Issue #453, opened 2016-12-05
helper tool for external middleware


miekg:
Caddy had `caddyext` as a tool for some time which was quite nice - we should probably create one for coredns.

Currently I'm not aware that Caddy has something other than the download system on <caddyserver.com> for getting middleware compiled into the binary, this makes development harder as well. Thus I think such a utility provides value.

<https://github.com/caddyserver/caddyext>
Plus this needs docs in `middleware.md`.

miekg:
Is this *actual* useful, or should we just document the process?

bwaz:
caddyext may lose support based on this forum post. 

> That tool will probably be phased out soon.
https://forum.caddyserver.com/t/caddys-source-not-found-on-any-gopath-directories/1118

miekg:
Yes, I know. But removing this tool is not helpful - I want something for CoreDNS to aid in developing and having external middleware - so not everything needs to be in the main repo.

bwaz:
This may work as a stop-gap solution for external middleware - please advise
We can setup a directives.yaml config file in the root directory 
```
...
-package "github.com/miekg/coredns/middleware/bind"
  -name "bind"
  -enabled true
-package "github.com/miekg/coredns/middleware/log"
  -name "log"
  -enabled true
...
```
core/cordns.go will be automatically created/updated on build-time to include only enabled packages
core/dnsserver/directives.go: var directives will be removed from this file,
it will be defined in a new file that is created at build-time as well.

I was successful in cloning a middleware  and moving it to a separate repo project, the build will pull from github when the package is defined in the coredns.go file.

miekg:
I like the idea (not sure about yaml :) ). Note a middleware also needs to be ordered.

But I like the fact that this can be kept separate from the source. A Go generate can automatically parse this and edit the source to make stuff work

bwaz:
This could be the middleware file - would generate the imports and define directives in the same order. 
The name of the package should be the same when caddy registering (I hope). We can ignore "#" lines. 
I'm assuming we should not import a package if it's not in the directives to make a smaller bin.
go:generate is not really needed in this case, we could just parse prior to build.

```
github.com/miekg/coredns/middleware/root
github.com/miekg/coredns/middleware/bind
github.com/miekg/coredns/middleware/trace
github.com/miekg/coredns/middleware/health
github.com/miekg/coredns/middleware/pprof
github.com/miekg/coredns/middleware/prometheus
github.com/miekg/coredns/middleware/errors
[...]

PR #463, opened 2016-12-21
WIP: gRPC Server Middleware

johnbelamaric:
Implements the server side of a client/server pair to allow
encrypted DNS over gRPC between two CoreDNS instances.

This is server-side only, not like the previous PR that was both client and server.

This still needs tests. Additionally, having a middleware that starts up another server that sits inside a server block that already defines a port seems weird:

~~~
.:53 {
  grpc :443 {
    tls cert.pem key.pem ca.pem
}
~~~

I would think it would look better like:

~~~
.:443 {
  grpc cert.pem key.pem ca.pem
}
~~~

And maybe this gRPC server should live alongside the existing udp/tcp server entries in the dnsserver.Server? Or replace them somehow?

Thoughts...?


johnbelamaric:
GitHub won't do dependent changes. The client code is dependent on this PR because it needs the TLS stuff and it needs the gRPC protobuf files. So, to see the client only changes from this one, look here:

https://github.com/johnbelamaric/coredns/pull/1

John

middleware/grpc/grpc.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (g *grpc) Query(ctx context.Context, in *pb.DnsPacket) (*pb.DnsPacket, error) {
+	msg := new(dns.Msg)
+	err := msg.Unpack(in.Msg)
+	if err != nil {
+		return nil, err
+	}
+
+	//TODO(jbelamaric): this is junk
```
miekg: what would be the idea when we replace this with non-junk?

middleware/grpc/grpc.go
```diff
@@ -0,0 +1,120 @@
...
+	r.Msg = m
+	return nil
+}
+
+func (r *response) Write(b []byte) (int, error) {
+	r.Msg = new(dns.Msg)
+	return len(b), r.Msg.Unpack(b)
+}
+
+func (r *response) Close() error {
```
miekg, line 112: you can put these on 1 line.

The probably needs some (bogus) comment of 'Implements the dns.ResponseWriter interface' to shut golint up

miekg (commented):
LGTM. The only biggy is maybe that we've just invented a new DNS protocol, or maybe DNS over gRPC. Might require an RFC :-)

miekg:
ah yes: tests...

middleware/grpc/grpc.go
johnbelamaric: Not sure. I don't fully understand what it is used for. I could put the local address that gRPC is listening on and the remote address of the gRPC connection, I think. I think I can get that info from the context...but what's it for? can these be nil?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] WIP: gRPC Serve..." ]
>
>> +	g.server = grpclib.NewServer()
>+	pb.RegisterDnsServiceServer(g.server, g)
>+	go func() {
>+		g.server.Serve(ln)
>+	}()
>+	return nil
>+}
>+
>+func (g *grpc) Query(ctx context.Context, in *pb.DnsPacket) (*pb.DnsPacket, error) {
>+	msg := new(dns.Msg)
>+	err := msg.Unpack(in.Msg)
>+	if err != nil {
>+		return nil, err
>+	}
>+
>+	//TODO(jbelamaric): this is junk
>
>Not sure. I don't fully understand what it is used for. I could put the local address that gRPC is listening on and the remote address of the gRPC connection, I think. I think I can get that info from the context...but what's it for? can these be nil?

This is getting a bit icky - if we want this we need to make grpc a first class
citizen - i.e. CoreDNS should do the parsing and then just give the `*dns.Msg`
to a `ServeDNS`

I'm like how this actually seems to work, but it somewhat of a layer violation
to do this in a middleware.

I'm also somewhat serious about writing a draft about this as it is exactly DNS
over TLS, but using gRPC as a transport.

miekg:
Doing this outside of a middleware might still be tricky, esp. since go DNS does not support grpc, so need to somehow know this came in via grpc and do the right thing when returns. Probably wrapping the ResponseWriter, with a gRPCResponseWriter that takes care of this.

johnbelamaric:
I think it would have to take over the TCP connection rather than have the standard DNS over TCP connection running. That way we could use the port from the main server definition as described in the comment above: https://github.com/miekg/coredns/pull/463#issue-197007201

I don't know that we could multiplex that connection.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] WIP: gRPC Serve..." ]
>I think it would have to take over the TCP connection rather than have the standard DNS over TCP connection running. That way we could use the port from the main server definition as described in the comment above: https://github.com/miekg/coredns/pull/463#issue-197007201

(can't reply to that commment). So I'll reply here:

> This still needs tests. Additionally, having a middleware that starts up another server that sits inside a server block that already defines a port seems weird:

yes, that is weird

>I don't know that we could multiplex that connection.

that's weird as well. You want something that tells what the protocol will be.

I.e. in the Corefile:

~~~
dns-tls://.:53 {...}

dns-grpc://.:53 { ...}
~~~

/Miek

--
Miek Gieben

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] WIP: gRPC Serve..." ]
>I think it would have to take over the TCP connection rather than have the standard DNS over TCP connection running. That way we could use the port from the main server definition as described in the comment above: https://github.com/miekg/coredns/pull/463#issue-197007201
>
>I don't know that we could multiplex that connection.

Might want to play with dns.DecorateReader to make grpc happen in a sane way on
the server side. No idea how viable that is though.

/Miek

--
Miek Gieben

johnbelamaric:
That looks good. Can you have multiple protocols though? For example udp and tcp.

---
John Belamaric
(240) 383-6963

On Dec 22, 2016, at 12:05 PM, Miek Gieben <notifications@github.com<mailto:notifications@github.com>> wrote:

[ Quoting <notifications@github.com<mailto:notifications@github.com>> in "Re: [miekg/coredns] WIP: gRPC Serve..." ]
>I think it would have to take over the TCP connection rather than have the standard DNS over TCP connection running. That way we could use the port from the main server definition as described in the comment above: https://github.com/miekg/coredns/pull/463#issue-197007201

(can't reply to that commment). So I'll reply here:

> This still needs tests. Additionally, having a middleware that starts up another server that sits inside a server block that already defines a port seems weird:

yes, that is weird

>I don't know that we could multiplex that connection.

that's weird as well. You want something that tells what the protocol will be.

I.e. in the Corefile:

~~~
dns-tls://.:53 {...}

dns-grpc://.:53 { ...}
~~~

/Miek

--
Miek Gieben

-

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] WIP: gRPC Serve..." ]
>That looks good. Can you have multiple protocols though? For example udp and tcp.

By default this always open a udp and tcp port.

johnbelamaric:
Ok, I have updated this to set the remote address and the local address when we can. I have NOT updated it to try and make gRPC a first-class citizen. It is still nested inside another server block.

I also update https://github.com/johnbelamaric/coredns/pull/1 for your review comments.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] WIP: gRPC Serve..." ]
>Ok, I have updated this to set the remote address and the local address when we can. I have NOT updated it to try and make gRPC a first-class citizen. It is still nested inside another server block.
>
>I also update https://github.com/johnbelamaric/coredns/pull/1 for your review comments.

Ack. Note that I'm very hesitant to merge the server part of this. Client side
proxy is OK (but I'll want to review it again)

johnbelamaric:
For the server side, what is the primary hesitation? That it should be a first-class citizen and not in a middleware?

FYI, the way it is written now the client PR is a patch on top of the server PR.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] WIP: gRPC Serve..." ]
>For the server side, what is the primary hesitation? That it should be a first-class citizen and not in a middleware?

Yes, that gets super-weird...OTOH doing a first-class citizen may involve
patching Go DNS itself (and I'm totally nacking that as gRPC dns is nowhere near
to a standard).

So yes, I wish there was a cleaner way of implementing this...

>FYI, the way it is written now the client PR is a patch on top of the server PR.

Ack. We should split this anyway IMHO - client side first and then discuss
serverside.

note: I'm well aware that this is nice to have (the serverside grpc stuff)

johnbelamaric:
The only thing is that without the server the client is useless :)

Anyway, you are right this was intended as just a prototype for now, there is no pressing need for it.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] WIP: gRPC Serve..." ]
>The only thing is that without the server the client is useless :)

Yes I know :-( To be fair, this could be the first example of an out-of-tree
middleware. We just lack an automated way to build a coredns binary without
fiddling with the source. See #453 - but I think I want something `go generate`
like, that will do all-the-things correctly during build time.

This could be a nice intermediate solution - which allows us to think/code what
we need be needed to make this a first class citizen.

(It does create a somewhat weird schism that the client side can speak grpc by
default and the serverside needs an external middleware).

/Miek

--
Miek Gieben

johnbelamaric:
Closing this in favor of https://github.com/infobloxopen/coredns-grpc

Still plan to include client though (and move .proto and pb files to CoreDNS) once #481 merges.

PR #464, opened 2016-12-21
handle A/PTR/SRV for headless services/endpoints

chrisohaver:
This is a complete refactor of the prior PR (which i recently cancelled), to minimize as much as possible the amount of code change.  I have only refactored existing code where absolutely necessary.

This adds...
* 'A' lookups for headless service targets (e.g. hostname.svc-name.namespace.cluster.local.)
* 'SRV' lookups for headless service
* 'SRV' lookups for normal services (prior implementation was wrong)
* 'PTR' lookups for headless service targets

There is some necessary refactoring to the way that we pass services around internally in k8s middleware.
Prior design assumed that all answers could be gleaned from api.Service objects.  However headless service endoints are stored in api.Endpoints objects.  Also, per port/protocol filtering requires finer resolution than a list of api.Service.  So new structs are added ...  `type service struct` and `type endpoint struct`

### DIFF
--- a/coremain/version.go
+++ b/coremain/version.go
@@ -2,7 +2,7 @@ package coremain
 
 const (
 	coreName    = "CoreDNS"
-	coreVersion = "003"
+	coreVersion = "004"
 
 	serverType = "dns"
 )
--- a/middleware/kubernetes/kubernetes.go
+++ b/middleware/kubernetes/kubernetes.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"regexp"
 	"strings"
 	"time"
 
@@ -64,18 +63,11 @@ var errInvalidRequest = errors.New("invalid query name")
 
 // Services implements the ServiceBackend interface.
 func (k *Kubernetes) Services(state request.Request, exact bool, opt middleware.Options) ([]msg.Service, []msg.Service, error) {
-
-	if state.Type() == "SRV" && !validSrvPrefix(state.QName()) {
+	if state.Type() == "SRV" && !ValidSRV(state.Name()) {
 		return nil, nil, errInvalidRequest
 	}
-
 	s, e := k.Records(state.Name(), exact)
 	return s, nil, e // Haven't implemented debug queries yet.
-}
-
-func validSrvPrefix(name string) bool {
-	matched, _ := regexp.MatchString("^_[^\\.]+\\._(tcp|udp|\\*|any)\\.", strings.ToLower(name))
-	return matched
 }
 
 // PrimaryZone will return the first non-reverse zone being handled by this middleware
@@ -195,10 +187,13 @@ func (k *Kubernetes) getZoneForName(name string) (string, []string) {
 	return zone, serviceSegments
 }
 
-func stripSrvPrefix(name []string) (port string, protocol string, nameOut []string) {
-	if strings.HasPrefix(name[0], "_") && strings.HasPrefix(name[1], "_") {
-		return strings.ToLower(name[0][1:]), strings.ToLower(name[1][1:]), name[2:]
+// stripSrvPrefix separates out the port and protocol segments, if present
+// If not present, assume all ports/protocols (e.g. wildcard)
+func stripSrvPrefix(name []string) (string, string, []string) {
+	if name[0][:1] == "_" && name[1][:1] == "_" {
+		return name[0][1:], name[1][1:], name[2:]
 	}
+	// no srv prefix present
 	return "*", "*", name
 }
 
@@ -430,4 +425,58 @@ func (k *Kubernetes) getServiceRecordForIP(ip, name string) []msg.Service {
 // symbolContainsWildcard checks whether symbol contains a wildcard value
 func symbolContainsWildcard(symbol string) bool {
 	return (strings.Contains(symbol, "*") || (symbol == "any"))
+}
+
+// ValidSRV parses a server record validating _port._proto. prefix labels.
+// The valid schema is:
+//   * Fist two segments must start with an "_",
+//   * Second segment must be one of _tcp|_udp|_*|_any
+func ValidSRV(name string) bool {
+
+	// Does it start with a "_" ?
+	if len(name) > 0 && name[0] != '_' {
+		return false
+	}
+
+	// First label
+	first, end := dns.NextLabel(name, 0)
+	if end {
+		return false
+	}
+	// Second label
+	off, end := dns.NextLabel(name, first)
+	if end {
+		return false
+	}
+
+	// first:off has captured _tcp. or _udp. (if present)
+	second := name[first:off]
+	if len(second) > 0 && second[0] != '_' {
+		return false
+	}
+
+	// A bit convoluted to avoid strings.ToLower
+	if len(second) == 5 {
+		// matches _tcp
+		if (second[1] == 't' || second[1] == 'T') && (second[2] == 'c' || second[2] == 'C') &&
+			(second[3] == 'p' || second[3] == 'P') {
+			return true
+		}
+		// matches _udp
+		if (second[1] == 'u' || second[1] == 'U') && (second[2] == 'd' || second[2] == 'D') &&
+			(second[3] == 'p' || second[3] == 'P') {
+			return true
+		}
+		// matches _any
+		if (second[1] == 'a' || second[1] == 'A') && (second[2] == 'n' || second[2] == 'N') &&
+			(second[3] == 'y' || second[3] == 'Y') {
+			return true
+		}
+	}
+	// matches _*
+	if len(second) == 3 && second[1] == '*' {
+		return true
+	}
+
+	return false
 }
--- a/middleware/kubernetes/kubernetes.go
+++ b/middleware/kubernetes/kubernetes.go
@@ -190,7 +190,7 @@ func (k *Kubernetes) getZoneForName(name string) (string, []string) {
 // stripSrvPrefix separates out the port and protocol segments, if present
 // If not present, assume all ports/protocols (e.g. wildcard)
 func stripSrvPrefix(name []string) (string, string, []string) {
-	if name[0][:1] == "_" && name[1][:1] == "_" {
+	if name[0][0] == '_' && name[1][0] == '_' {
 		return name[0][1:], name[1][1:], name[2:]
 	}
 	// no srv prefix present
--- a/middleware/kubernetes/kubernetes.go
+++ b/middleware/kubernetes/kubernetes.go
@@ -5,7 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"strconv"
+	"regexp"
 	"strings"
 	"time"
 
@@ -45,13 +45,37 @@ type Kubernetes struct {
 	Selector      *labels.Selector
 }
 
+type endpoint struct {
+	addr api.EndpointAddress
+	port api.EndpointPort
+}
+
+type service struct {
+	name      string
+	namespace string
+	addr      string
+	ports     []api.ServicePort
+	endpoints []endpoint
+}
+
 var errNoItems = errors.New("no items found")
 var errNsNotExposed = errors.New("namespace is not exposed")
+var errInvalidRequest = errors.New("invalid query name")
 
 // Services implements the ServiceBackend interface.
 func (k *Kubernetes) Services(state request.Request, exact bool, opt middleware.Options) ([]msg.Service, []msg.Service, error) {
+
+	if state.Type() == "SRV" && !validSrvPrefix(state.QName()) {
+		return nil, nil, errInvalidRequest
+	}
+
 	s, e := k.Records(state.Name(), exact)
 	return s, nil, e // Haven't implemented debug queries yet.
+}
+
+func validSrvPrefix(name string) bool {
+	matched, _ := regexp.MatchString("^_.*\\._.*\\..*", name)
+	return matched
 }
 
 // PrimaryZone will return the first non-reverse zone being handled by this middleware
@@ -77,7 +101,7 @@ func (k *Kubernetes) Lookup(state request.Request, name string, typ uint16) (*dn
 
 // IsNameError implements the ServiceBackend interface.
 func (k *Kubernetes) IsNameError(err error) bool {
-	return err == errNoItems || err == errNsNotExposed
+	return err == errNoItems || err == errNsNotExposed || err == errInvalidRequest
 }
 
 // Debug implements the ServiceBackend interface.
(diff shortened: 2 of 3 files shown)


middleware/kubernetes/kubernetes.go
```diff
@@ -45,15 +45,39 @@ type Kubernetes struct {
...
+	if state.Type() == "SRV" && !validSrvPrefix(state.QName()) {
+		return nil, nil, errInvalidRequest
+	}
+
 	s, e := k.Records(state.Name(), exact)
 	return s, nil, e // Haven't implemented debug queries yet.
 }
 
+func validSrvPrefix(name string) bool {
+	matched, _ := regexp.MatchString("^_.*\\._.*\\..*", name)
```
johnbelamaric: Second segment can only be _tcp or _udp, right?

chrisohaver: Practically speaking, yes in the sense that kubernetes only serves tcp and udp. However the RFC says ...

>... _TCP and _UDP are at present the most useful values for this field, though any name defined by Assigned Numbers or locally may be used (as for Service). ...

That said - its more efficient to short circuit and return an error before searching fruitlessly to produce the same result.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] handle A/PTR/SR..." ]
>+func validSrvPrefix(name string) bool {
>+	matched, _ := regexp.MatchString("^_.*\\._.*\\..*", name)
>
>Practically speaking, yes in the sense that kubernetes only serves tcp and udp. However the RFC says ...

We should probably create a small package for fiddling with labels of a domain
name and it should use dns.NextLabel en dns.PrevLabel or dns.Labels because
parsing domainnames is annoying (esp, esc. dots: \.)

Thanks for new PR Chris, appreciated.


/Miek

--
Miek Gieben

middleware/kubernetes/kubernetes.go
```diff
@@ -207,14 +247,16 @@ func (k *Kubernetes) Records(name string, exact bool) ([]msg.Service, error) {
...
+	protocolWildcard := symbolContainsWildcard(protocol)
 
 	// Abort if the namespace does not contain a wildcard, and namespace is not published per CoreFile
 	// Case where namespace contains a wildcard is handled in Get(...) method.
 	if (!nsWildcard) && (len(k.Namespaces) > 0) && (!dnsstrings.StringInSlice(namespace, k.Namespaces)) {
 		return nil, errNsNotExposed
 	}
 
-	k8sItems, err := k.Get(namespace, nsWildcard, serviceName, serviceWildcard, typeName)
+	k8sItems, err := k.Get(namespace, nsWildcard, serviceName, serviceWildcard, endpointname, port, portWildcard, protocol, protocolWildcard, typeName)
```
miekg: We should probably think of wrapping this stuff up in a struct that holds of of these parameters

middleware/kubernetes/kubernetes.go
```diff
@@ -227,40 +269,42 @@ func (k *Kubernetes) Records(name string, exact bool) ([]msg.Service, error) {
 	return records, nil
 }
 
-func (k *Kubernetes) getRecordsForServiceItems(serviceItems []*api.Service, zone string) []msg.Service {
+func endpointHostname(addr api.EndpointAddress) string {
+	if addr.Hostname != "" {
+		return strings.ToLower(addr.Hostname)
```
miekg, line 264: you return here, don't need to open an else block

middleware/kubernetes/kubernetes.go
```diff
@@ -227,40 +269,42 @@ func (k *Kubernetes) Records(name string, exact bool) ([]msg.Service, error) {
...
+	} else {
+		if strings.Contains(addr.IP, ".") {
+			return strings.Replace(addr.IP, ".", "-", -1)
+		}
+		if strings.Contains(addr.IP, ":") {
```
miekg: this will convert `fe80::00` to `fe80--00`, or is addr.IP always a complete v6 address? Also https://github.com/miekg/coredns/pull/452 has similar code, might be worth abstracting in the future.

middleware/kubernetes/kubernetes.go
```diff
@@ -270,32 +314,66 @@ func (k *Kubernetes) getRecordsForServiceItems(serviceItems []*api.Service, zone
...
-		return nil, fmt.Errorf("pod not implemented")
+		return nil, fmt.Errorf("%v not implemented", typeName)
 	default:
-		return k.getServices(namespace, nsWildcard, servicename, serviceWildcard)
+		return k.getServices(namespace, nsWildcard, servicename, serviceWildcard, endpointname, port, portWildcard, protocol, protocolWildcard)
 	}
 }
 
-func (k *Kubernetes) getServices(namespace string, nsWildcard bool, servicename string, serviceWildcard bool) ([]*api.Service, error) {
+func (k *Kubernetes) getServices(namespace string, nsWildcard bool, servicename string, serviceWildcard bool, endpointname string, port string, portWildcard bool, protocol string, protocolWildcard bool) ([]service, error) {
```
miekg: Wow, this 9 parameters :) We def. need some structure that holds these, maybe we can drop all those wildcard booleans into something like that?

miekg (commented):
Looks good. Nits and some questions.
the regexp we might want to leave to later - but still needs to be fixed at some point.

middleware/kubernetes/kubernetes.go
chrisohaver: addr.IP is the string IP exactly as returned to us from the kubernetes API.  So, we are trustting the data we get back from kubernetes to be correct.  Its not necessarily important that this value actually be a valid IP.  What matters is that the addr.Hostname (or if blank, the addr.IP) uniquely identifies the endpoint in the context of the service/namespace.  IPv6 collapsing doesnt jeopardize that.

middleware/kubernetes/kubernetes.go
chrisohaver: IMO, The wildcard params are clunky, and not really necessary to pass along through the cumbersome chain of private functions.  They can be computed where needed (at the end of the chain in getServices()), and still serve the purpose (which I presume was the efficiency of not having to check for wildcards at every comparison).

middleware/kubernetes/kubernetes.go
```diff
@@ -45,15 +45,39 @@ type Kubernetes struct {
...
+	if state.Type() == "SRV" && !validSrvPrefix(state.QName()) {
+		return nil, nil, errInvalidRequest
+	}
+
 	s, e := k.Records(state.Name(), exact)
 	return s, nil, e // Haven't implemented debug queries yet.
 }
 
+func validSrvPrefix(name string) bool {
+	matched, _ := regexp.MatchString("^_[^\\.]+\\._(tcp|udp|\\*|any)\\.", strings.ToLower(name))
```
miekg: do a case insensitive match so that you don't have to lowercase name and create garbage in the process.
(I'll write a parser that will do this for all cases - hopefully without allocating any memory)

miekg (commented):
Has some lingering comments.

middleware/kubernetes/kubernetes.go
miekg: https://gist.github.com/miekg/3f2d21e5c2cf74572d888626e1bf193e

Is a better way of doing this. Should prolly add such a thing to dnsutil

chrisohaver: So i can understand better:  What aspects of the original code create garbage... so i can avoid it in the future.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] handle A/PTR/SR..." ]
>
>>  	s, e := k.Records(state.Name(), exact)
> 	return s, nil, e // Haven't implemented debug queries yet.
> }
>
>+func validSrvPrefix(name string) bool {
>+	matched, _ := regexp.MatchString("^_[^\\.]+\\._(tcp|udp|\\*|any)\\.", strings.ToLower(name))
>
>So i can understand better:  What aspects of the original code create garbage... so i can avoid it in the future.

strings.ToLower creates a copy of `name` each time this is called, hence
"expensive".

/Miek

--
Miek Gieben

chrisohaver:
I have integrated your function - I had to modify slightly to fit. Changed to a bool function, added allowance for wildcards.

chrisohaver:
The kubernetes integration test passes for me, but not in Travis.  Any ideas?
Is there a way I can catch the travis CI instance to debug interactively?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] handle A/PTR/SR..." ]
>The kubernetes integration test passes for me, but not in Travis.  Any ideas?
>Is there a way I can catch the travis CI instance to debug interactively?

I wish :-( (travis is quite slow and hard to debug when stuff (only) fails
there. Wish we had something better...

I'll hope to take a look tomorrow (UTC) to see where I can help.

middleware/kubernetes/kubernetes.go
```diff
@@ -171,6 +187,23 @@ func (k *Kubernetes) getZoneForName(name string) (string, []string) {
 	return zone, serviceSegments
 }
 
+// stripSrvPrefix separates out the port and protocol segments, if present
+// If not present, assume all ports/protocols (e.g. wildcard)
+func stripSrvPrefix(name []string) (string, string, []string) {
```
miekg: Nit: Srv -> SRV to adhire to the naming convention for the other functions.

middleware/kubernetes/kubernetes.go
```diff
@@ -270,32 +303,71 @@ func (k *Kubernetes) getRecordsForServiceItems(serviceItems []*api.Service, zone
...
-		return nil, fmt.Errorf("pod not implemented")
+		return nil, fmt.Errorf("%v not implemented", typeName)
 	default:
-		return k.getServices(namespace, nsWildcard, servicename, serviceWildcard)
+		return k.getServices(namespace, servicename, endpointname, port, protocol)
 	}
 }
 
-func (k *Kubernetes) getServices(namespace string, nsWildcard bool, servicename string, serviceWildcard bool) ([]*api.Service, error) {
+func (k *Kubernetes) getServices(namespace string, servicename string, endpointname string, port string, protocol string) ([]service, error) {
```
miekg: nit: you don't need to repeat `string` for every parameter, only needed for the last one.
`namespace, servicename, endpointname, port, protocol string

test/kubernetes_test.go
```diff
@@ -105,105 +105,120 @@ var dnsTestCases = []test.Case{
...
+		Qname: "_*._*.svc-1-a.*.svc.cluster.local.", Qtype: dns.TypeSRV,
 		Rcode: dns.RcodeSuccess,
 		Answer: []dns.RR{
 			test.SRV("_http._tcp.svc-1-a.test-1.svc.cluster.local.      303    IN    SRV 10 100 80 svc-1-a.test-1.svc.cluster.local."),
 			test.SRV("_https._tcp.svc-1-a.test-1.svc.cluster.local.      303    IN    SRV 10 100 443 svc-1-a.test-1.svc.cluster.local."),
 		},
 	},
 	{
-		Qname: "svc-1-a.any.svc.cluster.local.", Qtype: dns.TypeSRV,
+		Qname: "_*._*.svc-1-a.any.svc.cluster.local.", Qtype: dns.TypeSRV,
```
miekg, line 129: Hmmm, I wasn't aware before; but this introduces new wildcard behavior, no? Now `_*` is *also* a wildcard?

miekg (commented):
I'll check on travis to see if I can fix the failure there.

miekg:
~~~
=== RUN   TestKubernetesIntegration
--- FAIL: TestKubernetesIntegration (5.02s)
	kubernetes_test.go:269: Could not send query: dns: buffer size too small
~~~
This is usually an indication that your proxying or returning a message when the caller is only expecting a standard 512 bytes one.

The diff. between travis and your env is prolly because travis re-downloads all deps on each run...

test/kubernetes_test.go
chrisohaver: Those are wild cards for the port/protocol request segments, which always start with an _.
Without the _ they would be considered part of the domain name.

chrisohaver:
I updated all dependencies and reproduced the error.  For now, I have reduced the SRV wildcard test cases to return different smaller subsets of services, instead of everything for every test.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] handle A/PTR/SR..." ]
>-		Qname: "svc-1-a.any.svc.cluster.local.", Qtype: dns.TypeSRV,
>+		Qname: "_*._*.svc-1-a.any.svc.cluster.local.", Qtype: dns.TypeSRV,
>
>Those are wild cards for the port/protocol request segments, which always start with an "_".
>Without the "_" they would be considered part of the domain name.

Ack. But SkyDNS only handles a single `*` as a wildcard... Are you saying 

 "_*._*.svc-1-a.any.svc.cluster.local." and "*.*.svc-1-a.any.svc.cluster.local." 
 or identical for CoreDNS, but only for SRV queries?

chrisohaver:
The port and protocol fields are identified by a leading _, followed by the value.
In the case of  " _*. " , the _ indicates its a port/protocol field.  The * is the value, which is a wildcard. 

 `"_*._*.svc-1-a.any.svc.cluster.local."` and `"*.*.svc-1-a.any.svc.cluster.local."` are not identical...

`"_*._*.svc-1-a.any.svc.cluster.local."` is a valid SRV query, because the port and protocol fields are present.
`"*.*.svc-1-a.any.svc.cluster.local."` is not a valid SRV query, because the port and protocol fields are not present.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] handle A/PTR/SR..." ]
>The port and protocol fields are identified by a leading _, followed by the value.
>In the case of  " _*. " , the _ indicates its a port/protocol field.  The * is the value, which is a wildcard.

Yes, but the `_` is just a letter, so the semantics of the wildcard have been
changed here. Just `*` should be enough for this to work and we shouldn't
special case `_*`.

miekg:
To be fair we can deal with this later.

coremain/version.go
```diff
@@ -2,7 +2,7 @@ package coremain
 
 const (
 	coreName    = "CoreDNS"
-	coreVersion = "003"
```
miekg, line 5: did you rebase?

chrisohaver:
I have not changed the wildcard semantics.

By RFC convention, all port and protocol fields begin with an underscore.  This is a literal underscore, not a mysql wildcard letter underscore.  The underscore identifies the field as a port/protocol field, differentiating it from the rest of the name string.  The underscore is not part of the value of a port or protocol field.

e.g. For the field `_http.` , the underscore identifies this as a port or protocol field, the value is "http".

The RFC does not cover wildcard values of course, but I have implemented them such that the _value_ can be a wildcard.  The leading underscore that identifies the field as a port/protocol field must still be present.

For the field `_*.` , the underscore identifies this as a port or protocol field, the value is "*", which is interpreted by CoreDNS as a wildcard under the established CoreDNS wildcard semantics.

miekg:
> For the field _*. , the underscore identifies this as a port or protocol field, the value is "*", which is interpreted by CoreDNS as a wildcard under the established CoreDNS wildcard semantics.

Ack, but CoreDNS only recognizes a single `*` (or `any`) as a wildcard, `_*` is *not* a wildcard in that sense.

coremain/version.go
chrisohaver: Yes - was i not supposed to? (i mean this genuinely - is there a different procedure I should follow spanning a release?)

miekg: Think it is OK. Just surprised this shows up as a diff after a rebase, but that maybe me not understanding git well enough.

PR #471, opened 2017-01-05
Add middleware/erratic

miekg:
This middleware allows playing with responses. Only one type is
implemented: it allows you to drop queries. I.e. withhold the response
from the client.

### DIFF
--- a/middleware/erratic/erratic.go
+++ b/middleware/erratic/erratic.go
@@ -1,0 +1,74 @@
+// Package erratic implements a middleware that returns erratic answers (delayed, dropped).
+package erratic
+
+import (
+	"sync/atomic"
+
+	"github.com/miekg/coredns/request"
+
+	"github.com/miekg/dns"
+	"golang.org/x/net/context"
+)
+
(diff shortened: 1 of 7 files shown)


middleware/erratic/erratic.go
```diff
@@ -0,0 +1,74 @@
...
+
+// ServeDNS implements the middleware.Handler interface.
+func (e *Erratic) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	state := request.Request{W: w, Req: r}
+
+	drop := false
+	if e.amount > 0 {
+		queryNr := atomic.LoadUint64(&e.q)
+
+		if queryNr%e.amount == 0 {
```
johnbelamaric, line 28: Nice! Should this be so uniform though? For example by default it drops every-other-query. I guess that's OK, but I wonder if there should be some randomness. I suppose if you have multiple clients it will appear essentially random, so it's probably not worth making it more complicated.

miekg:
Yeah, had those thoughts as well, but went for the most simple solution.
Upside of this is that testing is easier because of the non-randomness.

Issue #470, opened 2017-01-05
Clarify `*` (wildcard use) in CoreDNS


miekg:
SkyDNS introduced to ability to use a wildcard label in a query(!) to signify "I don't care about the value of this label". This was carried forward in CoreDNS.
While reviewing PR #464 I noticed the following being used as a wildcard label:
`Qname: "_*._*.bogusservice.test-1.svc.cluster.local.", Qtype: dns.TypeSRV`

Which adds `_*` as a wildcard label as well. We should see if this is needed and if so clarify the README and implement this in the *etcd* middleware as well.

miekg:
Comment from @chrisohaver 
> @miekg - When working with wild cards, i noticed some wild card unit tests which suggest that the prior implementation for * was incorrect. I have not changed the behavior. However, which of the following is the expected behavior for * wild cards?

> As implemented: One or more * present in a field will make that whole field wild, other characters ignored. e.g. delicious*taco* will match anything, including motor-oil
Or as one might expect wild cards to normally work: e.g. delicious*taco* matching delicious-beef-taco and delicious-taco-boat

chrisohaver:
Can we disable the comment markup?  It's super annoying.  The copy-pasted comment above is unintelligible...

chrisohaver:
The question is:
Should `*-apple` match `red-apple` and not `orange-orange`? This seems to be what one would expect.
However, as currently implemented, the presence of a `*` means that the entire field is wild, so `*-apple` is functionally equivalent to `*`, which means the expression matches both `red-apple` and `orange-orange`.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] Clarify `*` (wi..." ]
>The question is:
>Should `*-apple` match `red-apple` and not `orange-orange`?
>However, as currently implemented, the presence of a `*` means that the entire field is wild, so `*-apple` is functionally equivalent to `*`, which means the expression matches both `red-apple` and `orange-orange`.

The wildcard use should work exactly as the service side wildcard that we have
in the dns - so `*` will only work when it is the only thing in the label. The
only change made is that we allow multiple `*`-labels AFAICS.

Issue #458, opened 2016-12-14
middleware/kubernetes: remove name template feature


johnbelamaric:
This has very little utility and makes things a lot more complicated. Will plan to remove it in an upcoming release unless there are some strong objections here.

johnbelamaric:
@chrisohaver is this done now?

chrisohaver:
no

chrisohaver:
Removed and cleaned up in PR #472

PR #472, opened 2017-01-05
dont require/allow "_" prefix for srv wildcard fields

chrisohaver:
This addresses #470.  Do not require or allow SRV request underscore prefix in queries if the value is a wildcard.
Given a SRV request in the following format:
`_service._protocol.servicename.namespace.svc.cluster.local.`

The following requests are valid...
`_xyz._tcp.servicename.namespace.svc.cluster.local.`
`_xyz.*.servicename.namespace.svc.cluster.local.`
`*._tcp.servicename.namespace.svc.cluster.local.`
`*.*.servicename.namespace.svc.cluster.local.`

The following is not valid...
`_*._*.servicename.namespace.svc.cluster.local.`


middleware/kubernetes/kubernetes.go
```diff
@@ -190,11 +190,27 @@ func (k *Kubernetes) getZoneForName(name string) (string, []string) {
 // stripSRVPrefix separates out the port and protocol segments, if present
 // If not present, assume all ports/protocols (e.g. wildcard)
 func stripSRVPrefix(name []string) (string, string, []string) {
-	if name[0][0] == '_' && name[1][0] == '_' {
-		return name[0][1:], name[1][1:], name[2:]
+
+	var srv, prt string
+
+	if len(name) != 5 {
```
johnbelamaric: Are you stripping off the {zone} before doing these counts? While `cluster.local` is the usual one we use, people can use whatever they want - it could be `east.us.cloud.enterprise.com` or whatever.

chrisohaver: yes - name is perhaps a bad (or at least ambiguous) choice of variable name...  it does not contain the zone.  A length of 5 occurs when there is a service and protocol present.  In that case, it comprises _service, _protocol, service, namespace, svc

miekg: Would should assume anything about the lenght of name. This function (together with isValidSRV) should probably be moved to pkg/dnsutil (together with a few tests, etc).

middleware/kubernetes/kubernetes.go
```diff
@@ -432,49 +448,60 @@ func symbolContainsWildcard(symbol string) bool {
 //   * Fist two segments must start with an "_",
 //   * Second segment must be one of _tcp|_udp|_*|_any
 func ValidSRV(name string) bool {
```
miekg: IsValidSRV is even more idiomatic (used in go dns and in the std lib IIRC)

middleware/kubernetes/kubernetes.go
```diff
@@ -432,49 +448,60 @@ func symbolContainsWildcard(symbol string) bool {
...
+	// First label
+	first_ok := false
+
+	if  len(name) > 0 && name[0] == '_' {
```
miekg: Why does this function need changing?

middleware/kubernetes/kubernetes.go
```diff
@@ -432,49 +448,60 @@ func symbolContainsWildcard(symbol string) bool {
...
+	if len(second) == 5 && second[0] == '_' && (second[1] == 't' || second[1] == 'T') && 
+		(second[2] == 'c' || second[2] == 'C') && (second[3] == 'p' || second[3] == 'P') {
+		return true
+	}
+	// matches "_udp."
+	if len(second) == 5 && second[0] == '_' && (second[1] == 'u' || second[1] == 'U') && 
+		(second[2] == 'd' || second[2] == 'D') && (second[3] == 'p' || second[3] == 'P') {
+		return true
+	}
+	// matches "any."
```
miekg: We should probably *not* litter functions like these with `*` and `any` checks

middleware/kubernetes/kubernetes.go
chrisohaver: It needs to be changed to accept wildcard fields, which now do not have the SRV field underscore prefix.
Not allowing/requiring the standard SRV underscore prefixes for the SRV fields when the values are wild cards means we have to special-case the wildcard checks.

chrisohaver:
* collapse long repeated list parsed name query params into struct.
* streamline parse/validation of request name into one call (parseRequest) to avoid redundant re-parsing ... note that we don't actually need to check for uppercase characters because request.Request.Name() is already lower case.
* add unit tests for parseRequest 
* remove redundant wildcard value checks in symbolContainsWildcard
* change parse failures to return servfail instead of nxdomain

Not sure what happened there with the 11 changed files... they came in when i merged... which I may have done improperly...

chrisohaver:
@miekg - When working with wild cards, i noticed some wild card unit tests which suggest that the prior implementation for * was incorrect.  I have not changed the behavior.  However, which of the following is the expected behavior for * wild cards?

* As implemented:  One or more * present in a field will make that whole field wild, other characters ignored.  e.g. `delicious*taco*`  will match anything, including `motor-oil`
* Or as one might expect wild cards to normally work: e.g. `delicious*taco*`  matching `delicious-beef-taco` and `delicious-taco-boat`

chrisohaver:
To address #458, clean up remaining references to (now no longer used) NameTemplate.

middleware/kubernetes/kubernetes.go
```diff
@@ -165,88 +169,91 @@ func (k *Kubernetes) InitKubeCache() error {
...
+			segs = dns.SplitDomainName(lowerCasedName)
+			segs = segs[:len(segs)-dns.CountLabel(r.zone)]
 			break
 		}
 	}
 
-	return zone, serviceSegments
-}
+	offset := 0
+	if len(segs) == 5 {
```
miekg, line 206: should this be => 5 ?
(again scary about introducing arbitrary limits in code)

middleware/kubernetes/kubernetes_test.go
```diff
@@ -23,3 +23,119 @@ func TestSymbolContainsWildcard(t *testing.T) {
...
+	}
+	if expected := "svc"; r.typeName != expected {
+		t.Errorf("Expected parseRequest(%v) to get typeName == \"%v\". Instead got \"%v\".", query, expected, r.typeName)
+	}
+	if expected := "inter.webs.test"; r.zone != expected {
+		t.Errorf("Expected parseRequest(%v) to get zone == \"%v\". Instead got \"%v\".", query, expected, r.zone)
+	}
+
+	// Test invalid A request (with port protocol)
+	query = "_http._tcp.webs.mynamespace.svc.inter.webs.test."
```
miekg: could you add some malformed requests here? `_http._tcp` and just `.` and `_tcp.test.` maybe?

middleware/kubernetes/kubernetes_test.go
```diff
@@ -23,3 +23,119 @@ func TestSymbolContainsWildcard(t *testing.T) {
...
+func TestParseRequest(t *testing.T) {
+	k := Kubernetes{Zones: []string{"inter.webs.test"}}
+
+	// Test a valid SRV request
+	query := "_http._tcp.webs.mynamespace.svc.inter.webs.test."
+	r, e := k.parseRequest(query, "SRV")
+	if e != nil {
+		t.Errorf("Expected no error from parseRequest(%v). Instead got '%v'.", query, e)
+	}
+	if expected := "http"; r.port != expected {
```
miekg: simple for loop can prolly shorten this test code quite a bit

middleware/kubernetes/kubernetes.go
chrisohaver: `>` 5 fields would be invalid... there being extra fields ... if we allow them, we need to make rules for which ones we would ignore and how they would map to the expected ones.

miekg: Ok. Then I don't understand exactly how this work in k8s-land.

miekg:
lgtm, modulo a nit (and conflicts)

PR #473, opened 2017-01-06
middleware/proxy: make conn part of upstream host

miekg:
Move any connection related business to UpstreamHost. Define an new
interface 'Exchanger' that is used to talk to the upstream.


miekg:
fyi: @johnbelamaric this is changing/doing the same thing as you are doing in #463, but maybe more generic. Still WIP

middleware/proxy/dns.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/miekg/dns"
+)
+
+type dnsUpstream struct {
+	group *singleflight.Group
+}
+
+// ServeDNS does not satisfy middleware.Handler, instead it interacts with the upstream
+// and returns the respons or an error.
+func (d *dnsUpstream) Exchange(w dns.ResponseWriter, r *dns.Msg, address string) (*dns.Msg, error) {
```
johnbelamaric: Do we need the address? Shouldn't the UpstreamHost already know that? In the code below you pass host.Name, which obviously it already knows.

miekg: I could indeed set it in `*dnsUpstream` before calling this. Thanks!

Issue #465, opened 2016-12-21
tls middleware pkg


miekg:
There are multiple middleware creating a `*tls.Config` and code looks similar we should abstract this into middleware/pkg/tls

middleware doing this: *etcd*, *proxy* (soon)

johnbelamaric:
Just pushed this: https://github.com/johnbelamaric/coredns/pull/1 which does that for the most part. We could probably squeeze a little more out of parseUpstream and into pkg/tls if we really want to.

miekg:
Calling it done.

PR #474, opened 2017-01-09
Common tls

johnbelamaric:
Implements #465

### DIFF
--- a/middleware/tls/tls.go
+++ b/middleware/tls/tls.go
@@ -1,0 +1,53 @@
+package tls
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
+)
+
+// NewTLSConfig returns a TLS config that includes a certificate
+// Use for server TLS config or when using a client certificate
+// If caPath is empty, system CAs will be used
+func NewTLSConfig(certPath, keyPath, caPath string) (*tls.Config, error) {
+	cert, err := tls.LoadX509KeyPair(certPath, keyPath)
+	if err != nil {
+		return nil, fmt.Errorf("Could not load TLS cert: %s", err)
+	}
+
+	roots, err := loadRoots(caPath)
+	if err != nil {
+		return nil, err
+	}
+
+	return &tls.Config{Certificates: []tls.Certificate{cert}, RootCAs: roots}, nil
+}
+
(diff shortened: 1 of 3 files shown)


middleware/tls/tls.go
```diff
@@ -0,0 +1,81 @@
...
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
+)
+
+// NewTLSConfigFromArgs returns a TLS config based upon the passed
+// in list of arguments. Typically these come straight from the
+// Corefile
+func NewTLSConfigFromArgs(args []string) (*tls.Config, error) {
```
miekg: Think `args ...string` is nicer and conveys the intent slightly better. Also this should have some docs explaining the args....

middleware/tls/tls_test.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		t.Errorf("Failed to create TLSConfig: %s", err)
+	}
+}
+
+func TestNewTLSConfigFromArgs(t *testing.T) {
+	rmFunc, cert, key, ca := getPEMFiles(t)
+	defer rmFunc()
+
+	args := []string{}
```
miekg: with `...string` as the functions signature you don't need this hack.

johnbelamaric: Yeah, it's out of the latest version. I left it in for one but not all. Typical usage will be with args... being passed because it will come from the Corefile args.

johnbelamaric: (for some reason this comment is on an outdated version)

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] Common tls (#47..." ]
>(for some reason this comment is on an outdated version)

Yeah, I had dinner :) And then resumed the review. As said: LGTM

/Miek

--
Miek Gieben

miekg:
You're updating the caller in a seperate PR? (Would be my preference, as that keeps PRs small)

johnbelamaric:
Yes, that is the plan.

---
John Belamaric
(240) 383-6963

On Jan 10, 2017, at 3:00 AM, Miek Gieben <notifications@github.com<mailto:notifications@github.com>> wrote:


You're updating the caller in a seperate PR? (Would be my preference, as that keeps PRs small)

-

PR #475, opened 2017-01-09
Add insecure A records for pods

chrisohaver:
Add option for insecure A record responses for pods.
When a request comes in the following format: `1-2-3-4.namespace.pod.myzone.local.` an A record is returned for 1.2.3.4.  The IP is pulled directly from the request name, and is not verified/validated with k8s to be a valid ip.  This option will create an a record for any IP/namespace passed, even if the IP is not actually in the namespace. This is vulnerable to abuse: for example, it can allow a user to use a wild card certificate to establish a secure connection to a pod that is not actually in the namespace that the certificate is for. Hence the optioned is named "insecure"

### DIFF
--- a/middleware/kubernetes/README.md
+++ b/middleware/kubernetes/README.md
@@ -43,21 +43,36 @@ This is the default kubernetes setup, with everything specified in full:
         # Kubernetes data API resync period
         # Example values: 60s, 5m, 1h
         resyncperiod 5m
+		
         # Use url for k8s API endpoint
         endpoint https://k8sendpoint:8080
+		
         # The tls cert, key and the CA cert filenames
         tls cert key cacert
+		
         # Assemble k8s record names with the template
         template {service}.{namespace}.{type}.{zone}
+		
         # Only expose the k8s namespace "demo"
         namespaces demo
+		
         # Only expose the records for kubernetes objects
         # that match this label selector. The label
         # selector syntax is described in the kubernetes
         # API documentation: http://kubernetes.io/docs/user-guide/labels/
         # Example selector below only exposes objects tagged as
         # "application=nginx" in the staging or qa environments.
-        labels environment in (staging, qa),application=nginx
+        #labels environment in (staging, qa),application=nginx
+		
+		# The mode of responding to pod A record requests. 
+		# e.g 1-2-3-4.ns.pod.zone.  This option is provided to allow use of
+		# SSL certs when connecting directly to pods.
+		# Valid values: disabled, verified, insecure
+		#  disabled: default. ignore pod requests, always returning NXDOMAIN
+		#  insecure: Always return an A record with IP from request (without 
+		#            checking k8s).  This option is is vulnerable to abuse if
+		# 	         used maliciously in conjuction with wildcard SSL certs.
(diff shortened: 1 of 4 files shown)


middleware/kubernetes/README.md
```diff
@@ -43,21 +43,36 @@ This is the default kubernetes setup, with everything specified in full:
...
+        #labels environment in (staging, qa),application=nginx
+		
+		# The mode of responding to pod A record requests. 
+		# e.g 1-2-3-4.ns.pod.zone.  This option is provided to allow use of
+		# SSL certs when connecting directly to pods.
+		# Valid values: disabled, verified, insecure
+		#  disabled: default. ignore pod requests, always returning NXDOMAIN
+		#  insecure: Always return an A record with IP from request (without 
+		#            checking k8s).  This option is is vulnerable to abuse if
+		# 	         used maliciously in conjuction with wildcard SSL certs.
```
miekg, line 74: you haven't explained verified here...?

Also I think NXDOMAIN is not the best reply, proally SERVFAIL or a nodata reply?

middleware/kubernetes/kubernetes.go
```diff
@@ -42,8 +42,14 @@ type Kubernetes struct {
 	Namespaces    []string
 	LabelSelector *unversionedapi.LabelSelector
 	Selector      *labels.Selector
+	PodMode       string
 }
 
+const (
+	PodModeDisabled = "disabled" // default. pod requests are ignored
+	PodModeInsecure = "insecure" // ALL pod requests are answered without verfying they exist
```
miekg, line 50: There is no verified?

middleware/kubernetes/kubernetes.go
```diff
@@ -221,12 +233,9 @@ func (k *Kubernetes) Records(name string, exact bool) ([]msg.Service, error) {
...
 
-	// TODO: Implementation above globbed together segments for the serviceName if
-	//       multiple segments remained. Determine how to do similar globbing using
-	//		 the template-based implementation.
-	namespace = k.NameTemplate.NamespaceFromSegmentArray(serviceSegments)
-	serviceName = k.NameTemplate.ServiceFromSegmentArray(serviceSegments)
-	typeName = k.NameTemplate.TypeFromSegmentArray(serviceSegments)
+	serviceName = serviceSegments[0]
+	namespace = serviceSegments[1]
+	typeName = serviceSegments[2]
```
miekg, line 238: I don't see the code here, but I get uncomfortable when I see accesses to an array without bounds checking...

middleware/kubernetes/README.md
chrisohaver: verified is not yet implemented - hence not documented yet.

will change to servfail

middleware/kubernetes/kubernetes.go
chrisohaver: we abort with error if the length is < 3 a few lines above this code.

middleware/kubernetes/kubernetes.go
chrisohaver: verified is not yet implemented.

PR #476, opened 2017-01-10
Use common TLS parsing routine for etcd

johnbelamaric:
Change to use the new common routine, and update the docs to reflect
the different options for passing TLS configuration.


middleware/etcd/setup.go
```diff
@@ -12,6 +10,7 @@ import (
 	"github.com/miekg/coredns/middleware"
 	"github.com/miekg/coredns/middleware/pkg/dnsutil"
 	"github.com/miekg/coredns/middleware/pkg/singleflight"
+	mwtls "github.com/miekg/coredns/middleware/tls"
```
miekg: Just saw. Think we should move this under `middleware/pkg/tls`

miekg:
LGTM

middleware/etcd/setup.go
johnbelamaric: So how are the tests passing. WTF?

johnbelamaric: Oh, *should move*. I thought you were saying it *is already* under pkg. I will move it.

PR #480, opened 2017-01-13
middleware/proxy: implement Exchanger

miekg:
By defining and using an proxy.Exchanger interface we make the proxy
more generic and we can then fold back httproxy into proxy.

This overrides #463 and #473 and should make futures extensions rather
trivial

### DIFF
--- a/middleware/proxy/lookup.go
+++ b/middleware/proxy/lookup.go
@@ -11,9 +11,10 @@ import (
 	"github.com/miekg/dns"
 )
 
-// New create a new proxy with the hosts in host and a Random policy.
-func New(hosts []string) Proxy {
-	p := Proxy{Next: nil, Client: newClient()}
+// NewLookup create a new proxy with the hosts in host and a Random policy.
+func NewLookup(hosts []string) Proxy {
+	// TODO(miek): maybe add optional protocol parameter?
+	p := Proxy{Next: nil}
 
 	upstream := &staticUpstream{
 		from:        "",
@@ -21,7 +22,7 @@ func New(hosts []string) Proxy {
 		Policy:      &Random{},
 		Spray:       nil,
 		FailTimeout: 10 * time.Second,
-		MaxFails:    1,
+		MaxFails:    3,
 	}
 
 	for i, host := range hosts {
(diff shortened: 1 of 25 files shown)


middleware/proxy/lookup.go
```diff
@@ -84,7 +89,7 @@ func (p Proxy) lookup(state request.Request, r *dns.Msg) (*dns.Msg, error) {
 
 			atomic.AddInt64(&host.Conns, 1)
 
-			reply, backendErr := p.Client.ServeDNS(state.W, r, host)
+			reply, backendErr := host.Exchange(state)
 
 			atomic.AddInt64(&host.Conns, -1)
```
johnbelamaric, line 93: I think the increment/decrement of this belongs in the host.Exchange function.

middleware/proxy/metrics.go
```diff
@@ -12,7 +12,7 @@ import (
 var (
 	RequestDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: middleware.Namespace,
-		Subsystem: subsystem,
+		Subsystem: "proxy",
 		Name:      "request_duration_milliseconds",
 		Buckets:   append(prometheus.DefBuckets, []float64{50, 100, 200, 500, 1000, 2000, 3000, 4000, 5000, 10000}...),
```
johnbelamaric, line 17: Seem like really high numbers. I would think we would want at least 5 or 6 buckets in the < 20ms range.

middleware/proxy/lookup.go
miekg: Inclined to leave it here (for now). ATM all hosts will have a `Conns`.

middleware/proxy/metrics.go
miekg: Will do

miekg: misspoke, those buckets are in `prometheus.DefBuckets`

johnbelamaric: Ahh, I see. So this just adds longer time buckets since it's often going over the Internet in the proxy, I suppose.

miekg: Yeah, 8.8.8.8 has a default timeout of 5 seconds for instance.

PR #483, opened 2017-01-16
Add pod cache and verified pod responses

chrisohaver:
if pod mode is `verified` ...
* Init and start pods cache controller
* verify pod IP in cache  for pod A requests

### DIFF
--- a/middleware/kubernetes/kubernetes.go
+++ b/middleware/kubernetes/kubernetes.go
@@ -386,12 +386,17 @@ func (k *Kubernetes) findPods(namespace, podname string) (pods []pod, err error)
 	}
 
 	// PodModeVerified
-	podList, err := k.APIConn.podLister.List(labels.Everything())
+	objList, err := k.APIConn.podLister.Indexer.ByIndex(podIPIndex, ip)
 	if err != nil {
 		return nil, err
 	}
+
 	nsWildcard := symbolContainsWildcard(namespace)
-	for _, p := range podList {
+	for _, o := range objList {
+		p, ok := o.(*api.Pod)
+		if !ok {
+			return nil, errors.New("expected type *api.Pod")
+		}
 		// If namespace has a wildcard, filter results against Corefile namespace list.
 		if nsWildcard && (len(k.Namespaces) > 0) && (!dnsstrings.StringInSlice(p.Namespace, k.Namespaces)) {
 			continue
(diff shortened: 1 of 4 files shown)


middleware/kubernetes/kubernetes.go
```diff
@@ -384,9 +385,25 @@ func (k *Kubernetes) findPods(namespace, podname string) (pods []pod, err error)
 		return pods, nil
 	}
 
-	// TODO: implement cache verified pod responses
+	// PodModeVerified
+	podList, err := k.APIConn.podLister.List(labels.Everything())
```
johnbelamaric: Is there a way to index this list by IP? Seems like this is going to be relatively slow for a single pod-IP lookup.

chrisohaver: Probably? There is an indexing feature built in to the k8s api cache controller.  Not sure how it works yet. Will look into it.

chrisohaver:
Latest commit adds IP indexing for pods cache.
A couple of TODOs not directly related to this PR...
- The service cache is maintaining a namespace index that is not actually used anywhere.  It can probably be removed, and replaced with more appropriate index if necessary.
- Also, to reduce memory footprint, we can use our own structs which contain only the information we need, instead of caching the native kubernetes-client api structs in their entirety as currently implemented.

johnbelamaric:
LGTM. Does this also support PTR, or would that be separate? If it does, you should add tests for it.

chrisohaver:
I'll do PTR in a separate PR.  I think its a little complicated

PTRs would need to have a search order of preference.  e.g. ...

1. first search services  -> service.namespace.svc.zone.
2. then endpoints -> endpoint.service.namespace.svc.zone.
3. then pods -> pod-name.namespace.pod.zone.

What would we put as the pod-name for a pod PTR response?  The fake dashed IP? or the pod's hostname?

chrisohaver:
While poking around, I noticed that many PodIPs in the test environment are listed as loopback 127.0.0.1...
@johnbelamaric suggested these may be Host-Networked Pods.  If they are, then I should be validating the IP on the HostIP for those pods.  I'll investigate if this is the case.

chrisohaver:
OK - the loopback pods are actually from the system namespace, so although they get cached, they wont match any queries for other namespaces. So nothing to do here.

This raises a (minor?) issue: For all caches (service, endpoint, and pod caches), we are caching and watching all namespaces.  If a user is exposing a subset of all namespaces, then we are caching and watching the unexposed namespaces for no reason.  If one of the unexposed namespaces is a large and busy namespace, this would be a pointless drain of system resources.  IMO this doesn't need to be addressed in this PR.

johnbelamaric:
Ok, lgtm - @miekg, this good for you?

PR #485, opened 2017-01-18
truncated cache fix

kdima:
https://github.com/miekg/coredns/issues/484


miekg:
Adding Scrub here is a good idea, regardless.

PR #487, opened 2017-01-20
Add tracing option

johnbelamaric:
This adds an option to enable tracing with OpenTracing/OpenZipkin.
Enabling tracing will have a large impact on performance so it is
not advisable in production.

### DIFF
--- a/middleware/middleware.go
+++ b/middleware/middleware.go
@@ -71,8 +71,7 @@ func Error(name string, err error) error { return fmt.Errorf("%s/%s: %s", "middl
 // and a nil error.
 func NextOrFailure(name string, next Handler, ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 	if next != nil {
-		span := ot.SpanFromContext(ctx)
-		if span != nil {
+		if span := ot.SpanFromContext(ctx); span != nil {
 			child := span.Tracer().StartSpan(next.Name(), ot.ChildOf(span.Context()))
 			defer child.Finish()
 			ctx = ot.ContextWithSpan(ctx, child)
--- a/middleware/middleware.go
+++ b/middleware/middleware.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	"github.com/miekg/dns"
+        ot "github.com/opentracing/opentracing-go"
 	"golang.org/x/net/context"
 )
 
@@ -70,6 +71,12 @@ func Error(name string, err error) error { return fmt.Errorf("%s/%s: %s", "middl
 // and a nil error.
 func NextOrFailure(name string, next Handler, ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 	if next != nil {
+		span := ot.SpanFromContext(ctx)
+		if span != nil {
+			child := span.Tracer().StartSpan(next.Name(), ot.ChildOf(span.Context()))
+			defer child.Finish()
+			ctx = ot.ContextWithSpan(ctx, child)
+		}
 		return next.ServeDNS(ctx, w, r)
 	}
 
--- a/middleware/trace/README.md
+++ b/middleware/trace/README.md
@@ -6,23 +6,24 @@ middleware chain.
 ## Syntax
 
 ~~~
-trace [endpoint-type] [endpoint]
+trace [ENDPOINT-TYPE] [ENDPOINT]
 ~~~
 
 For each server you which to trace.
 
-It optionally the endpoint type and endpoint. The type defaults to Zipkin and the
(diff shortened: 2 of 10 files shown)


johnbelamaric:
@miekg let me know what you think of this. I didn't see any real impact to performance when it is disabled (though my testing was only with proxy module).

To try it out run Zipkin in docker:
```
```
Then when you run CoreDNS:

```
./coredns -conf Corefile -dns.trace zipkinhost:9411
```

Then you can navigate to the browser at zipkinhost:9411 to see the trace results.

miekg:
Ack. Currently AFK, but I assume you can just flip this on and off and
reload coredns?

Sounds useful although I haven't played with opentrace

johnbelamaric:
Yes, it is enabled with the `-dns.trace` option.

miekg:
I think for hot reloading it needs to be in the corefile?

johnbelamaric:
Ah, ok. Didn’t realize there was hot reloading. It’s not a middleware, and I didn’t see a way to do global settings in the Corefile. Right now it’s global but it could actually be per server - but two stanzas in the Corefile may apply to the same server so not sure how to handle that.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] Add tracing opt..." ]
>Ah, ok. Didn’t realize there was hot reloading. It’s not a middleware, and I didn’t see a way to do global settings in the Corefile. Right now it’s global but it could actually be per server - but two stanzas in the Corefile may apply to the same server so not sure how to handle that.
This could just be an option in the config of a server? Just like the *root*
directive?

johnbelamaric:
I could do that. But doesn't it gather all the zones on the same port into one server? So what would I do with this:

```
foo.com:53 {
  trace 1.2.3.4:9411
  log stdout
  errors
  proxy foo.com /etc/resolv.conf
}

foobar.com:53 {
  trace 9.9.9.9:12345
  log stdout
}
```

Don't those end up being service by the same dnsserver.Server?

I guess I can stick the tracer on the Config instead of the server, that would allow this to work as one would expect.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] Add tracing opt..." ]
>I could do that. But doesn't it gather all the zones on the same port into one server? So what would I do with this:
>
>```
>foo.com:53 {
>  trace 1.2.3.4:9411
>  log stdout
>  errors
>  proxy foo.com /etc/resolv.conf
>}
>
>foobar.com:53 {
>  trace 9.9.9.9:12345
>  log stdout
>}
>```
>
>Don't those end up being service by the same dnsserver.Server?
>
>I guess I can stick the tracer on the Config instead of the server, that would allow this to work as one would expect.

Yes

miekg:
Think this should be a middleware just like metrics (functionality is simular-ish anyway). Might need some bits in core, but those can be minimal, i guesss

johnbelamaric:
Ok

johnbelamaric:
Ok, it's a middleware now, that works fine. Also I put in the skeleton to allow other OpenTracing-compatible implementations other than Zipkin.

middleware/middleware.go
```diff
@@ -70,6 +71,12 @@ func Error(name string, err error) error { return fmt.Errorf("%s/%s: %s", "middl
 // and a nil error.
 func NextOrFailure(name string, next Handler, ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 	if next != nil {
+		span := ot.SpanFromContext(ctx)
+		if span != nil {
```
miekg: Nit: if statement can be done on 1 line.
`if span := ot.SpanFromContext(ctx); span != nil {`

middleware/trace/README.md
```diff
@@ -0,0 +1,46 @@
+# trace
+
+This module enables OpenTracing-based tracing of DNS requests as they go through the
+middleware chain.
+
+## Syntax
+
+~~~
+trace [endpoint-type] [endpoint]
```
miekg: please uppercase endpoint-type and endpoint to be inline with the other READMEs

middleware/trace/README.md
```diff
@@ -0,0 +1,46 @@
...
+~~~
+
+For each server you which to trace.
+
+It optionally the endpoint type and endpoint. The type defaults to Zipkin and the
```
miekg: `**ENDPOINT-TYPE**` and `**ENDPOINT**`

miekg:
lgtm. Please update; https://github.com/miekg/coredns/wiki/CoreDNS-005-Release
Maybe also add some works about the k8s middleware there?

johnbelamaric:
Ok, edits to release notes also done.

miekg (approved):

PR #488, opened 2017-01-20
only a * by itself is wild

chrisohaver:
Addressing issue #470:  Only a `*` by itself in a label is a wildcard.


johnbelamaric:
lgtm

PR #490, opened 2017-01-21
docs: update the zone specification

### LINKED ISSUE #489: Corefile weirdness
This doesn't do what I would expect. Either the last stanza should be a syntax error, or both domains should serve off of port 5354:

```
example.com:5000 {
	proxy . /etc/resolv.conf
}

foobar.com:5353 {
	log stdout
	proxy . /etc/resolv.conf
}

foo.com bar.com :5354 {
	log stdout
	errors
	proxy . /etc/resolv.conf
}
```

Instead, there is no error but the servers created are odd:

```
ubuntu@godev:~/go-work/src/github.com/miekg/coredns$ sudo ./coredns -conf Corefile-multi
example.com.:5000
foobar.com.:5353
foo.com.:53
bar.com.:53
.:5354
2017/01/20 18:22:46 [INFO] CoreDNS-004
CoreDNS-004
^C2017/01/20 18:22:47 [INFO] SIGINT: Shutting down
ubuntu@godev:~/go-work/src/github.com/miekg/coredns$
```

Or maybe I just don't quite understand how this is supposed to work?

miekg:
Add more words about how to specify the zone:port in a Corefile.

Fixes #489

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,13 +141,33 @@ nameserver *and* rewrite ANY queries to HINFO.
 }
 ~~~
 
+### Zone Specification
 
-## What Remains To Be Done
+The following Corefile fragment is legal, but does not explicitly define a zone to listen on:
 
-* Optimizations.
-* Load testing.
-* The [issues](https://github.com/miekg/coredns/issues).
+~~~ txt
+{
+   # ...
+}
+~~~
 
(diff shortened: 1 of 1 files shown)


johnbelamaric (approved):
Lgtm

PR #492, opened 2017-01-23
Fix panic caused by missing metric labels in proxy.

### LINKED ISSUE #491: kubernetes middleware port parsing breaks searches
The kubernetes middlware assumes all domains with 5 segments before the zone are of _port._protocol.service.namespace.type.zone format. it tests this assumption even for non-SRV requests, despite non-SRV having no such concept, and returns an SRVFAIL (not NXDOMAIN) when the check fails. as a consequence, the kubernetes search domains are broken as a lookup may accidentally match the 5-segment check.

as a simple example, running `apk update` in an alpine-based container breaks, as the `dl-cdn.alpinelinux.org` lookup becomes `dl-cdn.alpinelinux.org.default.svc.$ZONE`, matches the 5-segment check, and produces an error like
`23/Jan/2017:13:32:37 +0000 [ERROR 0 dl-cdn.alpinelinux.org.default.svc.$ZONE. A] srv port must start with an underscore or be a wildcard`. because this is a srvfail rather than nxdomain, the search stops there and the domain lookup fails.

devnev:


miekg:
Yeah, I fucked here :-( This will be fixed when https://github.com/miekg/coredns/pull/481 is merged (I backed out this change because I *thought* it was segfaulting).

Anyway using `Protocol()` like this is a bad idea, because `host` can be `nil` and then you don't know what the protocol should be. Also the `.From()` should give enough details because that identifies the proxy uniquely

devnev:
> Anyway using Protocol() like this is a bad idea, because host can be nil and then you don't know what the protocol should be.

It's using state.Proto(), not host.Protocol(). I.e. the protocol of the request.

> Also the .From() should give enough details because that identifies the proxy uniquely

 The Upstream's `Exchange` uses state.Proto() for its Dial, so that proto is still relevant, isn't it?

miekg:
a yes, sorry.

PR #493, opened 2017-01-23
correct k8s name parsing response codes

### LINKED ISSUE #491: kubernetes middleware port parsing breaks searches

chrisohaver:
Fixes #491.

When a query string is invalid for kubernetes, return NXDOMAIN, so that clients will continue to try other domains in the search list.

### DIFF
--- a/middleware/kubernetes/kubernetes.go
+++ b/middleware/kubernetes/kubernetes.go
@@ -225,7 +225,10 @@ func (k *Kubernetes) parseRequest(lowerCasedName, qtype string) (r recordRequest
 	}
 
 	offset := 0
-	if len(segs) == 5 {
+	if qtype == "SRV" {
+		if len(segs) != 5 {
+			return r, errInvalidRequest
+		}
 		// This is a SRV style request, get first two elements as port and
 		// protocol, stripping leading underscores if present.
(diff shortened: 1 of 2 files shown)


chrisohaver:
I fixed up the logic a bit too - so that we don't go down the SRV validation path for non-SRV requests (even if the outcome would be the same)

PR #496, opened 2017-01-30
middleware.cfg to configure middleware directives

bwaz:
This update will generate imports/directives based on middleware.cfg. 
(go generate is called in the Makefile before a build)

```
10:root:
20:bind:
30:trace:
40:health:
50:pprof:
60:prometheus:
70:errors:
80:log:
90:chaos:
100:cache:
110:rewrite:
120:loadbalance:
130:dnssec:
140:file:
150:auto:
160:secondary:
170:etcd:
180:kubernetes:
190:proxy:
200:httpprox:
210:whoami:
220:erratic:

```


bwaz:
Adding an external package is now easy. 
```
diff --git a/core/coredns-middleware.go b/core/coredns-middleware.go
index 38b2d52..271087f 100644
--- a/core/coredns-middleware.go
+++ b/core/coredns-middleware.go
@@ -1,4 +1,5 @@
 // generated by gen.go; DO NOT EDIT
 package core
 import (
+               _ "github.com/bwaz/hello"
 )
diff --git a/core/dnsserver/directives-config.go b/core/dnsserver/directives-config.go
index 7a7596d..e345f81 100644
--- a/core/dnsserver/directives-config.go
+++ b/core/dnsserver/directives-config.go
@@ -19,6 +19,7 @@ var directives = []string{
 "prometheus",
 "errors",
 "log",
+"hello",
 "chaos",
 "cache",
 "rewrite",
diff --git a/middleware.cfg b/middleware.cfg
index e164647..2abb43c 100644
--- a/middleware.cfg
+++ b/middleware.cfg
@@ -6,6 +6,7 @@
 60:prometheus:
 70:errors:
 80:log:
+85:hello:github.com/bwaz/hello
 90:chaos:
 100:cache:
 110:rewrite:

```

miekg:
Nice! Some nits:
1. the (sorta) defactor standard for generated file is to prefix them with `z`, so
`core/coredns-middleware.go` -> `core/zmiddleware.go`
2. maybe tiny overkill to give `gen.go` its own directory, just name `directives_generate.go`

And lastly; docs :)
Can `gen.go` ignore lines start with a `#` so we can put some docs in there as well?
IIRC there are probably other places where we reference how to add a middleware - could you update those as well. I'll take care of external refs. like on coredns.io etc.

Thanks!

bwaz:
1. done, 2. The build complains when there are more than one "main" found in a path. moving to an isolated path works.  3. Working on updating docs next , 4. will add # to parser ignore

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] middleware.cfg ..." ]
>1. done, 2. The build complains when there are more than one "main" found in a path. moving to an isolated path works.  3. Working on updating docs next , 4. will add # to parser ignore

awesome. Re 2: ack

bwaz:
Can we update the README.md to use 
`make gen && make`
?
I have issue following the instructions .

```
## Compilation

CoreDNS (as a servertype plugin for Caddy) has a dependency on Caddy, but this is not different than
any other Go dependency. If you have the source of CoreDNS, get all dependencies:

    go get ./...

And then `go build` as you would normally do:

    go build

This should yield a `coredns` binary.
```




```
rwasniak@rwasniak-VirtualBox:~/work/src/github.com/miekg/coredns$ go get ./...
package k8s.io/client-go/1.4/kubernetes: cannot find package "k8s.io/client-go/1.4/kubernetes" in any of:
	/usr/local/go/src/k8s.io/client-go/1.4/kubernetes (from $GOROOT)
	/mnt/sdb1/rwasniak/work/src/k8s.io/client-go/1.4/kubernetes (from $GOPATH)
package k8s.io/client-go/1.4/pkg/api: cannot find package "k8s.io/client-go/1.4/pkg/api" in any of:
	/usr/local/go/src/k8s.io/client-go/1.4/pkg/api (from $GOROOT)
	/mnt/sdb1/rwasniak/work/src/k8s.io/client-go/1.4/pkg/api (from $GOPATH)
package k8s.io/client-go/1.4/rest: cannot find package "k8s.io/client-go/1.4/rest" in any of:
	/usr/local/go/src/k8s.io/client-go/1.4/rest (from $GOROOT)
	/mnt/sdb1/rwasniak/work/src/k8s.io/client-go/1.4/rest (from $GOPATH)
package k8s.io/client-go/1.4/tools/clientcmd: cannot find package "k8s.io/client-go/1.4/tools/clientcmd" in any of:
	/usr/local/go/src/k8s.io/client-go/1.4/tools/clientcmd (from $GOROOT)
	/mnt/sdb1/rwasniak/work/src/k8s.io/client-go/1.4/tools/clientcmd (from $GOPATH)
rwasniak@rwasniak-VirtualBox:~/work/src/github.com/miekg/coredns$ 


```

bwaz:
```
## Compilation

CoreDNS (as a servertype plugin for Caddy) has a dependency on Caddy, but this is not different than
any other Go dependency. If you have the source of CoreDNS:

    make gen

To update the middleware directives and then `make` as you would normally do:

    make

This should yield a `coredns` binary.
```

bwaz:
@miekg  , Outside the scope of this ticket, some local mw can be identified as removable to build smaller binaries. We can then remove the imports from the core/coredns.go file and update the config to add as external even if it's in the local repo.

miekg:
The `gen` step should not be a normal part of the build - it's only useful for people willing to have external middleware - so I would remove that. Sorta ambivalent on the use of `make` here.

Yes, in the end all middleware is external and we default to the ones in the local coredns repo.

core/coredns.go
```diff
@@ -1,4 +1,7 @@
 // Package core registers the server and all plugins we support.
+
+// Additional middleware packages
+//go:generate go run ../gen/directives_generate.go ../middleware.cfg
```
miekg, line 4: don't we run it twice now? As there is also a go generate below?

miekg:
lgtm - 1 q on having 'go generate' twice

core/coredns.go
bwaz: directives_generate.go will generate the imports or directives based on the package go:generate was called from

miekg:
Thanks for your work!

PR #481, opened 2017-01-15
middleware/proxy: absorb httpproxy

miekg:

PR #498, opened 2017-01-31
Changes needed to support external gRPC server middleware

johnbelamaric:
The gRPC server middleware[1] needs access to the Server object
in order to push the unpacked Msg through the normal middleware
pipeline. These are the changes to core needed to make that
possible.

[1] https://github.com/infobloxopen/coredns-grpc


miekg:
Would be interesting to publish something blog.coredns.io about how to (ab)use this :)

johnbelamaric:
Still need the client, I am planning to clean it up once #481 merges.

PR #500, opened 2017-01-31
Allow cidr based reverse zone config

chrisohaver:
Instead of requiring declaration of in-addr.arpa style reverse zones for PTR lookups, allow simpler config in kubnenetes section.  This approach draws some, from the proposed PR middleware/reverse.

` kubernetes cluster.local 0.0.10.in-addr.arpa 0.17.172.in-addr.arpa {`
`                ...`
`}`

becomes...

`   kubernetes cluster.local {`
`                cidrs 172.17.0.0/24 10.0.0.0/24`
`                ...`
`}`

This allows for easier configuration of reverse ranges for cidrs that dont fall on an octet (e.g. /24 /16).
Existing style zone configs are still accepted.  Zone configurations take precedence (e.g. if both zones and cidrs are defined, the matching zone is always served in entirety, regardless of cidr configs).

@miekg, let me know if you are OK with this approach, and I'll add documentation and unit tests.

### DIFF
--- a/middleware/kubernetes/kubernetes.go
+++ b/middleware/kubernetes/kubernetes.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


middleware/kubernetes/kubernetes.go
```diff
@@ -128,6 +131,25 @@ func (k *Kubernetes) Reverse(state request.Request, exact bool, opt middleware.O
 	return records, nil, nil
 }
 
+func (k *Kubernetes) ExtractAddressFromReverse(state request.Request) string {
+	if state.Type() != "PTR" {
+		return ""
+	}
+	if k.extractAddressFromReverse == "" {
```
miekg: We get a new `*k` for each request? IOW: is this thread safe?

miekg:
lgtm

miekg:
if the thing was racy, I wonder why the tests didn't catch it btw.

johnbelamaric:
Oops, should have waited for test and docs. @chrisohaver to send those in a separate PR.

PR #504, opened 2017-02-06
 * Add script to generate manifest

johnbelamaric:


miekg:
Feel like we should have a separate repo for these kind of scripts.

PR #505, opened 2017-02-07
middleware/proxy: fix except keyword

### LINKED ISSUE #502: middleware/proxy: fix except keyword
This never worked, fix (copy code from Caddy?).

Mostly need on my home network where I forward all named (`.`) go Google via `https_google`, but some unqualified names need to be locally resolved. I.e. I (sadly) need this to work.

~~~
. {
   proxy . 1.2.3.4:53 {
      execpt pi gluon #some local hostnames
      protocol https_google
  }
  proxy pi 192.168.1.1:53
  proxy higgs 192.168.1.1:53
}

miekg:
Fix the except keyword usage - the config would allow it, but it was
not enforced in the code.
Turns out that **FROM** was also not enforced, fix both, by (basically)
copying the code from Caddy.

Update the README and tests.

Locally test as well, shows that this works:

~~~
.:1053 {
    proxy miek.nl 8.8.8.8:53 {
        except a.miek.nl
    }
    proxy a.miek.nl 8.8.4.4:53

    errors stdout
    log stdout
}
~~~

And gives the desired results, not having a proxy line for `a.miek.nl`
results in a SERVFAIL (as expected).

Fixes #502


johnbelamaric:
lgtm

Issue #503, opened 2017-02-06
kubernetes: PTR issue of some sort


johnbelamaric:
I have CoreDNS running in a test cluster and I am seeing an error on the client side when doing a reverse lookup and one on the server side when looking up something not in the default namespace (kube-dns.kube-system from a pod in default, for example).

Logs:
```
ubuntu@godev:~$ k logs --namespace kube-system coredns-243894951-qzbf1 --follow
2017/02/06 21:17:17 [INFO] Kubernetes middleware configured without a label selector. No label-based filtering will be performed.
.:53
2017/02/06 21:17:17 [INFO] CoreDNS-004
CoreDNS-004
10.2.6.75 - [06/Feb/2017:21:18:44 +0000] "A IN kubernetes.default.svc.cluster.local. udp 54 false 512" NOERROR 70 981.518µs
10.2.6.75 - [06/Feb/2017:21:18:44 +0000] "AAAA IN kubernetes.default.svc.cluster.local. udp 54 false 512" NOERROR 107 206.432µs
10.2.6.75 - [06/Feb/2017:21:18:44 +0000] "MX IN kubernetes.default.svc.cluster.local. udp 54 false 512" NOERROR 107 269.595µs
10.2.6.75 - [06/Feb/2017:21:18:46 +0000] "PTR IN 1.0.3.10.in-addr.arpa. udp 39 false 512" NOERROR 91 474.15µs
10.2.6.75 - [06/Feb/2017:21:23:11 +0000] "PTR IN 1.1.3.10.in-addr.arpa. udp 39 false 512" NXDOMAIN 115 4.210922ms
10.2.6.75 - [06/Feb/2017:21:23:24 +0000] "PTR IN 8.8.8.8.in-addr.arpa. udp 38 false 512" NOERROR 82 244.647573ms
10.2.6.75 - [06/Feb/2017:21:23:53 +0000] "A IN kube-dns.kube-system.default.svc.cluster.local. udp 64 false 512" NXDOMAIN 117 844.953µs
06/Feb/2017:21:23:53 +0000 [ERROR 0 kube-dns.kube-system.default.svc.cluster.local. A] no items found
10.2.6.75 - [06/Feb/2017:21:23:53 +0000] "A IN kube-dns.kube-system.svc.cluster.local. udp 56 false 512" NOERROR 72 702.577µs
[...]

PR #506, opened 2017-02-07
Fix no PTR records response for cidr ranges 

chrisohaver:
Partly addresses #503, the PTR malformed packet response.

The issue was specific to cidr based PTR responses, when there is no PTR record.
In this configuration, the authoritative range of IPs is defined in a cidr, not as a zone in the corefile config.
When a request comes in that is in range, but we don't have a record, we need to reply with what zone it does not exist.  In the case that the range is defined by a cidr, we'll claim authority for just the requested IP.


miekg:
lgtm

PR #510, opened 2017-02-08
Fix k8s chaining

chrisohaver:
I broke next middleware chaining from k8s in my last commit.  
Fixed here, and added chain test to appropriate integration tests.

### DIFF
--- a/middleware/kubernetes/handler.go
+++ b/middleware/kubernetes/handler.go
@@ -35,6 +35,8 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 			}
 			// Set the zone to this specific request.
 			zone = state.Name()
+		} else {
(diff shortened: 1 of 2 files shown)


middleware/kubernetes/handler.go
```diff
@@ -35,6 +35,8 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 			}
 			// Set the zone to this specific request.
 			zone = state.Name()
+		} else {
```
miekg: If you reverse the == "PTR" above like so it read better I think:
~~~
if state.Type() != "PTR" {
  return middleware.NextOrFailure(k.Name(), k.Next, ctx, w, r)
}
~~~
Then the rest can be dedented?

miekg:
One still nit.

LGTM

PR #511, opened 2017-02-08
Client-side of gRPC proxy

johnbelamaric:
This works with the out-of-tree gRPC server middleware here: https://github.com/infobloxopen/coredns-grpc.

Still need to do something about the context - would prefer to see that passed to Exchange()


middleware/proxy/grpc.go
```diff
@@ -0,0 +1,70 @@
...
+	err = d.Unpack(reply.Msg)
+	if err != nil {
+		return nil, err
+	}
+	return d, nil
+}
+
+func (g *grpcClient) Protocol() string { return "grpc" }
+
+func (g *grpcClient) OnShutdown(p *Proxy) error {
```
miekg, line 56: Nit: put on 1 line, and 1 line up, gofmt will then make it nice.

middleware/proxy/grpc.go
```diff
@@ -0,0 +1,70 @@
...
+	return nil
+}
+
+func (g *grpcClient) OnStartup(p *Proxy) error {
+	for _, host := range g.upstream.Hosts {
+		conn, err := grpc.Dial(host.Name, g.dialOpt)
+		if err != nil {
+			log.Printf("[WARNING] Skipping gRPC host '%s' due to Dial error: '%s'", host, err)
+		} else {
+			g.clients[host.Name] = pb.NewDnsServiceClient(conn)
```
miekg: Have you raced checked this? I think we're OK, because this is in the startup phase and thus sequential.

Also: there's no redial? *That* may trip you up and also coupled CoreDNS start to the backends which is a dependency you can do without IMHO.

miekg:
Some moaning. Love how short this PR has become since the recent changes in *proxy* :)

middleware/proxy/grpc.go
johnbelamaric: Yeah, I was thinking since it's in the startup phase it will only be called once, but I'll double-check.

Haven't figured out the redial and health check yet. I know the gRPC code will retry during dial but not sure if it will redial after a connection is lost.

miekg:
lgtm.

middleware/proxy/grpc.go
johnbelamaric: I don't think there is a race per se, because those are only called on startup and in sequence. However, I have been messing around with reloading the Corefile via USR1 signal. Definitely seeing oddness with this, so I think I need to work on the OnShutdown, etc. pieces.

It DOES redial though when the connection is lost; the grpc-go takes care of that for us.

middleware/proxy/upstream_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/mholt/caddy"
+	"github.com/miekg/coredns/middleware/test"
```
miekg: ordering

miekg:
re: context, yes makes sense to pass that.

middleware/proxy/upstream.go
```diff
@@ -197,6 +198,16 @@ func parseBlock(c *caddyfile.Dispenser, u *staticUpstream) error {
 			}
 
 			u.ex = newGoogle("", boot) // "" for default in google.go
+		case "grpc":
```
miekg, line 201: Think we should call this `grpc_infoblox` (or something).

johnbelamaric: Well, I moved the .proto into CoreDNS. That way coredns-grpc will depend on coredns instead of the other way around. It also means that anyone can implement a server (middleware or otherwise) using that .proto that is in CoreDNS. So it's not directly tied to what we're doing at Infoblox with coredns-grpc.

middleware/proxy/grpc.go
johnbelamaric: Ok, so here is the issue. The OnShutdown is not being called when I do a reload. This is because the port bind is failing, so the Restart method is bailing partway through:

```
2017/02/10 15:55:00 [INFO] Reloading
2017/02/10 15:55:00 [INFO] Creating gRPC connection 0 to 127.0.0.1:4443.
2017/02/10 15:55:00 [INFO] Creating gRPC connection 1 to 127.0.0.1:4444.
2017/02/10 15:55:00 [ERROR] SIGUSR1: listen tcp :1053: bind: address already in use
```

It seems the issue stems from this statement here: https://github.com/mholt/caddy/blob/master/caddy.go#L619 - it seems dnsserver.Server is NOT implementing the GracefulServer interface: https://godoc.org/github.com/mholt/caddy#GracefulServer

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] WIP: Client-sid..." ]
>
>```
>2017/02/10 15:55:00 [INFO] Reloading
>2017/02/10 15:55:00 [INFO] Creating gRPC connection 0 to 127.0.0.1:4443.
>2017/02/10 15:55:00 [INFO] Creating gRPC connection 1 to 127.0.0.1:4444.
>2017/02/10 15:55:00 [ERROR] SIGUSR1: listen tcp :1053: bind: address already in use
>```
>
>It seems the issue stems from this statement here: https://github.com/mholt/caddy/blob/master/caddy.go#L619 - it seems dnsserver.Server is NOT implementing the GracefulServer interface: https://godoc.org/github.com/mholt/caddy#GracefulServer

Ah :( Should fix then. Thought I did actually, but things might have moved on.

Thanks for finding. Also: tests (blame me)

johnbelamaric:
@miekg please take a look, I think it's ready

johnbelamaric:
oh, except the README

middleware/proxy/grpc.go
```diff
@@ -0,0 +1,79 @@
...
+		if err != nil {
+			log.Printf("[WARNING] Error closing connection %d: %s\n", i, err)
+		}
+	}
+	return nil
+}
+
+func (g *grpcClient) OnStartup(p *Proxy) error {
+	for i, host := range g.upstream.Hosts {
+		log.Printf("[INFO] Creating gRPC connection %d to %s.\n", i, host.Name)
```
miekg: Think we should be silent here.

middleware/proxy/upstream.go
miekg: ack

middleware/proxy/upstream_test.go
miekg: still ordering: move this up in its own section of the imports.

miekg (changes requested):
Missing the README.md updates for this?

miekg:
lgtm 
prolly will polish the README.md a bit, post merge.

miekg (approved):

middleware/proxy/README.md
```diff
@@ -40,7 +40,16 @@ proxy FROM TO... {
...
+  **TO** must be `dns.google.com`. The `grpc` option will talk to a server that has implemented the DnsService defined
+  in https://github.com/miekg/coredns/middleware/proxy/pb/dns.proto. An out-of-tree middleware that implements this can be
+  found at https://github.com/infobloxopen/coredns-grpc. Arguments to `protocol grpc` are as follows:
+  * None - No client authentication is used, and the system CAs are used to verify the server certificate.
+  * insecure - TLS is not used, the connection is made in plaintext (not good in production).
+  * CA-PEM - No client authentication is used, and the file CA-PEM is used to verify the server certificate.
+  * KEY-PEM CERT-PEM - Client authentication is used with the specified key/cert pair. The server certificate is verified
+    with the system CAs.
+  * KEY-PEM CERT-PEM CA-PEM - Client authentication is used with the specified key/cert pair. The server certificate is
+    verified using the CA-PEM file.
```
johnbelamaric: @miekg I am going to move all this explanation into the next section.

miekg: sgtm.

PR #516, opened 2017-02-09
middleware/reverse: random updates

miekg:
Make the documentation somewhat shorter (and hopefully clearer in the
process). Also to be on-par with the *auto* middleware, start counting
the referenced zones from 1 (instead of 0).
Some variable cleanups and use the NextOrFailure in the ServeDNS
function.


miekg:
@Project0 could you check if you agree with these changes? Thanks.

project0:
@miekg lgtm

PR #518, opened 2017-02-10
env `COREFILE` to setup coredns on docker

bwaz:
wrapper script to run coredns, pipe COREFILE if set.


miekg:
Note we already have `COREDNSPATH`. Not really used for anything. Maybe prefix everything with `CORE_` ? So `CORE_FILE` and `CORE_PATH`?

miekg:
oh, let's do this properly in Go and document it

PR #524, opened 2017-02-14
go gen improvements

miekg:
Remove the "gen" directory and move directives_generate.go out of it.
Add a build ignore tag so it isn't build by default. Cleanup the go ge
invocations so there are not seen as package docs.


miekg:
@bwaz I'm still fuzzy on why this 'go gen' needs to be called twice. Why is that?

bwaz:
go generate will provide the path and package name that where it was called from. So if the package layout changes the generated files will change package name and location path as well.  The imports and directives need to parse the same config file to provide output for each use.

miekg:
So the changes I've just pushed, will make `directives_generate.go` break somehow? I'm sorry, but I think I still don't see it. If you can show me some change/code that will break this PR I'm happy to rollback.

Makefile
```diff
@@ -50,7 +50,7 @@ clean:
 
 .PHONY: gen
 gen:
-	go generate ./core/...
+	go generate
```
bwaz: make sure this actually will generate the z files; delete them first and retry.

bwaz: I've tried `go generate ./... ` however this caused issues with vendor code - you could for now just explicitly generate the coredns.go

directives_generate.go
```diff
@@ -0,0 +1,120 @@
...
+					} else {
+						mi[items[1]] = middlewarePath + items[2]
+					}
+				}
+
+			}
+		}
+	}
+
+	var orders []int
```
bwaz: This is unused and redundant what is in the func.... I missed this earlier.

Makefile
miekg: done. Listed `coredns.go` explicitly.

directives_generate.go
miekg: removed.

Makefile
miekg: also removed the z* files and go genned - that works.

bwaz (approved):

PR #526, opened 2017-02-15
Fix make gen

johnbelamaric:
When using an external middleware, the generated file was failing to compile.
The issue is a typo in directives_generate.go that left imports with an
unterminated string.

### DIFF
--- a/directives_generate.go
+++ b/directives_generate.go
@@ -56,7 +56,7 @@ func genImports(file, pack string, mi map[string]string) {
 	}
 
 	for _, v := range mi {
-		outs += `_ "` + v + `\` + "\n"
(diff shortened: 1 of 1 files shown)


miekg:
of course :)

yongtang:
LGTM 👍

miekg:
Huh: --- FAIL: TestKubernetesIntegrationAllNSExposed (1.01s)
	kubernetes_test.go:415: Expected 0 answers but got 1 for query 123.0.0.10.in-addr.arpa., 12

johnbelamaric:
@chrisohaver can you take a look at that when you have a chance? Maybe it's a race or something?

PR #527, opened 2017-02-16
Trace improvements

johnbelamaric:


miekg:
lgtm, but this PR has some commits that don't belong? Like bbf6189?

johnbelamaric:
Yeah, something to do with the way I rebased it. I can clean it up.

PR #529, opened 2017-02-16
WIP:Edns0 rewrite for local codes

bwaz:
WIP for edns0, need to add tests documentation and improve the Corefile syntax.  PR for general discussions at this point.

### DIFF
--- a/middleware/rewrite/README.md
+++ b/middleware/rewrite/README.md
@@ -23,6 +23,10 @@ rewritten; e.g., to rewrite CH queries to IN use `rewrite class CH IN`.
 When the FIELD is `name` the query name in the message is rewritten; this
 needs to be a full match of the name, e.g., `rewrite name miek.nl example.org`.
 
+When the FIELD is `edns0` an EDNS0_LOCAL option can be appended to the request;
+e.g., to force a code on all requests, `rewrite edns0 replace_or_append 0xffad 55555555555555555555555555555555`;
+to replace present code values, `rewrite edns0 replace 0xffad 55555555555555555555555555555550`.
+
 If you specify multiple rules and an incoming query matches on multiple (simple) rules, only
 the first rewrite is applied.
 
--- a/middleware/rewrite/edns0.go
+++ b/middleware/rewrite/edns0.go
@@ -51,6 +51,11 @@ func (rule *Edns0Rule) SetEDNS0Attrs(r *dns.Msg) {
 
 // Initializer
 func (rule Edns0Rule) New(args ...string) Rule {
+	if len(args) < 3 {
+		log.Printf("[WARN] %s is invalid", args)
+		return rule
+	}
+
 	switch args[0] {
 	case "append":
 	case "replace":
--- a/middleware/rewrite/edns0.go
+++ b/middleware/rewrite/edns0.go
@@ -3,7 +3,7 @@ package rewrite
 
 import (
 	"encoding/hex"
-	"log"
+	"errors"
 	"strconv"
 
 	"github.com/miekg/dns"
@@ -51,10 +51,10 @@ func (rule *Edns0Rule) SetEDNS0Attrs(r *dns.Msg) {
 }
 
 // Initializer
-func (rule Edns0Rule) New(args ...string) Rule {
+func (rule Edns0Rule) New(args ...string) (Rule, error) {
+	var err error
 	if len(args) < 3 {
-		log.Printf("[WARN] %s is invalid", args)
-		return rule
+		return rule, errors.New("Wrong argument count")
 	}
 
 	switch args[0] {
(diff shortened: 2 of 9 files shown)


middleware/rewrite/edns0.go
```diff
@@ -0,0 +1,88 @@
...
+	code   uint16
+	data   []byte
+	match  []byte // TODO: add match input for replace rule
+}
+
+// SetEDNS0Attrs will alter the request
+func (rule *Edns0Rule) SetEDNS0Attrs(r *dns.Msg) {
+	o := r.IsEdns0()
+	if o == nil {
+		return
```
johnbelamaric: If we are doing append or set, then I would think we could be adding an EDNS0 opt record to a Msg that doesn't have one. So, I don't think we want to make this test here. Instead, we just want to skip the for loop below if o is nil.

middleware/rewrite/edns0.go
```diff
@@ -0,0 +1,88 @@
...
+		opt.Data = rule.data
+		o.Option = append(o.Option, opt)
+	}
+
+}
+
+// Initializer
+func (rule Edns0Rule) New(args ...string) Rule {
+	if len(args) < 3 {
+		log.Printf("[WARN] %s is invalid", args)
```
johnbelamaric: Instead of logging we should return an error.

miekg: New as a method is weird. Why is this needed?

miekg:
For discussion purposes the updates the README.md would be helpful as well.

middleware/rewrite/edns0.go
bwaz: I set Edns0 when nil, needs testing...

middleware/rewrite/edns0.go
bwaz: updated

johnbelamaric: I think Bob did that because then he sticks a new empty xyzRule in field.go and doesn't have to change the code in setup.go.

middleware/rewrite/setup.go
```diff
@@ -109,8 +109,13 @@ func rewriteParse(c *caddy.Controller) ([]Rule, error) {
...
 		default:
 			if _, ok := Fields[args[0]]; ok {
-				rule = Fields[args[0]].New(args[1:]...)
-				simpleRules = append(simpleRules, rule)
+				var err error
+				rule, err = Fields[args[0]].New(args[1:]...)
+				if err == nil {
+					simpleRules = append(simpleRules, rule)
+				} else {
+					log.Printf("[ERROR] %s, ignore %s", err, args)
```
johnbelamaric: Don't log, just return nil, err

middleware/rewrite/setup.go
```diff
@@ -109,8 +109,13 @@ func rewriteParse(c *caddy.Controller) ([]Rule, error) {
...
+				}
 			} else {
 				log.Printf("[WARN] %s is not a valid field, ignore %s", args[0], args)
```
johnbelamaric: Same here.

middleware/rewrite/README.md
```diff
@@ -23,6 +23,10 @@ rewritten; e.g., to rewrite CH queries to IN use `rewrite class CH IN`.
 When the FIELD is `name` the query name in the message is rewritten; this
 needs to be a full match of the name, e.g., `rewrite name miek.nl example.org`.
 
+When the FIELD is `edns0` an EDNS0_LOCAL option can be appended to the request;
```
miekg, line 26: What happens when I want to rewrite the EDNS0 bufsize and friends? Should would subdivide `edns0` in to `edns0 local` or something? Scrubbing stuff out of EDNS0 might also make sense (or at least the language should not preclude it)

middleware/rewrite/README.md
```diff
@@ -23,6 +23,10 @@ rewritten; e.g., to rewrite CH queries to IN use `rewrite class CH IN`.
...
+e.g., to force a code on all requests, `rewrite edns0 replace_or_append 0xffad 55555555555555555555555555555555`;
```
miekg, line 27: Just opinion, but I find `replace_or_append` annoyingly long to type...

johnbelamaric: Yeah. We couldn't think of a better name that captured the semantics. I guess we could change to DB-like terminology:

insert - append it
update - change it if it is there
upsert - made up word meaning update or insert if not found

middleware/rewrite/README.md
johnbelamaric: We can do that (`edn0 local`). FYI, Friday was @bwaz last day at Infoblox so someone else (maybe me?) will be wrapping this up.

johnbelamaric:
Replaced with #561

PR #533, opened 2017-02-17
ignore_serial option for the file middleware.

miekg:
WIP


miekg:
going to drop this for now. Will revive someday.

Issue #515, opened 2017-02-09
Rewrite middleware and fallthrough


miekg:
The rewrite middleware has `fallthrough` (at my request) to allows for falling through to the next middleware *even though* this middleware should be authoritative for this name.

q: should we keep something like `fallthrough` or just allow it? I'm not sure about the ramifications, but my gut feeling says that just allowing it can lead to weird to debug DNS issues... (and God forbid, maybe loops).
I'm also a fan of keeping things explicit; hence I think I'm in favor of keeping it.

johnbelamaric:
I think we should keep it. I'd like to add it to the k8s one too.

johnbelamaric:
Or maybe to etcd.

miekg:
Ack. I can live with that. *Maybe* we should elevate it to be part of the Handler interface, as a `Fallthrough` method? We can then put in the NextOrFailure or something.

miekg:
Best thing now is to document this behavior in `middleware.md`.

stp-ip:
Adding fallthrough in general as a highlevel option would enable workarounds like having multiple ```auto``` directives within one server. Basically abusing fallthrough and the proxy middleware.
5053 -> auto . {} + proxy -> 5054 -> auto . {} + proxy -> 5055 -> k8s + proxy -> public resolution.
Huge performance issue, but would enable various workarounds and enable testing various solutions in the wild (ignoring the performance hit) until we can implement or decide to implement it.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] Rewrite middlew..." ]
>Adding fallthrough in general as a highlevel option would enable workarounds like having multiple ```auto``` directives within one server. Basically abusing fallthrough and the proxy middleware.
>5053 -> auto . {} + proxy -> 5054 -> auto . {} + proxy -> 5055 -> k8s + proxy -> public resolution.
>Huge performance issue, but would enable various workarounds and enable testing various solutions in the wild (ignoring the performance hit) until we can implement or decide to implement it.

Not only a performance issue - what if your resolution fails, how do you debug
that? What about loops?

/Miek

--
Miek Gieben

stp-ip:
One could also produce loops with the rewrite fallthrough. Debugging could be harder, but we could help by writing to the log each time a request falls through.

PR #544, opened 2017-02-21
Grpc tracing

johnbelamaric:
Enables the tracing to pass through the gRPC client and into the server.

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -155,6 +155,10 @@ func (s *Server) Address() string { return s.Addr }
 // defined in the request so that the correct zone
 // (configuration and middleware stack) will handle the request.
 func (s *Server) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
+	s.ServeDNSWithContext(context.Background(), w, r)
+}
+
+func (s *Server) ServeDNSWithContext(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) {
 	defer func() {
 		// In case the user doesn't enable error middleware, we still
 		// need to make sure that we stay alive up here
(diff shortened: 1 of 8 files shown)


core/dnsserver/server.go
```diff
@@ -155,6 +155,10 @@ func (s *Server) Address() string { return s.Addr }
 // defined in the request so that the correct zone
 // (configuration and middleware stack) will handle the request.
 func (s *Server) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
+	s.ServeDNSWithContext(context.Background(), w, r)
+}
+
+func (s *Server) ServeDNSWithContext(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) {
```
miekg, line 161: This needs some docs. Also about where it is used IMO.

Also, I think we should add a TODO that we should find a better solution.

johnbelamaric:
FYI, here's the pretty kind of trace this enables. "dispatcher" and "grpcserver" are two different instances of CoreDNS.

coredns-grpc-trace

core/dnsserver/server.go
johnbelamaric: Ok. We can also wait for 007 on this, it's not a real rush or anything.

johnbelamaric: We could get a little crazy with this to, to help track down issues in fall through chains like described here https://github.com/coredns/coredns/issues/515. You could bundle the necessary trace data into an EDNS0 local option so you could actually trace through DNS calls not just gRPC.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Grpc tracing ..." ]
>FYI, here's the pretty kind of trace this enables. "dispatcher" and "grpcserver" are two different instances of CoreDNS.
>
>coredns-grpc-trace

Ack. Let's wait for 007, we do one relatively quickly, we don't have some
schedule (yet).

johnbelamaric:
@miekg any more thoughts on this one?

core/dnsserver/server.go
miekg: this all lgtm, with the note that we should go back and fix miekg/dns to make less intrusive

PR #545, opened 2017-02-21
Fix badge in README.md

yongtang:
This fix fixes badge in README.md so that it will correctly reflect github.com/coredns/coredns

Also, CodeCov and GoReportCard style has been changed to flat-square so that they are consistent with


yongtang:
@miekg The TravisCI is still not right. I think you may need to allow TravisCI to access CoreDNS.

You probably could find the setting in your "Personal Settings" -> "Authorized Applications".

Maybe you could also add me to CoreDNS team/organization.

PR #561, opened 2017-02-27
Rewrite edns0

johnbelamaric:
Taking over for Bob (@bwaz). Fix conflicts and go fmt.


johnbelamaric:
@miekg take a look and let me know what you think. I updated the README too. I haven't actually tested this live yet, just unit tests.

The `edns0` rule type now is more easily extended - in addition to `local` I did `NSID` just to show an example of something more.

I think at some point we will want complex rules, but this is a decent start.

middleware/rewrite/class.go
```diff
@@ -1,26 +1,33 @@
...
+type classRule struct {
 	fromClass, toClass uint16
 }
 
-// New initializes rule.
-func (rule ClassRule) New(args ...string) Rule {
-	from, to := args[0], strings.Join(args[1:], " ")
-	return &ClassRule{dns.StringToClass[from], dns.StringToClass[to]}
-
+func newClassRule(args ...string) (Rule, error) {
```
miekg: don't fully understand this, varargs but then check for enough params? (it's also internal, so it's not part of any API)

middleware/rewrite/edns0.go
```diff
@@ -0,0 +1,132 @@
...
+import (
+	"encoding/hex"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/miekg/dns"
+)
+
+// edns0LocalRule is a class rewrite rule.
```
miekg: class -> EDNS0?

middleware/rewrite/edns0.go
```diff
@@ -0,0 +1,132 @@
...
+	code   uint16
+	data   []byte
+	match  []byte // TODO: add match input for replace rule
+}
+
+type edns0NsidRule struct {
+	action string
+}
+
+func getEdns0Opt(r *dns.Msg) *dns.OPT {
```
miekg: This doesn't only get, but also set the edns0 option if not found. Needs better name + some function doc.

middleware/rewrite/edns0.go
```diff
@@ -0,0 +1,132 @@
...
+	// add option if not found
+	if !found && (rule.action == "append" || rule.action == "set") {
+		o.SetDo(true)
+		o.Option = append(o.Option, &dns.EDNS0_NSID{dns.EDNS0NSID, ""})
+	}
+
+	return RewriteDone
+}
+
+// SetEDNS0Attrs will alter the request
```
miekg: docs don't match function name: "Rewrite will set ..."

middleware/rewrite/edns0.go
```diff
@@ -0,0 +1,132 @@
...
+		opt.Data = rule.data
+		o.Option = append(o.Option, &opt)
+	}
+
+	return RewriteDone
+}
+
+// Initializer
+func newEdns0Rule(args ...string) (Rule, error) {
+	if len(args) < 2 {
```
miekg: some comment here as above.

middleware/rewrite/edns0.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, fmt.Errorf("Too few arguments for an EDNS0 rule")
+	}
+
+	action := strings.ToLower(args[0])
+	switch action {
+	case "append":
+	case "replace":
+	case "set":
+	default:
+		return nil, fmt.Errorf("invalid action: '%s'", action)
```
miekg: '%s' -> %q

middleware/rewrite/edns0.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func newEdns0LocalRule(action, code, data string) (*edns0LocalRule, error) {
+	c, err := strconv.ParseUint(code, 0, 16)
+	if err != nil {
+		return nil, err
+	}
+
+	decoded := []byte(data)
+	if strings.HasPrefix(data, "0x") {
```
miekg, line 133: should this error if it is not hex?

middleware/rewrite/edns0.go
```diff
@@ -0,0 +1,132 @@
...
+			return nil, fmt.Errorf("EDNS0 local rules require exactly three args")
+		}
+		return newEdns0LocalRule(action, args[1], args[2])
+	case "nsid":
+		if len(args) != 2 {
+			return nil, fmt.Errorf("EDNS0 NSID rules do not accept args")
+		}
+		return &edns0NsidRule{action: action}, nil
+	default:
+		return nil, fmt.Errorf("Invalid rule type '%s'", args[1])
```
miekg: Invalid -> invalid
'%s' -> %q

middleware/rewrite/edns0.go
```diff
@@ -0,0 +1,132 @@
...
+
+	decoded := []byte(data)
+	if strings.HasPrefix(data, "0x") {
+		decoded, err = hex.DecodeString(data[2:])
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return &edns0LocalRule{action, uint16(c), decoded, nil}, nil
```
miekg: Please use field named in struct literals

middleware/rewrite/name.go
```diff
@@ -1,26 +1,26 @@
...
-type NameRule struct {
+type nameRule struct {
 	From, To string
 }
 
-// New initializes a new rule.
-func (rule NameRule) New(args ...string) Rule {
-	from, to := args[0], strings.Join(args[1:], " ")
-	return &NameRule{middleware.Name(from).Normalize(), middleware.Name(to).Normalize()}
+func newNameRule(args ...string) (Rule, error) {
```
miekg: varargs

middleware/rewrite/rewrite.go
```diff
@@ -52,10 +56,27 @@ func (rw Rewrite) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
...
+	case "name":
+		return newNameRule(args[1:]...)
+	case "class":
+		return newClassRule(args[1:]...)
+	case "type":
+		return newTypeRule(args[1:]...)
+	case "edns0":
+		return newEdns0Rule(args[1:]...)
+	default:
+		return nil, fmt.Errorf("Invalid rule type '%s'", args[0])
```
miekg: Lowercase invalid, %q to quote the string

middleware/rewrite/rewrite_test.go
```diff
@@ -16,14 +17,105 @@ func msgPrinter(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, err
...
 }
 
+func TestNewRule(t *testing.T) {
+	r, err := newRule()
+	if err == nil {
+		t.Errorf("Expected error but got success for newRule()")
+	}
+	r, err = newRule("foo")
+	if err == nil {
+		t.Errorf("Expected error but got success for newRule(foo)")
```
miekg: Can't we add a Stringer for rules, so the error becomes: `"... %q", r)`

middleware/rewrite/rewrite_test.go
```diff
@@ -16,14 +17,105 @@ func msgPrinter(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, err
 	return 0, nil
 }
 
+func TestNewRule(t *testing.T) {
+	r, err := newRule()
```
miekg: Some looping of a slice of rules makes this test function more readable.

middleware/rewrite/setup_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"testing"
+
+	"github.com/mholt/caddy"
+)
+
+func TestParse(t *testing.T) {
+	c := caddy.NewTestController("dns", `rewrite`)
+	_, err := rewriteParse(c)
+	if err == nil {
+		t.Errorf("Expected error but found nil for `rewrite`")
```
miekg, line 13: Note `t.Error` is enough in this case. Think 'go vet' says this.

middleware/rewrite/type.go
```diff
@@ -2,24 +2,34 @@
...
-	from, to := args[0], strings.Join(args[1:], " ")
-	return &TypeRule{dns.StringToType[from], dns.StringToType[to]}
+func newTypeRule(args ...string) (Rule, error) {
+	if len(args) != 2 {
+		return nil, fmt.Errorf("Type rules must have exactly two arguments")
+	}
+	var from, to uint16
+	var ok bool
+	if from, ok = dns.StringToType[strings.ToUpper(args[0])]; !ok {
+		return nil, fmt.Errorf("Invalid type '%s'", strings.ToUpper(args[0]))
```
miekg: Lowercase 'Invalid' as it's an error

miekg (changes requested):
Logic looks OK.

Number of nits and comments

middleware/rewrite/class.go
johnbelamaric: Yeah. I went back and forth on it. I either have to put the checks that go with each rule type in the "newXXRule" or I have to put them in newRule. I can move them to newRule if that's more idiomatic.

middleware/rewrite/edns0.go
johnbelamaric: This one still needs varargs, because it will create different EDNS0 rules depending on the args, and those need different number of args. However, they all require at least 2 args: the action, and the EDNS0 option type.

middleware/rewrite/edns0.go
johnbelamaric: No, you can do `0x61` or just `a`, for example.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: Rewrite ..." ]
>+func newClassRule(args ...string) (Rule, error) {
>
>Yeah. I went back and forth on it. I either have to put the checks that go with each rule type in the "newXXRule" or I have to put them in newRule. I can move them to newRule if that's more idiomatic.

Think naming wise keeping it like this makes more sense.

johnbelamaric:
Ok, well, I made something of a change there. It actually turned out to be better as I could put all the checks for those rule types in a single if statement.

johnbelamaric:
Still need live testing/integration tests, so still WIP

johnbelamaric:
Ok, found some typos in README and also need to properly return RewriteIgnored when appropriate. Getting there.

johnbelamaric:
@miekg I think this is finally ready.

miekg (approved):

PR #564, opened 2017-03-01
Fix resolving CNAME with no proxy

miekg:
This fixes a crash when we resolve (or try to) an external CNAME
when no proxy is set.

Add test as well.

### DIFF
--- a/test/file_cname_proxy_test.go
+++ b/test/file_cname_proxy_test.go
@@ -4,7 +4,6 @@ import (
 	"io/ioutil"
 	"log"
 	"testing"
-	"time"
 
 	"github.com/coredns/coredns/middleware/proxy"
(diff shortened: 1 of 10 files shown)


test/file_cname_proxy_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/miekg/dns"
+)
+
+func TestZoneExternalCNAMELookup(t *testing.T) {
+	t.Parallel()
+	log.SetOutput(ioutil.Discard)
+
+	name, rm, err := TempFile(".", exampleOrg)
+	if err != nil {
+		t.Fatalf("Failed to created zone: %s", err)
```
0x333333: Small typo: `created` => `create` 😆

0x333333 (approved):
Thanks for the fix!

PR #570, opened 2017-03-05
Add `go lint` check and fix several lint issues

yongtang:
This fix updates the Makefile to add the `go lint` check for the build. This fix also fixes several go lint issues.

NOTE: This fix does not enforce `go lint` (suggestion only). This fix also ignores the `go lint` error:
```
middleware/middleware.go:72:1: context.Context should be the first parameter of a function
```
as it requires too many changes in API.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,23 +7,27 @@ all: coredns
 # Phony this to ensure we always build the binary.
 # TODO: Add .go file dependencies.
 .PHONY: coredns
-coredns: deps core/zmiddleware.go core/dnsserver/zdirectives.go
+coredns: check core/zmiddleware.go core/dnsserver/zdirectives.go
 	go build $(BUILD_VERBOSE) -ldflags="-s -w"
 
 .PHONY: deps
-deps: fmt
+deps:
 	go get ${BUILD_VERBOSE}
+	go get -u github.com/golang/lint/golint
+
+.PHONY: check
+check: fmt deps
 
 .PHONY: test
-test: deps
+test: check
 	go test -race $(TEST_VERBOSE) ./test ./middleware/...
 
 .PHONY: testk8s
-testk8s: deps
+testk8s: check
 	go test -race $(TEST_VERBOSE) -tags=k8s -run 'TestKubernetes' ./test ./middleware/kubernetes/...
 
 .PHONY: coverage
-coverage: deps
+coverage: check
 	set -e -x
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -52,6 +56,11 @@ fmt:
 	@test -z "$$(gofmt -s -l . | grep -v vendor/ | tee /dev/stderr)" || \
 		(echo "please format Go code with 'gofmt -s -w'" && false)
 
+.PHONY: lint
+lint: deps
+	## run go lint
+	@test -z "$$(golint ./... | grep -v vendor/ | grep -v ".pb.go:" | grep -vE "context\.Context should be the first parameter of a function" | tee /dev/stderr)"
```
miekg: This is I think the reason that we can't always enfore this, something `go lint` is too pendantic.

Makefile
```diff
@@ -7,23 +7,27 @@ all: coredns
 # Phony this to ensure we always build the binary.
 # TODO: Add .go file dependencies.
 .PHONY: coredns
-coredns: deps core/zmiddleware.go core/dnsserver/zdirectives.go
+coredns: check core/zmiddleware.go core/dnsserver/zdirectives.go
```
miekg, line 10: see other comment, I would not always enfore this.

yongtang:
Thanks @miekg. I updated the PR. Now `go lint` check is only suggestion only and is not a dependency of `make`.

It is still possible to invoke `make lint` to check for any go lint issues (after filtering out `vendor`, `pb.go`, and `Context` output).

Makefile
```diff
@@ -7,23 +7,27 @@ all: coredns
...
 	go build $(BUILD_VERBOSE) -ldflags="-s -w"
 
 .PHONY: deps
-deps: fmt
+deps:
 	go get ${BUILD_VERBOSE}
+	go get -u github.com/golang/lint/golint
+
+.PHONY: check
+check: fmt deps
```
yongtang, line 19: `check` now only requires `fmt`.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add `go lint`..." ]
>Thanks @miekg. I updated the PR. Now `go lint` check is only suggestion only and is not a dependency of `make`.
>
>It is still possible to invoke `make lint` to check for any go lint issues (after filtering out `vendor`, `pb.go`, and `Context` output).

lgtm

PR #574, opened 2017-03-06
core: add more transports

miekg:
Allow CoreDNS to listen for TLS request coming over port 853. This can
be enabled with `dns-tls` in the config file. Currently you can't
disable normal dns traffic, i.e. udp/tcp on the specified port.


miekg:
@johnbelamaric this is my new plan for supporting gRPC (and dns over TLS). This involves another hack in go DNS (https://github.com/miekg/dns/issues/464#issuecomment-284388274), but I think I'm fine with this.

This is of course WIP, and I still need to figure out what the cleanest way is to create a TLS listener (same holds true for gRPC).

Merging something like this mean ServeDNSWithContext() dies again, because Go DNS does not do that

miekg:
@johnbelamaric completely reworked :) I'm adding a new Server that handles TLS connections. The exact same thing can be done for gRPC. That will probably need its own `dns.ServeMux` implement to have a Handler that can read and write gRPC.
We should have our own Muxer anyway because the one from Go DNS redo's the work we're also doing - we can just call 1 entry point. See #575 for that.

I believe this is the cleanest way to support gRPC (and any other protocol that look like DNS)

README.md
```diff
@@ -169,6 +173,14 @@ example.org {
 }
 ~~~
 
+Listening on TLS and for gRPC? Use:
+
+~~~ txt
+tls://example.org grpc://example.org {
+    # ...
+}
+~~
```
johnbelamaric: Missing a ~

README.md
```diff
@@ -169,6 +173,14 @@ example.org {
 }
 ~~~
 
+Listening on TLS and for gRPC? Use:
+
+~~~ txt
+tls://example.org grpc://example.org {
```
johnbelamaric, line 181: Does this generalize to DNS too? Something like:

```
tls://example.org dns://example.org {
...
}
```

I guess I am trying to ask if we are differentiating between the network transport and the application protocol here, and how that is set up. This syntax makes me think this would be possible (but I don't think it can be):

```
tls:. dns://example.org grpc://example.com {
}
```

So I think a different syntax is better, one that separates layer 4 and layer 7 better. Maybe

```
tls:grpc://example.org {
}

tcp:dns://example.com {
}

tcp:grpc://example.net {
}
```

But we also need to be able to specify the addresses and ports in the above syntax, and for TLS we need a directive to specify the certs for the server. I am thinking we may be trying to put too much into the initial args of the server block. Instead, there could be server-level directives that are not necessarily implemented as middleware.

```
example.org:53 {
  protocol udp:dns tcp:dns
}

example.org:443 {
  protocol tls:grpc [....tls params...]
}

example.com:443 {
  protocol tls:dns [...tls params... - but they have to be the same for all servers on this port...]
}
```

middleware/tls/README.md
```diff
@@ -0,0 +1,13 @@
+# tls
```
johnbelamaric, line 1: ok, i see the tls stuff is here so no need for it like I talked about above. I still think though if you leave the server definition as just domain:hostport we are better off. Otherwise you could do dns-tls and grpc for example and they would look like different server blocks but in fact are on the same port and so need to share the TLS info.

johnbelamaric:
This comment: "That will probably need its own `dns.ServeMux` implement to have a Handler that can read and write gRPC."

I am wondering if we can avoid that. The gRPC piece is built on protobuf, so that DNS format is bound to a specific proto file. It seems too application-y to embed those proto definitions in go-dns. So, as long as CoreDNS can implement its own ServeMux and hand that into go-dns I think this works. That's probably what you were intending anyway. I would like that handler to be able to decode the context out of the gRPC request too and pass that along.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] core: add mor..." ]
>This comment: "That will probably need its own `dns.ServeMux` implement to have a Handler that can read and write gRPC."
>
>I am wondering if we can avoid that. The gRPC piece is built on protobuf, so that DNS format is bound to a specific proto file. It seems too application-y to embed those proto definitions in go-dns. So, as long as CoreDNS can implement its own ServeMux and hand that into go-dns I think this works. That's probably what you were intending anyway. I would like that handler to be able to decode the context out of the gRPC request too and pass that along.

Yes, I believe both are possible. Just working on incorperating the gRPC stuff
now.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] core: add mor..." ]
>Missing a ~

Done.

>> @@ -169,6 +173,14 @@ example.org {
>```
>tls://example.org dns://example.org {
>...

Yes (in the current PR it does)

>I guess I am trying to ask if we are differentiating between the network transport and the application protocol here, and how that is set up. This syntax makes me think this would be possible (but I don't think it can be):
>
>```
>tls:. dns://example.org grpc://example.com {
>}
>```

Nit `tls://.`. But what is wrong with the above?

>So I think a different syntax is better, one that separates layer 4 and layer 7 better. Maybe
>
>```
>tls:grpc://example.org {
>}
>
>tcp:dns://example.com {
>}
>
>tcp:grpc://example.net {
>}
>```

But didn't that ship sank when they defined the URL and URI? 

>But we also need to be able to specify the addresses and ports in the above syntax, and for TLS we need a directive to specify the certs for the server. I am thinking we may be trying to put too much into the initial args of the server block. Instead, there could be server-level directives that are not necessarily implemented as middleware.

Yes, in the PR the `tls` "middleware" does this. It similar to the "root"
middleware which is also not a read middleware.


>```
>example.org:53 {
>  protocol udp:dns tcp:dns
>}
>
>example.org:443 {
>  protocol tls:grpc [....tls params...]
>}
>
>example.com:443 {
>  protocol tls:dns [...tls params... - but they have to be the same for all servers on this port...]
>}
>```
>
>> @@ -0,0 +1,13 @@
>+# tls
>
>ok, i see the tls stuff is here so no need for it like I talked about above. I still think though if you leave the server definition as just domain:hostport we are better off. Otherwise you could do dns-tls and grpc for example and they would look like different server blocks but in fact are on the same port and so need to share the TLS info.

But how can they be in the same port? One of the listener would fail because the
port is already taken.

The point of not doing a `protocol` is that we up another server inside the
server, which feels a bit off. Esp. when transparant restarts and things are
considered.

/Miek

--
Miek Gieben

johnbelamaric:
> 
> >I guess I am trying to ask if we are differentiating between the network transport and the application protocol here, and how that is set up. This syntax makes me think this would be possible (but I don't think it can be):
> >
> >```
> >tls:. dns://example.org grpc://example.com {
> >}
> >```
> 
> Nit `tls://.`. But what is wrong with the above?
> 

I think it implies something broken. It’s trying to say make a TLS server port, then if the query coming in is for example.org, use plain DNS, if it’s for example.com, use GRPC. But we can’t know what zone it’s for until we know whether to use DNS or GRPC.

Take this Corefile today, for example:

```

foobar.com:5353 {
	log stdout
	proxy . /etc/resolv.conf
}

foo.bar:5353 {
	log stdout
	erratic
}

```

This works fine. CoreDNS will gather up the server blocks and mux the zones so that foo.bar goes to erratic and foobar.com goes to proxy. The syntax above makes met think it’s going to work the same way but it won’t.

> 
> 
> But didn't that ship sank when they defined the URL and URI? 
> 

Well, not sure that URL/URI is sufficient to describe everything we need. At least not in a natural way.

> 
> But how can they be in the same port? One of the listener would fail because the
> port is already taken.
> 

Yes, my bad. Same issue as what I talk about above. My point was muxing the port after the TLS connection but that doesn’t work across protocols.
> 
>

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] core: add mor..." ]
>> >tls:. dns://example.org grpc://example.com {
>> >}
>> >```
>>
>> Nit `tls://.`. But what is wrong with the above?
>>
>
>I think it implies something broken. It’s trying to say make a TLS server port, then if the query coming in is for example.org, use plain DNS, if it’s for example.com, use GRPC. But we can’t know what zone it’s for until we know whether to use DNS or GRPC.

But these are all running on diff. ports, wouldn't make that the choice for us?

>Take this Corefile today, for example:
>
>```
>
>foobar.com:5353 {
>	log stdout
>	proxy . /etc/resolv.conf
>}
>
>foo.bar:5353 {
>	log stdout
>	erratic
>}
>
>```
>
>This works fine. CoreDNS will gather up the server blocks and mux the zones so that foo.bar goes to erratic and foobar.com goes to proxy. The syntax above makes met think it’s going to work the same way but it won’t.

Hmmm, I see your point, but isn't this the same as with http and https? 

>Well, not sure that URL/URI is sufficient to describe everything we need. At least not in a natural way.

Ack. But I'm not 100% convinced such flexibility is needed. We don't have a now?
(discounting coredns/grpc atm :-) )

>> But how can they be in the same port? One of the listener would fail because the
>> port is already taken.
>>
>
>Yes, my bad. Same issue as what I talk about above. My point was muxing the port after the TLS connection but that doesn’t work across protocols.

To put it another way: In the current PR the different protocol don't share the
muxer.

/Miek

--
Miek Gieben

middleware/tls/README.md
miekg: Yes. I think the *root* middleware already has this problem. Last one counts there (I think).

miekg:
@johnbelamaric this folds gprc into coredns, please have look. CoreDNS starts with a `grpc://` stanza defined, no TLS there (yet). I also haven't got any simple testing tools to send some grpc queries.

Believe we should move the discussion about `tls://.` vs `protocol`-stanza into a bug. I'll try to do that.

core/dnsserver/server-grpc.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/coredns/coredns/middleware/proxy/pb"
+
+	"github.com/miekg/dns"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/peer"
+)
+
+// GRPCServer represents an instance of a DNS-over-gRPC.
+type GRPCServer struct {
```
johnbelamaric: Is there any real reason to export this?

core/dnsserver/server-grpc.go
```diff
@@ -0,0 +1,152 @@
...
+		return nil, errors.New("no peer in gRPC context")
+	}
+
+	a, ok := p.Addr.(*net.TCPAddr)
+	if !ok {
+		return nil, fmt.Errorf("no TCP peer in gRPC context: %v", p.Addr)
+	}
+
+	r := &net.IPAddr{IP: a.IP}
+	w := &gRPCresponse{localAddr: s.listenAddr, remoteAddr: r}
```
johnbelamaric: s.listenAddr is better than what is in coredns/grpc 👍

core/dnsserver/server-grpc.go
```diff
@@ -0,0 +1,152 @@
...
+	for zone, config := range s.zones {
+		fmt.Println(TransportGRPC + "://" + zone + ":" + config.Port)
+	}
+}
+
+func (s *GRPCServer) Stop() (err error) {
+	s.m.Lock()
+	defer s.m.Unlock()
+	if s.grpcServer != nil {
+		s.grpcServer.Stop()
```
johnbelamaric: What method is called for graceful restarts? We may have to use [GracefulStop](https://godoc.org/google.golang.org/grpc#Server.GracefulStop).

johnbelamaric (commented):
Yes!!

core/dnsserver/server-grpc.go
miekg: No. Same holds for TLS server, but tLS will make `go lint` complain...hence exported. I'll reverse the name serverTLS, servergRPC.

core/dnsserver/server-grpc.go
miekg: I didn't care about that (yet). Might be easy change in followup PRs

johnbelamaric: Yup. Existing out-of-tree doesn't worry about it yet either.

core/dnsserver/server-grpc.go
miekg: yeah :-) Saw those string conversions for every request... :-)

Can you easily check/add tests for this PR?

miekg:
I'll get the tls stuff into shape, fix the test and merge during the weekend

johnbelamaric:
Awesome. If I have time I will push some tests to your branch for this PR.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] core: add mor..." ]
>Awesome. If I have time I will push some tests to your branch for this PR.

That would be helpful, thanks

miekg:
@johnbelamaric Ok, this has most of the bits. PR is larger than I wanted (24 files touched...). And the test is throwing crap like:
~~~
2017/03/13 11:00:08 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: dial tcp [::]:38135: getsockopt: connection refused"; Reconnecting to {[::]:38135 <nil>}
2017/03/13 11:00:09 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: dial tcp [::]:38135: getsockopt: connection refused"; Reconnecting to {[::]:38135 <nil>}
2017/03/13 11:00:10 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: dial tcp [::]:38135: getsockopt: connection refused"; Reconnecting to {[::]:38135 <nil>}
2017/03/13 11:00:14 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: dial tcp [::]:38135: getsockopt: connection refused"; Reconnecting to {[::]:38135 <nil>}
2017/03/13 11:00:18 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: dial tcp [::]:38135: getsockopt: connection refused"; Reconnecting to {[::]:38135 <nil>}
PASS
~~~
Which I don't fully understand.

johnbelamaric:
Looks like the test is exiting, but we never call conn.Close() so the go routine is still running trying to connect after the server.Stop() is called in the defer.

core/dnsserver/server-grpc.go
```diff
@@ -0,0 +1,165 @@
...
+// NewGRPCServer returns a new CoreDNS GRPC server and compiles all middleware in to it.
+func NewServergRPC(addr string, group []*Config) (*servergRPC, error) {
+
+	s, err := NewServer(addr, group)
+	if err != nil {
+		return nil, err
+	}
+	gs := &servergRPC{Server: s}
+	gs.grpcServer = grpc.NewServer()
+	// trace foo... TODO(miek)
```
johnbelamaric, line 34: Still TODO here.

miekg: Yes, as-is the tls impl. Trace can now also be for *all* server, meaning it should be moved out of this PR anyway.

Want to do this in a follow up PR.

README.md
```diff
@@ -16,8 +17,8 @@ server that can talk to multiple backends (etcd, kubernetes, etc.).
 CoreDNS aims to be a fast and flexible DNS server. The keyword here is *flexible*: with CoreDNS you
 are able to do what you want with your DNS data. And if not: write some middleware!
 
-CoreDNS can listen for DNS request coming in over UDP/TCP (go'old DNS), TLS (RFC-XXXX) and gRPC (not
-a standard (yet?).
+CoreDNS can listen for DNS request coming in over UDP/TCP (go'old DNS), TLS (RFC 7858) and gRPC (not
```
johnbelamaric: A link to the RFC would be nice.

PR #576, opened 2017-03-08
core: remove dns.ServeMux

miekg:
We don't need to use the muxer in Go DNS s we only have one entry point
per server - and mux ourselves in ServeDNS.

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -26,7 +26,6 @@ import (
 // graceful termination (POSIX only).
 type Server struct {
 	Addr string // Address we listen on
-	mux  *dns.ServeMux
 
 	server [2]*dns.Server // 0 is a net.Listener, 1 is a net.PacketConn (a *UDPConn) in our case.
 	m      sync.Mutex     // protects the servers
@@ -44,9 +43,6 @@ func NewServer(addr string, group []*Config) (*Server, error) {
 		zones:       make(map[string]*Config),
 		connTimeout: 5 * time.Second, // TODO(miek): was configurable
 	}
-	mux := dns.NewServeMux()
-	mux.Handle(".", s) // wildcard handler, everything will go through here
-	s.mux = mux
 
 	// We have to bound our wg with one increment
 	// to prevent a "race condition" that is hard-coded
@@ -75,7 +71,10 @@ func NewServer(addr string, group []*Config) (*Server, error) {
 // This implements caddy.TCPServer interface.
 func (s *Server) Serve(l net.Listener) error {
 	s.m.Lock()
-	s.server[tcp] = &dns.Server{Listener: l, Net: "tcp", Handler: s.mux}
+	s.server[tcp] = &dns.Server{Listener: l, Net: "tcp", Handler: dns.HandlerFunc(func(w dns.ResponseWriter, r *dns.Msg) {
+		ctx := context.Background()
+		s.ServeDNS(ctx, w, r)
+	})}
 	s.m.Unlock()
 
 	return s.server[tcp].ActivateAndServe()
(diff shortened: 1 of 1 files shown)


core/dnsserver/server.go
```diff
@@ -155,10 +151,7 @@ func (s *Server) Address() string { return s.Addr }
 // defined in the request so that the correct zone
 // (configuration and middleware stack) will handle the request.
 func (s *Server) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
-	s.serveDNSWithContext(context.Background(), w, r)
-}
-
-func (s *Server) serveDNSWithContext(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) {
+	ctx := context.Background()
```
johnbelamaric: Can we put this back to ServeDNSWithContext? The last commit by Yong broke the gRPC middleware tracing since it can't call this method anymore. It looks like I never pushed that PR in coredns/grpc, but I'd like to.

johnbelamaric (commented):
Looks good except for the loss of the ServeDNSWithContext.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] core: remove ..." ]
>
>Looks good except for the loss of the ServeDNSWithContext.
>
>> @@ -155,10 +151,7 @@ func (s *Server) Address() string { return s.Addr }
> // defined in the request so that the correct zone
> // (configuration and middleware stack) will handle the request.
> func (s *Server) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
>-	s.serveDNSWithContext(context.Background(), w, r)
>-}
>-
>-func (s *Server) serveDNSWithContext(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) {
>+	ctx := context.Background()
>
>Can we put this back to ServeDNSWithContext? The last commit by Yong broke the gRPC middleware tracing since it can't call this method anymore. It looks like I never pushed that PR in coredns/grpc, but I'd like to.

Ah, I didn't see he downcased it.

yongtang:
We could upper case the
```
s/serveDNSWithContext/ServeDNSWithContext/
```
then adding a simple comment with `// ServeDNSWithContext ...` will address the export and lint issue at the same time.

johnbelamaric:
#577

johnbelamaric:
Here is why I need it: https://github.com/coredns/grpc/pull/7

core/dnsserver/server.go
miekg: I think we *can* do this simpler. The coredns.Server needs a dns.ReadDecorator to do some gRPC stuff while reading the packet. The writing can be done like it is done now in the coredns/grpc.
Keeping the Context can be done with using a function litteral when setting up the handler, where we pull out the context from gRPC and call ServeDNS.
We should also fix ServeDNS to take a context.Context. I'll make some changes that show what I mean here.

miekg:
Current PR neatly (IMHO) solves this. `ServeDNS` is now the canonical way to call into the middlewares.

johnbelamaric:
Yes, that LGTM. Assuming it passes CI 👍 

At some point we should switch to "context" from "golang.org/x/net/context" but that will require a lot of changes.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] core: remove ..." ]
>Yes, that LGTM. Assuming it passes CI 👍

:-)

>At some point we should switch to "context" from "golang.org/x/net/context" but that will require a lot of changes.

yes, thought about that too the last few days. Can the protobuf Go generator
already generate code with the new context?

Also bit worried about the vendored k8s stuff...


/Miek

--
Miek Gieben

PR #580, opened 2017-03-11
Some tests

johnbelamaric:
This adds a few basic tests for gRPC.

THIS DOESN'T WORK YET

But it's what I can do now so hopefully it gets things started.


johnbelamaric:
Ok, this test now creates a gRPC server and client and passes a query through it.

miekg:
pulled in to #574

PR #586, opened 2017-03-14
Add `force-tcp` to proxy over DNS

mwitkow:
This change adds a parameter to `proxy` middleware's `protocol dns` option of `force-tcp`.

This allows the proxying mode to be done over `tcp` instead of respecting the inbound request's protocol. This makes it significantly more reliable in case you're querying DNS over weird VPNs that have weird MTUs and blackhole packets.

### DIFF
--- a/middleware/proxy/README.md
+++ b/middleware/proxy/README.md
@@ -71,7 +71,8 @@ Currently `protocol` supports `dns` (i.e., standard DNS over UDP/TCP) and `https
 payload over HTTPS). Note that with `https_google` the entire transport is encrypted. Only *you* and
 *Google* can see your DNS activity.
 
-* `dns`: no options can be given at the moment.
+* `dns`: uses the standard DNS exchange. You can pass `force-tcp` to make sure that the proxied connection is performed
+  over TCP, regardless of the inbound request's protocol.
 * `https_google`: bootstrap **ADDRESS...** is used to (re-)resolve `dns.google.com` to an address to
   connect to. This happens every 300s. If not specified the default is used: 8.8.8.8:53/8.8.4.4:53.
   Note that **TO** is *ignored* when `https_google` is used, as its upstream is defined as
--- a/middleware/proxy/dns.go
+++ b/middleware/proxy/dns.go
@@ -52,7 +52,10 @@ func (d *dnsEx) Exchange(ctx context.Context, addr string, state request.Request
 	if err != nil {
 		return nil, err
 	}
-
+	if reply.Len() > allowedReplySize(state) {
+		// Maybe clear the RR fields so the truncated definitely gets there?
+		reply.Truncated = true
+	}
 	reply.Compress = true
 	reply.Id = state.Req.Id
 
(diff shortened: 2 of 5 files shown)


mwitkow:
@miekg and @kdima for context

miekg:
How do you make sure it fits in the UDP buffer of the original client? It also redials for every conn which is quite expensive.

mwitkow:
@miekg the redial, it already redials if a query needs to fall back from UDP to TCP. 

As for size of the response. Would checking `reply.Len() >= max(512, opt.UDPSize)` and setting `reply.Truncated` work? If not, any tips? :)

miekg:
> As for size of the response. Would checking reply.Len() >= max(512, opt.UDPSize) and setting reply.Truncated work? If not, any tips? :)

That might not be totally crazy.

As an alternative: woudn't doing UDP retries be simpler? (And more in the spirit of go'old DNS?)

mwitkow:
The problem is UDP and MTU issues. The downstream requests get lost and the whole lookup times out.

I'll add the reply.Len check

mwitkow:
@miekg the only thing I don't know is how to clear the RR records of the reply to make sure that the Golang resolver doesn't reject the response becauuse its bigger than 512B

miekg:
~~~
state.SizeAndDo(resp)
resp, _ = state.Scrub(resp)
w.WriteMsg(resp)
~~~
Should do trick. But double check, I'm only drinking my third coffee.

mwitkow:
Right, done it with `Scrub`. PTAL.

Tested on an SRV response of ~700B:
 * dig with bufsize=512, bufsize=2014 both work (512 retries)
 * using a Go resolver, retries over TCP

johnbelamaric (commented):
Please add unit tests and also an integration test (in top level test dir) trying the different packet sizes.

mwitkow:
@johnbelamaric I'd be happy to implement some, but there really isn't anything to piggy back off, either for unit tests or integraiton test:
 * `test/proxy_test.go` has no integration tests except for reading the config
 * `middleware/proxy` has no unit tests tests for DNS proxying 

And unfortunately I don't think I am confident enough to write an unit test/integration test suite for proxying myself.

johnbelamaric:
Ok on the unit tests. Integration tests go here: https://github.com/coredns/coredns/blob/master/test/proxy_test.go

johnbelamaric:
Oh, sorry, see you mentioned that. https://github.com/coredns/coredns/blob/master/test/proxy_test.go#L15 actually creates a server and passes a request through it.

The upstream_test.go has tests for parsing - you could add one unit test there.

miekg:
Yes, should be relatively easy to plumb this in test/proxy_test.go

other than that; lgtm

mwitkow:
@miekg added a test that makes the request to the proxied server over TCP. However the zone served in hte example doesn't really have a large enough responsen size.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add `force-tc..." ]
>@miekg added a test that makes the request to the proxied server over TCP. However the zone served in hte example doesn't really have a large enough responsen size.

Ack. I'll file an issue that we should still test this.

johnbelamaric:
Feature for erratic - add a way to bulk up the return size and/or set truncated flag.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add `force-tc..." ]
>Feature for erratic - add a way to bulk up the return size and/or set truncated flag.

Please file as an issue otherwise we will loose track

miekg:
superseded with #594

PR #603, opened 2017-03-16
middleware/proxy: fix race; add Go 1.7 backward compatibility

mfischer-zd:
* Fix race on backend health status update
* Ensure test case is compatible on Go 1.7


mfischer-zd:
I think the Codecov error was possibly caused by my doing a force-push on the branch and yanking the first ref out from under it.  Sorry about that.

PR #606, opened 2017-03-18
bump travis to 1.8

miekg:


.travis.yml
```diff
@@ -7,7 +7,7 @@ services:
 
 language: go
 go:
-  - 1.7
+  - 1.8
```
mfischer-zd, line 10: Do we want to replace, or add?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] bump travis t..." ]
>mfischer-zd commented on this pull request.
>
>
>
>> @@ -7,7 +7,7 @@ services:
>
> language: go
> go:
>-  - 1.7
>+  - 1.8
>
>Do we want to replace, or add?

Replace, mostly because travis already takes a long time.

Got me thinking about 1.7/1.8. Maybe we should just standardize on the
latest Go version (personally I don't have a problem with).

/Miek

--
Miek Gieben

mfischer-zd:
I think Travis runs the jobs in parallel - but I'd have to test to be sure.

At any rate, testing only against the latest major release is fine with me.  Most users will be using binary builds anyway, so they don't really care which version of Go was used to produce it.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] bump travis t..." ]
>I think Travis runs the jobs in parallel - but I'd have to test to be sure.
>
>At any rate, testing only against the latest major release is fine with me.  Most users will be using binary builds anyway, so they don't really care which version of Go was used to produce it.

I agree

mfischer-zd (approved):

PR #607, opened 2017-03-27
readme: fixed typos

michael-k:


miekg:
Thanks. It seems the last lines got some extra spaces on the end though.

michael-k:
Yeah, exactly 2 per line.  That's a feature of markdown to enforce line breaks.  Compare https://github.com/coredns/coredns#blog-and-contact with https://github.com/michael-k/coredns/blob/patch-1/README.md#blog-and-contact

I can simply remove them or convert the section to a list if you want.

miekg:
Hmmm, using that technique is a bit of an anti-pattern and I probably won't survive the next round of "trim-all-eol-whitespace". Is there nothing else you can use? Or maybe make it a bullet list?

michael-k:
converted to a list

https://github.com/michael-k/coredns/blob/patch-1/README.md#blog-and-contact

PR #611, opened 2017-04-06
this paragraph was redundant

goern:
The paragraph felt redundant while reading thru the README


miekg:
Thanks for doing cleanups like these!

PR #613, opened 2017-04-12
middleware/metrics: allow multiple listeners

miekg:
There was no inherent reason *not* to allow multiple listeners for the
monitoring data. Actually enforcing only one listener lead to more code
then just allowing multiple. It's probably not what you want; but
CoreDNS is happy to oblige.


miekg:
@johnbelamaric (diff from our email exchange)

PR #619, opened 2017-04-17
Tracing for gRPC Server

johnbelamaric:
This implements the tracing for the gRPC server. It does not implement it for other servers, which will require more extensive changes to pass context through the proxy.

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -13,9 +13,11 @@ import (
 	"github.com/coredns/coredns/middleware/metrics/vars"
 	"github.com/coredns/coredns/middleware/pkg/edns"
 	"github.com/coredns/coredns/middleware/pkg/rcode"
+	"github.com/coredns/coredns/middleware/pkg/trace"
 	"github.com/coredns/coredns/request"
 
 	"github.com/miekg/dns"
+	ot "github.com/opentracing/opentracing-go"
 	"golang.org/x/net/context"
 )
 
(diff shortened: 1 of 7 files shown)


core/dnsserver/server.go
```diff
@@ -59,6 +62,15 @@ func NewServer(addr string, group []*Config) (*Server, error) {
 		var stack middleware.Handler
 		for i := len(site.Middleware) - 1; i >= 0; i-- {
 			stack = site.Middleware[i](stack)
+			if s.trace == nil && stack.Name() == "trace" {
+				// we have to stash away the middleware, not the
+				// Tracer object, because the Tracer won't be initialized yet
+				if t, ok := stack.(trace.Trace); ok {
+					s.trace = t
+				} else {
+					log.Println("[WARNING] Invalid type found for trace middleware. This is a bug.")
```
miekg: This happens when we add a middleware with `Name()` is trace, but isn't a `trace.Trace`? Think it is save to drop the else here. The users of CoreDNS won't know what to do anyway.

(We could add a tests for middleware having unique names)

core/dnsserver/server.go
```diff
@@ -242,6 +254,14 @@ func (s *Server) OnStartupComplete() {
 	}
 }
 
+func (s *Server) Tracer() ot.Tracer {
```
miekg, line 255: Just uppercase Trace in server? Wont need this helper method then.

middleware/pkg/trace/trace.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"github.com/coredns/coredns/middleware"
+	ot "github.com/opentracing/opentracing-go"
+)
+
+// Trace holds the tracer and endpoint info
+type Trace interface {
+	middleware.Handler
+	Tracer() ot.Tracer
+}
```
miekg, line 12: ah, disregard prev. comment. Didn't see this interface yet.

miekg (approved):

PR #626, opened 2017-04-19
Add fallthrough support for Kubernetes

johnbelamaric:
This enables registering other services in the same zone as
Kubernetes services. This also re-orders the middleware chain
so that Kubernetes comes before other types, in order to make
this work out-of-the-box.


miekg:
lgtm. Might want to update the README as well (can be done in separate PR).

PR #627, opened 2017-04-19
msg.Service: add HostType() method

miekg:
This method parse the Host field in the service. It returns 1 or 3
things 1) it is a host 2) an IPv4 address or an 3) IPv6 address.
This simplifies some code quite a bit.

### DIFF
--- a/middleware/etcd/msg/type_test.go
+++ b/middleware/etcd/msg/type_test.go
@@ -1,0 +1,31 @@
+package msg
+
+import (
+	"testing"
+
+	"github.com/miekg/dns"
+)
+
+func TestType(t *testing.T) {
+	tests := []struct {
+		serv         Service
+		expectedType uint16
+	}{
+		{Service{Host: "example.org"}, dns.TypeANY},
+		{Service{Host: "127.0.0.1"}, dns.TypeA},
+		{Service{Host: "2000::3"}, dns.TypeAAAA},
+		{Service{Host: "2000..3"}, dns.TypeANY},
+		{Service{Host: "127.0.0.257"}, dns.TypeANY},
+		{Service{Host: "127.0.0.252", Mail: true}, dns.TypeA},
+		{Service{Host: "127.0.0.252", Mail: true, Text: "a"}, dns.TypeA},
+		{Service{Host: "127.0.0.254", Mail: false, Text: "a"}, dns.TypeA},
+	}
+
+	for i, tc := range tests {
+		what, _ := tc.serv.HostType()
+		if what != tc.expectedType {
+			t.Errorf("Test %d: Expected what %v, but got %v", i, tc.expectedType, what)
+		}
+	}
+
+}
--- a/middleware/backend_lookup.go
+++ b/middleware/backend_lookup.go
@@ -21,9 +21,11 @@ func A(b ServiceBackend, zone string, state request.Request, previousRecords []d
 	}
 
 	for _, serv := range services {
-		ip := net.ParseIP(serv.Host)
-		switch {
-		case ip == nil:
+
+		what, ip := serv.HostType()
+
+		switch what {
+		case dns.TypeANY:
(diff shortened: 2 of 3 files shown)


miekg:
@yongtang could you give this a glance over? Should be mostly OK (now waiting for tests to pass), but code review never hurts.

shashidharatd:
@miekg, Thanks for this.
It would have been great, if we can get the type of DNS record by looking at msg.Service.
Since CoreDNS also supports SRV, MX, TXT, PTR, NS records, we should be able to tell if a msg.Service is one of those.

miekg:
Good point. Maybe this should be more generic and just return the
dns.TypeXXX

On 20 Apr 2017 11:08 am, "Shashidhara T D" <notifications@github.com> wrote:

@miekg <https://github.com/miekg>, Thanks for this.
It would have been great, if we can get the type of DNS record by looking
at msg.Service.
Since CoreDNS also supports SRV, MX, TXT, PTR, NS records, we should be
able to tell if a msg.Service is one of those.

—
You are receiving this because you were mentioned.

<https://github.com/coredns/coredns/pull/627#issuecomment-295664688>, or mute
the thread

miekg:
@shashidharatd a single service can represent multiple RR, i.e. it depends what you want to do with it. I've narrowed it down to `HostType` which can be `dns.TypeA`, `dns.TypeAAAA` or `dns.TypeANY` (there is no dns equivalent to signal, it is just a name).

PTAL

middleware/etcd/msg/type_test.go
```diff
@@ -0,0 +1,31 @@
...
+		serv         Service
+		expectedType uint16
+	}{
+		{Service{Host: "example.org"}, dns.TypeANY},
+		{Service{Host: "127.0.0.1"}, dns.TypeA},
+		{Service{Host: "2000::3"}, dns.TypeAAAA},
+		{Service{Host: "2000..3"}, dns.TypeANY},
+		{Service{Host: "127.0.0.257"}, dns.TypeANY},
+		{Service{Host: "127.0.0.252", Mail: true}, dns.TypeA},
+		{Service{Host: "127.0.0.252", Mail: true, Text: "a"}, dns.TypeA},
```
shashidharatd, line 20: @miekg, does this mean A record could also be a MX record and a TXT record?
Slightly difficult to comprehend. I seriously hoped Service contained a field specifically to signify the record type.

shashidharatd: Since Service can represent multiple RR, isn't it better if we return multiple record types for Service Type function?
What is your opinion @miekg?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] msg.Service: ..." ]
>shashidharatd commented on this pull request.
>
>
>
>> +
>+	"github.com/miekg/dns"
>+)
>+
>+func TestType(t *testing.T) {
>+	tests := []struct {
>+		serv         Service
>+		expectedType uint16
>+	}{
>+		{Service{Host: "example.org"}, dns.TypeANY},
>+		{Service{Host: "127.0.0.1"}, dns.TypeA},
>+		{Service{Host: "2000::3"}, dns.TypeAAAA},
>+		{Service{Host: "2000..3"}, dns.TypeANY},
>+		{Service{Host: "127.0.0.257"}, dns.TypeANY},
>+		{Service{Host: "127.0.0.252", Mail: true}, dns.TypeA},
>+		{Service{Host: "127.0.0.252", Mail: true, Text: "a"}, dns.TypeA},
>
>@miekg, does this mean A record could also be a MX record and a TXT record?

Technically Yes. All you have in a msg.Service is something that is reacheable
on Host (and Port), it doesn't tell you what protocol it talks. (In hindsight
adding the Mail boolean might be considered a mistake)

>Slightly difficult to comprehend. I seriously hoped Service contained a field specifically to signify the record type.

There is no 1:1 mapping

/Miek

--
Miek Gieben

PR #639, opened 2017-04-21
Pprof listen

chrisohaver:
Add option to pprof middleware to specify address to listen on.

### DIFF
--- a/middleware/pprof/setup.go
+++ b/middleware/pprof/setup.go
@@ -9,9 +9,7 @@ import (
 	"github.com/mholt/caddy"
 )
 
-const (
-	default_addr = "localhost:6053"
-)
+const defaultAddr = "localhost:6053"
 
 func init() {
 	caddy.RegisterPlugin("pprof", caddy.Plugin{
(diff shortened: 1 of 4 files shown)


middleware/pprof/setup.go
```diff
@@ -1,13 +1,18 @@
...
 
 	"github.com/coredns/coredns/middleware"
 
 	"github.com/mholt/caddy"
 )
 
+const (
+	default_addr = "localhost:6053"
+)
+
```
miekg: you can drop the `(` and `)` as there is only 1 const defined. Also `default_addr` -> `defaultAddr`

middleware/pprof/setup.go
```diff
@@ -17,11 +22,21 @@ func init() {
 
 func setup(c *caddy.Controller) error {
 	found := false
+	h := &handler{}
+	h.addr = default_addr
```
miekg: you can do this in 1 go: `h := &handler{addr: defaultAddr}`

miekg:
some nits. LGTM

Issue #635, opened 2017-04-20
RRSIGs are served with expiration dates in the past


liamr4c:
The errors given by DNSViz can be seen [here](http://dnsviz.net/d/liams.io/WPUjgA/dnssec/). Invalid RRSIGs obviously makes resolvers like google not want to resolve your domain.

I use the `file` middleware to load the zones and the `dnssec` middleware to sign DNSSEC on the fly, which normally works. My Corefile seems to not have anything obviously wrong:
```
liams.io:53 {
    file liams.io.zone
    dnssec {
        key file Kliams.io.+013+55741
    }
    errors stdout
    rewrite type ANY HINFO
}
```

Restarting coredns usually fixes this, so maybe it is the `cache` not taking into account the fact that RRSIGs expire?

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] RRSIGs are served..." ]
>The errors given by DNSViz can be seen [here](http://dnsviz.net/d/liams.io/WPUjgA/dnssec/). Invalid RRSIGs obviously makes resolvers like google not want to resolve your domain.
>
>I use the `file` middleware to load the zones and the `dnssec` middleware to sign DNSSEC on the fly, which normally works. My Corefile seems to not have anything obviously wrong:
>```
>liams.io:53 {
>    file liams.io.zone
>    dnssec {
>        key file Kliams.io.+013+55741
>    }
>    errors stdout
>    rewrite type ANY HINFO
>}
>```
>
>Restarting coredns usually fixes this, so maybe it is the `cache` not taking into account the fact that RRSIGs expire?

you Corefile does not contain `cache` (in the above example). Regarding your
statement: #367

(so you're right; should be a simple fix though)

/Miek

--
Miek Gieben

PR #641, opened 2017-04-22
middleware/cache: don't cache expired RRSIGs

### LINKED ISSUE #367: middleware/cache signature lifetime is not considered
Signature lifetime does not get accounted for in any way. The assumption currently is, that the TTL should be lower than the signature lifetime.

The next step to tackle this might be to remove a cached item, when the signature lifetime has run out. (This should happen even when the cache TTL or the record TTL is no reached yet as serving invalid signatures should be prevented)

miekg:
Check message for expired sig and don't cache those.

Aside: This hack of caching entire messages is probably something we
should stop doing at some point in the future and do this on a per RRset
basis.

Fixes #367 #635


yongtang:
lgtm.

Issue #610, opened 2017-03-30
Add support for go vendor


andyxning:
Add support for go vendor to make it build with a reproducible binary. :) 

How about we use godep to maintain coredns dependencies?

miekg:
I don't like godep and I think we should wait until `dep` becomes the official way of handling dependencies in Go code.

andyxning:
@miekg I have no special tastes on go dependency tools. :) I just propose that we need to make the dependencies recored ASAP. 

I have read the [roadmap](https://github.com/golang/dep/wiki/Roadmap) and [readme](https://github.com/golang/dep). It seems that there will be at least one year for it to be production ready. I am afraid this takes too long. And the project also said that `dep` is not an golang official project and there is no guarantee it will be merged into go tool chain.

As for `godep`, Kubernetes project is using it. Maybe it is not perfect. But it can work now. :)

johnbelamaric:
We use https://glide.sh in our internal stuff. It's pretty straightforward. You create a YAML with the dependencies and it pulls in all the right vendor code. I've asked @mstaneuka to see what that would look like for CoreDNS.

miekg:
Major reason why i don't like vendoring is that we then have to deal with a
upgrade-all-vendored code every once ina while. I rather deal with smaller
breakages along the way...

johnbelamaric:
Yeah, I get that. We already have that issue with the one thing we vendor (k8s client), which we'll probably need to bump up sooner or later.

Without it though it makes a repeatable build harder. And if someone deletes an upstream repository we'll be hurting.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add support f..." ]
>Yeah, I get that. We already have that issue with the one thing we vendor (k8s client), which we'll probably need to bump up sooner or later.
>
>Without it though it makes a repeatable build harder. And if someone deletes an upstream repository we'll be hurting.

that's true...

andyxning:
> Major reason why i don't like vendoring is that we then have to deal with a
upgrade-all-vendored code every once ina while. I rather deal with smaller
breakages along the way...

@miekg I am afraid that it is a common and main job for vendor dependency management to upgrade vendored dependencies. It is unavoidable. :)

PR #653, opened 2017-04-27
Introduce go vendor management using 'glide' tool (https://github.com/coredns/coredns/issues/610)

mstaneuka:


johnbelamaric:
So, what would the process be for updating all the vendor code? Maybe we should have a make target that updates all vendor code.

What do we have to do if we wanted to add another package?

yongtang:
I could see there are so many subpackges for `k8s.io/client-go`. That means updating it will have to go through so many lines in config file, not even to mention the vendoring code?

Do we need to include sub packages level? Could we just list `client-go`?

mstaneuka:
> @johnbelamaric  So, what would the process be for updating all the vendor code? Maybe we should have a make target that updates all vendor code.

Generally it seems very hard to update glide.yaml automatically. What I see in other projects - vendor usually updated on demand either as a part of feature's PR or to fix some version specific issues. However some core dependencies could be updated from time to time.

Perhaps having reproducible environment is more important than keep in sync with latest version of every package we are using...

> @johnbelamaric  What do we have to do if we wanted to add another package?

We could either add some scripts similar to https://github.com/coreos/etcd/blob/master/scripts/updatedep.sh or add instructions e.g. to the README. I prefer not to use any scripts and use glide directly.

mstaneuka:
> @yongtang I could see there are so many subpackges for k8s.io/client-go. That means updating it will have to go through so many lines in config file, not even to mention the vendoring code?
>
>Do we need to include sub packages level? Could we just list client-go?

If I'm correct originally ```subpackages``` has been added to get rid of transitive dependencies fetch for a subpackages we don't really use. E.g.
```
- package: github.com/foo
  subpackages:
  - bar
``` 
may lead to skipping deps fetch for  github.com/foo/spam (if github.com/foo/bar doesn't use github.com/foo/spam).

What I understand currently glide uses some code scanning to resolve such cases. So yes, we can remove  ```subpackages```. At lest it doesn't make any changes in the vendor

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Introduce go ..." ]
>> @johnbelamaric  So, what would the process be for updating all the vendor code? Maybe we should have a make target that updates all vendor code.
>
>Generally it seems very hard to update glide.yaml automatically. What I see in other projects - vendor usually updated on demand either as a part of feature's PR or to fix some version specific issues. However some core dependencies could be updated from time to time.
>
>Perhaps having reproducible environment is more important than keep in sync with latest version of every package we are using...
>
>> @johnbelamaric  What do we have to do if we wanted to add another package?
>
>We could either add some scripts similar to https://github.com/coreos/etcd/blob/master/scripts/updatedep.sh or add instructions e.g. to the README. I prefer not to use any scripts and use glide directly.

I'm in favor of vendoring (reluctantly :) ). How anyone any ideas how `go dep`
will do this? Get we play with that, when is it ready?

mstaneuka:
Closing PR as vendoring was covered by https://github.com/coredns/coredns/pull/654

PR #654, opened 2017-04-28
Add vendor with `go dep`

### LINKED ISSUE #610: Add support for go vendor

yongtang:
This fix adds vendor with `go dep`

(`go dep` has not been officially released yet but if every tests pass we could think about using it)

This fix fixes #610

/cc @miekg @johnbelamaric @mstaneuka


miekg:
Ack. "lgtm" as I didn't review anything on the 8000+ files

yongtang:
Thanks. Merge ...

PR #658, opened 2017-05-03
middleware/file: correctly parse the stanza

### LINKED ISSUE #657: middleware/file: external names not resolved
External names in zone file are not resolved even with a upstream resolver configured:
```
example.org:1053 {
  file example.org example.org {
    upstream 8.8.8.8
  }

  log stdout
}
```

The `example.org` file has a CNAME pointing to an [xip.io](http://xip.io/) subdomain, which resolves to `10.0.0.1`:
```
@       IN      SOA     sns.dns.icann.org. noc.dns.icann.org. 2017042703 7200 3600 1209600 3600
        IN      NS      a.iana-servers.net.
        IN      NS      b.iana-servers.net.
www     IN      CNAME   foobar.10.0.0.1.xip.io.
```

The server however returns the CNAME record only:
```
$ dig www.example.org @127.0.0.1:1053
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 9812
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 1; AUTHORITY: 2; ADDITIONAL: 0

;; QUESTION SECTION:
;; www.example.org.    		IN	A

;; ANSWER SECTION:
www.example.org.    	3600	IN	CNAME	foobar.10.0.0.1.xip.io.

;; AUTHORITY SECTION:
example.org.        	3600	IN	NS	a.iana-servers.net.
example.org.        	3600	IN	NS	b.iana-servers.net.
```

The A record for `foobar.10.0.0.1.xip.io.` should be included in the answer section as well:
```
foobar.10.0.0.1.xip.io.	300	IN	A	10.0.0.1
```

Reproduced with latest build (4fc1318e).

miekg:
Parsing the file stanza would give precedence to 'transfer' and ignore
other bits if it wasn't specified.

This change fixes the parsing. The actually external CNAME retrieval is
working fine (once the upstream is correctly parsed).

This wasn't caught in tests, because we lack a parsing test for this.

Fixes #657

### DIFF
--- a/middleware/file/setup.go
+++ b/middleware/file/setup.go
@@ -92,12 +92,17 @@ func fileParse(c *caddy.Controller) (Zones, error) {
 
 			noReload := false
 			prxy := proxy.Proxy{}
+			t := []string{}
+			var e error
+
 			for c.NextBlock() {
-				t, _, e := TransferParse(c, false)
-				if e != nil {
-					return Zones{}, e
-				}
 				switch c.Val() {
+				case "transfer":
+					t, _, e = TransferParse(c, false)
+					if e != nil {
+						return Zones{}, e
+					}
+
 				case "no_reload":
 					noReload = true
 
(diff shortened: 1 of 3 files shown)


miekg:
Needs a tests which i will add.

johnbelamaric:
lgtm

PR #662, opened 2017-05-05
Handle K8s middleware NS record

chrisohaver:
Adds NS record for the dns service name and ip.
For an NS query: The service domain name (per k8s dns spec), and the service's Cluster IP are returned.
Also, an A record request for "ns.dns.zone" will return the service's Cluster IP. (to make the SOA response valid)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ all: coredns
 # Phony this to ensure we always build the binary.
 # TODO: Add .go file dependencies.
 .PHONY: coredns
-coredns: check
+coredns: fmt check
 	go build $(BUILD_VERBOSE) -ldflags="-s -w"
 
 .PHONY: deps
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ all: coredns
 # Phony this to ensure we always build the binary.
 # TODO: Add .go file dependencies.
 .PHONY: coredns
-coredns: fmt check
+coredns: check
 	go build $(BUILD_VERBOSE) -ldflags="-s -w"
 
 .PHONY: deps
@@ -15,7 +15,7 @@ deps: core/zmiddleware.go core/dnsserver/zdirectives.go
 	go get -u github.com/golang/lint/golint
 
 .PHONY: check
-check: deps
+check: fmt deps
 
 .PHONY: test
 test: check
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ deps: core/zmiddleware.go core/dnsserver/zdirectives.go
 	go get -u github.com/golang/lint/golint
 
 .PHONY: check
-check: fmt deps
+check: deps
 
 .PHONY: test
 test: check
--- a/Makefile.release
+++ b/Makefile.release
@@ -88,7 +88,7 @@ release:
 
 .PHONY: docker-build
 docker:
-	CGO_ENABLED=0 GOOS=linux /usr/local/go/bin/go build -ldflags="-s -w"
+	CGO_ENABLED=0 GOOS=linux go build -ldflags="-s -w"
 	docker build -t $(DOCKER_IMAGE_NAME) .
 	docker tag $(DOCKER_IMAGE_NAME):latest $(DOCKER_IMAGE_NAME):$(VERSION)
 
--- a/Makefile.release
+++ b/Makefile.release
@@ -38,7 +38,7 @@ NAME:=coredns
 VERSION:=$(shell grep 'coreVersion' coremain/version.go | awk '{ print $$3 }' | tr -d '"')
 ARCH:=$(shell uname -m)
 GITHUB:=coredns
-DOCKER:=coredns
+DOCKER:=chrisohaver
 DOCKER_IMAGE_NAME:=$(DOCKER)/$(NAME)
 
 all:	commit push build tar release
@@ -88,7 +88,7 @@ release:
 
 .PHONY: docker-build
 docker:
-	CGO_ENABLED=0 GOOS=linux go build -ldflags="-s -w"
+	CGO_ENABLED=0 GOOS=linux /usr/local/go/bin/go build -ldflags="-s -w"
 	docker build -t $(DOCKER_IMAGE_NAME) .
 	docker tag $(DOCKER_IMAGE_NAME):latest $(DOCKER_IMAGE_NAME):$(VERSION)
 
--- a/Makefile.release
+++ b/Makefile.release
@@ -38,7 +38,7 @@ NAME:=coredns
 VERSION:=$(shell grep 'coreVersion' coremain/version.go | awk '{ print $$3 }' | tr -d '"')
 ARCH:=$(shell uname -m)
 GITHUB:=coredns
-DOCKER:=chrisohaver
+DOCKER:=coredns
 DOCKER_IMAGE_NAME:=$(DOCKER)/$(NAME)
 
 all:	commit push build tar release
--- a/middleware/kubernetes/controller.go
+++ b/middleware/kubernetes/controller.go
@@ -35,7 +35,7 @@ func (s *storeToNamespaceLister) List() (ns api.NamespaceList, err error) {
 	return ns, nil
 }
 
-type dnsControllerItf interface {
+type dnsController interface {
 	ServiceList() []*api.Service
 	PodIndex(string) []interface{}
 	EndpointsList() api.EndpointsList
@@ -43,7 +43,7 @@ type dnsControllerItf interface {
 	Stop() error
 }
 
-type dnsController struct {
+type dnsControl struct {
 	client *kubernetes.Clientset
 
 	selector *labels.Selector
@@ -67,8 +67,8 @@ type dnsController struct {
 }
 
 // newDNSController creates a controller for CoreDNS.
-func newdnsController(kubeClient *kubernetes.Clientset, resyncPeriod time.Duration, lselector *labels.Selector, initPodCache bool) *dnsController {
-	dns := dnsController{
+func newdnsController(kubeClient *kubernetes.Clientset, resyncPeriod time.Duration, lselector *labels.Selector, initPodCache bool) *dnsControl {
+	dns := dnsControl{
 		client:   kubeClient,
 		selector: lselector,
 		stopCh:   make(chan struct{}),
@@ -294,12 +294,12 @@ func endpointsWatchFunc(c *kubernetes.Clientset, ns string, s *labels.Selector) 
 	}
 }
 
-func (dns *dnsController) controllersInSync() bool {
+func (dns *dnsControl) controllersInSync() bool {
 	return dns.svcController.HasSynced()
 }
 
 // Stop stops the  controller.
-func (dns *dnsController) Stop() error {
+func (dns *dnsControl) Stop() error {
 	dns.stopLock.Lock()
 	defer dns.stopLock.Unlock()
 
@@ -315,7 +315,7 @@ func (dns *dnsController) Stop() error {
 }
 
 // Run starts the controller.
-func (dns *dnsController) Run() {
+func (dns *dnsControl) Run() {
 	go dns.svcController.Run(dns.stopCh)
 	go dns.nsController.Run(dns.stopCh)
 	go dns.epController.Run(dns.stopCh)
@@ -325,7 +325,7 @@ func (dns *dnsController) Run() {
 	<-dns.stopCh
 }
 
-func (dns *dnsController) NamespaceList() *api.NamespaceList {
+func (dns *dnsControl) NamespaceList() *api.NamespaceList {
 	nsList, err := dns.nsLister.List()
 	if err != nil {
 		return &api.NamespaceList{}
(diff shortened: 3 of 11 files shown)


middleware/kubernetes/kubernetes.go
```diff
@@ -92,13 +101,27 @@ func (k *Kubernetes) Services(state request.Request, exact bool, opt middleware.
 		return nil, nil, e
 	}
 
+	if state.Name() == DefaultNSName+"."+r.zone && state.Type() == "A" {
```
johnbelamaric: Do we really need to do the additions with "." and r.zone or can we just do strings.Index(state.Name(), DefaultNSName) == 0 ?

johnbelamaric (changes requested):
Also you need to go fmt.

middleware/kubernetes/kubernetes.go
miekg: `DefaultNSName` could also include a `.`

Also: shall we move some of this logic on another file? This one getting big, `ns.go` or something?

middleware/kubernetes/kubernetes.go
```diff
@@ -80,10 +82,17 @@ type recordRequest struct {
 	port, protocol, endpoint, service, namespace, typeName, zone string
 }
 
+type nsRecord struct {
```
miekg: Can't we just use `dns.A`?

middleware/kubernetes/kubernetes.go
```diff
@@ -115,6 +138,73 @@ func (k *Kubernetes) recordsForTXT(r recordRequest) ([]msg.Service, error) {
 	return nil, nil
 }
 
+func (k *Kubernetes) CoreDNSRecord() nsRecord {
```
miekg: Can there only be 1?

miekg (changes requested):

middleware/kubernetes/kubernetes.go
chrisohaver: Realistically/Practically, yes.  Which is why I treat it this way.

In theory, there could be multiple services pointing to the coredns pod, and those services may or may not have the DNS port open.  So a complete answer would be to look at each service that points to the pod, and examine the port forwarding in the service and pod configs to see which ones forward dns to coredns.

But in this PR, I am assuming there is one service for the DNS pod, and that service forwards DNS port to the pod.

middleware/kubernetes/kubernetes.go
chrisohaver: @johnbelamaric : `strings.Index(state.Name(), DefaultNSName) == 0` would falsely catch an A record request for `ns.dns.svc.cluster.local`... which could be a valid domain name for service "ns" in  namespace "dns".

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Handle K8s mi..." ]
>
>
>
>> @@ -92,13 +101,27 @@ func (k *Kubernetes) Services(state request.Request, exact bool, opt middleware.
> 		return nil, nil, e
> 	}
>
>+	if state.Name() == DefaultNSName+"."+r.zone && state.Type() == "A" {
>
>@johnbelamaric : `strings.Index(state.Name(), DefaultNSName) == 0` would falsely catch an A record request for `ns.dns.svc.cluster.local`... which could be a valid domain name for service "ns" in  namespace "dns".

ah, we also claiming a namespace here. Should we rename 'dns' to
'kube-system'?

Is `r.zone` static, 'cause the above causes garbage for every query hitting this
middleware.

chrisohaver:
> ah, we also claiming a namespace here. Should we rename 'dns' to 'kube-system'?

Ideally, SOA would return the actual domain name of the dns service.  But currently, backend_lookup SOA returns a hard-coded record to `ns.dns.[zone]`, which means that all middleware needs to add handle an A record request for `ns.dns.[zone]`.

> Is `r.zone` static, 'cause the above causes garbage for every query hitting this middleware.

r.zone can be different for each request. A middleware can handle more then one zone, each request is only for one zone.

johnbelamaric:
Still, we can do the Index check and then double-check inside the if with the complete string. That way we avoid the garbage *unless* we enter the if - it's somewhat unlikely that we will see a service named `ns` in a namespace `dns`.

Or maybe we should change the `dns.dns.` hardcoding.

chrisohaver:
> Or maybe we should change the dns.dns. hardcoding.

Yes, IMO we should change SOA to ask the middleware what the record should be... ideally in a way that doest require us to update all other middleware.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Handle K8s mi..." ]
>> ah, we also claiming a namespace here. Should we rename 'dns' to 'kube-system'?
>
>Ideally, SOA would return the actual domain name of the dns service.  But currently, backend_lookup SOA returns a hard-coded record to `ns.dns.[zone]`, which means that all middleware needs to add handle an A record request for `ns.dns.[zone]`.
>
>> Is `r.zone` static, 'cause the above causes garbage for every query hitting this middleware.
>
>r.zone can be different for each request. A middleware can handle more then one zone, each request is only for one zone.

Ack. but we should try to do this with allocating a new string

chrisohaver:
Latest change ...
* Splits NS record functions into a separate file: ns.go
* Reduces heap allocations per kubernetes middleware query. Based on compile time escape analysis - not run time profiling.

Makefile.release
```diff
@@ -38,7 +38,7 @@ NAME:=coredns
 VERSION:=$(shell grep 'coreVersion' coremain/version.go | awk '{ print $$3 }' | tr -d '"')
 ARCH:=$(shell uname -m)
 GITHUB:=coredns
-DOCKER:=coredns
+DOCKER:=chrisohaver
```
johnbelamaric: This file shouldn't be included in the PR.

Makefile
```diff
@@ -15,7 +15,7 @@ deps: core/zmiddleware.go core/dnsserver/zdirectives.go
 	go get -u github.com/golang/lint/golint
 
 .PHONY: check
-check: fmt deps
+check: deps
```
johnbelamaric: this file shouldn't be in here either

chrisohaver: ugh... i wonder what else snuck in... My intent was to only commit ns.go and kubernetes.go

chrisohaver:
Looking into the pod lookup related integration test failures.
Pod lookups (both insecure and verified) work when i do manual testing in a real k8s deployment.
Not yet sure why it fails in travis integration.

Makefile
chrisohaver: OK - mystery solved... at some point I accidentally dropped the "docker" target off a  `make -f Makefile.release docker` command ...
`make -f Makefile.release` automatically does a `git commits -am"Release $(VERSION)"` ...

johnbelamaric:
Ah ha!

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Handle K8s mi..." ]
>
>
>
>> @@ -15,7 +15,7 @@ deps: core/zmiddleware.go core/dnsserver/zdirectives.go
> 	go get -u github.com/golang/lint/golint
>
> .PHONY: check
>-check: fmt deps
>+check: deps
>
>OK - mystery solved... at some point I accidentally dropped the "docker" target off a  `make -f Makefile.release docker` command ...
>`make -f Makefile.release` automatically does a `git commits -am"Release $(VERSION)"` ...

Ai, that's worth fixing

johnbelamaric:
@miekg any thoughts on this comment from earlier:

>> ah, we also claiming a namespace here. Should we rename 'dns' to 'kube-system'?

>Ideally, SOA would return the actual domain name of the dns service. But currently, backend_lookup SOA returns a hard-coded record to ns.dns.[zone], which means that all middleware needs to add handle an A record request for ns.dns.[zone].

I think that's the only outstanding question here.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Handle K8s mi..." ]
>@miekg any thoughts on this comment from earlier:
>
>>> ah, we also claiming a namespace here. Should we rename 'dns' to 'kube-system'?
>
>>Ideally, SOA would return the actual domain name of the dns service. But currently, backend_lookup SOA returns a hard-coded record to ns.dns.[zone], which means that all middleware needs to add handle an A record request for ns.dns.[zone].
>
>I think that's the only outstanding question here.

I think we should not claim another namespace and default to kube-system as that
would interfere least with the rest of the system? WDYT?

chrisohaver:
I think the PR is fine as is in this regard.

`dns` is not a valid object type (valid types are `svc` or `pod`), so there is no chance of namespace collision. `myservice.ns.dns.[zone].` cannot reference a valid kubernetes service.

The only trouble that the hard coded SOA response presents is that every middleware has to implement/handle an A record for that static entry.  This is already added for K8s (in this PR).  The dns service will have two A records...
- one at `ns.dns.zone.`
- the other typically at `kube-dns.kube-system.svc.zone.` 

The PTR record points to `kube-dns.kube-system.svc.zone.`

johnbelamaric:
Ok, so from a k8s perspective there isn't really an issue here. `ns.dns.cluster.local.` doesn't conflict with any other record that k8s middleware should handle. @miekg from a DNS perspective is there an issue with the PTR not pointing to the A record that matches the SOA?

I think we probably just need a separate issue to discuss/decide whether we want CoreDNS to hard code the SOA name and force an A record with that name.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Handle K8s mi..." ]
>Ok, so from a k8s perspective there isn't really an issue here. `ns.dns.cluster.local.` doesn't conflict with any other record that k8s middleware should handle. @miekg from a DNS perspective is there an issue with the PTR not pointing to the A record that matches the SOA?

Ok. Let's use 'dns'. Nothing should check a PTR to a SOA's MNAME (I think)

>I think we probably just need a separate issue to discuss/decide whether we want CoreDNS to hard code the SOA name and force an A record with that name.

Yes.

johnbelamaric (approved):

johnbelamaric:
@miekg I think this is ready - it doesn't hit the diff coverage number because of improvements to code that is tested by the integration but would require mocking the API to cover it in UT.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Handle K8s mi..." ]
>@miekg I think this is ready - it doesn't hit the diff coverage number because of improvements to code that is tested by the integration but would require mocking the API to cover it in UT.

Ack. I'll give it a review in ~2 hours

middleware/kubernetes/controller.go
```diff
@@ -35,6 +35,14 @@ func (s *storeToNamespaceLister) List() (ns api.NamespaceList, err error) {
 	return ns, nil
 }
 
+type dnsControllerItf interface {
```
miekg: Naming: Drop the Itf and just leave it as `dnsController` and rename the type below to `dnsControl`

middleware/kubernetes/kubernetes.go
```diff
@@ -234,7 +248,15 @@ func (k *Kubernetes) parseRequest(lowerCasedName, qtype string) (r recordRequest
 		}
 	}
 	if r.zone == "" {
-		return r, errors.New("zone not found")
+		return r, errZoneNotFound
+	}
+
+	if qtype == "NS" {
```
miekg: Most dns type comparisons use dns.TypeXX (`dns.TypeNS`), how did we end up with a string here?

middleware/kubernetes/ns.go
```diff
@@ -0,0 +1,111 @@
...
+		Host: ns.A.String(),
+		Key:  msg.Path(strings.Join([]string{ns.Hdr.Name, r.zone}, "."), "coredns")}
+	*svcs = append(*svcs, s)
+	return nil
+}
+
+// DefaultNSMsg returns an msg.Service representing an A record for
+// ns.dns.[zone] -> dns service ip. This A record is needed to legitimize
+// the SOA response in middleware.NS(), which is hardcoded at ns.dns.[zone].
+func (k *Kubernetes) DefaultNSMsg(r recordRequest) msg.Service {
```
miekg: Do these methods all need to be exported?

middleware/kubernetes/ns.go
```diff
@@ -0,0 +1,111 @@
...
+	return strings.Index(name, DefaultNSName) == 0 && strings.Index(name, r.zone) == len(DefaultNSName)
+}
+
+type ifAddrsFunc func() ([]net.Addr, error)
+
+func (k *Kubernetes) CoreDNSRecord() dns.A {
+	return k.DoCoreDNSRecord(net.InterfaceAddrs)
+}
+
+func (k *Kubernetes) DoCoreDNSRecord(ifAddrs ifAddrsFunc) dns.A {
```
miekg: Drop the `Do' and lowercase `Core`.

Why the `CoreDNSRecord()` -> `DoCoreDNSRecord` indirection? `DoCoreDNSRecord` is useful by itself?

miekg (commented):
Looks good, some nits and comments.

middleware/kubernetes/ns.go
chrisohaver: That was for quick and dirty stubbing.  I have replaced it with a more idiomatic struct wrapper and interface.

middleware/kubernetes/controller.go
chrisohaver: I did consider naming the interface dnsControllerer. :wink:

miekg (approved):

miekg:
Test fail, pasting here because I just took a look:
~~~
--- FAIL: TestKubernetesIntegrationFallthrough (1.04s)
	kubernetes_test.go:436: Expected rcode 0 but got 2 for query cluster.local., 2
	kubernetes_test.go:440: Expected 1 answers but got 0 for query cluster.local., 
~~~

Issue #66, opened 2016-04-02
middleware: file: uses RR for lookups


miekg:
For looksups in the tree we use a dns.RR, but we only use the name and type wasting the rest of the header. We should just use qname and qtype for all lookups

PR #664, opened 2017-05-11
Pr 651 1

miekg:
@ericyan This is what I meant, there are some open questions in the code, regarding when you don't find a CNAME to expand and the number of DNAMEs that can be present at a point.
This means we can also drop some of the added code to search for DNAMEs because it is not needed.

### DIFF
--- a/middleware/file/lookup.go
+++ b/middleware/file/lookup.go
@@ -48,19 +48,6 @@ func (z *Zone) Lookup(state request.Request, qname string) ([]dns.RR, []dns.RR, 
 		return nsrrs, nil, glue, Success
 	}
 
-	// Invoke DNAME redirection, if any
-	for _, dname := range z.dnames {
-		if cname := synthesizeCNAME(qname, dname); cname != nil {
-			answer, ns, extra, rcode := z.searchCNAME(state, nil, []dns.RR{cname})
-
-			// The relevant DNAME RR should be included in the answer section,
-			// if The DNAME is being employed as a substitution instruction.
-			answer = append([]dns.RR{dname}, answer...)
-
-			return answer, ns, extra, rcode
-		}
-	}
-
 	var (
 		found, shot    bool
 		parts          string
@@ -135,7 +122,13 @@ func (z *Zone) Lookup(state request.Request, qname string) ([]dns.RR, []dns.RR, 
 	// Found entire name.
 	if found && shot {
 
-		// DNAME...?
+		// Ignore records at any subdomain of the owner of a DNAME RR.
+		if z.hasDNAME && qtype != dns.TypeDNAME {
+			if answer, ns, extra, rcode := z.searchDNAME(state); answer != nil {
+				return answer, ns, extra, rcode
+			}
+		}
+
 		if rrs := elem.Types(dns.TypeCNAME); len(rrs) > 0 && qtype != dns.TypeCNAME {
 			return z.searchCNAME(state, elem, rrs)
 		}
@@ -167,6 +160,11 @@ func (z *Zone) Lookup(state request.Request, qname string) ([]dns.RR, []dns.RR, 
 	}
 
 	// Haven't found the original name.
+
+	// Invoke DNAME redirection, if any.
+	if z.hasDNAME {
+		return z.searchDNAME(state)
+	}
 
 	// Found wildcard.
 	if wildElem != nil {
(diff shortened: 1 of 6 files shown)


middleware/file/lookup.go
```diff
@@ -95,6 +95,22 @@ func (z *Zone) Lookup(state request.Request, qname string) ([]dns.RR, []dns.RR,
 			continue
 		}
 
+		// If we see DNAME records, we should return those.
+		if dnamerrs := elem.Types(dns.TypeDNAME); dnamerrs != nil {
+			// There can only be one?
```
ericyan, line 100: Yes, only one DNAME is allowed per name.

middleware/file/lookup.go
```diff
@@ -95,6 +95,22 @@ func (z *Zone) Lookup(state request.Request, qname string) ([]dns.RR, []dns.RR,
...
+			if cname := synthesizeCNAME(state.Name(), dname.(*dns.DNAME)); cname != nil {
+				answer, ns, extra, rcode := z.searchCNAME(state, elem, []dns.RR{cname})
+
+				// The relevant DNAME RR should be included in the answer section,
+				// if the DNAME is being employed as a substitution instruction.
+				answer = append([]dns.RR{dname}, answer...)
+
+				return answer, ns, extra, rcode
+			}
+			// What should happen here? Just the DNAME w/ SIG?
```
ericyan, line 111: The domain name that owns a DNAME record is allowed to have other RR types at that domain name, except those have restrictions on what they can coexist with (e.g. another DNAME). So there is nothing special left here.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Pr 651 1 (#66..." ]
>ericyan commented on this pull request.
>
>
>
>> @@ -95,6 +95,22 @@ func (z *Zone) Lookup(state request.Request, qname string) ([]dns.RR, []dns.RR,
> 			continue
> 		}
>
>+		// If we see DNAME records, we should return those.
>+		if dnamerrs := elem.Types(dns.TypeDNAME); dnamerrs != nil {
>+			// There can only be one?
>+			dname := dnamerrs[0]
>+			if cname := synthesizeCNAME(state.Name(), dname.(*dns.DNAME)); cname != nil {
>+				answer, ns, extra, rcode := z.searchCNAME(state, elem, []dns.RR{cname})
>+
>+				// The relevant DNAME RR should be included in the answer section,
>+				// if the DNAME is being employed as a substitution instruction.
>+				answer = append([]dns.RR{dname}, answer...)
>+
>+				return answer, ns, extra, rcode
>+			}
>+			// What should happen here? Just the DNAME w/ SIG?
>
>The domain name that owns a DNAME record is allowed to have other RR types at that domain name, except those have restrictions on what they can coexist with (e.g. another DNAME). So there is nothing special left here.

Ok. cool. Can you merge this PR in yours and put your comment here instead of
mine?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Pr 651 1 (#66..." ]
>ericyan commented on this pull request.
>
>
>
>> @@ -95,6 +95,22 @@ func (z *Zone) Lookup(state request.Request, qname string) ([]dns.RR, []dns.RR,
> 			continue
> 		}
>
>+		// If we see DNAME records, we should return those.
>+		if dnamerrs := elem.Types(dns.TypeDNAME); dnamerrs != nil {
>+			// There can only be one?
>
>Yes, only one DNAME is allowed per name.

I wonder what we need to do when we see multiple DNAME when loading the zone - I
believe we should just allow it and emit a warning here when we see multiple
DNAMEs and just pick the first one. This does mean though that when signed the
signature won't match. Hmm.. what does the spec say about this?

/Miek

--
Miek Gieben

ericyan:
The spec suggests that we should refuse to load the zone if it has multiple DNAMEs for the same owner. It is pretty interesting that they chose the word *ought to* instead of *SHOULD*.

> DNAME is a singleton type, meaning only one DNAME is allowed per name. [...] A server ought to refuse to load a zone that violates these rules.

I agree that it is better to handle this in a graceful manner, but it can be tricky especially when DNSSEC is involved. Maybe we should move further discussions on this to a new issue?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Pr 651 1 (#66..." ]
>The spec suggests that we should refuse to load the zone if it has multiple DNAMEs for the same owner. It is pretty interesting that they chose the word *ought to* instead of *SHOULD*.
>
>> DNAME is a singleton type, meaning only one DNAME is allowed per name. [...] A server ought to refuse to load a zone that violates these rules.
>
>I agree that it is better to handle this in a graceful manner, but it can be tricky especially when DNSSEC is involved. Maybe we should move further discussions on this to a new issue?

Agreed. Let's get this PR merged first.

PR #673, opened 2017-05-25
Fix docker-build target in Makefile.release

chrisohaver:
The typo prevented the `docker-upload` target from executing when doing `make docker`.


miekg:
ah, forgot this when I changed stuff :(

PR #677, opened 2017-05-26
Add k8s external service CNAMEs

chrisohaver:
- `CNAME` requests: `CNAME` records are returned for all k8s services pointing to an external name  
- `A` requests for k8s external services: `CNAME` record returned in addition to an `A` record for the`CNAME` target. `A` record retrieved from upstream server
- Add `upstream` option to k8s middleware config.
- Unit test and Integration tests
- Minor refactoring of msg.HostType(): Change TypeANY return code to TypeCNAME. I think TypeCNAME maps more logically to a host containing a name.
- Updated k8s middleware README

This doesn't yet handle SRV requests for external services.  I'd prefer to defer that to a later PR, if we determine it's needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ go:
 go_import_path: github.com/coredns/coredns
 
 env:
-  - ETCD_VERSION=2.3.1 K8S_VERSION=1.6.4 KUBECTL="docker exec hyperkube /hyperkube kubectl" DNS_ARGUMENTS=""
+  - ETCD_VERSION=2.3.1 K8S_VERSION=1.6.0 KUBECTL="docker exec hyperkube /hyperkube kubectl" DNS_ARGUMENTS=""
 
 # In the Travis VM-based build environment, IPv6 networking is not
 # enabled by default. The sysctl operations below enable IPv6.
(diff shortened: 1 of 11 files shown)


chrisohaver:
Looks like hyperkube version of k8s doesn't support external name services.
I use minikube with a newer k8s locally to test integrations.

chrisohaver:
Yeah - Looks like the ExternalName service type was added in kubernetes 1.4, and travis is using kubernetes 1.3.8.  Looking into how to update...

chrisohaver:
Tried a couple of 1.6 versions, they never started successfully.
Landed on 1.5, which works, but requires port spec declarations for externalname services (not required in 1.6).
Added port spec declarations... build tests passed.

johnbelamaric (approved):

johnbelamaric (changes requested):
Actually, one little thing. Since this feature was (inadvertently) left out of the 1.0.0 DNS schema spec, the TXT record should be changed to return 1.0.1. See https://github.com/kubernetes/dns/pull/89

johnbelamaric (approved):
You're up early 😀

PR #680, opened 2017-05-30
middleware/secondary: fix crash with no zone

### LINKED ISSUE #679: secondary: runtime error: invalid memory address or nil pointer dereference
I'm trying to add secondary zones to my configuration. When trying to resolve a secondary domain I get the following error:
```
30/May/2017:10:10:04 +0000 [PANIC foo.test.local. A] coredns/vendor/github.com/miekg/dns/ztypes.go:539 - runtime error: invalid memory address or nil pointer dereference
30/May/2017:10:10:09 +0000 [PANIC foo.test.local. A] coredns/vendor/github.com/miekg/dns/ztypes.go:539 - runtime error: invalid memory address or nil pointer dereference
30/May/2017:10:10:14 +0000 [PANIC foo.test.local. A] coredns/vendor/github.com/miekg/dns/ztypes.go:539 - runtime error: invalid memory address or nil pointer dereference
```

I'm using CoreDNS version 007 and here is my configuration:
```
.:5353 {
  log stdout
  errors stdout
  health

  secondary foo.local {
    transfer from 10.1.0.10
  }

  proxy . 208.67.220.220 208.67.222.222
  cache 30
}
```

miekg:
When CoreDNS starts up and can't get a zone transfer going the Apex is
empty. This `nil` is then transformed into wireformat, which fails with
a nil pointer dereference in Go DNS.

In this case we should just return SERVFAIL, because we don't have any
info (yet). Note the lookup code returned NXDOMAIN, which is correct
from a lookup standpoint, but also invalidates every name in the future
loaded zone.

Anyway, look for an apex before doing the lookup and return SERVFAIL if
nothing is found.

Fixes #679

### DIFF
--- a/middleware/file/lookup.go
+++ b/middleware/file/lookup.go
@@ -39,6 +39,14 @@ func (z *Zone) Lookup(state request.Request, qname string) ([]dns.RR, []dns.RR, 
 		}
 	}()
 
+	// If z is a secondary zone we might not have transfered it, meaning we have
+	// all zone context setup, except the actual record. This means (for one thing) the apex
+	// is empty and we don't have a SOA record.
+	soa := z.Apex.SOA
+	if soa == nil {
+		return nil, nil, nil, ServerFailure
+	}
+
 	if qtype == dns.TypeSOA {
 		return z.soa(do), z.ns(do), nil, Success
 	}
--- a/test/secondary_test.go
+++ b/test/secondary_test.go
@@ -1,0 +1,47 @@
+package test
+
+import (
+	"io/ioutil"
+	"log"
+	"testing"
+
+	"github.com/coredns/coredns/middleware/proxy"
+	"github.com/coredns/coredns/middleware/test"
+	"github.com/coredns/coredns/request"
+
+	"github.com/miekg/dns"
+)
+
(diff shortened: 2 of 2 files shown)


johnbelamaric:
The coverage report is implying that your tests misses the case it's trying to test, though I don't see anything obvious wrong with the test:

https://codecov.io/gh/coredns/coredns/compare/2f2c90f3917f5d2179874d7415e593e0ff0c0f79...e876f7b61c716e9385a10e1a963d9d3dfd45430c/diff

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/se..." ]
>The coverage report is implying that your tests misses the case it's trying to test, though I don't see anything obvious wrong with the test:
>
>https://codecov.io/gh/coredns/coredns/compare/2f2c90f3917f5d2179874d7415e593e0ff0c0f79...e876f7b61c716e9385a10e1a963d9d3dfd45430c/diff

I think codecov can't deal with the fact the we test some stuff in the top-level
test/ directory as opposed to a `_test.go` file in the middleware's directory.

Decrementing the test coverage in the PR is indeed a bit unfair.

PR #686, opened 2017-06-01
update deps

miekg:


miekg:
@yongtang this is a
~~~
dep ensure
dep prune
~~~
*But* is also pruned k8s' client 1.5 from the tree? Did we manually add that or is this a bug in `dep`?

yongtang:
@miekg The last time I added the vendor packages through `dep` I didn't run `dep prune`. So the removed go-client 1.5 in this PR should be fine (unneeded code are removed).

I run the `dep ensure` + `dep prune` and the result is the same:
https://github.com/coredns/coredns/compare/master...yongtang:vendor

miekg:
Ack. Thanks. I'll merge.

Issue #678, opened 2017-05-29
add middleware/hosts (for reading entries from /etc/hosts file)


ookangzheng:
My settings like this 

proxy . 127.0.0.1:5454 {
        policy round_robin
        fail_timeout 3s
*** lookup custom hosts file (ads block like, 0.0.0.0 ssl.google.com)
        spray
        protocol https_google
}
    cache 160
    log stdout
    errors stdout

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] How can I read my..." ]
>My settings like this
>
>~~~
>proxy . 127.0.0.1:5454 {
>        policy round_robin
>        fail_timeout 3s
>## lookup custom hosts file (ads block like, 0.0.0.0 ssl.google.com)
>        spray
>        protocol https_google
>}
>    cache 160
>    log stdout
>    errors stdout
>~~~

What's the question? You want to be able to block specific sites and getting
that info from the /etc/hosts file? I think we should do that in a separate
middleware, like `hosts` or something that does a lookup in the hosts file
first.

An actual blocking middleware might also be possible, but a `hosts` middleware
seems more useful at the moment, allthough a host file can only deal with a/aaaa
and ptr lookups.


/Miek

--
Miek Gieben

ookangzheng:
Yap, sorry for my english
I have a list of hosts file with 0.0.0.0 xxx.com 
I want it to lookup from hosts first, if match then return 0.0.0.0, else lookup from upstream dns like 8.8.8.8

ookangzheng:
my solution from now I just dnsmasq then forward upstream to coredns

miekg:
I'll open an issue for having a hosts middleware, seems like a nice
addition and maybe someone will have some fun adding it.

On 29 May 2017 12:13 pm, "Oo Kang Zheng" <notifications@github.com> wrote:

my solution from now I just dnsmasq then forward upstream to coredns

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/coredns/coredns/issues/678#issuecomment-304636528>, or mute
the thread

miekg:
Add a new middleware like this is also a good way to get started with coding in CoreDNS

miekg:
Specifically the middleware should:
* parse and reparse /etc/hosts every x seconds or on change
* probably needs `fallthrough` support, just like we did in the kubernetes middleware
* hook up the middleware into CoreDNS

ookangzheng:
[...]

PR #695, opened 2017-06-02
middleware/hosts for /etc/hosts parsing

pmoroney:
This adds a middleware to parse and serve hosts files from issue #678 

I pulled the hostsfile.go logic from https://github.com/golang/go/blob/master/src/net/hosts.go
I left the copyright notice in there, but if there is any additional attribution I need to add, please let me know.

I also left the cache for 5 seconds or rescan logic in from net/hosts.go
I figure that is probably good enough, or I can add a duration parameter for customizing the cache time.

I also didn't have the backend config accept a list of zones, but that should be easy enough for me to add if we want it.

Of course any suggestions are appreciated.


pmoroney:
Just realized I got PTR handling wrong while looking at the kubernetes middleware, I'll fix that.
I am also looking at adding the Fallthrough support.

I will have some more time this weekend hopefully.

miekg:
Np. I'll give this a first glance over today.

middleware.cfg
```diff
@@ -39,8 +39,9 @@
 170:auto:auto
 180:secondary:secondary
 190:etcd:etcd
-200:proxy:proxy
-210:whoami:whoami
-220:erratic:erratic
+200:hosts:hosts
```
miekg: We might want to put this early in the chain.

middleware/hosts/setup.go
```diff
@@ -0,0 +1,76 @@
...
+	})
+}
+
+func setup(c *caddy.Controller) error {
+	a, err := hostsParse(c)
+	if err != nil {
+		return middleware.Error("hosts", err)
+	}
+
+	// If we have enabled prometheus we should add newly discovered zones to it.
```
miekg: Don't think you'll need this, because you are not exporting metrics.

middleware/hosts/testdata/case-hosts
```diff
@@ -0,0 +1,2 @@
+127.0.0.1	PreserveMe	PreserveMe.local
```
miekg: Can we put these files into the source and make `Hostsfile` read from the strings? `strings.NewReader` or similar can be used.

We currently don't have any test data artifacts in the coredns code base which is something I like. Look in the file middleware to see how it is done.

middleware/hosts/hosts.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/coredns/coredns/request"
+	"github.com/miekg/dns"
+)
+
+type (
+	Hosts struct {
+		Next middleware.Handler
+		*Hostsfile
+
+		metrics *metrics.Metrics
```
miekg: metrics can be removed.

middleware/hosts/hosts.go
```diff
@@ -0,0 +1,104 @@
...
+			return middleware.NextOrFailure(h.Name(), h.Next, ctx, w, r)
+		}
+		for _, n := range names {
+			r := new(dns.PTR)
+			r.Hdr = dns.RR_Header{Name: zone, Rrtype: dns.TypePTR,
+				Class: dns.ClassINET, Ttl: 3600}
+			r.Ptr = n
+			answers = append(answers, r)
+		}
+	case dns.TypeA:
```
miekg: these address cases do a lot of the same work and this can probably be refactored a little.

miekg (changes requested):
This looks good already, add some comments.

middleware.cfg
pmoroney: I moved this up.

middleware/hosts/setup.go
pmoroney: Removed

middleware/hosts/testdata/case-hosts
pmoroney: I made them stings and separated the parsing of the hosts file from the opening of the file and the caching portion.

middleware/hosts/hosts.go
pmoroney: Removed

middleware/hosts/hosts.go
pmoroney: Working on this now as well as the PTR fixing.

pmoroney:
I think I addressed most of the issues.
I still need to add the Fallthrough setting, a list of zones in the config, and add comments.
I am hoping I will have some time tomorrow to get to these.

miekg:
Looks good, only need a README.md explaining the middleware, see other examples and `middleware.md` for some hand holding.

Another thing: do we need the `file` keyword in the config? Or just `hosts` with an optional file name that defaults to `/etc/hosts`?

pmoroney:
something like the following?
hosts [FILE] [ZONES...]

pmoroney:
Also right now it defaults to fallthrough to the next middleware. Do we want to make that optional?
And if the question is for a non-A,AAAA,PTR type right now it falls through, should it return a NOERROR with 0 answers for zones that do exist in the hosts file?
And if so, should there be a NextOrNoerror() function?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/ho..." ]
>Also right now it defaults to fallthrough to the next middleware. Do we want to make that optional?

Yes, we've decided this needs to be explicit, by using a 'fallthrough' keyword
in the config.

>And if the question is for a non-A,AAAA,PTR type right now it falls through, should it return a NOERROR with 0 answers for zones that do exist in the hosts file?

In the case of 'no fallthrough'? It should return refused, just let the
middleware return dns.RcodeRefused and CoreDNS will send the error back to the
client.

>And if so, should there be a NextOrNoerror() function?

Your middleware is different in the a /etc/hosts can basically contain any zone,
which means if you didn't find anything:
* with fallthrough; call next middleware, if none I think SERVFAIL is the
  correct response (it might be REFUSED, but I think SERVFAIL is OK as well).
* no fallthrough, could work with SERVFAIL as well, I think.


/Miek

--
Miek Gieben

pmoroney:
Ok, I added fallthrough as an explicit setting.
I also updated the README.md and tests accordingly.

Thanks for all of your advice and coaching!
Let me know if there is anything I have missed.

middleware/hosts/setup.go
```diff
@@ -0,0 +1,88 @@
...
+		Hostsfile: &Hostsfile{path: "/etc/hosts"},
+	}
+	defer h.ReadHosts()
+
+	config := dnsserver.GetConfig(c)
+
+	for c.Next() {
+		if c.Val() == "hosts" { // hosts [FILE] [ZONES...]
+			args := c.RemainingArgs()
+			if len(args) == 0 {
```
miekg: This check seems wrong as both FILE and ZONES are optional

middleware/hosts/setup.go
```diff
@@ -0,0 +1,88 @@
...
+
+			if !path.IsAbs(h.path) && config.Root != "" {
+				h.path = path.Join(config.Root, h.path)
+			}
+			_, err := os.Stat(h.path)
+			if err != nil {
+				if os.IsNotExist(err) {
+					log.Printf("[WARNING] File does not exist: %s", h.path)
+				} else {
+					return h, c.Errf("Unable to access hosts file '%s': %v", h.path, err)
```
miekg: nit: errors should not start with a capital letter.

middleware/hosts/hosts.go
```diff
@@ -0,0 +1,178 @@
...
+		names := h.LookupStaticAddr(dnsutil.ExtractAddressFromReverse(zone))
+		if len(names) == 0 {
+			// If this doesn't match we need to fall through regardless of h.Fallthrough
+			return middleware.NextOrFailure(h.Name(), h.Next, ctx, w, r)
+		}
+		answers = h.ptr(zone, names)
+	}
+
+	if zone != "" {
+		ips := h.LookupStaticHost(zone)
```
miekg: Wouldn't it be better to let `LookupStatisHost` returns `net.IP`s? Now we keep on parsing the same string over and over again in `ipv4Filter` and `ipv6Filter`...

miekg: Are even returns multiple sets; v4 and v6

middleware/hosts/hosts.go
```diff
@@ -0,0 +1,178 @@
...
+	state := request.Request{W: w, Req: r}
+	if state.QClass() != dns.ClassINET {
+		return dns.RcodeServerFailure, middleware.Error(h.Name(), errors.New("can only deal with ClassINET"))
+	}
+	qname := state.Name()
+
+	answers := []dns.RR{}
+
+	// Precheck with the origins, i.e. are we allowed to looks here.
+	if h.Origins != nil {
```
miekg: Should origins *always* have something?

miekg: I'm also having trouble understanding, why there is precheck? Most of this code is repeated below?

miekg (changes requested):
I check out the PR to play with it, but I couldn't get an answer for a name that I have in /etc/hosts. I.e. with a Corefile of:
~~~
.1053 {
   hosts
   chaos
}
~~~
And then `dig @localhost -p 1053 localhost A` it just hanged for me.
Reviewed the current PR before I delve into the code and debug this.

middleware/hosts/hosts.go
pmoroney: My assumption was if you are specifying the zones in the corefile, and the query is for a zone that is not in that list, then we want to failover to the next middleware even if failover is not specified. Is that assumption correct, or is there another way to handle that case?

Or, as I just realized may happen, does the c.ServerBlockKeys call return a zone regardless, "." perhaps? I'll test that out later tonight.

middleware/hosts/hosts.go
pmoroney: I thought about that initially but had erred on the side of modifying that file as little as possible. But, now that I have modified it quite a bit for the string vs file processing anyways, it makes more sense to modify it to return exactly what we need. I'll refactor this tonight.

middleware/hosts/setup.go
pmoroney: Yeah, this does look off. I think the cause of your issue above may be related to the changes I made to add the Fallthrough config. I'll look into this tonight and add more tests.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/ho..." ]
>Yeah, this does look off. I think the cause of your issue above may be related to the changes I made to add the Fallthrough config. I'll look into this tonight and add more tests.

Added an actual Serve test makes sense as well, see TestLookupDNAME in the file
middleware for inspiration.

/Miek

--
Miek Gieben

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/ho..." ]
>pmoroney commented on this pull request.
>
>
>
>> +	Origins     []string
>+	Fallthrough bool
>+}
>+
>+// ServeDNS implements the middleware.Handle interface.
>+func (h Hosts) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
>+	state := request.Request{W: w, Req: r}
>+	if state.QClass() != dns.ClassINET {
>+		return dns.RcodeServerFailure, middleware.Error(h.Name(), errors.New("can only deal with ClassINET"))
>+	}
>+	qname := state.Name()
>+
>+	answers := []dns.RR{}
>+
>+	// Precheck with the origins, i.e. are we allowed to looks here.
>+	if h.Origins != nil {
>
>My assumption was if you are specifying the zones in the corefile, and the query is for a zone that is not in that list, then we want to failover to the next middleware even if failover is not specified. Is that assumption correct, or is there another way to handle that case?

Yes, that assumption is correct, but each middleware is always responsible for a
set of zones. That set can't be empty.

>Or, as I just realized may happen, does the c.ServerBlockKeys call return a zone regardless, "." perhaps? I'll test that out later tonight.

The `Matches` check you do, check if the incoming query should be handled by
this middleware or directly given to the next middleware. Ideally this check
should only happen once in a middleware.

/Miek

--
Miek Gieben

pmoroney:
Just getting started on this, but was that supposed to be a colon as the first character of that corefile?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/ho..." ]
>Just getting started on this, but was that supposed to be a colon as the first character of that corefile?

No a dot: `.`

~~~
.:1053 {
    chaos
    hosts
}
~~~
from the top of my head. I.e. be authoritative for every domain name, this is
needed because most names in a /etc/hosts file are not qualified and basically
top-level domains, e.g. `localhost`.

/Miek

--
Miek Gieben

pmoroney:
I think I have it figured out now.
The problem was rooted in Hostsfile using a host normalization that didn't add the final . to dotless hostnames.
In my refactoring to make Hostsfile more efficient, I now use middleware.Name.Normalize() which fixed the problem.

miekg:
Just tried it again. Works nice! With and without fallthrough. I'll give it another pass-over.

miekg (approved):

PR #699, opened 2017-06-02
middleware/proxy: silence gprc errors

miekg:
Add discard logger


miekg:
@johnbelamaric should we silence glog stuff in coredns itself as well?

miekg:
silenced this in coredns

johnbelamaric:
With this we won't know if the connection is lost. Seems like it could lead to difficulties debugging in production if there is a disconnection.

PR #707, opened 2017-06-04
middleware/file: don't reload zone when SOA isn't changed

### LINKED ISSUE #415: middleware/file: dont reload zone with no SOA change
Right now we just reload - should check the SOA and if not changed -> do nothing.

miekg:
Give Parse an extra argument which is the SOA's serial, if > 0 we check
against the just parsed SOA and then just return.

Fixes #415

### DIFF
--- a/middleware/file/file.go
+++ b/middleware/file/file.go
@@ -110,8 +110,8 @@ func (f File) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (i
 func (f File) Name() string { return "file" }
 
 // Parse parses the zone in filename and returns a new Zone or an error.
-// If serial >= 0 it will not reload the zone, it returns an error indicating, nothing
-// was read.
+// If serial >= 0 it will reload the zone, if the SOA hasn't changed
+// it returns an error indicating nothing was read.
 func Parse(f io.Reader, origin, fileName string, serial int64) (*Zone, error) {
 	tokens := dns.ParseZone(f, dns.Fqdn(origin), fileName)
 	z := NewZone(origin, fileName)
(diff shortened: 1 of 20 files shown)


miekg:
Can you review @chrisohaver ?

chrisohaver:
Yes - I'll review it today.

middleware/file/file.go
```diff
@@ -109,14 +110,26 @@ func (f File) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (i
...
 	for x := range tokens {
 		if x.Error != nil {
-			log.Printf("[ERROR] Failed to parse `%s': %v", origin, x.Error)
 			return nil, x.Error
 		}
+
+		if !seenSOA && serial >= 0 {
+			if s, ok := x.RR.(*dns.SOA); ok {
+				if s.Serial == uint32(serial) { // same zone
+					return nil, fmt.Errorf("no change in seral: %d", serial)
```
chrisohaver: seral -> serial

middleware/file/file.go
```diff
@@ -109,14 +110,26 @@ func (f File) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (i
 func (f File) Name() string { return "file" }
 
 // Parse parses the zone in filename and returns a new Zone or an error.
-func Parse(f io.Reader, origin, fileName string) (*Zone, error) {
+// If serial >= 0 it will not reload the zone, it returns an error indicating, nothing
```
chrisohaver: I think the comment doesn't match the logic.  If serial >= 0 it will not reload the zone, _if the serial is unchanged from last load._

chrisohaver (commented):
LGTM.  Minor cosmetic issue in error message and a wording in a comment.

middleware/file/file.go
miekg: Done

middleware/file/file.go
miekg: done

chrisohaver (approved):

PR #710, opened 2017-06-06
WIP: Cache prefetch

miekg:


miekg:
@johnbelamaric could use some feedback here.

To track the frequency of incoming requests, I though of using a `*Fifo` per name+type pair and get the events in the last minute. If threshold exceeded -> prefetch. This will work I think. However:
This is unbounced memory use: for every new name/type pair I create a new `*Fifo`, is there a smarting thing I can do here?

Thanks!

johnbelamaric:
Ok, I gave it some thought. This might work.

What we really care about is that the cached item is in frequent use and used recently. We don't have to have a precise measure like "it was used x times in the last minute". Relaxing that a bit we could do this:

* For each element in the cache, add a `last_used` timestamp and `recent_hits` uint8.
* When we first store something in the cache, set `last_used` to now and the `recent_hits` to 1.
* Each time we get a cache hit, if the `last_used` is less than one minute ago we increment `recent_hits` by 1 (up to a max of threshold + 1 so we keep it < 255), otherwise we reset it to 1. We also update `last_used` to "now".

Here `recent_hits` isn't really the number in the last minute. But it will only grow if you are getting queries at least one per minute for that entry; it will grow faster if you get them even more often. So, if it gets above a threshold, then we could consider it reasonable to prefetch that entry when the TTL approaches, since we know we'll be getting at least one query for it soon. 

Instead of a timestamp you could use "seconds since startup" and a uint32 if you want to cut the size a little more. We also may want to make the "one minute" tunable.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: Cache pr..." ]
>Ok, I gave it some thought. This might work.
>
>What we really care about is that the cached item is in frequent use and used recently. We don't have to have a precise measure like "it was used x times in the last minute". Relaxing that a bit we could do this:
>
>* For each element in the cache, add a `last_used` timestamp and `recent_hits` uint8.
>* When we first store something in the cache, set `last_used` to now and the `recent_hits` to 1.
>* Each time we get a cache hit, if the `last_used` is less than one minute ago we increment `recent_hits` by 1 (up to a max of threshold + 1 so we keep it < 255), otherwise we reset it to 1. We also update `last_used` to "now".
>
>Here `recent_hits` isn't really the number in the last minute. But it will only grow if you are getting queries at least one per minute for that entry; it will grow faster if you get them even more often. So, if it gets above a threshold, then we could consider it reasonable to prefetch that entry when the TTL approaches, since we know we'll be getting at least one query for it soon.
>

Yes, awesome. This will work (*and* neatly ties into the cache we already have).

>Instead of a timestamp you could use "seconds since startup" and a uint32 if you want to cut the size a little more. We also may want to make the "one minute" tunable.

time.Time is quite readable, so I rather leave it a that. Making the '1 minute'
tuneable is def. something I should do.

Thanks!

/Miek

--
Miek Gieben

PR #711, opened 2017-06-06
middleware/dnstap add

2buisson:
Add a dnstap middleware, not to be merged!


johnbelamaric:
By convention we stick `WIP` (for work-in-progress) in front of PRs that we don't want someone to merge yet.

miekg:
I will review this on <strike>Saturday (or maybe tomorrow).</strike>now

middleware/dnstap/handler.go
```diff
@@ -0,0 +1,20 @@
+package dnstap
+
+import (
```
miekg, line 3: these import need to be sorted:
* stdlib, newline
* coredns ones, newline
* third party ones

middleware/dnstap/handler.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/coredns/coredns/middleware"
+	lib "github.com/dnstap/golang-dnstap"
```
miekg: rename lib to dnstap or tap

middleware/dnstap/handler.go
```diff
@@ -0,0 +1,20 @@
...
+	"golang.org/x/net/context"
+)
+
+type Dnstap struct {
+	out  *lib.FrameStreamOutput
+	Next middleware.Handler
+}
+
+func (h Dnstap) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	catch := ResponseWriter{w, h.out.GetOutputChannel()}
```
miekg: catch -> taprw ?

middleware/dnstap/handler.go
```diff
@@ -0,0 +1,20 @@
...
+	_, err := middleware.NextOrFailure(h.Name(), h.Next, ctx, &catch, r)
```
miekg: Why do you take the address of catch here?

middleware/dnstap/msg/msg.go
```diff
@@ -0,0 +1,78 @@
+package msg
+
+import (
+	"github.com/coredns/coredns/request"
+	lib "github.com/dnstap/golang-dnstap"
+	"github.com/pkg/errors"
+	"time"
+)
+
+type Msg struct {
```
miekg: Needs a comment

middleware/dnstap/msg/msg.go
```diff
@@ -0,0 +1,78 @@
...
+	Type        lib.Message_Type
+	Packed      []byte
+	SocketProto lib.SocketProtocol
+	SocketFam   lib.SocketFamily
+	Address     []byte
+	Port        uint32
+	TimeSec     uint64
+}
+
+func now(m *Msg) {
```
miekg: Would rename this to `epoch`

middleware/dnstap/msg/msg.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func NewClientResponse(state *request.Request, pack bool) (
+	*lib.Message, error) {
+
+	m := &Msg{
+		Type: lib.Message_CLIENT_RESPONSE,
+	}
+	if err := networkFromRequest(m, state); err != nil {
+		return nil, errors.Wrap(err, "network")
```
miekg: Why errors.Wrap?

middleware/dnstap/msg/msg.go
```diff
@@ -0,0 +1,78 @@
...
+		if err != nil {
+			return nil, errors.Wrap(err, "pack")
+		}
+		m.Packed = data
+	}
+	socket(m, state)
+	now(m)
+	return toResp(m), nil
+}
+func timeSec(dest **uint64) {
```
miekg: almost identical to now, except for the `*dest = &buf`, should be possible to life without this.

middleware/dnstap/msg/msg.go
```diff
@@ -0,0 +1,78 @@
...
+	buf := uint64(time.Now().Unix())
+	*dest = &buf
+}
+func socket(m *Msg, r *request.Request) {
```
miekg: socket -> family or socketFamily?

middleware/dnstap/msg/msg.go
```diff
@@ -0,0 +1,78 @@
...
+	if r.Family() == 2 {
+		m.SocketFam = lib.SocketFamily_INET6
+	}
+
+	m.SocketProto = lib.SocketProtocol_UDP
+	if r.Proto() == "tcp" {
+		m.SocketProto = lib.SocketProtocol_TCP
+	}
+}
+func networkFromRequest(m *Msg, r *request.Request) error {
```
miekg: I think its nicer to *not* give `m *Msg` to this function, but let this return the things you're interested in and then assign the things you need to the fieds of m?

middleware/dnstap/msg/msg.go
```diff
@@ -0,0 +1,78 @@
...
+	ip, port, err := ipPort(r.W.RemoteAddr())
```
miekg: why do you need ipPort?

miekg: You should be able to use a lot of stuff from `request.Request` and if not we should extend that package a little.

middleware/dnstap/msg/wrapper.go
```diff
@@ -0,0 +1,24 @@
+package msg
+
+import (
+	lib "github.com/dnstap/golang-dnstap"
+	"github.com/golang/protobuf/proto"
+	"github.com/pkg/errors"
+)
+
+func wrap(m *lib.Message) *lib.Dnstap {
```
miekg: Better named `New`?

middleware/dnstap/setup.go
```diff
@@ -0,0 +1,43 @@
...
+	})
+}
+func setup(c *caddy.Controller) error {
+	c.Next() // 'dnstap'
+	if c.NextArg() {
+		return middleware.Error("dnstap", c.ArgErr())
+	}
+
+	tap := Dnstap{}
+	out, err := lib.NewFrameStreamOutputFromFilename("/tmp/db")
```
miekg: I wonder how useful this is as opposed to a socket. Also this name should be configurable as having a default name that a process with root access creates can lead to security issues.

middleware/dnstap/setup.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		return errors.Wrap(err, "output")
+	}
+	tap.out = out
+
+	go out.RunOutputLoop()
+
+	c.OnShutdown(func() error {
+		out.Close()
```
miekg: This closes to above started go-routine?

middleware/dnstap/writer.go
```diff
@@ -0,0 +1,30 @@
...
+	dns.ResponseWriter
+	log chan<- []byte
+}
+
+func (w *ResponseWriter) WriteMsg(resp *dns.Msg) error {
+	state := request.Request{
+		Req: resp,
+		W:   w,
+	}
+	m, err := msg.NewClientResponse(&state, true)
```
miekg: We usually send state between function as-is, not as a pointer.

miekg (commented):
Starting to look good. I've left a whole bunch of comments. Don't feel intimidated, if something wasn't clear I left a comment or question, even if it was a small nit.

2buisson:
I did a lot of changes since then, I will try to follow your directives before pushing them, thanks.

miekg:
>I did a lot of changes since then, I will try to follow your directives before pushing them, thanks.

Ack. Overall looks good. Let us know when it is ready for a review.

/Miek

--
Miek Gieben

2buisson:
Here it is:
* I am wondering if this is the best approch: [writer.go](https://github.com/varyoo/coredns/blob/919bd9dd67d987b4f60cdcf6da2f4ad9442e8a92/middleware/dnstap/writer.go#L28), should it rather be done before the response is written back to the client, or asynchronously (a lot more of issues here).
* dnstap database VS socket: [README.md](https://github.com/varyoo/coredns/blob/919bd9dd67d987b4f60cdcf6da2f4ad9442e8a92/middleware/dnstap/README.md)

miekg:
you should do it synchronise. The goroutine itself runs async so there is not need to do this again.

I also prefer the socket option. But you're saying you can't read while we write? I don't fully understand that.

johnbelamaric:
If it's synchronous it will add some latency, I would think. But perhaps not enough to matter.

2buisson:
When the connection allows reading and writing, a bidirectional framestream can be used: before the client sends its data, the protocol ensure that the other party can read this data, by some read/write on both sides.

When using a socket the dnstap tool expects a bidirectional framestream. Only the bidirectional framestream decoder has been ported from the C library to the Go one, the encoder isn't there yet https://github.com/farsightsec/golang-framestream/pull/4.

miekg:
Ack. Thanks for the update.

middleware/dnstap/README.md
```diff
@@ -1,8 +1,22 @@
...
 go get github.com/dnstap/golang-dnstap
 cd $GOPATH/src/github.com/dnstap/golang-dnstap/dnstap
 go build
 ./dnstap -r /tmp/db
+./dnstap -r /tmp/db -y
 ```
+
+# dndstap binary database VS socket
+
+The database can become hudge with time, and cannot be cleared while the server is running that
```
alkino: huge

middleware/dnstap/README.md
```diff
@@ -1,8 +1,22 @@
...
+easly. This is why I believe the socket is prefered.
```
alkino: "cannot be cleared easily while the server is running"?

middleware/dnstap/README.md
```diff
@@ -1,8 +1,22 @@
...
+
+However the dnstap golang library https://github.com/dnstap/golang-dnstap dosen't support yet bidirectional framestreams.
```
alkino: doesn't

johnbelamaric:
@varyoo looks like you have made some progress on https://github.com/farsightsec/golang-framestream/pull/4 - nice! What's the next step?

2buisson:
I would appreciate a review at this point before I start writing tests.

middleware/dnstap/README.md
```diff
@@ -0,0 +1,18 @@
+# Directive
```
miekg: This README.md should follow the spec laid out in middleware.md

middleware/dnstap/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+`dnstap /tmp/dnstap.sock true`
```
miekg: Is this still current? What does `true` mean?

middleware/dnstap/handler.go
```diff
@@ -0,0 +1,39 @@
...
+func (h ClientTap) TapMessage(m *tap.Message) error {
+	frame, err := msg.Marshal(m)
+	if err != nil {
+		return errors.Wrap(err, "marshal")
+	}
+	_, err = h.Out.Write(frame)
+	return err
+}
+
+func (h ClientTap) ServeDNS(ctx context.Context, w dns.ResponseWriter,
```
miekg: Go doesn't have a line width, just put this on 1 line

middleware/dnstap/handler.go
```diff
@@ -0,0 +1,39 @@
...
+	r *dns.Msg) (int, error) {
+
+	taprw := ResponseWriter{ResponseWriter: w, Tap: &h, Query: r, Pack: h.Pack}
+	taprw.QueryEpoch()
```
miekg: `QueryEpoch()` seems to *set* the epoch value?

middleware/dnstap/handler.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/coredns/coredns/middleware"
+	"github.com/coredns/coredns/middleware/dnstap/msg"
+
+	tap "github.com/dnstap/golang-dnstap"
+	"github.com/miekg/dns"
+	"github.com/pkg/errors"
+)
+
+type ClientTap struct {
```
miekg: Just a nit on the name, but there is also a ServerTap? Otherwise just name this Tap

middleware/dnstap/writer.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/coredns/coredns/middleware/dnstap/msg"
+	"github.com/coredns/coredns/request"
+
+	tap "github.com/dnstap/golang-dnstap"
+	"github.com/miekg/dns"
+	"github.com/pkg/errors"
+)
+
+type Tap interface {
```
miekg: Tap -> Taper

middleware/dnstap/writer.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (w ResponseWriter) QueryEpoch() {
+	msg.Epoch(&w.queryData)
+}
+
+// Write back the response to the client and THEN work on logging the request
+// and response to dnstap.
+// Not sure on how to best report errors with this approch.
+func (w ResponseWriter) WriteMsg(resp *dns.Msg) error {
```
miekg: This method should not log by itself , but just return the error to the caller.

middleware/dnstap/writer.go
```diff
@@ -0,0 +1,79 @@
...
+	req := request.Request{W: w, Req: w.Query}
+	if err := msg.FromRequest(&w.queryData, req); err != nil {
+		return err
+	}
+	if w.Pack {
+		if err := msg.Pack(&w.queryData, w.Query); err != nil {
+			return errors.Wrap(err, "pack")
+		}
+	}
+	return errors.Wrap(
```
miekg: I don't see an immediate need for errors.Wrap here? (or in general)

middleware/dnstap/writer.go
```diff
@@ -0,0 +1,79 @@
...
+}
+func (w ResponseWriter) tapResponse(resp *dns.Msg) error {
+	d := &msg.Data{}
+	msg.Epoch(d)
+	d.Type = tap.Message_CLIENT_RESPONSE
+	req := request.Request{W: w, Req: resp}
+	if err := msg.FromRequest(d, req); err != nil {
+		return err
+	}
+	if w.Pack {
```
miekg: What does `Pack` mean?

miekg: Ah, I see, you need that binary for dnstap.

middleware/dnstap/msg/msg.go
```diff
@@ -0,0 +1,134 @@
...
+	if r.Family() == 2 {
+		d.SocketFam = tap.SocketFamily_INET6
+	}
+
+	d.SocketProto = tap.SocketProtocol_UDP
+	if r.Proto() == "tcp" {
+		d.SocketProto = tap.SocketProtocol_TCP
+	}
+}
+func networkFromWriter(d *Data, w dns.ResponseWriter) error {
```
miekg: Are you used to a lot of C? In Go you should just return the multiple values instead of giving the function a pointer which it modifies. Why can't this function just return the `net.Addr` and let the caller deal with that?

middleware/dnstap/msg/msg.go
```diff
@@ -0,0 +1,134 @@
...
+	switch *m.Type {
+	case tap.Message_CLIENT_QUERY,
+		tap.Message_RESOLVER_QUERY,
+		tap.Message_AUTH_QUERY,
+		tap.Message_FORWARDER_QUERY,
+		tap.Message_TOOL_QUERY:
+		// is query
+		m.QueryTimeSec = &d.TimeSec
+		m.QueryMessage = d.Packed
+	default:
```
miekg: might want to make this explicit instead of `default` and leave `default` when nothing matches.

middleware/dnstap/msg/msg.go
```diff
@@ -0,0 +1,134 @@
...
+		m.ResponseAddress = d.Address
+		m.ResponsePort = &d.Port
+	}
+
+	return &m
+}
+
+// Transform the data into a client response message.
+// Alternative to ToMsg.
+func ToClientResponse(d *Data) *tap.Message {
```
miekg: Drop `Client`, it's cleaner.

middleware/dnstap/msg/msg.go
```diff
@@ -0,0 +1,134 @@
...
+		ResponseTimeSec: &d.TimeSec,
+		ResponseMessage: d.Packed,
+		QueryAddress:    d.Address,
+		QueryPort:       &d.Port,
+	}
+}
+
+// Transform the data into a client query message.
+// Alternative to ToMsg.
+func ToClientQuery(d *Data) *tap.Message {
```
miekg: Drop `Client`, it's cleaner.

middleware/dnstap/msg/util.go
```diff
@@ -0,0 +1,39 @@
...
+	if err != nil {
+		return 0, errors.Wrap(err, "can't convert to int")
+	}
+	if i < 0 {
+		return 0, errors.New("can't be < 0")
+	}
+	return uint32(i), nil
+}
+
+func ipPort(a net.Addr) (ip []byte, port uint32, err error) {
```
miekg: Isn't it somewhat quaranteed the a `.RemoteAddr()` host the correct port ?

middleware/dnstap/msg/util.go
```diff
@@ -0,0 +1,39 @@
...
+	if err != nil {
+		err = errors.Wrap(err, "split ip:port")
+		return
+	}
+	port, err = parsePort(rawport)
+	if err != nil {
+		err = errors.Wrap(err, "port")
+		return
+	}
+	netip := net.ParseIP(rawip)
```
miekg: some here, you *have* a tcp connection... that can only come from a legal IP address, no?

middleware/dnstap/setup.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func wrapSetup(c *caddy.Controller) error {
+	if err := setup(c); err != nil {
+		return middleware.Error("dnstap", err)
+	}
+	return nil
+}
+
+func setup(c *caddy.Controller) error {
```
miekg: We should probably talk about how the *dnstap* directive should look in the Corefile. Can you send a proposal?

miekg (changes requested):
I'm proposing some method signature changes, which would impact any tests you'll write.

middleware/dnstap/msg/msg.go
2buisson: I did this because there could also be `ToResolverResponse` or `ToAuthResponse` in fact one per [event type](https://godoc.org/github.com/dnstap/golang-dnstap#Message_Type), we don't plan on adding these I could remove `Client` yes, your call.

This is the alternative to `ToMsg`: a funcion per type, I wrote both to see what's best.

middleware/dnstap/msg/msg.go
2buisson: I was thinking to combine `networkFromWriter` with for exemple `socketInfoFromResponse`, probably too much boilerplate.

middleware/dnstap/msg/util.go
2buisson: I guess, what does this mean for this function?

middleware/dnstap/msg/util.go
2buisson: Oh I see, if so I should take a `ResponseWriter` instead of the `net.Addr` and let it panic.

middleware/dnstap/msg/msg.go
2buisson: Will do.

middleware/dnstap/writer.go
2buisson: Yes this is the option we discussed, it's also the last shady `true` parameter in the directive parameters.

middleware/dnstap/writer.go
2buisson: I tend to over do it, but in some case it adds a usefull context to errors.

middleware/dnstap/writer.go
2buisson: Found a way to do that.

middleware/dnstap/handler.go
2buisson: The epoch value of the query request, without this method, the info would be lost when `ResponseWriter.Write` is called and logs both the query and response.

middleware/dnstap/handler.go
2buisson: Will rename to Dnstap.

miekg:
>+	return errors.Wrap(
>
>I tend to over do it, but in some case it adds a usefull context to errors.

Just return early with then error, you won't need errors.Wrap then.

/Miek

--
Miek Gieben

miekg:
>Yes this is the option we discussed, it's also the last shady `true` parameter in the directive options.

Ok. We may need to find something more descriptive than 'true' as a keyword in 
the config.

miekg:
>varyoo commented on this pull request.
>
>I did this because there could also be `ToResolverResponse` or `ToAuthResponse` in fact one per [event type](https://godoc.org/github.com/dnstap/golang-dnstap#Message_Type), we don't plan on adding these I could remove `Client` yes, your call.
>
>This is the alternative to `ToMsg`: a funcion per type, I wrote both to see what's best.

Ah, ok, sgtm - leave it like it is.

middleware/dnstap/msg/util.go
2buisson: Or maybe a type assertion:

~~~
if addr, ok := conn.RemoteAddr().(*net.TCPAddr); ok {
    fmt.Println(addr.IP.String())
}
~~~

miekg:
>varyoo commented on this pull request.
>
>> +	}
>+	return uint32(i), nil
>+}
>+
>+func ipPort(a net.Addr) (ip []byte, port uint32, err error) {
>+	rawip, rawport, err := net.SplitHostPort(a.String())
>+	if err != nil {
>+		err = errors.Wrap(err, "split ip:port")
>+		return
>+	}
>+	port, err = parsePort(rawport)
>+	if err != nil {
>+		err = errors.Wrap(err, "port")
>+		return
>+	}
>+	netip := net.ParseIP(rawip)
>
>Or maybe a type assertion:
>
>~~~
>if addr, ok := conn.RemoteAddr().(*net.TCPAddr); ok {
>    fmt.Println(addr.IP.String())
>}
>~~~

Let's recap. You need the string (well bytes actually) of the remote IP so you 
can put that in the proto for dnstap? The above would work, but, I think you'll
need another type asserting to get the IP address (this would then work for TCP 
and UDP).

if addr, ok := conn.RemoteAddr().(*net.TCPAddr); ok {
    if ipaddr, ok := addr.(net.IPAddr); ok {
        fmt.Println(ipaddr.IP.String())
    }
}

(not: not tested)


/Miek

--
Miek Gieben

johnbelamaric:
@varyoo what's the latest status here? I haven't had time to review but I may ask some other folks to take a look.

miekg:
>@varyoo what's the latest status here? I haven't had time to review but I may ask some other folks to take a look.

I believe status is: "ready for another review"

would be good to let someone else than me take a look.

2buisson:
>I believe status is: "ready for another review"

Yes, the middleware is responsible to log client query and responses, it has been done and tested. There is still other [events](https://godoc.org/github.com/dnstap/golang-dnstap#Message_Type) to log to dnstap, but I think it is to be done outside of this middleware *(happy to implement these)*:

~~~
proxy . 192.168.1.1:53 {
    dnstap /tmp/dnstap.sock
}
~~~

Unless a review :blush: tell me otherwise, I don't have more plans regarding this issue.

pmoroney:
I'm currently at GopherCon, but I can do a review on Saturday during the
community day if that's alright.

johnbelamaric:
@varyoo awesome! why are you thinking those other events should be out of this middleware? As opposed to optionally enabled in this middleware?

miekg:
>I'm currently at GopherCon, but I can do a review on Saturday during the
>community day if that's alright.

sgtm! Thank you

2buisson:
>why are you thinking those other events should be out of this middleware? As opposed to optionally enabled in this middleware?

If the proxy middleware is enabled, it will eventually call remote servers, those requests can be logged by dnstap, and they are invisble to the dnstap middleware. So why enable them under the dnstap middleware, since the the actual code can't be in it?

johnbelamaric:
I see. We have a similar thing with tracing, where we want to utilize the Tracer object to push data from other middleware. The dnstap middleware will have all the config info; we would want to re-use that any any sockets it keeps open to push the data from the other middleware.

johnbelamaric:
Is there a reason you are using `pkg/errors`? We don't use it anywhere else, we just use `errors` and `fmt`. Prefer not to add a dependency.

johnbelamaric:
You need to add the dnstap entry to `middleware.cfg`. I thought I saw it in an earlier version of the PR but it's not in this one.

johnbelamaric:
If I don't run the `dnstap` utility first, then I get this message. Instead, it should WARN and then have a goroutine that periodically checks if the socket exists. Or it could create the socket.

What happens if the pipe fills up? We should not block, but perhaps we should warn in this case.

```
ANN-M-JBELAMARIC:coredns jbelamaric$ ./coredns -conf Corefile-erratic
2017/07/15 10:13:51 middleware/dnstap: output: dial unix /etc/dnstap.sock: connect: no such file or directory
ANN-M-JBELAMARIC:coredns jbelamaric$
```

2buisson:
>The dnstap middleware will have all the config info; we would want to re-use that any any sockets it keeps open to push the data from the other middleware.

~~~Go
t := dnsserver.GetMiddleware(c, "trace")
P := &Proxy{Trace: t}
~~~

~~~Go
if p.Trace != nil {
        if t, ok := p.Trace.(trace.Trace); ok {
~~~

There is something else that can be done with a socket: the dnstap command line tool dosen't **yet** *(maybe the C version does)*, but could accept multiple connections! 

The following would be *then* completely transparent to the end user:

~~~Go
dnstap /tmp/dnstap.sock
proxy . 192.168.1.1:53 {
    dnstap /tmp/dnstap.sock
}
~~~

johnbelamaric:
Travis build is failing because dnstap packages aren't in the vendor dir. If you add those too (it's in one of the READMEs) then we can see if Travis passes.

middleware/dnstap/msg/msg.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/coredns/coredns/request"
+
+	tap "github.com/dnstap/golang-dnstap"
+	"github.com/miekg/dns"
+)
+
+// Data helps to build a dnstap Message.
+// It can be transformed into the actual Message using this package.
+type Data struct {
```
pmoroney, line 17: Is there a reason we are not just using the tap.Message struct here? It seems that we are using a subset of that struct.

Also, if we keep the Data struct, we may want to have all the functions below actually be methods of `*Data` since that is the first argument for each of them. i.e. `func (d *Data) FromRequest(r request.Request) error {`

middleware/dnstap/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+```sh
+go get github.com/dnstap/golang-dnstap
+cd $GOPATH/src/github.com/dnstap/golang-dnstap/dnstap
+go build
+./dnstap -u /tmp/dnstap.sock
+./dnstap -u /tmp/dnstap.sock -y
+```
+
+There is a buffer, expect at least 13 requests before the server sends its dnstap messages to the socket.
```
pmoroney, line 20: Should we have something that periodically flushes the buffer, so messages aren't queued for long periods of time on low volume servers? I sent one query and it only flushed to dnstap when I killed coredns an hour later.

pmoroney (changes requested):
Sorry for the delay. 
Besides the two questions I mentioned inline, it works for me.
Although, you should add comments to all exported types and functions. You can get a list by running `golint github.com/coredns/coredns/middleware/dnstap/...`. If you don't have golint, it can be gotten here: `github.com/golang/lint`

Thanks!

middleware/dnstap/msg/msg.go
2buisson: Yes, in dnstap command-line tool's code, the combinaison of fields displayed for each event type falls under the fields described in `msg.Data`. And it is handy because [tap.Message](https://godoc.org/github.com/dnstap/golang-dnstap#Message) fieds are only pointers.

Argeed for the methods.

middleware/dnstap/README.md
2buisson: Yes it could flushed periodically, the framestream library hides the buffer size.

johnbelamaric:
@varyoo, let's move the proxy additions to a separate PR. I'd like to get basic `dnstap` support in, and that could delay it.

I think the only things to address are:

1) Don't fail startup if the socket doesn't exist. Instead either create it or WARN and periodically check if it exists.
2) Periodically flush the buffer.

@coredns/maintainers please chime in if you disagree.

pmoroney:
@johnbelamaric I agree about splitting that to a new PR.
And on top of your things to address, comments should be placed on all exported types and functions so golint stops complaining.

miekg (approved):

miekg:
I haven't looked in detail, but other have, so lgtm from my original review.

miekg:
@varyoo this is not WIP anymore I think.
@pmoroney OK with merging?

pmoroney:
Yep, looks good to me

johnbelamaric:
👍 nice, @varyoo 

you should now rebase the other PR :)

alkino:
nice @varyoo 
But miekg you change the title of PR and not the title of commit. It is written "WIP" in history now. Too late but sad ^^

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/dn..." ]
>nice @varyoo
>But miekg you change the title of PR and not the title of commit. It is written "WIP" in history now. Too late but sad ^^

ah well :)

PR #713, opened 2017-06-06
WIP: middleware/cache: prefetch item upon expunge

miekg:


middleware/cache/README.md
```diff
@@ -32,8 +33,9 @@ cache [TTL] [ZONES...] {
   number of packets we cache before we start evicting (LRU). **TTL** overrides the cache maximum TTL.
 * `denial`, override the settings for caching denial of existence responses, **CAPACITY** indicates the maximum
   number of packets we cache before we start evicting (LRU). **TTL** overrides the cache maximum TTL.
-
-There is a third category (`error`) but those responses are never cached.
+  There is a third category (`error`) but those responses are never cached.
+* `prefetch`, will prefetch popular items when there are about to be expunged from the cache.
```
johnbelamaric: s/there/they are/

middleware/cache/README.md
```diff
@@ -32,8 +33,9 @@ cache [TTL] [ZONES...] {
...
+  Popular means **AMOUNT** queries seen within **DURATION**, which defaults to 1m.
```
johnbelamaric: Popular means **AMOUNT** queries have been seen no gaps of **DURATION** or more between them. **DURATION** defaults to 1m.

middleware/cache/freq/freq.go
```diff
@@ -0,0 +1,39 @@
...
+// If the last time we've seen this entity is within now - d, we increment hits, otherwise
+// we reset hits to 1. It returns the number of hits.
+func (f *Freq) Update(d time.Duration, now time.Time) int {
+	earliest := now.Add(-1 * d)
+	if f.last.Before(earliest) {
+		f.last = now
+		f.hits = 1
+		return f.hits
+	}
+	f.hits++
```
johnbelamaric: Actually what I was thinking is we always update "last" too. Right now, this will build up for 1 minute, then go back to 1, then build up, then go back to 1. You'll only prefetch if you hit the threshold in the minute right before you check for purging it.

Instead, if you update last every time, it means there have been no gaps between queries that are as long a **DURATION**, and that this has persisted for at least **AMOUNT** queries. It's not the same measure as before, but to me at least seems like a valid measure of "popularity".

johnbelamaric (changes requested):

miekg:
>Actually what I was thinking is we always update "last" too. Right now, this will build up for 1 minute, then go back to 1, then build up, then go back to 1. You'll only prefetch if you hit the threshold in the minute right before you check for purging it.
>
>Instead, if you update last every time, it means there have been no gaps between queries that are as long a **DURATION**, and that this has persisted for at least **AMOUNT** queries. It's not the same measure as before, but to me at least seems like a valid measure of "popularity".

You're correct. Let's just say I *forgot* to do that :)

miekg:
Looking through the caching code:
* we use LRU, but we don't expunge anything, so the LRU bit isn't used
* we store the TTL and check if an item is expired when we fetch it from the cache, if expired we re-fetch the item
* Added `freq.Freq` to this adds another LRU layer
* The reason to use a LRU cache wasn't LRU, but having a memory bounded cache the expunges older elements when we push newer ones
* If we create a new cache based on lru (or the original https://github.com/patrickmn/go-cache) we can add random eviction (easy to implement), together with prefetch this should give enough functionality. Random eviction means we don't need to track LRU (we use freq.Freq for that - maybe rename that as well).

The "new" impl. would obviously boil down to map + mutex, but this is closer to SkyDNS and probably faster, allthough I haven't benchmarked anything (but with have @chrisohaver's early k8s benchmark)

I think I will rip out most of the LRU stuff and go back to simpler cache.

WDYT?

johnbelamaric:
Makes sense to me. The earlier benchmarks showed the cache wasn't that efficient the way it is today.

miekg:
Ack I will go ahead and to do that. However #715 takes priority, so I'm not sure if we should do all this for 008.

johnbelamaric:
I agree, I think we should hold 008 for #715

miekg:
closing this for other inflight PRs that incorporate this.

miekg:
closes in favor of #731

PR #718, opened 2017-06-08
cache: add new cache implementation

miekg:
Add Cache impl and a few tests.


miekg:
Actually @johnbelamaric this is all that is need to rip out LRU (completely untested except for our unit tests)

miekg:
@johnbelamaric this is ready for early review. I'll put prefetch in this branch as well. We should merge after 008 release.

miekg:
@johnbelamaric changed it to a 256-way sharded cache impl, so we have 256 mutexes instead of one for cache access.

Please give this a lookover. It *should* improve concurrent throughput.

miekg:
superseded by #731

PR #721, opened 2017-06-08
No `f` in  in line 73 in coredns/middleware/secondary/setup.go

snowirbis:
Cant' transfer zone from masters without populating `f`.
This error prevents secondary zones recognized as "true" secondary, so secondary setting never worked.

### DIFF
--- a/middleware/secondary/setup.go
+++ b/middleware/secondary/setup.go
@@ -70,7 +70,7 @@ func secondaryParse(c *caddy.Controller) (file.Zones, error) {
 
 				switch c.Val() {
 				case "transfer":
-					t, _, e = file.TransferParse(c, true)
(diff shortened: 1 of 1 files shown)


johnbelamaric:
You mentioned you are new to GitHub. Here's hint - you should create a branch in your local clone, and then push that branch to GitHub. Create the PR from the branch. Otherwise, after this merges and then you pull from the coredns upstream to your master, you'll end up with an extra merge commit in your master and they'll always be out of sync.

snowirbis:
Thanks, John! 
Just did not work with git in multiuser environment.

johnbelamaric:
Yeah, takes a bit to get used to the workflow but it makes sense once you do.

PR #723, opened 2017-06-08
middleware/kubernetes: Implement current federation beta

chrisohaver:
This adds support for cross-cluster service discovery of federated services. Specifically, this implements the current beta proposed specification for k8s federation service discovery.

When querying CoreDNS for an A record of a federated service (e.g. service.ns.fed1.svc.cluster.local), coredns first checks the local cluster.  

- If the service is present and healthy, CoreDNS returns an A record for the service.  
- If the service is not present or unhealthy, CoreDNS returns a CNAME record for the service in the federated format (e.g. service.ns.fed1.svc.availability-zone.region.domain). Also included in the response is the CNAME chain and A record from the upstream server.

### DIFF
--- a/middleware/kubernetes/README.md
+++ b/middleware/kubernetes/README.md
@@ -115,6 +115,12 @@ kubernetes coredns.local {
 	# a path to a file structured like resolv.conf.
 	upstream 12.34.56.78:53
 	
+	# federation <federation-name> <federation-domain>
+	#
+	# Defines federation membership.  One line for each federation membership.
+	# Each line consists of the name of the federation, and the domain.
+	federation myfed foo.example.com
+	
 	# fallthrough
 	#
 	# If a query for a record in the cluster zone results in NXDOMAIN,
--- a/middleware/kubernetes/controller.go
+++ b/middleware/kubernetes/controller.go
@@ -69,8 +69,13 @@ type dnsControl struct {
 	stopCh   chan struct{}
 }
 
+type dnsControlOpts struct {
+	initPodCache  bool
+	initNodeCache bool
+}
+
 // newDNSController creates a controller for CoreDNS.
-func newdnsController(kubeClient *kubernetes.Clientset, resyncPeriod time.Duration, lselector *labels.Selector, initPodCache bool, initNodeCache bool) *dnsControl {
+func newdnsController(kubeClient *kubernetes.Clientset, resyncPeriod time.Duration, lselector *labels.Selector, opts dnsControlOpts) *dnsControl {
 	dns := dnsControl{
 		client:   kubeClient,
 		selector: lselector,
@@ -87,7 +92,7 @@ func newdnsController(kubeClient *kubernetes.Clientset, resyncPeriod time.Durati
 		cache.ResourceEventHandlerFuncs{},
 		cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
 
-	if initPodCache {
+	if opts.initPodCache {
 		dns.podLister.Indexer, dns.podController = cache.NewIndexerInformer(
 			&cache.ListWatch{
 				ListFunc:  podListFunc(dns.client, namespace, dns.selector),
@@ -99,7 +104,7 @@ func newdnsController(kubeClient *kubernetes.Clientset, resyncPeriod time.Durati
 			cache.Indexers{podIPIndex: podIPIndexFunc})
 	}
 
-	if initNodeCache {
+	if opts.initNodeCache {
 		dns.nodeLister.Store, dns.nodeController = cache.NewInformer(
 			&cache.ListWatch{
 				ListFunc:  nodeListFunc(dns.client, dns.selector),
@@ -424,6 +429,9 @@ func (dns *dnsControl) EndpointsList() api.EndpointsList {
 }
 
 func (dns *dnsControl) NodeList() api.NodeList {
-	nodeList, _ := dns.nodeLister.List()
+	nodeList, err := dns.nodeLister.List()
+	if err != nil {
+		log.Printf("[ERROR] Could not get NodeList: %v", err.Error())
+	}
 	return nodeList
 }
--- a/middleware/kubernetes/controller.go
+++ b/middleware/kubernetes/controller.go
@@ -206,15 +206,6 @@ func v1ToAPIFilter(in watch.Event) (out watch.Event, keep bool) {
 			return in, true
 		}
 		return watch.Event{Type: in.Type, Object: &apiObj}, true
-	case *v1.Node:
-		var apiObj api.Node
-		err := v1.Convert_v1_Node_To_api_Node(v1Obj, &apiObj, nil)
-		if err != nil {
-			log.Printf("[ERROR] Could not convert v1.Node: %s", err)
-			return in, true
-		}
-		return watch.Event{Type: in.Type, Object: &apiObj}, true
-
 	}
 
 	log.Printf("[WARN] Unhandled v1 type in event: %v", in)
--- a/middleware/kubernetes/controller.go
+++ b/middleware/kubernetes/controller.go
@@ -39,7 +39,9 @@ type dnsController interface {
 	ServiceList() []*api.Service
 	PodIndex(string) []interface{}
 	EndpointsList() api.EndpointsList
-	NodeList() api.NodeList
+
+	GetNodeByName(string) (api.Node, error)
+
 	Run()
 	Stop() error
 }
@@ -55,11 +57,10 @@ type dnsControl struct {
 	epController   *cache.Controller
 	nodeController *cache.Controller
 
-	svcLister  cache.StoreToServiceLister
-	podLister  cache.StoreToPodLister
-	nsLister   storeToNamespaceLister
-	epLister   cache.StoreToEndpointsLister
-	nodeLister cache.StoreToNodeLister
+	svcLister cache.StoreToServiceLister
+	podLister cache.StoreToPodLister
+	nsLister  storeToNamespaceLister
+	epLister  cache.StoreToEndpointsLister
 
 	// stopLock is used to enforce only a single call to Stop is active.
 	// Needed because we allow stopping through an http endpoint and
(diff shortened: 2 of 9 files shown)


middleware/kubernetes/federation_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	testStripFederation(t, k, []string{"service", "ns", "fed", "svc"}, "fed", "service.ns.svc")
+	testStripFederation(t, k, []string{"service", "ns", "foo", "svc"}, "", "service.ns.foo.svc")
+	testStripFederation(t, k, []string{"foo", "bar"}, "", "foo.bar")
+
+}
+
+type apiConnFedTest struct{}
+
+func (apiConnFedTest) Run() {
```
miekg: nit: you put these 1-line function on 1 line, even without newlines between them.

middleware/kubernetes/README.md
```diff
@@ -115,6 +115,12 @@ kubernetes coredns.local {
 	# a path to a file structured like resolv.conf.
 	upstream 12.34.56.78:53
 	
+	# federation <federation-name> <federation-domain>
```
miekg, line 118: This whole readme needs proper docs or more expansion (this is not relevant to this PR, just saying)

middleware/kubernetes/controller.go
```diff
@@ -67,7 +70,7 @@ type dnsControl struct {
 }
 
 // newDNSController creates a controller for CoreDNS.
-func newdnsController(kubeClient *kubernetes.Clientset, resyncPeriod time.Duration, lselector *labels.Selector, initPodCache bool) *dnsControl {
+func newdnsController(kubeClient *kubernetes.Clientset, resyncPeriod time.Duration, lselector *labels.Selector, initPodCache bool, initNodeCache bool) *dnsControl {
```
miekg: These bools needs to be backed in an Opts struct, then this function can use just an Opt parameter where these bools are fields in

middleware/kubernetes/controller.go
```diff
@@ -392,3 +446,8 @@ func (dns *dnsControl) ServiceInNamespace(namespace, servicename string) *api.Se
 	}
 	return svcObj
 }
+
+func (dns *dnsControl) NodeList() api.NodeList {
+	nodeList, _ := dns.nodeLister.List()
```
miekg: why disregard the return error?

middleware/kubernetes/federation.go
```diff
@@ -0,0 +1,76 @@
...
+
+const (
+	// TODO: Do not hardcode these labels. Pull them out of the API instead.
+	//
+	// We can get them via ....
+	//   import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	//     metav1.LabelZoneFailureDomain
+	//     metav1.LabelZoneRegion
+	//
+	// But importing above breaks coredns with flag collision of 'log_dir'
```
miekg, line 27: we don't have `log_dir`, maybe glog does? We already remove `log_dir` from coredns' flags.

middleware/kubernetes/federation.go
```diff
@@ -0,0 +1,76 @@
...
+			fed := segs[len(segs)-2]
+			segs[len(segs)-2] = segs[len(segs)-1]
+			segs = segs[:len(segs)-1]
+			return fed, segs
+		}
+	}
+	return "", segs
+}
+
+// federation CNAMRecord returns a service record for the requested federated service
```
miekg: federationCNAMERercord

middleware/kubernetes/federation.go
```diff
@@ -0,0 +1,76 @@
...
+					Host: strings.Join([]string{r.service, r.namespace, r.federation, r.typeName, node.Labels[LabelAvailabilityZone], node.Labels[LabelRegion], f.zone}, "."),
+				}
+			}
+			return msg.Service{
+				Key:  strings.Join([]string{msg.Path(r.zone, "coredns"), r.typeName, r.federation, r.namespace, r.service, r.endpoint}, "/"),
+				Host: strings.Join([]string{r.endpoint, r.service, r.namespace, r.federation, r.typeName, node.Labels[LabelAvailabilityZone], node.Labels[LabelRegion], f.zone}, "."),
+			}
+
+		}
+		break
```
miekg: what's this `break`? We only range for the first element?

miekg (changes requested):

middleware/kubernetes/federation.go
chrisohaver: We range until we find a node with non-empty region and availability-zone labels.
This mimics the kube-dns kludge.  The break is there for the unlikely case that the federation name in the request doesn't match a federation in the config, we don't bother checking other nodes.

middleware/kubernetes/controller.go
chrisohaver: which params do you think should go in it?
just the two bools? or also the resyncPeriod and label selector?

miekg:
>
>
>
>> @@ -67,7 +70,7 @@ type dnsControl struct {
> }
>
> // newDNSController creates a controller for CoreDNS.
>-func newdnsController(kubeClient *kubernetes.Clientset, resyncPeriod time.Duration, lselector *labels.Selector, initPodCache bool) *dnsControl {
>+func newdnsController(kubeClient *kubernetes.Clientset, resyncPeriod time.Duration, lselector *labels.Selector, initPodCache bool, initNodeCache bool) *dnsControl {
>
>which params do you think should go in it?
>just the two bools? or also the resyncPeriod and label selector?

just start with the bools, we can always do more if we want to, but it starts
a pattern which will be good.

/Miek

--
Miek Gieben

miekg:
>
>
>
>> +			if f.name != r.federation {
>+				continue
>+			}
>+			if r.endpoint == "" {
>+				return msg.Service{
>+					Key:  strings.Join([]string{msg.Path(r.zone, "coredns"), r.typeName, r.federation, r.namespace, r.service}, "/"),
>+					Host: strings.Join([]string{r.service, r.namespace, r.federation, r.typeName, node.Labels[LabelAvailabilityZone], node.Labels[LabelRegion], f.zone}, "."),
>+				}
>+			}
>+			return msg.Service{
>+				Key:  strings.Join([]string{msg.Path(r.zone, "coredns"), r.typeName, r.federation, r.namespace, r.service, r.endpoint}, "/"),
>+				Host: strings.Join([]string{r.endpoint, r.service, r.namespace, r.federation, r.typeName, node.Labels[LabelAvailabilityZone], node.Labels[LabelRegion], f.zone}, "."),
>+			}
>+
>+		}
>+		break
>
>We range until we find a node with non-empty region and availability-zone labels.
>This mimics the kube-dns kludge.  The break is there for the unlikely case that the federation name in the request doesn't match a federation in the config, we don't bother checking other nodes.

ack. Slight comments above the `break` that says this?

middleware/kubernetes/README.md
chrisohaver: Regarding visual formatting: I can move in-line comments into separate sections.
Can you point to an example README format that I should follow.

Regarding expansion: What areas need expansion/improvement? Federation I presume, and others...

middleware/kubernetes/kubernetes.go
```diff
@@ -236,16 +244,18 @@ func (k *Kubernetes) InitKubeCache() (err error) {
 		log.Printf("[INFO] Kubernetes middleware configured with the label selector '%s'. Only kubernetes objects matching this label selector will be exposed.", unversionedapi.FormatLabelSelector(k.LabelSelector))
 	}
 
-	k.APIConn = newdnsController(kubeClient, k.ResyncPeriod, k.Selector, k.PodMode == PodModeVerified)
+	// TODO: pass correct federation bool
+	k.APIConn = newdnsController(kubeClient, k.ResyncPeriod, k.Selector, k.PodMode == PodModeVerified, true)
```
chrisohaver: oops...

middleware/kubernetes/federation.go
chrisohaver: It may be unvendored glog conflicting with vendored glog.
We may need to vendor k8s.io/apimachinery

middleware/kubernetes/README.md
chrisohaver: Let's track this in a separate issue...

miekg:

I see:
~~~
./coredns flag redefined: log_dir
panic: ./coredns flag redefined: log_dir
~~~
that's pretty shit :(

/Miek

--
Miek Gieben

chrisohaver:
Regarding the latest commit to the PR.  I decided to implement the node lookup the "right way" instead of mimicking the kube-dns kludge.  We now pull the availability Zone and Region from the Node that coredns actually resides in (not just the first node in the cluster). It's also more efficient this way, doesn't require a watch on node objects.

middleware/kubernetes/controller.go
```diff
@@ -198,6 +206,15 @@ func v1ToAPIFilter(in watch.Event) (out watch.Event, keep bool) {
...
 		return watch.Event{Type: in.Type, Object: &apiObj}, true
+	case *v1.Node:
+		var apiObj api.Node
+		err := v1.Convert_v1_Node_To_api_Node(v1Obj, &apiObj, nil)
+		if err != nil {
+			log.Printf("[ERROR] Could not convert v1.Node: %s", err)
+			return in, true
+		}
+		return watch.Event{Type: in.Type, Object: &apiObj}, true
+
```
johnbelamaric: Since you went to the one-time lookup, we don't need this anymore.

johnbelamaric (changes requested):
I think you can pull that snippet out. Otherwise, LGTM.

johnbelamaric (approved):

middleware/kubernetes/README.md
chrisohaver: I've created #728 to track this issue.

middleware/kubernetes/federation.go
miekg: Checked this, indeed annoying :(

miekg (approved):

PR #735, opened 2017-06-13
middleware/debug: add

miekg:
Add a debug "middleware" that disables the recover() and just lets
CoreDNS crash; very useful for testing.

Fixes ##563


miekg:
@pmoroney you feel up to a quick review of this?

pmoroney:
It looks good to me, assuming the Plugin.Action() function is never called concurrently.
Should this also be a command line flag like dns_root?

pmoroney:
Since I am new at this, what is the procedure for reviewing?
Do I merge it if it is fairly simple, looks good, and the tests pass, or should I pull and test out the change manually?
Also which merge option should I use?

pmoroney:
Ahh, just noticed the "add your review" button on the top of the page, I assume that is what I need to be doing. reading through the github docs now.

miekg:
You can approve the review (somewhere) :) And then merge - or just merge

AFAICT we don't have a dns.root flag?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/de..." ]
>Since I am new at this, what is the procedure for reviewing?
>Do I merge it if it is fairly simple, looks good, and the tests pass, or should I pull and test out the change manually?
>Also which merge option should I use?

squash and merge

pmoroney (approved):
Looks good to me.

I verified that Plugin.Action() does not get called concurrently in Caddy, so operating on a pointer to the config should be safe.

Although, it'd also be nice if the test checked to see if it actually modified the Config.Debug value, but as this isn't meant for production use it is probably fine.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/de..." ]
>pmoroney approved this pull request.
>Although, it'd also be nice if the test checked to see if it actually modified the Config.Debug value, but as this isn't meant for production use it is probably fine.

Good point. Actually deleted some of that code - let me put it back.

pmoroney:
Ahh, sorry I was thinking of -dns.port

miekg:
Pushed 1 commit. PTAL - if it looks good, just merge. Thanks!

pmoroney (approved):
LGTM

pmoroney:
For future reference, Should I delete the branches after merging PRs?

miekg:
yes - I just did - otherwise these just accumulate (with no use)

pmoroney:
Awesome, thanks for teaching me how this all works!

PR #745, opened 2017-06-19
middleware/secondary: Fix issues

miekg:


miekg:
@johnbelamaric I'm in a bit of a bind here... If I use an external running process that provides `example.com` to the instance being tested I get *no* race detected.
If I uncomment these lines: https://github.com/coredns/coredns/pull/745/files#diff-aa0057fdcfcefafa23f0f1d86ecb8158R58 (and fix the code a bit to make the test run) I *do* get a race detected.
My gut says this is thus not a race in coredns, but maybe in the test... While writing this I think the race might well be in the re-use of the message `m` in both serving instances...

miekg:
Ok, worked around it by upping a separate coredns that as the zone and allows transfer. Pondering if and how to do this when doing automatic testing.

SuperQ:
This `upstream` thing makes no sense to me.

SuperQ:
There should be no way a request for a zone triggers a server to do a recursive resolve in the way this upstream thing seems to be intended.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/se..." ]
>There should be no way a request for a zone triggers a server to do a recursive resolve in the way this upstream thing seems to be intended.

This is needed in the case coredns is used as (recursive) proxy, clients then 
expects a complete response from the server, as corends is seen as fully recursive.
This is also the reason why it is in the file middleware. Docs could be 
improved here.

SuperQ:
When responding to a specific zone no recursive request should be necessary.  This seems like it's going to lead to very bad response behavior.

SuperQ:
Also, shouldn't any recursive response be handled by a standard root server request path by default rather than having to pass it up to a "smart proxy"?

johnbelamaric:
The test with an out-of-band server can be setup in the CI with Travis if you don’t want it in unit tests. There we know the environment so we can assume a given port - harder to do in unit tests which could be run on any machine.

On Jun 20, 2017, at 3:06 PM, Miek Gieben <notifications@github.com<mailto:notifications@github.com>> wrote:


@johnbelamaric<https://github.com/johnbelamaric> I'm in a bit of a bind here... If I use an external running process that provides example.com<http://example.com> to the instance being tested I get no race detected.
If I uncomment these lines: https://github.com/coredns/coredns/pull/745/files#diff-aa0057fdcfcefafa23f0f1d86ecb8158R58 (and fix the code a bit to make the test run) I do get a race detected.
My gut says this is thus not a race in coredns, but maybe in the test... While writing this I think the race might well be in the re-use of the message m in both serving instances...

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/se..." ]
>Also, shouldn't any recursive response be handled by a standard root server request path by default rather than having to pass it up to a "smart proxy"?

Why and how? There is no recursive code in CoreDNS...

SuperQ:
Well, even if there is no way to recurse from the root hints, having recursion in secondary middleware still makes no sense.  You would also need to have equivalent options in the `file` middleware as well.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/se..." ]
>Well, even if there is no way to recurse from the root hints, having recursion in secondary middleware still makes no sense.  You would also need to have equivalent options in the `file` middleware as well.

If we had a recursion middleware we could use that, yes. But we don't have it 
(and making a recursor is a 5+ year time commitment, I can't make), so in the 
mean time, a hack like this is the only option :-(

Use sparingly and only when CoreDNS is used as a proxy - we could go as far to 
disable this when the proxy middleware isn't loaded (maybe).

miekg:
I'm going to do the integration test in a later PR - the fixed ports I need are *really* annoying :(

PR #746, opened 2017-06-20
middleware/kubernetes: Add error logging to k8s list funcs

chrisohaver:
Log errors if we receive an error when referencing k8s object cache.

### DIFF
--- a/middleware/kubernetes/controller.go
+++ b/middleware/kubernetes/controller.go
@@ -344,6 +344,7 @@ func (dns *dnsControl) Run() {
 func (dns *dnsControl) NamespaceList() *api.NamespaceList {
 	nsList, err := dns.nsLister.List()
 	if err != nil {
+		log.Printf("[ERROR] Could not get namespace list: %s", err)
 		return &api.NamespaceList{}
 	}
 
@@ -353,6 +354,7 @@ func (dns *dnsControl) NamespaceList() *api.NamespaceList {
 func (dns *dnsControl) ServiceList() []*api.Service {
 	svcs, err := dns.svcLister.List(labels.Everything())
 	if err != nil {
+		log.Printf("[ERROR] Could not get service list: %s", err)
 		return []*api.Service{}
 	}
 
@@ -362,6 +364,7 @@ func (dns *dnsControl) ServiceList() []*api.Service {
 func (dns *dnsControl) PodIndex(ip string) []interface{} {
 	pods, err := dns.podLister.Indexer.ByIndex(podIPIndex, ip)
 	if err != nil {
+		log.Printf("[ERROR] Could not get pod index: %s", err)
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


chrisohaver:
Regarding the coverage ... the k8s controller is not yet covered by unit tests.  Hence the 0%.  Adding coverage is not trivial.

middleware/kubernetes/controller.go
```diff
@@ -371,6 +374,7 @@ func (dns *dnsControl) PodIndex(ip string) []interface{} {
 func (dns *dnsControl) EndpointsList() api.EndpointsList {
 	epl, err := dns.epLister.List()
 	if err != nil {
+		log.Printf("[ERROR] Could not get endpoint list: %s", err)
```
miekg, line 377: Can we make these methods just return the `error` and potentially print from the caller?

chrisohaver: We could.  It will require changing the interface, and all of its implementors.

miekg:
>
>
>
>> @@ -371,6 +374,7 @@ func (dns *dnsControl) PodIndex(ip string) []interface{} {
> func (dns *dnsControl) EndpointsList() api.EndpointsList {
> 	epl, err := dns.epLister.List()
> 	if err != nil {
>+		log.Printf("[ERROR] Could not get endpoint list: %s", err)
>
>We could.  It will require changing the interface, and all of its implementors.

Ours or *all of kubernetes*? The later is obviously not going to work, if it is 
the first: why not?

/Miek

--
Miek Gieben

chrisohaver:
> Ours or *all of kubernetes*? The later is obviously not going to work, if it is the first: why not?

The error is from kubernetes' methods that retrieve the cached data.  The data isn't coming directly from kubernetes at this stage, its being pulled from local cached copies of the objects.  I have never seen an error at this stage, but if we were to get errors here it might be because the cache stores became corrupted in memory somehow or a kubernetes bug of some kind.  If we get an error in these cases, we currently would return an empty list, which would result in an NXDOMAIN downstream.

chrisohaver:
Perhaps we should be returning SERVFAIL in these cases instead of NXDOMAIN...

miekg:
>Perhaps we should be returning SERVFAIL in these cases instead of NXDOMAIN...

Unless the error says 'no such name/service, making it an SERVFAIL is a good 
idea.

miekg:
>The error is from kubernetes' cache.

So... we can't change the signature of the function? If so, then lgtm

/Miek

--
Miek Gieben

chrisohaver:
> So... we can't change the signature of the function? If so, then lgtm

I can change the signatures.  It just impacts several unit test mocks of the controller interface, so i was trying to avoid it unless necessary.  But, I think I'll need to return the error to differentiate SERVFAIL vs NXDOMAIN anyways.  So I have to bite the bullet.

miekg:
Are you planning to do this for 009? Or should put it in 010?

chrisohaver:
We can push this to 010.  Though I may abandon the change.  Changing the function signatures was turning out to require more refactoring than I had anticipated.

miekg:
Think we close this?

PR #748, opened 2017-06-21
middleware/etcd: refactor etcd to backend

VioletRainbows:
#### Done
- Separated etcdv2 `ServiceBackend` implementation from the `Handler` implementation.
- Compiles and run
- Readme (was simply moved)
- Test (but they require doing: `e.ServiceBackend.(*etcdv2.EtcdV2).`). Also they are disabled.

#### Info
There are two struct: `Backend`, and `EtcdV2`.
`EtcdV2` contains only what is required by the `ServiceBackend` implementation.
Everything else is in `Backend`.

Please tell me what you guys are thinking about this approach.


middleware.cfg
```diff
@@ -40,7 +40,7 @@
 180:file:file
 190:auto:auto
 200:secondary:secondary
-210:etcd:etcd
+210:etcd:github.com/coredns/coredns/middleware/backend/etcdv2
```
miekg: ugh, because it is one dir deeper, we need the full path? (need to fix that in the generator)

middleware/backend/tests/debug_test.go
```diff
@@ -1,12 +1,12 @@
 // +build etcd
 
-package etcd
+package btests
```
miekg: why can't the test be in the same package scope?

middleware/backend/setup.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/coredns/coredns/middleware"
+	"github.com/coredns/coredns/middleware/backend/msg"
+	"github.com/coredns/coredns/middleware/proxy"
+)
+
+// StubServiceBackend is a ServiceBackend that also support stubs.
+type StubServiceBackend interface {
+	middleware.ServiceBackend
+
+	Records(name string, exact bool) ([]msg.Service, error)
```
miekg: Why the `Services` method in `middleware.ServiceBackend` not sufficient, why this extra method?

miekg: Ah, sorry, its for the `Stub`, makes sense. Should rename `Records -> Stubs` ?

miekg (commented):
The Stub interface abstraction make sense, but why did you move *backend* into `middleware`? Granted where it was, might not be the best place, but wound't the diff be way smaller if we left it there? (Ans easier to review).

Is the stub interface enough to get etcd3/etcd2 and consul fit the same interface?

middleware/backend/tests/debug_test.go
VioletRainbows: I had an issue with recursive imports (etcd importing backend, backend tests importing etcd).  I fixed it by moving it to another package. If you know a better way of doing it/naming, feel free to comment.

VioletRainbows:
When implementing etcdv3 and consul, the only function that needs to be implemented is `Records()` (and maybe `IsNameError()`) with the according `setup()`. Everything else stays the same.

To not copy/paste tests, I thought of keeping them in the same folder, but recursive imports blocked me on that.

I am still trying to figure out where would the (new?) backend implementation be best placed while keeping the tests in mind, but I am still unsure.

For the stub interface, the function used really returns all records. It is not stub-specific. Maybe the interface name is to be changed.

middleware.cfg
VioletRainbows: If need be, I'll modify the generator to support nested folders. In the meantime, I really appreciate your feedback and will keep updating the PR.

miekg:
>When implementing etcdv3 and consul, the only function that needs to be implemented is `Records()` (and maybe `IsNameError()`) with the according `setup()`. Everything else stays the same.

that makes sense.

>To not copy/paste tests, I thought of keeping them in the same folder, but recursive imports blocked me on that.

maybe shelve that problem for the time being? I.e. no tests for the new backends 
(yet)?

>I am still trying to figure out where would the (new?) backend implementation be best placed while keeping the tests in mind, but I am still unsure.

`middleware/backend/{etcd,etcd3,consul}` sounds good to me.

>For the stub interface, the function used really returns all records. It is not stub-specific. Maybe the interface name is to be changed.

Ok, please do. We only use this internally, so you're free to do whatever you/we 
want.

/Miek

--
Miek Gieben

miekg:
>VioletRainbows commented on this pull request.
>
>> @@ -1,12 +1,12 @@
> // +build etcd
>
>-package etcd
>+package btests
>
>I had an issue with recursive imports (etcd importing backend, backend tests importing etcd).  I fixed it by moving it to another package. If you know a better way of doing it/naming, feel free to comment.

Ack I will try to send you a PR that fixes this.

VioletRainbows:
* Renamed StubServiceBackend to ServiceBackendRecords 	
* Add relative path support in directives_generate.go
* Disabled tests

Tests are now in `middleware/backend` (but you need to add `-tags backend` to run them). They will fail due to import cycles.

Everything I have tried so far leads to either moving all tests to a different package or having to deal with import cycles.

miekg:
I tried getting rid of the import cycles - which is possible, but requires moving stuff around quite a bit. So I will not send a PR out for that.

But this comes back to my earlier question: why are you moving all this code around? If we just want to extend the interfaces a bit (work this is *very* welcome), we should just do that. Then fix the current *etcd* middleware; add a *etcd3* and *consul* variant? WDYT?

Should I extract `Add relative path support in directives_generate.go` from this PR and create a new one? So we can do that separately ? 

Reviewing also becomes harder with these code movements.

VioletRainbows:
I've created a few PRs to remove some burden from this one. Nothing major.

I'm not sure I understand what you mean by:
> If we just want to extend the interfaces a bit (work this is very welcome), we should just do that.

Should I modify the `ServiceBackend interface` with `Records` even though k8s does not need it? Or do you want me to create a separate PR where I only create and use the new interface, without moving or renaming code?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/et..." ]
>I've created a few PRs to remove some burden from this one. Nothing major.

Saw those, all merged. Thanks.

>I'm not sure I understand what you mean by:
>> If we just want to extend the interfaces a bit (work this is very welcome), we should just do that.
>
>Should I modify the `ServiceBackend interface` with `Records` even though k8s does not need it? Or do you want me to create a separate PR where I only create and use the new interface, without moving or renaming code?

I think just adding `Records` is the correct thing to do, and noop it in k8s.

/Miek

--
Miek Gieben

miekg:
I'm going close this as we are working on a different branch (almost accidentally merged it)

PR #749, opened 2017-06-23
middleware/proxy: async health checks

ghostflame:
Update to middleware/proxy:

The approach to this was discussed in the issue, but tl;dr -

Health checks are now done asynchronously in a go routine.  Rather than a healthy/unhealthy bool state, servers are marked healthy until a timestamp.  This is typically 2 * healthcheck-interval.  If they fail a health check, they are marked healthy until Unix(0, 0).

The mutex that was around the health check is now only around updating the health status (necessary to prevent a race).  The upshot of all this is that:

If a host is gone (and so health checking takes a very long time to discover it is dead) it's healthy-until will have expired already, and we do not block anything, critically the mutex, while we find this out.  A check in progress bool prevents concurrent checks overlapping.

The failure count mechanism is untouched.

### DIFF
--- a/middleware/proxy/google.go
+++ b/middleware/proxy/google.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"math"
 	"net"
 	"net/http"
 	"net/url"
@@ -236,6 +237,8 @@ func newUpstream(hosts []string, old *staticUpstream) Upstream {
 			}(upstream),
 			WithoutPathPrefix: upstream.WithoutPathPrefix,
 		}
+
+		log.Printf("Host %s marked healthy until %s.\n", uh.Name, uh.OkUntil.Local())
 		upstream.Hosts[i] = uh
 	}
 	return upstream
--- a/middleware/proxy/google.go
+++ b/middleware/proxy/google.go
@@ -218,7 +218,7 @@ func newUpstream(hosts []string, old *staticUpstream) Upstream {
 			Conns:       0,
 			Fails:       0,
 			FailTimeout: upstream.FailTimeout,
-			OkUntil:     time.Now().Add(upstream.HealthCheck.Future),
+			OkUntil:     time.Unix(math.MaxInt64, 0), // forever, initially
 
 			CheckDown: func(upstream *staticUpstream) UpstreamHostDownFunc {
 				return func(uh *UpstreamHost) bool {
--- a/middleware/proxy/google.go
+++ b/middleware/proxy/google.go
@@ -218,11 +218,12 @@ func newUpstream(hosts []string, old *staticUpstream) Upstream {
 			Conns:       0,
 			Fails:       0,
 			FailTimeout: upstream.FailTimeout,
-			Unhealthy:   false,
+			OkUntil:     time.Now().Add(upstream.HealthCheck.Future),
 
 			CheckDown: func(upstream *staticUpstream) UpstreamHostDownFunc {
 				return func(uh *UpstreamHost) bool {
-					if uh.Unhealthy {
+
+					if time.Now().After(uh.OkUntil) {
 						return true
 					}
 
--- a/middleware/proxy/google.go
+++ b/middleware/proxy/google.go
@@ -242,7 +242,6 @@ func newUpstream(hosts []string, old *staticUpstream) Upstream {
 			WithoutPathPrefix: upstream.WithoutPathPrefix,
 		}
 
-		log.Printf("[DEBUG] Ggl: Host %s marked healthy until %s.\n", uh.Name, uh.OkUntil.Local())
 		upstream.Hosts[i] = uh
 	}
 	return upstream
--- a/middleware/proxy/google.go
+++ b/middleware/proxy/google.go
@@ -218,26 +218,30 @@ func newUpstream(hosts []string, old *staticUpstream) Upstream {
 			Conns:       0,
 			Fails:       0,
 			FailTimeout: upstream.FailTimeout,
-			OkUntil:     time.Unix(4000000000, 0), // forever, initially
+			OkUntil:     time.Now().Add(time.Duration(24 * 365 * 100 * time.Hour)), // now + 100 years
 
 			CheckDown: func(upstream *staticUpstream) UpstreamHostDownFunc {
 				return func(uh *UpstreamHost) bool {
 
+					down := false
+
+					uh.checkMu.Lock()
 					if time.Now().After(uh.OkUntil) {
-						return true
+						down = true
 					}
+					uh.checkMu.Unlock()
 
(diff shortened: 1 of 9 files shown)


middleware/proxy/google.go
```diff
@@ -218,23 +219,30 @@ func newUpstream(hosts []string, old *staticUpstream) Upstream {
 			Conns:       0,
 			Fails:       0,
 			FailTimeout: upstream.FailTimeout,
-			Unhealthy:   false,
+			OkUntil:     time.Now().Add(time.Duration(24 * 365 * 100 * time.Hour)), // now + 100 years
```
miekg: probably better as a constant. HundredYears or something better, but pondering if the zero value of time.Time{} wouldn't be better.

middleware/proxy/google.go
```diff
@@ -218,23 +219,30 @@ func newUpstream(hosts []string, old *staticUpstream) Upstream {
...
+						down = true
 					}
-					return false
+					return down
 				}
 			}(upstream),
 			WithoutPathPrefix: upstream.WithoutPathPrefix,
 		}
+
+		log.Printf("[DEBUG] Ggl: Host %s marked healthy until %s.\n", uh.Name, uh.OkUntil.Local())
```
miekg: could be removed?

middleware/proxy/upstream.go
```diff
@@ -247,46 +258,93 @@ func parseBlock(c *caddyfile.Dispenser, u *staticUpstream) error {
...
-			host.Unhealthy = true
+
+			host.CheckUrl = "http://" + net.JoinHostPort(checkHostName, checkPort) + u.HealthCheck.Path
 		}
+
+		// calculate this before the get
+		nextTs := time.Now().Add(u.Future)
+
+		// locks/bools should prevent requests backing up
+		go healthCheckUrl(nextTs, host)
```
miekg, line 348: There should be a way to stop this goroutine on shutdown, so probably needs a stop channel or something?

miekg (commented):
General gist look OK. I want to play with this PR some more and probably have some additional feedback after that.
Thanks for the PR!

miekg:
Is it possible to think of a test that breaks with the current code and is fixed with this PR?

middleware/proxy/google.go
ghostflame: This is for the no-health-check case - in which case OkUntil has to always be after time.Now().  If healthchecks are configured, the first one will override this.  I needed to pick a value guaranteed to be later than Now() for all reasonable circumstances.  If there's a suitable constant, sure.

middleware/proxy/google.go
ghostflame: Yes.  Will do.

middleware/proxy/upstream.go
ghostflame: If it runs fractionally after the stop signal, does that matter?  It's only for a single in-flight health check - the go routine does just one check and then exits.  Next time we throw another.
I don't know the requirements of shutdown well enough, but a little googling suggested that go routines are not easily killed in the midst of blocking operations.  If this is important I can keep looking into it.

ghostflame:
The test cases suggested by @raspberrypython involved an IP address that doesn't respond, blocking for minutes.  The test framework can't - unless I am unaware of how - easily replicate this.  Any test on a single host misses out the network timeout problems it is meant to fix because the local kernel knows the state of local networks.

I guess one approach might be a test webserver that didn't time out, but dribbled a response back very slowly, a few bytes/second, enough to keep the health check interested but making the response stretch over minutes.  The old code would block lookups, with the mutex locked all that time, and this code would declare that server dead in a few seconds, depending on the interval settings, with lookups working until that moment.

That would, however, make the test suite take several minutes longer.  :-(

middleware/proxy/google.go
miekg: but you can special case `time.Time{}` in the code does that check as well, no? Yes, that makes that check somewhat more ugly, because you need to check for the special value. OTOH this 100 years is also a legal time vaue. I slightly lean into make an `IsZero` time the special thing here.

middleware/proxy/upstream.go
miekg: Ack. I'm somewhat concerned if we do a live reload do these live on? (not sure; I think they get killed). Also that would imply only leaking when reloading (if that is even true), which is not that important.

miekg:
yes, true, we discussed that on the issue - very hard to replicate in the (localhost) test suite.

middleware/proxy/google.go
ghostflame: OK.  I already use time.Unix(0,0) to signal a failed health check (because Now() is always after it).  So we'd make the check:
if ! uh.OkUntil.IsZero() && time.Now().After(uk.OkUntil) {
    // fail
}
?
That is different from time.Unix(0,0) so the mechanism would still work.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/pr..." ]
>ghostflame commented on this pull request.
>
>
>
>> @@ -218,23 +219,30 @@ func newUpstream(hosts []string, old *staticUpstream) Upstream {
> 			Conns:       0,
> 			Fails:       0,
> 			FailTimeout: upstream.FailTimeout,
>-			Unhealthy:   false,
>+			OkUntil:     time.Now().Add(time.Duration(24 * 365 * 100 * time.Hour)), // now + 100 years
>
>OK.  I already use time.Unix(0,0) to signal a failed health check (because Now() is always after it).  So we'd make the check:
>if ! uh.OkUntil.IsZero() && time.Now().After(uk.OkUntil) {
>    // fail
>}
>?
>That is different from time.Unix(0,0) so the mechanism would still work.

As said, I *think* this is slightly better

middleware/proxy/upstream.go
ghostflame: There's no loop inside the go routine healthCheckUrl - it just does a single Get() and exits.  It's the ticker loop inside healthCheckWorker() that does the loop and that is unchanged.  So other than a check to each host being in-flight when the stop is issued (and those single checks running to completion) it should show the same behaviour as before.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/pr..." ]
>ghostflame commented on this pull request.
>
>
>
>>  		}
>+
>+		// calculate this before the get
>+		nextTs := time.Now().Add(u.Future)
>+
>+		// locks/bools should prevent requests backing up
>+		go healthCheckUrl(nextTs, host)
>
>There's no loop inside the go routine healthCheckUrl - it just does a single Get() and exits.  It's the ticker loop inside healthCheckWorker() that does the loop and that is unchanged.  So other than a check to each host being in-flight when the stop is issued (and those single checks running to completion) it should show the same behaviour as before.

Ack. good.

Issue #747, opened 2017-06-21
middleware/kubernetes: Preemptive search path lookups


chrisohaver:
**Problem**:

A known issue with kube-dns and how K8s works is the long DNS search path and high ”ndots” in /etc/resolv.conf. https://github.com/kubernetes/kubernetes/issues/33554

For example, search path in a pod as dictated by the pod's dns policy is:
```
namespace.svc.cluster.local
svc.cluster.local
cluster.local
mydomain.com
```

If you try to lookup ”google.com” you will end up with queries:

```
google.com.namespace.svc.cluster.local.
google.com.svc.cluster.local.
google.com.cluster.local.
google.com.mydomain.com.
google.com.
```

This means that for every query of a host outside the cluster, the client has to make 5 queries.

**Proposed Solution**:

For queries from pods in the cluster, iterate on the search path in coredns, returning the first successful lookup back to the client.  This allows the client to get the answer in one query, instead of five. 

The condition for doing the preemptive lookups on an incoming question would have to meet all of the following...

- the incoming question would otherwise result in an NXDOMAIN (e.g. it doesn't match a service)
- the source is a pod in the kubernetes cluster
- the question ends with `namespace.svc.cluster.local.` where `namespace` is the namespace of the pod, and `cluster.local` is the zone

If these conditions are met, then the kubernetes middleware would strip the search path from the question, and try the remaining search paths until an answer is found.

miekg:
Would there be a way to pull the proposed hackery into its own middleware?

johnbelamaric:
It needs the pod watch that we put in for `pods verified` so it's probably not ideal to do that. It definitely though can be optionally enabled.

johnbelamaric:
Or are you saying so that it can be used in non-k8s cases?

miekg:
Reuse maybe. More from code complicity standpoint.

miekg:
Should we call it *autopath* (like the name, but it is a bit non descriptive) or maybe *serverpath*?  *serversearch*, *autosearch*?

Issue #728, opened 2017-06-10
middleware/kubernetes: Improve/Expand README


chrisohaver:
Improve presentation of the kubernetes middleware readme, and expand content.

chrisohaver:
Please assign this issue to me.

miekg:
done

PR #750, opened 2017-06-23
middleware/kubernetes: Server side path lookups

chrisohaver:
Implements #747 as proposed.

I've added docs inline with the current format. Will restructure and expand the docs later (#728)
I have not added a test in Travis CI k8s integration, since it currently doesn't run tests from within a pod.

### DIFF
--- a/middleware/kubernetes/README.md
+++ b/middleware/kubernetes/README.md
@@ -121,7 +121,7 @@ kubernetes coredns.local {
 	# Each line consists of the name of the federation, and the domain.
 	federation myfed foo.example.com
 	
-	# autopath [ndots:NDOTS] [resolv:RESOLV-CONF-FILE] [onnxdomain:RESPONSE] ...
+	# autopath [NDOTS [RESPONSE [RESOLV-CONF]]
 	#
 	# Enables server side search path lookups for pods.  When enabled, coredns
 	# will identify search path queries from pods and perform the remaining
@@ -149,24 +149,23 @@ kubernetes coredns.local {
 	#    ;; ANSWER SECTION:
 	#    google.com.		175	IN	A	216.58.194.206
 	#
-	# All 3 arguments are optional:
-	# 
-	# resolv:RESOLV-CONF-FILE (default: /etc/resolv.conf) If specified, coredns
-	# uses this file to get the host's search domains. CoreDNS performs a lookup
-	# on these domains if the in-cluster search domains in the path fail to 
-	# produce an answer. If not specified, the values will be read from the local 
-	# resolv.conf file (i.e the resolv.conf file in the pod containing coredns).
 	#
-	# ndots:NDOTS (default: 1) This provides an adjustable threshold to
+	# NDOTS (default: 0) This provides an adjustable threshold to
 	# prevent server side lookups from triggering. If the number of dots before
 	# the first search domain is less than this number, then the search path will
 	# not executed on the server side.
 	#
-	# onnxdomain:RESPONSE (default: NXDOMAIN) RESPONSE can be either SERVFAIL or
+	# RESPONSE (default: SERVFAIL) RESPONSE can be either NXDOMAIN, SERVFAIL or
 	# NOERROR. This option causes coredns to return the given response instead of
-	# NXDOMAIN when the all searches in the path produce no results. Doing this
-	# should prevent the client from fruitlessly continuing the client side 
-	# searches in the path.
+	# NXDOMAIN when the all searches in the path produce no results. Setting this
+	# to SERVFAIL or NOERROR should prevent the client from fruitlessly continuing
+	# the client side searches in the path after the server already checked them.
+	#
+	# RESOLV-CONF (default: /etc/resolv.conf) If specified, coredns uses this
+	# file to get the host's search domains. CoreDNS performs a lookup on these
+	# domains if the in-cluster search domains in the path fail to produce an
+	# answer. If not specified, the values will be read from the local resolv.conf
+	# file (i.e the resolv.conf file in the pod containing coredns).
 	#
 	# Enabling autopath causes coredns to use more memory since it needs to
 	# maintain a watch on all pods. If autopath and "pods verified" mode are
--- a/middleware/kubernetes/README.md
+++ b/middleware/kubernetes/README.md
@@ -121,6 +121,41 @@ kubernetes coredns.local {
 	# Each line consists of the name of the federation, and the domain.
 	federation myfed foo.example.com
 	
+	# autopath [<host-domain>]
+	#
+	# Enables server side search path lookups for pods.  When enabled, coredns
+	# will identify search path queries from pods and perform the remaining path
+	# lookups on the pod's behalf.  The search path used mimics the resolve.conf
+	# search path deployed to pods. E.g.
+	#
+	#     search namespace.svc.cluster.local svc.cluster.local cluster.local
+	#
+	# <host-domain> is optional. If specified, coredns performs a lookup on this
+	# path if the preceeding paths fail to produce an answer.
+	#
+	# If no paths produce an answer, coredns attempts the empty (.) path.
+	#
+	# A successful response will contain a question section with the original
+	# question, and an answer section containing the a record for a diffrent
+	# question.  The question and answer will not match. For example:
+	#
+	#    # host -v -t a google.com
+	#    Trying "google.com.default.svc.cluster.local"
+	#    ;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 50957
+	#    ;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0
+	#
+	#    ;; QUESTION SECTION:
+	#    ;google.com.default.svc.cluster.local. IN A
+	#
+	#    ;; ANSWER SECTION:
+	#    google.com.		175	IN	A	216.58.194.206
+	#
+	# Enabling autopath causes coredns to use more memory since it needs to
+	# maintain a watch on all pods. If autopath and "pods verified" mode are
+	# both enabled, they will use share the same watch. I.e. enabling both options
+	# should have an equivalent memory impact of just one.
+	autopath foo.example.com.
+
 	# fallthrough
 	#
 	# If a query for a record in the cluster zone results in NXDOMAIN,
--- a/middleware/kubernetes/README.md
+++ b/middleware/kubernetes/README.md
@@ -121,19 +121,22 @@ kubernetes coredns.local {
 	# Each line consists of the name of the federation, and the domain.
 	federation myfed foo.example.com
 	
-	# autopath [<host-domain>]
+	# autopath [ndots:NDOTS] [HOST-DOMAIN] [HOST-DOMAIN] ...
 	#
 	# Enables server side search path lookups for pods.  When enabled, coredns
-	# will identify search path queries from pods and perform the remaining path
-	# lookups on the pod's behalf.  The search path used mimics the resolve.conf
-	# search path deployed to pods. E.g.
+	# will identify search path queries from pods and perform the remaining
+	# lookups in the path on the pod's behalf.  The search path used mimics the
+	# resolv.conf search path deployed to pods. E.g.
 	#
 	#     search namespace.svc.cluster.local svc.cluster.local cluster.local
 	#
-	# <host-domain> is optional. If specified, coredns performs a lookup on this
-	# path if the preceeding paths fail to produce an answer.
+	# HOST-DOMAIN are optional. If specified, coredns performs a lookup on these
+	# domains if the preceeding searches fail to produce an answer. If not
+	# specified, the values will be read from the local resolv.conf file (i.e
+	# the resolv.conf file in the pod containing coredns).
 	#
-	# If no paths produce an answer, coredns attempts the empty (.) path.
+	# If no domains in the path produce an answer, a lookup on the bare question
+	# attempted.	
 	#
 	# A successful response will contain a question section with the original
 	# question, and an answer section containing the record for the question that
@@ -151,11 +154,16 @@ kubernetes coredns.local {
 	#    ;; ANSWER SECTION:
 	#    google.com.		175	IN	A	216.58.194.206
 	#
+	# NDOTS is optional, defaulting to 1. This provides an adjustable threshold to
+	# prevent server side lookups from triggering. If the number of dots before
+	# the first search domain is less than this number, then the search path will
+	# not executed on the server side.
+	#
 	# Enabling autopath causes coredns to use more memory since it needs to
 	# maintain a watch on all pods. If autopath and "pods verified" mode are
 	# both enabled, they will use share the same watch. I.e. enabling both options
 	# should have an equivalent memory impact of just one.
-	autopath foo.example.com.
+	autopath ndots:1 foo.example.com.
 
 	# fallthrough
 	#
--- a/middleware/kubernetes/README.md
+++ b/middleware/kubernetes/README.md
@@ -131,7 +131,7 @@ kubernetes coredns.local {
 	#  search ns1.svc.cluster.local svc.cluster.local cluster.local foo.com
 	#
 	# If no domains in the path produce an answer, a lookup on the bare question
-	# attempted.	
+	# will be attempted.	
 	#
 	# A successful response will contain a question section with the original
 	# question, and an answer section containing the record for the question that
@@ -171,7 +171,7 @@ kubernetes coredns.local {
 	# maintain a watch on all pods. If autopath and "pods verified" mode are
 	# both enabled, they will share the same watch. I.e. enabling both options
 	# should have an equivalent memory impact of just one.
-	autopath ndots:1 foo.example.com.
+	autopath 0 SERVFAIL /etc/resolv.conf
 
 	# fallthrough
 	#
--- a/middleware/kubernetes/README.md
+++ b/middleware/kubernetes/README.md
@@ -136,8 +136,9 @@ kubernetes coredns.local {
 	# If no paths produce an answer, coredns attempts the empty (.) path.
 	#
 	# A successful response will contain a question section with the original
-	# question, and an answer section containing the a record for a diffrent
-	# question.  The question and answer will not match. For example:
+	# question, and an answer section containing the record for the question that
+	# actually had an answer.  This means that the question and answer will not
+	# match. For example:
 	#
 	#    # host -v -t a google.com
 	#    Trying "google.com.default.svc.cluster.local"
--- a/middleware/kubernetes/README.md
+++ b/middleware/kubernetes/README.md
@@ -121,19 +121,14 @@ kubernetes coredns.local {
 	# Each line consists of the name of the federation, and the domain.
 	federation myfed foo.example.com
 	
-	# autopath [ndots:NDOTS] [HOST-DOMAIN] [HOST-DOMAIN] ...
+	# autopath [ndots:NDOTS] [resolv:RESOLV-CONF-FILE] [onnxdomain:RESPONSE] ...
 	#
 	# Enables server side search path lookups for pods.  When enabled, coredns
 	# will identify search path queries from pods and perform the remaining
 	# lookups in the path on the pod's behalf.  The search path used mimics the
 	# resolv.conf search path deployed to pods. E.g.
 	#
-	#     search namespace.svc.cluster.local svc.cluster.local cluster.local
-	#
-	# HOST-DOMAIN are optional. If specified, coredns performs a lookup on these
-	# domains if the preceeding searches fail to produce an answer. If not
-	# specified, the values will be read from the local resolv.conf file (i.e
-	# the resolv.conf file in the pod containing coredns).
+	#  search ns1.svc.cluster.local svc.cluster.local cluster.local foo.com
 	#
 	# If no domains in the path produce an answer, a lookup on the bare question
 	# attempted.	
@@ -154,14 +149,28 @@ kubernetes coredns.local {
 	#    ;; ANSWER SECTION:
 	#    google.com.		175	IN	A	216.58.194.206
 	#
-	# NDOTS is optional, defaulting to 1. This provides an adjustable threshold to
+	# All 3 arguments are optional:
+	# 
+	# resolv:RESOLV-CONF-FILE (default: /etc/resolv.conf) If specified, coredns
+	# uses this file to get the host's search domains. CoreDNS performs a lookup
+	# on these domains if the in-cluster search domains in the path fail to 
+	# produce an answer. If not specified, the values will be read from the local 
+	# resolv.conf file (i.e the resolv.conf file in the pod containing coredns).
+	#
+	# ndots:NDOTS (default: 1) This provides an adjustable threshold to
 	# prevent server side lookups from triggering. If the number of dots before
 	# the first search domain is less than this number, then the search path will
 	# not executed on the server side.
 	#
+	# onnxdomain:RESPONSE (default: NXDOMAIN) RESPONSE can be either SERVFAIL or
+	# NOERROR. This option causes coredns to return the given response instead of
+	# NXDOMAIN when the all searches in the path produce no results. Doing this
+	# should prevent the client from fruitlessly continuing the client side 
+	# searches in the path.
+	#
 	# Enabling autopath causes coredns to use more memory since it needs to
 	# maintain a watch on all pods. If autopath and "pods verified" mode are
-	# both enabled, they will use share the same watch. I.e. enabling both options
+	# both enabled, they will share the same watch. I.e. enabling both options
 	# should have an equivalent memory impact of just one.
 	autopath ndots:1 foo.example.com.
 
--- a/middleware/kubernetes/handler.go
+++ b/middleware/kubernetes/handler.go
@@ -43,35 +43,46 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 
 	records, extra, _, err := k.routeRequest(zone, state)
 
-	if k.AnticipatePaths && k.IsNameError(err) {
+	if k.AutoPath && k.IsNameError(err) {
 		p := k.findPodWithIP(state.IP())
-		if p != nil {
-			name, path, ok := splitSearchPath(zone, state.QName(), p.Namespace)
-			if ok {
-				// try the "svc.cluster.local" and "cluster.local" paths
-				for i := 0; i < 2; i++ {
-					path = strings.Join(strings.Split(path, ".")[1:], ".")
-					state = state.NewWithQuestion(strings.Join([]string{name, path}, "."), state.QType())
-					records, extra, _, err = k.routeRequest(zone, state)
-					if !k.IsNameError(err) {
-						break
-					}
-				}
-				if k.IsNameError(err) {
-					// Fallthrough with the host domain path (if set)
-					wr := rewrite.NewResponseReverter(w, r)
-					if k.HostDomainPath != "" {
-						r = state.NewWithQuestion(strings.Join([]string{name, k.HostDomainPath}, "."), state.QType()).Req
-						rcode, nextErr := middleware.NextOrFailure(k.Name(), k.Next, ctx, wr, r)
-						if rcode == dns.RcodeSuccess {
-							return rcode, nextErr
-						}
-					}
-					// Fallthrough with the . path
-					r = state.NewWithQuestion(strings.Join([]string{name, ""}, "."), state.QType()).Req
-					return middleware.NextOrFailure(k.Name(), k.Next, ctx, wr, r)
+		for p != nil {
+			name, path, ok := splitSearch(zone, state.QName(), p.Namespace)
+			if !ok {
+				break
+			}
+			if strings.Count(name, ".") < k.AutoPathNdots {
+				break
+			}
+			// Search "svc.cluster.local" and "cluster.local"
+			for i := 0; i < 2; i++ {
+				path = strings.Join(strings.Split(path, ".")[1:], ".")
+				state = state.NewWithQuestion(strings.Join([]string{name, path}, "."), state.QType())
+				records, extra, _, err = k.routeRequest(zone, state)
+				if !k.IsNameError(err) {
+					break
 				}
 			}
+			if !k.IsNameError(err) {
+				break
+			}
+			// Fallthrough with the host search path (if set)
+			wr := rewrite.NewResponseReverter(w, r)
+			for _, hostsearch := range k.HostSearchPath {
+				r = state.NewWithQuestion(strings.Join([]string{name, hostsearch}, "."), state.QType()).Req
+				rcode, nextErr := middleware.NextOrFailure(k.Name(), k.Next, ctx, wr, r)
+				if rcode == dns.RcodeSuccess {
+					return rcode, nextErr
+				}
+			}
+			// Search . in this middleware
+			state = state.NewWithQuestion(strings.Join([]string{name, "."}, ""), state.QType())
+			records, extra, _, err = k.routeRequest(zone, state)
+			if !k.IsNameError(err) {
+				break
+			}
+			// Search . in the next middleware
+			r = state.Req
+			return middleware.NextOrFailure(k.Name(), k.Next, ctx, wr, r)
 		}
 	}
 
--- a/middleware/kubernetes/handler.go
+++ b/middleware/kubernetes/handler.go
@@ -50,12 +50,12 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 			if !ok {
 				break
 			}
-			if strings.Count(name, ".") < k.AutoPath.NDots {
+			if (dns.CountLabel(name) - 1) < k.AutoPath.NDots {
 				break
 			}
 			// Search "svc.cluster.local" and "cluster.local"
 			for i := 0; i < 2; i++ {
-				path = strings.Join(strings.Split(path, ".")[1:], ".")
+				path = strings.Join(dns.SplitDomainName(path)[1:], ".")
 				state = state.NewWithQuestion(strings.Join([]string{name, path}, "."), state.QType())
 				records, extra, _, err = k.routeRequest(zone, state)
 				if !k.IsNameError(err) {
--- a/middleware/kubernetes/handler.go
+++ b/middleware/kubernetes/handler.go
@@ -43,14 +43,14 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 
 	records, extra, _, err := k.routeRequest(zone, state)
 
-	if k.AutoPath && k.IsNameError(err) {
+	if k.AutoPath.Enabled && k.IsNameError(err) {
 		p := k.findPodWithIP(state.IP())
 		for p != nil {
 			name, path, ok := splitSearch(zone, state.QName(), p.Namespace)
 			if !ok {
 				break
 			}
-			if strings.Count(name, ".") < k.AutoPathNdots {
+			if strings.Count(name, ".") < k.AutoPath.NDots {
 				break
 			}
 			// Search "svc.cluster.local" and "cluster.local"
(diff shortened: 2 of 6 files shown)


johnbelamaric:
Please rebase

middleware/kubernetes/README.md
```diff
@@ -121,6 +121,42 @@ kubernetes coredns.local {
 	# Each line consists of the name of the federation, and the domain.
 	federation myfed foo.example.com
 	
+	# autopath [<host-domain>]
+	#
+	# Enables server side search path lookups for pods.  When enabled, coredns
+	# will identify search path queries from pods and perform the remaining path
+	# lookups on the pod's behalf.  The search path used mimics the resolve.conf
```
johnbelamaric: s/resolve/resolv/

middleware/kubernetes/README.md
```diff
@@ -121,6 +121,42 @@ kubernetes coredns.local {
...
+	# search path deployed to pods. E.g.
+	#
+	#     search namespace.svc.cluster.local svc.cluster.local cluster.local
+	#
+	# <host-domain> is optional. If specified, coredns performs a lookup on this
```
johnbelamaric: 1) should we allow more than one domain here
2) should we have an ndots-like setting (default to 1)

middleware/kubernetes/README.md
```diff
@@ -121,6 +121,42 @@ kubernetes coredns.local {
 	# Each line consists of the name of the federation, and the domain.
 	federation myfed foo.example.com
 	
+	# autopath [<host-domain>]
```
miekg: Nit: `<host-domain>` -> `HOST-DOMAIN`

middleware/kubernetes/README.md
miekg: Is is really needed to *extend* to already ridiculous search path?

middleware/kubernetes/handler.go
```diff
@@ -39,37 +41,40 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
...
-		records, extra, _, err = middleware.SRV(&k, zone, state, middleware.Options{})
-	case "SOA":
-		records, _, err = middleware.SOA(&k, zone, state, middleware.Options{})
-	case "NS":
-		if state.Name() == zone {
-			records, extra, _, err = middleware.NS(&k, zone, state, middleware.Options{})
-			break
+	records, extra, _, err := k.routeRequest(zone, state)
+
+	if k.AnticipatePaths && k.IsNameError(err) {
```
miekg: AnticipatePath -> AutoPath?

middleware/kubernetes/handler.go
```diff
@@ -39,37 +41,40 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
...
+		p := k.findPodWithIP(state.IP())
+		if p != nil {
+			name, path, ok := splitSearchPath(zone, state.QName(), p.Namespace)
+			if ok {
```
miekg: `If !ok { break }` ? and dedent the rest?

middleware/kubernetes/handler.go
```diff
@@ -39,37 +41,40 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
...
+					wr := rewrite.NewResponseReverter(w, r)
+					if k.HostDomainPath != "" {
+						r = state.NewWithQuestion(strings.Join([]string{name, k.HostDomainPath}, "."), state.QType()).Req
+						rcode, nextErr := middleware.NextOrFailure(k.Name(), k.Next, ctx, wr, r)
+						if rcode == dns.RcodeSuccess {
+							return rcode, nextErr
+						}
+					}
+					// Fallthrough with the . path
+					r = state.NewWithQuestion(strings.Join([]string{name, ""}, "."), state.QType()).Req
```
miekg: This `join` reads a bit weird?

middleware/kubernetes/handler.go
chrisohaver: We're not in a loop here. I could use a goto I suppose.  Depends on how _much_ you dislike indentation.

middleware/kubernetes/README.md
chrisohaver: 1) It's a matter of what kubernetes allows you to do, and we need to be able to mimic the search path.  I'll have to see how these are configured in kubernetes, and if they allow more than one.  Or we can allow more than one regardless ...  cant hurt.

2) ... Yes perhaps ... Cant hurt to allow the user to tweak the behavior.

chrisohaver:
As i was sketching out scenarios of the ndots queries, I found some cases that I dont handle.

specifically , when trying "." (the last lookup), we cant fall though right away, we need to check locally in kubernetes first.  The reason is that a pod with ndots=5 will run a fully qualified domain through the search path...
```
# host -v -t a service.namespace.svc.cluster.local
Trying "service.namespace.svc.cluster.local.default.svc.cluster.local"
```
Which means we need to try `service.namespace.svc.cluster.local` in "." locally first, then fall through.

middleware/kubernetes/README.md
chrisohaver: for 1 above, would be better if i could pull this configuration from kubernetes somehow, rather than ask the user for it.

middleware/kubernetes/handler.go
johnbelamaric: GitHub won't let me respond to the question above this for some weird reason, so it's here. We don't WANT to extend the search path. But in the pod the search path can include more than the cluster domain, so we need to allow it. For example, on EC2 it had something like "ec2.internal" in there after the k8s paths.

middleware/kubernetes/handler.go
chrisohaver: I changed `if p != nil ` to `for p != nil ` which allowed me to reduce indentation in several places with early breaks... return at the end means the loop can only execute once...

middleware/kubernetes/README.md
chrisohaver: FYI, the kubelet config option to control the domain appears to be ...
`      --cluster-domain string                                   Domain for this cluster.  If set, kubelet will configure all containers to search this domain in addition to the host's search domains
`

... which implies only one is allowed.

chrisohaver: ok.  k8s allows > 1 host domain. they are supposedly copied from domain entries from the resolv.conf of the host node.

... which means it can be different on each node? hmm, i hope not.

johnbelamaric:
Ok, so we should be able to get them from there too, then. That is, from our resolv.conf

---
John Belamaric
(240) 383-6963

On Jun 23, 2017, at 10:48 PM, chrisohaver <notifications@github.com<mailto:notifications@github.com>> wrote:


@chrisohaver commented on this pull request.

In middleware/kubernetes/README.md<https://github.com/coredns/coredns/pull/750#discussion_r123868800>:

> @@ -121,6 +121,42 @@ kubernetes coredns.local {
        # Each line consists of the name of the federation, and the domain.
        federation myfed foo.example.com<http://foo.example.com>

+       # autopath [<host-domain>]
+       #
+       # Enables server side search path lookups for pods.  When enabled, coredns
+       # will identify search path queries from pods and perform the remaining path
+       # lookups on the pod's behalf.  The search path used mimics the resolve.conf
+       # search path deployed to pods. E.g.
+       #
+       #     search namespace.svc.cluster.local svc.cluster.local cluster.local
+       #
+       # <host-domain> is optional. If specified, coredns performs a lookup on this


ok. k8s allows > 1 host domain. they are supposedly copied from domain entries from the resolv.conf of the host node.

chrisohaver:
> Ok, so we should be able to get them from there too, then. That is, from our resolv.conf

Yes, because the coredns pod uses the "Default" dns-policy, which inherits resolve.conf from the node its running on.

If the resolv.conf configs can be different across different nodes in a cluster, we will need to add a caveat.... or allow the user to specify a host domain list per node (or figure out how to retrieve the resolv.conf from other nodes).  We can identify what node a pod is running on.

chrisohaver:
>> Ok, so we should be able to get them from there too, then. That is, from our resolv.conf

> Yes, because the coredns pod uses the "Default" dns-policy, which inherits resolve.conf from the node its running on.

Unless coredns is not running in cluster, in which case we cant use the local resolv.conf - So we'll need to allow the user to override with a specified list.

chrisohaver:
Another possible user option:
When a client makes an ill-fated query (ill-fated = one that wont be resolved by anything in the search path), the final upstream server side lookup will return NXDOMAIN, which is returned to the client.  This will trigger the client to pointlessly run the query through the entire path itself (client side).

If coredns were to return a SERVFAIL however, the client would not run through the path.
... or NOERROR, with an empty data set ... I think would accomplish the same.

We could make this an optional behavior.

middleware/kubernetes/setup.go
```diff
@@ -196,7 +229,33 @@ func kubernetesParse(c *caddy.Controller) (*Kubernetes, error) {
 	return nil, errors.New("kubernetes setup called without keyword 'kubernetes' in Corefile")
 }
 
+func getSearchPathFromResolvConf(filename string) ([]string, error) {
```
johnbelamaric: I think you could use https://github.com/miekg/dns/blob/master/clientconfig.go#L22 to do this parsing.

middleware/kubernetes/setup.go
```diff
@@ -187,7 +189,38 @@ func kubernetesParse(c *caddy.Controller) (*Kubernetes, error) {
...
+							if err != nil {
+								return nil, fmt.Errorf("invalid ndots option for autopath, got '%v', expected an integer", ndots)
+							}
+							k8s.AutoPathNdots = ndots
+							gotNdots = true
+							continue
+						}
+						hostdomains = append(hostdomains, arg)
+					}
+					k8s.HostSearchPath = hostdomains
```
johnbelamaric: I am wondering if we shouldn't bother with the additional search list and just use what comes from resolv.conf. Then we could allow the option of giving a different filename here (other than /etc/resolv.conf), which would give the same capability but without us having to do the special processing here.

johnbelamaric: Oh, I see from your comment above we do need to allow them to override in some way or another. I think though if we make them use a resolv.conf like file that is OK for the rare case of out of cluster.

johnbelamaric:
I think the SERVFAIL idea is a good one, otherwise bad queries will still have the high latency of all the extra requests. @miekg do you see any issue with that? In normal K8s usage, there are no other name servers in the resolv.conf so it will just stop there.

middleware/kubernetes/setup.go
chrisohaver: ok - Less code is good.  To clarify , the ndots config would not come from the resolv.conf.  It should still be a separate parameter.

johnbelamaric:
Yes, that makes sense.

On Jun 26, 2017, at 10:20 AM, chrisohaver <notifications@github.com<mailto:notifications@github.com>> wrote:


@chrisohaver commented on this pull request.

In middleware/kubernetes/setup.go<https://github.com/coredns/coredns/pull/750#discussion_r124021734>:

> +                                     for _, arg := range args {
+                                               if strings.HasPrefix(arg, ndotsOptPrefix) {
+                                                       if gotNdots {
+                                                               return nil, fmt.Errorf("found more than one ndots option, expected at most one")
+                                                       }
+                                                       ndots, err := strconv.Atoi(arg[len(ndotsOptPrefix):])
+                                                       if err != nil {
+                                                               return nil, fmt.Errorf("invalid ndots option for autopath, got '%v', expected an integer", ndots)
+                                                       }
+                                                       k8s.AutoPathNdots = ndots
+                                                       gotNdots = true
+                                                       continue
+                                               }
+                                               hostdomains = append(hostdomains, arg)
+                                       }
+                                       k8s.HostSearchPath = hostdomains


ok - Less code is good. To clarify , the ndots config would not come from the resolv.conf. It should still be a separate parameter.

middleware/kubernetes/setup.go
chrisohaver: Yeah - i'm not handling domain and search path correctly anyways.  I was appending the two.  Apparently the domain is only used if search is not specified

miekg:
>I think the SERVFAIL idea is a good one, otherwise bad queries will still have the high latency of all the extra requests. @miekg do you see any issue with that? In normal K8s usage, there are no other name servers in the resolv.conf so it will just stop there.

I agree.

miekg:
>I am wondering if we shouldn't bother with the additional search list and just use what comes from resolv.conf. Then we could allow the option of giving a different filename here (other than /etc/resolv.conf), which would give the same capability but without us having to do the special processing here.

Yes please.

chrisohaver:
The "autopath" setting will have 3 optional parameters.
- one for for setting ndots threshold
- one for specifying path to resolv.conf
- one for returning SERVFAIL (or NOERROR) instead of NXDOMAIN

@miekg do you have a established/preferred way to capture this in the config?
- one line, with named parameters e.g. 
`autopath ndots:1 resolv:/etc/resolv.conf onfail:SERVFAIL`

- separate setting for each e.g.
```
autopath
autopathndots 1
autopathresolv /etc/resolv.conf
autopathonfail SERVFAIL
```

- parameters in a nested block
```
autopath {
   nodts 1
   resolve /etc/resolv.conf
   onfail SERVFAIL
}
```

johnbelamaric:
The nested block is probably the best way to do this. Should "onfail" be "onnxdomain" ?

chrisohaver:
> The nested block is probably the best way to do this. Should "onfail" be "onnxdomain" ?

yes, and `nodts` should be `ndots`... and `resolve` should be `resolv` ...since its referring to the file.

chrisohaver:
> The nested block is probably the best way

May not be supported by the caddy definition...

from https://caddyserver.com/docs/caddyfile ...
> Subdirectives cannot open new blocks. In other words, nested directive blocks are not supported.

I think "directives" map to "middleware" for us... (and sub-directives are the options in the middleware)

johnbelamaric:
Bummer. I thought they were supported. Then single line format, in my opinion.

On Jun 26, 2017, at 3:14 PM, chrisohaver <notifications@github.com<mailto:notifications@github.com>> wrote:


The nested block is probably the best way

May not be supported by the caddy definition...

from https://caddyserver.com/docs/caddyfile ... I think "directives" map to "middleware" for us...

Subdirectives cannot open new blocks. In other words, nested directive blocks are not supported.

miekg:
single line sgtm, note we don't have these prefixed `ndots:x` yet, I think. And I do like sane defaults, so
`autopath` should do something sane. Is the `fail` really needed?

miekg:
Haven't following this discussion super closely, so sorry if I ask the obvious. Why is ndots needs when there is a resolv.conf? To override it? Why?

chrisohaver:
> Haven't following this discussion super closely, so sorry if I ask the obvious. Why is ndots needs when there is a resolv.conf? To override it? Why?

It's not obvious at all.  They are not the same thing.  ndots in this context refers to the minimum number of dots required to still do a server side lookup.  The number of dots here is in the (presumed) original search query.  It's there to allow the user to control default behavior for ambiguous cases. For example:

Case 1: user's non-fully qualified query gets expanded by search path
1. User searches for `google` (this is the original search term)
2. This gets expanded `google.namespace.svc.cluster.local.`, and sent to coredns.
3. `google` isn't a service in `namespace`, so we check to see if we should do server side expansion
4. coredns splits this into `google` and `namespace.svc.cluster.local.`, and assumes it was expanded by the client search path.
5. `google` has < 1 dots, so we skip server side lookups, and return NXDOMAIN (or fall though if thats configured)
6. The client then does the remaining search path lookups from the client side

Case 2: user makes fully qualified query
1. User searches for `google.namespace.svc.cluster.local.`
2. This doesnt expanded, because it's fully qualified, and sent to coredns as is.
(at this point, coredns receives the same exact query as in case 1. hence the ambiguity)
3. `google` isnt a service in `namespace`, so we check to see if we should do server side expansion
4. coredns splits this into `google` and `namespace.svc.cluster.local.`, and assumes it was expanded by the client search path.
5. `google` has < 1 dots, so we skip server side lookups, and return NXDOMAIN (or fall though if thats configured)
6. The client gets NXDOMAIN, but doesnt continue with search path lookups, since original query was fully qualified.

chrisohaver:
> It's not obvious at all. They are not the same thing.

I considered naming it something else.  Perhaps we should.

minDots? xDots? skipDots?  skipAutoPathIfLessThanNDots?

miekg:
>> Haven't following this discussion super closely, so sorry if I ask the obvious. Why is ndots needs when there is a resolv.conf? To override it? Why?
>
>Its not obvious at all.  They are not the same thing.  ndots in this context refers to the minimum number of dots required to still do a server side lookup.  The number of dots here is in the (presumed) original search query.  It's there to allow the user to control default behavior for ambiguous cases. For example:
>
>Case 1: user's non-fully qualified query gets expanded by search path
>1. User searches for `google` (this is the original search term)
>2. This gets expanded `google.namespace.svc.cluster.local.`, and sent to coredns.
>3. `google` isn't a service in `namespace`, so we check to see if we should do server side expansion
>4. coredns splits this into `google` and `namespace.svc.cluster.local.`, and assumes it was expanded by the client search path.
>5. `google` has < 1 dots, so we skip server side lookups, and return NXDOMAIN (or fall though if thats configured)
>6. The client then does the remaining search path lookups from the client side
>
>Case 2: user makes fully qualified query
>1. User searches for `google.namespace.svc.cluster.local.`
>2. This doesnt expanded, because its fully qualified, and sent to coredns as is.
>(at this point, coredns receives the same exact query as in case 1. hence the ambiguity)
>3. `google` isnt a service in `namespace`, so we check to see if we should do server side expansion
>4. coredns splits this into `google` and `namespace.svc.cluster.local.`, and assumes it was expanded by the client search path.
>5. `google` has < 1 dots, so we skip server side lookups, and return NXDOMAIN (or fall though if thats configured)
>6. The client gets NXDOMAIN, but doesnt continue with search path lookups, since original query was fully qualified.

thanks Chris, useful info. But can't we infer this from our /etc/resolv. We *do* 
assume we have access to the same one the client has? So we "subtract" our 
searchpaths, shortest name left wins. And then do steps 3 to 5 from above (which 
are the same in both lists). 6. is out of our control anyway?

Am I missing the nuance?

/Miek

--
Miek Gieben

johnbelamaric:
It's not actually related to the search path or ndots in the host or client resolv.conf. It works similarly to that ndots though. The idea is that when we "guess" the original query, we need to know if we should do the server side path search or not. That is, this is used as the ndots based on our guess as to what the real query was.

It will almost always be 1. If they search for "google", do not use server side search, if the search for google.com<http://google.com>, DO use the server side search path. Maybe it doesn't need to be configurable, but it's hard to predict all the use cases.

---
John Belamaric
(240) 383-6963

On Jun 26, 2017, at 5:59 PM, Miek Gieben <notifications@github.com<mailto:notifications@github.com>> wrote:

[ Quoting <notifications@github.com<mailto:notifications@github.com>> in "Re: [coredns/coredns] middleware/ku..." ]
>> Haven't following this discussion super closely, so sorry if I ask the obvious. Why is ndots needs when there is a resolv.conf? To override it? Why?
>
>Its not obvious at all. They are not the same thing. ndots in this context refers to the minimum number of dots required to still do a server side lookup. The number of dots here is in the (presumed) original search query. It's there to allow the user to control default behavior for ambiguous cases. For example:
>
>Case 1: user's non-fully qualified query gets expanded by search path
>1. User searches for `google` (this is the original search term)
>2. This gets expanded `google.namespace.svc.cluster.local.`, and sent to coredns.
>3. `google` isn't a service in `namespace`, so we check to see if we should do server side expansion
>4. coredns splits this into `google` and `namespace.svc.cluster.local.`, and assumes it was expanded by the client search path.
>5. `google` has < 1 dots, so we skip server side lookups, and return NXDOMAIN (or fall though if thats configured)
>6. The client then does the remaining search path lookups from the client side
>
>Case 2: user makes fully qualified query
>1. User searches for `google.namespace.svc.cluster.local.`
>2. This doesnt expanded, because its fully qualified, and sent to coredns as is.
>(at this point, coredns receives the same exact query as in case 1. hence the ambiguity)
>3. `google` isnt a service in `namespace`, so we check to see if we should do server side expansion
>4. coredns splits this into `google` and `namespace.svc.cluster.local.`, and assumes it was expanded by the client search path.
>5. `google` has < 1 dots, so we skip server side lookups, and return NXDOMAIN (or fall though if thats configured)
>6. The client gets NXDOMAIN, but doesnt continue with search path lookups, since original query was fully qualified.

thanks Chris, useful info. But can't we infer this from our /etc/resolv. We *do*
assume we have access to the same one the client has? So we "subtract" our
searchpaths, shortest name left wins. And then do steps 3 to 5 from above (which
are the same in both lists). 6. is out of our control anyway?

Am I missing the nuance?

/Miek

--
Miek Gieben

chrisohaver:
IMO it should be configurable. 2 is also valid (more conservative). When you consider the host domains, 0 is also valid. In fact it would make sense to make the check after the host domains searches - I.e. Don't skip all server side looks ups, just skip the last fallthrough.  This way a search for 'foo' would get tried on foo.my.domain.com<http://foo.my.domain.com>. On the server side...

Sent from my mobile

On Jun 26, 2017, at 7:33 PM, John Belamaric <notifications@github.com<mailto:notifications@github.com>> wrote:

It's not actually related to the search path or ndots in the host or client resolv.conf. It works similarly to that ndots though. The idea is that when we "guess" the original query, we need to know if we should do the server side path search or not. That is, this is used as the ndots based on our guess as to what the real query was.

It will almost always be 1. If they search for "google", do not use server side search, if the search for google.com<http://google.com><http://google.com>, DO use the server side search path. Maybe it doesn't need to be configurable, but it's hard to predict all the use cases.

---
John Belamaric
(240) 383-6963

On Jun 26, 2017, at 5:59 PM, Miek Gieben <notifications@github.com<mailto:notifications@github.com><mailto:notifications@github.com>> wrote:

[ Quoting <notifications@github.com<mailto:notifications@github.com><mailto:notifications@github.com>> in "Re: [coredns/coredns] middleware/ku..." ]
>> Haven't following this discussion super closely, so sorry if I ask the obvious. Why is ndots needs when there is a resolv.conf? To override it? Why?
>
>Its not obvious at all. They are not the same thing. ndots in this context refers to the minimum number of dots required to still do a server side lookup. The number of dots here is in the (presumed) original search query. It's there to allow the user to control default behavior for ambiguous cases. For example:
>
>Case 1: user's non-fully qualified query gets expanded by search path
>1. User searches for `google` (this is the original search term)
>2. This gets expanded `google.namespace.svc.cluster.local.`, and sent to coredns.
>3. `google` isn't a service in `namespace`, so we check to see if we should do server side expansion
>4. coredns splits this into `google` and `namespace.svc.cluster.local.`, and assumes it was expanded by the client search path.
>5. `google` has < 1 dots, so we skip server side lookups, and return NXDOMAIN (or fall though if thats configured)
>6. The client then does the remaining search path lookups from the client side
>
>Case 2: user makes fully qualified query
>1. User searches for `google.namespace.svc.cluster.local.`
>2. This doesnt expanded, because its fully qualified, and sent to coredns as is.
>(at this point, coredns receives the same exact query as in case 1. hence the ambiguity)
>3. `google` isnt a service in `namespace`, so we check to see if we should do server side expansion
>4. coredns splits this into `google` and `namespace.svc.cluster.local.`, and assumes it was expanded by the client search path.
>5. `google` has < 1 dots, so we skip server side lookups, and return NXDOMAIN (or fall though if thats configured)
>6. The client gets NXDOMAIN, but doesnt continue with search path lookups, since original query was fully qualified.

thanks Chris, useful info. But can't we infer this from our /etc/resolv. We *do*
assume we have access to the same one the client has? So we "subtract" our
searchpaths, shortest name left wins. And then do steps 3 to 5 from above (which
are the same in both lists). 6. is out of our control anyway?

Am I missing the nuance?

/Miek

--
Miek Gieben

miekg:
>It's not actually related to the search path or ndots in the host or client resolv.conf. It works similarly to that ndots though. The idea is that when we "guess" the original query, we need to know if we should do the server side path search or not. That is, this is used as the ndots based on our guess as to what the real query was.

understood, but if you have access to the searchpath yourself, that would be a 
more robust way of guessing what was added and thus what the original name 
should be?

chrisohaver:
We actually don't need data from resolv.conf to infer the first search path... in fact, we could not infer it from resolv.conf, because coredns will be running with a different dns policy than the pods.

The ambiguity that forces us to have this setting lies with whether the user did ...

`host test`
Vs
`host test.podns.svc.cluster.local.`

Assuming the client pods namespace is `podns`, the first query is expanded client side to `test.podns.svc.cluster.local.` the second query is not expanded client side.

When the queries reach coredns, they are indistinguishable from each other. They both appear to have been expanded by client side search path. The first should get server side expansion. The second should not. But without any way to tell the difference, we have to arbitrarily choose one way to behave.  With ndots of 1, the decision is to skip server side lookups. With ndots of 0, the decision would be to not skip.


Sent from my mobile

On Jun 27, 2017, at 5:48 AM, Miek Gieben <notifications@github.com<mailto:notifications@github.com>> wrote:

[ Quoting <notifications@github.com<mailto:notifications@github.com>> in "Re: [coredns/coredns] middleware/ku..." ]
>It's not actually related to the search path or ndots in the host or client resolv.conf. It works similarly to that ndots though. The idea is that when we "guess" the original query, we need to know if we should do the server side path search or not. That is, this is used as the ndots based on our guess as to what the real query was.

understood, but if you have access to the searchpath yourself, that would be a
more robust way of guessing what was added and thus what the original name
should be?

miekg:
>We actually don't need data from resolv.conf to infer the first search path... in fact, we could not infer it from resolv.conf, because coredns will be running with a different dns policy than the pods.

ah. OK. We *don't* share a /etc/resolv.conf, yes that makes it way harder.
(just checked on a GKE k8s node - indeed normal resolv.conf - now I'm left 
wondering where that crazy ndots:5 resolv.conf is configured.)

>The ambiguity that forces us to have this setting lies with whether the user did ...
>
>`host test`
>Vs
>`host test.podns.svc.cluster.local.`
>
>Assuming the client pods namespace is `podns`, the first query is expanded client side to `test.podns.svc.cluster.local.` the second query is not expanded client side.
>
>When the queries reach coredns, they are indistinguishable from each other. They both appear to have been expanded by client side search path. The first should get server side expansion. The second should not. But without any way to tell the difference, we have to arbitrarily choose one way to behave.  With ndots of 1, the decision is to skip server side lookups. With ndots of 0, the decision would be to not skip.

yeah. I see. But how should one make the tradeoff for setting this value?

chrisohaver:
>I'm left wondering where that crazy ndots:5 resolv.conf is configured.

It's on the pod itself, configured based on the dns policy assigned to the pod.  There are two dns policies, "Default" and "ClusterFirst".  The default dns policy for pods is "ClusterFirst" ... ironically.  The "ClusterFirst" policy uses the long search path and high ndots.

> But how should one make the tradeoff for setting this value?

The trade off is optimization vs correctness.  Lower ndots means we do server-side queries more often, but there's a higher chance we'll do server side search path for a query that we should not run through the search path. Higher ndots means less of a chance, at the expense if not optimizing any queries with ndots less than the setting.

johnbelamaric:
You mean >= the setting? If path-n-dots == 1, then we do want to run "google.com" through the optimization, correct? And also "www.google.com". But not bother with "google" because it probably won't properly resolve? Have I confused myself here?

chrisohaver:
>>Higher ndots means less of a chance, at the expense if not optimizing any queries with ndots less than the setting.

> You mean >= the setting? If path-n-dots == 1, then we do want to run "google.com" through the optimization, correct? And also "www.google.com". But not bother with "google" because it probably won't properly resolve? Have I confused myself here?

We do not optimize when ndots in the presumed original query is less than the setting. 
For ndots of 1:
- `google.com` has 1 dot. 1 is not less than 1. So we optimize
- `google` has 0 dots. 0 is less than 1.  So we don't optimize.

johnbelamaric:
Ok, that's what I thought. Thanks.

chrisohaver:
Any thoughts on my comment earlier...
> ...it would make sense to make the [ndots] check after the host domains searches - I.e. Don't skip all server side looks ups, just skip the last fallthrough.  This way a search for 'foo' would get tried on foo.my.domain.com. On the server side...

e.g. For ndots = 1, and search path contains local domain `bignumbers.com`.  If user searches for `google` should we still search `google.bignumbers.com.` ?   Probably yes.

johnbelamaric:
Ah, interesting. Ok. Probably yes. Does this path specific ndots piece make sense or are we overthinking this one?

chrisohaver:
>>For ndots = 1, and search path contains local domain bignumbers.com. If user searches for google should we still search google.bignumbers.com. ? Probably yes. 

>Ah, interesting. Ok. Probably yes.

I think I'm about to disagree with myself.  In same example above, if user queries for fully qualified `google.namespace.svc.cluster.local.`  (with trailing dot, so it doesn't expand on client side) and the service `google` doesn't exist, should we try `google.bignumbers.com`.  Probably no.

So, we should leave it to the user to decide if they want to change default ndots behavior.

> Does this path specific ndots piece make sense

IMO, setting this to 1 wont be the right choice for everyone.  I think 0 would be a popular choice (possibly more so than 1).
Due to the ambiguity, there isn't a 100% correct solution, so it stands to be user configurable ... default is 1, mileage may vary.  I'd rather not remove the option just to add it back later.

chrisohaver:
I'm leaning toward making the default options optimize as much as possible, and let the user configure more conservative options.

ndots:0
onnxdomain:SERVFAIL

miekg:
>I'm leaning toward making the default options optimize as much as possible, and let the user configure more conservative options.
>
>ndots:0
>onnxdomain:SERVFAIL

agreed.

middleware/kubernetes/README.md
```diff
@@ -121,6 +121,59 @@ kubernetes coredns.local {
 	# Each line consists of the name of the federation, and the domain.
 	federation myfed foo.example.com
 	
+	# autopath [ndots:NDOTS] [resolv:RESOLV-CONF-FILE] [onnxdomain:RESPONSE] ...
```
miekg: We should drop these option prefixes IMHO, we don't use those anywhere else.
So just `autopath [NDOTS] [RESOLVE-CONF] [RESPOSE]`

middleware/kubernetes/handler.go
```diff
@@ -39,37 +41,55 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
...
+	records, extra, _, err := k.routeRequest(zone, state)
+
+	if k.AutoPath.Enabled && k.IsNameError(err) {
+		p := k.findPodWithIP(state.IP())
+		for p != nil {
+			name, path, ok := splitSearch(zone, state.QName(), p.Namespace)
+			if !ok {
+				break
+			}
+			if strings.Count(name, ".") < k.AutoPath.NDots {
```
miekg: should be dns.LabelCount

middleware/kubernetes/handler.go
```diff
@@ -39,37 +41,55 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
...
+				break
+			}
+			// Search "svc.cluster.local" and "cluster.local"
+			for i := 0; i < 2; i++ {
+				path = strings.Join(strings.Split(path, ".")[1:], ".")
```
miekg: strings.Split does not work with paths that contain `\.`

middleware/kubernetes/README.md
johnbelamaric: Yeah, since now we have no variable length args, we can do that.

chrisohaver: It would have to be....

autopath [NDOTS [RESOLV-CONF [RESPONSE]]]

If a user wants to specify a RESPONSE, they need to specify the other two options at their default values.

Is this acceptable?

miekg:
>
>
>
>> @@ -121,6 +121,59 @@ kubernetes coredns.local {
> 	# Each line consists of the name of the federation, and the domain.
> 	federation myfed foo.example.com
> 	
>+	# autopath [ndots:NDOTS] [resolv:RESOLV-CONF-FILE] [onnxdomain:RESPONSE] ...
>
>It would have to be....
>
>[NDOTS [RESOLVE-CONF [RESPONSE]]]
>
>Since they are optional... if a user wants to specify a response, they need to specify the other two options at their default values.

Er.. yes, I always get lost when I try to type that :)

/Miek

--
Miek Gieben

middleware/kubernetes/README.md
chrisohaver: perhaps ...

autopath [NDOTS [RESPONSE [RESOLV-CONF]]]

I think a custom resolv.conf would be the least likely thing someone would change (out of cluster coredns only), so make that the most inconvenient thing to change.

miekg:
>
>> @@ -121,6 +121,59 @@ kubernetes coredns.local {
> 	# Each line consists of the name of the federation, and the domain.
> 	federation myfed foo.example.com
> 	
>+	# autopath [ndots:NDOTS] [resolv:RESOLV-CONF-FILE] [onnxdomain:RESPONSE] ...
>
>perhaps ...
>
>autopath [NDOTS [RESPONSE [RESOLV-CONF]]]
>
>I think a custom resolv.conf would be the least likely thing someone would change (out of cluster coredns only), so make that the most inconvenient thing to change.

Agreed.

miekg (approved):

miekg:
lgtm

middleware/kubernetes/README.md
```diff
@@ -121,6 +121,58 @@ kubernetes coredns.local {
...
+	# file to get the host's search domains. CoreDNS performs a lookup on these
+	# domains if the in-cluster search domains in the path fail to produce an
+	# answer. If not specified, the values will be read from the local resolv.conf
+	# file (i.e the resolv.conf file in the pod containing coredns).
+	#
+	# Enabling autopath causes coredns to use more memory since it needs to
+	# maintain a watch on all pods. If autopath and "pods verified" mode are
+	# both enabled, they will share the same watch. I.e. enabling both options
+	# should have an equivalent memory impact of just one.
+	autopath ndots:1 foo.example.com.
```
johnbelamaric: Need to fix this example.

middleware/kubernetes/README.md
```diff
@@ -121,6 +121,58 @@ kubernetes coredns.local {
...
+	#
+	# Enables server side search path lookups for pods.  When enabled, coredns
+	# will identify search path queries from pods and perform the remaining
+	# lookups in the path on the pod's behalf.  The search path used mimics the
+	# resolv.conf search path deployed to pods. E.g.
+	#
+	#  search ns1.svc.cluster.local svc.cluster.local cluster.local foo.com
+	#
+	# If no domains in the path produce an answer, a lookup on the bare question
+	# attempted.
```
johnbelamaric: "will be attempted" or "is attempted".

johnbelamaric (changes requested):
Minor tweaks.

chrisohaver:
Fixed the grammar error and the example.

miekg:
Just a thought: I still think this functionality is useful outside kubernetes - if only just for testing. If the server shares my /etc/resolv.conf I could just let it shortcut the search path. Would be superhandy for debugging (I think)

miekg:
Where are the actualy ServeDNS tests for this?

chrisohaver:
> Where are the actualy ServeDNS tests for this

There have never been serveDNS tests in the k8s middleware. I started converting the existing Services tests to ServeDNS tests a few days ago. I have most of them working plus some more coverage, but I have not worked out how to handle fall through to other MW in chain and upstreams yet (required for testing autopath and upstream).

miekg:
>> Where are the actualy ServeDNS tests for this
>
>There have never been serveDNS tests in the k8s middleware. I started converting the existing Services tests to ServeDNS tests a few days ago. I have most of them working plus some more coverage, but I have not worked out how to handle fall through to other MW in chain and upstreams yet (required for testing autopath and upstream).

Ack.

Full integration can be done in the uppper test/ dir.

(I didn't mean ServeDNS per se, just test for this new functionality)

chrisohaver:
The integration tests dont fit the feature well,....

I didn' tadd tests for autopath in Travis CI k8s integration, because it currently doesn't run the queries from within a pod. Autopath (intentionally) only triggers when the source of the query is from a pod in the cluster. So, we can test that it isn't working when it shouldn't, but not that it is when it should.

Changing k8s CI integration tests to run the queries run from within a pod is complex.  ... might be able to do it by dropping the queries out as system calls wrapped in `kubectl run`, then  parsing the output.  Still, a significant refactor.

miekg:
I see what you mean... But I don't like having such a hacky/important feature without tests what so ever. Could we have unit tests then?

chrisohaver:
Since the bulk of the logic is in woven into ServeDNS, I started to write unit tests for ServeDNS.
But I have not worked out how to test the logic that conditionally falls through to next middleware.
I need to create a stub middleware perhaps.

I'll send a WIP PR for what I have so far. Testing the k8s middleware lookups part of autopath is better than nothing.

pmoroney:
In middleware/test/helpers.go there is a `func NextHandler(rcode int, err error) Handler` that might help with the next middleware testing.

chrisohaver:
>In middleware/test/helpers.go there is a func NextHandler(rcode int, err error) Handler that might help with the next middleware testing.

Thanks @pmoroney, that did help!

PR #754, opened 2017-06-24
Adjust codecov config (codecov.yml) for commit status

### LINKED ISSUE #752: codecov should not fail the whole build
Codecov is useful, but is should not fail the whole build (in github) when it fails.

@yongtang is that possible?

yongtang:
This fix adjusted codecov config (codecov.yml) so that
commit status is success as long as coverage > 40%.

This fix fixes #752.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,6 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 40%
+        threshold: null


yongtang:
This is a trial. Let's see if it works with codecov.

yongtang:
The CodeCov config change works. Merging...

Issue #691, opened 2017-06-02
wildcard DNS / mimic xip.io


jremond:
I started to look at implementing a behaviour similar to http://xip.io/
It's my first code in Go and the name is definitively not final but it is in my gurdil branch https://github.com/jremond/coredns/tree/gurdil

The idea is to resolve 
any.thing.1.2.3.4.example.com 
to 
1.2.3.4 

Corefile looks like
```
.:1053 {
    gurdil example.com
    log stdout
    errors stdout
}

```

Do you think it's worth putting more effort into a proper Go implementation ?

Thanks

miekg:
Sure this looks useful. Does xip.io only do the forward A address? Or also reverse. What about ipv6?

Reminds me of the `reverse` middleware, btw: https://github.com/coredns/coredns/tree/master/middleware/reverse

Note that middleware can also be 'out of tree', by just editing `middleware.cfg` you can add new one from wherever.

jremond:
From my understanding when reading https://github.com/basecamp/xip-pdns/blob/master/bin/xip-pdns
xip.io only supports IPv4 and only when requesting a domain like [anything.]IP.example.com
It also adds SOA and NS records from the configuration file.

I don't see which behaviour would be the right one for the reverse
should 192.30.253.112 resolves to 
* github.com
* 192.30.253.112.example.com
* 192.30.253.112.reverse.example.com

For ipv6 I don't think ":" is a valid character for a domain, how would you represent the IP to keep it human friendly ? Just remove it  and support only address with all 0 expanded ?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] wildcard DNS ..." ]
>From my understanding when reading https://github.com/basecamp/xip-pdns/blob/master/bin/xip-pdns
>xip.io only supports IPv4 and only when requesting a domain like [anything.]IP.example.com
>It also adds SOA and NS records from the configuration file.

Ack.

>I don't see which behaviour would be the right one for the reverse
>should 192.30.253.112 resolves to
>* github.com
>* 192.30.253.112.example.com
>* 192.30.253.112.reverse.example.com

Depends who owns these IPs, reverse is always a bit tricky in that regard.

>For ipv6 I don't think ":" is a valid character for a domain, how would you represent the IP to keep it human friendly ? Just remove it  and support only address with all 0 expanded ?

Actually `:` should just work:  dig a:a:a.mx, although legacy tools might have
[...]

PR #755, opened 2017-06-26
add wildcard parameter to allow resolving multiple name to the same IP

jremond:
See discussion https://github.com/coredns/coredns/issues/691

Implementation of a separate wildcard keyword to allow to resolve
any.thing.ip-1.2.3.4.example.org to 1.2.3.4

### DIFF
--- a/middleware/reverse/setup.go
+++ b/middleware/reverse/setup.go
@@ -38,6 +38,7 @@ func reverseParse(c *caddy.Controller) (nets networks, fall bool, err error) {
 	// normalize zones, validation is almost done by dnsserver
 	// TODO(miek): need sane helpers for these.
 	zones := make([]string, len(c.ServerBlockKeys))
+	wildcard := false
 
 	for i, str := range c.ServerBlockKeys {
 		zones[i] = middleware.Host(str).Normalize()
@@ -85,6 +86,9 @@ func reverseParse(c *caddy.Controller) (nets networks, fall bool, err error) {
 						return nil, false, err
 					}
 
+				case "wildcard":
+					wildcard = true
+
 				case "fallthrough":
 					fall = true
 
@@ -117,8 +121,12 @@ func reverseParse(c *caddy.Controller) (nets networks, fall bool, err error) {
 				if ipnet.IP.To4() == nil {
 					regexIP = regexMatchV6
 				}
+				prefix := "^"
+				if wildcard {
+					prefix += ".*"
+				}
 				regex, err := regexp.Compile(
-					"^" + strings.Replace( // inject ip regex into template
+					prefix + strings.Replace( // inject ip regex into template
 						regexp.QuoteMeta(template), // escape dots
 						regexp.QuoteMeta(templateNameIP),
 						regexIP,
--- a/middleware/reverse/README.md
+++ b/middleware/reverse/README.md
@@ -9,6 +9,7 @@ reverse NETWORK... {
     hostname TEMPLATE
     [ttl TTL]
     [fallthrough]
+    [wildcard]
 ~~~
 
 * **NETWORK** one or more CIDR formatted networks to respond on.
@@ -33,6 +34,12 @@ With IPv6 lookups the ":" is removed, and any zero ranged are expanded, e.g.,
 The `{zone[i]}` symbol is **optional** and can be replaced by a fixed (zone) string.
 The zone will be matched by the zones listed in *this* configuration stanza.
 `i` needs to be replaced with the index of the configured listener zones, starting with 1.
+
+### `wildcard`
+
+If `wildcard` is true :
+any.thing.ip-1.2.3.4.example.org resolves to 1.2.3.4 (ip-{ip}.{zone} is the hostname/template)
+
 
 ## Examples
 
--- a/middleware/reverse/reverse_test.go
+++ b/middleware/reverse/reverse_test.go
@@ -1,0 +1,71 @@
+package reverse
+
+import (
+	"net"
+	"regexp"
+	"testing"
+
+	"github.com/coredns/coredns/middleware"
+	"github.com/coredns/coredns/middleware/pkg/dnsrecorder"
+	"github.com/coredns/coredns/middleware/test"
+
+	"github.com/miekg/dns"
+	"golang.org/x/net/context"
+)
+
+func TestReverse(t *testing.T) {
+	_, net4, _ := net.ParseCIDR("10.1.1.0/24")
+	regexIP4, _ := regexp.Compile("^.*ip-" + regexMatchV4 + "\\.example\\.org\\.$")
+
+	em := Reverse{
+		Networks: networks{network{
+			IPnet:        net4,
+			Zone:         "example.org",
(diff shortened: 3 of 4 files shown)


miekg:
@pmoroney Want to give this one a try?
(I will also give it a check)

pmoroney:
Sure, I'll look into it in a few

middleware/reverse/setup.go
```diff
@@ -117,8 +121,12 @@ func reverseParse(c *caddy.Controller) (nets networks, fall bool, err error) {
 				if ipnet.IP.To4() == nil {
 					regexIP = regexMatchV6
 				}
+				prefix := "^"
+				if wildcard {
+					prefix += ".*"
```
pmoroney, line 126: nit - Removing the explicit ^ should work without needing to specify the .* at the beginning, so we can just set prefix to the empty string.

pmoroney (commented):
Question: assuming the following template: {ip}.{zone}
do we want wildcard to return a result for test1.2.3.4.example.net or do we want to require a "." between the ip and the prefix if the template starts with "{ip}"? I am only asking because I am not sure about all the use cases for this middleware. 

Also, I would like to see a test of the wildcard functionality added.

jremond:
Do you have an example of test (in another middleware) that you think would make sense ?

middleware/reverse/setup.go
jremond: After testing, setting prefix to "" does not work.

miekg:
This will either make 009 or be pushed to 010 (I want to release 009 quickly because of bugs found)

miekg:
Tests can be found in basically any middleware `TestChaos` in middleware/chaos is I think simplest.

middleware/reverse/setup.go
miekg: That doesn't make sense. 

~~~
diff --git a/middleware/reverse/setup.go b/middleware/reverse/setup.go
index ee426f85..663ebea6 100644
--- a/middleware/reverse/setup.go
+++ b/middleware/reverse/setup.go
@@ -121,7 +121,7 @@ func reverseParse(c *caddy.Controller) (nets networks, fall bool, err error) {
                                if ipnet.IP.To4() == nil {
                                        regexIP = regexMatchV6
                                }
-                               prefix := "^"
+                               prefix := ""
                                if wildcard {
                                        prefix += ".*"
                                }
diff --git a/middleware/reverse/setup_test.go b/middleware/reverse/setup_test.go
index 2198c1e3..65671a72 100644
--- a/middleware/reverse/setup_test.go
+++ b/middleware/reverse/setup_test.go
@@ -14,10 +14,10 @@ func TestSetupParse(t *testing.T) {
        _, net4, _ := net.ParseCIDR("10.1.1.0/24")
        _, net6, _ := net.ParseCIDR("fd01::/64")
 
-       regexIP6, _ := regexp.Compile("^ip-" + regexMatchV6 + "\\.domain\\.com\\.$")
-       regexIpv4dynamic, _ := regexp.Compile("^dynamic-" + regexMatchV4 + "-intern\\.dynamic\\.domain\\.com\\.$")
-       regexIpv6dynamic, _ := regexp.Compile("^dynamic-" + regexMatchV6 + "-intern\\.dynamic\\.domain\\.com\\.$")
-       regexIpv4vpndynamic, _ := regexp.Compile("^dynamic-" + regexMatchV4 + "-vpn\\.dynamic\\.domain\\.com\\.$")
+       regexIP6, _ := regexp.Compile("ip-" + regexMatchV6 + "\\.domain\\.com\\.$")
+       regexIpv4dynamic, _ := regexp.Compile("dynamic-" + regexMatchV4 + "-intern\\.dynamic\\.domain\\.com\\.$")
+       regexIpv6dynamic, _ := regexp.Compile("dynamic-" + regexMatchV6 + "-intern\\.dynamic\\.domain\\.com\\.$")
+       regexIpv4vpndynamic, _ := regexp.Compile("dynamic-" + regexMatchV4 + "-vpn\\.dynamic\\.domain\\.com\\.$")
 
        serverBlockKeys := []string{"domain.com.:8053", "dynamic.domain.com.:8053"}
 
~~~
fixes the tests

jremond: I don't mean the tests are broken, I mean 
test.ip-127.0.3.1.example.org does not resolve to 127.0.3.1 if I set the prefix to ""
I get a SERVFAIL in the logs

miekg:
Ah. That's a different thing then.

jremond:
@pmoroney I just added a test based on the one from the chaos middleware, let me know if it is what you asked for.
I'm still unable to make it work with an empty prefix instead of ".*", any hint on that ?

pmoroney:
The new test in reverse_test.go looks good, thank's for adding that. 
On line 126 of setup.go, if I change it to: `prefix = ""` it seems to work for me.

And also, while I was trying to figure out why the patch code coverage target wasn't hit, I realized that the added test in setup_test.go might not actually be testing anything. It might just work to copy the first test in the table (the `// with defaults` one), add the wildcard config flag, and make sure the regexp changed appropriately. Hopefully then the check will pass :)

jremond:
my Corefile is 
```
.:1053 {
    log stdout
    errors stdout
    reverse 127.0.0.0/8 {
        hostname ip-{ip}.example.org
        wildcard
    }
}
```
if I change the line 126 of setup.go to `prefix = ""`
then `dig +short  @127.0.0.1 -p 1053 test.ip-127.0.3.1.example.org`
does not return anything and the corresponding log line is 
```
.:1053
2017/07/10 21:57:09 [INFO] CoreDNS-008
CoreDNS-008
127.0.0.1 - [10/Jul/2017:21:57:21 -0400] "A IN test.ip-127.0.3.1.example.org. udp 59 false 4096" SERVFAIL 59 180.182µs
```

Can you share your `Corefile` ?


I pushed a new test in setup_test.go to see if I understand correctly what is expected.

pmoroney:
I was using the following corefile:
```
example.org:5533 {
    reverse 10.0.0.0/8 {
        wildcard
    }
}
```
And I was getting the following response:
```
$ dig @127.0.0.1 -p 5533 asip-10.1.2.3.example.org                                                                                                                        

; <<>> DiG 9.10.3-P4 <<>> @127.0.0.1 -p 5533 asip-10.1.2.3.example.org
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 2423
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
;; QUESTION SECTION:
;asip-10.1.2.3.example.org.     IN      A

;; ANSWER SECTION:
asip-10.1.2.3.example.org. 60   IN      A       10.1.2.3

;; Query time: 0 msec
;; SERVER: 127.0.0.1#5533(127.0.0.1)
;; WHEN: Sat Jul 08 23:50:16 MDT 2017
;; MSG SIZE  rcvd: 70
```

But it works either way, the new test looks great, and all the checks pass. So I am going to merge it anyways. Thanks again!

PR #757, opened 2017-06-26
middleware/cache: fix race

miekg:
While adding a parallel performance benchmark I stumbled on a race
condition (another reason to add performance benchmarks!), so this
PR makes sure the msg is created in a race free manor and adds the
parallel benchmark.


johnbelamaric:
LGTM

PR #763, opened 2017-06-30
mw/kubernetes: rewrite README

### LINKED ISSUE #728: middleware/kubernetes: Improve/Expand README

miekg:
Make the README comply with the other READMEs

Example our still left out.

Fixes #728 (some of it)

### DIFF
--- a/middleware/kubernetes/README.md
+++ b/middleware/kubernetes/README.md
@@ -26,163 +26,116 @@ Endpoint `A` records are constructed as "epname.myservice.mynamespace.svc.coredn
 * "svc" indicates this is a service
 * "coredns.local" is the zone
 
-Also supported are PTR and SRV records for services/endpoints.
+Also supported are PTR and SRV records for services and endpoints.
 
 ## Syntax
 
 This is an example kubernetes configuration block, with all options described:
 
-```
-# kubernetes <zone> [<zone>] ...
-#
-# Use kubernetes middleware for domain "coredns.local"
-# Reverse domain zones can be defined here (e.g. 0.0.10.in-addr.arpa),
-# or instead with the "cidrs" option.
-#
-kubernetes coredns.local {
+~~~
+kubernetes [ZONES...]
+~~~
 
-	# resyncperiod <period>
-	#
-	# Kubernetes data API resync period. Default is 5m
-	# Example values: 60s, 5m, 1h
-	#
-	resyncperiod 5m
+* **ZONES** zones it should be authoritative for. If empty, the zones from the configuration block
+    are used.
 
-	# endpoint <url>
-	#
-	# Use url for a remote k8s API endpoint.  If omitted, it will connect to
-	# k8s in-cluster using the cluster service account.
-	#
-	endpoint https://k8s-endpoint:8080
+This is the most simple configuration you can use, but there are numerous other options:
 
-	# tls <cert-filename> <key-filename> <cacert-filename>
-	#
-	# The tls cert, key and the CA cert filenanames for remote k8s connection.
-	# This option is ignored if connecting in-cluster (i.e. endpoint is not
-	# specified).
-	#
-	tls cert key cacert
-
-	# namespaces <namespace> [<namespace>] ...
-	#
-	# Only expose the k8s namespaces listed.  If this option is omitted
-	# all namespaces are exposed
-	#
-	namespaces demo
-
-	# lables <expression> [,<expression>] ...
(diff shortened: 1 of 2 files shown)


middleware/kubernetes/README.md
```diff
@@ -26,163 +26,116 @@ Endpoint `A` records are constructed as "epname.myservice.mynamespace.svc.coredn
...
+  section containing the record for the question that actually had an answer.  This means that the
+  question and answer will not match. For example:
+
+   # host -v -t a google.com
+   Trying "google.com.default.svc.cluster.local"
+   ;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 50957
+   ;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0
+
+   ;; QUESTION SECTION:
+   ;google.com.default.svc.cluster.local. IN A
```
miekg, line 116: @chrisohaver @johnbelamaric I thought we were going to use CNAMEs for this?

chrisohaver:
I already have a rewrite of the readme locally. Was going to do a PR today.

Sent from my mobile

On Jun 30, 2017, at 6:47 AM, Miek Gieben <notifications@github.com<mailto:notifications@github.com>> wrote:


@miekg commented on this pull request.

In middleware/kubernetes/README.md<https://github.com/coredns/coredns/pull/763#discussion_r125012907>:

> +   ~~~
+   search ns1.svc.cluster.local svc.cluster.local cluster.local foo.com<http://foo.com>
+   ~~~
+
+  If no domains in the path produce an answer, a lookup on the bare question will be attempted.
+  A successful response will contain a question section with the original question, and an answer
+  section containing the record for the question that actually had an answer.  This means that the
+  question and answer will not match. For example:
+
+   # host -v -t a google.com<http://google.com>
+   Trying "google.com.default.svc.cluster.local"
+   ;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 50957
+   ;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0
+
+   ;; QUESTION SECTION:
+   ;google.com.default.svc.cluster.local. IN A


@chrisohaver<https://github.com/chrisohaver> @johnbelamaric<https://github.com/johnbelamaric> I thought we were going to use CNAMEs for this?

chrisohaver:
>  I thought we were going to use CNAMEs for this?

It was part of the discussion, but we didn't land on it. I had suggested it as a work around of the mismatch name concern.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] mw/kubernetes..." ]
>>  I thought we were going to use CNAMEs for this?
>

>It was part of the discussion, but we didn't land on it. I had suggested it as a work around of the mismatch name concern.

Ok. but the current impl. will not work. I've seen rewrite fail with Chrome that 
checks this - Chrome is likely not running in a k8s network, but this is just an 
example of a process that does more checking. The answer need to match the 
qname.

chrisohaver:
If someone runs into this issue, they can disable autopath.  It's not enabled by default.
We could add the CNAME record answer later as optional behavior ... or as the default non-user configurable behavior if you are opposed to adding another configuration option.

Or would you prefer we implement the CNAME now (in 009)?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] mw/kubernetes..." ]
>If someone runs into this issue, they can disable autopath.  It's not enabled by default.
>We could add the CNAME record answer later as optional behavior ... or as the default non-user configurable behavior if you are opposed to adding another configuration option.
>
>Or would you prefer we implement the CNAME now (in 009)?

Yes, we should; I'm afraid the current impl. is too fragile...

chrisohaver:
Regarding the fact that I had already re-written the K8s readme... we need to coordinate better.
Perhaps we need a project mgmt tool to avoid wasted efforts.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] mw/kubernetes..." ]
>Regarding the fact that I had already re-written the K8s readme... we need to coordinate better.
>Perhaps we need a project mgmt tool to avoid wasted efforts.

meh... I learned something and it wasn't a big deal. You assigned the issue to 
yourself... so I have only me to blame.

Please no mgmt tools

/Miek

--
Miek Gieben

PR #769, opened 2017-07-01
middleware/etcd: move NewHTTPTransport to pkg/tls

VioletRainbows:


miekg:
These are all nice. Thanks!

PR #770, opened 2017-07-01
middleware/backend: add Records() to ServiceBackend interface

VioletRainbows:


VioletRainbows:
I believe that the next PR will be mostly moving and renaming of etcd and splitting etcd struct between etcd and backend.

Issue #766, opened 2017-06-30
mw/k8s: add more test coverage for autopath function


chrisohaver:

chrisohaver:
I have unit test coverage working for the k8s object lookup (second path lookup) through ServeDNS().
Also added several other cases not related to AutoPath while I was at it (A, SRV, PTR, local/remote federated records, external CNAME).

I haven't worked out how to mock next middleware to test the host domain and "." searches.

chrisohaver:
#771 completes good coverage for autopath, including mocked next middleware to test host domain and "." searches.

Issue #765, opened 2017-06-30
mw/k8s: add cname records to autopath responses


chrisohaver:
add cname records to "bridge the gap" between question section, and answer section questions in autopath responses

chrisohaver:
I have this working for k8s object lookups (with unit test coverage).

Adding the CNAME to the next middleware fall-through cases is more complicated, it requires extending the reverter/rcode-rewriter, to add CNAMES RRs for each Answer with a name that doesnt match the request name.  I've got it coded/compiled, but it not working yet.

chrisohaver:
implemented in #771

Issue #761, opened 2017-06-29
middleware/kubernetes: fix response rewrite in autopath


chrisohaver:
The RESPONSE option from autopath isn't working as expected.  DNS client behavior seems to be traversing the search path until data is returned, regardless of response code.  In effect, setting the response code does nothing.

miekg:
Guessing it will continue until it see rcode.Success?
The less options to configure the better IMO

chrisohaver:
The issue may be that I am not setting the rcode in the return correctly.
I am doing ...
```
wr := rewrite.NewResponseReverter(w, r)
...
rcode, nextErr := middleware.NextOrFailure(k.Name(), k.Next, ctx, wr, r)
...
rcode = dns.RcodeSuccess
r.Rcode = rcode
return rcode, nextErr
```
But the setting the rcode this way seems to have no effect on the rcode actually returned to the client, which receives NXDOMAIN.

miekg:
`NewResponseReverter` will writte to the client, whatever you do afterwards will only make CoreDNS write *another* message (a SERVFAIL) or do nothing (assuming response was already written).

chrisohaver:
Ah, I see how it works.  I think that I can create a new implementation of ResponseWriter, similar to ResponseReverter but also conditionally overwrites the response code based on the number of answers.

chrisohaver:
cool - it works.  Need to test more, but may not need to remove the option after all ...

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/ku..." ]
>Ah, I see how it works.  I think that I can create a new implementation of ResponseWriter, similar to ResponseReverter but also conditionally overwrites the response code based on the number of answers.

That's sounds really dangerous... why can't you use the return code you already 
have?

chrisohaver:
>why can't you use the return code you already have?

Because we need to change the response code from NXDOMAIN to NOERROR.  Thats the mechanism of the kludge... return a NOERROR, so the client doesn't continue searching the path.

chrisohaver:
To clarify - the logic is not based on the number of answers as I said above... It's based on the return code... e.g. If the return code is NXDOMAIN, change it to NOERROR when writing it to the client.

... so yes, it uses the return code in the response writer to determine if the response code should be overridden with NOERROR.

chrisohaver:
fixed in #771

PR #771, opened 2017-07-02
k8s/autopath: Add CNAMES

chrisohaver:
* I added unit tests for ServeDNS, and coverage for autopath queries that can be handled in the k8s middleware.  no coverage for queries that fall though to next middleware. #766 
* Autopath adds CNAME records to "bridge the gap" between header question and the A record.  #765
* Fixes the NXDOMAIN override to cause the client to stop traversing the search path when there is no result. (almost) #761
* The CNAME and NXDOMAIN items above are in part handled by a new ResponseWriter implementor: AutoPathWriter, which handles deferring writes to the outer middleware (k8s), conditionally rewriting the rcode, and restoring the header question. 

Work in Progress.  Some issues...
* disabling the NXDOMAIN feature doesn't work.  Its always enabled (not sure why yet).
* AAAA requests get A responses when passing them through the next middleware (proxy).
* currently only tries AutoPath for A and AAAA.  Need to test other types.
* Still no test coverage for queries that need to get answers from next middleware.

### DIFF
--- a/middleware/kubernetes/README.md
+++ b/middleware/kubernetes/README.md
@@ -185,7 +185,7 @@ specified).
  
   **NDOTS** (default: `0`) This provides an adjustable threshold to prevent server side lookups from triggering. If the number of dots before the first search domain is less than this number, then the search path will not executed on the server side.  When autopath is enabled with default settings, the search path is always conducted when the query is in the first search domain `<pod-namespace>.svc.<zone>.`.
 	
-  **RESPONSE** (default: `SERVFAIL`) This option causes the kubernetes middleware to return the given response instead of NXDOMAIN when the all searches in the path produce no results. Valid values: `NXDOMAIN`, `SERVFAIL` or `NOERROR`. Setting this to `SERVFAIL` or `NOERROR` should prevent the client from fruitlessly continuing the client side searches in the path after the server already checked them.
+  **RESPONSE** (default: `NOERROR`) This option causes the kubernetes middleware to return the given response instead of NXDOMAIN when the all searches in the path produce no results. Valid values: `NXDOMAIN`, `SERVFAIL` or `NOERROR`. Setting this to `SERVFAIL` or `NOERROR` should prevent the client from fruitlessly continuing the client side searches in the path after the server already checked them.
 
   **RESOLV-CONF** (default: `/etc/resolv.conf`) If specified, the kubernetes middleware uses this file to get the host's search domains. The kubernetes middleware performs a lookup on these domains if the in-cluster search domains in the path fail to produce an answer. If not specified, the values will be read from the local resolv.conf file (i.e the resolv.conf file in the pod containing CoreDNS).  In practice, this option should only need to be used if running CoreDNS outside of the cluster and the search path in /etc/resolv.conf does not match the cluster's "default" dns-policiy.
 
@@ -245,7 +245,7 @@ specified).
 		cidrs 10.0.0.0/24
 		pods verified
 		upstream 10.102.3.10:53
-		autopath 0 SERVFAIL cluster.conf
+		autopath 0 NOERROR cluster.conf
 	}
 
 
--- a/middleware/kubernetes/README.md
+++ b/middleware/kubernetes/README.md
@@ -168,7 +168,7 @@ specified).
     ;google.com.default.svc.cluster.local. IN A
 
     ;; ANSWER SECTION:
-    google.com.default.svc.cluster.local. 211 IN CNAME google.com.
+    google.com.default.svc.cluster.local. 175 IN CNAME google.com.
     google.com.		175	IN	A	216.58.194.206
   ```
 
@@ -179,9 +179,11 @@ specified).
   ```
 	# host -t a google.com
 	google.com has address 216.58.194.206
+	google.com.default.svc.cluster.local is an alias for google.com.
 	
 	# host -t a google.com.default.svc.cluster.local.
 	google.com has address 216.58.194.206
+	google.com.default.svc.cluster.local is an alias for google.com.
   ```
  
   **NDOTS** (default: `0`) This provides an adjustable threshold to prevent server side lookups from triggering. If the number of dots before the first search domain is less than this number, then the search path will not executed on the server side.  When autopath is enabled with default settings, the search path is always conducted when the query is in the first search domain `<pod-namespace>.svc.<zone>.`.
--- a/middleware/kubernetes/README.md
+++ b/middleware/kubernetes/README.md
@@ -156,7 +156,7 @@ specified).
 
   If no domains in the path produce an answer, a lookup on the bare question will be attempted.	
 
-  A successful response will contain a question section with the original question, and an answer section containing the record for the question that actually had an answer.  This means that the question and answer will not match. For example:
+  A successful response will contain a question section with the original question, and an answer section containing the record for the question that actually had an answer.  This means that the question and answer will not match. To avoid potential client confusion, a dynamically generated CNAME entry is added to join the two. For example:
 
   ```
     # host -v -t a google.com
@@ -168,6 +168,7 @@ specified).
     ;google.com.default.svc.cluster.local. IN A
 
     ;; ANSWER SECTION:
+    google.com.default.svc.cluster.local. 211 IN CNAME google.com.
     google.com.		175	IN	A	216.58.194.206
   ```
 
--- a/middleware/kubernetes/autopath.go
+++ b/middleware/kubernetes/autopath.go
@@ -55,7 +55,9 @@ func (apw *AutoPathWriter) overrideMsg(res *dns.Msg, force bool) error {
 		if apw.original.Name == a.Header().Name {
 			continue
 		}
-		res.Answer = append(res.Answer, newCNAME(apw.original.Name, dns.Fqdn(a.Header().Name), a.Header().Ttl))
+		res.Answer = append(res.Answer, nil)
+		copy(res.Answer[1:], res.Answer)
+		res.Answer[0] = newCNAME(apw.original.Name, dns.Fqdn(a.Header().Name), a.Header().Ttl)
 	}
 	res.Question[0] = apw.original
 	apw.Sent = true
--- a/middleware/kubernetes/autopath.go
+++ b/middleware/kubernetes/autopath.go
@@ -31,39 +31,33 @@ func NewAutoPathWriter(w dns.ResponseWriter, r *dns.Msg) *AutoPathWriter {
 	}
 }
 
+// WriteMsg writes to client, unless response will be NXDOMAIN
 func (apw *AutoPathWriter) WriteMsg(res *dns.Msg) error {
-	return apw.OverrideMsg(res, false)
+	return apw.overrideMsg(res, false)
 }
 
-// Force the write to client regardless of response code
+// ForceWriteMsg forces the write to client regardless of response code
 func (apw *AutoPathWriter) ForceWriteMsg(res *dns.Msg) error {
-	return apw.OverrideMsg(res, true)
+	return apw.overrideMsg(res, true)
 }
 
-// OverrideMsg records the status code and calls the
-// underlying ResponseWriter's WriteMsg method unless
-// the write is deferred, or force = true.
-func (apw *AutoPathWriter) OverrideMsg(res *dns.Msg, force bool) error {
+// overrideMsg overrides rcode, reverts question, adds CNAME, and calls the
+// underlying ResponseWriter's WriteMsg method unless the write is deferred,
+// or force = true.
+func (apw *AutoPathWriter) overrideMsg(res *dns.Msg, force bool) error {
 	if res.Rcode == dns.RcodeNameError {
 		res.Rcode = apw.Rcode
+	}
+	if res.Rcode != dns.RcodeSuccess && !force {
+		return nil
 	}
 	for _, a := range res.Answer {
 		if apw.original.Name == a.Header().Name {
 			continue
 		}
-		cname := dns.CNAME{
-			Hdr: dns.RR_Header{
-				Name:   apw.original.Name,
-				Rrtype: dns.TypeCNAME,
-				Class:  dns.ClassINET,
-				Ttl:    a.Header().Ttl},
-			Target: dns.Fqdn(a.Header().Name)}
-		res.Answer = append(res.Answer, &cname)
+		res.Answer = append(res.Answer, newCNAME(apw.original.Name, dns.Fqdn(a.Header().Name), a.Header().Ttl))
 	}
 	res.Question[0] = apw.original
-	if res.Rcode != dns.RcodeSuccess && !force {
-		return nil
-	}
 	apw.Sent = true
 	return apw.ResponseWriter.WriteMsg(res)
 }
--- a/middleware/kubernetes/autopath.go
+++ b/middleware/kubernetes/autopath.go
@@ -1,0 +1,77 @@
+package kubernetes
+
+import "github.com/miekg/dns"
+
+// AutoPathWriter implements a ResponseWriter that also does the following:
+// * reverts question section of a packet to its original state.
+//   This is done to avoid fatal 'Question section mismatch:' errors in client.
+// * Defers write to the client if the response code is NXDOMAIN.  This is needed
+//   to enable further search path probing if a search was not sucessful.
+// * Allow forced write to client regardless of response code.  This is needed to
+//   write the packet to the client if the final search in the path fails to
+//   produce results.
+// * Overwrites response code with AutoPathWriter.Rcode if the response code
+//   is NXDOMAIN (NameError).  This is needed to support the AutoPath.OnNXDOMAIN
+//   function, which returns a NOERROR to client instead of NXDOMAIN if the final
+//   search in the path fails to produce results.
+
+type AutoPathWriter struct {
+	dns.ResponseWriter
+	original dns.Question
+	Rcode    int
+}
+
+// NewAutoPathWriter returns a pointer to a new AutoPathWriter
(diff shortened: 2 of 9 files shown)


middleware/kubernetes/autopath.go
```diff
@@ -0,0 +1,79 @@
...
+			continue
+		}
+		cname := dns.CNAME{
+			Hdr: dns.RR_Header{
+				Name:   r.original.Name,
+				Rrtype: dns.TypeCNAME,
+				Class:  dns.ClassINET,
+				Ttl:    a.Header().Ttl},
+			Target: dns.Fqdn(a.Header().Name)}
+		res.Answer = append(res.Answer, &cname)
```
miekg: Does appending, while ranging actually work

middleware/kubernetes/handler.go
```diff
@@ -40,10 +38,8 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 		// Set the zone to this specific request.
 		zone = state.Name()
 	}
-
 	records, extra, _, err := k.routeRequest(zone, state)
-
-	if k.AutoPath.Enabled && k.IsNameError(err) {
+	if k.AutoPath.Enabled && k.IsNameError(err) && (state.QType() == dns.TypeA || state.QType() == dns.TypeAAAA) {
 		p := k.findPodWithIP(state.IP())
 		for p != nil {
```
miekg: `if p == nil { continue}`

middleware/kubernetes/handler.go
```diff
@@ -53,38 +49,46 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
...
 			for i := 0; i < 2; i++ {
 				path = strings.Join(dns.SplitDomainName(path)[1:], ".")
 				state = state.NewWithQuestion(strings.Join([]string{name, path}, "."), state.QType())
 				records, extra, _, err = k.routeRequest(zone, state)
 				if !k.IsNameError(err) {
+					records = append(records, NewCNAME(origQName, records[0].Header().Name, records[0].Header().Ttl))
 					break
 				}
 			}
 			if !k.IsNameError(err) {
```
miekg: Why is this IsNameErrror check repeated outside the `for`-loop as well?

middleware/kubernetes/handler.go
```diff
@@ -53,38 +49,46 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 			if (dns.CountLabel(name) - 1) < k.AutoPath.NDots {
 				break
 			}
-			// Search "svc.cluster.local" and "cluster.local"
+			origQName := state.QName()
+			// Search "svc.cluster.local." and "cluster.local."
 			for i := 0; i < 2; i++ {
```
miekg, line 56: Why is this looping twice over the same thing? Unless my eyes are deceiving me?

chrisohaver:
The range is established once, before the loop starts. Appended items don't get ranged over, since they are outside the initial range.

Sent from my mobile

On Jul 5, 2017, at 8:46 AM, Miek Gieben <notifications@github.com<mailto:notifications@github.com>> wrote:


@miekg commented on this pull request.

In middleware/kubernetes/autopath.go<https://github.com/coredns/coredns/pull/771#discussion_r125632310>:

> +func (r *AutoPathWriter) WriteMsg(res *dns.Msg) error {
+       if res.Rcode == dns.RcodeNameError {
+               res.Rcode = r.Rcode
+       }
+       for _, a := range res.Answer {
+               if r.original.Name == a.Header().Name {
+                       continue
+               }
+               cname := dns.CNAME{
+                       Hdr: dns.RR_Header{
+                               Name:   r.original.Name,
+                               Rrtype: dns.TypeCNAME,
+                               Class:  dns.ClassINET,
+                               Ttl:    a.Header().Ttl},
+                       Target: dns.Fqdn(a.Header().Name)}
+               res.Answer = append(res.Answer, &cname)


Does appending, while ranging actually work

middleware/kubernetes/handler.go
```diff
@@ -53,38 +49,46 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
...
 			}
-			// Fallthrough with the host search path (if set)
-			wr := rewrite.NewResponseReverter(w, r)
+			// Try host search path (if set) in the next middleware
+			apw := NewAutoPathWriter(w, r)
 			for _, hostsearch := range k.AutoPath.HostSearchPath {
-				r = state.NewWithQuestion(strings.Join([]string{name, hostsearch}, "."), state.QType()).Req
-				rcode, nextErr := middleware.NextOrFailure(k.Name(), k.Next, ctx, wr, r)
-				if rcode == dns.RcodeSuccess {
+				state = state.NewWithQuestion(strings.Join([]string{name, hostsearch}, "."), state.QType())
```
miekg: We might want to have a dnsutil.Join function for these?

middleware/kubernetes/handler.go
chrisohaver: Not sure I follow this comment.

middleware/kubernetes/autopath.go
chrisohaver: The way it behaves works for what we want to do... the records appended in the loop are not ranged over.

middleware/kubernetes/handler.go
chrisohaver: it strips the first segment from path each time through.
The original path is ... `w.x.y.z`, then the first loop does `x.y.z`, and the second does `y.z`

middleware/kubernetes/handler.go
chrisohaver: if you mean to replace the `for p != nil {`  I don't think it would work. `continue` I thought only works inside a loop.

middleware/kubernetes/handler.go
chrisohaver: This breaks out of the outer loop (`for p!= nil`).
I could label the outer loop, and break out from the inner loop line 60 if you prefer.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: k8s/auto..." ]
>>  			for i := 0; i < 2; i++ {
> 				path = strings.Join(dns.SplitDomainName(path)[1:], ".")
> 				state = state.NewWithQuestion(strings.Join([]string{name, path}, "."), state.QType())
> 				records, extra, _, err = k.routeRequest(zone, state)
> 				if !k.IsNameError(err) {
>+					records = append(records, NewCNAME(origQName, records[0].Header().Name, records[0].Header().Ttl))
> 					break
> 				}
> 			}
> 			if !k.IsNameError(err) {
>
>This breaks out of the outer loop (`for p!= nil`).
>I could label the outer loop, and break out from the inner loop line 60 if you prefer.

OK, just leave it like this then.


/Miek

--
Miek Gieben

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: k8s/auto..." ]
>
>
>
>>  	records, extra, _, err := k.routeRequest(zone, state)
>-
>-	if k.AutoPath.Enabled && k.IsNameError(err) {
>+	if k.AutoPath.Enabled && k.IsNameError(err) && (state.QType() == dns.TypeA || state.QType() == dns.TypeAAAA) {
> 		p := k.findPodWithIP(state.IP())
> 		for p != nil {
>
>if you mean to replace the `for p != nil {`  I don't think it would work. `continue` I thought only works inside a loop.

True... I was just wondering if there was an "early out" here, because all the 
work happens in the if-body. Maybe not.

middleware/kubernetes/handler.go
miekg: On what line in that loop is `i` used?

chrisohaver: `i` is only used in the loop condition.

chrisohaver:
I've fixed the remaining bigger issues...
* OnNXDOMAIN fixed, unit test added.
* AAAA/A response switch fixed, unit test added.
* Added coverage for next middleware cases .

Only A and AAAA can take advantage of the feature for now.  I think we should leave that as is, and add other types later.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] k8s/autopath:..." ]
>I've fixed the remaining bigger issues...
>* OnNXDOMAIN fixed, unit test added.
>* AAAA/A response switch fixed, unit test added.
>* Added coverage for next middleware cases .
>
>Only A and AAAA can take advantage of the feature for now.  I think we should leave that as is, and add other types later.

Seems legit, that's probably 90% of the queries anyway.

middleware/kubernetes/README.md
```diff
@@ -168,6 +168,7 @@ specified).
     ;google.com.default.svc.cluster.local. IN A
 
     ;; ANSWER SECTION:
+    google.com.default.svc.cluster.local. 211 IN CNAME google.com.
```
miekg: Where does the TTL come from?

middleware/kubernetes/handler.go
miekg: I still don't see it :-( Where is the path changed for the 2nd run?

middleware/kubernetes/autopath.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Force the write to client regardless of response code
+func (apw *AutoPathWriter) ForceWriteMsg(res *dns.Msg) error {
+	return apw.OverrideMsg(res, true)
+}
+
+// OverrideMsg records the status code and calls the
+// underlying ResponseWriter's WriteMsg method unless
+// the write is deferred, or force = true.
+func (apw *AutoPathWriter) OverrideMsg(res *dns.Msg, force bool) error {
```
miekg: It is probably cleaner to have two separately named functions that do the 2 things the bool differentiates between. Like you have before

middleware/kubernetes/kubernetes.go
```diff
@@ -57,6 +57,10 @@ type AutoPath struct {
 	OnNXDOMAIN     int
 }
 
+type Proxyer interface {
```
miekg: Why the new interface?

middleware/kubernetes/handler.go
```diff
@@ -115,6 +121,10 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 	return dns.RcodeSuccess, nil
 }
 
+func NewCNAME(name string, target string, ttl uint32) *dns.CNAME {
```
miekg: You could use this in `OverrideMsg` as well. But this function duplicates some of the things that already exists in `etcd/msg/service.go` where `NewXXX` is defined for a bunch a RR types. Sad thing is they are methods on *Service. Probably needs to be factor out and put into `dnsutil`. Can you add a  TODO here that this needs to be done?

Also the function does not need to be expored

middleware/test/helpers.go
```diff
@@ -279,6 +279,25 @@ func NextHandler(rcode int, err error) Handler {
 	})
 }
 
+// MockHandler returns a Handler that returns an answer for the quesion in the
+// request per the question->answer map qMap.
+func MockHandler(qMap map[dns.Question]dns.Msg) Handler {
```
miekg: Please move this to `kubernetes/handler_test.go` and call it "autoPathHandler" or "testHandler"?

request/request.go
```diff
@@ -29,7 +29,7 @@ type Request struct {
 // section in the request.
 func (r *Request) NewWithQuestion(name string, typ uint16) Request {
 	req1 := Request{W: r.W, Req: r.Req.Copy()}
-	req1.Req.Question[0] = dns.Question{Name: dns.Fqdn(name), Qtype: dns.ClassINET, Qclass: typ}
+	req1.Req.Question[0] = dns.Question{Name: dns.Fqdn(name), Qclass: dns.ClassINET, Qtype: typ}
```
miekg, line 32: Ouch

middleware/kubernetes/README.md
chrisohaver: It's copied from the A record.

middleware/kubernetes/kubernetes.go
chrisohaver: ah - I was starting to mock out the upstream lookups so those cases could be tested.

I can remove it - its not directly related to this feature.

request/request.go
chrisohaver: This was the cause of the A/AAAA swap.
Class `IN` uses the same numerical value as type `A`. So when an `AAAA` record came in, this function was turning it into a `A` record with an (invalid) class `28` ... clients seem to ignore the invalid class.

middleware/kubernetes/handler.go
chrisohaver: first line in the loop does, effectively `path = path minus first segment` ... 
```
path = strings.Join(dns.SplitDomainName(path)[1:], ".")
```

middleware/kubernetes/autopath.go
chrisohaver: There is WriteMsg, and ForceWriteMsg.  The first calls OverrideMsg with force=false. the second calls OverrideMsg with force=true.

I should unexpose OverrideMsg, might make is clearer.

middleware/kubernetes/autopath.go
```diff
@@ -0,0 +1,82 @@
...
+			Hdr: dns.RR_Header{
+				Name:   apw.original.Name,
+				Rrtype: dns.TypeCNAME,
+				Class:  dns.ClassINET,
+				Ttl:    a.Header().Ttl},
+			Target: dns.Fqdn(a.Header().Name)}
+		res.Answer = append(res.Answer, &cname)
+	}
+	res.Question[0] = apw.original
+	if res.Rcode != dns.RcodeSuccess && !force {
```
chrisohaver: I should move this check before the for loop.

middleware/test/helpers.go
chrisohaver: I considered it to have a general purpose - not specific to autopath or kubernetes.

miekg:
Ah ok. True looks useful. Better name than Mock would help then :)

On Jul 7, 2017 2:46 PM, "chrisohaver" <notifications@github.com> wrote:

*@chrisohaver* commented on this pull request.
------------------------------

In middleware/test/helpers.go
<https://github.com/coredns/coredns/pull/771#discussion_r126150071>:

> @@ -279,6 +279,25 @@ func NextHandler(rcode int, err error) Handler {
 	})
 }

+// MockHandler returns a Handler that returns an answer for the quesion in the
+// request per the question->answer map qMap.
+func MockHandler(qMap map[dns.Question]dns.Msg) Handler {

I considered it to have a general purpose - not specific to autopath or
kubernetes.

—
You are receiving this because you were mentioned.
<https://github.com/coredns/coredns/pull/771#discussion_r126150071>, or mute
the thread

middleware/kubernetes/handler.go
miekg: ah, sorry. yes.

middleware/kubernetes/README.md
miekg: Why is the TTL on the A record diff. then?

miekg (approved):

middleware/kubernetes/README.md
chrisohaver: Ah - good eye... I'll fix the README.

miekg:
The more I play with this code the more I would like see this in a separate middleware (if possible). Lots of complexity here that makes the k8s middleware even more complex than it already it.

I should have a PR ready soonish that tries to do this in the *autopath* middleware.

johnbelamaric:
@miekg you want to hold off on the merge? I was thinking this was holding 009.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] k8s/autopath:..." ]
>@miekg you want to hold off on the merge? I was thinking this was holding 009.

Yes please. If the middleware *autopath* actually works, we might want to revert 
a few commits even...

middleware/kubernetes/handler.go
```diff
@@ -53,38 +51,46 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
...
 					return rcode, nextErr
 				}
 			}
 			// Search . in this middleware
-			state = state.NewWithQuestion(strings.Join([]string{name, "."}, ""), state.QType())
-			records, extra, _, err = k.routeRequest(zone, state)
+			newstate := state.NewWithQuestion(strings.Join([]string{name, "."}, ""), state.QType())
+			records, extra, _, err = k.routeRequest(zone, newstate)
 			if !k.IsNameError(err) {
+				records = append(records, newCNAME(origQName, records[0].Header().Name, records[0].Header().Ttl))
```
miekg: you need a put the CNAME first and then the original A record otherwise glibc won't understand it.

middleware/kubernetes/autopath.go
```diff
@@ -55,7 +55,9 @@ func (apw *AutoPathWriter) overrideMsg(res *dns.Msg, force bool) error {
 		if apw.original.Name == a.Header().Name {
 			continue
 		}
-		res.Answer = append(res.Answer, newCNAME(apw.original.Name, dns.Fqdn(a.Header().Name), a.Header().Ttl))
+		res.Answer = append(res.Answer, nil)
```
johnbelamaric: Maybe this instead: https://codingair.wordpress.com/2014/07/18/go-appendprepend-item-into-slice/

?

chrisohaver: I actually googled up the same link.  But after reading comment ...

```
A better prepend, as it generates less garbage:

data = append(data, “”)
copy(data[1:], data)
data[0] = “Prepend Item”
```

I opted for the less garbagey method.  I did not verify if it produces less garbage, but it looks like it does.

PR #774, opened 2017-07-07
mw/k8s: remove unused code

chrisohaver:
I found that kubernetes/lookup.go is actually not used by anything.  Functionally redundant, I think left over from early k8s middleware work.


chrisohaver:
>--- FAIL: TestReload (2.00s)
	reload_test.go:42: Could not send message: read udp [::1]:55711->[::1]:52121: i/o timeout

Any ideas?  Does this fail intermittently?

PR #775, opened 2017-07-11
middleware/autopath: new

miekg:
Abstract the autopath functionality into it's own middleware. This does server side search path expansion.

Note: there is still a lot to copy, I'm trying to use git for most things to keep the attribution correct.

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -204,7 +204,7 @@ func (s *Server) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 		if h, ok := s.zones[string(b[:l])]; ok {
 			if r.Question[0].Qtype != dns.TypeDS {
 				rcode, _ := h.middlewareChain.ServeDNS(ctx, w, r)
-				if rcodeNoClientWrite(rcode) {
+				if !middleware.ClientWriteDone(rcode) {
 					DefaultErrorFunc(w, r, rcode)
 				}
 				return
@@ -225,7 +225,7 @@ func (s *Server) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 	if dshandler != nil {
 		// DS request, and we found a zone, use the handler for the query
 		rcode, _ := dshandler.middlewareChain.ServeDNS(ctx, w, r)
-		if rcodeNoClientWrite(rcode) {
+		if !middleware.ClientWriteDone(rcode) {
 			DefaultErrorFunc(w, r, rcode)
 		}
 		return
@@ -234,7 +234,7 @@ func (s *Server) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 	// Wildcard match, if we have found nothing try the root zone as a last resort.
 	if h, ok := s.zones["."]; ok {
 		rcode, _ := h.middlewareChain.ServeDNS(ctx, w, r)
-		if rcodeNoClientWrite(rcode) {
+		if !middleware.ClientWriteDone(rcode) {
 			DefaultErrorFunc(w, r, rcode)
 		}
 		return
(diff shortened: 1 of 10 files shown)


miekg:
It's not working and no tests, but it should give a general feel of having this functionality live outside the kubernetes middeware
/cc @chrisohaver @johnbelamaric

johnbelamaric:
This can't do the same as the K8s auto path. The K8s auto path needs to look up the search path based on the *client source IP*. Each pod will have a search path that depends on the namespace in which the pod lives.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/au..." ]
>This can't do the same as the K8s auto path. The K8s auto path needs to look up the search path based on the *client source IP*. Each pod will have a search path that depends on the namespace in which the pod lives.

Damn. But that might still be doable, as it is just a way to get the searchpath 
for this request. Would be nice if there was a way to call into middleware - but 
that has complication as well. But that gets ugly as well :(

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/au..." ]
>This can't do the same as the K8s auto path. The K8s auto path needs to look up the search path based on the *client source IP*. Each pod will have a search path that depends on the namespace in which the pod lives.

But I "somehow" get this list, it should work fine, no?

johnbelamaric:
Yes, I think so. @chrisohaver ?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/au..." ]
>Yes, I think so. @chrisohaver ?

Then I have an even whackier idea: Implement this is a standalone middleware 
(which nobody is allowed to use? dunno). Wrap the kubernetes middleware in the
autopath middleware before we instantiate the kubernetes middleware. Add some 
special code that retrieve the domain search list from k8s before calling the 
ServeDNS from kubernetes. This should mean, all this special casing is kept
outside the kubernetes middleware and we may spin this off as another 
middleware; it also makes the testing easier (as said before).

chrisohaver:
Conceptually perhaps. Less efficient though, since all incoming queries are checked for inclusion in every domain in the path.

chrisohaver:
the goal is to make this _less_ complex, right?

chrisohaver:
Can you expand on how this would make testing easier? (testing of what, and how)

johnbelamaric:
Here's another idea. I think we could proceed with what we have for 009. It's a working PR. We can then look at how this fits in back with the policy piece. Not to pull another thing in, but there is some real overlap here.

Today, the policy integration at https://github.com/infobloxopen/themis/tree/master/contrib/coredns/policy is focused on looking at individual queries and deciding to allow, block, or redirect them. It can use the basic query data or edns0 data to make those decisions. We ALSO have plans to eventually allow an integration with K8s to allow K8s meta-data (for example, labels on the pod making the query) to be used in these kind of policy decisions. The search path is really just another piece of meta-data in this case. So, as we move forward on enabling the gathering of that meta-data for policy, it could also be re-used in an auto path middleware.

In a kind-of-sort-of related idea, we could actually use the policy stuff to determine not just whether to allow, block, or redirect (to another IP) a query, we could actually allow it to determine the downstream middleware chain. Sort of a programmable "fall through".

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/au..." ]
>the goal is to make this _less_ complex right?

This would mean no code changes in the kubernetes middleware. 

Re: testing: stand alone middleware: write Corefile, fire up CoreDNS, send 
queries; no need for k8s at all.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/au..." ]
>Here's another idea. I think we could proceed with what we have for 009. It's a working PR. We can then look at how this fits in back with the policy piece. Not to pull another thing in, but there is some real overlap here.
>
>Today, the policy integration at https://github.com/infobloxopen/themis/tree/master/contrib/coredns/policy is focused on looking at individual queries and deciding to allow, block, or redirect them. It can use the basic query data or edns0 data to make those decisions. We ALSO have plans to eventually allow an integration with K8s to allow K8s meta-data (for example, labels on the pod making the query) to be used in these kind of policy decisions. The search path is really just another piece of meta-data in this case. So, as we move forward on enabling the gathering of that meta-data for policy, it could also be re-used in an auto path middleware.
>
>In a kind-of-sort-of related idea, we could actually use the policy stuff to determine not just whether to allow, block, or redirect (to another IP) a query, we could actually allow it to determine the downstream middleware chain. Sort of a programmable "fall through".

that's a very nice idea, but we do have to be a bit careful not to reinvent a 
mini-coredns in each middleware :-)

I agree with 009, let's release. I'll whack a bit on this PR to see it pans out.

PR #780, opened 2017-07-19
middleware/kubernetes: fix aaaa response

### LINKED ISSUE #779: middleware/kubernetes: AAAA NXDOMAIN response
In k8s middleware `AAAA` queries always return an `NXDOMAIN` response.  The response should instead be conditional...

- If a service exists with that name, but there is not a `AAAA` record, the return should be a `NOERROR` with no answers.  `AAAA` is not supported in K8s yet, so currently there will never be a `AAAA` record.

- If a service exists doesn't exist with that name, the return should be a `NXDOMAIN`.

chrisohaver:
Fixes #779 ...

While testing this fix, we ran into what looks like the same bug in hosts middleware. I'll open an issue.

### DIFF
--- a/middleware/kubernetes/handler_test.go
+++ b/middleware/kubernetes/handler_test.go
@@ -81,6 +81,22 @@ var dnsTestCases = map[string](*test.Case){
 			test.CNAME("svc0.testns.fed.svc.cluster.local.	0	IN	CNAME	svc0.testns.fed.svc.fd-az.fd-r.federal.test."),
 		},
 	},
+	"AAAA Service (existing service)": {
+		Qname: "svc1.testns.svc.cluster.local.", Qtype: dns.TypeAAAA,
+		Rcode:  dns.RcodeSuccess,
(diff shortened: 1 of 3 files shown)


pmoroney:
I'll start on fixing the hosts middleware

chrisohaver:
> I'll start on fixing the hosts middleware

Thanks! #781

middleware/kubernetes/handler_test.go
```diff
@@ -81,6 +81,22 @@ var dnsTestCases = map[string](*test.Case){
 			test.CNAME("svc0.testns.fed.svc.cluster.local.	0	IN	CNAME	svc0.testns.fed.svc.fd-az.fd-r.federal.test."),
 		},
 	},
+	"AAAA Service (existing service)": {
```
miekg, line 84: Are we doing the correct thing for other types as well? i.e. MX lookups or other things that might get queried?

miekg:
lgtm

Issue #781, opened 2017-07-19
middleware/hosts: AAAA responses


chrisohaver:
For a name that has an `A` record, but not an `AAAA` record, If the name is queried for the `AAAA` record, hosts middleware should return a `NOERROR` with 0 answers.  Currently it returns `NXDOMAIN`, even if there exists an `A` record.

miekg:
as said in the other code review, should also double check other types. Does MX give back nodata as well?

pmoroney:
Good point, I'll verify

chrisohaver:
Can we close this one?

pmoroney:
Yes, #782 fixes this and has been merged.

PR #782, opened 2017-07-19
hosts middleware should return NoError if other records exist in the zone

pmoroney:
Addresses https://github.com/coredns/coredns/issues/781

### DIFF
--- a/middleware/hosts/hosts.go
+++ b/middleware/hosts/hosts.go
@@ -60,7 +60,7 @@ func (h Hosts) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 			return middleware.NextOrFailure(h.Name(), h.Next, ctx, w, r)
 		}
 		if !h.otherRecordsExist(state.QType(), qname) {
-			return dns.RcodeRefused, nil
+			return dns.RcodeNameError, nil
 		}
 	}
 
--- a/middleware/hosts/hosts.go
+++ b/middleware/hosts/hosts.go
@@ -39,6 +39,7 @@ func (h Hosts) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 		}
 	}
 
+	var otherRecordCount int
 	switch state.QType() {
 	case dns.TypePTR:
 		names := h.LookupStaticAddr(dnsutil.ExtractAddressFromReverse(qname))
@@ -49,9 +50,11 @@ func (h Hosts) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 		answers = h.ptr(qname, names)
 	case dns.TypeA:
 		ips := h.LookupStaticHostV4(qname)
+		otherRecordCount = len(h.LookupStaticHostV6(qname))
 		answers = a(qname, ips)
 	case dns.TypeAAAA:
 		ips := h.LookupStaticHostV6(qname)
+		otherRecordCount = len(h.LookupStaticHostV4(qname))
 		answers = aaaa(qname, ips)
 	}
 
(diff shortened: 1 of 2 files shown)


middleware/hosts/hosts.go
```diff
@@ -49,17 +50,21 @@ func (h Hosts) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
...
 		answers = aaaa(qname, ips)
 	}
 
 	if len(answers) == 0 {
 		if h.Fallthrough {
 			return middleware.NextOrFailure(h.Name(), h.Next, ctx, w, r)
 		}
-		return dns.RcodeRefused, nil
+		if otherRecordCount == 0 {
+			return dns.RcodeRefused, nil
```
chrisohaver: I realize the response code hasn't changed, but why `refused` instead of `nxdomain` when no records exist?

pmoroney: The hosts middleware is more like the auto middleware, in that the origin is generally defined as "."
We wanted to return a REFUSED instead of an NXDOMAIN since we don't want to say we are authoritative for that zone and we don't necessarily want it cached.
I am not necessarily married to that decision though, and if it will be better to return an NXDOMAIN I can go ahead and change that.

pmoroney:
hmm, why didn't this do the coverage check?

middleware/hosts/hosts.go
chrisohaver: OK, NXDOMAIN is only supposed to be used for authoritative answers... so makes sense.

chrisohaver:
LGTM.  One nit, you could defer the "other" lookup/count to after the record lookup, and only do it if answers == 0 (e.g. only do it when we need to).

pmoroney:
I'm actually going to modify it some more so I can get it to work for other record types.

johnbelamaric:
I assume Refused will trigger search list navigation just like NXDOMAIN?

miekg:
This is the code in gnu libc: http://sourceware.org/git/?p=glibc.git;a=blob;f=resolv/res_query.c;h=ebbe5a6a4ed86abe3fccd4a134bfcf6f613c9bbb;hb=HEAD#l436

chrisohaver:
I believe John is referring to the search path traversal behavior on the client (search path from resolv.conf).
If the client searches for `foo`, and the search path is `bar.com baz.com` ...
if hosts is configured to serve zone `bar.com`, and doesnt have an entry for `foo` ... will the `REFUSED` response cause the client to give up, or will the client continue to search for `foo` in the next search domain - `foo.baz.com`

johnbelamaric:
So based on that @miekg, it looks like it will NOT trigger search list. I would think we would want to in this case. That means either success with no data or NXDOMAIN, if I am reading the code right. Although I don't remember seeing that behavior with "no data".

chrisohaver:
I just verified by test that `REFUSED` halts the path search on the client, at least with the "host" command...
for search path = `testdom testdom2`

```
# host -v -ta test3
Trying "test3.testdom"
Host test3.testdom not found: 5(REFUSED)
```

it doesnt try `test3.testdom2`

pmoroney:
ok, I'll switch it to NXDOMAIN then :)

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] hosts middlew..." ]
>ok, I'll switch it to NXDOMAIN then :)

But NXDOMAIN is the wrong rcode:
https://tools.ietf.org/html/rfc8020

A Nodata response is better from a DNS perspective. The retry behavior also 
depends on the libc in the client; i.e. what does musl do?

/Miek

--
Miek Gieben

miekg:
Or maybe it *is* correct. I mean, *hosts* only knows what it knows.

johnbelamaric:
I have been playing with this, and I think how it is now works for most cases. Suppose your `hosts` file has an IPv4 address for `example.com`. If you as for an A record, it will return it. If you ask for AAAA or MX it will return a nodata response. However, if you enable `fallthrough`, instead of returning nodata, it will trigger the fall through, which will either return a real value or will eventually give an `NXDOMAIN` and trigger the search path.

miekg:
yes that makes sense, and mimics my (incoherent) thoughts. I'm (still) opposed to implicit **fallthrough** but would should probably mention this in the README.

chrisohaver:
It seems that any non-authoritative middleware should implicitly fall through.

PR #786, opened 2017-07-24
middleware/proxy: dnstap

2buisson:
Based on https://github.com/coredns/coredns/pull/711, this PR enables other middlewares to log their own events to dnstap.

### DIFF
--- a/middleware/dnstap/handler.go
+++ b/middleware/dnstap/handler.go
@@ -33,6 +33,10 @@ func (h Dnstap) TapMessage(m *tap.Message) error {
 	return tapMessageTo(h.Out, m)
 }
 
+func (h Dnstap) IncludeBinary() bool {
+	return h.Pack
+}
+
 func (h Dnstap) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 	rw := &taprw.ResponseWriter{ResponseWriter: w, Taper: &h, Query: r, Pack: h.Pack}
 	rw.QueryEpoch()
--- a/middleware/dnstap/handler.go
+++ b/middleware/dnstap/handler.go
@@ -34,13 +34,7 @@ func (h Dnstap) TapMessage(m *tap.Message) error {
 }
 
 func (h Dnstap) Tap(b *msg.Builder) error {
-	if h.Pack {
-		if err := b.IncludeBinary(); err != nil {
-			return fmt.Errorf("pack: %s", err)
-		}
-	}
-
-	m, err := b.Build()
+	m, err := b.Build(h.Pack)
 	if err != nil {
 		return err
 	}
--- a/middleware/dnstap/handler.go
+++ b/middleware/dnstap/handler.go
@@ -13,6 +13,7 @@ import (
 	"golang.org/x/net/context"
 )
 
+// Dnstap is the dnstap handler.
 type Dnstap struct {
 	Next middleware.Handler
 	Out  io.Writer
@@ -22,7 +23,8 @@ type Dnstap struct {
 type (
 	// Taper is implemented by the Context passed by the dnstap handler.
 	Taper interface {
-		Tap(*msg.Builder) error
+		TapMessage(*tap.Message) error
+		TapBuilder() msg.Builder
 	}
 	tapContext struct {
 		context.Context
@@ -45,20 +47,19 @@ func tapMessageTo(w io.Writer, m *tap.Message) error {
 	return err
 }
 
+// TapMessage implements Taper.
 func (h Dnstap) TapMessage(m *tap.Message) error {
 	return tapMessageTo(h.Out, m)
 }
 
-func (h Dnstap) Tap(b *msg.Builder) error {
-	m, err := b.Build(h.Pack)
-	if err != nil {
-		return err
-	}
-	return tapMessageTo(h.Out, m)
+// TapBuilder implements Taper.
+func (h Dnstap) TapBuilder() msg.Builder {
+	return msg.Builder{Full: h.Pack}
 }
 
+// ServeDNS logs the client query and response to dnstap and passes the dnstap Context.
 func (h Dnstap) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
-	rw := &taprw.ResponseWriter{ResponseWriter: w, Taper: &h, Query: r, Pack: h.Pack}
+	rw := &taprw.ResponseWriter{ResponseWriter: w, Taper: &h, Query: r}
 	rw.QueryEpoch()
 
 	code, err := middleware.NextOrFailure(h.Name(), h.Next, tapContext{ctx, h}, rw, r)
@@ -73,4 +74,6 @@ func (h Dnstap) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 
 	return code, nil
 }
+
+// Name returns dnstap.
 func (h Dnstap) Name() string { return "dnstap" }
--- a/middleware/dnstap/handler.go
+++ b/middleware/dnstap/handler.go
@@ -4,20 +4,36 @@ import (
 	"fmt"
 	"io"
 
-	"golang.org/x/net/context"
-
 	"github.com/coredns/coredns/middleware"
 	"github.com/coredns/coredns/middleware/dnstap/msg"
 	"github.com/coredns/coredns/middleware/dnstap/taprw"
 
 	tap "github.com/dnstap/golang-dnstap"
 	"github.com/miekg/dns"
+	"golang.org/x/net/context"
 )
 
 type Dnstap struct {
 	Next middleware.Handler
 	Out  io.Writer
 	Pack bool
+}
+
+type (
+	// Taper is implemented by the Context passed by the dnstap handler.
+	Taper interface {
+		Tap(*msg.Builder) error
+	}
+	tapContext struct {
+		context.Context
+		Dnstap
+	}
+)
+
+// TaperFromContext will return a Taper if the dnstap middleware is enabled.
+func TaperFromContext(ctx context.Context) (t Taper) {
+	t, _ = ctx.(Taper)
+	return
 }
 
 func tapMessageTo(w io.Writer, m *tap.Message) error {
@@ -45,7 +61,7 @@ func (h Dnstap) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 	rw := &taprw.ResponseWriter{ResponseWriter: w, Taper: &h, Query: r, Pack: h.Pack}
 	rw.QueryEpoch()
 
-	code, err := middleware.NextOrFailure(h.Name(), h.Next, ctx, rw, r)
+	code, err := middleware.NextOrFailure(h.Name(), h.Next, tapContext{ctx, h}, rw, r)
 	if err != nil {
 		// ignore dnstap errors
 		return code, err
--- a/middleware/dnstap/handler.go
+++ b/middleware/dnstap/handler.go
@@ -33,8 +33,18 @@ func (h Dnstap) TapMessage(m *tap.Message) error {
 	return tapMessageTo(h.Out, m)
 }
 
-func (h Dnstap) IncludeBinary() bool {
-	return h.Pack
+func (h Dnstap) Tap(b *msg.Builder) error {
+	if h.Pack {
+		if err := b.IncludeBinary(); err != nil {
+			return fmt.Errorf("pack: %s", err)
+		}
+	}
+
+	m, err := b.Build()
+	if err != nil {
+		return err
+	}
+	return tapMessageTo(h.Out, m)
 }
 
 func (h Dnstap) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
--- a/middleware/dnstap/handler.go
+++ b/middleware/dnstap/handler.go
@@ -21,8 +21,8 @@ type Dnstap struct {
 }
 
 type (
-	// Taper is implemented by the Context passed by the dnstap handler.
-	Taper interface {
+	// Tapper is implemented by the Context passed by the dnstap handler.
+	Tapper interface {
 		TapMessage(*tap.Message) error
 		TapBuilder() msg.Builder
 	}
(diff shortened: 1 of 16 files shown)


miekg:
We should probably merged #711, because it looks like got all those changes in here again?

johnbelamaric:
Yeah, it will need rebase after we merge that. I think there are one or two little things remaining there.

middleware/proxy/dns.go
```diff
@@ -43,6 +61,12 @@ func (d *dnsEx) Exchange(ctx context.Context, addr string, state request.Request
 		return nil, err
 	}
 
+	// log forwarded query to dnstap
+	taperr := d.tap(co, state.Req, true)
+	if taperr != nil {
+		taperr = fmt.Errorf("dnstap: %s", err)
```
miekg: Why don't you return here?

middleware/dnstap/msg/msg.go
```diff
@@ -87,3 +95,29 @@ func (d *Data) ToClientQuery() *tap.Message {
 		QueryPort:      &d.Port,
 	}
 }
+
+// Transform the data into a forwader or resolver query message.
```
miekg: Does this need to be exported? Also use `go vet` and `golint` to find lint violation about functions docs.

middleware/dnstap/msg/msg.go
```diff
@@ -24,8 +24,12 @@ type Data struct {
 	TimeSec     uint64
 }
 
-func (d *Data) FromRequest(r request.Request) error {
-	switch addr := r.W.RemoteAddr().(type) {
+type Conn interface {
```
miekg: Why does do we need this interface?

middleware/proxy/dns.go
2buisson: I don't return so the dnstap error won't stop the exchange process.

middleware/dnstap/msg/msg.go
2buisson: Yes! `msg` is a helper package, `ToOutsideQuery` may be useful for other middlewares.

middleware/dnstap/msg/msg.go
2buisson: `dns.ResponseWriter` (client messages) and `net.Conn` (proxy messages) both satisfy it.

middleware/dnstap/msg/msg.go
miekg: If it really usefll we should probably put it somewhere under middleware/pkg/...

middleware/dnstap/msg/msg.go
miekg: Ah yes. I see.

middleware/proxy/dns.go
```diff
@@ -32,7 +35,22 @@ func (d *dnsEx) Protocol() string          { return "dns" }
 func (d *dnsEx) OnShutdown(p *Proxy) error { return nil }
 func (d *dnsEx) OnStartup(p *Proxy) error  { return nil }
 
+func (d *dnsEx) tap(co net.Conn, m *dns.Msg, isQuery bool) error {
+	if d.Taper == nil {
+		// nothing to tap to
+		return nil
+	}
+	dat := msg.Data{}
+	dat.Epoch()
```
miekg: dat.Epoch() is weird. The Epoch() function should just return the `uint64`
~~~
dat := msg.Data{TimeSec: Epoch()}
~~~
Or make a `new.NewData()` is TimeSec is always set.

middleware/proxy/dns.go
miekg: But that's not true because `taperr` is returned at the end of the function.

middleware/proxy/proxy.go
```diff
@@ -21,6 +25,10 @@ var (
 	errInvalidDomain   = errors.New("invalid path for proxy")
 )
 
+type Taper interface {
```
miekg: Doesn't dnstap already define this interface?

middleware/proxy/lookup.go
```diff
@@ -108,11 +108,11 @@ func (p Proxy) lookup(state request.Request) (*dns.Msg, error) {
 
 			atomic.AddInt64(&host.Conns, 1)
 
-			reply, backendErr := upstream.Exchanger().Exchange(context.TODO(), host.Name, state)
+			reply, _ := upstream.Exchanger().Exchange(context.TODO(), host.Name, state)
```
miekg: Wait, why are you ignore this error?

miekg (changes requested):
I have a feeling we should probably do this similar to a tracing is done, which is mostly implemented in the ServeDNS method and seems more contained. @johnbelamaric what do you think?

middleware/proxy/lookup.go
2buisson: `backendErr` was only used to determine if the reply is valid
~~~ Go
if backendErr == nil {
    return reply, nil
}
~~~

I changed the `Exchange` behavior to use reply instead, so the (dnstap) error isn't fatal:

~~~ Go
// When *dns.Msg is not nil it is valid.
// When both *dns.Msg and error are not nil, the error should be reported.
func (d *dnsEx) Exchange(ctx context.Context, addr string, state request.Request) (*dns.Msg, error) {
~~~

miekg:
@varyoo you'll need some help with this to get started?

2buisson:
I feel it is ready for another review.

miekg:
>I feel it is ready for another review.

Yes, will do soon.

Thanks!

/Miek

--
Miek Gieben

middleware/dnstap/handler.go
```diff
@@ -20,6 +19,23 @@ type Dnstap struct {
...
+	}
+	tapContext struct {
+		context.Context
+		Dnstap
+	}
+)
+
+// TaperFromContext will return a Taper if the dnstap middleware is enabled.
+func TaperFromContext(ctx context.Context) (t Taper) {
+	t, _ = ctx.(Taper)
```
miekg: The Zero value of `Taper` is basically a noop? Then this is OK (and very nice)

middleware/dnstap/msg/msg.go
```diff
@@ -47,6 +54,59 @@ func (d *Data) FromRequest(r request.Request) error {
...
 
+type (
+	// Type is the dnstap message type.
+	Type func(*Data) *tap.Message
+
+	// Packer compiles the DNS message back to wire-format.
+	Packer func() ([]byte, error)
+
+	// Builder is a dnstap message builder.
+	// It aims to replace Data completely.
```
miekg: I don't understand what you mean with this second line of comment?

middleware/dnstap/msg/msg.go
```diff
@@ -47,6 +54,59 @@ func (d *Data) FromRequest(r request.Request) error {
...
+	Builder struct {
+		Type
+		Pack Packer
+		Data
+	}
+)
+
+// OutsideQuery is any query but a client query.
```
miekg: Can you elaborate a bit here?

middleware/dnstap/msg/msg.go
```diff
@@ -47,6 +54,59 @@ func (d *Data) FromRequest(r request.Request) error {
...
+// OutsideResponse is any response but a client response.
+func OutsideResponse(t tap.Message_Type) Type {
+	return func(d *Data) *tap.Message {
+		d.Type = t
+		return d.ToOutsideResponse()
+	}
+}
+
+// Build returns a dnstap message with the wire-format message
+// when full.
```
miekg: when `full` is enabled in the config?

middleware/dnstap/msg/msg.go
```diff
@@ -47,6 +54,59 @@ func (d *Data) FromRequest(r request.Request) error {
...
+			return nil, fmt.Errorf("pack: %s", err)
+		}
+		b.Packed = bin
+	}
+
+	b.Epoch()
+	return b.Type(&b.Data), nil
+}
+
+// FromRequest is deprecated.
```
miekg: ? delete it?

miekg (commented):
This looks like promising way of getting dnstap into other middleware. Few nit comments.

I want to double check what the problem is of doing this directly in proxy, because I dislike it only being available in the dns proxy.

middleware/dnstap/msg/msg.go
2buisson: Ultimately yes, but I meant the full parameter:
>func (b Builder) Build(**full bool**) (*tap.Message, error) {

middleware/dnstap/msg/msg.go
2buisson: It can be a `tap.Message_RESOLVER_QUERY` or `tap.Message_FORWARDER_QUERY`

middleware/dnstap/msg/msg.go
2buisson: Yes! But breaking changes in *middleware/dnstap*

middleware/dnstap/msg/msg.go
2buisson: I meant the `msg.Data struct`

middleware/dnstap/handler.go
2buisson: I hope it dosen't panic, this needs tests.

middleware/dnstap/handler.go
```diff
@@ -4,22 +4,40 @@ import (
...
+// Dnstap is the dnstap handler.
 type Dnstap struct {
 	Next middleware.Handler
 	Out  io.Writer
 	Pack bool
 }
 
+type (
+	// Taper is implemented by the Context passed by the dnstap handler.
+	Taper interface {
```
johnbelamaric: s/Taper/Tapper/

2buisson: Oops fixed.

johnbelamaric:
Echoing @miekg's question - why is this done in proxy/dns.go instead of proxy/proxy.go?

2buisson:
>why is this done in proxy/dns.go instead of proxy/proxy.go?

>Exchange(ctx context.Context, **addr** string, state request.Request) (*dns.Msg, error) {

Because *addr* is resolved inside Exchange and dnstap wants the IP.

johnbelamaric:
Hmm. So, you are saying

https://github.com/coredns/coredns/blob/master/middleware/proxy/proxy.go#L89

is sending `host.Name` which may be an FQDN instead of an IP?

I didn't think that was true. When creating the upstream, it uses 

https://github.com/coredns/coredns/blob/master/middleware/pkg/dnsutil/host.go#L14

which doesn't allow host names.

In fact, the examples in https://github.com/coredns/coredns/blob/master/middleware/proxy/README.md I believe are completely broken!

2buisson:
Oh you are right, meaning dnstap could fit in ServeDNS!

Small note: dnstap will need to rely on the DNS Exchanger indeed using TCP when ForceTCP is passed to it.

Ok then, I will try to come up with an implementation.

2buisson:
It is now done inside ServeDNS, with only a `Transport()` method added to the Exchanger interface.

yongtang:
@varyoo There are some additional `go lint` you may want to clean up:
```
root@98459fbd16c1:/go/src/github.com/coredns/coredns# make lint
go get -u github.com/golang/lint/golint
middleware/dnstap/msg/wrapper.go:18:1: exported function Marshal should have comment or be unexported
middleware/dnstap/test/helpers.go:13:6: exported type Context should have comment or be unexported
middleware/dnstap/test/helpers.go:18:1: exported function TestingData should have comment or be unexported
middleware/dnstap/test/helpers.go:58:1: exported function MsgEqual should have comment or be unexported
middleware/dnstap/test/helpers.go:62:6: exported type TrapTapper should have comment or be unexported
middleware/dnstap/test/helpers.go:67:1: exported method TrapTapper.TapMessage should have comment or be unexported
middleware/dnstap/test/helpers.go:72:1: exported method TrapTapper.TapBuilder should have comment or be unexported
middleware/rewrite/edns0.go:237:3: var serverIp should be serverIP
Makefile:73: recipe for target 'lint' failed
make: *** [lint] Error 1
root@98459fbd16c1:/go/src/github.com/coredns/coredns# 
```

yongtang:
ping @miekg for the review.

middleware/dnstap/msg/msg.go
```diff
@@ -24,8 +45,34 @@ type Data struct {
 	TimeSec     uint64
 }
 
-func (d *Data) FromRequest(r request.Request) error {
-	switch addr := r.W.RemoteAddr().(type) {
+// HostPort decodes into Data any string returned by dnsutil.ParseHostPortOrFile.
+func (d *Data) HostPort(addr string) (err error) {
+	ip, port, err := net.SplitHostPort(addr)
+	if err != nil {
+		return
```
miekg: with named params, don't do a naked return; just `return err` in this case.

middleware/dnstap/msg/msg.go
```diff
@@ -24,8 +45,34 @@ type Data struct {
...
+	}
+	p, err := strconv.ParseUint(port, 10, 32)
+	if err != nil {
+		return
+	}
+	d.Port = uint32(p)
+
+	if ip := net.ParseIP(ip); ip != nil {
```
miekg: reverse this:
```
ip := net.ParseIP(ip)
if ip == nil {
   return fmt.Errorf("not an in: %s", ip)
}
```
and then the test

middleware/proxy/dns.go
```diff
@@ -28,6 +28,12 @@ func newDNSExWithOption(opt Options) *dnsEx {
 	return &dnsEx{Timeout: defaultTimeout * time.Second, Options: opt}
 }
 
+func (d *dnsEx) Transport() (t string) {
```
miekg: remove this named param

middleware/proxy/dns.go
```diff
@@ -28,6 +28,12 @@ func newDNSExWithOption(opt Options) *dnsEx {
...
+	if d.Options.ForceTCP {
+		t = "tcp"
```
miekg: return "tcp"

middleware/proxy/dns.go
```diff
@@ -28,6 +28,12 @@ func newDNSExWithOption(opt Options) *dnsEx {
...
+	}
+	return
```
miekg: empty string signifies "not tcp" ?

middleware/proxy/exchanger.go
```diff
@@ -13,6 +13,9 @@ type Exchanger interface {
 	Exchange(ctx context.Context, addr string, state request.Request) (*dns.Msg, error)
 	Protocol() string
 
+	// Transport returns the only transport protocol used by this Exchanger or "".
```
miekg, line 16: But only tcp is significant? This doc string needs to be expanded so new implementers know what to do

miekg (commented):
A few nits, looks good.

Should prolly update this README.md or the one in proxy to tell about this feature.

middleware/dnstap/msg/msg.go
2buisson: It's a trick so I can use the same variable name for a different type under the `if` scope.

middleware/proxy/exchanger.go
2buisson: Now expanded.

PR #796, opened 2017-07-27
Add Dockerfile for automatic build in DockerHub

yongtang:
Add Dockerfile for automatic build in DockerHub

We need a special Dockerfile that pulls the release binary and not by using `ADD coredns /coredns` locally in DockerHub.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,7 +4,9 @@ MAINTAINER Miek Gieben <miek@coredns.io> @miekg
 # only need ca-certificates & openssl if want to use https_google
 RUN apk --update add bind-tools ca-certificates openssl && update-ca-certificates && rm -rf /var/cache/apk/*
 
-ENV COREDNS_VERSION 010
+ARG COREDNS_VERSION=latest
+
+RUN echo build coredns v$COREDNS_VERSION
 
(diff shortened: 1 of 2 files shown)


yongtang:
Ref #790

docker/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM alpine:latest
+MAINTAINER Miek Gieben <miek@miek.nl> @miekg
```
miekg: Should we change this, to (at least) something @coredns.io

docker/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+# only need ca-certificates & openssl if want to use https_google
+RUN apk --update add bind-tools ca-certificates openssl && update-ca-certificates && rm -rf /var/cache/apk/*
+
+ENV COREDNS_VERSION 010
```
miekg: this needs to be changed on every release?

yongtang:
@miekg It should be possible to invoke based on the SOURCE_BRANCH, let me play with it and see what we could do.

miekg (approved):

yongtang:
Let me merge and give it a try. I may need one or two additional rounds for further adjustment. The current code base will not be touched (other than that docker config) in the process.

PR #799, opened 2017-07-27
Use pre_build hook in docker hub to build the coredns binary

yongtang:
and skip the download process.

We don't need another Dockerfile, we actually only need to invoke a pre_build hook in DockerHub so that the coredns binary could be generated before we do a docker build with the existing `Dockerfile`.


miekg (approved):

miekg:
Do you reference docker/pre_build in the build system or is it a configurable thing? If configurable would rename the directory to `.docker`

yongtang:
@miekg it is in the build system. DockerHub invokes the following hook for automatic build:

https://docs.docker.com/docker-cloud/builds/advanced/#custom-build-phase-hooks

- hooks/post_checkout
- hooks/pre_build
- hooks/post_build
- hooks/pre_test
- hooks/post_test
- hooks/pre_push (only used when executing a build rule or automated build )
- hooks/post_push (only used when executing a build rule or automated build )
```

PR #805, opened 2017-08-01
mw/cache: small optimizations

miekg:
Scan for OPT records by checking last two records and then just
slice the slice.


miekg:
Not worth the hassle, also see https://github.com/miekg/dns/pull/500

PR #811, opened 2017-08-03
mw/kubernets: no need for interfaces

miekg:
Clean up the internfaces - they are not needed. And use k.LocalIP to
get the local IP.

### DIFF
--- a/middleware/kubernetes/kubernetes.go
+++ b/middleware/kubernetes/kubernetes.go
@@ -46,7 +46,6 @@ type Kubernetes struct {
 	ReverseCidrs  []net.IPNet
 	Fallthrough   bool
 	AutoPath
-	interfaceAddrs interfaceAddrser
 }
 
 type AutoPath struct {
@@ -651,7 +650,7 @@ func (k *Kubernetes) localPodIP() net.IP {
 	if localPodIP != nil {
 		return localPodIP
 	}
-	addrs, _ := k.interfaceAddrs.interfaceAddrs()
+	addrs, _ := net.InterfaceAddrs()
 
 	for _, addr := range addrs {
 		ip, _, _ := net.ParseCIDR(addr.String())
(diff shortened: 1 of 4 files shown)


chrisohaver:
I had added the interface to allow for unit tests to stub out the local interfaces function .  I thought this was idiomatic, but perhaps it's not - I'm still new to Go.  Is there a more idiomatic way to stub out a function?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] mw/kubernets:..." ]
>I had added the interface to allow for unit tests to stub out the interface .  I thought this was idiomatic, but perhaps it's not - I'm still new to Go.  Is there a more idiomatic way to stub out a function?

I deleted that test when I saw it failing. Then I realized why you went 
interfaces the first place. The main code got me quite puzzled and I don't think 
it is worth leaving it as-is. I think the test was mainly (re)testing if 
net.InterfaceAddrs() works.

chrisohaver:
The test removed is not testing `net.InterfaceAddrs()`... it's testing a method that relies on net.InterfaceAddrs() to work.... hence `net.InterfaceAddrs()` being stubbed out for the test.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] mw/kubernets:..." ]
>The test removed is not testing `net.InterfaceAddrs()`... it's testing a method that relies on net.InterfaceAddrs() to work.... hence `net.InterfaceAddrs()` being stubbed out for the test.

A better test (if needed at all) would be to just execute this and check if the 
IP of the machine is picked up (just check that it parses).

If you use the *bind* directive this can potentially pick up the wrong local ip,  
no? I.e. the one CoreDNS isn't listening on?

chrisohaver:
IMO, removing the test is a bad idea - because it reduces coverage.

What is being tested is the logic of determining the k8s service name based on the ip address of the pod.  This is not testing that we get the right IP.  This is only used ion the context of k8s in-cluster deployment (where coredns is a pod in k8s)... that pod, as we deploy it, will only have one ip.

miekg:
Fair enough. It does break when I add the test back in. But this code does not to be clean up a bit. localPodIP is reimplemented here. And those interfaces don't help you understand the code.

PR #820, opened 2017-08-04
Support multiple k8s api servers specification and load balance among api servers

### LINKED ISSUE #819: mw/k8s: load balance api servers (with multiple api server endpoint)
At the moment k8s middleware can only accept one api server endpoint. This is not ideal for HA.

In case of the api server error, k8s middleware will not be able to access k8s even though the etcd backend and other api servers might be functioning correctly.

It would be good to load balance the api server endpoint for HA.

yongtang:
This fix adds supports for multiple k8s api servers specification,
load balance among api servers.

When two or more api servers are specified in kubernetes block (endpoint ...),
a proxy is created locally (with randomly generately port). The coredns
will points to the generated proxy so that load balancing could be achieved.

This fix fixes #819

### DIFF
--- a/middleware/kubernetes/apiproxy.go
+++ b/middleware/kubernetes/apiproxy.go
@@ -1,0 +1,169 @@
+package kubernetes
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log"
+	"net"
+	"net/http"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+type proxyUpstream struct {
+	network   string
+	address   string
+	stop      chan struct{}
+	wait      sync.WaitGroup
+	unhealthy int32
+}
+
+type proxyHandler struct {
+	upstreams []proxyUpstream
+	pickIndex int32
+}
+
+type apiProxy struct {
+	http.Server
+	listener net.Listener
+	handler  proxyHandler
+}
+
+func (p proxyUpstream) Network() string {
+	return p.network
+}
+
+func (p proxyUpstream) String() string {
+	return p.address
+}
+
+func (p *proxyUpstream) Healthcheck() {
+	p.wait.Add(1)
+	defer p.wait.Done()
+
+	status := func(unhealthy int32) string {
+		if unhealthy == 0 {
+			return "healthy"
+		}
+		return "unhealthy"
+	}
+
+	client := &http.Client{
+		Transport: &http.Transport{
+			DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+				return net.Dial(p.Network(), p.String())
+			},
+		},
+		Timeout: 5 * time.Second,
+	}
+	tick := time.NewTicker(5 * time.Second)
+	// Start healthcheck immediately
+	unhealthy := int32(1)
+	if resp, err := client.Get("http://kubernetes/"); err == nil {
+		defer resp.Body.Close()
+		unhealthy = 0
+	}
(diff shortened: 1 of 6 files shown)


middleware/kubernetes/apiproxy.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (p proxyUpstream) Network() string {
+	return p.network
+}
+
+func (p proxyUpstream) String() string {
+	return p.address
+}
+
+func (p *proxyUpstream) Healthcheck(once bool) {
```
miekg: Did you look at the health checks in middleware/proxy? Looks awfully similar... Either using that directly or getting a new pkg/healthcheck subpackage might sense here.

middleware/kubernetes/setup.go
```diff
@@ -129,7 +135,7 @@ func kubernetesParse(c *caddy.Controller) (*Kubernetes, error) {
 				case "endpoint":
 					args := c.RemainingArgs()
 					if len(args) > 0 {
-						k8s.APIEndpoint = args[0]
+						k8s.APIServerList = strings.Split(args[0], ",")
```
miekg: this doesn't allow for spaces?

johnbelamaric:
I think getClientConfig is only called at setup time. So, to make this truly useful, we need to re-start the K8s middleware when the API connection fails.

chrisohaver:
README.md should be updated...

yongtang:
Thanks. Let me update the PR.

yongtang:
The API connection is actually pointing to the local proxy we generate so the address `127.0.0.1:0` (`127.0.0.1:xxxx`) is actually fixed for kubernetes to connect. I remember kubernetes will always re-try. 

I think it should be fine for out-of-cluster deployment.

In case of connection fails no config change in kubernetes side.

Not sure about in-cluster deployment.

yongtang:
For in-cluster deployment it seems that this is handled by k8s's API `rest.InClusterConfig`, which we don't have control.

However, k8s is not a true-HA platform anyway (you have to add additional components on top of k8s to achieve that). 

I had a solution of adding nginx (exposing 8080) to each node of a k8s cluster and points k8s scheduler/controller-manager to nginx (`--master=127.0.0.1:8080`):

https://github.com/yongtang/kubernetes-cfn/blob/master/cfn/vpc.yaml

But that puts additional requirement to user's k8s practice.

I think for CoreDNS deployment, we should suggest users to use out-of-cluster deployment if they want HA.

yongtang:
See related discussion about HA in k8s:
kubernetes/kubernetes#18174

johnbelamaric:
@yongtang HA for CoreDNS accessing K8s API when in-cluster is provided by kube-proxy. The discussion in that issue is that kube-proxy in that case. The discussion there is about how kube-proxy cannot use itself to round-robin HA its own API server connection.

middleware/kubernetes/setup.go
yongtang: Thanks. Done.

yongtang:
The README.md has been updated. The whitespace has been removed as well. Will look into health check consolidation.

yongtang:
Added PR #854 to move health check to `middleware/pkg/healthcheck`. Will update once #854 is merged.

yongtang:
All tests passed now.

yongtang:
Let me merge this PR now. The kube-proxy is probably outside the scope and has to be resolved in some other ways.

PR #822, opened 2017-08-04
middleware/kubernetes: there are no invalid requests

miekg:
First step to centralize the qtype checking in the kubernetes
middleware. Let parseRequest just take the name and find things there.
Remove the ErrInvalidRequest is such a thing does not exist, it also
makes it harder to return NODATA for responses that need it.

### DIFF
--- a/middleware/kubernetes/kubernetes_test.go
+++ b/middleware/kubernetes/kubernetes_test.go
@@ -1,7 +1,6 @@
 package kubernetes
 
 import (
-	"errors"
 	"net"
 	"reflect"
 	"testing"
@@ -15,7 +14,7 @@ import (
 
 func TestRecordForTXT(t *testing.T) {
 	k := Kubernetes{Zones: []string{"inter.webs.test"}}
-	r, _ := k.parseRequest("dns-version.inter.webs.test", dns.TypeTXT)
+	r, _ := k.parseRequest("dns-version.inter.webs.test")
 
 	expected := DNSSchemaVersion
 	svc := k.recordsForTXT(r)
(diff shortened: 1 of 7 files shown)


middleware/kubernetes/kubernetes_test.go
```diff
@@ -212,36 +194,6 @@ func TestParseRequest(t *testing.T) {
 	for field, expected := range tcs {
 		expectString(t, f, "TXT", query, &r, field, expected)
 	}
-
-	// Invalid query tests
```
chrisohaver, line 216: Perhaps changes these tests to validate NODATA response, instead of just removing them.

miekg: Good point. We seem to have integration tests doing that already (or at least that would be the spot to add them)

johnbelamaric: I think if you do the tests here they show up in coverage, but not if you do them in the integration tests.

test/kubernetes_test.go
```diff
@@ -472,8 +469,6 @@ func doIntegrationTests(t *testing.T, corefile string, testCases []test.Case) {
 		if len(res.Answer) != len(tc.Answer) {
 			t.Errorf("Expected %d answers but got %d for query %s, %d", len(tc.Answer), len(res.Answer), tc.Qname, tc.Qtype)
 		}
-
-		//TODO: Check the actual RR values
```
chrisohaver, line 476: IMO, This still needs to be done...

chrisohaver:
I think you'll run into problems/ambiguity/inefficiency problems down the road if you try to completely remove awareness of query type from k8s middleware.

With the recent `SRV` change (treat missing `SRV` prefix as wild card), that means we cant identify an SRV query just by the presence of service/protocol prefixes in the question ... we need to know the type.

Later when querying services, if we don't know the query type, we will have to build out records for all open ports on all matching services for every query (instead of just doing that extra work when we need to do it)...

miekg:
>I think you'll run into problems/ambiguity/inefficiency problems down the road if you try to completely remove awareness of query type from k8s middleware.
>
>With the recent `SRV` change (treat missing `SRV` prefix as wild card), that means we cant identify an SRV query just by the presence of service/protocol prefixes in the question ... we need to know the type.
>
>Later when querying services, if we don't know the query type, we will have to build out records for all open ports on all matching services for every query (instead of just doing that extra work when we need to do it)...

Ok. I haven't checked that last bit of the k8s API, i.e. where r.port and 
r.transport are put into k8s' API to get a result.

My issue is mainly that we already check the type in routRequest, so I don't 
think we should repeat those later in the process.

I'm fine with closing this for the time being, while I check ns.go to remove to 
rely on that global var.

Issue #808, opened 2017-08-03
mw/kubernetes: handle incomplete SRV requests


chrisohaver:
Treat missing `_service._protocol.` segments as wild cards.

miekg:
With qtype A this is already done? Any other types should have the same behavior btw. Should check that for etcd as well, can't give NXDOMAIN for any name that is known to exist.

chrisohaver:
i think you meant to comment on the other issue.  this one is about handling incomplete name.
But if you are speaking about this issue... the answer is no, and incomplete A request will not assume a wildcard for all requests...

e.g. A request for `namespace.svc.zone.` is not equivalent to `*.namespace.svc.zone.` ... A request for `namespace.svc.zone.` will not return a list `A` records for all services.

chrisohaver:
Looks like an incompletely incomplete query doesn't get handled properly...

e.g. an SRV request for `*.svc-1-a.test-1.svc.cluster.local.` returns `NXDOMAIN` when the service `svc-1-a.test-1.svc.cluster.local.` exists.

chrisohaver:
Closing this - `NXDOMAIN` is the expected behavior here per #975

PR #823, opened 2017-08-04
Kubernetes srv

CryptoCopter:
Changes behaviour of the k8s-middleware so that SRV-requests without a specified port & service are handled as wildcards.
Some applications use the k8s [peer-finder](https://github.com/kubernetes/contrib/tree/master/peer-finder) for bootstrapping during deployment. This uses net.LookupSRV("", "", service_name) for peer-discovery. If such requests are not treated as valid, these applications will not deploy.

Implements behaviour of #808

### DIFF
--- a/middleware/kubernetes/kubernetes.go
+++ b/middleware/kubernetes/kubernetes.go
@@ -297,34 +297,42 @@ func (k *Kubernetes) parseRequest(lowerCasedName string, qtype uint16) (r record
 
 	offset := 0
 	if qtype == dns.TypeSRV {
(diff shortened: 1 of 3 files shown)


chrisohaver (approved):

strigazi:
@johnbelamaric @chrisohaver Could we have a release to get this fix for kubernetes?

miekg:
see #1012

afaik there is a public holiday coming up in the US as well, might be nice if everybody is in the office when we do a 1.0.0 release.

Also: feel free to compile from source, current master will basically be 1.0.0

PR #826, opened 2017-08-04
Misspell correction

yongtang:
Did a `misspell . | grep -v ^vendor` and fixed several typos.

### DIFF
--- a/core/zmiddleware.go
+++ b/core/zmiddleware.go
@@ -10,6 +10,7 @@ import (
 	_ "github.com/coredns/coredns/middleware/chaos"
 	_ "github.com/coredns/coredns/middleware/debug"
 	_ "github.com/coredns/coredns/middleware/dnssec"
+	_ "github.com/coredns/coredns/middleware/dnstap"
 	_ "github.com/coredns/coredns/middleware/erratic"
 	_ "github.com/coredns/coredns/middleware/errors"
 	_ "github.com/coredns/coredns/middleware/etcd"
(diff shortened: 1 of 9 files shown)


core/zmiddleware.go
```diff
@@ -10,6 +10,7 @@ import (
 	_ "github.com/coredns/coredns/middleware/chaos"
 	_ "github.com/coredns/coredns/middleware/debug"
 	_ "github.com/coredns/coredns/middleware/dnssec"
+	_ "github.com/coredns/coredns/middleware/dnstap"
```
miekg, line 13: we didn't enabled it?

chrisohaver: Yeah - this has been really annoying.

chrisohaver: probably included by accident in this PR.  but should be fixed anyway...

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Misspell corr..." ]
>
>
>
>> @@ -10,6 +10,7 @@ import (
> 	_ "github.com/coredns/coredns/middleware/chaos"
> 	_ "github.com/coredns/coredns/middleware/debug"
> 	_ "github.com/coredns/coredns/middleware/dnssec"
>+	_ "github.com/coredns/coredns/middleware/dnstap"
>
>probably included by accident in this PR.  but should be fixed anyway...

Indeed. The release-build Makefile doesn't include a 'go generate' step, it 
probably should.

yongtang:
I did a `make` before push the PR. Might be generated automatically then.

johnbelamaric:
yeah, the regular make does that, but not the release make.

yongtang:
It would be good if we could consolidate Makefile and Makefile.release so that they have the same flow and flags.

Issue #814, opened 2017-08-03
mw/kubernetes: Use indexes for endpoints and services caches


chrisohaver:
Add indexing to endpoints and services caches, (pod cache already uses indexes).

chrisohaver:
The pods cache already uses indexes to look up pods `PodIndex(ip string)`.

The same should be done for services and endpoints, removing `ServiceList()` and `EndpointsList()`, and replacing with new methods `ServiceIndex(service, namespace string)`, and `EndpointIndex(service, namespace string)`.

miekg:
Is something happening here?

chrisohaver:
Yes. It's in progress.

johnbelamaric:
#1149

PR #830, opened 2017-08-04
WIP: set nsAddress

miekg:
Resolve the NS Record IP on startup.

### DIFF
--- a/middleware/kubernetes/kubernetes.go
+++ b/middleware/kubernetes/kubernetes.go
@@ -46,7 +46,9 @@ type Kubernetes struct {
 	ReverseCidrs  []net.IPNet
 	Fallthrough   bool
 	AutoPath
+
 	interfaceAddrsFunc func() net.IP
+	nsAddr             *dns.A // IP for the "fake" NS record
 }
 
 // AutoPath enables server side search path lookups for pods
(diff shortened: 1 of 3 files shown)


miekg:
@chrisohaver so the problem with this PR will be that we don't when exactly to get `k.nsAddr`? Mostly because the setup function are async.

chrisohaver:
The initial seed of the cache may not be async.
Periodic refresh of cache is async, and so are incremental updates via watchers.

As the PR is now, the code will look up the name in the service list every time an `NS` or `SOA`  is requested.  This incurs a penalty because the service cache is not indexed. The penalty greater for deployments with a larger number of services.
I have opened an issue to add indexing for services and endpoints (#814), and currently working on implementing it. I didnt open the issue for this `NS` record reason, rather for general performance improvement at scale in k8s.
Once it's indexed though, it's less of an penalty to look up every time (but still not as efficient as if we just stashed the result somewhere).

Still - it depends on your concern for performance of `NS` and `SOA` records.  If this isn't a concern, then none of this matters - We can look it up every time.  When the indexing comes along later, it will be a welcome (but not critical) improvement.

miekg:
Thanks for the write up. We should def. cache this.

miekg:
Looking through the kube-dns code it has "waitForResourceSyncedOrDie()" which seems to exactly what we need before setting this or letting queries through.

PR #834, opened 2017-08-05
middleware/secondary: pull all zones

miekg:
Use the same OnStartup setup as in the file middleware. We need to copy
the variable from range, otherwise it gets overwriten in the next loop
because of the async goroutine call.

Hard to test, we have secondary_test.go which we could extend with
multiple zones for instance. For now this fix does not have an test case
with it...

### DIFF
--- a/middleware/secondary/setup.go
+++ b/middleware/secondary/setup.go
@@ -25,12 +25,13 @@ func setup(c *caddy.Controller) error {
 
 	// Add startup functions to retrieve the zone and keep it up to date.
 	for _, n := range zones.Names {
-		if len(zones.Z[n].TransferFrom) > 0 {
+		z := zones.Z[n]
+		if len(z.TransferFrom) > 0 {
 			c.OnStartup(func() error {
-				zones.Z[n].StartupOnce.Do(func() {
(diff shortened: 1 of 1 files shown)


miekg:
Locally tested with:
~~~
. {
    secondary miek.nl coredns.io {
        transfer from 2a01:7e00::f03c:91ff:fe79:234c
    }
}
~~~

and:
~~~
2017/08/05 09:01:41 [INFO] Transferred: miek.nl. from [2a01:7e00::f03c:91ff:fe79:234c]:53
2017/08/05 09:01:41 [INFO] Transferred: coredns.io. from [2a01:7e00::f03c:91ff:fe79:234c]:53
~~~

kfox1111:
Is there a release schedule for v011?

johnbelamaric:
Next is v1.0.0. When all the issues tagged with 1.0.0 are fixed. Usually it's about monthly. No fixed schedule though.

johnbelamaric:
https://github.com/coredns/coredns/milestone/10

But subject to change of course.

kfox1111:
Ok. thanks.

PR #838, opened 2017-08-05
Core: convert IP addresses to reverse zone

miekg:
If we see IP/mask syntax and the mask mod 8 == 0 we assume a reverse
zone and convert to in-addr or .arpa.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,17 +168,18 @@ example.org {
 }
 ~~~
 
-IP address that are automatically converted to reverse zones is also allowed:
+IP addresses are also allowed. They are automatically converted to reverse zones:
 
 ~~~ txt
 10.0.0.0/24 {
     # ...
 }
 ~~~
-Means you are authoritative for `0.0.10.in-addr.arpa.`. The netmask must be dividable by 8, if it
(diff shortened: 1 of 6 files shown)


miekg:
@johnbelamaric this pull the reverse addresses in the corefile, this obviously does not support any netmasks that are not a multiple of 8.

In the kubernetes middeware the cidrs would only work if you would are authoritative for `.` in your zone stanza, otherwise the reverse queries wouldn't even reach the middleware.

README.md
```diff
@@ -168,6 +168,19 @@ example.org {
...
+IP addresses are also allowed. They are automatically converted to reverse zones:
+
+~~~ txt
+10.0.0.0/24 {
+    # ...
+}
+~~~
+Means you are authoritative for `0.0.10.in-addr.arpa.`. 
+
+The netmask must be dividable by 8, if it is not the reverse conversion is not done. This also works
```
johnbelamaric, line 180: Shouldn't we implement https://www.ietf.org/rfc/rfc2317.txt for classless networks?

miekg:
Only in the case when your parent has this delegated to you. In the case of
kubernetes for instance no client would generate a reverse address that
would match this zone name... But happy to add this for v4 addresses.

PR #843, opened 2017-08-06
Metalinter

miekg:


miekg:
This does change the signature of NextOrFailure to have the context first (should have been like this from the begining), but is still a backwards incompat change.

miekg:
I'm gonna revert that context change - fairly big change with no big benefits, except to keep some linter happy.

PR #845, opened 2017-08-07
middleware/log: add rflags to logging

miekg:
Add the DNS message flags as {rflags} to the default logging

### DIFF
--- a/core/dnsserver/zdirectives.go
+++ b/core/dnsserver/zdirectives.go
@@ -35,8 +35,8 @@ var directives = []string{
 	"secondary",
 	"etcd",
 	"proxy",
-	"whoami",
 	"erratic",
+	"whoami",
 	"startup",
 	"shutdown",
 }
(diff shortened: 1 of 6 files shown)


miekg:
I sneakily also switched the middleware ordering so `whoami` sits behind erratic, allowing this:
~~~
. {
    whoami
    log
    erratic {
        truncate 3
    }
}
~~~
And the `./coredns -dns.port=1053`, which then shows:

~~~
.:1053
2017/08/07 09:55:13 [INFO] CoreDNS-010
CoreDNS-010
::1 - [07/Aug/2017:09:55:14 +0000] "A IN haad.miek.nl. udp 42 false 4096" NOERROR qr,aa,tc,rd 58 186.067µs
::1 - [07/Aug/2017:09:55:14 +0000] "A IN haad.miek.nl. tcp 42 false 65535" NOERROR qr,aa,rd 58 80.653µs
~~~
Showing the TC and the immediate retry with tcp from dig.

PR #854, opened 2017-08-08
Move Healthcheck to middleware/pkg/healthcheck

yongtang:
So that it could be reused by other middleware.


miekg:
These are mostly moves I assume?

miekg (approved):

yongtang:
@miekg Yes mostly just move the files to middleware/pkg/healthcheck.

PR #864, opened 2017-08-09
middleware/kuberntes: add new parser

miekg:
Slow move stuff over - also test the new parseQuery.


miekg:
pulling txt handling out is worth-while, but this PR might want to try to do too much. I'll cut it a bit back (leave current parsing - do want to fix the tests for it though)

Issue #870, opened 2017-08-09
middleware/kubernets: fix AutoPath callback


miekg:
That function: https://github.com/coredns/coredns/blob/master/middleware/kubernetes/autopath.go#L11
returns a path that the autopath middleware can use.

The current code return some nonsense (couldn't figure out what it should do). This method should return the correct searchpath.

@chrisohaver could you help here?

chrisohaver:
It should be the 3 k8s domains ...

```
namespace.svc.zone.
svc.zone.
zone.
```
(namespace = namespace of the remote pod)
Followed by the domains pulled from the local `resolv.conf`

miekg:
Thanks!

and `zone` must be the zone from the query (if it falls in the zone list k8s is auth for)? Or only the primary domain of kubernetes?

chrisohaver:
zone from the query.

miekg:
ack. Bit of a bummer to reparse the query again. But I guess that makes sense. I'll submit a PR with those changes.

chrisohaver:
Correction... it can be the primary zone.  I don't understand the use case for multiple zones though.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/ku..." ]
>Correction... it can be the primary zone.  I don't understand the use case for multiple zones though.

If the query matches one of the zones that the k8s middleware handles this 
should not matter.

chrisohaver:
I think there can be only one cluster domain per cluster.

miekg:
But you might still want autopathing?

chrisohaver:
cluster domain = zone....  one zone per cluster.

PR #871, opened 2017-08-09
middleware/authpath: Fix return from k8s mw

miekg:
Return the correct search path from the kubernetes' AutoPath function.
Based on preliminary discussion in #870

### DIFF
--- a/middleware/kubernetes/autopath.go
+++ b/middleware/kubernetes/autopath.go
@@ -9,6 +9,8 @@ import (
 	"k8s.io/client-go/1.5/pkg/api"
 )
 
+// AutoPath implements the AutoPathFunc call from the autopath middleware.
+// It returns a per-query search path or an error indicating no searchpathing should happen.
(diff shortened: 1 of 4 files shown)


middleware/kubernetes/autopath.go
```diff
@@ -3,30 +3,46 @@ package kubernetes
...
+		return nil, fmt.Errorf("kubernetes: no authoriative for %s", state.Name())
+	}
+
 	ip := state.IP()
 
-	pod := k.PodWithIP(ip)
+	pod := k.podWithIP(ip)
 	if pod == nil {
-		return nil, fmt.Errorf("kubernets: no pod found for %s", ip)
+		return nil, fmt.Errorf("kubernetes: no pod found for %s", ip)
```
johnbelamaric: Do we need this to be an error? I think it should just mean "do not attempt the auto path optimization" but not be considered an error condition, per se

miekg: No it does not have to be an error - we can treat nil as a clear signal as well.

Issue #831, opened 2017-08-04
mw/k8s: Add RR checking to k8s integration tests


chrisohaver:
Currently the integration tests for kubernetes (tests/kubernetes_test.go) only validate the number of responses returned per each test.

The tests should also validate the content of the responses, much in the same way that middleware/kubernetes/handler_test.go TestServeDNS() unit test does.

rajansandeep:
Implemented via #884

PR #884, opened 2017-08-10
Middleware/Kubernetes: Add RR check to K8s integration tests

rajansandeep:
Added the RR checks to Kubernetes integration test as per issue #831.

### DIFF
--- a/middleware/test/helpers.go
+++ b/middleware/test/helpers.go
@@ -13,7 +13,7 @@ type sect int
 const (
 	// Answer is the answer section in an Msg.
 	Answer sect = iota
-	// Ns is the authrotitative section in an Msg.
+	// Ns is the authoritative section in an Msg.
 	Ns
 	// Extra is the additional section in an Msg.
 	Extra
@@ -262,6 +262,23 @@ func Section(t *testing.T, tc Case, sec sect, rr []dns.RR) bool {
 		}
 	}
 	return true
+}
+
+// CNAMEOrder makes sure that CNAMES do not appear after their target records
+func CNAMEOrder(t *testing.T, res *dns.Msg) {
+	for i, c := range res.Answer {
+		if c.Header().Rrtype != dns.TypeCNAME {
+			continue
+		}
+		for _, a := range res.Answer[:i] {
+			if a.Header().Name != c.(*dns.CNAME).Target {
+				continue
+			}
+			t.Errorf("CNAME found after target record\n")
+			t.Logf("%v\n", res)
+
+		}
+	}
 }
 
 // SortAndCheck sorts resp and the checks the header and three sections against the testcase in tc.
--- a/middleware/test/helpers.go
+++ b/middleware/test/helpers.go
@@ -308,6 +308,23 @@ func SortAndCheck(t *testing.T, resp *dns.Msg, tc Case) {
 	return
 }
 
+// Function to make sure that CNAMES do not appear after their target records
+func CNAMEOrder(t *testing.T, res *dns.Msg) {
+	for i, c := range res.Answer {
+		if c.Header().Rrtype != dns.TypeCNAME {
+			continue
+		}
+		for _, a := range res.Answer[:i] {
+			if a.Header().Name != c.(*dns.CNAME).Target {
+				continue
+			}
+			t.Errorf("CNAME found after target record\n")
+			t.Logf("%v\n", res)
+
+		}
+	}
+}
+
 // ErrorHandler returns a Handler that returns ServerFailure error when called.
 func ErrorHandler() Handler {
 	return HandlerFunc(func(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
--- a/middleware/test/helpers.go
+++ b/middleware/test/helpers.go
@@ -308,7 +308,7 @@ func SortAndCheck(t *testing.T, resp *dns.Msg, tc Case) {
 	return
 }
 
-// Function to make sure that CNAMES do not appear after their target records
+// CNAMEOrder makes sure that CNAMES do not appear after their target records
 func CNAMEOrder(t *testing.T, res *dns.Msg) {
 	for i, c := range res.Answer {
 		if c.Header().Rrtype != dns.TypeCNAME {
--- a/middleware/test/helpers.go
+++ b/middleware/test/helpers.go
@@ -308,23 +308,6 @@ func SortAndCheck(t *testing.T, resp *dns.Msg, tc Case) {
 	return
 }
 
-// CNAMEOrder makes sure that CNAMES do not appear after their target records
-func CNAMEOrder(t *testing.T, res *dns.Msg) {
-	for i, c := range res.Answer {
-		if c.Header().Rrtype != dns.TypeCNAME {
-			continue
-		}
-		for _, a := range res.Answer[:i] {
-			if a.Header().Name != c.(*dns.CNAME).Target {
-				continue
-			}
-			t.Errorf("CNAME found after target record\n")
-			t.Logf("%v\n", res)
-
-		}
-	}
-}
-
 // ErrorHandler returns a Handler that returns ServerFailure error when called.
 func ErrorHandler() Handler {
 	return HandlerFunc(func(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
--- a/test/kubernetes_test.go
+++ b/test/kubernetes_test.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"sort"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -200,10 +201,10 @@ var dnsTestCases = []test.Case{
 	{
 		Qname: "_c-port._UDP.*.test-1.svc.cluster.local.", Qtype: dns.TypeSRV,
 		Rcode: dns.RcodeSuccess,
-		Answer: append(headlessSRVResponse("_c-port._UDP.*.test-1.svc.cluster.local.", "TypeSRV", "headless-svc", "test-1"),
+		Answer: append(srvResponse("_c-port._UDP.*.test-1.svc.cluster.local.", "TypeSRV", "headless-svc", "test-1"),
 			[]dns.RR{
 				test.SRV("_c-port._UDP.*.test-1.svc.cluster.local.      303    IN    SRV 0 33 1234 svc-c.test-1.svc.cluster.local.")}...),
-		Extra: append(headlessSRVResponse("_c-port._UDP.*.test-1.svc.cluster.local.", "TypeA", "headless-svc", "test-1"),
+		Extra: append(srvResponse("_c-port._UDP.*.test-1.svc.cluster.local.", "TypeA", "headless-svc", "test-1"),
 			[]dns.RR{
 				test.A("svc-c.test-1.svc.cluster.local.	303	IN	A	10.0.0.115")}...),
 	},
@@ -250,14 +251,9 @@ var dnsTestCases = []test.Case{
 	},
 	{
 		Qname: "*.svc-1-a.test-1.svc.cluster.local.", Qtype: dns.TypeSRV,
-		Rcode: dns.RcodeSuccess,
-		Answer: []dns.RR{
-			test.SRV("*.svc-1-a.test-1.svc.cluster.local.   0       IN      SRV     0 50 443 svc-1-a.test-1.svc.cluster.local."),
-			test.SRV("*.svc-1-a.test-1.svc.cluster.local.   0       IN      SRV     0 50 80  svc-1-a.test-1.svc.cluster.local."),
-		},
-		Extra: []dns.RR{
-			test.A("svc-1-a.test-1.svc.cluster.local.	0	IN	A	10.0.0.100"),
-		},
+		Rcode:  dns.RcodeSuccess,
+		Answer: srvResponse("*.svc-1-a.test-1.svc.cluster.local.", "TypeSRV", "svc-1-a", "test-1"),
+		Extra:  srvResponse("*.svc-1-a.test-1.svc.cluster.local.", "TypeA", "svc-1-a", "test-1"),
 	},
 	{
 		Qname: "*._not-udp-or-tcp.svc-1-a.test-1.svc.cluster.local.", Qtype: dns.TypeSRV,
@@ -537,12 +533,12 @@ var dnsTestCasesFallthrough = []test.Case{
 	{
 		Qname: "_c-port._UDP.*.test-1.svc.cluster.local.", Qtype: dns.TypeSRV,
 		Rcode: dns.RcodeSuccess,
-		Answer: append(headlessSRVResponse("_c-port._UDP.*.test-1.svc.cluster.local.", "TypeSRV", "headless-svc", "test-1"),
+		Answer: append(srvResponse("_c-port._UDP.*.test-1.svc.cluster.local.", "TypeSRV", "headless-svc", "test-1"),
 			[]dns.RR{
 				test.SRV("_c-port._UDP.*.test-1.svc.cluster.local.      303    IN    SRV 0 33 1234 svc-c.test-1.svc.cluster.local."),
 			}...),
 
-		Extra: append(headlessSRVResponse("_c-port._UDP.*.test-1.svc.cluster.local.", "TypeA", "headless-svc", "test-1"),
+		Extra: append(srvResponse("_c-port._UDP.*.test-1.svc.cluster.local.", "TypeA", "headless-svc", "test-1"),
 			[]dns.RR{
 				test.A("svc-c.test-1.svc.cluster.local.	303	IN	A	10.0.0.115"),
 			}...),
@@ -590,13 +586,9 @@ var dnsTestCasesFallthrough = []test.Case{
 	},
 	{
 		Qname: "*.svc-1-a.test-1.svc.cluster.local.", Qtype: dns.TypeSRV,
-		Rcode: dns.RcodeSuccess,
-		Answer: []dns.RR{
-			test.SRV("*.svc-1-a.test-1.svc.cluster.local.	303	IN	SRV	0 50 443 svc-1-a.test-1.svc.cluster.local."),
-			test.SRV("*.svc-1-a.test-1.svc.cluster.local.	303	IN	SRV	0 50 80  svc-1-a.test-1.svc.cluster.local.")},
-		Extra: []dns.RR{
-			test.A("svc-1-a.test-1.svc.cluster.local.	0	IN	A	10.0.0.100"),
-		},
+		Rcode:  dns.RcodeSuccess,
+		Answer: srvResponse("*.svc-1-a.test-1.svc.cluster.local.", "TypeSRV", "svc-1-a", "test-1"),
+		Extra:  srvResponse("*.svc-1-a.test-1.svc.cluster.local.", "TypeA", "svc-1-a", "test-1"),
 	},
 	{
 		Qname: "*._not-udp-or-tcp.svc-1-a.test-1.svc.cluster.local.", Qtype: dns.TypeSRV,
@@ -851,9 +843,9 @@ func headlessAResponse(qname, namespace, name string) []dns.RR {
 	return rr
 }
 
-// headlessSRVResponse returns the answer to a SRV request for the specific name and namespace
+// srvResponse returns the answer to a SRV request for the specific name and namespace
 // responsetype is the type of answer to generate, eg: TypeSRV ( for answer section) or TypeA (for extra section)
-func headlessSRVResponse(qname, responsetype, namespace, name string) []dns.RR {
+func srvResponse(qname, responsetype, namespace, name string) []dns.RR {
 	rr := []dns.RR{}
 
 	str, err := endpointIPs(name, namespace)
@@ -866,11 +858,18 @@ func headlessSRVResponse(qname, responsetype, namespace, name string) []dns.RR {
 
 	for i := 0; i < lr; i++ {
 		ip := strings.Replace(result[i], ".", "-", -1)
-		if responsetype == "TypeSRV" {
-			rr = append(rr, test.SRV(qname+"   303    IN    SRV 0 33 1234  "+ip+"."+namespace+"."+name+".svc.cluster.local."))
-		} else {
+		t := strconv.Itoa(100 / (lr + 1))
+		if responsetype == "TypeA" {
 			rr = append(rr, test.A(ip+"."+namespace+"."+name+".svc.cluster.local.	0	IN	A	"+result[i]))
 		}
+		if responsetype == "TypeSRV" && namespace == "headless-svc" {
+			rr = append(rr, test.SRV(qname+"   303    IN    SRV 0 "+t+" 1234  "+ip+"."+namespace+"."+name+".svc.cluster.local."))
+		}
+		if responsetype == "TypeSRV" && namespace != "headless-svc" {
+			rr = append(rr, test.SRV(qname+"   303    IN    SRV 0 "+t+" 443  "+ip+"."+namespace+"."+name+".svc.cluster.local."))
+			rr = append(rr, test.SRV(qname+"   303    IN    SRV 0 "+t+" 80  "+ip+"."+namespace+"."+name+".svc.cluster.local."))
+		}
+
 	}
 	return rr
 }
--- a/test/kubernetes_test.go
+++ b/test/kubernetes_test.go
@@ -875,7 +875,7 @@ func headlessSRVResponse(qname, responsetype, namespace, name string) []dns.RR {
 	return rr
 }
 
-//endpointIPs retrieves the IP address for a given name and namespace by parsing json using kubectl command
+//endpointIPs retrieves the IP address for a given name and namespace by parsing json using the kubectl command
 func endpointIPs(name, namespace string) (cmdOut []byte, err error) {
 
 	kctl := os.Getenv("KUBECTL")
--- a/test/kubernetes_test.go
+++ b/test/kubernetes_test.go
@@ -875,7 +875,7 @@ func headlessSRVResponse(qname, responsetype, namespace, name string) []dns.RR {
 	return rr
 }
 
-//endpointIPs retrieves the IP address for a given name and namespace by parsing json using the kubectl command
+//endpointIPs retrieves the IP address for a given name and namespace by parsing json using kubectl command
 func endpointIPs(name, namespace string) (cmdOut []byte, err error) {
 
 	kctl := os.Getenv("KUBECTL")
--- a/test/kubernetes_test.go
+++ b/test/kubernetes_test.go
@@ -705,7 +705,7 @@ func doIntegrationTests(t *testing.T, corefile string, testCases []test.Case) {
 		// Before sorting, make sure that CNAMES do not appear after their target records
 		test.CNAMEOrder(t, res)
 
-		// Sorting the test cases to check with the responses
+		// Sorting the test cases to check with the response
 		sort.Sort(test.RRSet(tc.Answer))
 		sort.Sort(test.RRSet(tc.Ns))
 		sort.Sort(test.RRSet(tc.Extra))
--- a/test/kubernetes_test.go
+++ b/test/kubernetes_test.go
@@ -885,7 +885,13 @@ func endpointIPs(name, namespace string) (cmdOut []byte, err error) {
 	}
 	cmdArgs := kctl + " -n " + name + " get endpoints " + namespace + " -o jsonpath={.subsets[*].addresses[*].ip}"
 	if cmdOut, err = exec.Command("sh", "-c", cmdArgs).Output(); err != nil {
+<<<<<<< HEAD
 		return nil, err
+=======
+		println(" ")
+		println(os.Stderr, "There was an error running kubectl command: ", err.Error())
+		os.Exit(1)
+>>>>>>> additional mods to kubectl run try 2
 	}
 	return cmdOut, nil
 }
--- a/test/kubernetes_test.go
+++ b/test/kubernetes_test.go
@@ -252,11 +252,11 @@ var dnsTestCases = []test.Case{
 		Qname: "*.svc-1-a.test-1.svc.cluster.local.", Qtype: dns.TypeSRV,
 		Rcode: dns.RcodeSuccess,
 		Answer: []dns.RR{
-			test.SRV("*.svc-1-a.test-1.svc.cluster.local.   0       IN      SRV     0 50 443 10-1-0-10.svc-1-a.test-1.svc.cluster.local."),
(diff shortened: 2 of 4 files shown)


test/kubernetes_test.go
```diff
@@ -1,9 +1,8 @@
-// +build k8s
```
chrisohaver: Dont remove this line, its a build constraint ... see https://golang.org/pkg/go/build/
It is used by the make file to control when this test is executed.

chrisohaver:
The test failures are due to dynamic allocation of pod IPs in k8s.  i.e. Pod IPs different on your system than on the Travis system.  This breaks the headless services tests, since headless services answer with the pod IPs for the service (they don't have a virtual cluster IP).

You could replace the "expected" results for the headless services with a function call that looks up the endpoints (pods), for the service by escaping to shell and running `kubectl` ... or by connecting ot the API (`kubectl` probably easier).  For example

```
Answer: getEndpoints("headless-svc", "test-1")

```

where getEndpoints returns `[]dns.RR` ... one for each endpoint under the service/namespace passed.

chrisohaver:
Some `kubectl` magic to get the IPs ...

`kubectl -n test-1 get endpoints headless-svc -o 'jsonpath={.subsets[*].addresses[*].ip}'`

test/kubernetes_test.go
```diff
@@ -28,189 +26,256 @@ var dnsTestCases = []test.Case{
 		Qname: "bogusservice.test-1.svc.cluster.local.", Qtype: dns.TypeA,
 		Rcode:  dns.RcodeNameError,
 		Answer: []dns.RR{},
+		Ns: []dns.RR{
+			test.SOA("cluster.local.	300	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1502313310 7200 1800 86400 60"),
+		},
 	},
 	{
 		Qname: "bogusendpoint.svc-1-a.test-1.svc.cluster.local.", Qtype: dns.TypeA,
```
miekg: Can we also add a little comment above each of these of *what* it actually tests?

test/kubernetes_test.go
```diff
@@ -415,18 +504,355 @@ var dnsTestCasesAllNSExposed = []test.Case{
...
+	{
+		Qname: "bogusservice.*.svc.cluster.local.", Qtype: dns.TypeA,
+		Rcode:  dns.RcodeNameError,
+		Answer: []dns.RR{},
+		Ns: []dns.RR{
+			test.SOA("cluster.local.	3600	IN	SOA	sns.dns.icann.org. noc.dns.icann.org. 2015082541 7200 3600 1209600 3600"),
+		},
+	},
+	{
+		Qname: "bogusservice.any.svc.cluster.local.", Qtype: dns.TypeA,
```
miekg: can we also test longer names than this?

test/kubernetes_test.go
```diff
@@ -415,18 +504,355 @@ var dnsTestCasesAllNSExposed = []test.Case{
 
 var dnsTestCasesFallthrough = []test.Case{
 	{
+		Qname: "svc-1-a.test-1.svc.cluster.local.", Qtype: dns.TypeA,
```
miekg, line 362: What new code paths are tested by these? Do the increase the go coverage?

miekg (changes requested):
How are the test results established, just running kube-dns and using those results? Or running CoreDNS?

I rather see that this PR also *decreases* the number of tests while keeping the same coverage. `go test -coverage` is a very useful tool for that.

miekg:
We should probably also check other record types, HINFO (to name one) and NS for instance.

test/kubernetes_test.go
chrisohaver: The added test case are not new test cases, they are test case that have two different answers based on the coredns configuration.

chrisohaver:
> We should probably also check other record types, HINFO (to name one) and NS for instance.

Sure, In interest to keep this PR from running away, open a issue to add those.

test/kubernetes_test.go
chrisohaver: Yes - perhaps we should add an issue to expand test cases in integration tests.  This PR is focused on adding response record validation.

miekg:
Ok, but just in the interested of keeping the amount of work done low - we might want to cut back on the amount now, instead of fixing each and everyone right now and then deleting a bunch of them later?

chrisohaver:
> How are the test results established, just running kube-dns and using those results? Or running CoreDNS?

> I rather see that this PR also decreases the number of tests while keeping the same coverage. go 

Where are you seeing that?

> test -coverage is a very useful tool for that.

I think it's `go test -cover` ...

But not sure what value actually counts coverage for in this case.  It's an integration test that in some cases runs multiple instances of coredns (one main, and one upstream server).   It looks like its only counting coverage for the setting up of the test and the sending of queries.  The actually tests cases are executed against instances of coredns, which are not tracked by test coverage... So as is, the coverage numbers are useless here.  As far as I can tell.

chrisohaver:
> Ok, but just in the interested of keeping the amount of work done low - we might want to cut back on the amount now, instead of fixing each and everyone right now and then deleting a bunch of them later?

Why would we be deleting tests?

miekg:
This tests establishes that any and * are equivalent:
~~~
svc-1-a.*.svc.cluster.local.
svc-1-a.any.svc.cluster.local.
~~~
What does this test test?
~~~
bogusservice.*.svc.cluster.local.
bogusservice.any.svc.cluster.local.
~~~

We should delete tests that don't add any coverage

chrisohaver:
> We should delete tests that don't add any coverage

Ideally - yes.  The risk is removing tests that we think don't add coverage, but actually do.

We can remove tests where its obvious - i.e. redundant wildcard checks.

miekg:
Indeed. That's why adding a rational above each test as comment would make sense. It forces you to think about these issues. More tests != better testing.

Also: one big question that I have is left unanswered: how we do validate the results? Is this compared with kube-dns? Do we have a doc outlining the differences and why ?

chrisohaver:
> Also: one big question that I have is left unanswered: how we do validate the results? Is this compared with kube-dns? 

They are validated per the k8s spec, which kube-dns also follows.
However, some functions, such as wildcards, don't have kube-dns equivalents.

> Do we have a doc outlining the differences and why ?

No.

miekg:
> They are validated per the k8s spec, which kube-dns also follows.
> However, some functions, such as wildcards, don't have kube-dns equivalents.

Interesting, do you have example of those?

chrisohaver:
> Interesting, do you have example of those?

Any feature we support but kube-dns does not.  I thought wildcard queries was one... perhaps I am mistaken.  Queries that use features like autopath, pods verified mode, fall into the same category.
I think fallthrough also.  However I think as of k8s 1.6, kube-dns supports fallthrough-like behavior.

johnbelamaric:
I don't think they do `fallthrough` like behavior, but they do allow you to configure stub zones in the dnsmasq configuration, which works similarly to doing different server blocks for those domains to go to different upstreams in CoreDNS.

test/kubernetes_test.go
```diff
@@ -351,16 +696,40 @@ func doIntegrationTests(t *testing.T, corefile string, testCases []test.Case) {
...
+
+			}
 		}
-
-		if len(res.Answer) != len(tc.Answer) {
-			t.Errorf("Expected %d answers but got %d for query %s, %d", len(tc.Answer), len(res.Answer), tc.Qname, tc.Qtype)
+		sort.Sort(test.RRSet(res.Answer))
+		sort.Sort(test.RRSet(res.Ns))
+		sort.Sort(test.RRSet(res.Extra))
+		sort.Sort(test.RRSet(tc.Answer))
```
miekg: I don't think we need to sort the testcases as well.

test/kubernetes_test.go
```diff
@@ -479,8 +848,70 @@ func TestKubernetesIntegrationFallthrough(t *testing.T) {
...
+		cmd    []byte
+		cmdOut []byte
+		err    error
+	)
+	kctl := os.Getenv("$KUBECTL")
+	if cmd, err = exec.Command("env").Output(); err != nil {
+		println(os.Stderr, "There was an error running kubectl command: ", err.Error())
+	}
+	cmd_str := string(cmd)
+	println(cmd_str)
```
miekg: Left over debugging?

test/kubernetes_test.go
```diff
@@ -295,25 +354,311 @@ var dnsTestCasesAllNSExposed = []test.Case{
...
+		Answer: []dns.RR{
+			test.CNAME("ext-svc.test-1.svc.cluster.local. 0 IN	CNAME	example.net."),
+			test.A("example.net.		3600	IN	A	13.14.15.16"),
+		},
+	},
+	{
+		Qname: "ext-svc.test-1.svc.cluster.local.", Qtype: dns.TypeCNAME,
+		Rcode: dns.RcodeSuccess,
+		Answer: []dns.RR{
+			test.CNAME("ext-svc.test-1.svc.cluster.local. 0 IN	CNAME	example.net."),
```
miekg, line 636: This TTL is really low... Is that intentional?

test/kubernetes_test.go
```diff
@@ -228,22 +285,15 @@ var dnsTestCases = []test.Case{
...
 		Answer: []dns.RR{
-			test.TXT("dns-version.cluster.local. 28800 IN TXT \"1.0.0\""),
-		},
-	},
-	{
-		Qname: "cluster.local.", Qtype: dns.TypeNS,
-		Rcode: dns.RcodeSuccess,
-		Answer: []dns.RR{
-			test.NS("cluster.local.          0       IN      NS      kubernetes.default.svc.cluster.local."),
+			test.TXT("dns-version.cluster.local. 28800 IN TXT \"1.0.1\""),
```
miekg: If you use "`" (backtacks) you are use raw strings in Go, no need to escape the " in the TXT rdata.

test/kubernetes_test.go
chrisohaver: I updated the title with WIP, to reflect that this is a work in progress.

Sandeep is debugging a travis vs local environment issue at the moment, so he is committing changes to the PR to debug on travis.  i.e. yes - it is debug...

test/kubernetes_test.go
```diff
@@ -34,189 +36,244 @@ var dnsTestCases = []test.Case{
...
 		Qname: "*.svc-1-a.test-1.svc.cluster.local.", Qtype: dns.TypeSRV,
 		Rcode:  dns.RcodeNameError,
 		Answer: []dns.RR{},
+		Ns: []dns.RR{
+			test.SOA("cluster.local.	300	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1502313310 7200 1800 86400 60"),
+		},
 	},
 	{
 		Qname: "*._not-udp-or-tcp.svc-1-a.test-1.svc.cluster.local.", Qtype: dns.TypeSRV,
 		Rcode:  dns.RcodeNameError,
```
miekg: This Rcode is wrong; this should be dns.RcodeSucccess and the message reply should indicate a nodata reponse.

miekg (changes requested):
Yes, I think there is no other choice than `getDynamicEndpoints` and calling `kubectl`. I do note however that we are pinning ourselves to specific `kubect` output which may lead to a lot of churn in the future.

chrisohaver:
>  I do note however that we are pinning ourselves to specific kubect output which may lead to a lot of churn in the future.

Not really - the command gets json data (which is the same fields as the API).  So not nearly as bad as parsing human readable command line output.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: Middlewa..." ]
>>  I do note however that we are pinning ourselves to specific kubect output which may lead to a lot of churn in the future.
>
>Not really - the command gets json data (which is the same fields as the API).  So not nearly as bad as parsing human readable command line output.

Ah cool, I stand correct. nice!

test/kubernetes_test.go
chrisohaver: just set the TTL to 303...

chrisohaver: Actually - most of these should be 303... (which I believe is secret handshake for dont check the TTL)...

We might want to put more thought into what the TTLs should be later.

test/kubernetes_test.go
```diff
@@ -351,16 +696,40 @@ func doIntegrationTests(t *testing.T, corefile string, testCases []test.Case) {
 			t.Fatalf("Could not send query: %s", err)
 		}
 
-		// check the answer
-		if res.Rcode != tc.Rcode {
-			t.Errorf("Expected rcode %d but got %d for query %s, %d", tc.Rcode, res.Rcode, tc.Qname, tc.Qtype)
+		// Before sorting, make sure that CNAMES do not appear after their target records
```
chrisohaver: This can be made into a function and put in test.  e.g. test.CNAMEOrder(res.Answer)

test/kubernetes_test.go
rajansandeep: That means we have to sort the testcases manually?

test/kubernetes_test.go
chrisohaver: This CNAME answer order check can be made into a function and put in middleware/test/helpers.go. e.g. test.CNAMEOrder(res.Answer).

chrisohaver (changes requested):

middleware/test/helpers.go
```diff
@@ -264,7 +264,7 @@ func Section(t *testing.T, tc Case, sec sect, rr []dns.RR) bool {
 }
 
 // Function to make sure that CNAMES do not appear after their target records
```
chrisohaver: It's good practice to make the comment before a method start with the method name.  For example:

`// CNAMEOrder makes sure that CNAMES do not appear after their target records`

chrisohaver (changes requested):

chrisohaver (approved):

test/kubernetes_test.go
miekg: Yes 303 is special, obscurely documented some where :-)

miekg:
I wanted to raise the following point
With this PR we moving away from having our test data in version control, as there is now this dynamic component. This is a *big* change that surely impacts debugging.

Why does this need to be dynamic? Because of the difference between minikube and Travis? Why are those different? How can we make them less different?

chrisohaver:
> With this PR we moving away from having our test data in version control, as there is now this dynamic component. 

The only dynamic aspect is the IP address in the response records.  The other parts of those response records are hard coded (i.e. in version control). So, yes, the IPs are dynamic, but all other aspects of the response are not.

>Why does this need to be dynamic? Because of the difference between minikube and Travis? Why are those different? How can we make them less different?

The dynamic nature is not a difference between minikube and hyperkube.  In this regard, they both behave the same way.  The reason for the dynamic assignment is due to how kubernetes assigns IPs to pods.  Kubernetes (not just minikube or hyperkube) assigns IPs to pods dynamically from a pool.

I think retrieving the IPs for the pods is the right approach.  Nevertheless, some alternatives...
* Reserve a special IP to mean "don't compare this IP" ... similar to TTL 303. Then we can skip comparing the IP in these dynamic records.  This is the same as not testing the IPs at all. It provides less coverage.
* Hard code the IPs in the tests to the answers we get in the Travis environment.  Since we are nuking the environment every time, the IPs might be the same every time.  However, this isn't a guarantee - we would be relying on race conditions.  Also, it would cause test failures in local instances, since those systems will have other pods for testing (the pool isn't the same state).

miekg:
Thanks for the update; yes those alternative are bad as well. With ample t.Logf this is probably the best we can do.

test/kubernetes_test.go
```diff
@@ -351,16 +697,31 @@ func doIntegrationTests(t *testing.T, corefile string, testCases []test.Case) {
...
+			t.Logf("%v\n", res)
 		}
 
-		if len(res.Answer) != len(tc.Answer) {
-			t.Errorf("Expected %d answers but got %d for query %s, %d", len(tc.Answer), len(res.Answer), tc.Qname, tc.Qtype)
-		}
+		sort.Sort(test.RRSet(res.Answer))
+		sort.Sort(test.RRSet(res.Ns))
+		sort.Sort(test.RRSet(res.Extra))
+		sort.Sort(test.RRSet(tc.Answer))
```
miekg: tc does not need to be sorted. Unless we add dynamic ip but even then we could sort it in the go code calling kubectl

test/kubernetes_test.go
```diff
@@ -479,8 +840,63 @@ func TestKubernetesIntegrationFallthrough(t *testing.T) {
...
 `
-	cases := append(dnsTestCases, dnsTestCasesFallthrough...)
-	doIntegrationTests(t, corefile, cases)
+	doIntegrationTests(t, corefile, dnsTestCasesFallthrough)
+}
+
+func getDynamicEndpoints(Qname, namespace, name string) []dns.RR {
+	rr := []dns.RR{}
+
+	result := strings.Split(getCmdOutput(name, namespace), " ")
```
miekg: Go style is usualy to omit **get**, so just `cmdOutput`

test/kubernetes_test.go
```diff
@@ -479,8 +840,63 @@ func TestKubernetesIntegrationFallthrough(t *testing.T) {
...
+	len_result := len(result)
```
miekg: Err... @chrisohaver told me we parse the json from kubectl, here we just string split the output?

test/kubernetes_test.go
```diff
@@ -479,8 +840,63 @@ func TestKubernetesIntegrationFallthrough(t *testing.T) {
...
+		cmdOut []byte
+		err    error
+	)
+	kctl := os.Getenv("KUBECTL")
+	if kctl == "" {
+		kctl = "kubectl"
+	}
+	cmdArgs := kctl + " -n " + name + " get endpoints " + namespace + " -o jsonpath={.subsets[*].addresses[*].ip}"
+	if cmdOut, err = exec.Command("sh", "-c", cmdArgs).Output(); err != nil {
+		println(os.Stderr, "There was an error running kubectl command: ", err.Error())
```
miekg: Don't use `println`: return the error and neatly log the failures from the caller so it is clear what went wrong.

test/kubernetes_test.go
miekg: I should probably also run
`cmdArgs := kctl + " -n " + name + " get endpoints " + namespace + " -o jsonpath={.subsets[*].addresses[*].ip}"`
to see what this output, 'cause I don't now what happens if you specify a json path

test/kubernetes_test.go
```diff
@@ -479,8 +840,63 @@ func TestKubernetesIntegrationFallthrough(t *testing.T) {
 	drop 0
     }
 `
-	cases := append(dnsTestCases, dnsTestCasesFallthrough...)
-	doIntegrationTests(t, corefile, cases)
+	doIntegrationTests(t, corefile, dnsTestCasesFallthrough)
+}
+
+func getDynamicEndpoints(Qname, namespace, name string) []dns.RR {
```
miekg: Document this function was some example inputs

test/kubernetes_test.go
```diff
@@ -479,8 +840,63 @@ func TestKubernetesIntegrationFallthrough(t *testing.T) {
...
+	rr := []dns.RR{}
+
+	result := strings.Split(getCmdOutput(name, namespace), " ")
+	len_result := len(result)
+
+	if len_result > 1 {
```
miekg: Underscore names is not idiomatic go, just use `l`, or `lr`

test/kubernetes_test.go
```diff
@@ -479,8 +840,63 @@ func TestKubernetesIntegrationFallthrough(t *testing.T) {
...
+
+func getSRVrecords(Qname, Qtype, namespace, name string) []dns.RR {
+	rr := []dns.RR{}
+
+	result := strings.Split(getCmdOutput(name, namespace), " ")
+	len_result := len(result)
+
+	if len_result > 1 {
+		for i := 0; i < len_result; i++ {
+			ip := strings.Replace(result[i], ".", "-", -1)
```
miekg: don't we have a kubernetes helper function for this?

test/kubernetes_test.go
```diff
@@ -479,8 +840,63 @@ func TestKubernetesIntegrationFallthrough(t *testing.T) {
...
+			rr = append(rr, test.A(Qname+"    303    IN      A   "+result[i]))
+		}
+	} else {
+		rr = append(rr, test.A(Qname+"    303    IN      A   "+result[0]))
+	}
+
+	return rr
+}
+
+func getSRVrecords(Qname, Qtype, namespace, name string) []dns.RR {
```
miekg: Document this function and drop the `get`

test/kubernetes_test.go
```diff
@@ -479,8 +840,63 @@ func TestKubernetesIntegrationFallthrough(t *testing.T) {
...
+	rr := []dns.RR{}
+
+	result := strings.Split(getCmdOutput(name, namespace), " ")
+	len_result := len(result)
+
+	if len_result > 1 {
```
miekg: Revert the flow here:
`len_result == 1 {
 // ip stuff
return
}
// srv stuff

test/kubernetes_test.go
miekg: Short documentation for the function would help as well.

test/kubernetes_test.go
```diff
@@ -479,8 +840,63 @@ func TestKubernetesIntegrationFallthrough(t *testing.T) {
...
+	kctl := os.Getenv("KUBECTL")
+	if kctl == "" {
+		kctl = "kubectl"
+	}
+	cmdArgs := kctl + " -n " + name + " get endpoints " + namespace + " -o jsonpath={.subsets[*].addresses[*].ip}"
+	if cmdOut, err = exec.Command("sh", "-c", cmdArgs).Output(); err != nil {
+		println(os.Stderr, "There was an error running kubectl command: ", err.Error())
+		os.Exit(1)
+	}
+	return string(cmdOut)
```
miekg: The `bytes` pkg has almost all operations the string package has, there is no need to convert this into a `string`.

test/kubernetes_test.go
chrisohaver: > @chrisohaver told me we parse the json from kubectl, here we just string split the output?

It's splitting output produced by the json path filter.

> I don't now what happens if you specify a json path

It depends on what path you give it. In this case, the path will list all IPs for each endpoint.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: Middlewa..." ]
>
>
>
>> @@ -479,8 +840,63 @@ func TestKubernetesIntegrationFallthrough(t *testing.T) {
> 	drop 0
>     }
> `
>-	cases := append(dnsTestCases, dnsTestCasesFallthrough...)
>-	doIntegrationTests(t, corefile, cases)
>+	doIntegrationTests(t, corefile, dnsTestCasesFallthrough)
>+}
>+
>+func getDynamicEndpoints(Qname, namespace, name string) []dns.RR {
>+	rr := []dns.RR{}
>+
>+	result := strings.Split(getCmdOutput(name, namespace), " ")
>+	len_result := len(result)
>
>> @chrisohaver told me we parse the json from kubectl, here we just string split the output?
>
>It's splitting output produced by the json path filter.
>
>> I don't now what happens if you specify a json path
>
>It depends on what path you give it. In this case, the path will list all IPs for each endpoint.

Yeah, updated my comment. This function can however be improved with some 
documentation.

test/kubernetes_test.go
rajansandeep: What is the most idiomatic way to report an error when you're not in a test. This error would occur during the declaration of the global variable.

chrisohaver:
I think more descriptive names for CmdOutput, SRVRecords, and DynamicEndpoints could be...

`kubectlGetEndpointIPs` ... or maybe just `endpointIPs`
`headlessAResponse`
`headlessSRVResponse`

They are longer, but more descriptive of what the methods do.
Also, I don't think the methods need to be exported, so they should start with lower case letters.

test/kubernetes_test.go
miekg: You "bubble up" the error to the caller, which may handle the error, but calling log.Fatal or t.Fatal. So in this case the function should return a tuple of `[]byte, errror` and not log anything by itself.

test/kubernetes_test.go
```diff
@@ -479,8 +827,65 @@ func TestKubernetesIntegrationFallthrough(t *testing.T) {
 	drop 0
     }
 `
-	cases := append(dnsTestCases, dnsTestCasesFallthrough...)
-	doIntegrationTests(t, corefile, cases)
+	doIntegrationTests(t, corefile, dnsTestCasesFallthrough)
+}
+
+//headlessAResponse returns the answer to an A request for the specific name and namespace
```
miekg, line 829: Thanks for adding comments. You don't need to formally describe the input params this should be obvious from the function signature. Function comments should read like a short story. I.e. `godoc .` in this dir should be a "nice read"

miekg (commented):
This is starting to look good. A few small things and then this can be merged.

rajansandeep:
One integration test is failing due to issue #808

chrisohaver:
I think the `SRV` test for `*.svc-1-a.test-1.svc.cluster.local.` has the wrong expected results.  It should _not_ be expecting a `NODATA`, rather it should be expecting a `NOERROR`, with the all of the SRV records for that service.  All of the following `SRV` queries should have the same answer...
* `svc-1-a.test-1.svc.cluster.local.` 
* `*.svc-1-a.test-1.svc.cluster.local.`
* `*.*.svc-1-a.test-1.svc.cluster.local.`

...One minor difference being the question in each answer record would match the original question.... but all other parts should be identical.

chrisohaver:
> I think the `SRV` test for `*.svc-1-a.test-1.svc.cluster.local.` has the wrong expected results...

That said, even after correcting the results, the test will still fail, due to #808

middleware/test/helpers.go
```diff
@@ -291,6 +291,25 @@ func SortAndCheck(t *testing.T, resp *dns.Msg, tc Case) {
 	return
 }
 
+// CNAMEOrder makes sure that CNAMES do not appear after their target records
+func CNAMEOrder(t *testing.T, res *dns.Msg) bool {
```
miekg: Why does this return a boolean as the t.Errorf is already done in the function?

test/kubernetes_test.go
```diff
@@ -351,17 +697,19 @@ func doIntegrationTests(t *testing.T, corefile string, testCases []test.Case) {
...
-			t.Errorf("Expected rcode %d but got %d for query %s, %d", tc.Rcode, res.Rcode, tc.Qname, tc.Qtype)
+		// Before sorting, make sure that CNAMES do not appear after their target records
+		if !test.CNAMEOrder(t, res) {
+			t.Logf("%v\n", res)
 		}
 
-		if len(res.Answer) != len(tc.Answer) {
-			t.Errorf("Expected %d answers but got %d for query %s, %d", len(tc.Answer), len(res.Answer), tc.Qname, tc.Qtype)
-		}
+		sort.Sort(test.RRSet(tc.Answer))
```
miekg: Can you add a comment here on why this is needed?

miekg (approved):
Left two small nits.

test/kubernetes_test.go
```diff
@@ -250,8 +250,11 @@ var dnsTestCases = []test.Case{
 	},
 	{
 		Qname: "*.svc-1-a.test-1.svc.cluster.local.", Qtype: dns.TypeSRV,
-		Rcode:  dns.RcodeSuccess,
-		Answer: []dns.RR{},
+		Rcode: dns.RcodeSuccess,
+		Answer: []dns.RR{
+			test.SRV("svc-1-a.test-1.svc.cluster.local.	303	IN	SRV	0 50 443 svc-1-a.test-1.svc.cluster.local."),
```
chrisohaver: The answers should (I think), have the question in them ... e.g. 
`test.SRV("svc-1-a.test-1.svc.cluster.local.	303	IN	SRV	0 50 443 svc-1-a.test-1.svc.cluster.local.")`
should instead be ...
`test.SRV("*.svc-1-a.test-1.svc.cluster.local.	303	IN	SRV	0 50 443 svc-1-a.test-1.svc.cluster.local.")`

miekg:
What's the status here? There are a few conflicts in kubernetes.go that don't look too bad. 

I also want to tackle https://github.com/coredns/coredns/issues/942 but that should wait on this PR other it will create a whole bunch of conflicts

chrisohaver:
> What's the status here? 

Approved, waiting for you to merge. But looks like the PR has to be rebased again.

miekg:
Is this failure the issue @chrisohaver just opened? #975 ?
~~~
--- FAIL: TestKubernetesIntegration (3.03s)
	helpers.go:179: rr 0 should have a Target of "svc-1-a.test-1.svc.cluster.local.", but has "172-17-0-4.svc-1-a.test-1.svc.cluster.local."
	helpers.go:296: ;; opcode: QUERY, status: NOERROR, id: 5997
		;; flags: qr aa rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1
		
		;; QUESTION SECTION:
		;*.svc-1-a.test-1.svc.cluster.local.	IN	 SRV
		
		;; ANSWER SECTION:
		*.svc-1-a.test-1.svc.cluster.local.	0	IN	SRV	0 50 443 172-17-0-4.svc-1-a.test-1.svc.cluster.local.
		*.svc-1-a.test-1.svc.cluster.local.	0	IN	SRV	0 50 80 172-17-0-4.svc-1-a.test-1.svc.cluster.local.
		
		;; ADDITIONAL SECTION:
		172-17-0-4.svc-1-a.test-1.svc.cluster.local.	0	IN	A	172.17.0.4
~~~

chrisohaver:
> Is this failure the issue @chrisohaver just opened?

Yes, to the 2nd item in #975. Actually coredns responds with an `NXDOMAIN` which breaks this test. The same way kube-dns responds.  And the test used to expect `NXDOMAIN` ...

chrisohaver:
> Yes, to the 2nd item in #975. 

No, I'm wrong, related to, but not caused by the 3rd Item in #975...

But the failure is that we are getting a wrong answer here.  The answer and extra sections point to the service endpoint (wrong IP ).

chrisohaver:
Ah - i see whats going on...  There is query ambiguity..

The * in that position can either mean "any `_protocol`"  or "any `endpoint`" ... but we have no way of knowing what the intent was.

The expected value in the test is the answer for "all `_protocol`" in service `svc-1-a`
The coredns response is for "all `endpoint`" in service `svc-1-a`

So - the coredns behavior may be correct - if we decide thats how we want it to behave.

This is why kube-dns answers queries like `_tcp.svc-1-a .test-1.svc.cluster.local` as `NXDOMAIN`... because it's treading the `_tcp` as an endpoint... which is the correct answer.

miekg:
> This is why kube-dns answers queries like _tcp.svc-1-a .test-1.svc.cluster.local as NXDOMAIN... because it's treading the _tcp as an endpoint... which is the correct answer.

Ok we should change that then.

chrisohaver:
@miekg , OK to merge?

miekg:
Yes, if there are no failing tests.

Issue #881, opened 2017-08-10
dnsserver.GetMiddleware considered harmful


miekg:
If you call `dnsserver.GetMiddleware` the compiled chain is foobar... (I've seen this before, moved on, but now I have a use case with the *autopath* middleware).

*autopath* works when you **don't** call GetMiddleware in the setup, but fails otherwise; you get SERVFAIL instead of autopathing.

The middleware is still called, but somehow it has been reset. This `println`, prints `A.Next` in `ServeDNS` of autopath (when GetMiddleware is called)
~~~
autopath
(0x0,0x0) # Next is cleared
~~~
Without GetMiddleware:
~~~
autopath
(0x20d1ce0,0xc420570d80)
~~~
I.e. ap.Next *is* set here.

Now the strange thing is (I don't understand this), that GetMiddleware *does not* write to anything it just return a middleware.Middleware if found...

miekg:
The *one* thing GetMiddleware does is calling the middleware with `nil` to get the handler for which the name method is defined.

miekg:
commenting *that* call out, makes it work:
~~~
		h = h
		/*
			x := h(nil)
			if name == x.Name() {
				return x
			}
		*/
~~~

miekg:
Code-link in master: https://github.com/coredns/coredns/blob/master/core/dnsserver/config.go#L61

miekg:
Discussion on slack: The `h(nil`) calls the `AddMiddleware` with a nill handler, which get's it added.
The short term solution is to have another registry for middleware that want to talk to each other and just register the middleware's *Handler* there.
This means we can drop `Name()` as well, but have an ugly second registration step. This probalby needs to be per-server meaning it changes the semantics of `GetMiddleware` a bit... (I think)

PR #885, opened 2017-08-10
core: replace GetMiddleware

miekg:
See the discussion in #881. GetMiddleware would add a `nil` middleware
to the callstack thereby breaking functionality.

This PR drops it in favor of RegisterHandler which is a completely
standalone registry for middleware that want to let it self know to
other middleware.

Currenly *autopath* uses this to call *kubernetes*'s AutoPath method
for dynamic autopathing.

### DIFF
--- a/core/dnsserver/register.go
+++ b/core/dnsserver/register.go
@@ -124,6 +124,49 @@ func (c *Config) AddMiddleware(m middleware.Middleware) {
 	c.Middleware = append(c.Middleware, m)
 }
 
+// RegisterHandler adds a handler to a site's handler registration. Handlers
+// should use this if the want to announce that they exist to other middleware.
+func (c *Config) RegisterHandler(h middleware.Handler) {
+	if c.Registry == nil {
+		c.Registry = make(map[string]middleware.Handler)
+	}
+
+	if _, ok := c.Registry[h.Name()]; ok {
+		panic("foot!")
+	}
+	c.Registry[h.Name()] = h
+}
+
+// GetHandler returns the middleware handler that has been added to the config under its name.
+// This is useful to inspect if a certain middleware is active in this server.
+// Note that this is order dependent and the order is defined in directives.go, i.e. if your middleware
+// comes before the middleware you are checking; it will not be there (yet).
+// See RegisterHandler on how to register the middleware with this server.
+func (c *Config) GetHandler(name string) middleware.Handler {
+	if c.Registry == nil {
+		return nil
+	}
+	if h, ok := c.Registry[name]; ok {
+		return h
+	}
+	return nil
+}
+
+// GetMiddleware returns the middleware handler that has been added to the config under name.
+// This is useful to inspect if a certain middleware is active in this server.
+// Note that this is order dependent and the order is defined in directives.go, i.e. if your middleware
+// comes before the middleware you are checking; it will not be there (yet).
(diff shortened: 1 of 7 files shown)


miekg:
@johnbelamaric as discussed - a possible fix. Need another registration call, which is ugly, but couldn't figure out something nicer.

johnbelamaric:
Couldn't there just be a Next() function on the Middleware or Handler? Or maybe I am missing something.

miekg:
But you still need to get to the handler in the current setup. That can
only done by calling the middleware.Middleware function.

middleware.Middleware is not an interface...

A said, not a fan of this pr, but current code is sadly a bit broken.

core/dnsserver/register.go
```diff
@@ -124,6 +124,49 @@ func (c *Config) AddMiddleware(m middleware.Middleware) {
...
 
+// RegisterHandler adds a handler to a site's handler registration. Handlers
+// should use this if the want to announce that they exist to other middleware.
+func (c *Config) RegisterHandler(h middleware.Handler) {
+	if c.Registry == nil {
+		c.Registry = make(map[string]middleware.Handler)
+	}
+
+	if _, ok := c.Registry[h.Name()]; ok {
+		panic("foot!")
```
johnbelamaric: We can have multiple instances of the same mw in a given chain (e.g., rewrite). Not an issue now but maybe that's something we need to deal with here.

core/dnsserver/register.go
```diff
@@ -124,6 +124,49 @@ func (c *Config) AddMiddleware(m middleware.Middleware) {
...
+		return h
+	}
+	return nil
+}
+
+// GetMiddleware returns the middleware handler that has been added to the config under name.
+// This is useful to inspect if a certain middleware is active in this server.
+// Note that this is order dependent and the order is defined in directives.go, i.e. if your middleware
+// comes before the middleware you are checking; it will not be there (yet).
+func GetMiddleware(c *caddy.Controller, name string) middleware.Handler {
```
johnbelamaric: Why do we need this still?

miekg:
Haven't removed the old code yet. Wanted Ack first.

Re: multiple instances: didn't we already have that problem? You would get
the first I think. Now you would get the last....

On Aug 10, 2017 17:47, "John Belamaric" <notifications@github.com> wrote:

*@johnbelamaric* commented on this pull request.
------------------------------

In core/dnsserver/register.go
<https://github.com/coredns/coredns/pull/885#discussion_r132507619>:

> +// See RegisterHandler on how to register the middleware with this server.
+func (c *Config) GetHandler(name string) middleware.Handler {
+	if c.Registry == nil {
+		return nil
+	}
+	if h, ok := c.Registry[name]; ok {
+		return h
+	}
+	return nil
+}
+
+// GetMiddleware returns the middleware handler that has been added
to the config under name.
+// This is useful to inspect if a certain middleware is active in this server.
+// Note that this is order dependent and the order is defined in
directives.go, i.e. if your middleware
+// comes before the middleware you are checking; it will not be there (yet).
+func GetMiddleware(c *caddy.Controller, name string) middleware.Handler {

Why do we need this still?

—
You are receiving this because you authored the thread.
<https://github.com/coredns/coredns/pull/885#pullrequestreview-55596152>,
or mute the thread

johnbelamaric:
Ok. Looks good in general. But my point on multiple instances was that we shouldn't panic in that case.

miekg:
A yes, good point.

PR #889, opened 2017-08-10
mw/k8s: compress the README

miekg:
Remove the examples, save one in the syntax section. The example
section covers all.

This brings the kubernetes README inline with other READMEs.


miekg:
yeah! make README smaller increase test coverage!

PR #899, opened 2017-08-11
Push parseRequest into k.Entries()

miekg:
This is only needed in that function.


miekg:
This fails because the answer is:
~~~
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 4, AUTHORITY: 0, ADDITIONAL: 4
		
		;; QUESTION SECTION:
		;_c-port._UDP.*.test-1.svc.cluster.local.	IN	 SRV
		
		;; ANSWER SECTION:
		_c-port._UDP.*.test-1.svc.cluster.local.	0	IN	SRV	0 25 1234 172-17-0-6.headless-svc.test-1.svc.cluster.local.
		_c-port._UDP.*.test-1.svc.cluster.local.	0	IN	SRV	0 25 1234 172-17-0-7.headless-svc.test-1.svc.cluster.local.
		_c-port._UDP.*.test-1.svc.cluster.local.	0	IN	SRV	0 25 0 example.net.
		_c-port._UDP.*.test-1.svc.cluster.local.	0	IN	SRV	0 25 1234 svc-c.test-1.svc.cluster.local.
		
		;; ADDITIONAL SECTION:
		172-17-0-6.headless-svc.test-1.svc.cluster.local.	0	IN	A	172.17.0.6
		172-17-0-7.headless-svc.test-1.svc.cluster.local.	0	IN	A	172.17.0.7
		example.net.	3600	IN	A	13.14.15.16
		svc-c.test-1.svc.cluster.local.	0	IN	A	10.0.0.115
~~~
Assuming the `example.net` is extra record that gets added.
@chrisohaver @johnbelamaric : I see nothing wrong in this?

chrisohaver:
If example.net has the "port" service running on port "0", then the answer is correct. Otherwise it's wrong.

chrisohaver:
Or put differently: If example.net has an SRV record for `_port._udp.example.net.` on port 0, then the answer is correct...

miekg:
I see. Real world messing with my Zen. I'll put the HostType check back in

miekg:
So how does this end up in the k8s master?

miekg:
Inclined to change the test cases - as re-adding this bring back the CNAME query for federation... which is exactly why we shouldn't care about type of query, except for nodata/nxdomain responses.

Issue #907, opened 2017-08-12
mw/kubernetes: federation refactoring


miekg:
There only reason that `parseRequest` is still a method on `Kubernetes` is because of a possibility of federation. `k.Entries` also sneaks in a federation check, when nothing is found.

We should make the federation flow completely separate from the normal resolution flow. We can detect  if it is a federation query by checking the federation label in early in `k.ServeDNS` and call a possible `federation.ServeDNS()` that handles the federation.

There is also little reason the federation function need to be methods in `Kubernetes`, 
* `federationCNAMERecord` only needs the federation records and a `node.API`, this `node.API` should be parameter to the function
* `stripFederation()` only need the federation to work.

Ideally the whole thing must be put in a federation sub package, except it can't because of the use of `recordRequest` which then should also be put in the subpackage, leading to a cascade of small package that not improve readability.

miekg:
Motivation for putting federation in its own package is to have cleaner, shorter names for the methods.

yongtang:
👍 For separation of federation from main kubernetes usage. Federation itself is not a necessary component of kubernetes and there are lot of users want to get around multi-cluster deployment with other means. No reason to mixture them together.

miekg:
pulling this out into a separate middleware? That would even be better IMO (if possible - though I think it should, look at *autopath* that just calls one function in kubernetes).

miekg:
https://github.com/coredns/coredns/pull/914 (framework for new middleware)

miekg:
This has been merged.

PR #908, opened 2017-08-12
WIP: Federation (don't merge)

miekg:


miekg:
This is what I mean with https://github.com/coredns/coredns/issues/907. In this PR (fails tests) all federation stuff in handled from `federation.go` and the other code can just deal with retrieving services from the k8s API.

There numerous cleanups and bugfixes left though.

/cc @chrisohaver

miekg:
Keeping open for some copy and pasting, but we should make federation a separate middleware, see #914 for the proper fix.

chrisohaver:
> we should make federation a separate middleware

Can you expand on why you think this?

PR #912, opened 2017-08-13
Consolidation of Makefile and Makefile.release

yongtang:
Several changes:
1. All go specific target like `go generate`, etc. has
been moved to Makefile. Now Makefile.release does not
repeat go build, etc. related rules.
2. In Makefile.release, the binary build is done through
`docker run` and with a fixed specific go version (currently 1.8.3).
This will help making sure build process could be reproduced
on any dev platform, with no dependency to the golang version
installed on the dev machine used.
3. Platform related flags (e.g., "GOOS=darwin") are passed through
Makefile (not Makefile.release).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,6 @@
 GITCOMMIT:=$(shell git describe --dirty --always)
+BINARY:=coredns
+SYSTEM:=
 
 all: coredns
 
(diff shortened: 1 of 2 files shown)


miekg:
+1 on the cleanup, but now releasing depends on docker and with docker automated builds we were moving away from this? I.e. For me personally, I tend to not have docker installed on my systems...

yongtang:
Updated. The dependency on `docker run` has been removed.

PR #914, opened 2017-08-13
mw/federation: add

miekg:
Move federation out of the kubernetes middleware.


miekg:
My local gofmt thinks everything is fine... *sigh*

miekg:
Hmm, runing gofmt -w -s *.go did change the file; goimports may be old?

PR #916, opened 2017-08-13
mw/kubernetes: remove federation and cidr

miekg:
Remove both as we have a corefile syntax change that handles cidr and
remove federation because that is going to be its own middleware.


miekg:
This blatantly removes failing tests.

miekg:
Again, inclined to merge to re-add functionality and tests in other middleware (see outstanding PRs)

Issue #917, opened 2017-08-14
CoreDNS does not build with Go 1.7


insomniacslk:
As discussed on https://github.com/coredns/coredns/commit/26d8680a11cbe444546d9ec247837ddf0594350e#commitcomment-23618661 , the build fails on Go 1.7. It succeeds on 1.8 and 1.6.
The Travis-CI configuration[1] only enforces 1.8 but as said by @miekg there's no policy but it makes sense to support at least the last two versions

[1] https://github.com/coredns/coredns/blob/00f5c7797e265dc2fecad5acbfcfdad9797e48e5/.travis.yml#L10

johnbelamaric:
@yongtang can you take a look?

insomniacslk:
Currently the only issue is with the kubernetes middleware, as I reported in the linked comment. I managed to build it by replacing the call to `p.Close()` into `p.listener.Close()`, which is the only thing that made sense to me. I may be wrong, so please review it.

If that's the right fix, the only necessary step would be to add Go 1.7 to the `.travis.yml` configuration and trigger a build.

Happy to help get this sorted out

johnbelamaric:
Thanks, feel free to submit that as a PR, the fix and the 1.7 in Travis.

insomniacslk:
Sure, will do that tonight

johnbelamaric:
👍 thanks

insomniacslk:
For reference, the current `apiProxy.Close` is a call to `http.Server.Close` as defined in Go 1.8.x's code, https://github.com/golang/go/blob/f776b9d5fa3577fe2fdaa3d2ef9d48adee329f9c/src/net/http/server.go#L2454 . It is meant to close all the active listeners.

In `apiProxy.Run`, the only listener is defined in the `apiProxy.listener`, an attribute of type `net.Listener`. As per this, if I understand correctly the code, the right thing to do here is to call `Close` on `apiProxy.listener`. I am building this change on Travis right now

yongtang:
Closed through #919.

PR #919, opened 2017-08-14
Make CoreDNS build on Go 1.7

insomniacslk:
Two changes:
* Added a compiler target in .travis.yml to build against Go 1.7
* Replaced a call to `http.Server.Close` into a call to `net.Listener.Close`, since the former is not implemented in Go 1.7. Go 1.8 implements this as a wrapper that closes all the listeners, so this should be correct here.

The latter needs review from someone with more context on the specific code to make sure this doesn't introduce any regression. Additionally this should go away as soon as Go 1.7 will not be supported anymore. Further discussion on https://github.com/coredns/coredns/issues/917 .

The build on my branch passes correctly, https://travis-ci.org/insomniacslk/coredns/builds/264472119 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ services:
 
 language: go
 go:
+  - 1.7
   - 1.8
 
 go_import_path: github.com/coredns/coredns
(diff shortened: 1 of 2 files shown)


johnbelamaric:
@yongtang please verify this is the correct fix and merge if so

yongtang (approved):
LGTM

Issue #817, opened 2017-08-03
mw/kubernetes: SRV requests for external services


miekg:
@chrisohaver this todo:
https://github.com/coredns/coredns/blob/master/middleware/kubernetes/kubernetes.go#L135
What should happen there? If we have a proxy do an external query? CNAMEs can point to internal names?

chrisohaver:
corrected title... External services is already implemented for `A` records.  External services are registered services in k8s that point to an outside host. Wasn't sure how to handle `SRV` requests on one of these services.  Not sure if k8s allows you to define available ports/protocols for an external service.  If so, then we can check the local k8s object, if not, we'd need to proxy the `SRV` request upstream.

chrisohaver:
> CNAMEs can point to internal names?

Had not considered that case. It may be possible to point an external service to a service inside the k8s cluster.  I have not tried. Not sure if that makes sense to do this.  But if it _can_ be done, to handle it we'd have to conditionally upstream only if the zone it not in the current zone.  Otherwise, lookup the service locally in k8s middleware.

johnbelamaric:
We could set the upstream back to ourselves. See also https://github.com/kubernetes/dns/issues/131 which can be fixed the same way. You risk loops, though.

johnbelamaric:
See also https://github.com/kubernetes/dns/issues/140

chrisohaver:
@johnbelamaric opened issue kubernetes/dns#140 to get consensus on how we should handle SRV requests for external services.

chrisohaver:
IMO cases 1-3 (listed in kubernetes/dns#140) are the same as “combine answers from both external service def and remote SRV request”.

I also think that it makes sense to do the same for 4.  In this case you are presenting the "real" SRV records for the external service target, in addition to "virtual" ones defined by the kubernetes external service definition.

PR #921, opened 2017-08-14
[branched] Parse request

miekg:
https://github.com/coredns/coredns/pull/899 on a new branch


johnbelamaric:
FYI, I tried SRV for an external service in kube-dns, and it returns an incorrect SRV record (as this one did). The question is what to do in these scenarios:

@chrisohaver @miekg 
1) There *is not* an SRV record for the name in the upstream, and port info *is not* specified in the external service definition. I think return no data.
2) There *is* an SRV record in the upstream for that name, and port info *is not* specified in the external service definition. I think return the upstream SRV (which would be the normal DNS thing to do, correct?)
3) There *is not* an SRV record for the name in the upstream, and port info *is* specified in the external service definition. I think return a synthetic SRV based on external service definition.
4) There *is* an SRV record for the name in the upstream, and port info *is* specified in the external service definition. Umm...not sure here. Probably prefer the K8s external service definition. Or maybe return both.

We probably should open up this discussion in kubernetes/dns

chrisohaver:
@johnbelamaric, there is an open issue for how to handle SRV answers for external services: #817.

Agree on your suggestions - except for # 4.  IMO we'd probably want to merge the results of both.  In fact, that logic would apply to 1-3 as well. Just merging results of both local definitions and upstream records would be the same result as described in 1-3.

chrisohaver:
> except for # 4

Just noticed you said "Or maybe return both." :) Yes, i agree.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [branched] Pa..." ]
>FYI, I tried SRV for an external service in kube-dns, and it returns an incorrect SRV record (as this one did). The question is what to do in these scenarios:
>
>@chrisohaver @miekg
>1) There *is not* an SRV record for the name in the upstream, and port info *is not* specified in the external service definition. I think return no data.
>2) There *is* an SRV record in the upstream for that name, and port info *is not* specified in the external service definition. I think return the upstream SRV (which would be the normal DNS thing to do, correct?)
>3) There *is not* an SRV record for the name in the upstream, and port info *is* specified in the external service definition. I think return a synthetic SRV based on external service definition.
>4) There *is* an SRV record for the name in the upstream, and port info *is* specified in the external service definition. Umm...not sure here. Probably prefer the K8s external service definition. Or maybe return both.
>
>We probably should open up this discussion in kubernetes/dns

Yes that makes sense.

Not replying in depth here, but just a general feeling. For NSD the quote 
"Garbage in, Garbage out" worked really well. If you start to put smartness into 
the system it becomes hard to draw a line. For NSD this was mostly performance 
driven.

/Miek

--
Miek Gieben

miekg:
Merging this (now tests don't fail in chris-miek-k8s)

PR #922, opened 2017-08-15
mw/kubernetes: add reverse_test.go

miekg:
This PR adds a reverse tests that copies some of it from the main
integration test. This aids in local testing because you don't need
a full k8s setup running.

It does the most minimal testing, a successful return and a nodata
response that includes the SOA for in-addr.arpa.


miekg:
Oh and a few pointer cleanups that or not necessary.

chrisohaver:
There's a lot of duplicated code here.  Why not add these test cases to the existing ServeDNS tests.

miekg:
Yes, deliberate choice, most is boilerplating. The existing tests are long, this neatly tests a reverse and a nodata. Any failures are easy to spot and the test completeness is also easy to assert.

chrisohaver (approved):

PR #926, opened 2017-08-16
tests: add SortAndCheck helper

miekg:
There was quite some code duplication in a lot of tests to check if
an answer was considered Ok. Created a test.SortAndCheck helper function
that takes care of this.


miekg:
As commented by @chrisohaver there was indeed a lot of duplicated code - this cuts back considerably on it.

PR #930, opened 2017-08-16
mw/autopath: integration test with erratic

miekg:
Test autopath integration with erratic.

This led to a bunch of fixes in autopath.


miekg:
@chrisohaver or @johnbelamaric if you could this a glance over. This should fix the run-time issues you saw (and a bunch of other things)

Issue #802, opened 2017-07-29
middleware/rewrite: Support for variables in EDNS0 local options


johnbelamaric:
Right now, the rewrite allows you to append static data (from the Corefile) in EDNS0 local options. This works for some use cases, but for others there is a need to populate these options with various data about the original request (which will then be proxied so the receiving proxy wouldn't have this data otherwise). For example, the ability to grab:

* `qname`
* `qtype`
* `client_ip`
* `client_port`
* `protocol`
* `server_ip`
* `server_port`

Right now I have a specific use case for client_source_ip and not the others. But it seems like we should add the basic data we have in there.

I imagine something like:

```
rewrite edns0 local set 0xffee $client_ip
```

huynhmb359:
John,  I'll start with the list you provided.   The macro name will be exactly as in the list.

huynhmb359:
For qname and qtype,  the question section may have more than 1 question.   Which one should we select or are we going to report them all?

johnbelamaric:
@miekg I have noticed we only process the first question (https://github.com/coredns/coredns/blob/master/core/dnsserver/server.go#L188) is that normal?

miekg:
What? Multiple questions in the section section. No that will never work. Although DNS was designed to support this, a reply could only have rcode making the notion useless

huynhmb359:
Thanks for the response.  It's confirmed that we are expecting only one question in the "question section". 

I need to fetch the LocalAddr in the query (i.e. server IP/port).  The  Request object provides APIs to access to the RemoteAddr (i.e. client IP/port) like  IP()/Family()/Port().  Can we add the similar APIs to access the LocalAddr in the Request object?

miekg:
LocalAddr? Yes, this is available in the dns.ResponseWriter.

huynhmb359:
Yes LocalAddr is available in dns.ResponseWriter.  However,  the Request has a set of APIs that conveniently read the information in the RemoteAddr.   These include parsing IP/ IP version check ...etc.
Reading the raw LocalAddr is identical to the reading of raw RemoteAddr.   In stead of duplicating the code,  I wonder if we can add these into the Request so that the common code can be extracted and reused.

miekg:
Sorry, with Request you mean `request.Request`. If so you can just access the `responseWriter` with:
```
r = request.Request{...}
r.W.LocalAddr()
```

huynhmb359:
[...]

PR #937, opened 2017-08-18
Add set EDNS0 with variable substitution

huynhmb359:
This PR to implement the issue https://github.com/coredns/coredns/issues/802.


middleware/rewrite/README.md
```diff
@@ -54,6 +55,21 @@ rewrites the first local option with code 0xffee, setting the data to "abcd". Eq
...
 ~~~
 
+* A variable data must start with `$`. Following are the supported variables:
+    * $qname
+    * $qtype
+    * $client_ip
+    * $client_port
+    * $protocol
+    * $server_ip
+    * $server_port
```
miekg: Variables in templates in other middleware (*auto* and *reverse*) are using the `{}` syntax, so `$qname` would become `{qname}`. etc.

middleware/rewrite/edns0.go
```diff
@@ -137,13 +151,174 @@ func newEdns0LocalRule(action, code, data string) (*edns0LocalRule, error) {
...
 // These are all defined actions.
 const (
 	Replace = "replace"
 	Set     = "set"
 	Append  = "append"
 )
+
+// Supported local EDNS0 variables
+const (
+	QueryName  = "$qname"
```
miekg: Do these need to be exported?

miekg (changes requested):
Build failed with:
Reload_test.go:42: Could not send message: read udp [::1]:40413->[::1]:45671: i/o timeout
Which is #945 

Code looks good.

middleware/rewrite/edns0.go
huynhmb359: No they don't.  I will change to local constant.

middleware/rewrite/README.md
huynhmb359: I will change to use {} for variables.

middleware/rewrite/edns0.go
```diff
@@ -137,13 +153,175 @@ func newEdns0LocalRule(action, code, data string) (*edns0LocalRule, error) {
...
+	if i, ok := ip.(*net.TCPAddr); ok {
+		a = i.IP
+	}
+	if a.To4() != nil {
+		return 1
+	}
+	return 2
+}
+
+// getRuleData returns the data specified by the variable
```
miekg: Just drop the `get`, just ruleData

middleware/rewrite/edns0.go
```diff
@@ -137,13 +153,175 @@ func newEdns0LocalRule(action, code, data string) (*edns0LocalRule, error) {
...
+
 // These are all defined actions.
 const (
 	Replace = "replace"
 	Set     = "set"
 	Append  = "append"
 )
+
+// Supported local EDNS0 variables
+var checkVariable = regexp.MustCompile("{.*}")
```
miekg: You are now using a regular expression in the fast path, which is something we should not do. A custom parser doesn't seem too hard in this case?
(Although it would be nice to get some number on how heavy a regexp would be, compared to something custom)

huynhmb359: Ok,  I'll change to string match.

middleware/rewrite/edns0.go
huynhmb359: ok,  will do.

PR #938, opened 2017-08-18
mw/kubernetes: remove zone from parseRequest

miekg:
State has the zone info as well, so don't need to have it in
parseRequest anymore.

### DIFF
--- a/middleware/kubernetes/parse.go
+++ b/middleware/kubernetes/parse.go
@@ -20,7 +20,6 @@ type recordRequest struct {
 	// A each name can be for a pod or a service, here we track what we've seen. This value is true for
 	// pods and false for services. If we ever need to extend this well use a typed value.
 	podOrSvc string
-	zone     string
 }
 
 // parseRequest parses the qname to find all the elements we need for querying k8s.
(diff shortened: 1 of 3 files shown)


miekg:
I'll double check why test coverage is going down. Could be just the deleted lines.

middleware/kubernetes/parse.go
```diff
@@ -101,6 +98,5 @@ func (r recordRequest) String() string {
 	s += "." + r.service
 	s += "." + r.namespace
 	s += "." + r.podOrSvc
-	s += "." + r.zone
```
chrisohaver, line 104: This function will escape to stack.  Consider using strings.Join() instead of "+"

chrisohaver (changes requested):

middleware/kubernetes/parse.go
miekg: sorry, what? You mean each s is a new string and therefor adds GC?
Sure I can change this.

miekg: strings.Join requires a string slice to be constructed. If you worried about mem...

chrisohaver: Oh ok, never mind then, i must be mistaken.

chrisohaver (approved):

miekg:
Travis 1.7 build failed with
~~~
$ docker run -d --net=host --name=etcd quay.io/coreos/etcd:v$ETCD_VERSION
Unable to find image 'quay.io/coreos/etcd:v2.3.1' locally
docker: Error response from daemon: Get https://quay.io/v2/coreos/etcd/manifests/v2.3.1: Get https://quay.io/v2/auth?scope=repository%3Acoreos%2Fetcd%3Apull&service=quay.io: net/http: TLS handshake timeout.
See 'docker run --help'.
~~~

PR #939, opened 2017-08-18
mw/kubernetes: Rewrite parseRequest and Readability improvements

miekg:
This build on https://github.com/coredns/coredns/pull/938 (which has been merged, this has been rebased).

This rewrites parseRequest and considers a name invalid if its too long or if it is not pod or svc.
This drops the _udp and _tcp check, although we might bring that back, but my feeling there is that there could be more outside of k8s' control, unlike the pod/svc thing...

It removes to need to parseRequest to look at the type making nodata responses work (or you would just receive a reply if a service was found).

Commit has more details. Also lots of comments added because 'endpoints' are rather tricky to handle.

If we want to fully fix nodata vs nxdomain I believe we need to record that we've filtered out services on k.get()


miekg:
ping?

chrisohaver (approved):

PR #944, opened 2017-08-19
Dns.join

miekg:


miekg:
failed because of: https://github.com/coredns/coredns/issues/945

PR #946, opened 2017-08-19
Travis speed

miekg:


miekg:
don't know where these commits are coming from.

PR #947, opened 2017-08-19
travis: parallize builds

miekg:


miekg:
This gives 8 parallel test jobs, 4 for each Go version.

miekg:
As discussed in the bug the coverage is where the time is spend

PR #953, opened 2017-08-19
Disable Travis cache

yongtang:
It looks like the cached path  `$GOPATH/src`  overlaps with the source path of `$GOPATH/src/github.com/coredns/coredns`.

Disable the cache for now.

May enable cache in the future if we could exclude `$GOPATH/src/github.com/coredns/coredns` from `$GOPATH/src`.


miekg:
Ack. This was just some experiment.

Issue #945, opened 2017-08-19
mw/auto: time in test


miekg:
The auto_test.go has a time.Sleep() in the test making it flaky:
~~~
time.Sleep(1100 * time.Millisecond) // wait for it to be picked up
~~~
This should be reworked/removed.

miekg:
Also sometimes fails with:
~~~
- FAIL: TestReload (2.00s)
	reload_test.go:42: Could not send message: read udp [::1]:49889->[::1]:59288: i/o timeout
~~~

PR #957, opened 2017-08-20
mw/kubernetes: resync to opts

miekg:
* resyncPeriod to opts: Only used to initialize the cache that already has a dnsControlopts, so
remove it from the main kubernetes struct.
* labelSelector to opts: Only used to initialize the cache, put in dnsControlOpts.


miekg:
1 test failed with
~~~
- FAIL: TestReload (2.00s)
	reload_test.go:42: Could not send message: read udp [::1]:49889->[::1]:59288: i/o timeout
~~~
https://github.com/coredns/coredns/issues/945 updated

PR #963, opened 2017-08-21
WIP: K8s axfr

miekg:
doesn't work (yet), just prints stuff to stdout; don't merge.

### DIFF
--- a/middleware/auto/setup.go
+++ b/middleware/auto/setup.go
@@ -10,9 +10,9 @@ import (
 
 	"github.com/coredns/coredns/core/dnsserver"
 	"github.com/coredns/coredns/middleware"
-	"github.com/coredns/coredns/middleware/file"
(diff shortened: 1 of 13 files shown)


miekg:
A few things need to be fleshed out here, but it is not super difficult:
- [ ] notifies need to be pulled out of *file*. Create `pkg/notify`
- [ ] onstartup notify timer needs to be set
- [x] the dnscontroller needs a bunch a update functions that update the soa's serial
- [ ] the soa serial needs to be unix epoch, even in the normal case (nobody transfers, so nobody cares).
- [ ] the SOA is created in backend_lookup.go. For this to work, this needs to call out to a `Serial` method the returns the serial number - this means extending the `Backend` interface.

miekg:
The dynamic SOA (serial) stuff probably deserves a separate PR, before attempting AXFR

PR #964, opened 2017-08-21
core: hide registerHandler

miekg:
Remove RegisterHandler and just make it implicit when we look at the
handler compilation step.


miekg:
Has some extra commits to up the test coverage, some setup's went down with > 1%

miekg:
fyi: docker error in one of the builds

PR #973, opened 2017-08-23
core: block CH queries earlier

miekg:
block chaos queries, unless the chaos middleware is loaded. We respond
with REFUSED.

This removes the need for each middleware to do this class != ClassINET
if-then.


miekg:
~~~
--- FAIL: TestProxyToChaosServer (0.01s)
	server_test.go:53: Expected successful reply, got REFUSED
~~~
Inclined to remove/change this test as this failure is now expected - or allow proxy to receive CH class queries as well?

miekg:
@pmoroney do you have time to review this?

PR #976, opened 2017-08-23
mw/health: poll other middleware

miekg:
This add the infrastructure to let other middleware report their health
status back to the health middleware. A health.Healther interface is
introduced and a middleware only needs to implement that. A middleware
that support healthcheck is not statically configured, but this could be
made into a RegisterHealth thing so it can *also* be done by the
middleware implementing it.

Every second each supported middleware is queried and the global health
state is updated.

<strike>The erratic middleware implements Healther as an example.</strike> Removed as not relevant to this PR - and *erratic* might do something more whacky with health checks in the future (like flip-flopping between states)


yongtang (approved):
LGTM. That is a good enhancement overall!

PR #995, opened 2017-08-26
mw/kubernetes: add configurable TTL

### LINKED ISSUE #935: mw/kubernetes: returns answers with 0 TTL?
While coding federation I noticed that the *kubernetes* mw returns records with 0 TTL:
```
svc0.testns.prod.svc.cluster.local.	0	IN	CNAME	svc0.testns.cluster.local..svc.fd-az.fd-r.cluster.local.
```
Need to delve a bit deeper, but this seems weird.

miekg:
Add ttl option to kubernetes. This defaults to 5s but allows
configuration to go up to 3600.

Configure the tests so that a few actually check for the 5s, while the
rest use the TTL of 303 which is ignored by the checking code.

Fixes #935

### DIFF
--- a/middleware/kubernetes/README.md
+++ b/middleware/kubernetes/README.md
@@ -28,6 +28,7 @@ kubernetes [ZONES...] {
     labels EXPRESSION
     pods POD-MODE
     upstream ADDRESS...
+    ttl TTL
     fallthrough
 }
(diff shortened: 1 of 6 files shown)


yongtang (approved):
LGTM

PR #1002, opened 2017-08-29
IP endpoint for dnstap

### LINKED ISSUE #924: mw/dnstap: IP endpoint for dnstap
In a microservices environment, it makes more sense to send dnstap data over the network rather than to a domain socket. What level of effort would that be?

2buisson:
Fixes #924 thanks to https://github.com/dnstap/golang-dnstap/pull/13.

### DIFF
--- a/middleware/dnstap/README.md
+++ b/middleware/dnstap/README.md
@@ -27,6 +27,12 @@ Log information including the wire-format DNS message about client requests and 
 dnstap /tmp/dnstap.sock full
 ~~~
 
+Log to a remote endpoint.
+
+~~~ txt
+dnstap 127.0.0.1:6000 full
+~~~
+
 ## Dnstap command line tool
 
 ~~~ sh
@@ -46,4 +52,10 @@ The following command listens on the given socket and saves message payloads to 
 
 ~~~ sh
 dnstap -u /tmp/dnstap.sock -w /tmp/test.dnstap
+~~~
+
+Listen for dnstap messages on port 6000.
+
+~~~ sh
+dnstap -l 127.0.0.1:6000
 ~~~
--- a/middleware/dnstap/README.md
+++ b/middleware/dnstap/README.md
@@ -24,13 +24,13 @@ dnstap /tmp/dnstap.sock
 Log information including the wire-format DNS message about client requests and responses to */tmp/dnstap.sock*.
 
 ~~~ txt
-dnstap /tmp/dnstap.sock full
+dnstap unix:///tmp/dnstap.sock full
 ~~~
 
 Log to a remote endpoint.
 
 ~~~ txt
-dnstap 127.0.0.1:6000 full
+dnstap tcp://127.0.0.1:6000 full
 ~~~
 
 ## Dnstap command line tool
--- a/middleware/dnstap/out/tcp.go
+++ b/middleware/dnstap/out/tcp.go
@@ -14,45 +14,43 @@ type TCP struct {
 }
 
 // NewTCP returns a TCP writer.
-func NewTCP(address string) (s *TCP) {
-	s = &TCP{address: address}
+func NewTCP(address string) *TCP {
+	s := &TCP{address: address}
 	s.frames = make([][]byte, 0, 13) // 13 messages buffer
-	return
+	return s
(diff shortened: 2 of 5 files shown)


yongtang:
I think the common way to define unix vs. tcp socket is to specify protocol:
```
unix:///var/run/dnstap.sock
tcp://127.0.0.1:8080
```

Maybe we could add the protocol at the beginning (and default to `unix://` if it is not specified)?

middleware/dnstap/README.md
```diff
@@ -27,6 +27,12 @@ Log information including the wire-format DNS message about client requests and
 dnstap /tmp/dnstap.sock full
 ~~~
 
+Log to a remote endpoint.
+
+~~~ txt
+dnstap 127.0.0.1:6000 full
```
yongtang:
```
unix:///var/run/dnstap.sock
tcp://127.0.0.1:8080
```

yongtang (changes requested):

middleware/dnstap/out/tcp.go
```diff
@@ -0,0 +1,61 @@
...
+type TCP struct {
+	address string
+	frames  [][]byte
+}
+
+// NewTCP returns a TCP writer.
+func NewTCP(address string) (s *TCP) {
+	s = &TCP{address: address}
+	s.frames = make([][]byte, 0, 13) // 13 messages buffer
+	return
```
miekg: general observation: don't do these naked returns :-)

middleware/dnstap/out/tcp.go
```diff
@@ -0,0 +1,61 @@
...
+	for _, frame := range s.frames {
+		if _, err = enc.Write(frame); err != nil {
+			return
+		}
+	}
+	err = enc.Flush()
+	return
+}
+
+// Close is an alias to Flush.
```
miekg: Why is this needed?

middleware/dnstap/out/tcp_test.go
```diff
@@ -0,0 +1,66 @@
...
+	if err := tcp.Flush(); err != nil {
+		return err
+	}
+	return nil
+}
+func TestTcp(t *testing.T) {
+	tcp := NewTCP("localhost:14000")
+
+	if err := sendOneTcp(tcp); err == nil {
+		t.Fatal("not listening but no error")
```
miekg: Test messages are different from errors, as they will not be wrapped and will be directly read by humans. So make them a full sentence "Not listening, but no error."
Why is this an Fatal in the test btw?

middleware/dnstap/setup.go
```diff
@@ -26,30 +28,43 @@ func wrapSetup(c *caddy.Controller) error {
 	return nil
 }
 
-func parseConfig(c *caddyfile.Dispenser) (path string, full bool, err error) {
+func parseConfig(c *caddyfile.Dispenser) (path string, socket, full bool, err error) {
```
miekg: Shouldn't keep adding return parameters in this function. Either start an dnsOpt struct which has all these or, if possible, fill out a *dnstap structure with the correct elements.

middleware/dnstap/setup.go
```diff
@@ -26,30 +28,43 @@ func wrapSetup(c *caddy.Controller) error {
...
 	dnstap := Dnstap{Pack: full}
 
-	o, err := out.NewSocket(path)
-	if err != nil {
-		log.Printf("[WARN] Can't connect to %s at the moment", path)
+	var o io.WriteCloser
+	if socket {
+		o, err = out.NewSocket(path)
+		if err != nil {
+			log.Printf("[WARN] Can't connect to %s at the moment", path)
```
miekg: Include error in log message.

middleware/dnstap/out/tcp.go
2buisson: To satisfy `io.WriteCloser` similarly to type `out.Socket`.

yongtang (approved):

middleware/dnstap/out/tcp.go
miekg: Ok, please add that to the comment :)

PR #1003, opened 2017-08-29
mw/health: call Shutdown on FinalShutdown

### LINKED ISSUE #979: middleware/health: healthcheck fails after graceful reload CoreDNS with SIGUSR1 signal

miekg:
Reloading caddy won't kill the health handler. Only on final shutdown
we stop the handler.

Currently when reloading CoreDNS with -SIGUSR1 the health handler stops
answering - there is a test for this but it doesn't capture whole
process reloading, sadly. This PR keeps the handler alive during reloads
and only stops on process shutdown.

Fixes #979

### DIFF
--- a/middleware/health/setup.go
+++ b/middleware/health/setup.go
@@ -47,7 +47,7 @@ func setup(c *caddy.Controller) error {
 	})
 
 	c.OnStartup(h.Startup)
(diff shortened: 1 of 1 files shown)


miekg:
*manually* tested... :-(

PR #1004, opened 2017-08-30
mw/kubernetes: split integration tests

miekg:
* separate file and test for api fallthrough, does not need all other
  servers to be started.
* more split ups: make it clear when or when not we need an upstream server,
  as just needlessly start it in doIntegrationTests.


miekg:
This still leaves CNAME tests that should be pulled out and the fallthrough stuff to could be minimized.

rajansandeep (approved):
Looks good.

PR #1016, opened 2017-09-01
Cleaning up dnssec docs

bradbeam:
- Removing mention of autogeneration of dnssec keys since this feature isn't implemented
- Proposing to standardize on not specifying signing key suffix ( [code](https://github.com/coredns/coredns/blob/master/middleware/dnssec/setup.go#L113) can handle suffixes, but when providing examples it's unclear on what format should be used )
- Updating syntax to be all inclusive of options ( perhaps it's worth adding in a note on which fields are optional ). This is probably something we'd want to investigate changing for all middleware. 
- Added an additional example for kubernetes

### DIFF
--- a/middleware/dnssec/README.md
+++ b/middleware/dnssec/README.md
@@ -5,32 +5,33 @@
 ## Syntax
 
 ~~~
-dnssec [ZONES...]
-~~~
-
-* **ZONES** zones that should be signed. If empty, the zones from the configuration block
-    are used.
-
-If keys are not specified (see below), a key is generated and used for all signing operations. The
-DNSSEC signing will treat this key a CSK (common signing key), forgoing the ZSK/KSK split. All
-signing operations are done online. Authenticated denial of existence is implemented with NSEC black
-lies. Using ECDSA as an algorithm is preferred as this leads to smaller signatures (compared to
-RSA). NSEC3 is *not* supported.
-
-A single signing key can be specified by using the `key` directive.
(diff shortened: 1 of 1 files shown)


middleware/dnssec/README.md
```diff
@@ -5,32 +5,28 @@
...
+
+* `ZONES` zones that should be signed. If empty, the zones from the configuration block
+    are used.
+
 * `key file` indicates that key file(s) should be read from disk. When multiple keys are specified, RRsets
   will be signed with all keys. Generating a key can be done with `dnssec-keygen`: `dnssec-keygen -a
-  ECDSAP256SHA256 <zonename>`. A key created for zone *A* can be safely used for zone *B*.
+  ECDSAP256SHA256 <zonename>`. A key created for zone *A* can be safely used for zone *B*. The name of the
+  key should be the basename of the generated key -- `Kexample.org+013+45330` instead of specifying the
+  `.private` or `.key` suffix.
```
miekg: Note that using `.key` of `.private` is actually supported. Base name will be "calculated" from that.

miekg (changes requested):
.key/.private is supported. Apart from that lgtm.

bradbeam:
Yeah I saw it was supported. That gives 3 variations for how the key file can be specified which can lead to confusion. I'll update to show the various supported formats.

miekg (approved):

PR #1019, opened 2017-09-02
mw/proxy: simplify google code

miekg:
Minimize bootstrap code, use time.Timer for first-time bootstrap.

Also re-resolve every 120 seconds, instead of 300 (might eventually make
this an option).


miekg:
wip: because I'm running and testing this on my raspberry pi

PR #1023, opened 2017-09-05
Run Docker container as a non-root user

nrvnrvn:


yongtang:
@nicorevin Is there any reason to make the change?

nrvnrvn:
running as a non-root is a generally recommended practice.

miekg:
Outside of docker this would make perfect sense, in fact this is how I run my CoreDNS instances. Inside docker I don't think this adds anything?

nrvnrvn:
Running as a non-root prevents some popular possible attacks. And this is just a part of a "least privilege" principle approach.

miekg:
https://medium.com/@mccode/understanding-how-uid-and-gid-work-in-docker-containers-c37a01d01cf
Wow, way more complex (and stupid) than I would have thought. My take away though is that you should us --user in this case?

nrvnrvn:
`--user` would be generally ok if we were allowed to run on the privileged port:
```sh 
$ docker run --rm --user 1000 coredns/coredns
2017/09/07 16:10:44 listen tcp :53: bind: permission denied
```
Anyway passing `--user uid:gid` is a more ad-hoc feature which is useful when you need to jump as `root` into a running container launched with non-root user.

And yes, this is much more complex. There is even [a user namespace remapping feature](https://docs.docker.com/engine/security/userns-remap/) (still a WIP as far as I know hence very fragile and unpredictable).

miekg:
I feel strongly about *not* merging this, it complicates the Dockerfile for something that is essentially docker's problem.

yongtang:
@nicorevin Running `--user uid:gid` is actually pretty common with docker. I tends to agree with @miekg that we may not want to change that unless there are absolutely needed.

One other thing I noticed is that the Dockerfile now actually downloads the release (pre-built) binary from the GitHub. I think the intention of the `Dockerfile` in the directory is to test the generated binary from dev repo so that with a `docker build` you could test the dev branch on your local machine.

nrvnrvn:
@yongtang it is also a good practice to define a USER in Dockerfile. And I explained why `--user` won't work with coredns now.
As for downloading I used it to keep the image size small. Touching coredns binary adds an extra layer and doubles the size.

It is up to you to merge it or not. I decided to share the way I build and execute coredns once I discovered that there is an official `coredns/coredns` image.

miekg:
Closing for now

Issue #925, opened 2017-08-16
crash: metrics/vars.Report


elephantfries:
Hi, I am seeing coredns crashing every now and then with what appears to be 'index out of range'. I've seen at least three crashes like that, they all produce the same trace. This is with etcd backend with entries valid for a few hours before being discarded by etcd. Version CoreDNS-010. Trace captured by systemd.
[coredns.crash.1.txt](https://github.com/coredns/coredns/files/1227134/coredns.crash.1.txt)

miekg:
Interesting crash. I'll take a look

miekg:
The only reason this can crash if `req.Msg` is empty, meaning no incoming message has been seen. Which is odd to say the least.

miekg:
This crashes in `DefaultErrorMsg` which together with metrics can lead to this. This is bug 2.

There are two crashes:
~~~
Aug 13 11:34:06 : panic: runtime error: index out of range [recovered]
Aug 13 11:34:06 : panic: runtime error: index out of range
~~~
On that gets recovered and then in the recover it crashes again, showing the stack trace.

Can you run with `debug` middleware enabled? That should show the first crash and bug 1.

elephantfries:
Yes, added debug.  It may take a few days.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] crash: metric..." ]
>Yes, added debug.  It may take a few days.

Thank you.

miekg:
@elephantfries We will probably release a 1.0.0 before we get a handle on this bug. But once seen and fixed will do a 1.0.1 (or whatever we're at then)

elephantfries:
I just checked and saw all three of my test sites showing coredns crashing.  It took 11 days and then they all crashed within 8 hours of one another.  Strange.  Here's a trace from journald (removed host name from the log).  All three produce the same trace.
[coredns.crash.2.txt](https://github.com/coredns/coredns/files/1262406/coredns.crash.2.txt)

elephantfries:
Just wanted to add that my system is an experimental network that uses hack job resolvers. It is possible it might unintentionally generate malformed queries. I'm very interested on my side of things if that is the case. But then, again, coredns is serving a stub domain off a domain hosted by namecheap.com. I would think a malformed query would be rejected before reaching the stub.

miekg:
What, *etcd* and `stubzones`? You may be hitting:
~~~
[...]

PR #1026, opened 2017-09-06
mw/etcd: fix 'fallthrough'

miekg:
This adds 'fallthrough' for *etcd* in the correct place.

Thanks to @johnbelamaric for pointing this out in #925.

### DIFF
--- a/middleware/etcd/README.md
+++ b/middleware/etcd/README.md
@@ -23,6 +23,7 @@ If you want to `round robin` A and AAAA responses look at the `loadbalance` midd
 ~~~
 etcd [ZONES...] {
     stubzones
+    fallthrough
     path PATH
     endpoint ENDPOINT...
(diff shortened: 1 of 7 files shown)


johnbelamaric (commented):
Rather than removing it completely, why not just implement it as it is done in other middleware. Basically I believe that means moving the check for Fallthrough to here:

https://github.com/coredns/coredns/blob/9d44397827425e567af01d43564c4294b42e98c9/middleware/etcd/handler.go#L89

miekg:
that could be done, and is backward compatible. Looking through kubernetes' ServeDNS, it uses k.Fallthrough to call the next middleware when it isn't authoritative for the zone. That's wrong.

miekg:
updated (with change to kubernetes as well)

johnbelamaric (approved):

PR #1029, opened 2017-09-06
mw/autopath: correct type for type switch

### LINKED ISSUE #1025: middleware/autopath: @kubernetes option not working
Autopath server-side lookups are not being performed when issuing dns queries from pods in the cluster.

I tested with the following configuration...

```
    .:53 {
      health
      errors
      log
      autopath cluster.local @kubernetes
      kubernetes cluster.local {
        pods verified
      }
    }
```

I see the following warning in the logs ...
```
2017/09/06 17:48:53 [WARNING] No search path available for autopath
```
.. suggesting that autopath middleware is not able to get the kubernetes search path.

miekg:
Use pointer to kubernetes as that is what is registered. Fix up
federation which had the same mistake.

Fixes #1025

### DIFF
--- a/middleware/autopath/setup.go
+++ b/middleware/autopath/setup.go
@@ -28,12 +28,11 @@ func setup(c *caddy.Controller) error {
 
 	// Do this in OnStartup, so all middleware has been initialized.
 	c.OnStartup(func() error {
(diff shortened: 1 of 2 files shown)


miekg:
Other than integration tests I have no good idea how to test this.

PR #1043, opened 2017-09-10
documention: test README snippets

miekg:
If a README has a corefile snippet that is annotated with `corefile`,
this test will parse the instance and checks the snippet is legal.
This means a) we will get better docs b) we know for sure everything
still parses.

The test parses everything in middleware/*/README.md, it does not check
for README presence, just Corefile snippets. The port used is 10053 and
overrides whatever port is set in the docs.

The secondary middleware was used as an example and adds two examples
that should parse.

failures show up as:

~~~
--- FAIL: TestReadme (0.04s)
	readme_test.go:50: Testing ../middleware/secondary/README.md, with 100 byte snippet
	readme_test.go:50: Testing ../middleware/secondary/README.md, with 93 byte snippet
	readme_test.go:53: Failed to start server for input "middleware/secondary: Corefile:3 - Error during parsing: unknown property 'transfeT'":
		. {
		    secondary example.net {
		        transfeT from 10.1.2.1
		        transfer to *
		    }
		}
FAIL
~~~

### DIFF
--- a/middleware/secondary/README.md
+++ b/middleware/secondary/README.md
@@ -31,9 +31,24 @@ secondary [zones...] {
 
 ## Examples
 
+Transfer `example.org` from 10.0.1.1, and if that fails try 10.1.2.1.
+
+~~~ corefile
+example.org {
+    secondary {
+        transfer from 10.0.1.1
+        transfer from 10.1.2.1
+    }
+}
 ~~~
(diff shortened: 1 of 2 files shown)


miekg:
grr failing check, but unrelated to this PR.

PR #1057, opened 2017-09-11
mw/errors: doc

### LINKED ISSUE #1052: mw/error: update docs?
from: https://gist.github.com/jpmens/c3f593c81e630b4a616c6205a97fd7a4

> errors: "The only valid name for FILE is stdout". huh?!?

miek: yes

miekg:
Remove the [FILE] option from the docs, we still allow it, but this one
value that we allow now *stdout* isn't shown anymore.

Also test the snippet with readme_test.go by adding `~~ corefile`.

Fixes #1052

### DIFF
--- a/middleware/errors/README.md
+++ b/middleware/errors/README.md
@@ -2,26 +2,21 @@
 
 *errors* enables error logging.
 
-Any errors encountered during the query processing will be printed on standard output.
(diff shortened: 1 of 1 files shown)


miekg:
service=quay.io: net/http: TLS handshake timeout: on one of the checks

PR #1064, opened 2017-09-12
Update backend-work branch

VioletRainbows:
Update backend-work branch to match changes in master.


miekg:
@VioletRainbows IIRC you moved a lot of `ServiceBackend` interface code around to resolve cyclic dependencies? It would like to see this PR *not* do that. As said I've been playing with adding etcd3 myself and didn't see any need to fiddle with `ServiceBackend`,  the same should be true for consul...

VioletRainbows:
@miekg The cyclic dependencies should be resolved. They existed because I redeclared the `ServiceBackend` interface. It is no longer the case https://github.com/coredns/coredns/pull/770.

Moving from etcdv2 to etcdv3 should be a simple manner of reimplementing `etcd.go` to work with etcdv3. No other files should be affected (from what I remember).

This branch exist to share code and tests between etcd and consul. I still have much work to do, but I'm slowly getting there.

VioletRainbows:
@miekg I was wrong. I do not know how to resole the cyclic imports without moving things around.

I was thinking doing something like this:
```
└── backend
    ├── consul
    ├── etcd
    ├── msg
    └── tests
```

The issue comes when `setup.go` in `etcd` need to import `backend`, but `backend` also need to import `etcd` for tests.

Having two `setup.go` (one in `backend` and one in `etcd`) isn't working code wise.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Update backen..." ]
>@miekg I was wrong. I do not know how to resole the cyclic imports without moving things around.
>
>I was thinking doing something like this:
>```
>└── backend
>    ├── consul
>    ├── etcd
>    ├── msg
>    └── tests
>```

Which is basically the same layout as we have now, with backend in the `plugin/` 
directory.

>The issue comes when `setup.go` in `etcd` need to import `backend`, but `backend` also need to import `etcd` for tests.

What kind of test is this?

/Miek

--
Miek Gieben

VioletRainbows:
My case was about `lookup_test.go`. It was still in `middleware/backend`. It was required to import `middlerare/backend/etcd` (for `Etcdv2` definition) who is also required to import `middleware/backend` (for `Backend` definition)

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Update backen..." ]
>My case was about `lookup_test.go`. It was still in `middleware/backend`. It was required to import `middlerare/backend/etcd` (for `Etcdv2` definition) who is also required to import `middleware/backend` (for `Backend` definition)

that might have only been "github.com/coredns/coredns/plugin/etcd/msg" ? Which 
should be move elsewhere `plugin/pkg/msg` for instance.

/Miek

--
Miek Gieben

PR #1067, opened 2017-09-13
Remove the word middleware

miekg:


miekg:
mechanical changes. Note this breaks external middleware, as exported types (`AddMiddleware`) also have been renamed.
(Although for the time being we could add an alias)

miekg:
I've re-added AddMiddleware which print as warning and then calls AddPlugin, I *think* this should be enough to keep external middle^W plugins working

miekg:
here goes nothing...

PR #1072, opened 2017-09-13
Add inline support for middleware/hosts

### LINKED ISSUE #999: mw/hosts: inline /etc/hosts file?
In our production we were trying to use `middleware/hosts` to add just a couple of entries from `/etc/hosts` file. At the moment a separate file has to be created even though we only need a couple of lines.

Since `/etc/hosts` file itself is line based and is rather straightforward, I think it makes sense to support the inline of `/etc/hosts` inside Corefile. That will be especially convenient when only one or two lines of `/etc/hosts` are used.

yongtang:
This fix add inline support for middleware/hosts so that
it is possible to specify hosts file insides the Corefile:
```
hosts {
    127.0.0.1 localhost localhost.domain
    10.0.0.1 example.org
    fallthrough
}
```

This fix fixes #999.

### DIFF
--- a/plugin/hosts/README.md
+++ b/plugin/hosts/README.md
@@ -11,6 +11,7 @@ available hosts files that block access to advertising servers.
 
 ~~~
 hosts [FILE [ZONES...]] {
+    [INLINE]
     fallthrough
 }
 ~~~
@@ -18,7 +19,10 @@ hosts [FILE [ZONES...]] {
 * **FILE** the hosts file to read and parse. If the path is relative the path from the *root*
   directive will be prepended to it. Defaults to /etc/hosts if omitted
 * **ZONES** zones it should be authoritative for. If empty, the zones from the configuration block
-    are used.
+   are used.
+* **INLINE** the hosts file contents inlined in Corefile. If there are any lines before fallthrough
+   then all of them will be treated as the additional content for hosts file. The specified hosts
+   file path will still be read but entries will be overrided.
 * `fallthrough` If zone matches and no record can be generated, pass request to the next plugin.
 
 ## Examples
@@ -40,6 +44,15 @@ next plugin if query doesn't match.
 
 ~~~
 hosts example.hosts example.org example.net {
+    fallthrough
+}
+~~~
+
+Load hosts file inlined in Corefile.
+
+~~~
+hosts example.hosts example.org {
+    10.0.0.1 example.org
     fallthrough
 }
 ~~~
--- a/plugin/hosts/hostsfile.go
+++ b/plugin/hosts/hostsfile.go
@@ -53,6 +53,11 @@ type Hostsfile struct {
 	// We don't support old-classful IP address notation.
 	byAddr map[string][]string
 
+	// inline saves the hosts file is inlined in Corefile
+	// We need a copy here as we want to use inline to override
+	// the default /etc/hosts
+	inline []string
+
 	expire time.Time
 	path   string
 	mtime  time.Time
@@ -74,6 +79,10 @@ func (h *Hostsfile) ReadHosts() {
 
 	var file *os.File
 	if file, _ = os.Open(h.path); file == nil {
+		// If this is the first time then we will try to parse inline
+		if len(h.byAddr) == 0 && len(h.inline) > 0 {
+			h.Parse(nil)
+		}
 		return
 	}
 	defer file.Close()
@@ -92,7 +101,12 @@ func (h *Hostsfile) Parse(file io.Reader) {
 	hsv6 := make(map[string][]net.IP)
 	is := make(map[string][]string)
 
-	scanner := bufio.NewScanner(file)
+	var readers []io.Reader
+	if file != nil {
+		readers = append(readers, file)
+	}
+	readers = append(readers, strings.NewReader(strings.Join(h.inline, "\n")))
+	scanner := bufio.NewScanner(io.MultiReader(readers...))
 	for scanner.Scan() {
 		line := scanner.Bytes()
 		if i := bytes.Index(line, []byte{'#'}); i >= 0 {
--- a/plugin/hosts/setup_test.go
+++ b/plugin/hosts/setup_test.go
@@ -84,3 +84,77 @@ func TestHostsParse(t *testing.T) {
 		}
 	}
 }
+
+func TestHostsInlineParse(t *testing.T) {
+	tests := []struct {
+		inputFileRules      string
+		shouldErr           bool
+		expectedbyAddr      map[string][]string
+		expectedFallthrough bool
+	}{
+		{
+			`hosts highly_unlikely_to_exist_hosts_file example.org {
+                                10.0.0.1 example.org
+                                fallthrough
+                        }`,
+			false,
+			map[string][]string{
+				`10.0.0.1`: {
+					`example.org.`,
+				},
+			},
+			true,
+		},
+		{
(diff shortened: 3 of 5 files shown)


middleware/hosts/setup.go
```diff
@@ -80,9 +69,28 @@ func hostsParse(c *caddy.Controller) (Hosts, error) {
 				}
 				return h, c.ArgErr()
 			default:
+				if !h.Fallthrough {
+					line := strings.Join(append([]string{c.Val()}, c.RemainingArgs()...), " ")
+					h.inline = append(h.inline, line)
```
miekg: Why do you gather raw line instead of parsing the line here?

I also don't think we should have a line with a keyword in front of it (although I made that mistake in the federation middleware).

I.e. `host <hosty-stuff>` for each one you want to add.

middleware/hosts/hostsfile.go
```diff
@@ -66,50 +68,56 @@ func (h *Hostsfile) ReadHosts() {
 	if now.Before(h.expire) && len(h.byAddr) > 0 {
 		return
 	}
-	stat, err := os.Stat(h.path)
-	if err == nil && h.mtime.Equal(stat.ModTime()) && h.size == stat.Size() {
-		h.expire = now.Add(cacheMaxAge)
-		return
-	}
+	lines := h.inline
+	if len(lines) == 0 {
```
miekg: why the lines == 0 check?

miekg (changes requested):
Some comments, I'm also  missing the README.md changes.

middleware/hosts/setup.go
yongtang: @miekg Done. The line is parsed in place instead of gathered and parsed later.

middleware/hosts/hostsfile.go
yongtang: Removed.

yongtang:
@miekg Thanks for the review. The PR has been updated with READMD.md. Please take a look.

plugin/hosts/README.md
```diff
@@ -11,14 +11,18 @@ available hosts files that block access to advertising servers.
...
     fallthrough
 }
 ~~~
 
 * **FILE** the hosts file to read and parse. If the path is relative the path from the *root*
   directive will be prepended to it. Defaults to /etc/hosts if omitted
 * **ZONES** zones it should be authoritative for. If empty, the zones from the configuration block
-    are used.
+   are used.
+* **INLINE** the hosts file contents inlined in Corefile. If there are any lines before fallthrough
```
miekg, line 23: What about conflicts with /etc/hosts? Will they be overwritten? (would be useful I think)

yongtang: At the moment if there are any inlined content, then `/etc/hosts` will be ignored. I think it also make sense to 
1. Combine them
2. Report conflict.

But I am not sure which way to go. Do you have some thought on it?

plugin/hosts/setup_test.go
```diff
@@ -84,3 +84,77 @@ func TestHostsParse(t *testing.T) {
...
+
+func TestHostsInlineParse(t *testing.T) {
+	tests := []struct {
+		inputFileRules      string
+		shouldErr           bool
+		expectedbyAddr      map[string][]string
+		expectedFallthrough bool
+	}{
+		{
+			`hosts inline example.org {
```
miekg: wait, does this match what the README now says?

miekg (commented):
I think it would be useful to have these inline option overrule whatever is in /etc/hosts?

plugin/hosts/setup_test.go
yongtang: The example in README is:
```
hosts example.hosts example.org {
    10.0.0.1 example.org
    fallthrough
}
```
where `example.hosts` is the hosts file name. If it is a file path that does not exist (like `inline`) then it will be ignored.

yongtang:
Let me update the PR so that inline option overrule whatever is in `/etc/hosts`

plugin/hosts/README.md
miekg: I would think overriding would make sense, but not sure. @pmoroney any ideas?

yongtang:
@miekg The PR has been updated. Please take a look.

plugin/hosts/setup_test.go
```diff
@@ -84,3 +84,77 @@ func TestHostsParse(t *testing.T) {
...
+			false,
+			map[string][]string{
+				`10.0.0.1`: {
+					`example.org.`,
+				},
+			},
+			true,
+		},
+		{
+			`hosts inline example.org {
```
miekg: `inline` signals a non-existent /etc/hosts file?

There is no IPv6 address in these tests, nor an alias: ip name alias

plugin/hosts/hostsfile.go
```diff
@@ -74,60 +77,68 @@ func (h *Hostsfile) ReadHosts() {
...
-		if i := bytes.Index(line, []byte{'#'}); i >= 0 {
-			// Discard comments.
-			line = line[0:i]
-		}
-		f := bytes.Fields(line)
-		if len(f) < 2 {
-			continue
+// Parse read one line  the hostsfile and populates the byName and byAddr maps if needed.
+func (h *Hostsfile) Parse(line string) {
+	if i := strings.Index(line, "#"); i >= 0 {
```
miekg: I'm sorry, but I don't understand the need to move this to `strings`

miekg (commented):
I think we should overwrite the exsiting /etc/hosts entries and add the rest. I.e. this provides a layer on top of them.

I also see we're missing an integration test for this; might be worth adding that at some point in the future (i.e. not in this PR)

yongtang:
@miekg The PR stores the inlined content and override the `/etc/hosts`. Since we have to take into consideration that the records from `/etc/hosts` will be updated and picked up by CoreDNS from time to time, we do need to save the `inline` records "somewhere" outside of `byNameV4`/`byNameV6`/etc.

In this way, when CoreDNS updates records `/etc/hosts`, it will apply the `inline` records on top of it (to override).

In the current PR I saved the `inline` records into `[]string`. It is probably easy to avoid duplicated code as `inline` and `/etc/hosts` will share similar logic. We could also save `inline` to somewhere like,
```
inlineByNameV4 map[string][]net.IP
inlineByNameV6 map[string][]net.IP
inlineByAddr map[string][]string
```

That will avoid the re-parsing of inline content.

pmoroney:
Sorry it took so long for me to notice these changes. I changed my notification settings so I should see these now.

Instead of making parse work on single lines, it might be simpler to change line 95 of hostsfile.go from 
```
scanner := bufio.NewScanner(file)
```
to
```
inline := strings.NewReader(strings.Join(h.inline,"\n"))
scanner := bufio.NewScanner(io.MultiReader(file, inline))
```
That way it's as if the inline entries are added to the end of the hosts file.

plugin/hosts/setup_test.go
miekg: ok, cool, I think `highly-likely-to-exist` (used in other tests) is slightly more appropriate here.

miekg:
Thanks that comment from @pmoroney is a good change, some extra tests (as mentioned) and we're good to go.

yongtang:
@miekg @pmoroney The PR has been updated with the suggested changes and an integration test. Please take a look.

pmoroney:
Looks good to me. 👍

miekg (approved):

PR #1076, opened 2017-09-14
core: Remove AddMiddleware

miekg:
This does not help to make it backwards compatible. The middleware ->
plugin rename invalidates all this. External middleware won't compile
either way.


miekg:
3 checks are already Ok, one is still running.

PR #1081, opened 2017-09-15
Provide access to Dnstap writer

### LINKED ISSUE #1075: Make Writer in Tapper available to enable middleware to write custom DNS tap messages
Add a GetWriter() function into Tapper interface to return the Writer object in DnsTap so that any middleware can write its own custom DNS tap messages.

rdrozhdzh:
Enable other plugins (middleware) to write custom DNS tap messages

 Fixes #1075

### DIFF
--- a/plugin/dnstap/handler.go
+++ b/plugin/dnstap/handler.go
@@ -25,6 +25,7 @@ type (
 	Tapper interface {
 		TapMessage(*tap.Message) error
 		TapBuilder() msg.Builder
+		GetWriter() io.Writer
 	}
 	tapContext struct {
 		context.Context
@@ -55,6 +56,11 @@ func (h Dnstap) TapMessage(m *tap.Message) error {
 // TapBuilder implements Tapper.
 func (h Dnstap) TapBuilder() msg.Builder {
 	return msg.Builder{Full: h.Pack}
+}
+
+// GetWriter implements Tapper.
+func (h Dnstap) GetWriter() io.Writer {
+	return h.Out
 }
 
 // ServeDNS logs the client query and response to dnstap and passes the dnstap Context.


johnbelamaric:
I don't think this is needed, as `Out` is exported. Also `Get` is not idiomatic in Go.

rdrozhdzh:
Other middlewares can only get access to Tapper interface, not to Dnstap object itself. Please let me know if I missed anything

miekg:
If you want to access the dnstap struct, something like this should work
~~~
	c.OnStartup(func() error {
		m := dnsserver.GetConfig(c).Handler("dnstap")
		if m == nil {
			return nil
		}
		if x, ok := m.(dnstap.Dnstap); ok {
			x.Out // do you thing
		}
		return nil
	})
~~~

miekg:
bigger question is: why do we need or want this?

rdrozhdzh:
Thank you Miek.
The idea was to allow sending customized dnstap messages along with standard ones. By customized I mean the messages which are not defined in https://github.com/dnstap/golang-dnstap/blob/master/dnstap.pb/dnstap.proto

johnbelamaric:
Can we tweak the TapBuilder for this?

rdrozhdzh:
TapBuilder could only construct message if it knows the definition of the message. I think it is not desirable to make dnstap middleware dependent on some external definitions

2buisson:
Maybe `Dnstap(tap.Dnstap)` is what you want? It could replace `Tapper.TapMessage(*tap.Message)` I agree.

huynhmb359:
@varyoo,  not sure I understand your suggestion.   Could you please clarify and give an example?  Thnx.

rdrozhdzh:
Oh, @varyoo, you mean the changes from your new PR? Yes, it might work, assuming that dnstap middleware references extendable Dnstap message, see PR https://github.com/dnstap/golang-dnstap/pull/14

rdrozhdzh:
Closing this PR, since the approach proposed by @miekg can be used instead

PR #1083, opened 2017-09-15
Adds the dnstap I/O routines and should fix some issues

2buisson:
As pointed out, dnstap's `out` package Socket and TCP structures were not thread safe while being used by several DNS worker. This PR adds a dnstap IO thread as specified by http://dnstap.info/Architecture/ which solves the issue because Socket and TCP are now being used in a single thread context.

### DIFF
--- a/plugin/dnstap/dnstapio/io.go
+++ b/plugin/dnstap/dnstapio/io.go
@@ -1,0 +1,60 @@
+package dnstapio
+
+import (
+	"fmt"
+	"io"
+
+	tap "github.com/dnstap/golang-dnstap"
+	"github.com/golang/protobuf/proto"
+)
+
+type DnstapIO struct {
+	writer  io.WriteCloser
+	queue   chan tap.Dnstap
+	stop    chan bool
+	dropped int
+}
+
+func New(w io.WriteCloser) *DnstapIO {
+	dio := DnstapIO{}
+	dio.writer = w
+	dio.queue = make(chan tap.Dnstap, 10)
+	dio.stop = make(chan bool)
+	go dio.serve()
+	return &dio
+}
+
+func (dio *DnstapIO) Dnstap(payload tap.Dnstap) {
+	select {
+	case dio.queue <- payload:
+	default:
+		dio.dropped++
+		fmt.Println("[WARN] Dnstap payload dropped.")
+	}
+}
+
+func (dio *DnstapIO) serve() {
+	for {
+		select {
+		case payload := <-dio.queue:
+			frame, err := proto.Marshal(&payload)
+			if err == nil {
+				dio.writer.Write(frame)
+			} else {
+				dio.dropped++
+				fmt.Println("[ERROR] Invalid dnstap payload dropped.")
+			}
+		case <-dio.stop:
+			close(dio.queue)
+			dio.stop <- true
+			return
+		}
+	}
+}
+
+func (dio DnstapIO) Close() error {
+	dio.stop <- true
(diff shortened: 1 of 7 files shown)


plugin/dnstap/dnstapio/io.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+type DnstapIO struct {
+	writer  io.WriteCloser
+	queue   chan tap.Dnstap
+	stop    chan bool
+	dropped int
+}
+
+func New(w io.WriteCloser) *DnstapIO {
```
miekg: docs...

plugin/dnstap/handler.go
```diff
@@ -16,11 +16,14 @@ import (
 // Dnstap is the dnstap handler.
 type Dnstap struct {
 	Next plugin.Handler
-	Out  io.Writer
+	IO   IOThread
```
miekg: hmmm, don't call it Thread, we don't have those in Go (well the language, the underlying runtime obv has them, but we don't see them). Don't have an immediate idea for a better name.

miekg:
All tests are done with `-race`, is it possible to construct one that fails due to data races?

plugin/dnstap/handler.go
```diff
@@ -16,11 +16,14 @@ import (
...
 	Pack bool
 }
```
johnbelamaric: Ok, not in this PR but looking this over I am wondering why we didn't just put the Dnstap *in the context* rather than *extending the Context type*. Here: https://github.com/coredns/coredns/pull/1083/files#diff-9bba1e13b803e0b0fabe763bea4b5fcfR34

What @huynhmb359 and @rdrozhdzh are trying to do are two things:

1) Send a custom DNS tap message in a downstream middleware (see https://github.com/dnstap/golang-dnstap/pull/14).
2) Append some data to the CLIENT_RESPONSE message in a downstream middleware.

For the second one, the idea is that the downstream policy middleware will attach a bunch of EDNS0 options from the original request to the DNS Tap response message so we can correlate these things. So, what I was hoping we could do is have the Dnstap middleware stuff itself in the context, and also stuff a generic map[string]byte[] in the context with a specific name. Downstream middleware could then populate that extra data map by their name. When the DNStap middleware gets back control, it would see the data in that context and included it in the extra field of the CR message.

So...that's what I would like to see. I don't think it means we need to stuff the Dnstap into the Context right now, but I am not sure why we did it via the type.

And yes, I fully acknowledge that I should have said this a month or two ago!!! Clearly not the most thorough review I have done 😜 .

2buisson:
@miekg what is -race?

plugin/dnstap/handler.go
2buisson: EDNS0 options aren't into the response message?
~Also the dnstap CLI tool cannot decode/encode 3rd party payloads.~ it can, just can't display them.

johnbelamaric: They would be in the request but aren't sent back to the client, so wouldn't be in the response.

plugin/dnstap/handler.go
rdrozhdzh: Maybe just call the interface Sender and rename the Dnstap method to Send? So, it will reflect the purpose of the interface, not implementation details

johnbelamaric:
@varyoo the SideCI failures were go lint failures I believe. Run go lint and go vet and clean up anything. I disabled SideCI because it asks for read/write privileges on basically your whole GitHub account which is excessive.

Issue #1013, opened 2017-09-01
mw/file: drop fsnotify


miekg:
Drop the use of nsnotify in the *file* middleware, instead use 30s(?) loop that parses the SOA and if changed reads in the new zone.

It seems we either using fsnotify wrongly or there is some other reason we don't pick up zone changes (#430)

miekg:
So I dislike dropping fsnotify because of the push behavior it has. I do not we set a directory fsnotify for every zonefile: `err = watcher.Add(path.Dir(z.file))` to catch writes. This will add up for lots of zones file. One way to tweak this is to do this per-directory.

stp-ip:
The first iteration would be as easy as to let the current scanner also catch updates and not just creates/deletes. Afterwards one can work on tweaking fsnotify. That would enable a working, but "slow" fallback, which is consistent and then fsnotify could be improved.

miekg:
I'm hesitant to do *a lot* of refactoring because we haven't fully root caused this yet. I do believe that limiting the number of *fsnotify.Watchers is a good thing (TM) - as there might be some limit somewhere, and the current core discards errors (/me ducks)

stp-ip:
I would push the refactoring of fsnotify back. I would first see it as more important to make the fallback work as expected. The cron like file scanner is currently only picking up creates and deletes, which seems not really like a fallback. If that works, coredns already is fine for probably 95% of the use cases.

After that I feel fsnotify is an optimization to minimize time to see changes as we move from partly working pushes + working pull to fully working pushes.

Limiting the watchers I agree is a good optimization. The usual limit at "/proc/sys/fs/inotify/max_user_watches" is ~8000. Also documenting, that it might be a good idea to actually use sub directories, might be worth it.

_who needs errors..._

miekg:
watcher.Close():  https://github.com/coredns/coredns/blob/master/plugin/file/reload.go#L55
might be *a* culprit to this, that should have actually be watcher.Remove(), although this watcher is per zone... Anyway PR incoming; tests will be tricky though - want to give this a rough pounding.

stp-ip:
I'll look at the changes and see, what I can test. If you have some specific manual tests in mind, I can do my best to go through them within my problematic setup and also locally.

miekg:
https://github.com/coredns/coredns/pull/1089

miekg:
Simple script to move *copy* diff. zone files into place:

~~~
017/09/18 14:25:40 [INFO] CoreDNS-011
[...]

PR #1089, opened 2017-09-18
plugin/{file,auto}: improve reload robustness

miekg:
The reload (fsnotify) mechanism seems to "stop working" for some users,
it has been hard to pin down, what exactly. This PR makes a couple of
changes to help (hopefully) with this:

* Add more logging, including a panic (on startup) when the fsnotify
  can't be set
* Makes the fsnotify.Watcher global (we only need one), and keeps track
  a what directories we watch, as there may be a limit (there is a limit
  somewhere) to how many watcher you can have per directly.
* Only tries reloading a zone when there was a *write* to the zone, so
  moves are not seen anymore.

Testing this was a lot of zones is tricky-ish, so current test-suite
hasn't been updated. This might happen in subsequent comments because
the `watcher` struct does need unit tests.

#1013

### DIFF
--- a/plugin/auto/README.md
+++ b/plugin/auto/README.md
@@ -70,3 +70,7 @@ org {
     }
 }
 ~~~
+
+## Also See
+
+See the README.md of the *file* plugin for details on the reload behavior.
--- a/plugin/auto/zone.go
+++ b/plugin/auto/zone.go
@@ -2,6 +2,7 @@
 package auto
 
 import (
+	"log"
 	"sync"
 
 	"github.com/coredns/coredns/plugin/file"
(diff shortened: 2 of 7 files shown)


miekg:
lol @ failing test:
~~~
% go t -v -run=TestZoneReload
=== RUN   TestZoneReload
--- PASS: TestZoneReload (1.01s)
PASS
~~~

miekg:
Hmmm, the test is showing flakiness even locally:
~~~
[1+] % for i in {0..20}; do go test -v -run=TestZoneReload; done                                         10m 21s ~/g/src/github.com/coredns/coredns/test reload + deb
=== RUN   TestZoneReload
--- FAIL: TestZoneReload (2.01s)
	file_reload_test.go:62: Expected two RR in answer section got 2
FAIL
exit status 1
FAIL	github.com/coredns/coredns/test	2.912s
=== RUN   TestZoneReload
--- PASS: TestZoneReload (2.01s)
PASS
ok  	github.com/coredns/coredns/test	3.682s
=== RUN   TestZoneReload
--- PASS: TestZoneReload (2.01s)
PASS
ok  	github.com/coredns/coredns/test	3.729s
~~~

miekg:
a *shorter* timeout seems to have helped. Hmm, is the file deleted before the tests ends...

miekg:
Closing in favor of #1090

PR #1090, opened 2017-09-18
plugin/{file,auto}: drop fsnotify

miekg:
Try reloading every minute. This is more deterministic then fsnotify.
Also thing cropped up: sharing zone files between zone; there is only
1 fsnotify event and we need to fan out the reload to all zone files.
This is a large rewrite (which could still be done), for now, poll
the zone file on dist.

#1013

### DIFF
--- a/plugin/auto/README.md
+++ b/plugin/auto/README.md
@@ -28,8 +28,8 @@ are used.
   name `db.example.com`, the extracted origin will be `example.com`. **TIMEOUT** specifies how often
   CoreDNS should scan the directory, the default is every 60 seconds. This value is in seconds.
   The minimum value is 1 second.
-* `no_reload` by default CoreDNS will reload a zone from disk whenever it detects a change to the
-  file. This option disables that behavior.
+* `no_reload` by default CoreDNS will try to reload a zone every minute and reloads if the
+  SOA's serial has changed. This option disables that behavior.
 * `upstream` defines upstream resolvers to be used resolve external names found (think CNAMEs)
   pointing to external names. **ADDRESS** can be an IP address, and IP:port or a string pointing to
   a file that is structured as /etc/resolv.conf.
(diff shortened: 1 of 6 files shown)


stp-ip:
Didn't dive into the code yet, but tested with various setups:
* non symlinked directory
  -> works

* symlinked directory (symlink wasn't switched after startup)
  -> works

* symlinked directory (symlink was switched once after startup)
  -> works

* symlinked directory (symlink was switched more than once after startup)
  -> partly fails
  * create is recognized
  * delete is recognized
  * move is recognized
  * change is not recognized

Will see, if I can pinpoint it to anything in the code later today or tomorrow.

miekg:
> symlinked directory (symlink was switched more than once after startup)
-> partly fails
> change is not recognized

Hmmm, but no logging that it failed to read the zone's contents?

stp-ip:
Not at all. Was having a loop creating random files to see, if it was just not picking up something at all, but that worked, just the changes weren't picked up.

stp-ip:
One thing that also came up, is that between a delete from a zone file and the deletion of the zone are a few seconds. I assume cause it can't read the file for the first loop, but only recognized the deletion of the zone within another run?

miekg:
debug println's in plugin/file/reload.go,  plugin/auto/walk.go will probably help you pin down these...

Thanks for testing this btw! Appreciated

stp-ip:
Yeah will hopefully have some more time tomorrow to dig into that.
Thanks for trying to solve it.

stp-ip:
Tested it thoroughly with the result being the change not being recognized was a weird caching issue on my side. Gave the code an extra rigorous test as everything was already in place anyway. Seems to work good. No more issues so far. Will run it internally for the open source projects, but I assume no issues will come up.

I would say lgtm.

PR #1091, opened 2017-09-19
plugin/proxy: detect dead upstream

### LINKED ISSUE #486: middleware/proxy: unhealthy backend per query type/name?
While working on #484 I setup a proxy to 8.8.8.8:53 and then queried `isc.org.` ANY. Someone blocks these queries. This mean the proxy will flag 8.8.8.8 as unhealthy for some amount of time, but other domains and even types for `isc.org.` keep working. Clearly this needs to get more sophisticated, but not completely clear what is the solution.

Do we need to treat domain/8.8.8.8:53 as a tuple and make backend health tick on that? This would mean that `Select()` needs an argument which is the current domain begin looked up. (Or something like this).

miekg:
When the upstream query returns a *network* error, don't mark our
upstream as bad, but instead return an error and return SERVFAIL.
A question remains: is this too lenient ? Should we double check for
timeout?

See the following sequence when querying example.org and isc.org:

::1 - [19/Sep/2017:21:12:48 +0100] "ANY IN isc.org. udp 37 false 4096" SERVFAIL qr,rd 37 5.000933527s
19/Sep/2017:21:12:48 +0100 [ERROR 0 isc.org. ANY] bad upstream 8.8.8.8:53: read udp 192.168.1.148:47547->8.8.8.8:53: i/o timeout
read udp 192.168.1.148:46653->8.8.8.8:53: i/o timeout
127.0.0.1 - [19/Sep/2017:21:12:49 +0100] "ANY IN isc.org. udp 37 false 4096" SERVFAIL qr,rd 37 5.001221543s
19/Sep/2017:21:12:49 +0100 [ERROR 0 isc.org. ANY] bad upstream 8.8.8.8:53: read udp 192.168.1.148:46653->8.8.8.8:53: i/o timeout
::1 - [19/Sep/2017:21:12:50 +0100] "MX IN example.org. udp 41 false 4096" NOERROR qr,rd,ra,ad 104 24.634596ms

Fixes #486

### DIFF
--- a/plugin/proxy/lookup.go
+++ b/plugin/proxy/lookup.go
@@ -5,6 +5,7 @@ package proxy
 import (
 	"context"
 	"fmt"
+	"net"
 	"sync/atomic"
 	"time"
 
@@ -116,6 +117,10 @@ func (p Proxy) lookup(state request.Request) (*dns.Msg, error) {
 
 			if backendErr == nil {
 				return reply, nil
+			}
+			// If we get a network error, *our* upstream is broken, not us, SERVFAIL request to client.
+			if _, ok := backendErr.(*net.OpError); ok {
+				return dns.RcodeServerFailure, fmt.Errorf("bad upstream %s: %s", host.Name, backendErr)
 			}
 			timeout := host.FailTimeout
 			if timeout == 0 {
--- a/plugin/proxy/lookup.go
+++ b/plugin/proxy/lookup.go
@@ -120,7 +120,7 @@ func (p Proxy) lookup(state request.Request) (*dns.Msg, error) {
 			}
 			// If we get a network error, *our* upstream is broken, not us, SERVFAIL request to client.
 			if _, ok := backendErr.(*net.OpError); ok {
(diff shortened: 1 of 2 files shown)


miekg:
Maybe narrow this with (https://golang.org/pkg/net/#OpError) Temporary() and Timeout()

miekg:
Hmm, this still does not detect if the upstream's upstream is dead, just that our upstream has seen an error

johnbelamaric:
Are there any cases where the immediate upstream may decide to simply not return anything? Like DROP in https://tools.ietf.org/html/draft-vixie-dns-rpz-00#section-3 for example. That would still cause it to go to a dead state even if it's not really dead. In that case we would need some kind of "known good query" to make sure it's alive.

johnbelamaric:
I don't see how we could reasonable figure out if the upstream's upstream is dead. I think we just want to make sure that our upstream isn't dead.

miekg:
yes, PR already closed. Your suggestion (or Cricket's) to send a control message is much better.

PR #1093, opened 2017-09-20
WIP: healthcheck: move to DNS checks

miekg:
This removes HTTP healthchecks and moves to DNS (via tcp) system.
This directly queries the upstream DNS server by using a . NS
no-recursive query. Any non-error DNS rcode ?? TODO(miek): if this makes
sense is taken as success.

The proxy and lookup code does not set an upstream unhealthy anymore,
but relies on the healthcheck to do is things.

Obviously tests fails.

### DIFF
--- a/plugin/pkg/dnstest/server.go
+++ b/plugin/pkg/dnstest/server.go
@@ -11,7 +11,8 @@ import (
 type Server struct {
 	Addr string // Address where the server listening.
 
-	s *dns.Server
+	s1 *dns.Server // udp
+	s2 *dns.Server // tcp
 }
 
 // NewServer starts and returns a new Server. The caller should call Close when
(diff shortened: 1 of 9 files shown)


miekg:
@johnbelamaric you think this PR worth pursuing ?

johnbelamaric:
Yes, but if we are doing a lot of QPS, then I am concerned about waiting all the way until the health check. That is, we have to wait for a timeout, then do we try the next upstream? If that's how we do it and we're getting 10k QPS, there will be a lot of timeouts if an upstream goes down.

Maybe we need it to be a little smarter. Like, periodic health checks, or if we see X number of timeouts without a success, fire off a health check?

miekg:
> That is, we have to wait for a timeout, then do we try the next upstream? If that's how we do it and we're getting 10k QPS, there will be a lot of timeouts if an upstream goes down.
>
> Maybe we need it to be a little smarter. Like, periodic health checks, or if we see X number of timeouts without a success, fire off a health check?

Yes, that makes sense, and *should*(TM) be fairly easy to add.

miekg:
Some other idea I'm jotting down:
* We should reuse `Exchanger` to perform the healthcheck so e.g. gRPC also works
* UpstreamHost could get a Health function that gets called for healthchecks

miekg:
closing here, but leaving branch linger for tiny bit for easy copying :)

PR #1098, opened 2017-09-21
pkg: add dnstest

miekg:
Add a full test server impl in this new package + tests. Move
dnsrecorder into this package as well and finish up the commented out
tests that were left in the old dnsrecorder package.

Update all callers.


miekg:
The only bit again this change requires to include dnsstest even though you're not testing; upside is this mimics httptest package a bit.

johnbelamaric:
Can you explain the big picture of why we need this?

miekg:
https://github.com/coredns/coredns/blob/master/plugin/proxy/proxy_test.go#L44

can just use this. If I'm going to move it to dns healthcheck pings (closed PR #1093 used this already)

johnbelamaric:
Ok, I see. thanks.

miekg:
cool. This might also make sense in loads of other tests (instead of having a full blown coredns running on the other side)

PR #1101, opened 2017-09-21
Enable dnstap plugin to insert other plugin's specific data into extra field of tap.Dnstap message

huynhmb359:
### 1. What does this pull request do
This PR enables dnstap plugin to insert other plugin's specific data into extra field of tap.Dnstap message.    

- Dnstap allocates a map where other plugins can add their extra data.   Key is the plugin name and value is the DnstapExtra.    DnstapExtra is a map where extra data per message type are stored.  Key is the message type (i.e.  CQ,  CR  ...etc) .  Value is the wire data in byte array.  

- We define a new structure where the extra data for various plugins can be written into the tap.Dnstap's extra using protobuf.   The structure called Extra.  It's map  with key is the name of the plugin and value is the extra data. 

- When sending out the dnstap message,  dnstap plugin will use the data in the map in the context to construct Extra data and then sends in the tap.DnsTap. 

 2. Which issues (if any) are related?
There is no issue opened for this.

No document change

I'll add UTs after the approach accepted.


miekg:
/cc @varyoo 

I think this can work. Making a new map on every connection is somewhat expensive though - what amount of data can actually be put here? I only see a `[]byte`.

huynhmb359:
For now it's only []byte.  Middleware can put its own serialize data here.

2buisson:
If a middleware wants to send a custom dnstap client response maybe it should do it completely and signal the dnstap middleware not to do it before the response is written.

johnbelamaric:
@varyoo in that case I guess we can get rid of the map creation. on the other hand, only one middleware can add that data, rather than accumulating stuff from all the middleware.

As of today, that works for our use case - we only need to do it with one middleware.

@huynhmb359 can we create the map only if we need it - that is, in the policy MW? Though, then we take a perf hit which we won't in @varyoo's solution. Hmm..

I guess given that we don't *need* multiple middleware to do it today, and there is a performance cost to doing that, we are better off taking @varyoo's suggestion. I think that means sticking a boolean in the context, and having the dnstap plugin skip the CR in that case. Make sense?

huynhmb359:
OK, I will fix according to both @varyoo  and @johnbelamaric  suggestion.

miekg:
you can also just append to a `[][]byte` in the context?

huynhmb359:
Not sure why my recent commit :  https://github.com/huynhmb359/coredns/commit/8b33d0a26b591d7e7637113fac705af7beae361c  
does not go into this PR?    

Any suggestion?   Should I close this PR and recreate?

plugin/dnstap/taprw/writer.go
```diff
@@ -19,12 +26,15 @@ type Tapper interface {
 
 // ResponseWriter captures the client response and logs the query to dnstap.
 // Single request use.
+// DnsTapExtras is map of DnstapExtra.  Key is the name of the middlware that
+// saves the extra data.  Value is a DnstapExtra.
```
johnbelamaric: Fix the comment.

johnbelamaric (changes requested):
Nit: Just fix the comment then this LGTM.

plugin/dnstap/taprw/writer.go
huynhmb359: It's fixed.  Thnx.

johnbelamaric:
I think the test failed due to this merging in upstream miekg/dns: https://github.com/miekg/dns/pull/513

PR #1106, opened 2017-09-23
healhcheck: various cleanups

miekg:
Network wasn't used. IgnorePaths wasn't used. Move checkdown function to
common function shared between proxy protocols. And some naming fixed.


miekg:
1 big functional change: atomic.StoreInt32(&uh.Fails, 0)
we reset the fails on successful HC

Issue #1109, opened 2017-09-25
Pkg/healthcheck: no timeout of Get


miekg:
The health check is doing a Get without a timeout... I suspect this is the reason why it is so bad on a remote not being there. This came about in a or a while back (ping the authors later). 

This should be fixed. And we can probably remove the okUntil, so we only have one metric for health: Fails.

miekg:
This was done in #749 

/cc @ghostflame

ghostflame:
Hi,

So it's worth finding out if the timeout option in the HTTP implementation includes time spend in connect().  If it doesn't, then Get timeout will not solve the problem.  Also, okUntil as a concept means that we have to refresh healthy status, the default is unhealthy (because we don't, for whatever reason, know that it is healthy).

IIRC, HTTP timeout is meant to be taken after issuing the request, and limiting the time a server may take to answer.  So unless the Go implementation is very different, it won't include time spent in tcp connect(), which is where the issue with an absent server is.

I don't know the underlying behaviour, but that is why I didn't try to prevent a fetch taking a long time - I went for a method that worked even if it did.

miekg:
Thanks for chiming in :) 

See https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/ (I trust cloudflare on this):

> The easiest to use is the Timeout field of http.Client. It covers the entire exchange, from Dial (if a connection is not reused) to reading the body.

See  #1110 for a possible simplification.

ghostflame:
Yeah, the way that reads suggests http.Client.Timeout does what you need.

I would be tempted to try out net.Dialer.Timeout set to a pretty aggressive value, maybe 20 msec (or a config'd higher value for very distant remotes) and fail fast - it would have the secondary effect of making a server that is struggling to respond look dead, taking pressure off it.  That's not great in all circumstances - if every server is struggling (but alive) it might declare every upstream dead - but if its made configurable (0 disables dialer timeout checks) then the user can decide.  It's certainly a much clearer indication of an unhappy server than HTTP response.  If the kernel is too busy to handle a TCP connection quickly, leave it alone :-)

miekg:
fixed via #1110

Issue #1035, opened 2017-09-08
mw/proxy


UladzimirTrehubenka:
Faced with the issue:
on the project we use CoreDNS configured for using Unbound resolver as a caching DNS proxy.
If Unbound sends to external resolver some query and it responses with RcodeServerFailure (Unbound returns this answer to CoreDNS), CoreDNS makes decision "unreachable backend" (Unbound) and some queries in interval fail_timeout (default 10 sec) are also resolved as RcodeServerFailure (i.e. ignore proxy backend as down instance). As workaround we just set max_fails to 0 to avoid this.

Would be useful to implement enhancement - add option:
dns_health_check DURATION dns_name1, dns_name2, dns_name3 ...

On the project we use following list: "amazon.com", "google.com", "youtube.com", "facebook.com", "yahoo.com", "apple.com", "microsoft.com" (suppose that these names should be resolved always). The logic is: get first name and make sure that query ret code is RcodeSuccess, if not - get next name until RcodeSuccess or end of the list. This dns_health_check allows determine that proxy endpoint works (queries are resolved) and network route between proxy endpoint and the Internet is fine (option health_check not so useful - looks like inherited from caddy http feature).

miekg:
I think I know what you're hinting at, but RcodeServerFailure shouldn't trigger this - but I should double check

nrvnrvn:
I've hit the same issue.
Getting a bunch of errors like this:

```
[13/Sep/2017:09:57:13 +0000] "A IN notifications.google.com. udp 42 false 512" SERVFAIL 42 171.123µs
13/Sep/2017:09:57:13 +0000 [ERROR 0 notifications.google.com. A] unreachable backend
```

Less than a millisecond duration and eyeballing the tcpdumps show that coredns does not even make calls to the configured dns (google public).

Adding `max_fails 0` seems to really solve the issue.

Version: 010
Config: 
```
.:53 {
    proxy . 8.8.8.8:53 8.8.4.4:53 {
        max_fails 0
    }
    log {
        class error
    }
    errors
}
```

miekg:
Commenting here for further idea on making this more robust (also see #486). 

<strike>Should we also implement retries?</strike> multiple upstream are tried in succession. This probably means that using `proxy . /etc/resolv.conf` is worse than `proxy . 169.254.169.254 169.254.169.254` in GCE for instance.

miekg:
[...]

PR #1110, opened 2017-09-25
plugin/proxy: kick of HC on every 3rd failure

miekg:

Changed the HC back to some older state and allow proxy.go and lookup.go to kick off HCs on every 3rd failure.
Also make the http.Get have a timeout so I won't wait forever (original impetus of pr #749)

#1109 #1035 and PR #749

None yet, although we may need to describe how all various timeout interact.

### DIFF
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -175,7 +175,6 @@ func (k *Kubernetes) getClientConfig() (*rest.Config, error) {
 				HealthCheck: healthcheck.HealthCheck{
 					FailTimeout: 3 * time.Second,
 					MaxFails:    1,
-					Future:      10 * time.Second,
 					Path:        "/",
 					Interval:    5 * time.Second,
 				},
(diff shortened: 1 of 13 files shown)


miekg:
@yongtang  @johnbelamaric

miekg:
Look into this error:
~~~
=== RUN   TestStubLookup
2017/09/25 16:43:03 [WARNING] Host 127.0.0.1:666 health check probe failed: Get : unsupported protocol scheme ""
--- PASS: TestStubLookup (0.01s)
~~~

miekg:
With some more tests with but httptest and dnstest and this should be good to go.

miekg:
this could really use a sanity check @johnbelamaric

johnbelamaric:
I'll try to take a look later today.

johnbelamaric:
I think this is OK. Let me see if I understand - the logic is periodic health checks, unless we see three failures in a row, then we trigger a health check immediately?

miekg:
Yes, that's the general gist of it.

miekg:
@johnbelamaric merge? (Saw I added already to the release notes)

johnbelamaric:
Yes, but what are the failures? (I am on my phone so it's a hard to check)

miekg:
Hmm, the error seems unrelated to this PR. Looking in to it, but weird. https://travis-ci.org/coredns/coredns/jobs/288229221

miekg:
ah, this is again miekg/dns thingy. That would be *really* strange...

Issue #902, opened 2017-08-11
middleware/k8s: add integration tests for k8s+autopath


chrisohaver:
Add autopath+k8s configuration testing to k8s integration testing.

This will require restructuring the test, "fixing" some shortcuts that until now were innocuous...
* Need to execute the queries from within a pod in the cluster (in the test-1 namespace). Currently queries are executed from outside the cluster.
* Should run the main coredns instance inside the cluster, replacing kube-dns.

miekg:
We might want to do these tests completely outside coredns, i.e. different repo that gets pulled in to a running k8s cluster (running somewhere) and performs a bunch of tests.

chrisohaver:
IMO it's just as difficult to set up an external system as it would be to make the travis tests more realistic (currently it kind of "cheats").  Possibly even easier to modify the travis test.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/k8..." ]
>IMO it's just as difficult to set up an external system as it would be to make the travis tests more realistic (currently it kind of "cheats").  Possibly even easier to modify the travis test.

Ok, please elaborate in this bug. I.e. design, what needs changing etc.

/Miek

--
Miek Gieben

chrisohaver:
We should change the following ...

* We should execute the test queries from pods in the cluster. Currently queries are executed from outside the cluster.
* We should run the main coredns instance inside the cluster, replacing kube-dns.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] middleware/k8..." ]
>We should change the following ...
>
>* We should execute the test queries from pods in the cluster. Currently queries are executed from outside the cluster.
>* We should run the main coredns instance inside the cluster, replacing kube-dns.

Ack. So how are we going to run a pod inside a k8s cluster? We somehow build a 
executable from the tests and dockerize that and then start a pod with it?

chrisohaver:
> ... build a executable from the tests and dockerize that and then start a pod with it?

Yes.

miekg:
This is a good idea, but the experimentation of how this will work should be done outside the main coredns repo.

johnbelamaric:
Doing all of this is harder, but querying from pods is more easily done:

```
ANN-M-JBELAMARIC:deploy jbelamaric$ kubectl run dnstools -it --rm --image=infoblox/dnstools --restart=Never -- -c "host kubernetes"
[...]

PR #1111, opened 2017-09-25
plugin/kubernetes: Add in-cluster integration tests 

chrisohaver:

This PR adds adds integration tests for kubernetes+autopath, and migrates most kubernetes tests to execute in-cluster.
Some background:  Currently during integration testing, coredns and test queries are all executed external to the kubernetes cluster. To correctly test kubernetes+autopath, we need an in-cluster deployment.  Specifically, coredns needs to be deployed in place of kube-dns in a kubernetes cluster, and queries need to be executed from within pods in the cluster.  This is a more typical/realistic deployment in general (not just for autopath tests), so most integration tests have been migrated to use this in-cluster deployment. The one exception is the out-of-cluster API failover test, which only applies to an out-of-cluster-deployemnts.

Minikube is used for the in-cluster deployment testing.  I investigated using hyperkube, and real kubernetes single node deployment, both had too many problems (hyperkube acting unstable, real k8s not supported in trusty).

Coredns is deployed in cluster by preparing the docker container and pushing to a local docker repo (which itself runs in a container).  The coredns deployment pulls from the local container repo.

In-cluster queries are made by spinning up a dns/tools pod, executing a query from the pod using `dig`, and then parsing the results into a dns.Msg struct, allowing us to re-use the existing test helpers.

I've split up tests into parallel jobs in travis, the longest taking about 5 minutes typically. 
The in-cluster jobs do a double compile, once for building the .  Figuring out how to avoid that could shave another minute off or so.

#902 

None


.travis/kubernetes/coredns.yaml
```diff
@@ -0,0 +1,123 @@
...
+  selector:
+    k8s-app: coredns
+  clusterIP: 10.0.0.10
+  ports:
+  - name: dns
+    port: 53
+    protocol: UDP
+  - name: dns-tcp
+    port: 53
+    protocol: TCP
```
johnbelamaric, line 123: Missing newline.

johnbelamaric (changes requested):
Looks like pulling the etcd tag out affected the etcd *unit* tests not just the integration tests.

miekg:
I'm sorry: I don't want this merged in the main coredns repo for various reasons:

* double time spend in travis (3 -> 6 minutes)
* multiple timeout loops seen that can wait up to 5 minutes
* more external network dependencies (would like to rid of those, not add more)
* custom code that parses `dig` output and a lot more infrastructure code
* more dependencies on docker
* only works for 1 plugin: *kubernetes*

This should move to an already running cluster where we can pull in the code from a PR (webhooks) and then do the testing we need. For added fun we can do a chaos monkey testing there as well

johnbelamaric:
To start why don't we create a test repo and move these integration tests to there? I think it's still easier to run minikube than have to have an existing running cluster, though we could move to that at some point if we think it makes the most sense.

I think we need to have this level of integration test that actually runs queries from inside a pod.

miekg:
No disagreement about the need for these kind of tests. As said per
previous comment, other repo is fine and gives clearer ownership.

chrisohaver:
@miekg, would you be OK with a solution if  ...
* it builds kubernetes env from scratch each time instead of relying on an already running one.
* coredns repo tests builds and pushes coredns docker image and pushes to docker-hub (in parallel with existing tests, non-failing if push fails), then the ci repo integration test pull the image and runs tests.

miekg:
>@miekg, would you be OK with a solution if  ...

I think we should be very careful adding code into the main coredns repo. If we 
play/polish this in another repo I care much less.

>* it builds kubernetes env from scratch each time instead of relying on an already running one.

What does that buy us?

>* coredns repo tests builds and pushes coredns docker image and pushes to docker-hub (in parallel with existing tests, non-failing if push fails), then the ci repo integration test pull the image and runs tests.

If we are doing it separately from travis, this is all fine. Might move the 
run-etcd checks into this eventually as well, and have travis only do the unit 
tests.

johnbelamaric:
Building the K8s environment means we don't have to rely on something in the environment (an external cluster) to be set up and in a running state. If we use travis we have to set it up every time. If we use Packet we could leave one running, but then we may have issues doing parallel tests and we need to make sure everything is cleaned up before each run.

chrisohaver:
Additionally, building the K8s environment on demand  allows us to test on different versions of k8s more easily.

Issue #1094, opened 2017-09-20
plugin/kubernetes: Update to use protobuf K8s client


johnbelamaric:
Per https://github.com/kubernetes/community/pull/1100#issuecomment-330860057

miekg:
added for 1.0.0

chrisohaver:
I'll look into this one.

aledbf:
@johnbelamaric @chrisohaver  this can help https://github.com/kubernetes/ingress/blob/master/core/pkg/ingress/controller/launch.go#L281

chrisohaver:
merged #1114

PR #1114, opened 2017-09-27
plugin/kubernetes: Enable protobuf, Update client api package

chrisohaver:

* Enables protobuf in k8s api
* Update k8s.io/client-go and dependent vendored packages to latest
* Adjust code to api changes

#1094 

none


.gitignore
```diff
@@ -5,3 +5,9 @@ coredns
 kubectl
 go-test-tmpfile*
 coverage.txt
+.idea/
+.vscode/
+cover.out
+cover.txt
+coverage.out
+myhosts
```
miekg: what's myhost?

miekg:
What commit should we look at?

Can the dep stuff be separated from actual code changes? Maybe even putting a blanket 'go get' in the Makefile to pull in deps and then do the dep prune (all not ideal)

.gitignore
chrisohaver: I didn't intend to commit .gitignore. reverting.

chrisohaver:
I rebuilt the commit history.
code changes are in "code changes"
vendor changes are in "vendor changes"

plugin/kubernetes/kubernetes.go
```diff
@@ -240,6 +242,7 @@ func (k *Kubernetes) initKubeCache(opts dnsControlOpts) (err error) {
 	if err != nil {
 		return err
 	}
+	config.ContentType = "application/vnd.kubernetes.protobuf"
```
miekg: Why don't we set this in k.getClientConfig?

plugin/kubernetes/handler_test.go
```diff
@@ -206,10 +207,10 @@ func (APIConnServeTest) PodIndex(string) []interface{} {
 	return a
 }
 
-func (APIConnServeTest) ServiceList() []*api.Service {
+func (APIConnServeTest) ServiceList() []interface{} {
```
miekg: Why does this need to be moved to empty interface slice? We loose all type safety

miekg (changes requested):
Thanks for splitting  into 2 comments.

miekg:
Thanks, can you squash  the code changes again?

plugin/kubernetes/controller.go
```diff
@@ -341,62 +218,59 @@ func (dns *dnsControl) Stop() error {
...
 		return nil
 	}
-	pods, err := dns.podLister.Indexer.ByIndex(podIPIndex, ip)
+	os, err := dns.podLister.ByIndex(podIPIndex, ip)
 	if err != nil {
 		return nil
 	}
-
+	for _, o := range os {
+		p, ok := o.(*api.Pod)
```
miekg, line 249: again, is this really needed? Or can this API really return something else?

plugin/kubernetes/controller.go
```diff
@@ -341,62 +218,59 @@ func (dns *dnsControl) Stop() error {
...
-}
-
-func (dns *dnsControl) ServiceList() []*api.Service {
-	svcs, err := dns.svcLister.List(labels.Everything())
-	if err != nil {
-		return []*api.Service{}
+func (dns *dnsControl) ServiceList() (svcs []*api.Service) {
+	os := dns.svcLister.List()
+	for _, o := range os {
+		s, ok := o.(*api.Service)
```
miekg, line 231: Are these checks really needed?

miekg (commented):
I see what had these checks before. I always wondered, are they needed? It's an extra loop for a k8s queries, so if then can be dropped, it would matter a tiny bit.

plugin/kubernetes/controller.go
chrisohaver: I dont know. I was coding defensively...  Should I remove the checks?

miekg: dunno... @johnbelamaric any ideas?

miekg: we can remove them later.

miekg (approved):

plugin/kubernetes/controller.go
chrisohaver: > It's an extra loop for a k8s queries..

The loop is required to convert the types to the return type.  This is why I was doing the conversion outside the function originally.

> Are these checks really needed

This i dont know for sure ... But i think that if they always returned one type, the methods would be typed that way, hence the check.

johnbelamaric: I wonder how the well type assertions are optimized by the compiler. I would imagine the overhead would be really small.

johnbelamaric: Also none of this stuff is in-band.

johnbelamaric: I mean, none of this runs for a DNS query, it's all async stuff in other go routines.

johnbelamaric:
Oh never mind, @chrisohaver has corrected me, this is in-band. I was thinking this was in a different part of the code.

Nonetheless, I do think the overhead is likely very low for the type assertions. Would be interesting to check.

PR #1120, opened 2017-09-29
plugin/dnstap: drop payload dropped

miekg:
This spams the travis logs hard and I wonder how useful it is in
practise. Also didn't list the actual error it saw.

PR #1121, opened 2017-09-29
plugin/dnstap: don't spam Travis + valid dnstap payload

2buisson:
In response to: https://github.com/coredns/coredns/pull/1120

### DIFF
--- a/plugin/dnstap/dnstapio/io.go
+++ b/plugin/dnstap/dnstapio/io.go
@@ -1,17 +1,20 @@
 package dnstapio
 
 import (
-	"fmt"
 	"io"
+	"log"
 	"os"
 
 	tap "github.com/dnstap/golang-dnstap"
 	"github.com/golang/protobuf/proto"
 )
 
+func init() {
+	log.SetOutput(os.Stdout)
+}
+
 // DnstapIO wraps the dnstap I/O routine.
 type DnstapIO struct {
-	iolog  io.Writer
 	writer io.WriteCloser
 	queue  chan tap.Dnstap
 	stop   chan bool
@@ -27,7 +30,7 @@ type Protocol interface {
 
 // New dnstap I/O routine from Protocol.
 func New(w Protocol) *DnstapIO {
-	dio := DnstapIO{iolog: os.Stdout}
+	dio := DnstapIO{}
 	dio.writer = w
 	dio.queue = make(chan tap.Dnstap, 10)
 	dio.stop = make(chan bool)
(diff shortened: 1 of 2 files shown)


miekg:
Ah nice :) 

but this (can) potentially flood your local logs... Is it useful to make so much noise when this happens? I.e. if you don't log then... <what happen>?

2buisson:
They should never be invalid, however `log.Println("[WARN] Dnstap payload dropped.")` can be removed.

Is the secondary test error related to dnstap?

plugin/dnstap/dnstapio/io.go
```diff
@@ -1,13 +1,18 @@
...
 	"io"
+	"log"
+	"os"
 
 	tap "github.com/dnstap/golang-dnstap"
 	"github.com/golang/protobuf/proto"
 )
 
+func init() {
+	log.SetOutput(os.Stdout)
```
miekg: I think this is already the default, it's also binary wide if you play with this setting.
Think it can be just removed.

miekg (approved):
Only that nit.

plugin/dnstap/dnstapio/io.go
2buisson: Oops, I wasn't sure it would affect the whole binary or just the package who imports.

Also the default of log is os.Stderr.

plugin/dnstap/dnstapio/io.go
```diff
@@ -38,7 +42,7 @@ func (dio *DnstapIO) Dnstap(payload tap.Dnstap) {
 	select {
 	case dio.queue <- payload:
 	default:
-		fmt.Println("[WARN] Dnstap payload dropped.")
+		dio.log.Println("[WARN] Dnstap payload dropped.")
```
miekg: just use log.Println, no need for all this.

plugin/dnstap/dnstapio/io_test.go
```diff
@@ -26,6 +28,7 @@ func (b buf) Close() error {
 func TestRace(t *testing.T) {
 	b := buf{&bytes.Buffer{}, 100 * time.Millisecond}
 	dio := New(b)
+	dio.log = log.New(ioutil.Discard, "", log.LstdFlags) // don't flood Travis
```
miekg: log.SetOutput(ioutil.Discard)

plugin/dnstap/dnstapio/io.go
2buisson: log is os.Stderr, is it ok to log on standard error?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/dnstap..." ]
>-		fmt.Println("[WARN] Dnstap payload dropped.")
>+		dio.log.Println("[WARN] Dnstap payload dropped.")
>
>log is os.Stderr, is it ok to log on standard error?

We use log.Prin* everywhere. Think we have setup code somewhere, although that 
may only be called when you enabled the log plugin (can't remember)

miekg (approved):

PR #1123, opened 2017-09-29
core: drop invalid packets

miekg:
We can still be on the receiving end of invalid packet. Drop them here.

See #925


miekg:
better safe than sorry.

Issue #1063, opened 2017-09-12
Recommended changes from CII badge


johnbelamaric:
While filling out https://bestpractices.coreinfrastructure.org/projects/1250 to get the CII badge, there were several suggestions we should implement:

* Switch to SemVer
* ~~Document a separate process for vulnerabilities reporting that allows private reporting~~ #1136
* Increase coverage of unit tests and/or figure out away to get a coverage number from the integration tests.
* ~~Document the policy on creation of new tests with new code.~~ already in CONTRIBUTING.md
* ~~Integrate the warnings/lint/vet analysis into the CI.~~
* ~~Integrate dynamic analysis tools (e.g., `gofuzz`) into the build and/or release process.~~ See #1118

johnbelamaric:
I added https://sideci.com which does go vet and lint in the CI (separate from Travis). Let's see if we like it.

johnbelamaric:
I disabled SideCI because it was pointed out to me that it requests far more privileges than seem necessary for the function it provides.

miekg:
* [soon to be done] Switch to SemVer, will be done with our 1.0.0 release
* [needs owner] Document a separate process for vulnerabilities reporting that allows private reporting
* [yeah] Increase coverage of unit tests and/or figure out away to get a coverage number from the integration tests.
* [need owner, and where do we document this] Document the policy on creation of new tests with new code.

miekg:
apart from semver and getting tests coverage higher (which is already high but because of the layout we use the tests in tests/ don't get counted).

I think this can be closed?

PR #1136, opened 2017-10-07
docs: update security issues contact

miekg:
Use team@coredns.io for reporting sec. vulnerablities. Currenttly this
has john@coredns.io and miek@coredns.io


miekg:
See #1063

PR #1143, opened 2017-10-10
plugin/autopath: Add metrics and remove log line

### LINKED ISSUE #1139: plugin/autopath: add namespace
Wouldn't specifying a namespace for *autopath* fix the caching issue (diff. pods will see each other's resolved name if *caching* is used)?

We should also add a metrics that autopath has been done and make the slightly logging better.

miekg:
~~Add a namespace, so autopathing only is performed in this namespace.
This will make caching work for the cluster again.~~

Also export metrics that we've done a successful autopath

Fixes #1139

### DIFF
--- a/plugin/autopath/README.md
+++ b/plugin/autopath/README.md
@@ -19,10 +19,26 @@ autopath [ZONE...] RESOLV-CONF
   plugin. For instance `@kubernetes`, will call out to the kubernetes plugin (for each
   query) to retrieve the search list it should use.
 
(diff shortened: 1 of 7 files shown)


miekg:
need to fix test

miekg:
Still think the metric is useful, let me drop all the rest and keep that bit.

miekg:
Logging on a (potential) per query basis is not the best thing to do, so I like to minimize that.

PR #1149, opened 2017-10-17
Plugin/Kubernetes: Service and Endpoint Indexing

rajansandeep:
Add service and endpoint indexing

#814 

None

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -21,10 +21,18 @@ var (
 )
 
 const podIPIndex = "PodIP"
+const svcNameNamespaceIndex = "NameNamespace"
+const svcIPIndex = "ServiceIP"
+const epNameNamespaceIndex = "EndpointNameNamespace"
+const epIPIndex = "EndpointsIP"
 
 type dnsController interface {
 	ServiceList() []*api.Service
+	SvcIndex(string) []*api.Service
+	SvcIndexReverse(string) []*api.Service
 	PodIndex(string) []*api.Pod
+	EpIndex(string) []*api.Endpoints
+	EpIndexReverse(string) []*api.Endpoints
 	EndpointsList() []*api.Endpoints
 
 	GetNodeByName(string) (*api.Node, error)
@@ -44,7 +52,7 @@ type dnsControl struct {
 
 	svcLister cache.Indexer
 	podLister cache.Indexer
-	epLister  cache.Store
+	epLister  cache.Indexer
 
 	// stopLock is used to enforce only a single call to Stop is active.
 	// Needed because we allow stopping through an http endpoint and
@@ -77,7 +85,7 @@ func newdnsController(kubeClient *kubernetes.Clientset, opts dnsControlOpts) *dn
 		&api.Service{},
 		opts.resyncPeriod,
 		cache.ResourceEventHandlerFuncs{},
-		cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+		cache.Indexers{svcNameNamespaceIndex: svcNameNamespaceIndexFunc, svcIPIndex: svcIPIndexFunc})
 
 	if opts.initPodCache {
 		dns.podLister, dns.podController = cache.NewIndexerInformer(
@@ -90,14 +98,15 @@ func newdnsController(kubeClient *kubernetes.Clientset, opts dnsControlOpts) *dn
 			cache.ResourceEventHandlerFuncs{},
 			cache.Indexers{podIPIndex: podIPIndexFunc})
 	}
-	dns.epLister, dns.epController = cache.NewInformer(
+	dns.epLister, dns.epController = cache.NewIndexerInformer(
 		&cache.ListWatch{
 			ListFunc:  endpointsListFunc(dns.client, namespace, dns.selector),
 			WatchFunc: endpointsWatchFunc(dns.client, namespace, dns.selector),
 		},
 		&api.Endpoints{},
 		opts.resyncPeriod,
-		cache.ResourceEventHandlerFuncs{})
+		cache.ResourceEventHandlerFuncs{},
+		cache.Indexers{epNameNamespaceIndex: epNameNamespaceIndexFunc, epIPIndex: epIPIndexFunc})
 
 	return &dns
 }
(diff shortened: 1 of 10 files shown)


plugin/kubernetes/controller.go
```diff
@@ -21,10 +21,18 @@ var (
 )
 
 const podIPIndex = "PodIP"
+const svcNameNamespaceIndex = "NameNamespace"
```
miekg, line 24: More Go-like is to group these:
```
const (
   ...
)
```

plugin/kubernetes/controller.go
```diff
@@ -110,6 +119,38 @@ func podIPIndexFunc(obj interface{}) ([]string, error) {
 	return []string{p.Status.PodIP}, nil
 }
 
+func svcIPIndexFunc(obj interface{}) ([]string, error) {
```
miekg, line 122: Do these really need to take an empty interface?

plugin/kubernetes/controller.go
```diff
@@ -255,6 +296,79 @@ func (dns *dnsControl) PodIndex(ip string) (pods []*api.Pod) {
...
+func (dns *dnsControl) EpIndex(idx string) (ep []*api.Endpoints) {
+	if dns.epLister == nil {
+		return nil
+	}
+	os, err := dns.epLister.ByIndex(epNameNamespaceIndex, idx)
+	if err != nil {
+		return nil
+	}
+	for _, o := range os {
+		e, ok := o.(*api.Endpoints)
```
miekg, line 345: Are all these type checks really needed?

plugin/kubernetes/kubernetes.go
```diff
@@ -342,8 +342,18 @@ func (k *Kubernetes) findPods(r recordRequest, zone string) (pods []msg.Service,
...
-	for _, svc := range k.APIConn.ServiceList() {
+	var (
+		endpointsList []*api.Endpoints
+		serviceList   []*api.Service
+		idx           string
+	)
+	if wildcard(r.service) || wildcard(r.namespace) {
+		serviceList = k.APIConn.ServiceList()
+	} else {
+		idx = r.service + "." + r.namespace
```
miekg, line 353: dnsutil.Join()?

plugin/kubernetes/kubernetes.go
```diff
@@ -357,8 +367,13 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 
 		// Endpoint query or headless service
 		if svc.Spec.ClusterIP == api.ClusterIPNone || r.endpoint != "" {
-
-			for _, ep := range k.APIConn.EndpointsList() {
+			if wildcard(r.service) || wildcard(r.namespace) {
```
miekg, line 370: Is the check from line 350 now done again?

plugin/kubernetes/ns.go
```diff
@@ -23,7 +23,7 @@ func (k *Kubernetes) nsAddr() *dns.A {
 	rr.A = localIP
 
 FindEndpoint:
-	for _, ep := range k.APIConn.EndpointsList() {
+	for _, ep := range k.APIConn.EpIndexReverse(localIP.String()) {
```
miekg, line 26: Is it an idea to make `EpIndexReverse` take a `net.Addr`?

miekg:
these empty interfaces are not good and set a bad example for code in CoreDNS.
Is anyone volunteering to clean this up? Otherwise I'll probably need to revert.

miekg:
A revert of this passes travis: https://travis-ci.org/coredns/coredns/builds/290356063

johnbelamaric:
Yeah, this is a performance fix so it will pass without it. But the QPS and latency with 10,000 services go from 10k and 0.1ms to 400 and 1000ms without this. Because it is then looping through all (well half on average) services for every query.

I believe the empty interfaces are due to the way we have to interface to the the K8s client-go code. But we need real types so then we have to do type assertions so we can access the data on there.

On Oct 20, 2017, at 4:59 AM, Miek Gieben <notifications@github.com<mailto:notifications@github.com>> wrote:


A revert of this passes travis: https://travis-ci.org/coredns/coredns/builds/290356063

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Plugin/Kubern..." ]
>Yeah, this is a performance fix so it will pass without it. But the QPS and latency with 10,000 services go from 10k and 0.1ms to 400 and 1000ms without this. Because it is then looping through all (well half on average) services for every query.
>
>I believe the empty interfaces are due to the way we have to interface to the the K8s client-go code. But we need real types so then we have to do type assertions so we can access the data on there.

Just checked that. Yes they have an empty interface in there... that's pretty 
terrible

miekg:
Still leaves the failing test:
~~~
== RUN TestKubernetes
--- FAIL: TestKubernetes (3.00s)
	helpers.go:115: rcode is "NXDOMAIN", expected "NOERROR"
	helpers.go:291: ;; opcode: QUERY, status: NXDOMAIN, id: 41324
	;; QUESTION SECTION:
	;*.svc-1-a.test-1.svc.cluster.local.	IN	SRV
	;; AUTHORITY SECTION:
	cluster.local.	300	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1508500831 7200 1800 86400 60
~~~
Suspect we may be faster/slower to start which makes this fail?

johnbelamaric:
Yeah. This is why we need the readiness checks. Think we should have it before 1.0.0

On Oct 20, 2017, at 8:07 AM, Miek Gieben <notifications@github.com<mailto:notifications@github.com>> wrote:


Still leaves the failing test:

== RUN TestKubernetes
--- FAIL: TestKubernetes (3.00s)
        helpers.go:115: rcode is "NXDOMAIN", expected "NOERROR"
        helpers.go:291: ;; opcode: QUERY, status: NXDOMAIN, id: 41324
        ;; QUESTION SECTION:
        ;*.svc-1-a.test-1.svc.cluster.local.    IN      SRV
        ;; AUTHORITY SECTION:
        cluster.local.  300     IN      SOA     ns.dns.cluster.local. hostmaster.cluster.local. 1508500831 7200 1800 86400 60


Suspect we may be faster/slower to start which makes this fail?

PR #1152, opened 2017-10-18
plugin/kubernetes: Modify integration tests for coredns/ci

chrisohaver:

Modifications to kubernetes integration tests for integration with coredns/ci.
This is paired with PR coredns/ci#2

#902

nothing user facing.  may need to update dev docs


test/kubernetes/a_test.go
```diff
@@ -0,0 +1,105 @@
+// +build ci
+
+package kubernetes
+
+import (
+	"testing"
+
+	"github.com/coredns/coredns/plugin/test"
+
+	testk8s "github.com/coredns/ci/test/kubernetes"
```
miekg: fmt

test/kubernetes/a_test.go
```diff
@@ -0,0 +1,105 @@
+// +build ci
```
miekg: address_test.go would be more descriptive.

test/kubernetes/a_test.go
```diff
@@ -0,0 +1,105 @@
...
+            upstream ` + udp + `
+        }
+    }
+`
+
+	err := testk8s.LoadCorefile(corefile)
+	if err != nil {
+		t.Fatalf("Could not load corefile: %s", err)
+	}
+	testk8s.DoIntegrationTests(t, dnsTestCasesA, "test-1")
```
miekg: Can we please stop hidding away the test and just put the code here. I'm assuming that this  (DoIntegrationTests) does all the sorting, including any CNAME sorting by default.

Just copy those few lines here.

test/kubernetes/a_test.go
```diff
@@ -0,0 +1,105 @@
...
+        errors
+        log
+        kubernetes cluster.local 10.in-addr.arpa {
+            namespaces test-1
+            upstream ` + udp + `
+        }
+    }
+`
+
+	err := testk8s.LoadCorefile(corefile)
```
miekg: why can't we use the functions we have in test ?

miekg (changes requested):

test/kubernetes/a_test.go
chrisohaver: These are the same functions, moved to the ci repo, as you had requested.

test/kubernetes/a_test.go
chrisohaver: the A is for type "A" .... but I can change it to "address"

test/kubernetes/a_test.go
chrisohaver: can you clarify the comment?

test/kubernetes/a_test.go
chrisohaver: `DoIntegrationTests` is ~50  lines. It does the following...

- Starts the dns client pod
- The for each test case: 
  - Builds/executes dig command 
  - parses response, sorts, and compares to expected

test/kubernetes/a_test.go
chrisohaver: To clarify: These are "test framework" functions, (like dig parsing), which are in the ci repo.

test/kubernetes/a_test.go
chrisohaver: Do you mean that the `github.com/coredns/ci/test/kubernetes` import should be grouped with the `github.com/coredns/coredns/plugin/test` import?

test/kubernetes/a_test.go
chrisohaver: I can reorganize things so that the tests themselves are executed in the test files...

chrisohaver: To be more precise, you want that all test failures and errors are generated directly from the X_test.go files?

Makefile
```diff
@@ -54,6 +54,9 @@ ifeq ($(TEST_TYPE),coverage)
 	done
 endif
 
+.PHONY: ci
```
miekg: Why can't we re-use the existing tag?

I'm sorry, I completely fail to see why (at this stage) we need changes *at all* to coredns' source

johnbelamaric: Chris is trying to get the K8s integration testing out of the main repo and out of the main Travis build, which is what you asked for. I think he would have been happy to leave it all in here, but is trying to make it work in the way we understood you wanted. Maybe you can provide more detail on exactly what you are looking for, so we can stop going back and forth on this and wasting time?

miekg:
>+.PHONY: ci
>
>Chris is trying to get the K8s integration testing out of the main repo and out of the main Travis build, which is what you asked for. I think he would have been happy to leave it all in here, but is trying to make it work in the way we understood you wanted. Maybe you can provide more detail on exactly what you are looking for, so we can stop going back and forth on this and wasting time?

Ack. So proposal:

1. copy the few k8s test to the ci repo
2. get the kinks worked out there
3. remove from coredns/coredns
4. move etcd test as well

I'm wondering how white boxy the intergration tests will be anyway. I.e. how far 
off will those be from starting CoreDNS with a corefile and replying queries? 
That would imply you *only* need the latest coredns binary (possibly as a 
docker container).

Open question is where should reply data live, i.e in coredns/ci or 
coredns/coredns?

johnbelamaric:
> That would imply you *only* need the latest coredns binary (possibly as a
docker container).

We want to run these against PRs. You comment on the PR /integration and it will run them (@chrisohaver is that right or is it automatic?)

chrisohaver:
> We want to run these against PRs. You comment on the PR /integration and it will run them (@chrisohaver is that right or is it automatic?)

That is correct.

PR #1153, opened 2017-10-19
plugin/dnssec; insert and sign DS records

### LINKED ISSUE #698: middleware/dnssec: sign the DS in the delegation
If we sign a delegation, we return as-is. But there should be a DS and sig. Now should the DS be in the on-fly-signed zone or should we generated it (we have the key) and then insert+sign?

Questions... Questions.. . But should be fixed/looked into.

miekg:
Sign a delegation as well and insert DS records.

Fixes #698


miekg:
cc @Isolus you might have opinion on this.

PR #1155, opened 2017-10-20
plugin/kubernetes: implement HasSynced() 

### LINKED ISSUE #1006: mw/kubernetes: block until ready?
The kubernetes tests have this:
~~~
	// Work-around for timing condition that results in no-data being returned in test environment.
	time.Sleep(3 * time.Second)
~~~
Which means that even in our code we need a workaround. This should be replaced with either a call to a blocking channel that gets closed once we've fully started or something else.

This is also needed for #985

miekg:

Implements HasSynced() that holds startup until the API has fully synced. Removing the race in the startup we now have.

Fixes #1006

### DIFF
--- a/plugin/kubernetes/setup.go
+++ b/plugin/kubernetes/setup.go
@@ -39,8 +39,14 @@ func setup(c *caddy.Controller) error {
 	c.OnStartup(func() error {
 		go kubernetes.APIConn.Run()
 		if kubernetes.APIProxy != nil {
-			go kubernetes.APIProxy.Run()
+			kubernetes.APIProxy.Run()
 		}
+		synced := false
+		for synced == false {
+			synced = kubernetes.APIConn.HasSynced()
+			time.Sleep(100 * time.Millisecond)
+		}
+
 		return nil
 	})
 
--- a/plugin/federation/kubernetes_api_test.go
+++ b/plugin/federation/kubernetes_api_test.go
@@ -9,6 +9,7 @@ import (
 
 type APIConnFederationTest struct{}
 
+func (APIConnFederationTest) HasSynced() bool                        { return true }
 func (APIConnFederationTest) Run()                                   { return }
 func (APIConnFederationTest) Stop() error                            { return nil }
 func (APIConnFederationTest) SvcIndexReverse(string) []*api.Service  { return nil }
--- a/plugin/kubernetes/apiproxy.go
+++ b/plugin/kubernetes/apiproxy.go
@@ -65,7 +65,9 @@ func (p *proxyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 func (p *apiProxy) Run() {
 	p.handler.Start()
-	p.Serve(p.listener)
+	go func() {
+		p.Serve(p.listener)
+	}()
 }
 
 func (p *apiProxy) Stop() {
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -38,6 +38,7 @@ type dnsController interface {
 	GetNodeByName(string) (*api.Node, error)
 
 	Run()
+	HasSynced() bool
 	Stop() error
 }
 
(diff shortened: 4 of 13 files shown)


fturib:
Hum.. this PR includes change for optimization on endPoint list index ... @rajansandeep : can you check that this change has no side-effect ?

miekg:
@fturib  you mean this commit: https://github.com/coredns/coredns/pull/1155/commits/ec84ff2d99ebd47c9fc2917982927a29e83f6a51 ?

rajansandeep:
@fturib, the change is a cleanup and has no side effect.

miekg:
hmmm, keep being flaky
~~~
-- FAIL: TestKubernetes (0.23s)
	helpers.go:115: rcode is "NXDOMAIN", expected "NOERROR"
	helpers.go:291: ;; opcode: QUERY, status: NXDOMAIN, id: 62757
	;; QUESTION SECTION:
	;*.svc-1-a.test-1.svc.cluster.local.	IN	SRV
	;; AUTHORITY SECTION:
	cluster.local.	300	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1508510551 7200 1800 86400 60
~~~
faster flaky though

miekg:
breaks on different questions, which makes sense because the the `range` we use in the tests. So even with HasSynced() we need to wait for a bit?
~~~
-- FAIL: TestKubernetes (0.21s)
	helpers.go:115: rcode is "NXDOMAIN", expected "NOERROR"
	helpers.go:291: ;; opcode: QUERY, status: NXDOMAIN, id: 7853
	;; QUESTION SECTION:
	;headless-svc.test-1.svc.cluster.local.	IN	A
	;; AUTHORITY SECTION:
	cluster.local.	300	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1508513248 7200 1800 86400 60
=== RUN TestMetricsServer
~~~

miekg:
believe this is still a good PR to merge, not complete OK with the sleep(1s) that is still needed, but an improvement over the previous behavior.

fturib:
@miekg : we reviewed the change about endPointList() in findServices() of kubernetes.go. And would advise to revert your commit : it was somehow wrong already, and your change will make another wrong doing. I am opening an issue to take care of this findService() issue.

miekg:
Ok. sgtm.

Why about the HasSynced change?

miekg:
Note: nothing has been commited, you mean we should revert somethting?

fturib:
I am referring only to the changes in 9  plugin/kubernetes/kubernetes.go. it is about the optimization in findService()  (or cleanup), and seems not linked to HasSynced() feature.

I am not sure about the wording.. but I see that change in the commit https://github.com/coredns/coredns/pull/1155/commits/da0c8edced16dc82c1dfa4bea4d8509bf7883e86, although it is not merged.

miekg:
I don't fully understand. If it was wrong already, can you send a PR on how it should look.

The change I made in this PR is a noop - so I'm not sure what you're referring at

fturib:
I am doing the change in same time locally. 
Eventually, you are right,  it is a noop, only impacting efficiency. 
So .. finally you can keep. 

I am doing the change locally, but needs to run the Tests, and then find how to setup a PR. 
It will be my first .. so will take some time :)

miekg:
Ok, no problem.

So what about this pr? Merge? Or back out and wait for yours?
(we're "only" pushing to master so we can play around a bit :))

fturib:
For me you can merge.

fturib:
I opened an issue here for my concern on endPointList : #1156

fturib:
hum .. you may think I am a little instable, but looking into the code you propose here, I think you made a 'maybe' infinite loop in the Setup function.

=> you may want to break the wait of readiness after an amount of time or loops - and raise an Error is not HasSynced().

plugin/kubernetes/setup.go
```diff
@@ -39,8 +39,14 @@ func setup(c *caddy.Controller) error {
 	c.OnStartup(func() error {
 		go kubernetes.APIConn.Run()
 		if kubernetes.APIProxy != nil {
-			go kubernetes.APIProxy.Run()
+			kubernetes.APIProxy.Run()
 		}
+		synced := false
+		for synced == false {
```
aledbf, line 45: Please check https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go#L104

You can use
```
        if !cache.WaitForCacheSync(stopCh,
		kubernetes.APIConn.HasSynced,
	) {
		return fmt.Errorf("Timed out waiting for caches to sync")
	}
```
from package `k8s.io/client-go/tools/cache`

Basically this waits until the cache has synced.

miekg: Interesting. I'll check tomorrow - might still merge this and then follow up.

PR #1159, opened 2017-10-21
CIDR query routing

### LINKED ISSUE #1074: Classless reverse DNS handling
Currently, you can specify a reverse zone in the Corefile with the CIDR notation (e.g., 10.0.0.0/8). However, if the network is not on a classful boundary, it is silently ignored.

In the K8s use cases in particular, we would like to be able to serve PTR records for the address space that K8s manages, but proxy other PTR requests upstream. This is how the `cidrs` feature worked in the past, before we added the CIDR notation for zones.

I think that if there is a classless zone defined, we could automatically round it to the containing classful boundary. We could retain the original CIDR though, and in the middleware we can test if the request falls outside that original CIDR. If so, we `fallthrough`. This could be automatic or it could be an option we set on a given middleware (e.g., `ptr_fallthrough` or something).

If multiple reverse zone CIDRs are listed in the ZONES, then we should find the minimal set of containing classful zones and adjust to those, storing away all the specific CIDRs.

The end logic of doing the PTR fall through would need to be implemented in each middleware, but most of this should be able to be done in a pkg to make it easy to implement per middleware.

miekg:
Enables CIDR routing of queries on non `x%8==0` boundaries.

Fixes #1074

Yes, this needs to be documented, but it mostly should just work as intended.  But all docs on coredns.io need updating to this.


miekg:
No impact on perf:
~~~
classless:
pkg: github.com/coredns/coredns/core/dnsserver
BenchmarkCoreServeDNS-4   	 1000000	      1431 ns/op	      16 B/op	       1 allocs/op

pkg: github.com/coredns/coredns/core/dnsserver
BenchmarkCoreServeDNS-4   	 1000000	      1429 ns/op	      16 B/op	       1 allocs/op

master:
pkg: github.com/coredns/coredns/core/dnsserver
BenchmarkCoreServeDNS-4   	 1000000	      1412 ns/op	      16 B/op	       1 allocs/op

pkg: github.com/coredns/coredns/core/dnsserver
BenchmarkCoreServeDNS-4   	 1000000	      1429 ns/op	      16 B/op	       1 allocs/op
~~~
This does not test an (e.g.) /17 because there you would pay the price.

miekg:
Probably can do w/ some extra corner case testing, but main code should be good to go.

(/cc @chrisohaver / @johnbelamaric )

chrisohaver:
Looks good.  One thing, perhaps filter func should take query string and query type, so the func can return before trying to extract addresses from non PTR request query string...

miekg:
Good point and I thought about that as well. But a reverse zone is just a normal DNS zone and SOA and NS queries are valid as well (or whatever you put in there).

Making FilterFunc more general is thing we can do of course, but I like doing the minimal thing for the current problem.

chrisohaver:
Ah, I wasn't trying to generalize the function. I was trying to suggest an optimization.  In the proposed implementation, every query, regardless of type will get run through `dnsutil.ExtractAddressFromReverse`...

I have only seen PTR types for reverse queries.  Can you give an example of a non-PTR reverse zone query?

miekg:
~~~
 dig -x 8.8.8.8 soa                                                                                                              ~/g/src/github.com/coredns/coredns/test classless

; <<>> DiG 9.10.3-P4-Debian <<>> -x 8.8.8.8 soa
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 47247
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0

;; QUESTION SECTION:
;8.8.8.8.in-addr.arpa.		IN	SOA

;; AUTHORITY SECTION:
8.8.8.in-addr.arpa.	60	IN	SOA	ns2.google.com. dns-admin.google.com. 173100588 900 900 1800 60
~~~

miekg:
ExtractReverseFrom is broken because it checks  the validity of the address, which it shouldn't

miekg:
grr. dns.ReverseAddr does the same.

PR #1168, opened 2017-10-24
plugin/kubernetes: lazy initialze EndPointsList

### LINKED ISSUE #1156: plugin/kubernetes - improve performance in findService()
1- compute wildcard status for a service is costly and we call it twice. => save info in boolean.
2- building endPoint list of the Service is also costly and useful only when we met an endPoint query or an headless service. => change into a lazy initialization.

miekg:
If we don't need it, don't initialize it endPointsList

Fixes #1156


miekg:
sorry, I was bored and wanted to squash the last bug in the 1.0.0 milestone. PTAL @fturib

fturib (approved):
More elegant than the changes I started locally.
Good.

PR #1169, opened 2017-10-24
plugin/hosts: Change lock from a Mutex to a RWMutex

pmoroney:
Changes the lock on the hosts file data from a Mutex to a RWMutex.
This allows multiple concurrent reads from the hosts file.

N/A

N/A

### DIFF
--- a/plugin/hosts/hostsfile.go
+++ b/plugin/hosts/hostsfile.go
@@ -68,11 +68,17 @@ type Hostsfile struct {
 func (h *Hostsfile) ReadHosts() {
 	now := time.Now()
 
+	// Not deferring h.RUnlock() because we may need to remove the read lock and aquire a write lock
+	h.RLock()
 	if now.Before(h.expire) && len(h.byAddr) > 0 {
+		h.RUnlock()
 		return
 	}
 	stat, err := os.Stat(h.path)
 	if err == nil && h.mtime.Equal(stat.ModTime()) && h.size == stat.Size() {
+		h.RUnlock()
+		h.Lock()
+		defer h.Unlock()
 		h.expire = now.Add(cacheMaxAge)
 		return
 	}
@@ -83,10 +89,14 @@ func (h *Hostsfile) ReadHosts() {
 		if len(h.byAddr) == 0 && len(h.inline) > 0 {
 			h.Parse(nil)
 		}
+		h.RUnlock()
 		return
 	}
 	defer file.Close()
 
+	h.RUnlock()
+	h.Lock()
+	defer h.Unlock()
 	h.Parse(file)
 
 	// Update the data cache.
@@ -139,8 +149,6 @@ func (h *Hostsfile) Parse(file io.Reader) {
 			is[addr.String()] = append(is[addr.String()], name)
 		}
 	}
-	h.Lock()
-	defer h.Unlock()
 	h.byNameV4 = hsv4
 	h.byNameV6 = hsv6
 	h.byAddr = is
--- a/plugin/hosts/hostsfile.go
+++ b/plugin/hosts/hostsfile.go
@@ -36,7 +36,7 @@ func absDomainName(b string) string {
 
 // Hostsfile contains known host entries.
 type Hostsfile struct {
-	sync.Mutex
+	sync.RWMutex
 
 	// list of zones we are authoritive for
 	Origins []string
@@ -139,6 +139,8 @@ func (h *Hostsfile) Parse(file io.Reader) {
 			is[addr.String()] = append(is[addr.String()], name)
 		}
 	}
+	h.Lock()
+	defer h.Unlock()
 	h.byNameV4 = hsv4
 	h.byNameV6 = hsv6
 	h.byAddr = is
@@ -159,8 +161,8 @@ func ipVersion(s string) int {
 
 // LookupStaticHostV4 looks up the IPv4 addresses for the given host from the hosts file.
 func (h *Hostsfile) LookupStaticHostV4(host string) []net.IP {
-	h.Lock()
-	defer h.Unlock()
+	h.RLock()
+	defer h.RUnlock()
 	h.ReadHosts()
 	if len(h.byNameV4) != 0 {
 		if ips, ok := h.byNameV4[absDomainName(host)]; ok {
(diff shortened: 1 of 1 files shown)


plugin/hosts/hostsfile.go
```diff
@@ -159,9 +161,9 @@ func ipVersion(s string) int {
 
 // LookupStaticHostV4 looks up the IPv4 addresses for the given host from the hosts file.
 func (h *Hostsfile) LookupStaticHostV4(host string) []net.IP {
-	h.Lock()
-	defer h.Unlock()
 	h.ReadHosts()
```
miekg, line 172: `h.ReadHosts()` can execute without a lock?

pmoroney: `h.Readhosts()` only checks to see if the file has been modified, if it has it calls `h.Parse()`. Which is the only function that modifies the maps. I added the `h.Lock()` to that function.

pmoroney: Hmm.. looking at it closer, it does get the len() of one of the maps. I was hoping to limit the write lock to only writing the maps, but maybe I should take it in at the beginning of `h.ReadHosts()`.

miekg (dismissed):

plugin/hosts/hostsfile.go
miekg: Ah, I pressed approve too soon :) Yes, len need RLock

pmoroney: Sorry about that. I am fixing that and making it cleaner.

plugin/hosts/hostsfile.go
```diff
@@ -68,11 +68,17 @@ type Hostsfile struct {
...
+	h.RLock()
 	if now.Before(h.expire) && len(h.byAddr) > 0 {
+		h.RUnlock()
 		return
 	}
 	stat, err := os.Stat(h.path)
 	if err == nil && h.mtime.Equal(stat.ModTime()) && h.size == stat.Size() {
+		h.RUnlock()
+		h.Lock()
+		defer h.Unlock()
```
miekg, line 81: This `defer` is a bit superfluous as your return 2 lines below, I would just sandwitch the `h.expire ...` line between lock and unlock

pmoroney: Sure thing

pmoroney:
Ack! I just ran `git push` instead of `git push pmoroney` to go to my fork. Is it supposed to merge that automatically?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/hosts:..." ]
>Ack! I just ran `git push` instead of `git push pmoroney` to go to my fork. Is it supposed to merge that automatically?

Yes...?! Dunno, but you somehow merged this now :)

/Miek

--
Miek Gieben

miekg:
Instead of reverting this - it works, I suggest doing a followup PR with the (minor) changes.

pmoroney:
I was attempting to push the minor changes, it looked like it just applied it without attaching it to the PR:
https://github.com/coredns/coredns/commit/25367a43296bf681c984121ae7f40e08f32508a8

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/hosts:..." ]
>I was attempting to push the minor changes, it looked like it just applied it without attaching it to the PR:
>https://github.com/coredns/coredns/commit/25367a43296bf681c984121ae7f40e08f32508a8

Maybe because you're editing on the master branch as well. Best to branch 
locally.

If you put WIP: in front the PR it can't get merged, but I think that only holds 
true if you do merges via the Github UI.

miekg:
Uh @pmoroney can you stop pushing on this branch?
* 25367a43 
* 53d9bff7 
* 680e6bd5 

All seem directly pushed to master? We usually squash all these into 1 to have a clearer history (and confidence that everything builds, such that git bisect will be easier).

So create a non-master branch locally, make changes, send PR and keep pushing on that branch. In the end we squash when we merge. Or you can locally squash with `git rebase -i HEAD~X` where X the number of commits you have on that branch.

pmoroney:
I only did the one push accidentally. I have since removed that repo and re-cloned it from my fork, so `git push` won't accidentally push to where `go get` clones from. I think it just took all 4 changesets I had locally and pushed them to master. Three of them were already in my fork which is where I created the PR from. And I am guessing it marked this PR as merged since the commits were already in master.

I had been pushing to master in my fork, but I am going to start pushing to a branch in my fork instead in order to have 2 barriers to master. 

Also, it looks like Github has a setting for "Require pull request reviews before merging" if you add `master` as a protected branch. I think that might prevent this kind of thing from happening again.

Sorry for messing all that up.

miekg:
Ok, np we have a few extra commits.

While I'm going over the code - I do wonder about all locking. This keeps tripping me up:
~~~
	h.ReadHosts()
	h.RLock()
	defer h.RUnlock()
~~~
I think it's better remove remove `h.ReadHosts()` from all these lookup functions and have it running as a separate go-routine which is started with OnStartup (and stopped with OnShutdown) in the setup function. This leaves all the lookup functions with a lock-and-defer at the top that looks nice.

Speaking of `ReadHost`, I don't fully understand what's going from hostfile.go#L86 onwards
~~~
	var file *os.File
	if file, _ = os.Open(h.path); file == nil {
		// If this is the first time then we will try to parse inline
		if len(h.byAddr) == 0 && len(h.inline) > 0 {
			h.Parse(nil)
		}
		h.RUnlock()
		return
	}
	defer file.Close()
~~~
The err from os.Open seems unhandled, but I think that is OK. I don't see what `file == nil` is doing here, makes more sense to just tick of of the error (which is ingored)? Also h.Parse is called without a write lock.

And above the error from os.Stat is only checked for `nil`-ness. Meaning the last statement of this function can set garbage: `h.mtime` and `h.size` if the stat failed.

Seems to me that the whole first part L71-L85 can be put in a separate func `shouldUpdate` or something that checks if the re-read should be done, return the stat data and an error - that would at last split the all the locking and unlocking and somewhat distinct pieces.

pmoroney:
Ok, I can move that to a separate go-routine. I assume OnStartup happens after Setup?

That weird bit from `ReadHost` is there to handle the first load where there is an inline hosts file in the config and the file listed doesn't exist. I don't like it either. I tried creating a `shouldUpdate` function yesterday, but it was a little messy still because of that case. But I think moving the update to a go-routine that can sleep until needed will clean all that up.

I'll work on that when I get a chance today.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/hosts:..." ]
>Ok, I can move that to a separate go-routine. I assume OnStartup happens after Setup?

yes and before the server comes lives.

>That weird bit from `ReadHost` is there to handle the first load where there is an inline hosts file in the config and the file listed doesn't exist. I don't like it either. I tried creating a `shouldUpdate` function yesterday, but it was a little messy still because of that case. But I think moving the update to a go-routine that can sleep until needed will clean all that up.
>
>I'll work on that when I get a chance today.

Ok, cool. And thanks!

PR #1172, opened 2017-10-25
plugin/dnstap: some cleanup

miekg:
Some cleanup in proxy and dnstap:
* just use time pkg directly and side step the indirection for Epoch
* Use Set in SetQueryEpoch to be more Go like. (Looked like a reader)
* Don't maintain two sets of time, we already track start, so use that.
* Use time.Time and convert when needed
* dedent the toDnstap function and put in a separate file


miekg:
trigger because I'm adding dnstap support to miekg/forward
(tracer is not needed because no gRPC support there)

2buisson (approved):
It's only best!

miekg:
Fixed the tests.

PR #1173, opened 2017-10-25
WIP: plugin/kubernetes: unflak tests

miekg:
Make this less (hopefully non) flaky.


johnbelamaric:
Are these moving to the CI, then? Because we're losing a lot of coverage if not.

miekg:
They should (in the end). I rather figure why they are flaky, but I can see why or how.

johnbelamaric:
These are removed in https://github.com/coredns/coredns/pull/1152 and moved I think.

PR #1174, opened 2017-10-25
plugin/dnstap: create ToMessage function

miekg:
Add ToMessage function.

### DIFF
--- a/plugin/proxy/proxy.go
+++ b/plugin/proxy/proxy.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/dnstap"
 	"github.com/coredns/coredns/plugin/pkg/healthcheck"
 	"github.com/coredns/coredns/request"
 
@@ -97,7 +98,7 @@ func (p Proxy) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 				child.Finish()
 			}
 
-			taperr := toDnstap(ctx, host.Name, upstream.Exchanger(), state, reply, start)
+			taperr := dnstap.ToMessage(ctx, host.Name, upstream.Exchanger().Protocol(), state, reply, start)
 
 			if backendErr == nil {
 				w.WriteMsg(reply)
--- a/plugin/dnstap/dnstap.go
+++ b/plugin/dnstap/dnstap.go
@@ -1,9 +1,8 @@
-package proxy
+package dnstap
 
 import (
 	"time"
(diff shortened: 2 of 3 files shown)


miekg:
as said in slack a function inside plugin/dnstap might be better, but is more work

plugin/dnstap/const.go
```diff
@@ -0,0 +1,16 @@
...
+// Mirror some of the tap. constants so we can reference them from this plugin
+// and not the vendored directory meaning out-of-tree middleware can use this.
+
+import (
+	tap "github.com/dnstap/golang-dnstap"
+)
+
+// Mirror constants from the golang-dnstap package.
+const (
+	SocketProtocolTCP        = tap.SocketProtocol_TCP
```
2buisson: Would be best placed in dnstap/msg because those are values for field Type of payload type tap.Message, only a detail though.

miekg: all of them? sgtm

miekg:
note sure, if we should merge this.

johnbelamaric:
/integration

let's see what happens...

miekg:
@johnbelamaric wow! :)

johnbelamaric:
It's all @chrisohaver not me :)

miekg:
@varyoo Ok, I've done the better thing. Not sure about the ToDnstap name though - but too tired now to think of something better.

miekg:
@chrisohaver hats off! This is some nice work!

miekg:
@varyoo is this OK to merge?

plugin/proxy/proxy.go
```diff
@@ -97,7 +98,7 @@ func (p Proxy) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 				child.Finish()
 			}
 
-			taperr := toDnstap(ctx, host.Name, upstream.Exchanger(), state, reply, start)
+			taperr := dnstap.ToMessage(ctx, host.Name, upstream.Exchanger().Protocol(), state, reply, start)
```
2buisson, line 101: Isn't `upstream.Exchanger().Protocol()` returning "dns" or maybe "https_google"?

proxy/dns.go:
~~~ Go
func (d *dnsEx) Protocol() string { return "dns" }
~~~

This is why I introduced Exchanger.Transport() returning "tcp", "udp" or "".
"" if the request used `state.Proto()`

miekg: Fair enough, but let me rephrase: how do we want (maybe external) proxies to report dnstap? I.e. forward does not have an Exchanger?

PR #1178, opened 2017-10-27
doc: update README

miekg:
People are using it in prod and some other updates.


miekg:
ref https://github.com/coredns/deployment/issues/13

PR #1180, opened 2017-10-28
add goroutine to check hosts file for updates

pmoroney:
Creates a goroutine in `OnStartup()` to check the hosts file for changes.
Refactors how the corefile inline hosts file gets fed to the parser.

Basically I took the three maps that are used for lookups and created a struct, `hostsMap`, to hold them.
Then I made `parse()` return that struct and accept an override as a parameter.
So `parse()` is used to create a `hostsMap` from a reader.
Initially, if an inline hostsfile exists it is parsed and the resulting `hostsMap` is set to `Hostsfile.inline`
A goroutine is created that runs `ReadFile()` every 5 seconds. (I kept that duration from the previous implementation, but I'm not married to it)
`ReadFile()` checks the file status to see if it has changed, if it has, it calls `parse()` on the file with `h.inline` as the override and sets the resulting `hostsMap` to `Hostsfile.hmap` which is what is used by the lookup functions.

N/A

N.A

### DIFF
--- a/plugin/hosts/hostsfile.go
+++ b/plugin/hosts/hostsfile.go
@@ -19,8 +19,6 @@ import (
 	"github.com/coredns/coredns/plugin"
 )
 
-const cacheMaxAge = 5 * time.Second
-
 func parseLiteralIP(addr string) net.IP {
 	if i := strings.Index(addr, "%"); i >= 0 {
 		// discard ipv6 zone
@@ -34,13 +32,7 @@ func absDomainName(b string) string {
 	return plugin.Name(b).Normalize()
 }
 
-// Hostsfile contains known host entries.
-type Hostsfile struct {
-	sync.RWMutex
-
-	// list of zones we are authoritive for
-	Origins []string
-
+type hostsMap struct {
 	// Key for the list of literal IP addresses must be a host
 	// name. It would be part of DNS labels, a FQDN or an absolute
 	// FQDN.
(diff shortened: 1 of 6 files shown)


plugin/hosts/hostsfile.go
```diff
@@ -52,71 +44,79 @@ type Hostsfile struct {
...
 // ReadHosts determines if the cached data needs to be updated based on the size and modification time of the hostsfile.
 func (h *Hostsfile) ReadHosts() {
-	now := time.Now()
-
-	// Not deferring h.RUnlock() because we may need to remove the read lock and aquire a write lock
-	h.RLock()
-	if now.Before(h.expire) && len(h.byAddr) > 0 {
-		h.RUnlock()
+	file, err := os.Open(h.path)
+	if err != nil {
```
miekg: This might benefit from a short comment that we don't care if the file does not exist (hence we're not returning `err`)

plugin/hosts/hostsfile.go
```diff
@@ -52,71 +44,79 @@ type Hostsfile struct {
...
 	}
-	readers = append(readers, strings.NewReader(strings.Join(h.inline, "\n")))
-	scanner := bufio.NewScanner(io.MultiReader(readers...))
+
+	hmap := newHostsMap()
+	h.inline = h.parse(strings.NewReader(strings.Join(inline, "\n")), hmap)
+	*h.hmap = *h.inline
+}
+
+func (h *Hostsfile) parseFile(file io.Reader) {
```
miekg: `parseFile` -> `parseReader`? And just use `r` for the io.Reader instead of `file`?

plugin/hosts/hostsfile.go
```diff
@@ -52,71 +44,79 @@ type Hostsfile struct {
...
+	h.hmap = h.parse(file, h.inline)
+}
+
+// Parse reads the hostsfile and populates the byName and byAddr maps.
+func (h *Hostsfile) parse(file io.Reader, override *hostsMap) *hostsMap {
```
miekg: `file` -> `r` as a var

plugin/hosts/hostsfile.go
```diff
@@ -140,18 +140,29 @@ func (h *Hostsfile) Parse(file io.Reader) {
...
 			}
-			is[addr.String()] = append(is[addr.String()], name)
+			hmap.byAddr[addr.String()] = append(hmap.byAddr[addr.String()], name)
 		}
 	}
-	h.byNameV4 = hsv4
-	h.byNameV6 = hsv6
-	h.byAddr = is
+
+	if override != nil {
```
miekg: I would just return here if `override == nil` and then do the override thing below

plugin/hosts/setup.go
```diff
@@ -25,6 +26,33 @@ func setup(c *caddy.Controller) error {
 		return plugin.Error("hosts", err)
 	}
 
+	parseChan := make(chan bool)
+
+	c.OnStartup(func() error {
+		h.ReadHosts()
+		if err != nil {
```
miekg: which `err` is referenced here?

plugin/hosts/setup.go
```diff
@@ -25,6 +26,33 @@ func setup(c *caddy.Controller) error {
...
+			return err
+		}
+
+		go func() {
+			ticker := time.NewTicker(5 * time.Second)
```
miekg: This deserves a small sentence in the README.md (e.g. "We scan for changes every 5s." or something)

miekg (changes requested):
Looks good! Some nits and comments.

pmoroney:
I have made the requested changes. Thank's for catching those.

miekg (approved):

PR #1184, opened 2017-10-30
readme: more tests

miekg:
Add dnssec and file plugin to the test readme. This requires creating a
bunch of files with the right content. Doing so already unconvered an
unconditional type assertion in DNSSEC. This PR will include the fix for
that as well.

Also extended the snippets in the file plugin README, so that they are
whole Corefile - showing more value and checking all corefile snippets.

Create outliner right now is the kubernetes plugin, because even setting
the right env vars will result in:

open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory":

Which we can't create for a test.

### DIFF
--- a/plugin/dnssec/README.md
+++ b/plugin/dnssec/README.md
@@ -47,10 +47,10 @@ If monitoring is enabled (via the *prometheus* directive) then the following met
 
 Sign responses for `example.org` with the key "Kexample.org.+013+45330.key".
 
-~~~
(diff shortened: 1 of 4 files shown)


miekg:
It is somewhat ugly to have to create these files, we could potentially mock the fs access, but that is a big change for no apparent win, i.e. unless we have plans to use/support in-memory FS

PR #1190, opened 2017-11-02
Add option to use pod name rather than IP address for Kubernetes

bakins:
This PR is an idea in code form. 

This adds an option to use the name of a pod as the hostname rather the hostname for lookups when pods mode is set to insecure or verified for the Kubernetes plugin.

https://github.com/kubernetes/kubernetes/issues/47992 - I did not find a related coredns issue doing a quick issue search.

We are also using airflow and this change enables DNS for pod FQDNs without setting`hostname` in the pod specs. As we use deployments, we do not want each pod to have the same hostname.

We would need to document the additional, optional, argument to `pods` in the Kubernetes plugin.

I was unsure if this should be a separate option or part of the `pods` mode.

As I said, this is to capture an idea in code form. It is fairly simple and has tests that pass.

### DIFF
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -31,6 +31,7 @@ kubernetes [ZONES...] {
     namespaces NAMESPACE...
     labels EXPRESSION
     pods POD-MODE
+    endpoint_pod_names
     upstream ADDRESS...
     ttl TTL
     fallthrough
@@ -65,6 +66,16 @@ kubernetes [ZONES...] {
      option requires substantially more memory than in insecure mode, since it will maintain a watch
      on all pods.
 
+* `endpoint_pod_names` Use the pod name of the pod targeted by the endpoint as 
+   the endpoint name in A records, e.g.
+   `endpoint-name.my-service.namespace.svc.cluster.local. in A 1.2.3.4`
+   By default, the endpoint-name name selection is as follows: Use the hostname 
+   of the endpoint, or if hostname is not set, use the dashed form of the endpoint
+   ip address (e.g. `1-2-3-4.my-service.namespace.svc.cluster.local.`)
+   If this directive is included, then name selection for endpoints changes as
+   follows: Use the hostname of the endpoint, or if hostname is not set, use the
+   pod name of the pod targeted by the endpoint. If there is no pod targeted by 
+   the endpoint, use the dashed ip address form.
 * `upstream` **ADDRESS [ADDRESS...]** defines the upstream resolvers used for resolving services
   that point to external hosts (External Services).  **ADDRESS** can be an ip, an ip:port, or a path
   to a file structured like resolv.conf.
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -28,19 +28,20 @@ import (
 
 // Kubernetes implements a plugin that connects to a Kubernetes cluster.
 type Kubernetes struct {
-	Next          plugin.Handler
-	Zones         []string
-	Proxy         proxy.Proxy // Proxy for looking up names during the resolution process
-	APIServerList []string
-	APIProxy      *apiProxy
-	APICertAuth   string
-	APIClientCert string
-	APIClientKey  string
-	APIConn       dnsController
-	Namespaces    map[string]bool
-	podMode       string
-	Fallthrough   bool
-	ttl           uint32
+	Next             plugin.Handler
+	Zones            []string
+	Proxy            proxy.Proxy // Proxy for looking up names during the resolution process
+	APIServerList    []string
+	APIProxy         *apiProxy
+	APICertAuth      string
+	APIClientCert    string
+	APIClientKey     string
+	APIConn          dnsController
+	Namespaces       map[string]bool
+	podMode          string
+	endpointNameMode bool
+	Fallthrough      bool
+	ttl              uint32
 
 	primaryZoneIndex   int
 	interfaceAddrsFunc func() net.IP
@@ -276,9 +277,12 @@ func (k *Kubernetes) Records(state request.Request, exact bool) ([]msg.Service, 
 	return services, err
 }
 
-func endpointHostname(addr api.EndpointAddress) string {
+func endpointHostname(addr api.EndpointAddress, endpointNameMode bool) string {
 	if addr.Hostname != "" {
 		return strings.ToLower(addr.Hostname)
+	}
+	if endpointNameMode && addr.TargetRef != nil && addr.TargetRef.Name != "" {
+		return addr.TargetRef.Name
 	}
 	if strings.Contains(addr.IP, ".") {
 		return strings.Replace(addr.IP, ".", "-", -1)
@@ -375,7 +379,7 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 						// See comments in parse.go parseRequest about the endpoint handling.
 
 						if r.endpoint != "" {
-							if !match(r.endpoint, endpointHostname(addr)) {
+							if !match(r.endpoint, endpointHostname(addr, k.endpointNameMode)) {
 								continue
 							}
 						}
@@ -385,7 +389,7 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 								continue
 							}
 							s := msg.Service{Host: addr.IP, Port: int(p.Port), TTL: k.ttl}
-							s.Key = strings.Join([]string{zonePath, Svc, svc.Namespace, svc.Name, endpointHostname(addr)}, "/")
+							s.Key = strings.Join([]string{zonePath, Svc, svc.Namespace, svc.Name, endpointHostname(addr, k.endpointNameMode)}, "/")
 
 							err = nil
 
--- a/plugin/kubernetes/setup.go
+++ b/plugin/kubernetes/setup.go
@@ -104,6 +104,13 @@ func kubernetesParse(c *caddy.Controller) (*Kubernetes, dnsControlOpts, error) {
 
 		for c.NextBlock() {
 			switch c.Val() {
+			case "endpoint_pod_names":
+				args := c.RemainingArgs()
+				if len(args) > 0 {
+					return nil, opts, c.ArgErr()
+				}
+				k8s.endpointNameMode = true
+				continue
 			case "pods":
 				args := c.RemainingArgs()
 				if len(args) == 1 {
--- a/plugin/kubernetes/setup_test.go
+++ b/plugin/kubernetes/setup_test.go
@@ -471,3 +471,66 @@ func TestKubernetesParse(t *testing.T) {
 		}
 	}
 }
+
+func TestKubernetesEndpointsParse(t *testing.T) {
+	tests := []struct {
+		input                string // Corefile data as string
+		shouldErr            bool   // true if test case is exected to produce an error.
+		expectedErrContent   string // substring from the expected error. Empty for positive cases.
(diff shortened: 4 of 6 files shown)


miekg:
Reading through that whole thread, this seems a reasonable request.

chrisohaver:
This seems like a reasonable option.  This follows the endpoint naming spec, since the pod name is unique in the namespace.

To clarify, this is  "Adding option to use pod name rather than IP address for _k8s endpoint fqdns_" ...
It's changing the way we name endpoints.  The default behavior is to use the pod hostname, the optional behavior being added here is to use the pod name. 

For example: `my-pod-12341234-abc123.my-service.namespace.svc.cluster.local` instead of the default `1-2-3-4.my-service.namespace.svc.cluster.local`

This does not change the way pod lookups function.
For example: `1-2-3-4.namespace.pod.cluster.local`
This does not allow you to look up a pod using for example an query such as `my-pod-12341234-abc123.namespace.pod.cluster.local`

As far as option name goes: I think "endpointNameMode" might be a slightly better/clearer option name than "podNameMode".

plugin/kubernetes/kubernetes.go
```diff
@@ -276,10 +277,13 @@ func (k *Kubernetes) Records(state request.Request, exact bool) ([]msg.Service,
 	return services, err
 }
 
-func endpointHostname(addr api.EndpointAddress) string {
+func endpointHostname(addr api.EndpointAddress, podNameMode string) string {
 	if addr.Hostname != "" {
 		return strings.ToLower(addr.Hostname)
 	}
+	if podNameMode == "name" && addr.TargetRef != nil && addr.TargetRef.Name != "" {
```
miekg: Can you make "name" a constant, similar to add podMode ones

plugin/kubernetes/setup.go
```diff
@@ -106,13 +106,24 @@ func kubernetesParse(c *caddy.Controller) (*Kubernetes, dnsControlOpts, error) {
 			switch c.Val() {
 			case "pods":
 				args := c.RemainingArgs()
-				if len(args) == 1 {
+				k8s.podNameMode = "default"
```
miekg: default is just not setting this? I.e. no need for "default"?

miekg (changes requested):
Few nit comments and the Travis failures need a fx.
And this needs to be documented in the README/

johnbelamaric:
This changes the endpoint records; ie, with svc not pod type in the name. The pods config option is for the pod records. Maybe we need an endpoints config option instead.

bakins:
@chrisohaver this does allow lookups to work for queries such as `pod-phdc8.my-service.svc.cluster.local` if the pod is n th

bakins:
Changes made. I added an `endpoints` config option, as I agree that reflects its actual usage better.
I added documentation - I'd appreciate feedback on it.

I can squash the commits if/when ready to merge.

plugin/kubernetes/README.md
```diff
@@ -65,6 +66,16 @@ kubernetes [ZONES...] {
...
 
+* `endpoints` **ENDPOINT-MODE** sets the mode for handling service endpoint A
+   records, e.g.
+   `pod-name.my-service.svc.cluster.local. in A 1.2.3.4` if the pod is an
+   endpoint of the service `my-service`.  By default, these will not resolve.
+   Valid values for **ENDPOINT-MODE**:
+
+   * `name`: Return an A record if there exists a pod that is a valid endpoint
+    for the service.  This only works if `pods` is set to either `insecure` or
+    `verified`.  If `hostname` is set on a pod, then it always has precedence.
```
johnbelamaric: This should work regardless of the pods setting, I believe. We always watch endpoints, and we respond to their hostname records regardless of whether `pod` types are enabled.

We could add an option to `pods verified` that enables `pod-name.namespace.pod.cluster.local`.

plugin/kubernetes/README.md
```diff
@@ -65,6 +66,16 @@ kubernetes [ZONES...] {
      option requires substantially more memory than in insecure mode, since it will maintain a watch
      on all pods.
 
+* `endpoints` **ENDPOINT-MODE** sets the mode for handling service endpoint A
+   records, e.g.
+   `pod-name.my-service.svc.cluster.local. in A 1.2.3.4` if the pod is an
```
johnbelamaric: Normally if the endpoint has a hostname then these will resolve. This change just makes them use the pod name if it exists and the hostname doesn't. I think that needs to be made clearer here.

johnbelamaric (changes requested):

plugin/kubernetes/README.md
johnbelamaric: But that would be a separate PR...

plugin/kubernetes/kubernetes.go
```diff
@@ -276,10 +279,13 @@ func (k *Kubernetes) Records(state request.Request, exact bool) ([]msg.Service,
 	return services, err
 }
 
-func endpointHostname(addr api.EndpointAddress) string {
+func endpointHostname(addr api.EndpointAddress, endpointNameMode string) string {
 	if addr.Hostname != "" {
 		return strings.ToLower(addr.Hostname)
 	}
+	if endpointNameMode == endpointNameModeName && addr.TargetRef != nil && addr.TargetRef.Name != "" {
```
miekg: Actually now that I think some more; endpointNameMode is just a boolean, no need to carry a string with us all time and to a string compare - this is either on or off.

plugin/kubernetes/setup_test.go
```diff
@@ -32,6 +33,7 @@ func TestKubernetesParse(t *testing.T) {
 			defaultResyncPeriod,
 			"",
 			podModeDisabled,
+			"",
```
miekg: Please create a new parse test for just this option; this adds a whole bunch of empty strings for no purpose at all in this test.

plugin/kubernetes/setup.go
```diff
@@ -104,6 +104,17 @@ func kubernetesParse(c *caddy.Controller) (*Kubernetes, dnsControlOpts, error) {
 
 		for c.NextBlock() {
 			switch c.Val() {
+			case "endpoints":
+				args := c.RemainingArgs()
```
miekg: I would reverse this and not use a switch:
~~~
if len(args) != 1 { 
     return nil, opts. c.ArgErr()
}
if args[0] != endpointNameModeName {
    return nil, opts. fmt.Errorf("wrong value for endpoints: %s")
}

miekg (changes requested):

plugin/kubernetes/README.md
bakins: "If `hostname` is set on a pod, then it always has precedence." - is that enough?

plugin/kubernetes/README.md
bakins: @johnbelamaric you are correct, it works if `pods` is disabled. Updating doc.

bakins:
Changes made. Writing docs is the hardest part of any project for me, so let me know what you think.

plugin/kubernetes/README.md
```diff
@@ -65,6 +66,18 @@ kubernetes [ZONES...] {
      option requires substantially more memory than in insecure mode, since it will maintain a watch
      on all pods.
 
+* `endpoints` **ENDPOINT-MODE** sets the mode for handling service endpoint A
```
chrisohaver: How about this instead...

* `endpoints` **ENDPOINT-MODE** sets the mode for naming endpoint A
   records, e.g. `endpoint-name.my-service.namespace.svc.cluster.local. in A 1.2.3.4`
   By default, the `endpoint-name` is set to the `hostname` of the endpoint, or if `hostname` is not set, the dashed form of the endpoint ip address (e.g. `1-2-3-4.my-service.namespace.svc.cluster.local.`) 

   Valid values for **ENDPOINT-MODE**:

   * `name`: If the endpoint hostname is not set, use the pod name of the pod targeted by the endpoint.  If there is no pod targeted by the endpoint, use the dashed ip address form.

chrisohaver (changes requested):
Looks good.  I've proposed an alternate documentation entry.

miekg (approved):
Looks good from code standpoint.

plugin/kubernetes/README.md
johnbelamaric: ok, but as @miekg said, since there is only one value for endpoints-mode, maybe we should just say that the whole directive is `endpoints pod names` or similar. That will simplify the config and docs a bit. I don't want to just say `endpoints` because it's not clear enough how it changes the behavior.

Questions:
 * Will it still respond to the dashed-ip name even if this is set? That is, is this an additional name or a replacement name? Should we pick one as the A record and have a CNAME for the other?
 * Which name shows up in the SRV record for the service?

chrisohaver: Per my understanding of the code...

> Is this an additional name or a replacement name? 

This is a replacement name, not an additional name.

> Which name shows up in the SRV record for the service?

`SRV` targets will point to the `A` records of the endpoints.

chrisohaver: Agreed ... here is a "boolean" directive version...

* `endpoint_pod_names` Use the pod name of the pod targeted by the endpoint as the endpoint name in `A`  records, e.g. `endpoint-name.my-service.namespace.svc.cluster.local. in A 1.2.3.4`
   By default, the `endpoint-name` name selection is as follows: Use the `hostname` of the endpoint, or if `hostname` is not set, use the dashed form of the endpoint ip address (e.g. `1-2-3-4.my-service.namespace.svc.cluster.local.`) 
   If this directive is included, then name selection for endpoints changes as follows: Use the `hostname` of the endpoint, or if `hostname` is not set, _use the pod name of the pod targeted by the endpoint_.  If there is no pod targeted by the endpoint, use the dashed ip address form.

johnbelamaric: 👍 now @bakins please update the code to match :)

miekg:
think we fleshed out what needs to be done here. Shall we wait with 0.9.10 so this can go in?

johnbelamaric:
No, let's just get 0.9.10 out ASAP.

bakins:
Changes made. Let me know if this all looks good and I'll squash.

plugin/kubernetes/kubernetes.go
```diff
@@ -70,6 +71,8 @@ const (
 	podModeInsecure = "insecure"
 	// DNSSchemaVersion is the schema version: https://github.com/kubernetes/dns/blob/master/docs/specification.md
 	DNSSchemaVersion = "1.0.1"
+	// endpointNameModeName is where we us the pod name rather than the IP for the endpoint hostname.
+	endpointNameModeName = "name"
```
chrisohaver: We can remove `endpointNameModeName` now...

chrisohaver: We can remove `endpointNameModeName` now...

chrisohaver (changes requested):

chrisohaver (approved):

bakins:
Okay, pushed hopefully the last change - removed endpointNameModeName. Let me know and I will squash, if that's what you want.

miekg:
We squash on merge, but feel free to do it yourself and force push the
branch.

bakins:
squashed and pushed.

miekg:
@johnbelamaric are you happy w/ this?

tomplus:
When are you going to release next version with this feature ? I'm looking forward to use it. Thanks.

johnbelamaric:
Yes, 1.0.0 will be released this week or early next week.

tomplus:
Thank you, it works like a charm !

Issue #1177, opened 2017-10-27
Docker image should be a manifest


johnbelamaric:
In order to fully implement https://github.com/kubernetes/community/pull/1100 we need our Docker image to be a manifest list with various platforms built. From @luxas:

> k8s requires the image to be built for amd64,  arm, arm64, ppc64le and s390x

We may want to consider moving the Docker packaging step from the main repo to the https://github.com/coredns/deployment repo at the same time.

Examples for this are here: https://github.com/luxas/kubeadm-workshop

luxas:
Thanks! LMK if you have any questions

miekg:
If we are going to curl these from the official coredns releases we might want to build them there as well, means adding Linux/{arm64,ppc64le,s390x}

But really s390??

luxas:
> But really s390??

@miekg Yeah, people do care about that for k8s ;)

yongtang:
I have some free cycles. I can work on this issue.

johnbelamaric:
@rajansandeep is about to push this already

yongtang:
Ok. Reassign to @rajansandeep.

yongtang:
Let's re-open this issue.

miekg:
Closing because:
https://github.com/coredns/coredns/pull/1195#issuecomment-341722649

PR #1194, opened 2017-11-02
Use `amd64` instread of `x86_64` when building architecture

yongtang:

In Makefile.release `shell uname -m` was used to find the current architecture. On amd64 systems this might be reported as `x86_64` depending on the Linux.

However, in order to have manifest docker images we need to stick with golang's architecture tag of `amd64`.

This fix changes instead so that the correct architecture conforming to golang (`amd64`) could be build.

This PR is related to #1177.

N/A

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -35,12 +35,11 @@
 # Github.
 
 ifeq (, $(shell which gh-release))
-     $(error "No gh-release in $$PATH, install with `go get progrium/gh-release`")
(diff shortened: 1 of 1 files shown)


miekg:
Shouldn't we just hardcore this instead of taking the machines architecture?

yongtang:
@miekg Yes you are right. We will be building multi-arch anyway so there is no need to detect `ARCH`.

The PR has been updated.

PR #1195, opened 2017-11-02
Add docker image to be a manifest list

rajansandeep:
Includes Docker image to be a manifest list.

#1177

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -29,6 +29,12 @@
 # Steps for docker:
 #
 # * Login into docker: docker login (should have push creds for coredns registry)
+# * We use the manifest-tool from https://github.com/estesp/manifest-tool to build the manifest list
+# * Make sure you have the binary downloaded in the directory.
+# * To download manifest-tool (for Linux-amd64):
+# * curl -sSL https://github.com/estesp/manifest-tool/releases/download/v0.7.0/manifest-tool-linux-amd64 > manifest-tool
+# * chmod +x manifest-tool
+#
 # * Run: make -f Makefile.release docker
 #
 # Docker push should happen after you make the new release and uploaded it to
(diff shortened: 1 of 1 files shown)


Makefile.release
```diff
@@ -38,6 +44,10 @@ ifeq (, $(shell which gh-release))
      $(error "No gh-release in $$PATH, install with `go get github.com/progrium/gh-release`")
 endif
 
+ifeq (, $(shell ls | grep manifest-tool ))
```
miekg: Don't assume this is locally installed. Might be in $PATH.

miekg (changes requested):

miekg (approved):

yongtang:
@rajansandeep The base image has to be from different architecture and the binary should used the generated binary form different directories. It should not use the same binary.

yongtang:
The binary seems to be fine as `tar -xzf ` is used. Need to change the base image.

miekg:
@yongtang what remains to be done here?
(If we want to do a quick 0.9.10, this the last outstanding issue)

rajansandeep:
Per the docs: https://hub.docker.com/r/library/alpine/ 

Supported Architectures:
amd64, arm32v6, arm64v8, i386, ppc64le, s390x
As of 2017-09-12, these other architectures are included under the non-prefixed images via "manifest lists" .

So we should be fine as is.

yongtang:
@rajansandeep @miekg Let's hold on for that. The base image is multi arch. However, at the time the coredns docker image is built, the `docker build` is running on a amd64 machine (the dev machine) and the binary will only be added to amd64 docker image.

There are some additional steps to be done. See steps in:
https://github.com/luxas/kubeadm-workshop/blob/master/images/nodejs/Makefile

I am preparing the PR and will open the PR shortly.

miekg:
ok

PR #1207, opened 2017-11-08
plugin/k8s: fix pods disabled behavior

chrisohaver:

Fix return code for pods queries when pods are disabled - per docs, should be `NXDOMAIN`, not `SERVFAIL`...  Unit tests also expect `NXDOMAIN` (but the return code not actually checked in the tests due the the "expected" error returned)

### DIFF
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -143,7 +143,7 @@ func (k *Kubernetes) Lookup(state request.Request, name string, typ uint16) (*dn
 
 // IsNameError implements the ServiceBackend interface.
 func (k *Kubernetes) IsNameError(err error) bool {
-	return err == errNoItems || err == errNsNotExposed || err == errInvalidRequest
+	return err == errNoItems || err == errNsNotExposed || err == errInvalidRequest || err == errPodsDisabled
 }
 
 func (k *Kubernetes) getClientConfig() (*rest.Config, error) {
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -77,7 +77,6 @@ var (
 	errNoItems        = errors.New("no items found")
 	errNsNotExposed   = errors.New("namespace is not exposed")
 	errInvalidRequest = errors.New("invalid query name")
-	errPodsDisabled   = errors.New("pod records disabled")
 )
 
 // Services implements the ServiceBackend interface.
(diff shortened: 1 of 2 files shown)


chrisohaver:
I didn't get the integration tests right... missing expected authority section...

chrisohaver:
Added expected auth section in coredns/ci#4 ... retest integration when merged.

plugin/kubernetes/kubernetes.go
```diff
@@ -143,7 +143,7 @@ func (k *Kubernetes) Lookup(state request.Request, name string, typ uint16) (*dn
 
 // IsNameError implements the ServiceBackend interface.
 func (k *Kubernetes) IsNameError(err error) bool {
-	return err == errNoItems || err == errNsNotExposed || err == errInvalidRequest
+	return err == errNoItems || err == errNsNotExposed || err == errInvalidRequest || err == errPodsDisabled
```
miekg: This smells like a hack. errPodsDisabled should be used in a general IsNameError function.

chrisohaver: not following... (i.e. why do you think its a hack?)

miekg: What has podDisabled to do with nxdomain? errInvalidRequest is already border line

chrisohaver: If the podMode is set to podsDisabled (default setting in k8s plugin), then we don't produce `A` records for pods (e.g. 1-2-3-4.ns.pod.cluster.local)...

If some one tries to query a pod in this mode, then there will be no `A` record for it. `NXDOMAIN` seems like the correct answer here.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/k8s: f..." ]
>
>
>
>> @@ -143,7 +143,7 @@ func (k *Kubernetes) Lookup(state request.Request, name string, typ uint16) (*dn
>
> // IsNameError implements the ServiceBackend interface.
> func (k *Kubernetes) IsNameError(err error) bool {
>-	return err == errNoItems || err == errNsNotExposed || err == errInvalidRequest
>+	return err == errNoItems || err == errNsNotExposed || err == errInvalidRequest || err == errPodsDisabled
>
>If the podMode is set to podsDisabled (default setting in k8s plugin), there we don't produce A records for pods (e.g. 1-2-3-4.ns.pod.cluster.local)...
>
>If some one tries to query a pod in this mode, then there will be no `A` record for it. `NXDOMAIN` seems like the correct answer here.

There will be no records, so return errNoItems?

chrisohaver:
> return errNoItems?

OK - that works too...

PR #1209, opened 2017-11-08
Add source graph used by badge

yongtang:

We are used by 10+ other projects now 👍 . Think it makes sense to add this badge.

N/A

N/A


johnbelamaric:
Seems like this is more geared toward libraries. Would make more sense in miekg/dns rather than here.

miekg:
It should be possible to use coredns as a lib as well. But I see your point.

johnbelamaric:
10 is a low number of "users". So I don't think it reflects that well because I think it can be misunderstood. These 10 are probably the various external plugins, which is interesting, but doesn't reflect adoption or usage.

miekg:
Ack. Feel free to revert.

yongtang:
It is not necessary a fair comparison to libraries though it is still a good indicator. For example, caddy is used by 121 projects by the same metric. That is roughly proportional to the number of stars/forks compared with CoreDNS.

Not a big deal anyway so feel free to revert.

PR #1212, opened 2017-11-09
plugin/k8s: Use service IP index in reverse lookups

chrisohaver:

Uses the service IP index (instead of service list) for reverse lookups in kubernetes.

### DIFF
--- a/plugin/kubernetes/reverse_test.go
+++ b/plugin/kubernetes/reverse_test.go
@@ -14,16 +14,19 @@ import (
 
 type APIConnReverseTest struct{}
 
-func (APIConnReverseTest) HasSynced() bool                       { return true }
-func (APIConnReverseTest) Run()                                  { return }
-func (APIConnReverseTest) Stop() error                           { return nil }
-func (APIConnReverseTest) PodIndex(string) []*api.Pod            { return nil }
-func (APIConnReverseTest) SvcIndex(string) []*api.Service        { return nil }
-func (APIConnReverseTest) SvcIndexReverse(string) []*api.Service { return nil }
-func (APIConnReverseTest) EpIndex(string) []*api.Endpoints       { return nil }
-func (APIConnReverseTest) EndpointsList() []*api.Endpoints       { return nil }
+func (APIConnReverseTest) HasSynced() bool                 { return true }
+func (APIConnReverseTest) Run()                            { return }
+func (APIConnReverseTest) Stop() error                     { return nil }
+func (APIConnReverseTest) PodIndex(string) []*api.Pod      { return nil }
+func (APIConnReverseTest) SvcIndex(string) []*api.Service  { return nil }
+func (APIConnReverseTest) EpIndex(string) []*api.Endpoints { return nil }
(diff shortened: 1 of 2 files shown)


plugin/kubernetes/reverse_test.go
```diff
@@ -14,16 +14,19 @@ import (
...
+func (APIConnReverseTest) Stop() error                     { return nil }
+func (APIConnReverseTest) PodIndex(string) []*api.Pod      { return nil }
+func (APIConnReverseTest) SvcIndex(string) []*api.Service  { return nil }
+func (APIConnReverseTest) EpIndex(string) []*api.Endpoints { return nil }
+func (APIConnReverseTest) EndpointsList() []*api.Endpoints { return nil }
+func (APIConnReverseTest) ServiceList() []*api.Service     { return nil }
 
-func (APIConnReverseTest) ServiceList() []*api.Service {
+func (APIConnReverseTest) SvcIndexReverse(ip string) []*api.Service {
+	if ip != "192.168.1.100" {
```
miekg, line 27: this looks weirds

chrisohaver:
Integration tests seem to be passing now.

Perhaps additional load due to concurrent CPU heavy fuzz testing is exposing some timing flakes in the CI tests.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/k8s: U..." ]
>Integration tests seem to be passing now.
>
>Perhaps additional load due to concurrent CPU heavy fuzz testing is exposing some timing flakes in the CI tests.

that's good and bad I guess?

plugin/kubernetes/reverse_test.go
chrisohaver: In the unit test, there is only one service in the index, with IP "192.168.1.100".  To mimic the behavior of the real index function, if called with any IP other than than "192.168.1.100", then it should return nil.

miekg: I don't know, still looks weird. Can we have a separate APIConnReverseTest config for those tests then?

chrisohaver: I think what I have done makes the most sense.  Separating it out would mean more test code bloat.  This is also a better cleaner test, since it returns he correct answer based on the input.  For example, without the "if", the function will always return an answer.  So, you could build a test case that expects a wrong answer, and it would pass, as long as the expected answer is "192.168.1.100".

chrisohaver:
ok to merge?

rajansandeep (approved):

plugin/kubernetes/reverse_test.go
fturib: I would second Chris here : the mock API is dedicated for this set of 4 UT in this class. The overall is small and easy to read.
Having one API mock strictly per use case would overload the class UT. 
I agree we may have to do one day if it test cases start to overlapping .. and logic in mock is too complicated.
But for now, I would think it is better to keep simple.

miekg:
ok. Merging.

PR #1217, opened 2017-11-10
plugin/log: remove OutputFile

### LINKED ISSUE #1216: plugin/log: OutputFile in Rule struct is not used
`OutputFile` in `Rule` struct from log.go is not used and can be removed.

It can cause some confusion that an output file can be specified but the log plugin only logs to stdout anyway.

miekg:
We use stdout for everything, remove OutputFile as it isn't used and
causes confusion.

Fixes #1216

### DIFF
--- a/plugin/log/setup.go
+++ b/plugin/log/setup.go
@@ -29,13 +29,7 @@ func setup(c *caddy.Controller) error {
 	// Open the log files for writing when the server starts
 	c.OnStartup(func() error {
 		for i := 0; i < len(rules); i++ {
-			// We only support stdout
-			writer := os.Stdout
-			if rules[i].OutputFile != "stdout" {
-				return plugin.Error("log", fmt.Errorf("invalid log file: %s", rules[i].OutputFile))
-			}
-
-			rules[i].Log = log.New(writer, "", 0)
+			rules[i].Log = log.New(os.Stdout, "", 0)
 		}
 
 		return nil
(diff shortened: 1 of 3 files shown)


plugin/log/setup.go
```diff
@@ -57,19 +51,20 @@ func logParse(c *caddy.Controller) ([]Rule, error) {
...
-				OutputFile: DefaultLogFilename,
-				Format:     DefaultLogFormat,
+				NameScope: ".",
+				Format:    DefaultLogFormat,
 			})
 		} else if len(args) == 1 {
-			// Only an output file specified.
+			// Only an output file specified, can only be *stdout*
+			if args[0] != "stdout" {
+				return nil, fmt.Errorf("invalid log file: %s, only %s is allowed", args[0], "stdout")
```
nchrisdk: Consider shortening the message:
`fmt.Error("Only stdout is allowed")`

nchrisdk (approved):
Looks good.
Thanks for the quick response!

miekg:
We could remove 'stdout' in is entirety? But that's no backwards compatible, so... meh.

stp-ip:
We haven't released v1.0.0 so we could do a breaking change for v1.0.0.
There won't be a better time for such a breaking change in my view.

miekg:
Good point. @nchrisdk you want to take care of that?
(I'll merge this when tests are done)

PR #1220, opened 2017-11-10
Add docker to make for local dev

chrisohaver:

Adds `make docker` to build locally a dockerized coredns for local dev testing.  This is not associated with the release, just convenience for developer docker/k8s testing on a local system.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ BINARY:=coredns
 SYSTEM:=
 CHECKS:=check godeps
 VERBOSE:=-v
+DOCKER_IMAGE_NAME:=coredns/coredns
 
 all: coredns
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -4,6 +4,7 @@ BINARY:=coredns
 SYSTEM:=
 CHECKS:=check godeps
 VERBOSE:=-v
+DOCKER_IMAGE_NAME:=coredns/coredns
```
miekg, line 7: If its for local dev, you shouldn't name it like this, a simple docker push overwrites it on docker hub.

Also: I thought we had (some) agreement of pushing this into the deployment repo.
IOW: why not use a shell alias/function to do this?

Issue #1218, opened 2017-11-10
plugin/log: remove stdout from the config


miekg:
Currently in the extended config you need to specify 'stdout' as the output file.  This is weird because we only log to stdout. So we should remove it from the config:

I.e this:

~~~ txt
log [NAME] stdout [FORMAT]
~~~

Can be become:

~~~ txt
log [NAME] [FORMAT]
~~~

And

~~~ txt
log [NAME] stdout [FORMAT] {
    class [success|denial|error|all]
}
~~~
should just be

~~~ txt
log [NAME]  [FORMAT] {
    class [success|denial|error|all]
}
~~~

miekg:
@nchrisdk has volunteered to take this on. Thanks!

PR #1221, opened 2017-11-10
plugin/log: remove need to specify stdout

nchrisdk:
Clean up of log plugin to remove need to specify stdout in Corefile

#1218

README.md of log plugin has been changed to match new format

### DIFF
--- a/plugin/log/README.md
+++ b/plugin/log/README.md
@@ -10,14 +10,10 @@ log
 
 * With no arguments, a query log entry is written to *stdout* in the common log format for all requests
 
-~~~ txt
-log [stdout]
-~~~
-
 Or if you want/need slightly more control:
 
 ~~~ txt
-log [NAME] stdout [FORMAT]
+log [NAME] [FORMAT]
 ~~~
 
 * `NAME` is the name to match in order to be logged
@@ -26,7 +22,7 @@ log [NAME] stdout [FORMAT]
 You can further specify the class of responses that get logged:
 
 ~~~ txt
-log [NAME] stdout [FORMAT] {
+log [NAME] [FORMAT] {
     class [success|denial|error|all]
 }
 ~~~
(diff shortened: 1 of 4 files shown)


plugin/log/README.md
```diff
@@ -10,14 +10,10 @@ log
 
 * With no arguments, a query log entry is written to *stdout* in the common log format for all requests
 
-~~~ txt
-log [stdout]
-~~~
-
```
miekg, line 16: I would leave this in, but drop the `[stdout]` bit

plugin/log/setup_test.go
```diff
@@ -18,47 +18,43 @@ func TestLogParse(t *testing.T) {
 			NameScope: ".",
 			Format:    DefaultLogFormat,
 		}}},
-		{`log example.org stdout`, false, []Rule{{
+		{`log example.org.`, false, []Rule{{
```
miekg: leaving out the suffix `.` on `example.org` was actually part of the test.

plugin/log/setup_test.go
```diff
@@ -18,47 +18,43 @@ func TestLogParse(t *testing.T) {
...
-			NameScope: "example.org.",
-			Format:    DefaultLogFormat,
-		}}},
-		{`log example.org stdout {common}`, false, []Rule{{
+		{`log example.org. {common}`, false, []Rule{{
 			NameScope: "example.org.",
 			Format:    CommonLogFormat,
 		}}},
-		{`log example.org stdout {combined}`, false, []Rule{{
+		{`log example.org. {combined}`, false, []Rule{{
```
miekg: here too: don't add the dot

miekg (changes requested):
looks good, still some (small) comments.

Thanks for the quick turn around!

plugin/log/README.md
nchrisdk: Right but then the statement becomes the same as is on line 7-9...

miekg: /me checks out PR. Ah yes, you're richt.

nchrisdk: =)

miekg (approved):

PR #1222, opened 2017-11-10
plugin/k8s: fix endpoint index creation

chrisohaver:

Corrects the k8s endpoint ip index, which is used by reverse lookups in k8s plugin.

Sort of related: we should add (restore?) PTR tests to the CI (coredns/ci#5)

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -144,12 +144,17 @@ func epNameNamespaceIndexFunc(obj interface{}) ([]string, error) {
 	return []string{s.ObjectMeta.Name + "." + s.ObjectMeta.Namespace}, nil
 }
 
-func epIPIndexFunc(obj interface{}) ([]string, error) {
-	ep, ok := obj.(*api.EndpointAddress)
+func epIPIndexFunc(obj interface{}) (idx []string, err error) {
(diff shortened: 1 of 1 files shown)


plugin/kubernetes/controller.go
```diff
@@ -144,12 +144,17 @@ func epNameNamespaceIndexFunc(obj interface{}) ([]string, error) {
...
 
-func epIPIndexFunc(obj interface{}) ([]string, error) {
-	ep, ok := obj.(*api.EndpointAddress)
+func epIPIndexFunc(obj interface{}) (idx []string, err error) {
+	ep, ok := obj.(*api.Endpoints)
 	if !ok {
-		return nil, errors.New("obj was not an *api.EndpointAddress")
+		return nil, errors.New("obj was not an *api.Endpoints")
+	}
+	for _, eps := range ep.Subsets {
```
miekg: Named return params add not much here, it would leave them out. err is not even used.

miekg (changes requested):

rajansandeep (approved):
Looks good.

miekg (approved):
If we are in nit territory I would place the idx closer to where it is used.

Issue #1228, opened 2017-11-13
plugin/kubernetes: NXDOMAIN for default.pod.cluster.local is wrong


miekg:
It should be NODATA as it is a non-terminal in pods verified mode:

~~~
dig +search srv 10-40-4-3.default.pod.cluster.local.
;10-40-4-3.default.pod.cluster.local. IN	SRV

;; ANSWER SECTION:
10-40-4-3.default.pod.cluster.local. 0 IN SRV	0 100 0 10-40-4-3.default.pod.cluster.local.

;; ADDITIONAL SECTION:
10-40-4-3.default.pod.cluster.local. 0 IN A	10.40.4.3
~~~
and

~~~
dig +search srv default.pod.cluster.local.
 ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 59987
~~~

Probably hit this for `default.svc.cluster.local` as well, but there is a default svc running (always?)

miekg:
Gah. Looks like this needs to be special cased in `findPods`

johnbelamaric:
default is the namespace not the service. which I guess is why you are doing #1229

miekg:
uh, yes. And yes :)

miekg:
Fixed via #1229

PR #1229, opened 2017-11-13
plugin/kubernetes: correctly set NODATA for ns

miekg:
A bare or wildcard query for just the namespace should return NODATA,
not NXDOMAIN, otherwise we deny the entirety of the names under the
namespace.

Add test to check for this in pod verified mode.

#1228

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -36,6 +36,7 @@ type dnsController interface {
 	EndpointsList() []*api.Endpoints
 
 	GetNodeByName(string) (*api.Node, error)
+	GetNamespaceByName(string) (*api.Namespace, error)
 
 	Run()
 	HasSynced() bool
@@ -394,4 +395,12 @@ func (dns *dnsControl) GetNodeByName(name string) (*api.Node, error) {
 		return &api.Node{}, err
 	}
 	return v1node, nil
+}
+
+func (dns *dnsControl) GetNamespaceByName(name string) (*api.Namespace, error) {
+	v1ns, err := dns.client.Namespaces().Get(name, meta.GetOptions{})
+	if err != nil {
+		return &api.Namespace{}, err
+	}
+	return v1ns, nil
 }
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -389,6 +389,9 @@ func (dns *dnsControl) EndpointsList() (eps []*api.Endpoints) {
 	return eps
 }
 
+// GetNodeByName return the node by name. If nothing is found an error is
+// returned. This query causes a roundtrip to the k8s API server, so use
+// sparingly. Currently this is only used for Federation.
 func (dns *dnsControl) GetNodeByName(name string) (*api.Node, error) {
 	v1node, err := dns.client.Nodes().Get(name, meta.GetOptions{})
 	if err != nil {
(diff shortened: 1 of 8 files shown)


plugin/kubernetes/controller.go
```diff
@@ -395,3 +396,11 @@ func (dns *dnsControl) GetNodeByName(name string) (*api.Node, error) {
 	}
 	return v1node, nil
 }
+
+func (dns *dnsControl) GetNamespaceByName(name string) (*api.Namespace, error) {
+	v1ns, err := dns.client.Namespaces().Get(name, meta.GetOptions{})
```
chrisohaver: I think this is a round trip to the k8s API (slow).  We should probably maintain a cache of namespaces instead, like we do for services, endpoints, and pods in `pods verified` mode.

plugin/kubernetes/controller.go
```diff
@@ -395,3 +396,11 @@ func (dns *dnsControl) GetNodeByName(name string) (*api.Node, error) {
...
+	if err != nil {
+		return &api.Namespace{}, err
```
chrisohaver: IMO, un-exposed namespaces should return NXDOMAIN... even though they exist, they are unexposed.

chrisohaver (changes requested):

plugin/kubernetes/controller.go
miekg: But this check is not done here? Should it?

chrisohaver: Yes, before returning the namespace, verify that it is exposed.

plugin/kubernetes/controller.go
miekg: It is, as is `GetNodeByName` (which is only used by federation?). The use case here is also small For the time being update the comment and saying this is slow because of the round trip should be good enough? I.e. it is only called on "wildcard" requests.

chrisohaver: Federation in k8s is in limbo. Stuck in beta, it's likely to be scrapped and replaced with something entirely different. So, I would not be concerned for performance there.

Wildcards requests are non-standard, so I'm sure they are practically a non-existent.
I don't know how common bare requests are (e.g. `ns1.svc.cluster.local`), 

So as long as bare namespace requests are uncommon, then it's OK to be slow here.

plugin/kubernetes/controller.go
chrisohaver: If you are leaving namespaces uncached, you can "cheat" here and skip the API lookup if only some namespaces are exposed.  I.e. just verify the namespace is on the exposed list without verifying it is real in the API. This means that if a user exposes a non-existing namespace, we'd present it as an empty namespace... but I think thats OK.

miekg: Actually looking at the code, that is already checked before we go into findPods or findServices so I don't think we should do it here.
But some comments could clarify this.

plugin/kubernetes/controller.go
chrisohaver: Should add namespace exposure cases to unit tests.

plugin/kubernetes/controller.go
chrisohaver: Should add namespace exposure cases to unit tests.

miekg:
Per comments, not exposed namespaces should return NXDOMAIN, that code is not touched here. We do have test for those in the CI repo.

chrisohaver (commented):
lgtm

chrisohaver (approved):

PR #1230, opened 2017-11-13
plugin/kubernetes: check err in getClientConfig

ncdc:

Check the error returned by clientConfig.ClientConfig() before trying to set cc.ContentType. Otherwise, you can get an ugly nil pointer exception.


miekg:
thanks!

PR #1234, opened 2017-11-14
pl/kubernetes: small cleanups

miekg:
Move nodata check up, so that it is done before we look up
pods/services.
Smaller cleanups and micro optimizations.


miekg:
don't have the bandwidth to deal w/ the integration errors.

Issue #1202, opened 2017-11-04
Proxy doesn't recover from network outage during startup


lugoues:
If there is no internet connectivity when coredns is started the initial resolution of dns.google.com it enters a loop of death where it never ends up resolving.

There are several issues I see below
 - `https://8.8.4.4:53/resolve?name=google.com.&type=1` doesn't make sense, 8.8.8.8 is a dns server not a web server
- `Host 8.8.4.4:53 health check probe failed: Get http://8.8.4.4:53:` again 8.8.8.8 is not a web server but also there is no health check configured and per the documentation:
    > If this option is not provided then health checks are disabled
 - This puts the upstream hosts into a `down` state which it never seams to recover from: see the 10:23:35 portion of the log below.
 - I tried setting `max_falis 0` to try and force the hosts back into a healthy state but it doesn't cause a refresh of the upstream server status. Instead it just enters a fast retry loop that never goes anywhere (see log below)

Config
```
. {
    hosts {
      fallthrough
    }
    proxy . 8.8.8.8:53 8.8.4.4:53 {
        protocol https_google
        max_fails 0
    }
    cache
    errors
}
```
Log
```
│2017/11/03 10:23:35 [WARNING] Failed to connect to HTTPS backend "dns.google.com.": Get https://172.217.12.142:443/resolve?name=117.5.40.10.in-addr.arpa.&type=12: dial tcp 172.217.12.142:443: connect: can't assign requested address                                       ┤
│2017/11/03 10:23:35 [WARNING] Failed to connect to HTTPS backend "dns.google.com.": Get https://172.217.12.142:443/resolve?name=117.5.40.10.in-addr.arpa.&type=12: dial tcp 172.217.12.142:443: connect: can't assign requested address                                       ┤
│2017/11/03 10:23:35 [WARNING] Failed to connect to HTTPS backend "dns.google.com.": Get https://172.217.12.142:443/resolve?name=117.5.40.10.in-addr.arpa.&type=12: dial tcp 172.217.12.142:443: connect: can't assign requested address                                       ┤
│03/Nov/2017:10:23:35 -0400 [ERROR 2 117.5.40.10.in-addr.arpa. PTR] unreachable backend: no upstream host                                                                                                                                                                      ┤
│03/Nov/2017:10:23:35 -0400 [ERROR 2 21.client-channel.google.com. A] unreachable backend: no upstream host                                                                                                                                                                    ┤
[...]

PR #1246, opened 2017-11-21
plugin/k8s: fix external service type check

chrisohaver:

Fixes an issue found by kubernetes end to end tests where an existing ExternalName service is converted into a ClusterIP service and results in a ClusterIP service with a (vestigial) externalName field.  This case was confusing coredns.  Fix is to identify external services by checking the service type instead of the external name field.

The failed test is the last sub test of "should provide DNS for ExternalName services" in kubernetes/test/e2e/network/dns.go

kubernetes/test-infra#5601

### DIFF
--- a/plugin/kubernetes/handler_test.go
+++ b/plugin/kubernetes/handler_test.go
@@ -288,6 +288,7 @@ var svcIndex = map[string][]*api.Service{
 				Protocol: "tcp",
 				Port:     80,
 			}},
+			Type: api.ServiceTypeExternalName,
 		},
 	}},
 }
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -418,16 +418,15 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 		}
 
 		// External service
-		if svc.Spec.ExternalName != "" {
+		if svc.Spec.Type == api.ServiceTypeExternalName {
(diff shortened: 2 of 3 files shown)


plugin/kubernetes/handler_test.go
```diff
@@ -288,6 +288,7 @@ var svcIndex = map[string][]*api.Service{
 				Protocol: "tcp",
 				Port:     80,
 			}},
+			Type: api.ServiceTypeExternalName,
```
fturib, line 291: Agree.. view from e2e tests of k8s. The externalNames services are defined by type.

fturib (approved):

johnbelamaric:
So, we should release 1.0.0 soon with this and any other fixes we need for e2e. Then we can bump the version number in the deployment manifests in kubadm, etc.

miekg:
new release sgtm, for 1.0.0 I at least want to take a look at: #1247 and #1202

johnbelamaric:
Ok. We have at least one more person that is interested in contributing, and has some time right now. Maybe he can take one of those two.

fturib:
@johnbelamaric , @miekg : Chris and I talked about the critical of this issue. 
The use case is raised here because the Integration test in K8s is half-invalid.

We want CoreDNS to be fixed because we want to pass the Integration tests and the code will be more correct like that. But if someone is reaching the issue in teh field, that is mostly because the Service is not correctly defined in K8s.

In other word .. I think we should not rush CoreDNS for that particular issue.

miekg:
Thanks! Feel free to pick on those issues. Note that github does not send notification for assigning an issue to someone.

fturib:
@miekg : ?? - comment on wrong issue I guess ...

fturib:
oops ... no. I did not see the comment of JohnB above.

Issue #1251, opened 2017-11-22
Log line and metrics don't match


miekg:
Running a coredns with a zone configured and doing a query yields:
~~~
::1 - [22/Nov/2017:15:48:45 +0000] "SOA IN dns.miek.nl. udp 41 false 4096" NOERROR qr,aa,rd,ra 119 172.933µs
~~~

but metrics say:

~~~
# HELP coredns_dns_request_duration_milliseconds Histogram of the time (in milliseconds) each request took.
# TYPE coredns_dns_request_duration_milliseconds histogram
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="0.005"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="0.01"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="0.025"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="0.05"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="0.1"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="0.25"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="0.5"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="1"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="2.5"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="5"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="10"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="50"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="100"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="200"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="500"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="1000"} 1
coredns_dns_request_duration_milliseconds_bucket{zone="dns.miek.nl.",le="2000"} 1
[...]

PR #1253, opened 2017-11-22
Update timing histograms

SuperQ:
* Use seconds for bucketing.
* Add template for high accuracy timing buckets to avoid copy-pasta.

Updates histogram buckets

Fixes: https://github.com/coredns/coredns/issues/1251

### DIFF
--- a/plugin/metrics/vars/report.go
+++ b/plugin/metrics/vars/report.go
@@ -20,7 +20,7 @@ func Report(req request.Request, zone, rcode string, size int, start time.Time) 
 	typ := req.QType()
 
 	RequestCount.WithLabelValues(zone, net, fam).Inc()
-	RequestDuration.WithLabelValues(zone).Observe(float64(time.Since(start) / time.Millisecond))
+	RequestDuration.WithLabelValues(zone).Observe(time.Since(start).Seconds())
 
 	if req.Do() {
 		RequestDo.WithLabelValues(zone).Inc()
--- a/plugin/metrics/vars/vars.go
+++ b/plugin/metrics/vars/vars.go
@@ -18,9 +18,9 @@ var (
 	RequestDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: plugin.Namespace,
 		Subsystem: subsystem,
-		Name:      "request_duration_milliseconds",
-		Buckets:   append(prometheus.DefBuckets, []float64{50, 100, 200, 500, 1000, 2000, 3000, 4000, 5000, 10000}...),
-		Help:      "Histogram of the time (in milliseconds) each request took.",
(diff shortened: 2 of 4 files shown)


miekg:
problem is this changes the metric name which is not backwards compatible (in a way). So I rather would fix this without changing the metric' name. WDYT?

plugin/metrics/vars/report.go
```diff
@@ -20,7 +20,7 @@ func Report(req request.Request, zone, rcode string, size int, start time.Time)
 	typ := req.QType()
 
 	RequestCount.WithLabelValues(zone, net, fam).Inc()
-	RequestDuration.WithLabelValues(zone).Observe(float64(time.Since(start) / time.Millisecond))
+	RequestDuration.WithLabelValues(zone).Observe(time.Since(start).Seconds())
```
miekg, line 23: ah that is nice. Didn't know about that. (in this context). Prolly convert it back to milliseconds?

SuperQ:
The problem is, the metric name doesn't meet with Prometheus best practices.  We had a lot of cruft with random units and we're trying to get everything onto a set of standard base units.  In this case, it should be seconds.

It's ok to break metric names if it makes sense to do so.  In this case I strongly suggest fixing this sooner rather than later. :smile:

johnbelamaric:
Let's fix (change names) before 1.0!

miekg:
fair enough, but the prometheus clients is exporting it like this: `http_request_duration_microseconds` so I don't why this would be against best practices?

SuperQ:
Yup, those metrics are old and broken, it's been on the TODO list to fix for a long while. :cry:

plugin/metrics/setup.go
```diff
@@ -98,3 +98,8 @@ const (
 	todo = 1
 	done = 2
 )
+
+// Based on Prometheus client_golang prometheus.DefBuckets
+var (
+	DefTimeBuckets = []float64{0.00025, 0.0005, 0.001, 0.0025 .005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}
```
miekg: This doesn't need be exported and can just be `var timeBucket` (also drop the `var ( )`)

miekg (commented):
the one nit

miekg:
travis fails:
~~~
gometalinter --deadline=1m --disable-all --enable=gofmt --enable=golint --enable=vet --exclude=^vendor/ --exclude=^pb/ ./...
plugin/metrics/setup.go:104:1:error: expected declaration, found ')' (vet)
plugin/metrics/setup.go:104:1:warning: expected declaration, found ')' (golint)
~~~

SuperQ:
Ok, sorry for the delay, I've fixed all of the problems with the travis tests.

PR #1191, opened 2017-11-02
WIP: plugin/kubernetes: axfr

miekg:
### 1. What does this pull request do?

Hook up dynamic SOA serial by setting ResourceEventHandlerFuncs in
dnsController.

Add prototype of returns msg.Services via the Transfer function. Leave
expanding this to RRs out of scope for a bit.

### 3. Which documentation changes (if any) need to be made?

This needs to be documented in the README and requires extension of the setup to allow `transfer to` syntax to be added.

PR #1259, opened 2017-11-28
support for zone transfer for kubernetes

bradbeam:
Building on the work in #1191

```
[22:22:56] $ go test -v -run TestKubernetesXFR
=== RUN   TestKubernetesXFR
--- PASS: TestKubernetesXFR (0.00s)
	xfr_test.go:45: &{ResponseWriter:0x2d398f0 Rcode:0 Len:365 Msg:;; opcode: QUERY, status: NOERROR, id: 65471
		;; flags: qr aa ra; QUERY: 1, ANSWER: 9, AUTHORITY: 0, ADDITIONAL: 0
		
		;; QUESTION SECTION:
		;cluster.local.	IN	 AXFR
		
		;; ANSWER SECTION:
		cluster.local.	300	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 0 7200 1800 86400 30
		svc1.testns.svc.cluster.local.	5	IN	A	10.0.0.1
		svc6.testns.svc.cluster.local.	5	IN	AAAA	1234:abcd::1
		172-0-0-2.hdls1.testns.svc.cluster.local.	5	IN	A	172.0.0.2
		172-0-0-3.hdls1.testns.svc.cluster.local.	5	IN	A	172.0.0.3
		5678-abcd--1.hdls1.testns.svc.cluster.local.	5	IN	AAAA	5678:abcd::1
		5678-abcd--2.hdls1.testns.svc.cluster.local.	5	IN	AAAA	5678:abcd::2
		external.testns.svc.cluster.local.	5	IN	CNAME	ext.interwebs.test.
		cluster.local.	300	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 0 7200 1800 86400 30
		 Start:2017-11-27 22:23:15.177925535 -0600 CST m=+0.048178501}
PASS
ok  	github.com/coredns/coredns/plugin/kubernetes	0.060s
```

I think it could still use some polish and support for additional record types.


plugin/kubernetes/controller.go
```diff
@@ -410,3 +415,18 @@ func (dns *dnsControl) GetNamespaceByName(name string) (*api.Namespace, error) {
 	}
 	return v1ns, nil
 }
+
+func (dns *dnsControl) Modified() int64 {
```
miekg: This doesn't actually tell you if it is modified? Should we just the `Serial(state request.Request) uint32` for this? Or change that interface?

plugin/kubernetes/handler.go
```diff
@@ -30,29 +30,31 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 		err     error
 	)
 
-	switch state.Type() {
-	case "A":
+	switch state.QType() {
```
miekg: Type() should just return AXFR  - I hope? Makes the PR smaller because it cuts out of the churn here.

plugin/kubernetes/kubernetes.go
```diff
@@ -461,12 +468,3 @@ func match(a, b string) bool {
 func wildcard(s string) bool {
 	return s == "*" || s == "any"
 }
-
-const (
```
miekg: Please try (I know it is hard) to now do cleanups in the same PR

plugin/kubernetes/handler.go
```diff
@@ -30,29 +30,31 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
...
+	case dns.TypeSOA:
 		records, err = plugin.SOA(&k, zone, state, plugin.Options{})
-	case "NS":
+	case dns.TypeNS:
 		if state.Name() == zone {
 			records, extra, err = plugin.NS(&k, zone, state, plugin.Options{})
 			break
 		}
 		fallthrough
+	case dns.TypeAXFR:
```
miekg: In the *file* plugin there is this:
~~~
	if state.QType() == dns.TypeAXFR || state.QType() == dns.TypeIXFR {
		xfr := Xfr{z}
		return xfr.ServeDNS(ctx, w, r)
	}
~~~
I.e. we jump to another ServeDNS that handles all xfr logic

plugin/kubernetes/setup.go
```diff
@@ -201,6 +201,12 @@ func kubernetesParse(c *caddy.Controller) (*Kubernetes, dnsControlOpts, error) {
 					return nil, opts, c.Errf("ttl must be in range [5, 3600]: %d", t)
 				}
 				k8s.ttl = uint32(t)
+			case "transfer":
+				dests, err := TransferParse(c)
```
miekg: Can you not just call `file.TransferParse(c, false)`

plugin/kubernetes/xfr.go
```diff
@@ -1,17 +1,80 @@
...
+
+	"github.com/miekg/dns"
+	api "k8s.io/client-go/pkg/api/v1"
 )
 
 // Serial implements the Transferer interface.
-func (e *Kubernetes) Serial(state request.Request) uint32 {
-	return uint32(time.Now().Unix())
-}
+func (k *Kubernetes) Serial(state request.Request) uint32 { return uint32(k.APIConn.Modified()) }
```
miekg: ah you do it here. Ok, maybe `Modified()` does make sense; however the name 'modified` is somewhat misleading

miekg (changes requested):
Def. on the right path here. I need to fiddle with this code as well, 'cause I'm not 100% sure what then end result should look like. Would be good to try to re-use as much of the current file plugin code as possible, maybe moving some stuff in a pkg/transfer package. Again: not sure.

plugin/kubernetes/xfr.go
```diff
@@ -1,17 +1,80 @@
...
+
+func (k *Kubernetes) transfer(c chan msg.Service, zone string) {
+
+	defer close(c)
+
+	zonePath := msg.Path(zone, "coredns")
+	serviceList := k.APIConn.ServiceList()
+	for _, svc := range serviceList {
+		// Endpoint query or headless service
+		if svc.Spec.ClusterIP == api.ClusterIPNone {
```
johnbelamaric: You should check the service type not just the defined fields. For example https://github.com/coredns/coredns/pull/1246

plugin/kubernetes/setup.go
bradbeam: We could do that, but I didn't want to create a dependency between the file and Kubernetes plugin.

plugin/kubernetes/kubernetes.go
bradbeam: Should I just defer these for a later PR then?

plugin/kubernetes/handler.go
bradbeam: I thought it made sense to switch to using the constants for comparison versus a string comparison.

plugin/kubernetes/setup.go
bradbeam: Also, the `file.TransferParse` supports transfer from, which we don't want to support in the Kubernetes plugin.

bradbeam: But I guess we could just ignore/err if we get froms returned. So that's an option.

plugin/kubernetes/handler.go
bradbeam: Can you elaborate a bit?
What would the benefit be of implementing another ServeDNS to handle this?

plugin/kubernetes/xfr.go
bradbeam: How do we want to handle when type is empty? 
In reading through api/v1/types.go and api/v1/defaults.go.SetDefaults_Service it seems like this shouldn't ever be the case where svc.Spec.Type is "", but there are some test cases that don't seem to have this defined.

bradbeam: I've updated the tests to explicitly define the type.

bradbeam:
Added in some additional bits to include SRV records
```
		;; QUESTION SECTION:
		;cluster.local.	IN	 AXFR
		
		;; ANSWER SECTION:
		cluster.local.	300	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1511935515 7200 1800 86400 30
		svc1.testns.svc.cluster.local.	5	IN	A	10.0.0.1
		_http._tcp.svc1.testns.svc.cluster.local.	5	IN	SRV	0 100 80 svc1.testns.svc.cluster.local.
		svc6.testns.svc.cluster.local.	5	IN	AAAA	1234:abcd::1
		_http._tcp.svc6.testns.svc.cluster.local.	5	IN	SRV	0 100 80 svc6.testns.svc.cluster.local.
		172-0-0-2.hdls1.testns.svc.cluster.local.	5	IN	A	172.0.0.2
		_http._tcp.hdls1.testns.svc.cluster.local.	5	IN	SRV	0 25 80 172-0-0-2.hdls1.testns.svc.cluster.local.
		_http._tcp.hdls1.testns.svc.cluster.local.	5	IN	SRV	0 25 80 172-0-0-3.hdls1.testns.svc.cluster.local.
		_http._tcp.hdls1.testns.svc.cluster.local.	5	IN	SRV	0 25 80 5678-abcd--1.hdls1.testns.svc.cluster.local.
		_http._tcp.hdls1.testns.svc.cluster.local.	5	IN	SRV	0 25 80 5678-abcd--2.hdls1.testns.svc.cluster.local.
		172-0-0-3.hdls1.testns.svc.cluster.local.	5	IN	A	172.0.0.3
		5678-abcd--1.hdls1.testns.svc.cluster.local.	5	IN	AAAA	5678:abcd::1
		5678-abcd--2.hdls1.testns.svc.cluster.local.	5	IN	AAAA	5678:abcd::2
		external.testns.svc.cluster.local.	5	IN	CNAME	ext.interwebs.test.
		cluster.local.	300	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1511935515 7200 1800 86400 30
```

It kind of feels like we're absuing the etcd.msg package for passing around a dump of the zone information since we're not actually storing any of the values in etcd for this use case. Also,  extracting the record type from the service is pretty limited. I'm curious if it's worth adding some additional fields to the msg.Service struct ( such as record type ) to allow it to be more explicit and perhaps updating the RR method to return the proper record for the associated service.

plugin/kubernetes/handler.go
miekg: yes, probably faster as well - should do this in diff. PR

plugin/kubernetes/kubernetes.go
miekg: yep

plugin/kubernetes/setup.go
miekg: secondary also depends on file, if it becomes cyclic we move to pkg/...
transfer from is an error when called `secondary = false`

johnbelamaric:
FYI use this for `fallthrough` case: https://github.com/infobloxopen/themis/blob/850a2e0fdc10362d587e5e81541b7458b17843ff/contrib/coredns/policy/policy.go#L623

bradbeam:
The above kind of works. I'm able to get transfers to work some of the times, other times it seems like the messages are being received by the secondary out of order which is causing the client to keep the connection open for >2s which is causing the readdeadline timeout to get reached. It _feels_ like the issue is with https://github.com/coredns/coredns/pull/1259/files#diff-347a95a2d36b7b336cb55cc146bd16d6R81 when we start writing the message back. I put some debugging in file/secondary.go and I can see the records being received out of order sometimes.  May need some help in tracking that down.

plugin/etcd/msg/type.go
```diff
@@ -18,6 +19,10 @@ func (s *Service) HostType() (what uint16, normalized net.IP) {
 
 	ip := net.ParseIP(s.Host)
 
+	if s.Key != "" && strings.HasPrefix(Domain(s.Key), "_") {
```
miekg: Why do we need this?
HostType() can't return TypeSRV as descibed in the function docs.

plugin/kubernetes/handler.go
```diff
@@ -53,6 +57,48 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
...
+			fakew.RemoteIP = w.RemoteAddr()
+
+			_, err := plugin.NextOrFailure(k.Name(), k.Next, ctx, fakew, r)
+			if err != nil {
+				break
+			}
+
+			fakeMsg := fakew.ReadMsg()
+			// Strip out SOA records
+			records = append(records, fakeMsg.Answer[1:len(fakeMsg.Answer)-1]...)
```
miekg: I don't understand what's happening here... Why the `if k.Fallthrough`? and why the `fakewriter`

Maybe it neater rename plugin.Xfr to plugin.All() or have a private function that just grabs all records.

miekg (changes requested):

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [wip] support..." ]
>The above kind of works. I'm able to get transfers to work some of the times, other times it seems like the messages are being received by the secondary out of order which is causing the client to keep the connection open for >2s which is causing the readdeadline timeout to get reached. It _feels_ like the issue is with https://github.com/coredns/coredns/pull/1259/files#diff-347a95a2d36b7b336cb55cc146bd16d6R81 when we start writing the message back. I put some debugging in file/secondary.go and I can see the records being received out of order sometimes.  May need some help in tracking that down.

Do you have some info/details on this? For AXFR there is no out of order, only 
the first and last SOA matter.

/Miek

--
Miek Gieben

bradbeam:
Yeah, I added a simple print statement in https://github.com/coredns/coredns/blob/master/plugin/file/secondary.go#L34 to print out `env`

Works:
```
2017/12/18 05:34:28 [DEBUG] will we fail &{RR:[v803.puclf02-r01.cluster.local.	0	IN	A	10.178.245.7 v301.puclf01-r01.cluster.local.	0	IN	A	100.65.0.3 sp02-r01.cluster.local.	IN	A	10.30.0.17 v251.puclf01-r01.cluster.local.	0	IN	A	10.30.6.3 v300.puclf01-r01.cluster.local.	0	IN	A	100.65.0.1 v800.puclf01-r01.cluster.local.	0	IN	A	10.178.245.1 v255.puclf01-r01.cluster.local.	0	IN	A	10.30.6.11 puclf02-r01.cluster.local.	0	IN	A	10.30.0.11 v254.puclf01-r01.cluster.local.	0	IN	10.30.6.9 v253.puclf02-r01.cluster.local.	0	IN	A	10.30.6.7 v303.puclf02-r01.cluster.local.	0	IN	A	100.65.0.7 v256.puclf02-r01.cluster.local.	0	IN	A	10.30.6.13 sp01-r01.cluster.local.	0	IN	A	10.30.0.16 v252.puclf02-r01.cluster.local.	0	IN	A	10.30.6.5 v302.puclf02-r01.cluster.local.	0	IN	A	100.65.0.5 v801.puclf01-r01.cluster.local.	0	IN	A	10.178.245.3 v257.puclf02-r01.cluster.local.	0	IN	A	10.30.6.15 v802.puclf02-r01.cluster.local.	0	IN	A	10.178.245.5] Error:<nil>}
2017/12/18 05:34:28 [DEBUG] will we fail &{RR:[cluster.local.	0	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1513593277 7200 1800 86400 30] Error:<nil>}
2017/12/18 05:34:28 [INFO] Transferred: cluster.local. from 10.30.12.1:53
```

Failed
```
2017/12/18 05:34:24 [DEBUG] will we fail &{RR:[puc-vcsa01-witness.cluster.local.	0	IN	A	10.25.8.32 puc-vcsa01.cluster.local.	0	IN	A	10.25.3.11 v256.puclf02-r01.cluster.local.	0	IN	A	10.30.6.13 v250.puclf01-r01.cluster.local.	0	IN	A	10.30.6.1 v801.puclf01-r01.cluster.local.	0	IN	A	10.178.245.3 v252.puclf02-r01.cluster.local.	0	IN	A	10.30.6.5 v301.puclf01-r01.cluster.local.	0	IN	A	100.65.0.3 v303.puclf02-r01.cluster.local.	0	IN	A	100.65.0.7 v255.puclf01-r01.cluster.local.	0	IN	A	10.30.6.11 v254.puclf01-r01.cluster.local.	0	IN	A	10.30.6.9 v257.puclf02-r01.cluster.local.	0	IN	A	10.30.6.15 v802.puclf02-r01.cluster.local.	0	IN	A	10.178.245.5 v302.puclf02-r01.cluster.local.	0	IN	A	100.65.0.5 v253.puclf02-r01.cluster.local.	0	IN	A	10.30.6.7 v300.puclf01-r01.cluster.local.	0	IN	A	100.65.0.1 v251.puclf01-r01.cluster.local.	0	IN	A	10.30.6.3 v800.puclf01-r01.cluster.local.	0	IN	A	10.178.245.1 puclf01-r01.cluster.local.	0	IN	A	10.30.0.10 puclf02-r01.cluster.local.	0	IN	A	10.30.0.11] Error:<nil>}
2017/12/18 05:34:24 [DEBUG] will we fail &{RR:[cluster.local.	0	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1513593272 7200 1800 86400 30 sp01-r01.cluster.local.	0	IN	A	10.30.0.16 sp02-r01.cluster.local.	0	IN	A	10.30.0.17 v803.puclf02-r01.cluster.local.	0	IN	A	10.178.245.7] Error:<nil>}
2017/12/18 05:34:26 [DEBUG] will we fail &{RR:[] Error:read tcp 10.30.14.19:46276->10.30.12.1:53: i/o timeout}
2017/12/18 05:34:26 [ERROR] Failed to transfer `cluster.local.' from "10.30.12.1:53": read tcp 10.30.14.19:46276->10.30.12.1:53: i/o timeout
```

The out of order bits I'm referring to are specifically -- 
```
2017/12/18 05:34:24 [DEBUG] will we fail &{RR:[cluster.local.	0	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1513593272 7200 1800 86400 30 sp01-r01.cluster.local.	0	IN	A	10.30.0.16 sp02-r01.cluster.local.	0	IN	A	10.30.0.17 v803.puclf02-r01.cluster.local.	0	IN	A	10.178.245.7] Error:<nil>}
```
where we get a SOA record, and then two additional A records. 

For testing I also added in a debug statement here to print out `records` and the ordering looks correct there. https://github.com/coredns/coredns/pull/1259/files#diff-347a95a2d36b7b336cb55cc146bd16d6R77

plugin/kubernetes/handler.go
johnbelamaric: If fall through is set, then the k8s plugin does not know all of the records that are in the zone. Some other later plugin may be handling some records. So, we push a request down the plugin chain to gather up all those records that may be managed by someone else.

bradbeam: Yeah, what John said.
Our use case is to have host records at `cluster.local`, but then k8s services at `xxx.namespace.svc.cluster.local`. So we use kubernetes and then a fallthrough to file based for host entries. 
By using the `fakewriter` we can let the next plugin write out results locally and then inject the data into the payload we send back to the client.

plugin/etcd/msg/type.go
bradbeam: This is used in the backend_lookup.go AXFR() so we can transfer the SRV records. Otherwise, we're limited to transferring only the types that HostType() supports ( A/AAAA/CNAME ). I can update the comment to show that it's supported.

plugin/kubernetes/handler.go
miekg: hmm, I don't like the complexity this introduces. I thought it would be enough to have it just return the records it has in the cluster - not grabbing things from all over the place?

miekg:
How does AXFR fail? How are you testing? With dig?

miekg:
OK, I've just pushed https://github.com/coredns/coredns/pull/1325 which shows a bit what I meant. Showing the code is prolly more efficient than keep on commenting on a PR.

Also while doing saw the Transfer method is kinda silly and this may have led you astray. I.e we just want to grab the records from k8s and pipe them through as an axfr. The msg channel seems a bit of an overkill...

Currently broken: the records returns (I just made things up), tests obviously. 

I'll try to see if this can mimic the errors you saw and hopefully debug that.

If we want to do that `fallthrough` thing we should do that as another PR whenever the xfr stuff is merged.

bradbeam:
The tests were between coredns instances.

```
.:53 {
          cache 30
          errors
          health
          loadbalance round_robin
          log stdout
          prometheus 
          proxy . 8.8.8.8 
}
cluster.local {
          kubernetes {
            transfer to 10.30.14.19
            fallthrough
          }
          file /etc/coredns/zones.d/cluster.local.db {
            transfer to 10.30.14.19
          }
          import /etc/coredns/conf.d/cluster.local.conf
}
```

The secondary:
```
.:53 {
  errors
  health
  loadbalance
  log
  prometheus 
}
cluster.local {
  debug
  
  secondary {
    transfer from 10.30.12.1
  }	
}
```

miekg:
hmm, interesting

bradbeam:
But dig does reproduce the issue as well

miekg:
OK.

Btw you dnsrecorder change makes sense as well, however I don't think (hope) we need locking here. I'll send a PR for changing the interface a bit and the caller.

plugin/kubernetes/handler.go
johnbelamaric: Well, if we are doing a zone transfer shouldn't we include all the records in that zone, regardless of which plugin handles them? From the point of view of an external server, it doesn't care how we construct the records internally, it just needs to know that it has everything for the requested zone.

I don't think this is really any more complex than fall through in general - and allows help debugging fall through results.

miekg: No, we should do this Fallthrough crazyness. In fact it is already leading to weird to debug errors.

johnbelamaric: What errors? We should go ahead without it for this PR and then add another PR to do it, and discuss it there. To me, it seems we are responsible for making sure that the AXFR contains all the records we might return for the zone.

bradbeam:
Added in an additional test for fallthrough to at least test for what I'm seeing. Going to save off this and trim this PR down to just kubernetes records support and follow up with a fallthrough bit.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [wip] support..." ]
>Added in an additional test for fallthrough to at least test for what I'm seeing. Going to save off this and trim this PR down to just kubernetes records support and follow up with a fallthrough bit.

I won't merge this unless the fallthrough stuff is removed.

If we need this we should gather the extra records in the GetRecords() stage, 
not by abusing fallthrough like this.

bradbeam:
Understood, just wanted to finish this out train of thought. Working on removing the fallthrough bits.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [wip] support..." ]
>Understood, just wanted to finish this out train of thought. Working on removing the fallthrough bits.

Ack. We can discuss "getting all the other records" in another PR. Some thought 
on that:

* Don't use `fallthrough` but have new code that grabs all the records we 
  want/need
* Failure mode: what if you can't resolv 1 external name, do you fail the axfr?
* Failure mode: what if an external name changes? Do you check for that? 

And lastly:

If you load this incomplete axfr into another CoreDNS with the secondary plugin 
and the same upstream configured: will it have all the data?

bradbeam:
I've removed the fallthrough bits, should be ready.

johnbelamaric:
@bradbeam looks like a race condition in Travis for the XFR tests - have you checked into it?

bradbeam:
Yup, got that. 
Next issue is it looks like my changes to dnstest with buffering the whole message and not the last line is causing failures in other plugins. Probably worth checking first which should be the correct behavior. 

edit: Nevermind, they were issues caused by reusing the same `dnstest.responsewriter`. I changed the instantiation of the responsewriters so they'd be single use only and that's addressed the failures. CI now green 🕺

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [wip] support..." ]
>Yup, got that.
>Next issue is it looks like my changes to dnstest with buffering the whole message and not the last line is causing failures in other plugins. Probably worth checking first which should be the correct behavior.

please use dns.multireader

plugin/etcd/msg/type.go
miekg: HostType returns 3 things: an address, which can be a or aaaa and a name (typed as cname). Extending this is not an option.

plugin/kubernetes/setup.go
```diff
@@ -204,6 +210,32 @@ func kubernetesParse(c *caddy.Controller) (*Kubernetes, dnsControlOpts, error) {
 	return k8s, opts, nil
 }
 
+// TransferParse parses transfer statements: 'transfer to [address...]'.
+func TransferParse(c *caddy.Controller) ([]string, error) {
```
miekg: we have parse.Transfer now

plugin/loadbalance/loadbalance_test.go
```diff
@@ -79,13 +79,12 @@ func TestLoadBalance(t *testing.T) {
 		},
 	}
 
-	rec := dnstest.NewRecorder(&test.ResponseWriter{})
-
-	for i, test := range tests {
+	for i, lbtest := range tests {
+		rec := dnstest.NewRecorder(&test.ResponseWriter{})
```
miekg: please revert and use dns.multirecorder

plugin/pkg/dnstest/recorder.go
```diff
@@ -2,6 +2,7 @@
 package dnstest
 
 import (
+	"sync"
```
miekg: please revert and use dns.multirecorder

plugin/template/template_test.go
```diff
@@ -339,10 +339,9 @@ func TestMultiSection(t *testing.T) {
 
 	handler.Next = test.NextHandler(rcodeFallthrough, nil)
 
-	rec := dnstest.NewRecorder(&test.ResponseWriter{})
-
 	// Asking for test. IN A -> REFUSED
 
+	rec := dnstest.NewRecorder(&test.ResponseWriter{})
```
miekg: please revert and use dns.multirecorder

plugin/kubernetes/fake_responsewriter.go
```diff
@@ -0,0 +1,42 @@
+package kubernetes
```
miekg: not used anymore

plugin/kubernetes/handler.go
```diff
@@ -53,6 +57,37 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 			break
 		}
 		fallthrough
+	case dns.TypeAXFR:
+		records, err = plugin.AXFR(&k, zone, state, plugin.Options{})
```
miekg: please move this to a new file xfr.go

IXFR should be here as well

plugin/kubernetes/controller.go
```diff
@@ -41,9 +42,14 @@ type dnsController interface {
 	Run()
 	HasSynced() bool
 	Stop() error
+
+	// Modified returns last modified.
+	Modified() int64
 }
 
 type dnsControl struct {
+	modified int64 // modified tracks last modified
```
miekg: please add a comment that this needs to be first because we use `sync.LoadAtomatic` and therefor it needs to be on a 8 byte boundary.

plugin/kubernetes/controller.go
```diff
@@ -41,9 +42,14 @@ type dnsController interface {
 	Run()
 	HasSynced() bool
 	Stop() error
+
+	// Modified returns last modified.
```
miekg: Think this is better phrases as: `Modified returns the timestamp of the most recent changes`

plugin/kubernetes/xfr.go
```diff
@@ -1,17 +1,92 @@
...
+	serviceList := k.APIConn.ServiceList()
+	for _, svc := range serviceList {
+		switch svc.Spec.Type {
+		case api.ServiceTypeClusterIP, api.ServiceTypeNodePort, api.ServiceTypeLoadBalancer:
+
+			if net.ParseIP(svc.Spec.ClusterIP) != nil {
+				for _, p := range svc.Spec.Ports {
+
+					s := msg.Service{Host: svc.Spec.ClusterIP, Port: int(p.Port), TTL: k.ttl}
+					s.Key = strings.Join([]string{zonePath, Svc, svc.Namespace, svc.Name}, "/")
```
miekg: I think we should return RRs here if possible instead of msg.Services, you also don't need `HostType` because you make that decision here.

This also means the transfer interface need to be changed, just like it did in my PR

miekg (changes requested):

plugin/kubernetes/handler.go
bradbeam: As in create a new handler for `dns.TypeAXFR` and `dns.TypeIXFR` or move the `plugin.AXFR` out of `backend_lookup.go` ?

plugin/kubernetes/xfr.go
bradbeam: The one in #1325 ?

plugin/kubernetes/handler.go
bradbeam: Oh, I think I see what you mean based on that other PR. Following that.

bradbeam:
I think this should address most of the comments. The current test failures are due to the SRV record weight being set at 8080 versus the test cases using `25` and  `100`. Any concerns about changing these to 8080?

Edit: Changed these to 100 since that seemed to be the simplest change.

bradbeam:
Bummer, dont think my last comment made it through. Anyways, found out that it was a bad idea to just switch the weight to 100. There's two different code paths for generating SRV records, `msg.NewSRV`[1] and `plugin.SRV`[2]. Most tests utilize the dynamic weight/priority calculation of plugin.SRV, but since we're using msg.NewSRV, I needed to account for that. The path of least resistance was to copy the logic from plugin.SRV and add it to xfr.go. I'm guessing we'll need to find a better way to handle this. 

Anyways, I think I've got everything addressed and ci comes back green still. So I think we're in good shape. 

[1] https://github.com/coredns/coredns/blob/master/plugin/etcd/msg/service.go#L71-L76
[2] https://github.com/coredns/coredns/blob/master/plugin/backend_lookup.go#L152-L173

johnbelamaric:
@bradbeam is this ready for review again, all comments addressed?

bradbeam:
Yessir.

plugin/kubernetes/handler.go
```diff
@@ -53,6 +53,39 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 			break
 		}
 		fallthrough
+	case dns.TypeAXFR, dns.TypeIXFR:
+		k.Transfer(ctx, state)
+		/*
```
johnbelamaric: Why is all this commented out code still here?

plugin/kubernetes/xfr.go
```diff
@@ -1,17 +1,190 @@
...
+
+					host := emitRecord(c, s)
+
+					s.Host = host
+
+					// Need to generate this to handle use cases for peer-finder
+					// ref: https://github.com/coredns/coredns/pull/823
+					c <- s.NewSRV(msg.Domain(s.Key), 100)
+
+					s.Key = strings.Join(append(svcBase, strings.ToLower("_"+string(p.Protocol)), strings.ToLower("_"+string(p.Name))), "/")
```
johnbelamaric: Unnamed ports do not get an SRV record, according to the spec: https://github.com/kubernetes/dns/blob/master/docs/specification.md#232---srv-records

I think you are giving SRV records even for unnamed ports, unless I am reading this incorrectly?

plugin/kubernetes/handler.go
bradbeam: Poor memory. I'll get it cleaned up

plugin/kubernetes/xfr.go
bradbeam: I think you're reading it correctly. I'll clean that up.

bradbeam:
@johnbelamaric updated

plugin/kubernetes/xfr.go
```diff
@@ -1,17 +1,202 @@
...
+func (k *Kubernetes) MinTTL(state request.Request) uint32 { return 30 }
+
+// Transfer implements the Transferer interface.
+func (k *Kubernetes) Transfer(ctx context.Context, state request.Request) (int, error) {
+
+	// Get all services.
+	rrs := make(chan dns.RR)
+	go k.transfer(rrs, state.Zone)
+
+	records := []dns.RR{}
```
johnbelamaric, line 32: Is everything from here down non-K8s specific? Maybe (after merge?) this could move to a utility function that other plugins could use, so they only have to implement a function that returns a channel?

johnbelamaric (approved):
lgtm

plugin/etcd/xfr.go
```diff
@@ -4,6 +4,8 @@ import (
 	"time"
 
 	"github.com/coredns/coredns/request"
+	"github.com/miekg/dns"
```
miekg, line 7: import fmt (newline between our imports and 3rd party, sadly gofmt doesn't do this automatically)

plugin/kubernetes/xfr.go
```diff
@@ -1,17 +1,202 @@
...
-	"time"
+	"log"
+	"math"
+	"net"
+	"strings"
 
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/etcd/msg"
 	"github.com/coredns/coredns/request"
+	"golang.org/x/net/context"
```
miekg: import fmt

plugin/kubernetes/xfr.go
```diff
@@ -1,17 +1,202 @@
...
+	rrs := make(chan dns.RR)
+	go k.transfer(rrs, state.Zone)
+
+	records := []dns.RR{}
+	for r := range rrs {
+		records = append(records, r)
+	}
+
+	if len(records) == 0 {
+		return dns.RcodeServerFailure, nil
```
miekg, line 38: is this the right return code? (not super important, client will retry anyway).

plugin/kubernetes/xfr.go
```diff
@@ -1,17 +1,202 @@
...
+				if ep.ObjectMeta.Name != svc.Name || ep.ObjectMeta.Namespace != svc.Namespace {
+					continue
+				}
+
+				for _, eps := range ep.Subsets {
+					srvWeight := calcSRVWeight(len(eps.Addresses))
+					for _, addr := range eps.Addresses {
+						s := msg.Service{Host: addr.IP, TTL: k.ttl}
+						s.Key = strings.Join(svcBase, "/")
+						emitRecord(c, s)
```
miekg: why is the return value from emitRecord not used here , a few lines down it is used?
(warrants at least a comment)

plugin/kubernetes/xfr.go
```diff
@@ -1,17 +1,202 @@
...
+				c <- s.NewCNAME(msg.Domain(s.Key), s.Host)
+			}
+		}
+	}
+	return
+}
+
+// emitRecord generates a new A or AAAA record based on the msg.Service and writes it to
+// a channel. emitRecord returns the host name from the generated record.
+func emitRecord(c chan dns.RR, message msg.Service) string {
```
miekg: emitAddressRecord ? CNAME are sent using this for instance, nor SRV

plugin/kubernetes/xfr.go
```diff
@@ -1,17 +1,202 @@
...
+		host = arec.Hdr.Name
+		c <- arec
+	}
+
+	return host
+}
+
+// calcSrvWeight borrows the logic implemented in plugin.SRV for dynamically
+// calculating the srv weight and priority
+func calcSRVWeight(numservices int) uint16 {
```
miekg, line 185: prolly move this to a lib function as some point as well.

miekg (commented):
few question, but this is ready to go in.

Thanks for your efforts here!

miekg (approved):

plugin/kubernetes/xfr.go
bradbeam: It's the same one that's implemented for file.xfr  https://github.com/coredns/coredns/blob/master/plugin/file/xfr.go#L30

bradbeam:
Think I've got your comments addressed. I was a little unsure about the imports formatting, but gave it a shot. lmk if it's not quite right.

miekg:
let's not linger on that. I'm pressing buttons, stand back!

PR #1261, opened 2017-11-28
metrics: use exponentional buckets

miekg:
See  https://github.com/kubernetes/dns/pull/168


bboreham:
I would quibble over citing https://github.com/kubernetes/dns/pull/168 - that is a clear improvement from linear to exponential buckets, but your buckets were broadly exponential before.

PR #1267, opened 2017-11-30
Added reconnect feature for dnstap plugin

UladzimirTrehubenka:
In case if dnstap listener endpoint fails CoreDNS tries to reconnect to - this is helps avoid error in case network fail.
Previously CoreDNS won't start if dnstap listener endpoint has not been started yet.
none

### DIFF
--- a/plugin/dnstap/dnstapio/io.go
+++ b/plugin/dnstap/dnstapio/io.go
@@ -13,34 +13,39 @@ import (
 const (
 	tcpTimeout   = 4 * time.Second
 	flushTimeout = 1 * time.Second
-	queueSize    = 1000
+	queueSize    = 10000
 )
 
 type dnstapIO struct {
-	enc   *fs.Encoder
-	conn  net.Conn
-	queue chan tap.Dnstap
+	endpoint string
+	socket   bool
+	conn     net.Conn
+	enc      *fs.Encoder
+	queue    chan tap.Dnstap
 }
 
 // New returns a new and initialized DnstapIO.
-func New() DnstapIO {
-	return &dnstapIO{queue: make(chan tap.Dnstap, queueSize)}
+func New(endpoint string, socket bool) DnstapIO {
+	return &dnstapIO{
+		endpoint: endpoint,
+		socket:   socket,
+		queue:    make(chan tap.Dnstap, queueSize),
+	}
 }
 
 // DnstapIO interface
 type DnstapIO interface {
-	Connect(endpoint string, socket bool) error
+	Connect()
 	Dnstap(payload tap.Dnstap)
 	Close()
 }
 
-// Connect connects to the dnstop endpoint.
-func (dio *dnstapIO) Connect(endpoint string, socket bool) error {
+func (dio *dnstapIO) newConnect() error {
 	var err error
-	if socket {
-		dio.conn, err = net.Dial("unix", endpoint)
+	if dio.socket {
+		dio.conn, err = net.Dial("unix", dio.endpoint)
 	} else {
-		dio.conn, err = net.DialTimeout("tcp", endpoint, tcpTimeout)
+		dio.conn, err = net.DialTimeout("tcp", dio.endpoint, tcpTimeout)
 	}
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


plugin/dnstap/dnstapio/io.go
```diff
@@ -13,34 +13,40 @@ import (
 const (
 	tcpTimeout   = 4 * time.Second
 	flushTimeout = 1 * time.Second
-	queueSize    = 1000
```
miekg, line 16: We should remove this

plugin/dnstap/dnstapio/io.go
```diff
@@ -52,10 +58,18 @@ func (dio *dnstapIO) Connect(endpoint string, socket bool) error {
 	if err != nil {
 		return err
 	}
-	go dio.serve()
+	dio.connected = true
```
miekg: I this to be used from multiple threads? If so, this is a data race.

plugin/dnstap/dnstapio/io.go
```diff
@@ -52,10 +58,18 @@ func (dio *dnstapIO) Connect(endpoint string, socket bool) error {
...
 	return nil
 }
 
+// Connect connects to the dnstop endpoint.
+func (dio *dnstapIO) Connect() {
+	if err := dio.connect(); err != nil {
+		log.Printf("[ERROR] No connection to dnstap endpoint")
```
miekg: Should just return the error here

plugin/dnstap/dnstapio/io.go
```diff
@@ -65,36 +79,58 @@ func (dio *dnstapIO) Dnstap(payload tap.Dnstap) {
 	}
 }
 
+func (dio *dnstapIO) close() {
```
miekg: Why do we now also have `Close()`?

plugin/dnstap/dnstapio/io.go
UladzimirTrehubenka: Handling is in dio.serve(). It starts in Connect() only once.
Assume that we have Connect() on start CoreDNS and Close() on shutdown CoreDNS.
So all stuf is in one goroutine and we have no data race.

plugin/dnstap/dnstapio/io.go
UladzimirTrehubenka: Close() closes queue and serve() returns.
close() closes connection and encoder w/o stop serve()

plugin/dnstap/dnstapio/io.go
UladzimirTrehubenka: Without queue we will have "Dnstap payload dropped" all time.

plugin/dnstap/dnstapio/io.go
UladzimirTrehubenka: What about case when dnstap listener is working, we start CoreDNS and in the same time we have network fail that will be resolved quickly (on tcp Dial timeout interval)?

plugin/dnstap/dnstapio/io.go
miekg: But now we have at after N queries? It's just bad form a to have buffered channels, this always blows up in prod - but you didn't add it, so I'll let this pass. But we should follow up and remove the queueSize entirely. If nothing listens we should discard it or something?

plugin/dnstap/dnstapio/io.go
```diff
@@ -65,36 +77,59 @@ func (dio *dnstapIO) Dnstap(payload tap.Dnstap) {
 	}
 }
 
+func (dio *dnstapIO) closeConnection() {
```
miekg, line 80: My question was more; why can't all this moved to Close()? Why are there 2 functions that "close"?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Added reconne..." ]
>UladzimirTrehubenka commented on this pull request.
>
>
>
>> @@ -52,10 +58,18 @@ func (dio *dnstapIO) Connect(endpoint string, socket bool) error {
> 	if err != nil {
> 		return err
> 	}
>-	go dio.serve()
>+	dio.connected = true
>
>Handling is in dio.serve(). It starts in Connect() only once.
>Assume that we have Connect() on start CoreDNS and Close() on shutdown CoreDNS.
>So all stuf is in one goroutine and we have no data race.

Ok!

plugin/dnstap/dnstapio/io.go
UladzimirTrehubenka: Function closeConnection() is used on reconnect (before newConnection() call), Close() is used on shutdown - it just close queue and this cause closeConnection() and exit from serve().

PR #1273, opened 2017-12-03
Add NSID plugin support for CoreDNS

### LINKED ISSUE #1256: Meta data or label support for identity/whoami?
I have a unique scenario where I would like to deploy multiple CoreDNS servers in a cluster. I want each CoreDNS server in the cluster to expose certain information like label or meta-data.

CoreDNS currently support whoami plugin that could be used to show the local IP of the CoreDNS server. Though it would be nice to allow CoreDNS to be assigned with a arbitrary label or metadata and expose it through DNS or HTTP.

For example, let's say I want to deploy 3 CoreDNS servers on 3 hosts. One host is deployed in Tokyo and another two hosts are deployed in California. I would like to have the first host with the label of:
`location=tokyo` and the rest two host with the label of: `location=california`.

A feature like the above will helps a lot in a distributed system and expand the usage of CoreDNS in more interesting ways.

yongtang:

This fix adds NSID plugin support for CoreDNS, as was proposed in #1256.

This fix fixes #1256.

### DIFF
--- a/plugin/nsid/nsid.go
+++ b/plugin/nsid/nsid.go
@@ -1,0 +1,97 @@
+// Package nsid implements NSID protocol
+package nsid
+
+import (
+	"encoding/hex"
+
+	"github.com/coredns/coredns/plugin"
+
+	"github.com/miekg/dns"
+	"golang.org/x/net/context"
+)
+
+// Nsid plugin
+type Nsid struct {
+	Next plugin.Handler
+	Data string
+}
+
+// ResponseWriter is a response writer that adds NSID response
+type ResponseWriter struct {
+	dns.ResponseWriter
+	Data string
+}
+
+// ServeDNS implements the plugin.Handler interface.
+func (n Nsid) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	if option := r.IsEdns0(); option != nil {
+		for _, o := range option.Option {
+			if _, ok := o.(*dns.EDNS0_NSID); ok {
+				nw := &ResponseWriter{ResponseWriter: w, Data: n.Data}
+				return plugin.NextOrFailure(n.Name(), n.Next, ctx, nw, r)
+
+			}
+		}
+	}
+	return plugin.NextOrFailure(n.Name(), n.Next, ctx, w, r)
+}
+
+// WriteMsg implements the dns.ResponseWriter interface.
+func (w *ResponseWriter) WriteMsg(res *dns.Msg) error {
+	r1 := new(dns.Msg)
+
+	// Make a copy of Msg and update NSID
+	r1.MsgHdr = res.MsgHdr
+	r1.Compress = res.Compress
+	if len(res.Question) > 0 {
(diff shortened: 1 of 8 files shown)


plugin/nsid/nsid.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+	return plugin.NextOrFailure(n.Name(), n.Next, ctx, w, r)
+}
+
+// WriteMsg implements the dns.ResponseWriter interface.
+func (w *ResponseWriter) WriteMsg(res *dns.Msg) error {
+	r1 := new(dns.Msg)
+
+	// Make a copy of Msg and update NSID
```
miekg: Why do you copy, you should be able to get away by just adding the NSID?

plugin/nsid/setup.go
```diff
@@ -0,0 +1,43 @@
...
+
+	dnsserver.GetConfig(c).AddPlugin(func(next plugin.Handler) plugin.Handler {
+		return Nsid{Next: next, Data: nsid}
+	})
+
+	return nil
+}
+
+func nsidParse(c *caddy.Controller) (string, error) {
+	nsid := "NSID"
```
miekg: I believe the default was the hostname (I *think*), haven't read the RFC lately

miekg (commented):
Nice!
Some questions and needs (I can write it) a README.md

plugin/nsid/nsid.go
yongtang: Thanks @miekg.

Initially I modified `e.Nsid` directly, but incorrectly not converting buffer to hex in `e.Nsid`. That results in `overflow packing` error. I though the error was caused by direct modification of `e.Nsid` and copy is needed.

Later on I fixed the hex issue, but forgot to give the direct modification a try. So the usage of copy remains.

Just tried again with direct modification and it works now. Thanks!

plugin/nsid/setup.go
yongtang: Thanks @miekg. Not familiar with this area. Now reading the RFC again, it does not mention the hostname. However, I did a google search and it turns out Bind is using hostname as the default (the related config seems to be `request-nsid`). I think it makes sense to conform to bind behavior. The PR has been updated.

yongtang:
@miekg The PR has been updated. Please take a look.

miekg (approved):

miekg:
lgtm. Here's a README
```
# nsid

*nsid* add an identifier of this server to each reply.

This plugin implements RFC 5001 and adds an EDNS0 OPT resource record to replies that uniquely
identifies the server. This can be useful in anycast setups to see which server was responsible for
generating the reply and for debugging.

## Syntax

~~~ txt
nsid [DATA]
~~~

**DATA** is the string to use in the nsid record.

If **DATA** is not given, the host's name is used.

## Examples

Enable nsid:

~~~ corefile
. {
    nsid
}
~~~
```

yongtang:
@miekg Thanks 👍 ! Updated the PR with the README.md added.

PR #1285, opened 2017-12-07
Changing switch to compare on constants versus strings

bradbeam:
Minor change to kubernetes plugin to do switch comparison against constants instead of strings.

n/a

n/a


miekg:
Might make sense to do the same in etcd/handler.go (while we're at it). otherwise lgtm

bradbeam:
will do

Issue #1296, opened 2017-12-10
plugin/rewrite: regular expression and substring match/replace


greenpau:
Currently, the "rewrite" plugin matches the name in its entirety and re-writes it. It cannot match on a substring or regular expression.

greenpau:
Merged.

PR #1297, opened 2017-12-10
plugin/rewrite: regular expression and substring match/replace (#1296)

greenpau:

Provides the ability to `rewrite` plugin to match on a substring or regular expression.

#1296 

The PR contains changes to README.md file

### DIFF
--- a/plugin/rewrite/README.md
+++ b/plugin/rewrite/README.md
@@ -96,3 +96,48 @@ rewrite edns0 subnet set 24 56
 
 * If the query has source IP as IPv4, the first 24 bits in the IP will be the network subnet.
 * If the query has source IP as IPv6, the first 56 bits in the IP will be the network subnet.
+
+### Name Field Rewrites
+
+The `rewrite` plugin offers the ability to match on the name in the question section of
+a DNS request. The match could be exact, substring, or based on a prefix, suffix, or regular
+expression.
+
+The syntax for the name re-writing is as follows:
+
+```
+rewrite [continue|stop] name [exact|prefix|suffix|substring|regex] STRING STRING
+```
+
+The match type, i.e. `exact`, `substring`, etc., triggers re-write: 
+
+* **exact** (default): on exact match of the name in the question section of a request
+* **substring**: on a partial match of the name in the question section of a request
+* **prefix**: when the name begins with the matching string
+* **suffix**: when the name ends with the matching string
+* **regex**: when the name in the question section of a request matches a regular expression
+
+If the match type is omitted, the `exact` match type is being assumed.
+
+The following instruction allows re-writing the name in the query that
+contains `service.us-west-1.example.org` substring.
+
+```
+rewrite name substring service.us-west-1.example.org service.us-west-1.consul
+```
+
+Thus:
+* Incoming Request Name: `ftp.service.us-west-1.example.org`
+* Re-written Request Name: `ftp.service.us-west-1.consul`
+
+The following instruction uses regular expressions. The name in a request
+matching `(.*)-(us-west-1)\.example\.org` regular expression is being replaces with
+`{1}.service.{2}.consul`, where `{1}` and `{2}` are regular expression match groups.
+
+```
+rewrite name regex (.*)-(us-west-1)\.example\.org {1}.service.{2}.consul
+```
+
+Thus:
+* Incoming Request Name: `ftp-us-west-1.example.org`
+* Re-written Request Name: `ftp.service.us-west-1.consul`
--- a/plugin/rewrite/name.go
+++ b/plugin/rewrite/name.go
@@ -1,20 +1,59 @@
 package rewrite
 
 import (
+	"fmt"
 	"github.com/coredns/coredns/plugin"
-
 	"github.com/miekg/dns"
+	"regexp"
+	"strconv"
+	"strings"
 )
 
 type nameRule struct {
-	From, To string
+	NextAction string
+	From       string
+	To         string
 }
 
-func newNameRule(from, to string) (Rule, error) {
-	return &nameRule{plugin.Name(from).Normalize(), plugin.Name(to).Normalize()}, nil
+type prefixNameRule struct {
+	NextAction  string
+	Prefix      string
+	Replacement string
 }
 
-// Rewrite rewrites the the current request.
+type suffixNameRule struct {
+	NextAction  string
+	Suffix      string
+	Replacement string
+}
+
+type substringNameRule struct {
+	NextAction  string
+	Substring   string
+	Replacement string
+}
+
+type regexNameRule struct {
+	NextAction  string
+	Pattern     *regexp.Regexp
+	Replacement string
+}
+
+const (
+	// ExactMatch matches only on exact match of the name in the question section of a request
+	ExactMatch = "exact"
+	// PrefixMatch matches when the name begins with the matching string
+	PrefixMatch = "prefix"
+	// SuffixMatch matches when the name ends with the matching string
+	SuffixMatch = "suffix"
+	// SubstringMatch matches on partial match of the name in the question section of a request
+	SubstringMatch = "substring"
+	// RegexMatch matches when the name in the question section of a request matches a regular expression
+	RegexMatch = "regex"
+)
+
+// Rewrite rewrites the current request based upon exact match of the name
+// in the question section of the request
 func (rule *nameRule) Rewrite(w dns.ResponseWriter, r *dns.Msg) Result {
 	if rule.From == r.Question[0].Name {
 		r.Question[0].Name = rule.To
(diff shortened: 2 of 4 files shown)


greenpau:
[Getting Started with CoreDNS and Consul](https://www.linkedin.com/pulse/getting-started-coredns-consul-paul-greenberg)

plugin/rewrite/name.go
```diff
@@ -16,6 +18,36 @@ func newNameRule(from, to string) (Rule, error) {
 
 // Rewrite rewrites the the current request.
 func (rule *nameRule) Rewrite(w dns.ResponseWriter, r *dns.Msg) Result {
+	rname := r.Question[0].Name
+	// Rewrite based on substring
+	if strings.Contains(rule.From, "sub:") {
```
miekg: Shouldn't this be `strings.HasPrefix`?

plugin/rewrite/name.go
```diff
@@ -16,6 +18,36 @@ func newNameRule(from, to string) (Rule, error) {
...
+		regexPattern, err := regexp.Compile(pattern)
+		if err != nil {
+			return RewriteIgnored
+		}
+		regexGroups := regexPattern.FindStringSubmatch(rname)
+		if len(regexGroups) == 0 {
+			return RewriteIgnored
+		}
+		for i, v := range regexGroups {
+			rwi := fmt.Sprintf("{%d}", i)
```
miekg: `strconv.Itoa` will be slightly cheaper to convert the int to a string.

plugin/rewrite/README.md
```diff
@@ -96,3 +96,24 @@ rewrite edns0 subnet set 24 56
 
 * If the query has source IP as IPv4, the first 24 bits in the IP will be the network subnet.
 * If the query has source IP as IPv6, the first 56 bits in the IP will be the network subnet.
+
+### Substrings and Regular Expressions
+
+The `sub:` substring in `FROM` part of the `rewrite` syntax allows to match based
```
miekg: Wouldn't it be better to do this in the `FIELD` part, like `name_sub` or `name_regexp` ?

miekg (changes requested):
I don't think I understand why or how `sub` works exactly, but the regexp stuff is nice.

Does need tests though.

WDYT: @huynhmb359

plugin/rewrite/README.md
greenpau: @miekg , my thinking is that `name` is the `FIELD` and I would think that leaving it intact makes more sense because `name_sub` and `name_regexp` sound more line an "action".

In my mind having `sub:` and `regex:` as "prefix" works best in `FROM`, because, in theory, there should be no "name" with `:` (colon) in it.

However, as someone who coded my favorite DNS library 🥇  , I will side with whatever choice you prefer 😄

plugin/rewrite/name.go
greenpau: @miekg , for the context, the `sub:` was born prior to `regex:`. In its current state `rewrite` requires full match. The `sub:` allows to match and replace a substring in a name. For example, I want to replace `contoso.com` with `service.consul`. I would do it with `sub:contoso.com service.consul` and no regular expressions.

>  Shouldn't this be `strings.HasPrefix`?

Agree. 👍 will re-work

plugin/rewrite/name.go
greenpau: > strconv.Itoa will be slightly cheaper to convert the int to a string.

Agree 👍 will re-work

greenpau:
@miekg , I just discovered an issue with my code. In a nutshell, when I set `rule.From` for the purposes of `reverter`, it would successfully resolve a single request. However, it would permanently modify the `rule` (from its original regex version) and subsequent request would not match.

```
        rule.From = r.Question[0].Name
        r.Question[0].Name = rule.To
```

If I don't modify it, i.e. comment it out, it results in `;; Question section mismatch: got ... /A/IN`

```
        // rule.From = r.Question[0].Name
        r.Question[0].Name = rule.To
```

It is a good puzzle to solve.

... One way is to dynamically add new rules. Personally, I don't like the idea.

plugin/rewrite/name.go
```diff
@@ -16,6 +18,36 @@ func newNameRule(from, to string) (Rule, error) {
 
 // Rewrite rewrites the the current request.
 func (rule *nameRule) Rewrite(w dns.ResponseWriter, r *dns.Msg) Result {
+	requestName := r.Question[0].Name
+	// Rewrite based on substring match
+	if strings.HasPrefix(rule.From, "sub:") {
```
johnbelamaric: This is static to the rule. You should create a constant for matchType (full, substring, regex) and then set it in the rule as a field. Then you don't need string comparisons. Also, you would parse out the `substituteStr` and `regexPattern` and pre-compile the regex at config parse time rather than query processing time.

plugin/rewrite/name.go
```diff
@@ -16,6 +18,36 @@ func newNameRule(from, to string) (Rule, error) {
...
+		if len(regexGroups) == 0 {
+			return RewriteIgnored
+		}
+		for groupIndex, groupValue := range regexGroups {
+			groupIndexStr := "{" + strconv.Itoa(groupIndex) + "}"
+			if strings.Contains(rule.To, groupIndexStr) {
+				rule.To = strings.Replace(rule.To, groupIndexStr, groupValue, -1)
+			}
+		}
+		rule.From = r.Question[0].Name
```
johnbelamaric: Why do you change the rule.From here? To trigger the "exact match" code? You can't change the rule at this point (which you mention in your latest comment).

You can move the exact match code into a separate function and then call it from here as well as from the matchType `exact` (better constant name than `full` I said above). In fact each matchType probably should be in a function to make this a simple switch statement.

greenpau: > You can move the exact match code into a separate function and then call it from here as well as from the matchType exact (better constant name than full I said above). In fact each matchType probably should be in a function to make this a simple switch statement.

👍

johnbelamaric:
Regarding `;; Question section mismatch: got ... /A/IN`, I am not following. Why does rule.From have anything to do with the `ResponseReverter`, which should have already been initialized with the original question?

greenpau:
> Regarding ;; Question section mismatch: got ... /A/IN, I am not following. Why does rule.From have anything to do with the ResponseReverter, which should have already been initialized with the original question?

@johnbelamaric , I am still reading through the code. It likely has nothing to do with ResponseReverter.

plugin/rewrite/name.go
greenpau: > You should create a constant for matchType (full, substring, regex) and then set it in the rule as a field. 

@johnbelamaric , I like the idea. 

> Also, you would parse out the substituteStr and regexPattern and pre-compile the regex at config parse time rather than query processing time.

Makes sense.

greenpau:
@johnbelamaric, please check see the latest and let me know your thoughts.

plugin/rewrite/README.md
```diff
@@ -96,3 +96,44 @@ rewrite edns0 subnet set 24 56
...
+The syntax for the name re-writing is as follows:
+
+```
+rewrite [continue|stop] name [exact|partial|regex] STRING STRING
+```
+
+The match type, i.e. `exact|partial|regex`, triggers re-write: 
+
+* **exact**: on exact match of the name in the question section of a request
+* **partial**: on a partial match of the name in the question section of a request
```
johnbelamaric: Probably want `prefix` and `suffix` too, as they should be a lot faster than a general partial match (you can test and make sure).

plugin/rewrite/README.md
```diff
@@ -96,3 +96,44 @@ rewrite edns0 subnet set 24 56
...
+
+The syntax for the name re-writing is as follows:
+
+```
+rewrite [continue|stop] name [exact|partial|regex] STRING STRING
+```
+
+The match type, i.e. `exact|partial|regex`, triggers re-write: 
+
+* **exact**: on exact match of the name in the question section of a request
```
johnbelamaric: Is this the default? It should say so.

plugin/rewrite/name.go
```diff
@@ -1,29 +1,116 @@
...
+        case PartialMatch:
+            return &partialNameRule{nextAction, plugin.Name(args[1]).Normalize(), plugin.Name(args[2]).Normalize()}, nil
+        case RegexMatch:
+            regexPattern, err := regexp.Compile(args[1])
+            if err != nil {
+                return nil, fmt.Errorf("Invalid regex pattern in a name rule: %s", args[1])
+            }
+            return &regexNameRule{nextAction, regexPattern, plugin.Name(args[2]).Normalize()}, nil
+        default:
+            return nil, fmt.Errorf("A name rule supports only exact, partial, and regex name matching")
```
johnbelamaric: This is not backwards compatible.

`rewrite name foo.com bar.com`

will now fail to parse. I think you need a case with 2 args.

plugin/rewrite/rewrite.go
```diff
@@ -100,12 +97,12 @@ func newRule(args ...string) (Rule, error) {
...
-	if ruleType != "edns0" && expectNumArgs != 3 {
+	if ruleType != "edns0" && ruleType != "name" && expectNumArgs != 3 {
 		return nil, fmt.Errorf("%s rules must have exactly two arguments", ruleType)
 	}
 	switch ruleType {
 	case "name":
-		return newNameRule(args[startArg], args[startArg+1])
+		return newNameRule(mode, args[startArg:]...)
 	case "class":
 		return newClassRule(args[startArg], args[startArg+1])
```
johnbelamaric, line 110: Unrelated to this PR, but just seeing that these don't use `mode`. Does the continue/stop thing not work for these rule types? The docs imply otherwise.

plugin/rewrite/name.go
```diff
@@ -1,29 +1,116 @@
...
-	"github.com/coredns/coredns/plugin"
-
-	"github.com/miekg/dns"
+    "github.com/coredns/coredns/plugin"
+    "github.com/miekg/dns"
+    "fmt"
+    "strings"
+    "regexp"
+    "strconv"
 )
```
johnbelamaric: What are these changes? Is this a different `go fmt` or something?

johnbelamaric (changes requested):
Nice, looking better. A few comments and questions.

plugin/rewrite/rewrite.go
greenpau: Will check.

plugin/rewrite/name.go
greenpau: I use vim-go. Will format with “go fmt” next time I rebase.

plugin/rewrite/README.md
greenpau: Ack. Will rework.

plugin/rewrite/README.md
greenpau: Will add. Planning to use TrimLeft and TrimRight functions for HasPrefix and HasSuffix functions.

plugin/rewrite/rewrite.go
greenpau: Prior to this PR the name rule always returned “stop” when match was found. Here, i added NextAction to the rules. Then, on Mode(), the function returns the value in the rule.

Now, stop/continue actually works. 

Will rebase and push in the evening.

johnbelamaric: Yes. Can you do the same for the other rule types in a separate PR?

greenpau: >  Yes. Can you do the same for the other rule types in a separate PR?

@johnbelamaric , sure. Please create an issue so I can create a PR against it.

greenpau:
> @miekg: Does need tests though.

>>  The diff coverage is 100%.

@miekg, @johnbelamaric, any feedback?

plugin/rewrite/README.md
```diff
@@ -96,3 +96,47 @@ rewrite edns0 subnet set 24 56
 
 * If the query has source IP as IPv4, the first 24 bits in the IP will be the network subnet.
 * If the query has source IP as IPv6, the first 56 bits in the IP will be the network subnet.
+
+### Name Field Rewrites
+
+The `rewrite` plugin offers the ability to match on the name in the question section of
+a DNS request. The match could be exact, substring, or based on a regular expression, etc.
```
johnbelamaric: s/, etc//

plugin/rewrite/README.md
```diff
@@ -96,3 +96,47 @@ rewrite edns0 subnet set 24 56
...
+
+The match type, i.e. `exact`, `substring`, etc., triggers re-write: 
+
+* **exact** (default): on exact match of the name in the question section of a request
+* **substring**: on a partial match of the name in the question section of a request
+* **prefix**: when the name begins with the matching string
+* **suffix**: when the name ends with the matching string
+* **regex**: when the name in the question section of a request matches a regular expression
+
+If the match type is ommitted, the `exact` match type is being assumed.
```
johnbelamaric: s/ommitted/omitted/

plugin/rewrite/name.go
```diff
@@ -23,7 +62,100 @@ func (rule *nameRule) Rewrite(w dns.ResponseWriter, r *dns.Msg) Result {
...
+		r.Question[0].Name = strings.Replace(r.Question[0].Name, rule.Substring, rule.Replacement, -1)
+		return RewriteDone
+	}
+	return RewriteIgnored
+}
+
+// Rewrite rewrites the current request when the name in the question
+// section of the request matches a regular expression
+func (rule *regexNameRule) Rewrite(w dns.ResponseWriter, r *dns.Msg) Result {
+	regexGroups := rule.Pattern.FindStringSubmatch(r.Question[0].Name)
```
johnbelamaric, line 96: Is there a reason not to use https://godoc.org/regexp#Regexp.ReplaceAllString ?

johnbelamaric (changes requested):
Some nits and a question.

plugin/rewrite/rewrite.go
johnbelamaric: #1307

plugin/rewrite/name.go
greenpau: I don’t like $ (dollar signs) in configuration files. The {} brackets work better when doing cat EOF type pasting.

Also replaceallstring does not give you regex groups.

plugin/rewrite/rewrite.go
greenpau: Perfect. Will create a PR after the merge of this one.

plugin/rewrite/name.go
johnbelamaric: OK. {} matches other variable substitutions in this plugin.

johnbelamaric (approved):

johnbelamaric:
Awesome, thanks, LGTM!

Issue #1289, opened 2017-12-08
Feature request: response templating


rtreffer:
We have a few places where we resolve either A records or PTR records based on templates. So far we've usually templated out whole RFC 1035 zone files.
We also have cases where we know that an appended search domain won't resolve and we could short-circuit it to NXDOMAIN.

This is related to the `reverse` plugin which was not an option due to the limited templating and limited scope.

***I'd like to suggest (and implement) a template plugin that can generate responses based on the incoming query and a response template.***

It would combine
* A [regexp](https://golang.org/pkg/regexp/) match
* [Go templates](https://golang.org/pkg/text/template/)
* [RFC 1035 style RR fragments](https://github.com/miekg/dns/blob/3bbde607ac2132f47c307b292594a1af4f619fde/scan.go#L105)

I managed to get the following working locally (Note: this is only my draft to see if it would work, everything up for debate):
```
.:1053 {
    proxy . 8.8.8.8

    template ANY ANY "[.]invalid[.]$" {
      rcode NXDOMAIN
      answer "invalid. 60 {{ $class }} SOA a.invalid. b.invalid. (1 60 60 60 60)"
    }
    template IN A "ip-10-(?P<b>[0-9]*)-(?P<c>[0-9]*)-(?P<d>[0-9]*)[.]example[.]$" {
      answer "{{ $name }} 60 IN A 10.{{ $.b }}.{{ $.c }}.{{ $.d }}"
    }
    template IN PTR "(?P<d>[0-9]*).(?P<c>[0-9]*).(?P<b>[0-9]*).10.in-addr.arpa.arpa.$" {
      answer "{{ $name }} 60 IN PTR ip-10-{{ $.b }}-{{ $.c }}-{{ $.d }}.example."
    }
    template IN MX ip-10-(?P<b>[0-9]*)-(?P<c>[0-9]*)-(?P<d>[0-9]*)[.]example[.]$ {
      answer "{{ $name }} 60 IN MX 10 {{ $name }}"
      additional "{{ $name }} 60 IN A 10.{{ $.b }}.{{ $.c }}.{{ $.d }}"
    }
}
```

The only problematic thing I've found so far is that the caddyfile parser tries to replace `{$...}` which means `{{$name}}` would break caddy/coredns while `{{ $name }}` works.

Wiring this up is pretty straight forward, simpler than reverse and way more flexible at the cost of more you need to know to use it (regexp, go templates, RFC 1035 syntax)

miekg:
[...]

PR #1298, opened 2017-12-10
plugin/template

rtreffer:
The template plugin matches the incoming query by class, type and regex and templates a response with a go templates.

As suggested in #1289

A README.md is in place. I haven't touched the root README.md.

Missing: 
1. ~~@miekg I haven't changed the rcode directive yet. Can you think of any rcode besides NXDOMAIN and ServFail that might be useful? In that case I'd go with `error|denial`.~~
2. ~~Metrics (esp. internal failures to template)~~

### DIFF
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -1,30 +1,32 @@
 # template
 
-*template* allows for dynamic responses base on the incoming query.
+*template* allows for dynamic responses based on the incoming query.
 
 ## Syntax
 
 ~~~
 template CLASS TYPE [REGEX...] {
-    [answer section]
-    [answer section]
-    [additional section]
+    [answer RR]
+    [answer RR]
+    [additional RR]
     [...]
-    [rcode  responsecode]
+    [rcode responsecode]
 }
 ~~~
 
 * **CLASS** the query class (usually IN or ANY)
 * **TYPE** the query type (A, PTR, ...)
-* **REGEX** regex that are matched against the incoming query. Specifying no regex matches everything.
-* `section` A RFC 1035 style fragment build by a go template that contains the answer.
-* `resplonsecode` A response code (NXDOMAIN, SERVFAIL, ...)
+* **REGEX** [Go regexp](https://golang.org/pkg/regexp/) that are matched against the incoming query. Specifying no regex matches everything (default: `.*`). First matching regex wins.
+* `RR` A [RFC 1035](https://tools.ietf.org/html/rfc1035#section-5) style `<rr>` fragment build by a [Go template](https://golang.org/pkg/text/template/) that contains the answer.
+* `responsecode` A response code (`NXDOMAIN, SERVFAIL, ...`). The default is `SUCCESS`.
 
 At least one answer section or rcode is needed.
 
+[Also see](#also-see) contains an additional reading list.
+
 ## Templates
 
-Each answer section is a full-featured go templated with the following predefined data
+Each resource record is a full-featured [Go template](https://golang.org/pkg/text/template/) with the following predefined data
 * `.Name` the query name, as a string
 * `.Class` the query class (usually `IN`)
 * `.Type` the RR type requested (e.g. `PTR`)
@@ -33,15 +35,15 @@ Each answer section is a full-featured go templated with the following predefine
 * `.Message` the incoming DNS query message.
 * `.Question` the matched question section.
 
-The output of the template must be a RFC 1035 style RR line (commonly refered to as a "zone file").
+The output of the template must be a [RFC 1035](https://tools.ietf.org/html/rfc1035) style resource record line (commonly refered to as a "zone file").
 
-**WARNING** there is a syntactical problem with go templates and caddy config files. Expressions like `{{$var}}` will be interpreted as a reference to an envorionment variable while `{{ $var }}` will work. Try to avoid template variables.
+**WARNING** there is a syntactical problem with Go templates and caddy config files. Expressions like `{{$var}}` will be interpreted as a reference to an environment variable by caddy/coredns while `{{ $var }}` will work. Try to avoid template variables. See [Bugs](#bugs)
 
 ## Examples
 
 ### Resolve .invalid as NXDOMAIN
 
-The `.invalid` domain is a reserved TLD to indicate non-existing domains.
+The `.invalid` domain is a reserved TLD (see [RFC-2606 Reserved Top Level DNS Names](https://tools.ietf.org/html/rfc2606#section-2)) to indicate invalid domains.
 
 ~~~ corefile
 . {
@@ -62,7 +64,9 @@ The `.invalid` domain is a reserved TLD to indicate non-existing domains.
 
 Imagine you run `example.com` with a datacenter `dc1.example.com`. The datacenter domain
 is part of the DNS search domain.
-However `something.example.com.dc1.example.com` would usually indicate a wrong autocomplete.
+However `something.example.com.dc1.example.com` would indicates a fully qualified
+domain name (`something.example.com`) that inadvertely has the default domain or search
+path (`dc1.example.com`) added.
 
 ~~~ corefile
 . {
@@ -97,7 +101,7 @@ However `something.example.com.dc1.example.com` would usually indicate a wrong a
 }
 ~~~
 
-An IP consists of 4 bytes, `a.b.c.d`. Named groups make it less error prone to reverse the
+An IPv4 address consists of 4 bytes, `a.b.c.d`. Named groups make it less error prone to reverse the
 ip in the PTR case. Try to use named groups to explain what your regex and template are doing.
 
 Note that the A record is actually a wildcard, any subdomain of the ip will resolve to the ip.
@@ -133,3 +137,17 @@ Named capture groups can be used to template one response for multiple patterns.
     }
 }
 ~~~
+
+# Also see
+
+- [Go regexp](https://golang.org/pkg/regexp/) for details about the regex implementation
+- [RE2 syntax reference](https://github.com/google/re2/wiki/Syntax) for details about the regex syntax
+- [RFC-1034](https://tools.ietf.org/html/rfc1034#section-3.6.1) and [RFC 1035](https://tools.ietf.org/html/rfc1035#section-5) for the resource record format
+- [Go template](https://golang.org/pkg/text/template/) for the template language reference
+
+# Bugs
+
+CoreDNS supports [caddyfile environment variables](https://caddyserver.com/docs/caddyfile#env)
+with notion of `{$ENV_VAR}`. This parser feature will break [Go template variables](https://golang.org/pkg/text/template/#hdr-Variables) notations like`{{$variable}}`.
+The equivalent notation `{{ $variable }}` will work.
+Try to avoid Go template variables in the context of this plugin.
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -39,6 +39,15 @@ The output of the template must be a [RFC 1035](https://tools.ietf.org/html/rfc1
 
 **WARNING** there is a syntactical problem with Go templates and caddy config files. Expressions like `{{$var}}` will be interpreted as a reference to an environment variable by caddy/coredns while `{{ $var }}` will work. Try to avoid template variables. See [Bugs](#bugs)
 
+## Metrics
+
+If monitoring is enabled (via the *prometheus* directive) then the following metrics are exported:
+- `coredns_template_matches_total` the total number of matched requests.
+- `coredns_template_template_failures_total` the number of times the Go templating failed.
+- `coredns_template_rr_failures_total` the number of times the templated resource record was invalid and could not be parsed.
+
+Both failure cases indicate a problem with the template configuration.
+
 ## Examples
 
 ### Resolve .invalid as NXDOMAIN
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -41,7 +41,7 @@ Each resource record is a full-featured [Go template](https://golang.org/pkg/tex
 
 The output of the template must be a [RFC 1035](https://tools.ietf.org/html/rfc1035) style resource record line (commonly refered to as a "zone file").
 
-**WARNING** there is a syntactical problem with Go templates and caddy config files. Expressions like `{{$var}}` will be interpreted as a reference to an environment variable by caddy/coredns while `{{ $var }}` will work. Try to avoid template variables. See [Bugs](#bugs)
+**WARNING** there is a syntactical problem with Go templates and caddy config files. Expressions like `{{$var}}` will be interpreted as a reference to an environment variable by caddy/coredns while `{{ $var }}` will work. Try to avoid template variables. See [Bugs](#bugs).
 
 ## Metrics
 
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -31,7 +31,7 @@ Each resource record is a full-featured [Go template](https://golang.org/pkg/tex
 * `.Class` the query class (usually `IN`)
 * `.Type` the RR type requested (e.g. `PTR`)
 * `.Match` an array of all matches. `index .Match 0` refers to the whole match.
-* `.Group` a map fo the named capture groups.
+* `.Group` a map of the named capture groups.
 * `.Message` the incoming DNS query message.
 * `.Question` the matched question section.
 
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -13,6 +13,7 @@ template CLASS TYPE [REGEX...] {
     [answer RR]
     [answer RR]
     [additional RR]
+    [authority RR]
     [...]
     [rcode responsecode]
 }
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -1,6 +1,10 @@
 # template
 
-*template* allows for dynamic responses based on the incoming query.
+*template* - allows for dynamic responses based on the incoming query.
+
+## Description
+
+The *template* plugin allows you to dynamically repond to queries by just writing a (Go) template.
 
 ## Syntax
 
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -47,9 +47,9 @@ The output of the template must be a [RFC 1035](https://tools.ietf.org/html/rfc1
 ## Metrics
 
 If monitoring is enabled (via the *prometheus* directive) then the following metrics are exported:
-- `coredns_template_matches_total` the total number of matched requests.
-- `coredns_template_template_failures_total` the number of times the Go templating failed.
-- `coredns_template_rr_failures_total` the number of times the templated resource record was invalid and could not be parsed.
+- `coredns_template_matches_total{regex}` the total number of matched requests by regex.
+- `coredns_template_template_failures_total{regex,section,template}` the number of times the Go templating failed. Regex, section and template label values can be used to map the error back to the config file.
+- `coredns_template_rr_failures_total{regex,section,template}` the number of times the templated resource record was invalid and could not be parsed. Regex, section and template label values can be used to map the error back to the config file.
 
 Both failure cases indicate a problem with the template configuration.
 
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -110,9 +110,7 @@ Having templates to map certain PTR/A pairs is a common pattern.
 . {
     proxy 8.8.8.8
 
-    template IN A
-        ^ip-(?P<a>10)-(?P<b>[0-9]*)-(?P<c>[0-9]*)-(?P<d>[0-9]*)[.]dc[.]example[.]$
-        ^(?P<a>[0-9]*)[.](?P<b>[0-9]*)[.](?P<c>[0-9]*)[.](?P<d>[0-9]*)[.]ext[.]example[.]$ {
+    template IN A "^ip-(?P<a>10)-(?P<b>[0-9]*)-(?P<c>[0-9]*)-(?P<d>[0-9]*)[.]dc[.]example[.]$" "^(?P<a>[0-9]*)[.](?P<b>[0-9]*)[.](?P<c>[0-9]*)[.](?P<d>[0-9]*)[.]ext[.]example[.]$" {
       answer "{{ .Name }} 60 IN A {{ .Group.a}}.{{ .Group.b }}.{{ .Group.c }}.{{ .Group.d }}"
     }
 }
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -20,7 +20,7 @@ template CLASS TYPE [REGEX...] {
 
 * **CLASS** the query class (usually IN or ANY)
 * **TYPE** the query type (A, PTR, ...)
-* **REGEX** [Go regexp](https://golang.org/pkg/regexp/) that are matched against the incoming query. Specifying no regex matches everything (default: `.*`). First matching regex wins.
+* **REGEX** [Go regexp](https://golang.org/pkg/regexp/) that are matched against the incoming question name. Specifying no regex matches everything (default: `.*`). First matching regex wins.
 * `RR` A [RFC 1035](https://tools.ietf.org/html/rfc1035#section-5) style `<rr>` fragment build by a [Go template](https://golang.org/pkg/text/template/) that contains the answer.
 * `responsecode` A response code (`NXDOMAIN, SERVFAIL, ...`). The default is `SUCCESS`.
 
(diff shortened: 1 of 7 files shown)


miekg:
Besides NXDOMAIN there is also NODATA (name exist, type does not). But NODATA does not have an rcode it is just SUCCESS; so you need that anyway which means you'll need `denial`.
SERVFAIL, REFUSED, NOTIMPL are a few server failures that come up mind.

Update: ah the response code is only used in the reply, yeah then you don't need to do the above.

plugin/template/README.md
```diff
@@ -0,0 +1,135 @@
+# template
+
+*template* allows for dynamic responses base on the incoming query.
```
miekg: *based

plugin/template/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Syntax
+
+~~~
+template CLASS TYPE [REGEX...] {
+    [answer section]
```
miekg: is `answer` a keyword? should this be `answer SECTION`?
(not sure how to read this yet)

miekg: no authoritity section?

plugin/template/README.md
```diff
@@ -0,0 +1,135 @@
...
+    [additional section]
+    [...]
+    [rcode  responsecode]
+}
+~~~
+
+* **CLASS** the query class (usually IN or ANY)
+* **TYPE** the query type (A, PTR, ...)
+* **REGEX** regex that are matched against the incoming query. Specifying no regex matches everything.
+* `section` A RFC 1035 style fragment build by a go template that contains the answer.
```
miekg: *Go

plugin/template/README.md
```diff
@@ -0,0 +1,135 @@
+# template
+
+*template* allows for dynamic responses base on the incoming query.
+
+## Syntax
+
+~~~
+template CLASS TYPE [REGEX...] {
```
miekg: what happens when multiple template match? First one wins?

plugin/template/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+The output of the template must be a RFC 1035 style RR line (commonly refered to as a "zone file").
+
+**WARNING** there is a syntactical problem with go templates and caddy config files. Expressions like `{{$var}}` will be interpreted as a reference to an envorionment variable while `{{ $var }}` will work. Try to avoid template variables.
+
+## Examples
+
+### Resolve .invalid as NXDOMAIN
+
+The `.invalid` domain is a reserved TLD to indicate non-existing domains.
```
miekg: really? Didn't know that (can add a RFC reference, i.e. See RFC XXXX)

plugin/template/README.md
miekg: or better `answer RR`

plugin/template/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+1. A query to .invalid will result in NXDOMAIN (rcode)
+2. A dummy SOA record is send to hand out a TTL of 60s for caching
+3. Querying `.invalid` of `CH` will also cause a NXDOMAIN/SOA response
+
+### Block invalid search domain completions
+
+Imagine you run `example.com` with a datacenter `dc1.example.com`. The datacenter domain
+is part of the DNS search domain.
+However `something.example.com.dc1.example.com` would usually indicate a wrong autocomplete.
```
miekg: autocomplete -> indicates a non fully qualified domain that inadvertely has the default domain added or search path added.

plugin/template/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+    # d.c.b.a.in-addr.arpa PTR ip-a-b-c-d.example
+
+    template IN PTR ^(?P<d>[0-9]*)[.](?P<c>[0-9]*)[.](?P<b>[0-9]*)[.]10[.]in-addr[.]arpa[.]$ {
+      answer "{{ .Name }} 60 IN PTR ip-10-{{ .Group.b }}-{{ .Group.c }}-{{ .Group.d }}.example.com."
+    }
+}
+~~~
+
+An IP consists of 4 bytes, `a.b.c.d`. Named groups make it less error prone to reverse the
```
miekg: An v4 IP address

plugin/template/README.md
```diff
@@ -0,0 +1,135 @@
...
+### Resolve A/PTR for .example
+
+~~~ corefile
+. {
+    proxy . 8.8.8.8
+
+    # ip-a-b-c-d.example.com A a.b.c.d
+
+    template IN A (^|[.])ip-10-(?P<b>[0-9]*)-(?P<c>[0-9]*)-(?P<d>[0-9]*)[.]example[.]$ {
+      answer "{{ .Name }} 60 IN A 10.{{ .Group.b }}.{{ .Group.c }}.{{ .Group.d }}"
```
miekg: No RCODE mean SUCCESS?

plugin/template/README.md
```diff
@@ -0,0 +1,135 @@
...
+    
+    template IN A ^ip-10-(?P<b>[0-9]*)-(?P<c>[0-9]*)-(?P<d>[0-9]*)[.]example[.]$ {
+      answer "{{ .Name }} 60 IN A 10.{{ .Group.b }}.{{ .Group.c }}.{{ .Group.d }}"
+    }
+    template IN MX ^ip-10-(?P<b>[0-9]*)-(?P<c>[0-9]*)-(?P<d>[0-9]*)[.]example[.]$ {
+      answer "{{ .Name }} 60 IN MX 10 {{ .Name }}"
+      additional "{{ .Name }} 60 IN A 10.{{ .Group.b }}.{{ .Group.c }}.{{ .Group.d }}"
+    }
+}
+~~~
```
miekg, line 153: Might be worth adding a "See Also" section that points to Go templates, godoc.org link? And maybe RFC 1035 for the text presentation?

Also add BUG section that duplicates the warning in the intro about caddy file eating the templates when the spacing is off.

plugin/template/setup.go
```diff
@@ -0,0 +1,127 @@
+package template
+
+import (
+	"regexp"
+	gotmpl "text/template"
+
+	"github.com/coredns/coredns/core/dnsserver"
+	"github.com/coredns/coredns/plugin"
+	"github.com/mholt/caddy"
```
miekg: newline between "our" imports and third party.

plugin/template/setup.go
```diff
@@ -0,0 +1,127 @@
...
+func templateParse(c *caddy.Controller) (templates []template, err error) {
+	templates = make([]template, 0)
+
+	for c.Next() {
+		t := template{}
+		if !c.NextArg() {
+			return nil, c.ArgErr()
+		}
+
+		class, found := dns.StringToClass[c.Val()]
```
miekg: typical go is to use `ok` instead of found.

plugin/template/setup.go
```diff
@@ -0,0 +1,127 @@
...
+		if !found {
+			return nil, c.Errf("Invalid query class %s", c.Val())
```
miekg: Errors always start with a lowercase

plugin/template/setup.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+		t.class = class
+
+		if !c.NextArg() {
+			return nil, c.ArgErr()
+		}
+		queryType, found := dns.StringToType[c.Val()]
+		if !found {
+			return nil, c.Errf("Invalid query rr type %s", c.Val())
```
miekg: drop 'query' and uppercase rr -> RR

plugin/template/template.go
```diff
@@ -0,0 +1,138 @@
+package template
+
+import (
+	"bytes"
+	"context"
+	"regexp"
+	"strconv"
+
```
miekg: drop this newline

plugin/template/template.go
```diff
@@ -0,0 +1,138 @@
...
+	gotmpl "text/template"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/request"
+	"github.com/miekg/dns"
```
miekg: newline above miekg/dns

plugin/template/template.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	return plugin.NextOrFailure(h.Name(), h.Next, ctx, w, r)
+}
+
+// Name implements the plugin.Handler interface.
+func (h Handler) Name() string {
+	return "template"
+}
+
+func executeRRTemplate(template *gotmpl.Template, data templateData) (*dns.RR, error) {
```
miekg: why does this return a pointer to dns.RR?

plugin/template/template.go
```diff
@@ -0,0 +1,138 @@
...
+	buffer := &bytes.Buffer{}
+	err := template.Execute(buffer, data)
+	if err != nil {
+		return nil, err
+	}
+	rrDefinition := buffer.String()
+	rr, err := dns.NewRR(rrDefinition)
```
miekg: `rr, err := dns.NewRR(buffer.String())`

plugin/template/template.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &rr, nil
+}
+
+func (t template) match(r *dns.Msg) (templateData, bool) {
+	for _, q := range r.Question {
```
miekg: There is only 1 question in the question section (long story), so there is no need to range. CoreDNS quarantees you have one before calling into the plugins.

miekg (changes requested):
Whole bunch of comments, but this is a good PR. Mostly nits and smaller things.

rtreffer:
@miekg I've added some metrics, I think I've addressed all your concerns. I'm dropping the WIP, can you do another review? (I was less available over the holidays than I've hoped)

plugin/template/README.md
rtreffer: Added to the README

plugin/template/README.md
```diff
@@ -0,0 +1,162 @@
...
+[Also see](#also-see) contains an additional reading list.
+
+## Templates
+
+Each resource record is a full-featured [Go template](https://golang.org/pkg/text/template/) with the following predefined data
+* `.Name` the query name, as a string
+* `.Class` the query class (usually `IN`)
+* `.Type` the RR type requested (e.g. `PTR`)
+* `.Match` an array of all matches. `index .Match 0` refers to the whole match.
+* `.Group` a map fo the named capture groups.
```
johnbelamaric: s/fo/of/

rtreffer: fixed

plugin/template/README.md
```diff
@@ -0,0 +1,166 @@
...
+    [answer RR]
+    [additional RR]
+    [...]
+    [rcode responsecode]
+}
+~~~
+
+* **CLASS** the query class (usually IN or ANY)
+* **TYPE** the query type (A, PTR, ...)
+* **REGEX** [Go regexp](https://golang.org/pkg/regexp/) that are matched against the incoming query. Specifying no regex matches everything (default: `.*`). First matching regex wins.
```
miekg: incoming "question name"?

plugin/template/README.md
```diff
@@ -0,0 +1,166 @@
...
+* `.Type` the RR type requested (e.g. `PTR`)
+* `.Match` an array of all matches. `index .Match 0` refers to the whole match.
+* `.Group` a map of the named capture groups.
+* `.Message` the incoming DNS query message.
+* `.Question` the matched question section.
+
+The output of the template must be a [RFC 1035](https://tools.ietf.org/html/rfc1035) style resource record line (commonly refered to as a "zone file").
+
+**WARNING** there is a syntactical problem with Go templates and caddy config files. Expressions like `{{$var}}` will be interpreted as a reference to an environment variable by caddy/coredns while `{{ $var }}` will work. Try to avoid template variables. See [Bugs](#bugs)
+
```
miekg: Final dot is missing.

plugin/template/README.md
```diff
@@ -0,0 +1,166 @@
...
+## Metrics
+
+If monitoring is enabled (via the *prometheus* directive) then the following metrics are exported:
+- `coredns_template_matches_total` the total number of matched requests.
```
miekg: please include in labels in `{}` in the metric name. see other READMEs, if there are non just use `{}`

plugin/template/README.md
```diff
@@ -0,0 +1,166 @@
...
+- `coredns_template_template_failures_total` the number of times the Go templating failed.
+- `coredns_template_rr_failures_total` the number of times the templated resource record was invalid and could not be parsed.
+
+Both failure cases indicate a problem with the template configuration.
```
miekg: Are these runtime errors? I.e. can we catch the template config errors on startup?

plugin/template/template.go
```diff
@@ -0,0 +1,143 @@
...
+	Group    map[string]string
+	Class    string
+	Type     string
+	Message  *dns.Msg
+	Question *dns.Question
+}
+
+// ServeDNS implements the plugin.Handler interface.
+func (h Handler) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	for _, template := range h.Templates {
```
miekg: This needs a check if you should handle the zone in the first place and if not call NextOrFailure here.

plugin/template/template.go
```diff
@@ -0,0 +1,143 @@
...
+		msg.Rcode = template.rcode
+
+		for _, answer := range template.answer {
+			rr, err := executeRRTemplate("answer", answer, data)
+			if err != nil {
+				return dns.RcodeServerFailure, err
+			}
+			msg.Answer = append(msg.Answer, rr)
+		}
+		for _, additional := range template.additional {
```
miekg, line 71: Why are we excluding the authority section? I mean that's fine, but it should be documented?

plugin/template/template.go
```diff
@@ -0,0 +1,143 @@
...
+// Name implements the plugin.Handler interface.
+func (h Handler) Name() string {
+	return "template"
+}
+
+func executeRRTemplate(section string, template *gotmpl.Template, data templateData) (dns.RR, error) {
+	buffer := &bytes.Buffer{}
+	err := template.Execute(buffer, data)
+	if err != nil {
+		TemplateFailureCount.WithLabelValues(data.Regex, section, template.Tree.Root.String()).Inc()
```
miekg, line 102: Slightly nice to let the caller increment the failure counter? I.e similar to not log errors in a function, but just returning the error?

plugin/template/template.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	rr, err := dns.NewRR(buffer.String())
+	if err != nil {
+		TemplateRRFailureCount.WithLabelValues(data.Regex, section, template.Tree.Root.String()).Inc()
+		return rr, err
+	}
+	return rr, nil
+}
+
+func (t template) match(r *dns.Msg) (templateData, bool) {
```
miekg: I think is nicer and more inline with other template to create a `state := request.Request(...)` and use that throughout. Give you some small helper functions that speeds up things, no need to keep lowercasing the question name over and over.

miekg (changes requested):
Bunch of comments, but getting close to merging this.

plugin/template/README.md
rtreffer: Yes, even if the template is syntactical correct it's hard to tell if
1. it works with the input
2. the output is a valid RR

Regex and template syntax errors will break startup, though.

plugin/template/template.go
rtreffer: Well, I'm unsure what all side effects of that are.

1. Do we need flag setting in that case, too?
2. Should we allow anything in the authority section?

I would wait for a use case on that, I don't have one so I'm unsure about it.

plugin/template/template.go
rtreffer: This function combines 2 steps and their error handling.

Unrolling the error in the caller is close to inlining this function in the first place. Happy to do that if it makes sense.

plugin/template/template.go
miekg: I don't know; it up to the user; if you write crap in there... garbage in garbage out? You can shoot yourself in the foot already. We can postpone adding this in a follow up PR.

plugin/template/template.go
miekg: nah, leave it for this then.

plugin/template/template.go
rtreffer: Nah, in that case it's easy, I'll add it.

plugin/template/template.go
rtreffer: Where would the plugin get a zone? It handles regex only (for now)

plugin/template/template.go
rtreffer: authority section pushed.

plugin/template/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+The *template* plugin allows you to dynamically repond to queries by just writing a (Go) template.
+
+## Syntax
+
+~~~
+template CLASS TYPE [REGEX...] {
+    [answer RR]
+    [answer RR]
+    [additional RR]
```
miekg: this needs authortity as well now.

plugin/template/README.md
miekg: please do this before we can merge.

plugin/template/template.go
miekg:
```
	zone := plugin.Zones(e.Zones).Matches(state.Name())
	if zone == "" {
		return plugin.NextOrFailure(e.Name(), e.Next, ctx, w, r)
	}
```
as copied from the etcd plugin (there is more code there, take a look)

rtreffer: Yes, I saw that, but there is no Zone for this plugin: it acts on regex rather than zone which is a more sophisticated match. There is no `e.Zones` equivalent right now.

If nothing matches `return plugin.NextOrFailure()` is done. If a regex matches the answer will be generated (or a servfail if that causes an error).

rtreffer: So the question is mainly: should this plugin actually act on zones and use a zone scope? Or is "try to match everything (the regex can limit the zone anyway)"?

Or should this be changed to act on zones and move the regex into a directive?

plugin/template/README.md
rtreffer: fixed

plugin/template/README.md
rtreffer: fixed

plugin/template/template.go
rtreffer: Oh no, I was missing the ToLower piece completely. The input is now a a request.Request and state.Name() is used to retrieve the qname.

rtreffer:
@miekg next round of fixes.

The only thing I'm still missing is the zone limiting. The plugin currently operates on `regex` as limiting factor and not any particular zone (right now `template <regex>` instead of `template <zone> { regex <regex>`). This is more powerful but not in line with other plugins. (Matching `something.example.$` does set a zone). Fallthrough is currently always on. This is useful to template e.g. ip templates under a shared domain.

1. I could explain/highlight the current behavior in the documentation
1. I could externalize the main body after a match, making the regex matching / fallback logic obvious to read
1. I could change the plugin to be bound to zone(s) and add a configurable fallthrough behavior (this is the same as a match-all NXDOMAIN regex)

^ @miekg thoughts? 3.?

johnbelamaric:
You can limit it by zone, and make the regex a directive. That way it does a fast filter out of the irrelevant queries. You can always make the zone "." if you want it to check every query.

miekg:
+1 to what @johnbelamaric said.

Also we've decided to make `fallthrough` always explicit (in the other plugins)

miekg:
> I could change the plugin to be bound to zone(s) and add a configurable fallthrough behavior (this is the same as a match-all NXDOMAIN regex)
I think this is a sensible thing to do.

Bit tricky to fiddle that in the syntax as it should be optional (then default to the server's block zones):
`template [ZONES] CLASS TYPE [REGEX...] {`

Should work by checking for IN/CH/ANY as a class, which you then can't have as zonenames, by meh,

plugin/template/setup.go
```diff
@@ -0,0 +1,143 @@
...
+func templateParse(c *caddy.Controller) (templates []template, err error) {
+	templates = make([]template, 0)
+
+	for c.Next() {
+		t := template{}
+		if !c.NextArg() {
+			return nil, c.ArgErr()
+		}
+
+		class, ok := dns.StringToClass[c.Val()]
```
gibson042, line 45: As noted in https://github.com/miekg/dns/issues/577#issuecomment-346235028 , "ANY" is not a real class mnemonic. This should be prepared for its removal, e.g.
```go
if c.Val() == "ANY" {
	t.class = dns.ClassANY
} else {
	class, ok := dns.StringToClass[c.Val()]
	if !ok {
		return nil, c.Errf("invalid query class %s", c.Val())
	}
	t.class = class
}
```

rtreffer: The linked page lists ANY: `255 QCLASS * (ANY)`
https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-2

ANY is not ALL, so you are free to respond with nothing.

rtreffer: Ah, removing from the list, yeah, that's fine, we just have to handle the ordering of these 2 PRs

rtreffer:
@miekg I would in that case rather move the regex into a directive. Perhaps as "match"....

```
template example. {
  match IN A "....."
  anwser .....
}
```

Does that look reasonable?

miekg:
>gibson042 commented on this pull request.
>
>
>
>> +		return Handler{Next: next, Templates: templates}
>+	})
>+
>+	return nil
>+}
>+
>+func templateParse(c *caddy.Controller) (templates []template, err error) {
>+	templates = make([]template, 0)
>+
>+	for c.Next() {
>+		t := template{}
>+		if !c.NextArg() {
>+			return nil, c.ArgErr()
>+		}
>+
>+		class, ok := dns.StringToClass[c.Val()]
>
>As noted in https://github.com/miekg/dns/issues/577#issuecomment-346235028 , "ANY" is not a real class mnemonic. This should be prepared for its removal, e.g.
>```go
>if c.Val() == "ANY" {
>	t.class = dns.ClassANY
>} else {
>	class, ok := dns.StringToClass[c.Val()]
>	if !ok {
>		return nil, c.Errf("invalid query class %s", c.Val())
>	}
>	t.class = class
>}
>```

I'm not sure we're going to do that though, so it's fine to leave this as is.

plugin/template/setup.go
gibson042: btw, ANY is also an invalid RTYPE mnemonic so you'll want similar code for that.

gibson042:
> I'm not sure we're going to do that though, so it's fine to leave this as is.

It relies upon a bug in the base library that is trivial to avoid... why would you want to keep code that will break if the bug is fixed?

plugin/template/setup.go
rtreffer: It is valid for querying and matching. It's '*' or 'ANY'.
It is not valid in DNS responses, **but these values here are never used in RRs, they are used for matching**.

gibson042: I feel like you're misunderstanding me... ANY is not valid for either class or type in an RFC 1035 master file, even though the values are assigned and CLASS255 or TYPE255 _would_ be valid (if nonsensical). You're relying on an entry in the `StringToClass` map that doesn't belong there, and you should stop doing so in order to avoid breaking if the bug in the base library is fixed.

rtreffer: Yes, I totally understand that, but they are not used in the context of 1035 here. They are not even used in the context of an RR here.

They are used to *match incoming types*. The same way you can ask a server to return *something* by sending an ANY query.

gibson042: I'm not telling you to drop "ANY" from your matchers, I'm telling you to stop assuming that "ANY" will be in `StringToClass` (which _is_ used for parsing presentation format).

rtreffer: Which is your suggestion in another PR, I know that.

But I could also invert that: Why not fix the parsing to only use mnemonic of that context?

The documents you pull as reference call it ANY. It is commonly referred to as ANY. It is just not valid in a certain, different, context.

Anyway, fixing this either way is trivial. I don't see any value in the debate here.

miekg:
I propose merging this and iterating in a couple of new PRs.
(I also want to release 1.0.3 and have this plugin hidden until we call it final, this means not putting it in plugin.cfg for a short while)

rtreffer:
@miekg what's your planned timeline for 1.0.3? Today? Tomorrow?

miekg:
either of those :)

Also I can't read the PR anymore because of the stupid way github does comments

rtreffer:
So the only objection I would have is that this means the syntax would change afterwards again (to add zones)

miekg:
mostly I would guess. I'll also take a closer look and either file (smallish) bugs and/or PRs

miekg (approved):

PR #1304, opened 2017-12-12
Adding additional records for test cases

bradbeam:
Adds some additional test cases for k8s handler. This should now encompass all records that would be generated by a zone transfer.

### DIFF
--- a/plugin/kubernetes/handler_test.go
+++ b/plugin/kubernetes/handler_test.go
@@ -85,6 +85,22 @@ var dnsTestCases = []test.Case{
 			test.A("hdls1.testns.svc.cluster.local.	5	IN	A	172.0.0.3"),
 		},
 	},
+	// A pod ip
+	{
+		Qname: "172-0-0-2.hdls1.testns.svc.cluster.local.", Qtype: dns.TypeA,
+		Rcode: dns.RcodeSuccess,
+		Answer: []dns.RR{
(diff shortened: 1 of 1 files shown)


plugin/kubernetes/handler_test.go
```diff
@@ -85,6 +85,22 @@ var dnsTestCases = []test.Case{
 			test.A("hdls1.testns.svc.cluster.local.	5	IN	A	172.0.0.3"),
 		},
 	},
+	// A pod ip
+	{
+		Qname: "172-0-0-2.hdls1.testns.svc.cluster.local.", Qtype: dns.TypeA,
```
miekg, line 90: I wouldn't put it here and I think actually that this is in the wrong namespace, i.e. this should return anything.

bradbeam: What are you thinking then? They are returned as part of the `Qname: "_http._tcp.hdls1.testns.svc.cluster.local.", Qtype: dns.TypeSRV, ` query. 
I was trying to validate the xfr by iterating through the success test cases --  https://github.com/coredns/coredns/pull/1259/commits/d8763b8d572bc3db3b698bbf956297756d30a25c#diff-81ea786d0ec7d3b64b1433739a22a1b8R55

johnbelamaric: @miekg any update here, would like to get this merged so we can move forward with the XFR

miekg:
If this looks good to you, please go ahead and merge.

PR #1306, opened 2017-12-12
Instead of hardcoding plugin lists in autopath/health, use interfaces.

### LINKED ISSUE #1291: Check for all plugins being healthers not just hardcoded list
Instead of hard-coding a list of healthers (https://github.com/coredns/coredns/blob/master/plugin/health/healther.go#L41), can we instead just loop over the `registry` and cast them all to Healther to see if they implement it? We would need a new external method called `Handlers()` (or similar) instead of just `Handler()`.

Reasoning: We implement our own plugins and currently we can do that by just adding lines to `plugin.cfg` in the build process, but if our plugins want to expose health information, we'd have to also patch the healther.go file, and the less patching we have to do, the better.

I'm willing to write a PR to do this, but wanted to make an issue first to discuss.

jameshartig:
Switched health and autopath plugin to allow any plugins to be used instead
of a hardcoded list. I did not switch federation over since it wasn't
obvious that anything other than kubernetes could be used with it.

Fixes #1291

### DIFF
--- a/core/dnsserver/register.go
+++ b/core/dnsserver/register.go
@@ -167,6 +167,21 @@ func (c *Config) Handler(name string) plugin.Handler {
 	return nil
 }
 
+// Handlers returns a slice of plugins that have been registered. This can be used to
+// inspect and interact with registered plugins but cannot be used to remove or add plugins.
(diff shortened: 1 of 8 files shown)


miekg (approved):
n/a (mistyped something here)

Issue #1307, opened 2017-12-13
rewrite: continue/stop not working for all rule types


johnbelamaric:
See https://github.com/coredns/coredns/pull/1297#discussion_r156476494

PR #1308, opened 2017-12-13
plugin/rewrite: fix flow control logic for all rule types

greenpau:
Resolves: #1307

Fixes flow control logic, i.e. stop and continue, for class and type rewrite rules.

#1307 

N/A

### DIFF
--- a/plugin/rewrite/class.go
+++ b/plugin/rewrite/class.go
@@ -8,19 +8,22 @@ import (
 )
 
 type classRule struct {
-	fromClass, toClass uint16
+	fromClass  uint16
+	toClass    uint16
+	NextAction string
 }
 
-func newClassRule(fromS, toS string) (Rule, error) {
+// newClassRule creates a class matching rule
+func newClassRule(nextAction string, args ...string) (Rule, error) {
 	var from, to uint16
 	var ok bool
-	if from, ok = dns.StringToClass[strings.ToUpper(fromS)]; !ok {
-		return nil, fmt.Errorf("invalid class %q", strings.ToUpper(fromS))
+	if from, ok = dns.StringToClass[strings.ToUpper(args[0])]; !ok {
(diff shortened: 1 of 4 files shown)


greenpau:
@johnbelamaric , please review.

johnbelamaric:
Thanks, nice. Would you mind adding tests for:

- these without stop/continue
- a test that shows stopping/continuing for each

greenpau:
> these without stop/continue

@johnbelamaric , what do you mean?

greenpau:
> a test that shows stopping/continuing for each

@johnbelamaric , 👍

johnbelamaric:
I mean that we shouldn't have to specify stop or continue, it should default to stop. Maybe there is already a test for that.

greenpau:
> I mean that we shouldn't have to specify stop or continue, it should default to stop.

@johnbelamaric, If you are talking about rule creation, then, yes. The “stop” is the default.

If you are talking about rewrite test, then stop and continue are the must, because that is how the function is getting invoked.

Can you please point me to the sections of _test code where you see an opportunity to do testing. Line numbers :-)

johnbelamaric:
Ok, looks like creation is good. My other point was that we must not have had tests for the stop/continue behavior since it wasn't working. So, adding some "continue" rules in here:

https://github.com/greenpau/coredns/blob/01a9ed501e275ac7fe82222b5c0ef987b54c8269/plugin/rewrite/rewrite_test.go#L152

and having some tests that have multiple rules applied.

greenpau:
> My other point was that we must not have had tests for the stop/continue behavior since it wasn't working.

@johnbelamaric , 👍 will add the tests 🔢

greenpau:
@johnbelamaric , please see if I captured your intent with the tests I added.

johnbelamaric:
Awesome, thanks.

PR #1309, opened 2017-12-13
plugin/metrics: Added NewMetrics func

jameshartig:
If external plugins wanted to extend metrics there was no way since
zoneNames couldn't be initialized. Now plugins can call NewMetrics to
get an instance of Metrics that they can extend.

We need to extend the metrics plugin to integrate it with our default metrics and to include automatic metrics that our other plugins generate. As in the commit message, there's no way to do this since zoneNames is private. By making a `NewMetrics` method, we can call that and then use the `*Metrics` instance.

### DIFF
--- a/plugin/metrics/metrics.go
+++ b/plugin/metrics/metrics.go
@@ -36,6 +36,11 @@ type Metrics struct {
 	zoneMu    sync.RWMutex
 }
 
+// New returns a new instance of Metrics with the given address
+func New(addr string) *Metrics {
+	return &Metrics{Addr: addr, zoneMap: make(map[string]bool)}
+}
+
 // AddZone adds zone z to m.
 func (m *Metrics) AddZone(z string) {
 	m.zoneMu.Lock()
--- a/plugin/metrics/metrics_test.go
+++ b/plugin/metrics/metrics_test.go
@@ -13,7 +13,7 @@ import (
 )
 
 func TestMetrics(t *testing.T) {
-	met := &Metrics{Addr: "localhost:0", zoneMap: make(map[string]bool)}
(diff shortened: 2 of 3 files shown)


jameshartig:
Looks like the travis build failed because of:
> fatal: unable to access 'https://go.googlesource.com/text/': Failed to connect to go.googlesource.com port 443: Connection timed out

plugin/metrics/metrics.go
```diff
@@ -36,6 +36,11 @@ type Metrics struct {
 	zoneMu    sync.RWMutex
 }
 
+// NewMetrics returns a new instance of Metrics with the given address
+func NewMetrics(addr string) *Metrics {
```
miekg: Can be just `New`?

miekg (approved):
Just "New" as a name is cleaner - other lgtm.

I kicked travis.

jameshartig:
Thanks! Just pushed an update to rename to `New` and then I'll merge.

miekg:
Go ahead :)

miekg:
This *seems* (not 100% sure yet), have broken the ablity to just add metrics:
In the *forward* plugin no metrics where exported. And I think all other plugin metrics are broken as well (didn't double check).

For the forward plugin I'm now hitting:
```
		have Collect(chan<- "github.com/prometheus/client_golang/prometheus".Metric)
		want Collect(chan<- "github.com/coredns/coredns/vendor/github.com/prometheus/client_golang/prometheus".Metric)
```

How are you extending the number of metrics?

miekg:
Sorry wrong PR :/ This is Ok.
Looking back further.

Issue #1279, opened 2017-12-05
Rewrite: Move from Caddy plugin implementation to the std plugin pkg


mxschmitt:
Hello,

I'd like to discuss a bit about the `Caddy` plugin implementation in CoreDNS and the replacement of it to the std [plugin](https://golang.org/pkg/plugin/) package which was introduced in Go 1.8.
Generally, I like to use a plugin architecture with a given structure and wide usage like in Caddy. But I think that it can be easier and cleaner to move to the new plugin engine.

Pros:
- Easier plugin plugin and play system: there's no need to compile always completely new from scratch
- Cleaner code: there is no need to import as _ package which is in my opinion really dirty

Cons:
- It's a beta feature
> The plugin support is currently incomplete, only supports Linux, and has known bugs. Please report any issues. [Source](https://golang.org/pkg/plugin/)
- No windows support: Not all architectures are yet supported

In general, there isn't it so easy to switch from one to another plugin engine instantly. We need to support the old method additional some time span and then remove it.

Thank you for your ideas and minds about this topic.

miekg:
I agree the recompiling to add a plugin is somewhat cumbersome. I rather not go after this alone, but wait for Caddy to make a move and follow that.

yongtang:
See some discussion in Caddy:
https://caddy.community/t/go-1-8-plugin-model/934/30

miekg:
closing here, 'cause nothing will change in the foreseeable future.

PR #1311, opened 2017-12-14
[WIP] CoreDNS distributed plugin with NSID

### LINKED ISSUE #1295: CoreDNS distributed plugin with NSID
In a cloud environment when combined with auto-scaling groups (e.g., in AWS), there might be a need to assign each machine with a domain name that depends on its functionality. For example, let's say there are 5 machines in an auto-scaling groups and they are expected be assigned with domain name of:
```
worker1.tf.local. -> 172.28.10.1
worker2.tf.local. -> 172.28.10.5
worker3.tf.local. -> 172.28.10.9
ps1.tf.local.     -> 172.28.10.15
ps2.tf.local.     -> 172.28.10.16
```

It might be possible to achieve the above scenario in CoreDNS:
1. Each node announce the domain it want to use with CoreDNS' NSID plugin, e.g., node 1 (172.28.10.1) have NSID of `worker1.tf.local.` exposed, node 2 (172.28.10.5) have NSID of `worker2.tf.local.` exposed.
2. Each node tries to talk to every IP address within the subnet of `172.28.10.0/24`, and find out the NSID (if available) of all  the rest of the nodes. The list of the domains (and its IP) will be stored in memory.
3. Then each node will serve the DNS queries if the queried domain is in the above list in memory.
4. Each node can also update the domain name list (and IP) with a TTL.

Note in cloud like AWS duplication of domain names (from multiple nodes) could be avoided through information exposed in AWS API. Any even if there is a collision, the CoreDNS could simply publish the domain name with multiple A record.

This could be a useful feature where external storage or external k-v store (like zookeeper or etcd) is undesired, as they require additional maintenance efforts.

yongtang:
This fix tries to implement a CoreDNS distributed in a cluster, and each CoreDNS node could return the domain names specified by its peers (through NSID).

The basic idea is that each node will use NSID to publish its domain name. The cluster is specified through the list of IPs or the list of subnets. When every node is up, it will send NSID queries to every IP in the list, and receive the domain names by response. If the domain name is within the specified domain, then each node will expose to the outside. This is more of an attempt to assign each node within a cluster a unique name in a distributed fashion, without relying on external storages or K-V stores.

This fix fixes #1295

To be updated.

### DIFF
--- a/core/dnsserver/zdirectives.go
+++ b/core/dnsserver/zdirectives.go
@@ -39,6 +39,7 @@ var directives = []string{
 	"etcd",
 	"proxy",
 	"erratic",
+	"distributed",
 	"whoami",
 	"startup",
 	"shutdown",
--- a/core/zplugin.go
+++ b/core/zplugin.go
@@ -10,6 +10,7 @@ import (
 	_ "github.com/coredns/coredns/plugin/cache"
 	_ "github.com/coredns/coredns/plugin/chaos"
 	_ "github.com/coredns/coredns/plugin/debug"
+	_ "github.com/coredns/coredns/plugin/distributed"
 	_ "github.com/coredns/coredns/plugin/dnssec"
 	_ "github.com/coredns/coredns/plugin/dnstap"
 	_ "github.com/coredns/coredns/plugin/erratic"
--- a/plugin.cfg
+++ b/plugin.cfg
@@ -47,6 +47,7 @@ secondary:secondary
 etcd:etcd
 proxy:proxy
 erratic:erratic
+distributed:distributed
 whoami:whoami
 startup:github.com/mholt/caddy/startupshutdown
 shutdown:github.com/mholt/caddy/startupshutdown
--- a/plugin/distributed/distributed.go
+++ b/plugin/distributed/distributed.go
@@ -1,0 +1,218 @@
+// Package distributed implements a distributed CoreDNS with NSID
+package distributed
+
+import (
+	"encoding/hex"
+	"fmt"
+	"log"
+	"net"
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/request"
+
+	"github.com/miekg/dns"
+	"golang.org/x/net/context"
+)
+
+type endpoint struct {
+	addr string
+	port string
+	quit chan struct{}
+}
+
+type entries struct {
+	sync.RWMutex
+	byAddr map[string]string
+	byName map[string][]net.IP
+}
+
+// Distributed plugin
+type Distributed struct {
+	Next      plugin.Handler
+	Origin    string
+	Endpoints *[]endpoint
+	Entries   *entries
+}
+
+func (e *endpoint) update(d *Distributed) error {
+	name, _ := e.query(d)
+
+	d.Entries.Lock()
+	defer d.Entries.Unlock()
+
+	// Only update when data is dirty
+	dirty := false
+	if name == "" {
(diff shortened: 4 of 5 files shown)


miekg:
We could/should keep this seperate for the time being; i.e. create coredns/distrubuted and put it on coredns.io/explugins and then graduate it at some point. Just like I'm planning with coredns/forward.

yongtang:
@miekg That makes sense. Let me update and move to coredns/distributed.

johnbelamaric:
Instead of a separate repo for every incubated plugin, I think an incubator repository with directories for each external plugin will be cleaner.

yongtang:
One thing I would like to raise is how to integrate with external plugin. Ideally we want to either

1. distribute one coredns binary with all external plugins prebuilt by us (but not enabled by default), *OR*
2. distribute one coredns binary with the option of dynamically load an external plugin built by user.

There are some discussions about external plugin (#1279), but from
https://caddy.community/t/go-1-8-plugin-model/934?page=2
it looks like nothing will happen in the foreseeable future.

I think for most of the users, the issue is that how to use external plugin without download a coredns not built by default coredns repo.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [WIP] CoreDNS..." ]
>One thing I would like to raise is how to integrate with external plugin. Ideally we want to either
>
>1. distribute one coredns binary with all external plugins prebuilt by us (but not enabled by default), *OR*
>2. distribute one coredns binary with the option of dynamically load an external plugin built by user.
>
>There are some discussions about external plugin (#1279), but from
>https://caddy.community/t/go-1-8-plugin-model/934?page=2
>it looks like nothing will happen in the foreseeable future.
>
>I think for most of the users, the issue is that how to use external plugin without download a coredns not built by default coredns repo.

Caddy uses a build server where you can click things together and then get a 
binary to download. We of course also have docker....

I don't care *that* much if we just bundle them by default (there is little sign 
for that on the explugins page). It mostly maintence that I'm concerned about.

yongtang:
If we bundle external plugins I would prefer to have each plugins in different repos. Those repos could be scattered  around in github and maintained by different parties or orgs. 

We could build them when the coredns binary is released.

I think we area already doing that, e.g., in plugin.cfg: 
```
startup:github.com/mholt/caddy/startupshutdown
shutdown:github.com/mholt/caddy/startupshutdown
```
are not maintained by CoreDNS though it is in the binary.

miekg:
@yongtang agreed. Let's put this in a new repo as well.

Yes we already default to including external plugin.

johnbelamaric:
I am just saying putting our external plugins in an incubator repo. External plugins can still be pulled from different repos anywhere. This was really just an organizational issue to not proliferate a lot of repos in the coredns org.

miekg:
Note that there is also political angle: if we can default to include some external plugins, why not all? Where do we draw the line?

yongtang:
@miekg I would suggest to have a "graduation" process so that if an external plugin is
1. A stable codebase in GitHub
2. Well maintained by a third party or org

Then the author of the external could just create a pull request to add it to `External Plugins` page (or a `Bundled Plugins` page) :
 https://coredns.io/explugins/

And I would suggest we could add them in the binary.

yongtang:
@miekg by the way I think the "graduation" process could either be initialized by a coredns maintainer, or an external plugin author. They could just create a pull request for the process and discussion.

miekg:
+1 to that, note the current process of getting on explugin is just "send pr"

But I'm all onboard to have some kind of vetting process

miekg:
I think that's fair, having a second process for "default inclusion"

yongtang:
Also, let me close this PR for now.

PR #1312, opened 2017-12-14
plugin/metrics: Switch to using promhttp instead of deprecated Handler

jameshartig:
prometheus.Handler is deprecated according to the godoc for the package so
instead we're using promhttp.

Additionally, we are exposing the Registry that metrics is using so other
plugins that are not inside of coredns can read the registry. Otherwise, if
we kept using the Default one, there's no way to access that from outside
of the coredns repo since it is vendored.

I had to vendor `promhttp` which seemed to update thousands of files.


miekg:
Can you split the vendor and prom change? (I'll try to find the actual code change in the mean time)

miekg:
ah, github deals pretty well with this. nice

miekg:
I can't add external metrics anymore due to vendoring. How are you doing it?

miekg:
hmm, of course the local go build fails, but it works when you add it to coredns and then compile; which is the only useful way to compile a plugin. Ok fixed problem #1. Still remains the fact the plugin metrics (those that are included in coredns by default) don't export metrics any more

miekg:
Nope :(
~~~
# github.com/coredns/forward
../forward/metrics.go:59:22: cannot use RequestCount (type *"github.com/prometheus/client_golang/prometheus".CounterVec) as type "github.com/coredns/coredns/vendor/github.com/prometheus/client_golang/prometheus".Collector in argument to m.Reg.MustRegister:
	*"github.com/prometheus/client_golang/prometheus".CounterVec does not implement "github.com/coredns/coredns/vendor/github.com/prometheus/client_golang/prometheus".Collector (wrong type for Collect method)
		have Collect(chan<- "github.com/prometheus/client_golang/prometheus".Metric)
		want Collect(chan<- "github.com/coredns/coredns/vendor/github.com/prometheus/client_golang/prometheus".Metric)
../forward/metrics.go:60:22: cannot use RequestDuration (type *"github.com/prometheus/client_golang/prometheus".HistogramVec) as type "github.com/coredns/coredns/vendor/github.com/prometheus/client_golang/prometheus".Collector in argument to m.Reg.MustRegister:
	*"github.com/prometheus/client_golang/prometheus".HistogramVec does not implement "github.com/coredns/coredns/vendor/github.com/prometheus/client_golang/prometheus".Collector (wrong type for Collect method)
		have Collect(chan<- "github.com/prometheus/client_golang/prometheus".Metric)
		want Collect(chan<- "github.com/coredns/coredns/vendor/github.com/prometheus/client_golang/prometheus".Metric)
../forward/metrics.go:61:22: cannot use HealthcheckFailureCount (type *"github.com/prometheus/client_golang/prometheus".CounterVec) as type "github.com/coredns/coredns/vendor/github.com/prometheus/client_golang/prometheus".Collector in argument to m.Reg.MustRegister:
	*"github.com/prometheus/client_golang/prometheus".CounterVec does not implement "github.com/coredns/coredns/vendor/github.com/prometheus/client_golang/prometheus".Collector (wrong type for Collect method)
		have Collect(chan<- "github.com/prometheus/client_golang/prometheus".Metric)
		want Collect(chan<- "github.com/coredns/coredns/vendor/github.com/prometheus/client_golang/prometheus".Metric)
../forward/metrics.go:62:22: cannot use SocketGauge (type *"github.com/prometheus/client_golang/prometheus".GaugeVec) as type "github.com/coredns/coredns/vendor/github.com/prometheus/client_golang/prometheus".Collector in argument to m.Reg.MustRegister:
	*"github.com/prometheus/client_golang/prometheus".GaugeVec does not implement "github.com/coredns/coredns/vendor/github.com/prometheus/client_golang/prometheus".Collector (wrong type for Collect method)
		have Collect(chan<- "github.com/prometheus/client_golang/prometheus".Metric)
		want Collect(chan<- "github.com/coredns/coredns/vendor/github.com/prometheus/client_golang/prometheus".Metric)
~~~

jameshartig:
I'm not near my laptop but here's a gist of the handler from my phone: https://gist.github.com/fastest963/924a38a47f99b9c561cf5863e6b9019a

It isnt great but it's working for us. I haven't had time yet to work out a better solution.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/metric..." ]
>I'm not near my laptop but here's a gist of the handler from my phone: https://gist.github.com/fastest963/924a38a47f99b9c561cf5863e6b9019a
>
>It isnt great but it's working for us. I haven't had time yet to work out a better solution.

Why didn't the defaultRegistry work for you?

(Mind you I can't get the metrics from the forward plugin to show up at all 
anymore, so I'm pondering my own sanity)

jameshartig:
The default registry in the prometheus package didn't work because it's a different package (the vendored one) than our packages (outside of coredns) could publish to. Our plugin imports packages inside our internal stdlib that have their own metrics and we couldn't get them to use the same default registry.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/metric..." ]
>The default registry in the prometheus package didn't work because it's a different package (the vendored one) than our packages (outside of coredns) could publish to. Our plugin imports packages inside our internal stdlib that have their own metrics and we couldn't get them to use the same default registry.

See angry tweets. I kicked prometheus out of the vendor dir because of this. At 
least means using and registring metrics work again.

jameshartig:
Looks like now that you removed prometheus vendoring we might be able to revisit this on my end. To come up with something cleaner.

PR #1315, opened 2017-12-15
Expose directives in dnsserver to help external plugin developers

yongtang:

This fix expose directives in dnsserver package, so that external
plugin developers could easily build customerized coredns+plugin
without changing the code base tree of coredns.

The following is an example that could bundle coredns+example,
in one simple file without modifying coredns codebase.
With changes in this PR, an external plugin developer
could just import the plugin it want (like `_ "github.com/coredns/example"`), 
and update the `dnsserver.Directives`,
then invoking `go build -v sample.go` below will be enough

```
package main

import (
        _ "github.com/coredns/example"

        "github.com/coredns/coredns/coremain"
        "github.com/coredns/coredns/core/dnsserver"
)

var directives = []string{
        "example",
        "log",
        "errors",
        ...
        ...
        ...
        "whoami",
        "startup",
        "shutdown",
}

func init() {
        dnsserver.Directives = directives
}

func main() {
        coremain.Run()
}
```

N/A

N/A

### DIFF
--- a/core/dnsserver/register.go
+++ b/core/dnsserver/register.go
@@ -22,7 +22,7 @@ func init() {
 	flag.StringVar(&Port, serverType+".port", DefaultPort, "Default port")
 
 	caddy.RegisterServerType(serverType, caddy.ServerType{
-		Directives: func() []string { return directives },
+		Directives: func() []string { return Directives },
 		DefaultInput: func() caddy.Input {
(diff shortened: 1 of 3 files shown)


miekg:
Should also update coredns.io about includig external plugins?

yongtang:
@miekg Sure. Let me take a look to see where to add it.

Issue #1313, opened 2017-12-14
feature: plugin/rewrite: rewrite ANSWER SECTION


greenpau:
Consider a DNS request for `ftp-us-west-1.coredns.rocks`. DNS server is `coredns`.

```
$ dig @10.1.1.1 ftp-us-west-1.coredns.rocks

; <<>> DiG 9.8.3-P1 <<>> @10.1.1.1 ftp-us-west-1.coredns.rocks
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 8619
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;ftp-us-west-1.coredns.rocks. IN A

;; ANSWER SECTION:
ftp.service.us-west-1.consul. 0	IN A	10.10.10.10
ftp.service.us-west-1.consul. 0	IN A	10.20.20.20
ftp.service.us-west-1.consul. 0	IN A	10.30.30.30
```

The rule that made it happen is:

```
rewrite name regex (.*)-(us-west-1)\.coredns\.rocks {1}.service.{2}.consul
```

The `ANSWER SECTION` contains entries for `ftp.service.us-west-1.consul.`

An upstream DNS server may consider the mismatch between `QUESTION`  and `ANSWER` as a MITM attack and would not return the answer:

```
;; ANSWER SECTION:
ftp.service.us-west-1.consul. IN A
```

I propose extended the rule engine for response re-writing (but only when there is a match in a first place).

```
rewrite name regex REQ_MATCH REQ_REPLACE RESP_MATCH RESP_REPLACE
rewrite name regex (.*)-(us-west-1)\.coredns\.rocks {1}.service.{2}.consul (.*)\.service\.(.*)\.consul {1}-{2}.coredns.rocks
```

@johnbelamaric , @miekg , your thoughts?

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] feature: plugin/r..." ]
>Consider a DNS request for `ftp-us-west-1.coredns.rocks`. DNS server is `coredns`.
>
>```
>$ dig @10.1.1.1 ftp-us-west-1.coredns.rocks
>
>; <<>> DiG 9.8.3-P1 <<>> @10.1.1.1 ftp-us-west-1.coredns.rocks
>; (1 server found)
>;; global options: +cmd
[...]

PR #1318, opened 2017-12-17
feature: plugin/rewrite: rewrite ANSWER SECTION

greenpau:
Resolves: #1313

The change allows rewriting a DNS response when the response was matched by `name regex` rule.

```
.:5353 {
    rewrite stop name regex xxx.(google.com) www.{1}
    rewrite stop {
        name regex yyy.(google.com) www.{1}
        answer name www.(google.com) xxx.{1}
    }
    proxy . 8.8.8.8
    log
    errors
}
``` 

When a user queries `yyy.google.com`, the response that the user would see is `xxx.google.com`. However, the answer itself is based of `www.google.com`.

In a nutshell, the rule syntax is:

```
rewrite {
    name regex STRING STRING
    answer [name] STRING STRING
}
```

#1313

Amend documentation of `plugin/rewrite`.

### DIFF
--- a/plugin/rewrite/class.go
+++ b/plugin/rewrite/class.go
@@ -41,3 +41,8 @@ func (rule *classRule) Rewrite(w dns.ResponseWriter, r *dns.Msg) Result {
 func (rule *classRule) Mode() string {
 	return rule.NextAction
 }
+
+// GetResponseRule return a rule to rewrite the response with. Currently not implemented.
+func (rule *classRule) GetResponseRule() ResponseRule {
+	return ResponseRule{}
+}
--- a/plugin/rewrite/edns0.go
+++ b/plugin/rewrite/edns0.go
@@ -78,6 +78,11 @@ func (rule *edns0NsidRule) Mode() string {
 	return rule.mode
 }
 
+// GetResponseRule return a rule to rewrite the response with. Currently not implemented.
+func (rule *edns0NsidRule) GetResponseRule() ResponseRule {
+	return ResponseRule{}
+}
+
 // Rewrite will alter the request EDNS0 local options
 func (rule *edns0LocalRule) Rewrite(w dns.ResponseWriter, r *dns.Msg) Result {
 	result := RewriteIgnored
@@ -113,6 +118,11 @@ func (rule *edns0LocalRule) Rewrite(w dns.ResponseWriter, r *dns.Msg) Result {
 // Mode returns the processing mode
 func (rule *edns0LocalRule) Mode() string {
 	return rule.mode
+}
+
+// GetResponseRule return a rule to rewrite the response with. Currently not implemented.
+func (rule *edns0LocalRule) GetResponseRule() ResponseRule {
+	return ResponseRule{}
 }
 
 // newEdns0Rule creates an EDNS0 rule of the appropriate type based on the args
@@ -312,6 +322,11 @@ func (rule *edns0VariableRule) Mode() string {
 	return rule.mode
 }
 
+// GetResponseRule return a rule to rewrite the response with. Currently not implemented.
+func (rule *edns0VariableRule) GetResponseRule() ResponseRule {
+	return ResponseRule{}
+}
+
 func isValidVariable(variable string) bool {
 	switch variable {
 	case
@@ -421,6 +436,11 @@ func (rule *edns0SubnetRule) Rewrite(w dns.ResponseWriter, r *dns.Msg) Result {
 // Mode returns the processing mode
 func (rule *edns0SubnetRule) Mode() string {
 	return rule.mode
+}
+
+// GetResponseRule return a rule to rewrite the response with. Currently not implemented.
+func (rule *edns0SubnetRule) GetResponseRule() ResponseRule {
+	return ResponseRule{}
 }
 
 // These are all defined actions.
--- a/plugin/rewrite/name.go
+++ b/plugin/rewrite/name.go
@@ -37,6 +37,7 @@ type regexNameRule struct {
 	NextAction  string
 	Pattern     *regexp.Regexp
 	Replacement string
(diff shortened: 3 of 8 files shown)


greenpau:
@johnbelamaric , @miekg , please take a look at this PR and provide your feedback.

greenpau:
@johnbelamaric , @miekg , how would you test `ResponseReverter`? I am not finding any existing tests. 😞

johnbelamaric:
This looks pretty good. I am wondering though if we should make response re-writing a separate plugin that is independent of request rewriting, or may just a special type of rewrite (like name/type/class etc). You have it set up here so that it will only activate the response rewrite if the corresponding request rewrite was done. Is that needed? Could we just have a `rewrite response` option that engages only on the response?

The HTTP rewrite I think offers some conditional rewrites. We could consider also looking at that and seeing if we can do the conditional response rewriting with some combination of a general response rule and those conditionals.

greenpau:
> You have it set up here so that it will only activate the response rewrite if the corresponding request rewrite was done. Is that needed?

@johnbelamaric , imho, as I described in the #1313, it is absolutely necessary. In many instances the mismatch in a name between the question and answers causes not accepting the response due to it being considered MITM.

> I am wondering though if we should make response re-writing a separate plugin that is independent of request rewriting, or may just a special type of rewrite (like name/type/class etc).

Perhaps in a different PR 😄  I actually thought about it. That's why I have `answer name`. I am not sure that rewriting type and class is any useful. However. re-writing TTL might be 😃 , i.e. `answer ttl`.

miekg:
Can it be done in another plugin? That syntax is coming really unwieldy?

johnbelamaric:
You have it set up here so that it will only activate the response rewrite if the corresponding request rewrite was done. Is that needed?

@johnbelamaric<https://github.com/johnbelamaric> , imho, as I described in the #1313<https://github.com/coredns/coredns/issues/1313>, it is absolutely necessary. In many instances the mismatch in a name between the question and answers causes not accepting the response due to it being considered MITM.


What I meant is, is it necessary to ONLY do it if the original request rewrite hits? Or could we always do it? Because if we can always do it, then it is easier to do in a separate plugin or rule.

johnbelamaric:
Can it be done in another plugin? That syntax is coming really unwieldy?

That's what I was wondering too. Either a separate plugin or separate rule type.

greenpau:
@johnbelamaric, @miekg, this PR is solving one issue - the mismatch between question and answer.

Having a different rule or plugin would then force all responses to change, which means that the mismatch between Q and A will become an issue for all requests whose request were NOT rewritten.

johnbelamaric:
Ok, so the request rewrite and response rewrite need to be tied together. I am still concerned about the syntax though. Take a look at https://caddyserver.com/docs/rewrite - notice the form with the `{}`. We may want to do something like that for this complex type of rule. So it would look like:

```
.:5353 {
    rewrite name regex xxx.(google.com) www.{1}
    rewrite {
        name regex yyy.(google.com) www.{1}
        answer name www.(google.com) xxx.{1}
    }
    proxy . 8.8.8.8
    log
    errors
}
```

greenpau:
> We may want to do something like that for this complex type of rule. So it would look like:

```
    rewrite {
        name regex yyy.(google.com) www.{1}
        answer name www.(google.com) xxx.{1}
    }
```

@johnbelamaric , I like it! 👍

greenpau:
> Ok, so the request rewrite and response rewrite need to be tied together. 

@johnbelamaric , 💯 ! For example, in an enterprise setting, if you have F5 "proxy"-ing your DNS requests to `coredns`, you ***must*** rewrite the response. Otherwise, F5 would drop a response with Q&A mismatch.

masteinhauser:
@greenpau @johnbelamaric Howdy, this is impacting our new CoreDNS upgrade/deployment in Kubernetes. Is there anything I can help with to move this patch along? 

It appears from this thread the configuration syntax is the main blocker at this moment, but otherwise the plugin addition is a go-ahead.

greenpau:
> this is impacting our new CoreDNS upgrade/deployment in Kubernetes. Is there anything I can help with to move this patch along?

@masteinhauser , I will get back to it next week.

Out of curiosity, are you planning to use this feature? How does it help with k8s?

greenpau:
Amended the description of the PR https://github.com/coredns/coredns/pull/1318#issue-282700705

masteinhauser:
@greenpau Hey, thanks for the update!

Yes, we will be using it with a multi-region Kubernetes deployment architecture. Each region has a separate cluster, and a separate "root domain" that we use to address and route requests across to other clusters. However, we still need to support the `cluster.local` domain for Conformance tests.

Right now we can rewrite the Question part within coredns from `cluster.local` -> `cluster.specific-cluster-domain` but the query response results end up being ignored by most/all DNS libraries and tools we have found, thus breaking the overall desired architecture pattern.

greenpau:
> Yes, we will be using it with a multi-region Kubernetes deployment architecture. Each region has a separate cluster, and a separate "root domain" that we use to address and route requests across to other clusters. 

> but the query response results end up being ignored by most/all DNS libraries and tools we have found

@masteinhauser , thank you for the explanation! It is the same thing for Consul. I hope I will finish up this week. This syntax thing requires some work. Will have some fun tonight 😄

greenpau:
@masteinhauser , I finished the syntax work. Once Travis CI is back online and it passes, I will ask @johnbelamaric to review again.

greenpau:
@johnbelamaric, the build passed; please review and comment.

greenpau:
@miekg , could you add this one as part of #1397 ?

miekg:
1.0.4 will be cut from 1.0.3 and then recompiled against a (locally) patched miekg/dns - no new features.

PR #1320, opened 2017-12-17
WIP: Crazy idea: install certs on startup

miekg:


miekg:
@yongtang how crazy/stupid is this?

yongtang:
I think it makes sense in many situations. Though sometimes user may prefer not to start an outbound connection when they launch coredns. So maybe we could make it optional?

yongtang:
On a second thought, maybe we could ask user to mount the certs directory in the host to the directory in coredns container? Haven't tried it but I think it might be possible.

miekg:
I like your idea as this would side step of a lof problems. Let me experiment with that.

johnbelamaric:
Yes, I think that is much better. It also makes it easy to use your own CA certs.

PR #1333, opened 2017-12-27
Stop vendoring prometheus

miekg:


miekg:
Both this and #1332 are needed so that https://github.com/coredns/forward/pull/17 can be merged to fix the metrics. 3rd party plugin metrics were not working at all. I also think none of the other metrics in coredns where working. I'll double check that and possibly send another PR.

miekg:
This drops the vendoring and updates the makefile to pull the deps.

PR #1335, opened 2017-12-27
metrics: correctly register all metrics

miekg:
After initial startup, see if prometheus is loaded and if so, register
our metrics with it.
Stop doing the init() func and just use the sync.Once so we don't double
registrer our metrics.

### DIFF
--- a/plugin/autopath/metrics.go
+++ b/plugin/autopath/metrics.go
@@ -18,12 +18,4 @@ var (
 	}, []string{})
 )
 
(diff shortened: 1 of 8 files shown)


jameshartig (approved):
Woops! Sorry, didn't know I broke these plugins. Good catch. LGTM

miekg:
np. Should have some kind of test for this.

Anyway, external plugins can use the same trick: https://github.com/coredns/forward/blob/master/setup.go#L36

PR #1337, opened 2017-12-29
WIP: new plugin/allow, to block certain incoming requests

sbach:
This is a proposal to introduce a new plugin that can filter incoming requests. It returns the error code: `REFUSED` if the client's IP address isn't in one of the configured CIDR.

A use case would be to run CoreDNS into a Kubernetes cluster. If exposing the DNS service to the outside (to serve a domain's zone), it can be used to prevent public IPs to do requests to say: the Kubernetes plugin (and thus retrieve the internal IPs of the cluster's containers).

If it is interesting enough to be included into mainline, I will ensure to follow with documentation and test code.

### DIFF
--- a/plugin/allow/allow.go
+++ b/plugin/allow/allow.go
@@ -1,0 +1,40 @@
+package allow
+
+import (
+    "net"
+
+    "github.com/coredns/coredns/plugin"
+    "github.com/coredns/coredns/request"
+
+    "github.com/miekg/dns"
+    "golang.org/x/net/context"
+)
+
+type Allow struct {
+    Next     plugin.Handler
+    Cidrs  []string
+}
+
+func (rw Allow) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+    state := request.Request{W: w, Req: r}
+    ip := state.IP()
+
+    for _, it := range rw.Cidrs {
+        _, cidr, err := net.ParseCIDR(it)
+        if err != nil {
+            return 0, err
+        }
+        if cidr.Contains(net.ParseIP(ip)) {
+            return plugin.NextOrFailure(rw.Name(), rw.Next, ctx, w, r)
+        }
+    }
+
+    m := new(dns.Msg)
+    m.SetRcode(r, dns.RcodeRefused)
+    state.SizeAndDo(m)
+    w.WriteMsg(m)
+
+    return 0, nil
+}
+
+func (rw Allow) Name() string { return "allow" }
--- a/core/dnsserver/zdirectives.go
+++ b/core/dnsserver/zdirectives.go
@@ -22,6 +22,7 @@ var directives = []string{
 	"prometheus",
 	"errors",
 	"log",
+	"allow",
 	"dnstap",
 	"chaos",
 	"cache",
--- a/core/zplugin.go
+++ b/core/zplugin.go
@@ -4,6 +4,7 @@ package core
 
 import (
 	// Include all plugins.
+	_ "github.com/coredns/coredns/plugin/allow"
 	_ "github.com/coredns/coredns/plugin/auto"
 	_ "github.com/coredns/coredns/plugin/autopath"
 	_ "github.com/coredns/coredns/plugin/bind"
--- a/plugin.cfg
+++ b/plugin.cfg
@@ -50,3 +50,4 @@ erratic:erratic
 whoami:whoami
 startup:github.com/mholt/caddy/startupshutdown
 shutdown:github.com/mholt/caddy/startupshutdown
+allow:allow
--- a/plugin/allow/setup.go
+++ b/plugin/allow/setup.go
@@ -1,0 +1,43 @@
+package allow
+
+import (
+    "strings"
+
+    "github.com/coredns/coredns/core/dnsserver"
+    "github.com/coredns/coredns/plugin"
+
+    "github.com/mholt/caddy"
+)
+
+func init() {
+    caddy.RegisterPlugin("allow", caddy.Plugin{
+        ServerType: "dns",
+        Action: setup,
+    })
+}
+
+func setup(c *caddy.Controller) error {
+    cidrs, err := argsParse(c)
+    if err != nil {
+        return plugin.Error("allow", err)
+    }
+
+    dnsserver.GetConfig(c).AddPlugin(func(next plugin.Handler) plugin.Handler {
+	    return Allow{Next: next, Cidrs: cidrs}
+    })
+
+    return nil
+}
+
+func argsParse(c *caddy.Controller) ([]string, error) {
+	var cidrs []string
+
+	for c.Next() {
+		args := c.RemainingArgs()
+		for _, arg := range args {
+			cidrs = append(cidrs, strings.ToLower(arg))
+		}
+	}
+
+	return cidrs, nil
+}


miekg:
Yes, interesting. Think is small and new enough to warrant inclusion; might also lead to some kind of language of "blocking/allowing" requests, so I think it is good to have it in mainline.

johnbelamaric:
@fturib

We plan to do something like this but haven’t gotten to it yet. And I am sorry to say I am not planning on working on it from the Bahamas this week 😁

Happy New Year 🎉




On Dec 31, 2017, at 11:50 AM, Miek Gieben <notifications@github.com<mailto:notifications@github.com>> wrote:


Yes, interesting. Think is small and new enough to warrant inclusion; might also lead to some kind of language of "blocking/allowing" requests, so I think it is good to have it in mainline.

fturib:
We have in mind to have a simple ACL language in the config of the plugin to setup.
and the enforcement would be ensured by a policy engine deployed on the fly.

Would could be a first step here is to implement the ACL language thanks to this plugin (using allow/deny rules). and later think to change the interpreter to a full policy engine if more complexity is required.

fturib:
I am looking into my notes. 
I found this simple language we proposed:

block <qtype> ANY

it would be something like:

[block|allow] [qtype] [cidr]

we did not dive deeper to define the interpretation with several rules ... and what is the default (allow ANY or bloc ANY, until an explicit rule say something else ...)

we did not define neither what are the parameters for the filtering:
SourceAdr ? QType ? DestAddr ? options on the query ?

NOTE: here in the proposed ALLOW plugin, filter is on source address.

plugin/allow/allow.go
```diff
@@ -0,0 +1,40 @@
...
+func (rw Allow) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+    state := request.Request{W: w, Req: r}
+    ip := state.IP()
+
+    for _, it := range rw.Cidrs {
+        _, cidr, err := net.ParseCIDR(it)
+        if err != nil {
+            return 0, err
+        }
+        if cidr.Contains(net.ParseIP(ip)) {
```
johnbelamaric, line 27: Lines 23-26 can be done during config parse instead of inline with request handling.

miekg:
@fturib that looks indeed simple enough. `[block|allow] [qtype] [cidr]`

There is tsig as well, so I think it should be more wordy, like ipf.
~~~
block type QTYPE from CIDR key TSIG-SOMETHING 
~~~

We could split allow and block into different plugins, but may be overdoing it.

fturib:
@miekg : I agree with the format you propose. I think also it would be more easy if we extend the syntax to add more elements to the line to match the DNS request.

On the other hand, I think it should all be in the same plugin. 
Mimic the ACL of firewalls:

1- we have a list of filters we evaluate from top to bottom
2- when one filter match we apply immediately the corresponding action (allow|deny) - either REFUSED if denied, either process the query (in fact forward to evaluation to next plugin).

NOTE: we could also use the same keywords : [allow|deny] (instead of [allow|block]).
ANY would be for any value in the corresponding field.


eg, if you wish to allow only A and AAAA type queries from 10.0.0.016, except from 10.0.1.0/24

```
deny type A from 10.0.1.0/24 
allow type A from 10.0.0.0/16
deny type AAAA from 10.0.1.0/24 
allow type AAAA from 10.0.0.0/16
deny type ANY from ANY
```

eg, if you wish to allow everything except PTR queries from 127.0.0.1

```
deny type PTR from 127.0.0.1/32
allow type ANY from ANY
```

etc...
the ACL are builds with adding small filters in top.

NOTE: if no filter is hit during evaluation then the process goes as if no plugin was there, sending the request to next plugin. it correspond to a global "allow ANY from ANY" at the end of the ACLs.

@sbach, @miekg : what do you think ?

johnbelamaric:
So in the first example, when the query matches `allow type A from 10.0.0.0/16`, we stop processing rules and allow it to go through? If they didn't have the `deny type ANY from ANY`, the `allow` statements would all be unnecessary, correct? Just double-checking I understand the logic.

johnbelamaric:
I think in the second example you meant `allow` not `deny type ANY from ANY`?

fturib:
yes :). Let me change in the comment directly for clarification. Thanks!

fturib:
@johnbelamaric : about the logic. Yes - we need the last "deny type ANY from ANY" to block all incoming queries that does not match in rule above.

miekg:
Yes, this start looking good.

Do we need an 'or' construct and a 'not' as well?

sbach:
@fturib I really like the idea to mimic firewalls behavior!

I will attempt to code what has been suggested. However, I am not that experienced in Go. Is there any other plugin with a similar configuration parsing logic I could understand first?

miekg:
The rewrite plugin comes to mind.

fturib:
@miekg  : about the logic OR, NOT .. I do not think we need to.
NOT : can be simulated by reversing the action (deny/allow)
OR : it is already implemented in the sequence of lines.
AND : could make sense on some parameters, but not on the CIDRs ...

I do not see that the logic is implemented in the firewall ACLs.

miekg:
I don't see any progress here, probably best to work on it externally and then, at some point, as for inclusion in CoreDNS. I the mean time we can list this on coredns.io/explugins.

sbach:
Apologizes for the lack of updates. @miekg, I do have some updated code following @fturib idea. I have yet to find the time to make it pretty, for an inclusion in the official repository.

For those interested: https://github.com/sbach/coredns/commit/af8cb2a4b4514546a1e3db1961a4c0b3c130c3be

ihac:
Hi @miekg 
Recently I'm applying for the GSoC project (https://github.com/cncf/soc#support-source-ip-based-query-blockallow). Since this issue was closed 1 years ago, I wonder whether you have any other plan about this feature now (e.g. new demands not mentioned in this issue, and plugin inclusion)?

chrisohaver:
#2335 is more recent.  The plan is/was to move #2335 to be an external plugin.  This has not happened yet due to other priorities, so it's in a state of invisible flux right now (it's  closed PR).  There have been a couple of GSOC inquiries into this feature in the CNCF coredns slack channel.  To summarize, there is already a proposed firewall plugin - #2335.  IIUC, There isn't much left to do in #2335, other than to package it as an external plugin, and maybe 1-2 other final tweaks.
OTOH, #2335 does a lot more than just client ip/zone rules.  There could be value in a lighter weight client-ip/zone only ACL plugin, particularly if #2335 doesn't scale well.

yongtang:
@ihac There are reasons for GSoC to focus on source IP, as it is merely a security measure to prevent attacks of the DNS server itself (not necessarily to protect the IT infrastructure that will talk to the DNS server). Amplification attack is one example. So the focus is different here.

ihac:
@chrisohaver @yongtang Thanks for the information. Seems there isn't much work left to do based on [#2335](https://github.com/coredns/coredns/pull/2335). I'm a little worried about whether the work of students could pass the final evaluation of GSoC.

yongtang:
@ihac The GSoC project has something in common with PR #2335, however, as I specified they have completely different scope. There is no requirement to build GSoC on top of #2335.

For GSoC project, depending on the final quality of the plugin, it could be either an independent repo (external plugin), or a repo within coredns org, or a default plugin inside coredns repo. This will be decided at the end. The final evaluation of GSoC is based on the quality (not quantity) of the completed work during the summer.

ihac:
@yongtang Thanks. That makes sense.

PR #1340, opened 2018-01-02
Update k8s client-go to v6.0.0

yongtang:

This fix updates k8s client-go to v6.0.0 as CoreDNS is supported in 1.9 and v6.0.0 is the recommended version.

There are quite some massive changes that need to be made:
1. `k8s.io/client-go/pkg/api/v1` has been changed to `k8s.io/api/v1` (repo changed from `client-go` to `api`)
2. kubernetes.Clientset adds one extra layer, so that `kubernetes.Clientset.Services()` and like has been changed to `kubernetes.Clientset.CoreV1().Services()` (and similarly for others)

Also, we have to stick with specific commits of `k8s.io/apimachinery` and the newly introduced `k8s.io/api` because `go dep` still could not figure out the right version to fetch.

See https://github.com/kubernetes/kubernetes/pull/53651

N/A


yongtang:
@chrisohaver Was trying to invoke the integration test but failed. The log is not very informative. Is there anyway to see a verbose output?

chrisohaver:
@yongtang the information is in the log, it's just in an odd order, due to the way many commands are chained together in one command in the make file ... It looks like the local docker repo container failed to start, which caused the subsequent image push to fail.  I saw this happen once before... docker gets confused, and cant start containers.  I'll restart docker...  that fixed it last time... hopefully will fix again.

chrisohaver:
restarted docker on drone. re-running /integration

chrisohaver:
That worked - if we keep seeing this, we can add a docker restart to the script.

yongtang:
@chrisohaver Thanks 👍

Issue #1331, opened 2017-12-26
plugin/kubernetes: should we answer this?


miekg:
I have a service called coredns:
~~~
% kubectl get svc coredns
NAME      TYPE           CLUSTER-IP      EXTERNAL-IP     PORT(S)        AGE
coredns   LoadBalancer   10.104.110.40   192.168.1.252   53:32377/UDP   9m
~~~

Now this returns something:
```
% dig svc coredns.pod.cluster.local

;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 25762
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 4, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;coredns.pod.cluster.local.	IN	A

;; ANSWER SECTION:
coredns.pod.cluster.local. 5	IN	CNAME	*.
*.			5	IN	A	10.96.0.1
*.			5	IN	A	10.104.110.40
*.			5	IN	A	10.96.0.10

;; Query time: 27 msec
````

chrisohaver:
What does the corefile look like, and how is the service/endpoints defined (etc)?

miekg:
Cluster coredns is 1.9 kubeadm setup (can access cluster because not a home), so that standard that we had (didn't include `pod insecure` I think, but I'll double check).

coredns deploy is defined in the attached yaml file
[coredns.yaml.txt](https://github.com/coredns/coredns/files/1600168/coredns.yaml.txt)

chrisohaver:
The attached yaml doesn't have a `kubernetes` stanza...

miekg:
Sorry to be ignorant here, but what would that do?

chrisohaver:
Ah - i see - you didn't attach the yaml of the cluster-dns coredns.  I meant to ask for that corefile, not the corefile of the service you are querying. 

A few things that are confusing:

`dig svc coredns.pod.cluster.local` seems like an unusual query.  This is querying the name `svc` ... for type `coredns.pod.cluster.local `.  I'm guessing a typo in the comment? 

In `coredns.pod.cluster.local` the "coredns" segment is the namespace ... which means it's an incomplete name (behavior not defined in spec). Do you have a namespace named `coredns`?  Is the service `coredns` in a namespace  also named`coredns`?  This should not be a problem - just confusing.

[...]

PR #1349, opened 2018-01-05
plugin/kubernetes: partial fix for crazy pod queries

miekg:
This is probably the first in a series to fix "crazy" pod queries.
If the namespace doesn't exist return NXDOMAIN.

It might be worth extending this 1:1 to findServices as well.


chrisohaver (approved):
lgtm

chrisohaver:
partially fixes #1331

PR #1350, opened 2018-01-05
plugin/kubernetes: add explicit non-existing ns check

miekg:
We already do the right thing here, but add a test nonetheless.


miekg:
trivial enough, merging.

PR #1352, opened 2018-01-06
pl/kubernetes: fix pod lookups

miekg:
This fixes the lookup and stops special casing of PodInsecure. Think
I spotted another bug about exposed namespaces.

See #1331


miekg:
This actually removes PodModeInsecure wholly... Is that bad?

miekg:
*reads docs*

jesus kube-dns works like that. *fuck*

johnbelamaric:
Yeah. That's why we did it. But we did deprecate it in the spec: https://github.com/kubernetes/dns/blob/master/docs/specification.md#26---deprecated-records

So we should be able to kill it before too long.

PR #1353, opened 2018-01-06
kubernetes: Add zone filtering to fallthrough

johnbelamaric:

Adds zone filtering to fallthrough for K8s. Moves fallthrough logic to common plugin function.

https://github.com/coredns/deployment/issues/42

README updates pending.

If the overall approach looks OK I will add README updates as well as make similar changes in other plugins that implement `fallthrough`

### DIFF
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -40,7 +40,7 @@ type Kubernetes struct {
 	Namespaces       map[string]bool
 	podMode          string
 	endpointNameMode bool
-	Fallthrough      bool
+	Fallthrough      *[]string // nil = disabled, empty = all zones, o/w zones
 	ttl              uint32
 
 	primaryZoneIndex   int
--- a/plugin.md
+++ b/plugin.md
@@ -71,6 +71,9 @@ reverse cases and **all other** request are handled by the backing plugin. This 
 "fallthrough" does. To keep things explicit we've opted that plugins implement such behavior
 should implement a `fallthrough` keyword.
 
+The `fallthrough` directive should optionally accept a list of zones. Only queries for records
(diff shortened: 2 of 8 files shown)


plugin/kubernetes/kubernetes.go
```diff
@@ -40,7 +40,7 @@ type Kubernetes struct {
 	Namespaces       map[string]bool
 	podMode          string
 	endpointNameMode bool
-	Fallthrough      bool
+	Fallthrough      *[]string // nil = disabled, empty = all zones, o/w zones
```
miekg, line 43: a slice is already a pointer, so a pointer to a slice looks weird, unless you're doing something advanced/special?

miekg (commented):
LGTM, if it wasn't for the pointer thing. Also plugin.md talks about fallthrough; that might need a few extra words

plugin/kubernetes/kubernetes.go
johnbelamaric: The zero value for the slice is `[]string{}` not `nil`. I need three states: nil, empty, populated. Or I need two separate fields.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [WIP] Add zon..." ]
>
>
>
>> @@ -40,7 +40,7 @@ type Kubernetes struct {
> 	Namespaces       map[string]bool
> 	podMode          string
> 	endpointNameMode bool
>-	Fallthrough      bool
>+	Fallthrough      *[]string // nil = disabled, empty = all zones, o/w zones
>
>The zero value for the slice is `[]string{}` not `nil`. I need three states: nil, empty, populated. Or I need two separate fields.

Ah, yeah. Make sense.

LGTM

johnbelamaric:
Still to do: update other plugins, add integration tests.

PR #1355, opened 2018-01-07
Add pkg/fall for Fallthrough

miekg:
Move this into it's own package to facilitate tests. Important bug
was fixed: make the names fully qualified.

Add fall package to hosts, reverse, etcd, and fix kubernetes and any
tests. The k8s tests are still as-is, might need a future cleanup.

All README.mds have been updated as well.


johnbelamaric:
Nice cleanup. One thing though, I realized we don't need the nil/empty/zones thing if we just make the default when fallthrough is there but no zones ".". That will clean it up a lot more. No more need for the Nil business. In fact we can just *always* run fallthrough. If the directive were omitted, the list would be empty so it would never fallthrough. If the directive were included with no zones, we have stuck "." in as the only zone so everything will fallthrough. Code should be a lot simpler with that change.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add pkg/fall ..." ]
>Nice cleanup. One thing though, I realized we don't need the nil/empty/zones thing if we just make the default when fallthrough is there but no zones ".". That will clean it up a lot more. No more need for the Nil business. In fact we can just *always* run fallthrough. If the directive were omitted, the list would be empty so it would never fallthrough. If the directive were included with no zones, we have stuck "." in as the only zone so everything will fallthrough. Code should be a lot simpler with that change.

Merge this and then clean?

johnbelamaric:
Sure.

PR #1360, opened 2018-01-08
Align plugin/template usage and syntax with other plugins

rtreffer:

The plugin/template is currently scoped by regex with a fallthrough default.

This patch changes the config to be zone scoped and makes fallthrough an option instead of a default.

This change was discussed in #1298

README change is included, plugin/template wasn't shipped yet.

### DIFF
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -9,29 +9,34 @@ The *template* plugin allows you to dynamically repond to queries by just writin
 ## Syntax
 
 ~~~
-template CLASS TYPE [REGEX...] {
+template CLASS TYPE [ZONE...] {
+    [match REGEX...]
     [answer RR]
     [additional RR]
     [authority RR]
     [...]
-    [rcode CODE]
+    [rcode responsecode]
+    [fallthrough]
 }
 ~~~
 
-* **CLASS** the query class (usually IN or ANY)
-* **TYPE** the query type (A, PTR, ...)
+* **CLASS** the query class (usually IN or ANY).
+* **TYPE** the query type (A, PTR, ... can be ANY to match all types).
+* **ZONE** the zone scope(s) for this template. Defaults to '.' (everything).
 * **REGEX** [Go regexp](https://golang.org/pkg/regexp/) that are matched against the incoming question name. Specifying no regex matches everything (default: `.*`). First matching regex wins.
 * `answer|additional|authority` **RR** A [RFC 1035](https://tools.ietf.org/html/rfc1035#section-5) style resource record fragment
   build by a [Go template](https://golang.org/pkg/text/template/) that contains the reply.
 * `rcode` **CODE** A response code (`NXDOMAIN, SERVFAIL, ...`). The default is `SUCCESS`.
+* `fallthrough` Continue with the next plugin if the zone matched but no regex did not match.
 
-At least one answer section or rcode is needed.
+At least one `answer` or `rcode` directive is needed (e.g. `rcode NXDOMAIN`).
 
 [Also see](#also-see) contains an additional reading list.
 
 ## Templates
 
 Each resource record is a full-featured [Go template](https://golang.org/pkg/text/template/) with the following predefined data
+* `.Zone` the matched zone string (e.g. `example.`).
 * `.Name` the query name, as a string (lowercased).
 * `.Class` the query class (usually `IN`).
 * `.Type` the RR type requested (e.g. `PTR`).
@@ -57,6 +62,22 @@ Both failure cases indicate a problem with the template configuration.
 
 ## Examples
 
+### Resolve everything to NXDOMAIN
+
+The most simplistic template is
+
+~~~ corefile
+. {
+    template ANY ANY {
+      rcode NXDOMAIN
+    }
+}
+~~~
+
+1. This template uses the default zone (`.` or all queries)
+2. All queries will be answered (no `fallthrough`)
+3. The answer is always NXDOMAIN
+
 ### Resolve .invalid as NXDOMAIN
 
 The `.invalid` domain is a reserved TLD (see [RFC-2606 Reserved Top Level DNS Names](https://tools.ietf.org/html/rfc2606#section-2)) to indicate invalid domains.
@@ -65,7 +86,7 @@ The `.invalid` domain is a reserved TLD (see [RFC-2606 Reserved Top Level DNS Na
 . {
     proxy . 8.8.8.8
 
-    template ANY ANY "[.]invalid[.]$" {
+    template ANY ANY invalid. {
       rcode NXDOMAIN
       answer "invalid. 60 {{ .Class }} SOA a.invalid. b.invalid. (1 60 60 60 60)"
     }
@@ -75,6 +96,7 @@ The `.invalid` domain is a reserved TLD (see [RFC-2606 Reserved Top Level DNS Na
 1. A query to .invalid will result in NXDOMAIN (rcode)
 2. A dummy SOA record is send to hand out a TTL of 60s for caching
 3. Querying `.invalid` of `CH` will also cause a NXDOMAIN/SOA response
+4. The default regex is `.*`
 
 ### Block invalid search domain completions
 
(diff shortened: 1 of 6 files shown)


rtreffer:
@miekg hot off the press. This is an alternative syntax scoped by zone with an explicit regex match.

plugin/template/README.md
```diff
@@ -9,29 +9,34 @@ The *template* plugin allows you to dynamically repond to queries by just writin
...
 
-* **CLASS** the query class (usually IN or ANY)
-* **TYPE** the query type (A, PTR, ...)
+* **ZONE** the zone scope(s) for this template. Defaults to '.' (everything).
+* **CLASS** the query class (usually IN or ANY).
+* **TYPE** the query type (A, PTR, ... can be ANY to match all types).
 * **REGEX** [Go regexp](https://golang.org/pkg/regexp/) that are matched against the incoming question name. Specifying no regex matches everything (default: `.*`). First matching regex wins.
 * `RR` A [RFC 1035](https://tools.ietf.org/html/rfc1035#section-5) style `<rr>` fragment build by a [Go template](https://golang.org/pkg/text/template/) that contains the answer.
 * `responsecode` A response code (`NXDOMAIN, SERVFAIL, ...`). The default is `SUCCESS`.
+* `fallthrough` Continue with the next plugin if the zone matched but no regex did not match.
```
miekg: This also has a optional ZONES argument as well (as of yesterday evening) :)

plugin/template/README.md
```diff
@@ -9,29 +9,34 @@ The *template* plugin allows you to dynamically repond to queries by just writin
...
     [authority RR]
     [...]
     [rcode responsecode]
+    [fallthrough]
 }
 ~~~
 
-* **CLASS** the query class (usually IN or ANY)
-* **TYPE** the query type (A, PTR, ...)
+* **ZONE** the zone scope(s) for this template. Defaults to '.' (everything).
```
miekg: Put zone below **TYPE** as it is last here

plugin/template/README.md
```diff
@@ -9,29 +9,34 @@ The *template* plugin allows you to dynamically repond to queries by just writin
...
+* **CLASS** the query class (usually IN or ANY).
+* **TYPE** the query type (A, PTR, ... can be ANY to match all types).
 * **REGEX** [Go regexp](https://golang.org/pkg/regexp/) that are matched against the incoming question name. Specifying no regex matches everything (default: `.*`). First matching regex wins.
```
miekg: `match` **REGEX**

plugin/template/README.md
```diff
@@ -63,7 +84,7 @@ The `.invalid` domain is a reserved TLD (see [RFC-2606 Reserved Top Level DNS Na
 . {
     proxy . 8.8.8.8
 
-    template ANY ANY "[.]invalid[.]$" {
+    template ANY ANY invalid. {
```
miekg: Should we make this final dot implicit?

plugin/template/template_test.go
```diff
@@ -260,8 +226,12 @@ func TestHandler(t *testing.T) {
 
 	for _, tr := range tests {
 		handler := Handler{
-			Next:      test.NextHandler(rcodeFallthrough, nil),
-			Templates: []template{tr.tmpl},
+			Next:        test.NextHandler(rcodeFallthrough, nil),
+			Class:       tr.qclass,
+			Qtype:       tr.qtype,
+			Zones:       []string{"."},
+			Fallthrough: true,
```
miekg: See the new `fall` pkg in plugin/pkg/fall for the new way of doing fallthrough

miekg:
Let me merge #1361, but this PR looks good.

plugin/template/README.md
rtreffer: Good question, I'm using the same logic as etcd.... And I'm usually trying to mark qualified names as such.

Related: If nothing is specified I'm right now matching '.' - is there a reason to switch to the etcd/setup.go method of
```
		if len(etc.Zones) == 0 {
			etc.Zones = make([]string, len(c.ServerBlockKeys))
			copy(etc.Zones, c.ServerBlockKeys)
		}
```

And it looks like I should normalize, too (according to the etcd source)

rtreffer:
@miekg rebased on master and most points fixed.

Metrics are broken, I'll need another PR for that (and #1362 - remove regex from metric labels)

plugin/template/README.md
miekg: yes, this should be `Normalized`

plugin/template/README.md
```diff
@@ -88,14 +110,33 @@ path (`dc1.example.com`) added.
...
+    template IN ANY example.com. {
+      match "(example.com.dc1.example.com)$"
       rcode NXDOMAIN
       answer "{{ index .Match 1 }} 60 IN SOA a.{{ index .Match 1 }} b.{{ index .Match 1 }} (1 60 60 60 60)"
+      fallthrough
     }
 }
 ~~~
 
+<<<<<<< HEAD
```
miekg: This didn't survive the rebase.

plugin/template/README.md
```diff
@@ -9,29 +9,34 @@ The *template* plugin allows you to dynamically repond to queries by just writin
...
-template CLASS TYPE [REGEX...] {
+template CLASS TYPE [ZONE...] {
+    [match REGEX...]
     [answer RR]
     [additional RR]
     [authority RR]
     [...]
-    [rcode CODE]
+    [rcode responsecode]
+    [fallthrough]
```
miekg: fallthrough can now have optional ZONES

plugin/template/README.md
miekg: drop the final dot (as this is not needed anymore)

plugin/template/setup_test.go
```diff
@@ -93,30 +93,37 @@ func TestSetupParse(t *testing.T) {
...
 			}`,
 			false,
 		},
 		{
-			`template IN ANY "[.](example[.]com[.]dc1[.]example[.]com[.])$" {
+			`template IN ANY example.com. {
+				match "[.](example[.]com[.]dc1[.]example[.]com[.])$"
 				rcode NXDOMAIN
 				answer "{{ index .Match 1 }} 60 IN SOA a.{{ index .Match 1 }} b.{{ index .Match 1 }} (1 60 60 60 60)"
+				fallthrough example.com.
```
miekg: doesn't need closing dot

plugin/template/template.go
```diff
@@ -108,21 +132,16 @@ func executeRRTemplate(section string, template *gotmpl.Template, data templateD
 	return rr, nil
 }
 
-func (t template) match(state request.Request) (templateData, bool) {
+func (t template) match(state request.Request, zone string) (templateData, bool) {
 	q := state.Req.Question[0]
 	data := templateData{}
 
-	if t.class != dns.ClassANY && q.Qclass != dns.ClassANY && q.Qclass != t.class {
```
miekg: We don't need this shortcut any more?

miekg (changes requested):

miekg:
btw: I can wait with 1.0.3 until this is sorted out; there is not actual rush for 1.0.3 other than getting the code out there. Having a new plugin in a release always looks good :)

rtreffer:
Just got up. There is an issue in this PR around mutliple sections scoped by different zones....

I'll change this the following way: the plugin does a very rough / quick check (can I be responsible for this zone at all?) and the template does a more thorough one (actual zone, type, regex).

I'm adding a test for that now.

miekg:
> I'll change this the following way: the plugin does a very rough / quick check (can I be responsible for this zone at all?) and the template does a more thorough one (actual zone, type, regex).

This is how other plugins do this as well, so lgtm

rtreffer:
I'm now more carefully scoping the match checks and added a test to verify it works: https://github.com/soundcloud/coredns/blob/f047cb0085c3935e7491adb7541d2929189af557/plugin/template/template_test.go#L300

I still have to do the metrics change and the mentioned README cleanup.

rtreffer:
@miekg everything should be addressed, including #1360

miekg (approved):

Issue #1278, opened 2017-12-04
plugin/cache: prefetch doesn't actually prefetch


grobie:
We're about to consolidate our DNS infrastructure and plan to use CoreDNS everywhere (authoritative servers, resolvers, caches, forwarders). SoundCloud relies on DNS heavily for service discovery and quick DNS lookups are key for the overall latency.

While doing some load testing, I noticed that the prefetch option requested in #581 doesn't prefetch the record in the background, but blocks actual queries. This differs from implementations in other popular DNS servers (like bind9) and doesn't help to ensure fast lookups in case authoritative/upstream servers are slow. With the CoreDNS implementation, the prefetch option will simply cause an earlier request to become bound by the upstream latency, instead of allowing all queries to be answered from the cache.

> The intention is that the fresh copy would arrive before the existing copy expires, which ensures a uniform response time.

https://deepthought.isc.org/article/AA-01122/0/Early-refresh-of-cache-records-cache-prefetch-in-BIND-9.10.html

Question: Is this intentional? I'm happy to contribute a change to prefetch a soon-to-expire cache entry in the background (go routine) and ensuring uniform response times.

johnbelamaric:
Um. No, I am pretty sure that was not intentional...

grobie:
I started working on a new prefetch implementation.

miekg:
Just returning (the maybe somewhat stale) entry from the cache, while looking up the new value might be a relative small change?

Also in handler.go, if the response was in the cache it is written to the client immidiately, after that happened the prefetch stuff is checked. So why would this block new queries?

grobie:
> So why would this block new queries?

It was a combination of multiple issues leading to my false statement, mostly that prefetching didn't work at all in the default config for answers with low TTLs (<10s), and that both logs and metrics reported wrong response times for requests triggering the prefetch.

grobie:
I finally had some time to pick this story up again. I fixed a couple of issues with prefetching, see #1363. It's far from being perfect, but it provides the minimum changes we need to use CoreDNS as caching server.

In general, the cache implementation would benefit greatly from some refactoring and decoupling of concerns. I'll see if I can find some time to clean up the current state.

[...]

PR #1363, opened 2018-01-08
plugin/cache: Fix prefetching issues

grobie:

Several fixes around the prefetch behavior of plugin/cache. For more details, check the individual commit messages. The biggest behavioral changes are:

* new metric `coredns_cache_prefetch_total`
* allow prefetching of single requests
* allow prefetching of records with TTL < 10s (which was impossible before)
* prefetch in the background to not report wrong client request duration

#1278 

The documentation of the available metrics and their labels has changed.

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -32,6 +32,27 @@ type Cache struct {
 	prefetch   int
 	duration   time.Duration
 	percentage int
+
+	// Testing.
+	now func() time.Time
+}
+
+// New returns an initialized Cache with default settings. It's up to the
+// caller to set the Next handler.
+func New() *Cache {
+	return &Cache{
+		Zones:      []string{"."},
+		pcap:       defaultCap,
+		pcache:     cache.New(defaultCap),
+		pttl:       maxTTL,
+		ncap:       defaultCap,
+		ncache:     cache.New(defaultCap),
+		nttl:       maxNTTL,
+		prefetch:   0,
+		duration:   1 * time.Minute,
+		percentage: 10,
+		now:        time.Now,
+	}
 }
 
 // Return key under which we store the item, -1 will be returned if we don't store the
@@ -88,7 +109,7 @@ type ResponseWriter struct {
 // WriteMsg implements the dns.ResponseWriter interface.
 func (w *ResponseWriter) WriteMsg(res *dns.Msg) error {
 	do := false
-	mt, opt := response.Typify(res, time.Now().UTC())
+	mt, opt := response.Typify(res, w.now().UTC())
 	if opt != nil {
 		do = opt.Do()
 	}
@@ -140,11 +161,11 @@ func (w *ResponseWriter) set(m *dns.Msg, key int, mt response.Type, duration tim
 
 	switch mt {
 	case response.NoError, response.Delegation:
-		i := newItem(m, duration)
+		i := newItem(m, w.now(), duration)
 		w.pcache.Add(uint32(key), i)
 
 	case response.NameError, response.NoData:
-		i := newItem(m, duration)
+		i := newItem(m, w.now(), duration)
 		w.ncache.Add(uint32(key), i)
 
 	case response.OtherError:
--- a/plugin/cache/handler.go
+++ b/plugin/cache/handler.go
@@ -41,16 +41,18 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 
 			threshold := int(math.Ceil(float64(c.percentage) / 100 * float64(i.origTTL)))
 			if i.Freq.Hits() >= c.prefetch && ttl <= threshold {
-				cachePrefetches.Inc()
-				// When prefetching we loose the item i, and with it the frequency
-				// that we've gathered sofar. See we copy the frequencies info back
-				// into the new item that was stored in the cache.
-				prr := &ResponseWriter{ResponseWriter: w, Cache: c, prefetch: true}
-				plugin.NextOrFailure(c.Name(), c.Next, ctx, prr, r)
+				go func() {
+					cachePrefetches.Inc()
+					// When prefetching we loose the item i, and with it the frequency
+					// that we've gathered sofar. See we copy the frequencies info back
+					// into the new item that was stored in the cache.
+					prr := &ResponseWriter{ResponseWriter: w, Cache: c, prefetch: true}
+					plugin.NextOrFailure(c.Name(), c.Next, ctx, prr, r)
 
-				if i1 := c.exists(qname, qtype, do); i1 != nil {
-					i1.Freq.Reset(now, i.Freq.Hits())
-				}
+					if i1 := c.exists(qname, qtype, do); i1 != nil {
+						i1.Freq.Reset(now, i.Freq.Hits())
+					}
+				}()
 			}
 		}
 		return dns.RcodeSuccess, nil
--- a/plugin/cache/handler.go
+++ b/plugin/cache/handler.go
@@ -1,6 +1,7 @@
 package cache
 
 import (
+	"math"
 	"sync"
 	"time"
 
@@ -37,26 +38,21 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 
 		if c.prefetch > 0 {
 			i.Freq.Update(c.duration, now)
-		}
 
-		pct := 100
-		if i.origTTL != 0 { // you'll never know
-			pct = int(float64(ttl) / float64(i.origTTL) * 100)
-		}
+			threshold := int(math.Ceil(float64(c.percentage) / 100 * float64(i.origTTL)))
+			if i.Freq.Hits() >= c.prefetch && ttl <= threshold {
+				cachePrefetches.Inc()
+				// When prefetching we loose the item i, and with it the frequency
+				// that we've gathered sofar. See we copy the frequencies info back
+				// into the new item that was stored in the cache.
+				prr := &ResponseWriter{ResponseWriter: w, Cache: c, prefetch: true}
+				plugin.NextOrFailure(c.Name(), c.Next, ctx, prr, r)
 
-		if c.prefetch > 0 && i.Freq.Hits() >= c.prefetch && pct < c.percentage {
-			cachePrefetches.Inc()
-			// When prefetching we loose the item i, and with it the frequency
-			// that we've gathered sofar. See we copy the frequencies info back
-			// into the new item that was stored in the cache.
-			prr := &ResponseWriter{ResponseWriter: w, Cache: c, prefetch: true}
-			plugin.NextOrFailure(c.Name(), c.Next, ctx, prr, r)
-
-			if i1 := c.exists(qname, qtype, do); i1 != nil {
-				i1.Freq.Reset(now, i.Freq.Hits())
+				if i1 := c.exists(qname, qtype, do); i1 != nil {
+					i1.Freq.Reset(now, i.Freq.Hits())
+				}
 			}
 		}
-
 		return dns.RcodeSuccess, nil
 	}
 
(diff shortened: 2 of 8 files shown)


plugin/cache/cache.go
```diff
@@ -109,8 +110,9 @@ func (w *ResponseWriter) WriteMsg(res *dns.Msg) error {
 	if key != -1 && duration > 0 {
 		w.set(res, key, mt, duration)
 
-		cacheSize.WithLabelValues(Success).Set(float64(w.pcache.Len()))
-		cacheSize.WithLabelValues(Denial).Set(float64(w.ncache.Len()))
+		l := strings.Join(w.Zones, ",")
```
miekg: This might be pretty expensive and create zone gc, is there a way around it?

miekg: Why are we including this in the metrics?

plugin/cache/handler.go
```diff
@@ -37,25 +39,22 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
-			// that we've gathered sofar. See we copy the frequencies info back
-			// into the new item that was stored in the cache.
-			prr := &ResponseWriter{ResponseWriter: w, Cache: c, prefetch: true}
-			plugin.NextOrFailure(c.Name(), c.Next, ctx, prr, r)
-
-			if i1, _ := c.get(now, qname, qtype, do); i1 != nil {
-				i1.Freq.Reset(now, i.Freq.Hits())
+			threshold := int(math.Ceil(float64(c.percentage) / 100 * float64(i.origTTL)))
+			if i.Freq.Hits() > c.prefetch && ttl <= threshold {
+				go func() {
```
miekg: Why does this start another goroutine?

miekg (commented):
So TestPrefetch is unsufficient as well? (Not didn't touch any of the tests in this PR)

plugin/cache/handler.go
grobie: See the commit message and pr description.

> The logging and metrics plugins (among others) included the duration of
a cache prefetch in the request latency of client request. This change
fixes this wrong reporting and executes the prefetch request in a
goroutine in the background.

plugin/cache/cache.go
grobie: See the commit message and pr description.

> Partition metrics by zone label to prevent multiple caches from overwriting each other's metric states (e.g. capacity and size).

Every zone using a cache has a difference cache size and potentially a different capacity. So far we export the metrics of a random cache on scrape, with this change we're starting to export cache metrics per used cache.

grobie: This can be written GC free by currying metrics. I'll look into it.

miekg:
Please add a test showing prefetch *isn't* working.

I don't understand why the zones are included like this, there is no per-zone partition in the cache.

grobie:
> I don't understand why the zones are included like this, there is no per-zone partition in the cache.

Ugh, you're right, gonna remove it again. Not being able to configure cache TTL per zone will make things a lot harder.

I find this mismatch between the config (where I can happily define multiple independent zones with separate settings) and the actual behavior (global plugin instances, every plugin has to check again for each request whether it matches) very confusing in CoreDNS. Is there a design document explaining the rationale for not isolating different chains from each other? Shouldn't CoreDNS warn or error on a impossible config like this at least?

```
foo. {
  proxy . 8.8.8.8
  cache {
    prefetch 1
  }
}

bar. {
  proxy . 8.8.8.8
  cache 30 {
    prefetch 10
  }
}
```

grobie:
> So TestPrefetch is unsufficient as well?

Actually, the existing TestPrefetch is broken, it doesn't test anything. You can remove the `p = true` line and it'll still pass.

grobie:
Alright, I refactored the tests to actually test the prefetch behavior and added one for the broken prefetching of responses with TTL < 10s. If you revert the code change of the percentage calculation to the original implementation, the test will fail now:

```
--- FAIL: TestPrefetch (1.00s)
        prefech_test.go:114: redis.service.dc1.consul.: want request to trigger a prefetch
        prefech_test.go:124: redis.service.dc1.consul.: want answer redis.service.dc1.consul.   4       IN      A       127.0.0.2, got redis.service.dc1.consul.        5       IN      A       127.0.0.2
```

grobie:
This PR depends on #1380 being merged first.

plugin/cache/prefech_test.go
```diff
@@ -2,53 +2,181 @@ package cache
...
+					t.Errorf("%s: want request to trigger a prefetch", tt.qname)
+				}
+			}
+			if rec.Rcode != dns.RcodeSuccess {
+				t.Errorf("%s: want rcode %d, got %d", tt.qname, dns.RcodeSuccess, rec.Rcode)
+			}
+			if want, got := 1, len(rec.Msg.Answer); want != got {
+				t.Errorf("%s: want %d answer RR, got %d", tt.qname, want, got)
+			}
+			if want, got := test.A(v.answer), rec.Msg.Answer[0]; !reflect.DeepEqual(want, got) {
```
miekg: Not a superfan of using reflect here, can you just compare the strings?

plugin/cache/handler.go
```diff
@@ -25,37 +26,35 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
-			// that we've gathered sofar. See we copy the frequencies info back
-			// into the new item that was stored in the cache.
-			prr := &ResponseWriter{ResponseWriter: w, Cache: c, prefetch: true}
-			plugin.NextOrFailure(c.Name(), c.Next, ctx, prr, r)
 
-			if i1, _ := c.get(now, qname, qtype, do); i1 != nil {
-				i1.Freq.Reset(now, i.Freq.Hits())
+			threshold := int(math.Ceil(float64(c.percentage) / 100 * float64(i.origTTL)))
+			if i.Freq.Hits() >= c.prefetch && ttl <= threshold {
+				go func() {
```
miekg, line 44: (previous comment is gone?)

still don't see why this is needing to run in a goroutine?

miekg (commented):
good stuff.

I want to see if I can fail the prefetch test in master

miekg:
Test fails on master: (good!)

But it's quite elaborate (and from this output hard to debug). Can't we just have a simple prefetch/noprefetch (i.e. 2 or 3 test cases?)

~~~
--- FAIL: TestPrefetch (4.04s)
	prefech_test.go:153: prefetch.example.: want answer prefetch.example.	10	IN	A	127.0.0.1, got prefetch.example.	30	IN	A	127.0.0.1
	prefech_test.go:143: prefetch.example.: want request to trigger a prefetch
	prefech_test.go:153: prefetch.example.: want answer prefetch.example.	2	IN	A	127.0.0.1, got prefetch.example.	30	IN	A	127.0.0.1
	prefech_test.go:143: hits.reset.: want request to trigger a prefetch
	prefech_test.go:153: hits.reset.: want answer hits.reset.	7	IN	A	127.0.0.1, got hits.reset.	80	IN	A	127.0.0.1
	prefech_test.go:153: hits.reset.: want answer hits.reset.	73	IN	A	127.0.0.2, got hits.reset.	79	IN	A	127.0.0.1
	prefech_test.go:153: redis.service.dc1.consul.: want answer redis.service.dc1.consul.	4	IN	A	127.0.0.1, got redis.service.dc1.consul.	5	IN	A	127.0.0.1
	prefech_test.go:143: redis.service.dc1.consul.: want request to trigger a prefetch
	prefech_test.go:153: redis.service.dc1.consul.: want answer redis.service.dc1.consul.	1	IN	A	127.0.0.1, got redis.service.dc1.consul.	5	IN	A	127.0.0.1
	prefech_test.go:153: redis.service.dc1.consul.: want answer redis.service.dc1.consul.	4	IN	A	127.0.0.2, got redis.service.dc1.consul.	4	IN	A	127.0.0.1
	prefech_test.go:153: no.prefetch.: want answer no.prefetch.	15	IN	A	127.0.0.1, got no.prefetch.	30	IN	A	127.0.0.1
	prefech_test.go:153: no.prefetch.: want answer no.prefetch.	1	IN	A	127.0.0.1, got no.prefetch.	30	IN	A	127.0.0.1
	prefech_test.go:143: no.prefetch.: want request to trigger a prefetch
	prefech_test.go:153: no.prefetch.: want answer no.prefetch.	30	IN	A	127.0.0.2, got no.prefetch.	30	IN	A	127.0.0.1
FAIL
exit status 1
FAIL	github.com/coredns/coredns/plugin/cache	4.080s
~~~

plugin/cache/handler.go
grobie: Is the commit message not helpful?

> The logging and metrics plugins (among others) included the duration of
a cache prefetch in the request latency of client request. This change
fixes this wrong reporting and executes the prefetch request in a
goroutine in the background.

The prefetch should not be counted towards the client request latency, as that can trip alerts on high latency.

plugin/cache/prefech_test.go
grobie: Interesting, I've never heard an argument against reflect in test usage, especially DeepEqual is super common in all go code bases. But as you wish, will change it.

grobie:
@miekg Aaaalright, I removed one test case and some verifications, and made the test output more readable.

```
--- FAIL: TestPrefetch (2.01s)
    --- FAIL: TestPrefetch/hits.reset. (1.00s)
        prefech_test.go:123: after 1m13s: want request to trigger a prefetch
    --- FAIL: TestPrefetch/short.ttl. (1.00s)
        prefech_test.go:123: after 4s: want request to trigger a prefetch
FAIL
FAIL    github.com/coredns/coredns/plugin/cache 2.018s
```

Given the different behaviors of the cache plugin (handler fetches from the next handler and returns that response, handler returns cached item and does not hit the next handler, handler returns cached item and does hit the next handler), I don't see how to remove any more test cases.

plugin/cache/setup.go
```diff
@@ -115,6 +120,7 @@ func cacheParse(c *caddy.Controller) (*Cache, error) {
 					return nil, err
 				}
 				ca.ncap = ncap
+				ca.ncache = cache.New(ca.ncap)
```
miekg: you're now doing it in the loop which means if there are multiple of these specific in the Corefile you'll keep create caches (that are used, and should be garbage collected). But I don't see the improvement of this specific change?

plugin/cache/handler.go
miekg: but we already wrote to the client?

plugin/cache/prefech_test.go
```diff
@@ -6,49 +6,158 @@ import (
...
-	c.duration = 1 * time.Second
-	c.Next = PrefetchHandler(t, dns.RcodeSuccess, nil)
+	tests := []struct {
+		qname         string
+		ttl           int
+		prefetch      int
+		verifications []verification
+	}{
+		{
+			qname:    "hits.reset.",
```
miekg: nit: should still use the TLD example.{org,net}, hits.reset.example.org to not invent new TLDs

plugin/cache/cache.go
```diff
@@ -32,6 +32,27 @@ type Cache struct {
 	prefetch   int
 	duration   time.Duration
 	percentage int
+
+	// Testing.
+	now func() time.Time
+}
+
+// NewCache returns an initialized Cache with default settings. It's up to the
```
miekg: Just New()?

plugin/cache/handler.go
grobie: This doesn't matter. Other plugins like log and metrics wait for the termination of `ServeDNS()` (called by `NextOrFailure`) to determine the request latency.

plugin/cache/cache.go
grobie: Done.

plugin/cache/prefech_test.go
grobie: Done.

plugin/cache/setup.go
grobie: My assumption was they'll get gargage collected and it doesn't matter. Without this change, two caches will always be initialized even if now explicit capacity attribute is set. I see the actual issue in only having a single cache instance and a single mutex which is shared across all zones. I changed it back.

miekg:
Thanks for pushing this!

PR #1364, opened 2018-01-09
Overloaded

### LINKED ISSUE #1342: Metrics Request: I'm going to be overloaded
It'd be great if CoreDNS exposed a metric stating when it's becoming overloaded so users know when they need to scale.

miekg:
plugin/health: add 'I'm I overloaded metric"
   
    Query our on health endpoint and record (and export as a metric) the
    time it takes. The Get has a 5s timeout, that, when reached, will set
    the metric duration to 5s. The actually call "I'm I overloaded" is left
    to an external entity.

Fixes #1342

### DIFF
--- a/plugin/health/README.md
+++ b/plugin/health/README.md
@@ -25,6 +25,14 @@ supports health checks has a section "Health" in their README.
 
 Any plugin that implements the Healther interface will be used to report health.
 
+## Metrics
+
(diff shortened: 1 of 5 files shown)


miekg:
cc @swade1987

would something like this work for you?

swade1987:
@miekg LGTM!

PR #1366, opened 2018-01-09
plugin/template: small tweaks

miekg:
Small README updates, and fallthrough fixes (using less code)

(/cc @rtreffer small nits I saw when scanning your PR)


rtreffer (approved):
Yes, better!

PR #1370, opened 2018-01-10
Fix version of go get

### LINKED ISSUE #1368: version guard miekg/dns and caddy
As these two (among other) are not vendored it may make sense to check the version of the installed packages.

yongtang:
This fix fixes version fetched from `go get` so that versions are guarded.

```
github.com/mholt/caddy              v0.10.10
github.com/miekg/dns                v1.0.3
github.com/prometheus/client_golang v0.8.0
golang.org/x/net                    release-branch.go1.9 (branch)
golang.org/x/text                   e19ae1496984b1c655b8044a65c0300a3c878dd3
```

This fix is related #1368.

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,11 @@ godeps:
 	go get github.com/prometheus/client_golang/prometheus
 	go get golang.org/x/net/context
 	go get golang.org/x/text
+	(cd $(GOPATH)/src/github.com/mholt/caddy              && git checkout -q v0.10.10)
+	(cd $(GOPATH)/src/github.com/miekg/dns                && git checkout -q v1.0.3)
+	(cd $(GOPATH)/src/github.com/prometheus/client_golang && git checkout -q v0.8.0)
+	(cd $(GOPATH)/src/golang.org/x/net                    && git checkout -q release-branch.go1.9)
+	(cd $(GOPATH)/src/golang.org/x/text                   && git checkout -q e19ae1496984b1c655b8044a65c0300a3c878dd3)
 
 .PHONY: travis
 travis: check


miekg:
I was wondering if we should do this in code? I.e have a failing test have a version is too low?

This might be good enough though. WDYT?

miekg (approved):

yongtang:
@miekg I think we could use Makefile like that to make sure when we release, the versions of other packages are truly "pinned".

The version check in the code is also useful, as some developers may choose to rebuild CoreDNS with customized plugins (and not use the Makefile here).

miekg:
I did run into
~~~
cd /home/miek/g/src/golang.org/x/text                   && git checkout -q e19ae1496984b1c655b8044a65c0300a3c878dd3)
fatal: reference is not a tree: e19ae1496984b1c655b8044a65c0300a3c878dd3
Makefile:27: recipe for target 'godeps' failed
make: *** [godeps] Error 128
72.63s user 13.73s system 64.89s elapsed 133%CPU (make)
~~~
Maybe `get get -u` would be better?

yongtang:
@miekg I think `go get -u` is better. Also there are some additional issues. If repo is not a branch then `go get` will fail as well:
```
# git checkout eb22672bea55af56d225d4e35405f4d2e9f062a0
Previous HEAD position was e19ae14... currency: fix format in example
HEAD is now at eb22672... cmd/gotext: add update command
# go get -u golang.org/x/text
cd /go/src/golang.org/x/text; git pull --ff-only
You are not currently on a branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.

    git pull <remote> <branch>

package golang.org/x/text: exit status 1
```

Added #1372 to try to address the above issue.

PR #1371, opened 2018-01-10
Add version check for github.com/miekg/dns

yongtang:

This fix adds version check for github.com/miekg/dns in runtime so that only `v1.x.x` is supported.

This is related to #1368.


miekg:
Now that I see it in code we should probably not do it like this. I.e. in the makefile is sufficient.

yongtang:
@miekg Thanks. Let me close this PR.

PR #1385, opened 2018-01-12
plugin/dnssec: check validityperiod of RRSIGs

### LINKED ISSUE #1378: RRSIG caching issue
After 8 days of operation (with file and without changing the zone file) the cached signatures are still served (and they are only signed for 8 days).
[dnssec.go#L40](https://github.com/coredns/coredns/blob/master/plugin/dnssec/dnssec.go#L40) states that "Signatures will be cached for a short while.".
But I can't find any code where expired or old signatures are dropped.

miekg:
Somehow we missed implementing this. If a sig a retrieved from the
cache, but not valid anymore, regenerate it instead of server invalid
signatures.

Fixes #1378

### DIFF
--- a/plugin/dnssec/cache_test.go
+++ b/plugin/dnssec/cache_test.go
@@ -77,6 +77,6 @@ func TestCacheNotValidYet(t *testing.T) {
 
 	_, ok := d.get(k)
 	if ok {
-		t.Errorf("signature was added to the cache even though not valid")
+		t.Errorf("signature was added to the cache even though not valid yet")
 	}
 }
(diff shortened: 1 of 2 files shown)


Isolus:
Looks promising, but I think you meant `sixDays    = 6 * 24 * time.Hour` instead of `sixDays    = 8 * 24 * time.Hour`.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/dnssec..." ]
>Looks promising, but I think you meant `sixDays    = 6 * 24 * time.Hour` instead of `sixDays    = 8 * 24 * time.Hour`.

erh. yeah :)

PR #1392, opened 2018-01-15
Enhancement of external plugin enabling

yongtang:
This fix is an enhancement of external plugin enabling.

Previously, it was already able to build a customerized coredns with plugins enabled selectively, without changing coredns source code. However, all default plugins are actually bundled because of the import rule:
```
"github.com/coredns/coredns/coremain"
```

The issue is best described with the following:
```
root@localhost:/go/src/github.com/coredns/coredns/sample# cat sample.go
package main

import (
        _ "github.com/coredns/forward"

        "github.com/coredns/coredns/coremain"
        "github.com/coredns/coredns/core/dnsserver"
)

var directives = []string{
        "forward",
        "startup",
        "shutdown",
}

func init() {
        dnsserver.Directives = directives
}

func main() {
        coremain.Run()
}
root@localhost:/go/src/github.com/coredns/coredns/sample#
root@localhost:/go/src/github.com/coredns/coredns/sample# go build -v sample.go
root@localhost:/go/src/github.com/coredns/coredns/sample# ./sample -plugins
root@localhost:/go/src/github.com/coredns/coredns/sample# ./sample -plugins
Server types:
  dns

Caddyfile loaders:
  flag
  default

Other plugins:
  dns.auto
  dns.autopath
  dns.bind
  dns.cache
  dns.chaos
  dns.debug
  dns.dnssec
  dns.dnstap
  dns.erratic
  dns.errors
  dns.etcd
  dns.federation
  dns.file
  dns.forward
  dns.health
  dns.hosts
  dns.kubernetes
  dns.loadbalance
  dns.log
  dns.nsid
  dns.pprof
  dns.prometheus
  dns.proxy
  dns.reverse
  dns.rewrite
  dns.root
  dns.route53
  dns.secondary
  dns.template
....
```

This fix moves zplugins.go to a different package/directory so that
it is possible to "only import plugins as needed".

The following is the new output after this fix:
```
root@localhost:/go/src/github.com/coredns/coredns/sample# ./sample -plugins
Server types:
  dns

Caddyfile loaders:
  flag
  default

Other plugins:
  dns.forward
  dns.prometheus
  shutdown
  startup

root@localhost:/go/src/github.com/coredns/coredns/sample#
```


miekg:
note the plugin like proxy also provide external function that are used by other plugins like Lookup.

Issue #1382, opened 2018-01-11
plugin/fallback: proxy request in case of error returned by downstream plugins


fturib:
The idea is to check the return of regular downstream plugins, and be able to redirect the query to a 'fallback' proxy base on the code error returned.
NOTE: we would like use similar if the whole CoreDNS service is unhealthy. 

**proposition for config:**

```
fallback [ZONES…] {
  on RESPONSE-CODE [ IP:PORT or filename … ]
}
```
 
Example:
 
```
fallback example.com {
  on NXDOMAIN /etc/resolv.conf
  on SERVFAIL 10.10.10.10:100 8.8.8.8:53
}
```
  
So, you can have as many ”on” directives as you want. It could potentially re-use the proxy code so it could also accept any proxy options.

The usage right now would be to fallback in following cases:
1  CoreDNS is unhealthy (the whole stack is bypassed in this case)
2. Grpc communication is broken - one of the plugin cannot do its job.
3. REFUSED response.

NOTE: that could be en enhancement of proxy plugin, but we could make more generic as a separated plugin.
However, would need this plugin to be placed on top of order of the list of plugins.

NOTE: Right now the needs is for a real use case at Infoblox, proposition of plugin is from @johnbelamaric .

fturib:
Work in progress - by @yongtang

miekg:
icky, but interesting :)

about the proposed syntax: you probably want to prefix it (optionally) with a scheme: `grpc://`, `tls://`
and NXDOMAIN should maybe be DENIAL (to catch nodata)?

I think it would be good to keep this *outside* proxy, as that one is (too) complex already.

johnbelamaric:
wrt to the "NOTE" on "the whole CoreDNS being unhealthy" - I think you mean if all the proxy endpoints are unhealthy we would use this fallback list?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP - plugin/..." ]
>wrt to the "NOTE" on "the whole CoreDNS being unhealthy" - I think you mean if all the proxy endpoints are unhealthy we would use this fallback list?

Why wouldn't just adding these fallbacks in the proxy config result in the same 
behavior? (Except for the routing)

johnbelamaric:
[...]

PR #1398, opened 2018-01-17
Add fallback plugin (DNS endpoints only)

yongtang:

This fix adds a preliminary fallback plugin so that failed DNS queries could be sent to other endpoints.

This fix only accept DNS endpoints at the moment.

This fix is part of the #1382.

### DIFF
--- a/plugin/fallback/README.md
+++ b/plugin/fallback/README.md
@@ -1,0 +1,32 @@
+# fallback
+
+## Name
+
+*fallback* - send failed DNS queries to fallback endpoints.
+
+## Description
+
+This plugin sends failed DNS quieres (e.g., NXDOMAIN, SERVFAIL, etc)
+to fallback endpoints.
+
+## Syntax
+
+~~~ txt
+fallback [ZONE] {
+  on [FAILURE] [ENDPOINTS...]
+}
+~~~
+
+* **ZONE** the name of the domain to be accessed.
+* **FAILURE** the failure code of the DNS queries (NXDOMAIN, SERVFAIL, etc.).
+* **ENDPOINTS** the fallback endpoints to send to.
+
+## Examples
+
+~~~ corefile
+. {
+    fallback example.org {
+      on NXDOMAIN 10.10.10.10:53
+    }
+}
+~~~
--- a/plugin/fallback/fallback.go
+++ b/plugin/fallback/fallback.go
@@ -1,0 +1,42 @@
+package fallback
+
+import (
+	"log"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/request"
+
+	"github.com/miekg/dns"
+	"golang.org/x/net/context"
+)
+
+type processor struct {
+	rcode    int
+	endpoint string
+}
+
+func (p *processor) ErrorFunc(w dns.ResponseWriter, r *dns.Msg, rc int) error {
+	if rc == p.rcode {
+		state := request.Request{W: w, Req: r}
+		qname := state.Name()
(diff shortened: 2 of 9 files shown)


plugin/fallback/README.md
```diff
@@ -0,0 +1,32 @@
...
+* **ZONE** the name of the domain to be accessed.
+* **FAILURE** the failure code of the DNS queries (NXDOMAIN, SERVFAIL, etc.).
+* **ENDPOINTS** the fallback endpoints to send to.
+
+## Examples
+
+~~~ corefile
+. {
+    fallback example.org {
+      on NXDOMAIN 10.10.10.10:53
```
fturib, line 29: maybe exemple with several failures cases - to show the full usage. 

on NXDOMAIN
on SERVFAIL
on .. ???

is there a list of all events that can happen and we can use ?

fturib: yes : SERVFAIL, NXDOMAIN, ERROR

fturib: should NXDOMAIN be DENIAL ? or is there a NODATA ?

plugin/fallback/setup_test.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func TestSetupFallback(t *testing.T) {
+	c := caddy.NewTestController("dns", `fallback`)
+	if err := setup(c); err != nil {
+		t.Fatalf("Expected no errors, but got: %v", err)
+	}
+
+	c = caddy.NewTestController("dns", `fallback example.org {
+    on NXDOMAIN 10.10.10.10:100 8.8.8.8:53
```
fturib, line 16: test with several "on XXXX" ?
test with file (/etc/resolv.conf)

plugin/fallback/fallback.go
```diff
@@ -0,0 +1,42 @@
...
+	rcode    int
+	endpoint string
+}
+
+func (p *processor) ErrorFunc(w dns.ResponseWriter, r *dns.Msg, rc int) error {
+	if rc == p.rcode {
+		state := request.Request{W: w, Req: r}
+		qname := state.Name()
+		log.Printf("[INFO] Send fallback %q to %q", qname, p.endpoint)
+		_, err := dns.Exchange(r, p.endpoint)
```
fturib, line 23: I understand we use directly Exchange as most of plugin are using Proxy.newLookup(...).
What I see is that in proxy (that is calling exchange), there is a management of the TTL of the query. We do not care here in case of error ?

fturib: Knowing that this ErrorFunc can now be called whenever in the ServeDNS chain, including before any plugin is called, would it be safe to verify that the request is correctly defined ?

see in server.ServeDNS:
```
	if r == nil || len(r.Question) == 0 {
		DefaultErrorFunc(w, r, dns.RcodeServerFailure)
		return
	}
```

fturib (commented):
I am not a CoreDNS core dev -- very far from there :)
I like the way it is implement - I find it 'light", but also a bit scary as it is a centralized function that is overwritten. 
I hope my comments will help !

yongtang:
I think there are some miscommunications there. Let's figure out what is the purpose of `fallback`. I realized multiple endpoints does not make much sense. Each endpoint could return different errors NXDOMAIN/SERVFAIL/REFUSED again and the logic for errors in the middle will mess up the logic again.

yongtang:
I also initially though fallback was to send the failed query to some where else (thus no further handling).

Now it looks like the purpose is to `ignore the failure (like SERVFAIL)of certain queries and switch to another proxy endpoint`.

PR #1399, opened 2018-01-18
plugin/dnssec: fix blacklies for NXDOMAIN

miekg:
~~Test if we correctly filter out the correct type bitmap on an NODATA
response.~~
Fix the NSEC bitmap for NXDOMAIN responses and add tests.


miekg:
I don't understand wtf codecov is doing here.

PR #1405, opened 2018-01-18
Bypass predefined Rcode in proxy

yongtang:
This fix adds `fallthrough` keyword in proxy plugin
so that predefined Rcodes are bypassed if matches the
query response. The next plugin will pick up and process.


johnbelamaric:
This is not how `fallthrough` works in all other plugins, I don't think we should overload that term to mean something different.

miekg:
+1 to what John says.

The thing your making (reverse query router?) can be done in a plugin.

yongtang:
OK. Let me close this PR.

PR #1410, opened 2018-01-20
Emit InstanceStartupEvent after starting server

rdrozhdzh:
Add emitting InstanceStartupEvent after starting server instance. The similar piece of code exists in caddy project, see
https://github.com/mholt/caddy/blob/master/caddy/caddymain/run.go#L144
This allows coredns to run extra scripts after start or communicate about its readiness to parent process with "startup" plugin (github.com/mholt/caddy/startupshutdown) or it's modern replacement - "on" plugin (github.com/mholt/caddy/onevent).

none

none

### DIFF
--- a/coremain/run.go
+++ b/coremain/run.go
@@ -98,6 +98,9 @@ func Run() {
 		showVersion()
 	}
 
+	// Execute instantiation events
+	caddy.EmitEvent(caddy.InstanceStartupEvent, instance)
+
 	// Twiddle your thumbs
 	instance.Wait()
 }


miekg:
Nice.

Can/are you willing to work on porting "on" and remove startup (and shutdown as well?)?

miekg:
I've added the on plugin in #1412.  So scratch the previous comment.

I might be nice though to add some events, like for shutdown?

rdrozhdzh:
Thanks

rdrozhdzh:
I'm not sure what do you mean by "add some events". I believe, caddy will emit ShutdownEvent for us during SIGTERM processing (assuming https://github.com/mholt/caddy/pull/1995#pullrequestreview-90303615 is fixed). And user will be able to configure his action with "on shutdown" line in corefile.
The InstanceStartupEvent is also generated as part of restart (reload) during SIGUSR1 processing, that's fine. The only essential case when no events were generated was server startup. Therefore I requested the change in this PR.
Actually, caddy emits also StartupEvent during startup (https://github.com/mholt/caddy/blob/master/caddy/caddymain/run.go#L118), but I'm not sure how useful would it be, because seems this event is not supported by "on" plugin (https://github.com/mholt/caddy/blob/master/onevent/hook/config.go#L16).

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Emit Instance..." ]
>I'm not sure what do you mean by "add some events". I believe, caddy will emit ShutdownEvent for us during SIGTERM processing (assuming https://github.com/mholt/caddy/pull/1995#pullrequestreview-90303615 is fixed). And user will be able to configure his action with "on shutdown" line in corefile.
>The InstanceStartupEvent is also generated as part of restart (reload) during SIGUSR1 processing, that's fine. The only essential case when no events were generated was server startup. Therefore I requested the change in this PR.
>Actually, caddy emits also StartupEvent during startup (https://github.com/mholt/caddy/blob/master/caddy/caddymain/run.go#L118), but I'm not sure how useful would it be, because seems this event is not supported by "on" plugin (https://github.com/mholt/caddy/blob/master/onevent/hook/config.go#L16).

Ok, thanks for the info. Bit strange that the event publishing needs to be split 
between implementations.

So this `on.md` docs is correct in the description of the supported events?
https://github.com/coredns/coredns.io/pull/57/files

/Miek

--
Miek Gieben

rdrozhdzh:
mostly correct, added some my notes

PR #1412, opened 2018-01-22
Add on plugin

miekg:
Enable this Caddy plugin by default. Docs will go up coredns.io for this
as well.

See https://caddyserver.com/docs/on


rdrozhdzh:
looks good to me

PR #1415, opened 2018-01-23
fix computation of reverse zone based on Ipv6 CIDRs

fturib:

Fix computation of ipv6 reverse zone : for Ipv6 the value of one digit in the zone is 4 bit as for ipv4 it is 8 bits.

I did not opened an issue

no need to change.
I updated the UT, but did not run all of them .. we will see with Travis :)

### DIFF
--- a/plugin/normalize.go
+++ b/plugin/normalize.go
@@ -115,11 +115,16 @@ func SplitHostPort(s string) (host, port string, ipnet *net.IPNet, err error) {
 	if err == nil {
 		if rev, e := dns.ReverseAddr(ip.String()); e == nil {
 			ones, bits = n.Mask.Size()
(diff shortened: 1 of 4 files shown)


plugin/normalize.go
```diff
@@ -115,11 +115,15 @@ func SplitHostPort(s string) (host, port string, ipnet *net.IPNet, err error) {
 	if err == nil {
 		if rev, e := dns.ReverseAddr(ip.String()); e == nil {
 			ones, bits = n.Mask.Size()
+			sizeDigit := 8
```
miekg: Can you put a comment here that mirror your commit comment?

fturib: Sure.  I am adding this comment

fturib:
updated following review comments.

miekg (approved):

Issue #1416, opened 2018-01-23
plugin/cache: invalid truncated responses for large cached items


grobie:
The plugin/cache sends too large responses to clients, which breaks many DNS clients.

```
127.0.0.1 - [23/Jan/2018:12:36:53 +0100] "SRV IN redacted.example.org. udp 76 false 512" NOERROR qr,tc,rd,ra 474 19.132796ms
127.0.0.1 - [23/Jan/2018:12:36:53 +0100] "SRV IN redacted.example.org. tcp 76 false 65535" NOERROR qr,rd,ra 5676 30.502231ms
127.0.0.1 - [23/Jan/2018:12:36:56 +0100] "SRV IN redacted.example.org. udp 76 false 512" NOERROR qr,tc,rd,ra 5596 573.938µs
```

miekg:
closed via #1417

PR #1417, opened 2018-01-23
Fix truncation of messages longer than permitted by the client

grobie:

CoreDNS currently doesn't respect the maximum response size advertised
by the client and returns the full answer on a message with the TC bit
set. This breaks client implementations which rely on DNS servers
respecting the advertised size limit, for example the Ruby stdlib
client. It also has negative network performance implications, as large
messages will be split up into multiple UDP packets, even though the
client will discard the truncated response anyway.

While RFC 2181 permits the response of partial RRSets, finding the
correct number of records fitting into the advertised response size is
non-trivial. As clients should ignore truncated messages, this change
simply removes the full RRSet on truncated messages.

#1416, but all plugins using `request.Scrub()` are affected by the incorrect truncation handling.

n/a


grobie:
Any comments?

PR #1420, opened 2018-01-23
Fallback plugin

johnbelamaric:

Implements the `fallback` plugin.

#1382 

It will need its own README.


johnbelamaric:
@fturib

johnbelamaric:
This uses the proxy upstream parsing. It supports everything the proxy does. It should work like this:

```
{
  fallback nxdomain foo.com /etc/resolv.conf {
    # proxy options
  }
  fallback refused bar.com /etc/resolv.conf {
    # other options
  }
  fallback servfail . 8.8.8.8:53
  proxy . /etc/resolv.conf
}
```

for example. I haven't tried that yet though.

johnbelamaric:
at this point exactly ONE `fallback` is allowed for a given `rcode`. maybe that should change.

plugin/fallback/setup.go
```diff
@@ -0,0 +1,53 @@
...
+		}
+
+		fmt.Printf("Maps to rcode %d\n", rc)
+
+		u, err := proxy.NewStaticUpstream(&c.Dispenser)
+		if err != nil {
+			return plugin.Error("fallback", err)
+		}
+
+		f.rules[rc] = u
```
fturib: Hum.. for what I understand, this setup is called once per word 'fallback' found in the config file.
Therefore, the way the syntax of this stanza is designed, you can register only ONE rcode per 'fallback'. 
in other word this map named rules could be replaced by a couple (int, Upstream).
Am  I correct ?

johnbelamaric: yes, one `rcode` per `fallback` directive. See the comment above - I changed the syntax from the original, so it would be easy to re-use the `proxy` code.

There is only *one* fallback plugin, not *one per directive*. So, this one plugin processes *all* `fallback` directives, and so they end up in this map.

fturib: maybe I misread the setup call ...
Ok. I double checked when is called the setup .. and it says **once per key for each server block**.
Very good. - exactly what is needed :)

```
/ SetupFunc is used to set up a plugin, or in other words,
// execute a directive. It will be called once per key for
// each server block it appears in.
type SetupFunc func(c *Controller) error
```

johnbelamaric:
Ko added some tests and now this is ready for review.

core/plugin/zplugin.go
```diff
@@ -15,6 +15,7 @@ import (
 	_ "github.com/coredns/coredns/plugin/erratic"
 	_ "github.com/coredns/coredns/plugin/errors"
 	_ "github.com/coredns/coredns/plugin/etcd"
+	_ "github.com/coredns/coredns/plugin/fallback"
```
miekg, line 18: no reason to put it here? Can bump it to its own repo.

plugin/fallback/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Syntax
+
+```
+{
+    fallback RCODE PROXY_PARAMS
+}
+```
+
+* `RCODE` is the string representation of the error response code. The set of valid error
```
miekg, line 26: Nit: `**RCODE**` same for PROXY_PARAMS

plugin/fallback/README.md
```diff
@@ -0,0 +1,73 @@
...
+strings are defined as `RcodeToString` in <https://github.com/miekg/dns/blob/master/msg.go>
```
miekg, line 27: I would list them here? Instead of pointing to a (large) file

plugin/fallback/README.md
```diff
@@ -0,0 +1,73 @@
...
+* `PROXY_SPECS` accepts the same parameters as the *proxy* plugin
+<https://github.com/coredns/coredns/tree/master/plugin/proxy>.
```
miekg, line 29: coredns.io link instead?

plugin/fallback/fallback.go
```diff
@@ -0,0 +1,59 @@
+// Package fallback implements a fallback plugin for CoreDNS
+package fallback
+
+import (
+	"context"
+
+	"github.com/miekg/dns"
```
miekg, line 7: 3rd party should be last import.

plugin/fallback/fallback.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/pkg/nonwriter"
+	"github.com/coredns/coredns/plugin/proxy"
+)
+
+// Fallback plugin
```
miekg, line 14: That's not really a comment... (also needs closing dot)

plugin/fallback/fallback.go
```diff
@@ -0,0 +1,59 @@
...
+type Fallback struct {
+	Next  plugin.Handler
+	trace plugin.Handler
+	rules map[int]proxy.Upstream
+	proxy proxyCreator
+}
+
+// proxyCreator creates a proxy with the specified upstream
+type proxyCreator interface {
+	Create(trace plugin.Handler, upstream proxy.Upstream) plugin.Handler
```
miekg, line 24: Typically called `New` in Go

miekg: Why is this an interface? Only has one function so you can just use NewFunc with that signature?

plugin/fallback/fallback.go
```diff
@@ -0,0 +1,59 @@
...
+
+// fallbackProxyCreator implements the proxyCreator interface
+// Used by the fallback plugin to create proxy using specified for upstream
+type fallbackProxyCreator struct{}
+
+func (f fallbackProxyCreator) Create(trace plugin.Handler, upstream proxy.Upstream) plugin.Handler {
+	return &proxy.Proxy{Trace: trace, Upstreams: &[]proxy.Upstream{upstream}}
+}
+
+func newFallback(trace plugin.Handler) (f *Fallback) {
```
miekg, line 35: new or New

plugin/fallback/fallback.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// ServeDNS implements the plugin.Handler interface.
+func (f Fallback) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	nw := nonwriter.New(w)
+	rcode, err := plugin.NextOrFailure(f.Name(), f.Next, ctx, nw, r)
+	if err != nil {
+		return rcode, err
+	}
+	//WTF? WHY DOES PROXY ALWAYS RETURN 0
```
miekg, line 46: because it *always* writes to the client

plugin/fallback/setup.go
```diff
@@ -0,0 +1,54 @@
...
+
+func init() {
+	caddy.RegisterPlugin("fallback", caddy.Plugin{
+		ServerType: "dns",
+		Action:     setup,
+	})
+}
+
+func setup(c *caddy.Controller) error {
+	t := dnsserver.GetConfig(c).Handler("trace")
```
miekg, line 23: note: this only works because `trace` has been setup before fallback

plugin/fallback/setup.go
```diff
@@ -0,0 +1,54 @@
...
+	f := newFallback(t)
+
+	for c.Next() {
+		var rcode string
+		if !c.Dispenser.Args(&rcode) {
```
miekg, line 28: `c.Dispenser`? Why is that field named and used here?

plugin/fallback/fallback_test.go
```diff
@@ -0,0 +1,226 @@
+package fallback
+
+import (
+	"context"
+	"testing"
+
+	"github.com/miekg/dns"
+
+	"fmt"
```
miekg, line 9: formatting

plugin/fallback/fallback_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/coredns/coredns/plugin/pkg/healthcheck"
+	"github.com/coredns/coredns/plugin/proxy"
+	"github.com/coredns/coredns/plugin/test"
+	"github.com/coredns/coredns/request"
+)
+
+// stubNextHandler is used to simulate a proxy plugin.
```
miekg, line 19: the testing code has more comments on the non-testing code

plugin/fallback/fallback_test.go
```diff
@@ -0,0 +1,226 @@
...
+	expectedUpstream proxy.Upstream
+	called           int
+	t                *testing.T
+}
+
+func (c *testProxyCreator) Create(trace plugin.Handler, upstream proxy.Upstream) plugin.Handler {
+	c.called++
+
+	// Ensure that it is called with the expected Upstream
+	if c.expectedUpstream != nil && c.expectedUpstream != upstream {
```
miekg, line 40: why do you need to guard against this in a test that you control?

plugin/fallback/fallback_test.go
```diff
@@ -0,0 +1,226 @@
...
+		c.t.Fatalf("Expected upstream passed to proxyCreator is '%v', but got '%v'",
```
miekg, line 41: calling t.Fatalf in a non test function? Should return an err

plugin/fallback/fallback_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+// dummyExchanger implements the proxy.Exchanger interface
+// It is used solely to implement the dummyUpstream above
+type dummyExchanger struct{}
+
+func (e dummyExchanger) Exchange(ctx context.Context, addr string, state request.Request) (*dns.Msg, error) {
+	return &dns.Msg{}, nil
+}
+
+func (e dummyExchanger) Protocol() string {
```
miekg, line 82: These functions call all be put on 1 line

plugin/fallback/fallback_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+			// Ensure that the proxyCreator is called once
+			if proxyCreator.called != 1 {
+				t.Fatalf("Expect proxy creator to be called once, but got '%d", proxyCreator.called)
+			}
+		})
+	}
+}
+
+func TestFallbackNotCalled(t *testing.T) {
```
miekg, line 162: Why all this mocking and not just checking if the impl does the right thing if an upstream returns nxdomain (i.e. dnstest.Server should make this relatively painless)

plugin/fallback/README.md
yuewko: Sounds good. Will fix.

plugin/fallback/README.md
yuewko: Yeah... but RcodeToString is a rather long list.

How about including a handful of typical example rcodes here, and refer to that link for the complete list?

plugin/fallback/README.md
yuewko: Sounds good. Will switch.

plugin/fallback/fallback.go
yuewko: My bad. Will fix.

plugin/fallback/fallback.go
yuewko: OK. Will add to comments in fallback.go and setup.go as appropriate.

plugin/fallback/fallback.go
yuewko: This allows me to have a stub implementation used for unit test.

Plus, will rename method as per earlier comment.

plugin/fallback/fallback_test.go
yuewko: My bad. Will fix.

plugin/fallback/fallback_test.go
yuewko: Yes. Will address comments.

plugin/fallback/fallback_test.go
yuewko: These checks are not meant to guard against error conditions. They are there to check against expected behavior as per test case.

plugin/fallback/fallback_test.go
yuewko: As per comment above, the check here is indeed part of the test.

Having said that, I do think it is more appropriate to use t.Errorf (instead of t.Fatalf). Will fix.

plugin/fallback/fallback_test.go
yuewko: Good idea. Will fix.

plugin/fallback/fallback_test.go
yuewko: I must confess that I'm not familiar with dnstest. I just took a quick look and it seems, IMHO, that it looks more akin to integration test.

How about adding a couple of tests using dnstest to supplement?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fallback plug..." ]
>yuewko commented on this pull request.
>
>
>
>> +}
>+
>+// testProxyCreator implements the proxyCreator interface.
>+// Used by unit test to verify that the proxyCreator.Create() method is called as expected
>+type testProxyCreator struct {
>+	expectedUpstream proxy.Upstream
>+	called           int
>+	t                *testing.T
>+}
>+
>+func (c *testProxyCreator) Create(trace plugin.Handler, upstream proxy.Upstream) plugin.Handler {
>+	c.called++
>+
>+	// Ensure that it is called with the expected Upstream
>+	if c.expectedUpstream != nil && c.expectedUpstream != upstream {
>+		c.t.Fatalf("Expected upstream passed to proxyCreator is '%v', but got '%v'",
>
>As per comment above, the check here is indeed part of the test.
>
>Having said that, I do think it is more appropriate to use t.Errorf (instead of t.Fatalf). Will fix.

No you should return the error can not call fatalf here

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fallback plug..." ]
>Yeah... but RcodeToString is a rather long list.
>
>How about including a handful of typical example rcodes here, and refer to that link for the complete list?

Ah, true. Putting a couple here wont hurt, I guess.

yuewko:
re: https://github.com/coredns/coredns/pull/1420#issuecomment-363096087
In addition to replacing t.Fatalf with t.Errorf, I'll also have the Create method return an error.

re: https://github.com/coredns/coredns/pull/1420#issuecomment-363097273
Sound good, I'll do that.

Thanks!!

core/plugin/zplugin.go
johnbelamaric: yeah, we could do that. @yuewko I will create a separate repo for it in CoreDNS org.

yuewko:
Sounds good!

--
Yue Ko
Staff Software Engineer, Cloud and Network Automation
Infoblox  |  Control Your Network
(T) +1-443-603-0249  (M) +1-410-991-6293
yko@infoblox.com<mailto:yko@infoblox.com>  |  www.infoblox.com<http://www.infoblox.com/>

miekg:
moved to github.com/coredns/fallback, coredns.io/explugins also updated.

PR #1422, opened 2018-01-24
Remove pointers to labels.Selector and pass normally instead

ZloyDyadka:
Remove pointers to labels.Selector and pass normally instead, because the value of an uninitialized variable of interface type is nil.


johnbelamaric:
@miekg chris set it up so it now does the CI automatically and it shows up in the "checks"

PR #1425, opened 2018-01-25
plugin/forward: add out of tree forward plugin

miekg:
This is a simpler proxy than *proxy*, include by default so it is easier
to switch (i.e. no recompile). It lacks features compared to proxy (did I
say it was simpler), but does cache udp and tcp connection, so it is
faster than proxy.


miekg:
check if this compiles without mucking w/ `Makefile`'s godeps

miekg:
Can't make sense of the coredns/ci failure - *seems* unrelated?

chrisohaver:
It appears to get confused when doing `go get -u github.com/coredns/forward`, which was added in this PR...

It's pretty consistent, PRs of branches that don't include this change still work, whereas any PRs for branches that include this change fail in the same way.

Looking into the error...

PR #1426, opened 2018-01-25
Minor refactor of proxy parsing to make upstreams re-usable

johnbelamaric:

Refactors upstream parsing so we can re-use it in other plugins than proxy. For example, fallback. Doing this in a separate PR means fallback can be external if we want.

It creates a function NewStaticUpstream that performs the parsing starting at the FROM.

#1382 

None.


miekg:
I'm fine with this, but I'm not sure if you want proliferate *proxy* and maybe instead focus on *forward*
Also see #1443

johnbelamaric:
Ok. Well, I will merge anyway, we can see about trying it with `forward`.

PR #1427, opened 2018-01-25
plugin/kubernetes: Document Stub Domains configuration

chrisohaver:
This addition to the README discusses the caveats of using *proxy* to replicate kube-dns stub domain configurations.  Specifically, it identifies that upstream lookups do not support stub domains.   It also highlights the danger of using `upstream 127.0.0.1` as a workaround.


miekg:
Think we should try and make upstream do the right thing in this case (if possible)

chrisohaver:
Closing, since it looks like we are doing #1484.
Updates to documentation will be in that PR.

PR #1435, opened 2018-01-25
Tolerate go get failures for coredns/forward

### LINKED ISSUE #1442: Builds from within a branch fail
Since #1425, when doing a `make` while in a branch (not master), the build fails in the `godeps` target.  This is also the cause of the `coredns/ci` integration test failures.

```
go get -u github.com/coredns/forward
# cd /Users/cohaver/go/src/github.com/coredns/coredns; git pull --ff-only
There is no tracking information for the current branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.
...
```

Manually doing `go get -v -u github.com/coredns/forward` yields ...

```
go get -v -u github.com/coredns/forward
github.com/coredns/forward (download)
github.com/coredns/coredns (download)
# cd /Users/cohaver/go/src/github.com/coredns/coredns; git pull --ff-only
...
```

So - git appears to be getting confused when trying to get coredns, which happens to be the current repository.

chrisohaver:

Make will tolerate failures when go-getting coredns/forward. Specifically, the failure to get coredns/coredns, when coredns/coredns is the current project.

This fix is a hack. But its IMO a workable one until we identify a better fix.

Fixes #1442

n/a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,6 @@ testk8s: check
 
 .PHONY: godeps
 godeps:
-	-git checkout master
 	(cd $(GOPATH)/src/github.com/mholt/caddy 2>/dev/null              && git checkout -q master 2>/dev/null || true)
 	(cd $(GOPATH)/src/github.com/miekg/dns 2>/dev/null                && git checkout -q master 2>/dev/null || true)
 	(cd $(GOPATH)/src/github.com/prometheus/client_golang 2>/dev/null && git checkout -q master 2>/dev/null || true)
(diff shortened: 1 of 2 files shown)


miekg:
is *forward* being pulled in the dir where we just checkout coredns?

chrisohaver:
Travis builds ok... but it seems that the travis is building with a simple "go test"... which means its getting dependency heads, and building on those.
It's not being built on the same versions of dependencies checked out by that unusual "godeps" section in the Makefile (where it's checking out specific branches).  We should fix that.

chrisohaver:
Nevermind - i see that the "test" target does call "godeps"... so we should be good on the unit tests.
(it calls it as part of the "check" target ... which i was assuming was only doing "checks" for some reason)

I suppose travis build OK because its building in master... vs building in a branch in local builds and in the CI...

chrisohaver:
I confirmed that when building locally on my laptop, I get the same error when building from within a branch, but everything builds ok when building in master.  I haven't yet figured out why that makes a difference...

chrisohaver:
Well - the hack above (in "chop" 206817f) makes things work with brute force.

But this wont work for local builds with uncommitted changes.  To account for those we'd have to stash changes, switch to master, do the go gets, then switch back and unstash.  Which, IMO, is too obnoxious.

So we need a better fix.

chrisohaver:
Last commit tolerates failures in the go get of `coredns/forward`.  Since the only dependent package giving trouble is coredns/coredns (which we have already), it seems OK to ignore.

Travis and coredns/ci tests look like they are doing ok with this change.  Also this fix should work locally with uncommitted changes in a branch since it doesn't involve switching back to master.

It's still a hack... but IMO it's one we can work with.

fturib:
@miekg : are you ok to merge this fix so we can have build on local branches and ci working again ?

johnbelamaric:
Can we just remove the go get for the coredns/forward? Because this line is already there:

```
        (cd $(GOPATH)/src/github.com/coredns/forward 2>/dev/null          && git checkout -q master 2>/dev/null || true)
```

johnbelamaric:
Ok, I see now that `-f` basically lets `coredns/coredns` be a fork (or branch?), which otherwise go get checks on and rejects.

PR #1441, opened 2018-01-26
plugin/proxy: Don't enable HTTP healthchecking if not configured

### LINKED ISSUE #1440: plugin/proxy: hosts will be marked down forever without healthcheck config
With a minimal proxy config not setting a HTTP healthcheck endpoint, an upstream host which had occassionally issues will eventually be marked down forever, without the chance to recover.

The problem is that CoreDNS will implicitly activate HTTP healthchecks even if not configured. Most common DNS recursors don't provide a HTTP healtcheck interface and will therefore fail these healthchecks.

This is quite critical and has caused a small outage for us.

grobie:

HTTP healthchecking will be implicitely activated for proxy upstream
hosts, even if not configured. The README states that not using the
health_check directive will disable HTTP healthchecks though.

It seems to me that the availability of the HealthCheck.Path attribute
is used as indicator whether HTTP healthchecks should be used or not.
The normalizeCheckURL() function didn't check that attribute though,
always returning a CheckURL. This would increase the healthcheck failure
on every third failure in plugin/proxy, without any possibility for the
upstream host to be marked as healthy again. This would eventually
remove all upstream hosts from the serving pool.

Fixes #1440 

n/a

### DIFF
--- a/plugin/pkg/healthcheck/healthcheck.go
+++ b/plugin/pkg/healthcheck/healthcheck.go
@@ -212,6 +212,10 @@ func (u *HealthCheck) Select() *UpstreamHost {
 
 // normalizeCheckURL creates a proper URL for the health check.
 func (u *HealthCheck) normalizeCheckURL(name string) string {
+	if u.Path == "" {
+		return ""
+	}
+
 	// The DNS server might be an HTTP server.  If so, extract its name.
 	hostName := name
 	ret, err := url.Parse(name)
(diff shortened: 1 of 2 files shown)


plugin/pkg/healthcheck/healthcheck.go
```diff
@@ -212,6 +212,10 @@ func (u *HealthCheck) Select() *UpstreamHost {
 
 // normalizeCheckURL creates a proper URL for the health check.
 func (u *HealthCheck) normalizeCheckURL(name string) string {
+	if u.Path == "" {
```
fturib, line 215: That is a good test to avoid to check an unwanted path.

fturib: I see the point : HealthCheckURL is called directly by proxy, even though the workers are not triggered in Start() because path is empty.

fturib (approved):

PR #1445, opened 2018-01-26
Reload hook

johnbelamaric:

Creates a new plugin that monitors the Corefile and reloads it when it changes.

none

New readme added

### DIFF
--- a/plugin/reload/reload.go
+++ b/plugin/reload/reload.go
@@ -2,7 +2,7 @@ package reload
 
 import (
 	"crypto/md5"
-	"fmt"
+	"log"
 	"time"
 
 	"github.com/mholt/caddy"
@@ -11,12 +11,11 @@ import (
 // reload periodically checks if the Corefile has changed, and reloads if so
 
 type reload struct {
-	instance   *caddy.Instance
-	interval   time.Duration
-	sum        [md5.Size]byte
-	serverType string
-	stopped    bool
-	quit       chan bool
+	instance *caddy.Instance
+	interval time.Duration
+	sum      [md5.Size]byte
+	stopped  bool
+	quit     chan bool
 }
 
 func hook(event caddy.EventName, info interface{}) error {
@@ -49,7 +48,7 @@ func hook(event caddy.EventName, info interface{}) error {
 				if s != r.sum {
 					_, err := r.instance.Restart(corefile)
 					if err != nil {
-						fmt.Printf("[ERROR] Corefile changed but reload failed: %s\n", err)
+						log.Printf("[ERROR] Corefile changed but reload failed: %s\n", err)
 						continue
 					}
 					// we are done, this hook gets called again with new instance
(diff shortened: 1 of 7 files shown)


johnbelamaric:
I haven't actually checked if this works as expected in K8s. I think it will - I believe when you update a ConfigMap for a file it will get delivered to running pods.

johnbelamaric:
One weirdness. You have to put it in a server block. It's global and probably shouldn't have to be but I haven't dug deeply enough into Caddyfile parsing to know.

miekg:
Re: global. Yeah we need to think on how we deal with that. Caddyfile supports a global section, but I could never figure out how to make use of that.

plugin/reload/reload.go
```diff
@@ -0,0 +1,66 @@
...
+			case <-tick.C:
+				corefile, err := caddy.LoadCaddyfile(r.instance.Caddyfile().ServerType())
+				if err != nil {
+					continue
+				}
+				s := md5.Sum(corefile.Body())
+				if s != r.sum {
+					_, err := r.instance.Restart(corefile)
+					if err != nil {
+						fmt.Printf("[ERROR] Corefile changed but reload failed: %s\n", err)
```
miekg: `log`?

plugin/reload/reload.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/mholt/caddy"
+)
+
+// reload periodically checks if the Corefile has changed, and reloads if so
+
+type reload struct {
+	instance   *caddy.Instance
+	interval   time.Duration
+	sum        [md5.Size]byte
+	serverType string
```
miekg: why do we need serverType? which for us is always 'dns'?

plugin/reload/setup.go
```diff
@@ -0,0 +1,72 @@
...
+	c.OnFinalShutdown(func() error {
+		r.quit <- true
+		return nil
+	})
+
+	return nil
+}
+
+const (
+	defaultInterval = 30
```
miekg: just make this a time.Duration so you don't need the `* time.Second` above.

plugin/reload/reload.go
johnbelamaric: oops. I ended up getting it from Caddyfile Input. The way this is written it could be used in any server type.

plugin/reload/reload.go
johnbelamaric: OK. Other places we just fmt.Println but Caddy uses log.

plugin/reload/setup.go
johnbelamaric: ok

miekg:
Re: log vs fmt. Hmm.. another thing we should document. I'm not sure now.

johnbelamaric:
Just FYI for anyone that is looking, I did try this on K8s and config maps are updated. The jitter business probably wasn't necessary for that case as it seems the pod config maps don't all get updated at the same time anyway.

PR #1447, opened 2018-01-27
plugin/forward: add it

miekg:
This moves coredns/forward into CoreDNS.

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -1,0 +1,144 @@
+# forward
+
+*forward* facilitates proxying DNS messages to upstream resolvers.
+
+The *forward* plugin is generally faster (~30+%) than *proxy* as it re-uses already openened sockets
+to the upstreams. It supports UDP, TCP and DNS-over-TLS and uses inband healthchecking that is
+enabled by default.
+When *all* upstreams are down it assumes healtchecking as a mechanism has failed and will try to
+connect to a random upstream (which may or may not work).
+
+## Syntax
+
+In its most basic form, a simple forwarder uses this syntax:
+
+~~~
+forward FROM TO...
+~~~
+
+* **FROM** is the base domain to match for the request to be forwarded.
+* **TO...** are the destination endpoints to forward to. The **TO** syntax allows you to specify
+  a protocol, `tls://9.9.9.9` or `dns://` for plain DNS. The number of upstreams is limited to 15.
+
+The health checks are done every *0.5s*. After *two* failed checks the upstream is considered
+unhealthy. The health checks use a recursive DNS query (`. IN NS`) to get upstream health. Any
+response that is not an error (REFUSED, NOTIMPL, SERVFAIL, etc) is taken as a healthy upstream. The
+health check uses the same protocol as specific in the **TO**. On startup each upstream is marked
+unhealthy until it passes a healthcheck. A 0 duration will disable any healthchecks.
+
+Multiple upstreams are randomized (default policy) on first use. When a healthy proxy returns an
+error during the exchange the next upstream in the list is tried.
+
+Extra knobs are available with an expanded syntax:
+
+~~~
+forward FROM TO... {
+    except IGNORED_NAMES...
+    force_tcp
+    health_check DURATION
+    expire DURATION
+    max_fails INTEGER
+    tls CERT KEY CA
+    tls_servername NAME
+    policy random|round_robin
+}
+~~~
+
+* **FROM** and **TO...** as above.
+* **IGNORED_NAMES** in `except` is a space-separated list of domains to exclude from forwarding.
+  Requests that match none of these names will be passed through.
+* `force_tcp`, use TCP even when the request comes in over UDP.
+* `health_checks`, use a different **DURATION** for health checking, the default duration is 0.5s.
+  A value of 0 disables the healthchecks completely.
+* `max_fails` is the number of subsequent failed health checks that are needed before considering
+  a backend to be down. If 0, the backend will never be marked as down. Default is 2.
+* `expire` **DURATION**, expire connections after this time, the default is 10s.
+* `tls` **CERT** **KEY** **CA** define the TLS properties for TLS; if you leave this out the
+  system's configuration will be used.
+* `tls_servername` **NAME** allows you to set a server name in the TLS configuration; for instance 9.9.9.9
+  needs this to be set to `dns.quad9.net`.
+* `policy` specifies the policy to use for selecting upstream servers. The default is `random`.
+
+The upstream selection is done via random selection. If the socket for this client isn't known *forward*
+will randomly choose one. If this turns out to be unhealthy, the next one is tried. If *all* hosts
+are down, we assume healthchecking is broken and select (according to policy) an upstream to try.
+
+Also note the TLS config is "global" for the whole forwarding proxy if you need a different
+`tls-name` for different upstreams you're out of luck.
+
+## Metrics
+
+If monitoring is enabled (via the *prometheus* directive) then the following metric are exported:
+
+* `coredns_forward_request_duration_seconds{to}` - duration per upstream interaction.
+* `coredns_forward_request_count_total{to}` - query count per upstream.
+* `coredns_forward_response_rcode_total{to, rcode}` - count of RCODEs per upstream.
+* `coredns_forward_healthcheck_failure_count_total{to}` - number of failed healthchecks per upstream.
+* `coredns_forward_socket_count_total{to}` - number of cached sockets per upstream.
+
+Where `to` is one of the upstream servers (**TO** from the config), `proto` is the protocol used by
+the incoming query ("tcp" or "udp"), and family the transport family ("1" for IPv4, and "2" for
+IPv6).
+
+## Examples
+
+Proxy all requests within example.org. to a nameserver running on a different port:
+
+~~~ corefile
+example.org {
+    forward . 127.0.0.1:9005
+}
+~~~
+
+Load-balance all requests between three resolvers, one of which has a IPv6 address.
+
+~~~ corefile
+. {
+    forward . 10.0.0.10:53 10.0.0.11:1053 [2003::1]:53
+}
+~~~
+
+Forward everything except requests to `example.org`
+
+~~~ corefile
+. {
+    forward . 10.0.0.10:1234 {
+        except example.org
+    }
+}
+~~~
+
+Proxy everything except `example.org` using the host's `resolv.conf`'s nameservers:
+
+~~~ corefile
+. {
+    forward . /etc/resolv.conf {
+        except example.org
+    }
+}
+~~~
+
+Forward to a IPv6 host:
+
+~~~ corefile
+. {
+    forward . [::1]:1053
+}
+~~~
+
+Proxy all requests to 9.9.9.9 using the DNS-over-TLS protocol, and cache every answer for up to 30
+seconds.
+
+~~~ corefile
+. {
+    forward . tls://9.9.9.9 {
+       tls_servername dns.quad9.net
+       health_check 5s
+    }
+    cache 30
+}
+~~~
+
+## Also See
+
+RFC 7858 for DNS over TLS.
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -2,8 +2,8 @@
 
 *forward* facilitates proxying DNS messages to upstream resolvers.
 
-The *forward* plugin is generally faster (~30+%) than *proxy* as it re-uses already openened sockets
-to the upstreams. It supports UDP, TCP and DNS-over-TLS and uses inband healthchecking that is
+The *forward* plugin is generally faster (~30+%) than *proxy* as it re-uses already opened sockets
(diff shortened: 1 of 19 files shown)


miekg:
Inclusion as such is relative harmless, but this code hasn't been reviewed.
(/cc @grobie @johnbelamaric @pmoroney)

plugin/forward/README.md
```diff
@@ -0,0 +1,141 @@
...
+    tls CERT KEY CA
+    tls_servername NAME
+}
+~~~
+
+* **FROM** and **TO...** as above.
+* **IGNORED_NAMES** in `except` is a space-separated list of domains to exclude from forwarding.
+  Requests that match none of these names will be passed through.
+* `force_tcp`, use TCP even when the request comes in over UDP.
+* `health_checks`, use a different **DURATION** for health checking, the default duration is 2s.
```
pmoroney: I believe the default duration is 0.5s

plugin/forward/forward.go
```diff
@@ -0,0 +1,160 @@
...
+
+	for _, proxy := range f.list() {
+		if proxy.Down(f.maxfails) {
+			fails++
+			if fails < len(f.proxies) {
+				continue
+			}
+			// All upstream proxies are dead, assume healtcheck is complete broken and randomly
+			// select an upstream to connect to.
+			proxy = f.list()[0]
```
pmoroney: If proxy is already random, because f.list is random, why do we re-select here? Couldn't we just continue with the last proxy?

plugin/forward/forward.go
```diff
@@ -0,0 +1,160 @@
...
+
+		ret, err := proxy.connect(ctx, state, f.forceTCP, true)
+
+		if child != nil {
+			child.Finish()
+		}
+
+		if err != nil {
+			log.Printf("[WARNING] Failed to connect to %s: %s", proxy.host.addr, err)
+			if fails < len(f.proxies) {
```
pmoroney: Should we increment `fails` before this line? If this continues and we get to the last proxy, `fails` might be equal to `len(f.proxies)-1` which would then continue when there is no more proxies to continue to.

I think this block might be replaced with
```
fails++
continue
```
as continue acts like a break anyways if `fails == len(f.proxies)`

plugin/forward/forward.go
```diff
@@ -0,0 +1,160 @@
...
+
+	for _, ignore := range f.ignored {
+		if plugin.Name(ignore).Matches(name) {
+			return false
+		}
+	}
+	return true
+}
+
+// list returns a randomized set of proxies to be used for this client. If the client was
```
pmoroney: I'm not sure what the second sentence of this comment is saying.

plugin/forward/persistent.go
```diff
@@ -0,0 +1,147 @@
...
+				if time.Since(pc.used) < t.host.expire {
+					// found one, remove from pool and return this conn.
+					t.conns[proto] = t.conns[proto][i+1:]
+					t.ret <- connErr{pc.c, nil}
+					continue Wait
+				}
+
+				pc.c.Close()
+			}
+
```
pmoroney: Maybe add a comment about how we only get here if no t.conns are usable and what we are going to do in that case. It took me a while to figure it out.

pmoroney (commented):
Just a few things

plugin/forward/README.md
miekg: done

plugin/forward/forward.go
miekg: my thinking here was, re-select to get a new random one. Using the last one is not a bad idea, but may if we support policies it would be more sense to select again? [I could be convinced to use the last one)
(Just pushed some policy code as well)

plugin/forward/forward.go
miekg: you make a good point here.

I want to keep `fails` separate from healtchecking (if reasonable), it may fail for *this* query, but a new query may go through. HC will tell if upstream is really dead. One of the problems in *proxy* is that these two things overlap, sometimes leading to weird results.

plugin/forward/persistent.go
miekg: done

miekg:
Thanks @pmoroney appreciate your review.

plugin/forward/forward.go
pmoroney: I like the idea of different policies. If we do that, we might want to make this explicitly random since picking a random one if they all fail is a separate concern to which order we try them.
Maybe create a `RandomProxy` function on `Forward`, that way we can skip the ordering done by the policy.

miekg:
>I like the idea of different policies. If we do that, we might want to make this explicitly random since picking a random one if they all fail is a separate concern to which order we try them.
>Maybe create a `RandomProxy` function on `Forward`, that way we can skip the ordering done by the policy.

Good point! I'll implement that (just added policies)

plugin/forward/README.md
```diff
@@ -0,0 +1,145 @@
+# forward
+
+*forward* facilitates proxying DNS messages to upstream resolvers.
+
+The *forward* plugin is generally faster (~30+%) than *proxy* as it re-uses already openened sockets
```
johnbelamaric: s/openened/opened/

plugin/forward/README.md
```diff
@@ -0,0 +1,145 @@
...
+to the upstreams. It supports UDP, TCP and DNS-over-TLS and uses inband healthchecking that is
```
johnbelamaric: .,$s/healthchecking/health checking/g

(last time Ko complained I didn't git the complete command :-P)

plugin/forward/README.md
```diff
@@ -0,0 +1,145 @@
...
+enabled by default.
+When *all* upstreams are down it assumes healtchecking as a mechanism has failed and will try to
```
johnbelamaric: s/healtchecking/health checking/

plugin/forward/README.md
```diff
@@ -0,0 +1,145 @@
...
+    tls_servername NAME
+    policy random|round_robin
+}
+~~~
+
+* **FROM** and **TO...** as above.
+* **IGNORED_NAMES** in `except` is a space-separated list of domains to exclude from forwarding.
+  Requests that match none of these names will be passed through.
+* `force_tcp`, use TCP even when the request comes in over UDP.
+* `health_checks`, use a different **DURATION** for health checking, the default duration is 0.5s.
```
johnbelamaric, line 55: s/`health_checks`/`health_check`/

plugin/forward/README.md
```diff
@@ -0,0 +1,145 @@
...
+  A value of 0 disables the healthchecks completely.
```
johnbelamaric: s/healthchecks/health check/

plugin/forward/README.md
```diff
@@ -0,0 +1,145 @@
...
+* `max_fails` is the number of subsequent failed health checks that are needed before considering
+  a backend to be down. If 0, the backend will never be marked as down. Default is 2.
+* `expire` **DURATION**, expire connections after this time, the default is 10s.
```
johnbelamaric: Can you explain this? Are you saying it re-uses the same connection but only for 10s?

plugin/forward/README.md
```diff
@@ -0,0 +1,145 @@
...
+* `tls_servername` **NAME** allows you to set a server name in the TLS configuration; for instance 9.9.9.9
+  needs this to be set to `dns.quad9.net`.
+* `policy` specifies the policy to use for selecting upstream servers. The default is `random`.
+
+The upstream selection is done via random (default policy) selection. If the socket for this client
+isn't known *forward* will randomly choose one. If this turns out to be unhealthy, the next one is
+tried. If *all* hosts are down, we assume healthchecking is broken and select a *random* upstream to
+try.
+
+Also note the TLS config is "global" for the whole forwarding proxy if you need a different
```
johnbelamaric: Fine for now I suppose but I can see this being a feature request before too long:

```
 tls_servername 9.9.9.9 dns.quad9.net
 tls_servername 10.10.10.0 somethingelse.com
```

plugin/forward/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+Proxy all requests within example.org. to a nameserver running on a different port:
+
+~~~ corefile
+example.org {
+    forward . 127.0.0.1:9005
+}
+~~~
+
+Load-balance all requests between three resolvers, one of which has a IPv6 address.
```
johnbelamaric: s/Load-balance/Load balance/

plugin/forward/README.md
```diff
@@ -0,0 +1,145 @@
...
+       tls_servername dns.quad9.net
+       health_check 5s
+    }
+    cache 30
+}
+~~~
+
+## Also See
+
+RFC 7858 for DNS over TLS.
```
johnbelamaric: Can you link the RFC?

plugin/forward/setup.go
```diff
@@ -0,0 +1,250 @@
...
+			// Double check the port, if e.g. is 53 and the transport is TLS make it 853.
+			// This can be somewhat annoying because you *can't* have TLS on port 53 then.
+			switch protocols[i] {
+			case TLS:
+				h1, p, err := net.SplitHostPort(h)
+				if err != nil {
+					break
+				}
+
+				if p == "53" {
```
johnbelamaric: Kind of magic. Seems like we should instead just log a warning here "Found TLS host on port 53, are you sure you didn't want 853?".

plugin/forward/setup.go
```diff
@@ -0,0 +1,250 @@
...
+		f.ignored = ignore
+	case "max_fails":
+		if !c.NextArg() {
+			return c.ArgErr()
+		}
+		n, err := strconv.Atoi(c.Val())
+		if err != nil {
+			return err
+		}
+		f.maxfails = uint32(n)
```
johnbelamaric: I think instead we should error out on a negative number.

plugin/forward/setup.go
```diff
@@ -0,0 +1,250 @@
...
+	case "health_check":
+		if !c.NextArg() {
+			return c.ArgErr()
+		}
+		dur, err := time.ParseDuration(c.Val())
+		if err != nil {
+			return err
+		}
+		f.hcInterval = dur
```
johnbelamaric: should check for a negative duration

plugin/forward/setup.go
```diff
@@ -0,0 +1,250 @@
...
+		f.tlsServerName = c.Val()
+	case "expire":
+		if !c.NextArg() {
+			return c.ArgErr()
+		}
+		dur, err := time.ParseDuration(c.Val())
+		if err != nil {
+			return err
+		}
+		f.expire = dur
```
johnbelamaric: ditto

johnbelamaric (changes requested):

plugin/forward/README.md
miekg: done

plugin/forward/README.md
miekg: done and for 'healtcheck' as well.

plugin/forward/README.md
miekg: yes. You're proposing to re-use the conn until it break? Or go with a much higher expiration?

plugin/forward/README.md
miekg: true, but indeed fixable with that syntax (and backwards compatible).

Repeated this in a new section (Bugs)

plugin/forward/setup.go
miekg: This is more of a bug in dnsutil.ParseHostPortOrFile that defaults to 53 because it doesn't know about the tls:// and friends (that should be fixed). Hence I need to fixup the port number here.

Updated comment with the above text.

plugin/forward/setup.go
miekg: all done

plugin/forward/setup.go
johnbelamaric: ok

plugin/forward/README.md
johnbelamaric: wanted to make sure i understood what "expired" meant. The cost of a new connection amortized across 10s is miniscule. But we can test with different values and see if it makes a difference.

plugin/forward/README.md
johnbelamaric: Still need to fix this one. It's different from the value above...

plugin/forward/lookup.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, proxy := range f.list() {
+		if proxy.Down(f.maxfails) {
+			fails++
+			if fails < len(f.proxies) {
+				continue
+			}
+			// All upstream proxies are dead, assume healtcheck is complete broken and randomly
+			// select an upstream to connect to.
+			proxy = f.list()[0]
+			log.Printf("[WARNING] All upstreams down, picking random one to connect to %s", proxy.host.addr)
```
johnbelamaric, line 35: Not sure we want this here for *every* query. Imagine 10k QPS while all upstreams are down...

johnbelamaric (approved):

Issue #1450, opened 2018-01-27
Resurrect plugin.md


miekg:
Some plugins docs have been removed from that file. Add that back and extend it if possible.

miekg:
closed via #1451

PR #1451, opened 2018-01-27
plugin.md updates

### LINKED ISSUE #1449: log or fmt.Printf?
Do we use log to output or fmt. Check all plugins and core code and document which it is and converge everything to the correct one

miekg:
Talk about return code, logging and other bits.

Also checked the code for fmt.Printf logging, there is none.

Fixes #1449 #1450

### DIFF
--- a/plugin.md
+++ b/plugin.md
@@ -2,14 +2,14 @@
 
 ## Writing Plugins
 
-The main method that gets called is `ServeDNS` it has three parameters:
+The main method that gets called is `ServeDNS`. It has three parameters:
 
-* a `context.Context`
-* `dns.ResponseWriter` that is, basically, the client's connection.
+* a `context.Context`;
+* `dns.ResponseWriter` that is, basically, the client's connection;
 * `*dns.Msg` the request from the client.
 
-`ServeDNS` returns two values, a response code and an error. The the error is not nil CoreDNS
-will return a generic SERVFAIL to the client. The response code tell CoreDNS if a reply has been
+`ServeDNS` returns two values, a response code and an error. If the error is not nil CoreDNS
+will return a generic SERVFAIL to the client. The response code tells CoreDNS if a reply has been
 written by the plugin chain or not. In the latter case CoreDNS will take care of that.
 
 CoreDNS treats:
@@ -19,11 +19,11 @@ CoreDNS treats:
 * FORMERR (dns.RcodeFormatError)
 * NOTIMP (dns.RcodeNotImplemented)
 
-as special and will then assume *nothing* has written to the client. In all other cases it is
-assumes something has been written to the client (by the plugin).
+as special and will then assume *nothing* has written to the client. In all other cases it assumes
+something has been written to the client (by the plugin).
 
-The [*example*](https://github.com/coredns/example) plugin shows a bare bones implements that can be
-used as a starting point for your plugin.
+The [*example*](https://github.com/coredns/example) plugin shows a bare-bones implementation that
+can be used as a starting point for your plugin.
 
 ## Hooking It Up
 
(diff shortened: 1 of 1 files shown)


plugin.md
```diff
@@ -2,11 +2,42 @@
 
 ## Writing Plugins
 
+The main method that gets called is `ServeDNS` it has three parameters:
```
johnbelamaric: s/`ServeDNS` it has/`ServeDNS`. It has/

plugin.md
```diff
@@ -2,11 +2,42 @@
...
+
+* a `context.Context`
+* `dns.ResponseWriter` that is, basically, the client's connection.
+* `*dns.Msg` the request from the client.
+
```
johnbelamaric: A nit, but I would put a semi-colon at the end of each of the first two bullets (instead of the period in the second one). The period is only on the last bullet.

plugin.md
```diff
@@ -2,11 +2,42 @@
...
+`ServeDNS` returns two values, a response code and an error. The the error is not nil CoreDNS
```
johnbelamaric: s/The/If/

plugin.md
```diff
@@ -2,11 +2,42 @@
...
+will return a generic SERVFAIL to the client. The response code tell CoreDNS if a reply has been
```
johnbelamaric: s/tell/tells/

plugin.md
```diff
@@ -2,11 +2,42 @@
...
+written by the plugin chain or not. In the latter case CoreDNS will take care of that.
+
+CoreDNS treats:
+
+* SERVFAIL (dns.RcodeServerFailure)
+* REFUSED (dns.RcodeRefused)
+* FORMERR (dns.RcodeFormatError)
+* NOTIMP (dns.RcodeNotImplemented)
+
+as special and will then assume *nothing* has written to the client. In all other cases it is
```
johnbelamaric: s/has/has been/

johnbelamaric: also remove the " is" at the end of the line.

plugin.md
```diff
@@ -2,11 +2,42 @@
...
+assumes something has been written to the client (by the plugin).
+
+The [*example*](https://github.com/coredns/example) plugin shows a bare bones implements that can be
```
johnbelamaric: s/bare bones/bare-bones/

johnbelamaric: s/implements/implementation/

plugin.md
```diff
@@ -2,11 +2,42 @@
...
 
 * <https://blog.coredns.io/2017/03/01/how-to-add-plugins-to-coredns/>
 * <https://blog.coredns.io/2016/12/19/writing-plugin-for-coredns/>, slightly older, but useful.
 
+## Logging
+
+If your plugin needs to output a log line you should use the `log` package. CoreDNS does not
+implement log levels. The standard way of outputing is: `log.Printf("[LEVEL] ...")`, and LEVEL
+can be: `INFO`, `WARNING` or `ERROR`.
+
```
johnbelamaric: I would add something like:

In general, logging should be left to the higher layers by returning an error. However, if there is a reason to consume the error but notify the user, then logging in the plugin can be acceptable.

johnbelamaric (changes requested):

miekg:
thanks, all acted upon. PTAL

johnbelamaric (approved):

plugin.md
```diff
@@ -2,11 +2,44 @@
...
+written by the plugin chain or not. In the latter case CoreDNS will take care of that.
+
+CoreDNS treats:
+
+* SERVFAIL (dns.RcodeServerFailure)
+* REFUSED (dns.RcodeRefused)
+* FORMERR (dns.RcodeFormatError)
+* NOTIMP (dns.RcodeNotImplemented)
+
+as special and will then assume *nothing* has written to the client. In all other cases it assumes
```
yuewko: s/has written/has been written/

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin.md upd..." ]
>(BTW, it's John's fault - his original comment should have been `s/has written/has been written/g`   ;-) )

done

PR #1453, opened 2018-01-27
pl/kubernetes: drop the comma parsing

### LINKED ISSUE #1446: Comma-separated endpoint in coredns not working
I'm using the following for the endpoint parameter:

`endpoint https://fqdn1:443,https://fqdn2:443,https://fqdn3:443`

Here's what the logs are showing:

`2018/01/27 00:05:14 [WARNING] Host https://fqdn1:443 health check probe failed: Get http://fqdn1:443/: net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x15\x03\x01\x00\x02\x02"
2018/01/27 00:05:14 [WARNING] Host https://fqdn2:443 health check probe failed: Get http://fqdn2:443/: net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x15\x03\x01\x00\x02\x02"
2018/01/27 00:05:14 [WARNING] Host https://fqdn3:443 health check probe failed: Get http://fqdn3:443/: net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x15\x03\x01\x00\x02\x02"`

It looks as though it were trying to perform a health check using the http scheme (Get http://) whereas 443 is used for secure connection.

It works when only one endpoint is specified.

Thanks

miekg:
Let caddyfile parser give us the token, the comma sep thing also didn't
work.

Fixes #1446

### DIFF
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -30,7 +30,7 @@ all the zones the plugin should be authoritative for.
 ```
 kubernetes [ZONES...] {
     resyncperiod DURATION
-    endpoint URL
+    endpoint URL [URL...]
     tls CERT KEY CACERT
     namespaces NAMESPACE...
     labels EXPRESSION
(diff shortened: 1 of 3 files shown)


yongtang (approved):
LGTM

fturib:
I would like to raise a concern here:

It seems that there was a Integration Test validating the format of the endpoint. 
it is now failing. Can be repared..

However, that means that format was working for K8s plugin (maybe without using Healthcheck)
Therefore we may have users in the field using that format.
=> The corefile will be incompatible when they will upgrade coredns.

**What is our policy on backward compatibility for corefile ?**

johnbelamaric:
Should be backward compatible unless we change the major version number, I believe: https://semver.org

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] pl/kubernetes..." ]
>I would like to raise a concern here:
>
>It seems that there was a Integration Test validating the format of the endpoint.
>it is now failing. Can be repared..
>
>However, that means that format was working for K8s plugin (maybe without using Healthcheck)
>Therefore we may have users in the field using that format.
>=> The corefile will be incompatible when they will upgrade coredns.

I don't understand? Someone tried it; didn't work?

/Miek

--
Miek Gieben

chrisohaver:
The issue with #1446 may be that that the API failover functionality doesn't support https.  Looking at the error in #1446, it doesn't appear to be related to comma vs space delimiting - despite what the title of the issue might suggest.

The ci tests for this feature only test http endpoints, not https.

chrisohaver:
I have created 2 new tests in coredns/ci#29 that show ...

* Secure API connection to a single endpoint = PASS. (as expected)
* Secure API connection to a list of endpoints = FAIL. (as suspected)

The test that was already there is passing (after removing commas from the config) ...
* Insecure API connection to a list of endpoints works = PASS

Issue #1456, opened 2018-01-28
Plugin/Reload side effect - Health keep checking status of dead plugins


fturib:
The reason is not on the Reload plugin itself but rather on some recurrent go functions that does not shutdown properly when the server is shutdint down.

Effect is that Healther is checking ALL plugins that were created for ALL servers started with RELOAD since the first start of CoreDNS.

miekg:
closed via #1585 unless you come up with a test to disprove.

Issue #1455, opened 2018-01-28
Plugin/Reload - incompatible with global var in plugin packages


fturib:
Reloading the corefile is done in 2 steps:
1- verify that file is correct by applying once the configuration in Caddy
2- restart the servers

If the step 1 goes to the end, but with an error in the corefile, then each plugin has been called for 'setup'.
Some plugin are using package global vars (metrics, reload .. at least).
It can happen these variables are reinit with the nes corefile, but as the step1 globally fails, the service is still continuing with the former configuration. But meanwhile, those global vars are inited to the new values .... which leads to impredictable beahvior.

NOTE: that is very true with the plugin RELOAD itself that save the data in a global var (instead in configPlugin ... ). Which make the plugin confused after a reload that failed. It is not able to reload on the next period.

fturib:
Found it while using this plugin for my tests on proxy ...
I am working on finding a solution.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Plugin/Reload..." ]
>Found it while using this plugin for my tests on proxy ...
>I am working on finding a solution.

Ok, otherwise we can delay its use by removing it from plugin.cfg ?

johnbelamaric:
This likely applies to all reloads, via this plugin or signals. I did test this particular plugin in that circumstance, though, so I am surprised to see this.

miekg:
Is this what you're trying to fix in #1485?

miekg:
is this (or was this) a problem with reload? Or do you have specific examples of failing plugin (other than reload)?

fturib:
When I opened this issue, that was to raise a concern about the other plugins that does not support RELOAD operation.

I mean, the plugins that may not deallocate / reallocate the resources when the server is restarted. 

Most concerning are the "global" plugins, as they usually allocate resources in global way - global variable for example, and/or opened an http server on specific address that has to be closed on the right time.

I guess we should verify on each plugin:
* are the events balanced for allocation/deallocation - in other words what happen "onStart", "onShutdown", "onRestart", "onFinalShutdown".
* is there some global variable that may be accessed simultaneously by the server still running and the new server that is starting (due to graceful restart ...). Maybe that needs mutex, or redesign if the RESTART was not thought at first time.


NOTE: since then, some UT on RELOAD itself shows needs of mutex around some of the properties that can be accessed by several threads ... (specifically at time of restart). 
the fixes are in this PR : #1485

miekg:
[...]

PR #1457, opened 2018-01-28
Plugin/RELOAD - Tune usage of var global, add limit to options

fturib:
Tune the plugin RELOAD
- found that the RELOAD was going inactive after load of invalid corefile. That may depend where is the error on the corefile (I mean if the setup for RELOAD was fully completed or not)
- add limits to the parameter to avoid issues with 0.

fix partially #1455 

modified the README for minimal values. 

**IMPORTANT**: There is still an issue at shutdown: with no RELOAD, the server stop nicely, with several reload (and thus restart), the server does not stop nicely. I guess it needs to revisit the go function that runs in background and that should stop. Maybe link to #1456

### DIFF
--- a/plugin/reload/setup.go
+++ b/plugin/reload/setup.go
@@ -1,7 +1,7 @@
 package reload
 
 import (
-	"errors"
+	"fmt"
 	"math/rand"
 	"sync"
 	"time"
@@ -22,7 +22,7 @@ func init() {
 // it is used to transmit data between Setup and start of the hook called 'onInstanceStartup'
 // channel for QUIT is never changed in purpose.
 // WARNING: this data may be unsync after an invalid attempt of reload Corefile.
-var r = reload{interval: defaultInterval, stopped: true, quit: make(chan bool)}
+var r = reload{interval: defaultInterval, usage: unused, quit: make(chan bool)}
 var once sync.Once
 
 func setup(c *caddy.Controller) error {
@@ -42,7 +42,7 @@ func setup(c *caddy.Controller) error {
 		i = d
 	}
 	if i < minInterval {
-		return plugin.Error("reload", errors.New("interval value must be greater or equal to 2 sec"))
+		return plugin.Error("reload", fmt.Errorf("interval value must be greater or equal to %v", minInterval))
 	}
 
 	j := defaultJitter
@@ -54,7 +54,7 @@ func setup(c *caddy.Controller) error {
 		j = d
 	}
 	if j < minJitter {
-		return plugin.Error("reload", errors.New("jitter value must be greater or equal to 1 sec"))
+		return plugin.Error("reload", fmt.Errorf("jitter value must be greater or equal to %v", minJitter))
 	}
 
 	if j > i/2 {
@@ -66,7 +66,7 @@ func setup(c *caddy.Controller) error {
 
 	// prepare info for next onInstanceStartup event
 	r.interval = i
-	r.stopped = false
+	r.usage = used
 
 	once.Do(func() {
 		caddy.RegisterEventHook("reload", hook)
--- a/plugin/reload/setup.go
+++ b/plugin/reload/setup.go
@@ -1,6 +1,7 @@
 package reload
 
 import (
+	"errors"
 	"math/rand"
 	"sync"
 	"time"
@@ -17,7 +18,11 @@ func init() {
 	})
 }
 
-var r *reload
+// the info reload is global to all application, whatever number of reloads.
+// it is used to transmit data between Setup and start of the hook called 'onInstanceStartup'
+// channel for QUIT is never changed in purpose.
+// WARNING: this data may be unsync after an invalid attempt of reload Corefile.
+var r = reload{interval: defaultInterval, stopped: true, quit: make(chan bool)}
 var once sync.Once
 
 func setup(c *caddy.Controller) error {
@@ -32,18 +37,24 @@ func setup(c *caddy.Controller) error {
 	if len(args) > 0 {
 		d, err := time.ParseDuration(args[0])
 		if err != nil {
-			return err
+			return plugin.Error("reload", err)
 		}
 		i = d
+	}
+	if i < minInterval {
+		return plugin.Error("reload", errors.New("interval value must be greater or equal to 2 sec"))
 	}
 
 	j := defaultJitter
 	if len(args) > 1 {
 		d, err := time.ParseDuration(args[1])
 		if err != nil {
-			return err
+			return plugin.Error("reload", err)
 		}
 		j = d
+	}
+	if j < minJitter {
+		return plugin.Error("reload", errors.New("jitter value must be greater or equal to 1 sec"))
 	}
 
 	if j > i/2 {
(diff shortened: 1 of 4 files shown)


fturib:
Why is the Codecov failing on a commit that does not contain code ?

miekg:
Why don't we just send the reload signal to ourself?

miekg:
This seems to work fine (note: little testing)
~~~
diff --git a/plugin/reload/reload.go b/plugin/reload/reload.go
index e469b452..35b41997 100644
--- a/plugin/reload/reload.go
+++ b/plugin/reload/reload.go
@@ -1,20 +1,21 @@
+// Package reload periodically checks if the Corefile has changed, and sends the reload signal
+// signal (SIGUSR1) to itself to reload.
 package reload
 
 import (
 	"crypto/md5"
 	"log"
+	"os"
+	"syscall"
 	"time"
 
 	"github.com/mholt/caddy"
 )
 
-// reload periodically checks if the Corefile has changed, and reloads if so
-
 type reload struct {
 	instance *caddy.Instance
 	interval time.Duration
 	sum      [md5.Size]byte
-	stopped  bool
 	quit     chan bool
 }
 
@@ -23,13 +24,6 @@ func hook(event caddy.EventName, info interface{}) error {
 		return nil
 	}
 
-	// if reload is removed from the Corefile, then the hook
-	// is still registered but setup is never called again
-	// so we need a flag to tell us not to reload
-	if r.stopped {
-		return nil
-	}
-
 	// this should be an instance. ok to panic if not
 	r.instance = info.(*caddy.Instance)
 	r.sum = md5.Sum(r.instance.Caddyfile().Body())
@@ -46,13 +40,12 @@ func hook(event caddy.EventName, info interface{}) error {
 				}
 				s := md5.Sum(corefile.Body())
 				if s != r.sum {
-					_, err := r.instance.Restart(corefile)
+					pid := os.Getpid()
+					err := syscall.Kill(pid, syscall.SIGUSR1)
 					if err != nil {
-						log.Printf("[ERROR] Corefile changed but reload failed: %s\n", err)
+						log.Printf("[ERROR] Corefile changed but failed to sent SIGUSR1 to %d: %s\n", pid, err)
 						continue
 					}
-					// we are done, this hook gets called again with new instance
-					r.stopped = true
 					return
 				}
 			case <-r.quit:
~~~

fturib:
I think that the issue is not really on the Restart itself.
So you way to do the restart should work.


Issue is more that in the loop to loadCqaddyFile you depend of the value "r.instance" that is a global var.

When you restart, at setup time RELOAD will re-initialized this r variable with a null instance.
if during the same setup, a directive that is processed AFTER reload is failing .. then r will stay with a null instance, although the server is not restarted (beacause of failing setup.)

on next loop, you will use r.instance to load your corefile . .. but r.instance is NIL.


from there my wondering about global vars : some can be recomputed during the setup process, but that new setup may not be globally successful and the service will continue (without restarting), but with some global values that may not correspond to the right corefile.

fturib:
I can reproduce the case using this corefile:

```
.:5354 {
     errors
     health :8091
     proxy . 2.2.2.2:53 {
        fail_timeout 9s
        max_fails 1
     }
     log
     reload 2s
}
```

run coredns -> all is ok.

```
API server listening at: 127.0.0.1:51513
.:5354
2018/01/28 14:02:00 [INFO] CoreDNS-1.0.5
2018/01/28 14:02:00 [INFO] darwin/amd64, go1.9.2, 
CoreDNS-1.0.5
darwin/amd64, go1.9.2, 
```

then modify the corefile, changing "max_fails 1" to "max_fails 1s" and save the corefile

```
2018/01/28 14:02:05 [INFO] Reloading
2018/01/28 14:02:05 [ERROR] Corefile changed but reload failed: plugin/proxy: strconv.Atoi: parsing "1s": invalid syntax
```

then try to change back to : "max_fails 1"
=> Nothing happen, the loop of the hook is lost ...

running directly on mac (I mean without the debugger), I get:
(in fact the r.instance is nil)

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x1eb47c2]

goroutine 76 [running]:
github.com/mholt/caddy.(*Instance).Caddyfile(...)
        /Users/ftur/go/src/github.com/mholt/caddy/caddy.go:448
github.com/coredns/coredns/plugin/reload.hook.func1()
        /Users/ftur/go/src/github.com/coredns/coredns/plugin/reload/reload.go:43 +0x132
created by github.com/coredns/coredns/plugin/reload.hook
        /Users/ftur/go/src/github.com/coredns/coredns/plugin/reload/reload.go:37 +0x12b
ftur-MBP13:coredns ftur$ 
```

fturib:
In fact, this plugin RELOAD is VERY useful, especially if you try to tune your corefile, changing parameters to observe the behavior ...
That is how I found the issue.

fturib:
NOTE: I may have misplaced the 'r.stopped = true'. - need to rethink of it.

@miekg : if we use the SIGUSR1 .. will it work on Windows ? - I am thinking 'no' because SIGUSR1 is posix signal.

miekg:
Graceful reload doesn't work on windows. But good point. Sending signals
opens that can of worms.

plugin/reload/setup.go
```diff
@@ -32,19 +37,25 @@ func setup(c *caddy.Controller) error {
 	if len(args) > 0 {
 		d, err := time.ParseDuration(args[0])
 		if err != nil {
-			return err
+			return plugin.Error("reload", err)
 		}
 		i = d
 	}
+	if i < minInterval {
+		return plugin.Error("reload", errors.New("interval value must be greater or equal to 2 sec"))
```
johnbelamaric: fmt.Errorf("interval values must be greater or equal to %v", minInterval)

plugin/reload/setup.go
```diff
@@ -32,19 +37,25 @@ func setup(c *caddy.Controller) error {
...
 	if len(args) > 1 {
 		d, err := time.ParseDuration(args[1])
 		if err != nil {
-			return err
+			return plugin.Error("reload", err)
 		}
 		j = d
 	}
+	if j < minJitter {
+		return plugin.Error("reload", errors.New("jitter value must be greater or equal to 1 sec"))
```
johnbelamaric: ditto

johnbelamaric (changes requested):
Ok, this makes sense. Minor nit.

johnbelamaric:
By the way, since this has some odd edge conditions, you should build integration tests for it. You can watch the log to see if the right reloads happen...

miekg:
can we also unit test this?

johnbelamaric:
I think it's a bit tricky to unit test because it involves server restarts, etc. But maybe.

By the way, the issue with package-level vars in other plugins that this issue raises applies to signal-based restarts too. There is no real difference. We probably want to review all of those, as I am not sure they were all written with the idea of `setup` being called more than once.

johnbelamaric (approved):

johnbelamaric:
Additional integration tests coming in another PR.

PR #1458, opened 2018-01-28
plugin/Proxy - WIP implement Healther interface

### LINKED ISSUE #1456: Plugin/Reload side effect - Health keep checking status of dead plugins

fturib:
implement Healther interface for Proxy plugin.

a Proxy bloc will report its status to plugin HEALTH if:
- the option 'report_health' is set in the stanza (in order to not disrupt installations that does not report the health)
- the Exchanger report a connection valid (last use of dial was not failing)
- at least one of the Host is heathy (not down).
NOTE: the health of each host will follow usual rules (max-fails, timeout, check remote health .. )

fix #1456

proxy.README is updated

NOTE: I needed to workaround a weird timeout usecase in proxy.go. in order to not disrupt installations I workaround only if the option 'report_health' is set.

NOTE: I modified the workflow for up and down the Health http listener for reporting. That is to be consistent with RELOAD: we need to take care of new configuration.
the drawback is that the server will be interrupted if an invalid corefile is presented, until a valid one is processed and the instance restarted.

### DIFF
--- a/plugin/proxy/README.md
+++ b/plugin/proxy/README.md
@@ -29,6 +29,7 @@ proxy FROM TO... {
     fail_timeout DURATION
     max_fails INTEGER
     health_check PATH:PORT [DURATION]
+    report_health
     except IGNORED_NAMES...
     spray
     protocol [dns [force_tcp]|https_google [bootstrap ADDRESS...]|grpc [insecure|CACERT|KEY CERT|KEY CERT CACERT]]
@@ -49,6 +50,9 @@ proxy FROM TO... {
   200-399, then that backend is marked healthy for double the healthcheck duration.  If it doesn't,
   it is marked as unhealthy and no requests are routed to it.  If this option is not provided then
   health checks are disabled.  The default duration is 4 seconds ("4s").
+* `report_health` enable the inclusion of this proxy bloc as part of the global healh of CoreDNS 
+  reported by plugin HEALTH. The proxy bloc report OK if the connection is not in fail state and if at 
+  least one of the destination endpoint report itself as not down.    
 * **IGNORED_NAMES** in `except` is a space-separated list of domains to exclude from proxying.
   Requests that match none of these names will be passed through.
 * `spray` when all backends are unhealthy, randomly pick one to send the traffic to. (This is
--- a/plugin/health/health.go
+++ b/plugin/health/health.go
@@ -39,33 +39,42 @@ func (h *health) OnStartup() error {
 		h.Addr = defAddr
 	}
 
-	once.Do(func() {
-		ln, err := net.Listen("tcp", h.Addr)
-		if err != nil {
-			log.Printf("[ERROR] Failed to start health handler: %s", err)
+	ln, err := net.Listen("tcp", h.Addr)
+	if err != nil {
+		log.Printf("[ERROR] Failed to start health handler: %s", err)
+		return err
+	}
+
+	h.ln = ln
+
+	h.mux = http.NewServeMux()
+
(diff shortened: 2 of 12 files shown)


fturib:
There is another test that is disrupted by the change  ... need to dive into.

plugin/proxy/README.md
```diff
@@ -49,6 +50,9 @@ proxy FROM TO... {
   200-399, then that backend is marked healthy for double the healthcheck duration.  If it doesn't,
   it is marked as unhealthy and no requests are routed to it.  If this option is not provided then
   health checks are disabled.  The default duration is 4 seconds ("4s").
+* `report_health` enable the inclusion of this proxy bloc as part of the global healh of CoreDNS
```
miekg, line 53: It doesn't work like this. If you implement health.Healther the plugin's health is include in the global health.

fturib: Well, yes but that mean that everyone that have already a corefile with 'health' will have an immediate impact at upgrade.

this option would permit:
1- to have a ramp-up at upgrade
2- finally to have a fine grain selecting which proxy bloc should be added in the overall health status.

I am open for rollback this part if it is not considered worth it. let me know.

fturib: NOTE: I am using that flag also to workaround a weird timeout test in proxy.go. 
Can you provide your opinion there ? Thanks!

miekg: timeout test should be root-caused (time permitting).

I don't think we should push more into proxy at this time and work on making forward a better proxy. Reporting health in forward is 2 to 5 lines as the "all upstreams are dead" is handled in the foward ServeDNS handler and not sprinkled across the protocol implementations.

And on third thought: I think doing health for proxies is a bad idea. You're tying the process health to some upstream entity that we don't control; that a global outage of all your proxies right there.

fturib:
Will need to rebase and resolve after #1441 will be merged.

plugin/proxy/README.md
fturib: I created an issue #1474 for FORWARD. 
Please have a look at proposition of implementation. 

I will need policy option on that plugin. (Issue is already opened).

Thanks!

fturib: OOOPS .. I forgot an issue. 
I cannot witch to FORWARD without gRPC.!

fturib: and also that "Dnstap" is supported ...

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/Proxy ..." ]
>
>
>
>> @@ -49,6 +50,9 @@ proxy FROM TO... {
>   200-399, then that backend is marked healthy for double the healthcheck duration.  If it doesn't,
>   it is marked as unhealthy and no requests are routed to it.  If this option is not provided then
>   health checks are disabled.  The default duration is 4 seconds ("4s").
>+* `report_health` enable the inclusion of this proxy bloc as part of the global healh of CoreDNS

That you implemented this option means this is not a correct way forward.

IOW: I'm NACKING this PR: we don't implement healther for the proxy/forward.

Health should be a local decision and not involve the network. (Yes we almost/sort 
of do something similar for k8s - but I believe the semantics are slightly 
different.

fturib:
Based on Miek's feedback about implementation of Healther, this PR gets useless.
Thanks for reviewa.

PR #1459, opened 2018-01-28
Plugin/Proxy - add new policy always_first to mimic windows dns resolvers

fturib:
fill documentation, add UT and cleanup fmt

Add a new policy 'always_first' for Proxy load balancing : the upstream Host to use is always the first available in order of the list.
Purpose is to mimic basic behavior on windows laptop.

None

updated the README for proxy, along with the man documentation.

NOTE: I had to adapt a UT for policy/health that was flaky on my installation. Reason is that I have always a reply on http port 80.

### DIFF
--- a/man/coredns-proxy.7
+++ b/man/coredns-proxy.7
@@ -38,7 +38,7 @@ However, advanced features including load balancing can be utilized with an expa
 .nf
 
 proxy FROM TO\.\.\. {
-    policy random|least_conn|round_robin|always_first
+    policy random|least_conn|round_robin|first
     fail_timeout DURATION
     max_fails INTEGER
     health_check PATH:PORT [DURATION]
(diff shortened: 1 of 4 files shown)


man/coredns-proxy.7
```diff
@@ -38,7 +38,7 @@ However, advanced features including load balancing can be utilized with an expa
 .nf
 
 proxy FROM TO\.\.\. {
-    policy random|least_conn|round_robin
+    policy random|least_conn|round_robin|always_first
```
johnbelamaric: How about just `first`?

fturib: I am ok with 'first'. I will change for it.

miekg:
Note: I want to move from proxy to forward (or something else *not* proxy) in the near terrm, as I'm now convinced proxy's health checking is wrong (also see the problem @grobie had).

If we want policy to be a thing; we should implement in *forward* as well.

fturib:
@miekg : for the projects in Infoblox, to be able to use forward, I think we would need:
- healther interface on Forward
- policy option in Forward.

Let me finish first this change .. and then consider to report this options to Forward. Thanks.
I guess anyway we will keep the same "pkg/policy", and the same policies.

fturib:
NOTE: there is an issue #1462 for policy in forward

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Plugin/Proxy ..." ]
>@miekg : for the projects in Infoblox, to be able to use forward, I think we would need:
>- healther interface on Forward
>- policy option in Forward.
>
>Let me finish first this change .. and then consider to report this options to Forward. Thanks.
>I guess anyway we will keep the same "pkg/policy", and the same policies.

Current PR #1447 add policy (and implement random and round robin)
healter can be copied from your PR (I guess)

fturib:
Will need to rebase and resolve after #1441 will be merged.

PR #1468, opened 2018-01-30
Don't use standard lib context package

### LINKED ISSUE #1466: plugin/template: cannot use handler as type plugin.Handler in return argument
Test fails:

```
$ go test
# github.com/coredns/coredns/plugin/template
plugin/template/setup.go:33: cannot use handler (type Handler) as type plugin.Handler in return argument:
	Handler does not implement plugin.Handler (wrong type for ServeDNS method)
		have ServeDNS("context".Context, dns.ResponseWriter, *dns.Msg) (int, error)
		want ServeDNS("golang.org/x/net/context".Context, dns.ResponseWriter, *dns.Msg) (int, error)
```

miekg:
With Go 1.9 you *can* include the std lib's context package and nothing
breaks. However we never officially made the move (and grpc also doesn't
ues the std lib's one).

Standardize all plugins on using the external context package.

Fixes #1466

### DIFF
--- a/plugin/etcd/lookup_test.go
+++ b/plugin/etcd/lookup_test.go
@@ -3,7 +3,6 @@
 package etcd
 
 import (
-	"context"
 	"encoding/json"
 	"testing"
 	"time"
@@ -16,6 +15,7 @@ import (
 
 	etcdc "github.com/coreos/etcd/client"
 	"github.com/miekg/dns"
+	"golang.org/x/net/context"
 )
 
 func init() {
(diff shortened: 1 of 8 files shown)


johnbelamaric:
Seems like we should move to the std lib pkg eventually though.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Don't use sta..." ]
>Seems like we should move to the std lib pkg eventually though.

Yes.

At the time we couldn't because grpc (or some other dependency didn't do it 
yet?).

Issue #1454, opened 2018-01-27
revisit MAINTAINERS


miekg:
Check if the current set of maintainers is still up to date.

stp-ip:
While doing that we might go with the more general OWNER files similar to Kubernetes. Would also give the option to have an owner for plugins etc.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] revisit MAINT..." ]
>While doing that we might go with the more general OWNER files similar to Kubernetes. Would also give the option to have an owner for plugins etc.

Good plan! Was pondering a bit how to do this for plugins.

stp-ip:
fyi: https://github.com/kubernetes/community/blob/master/contributors/devel/owners.md

miekg:
I'll try to use some git magic to come up with a PR. Wish there was an easy way to use some github integration that uses OWNERS.

https://github.com/kubernetes/test-infra/tree/master/prow is mentioned, but I have no idea how to set that up.

stp-ip:
Haven't worked with the test-infra specifically. But yeah prow is the main infra used for providing commands such as /approve /assign /lgtm and it basically just needs a kubernetes cluster and github account and secrets.

miekg:
that almost sounds doable :-)

stp-ip:
Mungegithub was used before for /approve, but that was deprecated afaik, maybe one of the awesome people from test-infra can help especially with where the current docs differ. cc @spiffxp @fejta @cblecker (thanks for any help. really appreciated)

miekg:
This has been done. I'll need to look into setting up the whole framework for actually assigning owners to bugs and PRs.

PR #1469, opened 2018-01-30
Add OWNERS

miekg:
This adds the initial top-level OWNERS file. Each plugin should also get
one, but that will be more work.

Ref: #1454


yongtang (approved):
LGTM

PR #1470, opened 2018-01-30
Remove hooks directory

miekg:
This was used for docker automation or someting?
I don't think there are uses for this, lets remove it.


yongtang (approved):
LGTM

Issue #1478, opened 2018-01-30
Feature: Ability to listen locally IPv4 and Ipv6 in the same instance of CoreDNS


fturib:
### Target
Basically, we would like to support a configuration that target only local DNS request, but for both Ipv4 and Ipv6 stacks.

```
.:53 {
  bind 127.0.0.1
  .. whatever config here ...
  proxy 8.8.8.8:53
}

.:53 {
  bind ::1
  .. whatever config here ...
  proxy 8.8.8.8:53
}
```

Issue: there is an initial validation of the headers of bloc-lists that prevent to listen the same domains on the same port ... *but the listening address is not included*.

The reason is that this check is done *before* the configuration is setup. It is based *only* on the headers of the blocs.

### Proposition

Move up the bind address in the header, in a compatible way with today's syntax.
Something like : <domain>:IP:Port
if the IP is not present, then <domain>:Port would be understood.

NOTE: Bind is still supported, it would overwrite the proposed address in the header.
(no change in the BIND plugin).

### Implementation

I think that it is enough to modify the function that parse and check the header of each ServerBlock in the function InspectServerBlocks of core/dnsserver/register.go

miekg:
This has come up before (has an issue I believe). Extending the zone syntax is not an option; it complicates it and is not very extensible. Multiple IPs will be the next feature request.

fturib:
What is the proposition for defining multiple IP ? Using the BIND plugin ?

miekg:
#1208

fturib:
I agree with #1208 : the extension of bind is not possible as the Server are already build at time of parsing bind (which lead to an error on BIND if it is used twice on different stanza listening the same port)

However, this needs to run several coredns instance, is right now a blocker on Infoblox projects for using coredns as expected.

John B advised me a syntax on the zone that would be compatible with existing syntax.
I mean : whoever want to keep current syntax can continue to use it, who needs to use an extended syntax to specify the listening IP can use it.

current syntax is supported

- . {...} => will listen on *:53
- .:5353 {..} => will listen on *:5353
- myzone.com:53 {..} => listen on *:53

enhanced syntax is also supported

- .:127.0.0.1:54 => will listen on 127.0.0.1:54
- myzone.com:[::1]:78 => will listen on [::1]:78

In coredns, the impact is only on parsing this zone (the Config or server already accept a ListeningAddr) and the sorting of Server per ListeningAddAndPort is already implemented.
[...]

PR #1480, opened 2018-02-02
WIP - Proposition to allow several listening addresses (Ipv4, Ipv6) for same port

fturib:
Enhance the syntax of zone - headers of Blocs of directives, in a way that is backward compatible with current syntax:   `<zone>:<port> {...}` or `<zone>:<ip>:<port> {...}`

Initial syntax is still available ...

```
.:53 {
  proxy 8.8.8.8:53
}

example.com sample.com {
   log
   rewrite ....
   proxy 8.8.8.8:53
}
```

With the new enhanced syntax, one can now write a single bloc to listen on localhost for both Ipv4 and Ipv6.

```
.:127.0.0.1:53 .:[::1]:53 {
  log
  errors
  proxy 8.8.8.8:53
}
```

or use different way to process w request depending the source of the request (per interface, or localhost)

```
.:187.12.0.1:53 {
    log
    errors
    etcd ...
}

.:127.0.0.1:53 .:[::1]:53 {
  proxy 8.8.8.8:53
}
```
NOTE: it is not possible to listen on the same port with a unicast and  multicast address.

changes:
- normalizeZone - to detect port and ip (instead of only port formerly), host validation and reverse one is unchanged 
- register/InspectServerBlocks - add check overlapping between multicast listening and unicast listening, for a same zone, same port.
- address - extended zoneAddr definition and add zoneAddrChecker to simplify the check overlapping.

NOTE: once the listeningAddr is detected, all servers are starting as expected. Caddy is already organized to take into account this listening address in the zone syntax.

Added corresponding UT or Integration tests.

#1478 

Modified the README to inform of this enhanced syntax

### DIFF
--- a/core/dnsserver/address.go
+++ b/core/dnsserver/address.go
@@ -2,59 +2,112 @@ package dnsserver
 
 import (
 	"net"
+	"strconv"
 	"strings"
 
+	"fmt"
 	"github.com/coredns/coredns/plugin"
-
 	"github.com/miekg/dns"
+	"sort"
 )
 
-type zoneAddr struct {
-	Zone      string
-	Port      string
-	Transport string     // dns, tls or grpc
-	IPNet     *net.IPNet // if reverse zone this hold the IPNet
+//ZoneAddr is the memory structure representation of Keys for ServerBlocs
+// use String() to get a print representation
+// use asKey() if you need to use this ZoneAddr in a map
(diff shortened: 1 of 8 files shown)


fturib:
I change to WIP as there is some talk on the issue itself on what syntax using to implement the listening on several IPs.

see #1478

fturib:
I made another implementation that does not change the syntax on ZONES, but allow the plugin BIND to accept several addresses for the listening servers.

Will create a separated PR.

PR #1481, opened 2018-02-02
Liveness

miekg:

### DIFF
--- a/plugin/pkg/up/up_test.go
+++ b/plugin/pkg/up/up_test.go
@@ -1,0 +1,40 @@
+package up
+
+import (
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+)
+
+func TestUp(t *testing.T) {
+	pr := New()
+	wg := sync.WaitGroup{}
+	hits := int32(0)
+
+	upfunc := func(s string) bool {
+		atomic.AddInt32(&hits, 1)
+		// Sleep tiny amount so that our other pr.Do() calls hit the lock.
+		time.Sleep(3 * time.Millisecond)
+		wg.Done()
+		return true
+	}
+
+	pr.Start("nonexistent", 5*time.Millisecond)
+	defer pr.Stop()
+
+	// These functions AddInt32 to the same hits variable, but we only want to wait when
+	// upfunc finishes, as that only calls Done() on the waitgroup.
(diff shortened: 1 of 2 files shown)


miekg:
@johnbelamaric this is a generic implementation that can about as a result of our discussion in https://github.com/coredns/coredns/pull/1444#issuecomment-361972297

miekg:
I'll probably rework *forward*'s healtchecking to use this package and see if things actually become more efficient and simpler.

plugin/pkg/up/up_test.go
```diff
@@ -0,0 +1,71 @@
...
+			atomic.AddInt32(&hits, 1)
+		}))
+	defer web.Close()
+
+	pr := New()
+	defer pr.Stop()
+	i := 0
+	wg.Add(1)
+	upfunc := func(s string) bool {
+		http.Get(s)
```
johnbelamaric: your test could be a lot simpler if it didn't actually do an HTTP Get. You wouldn't need the `web` server at all. It doesn't matter to this "up" machinery what the upfunc actually does.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Liveness (#14..." ]
>your test could be a lot simpler if it didn't actually do an HTTP Get. You wouldn't need the `web` server at all. It doesn't matter to this "up" machinery what the upfunc actually does.

true, was more a test if it could actually be used for something like a health 
check.

simplified.

Thanks!

miekg:
All updated.

I still might revert this if I can't make *forward* use this properly or when the resulting code doesn't look all the simpler. For now I think this will make a difference, code wise and the number of HC qps that we send.

johnbelamaric:
Did you push new commits? I still see the HTTP server in there. Not a big deal, but I thought you were going to remove that.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Liveness (#14..." ]
>Did you push new commits? I still see the HTTP server in there. Not a big deal, but I thought you were going to remove that.

git troubles, had to reclone the repo - should be gone though. I'll double check 
once I'm back home (on my laptop)

miekg:
found the missing tests; code updated.

johnbelamaric (approved):

PR #1484, opened 2018-02-02
plugin/kubernetes: Add upstream @self and loop count

chrisohaver:

- Adds `@self` option to upstream for kubernetes and etcd plugins.  `upstream @self` will cause upstream `CNAME` target lookups to be processed by the current server (via recursive re-entry into dnsserver.ServeDNS()).
- Counts re-entries into ServeDNS(), and breaks loop if re-entries > limit

Still To Do:
- ~~Document `@self` option~~
- ~~Integration tests~~

coredns/deployment#55
#1427

Document `@self` option in `upstream` directive in k8s and etcd plugin READMEs.

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -95,7 +95,7 @@ func NewServer(addr string, group []*Config) (*Server, error) {
 func (s *Server) Serve(l net.Listener) error {
 	s.m.Lock()
 	s.server[tcp] = &dns.Server{Listener: l, Net: "tcp", Handler: dns.HandlerFunc(func(w dns.ResponseWriter, r *dns.Msg) {
-		ctx := context.Background()
+		ctx := context.WithValue(context.Background(), Key{}, s)
 		s.ServeDNS(ctx, w, r)
 	})}
 	s.m.Unlock()
@@ -108,7 +108,7 @@ func (s *Server) Serve(l net.Listener) error {
 func (s *Server) ServePacket(p net.PacketConn) error {
 	s.m.Lock()
 	s.server[udp] = &dns.Server{PacketConn: p, Net: "udp", Handler: dns.HandlerFunc(func(w dns.ResponseWriter, r *dns.Msg) {
-		ctx := context.WithValue(context.Background(), ServerKey, s)
+		ctx := context.WithValue(context.Background(), Key{}, s)
 		s.ServeDNS(ctx, w, r)
 	})}
 	s.m.Unlock()
@@ -207,7 +207,7 @@ func (s *Server) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 		return
 	}
 
-	ctx, err := depthCheck(ctx)
+	ctx, err := incrementDepthAndCheck(ctx)
 	if err != nil {
 		DefaultErrorFunc(w, r, dns.RcodeServerFailure)
 		return
@@ -321,33 +321,35 @@ func DefaultErrorFunc(w dns.ResponseWriter, r *dns.Msg, rc int) {
 	w.WriteMsg(answer)
 }
 
-// depthCheck increments the loop counter in the context, and returns an error if
+// incrementDepthAndCheck increments the loop counter in the context, and returns an error if
 // the counter exceeds the max number of re-entries
-func depthCheck(ctx context.Context) (context.Context, error) {
+func incrementDepthAndCheck(ctx context.Context) (context.Context, error) {
 	// Loop counter for self directed lookups
-	loop := ctx.Value(loopKey)
+	loop := ctx.Value(loopKey{})
 	if loop == nil {
-		ctx = context.WithValue(ctx, loopKey, 0)
-	} else {
-		iloop := loop.(int) + 1
-		if iloop > maxreentries {
-			return ctx, fmt.Errorf("too deep")
-		}
-		ctx = context.WithValue(ctx, loopKey, iloop)
+		ctx = context.WithValue(ctx, loopKey{}, 0)
+		return ctx, nil
 	}
+
+	iloop := loop.(int) + 1
+	if iloop > maxreentries {
+		return ctx, fmt.Errorf("too deep")
+	}
+	ctx = context.WithValue(ctx, loopKey{}, iloop)
 	return ctx, nil
 }
-
-type key int
 
 const (
 	tcp          = 0
 	udp          = 1
 	maxreentries = 10
-	// ServerKey is the context key for the current server
-	ServerKey key = 0
-	loopKey   key = 1
 )
+
+// Key is the context key for the current server
+type Key struct{}
+
+// loopKey is the context key for counting self loops
+type loopKey struct{}
 
 // enableChaos is a map with plugin names for which we should open CH class queries as
 // we block these by default.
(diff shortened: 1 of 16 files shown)


miekg:
We should drop the `@self` so we don't introduce a new namespace.

plugin/backend.go
```diff
@@ -18,7 +19,7 @@ type ServiceBackend interface {
 	Reverse(state request.Request, exact bool, opt Options) ([]msg.Service, error)
 
 	// Lookup is used to find records else where.
-	Lookup(state request.Request, name string, typ uint16) (*dns.Msg, error)
+	Lookup(state request.Request, name string, typ uint16, opt Options) (*dns.Msg, error)
```
miekg: Should we fumble context.Context in state? `opt` makes sense as well, but context.Context is not really an option.
opt is also nice because of the symmetry with `Reverse`

core/dnsserver/server.go
```diff
@@ -314,9 +320,32 @@ func DefaultErrorFunc(w dns.ResponseWriter, r *dns.Msg, rc int) {
...
+		iloop := loop.(int) + 1
+		if iloop > maxreentries {
+			return ctx, fmt.Errorf("too deep")
+		}
+		ctx = context.WithValue(ctx, loopKey, iloop)
+	}
+	return ctx, nil
+}
+
+type key int
```
miekg: docs?

core/dnsserver/server.go
```diff
@@ -107,7 +107,7 @@ func (s *Server) Serve(l net.Listener) error {
 func (s *Server) ServePacket(p net.PacketConn) error {
 	s.m.Lock()
 	s.server[udp] = &dns.Server{PacketConn: p, Net: "udp", Handler: dns.HandlerFunc(func(w dns.ResponseWriter, r *dns.Msg) {
-		ctx := context.Background()
+		ctx := context.WithValue(context.Background(), ServerKey, s)
```
miekg: What about tcp?

core/dnsserver/server.go
```diff
@@ -314,9 +320,32 @@ func DefaultErrorFunc(w dns.ResponseWriter, r *dns.Msg, rc int) {
 	w.WriteMsg(answer)
 }
 
+// depthCheck increments the loop counter in the context, and returns an error if
+// the counter exceeds the max number of re-entries
+func depthCheck(ctx context.Context) (context.Context, error) {
+	// Loop counter for self directed lookups
+	loop := ctx.Value(loopKey)
+	if loop == nil {
+		ctx = context.WithValue(ctx, loopKey, 0)
```
miekg: `return ctx, nil` and close the if

core/dnsserver/server.go
```diff
@@ -314,9 +320,32 @@ func DefaultErrorFunc(w dns.ResponseWriter, r *dns.Msg, rc int) {
...
+type key int
+
 const (
-	tcp = 0
-	udp = 1
+	tcp          = 0
+	udp          = 1
+	maxreentries = 10
+	// ServerKey is the context key for the current server
+	ServerKey key = 0
```
miekg: must it be exported?
the full var now reads `dnsserver.ServerKey` think just `Key` read better?

core/dnsserver/server.go
```diff
@@ -314,9 +320,32 @@ func DefaultErrorFunc(w dns.ResponseWriter, r *dns.Msg, rc int) {
 	w.WriteMsg(answer)
 }
 
+// depthCheck increments the loop counter in the context, and returns an error if
+// the counter exceeds the max number of re-entries
+func depthCheck(ctx context.Context) (context.Context, error) {
```
miekg: incrementDepthAndCheck (or something better - depthCheck doesn't convey it is also incrementing)

plugin/backend.go
johnbelamaric: fumble? What are you trying to suggest?

miekg: should request.Request get a context.Context member in the struct.

johnbelamaric: Oh, you mean "bundle" :)

Yeah, that could work. `opt` already existed in some places which is why Chris used it, I don't know if that place uses request.Request or not - @chrisohaver ?

johnbelamaric: Yeah, in all these https://github.com/coredns/coredns/pull/1484/files#diff-347a95a2d36b7b336cb55cc146bd16d6R36 - they have `state` too, so that would work and be less change. But what's the point of Options in all those...

core/dnsserver/server.go
johnbelamaric: I have seen others use a typed, empty struct as the key. That way there is no collision with someone else using the context. See https://github.com/opentracing/opentracing-go/blob/master/gocontext.go#L12

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: Add upst..." ]
>
>
>
>> @@ -18,7 +19,7 @@ type ServiceBackend interface {
> 	Reverse(state request.Request, exact bool, opt Options) ([]msg.Service, error)
>
> 	// Lookup is used to find records else where.
>-	Lookup(state request.Request, name string, typ uint16) (*dns.Msg, error)
>+	Lookup(state request.Request, name string, typ uint16, opt Options) (*dns.Msg, error)
>
>Yeah, in all these https://github.com/coredns/coredns/pull/1484/files#diff-347a95a2d36b7b336cb55cc146bd16d6R36 - they have `state` too, so that would work and be less change. But what's the point of Options in all those...

force_tcp is one of those options. There could be more in the future, having an 
Options struct would make adding those easier and does not break the API.

plugin/backend.go
chrisohaver: yeah - its not really an "option."  I'll add it to request.Request.

chrisohaver:
I seem to have botched up my fork's master when testing the docker release.

I'll try to restore it.

chrisohaver:
Testing new upstream integration tests coredns/ci#42: /integration-cipr42

chrisohaver:
Integration tests added in coredns/ci#42, will merge that once this merges.

request/request.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/coredns/coredns/plugin/pkg/edns"
 
+	"context"
```
johnbelamaric: We use the external context package everywhere else and should use it here too.

plugin/backend.go
```diff
@@ -45,4 +46,5 @@ type Transferer interface {
 }
 
 // Options are extra options that can be specified for a lookup.
-type Options struct{}
+type Options struct {
```
johnbelamaric: can be on one line, right?

johnbelamaric (changes requested):

johnbelamaric (approved):

johnbelamaric:
@miekg PTAL, we need to merge this soon as K8s deadlines are coming up (we'll need 1.0.6 this week)

johnbelamaric:
FYI, I believe we added `context` to the `state` for this. We may want to reconsider this at some point, given the explicit go recommendation NOT to do this. See https://golang.org/pkg/context/

We would have to make a breaking change to put Context as the first arg in all these calls. Maybe a 1.1 thing?

miekg:
Hmmm..good point. Let's revert some commits then...

johnbelamaric:
No, I think we should go with this for 1.0.6 and then modify it in 1.1. We can't fix the problem without a breaking change, otherwise. And we need it fixed for k8s.

miekg:
Ok. Clear.

PR #1485, opened 2018-02-02
plugin/reload: Tests for reload plugin

rajansandeep:
Add tests for the reload plugin

None

None

### DIFF
--- a/core/dnsserver/register.go
+++ b/core/dnsserver/register.go
@@ -20,13 +20,14 @@ const serverType = "dns"
 // wise they potentially clash with other server types.
 func init() {
 	flag.StringVar(&Port, serverType+".port", DefaultPort, "Default port")
+	port := Port
 
 	caddy.RegisterServerType(serverType, caddy.ServerType{
 		Directives: func() []string { return Directives },
 		DefaultInput: func() caddy.Input {
 			return caddy.CaddyfileInput{
 				Filepath:       "Corefile",
-				Contents:       []byte(".:" + Port + " {\nwhoami\n}\n"),
+				Contents:       []byte(".:" + port + " {\nwhoami\n}\n"),
 				ServerTypeName: serverType,
 			}
 		},
--- a/plugin/reload/reload.go
+++ b/plugin/reload/reload.go
@@ -21,6 +21,39 @@ type reload struct {
 	quit     chan bool
 }
 
+// The set and get functions are defined to protect from data race conditions.
+// It is safe to access the reload variables from these functions and not directly.
+
+// setInterval sets the reload interval.
+func (r *reload) setInterval(interval time.Duration) {
+	mu.Lock()
+	defer mu.Unlock()
+	r.interval = interval
+}
+
+// setUsage sets the reload usage.
+func (r *reload) setUsage(usage int) {
+	mu.Lock()
+	defer mu.Unlock()
+	r.usage = usage
+}
+
+// Interval gets the reload interval.
+func (r reload) Interval() time.Duration {
+	mu.Lock()
+	defer mu.Unlock()
+	i := r.interval
+	return i
+}
+
+// Usage gets the reload usage.
+func (r reload) Usage() int {
+	mu.Lock()
+	defer mu.Unlock()
+	u := r.usage
+	return u
+}
+
 func hook(event caddy.EventName, info interface{}) error {
 	if event != caddy.InstanceStartupEvent {
 		return nil
@@ -29,7 +62,7 @@ func hook(event caddy.EventName, info interface{}) error {
 	// if reload is removed from the Corefile, then the hook
 	// is still registered but setup is never called again
 	// so we need a flag to tell us not to reload
-	if r.usage == unused {
+	if r.Usage() == unused {
 		return nil
 	}
 
@@ -39,7 +72,7 @@ func hook(event caddy.EventName, info interface{}) error {
 	log.Printf("[INFO] Running configuration MD5 = %x\n", md5sum)
 
 	go func() {
-		tick := time.NewTicker(r.interval)
+		tick := time.NewTicker(r.Interval())
 
 		for {
 			select {
@@ -54,15 +87,15 @@ func hook(event caddy.EventName, info interface{}) error {
 					md5sum = s
 					// now lets consider that plugin will not be reload, unless appear in next config file
 					// change status iof usage will be reset in setup if the plugin appears in config file
-					r.usage = maybeUsed
+					r.setUsage(maybeUsed)
 					_, err := instance.Restart(corefile)
 					if err != nil {
 						log.Printf("[ERROR] Corefile changed but reload failed: %s\n", err)
 						continue
 					}
 					// we are done, if the plugin was not set used, then it is not.
-					if r.usage == maybeUsed {
-						r.usage = unused
+					if r.Usage() == maybeUsed {
+						r.setUsage(unused)
 					}
 					return
 				}
--- a/plugin/reload/setup.go
+++ b/plugin/reload/setup.go
@@ -22,9 +22,12 @@ func init() {
 // it is used to transmit data between Setup and start of the hook called 'onInstanceStartup'
 // channel for QUIT is never changed in purpose.
 // WARNING: this data may be unsync after an invalid attempt of reload Corefile.
-var r = reload{interval: defaultInterval, usage: unused, quit: make(chan bool)}
-var once sync.Once
-var shutOnce sync.Once
+var (
+	r        = reload{interval: defaultInterval, usage: unused, quit: make(chan bool)}
+	once     sync.Once
+	shutOnce sync.Once
+	mu       sync.Mutex
+)
 
 func setup(c *caddy.Controller) error {
 	c.Next() // 'reload'
@@ -66,8 +69,8 @@ func setup(c *caddy.Controller) error {
 	i = i + jitter
 
 	// prepare info for next onInstanceStartup event
-	r.interval = i
-	r.usage = used
+	r.setInterval(i)
+	r.setUsage(used)
 
 	once.Do(func() {
 		caddy.RegisterEventHook("reload", hook)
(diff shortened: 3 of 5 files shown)


test/plugin_reload_test.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	"bytes"
+	"log"
+	"strings"
+	"testing"
+
+	"github.com/mholt/caddy"
+)
+
+func TestPluginReload(t *testing.T) {
```
miekg, line 12: Nit: drop `Plugin` here.

test/plugin_reload_test.go
```diff
@@ -0,0 +1,103 @@
...
+			health localhost:0
+		}
+			`,
+			`.:0 {
+			reload 2s
+			whoami
+			health localhost
+		}
+			`,
+			"Corefile successfully loaded",
```
miekg: As you do a strings.Contains I would makes this as short as possible, probably just `successfully` to prevent simple log changes making this test fail.

test/server.go
```diff
@@ -15,17 +15,21 @@ import (
...
 	caddy.Quiet = true
 	dnsserver.Quiet = true
-	log.SetOutput(ioutil.Discard)
-
 	return caddy.Start(NewInput(corefile))
 }
 
+// CoreDNSServer returns a CoreDNS test server. It just takes a normal Corefile as input and discards logging.
+func CoreDNSServer(corefile string) (*caddy.Instance, error) {
+	log.SetOutput(ioutil.Discard)
```
miekg: Just remove this from the original function (or does that make the tests to spammy?) and only leave `CoreDNSServer`

miekg (changes requested):

test/plugin_reload_test.go
rajansandeep: There's another fn with the name `TestReload` and hence I added the word `Plugin`.

test/server.go
rajansandeep: Done.

test/plugin_reload_test.go
miekg: ah, yeah, we're in the global test/ dir; my bad.

miekg (approved):

miekg:
modulo failing checks

miekg:
checks are still failing, can you double check?

rajansandeep:
I enabled the reload plugin for the tests to pass.

test/server.go
```diff
@@ -21,8 +19,6 @@ func CoreDNSServer(corefile string) (*caddy.Instance, error) {
 	defer mu.Unlock()
 	caddy.Quiet = true
 	dnsserver.Quiet = true
-	log.SetOutput(ioutil.Discard)
```
fturib, line 24: This line was useless before this commit ?
I mean : what impact for existing Test that are using this function ?

plugin/reload/reload.go
```diff
@@ -54,7 +56,9 @@ func hook(event caddy.EventName, info interface{}) error {
 					md5sum = s
 					// now lets consider that plugin will not be reload, unless appear in next config file
 					// change status iof usage will be reset in setup if the plugin appears in config file
+					mu.Lock()
 					r.usage = maybeUsed
+					mu.Unlock()
 					_, err := instance.Restart(corefile)
 					if err != nil {
 						log.Printf("[ERROR] Corefile changed but reload failed: %s\n", err)
```
fturib: About the usage of the Mutex.
It is here a very usual issue of sharing the same variable based on struct.

I think would be much proper if you define Getter and Setter on this global variable with mutex inisde these Setter / Getter.
You can group the getting ans setting of the properties as needed.

and add a big comment : do not access out of these Getter/Setter.

core/dnsserver/register.go
```diff
@@ -20,13 +20,14 @@ const serverType = "dns"
...
 
 	caddy.RegisterServerType(serverType, caddy.ServerType{
 		Directives: func() []string { return Directives },
 		DefaultInput: func() caddy.Input {
 			return caddy.CaddyfileInput{
 				Filepath:       "Corefile",
-				Contents:       []byte(".:" + Port + " {\nwhoami\n}\n"),
+				Contents:       []byte(".:" + port + " {\nwhoami\n}\n"),
 				ServerTypeName: serverType,
 			}
```
fturib, line 32: Same thing here : if we want it to be clean. then Setter/Getter around the port variable, with mutex. and a cpmment say it is useful because reload

test/plugin_reload_test.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}
+}
+
+func reloadCorefile(corefile, corefileNew string) string {
+	c, _ := CoreDNSServer(corefile)
+	c, result := verifyRestart(c, corefileNew)
+	c, result = verifyRestart(c, corefile)
+
+	return result
```
fturib: Is there a missing Stop at the end, or I do not follow the logic ?

fturib: I mean, stopping the last instance created by the second restart ?

plugin.cfg
```diff
@@ -20,7 +20,7 @@
 # log:log
 
 tls:tls
-#reload:reload # enable by reverting #1511
+reload:reload
```
fturib: Hum.. I am not sure when we want to setup the 'reload' plugin.
this UT are good.

But I though were link to deeper tests and review compatibility with other plugins.

fturib (changes requested):
Please see comments inside

plugin.cfg
rajansandeep: I have enabled Reload in order for the tests to pass.

test/server.go
rajansandeep: This is redundant and has no impact on existing tests.

plugin/reload/reload.go
```diff
@@ -29,7 +60,7 @@ func hook(event caddy.EventName, info interface{}) error {
 	// if reload is removed from the Corefile, then the hook
 	// is still registered but setup is never called again
 	// so we need a flag to tell us not to reload
-	if r.usage == unused {
+	if r.getUsage() == unused {
```
chrisohaver: The "idiomatic" way of naming getters in Go is not to use "get" in the name.   e.g. `r.Usage()` ...

https://golang.org/doc/effective_go.html#Getters

rajansandeep: Thanks, @chrisohaver.

plugin.cfg
fturib: ok.

test/server.go
fturib: ok

test/plugin_reload_test.go
```diff
@@ -0,0 +1,100 @@
...
+			)
+		}
+	}
+}
+
+func reloadCorefile(corefile, corefileNew string) string {
+	c, _ := CoreDNSServer(corefile)
+	c, result := verifyRestart(c, corefileNew)
+	c, result = verifyRestart(c, corefile)
+	c.Stop()
```
fturib: ok for stop, although you shout make a real StopCoreDNSServer() function that call Stop() and the finalShutdownCallback().
Remember that Load is using that  finalShutDown() and other plugins too.

fturib (commented):
I have still a small concern around the Stop().
All other comments are addressed Thanks!

test/plugin_reload_test.go
rajansandeep: Done.

rajansandeep:
@fturib all changes are good for PR approval?

test/server.go
```diff
@@ -21,13 +19,14 @@ func CoreDNSServer(corefile string) (*caddy.Instance, error) {
...
-
 	return caddy.Start(NewInput(corefile))
 }
 
 // CoreDNSServerStop stops a server.
-func CoreDNSServerStop(i *caddy.Instance) { i.Stop() }
+func CoreDNSServerStop(i *caddy.Instance) {
+	i.Stop()
+	i.ShutdownCallbacks()
```
miekg, line 28: Please remove this. 

These instance starting never exectuted any of these (and some tests work around that), should just add this with this PR

test/plugin_reload_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	log.SetOutput(&buf)
+	coreInput := NewInput(corefile)
+	c, err := c.Restart(coreInput)
+
+	wantMsg := "Running configuration MD5"
+	msg := buf.String()
+	if err != nil {
+		if !strings.Contains(msg, wantMsg) {
+			result = "Unsuccessful"
```
miekg: why is `result` this not a boolean?

plugin/reload/reload.go
```diff
@@ -21,6 +21,54 @@ type reload struct {
...
+
+// setUsage sets the reload usage.
+func (r *reload) setUsage(usage int) {
+	mu.Lock()
+	defer mu.Unlock()
+	r.usage = usage
+}
+
+// setQuit sets the reload quit state.
+func (r *reload) setQuit(quit bool) {
```
miekg: why does sending on a channel need a lock

plugin/reload/setup.go
```diff
@@ -25,6 +25,7 @@ func init() {
 var r = reload{interval: defaultInterval, usage: unused, quit: make(chan bool)}
 var once sync.Once
 var shutOnce sync.Once
+var mu sync.Mutex
```
miekg: why is this global?

and please group these `var ( ... )`

miekg (changes requested):
Sorry, I'm looking again but I feel to see why all the `core/` changes are needed?

Please minimize this PR to the bare necessities

test/server.go
fturib: CoreDNSServerStop was not used so far ...
and if you do not call ShutdownCallback then the onFinalShutdown functions will not be called : the stop is not complete.

miekg:
test/reload_test.go has a bunch of reload tests; I theory those should test the same thing is what this is trying to do. Closing this PR for the time being; i.e. reload tests should be contained to the test/ or plugin/reload

Issue #1448, opened 2018-01-27
plugin/forward: add dnstap support


miekg:
Add dnstap support to forward plugin. This may also require refactoring dnstap to make it easier to use in a plugin.

2buisson:
Eventually I will find something to do about this...

miekg:
closed via #2703

PR #1496, opened 2018-02-07
Easier way to dnstap?

2buisson:

Does making dnstap easier to use in a plugin by putting the error handling outside of the plugin itself.

* https://github.com/coredns/coredns/issues/1448

The Tapper interface in the context has changed.


plugin/dnstap/handler.go
```diff
@@ -60,14 +62,19 @@ func tapMessageTo(w io.Writer, m *tap.Message) error {
...
+
 	h.IO.Dnstap(msg.Wrap(m))
-	return nil
+	return
 }
 
-// TapBuilder implements Tapper.
-func (h Dnstap) TapBuilder() msg.Builder {
-	return msg.Builder{Full: h.Pack}
+func (h Dnstap) Pack() bool {
```
miekg: nit: entire function can be put on a single line 
non nit: docs

plugin/dnstap/handler.go
```diff
@@ -15,9 +15,11 @@ import (
 
 // Dnstap is the dnstap handler.
 type Dnstap struct {
-	Next plugin.Handler
-	IO   IORoutine
-	Pack bool
+	Next           plugin.Handler
+	IO             IORoutine
+	JoinRawMessage bool
```
miekg: JoinRawMessage should probably have some docs explaining what it does when you set it.

plugin/dnstap/msg/msg.go
```diff
@@ -10,100 +9,89 @@ import (
...
-// Builder helps to build Data by being aware of the dnstap plugin configuration.
 type Builder struct {
-	Full bool
-	Data
+	Packed      []byte
+	SocketProto tap.SocketProtocol
+	SocketFam   tap.SocketFamily
+	Address     []byte
+	Port        uint32
+	TimeSec     uint64
```
miekg: Can TimeSec be a `time.Time`?
or does everything here needs to be in a lower level format for easier packing?

Same question for Address (i.e. why not  an `net.IP`?)

And Port, why not just `int`?

plugin/dnstap/msg/msg.go
```diff
@@ -10,100 +9,89 @@ import (
...
-// Data helps to build a dnstap Message.
-// It can be transformed into the actual Message using this package.
-type Data struct {
-	Packed      []byte
-	SocketProto tap.SocketProtocol
-	SocketFam   tap.SocketFamily
-	Address     []byte
-	Port        uint32
-	TimeSec     uint64
+func (b *Builder) Msg(m *dns.Msg) *Builder {
```
miekg: Docs

plugin/dnstap/msg/msg.go
```diff
@@ -10,100 +9,89 @@ import (
...
+		return b
+	}
+
+	b.Packed, b.err = m.Pack()
+	return b
 }
 
-// HostPort decodes into Data any string returned by dnsutil.ParseHostPortOrFile.
-func (d *Data) HostPort(addr string) error {
+func (b *Builder) HostPort(addr string) *Builder {
```
miekg: Docs

miekg (changes requested):
Welcome back! :)

Could you also document (in the README.md?) the easier way to use this?

plugin/dnstap/msg/msg.go
2buisson: Yes it is for the Dnstap struct to take the address of those fields, maybe they shouldn't be exported.

miekg:
You can still use net.IP for instance.

2buisson:
Oops something is wrong do not merge!

2buisson:
Hopefully it's all good now :grin:

miekg:
On holiday, trying to get a review done in the next few days.

miekg (approved):

PR #1497, opened 2018-02-07
plugin/kubernetes: Allow only one k8s section

chrisohaver:

Errors during setup if more than one kubernetes is specified in the same server block.

Currently we allow more than one, but the results are counterintuitive.


miekg (commented):
small setup test?

otherwise lgtm

PR #1498, opened 2018-02-07
core: unblock CH class for forward as well

miekg:
Retweak this a little to make it slightly easier to *not* forget this,
but it is hardly perfect. Should probably make it an interface a plugin
can implement and then unblock if we see that interface.

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -41,7 +41,7 @@ type Server struct {
 }
 
 // NewServer returns a new CoreDNS server and compiles all plugins in to it. By default CH class
-// queries are blocked unless the chaos or proxy is loaded.
+// queries are blocked unless queries from enableChaos are loaded.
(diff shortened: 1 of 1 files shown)


miekg:
cc  @UladzimirTrehubenka

johnbelamaric:
I would think the easiest thing would be to have each plugin implement a "Classes" method that returns the list of supported classes. Then we just loop through them all and aggregate them.

miekg:
I don't want to *force* each plugin to do this (and right now that would be a massive backwards incompatible change anyway). Was thinking more about the `health.Healther` route here.

johnbelamaric:
Yeah, I suppose. I didn't think about external plugins. So then an interface that returns the list of supported classes; if they don't implement it assume only `IN`.

miekg:
filed #1499 for this

Issue #1489, opened 2018-02-05
plugin/rewrite: incorrect D0 option in the REPLY


fturib:
## Description 

on any DNS Server, the REPLY will be filtered differently if the D0 flag of the QUERY is set or not.
(see RFC 4035 , https://www.ietf.org/rfc/rfc4035.txt, 3.2.1.  The DO Bit).

on CoreDNS, we change (sometime) the QUERY by adding a OPT record and set the D0 bit to 1.
That is done in the rewriting process of the QUERY.

The plugin ensure to revert back the QUERY when it is resolved, **but does not take care to re-write the REPLY depending the D0 bit of the initial query**.

## Reproduction case

Corefile to use:
```
.:5356 {
    bind 127.0.0.1
    log
    errors
    proxy . 8.8.8.8:53 {
        protocol dns force_tcp
    }
}
.:5358 {
    log
    errors
    rewrite edns0 local set 0xffee 0xfa123467
    proxy . 8.8.8.8:53 {
        protocol dns force_tcp
    }
}
```

Run the query : (through the rewrite plugin)

$ `dig @127.0.0.1 -p 5358 nxdomaingmail.com`

You will see the reply with all NSSEC records - where it should not
```
; <<>> DiG 9.8.3-P1 <<>> @127.0.0.1 -p 5358 nxdomaingmail.com
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 50989
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 8, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 512
;; QUESTION SECTION:
;nxdomaingmail.com.		IN	A

;; AUTHORITY SECTION:
com.			846	IN	SOA	a.gtld-servers.net. nstld.verisign-grs.com. 1517863063 1800 900 604800 86400
[...]

PR #1500, opened 2018-02-07
plugin/rewrite: remove do set in edns0

rajansandeep:
Removes the DO bit from edns0

#1489


chrisohaver (approved):
LGTM

chrisohaver:
Was going to suggest perhaps adding a test that verifies the DO bit it maintained.
We tested manually, but a unit test would be nice.

Issue #1208, opened 2017-11-08
Cannot define same zones on different listen addresses


yunzheng:
I cannot seem to define the same zones for different listen addresses, e.g.:

```
. {
  bind 10.0.0.1
  proxy . 8.8.8.8
}

. {
  bind 10.0.0.2
  proxy local.network 10.0.0.10
  proxy other.network 192.168.1.1
}
```

It will error with `cannot serve dns://.:53 - zone already defined for dns://.:53`. I would expect I could define a ServeDNS instance separately per listen address/port combination. Is there maybe another way?

miekg:
ah with `bind`, nice twist. Interesting. I agree with you that this should work.

miekg:
The `bind` plugin code runs after the server starts up and does this duplication check. A "fix" is unwieldy. Why can't you just start 2 instances of coredns?

yunzheng:
I currently do, I have like 5 instances now... and each instance takes up like 3% of memory. It kinda pollutes the process list imho, so I was wondering if this was possible at all.

Maybe do it outside the `bind` plugin and make it part of main coredns logic?

miekg:
I don't think "pollutes the process list" is good reason to push for this though.
@mholt is this something that works in Caddy (i.e. did I miss some updates)

yunzheng:
Fair enough, but the memory usage point still stands though. It's currently using 30mb total now for 5 instances. Each `coredns` instance takes up 6mb.

miekg:
How much of that memory is shared?

Anyway I don't believe this can be fixed in a good way - which means you're stuck with running multiple processes. Sorry.

stp-ip:
Not "fixed" in caddy yet. It's more a go limitation.
Issue for caddy: mholt/caddy#864
Issue for go: golang/go#9334

miekg:
Note that *this* issue is different a not related to ipv6

stp-ip:
Yes it just came up during that IPv6 discussion and was discussed within this issue the most. It basically formed into IPv6 not being the issue, but the issue not binding to multiple IPs. ;)

yunzheng:
ok, then i'm closing the issue.

fturib:
It is now implemented : BIND directive accept several addresses.

PR #1501, opened 2018-02-07
WIP - Allow BIND plugin to accept multiple addresses for listening incoming DNS requests

fturib:

Allow BIND to accept multiple addresses, therefore the ServerBloc that includes this BIND directive will be duplicated to allow listening the incoming requests on each addresses provided.

1- The "duplication" part is compatible with the caddy process of the BlocServer and directives.
2- BIND 'looks like' a regular plugin, but now register as a "KeyEnhancer" instead as a Plugin.
 
## Synopsis

Caddy delegate to the ServerType (in our case "dns" of CoreDNS) the responsibility to validate the ServerBlocs and Directives generated by the parsing of the corefile. in CoreDNS languages, it is the place we can reviews and modify the Zones and the Plugins involved, for each ServerBloc.

It is done in the function **InspectServerBlocks** that must be implemented by caddy ServerType's Context. That function, called after the parsing of the corefile - and before any instantiation of listening Servers or setup of Plugin, allow the ServerType to transform as needed these ServerBlocs.
This function is meant to customize the list of Zones, and for each, the list of Plugin to call. it is time to add or remove elements. Nothing is created yet.

This function, for CoreDNS, is implemented in the file dnserver/register.go.

I added a pre-processor to recompute the Zones based on the Plugins declared in the ServerBloc. For instance, BIND is binding the Zone with the address, and if any extra address, BIND will bind a duplicate of the Zone to those extra addresses.

- All Zones, including the duplicates are processed through the usual loop, with an enhanced validator for checking the overlapping zone (multicast address is not compatible with unicast address).

I had to focus on the KeyAddr.String() that define the key that is used to save the Config for each ServerBloc : the duplicate Zones need to have a dedicated key. And the Normalise() functions used by all plugin to extract the host from the ServerBlockKey have to understand that format.

### Choice of implementation

**Code change is in dnsserver (address, register) and plugin/bind** + Unit Tests. Integration tests to come.

Some of the choice can be re-arranged depending this review:

As BIND was already a plugin and there was a comment about 'future extensions' .. I implemented that pre-processor for Zones as kind of "mini-plugin" processor. It is dedicated to transform the Zones:
- I introduced the concepts of KeyEnhancers .. similar to Plugin (but simpler)
=> BIND is registering as keyEnhancer, instead of registering as Plugin. It is invoked for parsing and for computing Zones, but no more for serveDNS.
- each Plugin directive that is acting as 'KeyEnhancer' is removed of the Plugin list before running all Setups. BIND is invoked only as KeyEnhancer. (it is not incompatible to implement both interface for a plugin, though)

_That "mini-plugin" implementation allow several other BIND-like plugins. If we do not expect to have other directive similar to BIND, we may directly include the code of BIND into address.go and remove completely the  plugin, keeping only the README (because it looks like a plugin)._

- I extended ZoneAddr, which is the structured format of a Zone (adding ListeningAddr). 
_Here I also added an Options map for future extension. But it is not really useful for now. If we do not expect to have other mini-plugin like BIND, I maybe can revert_

- I made the ZoneAddrOverlapValidator a dedicated object, as it was a little more complex to compute the overlap than before, and I moved it to address.go. We need it.

- I modified slightly the log for listening servers, so we can understand which one is listening what zone on what address. 
_Here the format of logs can be adapted to wished .._

Example of stanza:

```
.:5354 {
    bind 127.0.0.1 ::1
    log
    errors
    proxy . 8.8.8.8:53 {
        protocol dns force_tcp
    }
}

example.com:5357 sample.com:5358 {
    log
    errors
    proxy . 8.8.8.8:53 {
        protocol dns force_tcp
    }
}
my.domain.com:5356 {
    bind ::1
    log
    errors
    proxy . 8.8.8.8:53 {
        protocol dns force_tcp
    }
    reload 3s
}
```

Corresponding logs ... 

```
.:5354 (->127.0.0.1)
.:5354 (->::1)
example.com.:5357
sample.com.:5358
my.domain.com.:5356 (->::1)
CoreDNS-1.0.5
darwin/amd64, go1.9.2, 

2018/02/07 17:16:43 [INFO] CoreDNS-1.0.5
2018/02/07 17:16:43 [INFO] darwin/amd64, go1.9.2, 
2018/02/07 17:16:43 [INFO] Running configuration MD5 = add2385596aa244249a5319083b3b4ec

```

#1478 and #1208 

BIND documentation was changed to show acceptance of multiple addresses.

### DIFF
--- a/core/dnsserver/address.go
+++ b/core/dnsserver/address.go
@@ -2,59 +2,112 @@ package dnsserver
 
 import (
 	"net"
+	"strconv"
 	"strings"
 
+	"fmt"
 	"github.com/coredns/coredns/plugin"
-
 	"github.com/miekg/dns"
+	"sort"
 )
 
-type zoneAddr struct {
-	Zone      string
-	Port      string
-	Transport string     // dns, tls or grpc
-	IPNet     *net.IPNet // if reverse zone this hold the IPNet
+//ZoneAddr is the memory structure representation of Keys for ServerBlocs
+// use String() to get a print representation
+// use asKey() if you need to use this ZoneAddr in a map
+// use ParseZoneAddr() if you need to retrieve the initial ZoneAddr from the key format
+type ZoneAddr struct {
+	// first 4 are part of the key
+	Transport     string // dns, tls or grpc
+	Zone          string
+	Port          string
+	ListeningAddr string
+	IPNet         *net.IPNet // if reverse zone this hold the IPNet
+	Options       map[string]string
 }
 
-// String return the string representation of z.
-func (z zoneAddr) String() string { return z.Transport + "://" + z.Zone + ":" + z.Port }
+const (
+	separateProtocol  = "://"
+	separatePortAndIP = ":"
+	separateCIDR      = "##"
+	separateOptions   = "#$#"
+	startOption       = "["
+	endOption         = "]"
+	partsSep          = "="
+)
 
-// Transport returns the protocol of the string s
-func Transport(s string) string {
-	switch {
-	case strings.HasPrefix(s, TransportTLS+"://"):
-		return TransportTLS
-	case strings.HasPrefix(s, TransportDNS+"://"):
-		return TransportDNS
-	case strings.HasPrefix(s, TransportGRPC+"://"):
-		return TransportGRPC
+var transports map[int]string
+
+// Copy : provide a duplicate of this ZoneAddr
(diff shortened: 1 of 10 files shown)


fturib:
Replaced by https://github.com/coredns/coredns/pull/1512, and a second part to handle another way the verification of overlapping zones.

PR #1507, opened 2018-02-08
plugin/reload: enable proper coredns instance shutdown

rajansandeep:
This PR enables proper shutdown of CoreDNS instance when using reload plugin more than once.

Example scenario:
```
.:5356 {
    reload 3s
    log
    errors
}
.:5358 {
    reload 3s
    proxy . 8.8.8.8:53 {
        protocol dns force_tcp
    }
}
```
Before Fix (Need to give SIGINT twice):
```
$ ./coredns
.:5356
.:5358
2018/02/08 10:47:50 [INFO] CoreDNS-1.0.5
2018/02/08 10:47:50 [INFO] darwin/amd64, go1.9.3, 
CoreDNS-1.0.5
darwin/amd64, go1.9.3, 
2018/02/08 10:47:50 [INFO] Running configuration MD5 = 8ee3104d51143cb2311f17075628eb5c
^C2018/02/08 10:47:50 [INFO] SIGINT: Shutting down
^C2018/02/08 10:47:52 [INFO] SIGINT: Force quit
$
```

After fix: 
```
$ ./coredns
.:5356
.:5358
2018/02/08 10:48:11 [INFO] CoreDNS-1.0.5
2018/02/08 10:48:11 [INFO] darwin/amd64, go1.9.3, 
CoreDNS-1.0.5
darwin/amd64, go1.9.3, 
2018/02/08 10:48:11 [INFO] Running configuration MD5 = 8ee3104d51143cb2311f17075628eb5c
^C2018/02/08 10:48:12 [INFO] SIGINT: Shutting down
$
```
None

None

### DIFF
--- a/plugin/reload/setup.go
+++ b/plugin/reload/setup.go
@@ -24,6 +24,7 @@ func init() {
 // WARNING: this data may be unsync after an invalid attempt of reload Corefile.
 var r = reload{interval: defaultInterval, usage: unused, quit: make(chan bool)}
 var once sync.Once
+var shutOnce sync.Once
 
 func setup(c *caddy.Controller) error {
 	c.Next() // 'reload'
@@ -73,9 +74,11 @@ func setup(c *caddy.Controller) error {
 	})
 
 	// re-register on finalShutDown as the instance most-likely will be changed
-	c.OnFinalShutdown(func() error {
-		r.quit <- true
-		return nil
+	shutOnce.Do(func() {
+		c.OnFinalShutdown(func() error {
+			r.quit <- true
+			return nil
+		})
 	})
 	return nil
 }


plugin/reload/setup.go
```diff
@@ -73,9 +74,11 @@ func setup(c *caddy.Controller) error {
 	})
 
 	// re-register on finalShutDown as the instance most-likely will be changed
-	c.OnFinalShutdown(func() error {
-		r.quit <- true
-		return nil
+	shutOnce.Do(func() {
```
fturib, line 77: Yes. we want only one QUIT signal to stop the onTimer loop of the last Hook.

fturib: Anyway all these onTimer launched on each reload, even though would not have stopped as expected, are listening the SAME channel. One QUIT is enough.
Two QUIT will block the sender and then the application cannot shutdown properly.

fturib (approved):

johnbelamaric:
Ok, but shouldn't we disallow two `reload` in the file?

miekg:
just use the last one? (or the first?)

miekg:
or maybe we *should* reconsider and making it a flag...

miekg:
should we disable *reload* for the time being, while we work out the issues ? Doing a release with it means we tied it, which may be annoying if we decide to do things differently.

johnbelamaric:
Hmm. Let's move that discussion to the release issue. I will post something there.

fturib:
@johnbelamaric : you may be right. But it is a separate pb. 
We have the issue of several finalShutDown each time we reload (whatever the number of directive 'reload' in the file).

johnbelamaric:
OK. Well, this fixes a problem with the current master, so I suppose we can merge. But we still need to figure out whether we keep it as is or do something else.

plugin/reload/setup.go
johnbelamaric: so now it *doesn't* re-register this, like the comment says?

fturib:
@johnbelamaric : Sandeep just made a test with only one directive 'reload' - without this change in the code, and the shutdown is working as expected.
I am wondering why it works, but looks like the description is correct: it is only when we have several directives that the process does not stop properly.

johnbelamaric:
I meant the comment in the code above, where it was 're-registering' because the instance changed. Is that no longer relevant?

fturib:
Sorry for back and forth .... 
But I made a test on my machine, building with 'make', without this change and with this change - for only one Directive 'reload' in the corefile.
And finally, my opinion is that we need this change for a proper stop.

fturib:
I am running a second test fully from the last master branch ...

fturib:
.. and I am wrong. We do not need for a single directive (a second reload directive was hidden in my corefile)

fturib:
Ok let me clarify:

Today with current code in master branch.
- If a user is using properly the corefile with only one directive RELOAD, coreDNS stop properly
- if a user is using several RELOAD directives in the corefile, then coreDNS need a kill -9 for stop

This PR fix the second case for current master, and I think is better to merge.


If we want to check and fail to start if corefile contains several RELOAD directives, then we should open an issue for next release.

johnbelamaric:
ok, it's better than what's in there now

Issue #1403, opened 2018-01-18
Check and test all plugins for case preservation


miekg:
DNS is case preservation protocol. I.e. a qname that comes in as Example.COM needs to come out like that.

The *file* plugin does some of this right (and has some casing tests IIIRC), but we should check all plugins for this by writing tests for them. 

This is a bit annoying to do but a great way to learn about the coredns code base (and about DNS).

yongtang:
Not familiar with DNS protocol overall though I remember DNS is also case-insensitive. So we may need to be careful to make sure it is both `case-insensitive` and `case-preserving`.

miekg:
it would be nice if this would be a top-level test function that rotates through all plugins - problem there is that you don't know a good default syntax before hand.

Another approach is to have a simple test helper function that does this per plugin as to minimize the code that needs to be added.

miekg:
See #1393

PR #1509, opened 2018-02-08
Add a test case for case insensitive/preserving with whoami plugin

yongtang:
This fix add a test case for case insensitive/preserving with
whoami plugin.

This fix is part of the effort for #1403.


miekg:
lgtm, I would wish we could something generate in test/ which would check all plugins? Rotate through all plugins by making a Corefile per plugin or something?

PR #1511, opened 2018-02-09
plugin/reload: disable for now

miekg:
Reloading should work (kill -TERM reload the coredns process), but a lot
of plugins can't handle it proper. Disable to reload plugin until we fix
(most) of the plugins.


johnbelamaric (approved):

johnbelamaric:
Modulo checks

PR #1512, opened 2018-02-09
Plugin/BIND - extend the syntax to allow multiple addresses

### LINKED ISSUE #1478: Feature: Ability to listen locally IPv4 and Ipv6 in the same instance of CoreDNS

fturib:
Extend BIND plugin to support multiple addresses.

Config is modified : ListenHost is now an arrays of string named ListenHosts
NOTE: any external plugin that was using the former name will fail at compilation. I updated all changes for internal plugins. (only TRACE was using it to build a name of the server)

These addresses are processed in  'register', while computing the number of Server to create.

UT are provided.

We can now support configurations like:

```
.:5364 {
    reload 2s
    bind 127.0.0.1 ::1  10.84.16.153
    log
    errors
    proxy . 8.8.8.8:53 {
        protocol dns force_tcp
    }
}

example.com:5365 {
    log
    errors
    proxy . 8.8.8.8:53 {
        protocol dns force_tcp
    }
}
```

First step to fix #1478 and #1208 

README of BIND has been updated.

### DIFF
--- a/core/dnsserver/config.go
+++ b/core/dnsserver/config.go
@@ -2,6 +2,7 @@ package dnsserver
 
 import (
 	"crypto/tls"
+	"net"
 
 	"github.com/coredns/coredns/plugin"
 
@@ -13,7 +14,8 @@ type Config struct {
 	// The zone of the site.
 	Zone string
 
-	// The hostname to bind listener to, defaults to the wildcard address
+	// one or several hostnames to bind the server to.
+	// defaults to a single empty string that denote the wildcard address
 	ListenHosts []string
 
 	// The port to listen on.
@@ -50,6 +52,22 @@ type Config struct {
 	registry map[string]plugin.Handler
 }
 
+//HostAddresses builds a representation of the addresses of this Config
+//after server is started ONLY, can be used as a Key for identifing that config
+// :53 or 127.0.0.1:53 or 127.0.0.1:53/::1:53
+func (c *Config) HostAddresses() string {
+	hosts := ""
+	for _, h := range c.ListenHosts {
+		host := net.JoinHostPort(h, c.Port)
+		if hosts == "" {
+			hosts = host
+			continue
+		}
+		hosts = hosts + "/" + host
+	}
+	return hosts
+}
+
 // GetConfig gets the Config that corresponds to c.
 // If none exist nil is returned.
 func GetConfig(c *caddy.Controller) *Config {
@@ -60,6 +78,6 @@ func GetConfig(c *caddy.Controller) *Config {
 	// we should only get here during tests because directive
 	// actions typically skip the server blocks where we make
 	// the configs.
-	ctx.saveConfig(c.Key, &Config{})
+	ctx.saveConfig(c.Key, &Config{ListenHosts: []string{""}})
 	return GetConfig(c)
 }
--- a/core/dnsserver/config.go
+++ b/core/dnsserver/config.go
@@ -14,7 +14,7 @@ type Config struct {
 	Zone string
 
 	// The hostname to bind listener to, defaults to the wildcard address
-	ListenHost string
+	ListenHosts []string
 
 	// The port to listen on.
 	Port string
--- a/core/dnsserver/config.go
+++ b/core/dnsserver/config.go
@@ -56,16 +56,16 @@ type Config struct {
 //after server is started ONLY, can be used as a Key for identifing that config
 // :53 or 127.0.0.1:53 or 127.0.0.1:53/::1:53
 func (c *Config) HostAddresses() string {
-	hosts := ""
+	all := ""
 	for _, h := range c.ListenHosts {
-		host := net.JoinHostPort(h, c.Port)
-		if hosts == "" {
-			hosts = host
+		addr := net.JoinHostPort(h, c.Port)
+		if all == "" {
+			all = addr
 			continue
 		}
-		hosts = hosts + "/" + host
+		all = all + "/" + addr
 	}
-	return hosts
+	return all
 }
 
 // GetConfig gets the Config that corresponds to c.
--- a/core/dnsserver/register.go
+++ b/core/dnsserver/register.go
@@ -70,9 +70,10 @@ func (h *dnsContext) InspectServerBlocks(sourceFile string, serverBlocks []caddy
 
 			// Save the config to our master list, and key it for lookups.
 			cfg := &Config{
-				Zone:      za.Zone,
-				Port:      za.Port,
-				Transport: za.Transport,
+				Zone:        za.Zone,
+				Port:        za.Port,
+				Transport:   za.Transport,
+				ListenHosts: []string{""},
 			}
 			if za.IPNet == nil {
 				h.saveConfig(za.String(), cfg)
@@ -191,13 +192,7 @@ func groupConfigsByListenAddr(configs []*Config) (map[string][]*Config, error) {
 	groups := make(map[string][]*Config)
 
 	for _, conf := range configs {
-		hosts := []string{}
-		if conf.ListenHosts == nil {
-			hosts = append(hosts, "")
-		} else {
-			hosts = append(hosts, conf.ListenHosts...)
-		}
-		for _, h := range hosts {
+		for _, h := range conf.ListenHosts {
 			addr, err := net.ResolveTCPAddr("tcp", net.JoinHostPort(h, conf.Port))
 			if err != nil {
 				return nil, err
--- a/core/dnsserver/register.go
+++ b/core/dnsserver/register.go
@@ -191,14 +191,21 @@ func groupConfigsByListenAddr(configs []*Config) (map[string][]*Config, error) {
 	groups := make(map[string][]*Config)
 
 	for _, conf := range configs {
-		addr, err := net.ResolveTCPAddr("tcp", net.JoinHostPort(conf.ListenHost, conf.Port))
-		if err != nil {
-			return nil, err
+		hosts := []string{}
+		if conf.ListenHosts == nil {
+			hosts = append(hosts, "")
+		} else {
+			hosts = append(hosts, conf.ListenHosts...)
 		}
-		addrstr := conf.Transport + "://" + addr.String()
-		groups[addrstr] = append(groups[addrstr], conf)
+		for _, h := range hosts {
+			addr, err := net.ResolveTCPAddr("tcp", net.JoinHostPort(h, conf.Port))
+			if err != nil {
+				return nil, err
+			}
+			addrstr := conf.Transport + "://" + addr.String()
+			groups[addrstr] = append(groups[addrstr], conf)
+		}
 	}
-
 	return groups, nil
 }
 
--- a/core/dnsserver/register_test.go
+++ b/core/dnsserver/register_test.go
@@ -32,57 +32,70 @@ func TestHandlers(t *testing.T) {
 
 func TestGroupingServers(t *testing.T) {
 	for i, test := range []struct {
-		configs []*Config
-		gpes    []string
-		failing bool
+		configs        []*Config
+		expectedGroups []string
+		failing        bool
 	}{
+		// single config -> one group
 		{configs: []*Config{
-			{Transport: "dns", Zone: ".", Port: "53"},
+			{Transport: "dns", Zone: ".", Port: "53", ListenHosts: []string{""}},
 		},
-			gpes:    []string{"dns://:53"},
-			failing: false},
+			expectedGroups: []string{"dns://:53"},
+			failing:        false},
+
+		// 2 configs on different port -> 2 groups
 		{configs: []*Config{
-			{Transport: "dns", Zone: ".", Port: "53"},
-			{Transport: "dns", Zone: ".", Port: "54"},
+			{Transport: "dns", Zone: ".", Port: "53", ListenHosts: []string{""}},
+			{Transport: "dns", Zone: ".", Port: "54", ListenHosts: []string{""}},
 		},
-			gpes:    []string{"dns://:53", "dns://:54"},
-			failing: false},
+			expectedGroups: []string{"dns://:53", "dns://:54"},
+			failing:        false},
+
+		// 2 configs on same port, same broadcast address, diff zones -> 1 group
 		{configs: []*Config{
-			{Transport: "dns", Zone: ".", Port: "53"},
-			{Transport: "dns", Zone: "com.", Port: "53"},
+			{Transport: "dns", Zone: ".", Port: "53", ListenHosts: []string{""}},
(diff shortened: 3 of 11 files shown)


core/dnsserver/register.go
```diff
@@ -191,14 +191,21 @@ func groupConfigsByListenAddr(configs []*Config) (map[string][]*Config, error) {
 	groups := make(map[string][]*Config)
 
 	for _, conf := range configs {
-		addr, err := net.ResolveTCPAddr("tcp", net.JoinHostPort(conf.ListenHost, conf.Port))
-		if err != nil {
-			return nil, err
+		hosts := []string{}
+		if conf.ListenHosts == nil {
```
miekg: when is this `nil`?

core/dnsserver/register_test.go
```diff
@@ -29,3 +29,80 @@ func TestHandlers(t *testing.T) {
 		t.Errorf("Expected [testPlugin] from Handlers, got %v", hs)
 	}
 }
+
+func TestGroupingServers(t *testing.T) {
+	for i, test := range []struct {
+		configs []*Config
+		gpes    []string
```
miekg: what does `gpes` stand for?

core/dnsserver/register_test.go
```diff
@@ -29,3 +29,80 @@ func TestHandlers(t *testing.T) {
...
+		failing bool
+	}{
+		{configs: []*Config{
```
miekg: Do all these test cases really test new functionality? Seems to me you can get a away with 2 or 3 cases here?

plugin/bind/README.md
```diff
@@ -6,16 +6,21 @@
...
+
+If several addresses are provided, a listener will be open on each of the IP provided.
+
+Each address has to be an IP of one of the interfaces of the host.
 
 ## Syntax
 
 ~~~ txt
-bind ADDRESS
+bind ADDRESS [ADDRESS] ...
```
miekg: Think you can write `ADDRESS...`

plugin/bind/README.md
```diff
@@ -6,16 +6,21 @@
...
 ~~~
 
-**ADDRESS** is the IP address to bind to.
+**ADDRESS** is an IP address to bind to.
+When several addresses are provides a listener will be opened on each of the addresses
```
miekg: Missing closing dot

plugin/bind/README.md
```diff
@@ -26,3 +31,11 @@ To make your socket accessible only to that machine, bind to IP 127.0.0.1 (local
     bind 127.0.0.1
 }
 ~~~
+
+To allow processing DNS requests only local host on both Ipv4 and Ipv6 stacks, use the syntax:
```
miekg: IPv4 and IPv6

plugin/trace/setup.go
```diff
@@ -34,14 +34,30 @@ func setup(c *caddy.Controller) error {
...
+	case 1:
+		host = hosts[0]
+	default:
+		host = hosts[0]
+		for _, h := range hosts[1:] {
+			host = host + "," + h
+		}
+		host = "(" + host + ")"
+	}
+	return host + ":" + port
```
miekg: This should actually be net.JoinHostPort

plugin/trace/setup.go
```diff
@@ -34,14 +34,30 @@ func setup(c *caddy.Controller) error {
 	return nil
 }
 
+func endPointName(hosts []string, port string) string {
```
miekg: What does this do ?

plugin/trace/setup.go
```diff
@@ -34,14 +34,30 @@ func setup(c *caddy.Controller) error {
...
+	host := ""
+	switch len(hosts) {
+	case 0:
```
miekg: can this happen?

core/dnsserver/config.go
```diff
@@ -14,7 +14,7 @@ type Config struct {
 	Zone string
 
 	// The hostname to bind listener to, defaults to the wildcard address
```
miekg: comments also needs updating

miekg (changes requested):
This change is good. Bunch of comments and questions.

core/dnsserver/register.go
fturib: if there is no BIND directive. I will change to have a default value of one listenHost with an empty string. will look closer to former data and will be easier to handle.

core/dnsserver/register_test.go
fturib: oops... that stands for 'group' .. (in French .. groupe') - will change

plugin/bind/README.md
fturib: ok.

plugin/bind/README.md
fturib: ok

plugin/trace/setup.go
fturib: JoinHostPort does not support Host or IP to be empty. Here we can have host empty.

plugin/trace/setup.go
fturib: it build a representation of empty, 1 or multiple addresses attached to the same port ..
trace.ServiceEndpoint use at as a name.

can be 
:53
127.0.0.1:53
(127.0.0.1,::1):53

plugin/trace/setup.go
fturib: no more.

fturib: although .. I do not like to make a len(xx) without testing the zero-length.

core/dnsserver/config.go
fturib: I agree.

plugin/bind/README.md
johnbelamaric: s/provides/provided/

core/dnsserver/register_test.go
fturib: I reviewd, eliminates few, but still have 8 uses case, combinations of ports, addresses (unicast, braadcast) and zones. Added comment in front of each.

plugin/bind/README.md
fturib: ok

plugin/trace/setup.go
fturib: not sure how to get better, but we need to distinguish each BlocList with is identified by Zone + bind addresses.

fturib: Right now, used only by Trace plugin. But should be available for all ... it is the way to have an identifier of the Config once the server is started.

plugin/trace/setup.go
miekg: this does the wrong thing for v6 then.. unless `trace` handles that?

plugin/trace/setup.go
miekg: ok, please comment this.

fturib: I changed it : move to Config as HostAddresses() and return something a bit more usual ..

fturib:
I think that is now good for a second review. I addresses comments, also add support for multiple BIND directives in the ServerBlock.
Thanks!

core/dnsserver/config.go
```diff
@@ -50,6 +52,22 @@ type Config struct {
 	registry map[string]plugin.Handler
 }
 
+//HostAddresses builds a representation of the addresses of this Config
+//after server is started ONLY, can be used as a Key for identifing that config
+// :53 or 127.0.0.1:53 or 127.0.0.1:53/::1:53
+func (c *Config) HostAddresses() string {
+	hosts := ""
+	for _, h := range c.ListenHosts {
+		host := net.JoinHostPort(h, c.Port)
```
miekg: JoinHostPort doesn't return an empty string

core/dnsserver/server.go
```diff
@@ -287,8 +287,10 @@ func (s *Server) OnStartupComplete() {
 		return
 	}
 
-	for zone, config := range s.zones {
-		fmt.Println(zone + ":" + config.Port)
+	for zone := range s.zones {
+		// if the server is listening on a specific address let's make it visible in the log,
+		// so one can differentiate between all active listeners
+		fmt.Println(zone + " on " + s.Addr)
```
miekg: Is it possible to keep this backwards compatible? We introduce spaces here.
Right now we have:
~~~
.:1053
~~~
Can we detect we did a `bind` and only extend it when we did?
~~~
.:1053
com.:1053 on 192.168.1.2 172.16.1.15
~~~
Typing that last line. How does this look with multiple `bind` statements or one with multiple addresses?

plugin/bind/setup.go
```diff
@@ -12,13 +12,21 @@ import (
 
 func setupBind(c *caddy.Controller) error {
 	config := dnsserver.GetConfig(c)
+
+	// addresses will be consolidated over all BIND directives available in that BlocServer
+	allAddresses := []string{}
```
miekg: nit: I Go shorter names are generally preferred, would just go with `all` here

core/dnsserver/config.go
fturib: I agree, but not sure why you are doing that remark.
Is it because I test if "hosts" is not empty on the next line ? 
=> mistmatch between host and hosts ?

core/dnsserver/server.go
fturib: At this point the Server is listening only one address (unicast or broadcast) 
NOTE:  the multiple BIND address BlocServers are broken into several Servers within dnsContext.MakeServers, which use  groupConfigsByListenAddr() to split by one address only.

Right now it looks like:
```
com. on dns://:6003
. on dns://127.0.0.1:6001
. on dns://[::1]:6001
```

I will change to get  (for the same servers), keeping  compatible with former version.

```
com.:6003
.:6001 on 127.0.0.1
.:6001 on ::1
```

plugin/bind/setup.go
fturib: ok.

core/dnsserver/server.go
fturib: QUESTION: we have 3 kind of Servers (dns, tls, grpc) .. do we want so show the lkine of server is listening. In other word, do we want so show TLS or gRPC as the listening protocol ?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Plugin/BIND -..." ]
>
>
>
>> @@ -50,6 +52,22 @@ type Config struct {
> 	registry map[string]plugin.Handler
> }
>
>+//HostAddresses builds a representation of the addresses of this Config
>+//after server is started ONLY, can be used as a Key for identifing that config
>+// :53 or 127.0.0.1:53 or 127.0.0.1:53/::1:53
>+func (c *Config) HostAddresses() string {
>+	hosts := ""
>+	for _, h := range c.ListenHosts {
>+		host := net.JoinHostPort(h, c.Port)
>
>I agree, but not sure why you are doing that remark.
>Is it because I test if "hosts" is not empty on the next line ?
>=> mistmatch between host and hosts ?

you specifically check is `host` is empty as returned from net.JoinHostPort. 
That does not make sense because it will not return an empty string; so what are 
you checking here?

core/dnsserver/config.go
```diff
@@ -56,16 +56,16 @@ type Config struct {
...
 // :53 or 127.0.0.1:53 or 127.0.0.1:53/::1:53
 func (c *Config) HostAddresses() string {
-	hosts := ""
+	all := ""
 	for _, h := range c.ListenHosts {
-		host := net.JoinHostPort(h, c.Port)
-		if hosts == "" {
-			hosts = host
+		addr := net.JoinHostPort(h, c.Port)
+		if all == "" {
```
fturib: only changed the names of the variables,.
what I test as empty s "all" in order to correctly insert the separator of addresses.

core/dnsserver/server.go
```diff
@@ -288,9 +288,21 @@ func (s *Server) OnStartupComplete() {
...
+			return
+		}
+		if ip == "" {
+			fmt.Println(zone + ":" + port)
+			return
+		}
 		// if the server is listening on a specific address let's make it visible in the log,
 		// so one can differentiate between all active listeners
-		fmt.Println(zone + " on " + s.Addr)
+		fmt.Println(zone + ":" + port + " on " + ip)
```
fturib: Changed the printed format so it present the same line as before, when BIND is not involved.
Sample of new log is:

```
com.:6003
.:5001 on 127.0.0.1
.:5002 on 127.0.0.1
.:5002 on ::1
.:5002 on 192.168.1.163

```
for the corefile below:
```
com.:6003 {
    log . "6001 com ::1 {type} {name} - to 5003"
    errors
    proxy . 127.0.0.1:5001 {
        protocol dns force_tcp
    }
}


.:5001 {
    bind 127.0.0.1
    log . "5001 (type} {name}"
    proxy . 8.8.8.8:53 {
        protocol dns force_tcp
    }
}

.:5002 {
    bind 127.0.0.1 ::1 192.168.1.163
    log . "5002 (type} {name}"
    proxy . 8.8.8.8:53 {
        protocol dns force_tcp
    }
}
```

plugin/bind/setup.go
```diff
@@ -14,19 +14,19 @@ func setupBind(c *caddy.Controller) error {
...
+		for _, addr := range addrs {
 			if net.ParseIP(addr) == nil {
 				return plugin.Error("bind", fmt.Errorf("not a valid IP address: %s", addr))
 			}
 		}
-		allAddresses = append(allAddresses, addresses...)
+		all = append(all, addrs...)
 	}
-	config.ListenHosts = allAddresses
+	config.ListenHosts = all
```
fturib: only change the name of the variables.

fturib:
Uploaded a new version with fixed based on review.

fturib:
@miekg : is that last version ok ? Thanks for review / merge when ok.

miekg (approved):

PR #1514, opened 2018-02-09
plugin/cache: Don't cache completely empty messages

grobie:
While a well-behaving server should always include a SOA record in the
Authority section for NXDOMAIN/no data responses, CoreDNS should err on
the side of caution and not cache responses for the configured maximum
TTL from such misbehaving servers.

For more information, see Section 3 of [RFC2308](https://tools.ietf.org/html/rfc2308).

@miekg

### DIFF
--- a/plugin/cache/item.go
+++ b/plugin/cache/item.go
@@ -95,6 +95,10 @@ func minMsgTTL(m *dns.Msg, mt response.Type) time.Duration {
 		return 0
 	}
 
+	if len(m.Answer) == 0 && len(m.Ns) == 0 {
+		return 0
+	}
+
 	minTTL := maxTTL
 	for _, r := range append(m.Answer, m.Ns...) {
 		switch mt {
--- a/plugin/cache/minttl_test.go
+++ b/plugin/cache/minttl_test.go
@@ -25,9 +25,8 @@ func TestMinMsgTTL(t *testing.T) {
 	if mt != response.NoData {
 		t.Fatalf("Expected type to be response.NoData, got %s", mt)
 	}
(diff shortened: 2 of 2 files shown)


miekg:
probably makes more sense to do this in pkg/response//typify.go and adding a new Type called Empty and checking for `response.Empty` in *cache*

grobie:
Hm, sure, if you prefer that, I can make that change. Can you explain the advantages?

miekg:
Because that's where we decide what kind of message we deal with and Empty is surely something we want to use elsewhere.

grobie:
I assume only messages will match that `Empty` type where all Answer, Additional and Authority sections are empty. This will cause messages with only records in the Additional section still to be cached. While such messages don't make sense, the whole idea of this PR is to make CoreDNS more robust against broken server implementations. So we would still need the validation added in this PR. Do you prefer to create that custom Empty type regardless?

p.s. And yes before anyone asks, our broken server had a bug where it returned a SOA record in the Additional section in one case ...

miekg:
>I assume only messages will match result in that `Empty` type where all Answer, Additional and Authority sections are empty. This will cause messages with only records in the Additional section still to be cached. While such messages don't make sense, the whole idea of this PR is to make CoreDNS more robust against broken server implementations. So we would still need the validation added in this PR. Do you prefer to create that custom Empty type regardless?
>
>p.s. And yes before anyone asks, our broken server had a bug where it returned a SOA record in the Additional section in one case ...

LOL

These types of messages should probably be lumped into OtherError, i.e. not 
creating another type.

* messages with no content
* message with only things in Additional section.

miekg:
closed via new PR #1538

Issue #1492, opened 2018-02-06
plugin/metrics: only metrics from a single zone block are reported [BUG]


achille-roussel:
I'm observing what I believe is not the expected behavior with the metrics plugin (or at least not the documented behavior). 

Considering a configuration like this:
```
.com {
  proxy . 127.0.0.1:53
  autopath /zones.conf
  prometheus :9053
}

. {
  proxy . 127.0.0.1:53
  prometheus :9053
}
```
The `coredns_dns_` metrics which are generated by each metrics plugin are properly reported for both zones, however the metrics of the proxy and autopath plugins are not. Either metrics of the `.com` and `.` zones are reported. I haven't been able to determine which block takes priority, it seems kind of random.

The issue I believe comes from the fact that plugins in each zone will be attached to the `*metrics.Metrics` instance of their block (see this example https://github.com/coredns/coredns/blob/master/plugin/cache/setup.go#L33-L48), but only one `*metrics.Metrics` instance is configured to listen on a pair of IP and port (see https://github.com/coredns/coredns/blob/master/plugin/metrics/setup.go#L32-L39 and https://github.com/coredns/coredns/blob/master/plugin/metrics/metrics.go#L87-L104).
This results in having some plugins _wired_ to a metrics instance that is not serving requests.

A possible solution would be to share the `*prometheus.Registry` between `*metrics.Metrics` instances that are configured to be served on the same port. It means the `(*metrics.Metrics).MustRegister` method must be revisited to not panic when a metric is registered multiple times by multiple instantiations of a plugin.

I'm available to send a fix, but I wanted to get another pair of eyes on it in case I missed something.

johnbelamaric:
See #1452. I think we need to decide on how we want it to work. Should metrics just be globally enabled, or per server block? Right now it's getting confused.

achille-roussel:
Global plugins is something I was thinking of as well, but that seems more disruptive than fixing the current broken behavior (it seems possible to do so in this case), what do you think?

fturib:
looking at https://github.com/coredns/coredns/blob/master/plugin/metrics/setup.go#L32-L39, I would think that the plugin is assuming that you would change the address for each of your directives.

then, you would have a series of stats on each of the listening addresses.

However, because of lines #47 o #49, there is maybe some race conditions where the last directive is read the the address may not be set soon enough in uniqAddr to be generate the m.OnStartup ...



```
.com {
  proxy . 127.0.0.1:53
  autopath /zones.conf
[...]

PR #1521, opened 2018-02-11
[WIP] Share listeners in metrics plugin.

achille-roussel:
Hi all,

this pull request is a work in progress on the issue discussed in #1492, I'm facing a couple of bumps that I would like to gather input on before I move forward.

The changes I'm making consist of adding a shared state indirection that the `*Metrics` type uses internally to share listeners, http handlers, and prometheus registries. That way metrics of each zones can be exposed to the port they're declared on, and multiple zones can be exposed under the same port.

However there are two issues that the current design makes hard to address:
- plugins used a `sync.Once` variable to prevent registering the metrics multiple times to the same registry, but we need to remove this global synchronization in order to allow metrics to register to the registry of their zone block. This is a significant design change so I want to make sure it's OK.
- prometheus allows collectors (metrics) to be registered to multiple registries, but the use of global variables for metrics in CoreDNS means that metrics of all zones end up being exposed to all listeners, which is not the intuitive behavior (but is better than not having the metrics at all).

Overall, I feel like there's a bit of a need for a better design in how metrics are collected in CoreDNS, which doesn't involve using global variables, so we can properly scope metrics to the zone that they were generated in.

Since prometheus has been used in the project we could keep metrics as fields of the plugin values, so each plugin instantiation only reports on its own metrics to the prometheus registry that it was attached to.

Now there's a catch since two instantiations of a plugin with the same configuration would likely collide and fail to register. We can handle it by inspecting the error returned by [Register](https://godoc.org/github.com/prometheus/client_golang/prometheus#Registerer) in that case, if it's of type [AlreadyRegisteredError](https://godoc.org/github.com/prometheus/client_golang/prometheus#AlreadyRegisteredError) we get returned the collector that the operation conflicted with, so we can use that one in the plugin in place of the one we were attempting to register (since they're basically the same, it won't be wrong for two plugins to use the same metric value).

Another way to work around this problem is two require having the zones that the plugin is part of as constant labels on the metrics it registers. That way there will never be a conflict (since zone blocks are already unique). It also has the nice side effect of providing a way to visualize all metrics for a zone instead of having global ones (like the cache metrics for example).

I think we can make those changes progressively and they don't need to be part of this pull request, if we relax the use of `sync.Once` we get a non-ideal but working behavior, then we can tackle each plugin independently to have them only report metrics for the zone block that they are used in.

Let me know what you think!

### DIFF
--- a/corefile.cfg
+++ b/corefile.cfg
@@ -1,0 +1,22 @@
+. {
+    proxy . /etc/resolv.conf
+    prometheus :9053
+}
+
+google.com {
+    proxy . /etc/resolv.conf
+    prometheus :9053
+    cache 60
+}
+
+github.com {
+    proxy . /etc/resolv.conf
+    prometheus :9053
+    cache 60
+}
+
+segment.com {
+    proxy . /etc/resolv.conf
+    prometheus :9054
+    cache 60
+}
--- a/plugin/cache/setup.go
+++ b/plugin/cache/setup.go
@@ -31,19 +31,17 @@ func setup(c *caddy.Controller) error {
 	})
 
 	c.OnStartup(func() error {
-		once.Do(func() {
-			m := dnsserver.GetConfig(c).Handler("prometheus")
-			if m == nil {
-				return
-			}
-			if x, ok := m.(*metrics.Metrics); ok {
-				x.MustRegister(cacheSize)
-				x.MustRegister(cacheCapacity)
-				x.MustRegister(cacheHits)
-				x.MustRegister(cacheMisses)
-				x.MustRegister(cachePrefetches)
-			}
-		})
+		m := dnsserver.GetConfig(c).Handler("prometheus")
+		if m == nil {
+			return nil
+		}
+		if x, ok := m.(*metrics.Metrics); ok {
+			x.MustRegister(cacheSize)
+			x.MustRegister(cacheCapacity)
+			x.MustRegister(cacheHits)
+			x.MustRegister(cacheMisses)
+			x.MustRegister(cachePrefetches)
+		}
 		return nil
 	})
 
--- a/plugin/metrics/metrics.go
+++ b/plugin/metrics/metrics.go
@@ -2,6 +2,7 @@
 package metrics
 
 import (
+	"fmt"
 	"log"
 	"net"
 	"net/http"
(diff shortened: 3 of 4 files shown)


corefile.cfg
```diff
@@ -0,0 +1,22 @@
...
+    proxy . /etc/resolv.conf
+    prometheus :9053
+    cache 60
+}
+
+segment.com {
+    proxy . /etc/resolv.conf
+    prometheus :9054
+    cache 60
+}
```
achille-roussel, line 22: This file is only here as an example, I'll remove it once the code is ready to be merged.

johnbelamaric:
I haven't had a chance to look at this in detail, so maybe you have covered these. But in case not, here a few additional considerations.

> Another way to work around this problem is two require having the zones that the plugin is part of as constant labels on the metrics it registers. That way there will never be a conflict (since zone blocks are already unique). It also has the nice side effect of providing a way to visualize all metrics for a zone instead of having global ones (like the cache metrics for example).

It's not just zone. You could have the same zone more than once in different servers. So:

```
.:53 {
  proxy . 8.8.8.8
  prometheus
}

.:54 {
 proxy . 1.2.3.4
 prometheus
}
```

is legitimate. We therefore need the server info too. Maybe it already does this, I haven't looked into the code - but I wanted to make sure you were aware of it. This may already be getting addressed in #1501, but you should make sure.

The other issue is config reloading. This has come up recently, we have found that many of the plugins do not properly handle this situation - in particular those that use global variables. So, please be sure to take that in consideration in your design.

achille-roussel:
Thanks for the feedback. You make a good point about the server part in the config, I had not considered it. We would indeed need to tag metrics with the server address to ensuite uniqness of the metrics. 
The current implementation does work for this use case tho, that shouldn’t be an issue.

miekg:
[also haven't looked at the code yet]

But are we considering labeling every metrics with a server label? (Something like that is fine by me, just checking what the TL;DR is here)

achille-roussel:
Yes that would be the idea, basically adding a label to scope the instantiation of the metric so it does no conflict with other server blocks.

corefile.cfg
miekg: I missing something here, I think. But when you register a metrics in prometheus, currently via the sync.Once, so say `.` registers the metrics. Why don't the other (say) proxies plugins re-use that one?

miekg:
Ok see my comments on #1492, I think this can be done simpler

miekg:
This is fixed by adding the server label and with WithServer function - still needs to be expanded to all metrics; but works. closing this PR

Issue #1100, opened 2017-09-21
Subscription plugin


johnbelamaric:
In some cases, a client may want to subscribe to updates for a particular name, in a manner similar to a _watch_ in etcd or consul.

We can implement a subscription plugin that uses the gRPC server and extends the API to allow clients to subscribe to a query. Upon receiving the request, it will call down into the chain and get the response, returning it to the client but also keeping track of the records for that query. It will provide a channel (or function call) that *other* plugins will call whenever their state changes. For example, if `auto` reloads a file, and the `subscription` (or `push`?) plugin is enabled,  it will call a function to indicate that "something in this zone may have change". The subscription plugin will then re-query the records for any subscriptions to records in that zone, compare to the current state, and push down the changes to the subscribers.

It's not the most efficient way, but in this simple implementation we have the bulk of the logic in the new middleware, and avoid other plugins having to do much work.

miekg:
At least in etcd this is called a Watch; but that doesn't do the diff. It send you "some changed" and an index. For simplicity this could be a first iteration; don't send anything back, just something has changed and let the client requery.

zhujinghui:
Actually, etcd will send client  not only index, but also the current and previous data, see doc here https://coreos.com/etcd/docs/latest/v2/api.html#waiting-for-a-change

etcd2 implement watch by long polling (etcd 2 using HTTP/1.1). And etcd3 implement watch by gRPC stream (which may leverage on HTTP/2 server push, I am not sure), see https://coreos.com/etcd/docs/latest/learning/api.html#watch-streams

In either case, server side will send client side updated data, client does not need requery.

johnbelamaric:
Yeah. I have a very crude not-working-yet start on this plugin on my laptop that I haven't pushed yet. It's basically etcd-like functionality using a grpc stream. Users connect and create that watch, then the watch plugin informs each supported plugin that there is a watch on some particular name and gives them a channel to tickle if they have a relevant change.

If we implement this, then etcd plugin would be notified when a user creates a watch, so it could in fact employ its own watch on the same name. Right now I am just doing a k8s implementation, which will basically notify the watch plugin if anything changes in k8s. First cut won't be selective but eventually it will get better.

zhujinghui:
[...]

PR #1527, opened 2018-02-14
Watch feature

johnbelamaric:

Implements a watch feature in the gRPC server interface. Plugins that implement a specific interface will be notified of watches when they are added and canceled. When they see a change in the records for a name that is being watched, they send those names down a channel, and the watch server notifies any clients watching that name.

There is still a lot of work needed on this.

#1100 

Lots will be needed.


johnbelamaric:
You need to run CoreDNS with a gRPC server:

```
grpc://.:5553 {
	kubernetes cluster.local {
		endpoint https://kubernetes:8443/
		tls /Users/jbelamaric/.minikube/client.crt /Users/jbelamaric/.minikube/client.key /Users/jbelamaric/.minikube/ca.crt
	}
}
```

Then you can use the `dnsgrpc` in client/dnsgrpc to try this out. Just go build in there. 

```
ANN-M-JBELAMARIC:dnsgrpc jbelamaric$ ./dnsgrpc -w -insecure foo.default.svc.cluster.local
foo.default.svc.cluster.local.	5	IN	A	10.108.75.187


foo.default.svc.cluster.local. A NXDOMAIN


foo.default.svc.cluster.local.	5	IN	A	10.106.97.184

```

Here I deleted then added the service back.

plugin/kubernetes/controller.go
```diff
@@ -79,14 +87,21 @@ type dnsControlOpts struct {
 	// Label handling.
 	labelSelector *meta.LabelSelector
 	selector      labels.Selector
+	// watch-related items channel
```
johnbelamaric: annoying that these are basically duplicated in Kubernetes and opts, but didn't want to refactor everything.

plugin/pkg/watch/watch.go
```diff
@@ -0,0 +1,12 @@
+package watch
+
+// Chan is used to inform the server of a change
+type Chan chan []string
+
+// Watchable is the interface watchable plugins should implement
+type Watchable interface {
+	Name() string
```
johnbelamaric: I don't think I need this anymore, I think I had it here for a reason but ended up changing things around so I don't need it now.

chrisohaver:
@johnbelamaric, I've added watching of endpoints and pods, addressed most of the TODOs you had in there.

Is there anything else you see I should add to your PR?

core/dnsserver/watch.go
```diff
@@ -0,0 +1,153 @@
...
+				return err
+			}
+
+			qname := msg.Question[0].Name
+			w.mutex.Lock()
+			if _, ok := w.watches[qname]; !ok {
+				w.watches[qname] = make(watchlist)
+			}
+			w.watches[qname][id] = stream
+
```
johnbelamaric: Can the mutex be released here instead? `w.plugins` is set on start and never changed (well, maybe on reload?)

core/dnsserver/watch.go
```diff
@@ -0,0 +1,153 @@
...
+func (w *watcher) processWatches() {
+	for {
+		select {
+		case changed := <-w.changes:
+			w.mutex.Lock()
+			for qname, wl := range w.watches {
+				if plugin.Zones(changed).Matches(qname) == "" {
+					continue
+				}
+				log.Printf("Matches %s\n", qname)
```
johnbelamaric: debug

pb/dns.proto
```diff
@@ -9,4 +9,40 @@ message DnsPacket {
...
+	// from the same stream.
+	// All replies sent to the created watcher will attach with the same watch_id.
+	bool created = 2;
+
+	// canceled is set to true if the response is for a cancel watch request.
+	// No further events will be sent to the canceled watcher.
+	bool canceled = 3;
+
+	string qname = 4;
+	//DnsPacket reply = 6;
```
johnbelamaric: remove extraneous entry.

And maybe we should add an error message here. Is the `qname` even used?

core/dnsserver/watch.go
```diff
@@ -0,0 +1,153 @@
...
+		plugins := config.Handlers()
+		for _, p := range plugins {
+			if x, ok := p.(watch.Watchable); ok {
+				x.SetWatchChan(w.changes)
+				w.plugins = append(w.plugins, x)
+			}
+		}
+	}
+
+	//TODO: maybe a stop channel, work properly with reloads?
```
johnbelamaric: I think we need a stop channel to handle reloads properly.

core/dnsserver/watch.go
```diff
@@ -0,0 +1,153 @@
...
+
+	w.counter++
+	id := w.counter
+
+	w.mutex.Unlock()
+	return id
+}
+
+// watch is used to monitor the results of a given query. CoreDNS will push updated
+// query responses down the stream.
```
johnbelamaric: Comment is not accurate anymore

core/dnsserver/watch.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+		if err != nil {
+			return err
+		}
+		create := in.GetCreateRequest()
+		if create != nil {
+			msg := new(dns.Msg)
+			err := msg.Unpack(create.Query.Msg)
+			if err != nil {
+				// TODO: should write back an error response not break the stream
```
johnbelamaric: should we have some sort of error message we can give back when we say that we couldn't create it?

core/dnsserver/watch.go
```diff
@@ -0,0 +1,153 @@
...
+				stream.Send(&pb.WatchResponse{Created: false})
+				return nil
```
johnbelamaric: I think you want to just `continue` here.

core/dnsserver/watch.go
```diff
@@ -0,0 +1,153 @@
...
+		cancel := in.GetCancelRequest()
+		if cancel != nil {
+			w.mutex.Lock()
+			for qname, wl := range w.watches {
+				ws, ok := wl[cancel.WatchId]
+				if !ok {
+					continue
+				}
+
+				// only allow cancels from the client that started it
```
johnbelamaric: Can you test what happens if a stream tries to cancel a watchID that it doesn't own?

core/dnsserver/watch.go
```diff
@@ -0,0 +1,153 @@
...
+			for qname, wl := range w.watches {
+				if plugin.Zones(changed).Matches(qname) == "" {
+					continue
+				}
+				log.Printf("Matches %s\n", qname)
+				for id, stream := range wl {
+					wr := pb.WatchResponse{WatchId: id, Qname: qname}
+					err := stream.Send(&wr)
+					if err != nil {
+						log.Printf("Error sending to watch %d: %s\n", id, err)
```
johnbelamaric: Should this be tagged with `[WARNING]`?

client/client.go
```diff
@@ -0,0 +1,166 @@
...
+// Watch is used to track access results from watching a particular query.
+type Watch struct {
+	WatchID int64
+	Msgs    chan *dns.Msg
+	stream  pb.DnsService_WatchClient
+	client  *Client
+}
+
+// NewClient establishes a connection to a server and returns a pointer to a Client.
+func NewClient(endpoint, cert, key, ca string) (*Client, error) {
```
johnbelamaric: We should add a `[]grpc.DialOption` parameter that we append all our options to. This way they can set backoff policies, etc.

plugin/kubernetes/controller.go
```diff
@@ -491,63 +508,167 @@ func (dns *dnsControl) updateModifed() {
...
+		p := oldObj.(*api.Endpoints)
+		// endpoint updates can come frequently, make sure it's a change we care about
+		if endpointsEquivalent(p, ob) {
 			return
 		}
+		dns.sendEndpointsUpdates(xorSubsets(p, ob))
+	case *api.Pod:
+		dns.sendPodUpdates(ob)
+	default:
+		fmt.Printf("Updates for %T not supported.", ob)
```
johnbelamaric: `[WARNING]`

plugin/kubernetes/controller.go
```diff
@@ -491,63 +508,167 @@ func (dns *dnsControl) updateModifed() {
...
+			}
+		}
+	}
+	if len(z) > 0 {
+		*dns.watchChan <- z
+	}
+}
+
+// xorSubsets returns an endpoints struct containing the Subsets that have changed (xor) between a and b
+func xorSubsets(a, b *api.Endpoints) *api.Endpoints {
```
johnbelamaric: I am not sure I understand this function.

plugin/kubernetes/controller_test.go
```diff
@@ -0,0 +1,53 @@
...
+	for _, ap := range addrs {
+		apa := strings.Split(ap, ":")
+		address := apa[0]
+		port, _ := strconv.Atoi(apa[1])
+		eps = append(eps, api.EndpointSubset{Addresses: []api.EndpointAddress{{IP: address}}, Ports: []api.EndpointPort{{Port: int32(port)}}})
+	}
+	return eps
+}
+
+func TestXorSubsets(t *testing.T) {
```
johnbelamaric: ok, this helps me understand it...

plugin/kubernetes/controller.go
```diff
@@ -491,63 +508,167 @@ func (dns *dnsControl) updateModifed() {
...
+func (dns *dnsControl) Add(obj interface{}) {
+	dns.updateModifed()
+	dns.sendUpdates(nil, obj)
+}
+func (dns *dnsControl) Delete(obj interface{}) {
+	dns.updateModifed()
+	dns.sendUpdates(obj, nil)
+}
+func (dns *dnsControl) Update(oldObj, newObj interface{}) {
+	dns.updateModifed()
```
johnbelamaric: We don't want to `updateModified` if endpoints are equivalent.

plugin/kubernetes/kubernetes.go
```diff
@@ -291,6 +299,29 @@ func (k *Kubernetes) Records(state request.Request, exact bool) ([]msg.Service,
 	return services, err
 }
 
+// ServiceFQDN returns the k8s cluster dns spec service FQDN for the service (or endpoint) object
```
johnbelamaric: In another PR we should update all the places we do this to use these functions.

plugin/kubernetes/controller.go
chrisohaver: It returns an endpoints struct containing the Subsets that have changed (xor) between a and b...

chrisohaver: We only want to notify client of things that changed...

pb/dns.proto
chrisohaver: :) I have no idea what this file does... `.proto`?  how does it fit in?

core/dnsserver/watch.go
chrisohaver: Sorry - can you explain more?  I understand that i need to do something to make the reload plugin work here... I don't understand specifically what I need to do with a stop channel.  Is there an example of something else that uses a stop channel in relation to the reload plugin?

pb/dns.proto
johnbelamaric: It is used to generate the grpc code and protobuf message formats. If you change it you have to run `protoc --go_out=plugins=grpc:. dns.proto` in the `pb` directory. I think `brew install protobuf` will get you `protoc`.

plugin/kubernetes/controller.go
johnbelamaric: I understand what it is *for*. But it's a bit opaque to understand what it *does*. But looking at it in combination with the test helps.

chrisohaver: I'll expand the description...

core/dnsserver/watch.go
johnbelamaric: Look at `plugin/pkg/up/up.go`. Or `grep stop -r plugin` and you'll see some more examples. Basically you have `processWatches` do its `select` on both channels, and if it gets anything on the stop channel, it exits (stops the go routine).

Then in OnShutdown you send `true` down the stop channel. We might also want to notify all the clients that we are stopping the watches at that time, too. And close the streams maybe? Clean up at least.

plugin/kubernetes/controller.go
chrisohaver: ... and i renamed it to endpointsSubsetDiffs()

core/dnsserver/watch.go
chrisohaver: OK - I guess we can log an error to the coredns log.  I didn't se a way to send an error message/code to the client.  Or do I need to expand that... OH I see.. thats what the `.proto` file is for...
OK, will add an message field to the WatchResponse definition.

```
string message = 5
```

... and then add the client part to dnsgrpc.go.

or am i still confused.

johnbelamaric:
You are not confused, that is exactly right.

core/dnsserver/watch.go
johnbelamaric: @chrisohaver did you test this out (a stream trying to cancel a watchID it doesn't own)?

johnbelamaric (commented):
In a recent commit, @miekg changed this line:

https://github.com/coredns/coredns/pull/1527/files#diff-77a5c4e2c5fea6e91e4761f4251536b1R24

to be just "context". We should do the same here (even though normally you wouldn't modify generated code).

johnbelamaric:
The commit comment is preventing the WIP check from passing.

johnbelamaric:
@miekg @yongtang PTAL, this is ready to review.

client/client.go
```diff
@@ -0,0 +1,182 @@
+package client
```
miekg: should prolly move these to a new repo. From a cursory look it is needed for the cli tool?

plugin/pkg/watch/watch.go
```diff
@@ -0,0 +1,12 @@
...
+	SetWatchChan(Chan)
+	Watch(qname string) error
```
miekg: is qname fully qualified or not? You probably want to document each method in the interface to help implementers.

plugin/kubernetes/kubernetes.go
```diff
@@ -294,6 +302,29 @@ func (k *Kubernetes) Records(state request.Request, exact bool) ([]msg.Service,
...
+
+// PodFQDN returns the k8s cluster dns spec FQDN for the pod
+func PodFQDN(p *api.Pod, zone string) string {
+	name := strings.Replace(p.Status.PodIP, ".", "-", -1)
+	name = strings.Replace(name, ":", "-", -1)
+	return dnsutil.Join(append([]string{}, name, p.GetNamespace(), Pod, zone))
+}
+
+// EndpointFQDN returns a list of k8s cluster dns spec service FQDNs for each subset in the endpoint
+func EndpointFQDN(ep *api.Endpoints, zone string, endpointNameMode bool) []string {
```
miekg: can probably be all private as well, i.e. no need to export these from the kubernetes plugin?

plugin/kubernetes/kubernetes.go
```diff
@@ -294,6 +302,29 @@ func (k *Kubernetes) Records(state request.Request, exact bool) ([]msg.Service,
 	return services, err
 }
 
+// ServiceFQDN returns the k8s cluster dns spec service FQDN for the service (or endpoint) object
+func ServiceFQDN(obj meta.Object, zone string) string {
+	return dnsutil.Join(append([]string{}, obj.GetName(), obj.GetNamespace(), Svc, zone))
+}
+
+// PodFQDN returns the k8s cluster dns spec FQDN for the pod
```
miekg: not ending dot and dns -> DNS (more comments have not period on the end)

plugin/kubernetes/kubernetes.go
```diff
@@ -262,6 +266,10 @@ func (k *Kubernetes) InitKubeCache() (err error) {
 
 	k.opts.initPodCache = k.podMode == podModeVerified
 
+	k.opts.watchChan = &(k.watchChan)
```
miekg: `&(k.watchChan)` looks out of place are you sure this is correct?

plugin/pkg/watch/watch.go
miekg: does this inherit downwards, i.e. is qname a domain and will I get updates for everything under it? I guess no, but docs should make this clearer

plugin/kubernetes/controller.go
```diff
@@ -491,63 +508,176 @@ func (dns *dnsControl) updateModifed() {
...
+	for _, zone := range dns.zones {
+		names := append(EndpointFQDN(ep, zone, dns.endpointNameMode), ServiceFQDN(ep, zone))
+		for _, name := range names {
+			if _, ok := dns.watched[name]; ok {
+				z = append(z, name)
+			}
+		}
+	}
+	if len(z) > 0 {
+		*dns.watchChan <- z
```
miekg: this needs to be queued, because .... the interface demands a `[]string`?
What's wrong with send z down where you do the `append`? Or is this prohibitive from a network traffic perspective?

plugin/kubernetes/controller.go
```diff
@@ -7,6 +7,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	dnswatch "github.com/coredns/coredns/plugin/pkg/watch"
```
miekg, line 10: why the named import?

pb/dns.proto
```diff
@@ -9,4 +9,41 @@ message DnsPacket {
 
 service DnsService {
 	rpc Query (DnsPacket) returns (DnsPacket);
+	rpc Watch (stream WatchRequest) returns (stream WatchResponse);
+}
+
+message WatchRequest {
```
miekg, line 15: We should add a Makefile to convert to protos to go code

core/dnsserver/watch.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func (w *watcher) stop() {
+	w.mutex.Lock()
+	for wn, wl := range w.watches {
+		for id, stream := range wl {
+			wr := pb.WatchResponse{WatchId: id, Canceled: true}
+			err := stream.Send(&wr)
+			if err != nil {
+				log.Printf("[WARNING] Error notifiying client of cancellation: %s\n", err)
```
miekg: log.Warningf (now we have it)

core/dnsserver/watch.go
```diff
@@ -0,0 +1,166 @@
...
+				// let the client know we canceled the watch
+				stream.Send(&pb.WatchResponse{WatchId: cancel.WatchId, Canceled: true})
+			}
+			w.mutex.Unlock()
+			continue
+		}
+	}
+}
+
+func (w *watcher) processWatches() {
```
miekg: just process as method name?

core/dnsserver/watch.go
```diff
@@ -0,0 +1,166 @@
...
+			w.mutex.Lock()
+			for qname, wl := range w.watches {
+				if plugin.Zones(changed).Matches(qname) == "" {
+					continue
+				}
+				for id, stream := range wl {
+					wr := pb.WatchResponse{WatchId: id, Qname: qname}
+					err := stream.Send(&wr)
+					if err != nil {
+						log.Printf("[WARNING] Error sending change for %s to watch %d: %s\n", qname, id, err)
```
miekg: how often can this happen? (log spamming)

core/dnsserver/watch.go
miekg: How do you stop this goroutine?

core/dnsserver/watch.go
```diff
@@ -0,0 +1,166 @@
...
+	changes watch.Chan
+	counter int64
+	watches map[string]watchlist
+	plugins []watch.Watchable
+	mutex   sync.Mutex
+}
+
+type watchlist map[int64]pb.DnsService_WatchServer
+
+func newWatcher(zones map[string]*Config) *watcher {
```
miekg: Is it possible to move all this to a watch plugin? Or only keep the absolute minimal here?
I.e. health some similar things and has all its code in the plugin dir.

client/client.go
johnbelamaric: yes. ok.

plugin/pkg/watch/watch.go
johnbelamaric: exact name only if I remember right. and yes, FQDN.

plugin/kubernetes/kubernetes.go
johnbelamaric: It's needed. due to the ordering of how servers and plugins get intialized.

miekg:
just a general comment; this is a pretty damn cool feature

core/dnsserver/watch.go
johnbelamaric: I originally tried to do it that way. We can probably move code to a pkg or something. Honestly it's been a while since I did that and I don't recall the reasons exactly, other than that it needs to be in the gRPC service and so at its core in the grpc server.

core/dnsserver/watch.go
johnbelamaric: I think when this happens it's likely that the connection is broken and will get cleaned up before long. Not sure though.

core/dnsserver/watch.go
johnbelamaric: I think we have a stop channel in there now.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Watch feature..." ]
>
>
>
>> @@ -262,6 +266,10 @@ func (k *Kubernetes) InitKubeCache() (err error) {
>
> 	k.opts.initPodCache = k.podMode == podModeVerified
>
>+	k.opts.watchChan = &(k.watchChan)
>
>It's needed. due to the ordering of how servers and plugins get intialized.

But as a pointer?

core/dnsserver/watch.go
johnbelamaric: hmm, don't see the stop channel in the select, @chrisohaver can you take a look

plugin/kubernetes/controller.go
johnbelamaric: So you are saying just send them one at a time as separate changes (change the interface to string not []string)? I think that would be OK, it would be a bit simpler.

johnbelamaric:
Yeah. At this point, watchChan is nil. So we get a pointer that so that when it gets set, we see it.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Watch feature..." ]
>+	var z []string
>+	for _, zone := range dns.zones {
>+		names := append(EndpointFQDN(ep, zone, dns.endpointNameMode), ServiceFQDN(ep, zone))
>+		for _, name := range names {
>+			if _, ok := dns.watched[name]; ok {
>+				z = append(z, name)
>+			}
>+		}
>+	}
>+	if len(z) > 0 {
>+		*dns.watchChan <- z
>
>So you are saying just send them one at a time as separate changes (change the interface to string not []string)? I think that would be OK, it would be a bit simpler.

Yes, unless that has implications that I can't forsee?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Watch feature..." ]
>Yeah. At this point, watchChan is nil. So we get a pointer that so that when it gets set, we see it.

`	watchChan        *dnswatch.Chan`

is wrong, needs to be just 'watchChan dnswatch.Chan`, you still check for 
nilness IIRC.

/Miek

--
Miek Gieben

johnbelamaric:
I think that should be fine. We’ll take a look.

plugin/kubernetes/controller.go
johnbelamaric: It also imports "k8s.io/apimachinery/pkg/watch"

client/client.go
johnbelamaric: done

plugin/pkg/watch/watch.go
johnbelamaric: done

plugin/kubernetes/kubernetes.go
johnbelamaric: done

plugin/kubernetes/kubernetes.go
johnbelamaric: done

plugin/kubernetes/kubernetes.go
johnbelamaric: done

plugin/kubernetes/controller.go
johnbelamaric: done

core/dnsserver/watch.go
johnbelamaric: done

core/dnsserver/watch.go
johnbelamaric: done

pb/dns.proto
johnbelamaric: Done in a separate PR

johnbelamaric:
Ok, I managed to move most stuff to plugin/pkg/watch.

Still to do:
 - clean up stale watches (client disconnected)

I think that's it...right now if we get an error here https://github.com/coredns/coredns/pull/1527/files#diff-13573d12f6c520c266b20a363950a2edR153 then we don't remove the watch so every new change tries to keep pushing down a dead stream.

chrisohaver:
@johnbelamaric, ptal.  Cleaning stale watches seemed too simple - i don't see how to formally close the stream other than to delete the watch from the map.

Also: I did not test this because the client code is still in transit to a new repo ... You removed it from this PR, and presumably intend to move it to another repo.  I'm guessing that coredns/client is the project you intend as the destination, but it has no code in it, no branches other than master, and no forks... 

I can retrieve this code from the prior commits of this PR, and commit it to coredns/client for you.  But I wasn't sure if you had made changes in a local branch you have not pushed to github yet.

johnbelamaric:
Ok, I pushed the files to coredns/client. No Makefile, just have to do a "go build" in the cmd/dnsgrpc directory.

johnbelamaric:
@chrisohaver @miekg I think this is finally ready to go!

PR #1530, opened 2018-02-14
Allow overlapping Zones if binding addresses are different

fturib:
## 1. What does this pull request do?
Change the validation of overlapping zones of  ServerBlocs, by integrating the IP address that ServerBloc is bound too. 
(so far validation of overlapping was on Zone and Port only).

The support of overlapping zones implies changes in 2 areas:

### Validation of overlapping
The validator itself is moved a bit later in the process, after BIND defined the addresses of each ServerBloc. It is now moved in the MakeServer() - after the directives are processed, before we are grouping the Configs in order to create the Servers.

NOTE: the validation is now a little more complex as it has to take care of bound zones and unbound zones. I created an object to take care of it. see zoneAddrOverlapValidator in addresse.go

### registering the Configs in dnsContext
The key used to save Config in the dnsContext has to change because the keys of ServerBloc are no more discriminant for the ServerBlocs. It was the 'zone:port' information, it is now the concat of 2 index: position of the ServerBloc in the list, position of the Key in that ServerBloc.
=> GetConfig and SaveConfig has been changed accordingly.

Unit Tests are provided for each modified areas.
I prepare integration tests to validate the workflow of DNS messages inside the BlocLists. But it involves too much changes to integrate in this PR.
Will open a PR after this one will be approved.

## 2. Which issues (if any) are related?
#1478  and #1208 
it is second part of PR #1512

## 3. Which documentation changes (if any) need to be made?
- the global README of coredns does not specify the validation of overlapping zones/ports
- do we want to add an information in that doc ?

NOTE: I did not browse yet the blogs to check if there is a mention of overlapping check.

### DIFF
--- a/core/dnsserver/address.go
+++ b/core/dnsserver/address.go
@@ -94,3 +94,74 @@ const (
 	TransportTLS  = "tls"
 	TransportGRPC = "grpc"
 )
+
+const keyPartsSeparator = "##"
+
+type addrKey struct {
+	Transport string // dns, tls or grpc
+	Zone      string
+	Address   string
+	Port      string
+}
+
+func (k *addrKey) asKey() string {
+	return k.Transport + keyPartsSeparator + k.Zone + keyPartsSeparator + k.Address + keyPartsSeparator + k.Port
+}
+
+func parseAddrKey(key string) (*addrKey, error) {
+	keyParts := strings.Split(key, keyPartsSeparator)
+	if len(keyParts) != 4 {
+		return nil, fmt.Errorf("Provided value is not a key for addrKey, expect 4 parts joined by '%s' : %s", key, keyPartsSeparator)
+	}
+	return &addrKey{keyParts[0], keyParts[1], keyParts[2], keyParts[3]}, nil
+}
+
+func (k *addrKey) isMulticast() bool {
+	return k.Address == ""
+}
+
+func (k *addrKey) copyAsMulticast() *addrKey {
+	return &addrKey{Zone: k.Zone, Address: "", Port: k.Port, Transport: k.Transport}
+
+}
+
+// String return the string representation of this ZoneAddr - for print
+func (k *addrKey) String() string {
+	if k.Address == "" {
+		return k.Transport + "://" + k.Zone + ":" + k.Port
+	}
+	return k.Transport + "://" + k.Zone + ":" + k.Address + ":" + k.Port
+}
+
+// Build a Validator that rise error if the bound addresses for listeners are overlapping
+type zoneAddrOverlapValidator struct {
+	registeredAddr   map[string]string
+	multicastOverlap map[string]string
+}
+
+func newZoneAddrOverlapValidator() *zoneAddrOverlapValidator {
+	return &zoneAddrOverlapValidator{registeredAddr: make(map[string]string), multicastOverlap: make(map[string]string)}
+}
+
+func (c *zoneAddrOverlapValidator) registerAndCheck(k *addrKey) (same bool, overlap bool, overlapKey string) {
+	key := k.asKey()
+	if _, ok := c.registeredAddr[key]; ok {
+		// exact same zone already registered
+		return true, false, ""
+	}
+	mkey := k.copyAsMulticast().asKey()
+	if already, ok := c.multicastOverlap[mkey]; ok {
+		if k.isMulticast() {
+			// there is already a unicast registered
+			return false, true, already
+		}
+		if _, ok := c.registeredAddr[mkey]; ok {
+			// the overlapping multicast is already registered
+			return false, true, mkey
+		}
+	} else {
+		c.multicastOverlap[mkey] = key
+	}
+	c.registeredAddr[key] = key
+	return false, false, ""
+}
--- a/core/dnsserver/address.go
+++ b/core/dnsserver/address.go
@@ -15,10 +15,17 @@ type zoneAddr struct {
 	Port      string
 	Transport string     // dns, tls or grpc
 	IPNet     *net.IPNet // if reverse zone this hold the IPNet
+	Address   string     // used for bound zoneAddr - validation of overlapping
 }
 
 // String return the string representation of z.
-func (z zoneAddr) String() string { return z.Transport + "://" + z.Zone + ":" + z.Port }
+func (z zoneAddr) String() string {
+	s := z.Transport + "://" + z.Zone + ":" + z.Port
+	if z.Address != "" {
+		s += " on " + z.Address
+	}
+	return s
+}
 
 // Transport returns the protocol of the string s
 func Transport(s string) string {
@@ -95,81 +102,42 @@ const (
 	TransportGRPC = "grpc"
 )
 
-const keyPartsSeparator = "##"
-
-type addrKey struct {
-	Transport string // dns, tls or grpc
-	Zone      string
-	Address   string
-	Port      string
-}
-
-func (k *addrKey) asKey() string {
-	return k.Transport + keyPartsSeparator + k.Zone + keyPartsSeparator + k.Address + keyPartsSeparator + k.Port
-}
-
-func parseAddrKey(key string) (*addrKey, error) {
-	keyParts := strings.Split(key, keyPartsSeparator)
-	if len(keyParts) != 4 {
-		return nil, fmt.Errorf("provided value is not a key for addrKey, expect 4 parts joined by '%s' : %s", key, keyPartsSeparator)
-	}
-	return &addrKey{keyParts[0], keyParts[1], keyParts[2], keyParts[3]}, nil
-}
-
-func (k *addrKey) isUnbound() bool {
-	return k.Address == ""
-}
-
-func (k *addrKey) copyAsUnbound() *addrKey {
-	return &addrKey{Zone: k.Zone, Address: "", Port: k.Port, Transport: k.Transport}
-
-}
-
-// String return the string representation of this ZoneAddr - for print
-func (k *addrKey) String() string {
-	if k.Address == "" {
-		return k.Transport + "://" + k.Zone + ":" + k.Port
-	}
-	return k.Transport + "://" + k.Zone + ":" + k.Address + ":" + k.Port
-}
-
 // Build a Validator that rise error if the bound addresses for listeners are overlapping
 // we consider that an unbound address is overlapping all bound addresses for same zome, zame port
 type zoneAddrOverlapValidator struct {
-	registeredAddr map[string]string // each zoneAddr is registered once by its key (bound and unbound)
-	unboundOverlap map[string]string // the unbound equiv ZoneAdddr is registered by its original key (bound or unbound)
+	registeredAddr map[zoneAddr]zoneAddr // each zoneAddr is registered once by its key (bound and unbound)
+	unboundOverlap map[zoneAddr]zoneAddr // the unbound equiv ZoneAdddr is registered by its original key (bound or unbound)
 }
 
 func newZoneAddrOverlapValidator() *zoneAddrOverlapValidator {
-	return &zoneAddrOverlapValidator{registeredAddr: make(map[string]string), unboundOverlap: make(map[string]string)}
+	return &zoneAddrOverlapValidator{registeredAddr: make(map[zoneAddr]zoneAddr), unboundOverlap: make(map[zoneAddr]zoneAddr)}
 }
 
-func (c *zoneAddrOverlapValidator) registerAndCheck(k *addrKey) (same bool, overlap bool, overlapKey string) {
+func (c *zoneAddrOverlapValidator) registerAndCheck(k *zoneAddr) (same bool, overlap bool, overlapKey *zoneAddr) {
 
 	// we consider there is an overlap if:
 	//  - the current zoneAddr is already registered is already registered for the same zone/port
 	//  - the current zoneAddr is unbound and a non unbound is already registered for the same zone/port
 	//  - the current zoneAddr is bound and an unbound is already registered for the same zone/port
-	key := k.asKey()
-	if _, ok := c.registeredAddr[key]; ok {
+	if _, ok := c.registeredAddr[*k]; ok {
 		// exact same zone already registered
-		return true, false, ""
+		return true, false, nil
 	}
-	mkey := k.copyAsUnbound().asKey()
-	if already, ok := c.unboundOverlap[mkey]; ok {
-		if k.isUnbound() {
+	uk := zoneAddr{Zone: k.Zone, Address: "", Port: k.Port, Transport: k.Transport}
+	if already, ok := c.unboundOverlap[uk]; ok {
+		if k.Address == "" {
 			// there is already a bound registered
-			return false, true, already
+			return false, true, &already
 		}
-		if _, ok := c.registeredAddr[mkey]; ok {
+		if _, ok := c.registeredAddr[uk]; ok {
 			// the overlapping zone+port for the unbound addr is already registered
-			return false, true, mkey
+			return false, true, &uk
 		}
 	} else {
 		// anyway add the unbound equiv of current zoneAddr for future tests
-		c.unboundOverlap[mkey] = key
+		c.unboundOverlap[uk] = *k
 	}
 	// anyway add the current zoneAddr for future tests
-	c.registeredAddr[key] = key
-	return false, false, ""
+	c.registeredAddr[*k] = *k
+	return false, false, nil
 }
--- a/core/dnsserver/address.go
+++ b/core/dnsserver/address.go
@@ -111,16 +111,16 @@ func (k *addrKey) asKey() string {
 func parseAddrKey(key string) (*addrKey, error) {
 	keyParts := strings.Split(key, keyPartsSeparator)
 	if len(keyParts) != 4 {
-		return nil, fmt.Errorf("Provided value is not a key for addrKey, expect 4 parts joined by '%s' : %s", key, keyPartsSeparator)
+		return nil, fmt.Errorf("provided value is not a key for addrKey, expect 4 parts joined by '%s' : %s", key, keyPartsSeparator)
 	}
 	return &addrKey{keyParts[0], keyParts[1], keyParts[2], keyParts[3]}, nil
 }
 
-func (k *addrKey) isMulticast() bool {
+func (k *addrKey) isUnbound() bool {
 	return k.Address == ""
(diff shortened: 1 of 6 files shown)


core/dnsserver/register.go
```diff
@@ -45,38 +45,41 @@ type dnsContext struct {
...
 			if err != nil {
 				return nil, err
 			}
-			s.Keys[i] = za.String()
-			if v, ok := dups[za.String()]; ok {
-				return nil, fmt.Errorf("cannot serve %s - zone already defined for %v", za, v)
-			}
-			dups[za.String()] = za.String()
-
+			s.Keys[ik] = za.String()
```
fturib, line 64: Duplication test (overlapping checking) is removed from this function

core/dnsserver/register.go
```diff
@@ -100,6 +103,13 @@ func (h *dnsContext) InspectServerBlocks(sourceFile string, serverBlocks []caddy
 // MakeServers uses the newly-created siteConfigs to create and return a list of server instances.
 func (h *dnsContext) MakeServers() ([]caddy.Server, error) {
 
+	// Now that all Keys and Directives are parsed and initialized
+	// lets verify that there is no overlap on the zones and addresses to listen for
+	errValid := h.validateZonesAndListeningAddresses()
```
fturib, line 100: The overlapping test is done here.

core/dnsserver/register.go
```diff
@@ -45,38 +45,41 @@ type dnsContext struct {
...
 	h.configs = append(h.configs, cfg)
 	h.keysToConfigs[key] = cfg
 }
 
+func (h *dnsContext) getConfig(blocIndex int, keyIndex int) (*Config, bool) {
+	key := fmt.Sprintf("%d%s%d", blocIndex, keySeparator, keyIndex)
+	c, ok := h.keysToConfigs[key]
+	return c, ok
+}
+
```
fturib: The keys for saving Configs in the dnsContext are changed.

core/dnsserver/address.go
```diff
@@ -94,3 +94,74 @@ const (
...
+		}
+		if _, ok := c.registeredAddr[mkey]; ok {
+			// the overlapping multicast is already registered
+			return false, true, mkey
+		}
+	} else {
+		c.multicastOverlap[mkey] = key
+	}
+	c.registeredAddr[key] = key
+	return false, false, ""
```
fturib: Introducing an object to validate the overlaping of addresses.

We need to check that Zone x IP x Port is not overlapping, 
the tricky part is to verify that unicast cannot overlap with multicast, but 2 unicast can coexist.

core/dnsserver/server.go
```diff
@@ -65,6 +65,10 @@ func NewServer(addr string, group []*Config) (*Server, error) {
 		// set the config per zone
 		s.zones[site.Zone] = site
 		// compile custom plugin for everything
+		if site.registry != nil {
+			// this config is already computed with the chain of plugin
+			continue
+		}
```
fturib, line 71: A ServerBlocList is not duplicated when  it contains a BIND with mutiple addresses (or several BIND that lead to several addresses). 
The SAME ServerBlocList that is used from all of the socket listeners to process the DNS query.

core/dnsserver/address.go
```diff
@@ -94,3 +94,74 @@ const (
...
+
+func parseAddrKey(key string) (*addrKey, error) {
+	keyParts := strings.Split(key, keyPartsSeparator)
+	if len(keyParts) != 4 {
+		return nil, fmt.Errorf("Provided value is not a key for addrKey, expect 4 parts joined by '%s' : %s", key, keyPartsSeparator)
+	}
+	return &addrKey{keyParts[0], keyParts[1], keyParts[2], keyParts[3]}, nil
+}
+
+func (k *addrKey) isMulticast() bool {
```
miekg: Multicast? We want to care about this as well? Why? I thought this was going to be a purely textual change/check. We now also want to care about the type of addresses. 

NACK

at least in in this pr

fturib: @miekg : what I name multicast is the usual key zone + port, when there is no address bound.

I guess we need to validate this overlapping use case:

**not overlapping, and therefore accepted**
NOTE: this one is the use general use case we want to handle

.:53 { bind ::1 ...}
.:53 { bind 127.0.0.1 ..}

**overlapping, and therefore not accepted**
NOTE: this overlapping use case will also be denied by the listener : you cannot have a server listening a bound addresses if it is already listening the unbound address.

.:53 {...}
.:53 { bind 127.0.0.1 ..}

I call ".:53" a "multicast" addrKey, but if that is confusing (because not the right name), I can name it "unbound" .. 

What do you think ?

**NOTE:** I do not introduce any new concept .. I only try to check the overlapping with existing concepts : having no bind directive, or having bind directive with one or several unicast addresses.

**NOTE:** overlapping between bound and unbound address, is forbidden by the listening server. If we do not check that overlapping here, it will be denied later in the process. 
=> do we want to simplify our code (we are coming back to almost existing code) and let the listener deny the overlapping  ?

fturib:
This PR is now complete.
@miekg :let me know if my answer to your concern about "multicast" is good enough. I have renamed the variables replacing 'multicast' by 'unbound'.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP - Allow o..." ]
>This PR is now complete.
>@miekg :let me know if my answer to your concern about "multicast" is good enough. I have renamed the variables replacing 'multicast' by 'unbound'.

Don't call it multicast; that is networking term. `unbound` might be better; 
have to look at the code to make sure.

fturib:
It is now udpated in this PR.
There is no more reference to multicast or unicast. (I verified accross all code of coredns).
naming is now relative to 'bound' Zone/Port or 'unbound' Zone/Port

core/dnsserver/address.go
```diff
@@ -94,3 +94,82 @@ const (
...
+	Address   string
+	Port      string
+}
+
+func (k *addrKey) asKey() string {
+	return k.Transport + keyPartsSeparator + k.Zone + keyPartsSeparator + k.Address + keyPartsSeparator + k.Port
+}
+
+func parseAddrKey(key string) (*addrKey, error) {
+	keyParts := strings.Split(key, keyPartsSeparator)
```
miekg: So a zone with `#` in it breaks this whole scheme?

core/dnsserver/address.go
```diff
@@ -94,3 +94,82 @@ const (
 	TransportTLS  = "tls"
 	TransportGRPC = "grpc"
 )
+
+const keyPartsSeparator = "##"
+
+type addrKey struct {
```
miekg: Why do we need this and can't extend zoneAddr (which has almost the same bits?)

core/dnsserver/address.go
```diff
@@ -94,3 +94,82 @@ const (
...
+	//  - the current zoneAddr is unbound and a non unbound is already registered for the same zone/port
+	//  - the current zoneAddr is bound and an unbound is already registered for the same zone/port
+	key := k.asKey()
+	if _, ok := c.registeredAddr[key]; ok {
+		// exact same zone already registered
+		return true, false, ""
+	}
+	mkey := k.copyAsUnbound().asKey()
+	if already, ok := c.unboundOverlap[mkey]; ok {
+		if k.isUnbound() {
```
miekg: Not completely seeing the value in having this function on only using it once(?). k.Address == "" is just a clear.

core/dnsserver/address.go
```diff
@@ -94,3 +94,82 @@ const (
...
+	//  - the current zoneAddr is already registered is already registered for the same zone/port
+	//  - the current zoneAddr is unbound and a non unbound is already registered for the same zone/port
+	//  - the current zoneAddr is bound and an unbound is already registered for the same zone/port
+	key := k.asKey()
+	if _, ok := c.registeredAddr[key]; ok {
+		// exact same zone already registered
+		return true, false, ""
+	}
+	mkey := k.copyAsUnbound().asKey()
+	if already, ok := c.unboundOverlap[mkey]; ok {
```
miekg: as the plugin is called `bind` this is actually `bindOverlap`?

miekg (changes requested):

miekg:
I can't really see yet why this needs an additional 261 lines (but I haven't code it myself - so it may just balloon when you do). Anyways:

1. 

> Validation of overlapping
The validator itself is moved a bit later in the process, after BIND defined the addresses of each ServerBloc. It is now moved in the MakeServer() - after the directives are processed, before we are grouping the Configs in order to create the Servers.

This is most certainly the correct thing to do
 
2.

> NOTE: the validation is now a little more complex as it has to take care of bound zones and unbound zones. I created an object to take care of it. see zoneAddrOverlapValidator in addresse.go

Why is the key in the configs still a string? Seems to me we should just make this a struct and add the extra bind info to it? 

3.
> registering the Configs in dnsContext
The key used to save Config in the dnsContext has to change because the keys of ServerBloc are no more discriminant for the ServerBlocs. It was the 'zone:port' information, it is now the concat of 2 index: position of the ServerBloc in the list, position of the Key in that ServerBloc.
=> GetConfig and SaveConfig has been changed accordingly.

This doesn't set well with me, why do we care about the position? To make it unique, because we still store it in a map? Why can't we put it in a list then?

I think we should concentrate on step 1 first (which probably mean doing .3 as well, maybe).

core/dnsserver/address.go
fturib: That is a good point. 
I do not understand why we accept all chars in Zone (when parsing the keys for instance), but the domain specified by RFC is limited to alpha + digit + hyphen.
What is the use case of # in a zone ?

core/dnsserver/address.go
fturib: That was my first idea .. but ZoneAddr is used to compute/populate the Keys of ServerBloc.
and we did not wanted Address there.

I preferred to make a separated small object used only for the overlapping check. and has not the constrainst of the Keys.

core/dnsserver/address.go
fturib: it gives a name to what it is ... all is about bound and unbound in this checker

core/dnsserver/address.go
fturib: I will change the name.

core/dnsserver/address.go
fturib: but I can remove and use directly the test, with a comment

core/dnsserver/address.go
fturib: Although, I can make the move if you think that is preferable. let me know. Thanks

core/dnsserver/address.go
fturib: What is the syntax of the zone ?
I have this RFC : https://tools.ietf.org/html/rfc1035 - 2.3.1. Preferred name syntax

fturib: I am interested by the reply.
But anyway, I think I can avoid to write and parse a key.

fturib:
About 3 - I may misunderstand your point.
FWIU, the access to the Config is through that function: `func GetConfig(c *caddy.Controller) *Config` and Controller is a caddy object (we cannot change).

if we put the Configs into a list, how to we find the right Config to return ? it has to be based on key values of the Controller, no ?

fturib:
Uploaded a new version:
- fix or address all comments of review (in fact point 1.)
- for point 3. - I moved the key building near to Config, but I  do not understand how we could avoid to save these configs using index here. 
NOTE: the keyToConfigs map is useful ONLY during the initialization of CoreDNS.  I cleaned it after that initialization.

about point 2 : "Why is the key in the configs still a string? Seems to me we should just make this a struct and add the extra bind info to it?" 
=> could you rephrase the comment or explain what key of what configs ... ?

core/dnsserver/address.go
```diff
@@ -94,3 +101,43 @@ const (
...
 )
+
+// Build a Validator that rise error if the bound addresses for listeners are overlapping
+// we consider that an unbound address is overlapping all bound addresses for same zome, zame port
+type zoneAddrOverlapValidator struct {
+	registeredAddr map[zoneAddr]zoneAddr // each zoneAddr is registered once by its key (bound and unbound)
+	unboundOverlap map[zoneAddr]zoneAddr // the unbound equiv ZoneAdddr is registered by its original key (bound or unbound)
+}
+
+func newZoneAddrOverlapValidator() *zoneAddrOverlapValidator {
```
miekg: `zoneAddrOverlapValidator` -> `zoneOverlap`

core/dnsserver/address.go
```diff
@@ -94,3 +101,43 @@ const (
 	TransportTLS  = "tls"
 	TransportGRPC = "grpc"
 )
+
+// Build a Validator that rise error if the bound addresses for listeners are overlapping
```
miekg: This comment doesn't really describe what this struct does, but what you're planning to do with it, so this should be mostly moved to `registerAndCheck`, there are also a bunch of typos in it.

core/dnsserver/address.go
```diff
@@ -94,3 +101,43 @@ const (
...
+type zoneAddrOverlapValidator struct {
+	registeredAddr map[zoneAddr]zoneAddr // each zoneAddr is registered once by its key (bound and unbound)
+	unboundOverlap map[zoneAddr]zoneAddr // the unbound equiv ZoneAdddr is registered by its original key (bound or unbound)
+}
+
+func newZoneAddrOverlapValidator() *zoneAddrOverlapValidator {
+	return &zoneAddrOverlapValidator{registeredAddr: make(map[zoneAddr]zoneAddr), unboundOverlap: make(map[zoneAddr]zoneAddr)}
+}
+
+func (c *zoneAddrOverlapValidator) registerAndCheck(k *zoneAddr) (same bool, overlap bool, overlapKey *zoneAddr) {
```
miekg: does zoneAddr need to be pointers here?

core/dnsserver/address.go
```diff
@@ -94,3 +101,43 @@ const (
...
+		if _, ok := c.registeredAddr[uk]; ok {
+			// the overlapping zone+port for the unbound addr is already registered
+			return false, true, &uk
+		}
+	} else {
+		// anyway add the unbound equiv of current zoneAddr for future tests
+		c.unboundOverlap[uk] = *k
+	}
+	// anyway add the current zoneAddr for future tests
+	c.registeredAddr[*k] = *k
```
miekg: when is this code exectuted?

core/dnsserver/config.go
```diff
@@ -68,16 +69,22 @@ func (c *Config) HostAddresses() string {
 	return all
 }
 
+// build a key for identifying the configs during setup time
```
miekg: This isn't a proper Go comments, start iwth `keyForConfig ...`

core/dnsserver/register_test.go
```diff
@@ -51,15 +51,15 @@ func TestGroupingServers(t *testing.T) {
 			expectedGroups: []string{"dns://:53", "dns://:54"},
 			failing:        false},
 
-		// 2 configs on same port, same broadcast address, diff zones -> 1 group
+		// 2 configs on same port, both undound, diff zones -> 1 group
```
miekg: - > not using bind

core/dnsserver/register_test.go
```diff
@@ -51,15 +51,15 @@ func TestGroupingServers(t *testing.T) {
...
 		{configs: []*Config{
 			{Transport: "dns", Zone: ".", Port: "53", ListenHosts: []string{""}},
 			{Transport: "dns", Zone: "com.", Port: "53", ListenHosts: []string{""}},
 		},
 			expectedGroups: []string{"dns://:53"},
 			failing:        false},
 
-		// 2 configs on same port, same address, diff zones -> 1 group
+		// 2 configs on same port, one addressed - one unbound, diff zones -> 1 group
```
miekg: 'one using bind`

miekg (changes requested):
Yes, this looks much better. Thanks. Still a bunch of comments though, but veering into nit territory (for some)

core/dnsserver/address.go
fturib: no. I change.

core/dnsserver/address.go
fturib: ok. move and fix typos

core/dnsserver/address.go
fturib: ok.

core/dnsserver/address.go
fturib: when there is no overlap. I need to keep that code.

core/dnsserver/config.go
fturib: ok. update

core/dnsserver/register_test.go
fturib: ok. I change.

core/dnsserver/register_test.go
fturib: ok. I change.

fturib:
This PR has been updated, addressing the comments of review. (I hope ..).
Ready for another review...
Thanks!!

core/dnsserver/address.go
```diff
@@ -94,3 +101,37 @@ const (
 	TransportTLS  = "tls"
 	TransportGRPC = "grpc"
 )
+
+type zoneOverlap struct {
+	registeredAddr map[zoneAddr]zoneAddr // each zoneAddr is registered once by its key
+	unboundOverlap map[zoneAddr]zoneAddr // the "no bind" equiv ZoneAdddr is registered by its original key
+}
+
+func newZoneValidator() *zoneOverlap {
```
miekg: newZoneOverlap

core/dnsserver/address.go
```diff
@@ -94,3 +101,37 @@ const (
...
+	return &zoneOverlap{registeredAddr: make(map[zoneAddr]zoneAddr), unboundOverlap: make(map[zoneAddr]zoneAddr)}
+}
+
+// register a new zoneAddr, return information about existing or overlapping with already registered
```
miekg: should start with registerAndCheck

core/dnsserver/address.go
fturib: oops. my bad.

core/dnsserver/address.go
fturib: sure.

miekg (approved):

fturib:
Thanks for review and merge!

miekg:
Np. Thanks for your perseverance.

PR #1531, opened 2018-02-15
pkg/typify: empty messages are OtherError

miekg:
Messages with nothing in them are considered OtherError, they can not
serve any purpose for normal clients (i.e. dyn update or notifies might
have a use for them).

Also update a test in the cache plugin, so that we explicitaly test for
this case.


miekg:
IMO a better of solving not caching empty responses than #1514, as this lumps all those kind of responses into OtherError

grobie:
Will this change the rcode of the message or is the type only used internally?

miekg:
This will cause the cache to not cache these kind of messages.

PR #1541, opened 2018-02-18
go 1.10: update travis and gofmt

miekg:
fmt now complains about on file tls_test.go, fix that. Also remove
gofmt, as goimports also checks, so this was done twice.

Put go 1.9 and 1.10 in travis for the time being.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,8 @@ services:
 
 language: go
 go:
-  - 1.9
+  - "1.9"
+  - "1.10"
 
 go_import_path: github.com/coredns/coredns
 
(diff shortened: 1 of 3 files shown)


miekg:
See: older go (on coredns/ci)
~~~
ometalinter --deadline=2m --disable-all --enable=golint --enable=vet --enable=goimports --vendor --exclude=^pb/ ./...
plugin/tls/tls_test.go:1::warning: file is not goimported (goimports)
Makefile:79: recipe for target 'linter' failed
make[1]: *** [linter] Error 1
make[1]: Leaving directory '/tmp/tmp.8wBUFanJxw/src/github.com/coredns/coredns
~~~
make lint should be a separate step that doesn't stop the build

PR #1543, opened 2018-02-20
Scrub: use binary search

miekg:
Use binary search to find the minimal message size, that contains whole
RRs and fits the client's buffer. This is better then just setting
entire sections to `nil`. Extend the tests to test for additional and
answer section truncation. In the first case we *don't* set the TC bit.

### DIFF
--- a/request/request.go
+++ b/request/request.go
@@ -179,34 +179,86 @@ type Result int
 const (
 	// ScrubIgnored is returned when Scrub did nothing to the message.
 	ScrubIgnored Result = iota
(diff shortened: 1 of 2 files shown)


johnbelamaric:
lgtm

grobie:
I expect most issues to result from the scrub bug fixed in https://github.com/coredns/coredns/pull/1477 as it causes proxy to truncate valid compressed messages received from upstream.

A better approach would truncate and count in one single loop over the resource records.

PR #1551, opened 2018-02-21
plugin/kubernetes: PTR/A reverse query corner cases

chrisohaver:

Corrects some answers for _corner case_ queries when reverse zones are in the zones list.
Assuming `kubernetes cluster.local in-addr.arpa`, and that `svc1.default.cluster.local A 10.0.0.100` exists ...

`svc1.default.svc.in-addr.arpa.` `A` => `NXDOMAIN` (previously returned an `A` record for the service)
`100.0.0.10.in-addr.arpa.` `A` => `NODATA` (previously returned `NXDOMAIN`)
`svc1.default.svc.in-addr.arpa.` `PTR` => `NXDOMAIN` (previously returned `NODATA`)
`not-a-service.default.svc.cluster.local.` `PTR` => `NXDOMAIN` (previously returned `NODATA`)

coredns/ci tests will need to be adjusted/amended...

### DIFF
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -122,7 +122,7 @@ func (k *Kubernetes) Services(state request.Request, exact bool, opt plugin.Opti
 		return []msg.Service{svc}, nil
 	}
 
-	if dnsutil.IsReverse(state.Name()){
+	if dnsutil.IsReverse(state.Name()) {
 		// If the query is in a reverse zone, don't look up services
 		// instead return no answer, with error code of reverse lookup
 		_, err := k.Reverse(state, exact, opt)
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -122,14 +122,6 @@ func (k *Kubernetes) Services(state request.Request, exact bool, opt plugin.Opti
 		return []msg.Service{svc}, nil
 	}
 
-	if dnsutil.IsReverse(state.Name()) {
-		// If the query is in a reverse zone, don't look up services
-		// instead return no answer, with error code of reverse lookup
-		_, err := k.Reverse(state, exact, opt)
-		return svcs, err
-
-	}
-
 	s, e := k.Records(state, false)
 
 	// SRV for external services is not yet implemented, so remove those records.
@@ -280,6 +272,10 @@ func (k *Kubernetes) Records(state request.Request, exact bool) ([]msg.Service, 
 	r, e := parseRequest(state)
 	if e != nil {
 		return nil, e
+	}
+
+	if dnsutil.IsReverse(state.Name()) {
+		return nil, errNoItems
 	}
 
 	if !wildcard(r.namespace) && !k.namespaceExposed(r.namespace) {
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -122,6 +122,14 @@ func (k *Kubernetes) Services(state request.Request, exact bool, opt plugin.Opti
 		return []msg.Service{svc}, nil
 	}
 
+	if dnsutil.IsReverse(state.Name()){
+		// If the query is in a reverse zone, don't look up services
+		// instead return no answer, with error code of reverse lookup
+		_, err := k.Reverse(state, exact, opt)
+		return svcs, err
+
+	}
+
 	s, e := k.Records(state, false)
 
 	// SRV for external services is not yet implemented, so remove those records.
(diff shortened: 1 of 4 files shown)


chrisohaver:
The reasoning behind these corrections are:
1. We should only be returning `NXDOMAIN` when there are no records of any type for the name.
2. We should return `NODATA` when there is no record for the requested type, but there are records of other types for the name.

plugin/kubernetes/reverse.go
```diff
@@ -14,7 +14,11 @@ func (k *Kubernetes) Reverse(state request.Request, exact bool, opt plugin.Optio
 
 	ip := dnsutil.ExtractAddressFromReverse(state.Name())
 	if ip == "" {
-		return nil, nil
+		if dnsutil.IsReverse(state.Name()) {
```
miekg: This is to shortcut and not perform the `k.Records` lookup? Most likely shouldn't care, and just do the `Records` lookup

chrisohaver: If the zone is a reverse zone, then we should never do a `k.Records` lookup.
If we did, we would answer things like ...
`s1.ns.svc.in-addr.arpa.` has `PTR`? => `NOERROR`

I may be able to move the reverse check into `k.Records`, that way it's only in one place.

chrisohaver: > I may be able to move the reverse check into k.Records, that way it's only in one place.

Nevermind - that would end up being uglier... requiring checking the query type again in Records.

chrisohaver: If the zone is a reverse zone, then we should never do a `k.Records` lookup.
If we did, we would answer things like ...
`s1.ns.svc.in-addr.arpa.` has `PTR`? => `NOERROR`

miekg: Still think IsReverse is not needed, ip is empty because we didn't find a reverse so your duplicating that check here.

chrisohaver: `IsReverse` returns true if name is in a reverse zone.
`ExtractAddressFromReverse` returns an IP if the name is in a reverse zone _and_ contains a valid IP.

So `ExtractAddressFromReverse('s1.ns.svc.in-addr.arpa.')` is ""  because does not contain a valid IP.

While `IsReverse('s1.ns.svc.in-addr.arpa.')` is true because it's in `in-addr.arpa.`

It differentiates these two cases:

- `s1.ns.svc.in-addr.arpa.` has a `PTR`? ...  which should return `NXDOMAIN`
- `s1.ns.svc.cluster.local.` has a `PTR`? ... which should return `NOERROR` (if `s1.ns.svc.cluster.local.` has an `A` or `AAAA` record)

miekg: But as we are in the Reverse method don't we already know this?

chrisohaver: Removing the check breaks test...

```
        helpers.go:118: rcode is "NOERROR", expected "NXDOMAIN"
        helpers.go:301: ;; opcode: QUERY, status: NOERROR, id: 4937
                
                ;; QUESTION SECTION:
                ;svc1.testns.svc.0.10.in-addr.arpa.     IN       A
                
                ;; AUTHORITY SECTION:
                0.10.in-addr.arpa.      300     IN      SOA     ns.dns.0.10.in-addr.arpa. hostmaster.0.10.in-addr.arpa. 0 7200 1800 86400 30

```

miekg: So doing some awful printf debugging, the actual msg.Service that comes back here is:
~~~
** REVERSE **
reverse
1
/coredns/arpa/in-addr/10/0/svc/testns/svc1
192.168.1.100
~~~
which indicates this should not be solved here.

chrisohaver: where should it be solved.

miekg: in k.Records()?

chrisohaver:
Thanks! Moving the checks from `Service()` and `Reverse()` into `Records()` ends up collapsing the logic to something much simpler/cleaner.

chrisohaver:
@miekg, ok to merge?

plugin/kubernetes/kubernetes.go
```diff
@@ -274,6 +274,10 @@ func (k *Kubernetes) Records(state request.Request, exact bool) ([]msg.Service,
 		return nil, e
 	}
 
+	if dnsutil.IsReverse(state.Name()) {
```
miekg, line 277: I think this is indeed a better place to put it. Ah yes, we and up here because of the type of the query, instead of in Reverse (as we do for PTR lookups).

miekg (approved):

PR #1559, opened 2018-02-23
return an error for multiple use of some plugins

miekg:
Return plugin.ErrOnce when a plugin that doesn't support it, is called
mutliple times.

WIP: because I need to go through all of the plugins


chrisohaver:
Code would be shorter/cleaner with `for i := 1; c.Next(); i++ {`...

miekg:
This now adds it for:
* cache
* dnssec 
* errors
* forward
* hosts
* nsid

and changes it slightly in
* kubernetes
* pprof
* reload
* root

miekg:
merging as is, holler when stuff breaks.

PR #1562, opened 2018-02-26
List SoundCloud as adopter

grobie:
Let me know if the description is too long and I'll remove the information about the throughput.

@johnbelamaric @miekg


miekg:
perfect, thanks!

Issue #1567, opened 2018-02-27
Build of CoreDNS on the CNCF project  "cross cloud CI" is constantly failing (Makefile.release ?)


fturib:
the CNCF Crosscloud CI Dashboard is failing for CoreDNS.
see here : https://cncf.ci/#/

Looking at the second line, based on the HEAD of the repo, the build is failing with a imports and/or vendorisation on caddy:

**NOTE that they are using Makefile.release for building coredns**

for information, the release line is still on 1.0.5, will be updated soon on 1.0.6.

```
Running with gitlab-ci-multi-runner 1.11.5 (cbfcb5c)
  on CNCF (a7809678)
Using Docker executor with image crosscloudci/debian-go ...
Pulling docker image crosscloudci/debian-go ...
Running on runner-a7809678-project-9-concurrent-1 via 104ad005bc0e...
Fetching changes...
Removing build/
HEAD is now at d9958ab doc updates (#1563)
From https://gitlab.cncf.ci/coredns/coredns
   d9958ab..b1ce9ed  master     -> origin/master
Checking out b1ce9ed6 as master...
Skipping Git submodules setup
$ export BASE_URL=${BASE_URL:-$(echo $CI_PROJECT_URL |  cut -d'/' -f1-3)}
$ export CI_COMMIT_SHA_SHORT=$(echo ${CI_COMMIT_SHA} | cut -c -8)
$ if [ -z "$CROSS_CLOUD_YML" ]; then
$ export CROSS_CLOUD_YML="$CROSS_CLOUD_YML"
$ fi

$ source /opt/local/etc/rvmrc ; source /opt/local/etc/profile.d/rvm.sh ; cp -a /opt/local/dashboard /dashboard ; pushd /dashboard ; source /opt/local/.env ; ./bin/update_dashboard ; popd
/dashboard /builds/coredns/coredns
Project build pipeline id: 8272
Running update_dashboard for job compile for release type head
Using Dashboard API server productionapi.cncf.ci
Project build pipeline id: 8272
json_data: {
  "source_key_project_monitor":
    {
      "source_project_id":"9",
      "source_pipeline_id":"8272",
      "source_pipeline_job_id":"55004",
      "pipeline_release_type":"head",
      "project_build_pipeline_id":"8272",
      "target_project_name":"coredns",
      "child_pipeline":"false",
      "cloud":"azure"
    }
}
JWT enabled
/builds/coredns/coredns
$ ln -s /builds /go/src/github.com
[...]

PR #1570, opened 2018-02-28
Makefile release: depend on godeps

### LINKED ISSUE #1429: makefile doesn't guard against using older version of miekg/dns
I just managed to build coredns 1.0.5 and release it with updating miekg/dns to 1.0.4 :/

Repushing binaries now and docker.

But the check out of the version does not work as intented (i.e. prevent this)

miekg:
godeps sets the external repos to the correct versions meaning we can
build CoreDNS.

fixes #1429 #1567

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -94,12 +94,12 @@ build:
 	@echo Cleaning old builds
 	@rm -rf build && mkdir build
 	@echo Building: darwin $(VERSION)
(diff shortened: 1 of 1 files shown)


fturib (approved):
I am not very familiar with the Makefile.
But these change make sense and seems consistent.

let see what the next build on CNCF/CCCI will be.

Thanks!

chrisohaver:
LGTM - and this should be fixed.

Though, regarding #1567, I think the use of `Makefile.release` is misguided in that context.  I think it may be perceived that `Makefile.release` is enabling the building of prior tagged releases.

PR #1575, opened 2018-03-01
plugin/deprecated: add an easy way to error on deprecated plugin

miekg:
See package docs; will send another PR that enables this for startup and shutdown; i.e. trigger erorr on startup.


stp-ip:
Looks good.
/LGMT

PR #1579, opened 2018-03-01
Metrics

miekg:


miekg:
@achille-roussel does something like this work for you?

core/dnsserver/server.go
```diff
@@ -277,6 +285,11 @@ func (s *Server) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 
 	// Wildcard match, if we have found nothing try the root zone as a last resort.
 	if h, ok := s.zones["."]; ok && h.pluginChain != nil {
+
+		// See comment above.
+		ctx = context.WithValue(ctx, plugin.ZoneCtx, ".")
+		ctx = context.WithValue(ctx, plugin.AddrCtx, s.Addr)
```
achille-roussel: What about making a type with both field and using a single context value? It doesn’t seem like the zone and address are ever accessed individually.

plugin/plugin.go
```diff
@@ -107,3 +117,11 @@ var TimeBuckets = prometheus.ExponentialBuckets(0.00025, 2, 16) // from 0.25ms t
 
 // ErrOnce is returned when a plugin doesn't support multiple setups per server.
 var ErrOnce = errors.New("this plugin can only be used once per Server Block")
+
+type info string // info the key used to retrieve zone and addr info from the context.
```
achille-roussel: The context package advises to use pointers to struct types as keys to avoid any conflicts that could come from comparing plain values like strings.

achille-roussel (approved):
Yep, that would help :+1:

miekg:
Thanks for the comments; I'll rework this a bit.

PR #1585, opened 2018-03-02
plugin/health: make reload work

### LINKED ISSUE #1473: plugin/health: does not support reload
The plugin Health keep listening on the same IP, during all life of coredns.
If the IP is changes in the corefile, at reload that new address will not be take into account.
NOTE: whatever the way to reload - plugin RELOAD or signal USR1

To consider : if we try to fix it, then we MUST stop listening during the change of configuration. And if the new configuration fails to load, then the HTTP service will be down until a new valid configuration is reloaded

miekg:
Remove the once.Do from the startup, so we can re-bind the HTTP
listener. Also clarify the usage of health in multiple server blocks
(this is not the best approach - but there isn't a generic solution at
this point).

Manual tested as we lack testing infra, i.e kill -SIGUSR1 and some
CURLing of the health endpoint.

Fixes #1473

### DIFF
--- a/plugin/health/README.md
+++ b/plugin/health/README.md
@@ -74,7 +74,7 @@ Set a lameduck duration of 1 second:
 
 ~~~ corefile
 . {
-    health localhost {
+    health localhost:8092 {
         lameduck 1s
     }
 }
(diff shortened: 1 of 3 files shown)


miekg:
possible also fixes #1456

PR #1587, opened 2018-03-02
plugin/reload: enable

miekg:
Couldn't revert the original commit, just manually perform the needed
tweaks.


rajansandeep:
@miekg  I've enabled the reload plugin in #1485 for the tests to pass. I think that PR is also ready to be merged.

PR #1588, opened 2018-03-03
Metrics2

miekg:
see commit logs...


miekg:
@chrisohaver I updated docker on drone.coredns.io (came with apt-get upgrade)

PR #1591, opened 2018-03-06
Use apk's no-cache flag in Dockerfile

unkaktus:
Simplifies Dockerfile.

None.

None.


miekg:
don't we need `-update` to refresh the versions of the packages we're fetching?

unkaktus:
@miekg no, index is being fetched on-demand.
Also see https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache.

PR #1592, opened 2018-03-06
Allow DNS name to be specified in place of IP address (forward plugin)

smira:

Configuring DNS upstreams via DNS names might be considered a bad idea,
but sometimes it's inevitable - if DNS server runs in a container platform,
getting fixed IP for a service might be really tricky.

This change tries to keep backwards compatibility - for non-IP config
entries files are checked first, and only if file doesn't exist and string
looks like DNS name, it's used as input.

I've verified this `forward` plugin, and it works as expected.

None that I'm aware of.

There should be a lot of documentation and test updates, but I would like to see if such change might be considered appropriate or not.

Sample config:

```
service.consul {
    log
    errors
    forward service.consul marathon.service.example.com:53 {
         force_tcp
    }
}
```

### DIFF
--- a/plugin/forward/setup.go
+++ b/plugin/forward/setup.go
@@ -109,7 +109,7 @@ func parseForward(c *caddy.Controller) (*Forward, error) {
 
 		// If parseHostPortOrFile expands a file with a lot of nameserver our accounting in protocols doesn't make
 		// any sense anymore... For now: lets don't care.
-		toHosts, err := dnsutil.ParseHostPortOrFile(to...)
+		toHosts, err := dnsutil.ParseHostPortOrFileOrDNSName(to...)
 		if err != nil {
 			return f, err
 		}
--- a/plugin/forward/setup_test.go
+++ b/plugin/forward/setup_test.go
@@ -28,7 +28,7 @@ func TestSetup(t *testing.T) {
 		{"forward . [::1]:53", false, ".", nil, 2, false, ""},
 		{"forward . [2003::1]:53", false, ".", nil, 2, false, ""},
 		// negative
-		{"forward . a27.0.0.1", true, "", nil, 0, false, "not an IP"},
+		{"forward . -27.0.0.1", true, "", nil, 0, false, "not an IP"},
 		{"forward . 127.0.0.1 {\nblaatl\n}\n", true, "", nil, 0, false, "unknown property"},
 		{`forward . ::1
 		forward com ::2`, true, "", nil, 0, false, "plugin"},
--- a/plugin/pkg/dnsutil/host.go
+++ b/plugin/pkg/dnsutil/host.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"regexp"
 
 	"github.com/miekg/dns"
 )
@@ -12,6 +13,17 @@ import (
 // address:port or a filename. The address part is checked and the filename case a
 // resolv.conf like file is parsed and the nameserver found are returned.
 func ParseHostPortOrFile(s ...string) ([]string, error) {
+	return parseHostPortOrFileOrDNSName(false, s...)
+}
+
+// ParseHostPortOrFileOrDNSName parses the strings in s, each string can either be a address,
+// address:port or a filename or a DNS name. The address part is checked, DNS name is validated
+//  and the filename case a resolv.conf like file is parsed and the nameserver found are returned.
+func ParseHostPortOrFileOrDNSName(s ...string) ([]string, error) {
+	return parseHostPortOrFileOrDNSName(true, s...)
+}
+
+func parseHostPortOrFileOrDNSName(allowDNSName bool, s ...string) ([]string, error) {
 	var servers []string
 	for _, host := range s {
 		addr, _, err := net.SplitHostPort(host)
@@ -19,11 +31,18 @@ func ParseHostPortOrFile(s ...string) ([]string, error) {
 			// Parse didn't work, it is not a addr:port combo
 			if net.ParseIP(host) == nil {
 				// Not an IP address.
+				if allowDNSName && isDNSName(host) {
+					// consider address to be DNS name
+					servers = append(servers, net.JoinHostPort(host, "53"))
+					continue
+				}
+
 				ss, err := tryFile(host)
 				if err == nil {
 					servers = append(servers, ss...)
 					continue
 				}
+
 				return servers, fmt.Errorf("not an IP address or file: %q", host)
 			}
 			ss := net.JoinHostPort(host, "53")
@@ -33,11 +52,18 @@ func ParseHostPortOrFile(s ...string) ([]string, error) {
 
 		if net.ParseIP(addr) == nil {
 			// No an IP address.
+			if allowDNSName && isDNSName(host) {
+				// consider address to be hostname
+				servers = append(servers, host)
+				continue
+			}
+
 			ss, err := tryFile(host)
 			if err == nil {
 				servers = append(servers, ss...)
 				continue
 			}
+
 			return servers, fmt.Errorf("not an IP address or file: %q", host)
 		}
 		servers = append(servers, host)
@@ -48,7 +74,7 @@ func ParseHostPortOrFile(s ...string) ([]string, error) {
 // Try to open this is a file first.
 func tryFile(s string) ([]string, error) {
 	c, err := dns.ClientConfigFromFile(s)
-	if err == os.ErrNotExist {
+	if os.IsNotExist(err) {
 		return nil, fmt.Errorf("failed to open file %q: %q", s, err)
 	} else if err != nil {
 		return nil, err
@@ -59,6 +85,13 @@ func tryFile(s string) ([]string, error) {
 		servers = append(servers, net.JoinHostPort(s, c.Port))
 	}
 	return servers, nil
+}
+
+const dnsName string = `^([a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62}){1}(\.[a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62})*[\._]?$`
+
+// Verify that string validates as DNS name
+func isDNSName(s string) bool {
+	return regexp.MustCompile(dnsName).MatchString(s)
 }
 
 // ParseHostPort will check if the host part is a valid IP address, if the


miekg:
What's resolving the name? The upstreams you happen to have in /etc/resolv.conf?  And what is re-resolving the name when it changes, i.e. now the resolve is only done on startup...

miekg:
Think this still worthwhile functionality; we could route the request into ourselves as we do with a bare `upstream` in 1.0.6.

But it will still require a way of re-resolving names.

smira:
> What's resolving the name? The upstreams you happen to have in /etc/resolv.conf? And what is re-resolving the name when it changes, i.e. now the resolve is only done on startup...

In this implementation, yep, it's using system resolver. I was taking `forward` plugin as an example, as it is tracking upstream status and doing healthchecks, it should detect broken connections and attempt `dns.Dial` which should lead to re-resolving the name.

miekg:
So for this kind of functionality you cannot depend on the system resolver because that may or may not work.

What do you mean with taking the *forward* pluign as an example? As in: if I were to work on this PR I would need to use the forward plugin to resolve names?

miekg:
Ok. I think I understand now; you pointed /etc/resolv.conf to CoreDNS and this happens to work OK?

smira:
> Ok. I think I understand now; you pointed /etc/resolv.conf to CoreDNS and this happens to work OK?

yep, it's just relying on Go DNS resolver (as coredns is built with cgo disabled). 

So it goes simply via `dns.Dial` which in turn calls `net.Dial` which does resolving if it's not IP address. So it works just fine in my test environment.

I agree it would be more beneficial to keep resolving more configurable, but at the same time one can run use another CoreDNS (or other DNS server) and put it as resolver in `/etc/resolv.conf` (e.g. easy to override DNS settings for Docker container).

miekg:
yeah, I'm not sure if we want to make this supported (by allowing hostnames). If you look at the current *proxy* plugin, it has a lot of code for `https_google` to re-resolve names; that code works, but proved to be problematic with health checks. I.e. what I'm getting at is that this a) adds a lot of complexity b) depends on your environment to work. If CoreDNS is compiled differently does it still work? c) because of b. will add to the ticket/issue load.

I can get behind a mechanism that allows you to do this, but
* It needs to work in all cases and not depend on your environment  
* Be explicit on how names are resolved
* Re-resolves names when TTLs are about to expire

fantuz:
I believe that @smira might refer at some common practice in parallel with CURL directive RESOLVE.

In CURL, RESOLVE and CONNECT clauses, along with DNS_CACHE TIMEOUT, trow everything in the blender, is the same exact concept of DNS example you bring along:
 - such fuzzy solution might work if you are careful to store the TTL + refresh query in time + if you are lucky to get response.... not applicable on larger scale, and not desirable as unreliable.

I didn't test it and personally I am against the double-trouble of re-resolving names. I understand the Docker problematic, but there are ways provided within Docker to maintain IP stable and so hereafter...

Honestly, why did we spent 35 years putting IP in /etc/hosts, if it was just practical to use hostnames in place ?? ....

As @miekg says, might be difficult to force the host to renew the query BEFORE TTL expires. Incredibly dangerous to believe a stale TTL, as after that expires, you take a leap of faith in believing the host did not move elsewhere. @smira You discuss about ARPANET not DNS enabled technologies, sorry.

The RESOLVE directive is applied once at startup in CURL, but maybe there is a specific SIGNAL to request a change ?? not sure, no. RESOLVE is meant to be somehow static...... as it looks to be in coredns now, and I agree with choice (even more, I believe you shall configure a pool of resolvers, and never a single one).

I gues a combination of the following CURLOPTS might somehow workout, in a CURL paradigm.
Yet, double configuration will be necessary, so I see it as loss of time and counter-standard.

Original DNS works on the 13 root servers, all IPs are known by heart. So why names @smira ??

https://curl.haxx.se/libcurl/c/CURLOPT_RESOLVE.html
https://curl.haxx.se/libcurl/c/CURLOPT_DNS_CACHE_TIMEOUT.html
https://curl.haxx.se/libcurl/c/CURLOPT_CONNECT_TO.html
https://curl.haxx.se/libcurl/c/CURLOPT_NOSIGNAL.html


And here is a reference to a Dockery answer to your issue, or the opposite of it more precisely :+1: 
```
# head -1 /etc/sysconfig/docker
OPTIONS="--dns=10.0.0.10 --dns=10.0.0.11 --dns-search=example.com --dns-opt=use-vc"
Specifying use-vc in resolv.conf tells the resolver to use TCP instead of UDP.
```

from https://forums.docker.com/t/dns-resolution-not-working-in-containers/36246/2
I know it doesn't fully solve your problem, but is an implementation guideline.
More, the problem of "not getting the good/fixed IP" in a docker is a DOCKER-specific issue, and an administration topic, not a DNS thing at all IMHO.

Let's imagine you configure the TTL for your 2 NS to 5 sec (bare minimum, if one cannot keep a DNS running 5 seconds at least, that one should consider changing job).

You can even have 20 NS, follow my example till the end and you'll figure out why is wrong: docker hosts go down, one by one, very quickly within the same 5 seconds irrespectively of valid or invalid TTL+IP.
This mapping is demanded to your utilities, as DNS and docker are absolutely not tight together.

In that moment, the bloody 5 seconds of cutting cable, loss of power, added delay due to parh migration etc ...... bind/coredns (the main auth/publisher DNS server outside docker in your example) cannot follow the speed of changing, docker publication mechanisms failed as well, and docker itself luckily restarts, let's say in a minute.

By then, you are left out of your own NS, loosing all of them and not kwnowing where they ended up (as you have 20 servers and 100 IPs, just for example) ....

No-one knows where the battery of your 20 dockery NS are, none of them might be mapped to any stale zone file. End of story.

By best practices, TTL for NS as well, should be guaranteed to be higher than other (very) shortened TTL, say 15 to 60 minutes at least. otherwhise the risk of negative caches and false-positive will make your worldwide customers UNABLE or SLOW to reach any of your server, on a regular basis if you do not help docker to correctly publish onto authoritative servers AND if you keep using wide IP range.
I undertand docker is cool, i also use it a lot, but this request seems a bit redundant, and even when part of a plugin, could considerably charge the load 2x roughly.

DNS is a serious thing, try to make your DOCKER setup more reliable, IMHO is not a coredns issue neither a feature. is totally disruptive, I am strongly against this re-resolving even when referring to more modern DNS services (DNS-over-HTTP for example) and even if it makes more "user-friendly".

Please read this https://docs.docker.com/v17.09/engine/userguide/networking/configure-dns/ 
this https://docs.docker.com/v17.09/engine/userguide/networking/default_network/configure-dns/
and this https://development.robinwinslow.uk/2016/06/23/fix-docker-networking-dns/

and tell us about a dig against 127.0.0.11
127.0.0.11
yes is .eleven.

Cheers, best regards,
Max

fantuz:
not saying is not a KISS and "configuration-easening" idea, but overcomplicates and confonds purposes of DNS...  ok, inside a plugin could make sense, but is misleading and creates extra load, which is opposite of optimising.

smira:
@fantuz my use-case is not about providing public DNS service. This is all inter-cluster DNS inside one DC. Say I have pretty small but important DNS server running on container platform (let's take K8s as an example). This DNS server is doing some simple mapping which I need to consume only internally inside my network (not a public service at all). This DNS server is using platform upstream DNS servers which have fixed IPs (no problem here), but as it's running on container platform it doesn't have fixed IP, but it has reliable service discovery (provided by the platform), let's say it's `private-dns.service`.

Now on another compute node I want to forward resolving of some DNS names to my specific `private-dns.service`. I don't have fixed IP, and all I have is name. I can do tricks with wrappers which resolve name to IP, write config, and restart service/re-generate configs if IP has changed. Or I could try to provision fixed IP for `private-dns.service`, but it's even more hacky.

In the end the best solution I have is to forward DNS resolving for some names to my `private-dns.service`. So I ended up with `coredns` and this simple patch. Yes, if service discovery fails and `private-dns.service` is not resolvalbe, my DNS forwarding would be broken, but that's fine as if service discovery is broken, I have bigger problem and my application won't work anyway. 

My problem from bird eye view is pretty simple. I have public-facing service with public-facing DNS which resolves service name to public IP. This works fine, but I need to consume my service internally (within same site), and I can't use public IPs due to various networking/routing issues. As I'm accessing service over HTTPS, I want to keep same hostname as public endpoint, but I need it to resolve to private IPs. So my solution was to build specific internal DNS service which resolves public endpoints to private IPs, and forward DNS resolving internally to that service.

johnbelamaric:
Yes, I agree that in an environment like K8s this sort of thing is very helpful. In fact, even better would be to be able to use an SRV request to get the IP/port combinations. So, for example, if I am running unbound as a headless service to provide recursive DNS, I can forward to that by name.

smira:
@fantuz I agree that this "feature" might be dangerous and take down public DNS service if it is used when it shouldn't be used, so I'm open to changing this PR to limit it to specific plugin configuration where it might be more "safe" (e.g. limiting it only to `forward` or `proxy` plugins)

miekg:
yes, doing SRV would also be nice(r) (and have a fallback to A/AAAA).

I struggling with coming up with an idea on how to do this fully in CoreDNS, maybe a (special) plugin that loads before others that can be used to intercept such queries and resolves them in whatever way you want (default to /etc/resolv.conf).

We then still need some code to re-resolve after X seconds (and update the IP addresses use in the other plugins). This requires some fleshing out (code and/or design wise)

johnbelamaric:
We have general use cases like this, not just forward/proxy.

For example in the [policy plugin](https://github.com/infobloxopen/themis/tree/master/contrib/coredns/policy) we talk to a grpc server for policy decisions. Right now we list all the IP/port combos in the Corefile and have some other code watching k8s that rewrites the config file. That wouldn't be necessary with something like this. And if it could do things *other* than pure DNS with period re-resolving (like k8s API watch, or, dare-i-say, CoreDNS grpc watch - #1527) that would be nifty.

miekg:
@smira allowing hostnames in the config is the easy part. Are you up to the task for fleshing out some kind of mechanism which allows coredns to resolve these names by itself (and updating them?)
(I have only vague ideas at this moment - as mentioned in previous comments, but I'll gladly help, probably don't have enough time to push this forward by myself)

smira:
@miekg I can work on that if you could provide me with some background on that. So far I haven't looked deep enough in every plugin to understand the consequences. 

If CoreDNS is establishing outbound connection to some other service (e.g. `forward` doing outbound DNS requests to other DNS service), it using `net.Dial` in the end which does resolving automatically via system resolver. If connection fails, times out, it gets re-established which forces re-resolving of the hostname. This is basically same as with every other service talking to other service, I'm not sure I follow why CoreDNS requires special treatment here. As I'm coming from different background, probably I don't see the full picture.

miekg:
> @miekg I can work on that if you could provide me with some background on that. So far I haven't looked deep enough in every plugin to understand the consequences.

I don't think that will be needed initially; just try to make one plugin work and go from there; otherwise things will become too complex anyway.

> If CoreDNS is establishing outbound connection to some other service (e.g. forward doing outbound DNS requests to other DNS service), it using net.Dial in the end which does resolving automatically via 
system resolver.

The thing to keep in mind when writing a DNS server is **there is no system resolvers** and if there is one we might be it; IOW: any resolving that may occur must be handled directly by CoreDNS otherwise there is a high change of not working.

> If connection fails, times out, it gets re-established which forces re-resolving of the hostname.

True.. and good point; except hooking into net.Dial might be tricky... I.e. we still want CoreDNS to resolve - not the system's resolver.

> This is basically same as with every other service talking to other service, I'm not sure I follow why CoreDNS requires special treatment here. As I'm coming from different background, probably I don't see the full picture.

See above, but CoreDNS is special because it's DNS a server that can resolve names and it may be the only thing on the system capable of doing so; there is no systems resolver that is functional

smira:
> The thing to keep in mind when writing a DNS server is there is no system resolvers and if there is one we might be it; IOW: any resolving that may occur must be handled directly by CoreDNS otherwise there is a high change of not working.

This is not my use-case, in our environment we have at least 2 layers of DNS services, and I'm building 3rd layer which sits on top of those two to provide additional features, so I actually want CoreDNS to rely on system resolver to handle resolving of upstream DNS hostname. 

I think if system resolver is this CoreDNS instance, or if it's not available, using hostnames instead of IPs might lead to things being really unreliable, including loops while performing lookups.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Allow hostnam..." ]
>> The thing to keep in mind when writing a DNS server is there is no system resolvers and if there is one we might be it; IOW: any resolving that may occur must be handled directly by CoreDNS otherwise there is a high change of not working.
>
>This is not my use-case, in our environment we have at least 2 layers of DNS services, and I'm building 3rd layer which sits on top of those two to provide additional features, so I actually want CoreDNS to rely on system resolver to handle resolving of upstream DNS hostname.

Ok, but what happens to work for you is not the generic case we should optimize 
for.

>I think if system resolver is this CoreDNS instance, or if it's not available, using hostnames instead of IPs might lead to things being really unreliable, including loops while performing lookups.

We've started doing for `upstream` recently and check for local loops, so this 
could tie into that.

smira:
I've updated PR to limit changes only to `forward` plugin, so that other plugins are not affected.

rzenker:
+1 empower users to handle this use-case. If they loop their config, it is not a bug in coredns.

miekg:
@rzenker understood, but they might still complain on github and that might add considerable load on the maintainers; I rather have a more robust solution in place; that a) works when it works :-) b) detects and tells why it *didn't* work.

miekg:
Closing. If this is desirable (which I can believe it is); we need a robust solution, not something that works because of configuration that happens to be present.

PR #1596, opened 2018-03-09
Add DataDog as a provider for trace plugin

elbuo8:

Adds support for the DataDog Trace Agent in the CoreDNS Trace plugin.

Didn't see any :/

https://coredns.io/plugins/trace/ needs to include it as a supported one. Zipkin is still the default.


johnbelamaric (changes requested):
Ok, looks good, thanks. Get the checks passing and also modify the README.

johnbelamaric (approved):

Issue #1357, opened 2018-01-07
plugin/kubernetes: DNS records for services


miekg:
I've got this service definition:
~~~
apiVersion: v1
kind: Service
metadata:
  name: coredns
spec:
  ports:
  - name: coredns
    port: 53
    protocol: UDP
    targetPort: 53
  selector:
    app: coredns
  type: LoadBalancer
  loadBalancerIP: 192.168.1.252
~~~

describing the service shows:
~~~
% kubectl describe service coredns
Name:                     coredns
Namespace:                default
Labels:                   <none>
Annotations:              kubectl.kubernetes.io/last-applied-configuration={"apiVersion":"v1","kind":"Service","metadata":{"annotations":{},"name":"coredns","namespace":"default"},"spec":{"loadBalancerIP":"192.168.1.252","port...
Selector:                 app=coredns
Type:                     LoadBalancer
IP:                       10.96.77.39
IP:                       192.168.1.252
LoadBalancer Ingress:     192.168.1.252
Port:                     coredns  53/UDP
TargetPort:               53/UDP
NodePort:                 coredns  30332/UDP
Endpoints:                10.244.1.46:53,10.244.1.47:53,10.244.2.51:53 + 1 more...
Session Affinity:         None
External Traffic Policy:  Cluster
Events:                   <none>
~~~

A dig returns the clusterIP, so far so good:
~~~
% dig coredns.default.svc.cluster.local.
...
coredns.default.svc.cluster.local. 5 IN	A	10.96.77.39
~~~
For a headless service I would get all pod IPs (so far so good).

Now I query: `*.coredns.default.svc.cluster.local`, which does return the pods:

~~~
*.coredns.default.svc.cluster.local. 5 IN A	10.244.1.46
*.coredns.default.svc.cluster.local. 5 IN A	10.244.1.47
*.coredns.default.svc.cluster.local. 5 IN A	10.244.2.51
*.coredns.default.svc.cluster.local. 5 IN A	10.244.3.41
~~~
Note that this is super handy, but it's a bit strange and makes headless vs a normal service rather special.

chrisohaver:
This is expected, and behaves per k8s cluster dns spec.
You can query endpoint dns names directly regardless of what type of service they sit behind.

rikatz:
As this is expected (and a great feature), this is not part of https://github.com/kubernetes/dns/blob/master/docs/specification.md

[...]

PR #1605, opened 2018-03-13
Document the service with ClusterIP wildcard

rikatz:
This PR documents the trick exposed in coredns/coredns#1357 about using wildcards to get all POD IPs from a service with ClusterIP (as used in 'endpoints')

#1357 

This is a Documentation change :)


chrisohaver (approved):
LGTM

PR #1606, opened 2018-03-13
Add PTR record support for Route53 plugin

### LINKED ISSUE #1595: Expanding the Route 53 plugin RR type coverage
Hi,

It would be interesting for me to have the route 53 support other RR types than A and AAAA. I'm specifically interested in PTR records and zones.

So I'm trying to understand what, if any, issues there may be in expanding the functionality of the plugin.

Looking at the plugin implementation, it seems it should be possible to expand with functions handling the additional types. But I don't have much experience with neither the Route 53 API, CoreDNS or Go, so that may just be my unawareness getting in the way.

Does anyone have any thoughts? Consider this an invite for discussion :)

yongtang:
This fix adds PTR record support for Route53 plugin

This fix fixes #1595.

### DIFF
--- a/plugin/route53/route53.go
+++ b/plugin/route53/route53.go
@@ -50,6 +50,8 @@ func (rr Route53) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 		answers = a(qname, output.ResourceRecordSets)
 	case dns.TypeAAAA:
 		answers = aaaa(qname, output.ResourceRecordSets)
+	case dns.TypePTR:
+		answers = ptr(qname, output.ResourceRecordSets)
 	}
 
 	if len(answers) == 0 {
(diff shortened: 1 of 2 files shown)


johnbelamaric (approved):
The change LGTM. But I noticed that MaxItems is set to 1 for everything. Why?

yongtang:
AWS will always output every records and will not allow you to limit the number of output based on filter for type restriction alone. However, they do allow you to specify the `StartRecordType` (e.g. `PTR` or `AA`) so that the first type returned will be the type you desire.

johnbelamaric:
So you are just supporting a single record of a given type and FQDN?

yongtang:
One record for now. Don't know if AWS have any restrictions on the records returned. May need to verify with reference.

PR #1611, opened 2018-03-14
plugin/secondary: fix for #1609

### LINKED ISSUE #1609: plugin/secondary: Denial-of-Service via endless Zone Transfer

miekg:
Fixes #1609

With this Corefile:

~~~
example.com {
    bind 127.0.0.1
    log
    cache
    secondary {
        transfer from 127.0.0.53
    }
}

. {
    bind 127.0.0.1
    forward . 8.8.8.8
    log
    cache
}
~~~

starting: `% ./coredns -dns.port=1053`

queries flow:

~~~
example.com.:1053 on 127.0.0.1
.:1053 on 127.0.0.1
2018/03/14 18:24:59 [INFO] CoreDNS-1.1.0
2018/03/14 18:24:59 [INFO] linux/amd64, go1.10,
CoreDNS-1.1.0
linux/amd64, go1.10,
127.0.0.1:35698 - [14/Mar/2018:18:25:26 +0000] 15569 "MX IN miek.nl. udp 37 false 4096" NOERROR qr,rd,ra,ad 170 57.437457ms
127.0.0.1:37304 - [14/Mar/2018:18:25:28 +0000] 27556 "MX IN miek.nl. udp 37 false 4096" NOERROR qr,rd,ra,ad 170 104.4µs
127.0.0.1:34495 - [14/Mar/2018:18:25:30 +0000] 18816 "MX IN miek.nl. udp 37 false 4096" NOERROR qr,rd,ra,ad 170 92.186µs
~~~

Without this patch:
~~~
% ./coredns -dns.port=1053
<nothing>
~~~

### DIFF
--- a/plugin/file/secondary_test.go
+++ b/plugin/file/secondary_test.go
@@ -83,8 +83,7 @@ func TestShouldTransfer(t *testing.T) {
 	}
 	defer s.Shutdown()
 
-	z := new(Zone)
-	z.origin = testZone
+	z := NewZone(testZone, "stdin")
 	z.TransferFrom = []string{addrstr}
 
 	// when we have a nil SOA (initial state)
@@ -128,9 +127,7 @@ func TestTransferIn(t *testing.T) {
 	}
 	defer s.Shutdown()
 
-	z := new(Zone)
-	z.Expired = new(bool)
-	z.origin = testZone
+	z := NewZone(testZone, "stdin")
 	z.TransferFrom = []string{addrstr}
 
 	err = z.TransferIn()
--- a/plugin/file/zone.go
+++ b/plugin/file/zone.go
@@ -20,7 +20,7 @@ type Zone struct {
 	origLen int
 	file    string
 	*tree.Tree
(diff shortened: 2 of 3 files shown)


johnbelamaric:
So, now we start serving traffic even if the transfer is not done yet? I assume that blocking was intentional. Maybe it should only block for that zone instead of all zones (so put some sort of flag in to say if initial transfer is complete...do we have that already?).

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/second..." ]
>So, now we start serving traffic even if the transfer is not done yet? I assume that blocking was intentional. Maybe it should only block for that zone instead of all zones (so put some sort of flag in to say if initial transfer is complete...do we have that already?).

That can happen on expire as well, so it's not unique to the starting up phase. 
But yes, it will now return SERVFAIL.

miekg:
The race it shows in Travis is already there as transferIn is called in a
goroutine, but not tested.

Issue #1618, opened 2018-03-18
SIGUSR1: listen tcp:8080: bind: address already in use


yybear:
i have tow questions:
1. the coredns1.1.0, i enabled the plugin of reload, after sending SIGUSR1, then log shows:

[INFO] SIGUSR1: Reloading
[INFO] Reloading
[ERROR] SIGUSR1: listen tcp:8080: bind: address already in use
how to fix this?

2. other question was at first time i used coredns1.0.4, in this version the plugin of reload can't be used.
so no above problem, but after "Reloading complete", i found this was no effect, i had to recreate coredns pod(removing pod then kubernetes recreated new pod) to make corefile changes to work, may i miss some config item?
my corefile like this (when used coredns1.0.4 )
. 53 {
      error
      health
      cache 30
      kubernetes {..}
      ...
}

fturib:
This issue is the same (I think) that #1604 

NOTE: there are now 2 ways to do a RELOAD:
- using SIGUSR1 (always been there, it is defined by the lib Caddy)
- the plugin RELOAD that aims to automatically reload if there is a change in the Corefile.

The both use the same mechanism.

Although SIGUSR1 was there since long time, it is only when testing RELOAD that we found some of the plugin are not compatible with the reload process - and we are in progress of tuning these plugins.

About your questions

1. #1621 should address your problem with SIGUSR1 and Health. (I understand here you are using the signal, not the RELOAD plugin).

2. I understand that for CoreDNS 1.0.4 used inside K8s, the reload using SIGUSR1 was not working. Where you modifing the Corefile inside the pod - or the Configmap named "coredns" using kubectl ?

yybear:
with CoreDNS 1.0.4, i used the ConfigMap to store config of dns.
so after changed ConfigMap, i used SIGUSR1 to reload coredns

yybear:
you said "for CoreDNS 1.0.4 used inside K8s, the reload using SIGUSR1 was not working", so how to make CoreDNS reload config?
I use the ConfigMap to store the Corefile, but the name is not "coredns", my situation is a bit special. there will be some CoreDNSs for different servers, I will change the corefile to add some ip and domain pairs  by myself, so I hope can reload Corefile to make it sense.
if  using SIGUSR1 was not working, is there other method?

fturib:
I cannot say why SIGUSR1 was not working, for you, using the 1.0.4
[...]

Issue #1604, opened 2018-03-13
plugin/health plugin/prometheus graceful restart bug


UladzimirTrehubenka:
Tested on 3e3d8cc845b8e072efc75043c9aa3df318ec3bf3 with Corefile
```
.:53 {
	health 127.0.0.1:8080
	proxy . 8.8.8.8:53
}
```
After graceful restart with "killall -USR1 coredns" we get an error:
```
[INFO] SIGUSR1: Reloading
[INFO] Reloading
[ERROR] SIGUSR1: listen tcp 127.0.0.1:8080: bind: address already in use
```

fturib:
Hi Uladimir,do you have a change in your Corefile between first start and the reload ?
If, yes, can you add both Corefile ?

UladzimirTrehubenka:
I didn't change Corefile between start and reload.

fturib:
@UladzimirTrehubenka : I mismatch the use case. I though was tied to BIND. But in fact your issue is tide to HEALTH plugin.

I guess that there is a complete restart of the Health HTTP listener on port 8080. 
To know : the restart of CoreDNS,  there is a "graceful" restart, which means the former service is not shutdown before the new one is established and validated. From there this issue on listening a second time the same port.
NOTE: for the main listeners (the DNS request listeners), there is some magic in Caddy to allow a second service during the graceful restart. That magic does not happen on the listener created at setup or start by the plugins (like HEALTH, METRICS, ...). The plugin has to deal with check the address before reloading, or not allowing change of address at reload ...

UladzimirTrehubenka:
BTW with Corefile
```
.:53 {
	proxy . 8.8.8.8:53
	prometheus 127.0.0.1:9253
}
```
We didn't get any error in logs but after graceful restart we get
```
curl 127.0.0.1:9253/metrics
curl: (7) Failed to connect to 127.0.0.1 port 9253: Connection refused
```
So this is another one bug.

UladzimirTrehubenka:
Looks like all plugins that use Listen() should do magic with clone binding port (like CoreDNS does for dns port).

ephur:
See this same behavior using the reload plugin when health plugin is enabled (others like prometheus are as well):

2018/03/17 05:23:03 [INFO] Reloading   
2018/03/17 05:23:03 [ERROR] Corefile changed but reload failed: listen tcp :8080: bind: address already in use

miekg:
[...]

PR #1621, opened 2018-03-19
WIP - Allocator for Graceful TCP servers - issue #1604 and #1618

fturib:
Implement a Listener Allocator for of TCPListeners for additional Servers used by plugin (Health, Metrics, ...) that is able to handle graceful restarts.

WIP : 
- Health is modified to use this allocator. (also modified to run only once per BlocServer)
- Metrics not yet updated 

Need to dive into problem with reallocating the SAME port but for a different usage. In that case the current graceful restart would not work.

#1604 
#1618 

not sure which one here.

### DIFF
--- a/plugin/health/health.go
+++ b/plugin/health/health.go
@@ -2,6 +2,7 @@
 package health
 
 import (
+	"github.com/coredns/coredns/plugin/pkg/listener"
 	"io"
 	"log"
 	"net"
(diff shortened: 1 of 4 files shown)


miekg:
Can we as another avenue see if we can pass the open sockets from parent to child process when we reload? I.e. in the same manner as the reload from caddy does this? Or would we then still hit "address in use issues"?

fturib:
@miekg : yes it is what I implemented so far. And that work for the HTTP Listeners.

I am pulled on others tasks, that is why I am later here to finish.
I was to fix the corner cases : 
- when the same socket is reused but by another plugin ... what should we do ?

miekg:
> when the same socket is reused but by another plugin ... what should we do ?

Good question. Panic? I.e. fail to reload and crash?

fturib:
This PR is replaced by #1688.

PR #1624, opened 2018-03-21
plugin/forward: retry udp queries if timed out

rdrozhdzh:
forward plugin will retry sending DNS queries to upstream server if UDP connection was used. The UDP is not reliable protocol, so it might worth to retry several times
In current implementation, in case if only one upstream was configured and if the response from upstream was lost the forward plugin returns SERVFAIL with error "no healthy proxies or upstream error" 

none

none

### DIFF
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -34,21 +34,16 @@ func (p *Proxy) connect(ctx context.Context, state request.Request, forceTCP, me
 		conn.UDPSize = 512
 	}
 
-	to := timeout
-	if proto == "udp" {
-		to = udpTimeout
-	}
-
 	udpRetry := udpRetryCnt
 	var ret *dns.Msg
 	for {
-		conn.SetWriteDeadline(time.Now().Add(to))
+		conn.SetWriteDeadline(time.Now().Add(timeout))
 		if err := conn.WriteMsg(state.Req); err != nil {
 			conn.Close() // not giving it back
 			return nil, err
 		}
 
-		conn.SetReadDeadline(time.Now().Add(to))
+		conn.SetReadDeadline(time.Now().Add(timeout))
 		ret, err = conn.ReadMsg()
 		if err != nil {
 			oe, ok := err.(*net.OpError)
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -5,6 +5,7 @@
 package forward
 
 import (
+	"net"
 	"strconv"
 	"time"
 
@@ -33,20 +34,39 @@ func (p *Proxy) connect(ctx context.Context, state request.Request, forceTCP, me
 		conn.UDPSize = 512
 	}
 
-	conn.SetWriteDeadline(time.Now().Add(timeout))
-	if err := conn.WriteMsg(state.Req); err != nil {
-		conn.Close() // not giving it back
-		return nil, err
+	to := timeout
+	if proto == "udp" {
+		to = udpTimeout
 	}
 
-	conn.SetReadDeadline(time.Now().Add(timeout))
-	ret, err := conn.ReadMsg()
-	if err != nil {
-		conn.Close() // not giving it back
-		return nil, err
+	udpRetry := udpRetryCnt
+	var ret *dns.Msg
+	for {
+		conn.SetWriteDeadline(time.Now().Add(to))
+		if err := conn.WriteMsg(state.Req); err != nil {
+			conn.Close() // not giving it back
+			return nil, err
+		}
+
+		conn.SetReadDeadline(time.Now().Add(to))
+		ret, err = conn.ReadMsg()
+		if err != nil {
+			oe, ok := err.(*net.OpError)
+			if ok && oe.Timeout() && proto == "udp" && udpRetry > 0 {
+				udpRetry--
+				continue
+			}
+			conn.Close() // not giving it back
+			return nil, err
+		}
+		break
 	}
 
-	p.Yield(conn)
+	if udpRetry != udpRetryCnt {
+		conn.Close() // to skip stale msgs which might come later
+	} else {
+		p.Yield(conn)
+	}
 
 	if metric {
 		rc, ok := dns.RcodeToString[ret.Rcode]
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -47,9 +47,11 @@ func (p *Proxy) connect(ctx context.Context, state request.Request, forceTCP, me
 		ret, err = conn.ReadMsg()
 		if err != nil {
 			oe, ok := err.(*net.OpError)
-			if ok && oe.Timeout() && proto == "udp" && udpRetry > 0 {
-				udpRetry--
-				continue
+			if ok && oe.Timeout() {
+				if _, ok = conn.Conn.(*net.UDPConn); ok && udpRetry > 0 {
+					udpRetry--
+					continue
+				}
 			}
 			conn.Close() // not giving it back
 			return nil, err
--- a/plugin/forward/proxy.go
+++ b/plugin/forward/proxy.go
@@ -89,4 +89,6 @@ const (
 	dialTimeout = 4 * time.Second
 	timeout     = 2 * time.Second
 	hcDuration  = 500 * time.Millisecond
+	udpTimeout  = 500 * time.Millisecond
+	udpRetryCnt = 2
 )
--- a/plugin/forward/proxy.go
+++ b/plugin/forward/proxy.go
@@ -89,6 +89,5 @@ const (
 	dialTimeout = 4 * time.Second
 	timeout     = 2 * time.Second
 	hcDuration  = 500 * time.Millisecond
-	udpTimeout  = 500 * time.Millisecond
 	udpRetryCnt = 2
 )


fantuz:
Hi!

Linux defaults to retry 3 times over 5 seconds timeouts in UDP,
configurable (i.e. lossy networks or very reliable ones) so i think the
retry is a good idea.

Anyhow, talking TCP just for reference for a second, on a similar topic:
the "retransmission timeout" is set in Linux Kernel to be *just* 1 second;
it used to be more reasonable 3 seconds less than 10 years ago and just
some algo adjust or compensate on the fly, other TCP algorithms don't.

Is like we assume we are (all the internet is?) all reachable within 1
second RTT, which itself causes more TCP retransmissions than we think...
ok big sites want to get customer fast, retransmission is not a problem to
them, but they want to rtx too fast, 1 second is "too optimistic". 3 was
good, 1 maybe aggressive.. once TCP goes through he just goes through!!

Sorry for rant.

Similarly, setting low timeouts on UDP applications is not desired over
un-benchmarkable WAN.

Agree with you, with UDP you should retry 3 times in total 15 sec by
default... maybe too long but some DNS are slower than we think :)


Cheers

rdrozhdzh:
Thank you @fantuz for review. Probably you are right regarding UDP timeout. I'm going to revert it back to 2 seconds

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin/forward: r..." ]
>### 1. What does this pull request do?
>forward plugin will retry sending DNS queries to upstream server if UDP connection was used. The UDP is not reliable protocol, so it might worth to retry several times
>In current implementation, in case if only one upstream was configured and if the response from upstream was lost the forward plugin returns SERVFAIL with error "no healthy proxies or upstream error"

Retry might work, but we may have to be more careful here. If you have two
upstream, trying the other might be a better choice then retrying the first one.
You're assuming the network is break; I'm (was) assuming in this code the server 
is breaking

[I haven't looked at the PR]

But we need some clever bits to try the other server (if specified first - or 
something) and then retry? WDYT?

rdrozhdzh:
I see your point.
But indeed, I think in this case (use of  UDP protocol) probability of broken network (lost response) is much higher then probability of server getting down. Especially taking into account that healthckeck mechanism most likely would already detected broken servers and we would not even try to communicate with the broken server at all.
The only case when your approach would give better results is trying to send requests during the fisrt several seconds after upstream server has got down. But I think this is quite rare case that upstream servers go down.
Anyway, if we want first trying next server on timeout error than we need to move retry logic at higher level because we iterate over upstream list directly in ServeDNS() function. And in this case we will have no (or less) control over connection type, i.e. retry logic will probably be the same for all connection types. Also pulling/creating/cleaning connections for another server may give us additional overhead.
So, please let me know which approach is more preferable for you. I believe 'for UDP only' my approach would work better

fantuz:
Linux retries with the next entry in /etc/resolv.conf, in a **sequential** order mostly, not randomised.
Is also a suggested best-practice to have 2 entries in that resolv.conf, rarely needs more.
It can be seen as broken design, to need more than 2 virtual IPs for that basic setup (so, intelligence in the balancers ?).

I guess this is more about defining "best strategy for the server plugin to be (appear) responsive" rather than thinking that network is NOT broken or cannot glitch ;)

And from an outside observer, latency, real distance or real brokenness cannot be distinguished in one-shot request (maybe think of a counter that triggers temporary blacklist of entry ?) ... also brokenness is not forever, the host might be brought back in a relatively short time (dockers?)

miekg:
I'm not sure what this adds to this code: https://github.com/coredns/coredns/blob/master/plugin/forward/forward.go#L76
Which should also requery in the case of 1 upstream.

Do have a test or experience a use-case where this would have made a difference?

rdrozhdzh:
@miekg the code you mentioned is called only in case if all upstreams were initially marked unhealthy. I.e. at this point (line 76) we didn't send query yet. And this random upstream - is the only attempt to send the query.
If initially at least one upstream was healthy, then we won't get to line 76. 

So, if we have configured single upstream then in both cases (was it initially healthy or not) we do only 1 attempt to send query.

I tested on my virtual centos box with following config
```
.:1053 {
    forward . 10.20.30.10:53
    errors
}
```
and created DNS load with dnsperf tool. The output of coredns
```
.:1053
2018/03/23 17:09:05 [INFO] CoreDNS-1.0.6
2018/03/23 17:09:05 [INFO] linux/amd64, go1.9.4, 
CoreDNS-1.0.6
linux/amd64, go1.9.4, 
23/Mar/2018:17:09:14 +0300 [ERROR 2 test.org. A] no healthy proxies or upstream error
23/Mar/2018:17:09:19 +0300 [ERROR 2 test.org. A] no healthy proxies or upstream error
23/Mar/2018:17:09:19 +0300 [ERROR 2 test.org. A] no healthy proxies or upstream error
23/Mar/2018:17:09:19 +0300 [ERROR 2 test.org. A] no healthy proxies or upstream error
23/Mar/2018:17:09:19 +0300 [ERROR 2 test.org. A] no healthy proxies or upstream error
23/Mar/2018:17:09:24 +0300 [ERROR 2 test.org. A] no healthy proxies or upstream error
23/Mar/2018:17:09:24 +0300 [ERROR 2 test.org. A] no healthy proxies or upstream error
23/Mar/2018:17:09:24 +0300 [ERROR 2 test.org. A] no healthy proxies or upstream error
```
The output of dnsperf
```
DNS Performance Testing Tool
Nominum Version 2.0.0.0
(example query data file: /usr/share/dnsperf/queryfile-example-current)

[Status] Command line: dnsperf -s 127.0.0.1 -p 1053 -d dns.txt -l 15 -c 10
[Status] Sending queries (to 127.0.0.1)
[Status] Started at: Fri Mar 23 17:09:08 2018
[Status] Stopping after 15.000000 seconds
[Status] Testing complete (time limit)

Statistics:

  Queries sent:         8149
  Queries completed:    8149 (100.00%)
  Queries lost:         0 (0.00%)

  Response codes:       NOERROR 8141 (99.90%), SERVFAIL 8 (0.10%)
  Average packet size:  request 26, response 41
  Run time (s):         16.270376
  Queries per second:   500.848905

  Average Latency (s):  0.185240 (min 1.813043, max 11.168182)
  Latency StdDev (s):   0.057066
```
The number of lost queries (SERVFAIL) varies from 0.1% to 0.3% in my tests

With my fix I was not able to reproduce the issue anymore

miekg:
>@miekg the code you mentioned is called only in case if all upstreams were initially marked unhealthy. I.e. at this point (line 76) we didn't send query yet. And this random upstream - is the only attempt to send the query.
>If initially at least one upstream was healthy, then we won't get to line 76.
>
>So, if we have configured single upstream then in both cases (was it initially healthy or not) we do only 1 attempt to send query.

Ah, yes good point. The problem with doing the retry one layer deeper is that it 
may incur (another) time out. In any event I think it will be cleaner to try to 
make this behavior happen in forward/forward.go rather than the layer below it.

The one thing I want to prevent is that when you have 2 upstreams (you probably 
should always have 2 upstreams) we will just try to connect to the one that is 
timing out, instead of trying the other one (under the assumption that still 
works).

I.e your patch can be mimicked on config with `upstream . 8.8.8.8 8.8.8.8`

rdrozhdzh:
I thought about repeating upstream addresses in config, but I'm afraid in this case we will break metrics logic

miekg:
>I thought about repeating upstream addresses in config, but I'm afraid in this case we will break metrics logic

Hmmm, metrics is a separate discussion. I'm planning to tackle the whole metrics 
thing in 1.1.2 (will release 1.1.1 soon)

rdrozhdzh:
Actually, in our test cluster with 3 upstream servers under high load we see "no healthy proxies or upstream error" in log several times per minute. So, even 3 attempts may be insufficient in some conditions.

In proxy plugin we retry sending query during 16 seconds. @miekg, would you like me to implement similar approach in forward plugin? Alternatively, we can count number of retries (one counter for all upstreams), or use combination of retry count and retry duration

miekg:
>Actually, in our test cluster with 3 upstream servers under high load we see "no healthy proxies or upstream error" in log several times per minute. So, even 3 attempts may be insufficient in some conditions.

Would be good (but hard) to actually see why this is happening... Is this all 
UDP drops?

>In proxy plugin we retry sending query during 16 seconds. @miekg, would you like me to implement similar approach in forward plugin? Alternatively, we can count number of retries (one counter for all upstreams), or use combination of retry count and retry duration

I think forward is a better proxy and more geared towards DNS like time outs. 
16s is just too long for waiting for an answer; most stubs wait a maximum of 5s.

As previous this is a tricky problem to get right. We should start by checking 
if the error was a timeout, check if not other upstreams have been tried and 
re-query. But this should ideally all happen within 5s; and note that stubs (also) 
retry, but only is *we* don't send an answer, i.e. we fake a timeout on our 
side.

The downside of all this trickery is that we are well into the 2+ second 
timeframe, but maybe it is good to salvage what we can on the proxy side.


/Miek

--
Miek Gieben

miekg:
[ Quoting <miek@miek.nl> in "Re: [coredns/coredns] plugin/forwar..." ]
>[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/forwar..." ]
>>Actually, in our test cluster with 3 upstream servers under high load we see "no healthy proxies or upstream error" in log several times per minute. So, even 3 attempts may be insufficient in some conditions.
>
>Would be good (but hard) to actually see why this is happening... Is 
>this all UDP drops?

Note: I really that this works well, but we need to understand better why this 
is failing.

/Miek

--
Miek Gieben

rdrozhdzh:
BTW, is there any specific reason why we don't return real error from ServeDNS() and mask it with "no healthy proxies or upstream error" instead?

miekg:
>BTW, is there any specific reason why we don't return real error from ServeDNS() and mask it with "no healthy proxies or upstream error" instead?

No, no real reason, we could carry the error we last saw and return that.

Also; I got a setup in Packet.net:i
querperf <--> coredns+forward <--> coredns+whoami

(using atom cpu) and with that setup I can replicate the timouts, 3 of them 
(usually) with 12K qps and 25 seconds of run time (proxy plugin performs worse 
in this setup).

Anyway; yes we can retry on timeout of packet loss, *but* there is also a client 
timeout which is usually 5s, I think we use 2s, so we can just about to 1 retry; 
if we exausted the upstream list.

rdrozhdzh:
What if we reduce UPD timout to 0.5 seconds and will do max 4 attempts to send-receive? We will spend the same 2 seconds time on waiting the first packet, 1.5 seconds on waiting the 2nd packet, 1 sec on 3rd packet and 0.5 sec on 4th packet. It doesn't matter which response comes first, since all these responses are sent for the same query. As soon as we have read the response we'll close the connection and drop remaining responses if they happen to arrive. 
This was the initial logic of my PR, before I restored the timeout to 2 seconds. Do you think this could work for us?

rdrozhdzh:
As a next step, we can retry UPD send-receive in parallel with sending the query via alternative upstream (if any), but this will make code quite complicated, since we will have to synchronize responses from several upstreams

miekg:
>What if we reduce UPD timout to 0.5 seconds and will do 4 attempts so 
>send-receive? We will spend the same 2 seconds time on waiting the first 
>packet, 1.5 seconds on waiting the 2nd packet, 1 sec on 3rd packet and 0.5 sec 
>on 4th packet. It doesn't matter which response comes first, since all these 
>responses are sent for the same query. As soon as we have read the response 
>we'll close the connection and drop remaining responses if they happen to 
>arrive.

But we don't know how close by the upstream is, 0.5 *should* be enough on the 
net, but it is borderline; to do this correctly we may need to track rtt 
dynamically and adjust timeouts to that.

/Miek

--
Miek Gieben

miekg:
maybe it *is* the best way to do it in `connect`. Right now debugging TCP failures to 1.1.1.1 which surface because the remote side closed the connection

miekg:
And `forward.go` *does* look to be a better fit, see #1651
(if we're going to retry we should do it there as well, also allows checking it we've "used up" all the upstreams and redoing the udp query is worth it).

rdrozhdzh:
btw, is it correct that coredns returns SERVFAIL in case if queries to upstreams were timed out? Maybe coredns also should simulate timeout? i.e. return from ServeDNS() without writing response. So, we'll just pass to client exactly the same response which we get from upstream
My understanding is that if client doesn't get response during timeout it will likely retry. But I'm not sure about SERVFAIL. Will client normally retry on SERVFAIL?

miekg:
>btw, is it correct that coredns returns SERVFAIL in case if queries to 
>upstreams were timed out? 

yes, that's the response.

> Maybe coredns also should simulate timeout? i.e. 
>return from ServeDNS() without writing response. So, we'll just pass to client 
>exactly the same response which we get from upstream

Depends on client retry mechanism, but I rather have us send back SERVFAIL 
instead of letting the client timeout *again*.

>My understanding is that if client doesn't get response during timeout it will 
>likely retry. But I'm not sure about SERVFAIL. Will client normally retry on 
>SERVFAIL?

It should try the second resolver that is configured. This also makes the case 
that we should retry ourselves, we may play with the truncated bit in some 
cases? Allthough even then some stub resolver don't play ball.

/Miek

--
Miek Gieben

fantuz:
just for the sake of another parallel, **pdns** implements concepts of "_max-failures_" (64 times by default to declare an upstream down) and "_throttle_" (30 second throttle added after marking server down, to encourage client look somewhere else), along with a per-query "_wall-clock_" timeout before closing/dropping/giving up with client...

maybe 2 seconds retries and 5 in total is not enough ?

miekg:
I'm closing the PR - we can keep discussing of course, but maybe an issue will be better.

PR #1629, opened 2018-03-23
Fix #1625

### LINKED ISSUE #1625: High CPU load with ETCD backend
Hi!

We are using Coredns with Etcd as backend and we had an issue today that the Coredns caused full CPU load on three instances. The three Coredns are installed on three Etcd instances and the Etcd instances are not use by anything else.

After adding some entries to the Etcd the Coredns'es spinned up to 100% CPU load (multicore on-prem server). After deleting the last entries and restarting the Coredns the CPU keeps on 1%.

Unfortunately I cannot see any option to enable some kind of debug output. Is there a kind of a switch to enable debug output? I would like to see what is going on inside the Coredns to triage what is causing the problem.

Thanks!

m4r10k:

Fixes a problem in the request.go when the rl and size values are equal. If this is the case the loop cannot break and spins up the CPU to full load.

#1625

### DIFF
--- a/request/request.go
+++ b/request/request.go
@@ -217,6 +217,9 @@ func (r *Request) Scrub(reply *dns.Msg) (*dns.Msg, Result) {
 			re = m - 1
 			continue
 		}
+		if rl == size {
+			break
+		}
 	}
 	// We may come out of this loop with one rotation too many as we don't break on rl == size.
 	// I.e. m makes it too large, but m-1 works.
@@ -244,6 +247,9 @@ func (r *Request) Scrub(reply *dns.Msg) (*dns.Msg, Result) {
 		if rl > size {
 			ra = m - 1
 			continue
+		}
+		if rl == size {
+			break
 		}
 	}
 	// We may come out of this loop with one rotation too many as we don't break on rl == size.


miekg:
Thanks for the epic debugging, I'll call this out in the release notes.

request/request.go
```diff
@@ -217,6 +217,9 @@ func (r *Request) Scrub(reply *dns.Msg) (*dns.Msg, Result) {
 			re = m - 1
 			continue
 		}
+		if rl == size {
+			break
+		}
 	}
 	// We may come out of this loop with one rotation too many as we don't break on rl == size.
```
chrisohaver, line 224: Does this comment need to be updated?  or is it unrelated to the break on rl == size added above?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fix #1625 (#1..." ]
>
>
>
>> @@ -217,6 +217,9 @@ func (r *Request) Scrub(reply *dns.Msg) (*dns.Msg, Result) {
> 			re = m - 1
> 			continue
> 		}
>+		if rl == size {
>+			break
>+		}
> 	}
> 	// We may come out of this loop with one rotation too many as we don't break on rl == size.
>
>Does this comment need to be updated?  or is it unrelated to the break on rl == size added above?

... probably ...

that last bit is now wrong anyway.

PR #1630, opened 2018-03-23
Remove comments 1625

m4r10k:
Remove comments which are useless due to code changes.

### DIFF
--- a/request/request.go
+++ b/request/request.go
@@ -217,6 +217,9 @@ func (r *Request) Scrub(reply *dns.Msg) (*dns.Msg, Result) {
 			re = m - 1
 			continue
 		}
+		if rl == size {
+			break
+		}
 	}
 	// We may come out of this loop with one rotation too many as we don't break on rl == size.
 	// I.e. m makes it too large, but m-1 works.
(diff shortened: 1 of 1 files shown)


miekg:
Note 100% removing the comments will make it right; I still think we can leave the loop normally and have something slightly too large; I double check after the 1.1.1 release.

m4r10k:
OK, I understand +1

m4r10k:
You have more overview about what exactly  is going on there in which case!
But if you need something to test, let me know then!

request/request.go
```diff
@@ -217,6 +217,9 @@ func (r *Request) Scrub(reply *dns.Msg) (*dns.Msg, Result) {
 			re = m - 1
 			continue
 		}
+		if rl == size {
+			break
+		}
 	}
 	// We may come out of this loop with one rotation too many as we don't break on rl == size.
```
chrisohaver: This one too, if I'm not mistaken.

chrisohaver:
lgtm

miekg:
Full fix is in https://github.com/coredns/coredns/pull/1631
Tried to make a test case for the new `break`, but ended up doing more as we also need to take that OPT record into account (get re-added in SizeAndDo)

PR #1634, opened 2018-03-25
Fix dns-01-003

miekg:
Add checks in forward and proxy and a backstop in the *cache* plugin.


miekg:
Review of (very) similar patch has been done over email.

Issue #1636, opened 2018-03-26
bug: CH class handling is unblocked for first bind address only


UladzimirTrehubenka:
Tested on 5546dbf3c6dc7a5b437f0b41534052395e20f329 with Corefile
```
.:53 {
        chaos
	bind 127.0.0.1
	bind 10.120.252.202
	forward . 8.8.8.8:53
}
```
Observed right response for first bind address:
dig  @127.0.0.1 version.bind txt ch
```
; <<>> DiG 9.10.5-P2-RedHat-9.10.5-2.P2.fc24 <<>> @127.0.0.1 version.bind txt ch
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 39379
```
but for second bind address we get refuse:
dig @10.120.252.202 version.bind txt ch
```
; <<>> DiG 9.10.5-P2-RedHat-9.10.5-2.P2.fc24 <<>> @10.120.252.202 version.bind txt ch
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: REFUSED, id: 54886

```

PR #1637, opened 2018-03-26
Fixed NewServer() init logic

UladzimirTrehubenka:
Fixed NewServer() init logic
https://github.com/coredns/coredns/issues/1636
none

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -66,6 +66,17 @@ func NewServer(addr string, group []*Config) (*Server, error) {
 		// compile custom plugin for everything
 		if site.registry != nil {
 			// this config is already computed with the chain of plugin
+			// set classChaos in accordance with previously registered plugins
+			for name := range enableChaos {
+				if _, ok := site.registry[name]; ok {
+					s.classChaos = true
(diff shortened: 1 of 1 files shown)


core/dnsserver/server.go
```diff
@@ -64,16 +64,19 @@ func NewServer(addr string, group []*Config) (*Server, error) {
 		// set the config per zone
 		s.zones[site.Zone] = site
 		// compile custom plugin for everything
+		registered := false
 		if site.registry != nil {
```
miekg: can we just re use `site.registry`?

rdrozhdzh: @miekg, actually, there is no problem with reusing site.registry. The problem is with initializing fields Server.trace and Server.classChaos which should be done for each Server object, even if they are reusing the same config, e.g. see line https://github.com/coredns/coredns/pull/1637/files#diff-a277767ea17c7dfa10555939729f0d65R90

UladzimirTrehubenka: Reimplemented with using site.registry

miekg:
thanks!

PR #1639, opened 2018-03-26
plugin/forward: set the RD bit in the hc

miekg:
My routers acts funny when it sees it non RD query; make this HC as
boring as possible


budaki:
Let it forward to right one

PR #1646, opened 2018-03-31
plugin/proxy: return client error

miekg:
Return the client error if there was one instead of the generic
"no healthy upstream or error"

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -65,6 +65,7 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 
 	fails := 0
 	var span, child ot.Span
+	var upstreamErr error
 	span = ot.SpanFromContext(ctx)
 
 	for _, proxy := range f.list() {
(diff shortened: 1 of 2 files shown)


miekg:
testing this now with `tls://1.1.1.1`:
~~~
01/Apr/2018:14:21:43 +0100 [ERROR 0 miek.nl. MX] EOF
[::1]:54809 - [01/Apr/2018:14:21:47 +0100] 12042 "MX IN miek.nl. udp 37 false 4096" NOERROR qr,rd,ra,ad 469 79.518246ms
[::1]:39883 - [01/Apr/2018:14:21:49 +0100] 18508 "MX IN miek.nl. udp 37 false 4096" SERVFAIL qr,rd 37 160.654µs
01/Apr/2018:14:21:49 +0100 [ERROR 0 miek.nl. MX] EOF
~~~
Seems to work.

Also seems to indicate I need to catch this error in *forward*

rdrozhdzh:
the fix looks good, just change the PR subj to "plugin/forward ..."

PR #1651, opened 2018-04-01
plugin/forward: TCP conns can be closed

miekg:
Only when we read and get a io.EOF we know the conn is closed (for TCP).
If this is the case Dial (again) and retry. Note that this new
connection can also be closed by the upstream, we may want to add a
DialForceNew or something to get a new TCP connection..

Simular to #1624, *but* this is by (TCP) design. We also don't have to
wait for a timeout which makes it easier to reason about.

### DIFF
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -5,6 +5,7 @@
 package forward
 
 import (
+	"io"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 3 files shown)


miekg:
testing this locally:
~~~
default/coredns-deployment-bc444446d-pvnms[coredns]: 10.244.2.0:37262 - [01/Apr/2018:15:48:56 +0000] 45286 "AAAA IN google.com. udp 28 false 512" SERVFAIL qr,rd 28 586.385059ms
default/coredns-deployment-bc444446d-pvnms[coredns]: 01/Apr/2018:15:48:56 +0000 [ERROR 0 google.com. AAAA] EOF
~~~
it's getting EOF because the second conn is also been closed.

rdrozhdzh:
@miekg why do you do only one attempt to reconnect on tcp EOF? Actually you may have many cached connections which already closed by server, by still not expired on client side.
We may want to add some flag to connErr struct to distinguish if we got cached connection or newly created one. So, we could retry unless we get new connection.
Also, user may decrease 'expire' parameter of forward plugin if his configured upstream server drops connection too fast

miekg:
Yes, I would like to not give these out in the first place. Maybe a better
place is persistent.go, but you can only see if the conn. is closed from a
failed read.

rdrozhdzh:
@miekg I meant improvement like #1655

rdrozhdzh:
I built a version of my PR#1655 with a line
`log.Println("cached closed")`
after
`if err != nil && err == errCachedClosed {`
and tested in config
`dnsperf ---> coredns1 --- TLS ---> coredns2 ---> external DNS server`
After completing dnsperf test, I waited 9 sec, and issued DNS query with dig, and got about 100 lines with "cached closed". So, all these connections were dropped, then new connection was created and the query was resolved successfully

PR #1655, opened 2018-04-02
plugin/forward: retry on cached tcp connection closed by peer

rdrozhdzh:
Improve #1651


plugin/forward/forward.go
```diff
@@ -176,6 +173,7 @@ var (
 	errInvalidDomain = errors.New("invalid domain for forward")
 	errNoHealthy     = errors.New("no healthy proxies")
 	errNoForward     = errors.New("no forwarder defined")
+	errCachedClosed  = errors.New("cached connection was closed by proxy")
```
miekg: nit: s/proxy/peer ?

miekg (approved):

plugin/forward/forward.go
rdrozhdzh: done

PR #1657, opened 2018-04-02
plugin/forward: add query timeout

rdrozhdzh:
- subsequently retry sending query via different proxies
   until query timeout occurs. A proxy can be tried again if
   all other proxies were already tried

this fix is an alternative to #1624

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -8,6 +8,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"io"
+	"math/rand"
 	"time"
 
 	"github.com/coredns/coredns/plugin"
@@ -65,24 +66,36 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	}
 
 	fails := 0
-	var span, child ot.Span
-	var upstreamErr error
-	span = ot.SpanFromContext(ctx)
+	i := 0
+	returnErr := errNoHealthy
+	span := ot.SpanFromContext(ctx)
+	proxyList := f.list()
+	deadline := time.Now().Add(queryTimeout)
 
-	for _, proxy := range f.list() {
-		if proxy.Down(f.maxfails) {
-			fails++
-			if fails < len(f.proxies) {
-				continue
+	for time.Now().Before(deadline) {
+		if i >= len(proxyList) {
+			if i == fails {
+				// All upstream proxies are dead, assume healtcheck is completely broken and randomly
+				// select an upstream to connect to.
+				i = rand.Intn(len(proxyList))
+				deadline = time.Now() // give it a last attempt
+
+				HealthcheckBrokenCount.Add(1)
+			} else {
+				// reached the end of list, reset to begin
+				i = 0
+				fails = 0
 			}
-			// All upstream proxies are dead, assume healtcheck is completely broken and randomly
-			// select an upstream to connect to.
-			r := new(random)
-			proxy = r.List(f.proxies)[0]
-
-			HealthcheckBrokenCount.Add(1)
 		}
 
+		proxy := proxyList[i]
+		i++
+		if proxy.Down(f.maxfails) {
+			fails++
+			continue
+		}
+
+		var child ot.Span
 		if span != nil {
 			child = span.Tracer().StartSpan("connect", ot.ChildOf(span.Context()))
 			ctx = ot.ContextWithSpan(ctx, child)
@@ -107,18 +120,13 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 		}
 
 		ret, err = truncated(ret, err)
-		upstreamErr = err
-
 		if err != nil {
+			returnErr = err
 			// Kick off health check to see if *our* upstream is broken.
 			if f.maxfails != 0 {
 				proxy.Healthcheck()
 			}
-
-			if fails < len(f.proxies) {
-				continue
-			}
-			break
+			continue
 		}
 
 		// Check if the reply is correct; if not return FormErr.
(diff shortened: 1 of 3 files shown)


miekg:
yes, I think this is a saner way of doing things. I'll take a closer look later (on holiday atm)

plugin/forward/forward.go
```diff
@@ -65,24 +66,36 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	}
 
 	fails := 0
-	var span, child ot.Span
-	var upstreamErr error
-	span = ot.SpanFromContext(ctx)
+	i := 0
+	returnErr := errNoHealthy
```
miekg, line 70: I thought a bit about the name `upstreamErr` and I think it's a better name than `returnErr`

plugin/forward/forward.go
```diff
@@ -185,3 +189,5 @@ const (
 	randomPolicy policy = iota
 	roundRobinPolicy
 )
+
+const queryTimeout = 5 * time.Second
```
miekg, line 193: `defaultTimeout` ?

plugin/forward/forward.go
```diff
@@ -65,24 +66,36 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
...
+	span := ot.SpanFromContext(ctx)
+	proxyList := f.list()
+	deadline := time.Now().Add(queryTimeout)
+
+	for time.Now().Before(deadline) {
+		if i >= len(proxyList) {
+			if i == fails {
+				// All upstream proxies are dead, assume healtcheck is completely broken and randomly
+				// select an upstream to connect to.
+				i = rand.Intn(len(proxyList))
```
miekg, line 80: Just leave the original code here?

plugin/forward/forward.go
```diff
@@ -65,24 +66,36 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
...
+				deadline = time.Now() // give it a last attempt
+
+				HealthcheckBrokenCount.Add(1)
```
miekg, line 83: add a `continue` here, so we can drop the else and dedent the body of the else.

plugin/forward/forward.go
```diff
@@ -65,24 +66,36 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	}
 
 	fails := 0
-	var span, child ot.Span
-	var upstreamErr error
-	span = ot.SpanFromContext(ctx)
+	i := 0
+	returnErr := errNoHealthy
+	span := ot.SpanFromContext(ctx)
```
miekg, line 71: why did you move the `child` to down below?

plugin/forward/forward.go
```diff
@@ -65,24 +66,36 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
...
+	proxyList := f.list()
```
miekg, line 72: just use `list`

plugin/forward/health_test.go
```diff
@@ -45,15 +45,21 @@ func TestHealth(t *testing.T) {
...
 	const expected = 1
 	i := uint32(0)
+	q := uint32(0)
 	s := dnstest.NewServer(func(w dns.ResponseWriter, r *dns.Msg) {
 		if r.Question[0].Name == "." {
 			// health check, answer
 			atomic.AddUint32(&i, 1)
 			ret := new(dns.Msg)
 			ret.SetReply(r)
 			w.WriteMsg(ret)
```
miekg, line 55: just `return` here

plugin/pkg/up/up.go
```diff
@@ -37,9 +38,11 @@ func (p *Probe) Stop() { p.stop <- true }
 func (p *Probe) Start(interval time.Duration) { go p.start(interval) }
 
 func (p *Probe) start(interval time.Duration) {
+	stop := uint32(0)
 	for {
 		select {
 		case <-p.stop:
+			atomic.StoreUint32(&stop, 1)
```
miekg, line 45: this is purely for testing?

miekg (changes requested):
Generally OK with this change; bunch of comments, some nits, some not.

plugin/forward/forward.go
rdrozhdzh: My intention was to use the same var for upstream error and no healthy error, and get rid of the extra 'if' statement at the end of ServeDNS(). Therefore, I renamed it

plugin/forward/forward.go
rdrozhdzh: I can change to defaultTimeout if you wish. But this name tell us less about purpose of the constant

plugin/forward/forward.go
rdrozhdzh: the original code was more expensive from performance point of view, so I changed it

plugin/forward/forward.go
rdrozhdzh: Well, actually here we detected that all proxies are down, and we need to do the last attempt to connect to random proxy. This logic was in original code. So, I calculate index of random proxy in line 80, specify last attempt by resetting `deadline`, and go to the main part of loop where proxy is taken from list by index and `Proxy.connect()` is called.
Adding `continue` here will skip connecting to proxy

plugin/forward/forward.go
rdrozhdzh: 1. To reduce the scope of `child`. It is not referred outside the loop
2. This way `child` is initialized with `nil` in each iteration. Theoretically `child` may not be assigned with value in current iteration, and later down the code we compare it with `nil` to call `Finish()`. Non-nil value of `child` from previous iteration could lead to repeated call to `Finish()` which seems incorrect. Though, I guess such case is unlikely on practice

plugin/pkg/up/up.go
rdrozhdzh: Actually not. The test run successfully without this change. The goroutine below is stopped at the end of test and doesn't affect results
This is rather a fix of probe logic. If we call `Probe.Stop()` we expect it will stop probing. But before my fix in case if probing function always return error it keeps probing forever, which seems incorrect

rdrozhdzh:
Thank you @miekg for review. It seems I did too much refactoring in this PR :) I can revert the changes not related directly to the topic of this PR. Just let me know if you don't want some changes to be merged here

miekg:
Quick reply: yes minimizing the changes helps to get a pr approved. If you
want subsequent changes just submit further PRs

rdrozhdzh:
The new PR#1665 created as a replacement for this one, so closing this

PR #1659, opened 2018-04-03
plugin/forward: use dynamic read timeout

rdrozhdzh:
- each proxy stores average RTT (round trip time) of last rttCount queries.
   For now, I assigned the value 4 to rttCount
 - the read timeout is calculated as doubled average RTT, but it cannot
   exceed default timeout
 - initial avg RTT is set to a half of default timeout, so initial timeout
   is equal to default timeout
 - the RTT for failed read is considered equal to default timeout, so any
   failed read will lead to increasing average RTT (up to default timeout)
 - dynamic timeouts will let us react faster on lost UDP packets
 - in future, we may develop a low-latency forward policy based on
   collected RTT values of proxies

### DIFF
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -6,6 +6,7 @@ package forward
 
 import (
 	"strconv"
+	"sync/atomic"
 	"time"
 
 	"github.com/coredns/coredns/request"
@@ -13,6 +14,19 @@ import (
 	"github.com/miekg/dns"
 	"golang.org/x/net/context"
 )
+
+func (p *Proxy) readTimeout() time.Duration {
+	rtt := time.Duration(atomic.LoadInt64(&p.avgRtt))
+	if rtt < timeout/2 {
+		return 2 * rtt
+	}
+	return timeout
+}
+
+func (p *Proxy) updateRtt(newRtt time.Duration) {
+	rtt := time.Duration(atomic.LoadInt64(&p.avgRtt))
+	atomic.AddInt64(&p.avgRtt, int64((newRtt-rtt)/rttCount))
+}
 
 func (p *Proxy) connect(ctx context.Context, state request.Request, forceTCP, metric bool) (*dns.Msg, error) {
 	start := time.Now()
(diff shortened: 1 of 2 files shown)


miekg (approved):

plugin/forward/connect.go
```diff
@@ -14,6 +15,19 @@ import (
...
+	rtt := time.Duration(atomic.LoadInt64(&p.avgRtt))
+	if rtt < timeout/2 {
+		return 2 * rtt
+	}
+	return timeout
+}
+
+func (p *Proxy) updateRtt(newRtt time.Duration) {
+	rtt := time.Duration(atomic.LoadInt64(&p.avgRtt))
+	atomic.AddInt64(&p.avgRtt, int64((newRtt-rtt)/rttCount))
```
johnbelamaric, line 28: The load is atomic and the add is atomic but the two together are not. So you could for example:

thread 1: load
thread 2: load
thread 2: add
thread 1: add

do we care?

rdrozhdzh: Yes, I know that combination of these two lines is not atomic. But we don't need precise value here, and to my mind the result looks acceptable.
Let me know if you prefer more safe and precise solution here (e.g. based on mutex)

miekg:
>
>
>
>> @@ -14,6 +15,19 @@ import (
> 	"golang.org/x/net/context"
> )
>
>+func (p *Proxy) readTimeout() time.Duration {
>+	rtt := time.Duration(atomic.LoadInt64(&p.avgRtt))
>+	if rtt < timeout/2 {
>+		return 2 * rtt
>+	}
>+	return timeout
>+}
>+
>+func (p *Proxy) updateRtt(newRtt time.Duration) {
>+	rtt := time.Duration(atomic.LoadInt64(&p.avgRtt))
>+	atomic.AddInt64(&p.avgRtt, int64((newRtt-rtt)/rttCount))
>
>Yes, I know that combination of these two lines is not atomic. But we don't need precise value here, and to my mind the result looks acceptable.
>Let me know if you prefer more safe and precise solution here (e.g. based on mutex)

Main concern not using a mutex was speed? I think using atomic is OK here.

rdrozhdzh:
@miekg , yes performance was the reason

Issue #1662, opened 2018-04-05
plugin/log - ability to log denial AND error DNS responses


fturib:
For monitoring purpose we would need the ability to log ont only one class, but several.
typically : Denial + Error

stanza would looks like:

```
. {
    log example.org {
        class denial error
    }
}
```

in another word, "class" parameter would be an array instead of a single value.

**PR is almost ready - WIP**

mparamonau:
PR is [here](https://github.com/coredns/coredns/pull/1664)

miekg:
closed via #1664

PR #1664, opened 2018-04-06
plugin/log: allow various combinations of classes of responses

mparamonau:
This allows to log responses of different classes, for example, denial and error.

I would propose to add an ability of the log plugin to mix classes of responses.
The goal is not to clutter up the log file with unnecessary messages.
The proposed changes in the log plugin could be useful, at least, in following cases:

1. we would want to log all queries which were not resolved successfully:
log . {
    class denial error
}

2. we would want to log all queries on which we did not get errors:
log . {
    class denial success
}

So, the classes "error, denial, success, all" could be mixed. Of course, the following combinations don't make sense (but allowed): "class denial error success" or even "class denial error success all" - it's just "class all" as well as we mix the class "all" with any other types - "class error all" and so on, the class "all" means that all messages will be logged whatever we mix together with "all".

Fixing [#1662](https://github.com/coredns/coredns/issues/1662)

Maybe it requires to fix the [documentation](https://coredns.io/plugins/log/) of the coredns log plugin.

### DIFF
--- a/plugin/log/log.go
+++ b/plugin/log/log.go
@@ -54,7 +54,9 @@ func (l Logger) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 
 		tpe, _ := response.Typify(rrw.Msg, time.Now().UTC())
 		class := response.Classify(tpe)
-		if rule.Class == response.All || rule.Class == class {
+		// If we don't set up a class in config, the default "all" will be added
+		// and we shouldn't have an empty rule.Class.
+		if rule.Class[response.All] || rule.Class[class] {
 			rep := replacer.New(r, rrw, CommonLogEmptyValue)
 			rule.Log.Println(rep.Replace(rule.Format))
 		}
@@ -71,7 +73,7 @@ func (l Logger) Name() string { return "log" }
 // Rule configures the logging plugin.
 type Rule struct {
 	NameScope string
-	Class     response.Class
+	Class     map[response.Class]bool
 	Format    string
 	Log       *log.Logger
 }
--- a/plugin/log/setup.go
+++ b/plugin/log/setup.go
@@ -52,11 +52,13 @@ func logParse(c *caddy.Controller) ([]Rule, error) {
 			rules = append(rules, Rule{
 				NameScope: ".",
 				Format:    DefaultLogFormat,
+				Class:     make(map[response.Class]bool),
 			})
 		} else if len(args) == 1 {
 			rules = append(rules, Rule{
 				NameScope: dns.Fqdn(args[0]),
 				Format:    DefaultLogFormat,
+				Class:     make(map[response.Class]bool),
 			})
 		} else {
 			// Name scope, and maybe a format specified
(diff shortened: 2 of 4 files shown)


fturib:
@mparamonau : there is extra check of style coding in the Makefile.
=> can you rune "make" in your local repository, and then fix the style issues ?

mparamonau:
sure, will do it

rdrozhdzh:
> we would want to log all queries which were not resolved successfully:

maybe "errors" plugin would be suitable for this

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/log: a..." ]
>> we would want to log all queries which were not resolved successfully:
>
>maybe "errors" plugin would be suitable for this

I don't think *errors* will see these as such.

plugin/log/log.go
```diff
@@ -54,7 +54,7 @@ func (l Logger) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 
 		tpe, _ := response.Typify(rrw.Msg, time.Now().UTC())
 		class := response.Classify(tpe)
-		if rule.Class == response.All || rule.Class == class {
+		if rule.Class[response.All] || rule.Class[class] {
```
miekg: Is `rule.Class[class]` always in the map? If not this crashes runtime

plugin/log/setup.go
```diff
@@ -74,28 +76,36 @@ func logParse(c *caddy.Controller) ([]Rule, error) {
...
 			})
 		}
 
 		// Class refinements in an extra block.
 		for c.NextBlock() {
 			switch c.Val() {
-			// class followed by all, denial, error or success.
+			// class followed by combinations of all, denial, error and success.
 			case "class":
+				if len(rules[len(rules)-1].Class) != 0 {
```
miekg: what does this actually do? And why doesn't it return an error?

plugin/log/setup_test.go
```diff
@@ -17,56 +18,64 @@ func TestLogParse(t *testing.T) {
 		{`log`, false, []Rule{{
 			NameScope: ".",
 			Format:    DefaultLogFormat,
+			Class:     map[response.Class]bool{response.All: true},
```
miekg, line 21: Can you just add a new test that checks the class stuff?

miekg (changes requested):

plugin/log/log.go
mparamonau: Yes, it is, even if we don't set up a class in config, the default "all" will be added, hence, we shouldn't have an empty rule.Class

plugin/log/setup.go
mparamonau: Potentially we could have several "class" lines - here we handle only first record (previously we applied the latest)

plugin/log/setup_test.go
mparamonau: ok

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/log: a..." ]
>mparamonau commented on this pull request.
>
>> @@ -54,7 +54,7 @@ func (l Logger) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
>
> 		tpe, _ := response.Typify(rrw.Msg, time.Now().UTC())
> 		class := response.Classify(tpe)
>-		if rule.Class == response.All || rule.Class == class {
>+		if rule.Class[response.All] || rule.Class[class] {
>
>Yes, it is, even if we don't set up a class in config, the default "all" will be added, hence, we shouldn't have an empty rule.Class

Can you put this as a comment above this line?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/log: a..." ]
>mparamonau commented on this pull request.
>
>
>
>>  			case "class":
>+				if len(rules[len(rules)-1].Class) != 0 {
>
>Potentially we could have several "class" lines - here we handle only first record (previously we applied the latest)

Woulnd't it be nicer/better to allow this, i.e. multiple statements that get 
OR-ed?

class denial
class all

mparamonau:
Woulnd't it be nicer/better to allow this, i.e. multiple statements that get
OR-ed?

class denial
class all

Initially I did it in this way but then decided that writing every time the same word "class" is redundant).

mparamonau:
Now it allows us to use multiple statements that get OR-ed.

fturib:
@miekg : 
- the UT are updated. 
- the choice of several values rather than several lines for class makes sense.

Can we validate this PR ?

**Idea for another extension of logs, on another PR**
NOTE: about the OR'ed operation, I though we could do **on another PR** : have an execution of all LOG stanza that match the domain:

```
Log  .  DENIED {name} {when}  {
class denial
}
Log  . ERROR {name} {proto}  {
class error
}
Log  google.com  ERROR Google {name} {proto}  {
class error
}
```
This one would provide differentiated LOG information depending the class, but also the domain matched.

NOTE: we would have an issue maybe on the "best match .." but there is a function for that, if several logs are matching per criteria (domain, class)

miekg (approved):

miekg:
Oh, can you send a new PR with the changes to the README.md for this plugin. Thanks!

mparamonau:
Sure

PR #1665, opened 2018-04-09
plugin/forward: add query timeout

rdrozhdzh:
subsequently retry sending query via different proxies
until query timeout occurs. A proxy can be tried again if
all other proxies were already tried


rdrozhdzh:
this is a cleaner version of #1657

Issue #1669, opened 2018-04-11
plugin/forward: can't handle truncated responses


grobie:
Tested against current master a0f294e550e9369bab711d9ae4ec79c942eeb792.

Any truncated message will be responded with a FormErr by plugin/forward, due to a state mismatch https://github.com/coredns/coredns/blob/a0f294e550e9369bab711d9ae4ec79c942eeb792/plugin/forward/forward.go#L133-L137

That's the result of the proxy implementation always returning `nil` if miekg/dns returns an error, even it's just the pseudo `ErrTruncated`. That affects the implementation in `truncated()` and later on in `Match()`. https://github.com/coredns/coredns/blob/a0f294e550e9369bab711d9ae4ec79c942eeb792/plugin/forward/connect.go#L69

It'd be great to see more test coverage in the plugin/forward implementation given its growing complexity.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin/forward: c..." ]
>It'd be great to see more test coverage in the plugin/forward implementation given its growing complexity.

Agreed. Things never stay simple :/

grobie:
In v1.1.1 plugin/forward will retry and eventually respond with ServFail to truncated messages.

Issue #1667, opened 2018-04-11
plugin/forward: panic in truncation handling


grobie:
Yay, truncation again :-) The miekg/dns library assume a non-`nil` message passed to `SetReply(*dns.Msg, err error)` and will panic otherwise. https://github.com/miekg/dns/blob/01d59357d468872339068bcd5d55a00e2463051f/defaults.go#L14-L17

In `plugin/forward/truncated.go` coredns explicitly calls it with a  nil message. https://github.com/coredns/coredns/blob/a0f294e550e9369bab711d9ae4ec79c942eeb792/plugin/forward/truncated.go#L17-L19

grobie:
Fixed by #1670.

PR #1670, opened 2018-04-11
plugin/forward: check TC correctly on reply.

miekg:
Add test for this.

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -103,7 +103,7 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 			child.Finish()
 		}
 
-		ret, err = truncated(ret, err)
+		ret, err = truncated(state, ret, err)
 		upstreamErr = err
 
 		if err != nil {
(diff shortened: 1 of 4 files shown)


miekg:
I'll work on adding another test that exercises the retry bit on TC and see what falls out of that.

miekg:
added Forward test as well in here.

grobie:
Thanks for working on this and adding a test!

I can confirm this fixes the panic reported in #1667. I have to admit I'd wish for a simpler implementation in general. A truncated response from an upstream server is first parsed, an internal error code gets returned with it, then the parsed response is thrown away, a new message is created based on the original request, and finally that message is compared against the original request, even though they will always be the same. I can see that you don't want to change miekg/dns anymore to not return `ErrTruncated`, but that seems to be the source of all the complication here.

It'd be great to extend the test to verify that a truncated response from an upstream server is actually returned by plugin/forward. At the moment, it's still thrown away and a custom message is returned due to the error handling in connect.go reported in #1669.

miekg:
>Thanks for working on this and adding a test!
>
>I can confirm this fixes the panic reported in #1667. I have to admit I'd wish for a simpler implementation in general. A truncated response from an upstream server is first parsed, an internal error code gets returned with it, then the parsed response is thrown away, a new message is created based on the original request, and finally that message is compared against the original request, even though they will always be the same. I can see that you don't want to change miekg/dns anymore to not return `ErrTruncated`, but that seems to be the source of all the complication here.
>
>It'd be great to extend the test to verify that a truncated response from an upstream server is actually returned by plugin/forward. At the moment, it's still thrown away and a custom message is returned due to the error handling in connect.go reported in #1669.

Good points. I'll follow up with these.

PR #1671, opened 2018-04-11
plugin/pkg/up: implement backoff

miekg:
The backoff is not exponential on every 3rd error increase the duration
with 'interval'.

Also log the errors we see from failed healthchecks.

### DIFF
--- a/plugin/pkg/up/up.go
+++ b/plugin/pkg/up/up.go
@@ -1,5 +1,6 @@
 // Package up is used to run a function for some duration. If a new function is added while a previous run is
-// still ongoing, nothing new will be executed.
+// still ongoing, nothing new will be executed. Every 3rd failure we increase the interval with the interval
+// duration (up to four times).
 package up
 
 import (
@@ -53,12 +54,21 @@ func (p *Probe) start(interval time.Duration) {
 			// Passed the lock. Now run f for as long it returns false. If a true is returned
 			// we return from the goroutine and we can accept another Func to run.
 			go func() {
+				i, j := 1, 1
 				for {
 					if err := f(); err == nil {
 						break
 					}
-					// TODO(miek): little bit of exponential backoff here?
-					time.Sleep(interval)
+					time.Sleep(time.Duration(i) * interval)
+					// Every third failure, we increase the interval, but only up to 4 * interval.
+					j++
+					if j%3 == 0 {
+						i++
+						if i > 4 {
+							i = 4
+						}
+					}
+
 				}
 				p.Lock()
 				p.inprogress = false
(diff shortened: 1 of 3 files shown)


miekg:
Also up the timeout to something more sensible then 1 sec. up to 4 sec.

plugin/pkg/up/up.go
```diff
@@ -53,12 +54,21 @@ func (p *Probe) start(interval time.Duration) {
...
+					time.Sleep(time.Duration(i) * interval)
+					// Every third failure, we increase the interval, but only up to 4 * interval.
+					j++
+					if j%3 == 0 {
+						i++
+						if i > 4 {
+							i = 4
+						}
+					}
+
```
fturib, line 71: If I look only to the diff, then the implementation looks ok.

But, I am wondering what happen when the f() always return an error (because the remote endpoint is down), and you Stop or Reload the plugin. => this "go func()" will run forever and continue to run f() on the same remote endpoint no ?


=> you may end-up with weird behavior with obsolete code still called in background. Especially in the the RELOAD use case. 

Am I right ?

fturib: Should we add a mechanism to get out of this"go func()" when the Proxy is deleted ?

miekg: This is stopped when this is called: https://github.com/coredns/coredns/blob/master/plugin/forward/proxy.go#L82

fturib: IMO, the proxy.go#L82 will stop the method "start" .. but it will not stop the launched go routine that have a possible endless loop. This one will continue to loop until f() return does not return an error.

NOTE: I understand this part is not the point of this review. I just wanted to highlight it.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/pkg/up..." ]
>
>
>
>>  				for {
> 					if err := f(); err == nil {
> 						break
> 					}
>-					// TODO(miek): little bit of exponential backoff here?
>-					time.Sleep(interval)
>+					time.Sleep(time.Duration(i) * interval)
>+					// Every third failure, we increase the interval, but only up to 4 * interval.
>+					j++
>+					if j%3 == 0 {
>+						i++
>+						if i > 4 {
>+							i = 4
>+						}
>+					}
>+
>
>IMO, the proxy.go#L82 will stop the method "start" .. but it will not stop the launched go routine that have a possible endless loop. This one will continue to loop until f() return does not return an error.
>
>NOTE: I understand this part is not the point of this review. I just wanted to highlight it.

No, good to highlight this. Let me double check; either in this PR or otherwise 
in a follow up.

Thanks!

miekg:
I might drop this one entirely  and just fix the thing your mentioned.

PR #1676, opened 2018-04-12
pkg/up: stop *all* goroutines

miekg:
Stop all goroutines after we get the stop signal.


fturib (approved):
I am ok with the logic.

Issue #1672, opened 2018-04-12
plugin/forward: no TLSConfig does not mean no TLS


miekg:
The healthchecks are doing tcp-tls *if* there is a tlsConfig specified, but by default you just want to use the system's certs meaning you don't have the `tls` directive in the config. So the healthchecks then default to `udp` on port 853.

In my k8s cluster I see a lot of:
```
default/coredns-deployment-66989b445f-v58zt[coredns]: 2018/04/12 06:20:40 [WARNING] Health check of "9.9.9.9:853" failed with: read udp 10.244.1.105:44135->9.9.9.9:853: read: no route to host
```
Which now all make sense.

It also doesn't account for forceTCP, as we want these checks to be in-band it should use tcp for the healthchecks as well.

miekg:
The nice thing is the fallback the assumes "all healthchecks are broken, lets just forward the request" works. I.e. my DNS resolution is working well :+1:

miekg:
#1714 is new PR

miekg:
merged da fix.

PR #1679, opened 2018-04-13
Fix TLS configuration for forward.

miekg:


miekg:
Not sure if related, but:
~~~
default/coredns-deployment-664ddc7d9-htl22[coredns]: 14/Apr/2018:13:51:48 +0000 [PANIC registry-1.docker.io. AAAA] /home/miek/upstream/go/src/sync/atomic/asm_linux_arm.s:109 - runtime error: invalid memory address or nil pointer dereference
~~~

miekg:
anyone else available for review as well? /cc @rdrozhdzh @grobie ?

plugin/forward/connect.go
```diff
@@ -33,11 +33,15 @@ func (p *Proxy) connect(ctx context.Context, state request.Request, forceTCP, me
 	start := time.Now()
 
 	proto := state.Proto()
-	if forceTCP {
+	if p.tlsConfig != nil {
+		proto = "tcp-tls"
+	}
+
+	if forceTCP && proto == "udp" {
```
rdrozhdzh, line 40: this overwrites "tcp-tls" if it was set 3 lines above. Consider `else if`

plugin/forward/setup.go
```diff
@@ -136,14 +140,25 @@ func parseForward(c *caddy.Controller) (*Forward, error) {
...
+	// Update the proxies with possibly new TLS Configs.
 	for i := range f.proxies {
-		// Only set this for proxies that need it.
-		if protocols[i] == TLS {
-			f.proxies[i].SetTLSConfig(f.tlsConfig)
+		pr := f.proxies[i]
+		pr.setExpire(f.expire)
+
+		if f.tlsConfig != nil { // forward changes
+			pr.setTLSConfig(f.tlsConfig)
```
rdrozhdzh, line 153: this will turn non-tls upstreams to tls, since you removed `if protocols[i] == TLS`

rdrozhdzh (changes requested):
What was the reason of moving tlsConfig from transport to Proxy? Now we have to pass it in each `dial()`. IMO the transport related changes were not necessary.

rdrozhdzh:
I think the basic fix for #1672 could be like adding
```
if f.tlsConfig == nil {
	f.tlsConfig = new(tls.Config)
}

```
before 
https://github.com/coredns/coredns/blob/9a7e487a5aac4643baf6b385cc2d8c444e648221/plugin/forward/setup.go#L138-L140

plus the change you made in Proxy.setTLSConfig() to update tlsClient

miekg:
Thanks for the review @rdrozhdzh . You raise a good point with moving the tls from transport, that should have been necessary. Let me see if I can put that back easily.

miekg:
Closing in favor of https://github.com/coredns/coredns/pull/1714

Thanks @rdrozhdzh, really helpful review - closing this PR.

PR #1680, opened 2018-04-14
plugin/forward: min and max for avgRTT

### LINKED ISSUE #1678: plugin/forward: make read timeout configurable
The recently added dynamic timeout feature (#1659) may give worse results than fixed timeout under some circumstances, e.g. at high load when upstream servers are overloaded. Dynamic timeout allows us re-sending unanswered queries faster, and this could be a problem, because we make upstream server load even higher.

In my tests I saw RTT (round trip time) as minimal as 10μs. For comparison, transmission of 50 bytes (400 bits) over 100Mbit network takes about 4μs. So, we can easily catch timeout at the moment when message was partially read which results in truncated messages 

Therefore, I propose to introduce minimal timeout. The reason why we added dynamic timeout feature is to have more attempts to re-send query (probably to different upstreams) during limited period (5 seconds).  Even if we set minimal timeout to 200ms we will get around 25 attempts which is quite enough. No sense to retry faster and create additional server load.

I propose to add a timeout option to forward plugin like below
```
forward .:53 {
    timeout MAX MIN RATIO WEIGHT
}

```
If user specifies only the first parameter (**MAX**) then dynamic timeout will not be used

**MAX** - maximum read timeout which is currently defined by constant `timeout ` 
https://github.com/coredns/coredns/blob/e671e22e657b2fcb1580abd311e9c340b20b9e96/plugin/forward/proxy.go#L93

**MIN** - minimum read timeout. Timeout is based on average RTT but cannot be less than MIN
**RATIO** - desired timeout/RTT ratio. I.e. timeout will be calculated as (average  RTT) * RATIO
**WEIGHT** - defines how much average RTT is influenced by current RTT, i.e. avgRTT = avgRTT + (newRTT-avgRTT)/WEIGHT like in function below
https://github.com/coredns/coredns/blob/e671e22e657b2fcb1580abd311e9c340b20b9e96/plugin/forward/connect.go#L27-L30

miekg:
make this bounded.

Fixes #1678

### DIFF
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -19,17 +19,13 @@ import (
 func (p *Proxy) readTimeout() time.Duration {
 	rtt := time.Duration(atomic.LoadInt64(&p.avgRtt))
 
-	if rtt < timeout/2 {
-		return 2 * rtt
-	}
 	if rtt < minTimeout {
 		return minTimeout
 	}
-	if rtt > maxTimeout {
-		return maxTimeout
+	if rtt < timeout/2 {
+		return 2 * rtt
 	}
-
-	return timeout
+	return maxTimeout
 }
 
 func (p *Proxy) updateRtt(newRtt time.Duration) {
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -18,9 +18,17 @@ import (
 
 func (p *Proxy) readTimeout() time.Duration {
 	rtt := time.Duration(atomic.LoadInt64(&p.avgRtt))
+
 	if rtt < timeout/2 {
 		return 2 * rtt
 	}
+	if rtt < minTimeout {
+		avg = minTimeout
+	}
+	if rtt > maxTimeout {
+		avg = maxTimeout
+	}
+
 	return timeout
 }
 
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -22,7 +22,7 @@ func (p *Proxy) readTimeout() time.Duration {
 	if rtt < minTimeout {
 		return minTimeout
 	}
-	if rtt < timeout/2 {
+	if rtt < maxTimeout/2 {
 		return 2 * rtt
 	}
 	return maxTimeout
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -23,10 +23,10 @@ func (p *Proxy) readTimeout() time.Duration {
 		return 2 * rtt
 	}
 	if rtt < minTimeout {
-		avg = minTimeout
+		return minTimeout
 	}
 	if rtt > maxTimeout {
-		avg = maxTimeout
+		return maxTimeout
 	}
 
 	return timeout
--- a/plugin/forward/proxy.go
+++ b/plugin/forward/proxy.go
@@ -92,6 +92,6 @@ const (
 	dialTimeout = 4 * time.Second
 	timeout     = 2 * time.Second
 	maxTimeout  = 2 * time.Second
-	minTimeout  = 500 * time.Millisecond
+	minTimeout  = 10 * time.Millisecond
 	hcDuration  = 500 * time.Millisecond
 )
--- a/plugin/forward/proxy.go
+++ b/plugin/forward/proxy.go
@@ -91,5 +91,7 @@ func (p *Proxy) start(duration time.Duration) { p.probe.Start(duration) }
 const (
 	dialTimeout = 4 * time.Second
 	timeout     = 2 * time.Second
+	maxTimeout  = 2 * time.Second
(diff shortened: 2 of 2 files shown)


plugin/forward/connect.go
```diff
@@ -87,4 +94,8 @@ func (p *Proxy) connect(ctx context.Context, state request.Request, forceTCP, me
 	return ret, nil
 }
 
-const rttCount = 4
+const (
+	rttCount   = 4
+	minTimeout = time.Millisecond * 500
+	maxTimeout = dialTimeout // 4 s
```
rdrozhdzh: I think 4s is too much. Such timeout will eat almost all our queryTimeout time during which we can retry the query. I would suggest 
`maxTimeout = timeout` as here https://github.com/coredns/coredns/blob/d8b9dec4b35932b35b44b17a4b68883fda7c30d0/plugin/forward/proxy.go#L93

plugin/forward/connect.go
```diff
@@ -26,7 +26,14 @@ func (p *Proxy) readTimeout() time.Duration {
 
 func (p *Proxy) updateRtt(newRtt time.Duration) {
 	rtt := time.Duration(atomic.LoadInt64(&p.avgRtt))
-	atomic.AddInt64(&p.avgRtt, int64((newRtt-rtt)/rttCount))
+	avg := rtt + ((newRtt - rtt) / rttCount)
+	if avg < minTimeout {
```
rdrozhdzh: consider moving `if` logic to `readTimeout()` method. It's desirable to execute the two atomic instructions as closely as possible

miekg: done

plugin/forward/connect.go
```diff
@@ -18,9 +18,17 @@ import (
 
 func (p *Proxy) readTimeout() time.Duration {
 	rtt := time.Duration(atomic.LoadInt64(&p.avgRtt))
+
 	if rtt < timeout/2 {
 		return 2 * rtt
 	}
+	if rtt < minTimeout {
```
rdrozhdzh: the previous condition is less strict, so this condition will never trigger. Let's make it consistent, i.e
```
	if rtt < minTimeout {
		return minTimeout
	}
	if rtt < maxTimeout/2 {
		return 2 * rtt
	}
	return maxTimeout

```

miekg:
>
>
>
>>  	if rtt < timeout/2 {
> 		return 2 * rtt
> 	}
>+	if rtt < minTimeout {
>
>the previous condition is less strict, so this condition will never trigger. Let's make it consistent, i.e
>```
>	if rtt < minTimeout {
>		return minTimeout
>	}
>	if rtt < maxTimeout/2 {
>		return 2 * rtt
>	}
>	return maxTimeout
>
>```

done

grobie:
I really like the timeout behavior in plugin/forward, it performs very well in my tests. For our internal usage, 500ms is way to high though as lower bound. Our 99th percentile response time is <60ms, our median round trip time is 1ms.

In #1678 you write:

> I think we should do this, but not export it as something you can tweak. We should dynamically do the right thing without bothering the user with config tweaks IMO

I'd find it sad having to maintain our own fork of the forward plugin just to adjust all the timeouts to match our internal / no-public internet deployment. Would you reconsider your stance here?

grobie:
Exporting these constants so that I can change them at compile time would be of great help already.

miekg:
>I really like the timeout behavior in plugin/forward, it performs very well in my tests. For our internal usage, 500ms is way to high though as lower bound. Our 99th percentile response time is <60ms, our median round trip time is 1ms.

This can be changed, the 500ms is entirely made up...

>In #1678 you write:
>
>> I think we should do this, but not export it as something you can tweak. We should dynamically do the right thing without bothering the user with config tweaks IMO
>
>I'd find it sad having to maintain our own fork of the forward plugin just to adjust all the timeouts to match our internal / no-public internet deployment. Would you reconsider your stance here?

to have configuration knobs?

plugin/forward/connect.go
```diff
@@ -18,10 +18,14 @@ import (
 
 func (p *Proxy) readTimeout() time.Duration {
 	rtt := time.Duration(atomic.LoadInt64(&p.avgRtt))
+
+	if rtt < minTimeout {
+		return minTimeout
+	}
 	if rtt < timeout/2 {
```
rdrozhdzh: s/timeout/maxTimeout

miekg: done

grobie (approved):
Looks great. If further tweaks are desired from our side, I'll open a new issue.

miekg:
Feel free to merge (or tweak some more). I'm AFK for a couple of hours

rdrozhdzh:
LGTM

PR #1682, opened 2018-04-15
plugin/metrics: add 'server' label

miekg:
This uses the new WithServer(ctx) to get the current server from the
context.

First in a larger refactor to make all plugins do this.

Update the corresponding docs as well.


achille-roussel (commented):
That’s great, thanks for keeping me in the loop :+1:

PR #1683, opened 2018-04-15
plugin/proxy: add per server metrics

miekg:
Adds per server metrics.


miekg:
needs to be done in one go. closing

Issue #1686, opened 2018-04-17
plugin/health: allow per server health


fturib:
## Description

When trying to run CoreDNS with this Corefile 👍 

```
.:5002 francois.com:5002 {
    errors
    log
    health localhost:8080
    prometheus localhost:8075
	proxy . 8.8.8.8 {
		protocol dns force_tcp
	}
	reload 2s
}
```

I get the error:
`2018/04/17 17:11:54 listen tcp 127.0.0.1:8080: bind: address already in use`

## Invstigation

the reason is that the Setup for any plugin is called once per each head zone/port (listener). In current use case these are 2 : ".:5002" and "francois.com:5002".

The Plugin create one HEALTH structure for each call, and then will try to open one HTTP Listener for each HEALTH structure.

## Resolution in progress ...

miekg:
yes, health needs to be the same thing as the prometheus plugin.
This is currently non-obvious but mentioned in the README;
~~~
If you have multiple Server Block and need to export health for each of the plugins, you must run health endpoints on different ports:
~~~

miekg:
Coming back to this, while trying to fix it, your corefile is similar to:
~~~
.:5002 {
    errors
    log
    health localhost:8080
    prometheus localhost:8075
	proxy . 8.8.8.8 {
		protocol dns force_tcp
	}
	reload 2s
}

francois.com:5002 {
    errors
    log
    health localhost:8080
    prometheus localhost:8075
	proxy . 8.8.8.8 {
		protocol dns force_tcp
	}
	reload 2s
}
~~~

This is explicitly called out in the README:
> If you have multiple Server Blocks and need to export health for each of the plugins, you must run
> health endpoints on different ports:
> 
> ~~~ corefile
>com {
>   whoami
>   health :8080
>}
>
> net {
>    erratic
>    health :8081
>}
>~~~

There is no way (as of yet) to limit this to a specific server, so this is WAI

miekg:
If we want to fix this we should do per-server health and have a process health scraping all those per-server health points.
In fact, I see no other way to support this.

fturib:
The point is to dissociate the HTTP Servers from the Health object itself.
- one side : HTTP Server : we want one per address specified
- other side : Health object : Only collecting the Health info for all the Handler chains that are created from this Plugin. 

That was implemented that way in this PR : #1688. (but using also variables per instance ...)

### Simple way - Global var, but not RELOAD compatible
[...]

PR #1688, opened 2018-04-17
RELOAD capability for HEALTH, METRICS and any plugin that would use TCP Listeners

fturib:

1- introduce reusable Listener to allow the graceful Restart process of Caddy to success with Plugin that instanciate HTTP Listeners.
2- Use that new capability to ensure HEALTH and METRICS plugins are compatible with RELOAD
3- Fix HEALTH issue : #1686, and now allow sharing the listener among BlocServer or the whole Corefile 
4- METRICS: allow the directives "prometheus" to share the same HTTP Listener. (fix partially #1492)

NOTE: this PR replace the PR https://github.com/coredns/coredns/pull/1621, that I will close.

#1604
#1618
#1686 
#1492 

### 3. High level description of the PR

NOTE: Each of the following part is commited in a separated commit. That may help for review.

#### Adding Listener Distributor
plugins/pkg/listener
A Listener Distributor is a tool that is able to take reservations (booking) for listener to setup. It is able to:
- verify the address is not booked yet - at time of booking
- when allocate the Listener, is able to chain/reuse any existing Listener from the instance of CoreDNS to be stopped

#### Allow the Listener Distributor to be shared inside one instance, and chained on subsequent instances (RESTART process)
core/register
dnsContext is changed to be able to register a global listener Distributor available for all plugins of the coreDNS instance, and linked together the Distributor of the instance to be stopped with the Distributor of the new instance to be created (in order to be able to reuse the Listeners)

#### Fix and Enhance plugin HEALTH
plugin/health
Use the distributor to allow reuse of same Listener when restart
Allow sharing the HEALTH object (same listener) between several ServerChain (several ServerBloc of the same ServerBloc re-instanciated for another DNS listener). 

#### Fix and Enhance plugin METRICS
plugin/metrics
Use the distributor to allow reuse of same Listener when restart
Separated the METRICS objects in 2 parts: the Metrics that collect the data, and the LsnMetrics that server HTTP request and can be shared among several Metrics.
NOTE: to complete the fix of #1492, the PR https://github.com/coredns/coredns/pull/1682 will be needed in order to make difference between DNS queries served by different servers.

### 4. Which documentation changes (if any) need to be made?
TODO:
- update README for Health and Metrics
- will need to enhance a plugin blogs to explain for developers the logic of Setup, and how to use this new Distributor tool.

### DIFF
--- a/plugin/health/README.md
+++ b/plugin/health/README.md
@@ -33,47 +33,42 @@ health [ADDRESS] {
 * Where `lameduck` will make the process unhealthy then *wait* for **DURATION** before the process
   shuts down.
 
-If you have multiple Server Block and need to export health for each of the plugins, 
-you can either run all health enpoints on the same port, or on different ports. 
-In that case, 
+If you have multiple Server Blocks and need to export health for each of the plugins, 
+you can either run all health enpoints, like below:
 
 ~~~ corefile
-com {
+com. {
+    whoami
+    health
+}
+
+net. {
+    erratic
+    health
+}
+~~~
+
+
+or on different ports:
+
+~~~ corefile
+com. {
     whoami
     health :8080
 }
 
-net {
+net. {
     erratic
     health :8081
 }
 ~~~
 
-~~~ corefile
-com {
-    whoami
-    health
-}
-
-net {
-    erratic
-    health
-}
-~~~
 
  
-~~~ corefile
-com org net {
-    whoami
-    health
-}
-~~~
-
 ## Plugins
 
 Any plugin that implements the Healther interface will be used to report health.
-When the same port is reused, all the instances of the all the plugins that are associated with "health" 
-will participate to the result
+If ports are shared, multiple server plugins will be included as well.
 
 ## Metrics
 
--- a/plugin/health/README.md
+++ b/plugin/health/README.md
@@ -33,8 +33,9 @@ health [ADDRESS] {
 * Where `lameduck` will make the process unhealthy then *wait* for **DURATION** before the process
   shuts down.
 
-If you have multiple Server Block and need to export health for each of the plugins, you must run
-health endpoints on different ports:
+If you have multiple Server Block and need to export health for each of the plugins, 
+you can either run all health enpoints on the same port, or on different ports. 
+In that case, 
 
 ~~~ corefile
 com {
@@ -48,9 +49,31 @@ net {
 }
 ~~~
 
+~~~ corefile
+com {
+    whoami
+    health
+}
+
+net {
+    erratic
+    health
+}
+~~~
+
+ 
+~~~ corefile
+com org net {
+    whoami
+    health
+}
+~~~
+
 ## Plugins
 
 Any plugin that implements the Healther interface will be used to report health.
+When the same port is reused, all the instances of the all the plugins that are associated with "health" 
+will participate to the result
 
 ## Metrics
 
--- a/plugin/health/health.go
+++ b/plugin/health/health.go
@@ -2,12 +2,13 @@
 package health
 
 import (
-	"github.com/coredns/coredns/plugin/pkg/listener"
 	"io"
 	"net"
 	"net/http"
 	"sync"
 	"time"
+
+	"github.com/coredns/coredns/plugin/pkg/listener"
 
 	"github.com/coredns/coredns/plugin/pkg/log"
 )
--- a/plugin/health/health.go
+++ b/plugin/health/health.go
@@ -2,6 +2,7 @@
 package health
 
 import (
+	"github.com/coredns/coredns/plugin/pkg/listener"
 	"io"
 	"net"
 	"net/http"
@@ -13,8 +14,8 @@ import (
 
 // Health implements healthchecks by polling plugins.
 type health struct {
-	Addr     string
 	lameduck time.Duration
+	alloc    listener.AllocationToken
 
 	ln  net.Listener
 	mux *http.ServeMux
@@ -29,16 +30,25 @@ type health struct {
 }
 
 // newHealth returns a new initialized health.
-func newHealth(addr string) *health {
-	return &health{Addr: addr, stop: make(chan bool), pollstop: make(chan bool)}
+func newHealth(alloc listener.AllocationToken) *health {
+	return &health{alloc: alloc, stop: make(chan bool), pollstop: make(chan bool)}
+}
+
+func (h *health) Tag() string {
+	return "health"
+}
+
+func (h *health) registerHealther(plugins []Healther) {
+	h.Lock()
+	for _, p := range plugins {
+		h.h = append(h.h, p)
+	}
+	h.Unlock()
 }
 
 func (h *health) OnStartup() error {
-	if h.Addr == "" {
-		h.Addr = defAddr
-	}
 
-	ln, err := net.Listen("tcp", h.Addr)
+	ln, err := h.alloc.AllocateListener()
 	if err != nil {
 		return err
 	}
--- a/plugin/health/health.go
+++ b/plugin/health/health.go
@@ -34,7 +34,7 @@ func newHealth(alloc listener.AllocationToken) *health {
 	return &health{alloc: alloc, stop: make(chan bool), pollstop: make(chan bool)}
 }
 
-func (h *health) Tag() string {
+func (h *health) Name() string {
 	return "health"
 }
 
--- a/plugin/health/health.go
+++ b/plugin/health/health.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/coredns/coredns/plugin/pkg/listener"
-
 	"github.com/coredns/coredns/plugin/pkg/log"
 )
 
--- a/plugin/pkg/listener/listener.go
+++ b/plugin/pkg/listener/listener.go
@@ -51,7 +51,7 @@ func areSameIP(ip1 string, ip2 string) bool {
 	return false
 }
 
-func (b booking) isSame(protocol string, hostname string, port string) bool {
+func (b booking) equal(protocol string, hostname string, port string) bool {
 	// Consider that port "0" means "a new port" and will never match another port
 	// Consider that "localhost" or "" is matching both "127.0.01" or "[::1]"
 	if protocol == b.protocol {
@@ -150,7 +150,7 @@ func (l *Distributor) IsBooked(network string, address string) (bool, Booker) {
 		return false, nil
 	}
 	for _, b := range l.booked {
-		if b.isSame(network, hostname, port) {
+		if b.equal(network, hostname, port) {
 			return true, b.booker
 		}
 	}
@@ -165,7 +165,7 @@ func (l *Distributor) BookListener(network string, address string, bookerInfo Bo
 	}
 	// check if already booked - raise error
 	for _, booking := range l.booked {
-		if booking.isSame(network, hostname, port) {
+		if booking.equal(network, hostname, port) {
 			return nil, fmt.Errorf("listener booking for %s, (%s, %s) - an overlaping listener is already booked for %s (%s, %s) - sharing is not allowed", bookerInfo.Tag(), network, address, booking.booker.Tag(), booking.protocol, net.JoinHostPort(booking.hostname, booking.port))
 		}
 	}
@@ -175,8 +175,8 @@ func (l *Distributor) BookListener(network string, address string, bookerInfo Bo
 		if reuse != nil {
 			ntw := (*reuse.allocated).Addr().Network()
 			host, port, _ := net.SplitHostPort((*reuse.allocated).Addr().String())
-			if toBook.isSame(ntw, host, port) {
-				if reuse.isSame(network, hostname, port) && reuse.allocated != nil {
+			if toBook.equal(ntw, host, port) {
+				if reuse.equal(network, hostname, port) && reuse.allocated != nil {
 					// check that we reuse for the same family of Booker (same Tag)
 					if reuse.booker.Tag() != bookerInfo.Tag() {
 						return nil, fmt.Errorf("listener booking for %s, (%s, %s) - an overlaping listener is already in use for %s (%s, %s)", bookerInfo.Tag(), network, address, reuse.booker.Tag(), reuse.protocol, net.JoinHostPort(reuse.hostname, reuse.port))
--- a/plugin/pkg/listener/listener.go
+++ b/plugin/pkg/listener/listener.go
@@ -28,8 +28,9 @@ type booking struct {
 // TODO: manage properly the difference between localhost/"" and Ipv4 or Ipv6
 var (
 	sameIPList = [][]string{
-		{"127.0.0.1", "localhost", ""},
-		{"::1", "localhost", ""},
+		{"0.0.0.0", "::", ""},
+		{"127.0.0.1", "localhost"},
+		{"::1", "localhost"},
 	}
 )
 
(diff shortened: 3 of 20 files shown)


fturib:
NOTE, this PR resolve that question:
- when the same socket is reused but by another plugin ... what should we do ?

By returning all the errors at Booking time (that means in the Setup of the plugin) rather at Allocation time when the servers are already started.

In case of RESTART, if the Setup of the Plugins return an Error, then the new Instance is cancelled, the RESTART is considered as failed, and the service continue properly with the old instance, until someone fix the Corefile for RESTART.

plugin/health/health.go
```diff
@@ -28,16 +30,25 @@ type health struct {
...
-func newHealth(addr string) *health {
-	return &health{Addr: addr, stop: make(chan bool), pollstop: make(chan bool)}
+func newHealth(alloc listener.AllocationToken) *health {
+	return &health{alloc: alloc, stop: make(chan bool), pollstop: make(chan bool)}
 }
 
-func (h *health) OnStartup() error {
-	if h.Addr == "" {
-		h.Addr = defAddr
+func (h *health) Tag() string {
```
miekg: Can't we reuse Name() for this?

plugin/pkg/listener/listener.go
```diff
@@ -0,0 +1,209 @@
...
+func areSameIP(ip1 string, ip2 string) bool {
+	for _, l := range sameIPList {
+		if contains(l, ip1) && contains(l, ip2) {
+			return true
+		}
+	}
+	return false
+}
+
+func (b booking) isSame(protocol string, hostname string, port string) bool {
```
miekg: s/isSame/equal

miekg (commented):
Yes, I think we need something like this. Can you write a small blurp in plugin.md on how plugins should this? This would also help me in reviewing this PR. Thanks!

fturib:
@miekg : sure, will update the plugin.md, along with metrics.md and health.md
DONE.

plugin/health/health.go
fturib: Name() is for the handlers .. and Tag() for the Listeners. 
most likely it will not be the same object. 
NOTE: Health as no Name() function because it is not a Handler. the handler Metric has a Name(), but is not used for Listener -> it is a separated object LsnMetric that has a Tag()

The whole point to take another name for this identification was to avoid to confuse the developer about Listeners and Handlers (which are already confused with Plugins ...)

plugin/pkg/listener/listener.go
fturib: sure.

fturib:
I validate usage of Metrics with shared endpoint address, and Server label provided through the cts.

#### Corefile:
```
.henry.com:5002 francois.com:5002 louise.com:5003 {
    errors
    log
    health localhost:8080
    prometheus localhost:8075
	proxy . 8.8.8.8 {
		protocol dns force_tcp
	}
	reload 2s
}
.:5004 other.org:5005 {
    errors
    log
    health localhost:8080
    prometheus localhost:8075
	proxy . 8.8.8.8 {
		protocol dns force_tcp
	}
}
```

#### Scraping (partially) the metric results:

`curl "http://127.0.0.1:8075/metrics" | grep -v "# " | grep "request_count_total"`

```
coredns_dns_request_count_total{family="1",proto="tcp",server="",zone="dropped"} 2
coredns_dns_request_count_total{family="1",proto="tcp",server="dns://:5002",zone="."} 2
coredns_dns_request_count_total{family="1",proto="tcp",server="dns://:5002",zone="francois.com."} 2
coredns_dns_request_count_total{family="1",proto="tcp",server="dns://:5002",zone="louise.com."} 1
coredns_dns_request_count_total{family="1",proto="tcp",server="dns://:5003",zone="louise.com."} 2
coredns_dns_request_count_total{family="1",proto="tcp",server="dns://:5004",zone="."} 1
coredns_dns_request_count_total{family="1",proto="tcp",server="dns://:5005",zone="other.org."} 1
coredns_dns_request_count_total{family="1",proto="udp",server="dns://:5002",zone="."} 5


```

plugin/health/health.go
fturib: However .. if you think it is preferable, I can change.

fturib:
I udpated the plugin.MD, along with

fturib:
Closed by error. Reopening this PR.

UladzimirTrehubenka:
**Still have issues (build from your branch).**

With Corefile
```
.:1053 {
	health
	forward . 8.8.8.8:53
	errors
	prometheus 127.0.0.1:9153
}
.:1054 {
	health
	forward . 208.67.222.222:53
	errors
	prometheus 127.0.0.1:9153
}
```
after graceful restart have error
```
[INFO] SIGUSR1: Reloading
[INFO] Reloading
[ERROR] SIGUSR1: listen tcp :8080: bind: address already in use
```
With Corefile
```
.:1053 {
	health 127.0.0.1:8080
	forward . 8.8.8.8:53
	errors
	prometheus :9153
}
.:1054 {
	health 127.0.0.1:8080
	forward . 208.67.222.222:53
	errors
	prometheus :9153
}
```
after graceful restart have error
```
[INFO] SIGUSR1: Reloading
[INFO] Reloading
[INFO] health - listening on network tcp - address 127.0.0.1:8080
[ERROR] Failed to start metrics handler: listen tcp :9153: bind: address already in use
[ERROR] SIGUSR1: listen tcp :9153: bind: address already in use
```
Working configuration
```
.:1053 {
	health 127.0.0.1:8080
	forward . 8.8.8.8:53
	errors
	prometheus 127.0.0.1:9153
}
.:1054 {
	health 127.0.0.1:8080
	forward . 208.67.222.222:53
	errors
	prometheus 127.0.0.1:9153
}
```
output
```
[INFO] SIGUSR1: Reloading
[INFO] Reloading
[INFO] health - listening on network tcp - address 127.0.0.1:8080
[INFO] prometheus - listening on network tcp - address 127.0.0.1:9153
[INFO] Reloading complete
```

fturib:
@UladzimirTrehubenka : I tested with the plugin RELOAD, not with the SIGUSR1 .. so maybe there is something that is not equivalent in the sequence of restart  ... let me have a look.

fturib:
@UladzimirTrehubenka  : I am testing with the same Corefile and have correct reload using SIGUSR1:

I am wondering if you have a correct start of the Caddy Server - I mean emitting the "InstanceStartupEvent" event AFETR the start.

```
	// Start your engines
	instance, err := caddy.Start(corefile)
	// Execute instantiation events
	caddy.EmitEvent(caddy.InstanceStartupEvent, instance)

	// Twiddle your thumbs
	instance.Wait()
```

Result on my laptop

`ftur-MBP13:coredns ftur$ ./coredns -conf ./corefile`

```
2018/04/19 11:35:06 [INFO] health - listening on network tcp - address 127.0.0.1:8080
2018/04/19 11:35:06 [INFO] prometheus - listening on network tcp - address 127.0.0.1:9153
.:1054
.:1053
2018/04/19 11:35:06 [INFO] CoreDNS-1.1.1
2018/04/19 11:35:06 [INFO] darwin/amd64, go1.9.4, 5943fe2f
CoreDNS-1.1.1
darwin/amd64, go1.9.4, 5943fe2f
2018/04/19 11:35:24 [INFO] SIGUSR1: Reloading
2018/04/19 11:35:24 [INFO] Reloading
2018/04/19 11:35:24 [INFO] health - listening on network tcp - address 127.0.0.1:8080
2018/04/19 11:35:24 [INFO] prometheus - listening on network tcp - address 127.0.0.1:9153
2018/04/19 11:35:24 [INFO] Reloading complete
2018/04/19 11:35:36 [INFO] SIGUSR1: Reloading
2018/04/19 11:35:36 [INFO] Reloading
2018/04/19 11:35:36 [INFO] health - listening on network tcp - address 127.0.0.1:8080
2018/04/19 11:35:36 [INFO] prometheus - listening on network tcp - address 127.0.0.1:9153
2018/04/19 11:35:36 [INFO] Reloading complete

```

UladzimirTrehubenka:
Corefile:
```
.:1053 {
	health
	forward . 8.8.8.8:53
	errors
	prometheus :9153
}
.:1054 {
	health
	forward . 208.67.222.222:53
	errors
	prometheus :9153
}
```
Full output:
```
[INFO] health - listening on network tcp - address [::]:8080
[INFO] prometheus - listening on network tcp - address [::]:9153
.:1053
.:1054
[INFO] CoreDNS-1.1.1
[INFO] linux/amd64, go1.9.4, 
CoreDNS-1.1.1
linux/amd64, go1.9.4, 
[INFO] SIGUSR1: Reloading
[INFO] Reloading
[ERROR] SIGUSR1: listen tcp :8080: bind: address already in use
```
From your output I see that listeneres are bind to 127.0.0.1 (not 0.0.0.0):
```
[INFO] health - listening on network tcp - address 127.0.0.1:8080
[INFO] prometheus - listening on network tcp - address 127.0.0.1:9153
```

fturib:
@UladzimirTrehubenka : you are right, that is linked to the address in format 0.0.0.0 or [::].

fturib:
my bad, I did not consider "0.0.0.0", [::] and "". Done, and seems ok for my tests.

UladzimirTrehubenka:
I rechecked too - looks working. But some test failed due race condition.

fturib:
Yes, I need to dive in the race condition. Looks we have 2 issues:
- the dup(FD) is not protected at low level (in the net libraries of go)
- but also, in health, AllocateListener() should not happen in same time as the http.Serve() - whatever reusage of Listeners.
I will dive into.

NOTE: I also need to add some mutex around shared variables.

UladzimirTrehubenka:
Observed that for reworked README.md we have no race:
```
~~~ corefile
com {
    whoami
    health
}

net {
    erratic
    health
}
~~~

~~~ corefile
com {
    whoami
    health :8081
}

net {
    erratic
    health :8082
}
~~~
 
~~~ corefile
com org net {
    whoami
    health
}
~~~
```
Chain: [8080, 8080] -> [8081, 8082] -> [8080].
If we have something like:
[8080, 8080] -> [8080, 8081] or [8080, 8080] -> [8080]
we get race.
Is it possible that the issue is in the test not in the code?

plugin/health/README.md
```diff
@@ -33,8 +33,9 @@ health [ADDRESS] {
 * Where `lameduck` will make the process unhealthy then *wait* for **DURATION** before the process
   shuts down.
 
-If you have multiple Server Block and need to export health for each of the plugins, you must run
-health endpoints on different ports:
+If you have multiple Server Block and need to export health for each of the plugins,
```
miekg: Server BlockS

plugin/health/README.md
```diff
@@ -33,8 +33,9 @@ health [ADDRESS] {
 * Where `lameduck` will make the process unhealthy then *wait* for **DURATION** before the process
   shuts down.
 
-If you have multiple Server Block and need to export health for each of the plugins, you must run
-health endpoints on different ports:
+If you have multiple Server Block and need to export health for each of the plugins, 
+you can either run all health enpoints on the same port, or on different ports. 
+In that case,
```
miekg: Not clear what case you are referring too, the one with same ports or different ports.

Also not completely clear on what the second Corefile shows here. (Or the third)

plugin/health/README.md
```diff
@@ -48,9 +49,31 @@ net {
...
+    whoami
+    health
+}
+~~~
+
 ## Plugins
 
 Any plugin that implements the Healther interface will be used to report health.
+When the same port is reused, all the instances of the all the plugins that are associated with "health" 
+will participate to the result
```
miekg: I think I know what you are trying to say, but doesn't seem to come through in this sentence.

```
Each health plugin will export the health of all the plugins (that export health) in the server. If ports are shared, multiple server plugins will be included as well.
```
or something like that.

plugin/health/health.go
miekg: I see you point, but right now it is exactly like Name()... (let me review further, but I think we should re-use name)

plugin/pkg/listener/listener.go
```diff
@@ -0,0 +1,208 @@
...
+type booking struct {
+	booker    Booker
+	hostname  string
+	port      string
+	protocol  string
+	reusable  *net.Listener
+	allocated *net.Listener
+}
+
+// a way to identify same hostname
```
miekg: this feels brittle *and* only deals with localhost. Can we get away with just not caring?

plugin/pkg/listener/listener.go
```diff
@@ -0,0 +1,208 @@
...
+// Booker - generic type providing information on the booker
+type Booker interface {
+	// Tag is used to verify re-usability of a Listener - listener is reusable only for Booker with the same Tag
+	Tag() string
+}
+
+// booking structure use at booking, and allocation time of the listener
+type booking struct {
+	booker    Booker
+	hostname  string
```
miekg: hostname, port and protocol looks awfully close to a net.Addr - sure you don't want to use that?

plugin/pkg/listener/listener.go
```diff
@@ -0,0 +1,208 @@
...
+
+		}
+	}
+	if ln == nil {
+		ln, err = net.Listen(b.protocol, net.JoinHostPort(b.hostname, b.port))
+		if err != nil {
+			return nil, err
+		}
+	}
+	b.allocated = &ln
```
miekg: net.Listen returns an interface type, looks odd to take address of this?

miekg (changes requested):
We do *we* need to do the adminstrator for this as the kernel also knows?

Is the following feasible:
* we have a special net.Listen variant that tries to set up the listener and if it fails do something simular do booking.AllocateListener() ?

miekg:
btw @fturib the documentation you added really helped my review. Thanks for that.

plugin/health/README.md
fturib: ok.

plugin/pkg/listener/listener.go
fturib: ok. changed to network/address

fturib:
About

> Is the following feasible:
> 
> we have a special net.Listen variant that tries to set up the listener and if it fails do something simular do booking.AllocateListener() ?


I looked around and found this library : https://github.com/libp2p/go-reuseport
It seems  is a complete re-implementation of Listener down to socket level ...  

WONDERING POINTS ARE:
- it requires SO_REUSEPORT and SO_REUSEADDR to be useful for us. But as I understand, we are using the same technique of duplicating the FD, so I guess should be ok for us.
- What happen when you stop the first listener ? - as it is the same technique, it would just close the first FD and other listener could go ahead.
- it is tested for linux and darwin - but seems implemented for linux/windows - is that enough for CoreDNS ?

NOTE: it would be very easy to use : just need to replace the Listeners to open.
We would not make any verification of reusing the port for ANOTHER plugin. (but I guess as it is transient during restart, it is not a big matter).


NOTE2: still need to fix sharing listener issue in Prometheus (but could be fixed separately)

fturib:
@miekg : as you are a network guy .. could you ahve a quick look at this lib `https://github.com/libp2p/go-reuseport` and give a feedback for usage within CoreDNS ?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] RELOAD capabi..." ]
>@miekg : as you are a network guy .. could you ahve a quick look at this lib `https://github.com/libp2p/go-reuseport` and give a feedback for usage within CoreDNS ?

Haven't seen it, but bit hesitant to go full in with something that is not in 
the stdlib.

I'm still not clear what a good solution for the prometheus metric plugin 
net.listener thing is - my PR looks OKish, but is also already quite big

fturib:
Main concerns are resolved by #1709 .
We do not need anymore the Listener distributor, and we may fix the now gap i RESTART fail another way.

I will close this PR, and maybe re-opened later small ones to address #1686, and #1492.

PR #1689, opened 2018-04-18
pkg/log: ability for debug logs

### LINKED ISSUE #1626: debug logging inside the plugins
#1625 does show the need for better logging when things fail; the log **plugin** can do this, but logs errors and queries *not* interesting stuff happening inside a plugin.

We should probably use/extend (find log package); that allows some of this to be possible. I personally don't want to clutter everything with log statement; but having a generic `debug` directive that enables excessive logging would be nice

miekg:
When the debug plugin is enabled all log.Debug calls will print to
standard; if not there are a noop (almost).

The log package wraps some standard log functions as well, so just
replacing "log" with "plugin/pkg/log" should be enough to use this
package.

Fixes #1626

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -11,6 +11,7 @@ import (
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/metrics/vars"
 	"github.com/coredns/coredns/plugin/pkg/edns"
+	"github.com/coredns/coredns/plugin/pkg/log"
 	"github.com/coredns/coredns/plugin/pkg/rcode"
 	"github.com/coredns/coredns/plugin/pkg/trace"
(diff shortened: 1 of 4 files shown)


miekg:
We might want to go full in here and introduce `Info`, `Warn` and `Error` as well, the `log.Printf("[INFO] ..")`, vs `log.Debug` might be a bit jarring... Opinions? /cc @johnbelamaric

johnbelamaric:
Yes, I think that would be better. Then in fact debug can control log level not just debug...

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] pkg/log: abil..." ]
>Yes, I think that would be better. Then in fact debug can control log level not just debug...

Ah, yeah, that could be nice extension in the future. Let me make this change 
and then update the callers in a another PR.

Issue #1687, opened 2018-04-17
kubernetes cluster domain returns services (dig cluster.local)


dlamotte:
Unsure if this is defined behavior or not, but I'm running against CoreDNS 1.1.1 and I noticed that when I `dig cluster.local` it returns `A` records for all services in the cluster (at least as far as I can tell).

I'm in the process of changing that base domain to my company's domain (k8s.company.com).  I was hoping to use k8s.company.com to resolve the apiserver IP, but if I choose to use CoreDNS as the authority for this as well, I cannot because it resolves to the kubernetes services.  I'm doing the following to host the kubernetes records and authority.

```
apiVersion: v1
data:
  Corefile: |
    .:53 {
        cache
        errors
        health
        file /etc/coredns/company.com.zone company.com
        kubernetes k8s.company.com in-addr.arpa ip6.arpa {
          fallthrough k8s.company.com in-addr.arpa ip6.arpa
        }
        log
        prometheus :9153
        proxy . /etc/resolv.conf
    }
  company.com.zone: |
    $ORIGIN company.com.

    ; default ttl
    $TTL 7200

    ;                   authority-domain  domain-of-zone-admin
    company.com.  IN  SOA   ns.company.com.       hostmaster.company.com. (
                        1       ; zone-serial-number
                        7200    ; refresh-time
                        900     ; retry-time
                        1209600 ; expire-time
                        86400 ) ; negative caching ttl

    ; resource record
    ; name   ttl    record-class  record-type   record-data
    ; company.com.      IN            NS            ns ; ns.company.com
    k8s.company.com  IN            A               10.1.1.5 ; kubernetes apiserver
```

With this configuration, I cannot do `dig k8s.company.com` and get `10.1.1.5`.  I get the services in the cluster.

I tried _hacking_ a solution with `rewrite` but was unsuccessful.

chrisohaver:
> returns A records for all services in the cluster. 

No, it’s not supposed to do that. There was a similar issue in the past when querying bare namespaces.  This should be fixed. Thanks for reporting it.

chrisohaver:
> With this configuration, I cannot do dig k8s.company.com and get 10.1.1.5. I get the services in the cluster.

This is partly because file comes after kubernetes in the plugin.cfg order...(order in your config doesn’t matter). If file was before kubernetes, then you could put “fallthrough” in file so that queries other than “k8s.company.com” fallthrough to k8s instead of sending nxdomain to the client.

[...]

PR #1692, opened 2018-04-18
plugin/kubernetes: check for bare zone query

chrisohaver:

checks for bare zone query and returns NODATA instead of all services.

#1687

### DIFF
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -273,7 +273,7 @@ func (k *Kubernetes) Records(state request.Request, exact bool) ([]msg.Service, 
 	if e != nil {
 		return nil, e
 	}
-	if r.podOrSvc == ""{
+	if r.podOrSvc == "" {
 		return nil, nil
 	}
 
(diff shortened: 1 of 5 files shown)


plugin/kubernetes/kubernetes.go
```diff
@@ -273,6 +273,9 @@ func (k *Kubernetes) Records(state request.Request, exact bool) ([]msg.Service,
 	if e != nil {
 		return nil, e
 	}
+	if r.podOrSvc == ""{
```
bradbeam: nit probably want a space between `""` and `{`

bradbeam (changes requested):

chrisohaver:
FYI - I also found that bare `svc.cluster.local.` and `pod.cluster.local.` had bad behaviors.

`svc.cluster.local.` returned all services.
`pod.cluster.local.` returned `NXDOMAIN`

Both fixed in this PR now to return `NODATA`.

bradbeam (approved):
nice

plugin/kubernetes/parse.go
```diff
@@ -35,6 +35,9 @@ func parseRequest(state request.Request) (r recordRequest, err error) {
 	// Federations are handled in the federation plugin. And aren't parsed here.
 
 	base, _ := dnsutil.TrimZone(state.Name(), state.Zone)
+	if base == "" || base == Svc || base == Pod {
```
miekg: Small comment here? I.e. "return NODATA for apex queries?"

miekg: apart from this nit: lgtm.

PR #1695, opened 2018-04-19
Add Z Lab to adopters list

summerwind:
We are using CoreDNS heavily in production. This PR adds Z Lab to adopters list.
Thank you for building great product!


miekg:
Thank you!

PR #1699, opened 2018-04-19
global: move to context

miekg:
Move from golang.org/x/net/context to std lib's context.

Change done with:

for i in $(grep -l '/context' **/*.go); do sed -e 's|golang.org/x/net/context|context|' -i $i; echo $i; done
for i in **/*.go; do goimports -w $i; done

~~Removed 'go get' from Makefile as well.~~

### DIFF
--- a/pb/dns.pb.go
+++ b/pb/dns.pb.go
@@ -18,7 +18,8 @@ import fmt "fmt"
 import math "math"
 
 import (
-	context "golang.org/x/net/context"
+	context "context"
+
 	grpc "google.golang.org/grpc"
 )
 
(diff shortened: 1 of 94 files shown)


pb/dns.pb.go
```diff
@@ -18,7 +18,8 @@ import fmt "fmt"
 import math "math"
 
 import (
-	context "golang.org/x/net/context"
+	context "context"
```
johnbelamaric, line 21: This is a generated file, we shouldn't be updating it. Does the latest version of `protoc-gen-go` use `context`?

johnbelamaric: Hmm: https://github.com/golang/protobuf/blob/master/protoc-gen-go/grpc/grpc.go#L56

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] global: move ..." ]
>
>
>
>> @@ -18,7 +18,8 @@ import fmt "fmt"
> import math "math"
>
> import (
>-	context "golang.org/x/net/context"
>+	context "context"
>
>Hmm: https://github.com/golang/protobuf/blob/master/protoc-gen-go/grpc/grpc.go#L56

What I thought, hence the manual edit. Note we haven't really document (by means 
of a Makefile) on how to generate this file. It also doesn't change much...

miekg:
Let's figure this out one we need to regen that file

PR #1700, opened 2018-04-19
Fix #1685

### LINKED ISSUE #1685: ETCD: A record and NS record for dns zone name
Hi!

Following situation: I have to provide a zone, for example `abc.def.gh`. This zone should contain an A record with the same name as the zone eg.  `abc.def.gh 1.1.1.1` and of course multiple host entries eg  `gg.abc.def.gh 2.2.2.2` and so on.

The problem is, that the path in the ETCD cannot be a directory and a key at the same time:
```
/my-prefix/gh/def/abc {"host": "1.1.1.1", "ttl":"60"}
/my-prefix/gh/def/abc/gg {"host": "2.2.2.2", "ttl":"60"}
```

This leads to the situation that I am not able to create the NS entries if I would like to use the zonename as host too, because the entry `/my-prefix/gh/def/abc/dns/ns/ns1` is not usable if the `/my-prefix/gh/def/abc` is already a *key*.

Bind config would be:
```
abc.def.gh.	IN	NS	bind.
abc.def.gh.	IN	A	1.1.1.1
```

Am I missing something?

m4r10k:

This is WIP. I have added a demo code to the A function of `backend_lookup.go` to allow the possibility to create A (and later AAAA) record(s) with the name of the zone itself. This is supported by bind and this should be possible with the ETCD backend too.

Therefore the ETCD path a.dns.zone or a1.a.dns.zone (for loadbalancing) is used to realize the behavior

#1685 

Should be made if we are sure how we like to handle it of course

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -14,7 +14,7 @@ import (
 
 // A returns A records from Backend or an error.
 func A(b ServiceBackend, zone string, state request.Request, previousRecords []dns.RR, opt Options) (records []dns.RR, err error) {
-	services, err := b.Services(state, false, opt)
+	services, err := checkZoneForRecord(b, zone, state, opt)
 	if err != nil {
 		return nil, err
 	}
@@ -78,7 +78,7 @@ func A(b ServiceBackend, zone string, state request.Request, previousRecords []d
 
 // AAAA returns AAAA records from Backend or an error.
 func AAAA(b ServiceBackend, zone string, state request.Request, previousRecords []dns.RR, opt Options) (records []dns.RR, err error) {
-	services, err := b.Services(state, false, opt)
+	services, err := checkZoneForRecord(b, zone, state, opt)
 	if err != nil {
 		return nil, err
 	}
@@ -401,6 +401,39 @@ func newAddress(s msg.Service, name string, ip net.IP, what uint16) dns.RR {
 	}
 	// Should always be dns.TypeAAAA
 	return &dns.AAAA{Hdr: hdr, AAAA: ip}
+}
+
+func checkZoneForRecord(b ServiceBackend, zone string, state request.Request, opt Options) ([]msg.Service, error) {
+	var services []msg.Service
+	var err error
+	// if the zone name itself is queried we fake the query to search for a special entry
+	// this is equivalent to the NS search code
+	if state.Name() == zone {
+		old := state.QName()
+		state.Clear()
+		state.Req.Question[0].Name = dnsutil.Join([]string{"apex.dns", zone})
+
+		services, err = b.Services(state, false, opt)
+		if err != nil {
+			// it might be possible, that "apex.dns.zone" does not exists
+			// we set back the query and try it once again to restore the backward compatibility behavior
+			state.Req.Question[0].Name = old
+			services, err = b.Services(state, false, opt)
+			// if it still errors, we return the error
+			if err != nil {
+				return services, err
+			}
+		}
+
+		state.Req.Question[0].Name = old
+	} else {
+		services, err = b.Services(state, false, opt)
+		if err != nil {
+			return services, err
+		}
+	}
+
+	return services, err
 }
 
 const hostmaster = "hostmaster"
--- a/plugin/etcd/README.md
+++ b/plugin/etcd/README.md
@@ -128,6 +128,48 @@ Querying with dig:
 reverse.skydns.local.
 ~~~
 
+### Zone name as A record
+
+The zone name itself can be used A record. This behavior can be achieved by writing special entries to the ETCD path of your zone. If your zone is named `skydns.local` for example, you can create an `A` record for this zone as follows:
+
+~~~
+% curl -XPUT http://127.0.0.1:2379/v2/keys/skydns/local/skydns/dns/apex -d value='{"host":"1.1.1.1","ttl":"60"}'
+~~~
+
+If you query the zone name itself, you will receive the created `A` record:
+
+~~~ sh
+% dig +short skydns.local @localhost
+1.1.1.1
+~~~
+
+If you would like to use DNS RR for the zone name, you can set the following:
+~~~
+% curl -XPUT http://127.0.0.1:2379/v2/keys/skydns/local/skydns/dns/apex/x1 -d value='{"host":"1.1.1.1","ttl":"60"}'
+% curl -XPUT http://127.0.0.1:2379/v2/keys/skydns/local/skydns/dns/apex/x2 -d value='{"host":"1.1.1.2","ttl":"60"}'
+~~~
+
+If you query the zone name now, you will get the following response:
+
+~~~ sh
+dig +short skydns.local @localhost
+1.1.1.1
+1.1.1.2
+~~~
+
+If you would like to use `AAAA` records for the zone name too, you can set the following:
+~~~
+% curl -XPUT http://127.0.0.1:2379/v2/keys/skydns/local/skydns/dns/apex/x3 -d value='{"host":"2003::8:1","ttl":"60"}'
+% curl -XPUT http://127.0.0.1:2379/v2/keys/skydns/local/skydns/dns/apex/x4 -d value='{"host":"2003::8:2","ttl":"60"}'
+~~~
+
+If you query the zone name now for `AAAA` now, you will get the following response:
+~~~ sh
+dig +short skydns.local AAAA @localhost
+2003::8:1
+2003::8:2
+~~~
+
 ## Bugs
 
 Only the etcdv2 protocol is supported.
--- a/plugin/etcd/lookup_test.go
+++ b/plugin/etcd/lookup_test.go
@@ -43,6 +43,25 @@ var services = []*msg.Service{
 	// Nameservers.
 	{Host: "10.0.0.2", Key: "a.ns.dns.skydns.test."},
 	{Host: "10.0.0.3", Key: "b.ns.dns.skydns.test."},
+	// Zone name as A record (basic, return all)
+	{Host: "10.0.0.2", Key: "x.skydns_zonea.test."},
+	{Host: "10.0.0.3", Key: "y.skydns_zonea.test."},
+	// Zone name as A (single entry).
+	{Host: "10.0.0.2", Key: "x.skydns_zoneb.test."},
+	{Host: "10.0.0.3", Key: "y.skydns_zoneb.test."},
+	{Host: "10.0.0.4", Key: "apex.dns.skydns_zoneb.test."},
+	// A zone record (rr multiple entries).
+	{Host: "10.0.0.2", Key: "x.skydns_zonec.test."},
+	{Host: "10.0.0.3", Key: "y.skydns_zonec.test."},
+	{Host: "10.0.0.4", Key: "a1.apex.dns.skydns_zonec.test."},
+	{Host: "10.0.0.5", Key: "a2.apex.dns.skydns_zonec.test."},
+	// AAAA zone record (rr multiple entries mixed with A).
+	{Host: "10.0.0.2", Key: "x.skydns_zoned.test."},
+	{Host: "10.0.0.3", Key: "y.skydns_zoned.test."},
+	{Host: "10.0.0.4", Key: "a1.apex.dns.skydns_zoned.test."},
+	{Host: "10.0.0.5", Key: "a2.apex.dns.skydns_zoned.test."},
+	{Host: "2003::8:1", Key: "a3.apex.dns.skydns_zoned.test."},
+	{Host: "2003::8:2", Key: "a4.apex.dns.skydns_zoned.test."},
 	// Reverse.
 	{Host: "reverse.example.com", Key: "1.0.0.10.in-addr.arpa."}, // 10.0.0.1
 }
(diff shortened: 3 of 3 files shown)


m4r10k:
@miekg Thank you for your clue about the tests, I have now integrated them too. Maybe you can have a look on it. The tests are covering three cases:
- the default behavior if you are not using A records on the zone level (backward compatibility)
- the case when you use a single A record entry for the zone
- the case when you use multiple A records (RR) for the zone

The PR itself resides now in a separate function to reduce code duplication. Hopefully this is OK?

plugin/backend_lookup.go
```diff
@@ -403,4 +403,37 @@ func newAddress(s msg.Service, name string, ip net.IP, what uint16) dns.RR {
 	return &dns.AAAA{Hdr: hdr, AAAA: ip}
 }
 
+func checkZoneForRecord(b ServiceBackend, zone string, state *request.Request, opt Options) ([]msg.Service, error) {
```
miekg: state should not be a pointer

plugin/backend_lookup.go
```diff
@@ -403,4 +403,37 @@ func newAddress(s msg.Service, name string, ip net.IP, what uint16) dns.RR {
...
+	var services []msg.Service
+	var err error
+	// if the zone name itself is queried we fake the query to search for a special entry
+	// this is equivalent to the NS search code
+	if state.Name() == zone {
+		old := state.QName()
+		state.Clear()
+		state.Req.Question[0].Name = "a.dns." + zone
```
miekg: Can we not *also* use `ns.dns.` ? I.e we put different records here, and I can see people wanting to have MX and TXT as well.
Also for address record we need A and AAAA for starts which makes `a` a weird choice.

miekg (changes requested):
This looks good, but I think we want to generalize it, i.e for this use case A and AAAA should be supported.

miekg:
damn, reusing ns.dns won't work because those already need to address msg record.

I think we should go for apex; and support A and AAAA records and later MX and TXT maybe

plugin/backend_lookup.go
m4r10k: In line [412](https://github.com/m4r10k/coredns/blob/b189084482f759c3b577c698e0a9be3e4a040c36/plugin/backend_lookup.go#L412) I am saving it in case to restore it as needed in [428](https://github.com/m4r10k/coredns/blob/b189084482f759c3b577c698e0a9be3e4a040c36/plugin/backend_lookup.go#L428) so that it is usable back in line [28](https://github.com/m4r10k/coredns/blob/b189084482f759c3b577c698e0a9be3e4a040c36/plugin/backend_lookup.go#L28) because the `state` is never returned and it must be corrected as otherwise dig will complain that the answer does not match the question. Which I saw in some cases for the `NS` query (I can provide more info tomorrow) If the state is passed by value, the possible correction of the state would maybe be lost on return and the rest of the function may dangle on it. But I think you are correct, as the change on the path is only made temporarily. I will change this.

m4r10k:
@miekg thats true. I've tried to implement it without breaking to much I think. But you are correct, MX and TXT are hard to manage this way. Maybe later it would be an idea to rethink the data model on how the records are saved to etcd (but that would be a huge change and not backward compatible). I mean, there can only be one (and exactly one) key for an entry like a.skydns.local. So maybe it would be better to have a value which reflects the different types of records for this key. For example: 
```
[
  {
    "type": "A",
    "host": "1.1.1.1"
  },
  {
    "type": "A",
    "host": "1.1.1.2"
  },
  {
    "type": "NS",
    "host": "1.1.1.2"
  },
  {
    "type": "TXT",
    "text": "mytext1"
  },
  {
    "type": "TXT",
    "text": "mytext2"
  },
  {
    "type": "AAAA",
    "host": "0:0:0:0:0:ffff:101:101"
  }
]
```

That won't save us from the `dns.skydns.local` special entry (or something similar) as the zone itself cannot be a key at this point, but it would be more flexible to handle. But I do not know what would be the performance impact on parsing such a structure on every query.... and this is only an idea but I think, it sticks more to the possibilities of a bind zone file combined with the options etcd can provide (unique key = unique entry). I think the difference of A vs AAAA is ip4 vs ip6 parsing, but I have to have look at it tomorrow - thats only a wild guess but maybe later on a real AAAA type would be better for example?

m4r10k:
@miek I have tested it now with the AAAA records manually too, and it works as expected. If I write an ipv6 address to a key eg `t.skydns.local {"host":"2003::8:1"}` I will not be able to write an `A` record with an ipv4 address to the same key anymore.

Should I create some tests for the AAAA case too now? This would mean, that I have to create some additional creative zone names :smirk:

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP Fix #1685..." ]
>@miekg thats true. I've tried to implement it without breaking to much I think. But you are correct, MX and TXT are hard to manage this way. Maybe later it would be an idea to rethink the data model on how the records are saved to etcd. I mean, there can only be one (and exactly one) key for an entry like a.skydns.local. So maybe it would be better to have a value which reflects the different types of records for this key. For example:
>```
>[
>  {
>    "type": "A",
>    "host": "1.1.1.1"
>  },
>  {
>    "type": "A",
>    "host": "1.1.1.2"
>  },
>  {
>    "type": "NS",
>    "host": "1.1.1.2"
>  },
>  {
>    "type": "TXT",
>    "text": "mytext1"
>  },
>  {
>    "type": "TXT",
>    "text": "mytext2"
>  },
>  {
>    "type": "AAAA",
>    "host": "0:0:0:0:0:ffff:101:101"
>  }
>]```
>
>That won't save us from the `dns.skydns.local` special entry (or something similar) as the zone itself cannot be a key at this point, but it would be more flexible to handle. But I do not know what would be the performance impact on parsing such a structure on every query.... and this is only an idea but I think, it sticks more to the possibilities of an bind zone file combined with the options etcd can provide (unique key = unique entry). I think the difference of A vs AAAA is ip4 vs ip6 parsing, but I have to have look at it tomorrow - thats only a wild guess but maybe later on, a real AAAA type would be better?

OK, point taken, let's do the minimal things here and extend the "theme"

ns.dns -> ns records
a.dns -> a records
aaaa.dns -> aaaa records

leave out the rest (i.e. txt and what not).

Would that works? And yes aaaa should also be tested :)

m4r10k:
> OK, point taken, let's do the minimal things here and extend the "theme"
> ns.dns -> ns records
> a.dns -> a records
> aaaa.dns -> aaaa records
> leave out the rest (i.e. txt and what not).
> Would that works? And yes aaaa should also be tested :)

Perhaps this would be possible, but this would change the common behavior:
> I have tested it now with the AAAA records manually too, and it works as expected. If I write an ipv6 address to a key eg t.skydns.local {"host":"2003::8:1"} I will not be able to write an A record with an ipv4 address to the same key anymore.

I would suggest to not change this, until we take the big picture into account. Otherwise there would be a different behavior for the zone name entries and the *normal* records. Currently, it is not possible to have A **and** AAAA records at the same time for any record. You can only have A **or** AAAA entries for a record. From the test I've made (without looking into the whole code) the situations is at follows.

I will add the tests for AAAA soon.

###  tl;dr

#### first use a ipv4 converted to ipv6 entry which response as ipv4 A record only
```
 λ curl -XPUT http://127.0.0.1:2379/v2/keys/skydns/test/skydns/t -d value='{"host":"0:0:0:0:0:ffff:101:101"}'
{"action":"set","node":{"key":"/skydns/test/skydns/t","value":"{\"host\":\"0:0:0:0:0:ffff:101:101\"}","modifiedIndex":1138,"createdIndex":1138}}
 λ dig +short t.skydns.test @127.0.0.1
1.1.1.1
 λ dig +short t.skydns.test AAAA @127.0.0.1
.... no entry found
```
#### then create an real ipv6 entry which resolves to an AAAA record but NOT to an ipv4 entry
```
λ curl -XPUT http://127.0.0.1:2379/v2/keys/skydns/test/skydns/t -d value='{"host":"2003::8:1"}'
{"action":"set","node":{"key":"/skydns/test/skydns/t","value":"{\"host\":\"2003::8:1\"}","modifiedIndex":1140,"createdIndex":1140}}
λ dig +short t.skydns.test @127.0.0.1
.... no entry found
λ dig +short t.skydns.test AAAA @127.0.0.1
2003::8:1
```
At this point, I **cannot** create an `A` record any more, as the key in etcd is already used up.

#### the rest of the code (RR) works as expected
```
λ curl -XPUT http://127.0.0.1:2379/v2/keys/skydns/test/skydns/dns/a/a2 -d value='{"host":"2003::8:2"}'
{"action":"set","node":{"key":"/skydns/test/skydns/dns/a/a2","value":"{\"host\":\"2003::8:2\"}","modifiedIndex":1148,"createdIndex":1148}}
λ curl -XPUT http://127.0.0.1:2379/v2/keys/skydns/test/skydns/dns/a/a1 -d value='{"host":"2003::8:1"}'
{"action":"set","node":{"key":"/skydns/test/skydns/dns/a/a1","value":"{\"host\":\"2003::8:1\"}","modifiedIndex":1149,"createdIndex":1149}}
λ dig +short t.skydns.test AAAA @127.0.0.1
2003::8:1
λ dig +short skydns.test AAAA @127.0.0.1
2003::8:2
2003::8:1
```
Adding the `type` `aaaa.dns...` would change this *default* behavior for the zone name only and we would not like to change it at this time maybe, as we have to change it once more if we are thinking about another data handling.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP Fix #1685..." ]
>> OK, point taken, let's do the minimal things here and extend the "theme"
>> ns.dns -> ns records
>> a.dns -> a records
>> aaaa.dns -> aaaa records
>> leave out the rest (i.e. txt and what not).
>> Would that works? And yes aaaa should also be tested :)
>
>Perhaps this would be possible, but this would change the common behavior:
>> I have tested it now with the AAAA records manually too, and it works as expected. If I write an ipv6 address to a key eg t.skydns.local {"host":"2003::8:1"} I will not be able to write an A record with an ipv4 address to the same key anymore.
>
>I would suggest to not change this, until we take the big picture into account. Otherwise there would be a different behavior for the zone name entries and the *normal* records. Currently, it is not possible to have A **and** AAAA records at the same time for any record. You can only have A **or** AAAA entries for a record. From the test I've made (without looking into the whole code) the situations is at follows:
>

hmmm, also true, but note that having a and aaaa should be supported by:

 x1.bla.cluster.local
 x2.bla.cluster.local

putting the a records under x1 and the aaaa under x2 and then query bla.cluster.local.

I think we do want a directory for apex records, which would be my previous 
suggestion of using apex. for this.

m4r10k:
Ah, OK, now I know what you mean. This case is already working but needs a test too:
```
λ curl -XPUT http://127.0.0.1:2379/v2/keys/skydns/test/skydns/dns/a/a1 -d value='{"host":"2003::8:1"}'
{"action":"set","node":{"key":"/skydns/test/skydns/dns/a/a1","value":"{\"host\":\"2003::8:1\"}","modifiedIndex":1151,"createdIndex":1151}}
λ curl -XPUT http://127.0.0.1:2379/v2/keys/skydns/test/skydns/dns/a/a2 -d value='{"host":"1.1.1.1"}'
{"action":"set","node":{"key":"/skydns/test/skydns/dns/a/a2","value":"{\"host\":\"1.1.1.1\"}","modifiedIndex":1152,"createdIndex":1152}}
λ dig +short skydns.test AAAA @127.0.0.1
2003::8:1
λ dig +short skydns.test A @127.0.0.1
1.1.1.1
```

The apex directory for this kind of entries of course is the `a` directory beneath `dns.skydns.local`, as the `dns` entry is already used for `NS`. So yes, the PR can handle this but maybe we should name it `zone` (or anything else, suggestions are welcome) as in `zone.dns.skydns.local` instead of `a` to avoid confusion?

m4r10k:
Work done:
- Use `zone` for the zone name entries under `dns...`
- Update readme to reflect the change
- Add test for the A/AAAA mixed records case

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP Fix #1685..." ]
>The apex directory for this kind of entries of course is the `a` directory beneath `dns.skydns.local`, as the `dns` entry is already used for `NS`. So yes, the PR can handle this but maybe we should name it `zone` as in `zone.dns.skydns.local` instead of `a` to avoid confusion?

zone may be OK, the (somewhat official word) for this part of a zone is 'apex' 
hence why I proposed that.

m4r10k:
I have updated the code, but the travis job failed and I cannot see why. May it be possible to re-run it once again? Maybe I did the two pushes to fast behind one another (because I forgot to change the readme). Thx!

chrisohaver:
I re-ran the failed job.  Every once in a while a travis test will flake... will see if that was the case.

m4r10k:
@chrisohaver Thank you too! All good now!

m4r10k:
@miekg Anything else that we need on this one? I am still open for suggestions and hints of course!

plugin/backend_lookup.go
```diff
@@ -403,4 +403,37 @@ func newAddress(s msg.Service, name string, ip net.IP, what uint16) dns.RR {
...
+	// if the zone name itself is queried we fake the query to search for a special entry
+	// this is equivalent to the NS search code
+	if state.Name() == zone {
+		old := state.QName()
+		state.Clear()
+		state.Req.Question[0].Name = "apex.dns." + zone
+
+		services, err = b.Services(state, false, opt)
+		if err != nil {
+			// it might be possible, that "a.dns.zone" does not exists
```
miekg: a.dns.zone -> apex.dns.zone

plugin/backend_lookup.go
```diff
@@ -403,4 +403,37 @@ func newAddress(s msg.Service, name string, ip net.IP, what uint16) dns.RR {
...
 
+func checkZoneForRecord(b ServiceBackend, zone string, state request.Request, opt Options) ([]msg.Service, error) {
+	var services []msg.Service
+	var err error
+	// if the zone name itself is queried we fake the query to search for a special entry
+	// this is equivalent to the NS search code
+	if state.Name() == zone {
+		old := state.QName()
+		state.Clear()
+		state.Req.Question[0].Name = "apex.dns." + zone
```
miekg: technically zone can be ".", this plussing of name leads to an invalid name, use `dnsutil.Join`

miekg (commented):
This code looks good, almost lgtm to that.

We also use this in the kubernetes plugin (I just realized), but I think that should be OK...
/cc @chrisohaver

miekg (approved):

plugin/backend_lookup.go
m4r10k: 😀 thats true, as it is in line [336](https://github.com/m4r10k/coredns/blob/ab9ef573e18e65343657edf762a03280b897e46f/plugin/backend_lookup.go#L336) for `NS`

plugin/backend_lookup.go
m4r10k: Ups, will correct this tomorrow!

m4r10k:
Just changed the comment, but it seems like Travis does not like me :grin:

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fix #1685 (#1..." ]
>Just changed the comment, but it seems like Travis does not like me :grin:

travis has shit itself. Don't know why:
https://github.com/coredns/coredns/issues/1741

m4r10k:
I thought so. Thanks for info!

plugin/backend_lookup.go
miekg: please fix so I can merge.
(also rebase on master so travis won't fail, please :-) )

m4r10k:
@miekg Seems like that I have done it the right way 😁

Issue #1666, opened 2018-04-09
plugin/forward: queries are getting stuck in forward plugin during restart (reload)


rdrozhdzh:
During graceful restart (reload) the coredns calls registered `onShutdown()` handlers for the plugin objects to be terminated. Unfortunately, at this time server is not stopped yes and it continues accepting and processing new queries. So, for the same plugin `ServeDNS()` and `onShutdown()` can be called in any order.

Calling `onShutdown()` in forward plugin leads to stopping transport object, i.e. it stops `connManager()` method which reads `Dial` requests from channel. However, `Dial` request may still arrive after stopping transport, and such request hangs during write to channel, because nobody reads the channel on another side.

The problem looks even more serious for requests received via grpc connection. Because grpc server is stopped with `GracefulStop()` method (from `google.golang.org/grpc` package) which waits until all incoming queries get processed. Therefore, in this case `caddy.Restart()` function is not finished at all

The problem can be fixed on forward plugin side by dropping requests which come after call of `onShutdown()`. But I guess other plugins may have similar issues. Maybe it's worth to revise caddy's restart process and first shutdown servers (I mean stop receiving new requests and wait until all in-progress requests leave plugin's pipeline) and then call the registered `onShutdown()` handlers

fturib:
Right now, the event sequence in Caddy is the event BEFORE the action:

START : onFirstStartup(), on Startup() .. then start the servers
RESTART: onRestart(),  onStartup(), start the new servers, onShutdown(), stop the old servers
STOP: (on SIGTERM/SIGQUIT) onShutdown(), onFinalShutdown(), then stop all servers

Looks like all the events are BEFORE the action on the servers.
Useful when you want to switch on the resources before the first Query. 
But for shutdown, if you switch down your resources you may not be able to serve up to the stop() of the servers.

WARNING: About reload, there is another drawback : if the start and Stop does not succeed, Caddy just stop to process the Restart. Depending WHERE is the error, you may end up with none of the new Server neither old server running correctly. (for instance if one of the Shutdown() is returning an error).

johnbelamaric:
Don't the sockets get passed on to the new servers before onShutdown is called? If so, then ServeDNS should be in the new server not the old server. We may still want to have onShutdown wait for the pending requests to clear out. But I don't think it should get new requests at that point.

fturib:
I am not very into the socket level.
But what I can see about the restart mechanisme is:
[...]

PR #1701, opened 2018-04-19
plugin/forward: gracefull stop

rdrozhdzh:
stop connection manager only when no queries in progress

#1666

none

### DIFF
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -35,6 +35,16 @@ func (p *Proxy) updateRtt(newRtt time.Duration) {
 }
 
 func (p *Proxy) connect(ctx context.Context, state request.Request, forceTCP, metric bool) (*dns.Msg, error) {
+	atomic.AddInt32(&p.inProgress, 1)
+	defer func() {
+		if atomic.AddInt32(&p.inProgress, -1) == 0 {
+			p.checkStopTransport()
+		}
+	}()
+	if atomic.LoadUint32(&p.state) != running {
+		return nil, errStopped
+	}
+
 	start := time.Now()
 
 	proto := state.Proto()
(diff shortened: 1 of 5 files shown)


plugin/forward/connect.go
```diff
@@ -34,6 +34,16 @@ func (p *Proxy) updateRtt(newRtt time.Duration) {
 }
 
 func (p *Proxy) connect(ctx context.Context, state request.Request, forceTCP, metric bool) (*dns.Msg, error) {
+	atomic.AddInt32(&p.inProgress, 1)
+	defer func() {
+		if atomic.AddInt32(&p.inProgress, -1) == 0 {
```
miekg, line 40: how is guaranteed to not hit zero while the proxy is still up and running? (i.e. we are not shutting down)

rdrozhdzh: it is ok to have zero here, method checkStopTransport() will not stop transport unless Proxy.state == stopping

plugin/forward/persistent.go
```diff
@@ -157,7 +157,18 @@ func (t *transport) Yield(c *dns.Conn) {
 }
 
 // Stop stops the transport's connection manager.
-func (t *transport) Stop() { t.stop <- true }
+func (t *transport) Stop() {
+	t.stop <- true
+
+	// at this point nothing should write to channels anymore
+	// closing channels to make tests panic on write
```
miekg: I find this `close`-list somewhat ugly, is this only needed for the test? If so can we just do it in the tests only?

miekg (commented):
overall lgtm

plugin/forward/persistent.go
rdrozhdzh: Ok, I will remove this.
The idea was is that panic on writing in closed channel is easier to detect and fix than hanging if nobody reads the channel

rdrozhdzh:
@miekg, the `proxy_test.go` has yet another reference to `golang.org/x/net/context`. After merge, this likely should be changed in PR #1699 as well. Or I can change it here, if you are going to merge 1699 first

miekg:
>@miekg, the `proxy_test.go` has yet another reference to `golang.org/x/net/context`. After merge, this likely should be changed in PR #1699 as well. Or I can change it here, if you are going to merge 1699 first

Merged 1699

miekg (approved):

Issue #522, opened 2017-02-11
provide an API to manage host records


anuaimi:
we have a traditional monolithic app that we are moving to microservices on kubernetes.  we would like to use coredns with our test and staging environments.  while coredns can detect the services running on kubernetes, it can not detect the services that are still running on a traditional VM.  we would like our orchestration code to populate the missing host overrides.  as a result, coredns should support one of the common dns apis.  examples might be dnssimple or zerigo.  I considered aws route53 but it is much more complicated.

dns services ruby (fog) supports:
http://fog.io/about/supported_services.html

dns services python (libcloud) support:
https://libcloud.readthedocs.io/en/latest/supported_providers.html#dns

Note, I would be willing to write the code for this issue

johnbelamaric:
Do you expect the services running on VMs to be in the same zone or will you have different zones for k8s and those services?

If they are different zones, you can use a separate middleware (say etcd) to manage those records.

If you want them to be the same zone, it's something we may be able to do with #515 in combination with etcd (or others). 

You can definitely build a middleware for any of these APIs. That can live in-tree or out-of-tree depending on how generally applicable it is. For an example out-of-tree middleware (the only one in existence so far as I know) take a look at https://github.com/infobloxopen/coredns-grpc.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/coredns] provide an API ..." ]
>Do you expect the services running on VMs to be in the same zone or will you have different zones for k8s and those services?
>
>If they are different zones, you can use a separate middleware (say etcd) to manage those records.
>
>If you want them to be the same zone, it's something we may be able to do with #515 in combination with etcd (or others).
>
>You can definitely build a middleware for any of these APIs. That can live in-tree or out-of-tree depending on how generally applicable it is. For an example out-of-tree middleware (the only one in existence so far as I know) take a look at https://github.com/infobloxopen/coredns-grpc.

To echo what John said, to "old" way of service discovery (before k8s) was
skydns + etcd; this is functionality coredns also supports. That API is just
the etcd2 API (etcd3 is an open issue: #341).

The generic DNS api for updating zones is called dynamic updates: rfc2136

anuaimi:
[...]

PR #1702, opened 2018-04-19
plugin/etcdv3: Add etcd v3 plugin

### LINKED ISSUE #341: middleware/etcd: etcd3 support
See https://github.com/skynetservices/skydns/pull/299

nitisht:
Add etcd v3 plugin

Fixes #341


plugin/etcdv3/README.md
```diff
@@ -0,0 +1,133 @@
...
+Querying with dig:
+
+~~~ sh
+% dig @localhost -x 10.0.0.127 +short
+reverse.skydns.local.
+~~~
+
+## Bugs
+
+Only the etcdv2 protocol is supported.
```
harshavardhana: Remove this message

plugin.cfg
```diff
@@ -48,6 +48,7 @@ file:file
 auto:auto
 secondary:secondary
 etcd:etcd
+etcdv3:etcdv3
```
harshavardhana: better to call this etcd3 instead of etcdv3 while the latter is technically correct

plugin/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,175 @@
...
+	if serv.TTL == 0 {
+		return etcdTTL
+	}
+	if etcdTTL < serv.TTL {
+		return etcdTTL
+	}
+	return serv.TTL
+}
+
+const (
```
harshavardhana: constants should be at the beginning of the file..

harshavardhana (changes requested):

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP plugin/et..." ]
>+const (
>
>constants should be at the beginning of the file..

There is no such rule AFAICT. At the bottom is fine (doing this myself as well - 
quite like it)

miekg:
you may need to tmp. add `go get github.com/coreos/etcd/clientv3`  to the make file so that we can build it in travis and friends (the godep update to fix the vendor dir can (and should) come later)

nitisht:
> you may need to tmp. add go get github.com/coreos/etcd/clientv3 to the make file so that we can build it in travis and friends (the godep update to fix the vendor dir can (and should) come later)

Thanks, will do that @miekg

miekg:
testing that makefile fix right now, CI barfs with
~~~
o test -v ./test/kubernetes/...
panic: http: multiple registrations for /debug/requests

goroutine 1 [running]:
net/http.(*ServeMux).Handle(0x28f91a0, 0x1bea346, 0xf, 0x285c720, 0x1c7c640)
	/usr/local/go/src/net/http/server.go:2270 +0x627
net/http.(*ServeMux).HandleFunc(0x28f91a0, 0x1bea346, 0xf, 0x1c7c640)
	/usr/local/go/src/net/http/server.go:2302 +0x55
net/http.HandleFunc(0x1bea346, 0xf, 0x1c7c640)
	/usr/local/go/src/net/http/server.go:2314 +0x4b
github.com/coreos/etcd/vendor/golang.org/x/net/trace.init.0()
	/tmp/tmp.AQ3DeOEBYO/src/github.com/coreos/etcd/vendor/golang.org/x/net/trace/trace.go:115 +0x42
github.com/coreos/etcd/vendor/golang.org/x/net/trace.init()
	:1 +0x1cd
github.com/coreos/etcd/vendor/google.golang.org/grpc.init()
	:1 +0x96
github.com/coreos/etcd/etcdserver/etcdserverpb.init()
	:1 +0x87
github.com/coreos/etcd/clientv3.init()
	:1 +0x5f
github.com/coredns/coredns/plugin/etcd3.init()
	:1 +0x7d
github.com/coredns/coredns/core/plugin.init()
	:1 +0x7f
github.com/coredns/ci/test/kubernetes.init()
	:1 +0x91
main.init()
	:1 +0x53
exit status 2
FAIL	github.com/coredns/ci/test/kubernetes	0.009s
Makefile:74: recipe for target 'test-k8s' failed
make: *** [test-k8s] Error 1
~~~

miekg:
oh shit, looks like you can't have etcd2 and etcd3 at the same time in the same binary - at least not how we do it

miekg:
yes, removing etcd makes it compile

miekg:
Interestingly we now (probably) need to modify etcd too to be able to merge this - not sure how that looks - we can also *delete* etcd and make this the default - which is fine, but means we need to support the old way of configuring etcd - which in turn means we may just as well overwrite etcd entirely and make this the new default.

The only downside to this is when something is using etcd2 and can't possibly upgrade to etcd3 - which I can't really imagine.

harshavardhana:
> There is no such rule AFAICT. At the bottom is fine (doing this myself as well -
> quite like it)

constants are like comments in the code sort of verbatim, it is better to define them at start of a file you can look at Go stdlib codebase for this approach. Not that its a rule but a good practice. I have never seen them added at the end of files which is quite confusing when navigating in editors.

nitisht:
> which in turn means we may just as well overwrite etcd entirely and make this the new default.

Shall I go ahead with this, then @miekg ?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP plugin/et..." ]
>> which in turn means we may just as well overwrite etcd entirely and make this the new default.
>
>Shall I go ahead with this, then @miekg ?

Yes, I see etcd3 is the better plugin and CoreOS also says (for years) that 
everyone should move... @m4r10k : you're a heavy user of the etcd plugin, what 
do you think? Specically: we keep the Corefile syntax exactly the same, but 
use the etcd3 protocol, instead of the v2 one?

/Miek

--
Miek Gieben

m4r10k:
> you're a heavy user of the etcd plugin, what
do you think? Specically: we keep the Corefile syntax exactly the same, but
use the etcd3 protocol, instead of the v2 one?

I've read through this PR and if the Corefile syntax stays the same, I see no problem in the change to the etcd v3 protocol the hard way, beside the migration of the entries from etcd v2 -> to etcd v3 for the inexperienced users.

Some background: We have etcd v3 clusters for kubernetes running (all on-prem) and the etcd v3 clusters can also be used with the etcd v2 protocol. So yes, we are storing the key/vales for coredns in the etcd v2 tree of the etcd v3 version currently. Using the etcd v3 tree won't be a great problem. Kubernetes support of etcd v2 as backend is [deprecated with kubernetes v 1.13 too](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.10.md) - in roughly 18 month I think.

Another benefit for us is, that the backup functionality of etcd v3 is much more usable than the one of etcd v2 for sure. So :+1: from my point of view.

miekg:
What do you mean with "migrating", I was under the impression you could
just flip the transport and the etcd as-is?

m4r10k:
@miekg I am not sure but if I use etcdctl to get key/values written with the v2 than they are not visible if they are asked by a v3 protocol query:

```
/ # etcdctl --cert-file /certs/coredns.pem --key-file /certs/coredns-key.pem --ca-file /certs/ca.pem --endpoint https://x.x.x.x:2379 get /dns-internal/com/aaa/bbb/ccc
{"host":"10.x.x.x","ttl":60}
/ # ETCDCTL_API=3 etcdctl --cert /certs/coredns.pem --key /certs/coredns-key.pem --cacert /certs/ca.pem --endpoints  https://x.x.x.x:2379 ge
t /dns-internal/com/aaa/bbb/ccc
```

The same is true vice versa:
```
/ # ETCDCTL_API=3 etcdctl --cert /certs/coredns.pem --key /certs/coredns-key.pem --cacert /certs/ca.pem --endpoints h https://x.x.x.x:2379 put /foo bar
OK
/ # ETCDCTL_API=3 etcdctl --cert /certs/coredns.pem --key /certs/coredns-key.pem --cacert /certs/ca.pem --endpoints h https://x.x.x.x:2379 get /foo
/foo
bar
/ # etcdctl --cert-file /certs/coredns.pem --key-file /certs/coredns-key.pem --ca-file /certs/ca.pem --endpoint ht https://x.x.x.x:2379 get /foo
Error:  100: Key not found (/foo) [132
```
I have not tested it with the etcd golang clientv3 library yet, to produce evidence if it is the case then too. If it is true, than someone have to migrate the entries from v2 to v3. This might be a special case for the etcdctl binary but not for the library - I do not know. Maybe I can test this later...

miekg:
Damn, then doing a switcheroo is not backwards compatible.

@nitisht how hard would the following be to implement:
make etcd3 the new etcd and add a directive `version VERSION` that *defaults* to version 2, but can take a `3` as well; for 2 we use etcdv2 for 3 we use etcdv3 - this is backwards compatible and gives people to upgrade to 3 on their leisure.

m4r10k:
> oh shit, looks like you can't have etcd2 and etcd3 at the same time in the same binary - at least not how we do it

I have not looked in this and I didn't tried it until yet, but I looked into https://github.com/coreos/etcd/tree/master/etcdctl briefly. I am not that coding guy, but I guess they made modules to separate the v2 code from the v3 code (ctlv2/ctlv3) and use accordingly [here](https://github.com/coreos/etcd/blob/b933603a4331054d0215819666b350ac3bba99f4/etcdctl/main.go#L30). 

So an option might be to use `etcd` as v2 plugin and `etcdv3` as v3 plugin and start from scratch there? Then it would be not a Version switch (additional option to the existing plugin) but a plugin switch which would be more "clear"?

nitisht:
> how hard would the following be to implement:

Should not be too difficult @miekg . I will take a dig at it during the coming weekend.

nchrisdk:
@nitisht any news on the progress of this PR?

nitisht:
been busy with some work stuff, but will update it in a couple of days @nchrisdk

miekg:
while talking with @m4r10k today at KubeCon I got the idea to go full on with etcd and allow all RR types in etcd instead of the subset we have now.

It does intersect with the sql work as well. I might abstract away the backend in the *file* plugin and see how hard it actually is (encoding RR's is probably the hard part)

johnbelamaric:
Also consider that we may want both a *read* backend interface and a *write* backend interface, ala #522. Plugins could implement one or both. Also then we can build APIs on top of the write interface - DDNS, gRPC, REST.

nchrisdk:
@nitisht I can understand that. Let me know if I can help.

nitisht:
> make etcd3 the new etcd and add a directive version VERSION that defaults to version 2, but can take a 3 as well; for 2 we use etcdv2 for 3 we use etcdv3 - this is backwards compatible and gives people to upgrade to 3 on their leisure.

@miekg sorry for the delay and thank you for your patience! I am going to work on this now. Since it has been a while, just wanted to confirm this is how we still want it ?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP plugin/et..." ]
>> make etcd3 the new etcd and add a directive version VERSION that defaults to version 2, but can take a 3 as well; for 2 we use etcdv2 for 3 we use etcdv3 - this is backwards compatible and gives people to upgrade to 3 on their leisure.
>
>@miekg sorry for the delay and thanks for your patience! I am going to work on this now. Since it has been a while, just wanted to confirm this is how we still want it ?

Cool! And yes, this is the (proposed) way forward as I haven't seen any other 
proposals.

miekg:
@nitisht this is ready for review again, isn't? We then need to release a 1.2.0 which deprecates some stuff and then work on merging this and dropping the current etcd2 plugin.

I don't see anything massive that is holding up a 1.2.0 release so we can do this pretty soon.

nitisht:
hey @miekg I have the changes ready as per the approach we discussed, need to test it still though. okay to do this in a couple of days?

miekg:
Sgtm. Thanks!

nitisht:
@miekg I justed tested locally and looks like it is working fine now

nitisht:
@miekg is there something else I can do here to get this in?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/etcdv3..." ]
>@miekg is there something else I can do here to get this in?

No I should review and release 1.2.0 because we can't merge this without 
breakage. Both should happen soon, 1.2.0 I can probably release this evening.

nitisht:
Great, thanks for the info

plugin/etcd3/etcd.go
```diff
@@ -56,7 +64,7 @@ func (e *Etcd) Lookup(state request.Request, name string, typ uint16) (*dns.Msg,
 
 // IsNameError implements the ServiceBackend interface.
 func (e *Etcd) IsNameError(err error) bool {
-	if ee, ok := err.(etcdc.Error); ok && ee.Code == etcdc.ErrorCodeKeyNotFound {
+	if err.Error() == keyNotFoundErr {
```
miekg: This can be simpllified

miekg: return err.Errror() == keyNotfoundErr

plugin/etcd3/etcd.go
```diff
@@ -132,16 +128,13 @@ Nodes:
...
 		}
-		b := msg.Service{Host: serv.Host, Port: serv.Port, Priority: serv.Priority, Weight: serv.Weight, Text: serv.Text, Key: n.Key}
-		if _, ok := bx[b]; ok {
-			continue
-		}
+		b := msg.Service{Host: serv.Host, Port: serv.Port, Priority: serv.Priority, Weight: serv.Weight, Text: serv.Text, Key: string(n.Key[:])}
 		bx[b] = true
 
-		serv.Key = n.Key
+		serv.Key = string(n.Key[:])
```
miekg: Why is the `[:]` need here?

plugin/etcd3/handler.go
```diff
@@ -83,7 +81,7 @@ func (e *Etcd) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 
 	m := new(dns.Msg)
 	m.SetReply(r)
-	m.Authoritative, m.RecursionAvailable = true, true
+	m.Authoritative, m.RecursionAvailable, m.Compress = true, true, true
```
miekg: compress is done separately now, please remove.

plugin/etcd3/lookup_test.go
```diff
@@ -43,25 +42,6 @@ var services = []*msg.Service{
 	// Nameservers.
 	{Host: "10.0.0.2", Key: "a.ns.dns.skydns.test."},
 	{Host: "10.0.0.3", Key: "b.ns.dns.skydns.test."},
```
miekg: why are these tests removed?

miekg (commented):
looks good. I think we should just name it etcd as we're planning to replace it wholesale.

plugin/etcd3/lookup_test.go
nitisht: thanks for the review, looks like there were few tests not running due to renaming. Now that I switched back to the plugin name `etcd` few tests are failing. I will try to fix these today hopefully

nitisht:
@miekg other failing tests are fixed now, only failing test is `TestOtherLookup` https://travis-ci.org/coredns/coredns/jobs/397180376#L898

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/etcdv3..." ]
>@miekg other failing tests are fixed now, only failing test is `TestOtherLookup` https://travis-ci.org/coredns/coredns/jobs/397180376#L898

Ah nice. I should have time this evening (UTC) to help debug this.

nitisht:
> Ah nice. I should have time this evening (UTC) to help debug this.

Awesome, thanks!

miekg:
The failing test is:
~~~
--- FAIL: TestOtherLookup (0.12s)
	helpers.go:131: Additional for "a.mail.skydns.test." contained 10 results, 2 expected
	helpers.go:301: ;; opcode: QUERY, status: NOERROR, id: 24903
		;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 10
		
		;; QUESTION SECTION:
		;a.mail.skydns.test.	IN	 MX
		
		;; ANSWER SECTION:
		a.mail.skydns.test.	300	IN	MX	50 mx.skydns.test.
		
		;; ADDITIONAL SECTION:
		a.ipaddr.skydns.test.	300	IN	A	172.16.1.1
		a.ipaddr.skydns.test.	300	IN	A	172.16.1.1
		a.ipaddr.skydns.test.	300	IN	A	172.16.1.1
		a.ipaddr.skydns.test.	300	IN	A	172.16.1.1
		b.ipaddr.skydns.test.	300	IN	A	172.16.1.2
		mx.skydns.test.	300	IN	CNAME	a.ipaddr.skydns.test.
		mx.skydns.test.	300	IN	CNAME	a.ipaddr.skydns.test.
		mx.skydns.test.	300	IN	CNAME	a.ipaddr.skydns.test.
		mx.skydns.test.	300	IN	CNAME	a.ipaddr.skydns.test.
		mx.skydns.test.	300	IN	CNAME	b.ipaddr.skydns.test.
~~~
If I look at `OtherServices` for the mx we only have:
```
	{Host: "mx.skydns.test", Priority: 50, Mail: true, Key: "a.mail.skydns.test."},
	{Host: "mx.miek.nl", Priority: 50, Mail: true, Key: "b.mail.skydns.test."},
	{Host: "a.ipaddr.skydns.test", Priority: 30, Mail: true, Key: "a.mx.skydns.test."},
```
So I don't understand why `b.ipaddr.skydns.test` pops up here?

nitisht:
IIUC, this is present as well

```
{Host: "b.ipaddr.skydns.test", Mail: true, Key: "b.mx2.skydns.test."},
```

here: https://github.com/coredns/coredns/blob/master/plugin/etcd/other_test.go#L51

miekg:
but that has `mx2` why is it being picked up?

nitisht:
> but that has mx2 why is it being picked up?

I am not really sure, is there a way to run these tests locally? I can debug and try to find out

miekg:
yes, download the docker container:
docker run -d --net=host --name=etcd quay.io/coreos/etcd:v3.3

and run the tests with : go test -tags etcd

nitisht:
> but that has mx2 why is it being picked up?

the problem is change in the way etcd handles directories. etcdv2 supported directories, but there are only flat keys in etcdv3. 

This means a get for `/skydns/test/skydns/mx` returned everything under `/skydns/test/skydns/mx/` in etcdv2. 

Now with etcdv3, a get for `/skydns/test/skydns/mx` with prefix enabled returns everything matching `/skydns/test/skydns/mx*` - meaning `mx/...`, `mx2/..` and `mx3/..` all match the prefix. 

I am not sure how we should handle it now..I have added a quick fix here https://github.com/coredns/coredns/pull/1702/commits/b05ff150c167aa6d5f02e9ba20d3fb50a8f33695#diff-6d0109286d6cd87bf48c9c87a938d4f5R90 but this probably is not the right way..

nitisht:
FWIW this is no clear response filtering in etcdv3 so we could filter out irrelevant responses in get [[ref](https://github.com/coreos/etcd/issues/8467)]

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/etcdv3..." ]
>> but that has mx2 why is it being picked up?
>
>the problem is change in the way etcd handles directories. etcdv2 supported directories, but there are only flat keys in etcdv3.
>
>This means a get for `/skydns/test/skydns/mx` returned everything under `/skydns/test/skydns/mx/` in etcdv2.
>
>Now with etcdv3, a get for `/skydns/test/skydns/mx` with prefix enabled returns everything matching `/skydns/test/skydns/mx*` - meaning `mx/...`, `mx2/..` and `mx3/..` etc match the prefix.

Agh, yes now that makes sense. Thanks for digging into this.

>I am not sure how we should handle it now..I have added a quick fix here https://github.com/coredns/coredns/pull/1702/commits/b05ff150c167aa6d5f02e9ba20d3fb50a8f33695#diff-6d0109286d6cd87bf48c9c87a938d4f5R90 but this probably is not the right way..

No, this needs to be more generic. Can we just always the closing / ?
I'm assuming that will fail, but only for leaves with nothing below, this would 
require two queries though :(

/Miek

--
Miek Gieben

nitisht:
> Can we just always the closing / ?
I'm assuming that will fail, but only for leaves with nothing below, this would
require two queries though :(

Done

miekg:
merged! :-)

Can you send a follow up PR which documents this 2-lookups bit?

m4r10k:
Wow! Really cool!

yongtang:
👍 🎉 🎈 

FYI @JiachengXu

jiachengxu:
Cool!

harshavardhana:
Very cool thanks @miekg

miekg:
@harshavardhana credit where credit is due. I clicked the merge button, @nitisht did the work!

nitisht:
> Can you send a follow up PR which documents this 2-lookups bit?

~Sure, will do~
Done #1918

harshavardhana:
> @harshavardhana credit where credit is due. I clicked the merge button, @nitisht did the work!

Agreed @miekg - Thanks @nitisht

nchrisdk:
Great work! @nitisht

Issue #1691, opened 2018-04-18
Forward plugin doesn't support a sequential policy


sdonovan01:
The forward plugin currently only support random and round_robin policies for upstream server selection.  The default DNS resolution behavior is to check the first nameserver... then the second... and so on until all nameservers have been queried or a match is found.  Neither of these policies accomplish this.  I am suggesting that a new policy be added to the forward plugin named 'sequential' that provides the default DNS resolution behavior mentioned above.

I've looked at the forwarder code it it should be pretty simple to add this functionality.  I'll submit a pull request if it is agreed that this functionality is useful and the pull request will be accepted.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] Forward plugin do..." ]
>The forward plugin currently only support random and round_robin policies for upstream server selection.  The default DNS resolution behavior is to check the first nameserver... then the second... and so on until all nameservers have been queried or a match is found.  Neither of these policies accomplish this.  I am suggesting that a new policy be added to the forward plugin named 'sequential' that provides the default DNS resolution behavior mentioned above.
>
>I've looked at the forwarder code it it should be pretty simple to add this functionality.  I'll submit a pull request if it is agreed that this functionality is useful and the pull request will be accepted.

Yep that will be accepted. Note we have that policy in *proxy* and it's called 
"first" there, allthough now that I see it mentioned I quite like 'sequential'.

miekg:
... and fixed!

PR #1704, opened 2018-04-19
adding sequential policy to forward plugin

sdonovan01:
It adds a sequential policy to the forward plugin.  This policy will just return the list of proxies in the order they appear in the config.  This will enable the plugin to provide default DNS resolution behavior.

#1691 

The README.md from the forward plugin needs to be synced to the coredns/coredns.io repo.

### DIFF
--- a/man/coredns-forward.7
+++ b/man/coredns-forward.7
@@ -56,7 +56,7 @@ forward FROM TO\.\.\. {
     max_fails INTEGER
     tls CERT KEY CA
     tls_servername NAME
-    policy random|round_robin
+    policy random|round_robin|sequential
     health_check DURATION
 }
 .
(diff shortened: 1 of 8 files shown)


man/coredns-forward.7
```diff
@@ -56,7 +56,7 @@ forward FROM TO\.\.\. {
     max_fails INTEGER
     tls CERT KEY CA
     tls_servername NAME
-    policy random|round_robin
+    policy random|round_robin|sequential
```
miekg, line 59: this is generated, no need to add this

sdonovan01: You want me to back it out or are you OK with it as is (since it's just going to get stepped on anyway)?

miekg: you can leave it

miekg:
We should keep the policy names between both proxy and forward in sync. Probably means making sequential an alias for first in proxy as well (in this PR)

miekg (approved):

Issue #1706, opened 2018-04-20
can't reload Corefile in coredns


feresberbeche:
Hello,
I am using coredns with kubernetes and I when I use 
`kubectl -n kube-system edit configmap coredns`
and make changes to the Corefile, the changed are copied to inside the pods cause when I access them manually I find they are changed.
I use `kubectl exec -n kube-system coredns-*-* -- kill -SIGUSR1 1` with both pods of coredns
but when I use a dnstool pod to test if the new config works, well it dosen't.
Please tell me what to do. 

Here is my Corefile
  

>  Corefile: |
    .:53 {
        errors
        health
        rewrite continue {
        name regex ldap.feres.lab ldap.default.svc.cluster.local
        answer name ldap.default.svc.cluster.local ldap.feres.lab
        type ANY A
        }
        kubernetes cluster.local in-addr.arpa ip6.arpa {
           pods insecure
           upstream 8.8.8.8
           fallthrough in-addr.arpa ip6.arpa
        }
        prometheus :9153
        proxy . /etc/resolv.conf
        cache 30
    }

fturib:
Hi, as of today, the plugins HEALTH and PROMETHEUS are incompatible with Restart of CoreDNS (both RELOAD abs signal SIGUSR1)
There is a fix in progress : #1688.

Right now, if you want to use RELOAD, you need to get rid of these 2 other plugins in your Corefile.

PROMETHEUS : it depends if you use it or not.
HEALTH : I am not sure how important it is for the management of this croedns Pod in Kubernetes : @chrisohaver - can you answer here ?

feresberbeche:
@fturib Thanks for your answer, btw 
`kubeadm init --feature-gates CoreDNS=true
`
generates that Corefile by default, if those two plugins are not compatible please tell the kubeadm guys to rectify their default Corefile

chrisohaver:
You’ll have to delete the pods to get them to reload the config, for now… the reload function has a bug relating to health and prometheus plugins that prevents it from reloading successfully.  There’s an active PR to fix the issue.

```kubectl -n kube-system delete pod coredns-xxxxxxxx-zzzz```

It's not generally feasible to deploy coredns by default without a liveness probe and metrics just to fix the reload issue, since you can still reload the config by restarting the pods.  IOW, for most people, it's better to reload configs by restarting pods for the occasional config file change, than to permanently live without liveness checks.

[...]

PR #1668, opened 2018-04-11
fix readme.md of log plugin

mparamonau:
Fixed the readme.md according the PR https://github.com/coredns/coredns/pull/1664.

PR #1709, opened 2018-04-20
reload: use OnRestart

### LINKED ISSUE #1604: plugin/health plugin/prometheus graceful restart bug

miekg:
Close the listener on OnRestart for health and metrics so the default
setup function can setup the listener when the plugin is "starting up".

Lightly test with some SIGUSR1-ing. Also checked the reload plugin with
this, seems fine:

.com.:1043
.:1043
2018/04/20 15:01:25 [INFO] CoreDNS-1.1.1
2018/04/20 15:01:25 [INFO] linux/amd64, go1.10,
CoreDNS-1.1.1
linux/amd64, go1.10,
2018/04/20 15:01:25 [INFO] Running configuration MD5 = aa8b3f03946fb60546ca1f725d482714
2018/04/20 15:02:01 [INFO] Reloading
2018/04/20 15:02:01 [INFO] Running configuration MD5 = b34a96d99e01db4015a892212560155f
2018/04/20 15:02:01 [INFO] Reloading complete
^C2018/04/20 15:02:06 [INFO] SIGINT: Shutting down

With this corefile:
.com {
  proxy . 127.0.0.1:53
  prometheus :9054
  whoami
  reload
}

. {
  proxy . 127.0.0.1:53
  prometheus :9054
  whoami
  reload
}

The prometheus port was 9053, changed that to 54 so reload would pick it
up.

From a cursory look it seems this also fixes:
Fixes #1604 #1706

### DIFF
--- a/plugin/health/health.go
+++ b/plugin/health/health.go
@@ -61,6 +61,17 @@ func (h *health) OnStartup() error {
 	return nil
 }
 
+func (h *health) OnRestart() error {
+	// relingish our listener as we re-listen on successfull reload
+	if h.ln != nil {
+		if err := h.ln.Close(); err != nil {
+			return err
+		}
+		h.ln = nil
+	}
+	return nil
+}
+
 func (h *health) OnShutdown() error {
 	// Stop polling plugins
 	h.pollstop <- true
--- a/plugin/health/health.go
+++ b/plugin/health/health.go
@@ -83,11 +83,14 @@ func (h *health) OnShutdown() error {
 		time.Sleep(h.lameduck)
 	}
 
+	h.stop <- true
+	return nil
+}
+
+func (h *health) OnFinalShutdown() error {
 	if h.ln != nil {
 		return h.ln.Close()
 	}
-
-	h.stop <- true
 	return nil
 }
 
--- a/plugin/health/health.go
+++ b/plugin/health/health.go
@@ -72,7 +72,7 @@ func (h *health) OnRestart() error {
 	return nil
 }
 
-func (h *health) OnShutdown() error {
+func (h *health) OnFinalShutdown() error {
 	// Stop polling plugins
 	h.pollstop <- true
 	// NACK health
@@ -83,14 +83,11 @@ func (h *health) OnShutdown() error {
 		time.Sleep(h.lameduck)
 	}
 
-	h.stop <- true
-	return nil
-}
-
-func (h *health) OnFinalShutdown() error {
 	if h.ln != nil {
 		return h.ln.Close()
 	}
+
+	h.stop <- true
 	return nil
 }
 
--- a/plugin/health/health.go
+++ b/plugin/health/health.go
@@ -16,8 +16,9 @@ type health struct {
 	Addr     string
 	lameduck time.Duration
 
-	ln  net.Listener
-	mux *http.ServeMux
+	ln      net.Listener
+	nlSetup bool
+	mux     *http.ServeMux
 
 	// A slice of Healthers that the health plugin will poll every second for their health status.
 	h []Healther
@@ -45,6 +46,7 @@ func (h *health) OnStartup() error {
 
 	h.ln = ln
 	h.mux = http.NewServeMux()
+	h.nlSetup = true
 
 	h.mux.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
 		if h.Ok() {
@@ -64,7 +66,7 @@ func (h *health) OnStartup() error {
 func (h *health) OnRestart() error { return h.OnFinalShutdown() }
 
 func (h *health) OnFinalShutdown() error {
-	if h == nil {
+	if !h.nlSetup {
 		return nil
 	}
 
@@ -81,7 +83,7 @@ func (h *health) OnFinalShutdown() error {
 	h.ln.Close()
 
 	h.stop <- true
-	h.ln = nil
+	h.nlSetup = false
 	return nil
 }
 
--- a/plugin/health/health.go
+++ b/plugin/health/health.go
@@ -61,18 +61,13 @@ func (h *health) OnStartup() error {
 	return nil
 }
 
-func (h *health) OnRestart() error {
-	// relingish our listener as we re-listen on successfull reload
-	if h.ln != nil {
-		if err := h.ln.Close(); err != nil {
-			return err
-		}
-		h.ln = nil
-	}
-	return nil
-}
+func (h *health) OnRestart() error { return h.OnFinalShutdown() }
 
 func (h *health) OnFinalShutdown() error {
+	if h == nil {
+		return nil
+	}
+
 	// Stop polling plugins
 	h.pollstop <- true
 	// NACK health
@@ -83,11 +78,10 @@ func (h *health) OnFinalShutdown() error {
 		time.Sleep(h.lameduck)
 	}
 
-	if h.ln != nil {
-		return h.ln.Close()
-	}
+	h.ln.Close()
 
 	h.stop <- true
+	h.ln = nil
 	return nil
 }
 
--- a/plugin/metrics/metrics.go
+++ b/plugin/metrics/metrics.go
@@ -116,8 +116,8 @@ func (m *Metrics) OnRestart() error {
 	return nil
 }
 
-// OnShutdown tears down the metrics listener on shutdown and restart.
-func (m *Metrics) OnShutdown() error {
+// OnFinalShutdown tears down the metrics listener on shutdown and restart.
+func (m *Metrics) OnFinalShutdown() error {
 	// We allow prometheus statements in multiple Server Blocks, but only the first
 	// will open the listener, for the rest they are all nil; guard against that.
 	if m.ln != nil {
--- a/plugin/metrics/metrics.go
+++ b/plugin/metrics/metrics.go
@@ -86,10 +86,6 @@ func (m *Metrics) ZoneNames() []string {
 
 // OnStartup sets up the metrics on startup.
 func (m *Metrics) OnStartup() error {
-	if m.ln != nil {
-		return nil
-	}
-
 	ln, err := net.Listen("tcp", m.Addr)
 	if err != nil {
 		log.Errorf("Failed to start metrics handler: %s", err)
--- a/plugin/metrics/metrics.go
+++ b/plugin/metrics/metrics.go
@@ -86,7 +86,10 @@ func (m *Metrics) ZoneNames() []string {
 
 // OnStartup sets up the metrics on startup.
 func (m *Metrics) OnStartup() error {
-	println("Listening for", m.Addr)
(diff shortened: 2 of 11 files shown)


rdrozhdzh:
see my comment https://github.com/coredns/coredns/issues/1604#issuecomment-383125368

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] reload: use O..." ]
>see my comment https://github.com/coredns/coredns/issues/1604#issuecomment-383125368

Well, true, but we in a rock and hard place, and the complexity of #1688 is 
terrifying IMO.

This can leave with a closed listener which means - crash? (have to test). Note 
the a failed reload is already pretty severe.

rdrozhdzh:
Yes, the problem I mentioned (failed to start new instance) is rather hypothetical, we first need to fix #1688 since it affects lot of people

miekg:
Hmmm, failing to reload is rather odd (from the logging):
~~~
.com.:1043
.:1043
2018/04/20 15:29:00 [INFO] CoreDNS-1.1.1
2018/04/20 15:29:00 [INFO] linux/amd64, go1.10, 21bb5dcb-dirty
CoreDNS-1.1.1
linux/amd64, go1.10, 21bb5dcb-dirty
2018/04/20 15:29:00 [INFO] Running configuration MD5 = b34a96d99e01db4015a892212560155f
2018/04/20 15:29:36 [INFO] Reloading
2018/04/20 15:29:36 [ERROR] Failed to start metrics handler: listen tcp :443: bind: permission denied
2018/04/20 15:29:36 [INFO] Running configuration MD5 = 45fdd9451627040d94a86e08a6adaeed
2018/04/20 15:29:36 [INFO] Reloading complete
~~~
So it reloaded?

miekg:
I think #1668 is adding too much complexity, we we may need at some point; right now this (not 100% solution - I agree) is at least helping in the happy path...

fturib:
The point is that the order of event is not very convenient in CoreDNS use case.
We would need that the new Caddyfile is validated BEFORE stopping in any way the former server.
But right now : "onRestart" is called before any check on the new Corefile.

### 1- Allow listeners to not be reloaded

So if the new Corefile has a syntax error, then HELATH and PROMETHEUS will not work anymore until a valid Corefile is reloaded.
Is that good enough ? 

### 2- keep listener with address defined at first load (reload is kind of ignored for the HTTP service)

Another shortcut would be to adapt the HEALTH/PLUGIN to not reload the listeners .. and keep the initial Listening configuration. (change of code would be only in the Plugins).

### 3 - Reuse a first proposition that mostly works but does not handle some corner cases

There is a PR that can fix that issue : #1621, it was the first one I proposed few weeks ago. It is much simpler than #1688, but has some drawback if you reuse the ports from one plugin to another.
let me know if you think that could work as temporary before the #1688 is fully validated.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] reload: use O..." ]
>The point is that the order of event is not very convenient in CoreDNS use case.
>We would need that the new Caddyfile is validated BEFORE stopping in any way the former server.
>But right now : "onRestart" is called before any check on the new Corefile.

Yes, I'm aware of the trade off that have to made here. And this isn't a 100% 
solution either. But consider this:

* the failure modes are clear; reload work, but you may loose your listener. 
  This is really bad, but then again you were asking coredns for the impossible
* The code is supershort
* It solves all *initial* bug reports

miekg:
just a though: but with the reload plugin we should be able to test this actually.

plugin/health/health.go
```diff
@@ -61,6 +61,17 @@ func (h *health) OnStartup() error {
...
+			return err
+		}
+		h.ln = nil
+	}
+	return nil
+}
+
 func (h *health) OnShutdown() error {
 	// Stop polling plugins
 	h.pollstop <- true
```
fturib: The lameduck part should happen on RESTART ?
I would think only on FINAL shutdown.

That means maybe having a separated function to close Listener and to process the Lameduck.

OnRestart -> close listener
OnFinalShutdown -> close listener + lameduck delay

plugin/metrics/metrics.go
```diff
@@ -104,6 +104,18 @@ func (m *Metrics) OnStartup() error {
...
+	// relingish our listener as we re-listen on successfull reload
+	if m.ln != nil {
+		if err := m.ln.Close(); err != nil {
+			return err
+		}
+		m.ln = nil
+	}
+	return nil
+}
+
```
fturib: So.. the OnRestart is the same (a little proper) than the OnShutdown .. no ? Maybe we can have only one to be called for the 2 events:
- onRestart
- onFinalShutdown
I would not report the error to Caddy as it will prevent to continue the RESTART or SHUTDOWN. Anyway if the listener cannot be closed it means it is already closed. Maybe INFO the  error.

NOTE: onShutdown works too .. but is confusing for the reader.

miekg:
https://github.com/coredns/coredns/pull/1710 has a test, I'll rework stuff during the weekend (unless your branch from the PR - fine with me) and get into a state where we can merge (also needs some doc updates in the reload plugin and else where).

plugin/reload/README.md
```diff
@@ -59,3 +60,27 @@ Check every 10 seconds (jitter is automatically set to 10 / 2 = 5 in this case):
...
+~~~
+
+CoreDNS starts and serves health from :8080. Now you change `:8080` to `:443` not knowing a process
+is already listening on that port. The process reloads, and performs the following steps:
+
+1. close the listener on 8080
+2. reload and parse the config again
+3. fail to start a new listener on 443
+
+Now failing step 3 *does not cause* the reload to fail, instead almost everything works, except the
```
fturib: Are you sure ?
I think that if the Listener fails to start on 443, then OnStartup will return an error, that will be returned back to Caddy .. and the Start of the new server will just fail, leading to the fail of Restart().

The old server will keep running, but the Listener will be closed on :8080

plugin/reload/README.md
```diff
@@ -59,3 +60,27 @@ Check every 10 seconds (jitter is automatically set to 10 / 2 = 5 in this case):
...
+health exporting because there is no new listener.
+
+Note the same is true for prometheus metrics plugin.
```
fturib: Right now, plugin METRICS is a strange animal : if you shutdown a listener, it will never restart on the same address ! (because the started listeners are saved in a global var map).
=> it is maybe better to close only on "OnFinalShutdown" and do not consider changing the address.

miekg: The reason that was done was to support both:
~~~
.com {
   prometheus
}
.net {
   prometheus
}
~~~

and

.com {
   prometheus :1200
}
.net {
   prometheus :8080
}
~~~
Maybe we should just give up no that or find a better way.

miekg:
Explicitly not support in the health middleware
~~~
If you have multiple Server Blocks and need to export health for each of the plugins, you must run
health endpoints on different ports:
~~~
For the prometheus plugin that global uniqAddr is uses precisely for this reason.

plugin/reload/README.md
miekg: changed

miekg:
Ok, this ballooned into way more LOCs then expected :-(
Trying to trim it a little now the tests pass.

PTAL

plugin/metrics/setup.go
```diff
@@ -15,7 +15,7 @@ func init() {
 		Action:     setup,
 	})
 
-	uniqAddr = addrs{a: make(map[string]int)}
+	uniqAddr = newAddress()
```
fturib, line 18: I think for Metrics the issue is here : Addresses is a global var that register addrss of listeners for metrics. 
- at Setup : new addrs goes as "todo"
- as Startup : this address is processed - listener is created and "todo" goes into "done"
- at Restart : the listener is closed
- at Setup again (restart)... the addrs is already "done" - no change
- as Startup : nothing happen as the addrs is "done- 

**AFTER RESTART if address does not change, the listener of prometheus is just not listening**

fturib: A short term easy mitigation : do not shutdown the listener at "OnRestart", but only at "OnFinalShutdown".

=> premetheus will not be very compatible with reload, but will not prevent it.

fturib (commented):
I think should work for HEALTH.
but for METRICS, see my comment : I guess you can reload, but I think you cannot collect data on the listener after RESTART (see my comment)

plugin/metrics/setup.go
fturib: Did you make a try, running Coredns with health, and Prometheus, reloading and verifying you can still collect HEALTH and METRICS on the correct IP ?

miekg: OK, done the following test - I can prolly actually make a Go test for this as well. Anway:

Corefile:
~~~
. {
    errors
    log
    prometheus :35053
    whoami
    health :35080
}
~~~
Start `./coredns -dns.port=1043`; query it for metrics to be published. Then
~~~
% curl http://localhost:35080/health
OK%
~~~
and
~~~
% curl -s http://localhost:35053/metrics|head -3
# HELP coredns_build_info A metric with a constant '1' value labeled by version, revision, and goversion from which CoreDNS was built.
# TYPE coredns_build_info gauge
coredns_build_info{goversion="go1.10.1",revision="",version="1.1.1"} 1
~~~
restart:
~~~
% kill -SIGUSR1 $(pidof coredns) 
2018/04/21 16:18:15 [INFO] SIGUSR1: Reloading
2018/04/21 16:18:15 [INFO] Reloading
2018/04/21 16:18:15 [INFO] Reloading complete
~~~
metrics indeed *dead*
health works: 

~~~
% curl -s http://localhost:35080/health
OK% 
~~~

Nice find!

miekg: Trying another fix that sets the m.Addr to todo again.

miekg:
travis is extremely flacky for me on testing this pr :/

miekg:
Ok, squashed some commits to kick off another travis check; seems OK now.

not super perfect; but good enough IMO. @fturib PTAL

fturib:
Great Miek ! I think you are good with the tests. 
Code looks good for me.

The only drawbacks we have for now, and can be addressed later:
Health or Metrics listeners may not be operational after a failed RESTART, until the next successful RESTART

miekg:
yes, I tried to document that in plugin/reload/README.md in the bugs section or you pointing to *another* issue? (but even so, file another issues :) )

fturib (approved):
LGTM.

PR #1714, opened 2018-04-21
plugin/forward: fix TLS setup

miekg:
Way smaller PR than #1679. Fixes same thing.

### DIFF
--- a/plugin/forward/proxy_test.go
+++ b/plugin/forward/proxy_test.go
@@ -1,65 +1,68 @@
 package forward
 
 import (
-	"context"
-	"sync"
 	"testing"
 
 	"github.com/coredns/coredns/plugin/pkg/dnstest"
 	"github.com/coredns/coredns/plugin/test"
-	"github.com/coredns/coredns/request"
 
+	"github.com/mholt/caddy"
 	"github.com/miekg/dns"
+	"golang.org/x/net/context"
 )
 
-func TestProxyClose(t *testing.T) {
+func TestProxy(t *testing.T) {
 	s := dnstest.NewServer(func(w dns.ResponseWriter, r *dns.Msg) {
 		ret := new(dns.Msg)
 		ret.SetReply(r)
+		ret.Answer = append(ret.Answer, test.A("example.org. IN A 127.0.0.1"))
 		w.WriteMsg(ret)
 	})
 	defer s.Close()
 
-	req := new(dns.Msg)
-	req.SetQuestion("example.org.", dns.TypeA)
-	state := request.Request{W: &test.ResponseWriter{}, Req: req}
-	ctx := context.TODO()
+	c := caddy.NewTestController("dns", "forward . "+s.Addr)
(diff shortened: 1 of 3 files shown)


rdrozhdzh:
Hm, it seems some change is needed in setup code which sets default tls config in case if no tls option in corefile. Something like I mentioned in https://github.com/coredns/coredns/pull/1679#issuecomment-382450510

plugin/forward/proxy_test.go
```diff
@@ -1,15 +1,16 @@
...
 	"sync"
 	"testing"
 
 	"github.com/coredns/coredns/plugin/pkg/dnstest"
 	"github.com/coredns/coredns/plugin/test"
 	"github.com/coredns/coredns/request"
 
+	"github.com/mholt/caddy"
 	"github.com/miekg/dns"
+	"golang.org/x/net/context"
```
rdrozhdzh: why not "context"?

miekg: done

miekg:
> Hm, it seems some change is needed in setup code which sets default tls config in case if no tls option in corefile. Something like I mentioned in #1679 (comment)

Don't fully understand, why? because the TLS is only set in the underlaying elements if its needed, in setup.go:
~~~
if protocols[i] == TLS {
	f.proxies[i].SetTLSConfig(f.tlsConfig
}
~~~

rdrozhdzh:
lets consider the following forward directive
```
forward tls://127.0.0.1:443 {
    tls
}
```
in this case all should work fine since option **tls** without parameters will create a tls config with system's CA cert. But if we use the forward directive without **tls** option like below
```
forward tls://127.0.0.1:443
```
the `f.tlsConfig` will remain equal to `nil` and no proxy will be considered as TLS. So, the code I proposed will create a default tls config in case if no **tls** option was specified in forward directive

I thought, this was your main idea in #1672, i.e.
> by default you just want to use the system's certs meaning you don't have the tls directive in the config

This is my understanding, please let me know if I'm missing anything..

miekg:
>lets consider the following forward directive
>```
>forward tls://127.0.0.1:443 {
>    tls
>}
>```
>in this case all should work fine since option **tls** without parameters will create a tls config with system's CA cert. But if we use the forward directive without **tls** option like below
>```
>forward tls://127.0.0.1:443
>```
>the `f.tlsConfig` will remain equal to `nil` and no proxy will be considered as TLS. So, the code I proposed will create a default tls config in case if no **tls** option was specified in forward directive

No (but not obvious): New() returns a forward with a tlsConfig by default:

   f := &Forward{maxfails: 2, tlsConfig: new(tls.Config), expire: defaultExpire, p: new(random), from: ".", hcInterval: hcDuration}

this was changed in the original PR, but now I went for "keep PR small", so I 
kept that in.

rdrozhdzh:
Oh, I see it now, thanks for clarification, LGTM

rdrozhdzh (approved):

PR #1715, opened 2018-04-21
Export metrics in setup; so it also works after reload

miekg:


miekg:
moves some code. Merging

PR #1718, opened 2018-04-22
Use logging

miekg:


miekg:
only thing that is strange is that secondary logs under plugin/file because of how this is setup code wise.

PR #1719, opened 2018-04-23
Presubmit

miekg:
Add presubmits to the build, i.e. little scripts that run first and catch obvious things.

### DIFF
--- a/.presubmit/context
+++ b/.presubmit/context
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+echo "** presubmit/$(basename $0)"
+
+if grep -lr "golang.org/x/net/context" "$@"; then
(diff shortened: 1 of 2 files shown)


miekg:
adds example script the checks for the right context to be used.

yongtang (approved):
Tested my self with the output looks good:
```
$ make check
** presubmit/context
plugin/forward/connect.go
** presubmit/context: please use std lib's 'context' instead
go get -u github.com/alecthomas/gometalinter
gometalinter --install golint
Installing:
```

LGTM.

yongtang:
Actually, it is not failing the build yet, only print a warning. Should the command return a error code to fail?

miekg:
yep, wanted to error; le'me check

yongtang (approved):
LGTM. Thanks!

PR #1726, opened 2018-04-24
plugin/forward: remove lenc and lencOut channels

miekg:
Remove these 2 channels, as they were only used in testing and
complicate things if we want to do a clean stop()

Less is more.

### DIFF
--- a/plugin/forward/persistent_test.go
+++ b/plugin/forward/persistent_test.go
@@ -30,28 +30,4 @@ func TestPersistent(t *testing.T) {
 	tr.Yield(c1)
 	tr.Yield(c2)
 	tr.Yield(c3)
-
-	if x := tr.Len(); x != 3 {
-		t.Errorf("Expected cache size to be 3, got %d", x)
-	}
-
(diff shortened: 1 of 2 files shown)


plugin/forward/persistent_test.go
```diff
@@ -30,28 +30,4 @@ func TestPersistent(t *testing.T) {
 	tr.Yield(c1)
 	tr.Yield(c2)
 	tr.Yield(c3)
-
-	if x := tr.Len(); x != 3 {
```
rdrozhdzh, line 34: can't we use tr.len() (non-threadsafe version of Len) in this test?

miekg:
>
>
>
>> @@ -30,28 +30,4 @@ func TestPersistent(t *testing.T) {
> 	tr.Yield(c1)
> 	tr.Yield(c2)
> 	tr.Yield(c3)
>-
>-	if x := tr.Len(); x != 3 {
>
>can't we use tr.len() (non-threadsafe version of Len) in this test?

I thought that would then trigger the -race detector

PR #1728, opened 2018-04-24
core: remove HostAddresses()

miekg:
config.HostAddresses() is a weird function that gathers
some data from the server and returns a string.

It is *only* used the trace plugin, to figure out what
server starts the trace.

Looks to be better to fit in the with metrics.WithServer label
on the trace itself to show which server handled the trace.

Remove HostAddresses() and cleanup trace a small bit

### DIFF
--- a/plugin/trace/setup.go
+++ b/plugin/trace/setup.go
@@ -41,7 +41,8 @@ func traceParse(c *caddy.Controller) (*trace, error) {
 	)
 
 	cfg := dnsserver.GetConfig(c)
-	tr.ServiceEndpoint = cfg.HostAddresses()
+	tr.serviceEndpoint = cfg.ListenHosts[0] + ":" + cfg.Port
+
 	for c.Next() { // trace
 		var err error
 		args := c.RemainingArgs()
--- a/core/dnsserver/config.go
+++ b/core/dnsserver/config.go
@@ -3,7 +3,6 @@ package dnsserver
 import (
 	"crypto/tls"
 	"fmt"
-	"net"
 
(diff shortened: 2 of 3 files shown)


johnbelamaric (approved):
LGTM

plugin/trace/setup.go
```diff
@@ -41,7 +41,8 @@ func traceParse(c *caddy.Controller) (*trace, error) {
 	)
 
 	cfg := dnsserver.GetConfig(c)
-	tr.ServiceEndpoint = cfg.HostAddresses()
+	tr.serviceEndpoint = cfg.ListenHosts[0] + ":" + cfg.Port
+
```
fturib, line 45: Here you will have incomplete information if BIND is used with several IP.
However I guess that you will not be able to reuse that IP on another ServerBloc .. so it is still an identifier.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] core: remove ..." ]
>
>
>
>> @@ -41,7 +41,8 @@ func traceParse(c *caddy.Controller) (*trace, error) {
> 	)
>
> 	cfg := dnsserver.GetConfig(c)
>-	tr.ServiceEndpoint = cfg.HostAddresses()
>+	tr.serviceEndpoint = cfg.ListenHosts[0] + ":" + cfg.Port
>+
>
>Here you will have incomplete information if BIND is used with several IP.
>However I guess that you will not be able to reuse that IP on another ServerBloc .. so it is still an identifier.

Yeah, and note I changes the Trace to include the identifer from the context 
which should give you enough info.

PR #1730, opened 2018-04-25
plugin/autopath: per server metrics

miekg:
Implement per server metrics in autopath; also don't export the metric.

Updated readme as well.


chrisohaver (approved):
lgtm

PR #1731, opened 2018-04-25
plugin/template: update metrics to include server

miekg:
Add server label to the metrics and update the README.


miekg:
going in.

PR #1733, opened 2018-04-25
plugin/pkg/uniq: add

miekg:
Spin this out the metrics package so we can use it in the health
one of well to fix some reload bugs.


miekg:
Moves some code around; (hopefully) improved naming by moving in to pkg/uniq.

PR #1734, opened 2018-04-25
plugin/forward: make stop stop

miekg:
make sure stop doesn't hang, but always returns

### DIFF
--- a/plugin/forward/persistent.go
+++ b/plugin/forward/persistent.go
@@ -44,14 +44,7 @@ func newTransport(addr string, tlsConfig *tls.Config) *transport {
 		ret:    make(chan connErr),
 		stop:   make(chan bool),
 	}
-	go func() {
-		t.connManager()
-		// if connManager returns it has been stopped.
-		close(t.stop)
-		close(t.yield)
-		close(t.dial)
-		// close(t.ret) // we can still be dialing and wanting to send back the socket on t.ret
-	}()
+	go func() { t.connManager() }()
 	return t
 }
 
@@ -120,6 +113,7 @@ Wait:
 			t.conns["tcp-tls"] = append(t.conns["tcp-tls"], &persistConn{conn.c, time.Now()})
 
 		case <-t.stop:
+
 			return
 		}
 	}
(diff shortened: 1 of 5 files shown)


miekg:
this *does* revert the previous graceful stop PR.

miekg:
fails on 
~~~=== RUN   TestReadme
--- FAIL: TestReadme (6.05s)
	readme_test.go:75: Failed to start server with ../plugin/erratic/README.md, for input "listen udp :30066: bind: address already in use":
		. {
		    erratic
		}
~~~

miekg:
travis... tssk:
~~~
go get -u golang.org/x/net/context
package golang.org/x/net/context: unrecognized import path "golang.org/x/net/context" (fetch https://golang.org/x/net?go-get=1: no go-import meta tag)
make: *** [godeps] Error 1
~~~

plugin/forward/persistent.go
```diff
@@ -132,18 +126,33 @@ func (t *transport) Dial(proto string) (*dns.Conn, bool, error) {
 		proto = "tcp-tls"
 	}
 
-	t.dial <- proto
-	c := <-t.ret
-	return c.c, c.cached, c.err
+	select {
+	case t.dial <- proto:
+	case <-t.stop:
```
rdrozhdzh, line 131: This way close() will unblock all pending writes to channels, but (I guess) it will not prevent upcoming writes which will hang

miekg: yes, this is not nice to the caller, but it will force a clean shutdown; so there's that. Should probably double check that this gives servfail back to the caller so it may retry (think it does)

miekg:
doing 100 test runs, seems it can still occur:
~~~
% for i in {0..100}; do go test -v 2>&1; done |grep '[0-9]s$'
ok  	github.com/coredns/coredns/plugin/forward	601.010s
~~~
WTAF

rdrozhdzh:
that's what I expected, see my previous comment

miekg:
indeed, this is still race:
~~~
	case t.dial <- proto:
	case <-t.stop:
		return nil, false, errStopped
	}
~~~

miekg:
~~We could make them all a 1-buffered channel, but yuck~~
test catches this at once (if you close the channels on return from connmanager)

miekg:
actually while looking through the current code (master branch); that looks pretty sane; so I'm still left wondering

miekg:
ok, I'm pulling out the dial from the connManager, that a) starts another goroutine b) can be lingering around for a long time

Just putting it in dial is relatively straight forward.

PR #1738, opened 2018-04-25
plugin/forward: Fix the close test and other things

miekg:
See commit mesg.

### DIFF
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -34,16 +34,6 @@ func (p *Proxy) updateRtt(newRtt time.Duration) {
 }
 
 func (p *Proxy) connect(ctx context.Context, state request.Request, forceTCP, metric bool) (*dns.Msg, error) {
-	atomic.AddInt32(&p.inProgress, 1)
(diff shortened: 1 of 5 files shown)


miekg:
I saw scary things yesterday when I used the new test ProxyClose on master, but I can't repeat those today; I'm merging this. Then we need to redo the graceful shutdown, possibly by just using waitgroups as proxy does.

rdrozhdzh:
@miekg , did you save those "scary" logs, or maybe they are still accessible in travis? I would like to look at them

miekg:
>@miekg , did you save those "scary" logs, or maybe they are still accessible in travis? I would like to look at them

no :-( Trying to look back in my terminal history, but there are soo many data 
races and close on send, that I can tie this to a particular code base...

I think the test was wrong (to begin with) as the closing the proxy happens from 
the same routine that started it.

And as said in the PR (I think), a sync.WaitGroup approach might be better (and 
mimic proxy which would also make the code base more similar)

/Miek

--
Miek Gieben

PR #1747, opened 2018-04-27
plugin/forward: clarify relation with proxy

miekg:
some other small bits as well.

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -21,6 +21,10 @@ connect to a random upstream (which may or may not work).
 
 This plugin can only be used once per Server Block.
 
+How does *forward* relate to *proxy*? This plugin is the "new" version of *proxy* and is faster
+because it re-uses connections to the upstreams. It also does in-band health checks - using DNS
+instead of HTTP. Since it is newer it has a little less (production) millage on it.
+
 ## Syntax
 
 In its most basic form, a simple forwarder uses this syntax:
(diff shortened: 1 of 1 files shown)


plugin/forward/README.md
```diff
@@ -21,6 +21,10 @@ connect to a random upstream (which may or may not work).
 
 This plugin can only be used once per Server Block.
 
+How does *forward* relate to *proxy*? This plugin is the "new" version of *proxy* and is faster
+because it re-uses connections to the upstreams. It also does in-band health checks - using DNS
+instead of HTTP. Since it is newer it has a little less (production) millage on it.
```
johnbelamaric, line 26: s/millage/milage/

chrisohaver: https://trends.google.com/trends/explore?q=milage,mileage

yuewko: Very persuasive!

fturib:
@miekg:
About Proxy vs Forward, I noticed that:
- proxy support gRPC protocol but NOT TLS
- forward support TLS but not gRPC

is that correct ?

miekg:
>@miekg:
>About Proxy vs Forward, I noticed that:
>- proxy support gRPC protocol but NOT TLS
>- forward support TLS but not gRPC
>
>is that correct ?

Yep. allthough proxy tls support should not be that hard, grpc in forward is 
harder.

PR #1749, opened 2018-04-27
plugin/etcd: small refactor

miekg:
I think this function can be smaller.


miekg:
@m4r10k can you double check my logic here; got a feeling the function could be shorter

m4r10k:
Yes, I will have a look at this during weekend! Your approach is nicer! I have to test it against the problem, that the Etcd path might not exist (thats why my code looked ugly) and I am currently not at home (family). :)

m4r10k:
@miekg Wow, your code is **much** clearer and it works pretty well! LGTM

PR #1754, opened 2018-04-28
plugin/forward: implement two phased stop

miekg:
First signal that we stop; this inhbits new dial into the connmanager.
Then sleep (*shrudder*) for a bit and close and return from the
connmanager.

It also cleans up the accumulated connections on stop.

### DIFF
--- a/plugin/forward/persistent.go
+++ b/plugin/forward/persistent.go
@@ -152,7 +152,7 @@ func (t *transport) Yield(c *dns.Conn) { t.yield <- c }
 // Stop stops the transport's connection manager.
 func (t *transport) Stop() {
 	t.signal <- true
-	time.Sleep(500 * time.Millisecond)
+	time.Sleep(100 * time.Millisecond)
 	close(t.stop)
 }
 
--- a/plugin/forward/persistent.go
+++ b/plugin/forward/persistent.go
@@ -3,6 +3,7 @@ package forward
 import (
 	"crypto/tls"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/miekg/dns"
@@ -16,15 +17,18 @@ type persistConn struct {
 
 // transport hold the persistent cache.
 type transport struct {
-	conns     map[string][]*persistConn //  Buckets for udp, tcp and tcp-tls.
+	conns     map[string][]*persistConn // Buckets for udp, tcp and tcp-tls.
 	expire    time.Duration             // After this duration a connection is expired.
 	addr      string
 	tlsConfig *tls.Config
 
-	dial  chan string
-	yield chan *dns.Conn
-	ret   chan *dns.Conn
-	stop  chan bool
+	dial   chan string
+	yield  chan *dns.Conn
+	ret    chan *dns.Conn
+	stop   chan bool
+	signal chan bool
+	sync.RWMutex
+	stopping bool
 }
 
 func newTransport(addr string, tlsConfig *tls.Config) *transport {
@@ -36,6 +40,7 @@ func newTransport(addr string, tlsConfig *tls.Config) *transport {
 		yield:  make(chan *dns.Conn),
 		ret:    make(chan *dns.Conn),
 		stop:   make(chan bool),
+		signal: make(chan bool),
 	}
 	go func() { t.connManager() }()
 	return t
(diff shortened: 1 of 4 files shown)


miekg:
@rdrozhdzh I think this is similar to what you where trying to do; still impairs performance a bit (due to the mutex), but *at least* it looks to be doing the correct thing without "send on closed channel" or data races.

miekg:
I this can be merged I look at lookup.go which needs things from forward.go ported over.

miekg:
I'm failing to come up with something smart than `sleep`.

We could range over t.dial to slurp remaining entries, but that requires closing it in the first place

miekg:
test failed: 
~~~
=== RUN   TestMetricsAuto
--- FAIL: TestMetricsAuto (2.26s)
	metrics_test.go:171: Expected value 1 for coredns_dns_request_count_total, but got 
~~~

plugin/forward/persistent.go
```diff
@@ -129,7 +150,18 @@ func (t *transport) Dial(proto string) (*dns.Conn, bool, error) {
 func (t *transport) Yield(c *dns.Conn) { t.yield <- c }
 
 // Stop stops the transport's connection manager.
-func (t *transport) Stop() { close(t.stop) }
+func (t *transport) Stop() {
+	t.signal <- true
+	time.Sleep(100 * time.Millisecond)
+	close(t.stop)
```
rdrozhdzh, line 156: Lets assume `Stop()` is called when we are in the middle of `Dial()`. After successful `Dial()` we normally send the query, read response and yield the connection back to cache. Reading response may take more time than 100ms, so connManager() may already get stopped and `Yield()` will hang.
Perhaps, the code `if t.Signal() {` should be added to `Yield()` as you did it for `Dial()`

plugin/forward/persistent.go
```diff
@@ -96,15 +100,32 @@ Wait:
 
 			t.conns["tcp-tls"] = append(t.conns["tcp-tls"], &persistConn{conn, time.Now()})
 
+		case <-t.signal:
+			t.Lock()
+			t.stopping = true
+			t.Unlock()
```
rdrozhdzh, line 106: consider using atomic instead of mutex

plugin/forward/persistent.go
```diff
@@ -96,15 +100,32 @@ Wait:
 
 			t.conns["tcp-tls"] = append(t.conns["tcp-tls"], &persistConn{conn, time.Now()})
 
+		case <-t.signal:
```
rdrozhdzh, line 103: consider reusing `stop` channel for both setting flag and actual closing connManager. The code which read from channel can distinguish between **read** and **close** operations by second boolean parameter, i.e.
```
case _, ok <-t.stop:
    if ok {
        // read from channel
        t.stopping = true
        continue
    }
    // channel was closed
    // close connection manager
    return
```

plugin/forward/proxy_test.go
```diff
@@ -25,10 +25,12 @@ func TestProxyClose(t *testing.T) {
 	state := request.Request{W: &test.ResponseWriter{}, Req: req}
 	ctx := context.TODO()
 
-	for i := 0; i < 100; i++ {
+	for i := 0; i < 10; i++ {
 		p := NewProxy(s.Addr, nil)
 		p.start(hcDuration)
 
+		go func() { p.connect(ctx, state, false, false) }()
+		go func() { p.connect(ctx, state, true, false) }()
```
rdrozhdzh, line 33: the drawback of the latest version of test is that it doesn't check if particular connect() hung or not. I believe this is the reason why you don't see failed tests anymore

miekg:

Thanks for you comments, I will address them, except - I think - the one on 
atomic. I benchmarked some of this and mutexes came out faster than atomic (at 
some point). So unless there is a efficiency reason for them, I rather just 
stick to a mutex.

rdrozhdzh:
> So unless there is a efficiency reason for them, I rather just stick to a mutex.

From what I can see in go library code, mutexes use atomics internally (in addition to other means), so I suppose using atomic directly would be faster. Also code with atomic would look more compact. That was my reasoning
Anyway, if you feel the mutex approach is more preferred here - I don't mind

Let me know if you need some help from my side

rdrozhdzh:
I just thought, couldn't we set finalizer (https://golang.org/pkg/runtime/#SetFinalizer) for Proxy object to cleanup connections and stop connManager() goroutine after Proxy object gets unreachable?

rdrozhdzh:
see #1768 with finalizer

PR #1760, opened 2018-05-01
Do Compress only when need in request.Scrub

miekg:


johnbelamaric:
What's the context here? Trying to reduce CPU? This looks fine, assuming you grepped and got all the instances where it was forcing Compress. Do you have any benchmarks?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Do Compress o..." ]
>What's the context here? Trying to reduce CPU? This looks fine, assuming you grepped and got all the instances where it was forcing Compress. Do you have any benchmarks?

Yep, see one of the commits has bench for cache.

johnbelamaric (approved):
Ok, LGTM!

PR #1768, opened 2018-05-04
plugin/forward: close connection manager in proxy finalizer

rdrozhdzh:
 - connManager() goroutine will stop when Proxy is about to be
   garbage collected. This means that no queries are in progress,
   and no queries are going to come

#1666

none

### DIFF
--- a/plugin/forward/proxy.go
+++ b/plugin/forward/proxy.go
@@ -2,6 +2,7 @@ package forward
 
 import (
 	"crypto/tls"
+	"runtime"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


rdrozhdzh:
this is yet another approach to fix #1666. 

remaining problems to be addressed:
- cache cleanup. No guarantee that finalizer will be called soon, or be called at all. We need to close and remove expired connections from cache as soon as possible. In current design, we check connections only when next request comes. If we have high load to some upstream server, say we have 1000 or more connections in cache, and at some moment the upstream gets unhealthy, forward will stop sending queries to this upstream. However, all expired connections in cache will still use system resources for an unpredictably long time
I propose to add timer in connManager() and do connection cleanup periodically with no matter come new queries from clients or not.
- Probe doesn't expect calling Do() after Stop() - it will hang. However, this is a possible case in current forward implementation.
I propose to simplify Probe and remove channels at all. The existing mutex would be enough to synchronize concurrent calls. I would even replace the mutex with atomic calls...  ;)

miekg (approved):

PR #1774, opened 2018-05-05
presubmit: check for uppercase

miekg:
Another thing we can test automatically, we sorta settled on using an
uppercase letter in in t.Log and t.Fatal calls.

Let's just check for this.


yongtang (approved):
LGTM

PR #1777, opened 2018-05-05
plugin/errors: remove panic/recover

### LINKED ISSUE #1776: plugin/error: remove `recover`
This came from caddy's error plugin, can't actually see why we need as we do this in server.go as well. Also make enabled `debug` *actually* disable the panic/recover thing.

miekg:
Remove panic/recover and also use pkg/log to print the error. This
brings some consistency into the logging.

Fixes #1776

### DIFF
--- a/plugin/errors/errors.go
+++ b/plugin/errors/errors.go
@@ -3,77 +3,27 @@ package errors
 
 import (
 	"context"
-	"fmt"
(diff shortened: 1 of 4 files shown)


yongtang (approved):
LGTM, Thanks!

PR #1778, opened 2018-05-05
plugin/metrics: add panic counter

### LINKED ISSUE #1294: add metrics tracking panics
We should add a metric tracking these. At least plugin/error needs to do this as well.

miekg:
Count and export number of panics we see.

Fixes #1294

### DIFF
--- a/plugin/metrics/vars/vars.go
+++ b/plugin/metrics/vars/vars.go
@@ -59,6 +59,12 @@ var (
 		Name:      "response_rcode_count_total",
 		Help:      "Counter of response status codes.",
 	}, []string{"server", "zone", "rcode"})
+
+	Panic = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: plugin.Namespace,
+		Name:      "panic_count_total",
+		Help:      "A metrics that counts the number of panics.",
+	})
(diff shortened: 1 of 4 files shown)


plugin/metrics/vars/vars.go
```diff
@@ -59,6 +59,12 @@ var (
 		Name:      "response_rcode_count_total",
 		Help:      "Counter of response status codes.",
 	}, []string{"server", "zone", "rcode"})
+
+	Panic = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: plugin.Namespace,
+		Name:      "panic_count_total",
+		Help:      "A metrics that counts the number of panics.",
```
bradbeam, line 66: Are there any labels we can add in for the panic to help identify the cause? Not sure if we could identify a plugin/pkg at the point of panic to add in as a label.

bradbeam (approved):

plugin/metrics/vars/vars.go
miekg: not sure, the error plugin had a whole bunch of stuff to show this, but I'm sorta hoping the stack trace lands in your logging frame work...

Issue #1762, opened 2018-05-02
Plugin/Forward - ability to access the Proxy from external plugin


fturib:
### Enhancement

For an external plugin, we need to get access to functions of the FORWARD package that are currently private:

- forward.list() - list all the upstream connection defined 
- proxy.connect(...) - send a DNS msg to the upstream connection

I think it is safe to make this 2 function public.

@ekleiner is currently validating before sending a PR for this enhancement.

### Context

We developed an external plugin dedicated for our use case. It is monitoring the upstream connection of a PROXY plugin.
We would like to move that monitoring over an upstream connection of a FORWARD plugin.

our plugin is a kind of superset of the "health check", but we have very specific options tied to our use case 
- DNS query is kind of dedicated
- We need to handle reporting and event when the connection if found down or not healthy.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] Plugin/Forward - ..." ]
>### Enhancement
>
>For an external plugin, we need to get access to functions of the FORWARD package that are currently private:
>
>- forward.list() - list all the upstream connection defined
>- proxy.connect(...) - send a DNS msg to the upstream connection
>
>I think it is safe to make this 2 function public.
>
>@ekleiner is currently validating before sending a PR for this enhancement.

OK, sgtm

PR #1781, opened 2018-05-07
plugin/forward: expose TLSConfig and error messages to public

eugen-kleiner-mesh:
* Adds public method IsTLS()
* Makes error messages public
https://github.com/coredns/coredns/issues/1762

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -104,7 +104,7 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 		)
 		for {
 			ret, err = proxy.Connect(ctx, state, f.forceTCP, true)
-			if err != nil && err == errCachedClosed { // Remote side closed conn, can only happen with TCP.
+			if err != nil && err == ErrCachedClosed { // Remote side closed conn, can only happen with TCP.
 				continue
 			}
 			break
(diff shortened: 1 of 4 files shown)


plugin/forward/proxy.go
```diff
@@ -63,6 +63,9 @@ func (p *Proxy) SetTLSConfig(cfg *tls.Config) {
 	p.client = dnsClient(cfg)
 }
 
+// TLSConfig provides current TLS config.
+func (p *Proxy) TLSConfig() *tls.Config { return p.transport.tlsConfig }
+
```
fturib: is it safer to return a copy  instead of a pointer ? 
I mean : can someone modify the tls.Config using this function ?

plugin/forward/forward.go
```diff
@@ -186,10 +186,12 @@ func (f *Forward) ForceTCP() bool { return f.forceTCP }
 func (f *Forward) List() []*Proxy { return f.p.List(f.proxies) }
 
 var (
-	errInvalidDomain = errors.New("invalid domain for forward")
-	errNoHealthy     = errors.New("no healthy proxies")
```
fturib, line 190: You are removing 4 and adding only 3.
I guess the errInvalidDomain is not used ...

eugen-kleiner-mesh: Yes. It is not used, so removed.

plugin/forward/proxy.go
eugen-kleiner-mesh: Yes. As it is pointer and it is possible to modify it. If we want to make it safe then deepcopy is required here. Also, as it may be nil, I will return copy with bool "ok" variable.

miekg (commented):
Why do you need access to the tls config?

eugen-kleiner-mesh:
In fact I just need to determine if it is TLS or not.

rdrozhdzh:
seems, you can just  add `IsTLS()` method to Proxy, instead of `TLSConfig()`

miekg (approved):

PR #1782, opened 2018-05-07
plugin/forward: erase expired connections by timer

rdrozhdzh:
 - in previous implementation, the expired connections resided in
   cache until new request to the same upstream/protocol came. In
   case if the upstream was unhealthy new request may come long time
   later or may not come at all. All this time expired connections
   held system resources (file descriptors, ephemeral ports). In my
   fix the expired connections and related resources are released
   by timer
 - decreased the complexity of taking connection from cache. The list
   of connections is treated as stack (LIFO queue), i.e. the connection
   is taken from the end of queue (the most fresh connection) and
   returned to the end (as it was implemented before). The remarkable
   thing is that all connections in the stack appear to be ordered by
   'used' field
 - the cleanup() method finds the first good (not expired) connection
   in stack with binary search, since all connections are ordered by
   'used' field

#1768

### DIFF
--- a/plugin/forward/persistent.go
+++ b/plugin/forward/persistent.go
@@ -3,6 +3,7 @@ package forward
 import (
 	"crypto/tls"
 	"net"
+	"sort"
 	"time"
 
 	"github.com/miekg/dns"
@@ -53,28 +54,24 @@ func (t *transport) len() int {
 
 // connManagers manages the persistent connection cache for UDP and TCP.
 func (t *transport) connManager() {
-
+	timer := time.NewTimer(t.expire)
 Wait:
 	for {
 		select {
 		case proto := <-t.dial:
-			// Yes O(n), shouldn't put millions in here. We walk all connection until we find the first
-			// one that is usuable.
-			i := 0
-			for i = 0; i < len(t.conns[proto]); i++ {
-				pc := t.conns[proto][i]
+			// take the last used conn - complexity O(1)
+			if stack := t.conns[proto]; len(stack) > 0 {
+				pc := stack[len(stack)-1]
 				if time.Since(pc.used) < t.expire {
 					// Found one, remove from pool and return this conn.
-					t.conns[proto] = t.conns[proto][i+1:]
+					t.conns[proto] = stack[:len(stack)-1]
 					t.ret <- pc.c
 					continue Wait
 				}
-				// This conn has expired. Close it.
-				pc.c.Close()
+				// clear entire cache if the last conn is expired
+				go closeConns(stack)
+				t.conns[proto] = nil
 			}
-
-			// Not conns were found. Connect to the upstream to create one.
-			t.conns[proto] = t.conns[proto][i:]
 			SocketGauge.WithLabelValues(t.addr).Set(float64(t.len()))
 
 			t.ret <- nil
@@ -96,10 +93,47 @@ Wait:
 
 			t.conns["tcp-tls"] = append(t.conns["tcp-tls"], &persistConn{conn, time.Now()})
 
+		case <-timer.C:
+			t.cleanup(false)
+			timer.Reset(t.expire)
+
 		case <-t.stop:
+			t.cleanup(true)
 			close(t.ret)
 			return
 		}
+	}
+}
+
+// closeConns closes connections.
+func closeConns(conns []*persistConn) {
+	for _, pc := range conns {
+		pc.c.Close()
+	}
+}
+
+// cleanup removes connections from cache.
+func (t *transport) cleanup(all bool) {
+	staleTime := time.Now().Add(-t.expire)
+	for proto, stack := range t.conns {
+		if len(stack) == 0 {
+			continue
+		}
+		if all {
+			go closeConns(stack)
+			t.conns[proto] = nil
+			continue
+		}
+		if stack[0].used.After(staleTime) {
+			continue
+		}
+
+		// connections in stack are sorted by "used"
+		good := sort.Search(len(stack), func(i int) bool {
+			return stack[i].used.After(staleTime)
+		})
+		go closeConns(stack[:good])
+		t.conns[proto] = stack[good:]
 	}
 }
 
--- a/plugin/forward/persistent.go
+++ b/plugin/forward/persistent.go
@@ -69,6 +69,8 @@ Wait:
 				}
 				// clear entire cache if the last conn is expired
 				t.conns[proto] = nil
+				// now, the connections being passed to closeConns() are not reachable from
+				// transport methods anymore. So, it's safe to close them in a separate goroutine
 				go closeConns(stack)
 			}
 			SocketGauge.WithLabelValues(t.addr).Set(float64(t.len()))
@@ -119,6 +121,8 @@ func (t *transport) cleanup(all bool) {
 		}
 		if all {
 			t.conns[proto] = nil
+			// now, the connections being passed to closeConns() are not reachable from
+			// transport methods anymore. So, it's safe to close them in a separate goroutine
 			go closeConns(stack)
 			continue
 		}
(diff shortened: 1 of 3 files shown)


plugin/forward/persistent.go
```diff
@@ -96,13 +92,49 @@ Wait:
...
+			t.cleanup(true)
 			close(t.ret)
 			return
 		}
 	}
 }
 
+// closeConns closes connections.
+func closeConns(conns []*persistConn) {
+	for _, pc := range conns {
```
miekg, line 110: Doing this in a goroutine is a data race on conns, no?

plugin/forward/persistent.go
```diff
@@ -96,13 +92,49 @@ Wait:
...
+			go closeConns(stack)
+			t.conns[proto] = nil
+			continue
+		}
+		if stack[0].used.After(staleTime) {
+			continue
+		}
+
+		// connections in stack are sorted by "used"
+		good := sort.Search(len(stack), func(i int) bool {
```
miekg, line 134: So you're moving the complexity to this function, which would be O(n*log(n)) I guess, best case. I opted not doing this because we just want a conn that is valid, not the last one used...

miekg (changes requested):
Thanks for picking this up. This bugged me as well, as my local monitoring showed 4 conns that never expired.

I think we don't really need the Sort. The closing is more tricky, as you don't want to hold up giving out new connections...

plugin/forward/persistent.go
rdrozhdzh: I remove these connections from `transport.conns` before calling `closeConns()`, so they are unreachable from transport anymore

plugin/forward/persistent.go
rdrozhdzh: > So you're moving the complexity to this function, which would be O(n*log(n))

I believe the complexity of Search() is O(log(n)). So, this should be faster than iterating over all cached connections.

> we just want a conn that is valid, not the last one used...

yes, and the fastest way to find a valid connection is to check the last one. Since it was added last, it has higher probability (comparing to other connections) that it is not expired. If the last (in the slice) connection is expired this means all previous connections are expired as well, we even don't need to check them

Another benefit of taking connection from the end of slice is that we would potentially process higher traffic with less connections in cache (i.e. consume less system resources). Lets consider the following case.

Expiration time is 10s. We have 100 connections in cache after some previous user activity. At some point user starts sending 1 DNS query each 100ms. The DNS processing time is 10ms.

The original code will take a cached connection **from the begin** of cached conns and 10ms later it will add the connection back to the end of conns list with updated "used" timestamp. So, it will just rotate the 100 connections in the cache without releasing them, because each connection will be periodically updated with new timestamp, and no connection will expire

The new code will take a connection **from the end** of list and 10ms later the connection will be added back to the end of conns list with updated "used" timestamp. On next request we will use same connection again. After 10s all connections except the last one will get expired and released (by timer). Only one connection will reside in cache and this single connection would be enough to process current load (i.e. 1 query per 100ms)

So, I would say the new approach is more adaptive to the load and more prudent about system resources

rdrozhdzh:
@miekg, please clarify which changes you want me to add here

miekg:
sorry to be late; busy w/ work lately. Looking now.

your replies are convincing

plugin/forward/persistent.go
miekg: But `conns` is and because we're only `ranging` over it; that makes it save?

plugin/forward/persistent.go
miekg: ok, trust you on this.

plugin/forward/persistent.go
rdrozhdzh: `conns` here is just a parameter passed to `closeConns()`, i.e. local variable. This is not the same as `transport.conns`

miekg: it's a slice, so the underlaying array is shared between the two; it still don't think this is safe to do...

miekg: To move this forward, can we not expire and sort in a separate goroutine? (We do need to quantify the potential stall then though)

rdrozhdzh: yes, the `conns` is a slice, but the underlaying array is not shared. The second reference to this array is erased from `transport.conns` right before calling `closeConns()`, see the lines
https://github.com/coredns/coredns/pull/1782/files#diff-c641d6db9fa84aad4501394efccea2dbR71
and
https://github.com/coredns/coredns/pull/1782/files#diff-c641d6db9fa84aad4501394efccea2dbR133

And seems I missed the line
https://github.com/coredns/coredns/pull/1782/files#diff-c641d6db9fa84aad4501394efccea2dbR122
which also should be moved before `closeConns()`

We actually don't expire and search connections in a separate goroutine. The `cleanup()` is called synchronously from `connManager()`. We only call  `closeConns()` in a separate goroutine which just closes connections which are not accessible from `connManager()` anymore

rdrozhdzh: Well, strictly speaking, in case of https://github.com/coredns/coredns/pull/1782/files#diff-c641d6db9fa84aad4501394efccea2dbR133 the same underlying array is still used, but slices are not intersected. I.e. `connManager()` has a slice with no references to the connections passed to `closeConns()` with another slice. Adding new connection to cache cannot affect the connections passed to `closeConns()`

miekg: ok

miekg:
@johnbelamaric do you have some spare cycles to take a look here. 

The final conclusion looks to be this is OK, but I like a second opinion.

plugin/forward/persistent.go
johnbelamaric: I don't think that's true. I *think* the little program below will simulate this situation:

```
package main

import (
        "log"
        "time"
)

func close(s []int) {
        log.Println("Pre-sleep closing: ", s)
        time.Sleep(1*time.Second)
        log.Println("Post-sleep closing: ", s)
}

func add(s []int) {
}

func main() {
        log.Println("Hi")

        s := make([]int,10)
        for i := 0; i < 10; i++ {
                s[i] = i
        }

        ss := s[5:]
        s = s[:5]
        go close(ss)

        time.Sleep(time.Second/2)
        s = append(s,100)

        time.Sleep(2*time.Second)
        log.Println("s = ", s)
}
```

Running this we see:

```
jbelamaric@ANN-M-JBELAMARIC [k8s:minikube] ~/slices$ go run main.go
2018/05/23 18:28:30 Hi
2018/05/23 18:28:30 Pre-sleep closing:  [5 6 7 8 9]
2018/05/23 18:28:31 Post-sleep closing:  [100 6 7 8 9]
2018/05/23 18:28:32 s =  [0 1 2 3 4 100]
jbelamaric@ANN-M-JBELAMARIC [k8s:minikube] ~/slices$
```

Is there something different about that and the situation here?

rdrozhdzh: @johnbelamaric , your case is different

I pass to the `closeConns()` goroutine the slice which points to the beginning part of the array. And the slice pointing to remaining part overwrites the original slice in transport object which previously pointed to whole array . So, the transport cannot access the elements which resided in the beginning of the original array anymore. If transport appends new element it will be either added to the end of original array, or added to a newly allocated array after copying elements which were accessible from the transport slice. In any case the elements passed to `closeConns()` will remain in place and will not change.

To simulate this situaltion, you need to replace
```
        ss := s[5:]
        s = s[:5]
```
in your example with 
```
        ss := s[:5]
        s = s[5:]
```

johnbelamaric: Ah, got it. Ok 👍

miekg: Thanks both.

We're good to go then?

miekg: I would like some comments in the code on this; it's not *that* obvious how the race is avoided.

rdrozhdzh: done

miekg:
@johnbelamaric happy as well?

miekg (approved):

johnbelamaric (approved):
👍

johnbelamaric:
Pending seeing the CI work!

PR #1784, opened 2018-05-08
Probe simplification

rdrozhdzh:
The main reason of rework is that previous implementation hung when calling Do() after Stop()

#1768

### DIFF
--- a/plugin/pkg/up/up.go
+++ b/plugin/pkg/up/up.go
@@ -3,14 +3,15 @@
 package up
 
 import (
-	"sync/atomic"
+	"sync"
 	"time"
 )
 
 // Probe is used to run a single Func until it returns true (indicating a target is healthy). If an Func
 // is already in progress no new one will be added, i.e. there is always a maximum of 1 checks in flight.
 type Probe struct {
-	inprogress int32
+	sync.Mutex
+	inprogress int
 	interval   time.Duration
 }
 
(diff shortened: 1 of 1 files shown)


plugin/pkg/up/up.go
```diff
@@ -3,80 +3,49 @@
...
-					}
-					time.Sleep(interval)
-					p.Lock()
-					if p.inprogress == stop {
-						p.Unlock()
-						return
-					}
-					p.Unlock()
+func (p *Probe) Do(f Func) {
+	if atomic.CompareAndSwapInt32(&p.inprogress, idle, active) {
```
miekg: Coming back to this, I'm not a fan of this sync/atomic usage; it less clear than the mutexes, probably not faster (if that would be a consideration) and requires 64 bit alignment of the fields.

If we call Do() after Stop() it would call that a mistake of the caller; but bluntly hanging is not use.
Can't we just check `p.inprogress`? When starting or not call Do after Stop from the higher layers?

miekg (changes requested):

plugin/pkg/up/up.go
rdrozhdzh: controlling `Do()` and `Stop()` at higher level is not so easy, since these calls are initiated from ServeDNS() and OnShutdown() which can be called by caddy in **any** order. 
I think controlling `Probe.Do()` is much easier. I'm going to submit another PR with minimal changes to Probe, which will check `p.inprogress` in `Do()`

rdrozhdzh: I thought about change to original code like
```
func (p *Probe) Do(f Func) {
  p.Lock()
  if p.inprogress == idle {
    p.do <- f
  }
  p.Unlock()
}
```
but it leads to deadlock, if `Do()` is called concurrently

I can replace atomics with mutex in this PR if you wish. BTW, 64-bit alignment is required only for 64-bit values (see Bugs at https://golang.org/pkg/sync/atomic/), so int32 should work fine

Or you can just close this PR if you think the original code is good enough

rdrozhdzh:
In last commit I replaced atomic operations with mutex. Now the code is closer to original one, except that it doesn't have channels and extra goroutine.
@miekg, could you please review this change?

plugin/pkg/up/up.go
```diff
@@ -10,73 +10,58 @@ import (
...
-	}
+// Stop stops the probing.
+func (p *Probe) Stop() {
+	p.Lock()
+	p.inprogress = stop
+	p.Unlock()
 }
 
+// Start sets probing interval, after which probes can be initiated with Do.
+func (p *Probe) Start(interval time.Duration) { p.interval = interval }
```
miekg: This assignment will be racy (if Start is done from a different routine than Do). Any reason to split this up?

Looks better with the mutexes, thanks for that.

rdrozhdzh: It's assumed that `Start()` is called only once before first call of `Do()`. The comment above states about this. I believe the original code also relies on this expectation.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Probe simplif..." ]
>
>
>
>>  }
>
>+// Start sets probing interval, after which probes can be initiated with Do.
>+func (p *Probe) Start(interval time.Duration) { p.interval = interval }
>
>It's assumed that `Start()` is called only once before first call of `Do()`. The comment above states about this. I believe the original code also relies on this expectation.

Yes, but now p.interval is pulled in the Probe struct and updated (potentially) 
from diff. go routines. The original code didn't have that (again potential) 
data race condition.

rdrozhdzh:
ok, I will fix this.
btw, don't you mind if I rename `Start()` to `SetInterval()`?

miekg:
Name change is fine. But it will still start probing? Also I'm
contemplating adding some backoff here, so setinterval might not cover the
entire thing

rdrozhdzh:
The original `Start()` method started goroutine, which listened to a channel. Now, after removing channels and goroutine there is no need in any preparation steps, except setting interval (which is 0 by default).
The actual probing will start after calling `Do()`
In future, if you will need to do some preparation logic you can add the method `Start()` again, in addition to `SetInterval()`.
Alternatively, we can keep method `Start()` for now and just call `SetInterval()` from it

miekg (approved):
Ok :- ) thanks

PR #1792, opened 2018-05-09
Up miekg/dns to 1.0.6

miekg:
This was just tagged and has the worker patch that improves performance.


miekg:
cobot wasn't working properly, still this review request could have been valid, merging as -is (1line diff)

PR #1793, opened 2018-05-10
plugin/reload: mention auto in reload

chrisohaver:

Mention early and clearly (within TLDR; threshold), that automatic zone file reloads can be done with `auto` plugin.

This has come up many times, people expecting the `reload` plugin to automatically reload changes to zone files.

### DIFF
--- a/plugin/reload/README.md
+++ b/plugin/reload/README.md
@@ -2,7 +2,9 @@
 
 ## Name
 
-*reload* - allows automatic reload of a changed Corefile
+*reload* - allows automatic reload of a changed _Corefile_<sup>*</sup>
+
+<sub>* To enable automatic reloading of a _zone file_, use the `auto` plugin.</sub>
 
 ## Description
 
--- a/plugin/reload/README.md
+++ b/plugin/reload/README.md
@@ -2,11 +2,12 @@
 
 ## Name
 
-*reload* - allows automatic reload of a changed _Corefile_<sup>*</sup>
-
-<sub>* To enable automatic reloading of a _zone file_, use the [`auto`](https://github.com/coredns/coredns/blob/master/plugin/auto/README.md) plugin.</sub>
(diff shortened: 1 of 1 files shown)


chrisohaver:
@fastest963, disregard the review request... our corebot seems to be momentarily confused.

stp-ip (approved):

stp-ip:
I think a link to the auto plugin docs would be even better, but happy to get this merged as is.

stp-ip:
`[`auto`](/plugin/auto/README.md)` as relative link also works. That would make it easier for it to be linked to each branches docs etc. Not sure what's the general stance in CoreDNS on this. In Kubernetes it is relative.

chrisohaver:
The relative link will (i think) result in a broken link in the web site presentation at http://coredns.io.
The website uses "plugins" with an "s" ...

stp-ip:
Just checked. We don't have any cross plugin links yet. This will be the first. We have various links to the codebase, so I think it's fine to default to a link to the github repo README instead of the website docs.

@miekg What's your stance?

plugin/reload/README.md
```diff
@@ -2,7 +2,9 @@
 
 ## Name
 
-*reload* - allows automatic reload of a changed Corefile
+*reload* - allows automatic reload of a changed _Corefile_<sup>*</sup>
+
+<sub>* To enable automatic reloading of a _zone file_, use the [`auto`](https://github.com/coredns/coredns/blob/master/plugin/auto/README.md) plugin.</sub>
```
miekg: Good to have, but this may (haven't check the sync script for getting it on coredns.io). Also `<sub>` def. break the conversion into man pages.

Good to have this warning but it probably needs to move into description, or See Also (but that may be too far down?)

chrisohaver: I've removed the links and sub tags and moved the text to the Description section.

PR #1795, opened 2018-05-11
plugin/rewrite: Try to clarify docs for response rewrites

chrisohaver:

Tries to clarify rewrite docs, Making response rewrites more prominent in docs (by including `answer name` in the valid list of FIELDs).
I also constructed an updated/complete syntax description, but I think it's probably too hard to digest.  So I left the existing on in place and put it at the bottom of the docs, just for completeness.

The complexity of the syntax is partly due to the multi line option for response re-writes.  The multi line option presents `answer` as a valid FIELD descriptor, when it actuality it is an optional extension of the `name` FIELD rewrite (but only if `regex` is used to match the name, so technically it's an optional extension of `regex`). Regardless, since we prefer  the friendlier multi line option in the docs, I am listing `answer name` as a valid field.

the syntax of response rewrites has come up in several issues.

### DIFF
--- a/plugin/rewrite/README.md
+++ b/plugin/rewrite/README.md
@@ -11,24 +11,20 @@ Rewrites are invisible to the client. There are simple rewrites (fast) and compl
 
 ## Syntax
 
+A simplified/easy to digest syntax for *rewrite* is...
 ~~~
 rewrite [continue|stop] FIELD FROM TO
 ~~~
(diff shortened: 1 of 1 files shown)


johnbelamaric (approved):
👍

greenpau (approved):
LGTM 👍

PR #1798, opened 2018-05-15
Change Dockerfle to run as unprivileged user, fix #1767

### LINKED ISSUE #1767: docker: run as non-root
Can't find the original issue anymore (think we had one on this). But after seeing Liz's keynote this morning at kubecon we should def. run as non-root for CoreDNS in the Docker container.

Couple a things:
1: https://github.com/coredns/deployment/blob/master/kubernetes/coredns.yaml.sed#L48 uses port 53 explicitly, meaning dns.port=xxxx won't work
2. binding on a higher port and then redirecting port 53 to this higher port can't be done in the Dockerfile?
3. If 2 is true, then this is a backwards incompatible change.

m4r10k:
Change the Dockerfile to run as unprivileged user on privileged port

#1767

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,13 @@
 FROM alpine:latest
 
-# Only need ca-certificates & openssl if want to use DNS over TLS (RFC 7858).
-RUN apk --no-cache add bind-tools ca-certificates openssl && update-ca-certificates
-
+# first add the file because we have to run setcap on the binary and we would not like to create a new image layer so the binary must exist on RUN
 ADD coredns /coredns
 
+# Only need ca-certificates & openssl if want to use DNS over TLS (RFC 7858).
+RUN apk --no-cache add bind-tools ca-certificates openssl libcap && update-ca-certificates && chown nobody:nobody /coredns && setcap CAP_NET_BIND_SERVICE=+eip /coredns
+
+USER nobody
+
 EXPOSE 53 53/udp
+
 ENTRYPOINT ["/coredns"]


m4r10k:
@chrisohaver Huh? This Dockerfile builds without any issues locally... does Travis run build in another way? 

`m[91mFailed to set capabilities on file `/coredns' (Not supported)`

```
docker build . -t testcoredns
Sending build context to Docker daemon  56.16MB
Step 1/6 : FROM alpine:latest
 ---> 3fd9065eaf02
Step 2/6 : ADD coredns /coredns
 ---> 333c8988962d
Step 3/6 : RUN apk --no-cache add bind-tools ca-certificates openssl libcap && update-ca-certificates && chown nobody:nobody /coredns && setcap CAP_NET_BIND_SERVICE=+eip /coredns
 ---> Running in 6e671ea7bfdb
fetch http://dl-cdn.alpinelinux.org/alpine/v3.7/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.7/community/x86_64/APKINDEX.tar.gz
(1/9) Installing libgcc (6.4.0-r5)
(2/9) Installing libxml2 (2.9.7-r0)
(3/9) Installing bind-libs (9.11.2_p1-r0)
(4/9) Installing bind-tools (9.11.2_p1-r0)
(5/9) Installing ca-certificates (20171114-r0)
(6/9) Installing libcap (2.25-r1)
(7/9) Installing libcrypto1.0 (1.0.2o-r0)
(8/9) Installing libssl1.0 (1.0.2o-r0)
(9/9) Installing openssl (1.0.2o-r0)
Executing busybox-1.27.2-r7.trigger
Executing ca-certificates-20171114-r0.trigger
OK: 13 MiB in 20 packages
WARNING: ca-certificates.crt does not contain exactly one certificate or CRL: skipping
Removing intermediate container 6e671ea7bfdb
 ---> 5a2d76f009c0
Step 4/6 : USER nobody
 ---> Running in 4c03f07baa90
Removing intermediate container 4c03f07baa90
 ---> e5f72715eded
Step 5/6 : EXPOSE 53 53/udp
 ---> Running in fb3389f0fd83
Removing intermediate container fb3389f0fd83
 ---> b62bf1b7500e
Step 6/6 : ENTRYPOINT ["/coredns"]
 ---> Running in 386ed3883afe
Removing intermediate container 386ed3883afe
 ---> 8dc327d63c07
Successfully built 8dc327d63c07
Successfully tagged testcoredns:latest
```

m4r10k:
Oh, and maybe we should avoid `:latest`! We should tag the image we would like to use and update it on every release. Otherwise there might be problems in the future maybe.

chrisohaver:
> Travis run build in another way ?

I don't think Travis builds the docker image... Travis is mostly just unit tests. 

The coredns/ci tests are a set of kubernetes integration tests, running coredns in a kubernetes pod, so all tests depend on a successful build of the docker image.

Full error is ... 
```
[0m[91mFailed to set capabilities on file `/coredns' (Not supported)
usage: setcap [-q] [-v] (-r|-|)  [ ... (-r|-|)  ]

 Note  must be a regular (non-symlink) file.
```

Could be a docker version difference affecting setcap.  What version of docker are you using?  I'll check the CI system docker version.

chrisohaver:
the Ci system was on Docker version 17.12.1-ce... i updated it to 18.03.1-ce ... re-runnning /integration to see if that fixes things...

chrisohaver:
Same error after upgrading to 18.03.1-ce.

m4r10k:
Hm, I have the same Docker version. I've read a little bit about it and it might be a file system issue because the CI job is running DIND and therefore it maybe is using AUFS (https://github.com/hashicorp/docker-vault/issues/22).

I have tried it with a DIND container but the build also works without nagging about the setcap. 

```
docker version
Client:
 Version:      18.03.1-ce
 API version:  1.37
 Go version:   go1.9.5
 Git commit:   9ee9f40
 Built:        Thu Apr 26 07:17:20 2018
 OS/Arch:      linux/amd64
 Experimental: false
 Orchestrator: swarm

Server:
 Engine:
  Version:      18.03.1-ce
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.9.5
  Git commit:   9ee9f40
  Built:        Thu Apr 26 07:15:30 2018
  OS/Arch:      linux/amd64
  Experimental: false
```

chrisohaver:
The CI is using AUFS, on which as you have seen, others have reported the setcap error.
The CI is not using DIND afaik.  I think AUFS used to be the default storage driver... so it could be a holdover config over several upgrades.

m4r10k:
Bad. As it would be a nice solution for the root problem. If found this too. https://github.com/moby/moby/issues/20658

So as I read it, the `CONFIGAUFSXATTR=y` may be missing in the Linux kernel configuration of this host. 

As an alternative there would be `authbind` ,but this is not available in the Alpine repos and the normal way should be using `libcap` quoting Justin Cormack https://github.com/gliderlabs/docker-alpine/issues/166 .

Dockerfile
```diff
@@ -1,9 +1,13 @@
...
 
+# first add the file because we have to run setcap on the binary and we would not like to create a new image layer so the binary must exist on RUN
+ADD coredns /coredns
+
 # Only need ca-certificates & openssl if want to use DNS over TLS (RFC 7858).
-RUN apk --no-cache add bind-tools ca-certificates openssl && update-ca-certificates
+RUN apk --no-cache add bind-tools ca-certificates openssl libcap && update-ca-certificates && chown nobody:nobody /coredns && setcap CAP_NET_BIND_SERVICE=+eip /coredns
 
-ADD coredns /coredns
+USER nobody
```
miekg, line 9: this works, but now we depend on setcap, and this further complicates building for other architectures.
IMO the fix should be run on a high port as user non-root.

But then we need to port forward which you can't specify in the Dockerfile :-(

m4r10k: :smile: that's the problem. Security is always a trade off and most of the time between easy and complicated.

m4r10k: What about the idea to create a pipeline which builds the binary and test it in the common way and then have the Dockerfiles for the different images (and different possibilities) in a separate directory in the repository with multiple pipelines for the image builds. Then it would be possible to create secure(er) images at least for some platforms. That won't be a 100 percent solution but anyway better than nothing.

chrisohaver:
testing /integration with overlay2 storage driver (it worked fwiw)

m4r10k:
@chrisohaver Wow thanks! Nice to know that it basically works. The problem always stays the same, here is a example: If you run Apache Tomcats in containers for example, than they run as tomcat user and on port 8080 which is perfectly ok. Also they are using port 8009 for the AJP protocol which is often preferred as load balancing solution in combination with Apache httpd. Now, you have to provide a Apache httpd as load balancer and this one will usually run on port 80 or 443 because no user would like to type in port numbers in the browser.

This leads to the problem, that the Apache httpd has to bind on privileged ports and that is done via a controlling process that runs as root (to bind to the low port number) which forks the workers that are running as unprivileged user.

```
root      1108     1  0 01:01 ?        00:00:00 /usr/sbin/apache2 -k start
www-data  4935  1108  0 06:25 ?        00:00:02 /usr/sbin/apache2 -k start
www-data  4936  1108  0 06:25 ?        00:00:02 /usr/sbin/apache2 -k start
www-data  4937  1108  0 06:25 ?        00:00:02 /usr/sbin/apache2 -k start
www-data  4938  1108  0 06:25 ?        00:00:02 /usr/sbin/apache2 -k start
...
```

So the core of the problem stays the same if the Apache httpd is running as container - you cannot change this default behavior (root process) until you patch it in the way we discuss it here - if the root process fails, anything can happen. You can lower the risk by putting the Apache httpd on a separate host (which should be done so) but this won't work for CoreDNS as it is the Kubernetes DNS integrated solution somewhere in the future. Beside that, UDP load balancing is not nice neither. It's a chicken-and-egg question.

This was hardly seen as a problem in past (it doesn't come to mind by most of the people), because you had running a lot of "root" daemons on one host. If one of the daemons was hijacked, all your other services might be suspected also. Now the host ist shared by loads of containers, where you do not know the source in every case. Therefore absolutely no container shall run root processes. If one (and only one) of probably thousands of containers runs a root process and this one is hijacked, everything is odd. So full security is hard to achieve (if ever).

I think every solution can do the best on its own to not run as root. And if others do so too, the situation gets better and for suspects harder to exploit. And of course, with Gvisor the situation gets better too.

Sorry for the long text :smirk: :blush: - this PR is probably only one solution but a good place to discuss.

Dockerfile
miekg: Yes, this needs more then just changing the Dockerfile.

Right now we also don't install certs in non-amd64 images which could be fixed in the same way.

I can build those images on my laptop, but that code is non-portable. We looked at docker automatic builds a while back, but nothing came out of it and I'm not sure it can actually do multi arch builds

m4r10k: Yes, that's true. I haven't looked at multi arch build too. And I have to say that we use Gitlab (on-prem) and the pipelines there and therefore I have no experience with Travis and the possibilities there.

We sometimes have to build build images. What I mean is, maybe it is possible to put your local build code into a prepared build image and to use this image then to build the coredns release  images. But this needs stacked pipelines. The first one build the build system and the second one build the release images with the result from the first one.

miekg: for Linux you'll need qemu and binfmt to be setup in your kernel, see https://github.com/miekg/dxbuild

This won't even run in Travis...

m4r10k: :smile: That's why we use Gitlab on-prem because than you are able to use Windows with Gitlab runners to run UI tests from the pipeline or to prepare special build hosts to run pipeline steps on them and so on for example. 80% of the builds are perfectly happy with the default solution but the other 20% are special, every single one of them. :rescue_worker_helmet:

miekg:
For k8s at least, this https://github.com/coredns/deployment/commit/aba0245609821e8bba8b4ecd6a32c1a02a7bf658 is a less invasive fix

m4r10k:
Yes it is! Thanks for the mentioning!

miekg:
We can prolly do something like this but it needs to be in another image, i.e. coredns-nobody or something.

yongtang:
+1 for a separate image.

miekg:
should we close this or can it be reworked into a second docker container we deploy coredns-nobody?  I think we only need to set `USER nobody`, some port foo and modify Makefile.release .

Issue #1785, opened 2018-05-08
TestNewServer hangs


rdrozhdzh:
```
=== RUN   TestNewServer
SIGQUIT: quit
PC=0x45b643 m=0 sigcode=0
goroutine 4 [syscall]:
runtime.notetsleepg(0x846700, 0x773556ee, 0x16)
	/home/travis/.gimme/versions/go1.10.1.linux.amd64/src/runtime/lock_futex.go:227 +0x42 fp=0xc420043760 sp=0xc420043730 pc=0x410322
runtime.timerproc(0x8466e0)
	/home/travis/.gimme/versions/go1.10.1.linux.amd64/src/runtime/time.go:261 +0x2e7 fp=0xc4200437d8 sp=0xc420043760 pc=0x449b07
runtime.goexit()
	/home/travis/.gimme/versions/go1.10.1.linux.amd64/src/runtime/asm_amd64.s:2361 +0x1 fp=0xc4200437e0 sp=0xc4200437d8 pc=0x45a0a1
created by runtime.(*timersBucket).addtimerLocked
	/home/travis/.gimme/versions/go1.10.1.linux.amd64/src/runtime/time.go:160 +0x107
goroutine 1 [chan receive]:
testing.(*T).Run(0xc4200ee5a0, 0x6da588, 0xd, 0x6e72d8, 0x475601)
	/home/travis/.gimme/versions/go1.10.1.linux.amd64/src/testing/testing.go:825 +0x301
testing.runTests.func1(0xc4200ee000)
	/home/travis/.gimme/versions/go1.10.1.linux.amd64/src/testing/testing.go:1063 +0x64
testing.tRunner(0xc4200ee000, 0xc420065d90)
	/home/travis/.gimme/versions/go1.10.1.linux.amd64/src/testing/testing.go:777 +0xd0
testing.runTests(0xc4200a6260, 0x83e940, 0x6, 0x6, 0x411059)
	/home/travis/.gimme/versions/go1.10.1.linux.amd64/src/testing/testing.go:1061 +0x2c4
testing.(*M).Run(0xc4200ec000, 0x0)
	/home/travis/.gimme/versions/go1.10.1.linux.amd64/src/testing/testing.go:978 +0x171
main.main()
	_testmain.go:102 +0x208
goroutine 24 [chan receive]:
github.com/coredns/coredns/plugin/pkg/dnstest.NewServer(0x6e72d0, 0x81fca7)
[...]

PR #1799, opened 2018-05-15
retry creating dnstest.Server if tcp or udp port is in use

rdrozhdzh:
retry creating tcp and udp listeners if desired UDP port is already in use

#1785

none

### DIFF
--- a/plugin/pkg/dnstest/server.go
+++ b/plugin/pkg/dnstest/server.go
@@ -26,10 +26,10 @@ func NewServer(f dns.HandlerFunc) *Server {
 	s1 := &dns.Server{} // udp
 	s2 := &dns.Server{} // tcp
 
-	for i := 0; i < 1000; i++ { // 1000 attempts
+	for i := 0; i < 5; i++ { // 5 attempts
 		s2.Listener, _ = net.Listen("tcp", ":0")
 		if s2.Listener == nil {
-			panic("dnstest.NewServer(): failed to create tcp listener")
+			continue
 		}
 
 		s1.PacketConn, _ = net.ListenPacket("udp", s2.Listener.Addr().String())
--- a/plugin/pkg/dnstest/server.go
+++ b/plugin/pkg/dnstest/server.go
@@ -23,18 +23,28 @@ func NewServer(f dns.HandlerFunc) *Server {
 	ch1 := make(chan bool)
 	ch2 := make(chan bool)
 
-	l, _ := net.Listen("tcp", ":0")
-	if l == nil {
-		return nil
+	s1 := &dns.Server{} // udp
+	s2 := &dns.Server{} // tcp
+
+	for i := 0; i < 1000; i++ { // 1000 attempts
+		s2.Listener, _ = net.Listen("tcp", ":0")
+		if s2.Listener == nil {
+			panic("dnstest.NewServer(): failed to create tcp listener")
+		}
+
+		s1.PacketConn, _ = net.ListenPacket("udp", s2.Listener.Addr().String())
+		if s1.PacketConn != nil {
+			break
+		}
+
+		// perhaps UPD port is in use, try again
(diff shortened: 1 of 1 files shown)


miekg:
The doing tcp first should be sufficient, complicating this further doesn't add much IMO

rdrozhdzh:
@miekg, I reopened #1785 because I caught the test failed with panic due to this issue, see https://github.com/coredns/coredns/issues/1785#issuecomment-389111894

miekg:
none of the tests assume the testserver will fail, so this is a pretty harmless panic

rdrozhdzh:
reopening since the issue is getting annoying

plugin/pkg/dnstest/server.go
```diff
@@ -23,18 +23,28 @@ func NewServer(f dns.HandlerFunc) *Server {
 	ch1 := make(chan bool)
 	ch2 := make(chan bool)
 
-	l, _ := net.Listen("tcp", ":0")
-	if l == nil {
-		return nil
+	s1 := &dns.Server{} // udp
+	s2 := &dns.Server{} // tcp
+
+	for i := 0; i < 1000; i++ { // 1000 attempts
```
miekg: Can we make it, 5 ? 1000 looks like very much an overkill?

plugin/pkg/dnstest/server.go
```diff
@@ -23,18 +23,28 @@ func NewServer(f dns.HandlerFunc) *Server {
...
+		s2.Listener, _ = net.Listen("tcp", ":0")
+		if s2.Listener == nil {
+			panic("dnstest.NewServer(): failed to create tcp listener")
```
miekg: shouldn't this be `continue`?

rdrozhdzh: Here we ask system for **any** available port, so if it fails here this means that system has no available ports or some error not related to port. So, I don't think we should retry in this case

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] retry creatin..." ]
>
>
>
>> @@ -23,18 +23,28 @@ func NewServer(f dns.HandlerFunc) *Server {
> 	ch1 := make(chan bool)
> 	ch2 := make(chan bool)
>
>-	l, _ := net.Listen("tcp", ":0")
>-	if l == nil {
>-		return nil
>+	s1 := &dns.Server{} // udp
>+	s2 := &dns.Server{} // tcp
>+
>+	for i := 0; i < 1000; i++ { // 1000 attempts
>+		s2.Listener, _ = net.Listen("tcp", ":0")
>+		if s2.Listener == nil {
>+			panic("dnstest.NewServer(): failed to create tcp listener")
>
>Here we ask system for **any** available port, so if it fails here this means that system has no available ports or some error not related to port. So, I don't think we should retry in this case

Why not, we do vagely is similar thing for the udp port, as ports come and go, 
we can just as well retry again...

plugin/pkg/dnstest/server.go
rdrozhdzh: It depends on how much free port system has. Yes, 5 attempts may be enough. On the other hand I don't see any drawback with 1000 attempts - it won't take much time. If you prefer 5 I can change this

rdrozhdzh:
> Why not, we do vagely is similar thing for the udp port, as ports come and go, we can just as well retry again...

I mean in case of TCP (first listener) the error is rather hypothetical and likely not related to free ports. In case of UDP (second listener) we request some specific port and the probability of failure for this specific port is quite high, though some other ports in the system is likely free.

I can change to continue if you insist

miekg:
It's just a number, but 1000 feels like overkill, so pretty please :)

I would also favor removing the first panic and just try again for the tcp port allocation.

rdrozhdzh:
Done :)

PR #1803, opened 2018-05-16
Dep ensure

miekg:
Cut back on the non-vendored packages

We move the context from the stdlib, so we should be able to remove some of the non-vendored deps.


yongtang (approved):
LGTM

PR #1806, opened 2018-05-16
etc/kubernetes: remove dedup

miekg:
Don't think this is actually needed and slows things down.


miekg:
Nice, this fails the exact thing it should prevent. Good chance to clean that up

chrisohaver:
Probably need to avoid adding duplicates in backend type funcs (e.g. A(), AAAA(), etc...)

miekg:
hmm I wonder if kubernetes is affected though, etcd def. is

chrisohaver:
k8s failed 15 tests... most look to be duplicates issues.

miekg:
ah thanks. We can choose to do a AppendUnique (or something). Might be a matter of benchmarking though; think dedup might be a tad smarter as well.

miekg:
abandoning this effort :)

miekg:
looking at dns.Dedup we should probably not add dups in the first place. Find a dup is tricky though as we need to compare rdata and miekg/dns doesn't really have a function for that.

chrisohaver:
The reason k8s add dups is because the record retrieval funcs (e.g. findServices) are unaware of the query type.  They will for example pull a service multiple times if it has more than one port.  The extra records are duplicates from an `A` response POV... but they are needed for an `SRV` response.

IMO, I think the fact that the record retrieval funcs are unaware of record they need to build doesn't help us that much.  It may reduce some code slightly ... but at the expense of doing extra work most of the time (most queries are not `SRV`).

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] etc/kubernete..." ]
>The reason k8s add dups is because the record retrieval funcs (e.g. findServices) are unaware of the query type.  They will for example pull a service multiple times if it has more than one port.  The extra records are duplicates from an `A` record POV... but they are needed for an `SRV` record.
>
>IMO, I think the fact that the record retrieval funcs are unaware of record they need to build doesn't help us that much.  It may reduce some code slightly ... but at the expense of doing extra work most of the time (most queries are not `SRV`).

Ack.

Note dedup was added because multiple identical RRs are a protocol violation.

miekg:
https://github.com/miekg/dns/issues/679 filed

jameshartig:
Somewhat related, I was added as a reviewer but I'm not in the OWNERS file for the kubernetes plugin. Any idea why it chose me?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] etc/kubernete..." ]
>Somewhat related, I was added as a reviewer but I'm not in the OWNERS file for the kubernetes plugin. Any idea why it chose me?

my guess (don't log this in corbot, i should), dnstest -> pkg -> plugin -> 
coredns only has an owner on the top level.

PR #1816, opened 2018-05-19
plugin/template: Support NODATA responses

grobie:
A NODATA response has no answers and rcode NOERROR, but should have a
SOA record in the authority section.


rtreffer (approved):
Yes please!

PR #1820, opened 2018-05-19
Enforcing gofmt -s on linter check

yongtang:
This fix enforces gofmt -s on linter check in Makefile,
and fixes plugin/kubernetes/handler_test.go with `gofmt -s`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ pb:
 linter:
 	go get -u github.com/alecthomas/gometalinter
 	gometalinter --install golint
-	gometalinter --deadline=2m --disable-all --enable=golint --enable=vet --vendor --exclude=^pb/ ./...
+	gometalinter --deadline=2m --disable-all --enable=gofmt --enable=golint --enable=vet --vendor --exclude=^pb/ ./...
 
 .PHONY: goimports
 goimports:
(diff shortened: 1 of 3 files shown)


greenpau:
@yongtang , the integration test is failing. Please address and ping me back.

greenpau (changes requested):
Please fix the integration test. Several files are not formatted in accordance with the new standard.

yongtang:
Interesting, golang 1.9 and 1.10 have differences with gofmt:
```
diff --git a/plugin/tls/tls_test.go b/plugin/tls/tls_test.go
index 84a74ae3..0bbba18a 100644
--- a/plugin/tls/tls_test.go
+++ b/plugin/tls/tls_test.go
@@ -14,8 +14,8 @@ func TestTLS(t *testing.T) {
                expectedRoot       string // expected root, set to the controller. Empty for negative cases.
                expectedErrContent string // substring from the expected error. Empty for positive cases.
        }{
-       // positive
-       // negative
+               // positive
+               // negative
        }
 
        for i, test := range tests {
```

I think coredns/ci may need to update to 1.10.

miekg:
Hmm. I've *manually* upgraded the go version to  go1.10.2 linux/amd64

chrisohaver:
> Hmm. I've manually upgraded the go version to go1.10.2 linux/amd64

Thanks, yes, manual is the way to upgrade go for coredns/ci.

chrisohaver:
How is the new format different than before?

Didn't we already enforce gofmt in make?  Why the change?

chrisohaver:
> Didn't we already enforce gofmt in make? Why the change?

I see.  It was moved to the linter in September, then removed from linter in February on the assumption it was being performed twice...  but it seems it was the only check at the time.

chrisohaver:
@greenpau, all checks pass now.

greenpau (approved):

PR #1822, opened 2018-05-20
[WIP] Dynamic updates API with listen server

syndbg:

Adds an "experimental" API for dynamic updates of resource records.

Currently I plan to just test the waters and validate the direction I'm going in.

Tests are lacking, it's just a basic proof-of-concept. I'll add relevant ones in the next 24 hours.

Feedback is greatly appreciated by maintainers and anyone interested.

Currently what works is creating a POST request with A record that gets added to a zone managed by the file plugin.  

Somewhat related in the concept of managing records

https://github.com/coredns/coredns/issues/522

This is still WIP, I'll update when I have a clear change/feature-set.


miekg:
[unrelated to this PR]:
@bradbeam reseeding the random now before choosing a reviewer :)
(testing corbot is a bit tricky, so some of it happens in prod)

miekg:
I have a (fairly) generic critique on the direction we take here. Why should 
coredns be in the business of provisioning records? Why not use the specific 
provisioning mechanism for the backend, a database, or kv store or whatever?

syndbg:
For me the use cases revolve around:

* using coredns as main DNS server,
* having the ability to manage (add mostly) DNS records via API,
* having no other dependencies and other sources of truth, such as databases.

My personal problem with the other source-of-truth approach such as databases is asking the questions:

* Do I need to persist the records? Is in-memory good enough for me?

My use-case answer: In-memory is good enough for me and it's better not to persist. It may create awkward sync scenarios if it was persisted. 

* What's the ease of communication with the other source of truth? It varies by database/kv store and etc.

Curl/whatever API is good enough and easily accessible in any language. Other stores require specific drivers/binary protocol. Rest API support is rare except for CouchDB, from what I can remember right now.

* What is CoreDNS doing to sync with the other source of truth? Do I do it via event listening and/or polling?

I would prefer to do it via event listening, but this is not supported by any store.

miekg:
kinda neat that this is contained in a plugin.

Another problem with providing these kind of interfaces is the feature creep the elect: user, permissions all that kind of stuff. Even with the DNS dynamic updates (which we probably should support at some point *sigh*) you get into this territory.

Potentially we could create a coredns/dynapi to flesh out the idea and have a official out-of-tree home for it.

johnbelamaric:
Yeah.  We have separate things:

1) The Go API (interface or interfaces) for plugins to implement to say they are "writable". 
2) The API used by clients to push actual writes.

Each plugin can decide what they want to do for the first - or if they want to support it at all. Then, for the second, we have a separate set of plugins to expose different ways to do it:
- DDNS
- HTTPS
- gRPC

The authentication and authorization are whole separate topics that would need to be addressed, hopefully in ways that are usable across client APIs as much as possible.

johnbelamaric:
> Each plugin can decide what they want to do for the first - or if they want to support it at all. Then, for the second, we have a separate set of plugins to expose different ways to do it:

I don't mean each one defines its own API - I mean they can implement "Writable" or not. When we get a write request in, we figure out which plugin manages that zone and if it doesn't implement the interface we fail.

syndbg:
@johnbelamaric I would say implementing `Writable` is the way to go. Doing a strategy pattern is much more manageable.

syndbg:
Just a heads up. I'll be with very limited available time during the next 1-2 weeks. 

Slow progress is expected.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [WIP] Dynamic..." ]
>Just a heads up. I'll be with very limited available time during the next 1-2 weeks.
>
>Slow progress is expected.

Appreciate the heads-up.

Thanks!

syndbg:
@miekg Input is appreciated for the direction I went in.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [WIP] Dynamic..." ]
>@miekg Input is appreciated for the direction I went in.

And don't think CoreDNS should be on the hook for providing an API that writes 
to some backend. Zone transfers do actually fall into that category as well, but 
meh, they are part of the standard.

All these things start simple and then balloon into access control and 
authorization.

As such this should move to be come an external plugin, which can live on 
coredns/dynapi or some other github repo. If there is a lof of demand for this, 
we can revisit.

/Miek

--
Miek Gieben

syndbg:
@miekg I've sent you a direct message in CNCF's slack on how to proceed. 

My handle is `syndbg`. If you can take a look that would be great.

It's regarding how to proceed.

piaoyu:
func (f File) Delete(request *dynapi.DynapiRequest) error {
	return errors.New("not supported")
}
 func (f File) Update(request *dynapi.DynapiRequest) error {
	return errors.New("not supported")
}
@syndbg why not support Delete and Update? 
When a traditional monolithic app move to the microservices on kubernetes. We want to open the delete and update to the user?

syndbg:
@piaoyu Hi, it looks like you're looking at an old commit. From the most recent diff, `delete` and `update` is supported.

piaoyu:
@syndbg  Oh,I put my foot in my mouth.  delete and update is supported. But why not use grpc ?

syndbg:
@miekg Ping. Can we proceed to move `dynapirest`  to a separate git repository inside `coredns` org?

`dyndns` will remain as part of `coredns/coredns`, right?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [WIP] Dynamic..." ]
>@miekg Ping. Can we proceed to move `dynapirest`  to a separate git repository inside `coredns` org?
>
>`dyndns` will remain as part of `coredns/coredns`, right?

yes, we can easily do this. What do you need coredns/<new-name> ?

syndbg:
@miekg `coredns/dynapirest`

vtolstov:
any progress? I'm realy want to try create/list/delete domains and records via api

syndbg:
@vtolstov Waiting for the repository.

Plugin-wise we've been using it since little before May 20th in a production environment where a small daemon reads DHCP-leases and sends them to coredns to dynamically register DNS A records.

Worked great, however one issue was found a while ago for our use case: that DNS records created with no TTL specified would use a default one of `3600`.  With that fixed with https://github.com/coredns/coredns/pull/1822/files#diff-9233dbaadd466f84dd978f0b7304e340R46 and specifiable `TTL` as part of the create/update request - it fits our need perfectly with short TTL of DNS records.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [WIP] Dynamic..." ]
>@vtolstov Waiting for the repository.

Let's get that fixed now. Making 'coredns/dynapi'

vtolstov:
Can you give some info how to setup if I'm already have working coredns
server.

сб, 1 Сен 2018, 16:33 Anton Antonov <notifications@github.com>:

vtolstov:
gentle ping..

syndbg:
@vtolstov Hi, hopefully this weekend I'll have the time to handle it.

I've seen that you saw the other PR.

miekg:
https://github.com/coredns/dynapi exists, closing here.

ventz:
@miekg @syndbg It looks like that repo has no code?

Would it be possible to upload the code somewhere?

It seems a bit crazy that there's no REST API to CoreDNS...just from combing the threads/issues/requests/wikis -- this seems something a lot of users are asking about...

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [WIP] Dynamic..." ]
>It seems a bit crazy that there's no REST API to CoreDNS...just from combing

??

Such API should work on the data. There could be a case made that we need one to push
config updates, but I fail to see how different that is from just pushing out a new config
file - which should be fairly static anyway.


/Miek

--
Miek Gieben

ventz:
@miekg One could argue that nsupdate/rndc and dynamic zones can push data, but for modern applications, having a programable way to update dns (configs and data) is critical.

There is a reason CloudFlare, Route53, Azure, Google Cloud DNS, Digital Ocean, DNSimple, Dyn, and many others all provide REST APIs.

Be it LetsEncrypt integration, or pushing microservice data (side cart, registration - really anything outside of K8 where you can rely on etcd/consul, and then "discover" the data with CoreDNS) becomes difficult and sometimes "impossible".

For security/blacklisting "rpz like" functionality is another use case. While you can rely on rndc/zone transfers, they are slow, "heavy", and quite involved for adding programmable services on top.

Essentially something like this would fix it: https://doc.powerdns.com/authoritative/http-api/index.html

miekg:
If there would be any api worth emulating it's xDS from Envoy. all others are pretty bad.

ventz:
Envoy's xDS API is great.

I believe they also have both gRPC xDS and REST, although the HTTP-JSON polling may not actually be needed here.
In general, some sort of a "web/rest" front-end would be needed at the end of the day on top of whatever protocol is used underneath.

CoreDNS is well positioned to be the "default" dns choice for everyone and everything. Between the portability, speed, and ease of config/use, along with things like Kubernetes *finally* replacing kube-dns as a default in the last year and a half, there really are just a couple of small things missing from making CoreDNS the go-to choice.

yongtang:
I would be very much in favor of APIs that are similar to Envoy's xDS, though one thing with CoreDNS is that, CoreDNS is more or less a `frontend` with data backends stored in many other places. For example, the DNS records could be coming from an local hostile, a local record file, from k8s (stored in etcd but exposed in api server), or from cloud vendors (route53, clouddns).

The question is, what kinds of data should (or could) be updated from coredns `frontend` logically? Should we modify a local host file if hosts plugin is enabled? Not sure this is something we want to do as it could be really messy. After all, when hosts plugin is enabled, it loads hostfile as `immutable`.

ventz:
@yongtang That's a good question.

If CoreDNS is acting as a "proxy-only" front-end to a full back-end service (service being the key part), then it probably makes sense for that back-end service (and not CoreDNS) to be responsible for the data as the "data custodian". Essentially making sure it's accurate, up to date, etc.

If the backend is not a service but purely a "storage" mechanism, then CoreDNS is no longer acting as just a "proxy-only" front-end, but it is the "data custodian" and source of truth/responsible for updates.

This is where it gets blurry. A file is clearly a storage mechanism. What about S3? (50-50? - it is technically a service and it has an API) What about a DB?

My main argument "for" file being included is that the majority of users utilizing CoreDNS as an authoritative DNS are not using a 3rd party service that already has an API, and thus not using CoreDNS as a "proxy".  Instead, they are using it as an authoritative (not in DNS terms, but as in data custodian) data service, and are relying on RFC 1035 (5.1) zones.

Given the immutable hostfile, it may make sense to either create an exception for "file" types, or better yet treat them as a read-only "stub zones" (where they would have to be empty or would be used only as a "load" mechanism), and create a "write cache" which can serialize the zone (ex: write-on-commit, and load-on-start)

If you are using a 3rd party service, from a practical point - one could argue that the "easiest" solution currently is the redis plugin, and simply adding "write" (update) ability, or potentially just interacting with redis via the API instead - but it feels "hacky", and it requires having redis.

yongtang:
@ventz I think starting with an external plugin with clearly defined backend storage and data model would make sense. Moving the plugin into default coredns could be considered if there is good quality and community interests. The data model might be the most challenging parts: what types of records are being covered?, a key-value of IP-domain mapping only, or any DNS record types (note DNS record types are quite complex and is not really a plain key-value mapping).

ventz:
> Moving the plugin into default coredns could be considered if there is good quality and community interests.
Makes sense.

>The data model might be the most challenging parts: what types of records are being covered?
Agreed. Even between A vs PTR, it's not quite "as easy" to implement PTRs.

I think the records that "matter" come down to:
* SOA
* NS
* A / AAAA
* PTR
* CNAME
* TXT
* SPF? - (I would argue against this one - deprecated for ~6 years now...)
* SRV
* MX

Edit:
Speaking of the redis plugin earlier (https://coredns.io/explugins/redis/), I think they have done a good job defining almost everything. Ironically enough, as I mention PTRs above not being "easy",  it seems that is the one that has not been defined. :)

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [WIP] Dynamic..." ]
>    Moving the plugin into default coredns could be considered if there is good
>    quality and community interests.
>    Makes sense.
>
>    The data model might be the most challenging parts: what types of records
>    are being covered?
>    Agreed. Even between A vs PTR, it's not quite "as easy" to implement PTRs.
>
>I think the records that "matter" come down to:

this is the wrong way to think about, it should be record agnostic which leaves you with a
double encoding problem. Either you store the RRs' data as binary encoded as hex, or store
every record as an unknown RRs (basically also binary encoded as hex). This also means new
RRs will be supported automatically (by your backend)

Getting data in your data store will mean to compile it into it.

/Miek

--
Miek Gieben

polarathene:
This PR code was never migrated to the created repository for the plugin? Development of it appears to have ceased?

Not needing an additional supporting service/store such as redis or etcd would be nice for updating records. 

A data format such as JSON, YAML or HCL(native Go by hashicorp for their configs) that represented the records would be easy to parse/write by third-parties, even without those, all you'd really need is for the `file` plugin to watch/poll for updates and update/refresh it's records when the file is updated? ([not been possible since 2017](https://github.com/coredns/coredns/issues/522#issuecomment-279171592)? **EDIT:** There is a [reload option](https://coredns.io/plugins/file/) but due to the file being a zone file, requires SOA record to also be updated)

ilyaigpetrov:
What changes are required for this PR to be merged?
Could you, please, provide a step-by-step todo list so me or someone else may pick up and finish this PR.

johnbelamaric:
https://github.com/coredns/coredns/pull/1822#issuecomment-390780769

LucaDev:
@johnbelamaric I'd love to get started building the foundation for this. Are there any plans for the API design yet?

PR #1619, opened 2018-03-18
make CoreDNS DoH Server

miekg:
### 1. What does this pull request do?
Add experimental DoH support.

PR #1829, opened 2018-05-23
Revert pkg/nonwriter changes

### LINKED ISSUE #1826: Self CNAME target resolution (upstream) stopped working
Recently, `upstream` stopped working for self lookups of CNAME targets.

It was working early yesterday.

Integration tests (coredns/ci) started failing today. I verified it also fails on a local build of master, so it's probably not a coredns/ci environment issue.
 
I don't see anything obvious in the commits.  Possibly something changed in an import (that is not version controlled).

miekg:
The DoH work (#1619) made changes to pkg/nonwriter.Writer that in
hindsight were not backwards compatible; it added override for the
LocalAddr() and RemoteAddr(). Instead of rolling back that PR, this PR
reverts those changes and creates a DoHWriter for use in the
https-server.go side of things.

This was only caught in the integration test making this hard to catch,
so we add a upstream_file_test.go that tries (doesn't work yet) to test
this in the unit tests as well. Esp. helpful when 'git bisecting'.

Fixes #1826

### DIFF
--- a/core/dnsserver/https.go
+++ b/core/dnsserver/https.go
@@ -4,8 +4,10 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 
+	"github.com/coredns/coredns/plugin/pkg/nonwriter"
 	"github.com/miekg/dns"
 )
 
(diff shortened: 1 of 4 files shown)


chrisohaver (approved):
Thanks for finding this.

PR #1830, opened 2018-05-23
Add upstream unit test

miekg:
Fix the upstream unit test to actually test it.


miekg:
just little test fix, merging

PR #1831, opened 2018-05-23
Enable new features of Corbot

miekg:
enable alias and add one, so that "/plugin: forward" adds a label
called plugin-forward to the issue.

enable branches that automatically delete merged branches.


fturib (approved):

miekg:
The branches thing might be scary; I'll keep an eye on it and revert asap if I see weird stuff.

Issue #1794, opened 2018-05-11
plugin/proxy: the proxy plugin doesn't iterate upstreams according a policy


mparamonau:
The config example is as following:
.:53 {
    proxy . 9.8.8.8 8.8.8.8 {
        policy sequential
    }
    errors
    log
}

If the first upstream (9.8.8.8) is invalid, the proxy won't try next one (8.8.8.8).

eugen-kleiner-mesh:
I am looking into it.

ekleiner:
Looks like the issue is that health_check is not set. As result all proxies are treated as healthy. This seems like more correct config here:
.:5354 {
    proxy . 9.8.8.8 127.0.0.53 {
        policy sequential
        health_check /:80
    }
}
Is there any other way (supported by proxy plugin) to check if upstream is healthy except probing it via http? As this check seems strange because dns server is not http server.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/proxy:..." ]
>Looks like the issue is that health_check is not set. As result all proxies are treated as healthy. This seems like more correct config here:
>.:5354 {
>    proxy . 9.8.8.8 127.0.0.53 {
>        policy sequential
>        health_check /:80
>    }
>}
>Is there any other way (supported by proxy plugin) to check if upstream is healthy except probing it via http? As this check seems strange because dns server is not http server.

No, not in the proxy plugin. The forward plugin takes a diff. approach here, and 
uses dns healthchecks.

miekg:
is anything happening here.

miekg:
proxy will be deprecated.

PR #1834, opened 2018-05-24
plugin/proxy: Mark upstream broken if i/o timeout, default max_fails=1

eugen-kleiner-mesh:
* Removes the logic not to mark upstream as broken in case of i/o timeout because it is common error in case if upstream dns server is unreachable. It is especially visible in case of sequential policy.
* Makes default max_fails=1. It gives time to try another servers in case of sequential policy and it is supposed to be 1 by default according to README.

Fixes https://github.com/coredns/coredns/issues/1794

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -156,16 +156,8 @@ func SRV(b ServiceBackend, zone string, state request.Request, opt Options) (rec
 		return nil, nil, err
 	}
 
-	type s struct {
-		n string
-		p uint16
-	}
-	dup := make(map[s]bool)
-	type a struct {
-		n string
-		a string
-	}
-	dupAddr := make(map[a]bool)
+	dup := make(map[item]bool)
+	lookup := make(map[string]bool)
 
 	// Looping twice to get the right weight vs priority. This might break because we may drop duplicate SRV records latter on.
 	w := make(map[int]int)
@@ -180,7 +172,6 @@ func SRV(b ServiceBackend, zone string, state request.Request, opt Options) (rec
 		}
 		w[serv.Priority] += weight
 	}
-	lookup := make(map[string]bool)
 	for _, serv := range services {
 		w1 := 100.0 / float64(w[serv.Priority])
 		if serv.Weight == 0 {
@@ -227,20 +218,18 @@ func SRV(b ServiceBackend, zone string, state request.Request, opt Options) (rec
 			if e1 == nil {
 				extra = append(extra, addr...)
 			}
-			// IPv6 lookups here as well? AAAA(zone, state1, nil).
+			// TODO(miek): AAAA as well here.
 
 		case dns.TypeA, dns.TypeAAAA:
 			addr := serv.Host
 			serv.Host = msg.Domain(serv.Key)
 			srv := serv.NewSRV(state.QName(), weight)
 
-			if _, ok := dup[s{srv.Target, srv.Port}]; !ok {
-				dup[s{srv.Target, srv.Port}] = true
+			if ok := isDuplicate(dup, srv.Target, "", srv.Port); !ok {
 				records = append(records, srv)
 			}
 
-			if _, ok := dupAddr[a{srv.Target, addr}]; !ok {
-				dupAddr[a{srv.Target, addr}] = true
+			if ok := isDuplicate(dup, srv.Target, addr, 0); !ok {
 				extra = append(extra, newAddress(serv, srv.Target, ip, what))
 			}
 		}
@@ -255,17 +244,7 @@ func MX(b ServiceBackend, zone string, state request.Request, opt Options) (reco
 		return nil, nil, err
 	}
 
-	type s struct {
-		n string
-		p uint16
-	}
-	dup := make(map[s]bool)
-	type a struct {
-		n string
-		a string
-	}
-	dupAddr := make(map[a]bool)
-
+	dup := make(map[item]bool)
 	lookup := make(map[string]bool)
 	for _, serv := range services {
 		if !serv.Mail {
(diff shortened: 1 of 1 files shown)


plugin/proxy/lookup.go
```diff
@@ -28,7 +28,7 @@ func NewLookupWithOption(hosts []string, opts Options) Proxy {
 		from: ".",
 		HealthCheck: healthcheck.HealthCheck{
 			FailTimeout: 5 * time.Second,
-			MaxFails:    3,
+			MaxFails:    1,
 		},
```
fturib: Here I understand we just do what is said in the readme : max_fails is 1 by default.

plugin/proxy/proxy.go
```diff
@@ -116,19 +115,6 @@ func (p Proxy) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
...
-				// might be long gone...
-				if oe.Timeout() {
-					// Our upstream's upstream is problably messing up, continue with next selected
-					// host - which my be the *same* one as we don't set any uh.Fails.
-					continue
-				}
-			}
-
 			timeout := host.FailTimeout
 			if timeout == 0 {
```
fturib: For this change, I am hesitant.
I do not know why this code was introduced.
We need validation of @miekg  here.

fturib (commented):
We need that Miek review the removal of the ISC/"timeout" special handling case.

Do we fit exactly in this issue when there upstream point to nowhere ?

fturib:
@miekg : can you review here the lines removed in proxy ? does it needs an option in "Proxy" stanza to let the user choose how to behave in case of those errors around timeout ?

plugin/proxy/proxy.go
fturib: After talk and review, if we remove these lines, the impact will be that this upstream will be "failed" during the "fail timeout" period.

- That will allow to choose the next Upstream in case of "sequential" policy
- That will not hit other use case as anyway another Upstream will be choose.

So maybe that code was introduced before the mechanism of Down/failtimeout

fturib:
@miekg : after second though and review with @ekleiner, I think it is safe to remove these lines and let the healhcheck mechanism manage the retries in case of timeout.

What do you think ?

fturib (approved):
After talk with Eugen, the change l- which is also a simplification - has a valid logic.
The natural logic around Healthcheck is taking care of the timeout use case.

plugin/proxy/proxy.go
miekg: You can't rip this out; and this piece of code shows that we should move to *forward* as the default proxy impl. As we attribute upstream dead nameservers to *our* local proxy. The forward plugin fires off a healthcheck in this case to re-asses (even there is can be a bit smarter).

fturib: @miekg : what is the purpose of this code ? 
I mean : we want to avoid the healthcheck if there is a timeout ? 
if YES, is that still valid after the fix #1441 ?

What I mean is if we get rid of this code:
-  a timeout will be considered as a "fail" for 2 sec, 
until either there is a healtcheck. 
- if healhcheck is not defined, or healtcheck is ok, this upstream will be available again for next Select().

fturib: Maybe I misunderstand the point of the logic around this timeout. Let me know.

miekg: Is there a test you can add that shows the issue which will be fixed by this PR?

fturib: Good point. Let's make a UT (or integration test) to validate before and after. Thanks

fturib:
@ekleiner : is there a way to create a UT that shows behavior in case of timeout is the same without this piece of code ? (see last comment in review)

eugen-kleiner-mesh:
I will try to create some test related to this changes.

eugen-kleiner-mesh:
I have added test which emulates https://github.com/coredns/coredns/issues/1794 and fails without the change.

fturib:
LGTM, and now we have a simple way to have create timing-out servers. Thanks!

miekg:
So, with this PR checked and this corefile:
~~~
. {
    proxy . 8.8.8.8
    log
    errors
}
~~~
It marks the upstream as dead; which is wrong.
~~~
[::1]:57641 - [13/Jun/2018:08:23:00 +0100] 5598 "ANY IN isc.org. udp 37 false 4096" SERVFAIL qr,rd 37 5.000874863s
2018/06/13 08:23:00 [ERROR] 0 isc.org. ANY: unreachable backend: no upstream host
127.0.0.1:33500 - [13/Jun/2018:08:23:01 +0100] 5598 "ANY IN isc.org. udp 37 false 4096" SERVFAIL qr,rd 37 5.001065326s
2018/06/13 08:23:01 [ERROR] 0 isc.org. ANY: unreachable backend: no upstream host
[::1]:42304 - [13/Jun/2018:08:23:04 +0100] 45927 "A IN www.miek.nl. udp 41 false 4096" SERVFAIL qr,rd 41 106.228µs
2018/06/13 08:23:04 [ERROR] 0 www.miek.nl. A: unreachable backend: no upstream host
[::1]:50510 - [13/Jun/2018:08:23:05 +0100] 7786 "A IN www.miek.nl. udp 41 false 4096" SERVFAIL qr,rd 41 85.494µs
2018/06/13 08:23:05 [ERROR] 0 www.miek.nl. A: unreachable backend: no upstream host
[::1]:40139 - [13/Jun/2018:08:23:07 +0100] 40830 "A IN www.miek.nl. udp 41 false 4096" NOERROR qr,rd,ra,ad 100 55.080316ms
~~~

While running from master, I get a response for www.miek.nl because the upstream isn't marked as dead

eugen-kleiner-mesh:
@miekg , could you provide more details. Because this pr and the mentioned config (I changed only listening port) works fine for me in case if 8.8.8.8 is reachable.
```
 .:5354 {
    proxy . 8.8.8.8
    log
    errors
}
```
```
dig ics.org -t ANY @127.0.0.1 -p 5354
dig www.miek.nl -t A @127.0.0.1 -p 5354
```
```
127.0.0.1:39751 - [13/Jun/2018:13:48:06 +0300] 3027 "ANY IN ics.org. udp 37 false 4096" NOERROR qr,rd,ra 636 213.816963ms
127.0.0.1:46026 - [13/Jun/2018:13:48:46 +0300] 55887 "A IN www.miek.nl. udp 41 false 4096" NOERROR qr,rd,ra,ad 100 536.871779ms
```

eugen-kleiner-mesh:
Now I got it. Making many digs in a cycle reproduced the issue.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/proxy:..." ]
>Closed #1834.

We should probably invest time in fully deprecating proxy in favor of forward.

PR #1839, opened 2018-05-29
Use _<something> to signal go language feature

miekg:
Move all underscores to hyphens. I started with server_grpc.go and
server_tls.go and then figured, let's do everything.

So _test.go indicates testing and _linux and friends (not used)
indicates compile for specific architectures.

Left _generate.go to indicate something that is run by go generate.
(messed the git stuff up, so that became a delete and add :P) But git is smart enough to not
show this.


grobie:
:man_shrugging: I'm used to underscores from most projects and find mixed hypens + underscores rather confusing to be honest. No strong objections if you're passionate about it.

johnbelamaric:
Ditto.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Use _<somethi..." ]
>:man_shrugging: I'm used to underscores from most projects and find mixed hypens + underscores rather confusing to be honest. No strong objections if you're passionate about it.

no strong feelings here, but I do dislike the inconsistency, all underscores 
sgtm and also be enforced with a .presubmit (to save a little time on future 
reviews)

PR #1843, opened 2018-06-01
Docker: drop alpine

miekg:
Create a multistage docker build image that uses debian to install certs
and then create the final image by using FROM: scratch. This creates a
(slightly) smaller images and drops busybox and alpine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,12 @@
-FROM alpine:latest
+FROM debian:stable-slim
 
-# Only need ca-certificates & openssl if want to use DNS over TLS (RFC 7858).
-RUN apk --no-cache add bind-tools ca-certificates openssl && update-ca-certificates
+RUN apt-get update && apt-get -uy upgrade
+RUN apt-get -y install ca-certificates && update-ca-certificates
 
+FROM scratch
+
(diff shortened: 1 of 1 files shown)


yongtang:
How about bind-tools in alpine? dnsutils in Debian?

nberlee:
`/etc/ssl/certs` is the minimum needed for go root ca checking. No need for the parent `/etc/ssl` directories and `/etc/ca-certificates`

miekg:
thanks @nberlee made that change

@yongtang does it add much to have those tools in the docker image? For that to work we need to copy /usr, /lib and maybe more (that added up to 200M in a quick test I did)

yongtang:
@miekg I am actually fine not to add those tools. I saw `bind-tools` was installed before with `alpine`, not sure the initial usage, so was asking about it.

miekg:
@yongtang I think this change also allows us to add certs to all other images instead of only amd64, because the second stage only uses COPY instead of running actually commands, also the docker makefile part is now broken I think

stp-ip:
Sidenote. Debugging can be done attaching to the process id. This will most likely become much more common in Kubernetes too.

PR #1845, opened 2018-06-01
presubmit: Check errorf as well

miekg:
Uppercase all these test errors as well. And extend the presubmit to
check for these in the future. Also do a slightly smarter grep to only
get t.<something>. as (because dump regexp) this also grep over non test
files.


yongtang (approved):
LGTM

PR #1846, opened 2018-06-01
Extend presubmit to covert test and request

miekg:
Fix the casing *and* fix use of context as we were still referencing
the non-std lib context - now wondering how this could have worked...


greenpau (approved):
LGTM

PR #1850, opened 2018-06-02
Fix presubmit

miekg:
Uppercase these test cases as well.


miekg:
If travis comes back Ok, I will merge.

Issue #1849, opened 2018-06-01
plugin/Forward : ability to configure the dialTimeout 


fturib:
Currently we are using CoreDNS with forward and upstream to cloud AWS (using dns over tls)
Depending the network used we may have very slow connection and need to extend the dial timeout.
(right now issue with network over wifi..)

I cannot say right now what would be the timeout to be defined, and if this dialTimeout would be tunable in our configuration setup.

There is a similar issue here about readTimeout : #1678 that has been solved without adding the option in stanza.

We are specifically stuck with the dialTimeout, but should we rather propose a global "timeout" in the stanza that would cover ALL timeouts, in order to simplify the configuration for everyone ?

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin/Forward : ..." ]
>Currently we are using CoreDNS with forward and upstream to cloud AWS (using dns over tls)
>Depending the network used we may have very slow connection and need to extend the dial timeout.
>(right now issue with network over wifi..)
>
>I cannot say right now what would be the timeout to be defined, and if this dialTimeout would be tunable in our configuration setup.
>
>There is a similar issue here about readTimeout : #1678 that has been solved without adding the option in stanza.
>
>We are specifically stuck with the dialTimeout, but should we rather propose a global "timeout" in the stanza that would cover ALL timeouts, in order to simplify the configuration for everyone ?

Can we not do the same thing as we did with readTimeout?

fturib:
Do you mean by increasing/decreasing the dial timeout based on precedent tentatives ?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/Forwar..." ]
>Do you mean by increasing/decreasing the dial timeout based on precedent tentatives ?

yes (we use avg in the read/write timeout)

fturib:
NOTE: looks like there is an issue in the underlying library that manage the Dial. Which make useless for now to change this dialTimeout.
I created the issue https://github.com/miekg/dns/issues/689 to clarify

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/Forwar..." ]
>NOTE: looks like there is an issue in the underlying library that mange the Dial. I created the issue https://github.com/miekg/dns/issues/689 to clarify

Ack, saw it. I might be time to create a dns.Client2 to create some sanity 
there...

PR #1852, opened 2018-06-04
Plugin/Forward - autotune the dialTimeout for connection

fturib:

- implement an auto-tunable dialTimeout for fallback in a way very similar to readTimeout.
- timeout will start at 4sec and can vary from 100ms to 30sec depending success of former Dial of the same fallback.Proxy.

NOTE: this change depends of the merge of https://github.com/miekg/dns/pull/691

#1849 

- good question. I do not see anything about timeout in README. Should we add one ?

### DIFF
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -16,21 +16,65 @@ import (
 	"github.com/miekg/dns"
 )
 
-func (p *Proxy) readTimeout() time.Duration {
-	rtt := time.Duration(atomic.LoadInt64(&p.avgRtt))
+// utility function to auto-tune timeout values
+// average observed time is moved towards the last observed delay moderated by a weight
+// next timeout to use will be the double of the computed average, limited by min and max frame.
+func limitTimeout(currentAvg *int64, minValue time.Duration, maxValue time.Duration) time.Duration {
+	rt := time.Duration(atomic.LoadInt64(currentAvg))
+	if rt < minValue {
+		return minValue
+	}
+	if rt < maxValue/2 {
+		return 2 * rt
+	}
+	return maxValue
+}
 
-	if rtt < minTimeout {
-		return minTimeout
+func averageTimeout(currentAvg *int64, observedDuration time.Duration, weight int64) {
+	dt := time.Duration(atomic.LoadInt64(currentAvg))
+	atomic.AddInt64(currentAvg, int64(observedDuration-dt)/weight)
+}
+
+func (t *transport) dialTimeout() time.Duration {
+	return limitTimeout(&t.avgDialTime, minDialTimeout, maxDialTimeout)
+}
+
+func (t *transport) updateDialTimeout(newDialTime time.Duration) {
+	averageTimeout(&t.avgDialTime, newDialTime, cumulativeAvgWeight)
+}
+
+// Dial dials the address configured in transport, potentially reusing a connection or creating a new one.
+func (t *transport) Dial(proto string) (*dns.Conn, bool, error) {
+	// If tls has been configured; use it.
+	if t.tlsConfig != nil {
+		proto = "tcp-tls"
 	}
-	if rtt < maxTimeout/2 {
-		return 2 * rtt
+
+	t.dial <- proto
+	c := <-t.ret
+
+	if c != nil {
+		return c, true, nil
 	}
-	return maxTimeout
+
+	reqTime := time.Now()
+	timeout := t.dialTimeout()
+	if proto == "tcp-tls" {
+		conn, err := dns.DialTimeoutWithTLS("tcp", t.addr, t.tlsConfig, timeout)
+		t.updateDialTimeout(time.Since(reqTime))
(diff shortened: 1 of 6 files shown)


fturib:
Looks like I have issues with gofmt version on my local machine ... or maybe go version.
working on it.

plugin/forward/persistent.go
```diff
@@ -6,6 +6,8 @@ import (
 	"sort"
 	"time"
 
+	"sync/atomic"
```
miekg: fmt

plugin/forward/persistent.go
```diff
@@ -26,17 +28,21 @@ type transport struct {
 	yield chan *dns.Conn
 	ret   chan *dns.Conn
 	stop  chan bool
+
+	avgDialTime int64 // kind of average time of dial time
```
miekg: this is now an atomic and must come first in the structure

plugin/forward/persistent.go
```diff
@@ -26,17 +28,21 @@ type transport struct {
...
+
```
miekg: superfluous newline?

plugin/forward/persistent.go
miekg: I see avgRtt makes the same mistake

plugin/forward/persistent.go
```diff
@@ -26,17 +28,21 @@ type transport struct {
...
-		ret:    make(chan *dns.Conn),
-		stop:   make(chan bool),
+		conns:       make(map[string][]*persistConn),
+		expire:      defaultExpire,
+		addr:        addr,
+		dial:        make(chan string),
+		yield:       make(chan *dns.Conn),
+		ret:         make(chan *dns.Conn),
+		stop:        make(chan bool),
+		avgDialTime: 0,
```
miekg: the default is 0?

plugin/forward/persistent.go
```diff
@@ -141,6 +147,20 @@ func (t *transport) cleanup(all bool) {
 	}
 }
 
+func (t *transport) dialTimeout() time.Duration {
```
miekg: Can this be put in connect.go next to the avg calculation? That will probably flag some similarities that could be factored out a bit.

plugin/forward/persistent.go
fturib: hum.. good point. Yes I want to be 0 to start with the right initialDialTime. But I should take care when computing the avg...

miekg:
This averaging should probably be documented in the README.md

plugin/forward/persistent.go
rdrozhdzh: IMO, it is better to initialize it once here with defaultDialTimeout and don't check it in every call to `dialTimeout()`

plugin/forward/persistent.go
rdrozhdzh: I'm not sure that this average timeout at transport level would work fine for all upstreams, all protocols. 
The UDP dial will work almost immediately, since (I believe) it doesn't communicate to peer during `Dial()`, TCP dial communicates to peer, so it will be slower, TLS dial is even more slower (TCP + TLS handshake). 
Also different proxies can have different routes which can also affect dial time.

plugin/forward/persistent.go
```diff
@@ -141,6 +147,20 @@ func (t *transport) cleanup(all bool) {
...
+	rtt := time.Duration(atomic.LoadInt64(&t.avgDialTime))
```
rdrozhdzh: `rtt` means Round Trip Time, not sure if this name is suitable here. Consider `dt` (dial time) or smth else

rdrozhdzh:
hm, why Plugin/Fallback in the PR title? Shouldn't it be Plugin/Forward?

fturib:
you are right. My bad.

plugin/forward/persistent.go
fturib: @rdrozhdzh : this average is per Forward.Proxy (each Forward.Proxy maintain its own Transport object). I mean each computation is dedicated for one upstream. The first dialTimeout is the default, and then, if the line fall because expireTime or else, the next retry Dial will use an average (sort-of) of the last real time needed to Dial.

or did I miss something in your comment ?

rdrozhdzh: oh, sorry, you are right.

fturib:
NOTE: this PR make sense only if we also update the library Miek/dns with the version that includes that PR: https://github.com/miekg/dns/pull/691

plugin/forward/connect.go
```diff
@@ -16,21 +16,65 @@ import (
 	"github.com/miekg/dns"
 )
 
-func (p *Proxy) readTimeout() time.Duration {
-	rtt := time.Duration(atomic.LoadInt64(&p.avgRtt))
+// utility function to auto-tune timeout values
```
miekg: limitTimout is a ...

miekg (approved):
please also re-run the tests.

PR #1854, opened 2018-06-04
Don't crash during test when message is nil

chrisohaver:

Fail test gracefully instead of crashing if message being validated is nil.


miekg:
I rather not have this in these generic helpers. Why and where are these given a nil msg?

chrisohaver:
You don't think its better to fail the test gracefully?

miekg:
Which tests? You're amending the helper functions not the tests that call/use these

chrisohaver:
OK

Issue #1841, opened 2018-05-31
Standardize Response Duration


makinje16:
Sometimes response duration comes in nano seconds and sometimes milliseconds. Are we able to pick a standard for the response duration in order to be more clear and needing to have conversions for them.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] Standardize Respo..." ]
>Sometimes response duration comes in nano seconds and sometimes milliseconds. Are we able to pick a standard for the response duration in order to be more clear and needing to have conversions for them.

All in-tree metrics export seconds:

% grep -ir observe
forward/connect.go:	RequestDuration.WithLabelValues(p.addr).Observe(time.Since(start).Seconds())
metrics/vars/report.go:	RequestDuration.WithLabelValues(server, zone).Observe(time.Since(start).Seconds())
metrics/vars/report.go:	ResponseSize.WithLabelValues(server, zone, net).Observe(float64(size))
metrics/vars/report.go:	RequestSize.WithLabelValues(server, zone, net).Observe(float64(req.Len()))
proxy/proxy.go:		RequestDuration.WithLabelValues(metrics.WithServer(ctx), state.Proto(), upstream.Exchanger().Protocol(), familyToString(state.Family()), host.Name).Observe(time.Since(start).Seconds())
health/overloaded.go:	HealthDuration.Observe(timeout.Seconds())
health/overloaded.go:	HealthDuration.Observe(time.Since(start).Seconds())



/Miek

--
Miek Gieben

makinje16:
Sorry @miekg I should be more specific. In the logs plugin, when unparsed, the response duration field can be in ms or nanoseconds. Are we able to standardize that part of the log plugin?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Standardize R..." ]
>Sorry @miekg I should be more specific. In the logs plugin, when unparsed, the response duration field can be in ms or nanoseconds. Are we able to standardize that part of the log plugin?

ah yeah. I can see that being annoying, but why aren't you using metrics for 
that?

Should probably make that seconds as well, just for overall consistency; PR 
welcome

PR #1860, opened 2018-06-06
Normalizing Response Duration in Log Plugin

makinje16:
The log plugins response duration is now standardized to seconds to keep in line with the metrics plugin.

[Issue #1841 ](https://github.com/coredns/coredns/issues/1841)

no documentation changes need to be done

### DIFF
--- a/plugin/log/setup_test.go
+++ b/plugin/log/setup_test.go
@@ -65,6 +65,7 @@ func TestLogParse(t *testing.T) {
 		}}},
 		{`log example.org {
 			class denial
+			timeunit ms
 		}`, false, []Rule{{
 			NameScope: "example.org.",
 			Format:    CommonLogFormat,
@@ -72,6 +73,7 @@ func TestLogParse(t *testing.T) {
 		}}},
 		{`log {
 			class denial
+			timeunit fvdw
 		}`, false, []Rule{{
 			NameScope: ".",
 			Format:    CommonLogFormat,
@@ -79,6 +81,7 @@ func TestLogParse(t *testing.T) {
 		}}},
 		{`log {
 			class denial error
+			timeunit micro
 		}`, false, []Rule{{
 			NameScope: ".",
 			Format:    CommonLogFormat,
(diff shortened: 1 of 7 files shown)


miekg (commented):
Thanks!

I think we should just default to seconds to be in line with the metrics, also removes the need to make this configurable (less knobs is better)

plugin/log/setup_test.go
```diff
@@ -19,41 +19,49 @@ func TestLogParse(t *testing.T) {
 			NameScope: ".",
 			Format:    DefaultLogFormat,
 			Class:     map[response.Class]bool{response.All: true},
+			TimeUnits: "",
 		}}},
-		{`log example.org`, false, []Rule{{
+		{`log example.org `, false, []Rule{{
```
pmoroney: Not sure I understand it, but why was a space added here? And why do we expect it to have TimeUnits of `"ms"`?

makinje16: sloppy checking on my part. Will fix

makinje16:
I updated to just have seconds displayed to make it easier and less knobs as @miekg said before.

plugin/pkg/replacer/replacer.go
```diff
@@ -55,7 +55,7 @@ func New(r *dns.Msg, rr *dnstest.Recorder, emptyValue string) Replacer {
 		}
 		rep.replacements["{rcode}"] = rcode
 		rep.replacements["{rsize}"] = strconv.Itoa(rr.Len)
-		rep.replacements["{duration}"] = time.Since(rr.Start).String()
+		rep.replacements["{duration}"] = NormalizeTime(time.Since(rr.Start))
```
miekg: `time.Since(rr.start).Seconds().String()` will probably do this as well. Saves a strconv.

Can you add a short test for this in replace_test.go as well? Thanks!

makinje16: .Seconds() returns a float64 which has no member function .String() so I think the strconv is necessary @miekg

miekg: ah of course. Then just inline NormalizeTime here, i.e. just to the strconv.FormatFloat

miekg (approved):

PR #1867, opened 2018-06-11
Remove dnsutil.Dedup

miekg:
Remove the code and remove the call in etcd and kubernetes handlers.
This does mean we should not add dups in the first place, which means
adding another map in backend_lookup to prevent dups from begin added.

This should cut doesn on the allocation because dnsutil.Dedup is very
expensive by converting everything to strings, we avoid doing that now.

`(s *Service) RR()` is not used anywhere - was for the debug queries - long since dropped. Killed that off as well.


miekg:
CI test is wrong, garbage in garbage out, there is noone to save you now:
~~~
--- FAIL: TestUpstreamLoopBreak (6.43s)
    --- FAIL: TestUpstreamLoopBreak/upriver.test-2.svc.cluster.local._A (0.31s)
    	helpers.go:123: Answer for "upriver.test-2.svc.cluster.local." contained 11 results, 2 expected
    	helpers.go:301: ;; opcode: QUERY, status: NOERROR, id: 37417
    		;; flags: rd; QUERY: 1, ANSWER: 11, AUTHORITY: 0, ADDITIONAL: 0
    		
    		;; QUESTION SECTION:
    		;upriver.test-2.svc.cluster.local.	IN	 A
    		
    		;; ANSWER SECTION:
    		up.river.local.	5	IN	CNAME	upriver.test-2.svc.cluster.local.
    		up.river.local.	5	IN	CNAME	upriver.test-2.svc.cluster.local.
    		up.river.local.	5	IN	CNAME	upriver.test-2.svc.cluster.local.
    		up.river.local.	5	IN	CNAME	upriver.test-2.svc.cluster.local.
    		up.river.local.	5	IN	CNAME	upriver.test-2.svc.cluster.local.
    		upriver.test-2.svc.cluster.local.	5	IN	CNAME	up.river.local.
    		upriver.test-2.svc.cluster.local.	5	IN	CNAME	up.river.local.
    		upriver.test-2.svc.cluster.local.	5	IN	CNAME	up.river.local.
    		upriver.test-2.svc.cluster.local.	5	IN	CNAME	up.river.local.
    		upriver.test-2.svc.cluster.local.	5	IN	CNAME	up.river.local.
    		upriver.test-2.svc.cluster.local.	5	IN	CNAME	up.river.local.
~~~
Should potentially be fixed in the *template* plugin though.

chrisohaver:
> CI test is wrong, garbage in garbage out...

No, I  think the expected value in the CI test is correct.  I think the duplicates are invalid, and should not be added. They violate DNS spec, right?  

> there is noone to save you now:

Not sure what you mean by this.

> Should potentially be fixed in the template plugin though.

If possible, fixed in the upstream shared code, so it gets fixed for all plugins that use it.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Remove dnsuti..." ]
>> CI test is wrong, garbage in garbage out...
>
>No, I  think the expected value in the CI test is correct.  I think the duplicates are invalid, and should not be added. They violate DNS spec, right?

Yes.

>> Should potentially be fixed in the template plugin though.
>
>If possible, fixed in the upstream shared code, so it gets fixed for all plugins that use it.

There is no shared code. If we deem this dangerous template should filter it 
out; in only works in *this* setting because the kubernetes plugin was dedupping 
it...

/Miek

--
Miek Gieben

chrisohaver:
I can make the CI test not care about the number of records it gets back, just that it gets some records back. The test is intended to verify that we don't loop forever resolving a circular CNAME chain - not necessarily to validate that we don't have duplicates in responses.

chrisohaver (approved):

PR #1869, opened 2018-06-12
Follow up on #1867 : use one map in backend_lookup

miekg:
Reduce map usage and combine two maps into one; add isDuplicate that
tests and adds to make things more readable (hopefully).

Didn't change the A and AAAA map usage as those already just use a `map[string]bool`


yongtang (approved):
LGTM

PR #1877, opened 2018-06-14
Plugin/Rewrite - review documentation of "append"

fturib:
Fix mistake in the  documentation of plugin rewrite
from the code, the command 'append' is adding the option ONLY if that option is not already matching existing option in the OPT record.
NOTE: the "matching" part encoded make sense, and then it is better to fix the documentation than the code.

NONE

Done in this PR.

### DIFF
--- a/plugin/rewrite/README.md
+++ b/plugin/rewrite/README.md
@@ -180,7 +180,7 @@ rewrite [continue|stop] name regex STRING STRING answer name STRING STRING
 Using FIELD edns0, you can set, append, or replace specific EDNS0 options on the request.
 
 * `replace` will modify any matching (what that means may vary based on EDNS0 type) option with the specified option
-* `append` will add the option regardless of what options already exist
+* `append` will add the option if does not find any matching option
 * `set` will modify a matching option or add one if none is found
 
 Currently supported are `EDNS0_LOCAL`, `EDNS0_NSID` and `EDNS0_SUBNET`.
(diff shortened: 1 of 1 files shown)


plugin/rewrite/README.md
```diff
@@ -180,7 +180,7 @@ rewrite [continue|stop] name regex STRING STRING answer name STRING STRING
 Using FIELD edns0, you can set, append, or replace specific EDNS0 options on the request.
 
 * `replace` will modify any matching (what that means may vary based on EDNS0 type) option with the specified option
-* `append` will add the option regardless of what options already exist
+* `append` will add the option if does not find any matching option
```
miekg: I find both the old and new ambiguous. `append` as the EDNS option regardless of what option are already in the section? I.e. you can get duplicates? Is that what this is trying to say?

chrisohaver: do you mean...

`append` will add the option if it does not already exist

chrisohaver: or perhaps even more precise...

`append` will add the option *only* if it does not already exist

fturib: @miekg : what is ambiguus in the new definition ?
@chrisohaver : is used in purpose 'matching' because it is defined in 'replace'.
=> (what that means may vary based on EDNS0 type)

we are doing the same 'matching' test for all those operations.

chrisohaver: OK, how about...

`append` will add the option only if no matching option exists

chrisohaver: also, IMO, the way that "matching" is defined is awkward and disrupts the sentence.  It would be better to say something like ...

 * `replace` will modify any "matching" option with the specified option.  The criteria for "matching" varies based on EDNS0 type.

fturib: @chrisohaver : Thanks. I will update it.

chrisohaver (approved):

Issue #1876, opened 2018-06-13
plugin/proxy: REFUSED from backend is surfaced as SERVFAIL


ahmetb:
I'm using CoreDNS to run an experimental public DNS nameserver.

I'm using `proxy` plugin with `protocol grpc insecure` option. In my gRPC backend, I respond to queries that I don't support with RCODE=Refused.

This response is recorded by the `log` plugin as follows:

`
10.8.0.1:36131 - [13/Jun/2018:18:53:11 +0000] 56254 "MX IN mydomain.redacted. udp 110 false 512" REFUSED qr,aa,cd 110 610.392µs
`

When I **indirectly** query the nameserver with dig, I get RCODE=SERVFAIL:

```
# dig MX mydomain.redacted.
...
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 2435
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1
...
;; Query time: 136 msec
;; SERVER: 2401:fa00:fa::1#53(2401:fa00:fa::1)
```

when I **directly** query the nameserver, I get rcode=REFUSED:

```
$ dig MX mydomain.redacted. @ns1.ns.mydomain.redacted

...;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: REFUSED, id: 8373
;; flags: qr aa rd; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0
;; WARNING: recursion requested but not available
...
;; Query time: 45 msec
;; SERVER: 35.225.x.x#53(35.225.x.x)
;; WHEN: Wed Jun 13 11:58:11 PDT 2018
;; MSG SIZE  rcvd: 110
```
(based on `SERVER:` this directly hits my coredns endpoint)

Could this be potentially related to CoreDNS, or is it my ISP/intermediate resolvers?

miekg:
https://github.com/coredns/coredns/pull/1878
add explicit tests for this (both in proxy and forward) and that echos back the origin RCODE

ahmetb:
I suspect the issue here might not be coredns. Since I'm on corporate network, the intermediate DNS server might be caching REFUSED this as SERVFAIL. Let me know if you want a repro.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/proxy:..." ]
[...]

PR #1878, opened 2018-06-14
plugin/forward: add REFUSED test

miekg:
add a test to see if we copy the rcode correctly. Some minor cleanup in
import ordering and renaming NewUpstream to New as we already are in the
upstream package.


miekg:
See #1876

chrisohaver (approved):
lgtm

PR #1882, opened 2018-06-19
plugin/cache: remove goroutine from prefetch handler

miekg:
We used to start the prefetch in a goroutine, but because the responsewriter is closed
by coredns when this handler returns, this leads to a crash when any other plugin
accesses the (now partly closed) response writer.

Optionally we can create a new ResponseWriter that fakes a few methods
to make it work, but this quickly becomes a larger refactor.


grobie:
While this avoids the possibility of causing a panic by accessing closed/removed state, it also breaks the prefetch behavior. The overall longtail latency will actually be worse than not using prefetch, so I'd expect a clear warning in the cache readme explaining the impact.

In order to reduce our 90th percentile, breaking prefetch isn't an option for us, so I'll look into an alternative solution.

PR #1884, opened 2018-06-19
Ensure cache.ResponseWriter can be used asynchronously during prefetch

grobie:
The default dns.Response implementation of a dns.ResponseWriter will
panic if RemoteAddr() is called after the connection to the client has
been closed already. The current cache implementation doesn't create a
new request+responsewriter during an asynchronous prefetch, but
piggybacks on the request triggering the prefetch.

This change copies the RemoteAddr first, so that it's safe to use it
later during the actual prefetch request.

A better implementation would be to completely decouple the prefetch
request from the client triggering a request.

Resolves #1882.

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -4,6 +4,7 @@ package cache
 import (
 	"encoding/binary"
 	"hash/fnv"
+	"net"
 	"time"
 
 	"github.com/coredns/coredns/plugin"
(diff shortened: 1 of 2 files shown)


plugin/cache/cache.go
```diff
@@ -105,7 +106,40 @@ type ResponseWriter struct {
...
 
-	prefetch bool // When true write nothing back to the client.
+	prefetch   bool // When true write nothing back to the client.
+	remoteAddr net.Addr
+}
+
+// newPrefetchResponseWriter returns a Cache ResponseWriter to be used in
+// prefetch requests. It ensures RemoteAddr() can be called even after the
+// original connetion has already been closed.
+func newPrefetchResponseWriter(server string, state request.Request, c *Cache) *ResponseWriter {
```
miekg, line 116: I think it is better to have a new type PrefetchResponseWriter that does the prefetch, the prefetch bool can then also be dropped from the original ResponseWriter.

I looked at bit; the only downside is that WriteMsg needs to be duplicated. WDYT?

grobie: I'd love to refactor plugin/cache at some point and decouple responsibilities. The prefetch bool is a great indicator that multiple concerns are handled by a single type. There is more to it though, the `Set()` method should really be defined on cache and not the responsewriter.

At the moment my main goal is to get a stable version deployed, so refactoring the existing package structure is a second priority. I'm happy to look into it end of the week.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Ensure cache...." ]
>grobie commented on this pull request.
>I'd love to refactor plugin/cache at some point and decouple responsibilities. The prefetch bool is a great indicator that multiple concerns are handled by a single type. There is more to it though, the `Set()` method should really be defined on cache and not the responsewriter.
>
>At the moment my main goal is to get a stable version deployed, so refactoring the existing package structure is a second priority. I'm happy to look into it end of the week.

Ok, cool. That works me for.

PR #1885, opened 2018-06-19
Rename forward metrics socket_count_total to sockets_open

grobie:
The prometheus naming convention states only counters should have a
`_total` suffix, so that gagues and counters can be easily
distinguished.

Most of the metrics don't follow the idiomatic naming of Prometheus, e.g. using `_count_total` suffixes.


miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] Rename forward me..." ]
>The prometheus naming convention states only counters should have a
>`_total` suffix, so that gagues and counters can be easily
>distinguished.

Do we have any more of these - rather do them all in one go.

grobie:
@miekg I checked and this seems to be the only gauge with the `_total` suffix reserved for counters.

PR #1886, opened 2018-06-19
plugin/template: fix CNAME upstream handling

chrisohaver:

Fixes a couple of bugs with template CNAME lookup handling...
1. if `upstream` is not set, we'd return a SERVFAIL if a CNAME was in the response.
2. CNAME target lookup was being performed on CNAME queries (not just A and AAAA queries).  Expected behavior is that a query for CNAME records only returns the CNAME records.  Lookup and adding of A/AAAA records should only be done for A/AAAA queries.

#1887

none


chrisohaver:
oops - i need to remove that extraneous unrelated file...

greenpau:
@chrisohaver , any unit test improvements?

plugin/template/template.go
```diff
@@ -84,8 +84,8 @@ func (h Handler) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 				return dns.RcodeServerFailure, err
 			}
 			msg.Answer = append(msg.Answer, rr)
-			if rr.Header().Rrtype == dns.TypeCNAME {
-				up, _ := template.upstream.Lookup(state, rr.(*dns.CNAME).Target, dns.TypeA)
+			if template.upstream != nil && (state.QType() == dns.TypeA || state.QType() == dns.TypeAAAA) && rr.Header().Rrtype == dns.TypeCNAME {
```
greenpau, line 87: @chrisohaver , please add unit tests.

greenpau (changes requested):

greenpau:
@chrisohaver , please let me know when you are done with this PR. I will approve.

chrisohaver:
I could not work out how to test upstream function with "unit" tests, so added an integration test.

greenpau:
> I could not work out how to test upstream function with "unit" tests, so added an integration test.

@miekg , I am also not sure about this one. The integration test is great, but unit test is better, imho. What is your take on this?

chrisohaver:
To be clear, the integration test is in `/test`, where other similar integration tests reside.  Most of these tests run a server instance with a given corefile - hence they are technically integration tests, not strictly unit tests.

There is a similar test that tests CNAME upstream lookups for the `file` plugin (`/test/file_upstream_test.go`).

There _may_ be a way to test upstream lookups without running a server instance... but I don't see a pressing reason to do so.

chrisohaver:
> There may be a _way_ to test upstream lookups without running a server instance.

Actually - I could do it by creating an interface for `upstream`, `upstreamer`, and mocking it out with a something that implements `upstreamer`...  this would require some refactoring in `template`.

IMO not worth the trouble unless people feel really strongly about it.

miekg (approved):
there is test, the only downside is that codecov doesn't count it correctly

greenpau:
> there is test, the only downside is that codecov doesn't count it correctly

@miekg , 👍 , will remember this for the future. Thought it is better to ask.

greenpau:
@chrisohaver , do you mind squashing? (or perhaps it is done automatically)

miekg:
@greenpau we squash on commit (which should a project wide setting). Don't care that much on how the exact commit message looks.

PR #1888, opened 2018-06-20
remove deprecated code

miekg:
This removes:
* reverse plugin from plugin.cfg
* https_google option from proxy
* the -log flag

The -log was not described in README.md anyway.


miekg:
just the deletes we've announced. Merging

PR #1889, opened 2018-06-20
plugin/forward: Increase minimum read timeout to 200ms

grobie:
After several experiments at SoundCloud we found that the current
minimum read timeout of 10ms is too low. A single request against a
slow/unavailable authoritative server can cause all TCP connections to
get closed. We record a 50th percentile forward/proxy latency of <5ms,
and a 99th percentile latency of 60ms. Using a minimum timeout of 200ms
seems to be a fair trade-off between avoiding unnecessary high
connection churn and reacting to upstream failures in a timely manner.

This change also renames hcDuration to hcInterval to reflect its usage,
and removes the duplicated timeout constant to make code comprehension
easier.

### DIFF
--- a/plugin/forward/proxy.go
+++ b/plugin/forward/proxy.go
@@ -34,7 +34,7 @@ func NewProxy(addr string, tlsConfig *tls.Config) *Proxy {
 		fails:     0,
 		probe:     up.New(),
 		transport: newTransport(addr, tlsConfig),
-		avgRtt:    int64(timeout / 2),
+		avgRtt:    int64(maxTimeout / 2),
 	}
 	p.client = dnsClient(tlsConfig)
 	runtime.SetFinalizer(p, (*Proxy).finalizer)
@@ -106,8 +106,7 @@ func (p *Proxy) start(duration time.Duration) {
 }
 
 const (
-	timeout    = 2 * time.Second
 	maxTimeout = 2 * time.Second
-	minTimeout = 10 * time.Millisecond
-	hcDuration = 500 * time.Millisecond
+	minTimeout = 200 * time.Millisecond
+	hcInterval = 500 * time.Millisecond
 )
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -83,8 +83,9 @@ Also note the TLS config is "global" for the whole forwarding proxy if you need 
 `tls-name` for different upstreams you're out of luck.
 
 On each endpoint, the timeouts of the communication are set by default and automatically tuned depending early results.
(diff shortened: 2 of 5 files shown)


rdrozhdzh:
for reference, original PR #1680

grobie:
> for reference, original PR #1680

Yes, I have to admit I didn't fully understand all implications when commenting on that PR.

plugin/forward/proxy.go
```diff
@@ -48,9 +48,8 @@ func (p *Proxy) Addr() (addr string) { return p.addr }
 func dnsClient(tlsConfig *tls.Config) *dns.Client {
 	c := new(dns.Client)
 	c.Net = "udp"
-	// TODO(miek): this should be half of hcDuration?
-	c.ReadTimeout = 1 * time.Second
-	c.WriteTimeout = 1 * time.Second
+	c.ReadTimeout = hcInterval / 2
+	c.WriteTimeout = hcInterval / 2
```
rdrozhdzh: not sure about this change, 250ms may be too few for healthcheck. @miekg are you ok with this change?

rdrozhdzh (commented):
LGTM, except the change mentioned in the comment below. I agree that 200ms is close to ideal min timeout - no sense to make it lower

plugin/forward/proxy.go
miekg: everything is now getting dynamic, except this, we might be making it too low when something is farther away. 
Can we drop the /2 here?

In the general the hc also also needs to be more complex and do an exp. backoff  - but that's unrelated to this PR

grobie: I believe it doesn't make sense to allow a longer timeout than the interval. I'll revert this change and send another PR.

rdrozhdzh: Actually in case if upstream is down we poll the upstream each `hcInterval + readTimeout` (or `hcInterval + dialTimeout` for TCP), i.e. the time between subsequent polls is `hcInterval + readTimeout`. We could consider decreasing `hcInterval` and increasing `c.ReadTimeout` if we want to tune this logic further

rdrozhdzh: alternatively we can consider using ticker in Probe instead of sleep
https://github.com/coredns/coredns/blob/e3534205c7c3a89b09c9a9e2aca1a2159eb9ecde/plugin/pkg/up/up.go#L41

grobie: Oh, I find it surprising to see that the hcInterval constant only describes a part of the actual interval. Anyway, I reverted this change to the health timeouts and will open another PR.

rdrozhdzh (approved):

Issue #1823, opened 2018-05-21
http server shoud set valid http max-age for caching


miekg:
This is currently hardcoded:
~~~
	w.Header().Set("Cache-Control", "max-age=128") // TODO(Minttl as done in cache.
~~~
This should use the minttl logic from the cache package which should be factored out into a `pkg/ttl` package or something.

PR #1890, opened 2018-06-20
Fix max-age in http server

miekg:
Move the minMsgTTL to dnsutil and rename it MinimalTTL, move some
constants there as well.
Use these new function in server_https to correctly set the max-age
HTTP header.

Fixes: #1823


miekg:
@bradbeam do you have time to look at this? Thanks!

bradbeam (approved):

miekg:
Thanks!

PR #1892, opened 2018-06-21
Makefile fixes:

### LINKED ISSUE #1891: Readme needs to be reread and validated.
tested on Ubuntu 16.04

Both git clone method and docker compiled methods of creating the end binary fail.  Housecleaning or editing should be in order for these steps to work or updated steps to documentation given.

miekg:
Correctly set the path for the presubmits, fix some typos and make
the goimport target do the linting to the lint target can be removed.

Also don't make it a fatal error because gofmt changes between releases.

Fixes #1891

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,5 @@ before_script:
 script:
   - make TEST_TYPE=$TEST_TYPE travis
 
-
 after_success:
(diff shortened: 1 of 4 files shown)


yongtang (approved):
LGTM

Issue #1638, opened 2018-03-26
Plugin/rewrite: Extend rewrite plugin to support external plugins


rajansandeep:
Implement a hook which allows the rewrite plugin to support external plugins that will be an extension to the rewrite plugin.

Eg. The rewrite plugin would support an external plugin which extends the edns0 variable rule which allows returning the client's mac address.

This would make extending the rewrite plugin easily and avoid clutter

budaki:
Pls help me open my plugins

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] Plugin/rewrite: E..." ]
>Implement a hook which allows the rewrite plugin to support external plugins that will be an extension to the rewrite plugin.
>
>Eg. The rewrite plugin would support an external plugin which extends the edns0 variable rule which allows returning the client's mac address.
>
>This would make extending the rewrite plugin easily and avoid clutter

Can't this be done, by just chaining plugins?

/Miek

--
Miek Gieben

chrisohaver:
> Can't this be done, by just chaining plugins?

@miekg, Do you mean by creating a new plugin that wraps the existing rewrite code (which would need to be exposed/refactored to accommodate), and adds the new feature? For example, as it is done similarly in `auto` (wrapping `file`) and `kubernetai` (wrapping `kubernetes`) - a new `rewriteclientmac` plugin wrapping `rewrite`?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Plugin/rewrit..." ]
>> Can't this be done, by just chaining plugins?
>
>@miekg, Do you mean by creating a new plugin that wraps the existing rewrite code (which would need to be exposed/refactored to accommodate), and adds the new feature? For example, as it is done similarly in `auto` (wrapping `file`) and `kubernetai` (wrapping `kubernetes`) - a new `rewriteclientmac` plugin wrapping `rewrite`?

I was thinking using the normal wrapping of a responseWriter, but that doesn't 
work, because you can then only influence the return packet. So disregard that 
:-)o

I have no good ideas for this yet, but want to warn to not proliferate too many 
plugin types, we already have quite a few.


/Miek

--
Miek Gieben

rajansandeep:
I was thinking of three ways where I can implement eg. getting the mac address of the client.

- Extend the existing rewrite plugin by adding an edns0 variable called {client_mac}
[...]

PR #1894, opened 2018-06-22
plugin/metadata: add metadata plugin

eugen-kleiner-mesh:
This PR implements metadata plugin. Made to be able to exchange by data between plugins.
Also make rewrite plugin support metadata.
https://github.com/coredns/coredns/issues/1638


plugin/metadata/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Syntax
+
+~~~
+metadata [ZONES... ]
+~~~
+
+## Plugins
+
+Any plugin that implements the Metadater interface will be called to get metadata information.
+If **ZONES** is used it specifies all the zones the plugin should add Metadata to.
```
miekg: I don't understand the second part of this paragraph on how ZONES is used?

miekg: Just saw an example below. So it just limits it to queries within that zone, just like other plugins, got it; that could be made clearer though

plugin/metadata/README.md
```diff
@@ -0,0 +1,39 @@
...
+. {
+    metadata
+}
+~~~
+
+Add metadata for all requests within `example.org.`.
+
+~~~ corefile
+. {
+    metadata example.org
```
miekg: shouldn't there be a *whoami* plugin or something loaded as well to make this config snipet actually do something?

plugin/metadata/metadata.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// Metadater interface needs to be implemented by each plugin willing to provide
+// metadata information for other plugins.
+// Note: this method should work quickly, because it is called for every request
+// from the metadata plugin.
+type Metadater interface {
+	// Metadata gets content, ResponseWriter and dns.Msg and returns context with
+	// additional values. Metadata must be thread safe.
+	Metadata(context.Context, dns.ResponseWriter, *dns.Msg) (context.Context, error)
```
miekg: I think it is more useful to describe what we expect from a call to Metadata. The line about what parameters it takes is redundant.

Why must it be thread safe btw?

plugin/metadata/metadata.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// Metadata implements collecting metadata information from all enabled plugins
+// which provide it
```
miekg: // Metadata implements collecting metadata information from all plugins that implement the Metadater interface.
￼

plugin/metadata/metadata.go
```diff
@@ -0,0 +1,48 @@
...
+func (m *Metadata) Name() string { return "metadata" }
+
+// ServeDNS implements the plugin.Handler interface.
+func (m *Metadata) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+
+	state := request.Request{W: w, Req: r}
+	if len(m.Zones) == 0 || plugin.Zones(m.Zones).Matches(state.Name()) != "" {
+		// Go through all metadaters and collect metadata
+		for _, metadater := range m.Metadaters {
+			if c, err := metadater.Metadata(ctx, w, r); err == nil {
```
miekg: probably document that we only include the metadata is err is nil (in the interface documentation)

plugin/metadata/metadata.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// Name implements the Handler interface.
+func (m *Metadata) Name() string { return "metadata" }
+
+// ServeDNS implements the plugin.Handler interface.
+func (m *Metadata) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+
+	state := request.Request{W: w, Req: r}
+	if len(m.Zones) == 0 || plugin.Zones(m.Zones).Matches(state.Name()) != "" {
```
miekg: m.Zones should never be 0; either **ZONES** is specified or the zones from the block are applied

plugin/metadata/metadata_test.go
```diff
@@ -0,0 +1,57 @@
+package metadata
+
+import (
+	"context"
+	"testing"
+
+	"github.com/coredns/coredns/plugin/test"
+	"github.com/miekg/dns"
```
miekg: fmt

plugin/metadata/setup.go
```diff
@@ -0,0 +1,49 @@
...
+	caddy.RegisterPlugin("metadata", caddy.Plugin{
+		ServerType: "dns",
+		Action:     setup,
+	})
+}
+
+func setup(c *caddy.Controller) error {
+	m := &Metadata{}
+	c.Next()
+	zones := c.RemainingArgs()
```
miekg: this should check the zones from the server block and use those as the default, potentially being overridden with these values, see other plugin for examples.

Also needs a setup_test.go

miekg (commented):
Nice. Do we have a pluign using this ?

plugin/metadata/metadata_test.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// testMetadater implements fake Metadaters (plugins which inmplement Metadater interface)
+type testMetadater struct {
+	key   interface{}
```
johnbelamaric: Each plugin can't define its own key - if we do that then there is no way for things that are going to *use* the metadata - like `rewrite` and `policy` to pull the data out. Instead, the `metadata` plugin should define a key, and methods that the plugins can use to add/remove their metadata from the context. I also suggest that internally this is just kept by `metadata` as a map in the context; similar to how [grpc metadata](https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md) works.

plugin/metadata/metadata.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/request"
+	"github.com/miekg/dns"
+)
+
+// Metadater interface needs to be implemented by each plugin willing to provide
+// metadata information for other plugins.
+// Note: this method should work quickly, because it is called for every request
+// from the metadata plugin.
+type Metadater interface {
```
johnbelamaric: It's awkward but I think this needs to be "Metadataer" - of if that is just too ugly we can do "MetadataProvider" or something like that. But "Metadater" is not right from an English perspective.

fturib:
First plugin to use it will be REWRITE - be able to extend the set of variable that can be used to fill esdn0 part.

One implementation currently in progress for providing the data is "mac" - that we expect as external plugin: it would provide the mac address of the client-ip based on the ARP tables.

In a second phase, we would consolidate all variables using that Metadata concept:
- the one used on REWRITE
- the one used in Log
- maybe other ...
before starting here, we have to think how we handle the backward compatibility of the corefile.

plugin/metadata/README.md
```diff
@@ -0,0 +1,39 @@
+# metadata
+
+## Name
+
+*metadata* - enable  a metadata collector.
+
+## Description
+
+By enabling *metadata* any plugin that implements [metadata.metadater interface](https://godoc.org/github.com/coredns/coredns/plugin/metadata#Metadater) will be called and will be able to add it's own Metadata to context. The metadata collected will be available for all plugins handler, via the Context parameter provided in the ServeDNS function
+
```
johnbelamaric: This plugin needs to do a few other things, too:

1) It needs to manage the "built-in" variables that you would find in `rewrite`. That is, when executed, it should create metadata entries for all those built-in variables. Note that it should still be possible to use `rewrite` *without* `metadata`, so you may want to refactor some stuff to enable some code sharing there.

2) It needs to enable the parsing of EDNS0 options in the same way that the [Themis policy plugin](https://github.com/infobloxopen/themis/tree/master/contrib/coredns/policy) does. This can come in a later PR, but put it in your plans.

The README should explain some use cases - I can add them later if you want.

plugin/metadata/README.md
```diff
@@ -0,0 +1,39 @@
+# metadata
+
+## Name
+
+*metadata* - enable  a metadata collector.
```
johnbelamaric: remove extra space

johnbelamaric (changes requested):

miekg:
>before starting here, we have to think how we handle the backward compatibility of the corefile.

We have a policy, but not documented (oops):

1. version 1.x.y: announce upcoming deprecation(s)
2. version 1.x.y+1: keep config options but remove code, doesn't break Corefile 
   parsing.
3. version 1.x+1.0: remove the config options.

We are going through this right now, next release will be 1.2.0 and 41c72478 
implemented step 3.

plugin/metadata/metadata_test.go
eugen-kleiner-mesh: I didn't get the point of this comment

plugin/metadata/README.md
johnbelamaric: Yeah, we will need better examples that also illustrate the point of this.

eugen-kleiner-mesh: Sure, I will add whoami plugin. But there is no public plugins which support metadataer now.

plugin/metadata/README.md
eugen-kleiner-mesh: Regarding part 1. Could you clarify the idea? I see at least 3 ways which may fit:
1. Rewrite plugin implements Metadataer to add them to metadata in case if other plugins need them. Other logic remains the same (support of additional variables by rewrite is added in any case).
2. "build-in" variables are moved to metadata plugin which implements Metadataer to add them to metadata. Rewrite plugin will remain functional, but "built-in" variables are not available if metadata is disabled.
3. Same as 2, but variables are supposed to be accessible somehow with no metadata plugin enabled. I don't see how this can be implemented with no conflict with conception of plugins.

fturib: @johnbelamaric : can you reply to Eugen as soon as you can so he can go ahead ? Thanks!

plugin/metadata/metadataer.go
```diff
@@ -0,0 +1,58 @@
...
+// MD is metadata information storage
+type MD map[string]interface{}
+
+// MDKeyType defines the type of key that is used to save metadata into the context
+type mDKeyType string
+
+// MDKey specifies
+const mDKey mDKeyType = "md-key"
+
+type metadataKey struct{}
```
johnbelamaric: you can still use this below instead of the string - it's fine now to be unexported because you have the FromContext function.

johnbelamaric (changes requested):
I suggest moving this functionality: https://github.com/coredns/coredns/blob/master/plugin/rewrite/edns0.go#L241 to this package, and exporting it. This plugin can then implement Metadataer and populate the MD with all those variables by simply calling this function for each name. In the `rewrite` plugin, you would try to get the MD from the context; if it's not there, you use the function directly, otherwise you use the MD.

plugin/metadata/metadataer.go
eugen-kleiner-mesh: It was my first implementation and I forgot to remove it when moving to "const mDKey mDKeyType". But I can switch back to metadataKey{} if you think it looks better.

eugen-kleiner-mesh:
@johnbelamaric, I hope, moving this variables is temporary until functionality is deprecated? If we want to have metadata as plugin, I think variables shouldn't be retrieved bypassing metadata from context.

plugin/metadata/metadataer.go
johnbelamaric: Yes, please do.

johnbelamaric:
I am not sure I follow the concern? I want `rewrite` to continue to work, at least for one dev cycle, even without the `metadata` plugin. We could deprecate that and require `metadata` to use the variables in `rewrite` in a later release.

eugen-kleiner-mesh:
Good. I just wanted to make sure.

eugen-kleiner-mesh:
I have moved the "built-in" variables to metadata.
I think that variables.go should be moved from metadata to pkg directory. Please, suggest the preferred naming if you agree with moving it to pkg.
I will update READMEs and UTs tomorrow.

johnbelamaric:
Yes, pkg makes sense.

plugin/rewrite/edns0.go
```diff
@@ -146,7 +147,7 @@ func newEdns0Rule(mode string, args ...string) (Rule, error) {
 		}
 		//Check for variable option
 		if strings.HasPrefix(args[3], "{") && strings.HasSuffix(args[3], "}") {
-			return newEdns0VariableRule(mode, action, args[2], args[3])
+			return newEdns0VariableRule(mode, action, args[2], strings.Trim(args[3], "{}"))
 		}
```
fturib: I am not sure what is will do if you have a bracket inside the variable name.
isn't is safer to remove first and last char, as you already tested them ?

plugin/metadata/metadataer.go
```diff
@@ -0,0 +1,53 @@
...
+func (m MD) Get(key string) (value interface{}, ok bool) {
+	value, ok = m[key]
+	return
+}
+
+// addValues adds metadata values
+func (m MD) addValues(src map[string]interface{}) {
+	for k, v := range src {
+		m[k] = v
+	}
```
fturib: I am not sure what John meant about the key "not defined by the metadater", but I guess here if we have 2 time the same key (from  different metadataer we will overwrite one).

=> Maybe good to warn the user - a log ? an error ? servfail the query ?

Or should we ensure a namespace per plugin ?

fturib:
CI Travis job is failing here. But looks like the go version of the container is "go version go1.7.4 linux/amd64". 
(see https://travis-ci.org/coredns/coredns/jobs/396960595).

In travis.yaml, the go version required is "1.10.x" (see also here : https://travis-ci.org/coredns/coredns/jobs/396960595/config)


So, what went wrong ?

miekg:
>@miekg : CI Travis job is failing here. But looks like the go version of the container is "go version go1.7.4 linux/amd64".
>(see https://travis-ci.org/coredns/coredns/jobs/396960595).
>
>In travis.yaml, the go version required is "1.10.x"
>
>So, what went wrong ?

you have to ask travis.

I restarted that build and seemed to work fine.

plugin/metadata/metadataer.go
eugen-kleiner-mesh: Decided to show error in log and remove both entries in case of duplicate.

plugin/metadata/README.md
```diff
@@ -0,0 +1,42 @@
+# metadata
+
+## Name
+
+*metadata* - enable a metadata collector.
+
+## Description
+
+By enabling *metadata* any plugin that implements [metadata.Metadataer interface](https://godoc.org/github.com/coredns/coredns/plugin/metadata#Metadataer) will be called and will be able to add it's own Metadata to context. The metadata collected will be available for all plugins handler, via the Context parameter provided in the ServeDNS function
```
fturib: .. will be called **for each DNS query, at being of the process for that query, in order** to add it's own ...

plugin/metadata/README.md
```diff
@@ -0,0 +1,42 @@
...
+
```
fturib: metadata plugin is automatically adding the so-called default medatada (extracted from the query) in the context. Those default metadata are : <list the defaults>

plugin/metadata/README.md
```diff
@@ -0,0 +1,42 @@
...
+~~~ corefile
+. {
+    metadata
+    rewrite edns0 local set 0xffee {client_ip}
+    forward . 8.8.8.8:53
+}
+~~~
+
+Add metadata for all requests within `example.org.`. Rewrite uses one of provided by default metadata variables.
+
```
fturib: question: in that case, what will be the value added to edns0 when the query is not for "example.org." ?

plugin/metadata/metadataer.go
```diff
@@ -0,0 +1,65 @@
...
+	"context"
+
+	"github.com/coredns/coredns/plugin/pkg/log"
+	"github.com/miekg/dns"
+)
+
+// Metadataer interface needs to be implemented by each plugin willing to provide
+// metadata information for other plugins.
+// Note: this method should work quickly, because it is called for every request
+// from the metadata plugin.
```
fturib: because it will be called for every DNS Query that match metadata plugin filter.

miekg:
>ekleiner commented on this pull request.
>
>
>
>> +		}
>+	}
>+	return nil, false
>+}
>+
>+// Get returns metadata value by key
>+func (m MD) Get(key string) (value interface{}, ok bool) {
>+	value, ok = m[key]
>+	return
>+}
>+
>+// addValues adds metadata values
>+func (m MD) addValues(src map[string]interface{}) {
>+	for k, v := range src {
>+		m[k] = v
>+	}
>
>Decided to show error in log and remove both entries in case of duplicate.

These would better named setValue and Value (instead of Get), to be more inline 
with Go style.

fturib:
@johnbelamaric : can you review this PR from Eugen ? he will be very soon (end of this week) out for 2 weeks. And that would be better if we can have it done before he leave. Thanks!
- And he also need this to be merged to be able to have the MAC plugin (external) working.

johnbelamaric:
I guess I thought Eugene was going to rework this to evaluate the variables only on use?

johnbelamaric:
As described here: https://github.com/coredns/coredns/issues/1638#issuecomment-400095954

fturib:
@johnbelamaric : that will be in second phase. it is an improvement.
Hum ... but that needs the interface Metadataer is already in place.

@ekleiner : could you change with using the new interface you proposed in your comment ?
(for now, the Metadata serveDNS can use both functions to collect all the data immediatly).

plugin/metadata/README.md
```diff
@@ -0,0 +1,44 @@
...
+## Syntax
+
+~~~
+metadata [ZONES... ]
+~~~
+
+## Plugins
+
+Any plugin that implements the Metadataer interface will be called to get metadata information.
+If **ZONES** is specified then metadata add is limited by zones. Metadata is added to all context going through Metadataer if **ZONES** are not specified.
```
miekg: "Metadata is added to all context going through Metadataer if **ZONES** are not specified."

"all context" looks weird here?

plugin/metadata/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Examples
+
+Enable metadata for all requests. Rewrite uses one of provided by default metadata variables.
```
miekg: "Rewrite uses one of provided by default metadata variables."

Missing "the"?
Also it's not clear what happens here, `{client_ip}` is not extracted from the metadata I assume

plugin/metadata/metadata.go
```diff
@@ -0,0 +1,51 @@
+package metadata
+
+import (
+	"context"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/pkg/variables"
+	"github.com/coredns/coredns/request"
+	"github.com/miekg/dns"
```
miekg: fmt: newline about /miekg/dns

plugin/metadata/metadata_test.go
miekg: newline above miekg/dns

plugin/metadata/metadataer.go
miekg: this should be put in the README as well

plugin/metadata/metadataer.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// MD is metadata information storage
+type MD map[string]interface{}
+
+// metadataKey defines the type of key that is used to save metadata into the context
+type metadataKey struct{}
+
+// newMD initializes MD and attaches it to context
+func newMD(ctx context.Context) (MD, context.Context) {
```
miekg, line 29: Is this only used once? Then I would just inline it and show what we're doing.

plugin/metadata/metadataer.go
```diff
@@ -0,0 +1,65 @@
...
+	m := MD{}
+	return m, context.WithValue(ctx, metadataKey{}, m)
+}
+
+// FromContext retrieves MD struct from context
```
miekg: Closing dot; it is an sentence.

plugin/metadata/metadataer.go
```diff
@@ -0,0 +1,65 @@
...
+func FromContext(ctx context.Context) (md MD, ok bool) {
+	if metadata := ctx.Value(metadataKey{}); metadata != nil {
+		if md, ok := metadata.(MD); ok {
+			return md, true
+		}
+	}
+	return nil, false
```
miekg: `nil`? That happens to work here? Why not `MD{}]`?

plugin/metadata/metadataer.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Get returns metadata value by key
```
miekg: closing dot

plugin/metadata/metadataer.go
```diff
@@ -0,0 +1,65 @@
...
+func (m MD) Value(key string) (value interface{}, ok bool) {
+	value, ok = m[key]
+	return
```
miekg: naked returns are discouraged, so add the returned variables.
I think btw in this function it make less sense to name the return parameters at all.

plugin/metadata/metadataer.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// addValues adds metadata values.
+// If variable with a new key already attached then new is not appllied, old is removed.
```
miekg: Comment doesn't reflect new name of the function. I also don't understand the reason for doing 'new not appllied, old is removed'. Why not just overwrite?

plugin/metadata/metadataer.go
```diff
@@ -0,0 +1,65 @@
...
+func (m MD) setValues(src map[string]interface{}) {
+	duplicates := []string{}
+	for k, v := range src {
+		if _, ok := m[k]; !ok {
+			m[k] = v
+		} else {
+			duplicates = append(duplicates, k)
+			log.Errorf("Metadata variable '%v' has duplicates. None of them is used.", k)
```
miekg: Never log in deep functions, return the error

plugin/metadata/setup.go
```diff
@@ -0,0 +1,61 @@
+package metadata
+
+import (
+	"github.com/coredns/coredns/core/dnsserver"
+	"github.com/coredns/coredns/plugin"
+	"github.com/mholt/caddy"
```
miekg: newline above caddy

plugin/metadata/setup_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+		if !test.shouldErr && err == nil {
+			if !reflect.DeepEqual(test.zones, m.Zones) {
+				t.Errorf("Test %d: Expected zones %s. Zones were: %v", i, test.zones, m.Zones)
+			}
+		}
+
+		// Check the full setup call too
+		c = caddy.NewTestController("dns", test.input)
+		err = setup(c)
```
miekg: this is a new test function and should probably be moved to setup_test

plugin/pkg/variables/variables.go
```diff
@@ -0,0 +1,108 @@
+package variables
+
+import (
+	"encoding/binary"
+	"fmt"
+	"net"
+	"strconv"
+
+	"github.com/coredns/coredns/request"
+	"github.com/miekg/dns"
```
miekg: newline above miekg/dns

plugin/pkg/variables/variables.go
```diff
@@ -0,0 +1,108 @@
...
+	queryName  = "qname"
+	queryType  = "qtype"
+	clientIP   = "client_ip"
+	clientPort = "client_port"
+	protocol   = "protocol"
+	serverIP   = "server_ip"
+	serverPort = "server_port"
+)
+
+// AllProvidedVars is a list of available variables provided by GetMetadataValue
```
miekg: Vars repeat variables. This is better named `All`?

plugin/pkg/variables/variables.go
```diff
@@ -0,0 +1,108 @@
...
+var AllProvidedVars = []string{queryName, queryType, clientIP, clientPort, protocol, serverIP, serverPort}
+
+// GetMetadataValue calculates and returns the data specified by the variable name.
+// Supported varNames are listed in allProvidedVars.
+func GetMetadataValue(varName string, w dns.ResponseWriter, r *dns.Msg) ([]byte, error) {
```
miekg: Just Metadata?

plugin/pkg/variables/variables.go
```diff
@@ -0,0 +1,108 @@
...
+	req := request.Request{W: w, Req: r}
+	switch varName {
+	case queryName:
+		//Query name is written as ascii string
```
miekg: why this comment?

plugin/pkg/variables/variables.go
miekg: Is returning `[]byte` wise here? Or is that done to not have an `interface{}` here?
Should we split this up and allow sane return values?

plugin/metadata/README.md
eugen-kleiner-mesh: > Also it's not clear what happens here, {client_ip} is not extracted from the metadata I assume

If metadata is enabled, then this variables are extracted from metadata. They are taken directly only in case if not avalable in metadata for backward compatibility.

miekg:
>ekleiner commented on this pull request.
>
>
>
>> +
>+
>+## Syntax
>+
>+~~~
>+metadata [ZONES... ]
>+~~~
>+
>+## Plugins
>+
>+Any plugin that implements the Metadataer interface will be called to get metadata information.
>+If **ZONES** is specified then metadata add is limited by zones. Metadata is added to all context going through Metadataer if **ZONES** are not specified.
>+
>+## Examples
>+
>+Enable metadata for all requests. Rewrite uses one of provided by default metadata variables.
>
>> Also it's not clear what happens here, {client_ip} is not extracted from the metadata I assume
>
>It is. If metadata is enabled, then this variables are taken from metadata. They are taken directly only in case if not avalable in metadata for backward compatibility.

Sorry a 'not' snuck into that sentence. Anyway this could/should use some extra 
wording to make it clearer to the reader.

plugin/metadata/metadataer.go
eugen-kleiner-mesh: I didn't get your point here. Could you clarify what should be done here?

plugin/metadata/metadataer.go
eugen-kleiner-mesh: I think that "ok" parameter name explains it's meaning.

plugin/metadata/metadataer.go
eugen-kleiner-mesh: The idea is to discourage implementation of variables with the same name by 2 different plugins. So, if the same metadata is implemented by 2 plugins, we will have a stable result at least. We can't predict which one will be applied if we overwrite.

plugin/pkg/variables/variables.go
eugen-kleiner-mesh: I just moved all logic from https://github.com/coredns/coredns/blob/master/plugin/rewrite/edns0.go#L246 . Will remove.

plugin/pkg/variables/variables.go
eugen-kleiner-mesh: It is used only in rewrite module now. So, returning in format of the only consumer makes sense from my point of view.

eugen-kleiner-mesh:
@johnbelamaric , @miekg : As Francois mentioned I am soon out for 2 weeks and taking into account other external work I need to do, this is the last day I can handle your comments. I would appreciate getting feedback as soon as possible and will try to make changes ASAP.
Note: I will make the improvement mentioned in [comment](https://github.com/coredns/coredns/issues/1638#issuecomment-400095954) after I return from VAC.

plugin/metadata/metadataer.go
```diff
@@ -12,10 +12,12 @@ import (
...
-	// Metadata is expected to return map with metadata information which can be
-	// later retrieved from context by any other plugin. It may return empty
-	// map if no metadata needs to be published.
-	Metadata(context.Context, dns.ResponseWriter, *dns.Msg) map[string]interface{}
+	// List of variables which are provided by current metadataer. Must remain constant.
+	MetadataVarNames() []string
+	// Metadata is expected to return a value with metadata information by the key
+	// from first argument. Value can be later retrieved from context by any other plugin.
+	// If value is not available by some reason returned boolean value should be false.
+	Metadata(string, context.Context, dns.ResponseWriter, *dns.Msg) (interface{}, bool)
```
eugen-kleiner-mesh: There is a linter warning here: "context.Context should be the first parameter of a function (golint)".
Should I move a key from first to the last argument, or we don't care much about linter?

miekg:
>ekleiner commented on this pull request.
>
>
>
>> @@ -12,10 +12,12 @@ import (
> // Note: this method should work quickly, because it is called for every request
> // from the metadata plugin.
> type Metadataer interface {
>-	// Metadata is expected to return map with metadata information which can be
>-	// later retrieved from context by any other plugin. It may return empty
>-	// map if no metadata needs to be published.
>-	Metadata(context.Context, dns.ResponseWriter, *dns.Msg) map[string]interface{}
>+	// List of variables which are provided by current metadataer. Must remain constant.
>+	MetadataVarNames() []string
>+	// Metadata is expected to return a value with metadata information by the key
>+	// from first argument. Value can be later retrieved from context by any other plugin.
>+	// If value is not available by some reason returned boolean value should be false.
>+	Metadata(string, context.Context, dns.ResponseWriter, *dns.Msg) (interface{}, bool)
>
>There is a linter warning here: "context.Context should be the first parameter of a function (golint)".
>Should I move a key from first to the last argument?

yep, please do

plugin/metadata/metadataer.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+
+	"github.com/miekg/dns"
+)
+
+// Metadataer interface needs to be implemented by each plugin willing to provide
+// metadata information for other plugins.
+// Note: this method should work quickly, because it is called for every request
+// from the metadata plugin.
+type Metadataer interface {
```
johnbelamaric: Since it is no longer a single function interface, can we change the interface name. Maybe "MetadataProvider"?

plugin/metadata/metadataer.go
johnbelamaric: It is used in both ServeDNS and in the test code.

plugin/metadata/metadataer.go
```diff
@@ -0,0 +1,53 @@
...
+		if md, ok := metadata.(MD); ok {
+			return md, true
+		}
+	}
+	return MD{}, false
+}
+
+// Value returns metadata value by key.
+func (m MD) Value(key string) (value interface{}, ok bool) {
+	value, ok = m[key]
```
johnbelamaric, line 46: Why not just "return m[key]"?

johnbelamaric (changes requested):

miekg:
>Since it is no longer a single function interface, can we change the interface name. Maybe "MetadataProvider"?

Or just Provider or Metadata. It's already prefixed with the pkg name.

--
Miek Gieben

johnbelamaric:
Yes, just `metadata.Provider` sounds best to me.

plugin/metadata/metadataer.go
eugen-kleiner-mesh:
```
plugin/metadata/metadataer.go:46:2: not enough arguments to return
	have (interface {})
	want (interface {}, bool)
```

johnbelamaric (approved):
The README needs work, but that can be a separate PR. Create an issue for it and assign it to me.

fturib:
Created issue #1903  for reworking the README and assigned to @johnbelamaric

eugen-kleiner-mesh:
@miekg, it there any pending changes left from your point of view?

miekg:
>@fturib, thank you.
>@miekg, it there any pending changes left from your point of view?

I might, be we can iterate on it later

lgtm

PR #1897, opened 2018-06-25
Fix: metric `cache_miss_total`  shall include ttl case

accepting:
In the case of ttl <= 0, `cacheMisses` should increase its counter.

Signed-off-by: Jun Li <lijun.git@gmail.com>

As for plugin cache, when it occurs that the dns result is out of date(which means its `ttl <= 0`), metric "cache_miss_total" should increase its counter.

Not yet.

None.

### DIFF
--- a/plugin/cache/handler.go
+++ b/plugin/cache/handler.go
@@ -27,8 +27,8 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 
 	server := metrics.WithServer(ctx)
 
-	i, ttl := c.get(now, state, server)
-	if i != nil && ttl > 0 {
+	i, found := c.get(now, state, server)
+	if i != nil && found {
 		resp := i.toMsg(r, now)
 
 		state.SizeAndDo(resp)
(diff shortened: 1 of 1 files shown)


plugin/cache/handler.go
```diff
@@ -68,13 +68,23 @@ func (c *Cache) get(now time.Time, state request.Request, server string) (*item,
 	k := hash(state.Name(), state.QType(), state.Do())
 
 	if i, ok := c.ncache.Get(k); ok {
-		cacheHits.WithLabelValues(server, Denial).Inc()
-		return i.(*item), i.(*item).ttl(now)
+		ttl := i.(*item).ttl(now)
+		if ttl > 0 {
```
grobie: It doesn't look like there is any need to return the ttl at all. I'd propose to clean this up:

```go
func (c *Cache) get(now time.Time, state request.Request, server string) (*item, bool) {
	k := hash(state.Name(), state.QType(), state.Do())

	if i, ok := c.ncache.Get(k); ok && i.(*item).ttl(now) > 0 {
		cacheHits.WithLabelValues(server, Denial).Inc()
		return i.(*item), true
	}
	if i, ok := c.pcache.Get(k); ok && i.(*item).ttl(now) > 0 {
		cacheHits.WithLabelValues(server, Success).Inc()
		return i.(*item), true
	}
	cacheMisses.WithLabelValues(server).Inc()
	return nil, false
}
```

accepting:
updated @grobie

grobie:
Thanks!

Issue #1810, opened 2018-05-17
CNAME target resolution plugin


chrisohaver:
Currently, adding `A` records for `CNAME` targets in a response must be implemented per plugin that may generate a `CNAME`. 

Alternately, We could do this with an early-in-the-chain plugin, using a non-writer, that sends requests down the chain, but then looks at the response for before writing to the client. For each unresolved `CNAME` it sees (a `CNAME` without an `A` record for the target), it creates an `A` record request by sending the target name down the chain and adding the result (if any) to the response. It would continue until it has tried looking up each unique `CNAME` in the response.

This method is also loop safe for `CNAME` chain loops without relying on a depth limit like the current recursive method.

As a separate plugin, it enables `CNAME` target resolution for any current/future plugin.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] CNAME target reso..." ]
>Currently, adding `A` records for `CNAME` targets in a response must be implemented per plugin that may generate a `CNAME`.
>
>Alternately, We could do this with an early-in-the-chain plugin, using a non-writer, that sends requests down the chain, but then looks at the result for before writing to the client. For each unresolved CNAME it sees (CNAME without A Records), it creates an A record request, and sends that down the chain, adding the result (if any) to the results. Continuing until it has tried looking up each unique CNAME in the result.
>
>This method is also loop safe for CNAME chain loops without relying on a depth limit like the current recursive method.
>
>As a separate plugin, it enables `CNAME` target resolution for any current/future plugin.

I like the plugin idea, but might re-inspecting every packet for cname impair 
performance too much?

chrisohaver:
It would look at the record type of every answer in the response one time.  I don't think that would be a big hit, but I could be wrong.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] CNAME target ..." ]
>It would look at the record type of every answer in the response one time.  I don't think that would be a big hit, but I could be wrong.

Yeah, I might be overly concerned. This will clean up a lot of places that need 
CNAME resolving, so I welcome a plugin that simplifies it.

PR #1898, opened 2018-06-25
New plugin for resolving CNAMEs and SRVs

chrisohaver:

Plugin to _resolve_ CNAMEs and SRVs produced by any other plugin, by looking up target name and adding it to the answer.

#1810

included


chrisohaver:
I dont like the name either.  I'm hoping you guys have better ideas.

stp-ip:
Some DNS providers doing CNAME resolution/flattening are using:
* ANAME
* cname flattening
* record flattening

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: New plug..." ]
>I dont like the name either.  I'm hoping you guys have better ideas.

hmm though one. Officialy this process is called 'additional query processing', 
in the current code we use Lookup. Both are rather generic.

Maybe 'postprocess' or just 'post' ?

chrisohaver:
I think `post` is too generic unless we want to leave the door open for other types of post processing in the same plugin, in which case we'd need to add an option for the cname function e.g. ...

```
post . {
    lookup cnames
}
```

But If we want to make this plugin just about adding records for CNAME targets ... how about just `cnames` ?

chrisohaver:
I guess "alias" might be a better generic term... if we want to add DNAME support here too eventually.  Although I think for some DNS providers they use the term "alias" to mean a type of alias other than a standard DNS records, so could be confusing.

alias
dealias
unalias
antialias - lol no
flattenalias

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: New plug..." ]
>I guess "alias" might be a better generic term... if we want to add DNAME support here too eventually.  Although I think for some DNS providers they use the term "alias" to mean a type of alias else other than a standard DNS records, so could be confusing.
>
>alias
>dealias
>unalias
>antialias - lol no
>flattenalias

How about resolving SRV records ?

chrisohaver:
> How about resolving SRV records ?

For now it only handles A/AAAA queries.  OK to add later? Or should we support that in first version?

johnbelamaric:
It makes sense to call it ”additional” or something along those lines if we’re going to add this sort of thing. No need to do it all in one PR though.

chrisohaver:
> How about resolving SRV records ?

Oh - do you mean adding the A/AAAA records to the extra/additional section for SRV records?

Or do you mean CNAMEs with a target that holds an SRV record.
Or SRV records that have have targets that are CNAMEs (i recall spec doesn't allow this).

chrisohaver:
I find "additional" confusing because it's also the name of a response section, which has nothing to do with CNAME target resolution.  CNAME records, and their target A records are included in the answer section, not the additional section (iiuc).

chrisohaver:
If we want to add `SRV` flattening later, we could name the plugin `flatten`, as stp-ip suggests...

```
flatten . {
   cname
   srv
}
```

chrisohaver:
or

```
resolve . {
   cname
   srv
}
```

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: New plug..." ]
>> How about resolving SRV records ?
>
>Oh - do you mean adding the A/AAAA records to the extra/additional section for SRV records?

Yes, we do this now as well (IIRC), this basically a requirement if we are used 
as the first hop recursor (or wait for the follow up query - if it will be 
done).

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: New plug..." ]
>If we want to add `SRV` flattening later, we could name the plugin `flatten`, as stp-ip suggests...
>
>```
>flatten . {
>   cname
>   srv
>}
>```

it's also not really flattening is it? We are putting A/AAAA in the additional 
with this? Or are we removing the CNAME entirely? Then flatten would be a good 
name.

Thinking further ahead we could/should extend this probably to all 
external names?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: New plug..." ]
>or
>
>```
>resolve . {
>   cname
>   srv
>}
>```

yes, why not!

chrisohaver:
> We are putting A/AAAA in the additional with this? Or are we removing the CNAME entirely?

For CNAME, we add records to the ANSWER section.
For SRV, we would add records to the ADDITIONAL section.

Never removing CNAMEs or SRVs.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: New plug..." ]
>> We are putting A/AAAA in the additional with this? Or are we removing the CNAME entirely?
>
>For CNAME, we add records to the ANSWER section.
>For SRV, we would add records to the ADDITIONAL section.
>
>Never removing CNAMEs or SRVs.

resolve captures this, we could have a `flatten` directive at some point (or 
whatever)

chrisohaver:
I think this is pretty much done... 

@miek: what kind of records are CNAMEs allowed to point to?  For sure A and AAAA, and SRV, and other CNAMEs...
I also put MX and TXT in there... but i suspect CNAMES can target pretty much any type record... in which case i'll need to handle every record type in rrDiff()...

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] New plugin fo..." ]
>I think this is pretty much done...
>
>@miek: what kind of records are CNAMEs allowed to point to?  For sure A and AAAA, and SRV, and other CNAMEs...
>I also put MX and TXT in there... but i suspect CNAMES can target pretty much any type record... in which case i'll need to handle every record type in rrDiff()...

cname points to a *name*

chrisohaver:
> cname points to a *name*

Um... So ... all types of records have a name, so, a CNAME can point to any type of record is what you are saying?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] New plugin fo..." ]
>> cname points to a *name*
>
>Um... So ... all types of records have a name, so, a CNAME can point to any type of record is what you are saying?

yep

plugin/resolve/README.md
```diff
@@ -0,0 +1,51 @@
...
+    resolve .
+}
+~~~
+
+Enable `CNAME` target resolution for all zones, but disable `SRV` target resolution.
+
+~~~ corefile
+. {
+    resolve . {
+        no srv
```
miekg, line 48: should we just make this: `exclude TYPE [TYPE]` ?

plugin/resolve/resolve.go
```diff
@@ -0,0 +1,123 @@
...
+type Resolve struct {
+	Next    plugin.Handler
+	Zones   []string
+	DoCNAME bool
+	DoSRV   bool
+}
+
+// Name implements the Handler interface.
+func (c Resolve) Name() string { return name() }
+func name() string             { return "resolve" }
```
miekg, line 24: why this indirection?

plugin/resolve/resolve.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/coredns/coredns/request"
+
+	"github.com/miekg/dns"
+)
+
+// Resolve performs CNAME target resolution on all CNAMEs in the response
+type Resolve struct {
+	Next    plugin.Handler
+	Zones   []string
+	DoCNAME bool
```
miekg, line 18: DoCNAME and DoServ probably don't need to be exported here?

plugin/resolve/resolve.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// Run the query through the rest of the plugin chain using a non-writer
+	nw := nonwriter.New(w)
+	rcode, err := plugin.NextOrFailure(c.Name(), c.Next, ctx, nw, r)
+	if err != nil || !plugin.ClientWrite(rcode) {
+		return rcode, err
+	}
+
+	// Look at each answer and do lookups for any CNAME/SRV answers
+	for i := 0; i < len(nw.Msg.Answer); i++ {
```
miekg, line 47: this should probably be done in parallel? (at some point??)

plugin/resolve/resolve.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil || target.Msg == nil || !plugin.ClientWrite(rcode) {
+		return nil
+	}
+	return target.Msg.Answer
+}
+
+// addTarget adds the answers from 'target' to the answers in 'clientResponse' ensuring that targets are not already in the
+// client response (not creating duplicates)
+func addTarget(clientRR, targetRR []dns.RR) []dns.RR {
+	for _, t := range targetRR {
```
miekg, line 82: This is, I think, O(n^2) and should be avoided. It's already called from a loop, so we're up to O(n^3), that stops performing pretty quickly.
(Related a benchmark test would help here as ll)

plugin/resolve/resolve.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		if unique {
+			clientRR = append(clientRR, t)
+		}
+	}
+	return clientRR
+}
+
+// rrDiff returns true if the two dns.RR are different in name, type, or target
+func rrDiff(a, b dns.RR) bool {
```
miekg, line 99: would dns.IsDuplicate help here?

plugin/resolve/README.md
chrisohaver: I dont care either way... I'm not as particular.

plugin/resolve/resolve.go
chrisohaver: Allows us to define the literal string "resolve" once.  It's a pattern followed in a few other plugins. It could also be a constant... or just have the literal name of the plugin pasted in all over the place like we do in most plugins.  IMO many valid ways to do the same thing.

plugin/resolve/resolve.go
chrisohaver: Perhaps - but for CNAMES at least I think usually you only get one CNAME in a response.  Exceptions are corner case (i.e. wildcard queries).

For SRV, you can get multiples if you leave the service protocol prefixes off the query name.  But not sure how common that really is. 

So yeah - theoretically would be better to do them in parallel, but not sure how much practical benefit we'd see.

plugin/resolve/resolve.go
chrisohaver: I think it also compares TTL...

plugin/resolve/resolve.go
chrisohaver: ... Should we abandon if I cant get it below O(n^3)?

chrisohaver:
@miek - you voiced concern again over performance.  I don't think we can avoid looping through each result to avoid adding duplicates.  Should we abandon the PR?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] New plugin fo..." ]
>+func (c Resolve) Name() string { return name() }
>+func name() string             { return "resolve" }
>
>Allows us to define the literal string "resolve" once.  It's a pattern followed in a few other plugins. It could also be a constant... or just have the literal name of the plugin pasted in all over the place like we do in most plugins.  IMO many valid ways to do the same thing.

Sure, but I think 'resolve' is used twice and now the reader needs to follow the 
call chain?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] New plugin fo..." ]
>
>
>
>> +// queryTarget looks up records for the qname by querying against the plugin chain, using a non-writer
>+func (c Resolve) queryTarget(state request.Request, qName string, qType uint16) []dns.RR {
>+	target := nonwriter.New(state.W)
>+	r := state.Req.Copy()
>+	r.SetQuestion(qName, qType)
>+	rcode, err := plugin.NextOrFailure(c.Name(), c.Next, state.Context, target, r)
>+	if err != nil || target.Msg == nil || !plugin.ClientWrite(rcode) {
>+		return nil
>+	}
>+	return target.Msg.Answer
>+}
>+
>+// addTarget adds the answers from 'target' to the answers in 'clientResponse' ensuring that targets are not already in the
>+// client response (not creating duplicates)
>+func addTarget(clientRR, targetRR []dns.RR) []dns.RR {
>+	for _, t := range targetRR {
>
>... Should we abandon if I cant get it below O(n^3)?

yes, totally.

I think a map can help, and not doing a query once we noticed we've seen the the 
target name

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] New plugin fo..." ]
>@miek - you voiced concern again over performance.  I don't think we can avoid looping through each result to avoid adding duplicates.  Should we abandon the PR?

A map may help (as said), but if you - as the author of this PR - think it will 
be hard - yes, we shoud abandon.

Issue #1899, opened 2018-06-27
Forward plugin sometimes returns FORMERR


zloo:
Hi there,

we're using coredns in our kubernetes clusters as the resolver. We have a rather basic setup:
```
    . {
      errors
      health
      prometheus 0.0.0.0:9153
      log . "{remote}:{port} - [{when}] {>id} {type} {class} {name} {proto} {size} {>do} {>bufsize} {rcode} {>rflags} {rsize} {duration}" {
        class error
      }
      cache 30
      kubernetes tt-k8s1.ko.seznam.cz 172.16.0.0/12 {
        tls /coredns/tls/coredns.pem /coredns/tls/coredns-key.pem /coredns/tls/ca.pem
        upstream 10.250.20.222 10.250.50.222
      }
      forward . 10.250.20.222 10.250.50.222
    }
```

with 10.250.20.222 and 10.250.50.222 being our internal DNS servers handling recursive DNS queries for the rest of the domains, including internal domains as well as all outgoing queries. 

We're using coredns 1.1.2 at the moment.

Sometimes FORMERR errors keep showing up for a few minutes, and they disappear after less than 10 minutes. They show up like this in the log:
```
10.64.38.104:59837 - [27/Jun/2018:12:13:14 +0200] 56625 A IN www.mapy.cz. udp 29 false 512 FORMERR qr,rd 29 205.372µs
10.66.206.49:37658 - [27/Jun/2018:12:12:51 +0200] 31387 AAAA IN grafana.com. udp 29 false 512 FORMERR qr,rd 29 219.374µs
```

It seems that the requests are served from cache because of the µs latencies - if they would try to connect to the upstream DNS servers, it would take at least a millisecond.

I managed to do a raw dig for a failing domain when the issue was happening:
```
$ dig puzzle.ng.seznam.cz @10.65.142.169

; <<>> DiG 9.10.3-P4-Ubuntu <<>> puzzle.ng.seznam.cz @10.65.142.169
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: FORMERR, id: 46786
;; flags: qr rd; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0
;; WARNING: recursion requested but not available

;; WARNING: EDNS query returned status FORMERR - retry with '+noedns'

;; QUESTION SECTION:
;puzzle.ng.seznam.cz.		IN	A

;; Query time: 1 msec
[...]

PR #1902, opened 2018-06-28
Add debug.Hexdump

miekg:
Allow plugins to dump messages in text pcap to the log. The forward
plugin does this when a reply does not much the query.

If the debug plugin isn't loaded Hexdump and Hexdumpf are noops.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,6 @@ Corefile
 coredns
 coredns.exe
 coredns.exe~
-debug
-debug.test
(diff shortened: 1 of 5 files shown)


miekg:
Not 100% convinced adding pcap actually delivers something useful or is just a gimmick.

We should debug log this though. I'll prepare another PR that adds those (and maybe a metric)

miekg:
merging to see if it can help with #1899

PR #1906, opened 2018-06-29
plugin/metadata: some cleanups

miekg:
Name to provider.go as that's what being defined right now in the file.
Use request.Request because that's done in variables.go anyway. Name the
main storage M, because there is no further meaning behind.

Remove superfluous methods.

### DIFF
--- a/plugin/metadata/metadata.go
+++ b/plugin/metadata/metadata.go
@@ -24,15 +24,16 @@ func (m *Metadata) Name() string { return "metadata" }
 // ServeDNS implements the plugin.Handler interface.
 func (m *Metadata) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 
-	md, ctx := newMD(ctx)
+	ctx = context.WithValue(ctx, metadataKey{}, M{})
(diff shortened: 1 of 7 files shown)


fturib (commented):
LGTM.


For my information:
Why ServeDNS had the 2 parameters ResponseWriter and Msg, and for Metadata you prefer to send the Request ?

Is that Request was created later and we keep the signature of ServeDNS ? or these 2 data :  Request vs couple ResponseWriter+Msg are not equivalent ?

miekg:
>
>LGTM.
>
>
>For my information:
>Why ServeDNS had the 2 parameters, and for Metadata you prefer to send the Request ?
>
>Is that Request was created later and we keep the signature of ServeDNS ? or these 2 thinks Request vs couple Write+Msg are not equivalent ?

Most plugins use request.Request because adds a nice abstraction and caches 
things like the lowercase Qname (and edns0), so it should be slightly faster.

As I tried to say in the commit msg, in variables.go request and response writer 
were wrapped in a request.Request anyway, which is sortof expensive if you do 
this for every variable checked.

/Miek

--
Miek Gieben

PR #1908, opened 2018-06-29
WIP: rework metadata

miekg:

### DIFF
--- a/plugin/metadata/metadata.go
+++ b/plugin/metadata/metadata.go
@@ -31,9 +31,16 @@ func (m *Metadata) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 	if plugin.Zones(m.Zones).Matches(state.Name()) != "" {
 		// Go through all Providers and collect metadata.
 		for _, provider := range m.Providers {
-			for _, varName := range provider.MetadataVarNames() {
-				if val, ok := provider.Metadata(ctx, state, varName); ok {
-					md.SetValue(varName, val)
+			for _, v := range provider.VarStrings() {
+				val := provider.String(ctx, state, v)
+				if val != "" {
+					md.SetString(v, val)
+				}
+			}
+			for _, v := range provider.VarInts() {
+				val := provider.Int(ctx, state, v)
+				if val != -1 {
+					md.SetInt(v, val)
 				}
 			}
 		}
@@ -44,13 +51,12 @@ func (m *Metadata) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 	return rcode, err
 }
 
-// MetadataVarNames implements the plugin.Provider interface.
-func (m *Metadata) MetadataVarNames() []string { return variables.All }
+// Metadata implements the plugin.Provider interface.
+func (m *Metadata) String(ctx context.Context, state request.Request, v string) string {
+	return variables.StringValue(state, v)
+}
 
 // Metadata implements the plugin.Provider interface.
-func (m *Metadata) Metadata(ctx context.Context, state request.Request, varName string) (interface{}, bool) {
-	if val, err := variables.GetValue(state, varName); err == nil {
-		return val, true
-	}
-	return nil, false
+func (m *Metadata) Int(ctx context.Context, state request.Request, v string) int {
+	return variables.IntValue(state, v)
 }
--- a/plugin/metadata/provider.go
+++ b/plugin/metadata/provider.go
@@ -11,16 +11,27 @@ import (
 // Note: this method should work quickly, because it is called for every request
 // from the metadata plugin.
 type Provider interface {
-	// List of variables which are provided by current Provider. Must remain constant.
-	MetadataVarNames() []string
+	// List of variables which are provided by current Provider. Must not be changed during run-time.
+	VarInts() []string
+	VarStrings() []string
+
 	// Metadata is expected to return a value with metadata information by the key
 	// from 4th argument. Value can be later retrieved from context by any other plugin.
 	// If value is not available by some reason returned boolean value should be false.
-	Metadata(ctx context.Context, state request.Request, variable string) (interface{}, bool)
+	String(ctx context.Context, state request.Request, variable string) string
+	Int(ctx context.Context, state request.Request, variable string) int
 }
 
 // M is metadata information storage.
-type M map[string]interface{}
+type M struct {
+	s map[string]string
+	i map[string]int
+}
+
+// New returns a new initialized M.
+func New() M {
+	return M{s: make(map[string]string), i: make(map[string]int)}
+}
 
 // FromContext retrieves the metadata from the context.
 func FromContext(ctx context.Context) (M, bool) {
@@ -32,15 +43,26 @@ func FromContext(ctx context.Context) (M, bool) {
 	return M{}, false
 }
 
+// String returns metadata string value by key.
+func (m M) String(key string) (string, bool) {
+	s, ok := m.s[key]
+	return s, ok
+}
+
+// SetString sets the metadata value under key.
+func (m M) SetString(key, s string) {
+	m.s[key] = s
+}
+
 // Value returns metadata value by key.
-func (m M) Value(key string) (value interface{}, ok bool) {
-	value, ok = m[key]
-	return value, ok
+func (m M) Int(key string) (int, bool) {
+	i, ok := m.i[key]
+	return i, ok
 }
 
 // SetValue sets the metadata value under key.
-func (m M) SetValue(key string, val interface{}) {
-	m[key] = val
+func (m M) SetInt(key string, i int) {
+	m.i[key] = i
 }
 
 // metadataKey defines the type of key that is used to save metadata into the context.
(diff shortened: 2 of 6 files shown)


miekg:
This is (I think) where should head with this metadata thing.

this at least makes it type checked.

I've haven't fixed rewrite yet, because I won't to see if we can agree and where we're heading.

Regardless of implementation; it's pretty hard to understand what kind of variables are supported by the metadata plugin.

johnbelamaric:
Yeah, this could be good. I want to be able to support ripping meta data fields out of EDNS0 options. This would become options on the metadata plugin, and would be at least functionally equivalent to what we can do here:

https://github.com/infobloxopen/themis/tree/master/contrib/coredns/policy#syntax

Those are typed as well. It reads in the data in one format and then provides the data in another format. So, incoming EDNS0 option could have data that is: raw bytes, a hex string, or an IP. We could extend that with an ordinary int too. Then the variable(s) that are extracted also get a type, currently only string and address are supported, and that plugin translates the inbound type to the outbound type. In our case we should decide if we want to convert to different types here or leave it to the consumer. If we leave it up to the consumer then "hex" probably isn't necessary - it's just a string and the consumer would have to know to treat it as hex.

plugin/metadata/provider.go
```diff
@@ -11,16 +11,27 @@ import (
...
+	// List of variables which are provided by current Provider. Must not be changed during run-time.
+	VarInts() []string
+	VarStrings() []string
+
 	// Metadata is expected to return a value with metadata information by the key
 	// from 4th argument. Value can be later retrieved from context by any other plugin.
 	// If value is not available by some reason returned boolean value should be false.
-	Metadata(ctx context.Context, state request.Request, variable string) (interface{}, bool)
+	String(ctx context.Context, state request.Request, variable string) string
+	Int(ctx context.Context, state request.Request, variable string) int
```
fturib, line 22: How will it work here when we find we would like a new type ? 
=> need to add 2 new functions to the interface Provider
Does it break the compatibility with all the already implemented plugin that will miss these 2 new interfaces ?
I guess yes ...

Would it be better, if we want to have typed metadata, to define enumeration of types and save the Name + Type + Value ?

And maybe the Metadata can offer the service to translate from Type to other Type of at least to storage (in the Getter/Setter)

Doing so, we keep only interface stable with only 2 functions:
- MetadataVars() => return array of (String, enumType)
- Metadata(ctx context.Context, state request.Request, variable string) (interface{}, bool) - here the type is already known.

and Metadata implement the convenient function for getting / setting the values. And these can be extended without changing the interface Provider.

What do you think ?

plugin/metadata/metadata.go
```diff
@@ -31,9 +31,16 @@ func (m *Metadata) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
...
+			for _, v := range provider.VarStrings() {
+				val := provider.String(ctx, state, v)
+				if val != "" {
+					md.SetString(v, val)
+				}
+			}
+			for _, v := range provider.VarInts() {
+				val := provider.Int(ctx, state, v)
+				if val != -1 {
+					md.SetInt(v, val)
```
fturib, line 43: See other comment.

My fear is also here : for the plugins that need data, but do not care the type, it will have to loop over each type to get the right info ...

fturib: Although here we could implement a convenient function ... but that will come as the same as defining a enumType for the variable.

fturib (commented):
I understand the point to have typed variables. But I was wondering on the lost of genericity,

and MOSTLY about compatibility when we want to implement a new type.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: rework m..." ]
>
>I understand the point to have typed variables. But I was wondering on the lost of genericity,
>
>and MOSTLY about compatibility when we want to implement a new type.
>
>>  	// Metadata is expected to return a value with metadata information by the key
> 	// from 4th argument. Value can be later retrieved from context by any other plugin.
> 	// If value is not available by some reason returned boolean value should be false.
>-	Metadata(ctx context.Context, state request.Request, variable string) (interface{}, bool)
>+	String(ctx context.Context, state request.Request, variable string) string
>+	Int(ctx context.Context, state request.Request, variable string) int
>
>How will it work here when we find we would like a new type ?
>=> need to add 2 new functions to the interface Provider
>Does it break the compatibility with all the already implemented plugin that will miss these 2 new interfaces ?
>I guess yes ...
>
>Would it be better, if we want to have typed metadata, to define enumeration of types and save the Name + Type + Value ?
>
>And maybe the Metadata can offer the service to translate from Type to other Type of at least to storage (in the Getter/Setter)
>
>Doing so, we keep only interface stable with only 2 functions:
>- MetadataVars() => return array of (String, enumType)
>- Metadata(ctx context.Context, state request.Request, variable string) (interface{}, bool) - here the type is already known.
>
>and Metadata implement the convenient function for getting / setting the values. And these can be extended without changing the interface Provider.
>
>What do you think ?

Yeah, I think you make some good points. Currently the discoverability 
of the variables are supported is not that good. The variables package is also 
not sitting well with me, but I don't fully understand why I think that.
(Just filed https://github.com/coredns/coredns/issues/1911).

Basically, there are two things:

1. Metadata adds something to the metadata - this is trivial
2. Plugins need to retrieve all, or some data from the metadata (and need to 
   know the type of the data they get - this is not trivial

How does the plugin know what metadata is there and what type has it?

Say plugin X adds a `time.Time` to the metadata and names it 'epoch'.

Then plugin Y either knows fully about epoch and the time.Time type, or it can 
find out runtime. The variables package is an attempt to fix the latter, but 
then omitting the types. I think you need a run time registry, just like the 
plugin registration for new metadata variables or something.

fturib:
About: 

> I think you need a run time registry, just like the
> plugin registration for new metadata variables or something.


In fact initial idea was only to collect ALL metadata that any plugin can provide about the incoming DNS query,  into a Map and let plugin Y (consumer) check what if what it needs is there ...

**idea is that we do not know what metadata we will get before the query exist.**
=> registration can happen ONLY query per query ....

introduction of initial "getMetadataName" is to allow a lazzy initialization for these metadata : at first your provide only the name .. and when one consumer NEED really that data, then we collect it. That is rather on optimization focus.

I think that "metadata" .. is all information we know about the query and its context.
the consumer cannot assume that metadata will always be there: it needs to check if this one exist or this one.
Example: we are thinking, for k8s, providing the labels attached to the pod or service. That would allow a plugin to filter the queries based on those labels.

There is kind of contract between Plugins that go by the configuration Corefile.
it is documentation on Plugin Y Consumer : I am able to process those metadata ... and they are usually provided by theses Plugin X, Z.

For the Type, we can keep simple - when we get the value, we get is into a standard format (let say a string or byte buffer), we also get the type and the plugin Metadata provides the convenient way to encode/decode the data from/to its original type.

We can get a small structure that embed the value encoded + type + way to decode it to standard types (string, int, ...)

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: rework m..." ]
>About:
>
>> I think you need a run time registry, just like the
>> plugin registration for new metadata variables or something.
>
>In fact initial idea was only to collect ALL metadata that any plugin can provide about the incoming DNS query,  into a Map and let plugin Y (consumer) check what if what it needs is there ...

But duplicating data that's easily extracted (and much faster!) from the request 
and response writer is the wrong way to go.

>I think that "metadata" .. is all information we know about the query and its context.
>the consumer cannot assume that metadata will always be there: it needs to check if this one exist or this one.

Yes, but how? 

>Example: we are thinking, for k8s, providing the labels attached to the pod or service. That would allow a plugin to filter the queries based on those labels.

Yep, totally legit. This is, like, real metadata and impossible to get from the 
query (directly).

>There is kind of contract between Plugins that go by the configuration Corefile.
>it is documentation on Plugin Y Consumer : I am able to process those metadata ... and they are usually provided by theses Plugin X, Z.

That's a really brittle way of doing it. Is also weird: *adding* metdata is 
really neat; you just implement an interface; *Getting* the requires you reading 
some docs. Think such getters could be added to an structure that is:

map[string]func(...)

Where the key would be plugin's name adding the function. That function can then 
be called to return the value (if they are there).
[could be extended to a slice of functions)

>For the Type, we can keep simple - when we get the value, we get is into a standard format (let say a string or byte buffer), we also get the type and the plugin Metadata provides the convenient way to encode/decode the data from/to its original type.

Yep. I think string will be most useful.

/Miek

--
Miek Gieben

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: rework m..." ]
>About:
>
>> I think you need a run time registry, just like the
>> plugin registration for new metadata variables or something.

Just had another idea; what we are implementing here is basically
labelling the query. I.e. just like k8s labels. We could try to do it exactly 
like that.

I.e.

map[string]string

has the labels.

Labels() []string:  returns all keys (sorted?)
Value(label string) string: return the value for label

Yes, you still need to define the label name, but is does remove the need for 
variables.go. But also depends on reading the configuration on what labels are 
added by a plugin.

miekg:
closing in favor of #1914

Issue #1901, opened 2018-06-27
auto plugin complains about ConfigMap mounted volumes


particledecay:
To be clear, this is only when a ConfigMap is mounted without specifying individual items.

```yaml
    containers:
      - name: my-container
        volumeMounts:
          - name: zonedata
            mountPath: /etc/myzones
    volumes:
      - name: zonedata
        configMap:
          name: my-config
```

The files that are placed into `/etc/myzones` by Kubernetes may look like this:
```
/etc/myzones/example.com.db
/etc/myzones/example.org.db
```

However, those files are only symlinks to a directory whose name is a timestamp (set to the last datetime that the ConfigMap was updated):
```
/etc/myzones/example.com.db -> /etc/myzones/..data/example.com.db
/etc/myzones/..data -> /etc/myzones/..2018_06_27_07_11_05.442838727
```

Because the auto plugin performs a "walk", it will match the files in the timestamped subdirectory instead of the symlink (since `filepath.Walk` does not follow symbolic links). This means that as soon as the ConfigMap is updated and all the top-level symlinks are repointed to the latest timestamped subdirectory, CoreDNS will emit log messages stating that it can no longer find the zone that it previously read in the no-longer-existent files:

```
2018/06/27 14:46:30 [ERROR] plugin/file: Failed to open zone "example.com." in "/etc/myzones/..2018_06_27_07_11_05.442838727/example.com.db": open /etc/myzones/..2018_06_27_07_11_05.442838727/example.com.db: no such file or directory
```

This makes sense, because the file no longer exists. However, a new file now exists:
```
/etc/myzones/..2018_06_27_07_21_08.389576271/example.com.db
```

chrisohaver:
Is CoreDNS is also double-processing each file?  Perhaps...

I suppose that if the links pointed to files in another non-subdirectory, this would not be an issue.
Still, we should handle this better.

chrisohaver:
> Is CoreDNS is also double-processing each file?

... and all previous timestamp dirs, if they are left there...

particledecay:
No, you're not processing the symlinks at all, because `filepath.Walk` doesn't care about them. Also, the old timestamped directories don't stick around, they go away.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] auto plugin compl..." ]
>Because the auto plugin performs a "walk", it will match the files in the 
>timestamped subdirectory instead of the symlink (since `filepath.Walk` does not 
[...]

PR #1910, opened 2018-06-30
Ensuring updated file path in `auto.Walk`

particledecay:
When using the "auto" plugin in conjunction with a Kubernetes ConfigMap-mounted volume, the files within the mounted directory are all actually symlinks. Every time the ConfigMap is updated, the symlinks are repointed to the latest files. This currently causes CoreDNS to complain about losing a file and it will not re-read a new file with the same zone information.

This PR updates the file path when an `auto.Walk` is performed, to ensure that the most up-to-date file is always being read.

[#1901](https://github.com/coredns/coredns/issues/1901) 

None. A test was added specifically for this case.

### DIFF
--- a/plugin/file/zone.go
+++ b/plugin/file/zone.go
@@ -124,6 +124,20 @@ func (z *Zone) Insert(r dns.RR) error {
 // Delete deletes r from z.
 func (z *Zone) Delete(r dns.RR) { z.Tree.Delete(r) }
 
+// File retrieves the file path in a safe way
+func (z *Zone) File() string {
+	z.reloadMu.Lock()
+	defer z.reloadMu.Unlock()
+	return z.file
+}
+
+// SetFile updates the file path in a safe way
+func (z *Zone) SetFile(path string) {
+	z.reloadMu.Lock()
+	z.file = path
+	z.reloadMu.Unlock()
+}
+
 // TransferAllowed checks if incoming request for transferring the zone is allowed according to the ACLs.
 func (z *Zone) TransferAllowed(state request.Request) bool {
 	for _, t := range z.TransferTo {
--- a/plugin/auto/walk.go
+++ b/plugin/auto/walk.go
@@ -31,9 +31,10 @@ func (a Auto) Walk() error {
 			return nil
 		}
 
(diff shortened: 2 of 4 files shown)


plugin/file/zone.go
```diff
@@ -56,6 +56,16 @@ func NewZone(name, file string) *Zone {
 	return z
 }
 
+// GetPath returns the currently known file path
+func (z *Zone) GetPath() string {
+	return z.file
```
miekg: If we export File we don't need these getter and setters

miekg (commented):
Thanks. Only one comment from me

plugin/file/zone.go
yongtang: +1 to avoid getter and setters.

particledecay: Yep, makes sense, I just wasn't sure how you guys felt about exporting that property. Will revise.

particledecay:
Done.

miekg:
thanks.

The tests fail with a data race though

particledecay:
@miekg sorry, golang n00b here. I'll fix and squash.

particledecay:
Couldn't get rid of the getter/setter... didn't want to export `reloadMu` just for the `Lock` and `Unlock`, so felt like it was safer to read and write to `Zone.File` by unexporting `file` and using the aforementioned locks instead.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Ensuring upda..." ]
>@miekg sorry, golang n00b here. I'll fix and squash.

no problem. Glad we have these tests :)

plugin/file/zone.go
```diff
@@ -124,6 +124,20 @@ func (z *Zone) Insert(r dns.RR) error {
 // Delete deletes r from z.
 func (z *Zone) Delete(r dns.RR) { z.Tree.Delete(r) }
 
+// GetFile retrieves the file path in a safe way
+func (z *Zone) GetFile() string {
```
miekg: Usually In Go you drop the Get, so this just becomes File.

other than that, lgtm

miekg (approved):

Issue #1913, opened 2018-06-30
plugin/rewrite: use request.Request


miekg:
Quite a lot of code can be deleted from rewrite if we move the request.Request when rewritting, most of the helper functions it provides are written again in rewrite.

miekg:
Closed via 6dd2cf8c

PR #1914, opened 2018-06-30
plugin/metadata: metadata is just label=value

miekg:
This reverts 17d807f0 and re-adds the metadata plugin as a plugin that
just sets a label to a value.
There are now - no in tree users. We could add the request metadata by
default under names that copy request.Request, i.e

request/ip - remote IP
request/port - remote port

Variables.go has been deleted.


miekg:
@fturib @johnbelamaric 

New PR: I think this captures better what we want. In the rewrite plugin we could use *labels* to fetch the metadata. I.e `client_port` return the client IP right now, we could use the syntax `request/client_port` to signal it's a *label* and we should check the metadata.

fturib:
So, what I understand is that we are back to the initial proposition from @ekleiner.

### Good points:
- With a nicer set of functions to manage the metadata from context (Labels, Value, ...)
- There is a concept of namespace (request/ip).

### Warning:
- REWRITE is back to original. So we will need to modify in order to add the lookup into the Metadata. But maybe that should happen after the refactoring #1913.

## QUESTIONS
- what plugin will provide the metadata for requests (ip, port, etc ...) ? John suggested to be Metadata plugin, but you have another ides ?
- will it be mandatory for Metadataer to present Labels within a namespace ?
- do we abandon idea of performance. ? [see comment from JohnB](https://github.com/coredns/coredns/pull/1894#issuecomment-400852405)
 I mean the way it is build, we will not be able to have a lazy init. and therefore all metadata will be built even though it is useless. Could the value of the Label be a getter to the value ?

miekg:
Thanks for reviewing.

>So, what I understand is that we are back to the initial proposition from @ekleiner.
>
>### Good points:
>- With a nicer set of functions to manage the metadata from context (Labels, Value, ...)
>- There is a concept of namespace (request/ip).
>
>### Warning:
>- REWRITE is back to original. So we will need to modify in order to add the lookup into the Metadata. But maybe that should happen after the refactoring #1913.

yes, I think we need to make the destinction here between labels (they have a 
slash in them) and the current rewrite variables. 

>## QUESTIONS
>- what plugin will provide the metadata for requests (ip, port, etc ...) ? John suggested to be Metadata plugin, but you have another ides ?

none - use request.Request. We could also add them in metadata as 
request/client_ip and etc.

>- will it be mandatory for Metadataer to present Labels within a namespace ?

I think we should enforce that, yes.

>- do we abandon idea of performance. ? [see comment from JohnB](https://github.com/coredns/coredns/pull/1894#issuecomment-400852405)
> I mean the way it is build, we will not be able to have a lazy init. and therefore all metadata will be built even though it is useless. Could the value of the Label be a getter to the value ?

I don't think I fully understand John's comment: What's evaluated more than once 
here?

fturib:
> I don't think I fully understand John's comment: What's evaluated more than once
> here?

the comment of John was:

> I guess I thought Eugene was going to rework this to evaluate the variables **only on use**?

### expose the problem
The idea is to avoid to have all the Medataer compute the values for the metadatas that are not consumed. Or at least to be able to have a 'clever' way to collect the values.

That means, that instead of asking all metadataer to provide blindly the labels=value at beginning or query processing, it was proposed to have a 2 steps process:

1- at begin of query processing, metadataer provide a Label + a getter to the value
2- when the value is needed by any consumer, the getter is triggered to get the value.

The problem if that even if we want to implement that later, we should prepare the interface of Metadater for a 2 phase.

### Proposition to implement lazzy init, keeping the same metadater interface
Hum ... there is maybe a way, if the metadata is able to detect that the value returned is a Getter.
That would need to implement the following (and we can do later in another PR):

- type Getter interface {
     GetValue(label string) string
}
- SetGetterValue(ctx context.Context, label string, getter Getter) bool
- and modify Value to verify if the value is a string or a Getter, and trigger the call in later case and register the value returned

### At the end 
I think we can delay the optimization. There should not be change of interface. Only change in the functions provided in the metadata plugin.

fturib:
About 

> none - use request.Request. We could also add them in metadata as
> request/client_ip and etc.

So finally we can either use directly request.Request (like in Rewrite).
but metadata plugin will provide a "string" version with names like request/ip, request/port .. etc.

Seems ok for me.

miekg:
>> I don't think I fully understand John's comment: What's evaluated more than once
>> here?
>
>the comment of John was:
>
>> I guess I thought Eugene was going to rework this to evaluate the variables **only on use**?

Ah, ok. Then we should not add the value, but a function that returns the value. If 
you're interested in that value you call the function. Trivial change.

miekg:
@fturib ptal

fturib:
*** Review in on PR 1920
Ooh ... looks like I did the review on this on this PR : https://github.com/coredns/coredns/pull/1920/files

it also contains changes for metadata and rewrite.

Is that ok or should I report the comments for metadata on this PR ?

*** In Summary
metadata :
- proposition to include the cache mechanism in the context, it would ensure
-- ensure the value is always returned the same
-- easier for the provider : just give a function that fetch/compute effectively the value
-- easier for consumer : just get the value. do not bother with a fetcher.
-- race conditions to access the value are managed (by metadata)

- question on the name (metadataer or provider ?)

miekg:
I need to rebase 1920 once this is merged (branched from this pr)

miekg:
* rewrite in it current form should not use metadata; why make the common use case slower?proposition 

>
> -- ensure the value is always returned the same
> -- easier for the provider : just give a function that fetch/compute effectively the value

fair enough, but it smells like premature optimization.

> -- easier for consumer : just get the value. do not bother with a fetcher.

you call a function are you get the value; it's need to be lazy anyway...

> -- race conditions to access the value are managed (by metadata)

This could prove problematic indeed. I would initially just go; not thread safe safe, unless the docs
say so.

> question on the name (metadataer or provider ?)

thought the names made sense during the changes I made (i.e. no change from the original)

miekg:
I'll look at your comments on #1920 and apply them here, sorry 'bout that.

miekg:
you mostly comment on the caching behavior; which is fair, but should probably be done in another PR - I like to see some metadata impl. that are more complex and come back to this caching subject.

fturib:
I am ok with this implementation.
I agree we can do an later PR:
- the cache behavior if needed
- extend Rewrite as a consumer of metadata, accepting labels name as variables for edns0 local (this is our use case: extend the ability of Rewrite, allowing to encode in edns0 some values that are provided by other plugins using the Provider interface and metadata plugin). To do after PR #1920 is merged.
- we have already one implementation the MAC address of the client - unfortunately implemented as external plugin in a private repo of IB. need to find a way to show it to you ...

fturib (approved):
> you mostly comment on the caching behavior; which is fair, but should probably be done in another PR
=> I agree we can do later

> wou

miekg:
>I am ok with this implementation.
>I agree we can do an later PR:
>- the cache behavior if needed

+1

>- extend Rewrite accepting labels name as variables for edns0 local (that is our use case: extend the ability of Rewrite, allowing to encode in edns0 some values that are provided by other plugins using the Provider interface and metadata plugin).

+1 (if we name labels differently i.e. with `name/something` could be done quite 
elegantly - if not I'm open to use metadata for everything.

>- we have already one implementation the MAC address of the client - unfortunately implemented as external plugin in a private repo of IB. need to find a way to show it to you ...

Ack. If you're allowed to send me a zip file I'll be happy to take a look.

/Miek

--
Miek Gieben

PR #1920, opened 2018-07-01
plugin/rewrite: make code more idomatic

miekg:
Cleanup rewrite a bit and use request.Request everywhere.

Can't be merged until #1914 is in and there is one test failing.


miekg:
This fails a single test:
~~~
--- FAIL: TestRewriteEDNS0LocalVariable (0.00s)
	rewrite_test.go:523: Test 6: Expected [65518:0x7f000001] but got 
		;; OPT PSEUDOSECTION:
		; EDNS: version 0; flags: do; udp: 4096
		; LOCAL OPT: 65518:0x3132372e302e302e31
FAIL
exit status 1
~~~
can't figure out why

plugin/metadata/provider.go
```diff
@@ -1,3 +1,33 @@
...
+//       f := func() string {
+//		if cached != "" {
+//                 return cached
+//             }
+//             cached = expensiveFunc()
+//             return cached
+//       }
+//       metadata.SetValueFunc(ctx, "test/something", f)
+//	 return ctx
+//    }
```
fturib: YES!. the inner function is a good utility that could be provided as utility for any metadater. 
Better maybe : could the cache be in the map of the context ?
=> I mean SetValueFunc(ctx, name, fetchingFunction)

fturib: There is maybe pro and cons .. 
But I am thinking that then we would have one place only to manage the cache and maybe access to this value (race conditition ?)

plugin/metadata/provider.go
```diff
@@ -8,40 +38,62 @@ import (
...
-	// List of variables which are provided by current Provider. Must remain constant.
-	MetadataVarNames() []string
-	// Metadata is expected to return a value with metadata information by the key
-	// from 4th argument. Value can be later retrieved from context by any other plugin.
-	// If value is not available by some reason returned boolean value should be false.
-	Metadata(ctx context.Context, state request.Request, variable string) (interface{}, bool)
+	// Metadata adds metadata to the context and returns a (potentially) new context.
+	// Note: this method should work quickly, because it is called for every request
+	// from the metadata plugin.
+	Metadata(ctx context.Context, state request.Request) context.Context
```
fturib: Yes. In that case, should we go back to the standard of go, and name the interface "Metadataer" like John was suggesting a the beginning ? (like healther ..)

plugin/metadata/provider.go
```diff
@@ -8,40 +38,62 @@ import (
...
+	return nil
 }
 
-// Value returns metadata value by key.
-func (m M) Value(key string) (value interface{}, ok bool) {
-	value, ok = m[key]
-	return value, ok
+// ValueFunc returns the value function of label. If none can be found nil is returned. Calling the
+// function returns the value of the label.
+func ValueFunc(ctx context.Context, label string) Func {
```
fturib: if we set the cache in the context (automatically). Then this method is back to "Value", and the caller do not care to call or not the fetching function.

plugin/metadata/provider.go
```diff
@@ -8,40 +38,62 @@ import (
...
+			return true
+		}
+	}
+	return false
 }
 
-// metadataKey defines the type of key that is used to save metadata into the context.
-type metadataKey struct{}
+// md is metadata information storage.
+type md map[string]Func
```
fturib: here we could have a little more complex:

cache struct {
    value string
    inited bool
    fetcher Func
}

type md map[string] cache

plugin/rewrite/README.md
```diff
@@ -206,17 +206,13 @@ rewrites the first local option with code 0xffee, setting the data to "abcd". Eq
...
+* A variable data is specified with a pair of curly brackets `{}`. Following are the supported variables:
   {qname}, {qtype}, {client_ip}, {client_port}, {protocol}, {server_ip}, {server_port}.
-Any plugin that can provide it's own additional variables by implementing metadata.Provider interface. If you are going to use metadata variables then metadata plugin must be enabled.
 
 Example:
 
-~~~ corefile
-. {
-    metadata
-    rewrite edns0 local set 0xffee {client_ip}
```
fturib: At the end, we will want with "metadata" (because we expect to be able to add some of the metadata in the edns0 records).

Can be in second time, once this refactoring is in.

plugin/rewrite/edns0.go
```diff
@@ -315,45 +369,38 @@ func (rule *edns0SubnetRule) fillEcsData(w dns.ResponseWriter, r *dns.Msg, ecs *
...
+	if rule.action == Append || rule.action == Set {
+		opt := &dns.EDNS0_SUBNET{Code: dns.EDNS0SUBNET}
+		if rule.fillEcsData(state, opt) == nil {
+			o.Option = append(o.Option, opt)
+			return RewriteDone
 		}
 	}
 
-	return result
+	return RewriteIgnored
```
fturib, line 336: changes looks ok.

plugin/rewrite/edns0.go
```diff
@@ -75,24 +72,19 @@ Option:
...
+// Rewrite will alter the request EDNS0 local options.
+func (rule *edns0LocalRule) Rewrite(state request.Request) Result {
 	result := RewriteIgnored
-	o := setupEdns0Opt(r)
+	o := setupEdns0Opt(state.Req)
 	found := false
 	for _, s := range o.Option {
-		switch e := s.(type) {
-		case *dns.EDNS0_LOCAL:
+		if e, ok := s.(*dns.EDNS0_LOCAL); ok {
```
fturib: here same refactor with using break and returns could happen.

plugin/metadata/provider.go
fturib: WARNING: I guess we need to support the value "" case => which means maybe have a boolean to remember the cache is set or not.

fturib (commented):
metadata : 
- proposition to include the cache mechanism in the context
- question on the name

For Rewrite : that looks good, although difficult to verify each change. I hope the UT are good enough :)

plugin/rewrite/edns0.go
miekg: done

miekg:
@fturib found why the test was failing, so that is fixed. Added some more cleanups.

Should be good to go, PTAL

fturib (approved):
LGTM

PR #1921, opened 2018-07-01
request.Request: cache a few more value

miekg:
Cache IP's and ports as well.

### DIFF
--- a/request/request.go
+++ b/request/request.go
@@ -26,8 +26,13 @@ type Request struct {
 	do   *bool // nil: nothing, otherwise *do value
 	// TODO(miek): opt record itself as well?
 
-	// Cache lowercase qname.
-	name string
+	// Caches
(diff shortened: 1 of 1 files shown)


SuperQ (approved):
LGTM

cobb-tx:
@miekg  hi  I noticed that the r.ip was always empty  and  net.SplitHostPort(r.W.RemoteAddr().String()) give me the docker0 network card‘s ip

I expect IP then 172.17.0.11  is pod ip not docker0 ip

···
func (r *Request) IP() string {
	if r.ip != "" {
		return r.ip
	}

	ip, _, err := net.SplitHostPort(r.W.RemoteAddr().String())
	if err != nil {
		r.ip = r.W.RemoteAddr().String()
		return r.ip
	}

	r.ip = ip
	return r.ip
}
···

PR #1928, opened 2018-07-03
plugin/rewrite - extend edns0 local variable support with metadata

fturib:
### 1. add support of metadata for edns0 local variable ?
* allow to encode any metadata related to current query into edns0

Readme updated
UT provided.

#1638 

not sure here ...

### DIFF
--- a/plugin/metadata/metadata_test.go
+++ b/plugin/metadata/metadata_test.go
@@ -10,9 +10,9 @@ import (
 	"github.com/miekg/dns"
 )
 
-type StaticProvider map[string]Func
+type testProvider map[string]Func
 
-func (tp StaticProvider) Metadata(ctx context.Context, state request.Request) context.Context {
+func (tp testProvider) Metadata(ctx context.Context, state request.Request) context.Context {
 	for k, v := range tp {
 		SetValueFunc(ctx, k, v)
 	}
@@ -29,7 +29,7 @@ func (m *testHandler) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns
 }
 
 func TestMetadataServeDNS(t *testing.T) {
-	expectedMetadata := []StaticProvider{
+	expectedMetadata := []testProvider{
 		{"test/key1": func() string { return "testvalue1" }},
 		{"test/key2": func() string { return "two" }, "test/key3": func() string { return "testvalue3" }},
 	}
--- a/plugin/metadata/metadata_test.go
+++ b/plugin/metadata/metadata_test.go
@@ -47,15 +47,8 @@ func TestMetadataServeDNS(t *testing.T) {
 	}
 
 	ctx := context.TODO()
-	if IsMetadataSet(ctx) {
-		t.Errorf("Context provide Metadata information whereas metadata is not yet activated")
-	}
 	m.ServeDNS(ctx, &test.ResponseWriter{}, new(dns.Msg))
 	nctx := next.ctx
-
-	if !IsMetadataSet(nctx) {
-		t.Errorf("Context does not provide Metadata information")
-	}
 
 	for _, expected := range expectedMetadata {
 		for label, expVal := range expected {
@@ -76,11 +69,14 @@ func TestLabelFormat(t *testing.T) {
 		isValid bool
 	}{
 		{"plugin/LABEL", true},
+		{"p/LABEL", true},
+		{"plugin/L", true},
 		{"LABEL", false},
 		{"plugin.LABEL", false},
-		{"/NO-PLUGIN-ACCEPTED", true},
-		{"ONLY-PLUGIN-ACCEPTED/", true},
-		{"PLUGIN/LABEL/SUB-LABEL", true},
+		{"/NO-PLUGIN-NOT-ACCEPTED", false},
+		{"ONLY-PLUGIN-NOT-ACCEPTED/", false},
+		{"PLUGIN/LABEL/SUB-LABEL", false},
+		{"/", false},
 	}
 
 	for _, test := range labels {
--- a/plugin/metadata/metadata_test.go
+++ b/plugin/metadata/metadata_test.go
@@ -10,9 +10,9 @@ import (
 	"github.com/miekg/dns"
 )
 
-type testProvider map[string]Func
+type StaticProvider map[string]Func
 
-func (tp testProvider) Metadata(ctx context.Context, state request.Request) context.Context {
+func (tp StaticProvider) Metadata(ctx context.Context, state request.Request) context.Context {
 	for k, v := range tp {
 		SetValueFunc(ctx, k, v)
 	}
@@ -29,9 +29,9 @@ func (m *testHandler) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns
 }
 
 func TestMetadataServeDNS(t *testing.T) {
-	expectedMetadata := []testProvider{
-		testProvider{"test/key1": func() string { return "testvalue1" }},
-		testProvider{"test/key2": func() string { return "two" }, "test/key3": func() string { return "testvalue3" }},
+	expectedMetadata := []StaticProvider{
+		{"test/key1": func() string { return "testvalue1" }},
+		{"test/key2": func() string { return "two" }, "test/key3": func() string { return "testvalue3" }},
 	}
 	// Create fake Providers based on expectedMetadata
 	providers := []Provider{}
--- a/plugin/metadata/metadata_test.go
+++ b/plugin/metadata/metadata_test.go
@@ -47,15 +47,45 @@ func TestMetadataServeDNS(t *testing.T) {
 	}
 
 	ctx := context.TODO()
+	if IsMetadataSet(ctx) {
+		t.Errorf("Context provide Metadata information whereas metadata is not yet activated")
+	}
 	m.ServeDNS(ctx, &test.ResponseWriter{}, new(dns.Msg))
 	nctx := next.ctx
 
+	if !IsMetadataSet(nctx) {
+		t.Errorf("Context does not provide Metadata information")
+	}
+
 	for _, expected := range expectedMetadata {
 		for label, expVal := range expected {
+			if !IsLabel(label) {
+				t.Errorf("Expected label %s is not considered a valid label", label)
+			}
 			val := ValueFunc(nctx, label)
 			if val() != expVal() {
 				t.Errorf("Expected value %s for %s, but got %s", expVal(), label, val())
 			}
+		}
+	}
+}
+
+func TestLabelFormat(t *testing.T) {
+	labels := []struct {
+		label   string
+		isValid bool
+	}{
+		{"plugin/LABEL", true},
+		{"LABEL", false},
+		{"plugin.LABEL", false},
+		{"/NO-PLUGIN-ACCEPTED", true},
+		{"ONLY-PLUGIN-ACCEPTED/", true},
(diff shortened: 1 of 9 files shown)


plugin/rewrite/README.md
```diff
@@ -209,12 +209,21 @@ rewrites the first local option with code 0xffee, setting the data to "abcd". Eq
...
+the variable data will be filled with the value associated with that label. If that label is not provided, 
+the variable will be silently substitute by an empty string.  
+
+Examples:
 
 ~~~
 rewrite edns0 local set 0xffee {client_ip}
 ~~~
 
+~~~
```
miekg: The 2nd example needs more docs, something, ala:

The following example uses the imaginary *mac* plugin to retrieve the mac address, it adds a label `mac/address`, this can be etc. etc.

plugin/rewrite/README.md
```diff
@@ -209,12 +209,21 @@ rewrites the first local option with code 0xffee, setting the data to "abcd". Eq
 * A variable data is specified with a pair of curly brackets `{}`. Following are the supported variables:
   {qname}, {qtype}, {client_ip}, {client_port}, {protocol}, {server_ip}, {server_port}.
 
-Example:
+* If metadata plugin is enabled, then labels are supported as variables if they are presented within curly brackets.
```
miekg: If *the*. And this needs to reference the label syntax (maybe that should be a section in the metadata plugin)

plugin/rewrite/edns0.go
```diff
@@ -199,12 +202,22 @@ func (rule *edns0VariableRule) ruleData(state request.Request) ([]byte, error) {
 		return []byte(state.Proto()), nil
 	}
 
+	if metadata.Labels(ctx) != nil {
```
miekg: don't think this check is needed

miekg: This also needs to check this is a label, meaning *metadata* needs to export a `IsLabel` functions that checks the syntax (2 words with a slash as seperator)

plugin/rewrite/edns0.go
```diff
@@ -249,6 +262,11 @@ func isValidVariable(variable string) bool {
 		serverPort:
 		return true
 	}
+	// if variable contains a /, then we can assume it is a metadata.
```
miekg: See previous comment.

plugin/rewrite/rewrite_test.go
```diff
@@ -434,12 +438,32 @@ func optsEqual(a, b []dns.EDNS0) bool {
 	return true
 }
 
+type StaticProvider map[string]metadata.Func
```
miekg: doesn't need to be exported.

why `static`?

plugin/rewrite/rewrite_test.go
```diff
@@ -6,6 +6,10 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/coredns/coredns/request"
+
+	"github.com/coredns/coredns/plugin/metadata"
```
miekg: fmt

miekg (commented):
Yes, this is direction we should move to

plugin/rewrite/edns0.go
fturib: Yes the check is needed to make the difference between 
- "metadata" is enabled : an unknown {label} is resolved as "".
- "metadata" is not enabled : {label} cannot be resolved and we return an error (current behavior)

=> I can add a IsMetadataEnabled() along with an IsLabel()

plugin/rewrite/rewrite_test.go
fturib: I reuse it in the rewrite_test
Static because the value provided is static. In fact it is better maybe to say Direct or Immediate or Constant

plugin/rewrite/README.md
fturib: ok.

plugin/rewrite/README.md
fturib: ok

plugin/rewrite/edns0.go
fturib: ok

fturib:
Thanks for review, will update later in the day.

plugin/rewrite/rewrite_test.go
fturib: hum.. you are right. Finally I was not able to reused it. I will revert to initial.

miekg:
>
>
>
>> @@ -199,12 +202,22 @@ func (rule *edns0VariableRule) ruleData(state request.Request) ([]byte, error) {
> 		return []byte(state.Proto()), nil
> 	}
>
>+	if metadata.Labels(ctx) != nil {
>
>Yes the check is needed to make the difference between
>- "metadata" is enabled : an unknown {label} is resolved as "".
>- "metadata" is not enabled : {label} cannot be resolved and we return an error (current behavior)

Why would this need to tri-state? "" := no-metadata?

>=> I can add a IsMetadataEnabled() along with an IsLabel()

Yes, IsLabel will be needed.

fturib:
@miekg : as metadata is label=value that depends of the query itself (not the plugin configuration), the lack of label in a metadata should be handled silently with value "" (IMO).

On the other hand, today, or when the "metadata" plugin is not enable, 
1- the labels are not allowed
2- we fail if a variable is unknown.

So we need:
1 - to know if metadata is enabled (to error at runtime if not a variable - could be at setup, but I think we do not know yet if metadata is enabled or not)
2- isLabel (to error at setup if it is not a variable neither a label).

miekg:
>@miekg : as metadata is label=value that depends of the query itself (not the plugin configuration), the lack of label in a metadata should be handled silently with value "" (IMO).
>
>On the other hand, today, or when the "metadata" plugin is not enable,
>1- the labels are not allowed
>2- we fail if a variable is unknown.
>
>So we need:
>1 - to know if metadata is enabled (to error at runtime if not a variable - could be at setup, but I think we do not know yet if metadata is enabled or not)
>2- isLabel (to error at setup if it is not a variable neither a label).

Disagree. We return a string (at least now). So no metadata set just returns 
empty. Could be a tristate by making it a `*string`, but I'm remain unconvinced.

fturib:
Hum ... I think we are not pointing the same piece of code, or there is a misunderstanding somehow ..
I am pushing a commit taking care of these comments.
let me know what you think. Thanks!

plugin/metadata/provider.go
```diff
@@ -48,6 +49,18 @@ type Provider interface {
 // Func is the type of function in the metadata, when called they return the value of the label.
 type Func func() string
 
+// IsLabel check that the provided name looks like a valid label name
+func IsLabel(label string) bool {
+	parts := strings.Split(label, "/")
+	return len(parts) == 2 && len(parts[0]) > 0 && len(parts[1]) > 0
+}
+
+// IsMetadataSet check that the ctx is initialized to receive metadata information
```
miekg: This is not needed

plugin/metadata/provider.go
```diff
@@ -48,6 +49,18 @@ type Provider interface {
 // Func is the type of function in the metadata, when called they return the value of the label.
 type Func func() string
 
+// IsLabel check that the provided name looks like a valid label name
+func IsLabel(label string) bool {
+	parts := strings.Split(label, "/")
```
miekg: This allocates quite a bit of memory. `strings.Index` is a better choice I think.

also needs a test

plugin/metadata/provider.go
fturib: you mean : verifying that a Label has only 2 parts and each is non empty ?

miekg: for the islabel part, yes.

plugin/metadata/provider.go
```diff
@@ -48,6 +49,17 @@ type Provider interface {
 // Func is the type of function in the metadata, when called they return the value of the label.
 type Func func() string
 
+// IsLabel check that the provided name looks like a valid label name
+func IsLabel(label string) bool {
+	return strings.Index(label, "/") >= 0
+}
+
+// IsMetadataSet check that the ctx is initialized to receive metadata information
+func IsMetadataSet(ctx context.Context) bool {
```
miekg: please remove this is not needed

plugin/metadata/metadata_test.go
```diff
@@ -47,15 +47,45 @@ func TestMetadataServeDNS(t *testing.T) {
...
 		}
 	}
 }
+
+func TestLabelFormat(t *testing.T) {
+	labels := []struct {
+		label   string
+		isValid bool
+	}{
+		{"plugin/LABEL", true},
```
miekg, line 71: `/` would default to true as well and is invalid. The IsLabel function can fairly easy be extended to count the character to the index of / and check if we see another slash in the next part of the string.

plugin/metadata/metadata_test.go
```diff
@@ -47,15 +47,45 @@ func TestMetadataServeDNS(t *testing.T) {
...
+		{"LABEL", false},
+		{"plugin.LABEL", false},
+		{"/NO-PLUGIN-ACCEPTED", true},
```
miekg: this should be false

plugin/metadata/metadata_test.go
```diff
@@ -47,15 +47,45 @@ func TestMetadataServeDNS(t *testing.T) {
...
+		{"ONLY-PLUGIN-ACCEPTED/", true},
```
miekg: this should be false

plugin/metadata/metadata_test.go
```diff
@@ -47,15 +47,45 @@ func TestMetadataServeDNS(t *testing.T) {
...
+		{"PLUGIN/LABEL/SUB-LABEL", true},
```
miekg: this should be false

miekg (changes requested):
Getting there.

The IsMetaData is superfluous and should be removed. The IsLabel can be easily extended to count the number of chars seen and should reject invalid labels

plugin/metadata/provider.go
fturib: Did you look at code of Rewrite. If I remove IsMetadataSet .. then I can ensure the compatibility with current code.
=> any variables that looks like a LABEL will be accepted silently with a rewrite with empty string, instead of raising an error.

plugin/metadata/metadata_test.go
fturib: I agree with all FALSE cases.

miekg:
>Did you look at code of Rewrite. If I remove IsMetadataSet .. then I can ensure the compatibility with current code.
>=> any variables that looks like a LABEL will be accepted silently with a rewrite with empty string, instead of raising an error.

you don't need it if the rewrite's variable is a label, i.e. if IsLabel returns 
true you call the function with the labels name and you're done.

/Miek

--
Miek Gieben

fturib:
Hum.. that cannot assure compatibility with former behavior in case if metadata is not in the stanza.
But finally that is maybe not such a big deal. The former behavior was just abandoning the rewrite chain without providing any warning.

So, I will go by just considering a Label view as a variable is valid, and if the Label is not provided .. then the rewrite will enforce en empty string.

miekg:
>So, I will go by just considering a Label view as a variable is valid, and if the Label is not provided .. then the rewrite will enforce en empty string.

Or not do anything? I'm mean you call the function and it returns "", I would 
just consider that a noop.

/Miek

--
Miek Gieben

miekg (approved):
Thanks!

Issue #1931, opened 2018-07-03
zones expire on secondary


Isolus:
I'm using CoreDNS as a secondary DNS server (with plugin/secondary).
If I just wait without touching a zone file or coredns, the zone expires on the secondary after the time given in the SOA record.

Looking at `plugin/file/secondary.go` this behaviour seems to be a result of the `z.shouldTransfer()` checks. If the check returns "no", nothing happens (but we need to reset the tickers).

miekg:
yep, the expire needs to be restarted on every refresh (regardless if we transfer).

Do you have a PR?

miekg:
fixed via #1933

PR #1933, opened 2018-07-03
plugin/file: fix zone expiration

Isolus:
Zones expire without any changes on the secondary dns.

#1931 

-


miekg (approved):

miekg:
ideally we should add a test for this as well

PR #1936, opened 2018-07-04
Include Gravwell external plugin

traetox:
This pull request adds an external plugin that enables a direct integration with Gravwell for DNS auditing, analytics, etc.  Gravwell is available for free to the general public in the form of a community edition.  The plugin enables a very clean integration with CoreDNS so that users can get a highly reliable stream of DNS data.

Plugin code is located at https://github.com/gravwell/coredns
Access to Gravwell is provided at https://dev.gravwell.io/docs/#!quickstart/community-edition.md
Free Community licenses are available at https://www.gravwell.io/activate-community-edition
Additional info about Gravwell can be found at https://www.gravwell.io

N/A

None, just includes the reference to the plugin.


miekg:
Thanks for this plugin! We don't usually include external plugins like this
per default.

If your happy to be included on coredns.io/explugins I can list this plugin
there.

traetox:
That works for me.  Is there a standard way to include the plug-in via
build args?

I saw the way the route53 build used some shell foo but wasn't sure if that
was the sanctioned method of inclusion.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Include Gravw..." ]
>That works for me.  Is there a standard way to include the plug-in via
>build args?

No you have to edit plugin.cfg. Build args would be nice option though, but you 
need to then figure out where in the chain you want to insert the plugin.

>I saw the way the route53 build used some shell foo but wasn't sure if that
>was the sanctioned method of inclusion.

Can't recall that, but probably could have been some sed command to insert the 
plugin into plugin.cfg and then go generate.

traetox:
Ok.  Thanks for the info and awesome work.  Big fan of a statically
compiled and pluggable DNS server.

miekg:
See https://github.com/coredns/coredns.io/pull/96

can you flesh that out and describe the various options and give some examples? Thanks!

yongtang:
You can build plugin either with plugin.cfg, or build with a small entry point main program. They are described in https://coredns.io/2017/07/25/compile-time-enabling-or-disabling-plugins/

+1 to give some examples.

miekg:
closing here as it will be an external plugin.

PR #1941, opened 2018-07-05
plugin/forward: transport type optimization

rdrozhdzh:
code optimization 
 - replace strings with integers for protocol (transport type)
 - replace map with array in transport object. So, access to cached
   connections should be faster and less CPU consuming
 - moved protocol (transport type) selection code into single place


miekg:
Is this really *that* much faster. I find a bit less readable, because of the constants being used.

It is possible to add some kind of micro bench mark to see what we win here? I'm suspecting it will be dwarfed by the network latency anyway.

jiachengxu:
This optimization is really simple but nice!!

rdrozhdzh:
taking a vacation, will work on benchmark later

PR #1947, opened 2018-07-06
plugin/loadbalance: add parse and tests

miekg:
Add a parse function and test this in a setup_test.go. Also brush up the
documentation a little and mention we shuffly MX as well.


yongtang (approved):
LGTM

PR #1766, opened 2018-05-03
plugin/forward: expose few methods and attributes to public

eugen-kleiner-mesh:
### 1. What does this pull request do?
 * Makes Proxy.Connect() and Forward.List() methods public
 * Adds getters for Proxy.addr, Forward.from and Forward.forceTCP

### 2. Which issues (if any) are related?
It implements https://github.com/coredns/coredns/issues/1762

PR #1949, opened 2018-07-07
plugin/forward: various cleanup

miekg:
Fix documentation and remove the unused From method.

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -63,8 +63,7 @@ forward FROM TO... {
 * `force_tcp`, use TCP even when the request comes in over UDP.
 * `prefer_udp`, try first using UDP even when the request comes in over TCP. If response is truncated
   (TC flag set in response) then do another attempt over TCP. In case if both `force_tcp` and
-  `prefer_udp`.
-  options specified the `force_tcp` takes precedence.
+  `prefer_udp` options specified the `force_tcp` takes precedence.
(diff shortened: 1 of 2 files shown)


rdrozhdzh:
@miekg, the method Forward.From() was requested in #1762, and added in #1766. It may be still needed.
@fturib, @ekleiner, could you confirm?

miekg:
Ah yes, damn, I remember.  Something, something that needed this. If there are no in-tree users of exported functions this is not sustainable (and there is no backwards compatibility guarantee)

fturib:
@rdrozhdzh : thank you for reminding these requests.
After check in our ATEPMonitor plugin, I think we are not using it anymore.

@miekg : what do you mean by "in-tree users of exported functions" ? I mean, how should we code the functions we need to be public for external plugins such a way they will not be deleted in refactoring ?

miekg:
>@rdrozhdzh : thank you for reminding these requests.
>After check in our ATEPMonitor plugin, I think we are not using it anymore.
>
>@miekg : what do you mean by "in-tree users of exported functions" ? I mean, how should we code the functions we need to be public for external plugins such a way they will not be deleted in refactoring ?
>

We can't be expected to maintain some sort of backwards compatibility for 
(probably, accidentally) export functions *in* plugins.

And now you tell me you actually don't need (nor use) those functions anyway, 
which just proofs my point.

/Miek

--
Miek Gieben

PR #1950, opened 2018-07-07
plugin/forward: add HealthChecker interface

miekg:
Make the HealthChecker interface and morph the current DNS health
checker into that interface.

Remove all whole bunch of method on Forward that didn't make sense.

This is done in preparation of adding a DoH client to forward - which
requires a completely different healthcheck implementation (and more,
but lets start here)

### DIFF
--- a/plugin/forward/health.go
+++ b/plugin/forward/health.go
@@ -1,17 +1,47 @@
 package forward
 
 import (
+	"crypto/tls"
 	"sync/atomic"
+	"time"
 
 	"github.com/miekg/dns"
 )
+
+// HealthChecker checks the upstream health.
+type HealthChecker interface {
+	Check(*Proxy) error
+	SetTLSConfig(*tls.Config)
+}
+
+type hc struct{ c *dns.Client }
+
+// NewHealthChecker returns a new HealthChecker based on protocol.
+func NewHealthChecker(protocol int) HealthChecker {
+	switch protocol {
+	case DNS, TLS:
+		c := new(dns.Client)
+		c.Net = "udp"
+		c.ReadTimeout = 1 * time.Second
+		c.WriteTimeout = 1 * time.Second
+
+		return &hc{c: c}
+	}
+
+	return nil
+}
+
+func (h *hc) SetTLSConfig(cfg *tls.Config) {
+	h.c.Net = "tcp-tls"
+	h.c.TLSConfig = cfg
+}
 
 // For HC we send to . IN NS +norec message to the upstream. Dial timeouts and empty
 // replies are considered fails, basically anything else constitutes a healthy upstream.
 
 // Check is used as the up.Func in the up.Probe.
-func (p *Proxy) Check() error {
-	err := p.send()
+func (h *hc) Check(p *Proxy) error {
+	err := h.send(p.addr)
 	if err != nil {
 		HealthcheckFailureCount.WithLabelValues(p.addr).Add(1)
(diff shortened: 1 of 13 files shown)


miekg:
cc @rdrozhdzh

plugin/forward/health.go
```diff
@@ -1,17 +1,47 @@
...
 	"github.com/miekg/dns"
 )
 
+// HealthChecker checks the upstream health.
+type HealthChecker interface {
+	Check(*Proxy) error
+	SetTLSConfig(*tls.Config)
+}
+
+type hc struct{ c *dns.Client }
```
rdrozhdzh: consider renaming hc to dnsHc, since we're going to have other kinds of hc in future. It would be nice that name to correspond the hc type

plugin/forward/proxy.go
```diff
@@ -7,78 +7,56 @@ import (
...
 
 	// health checking
-	probe *up.Probe
+	probe  *up.Probe
+	health HealthChecker
 }
 
 // NewProxy returns a new proxy.
-func NewProxy(addr string, tlsConfig *tls.Config) *Proxy {
+func NewProxy(addr string, protocol int, cfg *tls.Config) *Proxy {
```
rdrozhdzh: consider removing the `cfg` parameter - we never pass tls config to `NewProxy()`. The same for second parameter of `newTransport()`. The tls config is actually set with `SetTlsConfig()` method.

plugin/forward/proxy.go
```diff
@@ -7,78 +7,56 @@ import (
...
+	p.health = NewHealthChecker(protocol)
+	if cfg != nil {
+		p.health.SetTLSConfig(cfg)
+	}
 	runtime.SetFinalizer(p, (*Proxy).finalizer)
 	return p
 }
 
-// Addr returns the address to forward to.
-func (p *Proxy) Addr() (addr string) { return p.addr }
```
rdrozhdzh, line 45: this may be needed by external plugins, see #1762, #1766

plugin/forward/proxy.go
```diff
@@ -7,78 +7,56 @@ import (
...
-
 // SetTLSConfig sets the TLS config in the lower p.transport and in the healthchecking client.
 func (p *Proxy) SetTLSConfig(cfg *tls.Config) {
 	p.transport.SetTLSConfig(cfg)
-	p.client = dnsClient(cfg)
+	p.health.SetTLSConfig(cfg)
 }
 
-// IsTLS returns true if proxy uses tls.
-func (p *Proxy) IsTLS() bool { return p.transport.tlsConfig != nil }
```
rdrozhdzh, line 69: this may be needed by external plugins, see #1762, #1781

miekg:
All comments addressed.

Thanks for flagging those public funtions, but if there is no intree usage 
nothing breaks during compilation, nor was there a test (with comments), saying 
this should stay.

rdrozhdzh:
Looks good.
@fturib, @ekleiner, please add appropriate test for the public methods/attributes you need

fturib:
@rdrozhdzh : Thanks to catch that point ! 
At first sight, looks like we are not using it anymore, as we changed the way to monitor the upstream connection. I will let @ekleiner verify and add these tests if needed.


@miekg : that raise concern on how to keep the compatibility with external plugins.
Is adding a UT on needed public functions for external plugins the way to ensure they are not removed in later releases ?

PR #1951, opened 2018-07-08
plugin/metadata: finish documentation

miekg:
Finish the README.md, add corner case in the IsLabel test and reword
some code comments slightly.

Generate the man-pages and add man/coredns-metadata.7 as well.

### DIFF
--- a/man/coredns-metadata.7
+++ b/man/coredns-metadata.7
@@ -1,0 +1,44 @@
+.\" generated with Ronn/v0.7.3
+.\" http://github.com/rtomayko/ronn/tree/0.7.3
+.
+.TH "COREDNS\-METADATA" "7" "July 2018" "CoreDNS" "CoreDNS plugins"
+.
+.SH "NAME"
+\fImetadata\fR \- enable a meta data collector\.
+.
+.SH "DESCRIPTION"
+By enabling \fImetadata\fR any plugin that implements metadata\.Provider interface \fIhttps://godoc\.org/github\.com/coredns/coredns/plugin/metadata#Provider\fR will be called for each DNS query, at beginning of the process for that query, in order to add it\'s own meta data to context\.
+.
+.P
+The meta data collected will be available for all plugins, via the Context parameter provided in the ServeDNS function\. The package (code) documentation has examples on how to inspect and retrieve metadata a plugin might be interested in\.
+.
+.P
+The meta data is added by setting a label with a value in the context\. These labels should be named \fBplugin/NAME\fR, where \fBNAME\fR is something descriptive\. The only hard requirement the \fImetadata\fR plugin enforces is that the labels contains a slash\. See the documentation for \fBmetadata\.SetValueFunc\fR\.
+.
+.P
+The value stored is a string\. The empty string signals "no meta data"\. See the documentation for \fBmetadata\.ValueFunc\fR on how to retrieve this\.
+.
+.SH "SYNTAX"
+.
+.nf
+
+metadata [ZONES\.\.\. ]
+.
+.fi
+.
+.IP "\(bu" 4
+\fBZONES\fR zones metadata should be invoked for\.
+.
+.IP "" 0
+.
+.SH "PLUGINS"
+\fBmetadata\.Provider\fR interface needs to be implemented by each plugin willing to provide metadata information for other plugins\. It will be called by metadata and gather the information from all plugins in context\.
+.
+.P
+Note: this method should work quickly, because it is called for every request\.
+.
+.SH "EXAMPLES"
+The \fIrewrite\fR plugin uses meta data to rewrite requests\.
+.
+.SH "ALSO SEE"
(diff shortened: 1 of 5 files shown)


man/coredns-metadata.7
```diff
@@ -0,0 +1,44 @@
...
+By enabling \fImetadata\fR any plugin that implements metadata\.Provider interface \fIhttps://godoc\.org/github\.com/coredns/coredns/plugin/metadata#Provider\fR will be called for each DNS query, at beginning of the process for that query, in order to add it\'s own meta data to context\.
+.
+.P
+The meta data collected will be available for all plugins, via the Context parameter provided in the ServeDNS function\. The package (code) documentation has examples on how to inspect and retrieve metadata a plugin might be interested in\.
+.
+.P
+The meta data is added by setting a label with a value in the context\. These labels should be named \fBplugin/NAME\fR, where \fBNAME\fR is something descriptive\. The only hard requirement the \fImetadata\fR plugin enforces is that the labels contains a slash\. See the documentation for \fBmetadata\.SetValueFunc\fR\.
+.
+.P
+The value stored is a string\. The empty string signals "no meta data"\. See the documentation for \fBmetadata\.ValueFunc\fR on how to retrieve this\.
```
fturib, line 19: this info: `empty string signals "no meta data"`
we have 3 cases:
- Provider advertise a Label and when retrieving the value, that value is not empty (expected regular case)
- Provider advertise a Label and when retrieving the value, that value is empty
- Provider do not advertise a Label 

Which one do we call "no meta data", view from the consumer ?

Consumer knows if Label is advertised with ValueFunc not NIL
if consumer can retrieve a value for a Label with ValueFunc, then there is a meta data .. and maybe that value is empty.

IMO,  the "no meta data" case is when the Label is not advertised (ValueFunc return a nil)

man/coredns-rewrite.7
```diff
@@ -306,12 +306,16 @@ rewrites the first local option with code 0xffee, setting the data to "abcd"\. E
...
 .IP "" 0
 .
-.TP
-A variable data is specified with a pair of curly brackets \fB{}\fR\. Following are the supported variables
-{qname}, {qtype}, {client_ip}, {client_port}, {protocol}, {server_ip}, {server_port}\.
+.IP "\(bu" 4
+A variable data is specified with a pair of curly brackets \fB{}\fR\. Following are the supported variables: {qname}, {qtype}, {client_ip}, {client_port}, {protocol}, {server_ip}, {server_port}\.
+.
+.IP "\(bu" 4
+If the metadata plugin is enabled, then labels are supported as variables if they are presented within curly brackets\. the variable data will be filled with the value associated with that label\. If that label is not provided, the variable will be silently substitute by an empty string\.
```
fturib, line 313: If that label is not provided, the rewrite rule will be ignored\.

plugin/metadata/README.md
```diff
@@ -8,15 +8,20 @@
...
+metadata a plugin might be interested in.
 
-TODO: write about naming of the keys (labels).
-TODO: write about extracting and using
+The meta data is added by setting a label with a value in the context. These labels should be named
+`plugin/NAME`, where **NAME** is something descriptive. The only hard requirement the *metadata*
+plugin enforces is that the labels contains a slash. See the documentation for
+`metadata.SetValueFunc`. 
+
+The value stored is a string. The empty string signals "no meta data". See the documentation for
```
fturib, line 23: same note as above. I would just not mention "The empty string signals "no meta data". ".

plugin/metadata/metadata_test.go
```diff
@@ -68,20 +68,23 @@ func TestLabelFormat(t *testing.T) {
...
 		{"/", false},
+		{"//", false},
 	}
 
 	for _, test := range labels {
-		if IsLabel(test.label) != test.isValid {
-			t.Errorf("Label %v is expected to have this validaty : %v - and has the opposite", test.label, test.isValid)
+		if x := IsLabel(test.label); x != test.isValid {
+			t.Errorf("Label %v expected %v, got: %v", test.label, test.isValid, x)
 		}
```
fturib, line 88: thanks for cleanup.

fturib (commented):
I tried to explain what I think about the "empty" and "no metadata".

NOTE: for Rewrite, current coding will do:

- if Label is not advertised : just keep the rewrite (no edns0)
-if the Label is advertised and ValueFunc return empty (""), then the rewrite will happen and an edns0 entry will be created for the provide key and the empty "" value.

miekg:
>NOTE: for Rewrite, current coding will do:
>
>- if Label is not advertised : just keep the rewrite (no edns0)
>-if the Label is advertised and ValueFunc return empty (""), then the rewrite will happen and an edns0 entry will be created for the provide key and the empty "" value.

I don't think that should be the contract. If you can't set metadata, i.e. your 
value is empty, you don't set anything. I can't think of a scenario where this 
would not be the case.

fturib:
Hum.. I am not really convince it is better to merge Unknown and Empty. We just remove fine grain in the metadata information.
But I will not argue for long time.
So let's go like that.

miekg:
>So let's go like that.

/merge

PR #1959, opened 2018-07-10
Fix ineffassign

yongtang:

This fix tries to fix ineffassign, as was reported in:
https://goreportcard.com/report/github.com/coredns/coredns#ineffassign

N/A

N/A

### DIFF
--- a/plugin/federation/setup.go
+++ b/plugin/federation/setup.go
@@ -50,7 +50,7 @@ func federationParse(c *caddy.Controller) (*Federation, error) {
 	for c.Next() {
 		// federation [zones..]
 		zones := c.RemainingArgs()
-		origins := []string{}
+		var origins []string
 		if len(zones) > 0 {
 			origins = make([]string, len(zones))
 			copy(origins, zones)
--- a/plugin/file/setup.go
+++ b/plugin/file/setup.go
@@ -53,7 +53,6 @@ func setup(c *caddy.Controller) error {
 func fileParse(c *caddy.Controller) (Zones, error) {
 	z := make(map[string]*Zone)
 	names := []string{}
-	origins := []string{}
 
 	config := dnsserver.GetConfig(c)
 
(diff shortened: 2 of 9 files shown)


plugin/log/setup.go
```diff
@@ -62,7 +62,7 @@ func logParse(c *caddy.Controller) ([]Rule, error) {
 			})
 		} else {
 			// Name scope, and maybe a format specified
-			format := DefaultLogFormat
+			var format string
```
miekg: are these things really an improvement?

yongtang: @miekg I guess it is something "nice to have" instead of "true" improvement.

Though the reason `format := DefaultLogFormat` shown up as a warning is that down in Ln 72:
```
 			default:
 				format = args[1]
```
That means `DefaultLogFormat` is actually not in use.

miekg:
I agree it's not used, but it does convey it will be a format of some sort.
Applying this will remove that indication.
My feeling is this or is technical correct, but ignores the documentation
these assignments provide .

yuewko:
How about renaming `format` to `logFormat`?

plugin/federation/setup.go
```diff
@@ -50,7 +50,7 @@ func federationParse(c *caddy.Controller) (*Federation, error) {
 	for c.Next() {
 		// federation [zones..]
 		zones := c.RemainingArgs()
-		origins := []string{}
+		var origins []string
```
miekg, line 53: why is one better over the other?

plugin/file/setup.go
```diff
@@ -53,7 +53,6 @@ func setup(c *caddy.Controller) error {
 func fileParse(c *caddy.Controller) (Zones, error) {
 	z := make(map[string]*Zone)
 	names := []string{}
-	origins := []string{}
```
miekg, line 56: this seems to make sense

plugin/log/setup.go
miekg: This hurts readability and we don't care about performance in this piece of code. Can you remove this? The rest of this PR lgtm.

PR #1962, opened 2018-07-11
Fix and replace message truncation

grobie:
With the addition of the Truncate() function in miekg/dns, the Scrub
implementation can be replaced. This fixes the following issues:

- Scrub would return invalid messages larger than the advertised client
  buffer size by adding an OPT record to responses without making room
  for it. This breaks at least the Java DNS client.
- Scrub would not keep the OPT record in truncated messages, even though
  it's mandated by rfc6891.
- As Scrub doesn't have access to the internal DNS len() functions, it
  had to loop over all resource record lists multiple times in order to
  find the maximum number of records fitting the advertised size. This
  can be implemented more efficiently in miekg/dns.

---

Depends on https://github.com/miekg/dns/pull/708.

As a follow up change, I'd suggest to move SizeAndDo and Scrub from each plugin into a ResponseWriter injected in the Server. Handling the requirements of the DNS spec correctly shouldn't be a concern of plugins but of the server. That will reduce a lot copy&pasted code and remove the possibility that CoreDNS can return invalid messages when plugin authors forget to call these functions.

### DIFF
--- a/request/request.go
+++ b/request/request.go
@@ -248,6 +248,7 @@ const (
 // get the bit, the client should then retry with pigeons.
 func (r *Request) Scrub(reply *dns.Msg) (*dns.Msg, Result) {
 	size := r.Size()
+	extras := len(reply.Extra)
 
 	reply.Compress = false
 	rl := reply.Len()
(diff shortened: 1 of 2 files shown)


miekg:
> As a follow up change, I'd suggest to move SizeAndDo and Scrub from each plugin 

excellent idea!

miekg:
I suggest splitting this is up in 2 parts; one for the  OPT rr addition when purge the answer section and another one for potential optimizations.

miekg:
@grobie (maybe you're on holiday), but if this needs to go in (soonish), you'll need to split it and preferably leave out the dep on the change in miekg/dns for the bugfix.

Thanks

grobie:
Sorry, I don't have much computer time at the moment. I'll only be able to get back to this in ~3 weeks.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fix and repla..." ]
>Sorry, I don't have much computer time at the moment. I'll only be able to get back to this in ~3 weeks.

np, enjoy your time AFK

PR #1963, opened 2018-07-11
WIP - Ensure Graceful shutdown of DNS Servers

fturib:
We found that when "restarting" CoreDNS there were some query lost.
In our usage we try to minimize this lost.

I used prepared infrastructure to improve a little the shutdown process of the Server (TCP/UDP)
- During the closing process, i used DecoratorReader to stop accepting new queries.
- There is already a mechanism to wait completion before closing the service. 
=> I just added handle of the queries as part of the completion

NOTES: 
- the test performed so far show we stopped to loose queries when restarting.
- further tests are still in progress : 
-- same test but with higher QPS 
-- verification of impact of performance (we are using sync to count queries)

I am not sure how to proceed for UT here ...

#1666 

I am not sure here as it is an improvement of the graceful restart of CoreDNS.

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -33,6 +33,7 @@ type Server struct {
 	server    [2]*dns.Server // 0 is a net.Listener, 1 is a net.PacketConn (a *UDPConn) in our case.
 	udpReader *OnOffReader   // Reader for UDP protocol that can stop accepting in queries when asked
 	tcpReader *OnOffReader
+	readerOff uint32     // inform that reader should start off. Corner case where the Readers are created after we start shutdown the Server
 	m         sync.Mutex // protects the servers
 
 	zones       map[string]*Config // zones keyed by their address
@@ -85,6 +86,7 @@ func NewServer(addr string, group []*Config) (*Server, error) {
 		Addr:        addr,
 		zones:       make(map[string]*Config),
 		connTimeout: 5 * time.Second, // TODO(miek): was configurable
+		readerOff:   0,               // by default the Server accept queries
 	}
 
 	// We have to bound our wg with one increment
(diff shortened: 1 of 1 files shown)


miekg:
If you are in a controlled shutdown I don't see why you can't just use lameduck in the health plugin.
Also think this works should be done in the underlaying server in miekg/dns, but without any `defers` that are still slowish.

chrisohaver:
What happened to the WIP and CI checks?

fturib:
@chrisohaver : I guess my bad here. I made a last change changing a struct to a pointer of struct. and i did not run the test locally. working on it.

chrisohaver:
No, I mean - the WIP bot and coredns/ci bots?  they are missing from the checks list.

fturib:
@miekg : for health, lameduck is triggered "onRestart" that means before we try to stop the server. I am not sure what we can do here, unless we are sure to stop accepting input before the lameduck is triggered. - I do not see an event happening before onRestart..

@miekg : about implementing in Miek/dns - do you mean ensure gracefull stop in the underlying TCP Servers and UDP Servers ? (counting the queries, switch off input and wait all queries are processed)

chrisohaver:
OK - there are all our friendly bots ... i guess github flaked out earlier.

miekg:
@fturib yes and it's tricky to get right, we had an impl. which didn't really work, then someone tried to tackle that, but that led to code so complex that it had to to reverted.
If you have the cycles are sane and simple (and fast) impl. of graceful shutdown in miekg/dns would be appreciated.

fturib:
@miekg : I will look into. Mostly I will reproduce mechanism similar to the one implemented here. But on the level of Miek/dns servers.

fturib:
This PR is now proposed on the DNS level : see https://github.com/miekg/dns/pull/713

bjhaid:
Hi, when should we expect a coredns release containing https://github.com/miekg/dns/pull/717? Thanks!

Issue #1848, opened 2018-06-01
Use 'docker manifest' instead of 'manifest-tool' in Makefile.release


chrisohaver:
Manifest creation has recently been added to docker cli docker/cli#138.
So at some point we should move to the built-in docker manifest creation instead of using the (soon to be deprecated) manifest-tool.

chrisohaver:
FWIW, I just noticed that "docker manifest" is still labeled as "experimental" in docker's docs.
https://docs.docker.com/edge/engine/reference/commandline/manifest/

miekg:
still experimental

miekg:
current solution seems to work fine.

PR #1968, opened 2018-07-12
Plugin/metrics - Starting/stopping tuning for the listener and HTTPServer

fturib:
Initial problem is:
1- plugin metrics fail to follow metric directive after 2 restarts
2- plugin metrics fail to start listener (sometime ...)

Point 1 - is fixed by modifying the interaction with uniqAddr = Unset the usage of Addr when shutdown the listener. It ensures the new metrics setup will initialize - if needed - a new uniqAddr *with its own OnStart function*

Point 2, after investigation, seems due to latency in releasing the socket on some OS platform. Works great on Mac OS. But our install on Alpine needs a delay before being able to reuse the listener.
=> add a second retry for acquiring the listener after 1sec delay, if first is failing.

NOTE: In order to show the failing point 1, I created a test in "test_reload". However I found that the service metrics was still running after listener is down (???), unless we stop the HTTP Server with Shutdown instead of Close.
I added this change in the code to have a proper closing.

#1848 (for point 2)
There is no issue created for point 1.

I do not think it needs. only fix or tuning.

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -8,40 +8,30 @@
 # For example: https://coredns.io/2016/09/18/coredns-001-release/ Also send an
 # email to coredns-discuss@ to announce the new version.
 #
-# We use https://github.com/progrium/gh-release to automate github stuff be sure
-# to have that binary in your path.
-#
-# Get a list of authors for this release with:
-#
-# git log --pretty=format:'%an' v001..master | sort -u (where v001 is the
-# previous release, obviously you'll need to adjust this)
+# Getting the authors for this release is done with the following command line
+# git log --pretty=format:'%an' v$(VERSION)..master | sort -u
 #
 # Steps:
-#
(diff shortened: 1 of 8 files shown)


fturib:
Hum.. something is wrong in my PR.
It shows the diff on files that are just coming from a merge of main.

fturib:
I am a bit confused with the commit. And a rebase does not seems to help.
I will rebuild the diff on my code on top of a branch created from master.

chrisohaver:
> But our install on Alpine 

Are you referring to the docker container here?  CoreDNS dropped alpine in 1.1.4.

>  will rebuild the diff on my code on top of a branch created from master.

Make sure your master isn't corrupted.

fturib:
CorreDNS does not provide image with alpine, but our usage in IB is still based on alpine OS

PR #1969, opened 2018-07-12
plugin/metrics - fails to follow metric directive after 2 restarts

fturib:
Initial problem is:
Plugin metrics fail to follow metric directive after 2 restarts

is fixed by modifying the interaction with uniqAddr = Unset the usage of Addr when shutdown the listener. It ensures the new metrics setup will initialize - if needed - a new uniqAddr *with its own OnStart function*

NOTE: In order to show the failing point, I created a test in "test_reload". 
However I found that the service metrics was still running after listener is down (???), unless we stop the HTTP Server with Shutdown instead of Close.
I added this change in the code to have a proper closing.

#1982

I do not think it needs. only fix or tuning.

### DIFF
--- a/plugin/metrics/metrics.go
+++ b/plugin/metrics/metrics.go
@@ -2,10 +2,12 @@
 package metrics
 
 import (
+	"context"
 	"net"
 	"net/http"
 	"os"
 	"sync"
+	"time"
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/metrics/vars"
@@ -22,6 +24,7 @@ type Metrics struct {
 	ln      net.Listener
 	lnSetup bool
 	mux     *http.ServeMux
+	srv     *http.Server
 
 	zoneNames []string
 	zoneMap   map[string]bool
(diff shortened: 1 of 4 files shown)


miekg:
Isn't Go 1.11 always setting socket re-use on sockets, preventing this problem. Doing a sleep(1) isn't really solving this problem.
Are we're not returning the socket correctly when closing done?

fturib:
I will split this PR in two : one for lost of control from setup (point 1) that is a found issue in the plugin.
and after double-checking the 2 points you are mentioning see how to go ahead with our issue about the listener. (point 2)

fturib:
based on comment of Miek into the PR, I searched in goland and closing listeners ...

There is this issue in golang: https://github.com/golang/go/issues/22502
which is a duplication of https://github.com/golang/go/issues/21856

Looks like there is a race condition : if you do an accept() in the time you are closing the listener, then the listener stay up for a while, until the system figure out the listener is closed ...

and the fix seems here : https://github.com/golang/go/commit/382d4928b8a758a91f06de9e6cb10b92bb882eff

and the code fix is in Go v1.10 and later, but NOT in Go v 1.9 (and minor versions)

plugin/pkg/uniq/uniq.go
```diff
@@ -24,24 +24,24 @@ func (u U) Set(key string, f func() error) {
...
 }
 
 // ForEach iterates for u executes f for each element that is 'todo' and sets it to 'done'.
 func (u U) ForEach() error {
 	for k, v := range u.u {
 		if v.state == todo {
-			v.f()
+			err := v.f()
+			if err == nil {
+				v.state = done
```
fturib, line 41: Let's consider the state goes to "done" only if the start of the Listener does not fail.

plugin/metrics/metrics.go
```diff
@@ -22,6 +23,7 @@ type Metrics struct {
 	ln      net.Listener
 	lnSetup bool
 	mux     *http.ServeMux
+	srv     *http.Server
```
fturib, line 27: Adding the srv, allow to call the Shutdown() and ensure the service is really closed when shutting down the plugin.
now expected by UT.

plugin/metrics/metrics.go
```diff
@@ -95,35 +97,46 @@ func (m *Metrics) OnStartup() error {
...
 	m.lnSetup = false
+	log.Infof("listener stopped on %s", m.Addr)
+}
+
+func (m *Metrics) OnRestart() error {
+	m.stopService()
+
+	// un-register so a new metrics from the new config can register and be launched
+	uniqAddr.Unset(m.Addr)
+
```
fturib, line 132: Here is the change to fix the initial issue : the current plugin is ended. So let's just unset from uniqAddr and let any new plugin setup to register for the next launching.
(and if there is no "prometheus" in the new corefile, then no listener will be launched)

fturib:
The active part of this PR is superseded by #1983.
This one become useless.

PR #1972, opened 2018-07-13
plugin/rewrite: prevent illegal names

miekg:
Log and returns an error when the name rewrite creates a name that is
illegal. Add test in name_test.go to see if an error is returned.

Possible followup could be the only check this if a name-rewrite is
done.

Fixes: #1638

### DIFF
--- a/plugin/rewrite/name.go
+++ b/plugin/rewrite/name.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/request"
+
+	"github.com/miekg/dns"
 )
 
 type nameRule struct {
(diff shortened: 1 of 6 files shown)


plugin/rewrite/name.go
```diff
@@ -194,3 +196,16 @@ func (rule *substringNameRule) GetResponseRule() ResponseRule { return ResponseR
...
+// validName returns true if s is valid domain name and shortern than 256 characters.
+func validName(s string) bool {
+	_, ok := dns.IsDomainName(s)
+	if !ok {
+		return false
+	}
+	if len(dns.Name(s).String()) > 255 {
+		return false
+	}
+
```
fturib, line 209: if you return an error with the exact pb, it will be more easy to find out the issue and fix it.
=> err("name is not a valid domain name")
=> err("name is too long (over 255 chars)")

fturib (commented):
LGTM. 
although advise on returning a more precise error to help use to understand/fix the issue

plugin/rewrite/name.go
greenpau: > if you return an error with the exact pb, it will be more easy to find out the issue and fix it.

@fturib , good point 👍

greenpau (commented):
LGTM 👍

miekg:
push another commit to address @fturib 's comment, but meh. Good enough, might revisit this later to optimize

PR #1973, opened 2018-07-13
plugin/foward: Simplify connection expiration

miekg:


miekg:
@rdrozhdzh I think this is valid simplification that just turns the thing in to a FIFO stack. Are you available to take a look? Thanks!

plugin/forward/persistent.go
```diff
@@ -66,85 +64,49 @@ Wait:
 				if time.Since(pc.used) < t.expire {
 					// Found one, remove from pool and return this conn.
 					t.conns[proto] = stack[:len(stack)-1]
-					t.ret <- pc.c
+					t.ret <- pc
 					continue Wait
 				}
-				// clear entire cache if the last conn is expired
-				t.conns[proto] = nil
```
rdrozhdzh, line 73: so, now the cache is not cleaned until the plugin is stopped? Besides of memory leak this will lead to exhausting open file descriptors (sockets)

plugin/forward/persistent.go
```diff
@@ -66,85 +64,49 @@ Wait:
...
-		case conn := <-t.yield:
+		case pc := <-t.yield:
 
 			SocketGauge.WithLabelValues(t.addr).Set(float64(t.len() + 1))
 
 			// no proto here, infer from config and conn
-			if _, ok := conn.Conn.(*net.UDPConn); ok {
-				t.conns["udp"] = append(t.conns["udp"], &persistConn{conn, time.Now()})
+			if _, ok := pc.c.Conn.(*net.UDPConn); ok {
+				t.conns["udp"] = append(t.conns["udp"], pc)
```
rdrozhdzh, line 81: the time is not updated neither on yield nor on dial (when popping from the cache), so each connection will expire exactly in `t.expire` time after creation. This changes the logic. Currently we close the connection if it was **idle** for `t.expire` time

plugin/forward/persistent.go
```diff
@@ -66,85 +64,49 @@ Wait:
...
 			if t.tlsConfig == nil {
-				t.conns["tcp"] = append(t.conns["tcp"], &persistConn{conn, time.Now()})
+				t.conns["tcp"] = append(t.conns["tcp"], pc)
 				continue Wait
 			}
 
-			t.conns["tcp-tls"] = append(t.conns["tcp-tls"], &persistConn{conn, time.Now()})
-
-		case <-ticker.C:
-			t.cleanup(false)
```
rdrozhdzh, line 100: We need cleaning the cache by timer because there may be periods when no new requests comes and nothing else can trigger cleanup. Also, the finalizer which calls `stop()` may be invoked in unpredictable long time or even not invoked at all. Doing cleanup by timer we guarantee that unused resources (sockets) are released in controlled manner.

rdrozhdzh (changes requested):
@miekg, what is your concern about the current implementation? Do you think it is working incorrectly because of races during cleanup?

miekg:
Before I make any changes, do you agree with the general gist of this PR,
or should we drop it?

rdrozhdzh:
I would prefer to drop it, I don't think we can simplify anything here without losing important functionality.
The only thing which could be changed is not to spawn new goroutines for closing connections - if this is your main concern. But I personally don't see any issues here

miekg:
Ok. Understood. I'll drop this.

PR #1978, opened 2018-07-16
WIP: plugin/secondary: make initial zone transfer async

### LINKED ISSUE #1609: plugin/secondary: Denial-of-Service via endless Zone Transfer

miekg:
Just run the whole thing in (another) go routine. This doesn't block
startup and is an easy fix for #1609. With the erratic axfr improvements
going on some more failure modes can be tested.

Fixes: #1609

### DIFF
--- a/plugin/secondary/setup.go
+++ b/plugin/secondary/setup.go
@@ -29,9 +29,11 @@ func setup(c *caddy.Controller) error {
 		if len(z.TransferFrom) > 0 {
 			c.OnStartup(func() error {
 				z.StartupOnce.Do(func() {
-					z.TransferIn()
 					go func() {
-						z.Update()
+						z.TransferIn()
+						go func() {
+							z.Update()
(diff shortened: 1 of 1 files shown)


chrisohaver:
Go question: Is nesting the routines inside one another different than calling them one at a time?  e.g.

~~~
go func() {
   z.TransferIn()
 }()
go func() {
   z.Update()
}()
~~~

chrisohaver:
Ah - I see, the order is controlled.  So, then how is it different than....
~~~
go func() {
   z.TransferIn()
   z.Update()
}()
~~~

miekg:
Hmmm, `z.Apex` is not protected by a Mutex which I discovered before (PR long time ago), so this needs more mutexes. We already have `reloadMu`, which should/could be reused.

miekg:
closing. This needs another attempt

PR #1979, opened 2018-07-16
Clean up tests logging

miekg:
This cleans up the travis logs so you can see the failures better.

Older tests in tests/ would call log.SetOutput(ioutil.Discard) in
a haphazard way. This add log.Discard and put an `init` function in each
package's dir (no way to do this globally). The cleanup in tests/ is
clear.

All plugins also got this init function to have some uniformity and kill
any (future) logging there in the tests as well.

There is a one-off in pkg/healthcheck because that does log.

### DIFF
--- a/core/dnsserver/log_test.go
+++ b/core/dnsserver/log_test.go
@@ -1,4 +1,4 @@
-package log
+package dnsserver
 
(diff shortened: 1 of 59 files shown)


miekg:
err, remove log/log_test by accident

Issue #1868, opened 2018-06-11
plugin/metrics - error creating the listener does not prevent CoreDNS to start


fturib:
I think (from the code) that it happen on each start or reload, but my use case is on Reload where a failing to listen the port does not prevent to switch to the new CoreDNS instance, although the error is reported in the log.

Looking into the code, I think comes from these lines:

in file `plugin/pkg/uniq/uniq.go`
```
// ForEach iterates for u executes f for each element that is 'todo' and sets it to 'done'.
func (u U) ForEach() error {
	for k, v := range u.u {
		if v.state == todo {
			v.f() . <==== an ERROR raised by f() does not trigger an ERROR on ForEach()
		}
		v.state = done
		u.u[k] = v
	}
	return nil
}
```

miekg:
I think we discussed this at length in an issue and/or pull request? And this was the best we came up with?
(return err here is easy, but can/may have other implication when reloading)

fturib:
Hum .. let me re-read that review then.
Right now the problem is that we implemented an external plugin that kick-off the reload on demand, and reply if the reload was OK. With this error not raised, we reply OK but the Metrics feature does not work.

fturib:
Other implication is: if you raise an error at reload, then the server will keep running on the old instance. And there is nothing to cleanup the already started new instance. 
So, at the end we are in 'improper' state for the plugin that work by global variable (not limited to one instance). We are in improper state in both cases : going forward with the new instance or going backward with the old instance.

fturib:
Maybe a better way is to get a status of each of the Plugins.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/metric..." ]
>Hum .. let me re-read that review then.
>Right now the problem is that we implemented an external plugin that kick-off the reload on demand, and reply if the reload was OK. With this error not raised, we reply OK but the Metrics feature does not work.

If you raise an error, the reload fails and something else fails then. Can't 
remember...

fturib:
About raising an Error : you are right it will not cover all use cases.
Because we are stopping listener when restart . . that means before loading a new Instance of CoreDNS, if any error happen during the setup/start, this new instance will be cancelled and coreDNS will consider continuing on the initial instance - which has now the listener closed and no event to set it up again.

Any Error (in any plugin .. or invalid corefile) will cancel the new Instance and go back to the former instance. Linked or not to the Prometheus plugin.

[...]

Issue #1982, opened 2018-07-17
plugin/metrics - plugin loose control of the listener after several reloads


fturib:
On Restart, plugin close the listener and set the uniqAddr to "todo" in order to be restarted on next "onStart: event.

But the SAME function to call is registered, which means on next "uniqAddr.ForEach" the function 'OnStart" of the now old plugin will be called, and the corresponding "onRestart" will never be called.

After 2 reloads, the Listener is no more closed.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin/metrics - ..." ]
>On Restart, plugin close the listener and set the uniqAddr to "todo" in order to be restarted on next "onStart: event.
>
>But the SAME function to call is registered, which means on next "uniqAddr.ForEach" the function 'OnStart" of the now old plugin will be called, and the corresponding "onRestart" will never be called.
>
>After 2 reloads, the Listener is no more closed.

/label: bug
/plugin: reload

fturib:
Easy way to fix this is to ensure that the Metrics that is shutingdown is releasing its token from UniqAddr

change : uniqAddr.Todo(m.Addr)
by : uniqAddr.Unset(m.Addr)

Then next instance of Metrics, at reload, will refill that uniqAddr with a todo.

fturib:
NOTE: that code is in the PR #1969. But we need to keep only that change ... (make it simple and easy).

calebice:
@fturib I am looking into this. It seems like the code changes required were incorporated into [#1969](https://github.com/coredns/coredns/pull/1969). It seems straightforward enough, but it might take me a little longer to get this set as a pull request. I intend to use this as an opportunity to dig around the code base.

fturib:
@calebice : thanks!

fturib:
@zacheddy is currently working it.

ekleiner:
Note: I have opened an issue about adding OnRestartFailed() to caddy and plan to add it soon. So it may affect the decision about the fix for this bug. See: https://github.com/mholt/caddy/issues/2250

fturib:
I assigned this Issue to me, but that is a placeholder for @ZachEddy

PR #1983, opened 2018-07-17
Plugin/Metrics - add Healther interface and refactor uniqAddr storage

fturib:
Inform when Prometheus cannot provide service because HTTP server could not start.

NOTE: during testing, I found that the Tests could not run properly because the variable uniqAddr that is shared by all instance of metrics plugin was not reset  between Tests.
I moved that variable from global scope of "metrics" to the storage of each caddy instance.
(Caddy did that enhancement in purpose for the "reload" use case).

### Content
- register healther for metrics (return not ok if all listeners are not started properly)
- save the instance of uniqAddr, that is shared by all instances of metrics plugin, into the caddy's instance storage, instead global var.
- add UT with a sample of function RepairHealth for metrics

#1868. 
#1982 

#1868. is resolved by allowing a monitoring of the metrics using the health plugin. 
any monitor can then know when Prometheus does not provide the service and then decide to restart the process of CoreDNS.

#1982 is resolved because we create a new uniqAddr on every new instance : there is no more mismatch with setup of former instance.

README of the plugin updated.

### NOTE: 
this PR overlaps with #1969 - the second to be merged will need a rebase before the merge.

### DIFF
--- a/plugin/metrics/README.md
+++ b/plugin/metrics/README.md
@@ -39,6 +39,13 @@ name "dropped" (without a closing dot - this is never a valid domain name).
 
 This plugin can only be used once per Server Block.
 
+## Health
+
+This plugin implements dynamic health checking. 
+Prometheus plugin is considered unhealthy if the HTTP server instance was not successfully started.
+(see bug description below on server fails to start)   
+
+
 ## Syntax
 
 ~~~
--- a/plugin/pkg/uniq/uniq.go
+++ b/plugin/pkg/uniq/uniq.go
@@ -8,7 +8,7 @@ import (
 
 // U keeps track of item to be done.
 type U struct {
-	sync.Mutex
+	sync.RWMutex
 	u map[string]item
 }
 
@@ -31,18 +31,6 @@ func (u *U) Set(key string, f func() error) {
 	u.u[key] = item{todo, f}
 }
 
-// SetTodo sets key to 'todo' again.
-func (u *U) SetTodo(key string) {
-	u.Lock()
-	defer u.Unlock()
-	v, ok := u.u[key]
-	if !ok {
-		return
-	}
-	v.state = todo
-	u.u[key] = v
-}
-
 // ForEach iterates for u executes f for each element that is 'todo' and sets it to 'done' - if f executes w/o error
 func (u *U) ForEach() error {
 	u.Lock()
@@ -61,8 +49,8 @@ func (u *U) ForEach() error {
 
 // HasTodo inform on weather some things are still in todo mode
 func (u *U) HasTodo() bool {
-	u.Lock()
-	defer u.Unlock()
+	u.RLock()
+	defer u.RUnlock()
 	for _, v := range u.u {
 		if v.state == todo {
 			return true
--- a/plugin/pkg/uniq/uniq.go
+++ b/plugin/pkg/uniq/uniq.go
@@ -2,8 +2,13 @@
 // identical events will only be processed once.
 package uniq
 
+import (
+	"sync"
+)
+
 // U keeps track of item to be done.
 type U struct {
+	sync.Mutex
 	u map[string]item
 }
 
@@ -13,11 +18,13 @@ type item struct {
 }
 
 // New returns a new initialized U.
-func New() U { return U{u: make(map[string]item)} }
+func New() *U { return &U{u: make(map[string]item)} }
 
 // Set sets function f in U under key. If the key already exists
 // it is not overwritten.
-func (u U) Set(key string, f func() error) {
+func (u *U) Set(key string, f func() error) {
+	u.Lock()
+	defer u.Unlock()
 	if _, ok := u.u[key]; ok {
 		return
 	}
@@ -25,7 +32,9 @@ func (u U) Set(key string, f func() error) {
 }
 
 // SetTodo sets key to 'todo' again.
-func (u U) SetTodo(key string) {
+func (u *U) SetTodo(key string) {
+	u.Lock()
+	defer u.Unlock()
 	v, ok := u.u[key]
 	if !ok {
 		return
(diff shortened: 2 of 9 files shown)


plugin/pkg/uniq/uniq.go
```diff
@@ -13,19 +18,23 @@ type item struct {
...
+	defer u.Unlock()
 	if _, ok := u.u[key]; ok {
 		return
 	}
 	u.u[key] = item{todo, f}
 }
 
 // SetTodo sets key to 'todo' again.
-func (u U) SetTodo(key string) {
+func (u *U) SetTodo(key string) {
```
miekg: Why did you change this to  a pointer receiver?

plugin/pkg/uniq/uniq.go
```diff
@@ -34,18 +43,34 @@ func (u U) SetTodo(key string) {
...
 		}
-		v.state = done
-		u.u[k] = v
 	}
 	return nil
 }
 
+// HasTodo inform on weather some things are still in todo mode
+func (u *U) HasTodo() bool {
+	u.Lock()
```
miekg: This only needs a read lock

plugin/pkg/uniq/uniq.go
fturib: Well, by setting the a value in U, I expect to modify that U structure.
It should then be a pointer receiver (instead of a value receiver).

I understand that was working so far. But I guess because the copy is sharing the same map.

But as I introduced the Mutex, this one was not effective if I use a value receiver : each copy of U was running on its own Mutex => I add a warning that mutex was inefficient and I could verify on UT that I had race conditions.

see : https://tour.golang.org/methods/4
https://programming.guide/go/pointer-vs-value-receiver.html

plugin/pkg/uniq/uniq.go
fturib: ok.

plugin/pkg/uniq/uniq.go
fturib: Right. a map varlue is a pointer to the head of map. So when copy we copy that pointer. Which make any change in that map works from the original or from the copy.

plugin/metrics/README.md
```diff
@@ -39,6 +39,13 @@ name "dropped" (without a closing dot - this is never a valid domain name).
 
 This plugin can only be used once per Server Block.
 
+## Health
+
+This plugin implements dynamic health checking. 
+Prometheus plugin is considered unhealthy if the HTTP server instance was not successfully started.
```
miekg, line 45: so, we will not fail binding to the port? That was an earlier plan, no? Did we decide we could not do that ?

fturib: as you started to say in the issue #1868, the problem to fail is that if we fail to start the new server, the CoreDNS will continue service on the old server .. but the listener is in unknown state (most likely started to stop but not finished). Which means Metrics most likely stopped to serve information on either new or old server.

It does not help to automate the monitoring of CoreDNS, unless we completely stop CoreDNS.

Adding Metrics to Health and inform when metrics is unhealthy allow any monitor to take a decision for stop/start CoreDNS

fturib:
@miekg : let me know if you can approve and merge here. This fix is blocking a push to production on our side. Thanks.!

plugin/metrics/README.md
miekg: I keep finding this odd. We should probably crash the startup if we fail to bind, i.e. not fail the reload, crash the entire process. Seems simplest and avoid a whole bunch of code to work around it. I do wonder however if log.Fatalf would crash everything or just the reloading process.

I also don't understand why your monitoring doesn't see the monitoring endpoint disappearing and needs the health endpoint to do something

fturib: # Short

- **I can remove the healther implementation. I do not need it. But I would like to keep the move out of global var - which also fix #1868 **
- **I would not like to have to stop the process if Metrics cannot listen. At least for few weeks. That would block IB that try to go to production right now.**

# Longer reply 

I think this PR contains several changes, and maybe I should go one by one:

## 1- issue #1868
it is the impact of Reload on the Metrics plugin that is using uniq resources (the Listeners). 
IMO the root cause is the usage of global var to save information : we cannot cleanup this information during a reload. That get worse when the Reload fails for some reason and we expect to go back to initial server.
=> Her, I move the uniqAddr info from global var, to Instance related storage. But still share by all Metrics structures used during that instance.


## 2- Healther and what to do when the listener (uniq resource) cannot start.
I was going toward Health plugin for consistency : it should warn when CoreDNS service is not correct.
But I agree that EXIT the process is another valid solution.
NOTE: having the monitoring check the Metrics endpoint along with Health endpoint can work too .. but that is inconsistent with Health definition.

## 3- the size of the change ... hum here I am not following too much. 
The point is to have the plugin working correctly.
Here I tried to fix the following issues I see in the current code:
- move away from the global var 
- change init uniqAddr in Setup- away from the go function that we are not sure when it will run (I mean nothing ensure it will run before the Start of the server)
- set status in UniqAddr to "done" only when listener is really started. Seems sanity. But also allow to have either an external check (Health) or retry (if we want ...)
- mutex on UniqAddr functions. is sanity. We are lucky so far this lib is used only at setup or event that are synced.

As the code is done and proposed now, isn't it better to checked-in ?

- the one that want to monitor on endpoint Metrics can still continue
- the issue #1868 is fixed with this code
- lib uniqAddr is thread-safe and dan be reused 

and one usage I may have (depends internal Infoblox) is ability to retry the Start of the Metrics from within an external plugin which is doing CLI+monitoring.


## What I expect from metrics for IB usage
- fix of #1868 (after several reload, Metrics lost ability to stop/start the listener according to Corefile)
- ability to know it Metrics failed to launch the Listener
- ability to retry (ability to know is similar to Healther). and the retry is described into RepairHealth in UT.

NOTE: Right now, the EXIT in Metrics in case of fail to listen would hurt IB use case. 
They need to move up to Go 1.10. I am pushing on it. But not done yet.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Plugin/Metric..." ]
>- fix of #1868 (after several reload, Metrics lost ability to stop/start the listener according to Corefile)
>- ability to know it Metrics failed to launch the Listener
>- ability to retry (ability to know is similar to Healther). and the retry is described into RepairHealth in UT.
>
>NOTE: Right now, the EXIT in Metrics in case of fail to listen would hurt IB use case.
>They need to move up to Go 1.10. I am pushing on it. But not done yet.

go 1.10 has a fix, so you should just use that. I understand that moving an org 
to a new version is hard, but we can't just add code because that is somehow 
easier for you right now.

Coming back to failing the reload - metrics plugin should probably call panic 
and let the - whatever is managing it - restart it. Fun thing is that panic are 
captured by the errors plugin \o/

/Miek

--
Miek Gieben

miekg:
[ Quoting <miek@miek.nl> in "Re: [coredns/coredns] Plugin/Metric..." ]
>Coming back to failing the reload - metrics plugin should probably 
>call panic and let the - whatever is managing it - restart it. Fun 
>thing is that panic are captured by the errors plugin \o/

or log.Fatalf that calls os.Exit

fturib:
About go version, I asked IB people, and look like we have an issue besides this version.

> I have asked the QA and  Gowtham mentions that the issue is observed on all  3 platforms :  Baremetal,  OVA,  and NIOS.
>  
> We are using golang 1.10.x   version,  so I guess it’s not related to the go issue that we are thinking of.

fturib:
I will remove this PR, until we have the root cause of Listener failing to listen when restart.
Seems not the known Golang issue .. then, what else ?

the 2 issues are still there, but maybe not the important part.

PR #1985, opened 2018-07-19
plugin/hosts: add log.Debug

miekg:
Logs the amount of entries we parsed after every parse, but only when
debug is loaded.


yongtang (approved):
LGTM

PR #1989, opened 2018-07-20
New plugin: loop

miekg:
Add a plugin that detects loops. It does this by sending an unique query
to our selves. If we see the query more than twice we stop the process.
If there isn't a loop, the plugin disables it self and becomes a noop
plugin.


miekg:
Added @chrisohaver as reviewer as well, as I've volunteered him in the owners file

greenpau:
@miekg , this one is pretty cool! 👍

chrisohaver (approved):
lgtm

greenpau (approved):
LGTM 👍

plugin/loop/loop.go
```diff
@@ -0,0 +1,94 @@
...
+		return plugin.NextOrFailure(l.Name(), l.Next, ctx, w, r)
+	}
+
+	if state.Name() == l.qname {
+		l.inc()
+	}
+
+	if l.seen() > 2 {
+		// TODO(miek): add log.Fatal(f) and use here.
+		log.Errorf("Seen \"HINFO IN %s\" more than twice, loop detected", l.qname)
```
chrisohaver: Can we also log the IP address that the loop was detected in?  i.e the ip that the request came in on...

miekg: Added, I'm my local testing this will always be localhost because there is no remote server at play.

chrisohaver:
looks great.  Thanks Miek for whipping this one out lightning fast!

plugin/loop/loop.go
```diff
@@ -0,0 +1,94 @@
...
+	if zone == "" {
+		return plugin.NextOrFailure(l.Name(), l.Next, ctx, w, r)
+	}
+
+	if state.Name() == l.qname {
+		l.inc()
+	}
+
+	if l.seen() > 2 {
+		// TODO(miek): add log.Fatal(f) and use here.
```
fturib: I understand we create only one Loop struct. and we send several queries (every 2 sec).
so, this l.seen > 2 will be true at the third msg we send .. no ?

Should we have here a count for each msg.id we create for the test ?

plugin/loop/setup.go
```diff
@@ -0,0 +1,90 @@
...
+				for _, lh := range conf.ListenHosts {
+					addr := net.JoinHostPort(lh, conf.Port)
+					if _, err := l.exchange(addr); err != nil {
+						continue
+					}
+					ok++
+				}
+
+				if ok == len(conf.ListenHosts) {
+					go func() {
```
fturib: looks like a way to disable sooner if at least one msg sent per Listening host.
Because of the double loop (30 sec, and for each Host) - and because of 'continue" at line 46, we are not sure that all Host are tested. (some may be tested several time, some not at all).

plugin/loop/setup.go
```diff
@@ -0,0 +1,90 @@
...
+	dnsserver.GetConfig(c).AddPlugin(func(next plugin.Handler) plugin.Handler {
+		l.Next = next
+		return l
+	})
+
+	// Send query to ourselves and see if it end up with us again.
+	c.OnStartup(func() error {
+		// Another Go function, otherwise we block startup and can't send the packet.
+		go func() {
+			deadline := time.Now().Add(30 * time.Second)
```
fturib, line 38: I guess 30 sec could be adapted to >= 2 * Nb of Listening Host.
I agree 15 seems already a big number ...

chrisohaver:
> Added, I'm my local testing this will always be localhost because there is no remote server at play.

Ah - I meant to log the IP address that coredns received the packet on (i.e. the destination ip of the incoming packet).
In the Ubuntu resolved case, this would be 127.0.0.53, which may make it easier to understand (since that IP is in resolv.conf).

plugin/loop/setup.go
```diff
@@ -0,0 +1,90 @@
...
+
+func parse(c *caddy.Controller) (*Loop, error) {
+	_ = c.NextArg()
+	if c.NextArg() {
+		return nil, c.ArgErr()
+	}
+
+	zone := "."
+	if len(c.ServerBlockKeys) > 0 {
+		zone = plugin.Host(c.ServerBlockKeys[0]).Normalize()
```
fturib: I think that setup is called for each of the serverBlocKeys.
There is a way to know which one we are called for : use c.Key

fturib (commented):
I think there may be some issues if bloc has several keys, or if config has several hosts.
Please see my comments.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] New plugin: l..." ]
>I understand we create only one Loop struct. and we send several queries (every 2 sec).
>so, this l.seen > 2 will be true at the third msg we send .. no ?

excellent point! I've only the first listenhost now. We can't use the query Id, 
because if something else forwards this again, it will be a different Id.

>I guess 30 sec could be adapted to >= 2 * Nb of Listening Host.
>I agree 15 seems already a big number ...

I'm fine with 15, but I think load zones in the file plugin blocks startup, so 
that can be a long time...

>I think that setup is called for each of the serverBlocKeys.
>There is a way to know which one we are called for : use c.Key

I'll add the standard code that this can only be used once.

Thanks for reviewing! PTAL.

fturib:
> I'll add the standard code that this can only be used once.

It is not about having several time the "loop" plugin in the same stanza.
It is about having several keys on top of the stanza, like:

```
coredns.io:53, coredns.com:443, infobloxcom:667 {
    loop
    procxy ...
}
```

the setup of loop will be triggered 3 times : one for each key on the top.
and from the setup(...) function you get that key with : c.Key

miekg:
This should result in three different queries because I suffix the zone for
each generated name. Let me see if I can add a test for that.

miekg:
~~~
coredns.io:1053 coredns.com:1054 infoblox.com:1055 {
    loop
    forward . 127.0.0.1:1054
    log
}
~~~
crashes with:
~~~
127.0.0.1:47153 - [20/Jul/2018:18:57:23 +0100] 738 "HINFO IN 1687376426597664323.1119963552489870178.coredns.io. udp 68 false 512" REFUSED qr,rd 68 0.000187187s
127.0.0.1:34498 - [20/Jul/2018:18:57:23 +0100] 45646 "HINFO IN 1687376426597664323.1119963552489870178.coredns.io. udp 68 false 512" REFUSED qr,rd 68 0.000087221s
2018/07/20 18:57:23 [FATAL] plugin/loop: Seen "HINFO IN 1687376426597664323.1119963552489870178.coredns.io." to [::]:1053 more than twice, loop detected
~~~

luxas:
@rajansandeep @miekg @johnbelamaric Will this option make it into the next k8s manifest that has a new enough CoreDNS version?

miekg:
This is as yet unreleased, but CoreDNS 1.2.1 will have it. What's the
timeframe?

chrisohaver:
> Will this option make it into the next k8s manifest that has a new enough CoreDNS version?

Yes

chrisohaver:
> What's the timeframe?

Plenty of time to get it into the next k8s release.

chrisohaver:
> Plenty of time to get it into the next k8s release.

Oh... time moving more quickly than I reckoned. So not plenty of time.  But enough time ...

1.12 feature code freeze is July 31.  so... in a week...

fturib:
In that case it would not be considered as a feature.
rather update of existing feature. Code slush is Aug 28th.
We would need to have CoreDNS 1.2.1 released by Aug20th.

chrisohaver:
Yes, thanks. I relayed the wrong freeze ... FYI...

https://github.com/kubernetes/sig-release/blob/master/releases/release-1.12/release-1.12.md

PR #1991, opened 2018-07-21
release: add sha256sum files

### LINKED ISSUE #1964: Publish sha256sums on release
Since on release binaries for various platforms and architectures are published, it's important to have a declared sha256sum for each binary in order to be able to verify the downloaded archive.

The [`coredns-bin`](https://aur.archlinux.org/packages/coredns-bin/) package on aur would directly benefit from this for example.

miekg:
Generate the sha256 sum of each asset. These should all be automatically
be uploaded in the github-push target.

This isn't tested for obvious reason, other than `make -f Makefile.release sum` creates the correct files.

Fixes #1964

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -55,7 +55,7 @@ endif
 all:
 	@echo Use the 'release' target to build a release, 'docker' for docker build.
 
-release: pre build tar
+release: pre build tar sum
 
 docker: docker-build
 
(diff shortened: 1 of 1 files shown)


yongtang (approved):
LGTM

Makefile.release
```diff
@@ -86,6 +86,12 @@ tar:
 	    tar -zcf release/$(NAME)_$(VERSION)_linux_$$arch.tgz -C build/linux/$$arch $(NAME) ;\
 	done
 
+.PHONY: sum
+sum:
+	for asset in `ls -A release/*tgz`; do \
+	    sha256sum $$asset > $$asset.sha256; \
```
grobie, line 92: @miekg Why do you publish a separate sha256sum file for each asset? I think the idiomatic way is to provide a single file (often called `sha256sum.txt`, or `SHA256SUMS`) with one line per artifact. Providing separate files seems to just unnecessarily complicate the process of redistributing binaries while ensuring consistency.

PR #1993, opened 2018-07-21
Test cases for address normalization with specified protocols

fntlnz:
Test that the behavior of address normalization and transport extraction is consistent for all the transports.
Signed-off-by: Lorenzo Fontana <lo@linux.com>


miekg:
Nice. Thank you

PR #1996, opened 2018-07-23
Document current limitations of loop

chrisohaver:

Document current limitations of loop.  Basically to set expectations for users. The plugin doesn't try to detect every possible kind of loop.


chrisohaver:
@miekg, PTAL - this is per my understanding of the implementation, I have not tested these.   So may be wrong here in some places.

chrisohaver:
... actually, I think the bits about the default forwarder, and stub zones is wrong.  I'll remove those...

miekg (approved):

Issue #1995, opened 2018-07-23
plugin/loop - test request is sent only to the first domain


fturib:
Therefore, we do not test the subsequent domains.

For instance, with this corefile, that define a loop on the second domain only (isidore.com), the Loop does not detect anything:

```
francois.com:5362, isidore.com:5362 {
     loop
     proxy . 127.0.0.1:5361 {
        protocol dns force_tcp
     }
     log
     errors
}
```

if an another external CoreDNS is running with a configuration that loop only on the second domain, then it will not be detected by the above CoreDNS.

external DNS that reply a loop for isidore.com - and not for francois.com
```
# no loop on francois.com
francois.com:5361 {
     loop
     proxy . 8.8.8.8:53 {
        protocol dns force_tcp
     }
     log
     errors
}
# loop on isidore.com
isidore.com:5361 {
     loop
     proxy . 127.0.0.1:5362 {
        protocol dns force_tcp
     }
     log
     errors
}
```

fturib:
Wait a minute ... my test is wrong.

chrisohaver:
related ...  I believe the following loop would not get detected, since the zone of loop is `.`, and it wont try a query in `stub.domain.com.`

```
. {
     loop
     proxy stub.domain.com. 127.0.0.1
     proxy . 8.8.8.8
}
```

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/loop -..." ]
>related ...  I believe the following loop would not get detected, since the zone of loop is `.`, and it wont try a query in `stub.domain.com.`
>
>```
>. {
>     loop
>     proxy stub.domain.com. 127.0.0.1
>     proxy . 8.8.8.8
>}
>```

yes.

chrisohaver:
This is actively being working on.

miekg:
who's working on this and what's the eta for this relatively simple change?

chrisohaver:
@ZachEddy, were you still working on this?

PR #1997, opened 2018-07-23
plugin/loop - ensure zone/qname declared on the right domain

fturib:
error (I guess a corner case though) in the setup of plugin loop

#1995 

None.

### DIFF
--- a/plugin/loop/setup.go
+++ b/plugin/loop/setup.go
@@ -73,8 +73,8 @@ func parse(c *caddy.Controller) (*Loop, error) {
 			return nil, c.ArgErr()
 		}
 
-		if len(c.ServerBlockKeys) > 0 {
-			zone = plugin.Host(c.ServerBlockKeys[0]).Normalize()
+		if len(c.Key) > 0 {
+			zone = plugin.Host(c.Key).Normalize()
 		}
 	}
 	return New(zone), nil


plugin/loop/setup.go
```diff
@@ -73,8 +73,8 @@ func parse(c *caddy.Controller) (*Loop, error) {
 			return nil, c.ArgErr()
 		}
 
-		if len(c.ServerBlockKeys) > 0 {
-			zone = plugin.Host(c.ServerBlockKeys[0]).Normalize()
+		if len(c.Key) > 0 {
```
miekg, line 76: I don't understand this change

fturib: if you have a server bloc with that header:

domain1, domain2 {
   loop
}

the structure loop will be created twice, one for each Key of the bloc.

the setup of loop will be called 2 times

one with the following informations:

```
c.ServerBlockKeys = ["domain1", "domain2"]
c.Key = "domain1"
```
and second time with:
```
c.ServerBlockKeys = ["domain1", "domain2"]
c.Key = "domain2"
```

if you keep creating the Loop with c.ServerBlockKeys[0], you will always send a query to the domain "domain1" even though you are serving "domain2".

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/loop -..." ]
>if you keep creating the Loop with c.ServerBlockKeys[0], you will always send a query to the domain "domain1" even though you are serving "domain2".

so? This is to detect a loop? Loop detects a loop. If you have advanced config 
that requires multiple queries or what not then I hope you know what you are 
doing.

What does this add? 
Even then I don't understand the change, what's c.Key? No plugin uses this? How 
is this different. Why do we need it here?

fturib:
hum. 
I try to help detecting the loops.

Right now, if you have this config:

```
one.com, two.com {
   loop
   proxy . 8.8.8.8
}
```

If you have a loop on one.com, you will detect it with "loop' plugin.
if you have a loop on two.com,  you will not detect it with "loop" plugin, because it will not send a query to "two.com", but to "one.com".

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/loop -..." ]
>hum.
>I try to help detecting the loops.
>
>Right now, if you have this config:
>
>```
>one.com, two.com {
>   loop
>   proxy . 8.8.8.8
>}
>```
>
>If you have a loop on one.com, you will detect it with "loop' plugin.
>if you have a loop on two.com,  you will not detect it with "loop" plugin, because it will not send a query to "two.com", but to "one.com".

Ok, fair enough. I would say if you don't have a loop for first, you don't have 
a loop period and as Chris has flagged you can still create a loop, that even 
with this change won't be detected. We don't wont to kill the server because we 
send too many queries; so we want to be *very* conservative here.

With the whole looping business: "admin does stupid shit, gets shot in foot", 
you should know better.

fturib:
hum... I understand we do not need this PR.
Closing.

PR #2007, opened 2018-07-27
request: re-adjust size for edns0 RR

miekg:
Cherry-picked the test and benchmark commits from #1962 and make the failing test not fail, by adding minimal fix in request.go. This makes size 12 bytes smaller if we need to add back the OPT RR inthe message.
    
Closes: #1962


chrisohaver:
`gometalinter --install goimports > /dev/null` is failing.
Changes here recently?

grobie:
Awesome, thanks @miekg!

Issue #2010, opened 2018-07-27
Pull Scrub out of plugins


miekg:
https://github.com/coredns/coredns/pull/1962#issuecomment-404265089 suggested by @grobie .

Very good idea as it fits well with the idea that plugins not requires a whole bunch of (expert) knowledge of the DNS protocol.

PR #2012, opened 2018-07-28
Default to scrubbing replies in the server

miekg:
Every plugin needs to deal with EDNS0 and should call Scrub. Move this
into the server and wrapping the ResponseWriter into a ScrubWriter that
handles these bits for us. Result:

Less code and faster, because multiple chained plugins could all be
calling scrub and SizeAndDo - now there is just one place.

Most tests in file/* and dnssec/* needed adjusting because in those unit
tests you don't see OPT RRs anymore. The DNSSEC signer was also looking
at the returned OPT RR to see if it needed to sign - as those are now
added by the server (and thus later), this needed to change slightly.

Fixes: #2010

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -237,6 +237,9 @@ func (s *Server) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 
 	var dshandler *Config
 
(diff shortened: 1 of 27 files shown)


miekg:
I don't understand why the PR build keeps on failing - https://travis-ci.org/coredns/coredns/jobs/409281591, it's caching an older version or something.

miekg:
I'm going to merge this after the 1.2.1 release

PR #2013, opened 2018-07-28
Make presubmit actually fail the build

miekg:
The exit code wasn't properly acted upon in the makefile.
Make filename-hyphen actually return an non-zero exit code.


yongtang (approved):
LGTM

PR #2018, opened 2018-07-31
dnstap.TapperFromContext always returns nil

2buisson:

Because of this bug, plugin proxy wasn't sending its DNS messages to dnstap.

I don't know what caused this since it was introduced, but now there will be a test on this function.


yongtang (approved):
LGTM

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] dnstap.Tapper..." ]
>yongtang approved this pull request.

/merge

PR #2023, opened 2018-08-04
plugin/forward: add dnstap support

2buisson:

* https://github.com/coredns/coredns/issues/1448

### DIFF
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -114,12 +114,18 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts options
 
 	conn.SetReadDeadline(time.Now().Add(p.readTimeout()))
 	ret, err := conn.ReadMsg()
+
+	taperr := logToDnstap(ctx, p.transport.addr, proto, state.Req, ret, start)
+
+	// check for error from previous call to conn.ReadMsg()
 	if err != nil {
 		p.updateRtt(maxTimeout)
 		conn.Close() // not giving it back
 		if err == io.EOF && cached {
 			return nil, ErrCachedClosed
 		}
+
+		// Connect have priority over dnstap errors
 		return ret, err
 	}
 
@@ -136,7 +142,8 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts options
 	RcodeCount.WithLabelValues(rc, p.addr).Add(1)
 	RequestDuration.WithLabelValues(p.addr).Observe(time.Since(start).Seconds())
 
-	return ret, nil
+	// no Connect error, return the dnstap error instead
+	return ret, taperr
 }
 
 const cumulativeAvgWeight = 4
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -115,7 +115,7 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts options
 	conn.SetReadDeadline(time.Now().Add(p.readTimeout()))
 	ret, err := conn.ReadMsg()
 
-	taperr := logToDnstap(ctx, p.transport.addr, proto, state.Req, ret, start)
+	taperr := logToDnstap(ctx, conn.RemoteAddr(), state.Req, ret, reqTime)
 
 	// check for error from previous call to conn.ReadMsg()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


plugin/forward/connect.go
```diff
@@ -114,12 +114,18 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts options
 
 	conn.SetReadDeadline(time.Now().Add(p.readTimeout()))
 	ret, err := conn.ReadMsg()
+
+	taperr := logToDnstap(ctx, p.transport.addr, proto, state.Req, ret, start)
```
rdrozhdzh: `proto` here does not always corresponds to actual protocol, e.g. `transport.Dial()` can override to use `tcp-tls` if TLS was configured. Perhaps, we could address this issue in a separate PR

rdrozhdzh: you can pass `reqTime` instead of `start` as last argument

plugin/forward/dnstap.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/miekg/dns"
+)
+
+func logToDnstap(ctx context.Context, host, proto string, req, ret *dns.Msg, start time.Time) error {
+	t := dnstap.TapperFromContext(ctx)
+	if t == nil {
+		return nil
+	}
+
+	b := msg.New().Time(start).HostPort(host)
```
rdrozhdzh: Consider renaming `Time()` -> `WithTime()`, `HostPort()` -> `WithHostPort()`. Current names look like getters.

plugin/forward/dnstap_test.go
```diff
@@ -0,0 +1,79 @@
...
+		ret.SetReply(r)
+		ret.Answer = append(ret.Answer, test.A("example.org. IN A 127.0.0.1"))
+		w.WriteMsg(ret)
+	})
+	defer s.Close()
+
+	c := caddy.NewTestController("dns", "forward . "+s.Addr)
+	f, err := parseForward(c)
+	if err != nil {
+		t.Errorf("Failed to create forwarder: %s", err)
```
rdrozhdzh, line 42: consider using `t.Fatalf()` here, no sense to continue if forwarder has not been created

plugin/forward/dnstap_test.go
```diff
@@ -0,0 +1,79 @@
...
+	defer f.OnShutdown()
+
+	dtapHandler.Next = f
+
+	m := new(dns.Msg)
+	m.SetQuestion("example.org.", dns.TypeA)
+	rec := dnstest.NewRecorder(&test.ResponseWriter{})
+
+	if _, err := dtapHandler.ServeDNS(context.TODO(), rec, m); err != nil {
+		t.Fatal("Expected to receive reply, but didn't")
```
rdrozhdzh, line 54: consider using `t.Errorf()` here and below where we can safely do more checks

plugin/forward/dnstap.go
```diff
@@ -0,0 +1,55 @@
...
+		b.SocketProto = tap.SocketProtocol_TCP
+	} else {
+		b.SocketProto = tap.SocketProtocol_UDP
+	}
+
+	if t.Pack() {
+		b.Msg(req)
+	}
+
+	m, err := b.ToOutsideQuery(tap.Message_FORWARDER_QUERY)
```
rdrozhdzh, line 27: consider adding a config option to dnstap plugin which defines which types of dnstap messages should be logged and which ones should not, e.g.
```
dnstap <socket> {
    messages CQ CR FQ FR AR RQ
}
```
Other plugins could send (or not to send) particular dnstap messages based on this configuration

2buisson: I need to think about this

rdrozhdzh: This could be done in a separate PR

plugin/forward/dnstap_test.go
2buisson: Copy-paste from https://github.com/coredns/coredns/blob/master/plugin/forward/proxy_test.go#L53

plugin/forward/dnstap_test.go
2buisson: Also copy-paste

miekg:
fixed via 403d42ce

Issue #2017, opened 2018-07-31
Ability to configure a smaller size cache 


huynhmb359:
The current cache capacity has the minimum cache capacity of 131,072 entries.   We did some experiment and noticed that it would consume up to 500 MB after filling up the cache.   This causes issue for running coredns on platform where memory is limited. 
Could we add support to configure cache size smaller than the current imposed minimum?

fturib:
@huynhmb359 : what is the size you expect ?

NOTE: if you have several caches in your Corefile, please provide the size (in number of entries) you would like to be able to set up, so we have a global view of what is expected.

huynhmb359:
@fturib,  we have several caches in our corefile.    In term of entries,  we would like to have about 5K of entries for the smallest one and 65K of entries for average.

fturib:
Today the cache is build with 256 shards. 
Default size for each shard is 512 entries.

65k means - 256 shards of 128 entries which seems ok
5k  means - 256 shards of 20 entries which seems pretty small number of entries and will not work properly if the hashkey is not well balanced (eviction is per shard)

@miekg : what was the initial point to introduce shard ? is it concurrent access or efficiency to memory access ?

In other words, for a number as low as 20 entries, should we keep the 256 shards or switch to a lower number of shards if the max size of the cache is low ?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Ability to co..." ]
>Today the cache is build with 256 shards.
>Default size for each shard is 512 entries.
>
>65k means - 256 shards of 128 entries which seems ok
>5k  means - 256 shards of 20 entries which seems pretty small number of entries and will not work properly if the hashkey is not well balanced (eviction is per shard)
>
>@miekg : what was the initial point to introduce shard ? is it concurrent access or efficiency to memory access ?

concurrency. And I think we may want to up this to 1024 by default (for even 
better concurrency)

/Miek

--
Miek Gieben

fturib:
If we go with 1024 shards, concurrency will be better, but, what is the down side when entries per map is small ?
- memory point of view (pre-allocation of 1k maps ?)
- eviction more often (I am not sure how to compute it .. I guess that depend of the well building of the hashkey).

for information, I found here a link that show memory used by empy maps (for information): https://stackoverflow.com/questions/15313105/memory-overhead-of-maps-in-go

PR #2030, opened 2018-08-13
Remove setting minimum number of cache entries per shard

huynhmb359:
This pull request removes the minimum requirement for number of cache entries per shard. 
With minimum requirement,  the current cache capacity has the minimum cache capacity of 131,072 entries which could cause issue for running coredns on platform where memory is limited.

#2017 

README.md in plugin/cache need to be updated with the change.

### DIFF
--- a/plugin/pkg/cache/cache.go
+++ b/plugin/pkg/cache/cache.go
@@ -31,10 +31,6 @@ type shard struct {
 // New returns a new cache.
 func New(size int) *Cache {
 	ssize := size / shardSize
-	if ssize < 512 {
-		ssize = 512
-	}
-
 	c := &Cache{}
 
 	// Initialize all the shards
--- a/plugin/pkg/cache/cache.go
+++ b/plugin/pkg/cache/cache.go
@@ -31,6 +31,9 @@ type shard struct {
 // New returns a new cache.
 func New(size int) *Cache {
 	ssize := size / shardSize
+	if ssize < 4 {
+		ssize = 4
+	}
 	c := &Cache{}
 
 	// Initialize all the shards


plugin/pkg/cache/cache.go
```diff
@@ -31,10 +31,6 @@ type shard struct {
 // New returns a new cache.
 func New(size int) *Cache {
 	ssize := size / shardSize
-	if ssize < 512 {
-		ssize = 512
-	}
-
```
fturib: No minimum at all ?
I guess to work properly we should verify that each shard of the cache is at least 1 or 2 size large No ?

huynhmb359: I think without having the minimum,  0 will has the effect of disabling the cache.

fturib:
I made a test of Cache with debugger.
And, if the size of each shard is finally compute to 0, then you will get a cache of 1.
Which means a cache of 1 for each shard, potentially up to a cache of 512, - that  depends the distribution of the key.

plugin/pkg/cache/cache.go
miekg: are you sure?

What are you seeing OOMs or something?

fturib:
@miekg : are you ok to merge this PR and have it in 1.2.1 ? (or 1.2.2) ? 

after that investigation about size of the cache and memory used, https://github.com/kubernetes/kops/issues/5652, I think it is better we get a cache of size 10k by default (instead of 131k today).

and fit memory in current size of 170 MB, for CoreDNS pod in k8s.

What do you think ?

chrisohaver (changes requested):
Need to update docs too...  default cache size will be 10000 now I think (as i think it used to be before the refactor about a year ago).

(no code changes for 10K default needed, I think it will behave that way as the PR is now, just need to update README)

chrisohaver (approved):
i'll update the docs for you in another PR

plugin/pkg/cache/cache.go
fturib: Someone having OOM on k8s, did investigation on the memory used by cache : https://github.com/kubernetes/kops/issues/5652

Looks like we are using an average of 850 bytes for each entry of the cache.
and because we do not clean it when TTL are done (to avoid extra locks on the shards), then the cache keep growing up to 260MB. with other usage of memory, it is then recommended to have the memory allocated to 320MB **if you have only one cache**.
Things are getting worst if you use several stanza with cache on each of them ...

I guess that having ability to define small cache can help a lot (especially for stub domains).

With this change of code, we are now back to 10k entries as defaults (instead of 131k). Which means each cache, if you keep default value would use about 10Mb, and you could tune smaller cache for the stub domains.

chrisohaver:
see docs update PR here #2062 ...

plugin/pkg/cache/cache.go
miekg: i don't think we want to totally remove this check, but lowering is fine.

We also bluntly cache entire messages which is also inefficient memory, but easy to reason about.
And note, note: duing the cache refactor this was an oversight on my part, that I didn't lower this in the first place.
Also note: it needs a documentation update; as we mention the number in the README

fturib: @chrisohaver : can you update your PR of the README, adding the real change of check for cache. Thong is busy on production issue and will not be available soon to update this PR.

I would propose something like:

```
if ssize < 4 {
		ssize = 4
	}
```

A size of 4 per shard would do a total 2k entries for the smallest cache : I think that is good enough for who want a cache per simple domains. 

Default will remains at 10k cache.

And who want a big cache can still define it.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Remove settin..." ]
>
>
>
>> @@ -31,10 +31,6 @@ type shard struct {
> // New returns a new cache.
> func New(size int) *Cache {
> 	ssize := size / shardSize
>-	if ssize < 512 {
>-		ssize = 512
>-	}
>-
>
>@chrisohaver : can you update your PR of the README, adding the real change of check for cache. Thong is busy on production issue and will not be available soon to update this PR.
>
>I would propose something like:
>
>```
>if ssize < 4 {
>		ssize = 4
>	}
>```
>
>A size of 4 per shard would do a total 2k entries for the smallest cache : I think that is good enough for who want a cache per simple domains.
>
>Default will remains at 10k cache.
>
>And who want a big cache can still define it.

sgtm

fturib (approved):
LGTM

PR #2034, opened 2018-08-15
plugin/forward: Split setup to reuse it from external plugins

eugen-kleiner-mesh:
It splits setup of forward plugin to reuse it from external plugin "alternate".
No logic changed.
https://github.com/coredns/fallback/issues/9
No.


fturib (approved):
LGTM.

PR #2036, opened 2018-08-17
Metrics listener fix

### LINKED ISSUE #1982: plugin/metrics - plugin loose control of the listener after several reloads

ZachEddy:

When the metrics address is changed inside the `Corefile` and a reload is triggered, CoreDNS will continue exporting metrics to the old address. This was reported in #1982, so check out that issue for a more in-depth description and some discussion about the bug itself.

This PR  resolves the issue by modifying the metrics plugin's `OnRestart` behavior. Before CoreDNS reloads, the metrics plugin now stops the existing Prometheus exporter and removes the current metrics address from the `Metrics.uniqAddr.Todo` field. When CoreDNS spins back up, it creates `Metrics.uniqAddr` using the new metrics configuration.

Resolves #1982.

Other than inline documentation, I don't think any documentation changes are necessary; this is just a bug fix.

### DIFF
--- a/plugin/metrics/metrics.go
+++ b/plugin/metrics/metrics.go
@@ -147,7 +147,7 @@ var ListenAddr string
 
 // shutdownTimeout is the maximum amount of time the metrics plugin will wait
 // before erroring when it tries to close the metrics server
-var shutdownTimeout time.Duration = time.Second * 1
+const shutdownTimeout time.Duration = time.Second * 5
 
 var buildInfo = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 	Namespace: plugin.Namespace,
--- a/plugin/metrics/metrics.go
+++ b/plugin/metrics/metrics.go
@@ -2,10 +2,12 @@
 package metrics
 
 import (
+	"context"
 	"net"
 	"net/http"
 	"os"
 	"sync"
+	"time"
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/metrics/vars"
@@ -22,6 +24,7 @@ type Metrics struct {
 	ln      net.Listener
 	lnSetup bool
 	mux     *http.ServeMux
+	srv     *http.Server
 
 	zoneNames []string
 	zoneMap   map[string]bool
@@ -94,9 +97,9 @@ func (m *Metrics) OnStartup() error {
 
 	m.mux = http.NewServeMux()
 	m.mux.Handle("/metrics", promhttp.HandlerFor(m.Reg, promhttp.HandlerOpts{}))
-
+	m.srv = &http.Server{Handler: m.mux}
 	go func() {
-		http.Serve(m.ln, m.mux)
+		m.srv.Serve(m.ln)
 	}()
 	return nil
 }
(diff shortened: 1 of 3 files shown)


plugin/pkg/uniq/uniq.go
```diff
@@ -38,10 +45,11 @@ func (u U) SetTodo(key string) {
 func (u U) ForEach() error {
 	for k, v := range u.u {
 		if v.state == todo {
-			v.f()
+			if err := v.f(); err == nil {
```
fturib: What do we do if there is an error here ?
I think this change of code goes with #1868. 

Although it is not wrong, I think it is better to let that change for the PR of that issue.

plugin/metrics/metrics.go
```diff
@@ -138,6 +145,10 @@ func keys(m map[string]bool) []string {
 // we listen on "localhost:0" and need to retrieve the actual address.
 var ListenAddr string
 
+// shutdownTimeout is the maximum amount of time the metrics plugin will wait
+// before erroring when it tries to close the metrics server
+var shutdownTimeout time.Duration = time.Second * 1
```
fturib: I do not like too much a global var.
I guess we can do a constant for now - is 1 sec enough ?

plugin/metrics/metrics.go
```diff
@@ -94,9 +97,9 @@ func (m *Metrics) OnStartup() error {
 
 	m.mux = http.NewServeMux()
 	m.mux.Handle("/metrics", promhttp.HandlerFor(m.Reg, promhttp.HandlerOpts{}))
-
+	m.srv = &http.Server{Handler: m.mux}
```
fturib, line 100: For info : declaring http.Server allow to use Shutdown instead of Close.
and that is mandatory for the UT to work properly.

Good to have the timeout on this shutdown !

fturib (changes requested):
See comments inline.
Thanks for contribution !!

plugin/pkg/uniq/uniq.go
ZachEddy: Okay! I'll revert `ForEach` back.

plugin/metrics/metrics.go
ZachEddy: > I do not like too much a global var.

Yeah, a constant would be better! I'll update the pull request accordingly.

> is 1 sec enough ?

I chose one second arbitrarily. Since this runs on reload, I figured that a shorter timeout would be better because it could (maybe?) suspend the entire reload process (i.e. a twenty second timeout might block the reload for twenty seconds).

I don't fully understand the reload plugin works and am making some assumptions. Would do you think?

plugin/metrics/metrics.go
ZachEddy: Okay cool. I initially tried without the `http.Server` field and saw that it didn't fix the issue, so I figured it had something to do with that 🙂

plugin/metrics/metrics.go
fturib: we expect this timeout is reached only once a while when the server cannot finish its work normally. This server is only about sending data, most of the time fewer than one sec should be enough.

but for extreme case I would set to 5 secs.
Again, we expect the timeout is not used ... and the stop of server happen within the sec.

fturib (approved):

grobie (approved):
LGTM

fturib:
@ZachEddy - congrat for your first contribution !

PR #2037, opened 2018-08-19
plugin/forward: Add comment about external dependency

eugen-kleiner-mesh:
Add comment about external dependency separated setup function not to be refactored in future.
No code changes.
https://github.com/coredns/fallback/issues/9

### DIFF
--- a/plugin/forward/setup.go
+++ b/plugin/forward/setup.go
@@ -90,6 +90,7 @@ func parseForward(c *caddy.Controller) (*Forward, error) {
 }
 
 // ParseForwardStanza parses one forward stanza
+// Don't remove ParseForwardStanza because it is used by external plugin alternate
 func ParseForwardStanza(c *caddyfile.Dispenser) (*Forward, error) {
 	f := New()
 


miekg:
this previous PR that allowed for inter plugin dependency should not have been merged. Each plugin can do whatever it wants, external dependencies can't exist here. If things are important enough they should move the `plugin/pkg` and even then it is up to discussion if you want to hook into some else's setup function.

miekg:
For *forward* the setup was also move into pkg that still causes headaches because we can't easily deprecate that code, and that's even *in tree* code.

If you need this dep somehow, you probably just need to copy the code.

fturib:
The same way Fallback was depending of Proxy (setup and handler), Alternate is completely dependent of Forward.

Instead of moving the code or duplicate the whole code, would it not be more simple to just declare by documentation Alternate Syntax and Behavior completely dependent of Forward. Like a piggy-back plugin on Forward.

It would mean for end-users that any change of syntax or behavior for Forward will directly impact Alternate.

For the direct dependency of code, we can setup Travis on Alternate repo to check weekly that it is still compilable and run as expected. And we would be warn if a major change in Fallback break Alternate. (like removing this ParseForwardStanza function ...

fturib:
NOTE: I think we have similar dependency with Kubernetai/Kubernetes and Auto/File ... maybe others ..

miekg:
>NOTE: I think we have similar dependency with Kubernetai/Kubernetes and Auto/File ... maybe others ..

that's technical not a concern for coredns. You have an external plugin, so it 
would be strange that we (=coredns) are now changing code to accomedate you.

rdrozhdzh:
One of possible solutions is to copy/mimic parsing code of forward plugin into external plugin and add unit test to make sure that parsing code of forward plugin and external plugin give the same or compatible results.
I used similar approach in #1280 where I mimicked the functionality of framestream Encoder to increase performance and in the same time have compatible output

fturib:
> that's technical not a concern for coredns. You have an external plugin, so it
> would be strange that we (=coredns) are now changing code to accomedate you.

I am kind of perplex here ...
the whole point here is to reuse the code that is already debugged and is maintained ...

We are trying to have a way to reuse it, and sure the code needs then a minimal change to be reusable, as It is not yet.

1. Either opening the Setup : creation of the Forward object and later usage.
2. Either we copy the code of creation (I mean Setup part) - in that case we will need to add public function to Forward structure to allow init of fields. (maxFails, expire, from.. etc are not public).

Point 1 was the closest to what we want at the end ... and with the minimal change in Forward.
but if needed, we can do Point 2.

PR #2042, opened 2018-08-21
plugin/kubernetes: handle blank name and namespaces

### LINKED ISSUE #2028: plugin/kubernetes: Bare namespace query returns all service records in namespace
If you query a name in the form `<namespace>.<type>.<cluster-domain>.` ... e.g. `default.svc.cluster.local.`, all services in the namespace are returned.  The result is the same as if `*.default.svc.cluster.local.` was queried (in which case the result is expected), which implies this is a "query wildcard" handling mishap.

```
dnstools# dig  test-1.svc.cluster.local.

; <<>> DiG 9.11.3 <<>> test-1.svc.cluster.local.
;; global options: +cmd
;; Got answer:
;; WARNING: .local is reserved for Multicast DNS
;; You are currently testing what happens when an mDNS query is leaked to DNS
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 64388
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 14, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; COOKIE: bd1f5e3b1b7bcd0d (echoed)
;; QUESTION SECTION:
;test-1.svc.cluster.local.	IN	A

;; ANSWER SECTION:
test-1.svc.cluster.local. 5	IN	A	172.17.0.254
test-1.svc.cluster.local. 5	IN	A	172.17.0.255
test-1.svc.cluster.local. 5	IN	A	10.0.0.100
test-1.svc.cluster.local. 5	IN	A	10.0.0.115
test-1.svc.cluster.local. 5	IN	A	10.0.0.241
test-1.svc.cluster.local. 5	IN	A	10.0.0.110

;; Query time: 629 msec
;; SERVER: 10.0.0.10#53(10.0.0.10)
;; WHEN: Fri Aug 10 17:35:43 UTC 2018
;; MSG SIZE  rcvd: 704
```

Expected result should be a `NODATA` response, confirming the existence of records below, but not listing them all.

chrisohaver:
Fixes cases where missing names or namespaces in a query cause k8s plugin to return all services/pods.

Fixes #2028

none

### DIFF
--- a/plugin/kubernetes/handler_pod_insecure_test.go
+++ b/plugin/kubernetes/handler_pod_insecure_test.go
@@ -55,7 +55,7 @@ var podModeInsecureCases = []test.Case{
 	},
 	{
 		Qname: "podns.pod.cluster.local.", Qtype: dns.TypeA,
-		Rcode: dns.RcodeNameError,
+		Rcode: dns.RcodeSuccess,
 		Ns: []dns.RR{
 			test.SOA("cluster.local.	300	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1515173576 7200 1800 86400 30"),
 		},
--- a/plugin/kubernetes/handler_test.go
+++ b/plugin/kubernetes/handler_test.go
@@ -290,6 +290,13 @@ var dnsTestCases = []test.Case{
 			test.SOA("cluster.local.	303	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1499347823 7200 1800 86400 60"),
 		},
 	},
+	{
+		Qname: "testns.svc.cluster.local.", Qtype: dns.TypeA,
+		Rcode: dns.RcodeSuccess,
+		Ns: []dns.RR{
+			test.SOA("cluster.local.	303	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1499347823 7200 1800 86400 60"),
+		},
+	},
 }
 
 func TestServeDNS(t *testing.T) {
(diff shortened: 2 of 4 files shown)


miekg:
Wasn't this a feature we actually liked? (there was some discussion on twitter iirc).

That test case is def. correct in this PR though. But I'm left wondering why this is handled correctly in the current loop

chrisohaver:
> Wasn't this a feature we actually liked? (there was some discussion on twitter iirc).

I dont recall the discussion - but I know it happened...  @fturib brought the discussion to my attention when I brought this issue up recently.

**As a useful feature**: If a user wants to see all services in a namespace, they can use a wildcard (e.g. `*.my-namespace.svc.cluster.local.`).   So, it's not like this behavior provides a new feature.

**As general DNS protocol**: IIUC, the expected behavior when querying any subdomain, is to reply only with records that have that exact name; else `NODATA` if it is a subdomain of other records; else `NXDOMAIN`.

> But I'm left wondering why this is handled correctly in the current loop

I don't know for certain - but I suspect it is something to do with the way we mock the API connection with `APIConnServeTest`.

fturib:
The link to twitter is here : https://twitter.com/mauilion/status/1025160250880454656
I would think it is better to reply on this link to use explicitly the wildcard '*' to ask for a 'search' on services.

I guess I can reply on this tweet as soon as this PR is merged.

chrisohaver:
In https://twitter.com/mauilion/status/1025160250880454656, the _wildcard_ feature is touted... _not bare namespaces_ ...  the example given is `any.default.svc.cluster.local.` ... which is a wildcard query using the `any` reserved keyword.

He is using an `SRV` query which does support (per convention), a missing service/protocol prefix as "any service/protocol".

This PR doesn't change that behavior.

miekg:
fair enough.

Issue #1981, opened 2018-07-16
plugin/rewrite: set arbitrary TTL value


greenpau:
`consul` returns entries with TTL 0, i.e. no cache. Looking to rewrite the TTL to a value higher than 0. Currently, `plugin/rewrite` does not support it (or maybe it does).

Prior to modifying the `plugin/rewrite`, qq: does anyone have a way of modifying TTL value with another plugin?

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin/rewrite: s..." ]
>`consul` returns entries with TTL 0, i.e. no cache. Looking to rewrite the TTL to a value higher than 0. Currently, `plugin/rewrite` does not support it (or maybe it does).

but this is on the way back, i.e. replies from some upstream, so rewrite is of 
not use.

>Prior to modifying the `plugin/rewrite`, qq: does anyone have a way of modifying TTL value with another plugin?

you need a tiny new plugin that wraps the ResponseWriter and changes the TTL 
when it writes to the client. TTL code can copied/used from pkg/dnsutil/ttl.go

/Miek

--
Miek Gieben

greenpau:
> but this is on the way back, i.e. replies from some upstream, so rewrite is of not use.

@miekg,  e.g., currently, `plugin/rewrite` matches on name and then changes name in the answer:

```
    rewrite stop {
        name regex (.*)-(us-west-1)\.coredns\.rocks {1}.service.{2}.consul
        answer name (.*)\.service\.(us-west-1)\.consul {1}-{2}.coredns.rocks
    }
```

When it comes to TTL, I think this could be implemented [here](https://github.com/coredns/coredns/blob/master/plugin/rewrite/reverter.go#L37). 🤔 💭

```
    rewrite continue {
        name regex (.*)-(us-west-1)\.coredns\.rocks {1}.service.{2}.consul
        answer ttl 15
    }
```

@miekg, what do you think about this? 

> you need a tiny new plugin that wraps the ResponseWriter and changes the TTL

@miekg , ack. ... more 🤔

miekg:
Oh yes. You can do some noop rewrite and then adjust the ttl on the way
back. Should work as well.

PR #2048, opened 2018-08-22
plugin/rewrite: add handling of TTL field rewrites

greenpau:
Resolves: #1981

This PR allows the rewriting of TTL field in DNS answers.

#1981 

Add a section for TTL rewrites in the plugin's README.md

### DIFF
--- a/plugin/rewrite/README.md
+++ b/plugin/rewrite/README.md
@@ -13,7 +13,7 @@ Rewrites are invisible to the client. There are simple rewrites (fast) and compl
 
 A simplified/easy to digest syntax for *rewrite* is...
 ~~~
-rewrite [continue|stop] FIELD FROM TO
+rewrite [continue|stop] FIELD [FROM TO|FROM TTL]
 ~~~
 
 * **FIELD** indicates what part of the request/response is being re-written.
@@ -25,9 +25,11 @@ e.g., to rewrite ANY queries to HINFO, use `rewrite type ANY HINFO`.
      name, e.g., `rewrite name example.net example.org`. Other match types are supported, see the **Name Field Rewrites** section below.
    * `answer name` - the query name in the _response_ is rewritten.  This option has special restrictions and requirements, in particular it must always combined with a `name` rewrite.  See below in the **Response Rewrites** section.
    *  `edns0` - an EDNS0 option can be appended to the request as described below in the **EDNS0 Options** section.
+   * `ttl` - the TTL value in the _response_ is rewritten.
 
-* **FROM** is the name or type to match
+* **FROM** is the name (exact, suffix, prefix, substring, or regex) or type to match
 * **TO** is the destination name or type to rewrite to
+* **TTL** is the number of seconds to set the TTL value to
 
 If you specify multiple rules and an incoming query matches on multiple rules, the rewrite
 will behave as following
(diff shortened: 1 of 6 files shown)


greenpau:
@chrisohaver , I noticed Travis integration test did run. Is this the expected behavior for WIP PRs?

chrisohaver:
Yeah - tests still run on WIPs... it's free/cheap.... and useful depending on which stage of developing the PR you are in.

greenpau:
@chrisohaver , @makinje16 , please review if you can.

greenpau:
@johnbelamaric , congrats on your new position 🚀 👍 . I reassigned reviewers being conscious of you might be busy with on-boarding. Nevertheless, if you can review this one, please do 😄

chrisohaver (commented):
Looks good.  One concern I have though is that this new TTL rewrite rule is, essentially, an *answer* rewrite (it's re-writing TTLs in the response)...  but its syntax follows *query* rewrite syntax.

greenpau:
> One concern I have though is that this new TTL rewrite rule is, essentially, an answer rewrite (it's re-writing TTLs in the response)... but its syntax follows query rewrite syntax.

@chrisohaver , you are correct. However, there is not TTL in the request. That's why I am thinking making it `rewrite ttl` is less confusing than `rewrite answer ttl`. What is the syntax you propose?

@miekg , what is your thought on the above?

Whatever you guys decide is fine by me. I will implement any syntax, as long as there is a consensus 😃

miekg:
>> One concern I have though is that this new TTL rewrite rule is, essentially, an answer rewrite (it's re-writing TTLs in the response)... but its syntax follows query rewrite syntax.
>
>@chrisohaver , you are correct. However, there is not TTL in the request. That's why I am thinking making it `rewrite ttl` is less confusing than `rewrite answer ttl`. What is the syntax you propose?
>
>@miekg , what is your thought on the above?
>
>Whatever you guys decide is fine by me. I will implement any syntax, as long as there is a consensus 😃

Haven't followed this PR in detail. I care about simplicity and consistency (for 
the user). So if it is not confusing it's fine by me; if it needs (a lot) of 
extra docs so explain, we should probably revisit.

greenpau:
@chrisohaver , the current syntax is simple enough, no confusion. Let’s go with it?

chrisohaver:
The existing answer re-write is a bit unusual anyways ... Given that, maybe its a bad idea to try to be consistent with it.  So, your initial syntax may be best.

To be consistent with the existing answer name rewrite, the syntax would be something like...

```
rewrite [continue|stop] name regex STRING answer ttl TTL
```
or in the pseudo 2 rule format ...
```
rewrite [continue|stop] {
    name regex STRING 
    answer ttl TTL
}
```

perhaps adding a `match` keyword would make it less ambiguous to parse...

```
rewrite [continue|stop] match name regex STRING answer ttl TTL
```
or in the pseudo 2 rule format ...
```
rewrite [continue|stop] {
    match name regex STRING 
    answer ttl TTL
}
```
Whichever way, it should also be documented in the **Syntax** section, and not just in a separate section added to the end.

chrisohaver:
IMO, We should probably re-think the syntax on answer name re-writes...

The only use case is to restore names to original after a qname re-write.  In that sense, there should be no need to specify the query name string 4 times.  Two should be sufficient, e.g. ...

```
rewrite [continue|stop] name regex STRING STRING answer name restore
```

Or - we just always make it restore the name on a name re-writes by default.
Then we can deprecate, and eventually remove the answer name rewrite syntax.

greenpau:
@chrisohaver , let’s keep this PR separate from the bigger thing. I understand what you are saying, but it should have its own issue and its own PR (linked to this one by reference).

What are you proposing to change in this one, lines and suggestions? I will follow. I cannot think in abstact here.

chrisohaver:
> let’s keep this PR separate from the bigger thing. I understand what you are saying, but it should have its own issue and its own PR (linked to this one by reference).

Obviously.  Yes, I agree.

greenpau:
@chrisohaver , like the “match name” proposal, but let’s cover it in a separate issue.

chrisohaver:
> What are you proposing to change in this one, lines and suggestions? I will follow. I cannot think in abstact here.

Please also document the new option in the Syntax section of the README (i.e. in the "FIELD" section, which starts at line [19](https://github.com/coredns/coredns/pull/2048/files#diff-e805684cc3e7aef334b3ac12d4a47319R19)).

greenpau:
> Please also document the new option in the Syntax section of the README (i.e. in the "FIELD" section, which starts at line 19).

@chrisohaver , done.

chrisohaver (approved):

PR #2050, opened 2018-08-22
[plugin/route53]: Support batch mode operation.

dilyevsky:
Cache all Route53 records internally using `ListResourceRecordPagesWithContext`
paging query and serve them from memory. Improves latency and avoids throttling by AWS under load.

Bonus features:

  * Support additional r53 record types (`CNAME`, `SOA`, `NS` etc)
  * Support `upstream` option

Signed-off-by: Dmitry Ilyevskiy <dmitry.ilyevskiy@getcruise.com>
Signed-off-by: Dmitry Ilyevskiy <ilyevsky@gmail.com>

### DIFF
--- a/plugin/proxy/dns.go
+++ b/plugin/proxy/dns.go
@@ -93,9 +93,7 @@ func exchange(m *dns.Msg, co net.Conn) (*dns.Msg, error) {
 
 	writeDeadline := time.Now().Add(defaultTimeout)
 	dnsco.SetWriteDeadline(writeDeadline)
-	if err := dnsco.WriteMsg(m); err != nil {
-		return nil, err
-	}
+	dnsco.WriteMsg(m)
 
 	readDeadline := time.Now().Add(defaultTimeout)
 	co.SetReadDeadline(readDeadline)
--- a/plugin/proxy/dns.go
+++ b/plugin/proxy/dns.go
@@ -93,7 +93,9 @@ func exchange(m *dns.Msg, co net.Conn) (*dns.Msg, error) {
 
 	writeDeadline := time.Now().Add(defaultTimeout)
 	dnsco.SetWriteDeadline(writeDeadline)
-	dnsco.WriteMsg(m)
+	if err := dnsco.WriteMsg(m); err != nil {
+		return nil, err
+	}
 
 	readDeadline := time.Now().Add(defaultTimeout)
 	co.SetReadDeadline(readDeadline)
--- a/plugin/route53/README.md
+++ b/plugin/route53/README.md
@@ -7,13 +7,15 @@
 ## Description
 
 The route53 plugin is useful for serving zones from resource record sets in AWS route53. This plugin
-only supports A and AAAA records. The route53 plugin can be used when coredns is deployed on AWS.
+supports all Amazon Route 53 records (https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/ResourceRecordTypes.html).
+The route53 plugin can be used when coredns is deployed on AWS or elsewhere.
 
 ## Syntax
 
 ~~~ txt
 route53 [ZONE:HOSTED_ZONE_ID...] {
     [aws_access_key AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY]
+    upstream [ADDRESS...]
 }
 ~~~
 
@@ -23,18 +25,23 @@ route53 [ZONE:HOSTED_ZONE_ID...] {
    to be used when query AWS (optional).  If they are not provided, then coredns tries to access
    AWS credentials the same way as AWS CLI, e.g., environmental variables, AWS credentials file,
    instance profile credentials, etc.
+* `upstream` [**ADDRESS**...] specifies upstream resolver(s) used for resolving services that point
+   to external hosts (eg. used to resolve CNAMEs). If no **ADDRESS** is given, CoreDNS will resolve
+   against itself. **ADDRESS** can be an IP, an IP:port or a path to a file structured like
+   resolv.conf (**NB**: Currently a bug is preventing the use of self-resolver).
 
 ## Examples
 
-Enable route53, with implicit aws credentials:
+Enable route53 with implicit aws credentials and an upstream:
 
 ~~~ txt
 . {
     route53 example.org.:Z1Z2Z3Z4DZ5Z6Z7
+    upstream 10.0.0.1
 }
 ~~~
 
-Enable route53, with explicit aws credentials:
+Enable route53 with explicit aws credentials:
 
 ~~~ txt
 . {
--- a/plugin/route53/README.md
+++ b/plugin/route53/README.md
@@ -28,7 +28,7 @@ route53 [ZONE:HOSTED_ZONE_ID...] {
 * `upstream` [**ADDRESS**...] specifies upstream resolver(s) used for resolving services that point
    to external hosts (eg. used to resolve CNAMEs). If no **ADDRESS** is given, CoreDNS will resolve
    against itself. **ADDRESS** can be an IP, an IP:port or a path to a file structured like
-   resolv.conf (**NB**: Currently a bug is preventing the use of self-resolver).
+   resolv.conf (**NB**: Currently a bug (#2099) is preventing the use of self-resolver).
 
 ## Examples
 
--- a/plugin/route53/route53.go
+++ b/plugin/route53/route53.go
@@ -37,12 +37,7 @@ type zone struct {
 }
 
 // New returns new *Route53.
-func New(
-	ctx context.Context,
-	c route53iface.Route53API,
-	keys map[string]string,
-	up *upstream.Upstream) (*Route53, error) {
-
+func New(ctx context.Context, c route53iface.Route53API, keys map[string]string, up *upstream.Upstream) (*Route53, error) {
 	zones := make(map[string]*zone, len(keys))
 	zoneNames := make([]string, 0, len(keys))
 	for dns, id := range keys {
--- a/plugin/route53/route53.go
+++ b/plugin/route53/route53.go
@@ -130,27 +130,6 @@ func (h *Route53) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	return dns.RcodeSuccess, nil
 }
 
-func awsStrTypeToUint16(t string) (uint16, bool) {
-	var out uint16
-	switch t {
-	case "A":
-		out = dns.TypeA
-	case "AAAA":
-		out = dns.TypeAAAA
-	case "CNAME":
-		out = dns.TypeCNAME
-	case "PTR":
-		out = dns.TypePTR
-	case "SOA":
-		out = dns.TypeSOA
-	case "NS":
-		out = dns.TypeNS
-	default:
-		return dns.TypeNone, false
-	}
-	return out, true
-}
-
 func parseSOA(soa string) (ns, mbox string, serial, refresh, retry, expire, minttl int, err error) {
 	parts := strings.SplitN(soa, " ", 7)
 	if len(parts) != 7 {
@@ -217,7 +196,7 @@ func setRRValue(rr dns.RR, hdr *dns.RR_Header, value string) error {
 }
 
 func updateZoneFromRRS(rrs *route53.ResourceRecordSet, z *file.Zone) error {
-	t, ok := awsStrTypeToUint16(aws.StringValue(rrs.Type))
+	t, ok := dns.StringToType[aws.StringValue(rrs.Type)]
 	if !ok {
 		return fmt.Errorf("unsupported record type: %s", aws.StringValue(rrs.Type))
 	}
--- a/plugin/route53/route53.go
+++ b/plugin/route53/route53.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/file"
-	"github.com/coredns/coredns/plugin/pkg/log"
 	"github.com/coredns/coredns/plugin/pkg/upstream"
 	"github.com/coredns/coredns/request"
 
--- a/plugin/route53/route53.go
+++ b/plugin/route53/route53.go
@@ -5,10 +5,6 @@ package route53
 import (
 	"context"
 	"fmt"
-	"net"
-	"reflect"
-	"strconv"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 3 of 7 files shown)


plugin/proxy/dns.go
```diff
@@ -93,7 +93,9 @@ func exchange(m *dns.Msg, co net.Conn) (*dns.Msg, error) {
 
 	writeDeadline := time.Now().Add(defaultTimeout)
 	dnsco.SetWriteDeadline(writeDeadline)
-	dnsco.WriteMsg(m)
+	if err := dnsco.WriteMsg(m); err != nil {
```
miekg: this looks like an unrelated change

plugin/route53/route53.go
```diff
@@ -15,98 +24,275 @@ import (
...
+	id string
+	z  *file.Zone
+}
 
-	zone := plugin.Zones(rr.zones).Matches(qname)
-	if zone == "" {
-		return plugin.NextOrFailure(rr.Name(), rr.Next, ctx, w, r)
-	}
+// NewRoute53 returns new *Route53.
+func NewRoute53(
```
miekg: Can just be `New`

plugin/route53/route53.go
```diff
@@ -15,98 +24,275 @@ import (
...
-	switch state.QType() {
-	case dns.TypeA:
-		answers = a(qname, output.ResourceRecordSets)
-	case dns.TypeAAAA:
-		answers = aaaa(qname, output.ResourceRecordSets)
-	case dns.TypePTR:
-		answers = ptr(qname, output.ResourceRecordSets)
+// Run executes first update, spins up an update forever-loop.
+// Returns error if first update fails.
+func (h *Route53) Run(ctx context.Context) error {
```
miekg: Why `h` for the receiver's name?

plugin/route53/route53.go
```diff
@@ -15,98 +24,275 @@ import (
...
 
-func a(zone string, rrss []*route53.ResourceRecordSet) []dns.RR {
-	answers := []dns.RR{}
-	for _, rrs := range rrss {
-		for _, rr := range rrs.ResourceRecords {
-			r := new(dns.A)
-			r.Hdr = dns.RR_Header{Name: zone, Rrtype: dns.TypeA, Class: dns.ClassINET, Ttl: uint32(aws.Int64Value(rrs.TTL))}
-			r.A = net.ParseIP(aws.StringValue(rr.Value)).To4()
-			answers = append(answers, r)
+func awsStrTypeToUint16(t string) (uint16, bool) {
```
miekg: this can be replaced with `dns.TypeToString[]` lookups. Prolly still need a switch to filter on the knows records.

plugin/route53/route53.go
```diff
@@ -15,98 +24,275 @@ import (
...
+		out = dns.TypeSOA
+	case "NS":
+		out = dns.TypeNS
+	default:
+		return dns.TypeNone, false
+	}
+	return out, true
+}
+
+func parseSOA(soa string) (ns, mbox string, serial, refresh, retry, expire, minttl int, err error) {
```
miekg: you can just call `dns.NewRR1` with SOA string, and then access the members in the SOA struct.

plugin/route53/route53.go
```diff
@@ -15,98 +24,275 @@ import (
...
+	if expire, err = strconv.Atoi(parts[5]); err != nil {
+		return
+	}
+	if minttl, err = strconv.Atoi(parts[6]); err != nil {
+		return
+	}
+	return
+}
+
+func setRRValue(rr dns.RR, hdr *dns.RR_Header, value string) error {
```
miekg: I'm not sure, but I feel there should be a better way to do this as well...

miekg (changes requested):
These are quite substantial changes, so I would like to volunteer you for the OWNERS file here as well?

plugin/proxy/dns.go
dilyevsky: Yes, I realized the error was not propagated up the stack when I was testing this thing. Would you like me to send a separate PR for this?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [plugin/route..." ]
>dilyevsky commented on this pull request.
>
>
>
>> @@ -93,7 +93,9 @@ func exchange(m *dns.Msg, co net.Conn) (*dns.Msg, error) {
>
> 	writeDeadline := time.Now().Add(defaultTimeout)
> 	dnsco.SetWriteDeadline(writeDeadline)
>-	dnsco.WriteMsg(m)
>+	if err := dnsco.WriteMsg(m); err != nil {
>
>Yes, I realized the error was not propagated up the stack when I was testing this thing. Would you like me to send a separate PR for this?

We can't care about this error, because there is nothing we can do.

plugin/route53/route53.go
dilyevsky: Done.

plugin/route53/route53.go
dilyevsky: I think it's somewhat customary in Go to have this be shorthand for the entity (interface) being implemented (`Handler` in this case). Kind of like so: https://golang.org/src/net/udpsock.go

Anyways, I don't feel strongly about it.

plugin/route53/route53.go
dilyevsky: Actually this is supposed to be the reverse but I search a little more and found `StringToType`, thanks. Done.

plugin/route53/route53.go
dilyevsky: Awesome - I was able to get rid of all parsing code and just use `dns.NewRR` parser.

plugin/route53/route53.go
dilyevsky: ditto. done.

dilyevsky (commented):
> These are quite substantial changes, so I would like to volunteer you for the OWNERS file here as well?

Will be happy to. Do I just add myself into OWNERS file?

dilyevsky:
> We can't care about this error, because there is nothing we can do.

I mean, just having it printed out would've saved me an hour of debugging, so it's not completely unhandled.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [plugin/route..." ]
>Will be happy to. Do I just add myself into OWNERS file?

Ack. In a separate PR please. Sent you an invite to be a collaborator.

Thanks!

plugin/proxy/dns.go
miekg: yep, please do. Should probably also be `log.Debug`

plugin/route53/route53.go
```diff
@@ -15,98 +20,180 @@ import (
...
 
-	zone := plugin.Zones(rr.zones).Matches(qname)
-	if zone == "" {
-		return plugin.NextOrFailure(rr.Name(), rr.Next, ctx, w, r)
+// New returns new *Route53.
+func New(
+	ctx context.Context,
+	c route53iface.Route53API,
+	keys map[string]string,
+	up *upstream.Upstream) (*Route53, error) {
```
miekg: please put these on one line (there is no line length limit in go)

plugin/route53/route53.go
```diff
@@ -15,98 +20,180 @@ import (
...
-			r.Hdr = dns.RR_Header{Name: zone, Rrtype: dns.TypeA, Class: dns.ClassINET, Ttl: uint32(aws.Int64Value(rrs.TTL))}
-			r.A = net.ParseIP(aws.StringValue(rr.Value)).To4()
-			answers = append(answers, r)
+func updateZoneFromRRS(rrs *route53.ResourceRecordSet, z *file.Zone) error {
+	for _, rr := range rrs.ResourceRecords {
+		// Assemble RFC 1035 conforming record to pass into dns scanner.
+		rfc1035 := fmt.Sprintf("%s %d IN %s %s", aws.StringValue(rrs.Name), aws.Int64Value(rrs.TTL), aws.StringValue(rrs.Type), aws.StringValue(rr.Value))
+		r, err := dns.NewRR(rfc1035)
+		if err != nil {
+			return fmt.Errorf("failed to parse resource record: %v", err)
```
miekg, line 129: %v -> %q

plugin/route53/route53.go
```diff
@@ -15,98 +20,180 @@ import (
...
 	}
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				log.Infof("breaking out of Route53 update loop: %v", ctx.Err())
+				return
+			case <-time.After(1 * time.Minute):
+				if err := h.updateZones(ctx); err != nil && ctx.Err() == nil /* Don't log error if ctx expired. */ {
+					log.Errorf("failed to update zones: %v", err)
```
miekg: We've settled an starting log lines with an uppercase `failed` -> `Failed` (same for the other log lines)

plugin/route53/route53.go
```diff
@@ -15,98 +20,180 @@ import (
...
+// Returns error if first update fails.
+func (h *Route53) Run(ctx context.Context) error {
+	if err := h.updateZones(ctx); err != nil {
+		return err
 	}
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				log.Infof("breaking out of Route53 update loop: %v", ctx.Err())
```
miekg: log can be deleted here.

plugin/route53/route53.go
miekg: Have you given some though if all these log files are really helpful or just spam the log output? I.e. would debug.Log(f) helpful?

plugin/route53/setup.go
```diff
@@ -57,32 +60,35 @@ func setup(c *caddy.Controller, f func(*credentials.Credentials) route53iface.Ro
 					return c.Errf("invalid access key '%v'", v)
 				}
 				credential = credentials.NewStaticCredentials(v[0], v[1], "")
+			case "upstream":
```
miekg, line 66: can you copy how the etcd plugin does this? i.e. just
~~~
case "upstream":
					args := c.RemainingArgs()
					if len(args) == 0 {
						return nil, false, c.ArgErr()
					}
					u, err := upstream.New(args)
					if err != nil {
						return nil, false, err
					}
					etc.Upstream = u
~~~
(but then of course adjusted to this plugin)

miekg (commented):
Probably the README.md also needs updating?

plugin/route53/route53.go
dilyevsky: I think it would be useful to know why the context was cancelled and that (also when) `Run` terminated. It should only be printed once.

plugin/route53/route53.go
dilyevsky: I know but I think this makes it more readable when there are many arguments passed in (alternatively they should be packaged into some sort of "options" struct but I don't think it's reached this point yet).

plugin/route53/route53.go
dilyevsky: Done.

plugin/route53/setup.go
dilyevsky: I think returning `ArgErr` is incorrect because the argument should be optional (at least as described in https://github.com/coredns/coredns/blob/master/plugin/etcd/README.md and https://github.com/coredns/coredns/blob/master/plugin/kubernetes/README.md). It's only because there is a bug in `Upstream` currently that I have to return error here (hence the TODO).

plugin/route53/route53.go
dilyevsky: > %v -> %q

Hm, these `.Error()` outputs are frequently changed together like so `error 3: error 2: error 1` so putting it quotation marks can result in multiple nested quotes...

> Have you given some though if all these log files are really helpful or just spam the log output? I.e. would debug.Log(f) helpful?

I believe currently even errors (this one is a warning) are not printed unless explicitly turned on in the config so it shouldn't be spammy. I think it is useful to log a warning when something that is genuinely unexpected happens (like in this case aws returned some unparseable records).

plugin/proxy/dns.go
dilyevsky: Ack, reverted this change here.

dilyevsky:
> Probably the README.md also needs updating?

Also done. PTAL

plugin/route53/README.md
```diff
@@ -23,18 +25,23 @@ route53 [ZONE:HOSTED_ZONE_ID...] {
    to be used when query AWS (optional).  If they are not provided, then coredns tries to access
    AWS credentials the same way as AWS CLI, e.g., environmental variables, AWS credentials file,
    instance profile credentials, etc.
+* `upstream` [**ADDRESS**...] specifies upstream resolver(s) used for resolving services that point
+   to external hosts (eg. used to resolve CNAMEs). If no **ADDRESS** is given, CoreDNS will resolve
+   against itself. **ADDRESS** can be an IP, an IP:port or a path to a file structured like
+   resolv.conf (**NB**: Currently a bug is preventing the use of self-resolver).
```
miekg: What bug? Consider adding a Bug section in the end if this really is a bug.

plugin/route53/route53.go
miekg: This is not the current style in CoreDNS (nor other Go code I've seen), please put this on one line. It's confusing because it looks like an struct literal

plugin/route53/route53.go
```diff
@@ -15,98 +20,180 @@ import (
...
 	}
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				log.Infof("Breaking out of Route53 update loop: %v", ctx.Err())
+				return
+			case <-time.After(1 * time.Minute):
+				if err := h.updateZones(ctx); err != nil && ctx.Err() == nil /* Don't log error if ctx expired. */ {
+					log.Errorf("Failed to update zones: %v", err)
```
miekg, line 76: If you create a `log` variable with `var log = clog.NewWithPlugin("route53")` where `clog` is:
`clog "github.com/coredns/coredns/plugin/pkg/log"` all these log will have **route53** prefixed automatically.

plugin/route53/setup.go
miekg: hmm, it is optional, that's a bug in etcd's setup, #2099 filed.

plugin/route53/README.md
dilyevsky: Sorry, I was going to file one but then forgot. Added reference to #2099 in the comment.

plugin/route53/route53.go
dilyevsky: Ah, thanks. I missed that it wasn't already initialized for this plugin. Done.

plugin/route53/route53.go
dilyevsky: Done.

dilyevsky:
Hey @miekg, thanks again for review! Looks like we have most of the comments (except the one related to #2099) handled. Are you ok with this going in now?

miekg:
yep. Go ahead and merge!

Thanks

dilyevsky:
great, thanks!

PR #2052, opened 2018-08-23
plugin/kubernetes: Create records for portless services

### LINKED ISSUE #2046: plugin/kubernetes: replicate recent hdls svc bug fix from kube-dns
This bug fix ... kubernetes/kubernetes#67622 ... needs to be replicated in the kubernetes plugin.
And unit tests added.

chrisohaver:

If a service does not have a port declared, k8s plugin doesn't create an A/AAAA record for it.  K8s documentation says that a port declaration is not needed, hence, this is technically a bug.  

This PR injects sentinel port values for services without an declared ports.
This allows us to create `A`/`AAAA` records for services without any declared ports.
When answering `SRV`, ignore those sentinel ports.

This was recently fixed in kube-dns (see #2046). And upon reviewing the change there, I initially thought this would add significant complexity to the code.  After exploring how it would work in coredns, it turned out to be less ugly than I feared it would be. 

fixes #2046

none

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -172,9 +172,9 @@ func SRV(b ServiceBackend, zone string, state request.Request, opt Options) (rec
 		w[serv.Priority] += weight
 	}
 	for _, serv := range services {
-		// Don't add the entry if the port is 0 (invalid). The kubernetes plugin uses port 0 when a service/endpoint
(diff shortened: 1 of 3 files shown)


miekg (approved):

miekg:
hmmm, conflicts :/
other lgtm.

Issue #1965, opened 2018-07-12
[kubernetes] Accept Kubeconfig File for Kubernetes Auth


Rigdon:
Currently the kubernetes plugin only accepts separate cert, key, and cacert files for authentication to remote Kubernetes clusters. Many other applications and services accept a valid Kubeconfig file for cross-cluster authentication and it would be very convenient to be able to authenticate the same way. This also helps solve the issue that currently it is necessary to issue long-lived certificates and tie them into RBAC policies in the remote Kubernetes cluster. This feature would allow users to create Service Accounts instead.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] [kubernetes] Acce..." ]
>and services accept a valid Kubeconfig file for cross-cluster authentication 
>and it would be very convenient to be able to authenticate the same way. This 

/label: plugin/kubernetes
/label: enhancement

chrisohaver:
I suppose the option would require specifying the context, as well as the path to the kubeconfig file. For example...

~~~
kubernetes cluster.local  {
    kubeconfig context path-to-kubeconfig
}
~~~

johnbelamaric:
#2053

PR #2053, opened 2018-08-24
Support for kubeconfig files

ZachEddy:

This pull request adds support for `kubeconfig` files to the Kubernetes plugin. If users need to connect to a remote Kubernetes cluster, they can specify the path to a `kubeconfig` file and the context they wish to use.

When the `kubeconfig` option is present, the Kubernetes plugin assigns any auth-related information to the Kubernetes REST client. That could be a token, a username and password, or TLS certificates. 

The `Corefile` ends up looking like this.

```
.:53 {
    kubernetes cluster.local {
        kubeconfig /path/to/kubeconfig contextname
    }
}
```

This feature was requested in #1965.

I added `README` documentation to the Kubernetes plugin.

### 4. General Notes

I made this a WIP because...

1. I haven't tried the `kubeconfig` option in a running Kubernetes cluster. Is there a good way to go about testing this?
2. As a temporary measure, I copied/pasted the `Config` struct from [this repository](https://github.com/ericchiang/k8s/blob/master/config.go). Can this be included as a dependency, or is it preferable to maintain separate `Config` struct inside CoreDNS?


johnbelamaric:
Hi Zach. You shouldn't need all that. We use https://github.com/kubernetes/client-go which should already support the use of the kubeconfig file. There is an example here: https://github.com/kubernetes/client-go/blob/master/examples/out-of-cluster-client-configuration/main.go#L45

chrisohaver:
> haven't tried the kubeconfig option in a running Kubernetes cluster. Is there a good way to go about testing this?

Do you mean from within a cluster?  You could create a configmap and mount it in the pod. Or add them as new values in the existing configmap md mount them in the pod. Unusual use case, But I could see that being useful in conjunction with the kubernetai plugin.

ZachEddy:
Thanks for pointing that out @johnbelamaric and sorry for the delay! I'm still not sure if I've gone about this in a sensible way.

The `clientcmd` package uses the current context inside the `kubeconfig` file unless overrides are provided, so I ended up following some code I found in [this](https://github.com/kubernetes/client-go/issues/192) GitHub issue. I wanted to mention that in case the code in `setup.go` isn't immediately obvious.

>Do you mean from within a cluster?

@chrisohaver That's not what I had in mind. I'm just wondering if I've tested my code sufficiently enough 👍

Issue #2054, opened 2018-08-26
plugin/cache: Add minimum ttl setting


aaronriekenberg:
Unbound has a cache-min-ttl setting, which is helpful to limit the number of data fetches from the backend.  From https://nlnetlabs.nl/documentation/unbound/unbound.conf/:

```  
              Time  to  live  minimum  for  RRsets  and messages in the cache.
              Default is 0.  If the minimum kicks in, the data is  cached  for
              longer than the domain owner intended, and thus less queries are
              made to look up the data.  Zero makes sure the data in the cache
              is  as the domain owner intended, higher values, especially more
              than an hour or so, can lead to trouble as the data in the cache
              does not match up with the actual data any more.
```

Seems like this would be quite easy to implement here: https://github.com/coredns/coredns/blob/master/plugin/cache/cache.go#L166

Have already gotten TTL from the message and set duration based on this TTL and the max TTL configured.  Just one more check to also clamp duration no lower than min TTL if this new option is set.  Looks like this would affect both the time the entry is stored in the cache and the TTL in the response - which is the same behavior as unbound.

Anyone see a reason not to add an additional min TTL parameter to cache?  Thinking of submitting a PR for this.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin/cache: Add..." ]
>Unbound has a cache-min-ttl setting, which is helpful to limit the number of data fetches from the backend.  From https://nlnetlabs.nl/documentation/unbound/unbound.conf/:
>
>```
>              Time  to  live  minimum  for  RRsets  and messages in the cache.
>              Default is 0.  If the minimum kicks in, the data is  cached  for
>              longer than the domain owner intended, and thus less queries are
>              made to look up the data.  Zero makes sure the data in the cache
>              is  as the domain owner intended, higher values, especially more
>              than an hour or so, can lead to trouble as the data in the cache
>              does not match up with the actual data any more.
>```
>
>Seems like this would be quite easy to implement here: https://github.com/coredns/coredns/blob/master/plugin/cache/cache.go#L166
>
>Have already gotten TTL from the message and set duration based on this TTL and the max TTL configured.  Just one more check to also clamp duration no lower than min TTL if this new option is set.  Looks like this would affect both the time the entry is stored in the cache and the TTL in the response - which is the same behavior as unbound.
>
>Anyone see a reason not to add an additional min TTL parameter to cache?  Thinking of submitting a PR for this.

[...]

PR #2055, opened 2018-08-26
Add MINTTL parameter to cache configuration.

aaronriekenberg:
Adds a MINTTL parameter to allow specifying a minimum ttl for success and denial responses that can override the ttl for the backend.  This can be useful for limiting queries to the backend.

https://github.com/coredns/coredns/issues/2054

PR updates cache/plugin/README.md

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -22,9 +22,10 @@ type Cache struct {
 	Next  plugin.Handler
 	Zones []string
 
-	ncache *cache.Cache
-	ncap   int
-	nttl   time.Duration
+	ncache  *cache.Cache
+	ncap    int
+	nttl    time.Duration
+	minnttl time.Duration
 
 	pcache  *cache.Cache
 	pcap    int
@@ -52,6 +53,7 @@ func New() *Cache {
 		ncap:       defaultCap,
 		ncache:     cache.New(defaultCap),
 		nttl:       maxNTTL,
+		minnttl:    minNTTL,
 		prefetch:   0,
 		duration:   1 * time.Minute,
 		percentage: 10,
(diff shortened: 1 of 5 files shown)


grobie (commented):
Why didn't you add the minttl option to negative caches as well?

chrisohaver:
> Why didn't you add the minttl option to negative caches as well?

This would be a good option.  Specifically as a means to mitigate k8s n-dots/search-path NXDOMAIN spamming issue.

aaronriekenberg:
Added minttl option to negative cache also.

plugin/cache/cache.go
```diff
@@ -100,6 +104,17 @@ func hash(qname string, qtype uint16, do bool) uint32 {
 	return h.Sum32()
 }
 
+func computeTTL(msgTTL time.Duration, minTTL time.Duration, maxTTL time.Duration) (ttl time.Duration) {
```
miekg: you don't need to repeat time.Duration for all these parameters

plugin/cache/cache.go
```diff
@@ -100,6 +104,17 @@ func hash(qname string, qtype uint16, do bool) uint32 {
...
+	ttl = msgTTL
+	if ttl < minTTL {
+		ttl = minTTL
+	}
+	if ttl > maxTTL {
+		ttl = maxTTL
+	}
+	return
```
miekg: no naked returns please, name the variable

plugin/cache/cache.go
```diff
@@ -228,7 +241,9 @@ func (w *ResponseWriter) Write(buf []byte) (int, error) {
 
 const (
 	maxTTL  = dnsutil.MaximumDefaulTTL
+	minTTL  = 0
 	maxNTTL = dnsutil.MaximumDefaulTTL / 2
+	minNTTL = 0
```
miekg: in pkg/dnsutil/ttl.go we say the minimum ttl coredns allows is 5s, which may not be (currently) true, but certainly clashes with these values.

plugin/cache/cache.go
aaronriekenberg: Fixed

plugin/cache/cache.go
aaronriekenberg: Fixed

plugin/cache/cache.go
aaronriekenberg: Set minTTL and minNTTL to dnsutil.MinimalDefaultTTL.

miekg (approved):

grobie (approved):

PR #2063, opened 2018-08-28
release: split out sha256 upload

miekg:
For the 1.2.1 release we didn't upload the sha256 files. I've added
an echo, but some testing suggests that the files are created, so the
problem lays somewhere else.

Add a new upload section that sets the content type correctly and POST
the files as text/plain.


yongtang:
Ah yes that's the sha256 files.

yongtang (approved):
LGTM

PR #2069, opened 2018-08-29
release: kill sum target

miekg:
Just create the sha256 inline so we know there are there.
Add a 'version' target in Makefile.release that prints the version so we
can double check that easily in the release script.


chrisohaver:
lgtm

PR #2070, opened 2018-08-29
plugin/rewrite: add closing dot for suffix rewrite rule

### LINKED ISSUE #1881: Rewrite Plugin: Suffix Matching Doesn't Work
I am using CoreDNS 1.1.3, latest at time of writing, on CentOS 7 amd64.

I have the following configuration:

```
. {
  rewrite name suffix naftuli.local c.naftuli-test.internal
  forward . 169.254.169.254:53
}
```

The expectation is that the following will happen:

```shell
$ dig +short @127.0.0.1 bastion.naftuli.local
10.0.0.4
```

However, I get no results with the configuration above. If I replace `suffix` with `substring`, the query works as expected.

It appears that the suffix matcher is broken for the rewrite plugin.

> As an aside, regular expressions do also work, but my exact use case is suffixes only:
> ```
> . {
>   rewrite name regex ([^.]+).naftuli\.local\b {1}.c.naftuli-test.internal
>   forward . 169.254.169.254:53
> }
> ```

greenpau:
plugin/rewrite: add closing dot for suffix rewrite rule

* add closing dot for suffix rewrite rule
* improve rule syntax checks

Resolves: #1881

Bug fixes. See above.

#1881, #2041 

N/A.


greenpau:
@chrisohaver , while working on this PR, as part of tightening up syntax checks for rewrite rules, I discovered this [PR](https://github.com/coredns/coredns/commit/5e06687ee5a9522bfc4faa9b8f2adf6103eb016d#diff-7833ff341041ee754d2333ef983675c4).

Do you mind if I will rewrite this test? The rule is not a `regex` match, but rather `exact` match.

```
    c = caddy.NewTestController("dns",
        `rewrite stop {
    name regex foo bar
    answer name bar foo
}`)
```

greenpau:
@chrisohaver , alternatively, I could create answer rewriting for `substring` match.

```
rewrite stop {
    name substring foo bar
    answer name bar foo
}
```

chrisohaver:
IMO a plain string is valid regex...  Kind of vacuously, but still valid.

greenpau:
> IMO a plain string is valid regex... Kind of vacuously, but still valid.

@chrisohaver , alright, point taken. What do you think about warning a user about it, i.e. `log.Warnf(...`?

chrisohaver:
Im not clear on what the negative implications are of having a plain string in a regex are ...  It the concern for performance?

greenpau:
> Im not clear on what the negative implications are of having a plain string in a regex are ... It the concern for performance?

@chrisohaver , when one knows what s/he is doing they can do whatever they want. However, an adoption of a product, e.g. CoreDNS, depends on how user/admin friendly it is (of course they can buy IB 😄  for that). For people who are not familar with RegEx, syntax and logic checks are important.

> It the concern for performance?

No concern. A person concerned about the performance will probable spend time understanding the implication of using`rewrite` rules.

chrisohaver:
OK, then I see no need for a warning for when a user uses a plain string in a regex rule.

greenpau:
> OK, then I see no need for a warning for when a user uses a plain string in a regex rule.

@chrisohaver, please review.

plugin/rewrite/name_test.go
```diff
@@ -31,3 +31,58 @@ func TestRewriteIllegalName(t *testing.T) {
...
+		next         string
+		args         []string
+		expectedFail bool
+	}{
+		{"stop", []string{"exact", "srv3.coredns.rocks", "srv4.coredns.rocks"}, false},
+		{"stop", []string{"srv1.coredns.rocks", "srv2.coredns.rocks"}, false},
+		{"stop", []string{"suffix", "coredns.rocks", "coredns.rocks."}, false},
+		{"stop", []string{"suffix", "coredns.rocks.", "coredns.rocks"}, false},
+		{"stop", []string{"suffix", "coredns.rocks.", "coredns.rocks."}, false},
+		{"stop", []string{"regex", "srv1.coredns.rocks", "10"}, false},
```
chrisohaver, line 46: This test, and the following one replace a name with "10" ... seems unusual.  Does this test a special case/part of code?  Was it copied from a TTL test case?

plugin/rewrite/README.md
```diff
@@ -179,6 +179,16 @@ follows:
 rewrite [continue|stop] name regex STRING STRING answer name STRING STRING
 ```
 
+When using `exact` name rewrite rules, answer gets re-written automatically,
```
chrisohaver, line 182: IMO, we should also do this for regex matches.

plugin/rewrite/name.go
```diff
@@ -115,76 +116,141 @@ func (rule *regexNameRule) Rewrite(ctx context.Context, state request.Request) R
...
-			return &nameRule{nextAction, plugin.Name(args[1]).Normalize(), plugin.Name(args[2]).Normalize()}, nil
+			rewriteAnswerFromPattern, err := isValidRegexPattern(rewriteQuestionTo, rewriteQuestionFrom)
+			if err != nil {
+				return nil, err
+			}
+			return &exactNameRule{
+				nextAction,
+				rewriteQuestionFrom,
+				rewriteQuestionTo,
+				ResponseRule{
```
chrisohaver, line 162: I think if we check at client write time, that we can handle all cases (including regex), and we don't need to set up the response re-write rule at setup time (which only works for exact match).

chrisohaver (commented):
I think we should extend the "automatic" response question fix to all rewrites, by checking it at client write time.  What do you think?

plugin/rewrite/name_test.go
greenpau: > Does this test a special case/part of code?

@chrisohaver, No.

> Was it copied from a TTL test case?

No.

I think I just wanted to check for digits only rewrite.

plugin/rewrite/README.md
greenpau: > IMO, we should also do this for regex matches.

@chrisohaver , sure. It is a bit trickier.  If there is "answer", then don't do it automatically?

plugin/rewrite/name.go
greenpau: > I think if we check at client write time

@chrisohaver , what impact on the performance do you foresee?

chrisohaver: I dont foresee a significant impact on performance.

miekg:
is this ready to go in?

greenpau:
> is this ready to go in?

@miekg, by and large, yes. The suggestions by @chrisohaver (i.e. "I think we should extend the "automatic" response question fix to all rewrites, by checking it at client write time.") could be implemented in a separate PR.

chrisohaver:
> could be implemented in a separate PR 

I just opened #2202.

greenpau:
> I just opened ...

@chrisohaver , please complete the review.

chrisohaver:
@greenpau, 
I see that the fix for #1881 is to allow users to not specify a trailing dot in domains.
Can you describe specifically in what way this partially resolves #2041?

greenpau:
@chrisohaver, it does not partially resolve it. I will ipdate the commit message.

greenpau:
>  Can you describe specifically in what way this partially resolves #2041?

@chrisohaver , updated commit message.

chrisohaver (approved):

miekg:
Can someone write a small blurp for the 1.2.5 coredns release in the
coredns 125 release notes on the wiki (short oneliner will do).

Thanks!

chrisohaver:
> Can someone write a small blurb ...

done

Issue #2072, opened 2018-08-30
plugin/cache - What if 2 DNS requests are hashed to the same key ?


fturib:
## Description
The KEY for cache-entry is an int32 that is computed by the function `hash(qname string, qtype uint16, do bool) uint32`.

As this function is a Hash -and not a Identifier-, it is possible to get collisions : two different set of values are hashed into the same KEY.

In that case these 2 initial different values will be saved in the SAME cache entry - the last one will recover the first one.

### Issue
At GET time, the cache does not make the difference between these 2 set of values, and will return always the DNS response of the last DNS query set with this KEY.

NOTE: This issue is unlikely to happen, but can have severe important impact as CoreDNS will reply a wrong valid answer that is not related to the DNS query.

### Propositions to fix

Split the key in 2 parts:
1. a hash part between 0 to 255  that will identify surely the shard
1. a string key that completely identify the DNS query items : concat(qname, qtype, do, '|')

Then:
- select the shard with the hash part
- enter the item in the shard's map with the key that is an identifier.
There should not change the perf as the Map anyway will re-hash the key (whatever is int or string).

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin/cache - Wh..." ]
>The KEY for cache-entry is an int32 that is computed by the function `hash(qname string, qtype uint16, do bool) uint32`.

We could move to a hash.Hash64, and use fnv.NewHash64() which would be 3/4 line 
change

eugen-kleiner-mesh:
> We could move to a hash.Hash64, and use fnv.NewHash64() which would be 3/4 line change

I think, making bigger hash will decrease the possibility of collision, but will not fully solve the issue. You can't code uniquely 255 bytes (domain length) into 4 or 8 bytes and guarantee that hash will not intersect.

fturib:
@ekleiner : there is a check in the cache that validate the data retrieved is the reply of the same query. because of possible collision that data may not be the right one, but we know it and drop the query to avoid reply an inappropriate response

At the end, the collision has an effect only on efficiency of the cache.

Moving to 64bit will decrease the frequency of collision (to an average of 1 per 100 year for a default cache and 10kQPS).

And we keep the efficiency of a cache with a key based on an integer instead of string or complex structure which would trigger garbage collector ...

miekg:
Fixed via #2077

PR #2074, opened 2018-08-31
pkg/Cache adapted to accept string keys

fturib:
- change pkg/cache to accept string key
- compute the shard value of the entry based on an hash on that string key
- change key computation for Cache plugin
- adapt key computation for DNSSEC plugin

**Still some questions :**
1. for CACHE, should we compute the key based on the whole DNS query msg (including edns0 values) ?
2- I update the key computation of DNSSEC based now on full RR values. Need to be validated

#2072 

I do not think need another documentation.

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -64,7 +64,7 @@ func uint16ToWire(data uint16) string {
 }
 
 // hash serializes the RRset and return a signature cache key.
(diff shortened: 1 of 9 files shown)


miekg:
What is the memory (+gc) implication of this change. How does it effect 
benchmarks?

Issue #2079, opened 2018-09-03
Cache metrics not published after upgrade to v1.2.2


aleerizw-zz:
After upgrading to version v1.2.2 the cache plugin metrics (e.g coredns_cache_misses_total)  have stopped showing up. 

### Configuration 
```
    .:53 {
        errors
        health
        kubernetes cluster.local. in-addr.arpa ip6.arpa {
          pods insecure
          upstream
          fallthrough in-addr.arpa ip6.arpa
        }
        prometheus :9153
        loop
        proxy . /etc/resolv.conf
        cache 30
        loadbalance
        reload
    }
```

### Output of /metrics
```
# HELP coredns_build_info A metric with a constant '1' value labeled by version, revision, and goversion from which CoreDNS was built.
# TYPE coredns_build_info gauge
coredns_build_info{goversion="go1.11",revision="eb51e8b",version="1.2.2"} 1
# HELP coredns_dns_request_count_total Counter of DNS requests made per zone, protocol and family.
# TYPE coredns_dns_request_count_total counter
coredns_dns_request_count_total{family="1",proto="udp",server="dns://:53",zone="."} 516107
# HELP coredns_dns_request_duration_seconds Histogram of the time (in seconds) each request took.
# TYPE coredns_dns_request_duration_seconds histogram
coredns_dns_request_duration_seconds_bucket{server="dns://:53",zone=".",le="0.00025"} 437948
coredns_dns_request_duration_seconds_bucket{server="dns://:53",zone=".",le="0.0005"} 440592
coredns_dns_request_duration_seconds_bucket{server="dns://:53",zone=".",le="0.001"} 445773
coredns_dns_request_duration_seconds_bucket{server="dns://:53",zone=".",le="0.002"} 497477
coredns_dns_request_duration_seconds_bucket{server="dns://:53",zone=".",le="0.004"} 502192
coredns_dns_request_duration_seconds_bucket{server="dns://:53",zone=".",le="0.008"} 502938
coredns_dns_request_duration_seconds_bucket{server="dns://:53",zone=".",le="0.016"} 508867
coredns_dns_request_duration_seconds_bucket{server="dns://:53",zone=".",le="0.032"} 513420
coredns_dns_request_duration_seconds_bucket{server="dns://:53",zone=".",le="0.064"} 515604
coredns_dns_request_duration_seconds_bucket{server="dns://:53",zone=".",le="0.128"} 515914
[...]

PR #2080, opened 2018-09-03
Ensure Re-register of metrics variables after a reload

fturib:

Since the fix of reload issue in Metrics (issue #1982), the metrics plugin create a prom.Reg on each reload.
Therefore each plugin that register metrics variables **MUST** re-register after reload.

Each plugin registering metrics variables was limited to register only once, using a once.Do function.
- At first I proposed to re-trigger that once.Do function on each reload
- Then I found this limitation was useless : by Caddy architecture, each plugin will have its setup called only once per stanza, and the registering happen against the "prometheus" plugin of the SAME stanza. **There cannot be several calls**.
=> Removing the once.Do(..), allow each plugin to re-register after each Reload.

### Added a UT for Metrics / Forward / Cache
Other plugins : DNSSEC, Autopath, Health, Proxy, Template ..use the same mechanism.

#2079 

No change of documentation this is a fix.

### DIFF
--- a/test/reload_test.go
+++ b/test/reload_test.go
@@ -234,15 +234,8 @@ func TestMetricsAvailableAfterReload(t *testing.T) {
 	if _, _, err := cl.Exchange(m, tcp); err != nil {
 		t.Fatalf("Could not send message: %s", err)
 	}
-	if _, _, err := cl.Exchange(m, tcp); err != nil {
-		t.Fatalf("Could not send message: %s", err)
-	}
-	if _, _, err := cl.Exchange(m, tcp); err != nil {
-		t.Fatalf("Could not send message: %s", err)
-	}
 
 	// we should have metrics from forward, cache, and metrics itself
-	time.Sleep(500 * time.Millisecond)
 	if err := collectMetricsInfo(promAddress, procMetric, procCache, procForward); err != nil {
 		t.Errorf("Could not scrap one of expected stats : %s", err)
 	}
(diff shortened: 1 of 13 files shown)


johnbelamaric:
As long as we're sure there are no code paths that lead to OnStartup) being called multiple times I think this is good. What happens on a failed reload? Or a series of failed reloads?

fturib:
Yes, onStartup is called only once after creating the new Instance and called Setup on eahc of the plugin created for this new instance.

NOTE: when reload fails, the new instance is just dropped. And onStartup is not called again on the old instance. Right now Eugen is trying to have a new event OnFailedReload to be added in Caddy so we can, in some cases do something after a failed reload.

test/reload_test.go
```diff
@@ -202,4 +204,132 @@ func TestReloadSeveralTimeMetrics(t *testing.T) {
 	}
 }
 
+func TestMetricsAvailableAfterReload(t *testing.T) {
+	//TODO: add a tool that find an available port because this needs to be a port
```
johnbelamaric, line 208: If you use port 0, it should do that for you shouldn't it?

fturib: Yes, I can .. and it is done in some cases.
But I need to know what is the port to send an HTP request ...
there is a trick right now: the opened port is in a global var "ListenAddr". But global var is not compatible with Reload, neither with these Tests that are running in parallel in the same process.

We should definitively design a solution to avoid the global var. For prod and for Tests.

Waiting a solution, the shortcut is to use a different port for each Test.

miekg: you mean met.Addr? https://github.com/coredns/coredns/blob/master/plugin/metrics/metrics.go#L22

fturib: met.Addr has the value provided in the Stanza => "localhost:0"

Right now, the one that now the effective port associated to the listener is
https://github.com/coredns/coredns/blob/master/plugin/metrics/metrics.go#L96
```
ListenAddr = m.ln.Addr().String() // For tests
```

But it is a global var of that package .

fturib: But we could save it in met.EffectiveAddr and use it, and remove the global var.
Let me try this.
(will make more change though as I will have to update all UTs).

fturib: I had a look. It is useless to try to save the effective address into the metric plugins ... but it would be useful to save it into the instance.Storage that exist in caddy's instance.

We can now save any data from the Controller (in Setup) along with a Key, and access the same data using an Instance of Caddy.

But I think will be in another PR.

fturib:
@johnbelamaric , @miekg  : can we merge this PR ?
the TODO I set for the UT can be worked later ... this UT is valid.
Thanks!

PR #2082, opened 2018-09-04
Add failed request counter metric to proxy/forward

janeczku:
Metrics exported by proxy and forward plugins should provide visibility into failed upstream requests allowing operators to detect and troubleshoot service degradation that may result from dropped packets or other i/o errors in the upstream path.
This PR adds metric counters that get incremented when a request to an upstream server yields an error instead of a valid query response.
- `coredns_proxy_request_failure_count_total{server, proto, proto_proxy, family, to}` - failed queries per upstream.
- `coredns_forward_request_failure_count_total{to}` - failed queries per upstream.

NA
Included

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -97,6 +97,7 @@ If monitoring is enabled (via the *prometheus* directive) then the following met
 
 * `coredns_forward_request_duration_seconds{to}` - duration per upstream interaction.
 * `coredns_forward_request_count_total{to}` - query count per upstream.
+* `coredns_forward_request_failure_count_total{to}` - failed queries per upstream.
 * `coredns_forward_response_rcode_total{to, rcode}` - count of RCODEs per upstream.
 * `coredns_forward_healthcheck_failure_count_total{to}` - number of failed health checks per upstream.
 * `coredns_forward_healthcheck_broken_count_total{}` - counter of when all upstreams are unhealthy,
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -91,6 +91,15 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts options
 		proto = state.Proto()
 	}
 
+	var err error
+	defer func() {
+		if err != nil && err != dns.ErrTruncated && err != ErrCachedClosed {
+			RequestFailureCount.WithLabelValues(p.addr).Add(1)
+		}
+	}()
+
+	RequestCount.WithLabelValues(p.addr).Add(1)
+
 	conn, cached, err := p.transport.Dial(proto)
 	if err != nil {
 		return nil, err
@@ -104,10 +113,10 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts options
 
 	conn.SetWriteDeadline(time.Now().Add(maxTimeout))
 	reqTime := time.Now()
-	if err := conn.WriteMsg(state.Req); err != nil {
+	if err = conn.WriteMsg(state.Req); err != nil {
 		conn.Close() // not giving it back
 		if err == io.EOF && cached {
-			return nil, ErrCachedClosed
+			err = ErrCachedClosed
 		}
 		return nil, err
 	}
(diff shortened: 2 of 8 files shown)


plugin/forward/README.md
```diff
@@ -97,6 +97,7 @@ If monitoring is enabled (via the *prometheus* directive) then the following met
 
 * `coredns_forward_request_duration_seconds{to}` - duration per upstream interaction.
 * `coredns_forward_request_count_total{to}` - query count per upstream.
+* `coredns_forward_request_failure_count{to}` - failed query count per upstream.
```
SuperQ: This metric name does not fit the naming conventions. I would suggest this:

```
* `coredns_forward_request_failed_total{to}` - failed queries per upstream.
```

SuperQ (changes requested):

plugin/forward/README.md
janeczku: Thanks for the review @SuperQ. Changed the metric name to `request_failure_count_total` to align with Prometheus naming convention and existing failure counters like `healthcheck_failure_count_total`.

SuperQ: Yes, there's a bunch of metrics with `count` that is extraneous. I'll have to make a sweep through these and fix the names. Really, I would call it `request_failures_total`, or `requests_failed_total`.

janeczku: Makes sense. Will update tomorrow.

plugin/forward/connect.go
```diff
@@ -91,6 +91,15 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts options
 		proto = state.Proto()
 	}
 
+	var err error
+	defer func() {
```
miekg, line 95: why does this do a `defer`?

janeczku: Because there are two places (WriteMsg, ReadMsg) where the function returns early in case of error. The `defer` makes it so that in both cases the error is evaluated and the failure counter increased. I just thought it would be nicer than adding the same if-then-counter-increase logic twice. @miekg

plugin/forward/connect.go
```diff
@@ -91,6 +91,15 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts options
...
+		if err != nil && err != dns.ErrTruncated && err != ErrCachedClosed {
+			RequestFailureCount.WithLabelValues(p.addr).Add(1)
+		}
+	}()
+
+	RequestCount.WithLabelValues(p.addr).Add(1)
```
rdrozhdzh, line 101: What is actually the counters RequestCount and RequestFailureCount intended to count?
I believe, they should show how many **user** requests were handled/failed. In other words, how many times ServeDNS() was invoked successfully or not successfully.
The counters incremented here don't give us these numbers. Method Connect() can be called many times from the same ServeDNS() invocation because of retrying on some errors.
Perhaps, we need to update these metrics at higher level, i.e. in ServeDNS() itself.

janeczku: > What is actually the counters RequestCount and RequestFailureCount intended to count? I believe, they should show how many user requests were handled/failed.

I agree user requests should be measured at the highest level, ie. `core.ServeDNS`.

I don't think plugin metrics (at least as they are implemented now) were ever intended to measure user requests but rather provide plugin-level insight into state of the DNS service.

E.g.: https://github.com/coredns/coredns/blob/master/plugin/proxy/README.md
> coredns_proxy_request_count_total{server, proto, proto_proxy, family, to} - query count per upstream.

Both levels of metrics make sense, i want user request metrics to determine quality of service as seen by users. But i also want plugin-level metrics for troubleshooting and early detection of issues that might other wise be masked because of retrying (e.g. network congestion, dropped upstream requests).

rdrozhdzh: Yes, plugin-level metrics make sense. But we need clearly document what exactly each counter means. From current description it's not obvious to me

janeczku: Yeah let me improve the documentation of these metrics to make clear that they not necessarily correspond to rate of user requests.

plugin/forward/connect.go
miekg: defer's are still considered slow, so I rather not add more of them in the hot path.

miekg:
this looks very much like a stale pr - inclined to just close it in a few days.

janeczku:
Closing my PR as i dont have bandwidth currently to work on it.

Issue #2090, opened 2018-09-10
Make fails due to latest change in Prometheus instrumentation library 


bilbercode:
## Make outcome
```
# github.com/coredns/coredns/plugin/metrics
../../coredns/coredns/plugin/metrics/metrics.go:43:49: too many arguments in call to prometheus.NewProcessCollector
	have (int, string)
	want (prometheus.ProcessCollectorOpts)
```
## Vendor change
https://github.com/prometheus/client_golang/commit/7858729281ec582767b20e0d696b6041d995d5e0

PR #2091, opened 2018-09-10
fix: fixes issue with metrics process collector

bilbercode:
fixes call to create NewProcessCollector in metrics, ref https://github.com/prometheus/client_golang/commit/7858729281ec582767b20e0d696b6041d995d5e0
GitHub issue #2090
No changes required to documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ godeps:
 	go get -u github.com/prometheus/client_golang/prometheus
 	(cd $(GOPATH)/src/github.com/mholt/caddy              && git checkout -q v0.10.13)
 	(cd $(GOPATH)/src/github.com/miekg/dns                && git checkout -q v1.0.12)
-	(cd $(GOPATH)/src/github.com/prometheus/client_golang && git checkout -q v0.8.0)
+	(cd $(GOPATH)/src/github.com/prometheus/client_golang && git checkout -q v0.9.0)
 	@ # for travis only, if this fails we don't care, but don't see benchmarks
 	 go get -u golang.org/x/tools/cmd/benchcmp || true
 
--- a/plugin/metrics/metrics.go
+++ b/plugin/metrics/metrics.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"net"
 	"net/http"
-	"os"
 	"sync"
 	"time"
 
@@ -40,7 +39,7 @@ func New(addr string) *Metrics {
 	}
 	// Add the default collectors
 	met.MustRegister(prometheus.NewGoCollector())
-	met.MustRegister(prometheus.NewProcessCollector(os.Getpid(), ""))
+	met.MustRegister(prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{ }))
 
 	// Add all of our collectors
 	met.MustRegister(buildInfo)


jameshartig (approved):
LGTM

miekg:
Fails to build with:
~~~
plugin/metrics/metrics.go:42:49: not enough arguments in call to prometheus.NewProcessCollector
plugin/metrics/metrics.go:42:50: undefined: prometheus.ProcessCollectorOpts
~~~

jameshartig:
I don't understand why the build is failing. The latest github.com/prometheus/client_golang has that struct and I don't see us vendoring that. @miekg any ideas?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] fix: fixes is..." ]
>I don't understand why the build is failing. The latest github.com/prometheus/client_golang has that struct and I don't see us vendoring that. @miekg any ideas?

We pull down a specific version, see the `Makefile`, might be worth bumping that 
version.

(Unrelated we should probably bump everything in vendor as well).

miekg:
hmmm, now locally *without* this PR my test also doesn't compile, uh?
(doing something else [code wise], but I'll take a look later)

Update: cancelled a `make test` just before it checked out the version of prometheus we're currently using. 

If you bump the version in the  Makefile to something newer this is ready to be merged.

jameshartig:
@miekg how do you feel about v0.9.0-pre1 locking for now? The README talks about their philosophy: https://github.com/prometheus/client_golang/blob/master/README.md

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] fix: fixes is..." ]
>@miekg how do you feel about v0.9.0-pre1 locking for now? The README talks about their philosophy: https://github.com/prometheus/client_golang/blob/master/README.md

We use 'git checkout -q v0.8.0' now, is v0.9.0-pre1 worse/better? Can't tell, 
we're also not the most advanced users, so if you think is OK, I'm fine with 
that.

/Miek

--
Miek Gieben

bilbercode:
Hey guys, my bad on this PR. I did a `go get -u ./...` which pulled from master on Prometheus. Obviously they are doing some work on the process collector

jameshartig:
I updated to v0.9.0-pre1 but is that not new enough for the change @bilbercode? Looks like the travis build still failed saying the method signature is still the old one.

@bilbercode is it better if we just wait until v0.9.0 is tagged (and complete) before merging this change?

bilbercode:
I would stick with 0.8.0 for now and not roll to the pre release, the interface could change :)

jameshartig:
Sounds good. We'll punt on this until v0.9.0 is complete.

bilbercode:
I'm actively developing an in house plugin and regularly use prometheus (prom) for instrumentation, if a newer version is released I will try and fix any issues an update this PR 👍

hmage:
v0.9.0 was just released -- https://github.com/prometheus/client_golang/releases/tag/v0.9.0

Guess it's time to bump the version.

miekg:
yes, please rebase and do all the magic to make it pass tests.

bilbercode:
Hey folks
I rebased from master and updated the version to the latest prom, most tests pass, however this line in the benchmark block of the `Makefile`
https://github.com/coredns/coredns/blob/d9efa9637460e799afa065c771ff009057e2bf85/Makefile#L70 will mostly fail if there are any code changes in the vendor deps that were patching because we are checking out old code against a newer version of the deps.

yongtang:
Guess we could ignore the benchmark failure? If there is an issue after the merge we could revert I think.

bilbercode:
I agree, I don't think we should loose the benchmark comparison, when were dealing with issues like this is mostly an edge case :)

miekg:
This needs a rebase?

miekg:
want to rebase and see this still applies?

PR #2096, opened 2018-09-13
[plugin/proxy]: Return on WriteMsg err.

dilyevsky:
Followup PR on the heels of #2050. Short-circut and log on error from
`WriteMsg`. This will prevent code getting stuck on `ReadMsg` and allow
for easier debugging.

Also simply the `ReadMsg` calling code a little.


miekg:
`dnsco.Close()` isn't called anymore. Wrt to log.debug lgtm

dilyevsky:
> `dnsco.Close()` isn't called anymore

Yeah it's actually already done two layers up in `Exchane` (also better style to handle full lifecycle at a single layer).

PR #2104, opened 2018-09-19
Better naming

miekg:
Move functions from pkg/transport to pkg/parse
    
    Although "parse" is a fairly generic name I believe this is somewhat
    better named. pkg/transport keeps a few constants that are uses
    throughout for the rest is is renaming a bunch (and the fallout from
    there to make things compile again).
    

miekg:
apart from renaming not code inside functions was changed

SuperQ (approved):
LGTM

Issue #2003, opened 2018-07-26
plugin/forward: Panic while healthchecking


negz:
I encountered the following panic today while experimenting with deploying CoreDNS

```
$ kubectl --context tfk-uc14 -n kube-system logs coredns-7b58f898db-zzttn --previous
.:53
2018/07/25 23:17:32 [INFO] CoreDNS-1.2.0
2018/07/25 23:17:32 [INFO] linux/amd64, go1.10.3, 2e322f6
CoreDNS-1.2.0
linux/amd64, go1.10.3, 2e322f6
2018/07/25 23:17:32 [INFO] plugin/reload: Running configuration MD5 = df9eeab6ca5daae007e09be2306a3d29
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x11dd3b9]
goroutine 3941 [running]:
github.com/coredns/coredns/plugin/forward.(*Proxy).Healthcheck.func1(0xc4201b0cd0, 0x0)
        /tmp/tmp.RJbV5C8cVr/g/src/github.com/coredns/coredns/plugin/forward/proxy.go:54 +0x29
github.com/coredns/coredns/plugin/pkg/up.(*Probe).Do.func1(0xc42035d6d0, 0x1dcd6500, 0xc4204d2000)
        /tmp/tmp.RJbV5C8cVr/g/src/github.com/coredns/coredns/plugin/pkg/up/up.go:38 +0x36
created by github.com/coredns/coredns/plugin/pkg/up.(*Probe).Do
        /tmp/tmp.RJbV5C8cVr/g/src/github.com/coredns/coredns/plugin/pkg/up/up.go:36 +0xa2
```

My `Corefile` was as follows:
```
.:53 {
    errors
    nsid CoreDNS
    health
    kubernetes cluster.local in-addr.arpa ip6.arpa {
      pods insecure
      upstream
      fallthrough in-addr.arpa ip6.arpa
    }
    trace zipkin zipkin.kube-system:9411 {
	    every 1000
    }
    prometheus :9153
    forward . /etc/resolv.conf
    cache 30
    reload
    loadbalance
}
```

The `resolv.conf` of the affected CoreDNS pods was:
```
# This file is managed by man:systemd-resolved(8). Do not edit.
#
# This is a dynamic resolv.conf file for connecting local clients directly to
# all known uplink DNS servers. This file lists all configured search domains.
#
[...]

PR #2115, opened 2018-09-21
Add test for #2003

miekg:
This adds a test for cleanup in c349446a


miekg:
going in (just a test) and should have been added in the original PR

PR #2117, opened 2018-09-22
plugin/route53: fix multiple route 53 zones issue

canthefason:
When we pass multiple Route 53 Hosted Zones, only the last zone from the list returns answers.


miekg:
the typical range gotcha. thanks for catching this and the test.

PR #2118, opened 2018-09-22
plugins/route53: add AWS credentials file support

canthefason:
This PR creates AWS credentials file support and enables setting different AWS profile names for each server block

### DIFF
--- a/plugin/route53/README.md
+++ b/plugin/route53/README.md
@@ -16,6 +16,7 @@ The route53 plugin can be used when coredns is deployed on AWS or elsewhere.
 route53 [ZONE:HOSTED_ZONE_ID...] {
     [aws_access_key AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY]
     upstream [ADDRESS...]
+    [credentials_file PROFILE FILENAME]
 }
 ~~~
 
@@ -29,10 +30,13 @@ route53 [ZONE:HOSTED_ZONE_ID...] {
    to external hosts (eg. used to resolve CNAMEs). If no **ADDRESS** is given, CoreDNS will resolve
    against itself. **ADDRESS** can be an IP, an IP:port or a path to a file structured like
    resolv.conf (**NB**: Currently a bug (#2099) is preventing the use of self-resolver).
+* `credentials_file` used for reading the credential file and setting the profile name for a given zone.
+* **PROFILE** AWS account profile name. Defaults to `default`.
+* **FILENAME** AWS credentials filename. Defaults to `~/.aws/credentials`
 
 ## Examples
 
-Enable route53 with implicit aws credentials and an upstream:
+Enable route53 with implicit AWS credentials and an upstream:
 
 ~~~ txt
 . {
@@ -41,12 +45,22 @@ Enable route53 with implicit aws credentials and an upstream:
 }
 ~~~
 
-Enable route53 with explicit aws credentials:
+Enable route53 with explicit AWS credentials:
 
 ~~~ txt
 . {
     route53 example.org.:Z1Z2Z3Z4DZ5Z6Z7 {
       aws_access_key AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY
+  }
+}
+~~~
+
+Enable route53 with AWS credentials file:
+
+~~~ txt
+. {
+    route53 example.org.:Z1Z2Z3Z4DZ5Z6Z7 {
+      credentials_file some-user
   }
 }
 ~~~
--- a/plugin/route53/README.md
+++ b/plugin/route53/README.md
@@ -16,7 +16,7 @@ The route53 plugin can be used when coredns is deployed on AWS or elsewhere.
 route53 [ZONE:HOSTED_ZONE_ID...] {
     [aws_access_key AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY]
     upstream [ADDRESS...]
-    [credentials_file PROFILE FILENAME]
+    credentials PROFILE [FILENAME]
 }
 ~~~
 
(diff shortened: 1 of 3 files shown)


plugin/route53/README.md
```diff
@@ -29,10 +30,13 @@ route53 [ZONE:HOSTED_ZONE_ID...] {
    to external hosts (eg. used to resolve CNAMEs). If no **ADDRESS** is given, CoreDNS will resolve
    against itself. **ADDRESS** can be an IP, an IP:port or a path to a file structured like
    resolv.conf (**NB**: Currently a bug (#2099) is preventing the use of self-resolver).
+* `credentials_file` used for reading the credential file and setting the profile name for a given zone.
```
miekg: make it just `credentials`?

plugin/route53/README.md
```diff
@@ -16,6 +16,7 @@ The route53 plugin can be used when coredns is deployed on AWS or elsewhere.
 route53 [ZONE:HOSTED_ZONE_ID...] {
     [aws_access_key AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY]
     upstream [ADDRESS...]
+    [credentials_file PROFILE FILENAME]
```
miekg: just put it like :`credential PROFILE [FILENAME]`

that the entire option is optionally and require `[` and `]` is not something we do consistently, also introduces quite some noise in the docs IMO

miekg (commented):
some naming things, but otherwise lgtm

plugin/route53/README.md
yongtang: Agree with @miekg, just `credentials` would be better

plugin/route53/setup.go
```diff
@@ -75,11 +82,23 @@ func setup(c *caddy.Controller, f func(*credentials.Credentials) route53iface.Ro
...
 					return c.Errf("invalid upstream: %v", err)
 				}
+			case "credentials_file":
+				args := c.RemainingArgs()
+				if len(args) > 0 {
+					sharedProvider.Profile = args[0]
+				}
+				if len(args) > 1 {
+					sharedProvider.Filename = args[1]
+				}
```
yongtang, line 102:
```
if len(args) > 1 {
    sharedProvider.Filename = args[1]
```

That means additional args after 1 will be silently ignored?

yongtang (changes requested):

plugin/route53/README.md
canthefason: `credentials` by itself didn't seem intuitive to me and felt it could be easily confused with `aws_access_key`, but I'm ok with changing it.

plugin/route53/setup.go
canthefason: Yes, that's right. Would you like me to error out when len(args) > 2?

dilyevsky: I think this should be:

```
shared := &credentials.SharedCredentialsProvider{}
if c.NextArg() {
    shared.Profile = c.Val()
} else {
  return c.ArgErr()
}
if c.NextArg() {
    shared.Filename = c.Val()
    if c.NextArg() {
        return c.Err("number of arguments is too damn high")
    }
}
providers = append(providers,  shared)
```

Something like that. You don't check for required `PROFILE` argument currently. Also can you add a test for this.

plugin/route53/setup.go
```diff
@@ -75,11 +82,23 @@ func setup(c *caddy.Controller, f func(*credentials.Credentials) route53iface.Ro
...
 			default:
 				return c.Errf("unknown property '%s'", c.Val())
 			}
 		}
 	}
+	providers = append(providers, &credentials.EnvProvider{}, sharedProvider)
+
+	credential := credentials.NewChainCredentials(providers)
+
 	client := f(credential)
```
dilyevsky: You can inline `credentials.NewChainCredentials(providers)`, `credential` is not used anywhere else it seems.

plugin/route53/setup.go
```diff
@@ -35,7 +35,9 @@ func init() {
 
 func setup(c *caddy.Controller, f func(*credentials.Credentials) route53iface.Route53API) error {
 	keys := map[string]string{}
-	credential := credentials.NewEnvCredentials()
+	sharedProvider := &credentials.SharedCredentialsProvider{}
+	providers := []credentials.Provider{}
```
dilyevsky: Just `var provider []credentials.Provider` will do.

plugin/route53/setup.go
```diff
@@ -35,7 +35,9 @@ func init() {
 
 func setup(c *caddy.Controller, f func(*credentials.Credentials) route53iface.Route53API) error {
 	keys := map[string]string{}
-	credential := credentials.NewEnvCredentials()
+	sharedProvider := &credentials.SharedCredentialsProvider{}
```
dilyevsky, line 46: Why do you need this here? Don't you want to append shared provider only when `credentials` stanza is present? This comment also applies to line 98 below.

dilyevsky (changes requested):

canthefason:
@dilyevsky @yongtang @miekg Instead of responding to individual comments, it will be better to put a comment under the PR itself. 

First off, let me clarify the existing behavior and then we can discuss the next steps. The Route53 plugin doc claims this: `If they (aws credentials) are not provided , then coredns tries to access AWS credentials the same way as AWS CLI, e.g., environmental variables, AWS credentials file, instance profile credentials, etc.`. Unfortunately, this explanation is not satisfied by the existing implementation. Apart from the static keys provided by the user, the current setup only does Environment Variable check as far as I can tell, and the rest is left out. Please correct me if I’m wrong.

Depending on this doc (and most of the applications that leverages AWS APIs), all static aws keys, environment variables and credentials could be optional, and the precedence is as follows: Static aws keys -> Environment Variables -> Credentials file -> IAM role. This PR intends to ensure that this precedence is in place.

miekg:
Ok. Thanks. I'm tied up in other coredns work. But fixing the docs to match
implementation seems a good start.

plugin/route53/setup.go
canthefason: https://github.com/coredns/coredns/pull/2118#issuecomment-424049145

when the `credentials` stanza doesn't exist, this will still try to read that credentials file with the `default` profile depending on the precedence.

plugin/route53/setup.go
canthefason: Updated this one. Just realized that there's no need for that extra NextArg that checks extra args. This is something magically handled by Caddy Dispenser itself, however, still added another test case to validate that it works as expected.

canthefason:
@dilyevsky @yongtang Updated this PR. PTAL.

plugin/route53/setup.go
dilyevsky: I see. Could you add comment about this so that somebody doesn't remove it later by mistake.

dilyevsky (changes requested):
one comment request, then lgtm

dilyevsky (approved):

PR #2119, opened 2018-09-22
K8s remove string ops

miekg:
plugin/kubernetes: remove bunch a string ops
    
    This removes a bunch of appends to where not needed, makes dnsutil.Join
    take variadic args which removes the need to wrap in a new string slice.
    

yongtang (approved):
LGTM

PR #2122, opened 2018-09-22
plugin/kubernetes: remove var namespace

miekg:
The removes the var namespace which was not needed.

Also uncovered a bug because of this in the namespace selection or hiding, because namespace as a var existed.


yongtang (approved):
LGTM

PR #2124, opened 2018-09-22
plugin/kubernetes: remove unnecessary checks

miekg:
These checks are not needed and also use a var for all obj errors.


yongtang (approved):
LGTM

PR #2126, opened 2018-09-22
Unblock startup if kubernetes API is unavailable

### LINKED ISSUE #2116: startup blocked if kubernetes API is unavailable
#### Version
CoreDNS-1.2.2
linux/amd64, go1.11, eb51e8b

#### Description
I'm experimenting with a bit of a weird setup, where I'm attempting to use CoreDNS to serve some local records, proxy to an upstream provider, and act as a replica for our kubernetes service discovery, running outside of kubernetes. The problem I'm experiencing, is the kubernetes API might not be available when CoreDNS starts, and if it's not available, I'm experiencing a hang in the CoreDNS startup where the process never binds and isn't able to serve non-kubernetes records.

I've isolated it to this code:
https://github.com/coredns/coredns/blob/master/plugin/kubernetes/setup.go#L71-L75

Which appears to have been introduced to prevent a race / unit tests:
PR: #1155 
Issue: #1006

#### Desired Behaiour
I'm not sure if this is intended behaviour or not. 

My expectation would be, that the kubernetes API being unavailable/unreachable would not block startup of a server under partial outage, only return serv fail on the zones the kubernetes plugin is responsible for (depending on fall-through).

I haven't investigated the tests, to see what implications this approach might have from a unit test perspective.

Alternatively, as a simpler workaround, I was considering the option to simply time limit the start-up routine would be blocked waiting for the kubernetes resources to sync. I was thinking something like 10-15 seconds would work.

#### Closing
I'm willing to contribute a fix for this issue with the desired approach.

Thank You

knisbet:
CoreDNS startup will currently block indefinitely if the kubernetes plugin isn't able to sync with the kubernetes API. This shifts the startup to only block for up to 5 seconds, and otherwise causes the kubernetes plugin to return servfail if fall through isn't enabled until the kubernetes api becomes available and the sync completes.

Fixes #2116 

N/A


miekg:
I thought the original reason for doing this, was that the plugin would hand out 
NXDOMAIN that would be cached in the cluster (which would then be cached). This 
PR reinstates that behavior.

We should double check what the plugin returns while not connected to the API - 
this should be SERVFAIL and nothing else.

/Miek

--
Miek Gieben

knisbet:
@meikg shouldn't https://github.com/coredns/coredns/pull/2126/files#diff-347a95a2d36b7b336cb55cc146bd16d6R68 take care of that scenario, or did I miss something? 

At least in my own testing, I get SERVFAIL for kubernetes plugin controlled zones after the first 5 seconds of startup.
```
kevin-test1:/$ dig @127.0.0.2 kubernetes.default.svc.cluster.local. +time=1

; <<>> DiG 9.10.3-P4-Debian <<>> @127.0.0.2 kubernetes.default.svc.cluster.local. +time=1
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 13047
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
;; QUESTION SECTION:
;kubernetes.default.svc.cluster.local. IN A

;; AUTHORITY SECTION:
cluster.local.		300	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 0 7200 1800 86400 30

;; Query time: 0 msec
;; SERVER: 127.0.0.2#53(127.0.0.2)
;; WHEN: Sat Sep 22 17:24:49 UTC 2018
;; MSG SIZE  rcvd: 158
```
Plugin Config:
```
    endpoint https://leader.telekube.local:6443
    tls /var/state/apiserver-kubelet-client.cert /var/state/apiserver-kubelet-client.key /var/state/root.cert
    pods disabled
  }
```

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Unblock start..." ]
>@meikg shouldn't https://github.com/coredns/coredns/pull/2126/files#diff-347a95a2d36b7b336cb55cc146bd16d6R68 take care of that scenario, or did I miss something?

yes, nice - that should work.

/Miek

--
Miek Gieben

miekg:
is it possible to write a test for this?
otherwise: lgtm

plugin/kubernetes/setup.go
```diff
@@ -68,13 +68,19 @@ func (k *Kubernetes) RegisterKubeCache(c *caddy.Controller) {
...
-		return nil
+		timeout := time.After(5 * time.Second)
+		ticker := time.NewTicker(100 * time.Millisecond)
+		for {
+			select {
+			case <-ticker.C:
+				if k.APIConn.HasSynced() {
+					return nil
+				}
+			case <-timeout:
```
miekg, line 89: are we reconnecting to the API when it's available, or are we just dead in the water now?

knisbet: The API connection works in the background trying to do the sync. This just changes the behaviour, of blocking the startup indefinitely waiting for the initial sync (HasSynced) to be true by only waiting 5 seconds for the sync then letting startup continue. 

When the initial sync completes (HasSynced = true) ServeDNS will stop returning ServerFailure for kubernetes zones.

knisbet:
I spent a bit of time looking through the test infra, and I'm not sure how to cleanly inject a test for this in the existing table driven test. It looks possible to create a separate test case for this.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Unblock start..." ]
>I spent a bit of time looking through the test infra, and I'm not sure how to cleanly inject a test for this in the existing table driven test. It looks possible to create a separate test case for this.

Please check. Otherwise we merge as-is.

fturib:
maybe the test can be in coredns/ci, along with other tests of kubernetes plugin.

chrisohaver (changes requested):
Sorry for not responding until now.  Can you do a test a unit test by running the tests on a implementation a DnsController (as other k8s plugin unit tests do) with HasSynced() always returning false.

knisbet:
Hey, sorry I disappeared, I started working on a new project. I made an attempt at adding a test case, that doesn't seem overly messy. Let me know what you think.

chrisohaver (approved):
Looks good!

chrisohaver (approved):
LGTM

PR #2107, opened 2018-09-20
Benchmark for k8s services

bradbeam:
got curious about some of the k8s perf issues and thought I'd take a shot at putting together a little benchmark for it. unsure on the usefulness or if it's the right approach, but figured it was worth an attempt. 

kicking this test off with `go test -bench=. -benchmem -memprofile memprofile.out -cpuprofile profile.out` and reviewing the memprofile, 

```
$ go tool pprof memprofile.out
Type: alloc_space
Time: Sep 19, 2018 at 9:47pm (CDT)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 1669.92MB, 98.48% of 1695.64MB total
Dropped 65 nodes (cum <= 8.48MB)
Showing top 10 nodes out of 18
      flat  flat%   sum%        cum   cum%
  954.29MB 56.28% 56.28%   954.29MB 56.28%  github.com/coredns/coredns/plugin/kubernetes.APIBench.GetNamespaceByName
  191.51MB 11.29% 67.57%   693.52MB 40.90%  github.com/coredns/coredns/plugin/etcd/msg.Path
     134MB  7.90% 75.48%   229.01MB 13.51%  github.com/miekg/dns.SplitDomainName
   95.50MB  5.63% 81.11%    95.50MB  5.63%  path.(*lazybuf).append (inline)
      95MB  5.60% 86.71%       95MB  5.60%  github.com/miekg/dns.Split
   92.50MB  5.46% 92.17%    92.50MB  5.46%  path.(*lazybuf).string (inline)
      85MB  5.01% 97.18%       85MB  5.01%  strings.Join
   20.54MB  1.21% 98.39%    20.54MB  1.21%  github.com/coredns/coredns/plugin/kubernetes.generateEndpoints
    1.57MB 0.093% 98.48%    19.58MB  1.15%  github.com/coredns/coredns/plugin/kubernetes.generateSvcs
         0     0% 98.48%   954.29MB 56.28%  github.com/coredns/coredns/plugin/kubernetes.(*APIBench).GetNamespaceByName
```

i hardcoded it for a 8192 / `/19` endpoint test since it was 'easiest' to go with a cidr notation and the /20 was under 5k endpoints to try to get some data close to the 5k node test.

Issue #2095, opened 2018-09-13
plugin/kubernetes: Reduce memory footprint at scale


chrisohaver:
Relating to CoreDNS OOM death in kubernetes/kubernetes#68613.

Evidence suggests that the endpoint store grows too large in the scale tests, wherein ~100K Endpoints exist. In CoreDNS, all of these endpoints are stored in client-go endpoint store cache.

One partial solution would be to not store endpoints for cluster IP services.  Endpoints records are technically not needed per the DNS Spec.  I suspect that kube-dns operates this way.  Note that in the scale tests where CoreDNS is OOM killed, no headless services are created, which means that if we skip ClusterIP Endpoints, we would end up not stashing any Endpoints in the store at all.  This could explain why kube-dns does not get OOM killed in the same test.

Skipping ClusterIP endpoints would however disable a CoreDNS feature that allows wildcard searching for all endpoints in a cluster IP service. Someone tweeted about how awesome this feature was a while back.  So, it could be missed.  Perhaps we could make it an option to enable (with a warning about memory impact on large clusters, as we warn with `pods verified` mode).

Another solution would be to optimize the data we store for endpoints, so that only the data we need to construct the record is kept.

miekg:
I think OOMing trumps any feature that is not part of the DNS spec, so let's not complicate things by adding options that will lead to bad behavior later on.

*Not* storing something sounds like a simple fix...?

chrisohaver:
>  let's not complicate things by adding options

agreed.

And yes, It looks to be a simple fix. I think there would be some race conditions early, that might force us to have to store some endpoints if we receive events for them before we receive the service events.  But that is short lived, and i think by schema, Services must exist before an Endpoint can be created anyways.

However, @johnbelamaric did a quick test, and it seems that endpoint objects would not take up enough memory to account for the OOM.  So, there may be something else going on here.

chrisohaver:
> And yes, It looks to be a simple fix.

Actually, this looks to be more complicated than I thought.  I don't yet see how to hook into the watch store function.

miekg:
Why doesn't skydns have this issue?

chrisohaver:
skydns == kube-dns ?

On Sep 15, 2018, at 9:34 AM, Miek Gieben <notifications@github.com<mailto:notifications@github.com>> wrote:


Why doesn't skydns have this issue?

miekg:
[...]

PR #2128, opened 2018-09-23
plugin/kubernetes: allow trimming down of cached items.

miekg:
plugin/kubernetes: allow trimming down of cached items.
    
    Implement the simplest case for *api.Pod where we only need
    DeletionTimestamp, PodIP and the Namespace. Creates a new subdir
    'object' under kubernetes where we store all this.
    
### DIFF
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -397,7 +397,7 @@ func (k *Kubernetes) findPods(r recordRequest, zone string) (pods []msg.Service,
 		}
 
 		// exclude pods in the process of termination
-		if !p.DeletionTimestamp.IsZero() {
+		if p.Deleting {
 			continue
 		}
 
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/etcd/msg"
+	"github.com/coredns/coredns/plugin/kubernetes/object"
 	"github.com/coredns/coredns/plugin/pkg/dnsutil"
 	"github.com/coredns/coredns/plugin/pkg/fall"
 	"github.com/coredns/coredns/plugin/pkg/healthcheck"
@@ -304,18 +305,18 @@ func serviceFQDN(obj meta.Object, zone string) string {
 }
 
 // podFQDN returns the k8s cluster dns spec FQDN for the pod.
-func podFQDN(p *api.Pod, zone string) string {
-	if strings.Contains(p.Status.PodIP, ".") {
-		name := strings.Replace(p.Status.PodIP, ".", "-", -1)
+func podFQDN(p *object.Pod, zone string) string {
+	if strings.Contains(p.PodIP, ".") {
+		name := strings.Replace(p.PodIP, ".", "-", -1)
 		return dnsutil.Join(name, p.GetNamespace(), Pod, zone)
 	}
 
-	name := strings.Replace(p.Status.PodIP, ":", "-", -1)
+	name := strings.Replace(p.PodIP, ":", "-", -1)
 	return dnsutil.Join(name, p.GetNamespace(), Pod, zone)
 }
 
 // endpointFQDN returns a list of k8s cluster dns spec service FQDNs for each subset in the endpoint.
-func endpointFQDN(ep *api.Endpoints, zone string, endpointNameMode bool) []string {
+func endpointFQDN(ep *object.Endpoints, zone string, endpointNameMode bool) []string {
 	var names []string
 	for _, ss := range ep.Subsets {
 		for _, addr := range ss.Addresses {
@@ -325,12 +326,12 @@ func endpointFQDN(ep *api.Endpoints, zone string, endpointNameMode bool) []strin
 	return names
 }
 
-func endpointHostname(addr api.EndpointAddress, endpointNameMode bool) string {
+func endpointHostname(addr object.EndpointAddress, endpointNameMode bool) string {
 	if addr.Hostname != "" {
 		return addr.Hostname
 	}
-	if endpointNameMode && addr.TargetRef != nil && addr.TargetRef.Name != "" {
-		return addr.TargetRef.Name
+	if endpointNameMode && addr.TargetRefName != "" {
+		return addr.TargetRefName
 	}
 	if strings.Contains(addr.IP, ".") {
 		return strings.Replace(addr.IP, ".", "-", -1)
@@ -396,12 +397,12 @@ func (k *Kubernetes) findPods(r recordRequest, zone string) (pods []msg.Service,
 		}
 
 		// exclude pods in the process of termination
-		if !p.ObjectMeta.DeletionTimestamp.IsZero() {
+		if !p.DeletionTimestamp.IsZero() {
 			continue
 		}
 
 		// check for matching ip and namespace
-		if ip == p.Status.PodIP && match(namespace, p.Namespace) {
+		if ip == p.PodIP && match(namespace, p.Namespace) {
 			s := msg.Service{Key: strings.Join([]string{zonePath, Pod, namespace, podname}, "/"), Host: ip, TTL: k.ttl}
 			pods = append(pods, s)
 
@@ -424,9 +425,9 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 	}
 
 	var (
-		endpointsListFunc func() []*api.Endpoints
-		endpointsList     []*api.Endpoints
-		serviceList       []*api.Service
+		endpointsListFunc func() []*object.Endpoints
+		endpointsList     []*object.Endpoints
+		serviceList       []*object.Service
 	)
 
 	// handle empty service name
@@ -441,11 +442,11 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 
 	if wildcard(r.service) || wildcard(r.namespace) {
 		serviceList = k.APIConn.ServiceList()
-		endpointsListFunc = func() []*api.Endpoints { return k.APIConn.EndpointsList() }
+		endpointsListFunc = func() []*object.Endpoints { return k.APIConn.EndpointsList() }
 	} else {
-		idx := r.service + "." + r.namespace
+		idx := object.ServiceKey(r.service, r.namespace)
 		serviceList = k.APIConn.SvcIndex(idx)
-		endpointsListFunc = func() []*api.Endpoints { return k.APIConn.EpIndex(idx) }
+		endpointsListFunc = func() []*object.Endpoints { return k.APIConn.EpIndex(idx) }
 	}
 
 	for _, svc := range serviceList {
@@ -459,7 +460,7 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 			continue
 		}
 
-		if k.opts.ignoreEmptyService && svc.Spec.ClusterIP != api.ClusterIPNone {
+		if k.opts.ignoreEmptyService && svc.ClusterIP != api.ClusterIPNone {
 			// serve NXDOMAIN if no endpoint is able to answer
 			podsCount := 0
 			for _, ep := range endpointsListFunc() {
@@ -474,12 +475,12 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 		}
 
 		// Endpoint query or headless service
-		if svc.Spec.ClusterIP == api.ClusterIPNone || r.endpoint != "" {
+		if svc.ClusterIP == api.ClusterIPNone || r.endpoint != "" {
 			if endpointsList == nil {
 				endpointsList = endpointsListFunc()
 			}
 			for _, ep := range endpointsList {
-				if ep.ObjectMeta.Name != svc.Name || ep.ObjectMeta.Namespace != svc.Namespace {
+				if ep.Name != svc.Name || ep.Namespace != svc.Namespace {
 					continue
 				}
 
@@ -491,11 +492,6 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 							if !match(r.endpoint, endpointHostname(addr, k.endpointNameMode)) {
 								continue
 							}
-						}
-
-						if len(eps.Ports) == 0 {
-							// add a sentinel port (-1) entry so we create records for services without any declared ports
-							eps.Ports = append(eps.Ports, api.EndpointPort{Port: -1})
 						}
 
 						for _, p := range eps.Ports {
@@ -516,8 +512,8 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 		}
 
 		// External service
-		if svc.Spec.Type == api.ServiceTypeExternalName {
-			s := msg.Service{Key: strings.Join([]string{zonePath, Svc, svc.Namespace, svc.Name}, "/"), Host: svc.Spec.ExternalName, TTL: k.ttl}
+		if svc.Type == api.ServiceTypeExternalName {
+			s := msg.Service{Key: strings.Join([]string{zonePath, Svc, svc.Namespace, svc.Name}, "/"), Host: svc.ExternalName, TTL: k.ttl}
 			if t, _ := s.HostType(); t == dns.TypeCNAME {
 				s.Key = strings.Join([]string{zonePath, Svc, svc.Namespace, svc.Name}, "/")
 				services = append(services, s)
@@ -528,18 +524,14 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 		}
 
 		// ClusterIP service
-		if len(svc.Spec.Ports) == 0 {
-			// add a sentinel port (-1) entry so we create records for services without any declared ports
-			svc.Spec.Ports = append(svc.Spec.Ports, api.ServicePort{Port: -1})
-		}
-		for _, p := range svc.Spec.Ports {
+		for _, p := range svc.Ports {
 			if !(match(r.port, p.Name) && match(r.protocol, string(p.Protocol))) {
 				continue
 			}
 
 			err = nil
 
-			s := msg.Service{Host: svc.Spec.ClusterIP, Port: int(p.Port), TTL: k.ttl}
+			s := msg.Service{Host: svc.ClusterIP, Port: int(p.Port), TTL: k.ttl}
 			s.Key = strings.Join([]string{zonePath, Svc, svc.Namespace, svc.Name}, "/")
 
 			services = append(services, s)
--- a/plugin/kubernetes/object/endpoint.go
+++ b/plugin/kubernetes/object/endpoint.go
@@ -92,6 +92,8 @@ func ToEndpoints(obj interface{}) interface{} {
 		}
 	}
 
+	end = nil
+
 	return e
 }
 
--- a/plugin/kubernetes/object/endpoint.go
+++ b/plugin/kubernetes/object/endpoint.go
@@ -54,7 +54,6 @@ func ToEndpoints(obj interface{}) interface{} {
 		Name:      end.GetName(),
 		Namespace: end.GetNamespace(),
 		Index:     EndpointsKey(end.GetName(), end.GetNamespace()),
-		IndexIP:   make([]string, len(end.Subsets)),
 		Subsets:   make([]EndpointSubset, len(end.Subsets)),
 	}
 	for i, eps := range end.Subsets {
@@ -87,9 +86,9 @@ func ToEndpoints(obj interface{}) interface{} {
 		e.Subsets[i] = sub
 	}
 
-	for i, eps := range end.Subsets {
+	for _, eps := range end.Subsets {
 		for _, a := range eps.Addresses {
-			e.IndexIP[i] = a.IP
+			e.IndexIP = append(e.IndexIP, a.IP)
 		}
 	}
 
@@ -103,8 +102,9 @@ func (e *Endpoints) CopyWithoutSubsets() *Endpoints {
 		Name:      e.Name,
 		Namespace: e.Namespace,
 		Index:     e.Index,
-		IndexIP:   e.IndexIP,
+		IndexIP:   make([]string, len(e.IndexIP)),
 	}
+	copy(e1.IndexIP, e.IndexIP)
 	return e1
 }
 
--- a/plugin/kubernetes/object/endpoint.go
+++ b/plugin/kubernetes/object/endpoint.go
@@ -1,0 +1,146 @@
+package object
+
+import (
+	api "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// Endpoints is a stripped down api.Endpoints with only the items we need for CoreDNS.
+type Endpoints struct {
+	Version   string
+	Name      string
+	Namespace string
+	Index     string
+	IndexIP   []string
+	Subsets   []EndpointSubset
+
+	*Empty
+}
+
+// EndpointSubset is a group of addresses with a common set of ports. The
+// expanded set of endpoints is the Cartesian product of Addresses x Ports.
+type EndpointSubset struct {
+	Addresses []EndpointAddress
+	Ports     []EndpointPort
+}
+
+// EndpointAddress is a tuple that describes single IP address.
+type EndpointAddress struct {
+	IP            string
+	Hostname      string
+	NodeName      string
+	TargetRefName string
+}
+
+// EndpointPort is a tuple that describes a single port.
+type EndpointPort struct {
+	Port     int32
+	Name     string
+	Protocol string
+}
+
+// EndpointsKey return a string using for the index.
+func EndpointsKey(name, namespace string) string { return name + "." + namespace }
+
+// ToEndpoints converts an api.Service to a *Service.
+func ToEndpoints(obj interface{}) interface{} {
+	end, ok := obj.(*api.Endpoints)
+	if !ok {
+		return nil
+	}
+
+	e := &Endpoints{
+		Version:   end.GetResourceVersion(),
+		Name:      end.GetName(),
+		Namespace: end.GetNamespace(),
+		Index:     EndpointsKey(end.GetName(), end.GetNamespace()),
+	}
+	for _, eps := range end.Subsets {
+		sub := EndpointSubset{}
+		for _, a := range eps.Addresses {
+			ea := EndpointAddress{IP: a.IP, Hostname: a.Hostname}
+			if a.NodeName != nil {
+				ea.NodeName = *a.NodeName
+			}
+			if a.TargetRef != nil {
+				ea.TargetRefName = a.TargetRef.Name
+			}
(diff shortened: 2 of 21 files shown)


miekg:
Don't understand why the integration tests are failing:
~~~
go test -v ./test/kubernetes/...
=== RUN   TestKubernetesA
=== RUN   TestKubernetesA/svc-1-a.test-1.svc.cluster.local._A
=== RUN   TestKubernetesA/headless-svc.test-1.svc.cluster.local._A
=== RUN   TestKubernetesA/bogusservice.test-1.svc.cluster.local._A
=== RUN   TestKubernetesA/bogusendpoint.svc-1-a.test-1.svc.cluster.local._A
=== RUN   TestKubernetesA/svc-1-a.*.svc.cluster.local._A
=== RUN   TestKubernetesA/*.test-1.svc.cluster.local._A
=== RUN   TestKubernetesA/*.test-2.svc.cluster.local._A
=== RUN   TestKubernetesA/10-20-0-101.test-1.pod.cluster.local._A
=== RUN   TestKubernetesA/dns-version.cluster.local._TXT
=== RUN   TestKubernetesA/headless-svc.test-1.svc.cluster.local._AAAA
--- FAIL: TestKubernetesA (9.53s)
    --- FAIL: TestKubernetesA/svc-1-a.test-1.svc.cluster.local._A (0.32s)
        helpers.go:118: Rcode is "NXDOMAIN", expected "NOERROR"
        helpers.go:301: ;; opcode: QUERY, status: NXDOMAIN, id: 33536
            ;; flags: rd; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0
            
            ;; QUESTION SECTION:
            ;svc-1-a.test-1.svc.cluster.local.	IN	 A
            
            ;; AUTHORITY SECTION:
            cluster.local.	300	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1537713959 7200 1800 86400 30
            
        address_test.go:128: coredns log: .:53
            2018/09/23 14:45:59 [INFO] CoreDNS-1.2.2
            2018/09/23 14:45:59 [INFO] linux/amd64, go1.11, bc454373
            CoreDNS-1.2.2
            linux/amd64, go1.11, bc454373
            172.17.0.6:38636 - [23/Sep/2018:14:46:05 +0000] 5556 "A IN svc-1-a.test-1.svc.cluster.local. udp 74 false 4096" NXDOMAIN qr,aa,rd,ra 167 0.000115883s
    --- FAIL: TestKubernetesA/headless-svc.test-1.svc.cluster.local._A (0.31s)
        helpers.go:118: Rcode is "NXDOMAIN", expected "NOERROR"
        helpers.go:301: ;; opcode: QUERY, status: NXDOMAIN, id: 25532
            ;; flags: rd; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0
            
            ;; QUESTION SECTION:
            ;headless-svc.test-1.svc.cluster.local.	IN	 A
            
            ;; AUTHORITY SECTION:
            cluster.local.	300	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1537713959 7200 1800 86400 30
            
        address_test.go:128: coredns log: .:53
            2018/09/23 14:45:59 [INFO] CoreDNS-1.2.2
            2018/09/23 14:45:59 [INFO] linux/amd64, go1.11, bc454373
            CoreDNS-1.2.2
            linux/amd64, go1.11, bc454373
            172.17.0.6:38636 - [23/Sep/2018:14:46:05 +0000] 5556 "A IN svc-1-a.test-1.svc.cluster.local. udp 74 false 4096" NXDOMAIN qr,aa,rd,ra 167 0.000115883s
            172.17.0.6:38110 - [23/Sep/2018:14:46:05 +0000] 37904 "A IN headless-svc.test-1.svc.cluster.local. udp 79 false 4096" NXDOMAIN qr,aa,rd,ra 172 0.000058161s
    --- PASS: TestKubernetesA/bogusservice.test-1.svc.cluster.local._A (0.12s)
    --- PASS: TestKubernetesA/bogusendpoint.svc-1-a.test-1.svc.cluster.local._A (0.13s)
    --- FAIL: TestKubernetesA/svc-1-a.*.svc.cluster.local._A (0.31s)
        helpers.go:118: Rcode is "NXDOMAIN", expected "NOERROR"
        helpers.go:301: ;; opcode: QUERY, status: NXDOMAIN, id: 34017
            ;; flags: rd; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0
            
            ;; QUESTION SECTION:
            ;svc-1-a.*.svc.cluster.local.	IN	 A
            
            ;; AUTHORITY SECTION:
            cluster.local.	300	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1537713959 7200 1800 86400 30
            
 ~~~
It's all NXDOMAIN, which means something is not populated or I've dropped something, somewhere.

chrisohaver:
Have not reviewed code yet, but Pod probably isn't the best place to start, since no Pod cache is maintained by default (we default to `pods insecure`, which does not maintain a watch on pods).   But this would be nice to have for people who change that default setting. Endpoint and Service cache are currently always enabled - these are the ones affecting the cache in the tests.

chrisohaver:
> Have not reviewed code yet, but Pod probably isn't the best place to start

Nevermind that - I see you've done this for Services and Endpoints too...

chrisohaver (dismissed):
Cool!  We'll want to do this for Endpoint too of course. It's probable that the Endpoints have the largest default store.  Row-wise, it's the same as Services, but each Endpoints api object can contain many actual endpoints.  Pods would be the largest store, but in most cases the pod store is disabled.

chrisohaver:
> It's all NXDOMAIN, which means something is not populated or I've dropped something, somewhere.

Since wildcard queries work, and full name queries don't, it may be the indexing thats broken (or index lookups)...  That is, wildcard queries use the `List` funcs (they cant use the indexes), and literal name lookups use the indexes.

miekg:
>> It's all NXDOMAIN, which means something is not populated or I've dropped something, somewhere.
>
>Since wildcard queries work, and full name queries don't, it may be the indexing thats broken (or index lookups)...  That is, wildcard queries use the `List` funcs (they cant use the indexes), and literal name lookups use the indexes.

Ah, ok. And that where the indexes come in. I'll take another look; 'cause I 
though it took care of that.

Is there any of these tests that can be easily ported to be a unit test? Right 
now this only fails on CI :/

chrisohaver:
> Is there any of these tests that can be easily ported to be a unit test? 

Not easily.  Otherwise they'd already be unit tests.

chrisohaver:
@miekg - Functioning queries don't matter for the k8s e2e scale tests, so we can run this through now, to get a sneak peak of how much memory we shaved off.  I would like to get the Endpoints part in though.

Do you mind if I trim down the endpoints in this PR?  Or would you prefer to do them?

chrisohaver:
> Since wildcard queries work, and full name queries don't,

Scratch that... some wildcard tests are also failing... so that's not the pattern...

miekg:
>@miekg - Functioning queries don't matter for the k8s e2e scale tests, so we can run this through now, to get a sneak peak of how much memory we shaved off.  I would like to get the endpoint part in though.
>
>Do you mind if I migrate the endpoints in this PR?  Or would you prefer to do them?

So I have several question on this PR

this code:

~~~
func (dns *dnsControl) ServiceList() (svcs []*object.Service) {
	os := dns.svcLister.List()
	for _, o := range os {
		s, ok := o.(*object.Service)
		if !ok {
			continue
		}
		svcs = append(svcs, s)
	}
	return svcs
}
~~~

Are I'm actually getting *object.Service types here, or could they still be 
*api.Service and need to be converted? I.e. this is working of the cache or 
before.

Porting endpoints is super easy, once I understand a bit more here.

miekg:
Uhm ci-test don't fail anymore @chrisohaver

chrisohaver:
> Uhm ci-test don't fail anymore

@miekg, when you are ready, we can test the memory usage of this build at scale.
Let me know, and I can pull this PR down, build the docker image, push it to my repo, and then re-run the 500 and 2000 node scale tests in the open PR we have in kubernetes/kubernetes.  

The currently running 2000 node test should be done in a few hours.

miekg:
Might be hitting the same race you talked about. Any case: another pass
~~~
(+/-) Start local proxy (for out-of-cluster tests)
(+/-) Integration tests
(+/-) Clean up
(+/-) Test Summary
Integration test passed 1 tests and 10 subtests.
~~~

miekg:
@chrisohaver please do; this should work (not sure if I'm hitting a race or not). I'll can do endpoints, but it would be good to get some data out of it.

(endpoints are more of the same)

miekg:
yep, i'ts racy; but the same records fail when it fails, huh?

miekg:
@johnbelamaric do you have time to look into this PR and possibly understand why this is now racy. The client-go caching code is a dense wood of empty interfaces and abstractions

chrisohaver:
e2e tests, as the code is now, show that this PR decreased memory slightly at the 500 node test, but increased memory more significantly at the 2000 node test.

500 nodes: 30.543872 MB
2000 nodes: 103.923712 MB

Also, given the 0 load base point starting point at 19MB, the growth doesn't seem as linear as the other series/builds.  This could perhaps indicate a leak (more pronounced at higher scale).

[chart](https://docs.google.com/spreadsheets/d/e/2PACX-1vRJCdyf82bXGwox34qdzKk5NnPhvdVfPMl8skJH7lLU9NbP8EZhQpN22gH8rOMn1RxcC8auzER7GL8K/pubchart?oid=626024915&format=interactive) (this series is "custom store")
[2000 node test](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/68938/pull-kubernetes-e2e-gce-large-performance/17/artifacts/ResourceUsageSummary_load_2018-09-28T08:37:55Z.json)
[500 node test](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/68938/pull-kubernetes-e2e-gce-big-performance/19/artifacts/ResourceUsageSummary_load_2018-09-28T04:36:52Z.json)

johnbelamaric:
@miekg I am looking. Nothing jumps out at me at first glance. It looks all mechanical except for the additional files and https://github.com/coredns/coredns/pull/2128/files#diff-4cd0f3dda92302a24e69782c6a691439R533

But that removal of DeepCopy looks like it's OK to me.

johnbelamaric:
I wonder if it's always been racy but some change here (doing conversions? that should be fast...) just brought it out. In https://github.com/coredns/ci/blob/master/build/kubernetes/minikube_setup.sh we just have a short sleep. Maybe we need to check a few records or something via kubectl and make sure they are there and look like what we expect, before finishing that script.

johnbelamaric:
If we can stop the CI in the middle and examine the minikube instance that could help too.

johnbelamaric:
Hmm. Not sure it's related, but there is a bug here. GetResourceVersion always returns "" so this line https://github.com/coredns/coredns/pull/2128/files#diff-4cd0f3dda92302a24e69782c6a691439L552 will always return immediately.

johnbelamaric:
You know, there could be a similar check inside the client-go code that causes some update not to be stored, which could cause these errors.

miekg:
@chrisohaver I pushed the endpoints change as well. Looks like there isn't any *more* breakage due to that at least.

chrisohaver:
> I pushed the endpoints change as well. Looks like there isn't any more breakage due to that at least.

Cool.  I'll build and kick off tests.

miekg:
sorry had to force push (thereby negating @johnbelamaric changes)

johnbelamaric:
Doesn't seem to fix anything :(

miekg:
trying to figure out what these failures have in common - at least it is now always failing.

miekg:
If this PR doesn't show a massive decrease in mem use I'm ready to drop it

johnbelamaric:
w00t!!!

johnbelamaric:
@chrisohaver can you try this one out on the 2k nodes?

miekg:
ah nice. I still think your ResourceVersion change was good as well; or are we exercising that code path here in ci? Something gets updated, put it in the cache?

miekg:
scratch that; that change is also in.

miekg:
rebased, squashed. PTAL

and let's wait for scale test.

miekg:
Note that the Endpoints aren't actually any smaller (almost all fields are included in objects.Endpoint)

johnbelamaric:
I believe there are lots of things not included, like labels, etc. if you
%v them you will see a big difference.

miekg:
according to travis this didn't make a difference (which is understandable because we're not using the client-go API there).
~~~
benchmark                       old allocs     new allocs     delta
BenchmarkCacheResponse-2        29             29             +0.00%
BenchmarkFileLookupDNSSEC-2     194            194            +0.00%
BenchmarkFileParseInsert-2      678            678            +0.00%
BenchmarkFileLookup-2           64             64             +0.00%
BenchmarkController-2           30             30             +0.00%
BenchmarkCache-2                0              0              +0.00%
BenchmarkMinimalTTL-2           0              0              +0.00%
BenchmarkRequestDo-2            0              0              +0.00%
BenchmarkRequestSize-2          0              0              +0.00%
BenchmarkRequestScrub-2         99             99             +0.00%
BenchmarkCoreServeDNS-2         5              5              +0.00%
benchmark                       old bytes     new bytes     delta
BenchmarkCacheResponse-2        512           512           +0.00%
BenchmarkFileLookupDNSSEC-2     6720          6720          +0.00%
BenchmarkFileParseInsert-2      116321        116321        +0.00%
BenchmarkFileLookup-2           2064          2064          +0.00%
BenchmarkController-2           1600          1600          +0.00%
BenchmarkCache-2                0             0             +0.00%
BenchmarkMinimalTTL-2           0             0             +0.00%
BenchmarkRequestDo-2            0             0             +0.00%
BenchmarkRequestSize-2          0             0             +0.00%
BenchmarkRequestScrub-2         5808          5808          +0.00%
BenchmarkCoreServeDNS-2         160         
~~~

bradbeam:
Kind of related, with #2107  merged, do we want to shift away from using `APIConnServeTest` and use client-go/fake and newdnsController? I'm expecting it would simplify the tests a bit, but seems like it might be a fair amount of legwork not directly related to this.

plugin/kubernetes/object/endpoint.go
```diff
@@ -0,0 +1,146 @@
...
+	Hostname      string
+	NodeName      string
+	TargetRefName string
+}
+
+// EndpointPort is a tuple that describes a single port.
+type EndpointPort struct {
+	Port     int32
+	Name     string
+	Protocol string
```
bradbeam, line 39: Since this will only ever be two values - TCP or UDP - does it make sense to map it to an int/bool? Or is that adding more complexity than it's worth?

plugin/kubernetes/object/endpoint.go
```diff
@@ -0,0 +1,146 @@
...
+		e1.Subsets = append(e.Subsets, sub)
+	}
+	return e1
+}
+
+// GetNamespace implements the metav1.Object interface.
+func (e *Endpoints) GetNamespace() string { return e.Namespace }
+
+// SetNamespace implements the metav1.Object interface.
+func (e *Endpoints) SetNamespace(namespace string) {}
```
bradbeam, line 150: Should these actually do anything or are they only there to satisfy the interface?

plugin/kubernetes/kubernetes.go
```diff
@@ -396,12 +397,12 @@ func (k *Kubernetes) findPods(r recordRequest, zone string) (pods []msg.Service,
 		}
 
 		// exclude pods in the process of termination
-		if !p.ObjectMeta.DeletionTimestamp.IsZero() {
+		if !p.DeletionTimestamp.IsZero() {
```
bradbeam: Should this check get moved to the object/pod.go object.topod conversion instead of saving the deletion timestamp?

miekg:
>+// EndpointPort is a tuple that describes a single port.
>+type EndpointPort struct {
>+	Port     int32
>+	Name     string
>+	Protocol string
>
>Since this will only ever be two values - TCP or UDP - does it make sense to map it to an int/bool? Or is that adding more complexity than it's worth?

Not sure how bools are implemented in the lower levels, but if they occupy a 
byte we don't save much and we compare against the string in kubernetes.go, so I 
don't think we'll gain much here.

/Miek

--
Miek Gieben

miekg:
>Kind of related, with #2107  merged, do we want to shift away from using `APIConnServeTest` and use client-go/fake and newdnsController? I'm expecting it would simplify the tests a bit, but seems like it might be a fair amount of legwork not directly related to this.

The less code *we* maintain, the better it is.

miekg:
>bradbeam commented on this pull request.
>
>
>
>> @@ -396,12 +397,12 @@ func (k *Kubernetes) findPods(r recordRequest, zone string) (pods []msg.Service,
> 		}
>
> 		// exclude pods in the process of termination
>-		if !p.ObjectMeta.DeletionTimestamp.IsZero() {
>+		if !p.DeletionTimestamp.IsZero() {
>
>Should this check get moved to the object/pod.go object.topod conversion instead of saving the deletion timestamp?

possibly.

chrisohaver:
The memory tests results are better.  It drops total memory usage by about 10% vs the full api store.  This is lightly lower than total kube-dns usage.  It still projects to come in slightly higher than the currently required memory limit at 5K, but we may be able to lobby successfully to have the limit raised (due to the fact that for kube-dns the limit only applies to one of its 3 containers).

[chart](https://docs.google.com/spreadsheets/d/e/2PACX-1vRJCdyf82bXGwox34qdzKk5NnPhvdVfPMl8skJH7lLU9NbP8EZhQpN22gH8rOMn1RxcC8auzER7GL8K/pubchart?oid=626024915&format=interactive) (this series is "custom store")

I have kicked off a discussion of comparing the the various memory reduction efforts [here](https://github.com/coredns/coredns/issues/2095#issuecomment-425901771).

plugin/kubernetes/kubernetes.go
chrisohaver: makes sense to me.

plugin/kubernetes/object/service.go
```diff
@@ -0,0 +1,87 @@
...
+		Namespace:    svc.GetNamespace(),
+		Index:        ServiceKey(svc.GetName(), svc.GetNamespace()),
+		ClusterIP:    svc.Spec.ClusterIP,
+		Type:         svc.Spec.Type,
+		ExternalName: svc.Spec.ExternalName,
+	}
+
+	ports := len(svc.Spec.Ports)
+	if ports == 0 {
+		s.Ports = []api.ServicePort{{Port: -1}}
```
chrisohaver, line 44: IMO, should comment here as to why this is done (sentinel for services with no declared ports).

chrisohaver (commented):
LGTM - one minor nit.

plugin/kubernetes/object/endpoint.go
johnbelamaric: Actually SCTP is coming, and so we should leave these as strings. I don't want to have to update an enum to track K8s.

plugin/kubernetes/object/endpoint.go
johnbelamaric: they are just there to satisfy the interface but of course having them and them being non-functional is misleading

johnbelamaric:
Only 10%? That's disappointing. I really expected more. Are we sure we don't have any dangling references to anything? Maybe we can make it more compact somehow?

@chrisohaver you said some changes to not call `sendUpdates` made a big difference? Can we look at why? As I recall there was a `DeepCopy` in there, but @miekg replaced it with something else.

miekg:
I'll work on the nits.

10% does sound low. I'll double check if I've missed something.

The deep copy in send updates is replaced with a less deep copy, that
functions still looks expensive (but haven't measured) though

chrisohaver:
> you said some changes to not call sendUpdates made a big difference?

It makes a difference, I didn't say it was big ...  I think ~ 4-5%.

> Can we look at why?

I don't know for sure what the architecture is of the client-go informer, but if there is a client side receive message queue, and it processes messages in order,  then the extra work done in the ResourceEventHandlerFuncs could be slowing processing down, causing the queue to grow, consuming memory.

plugin/kubernetes/object/service.go
miekg: done

miekg:
not using `append` means we not doubling the size on every append that goes over cap; prolly with endpoints this leads to a lot of wasted bytes.

I might make a typo that makes CI fail again (although I double checked)

plugin/kubernetes/object/endpoint.go
```diff
@@ -54,33 +54,43 @@ func ToEndpoints(obj interface{}) interface{} {
...
-		for _, a := range eps.Addresses {
+	for i, eps := range end.Subsets {
+		sub := EndpointSubset{
+			Addresses: make([]EndpointAddress, len(eps.Addresses)),
+		}
+
+		if len(eps.Ports) == 0 {
+			sub.Ports = []EndpointPort{{Port: -1}}
+		} else {
+			// Add sentinal if there are no ports.
```
johnbelamaric: Comment is in the wrong place...

plugin/kubernetes/kubernetes.go
miekg: done

plugin/kubernetes/object/pod.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil
+	}
+
+	p := &Pod{
+		Version:   pod.GetResourceVersion(),
+		PodIP:     pod.Status.PodIP,
+		Namespace: pod.GetNamespace(),
+		Name:      pod.GetName(),
+	}
+	t := pod.ObjectMeta.DeletionTimestamp
```
bradbeam, line 32: Is it worth moving this check above where we define the pod ( line 26 ) and skip adding the pod if it's terminating?
Or rather, do we actually need to store the pod information if it's in a deleting state?

miekg:
>+// ToPod converts an api.Pod to a *Pod.
>+func ToPod(obj interface{}) interface{} {
>+	pod, ok := obj.(*api.Pod)
>+	if !ok {
>+		return nil
>+	}
>+
>+	p := &Pod{
>+		Version:   pod.GetResourceVersion(),
>+		PodIP:     pod.Status.PodIP,
>+		Namespace: pod.GetNamespace(),
>+		Name:      pod.GetName(),
>+	}
>+	t := pod.ObjectMeta.DeletionTimestamp
>
>Is it worth moving this check above where we define the pod ( line 26 ) and skip adding the pod if it's terminating?
>Or rather, do we actually need to store the pod information if it's in a deleting state?

Prolly best to do further tweaks in follow PRs

miekg:
ok, added explicit hack to set the original object to nil when we're done converting. This *should* trigger GC at some point.
Note that if you store elements in a map, that map is never GC-ed if you delete from it - you need to copy everything to a new map to clear the old (prometheus does tricks like this in their code)

chrisohaver:
> Note that if you store elements in a map, that map is never GC-ed if you delete from it ...

You mean the map _value_ is not GCed or the map _key_ is not GCed? Or both?

What if the map element is a pointer (e.g. `map[string]*obj` ), and the pointer is set to nil (e.g. `m["a"] = nil`)?  Will the object pointed to get GCed?

plugin/kubernetes/object/service.go
```diff
@@ -47,6 +47,8 @@ func ToService(obj interface{}) interface{} {
 		copy(s.Ports, svc.Spec.Ports)
 	}
 
+	svc = nil
```
chrisohaver: Can you help me understand this better?  ...

Per my understanding, `svc` is a local scope pointer ... nothing points to _it_, so it should get allocated on the stack.  It goes away when `ToService(obj)` pops off the stack.  The object `svc` points to is still pointed to by `obj` (which is also on the stack and will go away when the function ends).  The object pointed to would still be pointed to by the caller to `ToService(obj)` ... So, I don't see how setting `svc` to nil here makes any difference.

miekg:
>
>
>
>> @@ -47,6 +47,8 @@ func ToService(obj interface{}) interface{} {
> 		copy(s.Ports, svc.Spec.Ports)
> 	}
>
>+	svc = nil
>
>Can you help me understand this better?  ...
>
>Per my understanding, `svc` is a local scope pointer ... nothing points to _it_, so it should get allocated on the stack.  It goes away when `ToService(obj)` pops off the stack.  The object `svc` points to is still pointed to by `obj` (which is also on the stack and will go away when the function ends).  The object pointed to would still pointed to by the caller to `ToService(obj)` ... So, I don't see how setting it to nil here makes any difference.

that's... very much a correct statement. I need to dereference it, and nil that

miekg:
Can we just this PR again on the scale infra? To see if its worth it?

Thanks!

chrisohaver:
Bumping up against quota, only the 2k node test completed... results are much better!  67.6 MB at 2000 node test. Thats 27% less memory than the original client-go api object store.

miekg:
>Bumping up against quota, only the 2k node test completed... results are much better!  67.6 MB at 2000 node test. Thats 27% less memory than the original client-go api object store.

thanks!

27% at least bring us slightly under kube-dns!

chrisohaver:
> 27% at least bring us slightly under kube-dns!

At the 2000 node test, this PR uses 32% less than kube-dns. IMO, that's better than "slightly under".

CoreDNS has always used less memory than kube-dns in the 2000 node test.

Regarding 5k node tests: We don't actually have real 5k node test results.  All 5k numbers in the charts are linear projections of 0/500/2000 node test results.

fturib:
That is a great result !

@miekg : for the big picture, do you have an idea of size or work that remains here for a merge ? Is there some parts we can help on ?

fturib:
Unfortunately, when it comes to merge CoreDNS change into k8s, we now need a release.
And we need to validate the optimization part way before the freeze time so we can have planB if the straightforward plan fails during the test.

Which means, an expectation of having CoreDNS by 10/12 to 10/15, then image available on gcr.io by 10/16 and a merge CoreDNS deployment (with new image, new mem limit for test) right after.

I created a Milestone 1.2.3 for the release of CoreDNS needed for K8s v1.13. that includes 2 tickets about optimization.

@miekg : in order to get close to this timeline, how can we help here ?



SIDE NOTE: there is an invite for SIG-sclalability on 10/11 at 16h30 UTC for an presentation of our optimization work, and talk on how much we can bump the memory - if we can.

miekg:
>That is a great result !
>
>@miekg : for the big picture, do you have an idea of size or work that remains here for a merge ? Is there some parts we can help on ?

This PR is done.

miekg:
>Unfortunately, when it comes to merge CoreDNS change into k8s, we now need a release.
>And we need to validate the optimization part way before the freeze time so we can have planB if the straightforward plan fails during the test.
>
>Which means, an expectation of having CoreDNS by 10/12 to 10/15, then image available on gcr.io by 10/16 and a merge CoreDNS deployment (with new image, new mem limit for test) right after.
>
>I created a Milestone 1.2.3 for the release of CoreDNS needed for K8s v1.13. that includes 2 tickets about optimization.
>
>@miekg : in order to get close to this timeline, how can we help here ?

We need a release anyway - master should be always in releasable state - and it 
is, so we can just kick off one.

chrisohaver (approved):

miekg:
Here goes nothing peeps!

Issue #2114, opened 2018-09-20
The date format in logs should be the same


eugen-kleiner-mesh:
To reproduce you can start coredns with the config which forwards plugin which points to not existing  dns server.
Corefile:
```
.:53 {
    bind 127.0.0.2
    forward . 127.0.0.54:53
    log
    errors
}
```
Run ```dig github.com @127.0.0.2``` and see the logs from coredns:
```
127.0.0.1:42936 - [21/Sep/2018:00:09:46 +0300] 15164 "A IN github.com. udp 52 false 4096" SERVFAIL qr,rd 52 1.000780352s
2018/09/21 00:09:46 [ERROR] 0 github.com. A: read udp 127.0.0.1:59513->127.0.0.54:53: read: connection refused
```
```21/Sep/2018:00:09:46 +0300``` differs from ```2018/09/21 00:09:46``` though it is the same date.

eugen-kleiner-mesh:
All plugins use "github.com/coredns/coredns/plugin/pkg/log" to output logs, but log plugin uses different way. So, I think it should be fixed either in log plugin or in log pkg.

miekg:
Good point. Would be better if that was unified.

fturib:
### Re-install the TimeZone in the time format.

Please, review the change of code : the TimeZone has been shift out, and I think it is important to be able to compare the date with other logs ...

FYI - @ynikanchyk, @ibvrubleg

miekg:
Looking *again* at this change. Dropping of the timezone is the wrong thing today and still leaves the date format being different.

Also in `errors/errors.go`:
~~~
clog.Errorf("%d %s %s: %v", rcode, state.Name(), state.Type(), err)
~~~
`rcode` isn't the rcode but the return value from the handler which is something different. Should probably remove that as well.

rdrozhdzh:
I would prefer the time format like RFC3339 (`2006-01-02T15:04:05Z07:00`) to be used by all plugins, i.e. it should be implemented in` plugin/pkg/log` instead of standard go log format.

And all plugins (including log and errors) should use `plugin/pkg/log` package for printing logs

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] The date form..." ]
>I would prefer the time format like RFC3339 (`2006-01-02T15:04:05Z07:00`) to be used by all plugins, i.e. it should be implemented in` plugin/pkg/log` instead of standard go log format.
>
[...]

PR #2135, opened 2018-09-26
plugin/log: Change time format to be consistent with pkg/log

eugen-kleiner-mesh:
Changes time format in log plugin logs to be consistent with pkg/log
https://github.com/coredns/coredns/issues/2114

### DIFF
--- a/plugin/pkg/replacer/replacer.go
+++ b/plugin/pkg/replacer/replacer.go
@@ -164,6 +164,6 @@ func addrToRFC3986(addr string) string {
 }
 
 const (
-	timeFormat     = "02/Jan/2006:15:04:05 -0700"
+	timeFormat     = "2006/01/02 15:04:05"
 	headerReplacer = "{>"
 )


miekg:
The log plugin has a runtime option to set the output, so we can't statically 
set it something as fix. We need to plugin *into* the log plugin to get the 
current format and use that.

eugen-kleiner-mesh:
@miekg, this is the format which is used by log plugin. See:
https://github.com/coredns/coredns/blob/master/plugin/log/log.go#L61 and
https://github.com/coredns/coredns/blob/master/plugin/pkg/replacer/replacer.go#L43
Could you clarify your comment if I misunderstood something?

eugen-kleiner-mesh:
In fact, pkg/log plugin can be remade to use the constant from pkg/format. But in this case Logger instance should be created in pkg/log.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/log: C..." ]
>This is the format which is used by log plugin. See:
>https://github.com/coredns/coredns/blob/master/plugin/log/log.go#L61 and
>https://github.com/coredns/coredns/blob/master/plugin/pkg/replacer/replacer.go#L43
>Could you clarify your comment if I misunderstood something?

My bad; you can only override the contents of the log: https://github.com/coredns/coredns/blob/master/plugin/log/README.md#log-format
not the actual logging prefix

>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/coredns/coredns/pull/2135#issuecomment-424861450
/Miek

--
Miek Gieben

PR #2136, opened 2018-09-26
Use newer Prometheus API (fixes impossibility to build CoreDNS right after 'go get')

ghost:
CoreDNS can't be built now:
`# github.com/coredns/coredns/plugin/metrics
plugin\metrics\metrics.go:43:49: too many arguments in call to prometheus.NewProcessCollector
        have (int, string)
        want (prometheus.ProcessCollectorOpts)`

Signature of the function from the dependency was changed. This PR fixes it by using [newer](https://godoc.org/github.com/prometheus/client_golang/prometheus#NewProcessCollector) one.

### DIFF
--- a/plugin/metrics/metrics.go
+++ b/plugin/metrics/metrics.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"net"
 	"net/http"
-	"os"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ghost:
Seems like an older version of the dependency on the build server is used, so it has failed.

rdrozhdzh:
see related PR #2091

jameshartig:
This is a duplicate of #2091. See the discussion in that issue. We won't be updating this until v0.9.0 has been released.

PR #2137, opened 2018-09-27
WIP: Less mem ii

miekg:


miekg:
Well using our own metaNamespaceKeyFunc (in this fashion) breaks everything /cc @chrisohaver

chrisohaver:
IMO, the benchmarks reported as comments seem kind of obtrusive. We should just add a status to the checks, with a link to the travis output, so it shows up as a link in the checks.

miekg:
If you know how and then please do.

PR #2140, opened 2018-09-27
Add code block in benchcmp output

bradbeam:

Adds some formatting to the benchmark results on PR

### DIFF
--- a/.benchmark.sh
+++ b/.benchmark.sh
@@ -5,7 +5,10 @@ set -e +o pipefail
 if [ "$TRAVIS_PULL_REQUEST" != "false" ] ; then
     echo -e "NOTE: The CPU benchmarks are performed on Travis VMs and vary widly between runs," > .benchmark.body
     echo -e " you can't trust them. The memory benchmarks are OK\n\n" >> .benchmark.body
-    cat .benchmark.log >> .benchmark.body
+    awk '/^benchmark old/ { printf "%s\n%s\n", "```", $0 };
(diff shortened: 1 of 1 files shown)


.benchmark.sh
```diff
@@ -5,7 +5,7 @@ set -e +o pipefail
 if [ "$TRAVIS_PULL_REQUEST" != "false" ] ; then
     echo -e "NOTE: The CPU benchmarks are performed on Travis VMs and vary widly between runs," > .benchmark.body
     echo -e " you can't trust them. The memory benchmarks are OK\n\n" >> .benchmark.body
-    cat .benchmark.log >> .benchmark.body
+    cat .benchmark.log | awk '/^benchmark old/ { printf "%s\n%s\n", "```", $0 }; /^$/ { print "```" }; /^Bench/ { print $0 }; END{ print "```" }' >  .benchmark.body
```
miekg: had the same thought :)

However you need a `>>` instead of `>` because now you're overwritting .benchmark.body ?

bradbeam: whoops.

miekg:
it's doing it almost right: https://github.com/coredns/coredns/pull/2137#issuecomment-425136218

bradbeam:
Well that's just odd.  Looks like the headers are gone / missing.  Maybe I just copied old results.

bradbeam:
Yeah that was it. I copied from the comment versus editing and copying there. The whitespace was off.

PR #2143, opened 2018-09-27
WIP plugin/kubernetes: use less mem - reduce index size

chrisohaver:

Reduce the amount of memory used for indexes...

- Don't create indexes for headless service endpoints (big): Removes reverse endpoint Index, Replace with a smaller manually maintained map of ip -> endpoint for endpoints of headless services only.
- Remove name.namespace forward index for endpoint and services (small): Use built in namespace/name key and GetByKey() instead of Index.

#2095 

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -419,7 +419,11 @@ func (dns *dnsControl) EpIndex(key string) (*api.Endpoints) {
 
 func (dns *dnsControl) EpIndexReverse(ip string) (ep *api.Endpoints) {
 	if !dns.revIdxAllEndpoints {
-		return dns.EpIndex(*dns.headlessEndpoints.keys[ip])
+		key := dns.headlessEndpoints.keys[ip]
+		if key == nil {
+			return nil
+		}
+		return dns.EpIndex(*key)
 	}
 
 	if dns.epLister == nil {
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -367,7 +367,7 @@ func (dns *dnsControl) PodIndex(ip string) (pods []*api.Pod) {
 	return nil
 }
 
-func (dns *dnsControl) SvcIndex(key string) (*api.Service) {
+func (dns *dnsControl) SvcIndex(key string) *api.Service {
 	o, _, err := dns.svcLister.GetByKey(key)
 	if err != nil {
 		return nil
@@ -380,7 +380,7 @@ func (dns *dnsControl) SvcIndex(key string) (*api.Service) {
 	return s
 }
 
-func (dns *dnsControl) SvcIndexReverse(ip string) (*api.Service) {
+func (dns *dnsControl) SvcIndexReverse(ip string) *api.Service {
 	if dns.svcLister == nil {
 		return nil
 	}
@@ -399,7 +399,7 @@ func (dns *dnsControl) SvcIndexReverse(ip string) (*api.Service) {
 	return nil
 }
 
-func (dns *dnsControl) EpIndex(key string) (*api.Endpoints) {
+func (dns *dnsControl) EpIndex(key string) *api.Endpoints {
 	if dns.epLister == nil {
 		return nil
 	}
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -27,12 +27,14 @@ const (
 
 type dnsController interface {
 	ServiceList() []*api.Service
-	SvcIndex(string) []*api.Service
-	SvcIndexReverse(string) []*api.Service
-	PodIndex(string) []*api.Pod
-	EpIndex(string) []*api.Endpoints
-	EpIndexReverse(string) []*api.Endpoints
 	EndpointsList() []*api.Endpoints
+
+	SvcIndex(string) *api.Service
+	PodIndex(string) []*api.Pod
+	EpIndex(string) *api.Endpoints
+
+	SvcIndexReverse(string) *api.Service
+	EpIndexReverse(string) *api.Endpoints
 
 	GetNodeByName(string) (*api.Node, error)
 	GetNamespaceByName(string) (*api.Namespace, error)
@@ -360,12 +362,12 @@ func (dns *dnsControl) PodIndex(ip string) (pods []*api.Pod) {
 		if !ok {
 			continue
 		}
-		pods = append(pods, p)
+		return []*api.Pod{p}
 	}
-	return pods
+	return nil
 }
 
-func (dns *dnsControl) SvcIndex(key string) (svcs []*api.Service) {
+func (dns *dnsControl) SvcIndex(key string) (*api.Service) {
 	if dns.svcLister == nil {
 		return nil
 	}
@@ -374,16 +376,14 @@ func (dns *dnsControl) SvcIndex(key string) (svcs []*api.Service) {
 		return nil
 	}
 	s, ok := o.(*api.Service)
-
 	if !ok {
-		return
+		return nil
 	}
-	svcs = append(svcs, s)
 
-	return svcs
+	return s
 }
 
-func (dns *dnsControl) SvcIndexReverse(ip string) (svcs []*api.Service) {
+func (dns *dnsControl) SvcIndexReverse(ip string) (*api.Service) {
 	if dns.svcLister == nil {
 		return nil
 	}
@@ -397,12 +397,12 @@ func (dns *dnsControl) SvcIndexReverse(ip string) (svcs []*api.Service) {
 		if !ok {
 			continue
 		}
-		svcs = append(svcs, s)
+		return s
 	}
-	return svcs
+	return nil
 }
 
-func (dns *dnsControl) EpIndex(key string) (ep []*api.Endpoints) {
+func (dns *dnsControl) EpIndex(key string) (*api.Endpoints) {
 	if dns.epLister == nil {
 		return nil
 	}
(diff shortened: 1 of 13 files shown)


chrisohaver:
Todo:
- make it work :)
- expose option to enable reverse lookups of all endpoint IPs?
- rename `SvcIndex()` and `EpIndex()` to `SvcByKey()` and `EpByKey()` (and make them return one item instead of a list)?

chrisohaver:
Looks like the coredns/ci test failure is because the test case does a reverse lookup on an endpoint of a clusterIP service, which is no longer supported with these proposed changes.

chrisohaver:
travis unit test failures are probably due name/namespace retrieval

chrisohaver:
Memory tests in progress: 
[500 Node]( https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/68938/pull-kubernetes-e2e-gce-big-performance/18/)
[2000 Node]( https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/68938/pull-kubernetes-e2e-gce-large-performance/16/)

chrisohaver:
Scratching head... the memory tests went impossibly well... something doesn't add up...

chrisohaver:
> Scratching head... the memory tests went impossibly well... something doesn't add up...

So the results came in at significantly less than no index at all - which is sort of impossible...  unless...
- I did make an optimization to watch callback functions, which were executing even when no watches were active.  I just added a simple check to make sure there are > 0 watches before processing them.  This is mainly a CPU optimization, processing endpoint change notifications to client is expensive. But I suppose there could be some degree of memory leakage in those routines, which are now getting skipped in the scale e2e tests since no watches are active during those tests.
- ~~Parallel work going on for memory reduction - maybe something merged, and i'm seeing that.~~  No, i don't see anything in the recent commit history

miekg:
Nothing has been merged to master wrt to the memory work.

chrisohaver:
> I did make an optimization to watch callback functions,

OK, rolling the watch optimization back out accounted for 40% of the difference.  The remaining 60% remains a mysterious gift.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP plugin/ku..." ]
>> I did make an optimization to watch callback functions,
>
>OK, rolling the watch optimization back out accounted for 40% of the difference.  The remaining 60% remains a mysterious gift.

so the *experimental* watch feature accounts for a lot of memory? Can we back 
out that feature entirely?

chrisohaver:
>so the *experimental* watch feature accounts for a lot of memory? Can we back
out that feature entirely?

If client watches are connected, assuming linear growth, trajectory lands it at 12 MB during the 5k node scale (~5% total memory).  This is just to process the api events and determine if we need to send updates to clients.  Actual active client connections would take up some addtl memory.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP plugin/ku..." ]
>>so the *experimental* watch feature accounts for a lot of memory? Can we back
>out that feature entirely?
>
>If client watches are connected, assuming linear growth, trajectory lands it at 12 MB during the 5k node scale (~5% total memory).  This is just to process the api events and determine if we need to send updates to clients.  Actual active client connections would take up some addtl memory.

ok, that's quite significant, and noone is using this in k8s as it stands right 
now.

chrisohaver:
> ok, that's quite significant, 

Perhaps yes.  Although this is a scale test, i'm not sure how fast the API events actually come in. The speed of the test may be gated by slow compute nodes.  The 2000 node spins up and tears down 60K pods in ~3 hours.  60000 / (3*60*60) seconds = 6 changes/s.

Safe play is to disable.

johnbelamaric:
Seems though with Chris's optimization it doesn't add any overhead. I'd
rather not back it out completely, but making a change so it has no effect
when there are no watches makes sense.

However - that same code is not just for watches. It is also the code that
determines whether we tickle the serial number used for zone transfers.

miekg:
unit tests fail as well, making this easier to debug

chrisohaver:
> unit tests fail as well, making this easier to debug

I know the reason for the failures ... i just have not gotten around to fixing them.
They are (all but one) purely unit test failures.  The code itself works when I manually test it deployed into a cluster.

One "real" failure is a test case that checks for reverse cluster ip endpoint ip record - which is now invalid with these changes.
The others are (I'm pretty sure) test fixture issues.

miekg:
OK, I did make a mistake in the endpoints, rectified and pushed a fix; CI now builds and tests OK. Only the CI thingy is left.

chrisohaver:
> OK, I did make a mistake in the endpoints, rectified and pushed a fix; CI now builds and tests OK. Only the CI thingy is left.

Which PR/commit?  And what is the CI thingy?

miekg:
sorry, wrong PR :/

chrisohaver:
FYI - I ran into the service/endpoint race condition thing again - it's possible I just had a lucky streak before. But now it seems to always fail (endpoint arrives before service), when this happens, we dont know if the endpoint is for a cluster ip or not, and therefore we don't know if we should save it to the IP map.

I wrote some code that addresses this - It works, but it's not pretty.  If we get an endpoint before we know about its service, then we save the endpoint to the IP map anyways.  Then for every service that is added, we check if its a cluster ip.  If it is, we delete it's endpoint entry from the map.  It works, but this is a lot of constant overhead.

Another way to solve this might be to register the  services watch first, wait for it to be ready, and then register the endpoints watch.  If thats possible.  I think this "endpoint before service" thing is only possible when we get the initial state dumped on us - if so, waiting for service dump before getting endpoints should work.

johnbelamaric:
I think watches on endpoints and services are inherently racy. I don't
think we can rely on an ordering, even after initial setup. The pushing of
that info from the api server may be from separate go routines.

chrisohaver:
> I think watches on endpoints and services are inherently racy.

OK - then the ugly fix may be the only way to go here.

johnbelamaric:
For services, clusterIP is immutable. That means that you can't change a
service from clusterIP -> headless without deleting it and recreating it.
So, you only need to trigger your extra check for service ADD events, not
for updates. I think that should limit it to a pretty small amount of
overhead.

chrisohaver:
> ... clusterIP is immutable ... So, you only need to trigger your extra check for service ADD events.

Yep.  I only added to _Add_, for that reason.  I'll push the changes shortly.

miekg:
uhm... this `map`, is that used concurrently?

chrisohaver:
Another thing I added in the last commit is a mutex on the reverse endpoint ip map. But now not sure if it's really needed.

chrisohaver:
> uhm... this map, is that used concurrently?

I had added a mutex.  I was thinking it might be the cause of some of the raciness, but it didn't help, so I implemented the "put it in and take it out later" approach.   I left the mutex in there.

chrisohaver:
I also need to add unit tests for these changes.  We can test the endpoint maintenance routines by feeding Add()/Update()/Delete() a set of  endpoint/service objects and verifying the map.

miekg:
filed https://github.com/coredns/ci/issues/70
CoreDNS is build without `-race`.

plugin/kubernetes/controller.go
```diff
@@ -365,21 +365,21 @@ func (dns *dnsControl) PodIndex(ip string) (pods []*api.Pod) {
 	return pods
 }
 
-func (dns *dnsControl) SvcIndex(idx string) (svcs []*api.Service) {
+func (dns *dnsControl) SvcIndex(key string) (svcs []*api.Service) {
 	if dns.svcLister == nil {
```
miekg: these `== nil` check where not applied for all index functions, I think they add nothing and can be removed.

plugin/kubernetes/controller.go
```diff
@@ -365,21 +365,21 @@ func (dns *dnsControl) PodIndex(ip string) (pods []*api.Pod) {
...
-		if !ok {
-			continue
-		}
-		svcs = append(svcs, s)
+	s, ok := o.(*api.Service)
+
+	if !ok {
+		return
 	}
+	svcs = append(svcs, s)
```
miekg: this doesn't need append.
I've we decide to merge #2128 it has a whole slew of fixes here.

plugin/kubernetes/controller.go
```diff
@@ -548,6 +558,9 @@ func endpointsSubsetDiffs(a, b *api.Endpoints) *api.Endpoints {
 
 // sendUpdates sends a notification to the server if a watch is enabled for the qname.
 func (dns *dnsControl) sendUpdates(oldObj, newObj interface{}) {
+	if len(dns.watched) == 0 {
+		return
+	}
```
miekg: this would be good as a small separate PR. I've also have #2157 to disable/enable watching entire separately

plugin/kubernetes/epipmap.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	for _, os := range oldEp.Subsets {
+		for _, oa := range os.Addresses {
+			found := false
+		FindDelete:
+			for _, ns := range newEp.Subsets {
+				for _, na := range ns.Addresses {
+					if oa.IP == na.IP {
+						found = true
+						break FindDelete
```
miekg, line 106: you can just delete here.

plugin/kubernetes/epipmap.go
```diff
@@ -0,0 +1,177 @@
...
+
+func (dns *dnsControl) addEpToMap(ip string, ep *api.Endpoints) {
+	namespace := ep.GetNamespace()
+	name := ep.GetName()
+	dns.headlessEndpoints.mutex.Lock()
+	if dns.headlessEndpoints.store[ip] == nil {
+		dns.headlessEndpoints.store[ip] = make(endpoints)
+	}
+	if dns.headlessEndpoints.store[ip][namespace] == nil {
+		dns.headlessEndpoints.store[ip][namespace] = make(map[string]bool)
```
miekg: this map (and the others can be redefined as `map[string]struct{}`

plugin/kubernetes/controller.go
chrisohaver: yeah - thanks for the reminder (it's in my todos in the first comment in this PR). I intend to go in and change these once we decide this PR is a viable route ... It means changing the function signature, and its an interface function, so it's a lot of changes.

chrisohaver:
From @miekg 
> Note that if you store elements in a map, that map is never GC-ed if you delete from it - you need to copy everything to a new map to clear the old (prometheus does tricks like this in their code)

In this case, the "save it now, and remove it later" technique wont work to reduce index size without rebuilding the entire map each time we remove an element.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP plugin/ku..." ]
>From @miekg
>> Note that if you store elements in a map, that map is never GC-ed if you delete from it - you need to copy everything to a new map to clear the old (prometheus does tricks like this in their code)
>
>In this case, the "save it now, and remove it later" technique wont work to reduce index size without rebuilding the entire map each time we remove an element.

Well; you loose the value so that should be gc-ed, the map's entry will stay.
(Go may have changed this since though; don't follow core Go that closely)

plugin/kubernetes/controller.go
chrisohaver: `dns.epLister` and  `dns.podLister` and can be nil, depending on kubernetes plugin options.
`dns.svcLister` can never be nil, so we can remove the check here.

chrisohaver:
closing for now - if we need to reintroduce the endpoints map, I can re-open.

PR #2147, opened 2018-09-28
Add Kubernetes auth providers

johnbelamaric:
With the merge of the #2053, we can now use a kubeconfig file for out-of-cluster Kubernetes integrations. However, the different auth providers were not included in that PR. Add them, which requires adding new vendored libraries. Also add auto-pruning in Gopkg.toml.

None

None, these are the standard providers that client-go supports, so users will expect them to work already with #2053.


miekg:
Can you split the comments in a 'code added'/'dep updated'? Both commits now contain a lot of changed files.

johnbelamaric:
Rebased on #2151 and also split code change and dep update into two separate commits.

chrisohaver (approved):
LGTM, once the vendor issues are cleared up.

johnbelamaric:
I do not understand. This built for me on Friday.

PR #2154, opened 2018-09-29
update Caddy to v0.10.13

### LINKED ISSUE #2153: template: environment variable after go template causes panic
CoreDNS panics when a template uses a Go template and an environment variable after the template:
```
template IN ANY {
  match "^([a-zA-Z0-9]*\.){$CLUSTER_DOMAIN}\.$"
  answer "{{ .Name }} 60 IN CNAME loadbalancer.default.svc.{$CLUSTER_DOMAIN}"
  upstream
  fallthrough
}
```

The fix was released with caddy v0.10.13, but CoreDNS is still using v0.10.11: https://github.com/mholt/caddy/commit/64c18a7c6c8c3c179748486bbd06a93f15aa580c

bhcleek:
Update Caddy so that environment variables may be used after Go templates without causing a panic.

#2153

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ godeps:
 	go get -u github.com/miekg/dns
 	go get -u github.com/prometheus/client_golang/prometheus/promhttp
 	go get -u github.com/prometheus/client_golang/prometheus
-	(cd $(GOPATH)/src/github.com/mholt/caddy              && git checkout -q v0.10.11)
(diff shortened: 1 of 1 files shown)


miekg (approved):

bhcleek:
Unfortunately, that doesn't actually fix the problem; it fixes related, but slightly different problems. I'm working on a patch to Caddy now, though, that should resolve it. Until then Caddy windows-style environment variable substitution, `{%ENV_VAR%}`, can be used as a workaround.

Issue #2156, opened 2018-10-02
plugin/kubernetes: disable watch


miekg:
We should add a watch directive in kubernetes, saying you want to get watch updates. By default it is disabled and only the modified timestamp is increased so axfr works.

chrisohaver:
@miekg, Why do you think this is no longer necessary?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/kubern..." ]
>Why do you think this is no longer necessary?

I'm not going to work on it anymore

chrisohaver:
> I'm not going to work on it anymore

But someone else may ... right?

johnbelamaric:
Not sure there is any reason to - didn't you put in a small tweak to make it basically a no-op when no watches are defined?

It was the endpoints check I thought that was causing possible memory issues, but we need that for AXFR not just watch.

chrisohaver:
> didn't you put in a small tweak to make it basically a no-op when no watches are defined?

No, because of the AXFR timestamp issue.

> Not sure there is any reason to.

I was thinking someone might want to run a grpc server, and _not_ have watch enabled.  But if thats not a problem/concern then there is no need for it.

Anyways, Thanks.  I was just looking for a reason.  Perhaps I wasn't asking the question correctly.  Not trying to be a pain in the ass, but I think its a good idea to explain why we are closing tickets, so later we understand why we closed them.

PR #2157, opened 2018-10-02
plugin/kubernetes: add watch option

miekg:
Disable watch by default (takes up mem) and allows user to override by
setting watch.

Update the README.md to match.

Fixes: #2156

### DIFF
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -42,6 +42,7 @@ kubernetes [ZONES...] {
     ttl TTL
     transfer to ADDRESS...
     fallthrough [ZONES...]
+    watch
 }
 ```
 
@@ -106,6 +107,7 @@ kubernetes [ZONES...] {
 * `ignore empty_service` return NXDOMAIN for services without any ready endpoint addresses (e.g. ready pods).
   This allows the querying pod to continue searching for the service in the search path.
   The search path could, for example, include another kubernetes cluster.
+* `watch` enables updates being send to a server listening for watching on a gRPC channel.
 
 ## Health
 
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	dnswatch "github.com/coredns/coredns/plugin/pkg/watch"
@@ -97,6 +96,7 @@ type dnsControlOpts struct {
 
 	zones            []string
 	endpointNameMode bool
+	watch            bool
 }
 
 // newDNSController creates a controller for CoreDNS.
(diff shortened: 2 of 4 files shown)


plugin/kubernetes/README.md
```diff
@@ -106,6 +107,7 @@ kubernetes [ZONES...] {
 * `ignore empty_service` return NXDOMAIN for services without any ready endpoint addresses (e.g. ready pods).
   This allows the querying pod to continue searching for the service in the search path.
   The search path could, for example, include another kubernetes cluster.
+* `watch` enables updates being send to a server listening for watching on a gRPC channel.
```
johnbelamaric, line 110: How about:

`watch` enables sending updates to clients subscribed to records using the gRPC Watch streaming API call

plugin/kubernetes/controller.go
```diff
@@ -109,6 +109,14 @@ func newdnsController(kubeClient kubernetes.Interface, opts dnsControlOpts) *dns
 		zones:            opts.zones,
 		endpointNameMode: opts.endpointNameMode,
 	}
+	add := func(obj interface{}) { dns.updateModifed() }
+	del := func(obj interface{}) { dns.updateModifed() }
+	upd := func(old, new interface{}) { dns.updateModifed() }
```
johnbelamaric, line 114: I don't think this is a good idea. Without the endpoint subsets comparison operations, there will be a continuous stream of updates requiring new AXFRs even though there is no actual, relevant change.

johnbelamaric (changes requested):
I get the idea, but I am not sure it's viable. The endpoint comparison is necessary, because endpoints get lots of updates that are completely irrelevant and don't change the DNS records at all.

miekg:
>johnbelamaric requested changes on this pull request.
>
>I get the idea, but I am not sure it's viable. The endpoint comparison is necessary, because endpoints get lots of updates that are completely irrelevant and don't change the DNS records at all.
>
>> @@ -106,6 +107,7 @@ kubernetes [ZONES...] {
> * `ignore empty_service` return NXDOMAIN for services without any ready endpoint addresses (e.g. ready pods).
>   This allows the querying pod to continue searching for the service in the search path.
>   The search path could, for example, include another kubernetes cluster.
>+* `watch` enables updates being send to a server listening for watching on a gRPC channel.
>
>How about:
>
>`watch` enables sending updates to clients subscribed to records using the gRPC Watch streaming API call


ack.

>> @@ -109,6 +109,14 @@ func newdnsController(kubeClient kubernetes.Interface, opts dnsControlOpts) *dns
> 		zones:            opts.zones,
> 		endpointNameMode: opts.endpointNameMode,
> 	}
>+	add := func(obj interface{}) { dns.updateModifed() }
>+	del := func(obj interface{}) { dns.updateModifed() }
>+	upd := func(old, new interface{}) { dns.updateModifed() }
>
>I don't think this is a good idea. Without the endpoint subsets comparison operations, there will be a continuous stream of updates requiring new AXFRs even though there is no actual, relevant change.

Ok, I couldn't easily find anymore what we had here originally.

chrisohaver:
> Disable watch by default (takes up mem) ...

Do we know how much memory it takes up?

johnbelamaric:
We had what you are proposing originally, but while doing the watch, I
found that I was getting continuous updates, and the serial was changing
all the time. So that's why we implemented the other bits. It's still
relevant with or without the watch.

miekg:
dropping

PR #2160, opened 2018-10-04
plugin/route53: add split zone support

canthefason:
This PR introduces split-zone support for the Route53 plugin. 

When there are multiple hosted zones with overlapping domains (private vs. public hosted zone), previously this plugin was returning a zone conflict error. Now CoreDNS looks up for the queried domain starting from the leftmost hosted zone and moves forward to the next one until it finds a matching resource record. When there's no matching resource record, CoreDNS responds with the rightmost hosted zones SOA record.

Signed-off-by: Can Yucel <jon.yucel@getcruise.com>

### DIFF
--- a/plugin/route53/route53.go
+++ b/plugin/route53/route53.go
@@ -30,28 +30,44 @@ type Route53 struct {
 	upstream  *upstream.Upstream
 
 	zMu   sync.RWMutex
-	zones map[string]*zone
+	zones zones
 }
 
 type zone struct {
-	id string
-	z  *file.Zone
+	id  string
+	z   *file.Zone
+	dns string
+}
+
+type zones map[string][]*zone
+
+// AddHostedZone adds a new hosted zone into the zones map.
+// Returns true if the zone introduces a new domain.
+func (z zones) AddHostedZone(hz *zone) (newDomain bool) {
+	hz.z = file.NewZone(hz.dns, "")
+	if _, ok := z[hz.dns]; !ok {
+		z[hz.dns] = make([]*zone, 0)
+		newDomain = true
+	}
+	z[hz.dns] = append(z[hz.dns], hz)
+	return
 }
 
 // New returns new *Route53.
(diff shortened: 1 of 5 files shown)


yongtang:
SGTM

plugin/route53/route53.go
```diff
@@ -30,28 +30,44 @@ type Route53 struct {
...
+}
+
+type zones map[string][]*zone
+
+// AddHostedZone adds a new hosted zone into the zones map.
+// Returns true if the zone introduces a new domain.
+func (z zones) AddHostedZone(hz *zone) (newDomain bool) {
+	hz.z = file.NewZone(hz.dns, "")
+	if _, ok := z[hz.dns]; !ok {
+		z[hz.dns] = make([]*zone, 0)
```
dilyevsky: nit: this line is not needed - appending to nil slice works just as well.

plugin/route53/route53.go
```diff
@@ -30,28 +30,44 @@ type Route53 struct {
...
-			HostedZoneId: aws.String(id),
+			DNSName:      aws.String(hostedZone.dns),
+			HostedZoneId: aws.String(hostedZone.id),
 		})
 		if err != nil {
 			return nil, err
 		}
-		zones[dns] = &zone{id: id, z: file.NewZone(dns, "")}
-		zoneNames = append(zoneNames, dns)
+		if zones.AddHostedZone(hostedZone) {
```
dilyevsky: Kind of prefer the old `keys` rather than passing half-initialized structures around. Probably need to invert it though to allow for multiple ids for the same zone name. Also could you document input params in the docstring please.

canthefason: just inverting was my initial thought, however it was ending up with losing the hosted zone ordering.

I changed the structure to a map[string][]string instead where the key is domain name and the value is a list of hosted zone ids. Let me know how you feel about it.

dilyevsky (approved):
/merge

PR #2162, opened 2018-10-05
Add request and response context to traces

negz:
This PR adds some context, in the form of root span tags, to traces emitted by CoreDNS.

The existing tracing implementation is great for determining how much time CoreDNS spends in each plugin, but doesn't provide any way to relate traces back to the queries that invoked them. This PR appends the queried name (e.g. `example.org.`) to the span name, and annotates each root span with the queried name, query type (`A`, `CNAME`, etc) and response code (`NXDOMAIN`, etc).

At Planet Labs we've been trying to diagnose some strange DNS client behaviour, and it would be useful to see a sample of the incoming requests and their responses. The `log` plugin could help, but we don't want incur the performance hit of logging _every_ request, just a sample. We already trace one CoreDNS request in every thousand, and we've found having request and response context associated with spans useful when debugging other services (linkerd, Traefik, etc), so tracing seemed like an ideal solution.

Here's a screenshot of some traces emitted by this code in the Stackdriver Trace UI (via the Zipkin to Stackdriver Trace proxy):

dnstrace

I neglected to raise an issue for this relatively small change.

I don't feel any documentation changes are required, but am happy to update the docs if someone has a suggestion.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -348,6 +348,7 @@
     ".",
     "ext",
     "log",
+    "mocktracer",
   ]
   pruneopts = ""
   revision = "1949ddbfd147afd4d964a9f00b24eb291e0e7c38"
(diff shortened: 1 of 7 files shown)


johnbelamaric:
Very nice, thanks!

One step further you could take tracing, if it's useful to you, would be to use an EDNS0 option to contain the span ID, so you could connect each DNS request with the originating root span. Essentially what you would get if you used gRPC lookups, but using more-or-less ordinary DNS. Something to consider.

/lgtm
/merge

PR #2163, opened 2018-10-08
add username and password of etcd supporter

BolaNasr:
support etcd scurity ( user and password)  to help you to connect to it 

ETCD_USERNAME 
ETCD_PASSWORD

**New Feature**

### DIFF
--- a/plugin/etcd/setup.go
+++ b/plugin/etcd/setup.go
@@ -3,6 +3,7 @@ package etcd
 import (
 	"context"
 	"crypto/tls"
+	"os"
 
 	"github.com/coredns/coredns/core/dnsserver"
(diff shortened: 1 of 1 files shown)


miekg:
I'm not totally against this, although it's a bit hidden as a feature; I would prefer to access this from the Corefile instead: you can use env variables there is as well.

So this should be morphed into an extension of the allowed directives; i.e. add `username` and `password`. Also needs an updated README.md with the new directives.

yongtang:
+1 for get username/password from Corefile. I feel env variables could easily be hidden and mixed with env variables used by other programs.

miekg:
closing, as this needs probably a new PR  - as discussed above.

Issue #2161, opened 2018-10-04
Segfault when forwarder list contains local server itself


hhoffstaette:
I recently migrated from dnsmasq and very much like coredns overall (now running everywhere),
but of course also immediately found a bug. :)

With the following config:
```
. {
	forward . /etc/resolv.conf
	cache
	log
}
```
and the following resolv.conf:
```
# host-local instance for resilience/caching
nameserver 127.0.0.1
# LAN-local instance that forwards to my ISP etc.
nameserver 192.168.100.222
```
coredns repeatably crashes with the follwing log a few seconds after start & a bit of initial random
lookup activity:
```
coredns -conf crash.conf
.:53
2018/10/05 01:25:11 [INFO] CoreDNS-1.2.2
2018/10/05 01:25:11 [INFO] linux/amd64, go1.11, eb51e8bac90fac86d34c9e1cb89b04ea0936b034
CoreDNS-1.2.2
linux/amd64, go1.11, eb51e8bac90fac86d34c9e1cb89b04ea0936b034
127.0.0.1:56626 - [05/Oct/2018:01:25:15 +0200] 35427 "PTR IN 166.241.155.62.in-addr.arpa. udp 45 false 512" NXDOMAIN qr,rd,ra 139 0.017405332s
127.0.0.1:50454 - [05/Oct/2018:01:25:15 +0200] 35427 "PTR IN 166.241.155.62.in-addr.arpa. udp 45 false 512" NXDOMAIN qr,rd,ra 139 0.01795297s
127.0.0.1:56209 - [05/Oct/2018:01:25:15 +0200] 22338 "PTR IN 38.118.5.217.in-addr.arpa. udp 43 false 512" NXDOMAIN qr,rd,ra 140 0.000773091s
127.0.0.1:51105 - [05/Oct/2018:01:25:24 +0200] 57904 "AAAA IN ntp3.ptb.de. udp 29 false 512" NOERROR qr,rd,ra 112 0.000827708s
127.0.0.1:58737 - [05/Oct/2018:01:25:24 +0200] 61978 "A IN ntp3.ptb.de. udp 29 false 512" NOERROR qr,rd,ra 100 0.000752208s
127.0.0.1:56626 - [05/Oct/2018:01:25:24 +0200] 61978 "A IN ntp3.ptb.de. udp 29 false 512" NOERROR qr,rd,ra 100 0.001321987s
[...]

PR #2165, opened 2018-10-08
plugin/forward: fix healthchecker crash

rdrozhdzh:
this can fix health-check crash, also adding more logging
#2161
none


rdrozhdzh:
the issue may be a regression after #1950

PR #2167, opened 2018-10-08
plugin/kubernetes: Reduce memory usage: Use keys instead of indexes

chrisohaver:

Use the default namespace/name key built into the client-go store instead of maintaining indexes.
Reduces resource usage (memory and cpu).

#2095

None

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -20,21 +20,19 @@ import (
 )
 
 const (
-	podIPIndex            = "PodIP"
-	svcNameNamespaceIndex = "NameNamespace"
-	svcIPIndex            = "ServiceIP"
-	epNameNamespaceIndex  = "EndpointNameNamespace"
-	epIPIndex             = "EndpointsIP"
+	podIPIndex = "PodIP"
+	svcIPIndex = "ServiceIP"
+	epIPIndex  = "EndpointsIP"
 )
 
 type dnsController interface {
 	ServiceList() []*object.Service
 	EndpointsList() []*object.Endpoints
-	SvcIndex(string) []*object.Service
-	SvcIndexReverse(string) []*object.Service
+	SvcIndex(string) *object.Service
+	SvcIndexReverse(string) *object.Service
 	PodIndex(string) []*object.Pod
-	EpIndex(string) []*object.Endpoints
-	EpIndexReverse(string) []*object.Endpoints
+	EpIndex(string) *object.Endpoints
+	EpIndexReverse(string) *object.Endpoints
 
 	GetNodeByName(string) (*api.Node, error)
 	GetNamespaceByName(string) (*api.Namespace, error)
@@ -118,7 +116,7 @@ func newdnsController(kubeClient kubernetes.Interface, opts dnsControlOpts) *dns
 		&object.Service{},
 		opts.resyncPeriod,
 		cache.ResourceEventHandlerFuncs{AddFunc: dns.Add, UpdateFunc: dns.Update, DeleteFunc: dns.Delete},
-		cache.Indexers{svcNameNamespaceIndex: svcNameNamespaceIndexFunc, svcIPIndex: svcIPIndexFunc},
+		cache.Indexers{svcIPIndex: svcIPIndexFunc},
 		object.ToService,
 	)
 
@@ -145,7 +143,7 @@ func newdnsController(kubeClient kubernetes.Interface, opts dnsControlOpts) *dns
 			&api.Endpoints{},
 			opts.resyncPeriod,
 			cache.ResourceEventHandlerFuncs{AddFunc: dns.Add, UpdateFunc: dns.Update, DeleteFunc: dns.Delete},
-			cache.Indexers{epNameNamespaceIndex: epNameNamespaceIndexFunc, epIPIndex: epIPIndexFunc},
+			cache.Indexers{epIPIndex: epIPIndexFunc},
 			object.ToEndpoints)
 	}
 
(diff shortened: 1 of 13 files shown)


chrisohaver:
Most of the changes here are fallout from changing the signature of functions in the `dnsController` interface.  Several functions that used to return a slice of objects, now return a single object.

The guts of this change are simple, just that we now use  `GetByKey()` instead of `ByIndex()` in a couple places (`dns.svcLister.GetByKey()` and `dns.epLister.GetByKey()`).

Everything else is byproduct of changing the interface.

I could have left the interface function return values unchanged, but that would require packing single answers into slices just to keep the diff short.  Those slices only ever contained one object because we were using unique indexes.

miekg:
This make sense. Thanks for the extra explanation; that helped.

chrisohaver:
The merge was less horrifying than I thought it would be.  Only took a few hours.

chrisohaver:
OK - all tests pass... this one is ready for review...

plugin/kubernetes/controller.go
```diff
@@ -118,7 +116,7 @@ func newdnsController(kubeClient kubernetes.Interface, opts dnsControlOpts) *dns
 		&object.Service{},
 		opts.resyncPeriod,
 		cache.ResourceEventHandlerFuncs{AddFunc: dns.Add, UpdateFunc: dns.Update, DeleteFunc: dns.Delete},
-		cache.Indexers{svcNameNamespaceIndex: svcNameNamespaceIndexFunc, svcIPIndex: svcIPIndexFunc},
```
chrisohaver, line 121: remove the 'name.namspace' index - we will use keys instead.

plugin/kubernetes/controller.go
```diff
@@ -145,7 +143,7 @@ func newdnsController(kubeClient kubernetes.Interface, opts dnsControlOpts) *dns
 			&api.Endpoints{},
 			opts.resyncPeriod,
 			cache.ResourceEventHandlerFuncs{AddFunc: dns.Add, UpdateFunc: dns.Update, DeleteFunc: dns.Delete},
-			cache.Indexers{epNameNamespaceIndex: epNameNamespaceIndexFunc, epIPIndex: epIPIndexFunc},
```
chrisohaver, line 148: remove the 'name.namspace' index - we will use keys instead.

plugin/kubernetes/controller.go
```diff
@@ -368,27 +353,24 @@ func (dns *dnsControl) PodIndex(ip string) (pods []*object.Pod) {
 		if !ok {
 			continue
 		}
-		pods = append(pods, p)
+		return []*object.Pod{p}
```
chrisohaver, line 356: This index is unique, it will only return one pod, so just return the first pod object we find

plugin/kubernetes/controller.go
```diff
@@ -368,27 +353,24 @@ func (dns *dnsControl) PodIndex(ip string) (pods []*object.Pod) {
...
 	}
-	return pods
+	return nil
 }
 
-func (dns *dnsControl) SvcIndex(idx string) (svcs []*object.Service) {
-	os, err := dns.svcLister.ByIndex(svcNameNamespaceIndex, idx)
+func (dns *dnsControl) SvcIndex(key string) *object.Service {
+	o, _, err := dns.svcLister.GetByKey(key)
```
chrisohaver, line 362: Get by key instead of by index.

plugin/kubernetes/controller.go
```diff
@@ -399,27 +381,27 @@ func (dns *dnsControl) SvcIndexReverse(ip string) (svcs []*object.Service) {
...
+		return s
 	}
-	return svcs
+	return nil
 }
 
-func (dns *dnsControl) EpIndex(idx string) (ep []*object.Endpoints) {
-	os, err := dns.epLister.ByIndex(epNameNamespaceIndex, idx)
+func (dns *dnsControl) EpIndex(key string) (ep *object.Endpoints) {
+	o, _, err := dns.epLister.GetByKey(key)
```
chrisohaver, line 390: Get by key instead of by index.

plugin/kubernetes/object/service.go
```diff
@@ -20,7 +20,7 @@ type Service struct {
 }
 
 // ServiceKey return a string using for the index.
-func ServiceKey(name, namespace string) string { return name + "." + namespace }
+func ServiceKey(namespace, name string) string { return namespace + "/" + name }
```
chrisohaver, line 23: the built in keys are in the form `namespace/name`

plugin/kubernetes/kubernetes.go
```diff
@@ -444,9 +444,12 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 		serviceList = k.APIConn.ServiceList()
 		endpointsListFunc = func() []*object.Endpoints { return k.APIConn.EndpointsList() }
 	} else {
-		idx := object.ServiceKey(r.service, r.namespace)
-		serviceList = k.APIConn.SvcIndex(idx)
-		endpointsListFunc = func() []*object.Endpoints { return k.APIConn.EpIndex(idx) }
+		key := object.ServiceKey(r.namespace, r.service)
```
fturib, line 447: the reverse of the params here on ServiceKey is on purpose ?

plugin/kubernetes/kubernetes_test.go
```diff
@@ -59,43 +59,24 @@ func TestEndpointHostname(t *testing.T) {
...
-			ExternalName: "coredns.io",
-			Type:         api.ServiceTypeExternalName,
-			Ports: []api.ServicePort{
-				{Name: "http", Protocol: "tcp", Port: 80},
-			},
-		},
+func (a APIConnServiceTest) SvcIndex(key string) *object.Service {
+	for _, s := range a.ServiceList() {
+		if object.ServiceKey(s.Namespace, s.Name) == key {
+			return s
```
fturib, line 76: although here we have the same order or params : NS, Name

plugin/kubernetes/object/service.go
fturib: ok. I see reverse of params here..

plugin/kubernetes/kubernetes.go
fturib: never mind. I see the change in the def of the method.

plugin/kubernetes/kubernetes_test.go
fturib: ok.

fturib (approved):
LGTM. I am ok with the changes.
Better to plan re-run the memory usage and some sort of perf - because the change of usage in the K8s API.

johnbelamaric (approved):
/lgtm

PR #2171, opened 2018-10-09
Add cgo option to Makefile

chrisohaver:

Add cgo option to allow compile with race.

coredns/ci#70


yongtang (approved):
LGTM

PR #2172, opened 2018-10-09
Rename VERBOSE variable to BUILDOPTS

chrisohaver:

VERBOSE actually acts as a generic go build flags argument that happens to default to just "-v", but you can include other options.  Rename the variable to BUILDOPTS to better reflect the function of the variable.


yongtang:
SGTM. Can coredns/ci be retested (as it errors out)?

chrisohaver:
> SGTM. Can coredns/ci be retested (as it errors out)?

It errors out using coredns/ci#71 ...

PR #2177, opened 2018-10-10
plugin/federation: Add upstream option to federation

chrisohaver:
Adds upstream option to federation, so that answers for federated services (which are rendered as CNAME records) can include the records they point to.

### DIFF
--- a/plugin/federation/federation.go
+++ b/plugin/federation/federation.go
@@ -110,8 +110,6 @@ func (f *Federation) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.
 
 	if f.Upstream != nil {
 		aRecord, err := f.Upstream.Lookup(state, service.Host, state.QType())
-		if err != nil {
-		}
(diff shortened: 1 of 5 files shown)


chrisohaver:
I added a CI test for this - coredns/ci#72.  I don't think a unit test for upstream function is easy.

chrisohaver:
FYI @shashidharatd

plugin/federation/federation.go
```diff
@@ -105,6 +108,15 @@ func (f *Federation) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.
 
 	m.Answer = []dns.RR{service.NewCNAME(state.QName(), service.Host)}
 
+	if f.Upstream != nil {
+		aRecord, err := f.Upstream.Lookup(state, service.Host, state.QType())
+		if err != nil {
+		}
```
fturib: What is the usage of this test with empty action ?

fturib (commented):
LGTM, besides one comment that seems to be related to dead code.

fturib (approved):

PR #2184, opened 2018-10-11
Forward: introduce `Opt` builder for creating plugin instance.

### LINKED ISSUE #2182: Expose support for forward's properties settings
Hi, one of my plugin may require reuse the forward plugin's functionality. Firstly I try to open a nested block for forward:

```
. {
  myplugin {
    myplugin_settings xxx
    forward . 1.1.1.1 {
      duration 10
    }
  }
}
```

But with no luck, caddyfile's dispenser seems have no support for this nested usage (and I do agree it's a bad style for config).

Another approach looks like:

```
. {
  myplugin {
    myplugin_settings_foo foo
    forward_expire DURATION
    // ... other properties for forward, prefixed with `forward`
  }
}
```

```go
type MyPlugin struct {
  Forward forward.Forward
}

func setup(c *caddy.Controller) *MyPlugin {
  p := New()

  // do parse work here
  parseStanza(c, p)

  return c
}

// so in parseStanza we can setup the forward plugin with exposed setters:
func parseStanza(c *caddy.Controller, p *MyPlugin) {
  // reading from caddyfile.Dispenser
  switch dispenser.Val() {
    case "forward_duration":
      p.Forward.SetDuration(parsedDuration)
    case "other_property":
      // ...
  }
}
```

Some may ask why not just fall through and let the chain handle this case, but to my usage, there is a need for altering the chain flowing direction, like a `if` statement, generally speaking, the flow would look like that:

```
. {
  if {
    condition_to_check
    perform_when_condition_meet
  }

  fallthrough_plugin
}
```

Any suggestions for this change? If you agree with this modification, I can submit a pr.

bcho:
fixes: #2182

This pull request adds a simple implementation for #2182, which decouples the forward plugin building logic, and enables reusing later.

#2182 

forward plugin's `New` method may be marked as deprecated

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -13,6 +13,8 @@ import (
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/debug"
 	clog "github.com/coredns/coredns/plugin/pkg/log"
+	"github.com/coredns/coredns/plugin/pkg/parse"
+	"github.com/coredns/coredns/plugin/pkg/transport"
 	"github.com/coredns/coredns/request"
 
 	"github.com/miekg/dns"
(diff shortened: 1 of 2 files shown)


plugin/forward/forward.go
```diff
@@ -21,6 +23,72 @@ import (
...
+		TLSConfig:           new(tls.Config),
+		Maxfails:            2,
+		Expire:              defaultExpire,
+		HealthCheckInterval: hcInterval,
+		SelectPolicy:        new(random),
+	}
+}
+
+// Build creates a forward instance.
+func (o Opt) Build() *Forward {
```
johnbelamaric: This should be (o *Opt) or it will make a copy.

plugin/forward/setup.go
```diff
@@ -89,51 +88,34 @@ func parseForward(c *caddy.Controller) (*Forward, error) {
...
-			f.proxies[i].SetTLSConfig(f.tlsConfig)
-		}
-		f.proxies[i].SetExpire(f.expire)
-	}
-	return f, nil
+	return o.Build(), nil
 }
 
-func parseBlock(c *caddyfile.Dispenser, f *Forward) error {
+func parseBlock(c *caddyfile.Dispenser, opt *Opt) error {
```
johnbelamaric: Why name it "opt" here and "o" everywhere else?

johnbelamaric (changes requested):
A few minor tweaks but otherwise lgtm. I know thought that @miekg has had concerns in the past about exposing this sort of thing in a plugin, rather than moving to a common pkg.

bcho:
@johnbelamaric changes updated, PTAL.

> concerns in the past about exposing this sort of thing in a plugin, rather than moving to a common pkg

Yep, I have the same concern too, but I don't find any implementations applying this way. @miekg any suggestions here?

miekg:
How is this different than just exposing the bits you need?

The problem with the approach is this now puts a requirement on this pluign to not break this api - there was never a thing like this before, so it's a pretty big change.

What's the usecase here and why can this not be solved by the various Lookup functions that are sort of the entry point for other plugin into this one?
(even with something like this in place it will break at some point, because there isn't any inherit guarantee being agreed here)

bcho:
Hi @miekg 

> What's the usecase here

The main use case is to allow other plugins to embed forward plugin's capacity (basically, to use `Forward.ServeDNS`).

For now, `Forward` holds some internal fields and I think it can be instantiated by an `Opt` builder. This builder actually reduces some dependencies between the caller and the callee IMO. Of course, this kind of usage requires the API to be stable, as the use case is just want to create dependencies between plugins.

But I come up with a better way, like just copy the `ServeDNS` logic into my plugin like that:

```go
import (
  "github.com/coredns/coredns/plugin/forward"
)

func (p MyPlugin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) {
  if fallbackToForward(ctx, w, r) {
    p.serveDNSWithForward(ctx, w, r)
  }
}

func (p MyPlugin) serveDNSWithForward(// ...) {
  var proxies []*forward.Proxy
  // using proxy as the same way forward does
}
```

So I think it's ok to close this pull request by now.

PR #2185, opened 2018-10-11
plugin/loop: tweak loop detected msg, add troubleshooting section

chrisohaver:

Feedback from the field, the current message is confusing some people. This PR tweaks the message a bit, moving the long "confusing" HINFO probe detail to the end of the message.

In the context of kubernetes, the exit on detection behavior manifests as a "crashloop" on the pod, since k8s detects the container exit as a "crash", and will try to start it up again ad infinitum.  I think people also find this confusing. But, there isn't anything practical we could do to change that.  If for example, we didn't exit, and disabled all forwards/proxys to the offending zone (hard to do), it would leave CoreDNS in a "silently" semi-functional state, which is IMO worse.  The presence of the coredns pod "crashloop" is a breadcrumb of sorts, to go check out the coredns log.  So, I think the best we can do is to improve the message a little.

it has come up in slack many times, and in various github repos... 
e.g. kubernetes/kubeadm#1162

Included

### DIFF
--- a/plugin/loop/loop.go
+++ b/plugin/loop/loop.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 
 	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/pkg/dnsutil"
 	clog "github.com/coredns/coredns/plugin/pkg/log"
 	"github.com/coredns/coredns/request"
 
@@ -49,7 +50,8 @@ func (l *Loop) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 	}
 
 	if l.seen() > 2 {
-		log.Fatalf("Forwarding loop detected in \".\" zone. Probe \"HINFO IN %s\" was seen more than twice.", l.qname)
+		zone := dnsutil.Join(dns.SplitDomainName(l.qname)[2:]...) + "."
+		log.Fatalf("Forwarding loop detected in \"%s\" zone. Probe \"HINFO IN %s\" was seen more than twice.", zone, l.qname)
 	}
 
 	return plugin.NextOrFailure(l.Name(), l.Next, ctx, w, r)
--- a/plugin/loop/loop.go
+++ b/plugin/loop/loop.go
@@ -5,7 +5,6 @@ import (
 	"sync"
 
 	"github.com/coredns/coredns/plugin"
-	"github.com/coredns/coredns/plugin/pkg/dnsutil"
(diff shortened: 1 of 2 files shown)


chrisohaver:
actually - i need to pull the zone part from the probe message - not hard code to "."

plugin/loop/loop.go
```diff
@@ -49,7 +50,11 @@ func (l *Loop) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 	}
 
 	if l.seen() > 2 {
-		log.Fatalf("Seen \"HINFO IN %s\" more than twice, loop detected", l.qname)
+		zone := "."
+		if len(dns.SplitDomainName(l.qname)) > 2 {
+			zone = dnsutil.Join(dns.SplitDomainName(l.qname)[2:]...)
```
miekg: why is this not `l.zone`?

We should probably also point to https://coredns.io/plugins/loop#looping (and adjust loop/README.md to have a section on it under `#looping`)

chrisohaver: ha `l.zone` of course - the answer is already there.

> We should probably also point to

Do you mean put url in the error message? e.g. 

```
Forwarding loop detected in "." zone. See https://coredns.io/plugins/loop#looping. ...
```

rajansandeep (approved):

plugin/loop/loop.go
```diff
@@ -50,11 +49,7 @@ func (l *Loop) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 	}
 
 	if l.seen() > 2 {
-		zone := "."
-		if len(dns.SplitDomainName(l.qname)) > 2 {
-			zone = dnsutil.Join(dns.SplitDomainName(l.qname)[2:]...)
-		}
-		log.Fatalf("Forwarding loop detected in \"%s\" zone. Probe \"HINFO IN %s\" was seen more than twice.", zone, l.qname)
+		log.Fatalf("Forwarding loop detected in \"%s\" zone. Exiting. See https://coredns.io/plugins/loop#troubleshooting.", l.zone, l.qname)
```
johnbelamaric: You left l.qname in there, it should come out. Otherwise, looks good.

johnbelamaric: ooops, nevermind. was looking at a partial change

miekg:
Yes put the URL there. Means we don't need a new section on coredns.io and
the info is also in the readme and manual page

PR #2186, opened 2018-10-11
Revert "plugin/kubernetes: Reduce memory usage: Use keys instead of indexes"

chrisohaver:
Reverts coredns/coredns#2167

Inexplicably, these changes increased memory usage when merged with the custom store solution.

### DIFF
--- a/plugin/federation/kubernetes_api_test.go
+++ b/plugin/federation/kubernetes_api_test.go
@@ -16,13 +16,12 @@ type APIConnFederationTest struct {
 func (APIConnFederationTest) HasSynced() bool                           { return true }
 func (APIConnFederationTest) Run()                                      { return }
 func (APIConnFederationTest) Stop() error                               { return nil }
-func (APIConnFederationTest) SvcIndexReverse(string) *object.Service  { return nil }
-func (APIConnFederationTest) EpIndexReverse(string) *object.Endpoints { return nil }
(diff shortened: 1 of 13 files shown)


fturib (approved):
LGTM.
It is kind of sad to revert here, there are good points (using keys instead new index, optim on the transfer, ...)

miekg:
you're branch is not clean; you reverting something else as well

fturib:
the other revert is empty ...
and the revert content is automatically computed from the initial PR.

chrisohaver:
> you're branch is not clean; you reverting something else as well

I reverted it directly from the github website.  Clicking on "revert".  So it should be clean.
Perhaps the original PR was not clean.

chrisohaver:
> It is kind of sad to revert here

Yes - it makes no sense at all.  Removing 2 redundant indexes increases memory use?  But, we are under a time crunch... so it will probably remain a mystery.

miekg:
re: revert: i don't know, but it looks weird. Please fix.

without a proper pprof setup we will also be in the guessing game so that should somehow get supported or we need to do our own scale tests

miekg:
oh, you rebased! I think this might be hard then let me see locally.

miekg:
no, git revert 974ed086  just works.

chrisohaver:
> oh, you rebased! 

Should I not rebase going forward?

Issue #2035, opened 2018-08-16
No any DELETE-operation in Cache ?


ckcd:
Hi,

I just read the code of "Cache" plugin, I was wondering about the cache delete:

First, There are `Cache.Remove` and `Shard.Remove` which are used to delete items, However, no methods call these two func.

Second, In "handler.go / get" of cache plugin, if the `i.(*item).ttl(now) <= 0` then the item will NOT be used, but why not delete this item ?
```
func (c *Cache) get(now time.Time, state request.Request, server string) (*item, bool) {
	k := hash(state.Name(), state.QType(), state.Do())

	if i, ok := c.ncache.Get(k); ok && i.(*item).ttl(now) > 0 {
		cacheHits.WithLabelValues(server, Denial).Inc()
		return i.(*item), true
	}

	if i, ok := c.pcache.Get(k); ok && i.(*item).ttl(now) > 0 {
		cacheHits.WithLabelValues(server, Success).Inc()
		return i.(*item), true
	}
	cacheMisses.WithLabelValues(server).Inc()
	return nil, false
}
```

Third, when a shard is full, `Add` will call `evict` to randomly delete a item rather traverse the shard to find items whose ttl <=0 , so maybe there is a bad situation: a item is added and never be used again, but the randomly methods nerver chose these lucky item ,so the useless item will be always existed ?

fturib:
I think the cache has been reworked to be completely oriented for a better performance under high QPS. Which means high level of threading.
Which means for the cache, try to limit the usage of locks, and if possible keep usage of RLock. At this time was introduces the split into shards.

What is a little more concerning is the way we choose the one to remove from the shard : randomely instead of picking an obsolete entry.

Could fix that eviction choice without impacting the performance ?

miekg:
Random eviction is a performance choice as well. In my initial testing a
LRU required so many locks it didn't perform at all.

Prefetch should offset the random eviction.

We should delete the unused code btw.

ckcd:
@miekg Thks for your answer,

so in your test the "random eviction" performance is better, does it relay on number of items (e.g. 256*512 in defalut)?

And should we delete the obsoleted items in '"func (c *Cache) get" while `i.(*item).ttl(now) <= 0` ?

Thks~

miekg:
The sharded cache is *also* done for performance so concurrent go-routines don't all want to use the same lock.

If we have unused code a PR that deletes it is welcome.

PR #2190, opened 2018-10-12
Delete out-of-date item in cache when get it.

ckcd:
In cache.get(), first it will check the cache to find the item, when get it, secondly, use ttl() to check whether the item is out-of-date, if it's OK then this item is the final answer, but if it's NOT, which means the ttl of this item is < 0, for now, the code doing nothing, I thinks it should be: delete this useless item.
As we know, cache plugin will random replace a item when the cache is full, instead of traverse the cache and find a expired item, this is for performance and avoid using to many locks. But in my situation, we already find a expired item and don't need any traverse, so we should delete the item with a little attention.

#2035 

None.

### DIFF
--- a/plugin/cache/cache_test.go
+++ b/plugin/cache/cache_test.go
@@ -292,3 +292,26 @@ func TestComputeTTL(t *testing.T) {
 		}
 	}
 }
+
+func TestOutOfDateItem(t *testing.T) {
+	c := New()
+	c.pttl = 5 * time.Second
+	c.nttl = 5 * time.Second
+	c.Next = BackendHandler()
+
+	req := new(dns.Msg)
+	req.SetQuestion("example.org.", dns.TypeA)
+	ctx := context.TODO()
+
+	c.ServeDNS(ctx, &test.ResponseWriter{}, req)
+	time.Sleep(5 * time.Second)
+
+	c.Next = nil
+	c.ServeDNS(ctx, &test.ResponseWriter{}, req)
+	if c.pcache.Len() != 0 {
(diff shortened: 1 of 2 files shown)


yongtang:
lgtm

miekg:
This turns every read into a write, is that desirable?

ckcd:
@miekg 
1. This only occurs in: query a item which is expired, so it won't affect too many read.
2. With this delete, we can reduce the scene of randomly eviction (i.e. the cache is full), so to protect useful items.

So yes, it is desirable in my opinion.
Thanks~

grobie:
>     2\. With this delete, we can reduce the scene of randomly eviction (i.e. the cache is full), so to protect useful items.

I'm curious about this one, can you confirm that with metrics in production? I'd have expected that the Cache.get() code path is only triggered during a DNS request against this name. A cache miss would then result in a Cache.write shortly after with an up-to-date record. Therefore, this change seems to change the cache commands from "cache get (miss), next plugin, cache write" to "cache get (miss), cache delete, next plugin, cache write". What other paths would trigger a cache get+delete without an immediate write to the same location?

ckcd:
@grobie 
Thanks for your opinion! 
I didn't notice the "cache write" before, so I think we should do delete after get a up-to-date item, I'm wrong~

miekg:
would a separate go-routine that scans the cache and expunges items be better here?

ckcd:
@miekg 
Sorry for reply so late,
a new goroutine may delete unuse item, but also use too many lock,
whatever I will try this method and do some test~

PR #2191, opened 2018-10-12
pprof plugin -- survive coredns reload

hmage:
`pprof` plugin does not survive CoreDNS reload when asked to via SIGUSR1 or when `reload` plugin does it automatically.

Couldn't find any.

None, it was expected behaviour to survive CoreDNS reload.

### DIFF
--- a/plugin/pprof/setup.go
+++ b/plugin/pprof/setup.go
@@ -47,10 +47,8 @@ func setup(c *caddy.Controller) error {
 		}
 	}
 
-	pprofOnce.Do(func() {
-		c.OnStartup(h.Startup)
-		c.OnShutdown(h.Shutdown)
-	})
+	c.OnStartup(h.Startup)
+	c.OnRestart(h.Shutdown)
 
 	return nil
 }


hmage:
About Travis TestReadme failing because test port is already taken -- a rerun should fix the fail.

chrisohaver:
restarted failed job

plugin/pprof/setup.go
```diff
@@ -47,10 +47,8 @@ func setup(c *caddy.Controller) error {
 		}
 	}
 
-	pprofOnce.Do(func() {
-		c.OnStartup(h.Startup)
-		c.OnShutdown(h.Shutdown)
-	})
+	c.OnStartup(h.Startup)
+	c.OnRestart(h.Shutdown)
```
fturib, line 51: Hum.. that will work for the reload.
However, if the pprof setup is called several time for each reload, only the first setup will succeed, and following will fail the chain of startup event.

It may be a limitation we do not care ... just need to say it in the README.
or to not fail the the Startup() when cannot listen on the port. (only logging the error)

exemples:

```
mydomain.com, other.com {
     pprof ...
}

```
or

```
. {
   pprof ...
}
mydomain.com {
   pprof ...
}

```

fturib (commented):
Hi think would need to add info on limitation in the README of pprof.
(see comment of review)

plugin/pprof/setup.go
miekg: yes, good question; does this still work for this scenario?

miekg:
this seems to have stalled. Closing

Issue #2168, opened 2018-10-09
Plugin/Errors - add a "consolidate" option to limit the logging of errors


fturib:
## Presentation

@rdrozhdzh extended plugin ERRORS for Infoblox purposes.
This extension would be useful for everyone.

The changes added (from standard ERRORS plugin of CoreDNS) are:
* added a consolidate option to prevent intensive logging of errors of particular type
* updated README.md
* added unit tests, including concurrent one
* used github.com/coredns/coredns/plugin/pkg/log to make log consistent with other plugins

Please, have a look at this repo : https://github.com/infobloxopen/coredns-plugin-errors

The plan would be to just integrate this new ERRORS plugin as the standard one.

## Adding "consolidate" option in ERRORS

Extra knobs are available with an expanded syntax:

```
errors {
	consolidate DURATION REGEXP
}
```
Option consolidate allows collecting several error messages matching the regular expression REGEXP during DURATION. After the DURATION since receiving the first such message, the consolidated message will be printed to standard output, e.g.

> 2 errors like '^read udp .* i/o timeout$' occurred in last 30s

Exemple of syntax:
```
. {
    forward . 8.8.8.8
    errors {
        consolidate 5m ".* i/o timeout$"
        consolidate 30s "^Failed to .+"
    }
}

```

rdrozhdzh:
thank you @fturib for presenting this feature.

The changes comparing to original 'errors' plugin could be easily reviewed here https://github.com/infobloxopen/coredns-plugin-errors/pull/3/commits/eb323591a433420049495348c562ceb6f6f1d6bc

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] Plugin/Errors - a..." ]
>## Presentation
>
>@rdrozhdzh extended plugin ERRORS for Infoblox purposes.
>This extension would be useful for everyone.
>
>The changes added (from standard ERRORS plugin of CoreDNS) are:
>* added a consolidate option to prevent intensive logging of errors of particular type
>* updated README.md
>* added unit tests, including concurrent one
>* used github.com/coredns/coredns/plugin/pkg/log to make log consistent with other plugins
>
>Please, have a look at this repo : https://github.com/infobloxopen/coredns-plugin-errors

Nice! (haven't looked at the impl. yet)

Doing this also for logging makes sense.

fturib:
@rdrozhdzh : can you add the PR to integrate the coredns-plugin-errors in CoreDNS  as replacement of errors ?

PR #2192, opened 2018-10-12
plugin/errors: 'consolidate' option

rdrozhdzh:
add 'consolidate' option
see more details at https://github.com/infobloxopen/coredns-plugin-errors/pull/3
#2168
updated README.md

### DIFF
--- a/plugin/errors/README.md
+++ b/plugin/errors/README.md
@@ -6,15 +6,35 @@
 
 ## Description
 
-Any errors encountered during the query processing will be printed to standard output.
+Any errors encountered during the query processing will be printed to standard output. The errors of particular type can be consolidated and printed once per some period of time.
 
 This plugin can only be used once per Server Block.
 
 ## Syntax
 
+The basic syntax is:
+
 ~~~
 errors
 ~~~
+
+Extra knobs are available with an expanded syntax:
+
+~~~
+errors {
+	consolidate DURATION REGEXP
+}
+~~~
+
+Option `consolidate` allows collecting several error messages matching the regular expression **REGEXP** during **DURATION**. After the **DURATION** since receiving the first such message, the consolidated message will be printed to standard output, e.g.
+
+~~~
+2 errors like '^read udp .* i/o timeout$' occurred in last 30s
+~~~
+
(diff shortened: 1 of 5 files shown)


fturib:
@rdrozhdzh : thanks for putting all that together !

fturib:
@miekg : is that ok to merge this PR ?

miekg:
sorry, I'm late reviewing this. Gave it a quick glance, will proper review later today.

One question on all the atomic use, isn't there some nice package we can use to do the ratelimiting and if not, are we sure we've aligned all the structure used correctly for atomic use?

rdrozhdzh:
Hi @miekg, I guess your concern is about the bug mentioned at https://golang.org/pkg/sync/atomic/ , i.e. 

> On both ARM and x86-32, it is the caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically. The first word in a variable or in an allocated struct, array, or slice can be relied upon to be 64-bit aligned. 

In my change three fields are accessed atomically - `pattern.ptimer`, `pattern.count`, `errorHandler.stopFlag`. The last two are 32-bit values, so they should not be affected by the mentioned bug. The `pattern.ptimer` is a pointer and it could be 64-bit value, but anyway it is the first field in an allocated structure.

Tests use more atomically accessed data fields, but they all are 32-bit data.

So, I don't expect any issues regarding data alignment.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/errors..." ]
>Hi @miekg, I guess your concern is about the bug mentioned at https://golang.org/pkg/sync/atomic/ , i.e.
>
>> On both ARM and x86-32, it is the caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically. The first word in a variable or in an allocated struct, array, or slice can be relied upon to be 64-bit aligned.
>
>In my change three fields are accessed atomically - `pattern.ptimer`, `pattern.count`, `errorHandler.stopFlag`. The last two are 32-bit values, so they should not be affected by the mentioned bug. The `pattern.ptimer` is a pointer and it could be 64-bit value, but anyway it is the first field in an allocated structure.
>
>Tests use more atomically accessed data fields, but they all are 32-bit data.
>
>So, I don't expect any issues regarding data alignment.

Good. Good to see you thought about it. :-)

miekg (approved):

PR #2193, opened 2018-10-12
plugin/route53: add alias record support

canthefason:
This PR introduces alias record support to Route53 plugin. The existing implementation just ignores alias records, which ends up with not resolving those records at all. 

As the solution, while iterating through all resource record sets, this PR first adds that alias record as a CNAME record. Then in runtime, it removes that CNAME record and update names for all the answers which the CNAME address gets resolved. 

Before we modify the records, the initial response will be quite similar to this:

```
alias.example.org.	300	IN	CNAME	alias.us-west-2.elb.amazonaws.com.
alias.us-west-2.elb.amazonaws.com. 59 IN	A 1.2.3.4
alias.us-west-2.elb.amazonaws.com. 59 IN	A 1.2.3.5
alias.us-west-2.elb.amazonaws.com. 59 IN	A 1.2.3.6
```

After we modify the response during querying, the answer we return to the client will look like this:
```
alias.example.org. 59 IN	A 1.2.3.4
alias.example.org. 59 IN	A 1.2.3.5
alias.example.org. 59 IN	A 1.2.3.6
```

Signed-off-by: Can Yucel <jon.yucel@getcruise.com>

### DIFF
--- a/plugin/route53/route53.go
+++ b/plugin/route53/route53.go
@@ -34,9 +34,10 @@ type Route53 struct {
 }
 
 type zone struct {
-	id  string
-	z   *file.Zone
-	dns string
+	id    string
+	z     *file.Zone
+	dns   string
+	alias map[string]struct{}
 }
 
 type zones map[string][]*zone
@@ -112,7 +113,8 @@ func (h *Route53) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	m.SetReply(r)
 	m.Authoritative, m.RecursionAvailable = true, true
 	var result file.Result
-	for _, hostedZone := range z {
+	var hostedZone *zone
+	for _, hostedZone = range z {
 		h.zMu.RLock()
 		m.Answer, m.Ns, m.Extra, result = hostedZone.z.Lookup(state, qname)
 		h.zMu.RUnlock()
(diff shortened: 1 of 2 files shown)


yongtang:
AliasTarget is not a DNS standard but a Route53's workaround, so it may not be easy to fit into the existing code base. Having said that, using CNAME as a placeholder in between is not very straight ward to maintain in the future. Wondering if there are other ways to get around.

canthefason:
@yongtang Another option I've been evaluating was resolving the alias records before adding them to the corresponding zones, but it felt a bit against the overall CoreDNS code structure.

canthefason:
@yongtang I gave more thoughts around resolving the alias records before adding the answers to the zones, but it requires me to introduce a custom response writer, which looks more hacky than using CNAMEs as placeholders.

dilyevsky:
Per meatspace conversation with @canthefason we think it makes more sense to rework alias support into a separate go routine loop within this plugin that would just sync standard `A/AAAA` records. This would allow to support healthchecks (https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-values-alias.html#rrsets-values-alias-evaluate-target-health) more easily in the future. @yongtang any objection to this?

yongtang:
@canthefason @dilyevsky Thanks and sorry for the late reply (flooded by other things here and there recently). A go routine with the future potential of health check looks reasonable and interesting to me.

miekg:
this can be closed?

canthefason:
yes, it's ok to be closed at the moment. Unfortunately, I didn't have time to go back and work on this particular issue.

PR #2194, opened 2018-10-13
plugin/file: Fix memory leak in Parse

tmthrgd:

For zone files with more than 10,000 records, the goroutines and memory pinned by dns.ParseZone won't be released unless the tokens chan is drained. As Parse is called by (*Zone).Reload very frequently, this causes memory leaks and OOM conditions.

miekg/dns#786

N/A

### DIFF
--- a/plugin/file/file.go
+++ b/plugin/file/file.go
@@ -121,6 +121,12 @@ func (s *serialErr) Error() string {
 // it returns an error indicating nothing was read.
 func Parse(f io.Reader, origin, fileName string, serial int64) (*Zone, error) {
 	tokens := dns.ParseZone(f, dns.Fqdn(origin), fileName)
+	defer func() {
+		// Drain the tokens chan so that large zone files won't
+		// leak goroutines and memory.
+		for range tokens {
+		}
+	}()
 	z := NewZone(origin, fileName)
 	seenSOA := false
 	for x := range tokens {


plugin/file/file.go
```diff
@@ -121,6 +121,12 @@ func (s *serialErr) Error() string {
 // it returns an error indicating nothing was read.
 func Parse(f io.Reader, origin, fileName string, serial int64) (*Zone, error) {
 	tokens := dns.ParseZone(f, dns.Fqdn(origin), fileName)
+	defer func() {
```
miekg, line 124: This looks like a proper fix, but it sure makes for a messy API. Wonder if something can be done in dns.ParseZone itself.

tmthrgd: Not that I can see without adding some sort of stop signal (which would complicate the ParseZone code). Only the sender can close a chan, and we have no way to signal to the ParseZone goroutines that we’re no longer interested in the tokens. I think draining them like this is the only option and certainly the simplest.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/file: ..." ]
>> @@ -121,6 +121,12 @@ func (s *serialErr) Error() string {
> // it returns an error indicating nothing was read.
> func Parse(f io.Reader, origin, fileName string, serial int64) (*Zone, error) {
> 	tokens := dns.ParseZone(f, dns.Fqdn(origin), fileName)
>+	defer func() {
>
>Not that I can see without adding some sort of stop signal (which would complicate the ParseZone code). Only the sender can close a chan, and we have no way to signal to the ParseZone goroutines that we’re no longer interested in the tokens. I think draining them like this is the only option and certainly the simplest.

Yeah, I skimmed that code a bit and it certainly looks like major engineering to 
tweak that.

I'll update the docs, so we at least have that.

Thanks!

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/file: ..." ]
>> @@ -121,6 +121,12 @@ func (s *serialErr) Error() string {
> // it returns an error indicating nothing was read.
> func Parse(f io.Reader, origin, fileName string, serial int64) (*Zone, error) {
> 	tokens := dns.ParseZone(f, dns.Fqdn(origin), fileName)
>+	defer func() {
>
>Not that I can see without adding some sort of stop signal (which would complicate the ParseZone code). Only the sender can close a chan, and we have no way to signal to the ParseZone goroutines that we’re no longer interested in the tokens. I think draining them like this is the only option and certainly the simplest.

The lexer go-routine can relative easy be removed I think, the state can be 
kept in a new structure and we give `zlexer` a receiver. That's one gone as 
least.

But yeah, for the time being just draining sgtm.

PR #2196, opened 2018-10-13
plugin/dnssec: Add support for KSK/ZSK split key setups

mensi:
The dnssec plugin currently only supports (multiple) ZSKs. This adds support for setups with a ZSK/KSK split. The way it works is that if you specify multiple keys, we check if there's a positive amount of keys with the SEP bit set and also a positive amount of keys without the SEP bit set.

None

References to a limitation to only ZSK can be removed.

### DIFF
--- a/plugin/dnssec/README.md
+++ b/plugin/dnssec/README.md
@@ -21,8 +21,13 @@ dnssec [ZONES... ] {
 }
 ~~~
 
-The specified key is used for all signing operations. The DNSSEC signing will treat this key as a
-CSK (common signing key), forgoing the ZSK/KSK split. All signing operations are done online.
+The signing behavior depends on the keys specified. If multiple keys are specified of which there is
+at least one key with the SEP bit set and at least one key with the SEP bit unset, signing will happen
+in split ZSK/KSK mode. DNSKEY records will be signed with all keys that have the SEP bit set. All other
+records will be signed with all keys that do not have the SEP bit set.
+
+In any other case, each specified key will be treated as a CSK (common signing key), forgoing the
+ZSK/KSK split. All signing operations are done online.
 Authenticated denial of existence is implemented with NSEC black lies. Using ECDSA as an algorithm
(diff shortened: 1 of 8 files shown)


fturib:
I do not think that the error of CI is tied to this code.
And it is also hard to believe that this WaitGroup issue can happen in our server :  the s.dnsWg itself is can of useless (I think the initial intend of usage never hapenned or was deprecated).
The only s.dnsWg.Add is in the creation of the server, and the only s.dnsWg.Done is in a method of that server.

The only case I can see is that the Server object is created without calling dnsServer.NewServer(..) method.
But I do not see how that is happening in the code.


NOTE: the failing UT is a reload one, I mean we are using instance.Restart()...

Isolus:
The code looks good to me and I don't see either why the test fails.

But I think this change can break existing setups. The SEP bit is not mandatory (i.e. enforced) for KSKs. If someone currently uses two keys (one with the SEP bit and a second one without) and both are needed a setup will break after this patch. Therefore we should at least document the SEP bit depending behaviour in the readme and in the release notes.

mensi:
So that's actually a somewhat interesting point. Currently what would happen with multiple keys is that all records are signed with all keys, so you get RRSIGs for each key. This means that even a KSK/ZSK setup will work today - but will generate redundant RRSIGs and thus larger responses.

With this change, the only case where it will break is if you have both SEP-true and SEP-false keys at the same time - if you only have keys of one flavour it will default to the old behavior.

So ultimately the question is whether a setup where you have *both* SEP-true and SEP-false keys is ever intended to be used in "every key signs every RRSet" mode.

As an alternative, I could add another config directive where you can set splitkeys to {enable,disable,autodetect} for backwards compatibility. The default can then either be disable or autodetect.

miekg:
This PR should probably also contain some updates to the README.md and the setup function?

mensi:
Updated the readme. The setup function doesn't really need any changes - what did you have in mind?

miekg (approved):

Issue #2189, opened 2018-10-11
plugin/cache: min TTL defaults should be 0


chrisohaver:
Having a configurable minimum TTL is a nice option (#2055), but it should IMO default to zero.  Otherwise the default behavior will keep some records in cache longer than the authoritative source intended intended - which I don't think is a good default behavior.  

This should be the case for both positive and negative minimum cache TTLs. Currently they both default to 5 seconds.

I think #2055 unintentionally changed the default behavior from a 0 minimum TTL, to 5 seconds TTL.

chrisohaver:
closing - see comment in #2198

aaronriekenberg:
I think it might be reasonable to have 0 as the default minimum TTL for the cache plugin, but just want to note this change was not unintentional in https://github.com/coredns/coredns/pull/2055 to set the default minTTL and minNTTL to dnsutil.MinimalDefaultTTL or 5 seconds.

I originally set the default minTTL and minNTTL to 0 in that PR, and got a review comment from @mikeg to use dnsutil.MinimalDefaultTTL (5s) instead: https://github.com/coredns/coredns/pull/2055#discussion_r214089712

The concern was consistency with the commentary in pkg/dnsutil/ttl.go (https://github.com/coredns/coredns/blob/master/plugin/pkg/dnsutil/ttl.go#L68)

chrisohaver:
> just want to note this change was not unintentional in #2055

Ah OK, Thanks for clarifying.  I didn't see that comment.

PR #2198, opened 2018-10-15
plugin/cache: Default min cache TTL to zero

chrisohaver:
* Make default minimum TTL be zero, so the default behavior is not to cache entries for longer than the TTL set by authoritative upstream servers.
* Add min TTL default value to readme.

#2189

included

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -238,9 +238,9 @@ func (w *ResponseWriter) Write(buf []byte) (int, error) {
 }
 
 const (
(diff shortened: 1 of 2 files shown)


grobie (approved):
I was never a fan of overriding the authoritative decision of DNS servers, so :+1: from my side. I guess @miekg should describe his motivation though.

chrisohaver:
Reading the code more carefully... it appears that the constant `MinimalDefaultTTL`, despite its name and comment, is only used when creating a TTL for a response that has no TTL.  I.e. These are the "default negative response TTLs".

Closing...

PR #2199, opened 2018-10-15
plugin/cache: Set min TTL default to zero

chrisohaver:

IIUC, #2055 changed default behavior of caching:

Before #2055, we used the message TTL when caching, even if the message had a TTL below the `MinimalDefaultTTL`.  `MinimalDefaultTTL` was only used in cache when synthesizing TTLs for negative responses that had no SOA record.

After #2055, if any message has a TTL below the `MinimalDefaultTTL`, the cache uses the `MinimalDefaultTTL` value instead of the message TTL.

Setting the default values for `minTTL` and `minNTTL` to 0 instead of `MinimalDefaultTTL` should restore the original default behavior.

#2189

### DIFF
--- a/plugin/cache/README.md
+++ b/plugin/cache/README.md
@@ -41,10 +41,10 @@ cache [TTL] [ZONES...] {
 * **TTL**  and **ZONES** as above.
 * `success`, override the settings for caching successful responses. **CAPACITY** indicates the maximum
   number of packets we cache before we start evicting (*randomly*). **TTL** overrides the cache maximum TTL.
-  **MINTTL** overrides the cache minimum TTL, which can be useful to limit queries to the backend.
+  **MINTTL** overrides the cache minimum TTL (default 0), which can be useful to limit queries to the backend.
 * `denial`, override the settings for caching denial of existence responses. **CAPACITY** indicates the maximum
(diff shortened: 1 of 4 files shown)


grobie:
Sounds like we should add a test to ensure the cache plugin doesn't extends the original message TTL.

fturib (approved):
LGTM.

fturib:
@grobie : this PR is about ensuring that, BY DEFAULT, the cache will not extend the original TTL. 
But, if you change the minTTL of the cache - in purpose in the Corefile -, then the original TTL will be increased to match the minTLL of the cache.
Looks like it is expected behavior of the PR that introduced the minTTL/maxTTL for cache.
see #2055.

fturib:
hum... maybe you are right a UT that check we do not extend by default would ensure nobody will change that default behavior.

fturib (changes requested):
Please add UT to check that TTL is not change for: replies with TTL=0, TTL=1 at least.

chrisohaver:
@fturib, test added

fturib (approved):

chrisohaver:
Would have been good to get this into 1.2.3, so we didn't have a release with the regression present.

grobie (approved):

Issue #2206, opened 2018-10-17
Metrics counters leaking between servers


bilbercode:
Hey Folks 
I have noticed that the metric collectors bleed between servers due to the way that we are setting up and registering them. 

Using the following configuration I would expect metrics to be isolated

```
 foo.bat:53 {
  prometheus :8081
  forward . 8.8.8.8 8.8.4.4
}

fizz.bang:53 {
  prometheus: 8082
  forward . 1.1.1.1 1.0.0.1
}
```
however upon checking the metrics they bleed between each other (report on each others hits). This is not behaviour I was expecting. It would be better for reporting purposes if we instantiated unique collectors per server we create as opposed to creating them per plugin globally. I am currently working on a fix for this for you and will submit a WIP PR shortly so we can go though it and hopefully get it fixed 👍

miekg:
this should be mitigated with the server label. Why is that not working for you?

miekg:
This looks stale or fixed with #2246

PR #2207, opened 2018-10-17
WIP feat: metrics isolation per server

bilbercode:
Isolates the creation of prometheus collectors to a server per server basis to prevent statistical leaks
#2206 
None


bilbercode:
Currently WIP

miekg:
I still don't see the bug here, and we get the current server from the context now.

Issue #2205, opened 2018-10-17
Route53 fallthrough to Route53 - Credential Bug


SuperFunks:
Hi there,

I'm trying to fallthrough route53 to another route53 section. Seems like that currently has a bug where the last profile in the credentials lines is used on all route53 sections. This of course gives access denied on the other zones and keeps CoreDNS from loading the zones.

Example (ID's have been changed):

```
.{
        route53 zone1.com.:Z2CDP34BQB3GIC {
                credentials 280846555434 /tmp/credentials
                fallthrough
        }
        route53 zone2.com.:Z3QSOTUJVXOESI {
                credentials 280846555434 /tmp/credentials
                fallthrough
        }
        route53 zone3.com.:Z2GB9FKQHZH7FN {
                credentials 188228134507 /tmp/credentials
        }
}
```

AccessDenied: User: arn:aws:sts::188228134507:assumed-role/ADFS-Viewer/coredns is not authorized to access this resource

Note that this works fine with the 2 first entries in there as they have the same credential profile - but adding the last one breaks the entire thing.

Hoping this could be fixed soon :)

Best regards,

SuperFunks

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] Route53 fallthrou..." ]
>Hi there,
>
>I'm trying to fallthrough route53 to another route53 section. Seems like that currently has a bug where the last profile in the credentials lines is used on all route53 sections. This of course gives access denied on the other zones and keeps CoreDNS from loading the zones.
>
>Example (ID's have been changed):
>
>.{
>        route53 zone1.com.:Z2CDP34BQB3GIC {
>                credentials 280846555434 /tmp/credentials
>                fallthrough
>        }
>        route53 zone2.com.:Z3QSOTUJVXOESI {
>                credentials 280846555434 /tmp/credentials
>                fallthrough
>        }
>        route53 zone3.com.:Z2GB9FKQHZH7FN {
>                credentials 188228134507 /tmp/credentials
>        }
>}
>
>AccessDenied: User: arn:aws:sts::188228134507:assumed-role/ADFS-Viewer/coredns is not authorized to access this resource
>
>Note that this works fine with the 2 first entries in there as they have the same credential profile - but adding the last one breaks the entire thing.
>
>Hoping this could be fixed soon :)

Should be fairly simple, but not all plugins allow multiple stanza in the 
corefile as this requires some administration the plugin's code, as in this case 
[...]

PR #2218, opened 2018-10-19
Fix multiple credentials in route53

yongtang:

This fix tries to to address the issue in #2205 while multiple credentials in route53 plugin will be overwritten by previous ones.

#2205 

n/a

### DIFF
--- a/plugin/route53/setup.go
+++ b/plugin/route53/setup.go
@@ -44,8 +44,7 @@ func setup(c *caddy.Controller, f func(*credentials.Credentials) route53iface.Ro
 	// With that said, even though a user doesn't define any credentials in
 	// Corefile, we should still attempt to read the default credentials file,
 	// ~/.aws/credentials with the default profile.
-	sharedProvider := &credentials.SharedCredentialsProvider{}
-	var providers []credentials.Provider
+	providers := []credentials.Provider{&credentials.EnvProvider{}, &credentials.SharedCredentialsProvider{}}
 	var fall fall.F
 
 	up, _ := upstream.New(nil)
@@ -95,6 +94,7 @@ func setup(c *caddy.Controller, f func(*credentials.Credentials) route53iface.Ro
 					return c.Errf("invalid upstream: %v", err)
 				}
 			case "credentials":
+				sharedProvider := &credentials.SharedCredentialsProvider{}
 				if c.NextArg() {
 					sharedProvider.Profile = c.Val()
 				} else {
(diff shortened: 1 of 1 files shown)


dilyevsky:
Hm does this actually fix this? Not sure if chained provider is intelligent enough to pick an appropriate credentials for the zone. I think it might just try the first one set and fail. aws docs aren't super clear on that...

yongtang:
I tried with `non-existing provider + valid provider` and it looks to be working. Though whether or not it goes to zone level is unknown. I will try to see if I could setup different credentials with different account to verify.

plugin/route53/setup.go
```diff
@@ -44,8 +44,7 @@ func setup(c *caddy.Controller, f func(*credentials.Credentials) route53iface.Ro
 	// With that said, even though a user doesn't define any credentials in
 	// Corefile, we should still attempt to read the default credentials file,
 	// ~/.aws/credentials with the default profile.
-	sharedProvider := &credentials.SharedCredentialsProvider{}
-	var providers []credentials.Provider
+	providers := []credentials.Provider{&credentials.EnvProvider{}, &credentials.SharedCredentialsProvider{}}
```
canthefason, line 47: Passing an empty SharedCredentialsProvider ends up with incorrect credentials. This is the message I get: 
```
Error during parsing: failed to create Route53 plugin: ExpiredToken: The security token included in the request is expired"
```

canthefason:
I was able to test this out and unfortunately, it didn't work. There are two reasons for it. The first one is the empty SharedCredentialsProvider that I already left a comment ends up with incorrect credentials which is an easy fix. The second reason is, when we add multiple shared credential providers [ListHostedZonesByNameWithContext](https://github.com/coredns/coredns/blob/master/plugin/route53/route53.go#L54) uses the very first shared credentials we pass and as expected, errors out when we query a hosted zone that belongs to the other accounts.

canthefason:
Besides the credentials file, this should also be an issue for static credentials, as well. IMHO, unless AWS clients allow switching between credentials, we need to initialize multiple clients with different sessions instead of passing one for each top-level stanza.

yongtang:
@canthefason Yes I think the per host/session might make sense. Will take a look and see if I could update the PR>

yongtang:
Has been flooded by other things recently, still working on it. Will update (hopefully) soon.

miekg:
closing this for the time being

PR #2219, opened 2018-10-20
plugin/file: Use new zone parser API

miekg:
Use new dns lib 1.0.14 and default to using the new zone parser that
does not leak go-routines.


dilyevsky (approved):
LGTM

dilyevsky (approved):
LGTM

PR #2020, opened 2018-08-02
Invite users to fill the survey

fturib:
### 1. Why is this pull request needed and what does it do?

Invite users of CroeDNS in conjunction with kubernetes to fill the surver "CoreDNS adoption".
This is a proposition.
Top of the page has more change to be read.

Let me know what you think.

NOTE: That survey may be alive or several months

PR #2222, opened 2018-10-21
Revert "Invite users to fill the survey (#2020)"

miekg:
This reverts commit 073708bc6604585d257234128cb71a945c675691.

Has been running for 3 months, let's remove it again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,18 +7,6 @@
 [![Go Report Card](https://goreportcard.com/badge/github.com/coredns/coredns)](https://goreportcard.com/report/coredns/coredns)
 [![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/1250/badge)](https://bestpractices.coreinfrastructure.org/projects/1250)
 
-
-
-## Feedback wanted ...
-
-CoreDNS is now part of [Kubernetes](http://kubernetes.io), [delivering the DNS and Discovery Service for the cluster](https://kubernetes.io/blog/2018/07/10/coredns-ga-for-kubernetes-cluster-dns/).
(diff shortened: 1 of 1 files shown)


fturib (approved):
I was expecting to keep that survey until KubeCon.
But I agree the rate of reply is now very low.

We have total of 98 replies - that was really helpful to trigger the decision to integrate CoreDNS as default in Kubernetes (initial target of this survey).

An unformal summary of the replies is available here : https://docs.google.com/spreadsheets/d/1yhDyAMm5U-ueFE4qA9s3vjCIT5kFsKWILn73c16ewjo/edit?usp=sharing.

NOTE: there is useful replies on what kubernetes users would expect as next features for CoreDNS.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Revert "Invit..." ]
>fturib approved this pull request.
>
>I was expecting to keep that survey until KubeCon.
>But I agree the rate of reply is now very low.
>
>We have total of 98 replies - that was really helpful to trigger the decision to integrate CoreDNS as default in Kubernetes (initial target of this survey).
>
>An unformal summary of the replies is available here : https://docs.google.com/spreadsheets/d/1yhDyAMm5U-ueFE4qA9s3vjCIT5kFsKWILn73c16ewjo/edit?usp=sharing.
>
>NOTE: there is useful replies on what kubernetes users would expect as next features for CoreDNS.


Cool, that's an nice list. Might be worth opening issues for each of them, some 
we can do already, some not, and some require major surgery.

/Miek

--
Miek Gieben

PR #2225, opened 2018-10-22
server: actually scrub response

miekg:
Did all the worked, hooked it up wrongly :(

This also needs test, but those are hard(er) because we only receive
packets after they have been decoded; i.e. we never see the wirefmt.

### DIFF
--- a/request/request.go
+++ b/request/request.go
@@ -226,11 +226,7 @@ func (r *Request) SizeAndDo(m *dns.Msg) bool {
 	return true
 }
 
-// Scrub is a noop function, added for backwards compatibility reasons. The original Scrub is now called
-// automatically by the server on writing the reply. See ScrubWriter.
-func (r *Request) Scrub(reply *dns.Msg) (*dns.Msg, int) { return reply, 0 }
-
-// scrub scrubs the reply message so that it will fit the client's buffer. It will first
+// Scrub scrubs the reply message so that it will fit the client's buffer. It will first
 // check if the reply fits without compression and then *with* compression.
 // Scrub will then use binary search to find a save cut off point in the additional section.
 // If even *without* the additional section the reply still doesn't fit we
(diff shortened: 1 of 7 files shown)


miekg:
oh nice. Our CI integration is gone after github's database failover...

miekg:
Ok, tested added. PTAL.

request/request.go
```diff
@@ -226,19 +226,15 @@ func (r *Request) SizeAndDo(m *dns.Msg) bool {
 	return true
 }
 
-// Scrub is a noop function, added for backwards compatibility reasons. The original Scrub is now called
-// automatically by the server on writing the reply. See ScrubWriter.
```
johnbelamaric, line 230: Was this comment incorrect? I mean, was the real Scrub never getting called in this case? So, we're not calling it twice now are we?

johnbelamaric (approved):
I think this is fine but one question...

miekg:
Good point. It's the old comment I copied. I'll update

request/request.go
miekg: Let me properly reply: no we're not calling it twice: Scrub is the only method left. The *old* Scrub was a noop, the new one is not.

johnbelamaric: Ok, so we weren't calling Scrub at all in the most recent release? Or just in some cases? That's what's confusing me. Maybe I just need to dig into the previous PR to understand it.

miekg: yes, we we're calling the noop function.

johnbelamaric:
Eww. I am surprised more problems haven't come up.

/lgtm

johnbelamaric:
Looks like this is breaking the rewrite tests.

miekg:
huh, I don't understand that failure. The package send to Scrub also only contains the mx record @greenpau

miekg:
broke *erratic* by adding any RRs if e.large wasn't set.

greenpau:
@miekg, this is merged. Do you still need me to look over the test?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] server: actua..." ]
>@miekg, this is merged. Do you still need me to look over the test?

thanks, but not needed. I made a booboo in the eratic plugin, luckily this test 
caught that :) All good know.

PR #2227, opened 2018-10-22
plugin/cache: Make default min TTL 5

chrisohaver:

#2055 contained an _intentional_ (but undocumented) change in default behavior that I incorrectly identified as an unintentional regression.  This PR reverts #2199,  but leaves (adjusted) documentation in place so the behavior is documented somewhere and also leaves (adjusted) unit tests in place so the behavior is tested.

#2055 #2189
partly reverts #2199


greenpau (approved):
/lgtm

miekg:
>greenpau approved this pull request.
>
>/lgtm

/merge

PR #2240, opened 2018-10-25
log/forward plugins: Extend dns query logging

drazhanski:
In case using forward plugins it could be needed to see which of the forward resolvers replied to the query. To extend logging information the additional place holders were added to the log plugin.
It expected to be useful to understand customer's issues at least for Infoblox ATEP client which used Coredns.
Documentation for the log and probably the forward plugins needs to be updated. The changes in the corresponding README.md files were provided.

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -91,6 +91,10 @@ On each endpoint, the timeouts of the communication are set by default and autom
 * dialTimeout by default is 30 sec, and can decrease automatically down to 100ms
 * readTimeout by default is 2 sec, and can decrease automatically down to 200ms
 
+## Metadata
+
+* forward/resolving_proxy : provide the IP address and port of the upstream resolver used to resolve the current DNS query.
+
 ## Metrics
 
 If monitoring is enabled (via the *prometheus* directive) then the following metric are exported:
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/debug"
+	"github.com/coredns/coredns/plugin/metadata"
 	clog "github.com/coredns/coredns/plugin/pkg/log"
 	"github.com/coredns/coredns/request"
 
@@ -58,6 +59,15 @@ func (f *Forward) Len() int { return len(f.proxies) }
 
 // Name implements plugin.Handler.
 func (f *Forward) Name() string { return "forward" }
+
+//declareMetadata adds to the context a metadata parameter which will return the passed value.
+func (f *Forward) declareMetadata(ctx context.Context, name string, value string) bool {
+	label := f.Name()+"/"+name
+	if metadata.IsLabel(label) {
+		return metadata.SetValueFunc(ctx, label, func() string { return value })
+	}
+	return false
+}
 
 // ServeDNS implements plugin.Handler.
 func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
@@ -148,10 +158,12 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 
 			formerr := state.ErrorMessage(dns.RcodeFormatError)
 			w.WriteMsg(formerr)
+			f.declareMetadata(ctx, "resolving_proxy", proxy.addr)
 			return 0, nil
 		}
 
 		w.WriteMsg(ret)
+		f.declareMetadata(ctx, "resolving_proxy", proxy.addr)
 		return 0, nil
 	}
 
--- a/plugin/log/README.md
+++ b/plugin/log/README.md
@@ -72,6 +72,10 @@ The following place holders are supported:
 * `{>do}`: is the EDNS0 DO (DNSSEC OK) bit set in the query
 * `{>id}`: query ID
 * `{>opcode}`: query OPCODE
+* `{local}`: server's IP address, for IPv6 addresses these are enclosed in brackets: `[::1]`
+* `{A}`: number of the A type answers in the response
+* `{AAAA}`: number of the AAAA type answers in the response
+* `{forward/resolving_proxy}`: if the query was forwarded to a resolver, the IP address and port of that replied resolver, else empty. The metadata plugin needs to be enabled to use this placeholder
 
 The default Common Log Format is:
 
--- a/plugin/pkg/replacer/replacer.go
+++ b/plugin/pkg/replacer/replacer.go
@@ -1,11 +1,13 @@
 package replacer
 
 import (
+	"context"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/coredns/coredns/plugin/metadata"
 	"github.com/coredns/coredns/request"
 
 	"github.com/miekg/dns"
@@ -31,7 +33,7 @@ type replacer struct {
 // values into the replacer. rr may be nil if it is not
 // available. emptyValue should be the string that is used
 // in place of empty string (can still be empty string).
-func New(r *dns.Msg, rr *dnstest.Recorder, emptyValue string) Replacer {
+func New(ctx context.Context, r *dns.Msg, rr *dnstest.Recorder, emptyValue string) Replacer {
 	req := request.Request{W: rr, Req: r}
 	rep := replacer{
 		replacements: map[string]string{
@@ -45,6 +47,8 @@ func New(r *dns.Msg, rr *dnstest.Recorder, emptyValue string) Replacer {
 			"{size}":   strconv.Itoa(req.Len()),
 			"{remote}": addrToRFC3986(req.IP()),
 			"{port}":   req.Port(),
+			"{local}":  addrToRFC3986(req.LocalIP()),
+			"{forward/resolving_proxy}": getMetadata(ctx, "forward/resolving_proxy"),
 		},
 		emptyValue: emptyValue,
 	}
@@ -58,6 +62,8 @@ func New(r *dns.Msg, rr *dnstest.Recorder, emptyValue string) Replacer {
 		rep.replacements["{duration}"] = strconv.FormatFloat(time.Since(rr.Start).Seconds(), 'f', -1, 64) + "s"
 		if rr.Msg != nil {
 			rep.replacements[headerReplacer+"rflags}"] = flagsToString(rr.Msg.MsgHdr)
+			rep.replacements["{A}"] = answersCount(rr.Msg, dns.TypeA)
+			rep.replacements["{AAAA}"] = answersCount(rr.Msg, dns.TypeAAAA)
 		}
 	}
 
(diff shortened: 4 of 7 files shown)


plugin/pkg/replacer/replacer.go
```diff
@@ -163,6 +169,30 @@ func addrToRFC3986(addr string) string {
 	return addr
 }
 
+//getMetadata will return value from Metadata or empty string
+func getMetadata(ctx context.Context, name string) string {
+	if ctx != nil {
+		proxy_prm := metadata.ValueFunc(ctx, name)
```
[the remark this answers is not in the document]
drazhanski: Fixed

plugin/forward/forward.go
```diff
@@ -59,6 +60,11 @@ func (f *Forward) Len() int { return len(f.proxies) }
 // Name implements plugin.Handler.
 func (f *Forward) Name() string { return "forward" }
 
+//declareMetadata adds to the context a metadata parameter which will return the passed value.
+func (f *Forward) declareMetadata(param_name string, param_value string, ctx context.Context) {
```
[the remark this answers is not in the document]
drazhanski: Fixed

plugin/forward/README.md
```diff
@@ -39,7 +39,8 @@ forward FROM TO...
   limited to 15.
 
 Multiple upstreams are randomized (see `policy`) on first use. When a healthy proxy returns an error
-during the exchange the next upstream in the list is tried.
+during the exchange the next upstream in the list is tried. Metadata parameter "forward/resolving_proxy" 
+contains the ip-address and the port of the replied upstream on success, see the log plugin for detail.
```
fturib: I think we should make a specific section **Metadata** the same way we have a **Metrics** section.
And list the each metadata information we are gathering in this plugin.

Something like:

### Metadata

* `forward/resolving_proxy` : provide the IP and port of the upstream resolver used to resolve current DNS query.

fturib: No : the plugins are independents ... it is better to not mention the log plugin here. (IMO)

plugin/forward/forward.go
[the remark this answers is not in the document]
fturib: sorry .. my bad. Looks like you need to reorder the parameters

plugin/log/README.md
```diff
@@ -72,6 +72,10 @@ The following place holders are supported:
 * `{>do}`: is the EDNS0 DO (DNSSEC OK) bit set in the query
 * `{>id}`: query ID
 * `{>opcode}`: query OPCODE
+* `{local}`: server's IP address, for IPv6 addresses these are enclosed in brackets: `[::1]`
+* `{A}`: number of the A type answers in the reply
+* `{AAAA}`: number of the AAAA type answers in the reply
+* `{resolver}`: the IP address and port of the proxy which replied to the query (used with the forward plugin only)
```
fturib: I would propose:

* `{resolver}`: if query was forwarded to a resolver, the IP address and port of that resolver, else empty

fturib (commented):
some comment.
Mostly on documentation

plugin/pkg/replacer/replacer.go
```diff
@@ -31,7 +33,7 @@ type replacer struct {
 // values into the replacer. rr may be nil if it is not
 // available. emptyValue should be the string that is used
 // in place of empty string (can still be empty string).
-func New(r *dns.Msg, rr *dnstest.Recorder, emptyValue string) Replacer {
+func New(ctx context.Context, r *dns.Msg, rr *dnstest.Recorder, emptyValue string) Replacer {
```
rajansandeep, line 36: You'll have to update the tests to factor in the context.

drazhanski: Yes, you are right. I fixed it in the tests.

plugin/forward/README.md
drazhanski: Ok, added this section.

drazhanski: Ok, fixed.

plugin/log/README.md
drazhanski: Ok. I only added "that replied resolver"

plugin/pkg/replacer/replacer.go
```diff
@@ -45,6 +47,8 @@ func New(r *dns.Msg, rr *dnstest.Recorder, emptyValue string) Replacer {
 			"{size}":   strconv.Itoa(req.Len()),
 			"{remote}": addrToRFC3986(req.IP()),
 			"{port}":   req.Port(),
+			"{local}":  addrToRFC3986(req.LocalIP()),
+			"{resolver}": getMetadata(ctx, "forward/resolving_proxy"),
```
johnbelamaric: Two things:

1) in the docs, we should note which place holders require the metadata plugin to be enabled
2) the placeholders should match the metadata key

johnbelamaric: In fact, if we find an "invalid" placeholder, perhaps we should treat it as a metadata key? There are pros and cons. It means we don't need to touch this code in order to add things - it automatically can log whatever metadata there is. But it means we can't check the validity of placeholders, so typos could go unnoticed at parse time (do we do that anyway?).

plugin/pkg/replacer/replacer.go
```diff
@@ -58,6 +62,8 @@ func New(r *dns.Msg, rr *dnstest.Recorder, emptyValue string) Replacer {
 		rep.replacements["{duration}"] = strconv.FormatFloat(time.Since(rr.Start).Seconds(), 'f', -1, 64) + "s"
 		if rr.Msg != nil {
 			rep.replacements[headerReplacer+"rflags}"] = flagsToString(rr.Msg.MsgHdr)
+			rep.replacements["{A}"] = answersCount(rr.Msg, dns.TypeA)
```
johnbelamaric, line 65: I think in a separate PR, we should look into whether we can make all these replacer values do lazy-evaluation, so we only take the time hit if they are used.

fturib: I am thinking that if we move to use metadata instead of the replacement ... then we will have the lazy-evaluation and we can add any metadata reference we want.
it will simplify also dependency between Log and any other plugin.

fturib:
I think the Metadata enhancement proposed should be in another PR.
Let's fix here the only part of this resolver_proxy.
and think about moving Log using more often Metadata (I started to work on it)

plugin/pkg/replacer/replacer.go
fturib: the format of labels in metadata is <plugin>/<name>, so it should be easy to check if it 'looks like' a metadata or if it is an error.

plugin/pkg/replacer/replacer.go
fturib: If we want to keep lazy-initialization, we cannot work like a Replacer (that is trying to replace any possible label), we need to parse and query only the label that are in the format.
We need a parsing/checking at setup.

plugin/pkg/replacer/replacer.go
drazhanski: 1. Ok, I added it to log plugin README.md
2. What did you mean? Should we have {resolving_proxy} or {forward/resolving_proxy} placeholders? Or something else?

johnbelamaric: {forward/resolving_proxy}, and if the placholder contains a /, then you getMetaData for it - but as @fturib said in another comment, we probably can't do that last part without a refactor. So for now, just change the placeholder name so it won't change the interface later when we implement that.

drazhanski: Ok, I changed the placeholder to be {forward/resolving_proxy}.

fturib:
I prefer with the last change : when we move with generic metadata usage, it will keep the same name.

plugin/pkg/replacer/replacer.go
fturib: I am wrong in former comment. 
We can keep the Replacer and have Lazy-initialization with Metadata. And also add lazy initialization for existing "variables"

drazhanski:
@fturib just for assurance if I did not get your last comment. I don't need to return any code back, do I?

fturib:
restating my last comment:
I like that the name is now {forward/resolving_proxy}.
Doing so, there will be no change of name when we will add the ability to log any metadata information.

fturib (approved):
LGTM
Metadata for Log with lazy-init is expected for another PR.

miekg:
I wanted to look at this, but now it's in. I still want to look at it. Also the commit message it got merged with is 'extend metadata logging' *not* the extensive commit that's was typed in the PR, so that's now sadly lost.

miekg:
Couple of things of the bat: I can see this gives useful feedback on it would be nice to have it. 
* {`local}` is a nice addition and mirrors `{remote}` (should also be put slightly higher in the list)
* the number of RRs of a specific type is super weird to log, also it so slow that I wonder why you would even enable this? If the question is for A, you almost (not completely) guaranteed the answer section *count* will be the number you're looking for.
* A and AAAA should not be logged (maaaaaybe debug logging in the solver plugin), but not as generic place holders in the log plugin. This opens the floodgates to other things. Think you're lookgin for dnstap or maybe metrics.
* I thought the whole idea of metadata was that your implement the metadata.Provider interface. Here `declareMetadata` is used. If metadata.Provider doesn't work we should make it work, extend the interface or remove entirely?
* doing metadata type things is a nice idea, might be worth specifying a new prefix, ala the `>` that's already there? Maybe '/' would work here. Then I would be able to log any metadata if it exists.

fturib:
About metadata:

### usage of `metadata.Provider` and `declareMetadata`
Here `declareMetadata` is in the `forward` plugin (not `log`). 
- you can use `declareMetadata` function through the `Provider` interface or not.
- the `Provider` interface allow you to declare your metadata BEFORE the plugin chain is executed. 
- in current case, it is useless because the `resover_proxy` information is not known in front. it is known only after the `forward.serveDns()` is called
- because of the lazy initialization, we could provide the metadata in front. But here, there is no easy way to transmit the right information from a `serveDNS` local variable to a function created at `Provider` time. So the best was to add the metadata on the fly, when the information is known.

### Prefix for metadata in Log
we decided the metadata would have a label like <plugin>/<name>. That maybe is not very convenient for the informations based on the request data, because it always make a long label.
NOTE: I am ok with `>name`. 

An issue we will soon hit is ensure that those data are not provided as metadata by several plugins ... I though that Metadata would be the best location to define the 'standard' metadata. and maybe here we can have simplified labels.

fturib:
Counting the A or AAAA records in the msg could be out of `Log` plugin.
One the plugin support completely Metadata, any specific usage like this one can be define in the external plugins of the user.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] log/forward p..." ]
>About metadata:
>
>### usage of `metadata.Provider` and `declareMetadata`
>Here `declareMetadata` is in the `forward` plugin (not `log`).
>- you can use `declareMetadata` function through the `Provider` interface or not.
>- the `Provider` interface allow you to declare your metadata BEFORE the plugin chain is executed.
>- in current case, it is useless because the `resover_proxy` information is not known in front. it is known only after the `forward.serveDns()` is called

yes, and this was known when we did the metadata plugin, so if that turns out 
not be useful we should remove it.

>- because of the lazy initialization, we could provide the metadata in front. But here, there is no easy way to transmit the right information from a `serveDNS` local variable to a function created at `Provider` time. So the best was to add the metadata on the fly, when the information is known.

The log plugin wraps the writter, maybe something (cleaner) can be implemented 
there; if that turns out hard/impossible we found a clear limitation.

>### Prefix for metadata in Log
>we decided the metadata would have a label like <plugin>/<name>. That maybe is not very convenient for the informations based on the request data, because it always make a long label.
>NOTE: I am ok with `>name`.

Starting with a `/` will make it different and fast to check. You can then also 
make it generic.

>An issue we will soon hit is ensure that those data are not provided as metadata by several plugins ... I though that Metadata would be the best location to define the 'standard' metadata. and maybe here we can have simplified labels.

well, we it when we get there, I think we have 3 labels or less currently being 
used in tree.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] log/forward p..." ]
>Counting the A or AAAA records in the msg could be out of `Log` plugin.

yes, that doesn't belong here.

>One the plugin support completely Metadata, any specific usage like this one can be define in the external plugins of the user.

This needs a bit a think to make it work cleanly, currenty this looks like a 
hack for a specific use-case.

miekg:
We probably want to release soonish. This can't go out in a release as it
adds bit that are reverted in a currently active PR.

PR #2243, opened 2018-10-27
Check SO_REUSEPORT support

stuartnelson3:

The code currently assumes that, if compiled with go1.11, the underlying system has SO_REUSEPORT available.

coredns fails to start on older kernels (pre-3.9) if compiled with go1.11


miekg:
happy to drop my PR #2244, but we can just set it and ignore the error (and log).

stuartnelson3:
Up to you, code should be the same between our two branches.

PR #2245, opened 2018-10-27
logging: unify pkg/log and plugin/log

miekg:
Default to using pkg/log for all logging output and make that always use
UTC.

### DIFF
--- a/coremain/run.go
+++ b/coremain/run.go
@@ -62,7 +62,7 @@ func Run() {
 	}
 
 	log.SetOutput(os.Stdout)
-	log.SetFlags(log.LstdFlags)
+	log.SetFlags(0) // Set to 0 because we're doing our own time, with timezone
 
 	if version {
 		showVersion()
--- a/plugin/log/README.md
+++ b/plugin/log/README.md
@@ -7,9 +7,10 @@
 ## Description
 
 By just using *log* you dump all queries (and parts for the reply) on standard output. Options exist
-to tweak the output a little.
+to tweak the output a little. The date/time prefix on log lines is RFC3339 formatted with
+milliseconds.
 
-Note that for busy servers this will incur a performance hit.
+Note that for busy servers logging will incur a performance hit.
 
 ## Syntax
 
@@ -59,7 +60,6 @@ The following place holders are supported:
 * `{name}`: qname of the request
 * `{class}`: qclass of the request
 * `{proto}`: protocol used (tcp or udp)
-* `{when}`: time of the query
 * `{remote}`: client's IP address, for IPv6 addresses these are enclosed in brackets: `[::1]`
 * `{size}`: request size in bytes
 * `{port}`: client's port
@@ -76,8 +76,14 @@ The following place holders are supported:
 The default Common Log Format is:
 
 ~~~ txt
-`{remote}:{port} - [{when}] {>id} "{type} {class} {name} {proto} {size} {>do} {>bufsize}" {rcode} {>rflags} {rsize} {duration}`
+`{remote}:{port} - {>id} "{type} {class} {name} {proto} {size} {>do} {>bufsize}" {rcode} {>rflags} {rsize} {duration}`
 ~~~
+
+Each of these logs will be outputted with `log.Infof`, so a typical example looks like this:
+
+~~~ txt
+2018-10-30T19:10:07.547Z [INFO] [::1]:50759 - 29008 "A IN example.org. udp 41 false 4096" NOERROR qr,rd,ra,ad 68 0.037990251s
+~~~~
 
 ## Examples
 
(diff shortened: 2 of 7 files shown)


plugin/log/README.md
```diff
@@ -7,7 +7,7 @@
 ## Description
 
 By just using *log* you dump all queries (and parts for the reply) on standard output. Options exist
-to tweak the output a little.
+to tweak the output a little. The timezone is used is always UTC.
```
fturib: typo. "The timezone is always UTC"

coremain/run.go
```diff
@@ -62,7 +62,7 @@ func Run() {
 	}
 
 	log.SetOutput(os.Stdout)
-	log.SetFlags(log.LstdFlags)
+	log.SetFlags(log.Ldate | log.Ltime | log.LUTC)
```
fturib: Changing to UTC will surprise everyone, and the date does not say the TZ.
So maybe, while waiting we implement a complete date with TZ in our pkg/log, we can output an info here:
```
	clog.Info("log output - all dates are UTC")
```

Result is:

```
API server listening at: 127.0.0.1:55958
2018/10/29 14:51:55 [INFO] log output - all dates are UTC
.:1053
.:1054
2018/10/29 14:51:55 [INFO] CoreDNS-1.2.5
2018/10/29 14:51:55 [INFO] darwin/amd64, go1.10.2, 
CoreDNS-1.2.5
darwin/amd64, go1.10.2, 
2018/10/29 14:51:55 [INFO] plugin/reload: Running configuration MD5 = 67fbea4124263c10c9c1aa0fb55a538f
```

fturib (commented):
See my comments about UTC information proposition.
+ glitch in one comment.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] logging: unif..." ]
>+to tweak the output a little. The timezone is used is always UTC.
>
>typo. "The timezone is always UTC"

done.

>> @@ -62,7 +62,7 @@ func Run() {
> 	}
>
> 	log.SetOutput(os.Stdout)
>-	log.SetFlags(log.LstdFlags)
>+	log.SetFlags(log.Ldate | log.Ltime | log.LUTC)
>
>Changing to UTC will surprise everyone, and the date does not say the TZ.

it does not say the timezone because it's UTC, but I agree the std log package 
is pretty brain dead in this regard. Logging in local time with a time zone is 
even crazier.

~~~
>2018/10/29 14:51:55 [INFO] log output - all dates are UTC
~~~

this doesn't add much to saying it in the README (i.e. both will not be read)

/Miek

--
Miek Gieben

miekg:
I'll update this to spit the date in w/ a timezone and RFC3339

fturib (approved):
LGTM

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] logging: unif..." ]
>fturib approved this pull request.
>
>LGTM

thanks we have to decide how backwards incompatible this is; can we just push it 
in the next or will this lead to any degraded experience.

fturib:
I was thinking that if you enable Log plugin, then you already know you will have no perf ...
But maybe I am wrong.

I know Log is not enable on the default Corefile for Kubernetes  

@rdrozhdzh, @drazhanski, @huynhmb359  - will you be impacted if the log is less performant  ?

plugin/pkg/log/log.go
```diff
@@ -1,31 +1,36 @@
...
 var D bool
 
+// RFC3339Milli doesn't exist, invent it here.
+func clock() string { return time.Now().Format("2006-01-02T15:04:05.999Z07:00") }
+
 // logf calls log.Printf prefixed with level.
 func logf(level, format string, v ...interface{}) {
-	s := level + fmt.Sprintf(format, v...)
-	golog.Print(s)
+	golog.Print(clock() + " " + level + fmt.Sprintf(format, v...))
```
rdrozhdzh: minor optimizations -
1. `" " + level` - the leading space can be added to constants like `info = " [INFO] "`
2. I suppose the code `golog.Print(clock(), level, fmt.Sprintf(format, v...))` will concatenate strings more effectively

the same for `func log()` below

rdrozhdzh (approved):
LGTM
I don't think we'll get any significant performance penalty with this implementation. Earlier, golang log library did time formatting for us, now we do similar formatting ourselves.

jameshartig (approved):
LGTM

drazhanski:
> @drazhanski - will you be impacted if the log is less performant ?
I don't expect performance problems with this implementation. And usually we log only error/denial classes  of responses, so it can't be so critical.

miekg:
I'll make those changes and then merge.

plugin/pkg/log/log.go
miekg: done

Issue #2241, opened 2018-10-26
 Availability of coredns_cache metrics are affected by server block order


saj:
This is easier for me to explain by walking through a reproduction of unexpected behaviour.  I will walk through two configuration cases.  The first case results in expected behaviour.  The second case, which is only a slight change from the first, results in unexpected behaviour.  Apologies for the length.

---

**Case 1: First server block activates the `cache` plugin**

`Corefile`:

```
.:1053 {
    prometheus 0.0.0.0:9153
    errors

    cache {
        prefetch 2
    }

    forward . 8.8.8.8 {
        max_fails 0
    }
}

.:1054 {
    prometheus 0.0.0.0:9153
    errors

    forward . 8.8.8.8 {
        max_fails 0
    }
}
```

CoreDNS started with:

```
docker run --name coredns --rm \
  --mount type=bind,src=/private/tmp/Corefile,dst=/Corefile,readonly \
  -p 1053:1053/udp -p 1054:1054/udp -p 9153:9153/tcp \
  coredns/coredns:1.2.5
```

Queries to both server blocks works as expected:

```
% dig @127.0.0.1 -p 1053 google.com +noquestion +noadditional +noauthority

; <<>> DiG 9.10.6 <<>> @127.0.0.1 -p 1053 google.com +noquestion +noadditional +noauthority
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 10608
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 4, ADDITIONAL: 9

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; ANSWER SECTION:
google.com.             196     IN      A       216.58.203.110

;; Query time: 13 msec
;; SERVER: 127.0.0.1#1053(127.0.0.1)
;; WHEN: Sat Oct 27 06:35:47 AEDT 2018
;; MSG SIZE  rcvd: 505

% dig @127.0.0.1 -p 1053 google.com +noquestion +noadditional +noauthority

; <<>> DiG 9.10.6 <<>> @127.0.0.1 -p 1053 google.com +noquestion +noadditional +noauthority
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 35353
[...]

PR #2246, opened 2018-10-27
Metrics registered on wrong prometheus registry

fturib:

When metrics of 2 different blocs are available on the SAME tcp address, 
all the plugins of both blocs should register that collectors for Prometheus on the Registry that is associated with the listener.

#2241

NONE.
It is fixing a wrong behavior.

### DIFF
--- a/plugin/cache/handler.go
+++ b/plugin/cache/handler.go
@@ -89,6 +89,15 @@ func (c *Cache) exists(state request.Request) *item {
 	return nil
 }
 
+// CacheResetMetrics - ensure all metrics for cache re-start from 0
+func CacheResetMetrics() {
+	cacheSize.Reset()
+	cacheHits.Reset()
+	cacheMisses.Reset()
+	cachePrefetches.Reset()
+	cacheDrops.Reset()
+}
+
 var (
 	cacheSize = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: plugin.Namespace,
--- a/plugin/cache/handler.go
+++ b/plugin/cache/handler.go
@@ -89,15 +89,6 @@ func (c *Cache) exists(state request.Request) *item {
 	return nil
 }
 
-// CacheResetMetrics - ensure all metrics for cache re-start from 0
-func CacheResetMetrics() {
-	cacheSize.Reset()
-	cacheHits.Reset()
-	cacheMisses.Reset()
-	cachePrefetches.Reset()
-	cacheDrops.Reset()
-}
-
 var (
 	cacheSize = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: plugin.Namespace,
--- a/plugin/metrics/metrics.go
+++ b/plugin/metrics/metrics.go
@@ -57,7 +57,15 @@ func New(addr string) *Metrics {
 }
 
 // MustRegister wraps m.Reg.MustRegister.
-func (m *Metrics) MustRegister(c prometheus.Collector) { m.Reg.MustRegister(c) }
+func (m *Metrics) MustRegister(c prometheus.Collector) {
+	err := m.Reg.Register(c)
+	if err != nil {
+		// ignore any duplicate error, but fatal on any other kind of error
+		if _, ok := err.(prometheus.AlreadyRegisteredError); !ok {
+			log.Fatalf("cannot register metrics collector : %s", err)
+		}
+	}
+}
 
 // AddZone adds zone z to m.
 func (m *Metrics) AddZone(z string) {
--- a/plugin/metrics/metrics.go
+++ b/plugin/metrics/metrics.go
@@ -62,7 +62,7 @@ func (m *Metrics) MustRegister(c prometheus.Collector) {
 	if err != nil {
 		// ignore any duplicate error, but fatal on any other kind of error
 		if _, ok := err.(prometheus.AlreadyRegisteredError); !ok {
-			log.Fatalf("cannot register metrics collector : %s", err)
+			log.Fatalf("Cannot register metrics collector: %s", err)
 		}
 	}
 }
--- a/plugin/pkg/uniq/uniq.go
+++ b/plugin/pkg/uniq/uniq.go
@@ -10,6 +10,7 @@ type U struct {
 type item struct {
 	state int          // either todo or done
 	f     func() error // function to be executed.
+	obj   interface{}  // any object to return when needed
 }
 
 // New returns a new initialized U.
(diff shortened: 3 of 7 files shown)


fturib:
UT failing shows the bug described in #2241.

fturib:
hum ... still WIP. 
The code is in, but the UT TestMetricsSeveralBlocs fails at an unexpected point. and also does not fail when run alone.

plugin/cache/handler.go
```diff
@@ -89,6 +89,15 @@ func (c *Cache) exists(state request.Request) *item {
 	return nil
 }
 
+// CacheResetMetrics - ensure all metrics for cache re-start from 0
+func CacheResetMetrics() {
```
miekg: when is this needed?

plugin/metrics/metrics.go
```diff
@@ -57,7 +57,15 @@ func New(addr string) *Metrics {
 }
 
 // MustRegister wraps m.Reg.MustRegister.
-func (m *Metrics) MustRegister(c prometheus.Collector) { m.Reg.MustRegister(c) }
+func (m *Metrics) MustRegister(c prometheus.Collector) {
+	err := m.Reg.Register(c)
+	if err != nil {
+		// ignore any duplicate error, but fatal on any other kind of error
+		if _, ok := err.(prometheus.AlreadyRegisteredError); !ok {
+			log.Fatalf("cannot register metrics collector : %s", err)
```
miekg: Need captial C and there is a space before the :

miekg (commented):
looks ok

plugin/cache/handler.go
fturib: it is needed in the tests: https://github.com/coredns/coredns/blob/master/test/metrics_test.go

all the metrics collectors are declared as global vars.
All our UT or Tests are running on the same process. 
If you wish to test the value of these metrics inside one of the tests, you need to Reset the metrics before running that specific test, else you got whatever other test may have used.

I am introducing such a test with `TestMetricsSeveralBlocs`.
That was also the reason (IMO) of the existing test `TestMetricsCache` that was flaky.

That is also related to this issue #2206

plugin/metrics/metrics.go
fturib: ok.

plugin/cache/handler.go
miekg: ok, but then it does not have to be defined in the main code; just a test helper function.

plugin/pkg/uniq/uniq.go
```diff
@@ -10,37 +10,29 @@ type U struct {
...
+	if item, ok := u.u[key]; ok {
+		return item.obj
 	}
-	u.u[key] = item{todo, f}
+	u.u[key] = item{todo, f, o}
+	return o
 }
 
-// Unset removes the 'todo' associated with a key
+// Unset removes the key
```
miekg: nit: closing dot

test/metrics_test.go
```diff
@@ -182,3 +187,53 @@ func TestMetricsAuto(t *testing.T) {
 		t.Errorf("Expected value %s for %s, but got %s", "1", metricName, got)
 	}
 }
+
+// Show that when 2 blocs share the same metric listener (they have a prometheus plugin on the same listening address),
+// ALL the metrics of the second bloc in order are declared in prometheus, especially the plugins that are used ONLY in the second bloc
+func TestMetricsSeveralBlocs(t *testing.T) {
```
miekg, line 200: At some point this needs to be moved to proper test infra that prometheus now provides

miekg (changes requested):

test/metrics_test.go
fturib: Hum .. I am not sure what you mean.
There is a repo https://github.com/prometheus/test-infra, but that is a forked of Kubernetes/test-infra. And maybe the Team expect to adapt for Prometheus, but that is not yet done.

plugin/pkg/uniq/uniq.go
fturib: ok.

plugin/cache/handler.go
fturib: > ok, but then it does not have to be defined in the main code; just a test helper function.

There is an issue with the visibility of the metrics collectors - they are private to the cache package. 

The test function are the test CI functions (not UT that are in the same package).
Unless I miss something, I have either to keep this ResetMetrics() function here, wither to change the collectors to make then public.

Am I correct ?

miekg: sorry, i don't understand. If it needed in the tests it needs to put there, not here.

fturib: I need a public function that can reset the metrics collectors that are private to the `cache` package because the tests are NOT in the `cache` package.

miekg:
Where is it used?

fturib:
in https://github.com/coredns/coredns/blob/master/test/metrics_test.go

miekg:
So put it in the test code

fturib:
I see there is a misunderstanding ... but I do not see how to explain another way that issue: I cannot move that function in the `test` package.

- The code inside the function cannot compile or run outside of the `cache` package because it is using variables private to the package.
- The test that needs to call the function is in the package `test` (it is not a UT, it is a CI test)


```
package cache
...
// CacheResetMetrics - ensure all metrics for cache re-start from 0
func CacheResetMetrics() {
	cacheSize.Reset()
	cacheHits.Reset()
	cacheMisses.Reset()
	cachePrefetches.Reset()
	cacheDrops.Reset()
}
```

miekg:
I still don't understand. You want to do something in a test and you added
a public function in the non test code. That does not make sense?

miekg:
If this is non-ut we should fix it that we can do this as unit test

fturib:
Looks like we do not need fix.

miekg:
I don't know... The issue regarding this is not a bug?

fturib:
Yes it is a bug.

fturib:
But I do not expect to change the CI into UT.
I can remove the CI test if really that function cannot sit there.

rdrozhdzh:
@fturib, as an alternative to resetting counters, you can read the initial metrics values at the very beginning of your test. Then, further in your test just compare expected values with difference between current and initial values

fturib:
@rdrozhdzh : thanks for the hint! I did not thought using that way.

fturib:
@rdrozhdzh : thanks for your hint.

I changed that way. I mean removing CacheResetMetrics and test the diff of counting between start and end of the test case.
It works if you take care exactly to ensure the metrics are setup properly before gathering your "initial values".

It is a bit odd but work. for me the ResetMetrics() function is much more clear for everyone and the right operation to perform here.

I think if we do not want that ResetMetrics() we would better:
- either move the pseudo CI test to real CI test (in coredns/ci) that perform test on a new CoreDNS process
- or better, remove all global variables. Anyway those var are incompatible with reload ...

NOTE: I guess those var global are the reason of the metrics bleeding between listening servers (see #2206). So eventually all those global vars and Reset should be fixed by the corresponding PR.

test/metrics_test.go
miekg: iirc there is prometheus package where you can just read the metrics

miekg (approved):

PR #2247, opened 2018-10-28
cache: some optimizations

miekg:
Some optimization and lowercasing of the qname (in the end
miekg/dns should provide a fast and OK function for it).

* remove the make([]byte, 2) allocation in the key()
* use already lowercased qname in hash key calculation.

% benchcmp old.txt new.txt
benchmark                    old ns/op     new ns/op     delta
BenchmarkCacheResponse-4     9599          8735          -9.00%

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -2,7 +2,6 @@
 package cache
 
 import (
-	"encoding/binary"
 	"hash/fnv"
 	"net"
 	"time"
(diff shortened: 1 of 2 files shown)


fturib:
`BenchmarkCacheResponse-4 9599 8735 -9.00%` : is that mean we have a 9% gain in perf on the cache.

That worth the change !

fturib (approved):
LGTM

grobie (approved):

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] cache: some o..." ]
>`BenchmarkCacheResponse-4 9599 8735 -9.00%` : is that mean we have a 9% gain in perf on the cache.

Going back to strings (qname + qtype + do), was 15%, but adds more memory 
pressure. Also need way more changes, but might be a things we need to go back 
to.

Issue #2248, opened 2018-10-28
plugin/cache: README


miekg:
> A cache is divided into 256 shards, each holding up to 512 items by default - for a total size
of 256 * 512 = 131,072 items.

@fturib I can't remember (and not find it), where does the value of 512 come from?

fturib:
It used to be the minimum size of a shard.
But was changed in this PR: https://github.com/coredns/coredns/pull/2065

the minimum is now 4

```
 ssize := size / shardSize
 if ssize < 4 {
    ssize = 4
 }
```

chrisohaver:
Yeah - I missed that one.  Fixed in #2249.
Also, strictly speaking, our default cache is not 10000, due to shard size truncation, its 9984.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/cache:..." ]
>Yeah - I missed that one.  Fixed in #2249.
>Also, strictly speaking, our default cache is not 10000, due to shard size truncation, its 19*512 = 9728.

A thanks for the update. But the cache size is just diveded by the shardSize? We 
use 10,000 shardsize is 256, meaning per shard we only have 39 elements per map.

Where does the 512 come from in this calculation?

chrisohaver:
I corrected my comment several hours ago. But you can’t see corrections in email....

> Yeah - I missed that one. Fixed in #2249.
Also, strictly speaking, our default cache is not 10000, due to shard size truncation, its 9984.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/cache:..." ]
>I corrected my comment several hours ago. But you can’t see corrections in email....

I think you have like 10s to change your comment, and then an email is sent. 
I've never seen an update being send out.

/Miek

--
Miek Gieben

PR #2249, opened 2018-10-29
plugin/cache: Document actual cache defaults

chrisohaver:

Correct default cache capacity mentions in README:
* there was a stray mention of the old capacity
* Technically, the max capacity is actually 9984 (39*256), not exactly  10000, due to shard size truncation.

#2248

### DIFF
--- a/plugin/cache/README.md
+++ b/plugin/cache/README.md
@@ -25,7 +25,7 @@ cache [TTL] [ZONES...]
 
 Each element in the cache is cached according to its TTL (with **TTL** as the max).
 A cache is divided into 256 shards, each holding up to 39 items by default - for a total size
-of 256*39 = 9984 items.
+of 256 * 39 = 9984 items. 
 
 If you want more control:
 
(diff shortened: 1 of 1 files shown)


plugin/cache/README.md
```diff
@@ -24,8 +24,8 @@ cache [TTL] [ZONES...]
 * **ZONES** zones it should cache for. If empty, the zones from the configuration block are used.
 
 Each element in the cache is cached according to its TTL (with **TTL** as the max).
-A cache is divided into 256 shards, each holding up to 512 items by default - for a total size
-of 256 * 512 = 131,072 items.
+A cache is divided into 256 shards, each holding up to 19 items by default - for a total size
+of 256 * 19 = 9728 items.
```
fturib: it is 39 no ?
(256 * 19 = 4864)

fturib: maybe good to say it is the closest value to 10000 items .. (just because choosing 39 or 9728 seems really weird)

fturib (changes requested):
I think an error in numbers. Please re-check

chrisohaver:
> I think an error in numbers. Please re-check

Yep - realized the error, and committed a fix. I was using a wrong shard count of 512.

chrisohaver:
@fturib, I added a note in the docs about the rounding behavior.

fturib (approved):
Thanks!

grobie (approved):

johnbelamaric:
This is # of "items" - that means response messages? Does this mean a max
memory footprint for the default cache settings of 4096 * 9728? Although a
4k message would be rare.

miekg:
I wonder if a per RRset cache would help here, but it does complicate thing regarding lookups and I also wonder how much you win.
We don't max the response, so 60KB TCP returned packet would be cached.

Issue #2250, opened 2018-10-29
plugin/Log - accept Metadata in the logging string


fturib:
Allow log format string accept metadata information

Supposing that those metadata exist, provided by other plugin
- mac/address
- request/client_id
- ..
- <plugin>/<name>

Allow the Log format string to accept any of those metadata:
```
.:53 {
   log . "{name} {type} MAC:{mac/address} from {request/client_id}"
   ...
}
```
would be a valid format.

miekg:
implemented in #2251

PR #2251, opened 2018-10-30
plugin/log - Support for Metadata

fturib:
- add metadata support to Log and introduce laze init for replacer

**Support of Metadata for plugin Log**
Any place holder with a valid metadata label format : {<plugin>/<label>} we be accepted and replaced by the value of the corresponding metadata.
If metadata does not exist or provide an empty value, the default "emptyvalue" will be applied.

**Lazy initialization**
All Placeholder values are now given with a function that will be called ONLY of the placeholder is defined in the log. It allows lazy initialization.

**parsing the log format**
The process of replacement is now changed into a scan of {} and replacement placeholder by placeholder instead of the former try to replace for each possible placeholders.
(side-effect of lazy-initialization)

#2250 

NOTE: 3 place holders added recently within PR #2240 are removed:
- one was already a metadata and will be interpreted directly
- the 2 other are counting the number of A and AAAA records. Considered as very specific, and can now be added as metadata by who needs it. (@drazhanski : I proposed a PR for plugin AtepMonitor that can handle thoseMetadata specifics for your logging).

NOTE: this PR does not transform into Metadata the standard place holders of Log. That can be done later if we wish.

Updated the README of plugin/log

### DIFF
--- a/plugin/log/README.md
+++ b/plugin/log/README.md
@@ -61,6 +61,7 @@ The following place holders are supported:
 * `{class}`: qclass of the request
 * `{proto}`: protocol used (tcp or udp)
 * `{remote}`: client's IP address, for IPv6 addresses these are enclosed in brackets: `[::1]`
+* `{local}`: server's IP address, for IPv6 addresses these are enclosed in brackets: `[::1]`
 * `{size}`: request size in bytes
 * `{port}`: client's port
 * `{duration}`: response duration
@@ -72,6 +73,10 @@ The following place holders are supported:
 * `{>do}`: is the EDNS0 DO (DNSSEC OK) bit set in the query
 * `{>id}`: query ID
 * `{>opcode}`: query OPCODE
+* `{/[LABEL]}`: any metadata label is accepted as a place holder if it is enclosed between `{/` and  `}`.
+the place holder will be replaced by the corresponding metadata value or the default value `-` if label is not defined.
+
+
 
 The default Common Log Format is:
 
--- a/plugin/pkg/replacer/replacer.go
+++ b/plugin/pkg/replacer/replacer.go
@@ -1,10 +1,12 @@
 package replacer
 
 import (
+	"context"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/coredns/coredns/plugin/metadata"
 	"github.com/coredns/coredns/plugin/pkg/dnstest"
 	"github.com/coredns/coredns/request"
 
@@ -21,6 +23,7 @@ type Replacer interface {
 }
 
 type replacer struct {
+	ctx          context.Context
 	replacements map[string]string
 	emptyValue   string
 }
@@ -31,18 +34,20 @@ type replacer struct {
 // values into the replacer. rr may be nil if it is not
 // available. emptyValue should be the string that is used
 // in place of empty string (can still be empty string).
-func New(r *dns.Msg, rr *dnstest.Recorder, emptyValue string) Replacer {
+func New(ctx context.Context, r *dns.Msg, rr *dnstest.Recorder, emptyValue string) Replacer {
 	req := request.Request{W: rr, Req: r}
 	rep := replacer{
+		ctx: ctx,
 		replacements: map[string]string{
-			"{type}":   req.Type(),
-			"{name}":   req.Name(),
-			"{class}":  req.Class(),
-			"{proto}":  req.Proto(),
+			"{type}":  req.Type(),
+			"{name}":  req.Name(),
+			"{class}": req.Class(),
+			"{proto}": req.Proto(),
 			"{when}":   "", // made a noop
 			"{size}":   strconv.Itoa(req.Len()),
 			"{remote}": addrToRFC3986(req.IP()),
 			"{port}":   req.Port(),
+			"{local}":  addrToRFC3986(req.LocalIP()),
 		},
 		emptyValue: emptyValue,
 	}
@@ -71,22 +76,35 @@ func New(r *dns.Msg, rr *dnstest.Recorder, emptyValue string) Replacer {
 // Replace performs a replacement of values on s and returns
 // the string with the replaced values.
 func (r replacer) Replace(s string) string {
-	// Header replacements - these are case-insensitive, so we can't just use strings.Replace()
-	for strings.Contains(s, headerReplacer) {
-		idxStart := strings.Index(s, headerReplacer)
-		endOffset := idxStart + len(headerReplacer)
-		idxEnd := strings.Index(s[endOffset:], "}")
-		if idxEnd > -1 {
-			placeholder := strings.ToLower(s[idxStart : endOffset+idxEnd+1])
-			replacement := r.replacements[placeholder]
-			if replacement == "" {
-				replacement = r.emptyValue
+
+	// declare a function that replace based on header matching
+	fscanAndReplace := func(s string, header string, replace func(string) string) string {
+		b := strings.Builder{}
+		for strings.Contains(s, header) {
+			idxStart := strings.Index(s, header)
+			endOffset := idxStart + len(header)
+			idxEnd := strings.Index(s[endOffset:], "}")
+			if idxEnd > -1 {
+				placeholder := strings.ToLower(s[idxStart : endOffset+idxEnd+1])
+				replacement := replace(placeholder)
+				if replacement == "" {
+					replacement = r.emptyValue
+				}
+				b.WriteString(s[:idxStart])
+				b.WriteString(replacement)
+				s = s[endOffset+idxEnd+1:]
+			} else {
+				break
 			}
-			s = s[:idxStart] + replacement + s[endOffset+idxEnd+1:]
-		} else {
-			break
 		}
+		b.WriteString(s)
+		return b.String()
 	}
+
+	// Header replacements - these are case-insensitive, so we can't just use strings.Replace()
+	s = fscanAndReplace(s, headerReplacer, func(placeholder string) string {
+		return r.replacements[placeholder]
+	})
 
 	// Regular replacements - these are easier because they're case-sensitive
 	for placeholder, replacement := range r.replacements {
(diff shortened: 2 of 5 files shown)


plugin/log/README.md
```diff
@@ -72,10 +73,10 @@ The following place holders are supported:
 * `{>do}`: is the EDNS0 DO (DNSSEC OK) bit set in the query
 * `{>id}`: query ID
 * `{>opcode}`: query OPCODE
-* `{local}`: server's IP address, for IPv6 addresses these are enclosed in brackets: `[::1]`
-* `{A}`: number of the A type answers in the response
-* `{AAAA}`: number of the AAAA type answers in the response
-* `{forward/resolving_proxy}`: if the query was forwarded to a resolver, the IP address and port of that replied resolver, else empty. The metadata plugin needs to be enabled to use this placeholder
+* `{<plugin>/<name>}`: metadata labels are accepted as place holders and will be replaced by the corresponding metadata value.
```
miekg: for easy scaning these should start with a `/`

plugin/pkg/replacer/replacer.go
```diff
@@ -36,80 +41,92 @@ type replacer struct {
...
 }
 
 // Replace performs a replacement of values on s and returns
 // the string with the replaced values.
 func (r replacer) Replace(s string) string {
-	// Header replacements - these are case-insensitive, so we can't just use strings.Replace()
-	for strings.Contains(s, headerReplacer) {
-		idxStart := strings.Index(s, headerReplacer)
-		endOffset := idxStart + len(headerReplacer)
+	b := strings.Builder{}
```
miekg: this is a good fix, because looking through this replacement code I think it looks pretty heavy weight

plugin/pkg/replacer/replacer.go
```diff
@@ -36,80 +41,92 @@ type replacer struct {
...
+	for strings.Contains(s, "{") {
```
miekg: please add new section here for the label parsing

plugin/pkg/replacer/replacer.go
```diff
@@ -36,80 +41,92 @@ type replacer struct {
...
+				if fok {
+					replacement = f()
+				} else {
+					label := placeholder[1 : len(placeholder)-1]
+					fm := metadata.ValueFunc(r.ctx, label)
+					if fm != nil {
+						replacement = fm()
+					} else {
+						if !metadata.IsLabel(label) {
+							replacement = "<unknown-place-holder>"
```
miekg: this is not `emptyValue`

plugin/pkg/replacer/replacer.go
```diff
@@ -36,80 +41,92 @@ type replacer struct {
...
 			placeholder := strings.ToLower(s[idxStart : endOffset+idxEnd+1])
-			replacement := r.replacements[placeholder]
+			replacement, ok := r.evaluated[placeholder]
+			if !ok {
+				f, fok := r.replacements[placeholder]
+				if fok {
+					replacement = f()
+				} else {
+					label := placeholder[1 : len(placeholder)-1]
+					fm := metadata.ValueFunc(r.ctx, label)
```
miekg: so I thought we couldn't use this API because things are not picked up at the right time? Can you add some text as comments and/or in the commit text on how this will work?

plugin/pkg/replacer/replacer.go
```diff
@@ -36,80 +41,92 @@ type replacer struct {
...
-		endOffset := idxStart + len(headerReplacer)
+	b := strings.Builder{}
+	for strings.Contains(s, "{") {
+		idxStart := strings.Index(s, "{")
+		endOffset := idxStart + 1
 		idxEnd := strings.Index(s[endOffset:], "}")
 		if idxEnd > -1 {
 			placeholder := strings.ToLower(s[idxStart : endOffset+idxEnd+1])
-			replacement := r.replacements[placeholder]
+			replacement, ok := r.evaluated[placeholder]
```
miekg: why are we keeping track of this?

plugin/pkg/replacer/replacer.go
```diff
@@ -22,8 +22,13 @@ type Replacer interface {
 	Set(key, value string)
 }
 
+// Func is the type of function in the metadata, when called they return the value of the label.
+type Func func() string
+
 type replacer struct {
-	replacements map[string]string
+	ctx          context.Context
+	replacements map[string]Func   // replacements are given as function for lazy init
```
miekg: only the label replacements needs to be a function,  the rest can basically stay as-is?

miekg (commented):
Thanks, this looks better.

plugin/pkg/replacer/replacer.go
fturib: if the placeholder is not a standard and is has no format of a metadata label, then it is obviously an error in the Corefile. Here is a way to warn it.

plugin/pkg/replacer/replacer.go
fturib: `evaluated` is a cache for placeholder already evaluated.

plugin/pkg/replacer/replacer.go
fturib: I am not sure to understand here. do you mean Label as metadata label ?
in that case it is the same loop.

plugin/pkg/replacer/replacer.go
fturib: Hum.. I think you are referring to the part that gather data in Metadata. 
In PR #2240, this part is on the `Forward` plugin. 
You can either gather your data before any query processing (when Metadata call you with Provider), or later on when you data is known (in the use case of that PR, the data is known only after the ServeDNS action).

During the time your metadata is not gathered, its value is considered "".

plugin/pkg/replacer/replacer.go
fturib: It is a way to avoid multiple call to the function that provide the data. It depends how you use the Replacer.  I think we could re-use the same replacer for ALL the rules in Log.

plugin/pkg/replacer/replacer.go
fturib: Yes. I was addressing the comment of JohnB in PR #2240 for lazy initialization. It avoid doing the computation for values that are not used in the log format.

plugin/log/README.md
fturib: The name of the labels are defined by any plugin providing the metadata, not by the Log. 
I use here the generic label format we defined (see metadata.IsLabel())

That will not change the scanning that look for "{" and "}" anyway.

fturib: NOTE: maybe what you mean is that we could transform the standard placeholder of Log into metadata information that startwith "/".

eg : 
- {name} => could be a metadata /name and then used as {/name} in Log ?
- {type} => could be a metadata /type and then used as {/type} in Log ?

miekg: yes, each metadata thingy starts with `/` for easy detection and make it more visible

plugin/pkg/replacer/replacer.go
miekg: yes, but this only matters for labels, to we just needs a separate map just for that.

plugin/pkg/replacer/replacer.go
miekg: a completely new loop, this is getting unreadable

plugin/pkg/replacer/replacer.go
miekg: I don't think this is needed.

plugin/pkg/replacer/replacer.go
miekg: ok

plugin/pkg/replacer/replacer.go
miekg: well, you get EmptyValue then or something. If you want to warn a person you probably need to crash the process; no-one will see this or look for it in a running environment.

plugin/pkg/replacer/replacer.go
fturib: ok. will come backward a little bit to extend only for labels.

plugin/pkg/replacer/replacer.go
fturib: ok. will come backward a little. removing the cache for values

fturib:
I modified the commit:
- do not change anything about current standard placeholders
- process labels of metadata as a separated loop
- add corresponding UT
- remove extra std placeholders that can be treated as separated metadata (A, AAAA)

fturib:
I did not addressed the {/name} placeholders.
@miekg : want to be sure we are on the same page here.
Do you want I transform all existing standard placeholders and make them metadata with labels that starts directly with / ?
eg : {/name}, {/type}, {/>id}

In that case, everything will be metadata.

Is that correct ?

miekg:
uh, no, leave the current one. Just where you want to reference a thing that you know has a metadata label you use the new `{/` prefix.

fturib:
So what you mean is that Labels will be in the format : {/<plugin>/<name>}
for instance:
- `{/mac/address}`
- or `{/forward/resolving_addr}`

am I correct ?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/log - ..." ]
>So what you mean is that Labels will be in the format : {/<plugin>/<name>}
>for instance:
>- `{/mac/address}`
>- or `{/forward/resolving_addr}`
>
>am I correct ?

yes, that allows for quick identification and easy parsing.

/Miek

--
Miek Gieben

rdrozhdzh:
I would still prefer lazy initialization like `replacements map[string]Func` (though, I didn't see that version of code).
This way we could create the map only once (map creation/initialization is quite expensive). `New()` would just save internally the parameters like Context, Msg and Recorder and reuse the `replacements` map for each `Replacer` object. Each `Func` in map should have access to parameters saved in `Replacer` by `New()`

plugin/pkg/replacer/replacer.go
rdrozhdzh: ideally, single strings.Builder{} object should be used for building whole replacement string (in latest version of PR this is not true)

fturib:
I will open a separated PR for lazy initialization, as proposed by @rdrozhdzh which seems to be a better idea than what I proposed earlier.
That will improve the current PR at perf point of view, but keep the same behavior

fturib:
I create issue #2293 for the performance improvement.

Is that ok to validate this PR as is, or should I implement that perf improvement first ?

miekg (approved):

johnbelamaric (approved):

PR #2254, opened 2018-10-31
plugin/etcd: propagate recursion flag properly

callebjorkell:

When fetching records via the etcd plugin, the recursion flag was never set properly according to if the caller requested an exact record match or not. This causes problems especially in CNAME lookups, where recursion took place and a random RR was returned instead of the one that was specifically added for this key. Even when there is no service attached on the given path, it is still wrong to return a random one from the recursion.

To fix this issue, the `exact` flag is passed down to the etcd client as the `recursive` parameter to not perform recursive lookups when an exact match is requested by the caller.

A test is included to prove the right CNAME record is now returned (the test fails on the current master).

None

None


miekg:
think this is OK. Thanks!

PR #2255, opened 2018-10-31
plugin/loop: fix loop detection when an upstream DNS server is not reachable

matlec:

The loop plugin erroneously detects a loop (and causes the server startup to fail) when the server is configured with an upstream DNS server that is unavailable during startup. The issue can be reproduced with the following minimal configuration file (choose any IP address that is unavailable for you):
```
.:53 {
	proxy . 10.20.30.40
	loop
}
```
Running the server with this configuration file yields the following output:
```
[FATAL] plugin/loop: Forwarding loop detected in "." zone. Exiting. See https://coredns.io/plugins/loop#troubleshooting. Probe query: "HINFO 1132263262137561571.8670159584435333845.".
```

The loop detection fails for the following reason: when the server receives the HINFO query, the query first passes the loop plugin which registers an occurence of the query by calling inc(). When the query is passed to the proxy plugin, the plugin tries to forward the request to the (unavailable) upstream server which ultimately causes the HINFO query request to fail. In its current implementation, the loop plugin retries the query several times over a total time span of 30 seconds. With each retry the loop plugin registers another occurrence of the HINFO query request. When loop plugin receives the second retry (i.e. when seen() returns 3) , the plugin erroneously believes to have detected a loop.

None

None

### DIFF
--- a/plugin/loop/loop.go
+++ b/plugin/loop/loop.go
@@ -77,6 +77,12 @@ func (l *Loop) inc() {
 	l.i++
 }
 
+func (l *Loop) reset() {
+	l.Lock()
+	defer l.Unlock()
+	l.i = 0
+}
+
 func (l *Loop) setDisabled() {
 	l.Lock()
 	defer l.Unlock()
--- a/plugin/loop/setup.go
+++ b/plugin/loop/setup.go
@@ -42,6 +42,7 @@ func setup(c *caddy.Controller) error {
 
 			for time.Now().Before(deadline) {
 				if _, err := l.exchange(addr); err != nil {
+					l.reset()
 					time.Sleep(1 * time.Second)
 					continue
 				}


chrisohaver:
Ah ha! Thanks for identifying and fixing this.  And thanks for the clear explanation of how it was broken.

We should add a test for this ... possibly in coredns/ci.

chrisohaver (approved):

fturib:
@miekg , @chrisohaver : in your opinion, 
- How critical is this issue for CoreDNS in Kubernetes ? 
- Should we plan another release of CoreDNS to integrate into kubernetes v1.13 before its release ?

chrisohaver:
IMO - Critical enough that we should include this in the k8s 1.13 release (via a new release CoreDNS 1.2.6).

johnbelamaric:
Agreed.

miekg:
We can do a release easily.

Issue #2257, opened 2018-11-01
Cached responses incorrectly truncated


stuartnelson3:
coredns version: 1.2.5
kernel: 4.18.10-arch1-1-ARCH
dig version: 9.13.3

After making an SRV lookup request using TCP, and then making the same request forcing a UDP response and a limited buffer size against the now-cached lookup, the response received via UDP exceeds the requested buffer size.

The test was executed on my local machine running the release coredns binary 1.2.5, but was initially seen on our production instances running debian jessie + kernel 4.18.12+soundcloud1 (there are some minor patches to the kernel).

Corefile (mimics production):
```
. {
  prometheus :1853
  errors
  log

  loadbalance

  cache 3600 {
    prefetch 1 1m
  }

  proxy . <recursor-ip>:53 {
    max_fails 5
    spray
  }
}
```

Initial tcp dig:
https://gist.github.com/stuartnelson3/4bf9251e92032b7270aa2e9fdf70e059#file-tcp_lookup-txt
Cached, incorrectly truncated udp response:
https://gist.github.com/stuartnelson3/4bf9251e92032b7270aa2e9fdf70e059#file-udp_lookup-txt

stuartnelson3:
The issue was introduced in https://github.com/coredns/coredns/commit/074d176f039940d408bde2831c442500433f3ac4

stuartnelson3:
The error seems to lie here: https://github.com/coredns/coredns/blob/master/request/request.go#L313-L317

`rl` == 4160
`size` == 4084

Removing the single record results in a msg size of 4119, indicating that the answer that was removed was 41 bytes, but it appears it needed to remove 76 bytes in order to fit the buffer.

PR #2261, opened 2018-11-01
fix truncation bug

stuartnelson3:

There is a truncation bug. The two main changes are that, originally, the binary search algorithm was using the wrong value for `re`:
```
re := len(reply.Extra) - sub
size -= sub
```
In this code, `sub = optLen = 12`, which is in bytes. However, `len(reply.Extra)` is a number of records, not their byte amount. I figure that that was supposed to be "remove one record from reply.Extra", and not remove 12 from the length, which resulted in the extra section not being shortened since `l == 0` and `re` would be some negative number.

~~The second change was to pull the binary search code into its own function. After the change above, my test case was still failing, but once I moved to an isolated function (and variables) for the binary search, it fixed the tests. My guess is some variable was being shadowed or something unexpected otherwise mutated.~~

EDIT:
The second change is for the binary search, instead of terminating when the lower bound equals the upper bound, do one final iteration when they are equal (and then exit the loop). This will perform the last check that sets record length to the mid point that the binary search found.

https://github.com/coredns/coredns/issues/2257

None that I'm aware of.


request/testdata/additional.txt
```diff
@@ -0,0 +1,5 @@
+ip-10-10-52-54.y05.foo3.example.org. 5 IN A	10.10.52.54
```
miekg: Can these just be generated on the fly? Instead of adding these files?

miekg (commented):
Thanks, looking through this now. Wonder why the build/test failed.

miekg:
This fails in travis with:
~~~
=== RUN   TestRequestScrubExtra
--- FAIL: TestRequestScrubExtra (0.01s)
    request_test.go:105: Want scrub to not set truncated bit
=== RUN   TestRequestScrubExtraEdns0
--- FAIL: TestRequestScrubExtraEdns0 (0.01s)
    request_test.go:127: Want scrub to not set truncated bit
=== RUN   TestRequestScrubExtraRegression
--- FAIL: TestRequestScrubExtraRegression (0.00s)
    request_test.go:157: Want scrub to not set truncated bit
=== RUN   TestSoundCloudNoTruncate
~~~

miekg:
Would be good to get this in 1.2.6 though

stuartnelson3:
Fixed the issues that travis caught, I had introduced another regression :)

request/request.go
```diff
@@ -252,16 +252,16 @@ func (r *Request) Scrub(reply *dns.Msg) *dns.Msg {
 	// Account for the OPT record that gets added in SizeAndDo(), subtract that length.
 	sub := 0
 	if r.Req.IsEdns0() != nil {
-		sub = optLen
+		size -= optLen
+		sub = 1
```
miekg: are we still using sub?

request/request.go
```diff
@@ -252,16 +252,16 @@ func (r *Request) Scrub(reply *dns.Msg) *dns.Msg {
...
 	}
 
 	// subtract to make spaces for re-added EDNS0 OPT RR.
 	re := len(reply.Extra) - sub
-	size -= sub
 
 	l, m := 0, 0
 	origExtra := reply.Extra
-	for l < re {
+	for l <= re {
```
miekg, line 262: this warrants a comment to why `<=` needs to be used here

request/request.go
stuartnelson3: yeah, it's being used below on line 260

request/request.go
stuartnelson3: This is the standard way to implement a binary search. If we exit the loop when `l == re`, then we potentially aren't checking one of the indices in the slice. We might be, because of integer division, but there's no guarantee we are.

edit:

Skipping the last check (when we terminate if `l == re`) caused the off-by-one error that my tests reveal, where the loop exits before performing the final iteration that would bring it within either being one record below or one record above the desired bufsize.

miekg:
Suggesting this patch:
~~~
diff --git a/request/request.go b/request/request.go
index 72aec097..32587e7c 100644
--- a/request/request.go
+++ b/request/request.go
@@ -250,15 +250,15 @@ func (r *Request) Scrub(reply *dns.Msg) *dns.Msg {
        }
 
        // Account for the OPT record that gets added in SizeAndDo(), subtract that length.
-       sub := 0
+       re := len(reply.Extra)
        if r.Req.IsEdns0() != nil {
                size -= optLen
-               sub = 1
+               re--
+               if re < 0 {
+                       re = 0
+               }
        }
 
-       // subtract to make spaces for re-added EDNS0 OPT RR.
-       re := len(reply.Extra) - sub
-
        l, m := 0, 0
        origExtra := reply.Extra
        for l <= re {
~~~

miekg:
And actually `re` can never be 0 in this loop because we have an OPT RR, so the `if re< 0` check doesn't need to be done.

miekg (approved):

PR #2269, opened 2018-11-03
plugin/log: explain the {size} better

### LINKED ISSUE #2258: Logging reports incorrect message size
Discovered while reproducing https://github.com/coredns/coredns/issues/2257

```
# ... dig output ...
;; Query time: 1 msec
;; SERVER: ::1#1053(::1)
;; WHEN: Thu Nov 01 14:15:59 CET 2018
;; MSG SIZE  rcvd: 4119
```

```
[::1]:51981 - [01/Nov/2018:14:15:59 +0100] 263 "SRV IN http.web.stnn.rocket.srv.db.s-cloud.net. udp 81 false 4096" NOERROR qr,tc,rd,ra 6943 0.001206116s
```

It appears that dig reports the msg size as 4119, but the logging plugin says it is 6943.

I haven't checked, but this response bytes reported in the prometheus plugin could also be wrong.

miekg:
This is now the raw size which may be larger than what a particular
client actually sees.

Clarify this a bit.

Fixes #2258

### DIFF
--- a/plugin/log/README.md
+++ b/plugin/log/README.md
@@ -65,9 +65,9 @@ The following place holders are supported:
 * `{port}`: client's port
 * `{duration}`: response duration
 * `{rcode}`: response RCODE
-* `{rsize}`: response size
+* `{rsize}`: raw (uncompressed), response size (a client may receive a smaller response)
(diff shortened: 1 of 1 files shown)


yongtang (approved):
LGTM

PR #2270, opened 2018-11-03
plugin/host: parse file without holding lock

miekg:
Parse first and then swap the maps *while* holding the lock.


yongtang (approved):
LGTM. That should be a nice enhancement for large host file parsing I think.

yongtang (changes requested):
Looks like change in test might be needed.

yongtang (approved):
LGTM. Thanks!

PR #2279, opened 2018-11-07
add opts.initPodCache check to avoid panic

DavadDi:
When turn on autopath plugin, but kubenetes setting for pods != verified will cause panic and return SERVFAIL, expect worked as normal without autopath function.


miekg:
would be nice to have a test for this, but lgtm.

miekg:
though I'm not 100% sure. Any thoughts @chrisohaver ?

chrisohaver (approved):

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] add opts.init..." ]

/merge

Issue #2286, opened 2018-11-10
How  the `TransferTo` field of  kubernetes plugin used for?


DavadDi:
```
// Kubernetes implements a plugin that connects to a Kubernetes cluster.
type Kubernetes struct {
        // ...
	TransferTo         []string
}
```
```
func ParseStanza(c *caddy.Controller) (*Kubernetes, error) {
                //  .....
		case "transfer":
			tos, froms, err := parse.Transfer(c, false)
			if err != nil {
				return nil, err
			}
			if len(froms) != 0 {
				return nil, c.Errf("transfer from is not supported with this plugin")
			}
			k8s.TransferTo = tos
                        // ...
  ```

Only see  `TransferTo` is assigned on syntax parsed, but can't find any reference to using it? So what's the `TransferTo` used for? Any help?

chrisohaver:
On the surface this looks like unfinished code.  There should be a check that validates the operation is allowed (e.g. as is done with `TransferAllowed` in file plugin ).  Currently, it appears that kubernetes plugin ignores the `transfer to` option is ignored, allowing transfers anywhere.

chrisohaver:
fixed in #2292

PR #2292, opened 2018-11-12
plugin/kubernetes: Validate transfers are allowed

chrisohaver:

Actually check that the requesting IP is allowed to transfer the zone, as is described in the docs.

#2286 

none

### DIFF
--- a/plugin/kubernetes/xfr.go
+++ b/plugin/kubernetes/xfr.go
@@ -25,6 +25,10 @@ func (k *Kubernetes) MinTTL(state request.Request) uint32 { return 30 }
 // Transfer implements the Transferer interface.
 func (k *Kubernetes) Transfer(ctx context.Context, state request.Request) (int, error) {
 
+	if !k.transferAllowed(state) {
+		return dns.RcodeRefused, nil
+	}
+
 	// Get all services.
 	rrs := make(chan dns.RR)
 	go k.transfer(rrs, state.Zone)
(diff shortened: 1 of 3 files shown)


plugin/kubernetes/xfr.go
```diff
@@ -71,6 +75,25 @@ func (k *Kubernetes) Transfer(ctx context.Context, state request.Request) (int,
 	return dns.RcodeSuccess, nil
 }
 
+// transferAllowed checks if incoming request for transferring the zone is allowed according to the ACLs.
+func (k *Kubernetes) transferAllowed(state request.Request) bool {
```
miekg, line 80: Please add a comment that this is copy from zone.transferAllowed and should be (at somepoint) factored into a transfer pkg (or where ever we put it)

plugin/kubernetes/xfr_test.go
```diff
@@ -15,7 +15,7 @@ import (
 func TestKubernetesXFR(t *testing.T) {
 	k := New([]string{"cluster.local."})
 	k.APIConn = &APIConnServeTest{}
-	k.TransferTo = []string{"127.0.0.1"}
```
miekg, line 18: if you don't specify the port in the setup, it still works?

chrisohaver: I recall that it runs it through a normalizer, which adds the default 53 port...

I'll double check  ...

chrisohaver:
Added tests for transfer config parsing (there were none).  Verifies address with/without port.

plugin/kubernetes/xfr_test.go
bradbeam: https://github.com/coredns/coredns/blob/master/plugin/pkg/parse/host.go#L82 Yeah it should be fine to specify w/o a port.

bradbeam (approved):

Issue #2294, opened 2018-11-12
Negative caching not respecting minimum of TTL and MINIMUM


stuartnelson3:
The TTL for the SOA record resulting from no records (NXDOMAIN) set by backend_lookup.go is hardcoded to 300 seconds:

https://github.com/coredns/coredns/blob/master/plugin/backend_lookup.go#L394

According to https://tools.ietf.org/html/rfc2308#section-3:

>The TTL of this record is set from the minimum of the MINIMUM field of the SOA record and the TTL of the SOA itself, and indicates how long a resolver may cache the negative answer.

So it seems like line 394 should be returning the minimum of 300 and `b.minTTL(state)`?

It might also be nice to allow setting the max ttl for negative caching, like in bind9:

http://www.zytrax.com/books/dns/ch7/hkpng.html#max-ncache-ttl

stuartnelson3:
Having spoken with @grobie, it seems that it's fine for a SOA record to have a TTL that is higher than the MINIMUM, but then the caching server should only cache the response for the minimum of the two values. The issue we're encountering is that bind9 is always caching for the length of TTL, and not the minimum of the two.

So it seems like potentially this issue is requesting that we be able to configure the TTL and MINIMUM values, as they are both currently hardcoded at 300 and 30 seconds, respectively.

PR #2295, opened 2018-11-13
Set SOA record TTL as min of TTL/MINIMUM

stuartnelson3:

According to
https://tools.ietf.org/html/rfc2308#section-3:

  The TTL of this record is set from the minimum
  of the MINIMUM field of the SOA record and the
  TTL of the SOA itself, and indicates how long a
  resolver may cache the negative answer.

Set the SOA record TTL to the lesser of MINIMUM and TTL.

Servers that cache the SOA record might not properly respect the section quoted above. Changing the coredns response so that it always returns the smaller value as the record's TTL would prevent this error from happening.

These values are currently hardcoded in the kubernetes and etcd plugins, but they could be made configurable later if so desired.

#2294

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -389,17 +389,15 @@ func NS(b ServiceBackend, zone string, state request.Request, opt Options) (reco
 	return records, extra, nil
 }
 
-func min(a, b uint32) uint32 {
-	if a < b {
-		return a
-	}
-	return b
-}
-
 // SOA returns a SOA record from the backend.
 func SOA(b ServiceBackend, zone string, state request.Request, opt Options) ([]dns.RR, error) {
 	minTTL := b.MinTTL(state)
-	header := dns.RR_Header{Name: zone, Rrtype: dns.TypeSOA, Ttl: min(300, minTTL), Class: dns.ClassINET}
+	ttl := 300
+	if minTTL < ttl {
+		ttl = minTTL
+	}
+
+	header := dns.RR_Header{Name: zone, Rrtype: dns.TypeSOA, Ttl: ttl, Class: dns.ClassINET}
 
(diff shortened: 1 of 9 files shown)


plugin/backend_lookup.go
```diff
@@ -389,9 +389,17 @@ func NS(b ServiceBackend, zone string, state request.Request, opt Options) (reco
 	return records, extra, nil
 }
 
+func min(a, b uint32) uint32 {
+	if a < b {
+		return a
+	}
+	return b
+}
```
stuartnelson3: I didn't know where to put this function, so I just dropped it here. Assuming you're in favor of this PR, let me know if there's a better place for this.

grobie: Looks simple enough to me to just do that without a function.

```go
ttl, minTTL := 300, b.MinTTL(state)
if ttl > minTTL {
    ttl = minTTL
}
```

grobie (approved):

miekg:
Builds fails
~~~
 github.com/coredns/coredns/plugin
plugin/backend_lookup.go:396:12: invalid operation: minTTL < ttl (mismatched types uint32 and int)
plugin/backend_lookup.go:397:7: cannot use minTTL (type uint32) as type int in assignment
plugin/backend_lookup.go:400:59: cannot use ttl (type int) as type uint32 in field value
make: *** [travis] Error 1
~~~

miekg (approved):

stuartnelson3:
I can't seem to find the code to update for the failure, `--- FAIL: TestKubernetesA/10-20-0-101.test-1.pod.cluster.local._A (0.38s)`. Can you point it out?

stuartnelson3:
https://github.com/coredns/ci/pull/77

PR #2297, opened 2018-11-13
Create ISSUE_TEMPLATE.md

chrisohaver:

Requests that folks at least lets us know coredns version, corefile, and logs when reporting a bug or asking for assistance.


chrisohaver:
Should we add anything else here?

PR #2301, opened 2018-11-13
plugin/proxy: add jitter

miekg:
Add in a bunch of jitter when adding connection back into the map. This
help spread the load and keep some connection open for longer.

This does not implement keep a minimal number of connections open,
but jitter helps in making the expire less dependent on the query load.

### DIFF
--- a/plugin/file/secondary.go
+++ b/plugin/file/secondary.go
@@ -1,8 +1,9 @@
 package file
 
 import (
-	"math/rand"
 	"time"
+
+	"github.com/coredns/coredns/plugin/pkg/jitter"
 
 	"github.com/miekg/dns"
 )
@@ -134,7 +135,7 @@ Restart:
 				break
 			}
 
-			time.Sleep(jitter(2000)) // 2s randomize
+			time.Sleep(jitter.DurationMillisecond(2000))
 
 			ok, err := z.shouldTransfer()
 			if err != nil {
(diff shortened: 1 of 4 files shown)


rdrozhdzh:
lets assume config

user -> coredns1+forward1 ----tcp---> coredns2+forward2 -> external resolver

coredns2 will close idle TCP connection in 8 seconds, so it's no sense to keep connections in forward1 longer than 8s and it would be reasonable to set expire=8s . However if we add jitter than some connections could be used by forward1 after they are closed by coredns2. This will lead to extra round trip between coredns1 and coredns2.

Though, not sure how critical it is

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/proxy:..." ]
>lets assume config
>
>user -> coredns1+forward1 ----tcp---> coredns2+forward2 -> external resolver
>
>coredns2 will close idle TCP connection in 8 seconds, so it's no sense to keep connections in forward1 longer than 8s and it would be reasonable to set expire=8s . However if we add jitter than some connections could be used by forward1 after they are closed by coredns2. This will lead to extra round trip between coredns1 and coredns2.
>
>Though, not sure how critical it is

this will always be true, but with some jitter you decouple the incoming qps 
from the actual expire (I think)

miekg:
I think this will help in bursting traffic scenarios, but it needs something (maybe a test) that actually shows this is true.

miekg:
Closing this as a simpler fix has been merged and this add more complexity (seperate smaller jitter pkg may still be a good idea)

PR #2303, opened 2018-11-13
pkg/log: use 000 instead of 999 for the millis

miekg:
999 chops of suffix zero, 000 allows for the milliseconds to be always
printed with 3 chars. This makes the log the printed with the same
columns.

(partial logs below)

2018-11-13T21:13:28.249Z [INFO] [::1]
2018-11-13T21:13:48.414Z [INFO] [::1]
2018-11-13T21:13:49.1Z [INFO] [::1]

vs:

2018-11-13T21:20:22.262Z [INFO] [::1]
2018-11-13T21:20:22.436Z [INFO] [::1]
2018-11-13T21:20:22.608Z [INFO] [::1]

### DIFF
--- a/plugin/pkg/log/log.go
+++ b/plugin/pkg/log/log.go
@@ -20,7 +20,7 @@ import (
 var D bool
 
 // RFC3339Milli doesn't exist, invent it here.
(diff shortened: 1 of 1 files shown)


yongtang (approved):
LGTM

miekg:
Thanks!

/merge

PR #2296, opened 2018-11-13
Advertize email next to my github username

fturib:
### 1. Why is this pull request needed and what does it do?

Following Governance description, I add my email next to my github handle.

@miekg :
* I guess that will need change in the dreck project that parse this file.
* Do you want me to PR the change on that project (miek/dreck) ?

### 2. Which issues (if any) are related?

I would need for CNCF Graduation to show the different organizations the maintainers are from. Anyone with no email or personal email (like gmail) will be considered under its own organization.

That would also help when vote will come-in, as we restraint votes for people on the same organization.

### 3. Which documentation changes (if any) need to be made?

It is documented in [Governance](governance.md) already

PR #2304, opened 2018-11-13
OWNERS: add email

miekg:
Just use yaml support comments instead of hacking around it.


miekg:
Seems better and easier than #2296 

/cc @fturib

miekg:
ok, rebase hell. You get the gist.

PR #2308, opened 2018-11-14
[plugin/metadata] Add functionality for metadata to provide standard values.

rajansandeep:
Extend Metadata functionality to include the basic 'standard' request values from the Metadata plugin.
The standard values extended currently are the edns0 values from the `rewrite` plugin.

Document changes are made in the metadata Readme file.

### DIFF
--- a/plugin/metadata/metadata.go
+++ b/plugin/metadata/metadata.go
@@ -15,6 +15,7 @@ type Metadata struct {
 	Zones     []string
 	Providers []Provider
 	Next      plugin.Handler
+	options   map[uint16][]*edns0Map
 }
 
 // Name implements the Handler interface.
--- a/plugin/metadata/request.go
+++ b/plugin/metadata/request.go
@@ -1,0 +1,211 @@
+package metadata
+
+import (
+	"context"
+	"encoding/hex"
+	"fmt"
+	"net"
+	"strconv"
+
+	"github.com/coredns/coredns/request"
+	"github.com/miekg/dns"
+)
+
+const (
+	typeEDNS0Bytes = iota
+	typeEDNS0Hex
+	typeEDNS0IP
+)
+
+const (
+	QueryName  = "qname"
+	QueryType  = "qtype"
+	ClientIP   = "client_ip"
+	ClientPort = "client_port"
+	Protocol   = "protocol"
+	ServerIP   = "server_ip"
+	ServerPort = "server_port"
+	ResponseIP = "response_ip"
+)
+
+func newMetadata() *Metadata {
+	pol := &Metadata{options: make(map[uint16][]*edns0Map, 0)}
+	return pol
+}
+
+type edns0Map struct {
+	name     string
+	code     uint16
+	dataType uint16
+	size     uint
+	start    uint
+	end      uint
+}
+
+var stringToEDNS0MapType = map[string]uint16{
+	"bytes":   typeEDNS0Bytes,
+	"hex":     typeEDNS0Hex,
+	"address": typeEDNS0IP,
+}
+
+func newEDNS0Map(code, name, dataType, sizeStr, startStr, endStr string) (*edns0Map, error) {
+	c, err := strconv.ParseUint(code, 0, 16)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse EDNS0 code: %s", err)
+	}
+	size, err := strconv.ParseUint(sizeStr, 10, 32)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse EDNS0 data size: %s", err)
+	}
+	start, err := strconv.ParseUint(startStr, 10, 32)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse EDNS0 start index: %s", err)
+	}
+	end, err := strconv.ParseUint(endStr, 10, 32)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse EDNS0 end index: %s", err)
+	}
+	if end <= start && end != 0 {
+		return nil, fmt.Errorf("end index should be > start index (actual %d <= %d)", end, start)
+	}
+	if end > size && size != 0 {
+		return nil, fmt.Errorf("end index should be <= size (actual %d > %d)", end, size)
+	}
+	ednsType, ok := stringToEDNS0MapType[dataType]
+	if !ok {
+		return nil, fmt.Errorf("invalid dataType for EDNS0 map: %s", dataType)
+	}
+	ecode := uint16(c)
+	return &edns0Map{name, ecode, ednsType, uint(size), uint(start), uint(end)}, nil
+}
+
+func (m *Metadata) addEDNS0Map(code, name, dataType, sizeStr, startStr, endStr string) error {
+	meta, err := newEDNS0Map(code, name, dataType, sizeStr, startStr, endStr)
+	if err != nil {
+		return err
+	}
+	m.options[meta.code] = append(m.options[meta.code], meta)
+	return nil
+}
+
+func (m *Metadata) Metadata(ctx context.Context, state request.Request) context.Context {
+	return m.fillMetadata(ctx, state)
+}
+
+func (m *Metadata) fillMetadata(ctx context.Context, state request.Request) context.Context {
+
+	m.declareMetadata(QueryName, state.QName(), ctx)
+	m.declareMetadata(QueryType, dns.Type(state.QType()).String(), ctx)
+	m.declareMetadata(ClientIP, state.IP(), ctx)
+	m.declareMetadata(ServerIP, state.LocalIP(), ctx)
+	m.declareMetadata(ClientPort, state.Port(), ctx)
+	m.declareMetadata(ServerPort, state.LocalPort(), ctx)
+	m.declareMetadata(Protocol, state.Proto(), ctx)
+
+	m.getAttrsFromEDNS0(state.Req, ctx)
+
+	SetValueFunc(ctx, "metadata/"+ResponseIP, func() string {
+		ip := getRespIP(state.Req)
+		if ip != nil {
+			return ip.String()
+		}
+		return ""
+	})
+	return ctx
+
+}
+
+func (m *Metadata) declareMetadata(name string, value string, ctx context.Context) {
+	SetValueFunc(ctx, "metadata/"+name, func() string { return value })
+}
+
+func (m *Metadata) getAttrsFromEDNS0(r *dns.Msg, ctx context.Context) {
+	o := r.IsEdns0()
+	if o == nil {
+		return
+	}
+
+	for _, opt := range o.Option {
+		optLocal, local := opt.(*dns.EDNS0_LOCAL)
+		if !local {
+			continue
+		}
+		opts, ok := m.options[optLocal.Code]
+		if !ok {
+			continue
+		}
+		m.parseOptionGroup(optLocal.Data, opts, ctx)
+	}
+}
+
+func (m *Metadata) parseOptionGroup(data []byte, options []*edns0Map, ctx context.Context) {
+	for _, option := range options {
+		var value string
+		switch option.dataType {
+		case typeEDNS0Bytes:
+			value = string(data)
+		case typeEDNS0Hex:
+			value = parseHex(data, option)
+		case typeEDNS0IP:
+			ip := net.IP(data)
+			value = ip.String()
+		}
+		if value != "" {
+			m.declareMetadata(option.name, value, ctx)
+		}
+	}
+}
+
+func parseHex(data []byte, option *edns0Map) string {
+	size := uint(len(data))
+	// if option.size == 0 - don't check size
+	if option.size > 0 {
+		if size != option.size {
+			// skip parsing option with wrong size
+			return ""
+		}
+	}
+	start := uint(0)
+	if option.start < size {
+		// set start index
+		start = option.start
+	} else {
+		// skip parsing option if start >= data size
+		return ""
+	}
+	end := size
+	// if option.end == 0 - return data[start:]
+	if option.end > 0 {
+		if option.end <= size {
+			// set end index
+			end = option.end
+		} else {
+			// skip parsing option if end > data size
+			return ""
+		}
+	}
+	return hex.EncodeToString(data[start:end])
+}
+
+func getRespIP(r *dns.Msg) net.IP {
+	if r == nil {
+		return nil
+	}
+
+	var ip net.IP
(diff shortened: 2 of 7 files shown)


plugin/metadata/request.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/miekg/dns"
+)
+
+const (
+	typeEDNS0Bytes = iota
+	typeEDNS0Hex
+	typeEDNS0IP
+)
+
+const (
```
miekg, line 20: These const are all different than what we use in the `log` plugin, they should match for some consistency.

miekg: All these const are only used ones? doesn't really add anything to define them as const imo

plugin/metadata/metadata.go
```diff
@@ -15,6 +15,7 @@ type Metadata struct {
 	Zones     []string
 	Providers []Provider
 	Next      plugin.Handler
+	options   map[uint16][]*edns0Map
```
miekg, line 18: uhm, why is this not carried in the ctx like the rest?

plugin/metadata/request.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func (m *Metadata) getAttrsFromEDNS0(ctx context.Context, r *dns.Msg) {
+	o := r.IsEdns0()
+	if o == nil {
+		return
+	}
+
+	for _, opt := range o.Option {
+		optLocal, local := opt.(*dns.EDNS0_LOCAL)
```
miekg, line 130: odd to only care about EDNS0_LOCAL?

plugin/metadata/request.go
```diff
@@ -0,0 +1,212 @@
...
+	end      uint
+}
+
+var stringToEDNS0MapType = map[string]uint16{
+	"bytes":   typeEDNS0Bytes,
+	"hex":     typeEDNS0Hex,
+	"address": typeEDNS0IP,
+}
+
+func newEDNS0Map(code, name, dataType, sizeStr, startStr, endStr string) (*edns0Map, error) {
```
miekg, line 51: Don't understand the new for this parsing

plugin/metadata/request.go
```diff
@@ -0,0 +1,212 @@
...
+		if ip != nil {
+			return ip.String()
+		}
+		return ""
+	})
+	return ctx
+
+}
+
+func (m *Metadata) declareMetadata(ctx context.Context, name string, value string) {
```
miekg, line 119: really shallow function

plugin/metadata/request.go
```diff
@@ -0,0 +1,212 @@
...
+	var ip net.IP
+	for _, rr := range r.Answer {
+		switch rr := rr.(type) {
+		case *dns.A:
+			ip = rr.A
+
+		case *dns.AAAA:
+			ip = rr.AAAA
+		}
+		// If there are several responses, currently
```
miekg, line 205: what use is this function then? 
You also calling this function with state.Req.

plugin/metadata/metadata.go
miekg: options is a map to a slice of maps? You want to add such a heavy weight construct to every incoming request?

miekg (changes requested):

miekg:
There is also **zero** added benefit for adding `state` values into metadata. It just duplicates some values available to all plugins by default.

plugin/metadata/metadata.go
fturib: These are the parameters for extraction (location, name, format) that we get from the Stanza.
It is not the data itself - I mean that structure is only once for the life of the server.

plugin/metadata/request.go
fturib: These define the format of data in EDNS0.
Right now should be the same as REWRITE that do (for now) the same operation.

Once in Metadata, it would be expected that every plugins rather use the Metadata instead or redefining a way to extract the data from EDNS0 records.

plugin/metadata/request.go
fturib: Yes, maybe we can all it parseEDNS0Map(..)

plugin/metadata/request.go
fturib: Hum.. the issue here is the caching of the value between extraction and usage.
I tried several cases and find only this one, when the metadata value is a constant.

But maybe a good golang programmer would find another way.

plugin/metadata/request.go
```diff
@@ -0,0 +1,212 @@
...
+		case typeEDNS0Bytes:
+			value = string(data)
+		case typeEDNS0Hex:
+			value = parseHex(data, option)
+		case typeEDNS0IP:
+			ip := net.IP(data)
+			value = ip.String()
+		}
+		if value != "" {
+			m.declareMetadata(ctx, option.name, value)
```
fturib, line 155: Here, for EDNS0 that is much more heavy to handle, we should provide a real lazy initialization.
I mean : do not extract data until it is required by a usage of that data.

And by, the way we should to the same for any field of the Query : 
replace the "extract and cache the value" by "provide the function that extract on demand with a cache"

plugin/metadata/request.go
fturib: ok.

fturib (commented):
> There is also zero added benefit for adding state values into metadata. It just duplicates some values available to all plugins by default.

You are right. Each plugin can call these functions when needed.
However for the plugins that manipulate Metadata it is much more easy to have those values accessible through Metadata - it is about simplicity of usage.

And, anyway, the EDNS0 extraction part - which is very important for Firewall - needs to be defined somewhere.


So, I was thinking that Metadata was the right plugin to handle the very standard data (extraction form the Msg itself).

if we want to limit the impact on CPU, we should just ensure lazy init + cache of the value.

Exemple: (I agree we can do another way with the replacer that extract the same data)
```
   log . "from: {/metadata/client_ip}, looking for {/metadata/qname}"
```

``` 
   firewall query {
        refuse [metadata/qname] =~ 'infoblox.com' && [metadata/qtype == 'AAAA']
        allow true
   }
```

miekg:
A matter of convenience? That makes no sense  technically. Every plugin
uses state. It makes sense to use metadata to have plugins add things they
don't have access to. canonical example being mac address.

If you have something difficult to parse you add a function in dnsutil or
ednsutil. No need to lump that data with you (again) because you already
have access to it

fturib:
Sure, we can provide an internal lib that would provide the data the same way as in Metadata.
=> it would look like the replacer lib, but with lazy init in order to avoid call funtion when useless.

In that case the best would be to follow [proposition of @rdrozhdzh in this comment](https://github.com/coredns/coredns/pull/2251#issuecomment-434690523), and build a uniq map that can be instanciated against a "state". With lazy init. The global presentation will be almost the same as in Metadata.


There is still one concern about the EDNS0 data : we need a description for extracting the data out of the OPT record. And it make sense to have this description on only one place, and not on each plugin that need the data.

=> Can we keep that description on `metadata` plugin, or would it be preferable to create a `request` plugin that user includes in the stanza if he needs that extraction ?
NOTE: here, the same way, we can either provide this data in Metadata (lazy init still), or in a Map available on call through lib like `replacer`.

@miekg : let me know for the EDNS0 description and plugin associated. Thanks!

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [plugin/metad..." ]
>Sure, we can provide an internal lib that would provide the data the same way as in Metadata.
>=> it would look like the replacer lib, but with lazy init in order to avoid call funtion when useless.

No, you have some functions that get data from the reply.

>There is still one concern about the EDNS0 data : we need a description for extracting the data out of the OPT record. And it make sense to have this description on only one place, and not on each plugin that need the data.

parsing a EDNS_LOCAL option will be strange no matter how you spin it, it's a 
one off. As the comment says "edns.go:

    // The EDNS0_LOCAL option is used for local/experimental purposes.

>=> Can we keep that description on `metadata` plugin, or would it be preferable to create a `request` plugin that user includes in the stanza if he needs that extraction ?

metadata allows to insert metadata not available to a future plugin handling the 
request: this excludes anything availiable in the request or reply (which you can 
see and parse for yourself)

Sorry, but there is zero need for code that makes this more complex or adds a 
new naming scheme for "convience"

johnbelamaric:
Sounds like there are two separate issues. One is putting standard request stuff in metadata even if a plugin has access to it. I think the "convenience" here refers to whether plugin developers need to use a special case to differentiate local vs. metadata. So for example, in the log configuration, you have to have a switch statement in your "replacer" to decide where to pull the data from, rather than just always pulling it in the same way (I haven't looked closely at the code so correct me if my understanding is off here).

The second piece is the EDNS0 bit. So, this is used to coordinate requests between coredns instances. For example, in the Infoblox product, the on-prem (or in the workstation) CoreDNS instance will tag the outgoing DNS message with EDNS0 entries identifying the customer and user. This is done via rewrite. Then, on the receiving side, that data is parsed out and used to make decisions about how to reply. Various local EDNS0 options are used to encode that data and make it available on the upstream side.

This would be the same mechanism we can use for standard EDNS0 options. For example, if rewrite has access to the client IP it can use that to create an EDNS0 client subnet option; then on the upstream it can use that to respond appropriately. We can extend CoreDNS in that case with a way to support ECS, by having a general way to support response decisions based upon meta-data.

That general mechanism for response decisions based on meta-data is way more powerful than BIND's views, which IIUC are based on IP ACLs. We could base it on any metadata at all (EDNS0, kubernetes stuff, whatever).

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [plugin/metad..." ]
>Sounds like there are two separate issues. One is putting standard request stuff in metadata even if a plugin has access to it. I think the "convenience" here refers to whether plugin developers need to use a special case to differentiate local vs. metadata. So for example, in the log configuration, you have to have a switch statement in your "replacer" to decide where to pull the data from, rather than just always pulling it in the same way (I haven't looked closely at the code so correct me if my understanding is off here).
>
>The second piece is the EDNS0 bit. So, this is used to coordinate requests between coredns instances. For example, in the Infoblox product, the on-prem (or in the workstation) CoreDNS instance will tag the outgoing DNS message with EDNS0 entries identifying the customer and user. This is done via rewrite. Then, on the receiving side, that data is parsed out and used to make decisions about how to reply. Various local EDNS0 options are used to encode that data and make it available on the upstream side.

this can be done a small plugin that implements a Provider; and for this it most 
def. makes sense: a local edns0 option that has no use outside a specific 
context.

fturib:
I try to summarize here what is we be good to have. (please correct me if I am wrong).

I understand a plugin that describe how to extract that information from OPT records - the EDNS0 options.

Something like:

```
request {
   group_id edns0 0xffed byte
   client_id edns0 0xffee byte
}
```

And the plugin would implement the 'Provider' interface to gather that information in metadata. 

QUESTION:
- `request` plugin would be internal or external ?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [plugin/metad..." ]
>I try to summarize here what is we be good to have. (please correct me if I am wrong).
>
>I understand a plugin that describe how to extract that information from OPT records - the EDNS0 options.
>
>Something like:
>
>```
>request {
>   group_id edns0 0xffed byte
>   client_id edns0 0xffee byte
>}
>```
>
>And the plugin would implement the 'Provider' interface to gather that information in metadata.
>
>QUESTION:
>- `request` plugin would be internal or external ?

If you have locally defined EDNS0 options you're looking at a (small) private 
plugin to provide that metadata. We can't carry it in CoreDNS (well maybe we 
can), because it doesn't make sense to the rest of the world.

A *generic* plugin that just extracts EDNS0 options and presents them sanely 
(multiple metadata labels?) could make sense here.

fturib:
> A *generic* plugin that just extracts EDNS0 options and presents them sanely
> (multiple metadata labels?) could make sense here.

So,  the following plugin which would have a stanza for generic definition on how to extract the edn0 and give a label to that data falls in this "generic plugin" category. Correct ? 

```
extractEDNS0Plugin {
   <labelA> <edns0-addrA> <format>
   <labelB> <edns0-addrB> <format>
   etc ...
}
```

NOTE: <format> should provide the way to transform the binary data of the OPT into a string (all metadata are in string type). It can be just ASCII encoded, or a direct binary number, etc ...

So:
- What name for this plugin ?
- Can we plan to have it in coredns/coredns repo ?

johnbelamaric:
@miekg, the generic thing is the idea. That's the way the infoblox plugin
works today. it's also the mirror image of rewrite. So, rewrite let's you
pack local a local EDNS0 option with data pulled from the config and/or
from metadata. The idea of this is how to unpack it on the receiving side.
No specific local option codes are embedded in CoreDNS or even the plugin.
They are part of the Corefile. The original idea was to include this
functionality in the metadata plugin itself, but it could be a separate
plugin. @ftur I would think a name like metadata_edns0 would be good -
though I am not sure we allow _ in plugin names. We *could* have separate
plugins for different types of EDNS0 options: `md_edns0_local`,
`md_edns0_clientsubnet` etc. It proliferates plugins, but each of these
could have quite different requirements.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [plugin/metad..." ]
>@miekg, the generic thing is the idea. That's the way the infoblox plugin
>works today. it's also the mirror image of rewrite. So, rewrite let's you
>pack local a local EDNS0 option with data pulled from the config and/or
>from metadata. The idea of this is how to unpack it on the receiving side.
>No specific local option codes are embedded in CoreDNS or even the plugin.
>They are part of the Corefile. The original idea was to include this
>functionality in the metadata plugin itself, but it could be a separate
>plugin. @ftur I would think a name like metadata_edns0 would be good -
>though I am not sure we allow _ in plugin names. We *could* have separate
>plugins for different types of EDNS0 options: `md_edns0_local`,
>`md_edns0_clientsubnet` etc. It proliferates plugins, but each of these
>could have quite different requirements.

ok.

But this is mostly binary data; so I'm all for an **efficient** *new* plugin that 
implements a metadata.Provider and populates metadata labels for EDNS0 option. 
Exact label names TBD.

rajansandeep:
> But this is mostly binary data; so I'm all for an **efficient** *new* plugin that
implements a metadata.Provider and populates metadata labels for EDNS0 option.
Exact label names TBD.

@miekg @johnbelamaric @fturib 
I understand that we can have a new (internal) plugin to extract the edns0 options instead of the way it is currently proposed in this PR.

johnbelamaric:
yes

PR #2310, opened 2018-11-15
log: fix plugin logging

miekg:
Remove the code duplication and call of the "official" functions. This
is the second(?) time we forgot to update the other half, so remove that
problem entirely.

Also add a test if the correct (within limits) time in front of the log
line.

### DIFF
--- a/plugin/pkg/log/plugin.go
+++ b/plugin/pkg/log/plugin.go
@@ -19,7 +19,7 @@ func (p P) logf(level, format string, v ...interface{}) {
 }
 
 func (p P) log(level string, v ...interface{}) {
-	log(level+pFormat(p.plugin), v...)
+	log(level, pFormat(p.plugin), v...)
 }
 
 // Debug logs as log.Debug.
--- a/plugin/pkg/log/plugin.go
+++ b/plugin/pkg/log/plugin.go
@@ -15,7 +15,7 @@ type P struct {
 func NewWithPlugin(name string) P { return P{name} }
 
 func (p P) logf(level, format string, v ...interface{}) {
-	logf(level, pFormat(p.plugin)+fmt.Sprintf(format, v...))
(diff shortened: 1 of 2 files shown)


plugin/pkg/log/plugin.go
```diff
@@ -16,13 +15,11 @@ type P struct {
 func NewWithPlugin(name string) P { return P{name} }
 
 func (p P) logf(level, format string, v ...interface{}) {
-	s := level + pFormat(p.plugin) + fmt.Sprintf(format, v...)
-	golog.Print(s)
+	logf(level, pFormat(p.plugin)+fmt.Sprintf(format, v...))
```
rdrozhdzh:
```suggestion
	log(level, pFormat(p.plugin), fmt.Sprintf(format, v...))
```

plugin/pkg/log/plugin.go
```diff
@@ -16,13 +15,11 @@ type P struct {
...
 }
 
 func (p P) log(level string, v ...interface{}) {
-	s := level + pFormat(p.plugin) + fmt.Sprint(v...)
-	golog.Print(s)
+	log(level+pFormat(p.plugin), v...)
```
rdrozhdzh:
```suggestion
	log(level, pFormat(p.plugin), v...)
```

rdrozhdzh (approved):
Looks good. 
Some optimizations could be done (in this or another PR) - see below

formatting like in `pFormat()` should be done once in `NewWithPlugin()`.
`log()` and `logf()` should use the  stored formatted value

miekg:
that's a good point. Another PR seems in order.

miekg:
accepting your suggesting make it not compile and I can't revert anything.

rdrozhdzh:
Oh, sorry, I should have checked this first :(

PR #2315, opened 2018-11-15
Redo the plugin log PR

miekg:
Remove the code duplication and call of the "official" functions. This
is the second(?) time we forgot to update the other half, so remove that
problem entirely.

Also add a test if the correct (within limits) time in front of the log
line.


miekg:
this has been approved, but 'github suggestions' (or me) fucked it up. Retrying.

PR #2319, opened 2018-11-17
plugin/forward: remove dynamic read timeout

### LINKED ISSUE #2306: plugin/forward: dynamic read timeout - is it useful?
We have introduced this feature some time ago - see #1659, #1680, #1889.

The main reason for this was faster retrying in case of lost UDP query/response (i/o timeout) and having more attempts to retry in 5s interval, see also closed #1624

However, now I think this feature doesn't work fine in many cases and may be a reason for troubles like #2265. 

This is because we cannot reliably predict a reasonable timeout for next query based on RTT of previous queries. The time needed to resolve domain name varies significantly between different queries. It may depend on how many labels the queried domain name has, or which records have already been cached (or evicted from cache) on resolver side, or how responsive authoritative domain name servers for particular zone, the protocol used (UPD or TCP), etc

So, I propose to remove dynamic read timeout logic. I would also prefer to have the query timeout value configurable. I know @miekg you don't like knobs like this, but IMO this option would be really helpful for many people, since the timeout is highly dependent on environment in which coredns works, and we cannot predefine an ideal constant which would work perfectly for everyone.

As for similar feature for dynamic dial timeout (#1852) we may want to keep it, since dialTimeout doesn't depend on queried domain and dialing time may give more consistent values.

Any thoughts?

miekg:
We care about an upstream being there, so we still have a dynamic dial
time out (by way higher then 200ms) of 1s; this should be fairly stable
for an upstream. The read timeout if more variable because of cached and
non cached responses. As such remove his logic entirely.

Fixes #2306

### DIFF
--- a/plugin/forward/persistent.go
+++ b/plugin/forward/persistent.go
@@ -31,7 +31,7 @@ type Transport struct {
 
 func newTransport(addr string) *Transport {
 	t := &Transport{
-		avgDialTime: int64(defaultDialTimeout / 2),
+		avgDialTime: int64(maxDialTimeout / 2),
 		conns:       make(map[string][]*persistConn),
 		expire:      defaultExpire,
 		addr:        addr,
@@ -159,8 +159,8 @@ func (t *Transport) SetExpire(expire time.Duration) { t.expire = expire }
 func (t *Transport) SetTLSConfig(cfg *tls.Config) { t.tlsConfig = cfg }
 
 const (
-	defaultExpire      = 10 * time.Second
-	minDialTimeout     = 100 * time.Millisecond
-	maxDialTimeout     = 30 * time.Second
-	defaultDialTimeout = 30 * time.Second
+	defaultExpire  = 10 * time.Second
+	minDialTimeout = 1 * time.Second
+	maxDialTimeout = 30 * time.Second
+	readTimeout    = 5 * time.Second // this is the normal value for some resolve take quite a while, usually (cached) responses are fast.
 )
--- a/plugin/forward/persistent.go
+++ b/plugin/forward/persistent.go
@@ -162,5 +162,7 @@ const (
 	defaultExpire  = 10 * time.Second
 	minDialTimeout = 1 * time.Second
 	maxDialTimeout = 30 * time.Second
-	readTimeout    = 5 * time.Second // this is the normal value for some resolve take quite a while, usually (cached) responses are fast.
+
+	// Some resolves might take quite a while, usually (cached) responses are fast. Set to 2s to give us some time to retry a different upstream.
+	readTimeout = 2 * time.Second
 )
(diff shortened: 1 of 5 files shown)


miekg:
hmm these ci failures are strange because this code is not exercised. Nothing big changed in ci recently either... cc @chrisohaver

plugin/forward/persistent.go
```diff
@@ -159,8 +159,8 @@ func (t *Transport) SetExpire(expire time.Duration) { t.expire = expire }
...
 
 const (
-	defaultExpire      = 10 * time.Second
-	minDialTimeout     = 100 * time.Millisecond
-	maxDialTimeout     = 30 * time.Second
-	defaultDialTimeout = 30 * time.Second
+	defaultExpire  = 10 * time.Second
+	minDialTimeout = 1 * time.Second
+	maxDialTimeout = 30 * time.Second
+	readTimeout    = 5 * time.Second // this is the normal value for some resolve take quite a while, usually (cached) responses are fast.
```
rdrozhdzh: originally (before dynamic timeout) we used 2s as timeout (https://github.com/coredns/coredns/blob/master/plugin/forward/proxy.go#L84). 5s is too much because we won't be able to retry the query via different upstreams. 
As I mentioned in #2306, I would prefer to have this timeout configurable. If not, I would prefer value 1.5s to be able to make at least 3 attempts in 5s.
Yes, resolution of some domain names may take longer time on first query, but if client will send the same request again to the same resolver - this may get resolved quickly with previously cached value

miekg:
>> @@ -159,8 +159,8 @@ func (t *Transport) SetExpire(expire time.Duration) { t.expire = expire }
> func (t *Transport) SetTLSConfig(cfg *tls.Config) { t.tlsConfig = cfg }
>
> const (
>-	defaultExpire      = 10 * time.Second
>-	minDialTimeout     = 100 * time.Millisecond
>-	maxDialTimeout     = 30 * time.Second
>-	defaultDialTimeout = 30 * time.Second
>+	defaultExpire  = 10 * time.Second
>+	minDialTimeout = 1 * time.Second
>+	maxDialTimeout = 30 * time.Second
>+	readTimeout    = 5 * time.Second // this is the normal value for some resolve take quite a while, usually (cached) responses are fast.
>
>originally (before dynamic timeout) we used 2s as timeout (https://github.com/coredns/coredns/blob/master/plugin/forward/proxy.go#L84). 5s is too much because we won't be able to retry the query via different upstreams.
>As I mentioned in #2306, I would prefer to have this timeout configurable. If not, I would prefer value 1.5s to be able to make at least 3 attempts in 5s.
>Yes, resolution of some domain names may take longer time on first query, but if client will send the same request again to the same resolver - this may get resolved quickly with previously cached value

yeah, we might need to knob. I'll put in the original value; although I believe 
that only protects against a single failure mode: upstream crashes after we send 
it our query. Long resolution timeouts will most likely be mimicked on the other 
upstream as well.

PR #2320, opened 2018-11-17
plugin/forward: remove truncate logic

miekg:
I think this is causing problem and it will actually clash with the
scrubbing that now happens for all plugins anyway. We're assuming the
returned message will be valid even with tc being set. request.Scrub
follows that same logic.


miekg:
cc @grobie

rdrozhdzh (dismissed):
I see two issues here
1. function `truncated()` cleared error `dns.ErrTruncated`. With your change this error will prevent writing message to writer, i.e. `Scrub()` won't be called
2. function `truncated()` correctly handled `nil` messages in conjunction with error `dns.ErrTruncated`. `Scrub()` doesn't expect `nil` messages at all, it will panic

miekg:
>rdrozhdzh requested changes on this pull request.
>
>I see two issues here
>1. function `truncated()` cleared error `dns.ErrTruncated`. With your change this error will prevent writing message to writer, i.e. `Scrub()` won't be called
>2. function `truncated()` correctly handled `nil` messages in conjunction with error `dns.ErrTruncated`. `Scrub()` doesn't expect `nil` messages at all, it will panic

Upstream issue to remove this entire thing that requires(d) this logic: https://github.com/miekg/dns/issues/814

rdrozhdzh:
I'm ok with such approach
One more note. Method `Forward.Forward()` doesn't write message to writer, and therefore doesn't call `Scrub()`. Outer code should take care of calling Scrub() if needed. This may affect behavior of existing code

miekg:
>I'm ok with such approach
>One more note. Method `Forward.Forward()` doesn't write message to writer, and therefore doesn't call `Scrub()`. Outer code should take care of calling Scrub() if needed. This may affect behavior of existing code

that's a very good point, and I hadn't though about it. Let me fix
some stuff there as well.

rdrozhdzh (approved):
LGTM
The issues were addressed in #2331 and #2346

PR #2330, opened 2018-11-21
plugin/route53: Add the missed braces for the upstream example in README.md.

jiachengxu:
Add the missed braces for the route53 example with upstream.


yongtang (approved):
LGTM

PR #2331, opened 2018-11-21
plugin/forward: scrub lookup/forward replies as well

miekg:
Add Scrub here as well


rdrozhdzh (approved):

yongtang (approved):
LGTM

PR #2332, opened 2018-11-22
plugin/coredns: Add clouddns plugin logic

lainra:
This PR is implementing a new plugin called `CloudDNS`

This plugin implements Google CloudDNS service to be used as a DNS zone file backend by CoreDNS.
It is heavily inspired by the CoreDNS `route53` plugin.
It fetches resource record sets from CloudDNS to dynamically populate the `file` plugin.
By default, RRsets are pulled every minute.

Google authentication is done by using OAuth2 with a JSON credentials file.
It is also required to specify the project along with the DNS managed zone as the DNS API only allows to fetch a zone using a project name; whereas Route53 allow fetching using the zone ID.

Tests are yet to be written, however, the plugin is already working and tested.

### DIFF
--- a/plugin/clouddns/clouddns.go
+++ b/plugin/clouddns/clouddns.go
@@ -1,0 +1,275 @@
+package clouddns
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/file"
+	"github.com/coredns/coredns/plugin/pkg/fall"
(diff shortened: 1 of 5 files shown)


lainra:
Hi @miekg, sorry, I created this PR unintentionally.
I've been working on this plugin this week but I will properly make a PR from my company repository CoreDNS fork later, as the current plugin state doesn't meet the contribution requirements.

miekg:
Ah. Ok

Note it would be nice see an OWNERS file as well, with you as a new maintainer for this plugin. And probably make sense to remove any duplicated code and create a shared package in plugin/pkg.

Also note; I see no reason not to allow this plugin as we already have route53

PR #2333, opened 2018-11-22
Scrub: Do more to avoid fragmentation

miekg:
Add a check to avoid fragmentation by checking the transport and
protocol and if udp enable compression if the packet is larger than that
limit.

### DIFF
--- a/request/request.go
+++ b/request/request.go
@@ -240,6 +240,23 @@ func (r *Request) Scrub(reply *dns.Msg) *dns.Msg {
 	reply.Compress = false
 	rl := reply.Len()
 	if size >= rl {
+		if r.Proto() != "udp" {
+			return reply
+		}
+
+		// Last ditch attempt to avoid fragmentation, if the size is bigger than the v4/v6 UDP fragmentation
+		// limit and sent via UDP compress it (in the hope we go under that limit). Limits taken from NSD:
+		//
+		//    .., 1480 (EDNS/IPv4), 1220 (EDNS/IPv6), or the advertized EDNS buffer size if that is
+		//    smaller than the EDNS default.
+		// See: https://open.nlnetlabs.nl/pipermail/nsd-users/2011-November/001278.html
+		if rl > 1480 && r.Family() == 1 {
+			reply.Compress = true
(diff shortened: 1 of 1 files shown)


miekg:
ping @fturib

fturib:
oops sorry. Did not see I was reviewer here. Let me come back later today.

request/request.go
```diff
@@ -240,6 +240,23 @@ func (r *Request) Scrub(reply *dns.Msg) *dns.Msg {
...
+		//    .., 1480 (EDNS/IPv4), 1220 (EDNS/IPv6), or the advertized EDNS buffer size if that is
+		//    smaller than the EDNS default.
+		// See: https://open.nlnetlabs.nl/pipermail/nsd-users/2011-November/001278.html
+		if rl > 1480 && r.Family() == 1 {
+			reply.Compress = true
+		}
+		if rl > 1220 && r.Family() == 2 {
+			reply.Compress = true
+		}
+
```
fturib, line 259: For what I understand, default UDP msg size is:
- 512
- 1480 if EDNS and IPv4
- 1220 if EDNS and IPv6

Should not check if request contains EDNS flag ?  and, depending result and the Family, check against those 3 values ?

miekg:
No this is just an attempt to prevent fragmentation. This has nothing to do
with edns0 buffer sizes.

fturib:
In that case, how much deos that realted to the doc pointed in the comment of this code ? https://open.nlnetlabs.nl/pipermail/nsd-users/2011-November/001278.html

Looking about UDP and fragmentation, 

I can see 1440 seems to be the IP Paylod for IPv4
and for IPv6 I found this document https://labs.apnic.net/?p=813, that would recommend 1280 where you are using 1220 in your code.

fturib (approved):
for IPv6 this document (https://labs.apnic.net/?p=813) advise 1280. However 1220 is lower so should work too. 
for IPv4, this document (https://notes.shichao.io/tcpv1/ch10/#example-udpipv4-fragmentation) align on 1440 for the IPv4 paylod w/o fragmentation.

Issue #2280, opened 2018-11-07
Authorization plugin proposal


ronenschafferibm:
Hi Community,

I have a need to authorize (i.e., deny or allow) DNS requests based on sender IP and query (e.g., domain and/or type).
Has this been implemented or suggested and if so, what was the feedback?

If not, I would like to propose a new plugin to enable authorization. Please let me know if that would be of interest and I will submit a detailed design for discussion.

fturib:
@ronenschafferibm, I am currently working on a plugin named 'firewall'.
Unfortunately I did not propose yet the specs on coredns/coredns.

It is working with the `metadata` plugin, and an expression engine that allow to combine these metadata in order to provide an expression.

the Corefile looks like:

```
.:4560 {
   # provide the ability to have plugin adding metadata in the context
   metadata
   # request is converting data of the query into metadata (qname, atype, ...)
   request

   # firewall allow to filter the query or the response
   # each rule is evaluated in sequence and if the expression return a TRUE result, the query is processed according to the first keyword (allow, block, refuse, drop)
   firewall query {
      allow [request/qname] =~ 'infoblox.com'
      allow [request/qname] =~ 'google.com' && ([request/qtype] == 'A' || [request/qtype] == 'AAAA')
      block true
   }
   
   # same filtering on the reply, Metadata mechanism allow the request plugin to provide the address returned as a metadata
   firewall reply {
      refuse [response/address] =~ '172.217'   # refuse A replied by Google.com
      allow true
   }

   # regular processing of the query, once filtering by firewall is done
   forward . 8.8.8.8 {
      force_tcp
   }
   log
   errors
   reload 3s
}
```

Let me know if that correspond to what you are expecting.

NOTE: client_ip, qname, qtype are expected to be standard metadata, along with EDNS0 field extraction.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Authorization..." ]
>Let me know if that correspond what you where expected.
>
>NOTE: client_ip, qname, qtype are expected to be standard metadata, along with EDNS0 field extraction.

I'm exited to see this new *firewall* plugin!

Good stuff

elevran:
@fturib this sounds relevant for what @ronenschafferibm was proposing (he and I work together).

What is your ETA for the firewall plug-in? 
[...]

PR #2335, opened 2018-11-26
plugin/firewall - new plugin filtering Query and Response

### LINKED ISSUE #2293: plugin/log improve performance
### Reason for this issue

**In the pkg/replacer**

During implementation of metadata support for plugin Log
comment was raised on ability to improve the performance by creation of a single Map for lazy initialization.

See : https://github.com/coredns/coredns/pull/2251#issuecomment-434690523

fturib:
This PR introduce a new plugin : Firewall
it aims to allow disruption of the process based on rules applied on the Query and/or on the Response.

A sample of corefile using this plugin will be a better explanation:

```
.:53 {
   # firewall allow to filter the query or the response
   # each rule is evaluated in sequence and if the expression return a TRUE result, the query is processed according to the first keyword (allow, block, refuse, drop)
   firewall query {
      allow name =~ 'infoblox.com'
      allow name =~ 'google.com' && (type == 'A' || type == 'AAAA')
      block true
   }
   
   # same filtering on the reply, Metadata mechanism allow the request plugin to provide the address returned as a metadata
   firewall response {
      refuse response_ip =~ '172.217'   # refuse A replied by Google.com
      allow true
   }

   # regular processing of the query, once filtering by firewall is done
   forward . 8.8.8.8 {
      force_tcp
   }
   log
   errors
   reload 3s
}
```

Main points are:
- firewall allow only 2 points of enforcement : `query` and `response`
- each enforcement point is a RuleList : a list of rules that are evaluated in sequence. The first one that return a decision about the workflow (Allow, Refuse, Block, Drop) interrupt the sequence and the workflow is applied.
- it is designed to allow different policy engines : an interface `Engineer` is defined to allow any plugin to behave like a policy engine.
- in this PR, only one engine is provided, integrated in the plugin/pkg/plicy : ExpressionEngine based on the evaluator Knetic/govaluate. 
 - The expressions are pre-compiled at setup time to be executed faster at query time.
 - Expression engine support a data extractor from query/response (similar to replacer). It is designed in pkg/polic/rqdata
- Expression engine also support any metadata as designed by the metadata plugin.

NOTE: the lib github.com/Knetic/govaluate is added directly in vendor. I am not sure the is the appropriate way to add a library.

Similar plugin is proposed here : #2280 
NOTE: implementation of rqdata could be easily reused in replacer to resolve #2293 

See Firewall/README.


miekg:
looking into it. its 1500 lines so it may take a while

plugin/firewall/README.md
```diff
@@ -0,0 +1,114 @@
...
+Exemples of expression using the usual metadata:
+* `client_ip == '10.0.0.20'`
+* `type == 'AAAA'`
+* `type IN ('AAAA', 'A', 'TXT')`
+* `type IN ('AAAA', 'A') && name =~ 'google.com'`
+* `[mac/address] =~ '.*:FF:.*'`
+
+NOTE: because of the `/` separator included in a label of metadata, those labels must be enclosed on bracket [...] for a correct evaluation by the expression engine
+
+The following names are supported for qyer information
```
miekg: query

plugin/firewall/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+* `type`: qtype of the request
```
miekg: like that these copy the names we use in the log package

plugin/firewall/README.md
```diff
@@ -0,0 +1,114 @@
...
+   firewall query {
+      allow name =~ 'exemple.com'
+      allow name =~ 'google.com' && (type == 'A' || type == 'AAAA')
+      block true
+   }
+}
+~~~
+
+
+WARNING: following corefile description expect that metadata plugin know ensure definition of EDNS0 data
```
miekg: this sentence doesn't look right. Might also be just a Note: instead of a warning?

plugin/firewall/engineer.go
```diff
@@ -0,0 +1,9 @@
+package firewall
+
+import "github.com/coredns/coredns/plugin/pkg/policy"
+
+// Engineer allow registration of Engines for Policy. On plugin can declare several Engines, each of these are defined by a name.
```
miekg: this needs slightly better english

plugin/firewall/engineer.go
```diff
@@ -0,0 +1,9 @@
...
+// any duplication of name will raise an ERROR info and the corresponding Engine will be skipped
+type Engineer interface {
+	GetEngine(name string) policy.Engine
```
miekg: Just `Engine` is more idiomatic.

plugin/firewall/firewall.go
```diff
@@ -0,0 +1,125 @@
+// Package firewall enables filtering on query and response using direct expression as policy.
+// it allows interact with other Policy Engines if those are plugin implementing the Engineer interface
+package firewall
+
+import (
+	"context"
+	"errors"
+
+	"github.com/coredns/coredns/plugin/pkg/dnstest"
```
miekg: fmt the imports

plugin/firewall/firewall.go
```diff
@@ -0,0 +1,125 @@
...
+	default:
+		// Any other action returned by RuleLists is considered an internal error
+		status = dns.RcodeServerFailure
+		err = errInvalidAction
+	}
+	return p.buildDNSReply(status, false, err, w, r)
+
+}
+
+func (p *firewallPlugin) buildDNSReply(status int, errorExec bool, err error, w dns.ResponseWriter, r *dns.Msg) (int, error) {
```
miekg: `buildReply` ?

plugin/firewall/rulelist.go
```diff
@@ -0,0 +1,117 @@
+package firewall
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/coredns/coredns/plugin/pkg/policy"
```
miekg: fmt

Makefile
```diff
@@ -30,6 +30,7 @@ godeps:
 	go get -u github.com/miekg/dns
 	go get -u github.com/prometheus/client_golang/prometheus/promhttp
 	go get -u github.com/prometheus/client_golang/prometheus
+	go get -u github.com/Knetic/govaluate
```
miekg: why can't this be vendored?

plugin/pkg/policy/rqdata_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	w := dnstest.NewRecorder(&test.ResponseWriter{})
+
+	r := new(dns.Msg)
+	r.SetQuestion("example.org.", dns.TypeHINFO)
+	r.MsgHdr.AuthenticatedData = true
+	state := request.Request{Req: r, W: w}
+
+	return &RequestDataExtractor{state, mapping}
+
```
miekg: newline

plugin/pkg/policy/rqdata.go
```diff
@@ -0,0 +1,215 @@
...
+// addrToRFC3986 will add brackets to the address if it is an IPv6 address.
+func addrToRFC3986(addr string) string {
+	if strings.Contains(addr, ":") {
+		return "[" + addr + "]"
+	}
+	return addr
+}
+
+// getRespIP return the first A or AAAA records found in the Answer of the DNS msg
+func getRespIP(r *dns.Msg) net.IP {
```
miekg: s/getR/r/

plugin/pkg/policy/rqdata.go
```diff
@@ -0,0 +1,215 @@
...
+
+// RequestDataExtractor implements a GetValue(name) (value, valid) function
+// which allow to extract data from an existing DNS Request(or state)
+type RequestDataExtractor struct {
+	state     request.Request
+	requester *RequestExtractorMapping
+}
+
+// NewRequestExtractorMapping build the mapping name -> func to extract data from the Request
+func NewRequestExtractorMapping(emptyValue string) *RequestExtractorMapping {
```
miekg: this looks super expensive, is this done for all request? Why is this needed?

plugin/pkg/policy/rqdata.go
```diff
@@ -0,0 +1,215 @@
...
+			return ""
+		},
+	}
+	return &RequestExtractorMapping{replacements, emptyValue}
+}
+
+// GetValue extract the data that is mapped to this name and return the corresponding value as a string
+// if that value is empty then the defaultValue is returned
+// Second parameter is a boolean that inform if the name itself is supported in the mapping
+func (rd *RequestDataExtractor) GetValue(name string) (string, bool) {
```
miekg: s/GetV/V/

plugin/pkg/policy/rqdata.go
```diff
@@ -0,0 +1,215 @@
...
+	var ip net.IP
+	for _, rr := range r.Answer {
+		switch rr := rr.(type) {
+		case *dns.A:
+			ip = rr.A
+
+		case *dns.AAAA:
+			ip = rr.AAAA
+		}
+		// If there are several responses, currently
```
miekg: this keeps being a weird thing to do, because this is such  a narrow use-case. You'll have to deal with multiple returned IP addresses

plugin/pkg/policy/rqdata.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	if h.CheckingDisabled {
+		flags[i] = "cd"
+		i++
+	}
+	return strings.Join(flags[:i], ",")
+}
+
+// addrToRFC3986 will add brackets to the address if it is an IPv6 address.
+func addrToRFC3986(addr string) string {
```
miekg: this is an odd function to need? Why is it needed?

plugin/pkg/policy/rqdata.go
```diff
@@ -0,0 +1,215 @@
...
+func boolToString(b bool) string {
+	if b {
+		return "true"
+	}
+	return "false"
+}
+
+// flagsToString checks all header flags and returns those
+// that are set as a string separated with commas
+func flagsToString(h dns.MsgHdr) string {
```
miekg: why does this needs to become a string? For metadata?

plugin/pkg/policy/expression_test.go
```diff
@@ -0,0 +1,123 @@
...
+		if err != nil {
+			if !test.errorBuild {
+				t.Errorf("Test %d : unexpected error at build rule : %s", i, err)
+			}
+			continue
+		}
+		if test.errorBuild {
+			t.Errorf("Test %d : no error at BuilRule returned, when one was expected", i)
+		}
+
```
miekg: extra line

plugin/pkg/policy/expression_test.go
```diff
@@ -0,0 +1,123 @@
+package policy
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/coredns/coredns/plugin/pkg/dnstest"
+	tst "github.com/coredns/coredns/plugin/test"
+	"github.com/coredns/coredns/request"
```
miekg: fmt

plugin/pkg/policy/expression.go
```diff
@@ -0,0 +1,118 @@
...
+
+	if result {
+		return r.kind, nil
+	}
+	return TypeNone, nil
+}
+
+// Get return the value associated with the variable
+// DataRequestExtractor is evaluated first, and if the name does not match then metadata is evaluated
+func (p *dataAsParam) Get(name string) (interface{}, error) {
```
miekg: Get -> Value?

plugin/pkg/policy/expression.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	if kind == TypeNone {
+		return nil, fmt.Errorf("invalid keyword %s for a policy rule", keyword)
+	}
+	return &policyRuleExpression{kind, x.kindIfErrorEvaluation, e}, nil
+}
+
+func toBoolean(v interface{}) (bool, error) {
+	if s, ok := v.(string); ok {
+		return strings.ToLower(s) == "true", nil
```
miekg: why does this to a ToLower()?

plugin/pkg/policy/engine.go
```diff
@@ -0,0 +1,71 @@
...
+//   - one of the other TypeAllow/TypeRefuse/TypeDrop/TypeBlock otherwise
+type Rule interface {
+	Evaluate(data interface{}) (byte, error)
+}
+
+// Engine for Firewall plugin.
+// each Engine must be able to build rules (at setup time) base on one line of the corefile configuration
+// build at execution time some data extracted from context or query itself, needed to evaluate the rules builts.
+type Engine interface {
+	// BuildRules - create a Rule based on args or throw an error, This Rule will be evaluated during processing of DNS Queries
```
miekg: please normalize these doc comments  and unify the exported docs and the inline docs after each field that also(?) tell something I need to know?

plugin/pkg/policy/engine.go
```diff
@@ -0,0 +1,71 @@
+package policy
+
+import (
+	"context"
+
+	"github.com/coredns/coredns/request"
+)
+
+const (
+	// TypeNone a no value
```
miekg: typos in the comment

plugin/firewall/setup.go
```diff
@@ -0,0 +1,178 @@
...
+					eng[re.plugin] = names
+				}
+				if _, ok := names[re.name]; !ok {
+					names[re.name] = re.name
+				}
+			}
+		}
+	}
+
+	// retrieve all needed Engine.
```
miekg: typos

plugin/firewall/setup.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+	}
+	return p, nil
+}
+
+func (p *firewallPlugin) parseOptionOrRule(c *caddy.Controller) (*ruleElement, error) {
+	// by default, at least one engine is available : the ExpressionEngine
+	e := &policy.ExpressionEngine{}
+	switch c.Val() {
+	// check each possible option of the plugin
```
miekg: superfluous comment

miekg (changes requested):
good stuff. Probably will find some nits/comments. But very nice to see this plugin.

miekg:
I think the rule related code should be moved to a rule subdirectory and get a `package rule` to split things u pa bit, WDYT?

plugin/firewall/firewall.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/miekg/dns"
+)
+
+var (
+	errInvalidAction = errors.New("invalid action")
+)
+
+// firewallPlugin represents a plugin instance that can validate DNS
+// requests and replies using rulelists on the query and/or on the reply
+type firewallPlugin struct {
```
miekg: s/Plugin//

plugin/firewall/firewall.go
```diff
@@ -0,0 +1,125 @@
...
+	engines map[string]policy.Engine
+	query   *ruleList
+	reply   *ruleList
+
+	next plugin.Handler
+}
+
+func newFirewallPlugin() (*firewallPlugin, error) {
```
miekg: s/newFirewallPlugin/new/

plugin/firewall/firewall.go
```diff
@@ -0,0 +1,125 @@
...
+	pol := &firewallPlugin{}
+	pol.engines = make(map[string]policy.Engine, 0)
+	pol.engines["--default--"] = policy.NewExpressionEngine()
```
miekg: these 3 lines can be just 1 line does initialize the plugin and sets the default.

plugin/firewall/setup.go
```diff
@@ -0,0 +1,178 @@
...
+				return err
+			}
+		}
+		return nil
+	})
+
+	return nil
+}
+
+func firewallParse(c *caddy.Controller) (*firewallPlugin, error) {
```
miekg: s/firewallParse/parse/

plugin/firewall/setup.go
```diff
@@ -0,0 +1,178 @@
...
+				}
+			}
+		}
+	}
+
+	// build a list of all engines not found
+	allnames := make([]string, 0)
+	for pln, names := range eng {
+		for n := range names {
+			allnames = append(allnames, pln+"/"+n)
```
miekg: you can break here and report the erorr, removing the need for `allnames` and friends alltogther.

plugin/pkg/policy/engine.go
```diff
@@ -0,0 +1,71 @@
...
+	TypeNone = byte(iota)
```
miekg: why is this a byte?

plugin/pkg/policy/engine.go
```diff
@@ -0,0 +1,71 @@
...
+
+func initNameTypes() {
+	NameTypes[TypeNone] = nameTypeNone
+	NameTypes[TypeRefuse] = nameTypeRefuse
+	NameTypes[TypeAllow] = nameTypeAllow
+	NameTypes[TypeBlock] = nameTypeBlock
+	NameTypes[TypeDrop] = nameTypeDrop
+}
+
+func init() {
```
miekg: this shouldn't be done in an init function

plugin/pkg/policy/engine.go
```diff
@@ -0,0 +1,71 @@
...
+	nameTypeRefuse = "refuse"
+	nameTypeBlock  = "block"
+	nameTypeDrop   = "drop"
+)
+
+// NameTypes keep a mapping of the byte constant to the corresponding name
+var NameTypes [TypeCount]string
+
+func initNameTypes() {
+	NameTypes[TypeNone] = nameTypeNone
```
miekg: NameType is only use in the setup function of firewall, why is all this stuff needed and exported?

miekg (changes requested):

plugin/firewall/README.md
fturib: ok

plugin/firewall/README.md
fturib: I remove it. However the comment depends of the implementation of Metadata and its ability to transform EDNS0 local option into labels. 
I may have to change the readme depending the outcome of the PR #2308

plugin/firewall/engineer.go
fturib: I tried..

plugin/firewall/engineer.go
fturib: ok. changed

plugin/firewall/firewall.go
fturib: ok

plugin/firewall/firewall.go
fturib: ok. that is just a local function.

plugin/firewall/rulelist.go
fturib: ok

Makefile
fturib: It can be. I just do not know how to do it.

plugin/pkg/policy/rqdata_test.go
fturib: ok

plugin/pkg/policy/rqdata.go
fturib: It is expensive. But it is **created only once at setup of Firewall**.
Then, on each query, the map is associated with the State of the query for extraction of the data - if needed by the expression

plugin/pkg/policy/rqdata.go
fturib: ok

plugin/pkg/policy/rqdata.go
fturib: Coming as is from "replacer" : see plugin/pkg/replacer/replacer.go

plugin/pkg/policy/rqdata.go
fturib: Not related to metadata.
I just keep the same definition and format that is currently in "replacer" (for the Log plugin)

plugin/pkg/policy/expression_test.go
fturib: ok

plugin/pkg/policy/expression_test.go
fturib: ok (I think)

plugin/pkg/policy/expression.go
fturib: That is an interface of Expression to implement in order to evaluate the "variable" of the expression.
I have no choice but to use the required name of the interface

plugin/pkg/policy/expression.go
fturib: accept all forms of "TRUE", "True", true", etc ...

plugin/pkg/policy/engine.go
fturib: ok. I tried to simplify and clarify the comments. To be reviewed.

plugin/pkg/policy/engine.go
fturib: ok

plugin/firewall/setup.go
fturib: ok

plugin/firewall/setup.go
fturib: ok

plugin/firewall/firewall.go
fturib: ok

plugin/firewall/firewall.go
fturib: ok

plugin/firewall/firewall.go
fturib: ok

plugin/firewall/setup.go
fturib: ok

plugin/firewall/setup.go
fturib: Ok, but we will report only one missing engine at each setup where we could report all togethers.

plugin/pkg/policy/engine.go
fturib: not mandatory ... I removed.

plugin/pkg/policy/engine.go
fturib: and where then ? it is a Map and cannot be created as a constant.

plugin/pkg/policy/engine.go
fturib: Engine is an API for policy engines. The TypesXXX are predefined. 
I think that the names of these types should be visible for these policy engines that will implement Engine.

fturib:
@johnbelamaric : sorry, I did not see you selected for 1.3.0. I guess it will be in next release.

fturib:
I need to sort out why the code is no more compiling on my local machine. 
Several errors with an googleapis v2 somewhere ...
Seems to be tied to update of Kubernetes APIs ... (or my local repo is obsolete)

plugin/pkg/policy/rqdata.go
fturib: In case of firewall, it is not such narrow : you would like to block the reply depending the network that replie belongs to. 
However, I agree we need to take care of multiple replies.
I hope to tune that in a subsequent PR : 
- Expression engine support functions and parameters as arrays.
- Also I would like to add a in_network(...)  or in_cidr(...) function in order to make an easy test.

I would like to be able to enforce a rule that does not allow a reply out of corporate, or out of cluster:

```
firewall response {
    allow  in-cidr(response_ip, '10.0.0.0/16')
    refuse true
}
```

fturib:
Re-submitted, trying to address all comments.
Also found the way to vendorize the Knetic/govaluate library.

Thanks for a second review.

fturib:
hum.. still some trouble with the dependency... working on it.

fturib:
ok. the Expression library is now properly vendorized. All checks are ok.
Thanks for a continuing the review !

miekg:
Ok. I'll re-review in the coming days.

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+## Name
+
+*firewall* - enables filtering on query and response using direct expression as policy.
+
+## Description
+
+The firewall plugin defines a rule list of expressions that triggers workflow action on the DNS query or its response.
+A rule list is an ordered set of rules that are evaluated in sequence.
+Each rule has two parts: an action and an expression. When the rule is evaluated,
+first the expression get evaluated using Metadata and a go-like language.
```
chrisohaver:
```suggestion
first the expression is evaluated.
```

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+- If the expression evaluates to `true` the action is performed on the query and the rule list evaluation is considered over.
```
chrisohaver:
```suggestion
- If the expression evaluates to `true` the action is performed on the query and the rule list evaluation ceases.
```

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+- If the expression does not evaluates to `true` then next rule in sequence get evaluated
```
chrisohaver:
```suggestion
- If the expression does not evaluates to `true` then next rule in sequence is evaluated.
```

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+~~~ txt
+firewall APPLICATION {
+    ACTION EXPRESSION
+    ACTION EXPRESSION
+    ...
+}
+~~~~
+
+* **APPLICATION** defines where is applied the firewall rulelist. It can be `query` or `response`
```
chrisohaver: **APPLICATION** doesn't feel like the right word here... Is this a standard firewall term?

Perhaps "DIRECTION" instead?

chrisohaver:
```suggestion
* **APPLICATION** indicates if the _rule list_ will be applied to queries or responses.  It can be `query` or `response`
```

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+* **ACTION** defines the workflow action to apply to the DNS operation if the **EXPRESSION** evaluates to the boolean `true`
+action is one of:
+  - `allow` : continue the DNS resolution process
+  - `refuse` : interrupt the DNS resolution, just return the REFUSE result
+  - `block` : interrupt the DNS resolution, just return the NXDOMAIN result
+  - `drop` : interrupt the DNS resolution, just return the NODATA result
```
chrisohaver: Based on the name "drop" I would expect no answer at all, (i.e. clients times out).

For a NODATA response, I think a better name might be "hide"?

But, what are the use cases for forcing NXDOMAIN vs NODATA? They are both negative answers, and the distinction between them means something very specific in DNS.  Perhaps there should only be a single "block" action, that returns NXDOMAIN or NODATA, based on whether or not there are other record types with the same name?

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+* **EXPRESSION** defines the expression to evaluate in order to validate the action and interrupt the sequence of rules.
```
chrisohaver, line 36:
```suggestion
* **EXPRESSION** defines the expression
```

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+Expression is a [go-like language](https://github.com/Knetic/govaluate/blob/master/MANUAL.md)
```
chrisohaver, line 37: govaluate authors say the language is C-like in the README.

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+where the variables are either the `metadata` of CoreDNS either a list of names associated with the DNS query/response information.
```
chrisohaver, line 38:
```suggestion
where the variables are either the _metadata_ of CoreDNS either a list of properties of the DNS query/response.
```

Do we need to define what _metadata_ is here, or perhaps link to where we define it elsewhere?
Also, we should be consistent with how we mark up "metadata", either make it bold, italic, or literal (backticks)... whichever is appropriate.

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+Usual operators applies.
```
chrisohaver, line 39:
```suggestion
```

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+Exemples of expression using the usual metadata:
```
chrisohaver, line 41:
```suggestion
Examples of expressions:
```

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+* `client_ip == '10.0.0.20'`
+* `type == 'AAAA'`
+* `type IN ('AAAA', 'A', 'TXT')`
+* `type IN ('AAAA', 'A') && name =~ 'google.com'`
+* `[mac/address] =~ '.*:FF:.*'`
+
+NOTE: because of the `/` separator included in a label of metadata, those labels must be enclosed on bracket [...] for a correct evaluation by the expression engine
```
chrisohaver, line 48:
```suggestion
NOTE: metadata labels containing a `/` must be enclosed in brackets, e.g. `[mac/address]`
```

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+* `port`: client's port
+* `duration`: response duration
+* `rcode`: response CODE (NOERROR, NXDOMAIN, SERVFAIL, ...)
+* `rsize`: raw (uncompressed), response size (a client may receive a smaller response)
+* `>rflags`: response flags, each set flag will be displayed, e.g. "aa, tc". This includes the qr
+  bit as well
+* `>bufsize`: the EDNS0 buffer size advertised in the query
+* `>do`: is the EDNS0 DO (DNSSEC OK) bit set in the query
+* `>id`: query ID
+* `>opcode`: query OPCODE
```
chrisohaver, line 67: Curious, why do some variables start with `>`...

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+* `server_ip`: server's IP address, for IPv6 addresses these are enclosed in brackets: `[::1]`
+* `server_port` : client's port
+* `response_ip` : the IP returned in the first A or AAAA record of the Answer section
+
+
+## Examples
+
+Allow queries for exemple.com.
```
chrisohaver: This is a confusing example...

is the spelling error intentional, and org/com switch also intentional? 
The stanza is for "example.org" ... so it wont process "exemple.com" or "google.com"

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+   firewall query {
+      allow name =~ 'exemple.com'
+      allow name =~ 'google.com' && (type == 'A' || type == 'AAAA')
+      block true
+   }
+}
+~~~
+
+
+Define the metadata labels `group_id` and `client_id` based on the content of OPT records (EDNS0)
```
chrisohaver:
```suggestion
Define the metadata labels `group_id` and `client_id` with values extracted from EDNS0
```

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+and use those values to filter the DNS queries: any query that have not the group_id 123456789 AND that has not the client_id matching ABCDEF will be returned REFUSED
```
chrisohaver:
```suggestion
and use those values to filter the DNS queries: any query that doesn't have the group_id of `123456789` AND doesn't have the client_id of `ABCDEF` will be returned REFUSED.
```

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+Also filter the reply with disallowing IP returned within a specific regular expression.
```
chrisohaver:
```suggestion
In addition, respond with a REFUSE if the first `A`/`AAAA` record in the response contains an IP in 172.217.0.0/16.
```

Actually - i'm guessing that "=~" is a regex compare (a la Perl/Ruby). Do you need to escape the "." here?

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+      group_id edns0 0xffed bytes
+      client_id edns0 0xffee bytes
+   }
+   firewall query {
+      refuse [metadata/client_id] != 'ABCDEF'
+      refuse [metadata/group_id] != '123456789'
+      block true
+   }
+   firewall response {
+      refuse  response_ip =~ '172.217.*'   # refuse any IP that is in 172.217.0.0/16
```
chrisohaver, line 107: Is there a practical application of this type of rule? Refuse if the first A/AAAA record is in an IPv4 cidr? 
It hints that the intent is to hide all responses containing IPs in 172.217.0.0/16.  How would it handle multiple A records with same name, SRV responses with the IP in the additional section...

chrisohaver (commented):
First pass on docs

miekg:
Send //github.com/coredns/coredns/pull/2437 to move the rule stuff to it's own sub package.

* that pr still needs proper documentation on the now exported fields.
* I have a hard time wrapping my head around what all the different Ensure* function to and why there are so many
* pkg/policy needs to be move to this plugin's directory as well.

after those changes I can review more indepth

fturib:
I integrated the changes from [PR/2437](https://github.com/coredns/coredns/pull/2437)
- moving rulelist into a sub-package rule
- moving engine and expression from pkg/policy to firewall/policy
- moving rqdata from pkg/policy to pkg/rqdata 

I also integrated the changes of README propose by @chrisohaver

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+    ...
+}
+~~~~
+
+* **DIRECTION** indicates if the _rule list_ will be applied to queries or responses. It can be `query` or `response`.
+
+* **ACTION** defines the workflow action to apply to the DNS operation if the **EXPRESSION** evaluates to the boolean `true`
+action is one of:
+  - `allow` : continue the DNS resolution process
+  - `refuse` : interrupt the DNS resolution, just return the REFUSE result
```
miekg: what's "result"? Do you mean you return a packet with rcode (response code) REFUSE?

plugin/firewall/policy/engine.go
```diff
@@ -0,0 +1,71 @@
...
+const (
+	nameTypeNone   = "none"
+	nameTypeAllow  = "allow"
+	nameTypeRefuse = "refuse"
+	nameTypeBlock  = "block"
+	nameTypeDrop   = "drop"
+)
+
+// NameTypes keep a mapping of the byte constant to the corresponding name
+var NameTypes [TypeCount]string
```
miekg: this whole init section can be replaces with 
~~~
var NameType = [TypeCount]string{....}
~~~
the `nameTypeX` constant don't need to be defined, as they are only used here. Also wonder why you use an array and not an map? Having some inherit ordering is in the policy is also kind of confusing.

plugin/firewall/firewall.go
```diff
@@ -0,0 +1,126 @@
...
+		status    = -1
+		respMsg   *dns.Msg
+		err       error
+		queryData = make(map[string]interface{}, 0)
+	)
+
+	state := request.Request{W: w, Req: r}
+
+	// ask policy for the Query Rulelist
+	action, err := p.query.Evaluate(ctx, state, queryData, p.engines)
```
miekg, line 61: what does `queryData` do here? Why is it empty?

plugin/firewall/firewall.go
```diff
@@ -0,0 +1,126 @@
...
+	case policy.TypeDrop:
+		// One of the RuleList ended evaluation with typeDrop : simulate a drop
+		return dns.RcodeSuccess, nil
+	default:
+		// Any other action returned by RuleLists is considered an internal error
+		status = dns.RcodeServerFailure
+		err = errInvalidAction
+	}
+	return p.buildReply(status, false, err, w, r)
+
```
miekg: extra newline

plugin/firewall/firewall.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (p *firewall) buildReply(status int, errorExec bool, err error, w dns.ResponseWriter, r *dns.Msg) (int, error) {
```
miekg: buildReply looks odd, why do you give it an errorExec bool *and* and error? Is also doesn't use anything from `p`, so why it is a method instead of a function. 
And status should be rcode prolly

plugin/firewall/firewall.go
```diff
@@ -0,0 +1,126 @@
...
+
+	r.Rcode = status
+	r.Response = true
+	if status != dns.RcodeServerFailure || errorExec {
+		w.WriteMsg(r)
```
miekg: you're echoing the original reply, this is prolly not write. You need to create a new msg and use `SetReply`.

plugin/firewall/firewall.go
```diff
@@ -0,0 +1,126 @@
...
+		return nil, err
+	}
+	if pol.reply, err = rule.NewList(policy.TypeAllow, true); err != nil {
+		return nil, err
+	}
+	return pol, nil
+}
+
+// ServeDNS implements the Handler interface.
+func (p *firewall) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
```
miekg, line 50: this method doesn't check if it is authoritative for the domains it is seeing.

plugin/firewall/firewall.go
miekg: err can only be `errInvalidAction` it seems. Looks like you can just handle that separately without calling this function and giving it an `err`.

plugin/firewall/firewall.go
```diff
@@ -0,0 +1,126 @@
...
+	r.Rcode = status
+	r.Response = true
+	if status != dns.RcodeServerFailure || errorExec {
+		w.WriteMsg(r)
+	}
+
+	return dns.RcodeSuccess, err
+}
+
+// Name implements the Handler interface
```
miekg: closing dot

plugin/firewall/policy/expression.go
```diff
@@ -0,0 +1,120 @@
+package policy
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/coredns/coredns/plugin/pkg/rqdata"
+
```
miekg: fmt

plugin/pkg/rqdata/rqdata.go
```diff
@@ -0,0 +1,220 @@
+package rqdata
```
miekg, line 1: *why* is this package needed?

plugin/firewall/policy/expression.go
```diff
@@ -0,0 +1,120 @@
...
+func (x *ExprEngine) BuildReplyData(ctx context.Context, state request.Request, query interface{}) (interface{}, error) {
+	return &dataAsParam{ctx, rqdata.NewExtractor(state, x.dataFromReq)}, nil
+}
+
+//BuildRule create a rule for Expression Engine:
+// - first param is one of the action to return
+// - second and following param is a sentence the represent an Expression
+func (x *ExprEngine) BuildRule(args []string) (Rule, error) {
+	keyword := args[0]
+	exp := args[1:]
```
miekg, line 53: why does BuildRule get a slice of strings when the first item is clearly a keyword and the rest is an expression? Why not `keyword string, expr []string)?

plugin/firewall/policy/expression.go
```diff
@@ -0,0 +1,120 @@
...
+
+	if result {
+		return r.kind, nil
+	}
+	return TypeNone, nil
+}
+
+// Get return the value associated with the variable
+// DataRequestExtractor is evaluated first, and if the name does not match then metadata is evaluated
+func (p *dataAsParam) Get(name string) (interface{}, error) {
```
miekg, line 109: it's not clear that this needs to be implemented because of the github.com/Knetic/govaluate use

plugin/firewall/policy/expression.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"github.com/coredns/coredns/request"
+
+	"github.com/coredns/coredns/plugin/metadata"
+
+	expr "github.com/Knetic/govaluate"
+)
+
+type ruleExpr struct {
+	kind        int
```
miekg: I don't understand these `kind`s here. And also not the one below.

plugin/firewall/policy/expression.go
```diff
@@ -0,0 +1,120 @@
...
+
+// ExprEngine implement interface Engine for Firewall plugin
+// it evaluate the rues using an the lib Knetic/govaluate
+type ExprEngine struct {
+	kindIfErrorEvaluation int
+	dataFromReq           *rqdata.Mapping
+}
+
+type dataAsParam struct {
+	ctx         context.Context
```
miekg, line 29: why does this need to embed a context?

plugin/firewall/policy/expression.go
```diff
@@ -0,0 +1,120 @@
...
+	return &dataAsParam{ctx, rqdata.NewExtractor(state, x.dataFromReq)}, nil
+}
+
+//BuildRule create a rule for Expression Engine:
+// - first param is one of the action to return
+// - second and following param is a sentence the represent an Expression
+func (x *ExprEngine) BuildRule(args []string) (Rule, error) {
+	keyword := args[0]
+	exp := args[1:]
+	e, err := expr.NewEvaluableExpression(strings.Join(exp, " "))
```
miekg, line 54: hmm, here you glue all `args[1:]` back together, why do you give this function a string slice?

plugin/firewall/policy/expression.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	if kind == TypeNone {
+		return nil, fmt.Errorf("invalid keyword %s for a policy rule", keyword)
+	}
+	return &ruleExpr{kind, x.kindIfErrorEvaluation, e}, nil
+}
+
+func toBoolean(v interface{}) (bool, error) {
+	if s, ok := v.(string); ok {
+		return strings.ToLower(s) == "true", nil
```
miekg, line 73: the expr engine can return, True TrUe and whatever to mean true?

plugin/firewall/policy/expression.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	if b, ok := v.(bool); ok {
```
miekg, line 75: how is 'true' different from a boolean?

plugin/firewall/policy/expression.go
```diff
@@ -0,0 +1,120 @@
...
+		return b, nil
+	}
+	if i, ok := v.(int); ok {
+		return i != 0, nil
```
miekg, line 79: why is i != 0 considered here?

fturib:
@miekg : Thanks for the review. Will follow up today or tomorrow.

plugin/firewall/README.md
fturib: yes. will change the text with "reply with XXXX code".

plugin/firewall/policy/engine.go
fturib: Right. not sure why I started with an array. Will change and remove the nameXX constants

plugin/firewall/firewall.go
fturib: queryData is a map we need to keep for the response filter evaluation. Therefore it must be at ServeDNS level. But it is really filled when needed by the rule.list.Evaluate function.

plugin/firewall/firewall.go
fturib: removed.

plugin/firewall/firewall.go
fturib: buildReply is a  factorization of the reply ... but not really a big gain, and maybe confuse the code. I just removed it.

plugin/firewall/firewall.go
fturib: ok.

plugin/firewall/firewall.go
fturib: hum ... there is no ZONE in the directive .. should we add it ?

fturib: if we add, then should we consider that we could have different filter depending the zone ?
something like

```
firewall query google.com {
...
}
firewall query azure.com {
...
}
firewall query . {
...
}

```

plugin/firewall/firewall.go
fturib: ok

plugin/firewall/policy/expression.go
fturib: ok. Look like my tool is not set the same way as you wish the import to be formatted. 
But the validation in Makefile is ok ...

plugin/pkg/rqdata/rqdata.go
fturib: I am not sure I understand the question. It is a library .. therefore need to be in a package no ?

If the question is why the lib is needed: I cannot use directly "Replacer" because of the format of the variables (that includes the curly brace in replacer).
Therefor I created a tool similar, but with an optimization on the Mapping that is created only once at setup.
I expect to have Rewrite and Log using this rqdata instead of the "replacer" (or direct replacement in the case of Rewrite).

At the end, we will have a single tool that can be used by all Plugins that want to extract data from "Request". And we have #2293 fixed.

plugin/firewall/policy/expression.go
fturib: To follow the interface `Engine`.
Only the expression engine is using `keyword` + params
other engines .. we do not know that is only params.

plugin/firewall/policy/expression.go
fturib: I add a comment

plugin/firewall/policy/expression.go
fturib: agree. will replace the `kind` by `action`. It represent the "policy action" associated with the expression.

plugin/firewall/policy/expression.go
fturib: dataAsParam is the Evaluator for variable that are embeded in the expression.
Those variables can be either a name of a field of the request (resolved by the RqData), either a Label of Metadata. We need the context to resolve the Metadata labels.

plugin/firewall/policy/expression.go
fturib: To follow the interface of `Engine`.

plugin/firewall/policy/expression.go
fturib: Yes. We need a definition on what is "true".

plugin/firewall/policy/expression.go
fturib: I am not sure I understand the question.

plugin/firewall/policy/expression.go
fturib: Well, in case the epxression return a "int", I consider that if it return the value "0" then the result of expression is FALSE, else it is TRUE.

fturib:
I forgot to say, that I think I addressed all comments from last review. 
In some cases I added comment on the review comment.

It is now ready (IMO) for another review.
Thanks!

elico:
@fturib  @miekg Was this feature/plugin approved and is scheduled for merge?

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
+# firewall
+
+## Name
+
+*firewall* - enables filtering on query and response using direct expression as policy.
```
miekg, line 5: 'direct expressions' is a bit weird, what do you eamn?

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+Exemples of expression using the usual metadata:
+* `client_ip == '10.0.0.20'`
+* `type == 'AAAA'`
+* `type IN ('AAAA', 'A', 'TXT')`
+* `type IN ('AAAA', 'A') && name =~ 'google.com'`
+* `[mac/address] =~ '.*:FF:.*'`
+
+NOTE: because of the `/` separator included in a label of metadata, those labels must be enclosed on bracket [...] for a correct evaluation by the expression engine
+
+The following names are supported for querying information
```
miekg, line 50: Would link to the log's plugin and say we support the same syntax (plus a few new ones? Are there new ones?)

miekg: The `>` idea was that those are header related.

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+## Examples
+
+Allow queries for example.com.
+Allow also the queries for google.com if those are A or AAAA type of queries
+NXDOMAIN every other queries
+
+~~~ corefile
+. {
+   firewall query {
+      allow name =~ 'example.com'
```
miekg, line 82: So no closing dot? Can we detail what done here?

plugin/firewall/README.md
```diff
@@ -0,0 +1,112 @@
...
+      refuse [metadata/client_id] != 'ABCDEF'
+      refuse [metadata/group_id] != '123456789'
+      block true
+   }
+   firewall response {
+      refuse  response_ip =~ '172.217.*'   # refuse any IP that is in 172.217.0.0/16
+      allow true
+   }
+}
+~~~
```
miekg, line 111: Add `See Also` section to point to the language.

What does a minimal firewall plugin look like? I.e a block all example and a allow allow example would help here.

(the current file also has a spurious empty line on the end)

plugin/pkg/rqdata/rqdata.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// Extractor implements a Value(name) (value, valid) function
+// which allow to extract data from an existing DNS Request(or state)
+type Extractor struct {
+	state     request.Request
+	requester *Mapping
+}
+
+//NewExtractor return a new Extractor based on the Mapping and the Request provided
```
miekg, line 32: no space between // and comment. Also no closing dot (please check the rest of the code)

plugin/pkg/rqdata/rqdata.go
```diff
@@ -0,0 +1,220 @@
...
+)
+
+type requestFunc func(state request.Request) string
+
+// Mapping define the mapping between 'name' of data and the way to extract that data from the Request
+// it also defines what will be the empty value returned if the data behind the name is empty.
+// it is pretty static, and you should need to instantiate only once
+type Mapping struct {
+	replacements map[string]requestFunc
+	emptyValue   string
```
miekg, line 22: This emptyValue is odd? 1) why is `""` not enough  to signal "not found"

plugin/pkg/rqdata/rqdata.go
```diff
@@ -0,0 +1,220 @@
...
+		">opcode": func(state request.Request) string {
+			return strconv.Itoa(int(state.Req.Opcode))
+		},
+		">do": func(state request.Request) string {
+			return boolToString(state.Do())
+		},
+		">bufsize": func(state request.Request) string {
+			return strconv.Itoa(state.Size())
+		},
+
```
miekg, line 108: this one has a new line after it

plugin/pkg/rqdata/rqdata.go
```diff
@@ -0,0 +1,220 @@
...
+//NewExtractor return a new Extractor based on the Mapping and the Request provided
+func NewExtractor(r request.Request, m *Mapping) *Extractor {
+	return &Extractor{r, m}
+}
+
+// NewMapping build the mapping name -> func to extract data from the Request
+func NewMapping(emptyValue string) *Mapping {
+	replacements := map[string]requestFunc{
+		"type": func(state request.Request) string {
```
miekg, line 40: Why does all this stuff require a expansive setup of getting these values? you can just return state.Type() with the functions and without the map

plugin/firewall/firewall.go
miekg: querydata should not be a map to an empty interface, but use something typed. Make no sense to make it an `interface{}`.

Also: while looking to the `qrdata` plugin, *what* data isn't know at this point?
Also: this argument is  passed by pointer, instead of this method just returning it? It should just return it. Also very curious to learn what *extra* data will put in it that we don't already know

plugin/firewall/firewall.go
```diff
@@ -0,0 +1,128 @@
...
+			m := new(dns.Msg)
+			m = m.SetRcode(r, dns.RcodeServerFailure)
+			w.WriteMsg(m)
+			return dns.RcodeSuccess, err
+		}
+	}
+
+	// Now apply the action evaluated by the RuleLists
+	switch action {
+	case policy.TypeAllow:
```
miekg, line 100: please write back and do the house keeping in the switch statement. this will make this more readable.

plugin/firewall/policy/expression.go
miekg: It's very odd to put the ctx in the struct; I don't we do it currently. Should just give it the functions you're calling

plugin/firewall/policy/expression.go
miekg: but you control the interface definition? Isn't the rule language fixed? Why do we even need that flexibility?

plugin/firewall/policy/expression.go
miekg: which you control

plugin/firewall/policy/expression.go
miekg: please add comment to why and link to the code/documentation where this is defined (in the external code)

plugin/firewall/policy/expression.go
miekg: this is a very weird function; can't really see why an API would be so weird that we would need this function. As said above; please link to docs and impl. on why this is.

miekg (changes requested):
This starts to look much better, still some way to go though.

The rqdata package can also be moved in the firewall's directory at has no use outside of this plugin, rqdata needs a better name as well.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/firewa..." ]
>@fturib  @miekg Was this feature/plugin approved and is scheduled for merge?

not approved yet, but will be merged at some point.

plugin/firewall/README.md
fturib: ok. changed.

plugin/firewall/README.md
fturib: It is almost the same:
- but {combined} and {common} are not supported as it makes no sense
- we have few others: server_ip, server_port and response_ip
- we need for Firewall that those are not enclosed in braces {}.

the rqdata tool used here is expected to be reuse for `log` and then for `rewrite`, so we have a common reference to all names of data extracted from the query.

But, maybe we can wait it is really shared before having a common reference in README ?

plugin/firewall/README.md
fturib: Right, needs a closing dot and proper regexp.
tested with https://regex101.com/, should be `allow name =~ 'example\.com\.$'`

fturib: Does not make the syntax very easy, unless we add a function sub-zone(..) to the language

plugin/firewall/README.md
fturib: ok. removed last line. Added 2 examples.

plugin/pkg/rqdata/rqdata.go
fturib: ok

plugin/pkg/rqdata/rqdata.go
fturib: Because I plan this tool to be shared with Log. And Log replace an empty value with "-".

plugin/pkg/rqdata/rqdata.go
fturib: ok

plugin/pkg/rqdata/rqdata.go
fturib: The idea is to build the Mapping only once, when creating the plugin and reuse the same mapping with every DNS query coming in.
=> If I use direct function, I end-up creating something similar to `replacer` that created the map on each DNS request.

That is the performance improvement proposed here : https://github.com/coredns/coredns/issues/2293

plugin/firewall/firewall.go
fturib: can you rephrase ? I do not understand what is asked here.
You mean remove the comments?

fturib: about 
>querydata should not be a map to an empty interface, but use something typed. Make no sense to make it an interface{}.

I will try to simplify here.

miekg:
This is always an expansive regexp? That might make things too slow. This
probably needs splitting into a plain match and a regexp. Also, as said
closing dot or not needs to be consistent and documented.

miekg:
That's not clear and not done in this PR. Please remove any "may be handy
in the future" code from this.

miekg:
Ah. Ok that starts to make some sense, but this can be a lot cleaner as it
boils down to just this map

The duration thing that's happening here looks pretty bad and I would be
amazed if it actually works. Please remove.

miekg:
It should be a map to strings values

miekg:
I need to park this review for a bit as I want to work on more pressing
things than a feature.

I think I reviewed 1/3 of the code, but there is still plenty to go and
averall I'm not happy about the process. There wasn't an issue filled that
outlines the plan or the code. The code was just dumped in a giant PR. It
has lots of books that are not clear to me why we need them and it lack
documentation that would help a reviewer and future implementers.

I fully expect this review to go into about a 1000 comments before I'm
really happy with it, and will probably mean I need to send cleanup PRs
which means more work for me.

Package wise we getting into nits for the firewall package, but this leave
qrdata, policy and rules iirc.

fturib:
> I think I reviewed 1/3 of the code, but there is still plenty to go and
averall I'm not happy about the process. There wasn't an issue filled that
outlines the plan or the code. The code was just dumped in a giant PR. It
has lots of books that are not clear to me why we need them and it lack
documentation that would help a reviewer and future implementers.

Ok. Let's change of procedure.
It adds too much pressure on both of us and does not help for the code itself.

I propose to create an external Plugin with this Firewall, let say as an ALPHA version.
I will continue to add on it, maybe start to implement some of the enhancement for Policy Engine. 
But also some improvement on the expression language - like you suggest for performance (regexp).

Then you will have time, when you wish, to see if you want to move into an internal plugin, all or only a part, or with some refactoring if that is needed for perf.


That will allow each of us to go our pace without too much constrainst.

Issue #1851, opened 2018-06-03
Plugin/Kubernetes: A records for services with Load Balancer IP (External-IP)


leeliu:
#### Feature request:
Some type of DNS naming scheme to answer with `External-IP` for services in addition to Cluster IP/Pod IPs. (ie: `test.default.svc.cluster.local` currently points to `10.28.0.3`. Be nice if `test.default.extsvc.cluster.local` (or whatever naming scheme) could point to `35.129.2.3`.)

#### Use case:
- a coredns server within a cluster that an external domain can delegate a subdomain NS's to
- coredns answers with public load balancer IPs (`External-IP`) of running services
- perhaps an annotation on the service itself to allow different dns name than service name?
- we currently do this with route53 manually but would be nice to not have route53 as a dependency to manage internal cluster records

#### Example:
I have `example.com` where I delegate `NS` records for `core.example.com` to coredns pods running within the cluster. Allow coredns to configured to answer that it's the authoritative NS for `core.example.com`. Queries should resolve `test.core.example.com` to `35.129.2.3`. Bonus if `test` itself can be configured to be a different dns name than the internal service name through some type of annotation on the service.

ngortheone:
Very much interested in described functionality. Are there any existing solutions or at least workarounds to get this done?

chrisohaver:
This is feasible to add as a feature, and there is wide interest.  We discussed this very briefly during maintainers meeting.   We need to settle on how to implement it at a high level. @miekg, @johnbelamaric

dns-wise:
* should the external records be in an entirely new domain (e.g. `entirely.new.domain.com`)
* should the external records be in the same domain, under a different subdomain (e.g. `ext.cluster.local`)

code-wise:
* As a new option in _kubernetes_ plugin
* As a new plugin that shares the API connection with _kubernetes_ plugin.

johnbelamaric:
The user should be able to specify the domain to use for this, there is no reason to use something like `external.local`. Instead, this should work like any other plugin that manages a zone, except that it gets the data from the K8s api. for example, even if my cluster domain is `cluster.local` my external might be `services.mycompany.com`.

For this reason I think it makes sense to be a different plugin.

chrisohaver:
> For this reason I think it makes sense to be a different plugin.

A single plugin _can_ serve more than one zone...

```
kubernetes mycluster.local services.mycompany.com in-addr.arpa ip6.arpa {
  external services.mycompany.com
[...]

PR #2337, opened 2018-11-26
New Plugin: extkubernetes - Expose External IPs of Load Balancers

chrisohaver:

* Adds a new plugin (_extkubernetes_) that exposes external ip addresses of loadBalancer services
* Starts to moves some code common with _kubernetes_ to `pkg/kubernetes` (e.g. API connection, watch).

"Connection sharing" with the _kubernetes_ plugin is not implemented (that is, if both plugins are active, they do not share the same API connection).  I think the general mode of deployment would be to serve external IPs from a separate instance of CoreDNS, not from the internal cluster DNS, so connection sharing would seldom be a factor.  But if there is a demand to serve the data from the same instance, then we can implement connection sharing later.

No `coredns/ci` test yet.  It will be complicated to implement, but doable.  It will involve setting up virtual routers and load balancers (metalLB) in the framework of the test. I'll start on this only if it looks like the plugin will be accepted.

#1851
kubernetes/dns#242


chrisohaver:
FYI, @johnbelamaric

chrisohaver:
I should clarify that the state of this PR is that it is functional. I've tested it using minikube + metalLB, and it works as expected. Hence I am not marking this as a work in progress.
There is still some code duplication, more code could be moved to pkg/kubernetes.

Awaiting feedback and general acceptance of the approach before continuing.

plugin.cfg
```diff
@@ -44,6 +44,7 @@ hosts:hosts
 route53:route53
 federation:federation
 kubernetes:kubernetes
+extkubernetes:extkubernetes
```
miekg, line 47: don't really like the name, should we just go for external, or externelkubernetes or ... something?

plugin/extkubernetes/README.md
```diff
@@ -0,0 +1,97 @@
+# extkubernetes
+
+## Name
+
+*extkubernetes* - enables reading external zone data from a Kubernetes cluster for Load Balancer IPs.
+
+## Description
+
+Creates A/AAAA, SRV, and PTR records for the External IPs of each LoadBalancer type Service in a Kubernetes cluster.
```
miekg, line 9: Please full sentences. The *extkubernernetes* plugin ...

plugin/extkubernetes/README.md
```diff
@@ -0,0 +1,97 @@
...
+    kubeconfig KUBECONFIG CONTEXT
+    namespaces NAMESPACE...
+    labels EXPRESSION
+    ttl TTL
+    transfer to ADDRESS...
+    fallthrough [ZONES...]
+}
+```
+
+* `resyncperiod` specifies the Kubernetes API refresh period to be **DURATION**.
```
miekg, line 38: Just reference the kubernetes plugin here? Removes the need to duplicate all the info here.

plugin/extkubernetes/apiproxy.go
```diff
@@ -0,0 +1,75 @@
...
+type apiProxy struct {
+	http.Server
+	listener net.Listener
+	handler  proxyHandler
+}
+
+func (p *proxyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	upstream := p.Select()
+	network := "tcp"
+	address := upstream.Name
```
miekg, line 25: `address` is only used in the dial just below? No need to put this in a new var

plugin/extkubernetes/apiproxy.go
```diff
@@ -0,0 +1,75 @@
...
+
+	d, err := net.Dial(network, address)
+	if err != nil {
+		log.Errorf("Unable to establish connection to upstream %s://%s: %s", network, address, err)
+		http.Error(w, fmt.Sprintf("Unable to establish connection to upstream %s://%s: %s", network, address, err), 500)
+		return
+	}
+	hj, ok := w.(http.Hijacker)
```
miekg, line 33: why is this needed?

plugin/extkubernetes/handler.go
```diff
@@ -0,0 +1,90 @@
+package kubernetes
+
+import (
+	"context"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/request"
```
miekg, line 7: fmt

plugin/extkubernetes/handler.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/miekg/dns"
+)
+
+// ServeDNS implements the plugin.Handler interface.
+func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	opt := plugin.Options{}
```
miekg, line 14: where is opt used? Please move this closer

plugin/extkubernetes/handler_test.go
```diff
@@ -0,0 +1,309 @@
...
+	"github.com/coredns/coredns/plugin/pkg/kubernetes/object"
+	"github.com/coredns/coredns/plugin/pkg/watch"
+	"github.com/coredns/coredns/plugin/test"
+
+	"github.com/miekg/dns"
+	api "k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var dnsTestCases = []test.Case{
```
miekg, line 18: is this just duplicating stuff?

miekg:
This almost duplicates the kubernetes plugin. Please figure out a way to not do that and not fill up pkg/* with kubernetes stuff

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] New Plugin: e..." ]
>Closed #2337.

Note, Chris: I'm very much in favor of having this, but can we do it in much 
less than 3000 lines added?

fturib:
from @johnbelamaric:
> For this reason I think it makes sense to be a different plugin.

from @miekg :
> This almost duplicates the kubernetes plugin. Please figure out a way to not do that and not fill up pkg/* with kubernetes stuff

I guess there is too much constraints here.

If we want to make 2 separates plugins, then we'd better have common code (throuhg lib), and also maybe share the same connection to K8s API.


I am not sure why the number of lines make a difference: I would think the important is to have code that with good design,  and maintainable, and UT (which adds line for sure).

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] New Plugin: e..." ]
>I am not sure why the number of lines make a difference: I would think the important is to have code that with good design,  and maintainable, and UT (which adds line for sure).

It's not CLOC, but if duplicated, we need something to keep things in sync 
(automatically)

/Miek

--
Miek Gieben

fturib:
I do not find the meaning for CLOC (https://acronyms.thefreedictionary.com/CLOC). Can you translate ?

> but if duplicated, we need something to keep things in sync

I guess we do not want to duplicate but rather have a common library,  no ?
and the common part, packaged as a lib, is moved to plugin/pkg ... Is it not the purpose of this folder ?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] New Plugin: e..." ]
>I do not find the meaning for CLOC (https://acronyms.thefreedictionary.com/CLOC). Can you translate ?
>
>> but if duplicated, we need something to keep things in sync
>
>I guess we do not want to duplicate but rather have a common library,  no ?
>and the common part, packaged as a lib, is moved to plugin/pkg ... Is it not the purpose of this folder ?

yes, but everything within limits of course; if the entire plugin lives in pkg/ 
you have to ask what you're doing.

Seems doing this in the kubernets plugins seems more sensible?
(don't know what I said before, but 3000 cloc for a little feature seems to be 
pushing it)

PR #2342, opened 2018-11-27
pkg/up: implement backoff

miekg:
Every 2nd failure we double the interval until we hit 4 * interval. This
to have some sort of backoff, esp when a large cluster of coredns shares
an upstream (original intent of up package) they will hammer the
upstream. This put some back pressure on that.

### DIFF
--- a/plugin/pkg/up/up.go
+++ b/plugin/pkg/up/up.go
@@ -43,7 +43,7 @@ func (p *Probe) Do(f Func) {
 			}
 			time.Sleep(interval)
 			if i%2 == 0 && i < 4 { // 4 is 2 doubles, so no need to increase anymore - this is *also* checked in double()
-				p.Double()
+				p.double()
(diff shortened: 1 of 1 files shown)


plugin/pkg/up/up.go
```diff
@@ -34,17 +36,22 @@ func (p *Probe) Do(f Func) {
...
 	// we return from the goroutine and we can accept another Func to run.
 	go func() {
+		i := 1
 		for {
 			if err := f(); err == nil {
 				break
 			}
 			time.Sleep(interval)
+			if i%2 == 0 && i < 4 { // 4 is 2 doubles, so no need to increase anymore - this is *also* checked in double()
+				p.Double()
```
chrisohaver:
```suggestion
				p.double()
```

miekg:
thanks!

Issue #2327, opened 2018-11-20
CoreDNS caches out-of-date k8s dns results 


wavezhang:
CoreDNS version：
 
`1.1.3
`

Corefile:
```
    .:53 {
        errors
        health
        kubernetes cluster.local in-addr.arpa ip6.arpa {
           pods insecure
           transfer 127.0.0.1
           upstream
           fallthrough in-addr.arpa ip6.arpa
        }
        prometheus :9153
        proxy . /etc/resolv.conf
        cache 30
    }
```
I'm using coredns as dns service for k8s, when I recreate svc/pod or pod,  coredns cache plugin will cache NXDOMAIN reply even though svc/pod is alreay created.

wavezhang:
A possible solution is， CoreDNS cache plugin use k8s api to watch svc/pod changes, when changes detected, remove corresponds cache items if exists.

erikandLambo:
I don't know anything your talking about but can I help somehow

wavezhang:
@yongtang

chrisohaver:
Kubernetes plugin keeps it's own local store of objects that are updated when the objects change (in a sense, it keeps its own cache).

So, I think you can achieve what you want by disabling the cache plugin for kubernetes entries (i.e. `cluster.local` zone).  You can do this by splitting the config into two sections, and not including the cache plugin for the `cluster.local` zone... something like...

```
    cluster.local:53 in-addr.arpa:53 ip6.arpa:53 {
        errors
        kubernetes . {
           pods insecure
           transfer 127.0.0.1
           upstream
        }
    }
    
    .:53 {
        errors
        proxy . /etc/resolv.conf
        cache 30
        prometheus :9153
        health
    }
```

chrisohaver:
Actually - I think the example I gave wont work for reverse lookups of IPs that are not in the cluster.

yongtang:
I think a separate channel monitoring the changes in the source k8s api might help in this case.

chrisohaver:
> I think a separate channel monitoring the changes in the source k8s api might help in this case.

Since kubernetes plugin keeps its own cache - there is little/no benefit of also keeping it in the cache plugin.  This plays out in performance testing.  So, IMO, disabling caching for the k8s zones is the simplest most practical approach.

chrisohaver:
[...]

PR #2348, opened 2018-11-29
plugin/kubernetes: Allow 0 TTL

chrisohaver:

Throwing this out there as a possible solution for disabling cache for just kubernetes plugin created records.  Allowing 0 TTL should prevent kubernetes records from going into the cache.

The idea is that this could be used by folks who don't want any k8s records to persist in cache after they become invalid.  Operating with no cache for k8s plugin is not a big performance hit, since k8s objects are already  "cached" in a sense in the object store.

If a 0 TTL is problematic - we could instead extend _cache_ plugin with an option to exclude a list of domains.

Splitting the config up into two server stanzas, one without cache, is problematic for the reasons outlined in #2327 .

kubernetes/kubernetes#71308
#2327 

### DIFF
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -92,7 +92,7 @@ kubernetes [ZONES...] {
   will resolve External Services against itself. **ADDRESS** can be an IP, an IP:port, or a path
   to a file structured like resolv.conf.
 * `ttl` allows you to set a custom TTL for responses. The default (and minimum allowed) is
-  5 seconds, while the maximum is capped at 3600 seconds.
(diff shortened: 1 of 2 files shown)


chrisohaver:
e.g. ...

```
. {
      ttl 0
    }
    cache {
      success 10000 30 0
      denial 10000 30 0
    }
    forward . 8.8.8.8
}

```

yongtang (approved):
LGTM

chrisohaver:
@miekg, do you see any pitfalls by permitting TTL 0?

miekg:
Not sure. Should be ok in this context, I think.

PR #2350, opened 2018-11-29
Remove benchmark from travis

miekg:
We're not doing anything with the data so stop doing it. Also makes
travis 20% faster.


yongtang (approved):
LGTM

Issue #2345, opened 2018-11-28
plugin/route53: can't handle wildcard names properly


dilyevsky:
I noticed names like `*.example.com` don't work even though they are supported by the underlying plugin code (`file` plugin). Turns out AWS api actually returns stars double-escaped. Here a snippet from aws cli output:

```
 {
            "ResourceRecords": [
                {
                    "Value": "docker.example.com"
                }
            ], 
            "Type": "CNAME", 
            "Name": "\\052.docker.example.com.", 
            "TTL": 300
        }, 
```

Need to look out for this and un-escape the star.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] [route53]: can't ..." ]
>I noticed names like `*.example.com` don't work even though they are supported by the underlying plugin code (`file` plugin). Turns out AWS api actually returns stars double-escaped. Here a snippet from aws cli output:
>
>```
> {
>            "ResourceRecords": [
>                {
>                    "Value": "docker.example.com"
>                }
>            ],
>            "Type": "CNAME",
>            "Name": "\\052.example.com.",
>            "TTL": 300
>        },
>```
>
>Need to look out for this and un-escape the star.

Wow. Does Amazon actually claim this is supported? wildcard are tricky no matter 
what backend

/Miek

--
Miek Gieben

dilyevsky:
Yes but only if asterisk is the leftmost label of a record - https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DomainNameFormat.html#domain-name-format-asterisk

> wildcard are tricky no matter what backend

Yeah they are pretty messy. Luckily r53 is limited enough that it should be covered by existing code in `plugin/file`.

PR #2352, opened 2018-11-29
[plugin/route53] Support wildcards and other escaped chars.

dilyevsky:
Adds support for escaped octal sequences returned by AWS API. Follows RFC 1035 (only `a-z`, `0-9`, `-` are allowed) with one exception - `*` to allow for wildcard records.

Example usage (w/ Terraform RR definition):
```terraform
resource "aws_route53_record" "star_www_example_com" {
  name     = "*.www.example.com"
  type     = "CNAME"
  ttl      = "300"
  records  = ["www.example.com"]
}
```

```shell
$ dig a.www.example.com
...
;; QUESTION SECTION:
;a.www.example.com.		IN	A

;; ANSWER SECTION:
a.www.example.com.	2	IN	CNAME	www.example.com.
www.example.com.	24	IN	A	1.2.3.4
...
```
 
Fixes Issue #2345.

### DIFF
--- a/plugin/route53/route53.go
+++ b/plugin/route53/route53.go
@@ -4,7 +4,10 @@ package route53
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


dilyevsky:
also +r=@miekg. I'm thinking not to bother with rfc4343-style extensions (support all possible octects within a label) . Is there general guidance around that in Coredns?

plugin/route53/route53.go
```diff
@@ -140,10 +143,72 @@ func (h *Route53) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
...
+// "If the domain name includes any characters other than a to z, 0 to 9, -
+// (hyphen), or _ (underscore), Route 53 API actions return the characters as
+// escape codes."
+//
+// For our purposes (and with respect to RFC 1035), we'll fish for a-z, 0-9,
+// '-', '.' and '*' as the leftmost character (for wildcards) and throw error
+// for everything else.
+//
+// Example:
+//   `\\052.example.com.` -> `*.example.com`
```
miekg, line 162: this should be supported by miekg/dns transparantly, i.e you can just carry this as-is and things should work. This is valid DNS presentation format.

dilyevsky: hm, i could've sworn I tried to pass this in a test first and it didn't work. Maybe just single `\`?

dilyevsky: Nope. Didn't work with neither `\\` nor `\` when I short-circuit this func to just return argument.

miekg: you're right this actually does the wrong thing:
~~~
package main

import "github.com/miekg/dns"

func main() {
	rr, _ := dns.NewRR(`\052.example.com. IN A 192.168.1.1`)
	println(rr.String())
}
~~~
*I think*

plugin/route53/route53.go
```diff
@@ -140,10 +143,72 @@ func (h *Route53) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
...
+	var out string
+	for {
+		i := strings.Index(s, escapeSeq)
+		if i < 0 {
+			return out + s, nil
+		}
+
+		out += s[:i]
+
+		n, err := strconv.ParseInt(s[i+len(escapeSeq):i+len(escapeSeq)+3], 8, 8)
```
tmthrgd: You haven't checked the length here, this will panic if the `escapeSeq` occurs at the end of the string.

tmthrgd (commented):
Just a passing observation.

miekg:
was/is? this ready to be merged or are you waiting for more feedback?

dilyevsky:
Sorry, I've been overwhelmed with some other stuff. I need to confirm/fix a bug that @tmthrgd pointed out (thanks). Will update this weekend.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [plugin/route..." ]
>Sorry, I've been overwhelmed with some other stuff. I need to confirm/fix a bug that @tmthrgd pointed out (thanks). Will update this weekend.

no problem. Thanks for working on this.

dilyevsky:
Fixed issue that @tmthrgd pointed out (thanks!) and ascii code points were all messed up. Added tests. PTAL.

plugin/route53/route53.go
```diff
@@ -140,10 +143,79 @@ func (h *Route53) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
...
+// escape codes."
+//
+// For our purposes (and with respect to RFC 1035), we'll fish for a-z, 0-9,
+// '-', '.' and '*' as the leftmost character (for wildcards) and throw error
+// for everything else.
+//
+// Example:
+//   `\\052.example.com.` -> `*.example.com`
+//   `\\137.example.com.` -> error ('_' is not valid)
+func maybeUnescape(s string) (string, error) {
```
miekg, line 164: you should be able to lift the code out from the dns/msg.go in the packDomainName function which does something similar, but works on a []byte.

dilyevsky: Hm I took a quick pass - seems not a ton I could use as-is but `dddStringToByte` is kind of cool though - I think I'm going to plug modified version (`oooStringTobyte`) in place of `ParseInt`. Lmk.

miekg: ok, we may be able to make something more generic and useful in miekg/dns

plugin/route53/route53.go
```diff
@@ -140,10 +143,79 @@ func (h *Route53) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
...
+		// Parse `\\xxx` in base 8 (2nd arg) and attempt to fit into
+		// 8-bit result (3rd arg).
+		n, err := strconv.ParseInt(s[li:ri], 8, 8)
+		if err != nil {
+			return "", fmt.Errorf("invalid escape sequence: '%s%s'", escapeSeq, s[li:ri])
+		}
+
+		r := rune(n)
+		switch {
+		case r >= rune('a') && r <= rune('z'): // Route53 converts everything to lowercase.
```
miekg, line 187: are you dealing with `rune` or `byte` here?

miekg:
one nit question, otherwise lgtm

Issue #2349, opened 2018-11-29
Kubernetes Expose External Load Balancer IPs


chrisohaver:
Issue to collaboratively discuss possible implementations for Exposing External Load Balancer IPs in K8s.

related feature requests:
#1851
kubernetes/dns#242

chrisohaver:
**1) Separate plugin**:  make it a Separate plugin, and move code duplicated/shared with kubernetes plugin to shared pkg. (this approach is rejected, see #2337)

**2) Minimal lines of code**: Don't do it as a new plugin, add it as an option to the existing kubernetes plugin.  see example in https://github.com/coredns/coredns/issues/1851#issuecomment-431501196

**3) inter-plugin dependent**: make it a separate plugin, that directly includes parts of the existing kubernetes plugin.  shared code stays in kubernetes dir instead of moving to pkg.  would probably be similar in size to 1, just organized differently.

**4) Fake plugin approach**:  add the functionality to the existing k8s plugin, but expose the option via another plugin that does nothing except store options that the existing k8s plugin references.  In think this is ugly as hell, but it does present itself to the user as a separate plugin, and has minimal lines of code.

Other ideas?

chrisohaver:
@johnbelamaric @miekg

chrisohaver:
FWIW, I prefer 2, doing this in the kubernetes plugin. This was my original inclination.  It's much simpler.

miekg:
Yeah. I think your right in that assessment.

chrisohaver:
> Yeah. I think your right in that assessment.

Cool, @johnbelamaric if you are OK with it not being a separate plugin, then I can whip this out very quickly, probably in time for kube-con.

johnbelamaric:
Yes, I am OK with that. I would expect then that there would be two
recommended deployment models:

1) All-in-one: Same CoreDNS instances handle cluster DNS that handle
external resolution.
2) Segregated: Run a separate CoreDNS for the external domains, and don't
include the cluster domain in the zones list for the k8s plugin.

Does #2 work? I mean, do we make assumptions about the zones we are
handling being cluster domain?

chrisohaver:
both should work.  The existing code can be adapted.
I plan to add a single option called external... `external zone [zone...]`
It will define which zones will be used for exposing external IPs.
Those zones must be in the plugin's defined zones (or server block zones).
If all non-reverse zones are listed as external, then no internal records are exposed.

example: Expose only external (i.e. segregated)

```
[...]

PR #2356, opened 2018-11-30
plugin/kubernetes: Expose Load Balancer/External IPs as records

chrisohaver:

Adds option to expose Services with external Load Balancer IPs (and External IPs) in the form:

`service-name.namespace.zone`

where zone is a zone defined as an external zone in the options.  e.g.

```
kubernetes cluster.local myzone.com in-addr.arpa ip6.arpa {
    external myzone.com
}
```

#1851
kubernetes/dns#242
#2349

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -442,7 +442,7 @@ func newAddress(s msg.Service, name string, ip net.IP, what uint16) dns.RR {
 	return &dns.AAAA{Hdr: hdr, AAAA: ip}
 }
 
-// checkForApex checks the spcecial apex.dns directory for records that will be returned as A or AAAA.
+// checkForApex checks the special apex.dns directory for records that will be returned as A or AAAA.
 func checkForApex(b ServiceBackend, zone string, state request.Request, opt Options) ([]msg.Service, error) {
 	if state.Name() != zone {
 		return b.Services(state, false, opt)
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -374,6 +374,7 @@ func NS(b ServiceBackend, zone string, state request.Request, opt Options) (reco
 	// ... and reset
 	state.Req.Question[0].Name = old
 
+	hostSeen := make(map[string]bool)
 	for _, serv := range services {
 		what, ip := serv.HostType()
 		switch what {
@@ -382,7 +383,10 @@ func NS(b ServiceBackend, zone string, state request.Request, opt Options) (reco
 
 		case dns.TypeA, dns.TypeAAAA:
 			serv.Host = msg.Domain(serv.Key)
-			records = append(records, serv.NewNS(state.QName()))
+			if !hostSeen[serv.Host] {
+				records = append(records, serv.NewNS(state.QName()))
+				hostSeen[serv.Host] = true
+			}
 			extra = append(extra, newAddress(serv, serv.Host, ip, what))
 		}
 	}
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -45,6 +45,7 @@ kubernetes [ZONES...] {
     transfer to ADDRESS...
     fallthrough [ZONES...]
     ignore empty_service
+    external ZONES...
 }
 ```
 
@@ -109,6 +110,10 @@ kubernetes [ZONES...] {
 * `ignore empty_service` return NXDOMAIN for services without any ready endpoint addresses (e.g., ready pods).
   This allows the querying pod to continue searching for the service in the search path.
   The search path could, for example, include another Kubernetes cluster.
+* `external` will synthesize records in the listed **[ZONES...]** for all Kubernetes services with Load Balancer IPs.
+  All **[ZONES...]** must exist in the plugin's zones or server block.  If all non-reverse zones in the
+  plugin's zones or server block are listed, then no internal service records will be synthesized. Records are in the
+  form of `service-name.namespace.zone.`
 
 ## Health
 
@@ -149,6 +154,27 @@ Connect to Kubernetes with CoreDNS running outside the cluster:
 kubernetes cluster.local {
     endpoint https://k8s-endpoint:8443
     tls cert key cacert
+}
+~~~
+
+## Exposing Load Balancer IPs
+
+Expose internal service cluster ips in the `cluster.local` zone, and external load
+balancer ips in the `my.zone.com. zone.
+
+~~~ txt
+kubernetes cluster.local my.zone.com {
+    external my.zone.com
+}
+~~~
+
+Connect to Kubernetes with CoreDNS running outside the cluster, and only expose external IPs in the zone `my.zone.com`:
+
+~~~ txt
+kubernetes my.zone.com {
+    endpoint https://k8s-endpoint:8443
+    tls cert key cacert
+    external my.zone.com
 }
 ~~~
 
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -110,8 +110,8 @@ kubernetes [ZONES...] {
 * `ignore empty_service` return NXDOMAIN for services without any ready endpoint addresses (e.g., ready pods).
   This allows the querying pod to continue searching for the service in the search path.
   The search path could, for example, include another Kubernetes cluster.
-* `external` will synthesize records in the listed **[ZONES...]** for all Kubernetes services with Load Balancer IPs.
-  All **[ZONES...]** must exist in the plugin's zones or server block.  If all non-reverse zones in the
+* `external` **ZONES...** will synthesize records in the listed **ZONES...** for all Kubernetes services with Load Balancer IPs.
+  All **ZONES...** listed here must also exist in the plugin's zones or server block.  If all non-reverse zones in the
   plugin's zones or server block are listed, then no internal service records will be synthesized. Records are in the
   form of `service-name.namespace.zone.`
 
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -159,8 +159,8 @@ kubernetes cluster.local {
 
 ## Exposing Load Balancer IPs
 
-Expose internal service cluster ips in the `cluster.local` zone, and external load
-balancer ips in the `my.zone.com. zone.
+Expose internal service cluster IPs in the `cluster.local` zone, and external load
+balancer IPs in the `my.zone.com. zone.
 
 ~~~ txt
 kubernetes cluster.local my.zone.com {
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -110,7 +110,7 @@ kubernetes [ZONES...] {
 * `ignore empty_service` return NXDOMAIN for services without any ready endpoint addresses (e.g., ready pods).
   This allows the querying pod to continue searching for the service in the search path.
   The search path could, for example, include another Kubernetes cluster.
-* `external` **ZONES...** will synthesize records in the listed **ZONES...** for all Kubernetes services with Load Balancer IPs.
+* `external` **ZONES...** will synthesize records in the listed **ZONES...** for all Kubernetes services with Load Balancer IPs and External IPs.
   All **ZONES...** listed here must also exist in the plugin's zones or server block.  If all non-reverse zones in the
   plugin's zones or server block are listed, then no internal service records will be synthesized. Records are in the
   form of `service-name.namespace.zone.`
@@ -157,10 +157,9 @@ kubernetes cluster.local {
 }
 ~~~
 
-## Exposing Load Balancer IPs
+## Exposing Load Balancer and External IPs
 
-Expose internal service cluster IPs in the `cluster.local` zone, and external load
-balancer IPs in the `my.zone.com. zone.
+Expose internal service cluster IPs in the `cluster.local` zone, and external IPs in the `my.zone.com. zone.
 
 ~~~ txt
 kubernetes cluster.local my.zone.com {
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -25,6 +25,9 @@ const (
 	svcIPIndex            = "ServiceIP"
 	epNameNamespaceIndex  = "EndpointNameNamespace"
 	epIPIndex             = "EndpointsIP"
+	exposeAll             = 0
+	exposeCluster         = 1
+	exposeExternal        = 2
 )
 
 type dnsController interface {
@@ -97,6 +100,7 @@ type dnsControlOpts struct {
 
 	zones            []string
 	endpointNameMode bool
+	expose           int
 }
 
 // newDNSController creates a controller for CoreDNS.
@@ -110,6 +114,16 @@ func newdnsController(kubeClient kubernetes.Interface, opts dnsControlOpts) *dns
 		endpointNameMode: opts.endpointNameMode,
 	}
 
+	var svcIPIndexer cache.IndexFunc
+	switch opts.expose {
+	case exposeCluster:
+		svcIPIndexer = svcClusterIPIndexFunc
+	case exposeExternal:
+		svcIPIndexer = svcExternalIPsIndexFunc
+	case exposeAll:
+		svcIPIndexer = svcAllIPsIndexFunc
+	}
+
 	dns.svcLister, dns.svcController = object.NewIndexerInformer(
 		&cache.ListWatch{
 			ListFunc:  serviceListFunc(dns.client, api.NamespaceAll, dns.selector),
@@ -118,7 +132,7 @@ func newdnsController(kubeClient kubernetes.Interface, opts dnsControlOpts) *dns
 		&api.Service{},
 		opts.resyncPeriod,
 		cache.ResourceEventHandlerFuncs{AddFunc: dns.Add, UpdateFunc: dns.Update, DeleteFunc: dns.Delete},
-		cache.Indexers{svcNameNamespaceIndex: svcNameNamespaceIndexFunc, svcIPIndex: svcIPIndexFunc},
+		cache.Indexers{svcNameNamespaceIndex: svcNameNamespaceIndexFunc, svcIPIndex: svcIPIndexer},
 		object.ToService,
 	)
 
@@ -167,12 +181,28 @@ func podIPIndexFunc(obj interface{}) ([]string, error) {
 	return []string{p.PodIP}, nil
 }
 
-func svcIPIndexFunc(obj interface{}) ([]string, error) {
+func svcAllIPsIndexFunc(obj interface{}) ([]string, error) {
+	svc, ok := obj.(*object.Service)
+	if !ok {
+		return nil, errObj
+	}
+	return append(svc.ExternalIPs, svc.ClusterIP), nil
+}
+
+func svcClusterIPIndexFunc(obj interface{}) ([]string, error) {
 	svc, ok := obj.(*object.Service)
 	if !ok {
 		return nil, errObj
 	}
 	return []string{svc.ClusterIP}, nil
+}
+
+func svcExternalIPsIndexFunc(obj interface{}) ([]string, error) {
+	svc, ok := obj.(*object.Service)
+	if !ok {
+		return nil, errObj
+	}
+	return svc.ExternalIPs, nil
 }
 
 func svcNameNamespaceIndexFunc(obj interface{}) ([]string, error) {
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -93,7 +93,7 @@ var (
 // Services implements the ServiceBackend interface.
 func (k *Kubernetes) Services(state request.Request, exact bool, opt plugin.Options) (svcs []msg.Service, err error) {
 
-	external := k.externalZone(state.Name())
+	external := "" != plugin.Zones(k.externalZones).Matches(state.Name())
 
 	// We're looking again at types, which we've already done in ServeDNS, but there are some types k8s just can't answer.
 	switch state.QType() {
@@ -117,21 +117,22 @@ func (k *Kubernetes) Services(state request.Request, exact bool, opt plugin.Opti
 
 	case dns.TypeNS:
 		// We can only get here if the qname equals the zone, see ServeDNS in handler.go.
-		if external {
-			svc := msg.Service{Host: state.LocalIP(), Key: msg.Path(state.QName(), "coredns")}
-			return []msg.Service{svc}, nil
-		}
-		ns := k.nsAddr()
-		svc := msg.Service{Host: ns.A.String(), Key: msg.Path(state.QName(), "coredns")}
+		ns := k.nsAddr(external)
+		svc := msg.Service{Host: ns[0].A.String(), Key: msg.Path(state.QName(), "coredns")}
 		return []msg.Service{svc}, nil
 	}
 
-	if state.QType() == dns.TypeA && isDefaultNS(state.Name(), state.Zone) && !external {
-		// If this is an A request for "ns.dns", respond with a "fake" record for coredns.
-		// SOA records always use this hardcoded name
-		ns := k.nsAddr()
-		svc := msg.Service{Host: ns.A.String(), Key: msg.Path(state.QName(), "coredns")}
-		return []msg.Service{svc}, nil
+	if state.QType() == dns.TypeA && isDefaultNS(state.Name(), state.Zone) {
+		// create an A record for the NS target
+		nss := k.nsAddr(external)
+		var svcs []msg.Service
+		for _, ns := range nss {
+			if ns.A.String() == "0.0.0.0" {
+				continue
+			}
+			svcs = append(svcs, msg.Service{Host: ns.A.String(), Key: msg.Path(state.QName(), "coredns")})
+		}
+		return svcs, nil
 	}
 
 	s, e := k.Records(state, false)
@@ -290,18 +291,9 @@ func (k *Kubernetes) InitKubeCache() (err error) {
 // Records looks up services in kubernetes.
 func (k *Kubernetes) Records(state request.Request, exact bool) ([]msg.Service, error) {
 
-	external := k.externalZone(state.Name())
+	external := "" != plugin.Zones(k.externalZones).Matches(state.Name())
 
-	var (
-		r recordRequest
-		e error
-	)
-
-	if external {
-		r, e = parseRequestExternal(state)
-	} else {
-		r, e = parseRequest(state)
-	}
+	r, e := parseRequest(state, external)
 
 	if e != nil {
 		return nil, e
@@ -323,8 +315,20 @@ func (k *Kubernetes) Records(state request.Request, exact bool) ([]msg.Service, 
 		return pods, err
 	}
 
-	services, err := k.findServices(r, state.Zone, external)
-	return services, err
+	// handle empty service name
+	if r.service == "" {
+		if k.namespace(r.namespace) || wildcard(r.namespace) {
+			// NODATA
+			return nil, nil
+		}
+		// NXDOMAIN
+		return nil, errNoItems
+	}
+
+	if external {
+		return k.findServicesExternal(r, state.Zone)
+	}
+	return k.findServices(r, state.Zone)
 }
 
 // serviceFQDN returns the k8s cluster dns spec service FQDN for the service (or endpoint) object.
@@ -441,8 +445,8 @@ func (k *Kubernetes) findPods(r recordRequest, zone string) (pods []msg.Service,
 }
 
 // findServices returns the services matching r from the cache.
-func (k *Kubernetes) findServices(r recordRequest, zone string, external bool) (services []msg.Service, err error) {
-	if !external && k.opts.expose == exposeExternal {
+func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.Service, err error) {
+	if k.opts.expose == exposeExternal {
 		return nil, nil
 	}
 	zonePath := msg.Path(zone, "coredns")
@@ -460,16 +464,6 @@ func (k *Kubernetes) findServices(r recordRequest, zone string, external bool) (
 		endpointsList     []*object.Endpoints
 		serviceList       []*object.Service
 	)
-
-	// handle empty service name
-	if r.service == "" {
-		if k.namespace(r.namespace) || wildcard(r.namespace) {
-			// NODATA
-			return nil, nil
-		}
-		// NXDOMAIN
-		return nil, errNoItems
-	}
 
 	if wildcard(r.service) || wildcard(r.namespace) {
 		serviceList = k.APIConn.ServiceList()
@@ -491,7 +485,7 @@ func (k *Kubernetes) findServices(r recordRequest, zone string, external bool) (
 			continue
 		}
 
-		if k.opts.ignoreEmptyService && svc.ClusterIP != api.ClusterIPNone && !external {
+		if k.opts.ignoreEmptyService && svc.ClusterIP != api.ClusterIPNone {
 			// serve NXDOMAIN if no endpoint is able to answer
 			podsCount := 0
 			for _, ep := range endpointsListFunc() {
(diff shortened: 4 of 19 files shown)


chrisohaver:
Bulk of functionality is working...

Still to do:
- ~~zone transfers~~
- ~~unit tests~~
- ~~ci tests~~

plugin/kubernetes/controller.go
```diff
@@ -167,14 +181,30 @@ func podIPIndexFunc(obj interface{}) ([]string, error) {
 	return []string{p.PodIP}, nil
 }
 
-func svcIPIndexFunc(obj interface{}) ([]string, error) {
+func svcAllIPsIndexFunc(obj interface{}) ([]string, error) {
+	svc, ok := obj.(*object.Service)
+	if !ok {
+		return nil, errObj
+	}
+	return append(svc.ExternalIPs, svc.ClusterIP), nil
```
johnbelamaric, line 189: How does the indexing work? Does this create two index entries, one for each IP?

plugin/kubernetes/kubernetes.go
```diff
@@ -110,12 +118,16 @@ func (k *Kubernetes) Services(state request.Request, exact bool, opt plugin.Opti
 
 	case dns.TypeNS:
 		// We can only get here if the qname equals the zone, see ServeDNS in handler.go.
+		if external {
+			svc := msg.Service{Host: state.LocalIP(), Key: msg.Path(state.QName(), "coredns")}
```
johnbelamaric: so, this is interesting. what does the NS record look like for the external zone?

plugin/kubernetes/kubernetes.go
```diff
@@ -110,12 +118,16 @@ func (k *Kubernetes) Services(state request.Request, exact bool, opt plugin.Opti
...
+			return []msg.Service{svc}, nil
+		}
 		ns := k.nsAddr()
 		svc := msg.Service{Host: ns.A.String(), Key: msg.Path(state.QName(), "coredns")}
 		return []msg.Service{svc}, nil
 	}
 
-	if state.QType() == dns.TypeA && isDefaultNS(state.Name(), state.Zone) {
+	if state.QType() == dns.TypeA && isDefaultNS(state.Name(), state.Zone) && !external {
```
johnbelamaric: do we need some way to get the external IP for CoreDNS? I think we should set this up to support delegation. I think that means that we don't necessarily need that, because the parent zone file should contain the A record for the NS in the delegation. @miekg or @cricketliu can you clarify how that would work? For example, if the external zone is "cluster.example.com":

CoreDNS returns SOA for external domain:

```
cluster.example.com. IN SOA ns.dns.cluster.example.com hostmaster.cluster.example.com (...)
```

example.com contains

```
cluster.example.com. IN NS ns.dns.cluster.example.com
ns.dns.cluster.example.com IN A <coredns-service-external-ip>
```

Does CoreDNS need to have an A record for ns.dns.cluster.example.com, or is the glue record in example.com sufficient?

johnbelamaric (commented):
This will be very cool. A few comments. One question: here, we do this for load balancer services. But what about other ways to expose services? See https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#servicespec-v1-core

In particular, I am thinking about:

- externalIPs
- SRV records for NodePort services

plugin/kubernetes/controller.go
chrisohaver: Yes

plugin/kubernetes/kubernetes.go
chrisohaver: `ns.dns.zone`

johnbelamaric: Ok. but does it also create an A record? I don't get what LocalIP() is doing here?

chrisohaver:
> In particular, I am thinking about:
> * externalIPs

Regarding "External IPs",  I'm confused about the the exact relationship with load balancers ips... the docs say that load balancers are an example of external ips... are they the only example?  When I create load balancers in k8s, external IPs field remains blank on the service.  So I was thinking External IPs field may be used by some cloud providers to allow the user to declare the LB IP instead of it being dynamically provisioned.  The only examples of "external ips" I could find involve load balancers.  if you could point me to a non loadbalancer external IP example that would help tremendously.

 > * SRV records for NodePort services

That would be cool. I think it should be delivered in a separate PR.

johnbelamaric:
Agree on the SRVs for NodePort - separate PR is best.

For externalIPs, they are just managed outside of Kubernetes. So, the load
balancer ones are managed by the cloud controller (or plugin), but these
are just specified by the user. The user has to manually configure any
relevant nodes to accept traffic on those IPs. They also need to control
the pod placement to land on one of those nodes. SRV probably makes sense
here as well, because you may have several. Though, depending on your
routing (Anycast, for example), you could use the same IP on multiple nodes.

If we choose to do this one it should be a separate PR as well. Probably
need more real-world examples.

plugin/kubernetes/kubernetes.go
chrisohaver: We have code already to get the cluster IP of coredns in cluster.
If we need pull the external ip, I can adapt that code.
Of course, that assuming the in-cluster coreDNS _has_ an external ip... which is does not by default.

I was thinking that for most cases, if someone wants to expose external records, they'd do so in an external instance of coredns, not serve them from the in-cluster replicas... but i could be wrong.

chrisohaver:
> For externalIPs, they are just managed outside of Kubernetes. ...The user has to manually configure any relevant nodes to accept traffic on those IPs.

OK, makes sense.  In that case it's very easy to add.  I will do that in this PR, and update docs.

plugin/kubernetes/kubernetes.go
chrisohaver: yes, it creates an A record too...  I thought it was supposed to be the ip address of coredns itself.

for the in-cluster zone, this is the ip address of the coredns service (e.g. `10.96.0.10`).  so you have (i think)...
```
ns.dns.cluster.local IN NS kube-dns.kube-system.svc.cluster.local
kube-dns.kube-system.svc.cluster.local IN A 10.96.0.10
```

For the external zone, i think it should be something like ...
```
ns.dns.cluster.myzone.com IN NS kube-dns.kube-system.myzone.com
kube-dns.kube-system.myzone.com IN A 1.2.3.4
```

chrisohaver: isn't this the same question as below? "do we need some way to get the external IP for CoreDNS?"

plugin/kubernetes/kubernetes.go
chrisohaver: Looking a second time, I see some problems with the in-cluster kubernetes NS record code that need fixing ...

chrisohaver: Also, for external zones, `ns.dns.zone.` collides with a service named `ns` in namespace `dns`.

This whole mess is related to the backend hard coding `ns.dns.zone.` as the nameserver portion of an `SOA`  record.

miekg:
It needs to be somewhere..any suggestions?

chrisohaver:
> It needs to be somewhere..any suggestions?

I think in the case of kubernetes, it could point directly to the dns service... `kube-dns.kube-system.zone.`

```
zone. IN NS kube-dns.kube-system.zone.
kube-dns.kube-system.zone. IN A 1.2.3.4
```
This means the backend would need to get the NS name is from the plugin when generating an `SOA` ...

miekg:
We could easily change that and use the kube-system name space

cricketliu:
On Nov 30, 2018, at 12:01 PM, John Belamaric <notifications@github.com<mailto:notifications@github.com>> wrote:


do we need some way to get the external IP for CoreDNS? I think we should set this up to support delegation. I think that means that we don't necessarily need that, because the parent zone file should contain the A record for the NS in the delegation. @miekg<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_miekg&d=DwMCaQ&c=CWsWqoUynJrLESJQduOsxQ&r=uA3zJBZARmPCCWmTYBLk07UikWEyE9AKvAmI_Idoujg&m=WKiJZ8IOBtMAOh043-2de8vJoRzylFalwrXka2IK18s&s=LwRlYFcsgbHhWg8s5KdASaVmRmzaI_BjGsUKVm9ptTs&e=> or @cricketliu<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_cricketliu&d=DwMCaQ&c=CWsWqoUynJrLESJQduOsxQ&r=uA3zJBZARmPCCWmTYBLk07UikWEyE9AKvAmI_Idoujg&m=WKiJZ8IOBtMAOh043-2de8vJoRzylFalwrXka2IK18s&s=mxrOv80aG2GU_J9k-pW8yEVfstE76AwJWcM_UJiZicc&e=> can you clarify how that would work? For example, if the external zone is "cluster.example.com<http://cluster.example.com/>":

CoreDNS returns SOA for external domain:

cluster.example.com<http://cluster.example.com/>. IN SOA ns.dns.cluster.example.com<http://ns.dns.cluster.example.com/> hostmaster.cluster.example.com<http://hostmaster.cluster.example.com/> (...)


example.com<http://example.com/> contains

cluster.example.com<http://cluster.example.com/>. IN NS ns.dns.cluster.example.com<http://ns.dns.cluster.example.com/>
ns.dns.cluster.example.com<http://ns.dns.cluster.example.com/> IN A <coredns-service-external-ip>


Does CoreDNS need to have an A record for ns.dns.cluster.example.com<http://ns.dns.cluster.example.com/>, or is the glue record in example.com<http://example.com/> sufficient?

If CoreDNS is authoritative for cluster.example.com<http://cluster.example.com>, and ns.dns.cluster.example.com<http://ns.dns.cluster.example.com> is in that zone, then yes, CoreDNS needs an A RR for ns.dns.cluster.example.com<http://ns.dns.cluster.example.com>, too.  In other words, the A RR needs to appear twice:  Once in example.com<http://example.com> as glue, and once in the authoritative zone it's a part of.

cricket

cricketliu:
On Nov 30, 2018, at 12:36 PM, chrisohaver <notifications@github.com<mailto:notifications@github.com>> wrote:


yes, it creates an A record too... I thought it was supposed to be the ip address of coredns itself.

for the in-cluster zone, this is the ip address of the coredns service (e.g. 10.96.0.10). so you have (i think)...

ns.dns.cluster.local IN NS kube-dns.kube-system.svc.cluster.local
kube-dns.kube-system.svc.cluster.local IN A 10.96.0.10


For the external zone, i think it should be something like ...

ns.dns.cluster.myzone.com<http://ns.dns.cluster.myzone.com/> IN NS kube-dns.kube-system.myzone.com<http://kube-dns.kube-system.myzone.com/>
kube-dns.kube-system.myzone.com<http://kube-dns.kube-system.myzone.com/> IN A 1.2.3.4


We should probably test whether or not the A RR is necessary to complete the delegation.  The test is straightforward:  An A RR is needed as "glue" iff the name of the name server ends in the name of the subdomain.   So in this case, kube-dns.kube-system.myzone.com<http://kube-dns.kube-system.myzone.com> doesn't end in ns.dns.cluster.myzone.com<http://ns.dns.cluster.myzone.com>, so no A RR is needed.

cricket

plugin/kubernetes/README.md
```diff
@@ -45,6 +45,7 @@ kubernetes [ZONES...] {
     transfer to ADDRESS...
     fallthrough [ZONES...]
     ignore empty_service
+    external ZONES...
```
miekg, line 48: is ZONES optional?

plugin/kubernetes/README.md
```diff
@@ -152,6 +157,27 @@ kubernetes cluster.local {
 }
 ~~~
 
+## Exposing Load Balancer IPs
+
+Expose internal service cluster ips in the `cluster.local` zone, and external load
```
miekg: s/ips/IPs/g

plugin/kubernetes/controller.go
miekg: why is this an append?

plugin/kubernetes/kubernetes.go
miekg: yes, we need to think about it (probably in a new issue to not muddle this).

plugin/kubernetes/kubernetes.go
```diff
@@ -274,11 +289,24 @@ func (k *Kubernetes) InitKubeCache() (err error) {
 
 // Records looks up services in kubernetes.
 func (k *Kubernetes) Records(state request.Request, exact bool) ([]msg.Service, error) {
-	r, e := parseRequest(state)
+
+	external := k.externalZone(state.Name())
```
miekg: externalZones is a 1 -line function just inline it here (and the other place you call it)

plugin/kubernetes/kubernetes.go
```diff
@@ -274,11 +289,24 @@ func (k *Kubernetes) InitKubeCache() (err error) {
...
+
+	var (
+		r recordRequest
+		e error
+	)
+
+	if external {
+		r, e = parseRequestExternal(state)
```
miekg: Why does this need an almost identical parse function?

plugin/kubernetes/kubernetes.go
```diff
@@ -290,12 +318,12 @@ func (k *Kubernetes) Records(state request.Request, exact bool) ([]msg.Service,
...
 	}
 
-	if r.podOrSvc == Pod {
+	if r.podOrSvc == Pod && !external {
 		pods, err := k.findPods(r, state.Zone)
 		return pods, err
 	}
 
-	services, err := k.findServices(r, state.Zone)
+	services, err := k.findServices(r, state.Zone, external)
```
miekg: The external service search might actually be cleaner in a separate function.

plugin/kubernetes/object/service.go
```diff
@@ -12,6 +12,7 @@ type Service struct {
 	Namespace    string
 	Index        string
 	ClusterIP    string
+	ExternalIPs  []string
```
miekg, line 15: this adds extra memory, have we tested it?

plugin/kubernetes/setup.go
```diff
@@ -287,11 +275,68 @@ func ParseStanza(c *caddy.Controller) (*Kubernetes, error) {
 				continue
 			}
 			return nil, c.ArgErr()
+		case "external":
+			args := c.RemainingArgs()
+			if len(args) > 0 {
```
miekg: please return on len(args) == 0 and dedent the rest of this block

plugin/pkg/dnsutil/reverse.go
```diff
@@ -39,6 +39,12 @@ func IsReverse(name string) int {
 	if strings.HasSuffix(name, IP6arpa) {
 		return 2
 	}
+	if "."+name == IP4arpa {
```
miekg: this creates garbage.

also why?

plugin/kubernetes/setup.go
```diff
@@ -287,11 +275,68 @@ func ParseStanza(c *caddy.Controller) (*Kubernetes, error) {
...
+						if dnsutil.IsReverse(z) > 0 {
+							continue
+						}
+						if z == extZone {
+							found = true
+							break
+						}
+					}
+					if !found {
+						return nil, c.Errf("external zone '%s' is not in plugin/block zone list", c.Val())
```
miekg: none of the plugins have these checks, I'm also fan of garbage in, garbage out, so we should add it here.
As the corefile has all context we can easily write a tools that checks the config, ala named-checkconfig instead of pulling out that code directly into coredns

miekg (changes requested):

plugin/kubernetes/README.md
chrisohaver: No, Zones is not optional.  At least one is required.
You're basically defining which zones will be treated as external zones.

chrisohaver:
Cant reply to original code comment for some buggy reason.

> why is this an append?

Because i want to append a record to a slice.  Is there a better way?

plugin/kubernetes/controller.go
chrisohaver: Because i want to append a record to a slice.  Is there a better way?

plugin/kubernetes/kubernetes.go
chrisohaver: I felt it was better than cluttering the code with IFs...

plugin/kubernetes/kubernetes.go
chrisohaver: That would result in some code duplication.  You OK with that?

plugin/kubernetes/object/service.go
chrisohaver: No, I have not tested it yet. I don't expect it will add much.

plugin/pkg/dnsutil/reverse.go
chrisohaver: As it was, `IsReverse("in-addr.arpa.")` would return `0`, because it does not have the suffix ".in-addr.arpa.".

I'll try to fix this differently.

plugin/kubernetes/object/service.go
johnbelamaric: If it's not too hard we can not keep this if no external zone is set. But maybe we don't know that this deep in.

chrisohaver: What is the memory footprint for an empty slice?

johnbelamaric:
It's a pointer. So I would think 8 bytes typically.

plugin/kubernetes/object/service.go
chrisohaver: For most services, this will be empty.  I think the impact will be very minimal.  IMO, we can optimize later if we need to.

plugin/kubernetes/README.md
```diff
@@ -109,6 +110,10 @@ kubernetes [ZONES...] {
 * `ignore empty_service` return NXDOMAIN for services without any ready endpoint addresses (e.g., ready pods).
   This allows the querying pod to continue searching for the service in the search path.
   The search path could, for example, include another Kubernetes cluster.
+* `external` will synthesize records in the listed **[ZONES...]** for all Kubernetes services with Load Balancer IPs.
```
miekg: which ZONES is this referencing?

plugin/kubernetes/controller.go
miekg: oh, non obvious, can this be done in the caller or don't we control that?

plugin/kubernetes/kubernetes.go
```diff
@@ -110,21 +117,30 @@ func (k *Kubernetes) Services(state request.Request, exact bool, opt plugin.Opti
...
-		// If this is an A request for "ns.dns", respond with a "fake" record for coredns.
-		// SOA records always use this hardcoded name
-		ns := k.nsAddr()
-		svc := msg.Service{Host: ns.A.String(), Key: msg.Path(state.QName(), "coredns")}
-		return []msg.Service{svc}, nil
+		// create an A record for the NS target
+		nss := k.nsAddr(external)
+		var svcs []msg.Service
+		for _, ns := range nss {
+			if ns.A.String() == "0.0.0.0" {
```
miekg: what?

plugin/kubernetes/parse.go
```diff
@@ -57,13 +58,16 @@ func parseRequest(state request.Request) (r recordRequest, err error) {
...
 	}
-	r.podOrSvc = segs[last]
-	if r.podOrSvc != Pod && r.podOrSvc != Svc {
-		return r, errInvalidRequest
-	}
-	last--
-	if last < 0 {
-		return r, nil
+
+	if !external {
```
miekg, line 64: why can't the caller check the return value of this function as-is and then decide if the request is usable. Why does this function need changing?

plugin/pkg/dnsutil/reverse.go
```diff
@@ -33,12 +33,13 @@ func ExtractAddressFromReverse(reverseName string) string {
 // name is in a reverse zone. The returned integer will be 1 for in-addr.arpa. (IPv4)
 // and 2 for ip6.arpa. (IPv6).
 func IsReverse(name string) int {
-	if strings.HasSuffix(name, IP4arpa) {
+	if strings.HasSuffix(name, IP4arpa) || name == IP4arpaDom {
 		return 1
 	}
-	if strings.HasSuffix(name, IP6arpa) {
+	if strings.HasSuffix(name, IP6arpa) || name == IP6arpaDom {
```
miekg: this should be needed.

plugin/kubernetes/controller.go
chrisohaver: I don't think so, this is an indexer func thats called by the k8s client when k8s Services are changed/added/deleted, it expects a single slice of strings. 

An alternative would be to maintain a separate index.  But there's more overhead with a maintaining an additional index.

plugin/pkg/dnsutil/reverse.go
chrisohaver: ?

plugin/kubernetes/parse.go
chrisohaver: Because the request format is different for external vs cluster names.

external names: service-name.namespace.zone.
cluster names: service-name.namespace.svc.zone.

Also, cluster names accept an endpoint segment for A records of service endpoints (pod ips). (endpoint.service-name.namespace.svc.zone).  External names do not (they only expose external IPs).

plugin/kubernetes/README.md
chrisohaver: The local one in the option.  I'll clarify...

plugin/pkg/dnsutil/reverse.go
chrisohaver: Do you mean, "This should _not_ bet needed." ?

This does fix a bug in the existing code, where `in-addr.arpa.` is not identified as a reverse domain.
This can cause the _kubernetes_ plugin to select a reverse domain as its primary domain if `in-addr.arpa` is listed before `cluster.local`. e.g. `kubernetes in-addr.arpa cluster.local`

plugin/kubernetes/kubernetes.go
chrisohaver: Yeah - I should have commented the code here. I am using `0.0.0.0` to mean a null IP (i.e. we don't know the IP). I will refactor the nsAddr() function to make this less ugly.

plugin/kubernetes/parse.go
miekg: ah, that would be a good comment to put here.

plugin/pkg/dnsutil/reverse.go
miekg: yeah, missed *not*, just spotted that.

in-addr.arpa can never be a correct reverse domain in our context.
It most def. should *not* qualify as "true" in `IsReverse`

chrisohaver: Hmm. OK.  I'll create a different function that does what I need, and remove IsReverse(), since it will no longer be used by anything.

chrisohaver: ... actually - i'll create a new function that only checks the domain, and call both.

plugin/kubernetes/controller.go
johnbelamaric:
```
return []string{svc.ExternalIPs, svc.ClusterIP}, nil
```
Is that any better?

johnbelamaric: oh, nevermind. I see svc.ExternalIPs is already a slice

plugin/pkg/dnsutil/reverse.go
chrisohaver: Ive added a new function, and also call that when i'm comparing zones that may be a base reverse domain.

plugin/kubernetes/kubernetes.go
chrisohaver: I added comments, and use nil instead, which is a lot clearer.

plugin/kubernetes/object/service.go
miekg: https://play.golang.org/p/TB_uzDhcwXc
12 bytes, this has to be moved somewhere else

chrisohaver: Well, with 10,000 services max, 12 bytes per service comes to 0.12 MB
Which comes to about a 0.06% increase when comparing to scale results for clusters with ~8600 services (approx 189 MB total mb footprint + 0.12 MB).

Given that, I don't think this is a significant enough of an increase to warrant moving it.

chrisohaver: > with 10,000 services max ...

To clarify, 10,000 services is the maximum number of services allowed in a single Kubernetes cluster (the entire cluster. not per node).

If we consider the worst possible case (not realistic, but technically possible): a single node cluster with the maximum 10,000 services, the proportional increase due to empty 12 bytes slices would be about 0.21% (approx 58 MB total mb footprint + 0.12 MB).

chrisohaver:
Regarding the `NS` records - I have added ci integration tests for them for local and external zones.

The current behavior is to respond with an `NS` record for `ns.dns.<zone>` and also include `A` records in the additional section.  

- For the cluster's local zone, the `A` record is for the Cluster IP of the CoreDNS Service. 
- For the external zone, the additional section includes an `A` record for each External IP of the CoreDNS Service.

The CoreDNS Service is determined by looking the local IP in the Endpoint object store.

plugin/kubernetes/object/service.go
miekg: your almost doubling the default size  of this struct. And given that we had a lengthy  process of getting the memory down we should not add stuff here.

johnbelamaric:
Can we run a test to see how much of an effect this has? It's an empirical
question...

chrisohaver:
> Can we run a test to see how much of an effect this has?

Yes, I'll get one started today.

chrisohaver:
>>Can we run a test to see how much of an effect this has?

>Yes, I'll get one started today.

The 500 node test result shows a 1% increase in memory usage.
(compared to CoreDNS 1.2.4)

miekg:
regarding parseQuery: for a name like `service-name.namespace.zone` you want to know 2 things after you've established it's an external zone
1. the namespace (i.e. first label before the zone)
2. service-name, (i.e. the label before that)

So parse query should not have to be changed it all to make this work

miekg:
> The 500 node test result shows a 1% increase in memory usage.
> (compared to CoreDNS 1.2.4)

Good to know!

How hard is it to *not* change the Service struct?

plugin/kubernetes/controller.go
miekg: the append's return value is not assigned to anything

plugin/kubernetes/object/service.go
miekg: `ExternalIPs:  make([]string, len(svc.Status.LoadBalancer.Ingress)+len(svc.Spec.ExternalIPs))`
will probably somewhat more than 12 bytes

plugin/kubernetes/object/service.go
```diff
@@ -47,6 +49,16 @@ func ToService(obj interface{}) interface{} {
 		copy(s.Ports, svc.Spec.Ports)
 	}
 
+	for i, lb := range svc.Status.LoadBalancer.Ingress {
+		s.ExternalIPs[i] = lb.IP
+	}
+
+	ipCount := len(svc.Status.LoadBalancer.Ingress)
```
miekg, line 56: is `i` of the same value as you just iterate over the slice?

chrisohaver:
> regarding parseQuery: for a name like service-name.namespace.zone you want to know 2 things after you've established it's an external zone
>the namespace (i.e. first label before the zone)
>service-name, (i.e. the label before that)
>So parse query should not have to be changed it all to make this work

the unmodified parse function would puke on "namespace" because it is not "svc" or "pod"...

chrisohaver:
> How hard is it to not change the Service struct?

We could refactor to make all object.Service access through an interface, and make two structs that implement the interface (one with externalIPs and one without).

miekg:
>> How hard is it to not change the Service struct?
>
>We could refactor to make all object.Service access through an interface, and make two structs that implement the interface (one with externalIPs and one without).

hmmm, that would make things slower and can also increase memory.

I still would like to avoid changing this struct, unless it's a feature that 
benefits everyone. Right now every pays the price for a minority feature.

/Miek

--
Miek Gieben

chrisohaver:
FYI, @johnbelamaric, this wont be done by kube-con. Sorry!

plugin/kubernetes/controller.go
johnbelamaric: ? it's returned to the caller which is inside the client-go code.

plugin/kubernetes/object/service.go
johnbelamaric: Well, for most services the becomes make([]string, 0)

miekg:
>
>
>
>> @@ -167,14 +181,30 @@ func podIPIndexFunc(obj interface{}) ([]string, error) {
> 	return []string{p.PodIP}, nil
> }
>
>-func svcIPIndexFunc(obj interface{}) ([]string, error) {
>+func svcAllIPsIndexFunc(obj interface{}) ([]string, error) {
>+	svc, ok := obj.(*object.Service)
>+	if !ok {
>+		return nil, errObj
>+	}
>+	return append(svc.ExternalIPs, svc.ClusterIP), nil
>
>? it's returned to the caller which is inside the client-go code.
>

no other function does this in this file, why does this one need to do it?

johnbelamaric:
The other index funcs return a slice with a single element, or in the case
of the external IP one, the existing slice. In this case we want to index
by all the IPs, and we don't know the exact number (and it will be
different for each service).

@chrisohaver is this redundant with the separate cluster IP and external ip
indexes?

johnbelamaric:
I think this will be a pretty popular feature. We could make a separate
ToService function that doesn’t store the external ips, and choose which
one to use based on the setting. This means if you are not using the
feature, ExternalIPs field can always be nil.

miekg:
>I think this will be a pretty popular feature. We could make a separate
>ToService function that doesn’t store the external ips, and choose which
>one to use based on the setting. This means if you are not using the
>feature, ExternalIPs field can always be nil.

OTOH the 1% might not be much, but it's more than nothing and where to we draw 
the line? We were really counting the percentages back when.

chrisohaver:
Yes, we can easily do that. Though I don’t believe any services have external ips in the k8s cluster scale tests, so that test result would not be affected.

We have about 25% headroom under the memory limit at highest scale test.

On Dec 6, 2018, at 5:13 PM, John Belamaric <notifications@github.com<mailto:notifications@github.com>> wrote:

I think this will be a pretty popular feature. We could make a separate
ToService function that doesn’t store the external ips, and choose which
one to use based on the setting. This means if you are not using the
feature, ExternalIPs field can always be nil.

miekg:
...compared to kubedns. That 25% is not headroom, it's us being more efficient. In fact that scale test should slash 25% of that target and give those bytes to other processes.

Another thing this PR does is allowing a plugin to add zones to the initial list, no other plugin does this and changes the semantics of the Corefile considerably

chrisohaver:
Let me rephrase. We could grow 25% before hitting the coredns deployment’s memory limit.

This option does not add zones. The external zone must already be in the initial list.

On Dec 7, 2018, at 1:18 AM, Miek Gieben <notifications@github.com<mailto:notifications@github.com>> wrote:


...compared to kubedns. That 25% is not headroom, it's us being more efficient. In fact that scale test should slash 25% of that target and give those bytes to other processes.

Another thing this PR does is allowing a plugin to add zones to the initial list, no other plugin does this and changes the semantics of the Corefile considerably

plugin/backend_lookup.go
```diff
@@ -374,6 +374,7 @@ func NS(b ServiceBackend, zone string, state request.Request, opt Options) (reco
 	// ... and reset
 	state.Req.Question[0].Name = old
 
+	hostSeen := make(map[string]bool)
```
miekg, line 377: sure, another expensive map in the serving path, why not?

miekg:
Let me give a list on why this PR will not be merged. TL;DR: it increasing the maintenance burden in the future.

* it's a tiny feature
* functions get magical new boolean arguments that subtlety changes how they work.
* the hot serving path needs to be changed
* good documention in the sources is lacking

Good things:

* tests are good

Some other things that didn't help:

* putting an artificial timeframe on it (this has to be merged before xxxx)
* scaring us that its either this or a 3000 line copy/pasta PR (current alternative PR is ~500 lines minus tests)

PR #2357, opened 2018-12-01
Fix EDNS0 compliance

miekg:
Do SizeAndDo in the server and remove all uses of this from the plugins
and do it in the server in the ScrubWiriter. Also *always* do it. This
is to get into compliance  for https://dnsflagday.net/.

The pkg/edns0 now exports the EDNS0 options we understand; this is
exported to allow plugins add things here. This should be documented
still. We don't echo back options we don't understand.

Note I can't test it because of: https://github.com/miekg/dns/issues/857

Running a test instance and pointing the https://ednscomp.isc.org/ednscomp
to it shows the tests are now fixed:

~~~
EDNS Compliance Tester
Checking: 'miek.nl' as at 2018-12-01T17:53:15Z

miek.nl. @147.75.204.203 (drone.coredns.io.): dns=ok edns=ok edns1=ok edns@512=ok ednsopt=ok edns1opt=ok do=ok ednsflags=ok docookie=ok edns512tcp=ok optlist=ok
miek.nl. @2604:1380:2002:a000::1 (drone.coredns.io.): dns=ok edns=ok edns1=ok edns@512=ok ednsopt=ok edns1opt=ok do=ok ednsflags=ok docookie=ok edns512tcp=ok optlist=ok

All Ok
Codes
ok - test passed.
~~~

### DIFF
--- a/request/request.go
+++ b/request/request.go
@@ -194,7 +194,7 @@ func (r *Request) Size() int {
 // SizeAndDo adds an OPT record that the reflects the intent from request.
 // The returned bool indicated if an record was found and normalised.
 func (r *Request) SizeAndDo(m *dns.Msg) bool {
-	o := r.Req.IsEdns0() // TODO(miek): speed this up
+	o := r.Req.IsEdns0()
 	if o == nil {
 		return false
 	}
@@ -208,6 +208,10 @@ func (r *Request) SizeAndDo(m *dns.Msg) bool {
 		mo.SetUDPSize(o.UDPSize())
 		mo.Hdr.Ttl &= 0xff00 // clear flags
 
+		if len(o.Option) > 0 {
+			o.Option = supportedOptions(o.Option)
+		}
+
 		if odo {
 			mo.SetDo()
 		}
@@ -218,6 +222,10 @@ func (r *Request) SizeAndDo(m *dns.Msg) bool {
 	o.Hdr.Rrtype = dns.TypeOPT
 	o.SetVersion(0)
 	o.Hdr.Ttl &= 0xff00 // clear flags
+
+	if len(o.Option) > 0 {
+		o.Option = supportedOptions(o.Option)
+	}
 
 	if odo {
 		o.SetDo()
@@ -305,7 +313,6 @@ func (r *Request) Scrub(reply *dns.Msg) *dns.Msg {
 	}
 
 	if rl <= size {
-		r.SizeAndDo(reply)
 		return reply
 	}
 
@@ -341,7 +348,6 @@ func (r *Request) Scrub(reply *dns.Msg) *dns.Msg {
 		// this extra m-1 step does make it fit in the client's buffer however.
 	}
 
-	r.SizeAndDo(reply)
 	reply.Truncated = true
 	return reply
 }
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ godeps:
 	go get -u github.com/prometheus/client_golang/prometheus/promhttp
 	go get -u github.com/prometheus/client_golang/prometheus
 	(cd $(GOPATH)/src/github.com/mholt/caddy              && git checkout -q v0.11.1)
-	(cd $(GOPATH)/src/github.com/miekg/dns                && git checkout -q v1.1.0)
+	(cd $(GOPATH)/src/github.com/miekg/dns                && git checkout -q v1.1.1)
 	(cd $(GOPATH)/src/github.com/prometheus/client_golang && git checkout -q v0.9.1)
 
 .PHONY: travis
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -425,7 +425,6 @@ func BackendError(b ServiceBackend, zone string, rcode int, state request.Reques
 	m.Authoritative, m.RecursionAvailable = true, true
 	m.Ns, _ = SOA(b, zone, state, opt)
 
-	state.SizeAndDo(m)
 	state.W.WriteMsg(m)
 	// Return success as the rcode to signal we have written to the client.
 	return dns.RcodeSuccess, err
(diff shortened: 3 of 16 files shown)


miekg:
needs more tests and the edns.Supported is (of course) introducing a datarace.

miekg:
this breaks *rewrite* tests because we allow EDNS option rewriting there (/cc @greenpau )

rdrozhdzh:
from https://tools.ietf.org/html/rfc6891#section-6.2.6

>    Middleboxes that simply forward requests to a recursive resolver MUST
   NOT modify and MUST NOT delete the OPT record contents in either
   direction.

>    Middleboxes that have additional functionality, such as answering
   queries or acting as intelligent forwarders, SHOULD be able to
   process the OPT record and act based on its contents.  These
   middleboxes MUST consider the incoming request and any outgoing
   requests as separate transactions if the characteristics of the
   messages are different.

>    A more in-depth discussion of this type of equipment and other
   considerations regarding their interaction with DNS traffic is found
   in [RFC5625].

So, if we have simple config like
```
.:53 {
  forward 10.20.30.40
}
```
we "MUST NOT modify and MUST NOT delete the OPT record contents in either direction".

Here we don't know in advance which EDNS options are supported by upstream resolver or by authoritative servers responsible for some zones. So we cannot create a list of supported options.

Perhaps, it is responsibility of authoritative server (or plugins which behave like authoritative servser) to drop all unsupported options from response.

Please, let me know if my understanding of the standard is incorrect.

BTW, which config did you use when testing coredns for EDNS compliance?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fix EDNS0 com..." ]
>from https://tools.ietf.org/html/rfc6891#section-6.2.6
>
>>    Middleboxes that simply forward requests to a recursive resolver MUST
>   NOT modify and MUST NOT delete the OPT record contents in either
>   direction.
>
>>    Middleboxes that have additional functionality, such as answering
>   queries or acting as intelligent forwarders, SHOULD be able to
>   process the OPT record and act based on its contents.  These
>   middleboxes MUST consider the incoming request and any outgoing
>   requests as separate transactions if the characteristics of the
>   messages are different.
>
>>    A more in-depth discussion of this type of equipment and other
>   considerations regarding their interaction with DNS traffic is found
>   in [RFC5625].
>
>So, if we have simple config like
>```
>.:53 {
>  forward 10.20.30.40
>}
>```
>we "MUST NOT modify and MUST NOT delete the OPT record contents in either direction".

I'm not sure what the forward does right now here.

>Here we don't know in advance which EDNS options are supported by upstream resolver or by authoritative servers responsible for some zones. So we cannot create a list of supported options.

yes, this clashes with the standard and other plugins. Not sure what the 
solution should look like here.

>Perhaps, it is responsibility of authoritative server (or plugins which behave like authoritative servser) to drop all unsupported options from response.

Technically we *can* query the upstream and see what it supports? BIND9 does 
this as well.

>Please, let me know if my understanding of the standard is incorrect.

Think so, but I might need to read 6891 in its entirety, because OPT used to be 
hop by hop

>
>BTW, which config did you use when testing coredns for EDNS compliance?

~~~ corefile
. {
    file /tmp/miek.nl.signed miek.nl
    log
    errors
}
~~~

/Miek

--
Miek Gieben

rdrozhdzh:
> ~~~ corefile
> . {
>    file /tmp/miek.nl.signed miek.nl
>    log
>    errors
> }
> ~~~

I think in such config the `file` plugin should drop all unsupported ENDS options, since this plugin **creates** (not just relays or transmits) DNS response.

Other plugins and coredns itself should not touch any EDNS options in response

miekg:
the file plugin doesn't support edns0 option by itself, so I don't understand we you mean with drop. edns0 must be handled server side, the only things plugins can do is add options.

request/request.go
```diff
@@ -208,6 +208,10 @@ func (r *Request) SizeAndDo(m *dns.Msg) bool {
 		mo.SetUDPSize(o.UDPSize())
 		mo.Hdr.Ttl &= 0xff00 // clear flags
 
+		if len(o.Option) > 0 {
+			o.Option = supportedOptions(o.Option)
```
rdrozhdzh, line 212: Perhaps it should be smth like
```
mo.Option = append(mo.Option, supportedOptions(o.Option)...)
```
i.e. we should not remove (overwrite) existing options from response which might come from upstream server.
Also, we likely need to add check for duplicates in the code above, i.e. don't add option again if it already presents in response OPT record.

miekg: no, that will not work, because we whip the entire opt record anyway in forward and proxy right now

rdrozhdzh: @miekg, please check at least if `o.Option =` should be `mo.Option =`

Seems, in your solution you remove unknown options from query message, shouldn't you remove them from response?

See also line 23
```
m.Extra = append(m.Extra, o)
```
which is called only if response doesn't have OPT record

miekg: we've never removed OPT from the response, this PR doesn't change that behavior.

miekg:
I'll also fix the variable naming in that function `o` and `mo` are not helping at all

request/request.go
rdrozhdzh: I'm talking not about removing entire OPT RR, but about filtering individual options which you do in `supportedOptions()`. 
You take the options list from query OPT record, filter this list and assign back to **query** OPT. This doesn't affect **response** OPT record.

miekg: ... uh yes?

I don't understand what you're getting at: 
* is this something we need to do? 
* did we do this before and not anymore?
* Is there an actual bug in the PR?

rdrozhdzh: I meant to say that the code
https://github.com/coredns/coredns/blob/a2ad651405f1febaec4ffbd677f9ac190dada1a7/request/request.go#L211-L213
doesn't make any sense here in the block
```
if mo := m.IsEdns0(); mo != nil {
```
because we have `return` at the end of this block, and the modified object `o` is not used before `return`

For comparison, the similar code at
https://github.com/coredns/coredns/blob/a2ad651405f1febaec4ffbd677f9ac190dada1a7/request/request.go#L226-L228
may have sense, because the modified `o` object is used in line
https://github.com/coredns/coredns/blob/a2ad651405f1febaec4ffbd677f9ac190dada1a7/request/request.go#L233
before `return` statement

So, I just asked you to check if this code at lines 211-213 is really needed, or should be modified to make sense

miekg: you have a point

Issue #2364, opened 2018-12-04
plugin/rewrite: "name prefix" replacement does eat the first letter of the name


thomas-mangin:
Using CoreDNS 1.2.6 release

Using a configuration such as:
```
. {
        forward . 8.8.8.8

        rewrite stop name prefix test. not-a-test.
}
```

```
dig txt test.example.com @127.0.0.1

; <<>> DiG 9.12.3 <<>> txt test.example.com @127.0.0.1
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 36561
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;test.example.com.		IN	TXT

;; AUTHORITY SECTION:
xample.com.		300	IN	SOA	ns1.torresdns.com. admin.xample.com. 2018051801 86400 10800 604800 300

;; Query time: 48 msec
;; SERVER: 127.0.0.1#53(127.0.0.1)
;; WHEN: Tue Dec 04 07:56:15 GMT 2018
;; MSG SIZE  rcvd: 124
```
The "e" of "example get eaten away, same in the logs
```
2018-12-04T07:56:15.771Z [INFO] 127.0.0.1:51510 - 36561 "TXT IN not-a-test.xample.com. udp 63 false 4096" NOERROR qr,rd,ra 125 0.046009796s
```

daniel-garcia:
@chrisohaver I can take this issue.

thomas-mangin:
Thank you.

PR #2370, opened 2018-12-06
Use TrimPrefix instead of TrimLeft in rewrite prefix plugin (#2364)

daniel-garcia:
This PR fixes a bug where the rewrite plugin trims the prefix incorrectly due to using the TrimLeft function instead of the TrimPrefix function.

#2364 

None.


daniel-garcia:
There is an additional bug with suffix I suspect. Writing a test for that.

fturib (approved):
Good catch !
https://wrfly.kfd.me/posts/go-trimleft-and-trimprefix/

PR #2372, opened 2018-12-06
Use Trim(Prefix/Suffix) instead of Trim(Left/Right) in rewrite prefix…

daniel-garcia:
… plugin (#2364)


daniel-garcia:
@miekg @fturib  the suffix matching is also broken.

fturib (approved):
You are right ... needed to think also the balanced direction.
Thanks to update the UT accordingly.

PR #2374, opened 2018-12-06
msg.Service: optimize a bit

miekg:
Make the NewSRV and friends slightly smarter. Optimize the calling of
targetStrip which is almost certainly not used.

Added benchmark show a modest improvement:

benchmark             old ns/op     new ns/op     delta
BenchmarkNewSRV-4     300           283           -5.67%


nitisht (approved):
LGTM

PR #2376, opened 2018-12-07
fix a document error

9nix00:
fix a document error
the resolve file should be `/etc/resolv.conf` not `/etc/resolv/conf`


miekg:
yep, thanks for catching this.

Issue #2366, opened 2018-12-05
CAA record using etcd backend


jremond:
Hello,

I would like to add a CAA record in my coredns server. I use the etcd backend.
All I have been able to find is :
- https://github.com/coredns/coredns/pull/2084
- https://github.com/skynetservices/skydns/issues/321

My understanding is that the etcd backend mirrors the functionnalities of skydns (that's why I looked in that repository).

Is there any restriction on adding support for this type of record ?

Thanks

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] CAA record using ..." ]
>Hello,
>
>I would like to add a CAA record in my coredns server. I use the etcd backend.
>All I have been able to find is :
>- https://github.com/coredns/coredns/pull/2084
>- https://github.com/skynetservices/skydns/issues/321
>
>My understanding is that the etcd backend mirrors the functionnalities of skydns (that's why I looked in that repository).
>
>Is there any restriction on adding support for this type of record ?

Yes, it's not supported. The etcd is fairly limited in what record types it 
supports.

/plugin: etcd
/label: question

/Miek

--
Miek Gieben

jremond:
Is this by design or if a correct implementation is provided it could be merged ?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] CAA record us..." ]
>Is this by design or if a correct implementation is provided it could be merged ?

This was by design, we could extend msg.Service if we want to.

miekg:
I've come to the conclusion the current etcd plugin isn't a generic zone backend, so we can't treat it that way and adds all the RR types we need.

If we want to store all kinds of DNS data in etcd we need to new plugin.

Issue #2371, opened 2018-12-06
plugin/kubernetes: msg.Service has fields we don't use


miekg:
We're still using the msg.Service from the etcd plugin and further back from SkyDNS. We don't use half the fields in that struct:

* `Text`
* `Mail`
* `Group`
* `TargetStrip`

Text and group are a string, then a bool and a int. As we just this struct a lot I think we can save some memory by using a smaller one.

Will be fiddly to separate this out from the *etcd* plugin, but should be doable.

miekg:
No sure if they are that useful for etcd and who is still using those options. Probably someone, so we can rip them out easily.

miekg:
Generally we can actual sidestep backend.go and backend_lookup.go entirely and do this ourselves. The new k8s_external plugin shows had that can be done. The original etcd msg.Service had very different requirements and does a lot more - we do lesss in k8s so we don't need half of it. 

This should (we should measure) decrease memory while increasing qps

miekg:
Closing and deferring, because of time constraints.

PR #2377, opened 2018-12-07
WIP: Add caa etcd

jremond:

Add CAA record support to etcd plugin

https://github.com/coredns/coredns/issues/2366

I added an example in the README file of the etcd plugin

### DIFF
--- a/plugin/etcd/msg/service.go
+++ b/plugin/etcd/msg/service.go
@@ -34,6 +34,11 @@ type Service struct {
 
 	// Etcd key where we found this service and ignored from json un-/marshalling
 	Key string `json:"-"`
+
+	// CAA records require a Flag and a Tag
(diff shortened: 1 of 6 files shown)


plugin/etcd/msg/service.go
```diff
@@ -34,6 +34,11 @@ type Service struct {
 
 	// Etcd key where we found this service and ignored from json un-/marshalling
 	Key string `json:"-"`
+
```
miekg, line 37: your PR prompted  me to look into this struct and we should not bloat it anymore.

If we need this we should add a pointer to a CAA record struct and figure how to get the info out from etcd.
Also see #2371

miekg:
Closing as mentioned in the issue, needs to be implemented in another plugin that supports all RR types.

PR #2378, opened 2018-12-08
plugin/kubernetes: make prefix key in msg.Service a const

miekg:
This is a relic from the etcd integration and only used to remove in the
final answer returned to the client. We do need it. Make it a constant
and only one character wide.


yongtang (approved):
LGTM

PR #2379, opened 2018-12-08
Add new plugin: external - resolve k8s ingress and LB address with external names

miekg:
This plugin works in conjunction with the kubernetes plugin and exports
ingress and LB addresses as DNS records.

Alternative and much cleaner way to expose both of these elements.


miekg:
cc @johnbelamaric this is a much saner why to do #2356 (or the insanity that is #2337)

miekg:
if we agree this is better I can probably find some time to work this into something that works

miekg:
It still adds more code than I would like to see, but at least its contained in its own plugin. If the msg_to_dns.go just works we might consider moving that way for kubernetes as well and drop the the entire backend.go and backend_lookup.go to the etcd plugin.

miekg:
This does (almost) everything #2356 does without touching any of the existing files. It passes the same tests. The only thing it doesn't do is apex handling (hands out nodata right now).

As its own little plugin we are more flexible in axfr, soa parameters and other things people might want to set.

It still incurs the memory overhead of getting this data and caching it - this is done by default in the kubernetes plugin.

miekg:
this can be production ready and included in 1.2.7 easily (as it almost doesn't impact the k8s code).

miekg:
contentious bit is the change to `plugin/kubernetes/controller.go`, which does the same append.

Also this is still quite a bit of Go code, but mostly boilerplate and tests. Also fit federation and autopath.

plugin/external/README.md
```diff
@@ -0,0 +1,78 @@
+# external
+
+## Name
+
+*external* - resolve ingress and load balance IPs from outside kubernetes clusters.
```
johnbelamaric: Don't say "ingress" just say "load balancer and external IPs". Because "ingress" will imply L7 ingress to people, NOT the L3/L4 load balancer "ingress" IPs.

That's a completely different thing and would also be very cool to be able to manage, but not now...

plugin/external/README.md
```diff
@@ -0,0 +1,78 @@
...
+note the SOA's serial number is static. The IP addresses of the nameserver records are those of the
+CoreDNS service.
+
+The *external* plugin handles the subdomain `dns` and the apex of the zone by itself, all other
+queries are resolved to addresses in the cluster.
+
+## Syntax
+
+~~~
+external [ZONE...]
```
johnbelamaric: Can we think of a different name? I mean, "external" is very general and this is really k8s specific.

plugin/external/README.md
```diff
@@ -0,0 +1,78 @@
...
+				5          ; minimum (4 hours)
+				)
+example.org		5 IN	NS ns1.dns.example.org.
+
+ns1.dns.example.org.  5 IN  A    ....
+ns1.dns.example.org.  5 IN  AAAA ....
+~~~
+
+Note we use the `dns` subdomain to place the records the DNS needs (see the `apex` directive). Also
+note the SOA's serial number is static. The IP addresses of the nameserver records are those of the
```
johnbelamaric: Why is it static?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add new plugi..." ]
>+external [ZONE...]
>
>Can we think of a different name? I mean, "external" is very general and this is really k8s specific.

yes please! Any inspiration? we could underscore it with kubernetes i.e. 
`thing_kubenetes`, but that quite long, but may be applicable here.

>Why is it static?

lameness from my side, don't support AXFR in this plugin yet - so there is no 
reason to incremenent anything.

/Miek

--
Miek Gieben

johnbelamaric:
I would do kubernetes_external or k8s_external.

re: static - ok, once we support axfr then obviously that needs to change

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add new plugi..." ]
>I would do kubernetes_external or k8s_external.

Ok, i'll rename it to k8s_external and make the other changes.

plugin/external/README.md
miekg: done

plugin/external/README.md
miekg: done

miekg:
should probably also extend the example section with some actual setup in k8s and some dig examples etc. etc.

PR #2383, opened 2018-12-08
plugin/kubernetes: smaller map

miekg:
to continue with a theme: the map[string]bool can be reduced
to map[string]struct{} to reduce a tiny amount of memory.


yongtang (approved):
LGTM

yongtang (changes requested):
Actually, some changes in plugin/kubernetes/handler_test.go and plugin/kubernetes/handler_test.go and plugin/kubernetes/xfr_test.go are needed.

miekg:
yes I see. Done. Pushing now.

yongtang (approved):
LGTM, Thanks!

PR #2385, opened 2018-12-09
core: edns0 tweaks

miekg:
Per comment thread in https://github.com/coredns/coredns/pull/2357 which
spotted a bug; updated the code and added some comments.

This function should probably be redone as some point or made obsolete.

### DIFF
--- a/request/request.go
+++ b/request/request.go
@@ -206,10 +206,7 @@ func (r *Request) SizeAndDo(m *dns.Msg) bool {
 		mo.SetUDPSize(o.UDPSize())
 		mo.Hdr.Ttl &= 0xff00 // clear flags
 
-		// If the request had options set, but we don't support those, we should not echo those back.
-		if len(o.Option) > 0 {
-			mo.Option = supportedOptions(o.Option)
-		}
+		// Assume if the message m has options set, they are OK and represent what an upstream can do.
 
 		if o.Do() {
 			mo.SetDo()
--- a/request/request.go
+++ b/request/request.go
@@ -192,14 +192,12 @@ func (r *Request) Size() int {
 }
 
 // SizeAndDo adds an OPT record that the reflects the intent from request.
-// The returned bool indicated if an record was found and normalised.
+// The returned bool indicates if an record was found and normalised.
 func (r *Request) SizeAndDo(m *dns.Msg) bool {
 	o := r.Req.IsEdns0()
 	if o == nil {
 		return false
 	}
-
-	odo := o.Do()
 
 	if mo := m.IsEdns0(); mo != nil {
 		mo.Hdr.Name = "."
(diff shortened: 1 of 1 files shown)


request/request.go
```diff
@@ -192,32 +192,32 @@ func (r *Request) Size() int {
...
 		mo.Hdr.Name = "."
 		mo.Hdr.Rrtype = dns.TypeOPT
 		mo.SetVersion(0)
 		mo.SetUDPSize(o.UDPSize())
 		mo.Hdr.Ttl &= 0xff00 // clear flags
 
+		// If the request had options set, but we don't support those, we should not echo those back.
 		if len(o.Option) > 0 {
-			o.Option = supportedOptions(o.Option)
+			mo.Option = supportedOptions(o.Option)
```
rdrozhdzh: This code may erase existing options from response OPT record which seems incorrect. I would prefer to not merge this. The fix merged recently has noop code, but it at least doesn't break anything. 

Please consider proposals below
1. Just remove `if len(o.Option) > 0 {` block here. Since response message already has OPT record we can assume that other service (e.g. upstream server or another plugin) already set all options as needed
2. Add options supported by coredns to existing options (perhaps supported by upstream server) in response OPT record, i.e. add code like
```
if len(o.Option) > 0 {
  mo.Option = MERGE(mo.Option, supportedOptions(o.Option)...)
}
```
here `MERGE()` should add supported options but avoid duplicates

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] core: edns0 t..." ]
>
>
>
>>  		if len(o.Option) > 0 {
>-			o.Option = supportedOptions(o.Option)
>+			mo.Option = supportedOptions(o.Option)
>
>This code may erase existing options from response OPT record which seems incorrect. I would prefer to not merge this. The fix merged recently has noop code, but it at least doesn't break anything.
>
>Please consider proposals below
>1. Just remove `if len(o.Option) > 0 {` block here. Since response message already has OPT record we can assume that other service (e.g. upstream server or another plugin) already set all options as needed
>2. Add options supported by coredns to existing options (perhaps supported by upstream server) in response OPT record, i.e. add code like
>```
>if len(o.Option) > 0 {
>  mo.Option = MERGE(mo.Option, supportedOptions(o.Option)...)
>}
>```
>here `MERGE()` should add supported options but avoid duplicates

Both valid. For 1. I should check if built in plugins do this, I think not, so 
it should be OK to remove this and I like *not* doing things in the server path.

I'll ammend this PR

rdrozhdzh (approved):
Looks good to me.

Perhaps, the `if o.Do() {` block (line 211) may be removed as well, but I'm not sure.

miekg:
Hmmm. Yeah. Maybe. I'll propose something on the open bug(s), cause I don't
know right now.

miekg:
We're not compliant any more after this change:
~~~
miek.nl. @176.58.119.54 (linode.atoom.net.): dns=ok edns=ok edns1=noerror,badversion edns@512=ok ednsopt=ok edns1opt=noerror,badversion do=ok ednsflags=ok docookie=ok edns512tcp=ok optlist=ok 
miek.nl. @2a01:7e00::f03c:91ff:fe79:234c (linode.atoom.net.): dns=ok edns=ok edns1=noerror,badversion edns@512=ok ednsopt=ok edns1opt=noerror,badversion do=ok ednsflags=ok docookie=ok edns512tcp=ok optlist=ok 
~~~
this is with the file plugin. I'm going to make this compliant

PR #2386, opened 2018-12-10
map bool -> map struct{}

miekg:
This clears out the remaining map[x]bool usage and moves the bool to an
empty struct.

Nothing really in the serving path, so the saving are minimal, but good form to not use more than you need.

Two note worthy other changes:

* EnableChaos in the server is now also exported to make it show up in
  the documentation.
* The auto plugin is left as is, because there the boolean is
  explicitaly set to false to signal 'to-be-deleted' and the key is left
  as-is.


yongtang (approved):
LGTM, Thanks!

Issue #2394, opened 2018-12-14
setup functions


miekg:
Each plugin basically has the same setup() function that parses bits of the Corefile and sets up and structure to is the basis of the plugin.

From the  *erratic* plugin:
~~~
case "drop":
				args := c.RemainingArgs()
				if len(args) > 1 {
					return nil, c.ArgErr()
				}

				if len(args) == 0 {
					continue
				}

				amount, err := strconv.ParseInt(args[0], 10, 32)
				if err != nil {
					return nil, err
				}
				if amount < 0 {
					return nil, fmt.Errorf("illegal amount value given %q", args[0])
				}
				e.drop = uint64(amount)
				drop = true
~~~
RemainingArgs and the other bits are duplicated (and tested all over).

I propse we create a helper `plugin/pkg/setup` that abstract a lot of this way - not sure about the exact form. But the above could become
~~~
setup.Int64(c *caddy,Controller, min, max, error text) (int64, error)
~~~
Not sure about error text, but then the above becomes:
~~~
e.drop, err := setup.Int64(c, 0, -1)
if err != nil {
     return err
}
~~~

Of course we need 
* String
* Duration
* StringSlice
* Zone (?? maybe)

miekg:
We can then also cut down a bit (maybe) on all the testing - further along we might actually just want to generate this from the structure you describe and use tags `coredns:"int64"` and generate the lot

miekg:
nice, but *no*

PR #2397, opened 2018-12-14
Setup functions

miekg:
Proposal for #2394

### DIFF
--- a/plugin/erratic/setup.go
+++ b/plugin/erratic/setup.go
@@ -1,12 +1,11 @@
 package erratic
 
 import (
-	"fmt"
-	"strconv"
 	"time"
 
 	"github.com/coredns/coredns/core/dnsserver"
 	"github.com/coredns/coredns/plugin"
+	cs "github.com/coredns/coredns/plugin/pkg/coresetup"
 
 	"github.com/mholt/caddy"
 )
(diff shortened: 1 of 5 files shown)


miekg:
More descriptive error messages might be a thing. May need an extension of the Types to include a directive that can be used when erroring.

miekg:
this may also be overkill and doing it simpler in setup functions might be better.

fturib (commented):
I think we de not need the Value structure - because the only one that can use it is the one that already ask for the Parse with a defined Type.

= >Parse should return []interface{}

And the caller can just translate into the right type.

```
vals, err := cs.Parse(c, cs.Int{0, -1, 2})
if err != nil {
}
e.drop = uint64(vals[0])
```

fturib (commented):
I will be maybe more easy to split into 2 kinds of tools:
- check the number of parameters
- parse the value of each parameters depending the expected type and possible values.

We may rather need a set of simple helpers  that simplify the job:
```
args, err = ensureRemainingArgs(c, min, max)
myint, err := parseNum(args[0], 0, 32766, "delay of the xxx")
mydur,err  := parseDuration(arg[1], 10*Time.ms, 4*Time.Hour, "needed time for xxx")
```

miekg:
thanks for reviewing. I'll close this; nice exercise on what Go can do, doesn't help in readability at all.

PR #2398, opened 2018-12-14
[request] make proto return whether or not connection is using TLS

chantra:

this PR makes ResponseWriter.Proto return "tls" when the connection is over a TLS connection.

While technically this is just TCP, this aligns with the use of "tcp-tls" in (albeit I just return "tls" here): https://github.com/miekg/dns/blob/master/server.go#L127

I don't have strong opinions whether this should be in or not. I do find this useful and have something doing this already.... but figured it could be of some use to others if it was provided by request/request.go

### DIFF
--- a/plugin/test/responsewriter.go
+++ b/plugin/test/responsewriter.go
@@ -1,6 +1,7 @@
 package test
 
 import (
+	"crypto/tls"
 	"net"
 
 	"github.com/miekg/dns"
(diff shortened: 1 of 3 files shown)


chantra:
ok, new failure seems unrelated to this change and due to flaky tests over the network?

miekg:
if this is no redirect use  I don't see why we should merge it? Proto equaling 'tls' is technically not possible

chantra:
@miekg agreed that there is no such thing as Proto "tls". How about a new function that will return the listener type? It will leverage proto and add some syntax sugar to extract whether connection is using UDP, TCP, TLS.

I don't have strong opinions on this, but I found request package to offer so useful helper functions and being able to find out whether TLS is used or not in a simple call is useful to me and thought could be to others.

chantra:
feel free to close this  if you don't agree.

PR #2400, opened 2018-12-16
plugin/loop: show from -> to

miekg:
Show from and to address when detecting a loop they may aid in
debugging.

Hard to create a unit test, but this is a startup run with self induced
loop:

~~~ corefile
.:1053 {
    loop
    log
    forward . 127.0.0.1:1053
}
~~~~

~~~
:1053
2018-12-16T10:11:03.695Z [INFO] CoreDNS-1.3.0
2018-12-16T10:11:03.695Z [INFO] linux/amd64, go1.11,
CoreDNS-1.3.0
linux/amd64, go1.11,
2018-12-16T10:11:03.696Z [FATAL] plugin/loop: Loop (127.0.0.1:51384 -> :1053) detected for zone ".", see https://coredns.io/plugins/loop#troubleshooting. Query: "HINFO 2781022615773629442.4133547885299871809."
~~~

Update the docs and polished that a bit as well.

### DIFF
--- a/plugin/loop/README.md
+++ b/plugin/loop/README.md
@@ -7,10 +7,10 @@
 ## Description
 
 The *loop* plugin will send a random probe query to ourselves and will then keep track of how many times
-we see it. If we see it more than twice, we assume CoreDNS is looping and we halt the process.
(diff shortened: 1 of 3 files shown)


yongtang (approved):
LGTM

PR #2402, opened 2018-12-16
Fuzzing: add more fuzzing targets

miekg:
Also add Corefile fuzzing.


miekg:
fuzzed the caddyfile loader for a few hours (a short time) yesterday and nothing turned up, so that's good. I'm also fuzzing miekg/dns again - separately.

Issue #2024, opened 2018-08-06
RA flag is set unconditionally


Isolus:
While using coredns as authoritative name server for a .de domain, the predelegation check of the nic complains about "Recursive queries should not be allowed (resolver)".

In plugin/file (and in others) we have `m.Authoritative, m.RecursionAvailable = true, true` unconditionally. I think `m.RecursionAvailable` should only be true if recursion is available.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] RA flag is set un..." ]
>While using coredns as authoritative name server for a .de domain, the predelegation check of the nic complains about "Recursive queries should not be allowed (resolver)".
>
>In plugin/file (and in others) we have `m.Authoritative, m.RecursionAvailable = true, true` unconditionally. I think `m.RecursionAvailable` should only be true if recursion is available.

This means you need to check if proxy is loaded and only set it then. This 
creates problems for external plugins, because we can't compile in a list for 
all external plugins (explugins/unbound comes to mind). So this means a plugin 
must signal it can do recursive queries, so the server can see that. Probably a 
interface with a simple (noop) function so we can check for that?

miekg:
show me an implementation that actually cares about this, otherwise I'll close this issue as it's not that important.

Isolus:
The DENIC Predelegation Check (you can find an online version here: https://www.denic.de/en/service/tools/nast/ ) cares about this. Everytime you update a .de domain (e. g. DNS Servers, DNSSEC key material) the check is performed and reported. At the moment this is only displayed as a warning but if you have a registrar who treats warnings as errors, you can't change anything for your domain.
But even if you are allowed to skip this warning, it's a scary warning to everyone who doesn't have some deeper knowledge of DNS. And I don't like to ask friends or customers to skip a scary warning. That trains bad behavior.

I was thinking about your proposed solution. Is it clear at compile time whether a plugin allows recursive requests? Or you can imagine plugins where the behavior depends on the configuration?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] RA flag is se..." ]
>The DENIC Predelegation Check (you can find an online version here: https://www.denic.de/en/service/tools/nast/ ) cares about this. Everytime you update a .de domain (e. g. DNS Servers, DNSSEC key material) the check is performed and reported. At the moment this is only displayed as a warning but if you have a registrar who treats warnings as errors, you can't change anything for your domain.
[...]

PR #2408, opened 2018-12-18
disable ra flag for several plugins

Isolus:
This pull request disables the RA flag for several plugins. If CoreDNS is used as forwarder, the flag of forwarded answers of a recursive resolver remains unchanged.
A recursive resolver plugin should be the only case where the flag is set directly. So it shouldn't be set in the template either.
I'm not sure if the change in `plugin/etcd/stub_handler.go` breaks something. The flag does not seem to be used in stub zones in the sense that it advertises the general availability of recursion but rather that recursion is performed for this request and I'm not sure if other software depends on this behaviour.

#2024

### DIFF
--- a/plugin/auto/auto.go
+++ b/plugin/auto/auto.go
@@ -70,7 +70,7 @@ func (a Auto) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (i
 
 	m := new(dns.Msg)
 	m.SetReply(r)
-	m.Authoritative, m.RecursionAvailable = true, true
+	m.Authoritative, m.RecursionAvailable = true, false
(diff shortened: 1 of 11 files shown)


greenpau:
> I'm not sure if the change in plugin/etcd/stub_handler.go breaks something. 

@miekg , @nitisht, I noticed this comment and your review would be more appropriate because you are the owners of  `plugin/etcd`.

miekg:
this looks harmless. But we can remove the assignment entirely because it's `false` by default if you don't set it.

I think this is all copy/pasted code from skydns even.

miekg:
`m.Authoritative, m.RecursionAvailable = true, false` can be just

`m.Authoritative = true` throughout

PR #2418, opened 2018-12-27
Fix schema for federated services

shashidharatd:
Fix an outstanding issue when using federation plugin. Refer to this [issue](https://github.com/kubernetes/dns/issues/29). Federation service will use `fsvc` keyword instead of `svc` keyword in the kubernetes service schema. This will allow the controllers to clearly disambiguate the requests.
This PR also allow the federation service redirection to work when there are missing zone/region labels. This would help in some `on-prem` environments which does not have these labels.

https://github.com/kubernetes/dns/issues/29

Need to update `DNS spec for kubernetes` and `kubernetes documentation`. No updates required in this repo.

/cc @johnbelamaric @chrisohaver

### DIFF
--- a/plugin/federation/federation.go
+++ b/plugin/federation/federation.go
@@ -124,7 +124,7 @@ func (f *Federation) Name() string { return "federation" }
 
 // IsNameFederation checks the qname to see if it is a potential federation. The federation
 // label is always the 2nd to last once the zone is chopped of. For instance
-// "nginx.mynamespace.myfederation.svc.example.com" has "myfederation" as the federation label.
+// "nginx.mynamespace.myfederation.fsvc.example.com" has "myfederation" as the federation label.
 // IsNameFederation returns a new qname with the federation label and the label itself or two
 // empty strings if there wasn't a hit.
 func (f *Federation) isNameFederation(name, zone string) (string, string) {
--- a/plugin/federation/federation_test.go
+++ b/plugin/federation/federation_test.go
@@ -37,18 +37,18 @@ func TestFederationKubernetes(t *testing.T) {
 	tests := []test.Case{
 		{
 			// service exists so we return the IP address associated with it.
-			Qname: "svc1.testns.prod.svc.cluster.local.", Qtype: dns.TypeA,
+			Qname: "svc1.testns.prod.fsvc.cluster.local.", Qtype: dns.TypeA,
 			Rcode: dns.RcodeSuccess,
 			Answer: []dns.RR{
-				test.A("svc1.testns.prod.svc.cluster.local.      303       IN      A       10.0.0.1"),
+				test.A("svc1.testns.prod.fsvc.cluster.local.      303       IN      A       10.0.0.1"),
 			},
 		},
 		{
 			// service does not exist, do the federation dance.
-			Qname: "svc0.testns.prod.svc.cluster.local.", Qtype: dns.TypeA,
+			Qname: "svc0.testns.prod.fsvc.cluster.local.", Qtype: dns.TypeA,
 			Rcode: dns.RcodeSuccess,
 			Answer: []dns.RR{
-				test.CNAME("svc0.testns.prod.svc.cluster.local.  303       IN      CNAME   svc0.testns.prod.svc.fd-az.fd-r.federal.example."),
+				test.CNAME("svc0.testns.prod.fsvc.cluster.local.  303       IN      CNAME   svc0.testns.prod.fsvc.fd-az.fd-r.federal.example."),
 			},
 		},
 	}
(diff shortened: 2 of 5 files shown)


miekg:
That issue is closed a year ago, why is this relevant?

shashidharatd:
> That issue is closed a year ago, why is this relevant?

Even though the issue is closed, the feature is still supported in [federation-v2](https://github.com/kubernetes-sigs/federation-v2), The issue existed in `kube-dns`, and when it is ported to `CoreDNS`, the same issue still exist. This can be issue in configurations where `federation` plugin is enabled and also supports statefulsets. Until now this was not a hurdle as federation feature took a back seat for some time. Here is some description on how the [cross cluster service discovery](https://docs.google.com/presentation/d/1b6yI3FpysxvdLvCmrR5sgIeDcG7nMGdG3b169sed-pY/edit#slide=id.g465e004146_0_58) is supported in federation-v2.
Request to do the needful. Thanks!

miekg:
That's just a slide deck. Where is a design doc? What version of k8s
support this.

If we want to do something sane with federation (I'm all for it, current
situation is quite bad), put it in the DNS spec.

shashidharatd:
> That's just a slide deck. Where is a design doc? What version of k8s
support this.
If we want to do something sane with federation (I'm all for it, current
situation is quite bad), put it in the DNS spec.

would send out the update to DNS spec. `k8s version` does not matter as federation code is out of the core.

fturib:
@chrisohaver : what is your take on this ?

chrisohaver:
Have design proposals for federation-v2 gone through the KEP process?

chrisohaver:
Looking at the repo, it looks like federation-v2 is in the "prototype" phase...
@shashidharatd, how would you categorize the current state of federation-v2?
* pre-alpha
* alpha
* beta

Seems like pre-alpha? So, perhaps it makes sense, during this phase of development/prototyping, to develop the feature as a new external plugin, instead of modifying the existing built in plugin.

chrisohaver:
Yes - drilling a bit deeper, the design doc is still "a work in progress"... so I don't think it makes sense to merge code into the coredns release at this time.  Even after the design solidifies, it probably makes sense to leave the existing federation solution as is, until federation-v2 graduates past it.

The existing beta [federation solution](https://kubernetes.io/docs/concepts/cluster-administration/federation/) is still the only federation solution until federation-v2 is ready to replace it.

shashidharatd:
> Looking at the repo, it looks like federation-v2 is in the "prototype" phase...
@shashidharatd, how would you categorize the current state of federation-v2?
pre-alpha
alpha
beta
Seems like pre-alpha? So, perhaps it makes sense, during this phase of development/prototyping, to develop the feature as a new external plugin, instead of modifying the existing built in plugin.

currently, federation-v2 is kind of pre-release and soon would be an alpha release by release versioning.

> Yes - drilling a bit deeper, the design doc is still "a work in progress"... so I don't think it makes sense to merge code into the coredns release at this time. Even after the design solidifies, it probably makes sense to leave the existing federation solution as is, until federation-v2 graduates past it.
The existing beta federation solution is still the only federation solution until federation-v2 is ready to replace it.

federation v1, is not in active development for long time now. whoever interested has already been diverted to use federation-v2. coming to the [design](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/multicluster/federated-services.md), it was freezed long back, but it has implementation details specific to federation-v1. we want to provide exact same feature in federation-v2. we might need to rewrite that design doc using federation-v2 as implementation details.
Coming to this PR, this was basically to avoid a potential clash in DNS records of a federated service wrt statefulset as described in this [issue](https://github.com/kubernetes/dns/issues/29#issuecomment-273395067)

chrisohaver:
This change breaks the existing federation-v1 solution.  If merged now, anyone who is using federation-v1 would be forced to migrate their federation deployment to federation-v2 when upgrading to kubernetes 1.14 (or manually downgrade coredns after the update).  Therefore, I don't think we should be making this change to the federation plugin at this time.

An alternative could be to create a new plugin _federationv2_, initially as an external plugin.
When federation-v2 is field ready, it could be merged into core coredns - CoreDNS having both v1 and v2 implementations.
After a period of deprecation, we could then remove the existing _federation_ plugin.

Alternately we could add a version directive to the existing federation plugin, defaulting for now to v1, making the federation plugin support both versions. Although at the moment I prefer a separate plugin.

Several other approaches, e.g. branches, forks...

I defer to @miekg's decision here.

Despite federation-v1 being inactive for a long time, according to kubernetes docs, federation-v1 is still the current federation implementation at this time.  Kubernetes docs mention federation-v2, but only to say that it is still being built, and will replace federation-v1 in the future.  If follows that people would still be using using federation-v1.

miekg:
Agreeing with Chris here. Breaking change, needs config or a fallback. If
this change can be made without that in mind it says to me no one is using
federation and we can remove the entire plugin.

shashidharatd:
I do agree with  your suggestions @chrisohaver and @miekg. It makes perfect sense. Although making another plugin, we would have quite a lot of duplicated code. So i am inclined to introduce version directive as @chrisohaver suggested above. Do let me know if that's fine and thanks for your guidance.

miekg:
Version directive is fine.

miekg:
this PR as such won't go in; we need a newer one with a backwards compat path. Closing here.

Issue #2417, opened 2018-12-26
plugin/log - how to set multi name


ckcd:
Hi, I'm using CoreDNS  v1.3.0 in K8S v1.9,

I want to use `log` plugin to log multi domains, such as `a.com` and `b.com`, but i find the `log` plugin can only set ONE name.

```
log [NAME] [FORMAT] {
    class CLASSES...
}
```
e.g. 
```
log a.com
```

And obviously `log a.com b.com` is bad because the second args is `FORMAT`.

is there any methods to set multi name ?

miekg:
Either send a pr to fix this in the log plugin, or split your zones into
multiple server blocks and use log in each of them.

ckcd:
@miekg 
Thanks for you reply !

The second methods seems too tedious,

I will take the first suggestion and do some code, hope can do it and give a pr.

PR #2420, opened 2018-12-28
plugin/log: support multi nameScope

ckcd:
A large amount of logging will reduce performance, as the [plugin/log](https://coredns.io/plugins/log/) says:

> Note that for busy servers logging will incur a performance hit.

Fortunately `log` plugin can set the name that should be log:
```
log [NAME] [FORMAT] {
    class CLASSES...
}
```
So we can only log the domains that we concerned.

But, for now it can only set ONE name, and it's hard to make some workaround in `corefile` if we want to log more domains.

So this PR changes the `log` plugin to support multi name, and the `corefile` looks like:
```
log [NAMES...] {
    format FORMAT
    class CLASSES...
}
```

#2417 

See `log/README.md`

### DIFF
--- a/plugin/log/README.md
+++ b/plugin/log/README.md
@@ -23,10 +23,10 @@ log
 Or if you want/need slightly more control:
 
 ~~~ txt
-log [NAME] [FORMAT]
+log [NAMES...] [FORMAT]
 ~~~
 
-* `NAME` is the name to match in order to be logged
+* `NAMES` is the name list to match in order to be logged
 * `FORMAT` is the log format to use (default is Common Log Format), `{common}` is used as a shortcut
   for the Common Log Format. You can also use `{combined}` for a format that adds the query opcode
   `{>opcode}` to the Common Log Format.
@@ -34,7 +34,7 @@ log [NAME] [FORMAT]
 You can further specify the classes of responses that get logged:
 
 ~~~ txt
-log [NAME] [FORMAT] {
+log [NAMES...] [FORMAT] {
     class CLASSES...
 }
 ~~~
--- a/plugin/log/setup.go
+++ b/plugin/log/setup.go
@@ -1,6 +1,8 @@
 package log
 
 import (
+	"strings"
+
 	"github.com/coredns/coredns/core/dnsserver"
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/pkg/response"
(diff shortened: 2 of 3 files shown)


plugin/log/README.md
```diff
@@ -23,20 +23,21 @@ log
 Or if you want/need slightly more control:
 
 ~~~ txt
-log [NAME] [FORMAT]
```
miekg, line 26: If two parameters where specified we already assume the last one if the format - that should also make this change smaller (I guess).

Looking at the code and docs I see we also don't mention the use of `{common}` or `{combined}` as a shortcut. I'll send a PR for that.

plugin/log/log.go
```diff
@@ -28,42 +28,43 @@ type Logger struct {
 func (l Logger) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 	state := request.Request{W: w, Req: r}
 	for _, rule := range l.Rules {
-		if !plugin.Name(rule.NameScope).Matches(state.Name()) {
```
miekg: Would initiating multiple rules with the new names make sense here? We mean none of this code needs to change

ckcd: Hi,

I think it needs change, because in origin design, one `NAME` is associate with a `FORMAT`, so now in this PR, when initiating multiple rules with the new names, we have two choice:

1. just set one `FORMAT` for one`NAME` as origin, and don't need to change code here.
2. set multi `NAME` with one `FORMAT`, just as i do. and changes code here.

for method 1, That is a workaround of this situation, but it needs a format and a class for each name in `[]Rule`, which causes more memory. that's why i send this PR.

so i chose method 2, which needs to change code here.

thanks.

plugin/log/README.md
ckcd: @miekg  Hi,

Thanks for your review,

I understand the opinion of make smaller change, 

but for now the format is `NAME FORMAT`, i don't know how to changes to `NAMES... FORMAT` in the code because it's hard to judge the last parameter that whether it's a name or a format. Unless  changes to `FORMAT NAMES...` which also changes bigger.

Also, for this plugin, `NAME` or `NAMES` is necessary but `FORMAT` and `class` is optional, so maybe set these two in `{}` would be better.

miekg: No, we just say the last one is `FORMAT`, this is happening now anyway in the code so there would be no change

plugin/log/log.go
miekg: 1. is the better way

The map in the rule is a bit unfortunate (and causes most memory). Looks like an easy thing to remove/change.

ckcd:
@miekg 

Hi,

I have change the PR with your two opinions:

1. make the `FORMAT` be the last one as origin;
2. use method 1 in our conversation, which means do NOT need to change the `serveDNS`;

Would you please review it again?

Thank you !!!

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/log: s..." ]
>@miekg
>
>Hi,
>
>I have change the PR with your two opinions:
>
>1. make the `FORMAT` be the last one as origin;
>2. use method 1 in our conversation, which means do NOT need to change the `serveDNS`;
>
>Would you please review it again?
>
>Thank you !!!

Thanks. I'll take a look.

plugin/log/setup.go
```diff
@@ -34,6 +36,7 @@ func logParse(c *caddy.Controller) ([]Rule, error) {
 
 	for c.Next() {
 		args := c.RemainingArgs()
+		length := len(rules)
 
 		if len(args) == 0 {
```
miekg: while we're touching this, let's convert this if-else-if-else to a switch as well, looks somewhat cleaner.

plugin/log/setup.go
```diff
@@ -51,45 +54,57 @@ func logParse(c *caddy.Controller) ([]Rule, error) {
...
+					format = args[last]
+				}
 			}
 
-			rules = append(rules, Rule{
-				NameScope: dns.Fqdn(args[0]),
-				Format:    format,
-				Class:     make(map[response.Class]struct{}),
-			})
+			args = args[:last]
```
miekg: can we not just chopped of the last element in the `strings.Contains` bit above? Then `args` is always correct, and we don't need to last even.

plugin/log/setup.go
```diff
@@ -51,45 +54,57 @@ func logParse(c *caddy.Controller) ([]Rule, error) {
...
+				rules = append(rules, Rule{
+					NameScope: dns.Fqdn(str),
+					Format:    format,
+					Class:     make(map[response.Class]struct{}),
+				})
+			}
 		}
 
 		// Class refinements in an extra block.
+		classes := make(map[response.Class]struct{})
```
miekg, line 82: it's unclear to me why this change is needed?

plugin/log/setup.go
```diff
@@ -51,45 +54,57 @@ func logParse(c *caddy.Controller) ([]Rule, error) {
...
 			}
 		}
-		if len(rules[len(rules)-1].Class) == 0 {
-			rules[len(rules)-1].Class[response.All] = struct{}{}
+		if len(classes) == 0 {
+			classes[response.All] = struct{}{}
+		}
+
+		for i := len(rules) - 1; i >= length; i -= 1 {
+			rules[i].Class = classes
```
miekg, line 107: the class map is now shared between all rules? Might make sense, but should probably be done separately?

miekg:
Starting to look good.

plugin/log/setup.go
ckcd: Yes, the class map is shared, maybe we can add some optional to set class map for each nameScope, but it needs to change the format of this plugin also, e.g. 
```
log FORMAT {
NAME1 CLASSES...
NAME2 CLASSES...
...
}
```

plugin/log/setup.go
ckcd: Reason of this change is because each nameScope share class map as we talk in next comment.

plugin/log/setup.go
ckcd: Good idea~ doing it~

plugin/log/setup.go
ckcd: awesome !! i will take it~

ckcd:
Sorry for the twice force push because i forget to rebase master in first one.

change log:
1. using `switch` rather than `if/else`;
2. delete `last` even and chopped `args` directly;

and for the case of: class map shared for multi nameScope, i have some ideas but it needs more changes and thinking, maybe we should do it in another part.


and i don't understand why test for other irrelevant plugin is failed, test of this plugin is OK.
the error in travis is:
```
--- FAIL: TestStop (0.07s)
    --- FAIL: TestStop/Test_0 (0.01s)
        proxy_test.go:66: Test 0, Expected healthchecks to hit test server, got none
```
and it seems NO any  relation with this PR.
@miekg does it make sense ?

plugin/log/setup.go
miekg: ah yes I see. Let's not make this change right now though.

plugin/log/setup.go
ckcd: done, please check.

plugin/log/setup.go
ckcd: Hi,

I have make the changes, please have a check.

Thank you !!!

PR #2422, opened 2018-12-30
plugin/log: detail {combined} and {common}

miekg:
These can be used (and are handy), document them.


miekg:
seems dreck is not assigning reviewers... grrr.

PR #2425, opened 2019-01-02
plugin/etcd: Fix CNAME results for empty host fields

callebjorkell:

When a CNAME is requested, and a service with a filled out host field is not present for the exact key, a NODATA result should be returned. If the key exists, and the host field is empty, coredns erroneously returns a NOERROR response with a "." as the result. This would for example manifest when the requested key had a service representing a TXT record (Text field filled, empty Host field).

Fixing by specifically checking for empty Host fields when a CNAME lookup is done, and returning an empty answer, with the NOERROR Rcode.

N/A?

None

### DIFF
--- a/plugin/etcd/cname_test.go
+++ b/plugin/etcd/cname_test.go
@@ -58,6 +58,7 @@ var servicesCname = []*msg.Service{
 	{Host: "cname6.region2.skydns.test", Key: "cname5.region2.skydns.test."},
 	{Host: "endpoint.region2.skydns.test", Key: "cname6.region2.skydns.test."},
 	{Host: "mainendpoint.region2.skydns.test", Key: "region2.skydns.test."},
+	{Host: "", Key: "region3.skydns.test.", Text: "SOME-RECORD-TEXT"},
 	{Host: "10.240.0.1", Key: "endpoint.region2.skydns.test."},
 }
 
@@ -81,6 +82,13 @@ var dnsTestCasesCname = []test.Case{
 		Qname: "region2.skydns.test.", Qtype: dns.TypeCNAME,
 		Answer: []dns.RR{
 			test.CNAME("region2.skydns.test.	300	IN	CNAME	mainendpoint.region2.skydns.test."),
+		},
+	},
+	{
+		Qname: "region3.skydns.test.", Qtype: dns.TypeCNAME,
+		Rcode: dns.RcodeSuccess,
+		Ns: []dns.RR{
+			test.SOA("skydns.test.	303	IN	SOA	ns.dns.skydns.test. hostmaster.skydns.test. 1546424605 7200 1800 86400 30"),
 		},
 	},
 }
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -316,6 +316,9 @@ func CNAME(b ServiceBackend, zone string, state request.Request, opt Options) (r
 
 	if len(services) > 0 {
 		serv := services[0]
+		if len(serv.Host) == 0 {
+			return nil, nil
+		}
 		if ip := net.ParseIP(serv.Host); ip == nil {
 			records = append(records, serv.NewCNAME(state.QName(), serv.Host))
 		}


callebjorkell:
The declaration of an exported error from the plugin package was needed to be able to check for that specific one in the `etcd.IsNameError` function. The plugin and etcd packages are slightly tangled when it comes to this code path, and it seemed to me like this specific check is something that should be done up on the `plugin` package level since it _is_ something that applies to all potential plugins I think. Also, since the `ServiceBackend.Services` function is generically fetching services from the backend, it's not possible to do this filtering there without adding in more parameters. If you see a better way to structure it, I'm all ears.

callebjorkell:
Sorry for the force pushes... had some issues getting the tests to play nice.

plugin/etcd/etcd.go
```diff
@@ -66,7 +66,7 @@ func (e *Etcd) Lookup(state request.Request, name string, typ uint16) (*dns.Msg,
 
 // IsNameError implements the ServiceBackend interface.
 func (e *Etcd) IsNameError(err error) bool {
-	return err == errKeyNotFound
+	return err == errKeyNotFound || err == plugin.ErrEmptyHost
```
miekg: this should not be a name error. It should return a nodata response

plugin/etcd/cname_test.go
```diff
@@ -83,4 +84,11 @@ var dnsTestCasesCname = []test.Case{
 			test.CNAME("region2.skydns.test.	300	IN	CNAME	mainendpoint.region2.skydns.test."),
 		},
 	},
+	{
+		Qname: "region3.skydns.test.", Qtype: dns.TypeCNAME,
+		Rcode: dns.RcodeNameError,
```
miekg: For this test it needs to return Rcode: dns.RcodeSuccess with (still) the SOA in the Auth section.

callebjorkell: Hmmm. Yes, you're right of course, since there actually is a record (TXT) for the queried domain. I got hung up on the fact that the entry that was requested didn't exist for the requested type.

callebjorkell:
This was actually better because it also means that I didn't need the exported error anymore. Please have another look.

plugin/etcd/cname_test.go
```diff
@@ -58,6 +58,7 @@ var servicesCname = []*msg.Service{
 	{Host: "cname6.region2.skydns.test", Key: "cname5.region2.skydns.test."},
 	{Host: "endpoint.region2.skydns.test", Key: "cname6.region2.skydns.test."},
 	{Host: "mainendpoint.region2.skydns.test", Key: "region2.skydns.test."},
+	{Host: "", Key: "region3.skydns.test.", Text: "SOME-RECORD-TEXT"},
```
miekg, line 61: wouldn't this be also true for other types well? I.e A query when there's only a txt record?

pr looks good btw, but this might be a bigger issue than just for CNAME queries.

callebjorkell: Haven't really thought too much about it, but the main difference between the CNAME and the other record types is the recursion that doesn't take place in the CNAME lookups, so at the very least, the problem would be more complex (if it does exist) for the other record types.

callebjorkell: ... I just tried to add a TXT service to the `lookup_test`, and it does indeed seem like the same thing happens there. The record gets included in the Answer section as a CNAME record pointing to `.`

callebjorkell: I actually have another fix now that should also cover this case for A, AAAA, and SRV lookups. It's a fairly crude check if the host is empty or not, but it seems to be working. Would you prefer that I push that here, or should I open a new PR for that change?

miekg:
Not sure yet. Feels like a more generic fix is in order. Txt is probably
the only one where the host field doesn't matter.

callebjorkell:
Maybe we can outline a better solution in #2443 ?

callebjorkell:
Anything I can do right now to move this forwards?

miekg:
approach outlined in #2443 sounds better to me, see my latest comment.

callebjorkell:
Replaced by #2499

PR #2428, opened 2019-01-03
Add import plugin to give it  docs

miekg:
Fake the import plugin further to give it proper docs. This should help
make it more discoverable.


miekg:
note this reference the not-yet existing url coredns.io/plugins/import

yongtang (approved):
LGTM

Issue #1464, opened 2018-01-29
plugin/kubernetes: API connections to secure endpoint list fails


chrisohaver:
Secure API connections fail if the list of endpoints is > 1. 

See tests in `/coredns/ci/test/kubernetes/excluster_test.go` `TestKubernetesSecureAPIFallthrough`
The test is currently being skipped. coredns/ci#30

chrisohaver:
@yongtang - Have you had a chance to look into this?

chrisohaver:
IMO, we should either:
1. Fix this issue ... or 
2. Remove the api endpoint failover list feature

Without secure connections working, this feature is not very useful/practical.

yongtang:
What is the issue related to multiple api endpoint?

chrisohaver:
> What is the issue related to multiple api endpoint?

yes.

chrisohaver:
We should remove the feature if this is not going to be fixed.

miekg:
so, lets remove it

PR #2429, opened 2019-01-03
plugin/kubernetes: Remove multi API endpoint option

chrisohaver:

Removing multi k8s api connections option per #1464.

#1464

included

### DIFF
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -32,7 +32,7 @@ all the zones the plugin should be authoritative for.
 ```
 kubernetes [ZONES...] {
     resyncperiod DURATION
-    endpoint URL [URL...]
+    endpoint URL
     tls CERT KEY CACERT
     kubeconfig KUBECONFIG CONTEXT
     namespaces NAMESPACE...
@@ -51,9 +51,6 @@ kubernetes [ZONES...] {
 * `resyncperiod` specifies the Kubernetes data API **DURATION** period.
 * `endpoint` specifies the **URL** for a remote k8s API endpoint.
    If omitted, it will connect to k8s in-cluster using the cluster service account.
-   Multiple k8s API endpoints could be specified:
-   `endpoint http://k8s-endpoint1:8080 http://k8s-endpoint2:8080`. CoreDNS
-   will automatically perform a healthcheck and proxy to the healthy k8s API endpoint.
 * `tls` **CERT** **KEY** **CACERT** are the TLS cert, key and the CA cert file names for remote k8s connection.
    This option is ignored if connecting in-cluster (i.e. endpoint is not specified).
 * `kubeconfig` **KUBECONFIG** **CONTEXT** authenticates the connection to a remote k8s cluster using a kubeconfig file. It supports TLS, username and password, or token-based authentication. This option is ignored if connecting in-cluster (i.e., the endpoint is not specified).
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -30,7 +30,7 @@ type Kubernetes struct {
 	Next             plugin.Handler
 	Zones            []string
 	Upstream         upstream.Upstream
-	APIServer    string
+	APIServer        string
 	APIProxy         *apiProxy
 	APICertAuth      string
 	APIClientCert    string
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -6,15 +6,12 @@ import (
 	"fmt"
 	"net"
 	"strings"
-	"sync/atomic"
-	"time"
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/etcd/msg"
 	"github.com/coredns/coredns/plugin/kubernetes/object"
 	"github.com/coredns/coredns/plugin/pkg/dnsutil"
 	"github.com/coredns/coredns/plugin/pkg/fall"
-	"github.com/coredns/coredns/plugin/pkg/healthcheck"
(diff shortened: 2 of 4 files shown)


chrisohaver:
This change deprecates an option (multiple api endpoints).
I don't know the precise process for that.

* First merge/release a deprecation notice?
* Then merge this and release, adjusting version as appropriate?

yongtang:
I haven't looked into multiple secure https endpoint, but multiple http endpoint was fine and I were using that feature (with multiple http).

Could the coredns-ci be a certificate issue? I couldn't see enough context from coredns-ci.

chrisohaver:
It has been a while since I originally raised this issue, so I don't recall the details.
I could not get multiple api endpoints to work over a secure connection when testing in a local cluster.
The CI does successfully test a single endpoint over a secure connection, so the certs used by the CI in that test at least are valid. 

We could instead leave things as is, and document that we don't support secure connections for multiple api endpoints.

The code in `getClientConfig()` appears to assume an insecure connection, with "http://" hard coded in a couple of spots.

It may be easier to fix this than to deprecate, but the original bug wasn't getting any attention.
Also, no one else has raised this issue, and it would be IMO unusual for people to use this with an insecure connection, so I concluded that no one was actively using this feature.

yongtang:
Do you have a setup or test with certificate that could expose the issue? I always had some trouble (not even specific to multiple https endpoint) for certificate with k8s api, but I could look into the multiple https endpoint issue if a setup or test with valid cert is available.

chrisohaver:
There is a commented out test in the CI (in `test/kubernetes/excluster_test.go`)... but...

It should be easier to test locally.  I use minkube, and frequently test coredns builds running locally on my laptop configured to connect to the minikube cluster from outside of the cluster. For example, this kubernetes plugin config works for me...

```
    kubernetes cluster.local {
      endpoint https://192.168.99.100:8443
      tls /Users/cohaver/.minikube/client.crt /Users/cohaver/.minikube/client.key /Users/cohaver/.minikube/ca.crt
    }
```

If you cannot set up minikube locally, then you can open a PR in the ci repo to uncomment the test (but don't merge yet).  You can then run that integration test on a coredns repo PR by making a comment containing `/integration-ciprXX`, where XX is the number of the CI PR (as I have done in this PR above https://github.com/coredns/coredns/pull/2429#issuecomment-451209278).

miekg:
for backwards incompat changes:
* first announce
* remove code; leave option (don't break configs)
* remove option

See 3 releases in total

miekg:
lovely to see a LOC reduction here ,btw.

yongtang:
Depending on how certificate was issued or the common name field of the certificate, additional implementation work might be needed to have https load balancing.

At the time the feature was implemented we were still using golang 1.9 so there were quite a few limitations for http(s) proxy.

I think reporting error on multiple https endpoints while allowing http endpoints would be the short term fix.

Will take another look and extend the support for http(s).

yongtang:
Added PR #2438 for now.

miekg:
how btw, is a http only endpoint useful? Is that even allowed? I mean, before we even fix this functionality, if you can't really use it now in k8s, we better remove it. (Why? It uses pkg/healthcheck, the only other user is *proxy*, I want to get rid of proxy (and pkg/healthcheck)

miekg:
https://github.com/coredns/coredns/wiki/CoreDNS-1.3.1-Release

We are not/we are removing this in a future version? If we want to do something in an upcoming version it would be good to announce now, otherwise we need to wait another for another release.

yongtang:
@miekg A http only (single) endpoint is legitimate as localhost access through http is common. Multiple http endpoints for HA is less common nowadays, compared with a couple of years ago when that was implemented, as a lot of things has evolved.

Of course, in-cluster deployment is always possible, but in-cluster deployment is not always easy to debug and fix if you hit an issue. For production usage, many choose out-of-cluster deploy for reliability reasons.

For out-of-cluster deploy, from CoreDNS's point of view, it is possible to achieve HA even without multiple endpoints of api servers.

A configuration is to deploy coredns on the same host for every api server. So there is a 1:1 mapping of api server and coredns. Outside of api servers, you could deploy the coredns and set multiple upstream coredns servers that are located on api servers. The HA and load balancing is achieved through coredns itself now.

So multiple k8s endpoints usage is less common. The http endpoint still has its usage.

chrisohaver:
IMO:

**single https endpoint**: useful in conjunction with kubernetai plugin, wherein a coredns instance is connecting to other clusters.

**single http endpoint**: useful for local development and debugging kubernetes, since it's easy to debug a coredns running on on your local system.

**multiple https endpoint** : Currently does not work. Perhaps useful in conjunction with kubernetai plugin, if a remote cluster employs HA.

**multiple http endpoint**: I think unlikely this would get much use outside of sandboxes/POCs

miekg:
>IMO:
>
>**single https endpoint**: useful in conjunction with kubernetai plugin, wherein a coredns instance is connecting to other clusters.
>
>**single http endpoint**: useful for local development and debugging kubernetes, since it's easy to debug a coredns running on on the local host.
>
>**multiple https endpoint** : perhaps useful in conjunction with kubernetai plugin, if a remote cluster employs HA.
>
>**multiple http endpoint**: I think unlikely this would get much use outside of sandboxes/POCs

Seems that we could then remove the code that implements the multiple endpoints 
and noone will be off for the worse? Healthchecking a single endpoint also 
doesn't make a lot of sense, so that can go too? WDYT?

I want to kick out a 1.3.1 soonish and it would be good to get some kind of 
deprecation notice in there. See https://github.com/coredns/coredns/wiki/CoreDNS-1.3.1-Release
for the WIP notes.

miekg:
done in #2454

PR #2430, opened 2019-01-03
plugin/kubernetes: fix case preservation and add test

chrisohaver:

Adds case preservation in k8s plugin, and adds tests for it.

#1403

none

### DIFF
--- a/plugin/kubernetes/handler.go
+++ b/plugin/kubernetes/handler.go
@@ -18,11 +18,12 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 	m.SetReply(r)
 	m.Authoritative = true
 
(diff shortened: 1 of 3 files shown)


chrisohaver:
Regarding the change to the shared function `Matches()`...

I don't know if its best to fix that function to preserve the case (as I have done), or to create a new fuction e.g. `QMatches()`, that preserves the case, and leave the existing function as is.

I suppose the safe play is to create a new function... but I don't want to bloat code unnecessarily.

Opinions?

bradbeam (approved):

miekg:
should be fine. I'm leaning both ways. I think it will be better not to change the Matches function, and make the change in the k8s plugin.

In general we should work with QName() when responding, but we do things slightly different here. Think a helper function in dnsutil (ala) TrimZone but then the reverse makes sense here

PR #2442, opened 2019-01-07
support etcd credentials in etcd plugin

### LINKED ISSUE #2441: Support etcd username and password for etcd plugin
Following the discussion from https://github.com/coredns/coredns/pull/2163
I would like to had the support for username and password in the etcd configuration.

In the PR discussion @miekg said he would prefer to support this from the config file instead of some environment variable.

I can work on that and provide a PR that implement that feature.
Is there any other recommendation regarding this feature ?

About the format of the config file, I was thinking implementing something like
```
etcd [zone]{
   endpoint http://localhost:2379
   credentials username password
}
```
I can also be more explicity and do 
```
etcd [zone]{
   endpoint http://localhost:2379
   username username
   password password
}
```
What do you think ?
```
```

zaibon:
fixes #2441

### DIFF
--- a/plugin/etcd/README.md
+++ b/plugin/etcd/README.md
@@ -31,7 +31,8 @@ etcd [ZONES...] {
     fallthrough [ZONES...]
     path PATH
     endpoint ENDPOINT...
-    upstream
+    credentials USERNAME PASSWORD
+    upstream [ADDRESS...]
     tls CERT KEY CACERT
 }
 ~~~
@@ -42,8 +43,12 @@ etcd [ZONES...] {
   queries for those zones will be subject to fallthrough.
 * **PATH** the path inside etcd. Defaults to "/skydns".
 * **ENDPOINT** the etcd endpoints. Defaults to "http://localhost:2379".
-* `upstream` resolve names found in etcd (think CNAMEs) If you want CoreDNS to act as a proxy for clients,
-  you'll need to add the forward plugin. CoreDNS will resolve CNAMEs against itself.
+* `credentials` is used to set the **USERNAME** and **PASSWORD** for accessing the etcd cluster.
+* `upstream` upstream resolvers to be used resolve external names found in etcd (think CNAMEs)
+  pointing to external names. If you want CoreDNS to act as a proxy for clients, you'll need to add
+  the proxy plugin. If no **ADDRESS** is given, CoreDNS will resolve CNAMEs against itself.
+  **ADDRESS** can be an IP address, and IP:port or a string pointing to a file that is structured
+  as /etc/resolv.conf.
 * `tls` followed by:
 
     * no arguments, if the server certificate is signed by a system-installed CA and no client cert is needed
--- a/plugin/etcd/setup.go
+++ b/plugin/etcd/setup.go
@@ -48,6 +48,8 @@ func etcdParse(c *caddy.Controller) (*Etcd, error) {
 		tlsConfig *tls.Config
 		err       error
 		endpoints = []string{defaultEndpoint}
+		username  string
+		password  string
 	)
 	for c.Next() {
 		etc.Zones = c.RemainingArgs()
@@ -89,6 +91,15 @@ func etcdParse(c *caddy.Controller) (*Etcd, error) {
 					if err != nil {
 						return &Etcd{}, err
 					}
+				case "credentials":
+					args := c.RemainingArgs()
+					if len(args) == 0 {
+						return &Etcd{}, c.ArgErr()
+					}
+					if len(args) != 2 {
+						return &Etcd{}, c.Errf("credentials requires 2 arguments, username and password")
+					}
+					username, password = args[0], args[1]
 				default:
 					if c.Val() != "}" {
 						return &Etcd{}, c.Errf("unknown property '%s'", c.Val())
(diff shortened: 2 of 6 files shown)


plugin/etcd/README.md
```diff
@@ -113,6 +113,15 @@ etcd skydns.local {
 ...
 ~~~
 
+You can connect to an etcd cluster protected by a password by specifiying the credential in the Corefile.
+
+~~~
+etcd skydns.local {
+    endpoint http://localhost:2379
+    credentials username password
```
miekg: can you add this to the initial how-to-use the etcd plugin, with:

~~~
credentials USERNAME PASSWORD
~~~~
and explain it's use in the bullets below?

plugin/etcd/README.md
```diff
@@ -213,4 +222,4 @@ If you query the zone name for `TXT` now, you will get the following response:
 ~~~ sh
 % dig +short skydns.local TXT @localhost
 "this is a random text message."
-~~~
+~~~
```
miekg, line 213: what's this change? Fixing trailing whitespace?

miekg (commented):
is there a test that's easy to add?

plugin/etcd/README.md
zaibon: Yep looks like it

plugin/etcd/README.md
zaibon: Which "how-to-use" section are you talking about?

zaibon:
> is there a test that's easy to add?
Do you have some test about the parsing of the config file already somewhere? I didn't really find one.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] support etcd ..." ]
>> is there a test that's easy to add?
>Do you have some test about the parsing of the config file already somewhere? I didn't really find one.

test/etcd_test.go has tests.

If you're adding some please *don't* use proxy.Lookup as I'm removing all the 
code: new way of doing a lookup: https://github.com/coredns/coredns/pull/2436

tx

plugin/etcd/README.md
miekg: starting on line 30

zaibon: Done

zaibon:
I think the PR is ready now

plugin/etcd/README.md
```diff
@@ -45,6 +46,7 @@ etcd [ZONES...] {
   queries for those zones will be subject to fallthrough.
 * **PATH** the path inside etcd. Defaults to "/skydns".
 * **ENDPOINT** the etcd endpoints. Defaults to "http://localhost:2379".
+* `credentials` If the etcd cluster pointed by `endpoint` is password protectd, use this field to specify the username and password.
```
miekg: protectd -> protected

better will be:
~~~
`credentials` is used to set the **USERNAME** and **PASSWORD** for accessing the etcd cluster.
~~~

plugin/etcd/README.md
```diff
@@ -113,6 +115,7 @@ etcd skydns.local {
 ...
 ~~~
 
+
```
miekg: unintended newline?

plugin/etcd/setup.go
```diff
@@ -103,6 +105,15 @@ func etcdParse(c *caddy.Controller) (*Etcd, bool, error) {
 					if err != nil {
 						return &Etcd{}, false, err
 					}
+				case "credentials":
+					args := c.RemainingArgs()
+					if len(args) == 0 {
+						return &Etcd{}, false, c.ArgErr()
+					}
+					if len(args) != 2 {
+						return &Etcd{}, false, c.Errf("credentials required 2 arguments, username and password")
```
miekg: required -> requires

plugin/etcd/setup_test.go
```diff
@@ -14,43 +14,69 @@ func TestSetupEtcd(t *testing.T) {
...
 		},
 		// negative
 		{
 			`etcd {
 	endpoints localhost:300
 }
-`, true, "", []string{""}, "unknown property 'endpoints'",
+`, true, "", []string{""}, "unknown property 'endpoints'", "", "",
+		},
+		// with valid crendentials
```
miekg: credentials

miekg:
yes, starting to look good, but def. needs a test; I can be sure if this works or not.

In the top-level test/ dir you can copy etcd_cache_test.go and make a minimal etcd test that sets a username and password on the etcd backend (and deletes that again - because the testing etcd is used for more tests) and queries a single record from etcd and sees if it correct.

zaibon:
@miekg can you have another and see if https://github.com/coredns/coredns/pull/2442/files#diff-de7cfcdf070628c0ec52593ecfda6801R12 is what you are expecting for the tests.
I still need to figure out how to disable authentication to clean up the test. Everything Ive tried so far returns an error. If you have some input about that.

miekg:
you mean this code:
~~~// defer func() {
	// 	if _, err := etc2.Client.AuthDisable(ctx); err != nil {
	// 		t.Errorf("Fail to disable authentication: %v", err)
	// 	}
	// 	if _, err := etc.Client.RoleDelete(ctx, "root"); err != nil {
	// 		t.Errorf("Failed to delete root role: %v", err)
	// 	}
	// 	_, err := etc.Client.UserDelete(ctx, username)
	// 	if err != nil {
	// 		t.Errorf("Failed to delete user: %v", err)
	// 	}
	// }()
~~~
What happens?

Otherwise that test looks like the thing I want there

zaibon:
@miekg ok finally all tests are green.
Problem was I tried to remove root user and role, while etcd doesn't allow it.

miekg (approved):

Issue #2434, opened 2019-01-04
Headless service name resolution delay


svallala:
Hi,

     I have following configuration in the coredns configmap.

        .:53 {
            errors
            health
            kubernetes cluster.local 21.172.in-addr.arpa 19.172.in-addr.arpa 18.172.in-addr.arpa {
               pods verified
               fallthrough 21.172.in-addr.arpa 19.172.in-addr.arpa 18.172.in-addr.arpa
            }
            forward . /etc/resolv.conf {
               policy sequential
               health_check 5s
            }
            prometheus :9153
            cache 30
            reload
        }

I was testing the name resolution of a headless service. In the pod entrypoint I was timing how long it takes for name resolution to succeed using nslookup <hostname>, it takes 30 seconds. If I change the config with "cache 5", then it takes 5 seconds. Trying to understand the role of cache plugin here. If my application entrypoint relies on name resolution immediately on start, then should I just get rid of the cache plugin?

What are the drawbacks of removing the cache plugin, even after I remove the line, it takes couple of seconds for the name resolution to work. I am forced to put a sleep in my entrypoint, how long does it take for coredns to resolve a newly created headless service?

Any help is much appreciated.

Thanks
Shravan

chrisohaver:
The cache plugin keeps a local cache in coredns of dns responses.  It's mainly used to keep copies of upstream responses locally to reduce the traffic to upstream servers.  If you remove the cache plugin, the amount of traffic to your upstream dns servers will increase.

`cache 30` for example, will cache responses for a maximum of 30 seconds.  However, if the TTL of a record is < 30 then it will only be cached for a duration equal to its TTL.  

If you queried the headless service before it was created, then the negative response would be cached.  However, the TTL for records served by the kubernetes plugin defaults to 5 seconds, so those negative responses should only be cached by the cache plugin for 5 seconds.

If you set the TTL for kubernetes to zero (i.e. `ttl 0`, which is allowed in the latest version of coredns), then kubernetes records wont enter the cache.

The kubernetes plugin has a kind of built in cache of its own, which keeps indexed records of every service, endpoint, and pod (for pods verified). These records are updated via an asynchronous watch on the k8s API.  So, kubernetes plugin answers very quickly even if none of its records get cached by the cache plugin.

> how long does it take for coredns to resolve a newly created headless service?

[...]

PR #2446, opened 2019-01-08
plugin/kubernetes: Use the plugin's ttl setting for negative responses

chrisohaver:

Currently in the k8s plugin, the TTL of negative answers is hard coded to 30.  In this PR,  we instead use the same TTL in negative answers as in positive answers, which is user configurable by the `ttl` option.

#2434


miekg:
think we do the same in the caching (although you can define a specific TTL for denial answers, but that seems overkill here).

Think the current docs imply the TTL is applied to all answers.

miekg:
in the soa

~~~
test.SOA("ClUsTeR.lOcAl.	30	IN	SOA	ns.dns.ClUsTeR.lOcAl. hostmaster.ClUsTeR.lOcAl. 1499347823 7200 1800 86400 60"),
￼~~~

the last number in the SOA record (`60`) also needs to be set to k.ttl

PR #2452, opened 2019-01-10
plugin/kubernetes: set ttl for ns records

chrisohaver:

For NS record responses, TTL was not set, defaulting to 0.
This PR sets the TTL to `k.ttl` to match other responses.


chrisohaver:
This PR sets the TTL to `k.ttl` to match other responses - however due to the nature of the record (comparatively static compared to service/pod records), perhaps it makes more sense to give this a longer static TTL (as we do for example with the TXT record)?

miekg (approved):

miekg:
needs rebase and conflict resolution before merge @chrisohaver

chrisohaver:
thanks for the nudge

PR #2454, opened 2019-01-11
Deprecate multiple endpoints for out-of-cluster k8s api

yongtang:
### 1. Why is this pull request needed and what does it do?

This fix deprecates endpoints for out-of-cluster k8s api,
The Corefile still takes multiple endpoints though only
the first one is used. A warning is shown if there are
multiple endpoints.

### 2. Which issues (if any) are related?

This is the follow up on #2438

### 3. Which documentation changes (if any) need to be made?

Signed-off-by: Yong Tang <yong.tang.github@outlook.com>

PR #2461, opened 2019-01-13
Deprecate multiple endpoints for out-of-cluster k8s api

miekg:
This fix deprecates endpoints for out-of-cluster k8s api,
The Corefile still takes multiple endpoints though only
the first one is used. A warning is shown if there are
multiple endpoints.

This is #2454 rebased on the latest master.


miekg:
original pr is cleaned up.

PR #2466, opened 2019-01-15
fix tls_servername in cloudflare example

RobM83:
fix tls_servername in cloudflare example


yongtang (approved):
LGTM

PR #2470, opened 2019-01-16
readme: add text on backward incompat changes

miekg:
Document the process of releasing backwards incompatible changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,8 @@ When there is a backwards incompatible change in CoreDNS the following process i
 * Release x.y.z: Announce that in the next release we will make backward incompatible changes.
 * Release x.y+1.0: Make the changes, but allow the old configuration to be parsed. I.e. CoreDNS will
   start from an unchanged Corefile.
-* Release x.y+1.1: Remove the lenient parsing, so CoreDNS will not start.
+* Release x.y+1.1: Remove the lenient parsing, so CoreDNS will not start if those features are still
+  used.
 
 ## Security
 
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ CoreDNS can listen for DNS request coming in over UDP/TCP (go'old DNS), TLS ([RF
 
 Currently CoreDNS is able to:
 
-* Serve zone data from a file; both DNSSEC (NSEC only) and DNS are supported (*file*).
+* Serve zone data from a file; both DNSSEC (NSEC only) and DNS are supported (*file* and *auto*).
 * Retrieve zone data from primaries, i.e., act as a secondary server (AXFR only) (*secondary*).
 * Sign zone data on-the-fly (*dnssec*).
 * Load balancing of responses (*loadbalance*).
@@ -30,7 +30,7 @@ Currently CoreDNS is able to:
 * Caching (*cache*).
 * Use etcd as a backend (replace [SkyDNS](https://github.com/skynetservices/skydns)) (*etcd*).
 * Use k8s (kubernetes) as a backend (*kubernetes*).
-* Serve as a proxy to forward queries to some other (recursive) nameserver (*proxy*, and *forward*).
+* Serve as a proxy to forward queries to some other (recursive) nameserver (*forward*).
 * Provide metrics (by using Prometheus) (*metrics*).
 * Provide query (*log*) and error (*errors*) logging.
 * Support the CH class: `version.bind` and friends (*chaos*).
@@ -52,7 +52,7 @@ Check out the project:
 
 We vendor most (not all!) packages. This is mostly because vendoring isn't a perfect solution (in
 Go). We don't vendor `mholt/caddy` and `miekg/dns` for instance. Using `make` will pull down these
-dependencies and check out the correct version as well.
+dependencies and checks out the correct version as well.
 
 Next just run `make`:
 
@@ -192,6 +192,15 @@ More resources can be found:
 
 Examples for deployment via systemd and other use cases can be found in the [deployment
 repository](https://github.com/coredns/deployment).
+
+## Deprecation Policy
+
+When there is a backwards incompatible change in CoreDNS the following process is followed:
+
+* Release x.y.z: Announce that in the next release we will make backward incompatible changes.
+* Release x.y+1.0: Make the changes, but allow the old configuration to be parsed. I.e. CoreDNS will
+  start from an unchanged Corefile.
+* Release x.y+1.1: Remove the lenient parsing, so CoreDNS will not start.
 
 ## Security
 
--- a/README.md
+++ b/README.md
@@ -197,11 +197,12 @@ repository](https://github.com/coredns/deployment).
 
 When there is a backwards incompatible change in CoreDNS the following process is followed:
 
-* Release x.y.z: Announce that in the next release we will make backward incompatible changes.
-* Release x.y+1.0: Make the changes, but allow the old configuration to be parsed. I.e. CoreDNS will
-  start from an unchanged Corefile.
-* Release x.y+1.1: Remove the lenient parsing, so CoreDNS will not start if those features are still
-  used.
+*  Release x.y.z: Announce that in the next release we will make backward incompatible changes.
+*  Release x.y+1.0: Increase the minor version and set the patch version to 0. Make the changes,
+   but allow the old configuration to be parsed. I.e. CoreDNS will start from an unchanged
+   Corefile.
+*  Release x.y+1.1: Increase the patch version to 1. Remove the lenient parsing, so CoreDNS will
+   not start if those features are still used.
 
 E.g. 1.3.1 announce a change. 1.4.0 a new release with the change but backward compatible config.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -193,6 +193,16 @@ More resources can be found:
 Examples for deployment via systemd and other use cases can be found in the [deployment
 repository](https://github.com/coredns/deployment).
 
+## Deprecation Policy
+
+When there is a backwards incompatible change in CoreDNS the following process is followed:
+
+* Release x.y.z: Announce that in the next release we will make backward incompatible changes.
```
chrisohaver: should ".z" be in the others, or removed here?

README.md
```diff
@@ -193,6 +193,16 @@ More resources can be found:
...
+* Release x.y+1.0: Make the changes, but allow the old configuration to be parsed. I.e. CoreDNS will
+  start from an unchanged Corefile.
+* Release x.y+1.1: Remove the lenient parsing, so CoreDNS will not start if those features are still
```
chrisohaver: is this another +1, for a total of +2.1 from the original x.y.z ?  ...

Or are we going from ignored, to breakage in a minor release increment?

README.md
```diff
@@ -193,6 +193,16 @@ More resources can be found:
 Examples for deployment via systemd and other use cases can be found in the [deployment
 repository](https://github.com/coredns/deployment).
 
+## Deprecation Policy
+
+When there is a backwards incompatible change in CoreDNS the following process is followed:
+
+* Release x.y.z: Announce that in the next release we will make backward incompatible changes.
+* Release x.y+1.0: Make the changes, but allow the old configuration to be parsed. I.e. CoreDNS will
```
chrisohaver: is this adding 1.0 to y?  or adding 1 to x, and 0 to y?

README.md
miekg: why? Its 1.3.4 -> 1.4.0 -> 1.4.1 (maybe an example would clarify)

README.md
miekg: why would add a non-integer when talking about split out sem vers?

README.md
chrisohaver: >why?

well - you specify the patch in one example, and not the others... inconsistency

README.md
chrisohaver: Instead of `x.y+n.n` notation,  maybe instead we can define that x.y.z, is major.minor.patch, then clearly state when major, and minor versions are incremented at each stage.

README.md
chrisohaver: Ah - sorry i was mentally parsing this out as... `(x.y) + (1.0)`

README.md
chrisohaver: NM, i was confused by the notation.

stp-ip:
lgtm

fturib:
For the downstream products that are integrating CoreDNS, I am a bit worry about the high speed of integration of a backward incompatibility.

Current proposition does not provide room between the 2 last steps:
- x.(y+1).0 : the change is effective, but we ensure the config is still readable.
- x.(y+1).1 : the config is now incompatible.

If we need to stick a release for downstream product (I am thinking Kubernetes here) that includes `x.(y+1).0` changes but MUST NOT include the incompatibility because that could not be announced yet in a former Kubernetes release, there is no way to have an "in-between" release before `x.(y+1).1`.

**I am thinking of 2 uses cases:**
1. While working on CoreDNS for v1.14, **if CoreDNS 1.4.0 is already released, and we need to fix an important bug found on k8s side** .. . we will need to add an extra release 1.4.1 that does not include the backward incompatibility
1. **Security patch : imagine a vulnerability found on 1.4.0 that is embedded in K8s v1.14.** We will have to release a patch without the backward incompatibility. If 1.4.1 is already out with incompatibility, we have no proper way to release that security patch.


**Could we report the "backward incompatible" to next minor release (? (instead of next patch release)** 
And have the following rule:

- x.y.z : announce the future incompatibility (no change in the code)
- x.(y+1).0 : the change is effective, but we ensure the config is still readable.
- **x.(y+2).0 : the config is now incompatible.**

Such way, we still have room for quick patch release that does not embed the "incompatible" part.

@miekg , @johnbelamaric   what do you think ?

*ASIDE REMARK:* I guess that whatever the way we number the versions, as long as we introduce backward incompatibility, because downstream products release are slower, and because a vulnerability can be disclosed at anytime, **we will have one day to open maintenance branches for CoreDNS**.

johnbelamaric:
So, we use the minor version bump to indicate a deprecation, then the
second bump actually makes it incompatible. I think that's better - in that
it reduces the need for maintenance branches. If we want to avoid that as
much as possible, this is the way to go.

I hear you that we likely will need maintenance branches at some point, but
this will reduce the frequency and perhaps delay it for quite some time.

John

fturib:
@johnbelamaric : YES, good summary. And also good catch on my error about "next major version".  I meant "next minor version" I fixed directly by editing my comment above.

johnbelamaric:
@miekg any thoughts on changing it to deprecate in minor version +1, remove in minor version +2 (rather than minor version +1 patch version +1)? Before merging...

chrisohaver:
> So, we use the minor version bump to indicate a deprecation, then the
second bump actually makes it incompatible.

The PR is saying we will do a patch bump from x.x.0 to x.x.1 to make it a breaking backward incompatible change.  IMO, it should be a second minor bump.

miekg:
Then there is no use what's so ever to do this in 2 steps. Works for me,
because way less work to track this (which is sadly manual).

chrisohaver:
> Then there is no use what's so ever to do this in 2 steps.

This (doing it in 1 step) could result in a very short deprecation notice. e.g. If we announce a deprecation in 2.3.5 and the next release is 2.4.0, that gives users only one release notice (which is IMO short).

We can do it in one step, but I think we should have a fair period of time that a feature is deprecated before removing it. This can be an amount of calendar time, a set number of releases, or by other means, as long as users get a fair chance to see the deprecation.

miekg:
Odd reasoning. Why would it give users more time? To do what?

stp-ip:
A few points and the reasoning for my lgtm:
* deprecation policy was necessary to give user specific expectations
* breaking things is still mostly a no-do
* if someone can't upgrade in 1-2 months, they probably won't be able to in 2-3 months

It's mostly about maintenance burden and Kubernetes is lacking behind anyway. So if we take the stance, that we have deprecations synced more with K8s that would mean way longer deprecation notices of 6-18months.

For example we are still using proxy in K8s instead of forward, even so it's the "better" solution from current viewpoint.

So TL;DR:
Set user expectations low with current policy, try not to need the policy, if we do we can still discuss on the best way to handle something in terms of related features and code path and specific deprecation.

fturib:
@stp-ip : thanks for your clarification.

I kind of agree with you on the 3 points .. 
but as an integrator of CoreDNS in K8s I feel to be also a user. 

And my cycle for the release of k8s is 3 months. 
So, what I would like - as a user-integrator:
- knowing at beginning of cycle what will be the incompatibilities (so I can organize the upgrade path)
- be able to integrate in the code the changes to fix bugs (at least on k8s plugin) up to the code freeze  (2 weeks before end of cycle).


So maybe you are right, we can workaround that by internal discuss and organization about code path ...

 
By the way, the move from `proxy` to `forward` is on our agenda for the next release of k8s (v1.14).

miekg:
You can never be "held hostage" by your downstream. We release on our own
accord, k8s should fix the monolithic release process they have right now

chrisohaver:
> For example we are still using proxy in K8s instead of forward, even so it's the "better" solution from current viewpoint.

The plan has been to move to forward as soon as is was "stable" (subjective perhaps). We tried to make the move before coredns became default in k8s, but aborted due to some forward bugs raised at the time (i.e. not stable). We plan migrate to it in k8s 1.14.

fturib:
> You can never be "held hostage" by your downstream. We release on our own
accord,

@miekg : I agree with that point. That is why if we define a delay between announce and implementation, then any downstream product can check if its release cycle is compatible with this CoreDNS deprecation policy.

Looks like k8s guys missed to validate that point before accepting CoreDNS as default. But I guess it would be nice for all if we can manage that our policy is compatible with their release cycle.

chrisohaver:
If we have to, we could always hold the coredns version back in k8s.  For example, we may opt to deploy coredns 1.3.X in k8s 1.14 and announce deprecation of `upstream` in k8s 1.14 release notes, to avoid pulling the rug out from uses who happen to be using it.  Negatively, that would restrict which new features we can announce in k8s 1.14.  For security issues, we'd need to evaluate if we need to back-port security fixes to 1.3.X.  

All of that is doable.  But I think that the longer we keep features in a state of announced deprecation before removing them, the less likely we would need to hold back the version of coredns in k8s, and therefore more likely to be able to deploy the latest features, and less likely to have to back-port security fixes.

chrisohaver:
@miekg, ok to merge? (just confirming)

PR #2473, opened 2019-01-16
plugin hosts: seeking feedback on a feature allowing to use hashes instead of FQDN

thomas-mangin:
Hello,

As an ISP, we started using CoreDNS as DNS cache for some of our customers. We also provide a content filtering solution to a few thousands of UK schools.

Our solution normally involves explicit and transparent proxies (80 & 443), but when it is not possible, we currently use the host plugin to prevent school children to access some "unsavvy content" - you would be surprised at the wild range of interest teenagers can display :-)

I also want to be able to filter some "sensitive" hosts and point them when a proxy to perform L7 filtering.  For operational reasons, quite a few people have access to the content of CoreDNS.

As the content of the list is quite sensitive (children pornography and terrorist sites), I implemented a quick proof-of-concept allowing to use hashes of the domain names instead of the real name. It seems to work as expected but has not seen operational deployment.

As I never contributed to your projects, I would like your feedback on:
 * if this feature could be considered for upstreaming, so I do not have to maintain the code out-of-tree
 * any test change you would like to see to consider the patch

Should you have no objection with the feature, I would happily update the README.md of the plugin to reflect the new features (asking a native English speaker to do so tho :p).

Also AFACIS the plugin reloads its data every 5 seconds, which is not documented.

Thank you,

Thomas

### DIFF
--- a/plugin/hosts/hosts.go
+++ b/plugin/hosts/hosts.go
@@ -43,13 +43,13 @@ func (h Hosts) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 			// If this doesn't match we need to fall through regardless of h.Fallthrough
 			return plugin.NextOrFailure(h.Name(), h.Next, ctx, w, r)
 		}
-		answers = h.ptr(qname, names)
+		answers = h.ptr(qname, h.hmap.options.ttl, names)
 	case dns.TypeA:
 		ips := h.LookupStaticHostV4(qname)
-		answers = a(qname, ips)
+		answers = a(qname, h.hmap.options.ttl, ips)
 	case dns.TypeAAAA:
 		ips := h.LookupStaticHostV6(qname)
-		answers = aaaa(qname, ips)
+		answers = aaaa(qname, h.hmap.options.ttl, ips)
 	}
 
 	if len(answers) == 0 {
@@ -96,12 +96,12 @@ func (h Hosts) otherRecordsExist(qtype uint16, qname string) bool {
 func (h Hosts) Name() string { return "hosts" }
 
 // a takes a slice of net.IPs and returns a slice of A RRs.
-func a(zone string, ips []net.IP) []dns.RR {
+func a(zone string, ttl uint32, ips []net.IP) []dns.RR {
 	answers := []dns.RR{}
 	for _, ip := range ips {
 		r := new(dns.A)
 		r.Hdr = dns.RR_Header{Name: zone, Rrtype: dns.TypeA,
-			Class: dns.ClassINET, Ttl: 3600}
+			Class: dns.ClassINET, Ttl: ttl}
 		r.A = ip
 		answers = append(answers, r)
 	}
@@ -109,12 +109,12 @@ func a(zone string, ips []net.IP) []dns.RR {
 }
 
 // aaaa takes a slice of net.IPs and returns a slice of AAAA RRs.
-func aaaa(zone string, ips []net.IP) []dns.RR {
+func aaaa(zone string, ttl uint32, ips []net.IP) []dns.RR {
 	answers := []dns.RR{}
 	for _, ip := range ips {
 		r := new(dns.AAAA)
 		r.Hdr = dns.RR_Header{Name: zone, Rrtype: dns.TypeAAAA,
-			Class: dns.ClassINET, Ttl: 3600}
+			Class: dns.ClassINET, Ttl: ttl}
 		r.AAAA = ip
 		answers = append(answers, r)
 	}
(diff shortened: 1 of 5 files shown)


miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin hosts: see..." ]
> * if this feature could be considered for upstreaming, so I do not have to maintain the code out-of-tree

I'm not sure. Seems like something useful, but otoh it's only hashing; this is 
either something you want to encrypt on disk or use a policy engine to that 
queries something off-path to let a request through. Also the general concept of 
DNS filtering is often easily circumvented.

> * any test change you would like to see to consider the patch

haven't look yet.

>Also AFACIS the plugin reloads its data every 5 seconds, which is not documented.

that' true. Happy to merge a PR for that.

thomas-mangin:
@miekg it is not easily circumvented if your are the ISP and DNAT port 53 to your own servers :-)

I was trying to not get into too much details (It quickly become political and not technical) but the list of hostname we are filtering are the IWF CAI list -a list of sites with explicit child abuse images - and the CTIRU (Counter Terrorism Incident Response Unit) list, which are sites promoting terrorism. 

We have some legal requirements to make sure the content of these list are secure and having the hostname on all our caching DNS "in clear" seems "careless" as caching DNS are available to all our customers and I have to assume that we can not be sure that they could not be compromised.

Prior research have shown that you can reverse engineer the list but it required a copy of the whole zones file, which AFAIK is not something readily available. Not impossible but still better than nothing.

A auto-generated file with the hashes pushed a few times a day seems like a good technical compromise to me, the alternatives being overly complex and not prone to high availability.

fturib:
@thomas-mangin : FYI, there is currently a new plugin `Firewall` in review here : #2335.
The purpose is to filter the queries.

If is based on a filtering by expression. In case that your filtering is based on external information, this plugins propose an API in order to plug some external Policy Engine for the filtering (it can be a combination of both).

I would like to understand better your PoC:

1- I understand that your "authorization" is based on a blacklist and/or whitelist of domains.
2- by "allowing to use hashes of the domain names instead of the real name" -> do you mean you can by pass the filter if the domain is queried by its has ?

Is there a way you can share the Corefile using your PoC ?

yongtang:
@thomas-mangin I think that largely depending on the goal.
- If you are only looking for a way to hide the 'hosts' file's plain text domain names, then you should try encryption. It might be computational intensive but you could load the unencrypted content into the memory (so to avoid encryption).
- If you are looking for improving performance of hosts plugin when you have a very big black list set, then you may have better choices with a` cheap hash`.
  - For example, `crc32c` is part of many intel CPU's instruction set and that likely will give you ~1 cycles per 8 bytes (if I remember correctly). Even without crc32c, the normal `crc32` with software implementation could give you 2 cycles per bytes. That is much better throughput than md5/sha256/etc.
  - A cheap hash does not necessary equal more collusion. You could also try Bloom Filter which you could control the size.
- In any case if you want to block purely based on the hash then you have to consider the collision rate first.

thomas-mangin:
Answering the comments:

@fturib firewall is a nice plugin and I will welcome its inclusion but will not help with my use case just right now.

proof of concept (hand-made here and now)

dns cache
```
.:53 {
   forward . 100.64.0.1:53
   cache
}
```

filtering dns on 100.64.0.1
```
. {
.:53 {
	hosts sha1 {
		<ip of proxy>	9ce6e92eb0a2ff46b26a9fb4e403132cf247b3bb
		fallthrough
	}

	forward . <ip of unbound>:53
}
```
(this hash is `www.free.fr.` if you want to test the patch and the data would be in a file)

@yongtang I can not see a use for a Bloom Filter here (or HyperLogLog). The CPU overhead of SHA is noise, I have no scalability issue. I just want the host name to be obfuscated, the hashing is not for performance but obfuscation.

Collision is not an issue, as if we "collide" the only impact will be to redirect the HTTP/HTTPS connection to a proxy which will perform L7 inspection of the HTTP request and let the traffic through if it was not the the intended host.

DoH discussion are left for another day :p

thomas-mangin:
After rework, the impact of the patch for people not using it is now a simple integer test when the hostname can't be found in the hosts map. It has no impact if the host is present (but some code refactoring which cause an extra function call, perhaps mitigated by moving the normalisation of the hostname outside the locked section), so even if the feature is not widely used, if included it should not affect at all the performance of the plugin for other users.

As only one host plugin can not be used per section, the feature does not prevent the use of "normal" hostname, so both hashes and host can be used in the file/section. It however prevents the creation of the reverse entries (as hash reverse would make no sense, and in my case allow all the hashes to leak on PTR request). If this is an issue, the code can be change to generate reverse for normal hosts.

I also forgot to say that it also adds an option "no-autoreverse" to offer to not automatically generate reverse entries. This could be split into another patch if this feature is rejected but the no-reverse useful.

thomas-mangin:
added an option to allow to change the default 5s reload time (ie: "reload 10s"), allowed the configuration options to be used "inline". fixed an issue with how the configuration is store.

thomas-mangin:
Added TTL option, fixed the test suite, and at this point some feedback on the patch would be welcomed when people can find time.

miekg:
Coming back to this. DNS data is inherently public, so obfuscation will not
work (I can just throw a copy of the .com zone at it). Considering this,
and the non standard use of a hosts file I don't think this should live in
coredns as a standard supported feature.

I would suggest just making a new plugin that you use in your support, and
possibly publishing it on coredns.io/explugins

Thanks!

thomas-mangin:
Thank you for coming back to me with a position.

Issue #2472, opened 2019-01-16
Update README.md on how to compile from GitHub


thomas-mangin:
The main README.md is incomplete. Compilation can not be achieved by cloning the repo and running make (tested on OSX and Ubuntu bionic).

My solution was to run the following (which is surely what the instruction should contain too):

```
[ -z "$GOPATH" ] && export GOPATH=`pwd`
mkdir -p $GOPATH/src/github.com/coredns
cd $GOPATH/src/github.com/coredns
git clone https://github.com/coredns/coredns
cd coredns
make
```

The issue is discussed in https://github.com/coredns/coredns/issues/1693 and https://github.com/coredns/coredns/issues/1070 but did not lead to a change in the documentation

"Naively trying to compile the code fails:
```
> git clone https://github.com/coredns/coredns
> cd coredns
> make
** presubmit/context
** presubmit/filename-hyphen
** presubmit/test-lowercase
** presubmit/trailing-whitespace
(cd /Users/thomas/Library/Go/src/github.com/mholt/caddy 2>/dev/null              && git checkout -q master 2>/dev/null || true)
(cd /Users/thomas/Library/Go/src/github.com/miekg/dns 2>/dev/null                && git checkout -q master 2>/dev/null || true)
(cd /Users/thomas/Library/Go/src/github.com/prometheus/client_golang 2>/dev/null && git checkout -q master 2>/dev/null || true)
go get -u github.com/mholt/caddy
go get -u github.com/miekg/dns
go get -u github.com/prometheus/client_golang/prometheus/promhttp
go get -u github.com/prometheus/client_golang/prometheus
(cd /Users/thomas/Library/Go/src/github.com/mholt/caddy              && git checkout -q v0.11.1)
(cd /Users/thomas/Library/Go/src/github.com/miekg/dns                && git checkout -q v1.1.3)
(cd /Users/thomas/Library/Go/src/github.com/prometheus/client_golang && git checkout -q v0.9.1)
CGO_ENABLED=0  go build -v -ldflags="-s -w -X github.com/coredns/coredns/coremain.GitCommit=b0d37c54-dirty" -o coredns
coredns.go:9:2: cannot find package "github.com/coredns/coredns/core/plugin" in any of:
	/usr/local/Cellar/go/1.11.4/libexec/src/github.com/coredns/coredns/core/plugin (from $GOROOT)
	/Users/thomas/Library/Go/src/github.com/coredns/coredns/core/plugin (from $GOPATH)
[...]

PR #2477, opened 2019-01-18
better instruction on how to compile coredns

thomas-mangin:

Give steps by steps instructions on how to compile coredns

https://github.com/coredns/coredns/issues/2472

See patch

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,15 +44,15 @@ out-of-tree plugins.
 
 ## Compilation from Source
 
-Check out the project:
+Check out the project in your `$GOPATH` or set `$GOPATH` to be the folder where you want to work in
 
 ~~~
+% [ -z "$GOPATH" ] && export GOPATH=`pwd`
+% mkdir -p $GOPATH/src/github.com/coredns
+% cd $GOPATH/src/github.com/coredns
 % git clone https://github.com/coredns/coredns
+% cd coredns
 ~~~
-
-We vendor most (not all!) packages. This is mostly because vendoring isn't a perfect solution (in
-Go). We don't vendor `mholt/caddy` and `miekg/dns` for instance. Using `make` will pull down these
-dependencies and check out the correct version as well.
 
 Next just run `make`:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,16 +44,16 @@ out-of-tree plugins.
 
 ## Compilation from Source
 
-Check out the project:
+Check out the project in your `$GOPATH` or set `$GOPATH` to be the folder where you want to work in
 
 ~~~
+% [ -z "$GOPATH" ] && export GOPATH=`pwd`
```
jameshartig, line 50: This won't work on Windows where `export` is not defined. Also, this is going to create a bunch of folders under the current directory, when they might've wanted to just checkout in this directory.

Instead maybe a block after this that says "If you get `cannot find package` errors when running `make`, ensure that your checked out copy of coredns is available at `$GOPATH/src/github.com/coredns`. You may need to set `$GOPATH` if its not set already. See https://github.com/golang/go/wiki/GOPATH for more information.".

thomas-mangin: Hi James, I am happy with whatever you believe is best. The instructions as they stand just need a little bit of TLC, your wording would have given me the information I needed when make failed the first time. You can add this as easily as I - should I update the patch or can you just edit the file yourself ?

thomas-mangin: Trying to get this moving (one way or another):
* The new instructions may not work for windows, but then the current ones will not neither. Having not used windows since Windows XP, I will let someone else help them !
* Telling Unix users who are unfamiliar something about `$GOPATH` may point them in the right direction but does not provide the steps required before running `make` to "yield a `coredns` binary"
* For people attempting to use `$CWD`, if the user is savvy enough to have cloned CoreDNS within `$GOPATH/src/coredns` folder, I believe the current instructions will be understood (as they likely needed no instructions in the first place)

So unless some better step-by-step instructions can be found, or some change made to help windows users (which I can not provide having not used Windows since XP), I still believe the current patch is a step in the right direction and is worth merging.

miekg: Maybe just use `$GOPATH` and refer to some official GOPATH docs from the Go project otherwise. Setting GOPATH to pwd is a bit of a no-no, so don't do that.

thomas-mangin: As it is 5 lines of wording change and we spent more time discussing it in here than just having I changed can someone with commit access to README.md and authority to do so please update the page. I am closing the PR.

PR #2478, opened 2019-01-18
plugin/hosts reload the hosts file every 5 seconds

thomas-mangin:

It informs users that the hosts file is updated every 5 seconds


plugin/hosts/README.md
```diff
@@ -95,6 +95,10 @@ Load hosts file inlined in Corefile.
 }
 ~~~
 
+## Auto-update
```
miekg: Could you move this into the description above as new paragraph above "This plugin can only be used once per Server Block." ?

miekg: otherwise lgtm

thomas-mangin: done.

miekg (approved):

PR #2479, opened 2019-01-19
Stop importing testing in the main binary

miekg:
Stop importing "testing" into the main binary:
* test/helpers.go imported it; remote that and change function signature
* update all tests that use this


miekg:
only mechanical changes.

Issue #2484, opened 2019-01-21
plugin/rewrite: initial request is modified with OPT records


fturib:
### Context
`CoreDNS v 1.3.1`
`Corefile:`
```
.:53 {
                rewrite edns0 subnet append 24 56
                forward . 10.82.3.15:53
                log
                errors
}
```

### After analysis of the code (credits to @rdrozhdzh)

> I see that rewrite plugin creates OPT record in original request in case if the OPT record was absent from user query. > (https://github.com/coredns/coredns/blob/3533065ed20a7fe14cc00968e3fa8c1beff072b2/plugin/rewrite/edns0.go#L42-L49)

> Thus, when coredns gets response it checks in `func (r *Request) SizeAndDo(m *dns.Msg)` if original query had OPT records. However, since rewrite plugin has modified original query, then SizeAndDo() doesn't know what was in original user query. It supposes that the OPT record with UPD bufsize options came from user.

> Perhaps, rewrite plugin should create a copy of original query before adding OPT record. 

> Also, pay attention, that if coredns responds with OPT record to user query which didn't have OPT RR - this is a violation of RCF https://tools.ietf.org/html/rfc6891

### Initial symptoms (FYI)

CoreDNS returns a packet larger that the regular UDP size in our case
`dig -t TXT <fqdn> +short @<my-resolver-ip> +noedns`

- Client sends query to  CoreDNS  without any EDNS option  41  (i.e.  expect to use  packet size of 512 bytes) 
- CoreDNS inserts  ECS  EDNS option and EDNS option 41 with packet size of 4096 bytes  
- The downstream DNS server responds with  a packet greater than  512 bytes   which eventually forwarded to the client.

fturib:
As it seems simple to modify the initial request, but we need it for replies, and checks ...

Would it be better to keep aside a copy of that initial request (or at least the element that allow to send back properly the response to the client) ?

fturib:
the plugin Rewrite implements a `ResponseReverter` that could be enhanced somehow to ensure to revert also the EXTRA records when needed.
(it is already triggered by a boolean `needRevert`)

rdrozhdzh:
@fturib, I like your idea with `ResponseReverter`. This way we can only store a copy of original `OPT `record without copying entire DNS query message. And `ResponseReverter` will compare the response `OPT` record to original query record.

[...]

PR #2486, opened 2019-01-22
SizeAndDo() + rewrite - do not modify initial query, and do not response OPT if query does not ask to.

fturib:
### 1. why does it fix and why is it needed
- EDNS0 records are replied to a non EDNS0 client, when it should not
- effect is that the size of the packet can exceed default expectation of UDP client (512)

Explanation of fixe:
- in SizeAndDo() : remove OPT record from response, if the request has no OPT record at all
- in Rewrite ServeDNS() - copy the initial request before modifying it (be sure that you do not modify the OPT record of initial query).

NOTE: During investigation/test, I found that the Msg.Copy() was not complete for OPT records.
This PR requires the merge of this miekg/dns PR https://github.com/miekg/dns/pull/902.
 
#2484 

NONE - that is a fix of invalid behavior

### DIFF
--- a/plugin/rewrite/rewrite.go
+++ b/plugin/rewrite/rewrite.go
@@ -38,6 +38,9 @@ type Rewrite struct {
 
 // ServeDNS implements the plugin.Handler interface.
 func (rw Rewrite) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+
+	// ensure r is a copy in order to not pollute the initial request
+	r = r.Copy()
 	wr := NewResponseReverter(w, r)
 	state := request.Request{W: w, Req: r}
 
--- a/request/request.go
+++ b/request/request.go
@@ -196,6 +196,15 @@ func (r *Request) Size() int {
 func (r *Request) SizeAndDo(m *dns.Msg) bool {
 	o := r.Req.IsEdns0()
 	if o == nil {
+		// ensure remove OPT if initial query was not Edns0 - compliance with https://tools.ietf.org/html/rfc6891#page-10
+		for i := len(m.Extra) - 1; i >= 0; i-- {
+			if m.Extra[i].Header().Rrtype == dns.TypeOPT {
+				m.Extra[i] = m.Extra[len(m.Extra)-1]
+				m.Extra[len(m.Extra)-1] = nil
+				m.Extra = m.Extra[:len(m.Extra)-1]
+				break
+			}
+		}
 		return false
 	}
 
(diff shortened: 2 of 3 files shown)


plugin/rewrite/rewrite.go
```diff
@@ -38,6 +38,9 @@ type Rewrite struct {
 
 // ServeDNS implements the plugin.Handler interface.
 func (rw Rewrite) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+
+	// ensure r is a copy in order to not pollute the initial request
+	r = r.Copy()
```
fturib, line 43: We should not modify the initial query.
But creating  a Copy() for each query may be heavy (CPU / Memory).

Should we Copy ONLY when there will be a change on OPT ?

request/request.go
```diff
@@ -196,6 +196,15 @@ func (r *Request) Size() int {
 func (r *Request) SizeAndDo(m *dns.Msg) bool {
 	o := r.Req.IsEdns0()
 	if o == nil {
+		// ensure remove OPT if initial query was not Edns0 - compliance with https://tools.ietf.org/html/rfc6891#page-10
+		for i := len(m.Extra) - 1; i >= 0; i-- {
+			if m.Extra[i].Header().Rrtype == dns.TypeOPT {
+				m.Extra[i] = m.Extra[len(m.Extra)-1]
```
jameshartig: This will reorder the extra. I'm not sure if that violates any rules but typically the extra is ordered to match the answers.

jameshartig (commented):
Where are the EDNS0 options coming from in the response? Is rewrite adding them because you had edns0 rewrite rules? Should that check IsEdns0() before adding them instead?

fturib:
NOTE: test is failing, as expected, until PR https://github.com/miekg/dns/pull/902 is merged:  until then, Msg.Copy() is incomplete for OPT records and therefore, the copy of msg in Rewrite is ineffective.

```
rewrite_test.go:86: Expected EDNS_LOCAL data 'initial-data' but got "hello-world"
```

fturib:
@fastest963 : Yes, in the case I know, the EDNS0 were introduced by the `rewrite`.
`rewrite` aims to modify the request before resolution : upstream. So I think it is in purpose to add the the EDNS0 whatever the initial request. the EDNS0 option will help the resolution upstream. Once resolved, we want to return the response to the client, but at that point the EDNS0 is not more useful.

request/request.go
fturib: I am not aware of detail here. I can manage to keep the order.

rdrozhdzh:
see my comment https://github.com/coredns/coredns/issues/2484#issuecomment-456401325

fturib:
On suggestion of @rdrozhdzh , I will propose another implementation that involved ONLY the REWRITE plugin.

fturib:
this PR is replaced by this one : #2491

PR #2489, opened 2019-01-22
Making the hosts plugin more configurable

thomas-mangin:

This pull request is not needed. I happened to have worked on the hosts plugin and thought some of the improvement developed may be beneficial to the coredns community.

The patch allows the host plugin to be customised more with the following option:
 * reload \<duration\>|disabled (inline only atm)
   allows to change the reloading interval (default 5s)
 * ttl \<time in sec\> (inline only atm)
   allows to change the default TTL of the hosts and reverse records (default 5 minutes)
 * no-reverse (both as first argument on the plugin line and inline)
   disable the automatic generation of reverse entries for hosts defined in the file

within the patch is also a bug fix for the parsing of IPv4 with IPv6 ( ie: `::FFFF:127.0.0.1`) as the current code will consider this IP as an IPv6 IP when it is really an IPv4 one encoded using IPv6 syntax.

N/A

If the feature is considered for inclusion, the documentation will need to be updated to reflect the new features.

Feedback on the patch and word required to see it accepted welcomed. If this not of interest, please feel free to simply let me know and close the PR.


miekg:
> within the patch is also a bug fix for the parsing of IPv4 with IPv6 ( ie: ::FFFF:127.0.0.1) as the current code will consider this IP as an IPv6 IP when it is really an IPv4 one encoded using IPv6 syntax.

might be worth pulling out into a separate PR.

Your proposed changes should good. On disabling the reload, we usually just use 0 (prolly 0s as we want this to be a duration)

thomas-mangin:
I will break the fix into another PR.

plugin/hosts/hosts_test.go.orig
```diff
@@ -0,0 +1,79 @@
+package hosts
```
pmoroney: Should this file be included?

plugin/hosts/hostsfile.go
```diff
@@ -46,6 +66,9 @@ type hostsMap struct {
 	byAddr map[string][]string
 }
 
+var durationOf0s, _ = time.ParseDuration("0s")
```
pmoroney: nit: These can be constants instead of variables

```go
const durationOf0s = 0 * time.Second
const durationOf5s = 5 * time.Second
```

pmoroney (commented):
Thanks for these changes, they look awesome!
Besides splitting out the other PR, I just have one nit.

plugin/hosts/hostsfile.go
thomas-mangin: I thought it was a good idea, so I attempted to change the code but we use the the address of the variable in the code .. I could change the struct to not use a pointer. Not sure what is best practice.

plugin/hosts/hosts_test.go.orig
thomas-mangin: Thank you, no, it should not and was removed.

plugin/hosts/hostsfile.go
thomas-mangin: @pmoroney happy with a pointer or should it be changed to just a time.Duration ?

miekg:
Duration of zero is just 0, no?

thomas-mangin:
@miekg Yes, Duration is an type alias for int64 .. updated the code to not use parseDuration:
```
-var durationOf0s, _ = time.ParseDuration("0s")
-var durationOf5s, _ = time.ParseDuration("5s")
+var durationOf0s = time.Duration(0)
+var durationOf5s = time.Duration(5 * time.Second)
```

plugin/hosts/hostsfile.go
thomas-mangin: removed the pointer.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Making the ho..." ]
>@miekg Yes, Duration is an type alias for int64 .. updated the code to not use parseDuration:
>```
>-var durationOf0s, _ = time.ParseDuration("0s")
>-var durationOf5s, _ = time.ParseDuration("5s")
>+var durationOf0s = time.Duration(0)
>+var durationOf5s = time.Duration(5 * time.Second)
>```

this can just be 0, and in fact the can all be constants.

```
const (
    durationOf0s = 0
    durationOf5s = 5 * time.Second
)

```

thomas-mangin:
@miekg - yes, this is what the next patch did already (use constants and not use pointers for Duration). Sorry I should have updated my comment - me bad !

thomas-mangin:
Anything I can do to move this PR forward ?

pmoroney:
The changes look good to me, but I am now concerned about those extra commits in the PR, are those from another branch? Are they supposed to go into master?

miekg:
> The changes look good to me, but I am now concerned about those extra commits in the PR, are those from another branch? Are they supposed to go into master?

good call, these should be removed from this PR. Either squash the entire pr into one commit (a git rebase -i) or get the diff and re-do the PR with some cleaner history (close this one, do a new one and we'll merge that)

Issue #62, opened 2016-03-30
middleware file: wildcards


miekg:
implement wildcard responses for both DNS and DNSSEC.

miekg:
positive replies work. Now only nodata responses.

luk4hn:
Hi @miekg,
I have a wildcard entry in my zone file, like below: 

> *.example.org.                           300     IN A       1.2.3.4

With this config only query for subdomains start with letter from a to s (example: a.example.org, b.example.org, c.example.org, ...) can get correct response (ip: 1.2.3.4).
Query for subdomains start with letter from t, u, v ... y, z (ex: t.example.org, ta.example.org, taa.example.org, ....) resulted in no answer received.

Do you have any idea on how to make wildcard entry work correctly ? Thanks

miekg:
[ Quoting notifications@github.com in "Re: [miekg/coredns] middleware file..." ]

> Hi @miekg,
> I have a wildcard entry in my zone file, like below:
> 
> > *.example.org.                           300     IN A       1.2.3.4
> 
> With this config only query for subdomains start with letter from a to s (example: a.example.org, b.example.org, c.example.org, ...) can get correct response (ip: 1.2.3.4).
> Query for subdomains start with letter from t, u, v ... y, z (ex: t.example.org, ta.example.org, taa.example.org, ....) resulted in no answer received.
> 
> Do you have any idea on how to make wildcard entry work correctly ? Thanks

WAT?

Are you running CoreDNS from the git version? Can you post your zone + config
(or email me), I'll take a look.

/Miek

## 

Miek Gieben

luk4hn:
I'm using release v002 (coredns_002_linux_x86_64.tgz). 
My config is like below
- Corefile 

```
.:53 {
  rewrite ANY HINFO

  file records/example.org example.org

  proxy . 8.8.4.4:53 8.8.8.8:53 {
    policy least_conn
    fail_timeout 30s
    max_fails 1
    except example.org
    spray
  }

  cache 60 example.org
  errors stderr
  log stdout
}
```
- zone file:

```
$ORIGIN .
@       3600    IN      SOA     example.org.    root.example.org.       (
                2016102001      ; serial
                7200            ; refresh
                3600            ; retry
                86400           ; expire
                3600)           ; minimum

; NS
example.org.                             300     IN      NS    ns1.example.org.
example.org.                             300     IN      NS    ns2.example.org.

; CNAME

; A
*.example.org.                           300     IN A       172.46.0.15

[...]

PR #2491, opened 2019-01-23
plugin/rewrite - handle properly OPT records 

fturib:
### 1. why does it fix and why is it needed
- EDNS0 records are replied to a non EDNS0 client, when it should not
- effect is that the size of the packet can exceed default expectation of UDP client (512)

Explanation of fix:
- REWRITE may modify the OPT of the request : ensure to revert before writing the response
- if the initial request has no OPT, ensure the response has no OPT
- DO NOT fill the response with modified requests's OPT : if no OPT is returned in the response, the initial client will get the OPT from its own initial request (set by SizeAndDo) 

NOTE: During investigation/test, I found that the Msg.Copy() was not complete for OPT records.
This PR requires the merge of this miekg/dns PR https://github.com/miekg/dns/pull/902.
**1/25/2019 - the PR from miekg/dns is not more required**
 
#2484 

NONE - that is a fix of invalid behavior
enhanced the rewrite_test to show that a request with no OPT will not receive any OPT response.

### DIFF
--- a/plugin/rewrite/reverter.go
+++ b/plugin/rewrite/reverter.go
@@ -1,10 +1,11 @@
 package rewrite
 
 import (
-	"github.com/miekg/dns"
 	"regexp"
 	"strconv"
 	"strings"
+
+	"github.com/miekg/dns"
 )
 
 // ResponseRule contains a rule to rewrite a response with.
@@ -21,24 +22,61 @@ type ResponseRule struct {
 // dig will complain with ';; Question section mismatch: got example.org/HINFO/IN'
 type ResponseReverter struct {
 	dns.ResponseWriter
-	originalQuestion dns.Question
-	ResponseRewrite  bool
-	ResponseRules    []ResponseRule
+	receivedRequest *dns.Msg
+	sharedRequest   *dns.Msg
+	ResponseRewrite bool
+	ResponseRules   []ResponseRule
+}
+
+func removeOPT(m *dns.Msg) {
+	for i := len(m.Extra) - 1; i >= 0; i-- {
+		if m.Extra[i].Header().Rrtype == dns.TypeOPT {
+			copy(m.Extra[i:], m.Extra[i+1:])
+			m.Extra[len(m.Extra)-1] = nil
+			m.Extra = m.Extra[:len(m.Extra)-1]
+			return
+		}
+	}
+}
+
+// replaceOPTHeader set the OPT Header with the one provided.
+func replaceOPTHeader(m *dns.Msg, hdr *dns.RR_Header) {
+	for i := len(m.Extra) - 1; i >= 0; i-- {
+		if m.Extra[i].Header().Rrtype == dns.TypeOPT {
+			m.Extra[i].(*dns.OPT).Hdr = *hdr
+			return
+		}
+	}
+	opt := dns.OPT{*hdr, make([]dns.EDNS0, 0)}
+	m.Extra = append(m.Extra, &opt)
 }
 
 // NewResponseReverter returns a pointer to a new ResponseReverter.
 func NewResponseReverter(w dns.ResponseWriter, r *dns.Msg) *ResponseReverter {
-	return &ResponseReverter{
-		ResponseWriter:   w,
-		originalQuestion: r.Question[0],
+	rr := &ResponseReverter{
+		ResponseWriter:  w,
+		receivedRequest: r.Copy(),
+		sharedRequest:   r,
 	}
+	return rr
 }
 
 // WriteMsg records the status code and calls the underlying ResponseWriter's WriteMsg method.
-func (r *ResponseReverter) WriteMsg(res *dns.Msg) error {
-	res.Question[0] = r.originalQuestion
+func (r *ResponseReverter) WriteMsg(resp *dns.Msg) error {
+	receivedOpt := r.receivedRequest.IsEdns0()
+	if receivedOpt != nil {
+		replaceOPTHeader(r.sharedRequest, receivedOpt.Header())
+	} else {
+		removeOPT(r.sharedRequest)
+		removeOPT(resp)
+	}
+
+	resp.Question[0] = r.receivedRequest.Question[0]
+
 	if r.ResponseRewrite {
-		for _, rr := range res.Answer {
+
+		// revert some rewritten fields and ONLY for the Active rules.
+		for _, rr := range resp.Answer {
 			var isNameRewritten bool = false
 			var isTtlRewritten bool = false
 			var name string = rr.Header().Name
@@ -75,7 +113,7 @@ func (r *ResponseReverter) WriteMsg(res *dns.Msg) error {
 			}
 		}
 	}
-	return r.ResponseWriter.WriteMsg(res)
+	return r.ResponseWriter.WriteMsg(resp)
 }
 
 // Write is a wrapper that records the size of the message that gets written.
--- a/plugin/rewrite/rewrite.go
+++ b/plugin/rewrite/rewrite.go
@@ -41,13 +41,14 @@ func (rw Rewrite) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	wr := NewResponseReverter(w, r)
 	state := request.Request{W: w, Req: r}
 
+ApplyRules:
 	for _, rule := range rw.Rules {
 		switch result := rule.Rewrite(ctx, state); result {
 		case RewriteDone:
 			if !validName(state.Req.Question[0].Name) {
 				x := state.Req.Question[0].Name
 				log.Errorf("Invalid name after rewrite: %s", x)
-				state.Req.Question[0] = wr.originalQuestion
+				state.Req.Question[0] = wr.receivedRequest.Question[0]
 				return dns.RcodeServerFailure, fmt.Errorf("invalid name after rewrite: %s", x)
 			}
 			respRule := rule.GetResponseRule()
@@ -56,16 +57,14 @@ func (rw Rewrite) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 				wr.ResponseRules = append(wr.ResponseRules, respRule)
 			}
 			if rule.Mode() == Stop {
-				if rw.noRevert {
-					return plugin.NextOrFailure(rw.Name(), rw.Next, ctx, w, r)
-				}
-				return plugin.NextOrFailure(rw.Name(), rw.Next, ctx, wr, r)
+				break ApplyRules
 			}
 		case RewriteIgnored:
 			break
 		}
 	}
-	if rw.noRevert || len(wr.ResponseRules) == 0 {
+
+	if rw.noRevert {
 		return plugin.NextOrFailure(rw.Name(), rw.Next, ctx, w, r)
 	}
 	return plugin.NextOrFailure(rw.Name(), rw.Next, ctx, wr, r)
--- a/plugin/rewrite/rewrite_test.go
+++ b/plugin/rewrite/rewrite_test.go
@@ -20,6 +20,14 @@ func msgPrinter(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, err
 	return 0, nil
 }
 
+func handlerRequestSaver(m *dns.Msg) plugin.HandlerFunc {
+	return func(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+		*m = *r.Copy()
+		w.WriteMsg(r)
+		return 0, nil
+	}
+}
+
 func TestNewRule(t *testing.T) {
 	tests := []struct {
 		args        []string
@@ -247,11 +255,6 @@ func TestRewrite(t *testing.T) {
 }
 
 func TestRewriteEDNS0Local(t *testing.T) {
-	rw := Rewrite{
-		Next:     plugin.HandlerFunc(msgPrinter),
-		noRevert: true,
-	}
-
 	tests := []struct {
 		fromOpts []dns.EDNS0
 		args     []string
@@ -297,10 +300,22 @@ func TestRewriteEDNS0Local(t *testing.T) {
 	}
 
 	ctx := context.TODO()
+
+	// EDNS0 does not ensure to return the options in the DNS response
+	// thus, we need to save the modified request and verify it has the expected changes.
+	var savedReq dns.Msg
+	rw := Rewrite{
+		Next:     plugin.HandlerFunc(handlerRequestSaver(&savedReq)),
+		noRevert: false,
+	}
+
 	for i, tc := range tests {
 		m := new(dns.Msg)
 		m.SetQuestion("example.com.", dns.TypeA)
 		m.Question[0].Qclass = dns.ClassINET
+		if tc.doBool {
+			m.SetEdns0(4096, true)
+		}
 
 		r, err := newEdns0Rule("stop", tc.args...)
 		if err != nil {
@@ -312,9 +327,7 @@ func TestRewriteEDNS0Local(t *testing.T) {
 		rec := dnstest.NewRecorder(&test.ResponseWriter{})
 		rw.ServeDNS(ctx, rec, m)
 
-		resp := rec.Msg
-		o := resp.IsEdns0()
-		o.SetDo(tc.doBool)
+		o := savedReq.IsEdns0()
 		if o == nil {
 			t.Errorf("Test %d: EDNS0 options not set", i)
 			continue
@@ -335,8 +348,9 @@ func TestEdns0LocalMultiRule(t *testing.T) {
 	r, _ = newEdns0Rule("stop", "local", "set", "0xffee", "fedcba")
 	rules = append(rules, r)
 
+	var savedReq dns.Msg
 	rw := Rewrite{
-		Next:     plugin.HandlerFunc(msgPrinter),
+		Next:     plugin.HandlerFunc(handlerRequestSaver(&savedReq)),
 		Rules:    rules,
 		noRevert: true,
 	}
@@ -360,19 +374,15 @@ func TestEdns0LocalMultiRule(t *testing.T) {
 		m := new(dns.Msg)
 		m.SetQuestion("example.com.", dns.TypeA)
 		m.Question[0].Qclass = dns.ClassINET
+		m.SetEdns0(4096, false)
 		if tc.fromOpts != nil {
 			o := m.IsEdns0()
-			if o == nil {
-				m.SetEdns0(4096, true)
-				o = m.IsEdns0()
-			}
 			o.Option = append(o.Option, tc.fromOpts...)
 		}
 		rec := dnstest.NewRecorder(&test.ResponseWriter{})
 		rw.ServeDNS(ctx, rec, m)
 
-		resp := rec.Msg
-		o := resp.IsEdns0()
+		o := savedReq.IsEdns0()
 		if o == nil {
 			t.Errorf("Test %d: EDNS0 options not set", i)
 			continue
@@ -537,6 +547,7 @@ func TestRewriteEDNS0LocalVariable(t *testing.T) {
 	for i, tc := range tests {
 		m := new(dns.Msg)
 		m.SetQuestion("example.com.", dns.TypeA)
+		m.SetEdns0(4096, false)
 
 		r, err := newEdns0Rule("stop", tc.args...)
 		if err != nil {
@@ -661,6 +672,7 @@ func TestRewriteEDNS0Subnet(t *testing.T) {
 	for i, tc := range tests {
 		m := new(dns.Msg)
 		m.SetQuestion("example.com.", dns.TypeA)
+		m.SetEdns0(4096, false)
 
 		r, err := newEdns0Rule("stop", tc.args...)
 		if err != nil {
@@ -673,11 +685,11 @@ func TestRewriteEDNS0Subnet(t *testing.T) {
 
 		resp := rec.Msg
 		o := resp.IsEdns0()
-		o.SetDo(tc.doBool)
 		if o == nil {
 			t.Errorf("Test %d: EDNS0 options not set", i)
 			continue
 		}
+		o.SetDo(tc.doBool)
 		if o.Do() != tc.doBool {
 			t.Errorf("Test %d: Expected %v but got %v", i, tc.doBool, o.Do())
 		}
--- a/test/rewrite_test.go
+++ b/test/rewrite_test.go
@@ -26,6 +26,7 @@ func TestRewrite(t *testing.T) {
 
 	testMX(t, udp)
 	testEdns0(t, udp)
+	testNoEdns0(t, udp)
 }
 
 func testMX(t *testing.T, server string) {
@@ -52,6 +53,8 @@ func testMX(t *testing.T, server string) {
 func testEdns0(t *testing.T, server string) {
 	m := new(dns.Msg)
 	m.SetQuestion("example.com.", dns.TypeA)
+	// allow EDNS0
+	m.SetEdns0(4096, true)
 
 	r, err := dns.Exchange(m, server)
 	if err != nil {
@@ -69,9 +72,13 @@ func testEdns0(t *testing.T, server string) {
 		t.Errorf("Expected 192.0.2.53, got: %s", r.Answer[0].(*dns.A).A.String())
 	}
 	o := r.IsEdns0()
-	if o == nil || len(o.Option) == 0 {
-		t.Error("Expected EDNS0 options but got none")
+	if o == nil {
+		t.Error("Expected OPT Record but got none")
 	} else {
+		if len(o.Option) != 1 {
+			t.Errorf("Expected only one EDNS0 options but got %d", len(o.Option))
+			return
+		}
 		if e, ok := o.Option[0].(*dns.EDNS0_LOCAL); ok {
 			if e.Code != 0xffee {
 				t.Errorf("Expected EDNS_LOCAL code 0xffee but got %x", e.Code)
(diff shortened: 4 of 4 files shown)


plugin/rewrite/reverter.go
```diff
@@ -21,22 +22,82 @@ type ResponseRule struct {
...
+	originalOPT      *dns.OPT
 	ResponseRewrite  bool
 	ResponseRules    []ResponseRule
 }
 
+func dupOPT(opt *dns.OPT) (o *dns.OPT) {
+	// ensure remove OPT if initial query was not Edns0 - compliance with https://tools.ietf.org/html/rfc6891#page-10
+	edns0 := make([]dns.EDNS0, len(opt.Option))
+	for i, e := range opt.Option {
+		edns0[i] = e.Copy()
```
fturib: this e.Copy() needs that PR https://github.com/miekg/dns/pull/902 is merged.

test/rewrite_test.go
```diff
@@ -52,6 +53,8 @@ func testMX(t *testing.T, server string) {
 func testEdns0(t *testing.T, server string) {
 	m := new(dns.Msg)
 	m.SetQuestion("example.com.", dns.TypeA)
+	// allow EDNS0
+	m.SetEdns0(4096, true)
```
fturib, line 57: I set EDNS0 with a buffer size only.
What do we expect from the response in OPT ?

At this moment, there is no change of behavior (in comparison to precedent version).

However, I am wondering if we should expect the OPT generated by the REWRITE at query time ...

plugin/rewrite/reverter.go
```diff
@@ -21,22 +22,77 @@ type ResponseRule struct {
...
+	}
+	o = &dns.OPT{opt.Hdr, edns0}
+	return o
+}
+
+// return a deep copy of the OPT inside this msg
+func copyOPT(m *dns.Msg) (o *dns.OPT) {
```
rdrozhdzh: this func is not needed. You can use existing `IsEdns0()` method and then make a copy if needed

plugin/rewrite/reverter.go
```diff
@@ -21,22 +22,77 @@ type ResponseRule struct {
...
+func replaceOPT(m *dns.Msg, o *dns.OPT) {
+	for i := len(m.Extra) - 1; i >= 0; i-- {
+		if m.Extra[i].Header().Rrtype == dns.TypeOPT {
+			if o == nil {
+				copy(m.Extra[i:], m.Extra[i+1:])
+				m.Extra[len(m.Extra)-1] = nil
+				m.Extra = m.Extra[:len(m.Extra)-1]
+				break
+			}
+			m.Extra[i] = dupOPT(o)
```
rdrozhdzh: don't duplicate here, you can duplicate it in outer code

plugin/rewrite/reverter.go
```diff
@@ -21,22 +22,77 @@ type ResponseRule struct {
...
+			break
+		}
+	}
+	if o != nil {
+		m.Extra = append(m.Extra, dupOPT(o))
```
rdrozhdzh: don't duplicate here, you can duplicate it in outer code

plugin/rewrite/reverter.go
```diff
@@ -21,22 +22,77 @@ type ResponseRule struct {
...
+	}
+}
+
 // NewResponseReverter returns a pointer to a new ResponseReverter.
 func NewResponseReverter(w dns.ResponseWriter, r *dns.Msg) *ResponseReverter {
 	return &ResponseReverter{
 		ResponseWriter:   w,
+		request:          r,
 		originalQuestion: r.Question[0],
+		originalOPT:      copyOPT(r),
```
rdrozhdzh: I see you decided to not remove from response the options which absent from request. In such a case, you can just have a boolean value here, which just designate if original request has an OPT record. So, this line could look like
`originalOPT:      r.IsEDNS0() != nil,`

plugin/rewrite/reverter.go
```diff
@@ -21,22 +22,77 @@ type ResponseRule struct {
...
 	}
 }
 
 // WriteMsg records the status code and calls the underlying ResponseWriter's WriteMsg method.
 func (r *ResponseReverter) WriteMsg(res *dns.Msg) error {
 	res.Question[0] = r.originalQuestion
+
+	// set back the client request to its original OPT record (before changed by Rewrite)
+	replaceOPT(r.request, r.originalOPT)
```
rdrozhdzh: not sure if it make sense to restore original OPT in request. It's only important to remove the OPT record if it was not originally present in request, so that `SizeAndDo()` to not insert OPT record into response

plugin/rewrite/reverter.go
```diff
@@ -21,22 +22,77 @@ type ResponseRule struct {
...
+	if r.originalOPT == nil {
+		// ensure remove OPT if initial query was not EDNS0 - compliance with https://tools.ietf.org/html/rfc6891#page-10
+		replaceOPT(res, r.originalOPT)
```
rdrozhdzh: `replaceOPT(res, nil)` would be clearer

plugin/rewrite/reverter.go
```diff
@@ -21,22 +22,77 @@ type ResponseRule struct {
...
+		request:          r,
 		originalQuestion: r.Question[0],
+		originalOPT:      copyOPT(r),
 	}
 }
 
 // WriteMsg records the status code and calls the underlying ResponseWriter's WriteMsg method.
 func (r *ResponseReverter) WriteMsg(res *dns.Msg) error {
 	res.Question[0] = r.originalQuestion
+
```
rdrozhdzh: So far, I don't see why this `ResponseReverter` code will be called in case of EDNS rules. They don't have corresponding response rules and therefore `plugin.NextOrFailure()` will be called with original `ResponseWriter`, not with `ResponseReverter`. See the code at 
https://github.com/coredns/coredns/blob/92bec854dc609737e8686cdfb6bf61cdd9c07f7d/plugin/rewrite/rewrite.go#L68-L71

test/rewrite_test.go
```diff
@@ -72,15 +110,43 @@ func testEdns0(t *testing.T, server string) {
...
+		if len(o.Option) != 1 {
+			t.Errorf("Expected only one EDNS0 options but got %d", len(o.Option))
+		}
 		if e, ok := o.Option[0].(*dns.EDNS0_LOCAL); ok {
 			if e.Code != 0xffee {
 				t.Errorf("Expected EDNS_LOCAL code 0xffee but got %x", e.Code)
 			}
-			if !bytes.Equal(e.Data, []byte("hello-world")) {
-				t.Errorf("Expected EDNS_LOCAL data 'hello-world' but got %q", e.Data)
+			if !bytes.Equal(e.Data, []byte("initialization")) {
```
rdrozhdzh: Why did you change this test? It seems the "initialization" should be rewritten with "hello-world" by rewrite plugin

test/rewrite_test.go
```diff
@@ -72,15 +110,43 @@ func testEdns0(t *testing.T, server string) {
...
+				t.Errorf("Expected EDNS_LOCAL data 'initialization' but got %q", e.Data)
 			}
 		} else {
-			t.Errorf("Expected EDNS0_LOCAL but got %v", o.Option[0])
+			t.Errorf("Expected EDNS_COOKIE but got %v", o.Option[0])
```
rdrozhdzh: Why EDNS_COOKIE ?

test/rewrite_test.go
```diff
@@ -72,15 +110,43 @@ func testEdns0(t *testing.T, server string) {
...
+		t.Error("Expected to at least one RR in the answer section, got none")
+	}
+	if r.Answer[0].Header().Rrtype != dns.TypeA {
+		t.Errorf("Expected RR to A, got: %d", r.Answer[0].Header().Rrtype)
+	}
+	if r.Answer[0].(*dns.A).A.String() != "192.0.2.53" {
+		t.Errorf("Expected 192.0.2.53, got: %s", r.Answer[0].(*dns.A).A.String())
+	}
+	o := r.IsEdns0()
+	if !(o == nil || len(o.Option) == 0) {
```
rdrozhdzh: if o != nil

test/rewrite_test.go
```diff
@@ -72,15 +110,43 @@ func testEdns0(t *testing.T, server string) {
...
+		t.Error("Expected to be no EDNS0 options in the reply")
```
rdrozhdzh: Expected to be no OPT RR in the reply

rdrozhdzh (dismissed):
The main concerns are
- unneeded copying of OPT record in some cases, e.g. if all rewrite rules are not related to EDNS
- seems `ResponseReverter` is not used properly in case of EDNS rules
- tests should be fixed

test/rewrite_test.go
fturib: That is the whole point of this comment here : https://github.com/coredns/coredns/issues/2484#issuecomment-456433496

With my changes:

1- client EDNS_LOCAL("Initialization") no EDNS
2- Rewrite EDNS_LOCAL("Hello-World")
3- erratic resolution --> no EDNS
4- Rewrite writing response : let the response flow .. (do not alter the response at all)
5- SizeAndDo : if response has no EDNSO, fill with queries EDNS0 else let the response flow


Question: Should step 4 consolidate the response with the EDNS of the forged query at step 2 ?

I think no, but not sure here ... it is interpretation of the RFC.
I would think that the forged query of REWRITE is ONLY to get a resolution. It should be transparent for the initial client - but I maybe wrong.

Let me know what you think.

If we need to revert to former behavior, I guess that remove the last commit will do the trick.

fturib:
@rdrozhdzh : Thanks for the review.
Please assess my reply on your question here : https://github.com/coredns/coredns/pull/2491#discussion_r250245768

That part is important, and will decide, I guess if I need to change those functions.
Thanks!

test/rewrite_test.go
fturib: > Why did you change this test? It seems the "initialization" should be rewritten with "hello-world" by rewrite plugin

I do not think so.
Right now, for unclear reason, SizeAndDo() is just copying the request ENS0s  into the response OPT.
But i think we just should copy the header OPT (to confirm the version EDNS0). But there is no reason to copy the EDNS0 elements ... (Or I do not see it).

IMO, the initial UT that is testing 'Hello-world" is not correct.

rdrozhdzh: Ok, I see your point now. I didn't pay attention that `erratic` doesn't support EDNS.
Lets keep your changes for now. We may want to revise logic of `SizeAndDo()` in a separate PR if needed

plugin/rewrite/reverter.go
fturib: No. OPT has no public Copy().
However, now, EDNS0 has not neither.
so the last way to have a copy of OPT is : msg.Copy().isEdns0().

plugin/rewrite/reverter.go
fturib: yes. changed

plugin/rewrite/reverter.go
fturib: yes. changed.

plugin/rewrite/reverter.go
fturib: I think we MUST replace the OPT of the original query. 
At least size of buffer, etc ... 
I believe that the copy of sub-array of EDNS0 is not really useful.

plugin/rewrite/reverter.go
fturib: as if today it is important : you get the size of the buffer in the OPT header.
(used in SizeAndDo).

plugin/rewrite/reverter.go
fturib: that is interesting .. because the debug mode goes into the ResponseReverter ...

test/rewrite_test.go
fturib: error of test of code. forgot to revert back.

test/rewrite_test.go
fturib: yes.

test/rewrite_test.go
fturib: yes.

plugin/rewrite/reverter.go
fturib: because the default mode is STOP. and in case of STOP, then we go by line #62, into the next plugin with the RevertWriter.
That code seems messy ....

fturib:
I made change simpler : 
- only hasOPT and OPT.hdr are saved/restored.
- the Reverter is ALWAYS used to ensure OPT computation is correct. The Logic to avoid the reversion if not wanted was already in the reverter.
- the initial test_rewrite is back : 
   - the client will receive the EDNS0 options rewritten by plugins
   - unless the resolver provide its own EDNS0 in the response
   - **BUT the initial OPT header is restored : version, buffer size, flags are the one of the initial query and the response to the client will be tuned accordingly**.
   - if the client does not set EDNS0 flags, then it will not receive any OPT record back.
- internal UT were wrong, expecting OPT record return although the flag is not set in the query. Fixed those UT.

fturib:
I believe we should port some changes in SizeAndDo():
- save the initial OPT header, instead of the pointer to initial request (that is modified by other plugins)
- ensure to clean out the OPT from the response, if not expected by the client
- do not report the EDNS0 option from initial request to the response back (there is no reason here ...).

plugin/rewrite/reverter.go
rdrozhdzh: I meant one can use code like
```
o := dupOPT(msg.IsEnds0())
```
Though, `dupOPT()` should expect nil parameter for this to work. Otherwise, a bit more cumbersome construction will have to be used
```
o := msg.IsEnds0()
if o != nil {
  o := dupOPT(o)
}
```

plugin/rewrite/rewrite.go
```diff
@@ -51,23 +51,17 @@ func (rw Rewrite) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 				return dns.RcodeServerFailure, fmt.Errorf("invalid name after rewrite: %s", x)
 			}
 			respRule := rule.GetResponseRule()
-			if respRule.Active == true {
+			if !rw.noRevert && respRule.Active == true {
 				wr.ResponseRewrite = true
 				wr.ResponseRules = append(wr.ResponseRules, respRule)
 			}
 			if rule.Mode() == Stop {
-				if rw.noRevert {
```
rdrozhdzh, line 59: I'm afraid this is not correct. If the last rule in a series of continued rules is not triggered (RewriteIgnored), then you will skip ResponseReverter even if some previously triggered rules had corresponding response rules.

fturib: I do not think so. If we never go into the "stop" part, then the query will be processed by the "NextOrFailure" that is at the end of the procedure and use - the same way- the ResponseReverter.

Although I agree, would be better to show only one way to run the NextOfFailure.
Will try to propose another solution.

rdrozhdzh: Oh, I didn't see that the last `NextOrFailure()` also uses `wr`
Perhaps, you can just do `break` on Stop to avoid code duplication

fturib: yes .. that was my though too.

plugin/rewrite/reverter.go
```diff
@@ -21,23 +22,69 @@ type ResponseRule struct {
...
+		// ensure remove OPT out of the response
+		// if initial query was not EDNS0 - compliance with https://tools.ietf.org/html/rfc6891#page-10
+		replaceOPTHeader(res, false, r.originalOPTHeader)
+	}
+
+	// Revert part happen ONLY if flag is set
 	if r.ResponseRewrite {
+
+		// always revert the question
+		res.Question[0] = r.originalQuestion
```
rdrozhdzh: I believe this should be unconditional. The question name in response must be the same as in request

plugin/rewrite/reverter.go
```diff
@@ -21,23 +22,69 @@ type ResponseRule struct {
 // dig will complain with ';; Question section mismatch: got example.org/HINFO/IN'
 type ResponseReverter struct {
 	dns.ResponseWriter
-	originalQuestion dns.Question
-	ResponseRewrite  bool
-	ResponseRules    []ResponseRule
+	request           *dns.Msg
+	originalQuestion  dns.Question
+	originalOPTHeader dns.RR_Header
+	originalHasOPT    bool
```
rdrozhdzh: Perhaps, we can avoid the bool member and use method
```
func (rr *ResponseReverter) hasOriginalOPT() bool {
  return rr.originalOPTHeader.Rrtype == dns.TypeOPT
}
```

plugin/rewrite/reverter.go
```diff
@@ -21,23 +22,69 @@ type ResponseRule struct {
...
+	}
+	if hasOpt {
+		opt := dns.OPT{hdr, make([]dns.EDNS0, 0)}
+		m.Extra = append(m.Extra, &opt)
+	}
 }
 
 // NewResponseReverter returns a pointer to a new ResponseReverter.
 func NewResponseReverter(w dns.ResponseWriter, r *dns.Msg) *ResponseReverter {
+	hdr := dns.RR_Header{}
```
rdrozhdzh: consider code like
```
rr := &ResponseReverter{
	ResponseWriter:   w,
	request:           r,
	originalQuestion:  r.Question[0],
}
if o := r.IsEdns0(); o != nil {
	rr.originalOPTHeader = *o.Header()
}
return rr
```
This avoids extra Header copying to temp variable

plugin/rewrite/reverter.go
```diff
@@ -21,23 +22,69 @@ type ResponseRule struct {
...
+		originalOPTHeader: hdr,
 	}
 }
 
 // WriteMsg records the status code and calls the underlying ResponseWriter's WriteMsg method.
 func (r *ResponseReverter) WriteMsg(res *dns.Msg) error {
-	res.Question[0] = r.originalQuestion
+
+	// set back the client request to its original OPT header
+	replaceOPTHeader(r.request, r.originalHasOPT, r.originalOPTHeader)
```
rdrozhdzh:
```
replaceOPTHeader(r.request, r.hasOriginalOPT(), r.originalOPTHeader)
```

rdrozhdzh (commented):
Overall, looks good to me, except https://github.com/coredns/coredns/pull/2491#pullrequestreview-196995632
Please consider also couple of minor enhancements below

plugin/rewrite/reverter.go
fturib: You are right.
But here it is messy. 
- "noRevert" is only a UT case. and is used only to test if rewrite act as expected
- other case (with a response) needs to revert the OPT and the question
- in those other cases, some option requires a  "responseRewrite" (the rules that are Active)

I will change back to specific case for "noRevert" and let the use of Reverter only for real response use case.

plugin/rewrite/reverter.go
fturib: ok.

plugin/rewrite/reverter.go
fturib: ok.

plugin/rewrite/reverter.go
fturib: ok.

rdrozhdzh (approved):
LGTM

fturib:
@miekg : can I merge here, or you wish to review this PR ?

rdrozhdzh:
@miekg, please let us know if you have any concerns on this

plugin/rewrite/rewrite_test.go
```diff
@@ -20,6 +20,14 @@ func msgPrinter(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, err
 	return 0, nil
 }
 
+func handlerRequestSaver(m *dns.Msg) plugin.HandlerFunc {
```
miekg, line 23: a whole handler to return the message 'm', why is this needed?

plugin/rewrite/rewrite.go
```diff
@@ -42,30 +42,25 @@ func (rw Rewrite) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
...
-					return plugin.NextOrFailure(rw.Name(), rw.Next, ctx, w, r)
-				}
-				return plugin.NextOrFailure(rw.Name(), rw.Next, ctx, wr, r)
+				break
 			}
-		case RewriteIgnored:
```
miekg: In the new code I can't see this case, being handled...

Also don't understand why the 'switch' was changed to a if-then-elses

plugin/rewrite/reverter.go
```diff
@@ -21,23 +22,69 @@ type ResponseRule struct {
...
+
+	// set back the client request to its original OPT header
+	replaceOPTHeader(r.request, r.hasOriginalOPT(), r.originalOPTHeader)
+	if !r.hasOriginalOPT() {
+		// ensure remove OPT out of the response
+		// if initial query was not EDNS0 - compliance with https://tools.ietf.org/html/rfc6891#page-10
+		replaceOPTHeader(res, r.hasOriginalOPT(), r.originalOPTHeader)
+	}
+
+	// always revert the question
```
miekg: this comment doesn't add anything

plugin/rewrite/reverter.go
```diff
@@ -21,23 +22,69 @@ type ResponseRule struct {
...
 	res.Question[0] = r.originalQuestion
+
+	// Revert part happen ONLY if flag is set
```
miekg: This comment doesn't add anything (also misses closing dot)

plugin/rewrite/reverter.go
```diff
@@ -21,23 +22,69 @@ type ResponseRule struct {
 // dig will complain with ';; Question section mismatch: got example.org/HINFO/IN'
 type ResponseReverter struct {
 	dns.ResponseWriter
-	originalQuestion dns.Question
-	ResponseRewrite  bool
-	ResponseRules    []ResponseRule
+	request           *dns.Msg
```
miekg: not only do you put the original message in there; you're duplicating original question and original opt header as well. 
dont understand

plugin/rewrite/reverter.go
```diff
@@ -21,23 +22,69 @@ type ResponseRule struct {
...
+	originalQuestion  dns.Question
+	originalOPTHeader dns.RR_Header
+	ResponseRewrite   bool
+	ResponseRules     []ResponseRule
+}
+
+// align the OPT inside the msg the information provided - hasOPT and Header
+// the sub array of EDNS0 is not changed (unless OPT is removed)
+func replaceOPTHeader(m *dns.Msg, hasOpt bool, hdr dns.RR_Header) {
```
miekg: boolean params are a good indicators this should be two functions

plugin/rewrite/rewrite_test.go
fturib: It is for UT. 
I need to get the DNS request sent AFTER process by Rewrite in order to test that the OPT we modify o that DNS request is correct.

plugin/rewrite/rewrite.go
fturib: There was no code attached to `RewriteIgnore`.
I changed for switch to if, because it was possible and that allow to do a `break` in the loop.
Else I need a more complicated way to `break` out of the switch and the loop

plugin/rewrite/reverter.go
fturib: i'll remove then.

plugin/rewrite/reverter.go
fturib: i'll remove then.

plugin/rewrite/reverter.go
fturib: That is the whole point of this PR: (see description of the PR).

We need that `rewrite` return the `OPT Header` part of the `request` as it was before any modification by Rewrite. (because SizeAndDo will response to initial client according to that Header).

for that we need to keep a copy of the original opt header.

the `request` is only a pointer to the request that is flowing up to the pugin resolver. It can be changed by any of the plugin in the handler list.


About copy of `QUESTION` I guess that is a similar problem. But was there before this PR.

plugin/rewrite/reverter.go
fturib: I can change that.

plugin/rewrite/rewrite.go
fturib: ok. I found the way with a labelled break. Will revert to initial switch.

plugin/rewrite/reverter.go
fturib: not two functions but integrate the boolean inside the function.

fturib:
@miekg : I updated after your review.


I think the main concern of this PR is the way we want to handle the OPT record(s) in the plugin chain.
- SizeAndDo process the reply according to the OPT in the request. BUT, the request may be modified by the chain of plugins.

This PR is fixing the code for `Rewrite` as a plugin.
But the concern may be bigger, and may have to be fixed in `request.SizeAndDo`

**Responsibility of plugins or of SizeAndDo**
=> Do we want each plugin to return the OPT header that it received ? or do we want the SizeAndDo to keep a copy of the real initial OPT header received in order to be able to reply properley whateber happen in the plugins ?

**What should be the process of the OPT data ?**
In SizeAndDo, If there is no OPT returned in the response, we are copying the one from the request. 
https://github.com/coredns/coredns/blob/92836cc6f9aaf47a349993271ddae273d8e7d25d/request/request.go#L217-L227
=> Why ? And should we copy the one from the "initial" request, or the ones from the request modified by the plugin ?

NOTE: from the [RFC](https://tools.ietf.org/html/rfc6891#page-10), I understand we should not change the OPT records : just let flow the ones returned by the resolver.

miekg:
Which is the thing you send. Looks overly complex to me

miekg:
Your duplicating several things in this structure. There is no need for
that. Please fix

miekg:
Yes we want it there to show its handled. Please revert

miekg:
On Wed, 6 Feb 2019, 22:08 Francois Tur <notifications@github.com wrote:

> *@fturib* commented on this pull request.
> ------------------------------
>
> In plugin/rewrite/reverter.go
> <https://github.com/coredns/coredns/pull/2491#discussion_r254472138>:
>
> > @@ -21,23 +22,69 @@ type ResponseRule struct {
>  // dig will complain with ';; Question section mismatch: got example.org/HINFO/IN'
>  type ResponseReverter struct {
>  	dns.ResponseWriter
> -	originalQuestion dns.Question
> -	ResponseRewrite  bool
> -	ResponseRules    []ResponseRule
> +	request           *dns.Msg
> +	originalQuestion  dns.Question
> +	originalOPTHeader dns.RR_Header
> +	ResponseRewrite   bool
> +	ResponseRules     []ResponseRule
> +}
> +
> +// align the OPT inside the msg the information provided - hasOPT and Header
> +// the sub array of EDNS0 is not changed (unless OPT is removed)
> +func replaceOPTHeader(m *dns.Msg, hasOpt bool, hdr dns.RR_Header) {
>
> not two functions but integrate the boolean inside the function
>

Don't understand. You're splitting this up?

fturib:
I rebased and squash all commits for this PR.

About your comments of last review:

> Which is the thing you send. Looks overly complex to me

the UT needs to test the msg that is modified by `rewrite` and sent to next plugin. UT cannot prefigure if the input Request msg will be the same as the one rewritten. 
So the best is to catch that request msg after the plugin.

> Your duplicating several things in this structure. There is no need for that. Please fix

Hum... All the data saved in this structure is needed for the reverter ...
Ok. I changed to propose another implementation.

> Yes we want it there to show its handled. Please revert

I did the revert already. Can you double-check ?

> Don't understand. You're splitting this up?

Ok. Done.

miekg:
I have no idea what the state of this is and I intent to just close this soon

rdrozhdzh:
@miekg, I guess this was waiting for your review, and I'm afraid @fturib is not working on this anymore.
If you think this solution is too cumbersome, you can close this, just don't close the related issue, since it describes RFC violation. Perhaps, I would propose a simpler solution later - no time to work on this right now.

rdrozhdzh:
see #3015 for alternative fix

Issue #2492, opened 2019-01-23
fix kubernetes in-cluster CNAME lookup, proper fix?


wizard580:
**1.** What happened?
Intro: it's about not fully/properly fixed issue #2040 , related to the https://github.com/kubernetes/kubernetes/issues/67962

What happened:

I have a cluster with CoreDNS enabled

**First case:** Defined this service which should resolve to ClusterIP of the kubernetes service:
```apiVersion: v1
kind: Service
metadata:
  name: test
  namespace: default
spec:
  externalName: kubernetes-2.default.svc.cluster.local
  type: ExternalName
```
dig for this services DNS name will fail to show `CNAME` record

**Second case:** Having defined resource from the first case, define another resource:
```apiVersion: v1
kind: Service
metadata:
  name: kubernetes-2
  namespace: default
spec:
  type: ClusterIP
  selector:
    app: some-label
  ports:
  - name: some-port-8080
    port: 80
    protocol: TCP
    targetPort: 8080
```
Exact service definition **isn't important**. We just need **any normal service**.

dig for this services DNS name **will work** returning correct `A` record.
dig for first services DNS name **will still fail** to show `CNAME` record.

Then, **re-apply** first definition, without changing anything. 
And now dig will return correct `CNAME` record.

Looks like:
1) CoreDNS tries to check/resolve `CNAME-target` and if it fails (this could be temporary issue as I can imagine) it will not add `CNAME` record
2) When `CNAME-target` actually become available/created - we still miss `CNAME`

**2.** Which issues (if any) are related?

#2040 

Why DNS server cares about content of `CNAME` record? This is blocker for our use cases, and K8S regression after migration from `kube-dns`.
Can't we skip mentioned `CNAME` check?

chrisohaver:
IIUC, this works as intended.  I think it is an issue with timing, and cache.

In the first case, there is an external name that points to a non existing service.
This will result in a `CNAME` that points to a target that has no `A` record.
If you query that name for an `A` record (typically the default if you don't specify a record type), then you'll get an NXDOMAIN.  This answer gets cached.
If you query for a `CNAME` record, then you'll get the `CNAME`.
This is if I'm not mistaken, standard DNS behavior with `CNAMES` that point to non-existing targets.

In the second case, you create the service, so the `CNAME` now points to an existing record.
However, the response from when it did not exist is still cached.  So, you'll still get an NXDOMAIN until that cache entry expires.  The default TTL is for k8s records 5 seconds.

[...]

PR #2493, opened 2019-01-23
plugin/backend: CNAMEs to non-existing targets in same domain

chrisohaver:

Include CNAME record in a response where the CNAME target is in the same domain (or handled by same plugin), and the CNAME target does not exist. Prior to this change, we would reply NXDOMAIN for this case.

#2492 

none

### DIFF
--- a/plugin/etcd/lookup_test.go
+++ b/plugin/etcd/lookup_test.go
@@ -125,7 +125,9 @@ var dnsTestCases = []test.Case{
 	// CNAME (unresolvable internal name)
 	{
 		Qname: "cname.prod.region1.skydns.test.", Qtype: dns.TypeA,
-		Ns: []dns.RR{test.SOA("skydns.test. 30 SOA ns.dns.skydns.test. hostmaster.skydns.test. 0 0 0 0 0")},
+		Answer: []dns.RR{
+			test.CNAME("cname.prod.region1.skydns.test. 300 CNAME unresolvable.skydns.test."),
+		},
 	},
 	// Wildcard Test
 	{
@@ -181,7 +183,10 @@ var dnsTestCases = []test.Case{
 	// CNAME loop detection
 	{
 		Qname: "a.cname.skydns.test.", Qtype: dns.TypeA,
-		Ns: []dns.RR{test.SOA("skydns.test. 30 SOA ns.dns.skydns.test. hostmaster.skydns.test. 1407441600 28800 7200 604800 60")},
+		Answer: []dns.RR{
+			test.CNAME("a.cname.skydns.test. 300 CNAME b.cname.skydns.test."),
+			test.CNAME("b.cname.skydns.test. 300 CNAME a.cname.skydns.test."),
+		},
 	},
 	// NODATA Test
 	{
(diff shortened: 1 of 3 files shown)


chrisohaver:
@miekg, please validate if this is the proper way to reply in this case. Thanks!

chrisohaver:
Seems that in at least one test, we are expecting and validating the original behavior.

```
--- FAIL: TestLookup (0.13s)
    lookup_test.go:323: Answer for "cname.prod.region1.skydns.test." contained 1 results, 0 expected
    lookup_test.go:323: Answer for "a.cname.skydns.test." contained 2 results, 0 expected
```

This this a bad expectation of the test? Or are the tests and the existing behavior both correct as they are?

fturib (commented):
I understand that, now, we do not care anymore to be sure there is an A record to reply the CNAME.
CNAME is replied whatever as defined by the Service ExternalName
if it is in the domain we are authoritative, then we try to provide also the corresponding A record - but failing to get that extra information does not prevent replyin gthe CNAME.

LGTM.

fturib (approved):

plugin/etcd/lookup_test.go
```diff
@@ -181,7 +183,10 @@ var dnsTestCases = []test.Case{
 	// CNAME loop detection
 	{
 		Qname: "a.cname.skydns.test.", Qtype: dns.TypeA,
-		Ns: []dns.RR{test.SOA("skydns.test. 30 SOA ns.dns.skydns.test. hostmaster.skydns.test. 1407441600 28800 7200 604800 60")},
+		Answer: []dns.RR{
+			test.CNAME("a.cname.skydns.test. 300 CNAME b.cname.skydns.test."),
+			test.CNAME("b.cname.skydns.test. 300 CNAME a.cname.skydns.test."),
```
miekg, line 188: you create a cycle here, that's most def. not what you want to return

chrisohaver:
Actually - i may have prematurely closed this... 

@miekg, multiple choice ... are you saying ...
A) that this behavior is invalid in general?
B) Or are you saying that we should analyze our responses for CNAME loops, and return nothing (NODATA) if there is a loop.
C) both

miekg:
you don't want to return loops; how can that be useful for a client? 

I haven't look into this PR in detail, but I'm pretty sure we dont return loops at this moment

chrisohaver:
I'll re-open and address the loop issue if you think #2492 is a valid concern.  Until then i'll leave this closed.

chrisohaver:
IMO, I think we could justifiably take a garbage-in-garbage out stance here.  For example, if a user creates a CNAME loop of two ExternalName services pointing at each other, then we could pass the loop along to the client.  I'm pretty confident there is no use case for pointing two ExternalName Services at one another (for the same reason there should be no use case for pointing two CNAMEs at each other).

Filtering out loops here requires additional processing.

wizard580:
Sorry, is there any progress?

chrisohaver:
I think @miekg is not satisfied with the solution.

miekg:
I haven't seen a clear explanation on what the problem is. If only seen
ill-fated attempts to change something.

Issue #1523, opened 2018-02-13
Vendor all dependencies?


achille-roussel:
Hello,

I'm having a bit of trouble getting reproducible builds with CoreDNS, which makes tracking down issues much harder.

I have a build that I made 5 days ago and worked just fine, but the build I'm making today have broken the config we used.

Most of the pain comes from the _custom_ build process that CoreDNS uses with its makefile, which does `go get` + `git checkout` to lock down dependencies, where this could be better served by leveraging the vendor folder.

There are already a bunch of dependencies checked in into the vendor folder, my question is why not vendor all dependencies? That way building coredns would basically just be
```
go generate
go build
```

We can't just use one of the tagged releases because we add external plugins so we need to rebuild CoreDNS, and the `go get` steps fail if the coredns repo is checked out in a non-master branch or a branch that does not exist on the remote repo (due to interdependencies between coredns and plugins).

I'd love to hear about how people have addressed those kinds of problems as well, maybe I'm not approaching this the right way.

chrisohaver:
This go get "version guarding" was added about a month ago (#1370).  There isn't really an explanation as to why vendoring is not used.  Vendoring has its own set of pain points. @yongtang @miekg 

The change caused similar breakage as you describe in the continuous integration tests when an external plugin (forward) was included by default in the coredns Makefile.  The temporary fix (#1435) was to ignore the failure of the step in the Makefile. This had the effect of not re-getting coredns/coredns, which you already have at correct version if you are building from a branch.  IMO, it seems like Go should know this and not be confused, but maybe there is ambiguity there and I'm only seeing one case.  Since then, forward has been moved into coredns/coredns, so its no longer an external plugin.

miekg:
We try to vendor everything, but because of external plugins we can't sometimes. This is situation is pretty terrible. Go 1.9 and type aliases might make this better, see https://github.com/coredns/coredns/issues/792 for instance.

achille-roussel:
You make a good point about the difficulty that external plugins represent.

However, they can be vendored too for custom builds, if all dependencies live in `./vendor` there should be no issues to build coredns.

What do you think?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Vendor all de..." ]
>You make a good point about the difficulty that external plugins represent.
>
>However, they can be vendored too for custom builds, if all dependencies live in `./vendor` there should be no issues to build coredns.
>
>What do you think?

[...]

PR #2495, opened 2019-01-25
Vendor dependencies to allow for self-contained builds

MarcPow:

Consider vendoring dependencies.

I'll note that this suggestion has been rejected in the past, but I wanted to surface it again.

I'm relatively new to the world of Go, so it may be that there's a subtlety here of which I'm unaware, about which I'm happy to be educated.

If instead, perhaps, things have changed (and this is now acceptable), I'd love to see it merged.

https://github.com/coredns/coredns/issues/1523


miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] Vendor dependenci..." ]
>
>
>
>Consider vendoring dependencies.
>
>I'll note that this suggestion has been rejected in the past, but I wanted to surface it again.
>
>I'm relatively new to the world of Go, so it may be that there's a subtlety here of which I'm unaware, about which I'm happy to be educated.

external plugin will fail to compile if we do this:

	@ # Not vendoring these, so external plugins compile, avoiding:
	@ # cannot use c (type *"github.com/mholt/caddy".Controller) as type
	@ # *"github.com/coredns/coredns/vendor/github.com/mholt/caddy".Controller like errors.

so we can't. Maybe gomodules can make this work.

MarcPow:
> [ Quoting <notifications@github.com> in "[coredns/coredns] Vendor dependenci..." ]

> external plugin will fail to compile if we do this: @ # Not vendoring these, so external plugins compile, > avoiding: @ # cannot use c (type *"github.com/mholt/caddy".Controller) as type @ # >*"github.com/coredns/coredns/vendor/github.com/mholt/caddy".Controller like errors. so we can't. >Maybe gomodules can make this work.

Is there documentation flow on the external plugin compilation flow?   Naively, I've been assuming:  If "make coredns" works cleanly, then we should be good.    But that's apparently not the case... Put another way, can you give me an example of an external plugin whose compilation will fail in this manner, so that I have something to use as a litmus test - if I wanted to experiment to Go modules for example?

I realize it's not your job to educate me - thanks for being willing to engage in discussion on this! :)

miekg:
any external plugin will fail. See https://coredns.io/explugins/unbound/ for an example one, are coredns/example on github

PR #2496, opened 2019-01-25
Update plugin.md

AdamDang:
Fix a small error in line 11-12:
Original:
If the error is not nil CoreDNS, will return a SERVFAIL to the client. 
Should be:
If the error is not nil, CoreDNS will return a SERVFAIL to the client.


yongtang (approved):
LGTM

PR #2497, opened 2019-01-25
plugin/etcd: Remove unnecessary struct copy

callebjorkell:
The `b` struct is just copying all the fields into a new one, to
check if it already existed in a set. This isn't needed as all the
fields are identical, and a small rearrangement of the code solves the
same problem without the copy.


callebjorkell:
Found this when poking about in the code, so thought I would do a little housekeeping :slightly_smiling_face:

PR #2499, opened 2019-01-25
plugin/etcd: Filter empty host field by qtype

### LINKED ISSUE #2443: plugin/etcd: Empty host fields in services need to be ignored
In coredns 1.2.6, for a domain with services with empty `Host` fields, various queries (tested with `CNAME`, `A`, and `SRV` lookups) will contain answers with the domain `.`. The correct behaviour for this seems to be to skip all services that have an empty `Host` field for all queries other than `TXT`. For domains that have no other services defined, this should also work, since the expected result here is a NODATA response.

Initial discussion in PR #2425

callebjorkell:

When a query, different from a TXT lookup is performed, all services
with a missing `Host` field should be filtered out, as these otherwize
cause a line in the answer section with a single dot (`.`) as the
result. This behavior manifests for example when a TXT record is present
on a domain, eg. an A or SRV lookup is performed on said domain.

If there are no services containing a `Host` field, a `NODATA` response
should be given. If there are other Services, these alone should be
returned for the query.

Filter any service that has an empty Host field from all lookup types
other than TXT to solve this issue. At the same time the check for empty
`Text` fields in TXT queries are also moved to the same check in the
etcd ServiceBackend.

Replaces #2425 
Closes #2443 

N/A

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -331,9 +331,6 @@ func TXT(b ServiceBackend, zone string, state request.Request, opt Options) (rec
 	}
 
 	for _, serv := range services {
-		if serv.Text == "" {
-			continue
-		}
 		records = append(records, serv.NewTXT(state.QName()))
 	}
 	return records, nil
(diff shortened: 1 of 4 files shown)


callebjorkell:
I also checked that the kubernetes plugin behvior for TXT lookups wouldn't be affected by the moving of the empty Text field check.

miekg:
kubernetes txt sidesteps this entire code - hence my remark about forking backend.go into kubernetes - something we should just do ; and then strip the kubernetes side of it, because half of the features are not used there.

PR #2506, opened 2019-01-27
plugin/hosts: fix for ipv4-in-ipv6

thomas-mangin:

This fixes a bug when parsing IPv4 addresses using IPv6 encoding such as `::FFFF:127.0.0.1`

The work was part of https://github.com/coredns/coredns/pull/2489 but it was requested to have this as a different PR.

N/A

### DIFF
--- a/plugin/hosts/hostsfile.go
+++ b/plugin/hosts/hostsfile.go
@@ -185,9 +185,8 @@ func (h *Hostsfile) parse(r io.Reader, override *hostsMap) *hostsMap {
 }
 
 // ipVersion returns what IP version was used textually
-// gomacro> net.ParseIP("::FFFF:127.0.0.1")
-// 127.0.0.1	// net.IP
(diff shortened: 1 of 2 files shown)


plugin/hosts/hostsfile.go
```diff
@@ -185,8 +185,11 @@ func (h *Hostsfile) parse(r io.Reader, override *hostsMap) *hostsMap {
 }
 
 // ipVersion returns what IP version was used textually
+// gomacro> net.ParseIP("::FFFF:127.0.0.1")
```
miekg: Please drop the gomacro (no idea what that is) and just tell that `::FFFF` will be detected as v4. Also reference the RFC as. See RFC 4291 Section 2.5.5

thomas-mangin: gomacro: https://github.com/cosmos72/gomacro

Comment updated as requested.

Issue #2511, opened 2019-01-28
CoreDNS uses a lot memory during initialization phase


zhouhaibing089:
We are running CoreDNS 1.2.6 with the following Corefile:

```nginx
.:53 {
    errors
    health
    pprof :6060
    kubernetes cluster.local in-addr.arpa ip6.arpa {
        pods insecure
        upstream
        fallthrough in-addr.arpa ip6.arpa
    }
    prometheus :9153
    proxy . /etc/resolv.conf
    cache 30
    loop
    reload
    loadbalance
}
```

And we have around 5167 endpoints objects in the kubernetes cluster. We find whenever it restarts, it is going to consume a lot of memory. (approaching to 1G)

<img width="981" alt="screen shot 2019-01-28 at 3 34 13 pm" src="https://user-images.githubusercontent.com/4516999/51873489-38276480-2312-11e9-85f4-3b809036f406.png">

It will go back to a normal state with memory usage around 300MB.

This can also be confirmed with go pprof:

**With Alloc Space**

```console
$ go tool pprof --alloc_space http://10.22.205.211:6060/debug/pprof/heap
Fetching profile over HTTP from http://10.22.205.211:6060/debug/pprof/heap
Saved profile in /home/haibzhou/pprof/pprof.coredns.alloc_objects.alloc_space.inuse_objects.inuse_space.012.pb.gz
coredns
Type: alloc_space
Time: Jan 28, 2019 at 4:05pm (GMT+7)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 1336.06MB, 95.87% of 1393.59MB total
Dropped 55 nodes (cum <= 6.97MB)
Showing top 10 nodes out of 71
      flat  flat%   sum%        cum   cum%
  313.02MB 22.46% 22.46%   313.02MB 22.46%  strings.Join
  264.61MB 18.99% 41.45%   264.61MB 18.99%  bytes.makeSlice
  138.01MB  9.90% 51.35%   245.51MB 17.62%  github.com/coredns/coredns/vendor/k8s.io/api/core/v1.(*EndpointAddress).Unmarshal
  124.94MB  8.97% 60.32%   370.46MB 26.58%  github.com/coredns/coredns/vendor/k8s.io/api/core/v1.(*EndpointSubset).Unmarshal
[...]

PR #2513, opened 2019-01-29
plugin/kubernetes: Don't push watch updates when there are no watches

chrisohaver:

Do not calculate and push watch updates to watch channel, when no one is listening.

#2511 

none

### DIFF
--- a/plugin/kubernetes/watch.go
+++ b/plugin/kubernetes/watch.go
@@ -96,10 +96,16 @@ func (dns *dnsControl) sendUpdates(oldObj, newObj interface{}) {
 	switch ob := obj.(type) {
 	case *object.Service:
 		dns.updateModifed()
+		if len(dns.watched) == 0 {
+			return
+		}
 		dns.sendServiceUpdates(ob)
 	case *object.Endpoints:
 		if newObj == nil || oldObj == nil {
 			dns.updateModifed()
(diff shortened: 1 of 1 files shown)


zhouhaibing089:
Let me test it real. :) Thanks for the quick fix.

miekg:
this is the 2nd time that k8s watches create problems. Who's using this? Is this worth having? /cc @johnbelamaric

johnbelamaric:
I am not sure who is using it, but we have discussed possibly using it in
the context of the node-local DNS in Kubernetes.

In this case, it's actually the same problem as before. We backed out (or
maybe never merged) the change that fixed it before, because it also messed
up the serial updates for zone transfers. This latest change was more
surgical. The code no longer executes if there are no watches defined, it
should be quite safe to keep it.

John

miekg:
>In this case, it's actually the same problem as before. We backed out (or
>maybe never merged) the change that fixed it before, because it also messed
>up the serial updates for zone transfers. This latest change was more
>surgical. The code no longer executes if there are no watches defined, it
>should be quite safe to keep it.

Ok. But this essentially saying: no-one is using it, but it could be useful... 
Meanwhile every one looking at this code, needs to be aware of it's existence. 
That's not a good tradeoff. 

I'm going to propose we remove it for now and add it back in when there is an 
actual user.

Issue #2471, opened 2019-01-16
plugin/auto not reloading files


stp-ip:
Marking this down as a possible issue for now.
Will debug and try to pinpoint this sooner than later.

stp-ip:
So this seems to fail reloading:
```
      auto {
        directory /etc/source/files/zones/ (.*) {1} 30
        transfer to *
      }
```

This works:
```
      auto {
        directory /etc/source/files/zones/ (.*) {1}
        reload 30s
        transfer to *
      }
```

Debugging...

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/auto n..." ]
>So this seems to fail reloading:
>```
>      auto {
>        directory /etc/source/files/zones/ (.*) {1} 30
>        transfer to *
>      }
>```
>
>This works:
>```
>      auto {
>        directory /etc/source/files/zones/ (.*) {1}
>        reload 30s
>        transfer to *
>      }
>```
>
>Debugging...

geh. Odd. Curious to any outcomes here.

stp-ip:
Had a bit to take another stab at it.

So a few things to write it down already:
* `reload` is the interval to reload a zone. So if it changes on disk, the change will be read in.
* `timeout` is the interval to walk the directory. So if a file/zone got added or removed, the zone would be added or removed.
* we only test removing a file and switching symlinks within the watcher tests

The file plugin takes care of checking if a file is updated.

So they are needed in combination. While writing this #2516 comes up with defaulting it within auto, which sets the reload timer so file is actually checking it.
That way the removal and addition of files is checked (via auto) and the actual zones or SOA is checked (via file).

miekg:
this is now fixed?

stp-ip:
Yes fixed and in the process of moving to a simpler config.

PR #2516, opened 2019-01-30
plugin/auto: Reload zones every one minute

mrasu:
However README says "Default is one minute.", zone data is not reloaded without declaring `reload` option explicitly.
By this change, data is reloaded every one minute by default.

#2471 

None

### DIFF
--- a/plugin/auto/README.md
+++ b/plugin/auto/README.md
@@ -29,11 +29,12 @@ are used.
   used to extract the origin. **ORIGIN_TEMPLATE** will be used as a template for the origin. Strings
   like `{<number>}` are replaced with the respective matches in the file name, e.g. `{1}` is the
   first match, `{2}` is the second. The default is: `db\.(.*)  {1}` i.e. from a file with the
-  name `db.example.com`, the extracted origin will be `example.com`. **TIMEOUT** deprecated. 
-  `reload` will be used, if not defined, and it will be removed in a subsequent version
+  name `db.example.com`, the extracted origin will be `example.com`.
+  **TIMEOUT** is deprecated and will be removed in a subsequent version. 
+  `reload` will be used, if not defined
   (it specifies how often CoreDNS should scan the directory to watch for file removal and addition;
   the default is every 60 seconds. This value is in seconds. The minimum value is 1 second.)
-* `reload` interval to perform reload of zone if SOA version changes. Default is one minute.
+* `reload` interval to perform reloads of zones if SOA version changes and zonefiles. Default is one minute.
   Value of `0` means to not scan for changes and reload. eg. `30s` checks zonefile every 30 seconds
   and reloads zone when serial changes.
 * `no_reload` deprecated. Sets reload to 0.
--- a/plugin/auto/README.md
+++ b/plugin/auto/README.md
@@ -29,9 +29,10 @@ are used.
   used to extract the origin. **ORIGIN_TEMPLATE** will be used as a template for the origin. Strings
   like `{<number>}` are replaced with the respective matches in the file name, e.g. `{1}` is the
   first match, `{2}` is the second. The default is: `db\.(.*)  {1}` i.e. from a file with the
-  name `db.example.com`, the extracted origin will be `example.com`. **TIMEOUT** specifies how often
-  CoreDNS should scan the directory; the default is every 60 seconds. This value is in seconds.
-  The minimum value is 1 second.
+  name `db.example.com`, the extracted origin will be `example.com`. **TIMEOUT** deprecated. 
+  `reload` will be used, if not defined, and it will be removed in a subsequent version
+  (it specifies how often CoreDNS should scan the directory to watch for file removal and addition;
+  the default is every 60 seconds. This value is in seconds. The minimum value is 1 second.)
 * `reload` interval to perform reload of zone if SOA version changes. Default is one minute.
   Value of `0` means to not scan for changes and reload. eg. `30s` checks zonefile every 30 seconds
   and reloads zone when serial changes.
--- a/plugin/auto/setup.go
+++ b/plugin/auto/setup.go
@@ -77,16 +77,18 @@ func setup(c *caddy.Controller) error {
 }
 
 func autoParse(c *caddy.Controller) (Auto, error) {
+	nilInterval := -1 * time.Second
 	var a = Auto{
-		loader: loader{template: "${1}", re: regexp.MustCompile(`db\.(.*)`)},
-		Zones:  &Zones{},
+		loader: loader{
+			template:       "${1}",
+			re:             regexp.MustCompile(`db\.(.*)`),
+			ReloadInterval: nilInterval,
+			duration:       nilInterval,
+		},
+		Zones: &Zones{},
 	}
 
 	config := dnsserver.GetConfig(c)
-
-	nilInterval := -1 * time.Second
-	configReloadInterval := nilInterval
-	configDuration := nilInterval
 
 	for c.Next() {
 		// auto [ZONES...]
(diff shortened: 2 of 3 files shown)


stp-ip:
This fixes part of the problem. Question is, if this should be fixed differently.

I agree it should set a default, but we might want to fix a bit more.

* [x] fix documentation that `reload` is the SOA reload of changed files and `timeout` is the removal and addition timer
* [x] use `timeout` or `reload` to set the other, if not both are set (better default as this seems expected)
* [x] optionally fix it and only rely on reload and deprecate `timeout`

mrasu:
@stp-ip Thanks for feedback!

Following your response, I change behavior of `timeout` and `reload`.
But because I don't know a way to "deprecate" in this project, I'm not sure my change is exactly what you says.
Are belows fine for you?

After my change:

* when `reload` and `timeout` are set -> both use the value of `reload`.
* when only `reload` is set -> both use the value of `reload`.
* when only `timeout` is set -> both use the value of `timeout`.
* when `reload` and `timeout` are not set -> both are one minute.

If this is fine, I will update documentation.

stp-ip:
Paging @miekg as this might be a breaking or deprecation change depending on the decision.

> when reload and timeout are set -> both use the value of reload.

I would say, if both are set, that might be intentional. So overwriting timeout with reload might create unexpected outcomes until it's deprecated.
So first deprecate and then overwrite timeout with reload.

>when only reload is set -> both use the value of reload.

+1

>when only timeout is set -> both use the value of timeout.

+1

> when reload and timeout are not set -> both are one minute

+1

In general I think we should deprecate TIMEOUT.
The general process in my mind:
* document deprecation "reload will be used, if not defined, and it will be removed in a subsequent version"
* print log message that TIMEOUT within plugin/auto is deprecated and RELOAD should be used + a link to the docs
* 1st next release according to deprecation policy (https://github.com/coredns/coredns#deprecation-policy) overwrite TIMEOUT always with RELOAD
* 2nd next release according to deprecation policy remove TIMEOUT and let the config fail, if it's set

mrasu:
@stp-ip done!

plugin/auto/README.md
```diff
@@ -29,9 +29,10 @@ are used.
   used to extract the origin. **ORIGIN_TEMPLATE** will be used as a template for the origin. Strings
   like `{<number>}` are replaced with the respective matches in the file name, e.g. `{1}` is the
   first match, `{2}` is the second. The default is: `db\.(.*)  {1}` i.e. from a file with the
-  name `db.example.com`, the extracted origin will be `example.com`. **TIMEOUT** specifies how often
-  CoreDNS should scan the directory; the default is every 60 seconds. This value is in seconds.
-  The minimum value is 1 second.
+  name `db.example.com`, the extracted origin will be `example.com`. **TIMEOUT** deprecated.
```
miekg: The sentence starting with **TIMEOUT** looks broken

miekg:
we should just make sure reload = 60 is set and used. Why is this not the case?
We don't need to deprecate anything and add a new option, current set is fine.

stp-ip:
>we should just make sure reload = 60 is set and used. Why is this not the case?
>We don't need to deprecate anything and add a new option, current set is fine.

There are 2 options both with different outcomes.
The idea is to remove TIMEOUT or at least deprecate it. No need to keep a config value, that we basically overwrite in the long term.
My current idea is that, if TIMEOUT is set to something this should still take effect. Then we deprecate and later remote or overwrite the TIMEOUT with the value of RELOAD to make the transition easier.

So the main way to define both reloading files and reloading records is done via RELOAD.

The 60s default was in there, but was changed to nil. Not sure why so? Defaulting to 60s seems fine.

stp-ip:
Correction. The defaulting to 60s is done. It was just moved.

TL;DR:
If TIMEOUT is set, but not RELOAD -> it works
If RELOAD is set, but not TIMEOUT -> it works
If none is set, we default to 60s -> it works
If TIMEOUT and RELOAD is set, we honor the settings -> it works depending on user settings

mrasu:
Actually, my first commit (https://github.com/coredns/coredns/pull/2516/commits/c59bc4aba0a92ffda1fcba91e56d73a479f6cfea) just set 60 seconds as a default.
I deprecate TIMEOUT following with advice.
So either is fine

Honestly, because two values are for almost same purpose, I think it's fair to deprecate TIMEOUT and use the same configuration for watching files too.
But because I'm pretty new to this repository, I cannot judge whether is better.
So I'd like you to decide which is better.

miekg:
ok, I agree with deprecating TIMEOUT, we can offically drop no_reload as well at the same time. Need a release cycle (to announce and then drop) for it to become active. 

What should I write in the release notes?

stp-ip:
> Honestly, because two values are for almost same purpose, I think it's fair to deprecate TIMEOUT and use the same configuration for watching files too.
> But because I'm pretty new to this repository, I cannot judge whether is better.
> So I'd like you to decide which is better.

I think we agree that the purpose is similar, but changing it how it might be more logical might still break people. This is a change in how it works today so having this backwards compatible until it's deprecated seems better.

I'll test this again this week.

stp-ip:
> ok, I agree with deprecating TIMEOUT, we can offically drop no_reload as well at the same time. Need a release cycle (to announce and then drop) for it to become active.
> 
> What should I write in the release notes?

Suggestion:
`Deprecation: plugin/auto deprecates TIMEOUT and NO_RELOAD and recommends the use of RELOAD. This brings plugin/file and plugin/auto to use the same syntax for reload intervals.`

* [x] test current implementation on backwards compatibility and deprecation (me)
* [x] implementation
* [x] docs
* [x] release notes
* [x] create issue for next release to remove deprecated code (including NO_RELOAD from plugin/file)

mrasu:
Sorry for late reply. I'm sick these days(ful?).
I think I cannot contribute until weekend.
please feel free to commit or create new PR if needed.

:+1:

stp-ip:
> Sorry for late reply. I'm sick these days(ful?).
> I think I cannot contribute until weekend.
> please feel free to commit or create new PR if needed.

Get well. I'll take the time to test this PR and then we can iterate next week.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/auto: ..." ]
>> ok, I agree with deprecating TIMEOUT, we can offically drop no_reload as well at the same time. Need a release cycle (to announce and then drop) for it to become active.
>>
>> What should I write in the release notes?
>
>Suggestion:
>`Deprecation: plugin/auto deprecates TIMEOUT and NO_RELOAD and recommends the use of RELOAD. This brings plugin/file and plugin/auto to use the same syntax for reload intervals.`

sgtm

plugin/auto/README.md
stp-ip:
```suggestion
  name `db.example.com`, the extracted origin will be `example.com`. **TIMEOUT** is deprecated and will be removed in a subsequent version.
```

plugin/auto/README.md
```diff
@@ -29,9 +29,10 @@ are used.
...
   like `{<number>}` are replaced with the respective matches in the file name, e.g. `{1}` is the
   first match, `{2}` is the second. The default is: `db\.(.*)  {1}` i.e. from a file with the
-  name `db.example.com`, the extracted origin will be `example.com`. **TIMEOUT** specifies how often
-  CoreDNS should scan the directory; the default is every 60 seconds. This value is in seconds.
-  The minimum value is 1 second.
+  name `db.example.com`, the extracted origin will be `example.com`. **TIMEOUT** deprecated. 
+  `reload` will be used, if not defined, and it will be removed in a subsequent version
+  (it specifies how often CoreDNS should scan the directory to watch for file removal and addition;
+  the default is every 60 seconds. This value is in seconds. The minimum value is 1 second.)
 * `reload` interval to perform reload of zone if SOA version changes. Default is one minute.
```
stp-ip:
```suggestion
 * `reload` interval to perform reloads of zones if SOA version changes and zonefiles. Default is one minute.
```

stp-ip (changes requested):
Docs changes suggested.

mrasu:
@stp-ip Thanks! changed!

mrasu:
@stp-ip ping

stp-ip (approved):

stp-ip:
Thanks for the ping. This seems to have slipped through my notifications.

/approve
/lgtm

Issue #2518, opened 2019-01-31
plugin/kubernetes: ttl for different service type


ckcd:
Hi,

I have a question about `ttl` of services with different type:

In Kubernetes, dns `A` record of a `clusterIP service` is the IP address itself, but dns `A` record of a `headless service` is a list of endpoints addresses.

In my rough opinion, the `clusterIP service` is stable and we won't change it frequently, and it changes only when the service be deleted.  and it's IP address is stable which means the dns record will not change.

Meanwhile, the dns record of `headless service` is unstable because the `endpoints` will change when `pod` changes, and in k8s `pod` will be frequent kill and start so it's ip address has changed.

So i'm thinking about set different `ttl` for these different services. which may be present as:
1. for default ttl, give `clusterIP service` a long ttl (maybe 60s?) and give `headless service` a short ttl (maybe 5s?)
2. give more parameter of `plugin/kubernetes` in `Corefile`, so users can defines different ttl for different services.

I'm not sure whether it's necessary, please correct me if i have some misunderstanding.

Thanks !!!

chrisohaver:
>I'm not sure whether it's necessary

TL;DR: I think it could benefit node local dns cache deployments.

In CoreDNS, the _kubernetes_ plugin keeps its own cache, which is updated in "real time" asynchronously via a watch on the Kubernetes API.  So, even if you deploy _kubernetes_ plugin with a TTL of 0, or if you deploy CoreDNS with no _cache_ plugin at all, queries for Kubernetes records do not cause round trips to the K8s api per request.  In other words, there is not a significant performance difference when pulling a record from the _cache_ plugin vs the _kubernetes_ plugin.  So, from a coredns performance perspective, the TTL on _kubernetes_ records doesn't really factor in strongly.

But TTL would come into play with clients that might pay attention to the TTL, or if the the client reaching coredns through a local caching DNS, such as would happen with node local dns caching in k8s (which i think is becoming more common and may become standard soon).  In a node local dns caching deployment, a higher TTL for cluster ip services would reduce DNS traffic to the cluster DNS.

> give more parameter of `plugin/kubernetes` in Corefile, so users can defines different ttl for different services.

I think we could do this by extending the existing `ttl` directive, allowing an optional second parameter.
The second parameter if specified being the TTL to use for Cluster IP services, the first used for Pods, Endpoints, headless Services, and negative responses.

ckcd:
@chrisohaver 
Thanks for your reply !

[...]

PR #2519, opened 2019-02-01
plugin/kubernetes: set special ttl for clusterIP service

ckcd:

Unlike other elements, `clusterIP service` is stable and won't change it IP address unless been deleted, so maybe we can set a special ttl for it. So to help the performance and benefit node local dns cache deployments.

https://github.com/coredns/coredns/issues/2518

plugin/kubernetes/README.md

### DIFF
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -40,7 +40,7 @@ kubernetes [ZONES...] {
     pods POD-MODE
     endpoint_pod_names
     upstream [ADDRESS...]
-    ttl TTL
+    ttl TTL [clusterIPTTL]
     noendpoints
     transfer to ADDRESS...
     fallthrough [ZONES...]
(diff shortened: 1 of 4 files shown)


ckcd:
Sorry for the force push, 

in first push i set a different default value for `clusterIPTtl` but it failed in ci-test, i think this need some change in https://github.com/coredns/ci/tree/master/test/kubernetes , so i cancel this modify.

plugin/kubernetes/setup.go
```diff
@@ -252,6 +252,18 @@ func ParseStanza(c *caddy.Controller) (*Kubernetes, error) {
 				return nil, c.Errf("ttl must be in range [0, 3600]: %d", t)
 			}
 			k8s.ttl = uint32(t)
+
+			// set clusterIPTtl
+			if len(args) >= 2 {
```
chrisohaver: If there are > 2 TTLs provided, we should error.

plugin/kubernetes/README.md
```diff
@@ -40,7 +40,7 @@ kubernetes [ZONES...] {
     pods POD-MODE
     endpoint_pod_names
     upstream [ADDRESS...]
-    ttl TTL
+    ttl TTL [clusterIPTTL]
```
chrisohaver: To be consistent with surrounding style ...

```suggestion
    ttl TTL [CLUSTER-IP-TTL]
```

plugin/kubernetes/README.md
```diff
@@ -88,7 +88,8 @@ kubernetes [ZONES...] {
   that point to external hosts (aka External Services, aka CNAMEs).  If no **ADDRESS** is given, CoreDNS
   will resolve External Services against itself. **ADDRESS** can be an IP, an IP:port, or a path
   to a file structured like resolv.conf.
-* `ttl` allows you to set a custom TTL for responses. The default is 5 seconds.  The minimum TTL allowed is
+* `ttl` allows you to set a custom TTL for responses and a optional clusterIPTTL for clusterIP service
```
chrisohaver:
```suggestion
* `ttl` allows you to set a custom TTL for responses and a optional **CLUSTER-IP-TTL** for clusterIP service
```

chrisohaver (dismissed):

miekg:
if you can't set some of this in the API of k8s we should not add it here. Any k8s related should be configured with CRD not fiddling with Corefile configs.

Also this is a spec change, so the DNS spec should be updated to reflect that this is thing we do.
And another thing is that a TTL is just a hint to be ignored at will.

miekg:
Closing as this is not the right approach as I eluded in my comment.

PR #2527, opened 2019-02-04
cache: split out items

miekg:
Make each of the structs we store smaller, this should cut back on
memory use. The included benchmark shows same *number* of allocations,
but those allocation are smaller (still need to get that data out of the
test).

Cache was already using interface{} so this slit in nicely.

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -199,19 +199,26 @@ func (w *ResponseWriter) WriteMsg(res *dns.Msg) error {
 }
 
 func (w *ResponseWriter) set(m *dns.Msg, key uint64, mt response.Type, duration time.Duration) {
-	// duration is expected > 0
-	// and key is valid
+	// duration is expected > 0 and key to be valid
 	switch mt {
-	case response.NoError, response.Delegation:
-		i := newItem(m, w.now(), duration)
+	case response.NoError:
+		i := &noErrorItem{}
+		i.fromMsg(m, w.now(), duration)
+		w.pcache.Add(key, i)
+
+	case response.Delegation:
+		i := &delegationItem{}
+		i.fromMsg(m, w.now(), duration)
 		w.pcache.Add(key, i)
 
 	case response.NameError, response.NoData:
-		i := newItem(m, w.now(), duration)
+		i := &nameErrorItem{}
+		i.fromMsg(m, w.now(), duration)
 		w.ncache.Add(key, i)
 
 	case response.OtherError:
 		// don't cache these
+
 	default:
 		log.Warningf("Caching called with unknown classification: %d", mt)
 	}
--- a/plugin/cache/cache_memory_test.go
+++ b/plugin/cache/cache_memory_test.go
@@ -1,0 +1,54 @@
+package cache
+
+import (
+	"context"
+	"testing"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/test"
+
+	"github.com/miekg/dns"
+)
+
+func BenchmarkCacheMemory(b *testing.B) {
+	c := New()
+	c.Next = BackendMemoryHandler()
+	ctx := context.TODO()
+
+	names := []string{"example1", "example2", "a", "b", "c", "d", "e", "f", "g", "h", "i", "A", "B", "C", "D", "E", "F", "G", "H"}
+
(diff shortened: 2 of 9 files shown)


miekg:
@grobie do you have time to look this over?

grobie:
Thanks for that PR! Do you have any numbers showing the improvements in terms of memory needs?

miekg:
Sadly no, the benchmark shows _allocations_ not bytes allocated; I can do a one-off tests, but it would be much, *much* better to have a GitHub integration that reports this (although that more than a full weeks worth to setup)

miekg:
merge? @grobie

miekg:
I'll probably go ahead and merge this, after I self review it once again.

chrisohaver:
IMO, especially given the currently active issues (memory/cpu spike concerns), it would be better to quantify how much this change affects memory and cpu utilization, before merging.

miekg:
Sure. How and where?

chrisohaver:
Packet is a good resource for this type of testing. 

E.g. Do a load test using previous cache, measure cpu/mem.  Then do same load test using new cache, measure cpu/mem. Compare.

chrisohaver:
code lgtm fwiw

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] cache: split ..." ]
>Packet is a good resource for this type of testing.
>
>E.g. Do a load test using previous cache, measure cpu/mem.  Then do same load test using new cache, measure cpu/mem. Compare.

Sure. Doing these one-offs don't really help. What would you need to see and 
would you be willing to commit cycles to make something better?

chrisohaver:
By "something better" do you mean automated scale testing?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] cache: split ..." ]
>By "something better" do you mean automated scale testing?

Yes, or at least the metrics/data you're asking for. We have benchmark tests 
(this PRs adds them), but they can't be used in Travis.

Also cpu usage/global memory usage and not easy extractable.

chrisohaver:
An automated scale testing framework/setup doesn't matter in the context of this PR (because that would take a very long time to build, and this PR will have long since been merged before the framework is complete).

IMO, it would be _better_ to quantify the affect on memory footprint and cpu ... but not _required_.  But sounds like you are saying that it's not worth the effort to do a one-off test of memory/cpu footprint, because you have already added benchmarks. ok

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] cache: split ..." ]
>An automated scale testing framework/setup doesn't matters in the context of this PR (because that would take a very long time to build, and this PR will have long since been merged before the framework is complete).

Not true, it matters and has mattered before.

You're side stepping my question.

>IMO, it would be _better_ to quantify the affect on memory footprint and cpu ... but not _required_.  But sounds like you are saying that it's not worth the effort to do a one-off test of memory/cpu footprint, because you have already added benchmarks. ok

There are benchmarks added in this pr, but they show allocations. Not sure if 
they can be made to show more.

johnbelamaric:
On Tue, Mar 12, 2019 at 2:20 PM Miek Gieben <notifications@github.com>
wrote:

> [ Quoting <notifications@github.com> in "Re: [coredns/coredns] cache:
> split ..." ]
> >An automated scale testing framework/setup doesn't matters in the context
> of this PR (because that would take a very long time to build, and this PR
> will have long since been merged before the framework is complete).
>
> Not true, it matters and has mattered before.
>
>
I think Chris just meant that by the time we built that, this PR would have
already merged. It's still a good thing to build.

chrisohaver:
> I think Chris just meant that by the time we built that, this PR would have
already merged.

lol.  That is what I meant ... because thats literally what I wrote. But thanks for the clarification. :)

miekg:
Thanks. I'll take that as a 'no' then...

grobie:
I agree with @chrisohaver and would like to see some data baking the need for this change. A one-time test showing a significant reduction of memory RSS with change compared to current mater would be helpful already. I understand the wish for an automated suite which permanently monitors that, but I don't see this as a requirement for this PR.

Without any data it's not clear whether the increased complexity is really worth it.

miekg:
See https://github.com/coredns/coredns/pull/2694
This is already printing odd things (in master and on this branch), for total memory used (for which I need to read the actual docs of runtime)

Issue #2526, opened 2019-02-04
plugin/k8s_external - Support for zone transfer


fturib:
That would need:
- define what would be the SERIAL for SOA
- collect all Services using Externaler interface, and generate all the corresponding records.

NOTE: I have prepared locally a change that seems OK (not sure though).
Will shoot a PR later today.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin/k8s_extern..." ]
>That would need:
>- define what would be the SERIAL for SOA

this is already done, as we support axfr in k8s already.

>- collect all Services using Externaler interface, and generate all the corresponding records.

interface? why?

/Miek

--
Miek Gieben

johnbelamaric:
> this is already done, as we support axfr in k8s already.

You only need to bump the serial number if services of type loadbalancer or
with ExternalIPs change. So that should be a much smaller set of changes
than in the normal k8s plugin.

miekg:
see discussion in #2585 a new plugin is better.

PR #2528, opened 2019-02-04
[WIP] plugin/k8s_external implement transfer zone for k8s_external

fturib:
I need to be able to serve a secondary server from the Authoritative server of the public side of the Cluster.
In other word, I need AXFR to be implemented in `k8s_external` plugin

#2526 

oops .. will update on next commit

WIP, still todo:
- Update the README with 'tranferTo' and the support for AXFR/IXFR
- Implement a smart SERIAL for k8s_external
- add UT for Transfer support

### DIFF
--- a/plugin/k8s_external/README.md
+++ b/plugin/k8s_external/README.md
@@ -29,9 +29,8 @@ ns1.dns.example.org.  5 IN  A    ....
 ns1.dns.example.org.  5 IN  AAAA ....
 ~~~
 
-Note we use the `dns` subdomain to place the records the DNS needs (see the `apex` directive). Also
-note the SOA's serial number is static. The IP addresses of the nameserver records are those of the
-CoreDNS service.
+Note we use the `dns` subdomain to place the records the DNS needs (see the `apex` directive).
+The IP addresses of the nameserver records are those of the CoreDNS service.
 
 The *k8s_external* plugin handles the subdomain `dns` and the apex of the zone by itself, all other
 queries are resolved to addresses in the cluster.
@@ -51,11 +50,17 @@ this extended syntax.
 k8s_external [ZONE...] {
     apex APEX
     ttl TTL
+    transfer to ADDRESS...
 }
 ~~~
 
 * **APEX** is the name (DNS label) to use the apex records, defaults to `dns`.
 * `ttl` allows you to set a custom **TTL** for responses. The default is 5 (seconds).
+* `transfer` enables zone transfers. It may be specified multiples times. `To` signals the direction
+  (only `to` is allow). **ADDRESS** must be denoted in CIDR notation (127.0.0.1/32 etc.) or just as
+  plain addresses. The special wildcard `*` means: the entire internet.
+  Sending DNS notifies is not supported.
+
 
 # Examples
 
--- a/plugin/k8s_external/external.go
+++ b/plugin/k8s_external/external.go
@@ -38,6 +38,7 @@ type External struct {
 	hostmaster string
 	apex       string
 	ttl        uint32
+	transferTo []string
 
 	externalFunc     func(request.Request) ([]msg.Service, int)
 	externalAddrFunc func(request.Request) []dns.RR
--- a/plugin/k8s_external/msg_to_dns.go
+++ b/plugin/k8s_external/msg_to_dns.go
@@ -61,10 +61,7 @@ func (e *External) aaaa(services []msg.Service, state request.Request) (records 
 	return records
 }
 
-func (e *External) srv(services []msg.Service, state request.Request) (records, extra []dns.RR) {
-	dup := make(map[item]struct{})
-
-	// Looping twice to get the right weight vs priority. This might break because we may drop duplicate SRV records latter on.
+func buildWeights(services []msg.Service) map[int]int {
 	w := make(map[int]int)
 	for _, s := range services {
 		weight := 100
@@ -77,6 +74,13 @@ func (e *External) srv(services []msg.Service, state request.Request) (records, 
 		}
 		w[s.Priority] += weight
 	}
+	return w
+}
+
+func (e *External) srv(services []msg.Service, state request.Request) (records, extra []dns.RR) {
+	dup := make(map[item]struct{})
+	w := buildWeights(services)
+	// Looping twice to get the right weight vs priority. This might break because we may drop duplicate SRV records latter on.
 	for _, s := range services {
 		// Don't add the entry if the port is -1 (invalid). The kubernetes plugin uses port -1 when a service/endpoint
 		// does not have any declared ports.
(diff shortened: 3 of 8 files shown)


plugin/k8s_external/external.go
```diff
@@ -90,11 +91,11 @@ func (e *External) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 
 	switch state.QType() {
 	case dns.TypeA:
-		m.Answer = e.a(svc, state)
+		m.Answer = e.a(svc, &state)
```
miekg: state is not a pointer anywhere in coredns, why here?

plugin/k8s_external/xfr.go
```diff
@@ -0,0 +1,84 @@
...
+
+// Transfer implements the Transferer interface.
+func (e *External) Transfer(ctx context.Context, state request.Request) (int, error) {
+
+	if !e.transferAllowed(state) {
+		return dns.RcodeRefused, nil
+	}
+
+	serviceList, _ := e.externalFunc(state)
+	records, _ := e.srv(serviceList, nil)
```
miekg: can we just make a new function that returns the all the records we need?

plugin/k8s_external/external.go
fturib: See the change in corresponding functions : e.a(...), e.aaa(...). e.srv(...):
- the domain for the A record is usually the one of the `state`, but in case of transfer, it has to the be the one of the `svc.QName()`.
(Unless I am wrong here ... and there is another way).

So either I send state in pointer, either I change signature of those functions, either I duplicate those functions for AXFR use case.

plugin/k8s_external/xfr.go
fturib: you mean a new function e.axfr(state) ?

- Maybe that will resolve the point above with pointer of state.

miekg: I think that makes sense, and hopefully also minimizes the changes in k8s

fturib:
Pushed s sqush merge of commits. That includes:
- add setup for "transfer" and update README
- adding AXFR support for k8s_external (with its own translation of the msg.Service)
- change SERIAL of SOA to be timestamp instead of fixed value
- add building the list of All Svc in k8s/external
- fix generation of A record for NS (in k8s/ns)

Needs to do:
- add computation of a proper SERIAL (based on change in the services)
- UT

plugin/k8s_external/README.md
```diff
@@ -51,11 +50,17 @@ this extended syntax.
...
     apex APEX
     ttl TTL
+    transfer to ADDRESS...
 }
 ~~~
 
 * **APEX** is the name (DNS label) to use the apex records, defaults to `dns`.
 * `ttl` allows you to set a custom **TTL** for responses. The default is 5 (seconds).
+* `transfer` enables zone transfers. It may be specified multiples times. `To` signals the direction
+  (only `to` is allow). **ADDRESS** must be denoted in CIDR notation (127.0.0.1/32 etc.) or just as
```
miekg, line 60: allowed

plugin/k8s_external/msg_to_dns.go
```diff
@@ -61,10 +61,7 @@ func (e *External) aaaa(services []msg.Service, state request.Request) (records
 	return records
 }
 
-func (e *External) srv(services []msg.Service, state request.Request) (records, extra []dns.RR) {
-	dup := make(map[item]struct{})
-
-	// Looping twice to get the right weight vs priority. This might break because we may drop duplicate SRV records latter on.
+func buildWeights(services []msg.Service) map[int]int {
```
miekg, line 64: do we need this? I.e. can there be multiple msg where we need to calculate srv weights?

Also this change seems to be unrelated to the functionality of this pr, so please split it off and make it a global change in all plugins that do this

plugin/k8s_external/setup.go
```diff
@@ -3,6 +3,8 @@ package external
 import (
 	"strconv"
 
+	prs "github.com/coredns/coredns/plugin/pkg/parse"
+
```
miekg, line 7: has extra newline and why are you rewriting the import?

plugin/k8s_external/xfr.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/coredns/coredns/plugin/pkg/log"
+	"github.com/coredns/coredns/request"
+
+	"github.com/miekg/dns"
+)
+
+const transferLength = 2000
+
+// Serial implements the Transferer interface.
+func (e *External) Serial(state request.Request) uint32 { return uint32(time.Now().Unix()) }
```
miekg, line 17: In your PR description you say this is "smart", here it just takes time.Now(). So it's not smart?

plugin/k8s_external/xfr.go
```diff
@@ -0,0 +1,94 @@
...
+
+	tr.Out(state.W, state.Req, ch)
+	// Defer closing to the client
+	state.W.Hijack()
+	return dns.RcodeSuccess, nil
+}
+
+// transferAllowed checks if incoming request for transferring the zone is allowed according to the ACLs.
+// Note: This is copied from zone.transferAllowed, but should eventually be factored into a common transfer pkg.
+func (e *External) transferAllowed(state request.Request) bool {
```
miekg, line 78: prime candidate to factor this out (soonish), we have 3 of these (or more) identical functions in our plugins.

plugin/kubernetes/ns.go
```diff
@@ -23,6 +23,8 @@ func (k *Kubernetes) nsAddr() *dns.A {
 	)
 
 	rr := new(dns.A)
+	rr.Hdr.Class = dns.ClassINET
```
miekg, line 26: what does this fix?

plugin/kubernetes/external.go
```diff
@@ -19,6 +19,10 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
 	segs := dns.SplitDomainName(base)
 	last := len(segs) - 1
 	if last < 0 {
+		// this is a request for the zone itself. Return all services that are in exposed namespaces
```
miekg, line 22: this is only correct if there is an ixfr/axfr request, this should return the apex of the zone, not all records.

plugin/kubernetes/external.go
```diff
@@ -67,22 +71,51 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
...
 
-				services = append(services, s)
-			}
+func (k *Kubernetes) externalAll(state request.Request) ([]msg.Service, int) {
+
+	serviceList := k.APIConn.ServiceList()
+
+	services := []msg.Service{}
+	zonePath := msg.Path(state.Zone, coredns)
+	rcode := dns.RcodeNameError
```
miekg, line 88: this means, when we don't find any services we return nxdomain for the zone it self? That's not correct, should be nodata (at least)

miekg:
> Pushed s sqush merge of commits. That includes:
> 
> * add setup for "transfer" and update README
> * adding AXFR support for k8s_external (with its own translation of the msg.Service)
> * change SERIAL of SOA to be timestamp instead of fixed value
> * add building the list of All Svc in k8s/external
> * fix generation of A record for NS (in k8s/ns)
> 
> Needs to do:
> 
> * add computation of a proper SERIAL (based on change in the services)
> * UT

ah, missed your todo on smart soa serial. that will cause load on the api server though

plugin/k8s_external/setup.go
fturib: you mean : why I name the import "prs" ?
=> there is a confustion with the function parse .. but i can rename that one. ok.

plugin/kubernetes/external.go
fturib: I guess that is the responsibility of the caller to reply the correct records depending IXFR/AXFR.
this function `External` here is only to return Services.

plugin/kubernetes/external.go
fturib: hum. Anyway this `externallAll` should just return the services.

plugin/kubernetes/ns.go
fturib: that fixes the NS record of the k8s_external.
the function nsAddr() is called by ExternalAddress(..), which provide the RR for NS. And those RR are lacking the Class and Type.

plugin/k8s_external/msg_to_dns.go
fturib: my PR is to add AXFR, not to change existing SRV function.
So I keep the same computation ... cannot say really how much that is useful or not.

fturib: but .. sure, I can consider sharing this function in all plugins that have the same computation. Will be in anotehr PR.

miekg:
this is not going anywhere and #2585 (a new plugin) is the better alternative.

PR #2535, opened 2019-02-07
plugin/hosts provide more configuration flexibility

thomas-mangin:

This patch adds few features to the host plugin
 * no-reverse (both as first argument on the plugin line and inline)
   disable the automatic generation of reserve entries for hosts
 * ttl <duration> (inline only atm)
   allows to change the default ttl (default 5 minutes)
 * reload <duration> (inline only atm)
   allows to change the reloading interval (default 5s)

N/A

README.md updated to reflect the new features

### DIFF
--- a/plugin/hosts/README.md
+++ b/plugin/hosts/README.md
@@ -41,6 +41,9 @@ PTR records for reverse lookups are generated automatically by CoreDNS (based on
 ~~~
 hosts [FILE [ZONES...]] {
     [INLINE]
+    ttl SECONDS
+    no_reverse
+    reload DURATION
     fallthrough [ZONES...]
 }
 ~~~
@@ -53,6 +56,9 @@ hosts [FILE [ZONES...]] {
 * **INLINE** the hosts file contents inlined in Corefile. If there are any lines before fallthrough
    then all of them will be treated as the additional content for hosts file. The specified hosts
    file path will still be read but entries will be overrided.
+* `ttl` change the DNS TTL of the records generated (forward and reverse). The default is 3600 seonds (1 hour).
+* `reload` change the period between each hostsfile reload. A time of zero seconds disable the feature. Examples of valid durations: "300ms", "1.5h" or "2h45m" are valid duration with units "ns" (nanosecond), "us" (or "µs" for microsecond), "ms" (millisecond), "s" (second), "m" (minute), "h" (hour).
+* `no_reverse` disable the automatic generation of the the `in-addr.arpa` or `ip6.arpa` entries for the hosts
 * `fallthrough` If zone matches and no record can be generated, pass request to the next plugin.
   If **[ZONES...]** is omitted, then fallthrough happens for all zones for which the plugin
   is authoritative. If specific zones are listed (for example `in-addr.arpa` and `ip6.arpa`), then only
--- a/plugin/hosts/hostsfile.go
+++ b/plugin/hosts/hostsfile.go
@@ -32,6 +32,26 @@ func absDomainName(b string) string {
 	return plugin.Name(b).Normalize()
 }
 
+type options struct {
+	// automatically generate IP to Hostname PTR entries
+	// for host entries we parse
+	autoReverse bool
+
+	// The TTL of the record we generate
+	ttl uint32
+
+	// The time between two reload of the configuration
+	reload time.Duration
+}
+
+func newOptions() *options {
+	return &options{
+		autoReverse: true,
+		ttl:         3600,
+		reload:      durationOf5s,
+	}
+}
+
 type hostsMap struct {
 	// Key for the list of literal IP addresses must be a host
 	// name. It would be part of DNS labels, a FQDN or an absolute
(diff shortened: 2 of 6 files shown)


plugin/hosts/README.md
```diff
@@ -39,8 +39,11 @@ PTR records for reverse lookups are generated automatically by CoreDNS (based on
 ## Syntax
 
 ~~~
-hosts [FILE [ZONES...]] {
+hosts [no-reverse] [FILE [ZONES...]] {
```
miekg: Sorry, just now taking a closer look.

Can we not have `no-reverse` here, only in the block below?

plugin/hosts/README.md
```diff
@@ -53,6 +56,9 @@ hosts [FILE [ZONES...]] {
 * **INLINE** the hosts file contents inlined in Corefile. If there are any lines before fallthrough
    then all of them will be treated as the additional content for hosts file. The specified hosts
    file path will still be read but entries will be overrided.
+* `ttl` change the DNS TTL of the records generated (forward and reverse). The default is 3600 seonds (1 hour).
+* `reload` change the period between each hostsfile reload. A time of zero seconds disable the feature. Examples of valid durations: "300ms", "1.5h" or "2h45m" are valid duration with units "ns" (nanosecond), "us" (or "µs" for microsecond), "ms" (millisecond), "s" (second), "m" (minute), "h" (hour).
+* `no-autoreverse` disable the automatic generation of the the `in-addr.arpa` or `ip6.arpa` entries for the hosts
```
miekg: `no-autoreverse` -> `no_reverse` ? We default to using underscores for these things.

plugin/hosts/hostsfile.go
```diff
@@ -32,6 +32,26 @@ func absDomainName(b string) string {
 	return plugin.Name(b).Normalize()
 }
 
+type hostsOptions struct {
```
miekg: just 'options' ?

plugin/hosts/hostsfile.go
```diff
@@ -106,7 +133,7 @@ func (h *Hostsfile) readHosts() {
 		return
 	}
 
-	newMap := h.parse(file, h.inline)
+	newMap := h.parse(file, h.hmap, h.options.autoReverse)
```
miekg: just give parse the whole options struct

plugin/hosts/setup.go
```diff
@@ -22,28 +23,37 @@ func init() {
...
 		return plugin.Error("hosts", err)
 	}
 
-	parseChan := make(chan bool)
+	parseChan := periodicHostsUpdate(&h)
 
 	c.OnStartup(func() error {
 		h.readHosts()
-
-		go func() {
```
miekg, line 36: why can't we just keep this structure and just not execute it when the duration is 0?

plugin/hosts/README.md
thomas-mangin: sure, it removes code. will do. done.

plugin/hosts/hostsfile.go
thomas-mangin: sure, I tried to match the name of HostsMap ..

plugin/hosts/hostsfile.go
thomas-mangin: elsewhere this function is used to merge the inline data with the file and the code is
```h.hmap = h.parse(r, inline, h.options.autoReverse)```
so changing the API would not allow this anymore AFAICS.

File: plugin/hosts/setup.go (hunk 5 above)
Comment (thomas-mangin) [reply plugin/hosts/setup.go:36]: If you are commenting about the un-inlining, I could re-inline the code but I thought having a function naming the feature improved readability and simplified the code flow of the main `setup` function.

I thought about the code structure and decided for this one as to not require us to perform the test both in the `OnStartup` and `OnShutdown`. This way the test is within the code block but from the outside the behaviour is consistent.

What did you have in mind as change ?

File: plugin/hosts/hostsfile.go (hunk 3 above)
Comment (thomas-mangin) [reply plugin/hosts/hostsfile.go]: done, however in `setup` we now have a variable with the same name as the struct shadowing it.

File: plugin/hosts/hostsfile.go [hunk 6]
```diff
@@ -32,6 +32,26 @@ func absDomainName(b string) string {
 	return plugin.Name(b).Normalize()
 }
 
+type Options struct {
```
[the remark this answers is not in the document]
thomas-mangin: oops, fixed.

plugin/hosts/README.md
thomas-mangin: done.

plugin/hosts/hostsfile.go
miekg: I would just make that change, there is no backwards consideration for code living inside a plugin.

plugin/hosts/setup.go
miekg: fair enough

plugin/hosts/hostsfile.go
thomas-mangin: Ok, in that case, I think the code works only with the first parameter (removing the second and third). I first added a log.Fatalf in the `if override == nil {` before any change and none of the test code triggered it. It probably need a good pair of eyes to check this again. I left the change as a separate patch for ease of review.

miekg:
think this is good to go @pmoroney

pmoroney (approved):
LGTM thanks again for doing all this!

PR #2505, opened 2019-01-27
Upgrade caddy to 0.11.2

Xuanwo:
### 1. Why is this pull request needed and what does it do?

Caddy added `WrapListener` method in GracefulServer interface from 0.11.2, which made coredns plugin built failed. This pr fix it.

### 3. Which documentation changes (if any) need to be made?

Nope.

PR #2541, opened 2019-02-08
Upgrade caddy to 0.11.X

miekg:
Redo of #2505 (cherry-picked commit)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ godeps:
 	go get -u github.com/miekg/dns
 	go get -u github.com/prometheus/client_golang/prometheus/promhttp
 	go get -u github.com/prometheus/client_golang/prometheus
-	(cd $(GOPATH)/src/github.com/mholt/caddy              && git checkout -q v0.11.1)
+	(cd $(GOPATH)/src/github.com/mholt/caddy              && git checkout -q v0.11.2)
 	(cd $(GOPATH)/src/github.com/miekg/dns                && git checkout -q v1.1.4)
 	(cd $(GOPATH)/src/github.com/prometheus/client_golang && git checkout -q v0.9.1)
 
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ godeps:
 	go get -u github.com/miekg/dns
 	go get -u github.com/prometheus/client_golang/prometheus/promhttp
 	go get -u github.com/prometheus/client_golang/prometheus
-	(cd $(GOPATH)/src/github.com/mholt/caddy              && git checkout -q v0.11.2)
+	(cd $(GOPATH)/src/github.com/mholt/caddy              && git checkout -q v0.11.4)
 	(cd $(GOPATH)/src/github.com/miekg/dns                && git checkout -q v1.1.4)
 	(cd $(GOPATH)/src/github.com/prometheus/client_golang && git checkout -q v0.9.1)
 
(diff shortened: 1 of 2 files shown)


miekg:
failing with: 
```
--- FAIL: TestLookupCache (0.00s)
    cache_test.go:25: Could not get CoreDNS serving instance: unrecognized cluster plugin (was it included in the Caddy build?): file
~~~

**Comment by miekg** [comment]
Hmmm, @mholt this is an interesting failure...

**Comment by mholt** [comment]
It seems the caddytls package is not plugged in: https://sourcegraph.com/github.com/mholt/caddy@2ea544e9a03288118b4557d400dbc64031ae0198/-/blob/caddytls/setup.go#L39 - I suppose CoreDNS doesn't use the caddytls package?

I suppose I could change the logic [in here](https://sourcegraph.com/github.com/mholt/caddy@2ea544e9a03288118b4557d400dbc64031ae0198/-/blob/caddy.go#L484-487) to ignore a missing clustering plugin if the CADDY_CLUSTERING variable is not set...

**Comment by mholt** [comment]
Also, Caddy 0.11.2 is worth skipping. Most of its bugs are actually in the caddytls package, but 0.11.3 is out now.

**Comment by mholt** [comment]
@miekg I could fix this by moving all clustering plugin stuff to the `caddytls` package, so that plugins are registered with that package instead of with the `caddy` package, buuuut I'm not 100% sure that clustering plugins are going to only be used by the `tls` plugin in the future.

So, I think at this point I would recommend adding a default clustering plugin. You could do something like this:

```
caddy.RegisterClusterPlugin("file", func() (certmagic.Storage, error) {
	return &certmagic.FileStorage{Path: caddy.AssetsPath()}, nil
})
```

A clustering plugin basically defines how to access storage. Instances that share the same storage are part of the same cluster. (Right now it's for TLS assets but we may expand that in the future...)

File: Makefile [hunk 1]
```diff
@@ -31,7 +31,7 @@ godeps:
 	go get -u github.com/miekg/dns
 	go get -u github.com/prometheus/client_golang/prometheus/promhttp
 	go get -u github.com/prometheus/client_golang/prometheus
-	(cd $(GOPATH)/src/github.com/mholt/caddy              && git checkout -q v0.11.1)
+	(cd $(GOPATH)/src/github.com/mholt/caddy              && git checkout -q v0.11.2)
```
mholt:
```suggestion
	(cd $(GOPATH)/src/github.com/mholt/caddy              && git checkout -q v0.11.3)
```

miekg:
Thanks for looking into this. It doesn't sound too good though that we need
to work around things in caddy... The plugin should be complete standalone
and only affect coredns.

Can we find a way to decouple this? We only need the process management
side of things (mostly)

mholt:
@miekg I see your point. I guess that, since CertMagic storage is only dealt with directly in the caddytls package, it makes sense for clustering plugins to be a plugin to the tls plugin, not caddy core. 

Somewhat reluctantly (because I think I need to do more design work if we are to replace all of Caddy's storage interfacing), I have moved the clustering plugins solely into the caddytls package. Can you try building against https://github.com/mholt/caddy/pull/2459 ?

Then, when your PR builds pass, could you update the plugin on the Caddy website? Thanks!

miekg:
Waiting for caddy release with the fix: https://github.com/mholt/caddy/pull/2459

Xuanwo:
@miekg https://github.com/mholt/caddy/releases/tag/v0.11.4 has been released, maybe we can keep going for now.

miekg:
#2567 is needed to merge this one, as *reload* is racy

miekg:
merged both this one and #2567 locally and ran tests that seem to have done the trick - I'll merge the first and rebase this on top of that.

PR #2544, opened 2019-02-09
pkg/replace: make it more efficient.

miekg:
Remove the map that is allocated on every write and make it more static,
but just defining a function that gets called for a label and returns
its value.

Remove the interface definition and just implement what is needed in our
case. Add benchmark test for replace as well.

Extend metadata test to test multiple values (pretty sure this didn't
work, but there wasn't a test for it, so can't be sure).

Update all callers to use it - concurrent use should be fine as we pass
everything by value.

Benchmarks in replacer:

new: BenchmarkReplacer-4   300000      4717 ns/op     240 B/op       8 allocs/op
old: BenchmarkReplacer-4   300000      4368 ns/op     384 B/op      11 allocs/op

Added benchmark function to the old code to test it.

~~~
func BenchmarkReplacer(b *testing.B) {
	w := dnstest.NewRecorder(&test.ResponseWriter{})
	r := new(dns.Msg)
	r.SetQuestion("example.org.", dns.TypeHINFO)
	r.MsgHdr.AuthenticatedData = true
	b.ResetTimer()
	b.ReportAllocs()
	repl := New(context.TODO(), r, w, "")
	for i := 0; i < b.N; i++ {
		repl.Replace("{type} {name} {size}")
	}
}
~~~

New code contains (of course a different one). The amount of ops is
more, which might be good to look at some more. For all the allocations
is seems it was quite performant.

This looks to be 50% faster, and there is less allocations in log
plugin:

old: BenchmarkLogged-4   	   20000	     70526 ns/op
new: BenchmarkLogged-4   	   30000	     57558 ns/op

Fixes: #2293

No user visible backward incompatible changes are introduced.

### DIFF
--- a/plugin/pkg/replacer/replacer.go
+++ b/plugin/pkg/replacer/replacer.go
@@ -13,123 +13,133 @@ import (
 	"github.com/miekg/dns"
 )
 
-// Replacer is a type which can replace placeholder
-// substrings in a string with actual values from a
-// dns.Msg and responseRecorder. Always use
-// NewReplacer to get one of these.
-type Replacer interface {
-	Replace(string) string
-	Set(key, value string)
+type Replacer struct {
+	valueFunc func(request.Request, *dnstest.Recorder, string) string
+	labels    []string
 }
 
-type replacer struct {
-	ctx          context.Context
-	replacements map[string]string
-	emptyValue   string
+// labels are all supported labels that can be used in the default Replacer.
+var labels = []string{
+	"{type}",
+	"{name}",
+	"{class}",
+	"{proto}",
+	"{size}",
+	"{remote}",
+	"{port}",
+	"{local}",
+	// Header values.
+	headerReplacer + "id}",
+	headerReplacer + "opcode}",
+	headerReplacer + "do}",
+	headerReplacer + "bufsize}",
+	// Recorded replacements.
+	"{rcode}",
+	"{rsize}",
+	"{duration}",
+	headerReplacer + "rrflags}",
 }
 
-// New makes a new replacer based on r and rr.
-// Do not create a new replacer until r and rr have all
-// the needed values, because this function copies those
-// values into the replacer. rr may be nil if it is not
-// available. emptyValue should be the string that is used
-// in place of empty string (can still be empty string).
-func New(ctx context.Context, r *dns.Msg, rr *dnstest.Recorder, emptyValue string) Replacer {
-	req := request.Request{W: rr, Req: r}
-	rep := replacer{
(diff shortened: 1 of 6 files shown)


miekg:
Looks like replacer needs another test to increase coverage

plugin/pkg/replacer/replacer.go
```diff
@@ -13,123 +13,134 @@ import (
...
-	s = fscanAndReplace(s, headerReplacer, func(placeholder string) string {
-		return r.replacements[placeholder]
-	})
+	// Metadata label replacements. Scan for {/ and search for next }, replace that metadata label with
+	// any meta data that is available.
+	b := strings.Builder{}
+	for strings.Contains(s, labelReplacer) {
+		idxStart := strings.Index(s, labelReplacer)
+		endOffset := idxStart + len(labelReplacer)
+		idxEnd := strings.Index(s[endOffset:], "}")
```
fturib, line 124: Should the "}" be a constant (like "{/" is a constant) ?

plugin/pkg/replacer/replacer_test.go
```diff
@@ -87,46 +117,40 @@ func (m *testHandler) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns
...
-
 	tests := []struct {
 		expr   string
 		result string
 	}{
-		{"{name}", "example.org."},
-		{"{/test/key2}", "two"},
-		{"TYPE={type}, NAME={name}, BUFSIZE={>bufsize}, WHAT={/test/key2} .. and more", "TYPE=HINFO, NAME=example.org., BUFSIZE=512, WHAT=two .. and more"},
-		{"{/test/key2}{/test/key4}", "two-"},
-		{"{test/key2", "{test/key2"}, // if last } is missing, the end of format is considered static text
```
fturib, line 103: I guess it worth to keep the corner case of invalid format of the pattern.
=> at least verify that the code support invalid format.

fturib (approved):
Globally looks good to me. Made 2 small remarks. 
(clicked on approved by error. should wait the 2 remarks are addressed)

plugin/pkg/replacer/replacer.go
miekg: maybe, don't see the value of making the indirection here.

plugin/pkg/replacer/replacer_test.go
miekg: yep, let me add a test for that.

miekg:
thanks for your review; i;ll merge then ci is OK

Issue #2534, opened 2019-02-07
health: remove all per-plugin health status


miekg:
Using health to convey per plugin health status might make sense, but it get's tricky fast. Anything that requires network to be up, will in the event of a network hiccup report unhealthly. As this is a per process, the whole of coredns goes unhhealthy.

I..e. k8s can't connect to api -> unhealthy, while the cache might be perfectly fine to hand out requests in the mean time. 

Also when we report health OK; we're ready to process queries; if some plugin is not fully ready we return SERVFAIL, perfectly acceptable way of doing that in the DNS and the caller should know how to deal with that; retry; other server; retry again; etc.

IOW: per plugin health is useless and should be removed; all health code and be removed and concentrated in the (original) health plugin that says: We're up and running; some plugins might not be fully ready yet, but those will return SERVFAIL.

chrisohaver:
For k8s, would you suggest a new plugin to report readiness?
Or have the k8s plugin open a port and report it directly?

miekg:
Moving this to readiness might make a lot of sense. Have to figure out of health add anything at all, and what happens in k8s when readiness is toggling between on and off

miekg:
From k8s docs:

> Sometimes, applications are temporarily unable to serve traffic. For example, an application might need to load large data or configuration files during startup, or depend on external services after startup. In such cases, you don’t want to kill the application, but you don’t want to send it requests either. Kubernetes provides readiness probes to detect and mitigate these situations. A pod with containers reporting that they are not ready does not receive traffic through Kubernetes Services.

TL;DR: just a bad, cache might be perfectly fine in handling this, but we nack health for reasons and don't get any queries. That just leads to sadness.

PR #2547, opened 2019-02-09
plugin/health: remove ability to poll other plugins

miekg:
This mechanism defeats the purpose any plugin (mostly) caching can still
be alive, we can probably forward queries still. Don't poll plugins,
just tell the world we're up and running.

It was only actually used in kubernetes; and there specifically would
mean any network hiccup would NACK the entire server health.

Fixes: #2534


plugin/health/README.md
```diff
@@ -33,8 +29,8 @@ health [ADDRESS] {
 * Where `lameduck` will make the process unhealthy then *wait* for **DURATION** before the process
   shuts down.
 
-If you have multiple Server Blocks and need to export health for each of the plugins, you must run
-health endpoints on different ports:
+If you have multiple Server Blocks and need to export health, you must run health endpoints on
```
bradbeam: Would this still be the case if the health beck is being rolled up into a single check?

bradbeam (approved):

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/health..." ]
>bradbeam approved this pull request.
>
>
>
>> @@ -33,8 +29,8 @@ health [ADDRESS] {
> * Where `lameduck` will make the process unhealthy then *wait* for **DURATION** before the process
>   shuts down.
>
>-If you have multiple Server Blocks and need to export health for each of the plugins, you must run
>-health endpoints on different ports:
>+If you have multiple Server Blocks and need to export health, you must run health endpoints on
>
>Would this still be the case if the health beck is being rolled up into a single check?

There is really no need to do this, because it's back to health of the whole 
process. Delete it or expand that health is process wide and as-such can/should 
only be specified in one server block?

bradbeam:
Yeah I think it would be clearer By saying that the health check is process wide and should only be specified once.

chrisohaver:
From the k8s perspective, we need to replace the readiness function at the same time as removing it.  IOW, k8s should not move to this version of health (with the readiness function removed), until a new "readiness" plugin is added (or whatever we call it).

miekg:
readiness also implies don't send queries; I can't see how that should or can be used.

iow: this can be just merged and removes a large risk of saying coredns is unhealthy, while it's not.

chrisohaver:
fwiw, Im OK with splitting out readiness from health, as long as we are allowed to add the concept of readiness soon (e.g. a new readiness plugin). 

>I can't see how that should or can be used.

IMO it makes sense to only route queries to instances of coredns that are fully synced to the API, when there are some other instances that have not synced yet.  This could happen during rolling updates, wherein you don't want to continue a rollout to a new version of coredns until you know the new versions are successfully connected/synced to the API.

miekg:
the whole concept does not make sense if plugins are allowed to report back health and/or readiness; if that data is ORed together we will almost never go unhealthy; if's is ANDed together the smallest hiccup will cause the whole process to go unhealthy (or not ready - same difference). Doing it only during start up doesn't make sense either, because starting up isn't that special of a phase either.

chrisohaver:
It matters during a rolling update of the coredns Deployment, if for example a new version of CoreDNS is being rolled out, and that new version has some problem that prevents it from successfully connecting/syncing to the k8s API.  
If it reports "unready" in that state, then the rolling update will halt, seeing that the new version is never going "ready".
If it reports "ready" in that state, then the rolling update would continue to roll out the broken coredns to all instances, rendering `cluster.local` lookups non functional in the cluster.

johnbelamaric:
We added the "not healthy if k8s not synced" logic due to issues reported
by the field. If the sync takes a fair bit of time, but we say we are both
healthy and ready, then we get queries but are not ready, and incorrectly
respond or SERVFAIL.

This should have been done originally with a readiness check not a health
check. So, ultimately this is the right direction. But I agree
with @chrisohaver that K8s cannot move to this until the readiness check is
supported, or there will be production issues reported.

John

miekg:
ok, let's add health to all the other plugins as well then, rendering the whole thing useless... *kubernetes* is just a plugin; I don't see why this needs special treatment...

you get a perfectly fine SERVFAIL when it's not ready. 

Anything network related to report back health to the main instance is insane to support. A network hiccup and the whole process goes away - this is not how things should be.

Also: this came about from new issues "from the field".

chrisohaver:
Perhaps we can implement k8s readiness as an external plugin, and start doing a special build of coredns for k8s.

johnbelamaric:
Health and readiness aren't the same thing, at least not from a kubernetes
perspective. Readiness is more analogous to lame duck, but on start up. It
just means "don't send me traffic yet". I get that you don't want to
combine the plugins statuses together, because it's hard to understand what
that would mean. So, what about this:

1) Health goes back to global health for the process.
2) Add a readiness interface that plugins can choose to implement if they
want to.
3) Add a readiness plugin that has a per-plugin endpoint (e.g.,
/ready/kubernetes). Any plugin that implements the interface has this
endpoint and can use it to advertise their readiness. There is no global
"readiness".

This let's the user configuring the readiness check decide how do define
overall "readiness" as a function of the various plugins' readiness.
They're the only ones that really know their use case.

John

chrisohaver:
@johnbelamaric, agree on all points. Would prefer a generalized internal plugin. Each plugin would get its own path.  Or perhaps a list of plugins could be in the request, all statuses anded together...

miekg:
But even with readiness you're willing sacrifice the entire process health ? There are two things you most def. don't want:

* a single plugin determining the process in unhealthy
* doing so from with information you got over the network (you've basically created a query of death) - this is what the current health checks is

Per plugin health might be useful, but can't be used in k8s.

Where is it document that readyness check is only done on startup? Cause that might make it somewhat workable.

dbyron0:
Is https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes enough?

miekg:
Thanks @dbyron0, but that still doesn't fully specify if readyness is applied after a restart or during the entire lifetime of the pod. If that's the case you're still in the same boat; a single network hiccup kills all your pods

dbyron0:
Trying to dig through the kubelet code to find out the truth, but https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#container-v1-core says "Periodic probe of container service readiness."  Same for 1.12, 1.11 and 1.10.

I've convinced myself that readiness probes only happen on container startup, and liveness probes take over after that, but don't have concrete evidence of that yet.

More potential information here: https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0007-pod-ready++.md

dbyron0:
Well, from my reading of the code [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/prober/prober_manager.go#L156) and [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/prober/worker.go), readiness probes happen during the lifetime of the pod.

miekg:
We can still use it, but only at the start. I.e. a plugins emit an event
(or something else) and report ready, but only on startup and reload.

But this makes startup special and now it can fail. It also introduces
dependencies on other things in your stack being up when you restart
coredns - another thing you don't want.

Thanks for combing through the code.

Right now, I still consider health, as implemented by coredns and the
kubernetes plugin, as a loaded gun that can bring down all pods during
network disturbance. Moving this to readiness seems not to solve anything.

johnbelamaric:
It's not clear to me why you think this can bring everything down. I
suppose if the k8s plugin is saying not ready, it goes out of the load
balancer and so external queries don't work either. So it expands from
"can't answer internal queries" to "can't answer any queries". Bad, but not
much worse than "can't answer internal queries".

What network hiccup are you concerned about? Readiness checks themselves
are done on the node and so network is not an issue. Access to the API
server? Because if we lose that, we lose all internal cluster resolution
anyway - or at least it will go stale. Also, I don't think the readiness
check will fail just because the API server connection is lost. I believe
we only fail if we have *never synced*. Once we have synced once, then we
are always ready. So this is only a startup condition.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/health..." ]
>It's not clear to me why you think this can bring everything down. I
>suppose if the k8s plugin is saying not ready, it goes out of the load
>balancer and so external queries don't work either. So it expands from
>"can't answer internal queries" to "can't answer any queries". Bad, but not
>much worse than "can't answer internal queries".

No, it's probably a related event, so *all* your coredns instances are now dead; 
even though many could survive at least 30 seconds by utalizing the cache.

>What network hiccup are you concerned about? Readiness checks themselves
>are done on the node and so network is not an issue. Access to the API

ALL of them, anything that depends on the network to report health is inherently 
broken.

>we only fail if we have *never synced*. Once we have synced once, then we
>are always ready. So this is only a startup condition.

readyness makes some sense: filed #2611.

Issue #2548, opened 2019-02-09
Remove watch functionally


miekg:
See discussion in #2513

It's complex (and large), causes maintenance issue and is not used.

stp-ip:
We definitely need more data on how watch functionality might impact performance, but for this to be reasonable performance GRPC itself needs to be improved performance wise.

miekg:
This pr is not about performance, but cutting some code out.

We haven't got current perf figures for gRPC.

stp-ip:
Was adding the context that this currently is not a good option on production as far as our tests and talk prep went and this is mostly due to gRPC. Don't have a specific argument on removing the code or not.

miekg:
Ok and thanks for that.

Code can be brought back once we have a coherent plan to support and use
this. (Will still be fiddly, because there is little chance this will apply
cleanly by then)

PR #2549, opened 2019-02-09
Remove grpc watch functionality

miekg:
This was added, but didn't see any use. For a large, complex chunk of
code we should have some users of it.

Remove all watch functionally from plugins, servers and packages.

Fixes: #2548


miekg:
And tests failing, because its unclear where wanted watches should be kept

miekg:
oh, end we can't compile the protobuf anymore, something something package too old (had to comment that out to get things to compile)

chrisohaver (commented):
lgtm fwiw

chrisohaver:
Actually - i think i missed something in this change.  
Looks like updateModified is no longer called to update the modified time when objects change ... this affects zone transfers, since it is used as the serial.

Issue #1407, opened 2018-01-18
plugin/secondary: add metrics


miekg:
Metrics about zone transfers (and especially failures) should be added:
* failing to refresh
* retry active
* ... ?

SuperQ:
The ideas that come to mind for me are:

* gauge: last updated timestamp
* gauge: last updated serial number
* gauge: last update success boolean
* counter: refresh total
* counter: refresh time total
* counter: refresh errors total

calebice:
@fturib if this is something that is still relevant (hasn't been updated since April) I can add some of these listed metrics into the secondary plugin.

SuperQ:
Yes, I sadly haven't had the time to work on it myself. These metrics would be great.

fturib:
@miekg , @SuperQ : could you advice here on what you would like to implement. @calebice is willing to help on CoreDNS project, and is looking for first issue to fix ...

calebice:
I figured this would be a good beginning issue as I've got prior experience working with Prometheus metrics and collection!

SuperQ:
The metrics I listed above still seem like a good first pass. I'm happy to review any naming convention and related monitoring issues.

To add some literal metric names:

Given these prefixes:
```
Namespace: plugin.Namespace,
Subsystem: subsystem,
```

Gauges:
* `last_updated_seconds`
* `last_updated_serial`
* `last_update_success`

Counters:
* `refreshes_total`
* `refreshes_seconds_total`
* `refreshes_failed_total`

Each of these metrics should have a `zone` label.

miekg:
sgtm, happy review any pr

jiachengxu:
Hi @calebice , are you currently working on this issue? Can I take it?

calebice:
Hi @jiachengxu , 

Yes, I am working on this issue presently!

palash25:
~~Hi, so if I have understood this correctly solving this issue would require implementing the aforementioned metric types (just like prometheus does  https://github.com/prometheus/client_golang /blob/master/prometheus/gauge.go ) and then using them in the plugin or are theyalready implemented somewhere and that code could be reused here or can we just import and use the metrics type from the prometheus clients here? cc @miekg @SuperQ~~ 


Nevermind I found it.

miekg:
any work-in-progress or ideas we can look at?

palash25:
I just found how some of the metrics are used in the codebase, will try to make a WIP PR in few days if I can.

palash25:
[...]

PR #2550, opened 2019-02-09
plugin/secondary: Add metrics

palash25:
#1407

### DIFF
--- a/plugin/file/metrics.go
+++ b/plugin/file/metrics.go
@@ -1,0 +1,56 @@
+package file
+
+import (
+	"github.com/coredns/coredns/plugin"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const subsystem = "secondary"
+
+// Metrics for the secondary plugin
+var (
+	UpdatedSeconds = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: subsystem,
+		Name:      "update_duration_seconds",
+		Buckets:   plugin.TimeBuckets,
+		Help:      "Histogram of the duration of an update",
+	}, []string{"zone"})
+
+	UpdatedSerial = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: subsystem,
+		Name:      "serial",
+		Help:      "Gauge of the last updated serial",
+	}, []string{"zone"})
+
+	UpdatedSuccess = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: subsystem,
+		Name:      "last_updated_seconds",
+		Help:      "Gauge of the last updated timestamp",
+	}, []string{"zone"})
+
+	RefreshCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: subsystem,
+		Name:      "refreshes_total",
+		Help:      "Counter of the number of refreshes",
+	}, []string{"zone"})
+
+	RefreshSeconds = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: subsystem,
+		Name:      "refresh_seconds",
+		Buckets:   plugin.TimeBuckets,
+		Help:      "Histogram of the duration of a refresh",
+	}, []string{"zone"})
+
+	FailedRefreshCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: subsystem,
+		Name:      "refreshes_failed_total",
+		Help:      "Counter of the number of failed refresh attempts",
+	}, []string{"zone"})
+)
--- a/plugin/file/secondary.go
+++ b/plugin/file/secondary.go
@@ -123,6 +123,9 @@ Restart:
 	retryTicker := time.NewTicker(retry)
 	expireTicker := time.NewTicker(expire)
 
+	start := time.Now()
+	UpdatedSeconds.WithLabelValues(z.origin).Observe(time.Since(start).Seconds())
+
 	for {
 		select {
 		case <-expireTicker.C:
@@ -145,10 +148,12 @@ Restart:
 			}
 
 			if ok {
+				UpdatedSerial.WithLabelValues(z.origin).Set(float64(z.Apex.SOA.Serial))
 				if err := z.TransferIn(); err != nil {
 					// transfer failed, leave retryActive true
 					break
 				}
+				UpdatedSuccess.WithLabelValues(z.origin).Set(float64(time.Now().Unix()))
 			}
 
 			// no errors, stop timers and restart
@@ -160,23 +165,32 @@ Restart:
 
 		case <-refreshTicker.C:
 
+			start = time.Now()
+			RefreshSeconds.WithLabelValues(z.origin).Observe(time.Since(start).Seconds())
+
 			time.Sleep(jitter(5000)) // 5s randomize
 
 			ok, err := z.shouldTransfer()
 			if err != nil {
+				FailedRefreshCount.WithLabelValues(z.origin).Inc()
 				log.Warningf("Failed refresh check %s", err)
 				retryActive = true
 				continue
 			}
 
 			if ok {
+				UpdatedSerial.WithLabelValues(z.origin).Set(float64(z.Apex.SOA.Serial))
 				if err := z.TransferIn(); err != nil {
 					// transfer failed
+
+					FailedRefreshCount.WithLabelValues(z.origin).Inc()
 					retryActive = true
 					break
 				}
+				UpdatedSuccess.WithLabelValues(z.origin).Set(float64(time.Now().Unix()))
 			}
 
+			RefreshCount.WithLabelValues(z.origin).Inc()
 			// no errors, stop timers and restart
 			retryActive = false
 			refreshTicker.Stop()
--- a/plugin/secondary/README.md
+++ b/plugin/secondary/README.md
@@ -34,6 +34,19 @@ When a zone is due to be refreshed (Refresh timer fires) a random jitter of 5 se
 applied, before fetching. In the case of retry this will be 2 seconds. If there are any errors
 during the transfer the transfer fails; this will be logged.
 
+## Metrics
+
+If monitoring is enabled (via the *prometheus* directive) then the following metric is exported:
+
+* `coredns_secondary_update_duration_seconds{zone}` - histogram of the duration of an update.
+* `coredns_secondary_serial{zone}` - gauge of last updated serial number.
+* `coredns_secondary_last_updated_seconds{zone}` - gauge of successful updates.
+* `coredns_secondary_refreshes_total{zone}` - number of successful refreshes while updating.
+* `coredns_secondary_refreshes_failed_total{zone}` - number of failed refreshes while updating.
+* `coredns_secondary_refresh_seconds{zone}` - duration of the last refresh.
+
+Each metric has a label `zone` which is the zonename used for the request/response.
+
 ## Examples
 
 Transfer `example.org` from 10.0.1.1, and if that fails try 10.1.2.1.
--- a/plugin/secondary/setup.go
+++ b/plugin/secondary/setup.go
@@ -4,6 +4,7 @@ import (
 	"github.com/coredns/coredns/core/dnsserver"
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/file"
+	"github.com/coredns/coredns/plugin/metrics"
 	"github.com/coredns/coredns/plugin/pkg/parse"
 	"github.com/coredns/coredns/plugin/pkg/upstream"
 
@@ -17,6 +18,12 @@ func setup(c *caddy.Controller) error {
 	if err != nil {
 		return plugin.Error("secondary", err)
 	}
+
+	c.OnStartup(func() error {
+		metrics.MustRegister(c, file.UpdatedSeconds, file.UpdatedSerial, file.UpdatedSuccess,
+			file.RefreshCount, file.RefreshSeconds, file.FailedRefreshCount)
+		return nil
+	})
 
 	// Add startup functions to retrieve the zone and keep it up to date.
 	for _, n := range zones.Names {


miekg (commented):
Yes,this make sense.  Couple of point though:

* can you pull the metrics (and associated code) in a new file called metrics.go
* explain what are important metrics? Think the failures are very important and last_updated once
* explain the metrics in the readme
* add a server label in the metrics (see forward/metrics.go for an example)

Thanks!

palash25:
thanks for the quick review @miekg 

>  can you pull the metrics (and associated code) in a new file called metrics.go

by that, I am assuming that you mean `plugin/file/metrics.go` and not `plugin/secondary/metrics.go` (because the latter will lead to a cyclic import)

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [WIP] plugin:..." ]
>thanks for the quick review @miekg
>
>>  can you pull the metrics (and associated code) in a new file called metrics.go
>
>by that, I am assuming that you mean `plugin/file/metrics.go` and not `plugin/secondary/metrics.go` (because the latter will lead to a cyclic import)

yep, might make sense to put all metrics in there (file + secondary)

palash25:
@miekg I had a few more questions

would it make sense to make these two `last_updated_seconds`  `refreshes_seconds_total` histogram vectors instead of counter and gauges (as mentioned in the issue) since the other metrics in the codebase that are related to duration are also histogram vectors ?

how to I find an input for the server label argument in secondary plugin (the other plugins extract the server label from either a `ResponseWriter` struct or the `metrics.WithServer()` function using `context` but I don't see any of these being used in secondary)?


Also what is considered a successful update? Is it when the` Update` function finishes executing or everytime `goto Restart` in the `Update` function is executed?


I am also assuming that everytime the shouldTransfer method returns true the UpdatedSerial metric is supposed to be incremented and vice versa, hope that it is correct.

palash25:
ping @miekg

miekg:
> would it make sense to make these two `last_updated_seconds` `refreshes_seconds_total` histogram vectors instead of counter and gauges

yes, I would follow what we do in other plugins.

> how to I find an input for the server label argument in secondary plugin (the other plugins extract the server label from either a `ResponseWriter` struct or the `metrics.WithServer()` function using `context` but I don't see any of these being used in secondary)?

`WithServer` should be used here.

> Also what is considered a successful update? Is it when the` Update` function finishes executing or everytime `goto Restart` in the `Update` function is executed?

A successful update would be that we've incremented the local SOA from the remote server.

> I am also assuming that everytime the shouldTransfer method returns true the UpdatedSerial metric is supposed to be incremented and vice versa, hope that it is correct.

Depends, what does this signal to the user looking at the metric?

palash25:
@miekg updated the PR, still a little unsure about the difference betwee `UpdateSerial` and `UpdatedSuccess` metrics

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin: Add m..." ]
>@miekg updated the PR, still a little unsure about the difference betwee `UpdateSerial` and `UpdatedSuccess` metrics

pretty busy right now, but I'll try to take a look soon.

plugin/file/metrics.go
```diff
@@ -0,0 +1,53 @@
+package file
+
+import (
+	"github.com/coredns/coredns/plugin"
+	"github.com/prometheus/client_golang/prometheus"
```
miekg, line 6: these should be separated by a newline

plugin/file/secondary.go
```diff
@@ -112,6 +114,8 @@ func (z *Zone) Update() error {
 	}
 	retryActive := false
 
+	server := metrics.WithServer(context.Background())
```
miekg: this can't work, and can't give you the server. Which is a problem, because I don't think we have another way to get to it...

plugin/secondary/README.md
```diff
@@ -38,6 +38,19 @@ When a zone is due to be refreshed (Refresh timer fires) a random jitter of 5 se
 applied, before fetching. In the case of retry this will be 2 seconds. If there are any errors
 during the transfer the transfer fails; this will be logged.
 
+## Metrics
+
+If monitoring is enabled (via the *prometheus* directive) then the following metric is exported:
+
+* `coredns_secondary_last_updated_seconds{server, zone}` - duration of the last update.
+* `coredns_secondary_last_updated_serial{server, zone}` - gauge of serial update attempts.
```
miekg: I don't understand the gauges here, what are we trying to measure and why aren't we doing some like process_start_time_seconds but then for these updates?

plugin/file/secondary.go
palash25: So should I drop the server label from these metrics? because I tried finding other ways for it in the codebase but couldnt

plugin/secondary/README.md
palash25: even I don't really because I was trying to make sense of these guages from their names that were mentioned in the issue thread by @SuperQ here https://github.com/coredns/coredns/issues/1407#issuecomment-428276568 but not a lot of details were provided about them which made it hard to make sense of what they actually mean.

I could drop the guages for now and just submit this PR with the metrics that you find acceptable and if someone would like to open a discussion about guages again in the future and has a clearer description then I can implement them in another PR. Sounds good?

palash25: > why aren't we doing some like process_start_time_seconds but then for these updates?

this PR does have `coredns_secondary_last_updated_seconds` which records the duration of the last update. I don't see how `process_start_time_seconds` is different from that cc @miekg

SuperQ: last_updated_seconds is the timestamp of when the last update happened, not how long it took.

palash25: oh, thanks for clearing that up.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin: Add m..." ]
>> @@ -112,6 +114,8 @@ func (z *Zone) Update() error {
> 	}
> 	retryActive := false
>
>+	server := metrics.WithServer(context.Background())
>
>So should I drop the server label from these metrics? because I tried finding other ways for it in the codebase but couldnt

no you can't because this is def. tied to a server (in the future), except at 
this point you don't know it as this point.

I'm sorry, but I don't have a good solution for this right now. This requires 
some more thought and what is sensible here.

plugin/file/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+// Metrics for the secondary plugin
+var (
+	UpdatedSeconds = prometheus.NewHistogramVec(prometheus.HistogramOpts{
```
miekg, line 13: this file should be moved out of `file` and into `secondary`, the metrics variables are exported anyway, so they can still be accessed from *file*

plugin/file/secondary.go
```diff
@@ -170,11 +191,17 @@ Restart:
 			if ok {
 				if err := z.TransferIn(); err != nil {
 					// transfer failed
+					UpdatedSerial.WithLabelValues(z.origin).Dec()
+					UpdatedSuccess.WithLabelValues(z.origin).Dec()
```
miekg: these all miss the server label

miekg: Going further, I don't see a way to actually add the correct `server` label, so it should be removed from all of the metrics here, and just have the origin.

miekg (changes requested):

plugin/file/metrics.go
palash25: that would lead to an import cycle becuase then `file` will import `secondary` to be able to use the metrics and `secondary`  already imports `file`

I have discussed this with you before [here](https://github.com/coredns/coredns/pull/2550#issuecomment-462134982) and I think you had agreed to it if I understood it correctly @miekg

plugin/file/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+		Namespace: plugin.Namespace,
+		Subsystem: "secondary",
+		Name:      "last_updated_seconds",
```
dilyevsky: Name seems confusing (reads like a timestamp of last update). `last_update_dur_sec`?

plugin/file/secondary.go
```diff
@@ -138,15 +147,21 @@ Restart:
 
 			ok, err := z.shouldTransfer()
 			if err != nil {
+				UpdatedSerial.WithLabelValues(server, z.origin).Dec()
```
dilyevsky: Isn't this supposed to be gauge? It seems to be used as a counter and in lockstep with `UpdatedSuccess`. Is this intended?

miekg:
Ah yeah. True, I forgot. Proceed :)

plugin/file/secondary.go
palash25: I got confused initially about what this metric meant. I will be correcting this in my next push

palash25:
@miekg pushed some change, please review

palash25:
busy with some work, will get back to this next weekend and update the PR

palash25:
ping @miekg

palash25:
@miekg can you review this please?

plugin/file/metrics.go
SuperQ: There's no need to shorten names. It should be `update_duration_seconds`. The "last" implies it's more of a `happened at some time` timestamp metric like `process_start_time_seconds`.

plugin/file/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/coredns/coredns/plugin"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// Metrics for the secondary plugin
+var (
+	UpdatedSeconds = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "secondary",
```
SuperQ: Make this a `const` to avoid duplicate string typos.

plugin/file/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+		Name:      "refresh_seconds_total",
+		Buckets:   plugin.TimeBuckets,
+		Help:      "Histogram of the duration of a refresh",
+	}, []string{"zone"})
+
+	FailedRefreshCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "secondary",
+		Name:      "refresh_failed_total",
```
SuperQ:
```suggestion
		Name:      "refreshs_failed_total",
```

plugin/file/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+		Namespace: plugin.Namespace,
+		Subsystem: "secondary",
+		Name:      "refresh_total",
+		Help:      "Counter of the number of refreshes",
+	}, []string{"zone"})
+
+	RefreshSeconds = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "secondary",
+		Name:      "refresh_seconds_total",
```
SuperQ: Histograms don't need `_total`.

```suggestion
		Name:      "refresh_seconds",
```

plugin/file/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+		Namespace: plugin.Namespace,
+		Subsystem: "secondary",
+		Name:      "last_updated_serial",
+		Help:      "Gauge of the last updated serial",
+	}, []string{"zone"})
+
+	UpdatedSuccess = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "secondary",
+		Name:      "last_updated_success",
```
SuperQ:
```suggestion
		Name:      "last_updated_seconds",
```

plugin/file/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+		Help:      "Gauge of the last updated timestamp",
+	}, []string{"zone"})
+
+	RefreshCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "secondary",
+		Name:      "refresh_total",
```
SuperQ:
```suggestion
		Name:      "refreshes_total",
```

plugin/file/metrics.go
SuperQ:
```suggestion
		Name:      "update_duration_seconds",
```

plugin/file/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+		Subsystem: "secondary",
+		Name:      "last_updated_seconds",
+		Buckets:   plugin.TimeBuckets,
+		Help:      "Histogram of the duration of an update",
+	}, []string{"zone"})
+
+	UpdatedSerial = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "secondary",
+		Name:      "last_updated_serial",
```
SuperQ:
```suggestion
		Name:      "serial",
```

palash25:
@SuperQ changes made PTAL.

plugin/file/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+var (
+	UpdatedSeconds = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: subsystem,
+		Name:      "update_duration_seconds",
+		Buckets:   plugin.TimeBuckets,
+		Help:      "Histogram of the duration of an update",
+	}, []string{"zone"})
+
+	UpdatedSerial = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
miekg, line 21: I've recently come to the realisation that gauges are bad, like really bad for monitoring. And it's much better to make this a counter and use epoch timestamp to convey this information. Just like process_start_time_second which just stamps things with the unix epoch when it happens.

plugin/file/secondary.go
```diff
@@ -160,23 +165,32 @@ Restart:
 
 		case <-refreshTicker.C:
 
+			start = time.Now()
+			RefreshSeconds.WithLabelValues(z.origin).Observe(time.Since(start).Seconds())
```
miekg, line 169: this just sets it to 0? Or a value close to it?

plugin/file/secondary.go
```diff
@@ -160,23 +165,32 @@ Restart:
...
 					// transfer failed
+
+					FailedRefreshCount.WithLabelValues(z.origin).Inc()
 					retryActive = true
 					break
 				}
+				UpdatedSuccess.WithLabelValues(z.origin).Set(float64(time.Now().Unix()))
 			}
 
+			RefreshCount.WithLabelValues(z.origin).Inc()
```
miekg, line 193: Isn't refreshcount just updateSuccess+failedRefreshcount? If so, it can be removed.

plugin/file/metrics.go
SuperQ: +1, in general gauges can be bad for monitoring.

However, the zone serial is an integer, and could be useful for comparison. For example, if you have multiple secondaries serving a zone, you could compare something like this:

`coredns_secondary_serial != ignoring (instance) max without (instance) (coredns_secondary_serial)`

This would let you know if one of several secondaries isn't getting updated.

plugin/file/secondary.go
SuperQ: Yes, this observation doesn't make any sense, since there's nothing between the `Now()` and the observation.

plugin/file/secondary.go
SuperQ: No, `updateSuccess` is the timestamp of the last valid update. This looks correct.

miekg:
closing as this pr isn't going anywhere

PR #2559, opened 2019-02-15
Change http to https for security links

trungnvfet:
Signed-off-by: Nguyen Van Trung <trungnv@vn.fujitsu.com>

For security, should be change into https links


yongtang (approved):
LGTM

Issue #2570, opened 2019-02-17
plugin/cache: different between DO and no DO?


miekg:
Currently the cache generates a different key for requests that have the DO (DNSSEC OK) bit set. This is not *really* needed, but handy. But of course this increases the memory use. 

As an alternative we could just cache (no matter what) and filter out DNSSEC record for clients that don't need need them.

This also involved plugin/forward a bit, because it matters what we ask upstream there, also see #2569

Issue #2569, opened 2019-02-17
plugin/forward: do all queries with 4k buffer and handle truncation


miekg:
The *forward* plugin would be simpler and more robust if all queries are done using a 4k buffer and doing the TCP retry (on truncation) ourselves.

1. 4k buffer: regardless of the input buffer we use a 4k buffer; we already have code that makes responses fit in the client's buffer anyway.
2. TCP retry: this saves client trip *and* we can then cache the result - which should happen anyway.

These changes also impact caching; as every response from forward can now be cached, unlike the truncation response we would sent down - as we do now.

This will probably impact the `force_tcp` and `prefer_udp` option which may be an symptom that our current design doesn't cut it.

We may need to special case clients connecting via tcp - but I'm not sure that's worth it.

If this is done, we can take a fresh look at it's performance as well.

miekg:
overly optimistic that this would solve anything, closing

PR #2572, opened 2019-02-17
plugin/forward: all queries use 4k buffer

miekg:
Use 4k buffer for all queries and deprecate force_tcp and prefer_udp.
We do check of the clients wants DNSSEC or not.

See #2569 #2570


miekg:
this is def. backwards incompatible and we need to announce, etc.

greenpau (commented):
@miekg , you are removing the `force_tcp`, `prefer_udp` from `README.md`. What do you think about adding a section in the file and saying that these options are no longer valid and explaining why?

miekg:
yeah, I'm wonder if this change is a good change. The thing is; it seems the *cache* plugin implies we do things differently here. this probably means the cache plugin is doing things wrong. And the thing it's doing wrong is caching entire replies; meaning we should just implement an RRset cache.

PR #2577, opened 2019-02-18
DoH: Fixing panic in case if there's no response

ameshkov:

There could be a situation when there's no response after ServeDNS call. With the current implementation, this leads to panic.

I've not filed a bug report, went to the pull request right away:)

Nothing

No

### DIFF
--- a/core/dnsserver/server_https.go
+++ b/core/dnsserver/server_https.go
@@ -119,6 +119,11 @@ func (s *ServerHTTPS) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// We should expect a packet to be returned that we can send to the client.
 	s.ServeDNS(context.Background(), dw, msg)
 
+	if dw.Msg == nil {
+		http.Error(w, "No response", http.StatusInternalServerError)
+		return
+	}
+
 	buf, _ := dw.Msg.Pack()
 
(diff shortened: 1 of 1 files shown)


ameshkov:
Well, I don't think Travis has failed due to my change, but I cannot restart that build to prove it:)

core/dnsserver/server_https.go
```diff
@@ -119,6 +119,11 @@ func (s *ServerHTTPS) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// We should expect a packet to be returned that we can send to the client.
 	s.ServeDNS(context.Background(), dw, msg)
 
+	if dw.Msg == nil {
+		http.Error(w, "No response", http.StatusInternalServerError)
```
miekg, line 126: is this the correct response according to the DoH spec?

ameshkov: Well, there are no specific requirements for DOH responses with non-successful HTTP status codes. 

The only relevant part of the draft is this:
> DoH clients need to use the same semantic processing of non-successful HTTP status codes as other HTTP clients.

https://tools.ietf.org/html/rfc8484#page-7

miekg:
Can you add a comment that links to that rfc and section? I.e. "See RFC
8484, Section ....."

ameshkov:
Sure

PR #2578, opened 2019-02-18
Disable Proxy Recovery

askurydzin:

This PR includes new flag (disable_proxy_recovery) that lets all proxies to go down and never go up thus enabling following things:

 1. Other monitoring/healthcheck plugins could take control over health-check and recovery mechanisms reusing proxy.Down functionality. 
2. This also allows alternate plugin to switch faster without random upstream allocation.

1. Alternate plugin slow-down
2. No control over random upstream allocation [Not an issue, however worth to mention].

New forward flag `disable_proxy_recovery` to be introduced.

No. By default `disable_proxy_recovery` is set to false.

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -48,6 +48,7 @@ forward FROM TO... {
     except IGNORED_NAMES...
     force_tcp
     prefer_udp
+    disable_proxy_recovery
     expire DURATION
     max_fails INTEGER
     tls CERT KEY CA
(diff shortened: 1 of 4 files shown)


miekg:
Thanks, but I'm confused on what your fixing here? What is the core the problem and what does *forward* do wrong here?  I also prefer something that doesn't add a new user visible knob(s).

askurydzin:
It's not an issue, but maybe it makes sense to allow proxies to go down and let other techniques to be employed (restart or something) without random allocation of the upstream and healthcheck kick-off. Also alternate plugin is slowed down by random upstream allocation, which is not good as for me.

If it seems bizarre for you, I'm OK with closing this PR.

miekg:
doesn't max_fails do the same?

PR #2580, opened 2019-02-19
Adding '/usr/bin/env bash'

truongnh1992:
This commit aims to add '/usr/bin/env bash' as a shebang line
to indicates scripts use bash shell for interpreting.

Signed-off-by: Nguyen Hai Truong <truongnh@vn.fujitsu.com>

### DIFF
--- a/.presubmit/context
+++ b/.presubmit/context
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 echo "** presubmit/$(basename $0)"
 
--- a/.presubmit/filename-hyphen
+++ b/.presubmit/filename-hyphen
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 echo "** presubmit/$(basename $0)"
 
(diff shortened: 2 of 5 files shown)


miekg:
where is /bin/bash not working?

truongnh1992:
> 
> 
> where is /bin/bash not working?

@miekg 
Running a command through `/usr/bin/env` has the benefit of looking for whatever the default version of the program is in your current environment.
```
#!/usr/bin/env bash   #lends you some flexibility on different systems
#!/usr/bin/bash   #gives you explicit control on a given system of what executable is called
```
Moreover, `#!/usr/bin/env bash` makes the shell search for the first match of `bash` in the $PATH environment variable. It can be useful if you aren't aware of the absolute path or don't want to search for it.

*Source:* https://stackoverflow.com/questions/16365130/what-is-the-difference-between-usr-bin-env-bash-and-usr-bin-bash?answertab=votes#tab-top

miekg:
Sure. But if this PR doesn't really fix a bug that you experience, I don't
see why it should be merged.

truongnh1992:
@miekg As I known, almost shell scripts use the `shebang line`: #!/usr/bin/env bash
For example: Kubernetes project.

PR #2583, opened 2019-02-20
Setting authoritative flag to responses when served from cache.

uruddarraju:
This sets the authoritative flag on responses served from cache.

https://stackoverflow.com/questions/54778160/python-requests-library-not-resolving-dns-non-authoritative-dns-lookups. 

None.

No.


uruddarraju:
@miekg Please let me know if you have other thoughts on this and any specific reasons around why you don't want this flag set. I just verified an nscd installation, and it returns `aa` for responses served from its cache as well.

miekg:
This is wrong for cache responses. Which RFC says this should be done?

uruddarraju:
Thanks for responding on this one @miekg

PR #2588, opened 2019-02-20
Adding support for explicit no-cache policy for positive/negative dns…

### LINKED ISSUE #2586: Disable neg-cache for few domains
Is there a way to disable neg-caching for a particular domain similar to what dnsmasq provides with the no-neg-cache flag ?

I ask this to do a possible workaround to this issue I am facing [here](https://stackoverflow.com/questions/54778160/python-requests-library-not-resolving-dns-non-authoritative-dns-lookups), but also wanted to understand the reasoning behind not having a flag to disable neg-cache.

uruddarraju:
Please see https://github.com/coredns/coredns/issues/2586 for more context.
TLDR; This PR provides a way to disable one of the caches (positive/negative) and not the other. This enables usders to use coredns for positive caching with a ttl, while opting out of negative caching similar to the `no-neg-cache` support in dnsmasq

### 2. Which documentation changes (if any) need to be made?
cache.md should be updated, but will update this PR with the doc changes once I get some feedback.

### 3. Does this introduce a backward incompatible change or deprecation?
No

Also this enables us to have a cache policy like the following:
```
        cache {
          success 5000 5
          denial 0
        }
```
that would enable success cache but disable denial cache. 

fixes #2586

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -203,10 +203,18 @@ func (w *ResponseWriter) set(m *dns.Msg, key uint64, mt response.Type, duration 
 	// and key is valid
 	switch mt {
 	case response.NoError, response.Delegation:
+		// do not cache if cache disabled
+		if w.pcap < 0 {
+			return
+		}
 		i := newItem(m, w.now(), duration)
 		w.pcache.Add(key, i)
 
 	case response.NameError, response.NoData:
+		// do not cache if cache disabled
+		if w.ncap < 0 {
+			return
+		}
 		i := newItem(m, w.now(), duration)
 		w.ncache.Add(key, i)
 


uruddarraju:
cc @chrisohaver

stp-ip:
This seems to change the behavior even so it might be the correct behavior from a config perspective. 
`denial 0` did reduce the size to 1024 before and now would not cache at all. This would be a breaking change, but it might be ok, if documented correctly and announced in the release notes as it seems to be more in line with the actual intention.

uruddarraju:
@stp-ip Thanks for the comments. So would a doc change suffice for me to get this merged ? :)

Or, are there other things you'd need me to address before this gets merged ?

stp-ip:
From my side a docs change and a prepared release note to announce the breaking change would be enough as this seems to be a simple change, but I want another approval to get this merged as I might have missed some context. Just wanted to get feedback so that the second approval might already merge.
/lgtm

uruddarraju:
Thanks @stp-ip

miekg:
Setting size to zero seems natural to me. I don't see a backwards incompatible change.

But this is checking in the wrong place, nothing should be put in the cache in the first place.

uruddarraju:
@miekg yeah I initially thought of implementing this at Cache struct that is used commonly across. Or do you think it would be better for us to have the condition to check for the pcap/ncap here: https://github.com/coredns/coredns/blob/master/plugin/cache/cache.go#L201

uruddarraju:
Or would you think it is more appropriate to change the behavior of cache.Cache https://github.com/coredns/coredns/blob/master/plugin/pkg/cache/cache.go#L19 instead ? This would probably be a bigger change.

uruddarraju:
@miekg can you please take an other look ? Moved the logic to not put stuff into cache when the capacity is set to 0.

plugin/cache/cache.go
```diff
@@ -203,12 +203,16 @@ func (w *ResponseWriter) set(m *dns.Msg, key uint64, mt response.Type, duration
 	// and key is valid
 	switch mt {
 	case response.NoError, response.Delegation:
-		i := newItem(m, w.now(), duration)
-		w.pcache.Add(key, i)
+		if w.pcap > 0 {
```
chrisohaver: I recall Miek prefers minimizing indentation whenever possible ... e.g. ...

```
if w.pcap <= 0 {
    return
}
i := newItem(m, w.now(), duration)
w.pcache.Add(key, i)
```

uruddarraju:
Thanks @chrisohaver for the feedback. Addressed the indentation comment.

chrisohaver (approved):

chrisohaver (changes requested):
Looks OK as is codewise IMO: To be super tidy, we could also avoid initializing the cache if it's not going to be used.  i.e. In `cache.New(size)`, returning `nil` if the size is zero, and possible other tweaks elsewhere to avoid nil reference.

Will approve when docs are updated and unit test is added.

miekg:
lgtm, but please fix what Chris mentioned here.

chrisohaver:
ping @uruddarraju

uruddarraju:
on it, sorry for the miss.

miekg:
@uruddarraju can you fix the conflict? Thanks!

m4r10k:
@chrisohaver Thank you for picking this one up! We are also facing the same issue like mentioned [here on stackoverflow]( https://stackoverflow.com/questions/54778160/python-requests-library-not-resolving-non-authoritative-dns-lookups). In our case it seems like that with quite old Linux versions (Debian, CentOS,...) there might be a bug somewhere either in Python or in some case in C libraries?

Output from a failing rsync operation:
~~~
rsync: getaddrinfo: rsync.app.strabag.com 873: No address associated with hostname
rsync error: error in socket IO (code 10) at clientserver.c(122) [sender=3.0.9]
~~~

If the cache is disabled (removed from config), everything works neat. I think it would be helpful to be able to disable the **denied** cache to avoid this problems.

If I get some time today, we will try this PR in our environment, as we can test it easily - we have an reproducible case for this.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Adding suppor..." ]
>@chrisohaver Thank you for picking this one up! We are also facing the same issue like mentioned [here on stackoverflow]( https://stackoverflow.com/questions/54778160/python-requests-library-not-resolving-non-authoritative-dns-lookups). In our case it seems like that with quite old Linux versions (Debian, CentOS,...) there might be a bug somewhere either in Python or in some case in C libraries?
>
>Output from a failing rsync operation:
>~~~
>rsync: getaddrinfo: rsync.app.strabag.com 873: No address associated with hostname
>rsync error: error in socket IO (code 10) at clientserver.c(122) [sender=3.0.9]
>~~~
>
>If the cache is disabled (removed from config), everything works neat. I think it would be helpful to be able to disable the **denied** cache to avoid this problems.
>
>If I get some time today, we will try this PR in our environment, as we can test it easily - we have an reproducible case for this.

What does a tcpdump return here? I.e. would be good to get to the bottom. 
Disable the cache seems looks like a (good) mitigation, but doesn't sit well 
with me.

m4r10k:
> What does a tcpdump return here? I.e. would be good to get to the bottom. Disable the cache seems looks like a (good) mitigation, but doesn't sit well with me.

We are currently locked down in a meeting :smiley:, but I think we can provide a tcp dump later. I will send the dump via email to you for privacy reasons.

But are you sure that we should dig into it, as it seems like that only pretty old systems are affected?

miekg:
That works.

Yes the craziness of the DNS protocol requires that servers do their best
to facilitate stub clients. The ruby DNS client is also really bad, so this
may benefit those folks as well.

Would be good to include the initial answer , the cached answer and maybe
the answer without caching in the dump to compare

m4r10k:
OK, then we will try to get the dumps, write some description and send it over to you. This might take some time (maybe until tomorrow) ... :smiley: I will give feedback here, to let the others know about the progress!

m4r10k:
@miekg First tcpdump files should be in your inbox :smile:

miekg:
Thanks. We had a private discussion and I looked at the dump. It seems Ruby's "DNS" client disregards responses that don't the the AA (authoritative) bit set. I may be mistaken by I believe if an answer comes from a cache you shouldn't set this bit. OTOH I don't care that much, it's 1 bit and DNS is badly defined anyway; IOW: we can just set the bit even if we respond with an answer from the cache.

This would be a 1-line change - modulo some tests prolly.

stefanmb:
So for what it's worth I ran into a nasty production issue with Ubuntu 14.0.4 LTS (eglibc 2.19-0ubuntu6.14) where CoreDNS's lack of authoritative bit on cached responses caused failures for all apps using `getaddrinfo`, including Ruby and a MySQL client. 

I couldn't reproduce the behaviour after Ubuntu 14.10 (glibc 2.19-10ubuntu2.3). I assume it's a glibc bug, but I couldn't track down the exact lines/patch involved.

I ended up working around it by writing a CoreDNS "authoritative" plugin that simply sets the authoritative bit on all responses from a domain whitelist (e.g. *.cluster.local since I have autopath enabled), and then moving the legacy workloads away from Ubuntu 14.04.

m4r10k:
@miekg it seams like that what @stefanmb wrote confirms that there is some strange behavior of some older or buggy implementations which are caused by this "non/authoritative" flag switching in the cache plugin case. Can you point me the line, where the code can be changed to enforce the always authoritative response? I would hack it in tomorrow to prove this in our environment.

Afterwards it might be an option to make it configurable within the cache plugin maybe or if this flag doesn't really border anything of the DNS protocol you can pin it like you mentioned before.

Question: I don't want to hijack this issue for this - should I open a separate to not mix up the things here? Especially because  https://github.com/kubernetes/kubeadm/issues/1512 seems like to have the same root cause?

stefanmb:
@m4r10k All you need to do is clobber `Authoritative = true`: https://github.com/coredns/coredns/blob/master/plugin/cache/item.go#L59

Which OS and glibc version are you running?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Adding suppor..." ]
>@miekg it seams like that what @stefanmb wrote confirms that there is some strange behavior of some older or buggy implementations which are caused by this "non/authoritative" flag switching in the cache plugin case. Can you point me the line, where the code can be changed to enforce the always authoritative response? I would hack it in tomorrow to prove this in our environment.

https://github.com/coredns/coredns/blob/master/plugin/cache/item.go#L59

>Afterwards it might be an option to make it configurable within the cache plugin maybe or if this flag doesn't really border anything of the DNS protocol you can pin it like you mentioned before.

It's not super important to make this configurable. If things break we should 
just set it to true.

>Question: I don't want to hijack this issue for this - should I open a separate to not mix up the things here? Especially because  https://github.com/kubernetes/kubeadm/issues/1512 seems like to have the same root cause?

Yes, that makes sense. Thank you.

/Miek

--
Miek Gieben

m4r10k:
@stefanmb Thanks! Some legacies are always ghosting around like in your example and as always...  
@miekg Thanks too! 
:smile: - I will open an issue for this tomorrow.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Adding suppor..." ]
>@stefanmb Thanks! Some legacies are always ghosting around like in your example and as always...
>@miekg Thanks too!
>:smile: - I will open an issue for this tomorrow.

https://github.com/coredns/coredns/pull/2885 to remove (and save some memory in 
the process)

/Miek

--
Miek Gieben

m4r10k:
@miekg Wow, thanks!

miekg:
Fixed with 1.5.1

PR #2591, opened 2019-02-21
Replacing 'HTTP' by 'HTTPS' for securing links

longkb:
Currently, there are some links that we access with **HTTP**, it is
redirected to **HTTPS** automatically. So this commit aims to
replace **http://...** by **https://...** for security.

Co-Authored-By: Nguyen Phuong An <AnNP@vn.fujitsu.com>
Signed-off-by: Kim Bao Long <longkb@vn.fujitsu.com>

For securing HTTP link

### DIFF
--- a/plugin/file/tree/less.go
+++ b/plugin/file/tree/less.go
@@ -10,7 +10,7 @@ import (
 // >0 when a is larger than b.
 // The function orders names in DNSSEC canonical order: RFC 4034s section-6.1
 //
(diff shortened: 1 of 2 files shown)


yongtang:
The change in `vendor/github.com/apache/thrift/lib/go/thrift/simple_json_protocol.go` is actually in vendor directory which should not be modified. Created PR #2594 for the fix.

PR #2595, opened 2019-02-22
Adding detail of file ISSUE_TEMPLATE

truongnh1992:
Describe more detail in ISSUE_TEMPLATE:
	- Operating sytem, kernel version
	- Corefile?
	- Problem explanation, logs

Co-Authored-By: Dao Cong Tien tiendc@vn.fujitsu.com
Signed-off-by: Nguyen Hai Truong <truongnh@vn.fujitsu.com>

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -6,3 +6,13 @@ When reporting bugs and requesting assistance, please include:
 
 Thanks for using CoreDNS!
 -->
+
+### What version of coredns are you using?
+
+### What's your Corefile?
+
+### Environment information:
+- Operating system (output of `cat /etc/os-release`):
+- Kernel version (output of `uname -r`):
+
+### Your problem explanation, logs?


truongnh1992:
@pmoroney Could you help me to review this PR, please.

chrisohaver (changes requested):
Most of this seems redundant with the questions already present in the template.

pmoroney:
Also those instructions, `cat /etc/os-release`, and `uname -r`, are not OS agnostic. If someone used an OS that didn't use that file, or a non-linux OS, those commands might be confusing.

miekg:
Better addition is adding something like:
* Does your change increase resource usage
* Does your change decrease throughput
* Are there repeatable benchmarks

truongnh1992:
> 
> 
> Better addition is adding something like:
> 
>     * Does your change increase resource usage
> 
>     * Does your change decrease throughput
> 
>     * Are there repeatable benchmarks

Thank you. I see. I've just updated it.

chrisohaver:
```
* Does your change increase resource usage
* Does your change decrease throughput
```

... could probably boil those 2 down to one question.

What precisely do you mean by `Are there repeatable benchmarks`?  Are you asking for benchmark tests to be included?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Adding detail..." ]
>```
>* Does your change increase resource usage
>* Does your change decrease throughput
>```
>
>... could probably boil those 2 down to one question.
>
>What precisely do you mean by `Are there repeatable benchmarks`?  Are you asking for benchmark tests to be included?

yes, are a shell-script; how can you validate the claims otherwise. This is why 
I also want some kind of framework where these things can easily plugged in, 
benchmark tests are one such thing.

Issue #2600, opened 2019-02-22
plugin/template does not resolve CNAME as advertised


thomas-mangin:
Sorry to have to open a bug report, thankfully nothing is on fire :-)

Using master and the following (simplified) configurations.

```
. {
        template IN ANY . {
                match ^(www[.])?google[.](.*)$
                answer "{{ .Name }} 60 IN CNAME forcesafesearch.google.com."
                fallthrough
        }

        forward . IP:53 OTHER:53 {
                policy round_robin
        }

        # no tuning was done on these cache numbers, including prefetchs
        cache {
                success 10000
                denial 5000
                prefetch 250
        }
}
```

```
. {
        template IN ANY . {
                match ^(www[.])?google[.](.*)$
                answer "{{ .Name }} 60 IN CNAME forcesafesearch.google.com."
                upstream IP
                fallthrough
        }

        forward . IP:53 OTHER:53 {
                policy round_robin
        }

        # no tuning was done on these cache numbers, including prefetchs
        cache {
                success 10000
                denial 5000
                prefetch 250
        }
}
```

The first one does return the CNAME, when the second one return the expected CNAME *and* the A record for `forcesafesearch.google.com`. 

The documentation reports that `upstream` ```defines the upstream resolvers used for resolving CNAMEs. CoreDNS will resolve CNAMEs against itself.```.

I understood the sentence as saying ```upstream defines the upstream resolvers used for resolving CNAMEs.```  `Otherwise` ```CoreDNS will resolve CNAMEs against itself.``` so I would not have expected a difference in behaviour and always see the A record added.

Only returning CNAME does cause issues with clients, therefore the plugin should surely be updated to make sure to always add the A record even when no `upstream` option is used.

chrisohaver:
CoreDNS version?

chrisohaver:
The docs are unclear/wrong.  Without `upstream` present, template will not lookup CNAME targets.
This is as designed.  The intended behavior is:

no `upstream` present: no A records appended
`upstream` by itself (no ips): A record appended, resolved against self. 
`upstream 1.2.3.4`: A record appended, resolved against 1.2.3.4. 

However, we plan to change this behavior soon, removing the `upstream` option entirely, and always to resolve CNAMEs against "itself."

thomas-mangin:
@chrisohaver thank you very much for looking into this and providing the clarification patch.

PR #2603, opened 2019-02-23
use local server if no upstream in template

zaheerm:

When no upstream resolvers are defined in a template, make CoreDNS resolve CNAME targets against itself..

#2600

None

No


miekg:
this code is going away (and already not used - hopefully). Every `upstream` uses a proxy (if available) and depending on that config.

fturib:
@zaheerm : 
Thank you for proposing a PR. but the the fix for #2600 was rather a clarification of the documentation.
And it is now done here in #2613

PR #2604, opened 2019-02-24
plugin/kubernetes: create dns.Msg later

miekg:
There is no need to do this at the top of the handler, move this
down into the function.


yongtang (approved):
LGTM. This may potentially help a little I think.

PR #2602, opened 2019-02-23
refactor tests to remove duplicate code

darshanime:
Signed-off-by: darshanime <deathbullet@gmail.com>

### 1. Why is this pull request needed and what does it do?
It refactors `route_53` tests to remove duplicate code and improve readability.
### 2. Which issues (if any) are related?
NA

### 3. Which documentation changes (if any) need to be made?
None

### 4. Does this introduce a backward incompatible change or deprecation?
No

PR #2606, opened 2019-02-25
Print test case on UT failure

yuewko:
The change introduced by #2602 make it difficult, if not impossible, to identify the offending test case from the output should there be a UT failure. E.g.
```
$ go test
--- FAIL: TestSetupRoute53 (0.00s)
    setup_test.go:73: Unexpected errors: <nil>
FAIL
exit status 1
FAIL	github.com/coredns/coredns/plugin/route53	0.025s
```
Note that the line no. shown above is the line inside the test loop, not the test case itself.

The output of the same test failure with this PR looks like this:

```
go test
--- FAIL: TestSetupRoute53 (0.00s)
    --- FAIL: TestSetupRoute53/Test_case:_'route53_example.org:12345678_{_____upstream_}' (0.00s)
        setup_test.go:75: Unexpected errors: <nil>
FAIL
exit status 1
FAIL	github.com/coredns/coredns/plugin/route53	0.028s
```

None.

None.

No.

### DIFF
--- a/plugin/route53/setup_test.go
+++ b/plugin/route53/setup_test.go
@@ -1,6 +1,7 @@
 package route53
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/aws/aws-sdk-go/aws/credentials"
@@ -68,9 +69,11 @@ func TestSetupRoute53(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		c := caddy.NewTestController("dns", test.body)
-		if err := setup(c, f); (err == nil) == test.expectedError {
-			t.Errorf("Unexpected errors: %v", err)
-		}
+		t.Run(fmt.Sprintf("Test case: '%s'", test.body), func(t *testing.T) {
+			c := caddy.NewTestController("dns", test.body)
+			if err := setup(c, f); (err == nil) == test.expectedError {
+				t.Errorf("Unexpected errors: %v", err)
+			}
+		})
 	}
 }


plugin/route53/setup_test.go
```diff
@@ -68,9 +69,11 @@ func TestSetupRoute53(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		c := caddy.NewTestController("dns", test.body)
-		if err := setup(c, f); (err == nil) == test.expectedError {
-			t.Errorf("Unexpected errors: %v", err)
-		}
+		t.Run(fmt.Sprintf("Test case: '%s'", test.body), func(t *testing.T) {
```
miekg, line 72: Why not just print the test number that you can get from the range loop? We do that in a lot of other tests; that's a 2 line change

yuewko: Printing the test case itself allows you to see immediately what the error is. This is particularly useful when a change triggers failure in multiple test cases, printing the test case allows you to see immediately the failure pattern.

IMO, this is a big time saver.

miekg:
Yes. But we don't do this in any other plugin. So please follow the
(somewhat) standard and just print the sequence number.

yuewko:
If someone encounters UT failure as it is now (without this PR), I’m certain that the first thing he would do is to print out the test case, instead of the sequence no.

I’m afraid I cannot bring myself to fix it the way you suggested, especially when this PR doesn’t break anything.

If you feel strongly about it, please feel free to close this PR.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Print test ca..." ]
>If someone encounters UT failure as it is now (without this PR), I’m certain that the first thing he would do is to print out the test case, instead of the sequence no.
>
>I’m afraid I cannot bring myself to fix it the way you suggested, especially when this PR doesn’t break anything.
>
>If you feel strongly about it, please feel free to close this PR.

Yes, we either do this for all plugin which gives us consistency or not at all. 
As said, just printing `i` as we do everywhere else gives more context than we 
currently have here.

yuewko:
If you are implying that I have to update all the tests to be consistent, I'm afraid that would be out of the question.

Issue #2605, opened 2019-02-24
plugin/errors: both cLogger and eLogger are not needed


miekg:
The struct fields `cLogger` and `eLogger` are not needed and can be easily removed, they look only to be added for testing which can be done like is done in the *log* package.

Hence this can be removed and cleaned up.

rdrozhdzh:
@miekg could you please refer the similar code in _log_ package?

miekg:
This pattern can be re-used: https://github.com/coredns/coredns/blob/master/plugin/log/log_test.go#L29

also the use of regular expression might need a rethink, although this is not directly in the query path.

rdrozhdzh:
ok, I guess I understood now.
You want the test to parse log output instead of intercepting logging functions, right?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/errors..." ]
>ok, I guess I understood now.
>You want the test to parse log output instead of intercepting logging functions, right?

yes, just examing the output should work here as well.

rdrozhdzh:
ok, I will rework the test.

Regarding regular expressions, do you think we can replace them with smth more light-weight like comparing to substrings? We'll need to test if this could give us any benefits comparing to regexp

miekg:
Yes substring match will most definitely be faster. But as said this is
outside the direct serving path, so YMMV.

miekg:
closed via #2657

PR #2608, opened 2019-02-25
Fix error string should not be capitalized

huynguyennovem:
From [Golang coding convention](https://github.com/golang/go/wiki/CodeReviewComments#error-strings):
> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context. That is, use fmt.Errorf("something bad") not fmt.Errorf("Something bad"), so that log.Printf("Reading %s: %v", filename, err) formats without a spurious capital letter mid-message. This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.

N/A

- plugin/errors/errors_test.go
- plugin/etcd/etcd.go
- plugin/pkg/singleflight/singleflight_test.go
- plugin/proxy/grpc_test.go
- plugin/rewrite/name.go
- plugin/rewrite/ttl.go
- plugin/test/helpers.go
- test/external_test.go

N/A

Co-Authored-By: Nguyen Van Trung [trungnvfet@outlook.com](mailto:trungnvfet@outlook.com)
Signed-off-by: Nguyen Quang Huy [huynq0911@gmail.com](mailto:huynq0911@gmail.com)

### DIFF
--- a/plugin/etcd/etcd.go
+++ b/plugin/etcd/etcd.go
@@ -26,7 +26,7 @@ const (
 	etcdTimeout = 5 * time.Second
 )
 
-var errKeyNotFound = errors.New("Key not found")
+var errKeyNotFound = errors.New("key not found")
 
 // Etcd is a plugin talks to an etcd cluster.
 type Etcd struct {
--- a/plugin/pkg/singleflight/singleflight_test.go
+++ b/plugin/pkg/singleflight/singleflight_test.go
@@ -40,7 +40,7 @@ func TestDo(t *testing.T) {
 
 func TestDoErr(t *testing.T) {
 	var g Group
-	someErr := errors.New("Some error")
+	someErr := errors.New("some error")
 	v, err := g.Do(1, func() (interface{}, error) {
 		return nil, someErr
 	})
--- a/plugin/proxy/grpc_test.go
+++ b/plugin/proxy/grpc_test.go
@@ -42,7 +42,7 @@ func buildPool(size int) ([]*healthcheck.UpstreamHost, func(), error) {
 		for _, e := range errs {
 			valErr += fmt.Sprintf("%v\n", e)
 		}
-		return nil, nil, fmt.Errorf("Error at allocation of the pool : %v", valErr)
+		return nil, nil, fmt.Errorf("error at allocation of the pool : %v", valErr)
 	}
 	return ups, stopIt, nil
 }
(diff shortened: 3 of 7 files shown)


miekg:
*errors* is annoying here: `errors_test.go:317: Err0 was never reported by consLogger`

Can you back out the changes to the errors plugin? That code should be removed anyway, see #2605

huynguyennovem:
> _errors_ is annoying here: `errors_test.go:317: Err0 was never reported by consLogger`
> 
> Can you back out the changes to the errors plugin? That code should be removed anyway, see #2605

Should I revert that changes for issue [#2605](https://github.com/coredns/coredns/issues/2605) ?

rdrozhdzh:
> _errors_ is annoying here: `errors_test.go:317: Err0 was never reported by consLogger`
> 
> Can you back out the changes to the errors plugin? That code should be removed anyway, see #2605

I will look into this

rdrozhdzh:
The `TestServeDNSConcurrent` has some error patterns configured, in particular pattern "Failed", see
https://github.com/coredns/coredns/blob/c1534f32f3735902de87a26e3e3cfe20831bdaa3/plugin/errors/errors_test.go#L249

After changing the error message to lower case this pattern stopped matching and the test failed.

See also line
https://github.com/coredns/coredns/blob/c1534f32f3735902de87a26e3e3cfe20831bdaa3/plugin/errors/errors_test.go#L258
where the same pattern is used.

Probably it would be better to define those patterns ("Failed" and "down$") as constants next to error constants, and use them in the test where needed

huynguyennovem:
> The `TestServeDNSConcurrent` has some error patterns configured, in particular pattern "Failed", see
> [coredns/plugin/errors/errors_test.go](https://github.com/coredns/coredns/blob/c1534f32f3735902de87a26e3e3cfe20831bdaa3/plugin/errors/errors_test.go#L249)
> 
> Line 249 in [c1534f3](/coredns/coredns/commit/c1534f32f3735902de87a26e3e3cfe20831bdaa3)
> 
>  pattern: regexp.MustCompile("Failed"), 
> After changing the error message to lower case this pattern stopped matching and the test failed.
> 
> See also line
> [coredns/plugin/errors/errors_test.go](https://github.com/coredns/coredns/blob/c1534f32f3735902de87a26e3e3cfe20831bdaa3/plugin/errors/errors_test.go#L258)
> 
> Line 258 in [c1534f3](/coredns/coredns/commit/c1534f32f3735902de87a26e3e3cfe20831bdaa3)
> 
>  if pattern == "Failed" { 
> 
> where the same pattern is used.
> Probably it would be better to define those patterns ("Failed" and "down$") as constants next to error constants, and use them in the test where needed

If there are some pattern checks, I think we should revert this instead of creating a new one.

rdrozhdzh:
@huynq0911 , I think you may just remove plugin/errors/errors_test.go from this PR. I will fix it in a separate PR

huynguyennovem:
> @huynq0911 , I think you may just remove plugin/errors/errors_test.go from this PR. I will fix it in a separate PR

I have done this.

chrisohaver (approved):

PR #2613, opened 2019-02-25
plugin/template: Correct readme text for upstream option

chrisohaver:
Correct readme text for upstream option

#2600

included

no


miekg:
we're removing the possibility to do this

Issue #2611, opened 2019-02-25
add ready plugin


miekg:
add a new plugin ready, similar to health, but it waits for others plugins to signal readiness (prolly just using caddy events), when all plugins have signaled they are ready (maybe also have a timeout), this *ready* plugin goes into dormant mode and will always report 'ready'.

This solves the issues of a cold cache and other such startup events and prevents catastrophic failure that is inherent to the current health plugin (which will revert back to the process is alive).

miekg:
Events (sadly) don't work, because you just receive an event. This means you don't know when all plugins are done.
This is not too bad, as we can either reuse what we have for *health* or (ab)use `ParsingCallback`s

miekg:
closed via #2616

PR #2616, opened 2019-02-25
Add *ready* plugin

miekg:
Add a ready plugin that allows plugin to signal when they are ready.
This uses simple functions defined in the ready plugin; RegisterPlugin
and Signal.

Add README.md and tests moduled on the health plugin; which will be
relegated to just providing process health. In similar vein to health
this is a process wide setting meaning it will need to include all
plugins for all server block when signalling readiness.

During startup it block forever, while displaying the status every
second:

~~~
2019-02-25T16:09:02.168Z [INFO] plugin/ready: Waiting on 2/2: erratic
2019-02-25T16:09:03.168Z [INFO] plugin/ready: All plugins signal readiness
.:1053
bla.:1053
2019-02-25T16:09:03.168Z [INFO] CoreDNS-1.3.1
2019-02-25T16:09:03.168Z [INFO] linux/amd64, go1.11.4,
CoreDNS-1.3.1
linux/amd64, go1.11.4,
~~~

This was started with this Corefile
~~~
. {
    erratic
    whoami
    ready
}

bla {
    erratic
    whoami
}
~~~

Both *erratic* plugins are reporting readiness.

See #2611 and #2547

### DIFF
--- a/plugin/ready/ready.go
+++ b/plugin/ready/ready.go
@@ -1,0 +1,81 @@
+// Package ready is used to signal readiness of the CoreDNS process. Once all
+// plugins have called in the plugin will signal readiness by returning a 200
+// OK on the HTTP handler (on port 8181). If not ready yet, the handler will
+// return a 503.
+package ready
+
+import (
+	"io"
+	"net"
+	"net/http"
+	"sync"
+
+	clog "github.com/coredns/coredns/plugin/pkg/log"
+	"github.com/coredns/coredns/plugin/pkg/uniq"
+)
+
+var (
+	log      = clog.NewWithPlugin("ready")
+	plugins  = &list{}
+	uniqAddr = uniq.New()
+)
+
+type ready struct {
+	Addr string
+
+	sync.RWMutex
+	ln   net.Listener
+	done bool
+	mux  *http.ServeMux
+}
+
+func (rd *ready) onStartup() error {
+	if rd.Addr == "" {
+		rd.Addr = defAddr
+	}
+
+	ln, err := net.Listen("tcp", rd.Addr)
+	if err != nil {
+		return err
+	}
+
+	rd.Lock()
+	rd.ln = ln
+	rd.mux = http.NewServeMux()
+	rd.done = true
+	rd.Unlock()
+
+	rd.mux.HandleFunc("/ready", func(w http.ResponseWriter, _ *http.Request) {
+		ok, todo := plugins.Ready()
+		if ok {
+			w.WriteHeader(http.StatusOK)
+			io.WriteString(w, "OK")
+			return
+		}
+		log.Infof("Still waiting on: %q", todo)
+		w.WriteHeader(http.StatusServiceUnavailable)
+		io.WriteString(w, todo)
+	})
+
+	go func() { http.Serve(rd.ln, rd.mux) }()
+
+	return nil
+}
+
+func (rd *ready) onRestart() error { return rd.onFinalShutdown() }
+
+func (rd *ready) onFinalShutdown() error {
+	rd.Lock()
+	defer rd.Unlock()
+	if !rd.done {
+		return nil
+	}
+
+	uniqAddr.Unset(rd.Addr)
+
+	rd.ln.Close()
+	rd.done = false
+	return nil
+}
+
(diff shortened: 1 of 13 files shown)


plugin/ready/setup.go
```diff
@@ -0,0 +1,51 @@
...
+		return plugin.Error("ready", err)
+	}
+
+	r := new(addr)
+
+	c.OnStartup(r.onStartup)
+	c.OnStartup(r.wait)
+	c.OnRestart(r.onRestart)
+	c.OnFinalShutdown(r.onFinalShutdown)
+
```
fturib, line 49: Thanks to the new "OnRestartFailed" event that was introduced few months ago in Caddy.
We can handle the failed restart (when the Corefile is invalid).

i guess we could have:
```
        c.OnStartup(r.onStartup)
	c.OnStartup(r.wait)
	c.OnRestart(r.onRestart)
         c.OnRestartFailed(r.onStartup)
	c.OnRestartFailed(r.wait)
	c.OnFinalShutdown(r.onFinalShutdown)
```

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add *ready* p..." ]
>
>
>
>> +	})
>+}
>+
>+func setup(c *caddy.Controller) error {
>+	addr, err := parse(c)
>+	if err != nil {
>+		return plugin.Error("ready", err)
>+	}
>+
>+	r := new(addr)
>+
>+	c.OnStartup(r.onStartup)
>+	c.OnStartup(r.wait)
>+	c.OnRestart(r.onRestart)
>+	c.OnFinalShutdown(r.onFinalShutdown)
>+
>
>Thanks to the new "OnRestartFailed" event that was introduced few months ago in Caddy.
>We can handle the failed restart (when the Corefile is invalid).
>
>i guess we could have:
>```
>        c.OnStartup(r.onStartup)
>	c.OnStartup(r.wait)
>	c.OnRestart(r.onRestart)
>         c.OnRestartFailed(r.onStartup)
>	c.OnRestartFailed(r.wait)
>	c.OnFinalShutdown(r.onFinalShutdown)
>```

oh, that's awesome!

chrisohaver:
Thanks for adding this, although, I was thinking we'd just need _report_ readiness (and not actually block serving until all plugins are ready).

miekg:
fair point. 

It's also global in that it crosses server block boundaries in the current impl. That's not good and counter intuitive; meaning should look even more like health

miekg:
ok, reworked it to;
a) Address Chris' comment
b) make it local to a server block

greenpau (approved):

chrisohaver (approved):

plugin/ready/ready.go
```diff
@@ -0,0 +1,75 @@
...
+	rd.Unlock()
+
+	rd.mux.HandleFunc("/ready", func(w http.ResponseWriter, _ *http.Request) {
+		ok, todo := plugs.Ready()
+		if ok {
+			w.WriteHeader(http.StatusOK)
+			io.WriteString(w, "OK")
+			return
+		}
+		log.Infof("Still waiting on: %q", todo)
```
johnbelamaric, line 55: Is there a reason not to also write this to the body of the response? A 500 error can still have a body. Something like {"plugin-name": false, ...etc...}

I guess it exposes some internals we might now want to.

miekg:
Might be nice to add for when a human runs a curl

plugin/ready/ready.go
chrisohaver: +1 to make it json per your example (i.e. not just for curling humans).

It does expose internals, so might want to allow users to disable it.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add *ready* p..." ]
>+1 to make it json per your example (i.e. not just for curling humans).
>
>It does expose internals, so might want to allow users to disable it.

Thanks, but I think both suggestions are taking this a bit too far.

miekg:
This is now fully ready. PTAL

The proposed changes from @fturib will be done in a seperate PR for: health, prometheus and this plugin.

miekg:
quay.io is broken *sigh*

PR #2627, opened 2019-02-27
Add .gitignore files for Eclipse

vanduc95:
Some files need to ignore when use Eclipse.


yongtang (approved):
LGTM

PR #2630, opened 2019-02-27
plugin/template: fixed ret code

UladzimirTrehubenka:
```
	template IN ANY {
		rcode REFUSED
	}
```
causes write two responses instead one, so any following request fails with FORMERR until connection expired (~ 10 sec).

### DIFF
--- a/plugin/template/template.go
+++ b/plugin/template/template.go
@@ -105,7 +105,7 @@ func (h Handler) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 		}
 
 		w.WriteMsg(msg)
-		return template.rcode, nil
+		return dns.RcodeSuccess, nil
 	}
 
 	return h.Next.ServeDNS(ctx, w, r)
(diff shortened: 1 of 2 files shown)


plugin/template/template.go
```diff
@@ -105,7 +105,7 @@ func (h Handler) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 		}
 
 		w.WriteMsg(msg)
-		return template.rcode, nil
+		return dns.RcodeSuccess, nil
```
chrisohaver, line 110: I think this is correct, but it would be counter-intuitive to someone unfamiliar with the code.
So maybe a small comment in the code to describe why this is always set to Success, regardless of the actual response code.  Perhaps something like ...
```
// Template always writes to client, so always return success here to avoid a second write to client for "error" responses.
```

UladzimirTrehubenka: Done

chrisohaver (approved):

miekg:

/merge

PR #2635, opened 2019-03-01
Add Error log output after lookup failure

fanyeren:
    for example, error like "etcdserver: user name is empty" should be sent to warning administrator.
  nothing to be worried
  nothing to be worried
  no


miekg:
this should be logged if you have the *errors* plugin enable, is it not?

fanyeren:
> this should be logged if you have the _errors_ plugin enable, is it not?

i can not see this kind of errors. the output is quiet.

miekg:
It should... Can you figure out what happens here.

I'm not in favour of adding logs everywhere as we have a single place to do
this.

Issue #2633, opened 2019-02-28
Reload plugin does not detect changes in imported file


oria-startoscale:
Hash 2b6fb578c3d9689461040881ed2e41aa91d1e319
If a block contains an import file line and a reload line, and the imported file is modified, the reload plugin does not detect the change.

miekg:
make sense. not sure we want to support that as well as this requires a parsing of the complete file.

erikwilson:
Seems like it should @miekg , probably not much computation to parse out the complete file or maintain a list of import files to check. 

This would be helpful for separating out system configuration from user configuration and allowing graceful reloads for those conditions. 

At the very least would be nice to have an option such as in the hosts plugin.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Reload plugin..." ]
>Seems like it should @miekg , probably not much computation to parse out the complete file or maintain a list of import files to check.

dunno. If you change your imports it's also not hard to just write a timestamp 
in the main config file. I don't really see why this is up to the main binary to 
track import state separately.

erikwilson:
Because it violates expectations. If I tell users there is a special part of a config map they can edit for their changes, but they need to also modify the Corefile that I overwrite on update, that is a little weird. Or may I need to write something just to watch the user changes to inject data into the corefile, which is also not fun.

johnbelamaric:
I agree with Erik. The only reason it doesn’t is that I didn’t see an easy way to get the list of imports out of Caddy. I can take another look.

erikwilson:
I was looking through the code a bit thinking it might be easiest to implement with an option to the import plugin, would be nice to have an option to not error if an import file does not exist also so I don't need to use globs but I guess all of that stuff is built into caddyfile rather than a separate plugin. The only option I can see is to run caddyfile.Parse and take the md5 of that serialized output.

erikwilson:
Is it bit hacky dealing with import relative path stuff but could use caddyfile.ToJSON to achieve this, I have a change at https://github.com/erikwilson/coredns/commit/a69e3ae187505fdf64da1fb7d11f27a4d355f1b1 if you think it is PR worthy @johnbelamaric.

PR #2638, opened 2019-03-01
plugin/reload: add bug text about import

miekg:
See #2633


miekg:
And I thing to do lgmt and merge is either a bug or race (don't have time to look deeply)

PR #2644, opened 2019-03-03
Move *proxy* to external

miekg:
move the proxy plugin into coredns/proxy and remove it as a default
plugin. Link the proxy to deprecated in plugin.cfg

coredns/proxy doesn't compile because of the vendoring :(

This is backward incompatible (obvs).


bradbeam:
Is this replaced by #2651 ?

bradbeam (approved):

miekg:
yep, see #2651

PR #2645, opened 2019-03-03
Add release notes into coredns repo

### LINKED ISSUE #2355: missing changelog for all releases
Hi!!

I'd like to have a single changelog so I can grep through keywords of interest and easily find changes affecting these keywords.

From what I see there's currently only the Release Notes in the Wiki right?

I'm thinking of making a CHANGELOG.txt having all Release Notes appended to it for the moment. 

How does that sound?

miekg:
Verbatim copy of the current set of notes on coredns.io. Copy them here to
make them authoritative.

'make sync' on the coredns.io site copies them back in. They need to be 
formatted in Hugo style.

Fixes #2355

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -4,9 +4,10 @@
 # used to tag the git repo and to build the assets that are uploaded to github
 # (after some sanity checks).
 #
-# The release should be accompanied by release notes published on coredns.io.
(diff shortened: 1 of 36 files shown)


yongtang (approved):
LGTM

PR #2647, opened 2019-03-03
Remove internal loop detection

miekg:
I can't actually think of a situation where we can create an internal
loop. Sure externally triggered cycles can happen, but this is where the
*loop* plugin comes in that detects those.

Fixes #2602

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -230,12 +230,6 @@ func (s *Server) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 		return
 	}
 
-	ctx, err := incrementDepthAndCheck(ctx)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


chrisohaver (approved):
The loop plugin isn't air tight for some stub domain configurations.  I think there is an open issue for that.  Nevertheless, I think it's OK to remove this loop detection.  I'm a little leery of loop detection in general anyways.  IMO, loop plugin is better because it's A) optional, and B) it only does extra work at start up.

miekg:
quay.io is broken or something on travis
~~~
 docker run -d --net=host --name=etcd quay.io/coreos/etcd:v$ETCD_VERSION
Unable to find image 'quay.io/coreos/etcd:v3.3.8' locally
docker: Error response from daemon: Get https://quay.io/v2/: EOF.
See 'docker run --help'.
~~~

PR #2503, opened 2019-01-26
Add `go mod` support

yongtang:
### 1. Why is this pull request needed and what does it do?

This PR is an attempt to add `go mod` support for coredns.

### 2. Which issues (if any) are related?

n/a

### 3. Which documentation changes (if any) need to be made?

n/a

Signed-off-by: Yong Tang <yong.tang.github@outlook.com>

PR #2649, opened 2019-03-03
Update README.md to use go 1.12

yongtang:
Update README.md to use go 1.12, also check to see if #2503 works as expected with new PRs.


yongtang:
All tests passed now.

bradbeam (approved):

Issue #1476, opened 2018-01-30
plugin/forward: implement gRPC protocol for Upstream hosts


fturib:
Before being able to switch from PROXY to FORWARD, I need to talk gRPC language with the remote DNS server.

I am not sure how to implement, 
what can be reused from current PROXY plugin  ? 
I guess we need a design in current Proxy struct inside Forward to be able to add the gRPC client stuff.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] ExtPlugin/Forward..." ]
>Before being able to switch from PROXY to FORWARD, I need to talk gRPC language with the remote DNS server.
>
>I am not sure how to implement,
>what can be reused from current PROXY plugin  ?
>I guess we need a design in current Proxy struct inside Forward to be able to add the gRPC client stuff.

I'm inclined to go with a new plugin that does this - and not bolt it on 
forward.

Yes this will lead to some code and feature duplication, but will be far easier 
to maintain in the long run.

Thinking along similar lines with the https_google thing.

miekg:
coming back to this and now I think it should be done in *forward*. Mainly because grpc is a first class citizen in CoreDNS (we are a server as well). There are some tricky bits implementation wise:
* the `dial` in persistence.go  possible options, maybe as an empty interface to have the grpc.DialOpts
* the `yield` needs to specify what kind of conn it is yielding (udp/tcp/tls/grpc)
* I hope grpc.ClientConn is also a dns.Conn, we can probably fix that which means we would have a `WriteMsg`, which will probably be cleaner
* the health check needs a smaller tweak, because the grpc isn't support in miekg/dns

miekg:
if we make grpc.ClientConn and dns.Conn a lot of problems disappear (except the health check protocol)

miekg:
Even with a dns.Conn there is no generic dns.Dial so that will still have to be custom

miekg:
Ok, current IETF hackathon to get DNS over HTTPs in miekg/dns proofs this route is futile (I'm sending a report on this to the DoH list). So gRPC will forever be outside miekg/dns as will DoH (some helper functions may be upstreamed.

So for the *forward* plugin, I'll need to figure out how to add gRPC and DoH, which will probably mean it will not use the connection pooling from *forward* as those protcols have that all implement inside their (Go) implementation. This leaves connecting to a domain name (as proposed by DoH), i.e. not an IP. This is not supported yet in CoreDNS, but we're thinking about it in #1592)

[...]

PR #2655, opened 2019-03-05
[WIP] plugin/forward: gRPC support

inigohu:
To support gRPC connection in forward plugin

#1476 

Forward plugin documentation.

It shouldn't be


inigohu:
Here is my first approach of how implement a gRPC proxy in Forward plugin. The idea is the same as I mentioned in #1476 issue. Create a proxy interface that satisfied **gRPC** and **DNS** proxies, as well as **DoH** proxy in the future.

Here I attach a schema that helps me visualizing the new structure of the plugin.
coredns-forward-new 1

Here is the forward plugin's new workflow.
coredns-forward-new 2


/cc @glerchundi

miekg:
Thanks. Before I take a look I want to pose the question: how realistic is
it to make this a new plugin that only does gRPC forwarding?

I slightly concerned with making forward as complex as proxy was, which
will hamper future refactoring, etc.

miekg:
I'm really on two thoughts here, hence my question.

One thing in favour of doing a new plugin is TLS config which now applies
to the entire proxy. And there a bunch of options that make no sense for
gRPC and DoH

inigohu:
> how realistic is it to make this a new plugin that only does gRPC forwarding?

Is a good option and looks good to me. That would simplified de forward plugin implementation without mixing different proxies configurations like the TLS configuration as you mentioned.

I can make you a proposal of a plugin with just gRPC connection.

Which would be the name of the plugin? _gRPC-forward_?
Should be in this repository o maybe an external plugin?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [WIP] plugin/..." ]
>> how realistic is it to make this a new plugin that only does gRPC forwarding?
>
>Is a good option and looks good to me. That would simplified de forward plugin implementation without mixing different proxies configurations like the TLS configuration as you mentioned.
>
>I can make you a proposal of a plugin with just gRPC connection.
>
>Which would be the name of the plugin? _gRPC-forward_?

I think we can just name it grpc (I like short name for this)

>Should be in this repository o maybe an external plugin?

Useful enough to include by default.

/Miek

--
Miek Gieben

miekg:
closing in favor of #2667

PR #2661, opened 2019-03-08
[WIP]: test pr

rajansandeep:


miekg:
is there something we can do to avoid to test prs? They add a lot of noise to an already noisy inbox.

chrisohaver:
> is there something we can do to avoid to test prs? They add a lot of noise to an already noisy inbox.

Yes, we should be able to set things up so we can do them from our forks.

PR #2667, opened 2019-03-11
plugin/grpc: New gRPC plugin

inigohu:
For proxying DNS messages to upstream resolvers using gRPC connection

#1476 
-
No

### DIFF
--- a/plugin.cfg
+++ b/plugin.cfg
@@ -53,7 +53,7 @@ etcd:etcd
 loop:loop
 forward:forward
 proxy:deprecated
+grpc:grpc
 erratic:erratic
 whoami:whoami
 on:github.com/mholt/caddy/onevent
-grpc:grpc
--- a/plugin.cfg
+++ b/plugin.cfg
@@ -56,3 +56,4 @@ proxy:deprecated
 erratic:erratic
 whoami:whoami
 on:github.com/mholt/caddy/onevent
+grpc:grpc
--- a/plugin/grpc/grpc.go
+++ b/plugin/grpc/grpc.go
@@ -73,11 +73,6 @@ func (g *GRPC) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 			child.Finish()
 		}
 
-		// Continue if no answer has been found
-		if ret.Rcode == dns.RcodeNameError {
-			continue
-		}
-
 		// Check if the reply is correct; if not return FormErr.
 		if !state.Match(ret) {
 			debug.Hexdumpf(ret, "Wrong reply for id: %d, %s %d", ret.Id, state.QName(), state.QType())
--- a/plugin/grpc/grpc.go
+++ b/plugin/grpc/grpc.go
@@ -3,28 +3,27 @@ package grpc
 import (
 	"context"
 	"crypto/tls"
-	"errors"
 	"time"
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/debug"
 	"github.com/coredns/coredns/request"
+
 	"github.com/miekg/dns"
 	ot "github.com/opentracing/opentracing-go"
 )
 
-// GRPC is an grpc plugin to show how to write a plugin.
+// GRPC represents a plugin instance that can proxy requests to another (DNS) server via gRPC protocol.
+// It has a list of proxies each representing one upstream proxy.
 type GRPC struct {
-	proxies    []*Proxy
-	p          Policy
-	hcInterval time.Duration
+	proxies []*Proxy
+	p       Policy
 
 	from    string
 	ignored []string
 
 	tlsConfig     *tls.Config
 	tlsServerName string
-	maxfails      uint32
 
 	Next plugin.Handler
 }
@@ -40,7 +39,6 @@ func (g *GRPC) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 		span, child ot.Span
 		ret         *dns.Msg
 		err         error
-		fails       int
 		i           int
 	)
 	span = ot.SpanFromContext(ctx)
@@ -60,15 +58,6 @@ func (g *GRPC) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 		proxy := list[i]
 		i++
 
-		if proxy.down(g.maxfails) {
-			fails++
-			if fails < len(g.proxies) {
-				continue
-			}
-			// All upstream proxies are dead
-			return dns.RcodeServerFailure, ErrNoHealthy
-		}
-
 		if span != nil {
 			child = span.Tracer().StartSpan("query", ot.ChildOf(span.Context()))
 			ctx = ot.ContextWithSpan(ctx, child)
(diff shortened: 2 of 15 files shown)


plugin/grpc/grpc.go
```diff
@@ -0,0 +1,175 @@
...
+import (
+	"context"
+	"crypto/tls"
+	"errors"
+	"time"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/debug"
+	"github.com/coredns/coredns/request"
+	"github.com/miekg/dns"
```
miekg, line 12: seperate coredns imports from 3rd party with a newline

plugin/grpc/grpc.go
```diff
@@ -0,0 +1,175 @@
...
+	ot "github.com/opentracing/opentracing-go"
+)
+
+// GRPC is an grpc plugin to show how to write a plugin.
```
miekg: comment needs adjusting

plugin/grpc/proxy.go
```diff
@@ -0,0 +1,146 @@
...
+
+	if opts.TLSConfig != nil {
+		p.dialOpts = append(p.dialOpts, grpc.WithTransportCredentials(credentials.NewTLS(opts.TLSConfig)))
+	} else {
+		p.dialOpts = append(p.dialOpts, grpc.WithInsecure())
+	}
+
+	conn, err := grpc.Dial(p.addr, p.dialOpts...)
+	if err != nil {
+		log.Warningf("Skipping gRPC host '%s' due to Dial error: %s\n", p.addr, err)
```
miekg: should probably just return the error instead of logging or continuing?

plugin/grpc/proxy.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	RequestCount.WithLabelValues(p.addr).Add(1)
+	RcodeCount.WithLabelValues(rc, p.addr).Add(1)
+	RequestDuration.WithLabelValues(p.addr).Observe(time.Since(start).Seconds())
+
+	return ret, nil
+}
+
+// healthcheck kicks of a round of health checks for this proxy.
```
miekg: doesn't gRPC have all kinds of stuff for this? I.e reconnecting when the connection breaks etc?

plugin/grpc/proxy_test.go
```diff
@@ -0,0 +1,83 @@
...
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			var mock *mockDNSServiceClient
+			if tt.res != nil {
+				msg, err := tt.res.Pack()
+				if err != nil {
+					t.Fatalf("Error packing response: %s", err.Error())
+				}
+				mock = newMockDNSServiceClient(&pb.DnsPacket{Msg: msg}, nil)
```
miekg: would just call it testServiceClient or some such.

plugin/grpc/setup.go
```diff
@@ -0,0 +1,187 @@
...
+		g, err = ParseGRPCStanza(&c.Dispenser)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return g, nil
+}
+
+// ParseGRPCStanza parses one grpc stanza
+func ParseGRPCStanza(c *caddyfile.Dispenser) (*GRPC, error) {
```
miekg: doesn't have to be public

miekg (commented):
thanks!

there is no of course some code duplication between forward and grpc - which is fine, but it does make sense to cut back on that a little (there will at some point be a 'doh' proxy as well.

I'm fine with delaying de-dupping this, after this is merged.

plugin/grpc/proxy.go
inigohu: You're right. I've deleted all **healtcheck** stuff.

inigohu:
I've this error in the CI 

```
build github.com/coredns/coredns: cannot find module for path github.com/DataDog/dd-trace-go/tracer
Makefile:17: recipe for target 'coredns' failed
make[1]: *** [coredns] Error 1
```

I've realized that in _go.mod_ (master branch) file are 3 dependencies referencing to _DataDog/dd-trace-go_ . This may be the cause

```
github.com/DataDog/dd-trace-go v0.6.1
gopkg.in/DataDog/dd-trace-go.v0 v0.6.1
gopkg.in/DataDog/dd-trace-go.v1 v1.10.0
```

miekg:
Latest version fixes that. Fighting with it all morning

miekg:
rebasing on top of master (think you tried that w/ the force push) should fix it

inigohu:
> rebasing on top of master (think you tried that w/ the force push) should fix it

The branch is up to date with master and still get an error.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/grpc: ..." ]
>I've this error in the CI
>
>```
>build github.com/coredns/coredns: cannot find module for path github.com/DataDog/dd-trace-go/tracer
>Makefile:17: recipe for target 'coredns' failed
>make[1]: *** [coredns] Error 1
>```
>
>I've realized that in _go.mod_ (master branch) file are 3 dependencies referencing to _DataDog/dd-trace-go_ . This may be the cause
>
>```
>github.com/DataDog/dd-trace-go v0.6.1
>gopkg.in/DataDog/dd-trace-go.v0 v0.6.1
>gopkg.in/DataDog/dd-trace-go.v1 v1.10.0
>```

go modules has a learning curve.

anyway, things in master should be better now.

lopezator:
np @miekg we're working together. thanks!

plugin.cfg
```diff
@@ -56,3 +56,4 @@ proxy:github.com/coredns/proxy
 erratic:erratic
 whoami:whoami
 on:github.com/mholt/caddy/onevent
+grpc:grpc
```
miekg: this should be near *forward*, to keep forwarders together. Order matters here.

miekg:
this also needs an owner file with you in it, which also implies I will add you as a maintainer to the coredns project. (Unless you object, which is fine, but then we need to discuss what we do here)

inigohu:
> this also needs an owner file with you in it, which also implies I will add you as a maintainer to the coredns project. (Unless you object, which is fine, but then we need to discuss what we do here)

Of course, you can add me as a maintainer.
Thanks

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/grpc: ..." ]
>> this also needs an owner file with you in it, which also implies I will add you as a maintainer to the coredns project. (Unless you object, which is fine, but then we need to discuss what we do here)
>
>Of course, you can add me as a maintainer.
>Thanks

invite send

inigohu:
> invite send

@miekg do you want to make a review or can I merge it?

miekg:
I wanna do a last pass

plugin/grpc/grpc.go
```diff
@@ -0,0 +1,135 @@
...
+			// Continue with the next proxy
+			continue
+		}
+
+		if child != nil {
+			child.Finish()
+		}
+
+		// Continue if no answer has been found
+		if ret.Rcode == dns.RcodeNameError {
```
miekg: really, we do this this in forward as well? (assuming the code is copied). This odd

miekg (commented):
one question, for the rest lgtm

plugin/grpc/grpc.go
inigohu: No, I added this check to continue searching in other proxies if no results has been found.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/grpc: ..." ]
>+		// Continue if no answer has been found
>+		if ret.Rcode == dns.RcodeNameError {
>
>No, I added this check to continue searching in other proxies if no results has been found.

that's not the job of the proxy and can be removed

miekg (approved):

PR #2670, opened 2019-03-11
Remove pkg/healthcheck

miekg:
This original came from the proxy plugin, but as that is now external
this can be removed.
However the k8s plugin also used it. That code has been removed in a
recent PR; meaning this code has no user anymore and can be included
in the proxy repo and removed from this.


miekg:
@yongtang I can't get 'go mod' to just update the coredns/proxy dep; it *wants* to update the rest and this is leading to compile errors. Let me try another tack

PR #2682, opened 2019-03-12
Custom tag support in trace plugin

stefanmb:
This PR adds optional custom tag support to the `trace` plugin. Zero or more tags are supported, and they are added to the emitted spans.

I am managing a large deployment of CoreDNS instances with multiple configurations and exporting sampled traces to Stackdriver and Lightstep. Without custom tags it's hard to tell which instances the traces correspond to.

None

`trace` README should explain the use of the `tag` block, this PR includes the necessary changes. 

No

### DIFF
--- a/plugin/trace/README.md
+++ b/plugin/trace/README.md
@@ -30,6 +30,7 @@ trace [ENDPOINT-TYPE] [ENDPOINT] {
 	every AMOUNT
 	service NAME
 	client_server
+	tag key value
 }
 ~~~
 
@@ -38,6 +39,7 @@ trace [ENDPOINT-TYPE] [ENDPOINT] {
 * `service` **NAME** allows you to specify the service name reported to the tracing server.
   Default is `coredns`.
 * `client_server` will enable the `ClientServerSameSpan` OpenTracing feature.
+* `tag` is used to emit custom tags with each trace, it can be included multiple times
 
 ## Zipkin
 You can run Zipkin on a Docker host like this:
--- a/plugin/trace/README.md
+++ b/plugin/trace/README.md
@@ -40,7 +40,7 @@ trace [ENDPOINT-TYPE] [ENDPOINT] {
   Default is `coredns`.
 * `client_server` will enable the `ClientServerSameSpan` OpenTracing feature.
 * `tag` is used to emit custom tags with each trace, it can be included multiple times
-  If the `metadata` plugin used the value can be dynamic, use the format `{prefix/name}`
+  If the `metadata` plugin is used the value can be dynamic, use the format `{plugin/value}`
 
 ## Zipkin
 You can run Zipkin on a Docker host like this:
--- a/plugin/trace/README.md
+++ b/plugin/trace/README.md
@@ -40,6 +40,7 @@ trace [ENDPOINT-TYPE] [ENDPOINT] {
   Default is `coredns`.
 * `client_server` will enable the `ClientServerSameSpan` OpenTracing feature.
 * `tag` is used to emit custom tags with each trace, it can be included multiple times
+  If the `metadata` plugin used the value can be dynamic, use the format `{prefix/name}`
 
 ## Zipkin
 You can run Zipkin on a Docker host like this:
--- a/plugin/trace/trace.go
+++ b/plugin/trace/trace.go
@@ -9,6 +9,7 @@ import (
 	"sync/atomic"
 
 	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/metadata"
 	"github.com/coredns/coredns/plugin/metrics"
 	"github.com/coredns/coredns/plugin/pkg/dnstest"
 	"github.com/coredns/coredns/plugin/pkg/rcode"
@@ -122,10 +123,24 @@ func (t *trace) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 	span.SetTag(tagRcode, rcode.ToString(rw.Rcode))
 
 	for key, value := range t.tags {
-		span.SetTag(key, value)
+		value = tagValue(ctx, value)
+		if value != "" {
+			span.SetTag(key, value)
+		}
 	}
 
 	return status, err
+}
+
+func tagValue(ctx context.Context, value string) string {
+	if len(value) > 2 && value[0] == '{' && value[len(value)-1] == '}' {
+		fetcher := metadata.ValueFunc(ctx, value[1:len(value)-1])
+		if fetcher != nil {
+			return fetcher()
+		}
+		return ""
+	}
+	return value
 }
 
 func spanName(ctx context.Context, req request.Request) string {
--- a/plugin/trace/trace.go
+++ b/plugin/trace/trace.go
@@ -37,6 +37,7 @@ type trace struct {
 	serviceEndpoint string
 	serviceName     string
 	tags            map[string]string
+	tagFetchers     map[string]metadata.Func
 	clientServer    bool
 	every           uint64
 	count           uint64
@@ -122,25 +123,26 @@ func (t *trace) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 	span.SetTag(tagType, req.Type())
 	span.SetTag(tagRcode, rcode.ToString(rw.Rcode))
 
-	for key, value := range t.tags {
-		value = tagValue(ctx, value)
-		if value != "" {
-			span.SetTag(key, value)
+	for key, internalKey := range t.tags {
+		fetcher := metadata.ValueFunc(ctx, internalKey)
+		if fetcher != nil {
+			value := fetcher()
+			if value != "" {
+				span.SetTag(key, value)
+			}
 		}
 	}
 
 	return status, err
 }
 
-func tagValue(ctx context.Context, value string) string {
-	if len(value) > 2 && value[0] == '{' && value[len(value)-1] == '}' {
-		fetcher := metadata.ValueFunc(ctx, value[1:len(value)-1])
+func (t *trace) Metadata(ctx context.Context, state request.Request) context.Context {
+	for key, fetcher := range t.tagFetchers {
 		if fetcher != nil {
-			return fetcher()
+			metadata.SetValueFunc(ctx, key, fetcher)
 		}
-		return ""
 	}
-	return value
+	return ctx
 }
 
 func spanName(ctx context.Context, req request.Request) string {
--- a/plugin/trace/trace.go
+++ b/plugin/trace/trace.go
@@ -35,6 +35,7 @@ type trace struct {
 	tracer          ot.Tracer
 	serviceEndpoint string
 	serviceName     string
+	tags            map[string]string
 	clientServer    bool
 	every           uint64
 	count           uint64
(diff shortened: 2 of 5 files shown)


plugin/trace/trace.go
```diff
@@ -120,6 +121,10 @@ func (t *trace) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 	span.SetTag(tagType, req.Type())
 	span.SetTag(tagRcode, rcode.ToString(rw.Rcode))
 
+	for key, value := range t.tags {
+		span.SetTag(key, value)
```
johnbelamaric: This is a good change. One thing - rather than treating `value` as simple static text, it should be able to be a metadata value. See https://coredns.io/plugins/metadata/

stefanmb: @johnbelamaric Thank you for the suggestion!

I've implemented your request based on the `rewrite` example. This functionality is significantly more complex than what I need, and introduces more cross-plugin compile-time dependencies but I could see it would be useful to others.

Let me know if this is what you had in mind, thanks.

plugin/trace/README.md
```diff
@@ -38,6 +39,8 @@ trace [ENDPOINT-TYPE] [ENDPOINT] {
 * `service` **NAME** allows you to specify the service name reported to the tracing server.
   Default is `coredns`.
 * `client_server` will enable the `ClientServerSameSpan` OpenTracing feature.
+* `tag` is used to emit custom tags with each trace, it can be included multiple times
+  If the `metadata` plugin used the value can be dynamic, use the format `{prefix/name}`
```
miekg: the format should prolly be `{trace/tagname}`?

miekg (commented):
Where do you implement the `metadata.Provider` interface ?

plugin/trace/README.md
stefanmb: > Where do you implement the metadata.Provider interface ?

I don't. I'm not adding any new metadata from the `trace` plugin, I'm allowing it to be read from other plugins by using the syntax:

```
tag tagName {someOtherPlugin/dynamicValue}
```

The `prefix` should be the name of the plugin that *writes* the value (not `trace` which only *reads* the value).

I may have misunderstood what @johnbelamaric was asking for, but I believe what he wanted was the ability to read tag values dynamically from other plugins via the metadata API.

The implementation is based on the `rewrite` plugin which also doesn't implement `metadata.Provider`, but reads values from it if the `{...}` syntax is used in the rewrite rules.

**Edit:** Ok, I understand why implementing Provider is necessary, I'll make the changes, thanks!

plugin/trace/trace.go
```diff
@@ -120,9 +122,27 @@ func (t *trace) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
+		value = tagValue(ctx, value)
+		if value != "" {
+			span.SetTag(key, value)
+		}
+	}
+
 	return status, err
 }
 
+func tagValue(ctx context.Context, value string) string {
```
johnbelamaric: I think all this can be done during setup time. In the hot path it should just call the value func. That is, change `tags` to be type `map[string]metadata.Func`. Then you just call that func that you look up (which might return a static value if there are no metadata tags in it).

I think code in https://github.com/coredns/coredns/blob/master/plugin/metadata/provider.go can handle most of this for you. To avoid depending on the metadata plugin directly, it probably makes sense to move a lot of that stuff into a plugin/pkg that is used by any plugin processing or providing metadata. But that can be a different PR.

stefanmb: Ok so for the static values they'll simply be aded to the provider on setup?

That makes sense, I'll try to get this done tomorrow, thanks!

stefanmb: > To avoid depending on the metadata plugin directly, it probably makes sense to move a lot of that stuff into a plugin/pkg that is used by any plugin processing or providing metadata. But that can be a different PR.

I'm struggling with this aspect of CoreDNS in general. Not all plugins seem equal. Some provide discrete functionality, others do cross cutting concerns (e.g. tracing) and cross-plugin compile time dependencies are allowed (metadata, trace, dnstap, etc). The latter is scary because it means encumbrance between plugins only increases with time. Anyway, I think you're right - further separation could be done in follow-up work.

Thanks for the quick reviews!

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Custom tag su..." ]
>+func tagValue(ctx context.Context, value string) string {
>
>> To avoid depending on the metadata plugin directly, it probably makes sense to move a lot of that stuff into a plugin/pkg that is used by any plugin processing or providing metadata. But that can be a different PR.
>
>I'm struggling with this aspect of CoreDNS in general. Not all plugins seem equal. Some provide discrete functionality, others do cross cutting concerns (e.g. tracing) and cross-plugin compile time dependencies are allowed (metadata, trace, dnstap, etc). The latter is scary because it means encumbrance between plugins only increases with time. Anyway, I think you're right - further separation could be done in follow-up work.

As coredns evolved different use cases required different kinds of plugins, i.e. 
*cache* only wraps a dns.ReponseWriter, *autopath* required looking to the 
kubernetes plugin. We never categorized them, but I think it makes sense to 
start doing this in the documentation and provide an official list of what kind
of plugins exist.

stefanmb:
@miekg @johnbelamaric I've implemented what I understood was requested, but I think it's needlessly complex. Here is why:

1. What is the advantage of using a metadata provider as the backend for the tags inside the trace plugin? Do other plugins need to access OpenTracing tags using the metadata API? Why? This is an internal detail of the trace plugin.

1. We don't save much execution time in the "hot" path. The Provider interface is called *per request*, so we must populate the list of "static" tag getters on every request. We then have to call each of these getters from the same `trace` plugin. Why not just store our tags internally in the `trace` plugin if that's the only place they are used (see also question above)?

1. The dynamic tags cannot be resolved at setup time because we do not have a context, and tags that are resolved through metadata API can differ between requests. This means we need an additional indirection layer I called `innerKey`, examples:

```
trace {
  tag foo bar
}
```

^ this gets stored with the metadata API by the trace plugin, then requested back out of it by the same plugin. The metadata API wants a key with the format `trace/foo`, but I do not want to leak the `trace/` prefix to OpenTracing, because it's not necessary. Therefore we need to keep track of this internal key mapping: key is "foo", innerKey is "trace/foo", value is "bar".  This is wasteful for little gain IMO.

```
trace {
  tag foo {otherPlugin/bar}
}
```

^ this cannot be resolved at setup time, it must be resolved for each ServeDNS call because the value returned by `otherPlugin` can change with each call.

We still need an internal mapping that says `innerKey` is `otherPlugin/bar` but the OpenTracing key is `foo`.

Please let me know how you want to proceed.

If you want these features I'm ok with merging them, but I do not understand why the static tags must be stored as metadata.

Thanks!

johnbelamaric:
On Wed, Mar 13, 2019 at 1:56 PM Stefan Budeanu <notifications@github.com>
wrote:

> @miekg <https://github.com/miekg> @johnbelamaric
> <https://github.com/johnbelamaric> I've implemented what I understood was
> requested, but I think it's needlessly complex. Here is why:
>
>    1.
>
>    What is the advantage of using a metadata provider as the backend for
>    the tags inside the trace plugin? Do other plugins need to access
>    OpenTracing tags using the metadata API? Why? This is an internal detail of
>    the trace plugin.
>
>
None, I didn't intend to say that should be done. I don't think the trace
plugin needs to implement the provider interface. The trace plugin should
be a consumer but need not be a provider.


>
>    1.
>    2.
>
>    We don't save much execution time in the "hot" path. The Provider
>    interface is called *per request*, so we must populate the list of
>    "static" tag getters on every request. We then have to call each of these
>    getters from the same trace plugin. Why not just store our tags
>    internally in the trace plugin if that's the only place they are used
>    (see also question above)?
>
>
I don't think we need to do this. If it's a static shouldn't we just store
a func that returns the static value.

>
>    1.
>    2.
>
>    The dynamic tags cannot be resolved at setup time because we do not
>    have a context, and tags that are resolved through metadata API can differ
>    between requests. This means we need an additional indirection layer I
>    called innerKey, examples:
>
> trace {
>   tag foo bar
> }
>
> ^ this gets stored with the metadata API by the trace plugin, then
> requested back out of it by the same plugin. The metadata API wants a key
> with the format trace/foo, but I do not want to leak the trace/ prefix to
> OpenTracing, because it's not necessary. Therefore we need to keep track of
> this internal key mapping: key is "foo", innerKey is "trace/foo", value is
> "bar". This is wasteful for little gain IMO.
>
trace {
>   tag foo {otherPlugin/bar}
> }
>
> ^ this cannot be resolved at setup time, it must be resolved for each
> ServeDNS call because the value returned by otherPlugin can change with
> each call.
>
> We still need an internal mapping that says innerKey is otherPlugin/bar
> but the OpenTracing key is foo.
>
No, this is mixed up. I think things went sideways. This should work like
the log plugin. The metadata integration there is 2 lines of code:

                        rep := replacer.New(ctx, r, rrw,
CommonLogEmptyValue)
                        clog.Infof(rep.Replace(rule.Format))

I misfired when I said use a map[string]Func. You can go back to
map[string]string but just pass the tag value string through the Replacer.

Sorry about that.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Custom tag su..." ]
>None, I didn't intend to say that should be done. I don't think the trace
>plugin needs to implement the provider interface. The trace plugin should
>be a consumer but need not be a provider.

ok, I misunderstood your comment then

plugin/trace/trace.go
```diff
@@ -120,9 +123,28 @@ func (t *trace) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
+			if value != "" {
+				span.SetTag(key, value)
+			}
+		}
+	}
+
 	return status, err
 }
 
+func (t *trace) Metadata(ctx context.Context, state request.Request) context.Context {
```
miekg, line 139: So this isn't actually needed as trace will not provide metadata to other plugins?

johnbelamaric: Correct

plugin/trace/trace.go
```diff
@@ -120,9 +123,28 @@ func (t *trace) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 	span.SetTag(tagType, req.Type())
 	span.SetTag(tagRcode, rcode.ToString(rw.Rcode))
 
+	for key, internalKey := range t.tags {
```
johnbelamaric, line 126: Instead of implementing Provider below to handle static values, I think it's simpler to just treat static and dynamic values separately.

That mean:
1) your setup code is simpler, you just need the tags not the tagFetchers
2) this code will just loop through the tags, and it will look and see if the value has { }; if so, it calls ValueFunc; if not, it just writes the value directly
3) you don't need to implement provider

I think that will work, with less code, won't it?

miekg:
Is this being worked on? Otherwise I'll  close it

stefanmb:
@miekg I dropped the ball on this due to other priorities. I'm going to close it for now, once I get a chance to redo it based on the feedback I can open up a new PR. Thanks for your patience.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Custom tag su..." ]
>@miekg I dropped the ball on this due to other priorities. I'm going to close it for now, once I get a chance to redo it based on the feedback I can open up a new PR. Thanks for your patience.

no worries; we're all volunteers here :)

PR #2690, opened 2019-03-13
Metric for measuring DNS Programming Latency SLI.

mm4tt:
The DNS Programming Latency definition can be found [here](https://github.com/kubernetes/community/blob/master/sig-scalability/slos/dns_programming_latency.md).

The PR bears some similarities to https://github.com/kubernetes/kubernetes/pull/71999, which introduced the In-Cluster Network Programming Latency SLI metric.
The main difference is that there is no actual "programming" happening in the CoreDNS. 
The CoreDNS serves the content directly from the endpoints/service/pod cache, creating DNS records on the fly (for comparison, in kube-proxy the network programming consists of writing IPTables/IPVS rules).
Thus, this PR assumes that the programming of DNS ends in the moment when the endpoints/service/pod change reaches the CoreDNS via the watch mechanism.

This PR covers only "headless with selector" services, other service kinds are blocked on the impossibility of determining the LastUpdateTime of a service object.

It introduces a Prometheus metric for the DNS Programming Latency SLI.

None

Kubernetes plugin metric's documentation has changed, I've updated the plugin/kubernetes/README.md

No


mm4tt:
@wojtek-t, could you take a look?

CoreDNS folks, please wait for Wojtek's LGTM before reviewing.

plugin/kubernetes/controller.go
```diff
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/coredns/coredns/plugin/kubernetes/object"
+	"github.com/coredns/coredns/plugin/metrics/vars"
```
miekg: there is no reason to include this, you can just define a bunch of metrics in the plugin and export those.

plugin/kubernetes/object/endpoint.go
```diff
@@ -1,18 +1,23 @@
...
-	Index     string
-	IndexIP   []string
-	Subsets   []EndpointSubset
+	Version               string
+	Name                  string
+	Namespace             string
+	Index                 string
+	IndexIP               []string
+	Subsets               []EndpointSubset
+	LastChangeTriggerTime time.Time
```
miekg, line 20: this increases memory used

plugin/metrics/README.md
```diff
@@ -19,6 +19,7 @@ The following metrics are exported:
 * `coredns_dns_request_type_count_total{server, zone, type}` - counter of queries per zone and type.
 * `coredns_dns_response_size_bytes{server, zone, proto}` - response size in bytes.
 * `coredns_dns_response_rcode_count_total{server, zone, rcode}` - response per zone and rcode.
+* `coredns_dns_programming_latency_seconds{service_kind}` - [DNS programming latency SLI](https://github.com/kubernetes/community/blob/master/sig-scalability/slos/dns_programming_latency.md)
```
miekg: this metrics should be namespaced in the kubernetes plugin (and moved there) and named
`coredns_kubernetes_programming_latency`

plugin/metrics/README.md
```diff
@@ -33,6 +34,12 @@ Extra labels used are:
   NS, SRV, DS, DNSKEY, RRSIG, NSEC, NSEC3, IXFR, AXFR and ANY) and "other" which lumps together all
   other types.
 * The `response_rcode_count_total` has an extra label `rcode` which holds the rcode of the response.
+* The `coredns_dns_programming_latency_seconds` metric has an extra `service_kind` label that
```
miekg: this can be moved there as well, under a new section `## Metrics`

miekg:
why does this need to happen in coredns?

miekg:
also this: https://github.com/coredns/coredns/blob/master/plugin/kubernetes/object/object.go#L4

plugin/kubernetes/controller.go
mm4tt: Done.

plugin/metrics/README.md
mm4tt: Done

plugin/metrics/README.md
mm4tt: Done

plugin/kubernetes/object/endpoint.go
wojtek-t: I obviously don't know this code well enough, but I think this is acceptable.

In general, we are currently saying that we support up to 10k services in k8s. We are planning to bump it in the future to 100k. but given time.Time is 8B IIRC, we are talking about 80kB now and 800kB in the future.

I wouldn't consider it as a problem (epsecially given that a single endpoints object may have 1mB size...)

mm4tt:
> why does this need to happen in coredns?

CoreDns is becoming a standard way of doing DNS discovery in Kubernetes clusters. We'd like to cover it in our scalability SLI/sSLOs to guarantee it scales well in large clusters. 

More details about scalability scalability SLIs/SLOs can be found at https://github.com/kubernetes/community/blob/master/sig-scalability/slos/slos.md.
The DNS programming latency is described here - https://github.com/kubernetes/community/blob/master/sig-scalability/slos/dns_programming_latency.md

miekg:
I disagree. By adding this here the onus is now on us to make it work again
when it fails. This was no fun last time, so adding values here needs a
much stronger argument then you're giving.

Also these metrics can be retrieved completely independent from CoreDNS.

As such I'm NACKing is PR.

wojtek-t:
> Also these metrics can be retrieved completely independent from CoreDNS.

I disagree with that - how you can do that independently from CoreDNS - you can try to estimate with probes, but probes is not a very accurate mechanism. So I don't think it's a good idea.

@miekg - so just let me understand one thing - is your main problem with adding this field? If we wouldn't add it, you will be willing to accept it?

miekg:
This hasn't seen any public discussion, just this PR..what alternatives
have been considered and why was this approach choosen?

Not adding a time.Time is a good first step.

wojtek-t:
> This hasn't seen any public discussion, just this PR..what alternatives
have been considered and why was this approach choosen?

This had a public discussion when we were defining the SLI:
https://github.com/kubernetes/community/pull/2986
Though I agree it was generic discussion, not specific to CoreDNS.

We were considering the "probe-based" approach vs components directly exposing metrics and decided for the latter.

> Not adding a time.Time is a good first step.

Hmm - let me understand it better - I would like to understand if this has a chance to be approved if that is fixed, or it's still unclear.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Metric for me..." ]
>> This hasn't seen any public discussion, just this PR..what alternatives
>have been considered and why was this approach choosen?
>
>This had a public discussion when we were defining the SLI:
>https://github.com/kubernetes/community/pull/2986
>Though I agree it was generic discussion, not specific to CoreDNS.
>
>We were considering the "probe-based" approach vs components directly exposing metrics and decided for the latter.
>
>> Not adding a time.Time is a good first step.
>
>Hmm - let me understand it better - I would like to understand if this has a chance to be approved if that is fixed, or it's still unclear.

Yes, if this can be done without a memory increase in these components I don't 
see why extra metrics are bad.

Thanks for the background, would have helped to have seen that in the intial PR 
text.

wojtek-t:
Thanks Miek - @mm4tt is on vacation this week, but we will look into it after that.

mm4tt:
@miekg, thanks for your cooperation and sorry about not including more detailed background in the PR description.

I got rid of the lastChangeTriggerTimes map, which should reduce the amount of memory introduced by this code at least by half. 
Unfortunately, we still have the time.Time in the stripped down Endpoints object and honestly I don't see a way of getting rid of that. 
If we remove it from the stripped down object, then we will lose the LastChangeTriggerTime information completely, and there will be nothing to base the latency calculation on. 

Let me know what do you think.

@wojtek-t, I'd appreciate if you could also take a look.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Metric for me..." ]
>Unfortunately, we still have the time.Time in the stripped down Endpoints object and honestly I don't see a way of getting rid of that.

this is problematic and requires a full scale test to see what it adds. Because 
the feature itself is so small (this doesn't benefit a lot of users), I don't 
see how we can make a case for adding a time.Time here.... (even if the scale 
test says we're OK now). In fact we want to trim this struct further...

/Miek

--
Miek Gieben

wojtek-t:
> this is problematic and requires a full scale test to see what it adds. 

That's definitely something we should do.
Is there any procedure that describes how to build an experimental image of CoreDNS? [This is something we need to run a test, right?]

wojtek-t:
> Because the feature itself is so small (this doesn't benefit a lot of users), I don't
see how we can make a case for adding a time.Time here

@miekg - I disagree with that - I mean with the "this doesn't benefit a lot of users" part.
As a user, I'm really interested in knowing how fast changes in my k8s cluster are reflected
in DNS (that kind of answers the question how fresh data DNS serves).
And note that this is not something we can really answer within DNS server, because it may
be the resulted by say slow watch in apiserver or things like that.
So I actually claim that this will benefit every single user that is interested in knowing if DNS
is propagated fast or not (which I think is almost every user).

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Metric for me..." ]
>> this is problematic and requires a full scale test to see what it adds.
>
>That's definitely something we should do.
>Is there any procedure that describes how to build an experimental image of CoreDNS? [This is something we need to run a test, right?]

you can use the Makefile.release to build an image with a homebacked coredns. 
This can be uploaded to a registry of choice. Should be relatively easy to do.

mm4tt:
> [ Quoting <notifications@github.com> in "Re: [coredns/coredns] Metric for me..." ]
> > this is problematic and requires a full scale test to see what it adds. That's definitely something we should do. Is there any procedure that describes how to build an experimental image of CoreDNS? [This is something we need to run a test, right?]
> you can use the Makefile.release to build an image with a homebacked coredns. This can be uploaded to a registry of choice. Should be relatively easy to do.

Thanks, @miekg.
Will run the tests and come back with the results here once I have them. 

Update: I've reserved a scale test slot for this Thursday.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Metric for me..." ]
>> [ Quoting <notifications@github.com> in "Re: [coredns/coredns] Metric for me..." ]
>> > this is problematic and requires a full scale test to see what it adds. That's definitely something we should do. Is there any procedure that describes how to build an experimental image of CoreDNS? [This is something we need to run a test, right?]
>> you can use the Makefile.release to build an image with a homebacked coredns. This can be uploaded to a registry of choice. Should be relatively easy to do.
>
>Thanks, @miekg.
>Will run the tests and come back with the results here once I have them.

I get these are important latency metrics. *Technicaly* this does not have to 
happen in coredns, it can also be a probe running separately copying this code.

The problem with the scale test is that coredns query code path isn't exercised 
(at least that was last time), so any increase in memory will only become 
visible on deloy in prod.

mm4tt:
>  The problem with the scale test is that coredns query code path isn't exercised (at least that was last time), so any increase in memory will only become visible on deloy in prod.

I agree that we don't cover coredns extensively and there are many paths not exercised in our scalability tests. And in general some changes won't be visible until they are deployed to prod.

But not this change.
Even if we didn't call the coredns in our scalability tests at all, the changed code would be still executed. The change I made is in the watch (event listener) mechanism. Every time a new endpoints object is created, it will be delivered to coredns over the watch - the new code will be executed and additional memory would be allocated. It doesn't matter if coredns is later queried or not. 

By running the scale tests we'll see the memory increase and it should be exactly the same as if we deployed it in prod.

miekg:
Closing as I still don't want to take on the memory increase.

mm4tt:
Thanks Miek, for all the input. 

We'll rethink this carefully and probably meet with you before another attempts of implementing the SLI. 
While we understand the concerns about CoreDNS memory consumption, we still believe that this SLI will useful from the k8s user perspective. 

@oxddr will be taking over this effort in Q3.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Metric for me..." ]
>Thanks Miek, for all the input.
>
>We'll rethink this carefully and probably meet with you before another attempts of implementing the SLI.
>While we understand the concerns about CoreDNS memory consumption, we still believe that this SLI will useful from the k8s user perspective.
>
>@oxddr will be taking over this effort in Q3.

Thanks for circling back and noted. Please open an issue before the next PR 
attempt, so we can discuss there. To preempt any discussion: adding a time.Time 
will not fly.

There is no question that this metric is useful.

PR #2694, opened 2019-03-14
plugin/debug: add printing memory stats

miekg:
This eases looking at memory stats and doesn't require the use of
prometheus to extract it from the process.


grobie:
No strong objections, but using `while sleep(1); do curl http://localhost:1234/metrics | grep memory_usage_metric; done` sounds simple enough?

miekg:
Is see your point. This does allow more control on what's being printed
though.

grobie:
Given it doesn't require invasive changes to the rest of the code base, I
don't see harm in having helpful debug tools. Cool with me.

I wonder how accurate Go's view of memory consumption turns out to be. I
remember some problems in prometheus.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/debug:..." ]
>Given it doesn't require invasive changes to the rest of the code base, I
>don't see harm in having helpful debug tools. Cool with me.
>
>I wonder how accurate Go's view of memory consumption turns out to be. I
>remember some problems in prometheus.

There are pros and cons with this PR; pro: seems harmless enough; cons: it is 
externally visible and any change will be an API change.

I would be nice to have something that is accessible while running an automated 
tests; but I'm not sure how that would like. I'll hold off merging, but am going 
to use it for debugging; if useful enough I'll merge and we'll see where this 
leads.

/Miek

--
Miek Gieben

miekg:
this can be perfectly done in an external plugin. no need to do this here.

A set of *expert* debug plugin might actually be useful.

PR #2696, opened 2019-03-14
plugin/kubernetes: update serial when records change

chrisohaver:
Update serial when records change.
This is mostly restored functionality that was removed when watch feature was removed (#2549).

none

none

no

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -106,7 +106,7 @@ func newdnsController(kubeClient kubernetes.Interface, opts dnsControlOpts) *dns
 		},
 		&api.Service{},
 		opts.resyncPeriod,
-		cache.ResourceEventHandlerFuncs{},
+		cache.ResourceEventHandlerFuncs{AddFunc: dns.Add, UpdateFunc: dns.Update, DeleteFunc: dns.Delete},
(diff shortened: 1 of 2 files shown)


chrisohaver:
@miekg, PTAL, I missed this when I reviewed #2549.  Part of what the ResourceEventHandlerFuncs were doing was updating the SOA serial when services/endpoints/pods change.  This PR restores that part.

miekg:
this is lot of code being executed for just incrementing a SOA serial number

chrisohaver:
> this is lot of code being executed for just incrementing a SOA serial number

Yes, it only really matters for zone transfers.   We can skip this when `transfer` is not enabled, that way we don't do the extra work unless we need to.  Would that be ok?

miekg:
That would be better. Didn't we have something even simpler before watch
was merged?

johnbelamaric:
Not that worked. We had something simpler at one point but it bumped the
serial way way too often. We only noticed because of watch.

chrisohaver:
> Didn't we have something even simpler before watch was merged?

It was simpler.  Before, we updated the serial for _every_ notification received from the API.  The problem is that majority of updates we receive don't change our records (especially endpoint updates, which is a constant stream of various state changes unrelated to the name or IP of DNS records). This meant that the serial was constantly changing, even while no DNS records were changing. There does not yet exist in the k8s api, a way to subscribe to changes to certain fields only.

miekg:
ok, and there don't seem to be a lot of memory allocation going one.

PR #2698, opened 2019-03-15
pkg/log: fix data race on d

miekg:
Wrap d in a mutex to prevent data race. This makes is slower, but this
is a debugging aid anyway. It's not used normally.

### DIFF
--- a/plugin/pkg/log/log.go
+++ b/plugin/pkg/log/log.go
@@ -13,11 +13,33 @@ import (
 	"io/ioutil"
 	golog "log"
 	"os"
+	"sync"
 	"time"
 )
 
-// D controls whether we should output debug logs. If true, we do.
-var D bool
+// D controls whether we should output debug logs. If true, we do, once set
+// it can not be unset.
+var D = &d{}
+
+type d struct {
+	on bool
+	sync.RWMutex
+}
+
+// Set sets d to true.
+func (d *d) Set() {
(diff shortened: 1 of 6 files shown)


miekg:
https://github.com/coredns/coredns/pull/2694 triggered it

miekg:
I might be using it wrong, because currently this doesn't show up as a
datarace. So hold off on merging this just yet.

miekg:
This is def. a race that needs fixing. The usage pattern I had was completely valid.

miekg:
this def. needs to go in at some point. @varyoo

plugin/pkg/log/log.go
```diff
@@ -13,11 +13,33 @@ import (
...
+
+// Set sets d to true.
+func (d *d) Set() {
+	d.Lock()
+	d.on = true
+	d.Unlock()
+}
+
+// Value return the boolean value of d.
+func (d *d) Value() bool {
```
dilyevsky, line 37: Not sure how much it matters but:
```
type d int32
...
return atomic.LoadInt32((*int32)(&d)) > 0
```
is probably faster.

miekg: Missed your comment. I rather use the abstraction from the go library instead of doing the atomics myself.

dilyevsky: :+1:

Issue #2659, opened 2019-03-07
Add OnRestartFailed to reload, health and prometheus plugins


miekg:
As @fturib wrote:

Thanks to the new "OnRestartFailed" event that was introduced few months ago in Caddy.
We can handle the failed restart (when the Corefile is invalid).

i guess we could have:
```
        c.OnStartup(r.onStartup)
	c.OnStartup(r.wait)
	c.OnRestart(r.onRestart)
         c.OnRestartFailed(r.onStartup)
	c.OnRestartFailed(r.wait)
	c.OnFinalShutdown(r.onFinalShutdown)
```

miekg:
*ready* and *prometheus* are next.

PR #2700, opened 2019-03-15
plugin/metrics: Add a metric to monitor which plugin(s) is(are) enabled

### LINKED ISSUE #2273: plugin/metrics: add a metric to monitor which plugin(s) is(are) enabled
The initial idea is from #1520 , and we want to add a metric to monitor if the plugin `rewrite` is enabled.  @greenpau discussed this with me, and we think it would be better to have a global metric named `pulgin_enabled{server, name}`, so we can easily monitor all enabled plugins. 

I would like to work on this, but I don't really know how to get the list of enabled plugins. Does anyone know how to do that? 

Thanks,
Jiacheng

jiachengxu:
Fix #2273 
#2273 
README.md in the metrics plugin folder.
No.

### DIFF
--- a/plugin/metrics/README.md
+++ b/plugin/metrics/README.md
@@ -19,7 +19,7 @@ The following metrics are exported:
 * `coredns_dns_request_type_count_total{server, zone, type}` - counter of queries per zone and type.
 * `coredns_dns_response_size_bytes{server, zone, proto}` - response size in bytes.
 * `coredns_dns_response_rcode_count_total{server, zone, rcode}` - response per zone and rcode.
-* `coredns_plugin_enabled{server, zone, name}` - enabled pulgin(s) per server and zone.
+* `coredns_plugin_enabled{server, zone, name}` - indicates whether a plugin is enabled on per server and zone basis.
 
 Each counter has a label `zone` which is the zonename used for the request/response.
 
--- a/plugin/metrics/README.md
+++ b/plugin/metrics/README.md
@@ -76,3 +76,4 @@ When reloading, the Prometheus handler is stopped before the new server instance
 If that new server fails to start, then the initial server instance is still available and DNS queries still served,
 but Prometheus handler stays down.
 Prometheus will not reply HTTP request until a successful reload or a complete restart of CoreDNS.
+Only the plugins that register as Handler are visible in `coredns_plugin_enabled{server, zone, name}`. As of today the plugins reload and bind will not be reported.
--- a/plugin/metrics/README.md
+++ b/plugin/metrics/README.md
@@ -19,6 +19,7 @@ The following metrics are exported:
 * `coredns_dns_request_type_count_total{server, zone, type}` - counter of queries per zone and type.
 * `coredns_dns_response_size_bytes{server, zone, proto}` - response size in bytes.
 * `coredns_dns_response_rcode_count_total{server, zone, rcode}` - response per zone and rcode.
+* `coredns_plugin_enabled{server, zone, name}` - enabled pulgin(s) per server and zone.
 
 Each counter has a label `zone` which is the zonename used for the request/response.
 
--- a/plugin/metrics/setup.go
+++ b/plugin/metrics/setup.go
@@ -50,6 +50,12 @@ func setup(c *caddy.Controller) error {
 		})
 		return nil
 	})
+
+	c.OnRestart(func() error {
+		vars.PluginEnabled.Reset()
+		return nil
+	})
+
 	c.OnStartup(func() error {
 		conf := dnsserver.GetConfig(c)
 		plugins := conf.Handlers()
--- a/plugin/metrics/setup.go
+++ b/plugin/metrics/setup.go
@@ -1,12 +1,14 @@
 package metrics
 
 import (
+	"fmt"
 	"net"
 	"runtime"
 
 	"github.com/coredns/coredns/core/dnsserver"
 	"github.com/coredns/coredns/coremain"
 	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/metrics/vars"
 	clog "github.com/coredns/coredns/plugin/pkg/log"
 	"github.com/coredns/coredns/plugin/pkg/uniq"
 
@@ -49,7 +51,15 @@ func setup(c *caddy.Controller) error {
 		})
 		return nil
 	})
+	c.OnStartup(func() error {
+		plugins := dnsserver.GetConfig(c).Handlers()
+		for _, p := range plugins {
+			vars.PluginEnabled.WithLabelValues(p.Name()).Set(1)
+			fmt.Println(p.Name())
+		}
+		return nil
 
+	})
 	c.OnRestart(m.OnRestart)
 	c.OnFinalShutdown(m.OnFinalShutdown)
 
--- a/plugin/metrics/setup.go
+++ b/plugin/metrics/setup.go
@@ -1,7 +1,6 @@
 package metrics
 
 import (
-	"fmt"
 	"net"
 	"runtime"
 
@@ -52,10 +51,17 @@ func setup(c *caddy.Controller) error {
 		return nil
 	})
 	c.OnStartup(func() error {
-		plugins := dnsserver.GetConfig(c).Handlers()
-		for _, p := range plugins {
-			vars.PluginEnabled.WithLabelValues(p.Name()).Set(1)
-			fmt.Println(p.Name())
+		conf := dnsserver.GetConfig(c)
+		plugins := conf.Handlers()
+		for _, h := range conf.ListenHosts {
+			addr, err := net.ResolveTCPAddr("tcp", net.JoinHostPort(h, conf.Port))
+			if err != nil {
+				return err
+			}
+			addrstr := conf.Transport + "://" + addr.String()
+			for _, p := range plugins {
+				vars.PluginEnabled.WithLabelValues(addrstr, conf.Zone, p.Name()).Set(1)
+			}
 		}
 		return nil
 
--- a/plugin/metrics/setup.go
+++ b/plugin/metrics/setup.go
@@ -60,11 +60,7 @@ func setup(c *caddy.Controller) error {
 		conf := dnsserver.GetConfig(c)
 		plugins := conf.Handlers()
 		for _, h := range conf.ListenHosts {
-			addr, err := net.ResolveTCPAddr("tcp", net.JoinHostPort(h, conf.Port))
-			if err != nil {
-				return err
-			}
-			addrstr := conf.Transport + "://" + addr.String()
+			addrstr := conf.Transport + "://" + net.JoinHostPort(h, conf.Port)
 			for _, p := range plugins {
 				vars.PluginEnabled.WithLabelValues(addrstr, conf.Zone, p.Name()).Set(1)
 			}
--- a/plugin/metrics/vars/vars.go
+++ b/plugin/metrics/vars/vars.go
@@ -68,7 +68,6 @@ var (
 
 	PluginEnabled = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: plugin.Namespace,
-		Subsystem: subsystem,
 		Name:      "plugin_enabled",
 		Help:      "A metric that monitors which plugin(s) is(are) enabled.",
 	}, []string{"server", "zone", "name"})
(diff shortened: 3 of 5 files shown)


jiachengxu:
ping @miekg @greenpau

greenpau:
@jiachengxu , please squash in a single commit. It is easier to review it.

greenpau:
@jiachengxu , did you add relevant notes to `README.md` of the plugin?

jiachengxu:
@greenpau 

> please squash in a single commit. It is easier to review it. 

I think it doesn't matter how many commits are in the PR because you can click the `Files Changed` tab under the PR's title, then it will show you all the changes in this whole PR, and for the person who will merge it into the master, he(she) will squash all commits into one commit.

> did you add relevant notes to README.md of the plugin?

Yes, you will see it after you click the `Files Changed` tab.

plugin/metrics/README.md
```diff
@@ -19,6 +19,7 @@ The following metrics are exported:
 * `coredns_dns_request_type_count_total{server, zone, type}` - counter of queries per zone and type.
 * `coredns_dns_response_size_bytes{server, zone, proto}` - response size in bytes.
 * `coredns_dns_response_rcode_count_total{server, zone, rcode}` - response per zone and rcode.
+* `coredns_plugin_enabled{server, zone, name}` - enabled pulgin(s) per server and zone.
```
greenpau: @jiachengxu , "Indicates whether a plugin is enabled on per server and zone basis".

plugin/metrics/vars/vars.go
```diff
@@ -65,6 +65,12 @@ var (
 		Name:      "panic_count_total",
 		Help:      "A metrics that counts the number of panics.",
 	})
+
+	PluginEnabled = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: plugin.Namespace,
+		Name:      "plugin_enabled",
+		Help:      "A metric that monitors which plugin(s) is(are) enabled.",
```
greenpau: @jiachengxu , "Indicates whether a plugin is enabled on per server and zone basis".

greenpau (changes requested):
@jiachengxu , looks good with minor edits.

greenpau:
adding @fturib

plugin/metrics/README.md
fturib: The plugins are enabled by "ServerBloc", which means per Key at top of each of Stanza.
This Key or ServerBlocKey is available through the line:
```
```

Why not using this definition - which is used by almost all plugins ?

jiachengxu: This is much better than my previous one!

plugin/metrics/vars/vars.go
jiachengxu: Done

fturib (commented):
My main comment is on the tuple (server, zone, name).
Why not use the same indicator as all others (server, name) - where server includes the server and the zone information (in fact the ServerBloc key)

plugin/metrics/README.md
jiachengxu: If I understood correctly, you are asking why I need the `zone` label for this metric, I will give an example here:
Here is a `Corefile`:
```
coredns.io:1053 {
    prometheus
    errors
    log
}

example.io:1053 {
    errors
    log
}

coredns.io:53 {
    prometheus
    whoami
    log
}

example.io:53 {
    prometheus
    log
    errors
}
```
In this case, I can get information about which plugins are enabled not only on server but also on the zone:
```
coredns_plugin_enabled{name="errors",server="dns://:1053",zone="coredns.io."} 1
coredns_plugin_enabled{name="errors",server="dns://:53",zone="example.io."} 1
coredns_plugin_enabled{name="log",server="dns://:1053",zone="coredns.io."} 1
coredns_plugin_enabled{name="log",server="dns://:53",zone="coredns.io."} 1
coredns_plugin_enabled{name="log",server="dns://:53",zone="example.io."} 1
coredns_plugin_enabled{name="prometheus",server="dns://:1053",zone="coredns.io."} 1
coredns_plugin_enabled{name="prometheus",server="dns://:53",zone="coredns.io."} 1
coredns_plugin_enabled{name="prometheus",server="dns://:53",zone="example.io."} 1
coredns_plugin_enabled{name="whoami",server="dns://:53",zone="coredns.io."} 1
```

That's why I did this, I just want to get more detailed information from this metric.

But in this case it also has a drawback:
If you want to monitor the whole server, you need to enable `metrics` plugin for every zone of this server(In the above case, we can not see the enabled plugins of the zone="example.io", server="dns://:1053".

To be honest, I don't know which one is better, so maybe you have some ideas?

fturib: @jiachengxu : my bad !  I had not the correct view of what is in the "server" part (I though was including the zone, but no). 
So I agree with your view and then the way you implemented here.

plugin/metrics/setup.go
```diff
@@ -49,7 +50,22 @@ func setup(c *caddy.Controller) error {
...
+			}
+			addrstr := conf.Transport + "://" + addr.String()
+			for _, p := range plugins {
+				vars.PluginEnabled.WithLabelValues(addrstr, conf.Zone, p.Name()).Set(1)
+			}
+		}
+		return nil
 
+	})
 	c.OnRestart(m.OnRestart)
```
fturib, line 71: Did you try with the `reload` plugin ?
And then you change Corefile => CoreDNS will be reloaded .. with maybe different set of plugins ...
Will your metrics show this new set of plugins ?

fturib: I think you should take advantage of the "onRestart" event. But I am not sure how you can "clean" the metric values of the plugins that were loaded on firt time.

jiachengxu: > Did you try with the reload plugin ?
And then you change Corefile => CoreDNS will be reloaded .. with maybe different set of plugins ...
Will your metrics show this new set of plugins ?

I tried just now, and it doesn't work:
1. Every time I added a new plugin, after CoreDNS was reloaded, it will show the new set of plugins, but if I removed some plugins, it still showed them.
2. This mertic can not monitor if the `reload` plugin is enabled, which means I can not see `reload` even if I have `metrics` plugin already enabled... I don't really know why.

jiachengxu: > I think you should take advantage of the "onRestart" event. But I am not sure how you can "clean" the metric values of the plugins that were loaded on firt time.

Good point, I will try to play with that.

fturib: > This mertic can not monitor if the reload plugin is enabled, which means I can not see reload even if I have metrics plugin already enabled... I don't really know why.

I guess that `reload` does not register as a `handler` because it does not act on the DNS request: there is no `ServerDNS` function in the `reload` plugin.

fturib: > Every time I added a new plugin, after CoreDNS was reloaded, it will show the new set of plugins, but if I removed some plugins, it still showed them.

I guess you need to find a way to 'clean" the metrics variable on each reload, and re-register the whole new list of plugins.
=> I see a function `Reset()` on a `metricsMap` that seems to be a part of any metrics ...

jiachengxu: > I guess that reload does not register as a handler because it does not act on the DNS request: there is no ServerDNS function in the reload plugin.

So I just ignore the `reload` plugin?

> I see a function Reset() on a metricsMap that seems to be a part of any metrics ...

Actually, this works fine, I updated the `OnRestart` callback.

fturib: > So I just ignore the reload plugin?

I think we should fix that on a separated PR for the plugins that does not register a Handler.
So far, I think : `reload` and `bind`.
Best, here, would be to add a section `Bug` at the end of this readme, adding something like:
"Only the plugins that register as Handler are visible in this metrics. As of today the plugins `reload` and `bind` will not be reported" (or something similar).

> Actually, this works fine, I updated the OnRestart callback.

Great !!!
 ... but. I just remind now, that there will be an issue if the Restart fails. Which means if the new Corefile is invalid. In that case, your var is Reset, and CoreDNS will continue running w/o any plugin reported. (because onRestart is called ONLY at starting of CoreDNS).

There is a way to fix that : use the event "onRestartFailed" and let him do exactly the same thing as "onRestart".  See usage of event described in Caddy:
https://github.com/mholt/caddy/blob/f6e6a6be0446a8455001d6202032b20b64b37569/caddy.go#L110-L116


NOTE: [related issue to use onRestartFailed to ensure proper config if the Corefile is invalid at reload](https://github.com/coredns/coredns/issues/2659).

plugin/metrics/setup.go
```diff
@@ -50,6 +51,27 @@ func setup(c *caddy.Controller) error {
...
+	c.OnRestart(func() error {
+		vars.PluginEnabled.Reset()
+		return nil
+	})
+
+	c.OnStartup(func() error {
+		conf := dnsserver.GetConfig(c)
+		plugins := conf.Handlers()
+		for _, h := range conf.ListenHosts {
+			addr, err := net.ResolveTCPAddr("tcp", net.JoinHostPort(h, conf.Port))
```
miekg: Don't think you need `ResolveTCPAddr` here.

plugin/metrics/setup.go
```diff
@@ -50,6 +51,27 @@ func setup(c *caddy.Controller) error {
...
 	})
 
+	c.OnRestart(func() error {
+		vars.PluginEnabled.Reset()
+		return nil
+	})
+
+	c.OnStartup(func() error {
+		conf := dnsserver.GetConfig(c)
+		plugins := conf.Handlers()
```
miekg, line 61: Slightly shorter: `conf := dnsserver.GetConfig(c).Handlers()`

plugin/metrics/setup.go
jiachengxu: > I think we should fix that on a separated PR for the plugins that does not register a Handler.

I am willing to work on this, and I will create an issue for this, any ideas about how to fix this?

> Best, here, would be to add a section Bug at the end of this readme, adding something like:
"Only the plugins that register as Handler are visible in this metrics. As of today the plugins reload and bind will not be reported" (or something similar).

Great, I will do this.

> There is a way to fix that : use the event "onRestartFailed" and let him do exactly the same thing as "onRestart". See usage of event described in Caddy:

Did you mean that the "onRestartFailed" should do the same thing as **`onStart`** since the `onRestart` just resets the metric. We shouldn't reset the metric if it restarts failed, right?

plugin/metrics/setup.go
jiachengxu: So what should I do here then? Can you provide more hints, please?

plugin/metrics/setup.go
jiachengxu: If I don't need a for loop to iterate the hosts to `resolveTCPAddr` later, then I will not need this.

plugin/metrics/setup.go
fturib: Sorry, YES, I meant `onRestartFailed` should do the same thing as `onStart`.

jiachengxu: Hi @fturib, I tried to do the same thing in `onRestartFailed` as `onStart`, but it doesn't work, I think it is beccause:
 [If that new server fails to start, then the initial server instance is still available and DNS queries still served, **but Prometheus handler stays down. Prometheus will not reply HTTP request until a successful reload or a complete restart of CoreDNS.**](https://github.com/coredns/coredns/blob/master/plugin/metrics/README.md#bugs)

This has been already reported in the `metrics` README.md, so I think we should fix this at first.

fturib: You are right. The issue opened is here : #2659.

jiachengxu:
Hi @greenpau @fturib @miekg ,

What else should I do to merge this PR?

greenpau (approved):

fturib:
LGTM.
However there is a pending remark to be solved here:
> Don't think you need ResolveTCPAddr here.

@miekg : can you provide more hint on what you think should be used ?

miekg:
> @miekg : can you provide more hint on what you think should be used ?

nothing, this can be removed

plugin/metrics/setup.go
jiachengxu: Done

jiachengxu:
> > @miekg : can you provide more hint on what you think should be used ?
> 
> nothing, this can be removed

@miekg I have updated the PR, PTAL.

fturib (approved):

Issue #2387, opened 2018-12-11
plugin/proxy: gRPC BackoffMaxDelay


inigohu:
Hi,

I need the possibility to add **BackoffMaxDelay** parameter to grpc protocol in the Proxy plugin.

- Version: 1.2.6
- Corefile:

```
. {
    health :8081
    debug
    log
    proxy . 10.5.0.6:8080 {
    	protocol grpc insecure
	}
}
```

The reason is that I want to test my backend service that is being used by the Proxy plugin. To make an e2e test, I start my backend service and make a request to coreDNS.  

The grpc client backoff delay is exponential, and the maxDelay is 2 minutes. The coreDNS proxy has a 16 seconds of *tryDuration*.

This means that if the coreDNS is running, but not my backend service, the grpc client backoff delay will increase until 2 minutes. After that, if I run my test (start backend service and make a request), I get an error because the grpc connection has not been made yet (grpc backoff delay is bigger than *tryDuration*).

Would be great to have an extra parameter for **BackoffMaxDelay**

```
{
    ...
    proxy . 10.5.0.6:8080 {
    	protocol grpc insecure 1s
	}
    ...
}
```

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin/proxy: gRP..." ]
>Hi,
>
>I need the possibility to add **BackoffMaxDelay** parameter to grpc protocol in the Proxy plugin.

i want to actually retire this plugin; one on the things holding that back is 
that the forward plugin doesnt support gRPC.

To not answer your question: I would like to see the grpc support moved to grpc 
and then think about your feature request.

/plugin: proxy
/label: enhancement

PR #2701, opened 2019-03-15
plugin/grpc: configure BackoffMaxDelay parameter

inigohu:
Possibility to configure BackoffMaxDelay.
#2387


miekg:
Is this needed a user knob or can an sane default be set for DNS traffic?

inigohu:
I wouldn't configure it by default, this should be a customization. In my case I need to reduce backoff delay for testing purpose only.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/grpc: ..." ]
>I wouldn't configure it by default, this should be a customization. In my case I need to reduce backoff delay for testing purpose only.

then it should probably not be a user visibile knob.

inigohu:
> then it should probably not be a user visibile knob.

How can I set up this option without a visible knob?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/grpc: ..." ]
>> then it should probably not be a user visibile knob.
>
>How can I set up this option without a visible knob?

If you need it in testing you can just add to the testing code.

If you need it for *your* (external?) tests, that's a different problem. That's even more 
strongly points that this is a knob that is only useful for a very small subset 
of people and should not be there.

Maybe an exported variable you can set in code that you include in your testing 
setup or some such?

inigohu:
> If you need it in testing you can just add to the testing code. 
> 
> If you need it for *your* (external?) tests, that's a different problem. That's even more strongly points that this is a knob that is only useful for a very small subset of people and should not be there. 
> 
> Maybe an exported variable you can set in code that you include in your testing setup or some such?

I'm running my upstream service and CoreDNS inside a docker-compose to make integration tests. I can't set any external variable or add any configuration in my tests that modifies CoreDNS.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/grpc: ..." ]
>> If you need it in testing you can just add to the testing code.
>>
>> If you need it for *your* (external?) tests, that's a different problem. That's even more strongly points that this is a knob that is only useful for a very small subset of people and should not be there.
>>
>> Maybe an exported variable you can set in code that you include in your testing setup or some such?
>
>I'm running my upstream service and CoreDNS inside a docker-compose to make integration tests. I can't set any external variable or add any configuration in my tests that modifies CoreDNS.

Sure you can, except it requires some extra work from your side.

One of principles I try to apply to coredns (and miekg/dns) is that there should 
not be a need for knobs to users can twiddle. If such a requirement exists we 
should automate it as much as possible. And only as a last resort expose 
something to a user.

inigohu:
> One of principles I try to apply to coredns (and miekg/dns) is that there should not be a need for knobs to users can twiddle. If such a requirement exists we should automate it as much as possible. And only as a last resort expose something to a user.

I agree with those principles, but I think that if someone wants to make an end-to-end test of the backend service, should control de workflow of the test.

I've drawn the architecture of my test environment to explain my case in order to find a solution or finally close this PR.

Untitled Diagram
Backend depends on storage, so the first container to start is storage container and then, backend and CoreDNS containers.


Backend service needs to make some checks and builds before server and tests starts. Meantime, CoreDNS is already running and increasing the gRPC backoff delay connection until backend is up and ready to listen requests.

When the backend tests try to make DNS queries to CoreDNS will get an error because the gRPC backoff delay will be bigger than the DNS query timeout.

From the backend container I cannot control the CoreDNS  gRPC plugin status either, if is already connected or not.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/grpc: ..." ]
>I've drawn the architecture of my test environment to explain my case in order to find a solution or finally close this PR.
>
>Untitled Diagram
>Backend depends on storage, so the first container to start is storage container and then, backend and CoreDNS containers.
>
>
>Backend service needs to make some checks and builds before server and tests starts. Meantime, CoreDNS is already running and increasing the gRPC backoff delay connection until backend is up and ready to listen requests.
>
>When the backend tests try to make DNS queries to CoreDNS will get an error because the gRPC backoff delay will be bigger than the DNS query timeout.
>
>From the backend container I cannot control the CoreDNS  gRPC plugin status either, if is already connected or not.

This is a problem in itself and probably not only in your tests. Ideally the 
gRPC would try one-offs in-between the backoff delay to see if a upstream is 
respondning again.

A max of 5001 milleseconds of the backend delay might also work as a fix maximu, 
because anything bigger will not work anyway.

inigohu:
> This is a problem in itself and probably not only in your tests. Ideally the gRPC would try one-offs in-between the backoff delay to see if a upstream is respondning again. 

> A max of 5001 milleseconds of the backend delay might also work as a fix maximu, because anything bigger will not work anyway.

Ok. I think the best thing will be to make a fork with those custom features.
Thanks

PR #2704, opened 2019-03-17
server: cancel context after 501 milliseconds

miekg:
The context in each request is canceled after 501 ms. Plugins that check
ctx.Done() can abort (return) from whatever they are executing because
the waiting client will never see the reply.

The irony of this approach is that the cancelation is done by starting
another go-routine. It would be cooler if we would only start that
routine after x seconds, but doing that requires a go routine.

I don't believe Go provides out-of-the-box feature to look at the
packet's timestamp. Even if it did it requires each plugin to look at
it, the context.WithTimeout is somewhat cleaner.

### DIFF
--- a/core/dnsserver/cancel_test.go
+++ b/core/dnsserver/cancel_test.go
@@ -19,13 +19,11 @@ func (s sleepPlugin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.
 	m := new(dns.Msg)
 	m.SetReply(r)
 	for {
-		select {
-		case <-ctx.Done():
-			// use BadTime to return something time related
-			m.Rcode = dns.RcodeBadTime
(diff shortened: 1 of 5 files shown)


miekg:
The code from `sleepPlugin` can be used in *forward*  as a first step and to test (I can do this here in a separate commit).

It may be useful to make a generic 'check' function ala *Match* that checks the context in each plugin to prevent useless work.

johnbelamaric:
Why isn't this done in a plugin?

johnbelamaric:
This is also "gone in 1/2 second", not "gone in 6 seconds".

miekg:
heh, lol. Indeed.

And good point about doing this in a plugin, as we inherit the context anyway, further down the chain.

miekg:
See https://github.com/coredns/coredns/pull/2711

Issue #2624, opened 2019-02-27
plugin/forward:  maxes out CPU (TLS connection negotiating)


WinstonPrivacy:
We've noticed an occasional problem with using CoreDNS on a low-end ARM64 board (1Gb RAM, dual core 1.0 Ghz CPU). 

If another process temporarily maxes out CPU, CoreDNS CPU spikes to 80%+ and stays there indefinitely, even after the other process has been terminated. 

The issue only appears if there is a lot of DNS activity during the period of peak CPU activity. My guess is that the DNS queries pile up but CoreDNS is unable to ever catch up with them for some reason. As an example, here is a screenshot of CoreDNS log output taken a few minutes after terminating a high-CPU process. This will continue indefinitely and in fact, the response times almost seem to get worse.

The problem goes away once CoreDNS is restarted. Happy to try and help diagnose this but I'm not sure where to start.

![coredns failure](https://user-images.githubusercontent.com/36650369/53460250-15b46400-3a02-11e9-9e4b-40756c70221b.png)

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] CoreDNS maxes out..." ]
>We've noticed an occasional problem with using CoreDNS on a low-end ARM64 board (1Gb RAM, dual core 1.0 Ghz CPU).
>
>If another process temporarily maxes out CPU, CoreDNS CPU spikes to 80%+ and stays there indefinitely, even after the other process has been terminated.
>
>The issue only appears if there is a lot of DNS activity during the period of peak CPU activity. My guess is that the DNS queries pile up but CoreDNS is unable to ever catch up with them for some reason. As an example, here is a screenshot of CoreDNS log output taken a few minutes after terminating a high-CPU process. This will continue indefinitely and in fact, the response times almost seem to get worse.
>
>The problem goes away once CoreDNS is restarted. Happy to try and help diagnose this but I'm not sure where to start.
>
>![coredns failure](https://user-images.githubusercontent.com/36650369/53460250-15b46400-3a02-11e9-9e4b-40756c70221b.png)

What version are you running and with what Corefile?

Do you have some metrics (enable prometheus plugin), like number of goroutines, memory and cpu usage. A
pprof (enable pprof plugin) could help as well.

johnbelamaric:
At first glance this tells me we need to put a limit on our inbound request
queue and start dropping packets. There was another issue that seemed to
imply the same thing.

miekg:
[...]

PR #2706, opened 2019-03-18
plugin/throttle - limit impact burst of queries

fturib:

This plugin offers:
* a limitation of simultaneous inflight queries (drop any out-of-limit incoming queries)
* metrics for investigation (incoming, in process, dropped, etc ...)

So far, the test shows that the limitation of inflight queries is useful to prevent OOM crash in case of burst. However, that need to provide in Corefile a MAX-INFLIGHT, that is somehow related to the MAX-MEMORY of the deployment.

NOTE: Investigation of initial issue #2593 shows that this plugin is efficient for preventing OOM when incoming queries is getting high.

[request spike creates memory spike](https://github.com/coredns/coredns/issues/2593)

README included.

No.

### DIFF
--- a/plugin/throttle/README.md
+++ b/plugin/throttle/README.md
@@ -2,7 +2,7 @@
 
 ## Name
 
-*throttle* - limits the number of inflight queries.
+*throttle* - limits the maximum number of inflight queries.
 
 ## Description
 
--- a/plugin/throttle/README.md
+++ b/plugin/throttle/README.md
@@ -1,0 +1,57 @@
+# throttle
+
+## Name
+
+*throttle* - limits the number of inflight queries.
+
+## Description
+
+This plugin limit the max inflight simultaneous queries.
+While the limit is reached, any new queries received are dropped.
+
+
+The _throttle_ plugin prevents CoreDNS from consuming an unbound amount of memory
+when it receives a burst of incoming queries.
+
+One will need to tune the **MAX-INFLIGHT** with the memory allowed for this pod or process
+
+NOTE: `throttle` acts only on the server block is which it is defined.
+To be most effective, it should be enabled on all DNS server blocks.
+
+## Syntax
+
+~~~ txt
+throttle [MAX-INFLIGHT]
+~~~
+
+* The plugin check the number of in-process queries for each new incoming query.
+it it that number is > **MAX-INFLIGHT** then the query is immediatly dropped.
+* When **TIMEOUT** is defined (and non nul) this plugin verify i
+
+## Examples
+
+Check with the default intervals:
+
+~~~ corefile
+. {
+    throttle 1000
+}
+~~~
+
+Check that no more than 1000 queries are served in same time
+
+
+## Metrics
+
+If monitoring is enabled (via the *prometheus* directive) then the following metrics are exported:
+
+* `coredns_throttle_seen_queries{server, kind}` - Total queries seen by the throttle, per `server`.
+
+   Query `kind` can be either:
+   - `incoming` : total queries getting into this plugin chain
+   - `dropped`: total queries dropped because hit the limit **MAX-INFLIGHT**
+   - `served`: total queries served by throttle
+
+
+* `coredns_throttle_flight_queries{server}` - Total simultaneous queries currently processed by the throttle, per `server`.
+
--- a/plugin/throttle/setup_test.go
+++ b/plugin/throttle/setup_test.go
@@ -1,0 +1,30 @@
+package throttle
+
+import (
+	"testing"
+
+	"github.com/mholt/caddy"
+)
+
+func TestSetupThrottle(t *testing.T) {
+	c := caddy.NewTestController("dns", `throttle`)
+	if err := setup(c); err == nil {
+		t.Fatal("Expected errors, but got none")
+	}
+
+	c = caddy.NewTestController("dns", `throttle 10`)
+	if err := setup(c); err != nil {
+		t.Fatalf("Expected no errors, but got: %v", err)
+	}
+
+	c = caddy.NewTestController("dns", `throttle 10 whatever`)
+	if err := setup(c); err == nil {
+		t.Fatal("Expected errors, but got none")
+	}
+
+	c = caddy.NewTestController("dns", `throttle whatever`)
+	if err := setup(c); err == nil {
+		t.Fatal("Expected errors, but got none")
+	}
+
+}
--- a/plugin/throttle/throttle.go
+++ b/plugin/throttle/throttle.go
@@ -1,0 +1,68 @@
+package throttle
+
+import (
+	"context"
+	"sync/atomic"
+
+	"github.com/coredns/coredns/plugin/metrics"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/miekg/dns"
+)
+
+// Throttle holds the count of queries currently in process and the upper limit expected
+type Throttle struct {
+	workerCount int64
+	workerLimit int64
+	Next        plugin.Handler
+}
+
+// ServeDNS implements the plugin.Handler interface.
+func (t *Throttle) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+
+	count := atomic.AddInt64(&t.workerCount, 1)
+	defer atomic.AddInt64(&t.workerCount, -1)
+
+	server := metrics.WithServer(ctx)
+	throttleCount.WithLabelValues(server, incoming).Inc()
+	flightCount.WithLabelValues(server).Set(float64(count))
+
+	// just drop the queries when current inflight bucket is full
+	if count > t.workerLimit {
+		w.Close()
+		// just return success w/o writing in the writer -> it is a drop
+		// it is in purpose to not return a response as it would keep the goroutine alive the time of writing
+		throttleCount.WithLabelValues(server, dropped).Inc()
+		return dns.RcodeSuccess, nil
+	}
+
+	code, err := plugin.NextOrFailure(t.Name(), t.Next, ctx, w, r)
+	throttleCount.WithLabelValues(server, served).Inc()
+	return code, err
+}
+
+// Name implements the Handler interface.
+func (t Throttle) Name() string { return throttle }
+
+var (
+	throttleCount = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "throttle",
+		Name:      "seen_queries",
+		Help:      "Total queries seen by throttle mechanism.",
+	}, []string{"server", "type"})
+	flightCount = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "throttle",
+		Name:      "inflight_queries",
+		Help:      "The number of simultaneous queries inside throttle mechanism.",
+	}, []string{"server"})
+)
+
+const (
+	incoming = "incoming"
+	served   = "served"
+	dropped  = "dropped"
+)
--- a/plugin/throttle/throttle.go
+++ b/plugin/throttle/throttle.go
@@ -22,8 +22,8 @@ type Throttle struct {
 // ServeDNS implements the plugin.Handler interface.
 func (t *Throttle) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 
-	count := atomic.AddInt64(&t.workerCount, 1)
-	defer atomic.AddInt64(&t.workerCount, -1)
+	count := atomic.AddInt64(&(t.workerCount), 1)
+	defer atomic.AddInt64(&(t.workerCount), -1)
 
 	server := metrics.WithServer(ctx)
 	throttleCount.WithLabelValues(server, incoming).Inc()
@@ -44,7 +44,7 @@ func (t *Throttle) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 }
 
 // Name implements the Handler interface.
-func (t Throttle) Name() string { return throttle }
+func (t *Throttle) Name() string { return throttle }
 
 var (
 	throttleCount = prometheus.NewGaugeVec(prometheus.GaugeOpts{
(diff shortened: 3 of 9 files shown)


coremain/version.go
```diff
@@ -2,7 +2,7 @@ package coremain
 
 // Various CoreDNS constants.
 const (
-	CoreVersion = "1.4.0"
+	CoreVersion = "1.4.0-ftur-throttle"
```
rajansandeep: Maybe need to revert this change.

fturib: yes. Thanks :)

plugin/throttle/README.md
```diff
@@ -0,0 +1,70 @@
+# reload
```
chrisohaver:
```suggestion
# throttle
```

plugin/throttle/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Name
+
+*throttle* - limit the incoming traffic of DNS queries
```
chrisohaver: The **MAX-INFLIGHT** parameter does this, but the **TIMEOUT** does not do this.

Trying to describe both clearly and succinctly is tough.. perhaps...

```suggestion
*throttle* - limits/reduces the number of inflight queries.
```

~I was thinking that **TIMEOUT** would be ineffective anyways, since the bulk of the spike issue was with the number of inflight queries in the process of writing to the client.  Should this option be removed if it doesn't really help?  (it would reduce the code significantly)~

Ah never-mind. See #2711... looks like timeout will be a separate plugin.
So it should be removed from here.
Also, it's obvious to me now that a timeout should help out, since writing to the client is expensive it makes sense to skip this if the client is probably not listening for an answer anymore.

plugin/throttle/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Description
+
+This plugin limit the max inflight simultaneous queries.
```
chrisohaver, line 9:
```suggestion
This plugin limits the maximum number of inflight queries.
```

plugin/throttle/README.md
```diff
@@ -0,0 +1,70 @@
...
+Any extra query incoming is just dropped.
```
chrisohaver:
```suggestion
While the limit is reached, any new queries received are dropped.
```

plugin/throttle/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+The target is to ensure that CoreDNS survive a burst of incoming queries
```
chrisohaver:
```suggestion
The _throttle_ plugin prevents CoreDNS from consuming an unbound amount of memory when it receives a burst of incoming queries.
```

plugin/throttle/README.md
```diff
@@ -0,0 +1,70 @@
...
+without crashing or being deleted because of lack of memory.
```
chrisohaver:
```suggestion
```

plugin/throttle/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+One will need to tune the **MAX-INFLIGHT** with the memory allowed for this pod or process
+
+NOTE: `throttle` is acting ONLY on the server it is turned on.
```
chrisohaver:
```suggestion
NOTE: `throttle` acts only on the server block in which it is defined.
```

plugin/throttle/README.md
```diff
@@ -0,0 +1,70 @@
...
+To be effective, it should be enabled on each of DNS server defined.
```
chrisohaver:
```suggestion
To be most effective, it should be enabled on all DNS server blocks.
```

plugin/throttle/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+If monitoring is enabled (via the *prometheus* directive) then the following metrics are exported:
+
+* `coredns_throttle_seen_queries{server, kind}` - Total queries seen by the throttle, per `server`.
+
+   Query `kind` can be either:
+   - `incoming` : total queries getting into this plugin chain
+   - `overflight-dropped`: total queries dropped because hit the limit **MAX-INFLIGHT**
+   - `timedout-dropped`: total queries dropped because hit the limit **TIMEOUT**
+   - `served`: total queries served by throttle
```
chrisohaver, line 53: Is this metric for total queries received, or for total queries not dropped.

plugin/throttle/setup_test.go
```diff
@@ -0,0 +1,35 @@
+package throttle
+
+import (
+	"testing"
+
+	"github.com/mholt/caddy"
+)
+
+func TestSetupReload(t *testing.T) {
```
chrisohaver:
```suggestion
func TestSetupTrottle(t *testing.T) {
```

plugin/throttle/throttle.go
```diff
@@ -0,0 +1,124 @@
...
+	workerLimit  int64
+	writingCount int64
+	timeout      time.Duration
+	Next         plugin.Handler
+}
+
+// ServeDNS implements the plugin.Handler interface.
+func (t *Throttle) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+
+	droptIt := func(server string, dropttype string, w dns.ResponseWriter) (int, error) {
```
chrisohaver: Are "droptIt" and "dropttype" variable name typos? e.g.
```suggestion
	dropIt := func(server string, droptype string, w dns.ResponseWriter) (int, error) {
```

Why is this function a variable, and not a normal function?

chrisohaver (dismissed):
Some tweaks to README, and questions.

plugin/throttle/README.md
chrisohaver: Ah never-mind.  See  #2711... looks like timeout will be a separate plugin.

So it should be removed from here.

fturib: yes I removed it.

plugin/throttle/throttle.go
fturib: your are right: typo.... maybe glasses missing.

plugin/throttle/README.md
fturib: hum.. I forgot one word.

miekg:
We still don't have an actual root cause other than that go-routines pile up and what's causing that

Also this plugin assumes all further processing is slow which is not true. It might be as fast to allow *cache* or *file* to answer, instead of dropping the entire request.
I also monopolizes the entire server if one plugin is slow giving the others no change to reply.
*And* it adds a magic knob - which is great for us, 'cause we can now blame our users for "not holding it right"

chrisohaver:
> We still don't have an actual root cause other than that go-routines pile up and what's causing that

Given that that the memory issues have been replicated with **zero** plugins enabled, that hints that it may be caused by a bottleneck writing the response to the client. 

I vaguely recall, though it was more than a year ago last time i looked, when doing cpu pprof analysis of CoreDNS, most time was spent writing UDP responses to clients.  IMO, it's worth trying to replicate the memory spike issue with pprof on, and look at the time spend per routine stats.

willwp:
ARM64v8 A53, dual core CPU 1.0 GHz, 1 GB RAM, DDR4.   Tested out with throttle set at 1000

willwp:
ARM64v8 A53, dual core CPU 1.0 GHz, 1 GB RAM, DDR4.  Smoke Tested with throttle set at 1000
Built with go v1.12.  Noticed that memory demand stayed high, and not all of my dns queries in my test lab were not being responded to after I slammed with dnsperf:
```
dnsperf -s x.x.x.x -p 53 -d queryfile-example-10million-201202 -l 300 -c 30 -Q 2000
```
```
.:53 {                                                       
       cache 3600                                        
       forward . tls://1.1.1.1 tls://1.0.0.1 {
               tls_servername cloudflare-dns.com
               health_check 5s                  
               max_fails 1                      
               expire 90s                       
       }                                        
       alternate SERVFAIL . tls://9.9.9.9 tls://149.112.112.112 {
                tls_servername dns.quad9.net                     
                health_check 5s                                  
                max_fails 1                                      
                expire 90s                                       
       }                                    
       throttle 1000              
       prometheus 0.0.0.0:9090
       pprof 0.0.0.0:6053     
}                             
.:54 {                        
       cache 3600                                      
       forward . tls://1.1.1.1 tls://1.0.0.1 {         
               tls_servername cloudflare-dns.com       
               health_check 5s                         
               max_fails 1                             
               expire 90s                       
        }                                       
        alternate SERVFAIL . tls://9.9.9.9 tls://149.112.112.112 {
               tls_servername dns.quad9.net                       
               health_check 5s                                    
               max_fails 1                                        
               expire 90s                                         
        }                                                         
        throttle 1000                          
}                
```
Go_memstats_throttle
go_routines_throttle
dnsRequests_throttle

[pprof001_throttle.svg.zip](https://github.com/coredns/coredns/files/3009751/pprof001_throttle.svg.zip)

```
(pprof) top
Showing nodes accounting for 21.20s, 50.07% of 42.34s total
Dropped 486 nodes (cum <= 0.21s)
Showing top 10 nodes out of 145
      flat  flat%   sum%        cum   cum%
     3.22s  7.61%  7.61%     16.08s 37.98%  math/big.nat.divLarge /usr/local/go/src/math/big/nat.go
     2.98s  7.04% 14.64%      2.98s  7.04%  math/big.addMulVVW /usr/local/go/src/math/big/arith_arm64.s
     2.73s  6.45% 21.09%      3.41s  8.05%  runtime.findObject /usr/local/go/src/runtime/mbitmap.go
     2.46s  5.81% 26.90%      2.46s  5.81%  math/big.mulAddVWW /usr/local/go/src/math/big/arith_arm64.s
     2.33s  5.50% 32.40%      2.36s  5.57%  math/big.divWW_g /usr/local/go/src/math/big/arith.go
     2.15s  5.08% 37.48%      6.40s 15.12%  runtime.scanobject /usr/local/go/src/runtime/mgcmark.go
     1.80s  4.25% 41.73%      1.99s  4.70%  runtime.step /usr/local/go/src/runtime/symtab.go
     1.39s  3.28% 45.02%     11.32s 26.74%  runtime.mallocgc /usr/local/go/src/runtime/malloc.go
     1.13s  2.67% 47.69%      1.13s  2.67%  math/big.subVV /usr/local/go/src/math/big/arith_arm64.s
     1.01s  2.39% 50.07%      4.16s  9.83%  math/big.basicMul /usr/local/go/src/math/big/nat.go
```

chrisohaver (commented):
Looks good.  Since the primary use of this is to limit memory consumption of query processing,  it would be nice to see in the README some kind of ball park memory estimation distilled from some tests.

Something like ...

> Each query in flight consumes about __ KB. E.g. 1000 parallel queries consume approximately ___ MB of memory. This is an estimate, YMMV.

fturib:
@willwp : in my opinion the CoreDNS process you are showing is not `throttled`. 
Because if you have the`throttle 1000` active, then your number of go routines should not be above 1000 (maybe 1%). 

Maybe the process was not restarted ? 

NOTE: t`throttle` plugin also provide metrics that you can monitor that shows incoming, dropped and served queries.

Also, about the attached CPU profile result, seems the whole CPU is spend in TLS decryption ...

chrisohaver (approved):

fturib:
> @willwp : I take it back ... we have maybe a problem with TLS : the throttling is most likeley too late in the process.

>if TLS decryption is the one taking the CPU (as it seems based on your CPU profiling file) .. then we should throttle BEFORE the decryption ... and that means either in coredns.Server, either in lib Miekg/dns

Wrong: I delete my comment. TLS is not happening at listening incoming queries (at least in current configuration)

johnbelamaric:
@fturib the TLS here is only on outbound queries, not inbound. So, it is
happening after the throttling. Would be good to see the throttle metrics.

fturib:
@johnbelamaric : I agree, it replied too quicky. 
However, I dived into TLS client, and I think - but not sure yet - that the issue of @willwp is there. TLS is triggering a sub go routine to negotiate the handshake while listening timeout on main goroutine. I think that timeout happen, and the sub goroutine is not finished. 
A way to know it is to run the pprof with the option `block` to get the main point where all these routines are waiting.

I have already a modified `pprof` to accept extra parameter to enable the needed block parameter in order to use that profiling. Will PR asap.

fturib:
@willwp : I propose to continue investigations on your [initial ticket](https://github.com/coredns/coredns/issues/2624) (rather than on the PR). I added comment on that issue. Thanks.

miekg:
This will not be merged until #2593 has properly root-caused. Depending on that outcome we can have a discussion on this functionally, until then NACK

miekg:
@willwp the behaviour your seeing is the purpose of this plugin, it's intentionally lower the amount of queries the binary can handle making a ddos much more trivial.

Also a client retry behaviour is notoriously bad for DNS, you get to see the queries again and again

miekg:
Before I close this PR:
* I will not merge this as a default plugin
* The name *throttle* is not correct; as doesn't do anything as advanced as that, *kill* or *remote_kill* are more apt, because this lowers the bar for a remote DoS considerably.

chrisohaver:
My 2 cents on this plugin...

TLDR; The plugin has legitimate merit, but is hard to tune.

The goal of this plugin is to reduce maximum memory foot print to avoid OOM death.
It doesn't prevent failure, it trades out one mode of failure out for another.  But comparing the two modes of failure, _one is arguably worse than the other_:

1. w/o this plugin: during a spike, memory becomes exhausted, coredns is oom killed repeatedly.  CoreDNS drops queries while it is being restarted or crashed/out of memory.  During this time coredns is also consuming large amounts of memory negatively impacting other resources on the same system.  The speed at which the memory ramps up after each restart depends on the volume of the spike.
2. with this plugin: during a spike, memory is kept in check, but at the expense of dropping excess queries.  The portion of dropped queries depends on the volume of the spike and the max allowed in flight queries.

The plugin _does_ however have a major drawback IMO, that it is difficult to tune.  Proactively it's difficult to advise users what this should be set to, or for example give a fuzzy formula to translate a number inflight packets to an amount of memory.  If a user sets the value too low, then they are going to hobble coredns's capacity (although this could be mitigated by scaling horizontally across multiple hosts).  If they set it too high, then they will still suffer OOM kills during query spikes.  Even arriving at an appropriate value via experimentation could be difficult, since the time a query stays in flight is variable, and can be heavily influenced by outside factors (e.g. slow response from upstream servers).

I think due to the difficulty of tuning, and the negative of tuning too low make this a special use plugin someone should use only if they need to.  So far the need has not been widely reported.

FWIW, If someone finds they that need a feature like this, and they don't want to compile a custom version fo Coredns to use plugin, the same effect more or less could be accomplished by setting up `dnsmasq` in front of coredns configured with a `--dns-forward-max`, which I think defaults to 150.

Issue #2653, opened 2019-03-04
Support exposing records for namespaces by label expression


mdgreenfield:
Request that the `kubernetes` plugin supports a label expression configuration to expose records for matching namespaces.

Currently, the k8s API watches are created for a namespace and pods/services/endpoints using the `labels` expression. However, when a user configures a label on a namespace, they would not also add the same label to every pod/service/endpoint. As a result, trying to specify a `labels` configuration on a namespace restricts watches on pods/services/endpoints resulting in no records for the matched namespace.

I've got changes locally which can address this behavior, however, the biggest question is about the UX of the configuration. I'm happy to propose what the `kubernetes` configuration might look like as well as the PR to address this.

mdgreenfield:
https://github.com/coredns/coredns/pull/2707 addresses this issue. I'm happy to adjust this PR as needed.

PR #2707, opened 2019-03-19
Add `namespace_labels` configuration for kubernetes plugin

mdgreenfield:

The current `kubernetes` plugin does not correctly support a `labels` configuration for namespaces. 

This is due the way the API watches are created for a namespace and pods/services/endpoints. A namespace specific label would not be configured on pods/services/endpoints and as a result, trying to specify a label on a namespace restricts watches on pods/services/endpoints that CoreDNS is aware about.

https://github.com/coredns/coredns/issues/2653

The `kubernetes` plugin `README.md` documentation has been updated.


chrisohaver (approved):
Thanks, LGTM

PR #2711, opened 2019-03-22
plugin/cancel: add context cancelation plugin

miekg:
Per review comments on #2704, move this into a plugin that gets called.
Add the most minimal plugin, tests and documenation.

### DIFF
--- a/plugin/cancel/README.md
+++ b/plugin/cancel/README.md
@@ -12,7 +12,7 @@ triggered after 5001 milliseconds.
 The 5001 number is chosen because the default timeout for DNS clients is 5 seconds, after that they
 give up.
 
-A pluging interested in the cancelation status should call `plugin.Done()` on the context. If the
+A plugin interested in the cancellation status should call `plugin.Done()` on the context. If the
 context was canceled due to a timeout the plugin should not write anything back to the client and
 return a value indicating CoreDNS should not either; a zero return value should suffice for that.
 
@@ -20,7 +20,7 @@ return a value indicating CoreDNS should not either; a zero return value should 
 cancel [TIMEOUT]
 ~~~
 
-* **TIMEOUT** allowss setting a custom timeout. The default timeout is 5001 milliseconds (`5001 ms`)
+* **TIMEOUT** allows setting a custom timeout. The default timeout is 5001 milliseconds (`5001 ms`)
 
 ## Examples
 
--- a/plugin/cancel/README.md
+++ b/plugin/cancel/README.md
@@ -1,0 +1,34 @@
+# cancel
+
+## Name
+
+*cancel* - a plugin that cancels a request's context after 5001 milliseconds.
+
+## Description
+
+The *cancel* plugin creates a canceling context for each request. It adds a timeout that gets
+triggered after 5001 milliseconds.
+
+The 5001 number is chosen because the default timeout for DNS clients is 5 seconds, after that they
+give up.
+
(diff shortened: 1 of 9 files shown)


plugin/cancel/README.md
```diff
@@ -0,0 +1,34 @@
...
+## Name
+
+*cancel* - a plugin that cancels a request's context after 5001 milliseconds.
+
+## Description
+
+The *cancel* plugin creates a canceling context for each request. It adds a timeout that gets
+triggered after 5001 milliseconds.
+
+The 5001 number is chosen because the default timeout for DNS clients is 5 seconds, after that they
```
johnbelamaric, line 12: I know you like to avoid knobs, but I can see use cases for setting this even lower so making it customizable makes sense to me.

plugin/cancel/README.md
```diff
@@ -0,0 +1,34 @@
...
+give up.
+
+A pluging interested in the cancelation status should call `plugin.Done()` on the context. If the
```
johnbelamaric: Should this be plumbed through the higher level plugin.NextOrFailure? Or just a best practice, in which case we should update other plugins to use it.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/cancel..." ]
>I know you like to avoid knobs, but I can see use cases for setting this even lower so making it customizable makes sense to me.

I have no problem with that. The default can be explained and rationalized, if 
you want something else that's fine.

/Miek

--
Miek Gieben

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/cancel..." ]
>+## Description
>+
>+The *cancel* plugin creates a canceling context for each request. It adds a timeout that gets
>+triggered after 5001 milliseconds.
>+
>+The 5001 number is chosen because the default timeout for DNS clients is 5 seconds, after that they
>+give up.
>+
>+A pluging interested in the cancelation status should call `plugin.Done()` on the context. If the
>
>Should this be plumbed through the higher level plugin.NextOrFailure? Or just a best practice, in which case we should update other plugins to use it.

That pretty hard to do (and make genericly usable). Also most plugins won't need 
this as they access something in memory (e.g. cache). So first, and foremost I 
see this being used in *forward* to drop requests.

stp-ip:
lgtm, but waiting for @johnbelamaric to lgtm

plugin/cancel/README.md
```diff
@@ -0,0 +1,45 @@
...
```
johnbelamaric: s/pluging/plugin/

plugin/cancel/README.md
```diff
@@ -0,0 +1,45 @@
...
+context was canceled due to a timeout the plugin should not write anything back to the client and
+return a value indicating CoreDNS should not either; a zero return value should suffice for that.
+
+~~~ txt
+cancel [TIMEOUT]
+~~~
+
+* **TIMEOUT** allowss setting a custom timeout. The default timeout is 5001 milliseconds (`5001 ms`)
```
johnbelamaric: s/allowss/allows/

johnbelamaric (commented):
Just fix these two typos in the docs and it's ready to go.

miekg:
done.

I'll hold off on merging this to see if it actually makes sense to merge.

I would also be helpful to get the kernel timestamp for the received message so 
we can adjust the timeout accordingly.

miekg:
let's see if this works.

Issue #2715, opened 2019-03-24
plugin/log: don't use ErrorFunc


miekg:
plugin/log write a response back to the client if there is an error. This is a) unexpected and b) not needed and c) doesn't help with refactoring cyclic imports.

I think this can be removed entirely which makes the server do it. Also means DefaultErrorFunc can be a private function for dnsserver.

PR #2716, opened 2019-03-24
plugin/log: remove ErrorFunc

miekg:
The server handles this case no need to also do it in the log plugin.

Means DefaultErrorFunc can be private to the dnsserver and is now
renamed to just errorFunc

Fixes: #2715


yongtang (approved):
LGTM

Issue #2717, opened 2019-03-24
vars.Report shouldn't be called in server when metrics are already reported


miekg:
I think we're duplicating vars.Report-ing on error, as it's called from the metrics handler (unconditionally) and then again in core/dnsserver/server.go when there is an error.
This can be removed; and only done in the metrics plugin.

jamirais04:
# - - _~~~~_

PR #2718, opened 2019-03-24
Remove vars.Report from all but metrics plugin

miekg:
This should only happen there.

Fixes: #2717

It depends (and conflicts) with #2715


miekg:
ahh, this is of course undocumented, but there is a test. For names hitting the server that are refused, we're reporting those in the server.

miekg:
this needs some extra documentation which i will add in a new pr.

PR #2719, opened 2019-03-24
Don't double report metrics on error

miekg:
When there is an error use a different function to report the metrics,
in case the plugin chain handled the request the metrics are already
reported.

Builds up on #2716
Fixes: #2717


greenpau (approved):
LGTM 👍

miekg:
just rebased, merge when ci succeeds.

thanks!

Issue #2574, opened 2019-02-17
plugin/cache: cache failures?


miekg:
Does it make sense to cache (with some low TTL) upstream failures? This includes SERVFAIL, NOTIMPL and timeouts for names we're trying to query.

miekg:
This would be not hard to implement and would be nice; the only problem is what TTL to put on these.

miekg:
has been merged.

PR #2720, opened 2019-03-25
[plugin/cache] cache failures

ckcd:
cache upstream failures which includes `SERVFAIL` and `NOTIMPL`. for now the `TTL` is 5s.

https://github.com/coredns/coredns/issues/2574

No

No

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -163,6 +163,9 @@ func (w *ResponseWriter) WriteMsg(res *dns.Msg) error {
 	var duration time.Duration
 	if mt == response.NameError || mt == response.NoData {
 		duration = computeTTL(msgTTL, w.minnttl, w.nttl)
+	} else if mt == response.ServerFailureError || mt == response.NotImplementedError {
+		// use default ttl which is 5s
+		duration = minTTL
 	} else {
 		duration = computeTTL(msgTTL, w.minpttl, w.pttl)
 	}
@@ -206,7 +209,7 @@ func (w *ResponseWriter) set(m *dns.Msg, key uint64, mt response.Type, duration 
 		i := newItem(m, w.now(), duration)
 		w.pcache.Add(key, i)
 
-	case response.NameError, response.NoData:
+	case response.NameError, response.NoData, response.ServerFailureError, response.NotImplementedError:
 		i := newItem(m, w.now(), duration)
 		w.ncache.Add(key, i)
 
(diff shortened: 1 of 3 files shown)


plugin/pkg/response/typify.go
```diff
@@ -28,13 +32,15 @@ const (
...
-	NoError:    "NOERROR",
-	NameError:  "NXDOMAIN",
-	NoData:     "NODATA",
-	Delegation: "DELEGATION",
-	Meta:       "META",
-	Update:     "UPDATE",
-	OtherError: "OTHERERROR",
+	NoError:             "NOERROR",
+	NameError:           "NXDOMAIN",
+	ServerFailureError:  "SERVFAIL",
```
miekg: I think it makes sense to lump these together and have a single `ServerError` which for now encompasses both SERVFAIL and NOTIMPL

plugin/cache/cache.go
```diff
@@ -163,6 +163,9 @@ func (w *ResponseWriter) WriteMsg(res *dns.Msg) error {
 	var duration time.Duration
 	if mt == response.NameError || mt == response.NoData {
 		duration = computeTTL(msgTTL, w.minnttl, w.nttl)
+	} else if mt == response.ServerFailureError || mt == response.NotImplementedError {
+		// use default ttl which is 5s
+		duration = computeTTL(msgTTL, minTTL, maxTTL)
```
miekg: there should not be any data in these kind of message, so we can just assume 5s as the TTL

miekg (changes requested):
lgtm overall, some thoughts and making it slightly simpler

plugin/cache/cache.go
ckcd: Done~

plugin/pkg/response/typify.go
ckcd: It's hard to do because the `RCode` in dns message do NOT have a `ServerError`. So we need the accurate error type.

miekg: I want to limit the number of things in this list, OtherError also lumps together errors that are not `Rcode`s. So I still think this makes sense - i.e. a error category that says the server was unresponsive

ckcd: OK i got it~  the modify is DONE, please have a review~

Thanks !

ckcd:
@miekg 

Hi, i have do the change, please have a review, thanks~

miekg (approved):
lgtm, I'll merge after the 1.5.0 release

PR #2722, opened 2019-03-25
Remove server addr from the context

miekg:
This was added twice, just leave the server which also holds the address.

Conflicts with #2719 but should be easy to fix.


yongtang (approved):
LGTM

yongtang:
Actually, think several tests needs to be changes as well because of the PR.

PR #2726, opened 2019-03-26
Remove context.Context from request.Request

### LINKED ISSUE #2721: request.Request should not carry a context.Context
This struct should not carry a `Context`, this is only done for pkg/upstream, which should just receive a context in the funcion.

miekg:
This removes the context from request.Request and makes all the changes
in the code to make it compile again. It's all mechanical. It did
unearth some weirdness in that the context was kept in handler structs
which may cause havoc with concurrently handling of requests.

Fixes #2721


bradbeam (approved):

miekg:
thanks for reviewing!

PR #2729, opened 2019-03-26
plugin/pprof - add option to enable block profiling

fturib:

In order to collect profiling information for blocked routines, we need to set a specific function of runtime.

see changes in the README.

This pprof option is useful to investigate the reason of high number of goroutines and associated memory or CPU.

This profiling mode was needed to investigate: #2593
Will allow to investigate the Mem+CPU spike here : #2624

README is updated

no.

### DIFF
--- a/plugin/pprof/README.md
+++ b/plugin/pprof/README.md
@@ -17,10 +17,14 @@ This plugin can only be used once per Server Block.
 ## Syntax
 
 ~~~
-pprof [ADDRESS]
+pprof [ADDRESS] [BLOCKRATE]
 ~~~
 
-If not specified, ADDRESS defaults to localhost:6053.
+- If not specified, **ADDRESS** defaults to localhost:6053.
+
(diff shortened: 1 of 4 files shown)


rajansandeep (approved):
I have tested this and it works as expected.

plugin/pprof/README.md
```diff
@@ -42,11 +46,11 @@ Listen on an alternate address:
 }
 ~~~
 
-Listen on an all addresses on port 6060:
+Listen on an all addresses on port 6060: and enable block profiling
 
 ~~~ txt
 . {
-    pprof :6060
+    pprof :6060 1
```
miekg: lgtm, but I would suggest making this
~~~
pprof ADDRESS {
    blockrate [RATE]
}
~~~
where the default will be 1 is just `blockrate` is specified.

This change helps if we want even more options later.

fturib: ok updated.
```
pprof ADDRESS {
    block [RATE]
}
```

PR #2732, opened 2019-03-29
plugin/forward: bring back jitter

miekg:
Very minimal jitter implementation that adds up to 25% jitter to the
expiration for the connection. This helps when we see a burst in
connections which all race to get a TLS conn and then also are expired
at the same time, meaning we can see a repeat.

Another thing could be to keep a couple of connection in the cache to be
able to serve right away (not implemented in this PR)

### DIFF
--- a/plugin/forward/jitter.go
+++ b/plugin/forward/jitter.go
@@ -1,0 +1,12 @@
+package forward
+
+import (
+	"math/rand"
+	"time"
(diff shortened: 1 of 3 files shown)


miekg:
doesn't solve the core problem; only slightly makes it better; if at all (depends on query pattern)

Issue #2309, opened 2018-11-15
plugin/route53: NXDOMAIN instead of NOERROR with overlapping domains


mattlqx:
I'm doing split horizon with private and public Route 53 zones for the same domain name. I see that the standard behavior of returning NOERROR for an AAAA/MX record when an A exists for the query seems broken in this setup. I assume this is because when a query is a miss and falls to the next zone and is a miss there also, internally it's looking at that last zone for the A record instead of keeping track whether it saw one in any of the preceding zones. But I think this is definitely a bug.

```
example.com {
  route53 example.com.:ZZZZZZ3A0GBKDI example.com.:ZZZZZZ3ABBBBB {
    upstream 127.0.0.1:253  # unbound for recursion
  }
  prometheus
  cache 300
  errors stdout
  log stdout
}
```

For clarity, I'm expecting a response of NOERROR for an AAAA query for a record that has an A record in the left-most zone. Actual behavior is returning NXDOMAIN. This of course is throwing the `host` command for a loop with output like this as it tries to also lookup AAAA and MX records:

```
Using domain server:
Name: 127.0.0.1
Address: 127.0.0.1#53
Aliases: 

foo.example.com has address 127.0.0.1
Host foo.example.com not found: 3(NXDOMAIN)
Host foo.example.com not found: 3(NXDOMAIN)
```

I'm running CoreDNS 1.2.4 presently.

dilyevsky:
Yeah this should be fixed. For context, relevant rfc https://tools.ietf.org/html/rfc4074#section-3:

```
Suppose that an authoritative server has an A RR but has no AAAA RR
   for a host name.  Then, the server should return a response to a
   query for an AAAA RR of the name with the response code (RCODE) being
   0 (indicating no error) and with an empty answer section (see
   Sections 4.3.2 and 6.2.4 of [1]).  Such a response indicates that
   there is at least one RR of a different type than AAAA for the
   queried name, and the stub resolver can then look for A RRs.

   This way, the caching server can cache the fact that the queried name
   has no AAAA RR (but may have other types of RRs), and thus improve
   the response time to further queries for an AAAA RR of the name.
```

dilyevsky:
Oh and thanks for report, @mattlqx !

dilyevsky:
Hm, I'm having difficulty reproducing this (in test). I'll have some more time to work on it this week.

stp-ip:
This might be related with #2334

mattlqx:
[...]

PR #2734, opened 2019-03-29
[plugin/route53]: Do not return NXDOMAIN where it should be NODATA.

dilyevsky:
Fixes two cases of returning NXDOMAIN where it should be NODATA (name exists but not for the requested record type). This should address issues where some resolves make ANY requests and then may negative cache NXDOMAIN response which breaks resolution. Adds relevant unit tests.

Fixes issue: #2309 

Signed-off-by: Dmitry Ilyevskiy <dmitry.ilyevskiy@getcruise.com>


yongtang (approved):
LGTM

PR #2735, opened 2019-03-29
server: remove if registry != nil code

miekg:
I can't find what's this code should be doing and it looks like it's
not doing anything.

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -66,22 +66,8 @@ func NewServer(addr string, group []*Config) (*Server, error) {
 		}
 		// set the config per zone
 		s.zones[site.Zone] = site
+
 		// compile custom plugin for everything
-		if site.registry != nil {
-			// this config is already computed with the chain of plugin
-			// set classChaos in accordance with previously registered plugins
-			for name := range EnableChaos {
-				if _, ok := site.registry[name]; ok {
-					s.classChaos = true
-					break
-				}
-			}
-			// set trace handler in accordance with previously registered "trace" plugin
-			if handler, ok := site.registry["trace"]; ok {
-				s.trace = handler.(trace.Trace)
-			}
-			continue
-		}
 		var stack plugin.Handler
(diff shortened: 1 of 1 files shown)


fturib:
It was introduced to fix : https://github.com/coredns/coredns/issues/1636

miekg:
ah, thanks for your good memory, sadly that commit doesn't have a test; didn't add a comment in code ;(

miekg:
can't even reproduce the original issues

miekg:
Corefile
~~~
. {
        chaos
	bind 127.0.0.1
	bind 192.168.87.28
	forward . 8.8.8.8:53
}
~~~

and 
~~~
% dig @192.168.87.28 -p 1053 version.bind txt ch +noall +answer                                                                                                
version.bind.		0	CH	TXT	"CoreDNS-1.4.0"
% dig @127.0.0.1 -p 1053 version.bind txt ch +noall +answer                                                                                                    
version.bind.		0	CH	TXT	"CoreDNS-1.4.0"
~~~
Or I'm I wrong here?

<strike>Update: gah: wrong branch. *sigh*</strike>
No on this branch also *just works*

fturib:
Adding @UladzimirTrehubenka who raised the issue and the proposed the fix.

johnbelamaric:
You reverted more than just this code. The original did a "continue" if registry != nil. You instead go through and construct the plugin chain again.

That said, I don't understand why if it's already gone through that loop (thus making it != nil), the original bug would exist at all.

UladzimirTrehubenka:
Original Miek's code was (to avoid unnecessary construct the plugin chain):
```
if site.registry != nil {
    // this config is already computed with the chain of plugin
    // here the fix for set chaos/trace flags
    continue
}
```
So I cannot say "looks like it's not doing anything" but PR is correct.

miekg:
ok, this seems to be triggered only if you have multiple `bind` statements in a server. I can't anything at fault with removing this, other than doing slightly more work during setup. Result is less confusing

miekg:
removing as not needed.

PR #2736, opened 2019-03-30
plugin/cancel: enable by default

miekg:
Enable the cancel plugin (with some hackery to prevent cyclic imports)
to be active in every server. If the plugin *is* loaded that takes
precedence, allowing folks to set different timeouts than the default of
5001 ms.

Add several tests and document how to plugin.Done(ctx) in code.
Implement it in *forward* to actually make that return early if we're
only executing that plugin late in the process.

### DIFF
--- a/core/dnsserver/cancel_test.go
+++ b/core/dnsserver/cancel_test.go
@@ -1,0 +1,41 @@
+package dnsserver
+
+import (
+	"context"
+	"testing"
+
+	"github.com/miekg/dns"
+)
+
+// We can't include the actual cancel plugin here, because that create a cyclic dependency. Just make
+// a plugin with the same name because that's what triggers the check.
+type cancel struct{}
+
+func (c cancel) Name() string { return "cancel" }
+func (c cancel) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	return 0, nil
+}
+
+func TestCancelEnabled(t *testing.T) {
+	c := testConfig("dns", cancel{})
+	s, err := NewServer("127.0.0.1:53", []*Config{c})
+
+	if err != nil {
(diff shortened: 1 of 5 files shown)


miekg:
@WinstonPrivacy this hooks up the *cancel* plugin some more and has a check to return (earlier) in the *forward* plugin.

miekg:
apart from this a metric that tracks the average lifetime of a goroutine would be nice, but hard to implement.

core/dnsserver/cancel_test.go
```diff
@@ -0,0 +1,41 @@
+package dnsserver
+
+import (
+	"context"
+	"testing"
+
+	"github.com/miekg/dns"
+)
+
+// We can't include the actual cancel plugin here, because that create a cyclic dependency. Just make
```
chrisohaver, line 10:
```suggestion
// We can't include the actual cancel plugin here, because that creates a cyclic dependency. Just make
```

plugin/forward/forward.go
```diff
@@ -82,6 +82,10 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 			fails = 0
 		}
 
+		if plugin.Done(ctx) {
+			return dns.RcodeSuccess, plugin.Error(f.Name(), ctx.Err())
+		}
+
```
chrisohaver, line 88: Should this go [in here](https://github.com/coredns/coredns/pull/2736/files#diff-b12d06e8aa6117ce05d5715e1c8930c3R115) to break before tcp retry if we need to?

chrisohaver (commented):
LGTM, one question in comments

miekg:
I may close this as not needed. First I want to stress test the coredns binary and check if this would make any difference

miekg:
there seems to be no need for this. Closing for now.

PR #2738, opened 2019-03-30
WIP: add debian packaging

miekg:
Uses https://github.com/goreleaser/nfpm, nfpm pkg create a debian
foo.deb that can be installed.

nfpm can be used as a lib (and goreleaser uses it). We can do the
minimal thing add a little Go binary to create deb and maybe rpms.

### DIFF
--- a/coredns.postinst
+++ b/coredns.postinst
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+set -e
+
+case "$1" in
+    configure)
+        if ! getent passwd coredns > /dev/null; then
+            adduser --system --disabled-password --disabled-login --home /var/lib/coredns \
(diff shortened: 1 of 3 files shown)


miekg:
@yongtang something like this seems very sensible

miekg:
have to think a bit where we want to store the additional files postinstall and the systemd service file should live. They could technically live in another repository which has pros and cons.

yongtang:
I think having files in this repo (coredns) makes sense, if we want to have deb as part of the coredns release. This effectively promotes deb to first class citizen.

miekg:
going to do this outside the main coredns/coredns repo, because it's cleaner to keep this out.

Issue #2537, opened 2019-02-08
plugin/auto|file: fully deprecate TIMEOUT + NO_RELOAD


stp-ip:
Current (PR #2516) workings:
TIMEOUT & RELOAD not set -> 60s for both
TIMEOUT set -> use value for both
RELOAD set -> use value for both
TIMEOUT & RELOAD set -> use specific values

Deprecated implementation TIMEOUT:
TIMEOUT set -> error
RELOAD not set -> 60s default
RELOAD set -> use specific value

Deprectated implementation NO_RELOAD:
NO_RELOAD set -> error

Release notes:
```
Breaking: plugin/auto removes TIMEOUT and NO_RELOAD. plugin/file removes NO_RELOAD. Both plugin/file and plugin/auto now only allows the usage of RELOAD.
```

stp-ip:
@nzoueifi as you worked on the previous PR. Do you want to take this up?
That would mean the final deprecation would be done in 1.6.0

stp-ip:
Or @ihac :)

ihac:
@stp-ip Thanks for reminding. I'll create a pr today.

miekg:
We can just include it in 150, I can wait a bit

PR #2742, opened 2019-03-31
Fully deprecate TIMEOUT and NO_RELOAD

ihac:
Since the deprecation of `TIMEOUT` and `NO_RELOAD` was announced in v1.4.0, this PR fully deprecates them and refactors related unit tests.

[#2537: plugin/auto|file: fully deprecate TIMEOUT + NO_RELOAD](https://github.com/coredns/coredns/issues/2537)

This PR removes the docs related to `TIMEOUT` and `NO_RELOAD`.

Yes, old configuration with `TIMEOUT` or `NO_RELOAD` will not be parsed.

### DIFF
--- a/plugin/auto/README.md
+++ b/plugin/auto/README.md
@@ -17,7 +17,6 @@ zonefile. New or changed zones are automatically picked up from disk.
 auto [ZONES...] {
     directory DIR [REGEXP ORIGIN_TEMPLATE [TIMEOUT]]
     reload DURATION
-    no_reload
     upstream
 }
 ~~~
@@ -37,7 +36,6 @@ are used.
 * `reload` interval to perform reloads of zones if SOA version changes and zonefiles. Default is one minute.
   Value of `0` means to not scan for changes and reload. eg. `30s` checks zonefile every 30 seconds
   and reloads zone when serial changes.
-* `no_reload` deprecated. Sets reload to 0.
 * `upstream` defines upstream resolvers to be used resolve external names found (think CNAMEs)
   pointing to external names. CoreDNS will resolve CNAMEs against itself.
 
--- a/plugin/auto/README.md
+++ b/plugin/auto/README.md
@@ -15,7 +15,8 @@ zonefile. New or changed zones are automatically picked up from disk.
 
 ~~~
 auto [ZONES...] {
-    directory DIR [REGEXP ORIGIN_TEMPLATE [TIMEOUT]]
+    directory DIR [REGEXP ORIGIN_TEMPLATE]
+    transfer to ADDRESS...
     reload DURATION
     upstream
 }
@@ -29,11 +30,7 @@ are used.
   like `{<number>}` are replaced with the respective matches in the file name, e.g. `{1}` is the
   first match, `{2}` is the second. The default is: `db\.(.*)  {1}` i.e. from a file with the
   name `db.example.com`, the extracted origin will be `example.com`.
-  **TIMEOUT** is deprecated and will be removed in a subsequent version. 
-  `reload` will be used, if not defined
-  (it specifies how often CoreDNS should scan the directory to watch for file removal and addition;
-  the default is every 60 seconds. This value is in seconds. The minimum value is 1 second.)
-* `reload` interval to perform reloads of zones if SOA version changes and zonefiles. Default is one minute.
+* `reload` interval to perform reloads of zones if SOA version changes and zonefiles. It specifies how often CoreDNS should scan the directory to watch for file removal and addition. Default is one minute.
   Value of `0` means to not scan for changes and reload. eg. `30s` checks zonefile every 30 seconds
   and reloads zone when serial changes.
 * `upstream` defines upstream resolvers to be used resolve external names found (think CNAMEs)
@@ -73,7 +70,8 @@ where `example.org` is the origin. Scan every 45 seconds.
 ~~~ corefile
 org {
     auto {
-        directory /etc/coredns/zones/org www\.db\.(.*) {1} 45
+        directory /etc/coredns/zones/org www\.db\.(.*) {1}
+        reload 45s
     }
 }
 ~~~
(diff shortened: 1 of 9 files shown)


ihac:
Hi, @stp-ip @miekg 
This PR deprecates the use of `TIMEOUT` and `NO_RELOAD` fully as you planned in [issue 2537](https://github.com/coredns/coredns/issues/2537). However, according to the deprecation policy, CoreDNS is expected to be compatible with old configuration in `Release x.y+1.0` (i.e. 1.5.0). My PR seems to break the rule because all old Corefiles with `TIMEOUT` or `NO_RELOAD` used will not work. This is the reason why CI failed.
I'm a little bit confused~Do I need to update the CI tests?

miekg:
We're jumping to 150 from 140, I honestly can't remember what's now being deprecated other than this was on the list at some point. So I'll just merged when this is not failing CI.

miekg:
Yes tests need to be updated as well and some readme is now failing to parse:
~~~
    metrics_test.go:88: Could not get CoreDNS serving instance: plugin/auto: Corefile:3 - Error during parsing: Wrong argument count or unexpected line ending after '1'
=== RUN   TestMetricsSeveralBlocs
--- PASS: TestMetricsSeveralBlocs (0.02s)
=== RUN   TestMetricsPluginEnabled
--- PASS: TestMetricsPluginEnabled (0.00s)
=== RUN   TestLookupBalanceRewriteCacheDnssec
=== PAUSE TestLookupBalanceRewriteCacheDnssec
=== RUN   TestProxyThreeWay
--- PASS: TestProxyThreeWay (6.00s)
=== RUN   TestLookupProxy
=== PAUSE TestLookupProxy
=== RUN   TestReadme
--- FAIL: TestReadme (6.05s)
    readme_test.go:72: Failed to start server with ../plugin/auto/README.md, for input "plugin/auto: Corefile:3 - Error during parsing: Wrong argument count or unexpected line ending after '45'":
        org {
            auto {
                directory /etc/coredns/zones/org www\.db\.(.*) {1} 45
            }
        }
~~~

ihac:
Thanks @miekg 
Tests and readme now have been updated.

plugin/auto/README.md
```diff
@@ -15,9 +15,8 @@ zonefile. New or changed zones are automatically picked up from disk.
 
 ~~~
 auto [ZONES...] {
-    directory DIR [REGEXP ORIGIN_TEMPLATE [TIMEOUT]]
+    directory DIR [REGEXP [ORIGIN_TEMPLATE]]
```
miekg: you've added a `[` here that wasn't here before.

miekg: should it be here?

miekg (commented):
lgtm

1 question

plugin/auto/README.md
ihac: Actually I was trying to make it stay consistent with the code, where `TEMPLATE` is optional. 

``` golang
			case "directory": // directory DIR [REGEXP [TEMPLATE]]
				if !c.NextArg() {
					return a, c.ArgErr()
				}
				a.loader.directory = c.Val()
                                 // ...

				// regexp
				if c.NextArg() {
                                         // ...
			        }
				// template is optional here
				if c.NextArg() {
					a.loader.template = rewriteToExpand(c.Val())
				}
```

I'm not sure whether you were making `TEMPLATE` optional on purpose. I think it is fine, but we should set a default value (e.g. `{0}`). WDYT?
BTW, I personally prefer a `return fast` coding style (return as soon as no args can be parsed).

stp-ip:
Partially the warning was already present and the docs were already modified before 1.4.0, but we didn't cut a minor release and so these changes were officially announced with 1.4.0.
In my view deprecating them in 1.5.0 seems ok. Also it's not the main breaking change.
Also taking a look over the code in a sec.

stp-ip (approved):
/lgtm

Small nitpick related to the plugin/auto docs:
We show "transfer to/from" within the example, but don't actually document the variable etc.

I would add this from plugin/file:
```
transfer enables zone transfers. It may be specified multiples times. To or from signals the direction. ADDRESS must be denoted in CIDR notation (e.g., 127.0.0.1/32) or just as plain addresses. The special wildcard * means: the entire internet (only valid for 'transfer to'). When an address is specified a notify message will be send whenever the zone is reloaded.
```

plugin/auto/README.md
stp-ip: As far as my understanding of intent goes:
The optionality and the default is for both the regex and the template as they are quite correlated. That's why it wasn't separately shown as optional, but one "bundle".

`directory DIR [REGEXP ORIGIN_TEMPLATE]`

The code should reflect the default for this bundle to be `db\.(.*) {1}` afaik.

ihac:
@stp-ip Oops, I just forgot it~~
I'll include it in this PR.

plugin/auto/README.md
ihac: That make sense. So we should raise an error when `REGEXP` is set but `ORIGIN_TEMPLATE` is not?

stp-ip:
> @stp-ip Oops, I just forgot it~~
> I'll include it in this PR.

That sounds reasonable.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fully depreca..." ]
>ihac commented on this pull request.
>
>
>
>> @@ -15,9 +15,8 @@ zonefile. New or changed zones are automatically picked up from disk.
>
> ~~~
> auto [ZONES...] {
>-    directory DIR [REGEXP ORIGIN_TEMPLATE [TIMEOUT]]
>+    directory DIR [REGEXP [ORIGIN_TEMPLATE]]
>
>That make sense. So we should raise an error when `REGEXP` is set but `ORIGIN_TEMPLATE` is not?

That can wait (a bit) and should probably be done in a new PR (as it's not 
really related to this one)

ihac:
I've updated the PR (remove redundant `[`).

PR #2747, opened 2019-04-01
plugin/autopath: Document known autopath issue

chrisohaver:
Documents a known case where autopath doesn't work as one might expect.

None.

No.


rajansandeep (approved):
/lgtm

Issue #2660, opened 2019-03-08
CoreDNS eats up CPU and I/O


MartinEmrich:
Hi!

CoreDNS in my K8s cluster (3 Masters, 3 Workers, maybe 70 pods in all) is eating up all CPU (200-300%) and has immense I/O as seen with iotop:

```
Total DISK READ :     213.67 M/s | Total DISK WRITE :     156.35 K/s
Actual DISK READ:     214.11 M/s | Actual DISK WRITE:     159.83 K/s
    TID  PRIO  USER     DISK READ  DISK WRITE  SWAPIN     IO>    COMMAND                                                                                                                                                                                                       
 842404 be/4 root	48.65 M/s    0.00 B/s  0.00 % 24.67 % coredns -conf /etc/coredns/Corefile
 850515 be/4 root	35.77 M/s    0.00 B/s  0.00 % 21.27 % coredns -conf /etc/coredns/Corefile
 842289 be/4 root	47.30 M/s    0.00 B/s  0.00 % 21.12 % coredns -conf /etc/coredns/Corefile
 842420 be/4 root       45.04 M/s    0.00 B/s  0.00 % 20.85 % coredns -conf /etc/coredns/Corefile
 842413 be/4 root	12.47 M/s    0.00 B/s  0.00 % 18.88 % coredns -conf /etc/coredns/Corefile
 842435 be/4 root	16.34 M/s    0.00 B/s  0.00 % 12.10 % coredns -conf /etc/coredns/Corefile
```

K8s Version 1.13.0
CoreDNS version 1.2.6

CoreDNS Config (Default from kubeadm/kubernetes):
```
    .:53 {
        errors
        health
        kubernetes cluster.local in-addr.arpa ip6.arpa {
           pods insecure
           upstream
           fallthrough in-addr.arpa ip6.arpa
        }
        prometheus :9153
        proxy . /etc/resolv.conf
        cache 30
        loop
        reload
        loadbalance
    }
```

Host resolv.conf:
```
search some.sub.domain.company.com
nameserver 10.12.96.3
nameserver 172.24.187.220
nameserver 172.24.186.30
```

`kubectl logs` does not give any clues.

What could be wrong?

johnbelamaric:
No clue. It should be doing very very low reads (just the reloads of the config and resolv.conf periodically). Anything from `kubectl describe`?

MartinEmrich:
Sorry, nothing interesting there:

```
[...]

PR #2748, opened 2019-04-01
Bump the resync period to five hours

johnbelamaric:
Bumps the default client-go resync period to 5 hours from 5 minutes. Resync is no longer recommended to be done frequently, and uses extra CPU unnecessarily. Note that it does NOT refetch data from the API.

None.

None, the default wasn't documented before (maybe it should be though :) )

No.

### DIFF
--- a/plugin/kubernetes/setup.go
+++ b/plugin/kubernetes/setup.go
@@ -321,4 +321,4 @@ func searchFromResolvConf() []string {
 	return rc.Search
 }
 
-const defaultResyncPeriod = 5 * time.Minute
+const defaultResyncPeriod = 5 * time.Hour


chrisohaver (approved):
lgtm

miekg:
is this https://github.com/coredns/coredns/issues/2660?

And can we stop resyncing all together?

chrisohaver:
Could be. Although long ago I did not see high cpu, even when deploying on clusters with 10K services and 150k endpoints (the maximum allowed) with the 5 minute re-sync.  But that was some time before 1.2.6.

If resync does not refetch, what exactly does re-sync do? [Docs](https://github.com/kubernetes/client-go/blob/master/tools/cache/controller.go#L307) say  ...

```
* resyncPeriod: if non-zero, will re-list this often (you will get OnUpdate
  calls, even if nothing changed). Otherwise, re-list will be delayed as
  long as possible (until the upstream source closes the watch or times out,
  or you stop the controller).
```

It says that we would get update events for watch objects even if nothing changed. That seems to suggest refetching data (in the form of updates)...unless updates only contain deltas.

> Resync is no longer recommended to be done frequently, and uses extra CPU unnecessarily.

@johnbelamaric, Can you point to where this is recommended?  Perhaps there is some context explaining why.  Is the CPU concern primarily for the client or for the API server.

johnbelamaric:
See https://groups.google.com/forum/#!topic/kubernetes-sig-api-machinery/PbSCXdLDno0 for example. It's not clear if it re-lists or not, I have seen conflicting info on that. I haven't dug into the code.

Regardless, this was an "insurance policy" guarding against watch bugs. It now primarily guards against controller bugs. But since we are a readonly controller, I think it's not needed often.

johnbelamaric:
@miekg it can be disabled by setting it to zero. I'll mention it on that issue. Not sure if it's a good idea or not. It will relist on error even if resync is zero. I'll ask around.

johnbelamaric:
Ok, yes, @lavalamp has told me it is safe to disable it. It serves two primary purposes: paper over controller bugs; and trigger a "poll" of external systems if you need to integrate with them (essentially, it's a time.Tick trigger for your reconciliation loop at that point).

Neither really applies, we would rather fix any bugs we have than paper over them.

I will updated it to be disabled by default.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Bump the resy..." ]
>Ok, yes, @lavalamp has told me it is safe to disable it. It serves two primary purposes: paper over controller bugs; and trigger a "poll" of external systems if you need to integrate with them (essentially, it's a time.Tick trigger for your reconciliation loop at that point).
>
>Neither really applies, we would rather fix any bugs we have than paper over them.

+1 to that.

/Miek

--
Miek Gieben

PR #2752, opened 2019-04-02
plugin/kubernetes: Disable resync by default

johnbelamaric:
Disables the client-go resync by default.

Maybe #2660 

README (done)

No

### DIFF
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -48,7 +48,8 @@ kubernetes [ZONES...] {
 }
 ```
 
-* `resyncperiod` specifies the Kubernetes data API **DURATION** period.
+* `resyncperiod` specifies the Kubernetes data API **DURATION** period. By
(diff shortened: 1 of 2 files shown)


chrisohaver:
If there is no reason to make this anything other than zero, then IMO we should remove it as an option.
One less _kubernetes_ option would be nice.

bradbeam (approved):

bradbeam:
@chrisohaver Perhaps we can start the deprecation process and remove it later?

miekg:
>@chrisohaver Perhaps we can start the deprecation process and remove it later?

nooping the options is a good first start; we can still add this to the 1.5.0 
release and notes.

johnbelamaric:
How about we wait a release before nooping it? Because "should be safe" and "is proven safe" are two different things. Nice to have a configuration option to fallback until we are confident in the latter.

PR #2753, opened 2019-04-02
manifest-tool doesn't support mips

chrisohaver:

Looks like `linux/mips` is not supported by `manifest-tool` ...

```
manifest-tool push from-args --platforms linux/amd64,linux/arm,linux/arm64,linux/ppc64le,linux/s390x,linux/mips --template chrisohaver/coredns:coredns-ARCH --target chrisohaver/coredns:1.4.0-6f5b294-kubernetai
FATA[0007] Manifest entry for image chrisohaver/coredns:coredns-mips has unsupported os/arch combination: linux/mips
```

This removes mips from the manifest, and builds it separately.


miekg:
Hmm. That's unfortunate. Can you add a comment in the Makefile as well, why
mips is handled differently.

Otherwise lgmt

PR #2761, opened 2019-04-05
plugin/hosts: send nxdomain response

tracernz:

[As specified in the plugin documentation](https://github.com/coredns/coredns/blob/master/plugin.md#writing-plugins), CoreDNS assumes a plugin has sent a response in all cases except SERVFAIL, REFUSED, FORMERR, and NOTIMP. The hosts plugin was returning NXDOMAIN but sending no response in the case where it didn't have any records for the qname, and fallthrough was not available. __This resulted in no response being sent to the client.__

Additionally, when writing a test case, I encountered a segfault in plugin/test SortAndCheck because it dereferences the response without checking for nil (resp is nil when the plugin fails to write one). Perhaps this commit should be a separate PR?

I didn't find an issue on github.

None.

No.

### DIFF
--- a/plugin/hosts/hosts.go
+++ b/plugin/hosts/hosts.go
@@ -52,12 +52,14 @@ func (h Hosts) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 		answers = aaaa(qname, h.options.ttl, ips)
 	}
 
+	rcode := dns.RcodeSuccess
+
 	if len(answers) == 0 {
 		if h.Fall.Through(qname) {
 			return plugin.NextOrFailure(h.Name(), h.Next, ctx, w, r)
 		}
 		if !h.otherRecordsExist(state.QType(), qname) {
-			return dns.RcodeNameError, nil
+			rcode = dns.RcodeNameError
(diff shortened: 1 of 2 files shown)


miekg:
SortAndCheck might require another PR; generally I don't care that much if a test panic because something unexpected happen, because that's a problem that needs to be fixed anyway: IOW: just leave that current behavior

miekg (commented):
lgtm, but please revert the test-nil-check-change.

tracernz:
> generally I don't care that much if a test panic because something unexpected happen, because that's a problem that needs to be fixed anyway

Yeah, I could go either way. Removed that commit. :+1:

miekg:
can you rebase on master? That will make travis work again. Tx.

tracernz:
Done.

miekg:
oh, I was just about to merge this and then I remembered something. I don't think you can say nxdomain on anything here, because the hosts plugin just doesn't know. So the current behaviour is correct. although this could use a bunch of comments in the source....

tracernz:
So, the current behaviour is that coredns simply doesn't send a response at all.

PR #2764, opened 2019-04-06
manpages: chop off first header

miekg:
each readme starts with the plugin's name as a header, this needs to be
chopped off to provide a good manual page.


miekg:
all mechanical, apart from the Makefile.doc change.

PR #2766, opened 2019-04-06
WIP: travis changes

miekg:
Simplify Travis so it fails less often.

We don't need docker any more, let alone trusty and sudo, so simplifies
this, to just os: linux.


miekg:
*sigh* merging because we can now trust travis slightly more

PR #2769, opened 2019-04-07
kubernetes: never respond with NXDOMAIN for authority label

bhcleek:
Return a nodata response when trying to resolve the authority's label
for a record type that doesn't match the record type of the authority.

This guards against poisoning the authority record by requesting the
wrong record type for the authority label. For instance, given an
authoritative resolver that uses subdomain delegation for Kubernetes
services of a cluster that's configured to use IPv4, the parent may be
poisoned by querying it for the authority label of the cluster subdomain
with a AAAA record type, which would otherwise (i.e. without this
change) return an NXDOMAIN. That is, given
```
	cluster.example.com        NS 10800 ns.dns.cluster.example.com
	ns.dns.cluster.example.com A  10800 10.0.1.2
```
The parent may be poisoned for the SOA TTL by querying it for a AAAA
record of ns.dns.cluster.example.com, causing the parent to fail
delegate properly until the cached NXDOMAIN response expires.

I looked for an existing issue, but didn't find one. If the PR itself is not sufficient, I can create a new issue.

None.

Not to my knowledge.

### DIFF
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -112,10 +112,18 @@ func (k *Kubernetes) Services(ctx context.Context, state request.Request, exact 
 		return []msg.Service{svc}, nil
 	}
 
-	if state.QType() == dns.TypeA && isDefaultNS(state.Name(), state.Zone) {
+	if isDefaultNS(state.Name(), state.Zone) {
+		ns := k.nsAddr()
+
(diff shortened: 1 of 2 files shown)


chrisohaver (approved):

chrisohaver:
IIUC, this and related code will need to need to be updated once K8s supports dual stack services. But it doesn't yet, so OK for now.

PR #2771, opened 2019-04-07
[WIP] Dedup common code in forward and grpc

palash25:


miekg:
this seems to have stalled a bit, it also adds more code then it removes; sorta negating the effort.

PR #2772, opened 2019-04-08
plugins: remove `upstream [ADDRESS]` from docs

miekg:
Various plugins still had this documented, the setup functions still
allow it (which is fine), but update the docs that this arguments isn't
there any more.


yongtang (approved):
LGTM

PR #2782, opened 2019-04-11
Rebase coredns to distroless

yuwenma:
Advantage of using Distroless comparing to empty baseimage.
* Avoid the caveats when running go static binaries due to some missing non-binary dependencies.
* Distroless has ca-certifiactes preinstalled.
* A nit Dockerfile.

See detailed comparisons in https://github.com/kubernetes/enhancements/blob/master/keps/sig-release/20190316-rebase-images-to-distroless.md#type-1-from-scratch

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,4 @@
-FROM debian:stable-slim
-
-RUN apt-get update && apt-get -uy upgrade
-RUN apt-get -y install ca-certificates && update-ca-certificates
-
-FROM scratch
-
-COPY --from=0 /etc/ssl/certs /etc/ssl/certs
+FROM gcr.io/distroless/static:latest
 ADD coredns /coredns
 
 EXPOSE 53 53/udp


miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] Rebase coredns to..." ]
>Advantage of using Distroless comparing to empty baseimage.
>* Avoid the caveats when running go static binaries due to some missing non-binary dependencies.

We don't have that issue, coredns is staticly compiled and doesn't use 
/etc/nsswitch.

also installing ca-certificates has worked for ages, and doing the staged builds 
means the resulting image can be based on just scratch.

I don't know distroless; I don't know how to install extra dependencies should 
the need arise.

/Miek

--
Miek Gieben

yuwenma:
> [ Quoting <notifications@github.com> in "[coredns/coredns] Rebase coredns to..." ]
> Advantage of using Distroless comparing to empty baseimage. * Avoid the caveats when running go static binaries due to some missing non-binary dependencies.
> We don't have that issue, coredns is staticly compiled and doesn't use /etc/nsswitch. also installing ca-certificates has worked for ages, and doing the staged builds means the resulting image can be based on just scratch. I don't know distroless; I don't know how to install extra dependencies should the need arise. /Miek
> […](#)
> -- Miek Gieben

How about the reason of making the dockerfile nit? It's not a bad improvement and the coverage does increased :smile:

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Rebase coredn..." ]
>How about the reason of making the dockerfile nit? It's not a bad improvements though. (and the coverage does increased) :smile:

It makes us have to learn something we currently don't use, nor want. Pretty 
sure it will also increase the image's size.

We also want to run coredns non-root, no idea how I should go about doing this 
with distroless.

So NACK. sorry

/close

Issue #2779, opened 2019-04-10
Reverse zones should be calculated for the full CIDR range


mdgreenfield:
When specifying a reverse DNS zone using CIDR notation the full range of zones are not calculated.

For example, a CIDR of `10.129.60.0/22` currently only results in a single `60.129.10.in-addr.arpa.`, however, the full CIDR range should include zones for:

- `60.129.10.in-addr.arpa.`
- `61.129.10.in-addr.arpa.`
- `62.129.10.in-addr.arpa.`
- `63.129.10.in-addr.arpa.`

mdgreenfield:
I will see about putting together a PR to address this issue.

rdrozhdzh:
see related issue #2252

chrisohaver:
@miekg, in the README we say that CIDR zones work on non x8 boundaries (e.g. 25)

> This even works for CIDR (See RFC 1518 and 1519) addressing, i.e. 10.0.0.0/25, CoreDNS will then check if the in-addr request falls in the correct range.

Is this true?  I can't find where this check is done.

chrisohaver:
> For example, a CIDR of 10.129.60.0/22 currently only results in a single 60.129.10.in-addr.arpa., 

I think it would actually create `129.10.in-addr.arpa`, falling to /16, next lowest multiple of 8.

mdgreenfield:
In the beginning of the logs I'm seeing `60.129.10.in-addr.arpa.:53` for `10.129.60.0/22`

chrisohaver:
So, it casts a smaller net?  Thats broken...

mdgreenfield:
It looks like https://github.com/ynikanchyk/coredns/pull/2 addresses this issue.

I updated `core/dnsserver/address_test.go` with the example I provided (`10.129.60.0/22`) and found the actual result (`dns://60.129.10.in-addr.arpa.:53`) using `master`. 

I ran the same test with https://github.com/ynikanchyk/coredns/pull/2 and found that generated the following, which looks correct to me:

- `dns://60.129.10.in-addr.arpa.:53`
- `dns://61.129.10.in-addr.arpa.:53`
- `dns://62.129.10.in-addr.arpa.:53`
- `dns://63.129.10.in-addr.arpa.:53`

ynikanchyk:
[...]

PR #2783, opened 2019-04-12
Generate reverse zones for the full CIDR range

mdgreenfield:
This pull request calculates the full reverse zones for a given CIDR block. It is based on https://github.com/ynikanchyk/coredns/pull/2 although there were some small changes that needed to be updated to get all tests passing.

https://github.com/coredns/coredns/issues/2779
https://github.com/ynikanchyk/coredns/pull/2

Possibly will want to update https://coredns.io/2016/12/19/writing-plugins-for-coredns/

There is a lot that changed in this PR. Most of the changes affect tests and plugins but it should only affect reverse zones, not forward zones. And it should make it so that the full reverse zone is calculated instead of, in some cases, just a partial.


plugin/dnssec/setup.go
```diff
@@ -57,12 +57,27 @@ func dnssecParse(c *caddy.Controller) ([]string, []*DNSKEY, int, bool, error) {
 		}
 		i++
 
-		// dnssec [zones...]
-		zones = make([]string, len(c.ServerBlockKeys))
-		copy(zones, c.ServerBlockKeys)
-		args := c.RemainingArgs()
-		if len(args) > 0 {
-			zones = args
+		//// dnssec [zones...]
```
bradbeam: Is this intentionally commented?

bradbeam (approved):
lgtm, just a minor nit around some commented code

plugin/dnssec/setup.go
mdgreenfield: nope, thanks for catching that

mdgreenfield:
I'm working on an edge case for the following scenario where the PTR record is not successful:

```
10.129.60.0/22:0 10.131.192.0/18:0 {
	whoami
}
```

miekg:
please split this up and also don't change *all* setup functions. Instead add (a few) new tests that specifically test for the new behaviour (right now this is duplicated into all plugins)

.gitignore
```diff
@@ -13,3 +13,4 @@ coverage.txt
 .classpath
 .project
 .settings/**
+.vscode
```
miekg: this shouldn't be part of this pr

core/dnsserver/address.go
```diff
@@ -31,12 +31,12 @@ func (z zoneAddr) String() string {
 
 // normalizeZone parses a zone string into a structured format with separate
 // host, and port portions, as well as the original input string.
```
miekg: comment prolly needs update for reverse zone handling

go.sum
```diff
@@ -35,6 +35,7 @@ github.com/eapache/go-xerial-snappy v0.0.0-20180814174437-776d5712da21/go.mod h1
 github.com/eapache/queue v1.1.0 h1:YOEu7KNc61ntiQlcEeUIoDTJ2o8mQznoNvUhiigpIqc=
 github.com/eapache/queue v1.1.0/go.mod h1:6eCeP0CKFpHLu8blIFXhExK/dRa7WDZfr6jVFPTqq+I=
 github.com/estesp/manifest-tool v0.9.0/go.mod h1:w/oandYlJC/m8nkP8UaJVxsm/LwjurJQHXR27njws74=
+github.com/evanphx/json-patch v4.1.0+incompatible h1:K1MDoo4AZ4wU0GIU/fPmtZg7VpzLjCxu+UwBD1FvwOc=
```
miekg, line 38: if possible revert changes to these files as well (I know there are done automatically by go build now)

plugin/grpc/setup.go
```diff
@@ -69,7 +69,8 @@ func parseGRPCStanza(c *caddyfile.Dispenser) (*GRPC, error) {
 	if !c.Args(&g.from) {
 		return g, c.ArgErr()
 	}
-	g.from = plugin.Host(g.from).Normalize()
+	// We don't expect Normalize to expand
```
miekg: why not? Because it will not be a non-8-bit boundary zone because the setup function does not allow that?

plugin/metrics/metrics.go
```diff
@@ -75,6 +75,16 @@ func (m *Metrics) AddZone(z string) {
 	m.zoneMu.Unlock()
 }
 
+// AddZones adds zones from zs to m.
+func (m *Metrics) AddZones(zs []string) {
```
miekg: why is this needs? As it's just a loop around AddZone?

plugin/normalize.go
```diff
@@ -61,69 +61,119 @@ type (
...
+func SplitHostPort(s string) (hosts []string, port string, ipnet *net.IPNet, err error) {
 	// If there is: :[0-9]+ on the end we assume this is the port. This works for (ascii) domain
 	// names and our reverse syntax, which always needs a /mask *before* the port.
 	// So from the back, find first colon, and then check if its a number.
-	host = s
-
 	colon := strings.LastIndex(s, ":")
 	if colon == len(s)-1 {
-		return "", "", nil, fmt.Errorf("expecting data after last colon: %q", s)
+		return []string{""}, "", nil, fmt.Errorf("expecting data after last colon: %q", s)
```
miekg: `[]string{""}` can just be `nil`?

plugin/template/template_test.go
```diff
@@ -324,7 +324,7 @@ func TestMultiSection(t *testing.T) {
 		# Implicit section (see c.ServerBlockKeys)
 		# test.:8053 {
 
-	  # REFUSE IN A for the server zone (test.)
+		# REFUSE IN A for the server zone (test.)
```
miekg: unrelated whitespace change?

miekg (changes requested):

miekg:
Why can't this be done with the `FilterFunc`? https://github.com/coredns/coredns/blob/master/core/dnsserver/register.go#L82

plugins might not accept that or something?

plugin/grpc/setup.go
mdgreenfield: I think I misread the GRPC plugin docs. It looks like `FROM` supposed to be the `ZONE`. I'll get this fixed up.

plugin/template/template_test.go
mdgreenfield: changed to a tab during some testing, but will remove because nothing else was changed in this file.

go.sum
mdgreenfield: `go` keeps adding these even when i remove them from my changes. i'm still new to go and its ecosystem so if there is something I can do let me know.

mdgreenfield (commented):
> Why can't this be done with the FilterFunc? https://github.com/coredns/coredns/blob/master/core/dnsserver/register.go#L82
> 
> plugins might not accept that or something?

FilterFunc is not currently passed to the plugin. However, my understanding is that the FilterFunc will filter requests to a given server. But plugins can be zone specific, therefore, each plugin that is "zone aware" needs to normalize each zone as well so it can determine whether or not to handle the query. The query that is passed to a plugin is of the form `0.0.0.10.in-addr.arpa.` so doing something more clever with CIDR block contains isn't currently possible AFAICT.

Regarding the FilterFunc, I changed that to address an edge case. I've commented on the relevant change in the files.

> please split this up and also don't change all setup functions. Instead add (a few) new tests that specifically test for the new behavior (right now this is duplicated into all plugins)

If I'm completely off about the above can you point me in the right direction? AFAICT each plugin that supports zones will have to be altered.

core/dnsserver/register.go
```diff
@@ -78,18 +82,14 @@ func (h *dnsContext) InspectServerBlocks(sourceFile string, serverBlocks []caddy
 			}
 
 			ones, bits := za.IPNet.Mask.Size()
-			if (bits-ones)%8 != 0 { // only do this for non-octet boundaries
-				cfg.FilterFunc = func(s string) bool {
```
mdgreenfield, line 82: I removed this closure because the variable `za` was being set to the last `zoneAddr` seen in `InspectServerBlocks` due to the lexical scoping of variables within a closure.

miekg:
I'm looking into partially applying this PR and figuring some smaller fixes myself. First impression is that is becomes messy to do cleanly. I.e. expanding the zones in `InspectServerBlocks` is too late, so either this can be done (cleanly) earlier in the process, so that plugin also care less; things just work, or not at all; meaning that non octet boundaries was a cool plan, but doesn't works

mdgreenfield:
Thanks Miek, I've had some things come up and haven't been able to spend much time with this over the past week or so. Hoping to get back to it in the next week or two. Happy to adjust this PR as needed, if needed. Definitely agree that processing the cidr blocks earlier in the process is ideal to keep the plugins out of it as much as possible.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Generate reve..." ]
>Thanks Miek, I've had some things come up and haven't been able to spend much time with this over the past week or so. Hoping to get back to it in the next week or two. Happy to adjust this PR as needed, if needed. Definitely agree that processing the cidr blocks earlier in the process is ideal to keep the plugins out of it as much as possible.

Ack. I'm thinking Caddy should do this actually. The ServerBlockKeys 
(github.com/mholt/caddy/caddy.go) need to be expanded when a line
is read.

This can either be done directly in Caddy (means raising an issue there and see 
if they want to add that), or by moving to a custom Caddyloader - although for 
the latter case i'm not sure if you're exanding the list of zones at the right 
time.

Either way; some kind of generic callback *before* Caddy sees the list would be 
nice here; we can then just implement that then.

/Miek

--
Miek Gieben

mdgreenfield:
Following the conversation in https://github.com/mholt/caddy/issues/2596 it looks like I'm already adding the non-octal boundary reverse zones in the right spot (in `InspectServerBlocks`). But unfortunately, this PR still touches a lot of plugins because a plugin may be scoped to a subset of the serverblock reverse zone. `normalize.Normalize()` now returns a `[]string` so I think we either need to update all the plugins to support that or provide a `normalize.Normalize()` method that adheres to the previous contract (i.e. return `string`). Doing the later however, would result in similar edge cases as this PR is trying to solve.

Neither option is great. Any thoughts on where to go from here?

chrisohaver:
How about the suggestion discussed [starting here](https://github.com/coredns/coredns/issues/2779#issuecomment-483772569).

As you mentioned in that discussion, it has the caveat of not handling query count metrics properly. But perhaps this is an acceptable caveat given how minimal the code changes would be in comparison.

mdgreenfield:
I put together some very rough changes and ran through some tests to test rounding up to the nearest octal boundary. That does seem to work, however, more than just metrics are affected. The following configurations are invalid and even if the configurations didn't throw an error, all reverse queries would be resolved by the first server block.

```
192.168.1.0/22:0 192.168.4.0/22:0 {
	whoami
}
```
or
```
192.168.1.0/22:0  {
	whoami
}
192.168.4.0/22:0 {
	whoami
}
```

Result in an error like, `cannot serve dns://168.192.in-addr.arpa.:0 - it is already defined` (see https://github.com/coredns/coredns/blob/118b0c940890161d185b64497605a7ef84c38a0a/core/dnsserver/register.go#L205).

chrisohaver:
> The following configurations are invalid ...

Ah, yes of course.  The server would have to feed the query back "up" to be processed by the next server somehow - which I imagine would be difficult to pull off - and would defeat the purpose of it being a simpler solution.  Thanks for testing it out.

miekg:
Think to resolve this I would like to see:
* a PR that changes register.go and address.go that calculates the correct boundary for reverse + a test. This should also extend the `serverBlocks.Keys` slice to include the new zones. 

That should at least make plugins that don't call `Matches` be able to receive the query for the correct zone.

* another PR to extend `Matches` to implement reverse matching in a way that has minimal impact in the forward matching to keep that fast.

mdgreenfield:
> - a PR that changes register.go and address.go that calculates the correct boundary for reverse + a test. This should also extend the serverBlocks.Keys slice to include the new zones.
> 
> That should at least make plugins that don't call Matches be able to receive the query for the correct zone.

Makes sense. However, we'll still need to update `plugin/normalize.go` at a minimum with a new `Normalize()` func that returns a slice, I assume you'd rather keep a func that returns a string to maintain backwards compatibility so that we can then have separate followup PRs to update each/all plugin(s)?

> - another PR to extend Matches to implement reverse matching in a way that has minimal impact in the forward matching to keep that fast.

Right now, plugins maintain a single list of all zones so this will also require each plugin be updated such that the `setup` function maintains separate forward/reverse zones (I assume I'd probably use a heuristic like suffix match on `in-addr.arpa` to separate forward/reverse zones). But updating each plugin is something that could be done in followup PR(s) when `Normalize() string` gets updated to `Normalize() []string`. One thing not great is that plugin development would now probably want to strongly recommend keeping separate forward/reverse zones, which is not great.

With all these changes to each plugin, I suspect we'll want to get all the changes for all the plugins into a minor, possibly major, release?

(You can find me in the CoreDNS slack if you want to do a call to go over this in more detail.)

cpointner:
Any progress on this? If lack of time to code slows things down... we might be able to throw a coding intern (or two) at the problem...
However if i understand this correctly in order to fix the issue a breaking change for plugins is needed? In this case i personally would go with the rather-now-than-later approach... IMHO being able to freely configure classless reverse lookups that just work is totally worth it.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Generate reve..." ]
>Any progress on this? If lack of time to code slows things down... we might be able to throw a coding intern (or two) at the problem...
>However if i understand this correctly in order to fix the issue a breaking change for plugins is needed? In this case i personally would go with the rather-now-than-later approach... IMHO being able to freely configure classless reverse lookups that just work is totally worth it.

There are some requirements: 

* the plugin coode must most not require changes
* it may not slow down the forward path resolution
* preferably a small PR

cpointner:
Ok but what does this mean? Is anybody working on that? Is help wanted?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Generate reve..." ]
>Ok but what does this mean? Is anybody working on that? Is help wanted?

There also another point to consider which may warrant leaving the current impl 
(or ripping it all out):

A corefile like this:

~~~
org com {
  cache
  whoami
}
~~~

is identical to 

~~~
org {
  cache
  whoami
}

com {
  cache
  whoami
}
~~~

Ie. two (internal) servers and thus 2 caches, having something like 
192.168.1.0/17 expand automatically to multiple entries, hides this and make it 
even more un-expected.

cpointner:
@miekg: Sure but as long as the zones are not overlapping the 2 caches cannot contain the same entry twice. Also the way how caches are implemented isn't very predictable either... it is always possible that names get expunged from the cache before it is full isn't it?

miekg:
closing this as it has too many conflicts and discussion seems to have stalled. If we can't find a workable solution for v4 dns zones I rather remove this feature entirely

PR #2789, opened 2019-04-17
fallthrough support for forward plugin

achinthagunasekara:
fallthrough support for forward plugin

```
test {
    forward . 127.0.0.1:8600 {
        force_tcp
    }
    cache 300
}
```

In this situation, a plugin handles the query, but the reply it got from its backend (i.e. maybe it got NXDOMAIN) is such that it wants other plugins in the chain to take a look as well. If fallthrough is provided (and enabled!), the next plugin is called. A plugin that works like this is the hosts plugin. First, a lookup in the host table (/etc/hosts) is attempted, if it finds an answer, it returns that. If not, it will fallthrough to the next one in the hope that other plugins may return something to the client.

However looking at `forward` plugin, it doesn't seem to have support for `fallthrough` - https://coredns.io/plugins/forward/

This PR adds `fallthrough` support to `forward` plugin.

Not an issue. New feature

`forward` plugin documentation

Yes

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -50,6 +50,7 @@ forward FROM TO... {
     tls_servername NAME
     policy random|round_robin|sequential
     health_check DURATION
+    fallthrough [ZONES...]
 }
 ~~~
 
@@ -83,6 +84,10 @@ forward FROM TO... {
   * `round_robin` is a policy that selects hosts based on round robin ordering.
   * `sequential` is a policy that selects hosts based on sequential ordering.
 * `health_check`, use a different **DURATION** for health checking, the default duration is 0.5s.
+* `fallthrough` If zone matches and no record can be generated, pass request to the next plugin.
+  If **[ZONES...]** is omitted, then fallthrough happens for all zones for which the plugin
+  is authoritative. If specific zones are listed (for example `in-addr.arpa` and `ip6.arpa`), then only
+  queries for those zones will be subject to fallthrough.
 
 Also note the TLS config is "global" for the whole forwarding proxy if you need a different
 `tls-name` for different upstreams you're out of luck.
@@ -142,6 +147,17 @@ Proxy everything except `example.org` using the host's `resolv.conf`'s nameserve
 . {
     forward . /etc/resolv.conf {
         except example.org
+    }
+}
+~~~
+
+Load example.hosts file and only serve example.org and example.net from it and fall through to the
+next plugin if query doesn't match.
+
+~~~
+. {
+    hosts example.hosts example.org example.net {
+        fallthrough
     }
 }
 ~~~
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -65,7 +65,7 @@ func (f *Forward) Name() string { return "forward" }
 func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 
 	state := request.Request{W: w, Req: r}
-	qname := state.Name()
+	queryName := state.Name()
 	if !f.match(state) {
 		return plugin.NextOrFailure(f.Name(), f.Next, ctx, w, r)
 	}
@@ -153,10 +153,6 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 			formerr.SetRcode(state.Req, dns.RcodeFormatError)
 			w.WriteMsg(formerr)
 
-			if f.fall.Through(qname) {
-				return plugin.NextOrFailure(f.Name(), f.Next, ctx, w, r)
-			}
-
 			return 0, taperr
 		}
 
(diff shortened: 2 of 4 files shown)


miekg:
The host plugin is special because it's can't actually reply with NXDOMAIN
because that how /etc/hosts works (in conjunction with /etc/nswitch.conf).
It needs fall through.

No other plugin should have to do this and most certainly not the forward
plugin.

So NACK from me.

rdrozhdzh:
consider using plugin alternate
https://github.com/coredns/alternate

achinthagunasekara:
@miekg @rdrozhdzh 
My current issue is, my upstreams sometimes return `NXDOMAIN` for valid queries. In this situation I'd like to get the result from the cache.

Something like,

```
test {
    forward . 127.0.0.1:8600 {
        force_tcp
        fallthrough
    }
    cache 300
}
```

Is there any other way to do this?

Thanks

chrisohaver:
You may want to look into why your upstreams sometimes return NXDOMAIN for valid queries.  Is this something under your control? Is it an intentional behavior?  Perhaps there is an alternate solution that follows DNS standards.

That aside... 

Taken literally, you could do what you are asking for by setting your cache up to not cache negative responses.  Doing this means that if there is a cache entry, then it will not be an NXDOMAIN.  In the event that there is no cache entry, or it has expired, then there is no cache entry to use.

achinthagunasekara:
@chrisohaver

is `cache` plugin has any configuration that we can use to turn off caching of `NXDOMAIN` responses? I don't see anything like that in https://coredns.io/plugins/cache/

chrisohaver:
```
cache {
    denial 0
}
```

achinthagunasekara:
@chrisohaver What I'm trying to archive is,

1. Check the cache. If a non-expired response exists, return it.
2. If no non-expired response, ask upstream. If it returns a response, cache it and return it.
3. If no upstream response, check the cache for an expired response, return it.
4. If no expired response, return NXDOMAIN.

chrisohaver:
IMO, step 3, returning expired entries from cache, is too obscure of a behavior to add as an option to cache.

Why does your upstream “return NXDOMAIN sometimes for valid queries?” Is this something you control?

achinthagunasekara:
@chrisohaver 
Sorry about the delay getting back to you.
We don't have any control over the upstream, that was the reason we were looking into this solution.
Any suggestions on how I can proceed? :)

chrisohaver:
> Any suggestions on how I can proceed? :)

Maybe you could rewrite TTLs, making them really long so they stay in the cache for a really long time ... (see the _rewrite_ plugin).   It's definitely "hack" territory, but you're kind of already there if you're looking to serve expired entries from cache.

stp-ip:
How about cache + prefetch + denial cache disabled. Currently the denial cache can't be disabled afaik. But this might be something that could be supported via "denial 0".
Thoughts?

chrisohaver:
>How about cache + prefetch + denial cache disabled.

I suggested this [here](https://github.com/coredns/coredns/pull/2789#issuecomment-484511717) - sans `prefetch`.  I'm not seeing how `prefetch` helps solve the issue at hand though.

> Currently the denial cache can't be disabled afaik. But this might be something that could be supported via "denial 0".
> Thoughts?

I think negative cache can already be disabled like [this](https://github.com/coredns/coredns/pull/2789#issuecomment-484703225). IIRC it was enabled a couple of releases ago.

The TTL rewrite suggestion is in addition to no negative caching.  I.e. make the positive answers live in the cache for a very long time (but never put negatives in the cache).

stp-ip:
With prefetch the idea is to extend the time a positive answer is within the cache. But maybe that's not perfect.

Ah yeah read through another issue today (just checked it was before the mentioned release) that mentioned setting denial to 0 doesn't.

chrisohaver:
> I think negative cache can already be disabled like this. IIRC it was enabled a couple of releases ago.

Hmm. Looking at code , I don't see where we check for zero, and we seem to enforce an undocumented minimum size cache.  I keep tripping over this.  I think we should allow it.

chrisohaver:
Ah, This is why I thought it was done already... 

https://github.com/coredns/coredns/pull/2588

... it stalled 2 months ago.

achinthagunasekara:
@chrisohaver, @stp-ip  

What if we change the config to something like below, we can set `AMOUNT` to fairly low and `DURATION` fairly high. This is not going to solve the problem entirely, but should help?

```
test {
    cache 300 {
        prefetch 2 60 50%
    }
    forward . 127.0.0.1:8600 {
        force_tcp
    }
}
```

miekg:
closing this as it will not be merge and it seems to indicate some upstream problem

m-yosefpor:
@achinthagunasekara  I think this is what you are looking for https://coredns.io/explugins/fallback/

> The fallback plugin allows an alternate set of upstreams be specified which will be used if the plugin chain returns specific error message

```
. {
    forward . 127.0.0.1:8600 {
        force_tcp
    }
    fallback NXDOMAIN . 8.8.8.8 {
        protocol dns force_tcp
    }
    cache 300
}
```

PR #2793, opened 2019-04-22
Remove -cpu flag

miekg:
The -cpu flag is a weird one (and copied originally from Caddy), it
basically sets GOMAXPROCS which can be *easily* done by just setting
that environment variable. Also with systemd and containerized env you
set this externally *anyway*, so there is little use to do this again in
the binary.

Also the option's help was confusing (i.e. percentage of what?). Remove
the option and supporting files.


chrisohaver:
Do we need to follow deprecation policy here?  Or is it safe to assume that no one is using this flag?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Remove -cpu f..." ]
>Do we need to follow deprecation policy here?  Or is it safe to assume that no one is using this flag?

dunno, who's using this?

miekg:
and then we just move to 1.6 for the next release, problem solved.

To power of semver to trigger discussion is so annoying, that I'm ready to drop the whole idea #sigh

chrisohaver:
> and then we just move to 1.6 for the next release, problem solved.

Jumping to 1.6 to avert the deprecation policy is a dubious loop hole.  We haven't announce the flag as deprecated in a release, nor have we made it a no op in a release following the deprecation announcement.  So it doesn't follow our policy as written.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Remove -cpu f..." ]
>> and then we just move to 1.6 for the next release, problem solved.
>
>Jumping to 1.6 to avert the deprecation policy is a dubious loop hole.  We haven't announce the flag as deprecated in a release, nor have we made it a no op in a release following the deprecation announcement.  So it doesn't follow our policy as written.

you're making my point. Thank you #sigh

miekg:
No comments after 4 day, interest looks low, so I'll remove this. Means next release is 1.6.0; so we can maybe pile in some other incompatible changes. We already have the resyncperiod one in k8s.

bengadbois:
I'm on a team currently using the `-cpu` flag, so if this is merged, a deprecation notice would be appreciated

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Remove -cpu f..." ]
>I'm on a team currently using the `-cpu` flag, so if this is merged, a deprecation notice would be appreciated

Thanks! That implies the next release will probably be 1.6.0, unless we need 
some smaller fixes sooner (which will then be 1.5.2, with -cpu still in it)

PR #2794, opened 2019-04-22
plugin/chaos: randomize author list

miekg:
Randomize the author list on request; keep the zowners.go file stable so
a 'go generate' remain stable.

chaos.Owners could potentially be a map and be randomized by ranging
over it, but this seems simpler and fewer lines of code.

Bit of Easter hacking; seems more fair to randomize this list.


yongtang (approved):
LGTM

Issue #2790, opened 2019-04-17
Workaround for plugin/loop on Kubernetes not working


jforman:
Hi all,

I'm trying to diagnose a failing coredns deploy to my bare-metal kubernetes setup that fails on detecting a loop. I have followed the directions at https://coredns.io/plugins/loop/#troubleshooting, and specified my resolv conf in the kubelet config file, but something must be not quite right.

Any help is greatly appreciated.

Stats:

- Ubuntu 18.04 LTS libvirt/kvm VM
- Kubernetes 1.14.0
- Coredns 1.5.0

CoreDNS pod log:
```
.:53
2019-04-17T00:18:44.485Z [INFO] plugin/reload: Running configuration MD5 = fe280f2b52a55a24ba57f2b4328c2777
2019-04-17T00:18:44.485Z [INFO] CoreDNS-1.5.0
2019-04-17T00:18:44.485Z [INFO] linux/amd64, go1.12.2, e3f9a80
CoreDNS-1.5.0
linux/amd64, go1.12.2, e3f9a80
2019-04-17T00:18:44.486Z [FATAL] plugin/loop: Loop (127.0.0.1:34614 -> :53) detected for zone ".", see https://coredns.io/plugins/loop#troubleshooting. Query: "HINFO 6074102630077018525.5798689655824409358.
```

Pod description:
```
Name:               coredns-85767f454b-b5fcq
Namespace:          kube-system
Priority:           0
PriorityClassName:  <none>
Node:               devb-worker0.domain.obfuscated.net/10.10.2.133
Start Time:         Tue, 16 Apr 2019 20:17:17 -0400
Labels:             k8s-app=kube-dns
                    pod-template-hash=85767f454b
Annotations:        container.apparmor.security.beta.kubernetes.io/coredns: runtime/default
                    kubernetes.io/psp: 999-privileged
                    seccomp.security.alpha.kubernetes.io/pod: docker/default
Status:             Running
IP:                 10.244.3.10
Controlled By:      ReplicaSet/coredns-85767f454b
Containers:
  coredns:
    Container ID:  containerd://1e051fb5ae5d59f185996c4791d454d3b6359220cbda8f2925812aead8bb00c0
    Image:         coredns/coredns:1.5.0
    Image ID:      docker.io/coredns/coredns@sha256:e83beb5e43f8513fa735e77ffc5859640baea30a882a11cc75c4c3244a737d3c
    Ports:         53/UDP, 53/TCP, 9153/TCP
[...]

PR #2804, opened 2019-04-29
plugin/loop: Update troubleshooting step

chrisohaver:

Looks like `kubelet` has deprecated command flags, preferring options be specified in its config yaml file. This updates the corresponding troubleshooting step in *loop*.

#2790

### DIFF
--- a/plugin/loop/README.md
+++ b/plugin/loop/README.md
@@ -80,7 +80,7 @@ requests to itself.
 
 There are many ways to work around this issue, some are listed here:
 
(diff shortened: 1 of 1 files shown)


plugin/loop/README.md
```diff
@@ -80,7 +80,7 @@ requests to itself.
 
 There are many ways to work around this issue, some are listed here:
 
-* Add the following to `kubelet`: `--resolv-conf <path-to-your-real-resolv-conf-file>`.  Your "real"
+* Add the following to your `kubelet` config yaml: `ResolverConfig: <path-to-your-real-resolv-conf-file>` (or set it with the deprecated command flag `--resolv-conf`).  Your "real"
```
miekg: can you add in which version this was deprecated?

chrisohaver: done - it was deprecated quite some time ago ... in 1.10.

PR #2810, opened 2019-05-04
New option in forward plugin to retry requests

m-pavel:
Added new option to forward plugin to retry request to next server in case of given responses. It makes sense with `sequential` policy. For example, if zone have public and private DNS servers and they resolve same fqdn to different addressees (external and internal) and public server gives NXDOMAIN on internal resources that can be resolved by internal server. It possible to define `retry_on NXDOMAIN` and have prioritized public addresses and internal if only internal available.

See above.
None, it's a FR.
Already updated README.md
No.

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -50,6 +50,7 @@ forward FROM TO... {
     tls_servername NAME
     policy random|round_robin|sequential
     health_check DURATION
+    retry_on RESPONCES...
 }
 ~~~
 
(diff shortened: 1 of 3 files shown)


miekg:
I'm not inclined to merge this, because the behavior doesn't make a lot of sense from a DNS perspective. What are you trying to solve?

stp-ip:
I second that. It feels like this is working around a faulty upstream server.

m-pavel:
Ok, may be it's quite specific scenario:
Some organization's servers are available in internal and public networks, some in internal only. There is public DNS server which resolve external addresses of public servers but for internal resources it says NXDOMAIN. There is internal DNS servers that resolve public servers into internal addresses.
There are clients, (e.g. connected via VPN) that need to access public servers via external addresses and internal servers via internal addresses.

miekg:
This can also fail when responses are cached, because you will just get the first one (this PR doesn't create that failure mode, but this kind of code will eventually lead to that). Split horizon DNS is just a bad idea in general.

PR #2814, opened 2019-05-05
plugin/ready: fix starts and restarts

miekg:
Add OnRestartFailed to the ready plugin and some various cleanups.

Document slightly better how things are supposed to work with multiple
`ready`'s in the multiple Server Blocks.

All manually tested with this Corefile:
~~~
. {
    log
    ready
}

example.org {
    log
    chaos
    ready
}
~~~
And then `kill -SIGUSR1` and curling the ready endpoint. This works
well, the FailedReload is triggered by adding a syntax error in the
Corefile.

See #2659


miekg:
this should be good to go

PR #2817, opened 2019-05-05
plugin/autopath docs: remove last use of middleware

miekg:
This has somehow survived; the docs still used Middleware instead of
Plugins.


miekg:
too quick :)

Issue #2426, opened 2019-01-02
plugin/forward: forward to tls self-hosted coredns server not work


littleplus:
### Environment:
Both of client and server are in ```Centos 7 x86_64``` with ```CoreDNS 1.3.0```

### Problem:
I could resolve the hostname with ```kdig +tls @server```. But, when I use CoreDNS's plugin forward at client, it could not work.

when I use kdig , there is log in server show with noerror. when I use coredns forward, there's no any log show in server.

#### Client:
```
google.com {
    forward . tls://server{
        force_tcp
    }
    log
}
```
#### Server (the cert and ca cert are self-signed with ip as its hostname):
```
tls://.:853 {
        tls /etc/coredns/coredns.crt /etc/coredns/coredns.key /etc/coredns/ca.crt
        hosts /etc/coredns/hosts {
                fallthrough
        }
        log
        forward . 1.1.1.1 8.8.8.8
}
```

When I use client coredns forward's log and there is no any log show in server:
```
2019-01-02T23:09:37.286+08:00 [INFO] CoreDNS-1.3.0
2019-01-02T23:09:37.287+08:00 [INFO] linux/amd64, go1.11.4, c8f0e94
CoreDNS-1.3.0
linux/amd64, go1.11.4, c8f0e94
2019-01-02T23:09:40.828+08:00 [INFO] 10.0.1.2:60901 - 13307 "A IN www.google.com. udp 32 false 512" SERVFAIL qr,rd 32 1.210567836s
2019-01-02T23:09:40.832+08:00 [INFO] 10.0.1.2:60901 - 13307 "A IN www.google.com. udp 32 false 512" SERVFAIL qr,rd 32 0.213627591s
2019-01-02T23:09:40.855+08:00 [INFO] 10.0.1.2:20849 - 13307 "A IN www.google.com. udp 32 false 512" SERVFAIL qr,rd 32 0.02644381s
```

I try to use tcpdump to debug, but I have no idea:
Success with kdig tcpdump in client:
```
22:40:35.330685 IP client.61308 > server.853: Flags [S], seq 608450174, win 29200, options [mss 1460,sackOK,TS val 4139669014 ecr 0,nop,wscale 9], length 0
22:40:35.345340 IP server.853 > client.61308: Flags [S.], seq 2705609430, ack 608450175, win 63443, options [mss 1424,nop,nop,TS val 3854911046 ecr 4139669014,nop,wscale 6,nop,nop,sackOK], length 0
[...]

PR #2825, opened 2019-05-13
make sure client CA and auth type are set if CA is explicitly specified.

jinmeiib:

The current implementation of the tls plugin doesn't require client
certificate even if the CA parameter of the plugin configuration is
explicitly specified.  And, as a result, client authentication is
impossible.

It's not very clear whether it's intentional, but it doesn't seem to
make sense to allow the user to specify "CA" if the server doesn't
even request a client certificate.  So I assume at least the first
point is unintentional.

The change in this pull request is one easy way to fix it.  It now also
makes sure the client certificate is validated with the specified CA;
if it doesn't validate, the server won't accept the connection.

There may be other way to address these problems.  The change in this
PR is rather to demonstrate the issue and its cause.

No.  I've not filed a separate github issue on this.

README.md of the tls plugin will probably have to be updated to
clarify the effect of parameter 'CA'.  I'd also note that the current
description about the use of system CAs can be misleading, since the
server doesn't request client's certificate or validate it if
parameter 'CA' is omitted.

Yes, depending on the user's expectation.  Previously, even if
parameter CA was specified and the client wasn't configured with a
valid certificate, TLS connections from the client would be accepted.
Such connection attempts will now be rejected.

If this compatibility issue is a concern, we may have to extend the
configuration syntax to have another parameter that specifies the
certification validation policy.


johnbelamaric:
I agree this is necessary. The original intent was to validate the client cert, and so this is really a bug fix. However, I am concerned about two things:

1) The change in behavior is dramatic.
2) We should offer a way to have the system CAs validate the cert, this change won't allow that.

So...how about instead we:

1) Add a `client_auth` option with values corresponding to [tls.ClientAuthType](https://golang.org/pkg/crypto/tls/#ClientAuthType) `none` (default), `request`, `require`, `verify_if_given`, `require_and_verify`.
2) This would be added in version 1.6.0 - include a release note that in version 1.7.0, the default will change to `require_and_verify` if the CA is configured.

Then, what I'd really like to see (separate PR) is if `metadata` is enabled, then add some TLS-related metadata that can be used in logging and firewall/policy plugins, see: https://github.com/coredns/coredns/issues/2426#issuecomment-454141924. That makes `request` meaningful, because we can log - otherwise it's useless.

jinmeiib:
It makes sense to introduce an option to specify the client auth policy (as noted in the PR description I was aware of the concern).  Would you like me to update the code in this PR as such?  If so, do you mean the syntax of the tls plugin now looks like this?

```
tls CERT KEY [CA] {
  [client_auth none|request|require|verify_if_given|require_and_verify]
}
```

As for the `metadata` thing, that's actually another trouble I had while trying to use client authentication.  For now I have a workaround of extracting the TLS connection state from `dns.ResponseWriter`, but providing a generic way makes a lot more sense.

johnbelamaric:
Yes, that is correct, please implement that syntax change. Thanks!!

jinmeiib:
I've just pushed some new commits with the suggested changes.  I'd appreciate it if you could review it.

jinmeiib:
Hi @johnbelamaric 
If you have time, could you re-review the latest change?  I've realized a CI test failure, but it doesn't seem to be related to my change.  If there's anything else I should have done and you could point it out, I'd appreciate it very much. (I'm not particularly in a hurry, but I wanted to make sure I've done what I was supposed to do).

johnbelamaric:
/lgtm
/approve

Please though add a release note indicating that the default will change in 1.7

PR #2827, opened 2019-05-16
[test/responsewriter] provide mechanism to specify custom Remote IP a…

chantra:
…ddress and number of message written

This is useful to build test when the behaviour may be expected based on client IP (Geo IP and the likes)

GetWriteMsgCallCount() make it easier to validate if WriteMsg was called or not.

request/request_test.go contains a test validating the custom IP.
plugin/whoami/whoami_test.go test was modified to validate the
GetWriteMsgCallCount() function.

This PR allow customizing the source IP of a test.ResponseWriter

None

None that I know.

No

### DIFF
--- a/plugin/test/responsewriter.go
+++ b/plugin/test/responsewriter.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"net"
+	"sync/atomic"
 
 	"github.com/miekg/dns"
 )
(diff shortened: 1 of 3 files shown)


chantra:
`TestMetricsAvailableAfterReloadAndFailedReload` seems to get caught by the race condition detection 50% of the time, even on master so I don't think this diff is responsible for it, neither that c52d7c3fdd7e671dd614cc659c49d5a2c88ab787 is needed.

miekg:
I can see a need for this, but I'm not a fan of overloading this test writer, but rather see a completely new struct implementing this.

chantra:
Fine by me. My original implementation, the one I use, is actually a new
struct. Let me put it in a gist later.

chantra:
https://gist.github.com/chantra/cd8776fc38f9554c0055bdde56162c6a
was my original, and case-specific, approach.

Are you fine with this?
Also, do you agree `writeMsgCallCount` should be left in the original `ResponseWriter` struct?

chantra:
test failure is basically #2831

@miekg do you want something similar to https://gist.github.com/chantra/cd8776fc38f9554c0055bdde56162c6a ?

miekg:
from the looks of it you want to *just* inject a responsewriter for a test. If you generalize this it's gets a bit hazy, because you're injecting a response writer into a response writer which is the same as just a response writer.

I think it's better to leave this impl to a plugin that needs it and then if useful for other generalize it

chantra:
that's fair. let's close this then.

PR #2828, opened 2019-05-16
pkg/log: reset the buffer in the tests

miekg:
Reset the buf otherwise we're not checking the new value.


miekg:
Have a fix for the data race; this is non-controversial, so merging

PR #2829, opened 2019-05-17
Publish metadata from kubernetes plugin

johnbelamaric:
This pull request implements metadata.Provider in the Kubernetes plugin. It publishes the following metadata:
  * kubernetes/endpoint: the endpoint name in the query
  * kubernetes/kind: the resource kind (pod or svc) in the query
  * kubernetes/namespace: the namespace in the query
  * kubernetes/port-name: the port name in an SRV query
  * kubernetes/protocol: the protocol in an SRV query
  * kubernetes/service: the service name in the query
  * kubernetes/client-namespace: the client pod's namespace, if `pods verified` mode is enabled
  * kubernetes/client-pod-name: the client pod's name, if `pods verified` mode is enabled

With this and the `firewall` plugin, you could, for example limit pods to only see services in their own namespaces.

None.

README is updated as part of this PR

No.

### DIFF
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -225,3 +225,17 @@ or the word "any"), then that label will match all values.  The labels that acce
 *.service.default.svc.cluster.local. 5	IN A	192.168.25.15
 ```
  This response can be randomized using the `loadbalance` plugin
+
+## Metadata
+
+The kubernetes plugin will publish the following metadata, if the metadata
+plugin is also enabled:
+
+ * kubernetes/endpoint: the endpoint name in the query
+ * kubernetes/kind: the resource kind (pod or svc) in the query
+ * kubernetes/namespace: the namespace in the query
+ * kubernetes/port-name: the port name in an SRV query
+ * kubernetes/protocol: the protocol in an SRV query
+ * kubernetes/service: the service name in the query
+ * kubernetes/client-namespace: the client pod's namespace, if `pods verified` mode is enabled
+ * kubernetes/client-pod-name: the client pod's name, if `pods verified` mode is enabled
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -228,7 +228,7 @@ or the word "any"), then that label will match all values.  The labels that acce
 
 ## Metadata
 
-The kubernetes plugin will publish the following metadata, if the metadata
+The kubernetes plugin will publish the following metadata, if the _metadata_
 plugin is also enabled:
 
  * kubernetes/endpoint: the endpoint name in the query
--- a/plugin/kubernetes/metadata.go
+++ b/plugin/kubernetes/metadata.go
@@ -1,0 +1,58 @@
+package kubernetes
+
+import (
+	"context"
+
+	"github.com/coredns/coredns/plugin/metadata"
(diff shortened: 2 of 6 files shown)


johnbelamaric:
@chrisohaver @miekg @rajansandeep

chrisohaver:
Thanks John.  If we pull this data from the client-go cache instead of the query string, that would allow us to create metadata for service ports for non-SRV requests for example.  It's tough to extract interesting data from the trimmed down version of the client-go objects, but ... could also expose the number of ready endpoints of a service (perhaps not very useful).

If the client-go objects were not trimmed, we could expose other information such as pod node, and status.

johnbelamaric:
> Thanks John. If we pull this data from the client-go cache instead of the query string, that would allow us to create metadata for service ports for non-SRV requests for example. It's tough to extract interesting data from the trimmed down version of the client-go objects, but ... could also expose the number of ready endpoints of a service (perhaps not very useful).

But to do that, we have to lookup the service object, right? That means doing the whole lookup (which maybe is OK, especially if we hang on to it and reuse it). Is there a method that just returns the resource reference by the query?

> If the client-go objects were not trimmed, we could expose other information such as pod node, and status.

True, true. If we find enough valuable info in there, maybe we can have an option to use "full size" objects. There will also be a need for node info if we support the topology-based headless services.

chrisohaver:
Ah, never mind - I see now how its difficult due to how metadata works in relation to the plugin chain and other plugins.

chrisohaver (approved):
lgtm

plugin/kubernetes/README.md
```diff
@@ -225,3 +225,17 @@ or the word "any"), then that label will match all values.  The labels that acce
 *.service.default.svc.cluster.local. 5	IN A	192.168.25.15
 ```
  This response can be randomized using the `loadbalance` plugin
+
+## Metadata
+
+The kubernetes plugin will publish the following metadata, if the metadata
```
chrisohaver:
```suggestion
The kubernetes plugin will publish the following metadata, if the _metadata_
```

I think most of the time we _italicize_ plugin names when referring to the plugin.  So here, first metadata not italicized, second one italicized. Admittedly, we are not 100% consistent throughout all the READMEs (case in point `loadbalance` in the section above should be _loadbalance_).

plugin/kubernetes/metadata.go
```diff
@@ -0,0 +1,59 @@
+package kubernetes
+
+import (
+	"context"
+
+	"github.com/coredns/coredns/plugin/metadata"
+	"github.com/coredns/coredns/request"
+)
+
+// Metadata implements the metadata.Provider interface
```
miekg: nit: closing dot

plugin/metadata/metadata.go
```diff
@@ -20,10 +20,15 @@ type Metadata struct {
 // Name implements the Handler interface.
 func (m *Metadata) Name() string { return "metadata" }
 
+// ContextWithMetadata is exported for use by provider tests
```
miekg, line 23: Is this useful enough to add here?

johnbelamaric: Yes, without this you can't do a real unit test, you have to do more of an end-to-end test which is a lot more painful.

PR #2832, opened 2019-05-17
Add external mdns plugin

trown:
This adds a plugin that reads mDNS records from the local network
and responds to queries based on those records. It is useful for
providing mDNS records to non-mDNS-aware applications by making
them accessible through a standard DNS server.

### DIFF
--- a/plugin.cfg
+++ b/plugin.cfg
@@ -39,6 +39,7 @@ any:any
 chaos:chaos
 loadbalance:loadbalance
 cache:cache
+mdns:github.com/openshift-metal3/coredns-mdns
 rewrite:rewrite
 dnssec:dnssec
 autopath:autopath


trown:
I tried to follow the doc about adding external plugins [1](https://coredns.io/2017/07/23/add-external-plugins/) but it doesnt look like content/blog/explugins exists. Do I need to put it in content/explugins [2](https://github.com/coredns/coredns.io/tree/master/content/explugins)? If so, I can update the doc as well.

plugin.cfg
```diff
@@ -39,6 +39,7 @@ any:any
 chaos:chaos
 loadbalance:loadbalance
 cache:cache
+mdns:github.com/openshift-metal3/coredns-mdns
```
miekg, line 42: Thought this would be an entire plugin :) We don't merge external plugins like this. It might be worth while to file an PR on coredns/coredns.io to list this there as an external plugin (I could do that as well - if you like)

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add external ..." ]
>I tried to follow the doc about adding external plugins [1](https://coredns.io/2017/07/23/add-external-plugins/) but it doesnt look like content/blog/explugins exists. Do I need to put it in content/explugins[2](https://github.com/coredns/coredns.io/tree/master/content/explugins)? If so, I can update the doc as well.

Yep, content/explugins should be the spot.


Thanks!

polarathene:
Was this ever merged to the explugins? I don't see any related file there. Would have loved to have seen this on the sites plugin list, only just found out about it via google.

The URL now redirects to https://github.com/openshift/coredns-mdns

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add external ..." ]
>Was this ever merged to the explugins? I don't see any related file there.
>Would have loved to have seen this on the sites plugin list, only just found
>out about it via google.
>
>The URL now redirects to https://github.com/openshift/coredns-mdns

afaict there was never a pull request for coredns.io for this plugin

polarathene:
I have built CoreDNS with the external plugin successfully and got it to work. 

In my case mDNS wasn't exactly a zeroconfig solution, I had to configure avahi-daemon for each machine to publish it's hostname on the local network for the plugin to recognize them. I then ran into the issue of receiving the wrong IP due to multiple interfaces on a machine causing a race condition for the hostname (bridge network, wireless NIC, Docker, etc), so config needs to be update `allow-hosts` to the interface you want to advertise for the hostname.

Additionally, I assumed that I could use subdomains in with mDNS for a reverse proxy to manage. Either this requires additional configuration with Avahi or CoreDNS, or it's a limitation of mDNS.

---

The plugin does work, I'm not sure how useful it is to others given that experience though. Configuring each system to just route DNS through CoreDNS is probably less hassle.

PR #2840, opened 2019-05-24
Add server instance to the context in ServerTLS and ServerHTTPS

ameshkov:
The problem with the current code is that there's no way to get the server instance inside a plugin.
Because of that "metrics" plugin sets empty "server" label for requests served over TLS or HTTPS.


miekg:
Good find! Should this change also be applied to `server_grpc.go` line 132?

ameshkov:
Yeah, I guess it makes sense to have it in every server implementation

PR #2845, opened 2019-05-27
Update dns to v1.1.12, and remove duplicate dependencies

yongtang:
This fix:
1. Update dns to v1.1.12
2. Redirect dns v1.1.3 in caddy to v1.1.12 so that there is only one version of dns.
3. Run `go mod tidy` to prune unneeded dependencies.

n/a
n/a
n/a


yongtang:
Note `replace github.com/miekg/dns v1.1.3 => github.com/miekg/dns v1.1.12` line is used to override v1.1.3 used in caddy.

Issue #2842, opened 2019-05-25
Kubernetes autopath plugin bug - CoreDNS fails to resolve A records under circumstances


zendai:
# CoreDNS versions

All versions of CoreDNS are affected, including the latest one.

# Configuration

Sample configuration
```
.:53 {
    errors
    health
    kubernetes cluster.local. in-addr.arpa ip6.arpa {
      pods verified
      upstream
    }
    prometheus :9153
    proxy . /etc/resolv.conf
    loop
    cache 30
    loadbalance
    reload
    autopath @kubernetes
}
```

Relevant parts from the bug point of view are:
 - cache plugin
 - kubernetes plugin
 - autopath plugin using @kubernetes plugin

Relevant notes about the OS:
 - IPv6 enabled, so DNS requests are both for A and AAAA

# Failing component

Kubernetes plugin, [autopath](https://github.com/coredns/coredns/blob/v1.5.0/plugin/kubernetes/autopath.go) section.

# Symptom

Occasionally A record resolution fails within cluster. If you use curl in your Kubernetes cluster for whatever reason (many SDKs do), then you'll see sporadic `curl: (6) Could not resolve host: <hostname>`.

# Description

The scenario is fairly complex and took a few days to pin down the exact reason behind the symptom. Won't go deep here about the investigation, will present only the result which will make sense even without the investigation.

The root cause of the problem is that autopath is applied only if the first entry of the source pod's search path is equal or ends with the entry we want to resolve. This is checked [here](https://github.com/coredns/coredns/blob/v1.5.0/plugin/autopath/autopath.go#L88) using this function [here](https://github.com/coredns/coredns/blob/master/plugin/autopath/autopath.go#L150).

This search path function is implemented in modules which support autopath, including the Kubernetes plugin. The Kubernetes plugin is responsible to build the search path list based on the source pod of the request, this is what is used to "figure out" what the source Pod's search path looks like. This function is [here](https://github.com/coredns/coredns/blob/v1.5.0/plugin/kubernetes/autopath.go#L11).

This function [looks up](https://github.com/coredns/coredns/blob/v1.5.0/plugin/kubernetes/autopath.go#L29) the source IP of the Pod, then it reaches out to Kubernetes to find out which namespace it belongs to based on its IP address. Using that, it'll rebuild how the source Pod's search path looks like.

[...]

PR #2846, opened 2019-05-27
Fix for #2842, instead of returning the first Pod, return the one whi…

zendai:
…ch is Running

This fix is about the Kubenetes autopath module which occasionally identifies the source Pod incorrectly. It returns the first Pod entry in the list based on an IP search, which occasionally will be incorrect as we can have historical Pods with that IP in that list as well. If the first Pod in the list is a historical one, the autpath module will work with incorrect information.

This fix is returning a Pod object only if it's Running (which there should only be one, per IP). In order to do that I had to add a new field which I called Phase (after the original field in the Kubernetes spec, to remain consistent), and which represent the status of the Pod.

As a result now we look up the source Pod not just by IP, but also by status to make sure we get the correct one.
 

This is a bugfix for #2842.

Issue #2842.

None.

No.

### DIFF
--- a/plugin/kubernetes/object/pod.go
+++ b/plugin/kubernetes/object/pod.go
@@ -12,7 +12,7 @@ type Pod struct {
 	Name      string
 	Namespace string
 	Deleting  bool
-        Phase     string
+        Running   bool
 
 	*Empty
 }
@@ -29,12 +29,17 @@ func ToPod(obj interface{}) interface{} {
 		PodIP:     pod.Status.PodIP,
 		Namespace: pod.GetNamespace(),
 		Name:      pod.GetName(),
-                Phase:     string(pod.Status.Phase),
 	}
 	t := pod.ObjectMeta.DeletionTimestamp
 	if t != nil {
 		p.Deleting = !(*t).Time.IsZero()
 	}
+        s := string(pod.Status.Phase)
+        if s == "Running" {
+                p.Running = true
+        } else {
+                p.Running = false
+        }
 
 	*pod = api.Pod{}
 
--- a/plugin/kubernetes/object/pod.go
+++ b/plugin/kubernetes/object/pod.go
@@ -12,7 +12,6 @@ type Pod struct {
 	Name      string
 	Namespace string
 	Deleting  bool
-        Running   bool
 
 	*Empty
 }
@@ -21,6 +20,10 @@ type Pod struct {
 func ToPod(obj interface{}) interface{} {
 	pod, ok := obj.(*api.Pod)
 	if !ok {
+		return nil
+	}
+	s := string(pod.Status.Phase)
+	if s != "Running" {
 		return nil
 	}
 
@@ -34,12 +37,6 @@ func ToPod(obj interface{}) interface{} {
 	if t != nil {
 		p.Deleting = !(*t).Time.IsZero()
 	}
-        s := string(pod.Status.Phase)
-        if s == "Running" {
-                p.Running = true
-        } else {
-                p.Running = false
-        }
 
 	*pod = api.Pod{}
 
--- a/plugin/kubernetes/object/pod.go
+++ b/plugin/kubernetes/object/pod.go
@@ -22,10 +22,6 @@ func ToPod(obj interface{}) interface{} {
 	if !ok {
 		return nil
 	}
-	s := string(pod.Status.Phase)
-	if s != "Running" {
-		return nil
-	}
 
 	p := &Pod{
 		Version:   pod.GetResourceVersion(),
--- a/plugin/kubernetes/object/pod.go
+++ b/plugin/kubernetes/object/pod.go
@@ -12,6 +12,7 @@ type Pod struct {
 	Name      string
 	Namespace string
 	Deleting  bool
+        Phase     string
 
 	*Empty
 }
@@ -28,6 +29,7 @@ func ToPod(obj interface{}) interface{} {
 		PodIP:     pod.Status.PodIP,
 		Namespace: pod.GetNamespace(),
 		Name:      pod.GetName(),
+                Phase:     string(pod.Status.Phase),
 	}
 	t := pod.ObjectMeta.DeletionTimestamp
 	if t != nil {
--- a/plugin/kubernetes/autopath.go
+++ b/plugin/kubernetes/autopath.go
@@ -52,10 +52,8 @@ func (k *Kubernetes) AutoPath(state request.Request) []string {
 // podWithIP return the api.Pod for source IP ip. It returns nil if nothing can be found.
 func (k *Kubernetes) podWithIP(ip string) *object.Pod {
 	ps := k.APIConn.PodIndex(ip)
-        for _, p := range ps {
-          if p.Running {
-            return p
-            }
-        }
-	return nil
+	if len(ps) == 0 {
+		return nil
+	}
+	return ps[0]
 }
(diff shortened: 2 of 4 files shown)


yongtang:
@zendai Looks like the email address you used in git history (`A...@im...ve.com`) is not registered with GitHub. For that reason, the commit you contributed is not showing your account picture.

It is certainly up to you to decide one way or another. Though if you want the commit you contributed to be associated to you, you may consider link the email in GitHub (or use a already linked email when you commit the code).

zendai:
@yongtang thank you for the note, I wasn't even thinking about this. I added the email address to my account, I think that's the simpler solution for now. Next time I'll be smarter! (maybe)

plugin/kubernetes/object/pod.go
```diff
@@ -12,6 +12,7 @@ type Pod struct {
 	Name      string
 	Namespace string
 	Deleting  bool
+        Phase     string
```
miekg: ah, this is *very* expensive to add because it increases memory. Can't we use `Deleting` or figure out something there?

miekg (changes requested):
adding `Phase` is too expensive, esp considering autopath is sort of a hack

plugin/kubernetes/object/pod.go
zendai: @miekg Based on my limited understanding of the codebase (I just reviewed it today for the first time), this is the easiest way to fix this. If memory is a concern, instead of storing the status itself, we could store a bool to indicate whether it's running or not. What do you think?

The field will be less useful but will use less memory.

zendai: Btw, the status is about 5-6 characters long (not sure how much underlying memory this actually means though), is that really too much? 🙄

zendai: Let me check if we can use Deleting.

zendai: Can confirm, Deleting will be false for all the Pods, including the historical ones. We can't use Deleting here.

zendai:
We can't use the Deleting field, will submit a new version where we only use a bool instead of string to represent if the Pod is running. This is more CPU cycles but less memory.

zendai:
@miekg I rewrote it using bool instead of string. That's my best bet, if you want to optimise this further we have to make structural changes at many points.

zendai:
@miekg about "*esp considering autopath is sort of a hack*". Could you please elaborate what that means and what the consequences are?

Please don't take this the wrong way but I couldn't find any sign in the documentation that suggests that autopath is a hack, according to the docs it's as much of a plugin as any other. Add to this that CoreDNS is the default DNS for Kubernetes and it was recommended in the official Kubernetes blog (https://kubernetes.io/blog/2018/07/10/coredns-ga-for-kubernetes-cluster-dns/) just increases my confusion what hack means. 

Should we not use it in production? Is it not supported as much as the other plugins?

johnbelamaric:
How about just not keeping pods unless they are running?

No need to add anything to the struct, and we will actually save memory.

zendai:
As I see there are two ways not to keep Pods unless they are running.

 - Use the current Kubernetes API query mechanism and post process the results, which would still require a field in the struct (pointless).
 - Filter the Pods at the Kubernetes API query level, in which case we don't need the field in the struct, because Kubernetes will handle the filtering.

This second version is somewhere [here](https://github.com/coredns/coredns/blob/58c703f5ef212e3c9efbe31fa4d9332eb268ff0d/plugin/kubernetes/controller.go#L217). Potentially we could use a [FieldSelector](https://github.com/kubernetes/apimachinery/blob/7dcd37fca5430ef614dac1c0609605a3ec30336e/pkg/apis/meta/internalversion/types.go#L35) as part of the Kubernetes client options.

Also, this is theoretical as I have no in-depth knowledge of the codebase and I'm not sure about the consequences. Can we confirm that adding such a filter at the Kubernetes query layer won't have any negative side effect at any other component?

The current version only have one extra bool per Pod. Memory-wise that must be barely visible, even if we talk about tens of thousands of Pods. Does it make sense to go with a higher risk and deeper structural change to save a few kbytes, maybe tens of kbytes memory? 

Just asking, I honestly don't have internal knowledge of the code base, you know better than I do.

miekg:
Sounds like an excellent plan to me and we fix this bug as well. So win win

bradbeam:
> How about just not keeping pods unless they are running?

Would that enable us to drop p.Deleting? It looks like we're already withholding returning a response if p.Deleting == true.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fix for #2842..." ]
>> How about just not keeping pods unless they are running?
>
>Would that enable us to drop p.Deleting? It looks like we're already withholding returning a response if p.Deleting == true.

But running is not including pending is it? Think this may be a step too far.

Not adding pod with p.Deleting == true seems like another optimization we can 
make that should be backwards compatible.

zendai:
@miekg @johnbelamaric Please review this proposal. This won't have the extra field in struct, it'll return nil if the Pod is not Running at the conversion function `ToPod`. We already had a condition which returns nil from that function if there is an error fetching the Pod, so I assume it's safe to return nil.

Tested a few times, it works.

A more elegant solution would be what I mentioned before, ask Kubernetes to filter Pods using ` FieldSelector`, but that's a bigger change. That would go to two places, the [Pod listing](https://github.com/coredns/coredns/blob/58c703f5ef212e3c9efbe31fa4d9332eb268ff0d/plugin/kubernetes/controller.go#L217) function and the [Pod watcher](https://github.com/coredns/coredns/blob/29cb00aada276ad99e4f3aef380de9e68f2ebae8/plugin/kubernetes/watch.go#L15) function.

Please let me know if you're fine with this version, or you want the more elegant version with the `FieldSelector`.

zendai:
@miekg @johnbelamaric @bradbeam My latest commit is a proposal where we configure `FieldSelector` at the Kubernetes API query so the filtering is completely done at the Kubernetes level. Also, if there is a `FieldSelector` already we just append our filter. This is the cleanest version so far.

If you're sure that shortlisting Pods at that level is safe, I think this should be fine.

Please review and let me know what you think.

chrisohaver:
The current proposed solution LGTM. (filtering at API list/watch)

johnbelamaric:
Can we keep Running and Pending, as suggested by @miekg? That way we get the Pod info earlier which will reduce the likelihood of race conditions.

zendai:
@miekg @johnbelamaric @bradbeam Changed the `FieldSelector` to include `Running` and `Pending`. 

Please note that the `FieldSelector` allows [only for logical AND](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/#chained-selectors), and not OR. This means we can't filter for "include `Running` OR `Pending`" (explicit), but we can achieve the same result by excluding all the others (implicit) which are `Succeeded`, `Failed` and `Unknown`.

The result is the same, we're filtering for Pods that are `Running` or `Pending`. Tested, it works.

The available Pod states today are listed [here](https://github.com/kubernetes/kubernetes/blob/release-1.15/staging/src/k8s.io/api/core/v1/types.go#L2389), for the reference. The potential risk is if Kubernetes introduces new Pod states, which won't be covered by our exclusion list, still this is the best we can do now if we want to include multiple states.

Please let me know what you think.

chrisohaver:
Agree - It makes sense to include every status that eventually leads to Running, which IIUC is Pending and Running.  @zendai, Thanks for linking to the pod status definitions.  LGTM

bradbeam (approved):

miekg (approved):

PR #2853, opened 2019-05-30
plugin/kubernetes: skip deleting pods

miekg:
Don't add pods to our internal cache that are being deleted. This saves
a field in the struct as well.

Add (extra) comments about adding fields to the
object/{Pod,Service,Endpoint} structs.

See #2846 for a discussion. No documentation changes are needed and this
should be backwards compatible.


chrisohaver:
Looks good...

Perhaps we should also `continue` if `obj` is nil in `NewIndexerInformer()` after calling `convert()`.
Not sure how well client-go handles adds/updates/delete on a nil object.

bradbeam (approved):

PR #2855, opened 2019-05-31
plugin/template: allow an empty response

chrisohaver:
Allows _template_ to produce a NODATA response.
Desire to do this has come up several times in the past, specifically to create a black hole for AAAA requests.  This would allow users to do that with ...

```
   template IN AAAA .
``` 

But in general, I see _template_ as a flexible hacking/debugging tool, and I think allowing a NODATA response makes it a bit more flexible.


miekg (approved):

miekg:
I'll go ahead and merge this.

Issue #32, opened 2016-03-20
Dedup and round_robin middleware


miekg:
In SkyDNS I have to dedup message and correct the order, i.e. CNAMEs before A records and other fluff. This can of course be _also_ handled as a middleware.

See #9 for etcd implementation.

miekg:
has been merged. We can do more stuff here, but for now this is Ok

Issue #30, opened 2016-03-20
middleware: caching


miekg:
a caching middleware

miekg:
Should we cache RRset or complete messages? The former is how you should do it, but what it that leads to a partial answer? Then you need to get fetch it from somewhere.. do you then invoke the next middleware? That could work.
Or you could cache complete answers, but then there is a problem of multiple clients with different buffer sizes (or even transport), i.e. you can't re-use a TCP message when a client comes asking via UDP.

The first approach makes most sense, it's RFC compliant, and if you find that answer incomplete just hand off to the next middleware.

miekg:
question still is... when is an answer incomplete.... In this example it can be detected. `www.miek.nl` is a cname to `a.miek.nl`:

```
;; ANSWER SECTION:
www.miek.nl.        1800    IN  CNAME   a.miek.nl.
a.miek.nl.      1765    IN  A   139.162.196.78
```

So if I have the CNAME, but not the target, which is inbaliwick I should just hand hand off...?

PR #2866, opened 2019-06-05
be sure to close connection after completion of xfr out

jinmeiib:

It makes sure that the 'file' plugin closes the TCP connection when it responds to a zone transfer request.  Otherwise the connection and associated socket will stay in the CLOSE_WAIT state unless/until golang runtime performs GC (which is not documented, but it closes the socket when a TCP connection object is garbage collected).  This is the case even if the client closes the connection (as the original code comment suggests).  Maybe the original code assumed it's enough to let the GC do the job, but I think it's better to close it explicitly since sockets and file descriptors are a different kind of system resource than memory, and can sometimes be more scarce.

I've not filed a corresponding github issue.  It's easy to reproduce by running coredns with the file plugin and having it respond to an xfr request, and then checking the connection state using tools like netstat or lsof.

I don't think this requires a documentation change, as the effect is mostly invisible to users. (but if the proposed change and its rationale make sense, you may want to update the usage example for miekg/dns.Transfer.Out()).

I believe the new behavior backward compatible.  I realize that the connection can't accept any subsequent requests if we close it after xfr, but the original implementation can't either anyway, since it "hijacks" the writer.

### DIFF
--- a/plugin/file/xfr.go
+++ b/plugin/file/xfr.go
@@ -33,7 +33,10 @@ func (x Xfr) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (in
 	ch := make(chan *dns.Envelope)
 	defer close(ch)
 	tr := new(dns.Transfer)
-	go tr.Out(w, r, ch)
+	go func() {
+		tr.Out(w, r, ch)
+		w.Close()
+	}()
 
 	j, l := 0, 0
 	records = append(records, records[0]) // add closing SOA to the end
@@ -51,7 +54,6 @@ func (x Xfr) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (in
 	}
 
 	w.Hijack()
-	// w.Close() // Client closes connection
 	return dns.RcodeSuccess, nil
 }
 


miekg:
Thanks!

I think this needs some clarification in miekg/dns as well - I'll dig that up

miekg:
I think this is actually not needed. Do you have evidence it stays open, because it will (also) be closed in the scrubwriter?

This question is triggered by the data-race this introduces:
~~~
==================
WARNING: DATA RACE
Write at 0x00c0000a2a80 by goroutine 32:
  github.com/miekg/dns.(*response).Close()
      /home/miek/g/pkg/mod/github.com/miekg/dns@v1.1.12/server.go:734 +0xc7
  github.com/coredns/coredns/request.(*ScrubWriter).Close()
      <autogenerated>:1 +0x62
  github.com/coredns/coredns/plugin/file.Xfr.ServeDNS.func1()
      /home/miek/go/coredns/coredns/plugin/file/xfr.go:38 +0x74

Previous read at 0x00c0000a2a80 by goroutine 30:
  github.com/miekg/dns.(*Server).serveTCPConn()
      /home/miek/g/pkg/mod/github.com/miekg/dns@v1.1.12/server.go:518 +0x3fc

Goroutine 32 (running) created at:
  github.com/coredns/coredns/plugin/file.Xfr.ServeDNS()
      /home/miek/go/coredns/coredns/plugin/file/xfr.go:36 +0x58c
  github.com/coredns/coredns/plugin/file.File.ServeDNS()
      /home/miek/go/coredns/coredns/plugin/file/file.go:79 +0x13c0
  github.com/coredns/coredns/plugin/file.(*File).ServeDNS()
      <autogenerated>:1 +0x10e
  github.com/coredns/coredns/core/dnsserver.(*Server).ServeDNS()
      /home/miek/go/coredns/coredns/core/dnsserver/server.go:242 +0xbae
  github.com/coredns/coredns/core/dnsserver.(*Server).Serve.func1()
      /home/miek/go/coredns/coredns/core/dnsserver/server.go:102 +0xdc
  github.com/miekg/dns.HandlerFunc.ServeDNS()
      /home/miek/g/pkg/mod/github.com/miekg/dns@v1.1.12/server.go:37 +0x51
  github.com/miekg/dns.(*Server).serveDNS()
      /home/miek/g/pkg/mod/github.com/miekg/dns@v1.1.12/server.go:602 +0x4fe
  github.com/miekg/dns.(*Server).serveTCPConn()
      /home/miek/g/pkg/mod/github.com/miekg/dns@v1.1.12/server.go:517 +0x3ee

Goroutine 30 (finished) created at:
  github.com/miekg/dns.(*Server).serveTCP()
      /home/miek/g/pkg/mod/github.com/miekg/dns@v1.1.12/server.go:434 +0x33a
  github.com/miekg/dns.(*Server).ActivateAndServe()
      /home/miek/g/pkg/mod/github.com/miekg/dns@v1.1.12/server.go:337 +0x29c
  github.com/coredns/coredns/core/dnsserver.(*Server).Serve()
      /home/miek/go/coredns/coredns/core/dnsserver/server.go:106 +0x200
  github.com/mholt/caddy.startServers.func1.1()
      /home/miek/g/pkg/mod/github.com/mholt/caddy@v1.0.0/caddy.go:795 +0x86
==================
--- FAIL: TestSecondaryZoneTransfer (0.01s)
    testing.go:809: race detected during execution of test
~~~

jinmeiib:
> I think this is actually not needed. Do you have evidence it stays open, because it will (also) be closed in the scrubwriter?

Without my change it's not closed in the ScrubWriter in the parent goroutine (the one calling ServeDNS()) because it's hijacked.  As described in the description of this PR you should be able to see the "evidence" quite easily.  For example, run coredns with this configuration:
```
. {
    file example.org.zone {
        transfer to *
    }
}
```
(with some valid example.org zone file) and keep doing this:

```
lsof -n -p `pgrep coredns` | grep CLOSE_WAIT
```
and then perform xfr:
```
dig @127.0.0.1 example.org axfr
```
then you'll keep seeing output like this
```
coredns 40496 jtatuya    9u    IPv6 0x4fedb927f2253b47      0t0         TCP 127.0.0.1:domain->127.0.0.1:59397 (CLOSE_WAIT)
```
for some period (how long is not fully predictable - it depends on when the next GC takes place).

As for the race from the test, I can't reproduce it locally after a few tries, but I suspect what happened is this:

1. First, the child goroutine (#32) completes its xfr and (with my change) closes the writer.  It changes `w.closed` to true at line 734 of server.go:
```
func (w *response) Close() error {
	if w.closed {
		return &Error{err: "connection already closed"}
	}
	w.closed = true  // l.734
```

2. Then the parent goroutine (#30) completes the `ServeDNS()` chain and returns to `serveTCPConn()` and checks `w.closed` at line 518:
```
		if w.closed {  // l. 518
			break // Close() was called
		}
```

3. race checker detects these two events happening without explicit synchronization and complains about it.

If my guess is correct, this "race" is actually harmless since `w` has been hijacked by the parent itself and the parent doesn't use it anymore (other than checking `hijacked`).  But to make it cleaner we should probably change the order of checks for `closed` and `hijacked` in `serveTCPConn()`:

```
		if w.hijacked {
			break // the "hijacker" is responsible for calling Close().  Note that we should check 'hijacked' first; otherwise 'closed' is referenced/updated by two different goroutines, which can be considered a race condition (albeit harmless).
		}
		if w.closed {
			break // Close() was called
		}
```

miekg:
Ah we hijack. Thanks your explanation makes sense. Do we need to hijack in
the first place though?

miekg:
Might actually highlight a design flaw in miekg/dns. We can't close here
because of the race.

jinmeiib:
> Do we need to hijack in the first place though?

I actually wondered about that, too, when I noticed this issue.  I wasn't fully sure, but I was afraid that not hijacking would lead to a real race condition: the parent goroutine reads a new query with `w` and write a response to it, while the child is still writing its response with the same `w`.  So my tentative conclusion was that hijack is likely to be necessary.

> We can't close here because of the race.

If you refer to my original patch by "close" and the race reported in the test, I suspect this race doesn't do a real harm (and if we still don't like it can be avoided by switching the two checks) as I noted in my followup comment.  But I admit it may be less clean and probably more fragile against future changes.

Anyway, this is something I just happened to notice.  It's not urgent for me.  If you rather want to defer it until the miekg/dns design is revisited, please feel free to do so.

miekg:
We might actually want to close all conns explicitly...

PR #2870, opened 2019-06-06
plugin/kubernetes: Use target namespaces in endpoint fqdns

chrisohaver:

Endpoints can target Pods in other namespaces.  Although this doesn't happen normally when endpoints are automatically created based on Service selectors, it can occur when an endpoint is manually created.  

Since each target of an endpoint can be in a different namespace, this means we have to store the target's namespace in addition to the name in our custom object.  This increases memory footprint of the endpoint cache.

If the added memory is a concern, one strategy we could use to reduce memory in the k8s object cache could be to maintain a list of enumerated namespaces, which would allow us to replace all namespace strings in our custom objects with integers.

@johnbelamaric discovered this issue when looking into https://github.com/kubernetes/kubernetes/issues/47992

none?  Probably too esoteric to come up in the readme.

no

### DIFF
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -292,7 +292,7 @@ func endpointFQDN(ep *object.Endpoints, zone string, endpointNameMode bool) []st
 	var names []string
 	for _, ss := range ep.Subsets {
 		for _, addr := range ss.Addresses {
-			names = append(names, dnsutil.Join(endpointHostname(addr, endpointNameMode), serviceFQDN(ep, zone)))
+			names = append(names, dnsutil.Join(endpointHostname(addr, endpointNameMode), ep.GetName(), addr.TargetRefNamespace, Svc, zone))
 		}
 	}
 	return names
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -292,7 +292,13 @@ func endpointFQDN(ep *object.Endpoints, zone string, endpointNameMode bool) []st
 	var names []string
 	for _, ss := range ep.Subsets {
 		for _, addr := range ss.Addresses {
-			names = append(names, dnsutil.Join(endpointHostname(addr, endpointNameMode), ep.GetName(), addr.TargetRefNamespace, Svc, zone))
+			var namespace string
+			if addr.TargetRefNamespace != "" {
+				namespace = addr.TargetRefNamespace
+			} else {
+				namespace = ep.Namespace
+			}
+			names = append(names, dnsutil.Join(endpointHostname(addr, endpointNameMode), ep.GetName(), namespace, Svc, zone))
 		}
 	}
 	return names
--- a/plugin/kubernetes/object/endpoint.go
+++ b/plugin/kubernetes/object/endpoint.go
@@ -27,10 +27,11 @@ type EndpointSubset struct {
 
 // EndpointAddress is a tuple that describes single IP address.
 type EndpointAddress struct {
-	IP            string
-	Hostname      string
-	NodeName      string
-	TargetRefName string
+	IP                 string
+	Hostname           string
+	NodeName           string
+	TargetRefName      string
+	TargetRefNamespace string
 }
 
 // EndpointPort is a tuple that describes a single port.
@@ -75,6 +76,7 @@ func ToEndpoints(obj interface{}) interface{} {
 			}
 			if a.TargetRef != nil {
 				ea.TargetRefName = a.TargetRef.Name
+				ea.TargetRefNamespace = a.TargetRef.Namespace
 			}
 			sub.Addresses[j] = ea
 		}
--- a/plugin/kubernetes/kubernetes_test.go
+++ b/plugin/kubernetes/kubernetes_test.go
@@ -130,7 +130,7 @@ func (APIConnServiceTest) EpIndex(string) []*object.Endpoints {
 			Subsets: []object.EndpointSubset{
 				{
 					Addresses: []object.EndpointAddress{
-						{IP: "172.0.0.1", Hostname: "ep1a"},
+						{IP: "172.0.0.1", Hostname: "ep1a", TargetRefNamespace: "testns"},
 					},
 					Ports: []object.EndpointPort{
 						{Port: 80, Protocol: "tcp", Name: "http"},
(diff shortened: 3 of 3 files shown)


chrisohaver:
This raises at least one academic DNS issue relating to NODATA vs NXDOMAIN.
for example...

A service `svc1.ns1.svc.cluster.local` can have a endpoint `ep1.svc1.ns2.svc.cluster.local` ...
but `svc1.ns2.svc.cluster.local` may not exist...
but since `ep1.svc1.ns2.svc.cluster.local` does exist, then `svc1.ns2.svc.cluster.local` should be a NODATA, not NXDOMAIN.

plugin/kubernetes/kubernetes.go
```diff
@@ -292,7 +292,13 @@ func endpointFQDN(ep *object.Endpoints, zone string, endpointNameMode bool) []st
...
 	for _, ss := range ep.Subsets {
 		for _, addr := range ss.Addresses {
-			names = append(names, dnsutil.Join(endpointHostname(addr, endpointNameMode), serviceFQDN(ep, zone)))
+			var namespace string
+			if addr.TargetRefNamespace != "" {
+				namespace = addr.TargetRefNamespace
+			} else {
+				namespace = ep.Namespace
+			}
+			names = append(names, dnsutil.Join(endpointHostname(addr, endpointNameMode), ep.GetName(), namespace, Svc, zone))
```
miekg, line 301: This seems to remove the only used of serviceFQDN? If so we can get rid of that as well

plugin/kubernetes/object/endpoint.go
```diff
@@ -27,10 +27,11 @@ type EndpointSubset struct {
...
 type EndpointAddress struct {
-	IP            string
-	Hostname      string
-	NodeName      string
-	TargetRefName string
+	IP                 string
+	Hostname           string
+	NodeName           string
+	TargetRefName      string
+	TargetRefNamespace string
```
miekg, line 34: this is not cheap....

chrisohaver: Yes I know. I cover this in the description.

miekg:
>Yes I know. I cover this in the description.

I see. I think this is good change (after this is merged)

plugin/kubernetes/kubernetes.go
chrisohaver: Hmm - actually - I'm guessing things got refactored at some point in the past, and _ALL_ of these FQDN functions were abandoned/orphaned ... Only used in tests.

So, obviously this PR is invalid.  I'll close it...

chrisohaver:
Closing this PR.  It seems that none of the service/endpoint/podFQDN functions are used anymore.  Im guessing due to a past refactor or something.

chrisohaver:
i'll open another PR to remove the functions.

johnbelamaric:
That wasn't the right solution anyway. My description was not sufficient (or maybe even incorrect).  See https://github.com/kubernetes/kubernetes/issues/47992#issuecomment-347196086

Actually this is more of a refinement than a bug fix, and _maybe_ we shouldn't do it.

The point was in order to use the pod name in `endpoint_pod_names`, the namespace of the pod MUST match the service namespace. Otherwise you could end up with a name conflict because there may be a pod name in the original namespace. But now thinking about it, maybe that's not an issue - we just may end up with two IPs for that name, one for the the Pod in our namespace and one for another pod. As long as this is driven by the Endpoints record I don't really see an issue with that. @thockin was the one that mentioned restricting it by namespace - any comment, Tim?

chrisohaver:
So... lets have ...
a Pod named `abc` in namespace `space1` serving port 80
a Pod named `abc` in namespace `space2` serving port 443

A headless service in namespace `space1` that targets both above Pods.

The service fqdn is `s1.space1.svc.cluter.local`...

If we do a SRV query on the service, currently we probably get something like  ...

anwser:
```
s1.space1.svc.cluter.local  IN  SRV  80  abc.s1.space1.svc.cluter.local
s1.space1.svc.cluter.local  IN  SRV  443 abc.s1.space1.svc.cluter.local
```
addtional:
```
abc.s1.space1.svc.cluter.local IN A  172.0.0.1
abc.s1.space1.svc.cluter.local IN A  172.0.0.2
```
Which is problematic. Name collision; ambiguity.

If we change the response to ... 

anwser:
```
s1.space1.svc.cluter.local  IN  SRV  80  abc.s1.space1.svc.cluter.local
s1.space1.svc.cluter.local  IN  SRV  443 abc.s1.space2.svc.cluter.local
                                                ^^^^^^
```
addtional:
```
abc.s1.space1.svc.cluter.local IN A  172.0.0.1
abc.s1.space2.svc.cluter.local IN A  172.0.0.2
```
This is less problematic ...  except that  query on`s1.space2.svc.cluter.local` would probably return NXDOMAIN, when it should be NODATA, because `abc.s1.space2.svc.cluter.local` exists.

Doing it this way will require adjustments to our Endpoint object indexes (since endpoints can be in more than one namespace)

chrisohaver:
The spec naming scheme kind of implicitly assumes that endpoints will be in the same namespace as the service.  For an endpoint name `abc.s1.space1.svc.cluter.local`, is `space1` the namespace of the service `s1` or the pod `abc`?  IMO, it's kind of implied that they are both in that namespace.

It would be less ambiguous to stick the namespace of the pod in there too... e.g. `abc.space2.s1.space1.svc.cluter.local`, where `s1` is in `space1`, and `abc` is in `space2`.  But this is a _major_ change to the schema... and ndots+=1... and obtrusive to users, to which the extra namespace would almost always appear to be redundant.

chrisohaver:
Using the target namespace in the fqdn of an endpoint record would also result in some very confusing wildcard query results.  Since in the example [above](https://github.com/coredns/coredns/pull/2870#issuecomment-499958341), `abc.s1.space2.svc.cluter.local` may not actually be targeted by `s1.space2.svc.cluter.local`, but a wild card query for `*.s1.space2.svc.cluter.local ` would return that endpoint.  IMO, the usefulness of wildcard queries are dubious in general, so probably not a big deal.

thockin:
The fact that they cross namespaces doesn't actually matter does it?  I can have 2 pods in the same NS that have teh same hostname and different values for a given port name.

I think this is a "you broke it, you bought it" situation.  If I use the same hostname on 2 pods in a single Service and they don't have the same serving ports, I can't really expect it to work.  We should explode as gracefully as we can muster, but it's probably not worth hand-wringing over.

Thoughts?

johnbelamaric:
Yes, I agree. I am thinking we just do nothing here. I think we need to keep the namespace in the DNS name for the endpoint the same as the service namepace, not the backing pod. Way to confusing otherwise.

PR #2873, opened 2019-06-07
plugin/kubernetes: Remove unused functions

chrisohaver:

Removes 3 unused functions, and their unit tests.


miekg (approved):

bradbeam (approved):

johnbelamaric:
I think these were here from when we did `watch`. I would like to bring that back someday...sigh.

chrisohaver:
> I think these were here from when we did `watch`. I would like to bring that back someday...sigh.

Ah, right.  The issue of cross namespace endpoints remains though.  I think I'll open an issue to discuss it.  It's not so straightforward.

PR #2883, opened 2019-06-11
plugin/cache: allow disabling of positive/negative cache

chrisohaver:

Allows users to specify zero for positive or negative cache to disable the corresponding cache.

This expands on #2588, ...
* skips initialization of a cache if disabled (zero length)
* adds unit test
* updates README

#2588 (stalled)

included

not realistically

### DIFF
--- a/plugin/cache/README.md
+++ b/plugin/cache/README.md
@@ -53,8 +53,12 @@ cache [TTL] [ZONES...] {
 
 ## Capacity and Eviction
 
-If **CAPACITY** _is not_ specified, the default cache size is 9984 per cache. The minimum allowed cache size is 1024.
-If **CAPACITY** _is_ specified, the actual cache size used will be rounded down to the nearest number divisible by 256 (so all shards are equal in size).
+If **CAPACITY** is zero, the corresponding cache will be disabled.
+
+If **CAPACITY** _is not_ specified, the default cache size is 9984 per cache.
+
+If **CAPACITY** _is_ specified, the actual cache size used will be rounded down to the nearest 
+number divisible by 256 (so all shards are equal in size). The minimum allowed cache size is 1024.
 
 Eviction is done per shard. In effect, when a shard reaches capacity, items are evicted from that shard.
 Since shards don't fill up perfectly evenly, evictions will occur before the entire cache reaches full capacity.
@@ -100,6 +104,16 @@ Enable caching for all zones, keep a positive cache size of 5000 and a negative 
      cache {
          success 5000
          denial 2500
+    }
+ }
+ ~~~
+
+Enable caching for all zones but disable negative caching:
+
+~~~ corefile
+ . {
+     cache {
+         denial 0
     }
  }
  ~~~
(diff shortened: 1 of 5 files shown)


miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin/cache: all..." ]
>
>Allows users to specify zero for positive or negative cache to disable the corresponding cache.

If you don't want to cache you can disable the cache plugin.


/Miek

--
Miek Gieben

chrisohaver:
> If you don't want to cache you can disable the cache plugin. /Miek

Disabling the cache plugin disables both positive and negative cache.  The use case here is to disable only one.  For example, disabling the negative cache.

stp-ip:
Looks good from my point of view. Reasonable addition in functionality. Still would like an additional sign off by Miek or another reviewer.

miekg:
updating the readme and adding test is all good. The original PR was only a couple of lines though; this adds more. I'm not fond of the cache type as we already had a `switch` in the `set` method anyway - i.e. the cache type adds an indirection that not directly needed.

chrisohaver:
The original PR initializes an empty cache of minimum size, which is a waste of memory.
This PR skips initializing the corresponding cache if it's not going to be used.
The original PR also wastes time searching disabled caches for each get.
This PR skips searching a cache if it is disabled.

plugin/cache/cache.go
```diff
@@ -201,23 +207,31 @@ func (w *ResponseWriter) WriteMsg(res *dns.Msg) error {
...
-
+		return posCache
 	case response.NameError, response.NoData, response.ServerError:
-		i := newItem(m, w.now(), duration)
-		w.ncache.Add(key, i)
-
+		return negCache
 	case response.OtherError:
-		// don't cache these
+		return noCache
```
miekg, line 217: Please remove this indirection; it's not needed and it bolts an indirection on an indirection

plugin/cache/cache.go
```diff
@@ -242,4 +256,9 @@ const (
 	Success = "success"
 	// Denial is the class defined for negative caching.
 	Denial = "denial"
+
+	// Cache types for indicating which cache a response belongs
+	noCache  = 0 // should not be cached
+	posCache = 1 // should go in positive cache
+	negCache = 2 // should go in negative cache
```
miekg, line 263: please remove

plugin/cache/README.md
```diff
@@ -103,3 +107,13 @@ Enable caching for all zones, keep a positive cache size of 5000 and a negative
     }
  }
  ~~~
+
+Enable caching for all zones but disable negative caching:
+
+~~~ corefile
+ . {
```
miekg, line 114: We should get out of the habit of doing every example with `.`, please change this into into `example.org`, most examples also include a valid Corefile so add a forward in here or just *whoami*

miekg (changes requested):
We're also doing this because Python is stupid: https://stackoverflow.com/questions/54778160/python-requests-library-not-resolving-non-authoritative-dns-lookups

That's fixed with 1.5.1 so I don't see why this is needed, is it creates (slightly) more complexity on our side.

So NACK

chrisohaver:
This is done for a general use case.  Not just for the python issue.
One specific use case is that some people disable negative cache in micro-service type environments.
But I don't care about this strongly.  So closing...

PR #2884, opened 2019-06-12
Raise error if regexp and template are not specified together

ihac:
Signed-off-by: Xiao An <hac@zju.edu.cn>

Bond `regexp` and `template` together for the directive `directory`, so that users have to specify either both or none of them, as described in the README. 

None.

None.

No.

### DIFF
--- a/plugin/auto/setup_test.go
+++ b/plugin/auto/setup_test.go
@@ -75,6 +75,13 @@ func TestAutoParse(t *testing.T) {
 			}`,
 			true, "/tmp", "bliep", `(.*)`, 10 * time.Second, nil,
 		},
+		// no template specified.
+		{
+			`auto {
+				directory /tmp (.*)
+			}`,
+			true, "/tmp", "", `(.*)`, 60 * time.Second, nil,
+		},
 		// no directory specified.
 		{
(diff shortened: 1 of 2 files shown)


plugin/auto/setup_test.go
```diff
@@ -75,6 +75,13 @@ func TestAutoParse(t *testing.T) {
 			}`,
 			true, "/tmp", "bliep", `(.*)`, 10 * time.Second, nil,
 		},
+		// no template specified.
+		{
+			`auto {
+				directory /tmp (.*)
+			}`,
+			true, "/tmp", "bliep", `(.*)`, 60 * time.Second, nil,
```
chrisohaver: why "bliep"?

ihac: Well, actually it doesn't matter because we're expecting an error here (`shouldErr = true`). I just reused another test case. Anyway, I'll remove `bliep` to make it clear.

chrisohaver (approved):

plugin/auto/setup_test.go
chrisohaver: Thanks! lgtm

PR #2885, opened 2019-06-12
plugin/cache: remove item.Autoritative

### LINKED ISSUE #2887: cache plugin: authoritative/non-authoritative flag confuses old(er) clients
As described in https://github.com/coredns/coredns/pull/2588#issuecomment-501131738 and https://github.com/coredns/coredns/pull/2588#issuecomment-501353963 the cache plugin is currently confusing old(er) DNS clients or client implementations (Pyhton, Ruby, ...)

This happens because only the first answer from the DNS cache is provided as authoritative, all other ones (until the TTL is running out) are responded with non-authoritative.

@miekg has provided a PR which pins the answers given from the cache to authoritative.

PR: https://github.com/coredns/coredns/pull/2885
Kubernetes Issue trackback: https://github.com/kubernetes/kubeadm/issues/1512

@miekg  We have testet the PR in an own build on-premises and the old(er) clients can now work with the cached DNS records without any problem.

miekg:
Confuses clients if not set; remove it.

Fixes #2887

### DIFF
--- a/plugin/cache/cache_test.go
+++ b/plugin/cache/cache_test.go
@@ -17,7 +17,6 @@ type cacheTestCase struct {
 	test.Case
 	in                 test.Case
 	AuthenticatedData  bool
-	Authoritative      bool
 	RecursionAvailable bool
 	Truncated          bool
 	shouldCache        bool
@@ -25,7 +24,7 @@ type cacheTestCase struct {
 
 var cacheTestCases = []cacheTestCase{
 	{
-		RecursionAvailable: true, AuthenticatedData: true, Authoritative: true,
+		RecursionAvailable: true, AuthenticatedData: true,
 		Case: test.Case{
 			Qname: "miek.nl.", Qtype: dns.TypeMX,
 			Answer: []dns.RR{
@@ -43,7 +42,7 @@ var cacheTestCases = []cacheTestCase{
 		shouldCache: true,
 	},
 	{
-		RecursionAvailable: true, AuthenticatedData: true, Authoritative: true,
+		RecursionAvailable: true, AuthenticatedData: true,
 		Case: test.Case{
 			Qname: "mIEK.nL.", Qtype: dns.TypeMX,
 			Answer: []dns.RR{
@@ -70,7 +69,7 @@ var cacheTestCases = []cacheTestCase{
 		shouldCache: false,
 	},
 	{
-		RecursionAvailable: true, Authoritative: true,
+		RecursionAvailable: true,
 		Case: test.Case{
 			Rcode: dns.RcodeNameError,
 			Qname: "example.org.", Qtype: dns.TypeA,
@@ -88,7 +87,7 @@ var cacheTestCases = []cacheTestCase{
 		shouldCache: true,
 	},
 	{
-		RecursionAvailable: true, Authoritative: false,
+		RecursionAvailable: true,
 		Case: test.Case{
 			Rcode: dns.RcodeServerFailure,
 			Qname: "example.org.", Qtype: dns.TypeA,
@@ -102,7 +101,7 @@ var cacheTestCases = []cacheTestCase{
 		shouldCache: true,
 	},
 	{
-		RecursionAvailable: true, Authoritative: false,
+		RecursionAvailable: true,
 		Case: test.Case{
 			Rcode: dns.RcodeNotImplemented,
 			Qname: "example.org.", Qtype: dns.TypeA,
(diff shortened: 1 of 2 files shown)


chrisohaver:
When we cache a non-authoritative answer, now cache will send to the client as if it is was authoritative.  Seems like it would less wrong to retain the original value and pass that to the client.

miekg:
>When we cache a non-authoritative answer, now cache will send to the client as if it is was authoritative.  Seems like it would less wrong to retain the original value and pass that to the client.

That would make this failure case even harder to debug - because it will only 
happen for some names.

That a DNS client looks at this bit and *discards* the reply because of it is 
completely wrong.

m4r10k:
@miekg the PR patched version (1.5.0-pr2885) now runs on-premises without any problems now. LGTM.

Would you like to have a Readme update too, or should we got with it as it is?

Thanks!

miekg:
>@miekg the PR patched version (1.5.0-pr2885) now runs on-premises without any problems now. LGTM.

thanks for testing!

>Would you like to have a Readme update too, or should we got with it as it is?

What needs to be changed here?

/Miek

--
Miek Gieben

m4r10k:
Your are welcome!

As for the Reame: It was just a question if you would like to mention this change anyway in the readme or if it is enough to have it in the changelog. I only raised the question to fullfil the usual process. :see_no_evil:

miekg:
There is nothing about this in the README, so I'll won't add anything. For the release notes you're correct this will be on there.

chrisohaver:
Won't clients continue to be confused by non-authoritative answers when the cache plugin isn't being used?  If so, would this hack be better moved to a more global context than just in the cache plugin?

miekg:
>Won't clients continue to be confused by non-authoritative answers when the cache plugin isn't being used?  If so, would this hack be better moved to a more global context than just in the cache plugin?

possibly, but then it's out of our hands


/Miek

--
Miek Gieben

m4r10k:
Hmm, I am not that deep into DNS here and maybe I got this wrong (please forgive me) but as I understand, as long as the cache plugin is **not** enabled, every domain specified in the Corefile which coreDNS is responsible for **is** authoritative and if it is forwarded, it is not maybe.

The case here is, that the coreDNS, with enabled cache plugin, only responses the first time with an authoritative answer and later, as the data comes from the cache, it answers with non-authoritative, even if it is a domain it is always authoritative for (Corefile).

This behavior reflects, what the tcpdumps are showing, what the same client with the same software and the same DNS query is (wrongly) doing - only because the cache plugin is enabled.

But hey, I could be wrong here for sure 🙈 😄

miekg:
>Hmm, I am not that deep into DNS here and maybe I got this wrong (please forgive me) but as I understand, as long as the cache plugin is **not** enabled, ever domain specified in the Corefile which coreDNS is responsible for **is** authoritative. The case here is, that the coreDNS, with enabled cache plugin, only responses the first time with an authoritative answer and later, as the data comes from the cache, it answers with non-authoritative.
>
>This behavior reflects, what the tcpdumps are showing, what the same client with the same software and the same DNS query is (wrongly) doing - only because the cache plugin is enabled.
>
>But hey, I could be wrong here for sure 🙈 😄

The use-case is when you forward requests to another server, which may or may 
not be a cache. It then depends on the upstream if this bit is set or not. In 
general you don't want to fiddle with the bits going through unless mandated by 
the protocol, but in the DNS that ship has sailed a long time ago.

Is there an upstream bug the for Python resolver btw?

m4r10k:
> Is there an upstream bug the for Python resolver btw?

I do not know. What we encounter is, that with some quite old Linuxes we got resolving errors from different programs, like Puppet agent or rsync for example. Both can maybe tracked down to an old glibc (or something), as somehow mention on the internet. I guess that the program languages, in our case Ruby will somewhere be linked against a C library or uses some DNS code(?) of an C library which may cause this symptoms. There are legacy systems out there (also with Kubernetes and containers as such) which are not that actual as they should be of course, but nevertheless nobody is perfect. :smile: 

In our case, the cache **is** useful, as we use the ETCD backend massively and without caching, if the storages where the ETCD backends are running (virtual machines on SAN storage) have a hiccup, the coreDNS cannot answer the queries just in time. Especially in a container driven environment, like we have, this [can hurt cluster raft protocols](https://www.n0r1sk.com/post/2019-06-03-reviving-anear-death-docker-swarm-cluster/) - only some seconds (2-3) of a storage switchover are enough to face strange situations. Therefore, for reliability, we have to enable caching to mitigate these possible timeouts. But without this patch, we break older legacy services. Meh..

This use case of DNS - containers and fast dynamic service discovery - wasn't on the radar 20-30 years ago, but now it is and as @miekg said, who cares about this flag in case of cache? There seems to be, as I understand, no hard protocol standards for this in DNS.

So this PR does what it should with minimal impact on the rest. :+1: 

PS: containers do not have a DNS cache (nscd, systemd-resolved, ...)

PR #2886, opened 2019-06-13
core: Close (TCP) connections

miekg:
This was triggered by #2866 and #2872, but it triggers a race. It seems
simpler to just close this in the main server so none of the plugins
need to worry about it.

Unrelated: it may be worth checking/finding out how useful Hijack
actually is in miekg/dns

/cc @jinmeiib

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -188,6 +188,8 @@ func (s *Server) Address() string { return s.Addr }
 // defined in the request so that the correct zone
 // (configuration and plugin stack) will handle the request.
 func (s *Server) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) {
+	defer w.Close()
+
 	// The default dns.Mux checks the question section size, but we have our
 	// own mux here. Check if we have a question section. If not drop them here.
 	if r == nil || len(r.Question) == 0 {
--- a/plugin/file/xfr.go
+++ b/plugin/file/xfr.go
@@ -50,8 +50,6 @@ func (x Xfr) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (in
 		ch <- &dns.Envelope{RR: records[j:]}
 	}
 
-	w.Hijack()
-	// w.Close() // Client closes connection
 	return dns.RcodeSuccess, nil
 }
 


miekg:
This might be break cache's prefetch because that fires off another Go routine.

Also the travis failure need to be checked out.

jinmeiib:
> This might be break cache's prefetch because that fires off another Go routine.

And same for the file plugin in the case of xfr.  In fact, if I try to axfr a large zone (containing 64K AAAA RRs), it's interrupted in the middle of the transfer:

```
;; communications error to 127.0.0.1#53: end of file
```

It succeeds if I revert ef8d544.

If we want to close it in the main goroutine, I suspect the main goroutine will have to wait until the "child" goroutine completes its job (but that approach has its own drawbacks, of course).

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] core: Close (..." ]
>> This might be break cache's prefetch because that fires off another Go routine.
>
>And same for the file plugin in the case of xfr.  In fact, if I try to axfr a large zone (containing 64K AAAA RRs), it's interrupted in the middle of the transfer:

should probably convert that into a unit test?

Interestingly enough from (casual) looking at the code; it returns after the xfr 
is done. But apparently that isn't true.

>```
>;; communications error to 127.0.0.1#53: end of file
>```
>
>It succeeds if I revert ef8d544.
>
>If we want to close it in the main goroutine, I suspect the main goroutine will have to wait until the "child" goroutine completes its job (but that approach has its own drawbacks, of course).


Well, this is exactly what Hijack() should prevent, so at least that part make 
sense.

/Miek

--
Miek Gieben

jinmeiib:
On thinking about it more, I guess it actually not that a bad idea to have the "parent" goroutine wait until the "child" completes sending out the response.  the original implementation didn't accept more queries after handling an xfr query anyway and newer queries on newer connections will still be accepted by the listener, so the blocking behavior would be mostly the same as the original one.

I've implemented this approach on my fork for reference (not sent a PR yet).  See https://github.com/coredns/coredns/commit/7ced520792e6e4e0563ebceebac0385827d75be5

You may want to make it more generic, though, like the one in this PR.  In that case we'll probably need more organizational change (perhaps something more sophisticated/covoluted version of 'Hijack').

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] core: Close (..." ]
>On thinking about it more, I guess it actually not that a bad idea to have the "parent" goroutine wait until the "child" completes sending out the response.  the original implementation didn't accept more queries after handling an xfr query anyway and newer queries on newer connections will still be accepted by the listener, so the blocking behavior would be mostly the same as the original one.
>
>I've implemented this approach on my fork for reference (not sent a PR yet).  See https://github.com/coredns/coredns/commit/7ced520792e6e4e0563ebceebac0385827d75be5
>
>You may want to make it more generic, though, like the one in this PR.  In that case we'll probably need more organizational change (perhaps something more sophisticated/covoluted version of 'Hijack').

See lower level discussion in: https://github.com/miekg/dns/issues/983

/Miek

--
Miek Gieben

jinmeiib:
> See lower level discussion in: miekg/dns#983

I think my second fix (https://github.com/coredns/coredns/commit/7ced520792e6e4e0563ebceebac0385827d75be5) essentially implements the idea of "`ResponseWriter` must not outlive the `ServeDNS` handler and that methods can’t be called concurrently."

And, as for this:

> Interestingly enough from (casual) looking at the code; it returns after the xfr is done. But apparently that isn't true.

In the original code `Xfr.ServeDNS()` returns immediately after passing RRs to the spawned goroutine.  It doesn't necessarily guarantee that the other goroutine completes its work on the connection.  Using a larger size of zone just makes it more apparent.

If I find time I'll try to make it in the form of unit test and contribute it.

miekg:
I must check your code, but I think it's adds complexity to add this kind
of synchronization.

Hijack() does the same without caring about lifetimes (and is already part
of the DNS lib )

jinmeiib:
> Hijack() does the same without caring about lifetimes (and is already part of the DNS lib )

Sure, but then we have the issue of how to close the connection:

- If the hijacker doesn't close it, it will stay open until it's garbage collected
- If the hijacker closes it, it needs to be careful about spawning goroutines

If you prefer less complex fix, this set of change is probably (one of) the simplest, at least in terms of the number of lines to be modified:
- have the xfr's goroutine close the connection on exit (as my first patch did)
- the handler (parent) goroutine `Hijack()`s the writer (as the current implementation does)
- In `miekg/dns.Server.serveTCPConn()`, check `hijacked` before `closed` (to prevent the "race condition")

Although the race avoidance of this approach may be too sutble and fragile.

Anyway I didn't intend to insist on my second patch (https://github.com/coredns/coredns/commit/7ced520792e6e4e0563ebceebac0385827d75be5).  It's just for demonstrating one way to fix the issues identified so far.

miekg:
thanks for you comments. Tom also replied on the miekg/dns issue I opened.

We can (and should) close in the plugin (or add a WriteMsgAndClose function in pkg/dnsutil).

Closing this PR - because it's not the way forward.

Issue #2880, opened 2019-06-11
plugin/file: misbehaviour in zone with @


rgl:
When using CoreDNS 1.5.0 with `Corefile`:

```
. {
    file example.org.zone {
        transfer to *
    }
    log
    errors
}
```

And `example.org.zone` file:

```
$TTL 30m
$ORIGIN example.org.

@       in soa      ns  hostmaster (
    2019061101  ; serial number (this number should be increased each time this zone file is changed)
    4h          ; refresh (the polling interval that slave DNS server will query the master for zone changes)
                ; NB the slave will use this value instead of $TTL when deciding if the zone it outdated
    1h          ; update retry (the slave will retry a zone transfer after a transfer failure)
    7d          ; expire (the slave will ignore this zone if the transfer keeps failing for this long)
    4h          ; minimum (the slave stores negative results for this long)
)
        in ns      ns
ns      in a       10.0.0.2
ns      in a       10.0.0.3
www     in a       10.0.1.2
```

The resolution of `www` seems to fail (there is no `A RR` in the answers):

```
C:\Users\Rui Lopes>dig a www.example.org @127.0.0.1

; <<>> DiG 9.14.2 <<>> a www.example.org @127.0.0.1
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 42833
;; flags: qr rd; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 3
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
; COOKIE: 338ef5b5d416341d (echoed)
;; QUESTION SECTION:
;www.example.org.               IN      A

;; AUTHORITY SECTION:
example.org.            1800    IN      NS      ns.example.org.

;; ADDITIONAL SECTION:
ns.example.org.         1800    IN      A       10.0.0.2
ns.example.org.         1800    IN      A       10.0.0.3

;; Query time: 6 msec
;; SERVER: 127.0.0.1#53(127.0.0.1)
;; WHEN: Tue Jun 11 07:51:04 Hora de VerÒo de GMT 2019
;; MSG SIZE  rcvd: 155
```

But `axfr` seems to return the correct information:

```
C:\Users\Rui Lopes>dig axfr example.org @127.0.0.1

; <<>> DiG 9.14.2 <<>> axfr example.org @127.0.0.1
;; global options: +cmd
[...]

PR #2893, opened 2019-06-17
plugin/file: document using an explicit origin

miekg:
See #2880

/cc @rgl


stp-ip (commented):
/LGTM

chrisohaver:
"It's better to be explicit (which can be done in two ways) here and specify the correct origin:"

I think reads easier as ...

"It's better to be explicit here and specify the correct origin. This can be done in two ways:"

PR #2905, opened 2019-06-20
added a test for transferring a large zone in file plugin

jinmeiib:

See the discussion on PR #2886.  This test would have identified the issue of commit ef8d544.

No github issue, but the problem was discussed in PR #2886.

Shouldn't require a doc, as it only adds a new test.

Likewise, shouldn't introduce any incompatibility.

One additional note: this test also has code that exercises the issue raised in PR #2866.  It's currently disabled as it would fail, but if that additional check is deemed to be a distraction please feel free to remove it - it has nothing to do with transferring a large zone.

### DIFF
--- a/test/file_xfr_test.go
+++ b/test/file_xfr_test.go
@@ -13,8 +13,6 @@ import (
 )
 
 func TestLargeAXFR(t *testing.T) {
-	t.Parallel()
-
 	// Build a large zone in text format.  It contains 64K AAAA RRs.
 	var sb strings.Builder
 	const numAAAAs = 65536
--- a/test/file_xfr_test.go
+++ b/test/file_xfr_test.go
@@ -56,7 +56,7 @@ func TestLargeAXFR(t *testing.T) {
 		t.Fatalf("Unable to send TCP query: %s", err)
 	}
 
-	nrr := 0		// total number of transferred RRs
+	nrr := 0                                             // total number of transferred RRs
 	co.SetReadDeadline(time.Now().Add(60 * time.Second)) // use a longer timeout as it involves transferring a non-trivial amount of data.
 	for {
 		resp, err := co.ReadMsg()
@@ -78,7 +78,7 @@ func TestLargeAXFR(t *testing.T) {
 		}
 	}
 	// On successful completion, 2 SOA, 1 NS, and all AAAAs should have been transferred.
-	if nrr != numAAAAs + 3 {
+	if nrr != numAAAAs+3 {
 		t.Fatalf("Got an unexpected number of RRs: %d", nrr)
 	}
 
--- a/test/file_xfr_test.go
+++ b/test/file_xfr_test.go
@@ -1,0 +1,99 @@
+package test
+
+import (
+	"fmt"
+	"io"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/coredns/coredns/plugin/test"
+
+	"github.com/miekg/dns"
+)
+
+func TestLargeAXFR(t *testing.T) {
+	t.Parallel()
+
+	// Build a large zone in text format.  It contains 64K AAAA RRs.
+	var sb strings.Builder
+	sb.WriteString("example.com. IN SOA . . 1 60 60 60 60\n")
+	sb.WriteString("example.com. IN NS ns.example.\n")
+	for i := 0; i < 65536; i++ {
+		sb.WriteString(fmt.Sprintf("%d.example.com. IN AAAA 2001:db8::1\n", i))
+	}
+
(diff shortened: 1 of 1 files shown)


test/file_xfr_test.go
```diff
@@ -0,0 +1,99 @@
...
+			t.Fatalf("Expected SOA, but got type %d", resp.Answer[0].Header().Rrtype)
+		}
+		nrr += len(resp.Answer)
+		// If we see another SOA at the end of the message, we are done.
+		// Note that this check is not enough to detect all invalid responses, but checking those is not the purpose of this test.
+		if nrr > 1 && resp.Answer[len(resp.Answer)-1].Header().Rrtype == dns.TypeSOA {
+			break
+		}
+	}
+	if nrr != 65539 {
```
bradbeam: mind adding a comment here around the number 65537 records + soa + ns.

i wonder if it'd be better to do a `[]string` to house the generated records so we could do len(generated) + soa + ns instead of hardcoding a number.

bradbeam (commented):
overall lgtm. minor comment

test/file_xfr_test.go
jinmeiib: Thank you for the review, and that's a good point.  I've updated it to avoid hardcoding 65539.  I could use a `[]string`, but since only the number of RRs matters and we need `+1` (for the additional SOA) anyway, I just introduced a constant and referred to it elsewhere.

miekg:
opened #2929 to discuss further; test is good to have

jinmeiib:
btw, there was a report on integration test failure, but I don't think it's due to my change.  The failed test is not the one I added (that one passed) and I didn't change non-test code.

I'm not in a hurry, but just to avoid possible deadlock, please let me know if I'm supposed to do something more to close this PR (merge or not).

miekg:
Ah. I thought it was failing because of you testing something that should
fail.

I'll recheck Travis and then merge

miekg:
restarted multiple times, but keep hitting negative sync counter a lot

Have hunch that a test is too fast and server start and stop clash, maybe Go 1.12.6 made that worse

jinmeiib:
> restarted multiple times, but keep hitting negative sync counter a lot

hmm.  and it didn't happen without running my test?

miekg:
I don't know, haven't tried that

miekg:
It's go 1.12.6; now it fails locally as well

test/file_xfr_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"testing"
+	"time"
+
+	"github.com/coredns/coredns/plugin/test"
+
+	"github.com/miekg/dns"
+)
+
+func TestLargeAXFR(t *testing.T) {
+	t.Parallel()
```
miekg:
```suggestion
```

test/file_xfr_test.go
```diff
@@ -0,0 +1,101 @@
...
+       }
+}
+`
+
+	// Start server, and send an AXFR query to the TCP port.  We set the deadline to prevent the test from hanging.
+	i, _, tcp, err := CoreDNSServerAndPorts(corefile)
+	if err != nil {
+		t.Fatalf("Could not get CoreDNS serving instance: %s", err)
+	}
+	defer i.Stop()
```
miekg, line 44:
```suggestion
```

test/file_xfr_test.go
```diff
@@ -0,0 +1,101 @@
...
+		if nrr == 0 && resp.Answer[0].Header().Rrtype != dns.TypeSOA {
+			t.Fatalf("Expected SOA, but got type %d", resp.Answer[0].Header().Rrtype)
+		}
+		nrr += len(resp.Answer)
+		// If we see another SOA at the end of the message, we are done.
+		// Note that this check is not enough to detect all invalid responses, but checking those is not the purpose of this test.
+		if nrr > 1 && resp.Answer[len(resp.Answer)-1].Header().Rrtype == dns.TypeSOA {
+			break
+		}
+	}
```
miekg, line 79:
```suggestion
	}
	i.Stop()
```

miekg (commented):
suggestions

code is also not gofmt-ed

test/file_xfr_test.go
jinmeiib: removed as suggested.

jinmeiib: also applied `gofmt`.

test/file_xfr_test.go
jinmeiib: hmm...this may not be a big deal anyway as it's just a test, but is there anything wrong about guaranteeing to call `Stop` using `defer`?

PR #2918, opened 2019-06-24
Run gofmt -s and golint on the codebase

miekg:
Run golint and fix everythign it flagged (except the context arg
ordering), mostly edits in the rewrite plugin.


yongtang (approved):
LGTM

Issue #2911, opened 2019-06-20
plugin/kubernetes: deprecate resyncperiod


miekg:
from 1.5.0 changelog:
~~~
The resyncperiod option in kubernetes is deprecated and will be made a no-op in 1.6.0 and removed in 1.7.0.
~~~
Next release is likely 1.6.0 so we can start doing this.

sharmaansh21:
Get rid of `resyncperiod` from everywhere or they have the equivalent option?

chrisohaver:
The aim is to remove the option from the _kubernetes_ plugin with no equivalent replacement.
The option is already deprecated (as of 1.5.0).
In 1.6.0, it will be ignored if used in the Corefile.
In a following release it will cause an error if used in the Corefile.

IIUC, [this discussion](https://groups.google.com/forum/#!topic/kubernetes-sig-api-machinery/PbSCXdLDno0) reveals that `resyncperiod` is not useful in the context of how the _kubernetes_ plugin uses the API (a read-only watcher).  Furthermore, the function of the option is not what one might expect in this context, i.e. resync != relist.

sharmaansh21:
I can work on the deprecation part on the coredns side :bowtie:

sharmaansh21:
If I am going in the right direction, we need to wait for https://github.com/kubernetes/client-go/blob/master/tools/cache/controller.go#L52 to be removed first?
Used here https://github.com/coredns/coredns/blob/master/plugin/kubernetes/object/informer.go#L20

chrisohaver:
@sharmaanshul21,  Just pass `0`.  The option isn't going to be removed from the kubernetes API client. IIRC, `0` disables the resync.

```
FullResyncPeriod: 0,
```

miekg:
this has been merged and will go out in 1.6.0.

PR #2923, opened 2019-06-25
plugin/kubernetes: deprecate resyncperiod

sharmaansh21:
- Remove `resyncperiod` from Kubernetes plugin

#2911 

@chrisohaver something like this ?


plugin/kubernetes/setup.go
```diff
@@ -211,17 +210,6 @@ func ParseStanza(c *caddy.Controller) (*Kubernetes, error) {
 				continue
 			}
 			return nil, c.ArgErr()
-		case "resyncperiod":
```
chrisohaver: Instead of just removing, need to replace this case with a no-op... e.g.

```go
case "resyncperiod":
    continue
```

Following out deprecation policy, this will later be removed in a following version.

chrisohaver:
@sharmaanshul21, Yes, mostly lgtm, thanks!  See the one comment I made.
You may also need to update some unit tests (if there are any that check the setting of the option, the wont compile now).

sharmaansh21:
@chrisohaver Updated :+1:

chrisohaver:
Looks good, thanks. Holding off on merging until we know for certain that the next CoreDNS release number will be 1.6.0 (and not 1.5.2).

chrisohaver (approved):

miekg (approved):

Issue #2935, opened 2019-06-27
TLS config issues


bcebere:
Hi

The TLS plugin overwrites the tls.Config from Caddy, and zeroes the minimum version, maximum version and default ciphersuites.

This results in some issues when using CoreDNS with TLS support:
 - SSLv3, TLSv1.0 and TLSv1.1 are accepted. 
 - The are some weak ciphersuites accepted.
 - No server preferred ciphersuite order is provided.
 - No TLS 1.3 support

Given that DNS-over-TLS and DNS-over-HTTPS are relatively new, I think it shouldn't be backwards compatible with those weak configs.

Some tools to reproduce the issues:
https://github.com/drwetter/testssl.sh/
or 
https://github.com/ssllabs/ssllabs-scan

Run them against a TLS-enabled CoreDNS to reproduce the issues.

I generated a patch here https://gist.github.com/bcebere/cc12acb8ecbb7527061ecade69b1619b

I tried several times to push a branch and create a pull request in CoreDNS, but I guess you have some restrictions and it's not possible. It didn't happen with other repos.

So, I'll leave the patch here,maybe it's useful.

sharmaansh21:
@bcebere You need to fork the repo and push it to your fork and create a branch. I am happy to do that for you if you don't have time or do not wish to do it.
About the solution, I don't have much expertise in this area so maybe @chrisohaver @miekg can help here.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] TLS config issues..." ]
>I generated a patch here https://gist.github.com/bcebere/cc12acb8ecbb7527061ecade69b1619b

Some comments on that patch:

`const pluginName = "tls"` is an indirection that doesn't make things more 
readable, just use "tls" directly as it now.

For the `hardenTLS` function is there anything from the std lib we can use? How 
do we know this is a good config?

Do we need to do something server side? https://github.com/coredns/coredns/blob/master/core/dnsserver/server_https.go

bcebere:
I created a pull request now, thank you for the clarification @sharmaanshul21 
https://github.com/coredns/coredns/pull/2938


@miekg :
I removed the indirection, thank you for the insight.

I don't think there is some standard function, but if the tls plugin doesn't overwrite the default Caddy config, at least TLS 1.1 or earlier wouldn't be supported.

[...]

PR #2938, opened 2019-06-28
TLS hardening

bcebere:
The TLS plugin overwrites the tls.Config from Caddy, and zeroes the minimum version, maximum version and default ciphersuites.

This results in some issues when using CoreDNS with TLS support:

SSLv3, TLSv1.0 and TLSv1.1 are accepted.
The are some weak ciphersuites accepted.
No server preferred ciphersuite order is provided.
No TLS 1.3 support

https://github.com/coredns/coredns/issues/2935

Drops support for SSLv3, TLS v1.0 and TLS v1.1

### DIFF
--- a/plugin/tls/tls.go
+++ b/plugin/tls/tls.go
@@ -27,7 +27,6 @@ func setup(c *caddy.Controller) error {
 }
 
 func setTLSDefaults(tls *ctls.Config) {
-	os.Setenv("GODEBUG", "tls13=1")
 	tls.MinVersion = ctls.VersionTLS12
 	tls.MaxVersion = ctls.VersionTLS13
 	tls.CipherSuites = []uint16{
--- a/plugin/tls/tls.go
+++ b/plugin/tls/tls.go
@@ -26,7 +26,7 @@ func setup(c *caddy.Controller) error {
 	return nil
 }
 
-func hardenTLS(tls *ctls.Config) {
+func setTLSDefaults(tls *ctls.Config) {
 	os.Setenv("GODEBUG", "tls13=1")
 	tls.MinVersion = ctls.VersionTLS12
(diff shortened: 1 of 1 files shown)


plugin/tls/tls.go
```diff
@@ -25,6 +26,24 @@ func setup(c *caddy.Controller) error {
 	return nil
 }
 
+func hardenTLS(tls *ctls.Config) {
+	os.Setenv("GODEBUG", "tls13=1")
```
miekg: This looks odd and brittle, it also overwrites GODEBUG which is something we don't want. Please remove this line.

plugin/tls/tls.go
```diff
@@ -25,6 +26,24 @@ func setup(c *caddy.Controller) error {
 	return nil
 }
 
+func hardenTLS(tls *ctls.Config) {
```
miekg: It's not really harden, it's settting some defaults that we think are good.

setTLSParameters, or just setTLSDefaults?

plugin/tls/tls.go
bcebere: Unfortunately, this seems the only way to activate TLS 1.3 in Go right now.

I don't know how to do it cleaner.

bcebere: I could change it to
`os.Setenv("GODEBUG", os.Getenv("GODEBUG")+",tls13=1")`

At least I won't overwrite the variable.

I took the info from here https://github.com/golang/go/issues/30055

miekg (approved):

plugin/tls/tls.go
bcebere: Should we wait for the default TLS 1.3 support in Go fow now? 
I am not sure why they chose the opt-in via GODEBUG

miekg:
let's try again

/merge

PR #2942, opened 2019-06-29
Move to dns.Truncate

### LINKED ISSUE #2728: Use more efficient dns.Truncate
https://github.com/miekg/dns/commit/d051b464e9f024b65ce9674dece9d3e905b31828 adds dns.Truncate which linearly goes of the message and truncates. This is more efficient than the current binary search implementation we have.

coredns: binary search log2(n) Len's on the _entire message_
dns: Len(n) on the entire message.

The gist being: we are `Len`-ing mutiple times over the same RR currently, while the dns impl. just `Len`s once over each RRs.

miekg:
Ditch our truncation code and use the upstream one in miekg/dns.
This saves code on our end, end upstream is also more efficient as every
RR is Len-ed only once. With our bin-search this is not guaranteed.

Fixes #2728

### DIFF
--- a/request/request.go
+++ b/request/request.go
@@ -226,22 +226,17 @@ func (r *Request) SizeAndDo(m *dns.Msg) bool {
 
 // Scrub scrubs the reply message so that it will fit the client's buffer. It will first
 // check if the reply fits without compression and then *with* compression.
-// Scrub will then use binary search to find a save cut off point in the additional section.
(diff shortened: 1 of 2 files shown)


miekg:
I want to go on and just merge this, esp early-ish in the cycle, so it may see some testing.

miekg:
thanks!

PR #2943, opened 2019-06-29
file: close correctlty after AXFR

### LINKED ISSUE #2929: call Close() on the response writer
We're curently not calling Close which for TCP might be a problem. And it show up in AXFR. @jinmeiib is trying to fix things here.

Opening an issue to discuss option as we were having them on pull requests.

Options:

1. The server can call close; but this leads to data race, as these need Close() calls needs to be made by the goroutine writing to them
2. add a WriteMsgAndClose() call in coredns
3. add Close() to the plugins that need them

Note the cache plugin re-uses the response writer for prefetch, it should create a new responsewriter to do that and just let the old one be closed.

miekg:
Don't hijack, but wait for the writes to be done and then safely close
the connection.

Fixes: #2929

### DIFF
--- a/test/file_xfr_test.go
+++ b/test/file_xfr_test.go
@@ -2,7 +2,6 @@ package test
 
 import (
 	"fmt"
-	"net"
 	"strings"
 	"testing"
 	"time"
@@ -52,9 +51,17 @@ func TestLargeAXFR(t *testing.T) {
 	co.SetWriteDeadline(time.Now().Add(5 * time.Second))
 	err = co.WriteMsg(m)
 	if err != nil {
-		t.Fatalf("Unable to send TCP query: %s", err)
+		t.Fatalf("Unable to send AXFR/TCP query: %s", err)
 	}
 
+	// Then send another query on the same connection.  We use this to confirm that multiple outstanding queries won't cause a race.
+	m.SetQuestion("0.example.com.", dns.TypeAAAA)
+	err = co.WriteMsg(m)
+	if err != nil {
+		t.Fatalf("Unable to send AAAA/TCP query: %s", err)
+	}
+
+	// The AXFR query should be responded first.
 	nrr := 0 // total number of transferred RRs
 	for {
 		resp, err := co.ReadMsg()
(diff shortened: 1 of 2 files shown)


dilyevsky (approved):

test/file_xfr_test.go
```diff
@@ -82,18 +80,13 @@ func TestLargeAXFR(t *testing.T) {
 		t.Fatalf("Got an unexpected number of RRs: %d", nrr)
 	}
 
-	// At the time of the initial implementation of this test, the remaining check would fail due to the problem described in PR #2866, so we return here.
-	// Once the issue is resolved this workaround should be removed to enable the check.
-	return
-
-	// Once xfr is completed the server should close the connection, so a further read should result in an EOF error.
-	// This time we use a short timeout to make it faster in case the server doesn't behave as expected.
+	// Once xfr is completed the server should close the connection, so a further read should result in an error.
```
jinmeiib: The change to `file/xfr.go` doesn't seem to close the connection.  If that's the intent, this comment should be updated accordingly.  Also, in that case we don't have to bother to check the subsequent read times out.

miekg: It does, because it doesn't call Hijack(), so the server will close it.

Updated the comment; or are you proposing to remove the ending of this test?

jinmeiib: Ah, I see.  In that case checking timeout isn't really useful, since it wouldn't distinguish the updated behavior from the previous one (hijack and forget the connection).

So we could simply remove the additional checks.  But we can also confirm that the new implementation keeps the connection (rather than hijack) more explicitly.  The suggestion I'm pasting below implements that idea.  Feel free to use or ignore it.

```suggestion
diff --git a/test/file_xfr_test.go b/test/file_xfr_test.go
index bbdf2782..5e4df3a6 100644
--- a/test/file_xfr_test.go
+++ b/test/file_xfr_test.go
@@ -2,7 +2,6 @@ package test
 
 import (
 	"fmt"
-	"net"
 	"strings"
 	"testing"
 	"time"
@@ -52,9 +51,17 @@ func TestLargeAXFR(t *testing.T) {
 	co.SetWriteDeadline(time.Now().Add(5 * time.Second))
 	err = co.WriteMsg(m)
 	if err != nil {
-		t.Fatalf("Unable to send TCP query: %s", err)
+		t.Fatalf("Unable to send AXFR/TCP query: %s", err)
 	}
 
+	// Then send another query on the same connection.  We use this to confirm that multiple outstanding queries won't cause a race.
+	m.SetQuestion("0.example.com.", dns.TypeAAAA)
+	err = co.WriteMsg(m)
+	if err != nil {
+		t.Fatalf("Unable to send AAAA/TCP query: %s", err)
+	}
+
+	// The AXFR query should be responded first.
 	nrr := 0 // total number of transferred RRs
 	for {
 		resp, err := co.ReadMsg()
@@ -80,13 +87,15 @@ func TestLargeAXFR(t *testing.T) {
 		t.Fatalf("Got an unexpected number of RRs: %d", nrr)
 	}
 
-	// Once xfr is completed the server closes the connection, so a further read should result in an error.
-	co.SetReadDeadline(time.Now().Add(time.Second))
-	_, err = co.ReadMsg()
-	if err == nil {
-		t.Fatalf("Expected failure on further read, but it succeeded")
+	// The file plugin shouldn't hijack or (yet) close the connection, so the second query should also be responded.
+	resp, err := co.ReadMsg()
+	if err != nil {
+		t.Fatalf("Expected to receive reply, but didn't: %s", err)
+	}
+	if len(resp.Answer) < 1 {
+		t.Fatalf("Expected a non-empty answer, but it was empty")
 	}
-	if !err.(net.Error).Timeout() {
-		t.Fatalf("Expected timeout on further read, but got a different type of error: %v", err)
+	if resp.Answer[len(resp.Answer)-1].Header().Rrtype != dns.TypeAAAA {
+		t.Fatalf("Expected a AAAA answer, but it wasn't: type %d", resp.Answer[len(resp.Answer)-1].Header().Rrtype)
 	}
 }

```

miekg: applied. PTAL

jinmeiib: LGTM

PR #2944, opened 2019-06-29
plugin/file: load secondary zones lazily on startup

### LINKED ISSUE #1609: plugin/secondary: Denial-of-Service via endless Zone Transfer

miekg:
This fixes a long standing bug:
fixes: #1609

Load secondary zones in a go-routine; this required another mutex to
protect some fields; I think those were needded anyway because a
transfer can also happen when we're running; we just didn't have a test
for that situation.

The test had to be changed to wait for the transfer to happen at this is
async now.


jameshartig (approved):

yongtang (approved):
LGTM

Issue #2820, opened 2019-05-07
Support Azure DNS backend


krizhan:
From https://github.com/cncf/soc#support-azure-dns-backend, a suggested proposal was to implement a plugin to support azure dns backend (although I can't seem to find an open issue for it). However, it seems that this project was not selected. I am willing to implement this if help is still wanted for this plugin.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] Support Azure DNS..." ]
>From https://github.com/cncf/soc#support-azure-dns-backend, a suggested proposal was to implement a plugin to support azure dns backend (although I can't seem to find an open issue for it). However, it seems that this project was not selected. I am willing to implement this if help is still wanted for this plugin.

yep, this would still be welcome! I think/hope you can make this somewhat like 
the route53 plugin.

/Miek

--
Miek Gieben

yongtang:
Yes Azure dns backend could be a good addition 👍 . Would really help those with azure deployment.

palash25:
hi @yongtang , I had submitted a  GSoC proposal for this plugin and Nikhita Raghunath had reached out to me after the GSoC results saying that CNCF might be interested in sponsoring me to work on this project under the community bridge programme ( https://communitybridge.org/ ) but it wasn't decided when the work would start because CNCF is still busy deciding how many students to sponsor under the programme.

So I am a little confused that @krizhan wants to work on this now and someone is already working on the google cloud plugin too. Did Nikhita reach out to you regarding this @yongtang ?

yongtang:
@palash25 I was not aware of the community bridge program before, and have not been contacted. Maybe you can coordinate with @krizhan about this proposal?

palash25:
I have DM'd Nikhita on k8s slack, I was under the impression that she might have contacted you regarding this. 

Once I get a reply from her, I will update it here, thanks for the quick response though :smile:

krizhan:
Np. @palash25 can tackle this issue since its been hard to find free time w work anyways.

palash25:
thanks @krizhan

idvoretskyi:
@palash25 CNCF is happy to allocate one slot for CoreDNS at the Community Bridge. We'll follow up via email.

yongtang:
This issue could be closed as PR #2945 has been merged.

PR #2945, opened 2019-06-30
Add plugin for Azure DNS

darshanime:
Signed-off-by: darshanime <deathbullet@gmail.com>

Add a plugin for Azure DNS called `azure`, like we have `route53` for AWS.

Added a README.

No.


miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add plugin fo..." ]
>+package azureDNS

this would be better named 'azure', route53 is also just that.

yongtang:
+1 for renaming to `azure`, also, can you fix the stickler-ci warnings?

darshanime:
I have changed the name, fixed stikcler-ci warnings
@miekg @yongtang, please review.

plugin/azure/README.md
```diff
@@ -0,0 +1,87 @@
...
+## Syntax
+
+~~~ txt
+azure [resource-group:dns-zone] {
+    [azure_auth_location <path to azure_auth_location>]
+    [azure_tenant_id <azure_tenant_id>]
+    [azure_client_id <azure_client_id>]
+    [azure_client_secret <azure_client_secret>]
+    [azure_subscription_id <azure_subscription_id>]
+    upstream
```
miekg: `upstream` is a keyword is not needed anymore. Just make an upstream.New() in the code; that get's used is there is a forward (or something else to resolve) in this coredns

darshanime: gotcha, removed.

plugin/azure/setup.go
```diff
@@ -0,0 +1,139 @@
...
+	envSettings, keys, fall, err := parseCorefile(c)
+	if err != nil {
+		return err
+	}
+	up := upstream.New()
+	ctx := context.Background()
+	dnsClient := dns.NewRecordSetsClient(envSettings.Values[auth.SubscriptionID])
+	dnsClient.Authorizer, err = envSettings.GetAuthorizer()
+	if err != nil {
+		return c.Errf("\nfailed to create azure plugin:\n\n %v\n", err)
```
yongtang: Is it necessary to prefix and suffix with `\n`? I think those `\n` could be removed.

plugin/azure/azure.go
```diff
@@ -0,0 +1,226 @@
...
+
+// New validates the input dns zones and initializes the Azure struct
+func New(ctx context.Context, dnsClient azure.RecordSetsClient, keys map[string][]string, up *upstream.Upstream) (*Azure, error) {
+	zones := make(map[string][]*zone, len(keys))
+	zoneNames := make([]string, 0, len(keys))
+	for resourceGroup, inputZoneNames := range keys {
+		for _, zoneName := range inputZoneNames {
+			_, err := dnsClient.ListAllByDNSZone(context.Background(), resourceGroup, zoneName, nil, "")
+			if err != nil {
+				log.Fatalf("err not nil:%s", err.Error())
```
yongtang: The error should be more descriptive than `err not nil`.

plugin/azure/azure.go
```diff
@@ -0,0 +1,226 @@
...
+			}
+
+			if _, ok := zones[zoneName]; !ok {
+				zoneNames = append(zoneNames, fmt.Sprintf("%s.", zoneName))
```
yongtang: Not very sure about `"%s."`, can you explain a little about the intention here?

plugin/azure/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Description
+
+The azure plugin is useful for serving zones from resource record
+sets in Microsoft Azure DNS. The `azure` plugin can be used when coredns is deployed on Azure or elsewhere.
+
+## Syntax
+
+~~~ txt
+azure [resource-group:dns-zone] {
```
miekg: (i'm going over this PR slowly - hence a trickly of comments)

This syntax need to be styled different, see route53 readme for instance:
`<azure_tenant_id>` -> `TENTANT_ID` (for instance). 
Also not sure if prefixing everything with `azure` helps in readability.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add plugin fo..." ]
>yongtang commented on this pull request.
>
>
>
>> +	zoneNames []string
>+	client    azure.RecordSetsClient
>+	upstream  *upstream.Upstream
>+	zMu       sync.RWMutex
>+	zones     zones
>+}
>+
>+// New validates the input dns zones and initializes the Azure struct
>+func New(ctx context.Context, dnsClient azure.RecordSetsClient, keys map[string][]string, up *upstream.Upstream) (*Azure, error) {
>+	zones := make(map[string][]*zone, len(keys))
>+	zoneNames := make([]string, 0, len(keys))
>+	for resourceGroup, inputZoneNames := range keys {
>+		for _, zoneName := range inputZoneNames {
>+			_, err := dnsClient.ListAllByDNSZone(context.Background(), resourceGroup, zoneName, nil, "")
>+			if err != nil {
>+				log.Fatalf("err not nil:%s", err.Error())
>
>The error should be more descriptive than `err not nil`.

It should also be just returned and let the caller of New decide what to do. 
Note only loop and metrics call log.Fatalf currently, all other setups errors 
are handled by just returning an error from setup()

plugin/azure/azure.go
darshanime: cmiiw, CoreDNS adds a final trailing dot needed for the root of the DNS tree and makes the `qname` a fqdn. 
In the `zones` field of `Azure` struct, I added the same so that in `ServeDNS`, `plugin.Zones(h.zoneNames).Matches(qname)` results in a successful lookup.

yongtang: I know fqdn with dot, but I am asking about:
```
if _, ok := zones[zoneName]; !ok {
```

I would expect the zoneName is pre-normalized with (or without dot), then having one lookup only?

darshanime: hmm, if I understand correctly, you mean that `zoneName` would already be what it needs to be(be with or without dot `.`) in the `Corefile`? 
This won't work because azure doesn't allow creation of DNS zones with a trailing dot (throws `InvalidResourceNameTrailingCharacters`).
So, the user will always have a DNS zone like `foo.com`, we will fetch the records for `foo.com` using azure sdk but when serving via CoreDNS, it will query for a fqdn and lookup will fail if we don't make this transformation. 

let me know if I misunderstood your question though!

yongtang:
No I mean you should normalize zoneName before hand, so that if it does not contain dot it will be added. This should happen before you perform a lookup of

_, ok = zones[zoneName]

Sent from my iPhone

On Jul 4, 2019, at 2:32 PM, Darshan Chaudhary <notifications@github.com<mailto:notifications@github.com>> wrote:


@darshanime commented on this pull request.

In plugin/azure/azure.go<https://github.com/coredns/coredns/pull/2945#discussion_r300245739>:

> +     zones     zones
+}
+
+// New validates the input dns zones and initializes the Azure struct
+func New(ctx context.Context, dnsClient azure.RecordSetsClient, keys map[string][]string, up *upstream.Upstream) (*Azure, error) {
+       zones := make(map[string][]*zone, len(keys))
+       zoneNames := make([]string, 0, len(keys))
+       for resourceGroup, inputZoneNames := range keys {
+               for _, zoneName := range inputZoneNames {
+                       _, err := dnsClient.ListAllByDNSZone(context.Background(), resourceGroup, zoneName, nil, "")
+                       if err != nil {
+                               log.Fatalf("err not nil:%s", err.Error())
+                       }
+
+                       if _, ok := zones[zoneName]; !ok {
+                               zoneNames = append(zoneNames, fmt.Sprintf("%s.", zoneName))


hmm, if I understand correctly, you mean that zoneName would already be what it needs to be(be with or without dot .) in the Corefile?
This won't work because azure doesn't allow creation of DNS zones with a trailing dot (throws InvalidResourceNameTrailingCharacters).
So, the user will always have a DNS zone like foo.com<http://foo.com>, we will fetch the records for foo.com<http://foo.com> using azure sdk but when serving via CoreDNS, it will query for a fqdn and lookup will fail if we don't make this transformation.

let me know if I misunderstood your question though!

darshanime:
Ah, I gotcha. Sorry for the confusion. Made the change 🔝

darshanime:
Hello! Can we merge this?

go.mod
```diff
@@ -67,4 +72,4 @@ require (
 	sigs.k8s.io/yaml v1.1.0 // indirect
 )
 
-replace github.com/miekg/dns v1.1.3 => github.com/miekg/dns v1.1.15
```
yongtang: We want to leave 1.15 alone, as it is used to rewrite caddy's miekg/dns dependency to match the one CoreDNS depending on.

plugin/azure/azure.go
```diff
@@ -0,0 +1,223 @@
...
+	zones := make(map[string][]*zone, len(keys))
+	zoneNames := make([]string, 0, len(keys))
+	for resourceGroup, inputZoneNames := range keys {
+		for _, zoneName := range inputZoneNames {
+			_, err := dnsClient.ListAllByDNSZone(context.Background(), resourceGroup, zoneName, nil, "")
+			if err != nil {
+				return nil, err
+			}
+			// Normalizing zoneName to make it fqdn
+			zoneNameFQDN := fmt.Sprintf("%s.", zoneName)
```
yongtang: I think normalization might only be needed when zoneName does not ends with `.`. In other words, we could check if zoneName ends with `.`, and append only if it does not?

plugin/chaos/zowners.go
```diff
@@ -1,4 +1,4 @@
 package chaos
 
 // Owners are all GitHub handlers of all maintainers.
-var Owners = []string{"bradbeam", "chrisohaver", "dilyevsky", "ekleiner", "fastest963", "greenpau", "grobie", "inigohu", "isolus", "johnbelamaric", "miekg", "nchrisdk", "nitisht", "pmoroney", "rajansandeep", "rdrozhdzh", "rtreffer", "stp-ip", "superq", "varyoo", "yongtang"}
```
yongtang: There is a `no new line at end of file` after your update. Maybe you could exclude this file from the change?

plugin/azure/azure.go
```diff
@@ -45,8 +46,11 @@ func New(ctx context.Context, dnsClient azure.RecordSetsClient, keys map[string]
...
 				return nil, err
 			}
-			// Normalizing zoneName to make it fqdn
-			zoneNameFQDN := fmt.Sprintf("%s.", zoneName)
+			zoneNameFQDN := zoneName
+			// Normalizing zoneName to make it fqdn if required
+			if !strings.HasSuffix(zoneNameFQDN, ".") {
+				zoneNameFQDN = fmt.Sprintf("%s.", zoneNameFQDN)
+			}
 			if _, ok := zones[zoneNameFQDN]; !ok {
```
miekg, line 51: you can just use dns.Fqdn(zonename) in this look up and delete all the lines above

miekg:
> Hello! Can we merge this?

We're all volunteers here, so this won't get things merged faster (in fact it might slow things down).

I also still want to review this pr

plugin/azure/setup.go
```diff
@@ -0,0 +1,142 @@
...
+		return c.Errf("failed to run azure plugin: %v", err)
+	}
+	dnsserver.GetConfig(c).AddPlugin(func(next plugin.Handler) plugin.Handler {
+		h.Next = next
+		return h
+	})
+	return nil
+}
+
+func parseCorefile(c *caddy.Controller) (auth.EnvironmentSettings, map[string][]string, pkgFall.F, error) {
```
miekg: We usually just call this `parse`

plugin/azure/setup.go
```diff
@@ -0,0 +1,142 @@
...
+	keyPairs := map[string]struct{}{}
+	keys := map[string][]string{}
```
miekg: what's the difference between `keyPairs` and `keys`. Why do we need both?

plugin/azure/setup.go
```diff
@@ -0,0 +1,142 @@
...
+		for c.NextBlock() {
+			switch c.Val() {
+			case "auth_location":
+				var path string
+				if c.NextArg() {
+					path = c.Val()
+				} else {
+					return envSettings, keys, fall, c.ArgErr()
+				}
+				os.Setenv("AZURE_AUTH_LOCATION", path)
```
miekg: this env setting looks really weird and brittle. Can do we without it?

plugin/azure/azure_test.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+var demoAzure Azure = Azure{
+	Next:      returnTestHandler(),
+	Fall:      fall.Zero,
+	zoneNames: []string{"example.org", "www.example.org", "example.org.", "sample.example.org."},
+	zones:     returnTestZones(),
+}
+
+func returnTestZones() zones {
```
miekg: returnTestZones -> testZones

plugin/azure/azure_test.go
```diff
@@ -0,0 +1,149 @@
...
+		err := newZ.Insert(r)
+		if err != nil {
+			log.Fatalf("XXXXXXXXXXXXXX could not add file zone to zone: %s", err.Error())
+		}
+	}
+	zones["example.org."][0].z = newZ
+	return zones
+}
+
+func returnTestHandler() test.HandlerFunc {
```
miekg: -> testHandler

plugin/azure/azure_test.go
```diff
@@ -0,0 +1,149 @@
...
+	return func(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+		state := crequest.Request{W: w, Req: r}
+		qname := state.Name()
+		m := new(dns.Msg)
+		rcode := dns.RcodeServerFailure
+		if qname == "example.gov." {
```
miekg: example.org

plugin/azure/azure_test.go
```diff
@@ -0,0 +1,149 @@
...
+			m.SetReply(r)
+			rr, err := dns.NewRR("example.gov.  300 IN  A   2.4.6.8")
+			if err != nil {
+				log.Fatalf("Failed to create Resource Record: %v", err)
```
miekg: return the err

plugin/azure/azure_test.go
```diff
@@ -0,0 +1,149 @@
...
+	"context"
+	"reflect"
+	"testing"
+
+	"github.com/coredns/coredns/plugin/file"
+	"github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/coredns/coredns/plugin/pkg/fall"
+	"github.com/coredns/coredns/plugin/test"
+	crequest "github.com/coredns/coredns/request"
+	"github.com/miekg/dns"
```
miekg, line 14: newline before miekg/dns

plugin/azure/azure.go
```diff
@@ -0,0 +1,227 @@
...
+func (h *Azure) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	state := request.Request{W: w, Req: r}
+	qname := state.Name()
+
+	zName := plugin.Zones(h.zoneNames).Matches(qname)
+	if zName == "" {
+		return plugin.NextOrFailure(h.Name(), h.Next, ctx, w, r)
+	}
+
+	z, ok := h.zones[zName]
```
miekg: when does this happen if we're already are authoritative for the zone? Please add a comment

plugin/azure/azure.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	return &Azure{
+		client:    dnsClient,
+		zones:     zones,
+		zoneNames: zoneNames,
+		upstream:  up,
+	}, nil
+}
+
+// Run updates the zones once and starts a goroutine to do it every minute
```
miekg: dot at the end of this sentence

plugin/azure/setup.go
```diff
@@ -0,0 +1,142 @@
...
+var log = clog.NewWithPlugin("azure")
+
+func init() {
+	caddy.RegisterPlugin("azure", caddy.Plugin{
+		ServerType: "dns",
+		Action:     setup,
+	})
+}
+
+func setup(c *caddy.Controller) error {
```
miekg, line 28: I don't see you're restarting the go-routines to update the zones. Is this handled correctly after a reload?

miekg (changes requested):
Initial set of review nits/question/improvements.

plugin/azure/README.md
```diff
@@ -0,0 +1,83 @@
...
+    [auth_location <path to auth_location>]
```
miekg: we also usually don't have block quotes around these item, but say what the most minimal config looks like and then talk about extended config options (without the [ and ]) that are also possible.

Please take a look at route53/README.md and copy that.

plugin/azure/azure.go
```diff
@@ -0,0 +1,227 @@
...
+	return nil
+}
+
+func updateZoneFromResourceSet(recordSet azure.RecordSetListResultPage, zName string) *file.Zone {
+	newZ := file.NewZone(zName, "")
+	for _, result := range *(recordSet.Response().Value) {
+		if result.RecordSetProperties.ARecords != nil {
+			for _, a := range *(result.RecordSetProperties.ARecords) {
+				rfc1035 := fmt.Sprintf("%s %d IN %s %s", *(result.RecordSetProperties.Fqdn), *(result.RecordSetProperties.TTL), "A", *(a.Ipv4Address))
+				r, err := dns.NewRR(rfc1035)
```
miekg: This looks like a quite round about way of creating a new RR. You don't need to convert things to a string and parse the strings again. You can create a `dns.A` directly from these elements. Also removes the need to check for an error, becaue there is none

plugin/azure/azure.go
```diff
@@ -0,0 +1,227 @@
...
+	if len(errs) != 0 {
+		return fmt.Errorf("errors updating zones: %v", errs)
+	}
+	return nil
+}
+
+func updateZoneFromResourceSet(recordSet azure.RecordSetListResultPage, zName string) *file.Zone {
+	newZ := file.NewZone(zName, "")
+	for _, result := range *(recordSet.Response().Value) {
+		if result.RecordSetProperties.ARecords != nil {
```
miekg, line 115: a switch might be slightly more readable here. Also there are only 4 types a records in these zones? A/AAAA/NS and SOA, that seeems odd? If so this def. needs to be called out in the README.

plugin/azure/setup.go
```diff
@@ -0,0 +1,142 @@
...
+	"os"
+	"strings"
+
+	"github.com/coredns/coredns/core/dnsserver"
+	"github.com/coredns/coredns/plugin"
+	pkgFall "github.com/coredns/coredns/plugin/pkg/fall"
+	clog "github.com/coredns/coredns/plugin/pkg/log"
+	"github.com/coredns/coredns/plugin/pkg/upstream"
+
+	"github.com/Azure/azure-sdk-for-go/profiles/latest/dns/mgmt/dns"
```
miekg: other files have named this import "azure", please do this everywhere

plugin/azure/setup.go
```diff
@@ -0,0 +1,142 @@
...
+					if err != nil {
+						return envSettings, keys, fall, c.Errf("cannot set azure environment: %s", err.Error())
+					}
+				} else {
+					return envSettings, keys, fall, c.ArgErr()
+				}
+			case "fallthrough":
+				fall.SetZonesFromArgs(c.RemainingArgs())
+			default:
+				return envSettings, keys, fall, c.Errf("unknown property '%s'", c.Val())
```
miekg: `envSettings` can be shortened to just `env`

plugin/azure/azure.go
```diff
@@ -0,0 +1,227 @@
...
+	"github.com/coredns/coredns/request"
+
+	azure "github.com/Azure/azure-sdk-for-go/profiles/latest/dns/mgmt/dns"
+	"github.com/miekg/dns"
+)
+
+type zone struct {
+	id  string
+	z   *file.Zone
+	dns string
```
miekg: `dns` is oddly named, file.Zone also has a field origin that you mean to reuse that field? If it needs to be here (I think you can use file.Zone.origin) is should be named `name` or `origin`

plugin/azure/azure.go
```diff
@@ -0,0 +1,227 @@
...
+
+// Run updates the zones once and starts a goroutine to do it every minute
+func (h *Azure) Run(ctx context.Context) error {
+	if err := h.updateZones(ctx); err != nil {
+		return err
+	}
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
```
miekg, line 73: What's stopping this goroutine?

plugin/azure/azure.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+	}()
+	return nil
+}
+
+func (h *Azure) updateZones(ctx context.Context) error {
+	errc := make(chan error)
+	defer close(errc)
+	for zName, z := range h.zones {
+		go func(zName string, z []*zone) {
```
miekg: Is it important that these zones are updated in a go-routine each, or can update them in sequence (and so away with all the channel business here)

plugin/azure/azure.go
```diff
@@ -0,0 +1,227 @@
...
+	if len(m.Answer) == 0 && result != file.NoData && h.Fall.Through(qname) {
+		return plugin.NextOrFailure(h.Name(), h.Next, ctx, w, r)
+	}
+
+	switch result {
+	case file.Success:
+	case file.NoData:
+	case file.NameError:
+		m.Rcode = dns.RcodeNameError
+	case file.Delegation:
```
miekg, line 282: are delegations supported in azure dns? Seeing only 4 record type are supported in this plugin, delegations seems like an advanced feature (although NS records are there)

plugin/azure/azure_test.go
```diff
@@ -0,0 +1,149 @@
...
+import (
+	"context"
+	"reflect"
+	"testing"
+
+	"github.com/coredns/coredns/plugin/file"
+	"github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/coredns/coredns/plugin/pkg/fall"
+	"github.com/coredns/coredns/plugin/test"
+	crequest "github.com/coredns/coredns/request"
```
miekg: why are renaming the request import? No other plugin does this

plugin/azure/azure_test.go
```diff
@@ -0,0 +1,149 @@
...
+		"example.org.  300 IN  AAAA   2001:db8:85a3::8a2e:370:7334",
+		"www.example.org.  300 IN  A   1.2.3.4",
+		"www.example.org.  300 IN  A   1.2.3.4",
+		"org.	172800	IN	NS	ns3-06.azure-dns.org.",
+		"org.	300	IN	SOA	ns1-06.azure-dns.com. azuredns-hostmaster.microsoft.com. 1 3600 300 2419200 300",
+	} {
+		r, _ := dns.NewRR(rr)
+		err := newZ.Insert(r)
+		if err != nil {
+			log.Fatalf("XXXXXXXXXXXXXX could not add file zone to zone: %s", err.Error())
```
miekg: xxxxxx ?

miekg: also `log` in a test? Just return the error

plugin/azure/azure_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+func returnTestHandler() test.HandlerFunc {
+	return func(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+		state := crequest.Request{W: w, Req: r}
+		qname := state.Name()
+		m := new(dns.Msg)
+		rcode := dns.RcodeServerFailure
+		if qname == "example.gov." {
+			m.SetReply(r)
+			rr, err := dns.NewRR("example.gov.  300 IN  A   2.4.6.8")
```
miekg: dns.NewRR - > test.A

plugin/azure/azure_test.go
```diff
@@ -0,0 +1,149 @@
...
+			m.Authoritative = true
+			rcode = dns.RcodeSuccess
+		}
+		m.SetRcode(r, rcode)
+		w.WriteMsg(m)
+		return rcode, nil
+	}
+}
+
+func TestAzure(t *testing.T) {
```
miekg, line 69: how about non-supported RR types?

plugin/azure/setup.go
```diff
@@ -0,0 +1,142 @@
+package azure
+
+import (
+	"context"
+	"os"
+	"strings"
+
+	"github.com/coredns/coredns/core/dnsserver"
+	"github.com/coredns/coredns/plugin"
+	pkgFall "github.com/coredns/coredns/plugin/pkg/fall"
```
miekg: why this import rename?

plugin/azure/setup.go
```diff
@@ -0,0 +1,142 @@
...
+		return err
+	}
+	up := upstream.New()
+	ctx := context.Background()
+	dnsClient := dns.NewRecordSetsClient(envSettings.Values[auth.SubscriptionID])
+	dnsClient.Authorizer, err = envSettings.GetAuthorizer()
+	if err != nil {
+		return c.Errf("failed to create azure plugin: %v", err)
+	}
+	h, err := New(ctx, dnsClient, keys, up)
```
miekg: nit: upstream.New() can just be called here, no need to assign to `up`

plugin/azure/setup_test.go
```diff
@@ -0,0 +1,72 @@
+package azure
+
+import (
+	"testing"
+
+	"github.com/caddyserver/caddy"
+)
+
+func TestSetupRoute53(t *testing.T) {
```
miekg: uhm. Azure? Or just TestSetup

miekg:
this is heading in the right direction, but still needs quite a bit of work

plugin/azure/README.md
```diff
@@ -0,0 +1,83 @@
...
+~~~ txt
+{
+  "clientId": "123abc-123abc-123abc-123abc",
+  "clientSecret": "123abc-123abc-123abc-123abc",
+  "subscriptionId": "123abc-123abc-123abc-123abc",
+  "tenantId": "123abc-123abc-123abc-123abc"
+}
+
+~~~
+Enable route53 with explicit Azure credentials:
```
yongtang: `route53` -> `azure dns`?

palash25:
Hi @darshanime this was supposed to be my project under the [community bridge programme](https://communitybridge.dev.platform.linuxfoundation.org/). Since you have already worked on this I am not sure if I will be able to work on this as my project. Its ok to pick up open issues but at least care to go through the whole issue thread before sending a PR https://github.com/coredns/coredns/issues/2820#issuecomment-499859125

darshanime:
Hi @palash25, 
I was not aware of that when I started working on this, apologies for that. However, since I did this indepedently (and not under the community bridge programme), the CoreDNS slot is still open and you may be able to apply with some other issue. Best of luck!

plugin/azure/setup.go
darshanime: the [`auth.GetSettingsFromFile`](https://github.com/coredns/coredns/pull/2945/files#diff-12ffff5bbde7e904a84524ed836ceaa9R92) needs the env var. 
```
func GetSettingsFromFile() (FileSettings, error) {
	s := FileSettings{}
	fileLocation := os.Getenv("AZURE_AUTH_LOCATION")
	if fileLocation == "" {
		return s, errors.New("environment variable AZURE_AUTH_LOCATION is not set")
	}
        ...
```

We can remove the feature to supply creds via a file altogether.

plugin/azure/azure.go
darshanime: renamed it to `zone`

plugin/azure/setup.go
miekg: yes, please figure something out, this is a really weird way of doing things

plugin/azure/azure.go
darshanime: ah, missed passing it to [h.client.ListByDNSZone](https://github.com/coredns/coredns/pull/2945/files#diff-542a240e942503d1bcca9fd6f204f727R100)

plugin/azure/azure.go
darshanime: this is inspired by `route53` which uses multiple goroutines to update the zones.

plugin/azure/setup.go
darshanime: sure, I'll see if there is an alternative way, else will scrape the `auth_location` directive.

plugin/azure/setup.go
darshanime: `keyPairs` is a set for `resource-set:hosted-zone` entries in `Corefile`. 
`keys` stores the mappings `resource-set: [hosted-zone1, hosted-zone2]`. 

We can do away with `keyPairs` but then we will have to code to check for presence in the `[]string` slice of the corresponding `resource-set` in `keys`which is uglier.

miekg: don't sound this is critical for performance. Also the naming can be improved. `keys` is very generic, cause it is in effect zones?

plugin/azure/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+*   **`resource-group`** The resource group to which the dns hosted zones belong on Azure
+
+*   **`dns-zone`** the dns zone that contains the resource record sets to be
+    accessed.
+
+*   **`auth_location`** the path to the Azure CLI authentication file. You can also provide the credentials individually (`tenant_id`, `client_id`, `client_secret`, `subscription_id`)
+
+*   `fallthrough` If zone matches and no record can be generated, pass request to the next plugin.
+    If **[ZONES...]** is omitted, then fallthrough happens for all zones for which the plugin is
```
miekg: drop the block quotes here. (This should be changed in route53 as well)

plugin/azure/azure.go
darshanime: There are other Record types supported too, added them 👍

plugin/azure/azure.go
darshanime: Deletations are supported - 
> Azure DNS allows you to host a DNS zone and manage the DNS records for a domain in Azure. 

https://docs.microsoft.com/en-us/azure/dns/dns-domain-delegation

plugin/azure/azure_test.go
darshanime: this is to test for fallthrough, it should be `example.gov` no?

plugin/azure/azure_test.go
darshanime: added more tests

plugin/azure/setup.go
darshanime: renamed `keys` -> `resourceGroupSet`, `keyPairs` -> `resourceGroupMapping`

plugin/azure/setup.go
darshanime: Hmm, I tried this with `reload` plugin and the updates seem to work as is. 
CMIIW but doesn't [`instance.Restart(corefile)`](https://github.com/coredns/coredns/blob/f8bba51f844a9ba9add993dcd579288d3af3707e/plugin/reload/reload.go#L83) take care of this? I don't see `routet53` needing to handle this either.

darshanime:
@miekg @yongtang this is ready for review again
- [x] addressed PR comments
- [x] added all remaining DNS resource records supported by Azure, added specs
- [x] updated readme
- [x] rebased

plugin/azure/azure.go
miekg: I think this can be just sequential

plugin/azure/azure_test.go
miekg: ok, that warrents a comment then

plugin/azure/setup.go
miekg: ok

plugin/azure/setup_test.go
```diff
@@ -0,0 +1,72 @@
...
+}`, true},
+
+		{`azure resource-set:hosted-zone {
+    upstream
+}`, true},
+		{`azure resource-set:hosted-zone {
+    foobar
+}`, true},
+		{`azure resource-set:hosted-zone {
+    tenant_id <tenant_id>
```
miekg: don't bring these <foo> in back in the tests. This also still needs to be fixed in the README.md itself.

miekg (changes requested):
looks like we're getting there. Still a couple things to do, but then this should be good to go.

plugin/azure/azure.go
darshanime: sure will do, should I make it sequential for route53 too?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add plugin fo..." ]
>sure will do, should I make it sequential for route53 too?

... probably? Depends on why this was original done? Also depends how much 
in sync you want to keep the codebase for these 2 plugins /cc @dilyevsky

darshanime:
@miekg @yongtang please take a look

- [x] addressed PR comments
- [x] rebased

I've also added an `OWNERS` file. I'd be happy to help maintain this plugin.

plugin/azure/README.md
```diff
@@ -0,0 +1,52 @@
...
+## Description
+
+The azure plugin is useful for serving zones from resource record
+sets in Microsoft Azure DNS. The `azure` plugin can be used when coredns is deployed on Azure or elsewhere.
+This plugin supports all the DNS records supported by Azure, viz. A, AAAA, CAA, CNAME, MX, NS, PTR, SOA, SRV, and TXT record types.
+
+## Syntax
+
+~~~ txt
+azure resource-group:dns-zone {
```
miekg: please fix this readme (again)

plugin/azure/azure.go
```diff
@@ -0,0 +1,293 @@
...
+	"sync"
+	"time"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/file"
+	"github.com/coredns/coredns/plugin/pkg/fall"
+	"github.com/coredns/coredns/plugin/pkg/upstream"
+	"github.com/coredns/coredns/request"
+
+	azure "github.com/Azure/azure-sdk-for-go/profiles/latest/dns/mgmt/dns"
```
miekg: is good to rename is, but doesn't this rename now clash the with package name? Maybe azuredns as an alternative?

plugin/azure/azure.go
```diff
@@ -0,0 +1,293 @@
...
+
+type zone struct {
+	id   string
+	z    *file.Zone
+	zone string
+}
+
+type zones map[string][]*zone
+
+// Azure is the core struct of the azure plugin
```
miekg: closing dot

plugin/azure/azure.go
```diff
@@ -0,0 +1,293 @@
...
+	Next      plugin.Handler
+	Fall      fall.F
+	zoneNames []string
+	client    azure.RecordSetsClient
+	upstream  *upstream.Upstream
+	zMu       sync.RWMutex
+	zones     zones
+}
+
+// New validates the input dns zones and initializes the Azure struct
```
miekg: s/dns/DNS/
`... returns an initialized pointer to an Azure structure.`

plugin/azure/azure.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	return &Azure{
+		client:    dnsClient,
+		zones:     zones,
+		zoneNames: zoneNames,
+		upstream:  up,
+	}, nil
+}
+
+// Run updates the zones once and starts a goroutine to do it every minute.
```
miekg: `Run updates the zone from azure.`

plugin/azure/azure.go
```diff
@@ -0,0 +1,293 @@
...
+					Class:  dns.ClassINET,
+					Ttl:    resultTTL},
+				Target: dns.Fqdn(*CNAME)}
+			newZ.Insert(&cname)
+		}
+	}
+	return newZ
+}
+
+// ServeDNS uses the azure plugin to serve dns requests
```
miekg: `ServeDNS implements the plugin.Handler interface.`

plugin/azure/azure_test.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/coredns/coredns/plugin/file"
+	"github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/coredns/coredns/plugin/pkg/fall"
+	"github.com/coredns/coredns/plugin/test"
+	"github.com/coredns/coredns/request"
+
+	"github.com/miekg/dns"
+)
+
+var demoAzure Azure = Azure{
```
miekg:
```suggestion
var demoAzure  = Azure{
```

plugin/azure/azure_test.go
```diff
@@ -0,0 +1,180 @@
...
+	Next:      testHandler(),
+	Fall:      fall.Zero,
+	zoneNames: []string{"example.org", "www.example.org", "example.org.", "sample.example.org."},
```
miekg: half of these zones are not fully qualified.

plugin/azure/setup.go
```diff
@@ -0,0 +1,128 @@
...
+	"strings"
+
+	"github.com/coredns/coredns/core/dnsserver"
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/pkg/fall"
+	clog "github.com/coredns/coredns/plugin/pkg/log"
+	"github.com/coredns/coredns/plugin/pkg/upstream"
+
+	azure "github.com/Azure/azure-sdk-for-go/profiles/latest/dns/mgmt/dns"
+	azureCloud "github.com/Azure/go-autorest/autorest/azure"
```
miekg: azureCloud - azurerest ? (don't use uppercase for imports)

plugin/azure/README.md
darshanime: Sorry for the back and forth on the README. 
I am not sure about this particular case though, what changes do are you pointing at here?

plugin/azure/setup.go
```diff
@@ -0,0 +1,128 @@
...
+	caddy.RegisterPlugin("azure", caddy.Plugin{
+		ServerType: "dns",
+		Action:     setup,
+	})
+}
+
+func setup(c *caddy.Controller) error {
+	env, keys, fall, err := parse(c)
+	if err != nil {
+		return err
```
miekg: return a `plugin.Error("azure", err)` here.

plugin/azure/setup.go
```diff
@@ -0,0 +1,128 @@
...
+				return env, resourceGroupMapping, fall, c.Errf("conflict zone '%s'", args[i])
+			}
+
+			resourceGroupSet[args[i]] = struct{}{}
+			resourceGroupMapping[resourceGroup] = append(resourceGroupMapping[resourceGroup], zoneName)
+		}
+		for c.NextBlock() {
+			switch c.Val() {
+			case "subscription_id":
+				if c.NextArg() {
```
miekg: you can drop all the `else` parts here (and below) if you reverse the check
~~~
if !c.NextArg() {
   return ...
}
// set the value

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add plugin fo..." ]
>Sorry for the back and forth on the README.
>I am not sure about this particular case though, what changes do are you pointing at here?

values are CAPITALIZED, as they are in the route53 (and all other plugins)

plugin/azure/README.md
```diff
@@ -0,0 +1,52 @@
...
+~~~ txt
+azure RESOURCE-GROUP:DNS-ZONE {
+    tenant_id TENANT_ID
+    client_id CLIENT_ID
+    client_secret CLIENT_SECRET
+    subscription_id SUBSCRIPTION_ID
+}
+~~~
+
+*   **`resource-group`** The resource group to which the dns hosted zones belong on Azure
```
miekg: this now also needs to be fixed
~~~
**RESOURCE_GROUP** 
~~~
etc. Also please use underscores for resource-group and dns-zone
And dns-zone can be just 'ZONE'
CLIENT_SECRET -> SECRET
SUBSCRIPTION_ID -> SUBSCRIPTION

plugin/azure/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+*   **`dns-zone`** the dns zone that contains the resource record sets to be
```
miekg: 'the dns zone' -> 'the zone'

plugin/azure/README.md
```diff
@@ -0,0 +1,52 @@
...
+    accessed.
+
+*   `fallthrough` If zone matches and no record can be generated, pass request to the next plugin.
```
miekg, line 29: fallthrough is not mentioned in the syntax above

plugin/azure/README.md
```diff
@@ -0,0 +1,52 @@
...
+    If **ZONES** is omitted, then fallthrough happens for all zones for which the plugin is
+    authoritative. If specific zones are listed (for example `in-addr.arpa` and `ip6.arpa`), then
+    only queries for those zones will be subject to fallthrough.
+
+*   **ZONES** zones it should be authoritative for. If empty, the zones from the configuration block
```
miekg: where is ZONES?

plugin/azure/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+*   `environment` the azure environment to use. Defaults to `AzurePublicCloud`. Possible values: `AzureChinaCloud`, `AzureGermanCloud`, `AzurePublicCloud`, `AzureUSGovernmentCloud`.
```
miekg, line 34: where is environment?

plugin/azure/README.md
```diff
@@ -0,0 +1,52 @@
...
+Enable `azure` with Azure credentials and fallthrough:
+
+~~~ txt
+. {
+    azure resource-group-foo:foo.com {
+      tenant_id 123abc-123abc-123abc-123abc
+      client_id 123abc-123abc-123abc-123abc
+      client_secret 123abc-123abc-123abc-123abc
+      subscription_id 123abc-123abc-123abc-123abc
+      fallthrough example.gov.
```
miekg:
```suggestion
      fallthrough example.org.
```

plugin/azure/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+The azure plugin is useful for serving zones from resource record
+sets in Microsoft Azure DNS. The `azure` plugin can be used when coredns is deployed on Azure or elsewhere.
+This plugin supports all the DNS records supported by Azure, viz. A, AAAA, CAA, CNAME, MX, NS, PTR, SOA, SRV, and TXT record types.
+
+## Syntax
+
+~~~ txt
+azure RESOURCE-GROUP:DNS-ZONE {
+    tenant_id TENANT_ID
```
miekg: think it would be cleaner to drop all `_id`s from this; just "tenant" ?

miekg (changes requested):
README still needs a bit of work

miekg:
@yongtang I believe the code is ready to be merged, you?

The documentation still needs some work

plugin/azure/README.md
darshanime: The example mentions a minimal configuration. `environment` has a default value and need not be mentioned always. 
This documentation is here to let the user know that this is a possible configuration that can be changed.

plugin/azure/README.md
darshanime: `tenant_id`, `subscription_id` etc is the jargon used in the azure community. Should I drop the `_id`s?

plugin/azure/README.md
darshanime: the example above is minimal, should I include all possible configurations? How do you suggest I indicate it's optional without using the `[]` braces?

plugin/azure/README.md
darshanime: chaning the key to `tenant` but README value as `TENANT_ID` will keep things clear.

plugin/azure/README.md
```diff
@@ -0,0 +1,50 @@
...
+## Description
+
+The azure plugin is useful for serving zones from resource record
+sets in Microsoft Azure DNS. The `azure` plugin can be used when coredns is deployed on Azure or elsewhere.
+This plugin supports all the DNS records supported by Azure, viz. A, AAAA, CAA, CNAME, MX, NS, PTR, SOA, SRV, and TXT record types.
+
+## Syntax
+
+~~~ txt
+azure RESOURCE_GROUP:ZONE {
```
miekg:
```suggestion
azure RESOURCE_GROUP:ZONE... {
    tenant TENANT
    client CLIENT SECRET
    subscription SUBSCRIPTION_ID
}
~~~

* **RESOURCE_GROUP:ZONE** The resource group to which the hosted zones belong on Azure and the 
  **ZONE** where the *azure* plugin should be authoritative for. 

* `client` defines the **CLIENT** and **SECRET** used to connect to Azure.
```

plugin/azure/setup.go
```diff
@@ -0,0 +1,123 @@
...
+				if !c.NextArg() {
+					return env, resourceGroupMapping, fall, c.ArgErr()
+				}
+				env.Values[auth.ClientID] = c.Val()
+			case "secret":
+				if !c.NextArg() {
+					return env, resourceGroupMapping, fall, c.ArgErr()
+				}
+				env.Values[auth.ClientSecret] = c.Val()
+			case "environment":
```
miekg, line 104: what does environment does that secret doesn't do? Is set the secret and then uses that value to get the env from azure?
Is that how azure works?

plugin/azure/README.md
```diff
@@ -0,0 +1,50 @@
+# azure
+
+## Name
+
+*azure* - enables serving zone data from Microsoft Azure DNS service.
+
+## Description
+
+The azure plugin is useful for serving zones from resource record
```
miekg:
```suggestion
The *azure* plugin is useful for serving zones
```

plugin/azure/README.md
```diff
@@ -0,0 +1,50 @@
...
+sets in Microsoft Azure DNS. The `azure` plugin can be used when coredns is deployed on Azure or elsewhere.
```
miekg:
```suggestion
from Microsoft Azure DNS. 
```

plugin/azure/README.md
```diff
@@ -0,0 +1,50 @@
...
+This plugin supports all the DNS records supported by Azure, viz. A, AAAA, CAA, CNAME, MX, NS, PTR, SOA, SRV, and TXT record types.
```
miekg:
```suggestion
The *azure* plugins supports all the DNS records supported by Azure, viz. A, AAAA, CAA, CNAME, MX, NS, PTR, SOA, SRV, and TXT record types. For other types the result with be NOTIMPL.
```

miekg: what is the result for a query for a type that is not supported?

plugin/azure/README.md
```diff
@@ -0,0 +1,50 @@
...
+*   `fallthrough` If zone matches and no record can be generated, pass request to the next plugin.
+    If **ZONES** is omitted, then fallthrough happens for all zones for which the plugin is
+    authoritative. If specific zones are listed (for example `in-addr.arpa` and `ip6.arpa`), then
+    only queries for those zones will be subject to fallthrough.
+
+*   `environment` the azure environment to use. Defaults to `AzurePublicCloud`. Possible values: `AzureChinaCloud`, `AzureGermanCloud`, `AzurePublicCloud`, `AzureUSGovernmentCloud`.
+
+## Examples
+
+Enable `azure` with Azure credentials and fallthrough:
```
miekg:
```suggestion
Enable the *azure* plugin with Azure credentials and fallthrough:
```

plugin/azure/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+~~~ txt
+. {
+    azure resource_group_foo:foo.com {
```
miekg, line 42: this makes azure handle `foo.com`, the the fallthrough has `example.org`. example.org names will never be handled by the *azure* plugin so this example doesn't work.

Also I would just list a working minimal example first and then expand on that.

plugin/azure/README.md
```diff
@@ -0,0 +1,50 @@
...
+      tenant 123abc-123abc-123abc-123abc
+      client 123abc-123abc-123abc-123abc
+      secret 123abc-123abc-123abc-123abc
+      subscription 123abc-123abc-123abc-123abc
+      fallthrough example.org.
+    }
+}
+~~~
```
miekg, line 49: Are there links to Azure that we can put in a `## Also See` section for folks interested in Azure DNS?

plugin/azure/setup.go
darshanime: `environment` is used to talk to non public clouds, like  `AzureChinaCloud`, `AzureGermanCloud`, `AzurePublicCloud`, `AzureUSGovernmentCloud` etc. 
By default, the plugin talks to `AzurePublicCloud`

miekg:
I think this is good to go - I might have some nits, but I'll just send a follow PR.

Expect an invite for becoming a maintainer soon and can you ping me your email addr (miek@miek.nl) so I can add you to the maintainer lists? Thanks!

You only need to fix the conflicts - then we're good to go.

PR #2949, opened 2019-07-01
WIP: plugin/sign: add README for DNSSEC

miekg:
Talk about how DNSSEC options would look like, if implemented.

### DIFF
--- a/plugin/file/README.md
+++ b/plugin/file/README.md
@@ -6,27 +6,10 @@
 
 ## Description
 
-The *file* plugin is used for an "old-style" DNS server. It serves from a preloaded file that exists
-on disk.
-
-If the zone file contains signatures (i.e., is signed using DNSSEC), correct DNSSEC answers are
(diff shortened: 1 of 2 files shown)


miekg:
I forgot who else is interested in this; but this is an initial dump of what I'm planning to add to the file plugin.

johnbelamaric:
1) why not a separate plugin for this?
2) could coredns-keygen also be a plugin or is that nonsensical?

miekg:
Potentially it can. Only needs access to the zone files and keys.

On keygen: we can't have a main function in plugins.

miekg:
yep, separate plugin is, in fact, very easy to do. Propose the name *sign* for this, as *dnssec* is already taken.
Need to copy and paste some things from the *dnssec* plugin into this one and code a bit, but looks fairly easy.

Added benefit; I think this will now work out of the box with the *auto* plugin as well.

miekg:
coredns-keygen can be completely kept outside the main repo.

miekg:
closing for now, while I work on this.

PR #2956, opened 2019-07-02
plugins: set upstream unconditionally

### LINKED ISSUE #2954: Remove 'upstream' from plugins
In the 1.3.1 release notes we menioned we're going to change/do something about upstream. The time has come to clean this up in all plugins:

* for the time being allow `upstream` and any parameters, but ignore it entirely, so set a `upstream.New` in the code (this may or may not have been already done - haven't checked)

* some time later; remove upstream as a config option from each of the plugins and default to erroring when we see it.

miekg:
`upstream` is not needed as a setting; just set if unconditionally and
remove all documentation and tests for it.

At some point we want remove the hanlding for `upstream` as well and
error out on seeing it.

Fixes #2954

Remove 'upstream' from all docs and tests.

No, not yet; this has been promised since 1.3.1

### DIFF
--- a/plugin/auto/README.md
+++ b/plugin/auto/README.md
@@ -18,7 +18,6 @@ auto [ZONES...] {
     directory DIR [REGEXP ORIGIN_TEMPLATE]
     transfer to ADDRESS...
     reload DURATION
-    upstream
 }
 ~~~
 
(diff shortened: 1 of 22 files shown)


chrisohaver (approved):
lgtm

PR #2958, opened 2019-07-02
plugin/template: support metadata

johnbelamaric:
This allows metadata values to be used inside of template responses.

None.

README is updated in the PR.

No.

### DIFF
--- a/plugin/template/template.go
+++ b/plugin/template/template.go
@@ -8,6 +8,7 @@ import (
 	gotmpl "text/template"
 
 	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/metadata"
 	"github.com/coredns/coredns/plugin/metrics"
 	"github.com/coredns/coredns/plugin/pkg/fall"
 	"github.com/coredns/coredns/plugin/pkg/upstream"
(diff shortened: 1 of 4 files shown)


chrisohaver (approved):
LGTM.  I think adding an example using metadata would be good to have too.

plugin/template/template.go
```diff
@@ -47,6 +48,15 @@ type templateData struct {
 	Type     string
 	Message  *dns.Msg
 	Question *dns.Question
+	ctx      context.Context
```
miekg: Can we go forward without adding a context to a type? It should not be needed to have it here

johnbelamaric: The problem is that this is fed into `text/template` so it needs all that state, and doesn't have the Context. The real answer would be for `text/template` to accept a `Context` but that's not gonna happen anytime soon. I need the context to call `metadata.ValueFunc(ctx, key)`. The actual metadata object can't be pulled out of the context because it is keyed by a private type from `metadata`.

johnbelamaric: Options:

1) As is
2) Add a function in `metadata` that will return the map[string]func from the Context, and store that in the struct.

johnbelamaric: I implemented #2 here.

johnbelamaric:
One thought - maybe instead of just meta data, we should implement the full list from https://github.com/coredns/coredns/tree/master/plugin/log#log-format

johnbelamaric:
A bunch of those are circular. Mostly I am thinking the client IP:port and maybe server IP:port

miekg (approved):

PR #2967, opened 2019-07-03
plugin/metrics: remove unused code

miekg:
This isn't being used anymore. Wanted to remove the log.Fatalf, but the
whole funtion isn't called.


miekg:
-ENOCOFFEE, it's use, loosing log.Fatal here make sense.

PR #2976, opened 2019-07-04
Add whitelist option to forward plugin

eb4890:

I was thinking about ways of preventing DNS exfiltration.

Currently at $WORK we already whitelist the servers we talk to
Why not also allow a whitelist of dns names that you can request?

So I sat down to see how hard it would be and here is the first attempt.

I plan to write some more tests, but I thought I would find out if this is the right place in the plugins before committing more effort. 

None

I've made changes to the README in the plugin and the associated man page to document the new option.

Nope, it is adding a new option to a plugin.

### DIFF
--- a/man/coredns-forward.7
+++ b/man/coredns-forward.7
@@ -59,6 +59,7 @@ Extra knobs are available with an expanded syntax:
 .nf
 forward FROM TO... {
     except IGNORED\_NAMES...
+    whitelist ALLOWED_NAMES...
     force\_tcp
     prefer\_udp
     expire DURATION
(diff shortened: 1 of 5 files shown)


miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] Add whitelist opt..." ]
>
>I was thinking about ways of preventing DNS exfiltration.
>
>Currently at $WORK we already whitelist the servers we talk to
>Why not also allow a whitelist of dns names that you can request?
>
>So I sat down to see how hard it would be and here is the first attempt.

This can be done better with either:

* the rewrite or template plugin
* various block plugins like: https://github.com/miekg/block

eb4890:
This can't be done with the block plugin as it stands, as to prevent exfiltration you want to block everything *not* on a list.

The only problem with using the rewrite or template plugins is that if you have lots of domains to allow through the regexes might become large and hard to read (or you need multiple steps, such as marking the allowed domains and then rewriting all non-marked domains).

Happy to close it for now, but I think there is space for something that makes the UX nicer for this.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add whitelist..." ]
>Happy to close it for now, but I think there is space for something that makes the UX nicer for this.

Ack. But just blacklisting domain seems like an approach that is destined for 
failure. A better approach would be to do anomaly detection, and then (maybe) 
start blocking; this then integrated with metrics and alerting.

eb4890:
For context my use case is in a heavily locked down environment that only talks to a known handful of domains (this is for card payment processing, so things like payment providers/databasese and other supporting services). 

I wouldn't do a whitelist like I am proposing in an end user environment. What I am trying to avoid is https://blogs.akamai.com/2017/09/introduction-to-dns-data-exfiltration.html

Issue #2986, opened 2019-07-08
plugin/file: revisit file.Tree


miekg:
For the sign plugin I'm going over the file/tree implementation again (first time since a year I think). and I see gems like:
~~~
// Types returns the RRs with type qtype from e. If qname is given (only the
// first one is used), the RR are copied and the owner is replaced with qname[0].
func (e *Elem) Types(qtype uint16, qname ...string) []dns.RR {
	rrs := e.m[qtype]

	if rrs != nil && len(qname) > 0 {
		copied := make([]dns.RR, len(rrs))
		for i := range rrs {
			copied[i] = dns.Copy(rrs[i])
			copied[i].Header().Name = qname[0]
		}
		return copied
	}
	return rrs
}
~~~
... and I'm wondering WTH is this copying all this stuff.

miekg:
equalRdata is also not finished; we could
* finish it
* not care about this

miekg:
Also means revisiting the reload locking (#2975).

Pondering deleting/fixing a whole bunch of things. I also need to neat function that gives my all RRsets under a Elem for easy signing

miekg:
The above is wildcard related, but the code could have been documented better. Preparing a little cleanup PR once I (re-)figure out some bits.

miekg:
There is other stuff; now that I'm writing a signer. The tree.Do function is handy, but for instance, for *not* signing glue and would be useful to cull entire subtree to avoid signing glue records after seeing a delegated name. The tree contains all the info, so it would be a bit wasteful to has another thing carry this information

miekg:
I'm still planning to add a Walk function that skips certain subtree. For the rest this looks pretty good and #2987 made it better.

PR #2987, opened 2019-07-09
plugin/file: Rename do to walk and cleanup and document

miekg:
* This renames Do to Walk to be more inline with Go standards.
* It adds Types to TypesForWildcard to show the different functions
* *Removes* the identical RR check when inserting; this was only done
  for A, AAAA and MX and not finished; removed under the mantra garbage
  in garbage out.

### DIFF
--- a/plugin/file/tree/tree.go
+++ b/plugin/file/tree/tree.go
@@ -275,7 +275,8 @@ func (n *Node) deleteMax() (root *Node, d int) {
 	return
 }
 
-// Delete removes rr from the tree, is the node turns empty, that node is deleted with DeleteNode.
+// Delete removes all RRs of type rr.Header().Rrtype from e. If after the deletion of rr the node is empty the
+// entire node is deleted.
 func (t *Tree) Delete(rr dns.RR) {
 	if t.Root == nil {
 		return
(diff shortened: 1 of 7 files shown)


miekg:
See #2986

miekg:
this would  be good to merge as a simplifies some bits. It helps me for *sign* and I also have some ideas to fix the reload deadlock

plugin/file/tree/walk.go
```diff
@@ -0,0 +1,30 @@
+package tree
+
+import "github.com/miekg/dns"
+
+// Walk performs fn on all values stored in the tree. If a non-nill error is returned the
```
stp-ip:
```suggestion
// Walk performs fn on all values stored in the tree. If a non-nil error is returned the
```

plugin/file/tree/tree.go
```diff
@@ -286,12 +286,9 @@ func (t *Tree) Delete(rr dns.RR) {
...
 		return
 	}
-	// Delete from this element.
-	empty := el.Delete(rr)
-	if empty {
-		t.deleteNode(rr)
-		return
-	}
+	el.Delete(rr)
+	t.deleteNode(rr)
```
stp-ip: Why delete the node either way?

stp-ip (commented):
Still working through the code. I agree it makes it a bit cleaner, but not sure I am comfortable giving a lgtm yet.

plugin/file/tree/tree.go
miekg: good question. Let me also add some test for this, because we don't have many... Note delete/walk is currently unused.

miekg: Added test and change this bit. Thanks for catching this.

miekg:
I intent to merge this soon. I also need another function which is hard to do externally. A `WalkAuthoritative` this skips authoritative bits in a zone (i.e. records under a delegation).

Once these bits are in the locking can be simplified to only have 1 mutex, for reload and transfer.

Issue #2992, opened 2019-07-10
plugin/rewrite: false-negative check of long non-ascii domain name


rdrozhdzh:
corefile:

```
.:1053 {
    forward . 8.8.8.8
    rewrite edns0 subnet append 24 56
    errors
}
```
coredns responds with SERVFAIL to valid long non-ascii domain name
```
$ dig @127.0.0.1 -p 1053 длинноепредлинноенелатинскоеимя.toolongforrewrite.com

; <<>> DiG 9.11.3-1ubuntu1.8-Ubuntu <<>> @127.0.0.1 -p 1053 длинноепредлинноенелатинскоеимя.toolongforrewrite.com
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 9879
;; flags: qr rd; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
; COOKIE: ad423c33c9d81ebd (echoed)
;; QUESTION SECTION:
;\208\180\208\187\208\184\208\189\208\189\208\190\208\181\208\191\209\128\208\181\208\180\208\187\208\184\208\189\208\189\208\190\208\181\208\189\208\181\208\187\208\176\209\130\208\184\208\189\209\129\208\186\208\190\208\181\208\184\208\188\209\143.toolongforrewrite.com.	IN A

;; Query time: 0 msec
;; SERVER: 127.0.0.1#1053(127.0.0.1)
;; WHEN: Wed Jul 10 20:29:40 +03 2019
;; MSG SIZE  rcvd: 125
```

coredns output:

```
.:1053
2019-07-10T20:29:19.562+03:00 [INFO] CoreDNS-1.4.0
2019-07-10T20:29:19.562+03:00 [INFO] linux/amd64, go1.12, 
CoreDNS-1.4.0
linux/amd64, go1.12, 
[...]

PR #2995, opened 2019-07-11
plugin/rewrite: fix domain length validation

rdrozhdzh:
fix domain length validation in rewrite plugin
#2992
none
no

### DIFF
--- a/plugin/rewrite/rewrite.go
+++ b/plugin/rewrite/rewrite.go
@@ -44,12 +44,10 @@ func (rw Rewrite) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	for _, rule := range rw.Rules {
 		switch result := rule.Rewrite(ctx, state); result {
 		case RewriteDone:
-			_, validName := dns.IsDomainName(state.Req.Question[0].Name)
-			if !validName {
-				x := state.Req.Question[0].Name
-				log.Errorf("Invalid name after rewrite: %s", x)
+			if _, ok := dns.IsDomainName(state.Req.Question[0].Name); !ok {
+				err := fmt.Errorf("invalid name after rewrite: %s", state.Req.Question[0].Name)
 				state.Req.Question[0] = wr.originalQuestion
-				return dns.RcodeServerFailure, fmt.Errorf("invalid name after rewrite: %s", x)
+				return dns.RcodeServerFailure, err
 			}
 			respRule := rule.GetResponseRule()
 			if respRule.Active {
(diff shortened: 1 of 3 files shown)


miekg (commented):
lgtm, but do we really need a test for this? We can hope dns.IsDomainName is tested well enough

stp-ip:
/lgtm
But is hope a good strategy? :smiling_imp:

rdrozhdzh:
Yes, I'm sure `dns.IsDomainName()` is tested well. I'd rather want to make sure that rewrite plugin doesn't do anything freaky in addition to checking with `dns.IsDomainName()` like it was before my fix.

Though, I can remove my unit test if you wish. I just added it for confirmation that my fix works as expected.

miekg:
I think it's a rather convoluted way to test if IsDomainName does what it needs to do.

More generally it would be good to test this, but then for all plugins which is is way harder to setup.

plugin/rewrite/rewrite.go
```diff
@@ -44,7 +44,8 @@ func (rw Rewrite) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	for _, rule := range rw.Rules {
 		switch result := rule.Rewrite(ctx, state); result {
 		case RewriteDone:
-			if !validName(state.Req.Question[0].Name) {
+			_, validName := dns.IsDomainName(state.Req.Question[0].Name)
```
miekg: the `if` can be pulled up, and just using `ok` is more go like:
~~~
if _, ok := dns.IsDomainName; !ok {

}
~~~
also the `x :=` seemed superfluous and why are we logging this *and* return the error, seem we can drop the `log.Errrof` here as well

rdrozhdzh: Regarding `x :=` - we we need to save the name before overwriting `Question[0]` with `originalQuestion`, the name is needed to construct the error message to be returned by `ServeDNS()`. I'm going to update the code

rdrozhdzh:
Seems the build failure is not related to my change. Not sure if I can re-run the tests
```
--- FAIL: TestRequestScrubExtra (0.00s)
    request_test.go:102: Want scrub to not set truncated bit
=== RUN   TestRequestScrubExtraEdns0
--- FAIL: TestRequestScrubExtraEdns0 (0.00s)
    request_test.go:124: Want scrub to not set truncated bit
=== RUN   TestRequestScrubExtraRegression
--- FAIL: TestRequestScrubExtraRegression (0.00s)
    request_test.go:150: Want scrub to not set truncated bit
=== RUN   TestTruncation
--- PASS: TestTruncation (0.08s)
=== RUN   TestRequestScrubAnswerExact
--- PASS: TestRequestScrubAnswerExact (0.00s)
=== RUN   TestRequestMatch
--- PASS: TestRequestMatch (0.00s)
=== RUN   TestRequestClear
--- PASS: TestRequestClear (0.00s)
FAIL
coverage: 40.1% of statements
FAIL	github.com/coredns/coredns/request	0.092s
```

chrisohaver:
re-running tests...

miekg:
It is unrelated. Bit strange this would fail now, but will fix.

miekg:
please remove the test, other then that lgtm

PR #3001, opened 2019-07-12
Scrub: TC bit is always set

miekg:
miekg/dns recently became more aggressive in setting TC; anything that's
chopped of a response results in TC. Amend the tests. Disputable if
these TC checks still add something.


miekg:
this is build breaking, so'll merge after CI comes back

PR #3002, opened 2019-07-14
replacer: evaluate format once and improve perf by ~3x

charlievieth:

Why: request logging as it currently exists is incredibly slow and wasteful.  This improves the performance of logging by almost 3x and reduces memory usage by ~8x.

Benchmark results:

```
benchmark                                old ns/op     new ns/op     delta
BenchmarkReplacer-12                     644           324           -49.69%
BenchmarkReplacer_CommonLogFormat-12     4228          1471          -65.21%

benchmark                                old allocs     new allocs     delta
BenchmarkReplacer-12                     8              2              -75.00%
BenchmarkReplacer_CommonLogFormat-12     51             17             -66.67%

benchmark                                old bytes     new bytes     delta
BenchmarkReplacer-12                     240           48            -80.00%
BenchmarkReplacer_CommonLogFormat-12     3723          446           -88.02%
```

N/A

N/A

N/A

### DIFF
--- a/plugin/pkg/replacer/replacer.go
+++ b/plugin/pkg/replacer/replacer.go
@@ -211,10 +211,7 @@ func parseFormat(s string) replacer {
 		}
 		i := strings.LastIndexByte(s[:j], '{')
 		if i < 0 {
-			// Handle: "A } {foo}"
-			//            ^j
-			//
-			// By treating "A }" as a literal
+			// Handle: "A } {foo}" by treating "A }" as a literal
 			rep = append(rep, node{
 				value: s[:j+1],
 				typ:   typeLiteral,
@@ -270,7 +267,10 @@ func loadFormat(s string) replacer {
 	return v.(replacer)
 }
 
-// bufPool stores pointers to scratch buffers
+// bufPool stores pointers to scratch buffers.  We store a pointer to the slice
+// as it saves an allocation during the slice => interface{} conversion.  If we
+// were to store the slice value a pointer to it would have to be created on
+// each Put().
 var bufPool = sync.Pool{
 	New: func() interface{} {
 		b := make([]byte, 0, 256)
--- a/plugin/pkg/replacer/replacer.go
+++ b/plugin/pkg/replacer/replacer.go
@@ -129,7 +129,6 @@ func appendValue(b []byte, state request.Request, rr *dnstest.Recorder, label st
 		}
 		return append(b, EmptyValue...)
 	default:
-		// TODO (CEV): consider panicking since this should be impossible
 		return append(b, EmptyValue...)
 	}
 }
--- a/plugin/pkg/replacer/replacer.go
+++ b/plugin/pkg/replacer/replacer.go
@@ -267,20 +267,15 @@ func loadFormat(s string) replacer {
 	return v.(replacer)
 }
 
-// bufPool stores pointers to scratch buffers.  We store a pointer to the slice
-// as it saves an allocation during the slice => interface{} conversion.  If we
-// were to store the slice value a pointer to it would have to be created on
-// each Put().
+// bufPool stores pointers to scratch buffers.
 var bufPool = sync.Pool{
 	New: func() interface{} {
-		b := make([]byte, 0, 256)
-		return &b
+		return make([]byte, 0, 256)
 	},
 }
 
 func (r replacer) Replace(ctx context.Context, state request.Request, rr *dnstest.Recorder) string {
-	p := bufPool.Get().(*[]byte)
-	b := *p
+	b := bufPool.Get().([]byte)
 	for _, s := range r {
 		switch s.typ {
 		case typeLabel:
@@ -296,7 +291,6 @@ func (r replacer) Replace(ctx context.Context, state request.Request, rr *dnstes
 		}
 	}
 	s := string(b)
-	*p = b[:0]
-	bufPool.Put(p)
+	bufPool.Put(b[:0])
 	return s
 }
--- a/plugin/pkg/replacer/replacer.go
+++ b/plugin/pkg/replacer/replacer.go
@@ -35,44 +35,25 @@ const (
 )
 
 // labels are all supported labels that can be used in the default Replacer.
-var labels = [...]string{
-	"{type}",
-	"{name}",
-	"{class}",
-	"{proto}",
-	"{size}",
-	"{remote}",
-	"{port}",
-	"{local}",
+var labels = map[string]struct{}{
+	"{type}":   {},
+	"{name}":   {},
+	"{class}":  {},
+	"{proto}":  {},
+	"{size}":   {},
+	"{remote}": {},
+	"{port}":   {},
+	"{local}":  {},
 	// Header values.
-	headerReplacer + "id}",
-	headerReplacer + "opcode}",
-	headerReplacer + "do}",
-	headerReplacer + "bufsize}",
-	// Recorded replacements.
-	"{rcode}",
-	"{rsize}",
-	"{duration}",
-	headerReplacer + "rflags}",
-}
-
-var knownLabels = map[string]struct{}{
-	"{type}":                    {},
-	"{name}":                    {},
-	"{class}":                   {},
-	"{proto}":                   {},
-	"{size}":                    {},
-	"{remote}":                  {},
-	"{port}":                    {},
-	"{local}":                   {},
 	headerReplacer + "id}":      {},
 	headerReplacer + "opcode}":  {},
 	headerReplacer + "do}":      {},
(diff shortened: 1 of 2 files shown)


plugin/pkg/replacer/replacer_test.go
```diff
@@ -32,6 +37,178 @@ func TestReplacer(t *testing.T) {
...
+	}
+	for i, x := range tests {
+		r := parseFormat(x.Format)
+		if !reflect.DeepEqual(r, x.Expected) {
+			t.Errorf("%d: Expected:\n\t%+v\nGot:\n\t%+v", i, x.Expected, r)
+		}
+	}
+}
+
+func TestParseFormat_Nodes(t *testing.T) {
```
miekg: -> `TestParseFormatNodes` or does the `_` have any significance here?

plugin/pkg/replacer/replacer.go
```diff
@@ -41,168 +56,247 @@ var labels = []string{
...
-	return b.String()
-}
-
-func boolToString(b bool) string {
-	if b {
-		return "true"
+		return append(b, EmptyValue...)
+	default:
+		// TODO (CEV): consider panicking since this should be impossible
+		return append(b, EmptyValue...)
```
miekg, line 113: can we return an error?

plugin/pkg/replacer/replacer.go
```diff
@@ -41,168 +56,247 @@ var labels = []string{
...
+	value string // Literal value, label or metadata label
+	typ   nodeType
+}
+
+// A replacer is an ordered list of all the nodes in a format.
+type replacer []node
+
+func parseFormat(s string) replacer {
+	// Assume there is a literal between each label - its cheaper to over
+	// allocate once than allocate twice.
```
miekg, line 183: ... because `append` will double the capacity? And does it really matter that much here?

plugin/pkg/replacer/replacer.go
```diff
@@ -41,168 +56,247 @@ var labels = []string{
...
+		// We find the right bracket then backtrack to find the left bracket.
+		// This allows us to handle formats like: "{ {foo} }".
+		j := strings.IndexByte(s, '}')
+		if j < 0 {
+			break
+		}
+		i := strings.LastIndexByte(s[:j], '{')
+		if i < 0 {
+			// Handle: "A } {foo}"
+			//            ^j
```
miekg: just `^`? Don't understand the "symbol" `^j`

plugin/pkg/replacer/replacer.go
```diff
@@ -41,168 +56,247 @@ var labels = []string{
...
+// bufPool stores pointers to scratch buffers
+var bufPool = sync.Pool{
+	New: func() interface{} {
+		b := make([]byte, 0, 256)
+		return &b
+	},
+}
+
+func (r replacer) Replace(ctx context.Context, state request.Request, rr *dnstest.Recorder) string {
+	p := bufPool.Get().(*[]byte)
```
miekg: you have a pointer to a slice, so you can extend the slice or something? This requires some documentation detailing why it's a pointer to a slice. Probably where the syncmap is declared.

miekg (commented):
Nice improvements ! Generally looks good, but I do have a bunch of questions.

plugin/pkg/replacer/replacer_test.go
charlievieth: Nope, will change.

plugin/pkg/replacer/replacer.go
charlievieth: Yes and no the extra allocs don't matter so long as parsing doesn't need to be very fast.  That said, I really don't blindly appending to a slice when the required size can be estimated.

Comparison of pre-allocate (current method) to appending.
```
$ benchcmp pre-allocate.txt append.txt
benchmark                   old ns/op     new ns/op     delta
BenchmarkParseFormat-12     527           918           +74.19%

benchmark                   old allocs     new allocs     delta
BenchmarkParseFormat-12     1              6              +500.00%

benchmark                   old bytes     new bytes     delta
BenchmarkParseFormat-12     704           1520          +115.91%
```

plugin/pkg/replacer/replacer.go
charlievieth: Will fix.

plugin/pkg/replacer/replacer.go
charlievieth: I'll add a comment.  The reason we use store a pointer to the slice and not the slice itself is that it saves an alloc in the value => `interface{}` conversion.

plugin/pkg/replacer/replacer.go
charlievieth: This code will only be executed if there is a bug in the `parseFormat()` or `replacer.Replace()` logic and matches the previous behavior of the now removed `value()` function.  

Basically, this would be a hard programming error so I'm not sure there would be much value in propagating this to the user as there is nothing they can do about it (nor is there an easy way to propagate this error).

Personally, I would panic as I like to treat programming/logic errors harshly so that they do not make into production (I consider the behavior of unlocking an unlocked `sync.Mutex` to be a good example of this).  Additionally, if there is a bug here it is impossible to know if the logs being generated are correct.

That said, I'm open to suggestions as to how you think this should be handled.

**TLDR:** I don't think there is much value in returning an error - since this represents a logic bug.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] replacer: eva..." ]
>I'll add a comment.  The reason we use store a pointer to the slice and not the slice itself is that it saves an alloc in the value => `interface{}` conversion.

this sounds like a really micro optimizations we shouldn't really care about. I 
doubt this makes a dent in the 3x improvement we're seeing here.

Logging will always kill your performance, doing these micro optimizations won't 
make it that much better. Just use the slice directly.

/Miek

--
Miek Gieben

plugin/pkg/replacer/replacer.go
charlievieth: Comment added to bufPool https://github.com/coredns/coredns/blob/57fa92527aa00dc0b38db10eda5fd204218f1d46/plugin/pkg/replacer/replacer.go#L270-L274

charlievieth: > [ Quoting <notifications@github.com> in "Re: [coredns/coredns] replacer: eva..." ]
> I'll add a comment. The reason we use store a pointer to the slice and not the slice itself is that it saves an alloc in the value => `interface{}` conversion.
> this sounds like a really micro optimizations we shouldn't really care about. I doubt this makes a dent in the 3x improvement we're seeing here. Logging will always kill your performance, doing these micro optimizations won't make it that much better. Just use the slice directly. /Miek
> […](#)
> -- Miek Gieben

IMO, more a matter of correctness, but I'll change it to use the value so we don't have to deal with pointers.

miekg:
>**TLDR:** I don't think there is much value in returning an error - since this represents a logic bug.

ok, then remove the comment, no-one will read logs or anything, so if we can't 
make this a failure on startup it doesn't matter.

charlievieth:
> **TLDR:** I don't think there is much value in returning an error - since this represents a logic bug.
> ok, then remove the comment, no-one will read logs or anything, so if we can't make this a failure on startup it doesn't matter.

Comment removed.  Also, thank you for taking the time to review this and for the quick feedback!

plugin/pkg/replacer/replacer.go
charlievieth: Comment removed.

charlievieth:
Thanks again for the quick feedback, please let me know if there are any other changes you'd like for me to make.

plugin/pkg/replacer/replacer.go
charlievieth: Note the comment has been removed, but is not showing up in the diff.

plugin/pkg/replacer/replacer.go
```diff
@@ -14,13 +15,27 @@ import (
...
+
+const (
+	headerReplacer = "{>"
+	// EmptyValue is the default empty value.
+	EmptyValue = "-"
+)
+
 // labels are all supported labels that can be used in the default Replacer.
-var labels = []string{
+var labels = [...]string{
```
miekg: why the `...` here?

miekg (changes requested):
looks good, one nit.

plugin/pkg/replacer/replacer.go
charlievieth: That was a nit on my part.  I like to declare what are essentially `const slices` as arrays to indicate to other programers that they are const and not to be modified.

That said, the `labels` var is no longer necessary and I will replace it with the `knownLabels` map.

PR #3003, opened 2019-07-14
rewrite: remove condition file

charlievieth:

Nothing it declares is used by CoreDNS and it does not appear to be used by any third-party packages (according to [godoc.org](https://godoc.org/github.com/coredns/coredns/plugin/rewrite?importers)).

N/A

N/A

Yes.

The following constants are removed:

```golang
const (
       Is         = "is"
       Not        = "not"
       Has        = "has"
       NotHas     = "not_has"
       StartsWith = "starts_with"
       EndsWith   = "ends_with"
       Match      = "match"
       NotMatch   = "not_match"
)
```

The `If` type is also removed as are its constructor/methods:
```golang
// If is statement for a rewrite condition.
type If struct {
       A        string
       Operator string
       B        string
}

func (i If) True(r *dns.Msg) bool { ... }

func NewIf(a, operator, b string) (If, error) { ... }
```

A review of [godoc](https://godoc.org/github.com/coredns/coredns/plugin/rewrite?importers) show this code to be unused and thus safe to remove.


miekg:
Is there anything we can do to automatically detect these things?

charlievieth:
> Is there anything we can do to automatically detect these things?

Not that I know of.  The difficulty here is that some of these decls are exported, which makes checking a little more difficult ([godoc](https://godoc.org/) is pretty helpful here).  As for finding things that are unused internally I've used [sourcegraph](https://about.sourcegraph.com/) before and while it is OSS I'm not sure what support they offer for OSS projects.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] rewrite: remo..." ]
>> Is there anything we can do to automatically detect these things?
>
>Not that I know of.  The difficulty here is that some of these decls are exported, which makes checking a little more difficult ([godoc](https://godoc.org/) is pretty helpful here).  As for finding things that are unused internally I've used [sourcegraph](https://about.sourcegraph.com/) before and while it is OSS I'm not sure what support they offer for OSS projects.

Thanks. Yeah I thought so. We can grep our source and sort off make a judgement 
all from that. I.e. if it's exported, but not used within our tree we can remove 
it. Not aware of any tool that does this though.

PR #3004, opened 2019-07-14
Fix log plugin benchmark and slightly improve performance

charlievieth:

It fixes memory/alloc reporting for `BenchmarkLogged` and provides a small improvement to the logging logic.

Benchmark results:
```
benchmark              old ns/op     new ns/op     delta
BenchmarkLogged-12     5742          5573          -2.94%

benchmark              old allocs     new allocs     delta
BenchmarkLogged-12     60             60             +0.00%

benchmark              old bytes     new bytes     delta
BenchmarkLogged-12     4536          4145          -8.62%
```

N/A

N/A

N/A

### DIFF
--- a/plugin/log/log.go
+++ b/plugin/log/log.go
@@ -26,21 +26,24 @@ type Logger struct {
 // ServeDNS implements the plugin.Handler interface.
 func (l Logger) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 	state := request.Request{W: w, Req: r}
+	name := state.Name()
 	for _, rule := range l.Rules {
-		if !plugin.Name(rule.NameScope).Matches(state.Name()) {
+		if !plugin.Name(rule.NameScope).Matches(name) {
 			continue
 		}
 
 		rrw := dnstest.NewRecorder(w)
 		rc, err := plugin.NextOrFailure(l.Name(), l.Next, ctx, rrw, r)
 
-		tpe, _ := response.Typify(rrw.Msg, time.Now().UTC())
-		class := response.Classify(tpe)
 		// If we don't set up a class in config, the default "all" will be added
 		// and we shouldn't have an empty rule.Class.
 		_, ok := rule.Class[response.All]
-		_, ok1 := rule.Class[class]
-		if ok || ok1 {
+		if !ok {
+			tpe, _ := response.Typify(rrw.Msg, time.Now().UTC())
+			class := response.Classify(tpe)
+			_, ok = rule.Class[class]
+		}
+		if ok {
 			logstr := l.repl.Replace(ctx, state, rrw, rule.Format)
 			clog.Infof(logstr)
 		}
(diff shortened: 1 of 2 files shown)


plugin/log/log.go
```diff
@@ -26,21 +26,24 @@ type Logger struct {
...
 		rc, err := plugin.NextOrFailure(l.Name(), l.Next, ctx, rrw, r)
 
-		tpe, _ := response.Typify(rrw.Msg, time.Now().UTC())
-		class := response.Classify(tpe)
 		// If we don't set up a class in config, the default "all" will be added
 		// and we shouldn't have an empty rule.Class.
 		_, ok := rule.Class[response.All]
-		_, ok1 := rule.Class[class]
-		if ok || ok1 {
+		if !ok {
```
miekg, line 42: I don't find this use of the variable `ok` readable, the `ok || ok1` might have been slower, but at least it was clearer.
Does a `switch` help here?

charlievieth: The variable name isn't great, but the goal here is to skip the call to [`response.Typify()`](https://github.com/charlievieth/coredns/blob/cev/log-perf/plugin/pkg/response/typify.go#L57) if we already know the message needs to be logged: https://github.com/coredns/coredns/blob/72c84dcb1cb7e69941eca7050153e80a5c9f1cf7/plugin/log/log.go#L40-L45

Basically, this change makes the check lazier.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fix log plugi..." ]
>-		_, ok1 := rule.Class[class]
>-		if ok || ok1 {
>+		if !ok {
>
>The variable name isn't great, but the goal here is to skip the call to [`response.Typify()`](https://github.com/charlievieth/coredns/blob/cev/log-perf/plugin/pkg/response/typify.go#L57) if we already know the message needs to be logged: https://github.com/coredns/coredns/blob/72c84dcb1cb7e69941eca7050153e80a5c9f1cf7/plugin/log/log.go#L42

sure, but I had to read this 3 times to make sure it did the right thing. This 
doesn't sound super critical for performance, so I rather have it more readable.

plugin/log/log.go
charlievieth: Fixed.  I renamed `ok` => `shouldLog`.  Happy to add comments if you think they'll be helpful.

charlievieth:
Thanks again for the quick feedback - please let me know if there are any other changes you'd like for me to make.

plugin/log/log.go
```diff
@@ -26,21 +26,24 @@ type Logger struct {
...
 		// If we don't set up a class in config, the default "all" will be added
 		// and we shouldn't have an empty rule.Class.
-		_, ok := rule.Class[response.All]
-		_, ok1 := rule.Class[class]
-		if ok || ok1 {
+		_, shouldLog := rule.Class[response.All]
+		if !shouldLog {
+			tpe, _ := response.Typify(rrw.Msg, time.Now().UTC())
+			class := response.Classify(tpe)
+			_, shouldLog = rule.Class[class]
```
miekg: naming it `shouldLog` doesnt change anything, `shouldLog` is overwritten in this if-statement, that looks very weird.

charlievieth: Ok, I thought the confusion was around the variable name itself.  I changed it to back to using `ok` and `ok1`.

PR #3011, opened 2019-07-16
Add Google Cloud DNS plugin

### LINKED ISSUE #2822: Support Google Cloud DNS backend
This was one of the candidate proposals in GSoC: (https://github.com/cncf/soc#support-google-cloud-dns-backend). Think it is still very much relevant.

CoreDNS is able to serve DNS with cloud vendors (such as AWS) as the backend. The feature is very much useful in a hybrid environment where cloud-vendor specific service endpoints need to be exposed to clusters managed by Kubernetes. The goal of this feature is to support Google Cloud DNS (similar to already supported route53 plugin) as a backend for CoreDNS.

palash25:
Closes: #2822

### DIFF
--- a/plugin/clouddns/README.md
+++ b/plugin/clouddns/README.md
@@ -1,0 +1,67 @@
+# clouddns
+
+## Name
+
+*clouddns* - enables serving zone data from GCP clouddns.
+
+## Description
+
+The clouddns plugin is useful for serving zones from resource record
+sets in GCP clouddns. This plugin supports all [Google Cloud DNS records](https://cloud.google.com/dns/docs/overview#supported_dns_record_types).
+The clouddns plugin can be used when coredns is deployed on GCP or elsewhere.
+
+## Syntax
+
+~~~ txt
+clouddns [ZONE:PROJECT_NAME:HOSTED_ZONE_NAME...] {
+    credentials [FILENAME]
+    fallthrough [ZONES...]
+}
+~~~
+
+*   **ZONE** the name of the domain to be accessed. When there are multiple zones with overlapping
+    domains (private vs. public hosted zone), CoreDNS does the lookup in the given order here.
+    Therefore, for a non-existing resource record, SOA response will be from the rightmost zone.
+
+*   **HOSTED_ZONE_NAME** the name of the hosted zone that contains the resource record sets to be
+    accessed.
+
+*   `credentials` is used for reading the credential file.
+
+*   **FILENAME** GCP credentials file path.
+
+*   `fallthrough` If zone matches and no record can be generated, pass request to the next plugin.
+    If **[ZONES...]** is omitted, then fallthrough happens for all zones for which the plugin is
+    authoritative. If specific zones are listed (for example `in-addr.arpa` and `ip6.arpa`), then
+    only queries for those zones will be subject to fallthrough.
+
+*   **ZONES** zones it should be authoritative for. If empty, the zones from the configuration block
+
+## Examples
+
+Enable clouddns with implicit GCP credentials and resolve CNAMEs via 10.0.0.1:
+
+~~~ txt
+. {
+	clouddns example.org.:gcp-example-project:example-zone
+    forward . 10.0.0.1
+}
+~~~
+
+Enable clouddns with fallthrough:
+
+~~~ txt
+. {
+    clouddns example.org.:gcp-example-project:example-zone clouddns example.com.:gcp-example-project:example-zone-2 {
+      fallthrough example.gov.
+    }
+}
+~~~
+
+Enable clouddns with multiple hosted zones with the same domain:
+
+~~~ txt
+. {
+    clouddns example.org.:gcp-example-project:example-zone example.com.:gcp-example-project:other-example-zone
+}
+~~~
--- a/plugin/clouddns/clouddns.go
+++ b/plugin/clouddns/clouddns.go
@@ -1,0 +1,222 @@
+// Package clouddns implements a plugin that returns resource records
+// from GCP Cloud DNS.
+package clouddns
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/file"
+	"github.com/coredns/coredns/plugin/pkg/fall"
+	"github.com/coredns/coredns/plugin/pkg/upstream"
+	"github.com/coredns/coredns/request"
+
+	"github.com/miekg/dns"
+	gcp "google.golang.org/api/dns/v1"
+)
+
+// CloudDNS is a plugin that returns RR from GCP Cloud DNS.
+type CloudDNS struct {
+	Next plugin.Handler
+	Fall fall.F
+
+	zoneNames []string
+	client    gcpDNS
+	upstream  *upstream.Upstream
+
+	zMu   sync.RWMutex
+	zones zones
+}
+
+type zone struct {
+	projectName string
+	zoneName    string
+	z           *file.Zone
+	dns         string
+}
+
+type zones map[string][]*zone
+
+// New reads from the keys map which uses domain names as its key and a colon separated
+// string of project name and hosted zone name lists as its values, validates
+// that each domain name/zone id pair does exist, and returns a new *CloudDNS.
+// In addition to this, upstream is passed for doing recursive queries against CNAMEs.
+// Returns error if it cannot verify any given domain name/zone id pair.
+func New(ctx context.Context, c gcpDNS, keys map[string][]string, up *upstream.Upstream) (*CloudDNS, error) {
+	zones := make(map[string][]*zone, len(keys))
+	zoneNames := make([]string, 0, len(keys))
+	for dnsName, hostedZoneDetails := range keys {
+		for _, hostedZone := range hostedZoneDetails {
+			ss := strings.SplitN(hostedZone, ":", 2)
+			if len(ss) != 2 {
+				return nil, errors.New("either project or zone name missing")
+			}
+			err := c.zoneExists(ss[0], ss[1])
+			if err != nil {
+				return nil, err
+			}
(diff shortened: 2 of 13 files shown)


plugin/clouddns/README.md
```diff
@@ -0,0 +1,68 @@
...
+      fallthrough example.gov.
+    }
+}
+~~~
+
+Enable clouddns with multiple hosted zones with the same domain:
+
+~~~ txt
+. {
+    clouddns example.org.:gcp-example-project:example-zone example.org.:Z93A52145678156
```
maguro: Is `:Z93A52145678156` valid?

palash25: No, thanks for catching this, its a copy paste error from the route53 plugin will change it.

palash25:
@yongtang can you please review?

plugin/clouddns/clouddns.go
```diff
@@ -0,0 +1,231 @@
+package clouddns
```
dilyevsky, line 3: Missing doc string.

plugin/clouddns/clouddns.go
```diff
@@ -0,0 +1,231 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/coredns/coredns/plugin"
```
dilyevsky, line 13: This block and block below should be reversed - from less specific to more.

plugin/clouddns/clouddns.go
```diff
@@ -0,0 +1,231 @@
...
+		zoneNames: zoneNames,
+		zones:     zones,
+		upstream:  up,
+	}, nil
+}
+
+// Run executes first update, spins up an update forever-loop.
+// Returns error if first update fails.
+func (h *CloudDNS) Run(ctx context.Context) error {
+	fmt.Println("Enter Run")
```
dilyevsky: Remove stdout prints here and below.

plugin/clouddns/setup.go
```diff
@@ -0,0 +1,111 @@
...
+					return c.ArgErr()
+				}
+			case "fallthrough":
+				fall.SetZonesFromArgs(c.RemainingArgs())
+			default:
+				return c.Errf("unknown property '%s'", c.Val())
+			}
+		}
+
+		ctx := context.Background()
```
dilyevsky, line 87: Probably better to obtain context from `*caddy.Controller`.

plugin/clouddns/clouddns_test.go
```diff
@@ -0,0 +1,316 @@
...
+
+	return &gcp.ResourceRecordSetsListResponse{Rrsets: rr}, nil
+}
+
+func TestCloudDNS(t *testing.T) {
+	ctx := context.Background()
+
+	r, err := New(ctx, fakeGCPClient{}, map[string][]string{"bad.": {"bad-project:bad-zone"}}, &upstream.Upstream{})
+	if err != nil {
+		t.Fatalf("Failed to create Route53: %v", err)
```
dilyevsky: These prints still reference Route53. Other examples below.

plugin/clouddns/clouddns.go
```diff
@@ -0,0 +1,231 @@
...
+				errc <- err
+			}()
+
+			for i, hostedZone := range z {
+				fmt.Printf("Err %v", err)
+				newZ := file.NewZone(zName, "")
+				newZ.Upstream = h.upstream
+
+				zoneDetails := strings.Split(hostedZone.id, ":")
+
```
dilyevsky: nit: blank not needed.

plugin/clouddns/clouddns.go
```diff
@@ -0,0 +1,231 @@
...
+				rrListResponse, err = h.client.listRRSets(zoneDetails[0], zoneDetails[1])
```
dilyevsky: Probably want to test that `len(zoneDetails) == 2`

plugin/clouddns/clouddns.go
```diff
@@ -0,0 +1,231 @@
...
+// string of project name and hosted zone name lists as its values, validates
+// that each domain name/zone id pair does exist, and returns a new *CloudDNS.
+// In addition to this, upstream is passed for doing recursive queries against CNAMEs.
+// Returns error if it cannot verify any given domain name/zone id pair.
+func New(ctx context.Context, c gcpDNS, keys map[string][]string, up *upstream.Upstream) (*CloudDNS, error) {
+	zones := make(map[string][]*zone, len(keys))
+	zoneNames := make([]string, 0, len(keys))
+	for dnsName, hostedZoneDetails := range keys {
+		for _, hostedZone := range hostedZoneDetails {
+			zoneSlice := strings.SplitN(hostedZone, ":", 2)
```
dilyevsky: `zoneSlice` is long and generic. Better use idiomatic short name like `ss`

plugin/clouddns/clouddns.go
```diff
@@ -0,0 +1,231 @@
...
+			err := c.zoneExists(zoneSlice[0], zoneSlice[1])
```
dilyevsky: Didn't test for len == 2

plugin/clouddns/clouddns.go
```diff
@@ -0,0 +1,231 @@
...
+			if err != nil {
+				return nil, err
+			}
+			fqdnDNSName := dns.Fqdn(dnsName)
+			if _, ok := zones[fqdnDNSName]; !ok {
+				zoneNames = append(zoneNames, fqdnDNSName)
+			}
+			zones[fqdnDNSName] = append(zones[fqdnDNSName], &zone{id: hostedZone, dns: fqdnDNSName, z: file.NewZone(fqdnDNSName, "")})
```
dilyevsky: It's better to store project and managedZone as fields instead of parsing them out from .id every time.

plugin/clouddns/clouddns.go
dilyevsky: Also see my comment above about storing this as struct fields instead.

dilyevsky (commented):
Took initial pass.

plugin/clouddns/clouddns.go
palash25: I am sorry but I am not really sure what you mean. The imports are already in the order of less specific to more (`plugin` -> `plugin/file` -> `plugin/file/pkg/fall` and so on) so why should they be reversed?

plugin/clouddns/setup.go
palash25: I was following the same pattern as the route53 plugin as it also uses `context.Background()`, even the current [Azure plugin PR](https://github.com/coredns/coredns/pull/2945/files) also uses the same, so I used `context.Background()` for uniformity across the plugins.

plugin/clouddns/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Name
+
+*clouddns* - enables serving zone data from GCP clouddns.
+
+## Description
+
+The clouddns plugin is useful for serving zones from resource record
+sets in GCP clouddns. This plugin supports all Google Cloud DNS records
+([Look at the Supported DNS Record types section](https://cloud.google.com/dns/docs/overview)).
```
yongtang: Could change to 
```
This plugin supports all [Google Cloud DNS records](https://cloud.google.com/dns/docs/overview#supported_dns_record_types)
```

plugin/clouddns/README.md
```diff
@@ -0,0 +1,68 @@
...
+## Syntax
+
+~~~ txt
+clouddns [DNS_NAME:PROJECT_NAME:HOSTED_ZONE_NAME...] {
+    credentials [FILE_PATH]
+    fallthrough [ZONES...]
+}
+~~~
+
+*   **DNS_NAME** the name of the domain to be accessed. When there are multiple zones with overlapping
```
yongtang: `**DNS_NAME**` => `**ZONE**`

plugin/clouddns/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+The clouddns plugin is useful for serving zones from resource record
+sets in GCP clouddns. This plugin supports all Google Cloud DNS records
+([Look at the Supported DNS Record types section](https://cloud.google.com/dns/docs/overview)).
+The clouddns plugin can be used when coredns is deployed on GCP or elsewhere.
+
+## Syntax
+
+~~~ txt
+clouddns [DNS_NAME:PROJECT_NAME:HOSTED_ZONE_NAME...] {
```
yongtang: `DNS_NAME` => `ZONE`

plugin/clouddns/README.md
```diff
@@ -0,0 +1,68 @@
...
+    credentials [FILE_PATH]
```
yongtang: `FILE_PATH` => `FILENAME` to be consistent with route53

palash25:
@yongtang I have updated the PR, please take another look :smile:

plugin/clouddns/clouddns.go
palash25: you are right about that, thanks, changes made

yongtang:
@palash25 First pass looks ok. I will take another look during the weekend. Also, can you rebase the PR?

palash25:
done :+1:

yongtang:
I played with the plugin over the weekend, and look good to me.

One interesting find while I am testing on Google Cloud Platform, is that GCP actually allows you to create a zone without attaching to a VPC. Able to dig the DNS records while seeing `This zone isn't visible to any networks. Add networks to make this zone active.` on the GCP console.

This could be very useful as essentially you could use Google Cloud DNS as a database or datastore for your onprem usage. Essentially you could create a zone on Google Cloud DNS without attaching to a VPC, then you can fully access the record on your local cluster.

It does comes with a cost, though you could combine with cache and forward plugin to cache the result. The advantage is that there is no need to managed a local database or file.

Note on AWS, Route53 does require you to specify the VPC ID when create a Hosted Zone. So it is not applicable on AWS.

Overall the plugin is a useful addition.

yongtang (approved):

PR #3013, opened 2019-07-17
pkg/cache: fix race in Add() and Evict()

charlievieth:

This fixes a race in Add() when the shard is at max capacity and the key
being added is already stored.  Previously, the shard would evict a
random value - when all it needed to do was replace an existing value.
There was a race in how Evict() picked which key to remove, which would
cause concurrent calls to Evict() to remove the same key.

Additionally, this commit removes a lot of the lock contention and a
race around Add() and Evict() by changing them to immediately hold the
write lock.  Previously, they would check conditions with the read lock
held and not re-check those conditions once the write lock was acquired
(this is a race).

As an added bonus the cache is now slightly faster:
```
benchmark                      old ns/op     new ns/op     delta
BenchmarkCache-12              59.1          48.6          -17.77%
BenchmarkShard-12              64.5          59.6          -7.60%
BenchmarkShard_Parallel-12     119           111           -6.72%
```

N/A

N/A

N/A

### DIFF
--- a/plugin/pkg/cache/cache_test.go
+++ b/plugin/pkg/cache/cache_test.go
@@ -42,9 +42,8 @@ func TestCacheLen(t *testing.T) {
 }
 
 func TestCacheSharding(t *testing.T) {
-	const N = shardSize
-	c := New(N)
-	for i := 0; i < N*2; i++ {
+	c := New(shardSize)
+	for i := 0; i < shardSize*2; i++ {
 		c.Add(uint64(i), 1)
 	}
 	for i, s := range c.shards {
--- a/plugin/pkg/cache/cache_test.go
+++ b/plugin/pkg/cache/cache_test.go
@@ -3,11 +3,22 @@ package cache
 import "testing"
 
 func TestCacheAddAndGet(t *testing.T) {
-	c := New(4)
+	const N = shardSize * 4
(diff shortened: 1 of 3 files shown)


plugin/pkg/cache/shard_test.go
```diff
@@ -57,4 +83,57 @@ func TestShardLenEvict(t *testing.T) {
...
+	s := newShard(shardSize)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		k := uint64(i) % shardSize * 2
+		s.Add(k, 1)
+		s.Get(k)
+	}
+}
+
+func BenchmarkShard_Parallel(b *testing.B) {
```
miekg: this can drop the underscore

plugin/pkg/cache/cache_test.go
```diff
@@ -30,6 +41,19 @@ func TestCacheLen(t *testing.T) {
 	}
 }
 
+func TestCacheSharding(t *testing.T) {
+	const N = shardSize
```
miekg: nit: this const doesn't really add anything, and can be better removed

miekg (commented):
looks good. Thank you!

some nits

miekg (approved):

PR #3017, opened 2019-07-17
plugin/file: fix setting ReloadInterval

miekg:
The reload interval was only correctly set if there was an extra
block for the file. Move this down to set up.

Add test case that fails before, but now works.


stp-ip (approved):
/lgtm

PR #3025, opened 2019-07-20
plugin/file: New zone should have zero records

miekg:
After calling NewZone the number of records should be zero, but due to
how zone.All() was implemented so empty RRs would be added. This then
fails the == 0 check in xfr.go and put nil in the slice, this then
subsequently panics on the Len().

Fix this making All() smarter when adding records. Added little test to
enfore this.


yongtang (approved):
lgtm

miekg:
thanks @yongtang !

PR #3029, opened 2019-07-22
Fix response_rcode_count_total metric

mlowery:

Metric name in forward and grpc READMEs is wrong.

None.

I saw that man pages also have this wrong but not sure how they are built.

No.


miekg:
ah. Thank you!

PR #3038, opened 2019-07-23
Add Stringer to file.Success

maguro:
This was handy when I was debugging some stuff.

Useful for debugging purposes.

none

none

none


miekg:
I don't see an immediate need for this...

Also don't use testify. Now we need to learn some kind of test DSL

PR #3039, opened 2019-07-23
plugins: skip initializing if the origins don't match

### LINKED ISSUE #3026: multiple dns names and use of origin in server block
While write the *sign* plugin it ran into the following. With this Corefile something unexpected is happening:
~~~
miek.nl miek.org {
    file plugin/sign/testdata/db.miek.nl.signed miek.nl {
        reload 5s
    }
    file plugin/sign/testdata/db.miek.org.signed miek.org {
        reload 5s
    }

    sign plugin/sign/testdata/db.miek.nl {
        key file plugin/sign/testdata/Kmiek.nl.+013+59725
        directory plugin/sign/testdata
    }
}
~~~
If you start CoreDNS you'll get:

~~~
plugin/file: Successfully reloaded zone "miek.nl." in "plugin/sign/testdata/db.miek.nl.signed" with serial 1563737941
plugin/file: Successfully reloaded zone "miek.org." in "plugin/sign/testdata/db.miek.org.signed" with serial 1563737941
plugin/file: Successfully reloaded zone "miek.nl." in "plugin/sign/testdata/db.miek.nl.signed" with serial 1563737941
plugin/file: Successfully reloaded zone "miek.org." in "plugin/sign/testdata/db.miek.org.signed" with serial 1563737941
~~~
4 times, because the zone stanza is parsed twice, once for miek.nl and once for miek.org. This doesn't do what you expect.

This can be checked for what stanza you're actually parsing, this information *is* available in the `caddy.Controller`, but it might be worth adding a simple function for this to make sure you're setting up for the zone you're called for instead of all of them.

miekg:
ServerBlocks are parsed for each name in the serverblock keys, this may
lead to block being setup multiple times while the intent was probably
to do things once:

~~~
miek.nl miek.org {
    file plugin/sign/testdata/db.miek.nl.signed miek.nl {
        reload 5s
    }
    file plugin/sign/testdata/db.miek.org.signed miek.org {
        reload 5s
    }
}
~~~

Each file plugin should do a thing for the origin specified after it,
but this the first one is done for miek.nl and miek.org and the second
one too; so 4 in total. I don't believe this is the intent.

This add Origins function that checks for this. This returns an
ErrOrigin for which the plugin needs to check.

This PR makes *all* tests fail because a caddy.NewTestController doesn't
set up the full environment that Origins expects. Not sure how to handle
this case.

Fixes: #3026

Note all plugins that do this need this fix.

### DIFF
--- a/plugin.md
+++ b/plugin.md
@@ -12,7 +12,7 @@ The main method that gets called is `ServeDNS`. It has three parameters:
 will return a SERVFAIL to the client. The response code tells CoreDNS if a *reply has been
 written by the plugin chain or not*. In the latter case CoreDNS will take care of that.
 
-CoreDNS treats:
+CoreDNS treats (see plugin.ClientWrite):
 
 * SERVFAIL (dns.RcodeServerFailure)
 * REFUSED (dns.RcodeRefused)
--- a/plugin/file/setup.go
+++ b/plugin/file/setup.go
@@ -23,6 +23,9 @@ func init() {
 func setup(c *caddy.Controller) error {
 	zones, err := fileParse(c)
 	if err != nil {
+		if err == plugin.ErrOrigin {
+			return nil
+		}
 		return plugin.Error("file", err)
 	}
 
@@ -71,6 +74,9 @@ func fileParse(c *caddy.Controller) (Zones, error) {
 		copy(origins, c.ServerBlockKeys)
 		args := c.RemainingArgs()
 		if len(args) > 0 {
+			if err := plugin.Origins(c, args); err != nil {
+				return Zones{}, err
+			}
 			origins = args
 		}
 
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/caddyserver/caddy"
 	"github.com/miekg/dns"
 	ot "github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
@@ -99,11 +100,33 @@ func ClientWrite(rcode int) bool {
 	return true
 }
 
+// Origins returns nil if the c.ServerBlockKey[c.ServerBlockKeyIndex] is
+// contained as a subdomain (or equal) in the origins slice. If it can't be
+// found ErrOrigin will be returned. The ServerBlockKey names are Normalize-d
+// before comparing them to the fully qualifed names in origins
+func Origins(c *caddy.Controller, origins []string) error {
+	if len(c.ServerBlockKeys) == 0 {
+		return nil // needed because caddy.NewTestController doesn't set this, making all tests panic
+	}
+	key := Host(c.ServerBlockKeys[c.ServerBlockKeyIndex]).Normalize()
+	for i := range origins {
+		if dns.IsSubDomain(key, dns.Fqdn(origins[i])) {
+			return nil
+		}
+	}
+	return ErrOrigin
+}
+
 // Namespace is the namespace used for the metrics.
 const Namespace = "coredns"
 
 // TimeBuckets is based on Prometheus client_golang prometheus.DefBuckets
 var TimeBuckets = prometheus.ExponentialBuckets(0.00025, 2, 16) // from 0.25ms to 8 seconds
 
-// ErrOnce is returned when a plugin doesn't support multiple setups per server.
-var ErrOnce = errors.New("this plugin can only be used once per Server Block")
+var (
+	// ErrOnce is returned when a plugin doesn't support multiple setups per server.
+	ErrOnce = errors.New("this plugin can only be used once per Server Block")
+
+	// ErrOrigin is returned when a plugin is being setup, but specifies a more narrow set of origins.
+	ErrOrigin = errors.New("this plugin skipped the setup because the origin selection is more narrow")
+)
--- a/plugin/plugin_test.go
+++ b/plugin/plugin_test.go
@@ -1,0 +1,25 @@
+package plugin
+
+import (
+	"testing"
+
+	"github.com/caddyserver/caddy"
+)
+
+func TestOrigins(t *testing.T) {
+	c := caddy.NewTestController("dns", "")
+	c.ServerBlockKeys = []string{"example.net", "example.org"}
+
+	c.ServerBlockKeyIndex = 0
+	if err := Origins(c, []string{"example.net"}); err != nil {
+		t.Errorf("Expected nil error, got %s", err)
+	}
+	c.ServerBlockKeyIndex = 0
+	if err := Origins(c, []string{"sub.example.net"}); err != nil {
+		t.Errorf("Expected nil error, got %s", err)
+	}
+	c.ServerBlockKeyIndex = 1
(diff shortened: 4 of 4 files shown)


miekg:
the comparison should probably be "dns.IsSubDomain"

miekg:
the error check should move to the server, so the plugin can be kept simpler. Then in the server we can also emit a warning with some context: plugin not initialized for zone %s or something.
Plugin can keep doing `if err != nil; return err`

miekg:
skip that, that becomes really messy because the caddy code handles those initials error, we can catch them when we make the servers, but that seems quite late in the process. The current PR is more explicit and maybe that's for the better.

chrisohaver:
The following I think would still initialize the file plugin twice (for east.example.com and west.example.com)...

```
east.example.com west.example.com example.org {
   file dbfile example.com
}
```

chrisohaver:
> The following I think would still initialize the file plugin twice ...

Actually - I think I got the sub domain comparison backwards.  `example.com` is not a subdomain of any of the serverblock zones, so I think it would not get set up at all.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugins: skip..." ]
>The following I think would still initialize the file plugin twice (for east.example.com and west.example.com)...
>
>```
>east.example.com west.example.com example.org {
>   file dbfile example.com
>}
>```

true and good point

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugins: skip..." ]
>The following I think would still initialize the file plugin twice (for east.example.com and west.example.com)...
>
>```
>east.example.com west.example.com example.org {
>   file dbfile example.com
>}
>```

can we chuck this in "don't do that" ?

--
Miek Gieben

chrisohaver:
It becomes more of an issue when handling other plugins such as *forward*, which commonly uses "." as the zone... e.g.

```
zone1.com 168.192.inaddr.arpa {
   forward . 192.168.1.10
}
```

chrisohaver:
Nevermind, I just noticed that you have not listed forward as an affected plugin in #3026, which means I'm not understanding the underlying issue.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugins: skip..." ]
>It becomes more of an issue when handling other plugins such as *forward*, which commonly uses "." as the zone... e.g.
>
>```
>zone1.com 168.192.inaddr.arpa {
>   forward . 192.168.1.10
>}
>```

but that "." is not defined as a zone and this check is not applied. I do admit 
it becomes a bit fuzzy; but what currently happens is certainly not good.

johnbelamaric:
I am not sure this is the right fix. I think from a user point of view:

```
foo.com bar.com {
  some plugins
}
```

should result in a single plugin chain, and foo.com and bar.com get routed to the same plugin chain. Today it sounds like it results in two plugin chains?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugins: skip..." ]
>I am not sure this is the right fix. I think from a user point of view:
>
>```  
>foo.com bar.com {
>  some plugins
>}
>```
>
>should result in a single plugin chain, and foo.com and bar.com get routed to the same plugin chain. Today it sounds like it results in two plugin chains?

2 servers with a plugin chain each

johnbelamaric:
No, I think it's a single server with two `Config` objects in the [`zones` map](https://github.com/coredns/coredns/blob/master/core/dnsserver/server.go#L36). But yeah, looking at `Config`, it is for a [single zone](https://github.com/coredns/coredns/blob/master/core/dnsserver/config.go#L14), so we end up with multiple identical plugin chains - one for each zone. Which really sucks. Why can't a `Config` apply to more than one zone?

johnbelamaric:
Looking at `Config` more closely, the only pieces that could not be shared across all the zones listed in the server block keys are: Port, Transport, TLSConfig. In fact, if any of those are different, we create different *servers*. But I think that if those are all the same, we share the servers across Configs. In other words, if we bumped those pieces up to the server level (they are redundant at Config level anyway), then we can probably achieve this. It's a big refactor though.

johnbelamaric:
Unless my reading of this is wrong. It's been a while since I was digging in this code.

miekg:
I think what you propose will make things simpler. But you still need to skip a plugin's initialization like this in this PR?

johnbelamaric:
Maybe. But the reality is what I propose is going to take some time. It probably makes sense to make this fix now, and work on the bigger fix for later.

miekg:
I see the logic in your reasoning about `Config` and that's definitely something to be fixed. However I fail to understand how this is related to this PR/issue?

johnbelamaric:
The goal I was trying to achieve is to have the server block stanzas map to the internal structures, so that it is more intuitive for users. That is, a single server block means a single plugin chain, not potential duplicate plugin chains.

If we move those up to the server, then we can re-use a Config across all those zones in a server, meaning a single plugin chain in these cases. It's not really good enough to achieve the goal, though; the Corefile semantics would *still* be sort of weird. You would get a plugin chain for each unique protocol and port listed in the server block keys. Better, but not quiet there.

miekg:
Config can't be part of the server, or, at least the plugin.Chain can not.
~~~
foo.com bar.com {
    whoami
}

bloop.com {
    log
    whoami
}
~~~
creates 1 server, where only foo.com and bar.com share the plugin chain, bloop.com has completely different chain

miekg:
To skip init of the plugin you have to be early, e.g. for the *whoami* plugin:
~~~
	dnsserver.GetConfig(c).AddPlugin(func(next plugin.Handler) plugin.Handler {
		return Whoami{}
	})
~~~
the init can be skipped if `GetConfig(c)` return something we've seen already. But this clashes with plugins that can be allowed multiple times in a server block.

johnbelamaric:
My goal is not "one config per server" or "one plugin chain per server". It is "one plugin chain per server block in the Corefile". So, as it stands today, your Corefile would result in 3 plugin chains, one for each of foo.com, bar.com, and bloop.com. The two for foo.com and bar.com would be identical but distinct instances. When you add, say, cache in there, then this is bad (two separate caches). Even worse if kubernetes is in there, because you get multiple copies of the object caches.

miekg:
Not disagreeing here. Just echo-ing your sentiment that this is quite a big
fix. Also not sure why this code is so complicated, seems to be a relative
straight forward thing to do in the end.

mholt:
FYI:

When the Caddyfile is parsed, a server block is replicated once for each key at the heading of the block. So directives are executed "once per site" (or whatever the Corefile calls them) instead of "once per appearance".

However, there are ways to do things "once per appearance" regardless of the number of keys.

See the Controller type: https://sourcegraph.com/github.com/caddyserver/caddy@0ba427a6f44edb5169e022e3cbc23f7b6de35a81/-/blob/controller.go#L23-64 - there are fields that allow you to execute functions once per server block, as well as getting the positioning of the server block or the key in the block. That's how Caddy overcomes this problem when things need to happen "once per directive appearance" as opposed to "once per site".

miekg:
I'm inclined to close this PR and fix the docs in *sign*, as close it with WAI, nothing needs to be changed here, you just need to be aware of these things.

PR #3047, opened 2019-07-25
plugin/kubernetes: update to client-go 12.0.0

chrisohaver:

Updates client-go to 12.0.0 (released with k8s 1.15).

none

none

no


go.mod
```diff
@@ -59,13 +58,12 @@ require (
 	google.golang.org/grpc v1.22.0
 	gopkg.in/DataDog/dd-trace-go.v1 v1.16.0
 	gopkg.in/inf.v0 v0.9.1 // indirect
-	k8s.io/api v0.0.0-20190313235455-40a48860b5ab
-	k8s.io/apimachinery v0.0.0-20190313205120-d7deff9243b1
-	k8s.io/client-go v11.0.0+incompatible
+	k8s.io/api v0.0.0-20190620084959-7cf5895f2711
+	k8s.io/apimachinery v0.0.0-20190612205821-1799e75a0719
+	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
```
rajansandeep, line 63: Is this taking the dependency from `latest/master` instead of `12.0.0`?

chrisohaver:
Not sure why go.mod isn't tagging client-go on the tag that I specify in go get (v12.0.0)... but the commit in the go.mod _is_ the same as the v12.0.0 tag... so all is good I guess.

yongtang (approved):
lgtm

go.mod
chrisohaver: No.  The latest commit in client-go is "f16ff99...", the tag for v12.0.0 is "78d2af7..."

rajansandeep: Yeah, seems to be normal as described in https://github.com/kubernetes/client-go/issues/631

PR #3048, opened 2019-07-25
Update CONTRIBUTING.md for dependency update methods (go dep -> go mod)

yongtang:
We changed from go dep => go mod so updating the CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,7 @@ $ go get -u
 ```
 
 After the dependencies have been updated or added, you might run the following to
-prune vendored packages:
+cleanup the go module files:
 ```sh
 $ go mod tidy
 ```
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -44,25 +44,20 @@ faster.
...
-```
-
-Use the following to update the locked versions of all dependencies
-```sh
-$ make dep-ensure
+$ go get -u
 ```
 
 After the dependencies have been updated or added, you might run the following to
 prune vendored packages:
```
stp-ip: To be precise, we don't vendor any packages anymore, just cleanup the go module files.

stp-ip (approved):
/LGMT

small note on the text as we don't vendor anymore.

PR #3051, opened 2019-07-27
Fix corefile usage

stp-ip:

### DIFF
--- a/plugin/file/README.md
+++ b/plugin/file/README.md
@@ -44,7 +44,7 @@ file DBFILE [ZONES... ] {
 Load the `example.org` zone from `example.org.signed` and allow transfers to the internet, but send
 notifies to 10.240.1.1
 
-~~~ txt
+~~~ corefile
 example.org {
     file example.org.signed {
         transfer to *
@@ -55,7 +55,7 @@ example.org {
 
 Or use a single zone file for multiple zones:
 
-~~~ txt
+~~~ corefile
 . {
     file example.org.signed example.org example.net {
         transfer to *
@@ -67,7 +67,7 @@ Or use a single zone file for multiple zones:
 Note that if you have a configuration like the following you may run into a problem of the origin
 not being correctly recognized:
 
-~~~ txt
+~~~ corefile
 . {
     file db.example.org
 }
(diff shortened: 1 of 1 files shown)


miekg:
this will *now* work because we lazily start. It was failing the TestReadme test because the zone files don't exist.

Wish we could do the same for kubernetes config tests :/

miekg:
think you need
~~~
diff --git a/test/example_test.go b/test/example_test.go
index 852143fe..9136db30 100644
--- a/test/example_test.go
+++ b/test/example_test.go
@@ -2,14 +2,15 @@ package test
 
 const exampleOrg = `; example.org test file
 $TTL 3600
-example.org.           IN      SOA     sns.dns.icann.org. noc.dns.icann.org. 2015082541 7200 3600 1209600 3600
-example.org.           IN      NS      b.iana-servers.net.
-example.org.           IN      NS      a.iana-servers.net.
-example.org.           IN      A       127.0.0.1
-example.org.           IN      A       127.0.0.2
-short.example.org.     1       IN      A       127.0.0.3
-*.w.example.org.        IN      TXT     "Wildcard"
-a.b.c.w.example.org.    IN      TXT     "Not a wildcard"
-cname.example.org.      IN      CNAME   www.example.net.
-service.example.org.    IN      SRV     8080 10 10 example.org.
+@              IN      SOA     sns.dns.icann.org. noc.dns.icann.org. 2015082541 7200 3600 1209600 3600
+@              IN      NS      b.iana-servers.net.
+@              IN      NS      a.iana-servers.net.
+@              IN      A       127.0.0.1
+@              IN      A       127.0.0.2
+short  1       IN      A       127.0.0.3
+
+*.w        3600 IN      TXT     "Wildcard"
+a.b.c.w    IN      TXT     "Not a wildcard"
+cname      IN      CNAME   www
+service    IN      SRV     8080 10 10 @
 `
~~~
this works on master

stp-ip:
Not working for me yet, but will take another look.

miekg:
the importance of README testing. This is uncovering bugs :/

miekg:
see #3052 for the fix

PR #3052, opened 2019-07-27
plugin/file: allow README.md testing

miekg:
Allow readme testing for the file plugin and fix bugs that where found:

* the reader wasn't reset when re-reading the same io.reader for a
  different origin.

Builds upon #3051

### DIFF
--- a/test/example_test.go
+++ b/test/example_test.go
@@ -11,6 +11,6 @@ short	1	IN	A	127.0.0.3
 
 *.w        3600 IN      TXT     "Wildcard"
 a.b.c.w    IN      TXT     "Not a wildcard"
-cname      IN      CNAME   www
+cname      IN      CNAME   www.example.net.
 service    IN      SRV     8080 10 10 @
 `
--- a/test/example_test.go
+++ b/test/example_test.go
@@ -2,14 +2,15 @@ package test
 
 const exampleOrg = `; example.org test file
 $TTL 3600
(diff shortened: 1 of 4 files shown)


test/example_test.go
```diff
@@ -2,14 +2,15 @@ package test
...
+@		IN	SOA	sns.dns.icann.org. noc.dns.icann.org. 2015082541 7200 3600 1209600 3600
+@		IN	NS	b.iana-servers.net.
+@		IN	NS	a.iana-servers.net.
+@		IN	A	127.0.0.1
+@		IN	A	127.0.0.2
+short	1	IN	A	127.0.0.3
+
+*.w        3600 IN      TXT     "Wildcard"
+a.b.c.w    IN      TXT     "Not a wildcard"
+cname      IN      CNAME   www
```
stp-ip:
```suggestion
cname      IN      CNAME   www.example.net.
```

stp-ip: This should fix the non flake failing test.

stp-ip (changes requested):
Besides the CNAME fix and the separately filed flake, this looks good.

stp-ip:
restarting test due to external zip file failure

miekg:
well, so be it. Merging

stp-ip:
Cleared all travis caches. Test build on master should probably not fail. We'll see.

PR #3066, opened 2019-07-29
fix ineffectual assignment to err

zgfh:
fix ineffectual assignment to err
no
no
no

### DIFF
--- a/owners_generate.go
+++ b/owners_generate.go
@@ -40,7 +40,10 @@ func main() {
 			o, err = owners(p, o)
 			return err
 		})
-
+	if err != nil {
+		log.Fatal(err)
+	}
+	
 	// sort it and format it
 	list := []string{}
 	for k := range o {


plugin/kubernetes/controller_test.go
```diff
@@ -79,6 +79,9 @@ func generateEndpoints(cidr string, client kubernetes.Interface) {
 		}
 		ep.ObjectMeta.Name = "svc" + strconv.Itoa(count)
 		_, err = client.CoreV1().Endpoints("testns").Create(ep)
+		if err != nil {
```
miekg: this should return the err; but this being test code I don't think we care **that** much

miekg: idem for the other test functions

owners_generate.go
```diff
@@ -26,7 +26,7 @@ func main() {
 	}
 
 	// each of the plugins, in case someone is not in the top-level one
-	err = filepath.Walk("plugin",
+	_ = filepath.Walk("plugin",
```
miekg: keep the `err` and `log.Fatal(err)` afterwards

zgfh: updated

miekg:
the change in `owners_generate.go` is the only one I want to keep; there rest can be reverted.

zgfh:
@miekg updated

PR #3068, opened 2019-07-30
plugin/reload: Graceful reload of imported files

erikwilson:
Graceful reload of imports. 
If an imported file changes the reload plugin should detect the change.
https://github.com/coredns/coredns/tree/master/plugin/import

https://github.com/coredns/coredns/issues/2633
https://github.com/rancher/k3s/issues/462
https://github.com/Azure/AKS/issues/841

Import support on reload:
https://github.com/coredns/coredns/blob/master/plugin/reload/README.md#bugs

Unknown

### DIFF
--- a/plugin/reload/README.md
+++ b/plugin/reload/README.md
@@ -89,8 +89,9 @@ closed in step 1; so the health endpoint is broken. The same can hopen in the pr
 
 In general be careful with assigning new port and expecting reload to work fully.
 
-Also any `import` statement is not discovered by this plugin. This means if any of these imported files
-changes the *reload* plugin is ignorant of that fact.
+In CoreDNS v1.6.0 and earlier any `import` statements are not discovered by this plugin.
+This means if any of these imported files changes the *reload* plugin is ignorant of that fact.
+CoreDNS v1.7.0 and later does parse the Corefile and supports detecting changes in imported files.
 
 ## Metrics
 
--- a/plugin/reload/reload.go
+++ b/plugin/reload/reload.go
@@ -1,11 +1,14 @@
 package reload
 
 import (
+	"bytes"
 	"crypto/md5"
+	"encoding/json"
 	"sync"
 	"time"
 
 	"github.com/caddyserver/caddy"
+	"github.com/caddyserver/caddy/caddyfile"
 )
 
 // reload periodically checks if the Corefile has changed, and reloads if so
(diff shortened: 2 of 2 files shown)


plugin/reload/reload.go
```diff
@@ -46,6 +49,14 @@ func (r *reload) interval() time.Duration {
 	return r.dur
 }
 
+func parse(corefile caddy.Input) ([]byte, error) {
+	serverBlocks, err := caddyfile.Parse(corefile.Path(), bytes.NewReader(corefile.Body()), nil)
+	if err != nil {
+		return nil, err
+	}
+	return json.Marshal(serverBlocks)
```
johnbelamaric, line 57: Are we sure this is deterministic? That is, are we sure it always results in the same ordering? Otherwise we could end up with continuously reloading. We may need the MD5 to navigate through any map structure, sorting the map keys and creating a canonical string representation of the whole file.

johnbelamaric (changes requested):
I don't think we can rely on json.Marshal always producing identical output for the same structs. But I could be wrong.

Please update the readme to include reflect the change as well.

plugin/reload/reload.go
erikwilson: From my understanding `json.Marshal` is deterministic, although the documentation does a poor job at describing this. There is a part of the Marshal docs that states `map keys are sorted and used as JSON object keys` (https://github.com/golang/go/blob/5fae09b7386de26db59a1184f62fc7b22ec7667b/src/encoding/json/encode.go#L135). Looking at the code it appears map keys and struct fields are sorted going back to at least Go v1.4.

Information added to README about the behavior with version context.

plugin/reload/README.md
```diff
@@ -89,8 +89,9 @@ closed in step 1; so the health endpoint is broken. The same can hopen in the pr
 
 In general be careful with assigning new port and expecting reload to work fully.
 
-Also any `import` statement is not discovered by this plugin. This means if any of these imported files
-changes the *reload* plugin is ignorant of that fact.
+In CoreDNS v1.6.x and earlier any `import` statements are not discovered by this plugin.
```
johnbelamaric: 1.6.0

johnbelamaric (commented):
Ok, one minor tweak and I am good with it.

plugin/reload/README.md
erikwilson: Updated, thanks!

PR #3072, opened 2019-07-31
core: log panics

miekg:
These are too hidden now. They increase the issue-load, because people
don't see them.

Add log.Errorf in the core/dnsserver recover routine.


dilyevsky:
```
1428--- FAIL: TestSecondaryZoneTransfer (0.01s)
1429    secondary_test.go:83: Expected answer section
```

^ appears to be flaky

dilyevsky (approved):

miekg:
thanks (and yes that's flaky and should be fixed)

Issue #2353, opened 2018-11-30
plugin/route53: Throttling: Rate exceeded


mattlqx:
I have about 10 coredns daemons running with the route53 plugin setup for about 6 zones. All the daemons share the same instance profile credentials to read from Route 53. I'm seeing a good amount of `Throttling: Rate exceeded` messages in the log. Would it be difficult to make the refresh rate of the zones configurable? I don't necessarily need sub-10 second convergence of a zone and I'd like to not hammer the API and reach AWS's rate limits.

The throttling sometimes results in a daemon failing to start when it gets a 400 returned back from the API. I'd like to be able to control the rate the running daemons are refreshing zones to prevent these kinds of delayed starts as well.

I'm not entirely sure at what level AWS is rate limiting by account, by credential or something else, but I'd really not like to generate a whole new credential/user per coredns instance.

dilyevsky:
Currently the refresh is set to `1 * time.Minute` https://github.com/coredns/coredns/blob/95c9e14dcf65a9c3b97aa1be611fae664f99f7e6/plugin/route53/route53.go#L87 but it refreshes all zones at once. Shouldn't be difficult to have it configured but I'm thinking it probably has more to do with either the number of simultaneous zones or total number of records (they are paginated at 100 by default). Could you provide the total number of records you have configured between all of those zones?

dilyevsky:
Oh also just in case - your are using `>v1.2.3` right?

dilyevsky:
After reading through https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html#limits-api-requests this is likely:

```
All requests
Five requests per second per AWS account. If you submit more than five requests per
second, Amazon Route 53 returns an HTTP 400 error (Bad request). The response header
also includes a Code element with a value of Throttling and a Message element with a
value of Rate exceeded.
```

mattlqx:
Yes, v1.2.6.

I do have a fair amount of records...

Zone A  1600
Zone B  800
Zone C  100
Zone D  150

The remaining ones only have a handful.

mattlqx:
Thanks for the info on the rate limiting. Perhaps a configurable for pagination might help also or just make it something more like 1000 or 5000? I'm not sure there's a disadvantage to doing so.

dilyevsky:
1000 records is max that can be returned in a single call and with multiple
zones that are small it’s possible to hit this as well so I’m going to add
a ratelimiter.

[...]

PR #3073, opened 2019-07-31
[plugin/route53]: Increase ListResourceRecordSets paging size.

mattlqx:

without the paging parameter set, it will default to 100 records per request. with large enough zones and potentially multiple coredns daemons configured to pull from route 53, this can quickly add up and reach aws global api rate limits (5/sec per root account). increasing paging to max can help reduce the number of requests needed to pull records for a zone without no down side that i am aware of.

this helps issue #2353, but probably is not a complete fix.

None.

No.


dilyevsky (approved):
Thanks! Not sure what's up with those presubmits...

dilyevsky (changes requested):
Oh you need to rebase to pick up circle config.

PR #3054, opened 2019-07-27
plugin/kubernetes: remove some of the klog setup

miekg:
I don't believe this is actually needed (anymore). The:

os.Stderr = os.Stdout

is a crazy hack that def. needs to go.

PR #3076, opened 2019-07-31
plugin/kubernetes: remove some of the klog setup (#3054)

miekg:
This reverts commit 4fda9535d2f446de9dbaf8b397b0d871bdc882e5.

This uncovers another race that we didn't have before; color me
confused.


miekg:
holding off to see what #3079 does

PR #3079, opened 2019-08-01
Fixes races in test and klog

miekg:
Various fixes to make things less flaky:

* kubernetes: put klog.SetOutput in the setup function, not in the init
  function to see if that helps
* file: make z.Expired a boolean instead of a pointer to a boolean
* test: fix TestSecondaryZoneTransfer test, which wasn't actually
  testing in the right way. It's more right now, but may still be racy
  (race introduced because a file's lazy loading of zones)


yongtang (approved):
lgtm

PR #3083, opened 2019-08-01
plugin/route53: make refresh frequency adjustable

mattlqx:

the current update frequency for the refresh loop in the route 53 plugin is hard-coded
to 1 minute. aws rate-limits the number of api requests so less frequent record refreshes
can help when reaching those limits depending upon your individual scenarios. this pull
adds a configuration option to the route53 plugin to adjust the refresh frequency.

thanks for getting my last pull released so quickly. this is the last local change that
i have been running and would love to get it contributed back to the project.

Another step to solving #2353 

Included in pull.

No, maintains current behavior by default.

### DIFF
--- a/plugin/route53/README.md
+++ b/plugin/route53/README.md
@@ -18,6 +18,7 @@ route53 [ZONE:HOSTED_ZONE_ID...] {
     aws_access_key [AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY]
     credentials PROFILE [FILENAME]
     fallthrough [ZONES...]
+    refresh DURATION
 }
 ~~~
 
@@ -47,6 +48,14 @@ route53 [ZONE:HOSTED_ZONE_ID...] {
 
 *   **ZONES** zones it should be authoritative for. If empty, the zones from the configuration
     block.
+
+*   `refresh` can be used to control how long between record retrievals from Route 53. It requires
+    a duration string as a parameter to specify the duration between update cycles. Each update
+    cycle may result in many AWS API calls depending on how many domains use this plugin and how
+    many records are in each. Adjusting the update frequency may help reduce the potential of API
+    rate-limiting imposed by AWS.
+
+*   **DURATION** A duration string. Defaults to `1m`. If units are unspecified, seconds are assumed.
 
 ## Examples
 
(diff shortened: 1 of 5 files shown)


miekg:
there seems to be many issue to the frequency that allows you to fetch stuff from this API. A more user friendly approach is to figure out programmatically what the best freq. is and use that.

mattlqx:
The Terraform approach is to make as many requests as you can and then use an exponential back-off until you've made all the requests you want. The problem there is that you don't know if you're going to be limited until you are and once you are, you can only wait and keep trying. That works fine from the standpoint of a single client, but the rate-limits are *global* to your whole AWS account. So too many clients without knowledge of each other trying to pull as many records as fast as possible can quickly degrade your AWS API service quality.

As an operator, I have the view of how many services I have running that need to interface with it and I need to be able to control them to prevent rate-limits from being reached in the first place to the best of my ability. This isn't a huge deal when I'm running one CoreDNS daemon configured to pull from Route 53, but it increasingly becomes an issue for every additional replica added that is configured this way.

plugin/route53/README.md
```diff
@@ -18,6 +18,7 @@ route53 [ZONE:HOSTED_ZONE_ID...] {
     aws_access_key [AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY]
     credentials PROFILE [FILENAME]
     fallthrough [ZONES...]
+    frequency SECONDS
```
dilyevsky: Maybe call this `refresh`? `frequency` seems too generic...

dilyevsky: Why not use duration string here instead? Example:
```
  refresh 1s
```

Then you can use `time.ParseDuration` in the setup and skip all this strconv boilerplate.

plugin/route53/setup.go
```diff
@@ -98,6 +101,19 @@ func setup(c *caddy.Controller, f func(*credentials.Credentials) route53iface.Ro
...
+			case "frequency":
+				args := c.RemainingArgs()
+				if len(args) == 0 {
+					return c.ArgErr()
+				}
+				t, err := strconv.Atoi(args[0])
+				if err != nil {
+					return err
+				}
+				if t < 1 || t > 3600 {
```
dilyevsky: Why enforce arbitrary range? What if I want 100ms updates? Also see my comment above.

dilyevsky (changes requested):

plugin/route53/README.md
mattlqx: Updated and I think I got it as simple as I could while still considering numbers without units as seconds to keep it reasonable.

dilyevsky (approved):
Nice!

PR #3093, opened 2019-08-07
Add Continuous Fuzzing Integration to Fuzzit

yevgenypats:
This feature introduce continuous fuzzing with the following
features:

Do not merge yet as this is work in progress but comments are welcome.

* Ruzzing: fuzz-targets are run continuously on master
( the fuzzers are updated every time new code is pushed to master)
* Regresion: In addition to unit-tests travis runs all fuzz
targets through the generated corpus to catch bugs early  on
in the CI process before merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,26 @@ ifeq ($(TEST_TYPE),coverage)
 		fi; \
 	done
 endif
+ifeq ($(TEST_TYPE),fuzzit)
+	# skip fuzzing for PR
+	if [ "$(TRAVIS_PULL_REQUEST)" = "false" ] || [ "$(FUZZIT_TYPE)" = "local-regression" ] ; then \
+        export GO111MODULE=off; \
+        go get -u github.com/dvyukov/go-fuzz/go-fuzz-build; \
+        go get -u -v .; \
+        cd ../../go-acme/lego; \
+        git checkout v2.5.0; \
+        cd ../../coredns/coredns; \
+        LIBFUZZER=YES make -f Makefile.fuzz cache chaos file rewrite whoami corefile; \
+        wget -O fuzzit https://github.com/fuzzitdev/fuzzit/releases/download/v2.4.27/fuzzit_Linux_x86_64; \
+        chmod a+x fuzzit; \
+        ./fuzzit create job --type $(FUZZIT_TYPE) coredns/cache ./cache; \
+        ./fuzzit create job --type $(FUZZIT_TYPE) coredns/chaos ./chaos; \
+        ./fuzzit create job --type $(FUZZIT_TYPE) coredns/file ./file; \
+        ./fuzzit create job --type $(FUZZIT_TYPE) coredns/rewrite ./rewrite; \
+        ./fuzzit create job --type $(FUZZIT_TYPE) coredns/whoami ./whoami; \
+        ./fuzzit create job --type $(FUZZIT_TYPE) coredns/corefile ./corefile; \
+    fi;
+endif
 
 core/plugin/zplugin.go core/dnsserver/zdirectives.go: plugin.cfg
 	GO111MODULE=on go generate coredns.go
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 dist: xenial
-
+services:
+  - docker
 language: go
 go:
   - "1.12.x"
@@ -18,10 +19,16 @@ branches:
   - master
 
 env:
-  - TEST_TYPE=coverage
-  - TEST_TYPE=integration
-  - TEST_TYPE=core
-  - TEST_TYPE=plugin
+  global:
+    # This is FUZZIT_API_KEY
+    - secure: "IGpZAyt1e5BZ1C4LeJG+GrgFZzaBQkJ3BX/+MwWN85aJSDk5gwThS53OCr/7RFgBKBgP9xBv9i9hAv0PxVaRE0ETIzjc0rQzceJIWiYKfFYQyscFahKfSiGsWP32rMlU3K67tA7yITS+Z8mMyVH9Ndr1Fg9AmLL+WfATdrd6dP8hzsUpaghKlnJee9TycrfamDpISzecdOY9xzxcwRyphZxuCc/n236Nt7f7Ccz0zx/Qa5igX6mjKZpUyBpS2u02GmNJTfc3W5SbTRP5bSJ+ozSkZZyG3tTpYmeN87AQJ/oG7rUEzqGLt78i7jSYAXghJZT06H/fHFsOKssCj1m0hYiarnGoGzXScLDqp2fpkyzilsUT+W0VgXTy2Nq+88Sideiy6UwDwpqHr5ktyoYFeSVB/aCTJl5oxDxBqs9dfeJSEAy7/AYy8kJoIE/yPYsBnGw10CAED4Rf5mfDgstkZRBdAO0xLBihkPsgza2975DVf27YSjJZ4eKrnR+G/aNCKycLQvWD/5c2bcLCJqyz0uMLQC/4LspS9b5bAKurzqFRdrD5q78NDcbodHelc7zBlFrRwGFCUjXTbQoU6r+1FA8y2Z+n1bd7mIF1JBVHurYAygyYXOcry870hyucGojonvdgBvHp6txeYyPU14VvTNwkF2mddpBCvoSTSPZ5X64="
+  matrix:
+    - TEST_TYPE=coverage
+    - TEST_TYPE=integration
+    - TEST_TYPE=core
+    - TEST_TYPE=plugin
+    - TEST_TYPE=fuzzit FUZZIT_TYPE=local-regression
+    - TEST_TYPE=fuzzit FUZZIT_TYPE=fuzzing
 
 # In the Travis VM-based build environment, IPv6 networking is not
 # enabled by default. The sysctl operations below enable IPv6.
--- a/Makefile.fuzz
+++ b/Makefile.fuzz
@@ -14,6 +14,7 @@
 #$ go get github.com/dvyukov/go-fuzz/go-fuzz-build
 
 REPO:="github.com/coredns/coredns"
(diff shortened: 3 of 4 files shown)


Makefile
```diff
@@ -44,6 +44,33 @@ ifeq ($(TEST_TYPE),coverage)
...
+	./fuzzit create job --branch $(TRAVIS_BRANCH) --revision $(TRAVIS_COMMIT) whoami ./whoami
+endif
+ifeq ($(TEST_TYPE),fuzzit-regression)
+	export GO111MODULE=off
+	go get -u github.com/dvyukov/go-fuzz/go-fuzz-build
+	go get -u ./...
+	LIBFUZZER=YES make -f Makefile.fuzz cache chaos file rewrite whoami
+	wget -O fuzzit https://github.com/fuzzitdev/fuzzit/releases/download/v2.4.8/fuzzit_Linux_x86_64
+	chmod a+x fuzzit
+	./fuzzit create job --local coredns/cache ./cache
```
miekg: what's a local fuzzing job, something running on travis?

Makefile
```diff
@@ -44,6 +44,33 @@ ifeq ($(TEST_TYPE),coverage)
...
+	./fuzzit create job --branch $(TRAVIS_BRANCH) --revision $(TRAVIS_COMMIT) chaos ./chaos
+	./fuzzit create job --branch $(TRAVIS_BRANCH) --revision $(TRAVIS_COMMIT) file ./file
+	./fuzzit create job --branch $(TRAVIS_BRANCH) --revision $(TRAVIS_COMMIT) rewrite ./rewrite
+	./fuzzit create job --branch $(TRAVIS_BRANCH) --revision $(TRAVIS_COMMIT) whoami ./whoami
+endif
+ifeq ($(TEST_TYPE),fuzzit-regression)
+	export GO111MODULE=off
+	go get -u github.com/dvyukov/go-fuzz/go-fuzz-build
+	go get -u ./...
+	LIBFUZZER=YES make -f Makefile.fuzz cache chaos file rewrite whoami
```
miekg: Makefile.fuzz finds these targets by itself, no need to name them

Makefile
```diff
@@ -44,6 +44,33 @@ ifeq ($(TEST_TYPE),coverage)
...
 endif
+ifeq ($(TEST_TYPE),fuzzit)
+	export GO111MODULE=off
+	go get -u github.com/dvyukov/go-fuzz/go-fuzz-build
+	go get -u ./...
+	LIBFUZZER=YES make -f Makefile.fuzz cache chaos file rewrite whoami
+	wget -O fuzzit https://github.com/fuzzitdev/fuzzit/releases/download/v2.4.8/fuzzit_Linux_x86_64
+	chmod a+x fuzzit
+	./fuzzit auth $(FUZZIT_API_KEY)
+	./fuzzit create job --branch $(TRAVIS_BRANCH) --revision $(TRAVIS_COMMIT) cache ./cache
```
miekg: adding each new fuzz target here, can be be cumbersome, can this be automatically be figured out?

miekg (commented):
Thanks for picking this up so quickly. Think it's very good to have this automatically enabled.

Makefile
yevgenypats: Yes, it downloads the corpus and doing a regresion so you can immediately see if the PR breaks and introduce new bugs

Makefile
yevgenypats: I run `make -f Makefile.fuzz` but it's just echo the possible commands. I didn't see a `all` target implemented though I can add one.

Makefile
yevgenypats: Agree. I think we can just add another target for fuzzit in the Makefile where we have all targets and logic anyway? sounds reasonable?

yevgenypats: Also, because go-fuzz doesnt currently support GOMODULES do you know what would be the best way to compile it in Travis and get all the dependencies?

I tried `go get -u` but it didn't work

https://travis-ci.org/coredns/coredns/jobs/568993981

miekg:
>Agree. I think we can just add another target for fuzzit in the Makefile where we have all targets and logic anyway? sounds reasonable?

yes, please do.

miekg:
>Also, because go-fuzz doesnt support GOMODULES do you know what would be the best way to compile it in Travis and get all the dependencies?

did you open an issue on go-fuzz?

>I tried `go get -u` but it didn't work
>
>https://travis-ci.org/coredns/coredns/jobs/568993981

We can fork and host it on coredns/go-fuzz in the meantime

yevgenypats:
Regarding the go module it's an open [issue](https://github.com/dvyukov/go-fuzz/issues/195). I'm not sure it's an easy fix but it will be supported. Hopefully it will also be supported as part of go standard [toolchain](https://github.com/golang/go/issues/19109). We'll have to workaround GOMODULES for now.

yevgenypats:
@miekg I'm having trouble building the go-fuzz targets without the libfuzzer support (This is even not related to this PR).

`go-fuzz-build -tags fuzz  ./plugin/chaos`

prints
```bash
/go/src/github.com/go-acme/lego/certificate/authorization.go:6: cannot find package "github.com/go-acme/lego/v3/acme" in any of:
        /tmp/go-fuzz-build563253197/goroot/src/github.com/go-acme/lego/v3/acme (from $GOROOT)
        /tmp/go-fuzz-build563253197/gopath/src/github.com/go-acme/lego/v3/acme (from $GOPATH)
/go/src/github.com/go-acme/lego/certificate/certificates.go:16: cannot find package "github.com/go-acme/lego/v3/acme/api" in any of:
        /tmp/go-fuzz-build563253197/goroot/src/github.com/go-acme/lego/v3/acme/api (from $GOROOT)
        /tmp/go-fuzz-build563253197/gopath/src/github.com/go-acme/lego/v3/acme/api (from $GOPATH)
```

I guess something connected to go modules as well.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: Add Cont..." ]
>@miekg I'm having trouble building the go-fuzz targets without the libfuzzer support (This is even not related to this PR).

Having similar compile problems as well :/

Not sure who to fix, moving go-fuzz 100% to go modules would help I guess.

yevgenypats:
mm. Yes it might help though I was able to compile the targets on my mac somehow. I'm not able to reproduce it again on a linux docker. But it would be best not to wait for go-fuzz support imho and try to workaround the issue, this should work even with repos that support go modules.

Makefile
```diff
@@ -44,6 +44,26 @@ ifeq ($(TEST_TYPE),coverage)
 		fi; \
 	done
 endif
+ifeq ($(TEST_TYPE),fuzzit)
+	# skip fuzzing for PR
+	if [ "$(TRAVIS_PULL_REQUEST)" = "false" ] || [ "$(FUZZIT_TYPE)" = "local-regression" ] ; then \
```
miekg, line 49: this comes out as `if [ "3093" = "false" ] || [ "fuzzing" = "local-regression" ] ; then \` is that correct?

Makefile
```diff
@@ -44,6 +44,26 @@ ifeq ($(TEST_TYPE),coverage)
...
+        export GO111MODULE=off; \
+        go get -u github.com/dvyukov/go-fuzz/go-fuzz-build; \
+        go get -u -v .; \
+        cd ../../go-acme/lego; \
+        git checkout v2.5.0; \
```
miekg, line 54: oh wow, this is pretty horrible :/

I think it make sense to make this into Makefile.fuzz under some phony install target, so we can call `make -f Makefile.fuzz install` here. or $(MAKE) or somethign

yevgenypats: This is the workaround for the strange glitch with go-acme v3. How do you suggest working around that? (not sure I got it regarding the make install)

miekg:
thanks for making this work!

Got a couple of optimizations that could be done in this PR or in a follow up one.

The API key def. needs some docs and where to get it/change it/remove it. And maybe some words on that it is safe to put in source code.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add Continuou..." ]
>yevgenypats commented on this pull request.
>
>
>
>> @@ -44,6 +44,26 @@ ifeq ($(TEST_TYPE),coverage)
> 		fi; \
> 	done
> endif
>+ifeq ($(TEST_TYPE),fuzzit)
>+	# skip fuzzing for PR
>+	if [ "$(TRAVIS_PULL_REQUEST)" = "false" ] || [ "$(FUZZIT_TYPE)" = "local-regression" ] ; then \
>+        export GO111MODULE=off; \
>+        go get -u github.com/dvyukov/go-fuzz/go-fuzz-build; \
>+        go get -u -v .; \
>+        cd ../../go-acme/lego; \
>+        git checkout v2.5.0; \
>
>This is the workaround for the strange glitch with go-acme v3. How do you suggest working around that? (not sure I got it regarding the make install)

No, just moving it into Makefile.fuzz instead of having it here.

yevgenypats:
@miekg Thanks for your review:) I suggest to do the optimization in a following PR. 

You need to signup at https://app.fuzzit.dev and I'll add you to the core-dns group so you can view crashes as well as get notified of new crashes/bugs (there is one bug so far as far as I can see).

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add Continuou..." ]
>@miekg Thanks for your review:) I suggest to do the optimization in a following PR.
>
>You need to signup at https://app.fuzzit.dev and I'll add you to the core-dns group so you can view crashes as well as get notified of new crashes/bugs (there is one bug so far as far as I can see).

OK, done.

yevgenypats:
added you to coredns org: https://app.fuzzit.dev/orgs/coredns/dashboard

yevgenypats:
Success it works! https://travis-ci.org/coredns/coredns/jobs/573356225

Here is the link to the crash (it classifies incorrectly as leak, Im aware of that and will push a fix soon):
https://app.fuzzit.dev/orgs/coredns/targets/rewrite/ZcyF7mWmglds6Z589dKJ

miekg:
this is awesome!

PR #3098, opened 2019-08-08
plugin/kubernetes: Don't do a zone transfer for NS requests

chrisohaver:

A `NS` request for a non-apex zone (e.g. for `anything.cluster.local.`, when the zone is `cluster.local.`), would trigger a zone transfer to the client.

This was caused by the `AXFR` switch case being inserted between the `NS` case, and default case.  I'm still looking into whether the `NS` case's `fallthrough` is necessary. It looks to be in place to handle returning of NXDOMAIN vs NODATA responses, but does not appear to work.  If it is not necessary, we should remove it.

I discovered this when looking into kubernetes/kubernetes#81137

none.

no.

### DIFF
--- a/plugin/kubernetes/handler.go
+++ b/plugin/kubernetes/handler.go
@@ -51,6 +51,7 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 			records, extra, err = plugin.NS(ctx, &k, zone, state, plugin.Options{})
 			break
 		}
+		_, err = parseRequest(state)
 	default:
 		// Do a fake A lookup, so we can distinguish between NODATA and NXDOMAIN
 		_, err = plugin.A(ctx, &k, zone, state, nil, plugin.Options{})
--- a/plugin/kubernetes/handler.go
+++ b/plugin/kubernetes/handler.go
@@ -51,7 +51,6 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 			records, extra, err = plugin.NS(ctx, &k, zone, state, plugin.Options{})
 			break
 		}
-		fallthrough
 	default:
 		// Do a fake A lookup, so we can distinguish between NODATA and NXDOMAIN
 		_, err = plugin.A(ctx, &k, zone, state, nil, plugin.Options{})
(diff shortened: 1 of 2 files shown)


chrisohaver:
... also should add a test.

chrisohaver:
I removed the `NS` case fallthough, since it doesn't appear to be accomplishing what it is supposed to. Handling NXDOMAIN/NODATA responses properly in kubernetes is complicated by the generic backend.

chrisohaver:
> Handling NXDOMAIN/NODATA responses properly in kubernetes is complicated by the generic backend.

We can at least validate if the first subdomain exists or not by parsing the request (which we otherwise don't do in a NS request).  This allows us to return `NXDOMAIN` for `foo.cluster.local`, and `NODATA` for `svc.cluster.local`... this needs further work to handle existing/non-existing namespaces but i'll leave fixing that to future PR.

plugin/kubernetes/handler.go
```diff
@@ -44,14 +44,14 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 		records, extra, err = plugin.SRV(ctx, &k, zone, state, plugin.Options{})
 	case dns.TypeSOA:
 		records, err = plugin.SOA(ctx, &k, zone, state, plugin.Options{})
+	case dns.TypeAXFR, dns.TypeIXFR:
```
miekg: nit: I would put these first, or, at least, not in the middle as xfr is distinctly different from the other queries

plugin/kubernetes/handler.go
```diff
@@ -44,14 +44,14 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
...
+		k.Transfer(ctx, state)
 	case dns.TypeNS:
 		if state.Name() == zone {
 			records, extra, err = plugin.NS(ctx, &k, zone, state, plugin.Options{})
 			break
 		}
-		fallthrough
-	case dns.TypeAXFR, dns.TypeIXFR:
-		k.Transfer(ctx, state)
+		_, err = parseRequest(state)
```
miekg: why this?
the fallthrough was here to do nodata/nxdomain for ns queries other than the apex

chrisohaver: Yeah, I know _why_ it was there. But it wasn't working, (and there were no tests to validate that case) - it always returns NODATA, regardless of whether or not the subdomain exists.

Perhaps it makes better sense to restore the fallthrough, and then fix that issue in a separate PR.

chrisohaver:
@miekg, I moved the xfr case to the top as suggested.  I also restored the fallthrough.  I had to remove one of the test cases I added.  IMO that can be fixed in a separate PR since it's not directly related to this issue (I just ran into it when testing).

miekg:
>@miekg, I moved the xfr case to the top as suggested.  I also restored the fallthrough.  I had to remove one of the test cases I added.  IMO that can be fixed in a separate PR since it's not directly related to this issue (I just ran into it when testing).

sgtm, this good to have this fixed (and tested)

PR #3102, opened 2019-08-09
plugin/azure: clean up readme

miekg:
document the environment option and some cleanups. Go over the code and
fix/tweak random bits here and there.

Condense a few lines here and there.

### DIFF
--- a/plugin/azure/README.md
+++ b/plugin/azure/README.md
@@ -6,9 +6,9 @@
 
 ## Description
 
-The azure plugin is useful for serving zones from Microsoft Azure DNS.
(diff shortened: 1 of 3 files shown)


miekg:
@darshanime the example in the README needs some tweaking, those ID can't be all the same (if changed them a little, but more is needed).
**ENVIRONMENT** needs more explaining and an example. Also the minimal set of credentials to access stuff in Azure needs some kind of explanation.
(all of the above can happen in a subsequent PR by you)

yongtang (approved):
lgtm

PR #3105, opened 2019-08-09
plugin/kubernetes: Fix non-apex NS queries NXDOMAIN/NODATA responses

chrisohaver:
Previously non-apex NS queries always return NODATA, even if the subdomain does not actually exist (e.g. `foo.cluster.local`).  This PR checks the service records to determine if there is a record at the same level (NODATA), or if there is not (NXDOMAIN).

**This PR depends on, and includes parts of the fix in #3098.  So, #3098 should merge first, then this PR can be rebased with master, and merged.**

#3098

none

no


chrisohaver:
I think i see a cleaner fix... closing for now ...

PR #3108, opened 2019-08-10
plugin/route53: various updates

miekg:
In the setup function use plugin.Error() to wrap the errors with the
plugin name. Because there isn't a separate setup() function this is
done for all returned errors.

Remove *upstream.Upstream from the New parameters as this is always set
and adjust the tests to account for this.


miekg:
this is pure polish

yongtang (approved):
lgtm

PR #3110, opened 2019-08-11
Fix handling truncated responses in forward

ameshkov:
Currently, `forward` plugin does not handle truncated responses as it was supposed to because there's no error returned when a truncated response is received

None

None

On the contrary

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -109,9 +109,6 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 		opts := f.opts
 		for {
 			ret, err = proxy.Connect(ctx, state, opts)
-			if err == nil && ret != nil && !ret.Truncated {
-				break
-			}
 			if err == ErrCachedClosed { // Remote side closed conn, can only happen with TCP.
 				continue
 			}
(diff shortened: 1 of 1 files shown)


rdrozhdzh:
@ameshkov, could you clarify your use case? I guess the only problematic case here is when user always sends DNS queries by TCP (e.g. via TLS connection) and forward plugin is configured with `prefer_udp` option. 
In other cases, the response should get back to user with TC flag set and user should resend the query over TCP.

ameshkov:
@rdrozhdzh that's right, when `prefer_udp` is specified, `forward` plugin [was supposed](https://github.com/coredns/coredns/pull/3110/files#diff-b12d06e8aa6117ce05d5715e1c8930c3L118) to handle it and repeat the request automatically, and I guess it was working okay until some recent changes in `miekg/dns` -- when ErrTruncated was deprecated.

UladzimirTrehubenka:
ErrTruncated was deprecated in https://github.com/miekg/dns/pull/815 (27 Nov 2018) - delivered in v 1.1.0 (29 Nov 2018)

ameshkov:
So not so recent then:). It has been broken since then.

plugin/forward/forward.go
```diff
@@ -109,7 +109,7 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 		opts := f.opts
 		for {
 			ret, err = proxy.Connect(ctx, state, opts)
-			if err == nil {
+			if err == nil && ret != nil && !ret.Truncated {
```
rdrozhdzh: I would suggest to remove this `if` block at all. This would also fix the problem.

rdrozhdzh (approved):
looks good to me

Issue #3041, opened 2019-07-24
Remove federation plugin


miekg:
Kubernetes DNS federation has failed, and the federation plugin serve no useful services. I propose we move it to coredns/federation and out of the main tree

chrisohaver:
Sounds fair.  Following the deprecation policy, I propose we announce it as "deprecated" in the next release (1.6.0 i think). 
Then in a release after that make it a no-op, remove the coredns/ci tests, and move federation out to an external plugin.
Followed by complete removal of the no-op plugin in a release after that.

Regarding the Kubernetes project - we'll declare it deprecated there too in the next release (assuming we deploy 1.6.0).  If there is concern in the K8s dev community about nooping the plugin when it comes to that time, we can build coredns with the external plugin and deploy that image to the gcr.io image repo if necessary.

sharmaansh21:
I will send the PR :bowtie:

chrisohaver:
> I will send the PR :bowtie:

For which parts? (multiple PRs and multiple repositories involved here)

sharmaansh21:
coredns/coredns
coredns/ci

@chrisohaver  where else if you can please point me, I am more than happy to look into it.

chrisohaver:
Cool, another step would be moving the federation plugin to coredns/federation, and preparing it as an external plugin.

sharmaansh21:
@chrisohaver Sorry for the late reply I was on holidays, by external plugin you mean something like this https://github.com/coredns/proxy

yongtang:
Added a PR #3139.

miekg:
that pr was merged.

PR #3139, opened 2019-08-17
Move federation plugin to github.com/coredns/federation

yongtang:
This PR is part of the effort of #3041 to remove federation plugin.

This is the first step to move federation to github.com/coredns/federation

Note the github.com/coredns/federation was done through 
```
git filter-branch --subdirectory-filter plugin/federation -- --all
```

so related git history are kept.


miekg:
what do we do with `plugin/kubernetes/federation.go`? Is there any way this can be fully deprecated?

yongtang:
> what do we do with plugin/kubernetes/federation.go? Is there any way this can be fully deprecated?

@miekg It should be possible, although that may need some surgical moving around. Let me update the PR and see if works.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Move federati..." ]
>> what do we do with plugin/kubernetes/federation.go? Is there any way this can be fully deprecated?
>
>@miekg It should be possible, although that may need some surgical moving around. Let me update the PR and see if works.

but what I mean is, even with the code in a new repo; if we remove this 
interface into the kubernetes plugin it will stop working. So recompiling 
coredns to re-add *federation* will be more involved, because you have to patch 
more.

I'm fine with fully deprecating it, but it will be gone **gone** at some point.

yongtang:
@miekg We could move all related logic to coredns/federation there fully. I think we only need to export `localNodeName` in Kubernetes plugin (for now, until fully deprecated). Let me see if it works.

yongtang:
@miekg I think the PR is ready, please take a look and see if it works.

miekg (approved):

PR #3160, opened 2019-08-20
plugin/k8s_external/kubernetes: handle NS records

### LINKED ISSUE #3064: plugin/k8s_external
using coredns 1.3.1 and the k8s_external plugin i miss some A records in the dns subzone

<pre>
.:53 {
    errors
    health
    kubernetes k8s.cluster.local in-addr.arpa ip6.arpa {
       pods insecure
       upstream
       ttl 30
    }
    k8s_external k9s.cluster.local in-addr.arpa ip6.arpa {
       ttl 30
    }
    prometheus :9153
    forward . /etc/resolv.conf
    cache 30
    loop
    reload
    loadbalance
}
</pre>

the documentation notes that there is a stub/apex zone. But i'm missing an NS record and an A record
<pre>
dig NS k9s.cluster.local => invalid dns package
dig NS dns.k9s.cluster.local =>  no entry
dig A ns1.dns.k9s.cluster.local => no entry
</pre>

The kubernetes plugin handles them like
<pre>
dig NS k8s.cluster.local => ns.dns.k8s.cluster.local.
dig NS dns.k8s.cluster.local => no entry
dig A ns.dns.k8s.cluster.local => 10.96.x.y
</pre>

chrisohaver:

Fixes #3064:
* don't SERVFAIL when querying NS  `cluster.local.` in k8s_external
* reply with a record for `ns1.dns.cluster.local.`  in k8s_external

While reworking `nsAddr()` to fix #3064, I made the following related fixes/improvements which apply to both k8s_external and kubernetes:
* optimization to use svc object indexes (previously iterated entire service list, which is expensive)
* handle the ipv6 case (where coredns service is ipv6 - previously ipv4 was assumed)
* handle case where more than one k8s service targets the coredns endpoint
* handle corner case when a headless k8s service points to the coredns endpoint

#3064

None

No

### DIFF
--- a/plugin/kubernetes/external.go
+++ b/plugin/kubernetes/external.go
@@ -85,8 +85,10 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
 
 // ExternalAddress returns the external service address(es) for the CoreDNS service.
 func (k *Kubernetes) ExternalAddress(state request.Request) []dns.RR {
-	// This is probably wrong, because of all the fallback behavior of k.nsAddr, i.e. can get
-	// an address that isn't reachable from outside the cluster.
-	rrs := []dns.RR{k.nsAddr()}
-	return rrs
+	// If CoreDNS is running inside the Kubernetes cluster: k.nsAddrs() will return the external IPs of the services
+	// targeting the CoreDNS Pod.
+	// If CoreDNS is running outside of the Kubernetes cluster: k.nsAddrs() will return the first non-loopback IP
+	// address seen on the local system it is running on. This could be the wrong answer if coredns is using the *bind*
+	// plugin to bind to a different IP address.
+	return k.nsAddrs(true, state.Zone)
 }
--- a/plugin/kubernetes/ns.go
+++ b/plugin/kubernetes/ns.go
@@ -13,10 +13,8 @@ func isDefaultNS(name, zone string) bool {
 	return strings.Index(name, defaultNSName) == 0 && strings.Index(name, zone) == len(defaultNSName)
 }
 
-func (k *Kubernetes) nsAddr(zone string) dns.RR {
-	return k.nsAddrs(false, zone)[0]
-}
-
+// nsAddrs returns the A or AAAA records for the CoreDNS service in the cluster. If the service cannot be found,
+// it returns a record for the the local address of the machine we're running on.
 func (k *Kubernetes) nsAddrs(external bool, zone string) []dns.RR {
 	var (
 		svcNames []string
--- a/plugin/kubernetes/ns.go
+++ b/plugin/kubernetes/ns.go
@@ -4,6 +4,7 @@ import (
 	"net"
 	"strings"
 
+	"github.com/coredns/coredns/plugin/kubernetes/object"
 	"github.com/miekg/dns"
 	api "k8s.io/api/core/v1"
 )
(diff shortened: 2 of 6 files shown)


plugin/kubernetes/ns.go
```diff
@@ -12,54 +13,74 @@ func isDefaultNS(name, zone string) bool {
...
 	var (
-		svcName      string
-		svcNamespace string
+		svcNames []string
+		svcIPs   []net.IP
 	)
 
-	rr := new(dns.A)
+	// Find the CoreDNS Endpoint
 	localIP := k.interfaceAddrsFunc()
```
johnbelamaric, line 25: FYI, I just looked, and this always returns the IPv4 address. Not something for this PR, but perhaps it needs to go on the k8s IPv6 support list? It's probably going to need to return both...or all addresses? (right now it's just one).

plugin/kubernetes/external.go
```diff
@@ -85,8 +85,10 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
...
 func (k *Kubernetes) ExternalAddress(state request.Request) []dns.RR {
-	// This is probably wrong, because of all the fallback behavior of k.nsAddr, i.e. can get
-	// an address that isn't reachable from outside the cluster.
-	rrs := []dns.RR{k.nsAddr()}
-	return rrs
+	// If CoreDNS is running inside the Kubernetes cluster: k.nsAddrs() will return the external IPs of the services
+	// targeting the CoreDNS Pod.
+	// If CoreDNS is running outside of the Kubernetes cluster: k.nsAddrs() will return the first non-loopback IP
+	// address seen on the local system it is running on. This could be the wrong answer if coredns is using the *bind*
+	// plugin to bind to a different IP address.
```
johnbelamaric, line 92: Huh. Is there anything we can do about this? Maybe we need to get it from the server object? Can we get to that here?

plugin/kubernetes/ns_test.go
```diff
@@ -14,24 +15,48 @@ func (APIConnTest) HasSynced() bool                          { return true }
...
+			Namespace: "kube-system",
+			ClusterIP: "10::111",
+		},
 	}
 	return svcs
 }
 
-func (APIConnTest) EpIndexReverse(string) []*object.Endpoints {
+func (APIConnTest) EpIndexReverse(ip string) []*object.Endpoints {
+	if ip != "127.0.0.1" {
```
johnbelamaric: it's kind of confusing that we use loopback here, given that in general this wouldn't be an endpoint IP. unless I am not understanding something.

chrisohaver: Yeah - i think the test could stub in a different IP to be clearer.

plugin/kubernetes/external.go
chrisohaver: I think we could do it in the onStartUp event.

plugin/kubernetes/ns.go
chrisohaver: Yeah - it should return all addresses of all interfaces, and also check for the _bind_ plugin, if present. I was thinking to fixing this in a separate PR.  This PR getting a bit large, but I can fix it in this PR if thats preferable. 

I missed the IPV6 coredns pod ip bit, thanks! I think I should fix that in this PR, since I have other ipv6 handling in this PR.

johnbelamaric:
Ok, looks good. Please fix the `bind` and multiple IP issue in another PR.

/lgtm

chrisohaver:
@miekg - would you like to give this a once over or OK to merge?

@johnbelamaric, I do have the bind/multiple IP case (almost) working, but will submit that in a separate PR.

Issue #3164, opened 2019-08-21
Configuration validation


jgsqware:
Hello, when in my Corefile, I have a server block no space between the end of the name and the `{`, the server block is wrongly interpreted.

Corefile:
```
www.mydomain.com:1053 {
  cache 30
  forward . 8.8.8.8
}
```

CoreDNS logs:

```
# /coredns -conf ./c/Corefile
www.mydomain.com.:1053
2019-08-21T10:44:02.806Z [INFO] plugin/reload: Running configuration MD5 = 6740acee942add7a7ad7a726502e6ab4
2019-08-21T10:44:02.806Z [INFO] CoreDNS-1.6.2
2019-08-21T10:44:02.806Z [INFO] linux/amd64, go1.12.7, 25632d6a
CoreDNS-1.6.2
linux/amd64, go1.12.7, 25632d6a
^C[INFO] SIGINT: Shutting down
```

Corefile:
```
www.mydomain.com:1053{
  cache 30
  forward . 8.8.8.8
}
```

CoreDNS logs:

```
# /coredns -conf ./c/Corefile
www.mydomain.com:1053{.:53
2019-08-21T10:44:02.806Z [INFO] plugin/reload: Running configuration MD5 = 6740acee942add7a7ad7a726502e6ab4
2019-08-21T10:44:02.806Z [INFO] CoreDNS-1.6.2
2019-08-21T10:44:02.806Z [INFO] linux/amd64, go1.12.7, 25632d6a
CoreDNS-1.6.2
linux/amd64, go1.12.7, 25632d6a
^C[INFO] SIGINT: Shutting down
```

Is it possible to validate it before? I think `www.mydomain.com:1053{.:53` isn't valid

jgsqware:
what about validating here the zone url in the normalizeZone:
https://github.com/coredns/coredns/blob/master/core/dnsserver/address.go#L34

I guess it should be a `host:port` valid?

jgsqware:
this is panicking, https://play.golang.org/p/pTHNdIyFqxJ

PR #3165, opened 2019-08-21
Validate zone during normalization

jgsqware:
There is currently no validation on host zone, which lead to incorrect behavior after parsing zone in the Corefile.
I've added a RequestURL validation after the normalization to verify if the final zone is valid or not.
https://github.com/coredns/coredns/issues/3164
-
No

### DIFF
--- a/core/dnsserver/address.go
+++ b/core/dnsserver/address.go
@@ -3,6 +3,7 @@ package dnsserver
 import (
 	"fmt"
 	"net"
+	"net/url"
 	"strings"
 
 	"github.com/coredns/coredns/plugin"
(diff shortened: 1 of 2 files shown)


chrisohaver (approved):
lgtm, thanks

jgsqware:
:heart:

miekg:
The zone/ address being tested is a weird, but completely valid domain
name, so this patch is not correct.

jgsqware:
@miekg the adress can be anything?
I'm just wondering how to identify wrong zone because of missing space in the zone black to be honest. Like reference in the issue

chrisohaver:
> The zone/ address being tested is a weird, but completely valid domain
> name, so this patch is not correct.

Yeah - strictly speaking you are correct.  I figured you may take issue, which is why I approved, but didn't merge.

The domain `www.mydomain.com:1053{` is in itself a valid domain (because `com:1053{` is a valid dns segment according to RFC).

I guess Caddy doesn't care about the unmatched curly bracket.

yongtang:
@miekg @chrisohaver Didn't know `com:1053{` could be a valid domain before. Have created a PR #3167  to revert the patch.

@jgsqware I think something might be possible to check the caddy parser. Will have to take a deeper look first.

chrisohaver:
Yeah - In retrospect, this causes backward incompatibility for a user that is serving a domain with weird stuff in it... even if the domain is quoted ... e.g.
```
"my strange.but{completely valid}.domain.":53  {
}
```
... would stop working.

miekg:
It *is* strange that the caddyfile parser doesn't complain.

chrisohaver:
The caddy parser is more `\n` aware than it is `{` aware ... 

e.g. it accepts ...

```
.
}
```

miekg:
Ack. Might be worth sending a fix upstream to caddy if they want one.

jgsqware:
I will check it tomorrow

PR #3171, opened 2019-08-22
Measure and expose DNS programming latency from Kubernetes plugin.

oxddr:
For now metric is measure only for headless services. Informer has been slighlty
refactored, so the code can measure latency without storing extra fields on
Endpoint struct.

Measure and expose DNS programming latency from Kubernetes plugin.

ref https://github.com/kubernetes/perf-tests/issues/617

none 

no

### DIFF
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -229,3 +229,19 @@ plugin is also enabled:
  * kubernetes/service: the service name in the query
  * kubernetes/client-namespace: the client pod's namespace, if `pods verified` mode is enabled
  * kubernetes/client-pod-name: the client pod's name, if `pods verified` mode is enabled
+
+## Metrics
+
+The *kubernetes* plugin exports the following *Prometheus* metrics.
+* `coredns_kubernetes_dns_programming_latency_seconds{service_kind}` - exports the
+[DNS programming latency SLI](https://github.com/kubernetes/community/blob/master/sig-scalability/slos/dns_programming_latency.md).
+  The metrics has the `service_kind` label that identifies the kind of the
+  [kubernetes service](https://kubernetes.io/docs/concepts/services-networking/service).
+  It may take one of the three values:
+    * `cluster_ip`
+    * `headless_with_selector`
+    * `headless_without_selector`
+
+## Bugs
+
+ * add support for other service types; only "headless_with_selector" is supported now
--- a/plugin/kubernetes/metrics.go
+++ b/plugin/kubernetes/metrics.go
@@ -1,0 +1,76 @@
+package kubernetes
+
+import (
+	"time"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/kubernetes/object"
+	"github.com/prometheus/client_golang/prometheus"
+	api "k8s.io/api/core/v1"
+)
+
+const (
+	subsystem = "kubernetes"
+)
+
+var (
+	// DnsProgrammingLatency is defined as the time it took to program a DNS instance - from the time
+	// a service or pod has changed to the time the change was propagated and was available to be
+	// served by a DNS server.
+	// The definition of this SLI can be found at https://github.com/kubernetes/community/blob/master/sig-scalability/slos/dns_programming_latency.md
+	// Note that the metrics is partially based on the time exported by the endpoints controller on
+	// the master machine. The measurement may be inaccurate if there is a clock drift between the
+	// node and master machine.
+	// The service_kind label can be one of:
+	//   * cluster_ip
+	//   * headless_with_selector
+	//   * headless_without_selector
+	DnsProgrammingLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: subsystem,
+		Name:      "dns_programming_latency_seconds",
+		// From 1 millisecond to ~17 minutes.
+		Buckets: prometheus.ExponentialBuckets(0.001, 2, 20),
+		Help:    "Histogram of the time (in seconds) it took to program a dns instance.",
+	}, []string{"service_kind"})
+
+	// durationSinceFunc returns the duration elapsed since the given time.
+	// Added as a global variable to allow injection for testing.
+	durationSinceFunc = time.Since
+)
+
+func recordDNSProgrammingLatency(svcs []*object.Service, endpoints *api.Endpoints) {
+	// getLastChangeTriggerTime is the time.Time value of the EndpointsLastChangeTriggerTime
+	// annotation stored in the given endpoints object or the "zero" time if the annotation wasn't set
+	var lastChangeTriggerTime time.Time
+	stringVal, ok := endpoints.Annotations[api.EndpointsLastChangeTriggerTime]
+	if ok {
+		ts, err := time.Parse(time.RFC3339Nano, stringVal)
+		if err != nil {
+			log.Warningf("DnsProgrammingLatency cannot be calculated for Endpoints '%s/%s'; invalid %q annotation RFC3339 value of %q",
+				endpoints.GetNamespace(), endpoints.GetName(), api.EndpointsLastChangeTriggerTime, stringVal)
+			// In case of error val = time.Zero, which is ignored in the upstream code.
+		}
+		lastChangeTriggerTime = ts
+	}
+
+	// isHeadless indicates whether the endpoints object belongs to a headless
+	// service (i.e. clusterIp = None). Note that this can be a  false negatives if the service
+	// informer is lagging, i.e. we may not see a recently created service. Given that the services
+	// don't change very often (comparing to much more frequent endpoints changes), cases when this method
+	// will return wrong answer should be relatively rare. Because of that we intentionally accept this
+	// flaw to keep the solution simple.
+	isHeadless := len(svcs) == 1 && svcs[0].ClusterIP == api.ClusterIPNone
+
+	if endpoints == nil || !isHeadless || lastChangeTriggerTime.IsZero() {
+		return
+	}
+
+	// If we're here it means that the Endpoints object is for a headless service and that
+	// the Endpoints object was created by the endpoints-controller (because the
+	// LastChangeTriggerTime annotation is set). It means that the corresponding service is a
+	// "headless service with selector".
+	DnsProgrammingLatency.WithLabelValues("headless_with_selector").
+		Observe(durationSinceFunc(lastChangeTriggerTime).Seconds())
+
+}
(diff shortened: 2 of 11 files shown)


oxddr:
@johnbelamaric any chance you can have a look at this PR? It follows your suggestions on how measuring should be done without increasing memory footprint.

plugin/kubernetes/object/informer.go
```diff
@@ -5,19 +5,25 @@ import (
...
-func NewIndexerInformer(lw cache.ListerWatcher, objType runtime.Object, h cache.ResourceEventHandler, indexers cache.Indexers, convert ToFunc) (cache.Indexer, cache.Controller) {
+// NewIndexerInformer is a copy of the cache.NewIndexerInformer function, but allows custom process function
+func NewIndexerInformer(lw cache.ListerWatcher, objType runtime.Object, h cache.ResourceEventHandler, indexers cache.Indexers, builder ProcessorBuilder) (cache.Indexer, cache.Controller) {
 	clientState := cache.NewIndexer(cache.DeletionHandlingMetaNamespaceKeyFunc, indexers)
 
-	fifo := cache.NewDeltaFIFO(cache.MetaNamespaceKeyFunc, clientState)
-
 	cfg := &cache.Config{
-		Queue:            fifo,
+		Queue:            cache.NewDeltaFIFO(cache.MetaNamespaceKeyFunc, clientState),
```
johnbelamaric, line 13: in general please do this sort of cleanup in a separate PR to reduce the review burden

plugin/kubernetes/metrics_test.go
```diff
@@ -0,0 +1,131 @@
...
+const (
+	namespace = "testns"
+)
+
+func TestDnsProgrammingLatency(t *testing.T) {
+	client := fake.NewSimpleClientset()
+	now := time.Now()
+	controller := newdnsController(client, dnsControlOpts{
+		initEndpointsCache: true,
+		// This is needed as otherwise the fake k8s client doesn't work properly.
```
johnbelamaric, line 27: Can you explain this more? If we can eliminate `skipAPIObjectsCleanup` this all becomes a lot simpler. I don't remember why we need that as an option or see anywhere it gets set?

plugin/kubernetes/object/informer.go
oxddr: Sure thing, I can remove it. This one simply felt like a trivial change.

johnbelamaric: Not necessary, was just a note for the future.

plugin/kubernetes/metrics_test.go
oxddr: This is needed purely for testing purposes. If we do cleanup resources in test, cache doesn't work properly and tests fail. I am not at my workstation at the moment, so I can't check the exact effect. 

One option is to remove tests. However this non-trivial change to the code and I'd rather have it tested.

plugin/kubernetes/object/informer.go
oxddr: Noted :)

plugin/kubernetes/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+	return val
+}
+
+// isEndpointForHeadlessService returns true whether the endpoints object belongs to a headless
+// service (i.e. clusterIp = None). Note that this method may return false negatives if the service
+// informer is lagging, i.e. we may not see a recently created service. Given that the services
+// don't change very often (comparing to much more frequent endpoints changes), cases when this method
+// will return wrong answer should be relatively rare. Because of that we intentionally accept this
+// flaw to keep the solution simple.
+func isEndpointForHeadlessService(svcs []*object.Service, endpoints *api.Endpoints) bool {
```
johnbelamaric: This is only used once. Instead just inline it (along with the comment)

plugin/kubernetes/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+	// LastChangeTriggerTime annotation is set). It means that the corresponding service is a
+	// "headless service with selector".
+	DnsProgrammingLatency.WithLabelValues("headless_with_selector").
+		Observe(durationSinceFunc(lastChangeTriggerTime).Seconds())
+
+}
+
+// getLastChangeTriggerTime returns the time.Time value of the EndpointsLastChangeTriggerTime
+// annotation stored in the given endpoints object or the "zero" time if the annotation wasn't set
+func getLastChangeTriggerTime(endpoints *api.Endpoints) time.Time {
```
johnbelamaric: this is only used once, instead just inline it along with the comment

plugin/kubernetes/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+	// durationSinceFunc returns the duration elapsed since the given time.
+	// Added as a global variable to allow injection for testing.
+	durationSinceFunc = time.Since
+)
+
+func recordDNSProgrammingLatency(svcs []*object.Service, endpoints *api.Endpoints) {
+	lastChangeTriggerTime := getLastChangeTriggerTime(endpoints)
+
+	if endpoints == nil || !isEndpointForHeadlessService(svcs, endpoints) || lastChangeTriggerTime.IsZero() {
+		return
```
johnbelamaric, line 66: so we're not recording the latency for programming clusterIP or headless_without_selector? For clusterIP, I would think it's the service creation time vs. now - since clusterIP is immutable I think that should work.

For `headless_without_selector`, maybe the endpoints creation time...or update time if we have those in metadata. I recall you saying that we need the annotation though I can recall the exact reason offhand.

plugin/kubernetes/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+func getLastChangeTriggerTime(endpoints *api.Endpoints) time.Time {
+	stringVal, ok := endpoints.Annotations[api.EndpointsLastChangeTriggerTime]
+	if !ok {
+		// It's possible that the Endpoints object won't have the EndpointsLastChangeTriggerTime
+		// annotation set. In that case return the 'zero value', which is ignored in the upstream code.
+		return time.Time{}
+	}
+	val, err := time.Parse(time.RFC3339Nano, stringVal)
+	if err != nil {
+		log.Warningf("Error while parsing EndpointsLastChangeTriggerTimeAnnotation: '%s'. Error is %v",
```
johnbelamaric: Could this flood the logs?

If we're going to be wordy we should include enough info to debug and explain the actual effect we are warning about. The error value is not going to be useful I don't think, but knowing what it's supposed to look like will help.

```
log.Warningf("DnsProgrammingLatency cannot be calculated for Endpoints '%s/%s'; invalid %q annotation RFC3339 value of %q",
  endpoints.GetNamespace(), endpoints.GetName(), api.EndpointsLastChangeTriggerTime, stringVal)
```

johnbelamaric (changes requested):

plugin/kubernetes/metrics.go
oxddr: Thanks for the snippet!

I doubt. This annotations is machine-generated and it's not coming from user. I'd assume it should be formatted correctly unless we have a bug (should be easy to catch in tests) or a pathological case (user tampering with it).

plugin/kubernetes/metrics.go
oxddr: Yes, this one does work only for headless services as for now. I was planning to follow up with the rest once this PR is hopefully merged. I assume adding those two missing types should be relatively easy.

oxddr:
@johnbelamaric PTAL :)

johnbelamaric:
I think we're OK here but I'd like to have @chrisohaver take a look as well.

/assign: @chrisohaver 
/lgtm

plugin/kubernetes/metrics_test.go
```diff
@@ -0,0 +1,131 @@
+package kubernetes
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/coredns/coredns/plugin/kubernetes/object"
```
miekg, line 8: needs a newline after our local imports.

miekg (commented):
this misses an explanation of the new metrics in the README.md

plugin/kubernetes/object/object.go
```diff
@@ -16,14 +16,18 @@
 package object
 
 import (
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+        "k8s.io/apimachinery/pkg/apis/meta/v1"
```
chrisohaver: maybe run a `gofmt` to clean it up this kind of thing...

plugin/kubernetes/object/service.go
```diff
@@ -27,7 +27,13 @@ type Service struct {
 func ServiceKey(name, namespace string) string { return name + "." + namespace }
 
 // ToService converts an api.Service to a *Service.
```
chrisohaver:
```suggestion
// ToService returns a function that converts an api.Service to a *Service.
```

plugin/kubernetes/object/pod.go
```diff
@@ -17,7 +17,13 @@ type Pod struct {
 }
 
 // ToPod converts an api.Pod to a *Pod.
```
chrisohaver:
```suggestion
// ToPod returns a function that converts an api.Pod to a *Pod.
```

chrisohaver (approved):

plugin/kubernetes/object/object.go
oxddr: Formatted.

oxddr:
> this misses an explanation of the new metrics in the README.md

Explanation added.

plugin/kubernetes/README.md
```diff
@@ -229,3 +229,17 @@ plugin is also enabled:
...
+* `coredns_kubernetes_dns_programming_latency_seconds{service_kind}` - exports the
+[DNS programming latency SLI](https://github.com/kubernetes/community/blob/master/sig-scalability/slos/dns_programming_latency.md).
+  The metrics has the `service_kind` label that identifies the kind of the
+  [kubernetes service](https://kubernetes.io/docs/concepts/services-networking/service).
+  It may take one of the three values:
+    * `cluster_ip`
+    * `headless_with_selector`
+    * `headless_without_selector`
+
+TODO(oxddr): add support for other types - only "headless_with_selector" is supported now
```
miekg: just remove this TODO for now, or move it down below to a BUGS section.

oxddr: Added "Bugs" section below. I'd rather make it clear, that not all cases are supported. 

For the record, I'm planning to add a support for them soon. I'd rather have this PR merged so we can integrate Kubernetes testing pipeline in parallel.

miekg (commented):
no worries. I'll also propose some small changes to have things more inline with other READMEs. But let's merge this first.

oxddr:
Thank you all for the review!

PR #3175, opened 2019-08-22
fuzzing: allow setup function to be called

miekg:
This allows to fuzzing of more interesting targets that require setup.


miekg:
merging because trivial patch and allow us to work on more fuzzing thingies.

miekg:
reading the go-fuzz code and trying to make this it looks like this is not the right approach. Going to play some more and otherwise revert this.

PR #3177, opened 2019-08-22
plugin/file: respond correctly to IXFR message

### LINKED ISSUE #3176: plugin/file: stop doing transfer on ixfr
Thanks to some nice work from @wtoorop. 

CoreDNS sends out an AXFR when the remote server sends an IXFR request with a SOA. I should compare the SOA with the one it has and then answer with a (smaller) reply.

This explains the number of zone transfers I was seeing.

miekg:
Respond with a sing SOA record to an IXFR request if the SOA serials
match.

The added test fails on the current code with:

~~~
=== RUN   TestIxfrResponse
--- FAIL: TestIxfrResponse (0.00s)
    secondary_test.go:122: Expected answer section with single RR
FAIL
exit status 1
~~~

And obviously passes with the new code. This should cut down on the
weird number of zone transfers that I was seeing. At some point IXFR
support might be cool.

Fixes: #3176

### DIFF
--- a/plugin/file/xfr.go
+++ b/plugin/file/xfr.go
@@ -69,6 +69,9 @@ func (x Xfr) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (in
 	return dns.RcodeSuccess, nil
 }
 
+// Name implements the plugin.Handler interface.
+func (x Xfr) Name() string { return "xfr" }
+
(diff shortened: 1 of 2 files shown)


miekg:
relevant RFC 1995,  Section 2, para 4

miekg:
only fuzzing this is failling

miekg:
I need this to go in, so I can start testing it on my server.

PR #3178, opened 2019-08-22
fuzz: fix rewrite crash by fixing fuzz/do.go

miekg:
DNS packets that don't have a valid question section where fuzzed into
plugins. This can't happen in the server because they are handled before
the server calls the plugins. Amend do.go to do the same thing and
actually fuzz things.

### DIFF
--- a/plugin/pkg/fuzz/do.go
+++ b/plugin/pkg/fuzz/do.go
@@ -13,15 +13,19 @@ import (
 // Do will fuzz p - used by gofuzz. See Makefile.fuzz for comments and context.
 func Do(p plugin.Handler, data []byte) int {
 	ctx := context.TODO()
-	ret := 1
 	r := new(dns.Msg)
 	if err := r.Unpack(data); err != nil {
-		ret = 0
+		return 0 // plugin will never be called when this happens.
+	}
+	// If the data unpack into a dns msg, but does not have a proper question section discard it.
+	// The server parts make sure this is true before calling the plugins; mimic this behavior.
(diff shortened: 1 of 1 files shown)


plugin/pkg/fuzz/do.go
```diff
@@ -13,15 +13,19 @@ import (
...
 	ctx := context.TODO()
-	ret := 1
 	r := new(dns.Msg)
 	if err := r.Unpack(data); err != nil {
-		ret = 0
+		return 0 // plugin will never be called when this happens.
+	}
+	// If the data unpack into a dns msg, but does not have a proper question section discard it.
+	// The server parts make sure this is true before calling the plugins; mimic this behavior.
+	if len(r.Question) != 1 {
```
jameshartig: Should this only bail out if the len is 0 and continue if its >1?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] fuzz: fix rew..." ]
>>  	r := new(dns.Msg)
> 	if err := r.Unpack(data); err != nil {
>-		ret = 0
>+		return 0 // plugin will never be called when this happens.
>+	}
>+	// If the data unpack into a dns msg, but does not have a proper question section discard it.
>+	// The server parts make sure this is true before calling the plugins; mimic this behavior.
>+	if len(r.Question) != 1 {
>
>Should this only bail out if the len is 0 and continue if its >1?

good question. *looking at code*

`if r == nil || len(r.Question) == 0 {` is the actual check, so this is too 
strict.

Pushing fix

PR #3185, opened 2019-08-24
fuzz: use gofuzz build tag instead of fuzz

### LINKED ISSUE #3174: fuzz build should use gofuzz build tag
Mainly creating issue, so I wont forget. 
https://github.com/dvyukov/go-fuzz#random-notes notes the fuzzing is built with `gofuzz` build tag, we use `fuzz`. We should move to gofuzz.

this may be easy for a beginner, but you may run into go build + go fuzz issues

mfrw:
Since go-fuzz does not support Go modules yet, vendor dependencies.

Signed-off-by: Muhammad Falak R Wani <falakreyaz@gmail.com>

Use build tag `gofuzz` instead of `fuzz` for fuzzing.

fixes #3174

### DIFF
--- a/Makefile.fuzz
+++ b/Makefile.fuzz
@@ -24,13 +24,8 @@ FUZZ:=$(dir $(wildcard plugin/*/fuzz.go)) # plugin/cache/
 PLUGINS:=$(foreach f,$(FUZZ),$(subst plugin, ,$(f:/=))) # > /cache
 PLUGINS:=$(foreach f,$(PLUGINS),$(subst /, ,$(f))) # > cache
 
-.PHONY: vendor
-vendor:
-	@echo "Vendoring code till go-fuzz supports Go modules!"
-	@go mod vendor
-
 .PHONY: echo
-echo: vendor
+echo:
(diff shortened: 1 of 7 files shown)


Makefile.fuzz
```diff
@@ -24,29 +24,34 @@ FUZZ:=$(dir $(wildcard plugin/*/fuzz.go)) # plugin/cache/
...
 PLUGINS:=$(foreach f,$(PLUGINS),$(subst /, ,$(f))) # > cache
 
+.PHONY: vendor
+vendor:
+	@echo "Vendoring code till go-fuzz supports Go modules!"
+	@go mod vendor
+
 .PHONY: echo
-echo:
+echo: vendor
```
miekg: this change break the fuzz build

(I also find this confusing, the sooner gofuzz supports, go modules the better). For now it's probably best to not do this

mfrw: sure .. it was sketchy to me too.. but I put it just as an idea

miekg (approved):

PR #3188, opened 2019-08-25
plugin/forward: add fuzzing

miekg:
Add fuzz.go for forward


miekg:
so this works, but I'm waiting for the *forward* target to show up as a fuzz target

PR #3195, opened 2019-08-26
 return standardized text for ready and health endpoint

xieyanker:
return standardized text for ready and health endpoint


miekg:
the string "OK" is also used in health_test.go

xieyanker:
@miekg done!

miekg:
/lgtm
/merge

If this works

PR #3196, opened 2019-08-26
plugin/file: close reader for reload

### LINKED ISSUE #3191: plugin/auto: Number of handles Too many
I have 31 zone files，but use lsof to view 420 handles

information:
[root@coredns.a1-60-55.pub.unp /data/coredns/conf]lsof | grep '/data/coredns/zone/' |wc -l
420

miekg:
This reloader didn't close the openened file handle. Add a close. Can't
use `defer` because this is in a endless loop.

<strike>Fixes: #3191</strike>


miekg:
not fully certain if this really fixes the issue, but we do need to close this handle

PR #3200, opened 2019-08-26
Fuzz: fix forward build

miekg:
Fix this build - quite hidden in all the output.


miekg:
yeah, travis fixed. For circle we're hitting some kind of limit, unrelated to this.

PR #3208, opened 2019-08-26
plugin/kubernetes: Handle multiple local IPs and bind

chrisohaver:

Handle NS responses for following cases:
* CoreDNS is bound to more than one local address.
* CoreDNS is bound to a specific address/addresses with the bind plugin:  Only use the bound IPs.

example: CoreDNS bound to two IP addresses
```
;; ANSWER SECTION:
cluster.local.		5	IN	NS	ns.dns.cluster.local.

;; ADDITIONAL SECTION:
ns.dns.cluster.local.	5	IN	AAAA	1::2:3:4:5
ns.dns.cluster.local.	5	IN	A	10.84.16.183
```

Previously, we would always only use the _first_ non-loopback address seen on the local system.
This would fail in cases where:
* CoreDNS is using the bind plugin to bind to an IP other than the first non-loopback address. CoreDNS would then answer with an IP that it is not bound to.
* In a Kubernetes cluster, if CoreDNS runs in a dual stack pod, the first IP may not be targeted by a service (or each IP could be targeted by different services), in which case CoreDNS would answer incompletely, or with the Pod IP instead of the Service IP.

#3160
Also related to dual stack support in k8s.

no

no

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -381,14 +381,13 @@ func NS(ctx context.Context, b ServiceBackend, zone string, state request.Reques
 			return nil, nil, fmt.Errorf("NS record must be an IP address: %s", serv.Host)
 
 		case dns.TypeA, dns.TypeAAAA:
+			serv.Host = msg.Domain(serv.Key)
 			extra = append(extra, newAddress(serv, serv.Host, ip, what))
-
 			ns := serv.NewNS(state.QName())
-			if _, ok := seen[ns.Header().Name]; ok {
+			if _, ok := seen[ns.Ns]; ok {
 				continue
 			}
-			serv.Host = msg.Domain(serv.Key)
-			seen[ns.Header().Name] = true
+			seen[ns.Ns] = true
 			records = append(records, ns)
 		}
 	}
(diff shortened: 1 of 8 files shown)


plugin/backend_lookup.go
```diff
@@ -372,6 +372,8 @@ func NS(ctx context.Context, b ServiceBackend, zone string, state request.Reques
 	// ... and reset
 	state.Req.Question[0].Name = old
 
+	seen := map[string]bool{}
```
chrisohaver, line 375: Added `seen` to avoid answer duplication in cases where a single answer has multiple additional records.

plugin/kubernetes/local.go
```diff
@@ -2,41 +2,54 @@ package kubernetes
...
 		if ip6 != nil && !ip6.IsLoopback() {
-			return ip6
+			ips = append(ips, ip6)
 		}
 	}
-	return nil
+	return ips
 }
 
 // LocalNodeName is exclusively used in federation plugin, will be deprecated later.
```
johnbelamaric, line 39: > is this comment true?

At least partly ...

"LocalNodeName is exclusively used in federation plugin": True, only used by `FederationsFunc` in the federation plugin.

"... will be deprecated later": Maybe - depends on what "remove" means in #3041.  I thought it means deprecate/remove... but maybe it just means to move it to an external repo (since thats all that was done, and the issue was closed).

johnbelamaric (commented):
/lgtm

PR #3209, opened 2019-08-26
plugin/kubernetes: Restore k8s.io/client-go to v12.0.0

chrisohaver:

 #2945, (probably unintentionally) bumped the client-go module version _down_ from v12.0.0 to v11.0.0.  This PR restores the client-go version back to v12.0.0.


chrisohaver:
I had to go get the latest stable release (v13.0.0) of azure/go-autorest to get a successful compile.
However, when I do `go mod tidy`, it removes that dependency, and then fails to the import presubmit.

```
** presubmit/import-testing
build github.com/coredns/coredns: cannot load github.com/Azure/go-autorest/autorest: ambiguous import: found github.com/Azure/go-autorest/autorest in multiple modules:
        github.com/Azure/go-autorest v11.1.2+incompatible (/Users/cohaver/go/pkg/mod/github.com/!azure/go-autorest@v11.1.2+incompatible/autorest)
        github.com/Azure/go-autorest/autorest v0.9.0 (/Users/cohaver/go/pkg/mod/github.com/!azure/go-autorest/autorest@v0.9.0)

```
  So, the latest commit is sans-tidy.

PR #3213, opened 2019-08-27
fix mis-spelling in request.go

AllenZMC:
Make the comment more readable


miekg:
See this comment for future fixes: https://github.com/coredns/coredns/pull/3224#issuecomment-526034426

PR #3215, opened 2019-08-27
plugin/etcd: fully move to go.etcd.io

miekg:
This is needed as well to make it compile cleanly. Follow up to #3214.

Manually removed github.com/coreos/etcd/ from go.mod


miekg:
verified that this built with bazel

Issue #3121, opened 2019-08-13
plugin/transfer: New plugin for handling zone transfers


chrisohaver:
As discussed in #2585, the idea here is to implement zone transfers as a separate plugin.
Distilling what was discussed in #2585 ...

Syntax would look like ...
```
transfer [ZONES] {
    to HOSTS
}
```
 **ZONE** - defines a space delimited list of zones the plugin will handle.  If ZONES  is empty, all zones of the serverblock are handled.
**to HOSTS** - defines to which HOSTS transfers can be sent

The plugin...

* Defines interface `Transferer` that plugins must implement to enable zone transfers, comprising ...
   * `Records()` - returns all records
   * `SOA()` - returns SOA record
   * `Authoritative()` - Returns list of zones the plugin is authoritative for
* Is placed before other plugins in plugin.cfg. It would only handles transfer requests, falling through for other query types. 
   * Alternately it could be placed after, but then other plugins would need to fall through for transfer type queries.
* When processing a request, _transfer_ checks for plugins that implement `Transferer`.  Constructs zone transfer response from `Records()` and `SOA()` of first plugin that is `Authoritative()` for the requested zone.
* periodically checks `SOA()` for all plugins in the chain (with `Authoritative()` zones in ZONES), and sends notifies if SOA is changed.

Existing plugins (e.g. *kubernetes*) that already implement zone transfer should be updated to use the new interface, and docs updated.

See #2585 for prior discussion and context.

chrisohaver:
@miekg, @ramnes, @JensErat

chrisohaver:
If this hasn't been started, I can take a whack at it.

This is on my radar because I'd like to add zone transfer to _k8s_external_.

ramnes:
Awesome. I personally did not write a single line of code on this yet. I'd be glad to give my humble review if needed, or any other sort of help.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin/transfer: ..." ]
>As discussed in #2585, the idea here is to implement zone transfers as a separate plugin.
>Distilling what was discussed in #2585 ...
>
>Syntax would look like ...

This looks entirely reasonable.

It only covers *outgoing* transfers? 

Not sure about separating the notify hosts from the transfer to hosts.

chrisohaver:
> It only covers *outgoing* transfers?

[...]

Issue #2585, opened 2019-02-20
plugin/etcd: support zone transfers


ramnes:
As far as I understand, `plugin/etcd` doesn't support zone transfers so far, and that would be a great feature for us, in combination with ExternalDNS plugged to our Kubernetes cluster's CoreDNS.

I don't know much about Go nor AXFR, but if this is something that you guys are interested in, and if no one else does it, I'll eventually try to tackle this feature.

miekg:
This could be done in similar vein as the k8s implements it and should be fairly straight forward.

Note:
* you need some (cheap) way to detect changes in etcd (or use a time based system, where you assume things changes every minute or so)
* you can't get about fetching remote CNAMEs, i.e. everything transferable is in etcd.

miekg:
After this (or during) we can also unify all the code dealing with this, and just make each plugin implement an interface after which they can do zone transfers.

ramnes:
Why not before, so that we can directly use that said interface for the etcd plugin? :)

miekg:
sure; I think you may only need an Records() method that returns the records and a SOA() for the SOA. Still need some code in the plugin to handle to axfr request and send it though.

JensErat:
Same for the dnssec-plugin, which does not support AXFR.

miekg:
but the dnssec plugin doesn't own any data, so there is nothing transfer. Or are you proposing that we sign an axfr?

JensErat:
Yes. I somewhat have the feeling that AXFR should be its own plugin pretty much at the end of the plugin chain that just allows anything else to happen before. And dnssec is just in front -- consider signing Kubernetes (zone transfers), ...

Right now, it's pretty hard running CoreOS as primary DNS server, as there is no reliable zone transfer. In my opinion, a good solution for a CoreDNS operator would be any plugin has a "transfer" interface (like already somewhat proposed in here) and chained plugins can also enrich those transfer zones (like dnssec or any of the others rewriting). So in the end an AXFR provides a complete dump of what CoreOS would respond itself, at least wherever reasonably possible.

miekg:
thought about doing that (i.e. transfer as  a plugin). At the time I couldn't work out how that should work.

ramnes:
To be honest this is exactly what I thought I would see in the codebase before realizing that every plugin was implementing AXFR on its own side. So yeah, I'd love to see an AXFR plugin, that would make CoreDNS much simpler and accessible in my opinion.

[...]

PR #3223, opened 2019-08-28
plugin/transfer: Zone transfer plugin

chrisohaver:

New plugin *transfer* implementing the following parts of #3121
* AXFR
* Simple IXFR with AXFR fallback
* Define allowed transfer to hosts

NOTIFY is not implemented in this PR.

#3121 #2585

Included

### DIFF
--- a/plugin/kubernetes/xfr.go
+++ b/plugin/kubernetes/xfr.go
@@ -110,14 +110,15 @@ func (k *Kubernetes) transfer(c chan dns.RR, zone string) {
 		case api.ServiceTypeClusterIP, api.ServiceTypeNodePort, api.ServiceTypeLoadBalancer:
 			clusterIP := net.ParseIP(svc.ClusterIP)
 			if clusterIP != nil {
+				s := msg.Service{Host: svc.ClusterIP, TTL: k.ttl}
+				s.Key = strings.Join(svcBase, "/")
+
+				// Change host from IP to Name for SRV records
+				host := emitAddressRecord(c, s)
+
 				for _, p := range svc.Ports {
-
-					s := msg.Service{Host: svc.ClusterIP, Port: int(p.Port), TTL: k.ttl}
+					s := msg.Service{Host: host, Port: int(p.Port), TTL: k.ttl}
 					s.Key = strings.Join(svcBase, "/")
-
-					// Change host from IP to Name for SRV records
-					host := emitAddressRecord(c, s)
-					s.Host = host
 
 					// Need to generate this to handle use cases for peer-finder
 					// ref: https://github.com/coredns/coredns/pull/823
--- a/plugin/kubernetes/xfr.go
+++ b/plugin/kubernetes/xfr.go
@@ -110,15 +110,14 @@ func (k *Kubernetes) transfer(c chan dns.RR, zone string) {
 		case api.ServiceTypeClusterIP, api.ServiceTypeNodePort, api.ServiceTypeLoadBalancer:
 			clusterIP := net.ParseIP(svc.ClusterIP)
 			if clusterIP != nil {
-				s := msg.Service{Host: svc.ClusterIP, TTL: k.ttl}
-				s.Key = strings.Join(svcBase, "/")
-
-				// Change host from IP to Name for SRV records
-				host := emitAddressRecord(c, s)
-
 				for _, p := range svc.Ports {
-					s := msg.Service{Host: host, Port: int(p.Port), TTL: k.ttl}
+
+					s := msg.Service{Host: svc.ClusterIP, Port: int(p.Port), TTL: k.ttl}
 					s.Key = strings.Join(svcBase, "/")
+
+					// Change host from IP to Name for SRV records
+					host := emitAddressRecord(c, s)
+					s.Host = host
 
 					// Need to generate this to handle use cases for peer-finder
 					// ref: https://github.com/coredns/coredns/pull/823
--- a/plugin/transfer/README.md
+++ b/plugin/transfer/README.md
@@ -26,7 +26,7 @@ transfer [ZONE...] {
 * **ZONES** The zones *transfer* will answer zone requests for. If left blank,
   the zones are inherited from the enclosing server block. To answer zone 
   transfers for a given zone, there must be another plugin in the same server
-  server block that serves the same zone, and implements `transfer.Transferer`.
+  block that serves the same zone, and implements `transfer.Transferer`.
 
 * `to ` **HOST...** The hosts *transfer* will transfer to. Use `*` to permit
   transfers to all hosts.
--- a/plugin/transfer/README.md
+++ b/plugin/transfer/README.md
@@ -13,9 +13,7 @@ This plugin answers zone transfers for authoritative plugins that implement
 Transfer answers AXFR requests and IXFR requests with AXFR fallback if the
 zone has changed.
 
-Notifies are not supported.
-
-*transfer* is not currently implemented by any plugins.
+Notifies are not currently supported.
 
 ## Syntax
 
@@ -32,49 +30,3 @@ transfer [ZONE...] {
 
 * `to ` **HOST...** The hosts *transfer* will transfer to. Use `*` to permit
   transfers to all hosts.
-
-# Examples
-
-Answer zone transfers for `example.com.` using the theoretical plugin *plugin1*.
-
-~~~
-example.com {
-    transfer {
-      to *
-    }
-    plugin2
-}
-
-~~~
-
-You can enable zone transfer for more than one plugin with a single *transfer* instance.
-Here, *transfer* will send zone transfers to any source for the theoretical plugins *plugin1*
-and *plugin2* (both of which must both implement `transfer.Transferer`).
-
-~~~
-. {
-    transfer a.example.com b.example.com{
-      to *
-    }
-    plugin1 a.example.com
-    plugin2 b.example.com
-}
-
-~~~
-
-Building on the example above, you can use *transfer* more than once in a server block to specify
-separate options per zone.
-
-~~~
-. {
-    transfer a.example.com {
-      to *
-    }
-    transfer b.example.com {
-      to 192.168.0.100
-    }
-    plugin1 a.example.com
-    plugin2 b.example.com
-}
-
-~~~
--- a/plugin/transfer/README.md
+++ b/plugin/transfer/README.md
@@ -33,15 +33,46 @@ transfer [ZONE...] {
 
 # Examples
 
-Answer zone transfers for `my.zone.org.` using the theoretical plugin *myplugin*.
+Answer zone transfers for `example.com.` using the theoretical plugin *plugin1*.
 
 ~~~
-my.zone.org {
+example.com {
     transfer {
       to *
     }
-    myplugin
+    plugin2
 }
 
 ~~~
 
+You can enable zone transfer for more than one plugin with a single *transfer* instance.
+Here, *transfer* will send zone transfers to any source for the theoretical plugins *plugin1* 
+and *plugin2* (both of which must both implement `transfer.Transferer`).
+
+~~~
+. {
+    transfer a.example.com b.example.com{
+      to *
+    }
+    plugin1 a.example.com
+    plugin2 b.example.com
+}
+
+~~~
+
+Building on the example above, you can use *transfer* more than once in a server block to specify
+separate options per zone.
+
+~~~
+. {
+    transfer a.example.com {
+      to *
+    }
+    transfer b.example.com {
+      to 192.168.0.100
+    }
+    plugin1 a.example.com
+    plugin2 b.example.com
+}
+
+~~~
--- a/plugin/transfer/README.md
+++ b/plugin/transfer/README.md
@@ -1,0 +1,47 @@
+# transfer
+
+## Name
+
+*transfer* - answer zone transfers requests for compatible authoritative
+plugins.
+
+## Description
+
+This plugin answers zone transfers for authoritative plugins that implement
+`transfer.Transferer`.
+
+Transfer answers AXFR requests and IXFR requests with AXFR fallback if the
+zone has changed.
+
+*transfer* is not currently implemented by any plugins.
+
+## Syntax
+
+~~~
+transfer [ZONE...] {
+  to HOST...
+}
+~~~
+
+* **ZONES** The zones *transfer* will answer zone requests for. If left blank,
+  the zones are inherited from the enclosing server block. To answer zone 
+  transfers for a given zone, there must be another plugin in the same server
+  server block that serves the same zone, and implements `transfer.Transferer`.
+
+* `to ` **HOST...** The hosts *transfer* will transfer to. Use `*` to permit
+  transfers to all hosts.
+
+# Examples
+
+Answer zone transfers for `my.zone.org.` using the theoretical plugin *myplugin*.
+
+~~~
+my.zone.org {
+    transfer {
+      to *
+    }
+    myplugin
+}
+
+~~~
+
--- a/plugin/transfer/README.md
+++ b/plugin/transfer/README.md
@@ -26,7 +26,7 @@ transfer [ZONE...] {
 ~~~
 
 * **ZONES** The zones *transfer* will answer zone requests for. If left blank,
-  the zones are inherited from the enclosing server block. To answer zone 
+  the zones are inherited from the enclosing server block. To answer zone
   transfers for a given zone, there must be another plugin in the same server
   block that serves the same zone, and implements `transfer.Transferer`.
 
@@ -48,7 +48,7 @@ example.com {
 ~~~
 
 You can enable zone transfer for more than one plugin with a single *transfer* instance.
-Here, *transfer* will send zone transfers to any source for the theoretical plugins *plugin1* 
+Here, *transfer* will send zone transfers to any source for the theoretical plugins *plugin1*
 and *plugin2* (both of which must both implement `transfer.Transferer`).
 
 ~~~
--- a/plugin/transfer/README.md
+++ b/plugin/transfer/README.md
@@ -13,6 +13,8 @@ This plugin answers zone transfers for authoritative plugins that implement
 Transfer answers AXFR requests and IXFR requests with AXFR fallback if the
 zone has changed.
 
+Notifies are not supported.
+
 *transfer* is not currently implemented by any plugins.
 
 ## Syntax
--- a/plugin/transfer/setup.go
+++ b/plugin/transfer/setup.go
@@ -63,7 +63,7 @@ func parse(c *caddy.Controller) (*Transfer, error) {
 			}
 		}
 
-	BLOCK:
+	block:
 		for c.NextBlock() {
 			switch c.Val() {
 			case "to":
@@ -74,7 +74,7 @@ func parse(c *caddy.Controller) (*Transfer, error) {
 				for _, host := range args {
 					if host == "*" {
 						x.to = []string{"*"}
-						continue BLOCK
+						continue block
 					}
 					ip := net.ParseIP(host)
 					if ip == nil {
--- a/plugin/transfer/setup.go
+++ b/plugin/transfer/setup.go
@@ -29,35 +29,16 @@ func setup(c *caddy.Controller) error {
 	})
 
 	c.OnStartup(func() error {
-		// find all plugins that implement Transferer and add them to t.Transferers
+		// find all plugins that implement Transferer and add them to Transferers
 		plugins := dnsserver.GetConfig(c).Handlers()
 		for _, pl := range plugins {
 			tr, ok := pl.(Transferer)
 			if !ok {
 				continue
 			}
-			for _, z := range tr.Authoritative() {
-				for _, x := range t.xfrs {
-					var found bool
-					for _, xz := range x.Zones {
-						if z == xz {
-							found = true
-							break
-						}
-					}
-					if !found {
-						continue
-					}
-					if x.Transferers[z] != nil {
-						// Acknowledge only the first transferer for each zone.
-						// This doesn't play nice with fallthrough, where more than one plugin can
-						// collectively be authoritative for a zone.
-						continue
-					}
-					x.Transferers[z] = tr
-				}
+			for _, x := range t.xfrs {
+				x.Transferers = append(x.Transferers, tr)
 			}
-
 		}
 		return nil
 	})
@@ -110,8 +91,6 @@ func parse(c *caddy.Controller) (*Transfer, error) {
 		if len(x.to) == 0 {
 			return nil, plugin.Error("transfer", c.Errf("'to' is required", c.Val()))
 		}
-		x.Transferers = make(map[string]Transferer)
-		t.xfrs = append(t.xfrs, x)
 	}
 	return t, nil
 }
--- a/plugin/transfer/setup.go
+++ b/plugin/transfer/setup.go
@@ -54,12 +54,20 @@ func parse(c *caddy.Controller) (*Transfer, error) {
 		if len(zones) != 0 {
 			x.Zones = zones
 			for i := 0; i < len(x.Zones); i++ {
-				x.Zones[i] = plugin.Host(x.Zones[i]).Normalize()
+				nzone, err := plugin.Host(x.Zones[i]).MustNormalize()
+				if err != nil {
+					return nil, err
+				}
+				x.Zones[i] = nzone
 			}
 		} else {
 			x.Zones = make([]string, len(c.ServerBlockKeys))
 			for i := 0; i < len(c.ServerBlockKeys); i++ {
-				x.Zones[i] = plugin.Host(c.ServerBlockKeys[i]).Normalize()
+				nzone, err := plugin.Host(c.ServerBlockKeys[i]).MustNormalize()
+				if err != nil {
+					return nil, err
+				}
+				x.Zones[i] = nzone
 			}
 		}
 
--- a/plugin/transfer/setup.go
+++ b/plugin/transfer/setup.go
@@ -1,10 +1,10 @@
 package transfer
 
 import (
-	"net"
-
 	"github.com/coredns/coredns/core/dnsserver"
 	"github.com/coredns/coredns/plugin"
+	parsepkg "github.com/coredns/coredns/plugin/pkg/parse"
+	"github.com/coredns/coredns/plugin/pkg/transport"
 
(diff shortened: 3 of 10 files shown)


chrisohaver:
Status:
 * Basic AXFR structure in place.
 * Simple IXFR handling (fallback AXFR)

Todo:
 * Notify (in a following PR)

chrisohaver:
I copied chunks of the transfer code from the file implementation. I see that #3227 modifies/reworks some of that.  I'll pull those changes in here when it merges.

chrisohaver:
79d7b10 implements `transfer.Transferer` for *k8s_external*.  

This one was a bit tricky, as I had to extend the `external.Externaler` interface.  The result is a bit twisty. IMO, the code would be less verbose if we just allow *k8s_external* to use the kubernetes api connection directly instead of using a list of functions in the interface. That would make `external.Externaler` only implementable by plugins that have a k8s api connection... but thats a reasonable assumption.

plugin/transfer/transfer.go
```diff
@@ -0,0 +1,144 @@
...
+	Zones       []string
+	to          []string
+	Transferers map[string]Transferer // a map of zone->plugins which implement Transferer
+}
+
+// Transferer may be implemented by plugins to enable zone transfers
+type Transferer interface {
+	// AllRecords must return the SOA and all other records in the zone, including NS records
+	// for the given zone.
+	AllRecords(zone string) (*dns.SOA, []dns.RR, error)
```
miekg: This is, I think, not the correct interface. It needs to be more like an iterator.

```
type Transferer interface {
   Walk(zone string, serial uint32) ([]dns.RR, error)
}
~~~
Or something like that, see latest tweaks in `plugin/file/xfr.go`
Two things that:
1. First RR returned must be a SOA
2. `io.EOF` signals the end of the list

File: plugin/transfer/transfer.go [hunk 2]
```diff
@@ -0,0 +1,144 @@
...
+
+	// SoaSerial must return the current serial for the given zone.
+	SoaSerial(zone string) (uint32, error)
```
miekg: Probably can be folded into `Walk` as I did above

plugin/transfer/transfer.go
johnbelamaric: What about returning a channel?

plugin/transfer/setup.go
```diff
@@ -0,0 +1,114 @@
...
+								found = true
+								break
+							}
+						}
+						if !found {
+							continue
+						}
+						if x.Transferers[z] != nil {
+							// Acknowledge only the first transferer for each zone.
+							// This doesn't play nice with fallthrough, where more than one plugin can
```
johnbelamaric: Would be nice to see if a later version could do something for this case.

plugin/transfer/transfer.go
chrisohaver: Ah, a channel, so we can start responding to the client in parallel with the plugin generating the records.  Yeah, makes sense.

plugin/transfer/setup.go
chrisohaver: Yes, I thought about this and decided it was complicated enough to leave out for now, to keep this PR small and easily reviewable.  I think it's solvable.  The transfer plugin should be able to determine the "order" of plugins and merge them down with the records of other plugins later in the chain, preferring records from earlier plugins when there are collisions.  Multi instance plugins would need be handled by each plugin's implementation of transfer.

chrisohaver:
I've reworked the interface, and I have adjusted the _k8s_external_ and _kubernetes_ implementations.

I'm inclined to just implement the interface for kubernetes/k8s_external in this PR, other plugins to be migrated in separate PRs.

plugin/kubernetes/controller.go
```diff
@@ -43,13 +43,20 @@ type dnsController interface {
...
 
 type dnsControl struct {
-	// Modified tracks timestamp of the most recent changes
+	// modified tracks timestamp of the most recent changes
 	// It needs to be first because it is guaranteed to be 8-byte
 	// aligned ( we use sync.LoadAtomic with this )
 	modified int64
+	// extModified tracks timestamp of the most recent changes to
+	// services with external facing IP addresses
+    extModified int64
```
johnbelamaric: go fmt needed?

plugin/kubernetes/controller.go
```diff
@@ -423,6 +430,9 @@ func (dns *dnsControl) detectChanges(oldObj, newObj interface{}) {
 	switch ob := obj.(type) {
 	case *object.Service:
 		dns.updateModifed()
+		if len(ob.ExternalIPs) > 0 {
```
johnbelamaric: can we check if the actual external IPs have changed?

plugin/kubernetes/xfr.go
```diff
@@ -31,7 +31,7 @@ func (k *Kubernetes) Transfer(ctx context.Context, state request.Request) (int,
 
 	// Get all services.
 	rrs := make(chan dns.RR)
-	go k.transfer(rrs, state.Zone)
+	go k.transferAndClose(rrs, state.Zone)
```
johnbelamaric: So, after this merges, can we deprecate the transfer options within k8s, and move them all to transfer? Can we use transfer more than once with different zones so we can control the "allowed" IPs on a per-zone basis?

plugin/kubernetes/controller.go
chrisohaver: Yeah - we also dont check if other Service attributes have changed... e.g.
* externalName, which would change CNAME records
* defined ports, which would change SRV records

ClusterIP is immutable, so don't need to check that...

plugin/kubernetes/xfr.go
chrisohaver: Yes to both

plugin/kubernetes/controller.go
chrisohaver: Added checks in f430772.  I tested manually, with a local cluster.
We don't have good unit test coverage in the controller.

plugin/kubernetes/controller.go
```diff
@@ -510,6 +519,74 @@ func endpointsEquivalent(a, b *object.Endpoints) bool {
...
+		return true
+	}
+	for i := range oldSvc.ExternalIPs {
+		if oldSvc.ExternalIPs[i] != newSvc.ExternalIPs[i] {
+			return true
+		}
+	}
+
+	// All Port fields are mutable
+	for i := range oldSvc.Ports {
```
johnbelamaric: you forgot to check the len on the ports this time, so if newPorts > oldPorts this can return a false negative.

miekg:
I would prefer the entire xfr impl of k8s_external to be in another PR (I get the usefulness of testing your transfer impl, but it muddles this pr)

plugin/transfer/transfer.go
```diff
@@ -0,0 +1,135 @@
...
+	to          []string
+	Transferers map[string]Transferer // a map of zone->plugins which implement Transferer
+}
+
+// Transferer may be implemented by plugins to enable zone transfers
+type Transferer interface {
+	// AllRecords writes all records in the zone to the channel.  The
+	// first and last records written to the channel must be the SOA for the zone.
+	// If serial is non zero, an incremental zone transfer should be assumed.
+	AllRecords(zone string, serial uint32, ch chan dns.RR)
```
miekg: I don't like this interface yet. I think returning a channel is more natural, also the channel should be recieve only, not the send/receive it is now.
I.e. 
~~~
Transfer(zone string, serial uint32) (ch <-chan []dns.RR)
~~~
This also means the plugin closes the channel after is sent the last rrrs, which is more natural, not sure how that would work now.
And allowing a slice is also way more efficient, because that's probably what most plugins are holding anyway. Also sending the SOA twice seems axfr implemention detail and not super relevant to the implementation. The transfer plugin can take care of that.

plugin/transfer/transfer.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+	return false
+}
+
+// Name implements the Handler interface.
+func (Transfer) Name() string { return pluginName }
+
+const (
+	pluginName = "transfer"
```
miekg: meh, please just use "transfer" the few times it's need in the code.

plugin/transfer/transfer.go
```diff
@@ -0,0 +1,135 @@
...
+	// Authoritative must return a list of Zones that the plugin is authoritative for.
+	Authoritative() []string
+}
+
+// ServeDNS implements the plugin.Handler interface.
+func (t Transfer) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	state := request.Request{W: w, Req: r}
+
+	var serial uint32
+	if state.QType() == dns.TypeIXFR {
```
miekg, line 81: this needs the ixfr handling that was recently added in file's tranfer

plugin/transfer/transfer.go
```diff
@@ -0,0 +1,135 @@
...
+		soa, ok := r.Ns[0].(*dns.SOA)
+		if !ok {
+			return dns.RcodeServerFailure, nil
+		}
+		serial = soa.Serial
+	} else if state.QType() != dns.TypeAXFR {
```
miekg: `switch` is probably naturally

plugin/kubernetes/controller.go
chrisohaver: oops. good catch.

plugin/transfer/transfer.go
chrisohaver: I am currently handling this in the plugin implementations, to allow them to implement non-axfr-fallback if they could.  But implementing non-axfr-fallback is probably too much to expect of a plugin, and most plugins probably woudn't bother (the full ixfr answer protocol being a tad fiddly).  So, yeah , makes more practical sense to move the simple ixfr handling to the transfer plugin, and later we can extend the transfer plugin, perhaps with an IncrTransferer interface down the road.

chrisohaver:
@miekg, I think I got all of your feedback thus far rolled in...

* remove _kubernetes_ and _k8s_external_ implementations from this PR
* use RR slice in the interface channel. create and close the channel in sender.
* use literal instead of constant for plugin name
* move simple ixfr response logic from plugin implementations to transfer plugin
* use a `switch` instead of `if` in ServeDNS

Let me know if you'd like me to squash the commits.

miekg:
move the k8s transfer impl to a new pr.

Allrecords is oddly named and doesnt allow for incremental transfers to be
added, unless you go with some super weird naming scheme. SOA should not be
handled separately. And I don't think authoritative is needed as a method, you can just call Transfer and see what you get.
Probably best to let Transfer also return an predefined error or DNS rcode, to signal this

chrisohaver:
> move the k8s transfer impl to a new pr.

@miekg, It's already removed from this PR ...  Do you want me to open a new PR in parallel before this one is finalized?

chrisohaver:
@miekg, 

renamed AllRecords() -> Transfer() (per your earlier suggestion)
removed SOA() and Authoritative()

johnbelamaric:
There some k8s change in there still - looks like maybe a bug fix?

chrisohaver:
> There some k8s change in there still - looks like maybe a bug fix?

Sorry - i must have added it back in recently by accident ... trying to juggle the interdependent branches is challenging...

i'll rebase again...

chrisohaver:
> Sorry - i must have added it back in recently by accident ... trying to juggle the interdependent branches is challenging...

No, it's something i missed when I untangled this PR from the k8s change.  Sorry about that.

chrisohaver:
Actually, you are right - it was a bug fix... that was already merged in #3240 ...
not sure why it's coming up in my PR, my master is up to date, and I have rebased my branch with master ...

Maybe my local master branch is tainted.  I slipped off the git tightrope again.

Edit: Oh! Its the opposite of that bug fix... i.e. it's reverting the bug fix.  Not sure how I managed that...

chrisohaver:
I have reconciled that k8s change with master.
Also added unit tests.

johnbelamaric:
git is sharp and sometimes perplexing...

plugin/transfer/README.md
```diff
@@ -0,0 +1,47 @@
...
+~~~
+transfer [ZONE...] {
+  to HOST...
+}
+~~~
+
+* **ZONES** The zones *transfer* will answer zone requests for. If left blank,
+  the zones are inherited from the enclosing server block. To answer zone 
+  transfers for a given zone, there must be another plugin in the same server
+  server block that serves the same zone, and implements `transfer.Transferer`.
```
johnbelamaric: server is repeated here

plugin/transfer/transfer.go
```diff
@@ -0,0 +1,176 @@
...
+// Transfer is a plugin that handles zone transfers.
+type Transfer struct {
+	xfrs []*xfr
+	Next plugin.Handler
+}
+
+type xfr struct {
+	Zones       []string
+	to          []string
+	Transferers []Transferer // the list of plugins that implement Transferer
```
johnbelamaric: Why is this here instead of in the `Transfer` struct? In the setup you just add them all to every one anyway.

chrisohaver: ah, yes - it was there because it used to be a map of zones to Transferers ... only the zones for that instance were in it.  It can be moved to the main struct now.

chrisohaver: The list of Transferers is not likely to be large, and the load of transfer requests not likely to be high.  So the map optimization was not really needed.  Code is simpler without it.

chrisohaver:
@miekg, @johnbelamaric 

Would you guys like to see **new PRs** for *kubernetes* and *k8s_external* plugin implementations of `transfer.Transfer` now or wait until after this PR is merged (if it merges)?

plugin/transfer/README.md
```diff
@@ -0,0 +1,80 @@
...
+  block that serves the same zone, and implements `transfer.Transferer`.
+
+* `to ` **HOST...** The hosts *transfer* will transfer to. Use `*` to permit
+  transfers to all hosts.
+
+# Examples
+
+Answer zone transfers for `example.com.` using the theoretical plugin *plugin1*.
+
+~~~
```
miekg: this can use `corefile` is the language, so it's tested `plugin2` must be then renamed to an existing plugin.

plugin/transfer/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Description
+
+This plugin answers zone transfers for authoritative plugins that implement
+`transfer.Transferer`.
+
+Transfer answers AXFR requests and IXFR requests with AXFR fallback if the
+zone has changed.
+
+Notifies are not supported.
```
miekg: meaning... a plugin still needs to send those by itself? How does it know where to send them ?

plugin/transfer/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+*transfer* is not currently implemented by any plugins.
```
miekg: can be removed

plugin/transfer/README.md
```diff
@@ -0,0 +1,80 @@
...
+    plugin2 b.example.com
+}
+
+~~~
+
+Building on the example above, you can use *transfer* more than once in a server block to specify
+separate options per zone.
+
+~~~
+. {
```
miekg: this is not recommended, split these out on zone boundaries in diff. server blocks, the example itself also doesn't add much to the previous examples.

plugin/transfer/setup.go
```diff
@@ -0,0 +1,95 @@
...
+				x.Zones[i] = plugin.Host(x.Zones[i]).Normalize()
+			}
+		} else {
+			x.Zones = make([]string, len(c.ServerBlockKeys))
+			for i := 0; i < len(c.ServerBlockKeys); i++ {
+				x.Zones[i] = plugin.Host(c.ServerBlockKeys[i]).Normalize()
+			}
+		}
+
+	BLOCK:
```
miekg: s/BLOCK/block/

plugin/transfer/transfer.go
miekg: Yep, that's was my thinking why `AllRecords` would be a bad name for the functions in the interfaces. Now we can extend this and keep a similar naming scheme

plugin/transfer/transfer.go
```diff
@@ -0,0 +1,178 @@
...
+	return plugin.NextOrFailure(t.Name(), t.Next, ctx, w, r)
+
+}
+
+// send sends the response to the client.  It aborts if the first record received on the
+// channel is not an SOA.  send automatically adds the closing SOA to the end of an AXFR response.
+func (x xfr) send(state request.Request, fromPlugin <-chan []dns.RR) {
+	toRemote := make(chan *dns.Envelope)
+
+	go func() {
```
miekg: why is this done in a go-routine?

plugin/transfer/transfer.go
```diff
@@ -0,0 +1,178 @@
...
+		var serial uint32
+		var soa *dns.SOA
+		l, c := 0, 0
+		for rs := range fromPlugin {
+			for _, r := range rs {
+				if c == 0 {
+					var ok bool
+					soa, ok = r.(*dns.SOA)
+					if !ok {
+						log.Errorf("First record in transfer from zone %s was not an SOA", state.QName())
```
miekg: return the error?

plugin/transfer/transfer.go
```diff
@@ -0,0 +1,178 @@
...
+			toRemote <- &dns.Envelope{RR: records}
+		}
+
+		close(toRemote)
+		log.Infof("Outgoing transfer of %d records of zone %s to %s with %d SOA serial", c, state.QName(), state.IP(), serial)
+	}()
+
+	tr := new(dns.Transfer)
+	tr.Out(state.W, state.Req, toRemote)
+	state.W.Hijack()
```
miekg: hijack was removed from secondary because it's not thread safe, you're reintroducing it here.

plugin/transfer/transfer.go
```diff
@@ -0,0 +1,178 @@
...
+		return dns.RcodeSuccess, nil
+	}
+
+	return plugin.NextOrFailure(t.Name(), t.Next, ctx, w, r)
+
+}
+
+// send sends the response to the client.  It aborts if the first record received on the
+// channel is not an SOA.  send automatically adds the closing SOA to the end of an AXFR response.
+func (x xfr) send(state request.Request, fromPlugin <-chan []dns.RR) {
```
miekg: the code is file/xfr.go is remarkbly shorter than this; why not use that?

plugin/transfer/transfer.go
```diff
@@ -0,0 +1,178 @@
...
+	toRemote := make(chan *dns.Envelope)
+
+	go func() {
+		var records []dns.RR
+		var serial uint32
+		var soa *dns.SOA
+		l, c := 0, 0
+		for rs := range fromPlugin {
+			for _, r := range rs {
+				if c == 0 {
```
miekg: why do you introduce counters, and discard the counters from the `range`??

miekg (changes requested):
interface looks good, but I have a question about sending notifies

plugin/transfer/README.md
chrisohaver: I'll remove the examples for now, and add them back once when there are some plugins that implement transfer exist.

I'm hoping that this version of the README, where there are no plugins that implement transfer, will be short lived.

plugin/transfer/README.md
chrisohaver: It's not implemented yet.  I was hoping to implement notify in a separate PR, to keep the PR short.  Is that OK, or do you want notifies to be handled in this PR too?

plugin/transfer/transfer.go
chrisohaver: OK, the latest commit copies the code in from *file*, adapted to work with the channel returned by the plugin.

One functional difference I see is that *file*'s xfr implementation uses record count to cut/send envelopes, whereas previously I was using record data length. Not sure which one is more correct.

Another is that the code from *file* always sends the closing SOA in a separate envelope containing just that one record. Previously the closing SOA was included as the last record in the final envelope.

Not sure if these are significant.

plugin/transfer/transfer.go
chrisohaver: removed

plugin/transfer/transfer.go
chrisohaver: returns SERVFAIL with error now

plugin/transfer/transfer.go
chrisohaver: code replaced/refactored

plugin/transfer/transfer.go
chrisohaver: code replaced/refactored.

plugin/transfer/README.md
johnbelamaric: What are you thinking here? How will `transfer` know the zone has changed? Periodic polling?

We could bring back the `watch` infrastructure, which allowed plugins to push an event down a channel ... though it would likely need some refactoring to work in this internal manner (`transfer` subscribing to updates of zones from other plugins, rather than clients subscribing to updates of names).

plugin/transfer/README.md
johnbelamaric: I cannot really recommend always splitting in different server blocks until we fix the problems discussed in #3039. For any plugin that uses significant resources, duplicating it by having many server blocks is bad.

plugin/transfer/README.md
chrisohaver: I was thinking polling because it seems to me like a simpler/cleaner solution.

In a polling design, a plugin would implement a `Serial(zone) serial` interface, which would be trivial to implement - just returning the current serial for the zone. Transfer would periodically poll each plugin in Tranferers for each zone, then send the notifies to listed hosts if serials change (as compared to a state maintained by the transfer plugin).

In a channel push design, a plugin would need to add code to their existing areas that update their serial.  This would not be as trivial to implement from the plugin perspective (code spread out a bit more).  There may be a cleaner way to do a push design that I'm not seeing right now.

Channel push would get notifies sent more timely, but how timely are notifies expected to be? It could also be more spammy, if thats a concern.

johnbelamaric: Yeah, that's a reasonable argument.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/transf..." ]
>+~~~
>+. {
>
>I cannot really recommend always splitting in different server blocks until we fix the problems discussed in #3039. For any plugin that uses significant resources, duplicating it by having many server blocks is bad.

Hooking it in "." is wrong. That this is still done in coredns/deployment for 
k8s is dumb as well.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/transf..." ]
>> +# transfer
>+
>+## Name
>+
>+*transfer* - answer zone transfers requests for compatible authoritative
>+plugins.
>+
>+## Description
>+
>+This plugin answers zone transfers for authoritative plugins that implement
>+`transfer.Transferer`.
>+
>+Transfer answers AXFR requests and IXFR requests with AXFR fallback if the
>+zone has changed.
>+
>+Notifies are not supported.
>
>Yeah, that's a reasonable argument.

so say, the *auto* plugin loads 30k small zones, how do we make that work? That 
needs thought. Dealing with notifies needs to fleshed out in this PR and 
possible a small doc needs to be written on how this will work.

plugin/transfer/transfer.go
chrisohaver: > Not sure if these are significant.

Are they? (i.e. are these bugs in the *file* plugin's implementation of zone transfers, that I have now replicated here)

chrisohaver:
> Hooking it in "." is wrong. That this is still done in coredns/deployment for
k8s is dumb as well.

Please see this discussion in coredns/deployment#153. Thanks.

chrisohaver:
> so say, the *auto* plugin loads 30k small zones,

OK, to better support cases like that we could use channels instead.

Fleshing it out a bit, one way to do this ...

1. Create a _Notifier_ interface*, that contains one func, `Notify() <-chan  ZoneSerial`
which returns a channel to which the implementing plugin will send a `struct{Zone; Serial}` to the channel when a zone's serial changes.
2. In the On Startup caddy event, _transfer_ gets a channel from each plugin that implements Notify, and spins up listeners.
3. In each plugin, whenever the serial changes, sends an update to the channel.
4. When _transfer_ receives a message, it sends a NOTIFY to each static IP in the `to` option for the zone.

*If preferable, we could add `Notify()` to the `Transferer` interface, and permit plugins to return a nil channel (if they opt not to actually implement notify function).

chrisohaver:
Gentle ping.

chrisohaver:
@miekg, @johnbelamaric, please review my proposal for how we could add Notify support [above](https://github.com/coredns/coredns/pull/3223#issuecomment-534099365).

johnbelamaric:
looks reasonable to me

chrisohaver:
Thanks @johnbelamaric.  @miekg, is the notify concept [above](https://github.com/coredns/coredns/pull/3223#issuecomment-534099365) fleshed out enough, or would you like to see something more/different before merging this PR.

chrisohaver:
FWIW - I've got notify + k8s_external working in locally in a separate set of branches - tested with a NIOS dns grid member as the secondary.  I added `Notify` to the existing `Transferer` interface, which I think is cleaner than to create a new interface for it. One change from the proposed solution [above](https://github.com/coredns/coredns/pull/3223#issuecomment-534099365) is that the message only contains zone (serial not needed when sending a notify).

chrisohaver:
OK to merge?

plugin/transfer/transfer.go
```diff
@@ -0,0 +1,181 @@
...
+type Transferer interface {
+	// Transfer returns a channel to which it writes responses to the transfer request.
+	// If the plugin is not authoritative for the zone, it should immediately return the
+	// Transfer.ErrNotAuthoritative error.
+	//
+	// If serial is 0, handle as an AXFR request. Transfer should send all records
+	// in the zone to the channel. The SOA should be written to the channel first, followed
+	// by all other records, including all NS + glue records.
+	//
+	// If serial is not 0, handle as an IXFR request. If the serial is >
```
johnbelamaric: Was that supposed to be `<`?

chrisohaver: ~yep thanks!~  actually it's supposed to be >=, and the next comparison <.  I got them swapped, thanks for drawing my attention to it.

chrisohaver:
@miekg, I think all comments have been addressed and this is ready to go.  I'd like to merge this shortly unless you have any objections.

PR #3224, opened 2019-08-29
fix-up grammar-mistake

tanjunchen:
fix-up the the of  plugin/kubernetes/ns.go  document
None
None
None


miekg:
Thanks for these fixes, but we have seen quite a bunch of these PRs. I'm going to stop merging them unless they are batched up and fix at least a number of spelling mistake (like 10 to 20) in one go. The overhead is just not worth it.

thank you

PR #3227, opened 2019-08-29
plugin/file: rework outgoing axfr

miekg:


test/auto_test.go
```diff
@@ -142,14 +142,17 @@ func TestAutoAXFR(t *testing.T) {
...
-	resp, err := dns.Exchange(m, udp)
+	c, err := tr.In(m, tcp)
 	if err != nil {
 		t.Fatal("Expected to receive reply, but didn't")
 	}
-	if len(resp.Answer) != 5 {
-		t.Fatalf("Expected response with %d RRs, got %d", 5, len(resp.Answer))
+	l := len(c)
+
+	if l == 5 {
```
stp-ip: Shouldn't this be `!=`

stp-ip (commented):
/lgtm

One change request within the test. Rest looks good.

test/auto_test.go
miekg: yep, good catch!

Issue #3225, opened 2019-08-29
ascii logo


miekg:
I also wanted to print a small logo when coredns is starting up. Nothing too fancy, couple of lines and easily created (with `figlet` or some such).

After playing a tiny amount of time with I have some candidates I like.

~~~
% figlet -f digital CoreDNS
+-+-+-+-+-+-+-+
|C|o|r|e|D|N|S|
+-+-+-+-+-+-+-+
~~~
This can be amended slightly to have a more RFC protocol feel to it.

old style ascii:
~~~
% figlet -f slant CoreDNS
   ______                ____  _   _______
  / ____/___  ________  / __ \/ | / / ___/
 / /   / __ \/ ___/ _ \/ / / /  |/ /\__ \ 
/ /___/ /_/ / /  /  __/ /_/ / /|  /___/ / 
\____/\____/_/   \___/_____/_/ |_//____/  
~~~

Something along these lines,  bonus points for making it look like the actual logo with the styled chain on the left

miekg:
my terminal displays: https://www.compart.com/en/unicode/U+26D3
~~~
⛓CoreDNS
~~~

chrisohaver:
I like the retro slanted one.  But it takes up a lot of space.

The chain (⛓) one is short and sweet.

miekg:
figlet eats that unicode codepoint. I may try to "improve on it" and see what it leads. Afraid if we don't use plain ascii it will looks weird on most consoles

johnbelamaric:
+1 to:

% figlet -f slant CoreDNS
   ______                ____  _   _______
  / ____/___  ________  / __ \/ | / / ___/
 / /   / __ \/ ___/ _ \/ / / /  |/ /\__ \
/ /___/ /_/ / /  /  __/ /_/ / /|  /___/ /
\____/\____/_/   \___/_____/_/ |_//____/

PR #3230, opened 2019-08-30
startup: add logo

miekg:
As discussed in #3225, lets add a little logo. This PR incorperates this
in the startup text. It also removes the logging output as this outputs
identical lines of text, but for no clear reason. This means --quiet now
means "no output on startup at all".

Currently it looks like this:

~~~
.:1053
   ______                ____  _   _______
  / ____/___  ________  / __ \/ | / / ___/	~ CoreDNS-1.6.2
 / /   / __ \/ ___/ _ \/ / / /  |/ /\__ \ 	~ linux/amd64, go1.12.9,
/ /___/ /_/ / /  /  __/ /_/ / /|  /___/ /
\____/\____/_/   \___/_____/_/ |_//____/
~~~

We have 2 lines extra on the right if we need to print more (no ideas
come to mind currently).


thaJeztah:
perhaps put the version information on their own line after the logo, for easier `grep`-ing and for small terminals?

(perhaps put the logo on stderr to allow redirecting it)

miekg:
Quite like the way it looks now (otherwise I wouldn't have made it like
this). Can potentially add a pipe symbol before the version string as an
anchoring point?

Your terminal needs to be very narrow for this to look bad though

chrisohaver (approved):

mrueg:
ASCII logos could cause issues and difficulties in terms of accessibility. Has this been tested on screenreaders / other devices?

towolf:
Redis does this:
```
# By default Redis shows an ASCII art logo only when started to log to the
# standard output and if the standard output is a TTY. Basically this means
# that normally a logo is displayed only in interactive sessions.
#
# However it is possible to force the pre-4.0 behavior and always show a
# ASCII art logo in startup logs by setting the following option to yes.
always-show-logo yes
```

Why? Because multi-line logos look shitty in logging systems where one line is taken to be one message.

Issue #3235, opened 2019-08-31
Using docker to build coredns and consequently the docker image fails due to permissions.


orangelynx:
Hi,

when using a docker container of image `golang` to build the tarball (1.6.3 in my case), it creates files `go.mod` and `go.sum` during the build process inside the container as root user. If the source is mounted as a host bind volume (as instructed in the coredns main readme), the files will also be owned by root on the main file system.

Consequently, when using `docker build coredns-1.6.3` (specifying `coredns-1.6.3` as build context) as normal (non-root user), the command will fail when trying to add `go.sum` and `go.mod` to the buildcontext archive, as they are readonly by root.

this can of course be easily fixed locally, but ideally, these files should simply be added to .dockerignore so they will not be added to the build context (where they are not needed anyway).

PR #3236, opened 2019-08-31
Fix for 3235. Adds go.sum and go.mod to .dockerignore to avoid permis…

orangelynx:
…sion issues during docker build process.

Adds two files to .dockerignore that may be owned by root if the coredns binary was built in a golang docker container with root permissions and the docker container based on that build is built from a non-root context. See #3235 for details.

#3235 

None. Perhaps for clarity, if similar issues occur in the future, reference this issue in the documentation on docker build process.

No

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -10,3 +10,5 @@ test/*
 vendor/*
 build/*
 release/*
+go.sum
+go.mod


yongtang (approved):
LGTM

PR #3239, opened 2019-09-04
plugin/acl: whitespace cleanup

miekg:
OWNERS file was padded, remove empty lines. Fix code block in README,
and set codeblock type to 'corefile' without the upper case C.


ihac:
Thanks for fix. LGTM @miekg

PR #3242, opened 2019-09-05
Update README.md

cricketliu:
Minor textual cleanup.

PR #3254, opened 2019-09-05
Making README text consistent with other plugins' READMEs

cricketliu:


miekg:
The DCO will stay (see https://github.com/coredns/coredns/pull/3242#issuecomment-528616110), so you need to sign off all these commit: `commit -s` or `commit -s --amend`

bradbeam (approved):

Issue #3259, opened 2019-09-07
Compilation fails on a fresh install with Go 1.13


kasisnu:
Hi,

A fresh clone followed by `make` with Go `1.13` fails because a dependency on `golang.org/x/net` via `go.etcd.io/etcd` fails version validation checks.

https://tip.golang.org/doc/go1.13#version-validation

```sh
bash-3.2$ pwd
/Users/kasisnu.singh/go/src/github.com/coredns/coredns

bash-3.2$ git rev-parse HEAD
e4df752b17a908c0cef9d3f91cb45856466a08dc

bash-3.2$ go version
go version go1.13 darwin/amd64

bash-3.2$ make
** presubmit/context
** presubmit/filename-hyphen
** presubmit/import-testing
go: go.etcd.io/etcd@v0.0.0-20190823073701-67d0c21bb04c requires
        golang.org/x/net@v0.0.0-20190813000000-74dc4d7220e7: invalid pseudo-version: does not match version-control timestamp (2019-08-13T14:13:03Z)
** presubmit/test-lowercase
** presubmit/trailing-whitespace
GO111MODULE=on CGO_ENABLED=0  go build -v -ldflags="-s -w -X github.com/coredns/coredns/coremain.GitCommit=e4df752b" -o coredns
get "golang.org/x/net": found meta tag get.metaImport{Prefix:"golang.org/x/net", VCS:"git", RepoRoot:"https://go.googlesource.com/net"} at //golang.org/x/net?go-get=1
go: go.etcd.io/etcd@v0.0.0-20190823073701-67d0c21bb04c requires
        golang.org/x/net@v0.0.0-20190813000000-74dc4d7220e7: invalid pseudo-version: does not match version-control timestamp (2019-08-13T14:13:03Z)
make: *** [coredns] Error 1

bash-3.2$
```

Current documentation mentions that any version above 1.12 should work.

yongtang:
The issue has been resolved in #3260, thanks for the fix!

miekg:
the pinning still causes problems https://github.com/coredns/unbound/pull/24

miekg:
still broken, in spite of the pin

miekg:
fix the unbound repo by remove the go.mod and go.sum files. Plugins can't have these files for them to work... (go modules for you)

PR #3260, opened 2019-09-07
Pin golang.org/x/net in go.mod

kasisnu:
This PR pins an indirect dependency on `golang.org/x/net` that doesn't meet version validation checks introduced/enforced in `1.13`

https://github.com/coredns/coredns/issues/3259

I don't think so. Verified this patch with Go `1.12.9` and `1.13`.

```sh
bash-3.2$ go version
go version go1.13 darwin/amd64
bash-3.2$ make
** presubmit/context
** presubmit/filename-hyphen
** presubmit/import-testing
** presubmit/test-lowercase
** presubmit/trailing-whitespace
GO111MODULE=on CGO_ENABLED=0  go build -v -ldflags="-s -w -X github.com/coredns/coredns/coremain.GitCommit=9e363679" -o coredns
bash-3.2$```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -60,4 +60,5 @@ require (
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.0.0+incompatible
 	github.com/miekg/dns v1.1.3 => github.com/miekg/dns v1.1.16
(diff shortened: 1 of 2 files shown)


yongtang (approved):
LGTM. I also tried updating etcd but looks like have to wait for etcd update.

PR #3269, opened 2019-09-09
default config: add log

miekg:
When there is no Corefile found we load the default. Add the log plugin
to it, so you can see queries actually landing in CoreDNS.


stp-ip:
404 on the seed corpus for the fuzzing. Restarted the job for good measure, but lgtm.

/lgtm

Issue #2389, opened 2018-12-11
plugin/rewrite: rewrite answer for regexes based on question


SethCurry:
CoreDNS Version: 1.2.6

I have a use case to rewrite *.publicdomain.com such that it returns results for fqdn.publicdomain.com.  I tried using the rewrite plugin with this configuration:

```
rewrite name regex .*\.publicdomain.com fqdn.publicdomain.com
```

However that only rewrites the question, and not the answer, so the response is A records for fqdn.publicdomain.com.  Some clients (like curl) are okay with that and will use the response, but other clients (like Python2) seem to actually check whether the returned name matches the requested name.  Python2 reports that the domain could not be resolved with this rule in effect.

I also tried to use the file plugin, but I couldn't find a way to allow only fqdn.publicdomain.com to pass through to upstream DNS servers.

I don't know what the best solution would be, but one though I had was to propogate the matches from the `name` regex to `answer` regex if both are wrapped in a single regex statement like:

```
rewrite stop {
        name regex (.*)\.publicdomain\.com fqdn\.publicdomain\.com
        answer name fqdn\.publicdomain\.com {name-1}\.publicdomain\.com
}
```

johnbelamaric:
Yes, that's the basic idea. But you should anchor with ^ and $ too, or you'll have issues. And the syntax for the group is just {1}.  For example:

```
           name regex (.*)\.my\.domain\.$ {1}.default.svc.cluster.local.
           answer name (.*)\.default\.svc\.cluster\.local\.$ {1}.my.domain.
```

SethCurry:
Oh thanks for the tip there, but I think my issue is slightly different than that.  I need to squash many domains (*.my.domain) to a single domain (fqdn.my.domain) for the name regex, and then on the answer regex I need to restore the portion of the name that I rewrote.

So if someone requests `nada.my.domain` I want to forward a DNS request for `fqdn.my.domain` from CoreDNS, and then have CoreDNS take the response from the upstream server and rewrite it so that it keeps the IPs from the A records for `fqdn.my.domain`, but changes the DNS name in the response to `nada.my.domain` so clients don't freak out that they requested `nada.my.domain` and CoreDNS is returning records for `fqdn.my.domain`.

johnbelamaric:
Ah, so you want to use one of the matches from the name regex in the answer. I get it. @greenpau thoughts?

[...]

Issue #2202, opened 2018-10-16
plugin/rewrite: Always fix answer when question is rewritten


chrisohaver:
In `rewrite`, when we change the question in a request, we need to fix the question in the response to match the original question.  If you don't do this some DNS clients complain of a mismatch between the question and answer.  The `rewrite` plugin allows you to do this with "answer" rules.

IMO, since you always want to rewrite the question in the answer to match the original question, this should be the default behavior.  There should be no need to specify an answer rule.  This would make rewrite rules easier to write, and would do away with the verbose "answer" rewrite syntax.  It also cleans up the code a bit, removing the parsing and special checks for the answer rules.

greenpau:
> There should be no need to specify an answer rule. This would make rewrite rules easier to write, and would do away with the verbose "answer" rewrite syntax. 

@chrisohaver , hope you doing well! I am getting back to the rewrite code base. Have some some to tinker. I agree with what you are saying. At the same time, there could be times when "answer" clause is necessary. I can't think about a specific use case at the moment :thinking: :thinking: return multiple CNAMEs, etc.

I also think that we could improve on the syntax of `rewrite` rules. 

I want to take think opportunity and connect this issue with https://github.com/coredns/coredns/issues/2389. I am planning working on it. Let's see how we can improve.

PR #3273, opened 2019-09-10
plugin/rewrite: support 'answer question'

euank:

This adds the ability for a rewrite plugin to dictate that the result of a rewrite should just be whatever the question was.

An example of what this plugin allows:

```sh
# before this change
=== Corefile ===
.:1053 {
    log
    errors
    rewrite name regex (.*).foo.bar. google.com. answer name google.com foo.bar
    forward . /etc/resolv.conf
}

$ dig mail.foo.bar
;mail.foo.bar.			IN	A

;; ANSWER SECTION:
foo.bar.		32	IN	A	<google ip>

# No way to get mail.foo.bar here, unless we either only route known subdomains of `foo.bar`, or unless we route to {1}.google.com to retain that information (which is semantically very different).

# after this change
=== Corefile ===
.:1053 {                                                                        
    log
    errors
    rewrite name regex (.*).foo.bar. google.com.
    forward . /etc/resolv.conf
}

$  dig mail.foo.bar
                                                                
; <<>> DiG 9.12.3-P1 <<>> @127.0.0.1 -p 1053 mail.foo.bar
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 44000
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; COOKIE: 21fe703e266695d9bc8b1e1a5d7742b8dfa4e255943dea03 (good)
;; QUESTION SECTION:
;mail.foo.bar.			IN	A

;; ANSWER SECTION:
mail.foo.bar.		247	IN	A	172.217.25.206
```

#2389, #2202

Updated in this PR

Yes. It's technically possible people relied on the old broken default, though unlikely.

It's also possible people had incorrect Corefiles which the old rewrite rule parser didn't error on, but the new one will error on.

Changes:

* The rewrite plugin now rewrites all answers by default to match the question. The previous behavior resulted in invalid responses in many common cases. The old behavior may be induced by appending `answer name (.*)\. {1}` to a rewrite rule.
* The rewrite plugin's "answer name" syntax is now supported for all `rewrite name` match types. Previously, it was only supported for `regex` type rewrites. It is now supported by `exact`, `prefix`, `suffix`, and `substring` match types as well.
* The rewrite plugin is less lenient in parsing `name` type syntax. For example, invalid trailing data will now cause an error in more cases.


chrisohaver:
Thanks for the PR!

The first example in the description doesnt work because the answer rewrite doesnt match and backreference the subdomain ... If the rewrite plugin is written as ...

```
rewrite name regex (.*)\.foo\.bar\. {1}.coredns.io. answer name (.*)\.coredns\.io\. {1}.foo.bar.
```

Then for example `www.foo.bar` works...

```
$ dig @localhost www.foo.bar.

; <<>> DiG 9.10.6 <<>> @localhost www.foo.bar.
; (2 servers found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 29061
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
;; QUESTION SECTION:
;www.foo.bar.			IN	A

;; ANSWER SECTION:
www.foo.bar.		544	IN	A	176.58.119.54

;; Query time: 6 msec
;; SERVER: ::1#53(::1)
;; WHEN: Fri Sep 13 11:28:38 EDT 2019
;; MSG SIZE  rcvd: 67
```

Having to specify the precise reciprocal answer rewrite is admittedly tedious, and prone to error (as in this case).  IMO, making the answers match the question should be default behavior if no answer rewrite is specified. IOW...

```
rewrite name regex (.*)\.foo\.bar\. {1}.coredns.io.
```

... should automatically fix the answers to match the question, since that is what most (all?) users would want and expect. #2202

I don't think there is any reason a user would want to craft a response that had mismatched question and answer sections (since clients generally reject them).  But if they really wanted to, they could control that with an answer rewrite regex that does nothing, like the following... 

```
rewrite name regex (.*)\.foo\.bar\. {1}.coredns.io. answer name (.*) {1}
```

euank:
@chrisohaver I understand that I can write the first rewrite rule in such a way that I retain the information. However, that's semantically significantly different.

Your suggested solution assumes that the rewrite target has the same dns structure as the request, which typically won't be the case.

For a specific example, if I have a kubernetes cluster with a wildcard ingress entry for "*.example.com", then I might wish to write the following rule:

```
rewrite (.*)\.example\.com nginx-ingress-controller.kube-system.svc.cluster.local
```

In that case, if I used `{1}.nginx-ingress-controller...`, that wouldn't work, there aren't sub-dns entries.

I probably should have started with that example since it's a more concrete case where the existing setup doesn't work to my knowledge.

> Having to specify the precise reciprocal answer rewrite is admittedly tedious, and prone to error

I think it's not just tedious, but not possible in my case since I need to lossily flatten dns requests down, as above.

>  IMO, making the answers match the question should be default behavior if no answer rewrite is specified.

I agree! But in this case, doing the right thing is at odds with backwards compatibility.

I don't know what stance the Coredns project has on backwards compatibility, and if it's okay to break backwards compatibility, I'd be happy make `answer question` the default behavior.

chrisohaver:
Ah, I see now, your example rule is intentionally squashing all subdomains down to a single answer... behaving somewhat like a wild card record. Sorry, I didn't scroll right to read this comment fully:

> ```# No way to get mail.foo.bar here, unless we either only route known subdomains of `foo.bar`, or unless we route to {1}.google.com to retain that information (which is semantically very different).```

> I don't know what stance the CoreDNS project has on backwards compatibility, and if it's okay to break backwards compatibility, I'd be happy make answer question the default behavior.

We typically break backward compatibility every minor release (usually through deprecation of options or plugins). Yes, this is technically a backward compatibility... but I don't think it would be backward compatible in a significantly impactful way.  I don't think there is anyone out there using a rewrite rule without an answer re-write _intentionally_.  Doing so produces an invalid response.

There is one exception I'm not sure about ... there may be some way of crafting a chain of rewrites that depend on this behavior in some way.  I.e. A series of name rewrites of which the first rule _does not_ rewrite the answer, and the latter rules which _do_ rewrite the answer and depend on the original answer being unchanged. Possible? Realistic use case? Thoughts?

euank:
That seems like a possible, but unlikely, use case to me.

I do think the existing rewrite rules can always mimic the previous behavior (I think `rewrite ... answer name (.*) {1}` is identical to the current default behavior, right?), so people who intend to retain the un-rewritten answer would have an option at least.

If we're fine with breaking backwards compatibility, I would rather just flip the new behavior I added to be the default. If that sounds like a good plan / like something that we'd want to merge, I can do that and try to tackle the docs (though, fyi, I'm on vacation for ~a week).
I will want to keep a similar rewrite of the parser / add 'rewrite name' support for more rule types either way if we're okay with that too.

chrisohaver:
Pinging owners of this plugin @greenpau, @johnbelamaric.

Which do you prefer?
* Introduce the new `answer question` syntax currently proposed in this PR?
* Or leave syntax as is, and make answer rewrites the default behavior? (functionally backward incompatible in case described [above](https://github.com/coredns/coredns/pull/3273#issuecomment-531778213))

johnbelamaric:
Is it possible to simulate the old default, broken behavior? If so, I am OK treating this as a bug fix to make it work this way by default. As long as there is a way to workaround it in the unlikely event someone wants the old behavior.

chrisohaver:
> Is it possible to simulate the old default, broken behavior?

Yes, I proposed with something like ... 
```
rewrite name regex (.*)\.foo\.bar\. {1}.coredns.io. answer name (.*) {1}
```
... in which case coredns should use the user defined answer rewrite instead of using the default behavior.  The user defined answer rewrite here being a no-op (albeit a bit obscure).

johnbelamaric:
Ok, let's go ahead with this as a bug fix then, making it the default behavior. The release note and README should be updated to provide the work around.

euank:
I've updated this PR, including docs. I think the only remaining bit is changelog entries, double checking one possible bug, and possibly a bit more testing.

The one possible bug is that `rewrite .... answer name (.*) {1}` will rewrite an answer of `foo.com.` to `foo.com..`, which is obviously silly. I think that behavior existed before, and I need to double check if it did, because if so that's a changelog entry too when I fix it.

Edit: the above bug exists on master (due to doing a fqdn normalize on the value of a answer rewrite, even though it might not be a fqdn due to including regexes). I think that'll be easier to tackle separately, so I've updated the docs and punted it.

Is there a better place to leave a changelog than in a comment here btw?

------------------------------------------------------------------------

Changes:

* The rewrite plugin now rewrites all answers by default to match the question. The previous behavior resulted in invalid responses in many common cases. The old behavior may be induced by appending `answer name (.*)\. {1}` to a rewrite rule.
* The rewrite plugin's "answer name" syntax is now supported for all `rewrite name` match types. Previously, it was only supported for `regex` type rewrites. It is now supported by `exact`, `prefix`, `suffix`, and `substring` match types as well.
* The rewrite plugin is less lenient in parsing `name` type syntax. For example, invalid trailing data will now cause an error in more cases.

chrisohaver:
A rebase should fix the ci failure.

euank:
Updated and rebased; I think this is in a reviewable state now!

greenpau:
@euank , what about branch conflicts?

euank:
Rebased onto the current master, still should be good to review.

greenpau:
@euank , there is a negative coverage. Could you please go over the code and develop the test cases that would get that stat up?

euank:
@greenpau it's complaining that I lost coverage on `plugin/kubernetes/controller.go` and a bunch of other files I didn't touch.

It seems like the bot's information doesn't match up at all with this diff, so it's not really actionable.

Am I reading it wrong or misunderstanding something?

greenpau:
> Am I reading it wrong or misunderstanding something?

@euank , there are a lot of dependencies in the code. Inevitably, your changes may impact other plugins. A major change in this plugin will likely have an effect on other plugins.

I do understand that you worked hard on this one and, perhaps, in your mind you reached the "finish line." However, for this project, it is important to have good coverage for tests. It is hard work to write a test for this plugin, but it is totally worth it. One way of doing this is opening the coverage report and review the lines of code for which there were no tests. Then, engage in a creative process of test creation. With your changes, you can change behavior or break other plugins.

At the moment, you would need to understand how your changes impact these other files.

chrisohaver:
The man page changes should not be included.  These are autogenerated at release time from the READMEs.

The go.mod and go.sum changes don't seem related to this PR at all. I don't think they should be in the PR.

CodeCov is quirky/buggy sometimes.  Looking at the code changes, there is no reason to think that this actually reduces coverage in, for example, kubernetes/controller.go.  Seems to me this is CodeCov being confused.

Try pushing again, to see if that resolves the Code Cov confusion.

chrisohaver:
... and double check that your master is up to date with upstream master.

johnbelamaric:
Yeah, I think CodeCov may get confused if you don't rebase to upstream
master.

euank:
My PR was based against dfcb4edbe when it generated that codecov report, which is 3 commits behind master. The 3 commits since only touched go.mod, the readme, and comments, so I really don't get how it had such a strange diff.

All the same, I've rebased again against the current master.

@chrisohaver regarding the go.mod / go.sum changes, those are because I ran `go mod tidy`. There is no actual change in what dependencies are or aren't used, and it's a more clear/correct go.mod and go.sum file, so I'd rather keep that diff.

I also removed the man page change per your request.
Let me know if I missed any comments / steps!

chrisohaver:
> There is no actual change in what dependencies are or aren't used, and it's a more clear/correct go.mod and go.sum file, so I'd rather keep that diff.

Perhaps.  But they are not related at all to this change.

euank:
Rebased again, including making sure the `go.mod`/`go.sum` diff got removed

greenpau:
@euank , currently, there is failed coverage in:
* `name.go`
* `ttl.go`

Where are the additions to existing test cases?

image

image

plugin/rewrite/name.go
```diff
@@ -11,54 +11,60 @@ import (
...
-	// SubstringMatch matches on partial match of the name in the question section of a request
-	SubstringMatch = "substring"
-	// RegexMatch matches when the name in the question section of a request matches a regular expression
-	RegexMatch = "regex"
-)
-
 // Rewrite rewrites the current request based upon exact match of the name
 // in the question section of the request.
-func (rule *exactNameRule) Rewrite(ctx context.Context, state request.Request) Result {
+func (rule exactRewriter) Rewrite(ctx context.Context, state request.Request) Result {
```
greenpau, line 67: @euank , why removing the pointer?

plugin/rewrite/README.md
```diff
@@ -18,13 +18,20 @@ rewrite [continue|stop] FIELD [FROM TO|FROM TTL]
 
 * **FIELD** indicates what part of the request/response is being re-written.
 
-   * `type` - the type field of the request will be rewritten. FROM/TO must be a DNS record type (`A`, `MX`, etc.);
-e.g., to rewrite ANY queries to HINFO, use `rewrite type ANY HINFO`.
+   * `type` - the type field of the request will be rewritten. FROM/TO must be
+     a DNS record type (`A`, `MX`, etc.); e.g., to rewrite ANY queries to
+     HINFO, use `rewrite type ANY HINFO`.
    * `class` - the class of the message will be rewritten. FROM/TO must be a DNS class type (`IN`, `CH`, or `HS`); e.g., to rewrite CH queries to IN use `rewrite class CH IN`.
```
greenpau, line 24: @euank . please split this line too; That's pretty much the only line you did not touch :smile:

plugin/rewrite/reverter_test.go
```diff
@@ -17,17 +16,14 @@ var tests = []struct {
 	answer   []dns.RR
 	to       string
 	toType   uint16
-	noRevert bool
 }{
-	{"core.dns.rocks", dns.TypeA, []dns.RR{test.A("dns.core.rocks.  5   IN  A  10.0.0.1")}, "core.dns.rocks", dns.TypeA, false},
-	{"core.dns.rocks", dns.TypeSRV, []dns.RR{test.SRV("dns.core.rocks.  5  IN  SRV 0 100 100 srv1.dns.core.rocks.")}, "core.dns.rocks", dns.TypeSRV, false},
-	{"core.dns.rocks", dns.TypeA, []dns.RR{test.A("core.dns.rocks.  5   IN  A  10.0.0.1")}, "dns.core.rocks.", dns.TypeA, true},
-	{"core.dns.rocks", dns.TypeSRV, []dns.RR{test.SRV("core.dns.rocks.  5  IN  SRV 0 100 100 srv1.dns.core.rocks.")}, "dns.core.rocks.", dns.TypeSRV, true},
-	{"core.dns.rocks", dns.TypeHINFO, []dns.RR{test.HINFO("core.dns.rocks.  5  HINFO INTEL-64 \"RHEL 7.4\"")}, "core.dns.rocks", dns.TypeHINFO, false},
```
greenpau, line 26: @euank . what is the reason to remove some of the tests here?

greenpau:
@euank , please reinstate all of the removed and modified tests. I see some tests were re-written.

euank:
> @euank , please reinstate all of the removed and modified tests. I see some tests were re-written.

Could you please specifically comment on any such cases in the code review itself? I'd be happy to discuss any specific cases, but the vague "you removed/modified tests" doesn't help discussion at all. I think the code and test changes I made are an improvement. Specific comments will let us discuss this productively.

greenpau:
@euank , you added test cases and you also modified existing test cases. The test cases are there to test the new additions. If you modify them, then the tests that were previously passing will no longer apply. If you want to remove/modify existing test cases, please complete this PR without removing them first. Once you get it merged, you could submit another PR for removal.

Your PR does not merely add test cases; it uses completely new testing constructs, plus modifies/removed existing test cases. Please look at it from my perspective, how do I know you modification pass previously existing test?

To be more specific, I looked at side by side diff and noticed the above. You can do the same.

euank:
@greenpau I'll repeat that this discussion would be more helpful on any specific cases you see as problematic. Again, I think the changes I made make sense. If it would help, I'm happy to explain any specific changes or testing constructs, but you'd have to specifically ask about them.

> Please look at it from my perspective, how do I know you modification pass previously existing test?

Code review is supposed to catch that sort of thing. If you think I lost meaningful useful coverage, let's discuss it.
Tests are only a tool in order to assist in reasoning about code's correctness. They're not some holy end to themselves, so it's kinda a moot point if we lost an existing test that either no longer applies (because e.g. the code refactor made it moot), or because it serves no purpose towards that goal.

> To be more specific, I looked at side by side diff and noticed the above. You can do the same.

I can read the side-by-side diff, and I can read you saying "I'm concerned about some of the changes in the diff", but I can't read your mind to figure out which lines you mean. It would be easier for both of us, I suspect, to discuss any specific changes you have concerns about.

miekg:
>@greenpau I'll repeat that this discussion would be more helpful on any specific cases you see as problematic. Again, I think the changes I made make sense. If it would help, I'm happy to explain any specific changes or testing constructs, but you'd have to specifically ask about them.

Any PR sent should be as small and focussed as possible as this helps the 
reviewer/maintainer.

Touching code that should also be changed because you happen to spot it while 
working on your code should (almost) always be deferred to another PR, again to 
make things as small as possible.

euank:
@miekg I agree that PRs should be small and focused, of course, though only in-so-far as it aids in reviewing and understanding the change.
It isn't an end to itself either though; some things that reduce the size of a PR end up hurting readability.
For example, I could remove the docs changes and submit those as a separate PR, but that doesn't make sense to me because the docs and code changes go together and are easier to reason about together, not to mention that it would leave the master branch in an inconsistent state to merge them separately.

I want to help make this PR as easily reviewable as I can, but vague platitudes aren't getting us there. Discussing any specific technical details would be more helpful for me.

miekg:
>master branch in an inconsistent state to merge them separately.

yes, changes that naturally fit together should be done in a single PR.

>I want to help make this PR as easily reviewable as I can, but vague platitudes 
>aren't getting us there. Discussing any specific technical details would be 
>more helpful for me.

The ask is to back out changes to the already existing tests. Which is a fair 
comment from a maintainer IMO, so I would just comply.

euank:
@miekg 
> The ask is to back out changes to the already existing tests. Which is a fair
comment from a maintainer IMO, so I would just comply.

I want to comply. I asked for comments on specific changes that were at issue so we could make progress.

I obviously can't just revert all test changes. For example:

```diff
-  {[]string{"name", "a.com", "b.com"}, false, reflect.TypeOf(&exactNameRule{})}, 

+  {[]string{"name", "a.com", "b.com"}, false, reflect.TypeOf(&nameRule{})}, 
```
is a required change because the the type in the go code was changed, so the test's struct needs to change to compile at all.

The existing test code had no way to correctly test rewrites of responses, and since this change is entirely to do with rewrites of responses I had to improve the testing constructs just to be able to test my change at all, which also modified some tests a little that relied on the old test code that was incapable of testing this change.

Again, I'm happy to have this discussion and try to figure out the best way to get this in a reviewable/mergeable state, but I need someone to work with me, not just tell me "you can read the diff and figure out what to change yourself, I'm not going to tell you what or where it is"

greenpau:
@euank , please consider myself being your teammate. I understand you feel confident about your work. You outlined the the plugin, looked at it holistically and modified as you saw works best. I do not doubt your technical abilities. In fact, I saw your blog posts and was very impressed with the way you approached finding solutions.

A lot of people are using this plugin, and we don’t want it to break other people’s work. That is why I am being cautious.

I proposed you split your PR in multiple parts. For example.
- testing (... typing on my phone ...) that is where you have must compile type of rules
- argument handling; that’s iterator
- more tests
- ... some other chunks of code I missed
- plugin behavioral changes
- removal of unnecessary tests

If you want to chat over the phone to sync on this, let’s do it. Email me and we will agree on time.

miekg:
>If you want to chat over the phone to sync on this, let’s do it. Email me and we will agree on time.

I applaud this approach, but I fear I won't scale.

I will *always* back maintainers on this project. Those are the ones who put in 
the long term effort.

stp-ip:
Backing the maintainer yes, but inline code comments are definitely a bit more actionable for contributors than generalized statements about a whole commit especially as it was about a few specific instances.
What might feel clear to a maintainer around the codebase or a PR might not so for the contributor.
Also thanks @greenpau for trying to get this resolved via 1on1 time.

greenpau:
Side comment: I am ok with me being “overruled” or “not supported.”” I don’t even pretend to know “it all.” In fact, I am probably an amateur. My goal here is to collaborate and contribute, and make it such that other people (users) do not spend countless hours on realizing that there is a bug or an issue.

@stp-ip , the @euank’s picking up one test that did not change to make his argument is flawed. I specifically remember how the test with HINFO helped addressIng an issue in another plugin. That test was removed. Just search for HINFO in the PR. (Prior to writing that test I did not know what is HINFO type was 😄 ) You see :-) i am also picking one test that stands out LOL

There were some other questionS that went unanswered, e.g. about a pointer.

Unit tests is a tool that I use to offset my own lack of knowledge (not clear, forgot, etc.) about a subject. When the previously used test approach (not merely the lines that test; here, there is a new test function) changes with the changes of a code’s behavior (here, it is iterator over the arguments), then I am not as eager to just blindly trust what I see in the results of tests.

By and large, if a PR passes the test harness, it is good indicator that things are solid.

euank:
@greenpau

>  I specifically remember how the test with HINFO helped addressIng an issue in another plugin. That test was removed. Just search for HINFO in the PR

Thanks for being more specific. I've ctrl+f'd "HINFO" in this PR and found the following lines:

```diff
--- a/plugin/rewrite/reverter_test.go
+++ b/plugin/rewrite/reverter_test.go
-        {"core.dns.rocks", dns.TypeHINFO, []dns.RR{test.HINFO("core.dns.rocks.  5  HINFO INTEL-64 \"RHEL 7.4\"")}, "core.dns.rocks", dns.TypeHINFO, false},
+        {"core.dns.rocks", dns.TypeHINFO, []dns.RR{test.HINFO("core.dns.rocks.  5  HINFO INTEL-64 \"RHEL 7.4\"")}, "core.dns.rocks", dns.TypeHINFO},

--- a/plugin/rewrite/rewrite_test.go
+++ b/plugin/rewrite/rewrite_test.go

-       r, _ = newTypeRule("stop", "ANY", "HINFO")
-       rules = append(rules, r)
+       rules := []Rule{
#     ... lines omitted
+               mustRule(newTypeRule("stop", "ANY", "HINFO")),
#      ... lines omitted
+     }
---

-               {"a.nl.", dns.TypeANY, dns.ClassINET, "a.nl.", dns.TypeHINFO, dns.ClassINET},
+               {"a.nl.", dns.TypeANY, dns.ClassINET, "a.nl.", dns.TypeHINFO, dns.ClassINET, "a.nl."},

--- a/plugin/rewrite/ttl_test.go
+++ b/plugin/rewrite/ttl_test.go
-               {"srv15.coredns.rocks.", dns.TypeHINFO, []dns.RR{test.HINFO("srv15.coredns.rocks.  5  HINFO INTEL-64 \"RHEL 7.5\"")}, 15, false},
+               {"srv15.coredns.rocks.", dns.TypeHINFO, []dns.RR{test.HINFO("srv15.coredns.rocks.  5  HINFO INTEL-64 \"RHEL 7.5\"")}, 15},
```
In those lines, I don't see any loss of coverage. If you could comment on your concern on the PR diff itself, that would help me know which one or ones we're discussing.

I'll repeat, **specific review comments on the diff** would be massively helpful for understanding what concerns you have and what the next steps are for this PR.

I would rather not do a phonecall unless we've got specific topics or areas of the PR that we can't resolve by discussing them publicly on github.
I don't feel we're at that point because I don't feel as if we've made an attempt to discuss this code change on github yet.

> There were some other questionS that went unanswered, e.g. about a pointer.

I missed that question. Can you link me to it or re-post it?

greenpau:
> I missed that question. Can you link me to it or re-post it?

@euank , here is it 

image

greenpau:
@euank , per https://github.com/coredns/coredns/pull/3273/files#diff-65ca238a05b7d3d6dffa698a66f6d354R52-R53, what is the benefit(s) of having the `mockDnsResponder`?

image

https://github.com/coredns/coredns/pull/3273/files#diff-692d7c4bdc80bd64cfe7a2c7bfb425afR176-R204

greenpau:
@euank , why did you remove the `noRevert` flag? What is the benefit.

image

euank:
@greenpau it says `Pending` beside that comment, which means you never clicked submit review, and means no one else can see it.

Maybe that's why we're so confused here. I don't see any comments on the diff or questions because you never submitted your review, so we're looking at two different things. Want to make the above questions on the diff, click submit so I can see and respond to them, and we'll have the discussion there?

The yellow highlighter circles are harder to respond to than code-review comments, so I'll wait for those to be converted into code-review comments if that's okay with you :smiley:

greenpau:
> The yellow highlighter circles are harder to respond to than code-review comments, so I'll wait for those to be converted into code-review comments if that's okay with you 😃

@euank , oh well ...

greenpau (changes requested):

greenpau:
> euank , why did you remove the noRevert flag? What is the benefit.

@euank  , the ability not to rewrite “question” in answer back to what the question was must be preserved. Perhaps adding a flag to the rewrite config line. Perhaps ‘norevert’.

greenpau:
@euank, you need to split this PR into multiple. Let’s start with the argument handling pull first. This PR “as is” is “no go”. 

cc: @chrisohaver

plugin/rewrite/name.go
euank: Per the go FAQ: https://golang.org/doc/faq#methods_on_values_or_pointers

>  For types such as basic types, slices, and small structs, a value receiver is very cheap so unless the semantics of the method requires a pointer, a value receiver is efficient and clear. 

The semantics of Rewrite do not require a pointer receiver, so it's more clear to use a value receiver, as the Go language authors recommend.

I'm happy to split this into its own PR if you'd like since it is pretty independent of this change.

miekg:
Most plugins don't follow this rule. So nacking making that change.

I'm closing and locking this pr soon

euank:
@miekg The actual change here fixes two issues (#2389, #2202). It seems silly to close and lock a PR that fixes other user's issues. Especially given there's been only one day of actual PR review since, you know, @greenpau didn't post his review until this weekend. As above, I'm happy to work with the project to get this reviewed and mergeable. From my perspective, now that there's an actual review / conversation started, we can make progress on this.

plugin/rewrite/README.md
euank: Will do

plugin/rewrite/reverter_test.go
euank: The tests here tested the `noRevert` flag which was deleted. Since it was deleted they serve no purpose now.

I wrote this explanation in my commit message here: https://github.com/coredns/coredns/pull/3273/commits/1c9eca6ff574cd489148af12848850433378ea49

To duplicate that message:


This also removes the tests for the `noRevert` behavior. The behavior
was never publicly exposed anyway, and was just used because the dns
recorder didn't capture requests robustly. After adding the
'mockDnsResponder' helper, that problem no longer exists.

euank:
@greenpau 

> the ability not to rewrite “question” in answer back to what the question was must be preserved. Perhaps adding a flag to the rewrite config line. Perhaps ‘norevert’.

That ability still exists. See the original comment in this pr (and the first several comments of discussion):

> The rewrite plugin now rewrites all answers by default to match the question. The previous behavior resulted in invalid responses in many common cases. The old behavior may be induced by appending `answer name (.*)\. {1}` to a rewrite rule.

Originally, I didn't default to the new behavior, but @johnbelamaric (https://github.com/coredns/coredns/pull/3273#issuecomment-532793958) asked me to take the approach I have now.

-------

For a little more context on the `noRevert` bool that was removed in this PR:

That boolean was not exposed publicly. It was never set in any case other than unit tests, and in unit tests it was set in order to expose the request for unit testing (along with `msgPrinter`).

Because testing my functionality required both checking the request and response, the `noRevert + msgPrinter` style of testing was insufficient, so I had to write a better unit testing construct in order to test my change. That also resulted in `noRevert` being entirely unused anywhere since it already wasn't used in the real plugin's code.

euank:
@greenpau I'm happy to split out argument parsing into its own PR when I get time to, though I'll wait on whether @miekg thinks I shouldn't be making any of these changes at all. If this PR is closed/locked, clearly there's no interest in me working with this PR to try and make it mergable.

euank:
@miekg 

> Most plugins don't follow this rule

I assume this is in response to the pointer vs value receiver thing. It would be easier for me to see that if you replied to that comment rather than the whole thread.

I briefly perused the codebase, and it looks like other implementations of plugins do use value receivers. Specifically I ran the following:

<Details>

```
$ rg 'func .* ServeDNS' | grep -v plugin.go | grep -v "_test.go" | awk '{print $3}' | sed 's/)//g' | LANG=C sort | uniq
*AutoPath
*Azure
*Cache
*CloudDNS
*Erratic
*Etcd
*External
*Forward
*GRPC
*Loop
*Metadata
*Metrics
*Route53
*Server
*errorHandler
*trace
ACL
Any
Auto
Cancel
Chaos
Dnssec
Dnstap
File
Handler
HandlerFunc
Hosts
Kubernetes
Logger
Nsid
Rewrite
RoundRobin
Whoami
Xfr
```

</Details>

Now, that's not a really scientific grep, but it looks like at least the [Kubernetes](https://github.com/coredns/coredns/blob/d56dbffa95963f1e0ca14c0f75e42c816c0783a8/plugin/kubernetes/handler.go#L12-L13), [Loadbalance](https://github.com/coredns/coredns/blob/d56dbffa95963f1e0ca14c0f75e42c816c0783a8/plugin/loadbalance/handler.go#L18), the other part of the [rewrite plugin](https://github.com/coredns/coredns/blob/d56dbffa95963f1e0ca14c0f75e42c816c0783a8/plugin/rewrite/rewrite.go#L40), and 16 other implementations of the plugin.Handler interface intentionally use value receivers. By my naive grep, over half the implementations of `plugin.Handler` use a value receiver (note: I didn't actually check beyond that naive grep though).

Since the `Handler` implementation in the rewrite plugin already uses a value receiver, it also feels more consistent to use one for the rewrite portion as well.

Am I misunderstanding the results above or some other detail? Is there a doc-comment/document about this style choice somewhere, or an additional confounding detail that explains what I'm seeing here?

miekg:
Please split this PR no smaller pieces. Closing this for now.

Issue #3241, opened 2019-09-05
[Feature Request] plugin/hosts metrics


JasonSwindle:
Howdy,

Currently the plugin `hosts` is a bit of a black-box.   You could have an action take place, and there is no way to understand it happens from a metrics or logging point of view.  It would be great to have a metrics around the number of entries loaded in the hosts file, and if possible, how many have been actions on.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] [Feature Request]..." ]
>Howdy,
>
>Currently the plugin `hosts` is a bit of a black-box.   You could have an action take place, and there is no way to understand it happens from a metrics or logging point of view.  It would be great to have a metrics around the number of entries loaded in the hosts file, and if possible, how many have been actions on.

a gauge of the number of entries is useful, last reload timestamp probably as 
well.

yeya24:
I ' d like to help with this.

miekg:
done and merged. Thank you all.

PR #3277, opened 2019-09-15
add host metrics

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

add metrics to hosts plugin
#3241 
Maybe a doc
No

### DIFF
--- a/plugin/hosts/README.md
+++ b/plugin/hosts/README.md
@@ -72,6 +72,13 @@ hosts [FILE [ZONES...]] {
   is authoritative. If specific zones are listed (for example `in-addr.arpa` and `ip6.arpa`), then only
   queries for those zones will be subject to fallthrough.
 
+## Metrics
+
+If monitoring is enabled (via the *prometheus* directive) then the following metrics are exported:
+
+- `coredns_hosts_entries_count{}` - The combined number of entries in hosts and Corefile.
+- `coredns_hosts_reload_timestamp_seconds{}` - The timestamp of the last reload of hosts file.
+
 ## Examples
 
 Load `/etc/hosts` file.
--- a/plugin/hosts/hostsfile.go
+++ b/plugin/hosts/hostsfile.go
@@ -17,6 +17,8 @@ import (
 	"time"
 
 	"github.com/coredns/coredns/plugin"
+
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 // parseIP calls discards any v6 zone info, before calling net.ParseIP.
@@ -45,7 +47,7 @@ func newOptions() *options {
 	return &options{
 		autoReverse: true,
 		ttl:         3600,
-		reload:      time.Duration(5 * time.Second),
+		reload:      time.Duration(5 * time.Minute),
 	}
 }
 
@@ -135,6 +137,8 @@ func (h *Hostsfile) readHosts() {
 	h.mtime = stat.ModTime()
 	h.size = stat.Size()
 
+	hostsEntries.WithLabelValues().Set(float64(h.inline.Len() + h.hmap.Len()))
+	hostsReloadTime.SetToCurrentTime()
 	h.Unlock()
 }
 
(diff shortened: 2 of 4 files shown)


miekg:
can you pull out that health/overloaded.go change and put it up as another PR? Thanks

yeya24:
> can you pull out that health/overloaded.go change and put it up as another PR? Thanks

Done.

plugin/hosts/hostsfile.go
```diff
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/coredns/coredns/plugin"
+	"github.com/prometheus/client_golang/prometheus"
```
miekg, line 21: newline before this import

plugin/hosts/hostsfile.go
```diff
@@ -252,3 +255,19 @@ func (h *Hostsfile) LookupStaticAddr(addr string) []string {
 	copy(hostsCp[len(hosts1):], hosts2)
 	return hostsCp
 }
+
+var (
+	hostsEntries = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "hosts",
+		Name: "entries",
```
miekg: entries_count is a better name I think

plugin/hosts/hostsfile.go
```diff
@@ -252,3 +255,19 @@ func (h *Hostsfile) LookupStaticAddr(addr string) []string {
...
+		Help: "The number of entries in hosts file and corefile",
```
miekg: "The combined number of entries in the hosts and Corefile."
(and closing dot)

plugin/hosts/hostsfile.go
```diff
@@ -252,3 +255,19 @@ func (h *Hostsfile) LookupStaticAddr(addr string) []string {
...
+	}, []string{})
+
+	hostsReloadTime = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "hosts",
+		Name:      "reload_timestamp_seconds",
```
miekg, line 271: this clashes with 'reload'; or will this also be updated?
Should probably be namaed `reload_time_seconds`. The timestamp is unix epoch seconds?

miekg (changes requested):
thanks, a bunch of comments and the README needs to be updated to reflect that these metrics are added.

plugin/hosts/hostsfile.go
yeya24: This metric will not be upadated when reloading.  IMO timestamp is OK because prometheus also uses `_timestamp_seconds` to name the time of reload.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] add host metr..." ]
>This metric will not be upadated when reloading.  IMO timestamp is OK because prometheus also uses `_timestamp_seconds` to name the time of reload.

fair enough, can't argue with that :)

yeya24:
@miekg  I have addressed all of you mentioned. So I think I just need to add docs in the hosts plugin' s Readme? Do I need to add any other docs?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] add host metr..." ]
>@miekg  I have addressed all of you mentioned. So I think I just need to add docs in the hosts plugin' s Readme? Do I need to add any other docs?

Yes, it only needs to be added in this hosts' readme. Add a new section called 
"Metrics", see other readme's that do something similar.

yeya24:
Done. PTAL

plugin/hosts/hostsfile.go
```diff
@@ -135,6 +137,8 @@ func (h *Hostsfile) readHosts() {
 	h.mtime = stat.ModTime()
 	h.size = stat.Size()
 
+	hostsEntries.WithLabelValues().Set(float64(h.inline.Len() + h.hmap.Len()))
+	hostsReloadTime.SetToCurrentTime()
```
miekg: should we use `ModTime()` here?

miekg (commented):
1 question, otherwise lgtm

plugin/hosts/hostsfile.go
yeya24: Thanks for your suggestion. I have changed to use `stat.ModTime()`. It is more appropriate

plugin/hosts/README.md
```diff
@@ -72,6 +72,13 @@ hosts [FILE [ZONES...]] {
   is authoritative. If specific zones are listed (for example `in-addr.arpa` and `ip6.arpa`), then only
   queries for those zones will be subject to fallthrough.
 
+## Metrics
+
+If monitoring is enabled (via the *prometheus* directive) then the following metric is exported:
```
miekg:
```suggestion
If monitoring is enabled (via the *prometheus* directive) then the following metrics are exported:
```

miekg (approved):

PR #3285, opened 2019-09-19
modules: upgrade to latest k8s api/api-machinery

miekg:
Upgrade some k8s related packages, clean the go.sum file.
Remove the Initializers from object.go as they are now gone.

Trying to get client-go upgraded to something higher as well - but failed in this pr


plugin/kubernetes/object/object.go
```diff
@@ -16,7 +16,7 @@
 package object
 
 import (
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
miekg, line 19: gofmt introduces this; no idea why, so meh

miekg:
grr, conflicts, don't want to fix those.

PR #3286, opened 2019-09-19
Run go mod -tidy

miekg:
Tidy up and fix the require of x/net to the hash used in miekg/dns, this
seems to fix the "incompatible version bla bla"


yongtang (approved):
LGTM

Issue #3261, opened 2019-09-07
Stance/future with Caddy v2


stp-ip:
Caddy is working on a revamped v2 and we might wanna think about our future stance on this.

Do we want CoreDNS to migrate to v2?
Would this mean CoreDNS v2 as well?
What are the main pain points within Caddy v1?

Starting this discussion now as Caddy v2 is still inflight to make it possible to remove some pain points in the process, if we decide that v2 might be the way to go.

Would love your thoughts ideas etc.

For context:
https://github.com/caddyserver/caddy/wiki/v2:-Writing-a-Module
https://github.com/caddyserver/caddy/blob/v2/modules/caddyhttp/caddyhttp.go

stp-ip:
Looping in @mholt as well :)

mholt:
I think this would be a good idea, and I'd rather get feedback sooner than later especially if changes need to be made to the design (it's too late already for some things, but depends how critical it is).

However, I'm pretty confident we can meet most, if not all, of CoreDNS' needs already.

v2 _should_ be able to simplify quite a bit in general. The biggest improvement for plugins is that all config loading is done for you, yay! All you need to do is declare a type:

```go
package coredns 

type App struct {
   // config fields here
}

// since CoreDNS would be an "app" module, you'll need
// to implement the caddy.App interface with Start() and
// Stop() methods
```

Then implement the `caddy.Module` interface:

```go
func (App) CaddyModule() caddy.ModuleInfo {
	return caddy.ModuleInfo{
		Name: "dns",
		New:  func() caddy.Module { return new(App) },
	}
}
```

Then register it:

```go
package coredns

func init() {
    caddy.RegisterModule(App{})
}
```

Once you import your module's package, that's it! You can then run Caddy with a config like this:

```
{
    "apps": {
        "dns": {
             // coredns config goes here!
         }
     }
}
```

Most modules also have a `Provision()` method to set themselves up, especially if they need to load guest modules (plugins of their own). Loading modules out of a json.RawMessage field is pretty easy, but probably the "most complicated" single thing a host module will do:

```go
val, err := ctx.LoadModule("dns.whatever."+modName, app.WhateverRaw)
if err != nil {
	return fmt.Errorf("loading guest module '%s': %s", modName, err)
}
app.Whatever = val.(MyWhatever)
app.WhateverRaw = nil // allow GC to deallocate
```

That's about all there is to the basics. Hope that helps.

[...]

PR #3287, opened 2019-09-19
all: simplify registering plugins

miekg:
Abstract the caddy call and make it simpler.

See #3261 for some part of the discussion.

Go from:

~~~ go
func init() {
       caddy.RegisterPlugin("any", caddy.Plugin{
               ServerType: "dns",
               Action:     setup,
       })
}
~~~

To:

~~~ go
func init() { plugin.Register("any", setup) }
~~~

This requires some external documents in coredns.io to be updated as
well; the old way still works, so it's backwards compatible.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,18 +32,15 @@ require (
 	github.com/opentracing/opentracing-go v1.1.0
 	github.com/openzipkin-contrib/zipkin-go-opentracing v0.3.5
 	github.com/philhofer/fwd v1.0.0 // indirect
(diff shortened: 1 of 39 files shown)


johnbelamaric:
LGTM except the CI failures:

```
# github.com/coredns/coredns/plugin/deprecated
plugin/deprecated/setup.go:33:9: plugin.Register undefined (type string has no field or method Register)
Makefile:36: recipe for target 'travis' failed
make: *** [travis] Error 1
```

PR #3288, opened 2019-09-19
update doc to not use the root zone for everything

miekg:
Update all documentation in the tree to use example.org as an example
configuration (in so far possible). As to get out of the just use "."
and fallthrough and things would be fine.


yongtang (approved):
LGTM

PR #3292, opened 2019-09-22
request: use an optional bool type instead of a *bool pointer

charlievieth:
This saves us from having to allocate a `*bool` pointer when setting the Request.do field by replacing the `*bool` with an optional bool type.
N/A
N/A
N/A

### DIFF
--- a/request/bool.go
+++ b/request/bool.go
@@ -1,0 +1,22 @@
+package request
+
+// An optionalBool is a bool value that may not be set and is a replacement for
+// using a *bool pointer.
+type optionalBool int
+
+const (
+	boolUnset optionalBool = iota
+	boolFalse
+	boolTrue
+)
+
+func (o *optionalBool) Set(b bool) {
+	if b {
+		*o = boolTrue
+	} else {
+		*o = boolFalse
+	}
+}
+
+func (o optionalBool) IsSet() bool { return o != boolUnset }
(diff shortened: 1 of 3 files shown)


miekg:
If we want to make this smaller, it might be better to just remove `ip`, `port`, and other fields from the bottom of this struct.

If might be an idea to get the client's buffer size and encode the do bit in there as well, i.e. use an uint32 and use most significant bit for DO. (Have to check where that size is actually used though)

Adding a new tribool type, just for this doesn't sit well with me

rdrozhdzh:
I guess this is rather for improving performance. We'll save time on allocating and garbage collecting memory.

charlievieth:
> We'll save time on allocating and garbage collecting memory.

Yeah, this was just something small that I saw when looking at where the cache plugin was allocating memory.  The only goal here was to remove the allocation and extra pointer.  This is more of a nit / code cleanliness thing as the win here is tiny.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] request: use ..." ]
>> We'll save time on allocating and garbage collecting memory.
>
>Yeah, this was just something small that I saw when looking at where the cache plugin was allocating memory.  The only goal here was to remove the allocation and extra pointer.  This is more of a nit / code cleanliness thing as the win here is tiny.

I rather have something more comprehensive then this.

PR #3293, opened 2019-09-22
Improve the memory usage and performance of the cache plugin

charlievieth:
This PR reduces the memory usage of the cache plugin (14 vs 8 allocs and 540 vs 327 B) and improves performance as a side effect.

**PR:**
```
goos: darwin
goarch: amd64
pkg: github.com/coredns/coredns/plugin/cache
BenchmarkCacheResponse/NoPrefetch-12         	 1601374	       732 ns/op	     327 B/op	       8 allocs/op
BenchmarkCacheResponse/Prefetch-12           	 1434120	       830 ns/op	     327 B/op	       8 allocs/op
BenchmarkHash-12                             	130255627	         9.25 ns/op	       0 B/op	       0 allocs/op
PASS
```

master@ 004c5fca9d7dfb8b58d608e4d50833b40953441f
```
goos: darwin
goarch: amd64
pkg: github.com/coredns/coredns/plugin/cache
BenchmarkCacheResponse-12    	 1181955	       971 ns/op	     540 B/op	      14 allocs/op
PASS
```

N/A
N/A
N/A

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -2,7 +2,6 @@
 package cache
 
 import (
-	"hash/fnv"
 	"net"
 	"time"
 
@@ -76,21 +75,39 @@ func key(qname string, m *dns.Msg, t response.Type, do bool) (bool, uint64) {
 	return true, hash(qname, m.Question[0].Qtype, do)
 }
 
-var one = []byte("1")
-var zero = []byte("0")
+type fnv64 uint64
+
+const prime64 = 1099511628211
+
+func (s *fnv64) WriteString(data string) {
+	hash := *s
+	for i := 0; i < len(data); i++ {
+		hash *= prime64
+		hash ^= fnv64(data[i])
+	}
+	*s = hash
+}
+
+func (s *fnv64) WriteByte(c byte) {
+	hash := *s
+	hash *= prime64
+	hash ^= fnv64(c)
+	*s = hash
+}
+
+func (s fnv64) Sum64() uint64 { return uint64(s) }
 
 func hash(qname string, qtype uint16, do bool) uint64 {
-	h := fnv.New64()
-
+	var h fnv64
 	if do {
-		h.Write(one)
+		h.WriteByte('1')
 	} else {
-		h.Write(zero)
+		h.WriteByte('0')
 	}
 
-	h.Write([]byte{byte(qtype >> 8)})
-	h.Write([]byte{byte(qtype)})
-	h.Write([]byte(qname))
+	h.WriteByte(byte(qtype >> 8))
+	h.WriteByte(byte(qtype))
+	h.WriteString(qname)
 	return h.Sum64()
 }
 
--- a/plugin/cache/handler.go
+++ b/plugin/cache/handler.go
@@ -29,34 +29,39 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 	i, found := c.get(now, state, server)
 	if i != nil && found {
 		resp := i.toMsg(r, now)
-
 		w.WriteMsg(resp)
 
-		if c.prefetch > 0 {
-			ttl := i.ttl(now)
-			i.Freq.Update(c.duration, now)
-
-			threshold := int(math.Ceil(float64(c.percentage) / 100 * float64(i.origTTL)))
-			if i.Freq.Hits() >= c.prefetch && ttl <= threshold {
-				cw := newPrefetchResponseWriter(server, state, c)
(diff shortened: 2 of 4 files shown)


miekg:
Note there is also this closed PR https://github.com/coredns/coredns/pull/2527/files that tackles this on a more fundamental level, but not storing parts of a message that don't need to be stored. That didn't get merged because getting actual memory improvement data was hard.

plugin/cache/cache.go
```diff
@@ -76,21 +75,39 @@ func key(qname string, m *dns.Msg, t response.Type, do bool) (bool, uint64) {
 	return true, hash(qname, m.Question[0].Qtype, do)
 }
 
-var one = []byte("1")
-var zero = []byte("0")
+type fnv64 uint64
```
miekg, line 78: This needs some documentation on what's being done here. Is there any (semi) official hash algo that we can use? And why isn't fnv in the std library more efficient?

plugin/cache/handler.go
```diff
@@ -29,28 +29,10 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
-
-					// When prefetching we loose the item i, and with it the frequency
-					// that we've gathered sofar. See we copy the frequencies info back
-					// into the new item that was stored in the cache.
-					if i1 := c.exists(state); i1 != nil {
-						i1.Freq.Reset(now, i.Freq.Hits())
-					}
-				}(cw)
-			}
+		if c.shouldPrefetch(i, now) {
```
miekg, line 34: could change, but unrelated. Can you pull it into a new PR?

plugin/cache/item.go
```diff
@@ -9,15 +9,13 @@ import (
...
-	AuthenticatedData  bool
-	RecursionAvailable bool
 	Answer             []dns.RR
 	Ns                 []dns.RR
 	Extra              []dns.RR
-
-	origTTL uint32
-	stored  time.Time
-
+	AuthenticatedData  bool
```
miekg, line 15: unrelated change and these where grouped with the other msg related fields (i.e. there was some significance in that)

plugin/cache/item.go
```diff
@@ -62,9 +60,14 @@ func (i *item) toMsg(m *dns.Msg, now time.Time) *dns.Msg {
 	m1.RecursionAvailable = i.RecursionAvailable
 	m1.Rcode = i.Rcode
 
-	m1.Answer = make([]dns.RR, len(i.Answer))
-	m1.Ns = make([]dns.RR, len(i.Ns))
-	m1.Extra = make([]dns.RR, len(i.Extra))
+	m1.Answer = i.Answer
+	m1.Ns = i.Ns
+	m1.Extra = i.Extra
```
rdrozhdzh, line 65: do you really need these 3 lines above?

plugin/cache/item.go
```diff
@@ -62,9 +60,14 @@ func (i *item) toMsg(m *dns.Msg, now time.Time) *dns.Msg {
...
+
+	a := make([]dns.RR, len(i.Answer)+len(i.Ns)+len(i.Extra))
+	m1.Answer, a = a[:len(i.Answer)], a[len(i.Answer):]
```
rdrozhdzh, line 68: This is not quite safe. If any code later will want to append an RR to answer, it will overwrite NS record.
Consider assigning slice with capacity specified like
```
m1.Answer = a[:len(i.Answer):len(i.Answer)]
```

plugin/cache/cache.go
charlievieth: I'll document it, this is the same `fnv` function as the stdlib, but has a WriteByte and WriteString method.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Improve the m..." ]
>charlievieth commented on this pull request.
>
>
>
>> @@ -76,21 +75,39 @@ func key(qname string, m *dns.Msg, t response.Type, do bool) (bool, uint64) {
> 	return true, hash(qname, m.Question[0].Qtype, do)
> }
>
>-var one = []byte("1")
>-var zero = []byte("0")
>+type fnv64 uint64
>
>I'll document it, this is the same `fnv` function as the stdlib, but has a WriteByte and WriteString method.

I'm not a fan of rewriting parts of the stdlib, we'll now miss out any fixes 
(like ASM) they apply to this. Is there a smarter way of doing this? Dropping 
FNV all together?

miekg:
I'm moving some of this into their own PRs. For one is prefetch closure, see: #3353

miekg:
closing what's left of this pr

PR #3294, opened 2019-09-23
fix: delete unused var and const

hwdef:
delete unused var and const
NO

### DIFF
--- a/plugin/dnssec/handler.go
+++ b/plugin/dnssec/handler.go
@@ -55,13 +55,6 @@ var (
 		Help:      "The number of elements in the dnssec cache.",
 	}, []string{"server", "type"})
 
-	cacheCapacity = prometheus.NewGaugeVec(prometheus.GaugeOpts{
-		Namespace: plugin.Namespace,
-		Subsystem: "dnssec",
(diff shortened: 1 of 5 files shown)


plugin/dnssec/handler.go
```diff
@@ -55,13 +55,6 @@ var (
 		Help:      "The number of elements in the dnssec cache.",
 	}, []string{"server", "type"})
 
-	cacheCapacity = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
miekg, line 58: this needs to be removed from the README as well. (no-one ever complained this wasn't exported, so I'm OK with removing it)

miekg:
otherwise lgtm

plugin/dnssec/handler.go
hwdef: Thanks.

hwdef: @miekg 
I can not find it in README,
Can you help me find out it?
Thank you!

miekg: ah, it's not there. my mistake

PR #3301, opened 2019-09-24
Use strings.ToLower instead of manual implementation

### LINKED ISSUE #3291: strings.ToLower in core/dnsserver/server.go
Implement this:

https://github.com/miekg/dns/pull/1010

in core/dnsserver/server.go. This code is faster and less lines of code.

erbesharat:
Use strings.ToLower instead of manual implementation cause it's faster and has fewer lines of code.

fixes #3291 

Nothing

No

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"runtime"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


miekg:
sorry, this is not correct, a lot more code can be deleted.

miekg:
This might have been tricker than I anticipated, see #3304 for my attempt.

erbesharat:
@miekg I should've read the code more carefully. When I tried to remove the first for loop some of the coverage tests in `/tests` started to fail and I thought it's necessary to keep it. Should we close this PR and merge yours instead?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Use strings.T..." ]
> Should we close this PR and merge yours instead?

that's the easiest way forward, I think. The alternative is that you roughly 
implement the same, which seems a bit of an overkill :)

stp-ip:
Closing in favor of #3304

PR #3308, opened 2019-09-25
Modify iff to if in the documents

yuxiaobo96:
Signed-off-by: yuxiaobo <yuxiaobogo@163.com>

Make comments more perfect.
No.


miekg:
these actually might carry significance. We also stopped merging these one-off spelling fix.es

PR #3324, opened 2019-09-28
WIP: show how plugin.Controller will look

miekg:
This doesn't compile (easy to fix) but shows up a plugin.Controller would
look like. No plugin needs to import anything from caddy directly; it's
purely things from the plugin package.


miekg:
we *can* do this at some point, but now is not the time. Closing.

PR #3326, opened 2019-09-30
Makefiles: remove GO111MODULE

miekg:
This is the default in 1.13 it can be removed; also make the test output
log less, so failure are more pronounced in the travis output, hide
coverage stats for instance.


jameshartig (approved):

miekg:
thank you :)

PR #3334, opened 2019-09-30
WIP: plugin/forward: remove the channel communication

miekg:
Use three, protected by a mutex, slices of persistent connections. Just
pick one from the list. This removes the send and then receive; the
start/stop dance and simplifies the code.

This still needs to be load tested.

Possible things todo and fix is set a max on the number of conns in the
slice; 1000 each?

Another thing is the following scenario: big spike in qps, slice full,
then nothing for 5s; then all conns in the slice are too old to be used
and we will redial for the first 1000; is this problem? Or is this only
a problem when low qps (then it's not a problem).


miekg:
potentially I could bring back the routine that checks every x seconds if there are old connections in the list and if so truncate the list.

miekg:
hmm, the current code is *way* faster

PR #3336, opened 2019-10-01
plugin/forward: make Yield not block

miekg:
Yield may block when we're super busy with creating (and looking) for
connection. Set a small timeout on Yield, to skip putting the connecting
back in the queue.

Use persistentConn troughout the socket handling code to be more
consistent.

### DIFF
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -44,17 +44,18 @@ func (t *Transport) updateDialTimeout(newDialTime time.Duration) {
 }
 
 // Dial dials the address configured in transport, potentially reusing a connection or creating a new one.
-func (t *Transport) Dial(proto string) (*dns.Conn, bool, error) {
+func (t *Transport) Dial(proto string) (*persistConn, bool, error) {
 	// If tls has been configured; use it.
 	if t.tlsConfig != nil {
 		proto = "tcp-tls"
 	}
 
 	t.dial <- proto
-	c := <-t.ret
+	pc := <-t.ret
 
-	if c != nil {
-		return c, true, nil
+	if pc != nil {
+		pc.used = time.Now() // update used time
+		return pc, true, nil
 	}
 
 	reqTime := time.Now()
@@ -62,11 +63,11 @@ func (t *Transport) Dial(proto string) (*dns.Conn, bool, error) {
 	if proto == "tcp-tls" {
 		conn, err := dns.DialTimeoutWithTLS("tcp", t.addr, t.tlsConfig, timeout)
 		t.updateDialTimeout(time.Since(reqTime))
-		return conn, false, err
+		return &persistConn{conn, time.Now()}, false, err
 	}
 	conn, err := dns.DialTimeout(proto, t.addr, timeout)
 	t.updateDialTimeout(time.Since(reqTime))
-	return conn, false, err
+	return &persistConn{conn, time.Now()}, false, err
 }
 
 // Connect selects an upstream, sends the request and waits for a response.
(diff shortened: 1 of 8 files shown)


plugin/forward/connect.go
```diff
@@ -44,29 +44,29 @@ func (t *Transport) updateDialTimeout(newDialTime time.Duration) {
...
+		return pc, true, nil
 	}
 
 	reqTime := time.Now()
 	timeout := t.dialTimeout()
 	if proto == "tcp-tls" {
 		conn, err := dns.DialTimeoutWithTLS("tcp", t.addr, t.tlsConfig, timeout)
 		t.updateDialTimeout(time.Since(reqTime))
-		return conn, false, err
+		return &persistConn{conn, time.Now()}, false, err
```
rdrozhdzh: We should update `used` field in `func (t *Transport) Yield(pc *persistConn)` before sending to channel. This update should take place for all connections, not only to newly created.

plugin/forward/persistent_test.go
```diff
@@ -82,54 +82,6 @@ func TestCleanupByTimer(t *testing.T) {
 	tr.Yield(c4)
 }
 
-func TestPartialCleanup(t *testing.T) {
```
rdrozhdzh, line 85: do you plan to keep this test?

plugin/forward/connect.go
miekg: better is to rename used to created - not sure what my original intention were, but a def. lifetime was some part of it. I'll rename the field to reflect this - as this PR actually implement that.

plugin/forward/persistent_test.go
miekg: no, it was already not ideal with all those sleeps

plugin/forward/connect.go
rdrozhdzh: `used` is the time when this connection was last used. We need this time to detect expired connections, i.e. connection is considered expired when some time elapsed after last usage of this connection. And this is the reason why we need to update this field after each use of it.

miekg:
> 	reqTime := time.Now()
> 	timeout := t.dialTimeout()
> 	if proto == "tcp-tls" {
> 		conn, err := dns.DialTimeoutWithTLS("tcp", t.addr, t.tlsConfig, timeout)
> 		t.updateDialTimeout(time.Since(reqTime))
>-		return conn, false, err
>+		return &persistConn{conn, time.Now()}, false, err
>
>`used` is the time when this connection was last used. We need this time to detect expired connections, i.e. connection is considered expired when some time elapsed after last usage of this connection. And this is the reason why we need to update this field after each use of it.

yes, that's why I named it created. I want to drop conns *created*, not used,  X seconds ago.

rdrozhdzh:
> I want to drop conns *created*, not used,  X seconds ago.

Hm, which benefits are you going to achieve with this new approach? I think existing implementation uses cached connections more efficiently.

rdrozhdzh:
Also, the `connManager()` code relies on the fact that all connections in cache are ordered by `used`. With your new approach this logic will get broken.

miekg:
>Also, the `connManager()` code relies on the fact that all connections in cache are ordered by `used`. With your new approach this logic will get broken.

the intention is to minimze syscalls

Now they are ordered by created... how does that break the logic?

miekg:
>Also, the `connManager()` code relies on the fact that all connections in cache are ordered by `used`. With your new approach this logic will get broken.

hmmm, re-reading your original commit. I don't care so much about this, the 
important things is to not block. I'll revert some bits.

plugin/forward/connect.go
miekg: done - pushing in a minute

rdrozhdzh:
In `Yield()` we add connection to the end of slice. Since we always updated the `used` field with `Now()` the connections were automatically ordered by `used`
In `Dial()` we take the last (the freshest) connection.  If it appeared expired then we assume all connections before it are also expired. And we call cleanup for all those connections.

In you new approach, when used connection gets back to cache its `created` field may not be the greatest among other cached connection. So, cached connections will get unordered and at Dial time we'll have to check each connection if it's expired or not.

miekg:
>In `Yield()` we add connection to the end of slice. Since we always updated the `used` field with `Now()` the connections were automatically ordered by `used`
>In `Dial()` we take the last (the freshest) connection.  If it appeared expired then we assume all connections before it are also expired. And we call cleanup for all those connections.
>
>In you new approach, when used connection gets back to cache its `created` field may not be the greatest among other cached connection. So, cached connections will get unordered and at Dial time we'll have to check each connection if it's expired or not.

yep understood :) See my followup mail, reverted those changes. PTAL

plugin/forward/connect.go
```diff
@@ -44,29 +44,30 @@ func (t *Transport) updateDialTimeout(newDialTime time.Duration) {
...
 	}
 
 	t.dial <- proto
-	c := <-t.ret
+	pc := <-t.ret
 
-	if c != nil {
-		return c, true, nil
+	if pc != nil {
+		pc.used = time.Now() // update used time
```
rdrozhdzh: If we update `used` at dial time this doesn't guarantee that connections will be yielded in the same order. Ideally, `connManager()` should (and actually does) update time at yield time. As for me, I would keep the existing logic.
But I guess you want to move as much code as possible from connManager() loop. That's why I suggested to update `used` in `func (t *Transport) Yield(pc *persistConn)`. This could make cached connections slightly unordered, though.

miekg:
>But I guess you want to move as much code as possible from connManager() loop. That's why I suggested to update `used` in `func (t *Transport) Yield(pc *persistConn)`. This could make cached connections slightly unordered, though.

ack. That's indeed quite nice and removes a bunch of time.Now()'s. 

done - ptal

rdrozhdzh (approved):
lgtm

PR #3339, opened 2019-10-01
Move map to array

miekg:
The map was not needed move to an array, see #1941 for the original
idea. That of course didn't apply anymore; make a super minimal change
to implements the idea from #1941

### DIFF
--- a/plugin/forward/persistent.go
+++ b/plugin/forward/persistent.go
@@ -2,7 +2,6 @@ package forward
 
 import (
 	"crypto/tls"
-	"net"
 	"sort"
 	"time"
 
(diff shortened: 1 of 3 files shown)


miekg:
cc @drozhdzh (original PR was conflicting too much)

plugin/forward/persistent.go
```diff
@@ -17,9 +16,9 @@ type persistConn struct {
 
 // Transport hold the persistent cache.
 type Transport struct {
-	avgDialTime int64                     // kind of average time of dial time
-	conns       map[string][]*persistConn // Buckets for udp, tcp and tcp-tls.
-	expire      time.Duration             // After this duration a connection is expired.
+	avgDialTime int64             // kind of average time of dial time
+	conns       [3][]*persistConn // Buckets for udp, tcp and tcp-tls.
```
rdrozhdzh: don't you want to define a constant like `typeTotal` and put it last in the list of `typeUdp`, `typeTcp`, `typeTls`? 
Then you could specify array size as
```
    conns       [typeTotal][]*persistConn 
```
without any magic numbers

Also, in future, you will be able to add one more transport type without changing array definition/initialization code.

plugin/forward/persistent.go
```diff
@@ -32,7 +31,7 @@ type Transport struct {
 func newTransport(addr string) *Transport {
 	t := &Transport{
 		avgDialTime: int64(maxDialTimeout / 2),
-		conns:       make(map[string][]*persistConn),
+		conns:       [3][]*persistConn{},
```
rdrozhdzh: see above

rdrozhdzh (approved):
Overall looks good to me. Thanks for picking this up.
Is the test failure related to this change?

plugin/forward/persistent.go
miekg: yep good idea.

PR #3347, opened 2019-10-02
Use whitelist in .dockerignore

hloeffler:

Simplify the .dockerignore by using a white list instead of trying to blacklist everything else.

None.

None

No.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,14 +1,2 @@
-.git/*
-core/*
-coremain/*
-hooks/*
-man/*
-pb/*
-plugin/*
-request/*
-test/*
-vendor/*
-build/*
-release/*
-go.sum
-go.mod
+*
+!coredns


miekg:
I don't see how this is supposed to work, even with reading https://docs.docker.com/engine/reference/builder/#dockerignore-file myself. 

Detailing it some more in the PR description will help folks and cut down on the time needed to look at things

hloeffler:
Hello,

https://docs.docker.com/engine/reference/builder/#dockerignore-file
says:
```
Lines starting with ! (exclamation mark) can be used to make exceptions to exclusions. The following is an example .dockerignore file that uses this mechanism:

    *.md
    !README.md

All markdown files except README.md are excluded from the context.
```

And we only need to copy one file:
https://github.com/coredns/coredns/blob/master/Dockerfile#L9

So ignoring everything `*` and excluding `coredns` simplify the dockerignore file.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Use whitelist..." ]
>And we only need to copy one file:
>https://github.com/coredns/coredns/blob/master/Dockerfile#L9
>
>So ignoring everything `*` and excluding `coredns` simplify the dockerignore file.

ah, thanks, see this helps (also had some more coffee)

PR #3348, opened 2019-10-02
plugin/route53: remove amazon intialization from init

### LINKED ISSUE #3342: plugin/route53: init() allocates object we don't use
~~~
func init() {
	plugin.Register("route53",
		func(c *caddy.Controller) error {
			f := func(credential *credentials.Credentials) route53iface.Route53API {
				return route53.New(session.Must(session.NewSession(&aws.Config{
					Credentials: credential,
				})))
			}
			return setup(c, f)
		},
	)
}
~~~
this `init` allocates aws stuff even if I don't load the plugin. It should not do that.

miekg:
Don't perform this code in the init, this allocated 1 megabyte of memory
even if you don't use the plugin. Looks to be only there for testing,
adding a comment to reflect that.

Fixes #3342

### DIFF
--- a/plugin/route53/setup.go
+++ b/plugin/route53/setup.go
@@ -24,20 +24,14 @@ import (
 
 var log = clog.NewWithPlugin("route53")
 
-func init() {
-	plugin.Register("route53",
-		func(c *caddy.Controller) error {
-			f := func(credential *credentials.Credentials) route53iface.Route53API {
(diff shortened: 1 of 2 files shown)


miekg:
similar in vain, so /cc @yongtang

dilyevsky (approved):
heh, good catch! what the hell, aws?

PR #3349, opened 2019-10-02
plugin/clouddns: remove initialization from init

### LINKED ISSUE #3343: plugin/clouddns: remove allocation from init function
cloud dns allocates and initializes in the `init` function, this is bad because it does stuff even if I don't want to use the plugin.
~~~
func init() {
	plugin.Register("clouddns",
		func(c *caddy.Controller) error {
			f := func(ctx context.Context, opt option.ClientOption) (gcpDNS, error) {
				var err error
				var client *gcp.Service
				if opt != nil {
					client, err = gcp.NewService(ctx, opt)
				} else {
					// if credentials file is not provided in the Corefile
					// authenticate the client using env variables
					client, err = gcp.NewService(ctx)
				}
				return gcpClient{client}, err
			}
			return setup(c, f)
		},
	)
}
~~~
this should just call the registerPlugin function.

miekg:
Init should just call the plugin.Register with a setup function.

Fixes: #3343

### DIFF
--- a/plugin/clouddns/setup.go
+++ b/plugin/clouddns/setup.go
@@ -17,27 +17,23 @@ import (
 
 var log = clog.NewWithPlugin("clouddns")
 
-func init() {
-	plugin.Register("clouddns",
-		func(c *caddy.Controller) error {
-			f := func(ctx context.Context, opt option.ClientOption) (gcpDNS, error) {
-				var err error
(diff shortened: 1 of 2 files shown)


yongtang (approved):
lgtm, Thanks a lot for the fix!

PR #3358, opened 2019-10-05
circle CI: run with latest Go

miekg:
We compile coredns with latest Go on each release, do this same in
Circle-ci


miekg:
ci failures, have been fixed by upgrading etcd to the right version - which shows circle-ci can add value there as well. Merging.

PR #3363, opened 2019-10-07
Directive -> plugin

miekg:
Caught my eye, we name things directive still, esp when talking about
the prometheus *plugin*. Rename everything that needs to be plugin to
'plugin'. Also make sure Metrics is a H2 section (not H1).

### DIFF
--- a/plugin/autopath/README.md
+++ b/plugin/autopath/README.md
@@ -27,7 +27,7 @@ If a plugin implements the `AutoPather` interface then it can be used.
 
 ## Metrics
 
-If monitoring is enabled (via the *prometheus* directive) then the following metric is exported:
+If monitoring is enabled (via the *prometheus* plugin) then the following metric is exported:
 
 * `coredns_autopath_success_count_total{server}` - counter of successfully autopath-ed queries.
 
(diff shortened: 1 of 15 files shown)


johnbelamaric:
Hmm. Not sure this is a problem. `directive` refers to the keyword used in the Corefile. `plugin` refers to the code. So in fact the `prometheus` directive is used to enable the `metrics` plugin, if we are being precise.

stp-ip (approved):
/merge

stp-ip:
So we could say in general config = directive and implementation = plugin?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Directive -> ..." ]
>So we could say in general config = directive and implementation = plugin?

I think plugin is more clear. I think i tried defining them in the manual once, 
but haven't looked or updated that in ages. In this case you enable the metrics 
plugin, so it's very clear.

johnbelamaric:
Yeah, I am OK with blurring the line there, I think an extra word is not too helpful in most cases.

/lgtm

chrisohaver (approved):

rdrozhdzh:
My 5 cents. I like thinking of `directive` as a keyword in corefile and `plugin` as a program module. Using two different terms for different things makes documentation clearer.

miekg:
Well, if you have time and cycles to spare: https://coredns.io/manual/toc/

PR #3372, opened 2019-10-13
pkg/log: add Clear to stop debug logging

### LINKED ISSUE #3035: Removing debug plugin does not turn off debug logging
When the debug plugin is configured and then removed, the reloaded configuration does not turn off debug logging.  

One solution is to add a method, `log.D.Clear()`, that `dnsserver.NewServer()` can call before iterating through the configurations.  This seems like a good stop gap measure.

This is not ideal since any debug shutdown messages from the previous incarnation won't get printed.  A better solution may be to have a logger, specific to that plugin's incarnation, passed in during setup.  Old expiring plugins would still be able to print their debug information as they are shutting down while the next incarnations would rely on the reloaded configuration for their loggers.  An added benefit would be that if a plugin does not completely shutdown, it would still print debug logging.

miekg:
When reloading we need to disable debug output when the debug plugin is
removed from the config file. Add a `Clear` function to pkg/log and use
it in the server server.

Add test case in pkg/log, for actuall check I manually checked the
output by sprinkling some debug statements in the startup and checking
with sending SIGUSR1.

Also clear up the comments in pkg/log to remove the text about time
stamping.

Fixes: #3035

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -64,6 +64,10 @@ func NewServer(addr string, group []*Config) (*Server, error) {
 		if site.Debug {
 			s.debug = true
 			log.D.Set()
+		} else {
+			// When reloading we need to explicitly disable debug logging if it is now disabled.
+			s.debug = false
+			log.D.Clear()
 		}
 		// set the config per zone
 		s.zones[site.Zone] = site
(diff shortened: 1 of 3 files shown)


miekg:
ping @rajansandeep

rajansandeep (approved):
/lgtm

miekg:
thanks

Issue #3379, opened 2019-10-15
plugin/dnssec: ed25519 support


seankhliao:
*  the version of CoreDNS you are using: 
1.6.4, also on master
*  your Corefile
```
.:553 {
    auto {
        directory /etc/coredns/zone (.*).zone {1}
        transfer to *
        reload 10s
    }
    dnssec example.com {
        key file /etc/coredns/key/Kexample.com.+015+53911 /etc/coredns/key/Kexample.com.+015+40283
    }
    errors
    log
    reload 10s
    whoami
}
```
* logs, if applicable
```
plugin/dnssec: no private key found
```

Given that there is no mention of not supporting it, I expected support for ed25519 dnskeys (algorithm 15).  The upstream library github.com/miekg/dns already supports it, coredns just needs to allow its usage.  I can have submit a PR if needed.

also plugin/sign is probably wrong since the returned privatekey is `ed25519.PrivateKey` and not `*ed25519.PrivateKey`

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin/dnssec: ed..." ]
>*  the version of CoreDNS you are using:
>1.6.4, also on master
>*  your Corefile
>```
>.:553 {
>    auto {
>        directory /etc/coredns/zone (.*).zone {1}
>        transfer to *
>        reload 10s
>    }
>    dnssec example.com {
>        key file /etc/coredns/key/Kexample.com.+015+53911 /etc/coredns/key/Kexample.com.+015+40283
>    }
>    errors
>    log
>    reload 10s
>    whoami
>}
>```
>* logs, if applicable
>```
>plugin/dnssec: no private key found
>```
>
>Given that there is no mention of not supporting it, I expected support for ed25519 dnskeys (algorithm 15).  The upstream library github.com/miekg/dns already supports it, coredns just needs to allow its usage.  I can have submit a PR if needed.
>
>also plugin/sign is probably wrong since the returned privatekey is `ed25519.PrivateKey` and not `*ed25519.PrivateKey`

PR welcome!

PR #3380, opened 2019-10-15
plugin/dnssec, plugin/sign: ed25519 support

seankhliao:
adds support for ed25519 keys to plugin/dnssec,  and corrects plugin/sign ed25519 type assertion

#3379 

none

no

### DIFF
--- a/plugin/dnssec/dnskey.go
+++ b/plugin/dnssec/dnskey.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/coredns/coredns/request"
-
 	"github.com/miekg/dns"
 
 	"golang.org/x/crypto/ed25519"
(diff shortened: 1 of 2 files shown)


miekg:
This might need a tweak in plugin/sign as well. Could you check?

Current PR lgtm

seankhliao:
i just checked on my server, the second commit `f3e5eb7` does properly fix plugin/sign

plugin/dnssec/dnskey.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/coredns/coredns/request"
 
 	"github.com/miekg/dns"
+
```
miekg, line 13: nit: doesn't need newline

miekg:
I believe both README.md now also need to be updated?

seankhliao:
I'm not sure which part of the READMEs to update. People should probably still use ecdsa keys if they don't want things to break. With this change coredns supports almost everything `dnssec-keygen` will generate as a signing key (ed448 is the exception, no stdlib or x/crypto support)

miekg:
ok, there also is https://github.com/coredns/coredns-utils/tree/master/coredns-keygen which we may want to give a single switch to allow ed25519 keys

Issue #2898, opened 2019-06-18
Handling of invalid domain names in a Corefile


chrisohaver:
Across most plugins (all plugins?), when adding zones, we call `Normalize()` to make sure we have a valid set of zones.  However, if a zone is invalid (e.g. too long), then we normalize that zone to an empty string, without warning.  The empty string zone, i think gets interpreted as the root zone, which can result counterintuitive behavior, e.g. the plugin may begin to operate on all zones.

One way to address this would be to add a call to `dns.IsDomain(zone)` in each plugin where we normalize the zones, and exit if we encounter an invalid zone.  Or perhaps instead have all plugins call a new alternate version of `Normalize()` that also returns errors.  But there may be a more global way to fix this, hence opening this up for ideas before diving in with hammers and nails.

miekg:
Alternatively just add IsDomain in the core/dnsserver/register.go (or there 
abouts) and pre-validate the name - and error there

chrisohaver:
Ah - yes thanks!  That would require many fewer nails.

chrisohaver:
I cant think of a way to do this in `core/dnsserver/register.go` without some modifications to each plugin that we'd want to validate.  The least intrusive way i can think of is ...
* create a new interface with a zones getter `type Zoner interface { GetZones() []string }`
* for every plugin that has >= 1 zone in its config, add a `GetZones()` that returns a list of all zones.
* in `core/dnsserver/register.go` `NewServer()`, validate all zones for each plugin that implements `Zoner`.

chrisohaver:
I can kind of get it to work via reflection, but its ugly, and not all plugins store their zones in the same place.

But, a bigger problem with doing this at `register.go` time is that each plugin is already calling `Normalize()` on zones, so by the time they register, all the invalid zones have already been replaced with the root zone.   I.e. this cannot be done at register time unless we change all plugins to not Normalize their zones first.

miekg:
We can validate names in the corefile server stanza easy enough. What
plugins do is up to them, but we can give some helper function.

A MustNormalize or some such that returns an error and slowly move to that.

miekg:
are you working on this @chrisohaver ?

chrisohaver:
No

PR #3385, opened 2019-10-18
Add a parse friendly Normalize

chrisohaver:

Adds `MustNormalize()`, which is the same as `Normalize()` but does not ignore parse errors.
Going forward plugins should prefer `MustNormalize()` instead of `Normalize()` when reading parsing the Corefile.

 #2898

none

no


miekg:
thank you
/lgtm

PR #3400, opened 2019-10-25
bufsize plugin for preparing the DNS Flag Day and avoiding IP fragmentation

ykhr53:
In the next DNS Flag Day, message size should be considered to avoid IP fragmentation.  

https://dnsflagday.net/2020/
> Action: DNS Resolver Operators
> Requrirements on the resolver side are more or less the same as for authoritative: ensure that your servers can answer DNS queries over TCP (port 53), and configure an EDNS buffer size of 1232 bytes to avoid fragmentation. Remember to check your firewall(s) for problems with DNS over TCP!

Also we have to consider about fragment attack, *bufsize* can limit requester's UDP payload size.  
This plugin can follow this draft too.  

https://tools.ietf.org/html/draft-fujiwara-dnsop-fragment-attack-01#section-4.3
> 4.3.  Limit requestor's UDP payload size to 512
>    Limiting EDNS0 requestor's UDP payload size [RFC6891] to 512 may be a
>    measure of path MTU attacks.

This plugin prevents IP fragmentation so that to be ready for the DNS Flag Day 2020 and to deal with DNS vulnerability.

None.

None.


ykhr53:
Sorry this PR have many commits, I'll create clean one.

Issue #2972, opened 2019-07-03
2020 DNS Flagday


miekg:
For the 2020 DNS flag day there are still some remaining bits, see http://dnsflagday.net/2020/#action-authoritative-dns-operators

* the *forward* plugin should set edns buf size to 1220 or 1480 based on outgoing IP familty
* actually truncate on 1220 and 1480 ? We use client edns0 bufsize for this now (likely set to 4096)

stp-ip:
Do we need this for the deprecated proxy plugin?

Or can we see proxy as deprecated?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] 2020 DNS Flag..." ]
>Do we need this for the deprecated proxy plugin?

most likely yes

miekg:
I meant *forward* in the above

miekg:
See https://github.com/dns-violations/dnsflagday/issues/125

I will not implement "set TC" to prevent udp fragmentation, because there is no way of knowing what MTU exists for coredns in all the crazy env it runs. So we will remain compressing *harder* with whatever value that threads comes up with.

ykhr53:
Hi  
I would like to propose a plugin for modifying EDNS0 buffer size.  
*bufsize* plugin allows to overwrite EDNS bufsize. It doesn't need to change *forward* or *cache* plugin.

Here is my idea. https://github.com/ykhr53/coredns/tree/dev-bufsize/plugin/bufsize

Related issues;  
- forward: allow use edns0 https://github.com/coredns/coredns/issues/3386  
- 2020 DNS Flagday https://github.com/coredns/coredns/issues/2972  

It's just an idea came up with and I think it's not enough, much appreciate if you give me more suggestions.
Thanks,

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] 2020 DNS Flag..." ]
>Hi
>I would like to propose a plugin for modifying EDNS0 buffer size.
>*bufsize* plugin allows to overwrite EDNS bufsize. It doesn't need to change *forward* or *cache* plugin.
>
>Here is my idea. https://github.com/ykhr53/coredns/tree/dev-bufsize/plugin/bufsize
>
>Related issues;
>- forward: allow use edns0 https://github.com/coredns/coredns/issues/3386
>- 2020 DNS Flagday https://github.com/coredns/coredns/issues/2972
>
[...]

PR #3401, opened 2019-10-26
Add bufsize plugin for preparing the DNS Flag Day and avoiding IP fragmentation

ykhr53:
In the next DNS Flag Day, message size should be considered to avoid IP fragmentation.  

https://dnsflagday.net/2020/
> Action: DNS Software Vendors
> It is important for DNS software vendors to comply with DNS standards, and to use a default EDNS buffer size (1232 bytes) that will not cause fragmentation on typical network links.

Also we have to consider about fragment attack, *bufsize* can limit requester's UDP payload size.  
This plugin can follow this draft too.  

https://tools.ietf.org/html/draft-fujiwara-dnsop-fragment-attack-01#section-4.3
> 4.3.  Limit requestor's UDP payload size to 512
>    Limiting EDNS0 requestor's UDP payload size [RFC6891] to 512 may be a
>    measure of path MTU attacks.

This plugin prevents IP fragmentation so that to be ready for the DNS Flag Day 2020 and to deal with DNS vulnerability.

#2972

None.

None.

### DIFF
--- a/plugin/bufsize/README.md
+++ b/plugin/bufsize/README.md
@@ -1,0 +1,34 @@
+# bufsize
+## Name
+*bufsize* - modifies EDNS0 buffer size for protecting answers from vulnerability of IP flagmentation.
+
+## Description
+*bufsize* can limit requester's UDP payload size.  
+It prevents IP fragmentation so that to be ready for the DNS Flag Day 2020 and to deal with DNS vulnerability.
+
+## Syntax
+```text
+bufsize [SIZE]
+```
+
+**[SIZE]** is an int value for setting the buffer size.  
+The default value is 512, and the value must be within 512 - 4096.  
+Only one argument is acceptable.
+
+## Examples
+```text
+. {
+    bufsize 512
+    forward . 172.31.0.10
+    log
+}
+```
+
+If you run a resolver on 172.31.0.10, the bufsize of incoming query on the resolver will be set as 512 bytes.
+
+## Considerations
+For now, if a client does not use EDNS, this plugin adds OPT RR.  
+This behavior respect the following description on RFC 6891.
+
+https://tools.ietf.org/html/rfc6891#section-6.1.1
+> The OPT RR MAY be placed anywhere within the additional data section.
--- a/plugin/bufsize/README.md
+++ b/plugin/bufsize/README.md
@@ -1,22 +1,22 @@
 # bufsize
 ## Name
-*bufsize* - modifies EDNS0 buffer size for protecting answers from vulnerability of IP flagmentation.
+*bufsize* - sizes EDNS0 buffer size to prevent IP fragmentation.
 
 ## Description
-*bufsize* can limit requester's UDP payload size.  
-It prevents IP fragmentation so that to be ready for the DNS Flag Day 2020 and to deal with DNS vulnerability.
+*bufsize* limits a requester's UDP payload size.  
+It prevents IP fragmentation so that to deal with DNS vulnerability.
 
 ## Syntax
-```text
+```txt
 bufsize [SIZE]
 ```
 
 **[SIZE]** is an int value for setting the buffer size.  
 The default value is 512, and the value must be within 512 - 4096.  
-Only one argument is acceptable.
+Only one argument is acceptable, and it covers both IPv4 and IPv6.
 
 ## Examples
-```text
+```corefile
 . {
     bufsize 512
     forward . 172.31.0.10
@@ -24,11 +24,7 @@ Only one argument is acceptable.
 }
 ```
 
-If you run a resolver on 172.31.0.10, the bufsize of incoming query on the resolver will be set as 512 bytes.
+If you run a resolver on 172.31.0.10, the buffer size of incoming query on the resolver will be set to 512 bytes.
 
 ## Considerations
-For now, if a client does not use EDNS, this plugin adds OPT RR.  
-This behavior respect the following description on RFC 6891.
-
-https://tools.ietf.org/html/rfc6891#section-6.1.1
-> The OPT RR MAY be placed anywhere within the additional data section.
+For now, if a client does not use EDNS, this plugin adds OPT RR.
--- a/plugin/bufsize/bufsize.go
+++ b/plugin/bufsize/bufsize.go
@@ -1,3 +1,4 @@
+// Package bufsize implements a plugin that modifies EDNS0 buffer size.
 package bufsize
 
 import (
@@ -8,7 +9,7 @@ import (
 	"github.com/miekg/dns"
 )
 
-// Bufsize plugin
+// Bufsize implements bufsize plugin.
 type Bufsize struct {
 	Next plugin.Handler
 	Size int
--- a/plugin/bufsize/bufsize.go
+++ b/plugin/bufsize/bufsize.go
@@ -1,0 +1,30 @@
+package bufsize
+
+import (
+	"context"
+
+	"github.com/coredns/coredns/plugin"
+
+	"github.com/miekg/dns"
+)
+
+// Bufsize plugin
+type Bufsize struct {
+	Next plugin.Handler
+	Size int
+}
+
+// ServeDNS implements the plugin.Handler interface.
+func (buf Bufsize) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	if option := r.IsEdns0(); option != nil {
+		option.SetUDPSize(uint16(buf.Size))
+	} else {
+		// If a client does not use EDNS, add it
+		r.SetEdns0(uint16(buf.Size), false)
+	}
+
+	return plugin.NextOrFailure(buf.Name(), buf.Next, ctx, w, r)
+}
+
+// Name implements the Handler interface.
+func (buf Bufsize) Name() string { return "bufsize" }
--- a/plugin/bufsize/bufsize_test.go
+++ b/plugin/bufsize/bufsize_test.go
@@ -1,0 +1,72 @@
+package bufsize
+
+import (
+	"context"
+	"testing"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/test"
+	"github.com/coredns/coredns/plugin/whoami"
+
+	"github.com/miekg/dns"
+)
+
+func TestBufsize(t *testing.T) {
+	em := Bufsize{
+		Size: 512,
+	}
+
+	tests := []struct {
+		next            plugin.Handler
+		qname           string
+		inputBufsize    uint16
+		outgoingBufsize uint16
+		expectedErr     error
+	}{
+		// This plugin is responsible for limiting outgoing query's bufize
+		{
+			next:            whoami.Whoami{},
+			qname:           ".",
+			inputBufsize:    1200,
+			outgoingBufsize: 512,
+			expectedErr:     nil,
+		},
+		// If EDNS is not enabled, this plugin adds it
+		{
+			next:            whoami.Whoami{},
+			qname:           ".",
+			outgoingBufsize: 512,
+			expectedErr:     nil,
+		},
+	}
+
+	for i, tc := range tests {
+		req := new(dns.Msg)
+		req.SetQuestion(dns.Fqdn(tc.qname), dns.TypeA)
+		req.Question[0].Qclass = dns.ClassINET
+		em.Next = tc.next
+
+		if tc.inputBufsize != 0 {
+			req.SetEdns0(tc.inputBufsize, false)
+		}
+
+		_, err := em.ServeDNS(context.TODO(), &test.ResponseWriter{}, req)
+
+		if err != tc.expectedErr {
+			t.Errorf("Test %d: Expected error is %v, but got %v", i, tc.expectedErr, err)
+		}
+
+		if tc.outgoingBufsize != 0 {
+			for _, extra := range req.Extra {
+				if option, ok := extra.(*dns.OPT); ok {
+					b := option.UDPSize()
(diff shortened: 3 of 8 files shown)


plugin/bufsize/bufsize_test.go
```diff
@@ -0,0 +1,72 @@
...
+		req := new(dns.Msg)
+		req.SetQuestion(dns.Fqdn(tc.qname), dns.TypeA)
+		req.Question[0].Qclass = dns.ClassINET
+		em.Next = tc.next
+
+		if tc.inputBufsize != 0 {
+			req.SetEdns0(tc.inputBufsize, false)
+		}
+
+		_, err := em.ServeDNS(context.TODO(), &test.ResponseWriter{}, req)
```
dilyevsky: nit: just `context.Background`

plugin/bufsize/setup.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func bufsizeParse(c *caddy.Controller) (int, error) {
+	for c.Next() {
+		args := c.RemainingArgs()
+		switch len(args) {
+		case 0:
+			// Nothing specified; use 512 as default
+			bufsize := 512
+			return bufsize, nil
```
dilyevsky:
```suggestion
return defaultBufSize, nil
```

Define `const defaultBufSize = 512` and delete line above.

plugin/bufsize/setup.go
```diff
@@ -0,0 +1,56 @@
...
+		case 1:
+			// Specified value is needed to verify
+			bufsize, err := strconv.Atoi(args[0])
+			if err != nil {
+				fmt.Println("Invalid argument in bufsize.")
```
dilyevsky: I think just returning error should be enough here and below.

plugin/bufsize/bufsize.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"context"
+
+	"github.com/coredns/coredns/plugin"
+
+	"github.com/miekg/dns"
+)
+
+// Bufsize plugin
```
dilyevsky: nit: // Bufsize implements bufsize plugin.

plugin/bufsize/bufsize.go
```diff
@@ -0,0 +1,30 @@
+package bufsize
```
dilyevsky, line 2: Missing pack
age docstring.

dilyevsky (commented):
Thanks for the change! The functionality itself appears to be 👌 I have only a few style comments. +r @miekg for new plugin approval.

plugin/bufsize/bufsize.go
ykhr53: Thanks, I'll add it!

plugin/bufsize/bufsize.go
ykhr53: Thanks, I'll add it!

plugin/bufsize/setup.go
ykhr53: It means, I should delete the two fmt.Println (line 40, 45) ?

dilyevsky: Yup, also line 51

ykhr53: OK, thank you for the review! I'll fix them 😊

plugin/bufsize/README.md
```diff
@@ -0,0 +1,34 @@
+# bufsize
+## Name
+*bufsize* - modifies EDNS0 buffer size for protecting answers from vulnerability of IP flagmentation.
```
miekg:
```suggestion
*bufsize* - sizes EDNS0 buffer size to prevent IP fragmentation.
```

plugin/bufsize/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Description
+*bufsize* can limit requester's UDP payload size.
```
miekg:
```suggestion
*bufsize*  limits a requester's UDP payload size.  
```

plugin/bufsize/README.md
```diff
@@ -0,0 +1,34 @@
+# bufsize
+## Name
+*bufsize* - modifies EDNS0 buffer size for protecting answers from vulnerability of IP flagmentation.
+
+## Description
+*bufsize* can limit requester's UDP payload size.  
+It prevents IP fragmentation so that to be ready for the DNS Flag Day 2020 and to deal with DNS vulnerability.
```
miekg: I would drop the mention of 2020 flagday here as it's not that relevant for the user.

plugin/bufsize/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Syntax
+```text
```
miekg:
```suggestion
```txt
```

plugin/bufsize/README.md
```diff
@@ -0,0 +1,34 @@
...
+bufsize [SIZE]
+```
+
+**[SIZE]** is an int value for setting the buffer size.
```
miekg, line 14: shouldn't this take into account ipv6 and ipv4 and things like that? Should we allow a SIZE at all, or just provide the defaults?

plugin/bufsize/README.md
```diff
@@ -0,0 +1,34 @@
...
+```text
+bufsize [SIZE]
+```
+
+**[SIZE]** is an int value for setting the buffer size.  
+The default value is 512, and the value must be within 512 - 4096.  
+Only one argument is acceptable.
+
+## Examples
+```text
```
miekg:
```suggestion
```corefile
```

plugin/bufsize/README.md
```diff
@@ -0,0 +1,34 @@
...
+. {
+    bufsize 512
+    forward . 172.31.0.10
+    log
+}
+```
+
+If you run a resolver on 172.31.0.10, the bufsize of incoming query on the resolver will be set as 512 bytes.
```
miekg:
```suggestion
If you run a resolver on 172.31.0.10, the buffer size of incoming query on the resolver will be set to 512 bytes.
```

plugin/bufsize/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Considerations
+For now, if a client does not use EDNS, this plugin adds OPT RR.  
+This behavior respect the following description on RFC 6891.
```
miekg:
```suggestion
This behavior respects the following description on RFC 6891.
```

miekg: add the section here as well, RFC 6891 Section 6.1.1

plugin/bufsize/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+https://tools.ietf.org/html/rfc6891#section-6.1.1
```
miekg:
```suggestion

```

plugin/bufsize/README.md
```diff
@@ -0,0 +1,34 @@
...
+> The OPT RR MAY be placed anywhere within the additional data section.
```
miekg: on second though, this is not that relevant, I would remove the entire section.

plugin/bufsize/bufsize.go
```diff
@@ -0,0 +1,31 @@
...
+// Bufsize implements bufsize plugin.
+type Bufsize struct {
+	Next plugin.Handler
+	Size int
+}
+
+// ServeDNS implements the plugin.Handler interface.
+func (buf Bufsize) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	if option := r.IsEdns0(); option != nil {
+		option.SetUDPSize(uint16(buf.Size))
```
miekg, line 21: does this do the right thing for the DO bit? It keep it, if it's there?

plugin/bufsize/setup.go
```diff
@@ -0,0 +1,52 @@
...
+
+	dnsserver.GetConfig(c).AddPlugin(func(next plugin.Handler) plugin.Handler {
+		return Bufsize{Next: next, Size: bufsize}
+	})
+
+	return nil
+}
+
+func bufsizeParse(c *caddy.Controller) (int, error) {
+	const defaultBufSize = 512
```
miekg, line 28: this const is only use once - just use the 512 value (or whatever we come up with later) directly

miekg (changes requested):
The core of this plugin lgtm and I suspect it would be that small. A whole bunch of other comments and a big one on how to deal with the various network families.

plugin/bufsize/setup.go
ykhr53: Thank you for the comment.
This const is from https://github.com/coredns/coredns/pull/3401#discussion_r340229081 .
Which one is better?

plugin/bufsize/README.md
ykhr53: Thanks, I'll fix it.

plugin/bufsize/README.md
ykhr53: Thanks, I'll fix it.

plugin/bufsize/README.md
ykhr53: Thanks, I'll fix it.

plugin/bufsize/README.md
ykhr53: Thanks, I'll fix it.

plugin/bufsize/README.md
ykhr53: Thanks, I'll fix it.

plugin/bufsize/README.md
ykhr53: ok, I agree with that. I'll drop it.

plugin/bufsize/README.md
ykhr53: ok, I'll drop line 31 - 34, it would be fine!

plugin/bufsize/bufsize.go
ykhr53: yes, SetUDPSize() doesn't touch DO bit. It keeps original DO bit value.
https://github.com/miekg/dns/blob/master/edns.go#L127

Also I have tested and confirmed by pcap.

plugin/bufsize/README.md
ykhr53: In the Flag Day page, the recommended value is 1232 bytes, it covers both IPv4 and IPv6.
So I decided to accept one option regardless of address family. And it's simple too.
But adding option to classify by an address family could be useful!
If you think this feature is needed, I will implement it. Or could be a feature request for future?

miekg: did it, OK. Let's just leave one - we have to make this clear in the docs though

plugin/bufsize/setup.go
miekg: just leave it then, no reason to ping-ping between reviewers and @dilyevsky knows his stuff as well.

plugin/bufsize/README.md
ykhr53: Thank you for the review! I'll mention it on **Syntax** section in README. How about something like that?
"Only one argument is acceptable, and it covers both IPv4 and IPv6."

plugin/bufsize/setup.go
ykhr53: Thanks! ok I'll leave it.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add bufsize p..." ]
>ykhr53 commented on this pull request.
>
>
>
>> @@ -0,0 +1,34 @@
>+# bufsize
>+## Name
>+*bufsize* - modifies EDNS0 buffer size for protecting answers from vulnerability of IP flagmentation.
>+
>+## Description
>+*bufsize* can limit requester's UDP payload size.
>+It prevents IP fragmentation so that to be ready for the DNS Flag Day 2020 and to deal with DNS vulnerability.
>+
>+## Syntax
>+```text
>+bufsize [SIZE]
>+```
>+
>+**[SIZE]** is an int value for setting the buffer size.
>
>Thank you for the review! I'll mention it on **Syntax** section in README. How about something like that?
>"Only one argument is acceptable, and it covers both IPv4 and IPv6."

yes, something like that sgtm

ykhr53 (commented):
Thank you for reviewing! I fixed some comments on README, so please re-review it. thanks :)

plugin/bufsize/setup.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	dnsserver.GetConfig(c).AddPlugin(func(next plugin.Handler) plugin.Handler {
+		return Bufsize{Next: next, Size: bufsize}
+	})
+
+	return nil
+}
+
+func bufsizeParse(c *caddy.Controller) (int, error) {
```
miekg: nit: just call this `parse`

miekg (commented):
one nit, lgtm  otherwise

ykhr53 (commented):
you mean, need to change the function name to `parse` from `bufsizeParse`?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add bufsize p..." ]
>ykhr53 commented on this pull request.
>
>you mean, need to change the function name to `parse` from `bufsizeParse`?

yes, it's already namespaced to your package name anyway

ykhr53:
ok, thanks!
I've just followed some other plugins like `nsidParse` or `cacheParse`.
But I can see some plugins use `parse`, so I'll change to `parse`!

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add bufsize p..." ]
>ok, thanks!
>I've just followed some other plugins like `nsidParse` or `cacheParse`.

yeah, those should just use 'parse' as well, but c'est la vie.

miekg (approved):

PR #3406, opened 2019-10-27
Correct some if/iff mixups

euank:
This is a fixup for https://github.com/coredns/coredns/pull/3310 which
replaced some 'iff's with 'if' under the assumption they were typos.

I'm fairly confident they were "If and only if"
(https://en.wikipedia.org/wiki/If_and_only_if), which is commonly
shortened as "iff".

I believe the shortening is common enough while programming that it's
reasonable to keep them as they were rather than expand them to 'if and
only if' to avoid any other confusion, but I could see it either way.


chrisohaver:
Maybe spell out "if and only if"  to avoid further PR churn.

stp-ip:
As a non native speaker I would have thought they were a typo and wouldn't have understood the `iff` writing. So I vote for spelling it out.

euank:
Thanks, that's good to know. Will do!

PR #3419, opened 2019-11-04
plugin/k8s_external: zone transfers

chrisohaver:

Implements `transfer.Transferer` interface to enable zone transfers.
Add zone serial to for external k8s zones.
Separately track zone serial for internal and external ip services in _kubernetes_ plugin.

none

Included

no

### DIFF
--- a/plugin/kubernetes/external.go
+++ b/plugin/kubernetes/external.go
@@ -87,8 +87,30 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
 func (k *Kubernetes) ExternalAddress(state request.Request) []dns.RR {
 	// If CoreDNS is running inside the Kubernetes cluster: k.nsAddrs() will return the external IPs of the services
 	// targeting the CoreDNS Pod.
-	// If CoreDNS is running outside of the Kubernetes cluster: k.nsAddrs() will return the first non-loopback IP
-	// address seen on the local system it is running on. This could be the wrong answer if coredns is using the *bind*
-	// plugin to bind to a different IP address.
+	// If CoreDNS is running outside of the Kubernetes cluster: k.nsAddrs() will return the non-loopback IPs
+	// that CoreDNS is listening on.
 	return k.nsAddrs(true, state.Zone)
+}
+
+// ExternalServices returns all services with external IPs
+func (k *Kubernetes) ExternalServices(zone string) (services []msg.Service) {
+	zonePath := msg.Path(zone, coredns)
+	for _, svc := range k.APIConn.ServiceList() {
+		for _, ip := range svc.ExternalIPs {
+			for _, p := range svc.Ports {
+				s := msg.Service{Host: ip, Port: int(p.Port), TTL: k.ttl}
+				s.Key = strings.Join([]string{zonePath, svc.Namespace, svc.Name}, "/")
+				services = append(services, s)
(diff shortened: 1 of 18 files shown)


plugin/kubernetes/handler_test.go
```diff
@@ -525,7 +524,8 @@ func (APIConnServeTest) Run()                                      {}
 func (APIConnServeTest) Stop() error                               { return nil }
 func (APIConnServeTest) EpIndexReverse(string) []*object.Endpoints { return nil }
 func (APIConnServeTest) SvcIndexReverse(string) []*object.Service  { return nil }
-func (APIConnServeTest) Modified() int64                           { return time.Now().Unix() }
+func (APIConnServeTest) Modified() int64                           { return 2 }
+func (APIConnServeTest) ExtModified() int64                        { return 2 }
```
stp-ip, line 528: Reason for the changes?

plugin/kubernetes/external.go
```diff
@@ -87,8 +87,30 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
...
-	// address seen on the local system it is running on. This could be the wrong answer if coredns is using the *bind*
-	// plugin to bind to a different IP address.
+	// If CoreDNS is running outside of the Kubernetes cluster: k.nsAddrs() will return the non-loopback IPs
+	// that CoreDNS is listening on.
 	return k.nsAddrs(true, state.Zone)
 }
+
+// ExternalServices returns all services with external IPs
+func (k *Kubernetes) ExternalServices(zone string) (services []msg.Service) {
+	zonePath := msg.Path(zone, coredns)
```
stp-ip, line 97: Still wrapping my head around the whole structure, but shouldn't that be:
```suggestion
	zonePath := msg.Path(zone, "coredns")
```

stp-ip: Same thing with https://github.com/coredns/coredns/pull/3419/files#diff-7b62412de4a07f4b4ad2d746cd77c878R59.
Or what am I missing here?

stp-ip (changes requested):
Looks good. A few comments and questions.
Maybe @johnbelamaric can take a final look as well.

Haven't tested this manually.

plugin/kubernetes/handler_test.go
chrisohaver: For reference: For testing IXFR and IXFR->AXFR fallback, same thing discussed here: https://github.com/coredns/coredns/pull/3452/files#r417266369

plugin/kubernetes/external.go
chrisohaver: It's a constant. From kubernetes.go: `const coredns = "c" // used as a fake key prefix in msg.Service`

This is an artifact of the etcd-styled "Backend" integration in the kubernetes plugin, where records are passed to the back-end in etcd msg structs.

stp-ip: Makes sense. What tripped me up was the different usage within the kubernetes PR: 
https://github.com/coredns/coredns/pull/3452/files#diff-82592195683029ae911713a3c6a581fdR84

chrisohaver: Oh - good catch.  [That](https://github.com/coredns/coredns/pull/3452/files#diff-82592195683029ae911713a3c6a581fdR84) looks like a copy-forward bug from the original kubernetes transfer implementation.  I think it's harmless, but should be fixed there to be consistent with the rest of the k8s code.

stp-ip: Sounds good.

stp-ip (approved):

PR #3302, opened 2019-09-24
plugin: cleanup code based on staticcheck warnings

beautytiger:
### 1. Why is this pull request needed and what does it do?
cleanup codebase, reduce staticcheck warnings

### 2. Which issues (if any) are related?
No
### 3. Which documentation changes (if any) need to be made?
No
### 4. Does this introduce a backward incompatible change or deprecation?
No

PR #3433, opened 2019-11-07
Fix: Call wg.Add in main goroutine to avoid race conditons.

beautytiger:
If the goroutine call wg.Add by itself, what if the main goroutine run to the end before goroutine start runing? May there are chances.

No

### DIFF
--- a/plugin/transfer/transfer.go
+++ b/plugin/transfer/transfer.go
@@ -109,8 +109,8 @@ func (t Transfer) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	ch := make(chan *dns.Envelope)
 	tr := new(dns.Transfer)
 	wg := new(sync.WaitGroup)
+	wg.Add(1)
 	go func() {
-		wg.Add(1)
 		tr.Out(w, r, ch)
 		wg.Done()
 	}()


miekg:
how does this relate to the docs?

~~~
    40  // Add adds delta, which may be negative, to the WaitGroup counter.
    41  // If the counter becomes zero, all goroutines blocked on Wait are released.
    42  // If the counter goes negative, Add panics.
    43  //
    44  // Note that calls with a positive delta that occur when the counter is zero
    45  // must happen before a Wait. Calls with a negative delta, or calls with a
    46  // positive delta that start when the counter is greater than zero, may happen
    47  // at any time.
    48  // Typically this means the calls to Add should execute before the statement
    49  // creating the goroutine or other event to be waited for.
    50  // If a WaitGroup is reused to wait for several independent sets of events,
    51  // new Add calls must happen after all previous Wait calls have returned.
    52  // See the WaitGroup example.
~~~

beautytiger:
I think it would be this:
```
// Typically this means the calls to Add should execute before the statement
// creating the goroutine or other event to be waited for.
```
call the wg.Add before creating the goroutine.

chrisohaver:
The same change to `file/xfr.go` (from where this code was originally copied from) was merged in #3302.

chrisohaver (approved):
LGTM, especially since the same change was merged to the original code from where this was copied.  Referencing #3302 would have made this more immediately obvious.

PR #3437, opened 2019-11-10
plugin/bufsize: add usecase description

ykhr53:
the *bufsize* plugin was just added to coredns repo though, I realised the documentation could be better.
I added a usecase on examples section when coredns behaves as an authoritative nameserver.
It would be useful to explain the concept of this plugin precisely.

None.

None.

None.

### DIFF
--- a/plugin/bufsize/README.md
+++ b/plugin/bufsize/README.md
@@ -16,6 +16,7 @@ The default value is 512, and the value must be within 512 - 4096.
 Only one argument is acceptable, and it covers both IPv4 and IPv6.
 
 ## Examples
+### As a forwarder
 ```corefile
 . {
     bufsize 512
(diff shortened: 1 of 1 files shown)


plugin/bufsize/README.md
```diff
@@ -16,6 +16,7 @@ The default value is 512, and the value must be within 512 - 4096.
 Only one argument is acceptable, and it covers both IPv4 and IPv6.
 
 ## Examples
+### As a forwarder
```
miekg: just drop this subsection and add the usecase to the prose. Also I think it makes sense to say you want a larger bufsize instead of the 512 default, but then also warn about fragmentation, maybe (again?)

ykhr53: Thank you for reviewing!
Ok, I'll drop subsection and use prose style.
Also, I agree with mentioning about larger bufsize consideration. And I think it would be better to mention it on **Considerations** section.

rajansandeep (approved):

Issue #3012, opened 2019-07-16
Unable to resolve Route 53 Alias Records


wazoo:
I am using the following: CoreDNS-1.5.0 - linux/amd64, go1.12.2, e3f9a80

I am using CoreDNS on a regular linux machine in a docker container to provide some DNS routing and split horizon for a hybrid AWS and on-prem infrastructure.  Normal DNS resolution works great however when I try to resolve an [AWS Alias Record](https://aws.amazon.com/premiumsupport/knowledge-center/route-53-create-alias-records/) I get NXDOMAIN.

I have tried adding a forward and fallthrough to the zone config (shown below) and still get NXDOMAIN but querying the VPC resolver for the same record works fine.

I have all of the zones configured as follows:

```
(base) {
  log
  errors
}
(cache) {
  cache 30 {
    success 500 90
    denial 500
  }
}
sub.domain.com.:53 {
  import base
  import cache
  route53 sub.domain.com.:<private zone ID> sub.domain.com.io.:<public zone ID>  {
      import account-credential-file
  }
}
```
If I have a record (as an example) `record.sub.domain.com` that is an A record or a CNAME resolution works as expected and private records override public records.  

If I have a record (as an example) `alias.sub.domain.com` I receive the following:

```
2019-07-16T20:52:55.952Z [INFO] 172.17.0.1:35996 - 9166 "A IN alias.sub.domain.com. udp 48 false 4096" NXDOMAIN qr,aa,rd 134 0.0002356s
```

wazoo:
I did some additional testing and I was able to get it to work with the following config:

```
(base) {
  log
  errors
}
(cache) {
  cache 30 {
    success 500 90
    denial 500
  }
}
sub.domain.com.:53 {
  import base
  import cache
  route53 sub.domain.com.:<private zone ID> sub.domain.com.io.:<public zone ID>  {
      import account-credential-file
      fallthrough
  }
  forward <vpc-resolver>:53
}
```

I am not sure if this is just missing from the documentation or I missread it but I am not sure why I need to specifically forward every zone.  I don't have a root forwarder configured on these containers, if I did that would this have the same effect?

sharmaansh21:
@wazoo I will debug into it, is it possible for you to provide more details. or maybe you can send it to me sharmaanshul2102@gmail.com if you don't want to share here. 

Exact config will be helpful for debugging.

ykhr53:
Similar behavior is happened on my side too.

[...]

PR #3442, opened 2019-11-13
plugin/route53: Add support for aliases

ilijamt:

Allows us to use aliases in route 53
https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-values-alias.html
https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-choosing-alias-non-alias.html

Based on the documentation ALIAS records can only return A/AAAA, and the TTL is fixed at 60 seconds.
Possible options for health-checks for ALIAS are 15 second and 30 seconds.

https://github.com/coredns/coredns/issues/3012

Added a new configuration option

No

I do have some questions on the best way to implement the DNS lookups, I took a look at the Cache plugin which actually does something similar to what I need, the ability to do the query lookups without being in the ServeDNS function.

Are there any thought on how to implement this?
@yongtang @dilyevsky

I've implemented a working version by querying the authoritative server for the domain in question.

### DIFF
--- a/plugin/route53/alias_resolver.go
+++ b/plugin/route53/alias_resolver.go
@@ -1,0 +1,39 @@
+package route53
+
+import (
+	"context"
+	"github.com/miekg/dns"
+)
+
+// AliasResolver helps us in resolving ALIAS records from AWS.
+type AliasResolver interface {
+	Resolve(ctx context.Context, dnsName string, dnsType uint16, ns string, ttl int64) (rrs []dns.RR, err error)
+}
+
+type authoritativeNsResolver struct{}
+
+func (anr authoritativeNsResolver) Resolve(ctx context.Context, dnsName string, dnsType uint16, ns string, ttl int64) (rrs []dns.RR, err error) {
+	client := new(dns.Client)
+	req := new(dns.Msg)
+
+	req.RecursionDesired = true
+	req.SetQuestion(dnsName, dnsType)
+
+	r, _, err := client.ExchangeContext(ctx, req, ns)
+	if err != nil {
+		return
+	}
+
+	for _, resp := range r.Answer {
+		var rec dns.RR
+		switch response := resp.(type) {
+		case *dns.A:
+			rec, _ = remapDnsAliasRR(dnsName, response, ttl)
+		case *dns.AAAA:
+			rec, _ = remapDnsAliasRR(dnsName, response, ttl)
+		}
+		rrs = append(rrs, rec)
+	}
+
+	return
+}
--- a/plugin/route53/alias_resolver_test.go
+++ b/plugin/route53/alias_resolver_test.go
@@ -1,0 +1,66 @@
+package route53
+
+import (
+	"context"
+	"fmt"
+	"github.com/miekg/dns"
+	"net"
+	"testing"
+	"time"
+)
+
+func TestAuthoritativeNsResolver_Resolve(t *testing.T) {
+	resolver := authoritativeNsResolver{}
+
+	dnsAddr := "127.0.0.1:31726"
+	server := &dns.Server{
+		Addr:      dnsAddr,
+		Net:       "udp",
+		ReusePort: true,
+	}
+
+	dns.HandleFunc("example.dev.", func(w dns.ResponseWriter, r *dns.Msg) {
+		q := r.Question[0]
+		m := new(dns.Msg)
+		m.SetReply(r)
+		var rr dns.RR
+		switch q.Qtype {
+		case dns.TypeA:
+			rr = &dns.A{
+				Hdr: dns.RR_Header{Name: "example.dev.", Rrtype: dns.TypeA, Class: dns.ClassINET, Ttl: 60},
+				A:   net.ParseIP("10.0.0.0"),
+			}
+		case dns.TypeAAAA:
+			rr = &dns.AAAA{
+				Hdr:  dns.RR_Header{Name: "example.dev.", Rrtype: dns.TypeAAAA, Class: dns.ClassINET, Ttl: 60},
+				AAAA: net.ParseIP("fdf8:f53b:82e4::53"),
+			}
+		}
+		m.Answer = append(m.Answer, rr)
+		_ = w.WriteMsg(m)
+	})
+
+	go func() {
+		err := server.ListenAndServe()
+		fmt.Println(err)
+	}()
+	time.Sleep(time.Millisecond * 50) // so we can start the server
+
+	_, err := resolver.Resolve(context.Background(), "test.example.dev.", dns.TypeA, dnsAddr, 60)
+	if err != nil {
+		t.Fatalf("Failed to resolve A record: %v", err)
+	}
+
+	_, err = resolver.Resolve(context.Background(), "test.example.dev.", dns.TypeAAAA, dnsAddr, 60)
+	if err != nil {
+		t.Fatalf("Failed to resolve AAAA record: %v", err)
+	}
+}
+
+func TestAuthoritativeNsResolver_ResolveInvalidPort(t *testing.T) {
+	resolver := authoritativeNsResolver{}
+	_, err := resolver.Resolve(context.Background(), "test.example.dev.", dns.TypeA, "127.0.0.1:8054", 60)
+	if err == nil {
+		t.Fatalf("Why did it resolve the record, should have failed")
+	}
+}
--- a/plugin/route53/route53.go
+++ b/plugin/route53/route53.go
@@ -33,24 +33,36 @@ type Route53 struct {
 	upstream  *upstream.Upstream
 	refresh   time.Duration
 
+	aliasRefresh  time.Duration
+	aliasTTL      int64
+	aliasResolver AliasResolver
+
 	zMu   sync.RWMutex
 	zones zones
 }
 
 type zone struct {
-	id  string
-	z   *file.Zone
-	dns string
+	id      string
+	z       *file.Zone
+	dns     string
+	aliases map[string]aliasRecord
 }
 
 type zones map[string][]*zone
+
+type aliasRecord struct {
+	dnsName     string
+	dnsType     uint16
+	dnsTypeName string
+	dnsZoneId   string
+}
 
 // New reads from the keys map which uses domain names as its key and hosted
 // zone id lists as its values, validates that each domain name/zone id pair
 // does exist, and returns a new *Route53. In addition to this, upstream is use
 // for doing recursive queries against CNAMEs. Returns error if it cannot
 // verify any given domain name/zone id pair.
-func New(ctx context.Context, c route53iface.Route53API, keys map[string][]string, refresh time.Duration) (*Route53, error) {
+func New(ctx context.Context, c route53iface.Route53API, keys map[string][]string, refresh time.Duration, aliasResolver AliasResolver) (*Route53, error) {
 	zones := make(map[string][]*zone, len(keys))
 	zoneNames := make([]string, 0, len(keys))
 	for dns, hostedZoneIDs := range keys {
@@ -69,11 +81,14 @@ func New(ctx context.Context, c route53iface.Route53API, keys map[string][]strin
 		}
 	}
 	return &Route53{
-		client:    c,
-		zoneNames: zoneNames,
-		zones:     zones,
-		upstream:  upstream.New(),
-		refresh:   refresh,
+		client:        c,
+		zoneNames:     zoneNames,
+		zones:         zones,
+		upstream:      upstream.New(),
+		refresh:       refresh,
+		aliasRefresh:  time.Minute,
+		aliasTTL:      60,
+		aliasResolver: aliasResolver,
 	}, nil
 }
 
@@ -89,6 +104,17 @@ func (h *Route53) Run(ctx context.Context) error {
 			case <-ctx.Done():
 				log.Infof("Breaking out of Route53 update loop: %v", ctx.Err())
 				return
+			case <-time.After(h.aliasRefresh):
+				if h.aliasRefresh >= h.refresh {
+					// we don't need to do anything as the refresh period is lower than the alias refresh
+					// period, so we can skip the extra call to update the aliases
+					continue
+				}
+				// make this auto adjustable so we can adjust based on the TTL,
+				// health-checks can be 15, 30 or 60 seconds per AWS so worst case this has to execute every 15 seconds
+				if err := h.updateAliases(ctx); err != nil && ctx.Err() == nil /* Don't log error if ctx expired. */ {
+					log.Errorf("Failed to update aliases: %v", err)
+				}
 			case <-time.After(h.refresh):
 				if err := h.updateZones(ctx); err != nil && ctx.Err() == nil /* Don't log error if ctx expired. */ {
 					log.Errorf("Failed to update zones: %v", err)
@@ -96,6 +122,88 @@ func (h *Route53) Run(ctx context.Context) error {
 			}
 		}
 	}()
+	return nil
+}
+
+func updateAliasesForZone(ctx context.Context, zMu sync.RWMutex, zone *zone, resolver AliasResolver, aliasTTL int64) (err error) {
+	for name, alias := range zone.aliases {
+		// remove old records from the hosted zone file if they exist
+		zMu.RLock()
+		records, found := zone.z.Search(name)
+		zMu.RUnlock()
+		if found {
+			// we need to remove all of them so the next time we add it will be the new ones
+			zMu.Lock()
+			for _, record := range records.All() {
+				zone.z.Delete(record)
+			}
+			zMu.Unlock()
+		}
+
+		// check to make sure that we don't host the zone first, if we do
+		// we need to return the alias from our zone, no need for an external lookup
+		if zone.id == alias.dnsZoneId {
+			records, found := zone.z.Search(alias.dnsName)
+			if found {
+				for _, record := range records.All() {
+					rec, _ := remapDnsAliasRR(name, record, int64(record.Header().Ttl))
+					zMu.Lock()
+					_ = zone.z.Insert(rec)
+					zMu.Unlock()
+				}
+			}
+			continue // we are finished, onto the next alias
+		}
+
+		// go through all the ns servers, before giving up on the resolving
+		for _, ns := range zone.z.NS {
+			var rrs []dns.RR
+			nameserver := fmt.Sprintf("%s:53", ns.(*dns.NS).Ns)
+			rrs, err = resolver.Resolve(ctx, name, alias.dnsType, nameserver, aliasTTL)
+			if err != nil {
+				continue // we skip and continue onto the next one until we exhaust all of them
+			}
+			for _, rr := range rrs {
+				zMu.Lock()
+				_ = zone.z.Insert(rr)
+				zMu.Unlock()
+			}
+			break // we finished successfully no need to go over the next nameserver
+		}
+	}
+	return err
+}
+
+func (h *Route53) updateAliases(ctx context.Context) error {
+	errc := make(chan error)
+	defer close(errc)
+
+	for zName, z := range h.zones {
+		go func(zName string, z []*zone) {
+			var err error
+			defer func() {
+				errc <- err
+			}()
+			for _, hostedZone := range z {
+				err = updateAliasesForZone(ctx, h.zMu, hostedZone, h.aliasResolver, h.aliasTTL)
+			}
+		}(zName, z)
+	}
+
+	// Collect errors (if any). This will also sync on all zones updates
+	// completion.
+	var errs []string
+	for i := 0; i < len(h.zones); i++ {
+		err := <-errc
+		if err != nil {
+			errs = append(errs, err.Error())
+		}
+	}
+
+	if len(errs) != 0 {
+		return fmt.Errorf("errors updating aliases: %v", errs)
+	}
+
 	return nil
 }
 
@@ -207,25 +315,61 @@ func maybeUnescape(s string) (string, error) {
 	}
 }
 
-func updateZoneFromRRS(rrs *route53.ResourceRecordSet, z *file.Zone) error {
+func rrFromRR(name string, dnsType string, ttl int64, record route53.ResourceRecord) (dns.RR, error) {
+
+	v, err := maybeUnescape(aws.StringValue(record.Value))
+	if err != nil {
+		return nil, fmt.Errorf("failed to unescape `%s' value: %v", aws.StringValue(record.Value), err)
+	}
+
+	rfc1035 := fmt.Sprintf("%s %d IN %s %s", name, ttl, dnsType, v)
+	r, err := dns.NewRR(rfc1035)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse resource record: %v", err)
+	}
+	return r, nil
+}
+
+func remapDnsAliasRR(name string, record dns.RR, ttl int64) (dns.RR, error) {
+	var val string
+	var dnsTypeName string
+	switch r := record.(type) {
+	case *dns.A:
+		val = r.A.String()
+		dnsTypeName = "A"
+	case *dns.AAAA:
+		val = r.AAAA.String()
+		dnsTypeName = "AAAA"
+	}
+
+	return rrFromRR(name, dnsTypeName, ttl, route53.ResourceRecord{Value: &val})
+}
+
+func updateZoneFromRRS(rrs *route53.ResourceRecordSet, z *file.Zone, alias map[string]aliasRecord) error {
+	n, err := maybeUnescape(aws.StringValue(rrs.Name))
+	if err != nil {
+		return fmt.Errorf("failed to unescape `%s' name: %v", aws.StringValue(rrs.Name), err)
+	}
+
+	if rrs.AliasTarget != nil {
+		var dnsType uint16
+		var dnsName = aws.StringValue(rrs.AliasTarget.DNSName)
+		switch aws.StringValue(rrs.Type) {
+		case "A":
+			dnsType = dns.TypeA
+		case "AAAA":
+			dnsType = dns.TypeAAAA
+		default:
+			return fmt.Errorf("failed to process alias record for %v => %v from route53: %v", n, dnsName, err)
+		}
+		alias[n] = aliasRecord{dnsName: dnsName, dnsType: dnsType, dnsTypeName: aws.StringValue(rrs.Type), dnsZoneId: aws.StringValue(rrs.AliasTarget.HostedZoneId)}
+	}
+
 	for _, rr := range rrs.ResourceRecords {
-
-		n, err := maybeUnescape(aws.StringValue(rrs.Name))
+		r, err := rrFromRR(n, aws.StringValue(rrs.Type), aws.Int64Value(rrs.TTL), *rr)
 		if err != nil {
-			return fmt.Errorf("failed to unescape `%s' name: %v", aws.StringValue(rrs.Name), err)
+			return err
 		}
-		v, err := maybeUnescape(aws.StringValue(rr.Value))
-		if err != nil {
-			return fmt.Errorf("failed to unescape `%s' value: %v", aws.StringValue(rr.Value), err)
-		}
-
-		// Assemble RFC 1035 conforming record to pass into dns scanner.
-		rfc1035 := fmt.Sprintf("%s %d IN %s %s", n, aws.Int64Value(rrs.TTL), aws.StringValue(rrs.Type), v)
-		r, err := dns.NewRR(rfc1035)
-		if err != nil {
-			return fmt.Errorf("failed to parse resource record: %v", err)
-		}
-
 		z.Insert(r)
 	}
 	return nil
@@ -252,10 +396,11 @@ func (h *Route53) updateZones(ctx context.Context) error {
 					HostedZoneId: aws.String(hostedZone.id),
 					MaxItems:     aws.String("1000"),
 				}
+				alias := map[string]aliasRecord{}
 				err = h.client.ListResourceRecordSetsPagesWithContext(ctx, in,
 					func(out *route53.ListResourceRecordSetsOutput, last bool) bool {
 						for _, rrs := range out.ResourceRecordSets {
-							if err := updateZoneFromRRS(rrs, newZ); err != nil {
+							if err := updateZoneFromRRS(rrs, newZ, alias); err != nil {
 								// Maybe unsupported record type. Log and carry on.
 								log.Warningf("Failed to process resource record set: %v", err)
 							}
@@ -266,8 +411,10 @@ func (h *Route53) updateZones(ctx context.Context) error {
 					err = fmt.Errorf("failed to list resource records for %v:%v from route53: %v", zName, hostedZone.id, err)
 					return
 				}
+
 				h.zMu.Lock()
 				(*z[i]).z = newZ
+				(*z[i]).aliases = alias
 				h.zMu.Unlock()
 			}
 
@@ -285,6 +432,8 @@ func (h *Route53) updateZones(ctx context.Context) error {
 	if len(errs) != 0 {
 		return fmt.Errorf("errors updating zones: %v", errs)
 	}
+
+	_ = h.updateAliases(ctx)
 	return nil
 }
 
--- a/plugin/route53/route53_test.go
+++ b/plugin/route53/route53_test.go
@@ -79,7 +79,7 @@ func (fakeRoute53) ListResourceRecordSetsPagesWithContext(_ aws.Context, in *rou
 func TestRoute53(t *testing.T) {
 	ctx := context.Background()
 
-	r, err := New(ctx, fakeRoute53{}, map[string][]string{"bad.": {"0987654321"}}, time.Minute)
+	r, err := New(ctx, fakeRoute53{}, map[string][]string{"bad.": {"0987654321"}}, time.Minute, nil)
 	if err != nil {
 		t.Fatalf("Failed to create Route53: %v", err)
 	}
@@ -87,7 +87,7 @@ func TestRoute53(t *testing.T) {
 		t.Fatalf("Expected errors for zone bad.")
 	}
 
-	r, err = New(ctx, fakeRoute53{}, map[string][]string{"org.": {"1357986420", "1234567890"}, "gov.": {"Z098765432", "1234567890"}}, 90*time.Second)
+	r, err = New(ctx, fakeRoute53{}, map[string][]string{"org.": {"1357986420", "1234567890"}, "gov.": {"Z098765432", "1234567890"}}, 90*time.Second, nil)
 	if err != nil {
 		t.Fatalf("Failed to create Route53: %v", err)
 	}
--- a/plugin/route53/setup.go
+++ b/plugin/route53/setup.go
@@ -125,7 +125,7 @@ func setup(c *caddy.Controller) error {
 		})
 		client := f(credentials.NewChainCredentials(providers))
 		ctx := context.Background()
-		h, err := New(ctx, client, keys, refresh)
+		h, err := New(ctx, client, keys, refresh, authoritativeNsResolver{})
 		if err != nil {
 			return plugin.Error("route53", c.Errf("failed to create Route53 plugin: %v", err))
 		}
--- a/plugin/route53/route53_alias_test.go
+++ b/plugin/route53/route53_alias_test.go
@@ -1,0 +1,194 @@
+package route53
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/coredns/coredns/plugin/pkg/fall"
+	"github.com/coredns/coredns/plugin/test"
+	crequest "github.com/coredns/coredns/request"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/request"
+	"github.com/aws/aws-sdk-go/service/route53"
+	"github.com/aws/aws-sdk-go/service/route53/route53iface"
+	"github.com/miekg/dns"
+)
+
+type fakeAliasResolver struct{}
+
+func (far fakeAliasResolver) Resolve(ctx context.Context, dnsName string, dnsType uint16, ns string, ttl int64) (rrs []dns.RR, err error) {
+	val := "1.2.3.4"
+	rr, _ := rrFromRR(dnsName, "A", ttl, route53.ResourceRecord{Value: &val})
+	rrs = append(rrs, rr)
+	return
+}
+
+type fakeRoute53Alias struct {
+	route53iface.Route53API
+}
+
+func (fakeRoute53Alias) ListHostedZonesByNameWithContext(_ aws.Context, input *route53.ListHostedZonesByNameInput, _ ...request.Option) (*route53.ListHostedZonesByNameOutput, error) {
+	return nil, nil
+}
+
+func (fakeRoute53Alias) ListResourceRecordSetsPagesWithContext(_ aws.Context, in *route53.ListResourceRecordSetsInput, fn func(*route53.ListResourceRecordSetsOutput, bool) bool, _ ...request.Option) error {
+	if aws.StringValue(in.HostedZoneId) == "0987654321" {
+		return errors.New("bad. zone is bad")
+	}
+	rrsResponse := map[string][]*route53.ResourceRecordSet{}
+	for _, r := range []struct {
+		rType, name, value, hostedZoneID string
+		aliasDnsName, aliasHostedZoneId  string
+		alias                            bool
+	}{
+		{"A", "alias.example.dev.", "1.2.3.4", "1234567891", "random.alias.address.", "3234567891", true},
+		{"A", "same-zone-alias.example.dev.", "4.3.2.1", "1234567891", "target.example.dev.", "1234567891", true},
+		{"A", "example.dev.", "1.2.3.4", "1234567891", "", "", false},
+		{"A", "target.example.dev.", "4.3.2.1", "1234567891", "", "", false},
+		{"A", "target.example.dev.", "8.7.6.5", "1234567891", "", "", false},
+		{"SOA", "dev.", "ns-1536.awsdns-00.co.uk. awsdns-hostmaster.amazon.com. 1 7200 900 1209600 86400", "1234567891", "", "", false},
+		{"NS", "dev.", "ns-1536.awsdns-00.co.uk", "1234567891", "", "", false},
+		// ALIAS record
+	} {
+		rrs, ok := rrsResponse[r.hostedZoneID]
+		if !ok {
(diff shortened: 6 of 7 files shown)


plugin/route53/route53.go
```diff
@@ -99,6 +111,62 @@ func (h *Route53) Run(ctx context.Context) error {
...
+
+	for zName, z := range h.zones {
+		go func(zName string, z []*zone) {
+			var err error
+			defer func() {
+				errc <- err
+			}()
+			for _, hostedZone := range z {
+				for name, alias := range hostedZone.aliases {
+					var r, req *dns.Msg
```
ilijamt: I'm interested in this block on how to properly do the query. as we will need to query the authoritative domain for ALIAS. 

One example is if we have an alias like 

```
alias.example.com ALIAS d21v7mkn0v2jz6.cloudfront.net.
```
The DNS server will preform the lookup for us if we query for the record directly from AWS, but in our case, we need to query an external resolver to get the correct IP addresses for **d21v7mkn0v2jz6.cloudfront.net**

Or we can query the authoritative DNS server directly for the A/AAAA records for **alias.example.com** instead of us trying to resolve the alias address, as we have the NS records for the zone. (This one is simpler and easier to implement)

Any thought on both options?  @yongtang @dilyevsky

plugin/route53/README.md
```diff
@@ -57,6 +57,12 @@ route53 [ZONE:HOSTED_ZONE_ID...] {
 
 *   **DURATION** A duration string. Defaults to `1m`. If units are unspecified, seconds are assumed.
 
+*   `alias_refresh` can be used to control how often do we refresh the aliases from Route 53. Each alias
```
miekg: on a first impression, this looks odd that it needs to be called out explicitly.

What's a standalone test you can add that fails with the current impl. but works with your pr?

yongtang: The TTL of alias in route53 follows the source TTL. From that standpoint I think an alias_refresh is not needed. Instead the refresh interval could be based on the TTL of the source?

ilijamt: @yongtang I don't mind remove alias_refresh at all, the reason it's there is that the health-checks can be 15 seconds or 30 seconds when they are defined, so this allows you to override the refreshing period. As for the source TTL, there is no way to change it other than leave it at 60 seconds. But mind you this is valid for all records not just aliases. Should I remove **alias_refresh** @yongtang ?


> If an alias record points to an AWS resource, you can't set the time to live (TTL); Route 53 uses the default TTL for the resource. If an alias record points to another record in the same hosted zone, Route 53 uses the TTL of the record that the alias record points to. For more information about the current TTL value for Elastic Load Balancing, go to Request Routing in the Elastic Load Balancing User Guide and search for "ttl".

And this is only if we are aliasing in the same hosted zone, not an external one.
I think you are talking about this case, but then we run into the same issue with health-checks.

@miekg Any record returned by AWS that is an ALIAS will fail, 
https://github.com/coredns/coredns/pull/3442/files#diff-60d1faf3afabbffbb6c6dd8dc9e5d95bR64-R67
This will never work as AWS will never return an IP address when you get the AWS ZONE from them, it will only return a record that says it's an alias.

ilijamt: I've added a same zone lookup, but regarding the timing how do you want to resolve it? @yongtang any ideas? We can make it dynamic, based on the record TTL but that will get unnecessary complicated, if you ask me. But I'm open to any ideas. I still feel that having it configurable would offer the simplest and least complicated option.

@miekg this is what we get from the AWS API when we process the alias records.

```
(*route53.ResourceRecordSet)(0xc000746180)({
  AliasTarget: {
    DNSName: "d21v7mkn0v2jz6.cloudfront.net.",
    EvaluateTargetHealth: false,
    HostedZoneId: "Z2FDTNDATAQYW2"
  },
  Name: "files.msp.sh.",
  Type: "A"
})
````

File: plugin/route53/setup.go [hunk 3]
```diff
@@ -46,7 +62,8 @@ func setup(c *caddy.Controller) error {
 		var providers []credentials.Provider
 		var fall fall.F
 
-		refresh := time.Duration(1) * time.Minute // default update frequency to 1 minute
+		refresh := time.Duration(1) * time.Minute       // default update frequency to 1 minute
```
dilyevsky: You don't need to cast this to `time.Duration`. In fact this makes it less readable.

plugin/route53/route53.go
```diff
@@ -99,6 +120,74 @@ func (h *Route53) Run(ctx context.Context) error {
...
+	for zName, z := range h.zones {
+		go func(zName string, z []*zone) {
+			var err error
+			defer func() {
+				errc <- err
+			}()
+			var rrs []dns.RR
+			var nameserver string
+			for _, hostedZone := range z {
+				for name, alias := range hostedZone.aliases {
```
dilyevsky: There's a whole lot of nesting going on here. It makes sense to split this up into sub-routines to make this more comprehensible.

plugin/route53/alias_resolver.go
```diff
@@ -0,0 +1,39 @@
...
+
+// AliasResolver helps us in resolving ALIAS records from AWS.
+type AliasResolver interface {
+	Resolve(ctx context.Context, dnsName string, dnsType uint16, ns string, ttl int64) (rrs []dns.RR, err error)
+}
+
+type authoritativeNsResolver struct{}
+
+func (anr authoritativeNsResolver) Resolve(ctx context.Context, dnsName string, dnsType uint16, ns string, ttl int64) (rrs []dns.RR, err error) {
+	var client = new(dns.Client)
```
dilyevsky:
```suggestion
	client := new(dns.Client)
```

plugin/route53/alias_resolver.go
```diff
@@ -0,0 +1,39 @@
...
+	var r, req *dns.Msg
```
dilyevsky: Please remove this line and declare these where you initialize it below:

```
req := new(dns.Msg)
...
resp, _, err := client.ExchangeContext(...
```

dilyevsky (commented):
This is nice! I got some style comments.

plugin/route53/setup.go
ilijamt: It's the original code, it wasn't really changed by me, I just formatted it

ilijamt: https://github.com/coredns/coredns/blob/master/plugin/route53/setup.go#L49

plugin/route53/route53.go
ilijamt: I've refactored this part a bit so now it should be simpler and easier to read. Let me know what you think.

plugin/route53/setup.go
dilyevsky: 🤔 did i write it this way? sorry, nvm then, i'll fix it later

ilijamt:
No worries I'll make the change since I'm editing already. I'll just change
it to time.Minute

plugin/route53/setup.go
ilijamt: I changed it.

ilijamt:
@dilyevsky @yongtang Any more comments, can we get this merged?

plugin/route53/README.md
miekg: alias_refreshed needs to be removed - this doesn't warrant more config knobs

plugin/route53/alias_resolver.go
```diff
@@ -0,0 +1,39 @@
+package route53
+
+import (
+	"context"
+	"github.com/miekg/dns"
+)
+
+// AliasResolver helps us in resolving ALIAS records from AWS.
+type AliasResolver interface {
```
miekg, line 9: Why is this needed? Why is it an interface? And why can't the normal "get records from AWS" do this?

plugin/route53/alias_resolver_test.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/miekg/dns"
+	"net"
+	"testing"
+	"time"
+)
+
+func getFreePort() (int, error) {
```
miekg: not needed, pkg/test or upper level test should be added.

plugin/route53/route53_test.go
```diff
@@ -34,37 +43,61 @@ func (fakeRoute53) ListResourceRecordSetsPagesWithContext(_ aws.Context, in *rou
 	rrsResponse := map[string][]*route53.ResourceRecordSet{}
 	for _, r := range []struct {
 		rType, name, value, hostedZoneID string
+		aliasDnsName, aliasHostedZoneId  string
```
miekg: please add a new contained tests for this instead of overloading existing tests. This way we have more, finer grained tests which help debugging and pin pointing mistakes.

plugin/route53/setup.go
```diff
@@ -31,6 +31,22 @@ var f = func(credential *credentials.Credentials) route53iface.Route53API {
 	return route53.New(session.Must(session.NewSession(&aws.Config{Credentials: credential})))
 }
 
+func parseDurationFromCaddyController(cc *caddy.Controller, def time.Duration) (duration time.Duration, err error) {
```
miekg: this doesn't need a extra function just to parse a duration

plugin/route53/setup.go
```diff
@@ -31,6 +31,22 @@ var f = func(credential *credentials.Credentials) route53iface.Route53API {
...
+	var durationStr string = cc.Val()
+	_, err = strconv.Atoi(cc.Val())
```
miekg: why? Just call ParseDuration

miekg: If we're keeping this config knob in the first place; which I think we shouldn't

plugin/route53/alias_resolver_test.go
```diff
@@ -0,0 +1,82 @@
+package route53
+
+import (
+	"context"
+	"fmt"
+	"github.com/miekg/dns"
```
miekg, line 6: fmt

plugin/route53/alias_resolver.go
```diff
@@ -0,0 +1,39 @@
+package route53
+
+import (
+	"context"
+	"github.com/miekg/dns"
```
miekg, line 5: fmt

go.mod
```diff
@@ -4,12 +4,12 @@ go 1.12
 
 require (
 	cloud.google.com/go v0.41.0 // indirect
-	github.com/Azure/azure-sdk-for-go v32.6.0+incompatible
+	github.com/Azure/azure-sdk-for-go v32.4.0+incompatible
```
miekg: please rebase and back out changes to this file

plugin/route53/alias_resolver.go
ilijamt: AWS doesn't provide the records, aliases are fqdns that can only be of type A or AAAA.

This is all the information you would get from the AWS API regarding an alias. And AWS evaluates the IP addresses based on healthchecks so even if the DNSName in the record set below has 3 IP's if the health check fails for 1 of them, you will only get 2 when you query AWS dns server.

```
(*route53.ResourceRecordSet)(0xc000746180)({
  AliasTarget: {
    DNSName: "d21v7mkn0v2jz6.cloudfront.net.",
    EvaluateTargetHealth: false,
    HostedZoneId: "Z2FDTNDATAQYW2"
  },
  Name: "files.msp.sh.",
  Type: "A"
})
```

As for why it's an interface is so we can more easily test our code, without this it would actually go out and do queries when we run our tests.

And also if we want to provide different methods of resolving, we can just substitute the current implementation.

plugin/route53/route53_test.go
ilijamt: agreed, will extract the tests in a different file, and revert any changes to the original file

plugin/route53/setup.go
ilijamt: will remove, this was for when I added another option for refreshing. but since it's not needed anymore, I will revert the file to the original contents.

plugin/route53/setup.go
ilijamt: reverted the original file.

go.mod
ilijamt: done

plugin/route53/alias_resolver.go
ilijamt: ran ```gofmt -s -w -l plugin/route53```

plugin/route53/README.md
ilijamt: removed.

plugin/route53/alias_resolver_test.go
ilijamt: The authoritative resolver is not exposed, so not easily testable from the pkg/test package.

The code is actually quite simple don't know if this even warrants a test. It just queries the authoritative name-server for the records. By using the dns server library as a resolver during the test, which returns dummy data.

What do you think @miekg ?

miekg: Not sure, but this contraption should not be added, the current test code should be usable

ilijamt: I'll remove this then. Can you take a look if this is what you were thinking?

ilijamt:
Is it possible to get this merged? Are there any other comments, or feedback to implement?

yongtang:
@ilijamt Sorry for the late reply. I will setup a AWS route53 alias record today and give it a try. Will come back to you shortly.

yongtang:
@ilijamt I am getting a NXDOMAIN for alias pointing to an ELB dns name, while at the same time, a static record (e.g., `1.1.1.1` for `A`) in the same domain works. I will need to take another look to see if this is working as expected.

yongtang (changes requested):
I will need to take a further look. Still having trouble to get plugin to work with AWS's DNS Alias+ElasticLoadBalancer. Maybe it is due to just some configurations but this is not as smooth as I initially thought.

ilijamt:
@yongtang what exactly are you testing? how are you setting it up? 

Here is an example that I've set up.
Screenshot from 2020-01-20 17-46-07

Also I've set it up with terraform with the following code. 

```
resource "aws_route53_record" "files_msp_sh" {
  provider = aws.eu-west-1
  name     = aws_s3_bucket.files_msp_sh.bucket
  type     = "A"
  zone_id  = aws_route53_zone.msp_sh.zone_id

  alias {
    name                   = aws_cloudfront_distribution.files_msp_sh.domain_name
    zone_id                = aws_cloudfront_distribution.files_msp_sh.hosted_zone_id
    evaluate_target_health = false
  }
}
```

ilijamt:
Here is the Corefile I set up.

```
(base) {
  log
  errors
}

example.com.:9053 {
  import base
  route53 example.com:Z263LA6KRGXDBT  {
      credentials coredns
  }
}
```

yongtang:
@ilijamt  Here is the setup I use:
```
.:9053 {
    route53 special.example.org.:Z01234567890123456789Z {
      aws_access_key XXX XXX # <= AWS keys
    }
    log
    errors
}
```

For route53 `special.example.org` I created two record:
- The first one is a static A record of `record.special.example.org` with value `1.1.1.1`. 
- The second one is an alias target of `alias.special.example.org` pointing to a elastic load balancer `xxxxxxxx.elb.us-east-1.amazonaws.com`
```
# aws route53 list-resource-record-sets --hosted-zone-id Z1234567Z
{
    "ResourceRecordSets": [
        {
            "ResourceRecords": [
                {
                    "Value": "ns-1536.awsdns-00.co.uk."
                }, 
                {
                    "Value": "ns-0.awsdns-00.com."
                }, 
                {
                    "Value": "ns-1024.awsdns-00.org."
                }, 
                {
                    "Value": "ns-512.awsdns-00.net."
                }
            ], 
            "Type": "NS", 
            "Name": "special.example.org.", 
            "TTL": 172800
        }, 
        {
            "ResourceRecords": [
                {
                    "Value": "ns-1536.awsdns-00.co.uk. awsdns-hostmaster.amazon.com. 1 7200 900 1209600 86400"
                }
            ], 
            "Type": "SOA", 
            "Name": "special.example.org.", 
            "TTL": 900
        }, 
        {
            "AliasTarget": {
                "HostedZoneId": "ZXXXXXXX", 
                "EvaluateTargetHealth": false, 
                "DNSName": "XXXXXXX.elb.us-east-1.amazonaws.com."
            }, 
            "Type": "A", 
            "Name": "alias.special.example.org."
        }, 
        {
            "ResourceRecords": [
                {
                    "Value": "1.1.1.1"
                }
            ], 
            "Type": "A", 
            "Name": "record.special.example.org.", 
            "TTL": 5
        }
    ]
}

```

When query `record.special.example.org`:
```
$ dig @127.0.0.1 -p 9053 record.special.example.org

; <<>> DiG 9.11.3-1ubuntu1.11-Ubuntu <<>> @127.0.0.1 -p 9053 record.special.example.org
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 38914
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 4, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; COOKIE: fc353cbc6bf04a5d (echoed)
;; QUESTION SECTION:
;record.special.example.org.	IN	A

;; ANSWER SECTION:
record.special.example.org. 5	IN	A	1.1.1.1

;; AUTHORITY SECTION:
special.example.org.	172800	IN	NS	ns-1536.awsdns-00.co.uk.
special.example.org.	172800	IN	NS	ns-0.awsdns-00.com.
special.example.org.	172800	IN	NS	ns-1024.awsdns-00.org.
special.example.org.	172800	IN	NS	ns-512.awsdns-00.net.

;; Query time: 0 msec
;; SERVER: 127.0.0.1#9053(127.0.0.1)
;; WHEN: Sun Jan 26 21:16:21 UTC 2020
;; MSG SIZE  rcvd: 323
```

When query `alias.special.example.org`:
```
$ dig @127.0.0.1 -p 9053 alias.special.example.org

; <<>> DiG 9.11.3-1ubuntu1.11-Ubuntu <<>> @127.0.0.1 -p 9053 alias.special.example.org
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 54233

;; OPT PSEUDOSECTION:
; COOKIE: 8d6f5570f4994054 (echoed)
;; QUESTION SECTION:
;alias.special.example.org.	IN	A

;; AUTHORITY SECTION:
special.example.org.	900	IN	SOA	ns-1536.awsdns-00.co.uk. awsdns-hostmaster.amazon.com. 1 7200 900 1209600 86400

;; Query time: 0 msec
;; SERVER: 127.0.0.1#9053(127.0.0.1)
;; WHEN: Sun Jan 26 21:14:57 UTC 2020
;; MSG SIZE  rcvd: 172
```

I tries multiple times ( to allow potential data sync up), but so far I always get the NXDOMAIN whenever trying to access the alias target.

yongtang:
@ilijamt I noticed in your setup, you are pointing to cloudfront. Can you give ELB a try?

ilijamt:
@yongtang Yeah, here is my setup, I created a new zone, just for the test to make sure everything is working as expected.

The zone you defined is a public zone right? Because private zones only resolve from inside AWS. 
The only thing I can think of if we can arrange offline some access to see what's different between your setup and mine. 

```json
$ aws route53 list-resource-record-sets --hosted-zone-id Z2HLV6T7982LFR
{
    "ResourceRecordSets": [
        {
            "Name": "coredns.com.",
            "Type": "NS",
            "TTL": 172800,
            "ResourceRecords": [
                {
                    "Value": "ns-852.awsdns-42.net."
                },
                {
                    "Value": "ns-40.awsdns-05.com."
                },
                {
                    "Value": "ns-1026.awsdns-00.org."
                },
                {
                    "Value": "ns-1873.awsdns-42.co.uk."
                }
            ]
        },
        {
            "Name": "coredns.com.",
            "Type": "SOA",
            "TTL": 900,
            "ResourceRecords": [
                {
                    "Value": "ns-852.awsdns-42.net. awsdns-hostmaster.amazon.com. 1 7200 900 1209600 86400"
                }
            ]
        },
        {
            "Name": "cloudfront.coredns.com.",
            "Type": "A",
            "AliasTarget": {
                "HostedZoneId": "Z2FDTNDATAQYW2",
                "DNSName": "d21v7mkn0v2jz6.cloudfront.net.",
                "EvaluateTargetHealth": false
            }
        },
        {
            "Name": "elb.coredns.com.",
            "Type": "A",
            "AliasTarget": {
                "HostedZoneId": "Z32O12XQLNTSW2",
                "DNSName": "dualstack.test-1751199221.eu-west-1.elb.amazonaws.com.",
                "EvaluateTargetHealth": false
            }
        },
        {
            "Name": "test.new.elb.coredns.com.",
            "Type": "A",
            "AliasTarget": {
                "HostedZoneId": "Z32O12XQLNTSW2",
                "DNSName": "dualstack.test-1751199221.eu-west-1.elb.amazonaws.com.",
                "EvaluateTargetHealth": false
            }
        }
    ]
}
```

Corefile
```
(base) {
  log
  errors
}

.:9053 {
  import base
  route53 coredns.com.:Z2HLV6T7982LFR {
      credentials coredns
  }
}
```

* Elastic load balancer 

```shell
$ dig elb.coredns.com. -p 9053 @127.0.0.1

; <<>> DiG 9.11.5-P4-5.1+b1-Debian <<>> elb.coredns.com. -p 9053 @127.0.0.1
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 45113
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 4, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; COOKIE: 1940f507edd0ac7b (echoed)
;; QUESTION SECTION:
;elb.coredns.com.		IN	A

;; ANSWER SECTION:
elb.coredns.com.	60	IN	A	52.48.211.165

;; AUTHORITY SECTION:
coredns.com.		172800	IN	NS	ns-852.awsdns-42.net.
coredns.com.		172800	IN	NS	ns-40.awsdns-05.com.
coredns.com.		172800	IN	NS	ns-1026.awsdns-00.org.
coredns.com.		172800	IN	NS	ns-1873.awsdns-42.co.uk.

;; Query time: 0 msec
;; SERVER: 127.0.0.1#9053(127.0.0.1)
;; WHEN: Sun Jan 26 23:49:59 CET 2020
;; MSG SIZE  rcvd: 270

$ dig elb.coredns.com. @ns-40.awsdns-05.com.
; <<>> DiG 9.11.5-P4-5.1+b1-Debian <<>> elb.coredns.com. @ns-40.awsdns-05.com.
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 59494
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 4, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
;; QUESTION SECTION:
;elb.coredns.com.		IN	A

;; ANSWER SECTION:
elb.coredns.com.	60	IN	A	52.48.211.165

;; AUTHORITY SECTION:
coredns.com.		172800	IN	NS	ns-1026.awsdns-00.org.
coredns.com.		172800	IN	NS	ns-1873.awsdns-42.co.uk.
coredns.com.		172800	IN	NS	ns-40.awsdns-05.com.
coredns.com.		172800	IN	NS	ns-852.awsdns-42.net.

;; Query time: 28 msec
;; SERVER: 205.251.192.40#53(205.251.192.40)
;; WHEN: Sun Jan 26 23:52:12 CET 2020
;; MSG SIZE  rcvd: 196
```

```shell
$ dig test.new.elb.coredns.com. -p 9053 @127.0.0.1

; <<>> DiG 9.11.5-P4-5.1+b1-Debian <<>> test.new.elb.coredns.com. -p 9053 @127.0.0.1
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 12747
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 4, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; COOKIE: 1508fd85cab6af29 (echoed)
;; QUESTION SECTION:
;test.new.elb.coredns.com.	IN	A

;; ANSWER SECTION:
test.new.elb.coredns.com. 60	IN	A	52.48.211.165

;; AUTHORITY SECTION:
coredns.com.		172800	IN	NS	ns-852.awsdns-42.net.
coredns.com.		172800	IN	NS	ns-40.awsdns-05.com.
coredns.com.		172800	IN	NS	ns-1026.awsdns-00.org.
coredns.com.		172800	IN	NS	ns-1873.awsdns-42.co.uk.

;; Query time: 0 msec
;; SERVER: 127.0.0.1#9053(127.0.0.1)
;; WHEN: Mon Jan 27 00:15:05 CET 2020
;; MSG SIZE  rcvd: 288

$ dig test.new.elb.coredns.com. @ns-40.awsdns-05.com.                                         

; <<>> DiG 9.11.5-P4-5.1+b1-Debian <<>> test.new.elb.coredns.com. @ns-40.awsdns-05.com.
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 14595
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 4, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
;; QUESTION SECTION:
;test.new.elb.coredns.com.	IN	A

;; ANSWER SECTION:
test.new.elb.coredns.com. 60	IN	A	52.48.211.165

;; AUTHORITY SECTION:
coredns.com.		172800	IN	NS	ns-1026.awsdns-00.org.
coredns.com.		172800	IN	NS	ns-1873.awsdns-42.co.uk.
coredns.com.		172800	IN	NS	ns-40.awsdns-05.com.
coredns.com.		172800	IN	NS	ns-852.awsdns-42.net.

;; Query time: 30 msec
;; SERVER: 205.251.192.40#53(205.251.192.40)
;; WHEN: Mon Jan 27 00:15:19 CET 2020
;; MSG SIZE  rcvd: 205
```

* Cloudfront
```shell
$ dig cloudfront.coredns.com. -p 9053 @127.0.0.1
; <<>> DiG 9.11.5-P4-5.1+b1-Debian <<>> cloudfront.coredns.com. -p 9053 @127.0.0.1
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 20624
;; flags: qr aa rd; QUERY: 1, ANSWER: 4, AUTHORITY: 4, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; COOKIE: e6e84c40de9d53ab (echoed)
;; QUESTION SECTION:
;cloudfront.coredns.com.		IN	A

;; ANSWER SECTION:
cloudfront.coredns.com.	60	IN	A	54.240.168.26
cloudfront.coredns.com.	60	IN	A	54.240.168.125
cloudfront.coredns.com.	60	IN	A	54.240.168.79
cloudfront.coredns.com.	60	IN	A	54.240.168.36

;; AUTHORITY SECTION:
coredns.com.		172800	IN	NS	ns-852.awsdns-42.net.
coredns.com.		172800	IN	NS	ns-40.awsdns-05.com.
coredns.com.		172800	IN	NS	ns-1026.awsdns-00.org.
coredns.com.		172800	IN	NS	ns-1873.awsdns-42.co.uk.

;; Query time: 0 msec
;; SERVER: 127.0.0.1#9053(127.0.0.1)
;; WHEN: Sun Jan 26 23:50:49 CET 2020
;; MSG SIZE  rcvd: 398

$ dig cloudfront.coredns.com. @ns-40.awsdns-05.com.

; <<>> DiG 9.11.5-P4-5.1+b1-Debian <<>> cloudfront.coredns.com. @ns-40.awsdns-05.com.
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 56231
;; flags: qr aa rd; QUERY: 1, ANSWER: 4, AUTHORITY: 4, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
;; QUESTION SECTION:
;cloudfront.coredns.com.		IN	A

;; ANSWER SECTION:
cloudfront.coredns.com.	60	IN	A	54.240.168.36
cloudfront.coredns.com.	60	IN	A	54.240.168.125
cloudfront.coredns.com.	60	IN	A	54.240.168.26
cloudfront.coredns.com.	60	IN	A	54.240.168.79

;; AUTHORITY SECTION:
coredns.com.		172800	IN	NS	ns-1026.awsdns-00.org.
coredns.com.		172800	IN	NS	ns-1873.awsdns-42.co.uk.
coredns.com.		172800	IN	NS	ns-40.awsdns-05.com.
coredns.com.		172800	IN	NS	ns-852.awsdns-42.net.

;; Query time: 31 msec
;; SERVER: 205.251.192.40#53(205.251.192.40)
;; WHEN: Sun Jan 26 23:52:02 CET 2020
;; MSG SIZE  rcvd: 251
```

```
$ ./coredns
.:9053
CoreDNS-1.6.5
linux/amd64, go1.13.1, ab250485
[INFO] 127.0.0.1:53936 - 9248 "A IN elb.coredns.com. udp 56 false 4096" NOERROR qr,aa,rd 247 0.000333069s
[INFO] 127.0.0.1:43767 - 45113 "A IN elb.coredns.com. udp 56 false 4096" NOERROR qr,aa,rd 247 0.000287838s
[INFO] 127.0.0.1:46963 - 55845 "A IN elb.coredns.com. udp 56 false 4096" NOERROR qr,aa,rd 247 0.000344065s
[INFO] 127.0.0.1:43977 - 30802 "A IN elb.coredns.com. udp 56 false 4096" NOERROR qr,aa,rd 247 0.000220154s
[INFO] 127.0.0.1:49293 - 36772 "A IN cloudfront.coredns.com. udp 63 false 4096" NOERROR qr,aa,rd 375 0.000301676s
[INFO] 127.0.0.1:39418 - 20624 "A IN cloudfront.coredns.com. udp 63 false 4096" NOERROR qr,aa,rd 375 0.000237274s
[INFO] 127.0.0.1:55256 - 13127 "A IN elb.coredns.com. udp 56 false 4096" NOERROR qr,aa,rd 247 0.000226342s
[INFO] 127.0.0.1:60727 - 63571 "A IN cloudfront.coredns.com. udp 63 false 4096" NOERROR qr,aa,rd 375 0.000209574s
[INFO] 127.0.0.1:60805 - 28734 "A IN test.new.elb.coredns.com. udp 65 false 4096" NOERROR qr,aa,rd 265 0.000313612s
[INFO] 127.0.0.1:42292 - 58847 "A IN test.new.elb.coredns.com. udp 65 false 4096" NOERROR qr,aa,rd 265 0.000216541s
[INFO] 127.0.0.1:36199 - 58235 "A IN test.new.elb.coredns.com. udp 65 false 4096" NOERROR qr,aa,rd 265 0.00029881s
[INFO] 127.0.0.1:58097 - 12712 "A IN test.new.elb.coredns.com. udp 65 false 4096" NOERROR qr,aa,rd 265 0.000235106s
[INFO] 127.0.0.1:52848 - 12747 "A IN test.new.elb.coredns.com. udp 65 false 4096" NOERROR qr,aa,rd 265 0.000219961s

```

ilijamt:
There was one difference between your test and my more than one level deep for the subdomain. 

So I added multiple levels deep and this doesn't work, I'll take a look and fix this soon.

```json
{
    "ResourceRecordSets": [
        {
            "Name": "coredns.com.",
            "Type": "NS",
            "TTL": 172800,
            "ResourceRecords": [
                {
                    "Value": "ns-852.awsdns-42.net."
                },
                {
                    "Value": "ns-40.awsdns-05.com."
                },
                {
                    "Value": "ns-1026.awsdns-00.org."
                },
                {
                    "Value": "ns-1873.awsdns-42.co.uk."
                }
            ]
        },
        {
            "Name": "coredns.com.",
            "Type": "SOA",
            "TTL": 900,
            "ResourceRecords": [
                {
                    "Value": "ns-852.awsdns-42.net. awsdns-hostmaster.amazon.com. 1 7200 900 1209600 86400"
                }
            ]
        },
        {
            "Name": "cloudfront.coredns.com.",
            "Type": "A",
            "AliasTarget": {
                "HostedZoneId": "Z2FDTNDATAQYW2",
                "DNSName": "d21v7mkn0v2jz6.cloudfront.net.",
                "EvaluateTargetHealth": false
            }
        },
        {
            "Name": "elb.coredns.com.",
            "Type": "A",
            "AliasTarget": {
                "HostedZoneId": "Z32O12XQLNTSW2",
                "DNSName": "dualstack.test-1751199221.eu-west-1.elb.amazonaws.com.",
                "EvaluateTargetHealth": false
            }
        },
        {
            "Name": "test.new.elb.coredns.com.",
            "Type": "A",
            "AliasTarget": {
                "HostedZoneId": "Z32O12XQLNTSW2",
                "DNSName": "dualstack.test-1751199221.eu-west-1.elb.amazonaws.com.",
                "EvaluateTargetHealth": false
            }
        }
    ]
}
```

```
$ dig test.new.elb.coredns.com. -p 9053 @127.0.0.1
; <<>> DiG 9.11.5-P4-5.1+b1-Debian <<>> test.new.elb.coredns.com. -p 9053 @127.0.0.1
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 45682

;; OPT PSEUDOSECTION:
; COOKIE: c9f200d5a59012aa (echoed)
;; QUESTION SECTION:
;test.new.elb.coredns.com.	IN	A

;; AUTHORITY SECTION:
coredns.com.		900	IN	SOA	ns-852.awsdns-42.net. awsdns-hostmaster.amazon.com. 1 7200 900 1209600 86400

;; Query time: 0 msec
;; SERVER: 127.0.0.1#9053(127.0.0.1)
;; WHEN: Mon Jan 27 00:07:53 CET 2020
;; MSG SIZE  rcvd: 160

$ dig test.new.elb.coredns.com. @ns-40.awsdns-05.com.
; <<>> DiG 9.11.5-P4-5.1+b1-Debian <<>> test.new.elb.coredns.com. @ns-40.awsdns-05.com.
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 8710
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 4, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
;; QUESTION SECTION:
;test.new.elb.coredns.com.	IN	A

;; ANSWER SECTION:
test.new.elb.coredns.com. 60	IN	A	52.48.211.165

;; AUTHORITY SECTION:
coredns.com.		172800	IN	NS	ns-1026.awsdns-00.org.
coredns.com.		172800	IN	NS	ns-1873.awsdns-42.co.uk.
coredns.com.		172800	IN	NS	ns-40.awsdns-05.com.
coredns.com.		172800	IN	NS	ns-852.awsdns-42.net.

;; Query time: 31 msec
;; SERVER: 205.251.192.40#53(205.251.192.40)
;; WHEN: Mon Jan 27 00:08:23 CET 2020
;; MSG SIZE  rcvd: 205
```

```
$ ./coredns
.:9053
CoreDNS-1.6.7
linux/amd64, go1.13.1, 85cf73ab
[INFO] 127.0.0.1:42095 - 11491 "A IN elb.coredns.com. udp 56 false 4096" NOERROR qr,aa,rd 247 0.000295599s
[INFO] 127.0.0.1:38580 - 15843 "A IN test.new.elb.coredns.com. udp 65 false 4096" NXDOMAIN qr,aa,rd 137 0.000223761s
[INFO] 127.0.0.1:52180 - 1843 "A IN test.new.elb.coredns.com. udp 65 false 4096" NXDOMAIN qr,aa,rd 137 0.000211356s
[INFO] 127.0.0.1:46354 - 29467 "A IN test.new.elb.coredns.com. udp 65 false 4096" NXDOMAIN qr,aa,rd 137 0.000106034s
[INFO] 127.0.0.1:38162 - 45682 "A IN test.new.elb.coredns.com. udp 65 false 4096" NXDOMAIN qr,aa,rd 137 0.000125902s
```

yongtang:
@ilijamt There were several reasons (and several advantages) that coredns route53 plugin was implemented at the first place:
1. coredns' route53 plugin using https to access api service endpoint, and does not using unreliable udp, and does not need public DNS access.
2. coredns's route53 plugin is able to probe internal dns record (non-public domain), with security access controlled by much safer IAM roles.

From your implementation, it looks like it takes  mixture approach of
1. using aws https service endpoint to extract the alias record
2. use publicly accessible dns to resolve the alias record.

However, if the implementation requires public dns access, then you could very much just use coredns as a dns proxy and it is not necessary to have the alias support in coredns at the place? In this case, you even don't need route53 plugin I think?

To make alias support on rout53 plugin more useful, I think we should stick with AWS API's to retrieve the information. That will maintain the same benefit of 
1. only https transportation
2. security controlled by IAM roles.
3. No need for public DNS/UDP access

Essentially coredns's route53 plugin could just
1. access route53 API to get the alias target record
2. access appropriate service (route53 or elb or cloudfont) to resolve the record into IP addresses.

Since AliasTarget only has limited applicable service types, the above would be feasible I think?

ilijamt:
@yongtang 

> However, if the implementation requires public dns access, then you could very much just use coredns as a dns proxy and it is not necessary to have the alias support in coredns at the place? In this case, you even don't need route53 plugin I think?

In my case we needed to have the information sitting in coredns, as my setup had to do with a multi-cloud setup, and if our connection would break we would not be able to resolve the records. But in this case, coredns acts as a primary DNS server for each of our locations, even if the connection between the clouds or locations gets cut. And coredns in our case is used for accessing multiple DNS zones, not just from R53, but also from AD, and other DNS resolvers, luckily most of them support axfr so it was not an issue.

> 1. access route53 API to get the alias target record

Nothing to do here, we already do this

> 2. access appropriate service (route53 or LB or CloudFront) to resolve the record into IP addresses.

It is possible to get a list of the IP addresses from AWS for an alias but only for our account, not if you do an alias on a cross-account afaik, for LB, CloudFront or any of the other ones we do not have access to their hosted zones, as they are handled by AWS. It will require handling all the services that can have an alias record. Currently, at the top of my head that would be:

-  CloudFront
-  Elastic Beanstalk
-  ELB load balancer
-  S3 website endpoint
-  Resource record set in the same hosted zone
-  VPC endpoint
-  API Gateway custom regional API
-  Global Accelerator DNS name

This will increase the complexity quite a lot, also let's say we agree with the increase of the complexity. The second problem that I see is how do we handle the following.

- Simple routing policy – Use for a single resource that performs a given function for your domain, for example, a web server that serves content for the example.com website.
- Failover routing policy – Use when you want to configure active-passive failover.
- Geolocation routing policy – Use when you want to route traffic based on the location of your users.
- Geoproximity routing policy – Use when you want to route traffic based on the location of your resources and, optionally, shift traffic from resources in one location to resources in another.
- Latency routing policy – Use when you have resources in multiple AWS Regions and you want to route traffic to the region that provides the best latency.
- Multivalue answer routing policy – Use when you want Route 53 to respond to DNS queries with up to eight healthy records selected at random.
- Weighted routing policy – Use to route traffic to multiple resources in proportions that you specify.

The recordset options vary based on the type of routing policy we decide to use. 
And also for each routing policy, we can have health-checks associated with them, that are evaluated by Route 53 on the DNS server associated with the hosted zone. At this stage, I don't see a way to extract all the information regarding an alias record simply, by querying the API for this information, because even if we query the API for the IP addresses of an LB, we will get all of them, but what happens when one of them is not healthy? We will end up with a possibility that we may get wrong and the unhealthy IP address(es), or we need to do the health-checks on our own.

That was the reason why I decided to use the authoritative resolver for the hosted zone itself to resolve the special cases for ALIAS, otherwise, I would have to recreate the whole logic behind the routing policies, health-checks, and anything else that would come out of it.

List of available properties for each record (based on Route 53 documentation)

```
HTTP/1.1 200
<?xml version="1.0" encoding="UTF-8"?>
<ListResourceRecordSetsResponse>
   <IsTruncated>boolean</IsTruncated>
   <MaxItems>string</MaxItems>
   <NextRecordIdentifier>string</NextRecordIdentifier>
   <NextRecordName>string</NextRecordName>
   <NextRecordType>string</NextRecordType>
   <ResourceRecordSets>
      <ResourceRecordSet>
         <AliasTarget>
            <DNSName>string</DNSName>
            <EvaluateTargetHealth>boolean</EvaluateTargetHealth>
            <HostedZoneId>string</HostedZoneId>
         </AliasTarget>
         <Failover>string</Failover>
         <GeoLocation>
            <ContinentCode>string</ContinentCode>
            <CountryCode>string</CountryCode>
            <SubdivisionCode>string</SubdivisionCode>
         </GeoLocation>
         <HealthCheckId>string</HealthCheckId>
         <MultiValueAnswer>boolean</MultiValueAnswer>
         <Name>string</Name>
         <Region>string</Region>
         <ResourceRecords>
            <ResourceRecord>
               <Value>string</Value>
            </ResourceRecord>
         </ResourceRecords>
         <SetIdentifier>string</SetIdentifier>
         <TrafficPolicyInstanceId>string</TrafficPolicyInstanceId>
         <TTL>long</TTL>
         <Type>string</Type>
         <Weight>long</Weight>
      </ResourceRecordSet>
   </ResourceRecordSets>
</ListResourceRecordSetsResponse>
```

yongtang:
@ilijamt My concern is very much about the security. Route53 API gives you better security because 1) https instead of udp, and you don't need public DNS access or not even the internal DNS server set force by AWS in your VPC. 2) access is controlled through IAM which give you the option to fine tune the access.

In the implementation in this PR essentially a public DNS resolver is introduced. That could very much have unintended consequences. Also, this public DNS resolver model does not work if you are in a VPC and AWS DNS record is not public.

I very much still think get the DNS information from AWS API (https) is the way to go. In case this is too complicated for implementation, there are still other solutions.

If you really need to resolve the aliasTarget, your deployment is inside a VPC, at least you should use the internal DNS by AWS as it is much safer than recursively resolve DNS through public Internet. In case public DNS access is a must for your deployment, you still have the option to point to a designated DNS resolver (than implement a resolver in route53 plugin that only solved this specific case).

 I think what you want to achieve might be done in a way without implement a resolver here:
1) Modify the route53 plugin  and allow user to decide if NXDOMAIN, NODATA, or Fallthrough should be done if alias target is encountered (e.g., in this PR).
2) For Fallthrough, in Corefile configuration, add forward to forward the fallthrough DNS queries (by AliasTarget) to a designated DNS server (like the one by AWS), or to a designated DNS resolver.

Will the above resolve the issue you are encountering?

miekg:
should we close this PR?

yongtang:
@miekg I think we can close this PR right now. In case there are still some interest for AWS route53 Alias, it could be implemented in the way mentioned in the last comment. The implementation will not require a full-fledge resolver (quite hard to implement right and securely). It only need to have an option to specify what coredns will do (e.g., passthrough).

I can work on an implementation if interests from the community stays. For now let's close this PR.

Hronom:
@yongtang , @miekg I'm interested in this feature. Also expectation that plugin for Route 53 supports all types of cases that Route 53 has, otherwise(like in my case) I spend a lot of time debugging about what is going on...

Please, can you provide temporary workaround guide if we use private hosted zone in AWS Route 53?

PR #3420, opened 2019-11-04
plugin/transfer: notifies

chrisohaver:
### 1. Why is this pull request needed and what does it do?

Adds notifies to transfer plugin, per https://github.com/coredns/coredns/pull/3223#issuecomment-534099365

### 2. Which issues (if any) are related?
#3223 #3121

### 3. Which documentation changes (if any) need to be made?
included 

### 4. Does this introduce a backward incompatible change or deprecation?
no

PR #3452, opened 2019-11-15
plugin/kubernetes: use transfer plugin

chrisohaver:

Allow use of the _transfer_ plugin in _kubernetes_.

The existing transfer option is refactored, but still functional - to be removed when the existing transfer option is eventually removed.

This PR also removes the zone transfer interface from the backend, since only k8s actually used it.  Etcd is mildly refactored to remove the no-op zone transfer stub.

included.

No, the existing transfer option is still functional in this PR.


chrisohaver:
Travis tests passed, despite the "Pending" status.  I guess the status update from travis to github failed.

miekg:
@chrisohaver does it make sense to do all transfer capable plugins in 1 go and then release 1.7.0? From fixing/tweaks rolling in, it seems we can do one (or two?) more 1.6.x releases and then tackle this?

chrisohaver:
If you'd like to defer merging this in favor of getting bug fixes merged, that makes sense.

... especially since the migration (this PR) doesn't add new functionality.

The k8s_external PR (#3419) _does_ add new functionality.  It didn't handle zone transfers before, and IMO the use case for doing zone transfers on an external zone is more practical than zone transfers on the internal cluster ip space.  But it's still new functionality, and therefore less urgent than bug fixes.

How do you want to handle the Notify addition to the transfer plugin interface (#3420)?  Review and merge now? Or wait until 1.7.0 or later.  I think its easiest to have that merged _before_ opening any PRs that depend on it (locally, I have notify implemented for k8s_external and kubernetes branches on ice until that merges).

miekg:
He Chris, sorry for lagging on this but busy with work. What would really help to get all this transfer stuff pushed forward is impl. of the auto plugin using transfer and it all works. If transfer is only useful for a subset of plugin we need to have good think if that's the best way forward.

chrisohaver:
> What would really help to get all this transfer stuff pushed forward is impl. of the auto plugin using transfer and it all works

OK - I did this [in a branch](https://github.com/chrisohaver/coredns/tree/auto-file-notify) in January when you requested it in the Notify PR... https://github.com/coredns/coredns/pull/3420#issuecomment-574735747

~The commits that implement it are 3efabed3372d7792aee3150a229383c91bc02295 and c1d2623b775feeba01f4b4212316907a74bf79b2~ (commits squashed to 0d6b91c0ae7511f000ce23f81c572cb801d1c1dc)

The branch implements both transfer and notify, so it's dependent on the notify PR #3420.

Can we ahead and rebase/merge #3420 now, so I can push the branch up as a PR as is? 
Or do you want to see file/auto implemented _without_ notify first.

chrisohaver:
lol - looking back on that branch, I see i have a bunch of unintended changes of "notify" -> "notifyChan" in code comments and error messages.  I'll clean that up.

chrisohaver:
>I see i have a bunch of unintended changes of "notify" -> "notifyChan" in code comments and error messages. I'll clean that up.

OK, I cleaned that up and squashed to 0d6b91c0ae7511f000ce23f81c572cb801d1c1dc

chrisohaver:
@miekg, I went a head and created PR #3732 for just the transfer implementation without notify, since #3420 has not yet merged.

chrisohaver:
ugh - more conflicts.

... oh cool, its just a README this time ...

plugin/kubernetes/handler_test.go
```diff
@@ -525,7 +524,7 @@ func (APIConnServeTest) Run()                                      {}
 func (APIConnServeTest) Stop() error                               { return nil }
 func (APIConnServeTest) EpIndexReverse(string) []*object.Endpoints { return nil }
 func (APIConnServeTest) SvcIndexReverse(string) []*object.Service  { return nil }
-func (APIConnServeTest) Modified() int64                           { return time.Now().Unix() }
+func (APIConnServeTest) Modified() int64                           { return 2 }
```
stp-ip: What's the specific reason for this change?

stp-ip (commented):
Looks good. Tiny nit question.

Maybe @johnbelamaric can take a final look as well.

plugin/kubernetes/handler_test.go
chrisohaver: I don't recall for certain. I suspect so that the serial is static during the tests, not changing based on when the test is run, but I don't see a change that adds a Serial check, so it could be extraneous.

chrisohaver: Hmm ... I suspect it's part of the notify implementation, (which I had not posted because it depends on code that wasnt merged yet: #3420). 
Last November, I had Transfer+Notify working for both kubernetes and k8s_external. This may have snuck in from that.

stp-ip: Generally would say nit, which shouldn't be in the PR. Doesn't break or add anything so also ok to keep it, but it might be context that is then missing within the notify code. Aka a static serial being used within the tests, which was changed in an "unrelated" PR.

chrisohaver: Actually - it may be directly related to this change.  There are tests here for incremental transfers, which check the serial.   Making the initial serial a low static number may have been to induce a "serial not changed" scenario to verify the IXFR short response when the zone is not changed.

But looking at the IXFR tests now, I don't see how they are actually working. On the surface, they look invalid. Still re-catching my bearings.

Sorry, it's been so long I don't have a crisp memory of the work.

chrisohaver: OK, ... I see, the IXFR tests _are_ valid, and the "2" is there for the purpose of testing the IXFR fallback vs empty incremental response.  So it _is_ directly related to this change.

chrisohaver: I'll add a comment to the code to explain it.

stp-ip: Ahh now it makes sense. Thanks for bringing me up to speed.

Was missing it being used as the empty initialization and then check against hardcoded serials via k.Transfer().

All good.

stp-ip (approved):

plugin/kubernetes/handler.go
```diff
@@ -29,7 +29,7 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 
 	switch state.QType() {
 	case dns.TypeAXFR, dns.TypeIXFR:
-		k.Transfer(ctx, state)
+		k.TransferDeprecated(ctx, state)
```
miekg, line 32: not worth the effort imo. I *think* the entire case can be dropped and the plugin does the right thing

plugin/kubernetes/transfer.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/coredns/coredns/plugin/transfer"
+	"github.com/coredns/coredns/request"
+
+	"github.com/miekg/dns"
+	api "k8s.io/api/core/v1"
+)
+
+// Transfer implements transfer.Transferer
+func (k *Kubernetes) Transfer(zone string, serial uint32) (<-chan []dns.RR, error) {
+	ch := make(chan []dns.RR, 2)
```
miekg, line 21: make the channel later, so you don't need close it if the zone isn't found

plugin/kubernetes/transfer.go
```diff
@@ -0,0 +1,216 @@
...
+		return nil, transfer.ErrNotAuthoritative
+	}
+
+	// SOA record
+	soas, err := plugin.SOA(context.TODO(), k, zone, request.Request{}, plugin.Options{})
+	if err != nil {
+		close(ch)
+		return nil, errors.New("could not create soa")
+	}
+	soa, ok := soas[0].(*dns.SOA)
```
miekg, line 48: why does this need to be checked?

plugin/kubernetes/transfer.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	if !found {
+		close(ch)
+		return nil, transfer.ErrNotAuthoritative
+	}
+
+	// reverse zone transfers are not implemented
+	if rev := (plugin.Zones{"in-addr.arpa.", "ip6.arpa."}).Matches(zone); rev != "" {
+		close(ch)
+		return nil, transfer.ErrNotAuthoritative
```
miekg, line 39: technically we are authoritative to this is NotImp

miekg:
bunch of comments, but lgtm overall

plugin/kubernetes/transfer.go
miekg: I think in my PR I can get away with unbuffered and a go routine here. https://github.com/coredns/coredns/pull/3972/files#diff-f1591d056edea707afdeae51b8dc8b8fR25 onward

PR #3454, opened 2019-11-16
Fix reloading in plugin/pprof.

gonzalop:
is called for the old ones.

When pprof is enabled, reloading the server stops listening on the pprof port.

> g@t~/gonzalop/coredns#master$ curl http://127.0.0.1:6053/debug/pprof/heap > /dev/null 2>&1
> g@t~/gonzalop/coredns#master$ echo $?
> 0
> g@t~/gonzalop/coredns#master$ sudo service coredns reload
> g@t~/gonzalop/coredns#master$ curl http://127.0.0.1:6053/debug/pprof/heap > /dev/null 2>&1
> curl: (7) Failed to connect to 127.0.0.1 port 6053: Connection refused
> g@t~/gonzalop/coredns#master$ git checkout pprof_reload > /dev/null
> g@t~/gonzalop/coredns#pprof_reload$ make > /dev/null
> g@t~/gonzalop/coredns#pprof_reload$ sudo install coredns /usr/bin/
> g@t~/gonzalop/coredns#pprof_reload$ sudo service coredns restart
> g@t~/gonzalop/coredns#pprof_reload$ curl http://127.0.0.1:6053/debug/pprof/heap > /dev/null 2>&1
> g@t~/gonzalop/coredns#pprof_reload$ sudo service coredns reload
> g@t~/gonzalop/coredns#pprof_reload$ curl http://127.0.0.1:6053/debug/pprof/heap > /dev/null 2>&1
> g@t~/gonzalop/coredns#pprof_reload$ echo $?
> 0
> 

None (I can create one if needed)

None.

None.


miekg:
Ha, wasn't aware this package already exists. Note this is a thin wrapper around "golang.org/x/sys/unix"  which is already used via miekg/dns *and* that functionality also exists there https://github.com/miekg/dns/blob/master/listen_go111.go

IOW: it would be great to add a very minimal package as pkg/reuseport and use that.

Looking further we already have those bits in https://github.com/coredns/coredns/blob/master/core/dnsserver/listen_go111.go. We should extract that and make it more generally available so other plugins can easily use it.

miekg:
See https://github.com/coredns/coredns/pull/3455

gonzalop:
> See #3455 

Great! I can update this change once yours is approved.

miekg:
lgtm, the go.{mod,sum} changes should be reverted though

gonzalop:
> lgtm, the go.{mod,sum} changes should be reverted though

Done. Ty.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fix reloading..." ]
>> lgtm, the go.{mod,sum} changes should be reverted though
>
>Done. Ty.

... and merged. Thanks, this opens the door to a few other fixes/simplications 
in all other plugins that open sockets.

PR #3455, opened 2019-11-16
create pkg/reuseport

miekg:
Move the core server listening functions to a new package so plugins can
use them.

Also make *all* servers use the functions here; as only the udp/tcp
listeners where using SO_REUSEPORT (if available). This is the only
actual change in this PR; in it's core it's just a move of 2 files.

This can also be used to cleanup the dance we're doing now for
re-acquiring the sockets in e.g. the metrics plugins and the ready
plugin.


yongtang (approved):
LGTM

PR #3457, opened 2019-11-17
Fix incorrect sample configuration in clouddns

yongtang:

This PR fixes an incorect sample Corefile in clouddns,
where there is an extra `clouddns` entry in the middle.

n/a

n/a

n/a


miekg:
Ffs, once again

/merge

Issue #3158, opened 2019-08-19
question: serve stale records via cache when backend is down


JarenGlover:
At a high level, I am trying to understand if coredns can serve stale data when the backend/upstreams fails.

An example of this can be find via [Unbound](https://nlnetlabs.nl/projects/unbound/about/)'s [serve-expired](http://manpages.ubuntu.com/manpages/bionic/man5/unbound.conf.5.html) configuration option.

![image](https://user-images.githubusercontent.com/1428354/63303259-28627400-c294-11e9-9283-0eef70055361.png)

In short, I want to be able to serve stale results via the cache when the backend fail during an outage.

More [context](https://github.com/vttale/serve-stale) on why one might want to serve stale data.

uruddarraju:
+1. I think this is an important feature to have if someone wants to run an entire dns infrastructure just on coredns, like we are trying to do.

JarenGlover:
It appears that we are checking if TTL is greater than 0 [here](https://github.com/coredns/coredns/blob/master/plugin/cache/cache.go#L173). If true, then it will never return from cache when TTL is 0.

miekg:
currently cache doesn't implement this feature. Would be nice though; like how unbound implemented it. Doesn't seem that hard actually

chrisohaver:
> like how unbound implemented it. Doesn't seem that hard actually.

Ah, yes, as explained in the bind description. Inferring from that, it does change behavior slightly; It will serve TTL 0 records occasionally when the backend is _not_ down.  I think this is a fairly minor behavior change, but probably should make the feature optional for that reason.  

Thinking on it: It would require that _cache_ continue to pass the query down the plugin chain when sending a TTL 0 record to a client, so the backend plugin can process it and write to cache if the backend is back up.  When the backend plugin writes its answer, we would need to know if an answer was already written to the client, and if so just write to cache without sending a second answer to the client.

soffchen:
> At a high level, I am trying to understand if coredns can serve stale data when the backend/upstreams fails.
> 
> An example of this can be find via [Unbound](https://nlnetlabs.nl/projects/unbound/about/)'s [serve-expired](http://manpages.ubuntu.com/manpages/bionic/man5/unbound.conf.5.html) configuration option.
> 
> ![image](https://user-images.githubusercontent.com/1428354/63303259-28627400-c294-11e9-9283-0eef70055361.png)
> 
> In short, I want to be able to serve stale results via the cache when the backend fail during an outage.
> 
[...]

PR #3466, opened 2019-11-18
Serve expired

gonzalop:
Implements #3158 by adding a _serve_expired_ option to the cache directive.
From unbound's man page:
>  serve-expired: <yes or no>
>               If  enabled,  unbound attempts to serve old responses from cache with a TTL of 0 in
>               the response without waiting for the  actual  resolution  to  finish.   The  actual
>               resolution answer ends up in the cache later on.  Default is "no".

#3158 

The documentation for the cache plugin would need to add documentation for the new option.

No.


gonzalop:
Ugh. Something went wrong with the merges. I'll retry later.

PR #3468, opened 2019-11-18
Add a serve_stale option for plugin/cache

gonzalop:
Implements #3158 by adding a _serve_stale_ option to the cache directive.
From unbound's man page:
>  serve-expired: <yes or no>
>               If  enabled,  unbound attempts to serve old responses from cache with a TTL of 0 in
>               the response without waiting for the  actual  resolution  to  finish.   The  actual
>               resolution answer ends up in the cache later on.  Default is "no".

#3158 

Yes, the PR includes a documentation update.

No.

### DIFF
--- a/plugin/cache/README.md
+++ b/plugin/cache/README.md
@@ -51,9 +51,10 @@ cache [TTL] [ZONES...] {
   **DURATION** defaults to 1m. Prefetching will happen when the TTL drops below **PERCENTAGE**,
   which defaults to `10%`, or latest 1 second before TTL expiration. Values should be in the range `[10%, 90%]`.
   Note the percent sign is mandatory. **PERCENTAGE** is treated as an `int`.
-* `serve_expired`, when set to yes, it will serve expired responses from cache when there are no healthy
-  upstream backends. The responses have a TTL of 0.  **DURATION** is how far back to consider stale
-  responses as fresh. The default is: 'serve\_expired no 1h'. Hint: the bigger the **CAPACITY** the more
+* `serve_expired`, when serve\_expired is set, CoreDNS always will serve an expired cache to a client if an
+  expired cache entry exists. When this happens, CoreDNS will attempt to refresh the cache entry after sending
+  the expired cache entry to the client. The responses have a TTL of 0. **DURATION** is how far back to consider
+  stale responses as fresh. The default is: 'serve\_expired no 1h'. Hint: the bigger the **CAPACITY** the more
   expired responses there will be in the cache.
 
 ## Capacity and Eviction
--- a/plugin/cache/README.md
+++ b/plugin/cache/README.md
@@ -34,7 +34,7 @@ cache [TTL] [ZONES...] {
     success CAPACITY [TTL] [MINTTL]
     denial CAPACITY [TTL] [MINTTL]
     prefetch AMOUNT [[DURATION] [PERCENTAGE%]]
-    serve_expired yes|no [DURATION]
+    serve_stale DURATION
 }
 ~~~
 
@@ -51,11 +51,10 @@ cache [TTL] [ZONES...] {
   **DURATION** defaults to 1m. Prefetching will happen when the TTL drops below **PERCENTAGE**,
   which defaults to `10%`, or latest 1 second before TTL expiration. Values should be in the range `[10%, 90%]`.
   Note the percent sign is mandatory. **PERCENTAGE** is treated as an `int`.
-* `serve_expired`, when serve\_expired is set, CoreDNS always will serve an expired cache to a client if an
-  expired cache entry exists. When this happens, CoreDNS will attempt to refresh the cache entry after sending
-  the expired cache entry to the client. The responses have a TTL of 0. **DURATION** is how far back to consider
-  stale responses as fresh. The default is: 'serve\_expired no 1h'. Hint: the bigger the **CAPACITY** the more
-  expired responses there will be in the cache.
+* `serve_stale`, when serve\_stale is set, cache always will serve an expired entry to a client if there is one
+  available.  When this happens, cache will attempt to refresh the cache entry after sending the expired cache
+  entry to the client. The responses have a TTL of 0. **DURATION** is how far back to consider
+  stale responses as fresh. The default is: 'serve\_stale 0h'.
 
 ## Capacity and Eviction
 
@@ -75,6 +74,7 @@ If monitoring is enabled (via the *prometheus* plugin) then the following metric
 * `coredns_cache_hits_total{server, type}` - Counter of cache hits by cache type.
 * `coredns_cache_misses_total{server}` - Counter of cache misses.
 * `coredns_cache_drops_total{server}` - Counter of dropped messages.
+* `coredns_cache_served_stale_total{server}` - Counter of requests served from stale cache entries.
 
 Cache types are either "denial" or "success". `Server` is the server handling the request, see the
 metrics plugin for documentation.
--- a/plugin/cache/README.md
+++ b/plugin/cache/README.md
@@ -34,7 +34,7 @@ cache [TTL] [ZONES...] {
     success CAPACITY [TTL] [MINTTL]
     denial CAPACITY [TTL] [MINTTL]
     prefetch AMOUNT [[DURATION] [PERCENTAGE%]]
-    serve_stale DURATION
+    serve_stale [DURATION]
 }
 ~~~
 
@@ -54,7 +54,7 @@ cache [TTL] [ZONES...] {
 * `serve_stale`, when serve\_stale is set, cache always will serve an expired entry to a client if there is one
   available.  When this happens, cache will attempt to refresh the cache entry after sending the expired cache
   entry to the client. The responses have a TTL of 0. **DURATION** is how far back to consider
-  stale responses as fresh. The default is: 'serve\_stale 0h'.
+  stale responses as fresh. The default duration is 1h.
 
 ## Capacity and Eviction
 
--- a/plugin/cache/README.md
+++ b/plugin/cache/README.md
@@ -34,6 +34,7 @@ cache [TTL] [ZONES...] {
     success CAPACITY [TTL] [MINTTL]
     denial CAPACITY [TTL] [MINTTL]
     prefetch AMOUNT [[DURATION] [PERCENTAGE%]]
+    serve_expired yes|no [DURATION]
 }
 ~~~
 
@@ -50,6 +51,10 @@ cache [TTL] [ZONES...] {
   **DURATION** defaults to 1m. Prefetching will happen when the TTL drops below **PERCENTAGE**,
   which defaults to `10%`, or latest 1 second before TTL expiration. Values should be in the range `[10%, 90%]`.
   Note the percent sign is mandatory. **PERCENTAGE** is treated as an `int`.
+* `serve_expired`, when set to yes, it will serve expired responses from cache when there are no healthy
+  upstream backends. The responses have a TTL of 0.  **DURATION** is how far back to consider stale
+  responses as fresh. The default is: 'serve\_expired no 1h'. Hint: the bigger the **CAPACITY** the more
+  expired responses there will be in the cache.
 
 ## Capacity and Eviction
 
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -36,6 +36,9 @@ type Cache struct {
 	duration   time.Duration
 	percentage int
 
+	// Whether to serve cache stale records when upstream is down.
+	serveExpired bool
+
 	// Testing.
 	now func() time.Time
 }
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -36,8 +36,8 @@ type Cache struct {
 	duration   time.Duration
 	percentage int
 
-	serveExpired bool
-	expiredUpTo  time.Duration
+	serveStale bool
+	staleUpTo  time.Duration
 
 	// Testing.
 	now func() time.Time
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -36,8 +36,7 @@ type Cache struct {
 	duration   time.Duration
 	percentage int
 
-	serveStale bool
-	staleUpTo  time.Duration
+	staleUpTo time.Duration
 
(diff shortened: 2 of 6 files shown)


plugin/cache/handler.go
```diff
@@ -27,18 +27,33 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
-		return dns.RcodeSuccess, nil
+		servedExpired.WithLabelValues(server).Inc()
+		// Adjust the time to get a 0 TTL in the reply we got from a stale item.
+		t := i.ttl(now)
+		if t < 0 {
+			now = now.Add(time.Duration(t) * time.Second)
+		}
+		go func() {
+			r := r.Copy()
+			crr := &ResponseWriter{Cache: c, state: state, server: server, prefetch: true, remoteAddr: w.LocalAddr()}
```
chrisohaver, line 44: IIUC, this go routine is the "ends up in the cache later on" bit.
Will this cause CoreDNS to send a DNS response packet to itself?

gonzalop: No, nothing is written to the response when prefetch is true.
The LocalAddr is arbitrary and is only needed to avoid a nullref in  ResponseWriter.RemoteAddr (it is not called, but I did not want to take any chances). Once a backend returns a cacheable response, it is cached but never written.

plugin/cache/README.md
```diff
@@ -50,6 +51,10 @@ cache [TTL] [ZONES...] {
   **DURATION** defaults to 1m. Prefetching will happen when the TTL drops below **PERCENTAGE**,
   which defaults to `10%`, or latest 1 second before TTL expiration. Values should be in the range `[10%, 90%]`.
   Note the percent sign is mandatory. **PERCENTAGE** is treated as an `int`.
+* `serve_expired`, when set to yes, it will serve expired responses from cache when there are no healthy
+  upstream backends. The responses have a TTL of 0.  **DURATION** is how far back to consider stale
```
chrisohaver: "... when there are no healthy upstream backends."

IIUC, we are not literally watching the health of the upstream backends here.  So this may be misleading.  One might infer that if an expired entry is served to the client, that must mean that the backend/upstream is unhealthy/non-responsive - but that is not always the case.  The actual behavior may be harder to succinctly describe, but IMO we should try...

When serve_expired is set, CoreDNS always will serve an expired cache to a client if an expired cache entry exists.  When this is happens, CoreDNS will attempt to refresh the cache entry after sending the expired cache entry to the client.

gonzalop: You are correct.
I've updated  the text with your more accurate description. Let me know if there's anything that looks off.

Thanks

-Gonzalo

chrisohaver (approved):
Thanks for the contribution! LGTM. Although IMO we should get an approval from one of the plugin OWNERS before merging ...

gonzalop:
Ack. Thank you, Chris, for reviewing.

plugin/cache/README.md
```diff
@@ -50,6 +51,11 @@ cache [TTL] [ZONES...] {
   **DURATION** defaults to 1m. Prefetching will happen when the TTL drops below **PERCENTAGE**,
   which defaults to `10%`, or latest 1 second before TTL expiration. Values should be in the range `[10%, 90%]`.
   Note the percent sign is mandatory. **PERCENTAGE** is treated as an `int`.
+* `serve_expired`, when serve\_expired is set, CoreDNS always will serve an expired cache to a client if an
```
miekg: s/CoreDNS/*cache*/

plugin/cache/README.md
```diff
@@ -50,6 +51,11 @@ cache [TTL] [ZONES...] {
...
+  expired cache entry exists. When this happens, CoreDNS will attempt to refresh the cache entry after sending
+  the expired cache entry to the client. The responses have a TTL of 0. **DURATION** is how far back to consider
+  stale responses as fresh. The default is: 'serve\_expired no 1h'. Hint: the bigger the **CAPACITY** the more
```
miekg: I would remove the hint, sounds a bit pedantic

plugin/cache/README.md
```diff
@@ -34,6 +34,7 @@ cache [TTL] [ZONES...] {
     success CAPACITY [TTL] [MINTTL]
     denial CAPACITY [TTL] [MINTTL]
     prefetch AMOUNT [[DURATION] [PERCENTAGE%]]
+    serve_expired yes|no [DURATION]
```
miekg: the new metrics should also be added to the README

plugin/cache/handler.go
```diff
@@ -27,18 +27,33 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
 	i, found := c.get(now, state, server)
-	if i != nil && found {
-		resp := i.toMsg(r, now)
-		w.WriteMsg(resp)
-
-		if c.shouldPrefetch(i, now) {
-			go c.doPrefetch(ctx, state, server, i, now)
+	if i == nil || !found {
+		if c.serveExpired {
+			i, found = c.get(now.Add(-c.expiredUpTo), state, server)
```
miekg: why do we need to reget the item; we've already looked it up. It's it more natural to have `c.get` return an boolean that includes that's it expired?
Or maybe a new function getExpired, but we control the intree code so we can upgrade all callers.

plugin/cache/setup_test.go
```diff
@@ -9,72 +9,101 @@ import (
...
+		input                string
+		shouldErr            bool
+		expectedNcap         int
+		expectedPcap         int
+		expectedNttl         time.Duration
+		expectedMinNttl      time.Duration
+		expectedPttl         time.Duration
+		expectedMinPttl      time.Duration
+		expectedPrefetch     int
+		expectedServeExpired bool
```
miekg: just put these checks in a small new test that just checks the new elements. Now we just have a giant list of `false` and a few `true`'s that check the new setup code. A new selfcontained small test is almost certainly better.

miekg (changes requested):

plugin/cache/README.md
stp-ip: Also: This be true|false instead of yes|no as far as I can see.

stp-ip: Couldn't we assume serve_expired to be false by default and therefore not need the true|false at all?
This would simplify the config to 'serve_expired [duration]'
Also generally called 'stale' responses instead of expired.

plugin/cache/README.md
gonzalop: Done.

plugin/cache/README.md
gonzalop: This is something I would have never thought of without external feedback, but you are right. Removed. Thanks.

plugin/cache/README.md
gonzalop: I've added the metrics. renamed the option to serve_stale and got rid of the yes|no argument.
Thanks! :)

plugin/cache/handler.go
gonzalop: You are right.
I added a getItem to unconditionally get an item from the cache. PTAL.

plugin/cache/setup_test.go
gonzalop: Done. A lot cleaner now.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add a serve_e..." ]
>stp-ip commented on this pull request.
>
>
>
>> @@ -34,6 +34,7 @@ cache [TTL] [ZONES...] {
>     success CAPACITY [TTL] [MINTTL]
>     denial CAPACITY [TTL] [MINTTL]
>     prefetch AMOUNT [[DURATION] [PERCENTAGE%]]
>+    serve_expired yes|no [DURATION]
>
>Also: This be true|false instead of yes|no as far as I can see.

or just the presence of the string `serve_expired`, so the boolean is 
superfluous.

plugin/cache/README.md
```diff
@@ -50,6 +51,10 @@ cache [TTL] [ZONES...] {
   **DURATION** defaults to 1m. Prefetching will happen when the TTL drops below **PERCENTAGE**,
   which defaults to `10%`, or latest 1 second before TTL expiration. Values should be in the range `[10%, 90%]`.
   Note the percent sign is mandatory. **PERCENTAGE** is treated as an `int`.
+* `serve_stale`, when serve\_stale is set, cache always will serve an expired entry to a client if there is one
+  available.  When this happens, cache will attempt to refresh the cache entry after sending the expired cache
+  entry to the client. The responses have a TTL of 0. **DURATION** is how far back to consider
+  stale responses as fresh. The default is: 'serve\_stale 0h'.
```
miekg: `0h` or did you mean `1h` here?

plugin/cache/cache.go
```diff
@@ -36,6 +36,9 @@ type Cache struct {
 	duration   time.Duration
 	percentage int
 
+	serveStale bool
+	staleUpTo  time.Duration
```
miekg: these two values can be unified into 1: if staleUpTo is zero it is disabled, otherwise we use it.

plugin/cache/handler.go
```diff
@@ -83,6 +99,18 @@ func (c *Cache) get(now time.Time, state request.Request, server string) (*item,
 	return nil, false
 }
 
+func (c *Cache) getItem(state request.Request, server string) *item {
```
miekg: getIgnoreTTL or something? getItem is very simular to get. Might also benefit from some documentation :)

plugin/cache/setup.go
```diff
@@ -176,6 +177,19 @@ func cacheParse(c *caddy.Controller) (*Cache, error) {
...
+				args := c.RemainingArgs()
+				if len(args) != 1 {
+					return nil, errors.New("syntax is: serve_stale <duration>")
+				}
+				d, err := time.ParseDuration(args[0])
+				if err != nil {
+					return nil, fmt.Errorf("invalid duration: %v", args[0])
+				}
+				if d < 0 {
+					return nil, errors.New("negative duration does not make sense")
```
miekg: "invalid negative duration for serve_stale"

plugin/cache/setup.go
```diff
@@ -176,6 +177,19 @@ func cacheParse(c *caddy.Controller) (*Cache, error) {
...
 				}
 
+			case "serve_stale":
+				args := c.RemainingArgs()
+				if len(args) != 1 {
+					return nil, errors.New("syntax is: serve_stale <duration>")
+				}
+				d, err := time.ParseDuration(args[0])
+				if err != nil {
+					return nil, fmt.Errorf("invalid duration: %v", args[0])
```
miekg: just return `err` as in the other parseDurations here.

plugin/cache/setup.go
```diff
@@ -176,6 +177,19 @@ func cacheParse(c *caddy.Controller) (*Cache, error) {
 					ca.percentage = num
 				}
 
+			case "serve_stale":
+				args := c.RemainingArgs()
+				if len(args) != 1 {
+					return nil, errors.New("syntax is: serve_stale <duration>")
```
miekg: return return nil, c.ArgErr()

miekg (commented):
looks much better, thanks, still some (style) comments though

plugin/cache/README.md
gonzalop: I'm removing that. There is no default now and you have to specify the duration.

plugin/cache/cache.go
gonzalop: How would you know when to serve immediately + fetch vs the normal behavior when the options is not present?

plugin/cache/README.md
chrisohaver: Per setup.go, looks like duration is optional and default is 1h.

plugin/cache/setup.go
gonzalop: Done. Thanks.

plugin/cache/setup.go
gonzalop: done

plugin/cache/handler.go
gonzalop: Done.

...but I noticed the 2 metrics in the old get method would now never be incremented. I am now incrementing those metrics as well, PTAL.

plugin/cache/setup.go
gonzalop: done

plugin/cache/README.md
gonzalop: Yup, it's updated in my last push.

plugin/cache/cache.go
miekg: isn't `serveStale == staleUpTo.IsZero()` ?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add a serve_s..." ]
>gonzalop commented on this pull request.
>
>
>
>> @@ -83,6 +99,18 @@ func (c *Cache) get(now time.Time, state request.Request, server string) (*item,
> 	return nil, false
> }
>
>+func (c *Cache) getItem(state request.Request, server string) *item {
>
>I noticed the 2 metrics in the old get method would now never be incremented. Only getIgnoredTTL knows where the item came from (Denial, Success). What would be the right thing to do here?

file an issue, please? Leave it for now - bug fixing and features should be kept 
in separate PRs if possible.

plugin/cache/cache.go
gonzalop: If a configuration does not include serve_stale, we would have enabled the feature for ttl == 0.

gonzalop: It breaks a prefetch test that expects to trigger a fetch after 30s for a TTL=30.

gonzalop: Done.
I got waaaaaay derailed by one of my tests but in the end you were right and it was trivial to remove it. Thanks.

gonzalop:
Anything else I should do?

miekg:
I'm trying out the new and improved corbot (github.com/miekg/dreck)

/lgtm
/merge

miekg:
hmmm the lgtm led to a 
~~~
[ERROR] dreck: POST https://api.github.com/repos/coredns/coredns/pulls/3468/reviews: 422 Unprocessable Entity [{Resource: Field: Code: Message:}]
~~~
which is REST for 'no' ?

Issue #3101, opened 2019-08-08
plugin/route53: early NODATA returns for overlapping domains with other record types


mattlqx:
A side-effect of fixing #2309 appears to be that if a Route 53 zone contains any record types of a name subsequent configured zones will not be consulted, resulting in NODATA/NOERROR results even if subsequent zones *do* have a record for the type.

It's easier to illustrate than try and explain...

Given this config...
```
proofofconcept.me {
  route53 proofofconcept.me.:Z3M5VAIPSRDR8Z proofofconcept.me.:Z1OMBJKY7Y7LLQ {
    credentials default /configs/credentials
  }
  errors
}

. {
  loop
  forward . 127.0.0.1:253
}
```

Z3M5VAIPSRDR8Z is private, Z1OMBJKY7Y7LLQ is public.

Both zones contain A records for proofofconcept.me, but only the public zone contains an MX record. In version 1.4.0, an MX query for proofofconcept.me would return the MX record from the public zone. But in version 1.5.0, the same query returns NOERROR/NODATA.

I've bisected to confirm it was commit 1e150674c50a10b0665f6c4270a6c58f833e6ef6 that changed this behavior.

Working:
```
$ dig -t mx proofofconcept.me @127.0.0.1

; <<>> DiG 9.11.3-1ubuntu1.8-Ubuntu <<>> -t mx proofofconcept.me @127.0.0.1
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 56421
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 4, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
; COOKIE: 010053bd692af627 (echoed)
;; QUESTION SECTION:
;proofofconcept.me.		IN	MX

;; ANSWER SECTION:
proofofconcept.me.	300	IN	MX	10 an.mx.host.

;; AUTHORITY SECTION:
proofofconcept.me.	172800	IN	NS	ns-920.awsdns-51.net.
proofofconcept.me.	172800	IN	NS	ns-1920.awsdns-48.co.uk.
proofofconcept.me.	172800	IN	NS	ns-1132.awsdns-13.org.
proofofconcept.me.	172800	IN	NS	ns-421.awsdns-52.com.

;; Query time: 0 msec
[...]

PR #3470, opened 2019-11-19
plugin/route53: fix NODATA handling

ykhr53:
I've made a fix for NODATA handling in multiple hosted zones.
Request for comments!

https://github.com/coredns/coredns/issues/3101

I don't think so.

I don't think so.

### DIFF
--- a/plugin/route53/route53.go
+++ b/plugin/route53/route53.go
@@ -117,16 +117,26 @@ func (h *Route53) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	m.SetReply(r)
 	m.Authoritative = true
 	var result file.Result
+
+	tmpMsg := new(dns.Msg)
+	var tmpResult file.Result
+	isNodata := false
+
 	for _, hostedZone := range z {
 		h.zMu.RLock()
 		m.Answer, m.Ns, m.Extra, result = hostedZone.z.Lookup(ctx, state, qname)
 		h.zMu.RUnlock()
 
-		// Take the answer if it's non-empty OR if there is another
-		// record type exists for this name (NODATA).
-		if len(m.Answer) != 0 || result == file.NoData {
+		if len(m.Answer) != 0 {
 			break
+		} else if result != file.NameError {
+			tmpMsg.Answer, tmpMsg.Ns, tmpMsg.Extra, tmpResult = m.Answer, m.Ns, m.Extra, result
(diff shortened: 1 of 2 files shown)


miekg:
on first sight this looks a bit hackish and I feel there must be a cleaner solution as the original code already takes NODATA into consideration.
(Also this code is cut and pasted in googlcloud and azure if I'm not mistaken, so they may have the same failure)

ykhr53:
Thank you for the comment!
yeah, it does...
The reason why the code is a bit messy is we need to keep *NODATA data* in order to check the next zone.
I'll think about cleaner approach.

plugin/route53/route53.go
```diff
@@ -117,18 +117,28 @@ func (h *Route53) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
...
 			break
+		} else if result != file.NameError {
+			tmpMsg.Answer, tmpMsg.Ns, tmpMsg.Extra, tmpResult = m.Answer, m.Ns, m.Extra, result
+			isNodata = true
 		}
 	}
 
+	if isNodata && result != file.Success {
+		m.Answer, m.Ns, m.Extra, result = tmpMsg.Answer, tmpMsg.Ns, tmpMsg.Extra, tmpResult
+	}
```
ykhr53, line 140: How about something like this?
I added some comments and changed some styles.
The concept itself isn't changed though.

```
		// If there is an answer, it will be returned immediately.
		if len(m.Answer) != 0 {
			break
		}

		// Only NXDOMAIN doesn't need to be stored.
		// Other cases such as NODATA, Delegation, SERVFAIL should be respected.
		if result != file.NameError {
			tmpMsg.Answer, tmpMsg.Ns, tmpMsg.Extra, tmpResult = m.Answer, m.Ns, m.Extra, result
			isNodata = true
		}
	}

	// If the first zone returns NODATA and the second zone does not return an answer,
	// the first NODATA message will be used.
	if isNodata && result != file.Success {
		m.Answer, m.Ns, m.Extra, result = tmpMsg.Answer, tmpMsg.Ns, tmpMsg.Extra, tmpResult
	}
```

ykhr53:
Hi,
Could you give me your opinion of this change, @yongtang @dilyevsky ?
I think I should wait for being merged the PR 3442 anyway.

miekg:
closing because not actively discussed.

Issue #3464, opened 2019-11-18
use pkg/reuseport in more plugins


miekg:
Port the reuseport socket handling to all plugins to open sockets, i.e. one that comes to mind if `ready`, but there are others. 

See PR https://github.com/coredns/coredns/pull/3454 for an example. This should lead to massive simplification; less race conditions and easier to understand code .

miekg:
metrics can be fixed here. May need more work though to some crazyness we do there.
dnstest as some Listen's as well; would make testing more robust to use reuseport.Listen there as well

miekg:
these are left:
~~~
plugin/pkg/dnstest/server.go:		s2.Listener, _ = net.Listen("tcp", ":0")
plugin/metrics/metrics.go:	ln, err := net.Listen("tcp", m.Addr)
plugin/dnstap/dnstapio/io_test.go:		l, err := net.Listen(param[0], param[1])
plugin/dnstap/dnstapio/io_test.go:	l, err := net.Listen("tcp", endpointTCP)
plugin/dnstap/dnstapio/io_test.go:	l, err := net.Listen("tcp", endpointTCP)
plugin/dnstap/dnstapio/io_test.go:	l, err = net.Listen("tcp", addr)
~~~

PR #3473, opened 2019-11-20
Fix reloading in health and ready

zouyee:

is called for the old ones.
a part of https://github.com/coredns/coredns/issues/3464
xref https://github.com/coredns/coredns/pull/3454
None
None

### DIFF
--- a/plugin/health/health.go
+++ b/plugin/health/health.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	clog "github.com/coredns/coredns/plugin/pkg/log"
+	"github.com/coredns/coredns/plugin/pkg/reuseport"
 )
 
 var log = clog.NewWithPlugin("health")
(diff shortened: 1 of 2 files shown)


plugin/health/health.go
```diff
@@ -29,8 +30,10 @@ func (h *health) OnStartup() error {
 		h.Addr = ":8080"
 	}
 	h.stop = make(chan bool)
-
-	ln, err := net.Listen("tcp", h.Addr)
+	// Reloading the plugin without changing the listening address results
+	// in an error unless we reuse the port because Startup is called for
+	// new handlers before Shutdown is called for the old ones.
```
miekg: you don't have to repeat the comment everywhere

zouyee: Done

miekg:
done a bit of minimal manual testing by reloading and curling the health endpoint.
/lgtm

PR #3476, opened 2019-11-21
Doc change: gen -> generate

blaines:
`go gen` doesn't seem to be valid. I think this is the correct intended command!


miekg:
indeed, go gen is a local shortcut of mine.

Can you sign your commit to satisfy the dco requirement? (git commit -s)

PR #3479, opened 2019-11-24
plugin/sign: fix signing of authoritative data

### LINKED ISSUE #3354: plugin/sign: add Walk to exclude children and glue
The Walk function used in *sign* currently signs glue (is has a BUG section in the README). Fix this to exclude it, so we currently sign any zone - even those with subdomains.

miekg:
Don't sign data we are not authoritative for. This adds an AuthWalk
which skips names we should not authoritative for. Adds a few tests to
check this is the case. Generates zones have been compared to
dnssec-signzone.

A number of changes have been made:

* don't add DS records to the apex
* NSEC TTL is the SOA's minttl value (copying bind9)
* Various cleanups

Fixes: #3354

Bulk of the code is new tests cases. The gist is file/tree/authwalk.go and sign/signer.go

### DIFF
--- a/plugin/file/tree/auth_walk.go
+++ b/plugin/file/tree/auth_walk.go
@@ -1,0 +1,58 @@
+package tree
+
+import (
+	"github.com/miekg/dns"
+)
+
+// AuthWalk performs fn on all authoritative values stored in the tree in
+// pre-order depth first. If a non-nil error is returned the AuthWalk was interrupted
+// by an fn returning that error. If fn alters stored values' sort
+// relationships, future tree operation behaviors are undefined.
+//
+// The fn function will be called with 3 arguments, the current element, a map containing all
+// the RRs for this element and a boolean if this name is considered authoritative.
+func (t *Tree) AuthWalk(fn func(*Elem, map[uint16][]dns.RR, bool) error) error {
+	if t.Root == nil {
+		return nil
+	}
+	return t.Root.authwalk(make(map[string]struct{}), fn)
+}
+
+func (n *Node) authwalk(ns map[string]struct{}, fn func(*Elem, map[uint16][]dns.RR, bool) error) error {
+	if n.Left != nil {
+		if err := n.Left.authwalk(ns, fn); err != nil {
+			return err
+		}
+	}
+
+	// Check if the current name is a subdomain of *any* of the delegated names we've seen, if so, skip this name.
+	// The ordering of the tree and how we walk if guarantees we see parents first.
+	if n.Elem.Type(dns.TypeNS) != nil {
+		ns[n.Elem.Name()] = struct{}{}
+	}
+
+	auth := true
+	i := 0
+	for {
+		j, end := dns.NextLabel(n.Elem.Name(), i)
+		if end {
+			break
+		}
+		if _, ok := ns[n.Elem.Name()[j:]]; ok {
+			auth = false
+			break
+		}
+		i++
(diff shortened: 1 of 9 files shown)


plugin/file/tree/walk.go
```diff
@@ -2,25 +2,28 @@ package tree
 
 import "github.com/miekg/dns"
 
-// Walk performs fn on all values stored in the tree. If a non-nil error is returned the
-// Walk was interrupted by an fn returning that error. If fn alters stored values' sort
+// Walk performs fn on all authoritative values stored in the tree in
+// pre-order depth first. If a non-nil error is returned the Walk was interrupted
```
chrisohaver:
```suggestion
// in-order depth first. If a non-nil error is returned the Walk was interrupted
```

chrisohaver:
@miekg, I'm finding I need to familiarize my self with DNSSEC before I can understand the PR and do a full/fair review.
This will slow my review.  If you'd prefer a faster code review, feel free to rope in others.

plugin/file/tree/auth_walk.go
```diff
@@ -0,0 +1,58 @@
+package tree
+
+import (
+	"github.com/miekg/dns"
+)
+
+// AuthWalk performs fn on all authoritative values stored in the tree in
+// pre-order depth first. If a non-nil error is returned the AuthWalk was interrupted
```
chrisohaver:
```suggestion
// in-order depth first. If a non-nil error is returned the AuthWalk was interrupted
```

chrisohaver, line 8:
```suggestion
// in-order depth first. If a non-nil error is returned the AuthWalk was interrupted
```

plugin/file/tree/auth_walk.go
```diff
@@ -0,0 +1,58 @@
...
+func (n *Node) authwalk(ns map[string]struct{}, fn func(*Elem, map[uint16][]dns.RR, bool) error) error {
+	if n.Left != nil {
+		if err := n.Left.authwalk(ns, fn); err != nil {
+			return err
+		}
+	}
+
+	// Check if the current name is a subdomain of *any* of the delegated names we've seen, if so, skip this name.
+	// The ordering of the tree and how we walk if guarantees we see parents first.
+	if n.Elem.Type(dns.TypeNS) != nil {
```
chrisohaver, line 30: Will this tag the origin/apex as delegated?  Does this need check to see if `n.Elem.Name() == origin`?
maybe by checking for an SOA? e.g. `n.Elem.Type(dns.TypeNS) != nil && n.Elem.Type(dns.TypeSOA) == nil`

miekg: Yes, this is a bit icky; i will not by virtue for zones as these apex records somewhere else: https://github.com/coredns/coredns/blob/master/plugin/file/zone.go#L24

chrisohaver (approved):
LGTM - Although I don't yet fully understand NSEC type bitmaps, and therefore can't verify that change.  But otherwise LGTM.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/sign: ..." ]
>
>LGTM - Although I don't yet fully understand NSEC type bitmaps, and therefore can't verify that change.  But otherwise LGTM.

Thanks. I'm running this on my personal server and done some checks. I'll
inspect more, but it should be good (fingers crossed).

PR #3491, opened 2019-11-29
Update governance docs to reflect CODEOWNERS

miekg:
NO GOVERNANCE changes made!

Point to the correct files, now that I merged CODEOWNERS, OWNERS files
are gone and there is also only 1 CODEOWNERS.


miekg:
ha, lol 12 pending reviewers - this may not really what we want :) Probably can tweak CODEOWNERS a bit to spread this out, server/ stuff could be split etc. so toplevel changes are relatively rare.

miekg:
merging because governance docs are now out of date with actual impl, as said no changes in the project

PR #3492, opened 2019-11-29
use pkg/reuseport in rest plugins

### LINKED ISSUE #3464: use pkg/reuseport in more plugins

zouyee:

is called for the old ones.
fix https://github.com/coredns/coredns/issues/3464
xref https://github.com/coredns/coredns/pull/3454
None
None

### DIFF
--- a/plugin/dnstap/dnstapio/io_test.go
+++ b/plugin/dnstap/dnstapio/io_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/coredns/coredns/plugin/pkg/reuseport"
+	
 	tap "github.com/dnstap/golang-dnstap"
 	fs "github.com/farsightsec/golang-framestream"
 )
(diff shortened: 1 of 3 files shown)


plugin/dnstap/dnstapio/io_test.go
```diff
@@ -1,13 +1,13 @@
 package dnstapio
 
 import (
+	"github.com/coredns/coredns/plugin/pkg/reuseport"
+	tap "github.com/dnstap/golang-dnstap"
+	fs "github.com/farsightsec/golang-framestream"
```
miekg: the fmt here should be std lib, newline, our stuff, newline, external deps

plugin/pkg/dnstest/server.go
```diff
@@ -4,6 +4,7 @@ import (
 	"net"
 
 	"github.com/miekg/dns"
+	"github.com/coredns/coredns/plugin/pkg/reuseport"
```
miekg, line 6: see prev. comment

chrisohaver (approved):
lgtm

zouyee:
@miekg PTAL

miekg (approved):

PR #3503, opened 2019-12-06
redirect handler for pprof index

### LINKED ISSUE #3504: handler  redirect to pprof index

handler `/debug/pprof` redirect to `/debug/pprof/`

**Why is this needed**:

zouyee:
Signed-off-by: zouyee <zounengren@cmss.chinamobile.com>

handler `/debug/pprof` redirect to `/debug/pprof/`

fix https://github.com/coredns/coredns/issues/3504

NONE

NONE

### DIFF
--- a/plugin/pprof/pprof.go
+++ b/plugin/pprof/pprof.go
@@ -31,6 +31,9 @@ func (h *handler) Startup() error {
 	h.ln = ln
 
 	h.mux = http.NewServeMux()
+	h.mux.HandleFunc(path, func(rw http.ResponseWriter, req *http.Request) {
+		http.Redirect(rw, req, path+"/", http.StatusFound)
+	})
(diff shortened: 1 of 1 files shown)


plugin/pprof/pprof.go
```diff
@@ -31,6 +31,7 @@ func (h *handler) Startup() error {
 	h.ln = ln
 
 	h.mux = http.NewServeMux()
+	h.mux.HandleFunc(path, redirectTo(path+"/"))
```
miekg: I can see this being handy, but a whole `redirectTo` function seems overkill. I would just inline the handler directly here.

zouyee: Done, thanks! 

> I can see this being handy, but a whole `redirectTo` function seems overkill. I would just inline the handler directly here.

miekg (approved):

Issue #3509, opened 2019-12-07
add: go mod tidy github action


miekg:
see https://github.com/transcom/mymove/pull/2984

We should add something similar to automatically run this on (all?) PRs?

yongtang:
I can take a look at this one.

yongtang:
Added PR #3512

yongtang:
One glitch is the associated user of the commit: https://github.com/coredns/coredns/commit/35953235cf6138afd95bff3bf6b54465c66c5372 https://github.com/coredns/coredns/commit/a2481250e0bfeaa48159b744db62c8e0128415d4

It ties the user to the last commit's user. This could get annoying as technically the commit belongs to `go-mod-tidy[bot]`

I think ideally we want to tie the user to `go-mod-tidy[bot]`, similar to `dependabot-preview[bot]`.

Will take another look to see how to remove the tied user instead.

yongtang:
Added PR #3517 so that now commits goes to `coredns-go-mod-tidy[bot]` instead of real user.

yongtang:
Thanks @miekg. I played GitHub workflow actions a little further, looks like GitHub workflow actions by itself could replace Travis and Circle, as it is a CI by itself. There are also some bot workload that could just be replaced by GitHub workflow actions.

Maybe it is time to consolidate Travis and Circle into GitHub workflow?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] add: go mod t..." ]
>Thanks @miekg. I played GitHub workflow actions a little further, looks like GitHub workflow actions by itself could replace Travis and Circle, as it is a CI by itself. There are also some bot workload that could just be replaced by GitHub workflow actions.
>
>Maybe it is time to consolidate Travis and Circle into GitHub workflow?

thanks for bringing this up; yes this idea is also floating around in my mind.

But it's one of those things that end up quite a bit of work and the end the
result is basically the same (i.e. we have tests), so unless we have a good
reason to do this now I would just defer this.

miekg:
looks done (from the comments - didn't double check)

PR #3512, opened 2019-12-07
Setup GitHub Workflow for auto `go tidy`

yongtang:

Setup GitHub Workflow for auto `go tidy`, when
- 'go.mod'
- 'go.sum'
- '.github/workflows/go.tidy.yml'
has been touched.

Note it covers both `dependabot[bot]` and non `dependabot` (manual touching those two files) case.

#3509

### DIFF
--- a/.github/workflows/go.tidy.yml
+++ b/.github/workflows/go.tidy.yml
@@ -1,0 +1,43 @@
+name: go tidy
+
+on:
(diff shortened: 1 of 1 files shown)


miekg:
cool! Can't actually review because I don't know anything about github actions, but doesn't look all that dangerous.

PR #3516, opened 2019-12-07
Add private DNS support for azure plugin

### LINKED ISSUE #3113: plugin/azure: add support for private zones
Azure supports hosting private dns zones, `plugin/azure` should support it. 
https://docs.microsoft.com/en-us/azure/dns/private-dns-overview

darshanime:
Fixes a bug in `azure` plugin introduced during refactor

closes https://github.com/coredns/coredns/issues/3113

Updated the readme

Yes, changes the syntax for `azure` plugin block

### DIFF
--- a/plugin/azure/README.md
+++ b/plugin/azure/README.md
@@ -13,7 +13,7 @@ record types.
 ## Syntax
 
 ~~~ txt
-azure RESOURCE_GROUP:ZONE... {
+azure RESOURCE_GROUP:ZONE:ACCESS... {
     tenant TENANT_ID
     client CLIENT_ID
     secret CLIENT_SECRET
@@ -23,8 +23,8 @@ azure RESOURCE_GROUP:ZONE... {
 }
 ~~~
 
-*   **RESOURCE_GROUP:ZONE** is the resource group to which the hosted zones belongs on Azure,
-    and  **ZONE** the zone that contains data.
+*   **RESOURCE_GROUP:ZONE:ACCESS** is the resource group to which the hosted zones belongs on Azure,
+    **ZONE** the zone that contains data and **ACCESS** is either `public` or `private`.
 
 *   **CLIENT_ID** and **CLIENT_SECRET** are the credentials for Azure, and `tenant` specifies the
     **TENANT_ID** to be used. **SUBSCRIPTION_ID** is the subscription ID. All of these are needed
@@ -38,11 +38,11 @@ azure RESOURCE_GROUP:ZONE... {
 
 ## Examples
 
-Enable the *azure* plugin with Azure credentials for the zone `example.org`:
+Enable the *azure* plugin with Azure credentials for public zone `example.org` and for private zone `example.private`:
 
 ~~~ txt
 example.org {
-    azure resource_group_foo:example.org {
+    azure resource_group_foo:example.org:public resource_group_foo:example.private:private {
       tenant 123abc-123abc-123abc-123abc
       client 123abc-123abc-123abc-234xyz
       subscription 123abc-123abc-123abc-563abc
--- a/plugin/azure/README.md
+++ b/plugin/azure/README.md
@@ -13,18 +13,19 @@ record types.
 ## Syntax
 
 ~~~ txt
-azure RESOURCE_GROUP:ZONE:ACCESS... {
+azure RESOURCE_GROUP:ZONE... {
     tenant TENANT_ID
     client CLIENT_ID
     secret CLIENT_SECRET
     subscription SUBSCRIPTION_ID
     environment ENVIRONMENT
     fallthrough [ZONES...]
+    access private
 }
 ~~~
 
-*   **RESOURCE_GROUP:ZONE:ACCESS** is the resource group to which the hosted zones belongs on Azure,
-    **ZONE** the zone that contains data and **ACCESS** is either `public` or `private`.
+*   **RESOURCE_GROUP:ZONE** is the resource group to which the hosted zones belongs on Azure,
+    **ZONE** the zone that contains data
 
 *   **CLIENT_ID** and **CLIENT_SECRET** are the credentials for Azure, and `tenant` specifies the
     **TENANT_ID** to be used. **SUBSCRIPTION_ID** is the subscription ID. All of these are needed
@@ -36,17 +37,20 @@ azure RESOURCE_GROUP:ZONE:ACCESS... {
     If **ZONES** is omitted, then fallthrough happens for all zones for which the plugin is
     authoritative.
 
+*   `access`  specifies if the zone is `public` or `private`. Default is `public`.
+
 ## Examples
 
 Enable the *azure* plugin with Azure credentials for public zone `example.org` and for private zone `example.private`:
 
 ~~~ txt
 example.org {
-    azure resource_group_foo:example.org:public resource_group_foo:example.private:private {
+    azure resource_group_foo:example.org resource_group_foo:example.private {
       tenant 123abc-123abc-123abc-123abc
       client 123abc-123abc-123abc-234xyz
       subscription 123abc-123abc-123abc-563abc
       secret mysecret
+      access private
     }
 }
 ~~~
--- a/plugin/azure/README.md
+++ b/plugin/azure/README.md
@@ -25,7 +25,7 @@ azure RESOURCE_GROUP:ZONE... {
 ~~~
 
 *   **RESOURCE_GROUP:ZONE** is the resource group to which the hosted zones belongs on Azure,
-    **ZONE** the zone that contains data
+    and **ZONE** the zone that contains data.
 
 *   **CLIENT_ID** and **CLIENT_SECRET** are the credentials for Azure, and `tenant` specifies the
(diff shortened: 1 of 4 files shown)


plugin/azure/README.md
```diff
@@ -13,7 +13,7 @@ record types.
 ## Syntax
 
 ~~~ txt
-azure RESOURCE_GROUP:ZONE... {
+azure RESOURCE_GROUP:ZONE:ACCESS... {
```
miekg: I would put this in the options below, by default it's public, unless you have `private` as an option

darshanime: In that case, there will have to be a separate azure block for private and public zones. Currently it looks like this - https://github.com/coredns/coredns/pull/3516/files#diff-694fec5c821a1b8f941cbb681c604bdeR45

What do you think?

miekg: but then just allow 'private, public and both' as an option?

darshanime: sorry I did not fully understand.
Are you suggesting - having something like:

```
    azure resource_group_foo:example.org:both {
      tenant 123abc-123abc-123abc-123abc
      client 123abc-123abc-123abc-234xyz
      subscription 123abc-123abc-123abc-563abc
      secret mysecret
    }
```

or

```
    azure resource_group_foo:example.org resource_group_foo:example.private {
      tenant 123abc-123abc-123abc-123abc
      client 123abc-123abc-123abc-234xyz
      subscription 123abc-123abc-123abc-563abc
      secret mysecret
      access both
    }
```

Or neither? 

`azure-sdk-for-go` treats the public zones and private zones as two distinct and disjoint types, with different API to interact with them and hence we need to identify which is which. So, both the above won't work.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add private D..." ]
>```
>    azure resource_group_foo:example.org resource_group_foo:example.private {
>      tenant 123abc-123abc-123abc-123abc
>      client 123abc-123abc-123abc-234xyz
>      subscription 123abc-123abc-123abc-563abc
>      secret mysecret
>      access both
>    }
>```

yes some like this, I would remove the `access` and just have `private` and `public` as
keywords. Thus:

```
    azure resource_group_foo:example.org resource_group_foo:example.private {
      tenant 123abc-123abc-123abc-123abc
      client 123abc-123abc-123abc-234xyz
      subscription 123abc-123abc-123abc-563abc
      secret mysecret
      public
      private
    }
```

And default to `public` or so.

darshanime:
```  
azure resource_group_foo:example.org resource_group_foo:example.private {
      tenant 123abc-123abc-123abc-123abc
      client 123abc-123abc-123abc-234xyz
      subscription 123abc-123abc-123abc-563abc
      secret mysecret
      public
      private
}
```

Hmm, so here, `resource_group_foo:example.org` is `public` and `esource_group_foo:example.private` is `private`? 
Is that determined by the order of zones and the public/private marker in block's body?

darshanime:
Bump.
Can this be reviewed and merged please? @miekg @yongtang

plugin/azure/README.md
miekg: the latter, have a property the signals the access level, access [private|both] make sense and defaulting to public (which isn't an option for access because it is the default)

plugin/azure/azure.go
```diff
@@ -13,53 +14,72 @@ import (
 	"github.com/coredns/coredns/plugin/pkg/upstream"
 	"github.com/coredns/coredns/request"
 
-	azuredns "github.com/Azure/azure-sdk-for-go/profiles/latest/dns/mgmt/dns"
+	publicAzureDNS "github.com/Azure/azure-sdk-for-go/profiles/latest/dns/mgmt/dns"
```
miekg: you can use caps in imports , they look also a bit ugly, what the default import path here, just 'dns'?

plugin/azure/azure.go
```diff
@@ -196,6 +225,87 @@ func updateZoneFromResourceSet(recordSet azuredns.RecordSetListResultPage, zName
 	return newZ
 }
 
+func updateZoneFromPrivateResourceSet(recordSet privateAzureDNS.RecordSetListResultPage, zName string) *file.Zone {
+	newZ := file.NewZone(zName, "")
+
+	for _, result := range *(recordSet.Response().Value) {
+		resultFqdn := *(result.RecordSetProperties.Fqdn)
+		resultTTL := uint32(*(result.RecordSetProperties.TTL))
+		if result.RecordSetProperties.ARecords != nil {
```
miekg, line 229: why is all this code repeated? Please factor it out a bit and reuse it for both private and public

plugin/azure/azure.go
darshanime: we have 2 `dns` now, `publicdns` and `privatedns`. They are different packages and both need importing hence `publicAzureDNS`. 
Should I change it to `publicdns` and `privatedns`?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add private D..." ]
>darshanime commented on this pull request.
>
>
>
>> @@ -13,53 +14,72 @@ import (
> 	"github.com/coredns/coredns/plugin/pkg/upstream"
> 	"github.com/coredns/coredns/request"
>
>-	azuredns "github.com/Azure/azure-sdk-for-go/profiles/latest/dns/mgmt/dns"
>+	publicAzureDNS "github.com/Azure/azure-sdk-for-go/profiles/latest/dns/mgmt/dns"
>
>we have 2 `dns` now, `publicdns` and `privatedns`. They are different packages and both need importing hence `publicAzureDNS`.
>Should I change it to `publicdns` and `privatedns`?

that sounds a lot better, so yes.

keiransteele:
I have compiled and tested this and it works great. I prefer the access bit to stay as it is now in the initial plugin line, it is easier to read rather than just having keywords which are disconnected from their context, especially if there are multiple domains.

miekg:
> especially if there are multiple domains.

yep, but I don't want to encode a lot things there, i.e. `tls` is *not* encoded there, but in a `tls` property inside the server block; it also limits the rest of coredns if we start allowing all kind of strings here; this is what the properties in the server block are for.

plugin/azure/azure.go
darshanime: these are distinct types in the auzre sdk codebase. 
Without generics, I am not sure how to refactor this, any ideas?

darshanime:
@miekg ptal now.

lukasmrtvy:
@miekg ping

plugin/azure/azure.go
miekg: I don't see it, it looks idential to the public ones  `*(recordSet.Response().Value) ` how it this different?

plugin/azure/azure.go
```diff
@@ -13,53 +13,69 @@ import (
 	"github.com/coredns/coredns/plugin/pkg/upstream"
 	"github.com/coredns/coredns/request"
 
-	azuredns "github.com/Azure/azure-sdk-for-go/profiles/latest/dns/mgmt/dns"
+	publicdns "github.com/Azure/azure-sdk-for-go/profiles/latest/dns/mgmt/dns"
+	privatedns "github.com/Azure/azure-sdk-for-go/profiles/latest/privatedns/mgmt/privatedns"
+
```
miekg: drop the newline here.

miekg (commented):
the pull request text message doesn't convey you're adding private zones. Could you update that?

miekg:
rebasing on master might be worthwhile as well. Generally this looks good, but we need to fix the code duplication currently here.

darshanime:
> we need to fix the code duplication currently here.

sure, ideas welcome!

lukasmrtvy:
@darshanime @miekg any news ?

darshanime:
> rebasing on master

done

> but we need to fix the code duplication currently here.

let me know if you have any ideas for this 👍

plugin/azure/azure.go
darshanime: the type is `privatedns.RecordSetListResultPage` and the other one is `publicdns.RecordSetListResultPage`

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add private D..." ]
>darshanime commented on this pull request.
>
>
>
>> @@ -196,6 +225,87 @@ func updateZoneFromResourceSet(recordSet azuredns.RecordSetListResultPage, zName
> 	return newZ
> }
>
>+func updateZoneFromPrivateResourceSet(recordSet privateAzureDNS.RecordSetListResultPage, zName string) *file.Zone {
>+	newZ := file.NewZone(zName, "")
>+
>+	for _, result := range *(recordSet.Response().Value) {
>+		resultFqdn := *(result.RecordSetProperties.Fqdn)
>+		resultTTL := uint32(*(result.RecordSetProperties.TTL))
>+		if result.RecordSetProperties.ARecords != nil {
>
>the type of `privatedns.RecordSetListResultPage` and the otherone is `publicdns.RecordSetListResultPage``

that is silly...


/lgtm

miekg:
ah the thing is failing travis. Please fix (don't whats wrong, because I didn't look)

lukasmrtvy:
@darshanime Can You look at it, please? ^

PR #3527, opened 2019-12-10
Fix HostPortOrFile to support IPv6 addresses with zone

bcrochet:
1. The HostPortOrFile tests don't have any IPv6 tests. This adds some.
2. The HostPortOrFile breaks if any of the addresses have IPv6 zone
defined. ParseIP does not handle %zone anymore.

This was discovered using the forward plugin, but perhaps would affect others. The HostPortOrFile function does not properly support an IPv6 address of the form address%zone. 

None

None
No

### DIFF
--- a/plugin/pkg/parse/host.go
+++ b/plugin/pkg/parse/host.go
@@ -4,11 +4,22 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"strings"
 
 	"github.com/coredns/coredns/plugin/pkg/transport"
 
 	"github.com/miekg/dns"
 )
+
+// Strips the zone, but preserves any port that comes after the zone
+func stripZone(host string) string {
+	if strings.Contains(host, "%") {
+		lastPercent := strings.LastIndex(host, "%")
+		newHost := host[:lastPercent]
+		return newHost
+	}
+	return host
+}
 
 // HostPortOrFile parses the strings in s, each string can either be a
 // address, [scheme://]address:port or a filename. The address part is checked
@@ -21,10 +32,11 @@ func HostPortOrFile(s ...string) ([]string, error) {
 		trans, host := Transport(h)
 
 		addr, _, err := net.SplitHostPort(host)
+
 		if err != nil {
 			// Parse didn't work, it is not a addr:port combo
-			if net.ParseIP(host) == nil {
-				// Not an IP address.
+			hostNoZone := stripZone(host)
+			if net.ParseIP(hostNoZone) == nil {
 				ss, err := tryFile(host)
 				if err == nil {
 					servers = append(servers, ss...)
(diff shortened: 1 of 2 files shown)


plugin/pkg/parse/host.go
```diff
@@ -24,11 +24,17 @@ func HostPortOrFile(s ...string) ([]string, error) {
 		if err != nil {
 			// Parse didn't work, it is not a addr:port combo
 			if net.ParseIP(host) == nil {
-				// Not an IP address.
-				ss, err := tryFile(host)
+				// Not an IPv4 address, for sure. Could be IPv6 with %zone?
+				addrs, err := net.LookupHost(host)
```
miekg: you can't use `net.LookupHost` here because you may not have a resolver at your disposal

miekg (changes requested):
doesn't net.ParseIP do this correct but itself?

In general, I think this is a good change.

plugin/pkg/parse/host.go
bcrochet: I was afraid of that. I wasn't totally sure that this was the correct approach, but it seemed like a good first step.

bcrochet (commented):
> doesn't net.ParseIP do this correct but itself?
> 
> In general, I think this is a good change.

That's why I'm trying to figure this out. ParseIP itself is not sufficient. If you pass ParseIP an IP with a %zone, it returns nil. So, while that isn't the worst thing, it just means that there needs to be more checks. Feel free to try running the tests with the additional cases that I've added, without the net.LookupHost. The ones with the %zone will fail.

bcrochet:
Filed bug #3528

miekg:
I won't have to time to work on this myself. But iirc the go std library
should be able to parse and see the zone part. Tricky bit might be
separating from a weirdly named file

bcrochet:
net.ParseIP definitely does not handle an address with %zone in it. I added the test cases with IPv6 addresses first, and ran them before I added any code. The ones with %zone failed.

miekg:
I think you need to net.ParseIP trimming off the `%scope` suffix. If that's an IP `%scope` is a scope. Note sure if `scope` can be syntax checked.

(also note I said zone, but it is scope before)

bcrochet:
If I'm understanding correctly, you're saying to trim off %scope before passing to ParseIP?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fix HostPortO..." ]
>If I'm understanding correctly, you're saying to trim off %scope before passing to ParseIP?

yep. Makes sense, no?

ironcladlou:
@miekg looks like the zone stripping feedback was addressed, PTAL? Thanks!

miekg (approved):

Issue #2564, opened 2019-02-16
plugin/hosts fallthrough results in NXDOMAIN


juselius:
I'm trying to use the hosts plugin to augment and override host names in our organization's DNS domain. Using the Corefile below results in both resolving the name **and** in NXDOMAIN for every host defined in the hosts section. Everything works, but it's a bit confusing getting NXDOMAIN after a successful lookup.

I'm using coredns version 1.2.5.

```
.:10053 {
             errors
             health :10054
             kubernetes cluster.local in-addr.arpa ip6.arpa {
                  pods insecure
                  upstream
                  fallthrough in-addr.arpa ip6.arpa
              }
              prometheus :10055
              proxy . /etc/resolv.conf
              hosts  {
                    10.253.18.100 myservice.mydomain.com
                    fallthrough
              }
              cache 30
              loop
              reload
              loadbalance
}
```

miekg:
Can you show some responses, preferable using `dig`

juselius:
@miekg sorry for the late reply.  Here is the output from nslookup and dig on a container:

```
root@busynix-df88cb5d6-lhc9x /# nslookup myservice.mydomain.com
Server:		10.0.0.254
Address:	10.0.0.254#53

Name:	myservice.mydomain.com
Address: 10.253.18.100
** server can't find myservice.mydomain.com: NXDOMAIN
```

```
root@busynix-df88cb5d6-lhc9x /# dig myservice.mydomain.com
; <<>> DiG 9.12.3-P1 <<>> myservice.mydomain.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 11012
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;myservice.mydomain.com.		IN	A

;; ANSWER SECTION:
myservice.mydomain.com.	30	IN	A	10.253.18.100

;; Query time: 0 msec
;; SERVER: 10.0.0.254#53(10.0.0.254)
;; WHEN: Mon Feb 18 17:32:55 UTC 2019
;; MSG SIZE  rcvd: 70
```

As you can see, it both finds the host, and then it doesn't.

miekg:
host will be never a full fledged plugin, just like /etc/hosts isn't a complete zone.

mrdima:
[...]

PR #3536, opened 2019-12-12
plugin/hosts: Modifies NODATA handling

ykhr53:
Bug fixing for the issue https://github.com/coredns/coredns/issues/2564.
If hosts file has only A record and this plugin receives a query for AAAA, NODATA should be returned.
Now, this plugin fallthrough the query and a client going to get NXDOMAIN.

https://github.com/coredns/coredns/issues/2564

I don't think so.

No.

### DIFF
--- a/plugin/hosts/hosts.go
+++ b/plugin/hosts/hosts.go
@@ -52,15 +52,15 @@ func (h Hosts) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 		answers = aaaa(qname, h.options.ttl, ips)
 	}
 
-	if len(answers) == 0 {
+	// This condition hooks only NXDOMAIN.
+	if len(answers) == 0 && !h.otherRecordsExist(qname) {
 		if h.Fall.Through(qname) {
 			return plugin.NextOrFailure(h.Name(), h.Next, ctx, w, r)
 		}
+
 		// We want to send an NXDOMAIN, but because of /etc/hosts' setup we don't have a SOA, so we make it REFUSED
 		// to at least give an answer back to signals we're having problems resolving this.
-		if !h.otherRecordsExist(qname) {
-			return dns.RcodeServerFailure, nil
-		}
+		return dns.RcodeServerFailure, nil
 	}
 
 	m := new(dns.Msg)
--- a/plugin/hosts/hosts.go
+++ b/plugin/hosts/hosts.go
@@ -52,13 +52,13 @@ func (h Hosts) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 		answers = aaaa(qname, h.options.ttl, ips)
 	}
 
-	// This condition hooks only NXDOMAIN.
+	// Only on NXDOMAIN we will fallthrough.
 	if len(answers) == 0 && !h.otherRecordsExist(qname) {
 		if h.Fall.Through(qname) {
 			return plugin.NextOrFailure(h.Name(), h.Next, ctx, w, r)
 		}
 
-		// We want to send an NXDOMAIN, but because of /etc/hosts' setup we don't have a SOA, so we make it REFUSED
+		// We want to send an NXDOMAIN, but because of /etc/hosts' setup we don't have a SOA, so we make it SERVFAIL
 		// to at least give an answer back to signals we're having problems resolving this.
 		return dns.RcodeServerFailure, nil
 	}
--- a/plugin/hosts/hosts_test.go
+++ b/plugin/hosts/hosts_test.go
@@ -6,32 +6,46 @@ import (
 	"testing"
 
 	"github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/coredns/coredns/plugin/pkg/fall"
 	"github.com/coredns/coredns/plugin/test"
 
 	"github.com/miekg/dns"
 )
 
 func TestLookupA(t *testing.T) {
-	h := Hosts{
-		Next: test.ErrorHandler(),
-		Hostsfile: &Hostsfile{
-			Origins: []string{"."},
-			hmap:    newMap(),
-			inline:  newMap(),
-			options: newOptions(),
-		},
-	}
-	h.hmap = h.parse(strings.NewReader(hostsExample))
-
-	ctx := context.TODO()
-
 	for _, tc := range hostsTestCases {
 		m := tc.Msg()
 
+		var tcFall fall.F
+		isFall := tc.Qname == "fallthrough-example.org."
+		if isFall {
+			tcFall = fall.Root
+		} else {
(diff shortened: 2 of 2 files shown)


plugin/hosts/hosts.go
```diff
@@ -52,15 +52,15 @@ func (h Hosts) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
...
 		if h.Fall.Through(qname) {
 			return plugin.NextOrFailure(h.Name(), h.Next, ctx, w, r)
 		}
+
 		// We want to send an NXDOMAIN, but because of /etc/hosts' setup we don't have a SOA, so we make it REFUSED
 		// to at least give an answer back to signals we're having problems resolving this.
-		if !h.otherRecordsExist(qname) {
-			return dns.RcodeServerFailure, nil
-		}
+		return dns.RcodeServerFailure, nil
```
miekg, line 63: I just notice this comment doesn't reflect what we're sending, REFUSED->SERVFAIL

plugin/hosts/hosts.go
```diff
@@ -52,15 +52,15 @@ func (h Hosts) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 		answers = aaaa(qname, h.options.ttl, ips)
 	}
 
-	if len(answers) == 0 {
+	// This condition hooks only NXDOMAIN.
```
miekg: comment looks off, maybe something like:
Only on NXDOMAIN we will fallthrough

miekg:
needs a test as well

plugin/hosts/hosts.go
ykhr53: I haven't touched it, but you are right. I'm gonna change it in this PR.

ykhr53:
Thank you for the review!
The original code has some tests which cover this change, that's why I didn't add tests.
Do you think we need to add a new test for more specific case?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/hosts:..." ]
>Do you think we need to add a new test for more specific case?

yes, you're changing behavior, no? That should have a test that fails now, but is fixed
with this PR

ykhr53:
Ok, thanks. I modified two tests to ensure to get NODATA instead of NXDOMAIN.

plugin/hosts/hosts.go
```diff
@@ -52,15 +52,15 @@ func (h Hosts) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
...
 		if h.Fall.Through(qname) {
 			return plugin.NextOrFailure(h.Name(), h.Next, ctx, w, r)
 		}
+
 		// We want to send an NXDOMAIN, but because of /etc/hosts' setup we don't have a SOA, so we make it REFUSED
 		// to at least give an answer back to signals we're having problems resolving this.
-		if !h.otherRecordsExist(qname) {
-			return dns.RcodeServerFailure, nil
-		}
+		return dns.RcodeRefused, nil
```
miekg: I meant changing the comment, not the return code itself

plugin/hosts/hosts_test.go
```diff
@@ -82,11 +82,11 @@ var hostsTestCases = []test.Case{
 	},
 	{
 		Qname: "example.org.", Qtype: dns.TypeAAAA,
-		Answer: []dns.RR{},
+		Answer: []dns.RR{}, Rcode: dns.RcodeSuccess,
```
miekg: I would like to see a new test that is failing with the current code and working with the new

plugin/hosts/hosts.go
miekg: also making this change without failing tests means this is lacking coverage. As you are touching this code; please also add a test for this :)

ykhr53: OK, I'm going to add some new tests which will cover this line and my new code!

ykhr53:
I made a bit wide change to the **hosts_test.go** because we need to take care the cases whether fallthrough is enabled or not.
The last test will be fail in the current code, and work in the new code.
I hope it's good and enough :)

ykhr53:
@miekg @pmoroney @johnbelamaric 

Hi, could you review this code?

plugin/hosts/hosts_test.go
```diff
@@ -6,87 +6,158 @@ import (
...
-		Next: test.ErrorHandler(),
-		Hostsfile: &Hostsfile{
-			Origins: []string{"."},
-			hmap:    newMap(),
-			inline:  newMap(),
-			options: newOptions(),
-		},
-	}
-	h.hmap = h.parse(strings.NewReader(hostsExample))
+	for i, tc := range hostsTestCases {
```
miekg: adding the test is good, but why did you amend an existing test? Better to just have a small contained test with only 1 use case: one that fails now and is fixed with this PR

ykhr53: the Case structure doesn't have any optional fields, so I need to extend it to use `Fallthrough` flag in `func TestLookupA`, I could not add `Fallthrough` only for the last test case.
So... should I add a new function like `func TestFallthrough` which checks only one case?
I don't have an idea which one is better, I'm going to follow your advice.
Thanks!

ykhr53:
I updated the test code to have a small contained test for failing only the current code.
We need to check RCODE to fix and test this bug, but as you said we don't want to change the existing cases.
So in this code, just check RCODE only the case fallthrough is enabled.
(I would like to add test index like the other plugins, but in this PR, I'm focusing on fixing this bug.)

#### Current Code
```
--- FAIL: TestLookupA (0.00s)
    /Users/yukihira/develop/coredns/plugin/hosts/hosts_test.go:48: Expected rcode is 0, but got 3
FAIL
FAIL	github.com/coredns/coredns/plugin/hosts	0.018s
Error: Tests failed.
```

#### New Code by this PR
```
ok  	github.com/coredns/coredns/plugin/hosts	(cached)
Success: Tests passed.
```

ykhr53:
The CircleCI test was failed because of the following reason.
```
    --- FAIL: TestKubernetesA/dns-version.cluster.local._TXT (0.44s)
        address_test.go:173: RR 0 should have a Txt of "1.1.0", but has "1.0.1"
        address_test.go:176: coredns log: .:53
```
I found some same error cases, but to be honest I am not sure this error is related to *hosts* plugin changes.
Much appreciated if anyone takes a look at it. Thanks.
https://circleci.com/gh/coredns/coredns/1411
https://circleci.com/gh/coredns/coredns/1409

chrisohaver:
> The CircleCI test was failed because of the following reason. ...

Updating your master branch and rebasing this branch against master should fix it.

ykhr53:
@chrisohaver Thank you for the reply and giving me a solution! I'm gonna do that.

miekg (approved):

Issue #2695, opened 2019-03-14
Dedup some code between grpc and proxy plugin


miekg:
Now that there are two *forwarding* plugins (*forward* and *grpc*), there is some code duplicated between the two; it worth checking if some of it can be factored into a couple of common packages under /plugin/pkg

inigohu:
Both have the same _policy_ implemented. Maybe we can move that to pkg.

miekg:
yeah, could be although that not a lot of code, nor is it complex. I.e. this is not a call to avoid duplication at any cost. Maybe there is not tangible to extract.

palash25:
Hi, @miekg I would like to take this up. Besides `policy.go` I see the following duplicate methods between these two plugins

```go
func (f *Forward) Len() int { return len(f.proxies) }

func (f *Forward) isAllowedDomain(name string) bool

func (f *Forward) match(state request.Request) bool

func (f *Forward) List() []*Proxy { return f.p.List(f.proxies) }
```

could make small changes to them; to not be defined on a particular struct and pass an interface as an argument instead and move these to the pkg dir. 

For example

```go
func match(plugin interface{}, state request.Request) bool
```


not sure if this is a desirable change or not let me know if you would like something like this to be included in the PR too.

miekg:
I think match and isAllowedDomain (they are used together anyway) is good to be abstracted into a shared package.

The signature for match would be better as: `match(state request.Request, name ...string) bool`

What would be a good name for this package? We could put `isAllowedDomain` in pkg/dnsutil ? Or something?

palash25:
> What would be a good name for this package? We could put isAllowedDomain in pkg/dnsutil ? Or something?

For now I am putting them in a file called match.go under pkg/dnsutil

Other alternatives could be a separate package called `pkg/utils` or `pkg/forward` (since these are utility functions used only by the forwarding plugins)

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Dedup some co..." ]
>> What would be a good name for this package? We could put isAllowedDomain in pkg/dnsutil ? Or something?
>
>For now I am putting them in a file called match.go under pkg/dnsutil
>
>Other alternatives could be a separate package called `pkg/utils` or `pkg/forward` (since these are utility functions used only by the forwarding plugins)

dnsutil sgtm

sharmaansh21:
I don't think anyone is looking into it, I can work in it.

palash25:
[...]

PR #3537, opened 2019-12-13
Dedup policy implement between grpc and proxy plugin

zouyee:
Signed-off-by: zouyee <zounengren@cmss.chinamobile.com>

 `grpc and proxy` plugin  have the same policy implemented  which is worth to  be factored into a couple of common packages .

a part of https://github.com/coredns/coredns/issues/2695

NONE

NONE


miekg (changes requested):
thanks.

dnsutil is probably the wrong package, I suggest pkg/policy and put everything in there

zouyee:
@miekg PTAL

plugin/pkg/policy/policy.go
```diff
@@ -0,0 +1,76 @@
...
+	List(policy ...interface{}) []interface{}
+	String() string
+}
+
+// Random is a policy that implements random upstream selection.
+type Random struct{}
+
+var _ Policy = &Random{}
+
+// String return the name of policy Random
```
miekg:
```suggestion
// String returns the name of policy Random.
```

plugin/pkg/policy/policy.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// RoundRobin is a policy that selects hosts based on round robin ordering.
+type RoundRobin struct {
+	robin uint32
+}
+
+var _ Policy = &RoundRobin{}
+
+// String return the name of policy RoundRobin
```
miekg:
```suggestion
// String returns the name of policy RoundRobin.
```

plugin/pkg/policy/policy.go
```diff
@@ -0,0 +1,76 @@
...
+
+	return robin
+}
+
+// Sequential is a policy that selects hosts based on sequential ordering.
+type Sequential struct{}
+
+var _ Policy = &Sequential{}
+
+// String return the name of policy Sequential
```
miekg:
```suggestion
// String returns the name of policy Sequential.
```

miekg (changes requested):

miekg:
yeah, 50 lines reduction in code :)

PR #3538, opened 2019-12-13
add consul plugin

poppyred:


dilyevsky (commented):
Browsing briefly this change will likely need a lot of work. Previously we've discussed that new substantial plugins will need support from at least one of maintainers. Has anything like this been discussed anywhere?

poppyred:
It doesn't feel good

poppyred:
move to [here](https://github.com/coredns/coredns/pull/3545)

poppyred:
>Browsing briefly this change will likely need a lot of work. Previously we've discussed that new substantial plugins will need support from at least one of maintainers. Has anything like this been discussed anywhere?
@dilyevsky 

just a few modifications to the etcd base, adding wildcard query logic, I think maintenance costs are a little.
it is that I need and few issue, but I think that add a consul back-end very good

PR #3545, opened 2019-12-16
add consul plugin!

poppyred:
Signed-off-by: 高岩A <yan.gao.a@17zuoye.com>


miekg:
please open an issue first to discuss the inclusion of this backend.

miekg:
see discussion, yes more backend support is useful, but imo we've reach the limit and something more scalable must be sought.

PR #3548, opened 2019-12-17
use map to accelerate ignore name lookup

missdeer:

I have a very large except list (> 73660 items) for `forward` plugin, linear lookup is not efficient enough. I use std map to accelerate this operation.

N/A

N/A

N/A


miekg:
`plugin.Name(ignore).Matches(name)` does more than that, also a list that big isn't really support so no wonder it's slow.

A better change would be to just cap the maximum number of elements you can have to 15 (just like the number of upstreams)

PR #3551, opened 2019-12-18
Dedup grpc and proxy plugin's internal func

### LINKED ISSUE #2695: Dedup some code between grpc and proxy plugin

zouyee:

grpc and proxy plugin have the same internal func implemented which is worth to be factored into a couple of common packages .

fix https://github.com/coredns/coredns/issues/2695

NONE

NONE

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -12,8 +12,8 @@ import (
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/debug"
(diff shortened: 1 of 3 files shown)


miekg:
I don't think this should move to normalize.go and also have a lot of nits on this code.

I would just leave this as-is and duplicated.

Issue #3556, opened 2019-12-19
TXT Lookup does not resolve via CNAME


ghost:
**What would you like to be added**:

Allow CoreDNS to lookup TXT records via CNAME.

**Why is this needed**:

From my observations, DNS servers will normally facilitate an answer to a TXT query via CNAME.  Currently, to the best of my knowledge, this is not supported by CoreDNS.

An example of this request running against a local DNS Server is below.  I've also proven this functionality with several other global DNS servers as well.

```
$ dig @localhos -t TXT cname.example.local

; <<>> DiG 9.10.6 <<>> -t TXT cname.example.local
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 12473
;; flags: qr rd ra; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4000
;; QUESTION SECTION:
;cname1.example.local.		IN	TXT

;; ANSWER SECTION:
cname1.example.local.	60	IN	CNAME	cname2.example.local.
cname2.example.local.	60	IN	CNAME	server.example.local.
server.example.local	56	IN	TXT     "test"
```

miekg:
has been merged.

PR #3557, opened 2019-12-20
Resolve TXT records via CNAME

ghost:

This PR allows CoreDNS to facilitate the lookup of TXT records via CNAME by applying the same pattern that is used to lookup A, AAAA records.

#3556 

None that I'm aware of.

A new parameter has been introduced to the TXT method in `backend_lookup` which is not backward compatible for any plugins that use that API.  However, I have adapted both the etcd and kubernetes plugins.  The only perceivable impact would be to any third party plugins that do not exist in the CoreDNS repository.

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -327,13 +327,67 @@ func CNAME(ctx context.Context, b ServiceBackend, zone string, state request.Req
 // TXT returns TXT records from Backend or an error.
 func TXT(ctx context.Context, b ServiceBackend, zone string, state request.Request, previousRecords []dns.RR, opt Options) (records []dns.RR, err error) {
 
+	services, err := b.Services(ctx, state, true, opt)
 	if err != nil {
 		return nil, err
 	}
 
+	dup := make(map[string]struct{})
+
 	for _, serv := range services {
-		records = append(records, serv.NewTXT(state.QName()))
+
+		what, _ := serv.HostType()
+
+		switch what {
+		case dns.TypeCNAME:
+			if Name(state.Name()).Matches(dns.Fqdn(serv.Host)) {
+				// x CNAME x is a direct loop, don't add those
+				continue
+			}
+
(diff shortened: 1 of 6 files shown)


chrisohaver:
Thanks, LGTM.

**Other reviewers**: Is this the tipping point for consolidating this duplicated-yet-one-more-time CNAME code in backend?

chrisohaver (approved):

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Resolve TXT r..." ]
>Thanks, LGTM.
>
>**Other reviewers**: Is this the tipping point for consolidating this duplicated-yet-one-more-time CNAME code in backend?

yes, this contraption needs works. Do we believe backend.go serve enough purpose to keep
it? Etc. etc.

ghost:
Thanks for the review.

I would have refactored the multiple CNAME implementations into one, but I think that would have been a more drastic change and difficult to review.

I can have a look at doing it if necessary.  Would you guys want a separate PR for it?

chrisohaver:
> yes, this contraption needs works. Do we believe backend.go serve enough purpose to keep
it? Etc. etc.

I personally would like to see kubernetes refactored to not use the backend.  But this would be a very large refactoring task, affecting related plugins, and some open PRs. This is something I can tackle if we want to go that route.  If we do want to eventually retire the backend code, then duplicating the code here for now is probably fine IMO.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Resolve TXT r..." ]
>> yes, this contraption needs works. Do we believe backend.go serve enough purpose to keep
>it? Etc. etc.
>
>I personally would like to see kubernetes refactored to not use the backend.  But this would be a very large refactoring task, affecting related plugins, and some open PRs. This is something I can tackle if we want to go that route.  If we do want to eventually retire the backend code, then duplicating the code here for now is probably fine IMO.

I think we need to do something indeed. Note it would be good to have the 'transfer' thing
fixed fully first.

PR #3571, opened 2019-12-29
doc: fix generated manual pages

miekg:
Went over all generated manual pages and fixed some markdown issues,
mostly escaping "_" to avoid underlining entire paragraphs.

Some textual fixes in route53 and other cloud DNS plugins.

Regenerated the markdown with mmark.


SuperQ (approved):
LGTM

PR #3573, opened 2019-12-30
plugin/cache: update comment to conform to the implementation

xiez:

Update comment accordingly to the code fix in https://github.com/coredns/coredns/commit/159a8fb68c44b238ccf603e9d7bd822519dcfb41 :wink: 

no

no

no

### DIFF
--- a/plugin/cache/item.go
+++ b/plugin/cache/item.go
@@ -49,7 +49,12 @@ func newItem(m *dns.Msg, now time.Time, d time.Duration) *item {
 }
 
 // toMsg turns i into a message, it tailors the reply to m.
-// The Authoritative bit is always set to 0, because the answer is from the cache.
+// The Authoritative bit should be set to 0, but some client stub resolver implementations, most notably,
+// on some legacy systems(e.g. ubuntu 14.04 with glib version 2.20), low-level glibc function `getaddrinfo`
+// useb by Python/Ruby/etc.. will discard answers that do not have this bit set.
+// So we're forced to always set this to 1; regardless if the answer came from the cache or not.
+// On newer systems(e.g. ubuntu 16.04 with glib version 2.23), this issue is resolved.
+// So we may set this bit back to 0 in the future ?
 func (i *item) toMsg(m *dns.Msg, now time.Time) *dns.Msg {
 	m1 := new(dns.Msg)
 	m1.SetReply(m)


plugin/cache/item.go
```diff
@@ -49,7 +49,7 @@ func newItem(m *dns.Msg, now time.Time, d time.Duration) *item {
 }
 
 // toMsg turns i into a message, it tailors the reply to m.
-// The Authoritative bit is always set to 0, because the answer is from the cache.
```
miekg, line 52: good catch. Please make this comment longer. 

The Authoritative bit should be set to 0, but some client stub resolver implementations, most notably, the one from Python will discard answers that do not have this bit set. So we're forced to always set this to 1; regardless if the answer came from the cache or not.

xiez: updated. 😄 

Just a little FYI, Python use low-level glibc function `getaddrinfo`  (https://github.com/python/cpython/blob/3.6/Modules/socketmodule.c#L6061) to perform address resolution. On some legacy systems(e.g. ubuntu 14.04 with glib version 2.20), this function will discard answers that do not have Authoritative bit set, which means not only Python, other languages like ruby/C also suffer the same issue;  On newer systems(e.g. ubuntu 16.04 with glib version 2.23), this issue is resolved.

miekg: Can you add that too; I don't mind that we have a long comment here capturing all these bits.

xiez: ok, how about this:

```
// The Authoritative bit should be set to 0, but some client stub resolver implementations, most notably,
// on some legacy systems(e.g. ubuntu 14.04 with glib version 2.20), low-level glibc function `getaddrinfo`
// useb by Python/Ruby/etc.. will discard answers that do not have this bit set.
// So we're forced to always set this to 1; regardless if the answer came from the cache or not.
// On newer systems(e.g. ubuntu 16.04 with glib version 2.23), this issue is resolved.
// So we may set this bit back to 0 in the future ?
```

miekg:
>ok, how about this:
>
>```
>// The Authoritative bit should be set to 0, but some client stub resolver implementations, most notably,
>// on some legacy systems(e.g. ubuntu 14.04 with glib version 2.20), low-level glibc function `getaddrinfo`
>// useb by Python/Ruby/etc.. will discard answers that do not have this bit set.
>// So we're forced to always set this to 1; regardless if the answer came from the cache or not.
>// On newer systems(e.g. ubuntu 16.04 with glib version 2.23), this issue is resolved.
>// So we may set this bit back to 0 in the future ?
>```

LGTM!

PR #3575, opened 2019-12-31
correct plugin/kubernetes markdown format

zouyee:

correct plugin/kubernetes markdown format

NONE

NONE

NONE

### DIFF
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -20,7 +20,7 @@ This plugin can only be used once per Server Block.
 
 ## Syntax
 
-~~~
+~~~ txt
 kubernetes [ZONES...]
 ~~~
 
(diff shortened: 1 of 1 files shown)


miekg:
Is there anything really broken in the current markdown? I'm not a fan of
just pushing change for the sake of it.

zouyee:
I just found those warnings for the markdown format by the markdown plugin for vscode.
https://github.com/DavidAnson/markdownlint/blob/v0.18.0/doc/Rules.md#md005

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] correct plugi..." ]
>I just found those warnings for the markdown format by the markdown plugin for vscode.
>https://github.com/DavidAnson/markdownlint/blob/v0.18.0/doc/Rules.md#md005

hmmm, if we think something is needed we should format this automatically. We already use
mmark to generate the manual pages, mmark can also format markdown documents
(https://mmark.miek.nl/post/what/)

That renderer is marked experimental, so little bugs may still be there (although I'm
pretty happy with it)

/Miek

--
Miek Gieben

Issue #3541, opened 2019-12-15
Autoscaling CoreDNS Nodes on Kubernetes Clusters Through Exposed Metrics


yongtang:
**Please Note:** This is a tracking issue for [Community Bridge Program](https://github.com/cncf/soc/tree/master/communitybridge). Anyone interested in this implementation should check [link](https://github.com/cncf/soc/tree/master/communitybridge) there.

**Please Note:** This is reserved for Community Bridge Program, **we don't accept PRs outside of Community Bridge Program.**

**Autoscaling CoreDNS Nodes on Kubernetes Clusters Through Exposed Metrics**

- Description: [CoreDNS](https://github.com/coredns/coredns) is Kubernetes' default DNS server for service discovery. As service discovery with DNS is the (often overlooked) critical piece for clusters, it is important to maintain enough capacity for serving the cluster-wide service discovery. CoreDNS already have a `metrics` plugin with metrics information exposed. In this project, the student will work on the following tasks:
  - Write metrics scraping method to extract the metrics from exposed CoreDNS metrics plugin.
  - Develop a method in transforming collected metrics into desired capacity for CoreDNS nodes.
  - Implement a routine to interact with Kubernetes API to autoscaling CoreDNS node with designed capacity.
- Bonus Task: Use machine learning algorithm for transforming metrics to above mentioned designed capacity. We will provide a start algorithm to help student getting started with machine learning, if student is willing to take this bonus task.
- Recommended Skills: Golang and DNS for this project, basic machine learning for bonus task.
- Mentor(s): Yong Tang @yongtang

Implementation: This project will be implemented as a standalone program. It could be placed under coredns org if finished successfully.

rajibmitra:
Interested to complete all the mentioned task as part of community bridge program @yongtang  , could you please assign this to me.

yongtang:
@rajibmitra The application is done through the community bridge program. Please see https://github.com/cncf/soc/tree/master/communitybridge#how-to-apply for details on how to apply.

knrt10:
This looks interesting. @yongtang till now I have run coredns locally and gone through basic plugin tutorial. I am interested in this, can you please guide me through any of the next steps?

yongtang:
@knrt10 you can look into https://github.com/cncf/soc/tree/master/communitybridge#how-to-apply for details on how to apply.

Also, we are still waiting for the update from CNCF on how the slot is allocated (not confirmed yet).

knrt10:
Sure, thanks

[...]

PR #3576, opened 2020-01-01
[WIP] Added initial scraping method

knrt10:
Signed-off-by: knrt10 <tripathi.kautilya@gmail.com>

Related to #3541

- Initial work for scraping. 
- Added 2 files in `metrics` folder, `scrape.go` and `scrape_test.go`.

Currently, it is returning JSON with the URL for metrics.

cc @yongtang. Also can you please tell in what format do we want the metrics to be?


miekg:
I'm assuming this is for #3541 ? Can you reference that issue in your PR message? Can you *also* explain *why* this is needed and must be merged? Is your plan for #3541 describe somewhere (public) ?

SuperQ:
This doesn't make any sense to me. We already have a standard Prometheus format metrics output. There's no need to implement JSON output. JSON output is inefficient and obsolete. It's not been supported by Prometheus for many years now.

yongtang:
@knrt10 You don't need to scrape the metrics into json. What you need is to extract the metric and transform the "vector of metrics" into a scaling factor, so that this scaling factor could be feed into Kubernetes. The *transform* of the "vector of metrics" could initially be done in Golang, by some simple method (e.g., linear?). Later you can try to find a way to wire up external *transform* program into the pipeline. For example, maybe with keras?

Normally in machine learning, python is the first-class citizen. However, it is still possible to wire up into the Golang. Let me know if you want to have a pointer on how to achieve it.

miekg:
nacking this PR - I don't foresee a need to add special scraping code to coredns [for this usecase]

PR #3580, opened 2020-01-05
optimize the metrics test

zouyee:
Signed-off-by: zouyee <zounyee1989@gmail.com>

To optimize the monitoring test, we should not simply confirm the existence of metrics, but compare the results of metrics deeply.

NONE

NONE

### DIFF
--- a/plugin/test/scrape.go
+++ b/plugin/test/scrape.go
@@ -25,14 +25,17 @@ package test
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"mime"
 	"net/http"
+	"reflect"
 	"strconv"
+	"strings"
 
 	"github.com/matttproud/golang_protobuf_extensions/pbutil"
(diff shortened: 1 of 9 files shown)


plugin/test/scrape.go
```diff
@@ -25,14 +25,18 @@ package test
...
 	"strconv"
+	"strings"
 
 	"github.com/matttproud/golang_protobuf_extensions/pbutil"
-	"github.com/prometheus/common/expfmt"
-
 	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/common/model"
+	"k8s.io/apimachinery/pkg/util/sets"
```
miekg: please, let's not "pollute" non-kubernetes code with kubernetes code. You probably also don't actually need sets, but a simple Go `map` (didn't look at the code yet)`

miekg:
this looks stale, closing for now

PR #3589, opened 2020-01-13
add mips64le to released ARCH

ydcool:
Signed-off-by: Dominic Yin <yindongchao@inspur.com>

Add mips64le to released ARCH. Since the golang has a completely supporting for cross compile for mips64le, the coredns which written in pure go should have a release support for mips64le without any side effects.

On the other hand, when we build the test images for kubernetes, the `jessie-dnsutils` image will download the coredns released tar file directly from this release page, but currently the tar for mips64le are not avaliable.
https://github.com/kubernetes/kubernetes/blob/master/test/images/jessie-dnsutils/Dockerfile#L30

No
No
No

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -88,6 +88,8 @@ build:
 	mkdir -p build/windows/amd64 && $(MAKE) coredns BINARY=build/windows/amd64/$(NAME).exe SYSTEM="GOOS=windows GOARCH=amd64" CHECKS="" BUILDOPTS=""
 	@echo Building: linux/mips - $(VERSION)
 	mkdir -p build/linux/mips  && $(MAKE) coredns BINARY=build/linux/mips/$(NAME) SYSTEM="GOOS=linux GOARCH=mips" CHECKS="" BUILDOPTS=""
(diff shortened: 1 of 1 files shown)


miekg (approved):

ydcool:
@miekg thank you, but one more question, is it possiable to upload the mips64le tars of released version (1.5.0 and later ) into the release page? then how?

miekg:
Not easily I think. That would involve manually fiddling with those
releases and if we screw it up it needs to be made right again.

In rather do a smaller release to include this arch

ydcool:
Yes you're right, but I'm worried about that it's a little bothering if we release a new version only for this. So is there any exact time when will the next version (1.6.7) released?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] add mips64le ..." ]
>Yes you're right, but I'm worried about that it's a little bothering if we release a new version only for this. So is there any exact time when will the next version (1.6.7) released?

there is no schedule and our automated releases (if it works ;-) ) are very boring as well. I have no problems with releasing
just to get some stuff "out there"

Issue #3587, opened 2020-01-11
plugin/kubernetes: coredns container can not resolve new created k8s service after running 2 months


njuicsgz:
**What happened**:
I have 5 coredns pod running about 2 months, then I found 2 of them can not lookup new services created by k8s.
```
~/allen# kubectl -n core-dns-wlcb-prod get svc,po -o wide
NAME           TYPE        CLUSTER-IP   EXTERNAL-IP     PORT(S)                   AGE       SELECTOR
svc/core-dns   ClusterIP   172.29.8.8   192.168.2.184   49153/TCP,53/UDP,53/TCP   88d       name=core-dns

NAME                                 READY     STATUS    RESTARTS   AGE       IP              NODE
po/core-dns-bb32dbc62e244ff-6zzb5    1/1       Running   1          67d       172.16.84.14    192.168.2.41
po/core-dns-bb32dbc62e244ff-89rdx    1/1       Running   1          67d       172.16.117.10   192.168.2.203
po/core-dns-bb32dbc62e244ff-pbfqm    1/1       Running   1          67d       172.16.84.15    192.168.2.41
po/core-dns-bb32dbc62e244ff-rfqbt    1/1       Running   1          67d       172.16.202.12   192.168.2.79
po/core-dns-bb32dbc62e244ff-vx5dx    1/1       Running   1          67d       172.16.83.20    192.168.2.184
```

the problem pod (172.16.83.20) 
```
root@dns-check-e5b984de89ac472-dd98j:/# dig @172.16.83.20 core.titan-dev-open.svc.cluster.local

; <<>> DiG 9.10.3-P4-Ubuntu <<>> @172.16.83.20 core.titan-dev-open.svc.cluster.local
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 23894
;; flags: qr aa rd; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;core.titan-dev-open.svc.cluster.local. IN A

;; AUTHORITY SECTION:
cluster.local.          5       IN      SOA     ns.dns.cluster.local. hostmaster.cluster.local. 1578707575 7200 1800 86400 5

;; Query time: 0 msec
[...]

PR #3590, opened 2020-01-13
Pass the simple communication test result as a pre-condition for ready check

zouyee:
 Although the connection between coredns and apiserver is broken, but the HasSynced () method  of reflector will not report the error.

If this is "connection refused" error, it means that most likely apiserver is not responsive.
It doesn't make sense to re-list all objects because most likely we will be able to restart watch where we ended.
If that's the case wait and resend watch request.

https://github.com/coredns/coredns/issues/3587
none
none

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -312,6 +312,19 @@ func (dns *dnsControl) Run() {
 
 // HasSynced calls on all controllers.
 func (dns *dnsControl) HasSynced() bool {
+	// Pass the simple communication test result as a pre-condition for ready check.
+	// If this is "connection refused" error, it means that most likely apiserver is not responsive.
+	// It doesn't make sense to re-list all objects because most likely we will be able to restart
+	// watch where we ended.
+	// If that's the case wait and resend watch request.
+	v, err := dns.client.Discovery().ServerVersion()
+	if err != nil {
+		log.Warningf("While trying to communicate with apiserver found err: %v", err)
+		return false
+	}
+	log.Infof("Running with Kubernetes cluster version: v%s.%s. git version: %s. git tree state: %s. commit: %s. platform: %s",
+		v.Major, v.Minor, v.GitVersion, v.GitTreeState, v.GitCommit, v.Platform)
+	log.Infof("Communication with server successful")
 	a := dns.svcController.HasSynced()
 	b := true
 	if dns.epController != nil {
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -319,12 +319,11 @@ func (dns *dnsControl) HasSynced() bool {
 	// If that's the case wait and resend watch request.
 	v, err := dns.client.Discovery().ServerVersion()
 	if err != nil {
-		log.Warningf("While trying to communicate with apiserver found err: %v", err)
+		log.Warningf("While trying to communicate with apiserver found error: %v", err)
 		return false
 	}
 	log.Infof("Running with Kubernetes cluster version: v%s.%s. git version: %s. git tree state: %s. commit: %s. platform: %s",
 		v.Major, v.Minor, v.GitVersion, v.GitTreeState, v.GitCommit, v.Platform)
-	log.Infof("Communication with server successful")
 	a := dns.svcController.HasSynced()
 	b := true
 	if dns.epController != nil {


zouyee:
@miekg PTAL

miekg:
Good change. Should we extend it more and for instance log the API version
where getting back?

zouyee:
@miekg done

plugin/kubernetes/controller.go
```diff
@@ -312,6 +312,19 @@ func (dns *dnsControl) Run() {
...
 // HasSynced calls on all controllers.
 func (dns *dnsControl) HasSynced() bool {
+	// Pass the simple communication test result as a pre-condition for ready check.
+	// If this is "connection refused" error, it means that most likely apiserver is not responsive.
+	// It doesn't make sense to re-list all objects because most likely we will be able to restart
+	// watch where we ended.
+	// If that's the case wait and resend watch request.
+	v, err := dns.client.Discovery().ServerVersion()
+	if err != nil {
+		log.Warningf("While trying to communicate with apiserver found err: %v", err)
```
miekg:
```suggestion
		log.Warningf("While trying to communicate with apiserver found error: %v", err)
```

plugin/kubernetes/controller.go
```diff
@@ -312,6 +312,19 @@ func (dns *dnsControl) Run() {
...
+		return false
+	}
+	log.Infof("Running with Kubernetes cluster version: v%s.%s. git version: %s. git tree state: %s. commit: %s. platform: %s",
+		v.Major, v.Minor, v.GitVersion, v.GitTreeState, v.GitCommit, v.Platform)
+	log.Infof("Communication with server successful")
```
miekg:
```suggestion
```

miekg (changes requested):

chrisohaver:
Thanks for the PR.  The idea of checking the api connection makes sense.  But more is needed to resolve #3587.

As is, this change will cause the CoreDNS Pod to become unready if the API check fails (assuming the Deployment's readiness probe monitors the ready plugin).  Kubernetes will then stop routing traffic to the unready Pod.  However, it would be left in this unready state indefinitely.

chrisohaver:
To resolve an issue like #3587,  CoreDNS would need to do _one_ of ...
* re-establish its connection to the API
* exit/die
* report unhealthy such that the Deployment's liveness probe kills it

miekg:
Unready never goes unhealthy after it becomes healthy. Doing so would lead
to a cascading failure on every network hiccup

johnbelamaric:
Not sure I follow. All the unready does is remove it from the load balancer
pool, how does that cause a cascading failure? Maybe you mean - CoreDNS /
api server connection lost -> loss of all endpoints in k8s dns -> other
stuff failing? Other than synthetic data plane checks instead of readiness
probes I don't see getting around that.

K8s config also has tolerance built in to prevent transient failures from
cascading - it requires three consecutive readiness probe failures (by
default) in order to go unready.

chrisohaver:
> Unready never goes unhealthy after it becomes healthy. Doing so would lead
to a cascading failure on every network hiccup

We could add some buffer in the readiness probe, such that it would need to be a long hiccup to trigger it.

chrisohaver:
> K8s config also has tolerance built in ... it requires three consecutive readiness probe failures.

Ah, the default already has tolerance, makes sense.

miekg:
No, having them all sync on a shared resource is plain wrong. Adding more
smarts here is also wrong. Caching and forwarding continue to work so there
is no reason to not report readiness true.

chrisohaver:
> No, having them all sync on a shared resource is plain wrong. Adding more smarts here is also wrong. Caching and forwarding continue to work so there is no reason to not report readiness true.
> […](#)

The _readiness probe_, in the Kubernetes _Deployment_.  I wasn't suggesting adding buffer into the _ready plugin_.

chrisohaver:
> To resolve an issue like #3587, CoreDNS would need to do one of ...
> 1. re-establish its connection to the API
> 2. exit/die
> 3.  report unhealthy such that the Deployment's liveness probe kills it

I think a combo of both 1 and 3 would work best... Try to repair the connection, but if that fails for too long, try respawning the Pod.  Not sure of how to best report being "unlive" ... Currently thats per the liveness probe, which points to the health plugin.  Maybe we should point the liveness probe to the ready plugin (with longer tolerance than readiness probe).

johnbelamaric:
We can just exit. That will be faster than waiting for a probe anyway.

chrisohaver:
> We can just exit. That will be faster than waiting for a probe anyway.

I think that _could_ result in many coredns instances going down at the same time during a momentary network outage/glitch that otherwise might have resolved itself.

Tangentially ... are we sure that when the watch dies (as in #3587), that the api server version check fails? Or might it continue to work?

miekg:
+1

Also how does a watch die? And is HasSynced the only way to notice this?
There must be more trivial ways to restart a watch and/or getting status
back on it?

johnbelamaric:
FWIW, the api server isn't "a single shared resource", in that it should have several replicas running and CoreDNS instances should be randomly connected to different ones, and the backing etcd store is also distributed.

miekg:
True, but I've seen things go wrong specularly. The point is, coredns is
part of the data plane and you can't just let the control plane decide to
take it down because that's likely the end of your cluster.

chrisohaver:
> We can just exit. That will be faster than waiting for a probe anyway.

If we do that, then this check should not be in the "ready" path.  I.e. something else that monitors the health of the API connection and dies when it determines the connection is dead.  Is there a watch heartbeat we can monitor?  The watch should be pumping a fairly constant stream of changes to us... If we don't see _anything_ after some time, perhaps we could assume things have gone bad and re-establish the api connection.

yongtang:
I remember client-go is still not able to connect to multiple api servers natively:
https://github.com/kubernetes/kubernetes/issues/18174

That means some additional work from coredns implementation side if coredns is to handle api server endpoint failures.

chrisohaver:
> I remember client-go is still not able to connect to multiple api servers natively:
> [kubernetes/kubernetes#18174](https://github.com/kubernetes/kubernetes/issues/18174)
> 
> That means some additional work from coredns implementation side if coredns is to handle api server endpoint failures.

Multiple instances of coredns can connect to different api instances (each connecting to a single instance)... so there is that redundancy...

johnbelamaric:
Yeah, understood, but right now K8s doesn't offer data plane checks, only
the control plane checks. I have seen other similar instances of this with
updating the cloud loadbalancer (as opposed to just kube-proxy). So it's a
general class of problem in k8s. Though DNS has a special level of
criticality.

zouyee:
> If we do that, then this check should not be in the "ready" path. I.e. something else that monitors the health of the API connection and dies when it determines the connection is dead. 

Regarding the feature of handling connection disconnection, it should be supported by client-go. Readiness probes can be applied in scenarios where applications cannot temporarily provide services. A pod with containers reporting that they are not ready does not receive traffic through Kubernetes Services.

> We can just exit. That will be faster than waiting for a probe anyway.

At the same time, we need to consider the situation of network jitter, so we cannot kill the pod directly.

miekg:
should we merge this or figure out how we lost  a watch in the first place? I'm more inclined to go for the latter; blindly merging without a root cause doesn't sit well with me

chrisohaver:
> should we merge this or figure out how we lost a watch in the first place? I'm more inclined to go for the latter; blindly merging without a root cause doesn't sit well with me

I also lean toward not merging until we understand the issue more. I'm not yet convinced that this would fix the problem its aiming to resolve (watch stops receiving updates). It's quite possible that a synchronous query to the API would work fine in this scenario.  E.g. if the reason the watch stops receiving updates is some bug in the api client/server that only affects that watch.

miekg:
closing here, in favour of learning more about this failure case (which may be in client-go who knows?)

PR #3596, opened 2020-01-13
Remove replace in go.mod

miekg:
Seems these are absolete now?

/cc @yongtang


yongtang (approved):
Yes we can remove those two replaces. Azure was fixed in 3258 and follow up PRs. miekg/dns replace was due to Caddy's usage of miekg/dns in 1.1.3. This has been fixed in Caddy if I remember.

/lgtm

PR #3602, opened 2020-01-17
docs: update README and log plugin

miekg:
README: remove the logo thing as we stopped doing that
log: remote the lines about the clock output as that's gone as well and
     discuss the query log vs other logging a bit.


chrisohaver (approved):
lgtm

PR #3626, opened 2020-01-27
Fix acl page

### LINKED ISSUE #177: plugin page generation: acl link/page not rendering
The acl plugin link and page are not rendering:

https://coredns.io/plugins/

https://coredns.io/plugins/acl/

miekg:
It must have a Name section.

Fixes: https://github.com/coredns/coredns.io/issues/177
(We need more automation to catch this stuff)


bradbeam (approved):

chrisohaver (approved):

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fix acl page ..." ]

/merge

chrisohaver:
> We need more automation to catch this stuff

Could be done in a unit test, which would be a tad awkward since it doesn't actually test Go code.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fix acl page ..." ]
>> We need more automation to catch this stuff
>
>Could be done in a unit test, which would be a tad awkward since it doesn't actually test Go code.

we *can* do it in Go code, but if we go that route we should probably also start parsing
Go code and get rid of our presubmit stuff and make it run via go tests. (quite a bit of
work)

PR #3627, opened 2020-01-27
Use latest Github checkout action

stp-ip:
Removes our step fixing detached state.

-

-

-


yongtang (approved):
lgtm

PR #3629, opened 2020-01-28
presubmit: add whitespace test in go

miekg:
Add a test in test/ to detect whitespace. The presubmits are barely run,
moving this into a Go test will just always report these things.

The other presubmit's, i.e. context and the error messages in test files
can also easliy be moved into go tests

* rm .benchmark.sh as it was not used.
* Clean up the whitespace that this new test found.


stp-ip (approved):
/lgtm

PR #3631, opened 2020-01-28
Remove all shell presubmits

miekg:
This ports the shell presubmit to Go and makes them better; esp the
Errorf/Logf/Fatalf one because it actually parses the code and works of
the AST.

These error will now show up in the travis/circle CI runs where they
belong.


yongtang (approved):
Nice! 👍 /lgtm

test/presubmit_test.go
```diff
@@ -75,3 +78,133 @@ func hasHyphen(path string, info os.FileInfo, _ error) error {
...
+	}
+	se, ok := ce.Fun.(*ast.SelectorExpr)
+	if !ok {
+		return l
+	}
+	id, ok := se.X.(*ast.Ident)
+	if !ok {
+		return l
+	}
+	if id.Name != "t" { //t *testing.T
```
stp-ip, line 135: Is that comment intentional? Just trying to understand the specifics here.

stp-ip (approved):
Besides the question.
/lgtm

test/presubmit_test.go
miekg: yes, "t" signals that Errorf is called from a var named `t` which we assume is a *testing.T (although the type isn't checked)

stp-ip: Thanks for the context. Not sure I like that unchecked assumption, but ok with just the comment.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Remove all sh..." ]
>Thanks for the context. Not sure I like that unchecked assumption, but ok with just the comment.

"it's better than it was" - as it was only a grep

PR #3636, opened 2020-01-29
presubmit: check import path ordering

### LINKED ISSUE #3634: presubmit to check import ordering and layout
the presubmit tests - https://github.com/coredns/coredns/blob/master/test/presubmit_test.go now tests various things. 

One the thing that almost *always* comes up is the ordering of imports. CoreDNS does:
~~~
stdlib imports
<newline>
<coredns imports>
<newline>
<3rd party imports>
~~~
having a presubmit tests that parses the imports (ala https://github.com/coredns/coredns/blob/master/test/presubmit_test.go#L203) and *checks* for this would be very welcome. 

I'm not sure though if you parse the AST you will get this information...

Tentatively marking as "good first use" issue.

miekg:
Add a test for this as well as it's annoying to point out in every code
review.
Fix all the import paths that are flagged by this new test.

Fixes: #3634

### DIFF
--- a/coredns.go
+++ b/coredns.go
@@ -4,10 +4,8 @@ package main
 //go:generate go run owners_generate.go
 
 import (
+	_ "github.com/coredns/coredns/core/plugin" // Plug in CoreDNS.
 	"github.com/coredns/coredns/coremain"
-
-	// Plug in CoreDNS
-	_ "github.com/coredns/coredns/core/plugin"
 )
 
 func main() {
--- a/plugin/cache/prefech_test.go
+++ b/plugin/cache/prefech_test.go
@@ -8,8 +8,8 @@ import (
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/pkg/dnstest"
-
 	"github.com/coredns/coredns/plugin/test"
+
 	"github.com/miekg/dns"
 )
 
--- a/plugin/dnssec/dnskey.go
+++ b/plugin/dnssec/dnskey.go
@@ -9,8 +9,8 @@ import (
 	"time"
 
 	"github.com/coredns/coredns/request"
-	"github.com/miekg/dns"
 
+	"github.com/miekg/dns"
 	"golang.org/x/crypto/ed25519"
 )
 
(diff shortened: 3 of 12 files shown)


bradbeam (approved):
Have we already looked at something like https://github.com/golangci/golangci-lint? 
There's an option around checking for the code to be formatted, I can't remember off the top of my head if it check import path ordering.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] presubmit: ch..." ]
>bradbeam approved this pull request.
>
>Have we already looked at something like https://github.com/golangci/golangci-lint?
>There's an option around checking for the code to be formatted, I can't remember off the top of my head if it check import path ordering.

it doesn't check this particular case. But having something that nits about non-fmt-ed go
code is *also* welcome as I'm also tired of pointing that out in PRs.

miekg:
See https://twitter.com/peterbourgon/status/1222512198972514305 new goimport will do this differently, and just lump them together. I'll update to that version and check, then we can drop this and just check gofmted-ness of the code.

miekg:
I can't replicate the new behaviour of goimports

jameshartig:
It seems like the `-local` option might be what we want here but the docs say it puts local imports *after* the third-party imports. What do you think about that?

I can confirm with some light testing that it puts them after but according to https://github.com/golang/go/issues/31646 it puts them before?

Either way, it doesn't seem to reorganize existing blocks.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] presubmit: ch..." ]
>It seems like the `-local` option might be what we want here but the docs say it puts local imports *after* the third-party imports. What do you think about that?

that would be annoying; then every contributor requires to add that flag.... Right now
goimports leaves them alone *if* you format them as required.

>I can confirm with some light testing that it puts them after but according to https://github.com/golang/go/issues/31646 it puts them before?

if goimports (with no flag) would do something sensible I'm fine with that, but that's not
that case right now...

miekg:
Once goimports is changes/fixed we should probably revert this and stop caring about having three blocks.

Issue #3635, opened 2020-01-28
plugin/forward: max concurrent queries


chrisohaver:
**What would you like to be added**:

An option to set the maximum number of concurrent upstream queries allowed by the _forward_ plugin. If the number of concurrent queries to the upstreams is exceeded, then further queries to the upstreams would be dropped/rejected until the number of concurrent queries falls below the maximum.  This would be an optional setting, defaulting to disabled (no limit).

**Why is this needed**:

Without a maximum number of concurrent connections, a slow upstream connection can cause the number of concurrent connections to grow unbound and exhaust memory. The goal of this option is to lessen the impact of upstream query spikes, but in a way that does not also block other queries to low latency plugins (e.g. served by cache, file, etc).  By imposing the limit in the forward plugin, we don't prevent other plugins from answering while the forwarding limit is reached.

miekg:
this sounds like a good thing to add, think bind9 sets to a 1000 or so, don't really see
an alternative to a manual knob - maybe we think of something better in due time

--
Miek Gieben

miekg:
closed via #3640

PR #3640, opened 2020-01-29
plugins/forward: Add max_concurrent option

chrisohaver:

Adds a concurrent query limit option to the forward plugin, as described in #3635 

#3635

included

no

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -50,6 +50,7 @@ forward FROM TO... {
     tls_servername NAME
     policy random|round_robin|sequential
     health_check DURATION
+    max_queries MAX
 }
 ~~~
 
@@ -83,6 +84,8 @@ forward FROM TO... {
   * `round_robin` is a policy that selects hosts based on round robin ordering.
   * `sequential` is a policy that selects hosts based on sequential ordering.
 * `health_check`, use a different **DURATION** for health checking, the default duration is 0.5s.
+* `max_queries`, limit the number of concurrent queries to **MAX**.  Any new query that would
+  this raise the number of concurrent queries above the **MAX** will result in a SERVFAIL.
 
 Also note the TLS config is "global" for the whole forwarding proxy if you need a different
 `tls-name` for different upstreams you're out of luck.
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -85,7 +85,9 @@ forward FROM TO... {
   * `sequential` is a policy that selects hosts based on sequential ordering.
 * `health_check`, use a different **DURATION** for health checking, the default duration is 0.5s.
 * `max_concurrent` **MAX** will limit the number of concurrent queries to **MAX**.  Any new query that would
-  raise the number of concurrent queries above the **MAX** will result in a SERVFAIL.
+  raise the number of concurrent queries above the **MAX** will result in a REFUSED response. A refusal of
+  this kind does not in itself count as a health failure. When choosing a value for **MAX**, pick a number
+  at least greater than the expected forwarded query rate * the normal latency of the upstream servers.
 
 Also note the TLS config is "global" for the whole forwarding proxy if you need a different
 `tls-name` for different upstreams you're out of luck.
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -105,7 +105,8 @@ If monitoring is enabled (via the *prometheus* plugin) then the following metric
 * `coredns_forward_healthcheck_failure_count_total{to}` - number of failed health checks per upstream.
 * `coredns_forward_healthcheck_broken_count_total{}` - counter of when all upstreams are unhealthy,
   and we are randomly (this always uses the `random` policy) spraying to an upstream.
-
+* `max_concurrent_reject_count_total{}` - counter of the number of queries rejected because the
+  number of concurrent queries were at maximum.
 Where `to` is one of the upstream servers (**TO** from the config), `rcode` is the returned RCODE
 from the upstream.
 
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -84,7 +84,7 @@ forward FROM TO... {
   * `round_robin` is a policy that selects hosts based on round robin ordering.
   * `sequential` is a policy that selects hosts based on sequential ordering.
 * `health_check`, use a different **DURATION** for health checking, the default duration is 0.5s.
-* `max_queries` **MAX** will limit the number of concurrent queries to **MAX**.  Any new query that would
+* `max_concurrent` **MAX** will limit the number of concurrent queries to **MAX**.  Any new query that would
   raise the number of concurrent queries above the **MAX** will result in a SERVFAIL.
 
 Also note the TLS config is "global" for the whole forwarding proxy if you need a different
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -84,8 +84,8 @@ forward FROM TO... {
   * `round_robin` is a policy that selects hosts based on round robin ordering.
   * `sequential` is a policy that selects hosts based on sequential ordering.
 * `health_check`, use a different **DURATION** for health checking, the default duration is 0.5s.
-* `max_queries`, limit the number of concurrent queries to **MAX**.  Any new query that would
-  this raise the number of concurrent queries above the **MAX** will result in a SERVFAIL.
+* `max_queries` will limit the number of concurrent queries to **MAX**.  Any new query that would
+  raise the number of concurrent queries above the **MAX** will result in a SERVFAIL.
 
 Also note the TLS config is "global" for the whole forwarding proxy if you need a different
 `tls-name` for different upstreams you're out of luck.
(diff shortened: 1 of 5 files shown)


plugin/forward/forward.go
```diff
@@ -68,6 +72,14 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 		return plugin.NextOrFailure(f.Name(), f.Next, ctx, w, r)
 	}
 
+	if f.maxQueryCount > 0 {
+		count := atomic.AddInt64(&(f.queryCount), 1)
+		defer atomic.AddInt64(&(f.queryCount), -1)
+		if count > f.maxQueryCount {
```
johnbelamaric: In theory there is a race condition where we may drop some extra queries when we're hovering near the edge. I don't think we care though, not enough to mutex this whole thing.

plugin/forward/forward.go
```diff
@@ -37,6 +38,9 @@ type Forward struct {
 	maxfails      uint32
 	expire        time.Duration
 
+	maxQueryCount int64
+	queryCount int64
```
miekg: for sync atomic these need to be put first in the struct otherwise they don't align on ARM

plugin/forward/forward.go
```diff
@@ -68,6 +72,14 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 		return plugin.NextOrFailure(f.Name(), f.Next, ctx, w, r)
 	}
 
+	if f.maxQueryCount > 0 {
+		count := atomic.AddInt64(&(f.queryCount), 1)
+		defer atomic.AddInt64(&(f.queryCount), -1)
+		if count > f.maxQueryCount {
+			return dns.RcodeServerFailure, errors.New("inflight forward queries exceeded maximum")
```
miekg: Would refuse be better here?

plugin/forward/setup.go
```diff
@@ -211,6 +212,18 @@ func parseBlock(c *caddy.Controller, f *Forward) error {
 		default:
 			return c.Errf("unknown policy '%s'", x)
 		}
+	case "max_queries":
```
miekg: is there a better name with maybe 'concurrent' in it? (not sure there is, but max_queries is not what's implemented here)

miekg:
Also see #2593

plugin/forward/forward.go
rdrozhdzh: No needs to construct the error object here on each query, it can be constructed once as a global var, see `ErrNoHealthy` for example

chrisohaver:
I should add a metric here for number of refused packets.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugins/forwa..." ]
>I should add a metric here for number of refused packets.

yes, that would be a nice enhancement

plugin/forward/forward.go
chrisohaver: I'm not sure.  I think either could work. I initially thought REFUSED, because this is a known/expected failure (soft failure). ... but then changed it to SERVFAIL in a later commit after reading some discussions on the difference between the two. 

There isn't a strict difference: It's fuzzy...
REFUSE is supposed to indicate refusal due to policy. I read that as a function of the client or queried name.  But thats vague, and you could describe a query rate type limit like this as a "policy".
SERVFAIL is supposed to indicate something broken.

plugin/forward/forward.go
chrisohaver: Yes, I lingered on this issue for a bit after seeing it happen during testing, but decided to stick with simple/performant atomic despite this.  The count can go over the limit if overwhelmed, since we don't check for the limit in the same atomic action as the increment.  Would be nice if there was an atomic action "increment-if-less-than", but I don't think this is a big deal, since these increments above the limit are immediately decremented.

Adding an atomic check of the value before the increment lessens, but does not completely eliminate the issue.  But I don't think it's a problem enough to warrant the extra check.

plugin/forward/setup.go
chrisohaver: changed to `max_concurrent`

plugin/forward/forward.go
chrisohaver: Thanks, fixed.

plugin/forward/forward.go
chrisohaver: Thanks, fixed.

plugin/forward/README.md
```diff
@@ -83,6 +84,8 @@ forward FROM TO... {
   * `round_robin` is a policy that selects hosts based on round robin ordering.
   * `sequential` is a policy that selects hosts based on sequential ordering.
 * `health_check`, use a different **DURATION** for health checking, the default duration is 0.5s.
+* `max_concurrent` **MAX** will limit the number of concurrent queries to **MAX**.  Any new query that would
```
miekg, line 87: how should one determine this value, what's a good one for **MAX**?

plugin/forward/forward.go
```diff
@@ -36,12 +39,17 @@ type Forward struct {
...
+	maxConcurrent int64
 
 	opts options // also here for testing
 
 	Next plugin.Handler
 }
 
+// ErrLimitExceeded indicates that a query was rejected because the number of concurrent queries has exceeded
+// the maximum allowed (maxConcurrent)
+var ErrLimitExceeded = errors.New("concurrent queries exceeded maximum")
```
miekg: I think is valuable to include the configured max here.

plugin/forward/README.md
miekg: We should probably add words on why we're not nacking health if this happens (i.e. to avoid cascading failure)

plugin/forward/forward.go
```diff
@@ -68,6 +76,15 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 		return plugin.NextOrFailure(f.Name(), f.Next, ctx, w, r)
 	}
 
+	if f.maxConcurrent > 0 {
+		count := atomic.AddInt64(&(f.concurrent), 1)
+		defer atomic.AddInt64(&(f.concurrent), -1)
+		if count > f.maxConcurrent {
+			MaxConcurrentRejectCount.Add(1)
+			return dns.RcodeServerFailure, ErrLimitExceeded
```
miekg: Difference between SERVFAIL or REFUSED is fuzzy. REFUSED is better here, because the server is fine (no SERVFAIL), but actually refuses to do work for you

rdrozhdzh: from rfc1035, section 4.1.1.

> Server failure - The name server was unable to process this query due to a problem with the name server.

> Refused - The name server refuses to perform the specified operation for policy reasons.  For example, a name server may not wish to provide the information to the particular requester, or a name server may not wish to perform a particular operation (e.g., zone transfer) for particular data.

I think SERVFAIL suits better in this case. This is a temporary error. If we return REFUSED, client will probably decide that this query is not permitted and will not retry. In case of SERVFAIL it could try again.

chrisohaver: In a sense, this kind of failure is like a timeout, in that there is some transient condition (some action has exceeded a configured limit) that makes it so we are not going to answer.  Although in a timeout, we _cannot_ get an answer, and in the concurrent limit case, we _choose_ not to.

I think clients would treat either of these equally, as a negative response.  Since the definition of REFUSED/SERVFAIL is fuzzy, it would be unwise for a client to draw special conclusions in a general context for one answer vs the other.

If the desired behavior is to have the client retry, then I think we would need to drop the response (not respond at all).

miekg:
Fair enough, servfail works for me.

We should never not reply.

plugin/forward/setup.go
```diff
@@ -211,6 +213,19 @@ func parseBlock(c *caddy.Controller, f *Forward) error {
...
+			return c.ArgErr()
+		}
+		n, err := strconv.Atoi(c.Val())
+		if err != nil {
+			return err
+		}
+		if n < 0 {
+			return fmt.Errorf("max_concurrent can't be negative: %d", n)
+		}
+		ErrLimitExceeded = errors.New("concurrent queries exceeded maximum " + c.Val())
```
rdrozhdzh: Here may be a race condition. When coredns is reloaded (e.g. with SIGUSR1) the old instance of forward plugin is still handling traffic and may access `ErrLimitExceeded`, and the new instance of forward plugin can update `ErrLimitExceeded` at the same time

miekg:
Just create the err in forward when you return, this is getting silly. If
you worry about allocations, there is plenty of lowhanging fruit with more
bang for the buck

rdrozhdzh:
Well, since the error text is not a constant anymore, that would be a reasonable choice.
Alternatively, the error can be saved in a `Forward` object.

plugin/forward/README.md
chrisohaver: I put in notes for these.  For picking a value, I describe a lower bound.  

I didn't tackle describing an upper bound yet.  For an upper bound (to bind memory usage), it's a matter of how much memory an average query takes up inflight in CoreDNS - which is variable (e.g. encryption).  I can try to estimate this experimentally.  I don't know how globally applicable or accurate that would be.

miekg:
I think mem usage is a good one..with the assumption a goroutine is 2kb. I
was pondering a similar text for discarding packets directly in miekg/dns,
but this pr is probably better, because here you can actually have
lingering go-routines.
Also note we timeout after 5s (I think), we can probably discard writing an
answer if that happens, cause the client is long gone (but that's a
different pr)

miekg (approved):

PR #3642, opened 2020-01-30
plugin/kubernetes: fix metadata

chrisohaver:

Kubernetes metadata is empty when the matched serverblock zone is not an exact match for the plugin's handled zone ... e.g. if the serverbock's zone is ".", and the kubernetes plugin zone is "cluster.local".

None

None

No

### DIFF
--- a/plugin/kubernetes/metadata.go
+++ b/plugin/kubernetes/metadata.go
@@ -10,11 +10,9 @@ import (
 
 // Metadata implements the metadata.Provider interface.
 func (k *Kubernetes) Metadata(ctx context.Context, state request.Request) context.Context {
-	origZone := state.Zone
-	state.Zone = plugin.Zones(k.Zones).Matches(state.Name())
+	zone := plugin.Zones(k.Zones).Matches(state.Name())
 	// possible optimization: cache r so it doesn't need to be calculated again in ServeDNS
-	r, err := parseRequest(state)
-	state.Zone = origZone
+	r, err := parseRequest(state.Name(), zone)
 	if err != nil {
 		metadata.SetValueFunc(ctx, "kubernetes/parse-error", func() string {
 			return err.Error()
--- a/plugin/kubernetes/metadata.go
+++ b/plugin/kubernetes/metadata.go
@@ -10,9 +10,11 @@ import (
 
 // Metadata implements the metadata.Provider interface.
 func (k *Kubernetes) Metadata(ctx context.Context, state request.Request) context.Context {
+	origZone := state.Zone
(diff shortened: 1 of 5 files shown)


plugin/kubernetes/metadata.go
```diff
@@ -3,12 +3,14 @@ package kubernetes
...
 	"context"
 
+	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/metadata"
 	"github.com/coredns/coredns/request"
 )
 
 // Metadata implements the metadata.Provider interface.
 func (k *Kubernetes) Metadata(ctx context.Context, state request.Request) context.Context {
+	state.Zone = plugin.Zones(k.Zones).Matches(state.Name())
```
johnbelamaric: Does this mutate the state in some undesired way?

Also, can we have a unit test for the bug and show that this fixes it?

chrisohaver: > Does this mutate the state in some undesired way? 

Yes, see the second commit...

chrisohaver: A more graceful way to handle this would be to have parse accept the matched zone as a string parameter, instead of pulling it from state.

That the kubernetes plugin modifies state.Zone at all is potentially problematic for fallthrough cases.

chrisohaver: > Also, can we have a unit test for the bug and show that this fixes it?

Added a test and verified the unit test fails when without the fix.

chrisohaver:
FYI @johnbelamaric  - I found this while testing an example of multi-tenant DNS walling using the policy plugins.  Adding an example of this has been on my todo list since the the policy+opa plugins were discovered and demoed by the sig-multitenant work group in December.

johnbelamaric:
Nice! Have you looked at https://github.com/coredns/coredns-opa ? It's
just for bootstrapping the coredns+opa running in GKE.  Sounds like you are
a stop ahead of that. Torin & I will be doing a talk on this in Amsterdam,
I believe I mentioned that to you before.

plugin/kubernetes/metadata.go
johnbelamaric: can you go ahead with the "more graceful way"? Is there a blocker for that?

chrisohaver: Sure, no blocker - it just touches more code.

chrisohaver:
Latest commit changes `parse()` to require separate name/zone strings instead of relying on a  mutated state.  It does not address the fact that kubernetes' `ServeDNS()` still mutates the state, because some things other than `parse()` appear to depend on it.  IMO that should be investigated/fixed in a separate PR.

Issue #3620, opened 2020-01-26
plugin/forward: disable tls heath check?


miekg:
When an upstream conn is returning error we will heathcheck the endpoint until it comes, the TLS this means a new tls connection every 0.5s. We should probably do better:

we can a) do exponential backoff *with* random in-between retries or b) only do tcp conns and forgo TLS when error checking.

I think both are valid things to implement.

miekg:
The backoff stuff should help here.

PR #3643, opened 2020-01-31
plugin/forward: Add exponential backoff to TLS healthcheck

rickywraith:
Improves healthcheck retry by adding exponential backoff to avoid new TLS connection every 0.5s

#3620

plugin/forward README.md

No

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/apache/thrift v0.13.0 // indirect
 	github.com/aws/aws-sdk-go v1.28.9
 	github.com/caddyserver/caddy v1.0.4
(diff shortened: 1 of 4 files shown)


miekg:
Nice. Does this _def_ to random probes while backing off? This also warrant some README.md changes *and* it makes sense to do this for all health checks; singling out tls is a bit odd (although understandable)

rickywraith:
@miekg the `NextBackOff()` that is used to generate intervals in between retries will return a randomized interval while backing off. And I think it will work for all healthchecks since it is implemented as part of `Probe` in `plugin/pkg/up/up.go`. I updated the README to reflect the change.

miekg (approved):

PR #3651, opened 2020-02-04
plugin/pkg/up: make default intervals shorter

miekg:
I think 15 min is too high, make this lower to react faster.


miekg:
cc @RickyRajinder

yongtang (approved):
LGTM

rickywraith (approved):

rickywraith:
This would also require the README to be updated

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/pkg/up..." ]
>This would also require the README to be updated

duh, of course :-) will do

/Miek

--
Miek Gieben

Issue #2998, opened 2019-07-11
plugin/forward: Add new latency weighted policy


stp-ip:
With various upstreams configured and the network not always behaving as one has planned, it would be interesting to have a latency weighted policy.

We could send an incoming request in parallel every x timeframe and use the top performing upstream until the next test request.

Just a thought and wanted to write it down already.
Might add additional thoughts going forward.

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] [forward] Add new..." ]
>With various upstreams configured and the network not always behaving as one has planned, it would be interesting to have a latency weighted policy.
>
>We could send an incoming request in parallel every x timeframe and use the top performing upstream until the next test request.
>
>Just a thought and wanted to write it down already.
>Might add additional thoughts going forward.

I think other (recursive) server call this DNS RTT *banding*.

stp-ip:
Thanks, I think that was the keyword I was missing!

erbesharat:
If it's ok I want to try to work on this issue :)

stp-ip:
Any specific prerequisites for a PR @miekg ?

Some thoughts:
In the bootstrap phase we could either wait for the first query and then send it upstream in parallel, answer with the fastest response and note down response latency to be used to determine the fastest upstream server to be used until the next latency "test".
Alternatively we could do prefetching and contact upstream nameservers with a predefined query and note down the response latency as well. This would enable more continuous and standardizes testing, but might be different due to caches being warm in upstreams, but it might be easier to debug as it's always just one and the same query and could be tweaked via its TTL.

For the TCP/TLS case we could do a follow on query to remove any problems with hot/cold connections.
This could be alleviated by making upstream connections be pre warmed in all cases aka hold a standing connection.

Additionally if the currently top performing upstream doesn't respond or responds with worse latency than the second in the performance list, we will query the next one in line and initiate an additional latency test outside the usual interval.

Retesting will be done either on upstream failures, latency drop of top performing server or on the next interval.

Config suggestion:
```
forward FROM TO... {
    policy latency
    latency_interval 5m
    latency_query ZONE [TYPE] # fallback: ". IN NS" similar to health check
}
```

miekg:
Some quick notes. Maybe banding can be applied for other policies as well,
otherwise just a extra latency policy would suffice as a config option.
Impl:  Have 3 buckets: slow, medium and fast, and have some relative weight
for them. Fill the buckets during querying and move nameservers around in
the buckets as required.

[...]

PR #3652, opened 2020-02-04
WIP: feat: add plugin fanout

denis-tingaikin:
### Which issues (if any) are related?
https://github.com/coredns/coredns/issues/3653

### TODO
- [X] Select the solution for integration this plugin into CoreDNS https://github.com/coredns/coredns/issues/3653
- [X] Code cleanup
- [X] Add support for metrics
- [X] Add support for UDP, DNS TLS
- [X] Add tests
- [X] Add documentation
- [ ] Pass review


core/plugin/zplugin.go
```diff
@@ -22,6 +22,7 @@ import (
 	_ "github.com/coredns/coredns/plugin/erratic"
 	_ "github.com/coredns/coredns/plugin/errors"
 	_ "github.com/coredns/coredns/plugin/etcd"
+	_ "github.com/coredns/coredns/plugin/fanout"
```
chrisohaver: This should probably go after the file plugin.

chrisohaver, line 25: This needs to be added to `plugin.cfg`. zplugin.go is generated.

core/dnsserver/zdirectives.go
```diff
@@ -55,4 +55,5 @@ var Directives = []string{
 	"whoami",
 	"on",
 	"sign",
+	"fanout",
```
chrisohaver: This needs to be added to plugin.cfg. zdirectives.go is generated from plugin.cfg.

Order is important ... This controls the order of operation of plugins (order in Corefile does not).  Probably should be right after forward.

plugin/fanout/README.md
```diff
@@ -0,0 +1,108 @@
...
+* `coredns_fanout_response_rcode_count_total{to, rcode}` - count of RCODEs per upstream.
+* `coredns_fanout_healthcheck_failure_count_total{to}` - number of failed health checks per upstream.
+* `coredns_fanout_healthcheck_broken_count_total{}` - counter of when all upstreams are unhealthy,
+  and we are randomly (this always uses the `random` policy) spraying to an upstream.
+
+Where `to` is one of the upstream servers (**TO** from the config), `rcode` is the returned RCODE
+from the upstream.
+
+## Examples
+Proxy all requests within `example.org.` to a nameserver running on a different port:
```
chrisohaver: I don't think this single upstream example is particularly compelling (especially not to lead with).

plugin/fanout/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+~~~ corefile
+. {
+    fanout . tls://9.9.9.9 {
+       tls_servername dns.quad9.net
+    }
+}
+~~~
+
+Sends parallel requests between five resolvers via UDP uses two workers. The first positive response from proxy will be provided ad result.
```
chrisohaver:
```suggestion
Sends parallel requests between five resolvers via UDP uses two workers. The first positive response from a proxy will be provided as the result.
```

plugin/fanout/README.md
```diff
@@ -0,0 +1,108 @@
...
+## Examples
+Proxy all requests within `example.org.` to a nameserver running on a different port:
+
+~~~ corefile
+example.org {
+    fanout . 127.0.0.1:9005
+}
+~~~
+
+Sends parallel requests between three resolvers, one of which has a IPv6 address via TCP. The first success connected response from proxy will be provided ad result.
```
chrisohaver:
```suggestion
Sends parallel requests between three resolvers, one of which has a IPv6 address via TCP. The first positive response from a proxy will be provided as the result.
```

core/plugin/zplugin.go
denis-tingaikin: thanks, fixed.

core/dnsserver/zdirectives.go
denis-tingaikin: thanks, fixed.

plugin/fanout/README.md
denis-tingaikin: thanks, fixed.

plugin/fanout/README.md
denis-tingaikin: thanks, fixed.

plugin/fanout/README.md
denis-tingaikin: thanks, fixed.

plugin.cfg
```diff
@@ -64,3 +64,4 @@ erratic:erratic
 whoami:whoami
 on:github.com/caddyserver/caddy/onevent
 sign:sign
+fanout:fanout
```
chrisohaver: IMO, this should be moved to immediately after forward.

denis-tingaikin: sure, fixed

plugin/fanout/README.md
```diff
@@ -0,0 +1,108 @@
...
+* `tls_servername` **NAME** allows you to set a server name in the TLS configuration; for instance 9.9.9.9
+  needs this to be set to `dns.quad9.net`. Multiple upstreams are still allowed in this scenario,
+  but they have to use the same `tls_servername`. E.g. mixing 9.9.9.9 (QuadDNS) with 1.1.1.1
+  (Cloudflare) will not work.
+
+* `fail-count` is the number of subsequent failed health checks that are needed before client stops trying to connect.
+* `worker-count` is the number of parallel queries per request.
+* `policy` is specifies the policy to use for selecting response from proxies.
+  * `any` is a policy that returns any first response from success connected proxies.
+  * `first-positive` is a policy that returns any first non-negative response from proxies.
```
chrisohaver: should mention that this is the default policy.

denis-tingaikin: thanks! Fixed

plugin/fanout/README.md
```diff
@@ -0,0 +1,108 @@
...
+    The server certificate is verified with the system CAs
+  * `tls` **CERT** **KEY**  **CA** - client authentication is used with the specified cert/key pair.
+    The server certificate is verified using the specified CA file
+* `tls_servername` **NAME** allows you to set a server name in the TLS configuration; for instance 9.9.9.9
+  needs this to be set to `dns.quad9.net`. Multiple upstreams are still allowed in this scenario,
+  but they have to use the same `tls_servername`. E.g. mixing 9.9.9.9 (QuadDNS) with 1.1.1.1
+  (Cloudflare) will not work.
+
+* `fail-count` is the number of subsequent failed health checks that are needed before client stops trying to connect.
+* `worker-count` is the number of parallel queries per request.
```
chrisohaver: Need to describe the default value (4, or the upstream client count, whichever is higher i think).  Also should advise on why a user would change this, and how to select a value.

chrisohaver:
I assume that much of the framework of this code is copied from _forward_.

IMO, #3640 should be included in _fanout_ too.  Fanout will significantly increase the number of concurrent queries.

Probably should check to see if any other changes/fixes in forward have been made after the commit from which this code was copied.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] feat: add plu..." ]
>I assume that much of the framework of this code is copied from _forward_.
>
>IMO, #3640 should be included in _fanout_ too.  Fanout will significantly increase the number of concurrent queries.
>
>Probably should check to see if any other changes/fixes in forward have been made after the commit from which this code was copied.

hmmm. I think we should make this a minimal plugin that does what is supposed to be doing
and make clear it's a debug-level plugin. You do not want to run it in production. This
entails: minimal options, no health checking, etc. That should also minimize code
duplication with forward.

chrisohaver:
> Each incoming DNS query that hits the CoreDNS fanout plugin will be replicated in parallel to each listed IP (i.e. the DNS servers). The first non-negative response from any of the queried DNS Servers will be forwarded as a response to the application's DNS request.

How do things play out when the worker count is less than the number of upstream servers?

"First to reply" implies that _fastest_ server always wins, but if the number of workers is less than the number of servers, then which server answers "first" is also controlled by an arbitrary relation between the order of servers in the list, _and_ how fast early servers in the list respond with negative answers.

denis-tingajkin:
> How do things play out when the worker count is less than the number of upstream servers?

In this case `worker count` will be equal to a number of upstream servers. Take a look at `setup.go:121`

chrisohaver:
> > How do things play out when the worker count is less than the number of upstream servers?
> 
> In this case `worker count` will be equal to a number of upstream servers. Take a look at `setup.go:121`


```
	minWorkers := f.workerCount

 	if minWorkers > len(f.clients) {
 		minWorkers = len(f.clients)
 	}

 	f.workerCount = minWorkers
```
That code appears to enforce that the number of workers does not _exceed_ the number of upstream dns servers.

denis-tingaikin:
@chrisohaver 

> That code appears to enforce that the number of workers does not exceed the number of upstream dns servers.

Yes, you are right, it is related to another case...


> How do things play out when the worker count is less than the number of upstream servers?

You caught an interesting case. 

I think the worker count should be equal to the proxy count by default. 

Also, as noticed, the user can use the worker count less than the proxy count. 
In most cases, the windowing workers will do only one try to connect to next dns servers  (if one of the first proxies will able to connect and it return answer). It means that if the worker count is less than the proxy count then in most cases last proxies will not try to connect(they will if only the first proxies return negative responses or first proxies were not able to connect). So I think if the user uses the worker count less than the proxy count then he wants to solve some specific performance issue. For example, the User uses `resolv.conf` file with 100 entries and he doesn't want every time sent 100 parallel queries.

Is it make sense?

denis-tingaikin:
@chrisohaver 

> I assume that much of the framework of this code is copied from forward.

Plugin `fanout` has some similar concept things like `Transport`, `Health` (health using only if proxies have attempts to reconnect. In this case, if the first attempt was a not success then the health will check the proxy. If the check will not success then the proxy will stop tries).

Also, I used similar solutions for tls configs and for metrics and I copied some tests for example for setup and adapted it. 

> Probably should check to see if any other changes/fixes in forward have been made after the commit from which this code was copied.

I agreed with @miekg.
At this moment `fanout plugin` has minimal needed things to work and it has minimal intersect with forward plugin. 


Let me know If I need to add/remove options for fanout plugin.

plugin/fanout/utils.go
```diff
@@ -0,0 +1,60 @@
...
+
+func logIfNotNil(err error) {
+	if err == nil {
+		return
+	}
+	stack := debug.Stack()
+	log.Errorf("err :%v \nstack:\n%v", err, string(stack))
+}
+
+func toDnstap(ctx context.Context, host string, protocol string, state request.Request, reply *dns.Msg, start time.Time) error {
```
denis-tingaikin, line 22: @miekg I copied this func from `forward` plugin because in `forward` pkg this func is private https://github.com/coredns/coredns/blob/master/plugin/forward/dnstap.go#L15.
Can I make this func in forward pkg public and reuse it for this plugin?

plugin/fanout/README.md
denis-tingaikin: Should be fixed, thanks!

plugin/fanout/README.md
```diff
@@ -0,0 +1,108 @@
...
+    fanout . 127.0.0.1:9005 127.0.0.1:9006 127.0.0.1:9007 127.0.0.1:9008
+}
+~~~
+
+Sends parallel requests between three resolvers, one of which has a IPv6 address via TCP. The first response from proxy will be provided as the result.
+
+~~~ corefile
+. {
+    fanout . 10.0.0.10:53 10.0.0.11:1053 [2003::1]:53 {
+        protocol TCP
```
miekg: I consider this a debug aid, not something people need to run normally as it puts strain on the DNS system, so this should not support a lot of feature. Protocol selection is a maybe in my book, `policy` selection should be removed entirely. 
As said earlier: I'm not looking to have this a cut and paste job from *forward*. I'm fine if you think differently, but then we should make this an external plugin.

denis-tingaikin: Can you please describe a use case for debugging with this plugin?

denis-tingaikin: > policy selection should be removed entirely.

Thanks, done

denis-tingaikin: @miekg  I also think this plugin can be used not only for debugging.
I agree with this comment https://github.com/coredns/coredns/issues/3653#issuecomment-583280957
I think it also can be useful for people or for some specific goals where need functionality to fanout request.

denis-tingajkin: @miekg two open questions are here:
1. What a usecase for debugging? (It could help to make changes in the plugin and add cover the usecase  by unit tests)
2. Why the functionality of this plugin can not be used by people? // related to https://github.com/coredns/coredns/issues/3653#issuecomment-583280957

miekg:
yes debugging aid; we don't want to spray upstreams with multiple useless queries by default.

I see many many nits in the PR (new logging functions, channels being use to return values from functions ). And more than a fair share of cut and paste from forward.

I don't have the energy to beat this into shape, as such this will be an external plugin

denis-tingajkin:
@miekg 

> yes debugging aid; we don't want to spray upstreams with multiple useless queries by default.

You're right. And I see only the next ways of using this plugin: 
1. Debugging. But here I faced a problem(which I described above) that I need an example... 
2. Using for private upstreams servers (i.e. when the user knows that he will pay resources for fast responses and he will use private upstreams servers).

Q: Could you please share any example debugging? It will be super useful.

denis-tingaikin:
@miekg 
> I see many many nits in the PR (new logging functions, channels being use to return values from functions )

Thanks for the feedback.

- [X] fix logging
- [X] fix channels being use to return values from functions

Please let me know what else should be fixed. I am very interested to make this plugin better.

denis-tingajkin:
@miekg 
> And more than a fair share of cut and paste from forward.

Yes, it is ugly. Some super-useful functions on forward plugin are private. 

Q: Can some useful things of forward plugin be moved to some pkg for sharing(for example to `pkg/plugin`)?
I think it can be useful not only for this plugin but for many next new internal/external plugins.

denis-tingaikin:
> And more than a fair share of cut and paste from forward. 

@miekg Could you please explain this?

I looked into code and found that copied only next private funcs from forward plugin:
`isAllowedDomain`
`toDnstap`

NOTE: I asked about refactoring to remove code dup:
1) https://github.com/coredns/coredns/pull/3652#discussion_r381072157
2) https://github.com/coredns/coredns/issues/3653#issuecomment-589751369

denis-tingaikin:
>  we don't want to spray upstreams with multiple useless queries by default.

The fanout plugin returns the **first positive answer** by default (please read use case section in the attached issue). 
It means that  "multiple useless queries" could be useful. For example, we have a list of upstream servers: A, B, C, D.
And we have a query: "example.". 
Only server D can resolve this query.
In this case, `Forward` plugin will never return the IP from server D (if the first servers A,B,C are fine)

denis-tingaikin:
@miekg  ^^^

denis-tingaikin:
@miekg , @chrisohaver  I've added an example of usage https://github.com/coredns/coredns/issues/2998#issuecomment-596179403

plugin/fanout/README.md
edwarnicke: @miekg Actually, in [Network Service Mesh](https://networkservicemesh.io/) we use this not as a debugging aid, but a first class tool.  Network Service Mesh allows a workload to connect (at the level of the workload) to multiple mutually ignorant 'Network Services' simultaneously.  For example, imagine a large auto maker providing a 'partner-network-service' to thousands of its suppliers, and a parts supplier producing a workload that participates in both 'partner-network-service' from the automaker and 'manufacturing-network-service' from its manufacturing group.  Both of those have independent, mutually ignorant DNS systems.  The fanout plugin allows us to participate simultaneously in both.  Obviously in this case you *don't* want the enormous tax of having to manually coordinate those DNS systems into a single DNS system, and you also don't want the latency of having to resolve in serial.  That's what the fanout plugin solves for.  So not debugging, but rather simultaneous participation in multiple mutually ignorant DNS systems.

edwarnicke:
> yes debugging aid; we don't want to spray upstreams with multiple useless queries by default.

Since the purpose of the plugin is to allow simultaneous participation in multiple mutual ignorant DNS systems (rather than multiple representative servers of a single DNS system as forwarder does)... the multiple queries *aren't* useless.

> 
> I see many many nits in the PR (new logging functions, channels being use to return values from functions ). And more than a fair share of cut and paste from forward.
> 
> I don't have the energy to beat this into shape, as such this will be an external plugin

That's entirely up to you guys in coredns :)  Just want to make sure you understand *what* we are trying to accomplish here.  If we just wanted redundant providers of a single DNS system, we'd use forwarder (which is excellent for that)... but we simultaneous participation in multiple mutually ignorant DNS systems... that's what fanout does :)

chrisohaver:
If the zones that each "mutually ignorant" dns server is authoritative for are known, then you can create server stanzas for each zone... for example:

```
zone-a zone-b {
   forward . 10.0.0.10
}

zone-c zone-d {
   forward . 10.0.0.12
}
```

This way, a query for `foo.zone-c.` is routed directly to `10.0.0.12`, instead of also querying `10.0.0.10`.

edwarnicke:
@chrisohaver Thanks for the suggestion... on the principle of 'try the simple thing first' we'll try going that route :)  There still exists some possibility that we will hit cases that may require a more complex approach... but its probably better to *start* simple and adapt to reality rather than imagining ultra-complex reality and starting complex :)

I can't promise we won't be back with the fanout plugin... but if we are, it will be because we hit the limitations of your suggestion in the real world :)

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] WIP: feat: ad..." ]
>@miekg Actually, in [Network Service Mesh](https://networkservicemesh.io/) we use this not as a debugging aid, but a first class tool.  Network Service Mesh allows a workload to connect (at the level of the workload) to multiple mutually ignorant 'Network Services' simultaneously.  For example, imagine a large auto maker providing a 'partner-network-service' to thousands of its suppliers, and a parts supplier producing a workload that participates in both 'partner-network-service' from the automaker and 'manufacturing-network-service' from its manufacturing group.  Both of those have independent, mutually ignorant DNS systems.  The fanout plugin allows us to participate simultaneously in both.  Obviously in this case you *don't* want the enormous tax of having to manually coordinate those DNS systems into a single DNS system, and you also don't want the latency of having to resolve in serial.  That's what the fanout plugin solves for.  So not debugging, but rather simultaneous participation in multiple mutually ignorant DNS systems.

DNS in k8s clusters is already a mess because of the giant search path, adding (and
making it easier), to 2x or 3x the amount of useless queries send will only make that
worse.

This is really a case of: Do you want DNS outages, because this is how you get DNS outages

miekg:
put up as external plugin: https://coredns.io/explugins/fanout/

PR #3671, opened 2020-02-13
Cleanup metrics

SuperQ:
Cleanup a variety of metric issues.
* Eliminate department of redundancy "count_total" naming.
* Use the plural of the unit when appropriate. (ex, "requests")
* Remove label names from metric names where appropriate. (ex, "rcode")
* Simplify request metrics by consolidating type label in to the base
request counter.
* Re-generate man pages.

Yes, this is going to probably annoy and break a lot of users, but the current metrics violate a lot of Prometheus best practices.

### DIFF
--- a/plugin/acl/README.md
+++ b/plugin/acl/README.md
@@ -73,8 +73,8 @@ example.org {
 
 If monitoring is enabled (via the _prometheus_ plugin) then the following metrics are exported:
 
-- `coredns_request_block_count_total{server, zone}` - counter of DNS requests being blocked.
+- `coredns_dns_blocked_requests_total{server, zone}` - counter of DNS requests being blocked.
 
-- `coredns_request_allow_count_total{server}` - counter of DNS requests being allowed.
+- `coredns_dns_allowed_requests_total{server}` - counter of DNS requests being allowed.
 
 The `server` and `zone` labels are explained in the _metrics_ plugin documentation.
--- a/plugin/acl/metrics.go
+++ b/plugin/acl/metrics.go
@@ -11,14 +11,14 @@ var (
 	RequestBlockCount = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Namespace: plugin.Namespace,
 		Subsystem: "dns",
-		Name:      "request_block_count_total",
+		Name:      "acl_blocked_requests_total",
 		Help:      "Counter of DNS requests being blocked.",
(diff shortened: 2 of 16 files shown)


bradbeam (approved):
Overall I'm fine with the change, they make sense, but I'm curious how we want to handle the transition to these newer metrics. 
Should we add in the new ones for a release or two and then deprecate the old ones?

bradbeam:
@prologic fyi, not sure if you're still maintaining the grafana dashboard

SuperQ:
If this was one or two fixes, doubling the metrics for those instances would be reasonable. But this required a pretty major refactor in naming. Doubling would be a huge load increase for monitoring systems.

For the breaking changes in the node_exporter, we recommended a couple options.

* Doubling the queries in Grafana dashboards (no performance issue here, works well).
* Adding recording rules for forwards or backwards compatibility to transition users over.

If you already have recording rules in Prometheus for things, you can do something like this:

```
rate(new_metric_total[1m]) or rate(old_metric_total[1m])
```

But of course that depends on what kind of label matching complications you have.

johnbelamaric:
Thanks Ben. At the very least this needs to wait for 1.7.0, we shouldn't change these in a patch release. For config options we have a deprecation period. I don't think we ever made that promise for metrics. But I am hesitant to just cut over without any break in period.

SuperQ:
Absolutely, this should happen in the next minor version. There's no real rush for this.

SuperQ:
Rebased with https://github.com/coredns/coredns/pull/3685.

SuperQ:
@miekg Rebased this, removed the docs changes.

miekg:
closed via #3776

prologic:
> @prologic fyi, not sure if you're still maintaining the grafana dashboard

I am and do. Has something changed I need to be aware of? I wasn't following this thread until you tagged me :)

SuperQ:
@prologic Yes, many of the common metric names have changed.

There are two good options to produce backwards-compatible dashboards.

Use `or`.
* `rate(new_metric_name[$__interval]) or rate(old_metric_name[$__interval])`

Use duplicate queries.
* Query A: `rate(new_metric_name[$__interval])`
* Query B: `rate(old_metric_name[$__interval])`

prologic:
Right thanks! I'll update my version of CoreDNS and update and re-publish the Dashbaord.

PR #3679, opened 2020-02-17
Add configuration flag to set if RecursionDesired should be set on health checkers in Forward-plugin

ctryti:
We would like to configure the health checker for the proxies whether they should do recursive queries (default) or turn it off by setting health_check_non_recursive in the configuration.

Issues in our environment involve upstream queries to time out before the answer gets back because of how the dns servers (upstream) are configured. They are out of our control and we only care if they are alive.

Configuration file for the forward plugin should reflect the new optional flag to turn off recursion on the health checker.

No. The default is to keep recursion requests on by default as is coded in previous versions of forward plugin.

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -52,6 +52,7 @@ forward FROM TO... {
     tls_servername NAME
     policy random|round_robin|sequential
     health_check DURATION
+    health_check_non_recursive
     max_concurrent MAX
 }
 ~~~
@@ -86,6 +87,7 @@ forward FROM TO... {
   * `round_robin` is a policy that selects hosts based on round robin ordering.
   * `sequential` is a policy that selects hosts based on sequential ordering.
 * `health_check`, use a different **DURATION** for health checking, the default duration is 0.5s.
+* `health_check_non_recursive`, sets the RecursionDesired-flag of the dns-query used in health checking to `false`. The flag is default `true`.
 * `max_concurrent` **MAX** will limit the number of concurrent queries to **MAX**.  Any new query that would
   raise the number of concurrent queries above the **MAX** will result in a SERVFAIL response. This
   response does not count as a health failure. When choosing a value for **MAX**, pick a number
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -51,8 +51,7 @@ forward FROM TO... {
     tls CERT KEY CA
     tls_servername NAME
     policy random|round_robin|sequential
-    health_check DURATION
-    health_check_non_recursive
+    health_check DURATION [no_rec]
     max_concurrent MAX
 }
 ~~~
@@ -86,8 +85,10 @@ forward FROM TO... {
   * `random` is a policy that implements random upstream selection.
   * `round_robin` is a policy that selects hosts based on round robin ordering.
   * `sequential` is a policy that selects hosts based on sequential ordering.
-* `health_check`, use a different **DURATION** for health checking, the default duration is 0.5s.
-* `health_check_non_recursive`, sets the RecursionDesired-flag of the dns-query used in health checking to `false`. The flag is default `true`.
+* `health_check` configure the behaviour of health checking of the upstream servers
+  * `<duration>` - use a different duration for health checking, the default duration is 0.5s.
+  * `no_rec` - optional argument that sets the RecursionDesired-flag of the dns-query used in health checking to `false`.
+    The flag is default `true`.
 * `max_concurrent` **MAX** will limit the number of concurrent queries to **MAX**.  Any new query that would
   raise the number of concurrent queries above the **MAX** will result in a SERVFAIL response. This
   response does not count as a health failure. When choosing a value for **MAX**, pick a number
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -52,7 +52,7 @@ type Forward struct {
 
 // New returns a new Forward.
 func New() *Forward {
-	f := &Forward{maxfails: 2, tlsConfig: new(tls.Config), expire: defaultExpire, p: new(policy.Random), from: ".", hcInterval: hcInterval}
+	f := &Forward{maxfails: 2, tlsConfig: new(tls.Config), expire: defaultExpire, p: new(policy.Random), from: ".", hcInterval: hcInterval, opts: options{false, false, true}}
 	return f
 }
 
@@ -224,8 +224,9 @@ var (
 
 // options holds various options that can be set.
 type options struct {
-	forceTCP  bool
-	preferUDP bool
+	forceTCP           bool
+	preferUDP          bool
+	hcRecursionDesired bool
 }
 
 const defaultTimeout = 5 * time.Second
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -52,7 +52,7 @@ type Forward struct {
 
 // New returns a new Forward.
 func New() *Forward {
-	f := &Forward{maxfails: 2, tlsConfig: new(tls.Config), expire: defaultExpire, p: new(policy.Random), from: ".", hcInterval: hcInterval, opts: options{false, false, true}}
+	f := &Forward{maxfails: 2, tlsConfig: new(tls.Config), expire: defaultExpire, p: new(policy.Random), from: ".", hcInterval: hcInterval, opts: options{forceTCP: false, preferUDP: false, hcRecursionDesired: true}}
 	return f
 }
 
--- a/plugin/forward/health.go
+++ b/plugin/forward/health.go
@@ -14,10 +14,14 @@ import (
 type HealthChecker interface {
 	Check(*Proxy) error
 	SetTLSConfig(*tls.Config)
+	SetRecursionDesired(bool)
 }
 
 // dnsHc is a health checker for a DNS endpoint (DNS, and DoT).
-type dnsHc struct{ c *dns.Client }
+type dnsHc struct {
+	c                *dns.Client
+	recursionDesired bool
+}
 
 // NewHealthChecker returns a new HealthChecker based on transport.
 func NewHealthChecker(trans string) HealthChecker {
@@ -28,7 +32,7 @@ func NewHealthChecker(trans string) HealthChecker {
 		c.ReadTimeout = 1 * time.Second
 		c.WriteTimeout = 1 * time.Second
 
-		return &dnsHc{c: c}
+		return &dnsHc{c: c, recursionDesired: true}
 	}
 
 	log.Warningf("No healthchecker for transport %q", trans)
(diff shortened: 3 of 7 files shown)


miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] Make health-check..." ]
>&lt;!--
>Thank you for contributing to CoreDNS!
>Please provide the following information to help us make the most of your pull request:
>--&gt;
>
>The `forward`-plugin states in comments that it performs health-checks on upstream dns-servers by sending non-recursive `. IN NS` questions. Currently in master, it sends recursive requests.
>(https://github.com/coredns/coredns/blob/master/plugin/forward/health.go#L43)
>
>This may not be a problem in environments where dns is set up ideally. But on our premise, these recursive requests results in timeouts / 10 second late responses, which causes Coredns to think the upstreams are always unhealthy.
>
>We found this out when we took down 1 of 2 upstreams for maintainence, and we suddenly saw a high amount of failed dns-queries.

I've seen the opposite on my home router. Having recursion on is more standard than having
if off

What's your upstream running, because it sounds works than CPE

ctryti:
We're in an air-gapped environment. Our upstreams are Active Directory DNS, which we suspect it might not be configured ideally, and that it's asking for records out on the internet which gets blocked by our firewall.

Sorry, i'm unfamiliar with the term; what does CPE mean?

We tested the change in this PR on our kubernetes-cluster, and it works as we expect. 
So, if making the health-check do non-recursive queries default is a no-go, how about making it configurable?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Make health-c..." ]
>So, if making the health-check do non-recursive queries default is a no-go, how about making it configurable?

sadly, that is the only option, yes.

ctryti:
Updated title and PR-description after feedback from reviewer.

plugin/forward/README.md
```diff
@@ -52,6 +52,7 @@ forward FROM TO... {
     tls_servername NAME
     policy random|round_robin|sequential
     health_check DURATION
+    health_check_non_recursive
```
miekg: It's better/cleaner to make this an option property of `health_check`, probably something like
`healthcheck DURATION [no_rec]` or something

plugin/forward/forward.go
```diff
@@ -52,7 +52,7 @@ type Forward struct {
 
 // New returns a new Forward.
 func New() *Forward {
-	f := &Forward{maxfails: 2, tlsConfig: new(tls.Config), expire: defaultExpire, p: new(policy.Random), from: ".", hcInterval: hcInterval}
+	f := &Forward{maxfails: 2, tlsConfig: new(tls.Config), expire: defaultExpire, p: new(policy.Random), from: ".", hcInterval: hcInterval, opts: options{false, false, true}}
```
miekg: the options struct should use named paramaeters: `forceTCP: false` etc.

plugin/forward/health.go
```diff
@@ -14,10 +14,14 @@ import (
...
 
 // dnsHc is a health checker for a DNS endpoint (DNS, and DoT).
-type dnsHc struct{ c *dns.Client }
+type dnsHc struct {
+	c                *dns.Client
+	recursionDesired bool
+}
 
 // NewHealthChecker returns a new HealthChecker based on transport.
 func NewHealthChecker(trans string) HealthChecker {
```
miekg: this should probably receive a boolean or an `options` to set the recursion field

plugin/forward/health.go
```diff
@@ -40,7 +44,11 @@ func (h *dnsHc) SetTLSConfig(cfg *tls.Config) {
 	h.c.TLSConfig = cfg
 }
 
-// For HC we send to . IN NS +norec message to the upstream. Dial timeouts and empty
+func (h *dnsHc) SetRecursionDesired(recursionDesired bool) {
```
miekg, line 48: probably not needed.

plugin/forward/health_test.go
```diff
@@ -43,6 +43,37 @@ func TestHealth(t *testing.T) {
 	}
 }
 
+func TestHealthNoRecursion(t *testing.T) {
```
miekg: this doesn't test the functionality you've added. The only thing we care about is if the incoming query now has the recursion bit set or not

plugin/forward/health_test.go
```diff
@@ -83,6 +114,47 @@ func TestHealthTimeout(t *testing.T) {
 	}
 }
 
+func TestHealthTimeoutNoRecursion(t *testing.T) {
```
miekg: this is duplicating other tests

plugin/forward/setup_test.go
```diff
@@ -21,21 +21,24 @@ func TestSetup(t *testing.T) {
...
-		{"forward . 127.0.0.1 {\nexcept miek.nl\n}\n", false, ".", nil, 2, options{}, ""},
-		{"forward . 127.0.0.1 {\nmax_fails 3\n}\n", false, ".", nil, 3, options{}, ""},
-		{"forward . 127.0.0.1 {\nforce_tcp\n}\n", false, ".", nil, 2, options{forceTCP: true}, ""},
-		{"forward . 127.0.0.1 {\nprefer_udp\n}\n", false, ".", nil, 2, options{preferUDP: true}, ""},
-		{"forward . 127.0.0.1 {\nforce_tcp\nprefer_udp\n}\n", false, ".", nil, 2, options{preferUDP: true, forceTCP: true}, ""},
-		{"forward . 127.0.0.1:53", false, ".", nil, 2, options{}, ""},
-		{"forward . 127.0.0.1:8080", false, ".", nil, 2, options{}, ""},
-		{"forward . [::1]:53", false, ".", nil, 2, options{}, ""},
-		{"forward . [2003::1]:53", false, ".", nil, 2, options{}, ""},
+		{"forward . 127.0.0.1", false, ".", nil, 2, options{hcRecursionDesired: true}, ""},
```
miekg, line 24: please add a single and small, contained test to test for this new property

plugin/forward/health_test.go
ctryti: Redid both tests i added. Now checks the RD-flag in the dns-requests.

plugin/forward/health.go
ctryti: Removed

plugin/forward/health.go
ctryti: I added a bool to HealthCheckers signature. Are you ok with this being set to true default in NewProxy()? Or should I also edit the signature of NewProxy() to include this config?

plugin/forward/forward.go
ctryti: Done

plugin/forward/README.md
ctryti: Done. `health_check` now supports an optional `no_rec` argument.

plugin/forward/health_test.go
ctryti: Redid both tests i added. Now checks the RD-flag in the dns-requests.

plugin/forward/setup_test.go
ctryti: Added a new test testing the health_check properties.

plugin/forward/README.md
```diff
@@ -85,7 +85,10 @@ forward FROM TO... {
   * `random` is a policy that implements random upstream selection.
   * `round_robin` is a policy that selects hosts based on round robin ordering.
   * `sequential` is a policy that selects hosts based on sequential ordering.
-* `health_check`, use a different **DURATION** for health checking, the default duration is 0.5s.
+* `health_check` configure the behaviour of health checking of the upstream servers
```
miekg, line 88: Please keep the current text and just add your new property in a new sentence after it, i.e. "The optional `no_rec` argument...`

plugin/forward/health.go
miekg: either github is not working, or you have pushed your code?

plugin/forward/health.go
```diff
@@ -40,7 +45,14 @@ func (h *dnsHc) SetTLSConfig(cfg *tls.Config) {
...
+	h.recursionDesired = recursionDesired
+}
+func (h *dnsHc) GetRecursionDesired() bool {
```
miekg, line 51: Drop Get from this - do we really need it btw?

plugin/forward/health_test.go
```diff
@@ -14,10 +14,15 @@ import (
 )
 
 func TestHealth(t *testing.T) {
```
miekg, line 16: don't change this function for this PR

plugin/forward/health_test.go
```diff
@@ -39,7 +44,43 @@ func TestHealth(t *testing.T) {
...
+func TestHealthNoRecursion(t *testing.T) {
+	const expected = 1
+	i := uint32(0)
+	q := uint32(0)
+	s := dnstest.NewServer(func(w dns.ResponseWriter, r *dns.Msg) {
+		if atomic.LoadUint32(&q) == 0 { //drop the first query to trigger health-checking
+			atomic.AddUint32(&q, 1)
+			return
+		}
+		if r.Question[0].Name == "." && r.RecursionDesired == false {
```
miekg, line 60: Do a t.Fatal here if recursion is set

plugin/forward/health_test.go
```diff
@@ -39,7 +44,43 @@ func TestHealth(t *testing.T) {
 	time.Sleep(1 * time.Second)
 	i1 := atomic.LoadUint32(&i)
 	if i1 != expected {
-		t.Errorf("Expected number of health checks to be %d, got %d", expected, i1)
+		t.Errorf("Expected number of health checks with RecursionDesired==true to be %d, got %d", expected, i1)
+	}
+}
+
+func TestHealthNoRecursion(t *testing.T) {
+	const expected = 1
```
miekg, line 52: We only care that recursion is unset. The other tests takes care of counting. Please minimize what you're trying to test here.

plugin/forward/setup_test.go
miekg: have you pushed your updated code?

ctryti: Yeah, it starts at line 215

plugin/forward/health_test.go
ctryti: I can change the `expected` to a bool, but i still have to count. At least fail one query for the healthchecker to kick in.

plugin/forward/health.go
ctryti: It does feel a bity java'y. But i don't know how else to get the property out of the dnsHc-instance. Any tips?

plugin/forward/health.go
ctryti: Sorry, I thought i removed this function, but it was a similar one from another file. I'm not sure how else to bring the boolean from the setup-function over to the dnsHc without changing a couple of function-signatures.

miekg: yeah, looking back at this code - it looks a bit over-engineered and needs some drastic simplications.

miekg (approved):
as said in comment, we probably want to revisit this entire setup, looking back feel super generic for no good reasons.

PR #3683, opened 2020-02-18
Fix plugin name in project README 

chrisohaver:

Just noticed this on the project README.  The outward name of the metrics plugin is "prometheus" not "metrics".

none

none

no

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Currently CoreDNS is able to:
 * Use etcd as a backend (replacing [SkyDNS](https://github.com/skynetservices/skydns)) (*etcd*).
 * Use k8s (kubernetes) as a backend (*kubernetes*).
 * Serve as a proxy to forward queries to some other (recursive) nameserver (*forward*).
-* Provide metrics (by using Prometheus) (*metrics*).
+* Provide metrics (by using Prometheus) (*prometheus*).
 * Provide query (*log*) and error (*errors*) logging.
 * Integrate with cloud providers (*route53*).
 * Support the CH class: `version.bind` and friends (*chaos*).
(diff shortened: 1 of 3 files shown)


miekg:
this needs a search and replace in all readme's as well, and plugin.md and other docs
might need updating as well, as we call it (iirc) metrics plugin in a lot of places

yongtang:
The Corefile still uses `prometheus` as the keyword. Maybe this could be changed to metrics as well? (I can do that in a separate PR if this makes sense).

chrisohaver:
> The Corefile still uses `prometheus` as the keyword. Maybe this could be changed to metrics as well? (I can do that in a separate PR if this makes sense).

Do you mean to change the name of the plugin in the Corefile to `metrics`?  That would be backward incompatible, affecting too many users, and not really providing any benefit.

yongtang:
> Do you mean to change the name of the plugin in the Corefile to metrics? That would be backward incompatible, affecting too many users, and not really providing any benefit.

We could create a transition path for that. What I am concerned is that the keyword discrepancy `metrics` vs `prometheus`, which might be confusing to many users.

Last time I was playing with metrics plugin, I naturally thought of `metrics` but only saw:
```
%: ./coredns 
Corefile:2 - Error during parsing: Unknown directive 'metrics'
```

chrisohaver:
Most users who use CoreDNS are not diving into the code, so they would not be confused by the package name plugin name mismatch.  It's only an annoyance to those diving into the code, or who already have a familiarity of the code.

To address that confusion for _contributors_, we could consider changing the directory and package name to `prometheus`.  This would pose no _user_ facing changes.  IMO we should keep this kind of clean up from surfacing to users as much as possible.

chrisohaver:
> this needs a search and replace in all readme's as well, and plugin.md and other docs might need updating as well, as we call it (iirc) metrics plugin in a lot of places

`plugin.md` looks OK to me as is.  However, both cache and reload refer to the "metrics plugin", should be made to match the user facing name.  I didn't find anything else.

PR #3685, opened 2020-02-18
metrics: add type to latency as well

### LINKED ISSUE #3684: separate latency metrics A/AAAA

I would like to be able to separate metrics for A and AAAA records.

**Why is this needed**:
I investigate an issue, that requires understanding of response latencies per record type.

miekg:
Add the record type to the latency metric as well.

This effectively removes the need for the _total because the histogram
contains the same information, this should probably be included in #3671

Fixes: #3684

### DIFF
--- a/plugin/metrics/README.md
+++ b/plugin/metrics/README.md
@@ -13,7 +13,7 @@ The following metrics are exported:
 * `coredns_build_info{version, revision, goversion}` - info about CoreDNS itself.
 * `coredns_panic_count_total{}` - total number of panics.
 * `coredns_dns_request_count_total{server, zone, proto, family}` - total query count.
-* `coredns_dns_request_duration_seconds{server, zone}` - duration to process each query.
+* `coredns_dns_request_duration_seconds{server, zone, type}` - duration to process each query.
 * `coredns_dns_request_size_bytes{server, zone, proto}` - size of the request in bytes.
 * `coredns_dns_request_do_count_total{server, zone}` -  queries that have the DO bit set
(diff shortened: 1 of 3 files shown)


chrisohaver (approved):

SuperQ:
The requests total metric also includes `proto` and `family`. Not sure if we want to add that granularity to the histogram.

I'm also thinking about if we can come up with a better name than "type" for this label. It tends to be a bit generic. Maybe `query` or `record`?

SuperQ (approved):
LGTM, since this is additive compared to #3671, I don't see a reason to hold off on this. Unless we really want to keep full metric stability between minor releases.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] metrics: add ..." ]
>The requests total metric also includes `proto` and `family`. Not sure if we want to add that granularity to the histogram.

yeah, that's probably why I never added in the first place. It's 10x more in cardinality,
which is not disastrous.

>I'm also thinking about if we can come up with a better name than "type" for this label. It tends to be a bit generic. Maybe `query` or `record`?

'type' is what this is called. You can make the case for 'qtype', but that's another
annoying metrics migration.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] metrics: add ..." ]
>SuperQ approved this pull request.
>
>LGTM, since this is additive compared to #3671, I don't see a reason to hold off on this. Unless we really want to keep full metric stability between minor releses.

adding labels should be fine

szuecs:
@miekg Thanks so much !

Issue #3695, opened 2020-02-25
plugin/cache: Optional Cache with Non-Blocking Adds, Admission Policy, and LFU Eviction


huntharo:
# What would you like to be added

- Option to use a cache with the following features:
  - Non-blocking adds to the cache (not making the requester of the non-cached item wait for the write lock to be acquired)
  - Key collision protection (ensure that a key collision will not return an answer to the wrong question)
    - Moved to: #3698
  - Admission policy (ensure that single-request for an item does not push other items out of the cache)
  - LFU eviction instead of random (keep frequently and recently requested items in the cache)
  - TTL eviction option, `ristretto ttlEvict`, will evict expired items from the cache, when they expire (or when max serve stale is reached if enabled)
- Future
  - Eventually: make the new cache the default option if it is an improvement for most

# Why is this needed

- Possibly related issues
  - #3388 - Restart due to OOM
- The internal cache is hashing using FNV but is not saving anything in the cache to allow detection of collision on *reads*
  - It appears to me, that the cache can return an Answer for a non-matching Question
  - The Question is not stored at all in the cache and the key is a unit64 FNV hash, so it appears to be impossible to detect that asking the cache for `microsoft.com` is returning an answer that was retrieved for `someevilhost.com`
  - Returning Answers to the wrong Question a rather serious flaw to be allowed in a DNS server, but perhaps there are mitigating circumstances that I missed
- Eviction in the internal cache is random and all new items are admitted
  - This is noted in the docs and is not unexpected, but it is not desired
  - Eviction of heavily accessed items should be avoided (LRU should be used instead of random for eviction)
  - Admission to the cache should not be automatic when the cache is full
     - Only items that are requested with some frequency should be considered for admission into the cache
     - Implementation of this is non-trivial
- Writing to the internal cache makes the caller wait for the acquisition of the write lock
  - I may be mistaken, but it appears that neither `.set` nor `.Add` are called in a new Go Routine
  - The caller must wait for the write lock to be obtained for these functions to return
  - This effectively serializes all writes within a shard - This is primarily a problem when there are high numbers of unique records to be cached
  - This *may* be leading to the large number of concurrent requests and OOM conditions being reported by some
  - Acquiring the write lock to write a single item, then releasing it, is expensive, particularly if there are already other items that waiting to be added to the same shard
    - Other cache libraries return to the caller immediately and obtain the write lock to write a batch of records at once

# Benchmarks

[...]

PR #3696, opened 2020-02-25
plugin/cache: Add optional ristretto cache

huntharo:
## 1. Why is this pull request needed and what does it do?

- Option to use a cache with the following features:
  - Non-blocking adds to the cache (not making the requester of the non-cached item wait for the write lock to be acquired)
  - Key collision protection (ensure that a key collision will not return an answer to the wrong question)
  - Admission policy (ensure that single-request for an item does not push other items out of the cache)
  - LFU eviction instead of random (keep frequently and recently requested items in the cache)
- Future
  - Enable TTL eviction in the new cache
  - Eventually: make the new cache the default option if it is an improvement for most

#3695 

- DONE: Move references to specific behaviors of the internal cache module to sections that apply only to that particular cache module (e.g. discussion of eviction being random)
- DONE: Document the new `ristretto` cache config option

- Not at the moment - The new cache is added as a non-default option

### 5. To-Do

- [ ] Review and confirm metrics are still at least somewhat compatible with prior values
- [ ] Determine why the existing builtin cache got faster between the baseline and this PR
- [x] Review the Ristretto default hash and replace if slower than FNV64
   - Ristretto was using MemHash + xxHash for a 128 bit hash by default
   - xxHash alone was adding ~200 ns / op
   - Passed Ristretto.KeyToHash to revert to FNV64 hashes - nearly restored read performance to baseline


.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
SuperQ: Please drop this file from the PR. We don't need `.vscode` stuff in the codebase.

huntharo: Thanks for noticing.  I'm usually super clean about that... but this once slipped by.

miekg:
closing here, external plugin (for now) looks like the best way forward. See coredns/coredns.io how to get that listed on coredns.io's website.

Issue #3701, opened 2020-02-27
plugin/cache: Expired denial cache item obscures success cache item - repeated hits to backend / miscount of misses


huntharo:
Note: PR #3702 has the failing test case to demonstrate

**What happened**:

- A name that did not exist was requested
- The name was retrieved and entered into the denial cache
- The name was created and re-requested after the TTL on the denial cache item expired
- The name was retrieved and entered into the success cache
- Subsequent requests for the name continued to hit the backend (bypassing the cache)
- In this case the Cache Miss stat was undercounted as well:
  - No Hit was counted as the item was determined to be too stale to serve in `getIgnoreTTL` - This was correct
  - No Miss was counted as the return happened from within the negative cache block and misses are only counted if both a negative and positive item do not exist (regardless of TTL) - This was incorrect as a backend hit would be made
  - The proposed patch causes a jump in the number of Misses but it's a reporting bug fix not a change in the actual misses
  - Perhaps this logic should move up a level as it's much easier to count a miss when you know you're calling `NextOrFailure`

**What you expected to happen**:

- The 2nd+ requests should have returned from the success cache instead of hitting the backend

**How to reproduce it (as minimally and precisely as possible)**:

PR with failing test case incoming.

To reproduce in the real world:

- Enable postive / negative cache
- Set negative cache max TTL to, say, 10 seconds
- Query a name that responds with an NXDOMAIN
- Create a record for the queried name
  - This will return NOERROR
  - Item will be added to positive cache
- Change the record and query it again
  - For example, change the A answer
  - OR: just query it again if the upstream is far enough away that you can tell from response time that the item was not returned from cache
  - Backend will be hit, returning either updated value or having response time indicating backend was hit

**Anything else we need to know?**:

The problem is not due to any one thing in isolation.

The problem is instead due to an entire set of things; changing any one of these things can be enough to resolve the problem:

- CoreDNS does not call .Remove on the cache to remove any items
  - Note: considering the `plugin/pkg/cache` as external to CoreDNS itself
  - Items that are retrieved from the cache and found to be expired are simply left in the cache
  - Eviction is random, which means items will not age out quickly after they stop getting accessed in the negative cache
  - Code where an expired item *could* be removed: https://github.com/coredns/coredns/blob/6910c9a116d55b7910975bd4d153b7fe77822c8b/plugin/cache/handler.go#L35-L36
- `cache.ServeDNS` uses `Handler.getIgnoreTTL`
[...]

PR #3702, opened 2020-02-27
plugin/cache: Add a test for denial cache masking the success cache

huntharo:

- New failing test case for a bug as described in #3701 

```
go test
--- FAIL: TestNegativeExpiredMaskingPositiveCache (0.00s)
    cache_test.go:371: Test 3 NOERROR from Cache: expecting 0; got 255
FAIL
exit status 1
FAIL    github.com/coredns/coredns/plugin/cache 0.033s
```

#3701 

None.

None.

### DIFF
--- a/plugin/cache/cache_test.go
+++ b/plugin/cache/cache_test.go
@@ -202,7 +202,7 @@ func TestCache(t *testing.T) {
 			crr.set(m, k, mt, c.pttl)
 		}
 
-		i, _ := c.get(time.Now().UTC(), state, "dns://:53")
+		i := c.getIgnoreTTL(time.Now().UTC(), state, "dns://:53")
 		ok := i != nil
 
 		if ok != tc.shouldCache {
--- a/plugin/cache/cache_test.go
+++ b/plugin/cache/cache_test.go
@@ -382,6 +382,82 @@ func TestNegativeStaleMaskingPositiveCache(t *testing.T) {
 	}
 }
 
+func TestNegativeExpiredMaskingPositiveCache(t *testing.T) {
+	c := New()
+	c.Next = nxDomainBackend(60)
+
+	req := new(dns.Msg)
+	qname := "cached.org."
+	req.SetQuestion(qname, dns.TypeA)
+	ctx := context.TODO()
+
+	// Negative Cache example.org. - NXDOMAIN
+	expectedResult := dns.RcodeNameError
+	if ret, _ := c.ServeDNS(ctx, &test.ResponseWriter{}, req); ret != expectedResult {
+		t.Errorf("Test 0 Negative Cache Population: expecting %v; got %v", expectedResult, ret)
+	}
+
+	// Confirm item was added to negative cache and not to positive cache
+	if c.ncache.Len() == 0 {
+		t.Errorf("Test 0 Negative Cache Population: item not added to negative cache")
+	}
+	if c.pcache.Len() != 0 {
+		t.Errorf("Test 0 Negative Cache Population: item added to positive cache")
+	}
+
+	// Backend - Give error only
+	c.Next = plugin.HandlerFunc(func(context.Context, dns.ResponseWriter, *dns.Msg) (int, error) {
+		return 255, nil // Below, a 255 means we tried querying upstream.
+	})
+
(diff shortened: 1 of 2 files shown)


huntharo:
Possible fix below.  I'm really not a fan of this fix.  I think it really exemplifies why we should not allow the cache to store both a positive and a negative record.  It seems that if adding a key to one cache we should be removing it from the other cache.

All tests pass and the `get()` function can be removed.

- Drop this patch in here:
  - https://github.com/coredns/coredns/blob/6910c9a116d55b7910975bd4d153b7fe77822c8b/plugin/cache/handler.go#L99-L119
- Delete the get() function:
  - https://github.com/coredns/coredns/blob/6910c9a116d55b7910975bd4d153b7fe77822c8b/plugin/cache/handler.go#L83-L97
- Replace the `get()` call with a `getIgnoreTTL()` call - Remove the unused second return param
  - https://github.com/coredns/coredns/blob/6910c9a116d55b7910975bd4d153b7fe77822c8b/plugin/cache/cache_test.go#L205

```
// getIgnoreTTL will return the best unexpired or expired item
func (c *Cache) getIgnoreTTL(now time.Time, state request.Request, server string) *item {
	k := hash(state.Name(), state.QType(), state.Do())

	var ni, pi interface{}
	nexists, pexists := false, false
	nwillserve, pwillserve := false, false
	pexpired := false
	nttl, pttl := 0, 0

	// See if the negative cache has the item
	ni, nexists = c.ncache.Get(k)
	if nexists {
		nttl = ni.(*item).ttl(now)
		if nttl > 0 || (c.staleUpTo > 0 && -nttl < int(c.staleUpTo.Seconds())) {
			nwillserve = true
		}
	}

	// See if the positive cache has the item
	pi, pexists = c.pcache.Get(k)
	if pexists {
		pttl = pi.(*item).ttl(now)
		if pttl > 0 || (c.staleUpTo > 0 && -pttl < int(c.staleUpTo.Seconds())) {
			pwillserve = true
		}
		if pttl <= 0 {
			pexpired = true
		}
	}

	if pexists && (!pexpired || !nexists || (nexists && pttl >= nttl)) {
		// Return the positive item if:
		// it exists and is not expired
		// OR it exists and the negative item does not exist
		// OR it exists and the negative item exists but the
		//   positive has more TTL (note: both may be expired or non-expired)
		if pwillserve {
			cacheHits.WithLabelValues(server, Success).Inc()
		} else {
			cacheMisses.WithLabelValues(server).Inc()
		}
		return pi.(*item)
	} else if nexists {
		// Return the negative item if it exists (and we didn't return positive item already)
		if nwillserve {
			cacheHits.WithLabelValues(server, Denial).Inc()
		} else {
			cacheMisses.WithLabelValues(server).Inc()
		}
		return ni.(*item)
	} else {
		cacheMisses.WithLabelValues(server).Inc()
		return nil
	}
}
```

chrisohaver:
With #3744 merged, the travis tests _should_ pass once this PR is rebased with master and conflicts resolved.

huntharo:
@chrisohaver  - I started working on rebasing this... but it looks like the tests were already merged to master.

https://github.com/coredns/coredns/blob/021f761291f8d8cf6e392a3b7ae2d924530d7083/plugin/cache/cache_test.go#L299-L383

chrisohaver:
> but it looks like the tests were already merged to master.

`TestNegativeStaleMaskingPositiveCache()` is a different test, based on yours but not the same.  It testing a related but distinct case for _stale_ cache retrieval.

IOW, your test still has value IMO.

huntharo:
@chrisohaver - Ok.  Let me know if you want me to resubmit my test as a PR.  I can do it but if it's going to be rewritten and submitted by someone else then let's just short circuit that.

chrisohaver:
Can you reopen this PR?

`TestNegativeStaleMaskingPositiveCache()` tests a different case than `TestNegativeExpiredMaskingPositiveCache()`.

`TestNegativeStaleMaskingPositiveCache()` tests the case for stale cache retrieval, which has different set up steps and different expected values, it is not a re-written version of `TestNegativeExpiredMaskingPositiveCache()`.  They are not testing the same bug.

Issue #3703, opened 2020-02-28
plugin/cache: Metrics reporting on Inserts Responsible for 17-65% of ns/op


huntharo:
Please advise if you'd like me to open a PR with the benchmarks and a fix for this or if you'd prefer to wait for my other PRs to land (or not) first.

**What would you like to be added**:

- I have a branch read with just these additional benchmarks added
- Commenting out these two lines causes a 17-65% reduction in ns/op when performing inserts with high contention
  - https://github.com/coredns/coredns/blob/6910c9a116d55b7910975bd4d153b7fe77822c8b/plugin/cache/cache.go#L178-L179
- The metrics feature does not need to be removed
  - This can be changed to a go routine that updates these two stats on an interval rather than on every single insert
  - The functionality of the metrics would be preserved and performance under heavy load (particularly under startup or reload as a lot of cache inserts would happen) would be improved

Benchmarks are available here:
https://github.com/huntharo/coredns/tree/cache-benchmark

**Why is this needed**:

- Simple performance improvement that increases the scalability of the application while preserving all functionality (increases the max load that can be served on a single machine running a single process using multiple processors)

**Delta Between Metrics Enabled/Disabled on Inserts**

- BenchmarkBuiltinBalanced-8
  - Enabled: 4,256 ns / op
  - Disabled: 3,552 ns / op
  - Delta: -704 ns / op = 17% reduction
- BenchmarkBuiltinInserts-8
  - Enabled: 3,568 ns / op
  - Disabled: 2,686 ns / op
  - Delta: -882 ns / op = 25% reduction
- BenchmarkBuiltinParallelInserts-8 
  - Enabled: 7,640 ns / op
  - Disabled: 2,698 ns / op
  - Delta: -4,942 ns / op = 65% reduction

**Metrics Enabled on Inserts**

```
BenchmarkBuiltinReads-8                           712930              1547 ns/op             566 B/op         14 allocs/op
BenchmarkBuiltinParallelReads-8                  2479280               487 ns/op             566 B/op         14 allocs/op
BenchmarkBuiltinBalanced-8                        262920              4256 ns/op             815 B/op         19 allocs/op
BenchmarkBuiltinInserts-8                         323592              3568 ns/op             706 B/op         16 allocs/op
BenchmarkBuiltinParallelInserts-8                 131554              7640 ns/op            2472 B/op         58 allocs/op
[...]

PR #3705, opened 2020-02-28
Cache reduce insert metrics cost

huntharo:

Reduces lock contention on cache inserts.  Each insert would need to obtain a ReadLock on *all* of the Shards, one at a time, on the builtin cache to compute the overall positive *and* negative cache sizes (regardless of which one was actually updated).  The cache is pretty carefully implemented to only need to lock one shard, but the metrics update caused every write to need a lock on every shard, which partially defeats the purpose of using shards.

#3703 

None.

None.

#### Baseline - Metrics Updated After Every Add

```
BenchmarkBuiltinReads-8                           712930              1547 ns/op             566 B/op         14 allocs/op
BenchmarkBuiltinParallelReads-8                  2479280               487 ns/op             566 B/op         14 allocs/op
BenchmarkBuiltinBalanced-8                        262920              4256 ns/op             815 B/op         19 allocs/op
BenchmarkBuiltinInserts-8                         323592              3568 ns/op             706 B/op         16 allocs/op
BenchmarkBuiltinParallelInserts-8                 131554              7640 ns/op            2472 B/op         58 allocs/op
BenchmarkBuiltinParallelInsertsRead-8             193125             29294 ns/op            1255 B/op         30 allocs/op
```

#### Metrics Updated in GoRoutine

```
BenchmarkBuiltinReads-8                           751372              1552 ns/op             566 B/op         14 allocs/op
BenchmarkBuiltinParallelReads-8                  2354401               507 ns/op             566 B/op         14 allocs/op
BenchmarkBuiltinBalanced-8                        286875              4217 ns/op             790 B/op         18 allocs/op
BenchmarkBuiltinInserts-8                         443962              2617 ns/op             688 B/op         16 allocs/op
BenchmarkBuiltinParallelInserts-8                 414012              2693 ns/op            1804 B/op         40 allocs/op
BenchmarkBuiltinParallelInsertsRead-8             200334             28704 ns/op            1247 B/op         30 allocs/op
```

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -38,7 +38,8 @@ type Cache struct {
 
 	staleUpTo time.Duration
 
-	mMgr *MetricsManager
+	mMgr            *MetricsManager
+	metricsInterval time.Duration
 
 	// Testing.
 	now func() time.Time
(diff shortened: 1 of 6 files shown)


SuperQ:
This doesn't seem to be an improvement, and violates the way metrics are designed to be managed. There should be no internal timer loops for updating metrics, as it violates best practices for instrumentation.

Issue #3658, opened 2020-02-06
CoreDNS Projects for Summer of Code 2020


yongtang:
Please Note: This is a tracking issue for [Summer of Code](https://github.com/cncf/soc). Anyone interested in this implementation should check [link](https://github.com/cncf/soc) there.

Please Note: This is reserved for Summer of Code, we don't accept PRs outside of Summer of Code, or without coordination with maintainers.

Below are the list of candidate CoreDNS projects for Summer of Code 2020:

**External health check and orchestration of CoreDNS in Kubernetes clusters**

- Description: CoreDNS is the cluster DNS server for Kubernetes and is very much critical for the overall health of the Kubernetes cluster. It is very important to monitoring the health of CoreDNS itself and restarting or repairing any CoreDNS pods that are not behaving correctly. While CoreDNS exposes a health check itself, the health check is not UDP (DNS) based. The existing health check is also launched locally which could be very much different when accessed by other pods remotely.
  This project intends to build an application that checks CoreDNS health through UDP (DNS) externally, and, restart CoreDNS pods by interacting with Kubernetes API through golang. This is an important project that will greatly improve the overall health of Kubernetes clusters through automation.
- Recommended Skills:  Go, DNS, Kubernetes
- Mentor(s): Yong Tang (@yongtang)
- Implementation: The deliverable of this project is a golang program that could be deployed in a Kubernetes cluster independently while at the same time, monitoring CoreDNS pods in the same cluster and interacting Kubernetes API (server) to restart CoreDNS pods as needed.
- Reference: Please see #3617 for some related discussions.

**Anomaly detection of CoreDNS server through machine learning**

- Description: As a DNS server, CoreDNS is critical to overall devops infrastructure. Any anomaly related to CoreDNS server should be taken seriously. While altering rules (combined with monitoring tools such as Prometheus) helps in discovering issues, those rules are often crafted manually and requires human expertise. It would help a lot if machine learning could be utilized to further automate the monitoring/alerting in case of anomaly.
  This project intends to build and train a model that could be used for anomaly detection of CoreDNS server through metrics collected from Prometheus. Since the metrics pipeline to Prometheus is already available in CoreDNS, the project’s focus is mostly on model building. It is suggested to use tf.keras to build the model. A successful model should at least be able to detect a scenario that is alerting and requires further devops or security intervention.  
- Recommended Skills:  DNS/CoreDNS, Prometheus, Keras/TensorFlow, Python
[...]

PR #3715, opened 2020-03-02
Added GSoC Proposal

wjayesh:
I've added my GSoC proposal to a folder which can be used to store other proposals too.

#3658 
None
No


miekg:
you're doing the PR in the wrong repo - not sure what the right repo is atm, but not coredns/coredns.

PR #3716, opened 2020-03-02
Bumb travis Go version to 1.14

miekg:
Start using latest Go for travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ services:
   - docker
 language: go
 go:
-  - "1.13.x"
+  - "1.14.x"
 
 cache:
   directories:


SuperQ:
I might hold off on this.

See: https://github.com/golang/go/issues/37436

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Bumb travis G..." ]
>I might hold off on this.
>
>See: https://github.com/golang/go/issues/37436

that's CGO no?

Note, our release script (https://github.com/coredns/release/blob/master/release-coredns)
grabs latest Go by default.

SuperQ:
No, it seems like it's any Go binary.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Bumb travis G..." ]
>No, it seems like it's any Go binary.

Meh... seems the whatever is happening, is also happening in 1.13 except it results in a
crash. But we can wait for 1.14.1 or so to pop up; would be good to get this merged, then
we upgrade automatically to newer versions anyway, and, as said, we would release with
1.14 by default (so we might want to hold off doing a release as well)

karalabe:
> Meh... seems the whatever is happening, is also happening in 1.13 except it results in a
crash.

The problem is that detecting the bad kernel is broken, so it will crash on patched kernels too (pretty much any recent Ubuntu and many other distros).

miekg:
oooooh, I was planning to do a release today, but then I remembered. Better hold off until this is fixed. I'll file a release tracker bug

karalabe:
It's not really clear when/how this will get fixed. Probably a good idea for now is to stick with Go 1.13 until there's a way forward and not wait for something that might not come (the upstream bug is tagged for `1.15` currently).

miekg:
if the go team has done that, than that signals to me this isn't a blocker

karalabe:
I guess they just forgot to swap out the milestone on github :) https://groups.google.com/forum/#!msg/golang-dev/_FbRwBmfHOg/mmtMSjO1AQAJ

Edit: Got this link after I wrote above.

nberlee:
In case of CoreDNS it is the drop all capabilities that makes the mlock test fail. Adding capability `IPC_LOCK` makes it work when compiled under 1.14, having that said, I hope golang will change this test....

SuperQ:
Looks like the issue will have fixes/documentation in 1.14.1.

miekg:
go 1.14.1 has been released...

stp-ip:
Informative sidenote:

The Kubernetes community will be switching the promotion process for images in k8s.gcr.io.
The final flip is scheduled in the coming weeks. I have access to do promotions and it comes down to pushing to a staging repo and opening a specific PR for adding versions + hashes (https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/images/k8s-staging-coredns/images.yaml).
Others are aware of the process as well such as @johnbelamaric @rajansandeep.

Didn't promote the last releases as the process wasn't live. Will promote all releases once this one goes out.

Images then are available under k8s.gcr.io/coredns/coredns:vX.X.X in addition to our own images on dockerhub. Also have an issue in my backlog to automate the staging image population with a build repo (https://github.com/kubernetes/k8s.io/issues/533). Aka build CoreDNS from source, auto push to staging and then we (mostly I) only need to open the corresponding PR.

stp-ip:
Wrong tab. My bad. Removed comment ;)

miekg:
Read through the entire go issue (which is now closed), I'm going to merge this.

This is neat as well: https://github.com/golang/go/wiki/LinuxKernelSignalVectorBug

PR #3729, opened 2020-03-06
Add CI test for man page updates

SuperQ:

Run make -f Makefile.doc in CircleCI to test that man pages are updated
in PRs.

None

None

None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,6 +40,14 @@ buildCoreDNSImage: &buildCoreDNSImage
         kind load docker-image coredns
 
 jobs:
+  doc-test:
+    docker:
+      - image: circleci/golang:1.13
+    steps:
+      - checkout
+      - run: ./fixup_file_mtime.sh
+      - run: make -f Makefile.doc
+      - run: git diff --exit-code -- man/
   kubernetes-tests:
     <<: *integrationDefaults
     steps:
@@ -62,4 +70,5 @@ workflows:
   version: 2
   integration-tests:
     jobs:
+      - doc-test
       - kubernetes-tests
--- a/Makefile.doc
+++ b/Makefile.doc
@@ -3,20 +3,35 @@
 # updated before doing a release. The Debian package, for instance, looks at these pages
 # and will install them on your system.
 
-MMARK:=mmark -man
+MMARK_VERSION:=2.2.4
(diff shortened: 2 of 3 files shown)


miekg:
ooh, I like! #lessToil

SuperQ:
I have no idea why, but go get is downloading an old version of mmark. :confused:

.circleci/config.yml
```diff
@@ -40,6 +40,15 @@ buildCoreDNSImage: &buildCoreDNSImage
         kind load docker-image coredns
 
 jobs:
+  doc-test:
+    docker:
+      - image: circleci/golang:1.13
+    steps:
+      - checkout
+      - run: go get -u -v github.com/mmarkdown/mmark
```
miekg: `@latest` ?

SuperQ: No, it looks like it's a Go issue. Because it's doing `go get` from within the CoreDNS source, it's trying to minimize version dependency.

I guess we can do a binary fetch of the build tar file.

SuperQ:
Ok, the test works, but it now thinks that all of the files need a timestamp update. This might be with how CircleCI checks out the codebase.

miekg:
circle-ci probably has some kind of fixed time, just like Go playground?

SuperQ:
No, I think it's because it tries to fetch the code in a shallow way, to avoid having to clone the full repo history. This may not set the timestamps of the files to the same as what they are in the real codebase.

miekg:
oh, but that's OK, they get bumped monthy

SuperQ:
There, I added a hack that fixes the timestamps based on the git log.

Should that be part of the Makefile?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add CI test f..." ]
>There, I added a hack that fixes the timestamps based on the git log.
>
>Should that be part of the Makefile?

yes, that is much nicer

miekg:
is this merge ready?

SuperQ:
I realized that putting the mtime fixer into the makefile isn't going to work, because it's keying on the mtime of those files.

I moved the mtime fixer into a little shell script, so that it's easier for people to run outside of CI.

This is now ready for merge.

miekg (approved):

PR #3732, opened 2020-03-06
plugin/file+auto: implement transfer.Transferer

chrisohaver:

Implements zone transfer using the _transfer_ plugin.

Could mention it in README, but we should not deprecate existing functionality until Notify PR and  auto/file notify implementations merge.  Without notify, it could be confusing/esoteric to document in this state.

no

### DIFF
--- a/plugin/auto/transfer.go
+++ b/plugin/auto/transfer.go
@@ -5,7 +5,7 @@ import (
 	"github.com/coredns/coredns/plugin/transfer"
 	"github.com/miekg/dns"
 )
-
+// Transfer implements Transfer.Transferer
 func (a Auto) Transfer(zone string, serial uint32) (<-chan []dns.RR, error) {
 	// look for exact zone match
 	var z *file.Zone
--- a/plugin/auto/transfer.go
+++ b/plugin/auto/transfer.go
@@ -1,0 +1,22 @@
+package auto
+
+import (
+	"github.com/coredns/coredns/plugin/file"
+	"github.com/coredns/coredns/plugin/transfer"
+	"github.com/miekg/dns"
+)
+
+func (a Auto) Transfer(zone string, serial uint32) (<-chan []dns.RR, error) {
+	// look for exact zone match
+	var z *file.Zone
+	for fz, zo := range a.Z {
+		if zone == fz {
+			z = zo
+			break
+		}
+	}
+	if z == nil {
+		return nil, transfer.ErrNotAuthoritative
+	}
+	return z.Transfer(serial)
+}
--- a/plugin/file/transfer.go
+++ b/plugin/file/transfer.go
@@ -1,0 +1,52 @@
+package file
+
+import (
+	"github.com/coredns/coredns/plugin/file/tree"
+	"github.com/coredns/coredns/plugin/transfer"
+	"github.com/miekg/dns"
+)
+
+func (f File) Transfer(zone string, serial uint32) (<-chan []dns.RR, error) {
+	// look for exact zone match
+	var z *Zone
+	for fz, zo := range f.Z {
+		if zone == fz {
+			z = zo
+			break
+		}
+	}
+	if z == nil {
+		return nil, transfer.ErrNotAuthoritative
+	}
+	return z.Transfer(serial)
+}
+
+func (z Zone) Transfer(serial uint32) (<-chan []dns.RR, error) {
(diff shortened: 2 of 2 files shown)


stp-ip (approved):
Looks good and works as expected. Love the transfer plugin idea. Much cleaner flow.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/file+a..." ]
>@stp-ip approved this pull request.
>
>Looks good and works as expected. Love the transfer plugin idea. Much cleaner
>flow.

good! thanks for testing. I think I'll just merge a bunch, check my local flow and (maybe)
add some polish to the code.

stp-ip:
No deprecations and working with older config and older code simultaneously so fine to merge this for v1.7.0.
Would love to get a ping on the polish.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/file+a..." ]
>No deprecations and working with older config and older code simultaneously so
>fine to merge this for v1.7.0.

Well, we can leave the option 'transfer' but it will silently not work. This needs to be a
breaking change. Of which we have lined up bunch for 1.7.0 already


>Would love to get a ping on the polish.
>
>—
>You are receiving this because your review was requested.
>Reply to this email directly, view it on GitHub, or unsubscribe.*
>
/Miek

--
Miek Gieben

chrisohaver:
Thanks for reviewing.  The existing implementation shouldn't be deprecated until notifies are supported ...e.g. #3420 + auto/file implementation of  changes in #3420...

stp-ip:
As long as one doesn't use the transfer plugin, the old transfer code will work as expected. So not a breaking change with the old config. Posted a plan on rolling this out in the meta issue:
https://github.com/coredns/coredns/issues/3121#issuecomment-620717169

> With 1.7.0 transfer plugin  (merged) + related plugin interfaces (reviewed)
> With 1.7.x add notify functionality and deprecate transfer usage within plugins in docs
> With 1.8.0 deprecate transfer usage within plugins with error/warning
> With 1.9.0 remove transfer usage by plugins

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/file+a..." ]
>Thanks for reviewing. The existing implementation shouldn't be deprecated until
>notifies are supported ...e.g. #3420 + auto/file implementation of changes in #
>3420...

Ack, without notifies I can't actually test the code; because I want them to be supported.

plugin/auto/transfer.go
```diff
@@ -0,0 +1,22 @@
+package auto
+
+import (
+	"github.com/coredns/coredns/plugin/file"
+	"github.com/coredns/coredns/plugin/transfer"
+	"github.com/miekg/dns"
+)
+// Transfer implements Transfer.Transferer
+func (a Auto) Transfer(zone string, serial uint32) (<-chan []dns.RR, error) {
+	// look for exact zone match
```
miekg, line 10: this needs to be done as done here: https://github.com/coredns/coredns/pull/3972/files#diff-f1591d056edea707afdeae51b8dc8b8fR12

plugin/file/transfer.go
```diff
@@ -0,0 +1,54 @@
...
+	apex, err := z.ApexIfDefined()
+	if err != nil {
+		close(ch)
+		return nil, err
+	}
+
+	ch <- apex
+
+	// ApexIfDefined ensures that first record is an SOA
+	if serial >= apex[0].(*dns.SOA).Serial && serial != 0 {
```
miekg, line 39: this is wrong because it doesn't take into account serial arithmetic

plugin/file/transfer.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	if z == nil {
+		return nil, transfer.ErrNotAuthoritative
+	}
+	return z.Transfer(serial)
+}
+
+// Transfer returns a channel containing records of a zone transfer response for the zone
+func (z Zone) Transfer(serial uint32) (<-chan []dns.RR, error) {
+	ch := make(chan []dns.RR, 2)
```
miekg, line 27: why is this buffered, why is the value 2?

miekg:
btw. I'm not willing to make the effort to keep hauling all deprecated code forward. If folks are unable to change their corefile we could rewrite it for them (which in general would not be a bad tool to have)

all the current xfr stuff should be deleted.

Issue #3735, opened 2020-03-07
Recovered from panic in server


pytimer:
**What happened**:

I have two kubernetes cluster and use coredns with hostNetwork, i found one of cluster coredns pod output `[ERROR] Recovered from panic in server: "dns://:1553"` every few seconds. But the dns resolve is ok.

**What you expected to happen**:

coredns no error logs.

**How to reproduce it (as minimally and precisely as possible)**:

I don't know why have this problem, so i can't reproduce it.

**Anything else we need to know?**:

**Environment**:

- the version of CoreDNS: 1.6.5
- Corefile:
```yaml
apiVersion: v1
data:
  Corefile: |
    .:1553 {
        errors
        health :9180 {
           lameduck 5s
        }
        ready :9181
        kubernetes cluster.local in-addr.arpa ip6.arpa {
           pods insecure
           fallthrough in-addr.arpa ip6.arpa
           ttl 30
        }
        prometheus :9153
        forward . /etc/resolv.conf
        cache 30
        loop
        reload
        loadbalance
    }
kind: ConfigMap
metadata:
  name: coredns
  namespace: kube-system
```
- logs, if applicable:
```sh
[INFO] plugin/ready: Still waiting on: "kubernetes"
.:1553
[INFO] plugin/reload: Running configuration MD5 = ea45c18bb97e6b6409eba23a1166b357
CoreDNS-1.6.5
linux/amd64, go1.13.4, c2fd1b2
[ERROR] Recovered from panic in server: "dns://:1553"
[INFO] plugin/ready: Still waiting on: "kubernetes"
[ERROR] Recovered from panic in server: "dns://:1553"
[ERROR] Recovered from panic in server: "dns://:1553"
[ERROR] Recovered from panic in server: "dns://:1553"
[ERROR] Recovered from panic in server: "dns://:1553"
[ERROR] Recovered from panic in server: "dns://:1553"
[ERROR] Recovered from panic in server: "dns://:1553"
[ERROR] Recovered from panic in server: "dns://:1553"
[ERROR] Recovered from panic in server: "dns://:1553"
```
- OS (e.g: `cat /etc/os-release`):
```sh
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
[...]

PR #3741, opened 2020-03-12
Check if the returned result of  forward.parseStanza() is 0

zouyee:

fix recovered from panic in server
https://github.com/coredns/coredns/issues/3735
none
none

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -91,6 +91,9 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	span = ot.SpanFromContext(ctx)
 	i := 0
 	list := f.List()
+	if len(list) == 0 {
+		return dns.RcodeServerFailure, ErrNoForward
+	}
 	deadline := time.Now().Add(defaultTimeout)
 	start := time.Now()
 	for time.Now().Before(deadline) {
--- a/plugin/forward/setup.go
+++ b/plugin/forward/setup.go
@@ -95,7 +95,7 @@ func parseStanza(c *caddy.Controller) (*Forward, error) {
 
 	toHosts, err := parse.HostPortOrFile(to...)
 	if err != nil {
-		return f, err
+		log.Warningf("parse with err: %v", err)
 	}
 
 	transports := make([]string, len(toHosts))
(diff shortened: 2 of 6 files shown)


miekg:
thanks for picking this up; as dicussed on the bug, I'm not sure what the better option is. When I saw this PR I thought, I wish we would have had a presubmit check for this (for future occurrences). In that light, maybe we *should* make it an error if nothing is parsed by `HostPortOrFile`?

chrisohaver:
> thanks for picking this up; as dicussed on the bug, I'm not sure what the better option is. When I saw this PR I thought, I wish we would have had a presubmit check for this (for future occurrences). In that light, maybe we _should_ make it an error if nothing is parsed by `HostPortOrFile`?

I'm leaning slightly toward making it an error, but either way is fine with me.  I think getting zero results from `HostPortOrFile` would, in virtually all reasonable cases be an invalid result.  IOW, for most usage, it's an error. The error could be a public type in the package, so in case there is some one in need of accepting a zero result, they could disregard that error.  In the unlikely future event that accepting zero results becomes a common need, we could add a new function that calls `HostPortOrFile` and  disregards the zero length.

zouyee:
> thanks for picking this up; as dicussed on the bug, I'm not sure what the better option is. When I saw this PR I thought, I wish we would have had a presubmit check for this (for future occurrences). In that light, maybe we _should_ make it an error if nothing is parsed by `HostPortOrFile`?

In the forward plugin, it invokes HostPortOrFile to resolve the corresponding nameservers. If the file is empty or the nameserver is invalid, HostPortOrFile  will ignore the corresponding string. Therefore, when HostPortOrFile is revoked by forward plugin  returns the Error(not an IP address or file),  the caller could use the log mode instead of returning the error directly

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Check if the ..." ]
>> thanks for picking this up; as dicussed on the bug, I'm not sure what the better option is. When I saw this PR I thought, I wish we would have had a presubmit check for this (for future occurrences). In that light, maybe we _should_ make it an error if nothing is parsed by `HostPortOrFile`?
>
>I'm leaning slightly toward making it an error, but either way is fine with me.  I think
>getting zero results from `HostPortOrFile` would, in virtually all reasonable cases be an
>invalid result.  IOW, for most usage, it's an error.

I agree with that, let's make it an error is zero results would be returned.

PR #3736, opened 2020-03-08
Stop serving stale responses from negative cache.

gonzalop:
Serving stale responses from the negative cache interacts poorly with
other plugins and seems to confuse users.

Fixes #3586.

### 1. Why is this pull request needed and what does it do?
See issue #3586 

### 2. Which issues (if any) are related?
#3586 

### 3. Which documentation changes (if any) need to be made?
None.

### 4. Does this introduce a backward incompatible change or deprecation?
No.

Issue #3586, opened 2020-01-10
"serve_stale" option in the "cache" plugin behaves incorrectly


zuzzas:
**What happened**:
`serve_stale` does not update NXDOMAIN status if it gets constantly hammered by requests.

**What you expected to happen**:
CoreDNS updates the record once an upstream DNS starts to return an A record after returning NXDOMAIN.

**How to reproduce it (as minimally and precisely as possible)**:
Begin the exercise by repeatedly hammering a CoreDNS instance with requests to the non-existent domain. It correctly returns NXDOMAIN.
```bash
while sleep 0.5; do dig test.default.svc.cluster.local @169.254.20.10; done
```

Create an appropriate (I've just created a Service in Kubernetes) A record on the upstream DNS. Verify:
```
$ dig test.default.svc.cluster.local. @192.168.0.10 +short
10.10.50.53
```

The aforementioned `while` loop will return NXDOMAIN indefinitely.

Removing the `serve_stale` option alleviates the issue.

**Anything else we need to know?**:

Notice, that these tests are performed not against the primary CoreDNS of a Kubernetes cluster, but against a secondary one that forwards requests to the primary (a node-level caching mechanism).

**Environment**:

- the version of CoreDNS: 1.6.6
- Corefile:
```
.:53 {
  errors {
    consolidate 10s ".* i/o timeout$"
    consolidate 10s ".* write: operation not permitted$"
  }
  cache {
    success 39936
    denial 9984
    prefetch 10 1m 25%
    serve_stale
  }
  reload 2s
  loop
  bind 192.168.0.10 169.254.20.10
  forward . 192.168.0.10 192.168.0.10 192.168.0.10 {
    max_fails 0
  }
  prometheus 127.0.0.1:9254
  health 127.0.0.1:9225
}
```
- logs, if applicable:
- OS (e.g: `cat /etc/os-release`):
```
NAME="Ubuntu"
VERSION="18.04.3 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.3 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
[...]

PR #3744, opened 2020-03-13
plugin/cache: fix negative cache masking cases

chrisohaver:
**Note:** This PR fixes a bug that affects the default usage of cache, not just those who enable stale cache retrieval.

This fixes some bugs introduced when the stale cache retrieval feature was added (#3468):
1. An expired (beyond stale) negative cache would mask positive cache, sending a query upstream thus missing the cache. This affects the default usage of cache. (#3701 ) Cache misses would also not be counted in this case. 
2. A stale negative cache entry could be returned to a client despite a fresh positive cache being available. (#3586)

The first issue is fixed by tweaking the logic in getIgnoreTTL.
The other issue is fixed by having prefetch remove cache entries from the negative cache when writing to the positive cache.

The test added in this PR is based on the test in PR #3702 (thanks @huntharo !)
I also successfully tested this change with tests in PR #3702, which could therefore merge if this PR merges.

This PR should supersede #3736 

Issues:
#3701 
#3586 

PRs:
#3702
#3736 

none

no

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -174,7 +174,7 @@ func (w *ResponseWriter) WriteMsg(res *dns.Msg) error {
 
 	if hasKey && duration > 0 {
 		if w.state.Match(res) {
-			w.set(res, key, mt, duration, w.prefetch)
+			w.set(res, key, mt, duration)
 			cacheSize.WithLabelValues(w.server, Success).Set(float64(w.pcache.Len()))
 			cacheSize.WithLabelValues(w.server, Denial).Set(float64(w.ncache.Len()))
 		} else {
@@ -203,7 +203,7 @@ func (w *ResponseWriter) WriteMsg(res *dns.Msg) error {
 	return w.ResponseWriter.WriteMsg(res)
 }
 
-func (w *ResponseWriter) set(m *dns.Msg, key uint64, mt response.Type, duration time.Duration, prefetch bool) {
+func (w *ResponseWriter) set(m *dns.Msg, key uint64, mt response.Type, duration time.Duration) {
 	// duration is expected > 0
 	// and key is valid
 	switch mt {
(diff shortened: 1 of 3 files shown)


chrisohaver:
PTAL @gonzalop

gonzalop (commented):
Great! Thanks for the fix.

plugin/cache/cache.go
```diff
@@ -174,7 +174,7 @@ func (w *ResponseWriter) WriteMsg(res *dns.Msg) error {
 
 	if hasKey && duration > 0 {
 		if w.state.Match(res) {
-			w.set(res, key, mt, duration)
+			w.set(res, key, mt, duration, w.prefetch)
```
miekg: shouldn't `w.prefetch` be available in set as it's a method on `w` ?

chrisohaver: oh duh.  thanks!

plugin/cache/cache.go
```diff
@@ -203,13 +203,17 @@ func (w *ResponseWriter) WriteMsg(res *dns.Msg) error {
...
 
-func (w *ResponseWriter) set(m *dns.Msg, key uint64, mt response.Type, duration time.Duration) {
+func (w *ResponseWriter) set(m *dns.Msg, key uint64, mt response.Type, duration time.Duration, prefetch bool) {
 	// duration is expected > 0
 	// and key is valid
 	switch mt {
 	case response.NoError, response.Delegation:
 		i := newItem(m, w.now(), duration)
 		w.pcache.Add(key, i)
+		// when pre-fetching, remove the negative cache entry if it exists
```
miekg, line 213: hmm, is this *only* need on prefetch, or maybe always?

chrisohaver: I don't believe so, and it would be a performance impact if we always do it.  I think due to the way things fall out, it's only needed in prefetch...

During an inline query cache retrieval, if there is an unexpired negative cache entry, we reply with that entry - no writes to cache occur, so the positive cache would remain empty.  IOW, if it was not for prefetch, there could not be an unexpired entry in both negative and positive cache.

miekg (approved):

yifan-gu:
@chrisohaver Thank you for the fix!
FWIW, recently we also found that together with kubernetes' autopath, this bug will cause unexpected high cache miss rate, prob because the autopath generates some negative cache during its search, which masks the positive cache.

PR #3754, opened 2020-03-16
Fix cacheMisses counter to be Inc() on all codepaths

ugodiggi:
The existing code only increments the cacheMissed counter in some, not all codepaths.
This PR contains two commits:
* fix cacheMisses counter to be Inc() on all codepaths
* small code reorganization to simplify that code

None

None

No

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -232,6 +232,9 @@ func (w *ResponseWriter) Write(buf []byte) (int, error) {
 	return n, err
 }
 
+// CacheClass is the type of cache, one of Success or Denial
+type CacheClass = string
+
 const (
 	maxTTL  = dnsutil.MaximumDefaulTTL
 	minTTL  = dnsutil.MinimalDefaultTTL
@@ -241,7 +244,7 @@ const (
 	defaultCap = 10000 // default capacity of the cache.
 
 	// Success is the class for caching positive caching.
-	Success = "success"
+	Success CacheClass = "success"
 	// Denial is the class defined for negative caching.
-	Denial = "denial"
+	Denial CacheClass = "denial"
 )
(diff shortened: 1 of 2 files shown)


chrisohaver:
Cache miss metric is also fixed #3744 ...

... and originally reported in #3701

plugin/cache/cache.go
```diff
@@ -241,7 +244,7 @@ const (
 	defaultCap = 10000 // default capacity of the cache.
 
 	// Success is the class for caching positive caching.
-	Success = "success"
+	Success CacheClass = "success"
```
miekg: Don't think `CacheClass` needs to be exported

plugin/cache/handler.go
```diff
@@ -27,14 +27,16 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 	server := metrics.WithServer(ctx)
 
 	ttl := 0
-	i := c.getIgnoreTTL(now, state, server)
+	i, iType := c.getIgnoreTTL(now, state, server)
```
miekg: is it `type` or `class`?

plugin/cache/cache.go
ugodiggi: Done

plugin/cache/handler.go
ugodiggi: Done

ugodiggi:
> Cache miss metric is also fixed #3744 ...
> 
> ... and originally reported in #3701

Thank you, I had not seen either. I'm happy to retire this PR if you prefer.

chrisohaver:
> > Cache miss metric is also fixed #3744 ...
> > ... and originally reported in #3701
> 
> Thank you, I had not seen either. I'm happy to retire this PR if you prefer.

Dunno - there is more code reduction here, and it brings the actual behavior of the `getIgnoreTTL` function in line with its name and description.

miekg:
#3744 has been merged, closing here.

PR #3755, opened 2020-03-16
Do cache NXDOMAIN responses that don't have SOA set

ugodiggi:

I have observed on EKS (AWS hosted Kubernetes) that the AWS DNS returns NXDOMAIN responses that do not include an SOA. Without this change, it is impossible to cache those responses even for DefaultMinTTL, as seen as `Type`=`OtherError`.

This PR does:
* Add a new Type: NonAuthNameError for recognizing NXDOMAIN responses
  that don't have SOA set
* Treat NonAuthNameError as a Denial for logging
* Do apply the default minTTL to NonAuthNameError responses, i.e. the
  denial MINTTL or the DefaultMinTTL.

NXDOMAIN responses that include some Answer are still `Typify`-ed as `OtherError` in order to respect existing tests.

See slack thread: https://cloud-native.slack.com/archives/C4DF7FP71/p1584282386084400

I don't think any is needed.

Some DNS responses that were not being cached are now cached, up to DefaultMinTTL or the denial MINTTL. Personally I don't see it as a backward incompatible change but maybe it is a notable change in behavior.

### DIFF
--- a/plugin/pkg/response/typify.go
+++ b/plugin/pkg/response/typify.go
@@ -15,6 +15,8 @@ const (
 	NoError Type = iota
 	// NameError is a NXDOMAIN in header, SOA in auth.
 	NameError
+	// NonAuthNameError is a NXDOMAIN in header, no SOA in auth.
+	NonAuthNameError
 	// ServerError is a set of errors we want to cache, for now it contains SERVFAIL and NOTIMPL.
 	ServerError
 	// NoData indicates name found, but not the type: NOERROR in header, SOA in auth.
@@ -30,14 +32,15 @@ const (
 )
 
 var toString = map[Type]string{
-	NoError:     "NOERROR",
-	NameError:   "NXDOMAIN",
-	ServerError: "SERVERERROR",
-	NoData:      "NODATA",
-	Delegation:  "DELEGATION",
-	Meta:        "META",
-	Update:      "UPDATE",
-	OtherError:  "OTHERERROR",
+	NoError:          "NOERROR",
+	NameError:        "NXDOMAIN",
+	NonAuthNameError: "NXDOMAINNOSOA",
+	ServerError:      "SERVERERROR",
+	NoData:           "NODATA",
+	Delegation:       "DELEGATION",
+	Meta:             "META",
+	Update:           "UPDATE",
+	OtherError:       "OTHERERROR",
 }
 
 func (t Type) String() string { return toString[t] }
@@ -105,8 +108,16 @@ func Typify(m *dns.Msg, t time.Time) (Type, *dns.OPT) {
 	if soa && m.Rcode == dns.RcodeSuccess {
 		return NoData, opt
 	}
-	if soa && m.Rcode == dns.RcodeNameError {
-		return NameError, opt
+	if m.Rcode == dns.RcodeNameError {
+		if soa {
+			return NameError, opt
+		} else if len(m.Answer) == 0 {
+			return NonAuthNameError, opt
+		} else {
+			// This is an NXDOMAIN response that includes a non-empty Answer.
+			// It's self-contradicting and we are not going to cache it.
+			return OtherError, opt
+		}
 	}
 
 	if m.Rcode == dns.RcodeServerFailure || m.Rcode == dns.RcodeNotImplemented {
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -163,7 +163,7 @@ func (w *ResponseWriter) WriteMsg(res *dns.Msg) error {
 
 	msgTTL := dnsutil.MinimalTTL(res, mt)
 	var duration time.Duration
-	if mt == response.NameError || mt == response.NoData {
+	if mt == response.NameError || mt == response.NonAuthNameError || mt == response.NoData {
 		duration = computeTTL(msgTTL, w.minnttl, w.nttl)
 	} else if mt == response.ServerError {
 		// use default ttl which is 5s
@@ -211,7 +211,7 @@ func (w *ResponseWriter) set(m *dns.Msg, key uint64, mt response.Type, duration 
 		i := newItem(m, w.now(), duration)
 		w.pcache.Add(key, i)
 
-	case response.NameError, response.NoData, response.ServerError:
+	case response.NameError, response.NonAuthNameError, response.NoData, response.ServerError:
 		i := newItem(m, w.now(), duration)
 		w.ncache.Add(key, i)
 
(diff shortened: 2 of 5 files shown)


miekg:
See https://tools.ietf.org/html/rfc2308#section-2.1 are CNAMEs involved? If not, than AWS not following the spec and we should not bend over backwards to enable this.

ugodiggi:
> See https://tools.ietf.org/html/rfc2308#section-2.1 are CNAMEs involved? If not, than AWS not following the spec and we should not bend over backwards to enable this.

Thank you for your answer @miekg .

This is what I observe with dig:
```
dig @172.28.0.2 google.com.ec2.internal.

; <<>> DiG 9.10.3-P4-Ubuntu <<>> @172.28.0.2 google.com.ec2.internal.
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 48662
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;google.com.ec2.internal.	IN	A

;; Query time: 0 msec
;; SERVER: 172.28.0.2#53(172.28.0.2)
;; WHEN: Sun Mar 15 14:58:16 UTC 2020
;; MSG SIZE  rcvd: 41
```
I am engaging with AWS to figure out why their behavior is non-standard.

I tried to rad through the RFC, and I do see very explicit language in  https://tools.ietf.org/html/rfc2308#section-5:
```
   Negative responses without SOA records SHOULD NOT be cached as there
   is no way to prevent the negative responses looping forever between a
   pair of servers even with a short TTL.

   Despite the DNS forming a tree of servers, with various mis-
   configurations it is possible to form a loop in the query graph, e.g.
   two servers listing each other as forwarders, various lame server
   configurations.  Without a TTL count down a cache negative response
   when received by the next server would have its TTL reset.  This
   negative indication could then live forever circulating between the
   servers involved.
```
The part that is confusing to me is that in Kubernetes we are trying to use CoreDNS as the top level DNS server, see i.e. the efforts in creating the [nodelocaldns](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns/nodelocaldns) addon. In that case, the issue mentioned in the RFC (a query graph loop) does not seem relevant.

Would you be open in making this behavior conditional to some explicit configuration in the cache plugin?

ugodiggi:
Also, for comparison, `dnsmasq` seems to [provide this option](http://www.thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html):
```
--neg-ttl=<time>
Negative replies from upstream servers normally contain time-to-live information in SOA records which dnsmasq uses for caching. If the replies from upstream servers omit this information, dnsmasq does not cache the reply. This option gives a default value for time-to-live (in seconds) which dnsmasq uses to cache negative replies even in the absence of an SOA record.
```

plugin/pkg/response/typify.go
```diff
@@ -105,8 +108,16 @@ func Typify(m *dns.Msg, t time.Time) (Type, *dns.OPT) {
...
+	if m.Rcode == dns.RcodeNameError {
+		if soa {
+			return NameError, opt
+		} else if len(m.Answer) == 0 {
+			return NonAuthNameError, opt
+		} else {
+			// This is an NXDOMAIN response that includes a non-empty Answer.
+			// It's self-contradicting and we are not going to cache it.
+			return OtherError, opt
+		}
```
chestermjr, line 120:
```suggestion
	if m.Rcode == dns.RcodeNameError {
		if soa {
			return NameError, opt
		}
		if len(m.Answer) == 0 {
			return NonAuthNameError, opt
		}
		// This is an NXDOMAIN response that includes a non-empty Answer.
		// It's self-contradicting and we are not going to cache it.
		return OtherError, opt
	}
```

miekg:
DNS is a broken protocol mostly because of these work arounds, this needs to be fixed at the source not by complicating our code because upstream is doing the wrong thing.
Also: will you be on the hook for removing this work around once it's not needed? When will that be? How can we track? 

So NACK.

chrisohaver:
Another angle to remedying the issue could be to extend the _rewrite_ plugin to enable conditionally adding records to a response. In theory this would be more generally utilitarian,  but in reality, I cant immediately think of any meaningful use case other than this one.

Quentin-M:
We are also facing this right now. We are currently hammering AWS DNS servers because we are not able to negative-cache those.

```
dig foo.bar.eu-west-1.compute.internal.

; <<>> DiG 9.11.2-P1 <<>> foo.bar.eu-west-1.compute.internal.
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 8077
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;foo.bar.eu-west-1.compute.internal. IN	A

;; Query time: 0 msec
;; SERVER: 10.13.0.2#53(10.13.0.2)
;; WHEN: Mon May 04 22:00:32 UTC 2020
;; MSG SIZE  rcvd: 52
```

ugodiggi:
We are running a forked version of CoreDNS that does include this PR.

This is not optimal, of course, and we have in our pipeline a change to our nodes launch configuration that removes `ec2.internal` from the search path.

In a CloudFormation template, the diff looks something like this:
```
        UserData:
          Fn::Base64:
            !Sub |
              #!/bin/bash
              set -o xtrace
+             sed "s|ec2.internal||" /etc/resolv.conf > /etc/resolv.conf.kubelet
              /etc/eks/bootstrap.sh ${AWS::StackName}-Cluster \
                --kubelet-extra-args '--node-labels=kubernetes.io/role=node \
+              --resolv-conf=/etc/resolv.conf.kubelet' \

              /opt/aws/bin/cfn-signal --exit-code $? \
                       --stack  ${AWS::StackName} \
                       --resource NodeGroup  \
                       --region ${AWS::Region}
```
i.e. we create a copy of the node's resolv.conf without `ec2.internal`, and use it as the kubelet's `--resolv-conf`.

We have not yet tested this extensively, but preliminary tests suggest that it will work.

nishaad78:
Hi @ugodiggi did you get any response from AWS? We hit this problem today on our Kubernetes cluster as well.

nishaad78:
Just to follow up, we solved this for our use case by updating the `ndots` in resolv.conf

The DNS queries are now first made with the valid fqdn (without the search domain suffix).

dmizelle:
@nishaad78 We just got bit by this too, and are talking with AWS.

PR #3776, opened 2020-03-26
Cleanup metrics

miekg:
Cleanup a variety of metric issues.
* Eliminate department of redundancy "count_total" naming.
* Use the plural of the unit when appropriate. (ex, "requests")
* Remove label names from metric names where appropriate. (ex, "rcode")
* Simplify request metrics by consolidating type label in to the base
request counter.
* Re-generate man pages.

miek: this is a copy of #3671 with the manual pages changes backed out
and then rebased on top of master. This required some new git-foo from
myself so YMMV with what we've ended up with.

### DIFF
--- a/plugin/acl/README.md
+++ b/plugin/acl/README.md
@@ -73,8 +73,8 @@ example.org {
 
 If monitoring is enabled (via the _prometheus_ plugin) then the following metrics are exported:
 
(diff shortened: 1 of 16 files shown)


SuperQ:
Oh, Sure, I could just rebase the other PR if you want.

miekg:
this seems to have done the trick - history is also preserved. I'll merge.

PR #3779, opened 2020-03-26
Add 1.7.0 notes already

miekg:
This is the work-in-progress release notes for 1.7.0, adding it now so
we can craft it will we work towards 1.7.0.

Notes currently yell that 1.7.0 is backwards incompatible.

### DIFF
--- a/notes/coredns-1.7.0.md
+++ b/notes/coredns-1.7.0.md
@@ -10,7 +10,9 @@ author = "coredns"
 The CoreDNS team has released
 [CoreDNS-1.7.0](https://github.com/coredns/coredns/releases/tag/v1.7.0).
 
-This is a **backwards incompatible release**. Major changes include: better metrics names, and newly
+This is a **backwards incompatible release**. Major changes include:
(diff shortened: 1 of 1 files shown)


notes/coredns-1.7.0.md
```diff
@@ -0,0 +1,20 @@
...
+tags = ["Release", "1.7.0", "Notes"]
+release = "1.7.0"
+date = 2020-03-24T10:00:00+00:00
+author = "coredns"
++++
+
+The CoreDNS team has released
+[CoreDNS-1.7.0](https://github.com/coredns/coredns/releases/tag/v1.7.0).
+
+This is a **backwards incompatible release**. Major changes include: better metrics names, and newly
```
SuperQ: Let's split up this to make it easier to read.

```suggestion
This is a **backwards incompatible release**. Major changes include:
* Better metrics names (PR #3776)
* New `transfer` plugin that removes the need for plugins to perform their own zone transfers.
```

Issue #3778, opened 2020-03-26
codify metric naming?


miekg:
We just had a massive metric name clean up which is an annoying backward incompat change.
Now in `test/presubmit_test.go` we already have a bunch of "tests" that check style issues that are specific to CoreDNS.

Can we add a metric naming test to this as well? 

/cc @SuperQ

SuperQ:
Interesting idea. I don't know if there's any linter code out there for metric names. Maybe @beorn7 has some ideas here.

beorn7:
`promtool check metrics` is doing a bit of linting, as in

```
$ curl -s localhost:8080/metrics | promtool check metrics
foo counter metrics should have "_total" suffix
```

I'd say there could be more checks in it, and I guess PRs will be welcome.

`promtool check metrics` could be part of your CI pipeline. Or the `github.com/prometheus/prometheus/util/promlint` package could be used directly.

One might argue that package could live in `prometheus/client_golang`, but I'm getting ahead of myself.

miekg:
ah that's neat. Slight down side of having another tool and the need to stand up prometheus, feed it queries (not all metrics are initialized on start) and then check. But it's good to know it possible.

I want to take a slightly different route and just parse the go code, looks for a prometheus metric declaration, get the type and then check - this can be done from a Go test

beorn7:
No need to run the Prometheus server.

The `promlint` package takes MetricFamily protobufs, IIUC. So you can just write a Go test that calls `Gather` on your metrics registry and feeds them to `promlint`.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] codify metric..." ]
>No need to run the Prometheus server.
>
>The `promlint` package takes MetricFamily protobufs, IIUC. So you can just write a Go test that calls `Gather` on your metrics registry and feeds them to `promlint`.

ok, thanks. I'll look into it!

chrisohaver:
> The promlint package

nice

PR #3789, opened 2020-03-28
introduce metric naming test

zouyee:
Signed-off-by: zounengren <zounengren@cmss.chinamobile.com>

introduce metric naming test to correct metrics' name

https://github.com/coredns/coredns/issues/3778

none
none


plugin/hosts/metrics.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	// DeprecatedhostsEntries is the combined number of entries in hosts and Corefile.
+	// non-histogram and non-summary metrics should not have "_count" suffix
+	DeprecatedhostsEntries = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "hosts",
+		Name:      "entries_count",
```
SuperQ: Let's just drop this metric, we're breaking metric names in 1.7.x

plugin/hosts/metrics.go
```diff
@@ -0,0 +1,32 @@
...
+		Help:      "[]The combined number of entries in hosts and Corefile.",
+	}, []string{})
+	// HostsEntries is the combined number of entries in hosts and Corefile.
+	HostsEntries = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "hosts",
+		Name:      "entries_size",
```
SuperQ: This should just be "entries", because this is the "unit". "size", just like "count", is superfluous.

```suggestion
		Name:      "entries",
```

plugin/dnssec/metrics.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	// CacheSize is the number of elements in the dnssec cache.
+	CacheSize = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "dnssec",
+		Name:      "cache_size",
```
SuperQ: "size" is ambiguous. Metric names should have their "unit" to make them understandable with minimal context.

I would suggest "cache_elements" to match the help text. Or is this entries? "cache_entries"

SuperQ (changes requested):

SuperQ:
Perhaps we could refactor our per-package metrics by making a top-level exported var called "Metrics". This would allow the test to iterate over the plugins without having to know the names of each plugin metric name.

Something like this:
```golang
package cache

type Metrics struct {
  CacheSize, CacheHits, CacheMisses prometheus.Collector
}

var Metrics = Metrics{
  CacheSize: prometheus.NewGaugeVec(...)
  CacheHits: prometheus.NewCounterVec(...)
  CacheMisses: prometheus.NewCounterVec(...)
}
```

Or with a map:
```golang
package cache

var Metrics = map[string]prometheus.Collector{
  "CacheSize": prometheus.NewGaugeVec(...)
  "CacheHits": prometheus.NewCounterVec(...)
  "CacheMisses": prometheus.NewCounterVec(...)
}
```

test/metric_naming_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/coredns/coredns/plugin/template"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/prometheus/util/promlint"
+)
+
+func TestPrometheusMetrics(t *testing.T) {
+	m := metrics.New("")
+	collectors := []prometheus.Collector{
```
miekg: this is hard to keep to to date and also requires all metric variables are exported which is odd.
A better test is to iterator over the Go code and find the metrics by their type

SuperQ: I agree, what do you think of my `package.Metrics` idea?

miekg: there are already a bunch of unwritten rules about plugin development, i don't want to burden people with adding more. If we (=coredns core dev) care about this we should put in the effort to make this checking possible without burdening plugin developers.

plugin/cache/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	// cacheSize is total elements in the cache by cache type.
+	cacheSize = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "cache",
+		Name:      "size",
```
SuperQ, line 14: "size" is ambiguous, Size of what? I think you meant "entries" to match the similar metric in dnssec.

```suggestion
		Name:      "entries",
```

SuperQ (commented):
Minor naming nit.

test/metric_naming_test.go
zouyee: Currently,  it used `ast` to parse the file, and build the struct of MetricFamily

plugin/cache/metrics.go
zouyee: These metrics are split from `plugin/cache/handler.go`, could these modifications be placed in the next pr?

SuperQ: Sure, makes sense.

SuperQ (approved):
LGTM

man/coredns-hosts.7
```diff
@@ -1,5 +1,5 @@
 .\" Generated by Mmark Markdown Processer - mmark.miek.nl
-.TH "COREDNS-HOSTS" 7 "January 2020" "CoreDNS" "CoreDNS Plugins"
```
miekg, line 2: Can you remove these files from the PR?

PR #3797, opened 2020-03-31
Update more metrics names

miekg:
The subsystem of a metric should be it's plugin name, not something
generic like 'dns' (which is reserved for the core). (potentially this
can be tested as well).

Fix dns64 and acl plugin.

### DIFF
--- a/plugin/kubernetes/metrics.go
+++ b/plugin/kubernetes/metrics.go
@@ -9,10 +9,6 @@ import (
 	api "k8s.io/api/core/v1"
 )
 
-const (
-	subsystem = "kubernetes"
-)
-
 var (
 	// DnsProgrammingLatency is defined as the time it took to program a DNS instance - from the time
 	// a service or pod has changed to the time the change was propagated and was available to be
@@ -27,7 +23,7 @@ var (
 	//   * headless_without_selector
 	DnsProgrammingLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: plugin.Namespace,
-		Subsystem: subsystem,
+		Subsystem: "kubernetes",
 		Name:      "dns_programming_duration_seconds",
 		// From 1 millisecond to ~17 minutes.
 		Buckets: prometheus.ExponentialBuckets(0.001, 2, 20),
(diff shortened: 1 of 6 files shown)


ihac (approved):
LGTM (Seems doc test failed though~

plugin/kubernetes/metrics.go
```diff
@@ -27,7 +23,7 @@ var (
 	//   * headless_without_selector
 	DnsProgrammingLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: plugin.Namespace,
-		Subsystem: subsystem,
+		Subsystem: "kubernetes",
```
SuperQ, line 26: It's typically preferred to make these a const/var to avoid copy-paste mistakes when there are multiple metrics.

I might suggest we have a standard name like `pluginName` (doesn't need to be exported from the plugin package. This would make it easier and more clear than just having it documented as a must.

SuperQ (commented):
Naming changes look fine. But I feel like we should have a standard for the Subsystem name variable.

SuperQ:
Yay, doc test working as intended.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Update more m..." ]
>Naming changes look fine. But I feel like we should have a standard for the
>Subsystem name variable.

that's the plugin name, we should but using that here

I don't understand the 'doc test', if the intent is to have every PR now include
generated code it should be moved into another automated update like go mod
tidy

/Miek

--
Miek Gieben

SuperQ:
@miekg You approved the doc test.

SuperQ:
See https://github.com/coredns/coredns/pull/3799

SuperQ:
https://github.com/coredns/coredns/pull/3800 should take care of the doc test issue.

miekg:
closing this in favor for @SuperQ 's pr.

@SuperQ can you add a metrics section to dns64; https://github.com/coredns/coredns/pull/3797/files#diff-7cd15ffca9a2af1881c2ecee8409a8b1R57 ? Thanks
(*another* thing I should automate)

SuperQ:
@miekg Already done, I copied basically everything from this PR to mine. :smile:

miekg:
Excellent!

PR #3799, opened 2020-03-31
Update more metrics names

SuperQ:
The subsystem of a metric must be it's plugin name. "dns" is reserved
for the core. Use a const convention to avoid copy-paste mistakes in
naming.

Fix dns64 and acl plugin.

### DIFF
--- a/plugin.md
+++ b/plugin.md
@@ -56,7 +56,7 @@ server.
 ## Metrics
 
 When exporting metrics the *Namespace* should be `plugin.Namespace` (="coredns"), and the
-*Subsystem* should be the name of the plugin. The README.md for the plugin should then also contain
+*Subsystem* must be the name of the plugin. The README.md for the plugin should then also contain
 a *Metrics* section detailing the metrics.
 
 ## Readiness
--- a/plugin/acl/README.md
+++ b/plugin/acl/README.md
@@ -73,8 +73,8 @@ example.org {
 
 If monitoring is enabled (via the _prometheus_ plugin) then the following metrics are exported:
 
-- `coredns_dns_blocked_requests_total{server, zone}` - counter of DNS requests being blocked.
+- `coredns_acl_blocked_requests_total{server, zone}` - counter of DNS requests being blocked.
 
-- `coredns_dns_allowed_requests_total{server}` - counter of DNS requests being allowed.
+- `coredns_acl_allowed_requests_total{server}` - counter of DNS requests being allowed.
 
 The `server` and `zone` labels are explained in the _metrics_ plugin documentation.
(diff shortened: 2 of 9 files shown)


bradbeam (approved):

miekg:
uhm, no. We already know the plugin's name as this is a method in the interface: `plugin.Handler`, this duplicates it, add a constant (which is not really needed) as `Name()` returns it and it only need to be defined there.

The problem with all this, is that it does get a bit unwieldly, because you know want plugin devs to say
`&Forward{}.Name()` - not sure if that even compile or needs more braces

SuperQ:
The existing situation of multiple duplicate `"pluginName"` string literals is considered bad Golang style.

SuperQ:
To put it another way, right now after doing a `plugin.Register("name", setup)`, we use the same string literal on `plugin.Error("name", err)`. This itself is a bad practice, and should be replaced with a const. Same with `clog.NewWithPlugin("name")`.

stp-ip (approved):
looks good

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Update more m..." ]
>To put it another way, right now after doing a plugin.Register("name", setup),
>we use the same string literal on plugin.Error("name", err). This itself is a
>bad practice, and should be replaced with a const. Same with clog.NewWithPlugin
>("name").

well, yes, I don't consider the current thing perfect.

But - again - there is not test or presubmit test to enforce this, so it still up to the
reviewer to remember this new things - which is not perfect either.

SuperQ:
@miekg I agree, it's not perfect. IMO it's more a Go style thing.

There are code linters for this kind of thing:
```
$ go get github.com/jgautheron/goconst/cmd/goconst
$ goconst plugin/clouddns/...
plugin/clouddns/setup.go:18:30:2 other occurrence(s) of "clouddns" found in: plugin/clouddns/setup.go:20:31 plugin/clouddns/clouddns.go:222:43
plugin/clouddns/setup.go:20:31:2 other occurrence(s) of "clouddns" found in: plugin/clouddns/setup.go:18:30 plugin/clouddns/clouddns.go:222:43
plugin/clouddns/clouddns.go:222:43:2 other occurrence(s) of "clouddns" found in: plugin/clouddns/setup.go:18:30 plugin/clouddns/setup.go:20:31
plugin/clouddns/setup.go:49:19:1 other occurrence(s) of "invalid zone '%s'" found in: plugin/clouddns/setup.go:53:19
plugin/clouddns/setup.go:53:19:1 other occurrence(s) of "invalid zone '%s'" found in: plugin/clouddns/setup.go:49:19
```

PR #3801, opened 2020-03-31
Change the recommened version to 1.14 (as we use 1.14 by default)

yongtang:
Change the recommened version to 1.14 (as we use 1.14 by default now)


SuperQ (approved):
LGTM

PR #3807, opened 2020-04-01
Fix goconst warnings

SuperQ:
Replace duplicate string literals with const values.

### DIFF
--- a/plugin/autopath/setup.go
+++ b/plugin/autopath/setup.go
@@ -11,12 +11,14 @@ import (
 	"github.com/miekg/dns"
 )
 
-func init() { plugin.Register("autopath", setup) }
+const pluginName = "autopath"
+
+func init() { plugin.Register(pluginName, setup) }
 
 func setup(c *caddy.Controller) error {
 	ap, mw, err := autoPathParse(c)
 	if err != nil {
-		return plugin.Error("autopath", err)
+		return plugin.Error(pluginName, err)
 	}
 
 	c.OnStartup(func() error {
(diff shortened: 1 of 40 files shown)


miekg:
~~~
other occurrence(s) of "tcp" found in: core/dnsserver/server_tls.go:65:29  ..
~~~
a blanket "thou shall const" is not useful and actively hurts readability because now I need to go back to the original const def. which may be far away

SuperQ:
But having lots of copy-pasta also hurts because typos happen. By using a const, the typo is caught at compile time.

Some amount of style linting is good to catch these.

SuperQ:
I found a good example of a copy-paste mistake, slightly different from the rest of them.

azure/setup.go:132:
```
  return env, resourceGroupMapping, accessMap, fall, c.Errf("unknown property: %q", c.Val())
```

SuperQ:
I've removed this as a linter test, and just used the goconst tool to help find these style issues.

plugin/autopath/setup.go
```diff
@@ -11,12 +11,14 @@ import (
 	"github.com/miekg/dns"
 )
 
-func init() { plugin.Register("autopath", setup) }
+const pluginName = "autopath"
```
miekg, line 14: I'm still not a fan of this PR and intent to close it.

Technically we don't need this const because we can use `<plugin>.Name()` method

SuperQ: Let me see about making use of that. I think it's a bit expensive to call a string method, where a simple const at compile time will do.

darshanime:
IMO this PR helps unify the error message format and make it easy to adhere to the conventions (among other things) so 👍 from me.

PR #3818, opened 2020-04-08
Ignore vscode settings file

zouy414:
Signed-off-by: ZouYu <zouy.fnst@cn.fujitsu.com>

Ignore vscode settings file
None
None
None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ kubectl
 go-test-tmpfile*
 coverage.txt
 .idea/
+.vscode/
 .DS_Store
 *~
 .classpath


stp-ip:
Thanks for the PR.

This needs a bit of discussion in general. Using this PR as initial reasoning.

Overall it is the sentiment that adding gitignore options for individual editor choices shouldn't be done. With these coming up more and more we should once and for all take a stance. Either we merge and keep or we don't allow this and remove the options we already have.

cc. @miekg @yongtang (referenced as you previously merged the most of those)

In my view they are not hard PRs to merge, but without a clear yes/no policy one might argue it we should be open to merging every gitignore PR specific to an editor.

stp-ip:
Sidenote: https://github.com/coredns/coredns/pull/2628

SuperQ:
In Prometheus, I think we made the decision to not sprinkle editor file exceptions into our git ignores. 

We do exclude built binaries and build-related output. But nothing for specific editors.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Ignore vscode..." ]
>In Prometheus, I think we made the decision to not sprinkle editor file
>exceptions into our git ignores.
>
>We do exclude built binaries and build-related output. But nothing for specific
>editors.

yes! I want to do the same here

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Ignore vscode..." ]
>In Prometheus, I think we made the decision to not sprinkle editor file
>exceptions into our git ignores.
>
>We do exclude built binaries and build-related output. But nothing for specific
>editors.

see #3820

/close

PR #3820, opened 2020-04-08
Remove all editor stuff from .gitignore

miekg:
Remove all cruft from .gitignore and leave build artifacts and add
a comment.


johnbelamaric:
fine with me
/lgtm

Issue #2980, opened 2019-07-05
plugin/nsid: nsid not being set


stp-ip:
Neither the default nor any other value is being set via nsid.

Tested with the following config:
```
. {
  forward . 8.8.8.8
  errors stdout
  log
  nsid test
  cache
}
```

```
. {
  forward . 8.8.8.8
  errors stdout
  log
  nsid
  cache
}
```

```
. {
  forward . 8.8.8.8
  errors stdout
  log
  nsid {$HOSTNAME}
  cache
}
```

Result:
```
;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
; NSID
```

sharmaansh21:
I can look into it :bowtie:

stp-ip:
The culprit seems to be the cache, which might seem intentional. So debugging queries should probably not be cached and separated into their own stanza. Not sure, if this needs a fix.

stp-ip:
Another workaround is to have a separate stanza that forwards back to the main stanza (which still can have the cache).

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [nsid] nsid n..." ]
>The culprit seems to be the cache, which might seem intentional. So debugging queries should probably not be cached and separated into their own stanza. Not sure, if this needs a fix.

nsid is there to identify the server answering your query. This was never 
designed to be cached or survive being forward. If anything we fail to add
our own nsid when replying?

stp-ip:
So from the preliminary testing it is as follows:
nsid + cache: on first request it includes the nsid
nsid + cache: on cached responses it doesn't respond with the nsid (this seems like a bug)
nsid + no chache: nsid works as expected

To workaround this issue one could do:
```
.:53 {
  bind 127.0.0.1 # gets traffic here
  bind 127.0.0.2
  forward . 8.8.8.8
  cache
}
nsid.example.com:53 {
  forward . 127.0.0.2
  nsid
}
```
That way nsid adds its response to the cached items as well.

miekg:
Might be an indication of something deeper going on with OPT

stp-ip:
My guess would be cache stripping any data it deems unnecessary and nsid isn't in the path, when it responds, but that's a bad hunch. Hopefully can dive into the code soon, if @sharmaanshul21 hasn't jumped in :)

sharmaansh21:
Not yet, please go ahead.

stp-ip:
[...]

PR #3822, opened 2020-04-11
Fix NSID not being set on cached responses

erbesharat:
The cache plugin would remove the EDNS0 options from the responses before caching them. So this would create an edge-case in NSID plugin because the NISD plugin used to only add the NSID to the response if it contains the EDNS0 options.
This PR adds a check to bring back the EDNS0 section to the responses, if the incoming query contains EDNS0.

#2980 

None

No

### DIFF
--- a/plugin/nsid/nsid.go
+++ b/plugin/nsid/nsid.go
@@ -19,7 +19,8 @@ type Nsid struct {
 // ResponseWriter is a response writer that adds NSID response
 type ResponseWriter struct {
 	dns.ResponseWriter
-	Data string
+	Data    string
+	request *dns.Msg
 }
 
 // ServeDNS implements the plugin.Handler interface.
@@ -27,7 +28,7 @@ func (n Nsid) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (i
 	if option := r.IsEdns0(); option != nil {
 		for _, o := range option.Option {
 			if _, ok := o.(*dns.EDNS0_NSID); ok {
-				nw := &ResponseWriter{ResponseWriter: w, Data: n.Data}
+				nw := &ResponseWriter{ResponseWriter: w, Data: n.Data, request: r}
 				return plugin.NextOrFailure(n.Name(), n.Next, ctx, nw, r)
 			}
 		}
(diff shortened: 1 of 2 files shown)


yongtang:
@erbesharat Can you add a test case for this scenario?

miekg:
EDNS0 is hop-by-hop, so caching them is nonsensical because if will differ for each client; this why these are stripped off before being put in the cache. I.e this is WAI - probably makes sense to put something like this comment in the source it self.

miekg:
oh, I see we already care here - that's odd, but maybe I'm forgetting/misremembering the history here. In general the NSID plugin should just add it and we should not do anything here.

stp-ip:
> Well, when the response comes back from the cache it doesn't have the EDNS0 section and as far as I understood NSID only gets added to the response if it has the EDNS0 section:
> https://github.com/coredns/coredns/blob/0930eb8beb69684dfb1e1e066e2a7a6ed3827bf4/plugin/nsid/nsid.go#L40
> ah, that makes sense, then nsid is not doing the right thing because i needs to check the
> query
Quote from https://github.com/coredns/coredns/issues/2980.

Aka nsid checked the response, which doesn't work as the cache doesn't have a EDNS0 section.

So cache works as intended, but nsid doesn't. The PR is about nsid, but maybe I'm missing something.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fix NSID not ..." ]
>    Well, when the response comes back from the cache it doesn't have the EDNS0
>    section and as far as I understood NSID only gets added to the response if
>    it has the EDNS0 section:
>    https://github.com/coredns/coredns/blob/
>    0930eb8beb69684dfb1e1e066e2a7a6ed3827bf4/plugin/nsid/nsid.go#L40
>    ah, that makes sense, then nsid is not doing the right thing because i
>    needs to check the
>    query
>    Quote from #2980.
>
>Aka nsid checked the response, which doesn't work as the cache doesn't have a
>EDNS0 section.
>
>So cache works as intended, but nsid doesn't. The PR is about nsid, but maybe
>I'm missing something.

duh, mixed up the other PR, and didn't properly look at the actual files changed, just the
code.

ignore me - and continue; I'll just add some code comments then :)

miekg (approved):

miekg:
but before merging please add a test as @yongtang asks

yongtang (approved):
lgtm

PR #3830, opened 2020-04-14
GitHub action: Update Homebrew formula on new release

fhemberger:
This action checks out `coredns/deployment` and replaces values in `HomebrewFormula/coredns.rb` with the one from the new release tag (`GITHUB_REF`): version, download URL and SHA256 hash.

---

I already updated the Homebrew formula a couple of times manually in the past, but this is a bit tedious on every release and is sometimes forgotten. That's why I'm giving GitHub Actions a try.

Of course I couldn't test the Git push workflow itself (I took that from the other actions), also I'm not sure if `on release (published)` is the right trigger (as the assets may not be ready by the time the release is published), this still needs some testing.

However, the core logic in the Bash file works, and the values are replaced successfully. The script also detects if there are no assets for a release (e.g. for v1.6.8).

It would be great if you could have a look at it so macOS users get regular updates for coredns over Homebrew as well.

/cc @chrisohaver @miekg

### DIFF
--- a/.github/actions/update-homebrew/Dockerfile
+++ b/.github/actions/update-homebrew/Dockerfile
@@ -1,0 +1,6 @@
+FROM alpine
+
+RUN apk add curl git sed
+COPY entrypoint.sh /entrypoint.sh
+
+ENTRYPOINT ["/entrypoint.sh"]
--- a/.github/actions/update-homebrew/action.yml
+++ b/.github/actions/update-homebrew/action.yml
@@ -1,0 +1,6 @@
+---
+name: 'Update Homebrew'
+description: 'Updates Homebrew formula in coredns/deployment'
+runs:
+  using: 'docker'
+  image: 'Dockerfile'
--- a/.github/actions/update-homebrew/entrypoint.sh
+++ b/.github/actions/update-homebrew/entrypoint.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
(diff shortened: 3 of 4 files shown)


miekg:
I get the need for automation, but I don't want to be on the hook for maintaining all these github action hooks. This needs to live somewhere else and have some kind of mechanism of figuring out a new version was tagged.

fhemberger:
Yes, it would be nice if the Homebrew update was an integrated part of the release process itself. This was just the most straight forward "fix" I could come up with right now. I'm happy to discuss alternative solutions.

miekg:
you probably need to write a poller to see the change; then you can let is live externally.

closing this here; as this won't be merged.

fhemberger:
Maybe you should then drop support for Homebrew altogether if it's not maintained at all and there is no interest to automate updates as part of your build process. There is still the option to download the latest macOS binaries right from the release page.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] GitHub action..." ]
>Maybe you should then drop support for Homebrew altogether if it's not
>maintained at all and there is no interest to automate updates as part of your
>build process. There is still the option to download the latest macOS binaries
>right from the release page.

yes, we create binaries and a docker image as part of the release.

yongtang:
As coredns is a static binary without any dependencies, it actually is more convenience to just use binary than going through a release channel like brew I think. The release channels are mostly used for those packages with dependency management.

Issue #3810, opened 2020-04-03
plugin/etcd: lookup_test fails


takuhiro:
**What happened**:

I run tests for plugin/etcd, but lookup_test.go always fails.

**What you expected to happen**:

all tests pass.

**How to reproduce it (as minimally and precisely as possible)**:

1. Run etcd locally. I used [v3.4.5](https://github.com/etcd-io/etcd/releases/tag/v3.4.5) for darwin.
```
$ ./etcd &
```
2. Run tests with `-tags=etcd`
```
$ go test -tags=etcd ./plugin/etcd                                                                                                                                                         
--- FAIL: TestLookup (0.38s)
    lookup_test.go:331: rcode is "NXDOMAIN", expected "NOERROR"
{"level":"warn","ts":"2020-04-03T15:42:50.922+0900","caller":"clientv3/retry_interceptor.go:61","msg":"retrying of unary invoker failed","target":"endpoint://client-2ede7eb0-fbb2-4ef1-ba6d-469070a3ad33/localhost:2379","attempt":0,"error":"rpc error: code = FailedPrecondition desc = etcdserver: authentication is not enabled"}
FAIL
FAIL    github.com/coredns/coredns/plugin/etcd  0.903s
FAIL

```

**Anything else we need to know?**:

The following test case fails.
https://github.com/coredns/coredns/blob/53f3f0b666821588e721ceeea4766b76333b668b/plugin/etcd/lookup_test.go#L132-L138

The test seems to have passed before e3266d24

**Environment**:

- the version of CoreDNS: v1.6.7 or later
- Corefile: 
- logs, if applicable:
- OS (e.g: `cat /etc/os-release`): macOS
- Others:

msedzins:
Hi, I've looked into that and actually there are two separate issues with those tests. You can spot it when you run:
```
$go test -tags=etcd ./plugin/etcd   -v 
=== RUN   TestLookup
    TestLookup: lookup_test.go:331: rcode is "NXDOMAIN", expected "NOERROR"
--- FAIL: TestLookup (0.29s)

=== RUN   TestSetupEtcd
[...]

PR #3840, opened 2020-04-19
Fix to bug 3810

msedzins:

Hopefully it's a fix to:
https://github.com/coredns/coredns/issues/3810

I added a comment to the issue in which I provided more details.

In short:
The error was introduced in e3266d2 with the change in the line number 330, to be more specific:
value of "exact" parameter was changed from "false" to "true".

As a result of that change, when we make TXT record lookup in etcd, there needs to be an exact match between the value we look up for and the key in etcd.

in lookup_test.go we have the following record definition:
```
// TXT record in server1.
	{Host: "", Port: 8080, Text: "sometext", Key: "txt.server1.prod.region1.skydns.test."},
```

This record is stored in etcd. Later on, in our test case we  look up for that record:
```
	// TXT Test
	{
		Qname: "server1.prod.region1.skydns.test.", Qtype: dns.TypeTXT,
		Answer: []dns.RR{
			test.TXT("server1.prod.region1.skydns.test. 303 IN TXT sometext"),
		},
	},
```

As you can see:
QName!=Key -> that's why the record will not be found in etcd using exact match. 

After my proposed change the test runs successfully.

https://github.com/coredns/coredns/issues/3810

No changes

No.

### DIFF
--- a/plugin/etcd/lookup_test.go
+++ b/plugin/etcd/lookup_test.go
@@ -131,9 +131,9 @@ var dnsTestCases = []test.Case{
 	},
 	// TXT Test
 	{
-		Qname: "server1.prod.region1.skydns.test.", Qtype: dns.TypeTXT,
+		Qname: "txt.server1.prod.region1.skydns.test.", Qtype: dns.TypeTXT,
 		Answer: []dns.RR{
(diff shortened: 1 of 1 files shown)


miekg:
note the current CI doesn't run the etcd tests, because we need etcd to be up and running and we've stopped doing that...

Issue #3849, opened 2020-04-24
plugin/transfer: Transfer plugin should de documented as alpha or WIP


cardil:
**What happened**:
Transfer plugin is described on CoreDNS docs, and it looks like it's working supported solution. But in reality only implementation is in tests.

**What you expected to happen**:
I expected that plugin that is documented in docs will work as described. That is transfer my zone to other server.

**How to reproduce it (as minimally and precisely as possible)**:
Create a zone file. Create a CoreDNS server with that zone and transfer plugin pointing to external DNS like freedns.42.pl
Transfer will not work.

- the version of CoreDNS: 1.6.9

chrisohaver:
In its current state it is not a WIP or alpha, but no internal plugins actually implement the `transfer.Transferer` interface (hence the lack of examples in the README).  It's available for people to use with their own plugins, if they implement the `transfer.Transferer` interface.  I've opened a PR to make this clearer in the Transfer plugin README.

There are a few open PRs that implement the interface for some internal plugins (file, auto, kubernetes, and k8s_external).

cardil:
I mean _CoreDNS_ as a whole. I spend a couple of hours trying to figure out why my zone isn't transferred. Finally, I check out repo, and find that `transfer.Transferer` isn't implemented yet by any plugin. If I wasn't a Golang developer, I probably couldn't check that. 

Those docs should be for administrators. This plugin should be marked as non-functional (alpha, WIP, some other way) until some plugins (file, auto) implement it, and users can use it.

darshanime:
can this be closed now that #3851 is merged?

PR #3851, opened 2020-04-24
plugin/transfer: Mention that the transfer plugin is not implemented by any plugins

chrisohaver:

Clarifies that no internal plugins currently implement the transfer plugin interface.

#3849


yongtang:
@chrisohaver The test failure is being fixed in #3850

zouyee (approved):

PR #3855, opened 2020-04-25
Fix mixed indentation within tests

serverwentdown:

It removes all mixed indentation within tests and improves the formatting a little. I assumed there was no existing preference for the closing backtick. 

Rules used:
1. Corefiles: Tabs, following indentation of surrounding Go code. 
2. Corefiles: If only one zone, no preceding newline. 
3. Corefiles: Backtick follows the last closing brace.
4. Zones: Align fields using spaces

### DIFF
--- a/test/auto_test.go
+++ b/test/auto_test.go
@@ -22,8 +22,7 @@ func TestAuto(t *testing.T) {
 			directory ` + tmpdir + ` db\.(.*) {1}
 			reload 1s
 		}
-	}
-`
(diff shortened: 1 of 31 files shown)


miekg:
this, indeed, looks better. I'll merge, but without some automation there is no way future reviewers will know about this.

For better or worse, I think the above rules should also be put in plugin/plugin.md so we at least have them documented in the source tree.

Issue #3077, opened 2019-07-31
plugin/kubernetes: Support serving records for nodes


erikwilson:
## Feature Request

Please support an option for serving node records through the kubernetes plugin. 

https://kubernetes.io/docs/concepts/architecture/nodes/#addresses

The kubernetes plugin should be able to serve a record for every node Hostname, preferably the InternalIP but optionally the ExternalIP.

Currently this can be achieved with an external controller to watch node changes and updating a config map linked to by the hosts plugin, but it would be helpful for this functionality to be built into the kubernetes plugin.

chrisohaver:
If done, To avoid name collision with the existing internal fqdn schema, the internal fqdn of a node could be `<node-hostname>.node.<cluster-zone>` e.g. `foo.node.cluster.local.`.  

ExternalIPs probably would be better exposed by the _k8s_external_ plugin, producing records using the same schema, e.g. `foo.node.mydomain.com.`.

I recall this has been requested/suggested/discussed before.

erikwilson:
Sorry if a similar issue already exists, I thought one would exist so took a quick look at issues opened & closed but couldn't find anything.

Yah I agree putting the ExteralIP off to another plugin or not dealing with it is fine, however I need DNS entries for an unmodified Hostname rather than a FQDN. The reason being that some kubernetes projects depend on the node name resolving, such as metrics-server. For example, if I don't provide a hosts entry for the nodes then metrics-server will produce the following error when scraping for metrics:
`
E0801 14:31:41.074784       1 manager.go:111] unable to fully collect metrics: [unable to fully scrape metrics from source kubelet_summary:k3s-1: unable to fetch metrics from Kubelet k3s-1 (k3s-1): Get https://k3s-1:10250/stats/summary/: dial tcp: lookup k3s-1 on 10.43.0.10:53: no such host, unable to fully scrape metrics from source kubelet_summary:k3s-2: unable to fetch metrics from Kubelet k3s-2 (k3s-2): Get https://k3s-2:10250/stats/summary/: dial tcp: lookup k3s-2 on 10.43.0.10:53: no such host]
`
If it was just for me I would say FQDN is fine, but I need this to support the behavior of other projects. Perhaps having fqdn/hostname behavior as an option, or resolving the node hostname as a fall through, could help support the functionality I need without causing conflicts or collisions.

chrisohaver:
No worries, it may have been in one of the kubernetes repos.

[...]

PR #3863, opened 2020-04-28
 resolves node hostname and node IPs from Kubernetes clusters.

windskyer:
The node name cannot be resolved, but the pod needs to access the node by name
NO
NO
NO


.gitignore
```diff
@@ -4,3 +4,4 @@ coredns.exe
 build/
 release/
 vendor/
+.idea/
```
chrisohaver, line 7: Please don't add IDE specific items to the git ignore.  Only build artifacts should go here.

chrisohaver:
Why the changes to  CODE_OF_CONDUCT.md,  CONTRIBUTING.md and SECURITY.md ?

plugin/kubernetes/controller.go
```diff
@@ -191,6 +200,17 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
 		defaultResyncPeriod,
 		cache.ResourceEventHandlerFuncs{})
 
+	dns.nodeLister, dns.nodeController = object.NewIndexerInformer(
```
chrisohaver, line 203: if `k8s_node` is not enabled, we should not init/run the node informer (since without `k8s_node` the informer would be consuming resources for no purpose)

plugin/kubernetes/node.go
```diff
@@ -0,0 +1,47 @@
...
+	nodeList := k.APIConn.NodeIndex(node)
+	services := []msg.Service{}
+	zonePath := msg.Path(state.Zone, coredns)
+	rcode := dns.RcodeNameError
+
+	for _, n := range nodeList {
+		if node != n.Name {
+			continue
+		}
+
```
chrisohaver, line 21: Returning all internal and external IPs doesn't feel like the right thing to do here.  Seems like it should return internal or external, but not both.

plugin/k8s_node/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+~~~
+k8s_node
+~~~
+
+If you want to change the apex domain or use a different TTL for the returned records you can use
+this extended syntax.
+
+~~~
+k8s_node {
```
chrisohaver, line 25: This syntax doesn't line up the description above it... "If you want to change the apex domain ... "

chrisohaver:
The new plugin needs to be added to plugin.cfg.

chrisohaver:
FYI, Adding dns records for kubernetes nodes has been discussed in length in #3077 .

plugin/k8s_node/setup.go
```diff
@@ -0,0 +1,73 @@
...
+	return nil
+}
+
+func parse(c *caddy.Controller) (*Node, error) {
+	e := New()
+
+	for c.Next() { // external
+		zones := c.RemainingArgs()
+		e.Zones = zones
+		for i, str := range e.Zones {
```
chrisohaver, line 49: This doesn't look quite right - it's pulling zones from the local plugin declaration, but not from the server block if there are none defined in the plugin declaration.

plugin/kubernetes/node.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/coredns/coredns/request"
+	"github.com/miekg/dns"
+	"strings"
+)
+
+func (k *Kubernetes) Node(state request.Request) ([]msg.Service, int) {
+	node := state.Name()[:len(state.Name())-1]
+	nodeList := k.APIConn.NodeIndex(node)
+	services := []msg.Service{}
+	zonePath := msg.Path(state.Zone, coredns)
```
chrisohaver, line 14: IIRC, state.Zone contains the zone of the serverblock, not the zone that matched in the plugin zone declaration.

e.g. for ...

```
. {
  k8s_node myzone.
}
```

I think state.Zone is always ".", not "myzone.".

plugin/k8s_node/README.md
```diff
@@ -0,0 +1,54 @@
...
+~~~
+
+With the Corefile above, the following Service will get an `A` record for `vm10-0-10-80.ksc.com` with the IP address `10.0.10.80`.
+
+~~~
+status:
+  addresses:
+  - address: 10.0.10.80
+    type: InternalIP
+  - address: vm10-0-10-80.ksc.com
```
chrisohaver, line 48: I don't think the hostname type address typically holds a fully qualified domain name.

plugin/k8s_node/setup.go
```diff
@@ -0,0 +1,73 @@
...
+		return e
+	})
+
+	return nil
+}
+
+func parse(c *caddy.Controller) (*Node, error) {
+	e := New()
+
+	for c.Next() { // external
```
chrisohaver, line 46: Is `//external` copy-paste cruft from k8s_external? Or does it mean something here.

chrisohaver (changes requested):
Thanks for contributing!
Adding dns records for kubernetes nodes has been discussed at length in #3077 .

I have a few comments/questions.

plugin/k8s_node/README.md
```diff
@@ -0,0 +1,54 @@
...
+This plugin resolves node external IP and internal IP address(es) of Kubernetes clusters. 
+This plugin is only useful if the *kubernetes* plugin is also loaded.
+
+The plugin resolve node IP addresses. It only handles queries for A and AAAA records;
+all others result in NODATA responses.
+
+## Syntax
+
+~~~
+k8s_node
```
chrisohaver, line 18: Your parse function pulls zones from the plugin declaration.  i.e.  `k8s_node [ZONES...]`

plugin/kubernetes/node.go
Sh4d1: Agree. It could return externalIP if it exists, and if not the internal IP

johnbelamaric: Hmm. Everything else returned by cluster DNS is internal. If the externalIP is firewalled (highly likely), this could cause issues. Refer to https://github.com/kubernetes/kubernetes/issues/90764

If you want to serve both you could have a separate external zone, perhaps?

miekg:
very much a stale pr, with many open questions. closing

linwalth:
Hello! I am in a situation where I would profit from such a solution :)

Issue #3868, opened 2020-05-01
kubernetes/client-pod-name and kubernetes/namespace metadata values will not log


aaronbbrown:
**What happened**:

I attempted to log the metadata labels `/kubernetes/client-namespace` and `/kubernetes/client-pod-name` using a combination of the `log`, `metadata`, and `kubernetes` plugins and the client-namespace and client-pod-name are always `-` in the logs, e.g.:

```
[INFO] A IN kubernetes.default.svc.dev.local. - -
```

**What you expected to happen**:

I expected the value of the pod name and namespace to be output to the logs.

**How to reproduce it (as minimally and precisely as possible)**:

I configured coredns with the following `Corefile`:
```
    .:53 {
        metadata .
        log . "{type} {class} {name} {/kubernetes/client-namespace} {/kubernetes/client-pod-name}"
        errors
        ready
        health
        kubernetes dev.local in-addr.arpa ip6.arpa {
          pods verified
          fallthrough in-addr.arpa ip6.arpa
          ttl 30
        }
        prometheus :9153
        forward . /etc/resolv.conf
        cache 30
        loop
        reload
        loadbalance
    }
```

Otherwise, it's a vanilla installation of coredns, nearly identical to the manifest in the repo.

I then started an interactive pod and used `dig` to make queries against coredns:
```
$ kubectl run --rm -it --image=tutum/dnsutils  -- bash
root@bash-64cb67db8c-ph6bt:/# dig +short google.com
172.217.13.142
root@bash-64cb67db8c-ph6bt:/# dig +short +tcp kubernetes.default.svc.dev.local
10.125.128.1
root@bash-64cb67db8c-ph6bt:/# dig +short  kubernetes.default.svc.dev.local
10.125.128.1
```

The following output was in my logs:
```
[INFO] A IN kubernetes.default.svc.dev.local. - -
[INFO] A IN google.com. - - -
[INFO] A IN kubernetes.default.svc.dev.local. - -
```

If I add `{/kubernetes/namespace}` to the `log` line, that will correctly show `default` as the namespace. My assumption is that [this check](https://github.com/coredns/coredns/blob/c9d567ab4423d391e5b3bef874185c324b6ed2ed/plugin/kubernetes/metadata.go#L47-L50) is failing, but it leans on a lot of libraries and I haven't been able to debug that.

**Anything else we need to know?**:

**Environment**:

- the version of CoreDNS: 1.6.9
- Corefile:
```
    .:53 {
        metadata .
[...]

PR #3869, opened 2020-05-01
plugin/kubernetes: Don't panic with metadata enabled and pods mode not verified

chrisohaver:

Enabling _metadata_ without enabling `pods verified` in _kubernetes_ would cause a panic when building metadata. Users should be able to use kubernetes metadata without enabling `pods verified`. This PR prevents the panic by making sure `pods verified` is enabled before looking up the IP address in the Pod index.

This may also be fixing a panic if _autopath_ is enabled without enabling `pods verified` in _kubernetes_ (though that would be a misconfiguration anyways).

None.

None.

No.

### DIFF
--- a/plugin/kubernetes/autopath.go
+++ b/plugin/kubernetes/autopath.go
@@ -51,6 +51,9 @@ func (k *Kubernetes) AutoPath(state request.Request) []string {
 
 // podWithIP return the api.Pod for source IP. It returns nil if nothing can be found.
 func (k *Kubernetes) podWithIP(ip string) *object.Pod {
+	if k.podMode != podModeVerified {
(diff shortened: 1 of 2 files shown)


chrisohaver:
Noticed this while looking into #3868, though I don't think it's the cause at all.

I should add a unit test for this before it merges.

miekg (approved):

aaronbbrown:
> Noticed this while looking into #3868, though I don't think it's the cause at all.

I also ran into this while trying to figure out #3868. Thanks for fixing it.

bradbeam (approved):

johnbelamaric (approved):

PR #3886, opened 2020-05-12
plugin/kubernetes: nil check before MetaNamespaceKeyFunc

chrisohaver:

Avoid panic and subsequent "dead" endpoints informer by checking for nil in the index key func.
This is a _workaround_ what I suspect are two upstream bugs in the k8s api libs:
* that nil endpoints are added to the informer queue in the first place
* that the panic causes the informer to stop receiving/processing new events.

I verified that this not only avoids the panic, but it also allows the informer to process missed events, and continue processing endpoint changes after the nil endpoint is encountered. IOW, it prevents the informer from dropping dead.

#3860

none

no

### DIFF
--- a/plugin/kubernetes/object/informer.go
+++ b/plugin/kubernetes/object/informer.go
@@ -1,13 +1,16 @@
 package object
 
 import (
+	"errors"
+	"reflect"
+
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/cache"
 )
 
 // NewIndexerInformer is a copy of the cache.NewIndexerInformer function, but allows custom process function
 func NewIndexerInformer(lw cache.ListerWatcher, objType runtime.Object, h cache.ResourceEventHandler, indexers cache.Indexers, builder ProcessorBuilder) (cache.Indexer, cache.Controller) {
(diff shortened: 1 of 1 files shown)


johnbelamaric:
Chris this looks good to me. Can you look into the k8s upstream issue too?

miekg:
can we reference (any) upstream issue in a comment in the code as well?

chrisohaver:
Actually - I'm thinking now that this may be a problem in our processor functions - not an upstream problem...  looking into it ...

multi-io:
While we're waiting for this or an equivalent fix to be released, can we identify the coreos version where this bug came in, and would it be advisable to temporarily downgrade to a previous version? (when using coredns as the kube-dns provider in a k8s cluster)

chrisohaver:
> Actually - I'm thinking now that this may be a problem in our processor functions - not an upstream problem... looking into it ...

Yeah - I'm more certain that this is an issue with how we handle the api object conversions. They do not handle deletion tombstones.  For now, the conversion funcs just return nil when these are encountered.  I think I see how to fix it without too much refactoring...

Closing this draft PR, there is a better fix.

chrisohaver:
> Closing this draft PR, there is a better fix.

See #3887

PR #3887, opened 2020-05-13
plugin/kubernetes: Handle endpoint tombstones

### LINKED ISSUE #3860: CoreDNS pods panic with no probe failures
**What happened**:

Recently, we have seen CoreDNS pods panic with no probe failures.  The pod logs show a panic and error yet there are no liveness nor readiness probe failures.  As a result, the pod isn't working and isn't restarted thus leading to DNS failures within the cluster.

**Pod liveness probe:**
```
    livenessProbe:
      failureThreshold: 5
      httpGet:
        path: /health
        port: 8080
        scheme: HTTP
      initialDelaySeconds: 60
      periodSeconds: 10
      successThreshold: 1
      timeoutSeconds: 5
```

**Pod readiness probe:**
```
    readinessProbe:
      failureThreshold: 3
      httpGet:
        path: /ready
        port: 8181
        scheme: HTTP
      periodSeconds: 10
      successThreshold: 1
      timeoutSeconds: 1
```

CoreDNS liveness probe detects that the CoreDNS pod has failed and restarts it.

We don't have an easy recreate scenario.  The problem may be triggered by master network connectivity flakes.

No.

**Environment**:

- the version of CoreDNS: 1.6.7 and 1.6.9
- Corefile: See below
- logs, if applicable: Not available
- OS (e.g: `cat /etc/os-release`): Ubuntu 18.04.4 LTS
- Others: Failures seen on IBM Cloud Kubernetes Service clusters running Kubernetes version 1.16 and 1.17.

```
    .:53 {
        errors
        health {
            lameduck 10s
        }
        ready
            pods insecure
            ttl 30
        }
        prometheus :9153
        forward . /etc/resolv.conf
        cache 30
        loop
        reload
        loadbalance
    }
```

chrisohaver:

`ToEndpoints()` conversion was filtering out Endpoint deletion tombstones (where the object in the event is not an `*api.Endpoint`, but rather a `cache.DeletedFinalStateUnknown`).  The tombstones are essentially indicators that the Endpoint was deleted, without a containing a full copy of the Endpoints object - just a key value.  `ToEndpoints()`  was returning a nil `*object.Endpoints` when these were encountered, which caused failures downstream when deleting from the index.

This fix is allows `ToEndpoints()` to pass the tombstones through, so they can be deleted properly from the `clientState`.

Fixes #3860

none

no

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -173,7 +173,6 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
 						case cache.Deleted:
 							apiEndpoints, ok := d.Object.(*api.Endpoints)
 							if !ok {
-								println("CTO DEBUG: GOT TOMBSTONE")
 								// Assume that the object must be a cache.DeletedFinalStateUnknown.
 								// This is essentially an indicator that the Endpoint was deleted, without a containing a
 								// up-to date copy of the Endpoints object. We need to use cache.DeletedFinalStateUnknown
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -142,11 +142,10 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
 			func(clientState cache.Indexer, h cache.ResourceEventHandler) cache.ProcessFunc {
 				return func(obj interface{}) error {
 					for _, d := range obj.(cache.Deltas) {
-						var apiEndpoints *api.Endpoints
 						var obj interface{}
-						end, ok := d.Object.(*api.Endpoints)
+						apiEndpoints, ok := d.Object.(*api.Endpoints)
 						if ok {
-							apiEndpoints, obj = object.ToEndpoints(end)
+							obj = object.ToEndpoints(apiEndpoints)
 						} else {
 							// Assume that the object is cache.DeletedFinalStateUnknown.
 							// This is essentially an indicator that the Endpoint was deleted, without a containing a full copy of the
@@ -183,6 +182,7 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
 							dns.updateModifed()
 							recordDNSProgrammingLatency(dns.getServices(obj), apiEndpoints)
 						}
+
 						if !opts.skipAPIObjectsCleanup && !reflect.ValueOf(apiEndpoints).IsNil() {
 							*apiEndpoints = api.Endpoints{}
 						}
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -141,22 +141,14 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
 			func(clientState cache.Indexer, h cache.ResourceEventHandler) cache.ProcessFunc {
 				return func(obj interface{}) error {
 					for _, d := range obj.(cache.Deltas) {
-						var obj *object.Endpoints
-						var dfsu cache.DeletedFinalStateUnknown
-
-						apiEndpoints, ok := d.Object.(*api.Endpoints)
-						if ok {
-							obj = object.ToEndpoints(apiEndpoints)
-						} else {
-							// Assume that the object is cache.DeletedFinalStateUnknown.
-							// This is essentially an indicator that the Endpoint was deleted, without a containing a full copy of the
-							// Endpoints object - just a key value. We need to use cache.DeletedFinalStateUnknown
-							// object so it can be properly deleted by store.Delete() below, which knows how to handle it.
-							dfsu = d.Object.(cache.DeletedFinalStateUnknown)
-						}
-
 						switch d.Type {
 						case cache.Sync, cache.Added, cache.Updated:
+							apiEndpoints, ok := d.Object.(*api.Endpoints)
+							if !ok {
+								return errors.New("got non-endpoint add/update")
+							}
+							obj := object.ToEndpoints(apiEndpoints)
+
 							if old, exists, err := clientState.Get(obj); err == nil && exists {
 								if err := clientState.Update(obj); err != nil {
 									return err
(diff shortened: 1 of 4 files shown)


miekg:
thanks @chrisohaver . Was this knowable at the time when we did the refactoring to make the memory go down?
I think this warrants a copy of your commit message into the source code to warn future devs - possible pointing to code and or documentation this is a thing

chrisohaver:
Looks like this may have been an issue since CoreDNS 1.2.3, when API object trimming was introduced. I would _not_ recommend reverting back to a version before 1.2.3 to avoid this bug.

This bug should be relatively uncommon to encounter, however some things that could increase occurrence:
- unstable network resulting in extended periods of disconnect between CoreDNS and k8s API.
- a short etcd compaction interval in the k8s api server

chrisohaver:
> Was this knowable at the time when we did the refactoring to make the memory go down?

Theoretically? Probably.
Reasonably?  Not sure.  client-go is complex.

> I think this warrants a copy of your commit message into the source code

Makes sense. I'll add it in `ToEndpoints()`.

plugin/kubernetes/object/endpoint.go
```diff
@@ -44,10 +44,10 @@ type EndpointPort struct {
 func EndpointsKey(name, namespace string) string { return name + "." + namespace }
 
 // ToEndpoints converts an api.Endpoints to a *Endpoints.
-func ToEndpoints(obj interface{}) (*api.Endpoints, *Endpoints) {
+func ToEndpoints(obj interface{}) (*api.Endpoints, interface{}) {
```
johnbelamaric: This is kind of a weird change. I think it would be more readable if we changed the signature to `func ToEndpoints(*api.Endpoints) (*Endpoints)`, do the casting in the caller, and just not call it if `obj` is not a `*api.Endpoints`.

johnbelamaric: Assuming as I believe `ToEndpoints` is only called once anyway.

chrisohaver: yeah - i was just making this change. :)

plugin/kubernetes/object/endpoint.go
```diff
@@ -44,12 +44,7 @@ type EndpointPort struct {
 func EndpointsKey(name, namespace string) string { return name + "." + namespace }
 
 // ToEndpoints converts an api.Endpoints to a *Endpoints.
-func ToEndpoints(obj interface{}) (*api.Endpoints, *Endpoints) {
-	end, ok := obj.(*api.Endpoints)
-	if !ok {
-		return nil, nil
-	}
-
+func ToEndpoints(end *api.Endpoints) (*api.Endpoints, *Endpoints) {
```
johnbelamaric: I still don't get why we pass in `end` and then just return it.

johnbelamaric: I mean, why do we need to return it?

chrisohaver: > I mean, why do we need to return it?

You're right.  I should do more refactoring here...

chrisohaver:
I've further jostled the code around to reduce verbosity and scope of changes.
... need to retest it to verify things still work. (checked)

zouyee:
Many controllers have examples of handling this scenario

https://github.com/kubernetes/kubernetes/blob/f975b4b0dff48c6a580593f3bc05220c70afa367/pkg/controller/disruption/disruption.go#L363-L379

chrisohaver:
> Many controllers have examples of handling this scenario

Yes, I have essentially based this change on such examples.

chrisohaver:
The latest iteration moves the delta object inspection to inside the switch cases.  I think it reads better.  Note that now it sends the (possibly outdated) endpoints object from the tombstone to `store.Delete()` instead of passing in the tombstone itself (as is done in client-go's informer processor, from where this was originally copied), but this is ok , because Delete only uses immutable key fields.

plugin/kubernetes/controller.go
```diff
@@ -163,17 +166,40 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
...
+								// up-to date copy of the Endpoints object. We need to use cache.DeletedFinalStateUnknown
+								// object so it can be properly deleted by store.Delete() below, which knows how to handle it.
+								tombstone, ok := d.Object.(cache.DeletedFinalStateUnknown)
+								if !ok {
+									return errors.New("expected tombstone")
+								}
+								if err := clientState.Delete(tombstone); err != nil {
+									return err
+								}
+								apiEndpoints, ok = tombstone.Obj.(*api.Endpoints)
```
johnbelamaric, line 184: I don't fully understand this, why do we need this part if we already deleted it with the tombstone?

chrisohaver: It hasn't been deleted yet at this point. It gets deleted later with `clientState.Delete(obj)`.  If we don't call `clientState.Delete`, it will remain in the cache.

chrisohaver:
Not sure why I have not seen this for Service objects in my tests - perhaps the API doesn't send tombstones for Service objects, just the endpoints. 

I have not tested Pods, but I suspect we'll see this issue there too.  But I want to get this PR squared away in a state that is acceptable before replicating it to Pods and Services.

plugin/kubernetes/controller.go
johnbelamaric: So what does `clientState.Delete(tombstone)` do?

chrisohaver: arg! - I left that in by accident.  Thanks for catching that. I'll remove it and run the test again.

johnbelamaric:
Ok, that looks better!

ekeih:
Thanks for your work on that @chrisohaver! This fix will make live easier for us, we recently saw this issue a lot on clusters running in Azure VMs.
I don't know how the release cycle/management of coredns usually works. Is there any way to get a patch release in the near future? Or do you release in fixed intervals?

PR #3890, opened 2020-05-15
plugin/kubernetes: handle tombstones in default processor

chrisohaver:

Like #3887, this applies tombstone handling to the DefaultProcessor (which handles Pods and Services) so we don't panic when encountering tombstones.

It also fixes terminating Pod exclusion, which in the current implementation causes a panic every time a Pod is deleted when `pods verified` is enabled.

#3887

None

No

### DIFF
--- a/plugin/kubernetes/object/informer.go
+++ b/plugin/kubernetes/object/informer.go
@@ -1,6 +1,8 @@
 package object
 
 import (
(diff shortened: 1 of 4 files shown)


johnbelamaric:
Nice!
/lgtm
/merge

Issue #3900, opened 2020-05-19
Only first upstream server is used in forward plugin


martinhaus:
**What happened**:

The policy for selecting upstream servers in the forward plugin is not working properly. Only the first upstream server is always selected no matter which policy is set. The issue seems to occur in 1.6.7 and later.

The issue is introduced with this commit https://github.com/coredns/coredns/commit/5e04c272382a7d7aca15f8ac9a8100a373ca1d1e#diff-167a31fa4e59b31f81c85cd09c92415

I assume the issue is caused by using the `List` method  `policy.go` incorrectly.

**What you expected to happen**:

Requests should be made to multiple upstream servers based on the policy which is configured.
When looking at produced metrics I expect to see multiple upstream servers.

```
coredns_forward_requests_total{to="1.1.1.1:53"} 3
coredns_forward_requests_total{to="8.8.4.4:53"} 3
coredns_forward_requests_total{to="8.8.8.8:53"} 2

```

**How to reproduce it (as minimally and precisely as possible)**:

1. Configure multiple upstream servers

```
forward . 1.1.1.1 8.8.8.8 9.9.9.9 8.8.4.4 {
      force_tcp
      policy random
      #max_fails 0
      health_check 5s
  }
```

2. Make multiple requests
```
➜  ~ dig 2.mnt.sk -p 5300 @127.0.0.1 +tcp; dig 1.mnt.sk -p 5300 @127.0.0.1 +tcp; dig 0.mnt.sk -p 5300 @127.0.0.1 +tcp; dig 4.mnt.sk -p 5300 @127.0.0.1 +tcp; dig 5.mnt.sk -p 5300 @127.0.0.1 +tcp; dig 6.mnt.sk -p 5300 @127.0.0.1 +tcp; dig 7.mnt.sk -p 5300 @127.0.0.1 +tcp;

```

3. Looking at produced metrics we can see only one upstream server is used for all of them
```
➜  ~ curl -v localhost:9153/metrics| grep coredns_forward_requests_total

coredns_forward_requests_total{to="1.1.1.1:53"} 8

```

**Anything else we need to know?**:

**Environment**:

- the version of CoreDNS: 1.6.9
- Corefile:
```
.:5300 {
  whoami
  ready :8091
  health :8080
  prometheus :9153
  loadbalance round_robin
  loop
  log . {
    class all
  }
  errors
  forward . 1.1.1.1 8.8.8.8 9.9.9.9 8.8.4.4 {
      force_tcp
[...]

PR #3901, opened 2020-05-19
Fix forward policy

martinhaus:

The policy for selecting upstream servers in the forward plugin is not working properly. Only the first upstream server is always selected no matter which policy is set. The issue seems to occur in 1.6.7 and later.

I changed the way the `List` method in `proxy.go` is used. Instead of passing a slice of structs I converted `[]*Proxy` into `[]interface{}` used the proxy selection method and the converted back to  `[]*Proxy`.

https://github.com/coredns/coredns/issues/3900

none

no

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -207,10 +207,19 @@ func (f *Forward) PreferUDP() bool { return f.opts.preferUDP }
 
 // List returns a set of proxies to be used for this client depending on the policy in f.
 func (f *Forward) List() []*Proxy {
-	if len(f.p.List(f.proxies)) == 1 {
-		return f.p.List(f.proxies)[0].([]*Proxy)
+	var interfaceSlice []interface{} = make([]interface{}, len(f.proxies))
+	for i, p := range f.proxies {
+		interfaceSlice[i] = p
 	}
-	return nil
+
+	sortedProxies := f.p.List(interfaceSlice...)
+
+	proxies := make([]*Proxy, len(sortedProxies))
+	for i, p := range sortedProxies {
+		proxies[i] = p.(*Proxy)
+	}
+
+	return proxies
 }
 
 var (
--- a/plugin/forward/setup_policy_test.go
+++ b/plugin/forward/setup_policy_test.go
@@ -1,6 +1,7 @@
 package forward
 
 import (
+	"reflect"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


chrisohaver:
Looks like the bug was introduced in PR #3537.  Prior to that, this was a 1 line function.
https://github.com/coredns/coredns/pull/3537/files#diff-b12d06e8aa6117ce05d5715e1c8930c3L191-R197

~Not immediately obvious to me why this function was changed in #3537.~

chrisohaver:
I see the same logic in GRPC too, changed in #3537.

chrisohaver (commented):
LGTM, can you add a unit test? Thanks!

martinhaus:
@chrisohaver I've added the test.

chrisohaver (approved):
LGTM, thanks!

chrisohaver (changes requested):
Actually - on second look... something seems off ...  that 2 loops are needed ...

chrisohaver:
Does the following work?

```
// List returns a set of proxies to be used for this client depending on the policy in f.
func (f *Forward) List() []*Proxy {
	if len(f.p.List(f.proxies)) == 1 {
		return f.p.List(f.proxies)[0].([]*Proxy)
	}
	ps := f.p.List(f.proxies)
	var proxies = make ([]*Proxy, len(ps))
	for i, p := range ps {
		proxies[i] = p.(*Proxy)
	}
	return proxies
}
```

chrisohaver:
@zouyee, PTAL.

martinhaus:
No, that does not work.
The problem is related to `List(p ...interface{})` 
https://github.com/coredns/coredns/blob/c9c212703a33ee6588a6b77eed71f2c7fdad577d/plugin/pkg/policy/policy.go#L23

With the configuration pasted in the related issue https://github.com/coredns/coredns/issues/3900, a call to `List` results into getting a slice of slices assigned in `p`.

```
Result of fmt.Println(p) on L24 of `List` (linked above)
[[0xc0001e2a00 0xc0001e2a40 0xc0001e2a80 0xc0001e2ac0]]
```

If you make a request with the snippet you posted above it results in
```
[ERROR] Recovered from panic in server: "dns://:5300"
```

That's why I've added the conversion from []*Proxy into []interface{} and back to resolve this.

chrisohaver:
> That's why I've added the conversion from []*Proxy into []interface{} and back to resolve this.

Ah - I see... Yes. Perhaps `forward.proxies` should be a slice of non-empty interface that `forward.Proxy` and and `grpc.Proxy` implement.  Either that or make `forward.proxies` a `[]interface{}`, and assert elements where needed - but that makes it even more opaque.

martinhaus:
@chrisohaver I'm not sure that'd work. `grpc.Proxy` and `forward.Proxy` do not have a single function in common so I'm not sure how would they be able to implement a non-empty interface.

Also if we add an interface (let's name it `ProxyInteface`) for `forward.Proxy` and make `forward.proxies` a slice of `ProxyInterface` we still cannot pass this to `policy.Sort` as `[]interface{}` without the conversion.

For the second option - making `forward.proxies` a `[]interface{}` and asserting elements everywhere seems like the incorrect approach. 

Maybe changing `policy.List` instead would be a better idea? Or using a separate implementation of the List function for forward and gprc? I'm not sure what would be the correct approach. 

Would do you think?

chrisohaver:
Since the current behavior is broken, and the proposed fix adds additional processing per request, I'm leaning toward ... 
1. for now, revert #3537 to get this fixed
2. add unit tests to verify the expected behavior (in forward and grpc)
3. then revisit #3537

Owners of forward (@johnbelamaric @miekg @rdrozhdzh), what do you think?

chrisohaver:
Closing in favor of #3919

PR #3902, opened 2020-05-20
Oss-fuzz ideal integration

catenacyber:
This is needed for oss-fuzz _ideal_ integration cf https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/

It adds 2 GitHub workflows 
- CIFuzz cf https://google.github.io/oss-fuzz/getting-started/continuous-integration/
- One workflow to check that we do not break the fuzz targets build

None

None

No

This was inspired by https://github.com/miekg/dns/pull/1091
The build fuzz targets workflow is not yet tested

### DIFF
--- a/.github/workflows/buildfuzz.yml
+++ b/.github/workflows/buildfuzz.yml
@@ -1,0 +1,25 @@
+name: Build Fuzz Targets
+on: [push, pull_request]
+jobs:
+
+  build:
+    name: Build and Test
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Set up Go 1.14
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.14
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v1
+
+    - name: Build Fuzz Targets
+      run: |
+        go get github.com/dvyukov/go-fuzz/go-fuzz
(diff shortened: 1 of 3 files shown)


miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] Oss-fuzz ideal in..." ]
>
>This is needed for oss-fuzz ideal integration cf https://google.github.io/
>oss-fuzz/advanced-topics/ideal-integration/

why is another fuzzing thing needed? What does this add, what's the benefit?

catenacyber:
> why is another fuzzing thing needed? What does this add, what's the benefit?

I would not say this is another fuzzing.
These are still the same fuzz targets
But now, the fuzz targets are also run by CIFuzz on each Pull Request when oss-fuzz runs only the master branch (soothes can prevent a bug of being merged)

Is this clearer ?

miekg:
ack thanks. I don't think that really adds any additional benefit, but it does add more things you potentially need to wait on. The current way of fuzzing is good enough imo

catenacyber:
Ok for CIfuzz.

And what about the first commit only ?
https://github.com/coredns/coredns/pull/3902/commits/db1478a69719352d041505c20454edfbb8c35747

This checks that fuzz targets keep consistent with the rest of the code

Issue #3878, opened 2020-05-07
k8s_external plugin doesn't work with AWS ELB/NLB


networkop:
When running a Kubernetes cluster on AWS (with AWS cloud provider) k8s_external plugin returns an empty response with just AUTHORITY SECTION included.

**What happened**:
I deployed a `k8s_external` plugin on an AWS-based k8s cluster, queried an external domain and got the following response:

```
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 581
;; flags: qr rd; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;my-svc.default.foo.com. IN	A

;; AUTHORITY SECTION:
foo.com.	5	IN	SOA	ns1.dns.foo.com. hostmaster.dns.foo.com. 12345 7200 1800 86400 5
```

This happens because the AWS ELB assigns a DNS alias instead of just an IP address. 

In order to differentiate between AWS and other clouds, the loadbalancer's status field looks different:

```
status:
  loadBalancer:
    ingress:
    - hostname: uuid.eu-west-2.elb.amazonaws.com
```

Compared to, for example, GKE:

```
status:
  loadBalancer:
    ingress:
    - ip: 35.123.9.321

```

**What you expected to happen**: 
I was expecting that `k8s_external` would return a CNAME that points to the `uuid.eu-west-2.elb.amazonaws.com`, e.g.

```
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 57815
;; flags: qr rd; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
; COOKIE: 1b5b8301e38cd740 (echoed)
;; QUESTION SECTION:
;my-svc.default.foo.com. IN A

;; ANSWER SECTION:
my-svc.default.foo.com. 30 IN CNAME uuid.eu-west-2.elb.amazonaws.com.
```

**How to reproduce it (as minimally and precisely as possible)**:
1. Create an EKS cluster
2. Create a service type:Loadbalancer
3. Add the following to Corefile's default zone `k8s_external foo.com`
[...]

PR #3916, opened 2020-05-27
k8s_external can now resolve CNAME returned by AWS ELB/NLB

networkop:

This PR allows `k8s_external` plugin to resolve CNAME returned by NLB/ELB when creating a service of type LoadBalancer on AWS-based clusters.

issue #3878 

None

No

### DIFF
--- a/plugin/k8s_external/external_test.go
+++ b/plugin/k8s_external/external_test.go
@@ -31,6 +31,8 @@ func TestExternal(t *testing.T) {
 		r := tc.Msg()
 		w := dnstest.NewRecorder(&test.ResponseWriter{})
 
+		e.upstream = &fakeUpstream{t, tc.Qname, fakeResponse}
+
 		_, err := e.ServeDNS(ctx, w, r)
 		if err != tc.Error {
 			t.Errorf("Test %d expected no error, got %v", i, err)
@@ -41,6 +43,7 @@ func TestExternal(t *testing.T) {
 		}
 
 		resp := w.Msg
+
 		if resp == nil {
 			t.Fatalf("Test %d, got nil message and no error for %q", i, r.Question[0].Name)
 		}
@@ -156,6 +159,8 @@ var tests = []test.Case{
 	{
 		Qname: "svc12.testns.example.com.", Qtype: dns.TypeA, Rcode: dns.RcodeSuccess,
 		Answer: []dns.RR{
+			test.A("dummy.hostname.	60 IN A 192.0.2.42"),
+			test.A("dummy.hostname.	60 IN A 192.0.2.43"),
 			test.CNAME("svc12.testns.example.com.	5	IN	CNAME	dummy.hostname"),
 		},
 	},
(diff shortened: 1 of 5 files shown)


plugin/kubernetes/object/service.go
```diff
@@ -63,6 +63,7 @@ func toService(skipCleanup bool, svc *api.Service) *Service {
 	li := copy(s.ExternalIPs, svc.Spec.ExternalIPs)
 	for i, lb := range svc.Status.LoadBalancer.Ingress {
 		s.ExternalIPs[li+i] = lb.IP
+		s.ExternalIPs[li+i] = lb.Hostname
```
chrisohaver, line 69: We should only use Hostname if IP is empty.

chrisohaver (changes requested):
Thanks! Looks good! Need to change the way IP/Hostname is retrieved from the API object, but looks good otherwise.

Please also add some unit tests.

plugin/kubernetes/object/service.go
networkop: all done

networkop:
add a couple of simple unit tests, please let me know if this is enough

plugin/kubernetes/object/service.go
```diff
@@ -62,7 +62,12 @@ func toService(skipCleanup bool, svc *api.Service) *Service {
 
 	li := copy(s.ExternalIPs, svc.Spec.ExternalIPs)
 	for i, lb := range svc.Status.LoadBalancer.Ingress {
-		s.ExternalIPs[li+i] = lb.IP
+		if &lb.IP != nil {
+			s.ExternalIPs[li+i] = lb.IP
+		} else if &lb.Hostname != nil {
+			s.ExternalIPs[li+i] = lb.Hostname
+		}
+
```
chrisohaver: pretty sure that the dereferenced string would never be nil.

```suggestion
		if lb.IP != "" {
			s.ExternalIPs[li+i] = lb.IP
			continue
		}
		s.ExternalIPs[li+i] = lb.Hostname
```

plugin/k8s_external/external_test.go
```diff
@@ -147,21 +147,33 @@ var tests = []test.Case{
...
+	{
+		Qname: "svc11.testns.example.com.", Qtype: dns.TypeA, Rcode: dns.RcodeSuccess,
+		Answer: []dns.RR{
+			test.A("svc11.testns.example.com.	5	IN	A	1.2.3.4"),
+		},
+	},
+	{
+		Qname: "svc12.testns.example.com.", Qtype: dns.TypeA, Rcode: dns.RcodeSuccess,
+		Answer: []dns.RR{
+			test.CNAME("svc12.testns.example.com.	5	IN	CNAME	dummy.hostname"),
```
chrisohaver, line 160: We should set the test up to expect an `A` record for `dummy.hostname` in the answer also.

chrisohaver (changes requested):

plugin/kubernetes/object/service.go
networkop: yep, not sure what i was doing here. fixed now

plugin/k8s_external/external_test.go
networkop: I've borrowed some code from dns64 testing and added it to k8s_external. Is this what you had in mind?

plugin/kubernetes/object/service.go
```diff
@@ -62,7 +62,12 @@ func toService(skipCleanup bool, svc *api.Service) *Service {
 
 	li := copy(s.ExternalIPs, svc.Spec.ExternalIPs)
 	for i, lb := range svc.Status.LoadBalancer.Ingress {
-		s.ExternalIPs[li+i] = lb.IP
+		if lb.IP == "" {
```
chrisohaver:
```suggestion
		if lb.IP != "" {
```

plugin/k8s_external/external_test.go
chrisohaver: Actually, I think just leave it out for now. I'd rather not introduce an interface just for testing purposes.  Sorry about that.

chrisohaver (changes requested):

networkop:
yep, all good. rolled back to the original test with no upstream.

chrisohaver (commented):
Looks good!  Before this merges, can you give it a quick sanity test in an AWS cluster with name-based LBs?  Just to make sure it works as expected. Thanks.

chrisohaver (approved):

networkop:
Tested this change on a few of AWS, GCP and private (Metallb) clusters and can confirm that it worked as expected in all of these cases.

PR #3920, opened 2020-05-30
Add timeouts for http server

milgradesec:
Currently the http.Server created for DoH does not have timeouts, it can lead to file descriptors leakage and DoS attacks by malicious clients.
Values from https://blog.cloudflare.com/exposing-go-on-the-internet/
No
No
No

### DIFF
--- a/core/dnsserver/server_https.go
+++ b/core/dnsserver/server_https.go
@@ -40,7 +40,12 @@ func NewServerHTTPS(addr string, group []*Config) (*ServerHTTPS, error) {
 		tlsConfig = conf.TLSConfig
 	}
 
-	sh := &ServerHTTPS{Server: s, tlsConfig: tlsConfig, httpsServer: new(http.Server)}
+	srv := &http.Server{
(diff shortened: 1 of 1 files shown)


yongtang:
The connection timeout value that are used, are too much tied to the underlying networking infrastructure (unless the traffic is point to point through the internet directly). In many cases, traffic need to go through different layers of load balancers, firewalls, applications security devices, and sometimes self-configured routers/switches (in internal or vpn). Hardcoded values could lead to many hard-to-trouble-shoot issues.

If timeout is needed then the value ideally should be adjustable.

miekg:
we could make them large enough that we have timeouts (so resources get released), and then figure out how to tune. I think server side timeout (with the multitude) of clients out there are very hard to tune runtime - most values we like to auto-tune in CoreDNS are client side - i.e. us dealing with some upstream, so we only need to deal with 1 network path. I think this can be merged as is

miekg:
i dont have better ideas then to just merge this now

PR #3923, opened 2020-06-02
Use nonroot distroless base image

sozercan:
Uses non root distroless minimal base image that already has ca-certificates included

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,7 @@
-FROM debian:stable-slim
+FROM gcr.io/distroless/static:nonroot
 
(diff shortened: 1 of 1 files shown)


sozercan:
Looks like there is an issue due to binding to privileged port 53.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Use nonroot d..." ]
>Looks like there is an issue due to binding to privileged port 53.

yes, this cannot work. There have been various issue to bind to a higher port, but that
creates a backward incompatible docker image.

/close

Issue #3879, opened 2020-05-08
Data is not synced when CoreDNS reconnects to kubernetes api server after protracted disconnection


GsssC:
**What happened**:
For some reason, our project needs to deploy coreDNS away from k8s apiserver , at the same time the network is not reliable. Therefore, we tested whether the event of k8s SVC when the network was disconnected would be detected by CoreDNS after reconnetcion. The result showed that when SVC was deleted during a long period of disconnection from k8s API server, and CoreDNS still had the parsing record of the SVC after reconnection.
**What you expected to happen**:
When CoreDNS is disconnected from API server, SVC is deleted. After reconnection, CoreDNS should not have the parsing record of this SVC
**How to reproduce it (as minimally and precisely as possible)**:
Virtural Machine a : run kubernetes api server
Virtural Machine b : run CoreDNS by binary+systemed
- Short time network failure
1. In a, use kubectl to create a svc relating existing pod
2. In b, use dig  to resolve svc `dig svc.namespace.service.cluster.local @debianIP -p 1053`, successfully return svc vip
3. In b, **shut down** the network of b
4. In a, use kubectl delete svc
5. In a, use kubectl create svc2
6. In b, **resume** the network of b after 1 min
7. In b, use dig to resolve svc , successfully return none
8. Inb, use dig to resolve svc2, successfully return svc2 vip
- Long time network failure
1. In a, use kubectl to create a svc
2. In b, use dig to resolve svc, successfully return svc vip
3. In b, **shut down** the network of b
4. In a, use kubectl delete svc
5. In a, use kubectl create svc2
6. In b, **resume** the network of b after 30 min
7. In b, use dig to resolve svc , **still return svc vip**
8. In b, use dig to resolve svc2, successfully return svc2 vip
**Anything else we need to know?**:
Could this be due to a design flaw in the ListWatch itself?

**Environment**:
- the version of CoreDNS:
- Corefile:
```
.:1053 {
        errors
        health {
           lameduck 5s
        }
        ready
        kubernetes cluster.local in-addr.arpa ip6.arpa {
           endpoint http://10.10.103.98:8080
           pods insecure
           fallthrough in-addr.arpa ip6.arpa
           ttl 0
        }
        forward . /etc/resolv.conf
        loop
        reload
        loadbalance
    }
```
- logs, if applicable:
- OS (e.g: `cat /etc/os-release`):
- Others:

miekg:
[...]

PR #3924, opened 2020-06-03
plugin/kubernetes: fix tombstone unwrapping

chrisohaver:

This corrects the handling of tombstones - specifically that the objects imbedded in the tombstone deltas are actually `coredns/object` types not `k8s/api` (as in normal delete events).  The delete case in the processor now makes no type assertion on the object in the tombstone, and simply passes the tombstone itself to `index.Delete()`.

**Unit test**:  coverage added for the DefaultProcessor for adds, updates, "normal" deletes, and "tombstone" deletes.

**Refactoring**: Consolidated the in-line Endpoints Process function and the DefaultProcessor so all watches now use the DefaultProcessor.

#3879 
#3860 
PRs #3887 and #3890

none

no

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -153,6 +153,7 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
 								if err := clientState.Update(obj); err != nil {
 									return err
 								}
+								h.OnUpdate(old, obj)
 								// endpoint updates can come frequently, make sure it's a change we care about
 								if !endpointsEquivalent(old.(*object.Endpoints), obj) {
 									dns.updateModifed()
@@ -162,6 +163,7 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
 								if err := clientState.Add(obj); err != nil {
 									return err
 								}
+								h.OnAdd(d.Object)
 								dns.updateModifed()
 								recordDNSProgrammingLatency(dns.getServices(obj), apiEndpoints)
 								if !opts.skipAPIObjectsCleanup {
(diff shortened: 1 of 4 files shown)


plugin/kubernetes/controller.go
```diff
@@ -125,7 +125,7 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
 			&api.Pod{},
 			cache.ResourceEventHandlerFuncs{AddFunc: dns.Add, UpdateFunc: dns.Update, DeleteFunc: dns.Delete},
 			cache.Indexers{podIPIndex: podIPIndexFunc},
-			object.DefaultProcessor(object.ToPod(opts.skipAPIObjectsCleanup)),
+			object.DefaultProcessor(object.ToPod(opts.skipAPIObjectsCleanup), nil),
```
miekg, line 128: this boolean arg is is only used for a single test? that seems excessive for adding it?

plugin/kubernetes/controller.go
```diff
@@ -491,12 +432,16 @@ func (dns *dnsControl) detectChanges(oldObj, newObj interface{}) {
...
+		if !endpointsEquivalent(oldObj.(*object.Endpoints), newObj.(*object.Endpoints)) {
+			dns.updateModifed()
+		}
 	default:
 		log.Warningf("Updates for %T not supported.", ob)
 	}
 }
 
-func (dns *dnsControl) getServices(endpoints *object.Endpoints) []*object.Service {
+func (dns *dnsControl) getServices(endpoints *api.Endpoints) []*object.Service {
```
miekg, line 444: why is this now an api.Endpoint instead of object?

plugin/kubernetes/object/endpoint.go
```diff
@@ -43,8 +45,19 @@ type EndpointPort struct {
 // EndpointsKey return a string using for the index.
 func EndpointsKey(name, namespace string) string { return name + "." + namespace }
 
-// ToEndpoints converts an api.Endpoints to a *Endpoints.
-func ToEndpoints(end *api.Endpoints) *Endpoints {
+// ToEndpoints returns a function that converts an *api.Endpoints to a *Endpoints.
+func ToEndpoints(skipCleanup bool) ToFunc {
```
miekg, line 49: Where does the ToFunc requirement come from? Just testing?

plugin/kubernetes/object/informer.go
```diff
@@ -20,8 +20,10 @@ func NewIndexerInformer(lw cache.ListerWatcher, objType runtime.Object, h cache.
 	return clientState, cache.New(cfg)
 }
 
-// DefaultProcessor is a copy of Process function from cache.NewIndexerInformer except it does a conversion.
-func DefaultProcessor(convert ToFunc) ProcessorBuilder {
+type recordLatencyFunc func(interface{})
```
miekg: empty interface? why?

plugin/kubernetes/controller.go
chrisohaver: The boolean arg was already there. IIUC, it was added when the record latency metrics were added, to allow the test to work.

The new parameter is a function that calculates the record latency metrics for that object type.  Currently it's only implemented for the Endpoints object.  So, for Services and Pods, we pass nil.

plugin/kubernetes/object/endpoint.go
chrisohaver: This is required to allow the Endpoints watch to use the default processor, instead of the current in-line processor. (it is similar in structure to the Pod and Service objects, which also use the default processor)

plugin/kubernetes/object/informer.go
chrisohaver: Ah, yes, I think I could tighten it down to `meta.Object`, which is implemented by all k8s api objects (e.g. Service, Pod, Endpoints, )...

plugin/kubernetes/controller.go
chrisohaver: `GetServices` can construct the index key from either type. Changing it to `api.Endpoint ` allows us to pass a single argument to `recordDNSProgrammingLatency()` instead of two (both `api.Endpoint` and `object.Endpoint` for the same record).

plugin/kubernetes/object/informer.go
chrisohaver: > I could tighten it down to meta.Object

Done in 8aabd9c

chrisohaver:
@johnbelamaric PTAL

luks:
Hi guys, i am wonder if this fix will be in release 1.7.0, we are having same problem, we have also coredns separated on masters node outside of k8s.

chrisohaver:
> Hi guys, i am wonder if this fix will be in release 1.7.0, we are having same problem, we have also coredns separated on masters node outside of k8s.

Yes. It will.

luks:
Super :), thank you

Issue #3930, opened 2020-06-08
weight for SRV records should be at least 1


yangbor:
**What happened**:
The SRV records returned by coredns in kubernetes cluster is 0 when number of pods for a service is greater than 100.

**What you expected to happen**:
The weight should be at least 1.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

- the version of CoreDNS:  1.6.2
- Corefile:
- logs, if applicable:
- OS (e.g: `cat /etc/os-release`):
- Others:

PR #3931, opened 2020-06-08
Fix 0 weight in SRV records with 100 more records in answer

yangbor:
Currently the SRV records' weight is 0 if the number of endpoints
is greater than 100. This will cause the upstream lb(haproxy e.g.)
using SRV records for service-discovery to mark the servers as
unavailable due to 0 weight.

The fix is to use math.Ceil instead of math.Floor for weight
calculation to make sure that the weight is at least 1.

#3930

none

none

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -185,6 +185,10 @@ func SRV(ctx context.Context, b ServiceBackend, zone string, state request.Reque
 			w1 *= float64(serv.Weight)
 		}
 		weight := uint16(math.Floor(w1))
+		// weight should be at least 1
+		if weight == 0 {
+			weight = 1
+		}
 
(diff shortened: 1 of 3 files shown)


yangbor:
Hi team, the circleci failed due to weight changed from 36 to 37. But I can't find the corresponding test code to fix.

chrisohaver:
`Math.Ceil` changes behavior for non-zero weights (hence the test failure).
If you check to see if weight is 0, and force it to 1, then other tests won't be affected.

yangbor:
@chrisohaver  Thanks, I updated the PR according to your suggestion.

chrisohaver (approved):
LGTM.  AFAICT, the backend does not have a facility for intentionally creating SRV records with zero weight (it defaults zero weights to 100). So this seems safe.

Issue #2465, opened 2019-01-15
plugin/k8s_external: Support Ingress-resources


voidengineer:
Since 1.3.0 CoreDNS supports exposing k8s LB services with the k8s_external plugin.

It would be great to support also ingress-resources (with k8s_external or a new plugin).
An ingress-resources spec contains a list of rules (each with a hostname) and its status can contain the associated ip-addresses or hostnames.

johnbelamaric:
See also https://groups.google.com/d/topic/coredns-discuss/7CBvHRgENg0/discussion

chrisohaver:
Ingress objects wont always all sit in the same zone.  Before starting on this, we should get consensus and approval up front on how this will be handled in an acceptable way.

This could be done as an extension of _k8s_external_, but doing so would require that we treat certain zones in the plugin's zone list with special/different significance.  While this is possible to do, that kind of pattern has been rejected in prior plugin PRs.  So I think we need to do this as a separate plugin.

As a separate plugin, one could list each ingress domain name in the plugin's zone list.  This could lend itself to a long zone list. These zones are not strictly "zones"... they are FQDNs, which doesn't fit snuggly with how the zone routing works in CoreDNS. In the example below, the plugin would process queries `moo.foo.example.com` because it's in the zone `foo.example.com`.

```
k8s_ingress foo.example.com bar.example.org baz.example.net {
    fallthrough
}
```

The `fallthrough` directive is included here because like the _host_ plugin, _k8s_ingress_ would not be authoritative for the zones (e.g. we'd need to fall through to next plugin for `moo.foo.example.com`).

One could have the plugin process everything, and fall through when the query does not match an ingress host ... Cleaner configuration, perhaps a bit more overhead (but i don't think much).  

```
k8s_ingress . {
    fallthrough
}
```

From this perspective, zones are still zones, and someone could selectively expose ingress names falling within certain zones if required.

voidengineer:
I think an additional plugin `k8s_ingress` makes sense. Most likely someone will ask for `k8s_istio_gateway` in the near future.

Maybe it should be possible to restrict the kubernetes namespaces where the ingress resources are searched and to specifiy the plugin multiple times.

```
k8s_ingress qa.example.org {
  namespace qa
  fallthrough
}

k8s_ingress production.example.org {
  namespace production
  fallthrough
}
```

chrisohaver:
I think that would be equivalent to ... 
```
k8s_ingress qa.example.org production.example.org {
[...]

PR #3935, opened 2020-06-08
Plugin/k8s_ingress

networkop:
This PR introduces a new `k8s_ingress` plugin to support external resolution of Ingress domain names, similar to what `k8s_external` does for services.

Issue #2465 

The new plugin readme (part of the PR)
No, however this would require a new set of RBAC rules, for example:
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
  name: excoredns-coredns
rules:
- apiGroups:
  - ""
  resources:
  - endpoints
  - services
  - pods
  - namespaces
  verbs:
  - list
  - watch
- apiGroups:
  - extensions
  resources:
  - ingresses
  verbs:
  - list
  - watch
```


networkop:
oops, I didn't know there were other CI tests in a separate repo. I'm happy to fix them, once there's a consensus on this PR.

chrisohaver:
Thanks for the PR! I have not done a full review yet.  Initial thoughts:
* There is a significant amount of code duplication with _k8s_external_. So that has me pause to think on how we could do things differently.  For example, how to more easily allow other plugins to "piggy-back" on the kubernetes API connection.
* This appears to cause the _kubernetes_ plugin to maintain a watch on ingresses, even if _k8s_ingress_ is not in use (i.e. affecting all kubernetes users ).  The Kubernetes scalability team is especially sensitive to bumps in memory, and and pressure on the API server.

networkop:
Thanks for your feedback!

> * There is a significant amount of code duplication with _k8s_external_. So that has me pause to think on how we could do things differently.  For example, how to more easily allow other plugins to "piggy-back" on the kubernetes API connection.

This is exactly what I thought as well. Not only that, but to make things work between the two plugins, one has to fallthrough to the next when there's no match found, instead of returning zone apex, like [this](https://github.com/coredns/coredns/pull/3935/files#diff-09e5fe140b7035bc7950cfc93ee6555dR107)


> * This appears to cause the _kubernetes_ plugin to maintain a watch on ingresses, even if _k8s_ingress_ is not in use (i.e. affecting all kubernetes users ).  The Kubernetes scalability team is especially sensitive to bumps in memory, and and pressure on the API server.

That is true, and this can be an argument against merging with k8s_external. 
I'm happy to investigate either option.

networkop:
hi @chrisohaver :wave: 
did you have any further thoughts or feedback on this PR?

chrisohaver:
> did you have any further thoughts or feedback on this PR?

I'll try to carve out time next week to review it in more detail.

plugin/k8s_ingress/README.md
```diff
@@ -0,0 +1,93 @@
...
+				5          ; minimum (4 hours)
+				)
+example.org		5 IN	NS ns1.dns.example.org.
+
+ns1.dns.example.org.  5 IN  A    ....
+ns1.dns.example.org.  5 IN  AAAA ....
+~~~
+
+Note that we use the `dns` subdomain for the records DNS needs (see the `apex` directive). Also
+note the SOA's serial number is static. The IP addresses of the nameserver records are those of the
```
chrisohaver, line 32: Static serial: This should probably move to a "Known Issues" section... since its not a desired behavior.

plugin/k8s_ingress/README.md
```diff
@@ -0,0 +1,93 @@
...
+CoreDNS service.
```
chrisohaver, line 33: > The IP addresses of the nameserver records are those of the CoreDNS service.

The Cluster IP of the service?  For clients outside fo the cluster, this would be an unreachable IP address.

plugin/k8s_ingress/README.md
```diff
@@ -0,0 +1,93 @@
...
+    ttl TTL
+}
+~~~
+
+* **APEX** is the name (DNS label) to use for the apex records; it defaults to `dns`.
+* `ttl` allows you to set a custom **TTL** for responses. The default is 5 (seconds).
+
+## Examples
+
+Enable names under `example.org` to be resolved to in-cluster DNS addresses.
```
chrisohaver, line 61: Can you rephrase this? It doesn't seem to describe what the example does?

plugin/k8s_ingress/README.md
```diff
@@ -0,0 +1,93 @@
...
+## Name
+
+*k8s_ingress* - resolves ingress FQDNs from outside Kubernetes clusters.
+
+## Description
+
+This plugin allows an additional zone to resolve the external IP address(es) of a Kubernetes
+Ingresses, similar to what `k8s_external` does for services. This plugin is only useful if the *kubernetes* plugin is also loaded.
+
+The plugin uses an external zone to resolve in-cluster IP addresses. It only handles queries for A,
```
chrisohaver, line 12:
```suggestion
It only handles queries for A,
```

plugin/k8s_ingress/README.md
```diff
@@ -0,0 +1,93 @@
+# k8s_ingress
+
+## Name
+
+*k8s_ingress* - resolves ingress FQDNs from outside Kubernetes clusters.
+
+## Description
+
+This plugin allows an additional zone to resolve the external IP address(es) of a Kubernetes
+Ingresses, similar to what `k8s_external` does for services. This plugin is only useful if the *kubernetes* plugin is also loaded.
```
chrisohaver, line 10:
```suggestion
This plugin creates records for Kubernetes Ingresses.  This plugin is only useful if the *kubernetes* plugin is also loaded.
```

plugin/k8s_ingress/external.go
```diff
@@ -0,0 +1,115 @@
+/*
+Package ingress implements ingress names for kubernetes clusters.
+
+This plugin only handles three qtypes (except the apex queries, because those are handled
+differently). We support A, AAAA and SRV request, for all other types we return NODATA or
```
chrisohaver, line 5: SRV records don't appear to be supported in this PR.

plugin/k8s_ingress/external.go
```diff
@@ -0,0 +1,115 @@
...
+NXDOMAIN depending on the state of the cluster.
+
+A plugin willing to provide these services must implement the Externaler interface, although it
```
chrisohaver, line 8: Externaler -> Ingresser

plugin/k8s_ingress/external.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/coredns/coredns/plugin/etcd/msg"
+	"github.com/coredns/coredns/plugin/pkg/upstream"
+	"github.com/coredns/coredns/request"
+
+	"github.com/miekg/dns"
+)
+
+// Ingresser defines the interface that a plugin should implement in order to be used by External.
+type Ingresser interface {
+	// External returns a slice of msg.Services that are looked up in the backend and match
```
chrisohaver, line 27: Copy-pasted comments don't match property names in this interface declaration.

plugin/k8s_ingress/external.go
```diff
@@ -0,0 +1,115 @@
...
+func New() *Ingress {
+	e := &Ingress{hostmaster: "hostmaster", ttl: 5, apex: "dns"}
+	return e
+}
+
+// ServeDNS implements the plugin.Handle interface.
+func (e *Ingress) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	state := request.Request{W: w, Req: r}
+
+	//log.Printf("Service query %s", state.Name())
```
chrisohaver, line 59: These commented-out debugs should be removed before merging.

plugin/k8s_ingress/external.go
```diff
@@ -0,0 +1,115 @@
...
+	switch state.QType() {
+	case dns.TypeA:
+		m.Answer = e.a(ctx, svc, state)
+	case dns.TypeAAAA:
+		m.Answer = e.aaaa(ctx, svc, state)
+	default:
+		m.Ns = []dns.RR{e.soa(state)}
+	}
+
+	// If we did have records, we have to continue to let other plugins pick it up.
```
chrisohaver, line 105: missing "not" ? "If we did _not_ have records," ...

This is implicit "fallthrough" behavior, which we should avoid I think.  Instead, add a "fallthough" option to the plugin, and continue down the plugin chain only if fallthough is enabled (see other plugins: kubernetes, hosts).  Also, to be consistent with other fallthrough implementations, we should only be falling through when the result is NXDOMAIN.  e.g. if someone does a TXT query on name that exists, we should return NODATA instead of allowing another plugin to answer.

plugin/k8s_ingress/external.go
```diff
@@ -0,0 +1,115 @@
...
+		return plugin.NextOrFailure(e.Name(), e.Next, ctx, w, r)
+	}
+
+	if e.externalFunc == nil {
+		return plugin.NextOrFailure(e.Name(), e.Next, ctx, w, r)
+	}
+	//log.Printf("Found matching zone %s", zone)
+	state.Zone = zone
+	for _, z := range e.Zones {
+		// TODO(miek): save this in the External struct.
```
chrisohaver, line 71: I assume this TODO is copy-pasted?  Does it still apply in this context?

plugin/k8s_ingress/external.go
```diff
@@ -0,0 +1,115 @@
+/*
```
chrisohaver, line 1: Should this file be named `ingress.go`?  is `external.go` a remanent of k8s_external?

plugin/k8s_ingress/msg_to_dns.go
```diff
@@ -0,0 +1,105 @@
...
+				rr := s.NewAAAA(state.QName(), ip)
+				rr.Hdr.Ttl = e.ttl
+				records = append(records, rr)
+			}
+		}
+	}
+	return records
+}
+
+// not sure if this is even needed.
```
chrisohaver, line 81: Please find out, and remove if not needed.

plugin/kubernetes/controller.go
```diff
@@ -116,6 +124,17 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
 		object.DefaultProcessor(object.ToService(opts.skipAPIObjectsCleanup)),
 	)
 
+	dns.ingLister, dns.ingController = object.NewIndexerInformer(
```
chrisohaver, line 127: This informer must be conditionally created, like the pod informer.
When k8s_ingress is not in use, kubernetes should not establish a watch on ingress objects.

plugin/k8s_ingress/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+The plugin uses an external zone to resolve in-cluster IP addresses. It only handles queries for A,
+AAAA; all others result in NODATA responses. To make it a proper DNS zone, it handles SOA and NS queries for the apex of the zone.
```
chrisohaver, line 13: Even though it's implied, probably should call out specifically that PTR requests (reverse lookups) are not handled.

plugin/kubernetes/controller.go
```diff
@@ -21,13 +22,18 @@ import (
...
+	ingHostIndex          = "IngressHost"
 	epNameNamespaceIndex  = "EndpointNameNamespace"
 	epIPIndex             = "EndpointsIP"
 )
 
 type dnsController interface {
 	ServiceList() []*object.Service
+	IngressList() []*object.Ingress
+	IngIndex(string) []*object.Ingress
+	IngIndexReverse(string) []*object.Ingress
```
chrisohaver, line 36: This isn't ideal - that we have to modify the controller interface like this.
Perhaps there is another approach...

chrisohaver (changes requested):
@networkop, I spent a few hours going over this.  IMO there is too much code (and comment) duplication for this to be feasibly included as an internal plugin.

chrisohaver:
An alternate approach could be to establish a separate connection and watch on the API.  I think the k8s client-go consolidates connections to minimize overhead.

miekg:
please send a PR to coredns/coredns.io where this can live as an external plugin. ta.

networkop:
@chrisohaver thanks for taking time to review this. I also felt there was too much code duplication and changes to other modules/plugins. This was a "naive" implementation where I tried to follow the current plugin (k8s_external) as much as possible. The goal was to open a discussion and get your opinion about what needs to change. So the main two issues with the logic (excluding comments/readme) are:
1. Modifying the k8s controller interface
2. Wrong fallthrough behaviour

I will try to refactor the code to make sure all k8s controller code is pulled into the ingress plugin itself, so it becomes more self-contained. 
However, this still would require changes to RBAC and, consequently, helm charts. So if you think that there's no point in continuing with this PR here, I'm happy work on it externally and, as @miekg suggested, simply list it as one of the external plugins.

chrisohaver:
The fallthrough behavior is easily fixable.  Avoiding the interface changes less so.  

The k8s controller code is somewhat complex due to a requirement of us keeping under a target memory limit set by Kubernetes.  E.g. we use customized indexer informers and do object conversions. Copying that structure verbatim would carry a lot of bulk and complexity you don't need - you should be able to use the default informers included in client-go.

Another other issue is the "always-on" watch on ingresses, even if k8s_ingress is not enabled in the corefile.
With that always on,  the RBAC rules would need to be updated.  However, if ingresses are only watched if k8s_ingress is enabled, then the RBAC rules only need to permit ingress list/watch when k8s_ingress is enabled.  Since k8s_ingress would not be enabled by default, the RBAC rules don't need to allow it by default.  If a user enables k8s_ingress, they would need to edit the RBAC rules.  This could be described in the README.  In the helm chart, if k8s_ingress is added as a configurable option, then it would need to adjust RBAC rules accordingly.

I think for now this should be an external plugin, and refactored to not require changes to the kubernetes plugin controller interface.

In the future, it would be nice for kubernetes plugin to make the informers/indexers generally available to other plugins (i.e. a way for other plugins to add informer/indexers to the controller or access existing ones).  If we work that out, and k8s_ingress proves to be a popular feature, we can consider integrating it into the tree as an internal plugin.

networkop:
Understood. I'll close this PR then. I'll update the issue #2465 if/when the external plugin is ready.

PR #3936, opened 2020-06-08
plugin/dnssec: fail if key configured twice

JensErat:

If a key is configured twice, the plugin will actually sign records twice. This commit adds two test cases (in especially the one I stumbled into, where I registered both public and private keys  separately) and a check returning an error if a key is registered twice.

The very flexible key configuration invites of putting both public and private keys in there. Alternatively, we might consider simply merging duplicate keys when loading them.

Please advise whether this might be a breaking change or requires changelog entries.

In case of existing but undetected misconfiguration, CoreDNS will fail.

### DIFF
--- a/plugin/dnssec/setup.go
+++ b/plugin/dnssec/setup.go
@@ -115,6 +115,19 @@ func dnssecParse(c *caddy.Controller) ([]string, []*DNSKEY, int, bool, error) {
 		}
 	}
 
+	// check if a key is configured multiple times
+	seen := make(map[string]bool)
+	for _, k := range keys {
+		// compare based on DNSKEY resource record that already includes key type, using same key for both KSK and ZSK
+		// is not an issue on CoreDNS side
+		record := k.K.String()
+		_, found := seen[record]
+		if found {
+			return zones, keys, capacity, splitkeys, fmt.Errorf("key %s (keyid: %d) loaded twice", k.K.Header().Name, k.tag)
+		}
+		seen[record] = true
+	}
+
 	return zones, keys, capacity, splitkeys, nil
 }
 
--- a/plugin/dnssec/setup_test.go
+++ b/plugin/dnssec/setup_test.go
@@ -78,6 +78,18 @@ func TestSetupDnssec(t *testing.T) {
 				key file ksk_Kcluster.local
 			}`, false, []string{"cluster.local."}, nil, true, defaultCap, "",
 		},
+		{
+			`dnssec cluster.local {
+				key file Kcluster.local
+                key file Kcluster.local
+			}`, true, []string{"example.org."}, nil, false, defaultCap, "loaded twice",
+		},
+		{
+			`dnssec cluster.local {
+				key file Kcluster.local.private
+                key file Kcluster.local.key
+			}`, true, []string{"example.org."}, nil, false, defaultCap, "loaded twice",
+		},
 	}
 
 	for i, test := range tests {


miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin/dnssec: fa..." ]
>
>If a key is configured twice, the plugin will actually sign records twice. This
>commit adds two test cases (in especially the one I stumbled into, where I
>registered both public and private keys separately) and a check returning an
>error if a key is registered twice.

How did you end up adding a key twice?

JensErat:
> How did you end up adding a key twice?

I did pretty much exactly what is described in the [second unit test](https://github.com/coredns/coredns/pull/3936/files#diff-e70f119dbc3de717ff0622808ec2a4e2R88-R90) with different names:

```
dnssec cluster.local {
  key file Kcluster.local.private
  key file Kcluster.local.key
}
```

Why I did this -- don't know. Not properly reading the docs. Looked reasonable, and it worked out. The keys have been in place with some bind server before, and bind requires configuring both. I told config management to dump all the files and add records in there. Probably all of those potential reasons together, and I didn't even realize something's wrong for more than a year until I digged into DNSSEC transfer issues and the transfers looked weird (which will probably bring another one or two issues and/or pull requests, but not quite sure what's wrong yet).

What I'm sure about is that providing such a configuration is not something CoreDNS likes, and I'd love if CoreDNS either fails clearly or can deal with such a configuration. I'm also fine with putting together code that actually merges duplicate keys instead, but I guess telling the user something is wrong is the better approach (did he maybe acutally forget another key instead?).

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/dnssec..." ]
>    How did you end up adding a key twice?
>
>I did pretty much exactly what is described in the second unit test with
>different names:
>
>dnssec cluster.local {
>  key file Kcluster.local.private
>  key file Kcluster.local.key
>}

ah, fair. Yes, this is due to be helpful and figure out what keys are actually meant to be
used. Wonder what we should do, merging the keys sounds like the wrong thing to do, so I'm
leaning to erroring. BUT we also (not implemented) have a key *directory* where you can
just drop keys and they will be picked up; log.Fataling there is a bit nasty because you
will crash a fully running coredns process. Hence I think we should merge those keys and
do a log.Debug/Warningf or something.

/plugin dnssec
/label enhancement

JensErat:
I cannot find a reference to this per-directory configuration in the [`dnssec` plugin docs](https://coredns.io/plugins/dnssec/) (haven't checked the code though) -- can you point me towards something regarding that option? How dynamic is reconfiguration of that directory (just considering use cases where config management rolls keys or something like that with intermediate copies, whatever might happen here).

So it seems we have two different cases with several options. Mostly for myself to get an overview:

| Configuration  | use duplicates | merge-and-ignore | merge-and-warn | error-out | "deprecation" warn -> err | 
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| per key file | :x: (current behavior) | :x: | :x: | :heavy_check_mark: | :grey_question: |
| directory  | :x: (current behavior)  | :x: | :white_check_mark: | :x:  |  :grey_question: |

My feeling was that there is mostly agreement for the err-out approach for file-based configuration and some preference for merging and warning (but not silently merging) in case of per-directory configuration.

Still not sure if I like the deprecation path anyway, but at least it might prove a way to handle both cases the same way (given there is CoreDNS handler picking up changes very fast, like observing the file system).

If there is consent, I will gladly get hands-on and extend the PR.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/dnssec..." ]
>If there is consent, I will gladly get hands-on and extend the PR.

consent is between you and me (and anyone else that may have an opinion). Doing the merge
seems most logical to me.

JensErat:
I also grepped through the code again, and could not find anything that looks like "reading keys from a directory" (although I must admit just having grepped for some keywords and looked at potentially relevant plugins, and _key_ is a pretty unspecific _key_word...). Can you point me towards some docs or code where keys are loaded from a directory?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/dnssec..." ]
>I also grepped through the code again, and could not find anything that looks
>like "reading keys from a directory" (although I must admit just having grepped
>for some keywords and looked at potentially relevant plugins, and key is a
>pretty unspecific _key_word...). Can you point me towards some docs or code
>where keys are loaded from a directory?

this not implemented by hinted in the README (I think)

miekg:
can you change this to drop identical keys from the list, possibly with a log.Warningf

miekg:
this should be redone as merging the keys. tx

Issue #3938, opened 2020-06-10
forward: TCP upstreams are health checked with UDP


pnovotnak:
**What happened**:

I have a DNS upstream that only responds to TCP requests. Unfortunately, coredns `forward` health checks always fail, it looks like the cause is that [CoreDNS only tries UDP](https://github.com/coredns/coredns/blob/master/plugin/forward/health.go#L32) ([it can't differentiate between TCP DNS and UDP DNS](https://github.com/coredns/coredns/blob/master/plugin/pkg/transport/transport.go#L5)). The most unfortunate side-effect of this bug is that even when `policy round_robin` is used to work around [a bug](https://github.com/coredns/coredns/issues/3900) in `policy random` coredns then falls back to a broken `random` policy.

**What you expected to happen**:

Health checks to TCP upstreams should be checked via TCP protocol.

**How to reproduce it (as minimally and precisely as possible)**:

Create a forwarder that forwards requests to an upstream that only has a TCP listener. Observe health check metrics and upstream request rates.

**Anything else we need to know?**:

**Environment**:

- the version of CoreDNS: kubernetes node-cache v1.15.13 (coredns v1.6.7).
- Corefile:
- logs, if applicable:

        $ k logs -n kube-system node-local-dns-22csh | grep healthchecker
        $ echo $?
        1
- OS (e.g: `cat /etc/os-release`):
- Others:

pnovotnak:
This is not a valid bug, actually. Just caught a query on the wire and it's TCP – the query itself is failing:

<img width="1065" alt="Screen Shot 2020-06-10 at 4 09 47 PM" src="https://user-images.githubusercontent.com/5713245/84327710-ea8d4f80-ab34-11ea-940f-dbc68de5297e.png">

pnovotnak:
I thought I was mistaken, but I managed to capture some unanswered UDP NS queries 
<img width="1098" alt="Screen Shot 2020-06-10 at 5 30 06 PM" src="https://user-images.githubusercontent.com/5713245/84333269-24198700-ab44-11ea-9656-445bb0b11081.png">

pnovotnak:
Here are some graphs of metrics collected from cache instances. We have 3 upstream IPs for our DNS infra (server is in GMT).

Response volume by code across all three servers. You can see the results of a config reload at around 12:00AM GMT – things are OK for a while, but over time the first IP begins handling all of the traffic. At the very end I reordered the upstream IPs, and the first one was once again selected for the majority of requests.
[...]

PR #3940, opened 2020-06-11
forward: Allow selection of health check transport

pnovotnak:
And select TCP when `force_tcp` is set.

It sets the transport used to query upstream servers in health checks to TCP when `force_tcp` is set true.

https://github.com/coredns/coredns/issues/3938

The documentation should call the switch out.

Yes, this will need to be called out in release notes. HOWEVER, it can be assumed that it won't break things since all queries to these upstreams will be over TCP anyway.

### DIFF
--- a/plugin/forward/fuzz.go
+++ b/plugin/forward/fuzz.go
@@ -16,8 +16,8 @@ var f *Forward
 func init() {
 	f = New()
 	s := dnstest.NewServer(r{}.reflectHandler)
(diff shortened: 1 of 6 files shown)


chrisohaver:
LGTM.  The circleci failure is unrelated to this change.

miekg:
this should probably be put in the `options` struct just as hcRecursionDesired

miekg:
Pr looks stale, will close tomorrow

PR #3944, opened 2020-06-12
forward: retry failed

huangnauh:

When some resolves might return ServerFailure or Refused, retry a different upstream. 
 
```
forward FROM TO... {
    ...
    retry_failed MAX
}
```

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -53,6 +53,7 @@ forward FROM TO... {
     policy random|round_robin|sequential
     health_check DURATION [no_rec]
     max_concurrent MAX
+    retry_failed MAX
 }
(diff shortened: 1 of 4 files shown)


chrisohaver:
Please provide some justification for adding the option.  e.g. Describe why/when someone would want to use this feature.

huangnauh:
> Please provide some justification for adding the option. e.g. Describe why/when someone would want to use this feature.

In my case,  upstream servers fail occasionally with a SERVFAIL.

miekg:
looks stale(-ish), but also very much a fringe case. Please re-open with some data on why this is a good feature to have and why it needs a control knob

PR #3955, opened 2020-06-20
Update dns64 plugin bugs list

SuperQ:

Remove the "Follow CNAMEs" bug, CNAME following is handled by the
forward plugin.

Docs only change.

Nope


SuperQ:
With a typical DNS64 lookup config like this:

```
. {
  dns64
  forward . 2001:4860:4860::8888
}
```

The handling of CNAMEs works:

```
$ host -t AAAA www.github.com ::1

www.github.com is an alias for github.com.
github.com has IPv6 address 64:ff9b::8c52:7604
```

zouyee (approved):
/lgtm

PR #3966, opened 2020-06-23
Improve gRPC Plugin when backend is not available

### LINKED ISSUE #3965: gRPC plugin support retries and timeout
**What would you like to be added**: It seems that gRPC plugin doesn't support configuring a timeout and the retries. Also it doesn't fit well with template and cancel plugin :/

In an example configuration:
 ```
solver.example.com {

    template IN SOA {
      authority "{{ .Zone }} 60 IN SOA {{ .Zone }} admin1.{{ .Zone }} (1 60 60 60 60)"
        fallthrough
    }

    template IN A {
      answer "{{ .Name }} 60 IN A 123.123.123.123"
        fallthrough
    }

    log
    errors
    reload
    cancel 2s
    grpc . 127.0.0.1:18853 {
  
    }
}
```

It doesn't respect the cancel context, and spends 5s in a query. As an example:

```
time dig txt  aaa.solver.example.com

real    0m15,035s
user    0m0,014s
sys     0m0,021s

* In CoreDNS Server:
[INFO] 123.123.123.111:11879 - 16332 "TXT IN aaa.solver.example.com. udp 61 false 4096" NOERROR - 0 0.000104219s
[INFO] 123.123.123.111:11879 - 16332 "TXT IN aaa.solver.example.com. udp 61 false 4096" NOERROR - 0 0.000196813s
[INFO] 123.123.123.111:11879 - 16332 "TXT IN aaa.solver.example.com. udp 61 false 4096" NOERROR - 0 0.000105267s
```

So with cancel = 2s it should take max of 6s (as the DNS retries 5 times) but instead it's failing in the client timeout and not answering with an error.

I'm proposing this as a feature so maybe gRPC plugin could support a Dial Timeout (as this is already supported in gRPC client), but it's strange to me that the context timeout is 2s but CoreDNS doesn't answer in this 2s :(

If inserting this as a directive to gRPC plugin is ok to you I can create the PR, otherwise need to discover here why the cancel plugin is not respected by gRPC :)

**Why is this needed**:

### PR DESCRIPTION
**Pull request by rikatz**
Signed-off-by: Ricardo Pchevuzinske Katz <ricardo.katz@serpro.gov.br>

1. Why is this pull request needed and what does it do? This PR always adds a return to gRPC plugin, even if the backend is unavailable. This way the client does not wait 5s (DNS Timeout) and got an error if no backend can provide answer

2. Which issues (if any) are related? Fixes #3965

3. Which documentation changes (if any) need to be made? None

4. Does this introduce a backward incompatible change or deprecation? No

### DIFF
--- a/plugin/grpc/grpc.go
+++ b/plugin/grpc/grpc.go
@@ -3,6 +3,7 @@ package grpc
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"time"
 
 	"github.com/coredns/coredns/plugin"
@@ -87,6 +88,9 @@ func (g *GRPC) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 		return 0, nil
 	}
 
+	if ret == nil {
+		return dns.RcodeServerFailure, errors.New("no healthy gRPC backend")
+	}
 	return 0, nil
 }
 
--- a/plugin/grpc/grpc.go
+++ b/plugin/grpc/grpc.go
@@ -37,10 +37,10 @@ func (g *GRPC) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 	}
 
 	var (
-		span, child ot.Span
-		ret         *dns.Msg
-		err         error
-		i           int
+		span, child      ot.Span
+		ret              *dns.Msg
+		upstreamErr, err error
+		i                int
 	)
 	span = ot.SpanFromContext(ctx)
 	list := g.list()
@@ -74,6 +74,8 @@ func (g *GRPC) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 			child.Finish()
 		}
 
+		upstreamErr = err
+
 		// Check if the reply is correct; if not return FormErr.
 		if !state.Match(ret) {
 			debug.Hexdumpf(ret, "Wrong reply for id: %d, %s %d", ret.Id, state.QName(), state.QType())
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: plugin/grpc/grpc.go [hunk 1]
```diff
@@ -87,6 +87,11 @@ func (g *GRPC) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 		return 0, nil
 	}
 
+	if ret == nil {
```
miekg: this is a good change, and should follow what forward does in forward/forward.go:

~~~
       if upstreamErr != nil {
|       |       return dns.RcodeServerFailure, upstreamErr
|       }           
                    
|       return dns.RcodeServerFailure, ErrNoHealthy
}                   
                 
~~~

rikatz: Replaced everything here to the same error of forward, but instead of creating a new const returned directly a new error.

rikatz:
@miekg changed as asked :)

I'll maybe later this week in other PR improve the ServeDNS and the unit test of this plugin, turning this more like the forward one (taking the good ideas and copying here)

Tks

rikatz:
@miekg anything here in this PR? Thanks :D

plugin/grpc/grpc.go
```diff
@@ -87,6 +88,9 @@ func (g *GRPC) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
...
+		return dns.RcodeServerFailure, errors.New("no healthy gRPC backend")
+	}
 	return 0, nil
```
miekg: this needs to be changed as well, and brought inline with the forward plugin:
`return dns.RcodeServerFailure, ErrNoHealthy`

plugin/grpc/grpc.go
miekg: this needs to track the error, like upstreamErr in forward plugin

rikatz: Got it, changed

plugin/grpc/grpc.go
rikatz: Also changed, replaced the nil return with the error assuming that if the plugin reached this step something ocurred

rikatz:
@miekg new changes made as asked :D Thank you very much

miekg (approved):

PR #3970, opened 2020-06-24
plugin/forward: remove exp backoff stuff

miekg:
we hc every 0.5s, doing exp backoff will create a large gap in the
ability to re-use an upstream. Doing a exp. backoff up to (say) 3s,
isn't really exp backoff either.

Remove the wording from the documentation.


zouyee (approved):

yongtang (approved):
lgtm

wisre:
thx

Issue #3929, opened 2020-06-07
Make notify client IP configurable


JensErat:
**What would you like to be added**:

Add a configuration option to select client IP notify is sent from.

**Why is this needed**:

Usually other name servers only accept notifies from a whitelisted IP address. If you have a machine with multiple interfaces attached, CoreDNS just chooses "some" address without the operator having proper influence on selection.

- - -

I guess all that's needed is adding a configured `Dialer` (https://github.com/miekg/dns/blob/master/client.go#L34) in the respective `notify(...)` functions (eg. https://github.com/coredns/coredns/blob/87bd9dec856b743d65d71f90accc4296d767b109/plugin/file/notify.go#L48). Maybe this is something rather to be implemented in #3121? Otherwise, how would a proper configuration look like -- a `notify from <client-ip>` option in `file` plugins?

chrisohaver:
I assume that the system is (consistently) deciding which interface to use based on the system routing table and destination IP of the notify packet.

JensErat:
I should have rather explained another situation with a VIP for HA purpose added to the machine. You don't really have control over the IP to choose by default, especially if the IP is reconfigured dynamically. You could apply dirty policy based routing hacks, but this calls for issues.

The canonic way of selecting an outgoing IP address is on the socket side. Similar issues will occur with IPv6 notifies, multiple IPv6 addresses on the same interface are much too common.

JensErat:
[Bind has an option](https://serverfault.com/a/692917/98727), [PowerDNS got one](https://doc.powerdns.com/authoritative/settings.html#query-local-address), I guess all the others will also have this address configurable.

chrisohaver:
OK - I think a reasonable behavior would be something like...  If the notify destination IP falls within the subnet of a configured "notify from" interface, sending using that use that interface.  If not, let the system route it.

JensErat:
I don't think this would solve any kind of real world issue. Setting up a proper source IP is important especially when you do _not_ have control over the receiving side, you can usually setup exactly one IP allowed to notify. If the receiving system is in your own subnet, you'll probably be able to configure pretty arbitrary trust anyway.

Example setup: Consider

- a HA pair of DNS servers 192.0.2.10/24 and 192.0.2.11/24
- one of them holding a failover IP from the same network 192.0.2.53/24 (also listed as primary NS)
- you want to notify a nameserver service operator on 198.51.100.100 to update the zones
- the service operator allows exactly one IP address per zone as notify source

[...]

PR #3972, opened 2020-06-26
Implement notifies for transfer plugin

miekg:
* Fix all plugins to use the new implementation (this is backwards
  incompatible)
* Implement notifies in transfer
* Make file and auto do notifies using transfer

This copied stuff from #3452. It leave out (PR is long enough as it is) the notify source port PR: #3420. That can and should probably bo cherry-picked in .

transfer/README.md is updated and the plugins implementing `Transferer` have seen updates as well.

Yes, this rips out the old code transferring and also makes the Corefile parsing a fatal error.


miekg:
also need to check what happens when a plugin gets hit TypeAXFR and how they respond. AXFR fallback for IXFR may need work, the *file* plugin did something special there.

chrisohaver:
I suppose this supersedes the various open PRs that implement parts of this?
e.g. #3732 #3452 #3420 #3419

I don't think #3929 is addressed here (implemented in #3420).

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Implement not..." ]
>I suppose this supersedes the various open PRs that implement parts of this?
>e.g. #3732 #3452 #3420 #3419

oh sorry, I missed those. Let me review with my new knowledge (happy to drop this PR)

/Miek

--
Miek Gieben

chrisohaver:
> happy to drop this PR

No, that's not necessary.  I was just verifying whether or not these were intended to replace, or to expand upon any of the existing related PRs.  I'll close my PRs, and we can move forward with this one.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Implement not..." ]
>    happy to drop this PR
>
>No, that's not necessary. I was just verifying whether or not these were
>intended to replace, or to expand upon any of the existing related PRs. I'll
>close my PRs, and we can move forward with this one.

left a bunch of comments on those. Some stuff is better there than here. I can probably
trim this back down to *just* the notifies and merge your other PRs?

chrisohaver:
> left a bunch of comments on those. Some stuff is better there than here. I can probably
trim this back down to *just* the notifies and merge your other PRs?

I think the easiest way forward (least amount of total work), would be to pull aspects of those PRs that you like into this one, rather than the other way around.   The implementation in this PR is naturally fresh in your mind.  The prior PRs have been sitting around for 4 - 8 months... I have lost some familiarity ...  some have collected merge conflicts. etc.

miekg:
I have rebased and merged this branch, so I can't easliy squashed.

This implements notifies and fixed a bunch of bugs in transfer along the way. All supporting plugin explicitly implement ixfr to axfr fallback.

miekg:
this acl checking is problematic with this plugin (i've changed it here, but it's still wrong)

~~~
org {
    auto {
        directory /etc/coredns/zones/org
    }
 
    transfer {
        to *
        to 10.240.1.1
    }
}
~~~
so *transfer* has `org` in it's auth list. An axfr for example.org won't match (in the current code, it did before). Moving that back to `Matches` would fix it, but this is also valid:
~~~
org {
    auto {
        directory /etc/coredns/zones/org
    }
 
    transfer example.org {
        to *
        to 10.240.1.1
    }

    transfer sub.example.org {
        to 10.240.1.2
    }
}
~~~
The key problem is that matching the ACL is different that matching the authoritativeness of the zone which is done with ErrNotAuthoritative.

miekg:
*maybe* `Matches` was the right thing to do, left me see if I can test this properly

removing upstream stuff *in this PR* was stupid, I'll revert that.

miekg:
ok PTAL, a build from this branch is now serving coredns.io, accoding to the logging it's all working nicely

chrisohaver (approved):

chrisohaver:
>  It leave out (PR is long enough as it is) the notify source port PR: #3420.

FYI, #3420 was not simply a notify source port PR it was an entire implementation of notify (hence its length), created in November 2019.  The notify source port issue had only recently been raised, and I added that case to the PR only recently.

yongtang (approved):

chrisohaver:
This was a backward incompatible change. Probably should have held on to it before merging right away in case any other critical bugs come up before next release.   With this merged, I think Next release Needs to be 1.8.0 ?

On Jul 7, 2020, at 3:38 PM, Yong Tang <notifications@github.com> wrote:

﻿

Merged #3972<https://urldefense.com/v3/__https://github.com/coredns/coredns/pull/3972__;!!JYsgTRAg6ZQ!ZP4DugOLWxwDV0vhWUGgPZJngg05c1M2tbIiN4Kfd8dDolWJmHtMmTl4t_JwPJRe$> into master.

yongtang:
Ah yes that might be a concern. We normally release minor versions every several months, and release patch versions every couple of weeks. The 1.7.0 was released only 3 weeks ago.

Maybe we can go directly to 1.8.0 in the next round?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Implement not..." ]
>Ah yes that might be a concern. We normally release minor versions every
>several months, and release patch versions every couple of weeks. The 1.7.0 was
>released only 3 weeks ago.
>
>Maybe we can go directly to 1.8.0 in the next round?

we can revert this for the time being? No matter what else we end up merging, if this goes
in we need to make it a 1.8.0 release. It's not as bad as it's sounds unless you use zone
transfers. I.e. for k8s this would most likely be a drop-in replacement.

johnbelamaric:
In the meantime, can we build translation support into the migrate tool, if we haven't already?

chrisohaver:
> In the meantime, can we build translation support into the migrate tool, if we haven't already?

Should be easy to add once it's released in CoreDNS.

PR #3987, opened 2020-07-02
complete auto plugin  test case

zouyee:
Signed-off-by: zounengren <zounengren@cmss.chinamobile.com>

complete auto plugin  test case

none

none

none

### DIFF
--- a/plugin/auto/auto_test.go
+++ b/plugin/auto/auto_test.go
@@ -1,0 +1,172 @@
+package auto
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/coredns/coredns/plugin/file"
+	"github.com/coredns/coredns/plugin/pkg/dnstest"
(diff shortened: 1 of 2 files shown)


plugin/auto/auto_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	"github.com/coredns/coredns/plugin/file"
+	"github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/coredns/coredns/plugin/test"
+
+	"github.com/miekg/dns"
+)
+
+// TODO: consider variables reuse，move to pkg/test?
+var miekAuth = []dns.RR{
```
miekg, line 16: we should not further using `miek.nl` for testing purposes. example.org,com,net is better.

also this looks wholesale copied. What tests are you adding, your commit msg is silent on this?

PR #4016, opened 2020-07-20
Disable debug mode for tracing - removes extra logging

uthark:
With debug mode tracer generates too many logs and they are written even without `log` plugin enabled.

This is an easiest option.
Other are: respect `log` plugin (might be difficult) or add configuration to a plugin.

### DIFF
--- a/plugin/trace/trace.go
+++ b/plugin/trace/trace.go
@@ -10,6 +10,7 @@ import (
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/metrics"
 	"github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/coredns/coredns/plugin/pkg/log"
 	"github.com/coredns/coredns/plugin/pkg/rcode"
 	_ "github.com/coredns/coredns/plugin/pkg/trace" // Plugin the trace package.
 	"github.com/coredns/coredns/request"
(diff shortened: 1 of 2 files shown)


plugin/trace/trace.go
```diff
@@ -52,7 +52,7 @@ func (t *trace) OnStartup() error {
 		case "zipkin":
 			err = t.setupZipkin()
 		case "datadog":
-			tracer := opentracer.New(tracer.WithAgentAddr(t.Endpoint), tracer.WithServiceName(t.serviceName), tracer.WithDebugMode(true))
+			tracer := opentracer.New(tracer.WithAgentAddr(t.Endpoint), tracer.WithServiceName(t.serviceName), tracer.WithDebugMode(false))
```
miekg: lgtm, but we can remove the entire `WithDebugmode` option?

chrisohaver: How about enabling it when debug is on: e.g.  `tracer.WithDebugMode(log.D.Value())`

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Disable debug..." ]
>
>
>In plugin/trace/trace.go:
>
>> @@ -52,7 +52,7 @@ func (t *trace) OnStartup() error {
>                case "zipkin":
>                        err = t.setupZipkin()
>                case "datadog":
>-                       tracer := opentracer.New(tracer.WithAgentAddr(t.Endpoint), tracer.WithServiceName(t.serviceName), tracer.WithDebugMode(true))
>+                       tracer := opentracer.New(tracer.WithAgentAddr(t.Endpoint), tracer.WithServiceName(t.serviceName), tracer.WithDebugMode(false))
>
>How about enabling it when debug is on: e.g. tracer.WithDebugMode(log.D.Value
>())

nice one!

plugin/trace/trace.go
uthark: That's even better. Updated PR.

miekg:
documentation now needs to reference the *debug* plugin. This should be in the introduction and also add a See Also section that points to the debug plugin - see other plugins that have this section for inspiration.

uthark:
@miekg please take another look.

miekg (approved):

PR #4017, opened 2020-07-21
Fix metric naming test

yeya24:
Signed-off-by: Ben Ye <yb532204897@gmail.com>

1. Change the `Visit` method https://github.com/coredns/coredns/blob/master/test/metric_naming_test.go#L89 to use a  pointer receiver, otherwise, it always returns nil.

2. Remove the `""` quotes when getting string literal value from `ast.BasicLit.Value`

### DIFF
--- a/test/metric_naming_test.go
+++ b/test/metric_naming_test.go
@@ -1,20 +1,20 @@
 package test
 
 import (
-	"bytes"
 	"go/ast"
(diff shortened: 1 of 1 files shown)


miekg (approved):

SuperQ (approved):
LGTM

PR #4018, opened 2020-07-21
For caddy v1 in our org

miekg:
This RP changes all imports for caddyserver/caddy to coredns/caddy. This
is the v1 code of caddy.

For the coredns/caddy repo the following changes have been made:

* anything not needed by us is deleted
* all `telemetry` stuff is deleted
* all its import paths are also changed to point to coredns/caddy
* the v1 branch has been moved to the master branch
* a v1.1.0 tag has been added to signal the latest release


miekg:
See https://github.com/coredns/ci/pull/131 for the circle-ci fix

miekg:
As to why?

Caddy is moving to a v2 architecture, which is backwards incompatible so I don't see why we should follow that. However that implies v1 will be unmaintained so it makes sense to pull this bits towards us and take owernship

ihac (approved):

miekg:
this is backwards incompatible for plugin devs... it's a rename of the import and that's it, but still annoying

darshanime:
Any plan of moving to caddy v2? It might be worthwhile not having to maintain caddy v1 ourselves.

miekg:
~~~
find -type f -name "*.go" -exec sed -i 's|github.com/caddyserver/caddy|github.com/coredns/caddy|' {} \;
~~~

miekg:
@darshanime I don't want to move to the new plugin system, for the core we may want to move? Haven't looked a the code changes. Regardless that would also be a backward incompatible change I think

What benefits do you see of moving to v2 @darshanime  ?

stp-ip:
Not sure I agree with this due to the following points:
* we need to maintain caddy v1
* breaking change for plugin devs
* if we decide to move to caddy v2 we would be breaking plugin devs again

If we break them, we should probably only do it once and with the benefit of having a maintained base such as caddy v2.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] For caddy v1 ..." ]
>Not sure I agree with this due to the following points:
>
>  • we need to maintain caddy v1
>  • breaking change for plugin devs
>  • if we decide to move to caddy v2 we would be breaking plugin devs again
>
>If we break them, we should probably only do it once and with the benefit of
>having a maintained base such as caddy v2.

I haven't seen any evidence v2 will improve anything for us; I've only seen v1 being
deprecated meaning we need to do something just to maintain the status quo

stp-ip:
> I haven't seen any evidence v2 will improve anything for us; I've only seen v1 being deprecated meaning we need to do something just to maintain the status quo

Moving to v2 currently doesn't improve much for us, but keeping and maintaining v1 doesn't improve much for us either and the downside of breaking our plugin devs seems to be similar.

One potential user upside with v2 is the additional configuration possibilities.

Mostly against us maintaining v1, but going with what the consensus is or the one doing the work ;)

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] For caddy v1 ..." ]
>    I haven't seen any evidence v2 will improve anything for us; I've only seen
>    v1 being deprecated meaning we need to do something just to maintain the
>    status quo
>
>Moving to v2 currently doesn't improve much for us, but keeping and maintaining
>v1 doesn't improve much for us either and the downside of breaking our plugin
>devs seems to be similar.

you seem to be overstating this; it's changing an import path, which you can automate.

>One potential user upside with v2 is the additional configuration
>possibilities.

such as? As this requires plugin author to completely rewrite their setup functions, this
should be on the level 2x/3x better than what we currently have (i.e. super-must-have
features).

>Mostly against us maintaining v1, but going with what the consensus is or the
>one doing the work ;)

We could potentially just use the v2 bits we need and not pull in the plugin/module
changes, which we most definitely don't need. And if we're going this route we _still_
want to fork it into coredns to prevent this happening again in the future.

Note I haven't looked into this, because it's toil and we're perfectly happy with what
caddy's v1 libs provide for us.

/Miek

--
Miek Gieben

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] For caddy v1 ..." ]
>Mostly against us maintaining v1, but going with what the consensus is or the
>one doing the work ;)

I think we can even ask Matt to add a type alias in the caddy codebase. I *think* this
should make the transition painless. (Not sure: never need a type alias before)

stp-ip:
> such as? As this requires plugin author to completely rewrite their setup functions, this should be on the level 2x/3x better than what we currently have (i.e. super-must-have features).

Mainly the ability to programmatically configure via an HTTP endpoint and JSON, while still being able to use the caddyfile style.
In smaller setups it probably doesn't make a huge difference. For more complex setups it could be interesting, but I would agree not easily 3x better.

dilyevsky (approved):

miekg:
circle ci will be broken by this because:
~~~
go: found github.com/caddyserver/caddy in github.com/caddyserver/caddy v1.0.5
# github.com/coredns/ci/test/kubernetes [github.com/coredns/ci/test/kubernetes.test]
./tools.go:147:2: cannot use server (type *"github.com/coredns/caddy".Instance) as type *"github.com/caddyserver/caddy".Instance in return argument
~~~
this can only be fixed *after* this is merged.

PR #4023, opened 2020-07-24
Add support for five TLSv1.3 ciphersuites

OleksandrBlack:
Add support for five TLSv1.3 ciphersuites


miekg:
Why is this needed?

OleksandrBlack:
@miekg https://github.com/ZcashFoundation/coredns-zcash  (dnsseeder) uses coredns 
safecoin.org forked coredns-zcash to coredns-safecoin but 
safecoin daemon uses TLS 1.3 for p2p connection (cipher: TLS_AES_256_GCM_SHA384)
for this reason gets the error TLS handshake fails

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add support f..." ]
>@miekg https://github.com/ZcashFoundation/coredns-zcash (dnsseeder) uses
>coredns
>safecoin.org forked coredns-zcash to coredns-safecoin but
>safecoin daemon uses TLS 1.3 for p2p connection (cipher:
>TLS_AES_256_GCM_SHA384)
>for this reason gets the error TLS handshake fails

Is there some kind of function from the go std lib we can call that does the same?

PR #4034, opened 2020-07-30
plugin/template: Add client IP data

Miouge1:
This PR adds a `{{ .Remote }}` data in the template module.

I have tested it with the following minimal `Corefile` (kind of emulating the whoami plugin):
```
. {
    template IN A {
      answer "{{ .Name }} 60 {{ .Class }} {{ .Type }} {{ .Remote }}"
    }
}
```

```console
$ dig @127.0.0.1 A example.com +short
127.0.0.1
```

In terms of use case, the client IP can be useful to make decisions.

None
`plugin/template/README.md` included in the PR
Adding a field should be backwards compatible.

### DIFF
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -48,6 +48,7 @@ Each resource record is a full-featured [Go template](https://golang.org/pkg/tex
 * `.Group` a map of the named capture groups.
 * `.Message` the complete incoming DNS message.
 * `.Question` the matched question section.
+* `.Remote` client’s IP address
 * `.Meta` a function that takes a metadata name and returns the value, if the
   metadata plugin is enabled. For example, `.Meta "kubernetes/client-namespace"`
 
--- a/plugin/template/template.go
+++ b/plugin/template/template.go
@@ -48,6 +48,7 @@ type templateData struct {
 	Type     string
 	Message  *dns.Msg
 	Question *dns.Question
+	Remote   string
 	md       map[string]metadata.Func
 }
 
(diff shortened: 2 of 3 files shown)


rtreffer (approved):
Looks cool, although it would be nice to say it might be the IP of an intermediate (at which point it becomes pretty pointless for the original client).

miekg:
In the *log* plugin we call this (for better or worse) **remote**. i think we should have this consistent between plugin, by either renaming it here; or adding `{ip}` to log. Think the more generic name 'remote' is slightly better

Miouge1:
Good points.  I've replaced `.IP` to `.Remote` in the latest commit and re-tested this locally:
```
. {
    log
    template IN ANY {
      answer "{{ .Name }} 60 {{ .Class }} {{ .Type }} {{ .Remote }}"
    }
}
```


```console
$ dig @127.0.0.1 TXT example.com +short
"127.0.0.1"
$ dig @::1 TXT example.com +short
"::1"
```


The log module documents `{{remote}}` as `client's IP address, for IPv6 addresses these are enclosed in brackets: [::1]`, since the brackets enclosing doesn't happen here I've skipped that from the description in this PR.

miekg:
>Good points. I've replaced .IP to .Remote in the latest commit and re-tested
>this locally:
>
>. {
>    log
>    template IN ANY {
>      answer "{{ .Name }} 60 {{ .Class }} {{ .Type }} {{ .Remote }}"
>    }
>}
>
>$ dig @127.0.0.1 TXT example.com +short
>"127.0.0.1"
>$ dig @::1 TXT example.com +short
>"::1"

cool
>
>The log module documents {{remote}} as client's IP address, for IPv6 addresses
>these are enclosed in brackets: [::1], since the brackets enclosing doesn't
>happen here I've skipped that from the description in this PR.

interesting, should we make it do that? I think we should, because
`{{ .Remote }} {{ .Port }}` doesn't allow you to see there is a port number

(if .Port even exists - dunno)

/Miek

--
Miek Gieben

Miouge1:
`{{ .Port }}` does not exist, but would not be too hard to add.

Are brackets enclosed IPv6 IPs supported in AAAA records? (I couldn't find an example of that in the codebase).

miekg:
>{{ .Port }} does not exist, but would not be too hard to add.
>
>Are brackets enclosed IPv6 IPs supported in AAAA records? (I couldn't find an
>example of that in the codebase).

oh sorry, this is aaaa, then no. Thought we where somehow returning a text representation
of the record.

Issue #4036, opened 2020-07-30
log.Debugf is silent even when `debug` plugin is loaded


ashtw:
**What happened**:

The `debug` plugin is loaded but calls to log.Debugf() in my plugin are silent

**What you expected to happen**:

I expected to see log.Debugf() messages in program output

**How to reproduce it (as minimally and precisely as possible)**:

Enable `debug` plugin and look for output from log.Debugf()

**Anything else we need to know?**:

**Environment**:

- the version of CoreDNS: master branch

ashtw:
See pull request [4037](https://github.com/coredns/coredns/pull/4037)

miekg:
see https://github.com/coredns/coredns/pull/4007

PR #4037, opened 2020-07-30
enable log.Debugf messages when debug plugin is loaded

ashtw:

Causes behavior of `debug` plugin to match [documentation](https://github.com/coredns/coredns/tree/master/plugin/debug/README.md): "A side effect of using *debug* is that `log.Debug` and `log.Debugf` messages will be printed to standard output."

[4036](https://github.com/coredns/coredns/issues/4036)

None

No

### DIFF
--- a/plugin/debug/debug.go
+++ b/plugin/debug/debug.go
@@ -5,6 +5,8 @@ import (
 	"github.com/coredns/coredns/plugin"
 
 	"github.com/caddyserver/caddy"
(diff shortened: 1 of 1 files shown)


miekg:
this is not what an init should do and we doing this already from the main server startup and we unset it as well.

ashtw:
@miekg Are you saying #4036 is invalid?

PR #4040, opened 2020-07-31
plugin/cache: Metadata cache key

chrisohaver:

I'm proposing this addition to the cache plugin to help support caching of policy-based responses (IOW, when the response for a given record can differ based on attributes other than name/class/type).  Normally, such responses cannot be safely cached, because the answer differs based on attributes that are not part of the hash function.

The idea here is to allow the cache plugin to compute the cache hash using the normal response attributes in conjunction with metadata. For example, another plugin could expose metadata that represents a policy-group that the remote client belongs to. When the cache is configured to use that metadata, the result is that a client only hits cache if the name/type/class _and_ the policy group match it.


miekg:
how does work we would rework how the cache caches? If we go for a per-RRset cache, will this not invalidate this approach?

plugin/cache/README.md
```diff
@@ -55,6 +56,9 @@ cache [TTL] [ZONES...] {
   available.  When this happens, cache will attempt to refresh the cache entry after sending the expired cache
   entry to the client. The responses have a TTL of 0. **DURATION** is how far back to consider
   stale responses as fresh. The default duration is 1h.
+* `addkey` adds the listed **METADATA** to the cache key when writing and reading from cache.
+  This can be used to cache policy based responses which contain answers that are are dependent
```
rdrozhdzh, line 60: are are dependent -> are dependent

chrisohaver:
> how does [this] work [if] we would rework how the cache caches? If we go for a per-RRset cache, will this not invalidate this approach?

I don't _think_ so... but I don't have a clear picture of how per-RRset cache is intended to work.  _IIUC_, In a per-RRset solution we keep a RR-set only cache (that is a cache of RR-sets, without any question data).  But something still needs to be able to join an incoming request to the valid RR-sets that will comprise the response.  And for that I think you'd need to keep a query cache of query-hash-id -> question-data + rrset-hash-id. So given an incoming query, we'd look up the query in the query-cache, and then construct the response based on the question-data and rrsets with the rrset-hash-id.

rdrozhdzh (commented):
LGTM

miekg:
a per-RR cache would change everything, so there is nothing that can be made to keep this backwards compatible.

If you need to calculate the has you'll need to list the code from the cache: https://github.com/coredns/coredns/blob/master/plugin/pkg/cache/cache.go#L12

chrisohaver:
> a per-RR cache would change everything, so there is nothing that can be made to keep this backwards compatible.
> 
> If you need to calculate the has you'll need to list the code from the cache: https://github.com/coredns/coredns/blob/master/plugin/pkg/cache/cache.go#L12

I'm not quite following... calculating the hash is based on inputs derived from the request. What do you mean by "list the code from the cache?"

PR #4044, opened 2020-08-03
plugin/template: EDNS local option

chrisohaver:

Adds the option to respond with an OPT pseudo-record with EDNS local options.
I used this to test an external plugin that depends on an upstream response including edns local options.

I could not puzzle out a way of adding edns options using the existing additional section RR string parsing. Looks like it has to do with the way RR string parser treats OPT records as RFC-3597 (unknown type).

### DIFF
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -12,7 +12,7 @@ The *template* plugin allows you to dynamically respond to queries by just writi
 
 ~~~
 template CLASS TYPE [ZONE...] {
-    match [REGEX...]
+    match REGEX...
     answer RR
     additional RR
     authority RR
@@ -25,7 +25,7 @@ template CLASS TYPE [ZONE...] {
 * **CLASS** the query class (usually IN or ANY).
 * **TYPE** the query type (A, PTR, ... can be ANY to match all types).
 * **ZONE** the zone scope(s) for this template. Defaults to the server zones.
-* `match` **REGEX** [Go regexp](https://golang.org/pkg/regexp/) that are matched against the incoming question name. Specifying no **REGEX** matches everything (default: `.*`). First matching regex wins.
+* `match` **REGEX** [Go regexp](https://golang.org/pkg/regexp/) that are matched against the incoming question name. Omitting `match` will match everything (default: `.*`). First matching regex wins.
 * `answer|additional|authority` **RR** A [RFC 1035](https://tools.ietf.org/html/rfc1035#section-5) style resource record fragment
   built by a [Go template](https://golang.org/pkg/text/template/) that contains the reply.
 * `edns local` add an EDNS0 local OPT pseudo-RR to the response. Each **CODE**=**DATA** pair adds an EDNS0 local value.
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -12,10 +12,11 @@ The *template* plugin allows you to dynamically respond to queries by just writi
 
 ~~~
 template CLASS TYPE [ZONE...] {
-    match REGEX...
+    match [REGEX...]
     answer RR
     additional RR
     authority RR
+    ednslocal CODE=DATA...
     rcode CODE
     fallthrough [ZONE...]
 }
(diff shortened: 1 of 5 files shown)


plugin/template/setup.go
```diff
@@ -109,6 +109,22 @@ func templateParse(c *caddy.Controller) (handler Handler, err error) {
 					t.additional = append(t.additional, tmpl)
 				}
 
+			case "edns":
```
miekg, line 112: should this not be documented somewhere?

plugin/template/README.md
```diff
@@ -24,9 +25,10 @@ template CLASS TYPE [ZONE...] {
 * **CLASS** the query class (usually IN or ANY).
 * **TYPE** the query type (A, PTR, ... can be ANY to match all types).
 * **ZONE** the zone scope(s) for this template. Defaults to the server zones.
-* **REGEX** [Go regexp](https://golang.org/pkg/regexp/) that are matched against the incoming question name. Specifying no regex matches everything (default: `.*`). First matching regex wins.
+* `match` **REGEX** [Go regexp](https://golang.org/pkg/regexp/) that are matched against the incoming question name. Omitting `match` will match everything (default: `.*`). First matching regex wins.
 * `answer|additional|authority` **RR** A [RFC 1035](https://tools.ietf.org/html/rfc1035#section-5) style resource record fragment
   built by a [Go template](https://golang.org/pkg/text/template/) that contains the reply.
+* `edns local` add an EDNS0 local OPT pseudo-RR to the response. Each **CODE**=**DATA** pair adds an EDNS0 local value.
```
miekg, line 31: `local` is option, so **OPTION** or something, where `local` is the only one allowed?

chrisohaver: Yes, for now only local is allowed.  My intent was to make the syntax flexible enough to gracefully support adding other EDNS0 option types down the road.  Since the **DATA** format would differ based on the type, IMO, even if we add more types, it would be easier to express the syntax for each supported type individually.

miekg: ack, but where do I put those `CODE=DATA` values. I think this can be expanded upon  and made clearer on what keywords are and what the option are

chrisohaver: OK.  ... and an example would help ...

chrisohaver: After iterating on this a bit more, I think it would be clearer to make the syntax be  ... 

`edns-option CODE VALUE`

... multiple occurrences allowed, each occurrence adding an option code/value pair to the OPT record.

In the future, if there is a need for it, we could add an `edns-header` option that allows customization of other edns bits (i.e. extended rcode and other flags) ... e.g.  something like `edns-header X-RCODE DO` ... of which only one occurrence is allowed.

WDYT?

chrisohaver:
Any interest in this?  Or should I close it?

plugin/template/template.go
```diff
@@ -128,6 +142,29 @@ func (h Handler) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
+	pos := strings.Index(buffer.String(), "=")
+	i, err := strconv.Atoi(buffer.String()[0:pos])
+	if err != nil {
+		templateRRFailureCount.WithLabelValues(server, data.Zone, data.Class, data.Type, section, template.Tree.Root.String()).Inc()
+		return err
+	}
+	code := uint16(i)
+	var edata []byte
+	if buffer.Len() > pos {
+		edata = buffer.Bytes()[pos+1:]
```
johnbelamaric, line 161: In `metadata_edns0` we allow decoding hex, addresses, and bytes. Wouldn't it make sense to allow the encodings of that here, especially given that data may come from template data (which IIRC could be metadata added by another plugin).

https://github.com/coredns/metadata_edns0

What I am getting at is, shouldn't we be able to encode any binary data in the value?

johnbelamaric:
Looks potentially useful to me, though I don't have any specific use cases I need.

plugin/template/template.go
chrisohaver: I was assuming we could do this with escapes .... like "\x1f" ...
I'll check to see if that actualy works.

chrisohaver: darn that doesn't work.  Plain strings is all I need for my tests, so i'll just keep this around for my own personal use.

chrisohaver:
Closing this.  I've been thinking this feature might be better as a separate plugin, where it adds edns options to the response.  It could be used in conjunction with template, or other plugins.  Also, that way it can exist as an external plugin.

miekg:
>Closing this. I've been thinking this feature might be better as a separate
>plugin, where it adds edns options to the response. It could be used in
>conjunction with template, or other plugins. Also, that way it can exist as an
>external plugin.

+1

PR #4049, opened 2020-08-10
plugin/etcd: update documention

### LINKED ISSUE #4048: etcd plugin delivers subdomain A record when queried for main domain, TLD or root
**What happened**:

I use CoreDNS with the etcd plugin. The etcd store is populated by ExternalDNS with currently only one record for `foo.example.com`. When I query CoreDNS for `foo.example.com.` the reply correctly contains an `A` record with the IP address. However, that same `A` record is also returned when I query CoreDNS for `example.com.`, `com.` or `.`.

When I query for `example.com.`, `com.` or `.`

* I would expect a reply without `ANSWER SECTION` and with `AUTHORITY SECTION` if the etcd plugin is configured **without** `fallthrough`;
* and I would expect a fallthrough if the etcd plugin is configured **with** `fallthrough`.

```
.:53 {
    etcd . {
        fallthrough
        stubzones
        endpoint https://cilium-etcd-client.coredns.svc:2379
        tls /etc/coredns/tls/etcd/etcd-client.crt /etc/coredns/tls/etcd/etcd-client.key /etc/coredns/tls/etcd/etcd-client-ca.crt
    }
    forward . /etc/resolv.conf
}
```

In my case I use `fallthrough` (see above) and the fallthrough does work when I query `bar.example.com`.

**Environment**:

- the version of CoreDNS: 1.7.0
- OS: Debian GNU/Linux 10 (buster)

miekg:
Make clear(er) how this plugin find records.

Fixes: #4048

### DIFF
--- a/plugin/etcd/README.md
+++ b/plugin/etcd/README.md
@@ -62,8 +62,8 @@ The *etcd* plugin leverages directory structure to look for related entries. For
 an entry `/skydns/test/skydns/mx` would have entries like `/skydns/test/skydns/mx/a`,
 `/skydns/test/skydns/mx/b` and so on. Similarly a directory `/skydns/test/skydns/mx1` will have all
 `mx1` entries. Note this plugin will search through the entire (sub)tree for records. In case of the
-first example a query for `mx.skydns.text` will return both the contents of the `a` and `b` records,
-if the directory extends deeper those records are returned as well.
+first example, a query for `mx.skydns.text` will return both the contents of the `a` and `b` records.
+If the directory extends deeper those records are returned as well.
 
 With etcd3, support for [hierarchical keys are
 dropped](https://coreos.com/etcd/docs/latest/learning/api.html). This means there are no directories
--- a/plugin/etcd/README.md
+++ b/plugin/etcd/README.md
@@ -8,7 +8,7 @@
 
 The *etcd* plugin implements the (older) SkyDNS service discovery service. It is *not* suitable as
 a generic DNS zone data plugin. Only a subset of DNS record types are implemented, and subdomains
-and delegations are not handled at all. The plugin will also recursively decent the tree and return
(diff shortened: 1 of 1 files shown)


plugin/etcd/README.md
```diff
@@ -8,14 +8,15 @@
 
 The *etcd* plugin implements the (older) SkyDNS service discovery service. It is *not* suitable as
 a generic DNS zone data plugin. Only a subset of DNS record types are implemented, and subdomains
-and delegations are not handled at all.
+and delegations are not handled at all. The plugin will also recursively decent the tree and return
+all records found, see "Special Behavior" below for details.
```
chrisohaver:
```suggestion
and delegations are not handled at all. The plugin will also recursively descend the tree and return
all records found, see "Special Behavior" below for details.
```

plugin/etcd/README.md
```diff
@@ -61,16 +61,18 @@ etcd [ZONES...] {
 The *etcd* plugin leverages directory structure to look for related entries. For example
 an entry `/skydns/test/skydns/mx` would have entries like `/skydns/test/skydns/mx/a`,
 `/skydns/test/skydns/mx/b` and so on. Similarly a directory `/skydns/test/skydns/mx1` will have all
-`mx1` entries.
+`mx1` entries. Note this plugin will search through the entire (sub)tree for records. In case of the
+first example a query for `mx.skydns.text` will return both the contents of the `a` and `b` records,
+if the directory extends deeper those records are returned as well.
```
chrisohaver:
```suggestion
first example, a query for `mx.skydns.text` will return both the contents of the `a` and `b` records.
If the directory extends deeper those records are returned as well.
```

chrisohaver (changes requested):
LGTM + minor spelling/grammar nits.

PR #4055, opened 2020-08-10
pkg/upstream: set edns0 and Do when required

miekg:
Set these options of the requests warrents them.

### DIFF
--- a/plugin/pkg/upstream/upstream.go
+++ b/plugin/pkg/upstream/upstream.go
@@ -25,8 +25,11 @@ func (u *Upstream) Lookup(ctx context.Context, state request.Request, name strin
 		return nil, fmt.Errorf("no full server is running")
 	}
 
+	size := state.Size()
+	do := state.Do()
 	req := new(dns.Msg)
 	req.SetQuestion(name, typ)
+	req.SetEdns0(uint16(size), do)
 
 	nw := nonwriter.New(state.W)
 


plugin/pkg/upstream/upstream.go
```diff
@@ -25,8 +25,11 @@ func (u *Upstream) Lookup(ctx context.Context, state request.Request, name strin
 		return nil, fmt.Errorf("no full server is running")
 	}
 
+	size := state.Size()
+	do := state.Do()
 	req := new(dns.Msg)
 	req.SetQuestion(name, typ)
+	req.SetEdns0(uint16(size), do)
```
chrisohaver, line 32: Perhaps we should be retaining the entire OPT record here. IIUC, ideally upstream should act as if the client made the request itself, in which case the entire OPT record, if present,  would be in the request.

miekg: this is a new request for CNAMEs usually (when coming out the file plugin). So we don't have an original to speak off

chrisohaver: It's not in `state.Req`?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] pkg/upstream:..." ]
>
>
>In plugin/pkg/upstream/upstream.go:
>
>>       req := new(dns.Msg)
>        req.SetQuestion(name, typ)
>+       req.SetEdns0(uint16(size), do)
>
>It's not in state.Req?

yes, but that may still be entirely new.

If you suggest copying the request OPT into a new message; that might work, but you're not
supposed to reuse opt RRs like that, as they are hop-by-hop...

chrisohaver:
> yes, but that may still be entirely new. If you suggest copying the request OPT into a new message; that might work

It could be, but afaict, it is not implemented in that way in .. at least file and etcd/k8s.

> but you're not supposed to reuse opt RRs like that, as they are hop-by-hop...

Yes - that _seems_ broken in the case of server-side CNAME lookups, but that _is_ what the RFC says.  I guess if the upstream requires the ENDS options set by the client, this server-side lookup would fail, triggering the client to do this lookup itself, in which case forward would retain the OPT record and pass it upstream.  So it works in the end, just not directly.

chrisohaver:
> Yes - that _seems_ broken in the case of server-side CNAME lookups, but that _is_ what the RFC says.

For sake of discussion... I suppose that to retain options here in an RFC compliant way, the decision to copy EDNS options forward to the next hop should be made on a per option basis, defaulting to not copy.  IOW, Upstream could be made aware of the hop policy of any EDNS options it expects see (in simplest form, I think, a list of EDNS opt codes that should be copied forward).

chrisohaver (approved):

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] pkg/upstream:..." ]
>    Yes - that seems broken in the case of server-side CNAME lookups, but that
>    is what the RFC says.
>
>For sake of discussion... I suppose that to retain options here in an RFC
>compliant way, the decision to copy EDNS options forward to the next hop should
>be made on a per option basis, defaulting to not copy. IOW, Upstream could be
>made aware of the hop policy of any EDNS options it expects see (in simplest
>form, I think, a list of EDNS opt codes that should be copied forward).

think this also comes down how to cache (whole packets), with just RRsets you have some
more leeway. I probably spend some cycles to rework the caching to do RRsets.

PR #4078, opened 2020-08-17
Add pkg/dnsutil.RRsets function

miekg:
This function returns the rrset from a slice of rrs. Lifted from the
dnssec plugin (which is now the only user).

This is *also* needed if we're doing per RRset caching which prompted
this PR.

Pros for merging: this function is now tested
Cons: might be less useful than I think


miekg:
premature optimization

PR #4079, opened 2020-08-18
Add proto and client-ip tags to trace plugin

adammw:
Add proto and client-ip tags to traces from the trace plugin. Not sure if there a limit or cost to adding tags, but these ones seem useful to analyse if certain types of requests are slower than others, or to identify which client a particular request comes from.

N/A

I don't see the tag keys documented anywhere, perhaps they should be?

No.

### DIFF
--- a/plugin/trace/trace.go
+++ b/plugin/trace/trace.go
@@ -26,6 +26,8 @@ const (
 	tagName  = "coredns.io/name"
 	tagType  = "coredns.io/type"
 	tagRcode = "coredns.io/rcode"
+	tagProto = "coredns.io/proto"
+	tagClientIP = "coredns.io/client-ip"
 )
 
 type trace struct {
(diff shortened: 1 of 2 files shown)


plugin/trace/trace.go
```diff
@@ -26,6 +26,8 @@ const (
 	tagName  = "coredns.io/name"
 	tagType  = "coredns.io/type"
 	tagRcode = "coredns.io/rcode"
+	tagProto = "coredns.io/proto"
+	tagClientIP = "coredns.io/client-ip"
```
miekg, line 30: to match the log plugin and others this should probably be renamed to:

```suggestion
	tagRemote = "coredns.io/remote"
```

miekg:
regardless of costs, these looks like useful additions

miekg:
ping?

johnbelamaric:
this looks good to me, pending the fix Miek mentioned. cost is not a huge issue as tracing every query is already very high cost - the expectation is that you trace a small subset of queries for statistical analysis

another feature that would be really cool would be an EDNS0 option that you could set on the query to turn on tracing. not a great idea in an open server or prod due to DDoS but could be a nice debugging tool in a dev / test env

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add proto and..." ]
>another feature that would be really cool would be an EDNS0 option that you
>could set on the query to turn on tracing. not a great idea in an open server
>or prod due to DDoS but could be a nice debugging tool in a dev / test env

that's a really cool idea!

miekg:
closed via #4159

PR #4085, opened 2020-08-28
cache: default to DNSSEC

### LINKED ISSUE #3836: Cache plugin does not work when DO bit is set
**What happened**:
DNS queries with DO bit set will miss cache 100% of the time

CoreDNS to serve subsequent queries from cache

Running `dig @<core-dns-ip> +dnssec google.com` results in a cache miss while removing the `+dnssec` will result in a cache hit. After debugging, https://github.com/coredns/coredns/blob/master/plugin/cache/cache.go#L78 is always false while https://github.com/coredns/coredns/blob/master/plugin/cache/handler.go#L100 reflects the correct value set by the client.

- the version of CoreDNS: master (https://github.com/coredns/coredns/commit/8013925b532d8a9e0d94a583c686a483cb665b80)
Debug logging:
```
[INFO] plugin/cache: handler.go: qname=google.com., qtype=1, do=true
[INFO] plugin/cache: looking up key in cache: 15668015455082718403
[INFO] plugin/cache: cache miss
[INFO] plugin/cache: cache.go: qname=google.com., qtype=1, do=false
[INFO] plugin/cache: [IN google.com.] key: 3507767526084950336 - ttl: 3m20s
[INFO] 127.0.0.1:55203 - 18136 "A IN google.com. udp 39 true 4096" NOERROR qr,rd,ra 54 0.004500154s
```

miekg:
This change does away with the DNS/DNSSEC distinction the cache
currently makes. Cache will always make coredns perform a DNSSEC query
and store that result. If a client just needs plain DNS, the DNSSEC
records are stripped from the response.

It should also be more memory efficient, because we store a reply once
and not one DNS and another for DNSSEC.

Fixes: #3836

### DIFF
--- a/plugin/cache/handler.go
+++ b/plugin/cache/handler.go
@@ -26,7 +26,7 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 	server := metrics.WithServer(ctx)
 
 	// On cache miss, if the request has the OPT record and the DO bit set we leave the message as-is. If there isn't a DO bit
-	// set we will modify the request to _add_ one. This means we will always do DNSSEC lookups on cache misses. 
+	// set we will modify the request to _add_ one. This means we will always do DNSSEC lookups on cache misses.
 	// When writing to cache, any DNSSEC RRs in the response are written to cache with the response.
 	// When sending a response to a non-DNSSEC client, we remove DNSSEC RRs from the response. We use a 2048 buffer size, which is
 	// less than 4096 (and older default) and more than 1024 which may be too small. We might need to tweaks this
--- a/plugin/cache/handler.go
+++ b/plugin/cache/handler.go
@@ -15,6 +15,7 @@ import (
 // ServeDNS implements the plugin.Handler interface.
 func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 	state := request.Request{W: w, Req: r}
+	do := state.Do()
 
 	zone := plugin.Zones(c.Zones).Matches(state.Name())
 	if zone == "" {
@@ -22,8 +23,13 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 	}
 
 	now := c.now().UTC()
-
 	server := metrics.WithServer(ctx)
+
+	// If the request has the OPT record and the DO bit set we leave the message as-is. If there isn't a DO bit
+	// set we will modify the request to _add_ one. This means we will always do DNSSEC lookups. When the reply
+	// comes back we will remove DNSSEC RRs to make non-DNSSEC clients happy. We use a 2048 buffer size, which is
+	// less than 4096 (and older default) and more than 1024 which may be too small. We might need to tweaks this
+	// value to be smaller still to prevent UDP fragmentation?
 
 	ttl := 0
 	i := c.getIgnoreTTL(now, state, server)
@@ -31,6 +37,9 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 		ttl = i.ttl(now)
 	}
 	if i == nil {
+		if !do {
+			setDo(r)
+		}
 		crr := &ResponseWriter{ResponseWriter: w, Cache: c, state: state, server: server}
 		return plugin.NextOrFailure(c.Name(), c.Next, ctx, crr, r)
 	}
@@ -40,11 +49,14 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 		now = now.Add(time.Duration(ttl) * time.Second)
 		go func() {
 			r := r.Copy()
+			if !do {
+				setDo(r)
+			}
 			crr := &ResponseWriter{Cache: c, state: state, server: server, prefetch: true, remoteAddr: w.LocalAddr()}
 			plugin.NextOrFailure(c.Name(), c.Next, ctx, crr, r)
 		}()
 	}
-	resp := i.toMsg(r, now)
+	resp := i.toMsg(r, now, do)
 	w.WriteMsg(resp)
 
 	if c.shouldPrefetch(i, now) {
@@ -80,7 +92,7 @@ func (c *Cache) shouldPrefetch(i *item, now time.Time) bool {
 func (c *Cache) Name() string { return "cache" }
 
 func (c *Cache) get(now time.Time, state request.Request, server string) (*item, bool) {
-	k := hash(state.Name(), state.QType(), state.Do())
+	k := hash(state.Name(), state.QType())
 
 	if i, ok := c.ncache.Get(k); ok && i.(*item).ttl(now) > 0 {
 		cacheHits.WithLabelValues(server, Denial).Inc()
@@ -97,7 +109,7 @@ func (c *Cache) get(now time.Time, state request.Request, server string) (*item,
 
 // getIgnoreTTL unconditionally returns an item if it exists in the cache.
 func (c *Cache) getIgnoreTTL(now time.Time, state request.Request, server string) *item {
-	k := hash(state.Name(), state.QType(), state.Do())
+	k := hash(state.Name(), state.QType())
 
 	if i, ok := c.ncache.Get(k); ok {
 		ttl := i.(*item).ttl(now)
@@ -118,7 +130,7 @@ func (c *Cache) getIgnoreTTL(now time.Time, state request.Request, server string
 }
 
 func (c *Cache) exists(state request.Request) *item {
-	k := hash(state.Name(), state.QType(), state.Do())
+	k := hash(state.Name(), state.QType())
 	if i, ok := c.ncache.Get(k); ok {
 		return i.(*item)
 	}
@@ -127,3 +139,14 @@ func (c *Cache) exists(state request.Request) *item {
 	}
 	return nil
 }
+
+func setDo(m *dns.Msg) {
+	o := &dns.OPT{Hdr: dns.RR_Header{Name: ".", Rrtype: dns.TypeOPT}}
+	o.SetDo()
+	o.SetUDPSize(defaultUDPBufSize)
+	m.Extra = append(m.Extra, o)
+}
+
+// defaultUDPBufsize is the bufsize the cache plugin uses on outgoing requests that don't
+// have an OPT RR.
+const defaultUDPBufSize = 2048
--- a/plugin/cache/handler.go
+++ b/plugin/cache/handler.go
@@ -140,8 +140,16 @@ func (c *Cache) exists(state request.Request) *item {
 	return nil
 }
 
+// setDo sets the DO bit and UDP buffer size in the message m.
 func setDo(m *dns.Msg) {
-	o := &dns.OPT{Hdr: dns.RR_Header{Name: ".", Rrtype: dns.TypeOPT}}
+	o := m.IsEdns0()
+	if o != nil {
+		o.SetDo()
+		o.SetUDPSize(defaultUDPBufSize)
+		return
+	}
+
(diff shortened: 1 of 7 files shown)


plugin/cache/handler.go
```diff
@@ -127,3 +139,14 @@ func (c *Cache) exists(state request.Request) *item {
 	}
 	return nil
 }
+
+func setDo(m *dns.Msg) {
+	o := &dns.OPT{Hdr: dns.RR_Header{Name: ".", Rrtype: dns.TypeOPT}}
+	o.SetDo()
+	o.SetUDPSize(defaultUDPBufSize)
+	m.Extra = append(m.Extra, o)
```
chrisohaver, line 156: If an OPT record is present in the response without DO set this would result in two OPT records in the request.

plugin/cache/item.go
```diff
@@ -72,23 +72,60 @@ func (i *item) toMsg(m *dns.Msg, now time.Time) *dns.Msg {
...
+	m1.Extra = m1.Extra[:j]
 	return m1
 }
 
 func (i *item) ttl(now time.Time) int {
 	ttl := int(i.origTTL) - int(now.UTC().Sub(i.stored).Seconds())
 	return ttl
 }
+
+// filterDNSSEC filters the dns.RR slice and removes DNSSEC records. This removes:
```
chrisohaver: Comment is for `filterDNSSEC`, function is `isDNSSEC`.
Seems like implementing `filterDNSSEC` might reduce lines of code a little.

plugin/cache/handler.go
```diff
@@ -15,22 +15,31 @@ import (
...
 
 	now := c.now().UTC()
-
 	server := metrics.WithServer(ctx)
 
+	// If the request has the OPT record and the DO bit set we leave the message as-is. If there isn't a DO bit
+	// set we will modify the request to _add_ one. This means we will always do DNSSEC lookups. When the reply
+	// comes back we will remove DNSSEC RRs to make non-DNSSEC clients happy. We use a 2048 buffer size, which is
+	// less than 4096 (and older default) and more than 1024 which may be too small. We might need to tweaks this
+	// value to be smaller still to prevent UDP fragmentation?
```
chrisohaver:
```suggestion
	// If the request has the OPT record and the DO bit set we leave the message as-is. If there isn't a DO bit
	// set we will modify the request to _add_ one. This means we will always do DNSSEC lookups. When the reply
	// comes back we will remove DNSSEC RRs to make non-DNSSEC clients happy. We use a 2048 buffer size, which is
	// less than 4096 (and older default) and more than 1024 which may be too small. We might need to tweaks this
	// value to be smaller still to prevent UDP fragmentation?
```
```suggestion
	// On cache miss, if the request has the OPT record and the DO bit set we leave the message as-is. If there isn't a DO bit
	// set we will modify the request to _add_ one. This means we will always do DNSSEC lookups on cache misses. 
	// When writing to cache, any DNSSEC RRs in the response are written to cache with the response.
	// When sending a response to a non-DNSSEC client, we remove DNSSEC RRs from the response. We use a 2048 buffer size, which is
	// less than 4096 (and older default) and more than 1024 which may be too small. We might need to tweaks this
	// value to be smaller still to prevent UDP fragmentation?
```

chrisohaver (changes requested):
There is some overhead added for non dnssec clients, which always applies and cannot be disabled.
But it only occurs on cache misses, So I don't _think_ it should impact performance in any big way.

plugin/cache/handler.go
miekg: good catch. The check before this only checks the Do bit which leaves the possibility of having an opt without Do, then we would add another. will fiix

plugin/cache/item.go
miekg: ack. let me change this - and now looking at it, I think capping at `j` is actually wrong here...

miekg: nope 'j` is correct, the loops where we use this aren't identical; one does r.Copy() the otherone doesn't. I'll fix the comment

plugin/cache/item.go
```diff
@@ -111,9 +111,9 @@ func (i *item) ttl(now time.Time) int {
 	return ttl
 }
 
-// filterDNSSEC filters the dns.RR slice and removes DNSSEC records. This removes:
-// NSEC,NSEC3,DS and RRSIG/SIG records. DNSKEYs are left alone on the assumptions that if they are
-// there the client explictly asked for them.
+// isDNSSEC return true id r is a DNSSEC record. NSEC,NSEC3,DS and RRSIG/SIG
```
chrisohaver:
```suggestion
// isDNSSEC returns true if r is a DNSSEC record. NSEC,NSEC3,DS and RRSIG/SIG
```

plugin/cache/item.go
```diff
@@ -111,9 +111,9 @@ func (i *item) ttl(now time.Time) int {
...
+// are DNSSEC records. DNSKEYs are left alone on the assumptions that if they
+// are there the client explictly asked for them.
```
chrisohaver: "DNSKEYs are left alone ..."  sentence doesn't apply to this function (but would to a "filter" function)

chrisohaver (changes requested):

plugin/cache/item.go
miekg: done

stp-ip (approved):
~lgtm~

Grr sorry Github UI mishap ^^

miekg:
all comments addressed. PTAL

plugin/cache/handler.go
```diff
@@ -15,22 +15,32 @@ import (
...
 	if zone == "" {
 		return plugin.NextOrFailure(c.Name(), c.Next, ctx, w, r)
 	}
 
 	now := c.now().UTC()
-
 	server := metrics.WithServer(ctx)
 
+	// On cache miss, if the request has the OPT record and the DO bit set we leave the message as-is. If there isn't a DO bit
+	// set we will modify the request to _add_ one. This means we will always do DNSSEC lookups on cache misses.
```
chrisohaver:
```suggestion
	// set we will modify the request to _add_ one. This means we will always do DNSSEC lookups on cache misses.
```

elliotfehr (approved):

chrisohaver (approved):

igcherkaev:
Hello.

We believe either this PR or https://github.com/coredns/coredns/pull/4055 has broken DNS resolution for certain domains.
It looks like CoreDNS would intermittently disregard +noedns flag and serve incomplete record (probably from cache?).
Reverting to 1.7.0 immediately fixes the issue.

Here's what I am talking about (where `10.23.136.9` is our pod of coredns):

```
bash-4.1$ dig +nodnssec +noedns @10.23.136.9 account.docusign.com

; <<>> DiG 9.8.2rc1-RedHat-9.8.2-0.68.rc1.el6_10.1 <<>> +nodnssec +noedns @10.23.136.9 account.docusign.com
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 43409
;; flags: qr rd ra; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 2048
;; QUESTION SECTION:
;account.docusign.com.		IN	A

;; ANSWER SECTION:
account.docusign.com.	16	IN	CNAME	account-geo.docusign.com.akadns.net.
account.docusign.com.	16	IN	RRSIG	CNAME 8 3 30 20201022163011 20201019153011 47556 docusign.com. SdDoe7IbyMar7i3bzZB6NJqEAvvmfBXFRhHg/F9QM4YEcDRVJMu2bEj1 JpbcQjsCPkQqUnLfRaW+Q+2XTUWCxAOMQlVndWmW4XyATxIHKRnz8mL6 BlnfIhJFxHnVEe+1ugUfgq3HYtrQWgetcvXJugtEbRLA3UskEZL3NoKx dTw=
account-geo.docusign.com.akadns.net. 16	IN CNAME account-eu.docusign.com.akadns.net.

;; Query time: 92 msec
;; SERVER: 10.23.136.9#53(10.23.136.9)
;; WHEN: Mon Oct 19 18:21:41 2020
;; MSG SIZE  rcvd: 393

```

Note the OPT section? 

Eventually it would respond properly after few tries:
```
bash-4.1$ dig +nodnssec +noedns @10.23.136.9 account.docusign.com

; <<>> DiG 9.8.2rc1-RedHat-9.8.2-0.68.rc1.el6_10.1 <<>> +nodnssec +noedns @10.23.136.9 account.docusign.com
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 794
;; flags: qr rd ra; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 2048
;; QUESTION SECTION:
;account.docusign.com.		IN	A

;; ANSWER SECTION:
account.docusign.com.	29	IN	CNAME	account-geo.docusign.com.akadns.net.
account.docusign.com.	29	IN	RRSIG	CNAME 8 3 30 20201022163011 20201019153011 47556 docusign.com. SdDoe7IbyMar7i3bzZB6NJqEAvvmfBXFRhHg/F9QM4YEcDRVJMu2bEj1 JpbcQjsCPkQqUnLfRaW+Q+2XTUWCxAOMQlVndWmW4XyATxIHKRnz8mL6 BlnfIhJFxHnVEe+1ugUfgq3HYtrQWgetcvXJugtEbRLA3UskEZL3NoKx dTw=
account-geo.docusign.com.akadns.net. 29	IN CNAME account-eu.docusign.com.akadns.net.

;; Query time: 43 msec
;; SERVER: 10.23.136.9#53(10.23.136.9)
;; WHEN: Mon Oct 19 18:22:17 2020
;; MSG SIZE  rcvd: 393

bash-4.1$ dig +nodnssec +noedns @10.23.136.9 account.docusign.com

; <<>> DiG 9.8.2rc1-RedHat-9.8.2-0.68.rc1.el6_10.1 <<>> +nodnssec +noedns @10.23.136.9 account.docusign.com
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 32079
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;account.docusign.com.		IN	A

;; ANSWER SECTION:
account.docusign.com.	26	IN	CNAME	account-geo.docusign.com.akadns.net.
account-geo.docusign.com.akadns.net. 26	IN CNAME account-eu.docusign.com.akadns.net.
account-eu.docusign.com.akadns.net. 26 IN A	185.81.100.38

;; Query time: 0 msec
;; SERVER: 10.23.136.9#53(10.23.136.9)
;; WHEN: Mon Oct 19 18:22:20 2020
;; MSG SIZE  rcvd: 240

```

All logs are clear, no error or anything. Access logs:
```
coredns-85667845f6-5vzxv coredns [INFO] 10.23.237.242:46005 - 43409 "A IN account.docusign.com. udp 49 true 2048" NOERROR qr,rd,ra 394 0.091970889s
coredns-85667845f6-5vzxv coredns [INFO] 10.23.237.242:58404 - 794 "A IN account.docusign.com. udp 49 true 2048" NOERROR qr,rd,ra 394 0.041991402s
coredns-85667845f6-5vzxv coredns [INFO] 10.23.237.242:35732 - 32079 "A IN account.docusign.com. udp 38 false 512" NOERROR qr,aa,rd,ra 240 0.000067694s

```

Note the DO flag.

Our configuration:
```
    .:53 {
        errors
        log
        debug
        health
        kubernetes dep.example.com 10.23.0.0/16 {
          pods verified
        }
        prometheus :9153
        forward . /etc/resolv.conf {
          max_fails 0
        }
        cache 60
        autopath @kubernetes
        loop
        reload
    }
```
I could not find any way to make it behave as in 1.7.0, so we had to rollback for now. Any ideas/thoughts?

igcherkaev:
OK, there's already an issue opened for this https://github.com/coredns/coredns/issues/4189

PR #4096, opened 2020-08-31
clouddns,route53: fix lingering goroutines after restart

### LINKED ISSUE #3815: USR1 signal still spawns additional loops on plugins that use refresh
**What happened**:

Using the `route53` plugin (others like `azure` and `clouddns` probably as well), can spawn an additional concurrent refresh loop for every USR1 signal received.

I added a debug line to the route53 updateZones function:

```diff
@@ -238,6 +242,7 @@ func updateZoneFromRRS(rrs *route53.ResourceRecordSet, z *file.Zone) error {
 func (h *Route53) updateZones(ctx context.Context) error {
 	errc := make(chan error)
 	defer close(errc)
+	fmt.Println(time.Now().String(), "running refresh loop: ", h.zones)
 	for zName, z := range h.zones {
 		go func(zName string, z []*zone) {
 			var err error
```

My debugging output is recorded here w/ 3 server blocks configured with the route53 plugin and configured for a 5 minute refresh interval. https://gist.github.com/mattlqx/7e8126c42a02cf6caa0982adb8fb8207

A loop is started when the server is started. I let it it refresh once normal, which happens at 5 minutes, then shortly there after I sent a USR1 signal. Another refresh occurs, and then approximately 4.5 minutes and 5 minutes after that you can see both loops refresh again.

USR1 should only refresh configuration information and clean-up any existing loops to get external data that a plugin may have open.

Configure a route53 plugin to get zone from Route 53.
Add a debugging print statement.
Run CoreDNS and observe the logging statements.
Send USR1 as many times as you like to observe n loops occurring in the background.

There was an underlying problem with Caddy a while ago that seemed to be related to the issue. https://github.com/coredns/coredns/issues/2396 But CoreDNS now uses the fixed version of Caddy and the problem still exists. I'm not really sure how to debug/fix further. I do note that this log statement is never printed: https://github.com/coredns/coredns/blob/master/plugin/route53/route53.go#L90. So I would imagine the context's Done channel is never closing.

**Environment**:

- the version of CoreDNS: 1.6.9
- Corefile:
```
(cacheandlog) {
  prometheus
  cache 300
  errors
  log . {
    class error
  }
}

example.com {
  route53 example.com.:Z1GCMBPMI21YYY example.com.:Z3IZZERHBUPZZZ {
    credentials default /configs/credentials
    refresh 5m
    fallthrough
  }
  forward . 172.30.13.132 172.30.14.132
  import cacheandlog
}

16.172.in-addr.arpa {
  route53 16.172.in-addr.arpa.:Z33QPSXQCQT8W1 {
    credentials default /configs/credentials
    refresh 5m
    fallthrough
  }
  forward . 172.30.13.132 172.30.14.132
  import cacheandlog
}

30.172.in-addr.arpa {
  route53 30.172.in-addr.arpa.:Z1CYHZ3A0GBKDI {
    credentials default /configs/credentials
    refresh 5m
    fallthrough
  }
  forward . 172.30.13.132 172.30.14.132
  import cacheandlog
}
```
- OS (e.g: `cat /etc/os-release`): PRETTY_NAME="Debian GNU/Linux 10 (buster)" (Golang 1.14.1 docker container)

miekg:
Stop the context so the refresh loop terminates on restart.

Fixes: #3815

### DIFF
--- a/plugin/clouddns/setup.go
+++ b/plugin/clouddns/setup.go
@@ -46,14 +46,14 @@ func setup(c *caddy.Controller) error {
 		for i := 0; i < len(args); i++ {
 			parts := strings.SplitN(args[i], ":", 3)
 			if len(parts) != 3 {
-				return c.Errf("invalid zone '%s'", args[i])
+				return plugin.Error("clouddns", c.Errf("invalid zone %q", args[i]))
 			}
 			dnsName, projectName, hostedZone := parts[0], parts[1], parts[2]
 			if dnsName == "" || projectName == "" || hostedZone == "" {
-				return c.Errf("invalid zone '%s'", args[i])
+				return plugin.Error("clouddns", c.Errf("invalid zone %q", args[i]))
 			}
 			if _, ok := keyPairs[args[i]]; ok {
-				return c.Errf("conflict zone '%s'", args[i])
+				return plugin.Error("clouddns", c.Errf("conflict zone %q", args[i]))
 			}
 
 			keyPairs[args[i]] = struct{}{}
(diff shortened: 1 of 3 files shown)


yongtang (approved):
/lgtm

PR #4108, opened 2020-09-07
plugin/forward: init ClientSessionCache in tls.Config

rdrozhdzh:
initialize ClientSessionCache in tls.Config. This may speed up a TLS handshake in upcoming connections to the same TLS server, i.e. the main reason is increasing performance

none
none
none

### DIFF
--- a/plugin/forward/setup.go
+++ b/plugin/forward/setup.go
@@ -118,6 +118,9 @@ func parseStanza(c *caddy.Controller) (*Forward, error) {
 	if f.tlsServerName != "" {
 		f.tlsConfig.ServerName = f.tlsServerName
 	}
+
+	// Initialize ClientSessionCache in tls.Config. This may speed up a TLS handshake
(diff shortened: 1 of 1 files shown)


plugin/forward/setup.go
```diff
@@ -117,6 +118,8 @@ func parseStanza(c *caddy.Controller) (*Forward, error) {
 	if f.tlsServerName != "" {
 		f.tlsConfig.ServerName = f.tlsServerName
 	}
+	f.tlsConfig.ClientSessionCache = tls.NewLRUClientSessionCache(len(f.proxies))
```
miekg, line 124:
```suggestion
        // Initialize ClientSessionCache in tls.Config. This may speed up a TLS handshake in upcoming connections to the same TLS server.
	f.tlsConfig.ClientSessionCache = tls.NewLRUClientSessionCache(len(f.proxies))
```

chrisohaver (approved):

PR #4110, opened 2020-09-08
.gitignore: add coverage output

zhijianli88:
delete these files when make clean as well

Signed-off-by: Li Zhijian <lizhijian@cn.fujitsu.com>


zouyee:
https://github.com/coredns/coredns/pull/3820

PR #4115, opened 2020-09-11
plugin/forward: fix panic when `expire` is configured as 0s

rdrozhdzh:
Signed-off-by: Ruslan Drozhdzh <rdrozhdzh@infoblox.com>

fixed panic when `expire` is configured as 0s. Instead, the ticker will be initialized with default value. This will also prevent calling `cleanup()` func too often.
none
none
no

### DIFF
--- a/plugin/forward/persistent.go
+++ b/plugin/forward/persistent.go
@@ -44,7 +44,7 @@ func newTransport(addr string) *Transport {
 
 // connManagers manages the persistent connection cache for UDP and TCP.
 func (t *Transport) connManager() {
-	ticker := time.NewTicker(t.expire)
+	ticker := time.NewTicker(defaultExpire)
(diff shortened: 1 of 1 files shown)


miekg:
probably cleaner to do this in the setup function, `t.expire` is also used throughout the code, so this fix looks incomplete.

rdrozhdzh:
`t.expire` configured as `0s` is not an error by itself. This value means that forward plugin will not use the cached connections as they are getting expired as soon as they are put to the cache. This behavior may look unpractical but it may be useful for running some tests.

My fix intentionally affects only the period of invocation of cleanup procedure which removes the expired connections and closes them.

miekg:
ok fair.

miekg (approved):

PR #4131, opened 2020-09-18
plugin/hosts: make hosts plugin cursive

miekg:
nit-picking PR, when using the name of the plugin make it cursive.


yongtang (approved):
lgtm

PR #4132, opened 2020-09-18
plugin/records: a new plugin

miekg:
This plugin serves records specified inline from the config file.

### DIFF
--- a/core/dnsserver/zdirectives.go
+++ b/core/dnsserver/zdirectives.go
@@ -39,6 +39,7 @@ var Directives = []string{
 	"template",
 	"transfer",
 	"hosts",
+	"records",
 	"route53",
 	"azure",
 	"clouddns",
--- a/core/plugin/zplugin.go
+++ b/core/plugin/zplugin.go
@@ -38,6 +38,7 @@ import (
 	_ "github.com/coredns/coredns/plugin/nsid"
 	_ "github.com/coredns/coredns/plugin/pprof"
 	_ "github.com/coredns/coredns/plugin/ready"
+	_ "github.com/coredns/coredns/plugin/records"
 	_ "github.com/coredns/coredns/plugin/reload"
 	_ "github.com/coredns/coredns/plugin/rewrite"
 	_ "github.com/coredns/coredns/plugin/root"
--- a/plugin.cfg
+++ b/plugin.cfg
@@ -48,6 +48,7 @@ autopath:autopath
 template:template
 transfer:transfer
 hosts:hosts
+records:records
 route53:route53
 azure:azure
 clouddns:clouddns
--- a/plugin/records/README.md
+++ b/plugin/records/README.md
@@ -1,0 +1,69 @@
+# records
+
+## Name
+
+*records* - enables serving zone data directly from the Corefile.
+
+## Description
+
+The *records* plugin is useful for serving zone data that is specified inline in the configuration
+file. As opposed to the *hosts* plugin, this plugin supports **all** record types.
+If no TTL is specified in the records, a default TTL of 3600s is assumed. For negative responses a
+SOA record should be included in the response, this will only be done when a SOA record is included
+in the data.
+
+Currently not implemented is DNSSEC. If RRSIG records are added they will not be returned in the
+reply even if the client is capable of handling them. If you need signed replies use the *dnssec*
+plugin in conjunction with this one.
+
+This plugin can only be used once per Server Block.
+
+## Syntax
+
+~~~
+records [ZONES...] {
+    [INLINE]
+}
+~~~
+
+* **ZONES** zones it should be authoritative for. If empty, the zones from the configuration block
+  are used.
+* **INLINE** the resource record that are to be served. These must be specified as the text
+  represenation of the record, see the examples below. Each record must be on a single line.
+
+If domain name in **INLINE** are not fully qualifed each of the **ZONES** are used as the origin and
+added to the names.
+
+## Examples
+
+Serve a MX records for example.org *and* give the MX server the name `mx1` and address 127.0.0.1.
+
+~~~ corefile
+example.org {
+    records {
+        @   60  IN SOA ns.icann.org. noc.dns.icann.org. 2020091001 7200 3600 1209600 3600
+        @   60  IN MX 10 mx1
+        mx1 60  IN A  127.0.0.1
+    }
+}
+~~~
+
+Create 2 zones, each will have a MX record. Note the no SOA record has been given.
+
+~~~
+. {
+    records example.org example.net {
+        mx1 IN MX 10 mx1
+    }
+}
+~~~
+
+## Bugs
+
+DNSSEC is not implemented.
+
+## See Also
+
+See the *hosts*' plugin documentation if you just need to return address records. Use the *reload*
+plugin to reload the contents of these inline records automatically when they are changed. The
+*dnssec* plugin can be used to sign replies.
--- a/plugin/records/README.md
+++ b/plugin/records/README.md
@@ -7,14 +7,19 @@
 ## Description
 
 The *records* plugin is useful for serving zone data that is specified inline in the configuration
-file. As opposed to the *hosts* plugin, this plugin supports **all** record types.
-If no TTL is specified in the records, a default TTL of 3600s is assumed. For negative responses a
-SOA record should be included in the response, this will only be done when a SOA record is included
-in the data.
+file. As opposed to the *hosts* plugin, this plugin supports **all** record types. Records need to
+be specifed in text representation as specified in RFC 1035. If no TTL is specified in the records,
+a default TTL of 3600s is assumed.
+
+For negative responses a SOA record should be included in the response, this will only be done when
+a SOA record is included in the data.
 
 Currently not implemented is DNSSEC. If RRSIG records are added they will not be returned in the
 reply even if the client is capable of handling them. If you need signed replies use the *dnssec*
 plugin in conjunction with this one.
+
+Note the *host* plugin is configured before *records* in `plugin.cfg`, which means that when both
+are being specified in a server block, the *host* plugin will get preference.
 
 This plugin can only be used once per Server Block.
 
@@ -27,9 +32,10 @@ records [ZONES...] {
 ~~~
 
 * **ZONES** zones it should be authoritative for. If empty, the zones from the configuration block
-  are used.
+   are used.
 * **INLINE** the resource record that are to be served. These must be specified as the text
-  represenation of the record, see the examples below. Each record must be on a single line.
+   represenation (as specifed in RFC 1035) of the record. See the examples below. Each record must
+   be on a single line.
 
 If domain name in **INLINE** are not fully qualifed each of the **ZONES** are used as the origin and
 added to the names.
@@ -66,4 +72,6 @@ DNSSEC is not implemented.
 
 See the *hosts*' plugin documentation if you just need to return address records. Use the *reload*
 plugin to reload the contents of these inline records automatically when they are changed. The
-*dnssec* plugin can be used to sign replies.
+*dnssec* plugin can be used to sign replies. See RFC 1035 and subsequent RFCs defining new record
+types for the text representation that must be used in this plugin. Note RFC 3597 (Handling of
+Unknown DNS Resource Record) syntax is also supported.
--- a/plugin/records/README.md
+++ b/plugin/records/README.md
@@ -2,7 +2,7 @@
 
 ## Name
 
-*records* - enables serving zone data directly from the Corefile.
+*records* - enables serving (basic) zone data directly from the Corefile.
 
 ## Description
 
@@ -75,7 +75,8 @@ RFC 1035 zone file and everything after it will be ignored, hence the need for q
 
 ## Bugs
 
-DNSSEC, nor wildcards are implemented. The lookup algorithm is pretty basic.
+DNSSEC, nor wildcards are implemented. The lookup algorithm is pretty basic. Future enhancements
+could leverage the code from the *file* plugin to make more compliant with the DNS specification.
 
 ## See Also
 
--- a/plugin/records/README.md
+++ b/plugin/records/README.md
@@ -14,9 +14,15 @@ a default TTL of 3600s is assumed.
 For negative responses a SOA record should be included in the response, this will only be done when
 a SOA record is included in the data.
 
-Currently not implemented is DNSSEC. If RRSIG records are added they will not be returned in the
-reply even if the client is capable of handling them. If you need signed replies use the *dnssec*
-plugin in conjunction with this one.
+The *records* plugin uses a simple algorithm for find the correct record(s) to return. This means
+some more advanced features are currently not available, such as:
+
+* DNSSEC, if RRSIG records are added they will not be returned in the reply even if the client is
+  capable of handling them. If you need signed replies use the *dnssec* plugin in conjunction with
+  this one.
+* Wildcards, i.e. `*.example.org`, will not be detected as a wildcard record.
+
+If you need a more robust implementation you probably want to use the *file* plugin.
 
 Note the *host* plugin is configured before *records* in `plugin.cfg`, which means that when both
 are being specified in a server block, the *host* plugin will get preference.
@@ -69,7 +75,7 @@ RFC 1035 zone file and everything after it will be ignored, hence the need for q
 
 ## Bugs
 
-DNSSEC is not implemented.
+DNSSEC, nor wildcards are implemented. The lookup algorithm is pretty basic.
 
 ## See Also
 
@@ -78,3 +84,5 @@ plugin to reload the contents of these inline records automatically when they ar
 *dnssec* plugin can be used to sign replies. See RFC 1035 and subsequent RFCs defining new record
 types for the text representation that must be used in this plugin. Note RFC 3597 (Handling of
 Unknown DNS Resource Record) syntax is also supported.
+
+Use the *file* plugin for a more fully featured DNS implementation (including DNSSEC).
--- a/plugin/records/README.md
+++ b/plugin/records/README.md
@@ -54,12 +54,15 @@ example.org {
 }
 ~~~
 
-Create 2 zones, each will have a MX record. Note the no SOA record has been given.
+Create 2 zones, each will have a MX record. Note that no SOA record has been given. Also note you
+need to quote the `;` in the TXT record's data to make the parser happy. (A `;` is a comment in a
+RFC 1035 zone file and everything after it will be ignored, hence the need for quoting it here.)
 
 ~~~
 . {
     records example.org example.net {
         mx1 IN MX 10 mx1
+        dkim._domainkey.relay 3600 IN TXT "v=DKIM1\; h=sha256\; k=rsa\; s=email\; p=MIIBIj ..."
     }
 }
 ~~~
--- a/plugin/records/records.go
+++ b/plugin/records/records.go
@@ -1,0 +1,74 @@
+package records
+
+import (
+	"context"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/request"
+
+	"github.com/miekg/dns"
+)
+
+// Records is the plugin handler.
+type Records struct {
+	origins []string // for easy matching, these strings are the index in the map m.
+	m       map[string][]dns.RR
+
+	Next plugin.Handler
+}
+
+func (re *Records) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	state := request.Request{W: w, Req: r}
+	qname := state.Name()
+	zone := plugin.Zones(re.origins).Matches(qname)
+	if zone == "" {
+		return plugin.NextOrFailure(re.Name(), re.Next, ctx, w, r)
+	}
+
+	// New we should have some data for this zone, as we just have a list of RR, iterate through them, find the qname
+	// and see if the qtype exists. If so reply, if not do the normal DNS thing and return either NXDOMAIN or NODATA.
+	m := new(dns.Msg)
+	m.SetReply(r)
+	m.Authoritative = true
+
+	nxdomain := true
+	var soa dns.RR
+	for _, r := range re.m[zone] {
+		if r.Header().Rrtype == dns.TypeSOA && soa == nil {
+			soa = r
+		}
+		if r.Header().Name == qname {
+			nxdomain = false
+			if r.Header().Rrtype == state.QType() {
+				m.Answer = append(m.Answer, r)
+			}
+		}
+	}
+
+	// handle NXDOMAIN, NODATA and normal response here.
+	if nxdomain {
+		m.Rcode = dns.RcodeNameError
+		if soa != nil {
+			m.Ns = []dns.RR{soa}
+		}
+		w.WriteMsg(m)
+		return dns.RcodeSuccess, nil
+	}
+
+	if len(m.Answer) == 0 {
+		if soa != nil {
+			m.Ns = []dns.RR{soa}
+		}
+	}
+
+	w.WriteMsg(m)
+	return dns.RcodeSuccess, nil
+}
+
+func (re *Records) Name() string { return "records" }
+
+func New() *Records {
+	re := new(Records)
+	re.m = make(map[string][]dns.RR)
+	return re
+}
--- a/plugin/records/records.go
+++ b/plugin/records/records.go
@@ -17,6 +17,7 @@ type Records struct {
 	Next plugin.Handler
 }
 
+// ServeDNS implements the plugin.Handle interface.
 func (re *Records) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 	state := request.Request{W: w, Req: r}
 	qname := state.Name()
@@ -65,8 +66,10 @@ func (re *Records) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 	return dns.RcodeSuccess, nil
 }
 
+// Name implements the plugin.Handle interface.
 func (re *Records) Name() string { return "records" }
 
+// New returns a pointer to a new and intialized Records.
 func New() *Records {
 	re := new(Records)
 	re.m = make(map[string][]dns.RR)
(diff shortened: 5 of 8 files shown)


polarathene:
This looks awesome! :grinning: 

I'll be giving it a go shortly, is it expecting RFC 1053 zone file record entries? That makes them easy to port over, as someone who's not super DNS savvy, it might help to mention the RFC in the docs rather than just zones? 

I had a response for the original issue that I forgot to send, I didn't expect to see a PR ready so fast :open_mouth:  I kept the two examples I came up with below if they're useful to consider from a UX point of view?:

<details>
<summary>A) By DNS Record Type</summary>

```corefile
example.org {
    records 300 { # 2nd value sets implicit TTL for records
        A {
            @ 127.0.0.1 # @ uses origin here: example.org
            mail 192.168.1.42 # example.org appended
        }
        # Single record entry, no {} block
        CNAME blog @
        MX {
            # hello@example.org => A mail.example.org 192.168.1.42
            @ 10 mail
            # hello@mail.example.org => A example.org 127.0.0.1
            mail 10 @
            # hello@relay.example.org => A mxa.smtp-relay-service.io
            relay 10 mxa.smtp-relay-service.io.
            relay 20 mxb.smtp-relay-service.io.
        }
        TXT 3600 { # override TTL
            # hello@example.org can only come from IP of A record mail.example.com
            @ "v=spf1 a:mail.example.org -all"
            # hello@mail.example.org is valid when sent from IP of any MX record for mail.example.org
            mail "v=spf1 mx -all"
            # hello@relay.example.org passes when third-party provides valid SPF record at domain with match returned
            relay "v=spf1 include:smtp-relay-service.io ~all"
            dkim._domainkey.relay "v=DKIM1; h=sha256; k=rsa; s=email; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAyPSKcUBKxwMBu594lJpRbUH3aw6jCrFLnHllovU5eFV96qc1bOGAzjYczoDGpWm141720KIL04Ev6vTzIWO2vd7+VVm9CTnCPrrBuhGWuUABWOY36UJfj7gs5BirW/7Tf6C8ksxKE1PnAoDaqn1VpI+GGm+w+N7khEOd9ESNvIMqoiBuzBzOB+r4T5MC8U9rq0m0Y3U5sA+E3s4+uHfaelL4v+Sd9aVcI+ysJKJiKr55Z6XQkBqImMZebXGCyeVKiOyVVAtS/TeDxGv2eu2/K0Cg0AsDYgfW066eRiY5OthPb6tZbuaZbbgryHw5G87R85Jbeuo6cuYDxf3gAIng4QIDAQAB"
        }
        PTR { # Probably better if these could be implicitly generated?
            127.0.0.1 @
            192.168.1.42 mail
        }
    }
}
```

</details>

<details>
<summary>B) By Domain (Seems best)</summary>

```corefile
example.org {
    records 300 { # 2nd value sets implicit TTL for records
        @ { # @ uses origin here: example.org
            A 127.0.0.1
            # hello@example.org => A mail.example.org 192.168.1.42
            MX 10 mail
            # hello@example.org can only come from IP of A record mail.example.com
            TXT "v=spf1 a:mail.example.org -all"
        }
        mail { # example.org appended
            A 192.168.1.42
            # hello@mail.example.org => A example.org 127.0.0.1
            MX 10 @
            # hello@mail.example.org is valid when sent from IP of any MX record for mail.example.org
            TXT "v=spf1 mx -all"
        }
        relay { 
            # hello@relay.example.org => A mxa.smtp-relay-service.io
            MX 10 mxa.smtp-relay-service.io.
            MX 20 mxb.smtp-relay-service.io.
            # hello@relay.example.org passes when third-party provides valid SPF record at domain with match returned
            TXT "v=spf1 include:smtp-relay-service.io ~all"
            # Non-default TTL, sub-label prefix before TXT value, matches to: `dkim._domainkey.relay.example.org`
            3600 TXT dkim._domainkey "v=DKIM1; h=sha256; k=rsa; s=email; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAyPSKcUBKxwMBu594lJpRbUH3aw6jCrFLnHllovU5eFV96qc1bOGAzjYczoDGpWm141720KIL04Ev6vTzIWO2vd7+VVm9CTnCPrrBuhGWuUABWOY36UJfj7gs5BirW/7Tf6C8ksxKE1PnAoDaqn1VpI+GGm+w+N7khEOd9ESNvIMqoiBuzBzOB+r4T5MC8U9rq0m0Y3U5sA+E3s4+uHfaelL4v+Sd9aVcI+ysJKJiKr55Z6XQkBqImMZebXGCyeVKiOyVVAtS/TeDxGv2eu2/K0Cg0AsDYgfW066eRiY5OthPb6tZbuaZbbgryHw5G87R85Jbeuo6cuYDxf3gAIng4QIDAQAB"
        }
        blog {
            CNAME @
        }
        . { # Wildcard subdomain CNAME to A example.org 127.0.0.1
            CNAME @
        }
    }
}
```
</details>

**Pros:**

- Reduces noise on the left (no repetition of domain, no `IN`)
- Easy/clear setting of non-default TTL to fallback to (instead of declaring for entries differing from `3600`?)
- Automatic `PTR`?
- Possible for things like `NS` records using an array of values syntax (depending on parser logic)

**Cons:**

- More code to implement and maintain? Not as generic for handling/supporting records?

It would probably make most sense here to go with B, there's nice UX with the domain grouping of related records, supporting the TXT DKIM record as shown should be ok? A TXT value would be wrapped in quotes and if there is an extra value after the record type, the quoted value should be last, with the first being prepended to the domain label.

---

If either requires to much effort to implement differently (I assume the `file` plugin RFC 1035 parser is being used for this), I wouldn't want to burden you further. Just having the functionality to support these records in the first place is valuable! Thank you.

miekg:
yep, it's 1035 text representation.

"IN" will be needed, because the parsing is line by line - automatic NS/SOA is not done - if you need it you need to specify it yourself. This is hard to automate; so it's garbage in, garbage out (not specifying SOA will lead to incomplete replies for instance).

polarathene:
> "IN" will be needed, because the parsing is line by line

Sorry, I'm not too familiar with `IN` other than it's considered a "class", but seems to be the general default, so my two alternative config styles omit it, if it's relevant to specify a different value I'd assume that'd be better suited as opt-in. Looking at an exported zone file even my SOA and NS records use IN.

> automatic NS/SOA is not done - if you need it you need to specify it yourself.

Perfectly ok :+1: 

> This is hard to automate; so it's garbage in, garbage out (not specifying SOA will lead to incomplete replies for instance).

Is that comment about SOA/NS records only?

> yep, it's 1035 text representation.

That's ok, thank you for making this possible. If I learn Go some day, and creating a plugin for the B style above isn't too difficult I might give it a shot :)

---

I tested this PR via a docker image I wrote, I used the mx1 record example from the docs and tested with `drill` on Linux. 

It worked well, except if I also had a `hosts` plugin active, even if unrelated to what `records` was responsible for it seemed `records` got ignored.

```corefile
test.internal {
    records {
        @   60  IN MX 10 mx1
        mx1 60  IN A  192.168.1.23
    }
    hosts {
        192.168.1.42 sub.test.internal
    }
    log
}
```

<details>
<summary>drill @localhost mx1.test.internal (records)</summary>

```
$ drill @localhost mx1.test.internal
;; ->>HEADER<<- opcode: QUERY, rcode: SERVFAIL, id: 20733
;; flags: qr rd ; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0 
;; QUESTION SECTION:
;; mx1.test.internal.   IN      A

;; ANSWER SECTION:

;; AUTHORITY SECTION:

;; ADDITIONAL SECTION:

;; Query time: 0 msec
;; SERVER: ::1
;; WHEN: Tue Sep 22 19:42:06 2020
;; MSG SIZE  rcvd: 35
```
</details>

<details>
<summary>drill @localhost test.internal MX (records)</summary>

```
$ drill @localhost test.internal MX
;; ->>HEADER<<- opcode: QUERY, rcode: SERVFAIL, id: 33484
;; flags: qr rd ; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0 
;; QUESTION SECTION:
;; test.internal.       IN      MX

;; ANSWER SECTION:

;; AUTHORITY SECTION:

;; ADDITIONAL SECTION:

;; Query time: 1 msec
;; SERVER: 127.0.0.1
;; WHEN: Tue Sep 22 19:41:59 2020
;; MSG SIZE  rcvd: 31
```
</details>

<details>
<summary>drill @localhost sub.test.internal (hosts)</summary>

```
$ drill @localhost sub.test.internal
;; ->>HEADER<<- opcode: QUERY, rcode: NOERROR, id: 15706
;; flags: qr aa rd ; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0 
;; QUESTION SECTION:
;; sub.test.internal.   IN      A

;; ANSWER SECTION:
sub.test.internal.      3600    IN      A       192.168.1.42

;; AUTHORITY SECTION:

;; ADDITIONAL SECTION:

;; Query time: 3 msec
;; SERVER: ::1
;; WHEN: Tue Sep 22 19:53:45 2020
;; MSG SIZE  rcvd: 68
```
</details>

If I remove the `hosts` block it'll be fine. If that's intentional, then the docs should also indicate it's incompatible to use with a `hosts` block as well.

miekg:
good points, I'll polish the documentation a bit more. Thanks for the feedback

polarathene:
TXT records are splitting by whitespace instead of returning the entire string wrapped in quotes. The DKIM record also dropped `;` in it's returned value (I think that is what prevented the remainder of the string from being included unlike the TXT spf example?).

```corefile
example.test {
    records {
        @ 60 IN A 127.0.0.1
        @ 60 IN MX 10 mail
        @ 60 IN TXT "v=spf1 a:mail.example.test -all"

        mail 60 IN A 192.168.1.42
        mail 60 IN MX 10 @
        mail 60 IN TXT "v=spf1 mx -all"

        relay 60 IN MX 10 mxa.smtp-relax-service.io.
        relay 60 IN MX 20 mxb.smtp-relax-service.io.
        relay 60 IN TXT "v=spf1 include:smtp-relay-service.io ~all"
        dkim._domainkey.relay 3600 IN TXT "v=DKIM1; h=sha256; k=rsa; s=email; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAyPSKcUBKxwMBu594lJpRbUH3aw6jCrFLnHllovU5eFV96qc1bOGAzjYczoDGpWm141720KIL04Ev6vTzIWO2vd7+VVm9CTnCPrrBuhGWuUABWOY36UJfj7gs5BirW/7Tf6C8ksxKE1PnAoDaqn1VpI+GGm+w+N7khEOd9ESNvIMqoiBuzBzOB+r4T5MC8U9rq0m0Y3U5sA+E3s4+uHfaelL4v+Sd9aVcI+ysJKJiKr55Z6XQkBqImMZebXGCyeVKiOyVVAtS/TeDxGv2eu2/K0Cg0AsDYgfW066eRiY5OthPb6tZbuaZbbgryHw5G87R85Jbeuo6cuYDxf3gAIng4QIDAQAB"
        
        blog 60 IN CNAME @
        * 60 IN CNAME @
    }
    log
}

42.1.168.192.in-addr.arpa {
    records {
        @ 60 IN PTR mail.example.test.
    }
}
```

<details>
<summary>drill @localhost mail.example.test TXT</summary>

```
$ drill @localhost mail.example.test TXT
;; ->>HEADER<<- opcode: QUERY, rcode: NOERROR, id: 5792
;; flags: qr aa rd ; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0 
;; QUESTION SECTION:
;; mail.example.test.   IN      TXT

;; ANSWER SECTION:
mail.example.test.      60      IN      TXT     "v=spf1" "mx" "-all"

;; AUTHORITY SECTION:

;; ADDITIONAL SECTION:

;; Query time: 0 msec
;; SERVER: 127.0.0.1
;; WHEN: Tue Sep 22 21:06:32 2020
;; MSG SIZE  rcvd: 79
```
</details>


<details>
<summary>drill @localhost dkim._domainkey.relay.example.test TXT</summary>

```
$ drill @localhost dkim._domainkey.relay.example.test TXT
;; ->>HEADER<<- opcode: QUERY, rcode: NOERROR, id: 59455
;; flags: qr aa rd ; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0 
;; QUESTION SECTION:
;; dkim._domainkey.relay.example.test.  IN      TXT

;; ANSWER SECTION:
dkim._domainkey.relay.example.test.     3600    IN      TXT     "v=DKIM1"

;; AUTHORITY SECTION:

;; ADDITIONAL SECTION:

;; Query time: 0 msec
;; SERVER: 127.0.0.1
;; WHEN: Tue Sep 22 21:09:34 2020
;; MSG SIZE  rcvd: 106
```
</details>

---

Additionally the wildcard syntax `*` is not working for the CNAME nor `*.sub`. If this is difficult to support it needs to also be mentioned that wildcard addressing is unsupported.

miekg:
you sadly need to quote the semi colons, found this by some experimenting, the `;` is a comment in zones files, so it needs to be escaped here to make the parser _not_ parse these as comments (and ignore the rest of the line): `v=DKIM1\;` . etc

polarathene:
> you sadly need to quote the semi colons

The entire value is enclosed between `"` as shown in the example config I provided. The line is very similar to my RFC 1035 zone file from my previous registrar and cloudflare DNS exports. I assume the plugin is dropping the quotes that wrap the string for parsing?

polarathene:
Besides the latest commit adding documentation on escaping `;` in a TXT value (note your example should change the `relay` sub-domain to `mx1` to match the MX record), what about the other TXT record for spf? 

That had no `;` but shows a bug with white-spacing. If you're unable to capture the `"` in the Corefile, are you going to concatenate all values onwards for that line? Wouldn't it be better to address the `"` issue which should also avoid the need to escape `;`? It would be just like in the zone file then. Would `\"` work instead to keep the quotes? I've not tried loading the zone  file itself, but I assume it works without having to escape `;` in my dkim record.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/record..." ]
>    you sadly need to quote the semi colons
>
>The entire value is enclosed between " as shown in the example config I
>provided. The line is very similar to my RFC 1035 zone file from my previous
>registrar and cloudflare DNS exports. I assume the plugin is dropping the
>quotes that wrap the string for parsing?

yes, an artefact of the Corefile parser we're using for this. I don't think I can see if
quotes where there in the first place and re-add them.

/Miek

--
Miek Gieben

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/record..." ]
>Besides the latest commit adding documentation on escaping ; in a TXT value
>(note your example should change the relay sub-domain to mx1 to match the MX
>record), what about the other TXT record for spf?

That should not matter.

`\"` didn't work for me, maybe there is another incantation (never delved this deep in the
parser before)

/Miek

--
Miek Gieben

polarathene:
> I don't think I can see if quotes where there in the first place and re-add them.


> That should not matter.

It shouldn't matter that it returns an inconsistent value? I haven't tried the result with any software that would consume it but `drill` does provide this multiple quoted values for TXT records instead of the single quoted value I get when I query the DNS records of my actual domain.

> `\"` didn't work for me, maybe there is another incantation (never delved this deep in the parser before)

Yes sorry I should perhaps have tried it first. I did this and interestingly it included the exterior quotes with the slash: `"\"v=spf1" "a:mail.example.test" "-all\""`. I also tried single quotes on another TXT record in the config: `"'v=spf1" "mx" "-all'"`, it seems in both cases the quotes were generated after the fact, for each value instead of just returning the quoted value as is in the config. There shouldn't be any need to parse that, perhaps the parser could ignore this value, I'm not familiar with the underlying code.

In rust we can pass a raw string, I wonder if there's anyway to indicate this here? If nothing can be done, then the docs for TXT records should perhaps mention the non-standard return output?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/record..." ]
>    I don't think I can see if quotes where there in the first place and re-add
>    them.
>
>    That should not matter.
>
>It shouldn't matter that it returns an inconsistent value? I haven't tried the
>result with any software that would consume it but drill does provide this
>multiple quoted values for TXT records instead of the single quoted value I get
>when I query the DNS records of my actual domain.

that should be equivalent

>    \" didn't work for me, maybe there is another incantation (never delved
>    this deep in the parser before)
>
>Yes sorry I should perhaps have tried it first. I did this and interestingly it
>included the exterior quotes with the slash: "\"v=spf1" "a:mail.example.test"
>"-all\"". I also tried single quotes on another TXT record in the config: "'v=
>spf1" "mx" "-all'", it seems in both cases the quotes were generated after the
>fact, for each value instead of just returning the quoted value as is in the
>config. There shouldn't be any need to parse that, perhaps the parser could
>ignore this value, I'm not familiar with the underlying code.
>
>In rust we can pass a raw string, I wonder if there's anyway to indicate this
>here? If nothing can be done, then the docs for TXT records should perhaps
>mention the non-standard return output?

If think the Corefile parser is pretty basic, we might be able to add `\"` and/or make
escaping more generic, but this is a minor use-case imo

/Miek

--
Miek Gieben

polarathene:
> If think the Corefile parser is pretty basic, we might be able to add `\"` and/or make escaping more generic, but this is a minor use-case imo

If it helps at all, Caddy's parser handles `"` wrapped values IIRC, I needed to use it when specifying header values for `Cache-Control`. If you could leverage that, then you wouldn't need to have any escaping involved and the entries would be like their RFC 1035 zone file counterparts?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/record..." ]
>    If think the Corefile parser is pretty basic, we might be able to add \"
>    and/or make escaping more generic, but this is a minor use-case imo
>
>If it helps at all, Caddy's parser handles " wrapped values IIRC, I needed to
>use it when specifying header values for Cache-Control. If you could leverage
>that, then you wouldn't need to have any escaping involved and the entries
>would be like their RFC 1035 zone file counterparts?

this is that parser (from caddy v1, mind you). What do you mean with `" wrapped" ? Double
quoting? (just checked that double quoting doesn't work).

polarathene:
> this is that parser (from caddy v1, mind you). 

I'm only familiar with Caddy v2, so perhaps the parser had been improved since. When I was learning Caddyfile config with the response header value, it was doing a similar problem without the quotes IIRC. No clue how header values were handled with v1.

> What do you mean with `" wrapped" ? Double quoting? (just checked that double quoting doesn't work).

I mean exactly like the example Corefile I provided earlier in this PR? The values are effectively the same as the exported zone files (which I referenced for record structure and values), both providers have the TXT records value starting and ending with double quote `"`. The DKIM entry doesn't have the `;` escaped as it's within that quoted string for the TXT value.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/record..." ]
>    What do you mean with `" wrapped" ? Double quoting? (just checked that
>    double quoting doesn't work).
>
>I mean exactly like the example Corefile I provided earlier in this PR? The
>values are effectively the same as the exported zone files (which I referenced
>for record structure and values), both providers have the TXT records value
>starting and ending with double quote ". The DKIM entry doesn't have the ;
>escaped as it's within that quoted string for the TXT value.

No, that's not working with the parser we currently have, you need to escape things in
this specific way to make those TXT records work. (Other records might see similar problems).

If you need to more there is the file plugin

/Miek

--
Miek Gieben

polarathene:
> If you need to more there is the file plugin

All good :+1: 

I guess it's something that might improve if ever in future updating the parser to whatever Caddy v2 is using happens.

Besides those parsing TXT value concerns, do you plan to support the `*` wildcard, or is that something that should be left to the `file` plugin too?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/record..." ]
>    If you need to more there is the file plugin
>
>All good 👍
>
>I guess it's something that might improve if ever in future updating the parser
>to whatever Caddy v2 is using happens.
>
>Besides those parsing TXT value concerns, do you plan to support the *
>wildcard, or is that something that should be left to the file plugin too?

Another good point, that's not supported either. Maybe re-using bits from the file plugin
would be a good iteration OTOH *file* plugin exists :)

/Miek

--
Miek Gieben

polarathene:
> Another good point, that's not supported either. 

Is that likely to always be the case? I am not sure what other gotchas there might be from this plugin approach, it's mostly the fault of the parser being used for the Corefile though right?

Is supporting a different file format like JSON or TOML easier than the Corefile DSL approach? Or would it also require a fair bit of code to support different record types? [OctoDNS uses YAML](https://github.com/github/octodns/blob/master/docs/records.md)(CoreDNS being able to use OctoDNS would equally be pretty cool actually), they've got wildcard support there, would it be easier or more valuable to leverage OctoDNS?

Here's a possible TOML format:

```TOML
origin = 'example.org'
ttl = 300 # sets implicit TTL for records

# Labels below are defined as objects within a 'labels' array, each label has it's own array of DNS records:
[[labels]]
name = '@' # '@' is replaced with origin
records = [
  ['A', '127.0.0.1'],
  ['MX', '10', 'mail'], # domain is not FQDN, appends origin to be 'mail.example.org'
  ['TXT', 'v=spf1 a:mail.example.org -all'],
]

[[labels]]
name = 'mail' # domain is not FQDN, appends origin to be 'mail.example.org'
records = [
  ['A', '192.168.1.42'],
  ['MX', '10', '@'],
  ['TXT', 'v=spf1 mx -all'],
]

[[labels]]
name = 'relay'
records = [
  ['MX', '10', 'mxa.smtp-relay-service.io.'],
  ['MX', '20', 'mxb.smtp-relay-service.io.'],
  ['TXT', 'v=spf1 include:smtp-relay-service.io ~all'],
]

[[labels]]
name = 'dkim._domainkey.relay' # kept simple, no nesting of 'relay' label
ttl = 3600 # set TTL for this labels records, could alternatively support TTL per record in records array?
records = [
  ['TXT', 'v=DKIM1; h=sha256; k=rsa; s=email; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAyPSKcUBKxwMBu594lJpRbUH3aw6jCrFLnHllovU5eFV96qc1bOGAzjYczoDGpWm141720KIL04Ev6vTzIWO2vd7+VVm9CTnCPrrBuhGWuUABWOY36UJfj7gs5BirW/7Tf6C8ksxKE1PnAoDaqn1VpI+GGm+w+N7khEOd9ESNvIMqoiBuzBzOB+r4T5MC8U9rq0m0Y3U5sA+E3s4+uHfaelL4v+Sd9aVcI+ysJKJiKr55Z6XQkBqImMZebXGCyeVKiOyVVAtS/TeDxGv2eu2/K0Cg0AsDYgfW066eRiY5OthPb6tZbuaZbbgryHw5G87R85Jbeuo6cuYDxf3gAIng4QIDAQAB'],
]

[[labels]]
name = 'blog'
records = [
  ['CNAME', '@'],
]

[[labels]]
name = '*' # wildcard
records = [
  ['CNAME', '@'],
]
```

[Single quoted strings are literals](https://learnxinyminutes.com/docs/toml), so no escaping required, there's also syntax support for multi-line strings. Perhaps that's easier to parse? (and easier for any third-parting tooling to generate/modify such as a UI config tool)

The RFC 1035 parser could probably be avoided then, you have an array of labels that can be looked up for keys, followed by their related record types as another array of values where you can inspect record type for handling, most are key-value except say `MX`, or if you were to check record array length for additional values like non-default TTL or CH(Chaos) instead of IN(Internet) record class.

What do you think? I realize the Corefile is preferable, and that the TOML syntax is more verbose. Just an alternative idea since the Corefile approach has demonstrated several unexpected drawbacks.

polarathene:
YAML variant of above config example:

```YAML
origin: example.org
ttl: 300
labels:
  - name: '@'
    records:
      - - A
        - 127.0.0.1
      - - MX
        - '10'
        - mail
      - - TXT
        - v=spf1 a:mail.example.org -all
  - name: mail
    records:
      - - A
        - 192.168.1.42
      - - MX
        - '10'
        - '@'
      - - TXT
        - v=spf1 mx -all
  - name: relay
    records:
      - - MX
        - '10'
        - mxa.smtp-relay-service.io.
      - - MX
        - '20'
        - mxb.smtp-relay-service.io.
      - - TXT
        - v=spf1 include:smtp-relay-service.io ~all
  - name: dkim._domainkey.relay
    ttl: 3600
    records:
      - - TXT
        - v=DKIM1; h=sha256; k=rsa; s=email; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAyPSKcUBKxwMBu594lJpRbUH3aw6jCrFLnHllovU5eFV96qc1bOGAzjYczoDGpWm141720KIL04Ev6vTzIWO2vd7+VVm9CTnCPrrBuhGWuUABWOY36UJfj7gs5BirW/7Tf6C8ksxKE1PnAoDaqn1VpI+GGm+w+N7khEOd9ESNvIMqoiBuzBzOB+r4T5MC8U9rq0m0Y3U5sA+E3s4+uHfaelL4v+Sd9aVcI+ysJKJiKr55Z6XQkBqImMZebXGCyeVKiOyVVAtS/TeDxGv2eu2/K0Cg0AsDYgfW066eRiY5OthPb6tZbuaZbbgryHw5G87R85Jbeuo6cuYDxf3gAIng4QIDAQAB
  - name: blog
    records:
      - - CNAME
        - '@'
  - name: '*'
    records:
      - - CNAME
        - '@'
```

Both formats can be pasted into [this site](https://toolkit.site/format.html) and inspected as a parsed data object interactively.

miekg:
No, we can't just switch parsers midway. You will need to quote certain fields to get
correct output. If need something better than you will want to use the *file* plugin.

Cheers

miekg:
I've moved this to https://github.com/coredns/records

polarathene:
Awesome @miekg , thanks for taking the time to put this together!

anoncam:
> you sadly need to quote the semi colons, found this by some experimenting, the `;` is a comment in zones files, so it needs to be escaped here to make the parser _not_ parse these as comments (and ignore the rest of the line): `v=DKIM1\;` . etc

Is this still the case for DKIM records?  I have found that when trying to add a DKIM key of 2048 bits I get a failure from both `named-checkzone` and coredns does not update my record...

Issue #3033, opened 2019-07-23
plugin/auto Non existent responses return SERVFAIL instead of NXDOMAIN


stp-ip:
This works and returns NXDOMAIN:
```
.:5053 {
  file /zones/example.com example.com
  forward . 8.8.8.8
}
```

This will result in a SERVFAIL:
```
.:5053 {
  auto {
    directory /zones/ (.*) {1}
  }
  forward . 8.8.8.8
}
```

miekg:
>This works and returns NXDOMAIN:
>```
>.:5053 {
>  file /zones/example.com
>  forward . 8.8.8.8
>}
>```

I wouldn't call it working because file is loading example.com with an origin of 
`.`  This looks WAI to me.

>This will result in a SERVFAIL:
>```
>.:5053 {
>  auto {
>    directory /zones/ {.*} {1}
>  }
>  forward . 8.8.8.8
>}
>```

yes, because presumably nothing is detected on disk, so SERVFAIL, also looks 
WAI, to me?

/Miek

--
Miek Gieben

stp-ip:
With a query such as `dig example.net` the first config (using file) will use forward to resolve it or give back a NXDOMAIN, if it can't find it upstream.
The second config (using auto) will SERVFAIL and never jump into forward.
This feels wrong as it should just detect the zone and push anything else to forward.

1:
```
.:5053 {
  file /zones/example.com example.com
  forward . 8.8.8.8
}
```
2:
```
.:5053 {
  auto {
    directory /zones/ (.*) {1}
  }
  forward . 8.8.8.8
}
```

Both only have the single zone file for example.com on disk.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/auto N..." ]
>With a query such as `dig example.net` the first config (using file) will use forward to resolve it or give back a NXDOMAIN, if it can't find it upstream.
>The second config (using auto) will SERVFAIL and never jump into forward.
>This feels wrong as it should just detect the zone and push anything else to forward.

you made file authoritative for example.com, everything else is handled for 
forward.

you made auto authoritative for ".", there is nothing left to forward.

this WAI. Use separate server block that specify for specific origins, don't 
assume "." will suffice

>
>1:
>```
>.:5053 {
>  file /zones/example.com example.com
>  forward . 8.8.8.8
>}
>```
>2:
>```
>.:5053 {
>  auto {
>    directory /zones/ (.*) {1}
>  }
[...]

PR #4134, opened 2020-09-19
plugins/auto: allow fallthrough if no zone match

imacks:
auto plugin currently responds to client with a failure if non of the zonefiles match the query zone. right now the only way for it to fallthrough to the next plugin is by explicitly naming the authoritative zones for this plugin, e.g. `auto example.com example.org`. This is not ideal if there are a lot of zones, or the zonefiles directory changes dynamically. This PR makes the auto plugin pass the query to the next plugin if the query zone doesn't match any of the loaded zonefiles.

this is a solution to #3033

current doc is silent on the issue. 

no.

### DIFF
--- a/plugin/auto/auto.go
+++ b/plugin/auto/auto.go
@@ -50,6 +50,9 @@ func (a Auto) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (i
 
 	// Now the real zone.
 	zone = plugin.Zones(a.Zones.Names()).Matches(qname)
+	if zone == "" {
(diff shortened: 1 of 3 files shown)


miekg:
can you sign the commit? git -s or just manually adding 'Signed-off-by: <your name>', can't merge otherwise

ta

imacks:
done :)

miekg:
argh, you need to rebase, picking up changes in another plugin

imacks:
i created a new pr that should be cleaner. #4166. feel free to close this one.

PR #4109, opened 2020-09-08
bump prometheus with zipkin

zouyee:
Signed-off-by: zounengren <zounengren@cmss.chinamobile.com>

### 1. Why is this pull request needed and what does it do?
bump prometheus with zipkin

### 2. Which issues (if any) are related?
xref https://github.com/coredns/coredns/pull/4071

### 3. Which documentation changes (if any) need to be made?
none

### 4. Does this introduce a backward incompatible change or deprecation?
none

PR #4137, opened 2020-09-21
Revert "bump prometheus with zipkin (#4109)"

yongtang:
This reverts commit 8e1ab71720d09ad1d9b7f5cd55eac4705b4b047e.

See issue #4135 for details.


miekg:
this doesn't removed the offending go.sum import. Needs a 'go mod tidy' ? Or did I point to the wrong commit?

yongtang:
@miekg Actually I don't see the reference of `sourcegraph.com` in coredns after double-check. Tend to believe the issue is on Dependabot side?

miekg:
ok, then we'll check with them. I'll kick off a release then :)

PR #4159, opened 2020-09-29
Add proto and remote tags to trace plugin

miekg:
See #4079 this is that PR with my suggestion committed.


miekg:
ci fails because of the caddyserver -> coredns/caddy move

PR #4161, opened 2020-09-29
fix goroutine leak when writeMsg failed

mimuret:
Fix transfer plugin goroutine leak.
When an error occurs during writeMsg, Transfer can't sent dns.Envelope into channel.

nothing

no need

no

### DIFF
--- a/plugin/transfer/transfer.go
+++ b/plugin/transfer/transfer.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"net"
-	"sync"
 
 	"github.com/coredns/coredns/plugin"
 	clog "github.com/coredns/coredns/plugin/pkg/log"
@@ -50,6 +49,8 @@ type Transferer interface {
 var (
 	// ErrNotAuthoritative is returned by Transfer() when the plugin is not authoritative for the zone.
 	ErrNotAuthoritative = errors.New("not authoritative for zone")
+	// ErrWriteMessage is returned by ServeDNS() when failed to write message.
+	ErrWriteMessage = errors.New("not authoritative for zone")
 )
 
 // ServeDNS implements the plugin.Handler interface.
@@ -107,11 +108,13 @@ func (t *Transfer) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 	// Send response to client
 	ch := make(chan *dns.Envelope)
 	tr := new(dns.Transfer)
-	wg := new(sync.WaitGroup)
-	wg.Add(1)
+	closeCh := make(chan struct{})
 	go func() {
-		tr.Out(w, r, ch)
-		wg.Done()
+		err := tr.Out(w, r, ch)
+		if err != nil {
+			log.Infof("failed to write message zone %q to %s for %d SOA serial: %s", state.QName(), state.IP(), serial, err.Error())
+		}
+		close(closeCh)
 	}()
 
 	rrs := []dns.RR{}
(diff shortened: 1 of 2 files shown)


plugin/transfer/transfer.go
```diff
@@ -107,11 +108,13 @@ func (t *Transfer) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 	// Send response to client
 	ch := make(chan *dns.Envelope)
 	tr := new(dns.Transfer)
-	wg := new(sync.WaitGroup)
-	wg.Add(1)
+	closeCh := make(chan struct{})
 	go func() {
-		tr.Out(w, r, ch)
-		wg.Done()
+		err := tr.Out(w, r, ch)
```
miekg, line 113: we could potentially save this error by using a closure and then return it later when closeCh is closed. Also means we don't need to reinvent a new error

plugin/transfer/transfer.go
```diff
@@ -107,11 +108,13 @@ func (t *Transfer) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
...
+		if err != nil {
+			log.Infof("failed to write message zone %q to %s for %d SOA serial: %s", state.QName(), state.IP(), serial, err.Error())
```
miekg:
```suggestion
			log.Warningf("Failed to write zone transfer of zone %q to %s for %d SOA serial: %s", state.QName(), state.IP(), serial, err)
```

plugin/transfer/transfer.go
```diff
@@ -123,7 +126,11 @@ func (t *Transfer) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 		}
 		rrs = append(rrs, records...)
 		if len(rrs) > 500 {
-			ch <- &dns.Envelope{RR: rrs}
+			select {
+			case ch <- &dns.Envelope{RR: rrs}:
+			case <-closeCh:
+				return dns.RcodeServerFailure, ErrWriteMessage
```
miekg, line 132: this would then returned the saved error from above

plugin/transfer/transfer_test.go
```diff
@@ -276,3 +283,36 @@ func TestTransferNotAllowed(t *testing.T) {
 		t.Errorf("Expected REFUSED response code, got %s", dns.RcodeToString[w.Msg.Rcode])
 	}
 }
+
+type badwriter struct {
```
miekg, line 287: I would rather see a new test which shows we are erroring

miekg:
good catch, bunch of comments, general direction looks good.

plugin/transfer/transfer_test.go
mimuret: Change the test code to see if a message write error was returned.

plugin/transfer/transfer.go
mimuret: https://github.com/coredns/coredns/blob/ea692844b17926ffae7bfdc9066c8abb89d5a1fc/plugin/transfer/transfer.go#L110-L117

change channel type from struct{} to error, and send  tr.Out result to channel.

plugin/transfer/transfer.go
miekg: we should really save this error and return it, no need to make up new error messages

plugin/transfer/transfer_test.go
miekg: please revert changes to this test and add a new one

miekg:
which routine leaks? I believe this is the tests I asked about: https://github.com/coredns/coredns/pull/4283 and I can't find what go routine is not returning there?

mimuret:
Maybe, channel write blocking on transfer.ServDNS.

https://github.com/coredns/coredns/blob/619a0aedc0331bae11d7f261588ab161debce53a/plugin/transfer/transfer.go#L126
https://github.com/coredns/coredns/blob/619a0aedc0331bae11d7f261588ab161debce53a/plugin/transfer/transfer.go#L149

My configure that primary DNS servers are coredns and secondary DNS servers are NSD.
example: https://github.com/mimuret/coredns-leak

NSD's serial check method is used AXFR.
NSD check SOA records in first message,
and Disconnect without reading the rest of the message.

Thereby, an error occurs in the tr.Out, to the reading of the channel will be stopped.
https://github.com/coredns/coredns/blob/619a0aedc0331bae11d7f261588ab161debce53a/plugin/transfer/transfer.go#L113

PR #4162, opened 2020-09-30
doh support: make no TLS config fatal

miekg:
without TLS you can't have a functioning DoH server as no client will be
able to talk to it. Make this a fatal failure.

Add some extra docs on how to start a DoH capable server.


chrisohaver (approved):

miekg:
thank you

PR #4166, opened 2020-10-01
plugins/auto: allow fallthrough if no zone match

imacks:
this is a solution to #3033

auto plugin currently responds to client with a failure if non of the zonefiles match the query zone. right now the only way for it to fallthrough to the next plugin is by explicitly naming the authoritative zones for this plugin, e.g. auto example.com example.org. This is not ideal if there are a lot of zones, or the zonefiles directory changes dynamically. This PR makes the auto plugin pass the query to the next plugin if the query zone doesn't match any of the loaded zonefiles.

this is a solution to #3033

current doc is silent on the issue.

no.

### DIFF
--- a/plugin/auto/auto.go
+++ b/plugin/auto/auto.go
@@ -50,7 +50,10 @@ func (a Auto) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (i
 
 	// Now the real zone.
 	zone = plugin.Zones(a.Zones.Names()).Matches(qname)
-
+	if zone == "" {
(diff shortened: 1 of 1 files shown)


miekg (approved):

miekg:
~~~
--- FAIL: TestTrailingWhitespace (0.02s)

    presubmit_test.go:28: file "/home/travis/gopath/src/github.com/coredns/coredns/plugin/auto/auto.go" has trailing whitespace at line 56, text: "\t"

FAIL
~~~

Issue #4099, opened 2020-08-31
Ready plugin continues to answer OK during lameduck period on existing connections


UnwashedMeme:
**What happened**:

The CoreDNS server continued to report it was "ready" during the lameduck period and only stopped after the server fully shutdown. 

After some investigation (and lots of frustration) I found that the loadbalancer (Azure's) was maintaining an existing TCP connection to CoreDNS and using HTTP pipelining to issue new requests.  CoreDNS closes the listening port right away on entering lameduck so testing with curl looked like it was denying ready requests. 

**What you expected to happen**:

The server to stop reporting it was ready, not just stop listening for new connections.

**How to reproduce it (as minimally and precisely as possible)**:

1. Setup CoreDNS with ready plugin, health plugin, lamduck configured (for suitable long period)
2. `netcat localhost 8181`  to open a connection, leave this open
3. `pkill -SIGTERM coredns`
4. In the netcat session: 
```
GET /ready HTTP/1.1
Host: localhost:8181


```
**Environment**:

- the version of CoreDNS: 1.7.0
- Corefile:
```
. {
    errors
    ready                  # what the LB will probe
    health {
        lameduck 60s
    }
    forward . 168.63.129.16
}
```
- logs, if applicable:
I stopped coredns about half way through here -- did it twice to ensure I could issue repeat queries. Then stopped CoreDNS and did it again.
```
root@coredns100000C:~# netcat localhost 8181
GET /ready HTTP/1.1
Host: localhost:8181

HTTP/1.1 200 OK
Date: Mon, 31 Aug 2020 18:08:05 GMT
Content-Length: 2
Content-Type: text/plain; charset=utf-8

OKGET /ready HTTP/1.1
Host: localhost:8181

HTTP/1.1 200 OK
Date: Mon, 31 Aug 2020 18:08:19 GMT
Content-Length: 2
Content-Type: text/plain; charset=utf-8

OKGET /ready HTTP/1.1
Host: localhost:8181

HTTP/1.1 200 OK
Date: Mon, 31 Aug 2020 18:09:28 GMT
Content-Length: 2
Content-Type: text/plain; charset=utf-8

OK
```

What finally clued  me in was when i stopped asking for listening ports, and show open connections:
```
root@coredns100000C:~# ss -tnp
State   Recv-Q    Send-Q              Local Address:Port                 Peer Address:Port
Process
ESTAB   0         0            [::ffff:172.20.0.18]:8181       [::ffff:168.63.129.16]:57406
[...]

Issue #4160, opened 2020-09-29
localEndpoint is missing in span, in tracer plugin


nemanjan00:
**What happened**:

When sending trace to zipkin, localEndpoint is missing and result of that is the name of service is "unknown-service-name" when sending trace to jaeger. 

**What you expected to happen**:

Jaeger expects localEndpoint that looks something like this:

```
    "localEndpoint": {
      "serviceName": "backend",
      "ipv4": "192.168.99.1",
      "port": 3306
    },
```

It is expected to have `localEndpoint` object with `serviceName` that is the same as service name from config, but, `localEndpoint` is missing. 

**How to reproduce it (as minimally and precisely as possible)**:

Just send trace to jaeger. 

docker-compose.yml

```
version: '3.5'

services:
  dns:
    image: coredns/coredns:latest

    command: ["-conf", "/config/Corefile"]

    ports:
      - "53:53/udp"

    volumes:
      - ./Corefile:/config/Corefile

  jaeger:
    image: 'jaegertracing/all-in-one:1.8'
    environment:
      - COLLECTOR_ZIPKIN_HTTP_PORT=9411
    ports:
      - '8086:8086'
      - '5775:5775/udp'
      - '6831:6831/udp'
      - '6832:6832/udp'
      - '5778:5778'
      - '16686:16686' 
      - '14268:14268'
      - '9411:9411'
```

Corefile

```
. {
  whoami
  health

  trace zipkin jaeger:9411 {
    service dns
  }
}
```

**Anything else we need to know?**:

**Environment**:

- the version of CoreDNS: linux/amd64, go1.15.2, aa82ca6 
- Corefile: 

```
. {
  whoami
  health

  trace zipkin jaeger:9411 {
    service dns
  }
}
```

zouyee:
```
recorder, err := zipkin.NewEndpoint(t.serviceName, t.serviceEndpoint)
	if err != nil {
		log.Warningf("build Zipkin endpoint found err: %v", err)
	}
```
When the configured `serviceEndpoint` does not exist, only the warning log will be displayed.

nemanjan00:
I am not sure how your response is related to my question :D

miekg:
Does this need code changes?

nemanjan00:
Yes, I am not sure if it is not respecting protocol or just using v1, but, it is not sending traces properly. 

Even zipkin is returning error 400 (bad request) for traces from coredns

PR #4167, opened 2020-10-01
plugin/ready: Don't return 200 OK during shutdown

chrisohaver:

Don't reply that we are ready during shutdown.  We already stop listening for new connections when we shutdown, but for clients that re-use connections, ready still answered OK during shutdown (extended by health's lameduck mode). This change makes ready report "not ready" during the shutdown period.

#4099

### DIFF
--- a/plugin/ready/ready.go
+++ b/plugin/ready/ready.go
@@ -43,6 +43,13 @@ func (rd *ready) onStartup() error {
 	rd.Unlock()
 
 	rd.mux.HandleFunc("/ready", func(w http.ResponseWriter, _ *http.Request) {
+		rd.Lock()
+		defer rd.Unlock()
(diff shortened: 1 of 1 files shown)


bradbeam (approved):

mrueg:
@miekg I think we're seeing an issue that will be solved by this on our kubernetes clusters. Is there a possibility to get a 1.7.2 or 1.8.0 release including this change soon?

chrisohaver:
1.7.2 -probably not, since backward incompatible changes have already been merged into master (_the_ release branch).

The 1.8.0 release is currently [waiting](https://github.com/coredns/coredns/issues/4172#issuecomment-703052693) on a fix for #4160 PR in progress #4171.

PR #4169, opened 2020-10-01
plugin/health: fix lameduck docs

### LINKED ISSUE #4084: health plugin `lameduck` option has misleading documentation.
**What happened**:

CoreDNS returns `200 OK` after SIGTERM while in lameduck mode.  

As soon as the SIGTERM was received and lameduck mode was entered  I expected `/health` endpoint would no longer return 200 OK.

Run CoreDNS with
```
    health {
        lameduck 21s   # Specific time isn't really relevant, just make it long enough to observe
    }
```
In a separate terminal run `sudo pkill -SIGTERM coredns; while curl http://localhost:8080/health; do printf "\n\n"; sleep 1; date; done` 

[The docs about lameduck say](https://coredns.io/plugins/health/):

>    Where lameduck will make the process unhealthy then wait for DURATION before the process shuts down.

* It looks like the lameduck changed in https://github.com/coredns/coredns/commit/c778b3a67ca43124a539acda21eca4934c8110f6#diff-8244c8b222eeb3813330b611798eb81bL78
  * L78 here is inside of `OnFinalShutdown`
  * It looks like that patch was addressing a slightly different situation but along the way it became "health always reports healthy"
* The ready plugin works exactly as expected in this situation:  `localhost:8181/ready` fails immediately upon entering lameduck mode.

I think the thing to do here is just fix up the docs, something like:

>  Where lameduck will delay shutdown for DURATION. `/health` will still report 200 OK and DNS requests are served but the [ready](https://coredns.io/plugins/ready/) plugin will start failing and can be used to observe that CoreDNS is in lameduck mode prior to shutdown.

**Environment**:

- the version of CoreDNS: 1.7.0
- Corefile:
```
. {
    errors
    ready
    health {
        lameduck 21s
    }
    prometheus :9153
    forward . 168.63.129.16
}
```
```
NAME="Ubuntu"
VERSION="20.04.1 LTS (Focal Fossa)"
```
- CoreDNS Logs:

```
Aug 27 20:45:40 coredns1000002 coredns[3415]: [INFO] SIGTERM: Shutting down servers then terminating
Aug 27 20:45:40 coredns1000002 coredns[3415]: [INFO] plugin/health: Going into lameduck mode for 21s
```
- Separate window running curl http://localhost:8080/health in a loop:
```
Thu Aug 27 20:45:39 UTC 2020
OK
Thu Aug 27 20:45:40 UTC 2020
OK
Thu Aug 27 20:45:41 UTC 2020
OK
Thu Aug 27 20:45:42 UTC 2020
OK
...
Thu Aug 27 20:46:01 UTC 2020
OK
Thu Aug 27 20:46:02 UTC 2020
curl: (7) Failed to connect to localhost port 8080: Connection refused
```

chrisohaver:

Fixes lameduck docs.  Also mention _ready_ plugin behavior during lameduck.

fixes #4084

### DIFF
--- a/plugin/health/README.md
+++ b/plugin/health/README.md
@@ -26,8 +26,8 @@ health [ADDRESS] {
 }
 ~~~
 
-* Where `lameduck` will make the process unhealthy then *wait* for **DURATION** before the process
-  shuts down.
+* Where `lameduck` will delay shutdown for **DURATION**. /health will still answer 200 OK.
+  Note: The *ready* plugin will not answer OK while CoreDNS is in lameduck mode prior to shutdown.
 
 If you have multiple Server Blocks, *health* can only be enabled in one of them (as it is process
 wide). If you really need multiple endpoints, you must run health endpoints on different ports:


yongtang (approved):
lgtm

PR #4171, opened 2020-10-01
plugin/trace: Have a consistent spanName

nyodas:
The current trace implementation create too many top level trace.
image

The spans are very varied and so can't be compared against each other.
But also bog down the trace store behind it.

The book [Distributed Tracing in Practice by Austin Parker, Daniel Spoonhower, Jonathan Mace, Ben Sigelman, Rebecca Isaacs](https://www.oreilly.com/library/view/distributed-tracing-in/9781492056621/ch04.html)

Mention that span names do not have to be the name of the function nor be unique

> First, names should be aggregable. In short, you want to avoid span names that are unique to each execution of a service. One antipattern we see, especially in HTTP services, is implementers making the span name the same as the fully matched route (such as GET /api/v2/users/1532492). This pattern makes it difficult to aggregate operations across thousands or millions of executions, severely reducing the utility of your data. Instead, make the route more generic and move parameters to tags, such as GET /api/v2/users/{id} with an associated tag of userId: 1532492.

This allows us to have APM based Graph such as

image

image

image

image

image

I moved the sampling to the implementation of each trace provider
To let them decide how to best decide when to drop the traces.
It allows the datadog provider to smoothly provide the right effective throughput ( in the UI )

none
Depends on the answer to the question in 5
Possible removal of `client_server` 
Possible addition of a datadog only option regarding trace analyze 
Possible addition of a `TopLevelSpanName` only option to remain backward compatible

Depends on the answer to the question in 5
Possible backward incompatible change regarding the `TopLevelSpanName`

### 5. Additional Question.

I have a few additional question.

Apparently the `clientServer` options was zipkin only and has been removed from the init of the zipkin tracer in #4109 
The configuration and documentation still point to it. Should i drop that completely or should i reinstate it ?

The change in traceName create a backward incompatible (UI wise) change.
I'm not that familiar with zipkin to be certain if that's an issue or not.
Should I create a configuration to toggle that on or off ?

Related to the first question, i didn't add it yet in this PR but to get the nice *table* and *graph* showed here, datadog need an additional tags on the trace it sends to analyze the trace further.
The trace getting analyzed have an additional cost for datadog customer which is why i didn't set that up by default.

Is it alright to create a Datadog only option in the plugin configuration like `client_server` used to be ?

### DIFF
--- a/plugin/trace/setup.go
+++ b/plugin/trace/setup.go
@@ -83,6 +83,21 @@ func traceParse(c *caddy.Controller) (*trace, error) {
 				if err != nil {
 					return nil, err
 				}
+			case "datadog_analytics_rate":
+				args := c.RemainingArgs()
+				if len(args) > 1 {
+					return nil, c.ArgErr()
+				}
+				tr.datadogAnalyticsRate = 0
+				if len(args) == 1 {
+					tr.datadogAnalyticsRate,err = strconv.ParseFloat(args[0], 64)
+				}
+				if err != nil {
+					return nil, err
+				}
+				if tr.datadogAnalyticsRate > 1 || tr.datadogAnalyticsRate < 0 {
+					return nil,fmt.Errorf("datadog analytics rate must be between 0 and 1, '%f' is not supported", tr.datadogAnalyticsRate )
+				}
 			}
 		}
 	}
--- a/plugin/trace/trace.go
+++ b/plugin/trace/trace.go
@@ -8,7 +8,6 @@ import (
 	"sync/atomic"
 
 	"github.com/coredns/coredns/plugin"
-	"github.com/coredns/coredns/plugin/metrics"
 	"github.com/coredns/coredns/plugin/pkg/dnstest"
 	"github.com/coredns/coredns/plugin/pkg/log"
 	"github.com/coredns/coredns/plugin/pkg/rcode"
@@ -20,30 +19,33 @@ import (
 	zipkinot "github.com/openzipkin-contrib/zipkin-go-opentracing"
 	"github.com/openzipkin/zipkin-go"
 	zipkinhttp "github.com/openzipkin/zipkin-go/reporter/http"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/opentracer"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 )
 
 const (
-	tagName   = "coredns.io/name"
-	tagType   = "coredns.io/type"
-	tagRcode  = "coredns.io/rcode"
-	tagProto  = "coredns.io/proto"
-	tagRemote = "coredns.io/remote"
+	tagName                 = "coredns.io/name"
+	tagType                 = "coredns.io/type"
+	tagRcode                = "coredns.io/rcode"
+	tagProto                = "coredns.io/proto"
+	tagRemote               = "coredns.io/remote"
+	defaultTopLevelSpanName = "servedns"
 )
 
 type trace struct {
 	count uint64 // as per Go spec, needs to be first element in a struct
 
-	Next            plugin.Handler
-	Endpoint        string
-	EndpointType    string
-	tracer          ot.Tracer
-	serviceEndpoint string
-	serviceName     string
-	clientServer    bool
-	every           uint64
-	Once            sync.Once
+	Next                 plugin.Handler
+	Endpoint             string
+	EndpointType         string
+	tracer               ot.Tracer
+	serviceEndpoint      string
+	serviceName          string
+	clientServer         bool
+	every                uint64
+	datadogAnalyticsRate float64
+	Once                 sync.Once
 }
 
 func (t *trace) Tracer() ot.Tracer {
@@ -58,7 +60,13 @@ func (t *trace) OnStartup() error {
 		case "zipkin":
 			err = t.setupZipkin()
 		case "datadog":
-			tracer := opentracer.New(tracer.WithAgentAddr(t.Endpoint), tracer.WithServiceName(t.serviceName), tracer.WithDebugMode(log.D.Value()))
+			tracer := opentracer.New(
+				tracer.WithAgentAddr(t.Endpoint),
+				tracer.WithDebugMode(log.D.Value()),
+				tracer.WithGlobalTag(ext.SpanTypeDNS, true),
+				tracer.WithServiceName(t.serviceName),
+				tracer.WithAnalyticsRate(t.datadogAnalyticsRate),
+			)
 			t.tracer = tracer
 		default:
 			err = fmt.Errorf("unknown endpoint type: %s", t.EndpointType)
@@ -73,7 +81,10 @@ func (t *trace) setupZipkin() error {
 	if err != nil {
 		log.Warningf("build Zipkin endpoint found err: %v", err)
 	}
-	tracer, err := zipkin.NewTracer(reporter, zipkin.WithLocalEndpoint(recorder))
+	tracer, err := zipkin.NewTracer(
+		reporter,
+		zipkin.WithLocalEndpoint(recorder),
+	)
 	if err != nil {
 		return err
 	}
@@ -100,7 +111,7 @@ func (t *trace) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 	}
 
 	req := request.Request{W: w, Req: r}
-	span = t.Tracer().StartSpan(spanName(ctx, req))
+	span = t.Tracer().StartSpan(defaultTopLevelSpanName)
 	defer span.Finish()
 
 	rw := dnstest.NewRecorder(w)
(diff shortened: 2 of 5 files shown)


plugin/trace/trace.go
```diff
@@ -112,10 +116,9 @@ func (t *trace) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 	span.SetTag(tagProto, req.Proto())
 	span.SetTag(tagRemote, req.IP())
 	span.SetTag(tagRcode, rcode.ToString(rw.Rcode))
+	if t.EndpointType == "datadog" {
```
johnbelamaric: Maybe put this type-specific logic up in the `OnStartup` by adding a function pointer to the `trace` struct and calling that here?

johnbelamaric (changes requested):
Thanks! Does this have any backwards-incompatibility issues? It sounds like span names might change? Or was it just broken the way it worked before?

johnbelamaric:
Sorry, I see your questions now.

1) I don't know if anyone is using `client_server`, but at the very least we cannot kill the config option and have any existing configs fail to load. If this feature is still available in the latest OT/Zipkin libs then please do re-instate it (separate PR is OK).

2) Can you propose a configuration syntax for the datadog specific options? I would think any type-specific options should start with the type name, ideally it would all be done in one line.

3) The intent was always a single span for a given request, from your discription it sounds like we were getting multiple top-level spans for a single request? I would consider that a bug and therefore no backwards compatibility issue exists with respect to span names.

nyodas:
> 1. I don't know if anyone is using `client_server`, but at the very least we cannot kill the config option and have any existing configs fail to load. If this feature is still available in the latest OT/Zipkin libs then please do re-instate it (separate PR is OK).

Cool will revert the change an verify it's still in the latest version.

> 2. Can you propose a configuration syntax for the datadog specific options? I would think any type-specific options should start with the type name, ideally it would all be done in one line.

Would something like `datadog_analyzed_rate = 1` or `datadog.analyzed_rate = 0.001` work
I will POC something and get back to you.

> 3. The intent was always a single span for a given request, from your description it sounds like we were getting multiple top-level spans for a single request? I would consider that a bug and therefore no backwards compatibility issue exists with respect to span names.

I think this is a bug, but it will change the established behavior.

Which is something like so.
Every request will have a different top level name. like in the first screenshot.
So for request A you would get 
`servedns_dns_53_www.coredns.io`
And for request B you would get 
`servedns_dns_53_www.github.com`

Which is in my opinion *broken*
At least in Datadog this break the usability.

However for request A  the `TopLevelSpanName` will always be the same.
The issue is that the top span should always be the same for every request ( the tags associated to the trace provide the context ) 
Currently for every query having a different request you will get another `TopLevelSpanName`.

plugin/trace/trace.go
nyodas: Moved that up there

plugin/trace/setup.go
```diff
@@ -100,7 +115,7 @@ func normalizeEndpoint(epType, ep string) (string, string, error) {
 
 	if epType == "zipkin" {
 		if !strings.Contains(ep, "http") {
-			ep = "http://" + ep + "/api/v1/spans"
+			ep = "http://" + ep + "/api/v2/spans"
```
nyodas: Trace weren't being accepted by my local zipkin server with 
```
Expected a JSON_V1 encoded list, but received: JSON_V2
```
Due to the update of the vendor in #4109

plugin/trace/trace.go
```diff
@@ -73,7 +84,12 @@ func (t *trace) setupZipkin() error {
 	if err != nil {
 		log.Warningf("build Zipkin endpoint found err: %v", err)
 	}
-	tracer, err := zipkin.NewTracer(reporter, zipkin.WithLocalEndpoint(recorder))
+	tracer, err := zipkin.NewTracer(
+		reporter,
+		zipkin.WithLocalEndpoint(recorder),
+		zipkin.WithSharedSpans(t.clientServer),
```
nyodas: Reinstated the `client_server` usage

plugin/trace/setup.go
```diff
@@ -83,6 +83,21 @@ func traceParse(c *caddy.Controller) (*trace, error) {
 				if err != nil {
 					return nil, err
 				}
+			case "datadog_analytics_rate":
```
nyodas, line 86: I went with this
I will write the doc and test once we're in agreement over this :)

nyodas:
> I don't know if anyone is using client_server, but at the very least we cannot kill the config option and have any existing configs fail to load. If this feature is still available in the latest OT/Zipkin libs then please do re-instate it (separate PR is OK).

Brought it back 
If it starts to bloat this PR i will open another one

While testing that change i noticed that the zipkin provider was borked and couldn't push traces anymore
Tcpdump showed that the zipkin server was responding with

 ```
    Expected a JSON_V1 encoded list, but received: JSON_V2
 ```

There is still an issue with
```
recorder, err := zipkin.NewEndpoint(t.serviceName, t.serviceEndpoint)
```
As `t.serviceEndpoint` is `.:53` in most cases
Zipkin can't parse the `.` and send a warning with
```
[WARNING] build Zipkin endpoint found err: lookup : no such host
```
cf: [openzipkin/endpoint.go](https://github.com/openzipkin/zipkin-go/blob/master/endpoint.go#L27-L80)

plugin/trace/setup.go
johnbelamaric: So this probably fixes #4160 bug?

johnbelamaric: So this means we no longer support v1, since the #4109 ? That needs to be documented.

plugin/trace/trace.go
```diff
@@ -58,7 +59,17 @@ func (t *trace) OnStartup() error {
...
 			err = t.setupZipkin()
 		case "datadog":
-			tracer := opentracer.New(tracer.WithAgentAddr(t.Endpoint), tracer.WithServiceName(t.serviceName), tracer.WithDebugMode(log.D.Value()))
+			samplingRules := []tracer.SamplingRule{
+				tracer.RateRule(1 / float64(t.every)),
+			}
+			tracer := opentracer.New(
+				tracer.WithAgentAddr(t.Endpoint),
+				tracer.WithDebugMode(log.D.Value()),
+				tracer.WithGlobalTag(ext.SpanTypeDNS, true),
```
johnbelamaric, line 66: nice, that's better

johnbelamaric (approved):
Those traces with the name in them look totally wrong, I don't think that is how this always worked. 99% sure someone introduced it later. So definitely a bug. It should be documented but we do not need to keep that backward compatible.

analytics option looks fine. if we end up with wanting lots of options we may need a different way, but good for now.

johnbelamaric:
Please answer the questions above, also open a separate PR for the release note about v1 support, then I can merge.

plugin/trace/setup.go
nyodas: > So this probably fixes #4160 bug?
This fix the 400 that is mentioned in the response

There is still the issue of the warning in my comment bellow ( i think i need to do a replace of `.` to `0.0.0.0` to have it work as expected

> So this means we no longer support v1, since the #4109 ? That needs to be documented.

Yes v1 is not supported since #4109

zouyee:
According to #4160 bug, the following tasks need to be completed
1. we no longer support v1 needs to be documented(update related readme), it is my mistake.
2. update 1.8.0 notes
4. according to #4160 bug, do we need to add the corresponding interface data test ?
thanks!

nyodas:
Created #4180 to fix the #4160 issue
We can continue to discuss over there for that particular issue.

plugin/trace/setup.go
nyodas: ✅  Wrote the doc and added a few test case.

plugin/trace/trace.go
```diff
@@ -86,21 +101,13 @@ func (t *trace) Name() string { return "trace" }
...
-	}
 	span := ot.SpanFromContext(ctx)
-	if !trace || span != nil {
+	if span != nil {
 		return plugin.NextOrFailure(t.Name(), t.Next, ctx, w, r)
 	}
 
 	req := request.Request{W: w, Req: r}
-	span = t.Tracer().StartSpan(spanName(ctx, req))
+	span = t.Tracer().StartSpan(defaultTopLevelSpanName)
```
johnbelamaric, line 114: So, I just realized a potential problem. I am wondering about the performance impact here. Previously, because the `every` was implemented outside, we never allocated a span unless we were hitting the sample. Now, we allocate a span for every single query. What is the cost of allocating a span, in memory and time? Can you check some benchmarks before and after?

nyodas: Will check this and get back to you with graphs.

nyodas: Doesn't look good there is a 20% hit
Memory
image

CPU
image

Corresponding req/s
image

Will leave that up a little to see if it change

Will touch base with the integration team at datadog if that's expected.

nyodas: After letting it sit for the night, things are definitely not looking good
40% increase in memory usage with a every at 100 using the datadog provider.
Currently touching base to see if the lib has a way to decrease that significantly else i will rollback that change and go back to the purest expression of the sampling we had before 

🟠  Baseline (yesterday values)
🔵  Current values
image

image

nyodas:
Following the increase in resource consumption due to the sampling happening on the apm provider (up to 40% in my tests).
I removed the use of the sampling by the lib themselves to do it early.

johnbelamaric:
Thank you!

Issue #4165, opened 2020-10-01
Add query_timestamp to RESPONSE events


SuperFunks:
**What would you like to be added**:
Add query_timestamp to RESPONSE events

**Why is this needed**:
In order to diff query_timestamp and response_timestamp to calculate the responsetime/resolutiontime performance.
I've started the discussion about this on the golang-dnstap github repository: https://github.com/dnstap/golang-dnstap/issues/36, but it appears to be a missing population in the RESPONSE events created in the CoreDNS DNSTAP Plugin.

"The plugin code appears to keep the query timestamp around as the QueryEpoch field of the dnstap plugin ResponseWriter, but does not use it when generating response reports."

miekg:
I think https://github.com/coredns/coredns/pull/4179 fixes this. Could you double check?

SuperFunks:
Yes for sure! - ill check this out first thing tomorrow.

SuperFunks:
I have tested it out - and something is good - but other stuff is bad.

I did a readout from DIG against CoreDNS using the dnstap2 branch - and this was the results:

**From the JSON output:**
"query_time":"2020-10-06T12:34:33.091035862Z",
"response_time":"2020-10-06T12:34:33.331921193Z",

**Nanoseconds difference between timestamps:** 
240885331 = 240.88 milliseconds

**DiG tool reported this delay:**
;; Query time: 241 msec

I'm guessing DIG rounds up - and DNSTAP has the more precise timing - so this is great!

One bad thing i saw though was that all events returned from DNSTAP is now of type CLIENT_RESPONSE or FORWARDER_RESPONSE. What happened to CLIENT_QUERY or FORWARDER_QUERY?

It looks to me like you have merged all query and response type events - effectively cutting the number of events in half ( i like that :) ) - but increasing the size for all events ( i don't like that :P ).

Also i have found that if CoreDNS experiences a timeout to an upstream DNS provider in the Forward plugin - no query event is shipped by DNSTAP - but in this case we do get a FORWARDER_QUERY event but that event doesn't contain the query_address of the real client. This is pretty bad in terms of insights as we have no way of knowing who called us during an upstream outage.

The same thing happens with i use the Unbound plugin - no client_query event is generated - even if DNS Recursion somehow goes wrong.

Merging Query and Response events when both is present is a good idea - but losing the query events when something goes wrong is not good :(

SuperFunks:
[...]

PR #4179, opened 2020-10-05
plugin/dnstap: various cleanups

miekg:
Remove identical functions ToClientQuery etc, and just use ToTapMessage
with the type. Remove the Builder type and just make it Msg and make
error handling explicit instead of using a Err field that gets checked
only at the end.

A recent issue made me look into this plugin, I suspect various other
cleanups (hopefully deletion of code) can be made as well


miekg:
ok, this needs more work, msg/msg.go just wraps the tap struct which seems to add an indirection that creates confusion. I'm going to see where this ends and how big of a code reduction I can create (if any)

plugin/dnstap/context.go
```diff
@@ -6,9 +6,9 @@ type contextKey struct{}
 
 var dnstapKey = contextKey{}
 
-// ContextWithTapper returns a new `context.Context` that holds a reference to
+// contextWithTapper returns a new `context.Context` that holds a reference to
 // `t`'s Tapper.
-func ContextWithTapper(ctx context.Context, t Tapper) context.Context {
+func contextWithTapper(ctx context.Context, t Tapper) context.Context {
```
rdrozhdzh: Actually I would rather remove this logic as well. 
Any plugin can get a reference to another plugin in the same server block. This could be done in a plugin startup procedure once with a code like
```go
if taph := dnsserver.GetConfig(c).Handler("dnstap"); taph != nil {
	tapPlugin, ok := taph.(dnstap.Dnstap)
}
```
This way we can avoid adding (needs memory allocation) and retrieving the same data to/from the context for each query.

plugin/dnstap/handler.go
```diff
@@ -32,14 +31,6 @@ type (
 	}
 )
 
-// ContextKey defines the type of key that is used to save data into the context.
-type ContextKey string
-
-const (
-	// DnstapSendOption specifies the Dnstap message to be send.  Default is sent all.
-	DnstapSendOption ContextKey = "dnstap-send-option"
```
rdrozhdzh, line 40: Removing DnstapSendOption is backward incompatible. With this option other plugins can control which of CQ or CR messages (or both) should be generated.

To be honest, I would prefer to have this config options in dnstap plugin configuration rather than sending them via context with each query. Also, it may be worth to configure generation of FQ, FR messages as well

plugin/dnstap/msg/msg.go
```diff
@@ -1,159 +1,78 @@
...
+// SetQueryAddress adds the query address to the message. This also sets the SocketFamily and SocketProtocol.
+func SetQueryAddress(t *tap.Message, addr net.Addr) error {
+	inet := tap.SocketFamily_INET
+	t.SocketFamily = &inet
+	switch a := addr.(type) {
 	case *net.TCPAddr:
-		b.Address = addr.IP
-		b.Port = uint32(addr.Port)
-		b.SocketProto = tap.SocketProtocol_TCP
+		t.QueryAddress = a.IP
```
rdrozhdzh, line 24: consider code de-duplication and optimization like below
```go
// these could be defined once at a packet level and shared between all messages since they are read only
var (
  protoUDP = tap.SocketProtocol_UDP
  protoTCP = tap.SocketProtocol_TCP
  familyINET = tap.SocketFamily_INET
  familyINET6 = tap.SocketFamily_INET6
)

func SetQueryAddress(t *tap.Message, addr net.Addr) error {
  var (
    ip net.IP
    port uint32
  )

  t.SocketFamily = &familyINET

  switch a := addr.(type) {
  case *net.TCPAddr:
    ip, port = a.IP, uint32(a.Port)
    t.SocketProtocol = &protoTCP
  case *net.UDPAddrr:
    ip, port = a.IP, uint32(a.Port)
    t.SocketProtocol = &protoUDP
  default:
    return fmt.Errorf("unknown address type: %T", a)
  }

  t.QueryAddress = ip
  t.QueryPort = &port
  if ip.To4() == nil {
    t.SocketFamily = &familyINET6
  }
}
```

plugin/dnstap/msg/msg.go
```diff
@@ -1,159 +1,78 @@
...
-		}
-		return b
+// SetResponseAddress the response address to the message.
+func SetResponseAddress(t *tap.Message, addr net.Addr) error {
+	switch a := addr.(type) {
+	case *net.TCPAddr:
+		t.ResponseAddress = a.IP
+		p := uint32(a.Port)
+		t.ResponsePort = &p
+		return nil
```
rdrozhdzh, line 63: I believe the proto and family should be set as well, since `SetQueryAddress()` and `SetResponseAddress()` can be called independently.

plugin/dnstap/writer.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if w.Pack() {
+		buf, err := resp.Pack()
+		if err != nil {
+			w.Err = err
+			return err
+		}
+		b.ResponseMessage = buf
+	}
+	msg.SetType(b, tap.Message_CLIENT_RESPONSE)
```
rdrozhdzh: `b.QueryMessage` should be erased if you reuse the same message structure. The same in forward plugin

plugin/dnstap/writer.go
```diff
@@ -0,0 +1,62 @@
...
+
+	Err error
+}
+
+// WriteMsg writes back the response to the client and THEN works on logging the request
+// and response to dnstap.
+func (w *ResponseWriter) WriteMsg(resp *dns.Msg) error {
+	writeErr := w.ResponseWriter.WriteMsg(resp)
+
+	b := new(tap.Message)
```
rdrozhdzh: Previously,  var `b`  was of type `Builder`, so it was called `b`. Now it is of type *tap.Message and the old name is confusing. The same in forward plugin.

rdrozhdzh (commented):
I like your idea to simplify this plugin. However, your changes make new version of the plugin incompatible with old version. There were a lot of public properties and methods in dnstap package. I think we should not simply remove them without deprecation for some reasonable period of time.

rdrozhdzh:
@miekg , I think we can just fix the issue #4165 now and rework dnstap plugin later. What do you think?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/dnstap..." ]
>@miekg , I think we can just fix the issue #4165 now and rework dnstap plugin
>later. What do you think?

there is no backward compat quarantee for this *and* we releasing 1.8.0 anyway, so I can't
see the point. Fixing code to be thrown away a day later seems a non-optimal time spend

rdrozhdzh:
So, are you going to merge your version? And release it in 1.8.0? 
Please, add notes about backward incompatibility then. And I will close my PR, if so.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/dnstap..." ]
>So, are you going to merge your version? And release it in 1.8.0?

well, once it's in a good shape, yes

>Please, add notes about backward incompatibility then. And I will close my PR,
>if so.

yep this will get a note in the relase notes. How to use is updated in the README of
dnstap.

/Miek

--
Miek Gieben

plugin/dnstap/context.go
miekg: good point

plugin/dnstap/handler.go
miekg: the plugin only has one option `full`, for the rest this was set to true. So I that flexibility is needed (it never was) config options should be added. 
Per previous comment the whole context stuff is gone, so this can be this be a bunch of booleans in `Dnstap`

plugin/dnstap/msg/msg.go
miekg: done. I don't have `a` outside of the type switch, so extracting out the `a.IP.To4()` becomes ikcy.

plugin/dnstap/msg/msg.go
miekg: done

plugin/dnstap/writer.go
miekg: use `tm` for tap message through out

miekg: correct comments to the wrong thread - but this is done as well

miekg:
comments addressed. PTAL

plugin/dnstap/handler.go
rdrozhdzh: Yes, these options are really needed. AFAIK bind and unbound allow configuring per message type as well.
The config may look like
```
  dnstap tcp://127.0.0.1:9123 full CR FR
```
where all parameters after endpoint are optional
or
```
  dnstap tcp://127.0.0.1:9123 full {
    allow CR FR
  }
```
Other plugins should be able to check these config options. Bunch of booleans is ok.

plugin/dnstap/writer.go
```diff
@@ -0,0 +1,63 @@
...
+	if w.IncludeRawMessage {
+		buf, err := w.Query.Pack()
+		if err != nil {
+			w.Err = err
+			return err
+		}
+		tm.QueryMessage = buf
+	}
+	msg.SetType(tm, tap.Message_CLIENT_QUERY)
+	w.TapMessage(tm)
```
rdrozhdzh: ideally, CQ message should be loged from `func (h Dnstap) ServeDNS()`. So that in case if some plugin chooses to drop DNS response (i.e. doesn't write DNS response), the CQ still to be sent. Though, this was never implemented

plugin/forward/dnstap.go
```diff
@@ -1,61 +1,60 @@
...
+	port, _ := strconv.ParseUint(p, 10, 32) // same here
+
 	opts := f.opts
-	t := ""
+	t := state.Proto()
 	switch {
 	case opts.forceTCP: // TCP flag has precedence over UDP flag
 		t = "tcp"
 	case opts.preferUDP:
 		t = "udp"
```
rdrozhdzh, line 30: this is not quite reliable. Forward may try udp and finally use TCP. The `opts` from the line https://github.com/coredns/coredns/blob/02f52b5d271d47a035c5432cca4fcf4174c023fa/plugin/forward/forward.go#L126 has more accurate value, since it is updated after UDP failure

rdrozhdzh (commented):
Overall looks good to me. The config options for message types are highly desired

miekg:
Thanks for reviewing.

Adding those options lgtm and should be done in follow up PR

plugin/forward/dnstap.go
miekg: and done.

plugin/forward/dnstap.go
```diff
@@ -1,61 +1,58 @@
...
 	tap "github.com/dnstap/golang-dnstap"
 	"github.com/miekg/dns"
 )
 
-func toDnstap(ctx context.Context, host string, f *Forward, state request.Request, reply *dns.Msg, start time.Time) error {
-	tapper := dnstap.TapperFromContext(ctx)
-	if tapper == nil {
-		return nil
-	}
+// oDnstap will send the forward and received message to the dnstap plugin.
```
rdrozhdzh:
```suggestion
// toDnstap will send the forward and received message to the dnstap plugin.
```

plugin/forward/dnstap.go
```diff
@@ -30,30 +30,31 @@ func toDnstap(f *Forward, host string, state request.Request, opts options, repl
 
 	if t == "tcp" {
 		ta := &net.TCPAddr{IP: net.ParseIP(ip), Port: int(port)}
-		msg.SetQueryAddress(tm, ta)
+		msg.SetQueryAddress(q, ta)
 	} else {
 		ta := &net.UDPAddr{IP: net.ParseIP(ip), Port: int(port)}
-		msg.SetQueryAddress(tm, ta)
+		msg.SetQueryAddress(q, ta)
```
rdrozhdzh: `msg.SetQueryAddress()` should be called for response as well. Though, not sure which of `msg.SetResponseAddress()` or `msg.SetQueryAddress()` is better for FQ, FR messages, I can't see such details in https://github.com/dnstap/golang-dnstap/blob/master/dnstap.pb/dnstap.proto comments. Perhaps, we could add both

plugin/forward/setup.go
```diff
@@ -41,6 +35,14 @@ func setup(c *caddy.Controller) error {
 	c.OnStartup(func() error {
 		return f.OnStartup()
 	})
+	c.OnStartup(func() error {
```
rdrozhdzh, line 38: lets also mention `c.OnStartup()` in README example

plugin/forward/dnstap.go
```diff
@@ -48,9 +47,8 @@ func toDnstap(f *Forward, host string, state request.Request, opts options, repl
 	r := new(tap.Message)
```
rdrozhdzh: lets allocate the `r` object inside the `if` section below

miekg: done

plugin/forward/dnstap.go
miekg: and done. Also unsure, use QueryAddress()

plugin/forward/setup.go
miekg: good point. done

rdrozhdzh (approved):
Thanks, LGTM

miekg:
thanks for the thorough review @rdrozhdzh

PR #4180, opened 2020-10-05
plugin/trace: Fix zipkin json_v2 and #4160

nyodas:
In #4171 i noticed that the client_server was not being used anymore.
It appears that it should still be there.

While testing that change i noticed that the zipkin provider was borked and couldn't push traces anymore
Tcpdump showed that the zipkin server was responding with

```
Expected a JSON_V1 encoded list, but received: JSON_V2
```

There was also an issue (#4160)  with
```

	recorder, err := zipkin.NewEndpoint(t.serviceName, t.serviceEndpoint)
```
As `t.serviceEndpoint` is `:53` in most cases
Zipkin can't parse the ip:port tuple and send a warning with
```
[WARNING] build Zipkin endpoint found err: lookup : no such host
```
cf: [openzipkin/endpoint.go](https://github.com/openzipkin/zipkin-go/blob/master/endpoint.go#L27-L80)

To actively set a correct values in serviceEndpoint you need to use the bind plugin
```
.:53 {
    bind 127.0.0.1
    trace
}
```

#4160 
#4109 
#4171 

Indicate that the zipkin api v1 is not working as of `1.7.1`

Yes the api version of zipkin is now using `v2`
Though that fix a bug introduced with #4109

### DIFF
--- a/plugin/trace/README.md
+++ b/plugin/trace/README.md
@@ -56,6 +56,8 @@ You can run Zipkin on a Docker host like this:
 docker run -d -p 9411:9411 openzipkin/zipkin
 ```
 
+:warning: The zipkin provider does not support the v1 API since coredns 1.7.1
+
 ## Examples
 
 Use an alternative Zipkin address:
--- a/plugin/trace/setup.go
+++ b/plugin/trace/setup.go
@@ -35,7 +35,9 @@ func traceParse(c *caddy.Controller) (*trace, error) {
 	)
 
 	cfg := dnsserver.GetConfig(c)
-	tr.serviceEndpoint = cfg.ListenHosts[0] + ":" + cfg.Port
+	if cfg.ListenHosts[0] != "" {
+		tr.serviceEndpoint = cfg.ListenHosts[0] + ":" + cfg.Port
+	}
 
 	for c.Next() { // trace
 		var err error
@@ -115,7 +117,7 @@ func normalizeEndpoint(epType, ep string) (string, string, error) {
 
 	if epType == "zipkin" {
 		if !strings.Contains(ep, "http") {
-			ep = "http://" + ep + "/api/v1/spans"
+			ep = "http://" + ep + "/api/v2/spans"
 		}
 	}
 
--- a/plugin/trace/setup_test.go
+++ b/plugin/trace/setup_test.go
@@ -16,17 +16,17 @@ func TestTraceParse(t *testing.T) {
 		clientServer bool
 	}{
 		// oks
-		{`trace`, false, "http://localhost:9411/api/v1/spans", 1, `coredns`, false},
-		{`trace localhost:1234`, false, "http://localhost:1234/api/v1/spans", 1, `coredns`, false},
+		{`trace`, false, "http://localhost:9411/api/v2/spans", 1, `coredns`, false},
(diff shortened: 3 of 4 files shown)


plugin/trace/README.md
```diff
@@ -47,6 +47,8 @@ You can run Zipkin on a Docker host like this:
 docker run -d -p 9411:9411 openzipkin/zipkin
 ```
 
+:warning: The zipkin provider does not support the v1 API since coredns 1.7.1
```
miekg, line 59: is this useful information for the reader or can we just not mention it?

plugin/trace/setup.go
```diff
@@ -35,7 +35,9 @@ func traceParse(c *caddy.Controller) (*trace, error) {
 	)
 
 	cfg := dnsserver.GetConfig(c)
-	tr.serviceEndpoint = cfg.ListenHosts[0] + ":" + cfg.Port
+	if cfg.ListenHosts[0] != "" {
+		tr.serviceEndpoint = cfg.ListenHosts[0] + ":" + cfg.Port
```
miekg, line 39: should we *not* add the port here? As you say in the description this doesn't work with zipkin...

plugin/trace/setup_test.go
```diff
@@ -8,31 +8,34 @@ import (
...
-		every        uint64
-		serviceName  string
-		clientServer bool
+		input           string
+		shouldErr       bool
+		endpoint        string
+		every           uint64
+		serviceName     string
+		clientServer    bool
+		serviceEndpoint string
```
miekg: serviceEndpoint is empty in all tests? In general if you want to test a little new thing, just add a small new test.

plugin/trace/README.md
nyodas: This is a "breaking"  change compared to pre-1.7.1
I don't mind removing the note, I made it in response to this [comment](https://github.com/coredns/coredns/pull/4171#issuecomment-703038038)

plugin/trace/setup_test.go
nyodas: Fair.

plugin/trace/setup.go
nyodas: Zipkin does understand the port when associated with a resolvable hostname / an IP, and there is always a port in the config.

I assume to track the the correct listener you need the right couple of bind + port

That would also work with no port but would be less precise

plugin/trace/setup_test.go
nyodas: Removed it.

johnbelamaric:
@nyodas can you resolve conflicts please?

nyodas:
Done.
My mistake for all the new reviewer.
My rebase added many commit unrelated to this and auto-added Codeowner as reviewer.
I can't seem to remove the new ones :/

mindw:
@nyodas The `bind` workaround + explicit URL works with 1.8.0:
```
. {
    ...

    bind 0.0.0.0

    # enable to output traces to jaeger
    trace zipkin http://{$HOST_IP}:30411/api/v2/spans {
      every 100
      client_server
    }
   ...
}
```

johnbelamaric:
Thanks!

/lgtm
/merge

PR #4182, opened 2020-10-06
fix alpn for http/2 upgrade when using DoH

### LINKED ISSUE #2566: dns over https not support HTTP/2 and TLS 1.3
I have the following configuration, and finally found that it does not support http2, nor does it support tls1.3. How do I need to configure it to support it?
please help me!

I'm using coredns version CoreDNS-1.3.1

https://.:443 {
  hosts xxx.hosts {
    fallthrough
  }

  forward . tls://xxx {
    tls_servername xxx
    force_tcp
    max_fails 3
    expire 10s
    health_check 5s
    policy sequential
    except company.com
  }

  tls xxx.cer xxx.key
  loadbalance round_robin
  reload 6s
  log
  errors
  rewrite edns0 subnet set 24 56
}

balboah:
### What

Provide "h2" in NextProtos for the TLS config that are set on the https server.

### Why

This is used for ALPN which is required to be able to upgrade the connection to http/2 which is recommended for DoH.
Without this we will be left with http/1.1 only.

Fixes #2566

### DIFF
--- a/core/dnsserver/server_https.go
+++ b/core/dnsserver/server_https.go
@@ -41,6 +41,9 @@ func NewServerHTTPS(addr string, group []*Config) (*ServerHTTPS, error) {
 	if tlsConfig == nil {
 		return nil, fmt.Errorf("DoH requires TLS to be configured, see the tls plugin")
 	}
+	// http/2 is recommended when using DoH. We need to specify it in next protos
+	// or the upgrade won't happen.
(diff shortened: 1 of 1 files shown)


miekg (approved):

miekg:
next to this, it's probably time to start testing DoH queries in our testing infra

PR #4188, opened 2020-10-09
plugin/autopath: slightly tweaks the docs

miekg:
Make the first sentence of the intro slightly easier to read. Refer to a
bugs section, just like other plugins do.

### DIFF
--- a/plugin/autopath/README.md
+++ b/plugin/autopath/README.md
@@ -6,13 +6,13 @@
 
 ## Description
 
-If it sees a query that matches the first element of the configured search path, *autopath* will
+If the *autopath* plugins sees a query that matches the first element of the configured search path, it will
 follow the chain of search path elements and return the first reply that is not NXDOMAIN. On any
 failures, the original reply is returned. Because *autopath* returns a reply for a name that wasn't
-the original question it will add a CNAME that points from the original name (with the search path
+the original question, it will add a CNAME that points from the original name (with the search path
 element in it) to the name of this answer.
 
-**Note**: There are several known issues.  See section below.
+**Note**: There are several known issues, see the "Bugs" section below.
 
 ## Syntax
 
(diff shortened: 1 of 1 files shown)


plugin/autopath/README.md
```diff
@@ -6,13 +6,13 @@
 
 ## Description
 
-If it sees a query that matches the first element of the configured search path, *autopath* will
+If the *autopath* plugins sees a query that matches the first element of the configured search path, it will
```
chrisohaver:
```suggestion
If the *autopath* plugin sees a query that matches the first element of the configured search path, it will
```

plugin/autopath/README.md
```diff
@@ -50,18 +50,19 @@ autopath @kubernetes
...
+Pod cache to resolve the client's IP address to a Pod. The Pod cache is maintained by an API watch
+on Pods. When Pod IP assignments change, the Kubernetes API notifies CoreDNS via the API watch.
 However, that notification is not instantaneous. In the case that a Pod is deleted, and it's IP is
-immediately provisioned to a Pod in another namespace, and that new Pod make a DNS lookup *before* the API watch
-can notify CoreDNS of the change, *autopath* will resolve the IP to the previous Pod's namespace.
+immediately provisioned to a Pod in another namespace, and that new Pod make a DNS lookup *before*
+the API watch can notify CoreDNS of the change, *autopath* will resolve the IP to the previous Pod's
+namespace.
 
 In Kubernetes, *autopath* is not compatible with Pods running from Windows nodes.
```
chrisohaver, line 65: Maybe move this out of the Bugs section... It's a limitation of k8s/windows, which doesn't support a domain search list.  Or perhaps just remove it.

chrisohaver (changes requested):

plugin/autopath/README.md
miekg: maybe indeed just remove it. On that bugs section if we only get NXDOMAIN we may be able to return a non-nonsensical answer that isn't NXDOMAIN to still short cut the resolution process; not sure what the client side implication are though

chrisohaver: The original incarnation of autopath had this option - to return another response to make the client stop searching. IIRC, SERVFAIL worked in my tests.  But it was an option, since I was unsure how various client resolvers, or applications would handle various negative responses.  It's probably too fiddly of a detail to clutter the config with. The failure mode is just "more work", not "wrong answer".

chrisohaver (approved):

Issue #4183, opened 2020-10-06
forward plugin issues with truncated responses without force_tcp or prefer_udp config


kishorj:
**What happened**:
When either prefer_udp or force_tcp options are not set in the forward config, CoreDNS follows the client protocol when talking to the upstream servers. If a client sent DNS request over UDP without the EDNS0 option, and the upstream server sent a truncated response, CoreDNS doesn't retry the request over TCP. It also doesn't set the TC flag in the response sent to the client. As a result, client ends up with a poetntially incomplete DNS response, and potentially fail to resolve the hostname permanently.

**What you expected to happen**:
In case the client sent requests over UDP and force_tcp or perfer_udp option is not set in the config, handle the truncated responses by either retrying over TCP or pass on the TC flag to the client.  

**How to reproduce it (as minimally and precisely as possible)**:
Use forward config without force_tcp or prefer_udp option, dns client use UDP, and not set the EDNS0 option. For example
```
dig +noedns site.with.truncated.response
```
**Anything else we need to know?**:

**Environment**:

- the version of CoreDNS: 1.6.4, 1.6.6, 1.6.7, 1.7.0
- Corefile:
```
  Corefile: |
    .:53 {
        errors
        health
        kubernetes cluster.local in-addr.arpa ip6.arpa {
          pods insecure
          fallthrough in-addr.arpa ip6.arpa
        }
        prometheus :9153
        forward . /etc/resolv.conf
        cache 30
        loop
        reload
        loadbalance
    }
```
- logs, if applicable:
- OS (e.g: `cat /etc/os-release`):
- Others:

chrisohaver:
> `dig +noedns site.with.truncated.response`

dig will automatically retry with tcp if the response is truncated. When it does, it doesn't show the intermediate truncated response.  If you use the `+ignore` flag, it will not retry, and display the truncated response.

chrisohaver:
I verified that CoreDNS passes along truncated responses, with the TC bit set to the client.  I tested with CoreDNS on current master branch configured to run 2 servers, one forwarding to the other using template to create a large response.

kishorj:
> > `dig +noedns site.with.truncated.response`
> 
> dig will automatically retry with tcp if the response is truncated. When it does, it doesn't show the intermediate truncated response. If you use the `+ignore` flag, it will not retry, and display the truncated response.

[...]

PR #4190, opened 2020-10-10
plugin/cache: fix removing OPT

miekg:
By checking state.Do() were are checking if the request had DO, but
we are _always_ adding Do now - do we need to save the DO from the
ORIGINAL request, which must be done in the ResponseWriter.

Also skip OPT records in filterDNSSEC as we can't set the TTL on those
records, this prevents writing a number to OPT's MBZ.

Note none of the tests have changed and still PASS. <strike>This is due to
the fact that CoreDNSServerAndPorts isn't a full server as we start in
main, it lacks the scrubwriter for instance. This is not bad per se, but
should be documented in the test code.</strike>
I don't know why this is.

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -104,6 +104,7 @@ type ResponseWriter struct {
 	state  request.Request
 	server string // Server handling the request.
 
+	do         bool // When true the original request had the DO bit set.
 	prefetch   bool // When true write nothing back to the client.
 	remoteAddr net.Addr
 }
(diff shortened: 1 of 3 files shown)


chrisohaver:
Likely fixes this: https://github.com/coredns/coredns/issues/4183#issuecomment-705685344 (not the whole issue, but the "cached response is always edns" observation)

chrisohaver (approved):
LGTM - would be nice to have a test covering, but understand that can be complicated.  Perhaps we could add an integration test (circle ci).

Issue #4202, opened 2020-10-14
plugin/kubernetes: Adjust Endpoints watch to support dual stack


chrisohaver:
**What would you like to be added**:

Migrate Endpoints watch to EndpointSlice to add support for dual stack headless services/endpoints.

**Why is this needed**:

In the latest Kubernetes implementation of dual-stack, the core Endpoints object will not be populated with both IPv6/IPv4 IPs for dual stack headless Services. Both IPs are available in the EndpointSlice object. EndpointSlice is currently beta, but should be GA next k8s release.

chrisohaver:
Work in progress: https://github.com/coredns/coredns/compare/master...chrisohaver:k8s-epslices

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/kubern..." ]
>Work in progress: master...chrisohaver:k8s-epslices

I wondered when we would need to look at endpoint slices. How do we know to use it?
Version check on the api?

chrisohaver:
> Version check on the api?

Not sure yet. I think whether or not kube-proxy etal uses EndpointSlice as the "source of truth" for service->pod routing was optional/configurable in k8s for some number of releases .... so I may need to hunt down that config setting (hopefully stashed somewhere in the API).

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/kubern..." ]
>    Version check on the api?
>
>Not sure yet. I think whether or not kube-proxy etal uses EndpointSlice as the
>"source of truth" for service->pod routing was optional/configurable in k8s for
>some number of releases .... so I may need to hunt down that config setting
>(hopefully stashed somewhere in the API).

linkerd and prometheus have added EndpointSlice, so here's hoping that they already solved
that problem.

Can you just query for it and see what the API returns?

chrisohaver:
Of course, It would be nice to just cut over to EndpointSlice instead of supporting both in parallel. And I think we _almost_ could have ...

EndpointSlice has been populated with Service->Pod targets by default since 1.18. ... but ...

[...]

PR #4209, opened 2020-10-15
plugin/kubernetes: Watch EndpointSlices

chrisohaver:

Watch EndpointSlice instead of Endpoints if the k8s server supports it.

I've done light manual testing on a kind cluster with a headless service that selects both ipv4 and ipv6 endpoints (in terms of service discovery, effectively a dual-stack headless service). I have not tested this on a "true" dual-stack cluster (but I don't think that matters in the context of just service discovery).

#4202

probably none

no


chrisohaver:
I suspect the circleci tests might fail because it's running a version of k8s that does not support endpoint mirroring, but might support endpoint slices v1beta1.

chrisohaver:
The last few commits add de-dup (where not already present) and tests for EndpointSlice skew address duplication.  With EndpointSlices, duplication of an Endpoint's Address is possible since we don't receive all EndpointSlices for any given Service at one time. https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#duplicate-endpoints

chrisohaver:
> I have not tested this on a dual-stack cluster.

The [headless Service with custom Endpoints containing both IPv4 and IPv6 addresses](https://github.com/coredns/ci/blob/master/build/kubernetes/dns-test.yaml#L178-L204) in coredns/ci tests more or less covers this.  The [`SRV` test](https://github.com/coredns/ci/blob/master/test/kubernetes/srv_test.go#L73-L90) verifies that both IPv4 and IPv6 addresses of the headless Service are present in the response.

chrisohaver:
EndpointSlices will be GA in K8s 1.20.  It would be cool if we are able get a version of CoreDNS with EndpointSlice support released in time to include with the K8s 1.20 release (code freeze Nov 12).

plugin/kubernetes/controller.go
```diff
@@ -246,6 +272,62 @@ func namespaceListFunc(ctx context.Context, c kubernetes.Interface, s labels.Sel
...
+	return func(options meta.ListOptions) (watch.Interface, error) {
+		if s != nil {
+			options.LabelSelector = s.String()
+		}
+		w, err := c.CoreV1().Services(ns).Watch(ctx, options)
+		return w, err
+	}
+}
+
+func podWatchFunc(ctx context.Context, c kubernetes.Interface, ns string, s labels.Selector) func(options meta.ListOptions) (watch.Interface, error) {
```
bradbeam, line 286: is this being used anywhere?

plugin/kubernetes/controller.go
```diff
@@ -226,6 +242,16 @@ func podListFunc(ctx context.Context, c kubernetes.Interface, ns string, s label
 	}
 }
 
+func endpointSliceListFunc(ctx context.Context, c kubernetes.Interface, ns string, s labels.Selector) func(meta.ListOptions) (runtime.Object, error) {
+	return func(opts meta.ListOptions) (runtime.Object, error) {
+		if s != nil {
+			opts.LabelSelector = s.String()
+		}
+		listV1, err := c.DiscoveryV1beta1().EndpointSlices(ns).List(ctx, opts)
```
bradbeam: nit, should we just `return c.DiscoveryV1beta1().EndpointSlices(ns).List(ctx, opts)`

plugin/kubernetes/controller.go
chrisohaver: Yes.  It's used when we watch Pods in with the option `pods verified`.

plugin/kubernetes/controller.go
chrisohaver: Did it in the other ListFunc and WatchFunc Functions too (for Services, Pods, and Endpoints).

rajansandeep (approved):
Thank you for adding this @chrisohaver 
This looks good to me!

plugin/kubernetes/controller.go
```diff
@@ -130,16 +133,29 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
...
-			&cache.ListWatch{
-				ListFunc:  endpointsListFunc(ctx, dns.client, api.NamespaceAll, dns.selector),
-				WatchFunc: endpointsWatchFunc(ctx, dns.client, api.NamespaceAll, dns.selector),
-			},
-			&api.Endpoints{},
-			cache.ResourceEventHandlerFuncs{AddFunc: dns.Add, UpdateFunc: dns.Update, DeleteFunc: dns.Delete},
-			cache.Indexers{epNameNamespaceIndex: epNameNamespaceIndexFunc, epIPIndex: epIPIndexFunc},
-			object.DefaultProcessor(object.ToEndpoints(opts.skipAPIObjectsCleanup), dns.recordDNSProgrammingLatency),
-		)
+		if opts.useEndpointSlices {
```
miekg, line 142: this needs to be simplified; seems `&api.EndpointSlice{}` is the only difference? That can be set latter in a small if-statement.

The endpoint slice also missed the dns latency for some reason?

Update: only the ListWatch seems difference, but the same argument holds

plugin/kubernetes/controller.go
```diff
@@ -241,8 +263,56 @@ func namespaceListFunc(ctx context.Context, c kubernetes.Interface, s labels.Sel
 		if s != nil {
 			opts.LabelSelector = s.String()
 		}
-		listV1, err := c.CoreV1().Namespaces().List(ctx, opts)
-		return listV1, err
+		return c.CoreV1().Namespaces().List(ctx, opts)
+	}
+}
+
+func serviceWatchFunc(ctx context.Context, c kubernetes.Interface, ns string, s labels.Selector) func(options meta.ListOptions) (watch.Interface, error) {
```
miekg, line 277: are these functions moved? Or why do we need these now?

plugin/kubernetes/handler_test.go
```diff
@@ -643,8 +643,9 @@ var epsIndex = map[string][]*object.Endpoints{
 				},
 			},
 		},
-		Name:      "svc1",
+		Name:      "svc1-aaaaa",
```
miekg: why is the -`aaaaa` added here?

plugin/kubernetes/kubernetes.go
```diff
@@ -244,6 +245,17 @@ func (k *Kubernetes) InitKubeCache(ctx context.Context) (err error) {
 
 	k.opts.zones = k.Zones
 	k.opts.endpointNameMode = k.endpointNameMode
+	// Enable use of endpoint slices if the API supports the discovery v1 beta1 api
+	if _, err := kubeClient.Discovery().ServerResourcesForGroupVersion(discovery.SchemeGroupVersion.String()); err == nil {
+		k.opts.useEndpointSlices = true
```
miekg, line 251: log.Infof this?

plugin/kubernetes/kubernetes.go
```diff
@@ -244,6 +245,17 @@ func (k *Kubernetes) InitKubeCache(ctx context.Context) (err error) {
...
+	}
+	// Disable use of endpoint slices for k8s versions 1.17 and 1.18. Endpoint slices were
+	// introduced in 1.17 but not fully integrated and enabled by default until k8s 1.19.
+	sv, _ := kubeClient.ServerVersion()
+	if strings.HasPrefix(sv.String(), "v1.18") || strings.HasPrefix(sv.String(), "v1.17") {
```
miekg: hasn't the version stuff of k8s a whole api to check this? Meaning sv.Major() and friends exist and can be checked here?

plugin/kubernetes/kubernetes.go
```diff
@@ -455,6 +469,11 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 							s := msg.Service{Host: addr.IP, Port: int(p.Port), TTL: k.ttl}
 							s.Key = strings.Join([]string{zonePath, Svc, svc.Namespace, svc.Name, endpointHostname(addr, k.endpointNameMode)}, "/")
 
+							if _, dup := uniq[s]; dup {
```
miekg: unclear why this is needed and I think we _alread_ remove duplicates from the answer

plugin/kubernetes/kubernetes_test.go
```diff
@@ -151,22 +152,24 @@ func (APIConnServiceTest) EpIndex(string) []*object.Endpoints {
...
-						{IP: "172.0.0.3"},
+						{IP: "172.0.0.2"},
 					},
 					Ports: []object.EndpointPort{
 						{Port: 80, Protocol: "tcp", Name: "http"},
 					},
 				},
 			},
-			Name:      "hdls1",
+			Name:      "hdls1-bbbbb",
```
miekg: unclear what these suffixes mean and why they need to be added to existing test cases?

plugin/kubernetes/ns.go
```diff
@@ -82,9 +88,10 @@ func (k *Kubernetes) nsAddrs(external bool, zone string) []dns.RR {
 		rr.Hdr.Name = svcNames[i]
 		rr.A = ip
 		rrs[i] = rr
+		i++
 	}
 
-	return rrs
+	return rrs[0:i]
```
miekg: this can be shorted to `rrs[:i]` but I don't think we need this because uniqueness is done else where.

plugin/kubernetes/ns_test.go
```diff
@@ -96,8 +98,23 @@ func (APIConnTest) EpIndexReverse(ip string) []*object.Endpoints {
 					},
 				},
 			},
-			Name:      "dns6-service",
+			Name:      "dns6-service-aaaaa",
```
miekg: please make this meaningful or add a completely new test

plugin/kubernetes/object/endpoint.go
```diff
@@ -56,6 +57,17 @@ func ToEndpoints(skipCleanup bool) ToFunc {
 	}
 }
 
+// SliceToEndpoints returns a function that converts an *discovery.EndpointSlice to a *Endpoints.
+func SliceToEndpoints(skipCleanup bool) ToFunc {
```
miekg: inclined to say this should move to an `endpointslice.go` file - if these are actual objects in the k8s api, which I think they are?
function name should also be `EndpointSliceToEndpoints` to make super clear what being converrted.

plugin/kubernetes/object/endpoint.go
```diff
@@ -108,6 +120,49 @@ func toEndpoints(skipCleanup bool, end *api.Endpoints) *Endpoints {
...
+			if end.TargetRef != nil {
+				ea.TargetRefName = end.TargetRef.Name
+			}
+			// EndpointSlice does not contain NodeName, leave blank
+			e.Subsets[0].Addresses = append(e.Subsets[0].Addresses, ea)
+			e.IndexIP = append(e.IndexIP, a)
+		}
+	}
+
+	if !skipCleanup {
```
miekg, line 159: Only metrics tests seems to use this boolean?
~~~
 grep skipAPIObjectsCleanup -ri 
metrics_test.go:		skipAPIObjectsCleanup: true,
~~~
If so, please remove this from the API

plugin/kubernetes/xfr.go
```diff
@@ -84,10 +84,6 @@ func (k *Kubernetes) Transfer(zone string, serial uint32) (<-chan []dns.RR, erro
 				endpointsList := k.APIConn.EpIndex(svc.Name + "." + svc.Namespace)
 
 				for _, ep := range endpointsList {
-					if ep.Name != svc.Name || ep.Namespace != svc.Namespace {
```
miekg, line 87: why this change?

plugin/kubernetes/object/endpoint.go
miekg: this can be done in a new PR

plugin/kubernetes/controller.go
chrisohaver: Thanks for reviewing! Agreed, I can reduce some code there.

> The endpoint slice also missed the dns latency for some reason?

When I tested in a local cluster, I saw that EndpointSlices did not have the annotation (or equivalent other field) required to calculate latency (something that holds the date of the last update).  I'll double check in latest k8s release.

plugin/kubernetes/controller.go
chrisohaver: They are not new, they are moved from `watch.go`, which was originally held an implementation of the since removed Watcher interface. At some point, these completely unrelated (except by name only) k8s api WatchFunc functions snuck in there.  When the CoreDNS watch feature was removed, the Watcher implementation was removed, and all that was left were the k8s api WatchFunc.

plugin/kubernetes/handler_test.go
chrisohaver: This is what endpointSlice names look like... by convention: the service name, dash, 5 random alpha-nums.
Having this reflected in the tests that we do not have assumptions in the code that and endpoint name = service name, which is no longer the case for EndpointSlice.

plugin/kubernetes/kubernetes.go
chrisohaver: IIRC, the backend does duplicate removal for some query types, but not all (e.g. not for A/AAAA).

I added de-dup (where not already present) and tests for EndpointSlice skew address duplication. With EndpointSlices, duplication of an Endpoint's Address is possible since we don't receive all EndpointSlices for any given Service at one time. https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#duplicate-endpoints

plugin/kubernetes/kubernetes_test.go
chrisohaver: This is what endpointSlice names look like... by convention: the service name, dash, 5 random alpha-nums.
Having this reflected in the tests that we do not have assumptions in the code that and endpoint name = service name, which is no longer the case for EndpointSlice.

plugin/kubernetes/xfr.go
aojea: I assume that this is because previously there was a 1:1 relation service - endpoint, referenced by name.
with endpoint slices service to slices is 1:N and the slices reference the service by label
xref: https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#ownership

chrisohaver: It was a pointless check. Name and namespace would always match, since we are getting the list from an index lookup on name and namespace. It was probably copy-pasted from `findServices()` which does need to check due to wildcard queries. 
Anyways, now comparing object.Endpoint.Name to object.Service.Name is no longer valid, since the object.Endpoint could have been derived from an api.EndpointSlice instead of a api.Endpoint.

plugin/kubernetes/object/endpoint.go
chrisohaver: EndpointSlices _are_ separate objects, but to simplify supporting both Endpoint and EndpointSlice in the code at the same time, I am distilling both Endpoints and EndpointSlice into object.Endpoint, instead of adding a new struct called object.EndpointSlice.  Since the new conversion functions work with the existing object.Endpoints struct, I put them in `object/endpoints.go`, but I can move them to a separate file if that makes more sense.

> function name should also be EndpointSliceToEndpoints to make super clear what being converrted.

Yeah - I can change that.

aojea:
@chrisohaver unrelated to this PR, but taking the opportunity to comment because it will affect CoreDNS.
Something new in tKuberntes dual-stack is that Services will pluralize the ClusterIP, so dual-stack services may also have ClusterIPs as an array with one IPv4 and one IPv6 address

chrisohaver:
> @chrisohaver unrelated to this PR, but taking the opportunity to comment because it will affect CoreDNS.
> Something new in tKuberntes dual-stack is that Services will pluralize the ClusterIP, so dual-stack services may also have ClusterIPs as an array with one IPv4 and one IPv6 address

@aojea, OK thanks (IIRC, I recently reviewed the PR that added this). Has that landed in a released version of the API yet?

aojea:
> @aojea, OK thanks (IIRC, I recently reviewed the PR that added this). Has that landed in a released version of the API yet?

it landed yesterday https://github.com/kubernetes/kubernetes/pull/91824, 
but will be under a feature gate in alpha and not enabled by default in 1.20

chrisohaver:
> > @aojea, OK thanks (IIRC, I recently reviewed the PR that added this). Has that landed in a released version of the API yet?
> 
> it landed yesterday [kubernetes/kubernetes#91824](https://github.com/kubernetes/kubernetes/pull/91824),
> but will be under a feature gate in alpha and not enabled by default in 1.20

OK.  IIUC, we shouldn't add support for _that_ to a CoreDNS release until a version of the K8s API is released with that in it. Otherwise I think we'd need to release a CoreDNS which was compiled against master k8s-api or a pre-release tag.

chrisohaver:
A general note: Supporting both Endpoints and EndpointSlices in the codebase can be relatively short lived.  When K8s 1.21 is released, the 3 supported k8s would be 1.21, 1.20, 1.19, all of which fully support EndpointSlices. So IMO, we can drop the Endpoints watch code with the version of CoreDNS that is released with K8s 1.21.

johnbelamaric:
Support is now 12 months not 9, so typically that means 4 releases. FYI.

plugin/kubernetes/ns_test.go
chrisohaver: Thanks. I've refactored, added comments and made the test data more compact and (i think) easier to read.  The test I had modified wasn't testing the right thing, and fixing it exposed a bug in nsAddrs, which is ~not~ _now_ fixed.

plugin/kubernetes/controller.go
chrisohaver: ~Ah, I found an updated timestamp in another place in the EndpointSlice object, I'll adjust the code to use it.~

No, OK , I see now that the same annotation _is_ added when EndpointSlices are created by the EndpointSliceController, but not when created by the EndpointSliceMirrorController.  So the existing code should work, with some generification to handle both Endpoints and EndpointSlice.

miekg: listWatch can be moved out of the if-then, and just assigned, making this more readeble

plugin/kubernetes/controller.go
```diff
@@ -130,15 +133,35 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
...
+			toFuncFunc        func(bool) object.ToFunc
+			recordLatencyFunc object.RecordLatencyFunc
+		)
+		if opts.useEndpointSlices {
+			apiObj = &discovery.EndpointSlice{}
+			listWatch = &cache.ListWatch{
+				ListFunc:  endpointSliceListFunc(ctx, dns.client, api.NamespaceAll, dns.selector),
+				WatchFunc: endpointSliceWatchFunc(ctx, dns.client, api.NamespaceAll, dns.selector),
+			}
+			toFuncFunc = object.EndpointSliceToEndpoints
```
miekg: why FuncFunc?

plugin/kubernetes/handler_test.go
miekg: really, wow? Please add a comment with ref to some docs, this looks like someone was just banging the keyboard, not a conscious design decision.

plugin/kubernetes/kubernetes.go
miekg: the dedup exists in backend.go (other discussion if that is still useful), but we should add it there, not here.
Thanks the ref to those docs

plugin/kubernetes/object/endpoint.go
miekg: https://github.com/coredns/coredns/pull/4244

plugin/kubernetes/xfr.go
miekg: ack ,thanks

plugin/kubernetes/handler_test.go
chrisohaver: I'll make them just incremental numbers.  The random string is the name scheme that K8s controllers use when creating endpoints (similar to how Pods are named when deployed per Deployments). I was just imitating that to make the data more "real", but it's not necessary at all.  All that matters is that they don't match the service name.

plugin/kubernetes/controller.go
chrisohaver: It's a function that returns function that returns a function.  IOW, a function that returns an object.ToFunc, not an object.ToFunc itself. Made sense to me, but I'll rename it to just `to`.

plugin/kubernetes/kubernetes.go
chrisohaver: Yes, the backend already does de-dup for all types that I can see. I was staring right at it and didn't see it.  I've removed all de-dup from this PR.

plugin/kubernetes/kubernetes.go
miekg: is this done? I want to resolve my remaining comments, so I can lgtm this

chrisohaver: Yes. A message is logged when disabling endpointslice watch for k8s before 1.19.

miekg (approved):
looks good enough

pires:
It seems this PR actually introduces a breaking change as CoreDNS will not report `kubernetes` plug-in is ready if it's not given permissions to watch `EndpointSlice`.

I just reproduced this w/ a new cluster provisioned w/ kubeadm 1.19.10.

aojea:
I have a PR for fixing it, kubeadm should provide the permissions for slices

chrisohaver:
> It seems this PR actually introduces a breaking change

~@pires, Correct.  This is fixed in #4580.~

Correction: This PR did introduce a breaking change, but not the one you refer to.

> CoreDNS will not report kubernetes plug-in is ready if it's not given permissions to watch EndpointSlice

This is intended behavior.  CoreDNS needs to have access to endpointslice to work in k8s >= 1.19.

pires:
>CoreDNS needs to have access to endpointslice to work in k8s >= 1.19.

I can't find any reference to this in the release docs for any of the 1.8.x patches. So let's agree to disagree.

Thank you, @aojea for addressing this on the kubeadm side.

chrisohaver:
@pires, https://github.com/coredns/deployment/pull/246

chrisohaver:
In retrospect, we should have bumped the minor version and documented this as a backward incompatible deprecation. 

... I've added it to the 1.8.1 notes retroactively, FWIW.

pires:
Thank you!

Issue #3758, opened 2020-03-19
Add coredns_hosts_request_count_total metric


Diaoul:
**What would you like to be added**:
I would like a `coredns_hosts_request_count_total` to be added in the same way `coredns_forward_request_count_total` works.
Could add the `file` sub-filter similar to the `to` one in the `forward` plugin.

**Why is this needed**:
There is currently no metric as to what goes through the `hosts` plugin. The `file` and `auto` plugins have the `zone` sub-filter available, this is not possible for `hosts` as it is responsible for potentially a huge amount of "zones".

sharmaansh21:
I can work on this :)

miekg:
it probably makes sense to remove this kind of per plugin metrics and have a builtin one that export queries seen per plugin. Now adding zones to such a metrics will blow up the metrics from a label cardinality standpoint...

We have metric changes queued for 1.7.0 so the time is now to make this change.

miekg:
hmm, just looking at metrics; we obvs don't have previous plugin information there.

Why do you care about this metric here for host?

stp-ip:
> it probably makes sense to remove this kind of per plugin metrics and have a builtin one that export queries seen per plugin. Now adding zones to such a metrics will blow up the metrics from a label cardinality standpoint...
> 
> We have metric changes queued for 1.7.0 so the time is now to make this change.

Thinking about that. Most of the plugins seem to want to know how often a specific plugin is hit in total.
We could make this into a general top level metric instead. As the middleware handover code caddy it might mean it's hard to do so.

Something along the lines of:
`coredns_plugin_count_total{server, zone, plugin}`

And even the plugin flow total might be interesting, but increases cardinality:
`coredns_plugin_count_total{server, zone, plugin-chain}`

*plugin-chain being "forward-nsid-custom" for example.

Diaoul:
In my case I use hosts as a blacklist, I can see other uses cases where hosts is used for some extra local resolution. It is useful to know how often the hosts is hit. For example to compare to other plugins.
I like the top level idea but we would loose some plugin-specific cardinality.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add coredns_h..." ]
>> it probably makes sense to remove this kind of per plugin metrics and have a builtin one that export queries seen per plugin. Now adding zones to such a metrics will blow up the metrics from a label cardinality standpoint...
>>
>> We have metric changes queued for 1.7.0 so the time is now to make this change.
[...]

PR #4221, opened 2020-10-20
plugins/hosts: Added prometheus metric "coredns_hosts_hits_total"

notanton:
This pull request adds a prometheus metric to the `hosts` plugin `coredns_hosts_hits_total` which is a counter for the number of requests served from the hosts file.

N/A, I made this change for my local setup.

I have updated the README for the plugin. If there is an intent to merge, I will make a PR to [coredns/coredns.io](https://github.com/coredns/coredns.io) to update the plugin page to include the new metric.

In order to accurately capture data for the metric, I have changed the order of plugin execution moving `hosts` before `cache` in `plugin.cfg` to ensure cached requests correctly increment the counter. I don't believe this causes any issues, but this is my first contribution to the project.


miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugins/hosts: Ad..." ]
>2. Which issues (if any) are related?
>
>N/A, I made this change for my local setup.

this has been discussed before (though I can't find it now). But metrics are not free and
if something like this is going to be added it will be generic for all plugins, not just
one (but that's harder than it sounds).

so this will not be merged.

chrisohaver:
Here's the prior discussion #3758. 

I'll spend some time today thinking on a generic approach.

chrisohaver:
Closing per https://github.com/coredns/coredns/pull/4223#issuecomment-714288729.

PR #4223, opened 2020-10-21
plugin/metrics: add metric for "plugin that served the response"

chrisohaver:

This adds a response metric per the plugin that "served" the response. 
It's based on the assumption that the deepest plugin reached in the chain is the plugin that "originated" the response. In most/all cases, this _is_ the plugin that essentially "served" the data. Perhaps a more truthful/literal name for the metric would be "coredns_deepest_plugin_reached" or something.  But IMO that is too cryptic.

Plugins like rewrite make this a little fuzzy. E.g. if a plugin like re-wrtite modifies a response, the original creator of the response is logged as the plugin, not rewrite.

In theory, one could write plugins that break this assumption.  e.g. a plugin that looks at a response from a next plugin, and replaces the response with something else entirely - like a response based policy.

This kind of metric has been requested for the host plugin a couple of times.

#3758
#4221

document the metric

### DIFF
--- a/plugin/metrics/handler.go
+++ b/plugin/metrics/handler.go
@@ -16,6 +16,9 @@ import (
 func (m *Metrics) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 	state := request.Request{W: w, Req: r}
 
+	var p string
+	ctx = context.WithValue(ctx, plugin.Key{}, &p)
+
 	qname := state.QName()
 	zone := plugin.Zones(m.ZoneNames()).Matches(qname)
 	if zone == "" {
@@ -27,6 +30,9 @@ func (m *Metrics) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	status, err := plugin.NextOrFailure(m.Name(), m.Next, ctx, rw, r)
 
 	vars.Report(WithServer(ctx), state, zone, rcode.ToString(rw.Rcode), rw.Len, rw.Start)
+	if p != "" {
+		PluginResponseCount.WithLabelValues(p).Add(1)
+	}
 
 	return status, err
 }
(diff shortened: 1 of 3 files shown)


chrisohaver:
Anyways - just opening up the PR to see others think this is a viable approach. If so, I can dress it up further with docs, tests, etc.  If no, we can close it. Thanks!

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] plugin/metrics: a..." ]
>This kind of metric has been requested for the host plugin a couple of times.

the 'why' has never been answered. You can even externally determine these value as you
know what's in the host file (or any other plugin) and the query logs (or dnstap).

chrisohaver:
> [ Quoting <notifications@github.com> in "[coredns/coredns] plugin/metrics: a..." ]
> This kind of metric has been requested for the host plugin a couple of times.
> the 'why' has never been answered. You can even externally determine these value as you know what's in the host file (or any other plugin) and the query logs (or dnstap).

Ah OK.  I didn't realize that you see no value in the feature at all. (I was thinking your concern for prior requests and PR was that the solution was specific to the hosts plugin, not generic to all plugins.  But if the generic feature itself has no value, then that kind of trumps all.)

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/metric..." ]
>Ah OK. I didn't realize that you see no value in the feature at all.

Not saying that, nothing comes for free and we already have metrics, *why* is this a good
one to add?

chrisohaver:
> [ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/metric..." ]
> Ah OK. I didn't realize that you see no value in the feature at all.
> Not saying that, nothing comes for free and we already have metrics, *why* is this a good one to add?

Gotcha - May have some value, but not convinced enough to warrant adding it.

Issue #4238, opened 2020-10-26
plugin/dnstap: slow test 


miekg:
The dnstapio tests a very slow:
~~~
% go t
2020/10/26 09:04:09 [WARNING] plugin/dnstap: Connection lost: write tcp 127.0.0.1:39918->127.0.0.1:41155: write: broken pipe
2020/10/26 09:04:09 [INFO] plugin/dnstap: Reconnected to dnstap
PASS
ok  	github.com/coredns/coredns/plugin/dnstap/dnstapio	9.046s
~~~
This needs to be sped up (< 1s). 

Also the `d.queue` uses a buffered channel (10,000 elements), no rational is given, so this should be made blocking. Further more for TCP we fiddle with nagle and set a (random sized) write buffer  - again without any rational. These need to be removed and/or given proper rational from the dnstap project.
*And* `d.socket` seem superfluous and can be removed.

miekg:
the buffered channel and the internal buffer used gives the impression this code can't block; TL:DR; it can block. I've got a WIP PR
that removes another 246 lines of code, but currently failing over the concurrency issues this uncovered.

miekg:
this has useful bits: https://github.com/coredns/coredns/pull/4240
but trying it with coredns, yields problems:
~~~
oreDNS-1.8.0
linux/amd64, go1.15.3, 
[ERROR] plugin/dnstap: Failed to flush dnstap: no connection
[INFO] plugin/dnstap: Reconnected to dnstap
^C[INFO] SIGINT: Shutting down
~~~
coredns now hangs

~~~
dnstap: opened input socket /tmp/dnstap.sock
dnstap.NewFrameStreamInput() failed: decoding error
dnstap.NewFrameStreamInput() failed: decoding error
~~~
on the dnstap side.

rdrozhdzh:
Hi @miekg, please review #1280 again before doing your changes.

Also, just for reference, see the recent perf problem which unbound team faced recently - 
https://github.com/NLnetLabs/unbound/issues/305

I hope, the dnstap plugin will not experience similar problems

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/dnstap..." ]
>Hi @miekg, please review #1280 again before doing your changes.
>
>Also, just for reference, see the recent perf problem which unbound team faced
>recently -
>NLnetLabs/unbound#305
>
>I hope, the dnstap plugin will not experience similar problems

[...]

PR #4239, opened 2020-10-26
Speed up testing

miekg:
* make notification run in the background, this recudes the test_readme
time from 18s to 0.10s
* reduce time for zone reload

* TestServeDNSConcurrent remove entirely. This took a whopping 58s for
  ... ? A few minutes staring didn't reveal wth it is actually testing.
  Making values smaller revealed race conditions in the tests. Remove
  entirely.

* Move many interval values to variables so we can reset them to short
  values for the tests.

* test_large_axfr: make the zone smaller. The number used 64K has no
  rational, make it 64/10 to speed up.
* TestProxyThreeWay: use client with shorter timeout

A few random tidbits in other tests.

Total time saved: 177s (almost 3m) - which makes it worthwhile again to
run the test locally:

this branch:

~~~
ok  	github.com/coredns/coredns/test	10.437s
cd plugin; time go t ./...
5,51s user 7,51s system 11,15s elapsed 744%CPU (
~~~

master:

~~~
ok  	github.com/coredns/coredns/test	35.252s
cd plugin; time go t ./...
157,64s user 15,39s system 50,05s elapsed 345%CPU ()
~~~
tests/ -25s
plugins/ -152s


miekg:
Another 10s can be shaved off as we fix https://github.com/coredns/coredns/issues/4238, meaning all tests run 10s which looks like nice ballpark figure to make local dev. fun again

miekg:
let me fix the timing as I took the wrong value from `time`

miekg:
ah flakyness in travis (of course  it's travis)

miekg:
travis is becoming very very slow

SuperQ:
I'm a fan of CircleCI :grin:

miekg:
yes, would also cut down on external deps, circle-ci is already finished, and that does a minikube k8s test!

miekg:
hmm, my initial commit msg had section headers (#) but these where filtered out because that text is semi-markdown and #-sign are comments... damnit

miekg:
if there are no objects I will merge this

stp-ip (approved):
/lgtm

test/reload_test.go
```diff
@@ -101,8 +100,6 @@ func TestReloadMetricsHealth(t *testing.T) {
 	}
 	defer c1.Stop()
 
-	time.Sleep(100 * time.Millisecond)
```
chrisohaver, line 104: the sleep was added here... https://github.com/coredns/coredns/pull/1715
not obvious what the rationale was.

chrisohaver (approved):

PR #4241, opened 2020-10-27
Fix golint warnings

zouy414:
Include:
1. plugin/forward/type.go:8:2: const typeUdp should be typeUDP
2. plugin/forward/type.go:9:2: const typeTcp should be typeTCP
3. plugin/forward/type.go:10:2: const typeTls should be typeTLS
4. plugin/kubernetes/metrics.go:24:2: var DnsProgrammingLatency should be DNSProgrammingLatency
5. plugin/kubernetes/metrics_test.go:124:102: func parameter clusterIp should be clusterIP

Signed-off-by: zouyu <zouy.fnst@cn.fujitsu.com>

Fix golint warning

### DIFF
--- a/plugin/forward/type.go
+++ b/plugin/forward/type.go
@@ -5,33 +5,33 @@ import "net"
 type transportType int
 
 const (
-	typeUdp transportType = iota
-	typeTcp
-	typeTls
+	typeUDP transportType = iota
+	typeTCP
+	typeTLS
 	typeTotalCount // keep this last
 )
 
 func stringToTransportType(s string) transportType {
 	switch s {
 	case "udp":
-		return typeUdp
+		return typeUDP
 	case "tcp":
-		return typeTcp
+		return typeTCP
 	case "tcp-tls":
-		return typeTls
+		return typeTLS
 	}
 
-	return typeUdp
+	return typeUDP
 }
 
 func (t *Transport) transportTypeFromConn(pc *persistConn) transportType {
 	if _, ok := pc.c.Conn.(*net.UDPConn); ok {
-		return typeUdp
+		return typeUDP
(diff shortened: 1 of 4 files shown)


plugin/acl/acl.go
```diff
@@ -77,7 +77,7 @@ RulesCheckLoop:
 	}
 
 	RequestAllowCount.WithLabelValues(metrics.WithServer(ctx)).Inc()
-	return plugin.NextOrFailure(state.Name(), a.Next, ctx, w, r)
+	return plugin.NextOrFailure(ctx, state.Name(), a.Next, w, r)
```
miekg: this is unfortunate, but it is backwards incompatible as a lot of plugins use this; so we've never changed it and apart from golint complaining, it's not _that_ bad

plugin/kubernetes/metrics_test.go
```diff
@@ -20,7 +20,7 @@ const (
 	namespace = "testns"
 )
 
-func TestDnsProgrammingLatency(t *testing.T) {
+func TestDNSProgrammingLatency(t *testing.T) {
```
miekg, line 23: lgtm

plugin/forward/type.go
```diff
@@ -5,33 +5,33 @@ import "net"
 type transportType int
 
 const (
-	typeUdp transportType = iota
-	typeTcp
-	typeTls
+	typeUDP transportType = iota
```
miekg, line 8: lgtm

plugin/acl/acl.go
zouy414: Thanks, I will restore it.

miekg (approved):

PR #4242, opened 2020-10-27
plugin/dnstap: remove custom encoder

miekg:
Those files reimplemented parts of the dnstap spec, we can just use the
dnstap functions for that. This leaves all the queuing that is enabled
and drops messages if the dnstap reader can't keep up. In the new code
flush() would never return an error (at least I couldn't make it do so),
so the reconnect functionally is moved to kick off when we get write
errors.

Some smaller cosmetic changes as well, `d.socket` is now `proto`, which
makes the dial() function smaller.

Total testing time is now <1s (which was the impetus to look into this
plugin *again*).

See #4238
The buffered channel needs to be sized correctly, as we may need to do
some queing if the dnstap reader can't keep up.


miekg:
if there are no objections I will merge this

zouyee (approved):

PR #4244, opened 2020-10-28
plugin/kubernetes: remove skipApiCleanup boolean

miekg:

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -89,9 +89,8 @@ type dnsControlOpts struct {
 	namespaceLabelSelector *meta.LabelSelector
 	namespaceSelector      labels.Selector
 
-	zones                 []string
-	endpointNameMode      bool
-	skipAPIObjectsCleanup bool
+	zones            []string
+	endpointNameMode bool
 }
 
 // newDNSController creates a controller for CoreDNS.
@@ -113,7 +112,7 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
 		&api.Service{},
 		cache.ResourceEventHandlerFuncs{AddFunc: dns.Add, UpdateFunc: dns.Update, DeleteFunc: dns.Delete},
 		cache.Indexers{svcNameNamespaceIndex: svcNameNamespaceIndexFunc, svcIPIndex: svcIPIndexFunc},
-		object.DefaultProcessor(object.ToService(opts.skipAPIObjectsCleanup), nil),
+		object.DefaultProcessor(object.ToService(), nil),
 	)
 
 	if opts.initPodCache {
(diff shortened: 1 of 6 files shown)


chrisohaver (commented):
IIRC, you had originally added the clean up step at the end of the To functions to help with memory footprint reduction.

chrisohaver:
> IIRC, you had originally added the clean up step at the end of the To functions to help with memory footprint reduction.

Presumably because the object is still referenced by the informer, so “zeroing” it out at least gets the things it referenced to be GCd. (I’m just presuming. I don’t know for sure)

miekg:
>    IIRC, you had originally added the clean up step at the end of the To
>    functions to help with memory footprint reduction.

hmm, good point, but wrapped in an if-statement iirc?

miekg:
yes, zeroing those values makes tests fail; this means that those tests are using code that's not used in prod. 

I'm not sure *who* much memory we are saving with doing this, or if it can be GCed... ?

chrisohaver:
> yes, zeroing those values makes tests fail; this means that those tests are using code that's not used in prod.
> 
> I'm not sure _who_ much memory we are saving with doing this, or if it can be GCed... ?

FWIW, I think this is the squashed commit message for adding that (from https://github.com/coredns/coredns/commit/830e97f8008ed83abb6b293a12a87f3fce34e423#diff-642b4aa397e466d410076c727182863bf09f5724a41f20ad14431bc3e86599ae)...


```
* Set incoming object to nil

Explicataliy discard the converted object; we did a deep copy it's
not needed anymore.

```

Unclear if it was done based on test results.

chrisohaver:
... and the PR comment: https://github.com/coredns/coredns/pull/2128#issuecomment-426531258

miekg:
>    yes, zeroing those values makes tests fail; this means that those tests are
>    using code that's not used in prod.
>
>    I'm not sure who much memory we are saving with doing this, or if it can be
>    GCed... ?
>
>FWIW, I think this is the squashed commit message for adding that (from 830e97f
>#diff-642b4aa397e466d410076c727182863bf09f5724a41f20ad14431bc3e86599ae)...
>
>* Set incoming object to nil
>
>Explicataliy discard the converted object; we did a deep copy it's
>not needed anymore.
>
>Signed-off-by: Miek Gieben <miek@miek.nl>
>
>Unclear if it was done based on test results.

it wasn't, difficult to test properly as well, setting to `nil` is save and quarantees the
object is gone and not referenced anymore (and thus gc-ed)

chrisohaver:
Actually - running through the PR conversation - it looks like memory footprint shrank significantly with this change.
footprint reduced by 10% prior to nilling, and then 27%/32% with the nilling.

miekg:
>Actually - running through the PR conversation - it looks like memory footprint
>shrank significantly with this change.
>footprint reduced by 10% prior to nilling, and then 27% with the nilling.

thanks for digging that up.

Then we left with the question that tests need a flag the specifically changes how
production works. I don't think that are good tests.

chrisohaver:
> Then we left with the question that tests need a flag the specifically changes how production works. I don't think that are good tests.

It's not ideal, but I think better than not having tests at all.

The valid concern is that the latency metric feature itself (not just the tests) might depend on the original api objects being left intact.

The [original defense of adding the `skipCleanup` boolean](https://github.com/coredns/coredns/pull/3171#discussion_r323463949), is a bit vague on precisely why this is needed.  Unclear if it's an artifact of how the fakeClient cache operates.

We might need to test this in coredns/ci ... in the kubernetes cluster tests.  We would not be able to check for specific latency values, like we do in these unit tests, but we could test that the metric produces "something reasonable".  Or perhaps we could if we manually modify the endpoint annotations (assuming that doesn't break stuff).

miekg:
>We might need to test this in coredns/ci ... in the kubernetes cluster tests.

I think that's a better approach, cleaner code and and better e2e testing.

/Miek

--
Miek Gieben

PR #4245, opened 2020-10-28
docs: move Also See to See Also

### LINKED ISSUE #4196: README's it's 'SEE ALSO' not Also See
Just looked at some man pages and they use 'See Also', not 'Also See' as the reference section. Should probably refer those words in all README.md's and in plugin.md where we detail this.

miekg:
sed -i 's/Also See/See Also/' plugin/**/README.md

Some plugins did already use 'See Also', so it's all consistent now.

Fixes: #4196


yongtang (approved):
LGTM

PR #4253, opened 2020-11-02
Remove boolean parameter from object.To() functions

miekg:
This was purely added to facilitate some testing, meaning the tests
don't test production code.

Remove the metrics and informer tests, these should be moved to
coredns/ci to be properly e2e tested.

Closes: #4244


chrisohaver:
Curious to see what the code cov diff is, and if we should be adding some unit tests (non-e2e) to cover some of the latency functions individually.

Also, perhaps we should add the e2e tests in coredns/ci _before_ removing the coredns/coredns tests so there is no coverage gap.

chrisohaver:
> ... perhaps we should add the e2e tests in coredns/ci before removing the coredns/coredns tests so there is no coverage gap.

I'm fiddling with this today.

plugin/kubernetes/metrics_test.go
```diff
@@ -1,223 +0,0 @@
...
-	now := time.Now()
-	ctx := context.TODO()
-	controller := newdnsController(ctx, client, dnsControlOpts{
-		initEndpointsCache: true,
-		useEndpointSlices:  true,
-		// This is needed as otherwise the fake k8s client doesn't work properly.
-		skipAPIObjectsCleanup: true,
-	})
-
-	durationSinceFunc = func(t time.Time) time.Duration {
```
chrisohaver, line 63: Probably should remove `durationSinceFunc` too (from `metrics.go`), since it was only an overridable func for the purpose of these tests.

chrisohaver:
> > ... perhaps we should add the e2e tests in coredns/ci before removing the coredns/coredns tests so there is no coverage gap.
> 
> I'm fiddling with this today.

Well, about that.  I'm _probably_ (hopefully) missing something, but I'm manually testing on a CoreDNS 1.7.0 running in a k8s cluster, and I cannot seem to get any latency metrics to actually appear.  I verified that there are endpoints created/updated with the annotation that the metric requires, and that CoreDNS answers with records for those endpoints, but scraping metrics from CoreDNS does not yield any `coredns_kubernetes_dns_programming_duration_seconds` entries.

chrisohaver:
> I cannot seem to get any latency metrics to actually appear

OK. I have confirmed that latency metrics are only recorded when skip cleanup is set to true.  IOW, the feature doesn't work in production.  Not sure if it has always been this way from the outset, or if some refactoring broke things (e.g. refactoring of controller with tombstone deletion fixes).

chrisohaver:
I've got a fix (#4255) that looks like it's working. It moves the "cleanup" step from the `toFuncs` to _after_ the latency calc in the `defaultProcessor`.  ~It does require a new "cleanup" function parameter to be passed to the default processor.~    In this way, the existing unit tests should be able to work with production code (i.e the clean up happens in both tests and produciton).

chrisohaver:
object.To() function bool parameters removed in #4255

PR #4255, opened 2020-11-03
plugin/kubernetes: Fix dns programming duration metric

chrisohaver:

Currently, the DNSProgrammingLatency metric does not produce any data.  This is because it relies on information which has been cleared by the time we calculate the latency and record the metric.  The unit tests for the metric pass, because they dubiously disable the clearing of the data during the test (skipCleanup bool).

This PR does the following

1. Fix the DNSProgrammingLatency metric by collecting required info from original object before clearing it out.
2. Remove skipCleanup boolean to simplify code
4. Correct the tests that relied on skipCleanup

 Closes: #4244  #4253


plugin/kubernetes/controller.go
```diff
@@ -173,6 +175,30 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
 	return &dns
 }
 
+func cleanEndpointSlice(i interface{}) {
+	if e, ok := i.(*discovery.EndpointSlice); ok {
+		*e = discovery.EndpointSlice{}
+	}
+}
```
bradbeam: I think this would be functionally the same, but is it worth making a `clean` function that does a type switch? That'd keep the informer.DefaultProcessor function parameters unchanged.

```
func cleanup(I interface{}) {
  switch item := i.(type) {
  case *api.Endpoints:
    *item = api.Endpoints{}
  case *api.Service:
    *item = api.Service{}
  ...
  }
}
```

bradbeam (approved):

plugin/kubernetes/controller.go
chrisohaver: Yes, thats better.

chrisohaver: done, thanks!

chrisohaver:
... a data race.  I'm guessing _this_ might have been the specific "problem" the original author was averting with the "skipCleanup" bool.  I wonder if this is really just an issue with "fakeClient" or if setting the objects to empty structs is inherently unsafe.  If we can confidently know that it's just an issue with the "fakeClient" we can exclude this test from race testing.  But seems reasonable that it could be an issue with a real client.

chrisohaver:
> But seems reasonable that it could be an issue with a real client.

I think the issue is that we are adding the object ourselves in the test.  If we ever had CoreDNS actually add a service pod or endpoint, this might be a real issue, but we do not.  If thats correct, I think we can safely skip race detection here.

chrisohaver:
I've refactored the latency unit test to not use the fakeClient.

plugin/kubernetes/object/informer.go
```diff
@@ -68,10 +70,25 @@ func DefaultProcessor(convert ToFunc, recordLatency RecordLatencyFunc) Processor
...
 					}
 				}
+				cleanObj(d.Object)
 			}
 			return nil
 		}
 	}
 }
 
+func cleanObj(i interface{}) {
```
miekg: I find it way more obvious if the endpoint cleanup happens in the To functions, this is too hidden

plugin/kubernetes/object/service.go
```diff
@@ -29,17 +29,11 @@ type Service struct {
...
 
 // ToService returns a function that converts an api.Service to a *Service.
-func ToService(skipCleanup bool) ToFunc {
-	return func(obj interface{}) (interface{}, error) {
-		svc, ok := obj.(*api.Service)
-		if !ok {
-			return nil, fmt.Errorf("unexpected object %v", obj)
-		}
-		return toService(skipCleanup, svc), nil
+func ToService(obj interface{}) (interface{}, error) {
```
miekg: we should not facilitate the use of `interface{}` any more than needed. It with my previous comment to keep the `nil`ing this can be kept as-is.

miekg:
I really don't care about this metrics tests. I care about clean(er) code in this plugin. If (even) the reworked tests, still needs tweaks in the code used for talking to k8s it should just go (as I did in my PR). Moving to circle-ci seems to be the best option.

plugin/kubernetes/object/informer.go
chrisohaver: If it happens in the To functions, the data is erased before it is needed when calculating latency after processing is done.  Currently, this happens, and it causes the latency metric to never report anything.

chrisohaver:
> I really don't care about this metrics tests. I care about clean(er) code in this plugin. If (even) the reworked tests, still needs tweaks in the code used for talking to k8s it should just go (as I did in my PR). Moving to circle-ci seems to be the best option.

Yeah.  Although this PR also fixes the latency metric, which does not work in production.

plugin/kubernetes/object/service.go
chrisohaver: > If (even) the reworked tests, still needs tweaks in the code used for talking to k8s it should just go

The tweaks in the production code in this PR are only done to remove skipCleanup boolean, and to fix the latency feature, which currently doesn't work in reality (because it relies on skipCleanup = true ).

chrisohaver:
Apologies, I really didn't make it clear at all that I was primarily fixing the latency reporting metric in this PR, and then adapting the tests to be able to work with the fix.

The 3 things done in this PR...
1. Fix endpoint object latency metrics (previously the metric was not reported) - this entailed moving the cleanup to after the latency calculation.
2. remove the skipCleanup boolean - resulting in always cleaning up, but this breaks the unit test, which relies on skip cleanup...
3. refactor unit test to work without skip cleanup.  Ultimately i had to ditch fakeclient, because the object clean up causes a race condition when the test is acting as both api client and server.

plugin/kubernetes/object/informer.go
miekg: no. cleaning up in the To functions is where this code belong, now it becomes hidden and we can only hope the switch will catch all different types

miekg:
nacking this.
This should be tested in some e2e fashion, not but again making code less readable esp, in an area where we need to jump to a lot of hoops to get where we are.

chrisohaver:
OK.  How about we just fix the metric in this PR, and drop all the metric related unit tests.

chrisohaver:
OK, I have refactored things to move the cleanup step back into the toFuncs.

I replaced the latencyFunc with a LatencyRecorder, which has an _init_ separate from a _record_ function, to allow the Processor to get the trigger timestamp and parent services before calling the toFuncs, and write the metric after calling toFuncs and updating index.

I left the unit tests in for now.   Since they test production code, I think they have value, but I will remove them if you insist.  None of the code changes here are made to assist or enable the unit tests.  The tests are _unit_ tests, i.e. they don't  do a full e2e test, so adding e2e tests are still useful from a e2e standpoint.

chrisohaver:
@miekg, PTAL.  I've taken your advice and moved the cleanup back into the ToFuncs, which has resulted in easier to read code.

chrisohaver:
The DNS programming latency metric tests in coredns/ci pass with this PR.

chrisohaver:
@miekg, PTAL. Per your request, I have ...
* removed the latency metric unit tests
* changed the ToFuncs and ToFunc type to not use empty interfaces

chrisohaver:
pinging plugin/kubernetes owners: @bradbeam @johnbelamaric @miekg @rajansandeep @yongtang @zouyee

zouyee (approved):

Issue #4220, opened 2020-10-20
add *minimal* plugin


miekg:
**What would you like to be added**:
Add a new plugin that minimizes responses as describe (e.g.) in the BIND manual
~~~
minimal-responses

 minimal-responses yes | no ;
 minimal-responses yes ;

If yes the server will only add records to the authority and additional data sections when they are required e.g. delegations, negative responses. This may improve the performance of the server. The BIND default is no. This statement may be used in a view or a global options clause.
~~~

**Why is this needed**:
This is another way to help UDP fragmentation. Unlike BIND this won't be a speed up because this plugin will only strip records that have been added by other plugins.

chrisohaver:
IMO this would be a good "first plugin" for a contributor who hasn't written a plugin yet, and as looking for such an opportunity.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] add *minimal*..." ]
>IMO this would be a good "first plugin" for a contributor who hasn't written a
>plugin yet, and as looking for such an opportunity.

agreed

bharath-12345:
@miekg would like to pick this up. Please do assign me to this.

bharath-123:
@miekg  and all. I would like to pick this issue up. Please assign me to this.

bharath-123:
I am new to coredns. I will take some time to do my research to understand the plugin requirements and start implementing it.

bharath-123:
Any references or suggestions as to how to begin would be really useful :)

chrisohaver:
There is https://coredns.io/2016/12/19/writing-plugins-for-coredns/, which is quite old but still mostly relevant.
For this kind of plugin, one that is modifying the results produced by other plugins, you should use a wrapper around the responsewriter that stores the response instead of writing to the client when calling the next plugin. When you get a result fro the next plugin, you can then strip the unnecessary records response and write to the client.  A great example of this comes from the Learning CoreDNS O'Reilly Book, the _onlyone_ plugin: https://github.com/coredns/learning-coredns/blob/master/plugins/onlyone/onlyone.go.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] add *minimal*..." ]
>There is https://coredns.io/2016/12/19/writing-plugins-for-coredns/, which is
>quite old but still mostly relevant.
>For this kind of plugin, one that is modifying the results produced by other
>plugins, you should use a wrapper around the responsewriter that stores the
[...]

PR #4257, opened 2020-11-04
[WIP] Add minimal plugin

bharath-123:

This PR adds the minimal plugin. It minimizes the size of the dns query response returned by coredns.

This PR was suggest by @miekg in issue https://github.com/coredns/coredns/issues/4220

A README.md file has been added to the plugin directory

No

I have written up the README.md indicating the implemenation and design of the minimal plugin. I am currently working on tests. Working on getting some good tests to write for this plugin. 

I feel that this plugin should be at the end of the plugin chain. Since it is stripping away the records right before the client receives it.

Looking forward to some running feedback on this plugin.

## Design and Implementation details

If minimal is set as yes, this plugin wraps a response writer around dns.ResponseWriter and passes it on the to the next plugin 
in the plugin chain. If minimal is explicitly set as no, then this plugin will not be added at all. This should be better than 
making this plugin a passthrough plugin

This response writer implements the dns.ResponseWriter interface. 

The writer first checks the type of the dns.Msg. If the response type is any type apart from NoError, we just return the response back to the 
client. We even return back any Delegations because delegations sometimes contain glue records which are essential for dns resolution.

If the response type is NoError, then we strip away the ADDITIONAL and AUTHORITY sections away and ensure that only the answer section is
present in the dns response which is essential to complete the dns query succesfully.

Pseudo code:

```
type Minimal struct {
  Next plugin.Handler
}

func (m Minimal) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
	mm := &MinimalResponseWriter{w}
	return plugin.NextOrFailure(rr.Name(), mm.Next, ctx, mm, r)
}

type MinimalResponseWriter struct {
  dns.ResponseWriter
} 

func (m *MinimalResponseWriter) WriteMsg(res *dns.Msg) error {
  type := response.Typify(res)

  if type != response.NoError {
    m.ResponseWriter.WriteMsg(m)
  }

  // will such a case ever arise?????
  if len(res.Answer) == 0 {
    return m.ResponseWriter.WriteMsg(m)
  }

  // set the AUTHORITY and ADDITIONAL record to nil. Empty the records.
  res.Ns = nil
  res.Extra = nil

  return m.ResponseWriter.WriteMsg(m)
}

```

### DIFF
--- a/plugin.cfg
+++ b/plugin.cfg
@@ -64,3 +64,4 @@ erratic:erratic
 whoami:whoami
 on:github.com/coredns/caddy/onevent
 sign:sign
+minimal:minimal
--- a/plugin/minimal/README.md
+++ b/plugin/minimal/README.md
@@ -39,11 +39,20 @@ present in the dns response which is essential to complete the dns query succesf
 
 Pseudo code:
 
-type MinimalResponseWriter struct {
-  dns.ResponseWriter
+type Minimal struct {
+  Next plugin.Handler
 }
 
-func (m *MinimalResponseWriter) minimalMsg(res *dns.Msg) error {
+func (m Minimal) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	mm := &MinimalResponseWriter{w}
+	return plugin.NextOrFailure(rr.Name(), mm.Next, ctx, mm, r)
+}
+
+type MinimalResponseWriter struct {
+  dns.ResponseWriter
+} 
+
+func (m *MinimalResponseWriter) WriteMsg(res *dns.Msg) error {
   type := response.Typify(res)
 
   if type != response.NoError {
--- a/plugin/minimal/README.md
+++ b/plugin/minimal/README.md
@@ -1,0 +1,65 @@
+# minimal
+
+## Name
+
+*minimal* - minimizes the size of the DNS query response 
+
+## Description
+
+The minimal plugin controls the addition of records to the AUTHORITY and ADDITIONAL section of the dns query response.
+
+## Syntax
+
+~~~
+minimal yes | no
+~~~
+
+## Examples
+
+~~~ corefile
+. {
+    minimal yes
+    forward . 8.8.8.8 
+}
+~~~
+
+## Design and Implementation details
+
+If minimal is set as yes, this plugin wraps a response writer around dns.ResponseWriter and passes it on the to the next plugin 
+in the plugin chain. If minimal is explicitly set as no, then this plugin will not be added at all. This should be better than 
+making this plugin a passthrough plugin
+
+This response writer implements the dns.ResponseWriter interface. 
+
+The writer first checks the type of the dns.Msg. If the response type is any type apart from NoError, we just return the response back to the 
+client. We even return back any Delegations because delegations sometimes contain glue records which are essential for dns resolution.
+
+If the response type is NoError, then we strip away the ADDITIONAL and AUTHORITY sections away and ensure that only the answer section is
+present in the dns response which is essential to complete the dns query succesfully.
+
+Pseudo code:
+
+type MinimalResponseWriter struct {
+  dns.ResponseWriter
+}
+
+func (m *MinimalResponseWriter) minimalMsg(res *dns.Msg) error {
+  type := response.Typify(res)
+
+  if type != response.NoError {
+    m.ResponseWriter.WriteMsg(m)
+  }
+
+  // will such a case ever arise?????
+  if len(res.Answer) == 0 {
+    return m.ResponseWriter.WriteMsg(m)
+  }
+
+  // set the AUTHORITY and ADDITIONAL record to nil. Empty the records.
+  res.Ns = nil
+  res.Extra = nil
+
+  return m.ResponseWriter.WriteMsg(m)
+}
+
+Tests: 


bharath-123:
Beginners question: What would the type of the following response be in coredns?
```
; <<>> DiG 9.17.6 <<>> -p 31488 -t NS rt.example @10.53.0.1
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 29595
;; flags: qr aa rd; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 2

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: 114aa6fc5b6c9f64010000005fa2ed44898c47b0bf66d211 (good)
;; QUESTION SECTION:
;rt.example.			IN	NS

;; ANSWER SECTION:
rt.example.		86400	IN	NS	ns1.rt2.example.
rt.example.		86400	IN	NS	ns1.rt.example.

;; ADDITIONAL SECTION:
ns1.rt.example.		86400	IN	A	10.53.0.1

;; Query time: 0 msec
;; SERVER: 10.53.0.1#31488(10.53.0.1)
;; WHEN: Wed Nov 04 23:34:52 IST 2020
;; MSG SIZE  rcvd: 130
```

This should be marked as a delegation right? ^

bharath-123:
For tests, I am using the tests in bind9. They do multiple dig calls to the local nameservers setup by them. Fortunately they store the request response in files after the tests are done. They give a really good idea for the expected output with and without minimal.

plugin/minimal/README.md
```diff
@@ -0,0 +1,74 @@
+# minimal
+
+## Name
+
+*minimal* - minimizes the size of the DNS query response
```
miekg, line 5:
```suggestion
*minimal* - minimizes the size of the DNS response message.
```

plugin/minimal/README.md
```diff
@@ -0,0 +1,74 @@
+# minimal
+
+## Name
+
+*minimal* - minimizes the size of the DNS query response 
+
+## Description
+
+The minimal plugin controls the addition of records to the AUTHORITY and ADDITIONAL section of the dns query response.
```
miekg, line 9:
```suggestion
The *minimal* plugin tries the minimize the size of the response. Depending the response type it removes resource records from the  AUTHORITY and ADDITIONAL sections.
```

plugin/minimal/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Syntax
+
+~~~
+minimal yes | no
```
miekg, line 14:
```suggestion
minimal
```

plugin/minimal/README.md
```diff
@@ -0,0 +1,74 @@
...
+## Examples
+
+~~~ corefile
+. {
+    minimal yes
+    forward . 8.8.8.8 
+}
+~~~
+
+## Design and Implementation details
```
miekg, line 26: entire section can be removed - this does not need to be in the readme. Add it to the PR or as a comment on the issue so we can review/discuss.

plugin/minimal/README.md
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  // set the AUTHORITY and ADDITIONAL record to nil. Empty the records.
+  res.Ns = nil
+  res.Extra = nil
+
+  return m.ResponseWriter.WriteMsg(m)
+}
+
+Tests:
```
miekg, line 74: Add a "See Also" section and reference the bind documentation for this feature.

plugin/minimal/README.md
chrisohaver:
```suggestion
*minimal* - minimizes the number of records in response messages.
```

plugin.cfg
```diff
@@ -64,3 +64,4 @@ erratic:erratic
 whoami:whoami
 on:github.com/coredns/caddy/onevent
 sign:sign
+minimal:minimal
```
chrisohaver, line 67: We'd want to cache minimal responses (so it needs to be after cache), but minimal will want to minimize all responses created by other plugins (so it needs to go before them). 

So, i'm thinking right before template, the first plugin in the chain that creates responses.

plugin/minimal/README.md
bharath-123: Ack these suggestions. They make the README.md look better. Removing the yes | no is a better choice too. Will incorporate them.

plugin/minimal/README.md
bharath-123: Yes. I intended to add that for review and discussion. I planned to remove that for the final PR. I shall seperate out the design and implementation sections.

plugin/minimal/README.md
bharath-123: Yes

plugin.cfg
bharath-123: My thinking was that it should be the last plugin to be processed since it is potentially stripping off certain records from the response. Caching makes sense here. 

Ahh I think I missed out the fact that minimal is wrapping up a response writer. This makes sense! Thank you for the insights!

bharath-123:
I have seperated out the design and implementation details. Would be great if you all can review that. I am currently working on the test cases. I am following the test cases written in the load balancer plugin. I feel the test cases for minimal will be somewhat similar.

chrisohaver:
```
  if type != response.NoError {
    m.ResponseWriter.WriteMsg(m)
  }

  // will such a case ever arise?????
  if len(res.Answer) == 0 {
    return m.ResponseWriter.WriteMsg(m)
  }
```
>   // will such a case ever arise?????

Yes, for NODATA responses (no records of requested type for a name that has records of other types).

chrisohaver:
The proposed design looks ok to me. It might require _slightly_ less code to use the pre-existing `nonwriter` (like in the onlyone example) instead of making a new ResponseWriter wrapper.  But I think it's fine either way.

bharath-123:
Will look into this. I ll start coding and raise a full fledged PR with tests.

plugin/minimal/README.md
bl-ue: @miekg updated suggestion:
```suggestion
The *minimal* plugin tries to minimize the size of the response. Depending the response type it removes resource records from the AUTHORITY and ADDITIONAL sections.
```
There were two spaces before AUTHORITY in your suggestion @miekg and I changes `tries the minimize` -> `tries to minimize` :)

miekg:
where do we stand here?

bharath-12345:
hey. I wasnt able to work on this due to some personal issues which cropped up. I apologize for this. I can pick this up sometime in the future or we can let someone else work on this.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] [WIP] Add min..." ]
>hey. I wasnt able to work on this due to some personal issues which cropped up.
>I apologize for this. I can pick this up sometime in the future or we can let
>someone else work on this.

there is no rush, but having something mergeable at some point in the future would be nice
:) As long as it is explicit if you're working on it or not.

miekg:
see #4380

PR #4267, opened 2020-11-06
Remove travis and move to github workflow

### LINKED ISSUE #4266: Deprecate Travis and move to github workflows
Travis is not working well anymore. Let's move our tests into a github workflow.

This means:
* adding a workflow into .github/workflows for tests
* copying the relavent bits from .travis.yml into it
* Running the multiple tests in parallel (if possible), this is `TEST_TYPE`
* See if the Makefile needs tweaking

miekg:
Add github testing workflow, simplify the Makefile because that was
complex because of Travis. Remove the fuzzing, needs to be re-added when
that works properly with go modules (it has been disabled for quite some
time)

Closes: #4266


miekg:
some unification of the go version make might sense, it's specified 4 times in two different files, but this can be done in follow up PRs

chrisohaver (commented):
Are the `fmt` and `metrics` tests intentionally dropped?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Remove travis..." ]
>
>Are the fmt and metrics tests intentionally dropped?

from the commit: "Move metric's naming test to test/presubmit_test.go"

I think moving the fmt to a tidy action is better, I've never seen it complain, even
though fmt-ing current master shows a diff... ?

/Miek

--
Miek Gieben

chrisohaver (approved):

yongtang:
@miekg Wondering if it makes sense to move `TestTrailingWhitespace` to Tidy Actions as well?

Since `go fmt` will automatically remove trailing whitespaces from `.go` files, we only need to remove trailing whitespaces from non-go text files (README.md, etc). I can help update the tidy actions if you think this makes sense.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Remove travis..." ]
>@miekg Wondering if it makes sense to move TestTrailingWhitespace to Tidy
>Actions as well?
>
>Since go fmt will automatically remove trailing whitespaces from .go files, we
>only need to remove trailing whitespaces from non-go text files (README.md,
>etc). I can help update the tidy actions if you think this makes sense.

I think that makes sense, annoy folks less, but still do the right thing.

PR #4277, opened 2020-11-10
Fix missing `*` in Go Fmt GitHub Actions

yongtang:
Looks like one `*` is missing in Go Fmt GitHub Actions
and that causes the Actions not being triggered.

### DIFF
--- a/.github/workflows/go.fmt.yml
+++ b/.github/workflows/go.fmt.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - 'master'
     paths:
-      - '*.go'
+      - '**.go'
 
 jobs:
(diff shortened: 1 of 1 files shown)


miekg:
/lgtm

In zsh this is valid, in actions as well I assume?

yongtang:
@miekg Ah this is the syntax from GitHub Actions: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

yongtang:
This GitHub Action has been triggered in 4b87be2  and it looks like it works correctly as well 🎉

PR #4279, opened 2020-11-11
Remove newlines between braces

miekg:
These are found with: `pcregrep -M "}\n\n}" **/*.go`
Sometimes a unneeded newline is inserted, remove those.


miekg:
merging, it's a trivial pr

Issue #4271, opened 2020-11-08
Segfault with cache running again itself


owengo:
**What happened**:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xb1f392]

goroutine 7095 [running]:
github.com/coredns/coredns/plugin/cache.(*ResponseWriter).LocalAddr(0xc000c81200, 0xc0, 0xc0)
        <autogenerated>:1 +0x32
github.com/coredns/coredns/plugin/cache.(*Cache).ServeDNS.func1(0x1, 0xc001000f30, 0xc0018fb4d0, 0xc00007f200, 0xc000e177e5, 0x9, 0x2077de0, 0xc0017b85c0, 0x2064f20, 0xc001333740)
        /tmp/tmp.yM3x7N1L0w/g/src/github.com/coredns/coredns/plugin/cache/handler.go:56 +0x4b
created by github.com/coredns/coredns/plugin/cache.(*Cache).ServeDNS
        /tmp/tmp.yM3x7N1L0w/g/src/github.com/coredns/coredns/plugin/cache/handler.go:52 +0x694

```

**What you expected to happen**:

no segfault

**How to reproduce it (as minimally and precisely as possible)**:

I think the cause is the mix of rewrite + cache. I had the issue with 1.7.0 eksbuild.1 and 1.8.0 from dockerhub.
The configuration is  quite big but here are the relevant parts:
```

apiVersion: v1
data:
  Corefile: |
    .:53 {
        errors
        health
        rewrite continue {
            name  regex (.*)\.(list-of-private-suffixes|amazonaws.com)\.cluster\.local {1}.{2}
            answer name (.*)\.(list-of-private-suffixes|amazonaws.com) {1}.{2}.cluster.local
        }
        rewrite stop {
            type AAAA A
        }
        kubernetes cluster.local {
          pods insecure
          fallthrough in-addr.arpa ip6.arpa
        }
        prometheus :9153
        forward . /etc/resolv.conf
        cache {
          prefetch 1
          success 1000 900
          denial 1000 60 30
          prefetch 1
          serve_stale
        }
        reload
        log
        loadbalance
        file /etc/node-local-zones/private-zone-file private-tld {
          reload 10s
        }
...

```

**Anything else we need to know?**:

Here the last lines of log before the crash:

```
[...]

PR #4281, opened 2020-11-11
Move .LocalAddr() out of goroutine

miekg:
I don't have theory why this panics, but concurrency has something to do
with it, so get the address before we call the goroutine.

See #4271


chrisohaver (approved):
lgtm

PR #4282, opened 2020-11-11
auto test: increase sleep

miekg:
Doing a reload every 10ms and sleeping 10ms makes this too racy,
increase all those sleeps to 50ms. A better method would be to get
notified of the reload/change, this would require some polling to get
the RR we expect, with a much longer timeout to stop the test
eventually.

### DIFF
--- a/test/auto_test.go
+++ b/test/auto_test.go
@@ -46,7 +46,7 @@ func TestAuto(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	time.Sleep(10 * time.Millisecond) // wait for it to be picked up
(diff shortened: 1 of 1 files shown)


miekg:
@stp-ip you can't merge without an lgtm, see https://github.com/miekg/dreck/blob/master/commands.md
(this could be changed of course...?)

stp-ip:
Damn. Too many different variations in various projects of similar commands. Sometimes I mess them up in my head. Was about to look at the commands file as it wasn't actuated yet :P

miekg:
getting that table right took _a lot of_ staring at the markdown

Issue #4285, opened 2020-11-12
plugin/azure - server fails to send response when querying AAAA of Azure private DNS zone 


mancaus:
**What happened**:

With CoreDNS configured with the azure plugin, pointing at an Azure Private DNS Zone. Querying AAAA for for a domain with an A record in that zone:
```
dig @127.0.0.1 -p 54545 +tries=1 <domain> AAAA

; <<>> DiG 9.11.5-P4-5.1ubuntu2.2-Ubuntu <<>> @127.0.0.1 -p 54545 +tries=1 <domain> AAAA
; (1 server found)
;; global options: +cmd
;; connection timed out; no servers could be reached
```

Logs from server:
```
./coredns -dns.port 54545
azdevint.dh.:54545
[INFO] plugin/reload: Running configuration MD5 = 91036d7b12759772ac5d002ada9c41fe
CoreDNS-1.8.0
linux/amd64, go1.15.4,
[INFO] 127.0.0.1:47245 - 28160 "AAAA IN shop-job-adapter.sxdb-blake-dev-neu.azdevint.dh. udp 88 false 4096" NOERROR qr,aa,rd 164 0.000053112s
```

**What you expected to happen**:

Server to respond with empty records and an SOA for the zone.

**How to reproduce it (as minimally and precisely as possible)**:

Per above, create an Azure Private Zone containing a single A record. Configure azure plugin with that zone. Query for AAAA.

**Anything else we need to know?**:

I'm creating an MR. The plugin does not check the return code of dns.ResponseWriter WriteMsg(m) in Serve(...). Adding that check shows the SOA host / ns is not a FQDN. Passing that throuth dns.Fqdn(...) when reading private zone records resolves the issue. The issue may be upstream, in that the the source of the hostname that is not an FQDN is the Azure-supplied SOA record for the private zone.

**Environment**:

- the version of CoreDNS: Built from master
- Corefile:   
- logs, if applicable: See above
- OS (e.g: `cat /etc/os-release`): Ubunto 19.10 amd64
- Others:

PR #4286, opened 2020-11-12
plugin/azure return FQDN as MNAME in SOA record

mancaus:

See #4285 - adds error handling for failure to send DNS response and ensures SOA host is an FQDN, as Azure private zones return a DNS name here currently.

#4285 

None

No

### DIFF
--- a/plugin/azure/azure.go
+++ b/plugin/azure/azure.go
@@ -286,7 +286,7 @@ func updateZoneFromPrivateResourceSet(recordSet privatedns.RecordSetListResultPa
 				Refresh: uint32(*(SOA.RefreshTime)),
 				Serial:  uint32(*(SOA.SerialNumber)),
 				Mbox:    dns.Fqdn(*(SOA.Email)),
-				Ns:      *(SOA.Host)}
+				Ns:      dns.Fqdn(*(SOA.Host))}
 			newZ.Insert(soa)
 		}
 


plugin/azure/azure.go
```diff
@@ -346,7 +346,14 @@ func (h *Azure) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 		return dns.RcodeServerFailure, nil
 	}
 
-	w.WriteMsg(m)
+	err := w.WriteMsg(m)
```
miekg: we don't really care if this fails, there's nothing what we can do about, logging it will allows for easy DoS-ing of whatever these logs end up in. Please remove.

original change lgtm

mancaus: Thanks, and good point - will do. Just to check - should we log as Debug? I only ask because the logs that were written gave the impression that a response was returned to the client, when it wasn't.

I'm happy either way and will adjust the PR once you can confirm.

miekg: hmm, true point. We can save the err and return it from the plugin, then it will be logged by errors plugin, but that would be a system wide change. 
Apart from we can't pack a message because of an issue like this, failing a write to a client is not important. We _could_ check for dns: type errors and only log those...

mancaus: I think what you propose makes sense - just return any error directly without logging. I just tried that and, as you say, it causes the errors plugin to pick it up and I see a reasonable log entry:
```
[ERROR] plugin/errors: 2 shop-job-adapter.sxdb-blake-dev-neu.azdevint.dh. AAAA: dns: domain must be fully qualified 
```

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] plugin/azure ..." ]
>@mancaus commented on this pull request.
>
>
>In plugin/azure/azure.go:
>
>> @@ -346,7 +346,14 @@ func (h *Azure) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
>                return dns.RcodeServerFailure, nil
>        }
>
>-       w.WriteMsg(m)
>+       err := w.WriteMsg(m)
>
>I think what you propose makes sense - just return any error directly without
>logging. I just tried that and, as you say, it causes the errors plugin to pick
>it up and I see a reasonable log entry:
>
>[ERROR] plugin/errors: 2 shop-job-adapter.sxdb-blake-dev-neu.azdevint.dh. AAAA: dns: domain must be fully qualified

yep. But this is an exception as this is a coding error in coredns, generally we don't
care about the WriteMsg error, this with rogue clients this becomes usefully wasteful.

An no matter what: this needs to be done in a different PR/issue anyway (and I'm not a
proponent of it for the reasons explained here)

/Miek

--
Miek Gieben

mancaus:
Thanks @miekg - this latest change to the PR reverts the error checking, so the change is back to the one-line fix for the issue itself.

Issue #4284, opened 2020-11-11
SRV Record Target names not rewritten by Rewrite Plugin


slick-nic:
**What happened**:
We use Kubernetes/Calico CNI with headless services and the coredns rewrite plugin to allow us to resolve addresses outside of the cluster.

This works fine for A records, but SRV Target domains in the answers for SRV records are not rewritten.

The A records in the Additional section of the SRV lookup response are also not rewritten.

See attached file for A + SRV dig lookup examples.
[Dig Output.txt](https://github.com/coredns/coredns/files/5525815/Dig.Output.txt)

**What you expected to happen**:
I expect rewrite to work for SRV Target.
Looking at the code, this probably also applies to other record types that return a domain name in the answer e.g. MX

**How to reproduce it (as minimally and precisely as possible)**:
Set up a headless service in kubernetes (with the config required to get headless service SRV records created)
Set up a rewrite from cluster.local to/from another.domain.uk
Resolve the SRV records for the rewritten domain, observe that the response still contains cluster.local

**Anything else we need to know?**:
I have written a unit test and a fix which I think identifies/fixes this issue _for me_, but I'm new to Go, I dont think this is the finished product in terms of a fix more of a starting point to discuss. Comments are welcome. I still have not run it in a K8s cluster to test fully.
[srv_revert.patch.txt](https://github.com/coredns/coredns/files/5525823/srv_revert.patch.txt)

**Environment**:
Kubernetes 1.19 / Calico CNI / CRI-O on Ubuntu Server 20.04
Core DNS 1.7.0 from k8s.gcr.io/coredns:1.7.0
YAML to update corefile in Kubernetes is attached.
[CoreFileUpdate.yaml.txt](https://github.com/coredns/coredns/files/5525811/CoreFileUpdate.yaml.txt)

Apologies for yaml / patch files as .txt - github doesnt accept patch or yaml files.

chrisohaver:
FWIW, This has been discussed before in https://github.com/coredns/coredns/issues/2770. I knew it seemed familiar, but took me a while to find it.

chrisohaver:
Does the current behavior cause a practical problem for SRV responses?
Currently neither the target or the addtl section are rewritten, so they match up.
If a client decides to ignore the additonal section records, and query the target names, they'd get the proper responses.

i.e.  is this fixing something that doesn't work now, or is it cosmetic in nature.

slick-nic:
> Does the current behavior cause a practical problem for SRV responses?
[...]

PR #4287, opened 2020-11-12
Rewrite SRV targets and additional names in response

slick-nic:
Signed-off-by: Nic Colledge <nic@njcolledge.net>

~~Fixes rewrite plugin so that it rewrites SRV targets and additional record names in the DNS response see #4284~~
~~Could also be useful to go further and add other record types which return names (e.g. MX)~~
The rewrite plugin successfully rewrites the names returned in the ANSWER section so they match the names in the request. However, it does not currently rewrite other names returned in record types such as SRV and MX. This PR adds a “answer value” configuration option to the regex rewrite rule to enable this functionality. It currently works on values returned by SRV, MX, CNAME, NS, DNAME, NAPTR and SOA records. More can be added if required.
Based on feedback on an earlier version of the PR, the rewrite functionality has also been expanded to consider records in the NS and EXTRA sections in addition ANSWER section.

#4284
~~None unless we add an option to enable/disable this functionality.~~
Added content to README.md in response to comment from @chrisohaver 
~~I dont think so, but happy to discuss adding an option to disable/enable it.~~
No, existing configs should not include the new option so there should be no impact.


plugin/rewrite/reverter.go
```diff
@@ -40,45 +40,71 @@ func (r *ResponseReverter) WriteMsg(res *dns.Msg) error {
...
+	}
+	if isTTLRewritten {
+		rr.Header().Ttl = ttl
+	}
+	if isSrvTargetRewritten {
+		rr.(*dns.SRV).Target = srvTarget
+	}
+}
+
+func rewriteName(rule ResponseRule, name *string, isNameRewritten *bool) {
```
chrisohaver: I think it would be more idiomatic for Go, written in this way ...

```
func rewriteName(rule ResponseRule, name string) (rewrittenName string, isNameRewritten bool) {
```

chrisohaver:
I think this new behavior should get a mention in the README.

plugin/rewrite/reverter.go
slick-nic: Yeah, I actually started with it the way you suggest, but switched to pointers to simplify the calling-code because of the way the isNameRewritten/isSrvTargetRewritten boolean works inside the existing loop (line 67) it saves a few lines of code for each call.
If my memory is correct each call would need an additional if and/or |= to retain the current behaviour of the loop/flag(s).
Im new to Go so I'll take your lead and chage it if you prefer the function signature in the more idiomatic way.

chrisohaver: I'll defer to the maintainers of this plugin.

plugin/rewrite/README.md
```diff
@@ -151,6 +151,62 @@ ftp-us-west-1.coredns.rocks. 0    IN A    10.20.20.20
 ftp-us-west-1.coredns.rocks. 0    IN A    10.30.30.30
 ```
 
+Note that names in SRV Record Targets and records in the `ADDITIONAL SECTION` 
+will also be rewritten following the `answer name` rule to keep the response
```
chrisohaver: > and records in the `ADDITIONAL SECTION` will also be rewritten

but not always. Only for SRV requests (for now)...

"For SRV responses, SRV targets and records in the `ADDITIONAL SECTION`  will also be rewritten ..."

I don't think the following examples are necessary in context with the prior examples, it should be clear what this means.

chrisohaver: hmm - i just re-read the code - and it looks like it may _always_ be rewriting the additional section?  If so, that will break some responses where targets no longer match the additional section entries.

chrisohaver:
It might be better to make this consistent for all types/sections, so everything lines up.  Authority, Answer, and Additional, and target names in all record types that have them ... SRV, CNAME, MX, NS, others? ... primary ns in the SOA record?

plugin/rewrite/README.md
slick-nic: Agreed the wording here is difficult.

But I dont think we can limit the description to just "For SRV responses..." consider the following answer/additional sections:

```
;; ANSWER SECTION:
_test._tcp.test-service.cluster.local. 15 IN SRV 0 50 1234 10-1-1-2.test-service.cluster.local.

;; ADDITIONAL SECTION:
10-1-1-2.test-service.cluster.local. 15 IN A 10.1.1.2
```

The code in this PR will always rewrite the names on the "left hand side" of both the answer and additional sections regardless of if the additional section relates to a SRV answer or something else.
However, you are correct that for names on the "right hand side" that they will only be rewritten for SRV (for now).

I'll try and come up with something better (suggestions are welcome).

chrisohaver: > The code in this PR will always rewrite the names on the "left hand side" of both the answer and additional sections regardless of if the additional section relates to a SRV answer or something else.

OK.  IIUC, this will break some responses, causing named targets in non-SRV records not to match the entries in the additional section.

slick-nic:
@chrisohaver just pushed some changes to handle more record types and the authority section as discussed. I dont really like the two switch-statements, maybe theres a better way?
Not the finished article just yet need to do some more reading about NAPTR, DNAME and think more about if we really want to update the SOA record, I'll also add some more unit testing in future I just wanted to get feedback first.

chrisohaver:
@slick-nic, Thanks, I think it looks promising!  Before finishing it off though, let's get feedback from the code owners of the plugin to see if this is a feature they deem it appropriate to add.

slick-nic:
Just had a bit of spare time so updated a few unit tests related to this change.
No test added for NAPTR records yet, we should think about if we want to rewrite the regex-bit of these also.
SOA records have a similar question around rewriting the email address returned in them (currently not rewriting).

Also need to add tests for Authority section rewriting - its not really different but probably good to get the coverage.

As mentioned in a previous comment from @chrisohaver - it would be good to get feedback from plugin owners before continuing.

plugin/rewrite/README.md
slick-nic: Resolving this as we have changed direction based on the discussion below and in the issue thread.

slick-nic:
Hi there,
@chrisohaver @johnbelamaric 

Sorry to bump but I have some free time coming up and was wondering if there was anything I could do to move this forward?
As mentioned by Chris previously do we think its appropriate to include these changes?
Any other feedback?

Thanks,
Nic.

chrisohaver:
https://github.com/coredns/coredns/blob/master/CODEOWNERS#L46

cc @greenpau

greenpau:
> ci/circleci: kubernetes-tests — Your tests failed on CircleCI

@slick-nic , please check this failure.

> Fixes rewrite plugin so that it rewrites SRV targets and additional record names in the DNS response see #4284
Could also be useful to go further and add other record types which return names (e.g. MX)

Here is my understanding of the issue #4284. Provided the following config:

```
name regex (.*)\.my\.domain\.uk {1}.cluster.local
answer name (.*)\.cluster\.local {1}.my.domain.uk
```

You get the response where the the name rewrites correctly, but the values are not being rewritten. 

```
;; ANSWER SECTION:
_test._tcp.test-service.namespace.svc.my.domain.uk. 15 IN SRV 0 50 1234 10-11-109-126.test-service.namespace.svc.cluster.local.
_test._tcp.test-service.namespace.svc.my.domain.uk. 15 IN SRV 0 50 1234 10-11-219-6.test-service.namespace.svc.cluster.local.
```

And you want them to be rewritten following this schema: `10-11-109-126.test-service.namespace.svc.cluster.local.` to 10-11-109-126.test-service.namespace.svc.my.domain.uk.`.

@slick-nic , correct?

Your current solution involves adding `rewriteResourceRecord()` where you determine the value (which you call a target). 

@slick-nic, If so, why not create `answer value (.*)\.cluster\.local {1}.my.domain.uk`? I did not look deep into the function. Nevertheless, what are cons/pros of your approach? Is there another alternative?

greenpau:
@slick-nic , I suggest you put a log more explanation in the docs/PR about the mechanics of this, i.e. the elaboration on "You get the response where the the name rewrites correctly, but the values are not being rewritten." Specifically, I think this is a good idea adopt this. We just need more documentation, discussion, and code review.

Consider that the docs will be read by newbies and you want to over-inform them on the subject. 😃

greenpau:
@slick-nic , side note, I see your `njcolledge.net`. I am from NY areas. Please connect with me via LinkedIn, we can connect offline, and do your code review over google meet or something. Will help you out!

slick-nic:
> And you want them to be rewritten following this schema: `10-11-109-126.test-service.namespace.svc.cluster.local.` to 10-11-109-126.test-service.namespace.svc.my.domain.uk.`.
> 
> @slick-nic , correct?

@greenpau - Yes that is what I want.
 
> Your current solution involves adding `rewriteResourceRecord()` where you determine the value (which you call a target).
> 
> @slick-nic, If so, why not create `answer value (.*)\.cluster\.local {1}.my.domain.uk`? I did not look deep into the function. Nevertheless, what are cons/pros of your approach? Is there another alternative?

@greenpau - Yeah the adding a new "answer value" to the config is a nice idea.
When I started using coredns and kubernetes I expected rewrite to "just work" as described above (based on my understanding of the documentation at the time), so for simplicity the the initial PR did just that for SRV records.
I see this as a "pro" of my approach, and I dont think I'm the first to get burned by this (IIRC a previous similar issue was referenced by Chris earlier).
The "con" is the change messes someone else up who is relying on it not getting rewritten, but because these otherwise would have been .local addresses its _probably_ not that big of a deal.

P.S. the "target" naming comes from this initially being related to SRV record Targets only (so adopted the naming from them), but the change was later broadened to include other records so maybe calling it something else would be better.

slick-nic:
> @slick-nic , I suggest you put a log more explanation in the docs/PR about the mechanics of this.

@greenpau Agreed, the initial version of this PR was intended only as a starting-point for discussion rather than a finished solution :-)

> @slick-nic , side note, I see your...

I'm not on LinkedIn so we should find another way to connect, I joined the CNCF/Slack #coredns channel today if you are on there that could be a way. Despite the .net domain name I'm currently in the UK.

greenpau:
@slick-nic , you are still in draft. I reviewed the changes and it looks good. However, is there a way you could improve your coverage? Also, please address `TODO`.

cc: @johnbelamaric

slick-nic:
> 
> 
> @slick-nic , you are still in draft. I reviewed the changes and it looks good. However, is there a way you could improve your coverage? Also, please address `TODO`.
> 
> cc: @johnbelamaric

@greenpau Yeah I'm still in draft; Just wanted to see if the merge-in from the origin and the updates fixed the failing kubernetes-tests from before. I'll try and take a look at the TODO and coverage tomorrow. Thanks for the quick response.

greenpau:
LGTM 👍

greenpau:
@miekg , this looks good, but I don’t have right not approve/merge. Please assist.

miekg:
You have not? Will fix. In the mean time

/lgtm
/merge

slick-nic:
Apologies to all the people on the review list getting spammed about this :-) I have tried to edit the list to remove you but github wont let me.

greenpau:
@miekg , thank you! 👍

greenpau:
@miekg , it no longer works and I also don’t get invitations to review PRs.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Rewrite SRV t..." ]
>@miekg , it no longer works and I also don’t get invitations to review PRs.

just sent one

greenpau:
@miekg , 👍

PR #4303, opened 2020-11-20
plugin/file: Use NXDOMAIN response if CNAME target is NXDOMAIN

### LINKED ISSUE #4288: Non-existent CNAME target in the same zone should be returned with NXDOMAIN instead of NOERROR rcode
Hi,

This is related to if 'A CNAME B' exists in a zone, but B (related to the same zone) does not exist, then the return code should be NXDOMAIN.  [RFC 6604](https://tools.ietf.org/html/rfc6604#section-3) mentions that 
> When chains are followed the RCODE in the ultimate DNS response MUST BE set based on the final query cycle leading to that
>       response.

Consider the following sample zone file

|  	|  	|  	|  	|
|-	|-	|-	|-	|
| example.com. 	| 500 | SOA 	| ns1.outside.com. root.example.com. 3 604800 86400 2419200 604800 	|
| example.com. 	| 500 | NS 	| ns1.outside.com.  	|
| b.c.example.com. 	| 500 | CNAME | c.c.example.com.  	|
| c.c.example.com. 	| 500 | CNAME	| d.c.example.com.  	|

For the query _<b.c.example.com. , A>_, the Coredns server returns the following response:
```
          "opcode QUERY",
          "rcode NOERROR",
          "flags QR AA",
          ";QUESTION",
          "b.c.campus.edu. IN A",
          ";ANSWER",
          "b.c.example.com. 500 IN CNAME c.c.example.com.",
          "c.c.example.com. 500 IN CNAME d.c.example.com.",
          ";AUTHORITY",
          "example.com. 500 IN NS ns1.outside.com.",
          ";ADDITIONAL"
```

The expected response is the same as above except that the `rcode` should be `NXDOMAIN`. 

**Environment**:

- the version of CoreDNS: `CoreDNS-1.8.0  linux/amd64, go1.15.3, 054c9ae`
- Corefile:
```
example.com:53 {
    file db.example.com
    log
    errors
}
```
- OS: Ubuntu 18

chrisohaver:

As cited in  #4288, per https://tools.ietf.org/html/rfc6604#section-3, when following a CNAME chain the response code sent to the client should be the same response code received when resolving the target of the final link in the chain.  Strictly this means we should be passing through the exact Rcode received when looking up the final target. But since _file_ uses an alternate response code datatype `file.Result` which does not map 1:1 to DNS response codes, making that happen might require some larger scope refactoring.  So for now, just addressing response codes that do map.

fixes #4288

### DIFF
--- a/plugin/file/lookup.go
+++ b/plugin/file/lookup.go
@@ -307,8 +307,9 @@ func (z *Zone) externalLookup(ctx context.Context, state request.Request, elem *
 	targetName := rrs[0].(*dns.CNAME).Target
 	elem, _ = z.Tree.Search(targetName)
 	if elem == nil {
-		rrs = append(rrs, z.doLookup(ctx, state, targetName, qtype)...)
-		return rrs, z.Apex.ns(do), nil, Success
+		lookupRRs, result := z.doLookup(ctx, state, targetName, qtype)
+		rrs = append(rrs, lookupRRs...)
+		return rrs, z.Apex.ns(do), nil, result
 	}
 
 	i := 0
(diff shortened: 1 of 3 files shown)


miekg:
looks good. Maybe add a little test?

chrisohaver:
> looks good. Maybe add a little test?

Thanks, test added.

PR #4307, opened 2020-11-23
add hairpin option to kubernetes plugin

yingnanzhang666:

In practice, we can use SLB or HLB as the lb provider of servicecontroller. For SLB, TLS terminating can be done in the pod sidecar. But for HLB, TLS certs are bound to LB devices, then if the end user wants to access this kind of HLB-TLS services from the cluster inside, the requests must be transferred to ExternalIP instead of ClusterIP.

This PR introduces a new option for kubernetes plugin, named 'hairpin'. This option is used to resolve specific labeled service fqdn to ExternalIP instead of clusterIP.

For example:
```
    hairpin labels hlb-tls=true
}

    hairpin annotations hlb-tls=true
}
```

The kubernetes plugin README.md documentation has been updated.

No.

### DIFF
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -98,6 +98,12 @@ kubernetes [ZONES...] {
 * `ignore empty_service` returns NXDOMAIN for services without any ready endpoint addresses (e.g., ready pods).
   This allows the querying pod to continue searching for the service in the search path.
   The search path could, for example, include another Kubernetes cluster.
+* `hairpin` **HAIRPIN-TYPE** **EXPRESSION** exposes ExternalIP instead of ClusterIP for Kubernetes services
+   that match this selector. Valid values for **HAIRPIN-TYPE**:
+
+   * `labels`: the labels of Services match this selector.
+   * `annotations`: the annotations of Services match this selector.
+
 
 Enabling zone transfer is done by using the *transfer* plugin.
 
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -95,6 +95,11 @@ type dnsControlOpts struct {
 	zones                 []string
 	endpointNameMode      bool
 	skipAPIObjectsCleanup bool
+
+	hairpinLabelSelector        *meta.LabelSelector
+	hairpinLabelAsSelector      labels.Selector
+	hairpinAnnotationSelector   *meta.LabelSelector
+	hairpinAnnotationAsSelector labels.Selector
 }
 
 // newDNSController creates a controller for CoreDNS.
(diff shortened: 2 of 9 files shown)


chrisohaver:
While this looks like it could possibly be useful, the general consensus has been not to extend the existing _kubernetes_ plugin with new non-spec features, but rather implement them as separate plugins.

However, it might be worth discussing the problem in more detail before going further, in case there are any alternate workarounds that would not require code changes in CoreDNS. e.g. The _rewrite_ plugin can help with TLS cert name mismatch issues.

mlowery:
Thanks @chrisohaver. Would you be open to this change instead being added to the `k8s_external` plugin? The usage might look like this:

```
k8s_external example.org {
    labels hlb-tls=true
}
kubernetes example.org in-addr.arpa ip6.arpa {
    ...
}
```

To note here is that the same zone is used for both plugins. And since `k8s_external` [precedes](https://github.com/coredns/coredns/blob/56eea6e6099c41e8913b6c24fbbc156133f22c62/core/dnsserver/zdirectives.go#L46-L47) `kubernetes`, it will have the first chance to answer. If the label doesn't match, it falls through.

chrisohaver:
> Thanks @chrisohaver. Would you be open to this change instead being added to the `k8s_external` plugin?

IMO, it would be better as a separate plugin.  As an extension to the  k8s_external plugin, it would still require code changes to the kubernetes plugin code.

yingnanzhang666:
> While this looks like it could possibly be useful, the general consensus has been not to extend the existing kubernetes plugin with new non-spec features, but rather implement them as separate plugins.

@chrisohaver But these services are also managed by kubernetes with different cloud provider. Isn't it better to extend existing kubernetes plugin to support this potential requirement?

chrisohaver:
> But these services are also managed by kubernetes with different cloud provider. Isn't it better to extend existing kubernetes plugin to support this potential requirement?

A separate plugin could provide the same functionality without further complicating the existing kubernetes plugin with more non-spec features.

As a separate plugin, it could be called k8s_hairpin, or something like that:
1. k8s_hairpin would establish an api watch on Services with a label selector.
2. dns requests would be processed by k8s_hairpin first. If a match is found, the external IP is returned.
3. If no match is found, it falls through to the next plugin, which would be kubernetes.

Though, again, it's probably worth discussing the original problem you are attempting to solve in more detail before going further, in case there are any alternate workarounds that would not require a new plugin or changes to code.

yingnanzhang666:
> Though, again, it's probably worth discussing the original problem you are attempting to solve in more detail before going further, in case there are any alternate workarounds that would not require a new plugin or changes to code.

@ We support different cloud providers in k8s cluster for service, such as software LB provider and centralized hardware LB provider. For the decentralized software LB service, TLS can be terminated to the sidecar container of pod. There's no problem to access these https service via ClusterIP. But for the centralized hardware LB service, certs will be bound to LB device and TLS will be terminated on LB, and then the backend pod use http without TLS, so if we bypass the LB and DNAT the HTTPS traffic to pod directly, it does't work. That is to say, we should resolve service fqdn to IngressAddress instead of ClusterIP, and then skip the DNAT rule.

These services in the same k8s cluster, handled by different cloud providers, use the same zone. And the fqdn for services align with the same pattern `<servicename>.<namespace>.svc.<zone>`. So it's not very suitable to classify them by zone or namepattern.

Consider this kind of centralized LB service, not only the TLS terminating issue, but also the traffic audit by centralized LB, it needs a way to transmit the traffic to real LB. So it's better to let CoreDNS have the capability to resolve service fqdn to ExternalIP on purpose.

chrisohaver:
Thanks for the context. That makes more sense to me now.
I think a separate k8s_hairpin as sketched out above should work.

If you think this is a feature that should be part of the [Kubernetes Service Discovery Spec](https://github.com/kubernetes/dns/blob/master/docs/specification.md), then it should be proposed in the Kubernetes project.  If that is accepted, we would then add the feature to the _kubernetes_ plugin in CoreDNS. In lieu of that, it should be a separate plugin.

chrisohaver:
A side note regarding the labels/annotations mechanic:  The trend I seen in K8s now is to use CRDs to drive this kind of per Service customization instead of labels and or annotations.

chrisohaver:
Closing, since this feature should be implemented as a separate plugin.

Issue #4324, opened 2020-12-09
plugins/cache: serve_stale + prefetch does redundant upstream fetches


coops:
**What happened**:

We have a DNS client in our environment that is... "happy". It does a lot of identical DNS queries, which we'd like to cache at CoreDNS to help protect the upstream resolvers. We use the cache plugin for this.

Conversely, we'd like the cache to remain usable by clients even if the upstream DNS resolver becomes unavailable. So we introduced `serve_stale`. This works well for that purpose, but there is no concurrency control around the refresh code, so during the refresh window CoreDNS proxies redundant queries upstream while awaiting cache fill.

We then turned to `prefetch`, but the same problem exists here; once CoreDNS starts proxying queries upstream to fill the cache it doesn't stop. The `max_concurrent` setting of the `forward` plugin doesn't help -- since it returns SERVFAIL once concurrency limits are exceeded we just end up with cache poisoning.

**What you expected to happen**:

Ideally we could fix this by allowing only 1 prefetch request in flight for a given DNS query. Then we can just tune cache settings until the cache is warm, and we don't have to touch the hot path.

**How to reproduce it (as minimally and precisely as possible)**:

Configure CoreDNS with `forward` and `cache` plugins, as below, making sure to set `serve_stale`. Then spam it with queries for a low-TTL cached response. During the cache-fill period, multiple queries will be proxied upstream.

**Anything else we need to know?**:

**Environment**:

- the version of CoreDNS: 1.8.0
- Corefile:
```
upstream {

    forward . 127.0.0.1:8600
    cache 300 {
        # Serve a response up to 7 days old from the cache
        # Attempts to refresh cache entry immediately after serving
        serve_stale 168h
        prefetch 10 1m 50%
    }
}
```
- OS (e.g: `cat /etc/os-release`): Ubuntu, 16.04 / 18.04

chrisohaver:
I recall that this more or less was requested "relatively recently", but I cannot find the issue.  The request was that a single packet be "in-flight" for any given query name, such that if 10 queries for the same name are received very quickly (before the first gets a response), the latter 9 queries are withheld by the cache plugin until the first gets a response and is written to cache, at which point the latter queries can be answered from cache.

coops:
@chrisohaver yes, exactly. We're happy to fix it. Do you have ideas about how?

kuroneko25:
I think we can synchronize the prefetches and stale fetches and check for freshness in the cache first before spawning those goroutines?

[...]

PR #4326, opened 2020-12-09
plugin/forward: respond with REFUSED when max_concurrent is exceeded

chrisohaver:

Respond with `REFUSED` when `max_concurrent` is exceeded to avoid caching the repsonse.  `REFUSED` responses are typified as `OTHERERROR`, which are not cached by the _cache_ plugin.  Currently, the response code used is `SERVFAIL`, which is cached. If `max_concurrent` exceeded responses are cached, those answer are potentially cached beyond the time the load has dropped, barring those clients from getting a valid answers for some time after the limit has dropped below the threshold.  So, yes, [`REFUSED` works better that `SERVFAIL`](https://github.com/coredns/coredns/pull/3640#discussion_r372902512).

This issue was raised tangentially in #4324


zouyee (approved):

miekg:
REFUSED is actually also the better rcode I think

Issue #4322, opened 2020-12-08
CloudDNS plugin fails to resolve some records


luanphantiki:
**What happened**: 
I am trying to configure clouddns which proxies pass On-prem dns requests to my private clouddns zone. But it seems failed to resolve some records (not at all).

Sample failure record: `uat-gateway-1.svr.huyfa.services.`

```

# dig @localhost uat-gateway-1.svr.huyfa.services.

; <<>> DiG 9.11.4-P2-RedHat-9.11.4-26.P2.el7_9.2 <<>> @localhost uat-gateway-1.svr.huyfa.services.
; (2 servers found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 56932
;; flags: qr aa rd; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;uat-gateway-1.svr.huyfa.services. IN	A

;; AUTHORITY SECTION:
huyfa.services.		900	IN	SOA	ns-gcp-private.googledomains.com. root.huyfa.services. 1607419406 7200 3600 86400 3600

;; Query time: 0 msec
;; SERVER: ::1#53(::1)
;; WHEN: Tue Dec 08 16:50:42 +07 2020
;; MSG SIZE  rcvd: 159

```

All records which relate to this record from my zone:

```

acp-api.dev.huyfa.services.                                                            CNAME  300    uat-gateway-1.svr.huyfa.services.
auth.dev.huyfa.services.                                                               CNAME  300    uat-gateway-1.svr.huyfa.services.
bulma.dev.huyfa.services.                                                              CNAME  300    uat-gateway-1.svr.huyfa.services.
catalog.dev.huyfa.services.                                                            CNAME  300    uat-gateway-1.svr.huyfa.services.
uat.esb.huyfa.services.                                                                CNAME  300    uat-gateway-1.svr.huyfa.services.
uat.graylog.huyfa.services.                                                            CNAME  300    uat-gateway-1.svr.huyfa.services.
uat-gateway-1.svr.huyfa.services.                                                      A      300    10.20.43.53
frontend-staging.huyfa.services.                                                       CNAME  300    uat-gateway-1.svr.huyfa.services.

```

**What you expected to happen**:

It would be like this:

```

[...]

PR #4328, opened 2020-12-10
fix record missing for zone with many of records

luanphantiki:

My zone has more than 2500 records and it misses many of records while using clouddns plugin. So it looks like a bug, where this should be paging through all results and instead is just taking the first page. Ref: https://github.com/coredns/coredns/issues/4322#issuecomment-741946091

https://github.com/coredns/coredns/issues/4322

### DIFF
--- a/plugin/clouddns/gcp.go
+++ b/plugin/clouddns/gcp.go
@@ -1,10 +1,14 @@
 package clouddns
 
-import gcp "google.golang.org/api/dns/v1"
+import (
+	"context"
+
+	gcp "google.golang.org/api/dns/v1"
+)
 
 type gcpDNS interface {
 	zoneExists(projectName, hostedZoneName string) error
-	listRRSets(projectName, hostedZoneName string) (*gcp.ResourceRecordSetsListResponse, error)
+	listRRSets(ctx context.Context, projectName, hostedZoneName string) (*gcp.ResourceRecordSetsListResponse, error)
 }
 
 type gcpClient struct {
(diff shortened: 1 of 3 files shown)


plugin/clouddns/gcp.go
```diff
@@ -23,10 +27,16 @@ func (c gcpClient) zoneExists(projectName, hostedZoneName string) error {
 
 // listRRSets is a wrapper method around `gcp.Service.ResourceRecordSets.List`
 // it fetches and returns the record sets for a hosted zone.
-func (c gcpClient) listRRSets(projectName, hostedZoneName string) (*gcp.ResourceRecordSetsListResponse, error) {
-	rr, err := c.ResourceRecordSets.List(projectName, hostedZoneName).Do()
-	if err != nil {
+func (c gcpClient) listRRSets(ctx context.Context, projectName, hostedZoneName string) (*gcp.ResourceRecordSetsListResponse, error) {
+	req := c.ResourceRecordSets.List(projectName, hostedZoneName)
```
chrisohaver, line 31: No longer need to check for error here?

plugin/clouddns/gcp.go
```diff
@@ -23,10 +27,16 @@ func (c gcpClient) zoneExists(projectName, hostedZoneName string) error {
...
+	var rs []*gcp.ResourceRecordSet
+	if err := req.Pages(ctx, func(page *gcp.ResourceRecordSetsListResponse) error {
```
chrisohaver, line 33: without error check from `c.ResourceRecordSets.List`, can `req` be nil?

plugin/clouddns/gcp.go
lphan-clv: 👍  correct, we should validate here, I'll add & push soon.

plugin/clouddns/gcp.go
lphan-clv: yes I think so, `c.ResourceRecordsets.List` can be checked when listing `Pages`

plugin/clouddns/gcp.go
lphan-clv: Yes, it's be checked while listing Pages

bilaltahirx:
Screenshot from 2020-04-07 01-36-53

plugin/clouddns/gcp.go
chrisohaver: If req is nil, this will crash.

chrisohaver (dismissed):

plugin/clouddns/gcp.go
luanphantiki: @chrisohaver : "req" type is always an object: ResourceRecordSetsListCall, always not nil, read more: https://sourcegraph.com/github.com/googleapis/google-api-go-client/-/blob/dns/v1/dns-gen.go#L5532:37

and "req.Pages" is used to iterate pages and records, error check is implemented already: 
- https://github.com/googleapis/google-api-go-client/blob/074c16e73361434fc3d1f6ef62585d57b70a9d1b/dns/v1/dns-gen.go#L5730
- line 33.

chrisohaver: Ok. Above you said it could be nil. If it cannot, then thats fine.

miekg:
is this ready?

luanphantiki:
> is this ready?

I think so, real PoC that I have applied it on my production for 1 month.

Issue #4323, opened 2020-12-08
core/dnsserver: add ability for plugins to read the http request


balboah:
**What would you like to be added**:

I'm proposing to add a callback similar to the dnsserver Config.FilterFunc but for the HTTP handler to be able to have a dynamic request path.

**Why is this needed**:

I'm looking for a way to write a plugin that needs access to the DoH request path, that can replace the currently static */dns-query*.

---

Let me know what you think, I'm open for suggestions on how to accomplish this. I'd prefer to not have to maintain a fork if possible and will prepare a PR with the agreed solution

johnbelamaric:
Can this just be published as metadata?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] core/dnsserve..." ]
>Can this just be published as metadata?

something like that.

I think the request path is fixed in the standard, so adding special code to work around
that isn't the right approach.

balboah:
Thanks for the input.
At least in for example Firefox, you need to specify the path as well. Also services like next dns uses the custom path as identification. This makes me think it's OK.

I'm not sure I follow the metadata approach.
Are you suggesting to have the path routing at the load balancer/proxy in front of coredns and then push some metadata somehow in the request?

So my end goal with this is to have the capability for a plugin to be able to identify different clients, to provide different configuration. Without changing the client side which are out of my control.

johnbelamaric:
I don't know the details of this. Is this something you just need to *read*, or is it something that you need to configure in CoreDNS that's not currently configurable? If you just need to read it, then you can add code in your plugin (or in the metadata plugin if this is something that is processed at the server level not within your plugin itself). See https://coredns.io/plugins/metadata/ and the associated code.

balboah:
As I see it, there are two changes that needs to be done for my use case:

1. (core) change the static */dns-query* path in https://github.com/coredns/coredns/blob/6bbb48d403ce3e0974809200ef1da2a20e969c70/core/dnsserver/server_https.go#L117 into a function similar to how the tls plugin hooks up its configuration
2. (external plugin) function passed through the configuration could possibly setup the metadata extracted from the http request (url path or other)

I haven't tried changing the code yet. But in theory if it's possible to do it similar to the tls plugin, the core changes outside of my custom plugin would be minimal. Something like "if config func is defined, call that instead of comparing request path with /dns-query"

balboah:
I went ahead and created a POC so you can easier see what I'm trying to explain :) 
[...]

Issue #4259, opened 2020-11-05
track DoH protocol in metrics in logging 


miekg:
Right now we track udp/tcp as the proto family (which is technically the correct thing), but it would be nice to track doh/doc in the future, i.e. the proto for doh would be `https` and for quic .... something else?

miekg:
this would probably mean amending the context to add a "protocol modifier", to we'll have 'tcp', 'udp', 'https', 'tls' for the various protos.

Not sure how to add DoQ when it comes out, as that's http/3 and probably will still use https:// as the prefix for websites

Issue #3460, opened 2019-11-18
https: automatic tls certificate (or in plugin/tls)


zzxap:
start a https server
https://www.mydomain.com  {
  bind myip
  hosts {
    10.6.6.2 sms.service
    10.6.6.3 search.service
  }
}
 how to set https certificate？

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] start a https ser..." ]
>start a https server
>https://www.mydomain.com  {
>  bind myip
>  hosts {
>    10.6.6.2 sms.service
>    10.6.6.3 search.service
>  }
>}
> how to set https certificate？

this has to be done manually. I also think Let's Encrypt; or generally the CA 
industry doesn't handout certs for IP addresses; so you need some kind of 
domain as well.

I'm eyeing caddy 2 ways of setting and getting certs as this has been automated 
already (cc @mholt)

mholt:
Hi - anything I can help with?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] start a https..." ]
>Hi - anything I can help with?

wondering if caddy's acme code can be tranferred/used from coredns to perform a 
simular function, but now for a DNS server.

mholt:
Yep, totally.

In Caddy 2, there is the `tls` app which can manage the certs for you.

Or if you just need cert management features in any other Go program, you can use CertMagic: https://github.com/mholt/certmagic -- this is the same library that Caddy uses. (Caddy 2 uses it also, but Caddy 2 has the `tls` app which is a more centralized way to do it.)

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] start a https..." ]
>Yep, totally.
>
>In Caddy 2, there is the `tls` app which can manage the certs for you.
>
>Or if you just need cert management features in any other Go program, you can use CertMagic: https://github.com/mholt/certmagic -- this is the same library that Caddy uses.

thanks! Would be good to see what's easiest to integrate (at some point), but 
that's homework we need to do.

/Miek

--
Miek Gieben

mholt:
No problem, let me know if you want any more specific guidance or have any questions!

polarathene:
It would be nice to see CoreDNS able to handle an ACME DNS challenge. 

[...]

PR #4329, opened 2020-12-10
custom DoH request validation

balboah:
This is a POC for #4323

### DIFF
--- a/core/dnsserver/config.go
+++ b/core/dnsserver/config.go
@@ -33,7 +33,8 @@ type Config struct {
 	Transport string
 
 	// If this function is not nil it will be used to inspect and validate
-	// HTTP requests.
+	// HTTP requests. Although this isn't referenced in-tree, external plugins
+	// may depend on it.
 	HTTPRequestValidateFunc func(*http.Request) bool
 
 	// If this function is not nil it will be used to further filter access
--- a/core/dnsserver/config.go
+++ b/core/dnsserver/config.go
@@ -3,6 +3,7 @@ package dnsserver
 import (
 	"crypto/tls"
 	"fmt"
+	"net/http"
 
 	"github.com/coredns/caddy"
 	"github.com/coredns/coredns/plugin"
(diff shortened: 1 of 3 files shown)


johnbelamaric:
So the idea then would be some sort of plugin that installs a new validation function? Like: `doh path /my-weird-path` with possible extensions like:

```
doh {
  path /foo
  read-timeout 100
}
```

Or something like that? The (completely made up) `read-timeout` would need more than a validator function of course but from a syntax point of view, is this what you are thinking?

If so, it seems reasonable to me. The `doh` plugin in this case would then publish the configured path as metadata by implementing the appropriate interface.

John

balboah:
My idea is more like http middleware, but only for reading the requests and only one global middleware :)
I may have two clients on DoH where each uses their own dynamic uuid like `https://example.com/<uuid>`

Then my plugin (which doesn't have to be included in this repo) would extract this ID and possibly pass it along using the metadata plugin.
It may for example be for things such as using rewrite plugin to set this uuid in edns0 for special handling upstream:
`rewrite edns0 local set 0xffee {uuid}`

Others could potentially use this in their plugin for collecting HTTP specific stats such as which user agent is used. 
It's not possible do to this afaik without something like these changes

balboah:
Not exactly sure of the details in these other issues, but it might be possible to use this as a first step for hooking up plugins that supports #3460 and #4259 as well

johnbelamaric:
Why do they use UUID in the path? Wouldn't it make more sense to use the standard path and put the UUID in a header?

In the latter case, this change would no longer be the same; instead of a validator, it would just be some sort of getter for the original HTTP request or some sort of inline parser function.

balboah:
The reason why it can’t be a header is that the clients aren’t modified or under my control. They can be any client, as long as they can specify a custom path. Which most can

johnbelamaric:
Ah, those pesky clients.

Ok. So in your case you will install a custom validator function that just checks a regex or something. So that's why my description above is not accurate for you, and we don't need the `doh` options plugin right now. Got it. You just want access to this in your external plugin.

This approach is fine with me.

balboah:
Yes. This sounds spot on 🙂

johnbelamaric:
Unless @miekg has some strong objections, I would merge this but it needs the following:

1) Add a comment that although this override is not used in-tree today, external plugins depend upon it. Otherwise it may get pulled later...
2) Add a test

balboah:
@johnbelamaric up for grabs!

PR #4331, opened 2020-12-11
Fix wrong Datadog port in trace plugin example

chotiwat:
It fixes the Datadog trace port in the example. 8125 is the StatsD port while 8126 is the trace (APM) port.

N/A

It seems only this document needs to be updated. The default port seems to be correctly set to 8126 in the code.

No


johnbelamaric:
Please sign the commit, that's where the DCO failure comes from.

chotiwat:
> Please sign the commit, that's where the DCO failure comes from.

Thanks. It's fixed.

PR #4246, opened 2020-10-28
plugin/kubernetes: Add support for dual stack ClusterIP Services

chrisohaver:
### 1. Why is this pull request needed and what does it do?

This PR adds support for dual stack ClusterIP Services.

PR #4339, opened 2020-12-15
plugin/kubernetes: Add support for dual stack ClusterIP Services

chrisohaver:

Kubernetes 1.20 added support for dual stack Cluster IP services, such that a Cluster IP Service can now have two IP addresses (1 v4 and 1 v6).

replaces #4246


plugin/kubernetes/object/service.go
```diff
@@ -40,13 +40,19 @@ func ToService(obj meta.Object) (meta.Object, error) {
...
 		Index:        ServiceKey(svc.GetName(), svc.GetNamespace()),
-		ClusterIP:    svc.Spec.ClusterIP,
 		Type:         svc.Spec.Type,
 		ExternalName: svc.Spec.ExternalName,
 
 		ExternalIPs: make([]string, len(svc.Status.LoadBalancer.Ingress)+len(svc.Spec.ExternalIPs)),
 	}
 
+	if len(svc.Spec.ClusterIPs) > 0 {
+		s.ClusterIPs = make([]string, len(svc.Spec.Ports))
```
aojea: why  `len(svc.Spec.Ports))` instead of ` len(svc.Spec.ClusterIPs))` ??

chrisohaver: oops. Thats a mistake. Thanks!

plugin/kubernetes/kubernetes.go
```diff
@@ -502,10 +501,11 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 
 			err = nil
 
-			s := msg.Service{Host: svc.ClusterIP, Port: int(p.Port), TTL: k.ttl}
-			s.Key = strings.Join([]string{zonePath, Svc, svc.Namespace, svc.Name}, "/")
-
-			services = append(services, s)
+			for _, ip := range svc.ClusterIPs {
```
zouyee, line 504: If the dual stack feature is not enabled, do we need to preserve signal ClusterIP processing？

chrisohaver: If dual stack is not enabled, ClusterIPs will contain the single ClusterIP as the first element.

chrisohaver:
I'm having trouble creating a functioning dual-stack k8s cluster to actually test these changes.

aojea:
> I'm having trouble creating a functioning dual-stack k8s cluster to actually test these changes.

@chrisohaver https://github.com/aojea/kind/releases/tag/dualstack
ping me in slack if it still doesn´t work

chrisohaver:
@aojea, Thanks so much! That was _exactly_ what I needed.

I created a dual-stack clusterIP service, and the coredns responds as expected when testing A, AAAA, and SRV.

aojea:
great LGTM then :smile:

yongtang (approved):
lgtm

PR #4352, opened 2020-12-22
plugin/forward: fix health check backoff interval between checks

chrisohaver:

The existing implementation of health check backoff is broken.

Currently during a health probe of an upstream, the same sleep interval is used between each check.  The interval only changes/backs-off after the upstream becomes healthy and then goes unhealthy again and another probe is launched.  Since an upstream isn't considered "down" until 3 failed checks, this could potentially delay detection of a down upstream by up to 30 seconds (2X the maximum interval).

This change makes the interval back off between each retry while probing an unhealthy upstream. It also resets the backoff interval at the start of each probe.

### DIFF
--- a/plugin/pkg/up/up.go
+++ b/plugin/pkg/up/up.go
@@ -49,7 +49,7 @@ func (p *Probe) Do(f Func) {
 			}
 			interval := p.expBackoff.NextBackOff()
 			// If exponential backoff has reached the maximum elapsed time, reset it
-			if interval == -1 {
+			if interval == backoff.Stop {
 				p.expBackoff.Reset()
 				interval = p.expBackoff.NextBackOff()
 			}
--- a/plugin/pkg/up/up.go
+++ b/plugin/pkg/up/up.go
@@ -32,13 +32,7 @@ func (p *Probe) Do(f Func) {
 		return
 	}
 	p.inprogress = active
-	interval := p.expBackoff.NextBackOff()
-	// If exponential backoff has reached the maximum elapsed time (15 minutes),
-	// reset it and try again
-	if interval == -1 {
-		p.expBackoff.Reset()
-		interval = p.expBackoff.NextBackOff()
-	}
+	p.expBackoff.Reset()
 	p.Unlock()
 	// Passed the lock. Now run f for as long it returns false. If a true is returned
 	// we return from the goroutine and we can accept another Func to run.
@@ -48,13 +42,19 @@ func (p *Probe) Do(f Func) {
 			if err := f(); err == nil {
 				break
 			}
-			time.Sleep(interval)
 			p.Lock()
 			if p.inprogress == stop {
 				p.Unlock()
 				return
 			}
+			interval := p.expBackoff.NextBackOff()
+			// If exponential backoff has reached the maximum elapsed time, reset it
+			if interval == -1 {
+				p.expBackoff.Reset()
+				interval = p.expBackoff.NextBackOff()
+			}
 			p.Unlock()
+			time.Sleep(interval)
 			i++
 		}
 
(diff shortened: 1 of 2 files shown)


zouyee (approved):

miekg:
this is why the hc is 0.5s (fixed). With exp backoff you also need random probes to see if upstream is healthy again, that's not done.
Please explain why 2 HC checks per second are problematic, this is a tiny amount of traffic and you want to react fast when an upstream is healthy again.

plugin/pkg/up/up.go
```diff
@@ -48,13 +42,19 @@ func (p *Probe) Do(f Func) {
...
 			}
-			time.Sleep(interval)
 			p.Lock()
 			if p.inprogress == stop {
 				p.Unlock()
 				return
 			}
+			interval := p.expBackoff.NextBackOff()
+			// If exponential backoff has reached the maximum elapsed time, reset it
+			if interval == -1 {
```
rdrozhdzh: consider
```go
if interval == backoff.Stop {
```
Shouldn't we stick with max interval (`p.expBackoff.MaxInterval`) in this case?

plugin/pkg/up/up_test.go
```diff
@@ -38,3 +39,35 @@ func TestUp(t *testing.T) {
...
+
+		return errors.New("test")
+	}
+	start := time.Now()
+	wg.Add(1)
+	pr.Do(upfunc)
+	wg.Wait()
+
+	elapsed := time.Now().Sub(start)
+	expected := time.Millisecond * (1 + 2 + 4 + 5 + 6) // plus execution time
```
rdrozhdzh: time.Millisecond * (1 + 2 + 4 + 8 +16)

plugin/pkg/up/up.go
chrisohaver: > Shouldn't we stick with max interval (p.expBackoff.MaxInterval) in this case?

IMO yes, that all makes sense - but in this case i was retaining the original logic (and comment), which intentionally resets.

plugin/pkg/up/up_test.go
chrisohaver: > this is why the hc is 0.5s (fixed). With exp backoff you also need random probes to see if upstream is healthy again, that's not done.
> Please explain why 2 HC checks per second are problematic, this is a tiny amount of traffic and you want to react fast when an upstream is healthy again.

@miekg, if the current backoff behavior is intentional and acceptable to you, I can close this PR.

chrisohaver: @miekg, Ah - are you saying you would rather remove the existing HC backoff behavior?

chrisohaver: Was the intent of #3970 to revert the incremental backoff code added in #3643?

plugin/pkg/up/up.go
rdrozhdzh: ok, I see.
please replace the constant `-1` with `backoff.Stop`

rdrozhdzh:
@miekg, Could you please clarify what is your current position regarding exponential backoff feature?
You added this feature in https://github.com/coredns/coredns/pull/2342 because 

>  when a large cluster of coredns shares
an upstream (original intent of up package) they will hammer the
upstream.

Did you intend to remove this feature as part of https://github.com/coredns/coredns/pull/3970? This PR has only doc change

miekg:
>
>
>In plugin/pkg/up/up_test.go:
>
>> +     upfunc := func() error {
>+               if i > 5 {
>+                       wg.Done()
>+                       return nil
>+               }
>+               i++
>+
>+               return errors.New("test")
>+       }
>+       start := time.Now()
>+       wg.Add(1)
>+       pr.Do(upfunc)
>+       wg.Wait()
>+
>+       elapsed := time.Now().Sub(start)
>+       expected := time.Millisecond * (1 + 2 + 4 + 5 + 6) // plus execution time
>
>@miekg, Ah - are you saying you would rather remove the existing HC backoff
>behavior?

yes, why is it here?

PR #4354, opened 2020-12-23
plugin/forward: Remove HC backoff

chrisohaver:

Remove the existing Health Check backoff behavior, per discussion in #4352.

#2342 - original backoff implementation
#3643 - modified to use exponential backoff package
#3970 - removes backoff from docs - may have intended to also remove code, or assumed code was already removed
#4352 - proposed fix to existing backoff implementation (superseded by this PR)

None - the existing HC behavior is not documented.

### DIFF
--- a/plugin/pkg/up/up.go
+++ b/plugin/pkg/up/up.go
@@ -28,6 +28,7 @@ func (p *Probe) Do(f Func) {
 		p.Unlock()
 		return
 	}
+	p.inprogress = active
 	interval := p.interval
 	p.Unlock()
 	// Passed the lock. Now run f for as long it returns false. If a true is returned
--- a/plugin/pkg/up/up.go
+++ b/plugin/pkg/up/up.go
@@ -9,6 +9,13 @@ import (
 
 // Probe is used to run a single Func until it returns true (indicating a target is healthy). If an Func
 // is already in progress no new one will be added, i.e. there is always a maximum of 1 checks in flight.
+//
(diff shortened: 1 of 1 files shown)


plugin/pkg/up/up.go
```diff
@@ -5,17 +5,14 @@ package up
 import (
 	"sync"
 	"time"
-
-	"github.com/cenkalti/backoff/v4"
 )
 
 // Probe is used to run a single Func until it returns true (indicating a target is healthy). If an Func
 // is already in progress no new one will be added, i.e. there is always a maximum of 1 checks in flight.
```
miekg, line 11:
```suggestion
// is already in progress no new one will be added, i.e. there is always a maximum of 1 checks in flight.
//
// There is a tradeoff to be made in figuring out quickly that an upstream is healthy and not doing to much work 
// (sending queries) to find that out. Having some kind of exp. backoff here won't help much, because you don't won't
// to backoff too much. You then also need random queries to be perfomed every so often to quickly detect a working
// upstream. In the end we just send a query every 0.5 second to check the upstream. This hopefully strikes a balance
// between getting information about the upstream state quickly and not doing too much work. Note that 0.5s is still an 
// eternity in DNS, so we may actually want to shorten it.   
```

miekg:
?
no idea why this is closed. But I'll re-add the PR myself then? The readme still lists this as they the current way https://github.com/coredns/coredns/tree/master/plugin/forward#description so I have no idea why this was changed.

Issue #4364, opened 2021-01-04
Race condition on cache plugin


sschepens:
It seems that coredns cache plugin has a race condition when using `serve_stale` with `prefetch`.
Serve stale, allows to respond a stale record but then fetches the record in a new goroutine with [this code](https://github.com/coredns/coredns/blob/master/plugin/cache/handler.go#L48).
Prefetch allows to perform a out of process refresh when exceeding a certain threshold of the ttl with [this code](https://github.com/coredns/coredns/blob/master/plugin/cache/handler.go#L64).

The problem is that both refreshes can occur at the same time and there is no condition guarding bot refreshes.

This issue also raises another question, why are there two separate ways of handling a cache refresh? I can see that one sets de `DO` bit, but the prefetch doesn't set it, shouldn't this be normalized?

Related to #4298

sschepens:
Another thing, issue #4271 added fix to get the remote address before firing the goroutine in the `serve_stale` refresh, but this was something that was actually already fixed in the `prefetch` [code](https://github.com/coredns/coredns/blob/master/plugin/cache/cache.go#L118) and also explains why it actually fails.

sschepens:
Tagging @gonzalop who seems to be the author of this code and @miekg who reviewed in #3468

PR #4367, opened 2021-01-06
plugin/cache: eliminate stale cache refresh and prefetch race

chrisohaver:

This addresses the redundant fetch and race issue raised in #4324 and #4364.
It adds a test that detects the race condition in the existing code, while passing with these changes.
It consolidates stale cache async refresh and the async prefetch to use the same prefetch code.
Doing this normalizes some different behavior between refresh and prefetch:
1. Do bit is now always set.  Previously it was set during cache refresh, but not prefetch.
2. RemoteIP is used as the remoteIP during cache refresh. Previously it was set to LocalIP during cache refresh.

#4324, #4364

### DIFF
--- a/plugin/cache/prefetch_test.go
+++ b/plugin/cache/prefetch_test.go
@@ -18,6 +18,7 @@ func TestPrefetch(t *testing.T) {
 		qname         string
 		ttl           int
 		prefetch      int
+		staleUpTo     time.Duration
 		verifications []verification
 	}{
 		{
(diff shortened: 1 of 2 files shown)


chrisohaver:
Another thing this changes:  It ticks the prefetch count metric up when fetching a stale cache cache item.  IMO, this is OK, since it is a kind of "prefetch" in itself.

sschepens:
@chrisohaver i hadn't seen this PR and i pushed one myself #4368

I can see we took different approaches.

To make sure we don't hit #4271 we should probably move `newPrefetchResponseWriter` creation outside the goroutine.

plugin/cache/prefetch_test.go
```diff
@@ -18,6 +18,7 @@ func TestPrefetch(t *testing.T) {
 		qname         string
 		ttl           int
 		prefetch      int
+		staleUpTo     time.Duration
```
miekg, line 21: where is this used? I only see assignments to staleUpTo ?

chrisohaver: used here... https://github.com/coredns/coredns/pull/4367/files#diff-b6eb3644f59cc77177fa546593ae1b637cfb4699ba5b1664df21574eb874e2acR125

chrisohaver:
duplicate of #4368

plugin/cache/prefetch_test.go
miekg: I see. lgtm

miekg (approved):

Issue #4298, opened 2020-11-16
Panic during coreDNS runtime


mnevadaf:
**What happened**:
I was running coredns v1.8.0 within an alpine docker container, and received the below error during runtime:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x55bdbece1f4f]
goroutine 116139 [running]:
github.com/miekg/dns.(*Msg).IsEdns0(0xc00074c3f0, 0x0)
	/home/buildozer/go/pkg/mod/github.com/miekg/dns@v1.1.34/defaults.go:153 +0x6f
github.com/coredns/coredns/request.(*Request).Size(0xc000223af0, 0x55bdbff898e0)
	/home/buildozer/aports/testing/coredns/src/coredns-1.8.0/request/request.go:168 +0x45
github.com/coredns/coredns/plugin/forward.(*Proxy).Connect(0xc000272580, 0x55bdc0fd2400, 0xc00064cde0, 0xc00074c3f0, 0x55bdc0fe5080, 0xc000713ec0, 0x0, 0x0, 0x0, 0x0, ...)
	/home/buildozer/aports/testing/coredns/src/coredns-1.8.0/plugin/forward/connect.go:94 +0xcd
github.com/coredns/coredns/plugin/forward.(*Forward).ServeDNS(0xc0004ee480, 0x55bdc0fd2400, 0xc00064cde0, 0x55bdc0fe5080, 0xc000713ec0, 0xc00074c3f0, 0x0, 0x0, 0x0)
	/home/buildozer/aports/testing/coredns/src/coredns-1.8.0/plugin/forward/forward.go:131 +0x530
github.com/coredns/coredns/plugin.NextOrFailure(0x55bdbff8acb3, 0x5, 0x55bdc0f9b340, 0xc0004ee480, 0x55bdc0fd2400, 0xc00064cde0, 0x55bdc0fe5080, 0xc000713ec0, 0xc00074c3f0, 0x0, ...)
	/home/buildozer/aports/testing/coredns/src/coredns-1.8.0/plugin/plugin.go:80 +0xfa
github.com/coredns/coredns/plugin/cache.(*Cache).doPrefetch(0xc000161d40, 0x55bdc0fd2400, 0xc00064cde0, 0xc00074c3f0, 0x55bdc0fe5260, 0xc0004a8680, 0x0, 0x0, 0x200, 0xc0001bfd80, ...)
[...]

PR #4368, opened 2021-01-07
Prevent race from prefetching

### LINKED ISSUE #4364: Race condition on cache plugin

sschepens:
Normalizes fetching and prefetching behavior in cache plugin and prevents multiple prefetches from firing from a single request.

Should fix #4364 and probably #4298

This introduces some slight behavior changes:
- On normal prefetching, the DO bit is now being set if not already set, which was actually already being set for other fetch requests in the cache plugin.
- On stale prefetch it also fakes client address to always be TCP, which was already being done in normal prefetch.

Doubts:
When doing normal prefetch, cache frequency is restored, but when doing a stale prefetch it's not being restored, is this something that should be done?

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -112,7 +112,7 @@ type ResponseWriter struct {
 // newPrefetchResponseWriter returns a Cache ResponseWriter to be used in
 // prefetch requests. It ensures RemoteAddr() can be called even after the
 // original connection has already been closed.
-func newPrefetchResponseWriter(server string, state request.Request, c *Cache, do bool) *ResponseWriter {
+func newPrefetchResponseWriter(server string, state request.Request, c *Cache) *ResponseWriter {
 	// Resolve the address now, the connection might be already closed when the
 	// actual prefetch request is made.
 	addr := state.W.RemoteAddr()
@@ -130,7 +130,6 @@ func newPrefetchResponseWriter(server string, state request.Request, c *Cache, d
 		server:         server,
 		prefetch:       true,
 		remoteAddr:     addr,
-		do:             do,
 	}
 }
 
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -112,7 +112,7 @@ type ResponseWriter struct {
 // newPrefetchResponseWriter returns a Cache ResponseWriter to be used in
 // prefetch requests. It ensures RemoteAddr() can be called even after the
 // original connection has already been closed.
-func newPrefetchResponseWriter(server string, state request.Request, c *Cache) *ResponseWriter {
+func newPrefetchResponseWriter(server string, state request.Request, c *Cache, do bool) *ResponseWriter {
 	// Resolve the address now, the connection might be already closed when the
 	// actual prefetch request is made.
 	addr := state.W.RemoteAddr()
@@ -130,6 +130,7 @@ func newPrefetchResponseWriter(server string, state request.Request, c *Cache) *
 		server:         server,
 		prefetch:       true,
 		remoteAddr:     addr,
+		do:             do,
 	}
 }
 
--- a/plugin/cache/handler.go
+++ b/plugin/cache/handler.go
@@ -39,46 +39,40 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 		ttl = i.ttl(now)
 	}
 	if i == nil {
-		if !do {
-			setDo(rc)
-		}
 		crr := &ResponseWriter{ResponseWriter: w, Cache: c, state: state, server: server, do: do}
-		return plugin.NextOrFailure(c.Name(), c.Next, ctx, crr, rc)
+		return c.doRefresh(ctx, state, crr, do)
 	}
 	if ttl < 0 {
 		servedStale.WithLabelValues(server).Inc()
 		// Adjust the time to get a 0 TTL in the reply built from a stale item.
 		now = now.Add(time.Duration(ttl) * time.Second)
-		addr := w.LocalAddr() // See https://github.com/coredns/coredns/issues/4271, unclear how, but pull this out of the goroutine, and get the address here.
+		cw := newPrefetchResponseWriter(server, state, c, do)
+		go c.doRefresh(ctx, state, cw, do)
+	} else if c.shouldPrefetch(i, now) {
+		cachePrefetches.WithLabelValues(server).Inc()
+		cw := newPrefetchResponseWriter(server, state, c, do)
 		go func() {
-			if !do {
-				setDo(rc)
+			c.doRefresh(ctx, state, cw, do)
+
+			// When prefetching we loose the item i, and with it the frequency
+			// that we've gathered sofar. See we copy the frequencies info back
+			// into the new item that was stored in the cache.
+			if i1 := c.exists(state); i1 != nil {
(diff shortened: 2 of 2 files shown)


chrisohaver:
@sschepens thanks.  Did you see #4367?

sschepens:
> @sschepens thanks. Did you see #4367?

Yes @chrisohaver, saw it after i pushed this, i made a comment there

plugin/cache/cache.go
```diff
@@ -112,7 +112,7 @@ type ResponseWriter struct {
 // newPrefetchResponseWriter returns a Cache ResponseWriter to be used in
 // prefetch requests. It ensures RemoteAddr() can be called even after the
 // original connection has already been closed.
-func newPrefetchResponseWriter(server string, state request.Request, c *Cache) *ResponseWriter {
+func newPrefetchResponseWriter(server string, state request.Request, c *Cache, do bool) *ResponseWriter {
```
miekg: why does this need a `do` bool? We always want to do a DO enabled lookup and fix the reply for clients that didn't ask for it.

sschepens: i made this because refresh when stale did this, it keeps the original do and not the modified

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Prevent race ..." ]
>@sschepens commented on this pull request.
>
>
>In plugin/cache/cache.go:
>
>> @@ -112,7 +112,7 @@ type ResponseWriter struct {
> // newPrefetchResponseWriter returns a Cache ResponseWriter to be used in
> // prefetch requests. It ensures RemoteAddr() can be called even after the
> // original connection has already been closed.
>-func newPrefetchResponseWriter(server string, state request.Request, c *Cache) *ResponseWriter {
>+func newPrefetchResponseWriter(server string, state request.Request, c *Cache, do bool) *ResponseWriter {
>
>i made this because refresh when stale did this, it keeps the original do and
>not the modified

The original 'do' from .. the client? For prefetch we (technically) don't have that
anymore?

sschepens:
> The original 'do' from .. the client? For prefetch we (technically) don't have that anymore?

Seems so, i don't know why this si done, but it's stored in the response writer, so it maybe has something to do when handling the response from the backend or storing something in the cache.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Prevent race ..." ]
>    The original 'do' from .. the client? For prefetch we (technically) don't
>    have that anymore?
>
>Seems so, i don't know why this si done, but it's stored in the response
>writer, so it maybe has something to do when handling the response from the
>backend or storing something in the cache.

not in this path. I've recently changed *cache* to do a DO lookup, always. (To fix another
bug). So we unconditionally want to do this with prefetch as well.
When replying to a client we filter out the sigs to make non-DO replies work. In this
codepath would should not care about 'do' to add a bool, this should be done implicitly
anyway.

plugin/cache/cache.go
sschepens: Ok, removing this

sschepens:
@miekg I removed the setting of Do on the response writer

plugin/cache/handler.go
```diff
@@ -39,39 +39,28 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 		ttl = i.ttl(now)
 	}
 	if i == nil {
-		if !do {
-			setDo(rc)
-		}
 		crr := &ResponseWriter{ResponseWriter: w, Cache: c, state: state, server: server, do: do}
-		return plugin.NextOrFailure(c.Name(), c.Next, ctx, crr, rc)
+		return c.doRefresh(ctx, state, crr, do)
```
miekg: this bool `do` should also be removed

sschepens: Why should it be removed? it's used to later set do if not already set.
Maybe we can always run this code regardless we are doing a fetch?

miekg: that should already be in state or said differently the message in state should either have DO set or we should check and set it, no reason to carry a boolean argument here.

sschepens: ok, removed

plugin/cache/handler.go
```diff
@@ -81,6 +70,13 @@ func (c *Cache) doPrefetch(ctx context.Context, state request.Request, server st
 	}
 }
 
+func (c *Cache) doRefresh(ctx context.Context, state request.Request, cw *ResponseWriter) (int, error) {
+	if !state.Do() {
+		setDo(state.Req)
```
miekg, line 75: So, now we have this operating on the state.Req which is a pointer to a dns.Msg. Is this thread safe? It's still done in a go function? So I have some trouble seeing how this change would now fix the race?

sschepens: there is no race because no one else is accessing `state.Req`, we could move all prefetches after response is written to make sure no one really uses state.Req, but this should already be true

sschepens: or we could modify the request on the main thread and then fire a goroutine

miekg:
OK, thanks for your work on this. I think this should be close to being mergeable, but I'm still fuzzy on your commit msg in the PR:
"Normalizes fetching and prefetching behavior in cache plugin and prevents multiple prefetches from firing from a single request."
And this probably me, not paying enough attention, but can you make clearer there how this PR prevents that?

plugin/cache/handler.go
```diff
@@ -39,39 +39,28 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
-		go func() {
-			if !do {
-				setDo(rc)
-			}
-			crr := &ResponseWriter{Cache: c, state: state, server: server, prefetch: true, remoteAddr: addr, do: do}
-			plugin.NextOrFailure(c.Name(), c.Next, ctx, crr, rc)
-		}()
+		cw := newPrefetchResponseWriter(server, state, c)
+		go c.doPrefetch(ctx, state, cw, i, now)
+	} else if c.shouldPrefetch(i, now) {
```
sschepens, line 51: @miekg this `else if` is the change that prevents from firing multiple prefetches, basically, we only prefetch if we haven't triggered a refresh because of a stale item

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Prevent race ..." ]
>@miekg this else if is the change that prevents from firing multiple
>prefetches, basically, we only prefetch if we haven't triggered a refresh

ok!

/lgtm
/merge

sschepens:
@miekg thanks! is there a chance of getting a release out with this change?

chrisohaver:
yes https://github.com/coredns/coredns/issues/4369#issuecomment-760917393

PR #4371, opened 2021-01-09
forward/up: HC every 0.5 seconds, do not do exponential backoff

miekg:
See https://github.com/coredns/coredns/pull/4354/

For some reason that was closed.

### DIFF
--- a/plugin/pkg/up/up.go
+++ b/plugin/pkg/up/up.go
@@ -28,6 +28,7 @@ func (p *Probe) Do(f Func) {
 		p.Unlock()
 		return
 	}
+	p.inprogress = active
 	interval := p.interval
 	p.Unlock()
 	// Passed the lock. Now run f for as long it returns false. If a true is returned
--- a/plugin/pkg/up/up.go
+++ b/plugin/pkg/up/up.go
@@ -9,6 +9,13 @@ import (
 
 // Probe is used to run a single Func until it returns true (indicating a target is healthy). If an Func
 // is already in progress no new one will be added, i.e. there is always a maximum of 1 checks in flight.
+//
+// There is a tradeoff to be made in figuring out quickly that an upstream is healthy and not doing to much work 
+// (sending queries) to find that out. Having some kind of exp. backoff here won't help much, because you don't won't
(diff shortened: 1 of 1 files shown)


plugin/pkg/up/up.go
```diff
@@ -5,17 +5,21 @@ package up
...
 	"time"
-
-	"github.com/cenkalti/backoff/v4"
 )
 
 // Probe is used to run a single Func until it returns true (indicating a target is healthy). If an Func
 // is already in progress no new one will be added, i.e. there is always a maximum of 1 checks in flight.
-// When failures start to happen we will back off every second failure up to maximum of 4 intervals.
+//
+// There is a tradeoff to be made in figuring out quickly that an upstream is healthy and not doing to much work
```
chrisohaver, line 13: FWIW, I think there is a benefit to doing retry backoff.  It helps by very quickly restoring forwarding to upstreams that may only have been momentarily unreachable, but when the upstream is down for a long time it reduces sustained strain by sending less frequent health checks.  The significance is more pronounced when CoreDNS is deployed as a horizontally scaled microservice where there can be many CoreDNS instances sending health checks to the same upstream.

Retry backoff was added a couple of years ago, but it appears to have never worked as intended ...

The bug in the initial backoff implementation #2342 prevented the backoff interval from increasing between retries.

The followup PR #3643 essentially added random jitter to the backoff interval.  But it did not resolve the bug in #2342 so the backoff interval was still static.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] forward/up: H..." ]
>FWIW, I think there is a benefit to doing retry backoff. It helps by very
>quickly restoring forwarding to upstreams that may only have been momentarily
>unreachable, but when the upstream is down for a long time it reduces sustained
>strain by sending less frequent health checks. The significance is more
>pronounced when CoreDNS is deployed as a horizontally scaled microservice where
>there can be many CoreDNS instances sending health checks to the same upstream.

yes, 2 qps per server, that's not much even if you have 300 instances running. And you
want to quickly react so even with exp backoff you will be sending way more queries in
between.

PR #4376, opened 2021-01-10
plugin/file: document wrong behavior in lookup fox Apex

### LINKED ISSUE #4374: NXDOMAIN returned when the domain exists
**What happened**:

Hi,

CoreDNS returns `NXDOMAIN` for queries with no matching type, but the domain name exists.

|  	|  	|  	|  	|
|-	|-	|-	|-	|
| example.com. 	| 500 | SOA 	| ns1.outside.com. root.example.com. 3 604800 86400 2419200 604800 	|
| example.com. 	| 500 | NS 	| ns1.outside.com.  	|

For the query _<example.com. , A>_, the Coredns server returns the following response:
```
          "opcode QUERY",
          "rcode NXDOMAIN",
          "flags QR AA",
          ";QUESTION",
          "example.com. IN A",
          ";ANSWER",
          ";AUTHORITY",
          "example.com. 500 IN SOA ns1.outside.com. root.example.com. 3 604800 86400 2419200 604800",
          ";ADDITIONAL"
```

The query name `example.com.` exists, but the requested type `A`  does not, so the rcode should be `NOERROR` instead of `NXDOMAIN.`  The response code is important as resolvers can use QNAME minimization as described in [RFC 7816: QNAME Minimization](https://tools.ietf.org/html/rfc7816) when resolving names like `child.foo.example.com` and will wrongly conclude that there is nothing at `example.com.` as well as below it.

**Environment**:

- the version of CoreDNS: `CoreDNS-1.8.0  linux/amd64, go1.14.13, 31abdc2f`
- Corefile:
```
example.com:53 {
    file db.example.com
    log
    errors
}
```
- OS: Ubuntu 18
-

miekg:
An apex only zone returns the wrong answer, document this by adding a
test.

Closes: #4374

### DIFF
--- a/plugin/file/apex_test.go
+++ b/plugin/file/apex_test.go
@@ -1,0 +1,45 @@
+package file
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/coredns/coredns/plugin/test"
+
+	"github.com/miekg/dns"
+)
+
+const exampleApexOnly = `$ORIGIN example.com.
+@   IN      SOA     ns1.example.com. admin.example.com.  (
+                               2005011437 ; Serial
+                               1200       ; Refresh
+                               144        ; Retry
+                               1814400    ; Expire
+                               2h )       ; Minimum
+@           IN  NS      ns1.example.com.
+`
+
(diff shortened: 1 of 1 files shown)


miekg:
merging as just a test and no functionality change.

PR #4385, opened 2021-01-12
[plugin][forward] - Populate return code on upstream error.

nyodas:
Following an incident we noticed that the global metrics did not have the right return code when upstream failure were happening.
More specifically `i/o timeout` errors
~Also if the client is patient enough (>30s) the query is returned with a *noerror* code.~

Graph showing app errors
👆 This is a client metrics showing dns error (timeout / servfail)
Graph showing apm errors
👆 Note the _30s_ timeout
Graph showing no servfail errors
☝️ Blue is NOERROR , which should have been another color because of a different error code.

~In order to fix that particular issue, i thought of 2 ways to go about it.~
~The one in the PR~
 modifying the [metrics/handler.go:ServeDns](https://github.com/coredns/coredns/blob/master/plugin/metrics/handler.go#L29) function to use the `status` instead of `rw.Rcode`
~Let me know if the second one is better or if you see a better way of doing this.~

None that i found

None

No

### 5. Additional Question / Info

Additionally the [maxDialTimeout](https://github.com/coredns/coredns/blob/master/plugin/forward/persistent.go#L154) seems a bit high at *30s* with client usually having the default *5s* timeout.
It tends to create a queuing problem.
I saw the cancel plugin but did not find the wiring in the forward plugin. 
I'm curious about how you see this working.

Edit: Strikethrough based on comment.

### DIFF
--- a/plugin/metrics/handler.go
+++ b/plugin/metrics/handler.go
@@ -24,11 +24,15 @@ func (m *Metrics) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 
 	// Record response to get status code and size of the reply.
 	rw := dnstest.NewRecorder(w)
-	status, err := plugin.NextOrFailure(m.Name(), m.Next, ctx, rw, r)
+	rcodeFromNext, err := plugin.NextOrFailure(m.Name(), m.Next, ctx, rw, r)
+	rcodeRet := rcode.ToString(rw.Rcode)
+	if !plugin.ClientWrite(rcodeFromNext) && err != nil {
+		rcodeRet = rcode.ToString(rcodeFromNext)
+	}
 
-	vars.Report(WithServer(ctx), state, zone, rcode.ToString(rw.Rcode), rw.Len, rw.Start)
+	vars.Report(WithServer(ctx), state, zone, rcodeRet, rw.Len, rw.Start)
 
-	return status, err
+	return rcodeFromNext, err
 }
 
 // Name implements the Handler interface.
--- a/plugin/trace/trace.go
+++ b/plugin/trace/trace.go
@@ -117,13 +117,17 @@ func (t *trace) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 
 	rw := dnstest.NewRecorder(w)
 	ctx = ot.ContextWithSpan(ctx, span)
-	status, err := plugin.NextOrFailure(t.Name(), t.Next, ctx, rw, r)
-
+	rcodeFromNext, err := plugin.NextOrFailure(t.Name(), t.Next, ctx, rw, r)
+	rcodeRet := rcode.ToString(rw.Rcode)
+	if !plugin.ClientWrite(rcodeFromNext) && err != nil {
+		rcodeRet = rcode.ToString(rcodeFromNext)
+	}
+	
 	span.SetTag(tagName, req.Name())
 	span.SetTag(tagType, req.Type())
 	span.SetTag(tagProto, req.Proto())
 	span.SetTag(tagRemote, req.IP())
-	span.SetTag(tagRcode, rcode.ToString(rw.Rcode))
+	span.SetTag(tagRcode, rcodeRet)
 
-	return status, err
+	return rcodeFromNext, err
 }


plugin/forward/forward.go
```diff
@@ -177,6 +177,9 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	}
 
 	if upstreamErr != nil {
+		serverFailureErr := new(dns.Msg)
```
miekg: this is done here: https://github.com/coredns/coredns/blob/master/core/dnsserver/server.go#L250

nyodas: In that case it is done to late in the flow for other plugin to use.
Namely metrics and apm.
It make sense to do it at the end of the chain as a sort of catch all.but the next plugin apparently do not have the info. unless i go to option 2 by changing the usage of the [ metrics/handler.go:ServeDns](https://github.com/coredns/coredns/blob/master/plugin/metrics/handler.go#L29)
And possibly in the [trace/trace.go:ServeDns](https://github.com/coredns/coredns/blob/master/plugin/trace/trace.go#L126)
Or am i missing something.

nyodas:
Implemented option 2 and updated [metrics/handler.go:ServeDns](https://github.com/coredns/coredns/blob/master/plugin/metrics/handler.go#L29)
And [trace/trace.go:ServeDns](https://github.com/coredns/coredns/blob/master/plugin/trace/trace.go#L126)

miekg:
ah I see. You may indeed want to care about status (although the only conveys "have we written to the client", nothing more really". The rcode and error are the important bits here.
I can (maybe) see what you want, but the PR as it's stands is not the correct way

nyodas:
> ah I see. You may indeed want to care about status (although the only conveys "have we written to the client", nothing more really". The rcode and error are the important bits here.
> I can (maybe) see what you want, but the PR as it's stands is not the correct way

Shame, indeed that won't do.if status can be virtually anything and is not limited to existing rcode.

How should i approach it instead ? 
Because apart from those 2 i don't see another way to introspect the earlier failure of `forward` to have the right metrics set up.

nyodas:
Pushed a third option, though to be fair i think previous attempt was more in line with the rest of the codebase.

Basically check if ~`rw.Rcode` is `0` (initial value and noerror) and that we do have an error in previous plugins (This feel clunky if further plugin manage to resolve, though metrics and trace are on top of the stack so the last one to go)~ and hasn't been written to the client yet (the same way [servedns](https://github.com/coredns/coredns/blob/master/core/dnsserver/server.go#L249) does

> (although the only conveys "have we written to the client", nothing more really".

Weirdly the return of that func is used as rcode fairly widely 
In the [error plugin](https://github.com/coredns/coredns/blob/master/plugin/errors/errors.go#L83) to communicate to the user that an error occurred with the right rcode
The *error* plugin is also lower down the stack of plugin than both *metrics* and *trace* 
In the root [servedns](https://github.com/coredns/coredns/blob/master/core/dnsserver/server.go#L249) -> [checking that the answer is still missing depending on the rcode](https://github.com/coredns/coredns/blob/master/plugin/plugin.go#L88-L99)
And in the [go-doc](https://github.com/coredns/coredns/blob/master/plugin/plugin.go#L23-L49) which outline the return expected of `ServeDNS` and thus of `NextOrFailure`
So `status` is in fact an artifact of naming and should be called rcode or something to that effect.

plugin/metrics/handler.go
```diff
@@ -24,11 +24,15 @@ func (m *Metrics) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 
 	// Record response to get status code and size of the reply.
 	rw := dnstest.NewRecorder(w)
-	status, err := plugin.NextOrFailure(m.Name(), m.Next, ctx, rw, r)
+	rcodeFromNext, err := plugin.NextOrFailure(m.Name(), m.Next, ctx, rw, r)
```
miekg, line 27: plugins don't return an rcode, it's code to see if the server needs to write something to the client. So it next to useless for the purpose you want to use it for.

miekg:
> Following an incident we noticed that the global metrics did not have the right return code when upstream failure were happening.
> More specifically i/o timeout errors

This is a valid point and I'm open to add what's need here, but this PR is not doing that as explained in comments. Closing

nyodas:
> > Following an incident we noticed that the global metrics did not have the right return code when upstream failure were happening.
> > More specifically i/o timeout errors
> 
> This is a valid point and I'm open to add what's need here, but this PR is not doing that as explained in comments. Closing

Should i open an issue ?

Or can you point me in the right direction ?

PR #4386, opened 2021-01-13
Add git plugin

miekg:
This adds the git plugin from github.com/miekg/coredns-git

It's *does not* add a e2e tests, as (1) this will take more than a
second, and (2) is not hermetic.

I'll still need to do some local testing to see if this works correctly.

### DIFF
--- a/plugin/git/commands.go
+++ b/plugin/git/commands.go
@@ -1,0 +1,48 @@
+package git
+
+import (
+	"bytes"
+	"os/exec"
+	"sync"
+)
+
+type gitCmd struct {
+	command string
+	args    []string
+	dir     string
+
+	sync.RWMutex
+}
+
+// Exec executes the command initiated in gitCmd.
+func (g *gitCmd) Exec(dir string) error {
+	g.Lock()
+	g.dir = dir
+	g.Unlock()
+	return runCmd(g.command, g.args, dir)
+}
+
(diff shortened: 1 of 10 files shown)


plugin/git/commands.go
```diff
@@ -0,0 +1,48 @@
+package git
+
+import (
+	"bytes"
+	"os/exec"
```
SuperQ, line 5: I'm a big un-fan of having `os/exec` in a service like this. Having CoreDNS fork processes in the background seems like a bad idea.

Most people who wan to store config in git will have some kind of external management like Ansible/Chef/Kubernetes.

SuperQ:
Is there a monitoring plan for when git processes get stuck?

plugin/git/commands.go
miekg: so don't use this plugin?
I don't see why executing a git command every hour would be a bad thing? Can you explain?

miekg:
a "monitoring plan?"

plugin/git/commands.go
SuperQ: Because these things are better handled async, outside of a serving process. For example, a cron job.

Counter question, what is the advantage of re-inventing cron here?

miekg:
it seems noone else is chiming in. Less is more, so closing

PR #4387, opened 2021-01-13
plugin/file: guard against cname loops

miekg:
This adds a loop counter to the context, which is checked. If we get
more than 9 loops, we break the off the lookup and return whatever we
have found.

This is different than the normal cname loop (which we already check) because we call z.externalLookup which goes back to the main process.

Adding the following comment in the code:

~~~
Note the answer we're returning will be incomplete (more cnames to be followed) or
illegal (wildcard cname with multiple identical records). For now it's more important
to protect ourselves then to give the client a valid answer.

Note under the moto of garbage in, garbage out, we may actually _not_ care.
~~~

### DIFF
--- a/plugin/file/lookup.go
+++ b/plugin/file/lookup.go
@@ -3,6 +3,7 @@ package file
 import (
 	"context"
 
+	"github.com/coredns/coredns/core/dnsserver"
 	"github.com/coredns/coredns/plugin/file/rrutil"
 	"github.com/coredns/coredns/plugin/file/tree"
 	"github.com/coredns/coredns/request"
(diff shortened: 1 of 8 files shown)


plugin/file/lookup.go
```diff
@@ -62,6 +61,11 @@ func (z *Zone) Lookup(ctx context.Context, state request.Request, qname string)
 		elem, wildElem *tree.Elem
 	)
 
+	loop := 0
+	if x := ctx.Value(loopKey); x != nil {
```
bradbeam: Do we need to be concerned about leaking the loop data to other plugins in the chain?

bradbeam (approved):

plugin/file/lookup.go
miekg: as far as I understand context.Context you can't access this because `loopKey` is private.

miekg: OTOH having the ability for other plugins (or the server?) to check this might be a possibility

miekg: maybe doing this more generally is actually a good thing. Also returning SERVFAIL might be an option here, and not a garbled response

miekg:
@bradbeam  rework this a bit and made it available for all plugins. Could you take another look, would be good to have this in coredns 1.8.1 (but not super critical)

bradbeam (approved):

miekg:
Thanks!

/merge

PR #4389, opened 2021-01-14
plugin/acl: add the ability to filter records

georgyo:
Currently ACLs only allow for allow and block, however it isn't
always desirable to set the status code to REFUSED. Often times
you want to completely hide the fact that those records even exist.

Adding the ability to acl to filter results makes it significantly
harder for a third party to know that the records are being masked.

None
I made a few in this module, but maybe more?

None

### DIFF
--- a/plugin/acl/acl.go
+++ b/plugin/acl/acl.go
@@ -45,6 +45,8 @@ const (
 	actionAllow
 	// actionBlock blocks unauthorized queries towards protected DNS zones.
 	actionBlock
+	// actionFilter returns empty sets for queries towards protected DNS zones.
+	actionFilter
 )
 
 // ServeDNS implements the plugin.Handler interface.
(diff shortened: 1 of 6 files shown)


plugin/acl/acl.go
```diff
@@ -73,7 +75,16 @@ RulesCheckLoop:
 			{
 				break RulesCheckLoop
 			}
+		case actionFilter:
+			{
+				m := new(dns.Msg)
+				m.SetRcode(r, dns.RcodeSuccess)
+				w.WriteMsg(m)
+				RequestBlockCount.WithLabelValues(metrics.WithServer(ctx), zone).Inc()
```
ihac: It should have its own metric?

plugin/acl/acl_test.go
```diff
@@ -145,6 +145,34 @@ func TestACLServeDNS(t *testing.T) {
 			dns.RcodeSuccess,
 			false,
 		},
+		{
+			"Filter 1 BLOCKED",
```
ihac: `BLOCKED` does not make sense here. Could use `FILTERED` or sth else

plugin/acl/acl.go
georgyo: I was on the fence about this before the initial submit, however you're likely right. Added.

ihac: Yea, understand. Just my suggestion though.

ihac (approved):
LGTM.

yongtang (approved):
/lgtm

PR #4390, opened 2021-01-14
fix useless proxy probing due to NS query not being restored

### LINKED ISSUE #4388: Etcd + forward for . causes upstream query flood on "NS ." query
**What happened**:

DNS servers used for forwarding are hammered with queries.

An error should be returned

docker-compose.yml:

```
version: '3.5'
  
services:

  coredns:
    image: mycoredns
    container_name: coredns
    hostname: coredns
    restart: always
    ports:
      - "127.0.0.1:5553:53/udp"
    environment:
      COREDNS_ETCD: coredns-etcd

  coredns-etcd:
    container_name: coredns-etcd
    image: quay.io/coreos/etcd
    restart: always
    volumes:
      - ./coredns-etcd:/etcd-data
    environment:
      - ETCD_NAME=coredns-etcd
      - ETCD_DATA_DIR=/etcd-data
      - ETCD_LISTEN_CLIENT_URLS=http://0.0.0.0:2379
      - ETCD_ADVERTISE_CLIENT_URLS=http://coredns-etcd:2379
```

Dockerfile (to build mycoredns):

```
FROM coredns/coredns:1.8.0
COPY Corefile /
CMD [ "-conf", "/Corefile" ]
```

Corefile:

```
. {
    etcd {
        fallthrough
        path /coredns
        endpoint http://{$COREDNS_ETCD}:2379
    }
    forward . 8.8.8.8 9.9.9.9
    log
    errors
}
```

```
$ docker build -t mycoredns .
$ docker-compose up
```

With the container running, `dig @127.0.0.1 -p 5553 . NS` causes a huge flood of queries towards 8.8.8.8 or whatever server is listed in the `forward` line.

After some seconds, the following error is logged:

```
[ERROR] plugin/errors: 2 ns.dns.. NS: dns: bad rdata
```

and the dig query fails. 
I'm not going into whether the Corefile is correct, but I think the observed behavior should not happen regardless.

EDIT: the flood appears to come from the healthchecker, which also sends a ". NS" probe query, but since the probes are successful, the proxy is not declared down and the retry loop keeps spinning.

waldner:
Should fix #4388

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -424,11 +424,11 @@ func NS(ctx context.Context, b ServiceBackend, zone string, state request.Reques
 	state.Clear()
 	state.Req.Question[0].Name = dnsutil.Join("ns.dns.", zone)
 	services, err := b.Services(ctx, state, false, opt)
+	// ... and reset
+	state.Req.Question[0].Name = old
 	if err != nil {
 		return nil, nil, err
 	}
-	// ... and reset
-	state.Req.Question[0].Name = old
 
 	seen := map[string]bool{}
 
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -151,6 +151,11 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 		upstreamErr = err
 
 		if err != nil {
+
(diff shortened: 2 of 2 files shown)


miekg:
we shouldn't break here, backups like tcp might be in play and not used.

PR #4391, opened 2021-01-15
plugin/forward Add rcode and rtype to request_duration_seconds metric

wayt:

This adds the response code (`rcode`) and request type (`type`) to `coredns_forward_request_duration_seconds` metric.

We've been using these extra tags in production for a while to help us debug unexpected behavior. It could be a nice observability addition to CoreDNS.

Documentation updated in forward's README.

no

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -107,7 +107,7 @@ If monitoring is enabled (via the *prometheus* plugin) then the following metric
 
 * `coredns_forward_requests_total{to}` - query count per upstream.
 * `coredns_forward_responses_total{to}` - Counter of responses received per upstream.
-* `coredns_forward_request_duration_seconds{to}` - duration per upstream interaction.
+* `coredns_forward_request_duration_seconds{to, rcode, type}` - duration per upstream, RCODE, type
 * `coredns_forward_responses_total{to, rcode}` - count of RCODEs per upstream.
 * `coredns_forward_healthcheck_failures_total{to}` - number of failed health checks per upstream.
 * `coredns_forward_healthcheck_broken_total{}` - counter of when all upstreams are unhealthy,
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -11,6 +11,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/coredns/coredns/plugin/pkg/dnsutil"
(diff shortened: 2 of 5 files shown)


wayt:
Hey @miekg @johnbelamaric , this is my first contribution to CoreDNS. Let me know if I'm doing something wrong with the process.

Thanks!

miekg:
rtype contains too many values and will blow up whatever is scraping coredns - we have a hack to lumps all lesser know rtypes into  others to make that work

miekg:
rcode might be useful, but if there is a timeout you don't have a timeout, so I don't think we can add this.

your commit/description also doesn't say why you need this?

wayt:
I understand the concern about metric cardinality, I'll leave it up to you.

> your commit/description also doesn't say why you need this?

We've been using these extra tags in production for a while to help us debug unexpected behavior. Thought it would be a good addition to roll that out by default in CoreDNS.

miekg:
>I understand the concern about metric cardinality, I'll leave it up to you.
>
>    your commit/description also doesn't say why you need this?
>
>We've been using these extra tags in production for a while to help us debug
>unexpected behavior. Thought it would be a good addition to roll that out by
>default in CoreDNS.

ack, please add that to the description.

Rcode won't fly as I said. rtype will, but you need to do:
https://github.com/coredns/coredns/blob/master/plugin/metrics/vars/report.go#L28
(which may need to be extracted and put in pkg/dnsutil or something)

>
>—
>You are receiving this because you were mentioned.
>Reply to this email directly, view it on GitHub, or unsubscribe.*
>
/Miek

--
Miek Gieben

rdrozhdzh:
@miekg Rcode is already used in another metrics `RcodeCount`, so why it won't work for `RequestDuration`?
In case of timeout (error) we don't update the metrics at all

miekg:
How does that stance help with the intent to use this for debugging?

wayt:
Here is an example of a graph where we had an elevated and inconsistent response duration, and when splitting by `rcode` it appeared that it was because of `rcode:servfail` (lighblue), while `rcode:noerror` (blue) is fine.

Screen Shot 2021-01-15 at 1 28 42 PM

miekg:
thanks @wayt that image helps.

wayt:
Thank you!

SuperQ:
Late to the party, but this seems like a useful label.

PR #4408, opened 2021-01-19
plugin/trace - Use compatible tag name for datadog

nyodas:
Traces are currently working properly with datadog with the exception of having the ability to facet the tags

image

In order to get valuable graph in dashboard the tag on the metrics need to be faceted by datadog.
This PR update the tags with facetable path.
While keeping backward compatibility with existing path for zipkin.

Graph like:
image
image
image
image
image

None
None that i'm aware
I don't think so.

### DIFF
--- a/plugin/trace/trace.go
+++ b/plugin/trace/trace.go
@@ -25,13 +25,33 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


miekg:
this looks good to me

PR #4409, opened 2021-01-19
plugin/file: Make copies of RRs before returning them to avoid tree mutation 

chrisohaver:

This PR makes copies of RRs from the tree before writing them to the client.  If the original RRs are written to the client, a mutating response writer (e.g. rewrite's `ResponseReverter`) will mutate the RR elements in the file's in memory tree, affecting the responses to later queries.

[Reported on slack](https://cloud-native.slack.com/archives/C4DF7FP71/p1611065238019400?thread_ts=1611051481.018900&cid=C4DF7FP71) by @svinther

### DIFF
--- a/plugin/file/lookup.go
+++ b/plugin/file/lookup.go
@@ -171,7 +171,12 @@ func (z *Zone) Lookup(ctx context.Context, state request.Request, qname string) 
 			return z.externalLookup(ctx, state, elem, rrs)
 		}
 
-		rrs := elem.Type(qtype)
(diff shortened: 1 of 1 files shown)


chrisohaver:
Not sure if it's better to do this in _file_, or in ResponseReverter ...

miekg (approved):

miekg:
actually, I agree with your comment, this should be done in mutating plugins. Lets revert update docs and fix in tree users, *loadbalance* might also do this, but less severe

PR #4411, opened 2021-01-20
Add cache eviction metrics to the cache plugin

fhriley:

Adds the cache evictions count to the cache plugin. This is a useful metric for determining if your cache capacity is too small.

None

Added to the cache plugin readme.

No


miekg:
don't think this hurts.

Issue #4369, opened 2021-01-09
CoreDNS 1.8.1 Release Tracking


miekg:
Tracking bug for coredns 1.8.1 release.

1 or 2 PRs needs merging, but then we can kick off a 1.8.1 release.

yongtang:
See PR #4370

miekg:
note to self: the version is already incremented to 1.8.1!

miekg:
the cache race issue has been merged, and the HC is now fired every 0.5s, those where important PR from my side.

Anything more we want to put in? Otherwise I'll kick off 1.8.1 release soonish (yes still need to finalize the notes)

miekg:
Oh I also want this: https://github.com/coredns/coredns/pull/4387

sschepens:
@miekg seems like everything's merged now, can we get the release?

yongtang:
Added an additional PR #4410 to update the release note with PR #4380 and #4387. 

@miekg if everything is ok I think we can release the 1.8.1 after the merge.

miekg:
seen that, excellent. Pushed some minor polish to the notes and merged that.

miekg:
oh joy, the docker bullcrap is broken again - getting this || close to drop that stuff altogether

miekg:
A likely casualty from this is the braindead docker "multi-arch" stuff that requires this crap to work. Meaning going forward we'll only have a amd64 docker image

miekg:
ok, doing some minimal fixes in this area is not working, ready to rip it all out

miekg:
can probably install manifest-tool manually and remove the compilation step from Makefile.release, for the time being

miekg:
See https://github.com/coredns/coredns/pull/4414

miekg:
fuck this docker shit, I'm going to rip that stuff completely out

miekg:
there are images in docker, but crap manifest-tool didn't run, so I have no idea. 

I have no sympathy left for this stuff, going forward an amd64 image will be pushed and even for that thing I'm getting itchy

miekg:
I'm calling it done

PR #4415, opened 2021-01-21
release: remove manifest tool

miekg:
Also include mips and mips64le in the list LINUX_ARCH, that docker can't
deal with this is not our problem; unify and simplify our stuff.

This simple stops using the manifest-tool; I don't know what will break
for those poor souls that actually use non-amd64 arches and docker.

Docker still warns that the manifest tool support is experimental, so
going forward we will highly likely only publish amd64 images.

See #4369 for context on how badly this is working

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -47,17 +47,13 @@ COMMA:=$(EMPTY),$(EMPTY)
 ifeq (, $(shell which curl))
     $(error "No curl in $$PATH, please install")
 endif
-ifeq (, $(shell which manifest-tool))
-    $(error "No manifest-tool in $$PATH, please install")
-endif
 
 DOCKER:=
 NAME:=coredns
 VERSION:=$(shell grep 'CoreVersion' coremain/version.go | awk '{ print $$3 }' | tr -d '"')
 GITHUB:=coredns
 DOCKER_IMAGE_NAME:=$(DOCKER)/$(NAME)
-# mips is not in LINUX_ARCH because it's not supported by the manifest-tool.
-LINUX_ARCH:=amd64 arm arm64 ppc64le s390x
+LINUX_ARCH:=amd64 arm arm64 ppc64le s390x mips mips64le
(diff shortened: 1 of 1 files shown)


mrueg:
@miekg Would you be interested in using https://github.com/goreleaser/goreleaser and https://www.docker.com/blog/multi-arch-images/ for the project?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] release: remo..." ]
>@miekg Would you be interested in using https://github.com/goreleaser/
>goreleaser and https://www.docker.com/blog/multi-arch-images/ for the project?

if they don't break midway in the next year, then yes

SuperQ:
I've used goreleaser in several projects, works pretty well as long as you don't need CGO.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] release: remo..." ]
>I've used goreleaser in several projects, works pretty well as long as you
>don't need CGO.

ack.

Note the normal releasing works perfectly well.

mrueg:
I can try to provide a PR for it next week. It would be good to have container images built for non-amd64 arches as well, since there is a growing number of kubernetes users on arm/arm64 platform for example.

chrisohaver:
I have used docker's built-in multi-arch build tool in [another project](https://github.com/chrisohaver/dnsdrone/blob/master/Makefile).  It works, however it was still in beta when I last played with it.

sschepens:
@miekg I don't see version `1.8.1` released in docker hub, and also `latest` isn't updated, I do see some other tags have been pushed.

Is this related to PR?

zouyee:
- Tags missing  
截屏2021-01-22 下午4 04 53

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] release: remo..." ]
>@miekg I don't see version 1.8.1 released in docker hub, and also latest isn't
>updated, I do see some other tags have been pushed.

probably, the docker crap blew up, for the umph time, and I'm not going to work around this
time.

miekg:
see #4421 for the better fix

PR #4416, opened 2021-01-21
copy rrs

miekg:
- Revert "make copies of RRs before returning them (#4409)"
- Document copying responses

### DIFF
--- a/plugin.md
+++ b/plugin.md
@@ -131,6 +131,14 @@ The `fallthrough` directive should optionally accept a list of zones. Only queri
 in one of those zones should be allowed to fallthrough. See `plugin/pkg/fallthrough` for the
 implementation.
 
+## Mutating a Response
+
+If the plugin mutates a response it MUST make a copy of the entire response before doing so. A
+response is a pointer to a `dns.Msg` and as such you will be manipulating the original response,
+which could have been generated from a data store. E.g. the *file* plugin creates a response that
+the *rewrite* plugin then rewrites; not copying the data, means it's **also** mutating the data of
+the *file*'s data store. A response can be copied by using the `Copy()` method.
+
 ## General Guidelines
 
 Some general guidelines:
--- a/plugin.md
+++ b/plugin.md
@@ -133,7 +133,8 @@ implementation.
 
 ## Mutating a Response
 
-If the plugin mutates a response it MUST make a copy of the entire response before doing so. A
+Using a custom `ResponseWriter`, a plugin can mutate a response when another plugin further down the chain writes the response to the client.
+If a plugin mutates a response it MUST make a copy of the entire response before doing so. A
 response is a pointer to a `dns.Msg` and as such you will be manipulating the original response,
 which could have been generated from a data store. E.g. the *file* plugin creates a response that
 the *rewrite* plugin then rewrites; not copying the data, means it's **also** mutating the data of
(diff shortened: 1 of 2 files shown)


miekg:
cc @chrisohaver

chrisohaver:
Is there a potential write/read race condition here when a plugin such as _file_ or _auto_ is updating an RR in the tree, at the same time the response writer is reading that RR?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] copy rrs (#44..." ]
>Is there a potential write/read race condition here when a plugin such as file
>or auto is updating an RR in the tree, at the same time the response writer is
>reading that RR?

likely yes... :-( so maybe not revert?

/Miek

--
Miek Gieben

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] copy rrs (#44..." ]
>Is there a potential write/read race condition here when a plugin such as file
>or auto is updating an RR in the tree, at the same time the response writer is
>reading that RR?

although file doesn't update a single RR (no dynamic updates)

chrisohaver:
> although file doesn't update a single RR (no dynamic updates)

Right - I suppose it replaces the entire tree, leaving the previous tree to be GC'ed. I see in file that only the tree/apex pointers are locked, not the contents of the trees. So, I think it's safe as is.  

Ultimately, putting the responsibility of fixing this on response writers seems better because it only copies when needed.

plugin.md
```diff
@@ -131,6 +131,14 @@ The `fallthrough` directive should optionally accept a list of zones. Only queri
 in one of those zones should be allowed to fallthrough. See `plugin/pkg/fallthrough` for the
 implementation.
 
+## Mutating a Response
+
+If the plugin mutates a response it MUST make a copy of the entire response before doing so. A
```
chrisohaver:
```suggestion
Using a custom `ResponseWriter`, a plugin can mutate a response when another plugin further down the chain writes the response to the client.
If a plugin mutates a response it MUST make a copy of the entire response before doing so. A
```

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] copy rrs (#44..." ]
>    although file doesn't update a single RR (no dynamic updates)
>
>Right - I suppose it replaces the entire tree, leaving the previous tree to be
>GC'ed. I see in file that only the tree/apex pointers are locked, not the
>contents of the trees. So, I think it's safe as is.
>
>Ultimately, putting the responsibility of fixing this on response writers seems
>better because it only copies when needed.

Agreed

miekg:
still needs fixed in rewrite

chrisohaver:
> still needs fixed in rewrite

was expecting that to be added to this pr before it merged ...

chrisohaver:
@miekg , Should we open an issue so we don't forget about it?  Or do you have a fix in the works already?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] copy rrs (#44..." ]
>    still needs fixed in rewrite
>
>was expecting that to be added to this pr before it merged ...

I suspect more plugins we need something, so didn't want to conflate and enlarge this
(which was essentally a revert only)

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] copy rrs (#44..." ]
>@miekg , Should we open an issue so we don't forget about it? Or do you have a
>fix in the works already?

feel free. I have no code laying around

PR #4417, opened 2021-01-22
Added minimal-responses plugin

### LINKED ISSUE #4220: add *minimal* plugin

gdsoumya:
Signed-off-by: Soumya Ghosh Dastidar <gdsoumya@gmail.com>

- This pull requests adds the minimal-responses plugin mentioned in the [BIND 9 Configuration Reference](https://bind9.readthedocs.io/en/latest/reference.html#boolean-options)
- Corrects comment for `NextOrFailure` function for plugins
closes #4220 
A new README.md has been added for the plugin which should serve as it's documentation
No.


plugin/minimal/minimal.go
```diff
@@ -0,0 +1,41 @@
...
+
+// minimalResponse implements the plugin.Handler interface.
+type minimalResponse struct {
+	Next plugin.Handler
+}
+
+func (m *minimalResponse) Name() string { return "minimal-responses" }
+
+func (m *minimalResponse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	// if plugin active, try minimizing response
```
miekg:
```suggestion
```

plugin/minimal/minimal.go
```diff
@@ -0,0 +1,41 @@
...
+	nw := nonwriter.New(w)
+
+	// Call all the next plugin in the chain.
```
miekg:
```suggestion
```

plugin/minimal/minimal.go
```diff
@@ -0,0 +1,41 @@
...
+	rcode, err := plugin.NextOrFailure(m.Name(), m.Next, ctx, nw, r)
+	if err != nil {
+		// if error received then just return the error
```
miekg:
```suggestion
```

plugin/minimal/minimal.go
```diff
@@ -0,0 +1,41 @@
...
+		return rcode, err
+	}
+	// else write minimized response
```
miekg:
```suggestion
```

plugin/minimal/README.md
```diff
@@ -0,0 +1,31 @@
+# minimal-responses
```
miekg: Just name it `minimal` and drop the `-responses` part, this can be done throughout

plugin/minimal/README.md
```diff
@@ -0,0 +1,31 @@
...
+~~~ corefile
+example.org {
+    whoami
+    minimal-responses
+}
+~~~
+
+## See Also
+
+[BIND 9 Configuration Reference](https://bind9.readthedocs.io/en/latest/reference.html#boolean-options)
```
miekg, line 36: still bit hard to find the actual docs for minimal responses there. Is there a more direct link? Also interesting to see the amount of options there...

plugin/minimal/minimal.go
```diff
@@ -0,0 +1,41 @@
...
+	w.WriteMsg(m.minimizeResponse(nw.Msg))
+	return rcode, err
+}
+
+func (m *minimalResponse) minimizeResponse(msg *dns.Msg) *dns.Msg {
```
miekg: so yeah, this is the thing that needs fleshing out

plugin/minimal/minimal_test.go
```diff
@@ -0,0 +1,120 @@
...
+type responseStruct struct {
+	answer []dns.RR
+	ns     []dns.RR
+	extra  []dns.RR
+	rcode  int
+}
+
+func TestMinimizeResponse(t *testing.T) {
+	baseAnswer := []dns.RR{
+		test.A("google.com.  293 IN A 142.250.76.46"),
```
miekg: please use example.org (or example.com/net) in tests

plugin/minimal/minimal_test.go
```diff
@@ -0,0 +1,120 @@
+package minimal
+
+import (
+	"testing"
+
+	"github.com/coredns/coredns/plugin/test"
+	"github.com/miekg/dns"
+)
+
+type responseStruct struct {
```
miekg: that test package has Case, i.e. `test.Case` that might be useful here as well.

plugin/minimal/setup.go
```diff
@@ -0,0 +1,27 @@
+package minimal
+
+import (
+	"github.com/coredns/caddy"
+	"github.com/coredns/coredns/core/dnsserver"
+	"github.com/coredns/coredns/plugin"
+)
+
+func init() {
+	caddy.RegisterPlugin("minimal-responses", caddy.Plugin{
```
miekg: there is a 1-line short cut for this now

miekg (commented):
basic all look good, but still bunch of comments.

Let me check what we can really strip from responses.

plugin/minimal/README.md
gdsoumya: I tried to find other links there was one but it wasn't directly from bind : https://www.zytrax.com/books/dns/ch7/queries.html#minimal-responses

plugin/minimal/minimal_test.go
gdsoumya: Okay I will take a look at that, I was following the general pattern in the plugins and examples I found

plugin/minimal/setup.go
gdsoumya: I see, is this the function `plugin.Register("minimal", setup)`

plugin/minimal/minimal.go
gdsoumya: Is this logic enough or is there anything else needed?

plugin/minimal/minimal_test.go
gdsoumya: @miekg I searched for `Case` in the default go testing pkg but i was unable to find it. Can you provide a link to the doc maybe?

plugin/minimal/setup.go
gdsoumya: Made the change

plugin/minimal/minimal_test.go
gdsoumya: done

plugin/minimal/README.md
gdsoumya: done

.gitignore
```diff
@@ -4,3 +4,4 @@ coredns.exe
 build/
 release/
 vendor/
+.idea/
```
miekg: please remove

plugin/minimal/minimal.go
```diff
@@ -0,0 +1,37 @@
...
+	rcode, err := plugin.NextOrFailure(m.Name(), m.Next, ctx, nw, r)
+	if err != nil {
+		return rcode, err
+	}
+	w.WriteMsg(m.minimizeResponse(nw.Msg))
+	return rcode, err
+}
+
+func (m *minimalHandler) minimizeResponse(msg *dns.Msg) *dns.Msg {
+	if ty, _ := response.Typify(msg, time.Now().UTC()); ty == response.NoError {
```
miekg: you need to create a new message here to prevent modifying an existing dns.Msg that we cache for instance, also worth changing this into a switch statement and at least have all the response.* handled there (some will be empty).

Also this function can be removed and the code moved into the `ServeDNS` function

gdsoumya: instead of mentioning each `response.*` type as a case would it make more sense to keep one case for `response.NoError` and a default case (empty) for others? currently the other cases don't have any logic so until new cases are required we can use the default case directly.

plugin/minimal/minimal.go
miekg: This is a diff of what needs added here
~~~
diff --git plugin/minimal/README.md plugin/minimal/README.md
index 1781fb3d..165bb84e 100644
--- plugin/minimal/README.md
+++ plugin/minimal/README.md
@@ -6,8 +6,12 @@
 
 ## Description
 
-The *minimal* plugin tries to minimize the size of the response. Depending on the response type it removes resource records from the AUTHORITY and ADDITIONAL sections.
+The *minimal* plugin tries to minimize the size of the response. Depending on the response type it
+removes resource records from the AUTHORITY and ADDITIONAL sections.
 
+Specifically this plugin looks at successful responses (this excludes negative responses, i.e.
+nodata or name error). If the successful response isn't a delegation only the RRs in the answer
+section are written to the client.
 
 ## Syntax
 
@@ -21,7 +25,7 @@ Enable minimal responses:
 
 ~~~ corefile
 example.org {
-    whoami
+    forward . 8.8.8.8
     minimal
 }
 ~~~
diff --git plugin/minimal/minimal.go plugin/minimal/minimal.go
index c63cdeb0..bdd82019 100644
--- plugin/minimal/minimal.go
+++ plugin/minimal/minimal.go
@@ -2,11 +2,13 @@ package minimal
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/coredns/coredns/plugin/pkg/nonwriter"
 	"github.com/coredns/coredns/plugin/pkg/response"
+
 	"github.com/miekg/dns"
 )
 
@@ -24,14 +26,34 @@ func (m *minimalHandler) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *
 	if err != nil {
 		return rcode, err
 	}
-	w.WriteMsg(m.minimizeResponse(nw.Msg))
-	return rcode, err
-}
+	// Depending on the type of response we can remove RRs from auth or additional section. Create a new message and
+	// copy the RRs we need into it.
+	ty, _ := response.Typify(nw.Msg, time.Now().UTC())
+	cl := response.Classify(ty)
+	// Denial or Error of left through, note this in includes Type=Meta and Type=Update, but that fine for  now
+	if cl == response.Denial || cl == response.Error {
+		w.WriteMsg(nw.Msg)
+		return 0, nil
+	}
+
+	// Further look at the response. We leave delegations alone (although we may prune those, but lets make it too difficult for ourselves
+	if ty == response.Delegation {
+		w.WriteMsg(nw.Msg)
+		return 0, nil
+	}
 
-func (m *minimalHandler) minimizeResponse(msg *dns.Msg) *dns.Msg {
-	if ty, _ := response.Typify(msg, time.Now().UTC()); ty == response.NoError {
-		msg.Extra = nil
-		msg.Ns = nil
+	if ty != response.NoError {
+		w.WriteMsg(nw.Msg)
+		return 0, plugin.Error("minimal", fmt.Errorf("unhandled response type %q for %q", ty, nw.Msg.Question[0].Name))
+	}
+	d := new(dns.Msg)
+	d.SetReply(r)
+	// Only copy the RRs in the answer section and write that to the client
+	d.Answer = make([]dns.RR, len(nw.Msg.Answer))
+	for i := range nw.Msg.Answer {
+		d.Answer[i] = nw.Msg.Answer[i] // this doesn't need to be a deep copy, we're not changing the RRs.
 	}
-	return msg
+
+	w.WriteMsg(d)
+	return 0, nil
 }
~~~

gdsoumya: As we don't need a deep copy can't we just set `d.Answer = nw.Msg.Answer`

gdsoumya: Also shouldn't the other parameters from the original msg be copied over too like header?

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Added minimal..." ]
>@gdsoumya commented on this pull request.
>
>
>In plugin/minimal/minimal.go:
>
>> +func (m *minimalResponse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
>+       // if plugin active, try minimizing response
>+       nw := nonwriter.New(w)
>+
>+       // Call all the next plugin in the chain.
>+       rcode, err := plugin.NextOrFailure(m.Name(), m.Next, ctx, nw, r)
>+       if err != nil {
>+               // if error received then just return the error
>+               return rcode, err
>+       }
>+       // else write minimized response
>+       w.WriteMsg(m.minimizeResponse(nw.Msg))
>+       return rcode, err
>+}
>+
>+func (m *minimalResponse) minimizeResponse(msg *dns.Msg) *dns.Msg {
>
>As we don't need a deep copy can't we just set d.Answer = nw.Msg.Answer

right

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Added minimal..." ]
>@gdsoumya commented on this pull request.
>
>
>In plugin/minimal/minimal.go:
>
>> +func (m *minimalResponse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
>+       // if plugin active, try minimizing response
>+       nw := nonwriter.New(w)
>+
>+       // Call all the next plugin in the chain.
>+       rcode, err := plugin.NextOrFailure(m.Name(), m.Next, ctx, nw, r)
>+       if err != nil {
>+               // if error received then just return the error
>+               return rcode, err
>+       }
>+       // else write minimized response
>+       w.WriteMsg(m.minimizeResponse(nw.Msg))
>+       return rcode, err
>+}
>+
>+func (m *minimalResponse) minimizeResponse(msg *dns.Msg) *dns.Msg {
>
>Also shouldn't the other parameters from the original msg be copied over too
>like header?

yes, that too :)

plugin/minimal/minimal.go
gdsoumya: I have pushed the suggested changes.

miekg:
oh wow that diff looks horrible in github.

miekg:
I'll take a closer look tomorrow and if need do a small PR on this branch, but I think we're close

gdsoumya:
I also added one more minor correction to the comments for `NextOrFailure`. According to the doc string provided here https://github.com/coredns/coredns/blob/6ff19342d3b6d354daebbbe131514f2d6514a905/plugin/plugin.go#L72 it should return nil as error if the next handler is absent but in the code it does return an error `Error(name, errors.New("no next plugin found"))`. 

I asked in the slack channel whether this is the correct behaviour and was told that it is, so I updated the comment to reflect the same.

gdsoumya:
@miekg any update on this PR?

miekg:
I need to look into some recent CI failures, but I think this can be just merged.

plugin/minimal/minimal_test.go
miekg: https://github.com/coredns/coredns/blob/master/plugin/test/helpers.go#L31

gdsoumya: Updated the test cases to use `test.Case` as suggested

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Added minimal..." ]
>@gdsoumya commented on this pull request.
>
>
>In plugin/minimal/minimal_test.go:
>
>> @@ -0,0 +1,120 @@
>+package minimal
>+
>+import (
>+       "testing"
>+
>+       "github.com/coredns/coredns/plugin/test"
>+       "github.com/miekg/dns"
>+)
>+
>+type responseStruct struct {
>
>Updated the test cases to use test.Case as suggested

excellent. I review/merge on Monday

gdsoumya:
@miekg any update on this?

PR #4420, opened 2021-01-23
core: flag blacklisting not needed anymore

miekg:
The glog flag blacklisting doesn't seem to be necessary anymore as k8s
moved to klog. This was added back in the glog days.


yongtang (approved):
/lgtm

PR #4421, opened 2021-01-24
Makefile.release: Replace manifest-tool with docker manifest

mrueg:
This commit replaces `manifest-tool` with `docker manifest`, which should
hopefully provide a better experience for releasing coredns.

In addition it integrates mips64le (since it's supported by `docker
manifest`) in LINUX_ARCHES, so docker images for mips64le are pushed as
well.

Alternative to https://github.com/coredns/coredns/pull/4415
None
If people use them docker.io/coredns/coredns:$arch tags are not updated anymore and changed to docker.io/coredns/coredns-$arch:$version

I've tested it locally and pushed to my dockerhub namespace.

```
docker manifest inspect mrueg/coredns:1.8.1
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 739,
         "digest": "sha256:4e7a16c46bb028aae66635b2131318df3015a32ff0cfee4d4086f9a8d2675af9",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 739,
         "digest": "sha256:7a46d6f4fbfc1a0c2443a81f98a709d75d8cdf7da00419f3c697ef9c8d0d35b0",
         "platform": {
            "architecture": "arm",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 739,
         "digest": "sha256:c8aba8abbb3478aea65ccc6678e2edbffe113a6298375e5cec99347b16f645d7",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 739,
         "digest": "sha256:01019c876f207abb7d868e60728468dafcc678f19323ca221b2ddf31d8d96678",
         "platform": {
            "architecture": "mips64le",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 739,
         "digest": "sha256:4b51c8aa0826f7af32b557eae79849a0a26c351e39d7e85b8115c236b5bb592b",
         "platform": {
            "architecture": "ppc64le",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 739,
         "digest": "sha256:d0f3e086c01762d125ac9c78765a60a6a97bd1ef430e4b65d632576cd8e74e04",
         "platform": {
            "architecture": "s390x",
            "os": "linux"
         }
      }
   ]
}
```

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -44,11 +44,8 @@ EMPTY:=
 SPACE:=$(EMPTY) $(EMPTY)
 COMMA:=$(EMPTY),$(EMPTY)
 
-ifeq (, $(shell which curl))
+ifeq (, $(shell command -v curl))
     $(error "No curl in $$PATH, please install")
-endif
-ifeq (, $(shell which manifest-tool))
-    $(error "No manifest-tool in $$PATH, please install")
 endif
 
 DOCKER:=
@@ -56,8 +53,8 @@ NAME:=coredns
 VERSION:=$(shell grep 'CoreVersion' coremain/version.go | awk '{ print $$3 }' | tr -d '"')
 GITHUB:=coredns
 DOCKER_IMAGE_NAME:=$(DOCKER)/$(NAME)
-# mips is not in LINUX_ARCH because it's not supported by the manifest-tool.
-LINUX_ARCH:=amd64 arm arm64 ppc64le s390x
+# mips is not in LINUX_ARCH because it's not supported by docker manifest
+LINUX_ARCH:=amd64 arm arm64 mips64le ppc64le s390x
 PLATFORMS:=$(subst $(SPACE),$(COMMA),$(foreach arch,$(LINUX_ARCH),linux/$(arch)))
 
 all:
@@ -77,8 +74,6 @@ build:
 	mkdir -p build/windows/amd64 && $(MAKE) coredns BINARY=build/windows/amd64/$(NAME).exe SYSTEM="GOOS=windows GOARCH=amd64" CHECKS="" BUILDOPTS=""
 	@echo Building: linux/mips - $(VERSION)
 	mkdir -p build/linux/mips  && $(MAKE) coredns BINARY=build/linux/mips/$(NAME) SYSTEM="GOOS=linux GOARCH=mips" CHECKS="" BUILDOPTS=""
-	@echo Building: linux/mips64le - $(VERSION)
-	mkdir -p build/linux/mips64le  && $(MAKE) coredns BINARY=build/linux/mips64le/$(NAME) SYSTEM="GOOS=linux GOARCH=mips64le" CHECKS="" BUILDOPTS=""
 	@echo Building: linux/$(LINUX_ARCH) - $(VERSION) ;\
 	for arch in $(LINUX_ARCH); do \
 	    mkdir -p build/linux/$$arch  && $(MAKE) coredns BINARY=build/linux/$$arch/$(NAME) SYSTEM="GOOS=linux GOARCH=$$arch" CHECKS="" BUILDOPTS="" ;\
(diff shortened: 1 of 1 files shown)


chrisohaver (commented):
Thanks! It should also update/push the `latest` tag.

chrisohaver (commented):
looks good thanks.

IMO/FYI: I had always viewed the single arch tags as "cruft" left behind from the multi-arch build process that was not cleaned up afterwards (either because doing do broke stuff, or because its a pita to do so).  From that POV, tagging them with a version results in increasing the amount of cruft every release.

mrueg:
> looks good thanks.
> 
> IMO/FYI: I had always viewed the single arch tags as "cruft" left behind from the multi-arch build process that was not cleaned up afterwards (either because doing do broke stuff, or because its a pita to do so). From that POV, tagging them with a version results in increasing the amount of cruft every release.

unfortunately we need them otherwise docker manifest won't work.

chrisohaver:
> > looks good thanks.
> > IMO/FYI: I had always viewed the single arch tags as "cruft" left behind from the multi-arch build process that was not cleaned up afterwards (either because doing do broke stuff, or because its a pita to do so). From that POV, tagging them with a version results in increasing the amount of cruft every release.
> 
> unfortunately we need them otherwise docker manifest won't work.

I know that they are needed _while_ creating the manifest (I found it very kludgey that it cannot work with local copies, but yes I'm aware of the limitation).  Are you saying they must also be tagged with versions now?

mrueg:
> I know that they are needed _while_ creating the manifest (I found it very kludgey that it cannot work with local copies, but yes I'm aware of the limitation). Are you saying they must also be tagged with versions now?

They don't need to be tagged with versions. I found it cleaner and more convenient in case people for some reason want tags for arch specific versions (eventually they don't take up any storage since it's pointing to the same content as the manifest).
If you want me to, I can revert it back and create all manifests from :latest

chrisohaver:
> If you want me to, I can revert it back and create all manifests from :latest

Fine as is, if others are not concerned.

Makefile.release
```diff
@@ -44,20 +44,17 @@ EMPTY:=
...
-endif
 
 DOCKER:=
 NAME:=coredns
 VERSION:=$(shell grep 'CoreVersion' coremain/version.go | awk '{ print $$3 }' | tr -d '"')
 GITHUB:=coredns
 DOCKER_IMAGE_NAME:=$(DOCKER)/$(NAME)
-# mips is not in LINUX_ARCH because it's not supported by the manifest-tool.
-LINUX_ARCH:=amd64 arm arm64 ppc64le s390x
+DOCKER_CLI:=docker
```
miekg: why this? It's even long to type?

Makefile.release
```diff
@@ -148,13 +143,25 @@ docker-push:
...
+	@# Pushes multi-arch image to coredns/coredns:$version
+	export DOCKER_CLI_EXPERIMENTAL=enabled
+	@echo $(DOCKER_PASSWORD) | $(DOCKER_CLI) login -u $(DOCKER_LOGIN) --password-stdin
 	@echo Pushing: $(VERSION) to $(DOCKER_IMAGE_NAME)
 	for arch in $(LINUX_ARCH); do \
-	    docker push $(DOCKER_IMAGE_NAME):coredns-$$arch ;\
+		$(DOCKER_CLI) push $(DOCKER_IMAGE_NAME)-$${arch}:$(VERSION) ;\
+		$(DOCKER_CLI) push $(DOCKER_IMAGE_NAME)-$${arch}:latest ;\
+	done
+	$(DOCKER_CLI) manifest create --amend $(DOCKER_IMAGE_NAME):$(VERSION) $(shell echo $(LINUX_ARCH) | sed -e "s~[^ ]*~$(DOCKER_IMAGE_NAME)\-&:$(VERSION)~g")
```
miekg: can we pull the sed command out of this and assign to variable with a sensible name?

Makefile.release
mrueg: Extracting it, so at some point folks can replace it with buildah or similar alternatives.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Makefile.rele..." ]
>@mrueg commented on this pull request.
>
>
>In Makefile.release:
>
>>
> DOCKER:=
> NAME:=coredns
> VERSION:=$(shell grep 'CoreVersion' coremain/version.go | awk '{ print $$3 }' | tr -d '"')
> GITHUB:=coredns
> DOCKER_IMAGE_NAME:=$(DOCKER)/$(NAME)
>-# mips is not in LINUX_ARCH because it's not supported by the manifest-tool.
>-LINUX_ARCH:=amd64 arm arm64 ppc64le s390x
>+DOCKER_CLI:=docker
>
>Extracting it, so at some point folks can replace it with buildah or similar
>alternatives.

Sure, but we will not do that

Makefile.release
mrueg: Done!

mrueg:
@miekg reverted the DOCKER_CLI changes.

miekg (approved):

miekg:
Thanks!

I think we should do a 1.8.2 rather soon to test this (testing this stuff is too hard - we should either fix that, or release way more often).

nerzhul:
any plan to release 1.8.2 soon ?

miekg:
after fixing the Makefile.release and then testing it, it seems still broken, but I have no idea how bad that is:
~~~
he push refers to repository [docker.io/miekg/coredns-amd64]
b273bfd4251d: Preparing 
225df95e717c: Preparing 
make: *** [Makefile.release:156: docker-push] Error 1
~~~
update: ok this final thing is me being stupid

Issue #4428, opened 2021-01-26
Detection of K8s minor version faulty


uablrek:
**What happened**:

When making a `nslookup` on a headless dual-stack service in `v1.21.0-alpha.1` only addresses of the base ipFamily are returned.

**What you expected to happen**:

Addresses of both families shall be returned.

**How to reproduce it (as minimally and precisely as possible)**:

On `v1.21.0-alpha.1` or K8s "master", create a headless dual-stack service with some endpoints;
```
apiVersion: v1
kind: Service
metadata:
  name: mserver-headless-dual
spec:
  ipFamilyPolicy: PreferDualStack
  clusterIP: None
  selector:
    app: mserver-deployment
  ports:
  - port: 5001
    name: mconnect
```
Do a "nslookup";
```
nslookup mserver-headless-dual.default.svc.xcluster
Server:         192.168.0.1
Address:        192.168.0.1:53

Name:   mserver-headless-dual.default.svc.xcluster
Address: 11.0.1.2
Name:   mserver-headless-dual.default.svc.xcluster
Address: 11.0.1.5
Name:   mserver-headless-dual.default.svc.xcluster
Address: 11.0.2.4
Name:   mserver-headless-dual.default.svc.xcluster
Address: 11.0.4.4
```

Also check the coredns log;
```
[INFO] plugin/kubernetes: watching Endpoints instead of EndpointSlices in k8s versions < 1.19
```

**Anything else we need to know?**:

The problem is here;
https://github.com/coredns/coredns/blob/a9206996eb89314f7bf1c0b6c87e5da083615b8e/plugin/kubernetes/kubernetes.go#L259-L264

`sv.Minor` will contain an appended `+` sign like "21+", Atoi will fail and the "minor" variable become 0 (zero).

**Environment**:

- the version of CoreDNS: v1.8.1
- Corefile:
- logs, if applicable: (see above)
- OS (e.g: `cat /etc/os-release`):
- Others:

uablrek:
Here is a comment about the appended "+";
https://github.com/kubernetes/client-go/blob/b72204b2445de5ac815ae2bb993f6182d271fdb4/pkg/version/base.go#L43

uablrek:
Suggested fix;
```diff
diff --git a/plugin/kubernetes/kubernetes.go b/plugin/kubernetes/kubernetes.go
index a7fa378..ce9ca60 100644
--- a/plugin/kubernetes/kubernetes.go
[...]

PR #4430, opened 2021-01-26
Corrected detection of K8s minor version

uablrek:

For not-yet-released K8s versions i.e. alpha, rc and local builds the minor version was not detected correctly.

This caused endpoints to be used when endpointslices should be used.

#4428

N/A

No


chrisohaver (approved):
LGTM Thanks!

PR #4431, opened 2021-01-27
plugin/sign: track zone file's mtime

### LINKED ISSUE #4406: plugin/sign: check if source zone is changed
When adding or removing records from a zone file it would be nice if the *sign* plugin would detect this is re-signs the zone, right now you need to remove the signed zone and SIGUSR1 the coredns process to force a resign and then wait a minute for the file plugin to pick up the new zone.

md5summing the source zone and using that to detect a change would facilitate that, iirc the *file* plugin already does this.

### LINKED ISSUE #4407: Answer rewrites not performed if a subsequent plugin returns RcodeServerFailure
**What happened**: When using the rewrite plugin with answer rewrites, if a subsequent plugin returns RcodeServerFailure, the rewrite plugin answer rewrites do not execute. As a result, a reply is sent with a question field whose name does not match the original query name, causing "Question section mismatch" in Dig.

**What you expected to happen**: The question section of the reply should be rewritten according to the rewrite rules, even if the response is an RcodeServerFailure.

**How to reproduce it (as minimally and precisely as possible)**: Configure CoreDNS with name and answer rewrites, and a subsequent plugin that causes RcodeServerFailure, then query via dig.

**Environment**:

- the version of CoreDNS: Commit f286a24b4e2c8027b9045b4fb36f92ecf5ed33f1

miekg:
Resign if the original zone's mtime is change in some way.

Closes #4406

### DIFF
--- a/plugin/sign/README.md
+++ b/plugin/sign/README.md
@@ -32,7 +32,7 @@ it do key or algorithm rollovers - it just signs.
     Both these dates are only checked on the SOA's signature(s). This concerns the DNSSEC data, the
     *sign* plugin will also take into account and resign if:
 
-     - the **mtime** of the zone has changed, since the last time it was checked.
+     - the **mtime** of the zone file has changed, since the last time it was checked.
 
      - the signed zone file doesn't exist on disk.
 
--- a/plugin/sign/README.md
+++ b/plugin/sign/README.md
@@ -7,9 +7,9 @@
 ## Description
 
 The *sign* plugin is used to sign (see RFC 6781) zones. In this process DNSSEC resource records are
-added. The signatures that sign the resource records sets have an expiration date, this means the
-signing process must be repeated before this expiration data is reached. Otherwise the zone's data
-will go BAD (RFC 4035, Section 5.5). The *sign* plugin takes care of this.
+added to the zone. The signatures that sign the resource records sets have an expiration date. This
+means the signing process must be repeated before this expiration data is reached. Otherwise the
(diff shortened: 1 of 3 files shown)


plugin/sign/README.md
```diff
@@ -29,7 +29,12 @@ it do key or algorithm rollovers - it just signs.
 
      -  the signature only has 14 days left before expiring.
 
-    Both these dates are only checked on the SOA's signature(s).
+    Both these dates are only checked on the SOA's signature(s). This concerns the DNSSEC data, the
+    *sign* plugin will also take into account and resign if:
+
+     - the **mtime** of the zone has changed, since the last time it was checked.
```
chrisohaver:
```suggestion
     - the **mtime** of the zone file has changed, since the last time it was checked.
```

chrisohaver (commented):
I recall that _file_ monitors for changes in the serial before reloading the zone tree.  Is resigning based on file mod time a potential issue if someone updates the zone file but forgets to update the serial?

chrisohaver:
> I recall that _file_ monitors for changes in the serial before reloading the zone tree. Is resigning based on file mod time a potential issue if someone updates the zone file but forgets to update the serial?

I suppose not.  It would just resign again when the file is modified to update the serial.

chrisohaver (approved):

miekg:
a better way is actually to copy what the file plugin does, check the SOA serial. That's less fragile and I like the similarity between file and sign

PR #4443, opened 2021-02-01
plugin/rewrite: copy msg before rewritting

miekg:
Copy the msg to prevent messing with the (via the pointer) original
created message that may be stored in the cache or anything other data
store.


yongtang (approved):
/lgtm

PR #4445, opened 2021-02-01
doh: set http request in writer

balboah:
### Why 

Makes it possible to read the current http request while serving DNS and in turn modify metadata or do other custom processing based on the HTTP request rather than the DNS message.

### DIFF
--- a/core/dnsserver/https.go
+++ b/core/dnsserver/https.go
@@ -2,6 +2,7 @@ package dnsserver
 
 import (
 	"net"
+	"net/http"
 
 	"github.com/coredns/coredns/plugin/pkg/nonwriter"
 )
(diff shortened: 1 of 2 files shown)


miekg:
why?

balboah:
This is kind of a follow up on #4329

Here is my plugin that needs to extract metadata from the HTTP request itself:

```go
func (id Identity) Metadata(ctx context.Context, state request.Request) context.Context {
	switch w := state.W.(type) {
	case *request.ScrubWriter:
		switch w := w.ResponseWriter.(type) {
		case *dnsserver.DoHWriter:
			path := w.Request().URL.Path
			if path == "/dns-query" {
				break
			}
			id := strings.TrimLeft(path, "/")
			metadata.SetValueFunc(ctx, name+"/device", func() string { return id })
		default:
		}
	default:
	}

	return ctx
}
```

The plugin sets metadata for what is in the request path to be used as an identifier.
I could not find another way to get to the HTTP request

miekg:
thanks that makes sense. Including this in the initial PR would have sped things up.

Can you send another PR that updates the comment on this struct as it now also adds the original http request?

Issue #2833, opened 2019-05-17
Proposal: split horizon and advanced routing thoughts


johnbelamaric:
# Advanced routing

I am thinking about how we can do split horizon and other advanced routing (e.g., metadata-based). I'd like to get into the routing logic somehow and redirect queries to the stanza I want (override the normal zone-based logic). Here are some ideas, please give some thoughts.

## Use Rewrite

We can *almost* do this already. If we had a conditional on the `rewrite` rules, we could rewrite to some fake intermediate domain, and then forward that back to ourselves. This is very hacky though.

## Firewall

The [firewall](https://github.com/coredns/policy) plugin has allow/refuse/block/drop actions now. We could add a `forward` action that forwards it to another DNS server, or back to ourselves.

```
  firewall query {
    forward dns://127.0.0.1:5432 client_ip =~ '10\.120\.0\..*'
  }
```

(As an aside, the expression engine here could be used with `rewrite` in the "Use Rewrite" option above, if we chose that).

(As another aside, we should add IP -> int conversion in [govaluate](https://github.com/Knetic/govaluate) or direct "in-cidr" operations).

## New Split Plugin

We can add a simple `split` plugin that allows forwarding based on common criteria, for example source IP. This is not mutually exclusive with `firewall`, but may be a little simpler for common cases.
```
example.com {
  split {
    source 10.0.0.0/8 dns://127.0.0.1:5533
    source 172.0.0.0/8 dns://127.0.0.1:5534
  }
  forward . /etc/resolv.conf
}

example.com:5533 {
  file internal-example.db
}

example.com:5534 {
  log
  file internal-example-2.db
}
```

## Named Stanzas

I don't like the separate servers on specific localhost ports. To alleviate that, we can allow name:// as a scheme. This also lets us keep these out of the normal routing. For example:

```
example.com {
  split {
    source 10.0.0.0/8 name://internal-example
    source 172.0.0.0/8 name://internal-example-2
  }
  firewall query {
    jump name://internal-example proto == "tcp"
  }
  forward example.com /etc/resolv.conf
}

name://internal-example {
  file internal-example.db
}

name://internal-example-2 {
  log
  file internal-example.db
}
```

@fturib suggested a `switch` plugin at some point that is I think a little like this - but I couldn't find the issue where he wrote it up.

chantra:
#2827 would actually be useful to add unittest to this :) cc @miekg

miekg:
Nice ideas.

[...]

PR #4446, opened 2021-02-01
plugin/forward: Extensible Forwarding Policies

chrisohaver:

This PR enables the _forward_ plugin to load forwarding policies defined in other plugins.  My thinking is that it would be useful for others that need to have forwarding policies which may not be appropriate for merging in tree.

An example of such a plugin: a conditional forwarding policy, https://github.com/chrisohaver/conditional

with the example conditional plugin: #2833

ToDo: document interface in readme.

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -41,7 +41,7 @@ type Forward struct {
 	expire        time.Duration
 	maxConcurrent int64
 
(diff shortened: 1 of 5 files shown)


miekg:
I can see some use, but I rather not complicate forward further, complexity is what killed the proxy plugin

PR #4451, opened 2021-02-05
make kubernetes plugin kubeconfig argument 'context' optional

answer1991:

Make kubernetes plugin kubeconfig argument 'context' optional.

This corefile should also work fine:
```bash
kubernetes coredns.local {
	kubeconfig file
}
```

None.

kubernetes plugin document.

None.

### DIFF
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -33,7 +33,7 @@ all the zones the plugin should be authoritative for.
 kubernetes [ZONES...] {
     endpoint URL
     tls CERT KEY CACERT
-    kubeconfig KUBECONFIG CONTEXT
+    kubeconfig KUBECONFIG [CONTEXT]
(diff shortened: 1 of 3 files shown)


miekg:
lgtm, but needs more text in the README. IIRC this now defaults to the first context (or whatever) in the kubeconfig?

answer1991:
> IIRC this now defaults to the first context (or whatever) in the kubeconfig?

Yes, current context could be set and specified in kubeconfig file, and the user or program could use kubeconfig without specified the context. For example, when we run `kubectl --kubeconfig file get pods` then we used the current context specified in kubeconfig file. And we can switch the kubeconfig current context use `kubectl config use-context`. 

> but needs more text in the README.

Let me update it.

chrisohaver (approved):
LGTM Thanks!

yongtang (approved):
lgtm

PR #4452, opened 2021-02-05
plugin/transfer: only allow outgoing axfr over tcp

### LINKED ISSUE #4450: Prevent AXFR over UDP
**What happened**:  When sending an AXFR request over UDP (which typical clients like `dig` prevents you from doing), CoreDNS will happily reply over UDP, which has the potential of generating big payloads and used as an amplification vector.

**What you expected to happen**: AXFR to be limited in size and force TC or possibly return NOTIMPL (like root AXFR server do for instance)

Using a modified [miekg/exdns 's q.o](https://gist.github.com/chantra/6e4a19e06ccbc2f2c35d650295e37e7c)

Running:
```
./coredns -dns.port=1053
```

With a Corefile:
```
$ cat Corefile
. {
 file somezonefile
    log
 transfer {
    to *
 }
}
``` 

using the modified q.go:
```
$ go run q.go -port 1053   @::1  AXFR
```

we can see that CoreDNS would reply to the AXFR request over UDP.

**Environment**:

- the version of CoreDNS:
```
CoreDNS-1.8.1
linux/amd64, go1.14.7, 03812bb1

```
- Corefile:
```
. {
 file somezonefile
    log
 transfer {
    to *
 }
}
```
Ubuntu 20.10

miekg:
Return refused when the query comes in over udp.
No need to add a new test case as the current crop needed to be changed
to use TCP.

Fixes: #4450

### DIFF
--- a/plugin/transfer/failed_write_test.go
+++ b/plugin/transfer/failed_write_test.go
@@ -20,7 +20,7 @@ func (w *badwriter) WriteMsg(res *dns.Msg) error { return fmt.Errorf("failed to 
 func TestWriteMessageFailed(t *testing.T) {
 	transfer := newTestTransfer()
 	ctx := context.TODO()
-	w := &badwriter{ResponseWriter: &test.ResponseWriter{}}
+	w := &badwriter{ResponseWriter: &test.ResponseWriter{TCP: true}}
 	m := &dns.Msg{}
 	m.SetAxfr("example.org.")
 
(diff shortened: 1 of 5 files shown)


SuperQ (approved):
LGTM

SuperQ:
Based on https://tools.ietf.org/html/rfc5936#section-4.2, seems legit.

PR #4453, opened 2021-02-06
Allow dependabot to check go dependencies

jlosito:

I wouldn't say this pull requested is *needed* per se. Rather it's a nice to have.

This change allows dependabot to check any Go dependency which this projects uses on a weekly basis and submit pull requests with version bumps in order to keep packages up-to-date.

https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates

None.

None.

No.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "gomod"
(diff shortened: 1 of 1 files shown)


miekg:
we already do this, see https://github.com/coredns/coredns/pull/4442

PR #4461, opened 2021-02-11
Add 1.8.2 notes

miekg:

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -18,7 +18,7 @@
 # Steps to release, first:
 #
 # 1. Up the version in coremain/version.go
-# 2. Do a make -f Makefile.doc
+# 2. Do a make -f Makefile.doc # This has been automated in GitHub, so you can probably skip this step
 # 3. go generate
 # 4. Send PR to get this merged.
 #
@@ -44,7 +44,7 @@ EMPTY:=
 SPACE:=$(EMPTY) $(EMPTY)
 COMMA:=$(EMPTY),$(EMPTY)
 
-ifeq (, $(shell command -v curl))
+ifeq (, $(shell which -v curl))
     $(error "No curl in $$PATH, please install")
 endif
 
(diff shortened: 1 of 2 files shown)


miekg:
also fixed a zsh-ism in checking for the curl command

Makefile.release
```diff
@@ -44,7 +44,7 @@ EMPTY:=
 SPACE:=$(EMPTY) $(EMPTY)
 COMMA:=$(EMPTY),$(EMPTY)
 
-ifeq (, $(shell command -v curl))
```
mrueg, line 47: Why moving back to which? `command -v` is posix https://pubs.opengroup.org/onlinepubs/009604499/utilities/command.html https://github.com/koalaman/shellcheck/wiki/SC2230

yongtang: On Ubuntu 20.04 with `command` I noticed the following errors:
```
$ make -f Makefile.release  tar
make: command: Command not found
Makefile.release:48: *** "No curl in $PATH, please install".  Stop.
```

yongtang:
@miekg `which -v` may not work on Ubuntu 20.04 as well:
```
$ which -v curl
Illegal option -v
Usage: /usr/bin/which [-a] args
```

Maybe we can simply go with `which curl`?

Makefile.release
mrueg: Okay, I don't have a ubuntu machine to test here. Then yes revert back to `which curl` as @yongtang , I was hoping that we could avoid relying on which.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Add 1.8.2 not..." ]
>@miekg which -v may not work on Ubuntu 20.04 as well:
>
>$ which -v curl
>Illegal option -v
>Usage: /usr/bin/which [-a] args
>
>Maybe we can simply go with which curl?

yes, that should be enough, and was what was there for the manifest-tool

yongtang (approved):
LGTM

PR #4462, opened 2021-02-12
A noop target to Makefile.release and run it in CI

miekg:
This tests the Makefile.release (at least on the CI) to be correct and
executes some of the init code. This is to catch problems early and
before we actually release and enter some half released state.

### DIFF
--- a/.github/workflows/go.test.yml
+++ b/.github/workflows/go.test.yml
@@ -65,3 +65,14 @@ jobs:
       run: |
         go install github.com/fatih/faillint
         ( cd test; go test -race ./... )
+
+  test-makefile-release:
+    name: Test Makefile.release
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


yongtang:
@miekg wondering if it makes sense to just re-use the `tar` or `docker-build` target for testing the Makefile.release on CI? That may provide better coverage of the Makefile.release (without triggering true release).

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] A noop target..." ]
>@miekg wondering if it makes sense to just re-use the tar or docker-build
>target for testing the Makefile.release on CI? That may provide better coverage
>of the Makefile.release (without triggering true release).

that's a good point, will make the change to run tar

yongtang (approved):
lgtm

PR #4463, opened 2021-02-12
CoreDNS "Views"

chrisohaver:

This proposal introduces a feature in the vein of BIND's views, where a server-block can hold configurations that control what queries are routed to it.

At a high level, it allows plugins to define filters that `dnsserver.ServeDNS()` will use to control how queries are to handlers.

In more detail ...

* It extends server Configs to hold a list of FilterFuncs (currently only one is stored - for cidr reverse zone filtering)
* It allows plugins to define a filterFunc (via a new interface `dnsserver.Viewer`), which is added to the FilterFuncs when initializing the server.
* It accomodates multiple configs per zone for configs have a filter function defined.  If a config has no filter, it must be the last config for that zone.
* CoreDNS will route a query to the server config with the longest matching zone AND with all defined FilterFuncs evaluating to true. If more than one server config is defined for the zone, the query is routed to the first one with all defined FilterFuncs evaluating to true. If no server configs meet the above criteria for a zone, then the next longest matching zone is attempted.

I created an example plugin that exposes this as user defined expression called _[serve](https://github.com/chrisohaver/conditional/tree/advanced-routing)_.
An exerpt from the [README](https://github.com/chrisohaver/conditional/blob/advanced-routing/README.md):

> ## Syntax
>```
>serve {
>    EXPRESSION
>}
>```
>
> * **EXPRESSION** - CoreDNS will not route incoming queries to the enclosing server block
>  if any **EXPRESSION** evaluates to false. Multiple **EXPRESSION** can be given on separate lines.
>  See the **Expression Syntax** section below for available variables and functions.
>
>CoreDNS will only route a query to a given server block if the query falls within the server block's
zone (per normal routing behavior), and all **EXPRESSION** listed in *serve* evaluate to true.

#2833

This approach may also help facilitate resolving cidr reverse zones issues such as #2779, where reverse zones defined by cidrs in the same octet boundary collide.


miekg:
views can be easily mimicked by running multiple coredns processes, there is no need to do this in process

chrisohaver:
> views can be easily mimicked by running multiple coredns processes, there is no need to do this in process

Multiple CoreDNS processes cannot emulate views without some mechanism for selectively routing queries to them.

miekg:
the routing is usually done by listening on different interfaces. Just like NSD decided this non-feature of BIND does not need to be replicated

chrisohaver:
> the routing is usually done by listening on different interfaces.  Just like NSD decided this non-feature of BIND does not need to be replicated

I think harkening to Bind Views is muddying the water.

_This_ PR enables the option for split DNS as well as other conditional serving options, which are not possible in CoreDNS now by listening on different interfaces.  E.g. #2833

johnbelamaric:
The most common use case I am familiar with is split horizon DNS. In that
case we need to differentiate between which instance to forward to based
upon the source IP CIDR, IIUC. Right now we only support routing on the
zone.

I suppose if the client resolvers are configured with an internal IP then
we can differentiate based on the inbound interface. But I certainly see
the desire for more flexibility in the routing. I wonder if we can refactor
so that the routing decision is pluggable. That would allow plugins that
alter routing.

chrisohaver:
Since this PR is in the form of an interface that is not implemented by any internal plugins yet, there is no concrete example in the PR, hence I included the example ~_conditional_~ _serve_ plugin to provide more concrete context.

With that in mind, ~I've updated the README in the example plugin to hopefully make it more clear what can be accomplished:  https://github.com/chrisohaver/conditional#coredns-advanced-routing~

I created a new branch in the example plugin repo that removes the "noise" of unrelated implementations, and renames the example implementation to `serve`, which  I think is more clear. https://github.com/chrisohaver/conditional/tree/advanced-routing

chrisohaver:
FWIW, I think I've got the reverse zone cidr coverage/collision issues resolved with this PR in conjunction with some fixes to `plugin.SplitHostPort`... but I think it's too off topic to include in this PR.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] CoreDNS "View..." ]
>FWIW, I think I've got the reverse zone cidr coverage/collision issues resolved

that would be a nice fix.

Still if you want advanced routing you need to write a DNS routing service I don't want
more complexity in the core bit of coredns.

chrisohaver:
> Still if you want advanced routing you need to write a DNS routing service I don't want more complexity in the core bit of coredns.

This PR off-loads that complexity to external plugins.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] CoreDNS "View..." ]
>    Still if you want advanced routing you need to write a DNS routing service
>    I don't want more complexity in the core bit of coredns.
>
>This PR off-loads that complexity to external plugins.

??????


  core/dnsserver/address.go      | 22 ++++++++++++++++++----
  core/dnsserver/config.go       |  7 ++++---
  core/dnsserver/onstartup.go    |  2 +-
  core/dnsserver/register.go     | 51 +++++++++++++++++++++++++++++++++++++++++----------
  core/dnsserver/server.go       | 90 +++++++++++++++++++++++++++++++++++++++++++++++++++++-------------------------------------
  core/dnsserver/server_grpc.go  |  8 +++++---
  core/dnsserver/server_https.go | 14 +++++++++-----
  core/dnsserver/server_tls.go   |  8 +++++---
  core/dnsserver/view.go         |  9 +++++++++

PR #4464, opened 2021-02-12
Enforce server side MaxUDPSize

### LINKED ISSUE #4449: Enforce server side MaxUDPSize
**What happened**: Currently, the maximum size of an answer is informed by the client's accepted Bufsize (or optionally clamped by the bufsize plugin)

The request UDP Size is only a signal of how much a client is willing to receive, but it should not be dictating how big an answer a server should send.
The server should be able to configure a MaxUDPSize value and the max answer size, for UDP should be:

```
max(MinUDPSize, min(ServerMaxUDPOutSize, client.UDPSize))
```
where client.UDPSize defaults to MinUDPSize if EDNS is not set.

When Scrubbing the reply:
https://github.com/coredns/coredns/blob/03812bb1e7ec6af384976bc4a9fb22f054dece89/request/request.go#L220-L221)
The message is truncate to request.Size() size. This size is computed by looking at the client's request accepted bufsize, and then normalize.
https://github.com/coredns/coredns/blob/03812bb1e7ec6af384976bc4a9fb22f054dece89/request/request.go#L167-L174

This has the effect that if a client send a Bufsize of 4k, the server will reply up to 4k.

I am aware of the `bufsize` plugin, but I would argue it does not do the job, albeit there may be ways to change it in a way that make it compatible.

For one, the `bufsize` plugin adds EDNS when there is no EDNS. This is wrong for 2 reasons IMO:
- if there were no EDNS in the request, we should not add EDNS in the reply ([RFC6891 Section 7](https://tools.ietf.org/html/rfc6891#section-7)
> Lack of presence of an OPT record in a request MUST be taken as an
   indication that the requestor does not implement any part of this
   specification and that the responder MUST NOT include an OPT record
   in its response.
- by default, this would mean that the client accepts max 512, for any value of bufsize > 512, we would now possibly return payload bigger than accepted by the client.

If EDNS is set, the plugin may allow for a bigger payload than accepted by the requester (say requester allow 512, and we clamp to 1024, 1024 is now going to be the payload size allowed).

**Proposed solution**:

`Scrub` and `Size` being request methods, we somehow need to set the maximum UDP size as part of the request struct.

A possible solution would be to have a new attribute in the `Request` struct, say `MaxUDPSize`. The `ScrubWriter` could then set that value to whatever the server policy is willing to do when created the request object: https://github.com/coredns/coredns/blob/7339235de1663cd2cabb4e1f90b18eabcb799a65/request/writer.go#L17

`bufsize` plugin could then set this value instead e.g `w.MaxUDPSize = uint16(bufsize)`

**Considerations**:

* this MaxUDPSize should probably have a sane default. I would be tempted to go with [DNSFlagDay 2020](https://dnsflagday.net/2020/) values, but for backward compatibility, it is probably better to use 4096?
* MaxUDPSize == 0 would need to be handled by whichever sane default previously chosen.
* We may be willing to have a separate size for IPv4 and IPv6
* If the intent of `bufsize` was to set the max UDP payload the server was willing to receive, then this would not work as expected.

** Further considerations**:

Setting the response EDNS size to the request EDNS size is still not correct, but I view this more as cosmetic than a practical problem (based on my use-case), but the responder EDNS bufsize should reflect what the server is willing to accept as a UDP payload sent their way, typically for UPDATE for instance.

[RFC6891 Section 6.2.4](https://tools.ietf.org/html/rfc6891#section-6.2.4):
> The responder's maximum payload size can change over time but can
   reasonably be expected to remain constant between two closely spaced
   sequential transactions, for example, an arbitrary QUERY used as a
   probe to discover a responder's maximum UDP payload size, followed
   immediately by an UPDATE that takes advantage of this size.  This is
   considered preferable to the outright use of TCP for oversized
   requests, if there is any reason to suspect that the responder
   implements EDNS, and if a request will not fit in the default
   512-byte payload size limit.
   
   So likely, when calling `SizeAndDo` https://github.com/coredns/coredns/blob/03812bb1e7ec6af384976bc4a9fb22f054dece89/request/request.go#L191, this value should probably also be set to a server controlled value, similar to this bug report. And it could also probably be handled by the bufsize plugin.
   I am happy to open a separate issue for that though.

chantra:
This patch allows the server to enforce a maximum size that they are
willing to send over UDP.
The server will honor the max UDP size specificed by the client if
possible, yet will cap this to request.MaxUDPSize (which defaults to
request.DefaultMaxUDPSize).
If EDNS0 is not set, the logic will default to a MaxSize of 0, which
will be normalized to dns.MinMsgSize for UDP, or dns.MaxMsgSize for TCP.

Fixes #4449

This PR allows an administrator to control the maximum size of a DNS message to be sent over UDP while honoring the max UDP size advertised by the client (via EDNS0) if possible.
Failing to do so, a client could coerce a CoreDNS server to send large answers that can be used in reflection/amplification attacks.
The maximum size of the answer is currently limited to the maximum size that a server has in its database.

#4449

The behaviour would need to be documented (I don't think the current behaviour is documented either). A separate patch, which I am happy to provide, is needed to allow configuring this MaxUDPSize.

If people were relying on enormous answers, this would probably be a change. Defaulting server max size to 4096 is probably going to cover base for almost all cases. Optimally, it could be defaulted to 1232 (dns flagday 2020 value), but that would probably change the behaviour of existing installs.

### DIFF
--- a/request/request.go
+++ b/request/request.go
@@ -10,6 +10,10 @@ import (
 	"github.com/miekg/dns"
 )
 
+// DefaultMaxUDPSize is the maximum size of a packet that we would send over UDP.
+// 1232 is based on [DNS Flag Day 2020](https://dnsflagday.net/2020/) value.
+const DefaultMaxUDPSize = uint16(1232)
+
 // Request contains some connection state and is useful in plugin.
 type Request struct {
 	Req *dns.Msg
@@ -17,6 +21,8 @@ type Request struct {
 
 	// Optional lowercased zone of this query.
 	Zone string
+	// The maximum UDP buffer size the server is willing to send out. 0 for default value
+	MaxUDPSize uint16
 
 	// Cache size after first call to Size or Do. If size is zero nothing has been cached yet.
 	// Both Size and Do set these values (and cache them).
@@ -35,7 +41,7 @@ type Request struct {
 // NewWithQuestion returns a new request based on the old, but with a new question
 // section in the request.
 func (r *Request) NewWithQuestion(name string, typ uint16) Request {
-	req1 := Request{W: r.W, Req: r.Req.Copy()}
+	req1 := Request{W: r.W, Req: r.Req.Copy(), MaxUDPSize: r.MaxUDPSize}
 	req1.Req.Question[0] = dns.Question{Name: dns.Fqdn(name), Qclass: dns.ClassINET, Qtype: typ}
 	return req1
 }
@@ -157,7 +163,9 @@ func (r *Request) Do() bool {
 // Len returns the length in bytes in the request.
 func (r *Request) Len() int { return r.Req.Len() }
 
-// Size returns if buffer size *advertised* in the requests OPT record.
+// Size returns the size that can be sent on the wire.
+// If buffer size is *advertised* in the requests OPT record, Size will be
+// min(advertised buffer size, maximum allowed by the server).
 // Or when the request was over TCP, we return the maximum allowed size of 64K.
 func (r *Request) Size() int {
 	if r.size != 0 {
@@ -165,9 +173,18 @@ func (r *Request) Size() int {
 	}
 
 	size := uint16(0)
+
 	if o := r.Req.IsEdns0(); o != nil {
 		r.do = o.Do()
-		size = o.UDPSize()
+		// client is providing EDNS0, make sure the provided bufsize
+		// is not greater than our MaxUDPSize
+		size = r.MaxUDPSize
+		if size == 0 {
+			size = DefaultMaxUDPSize
+		}
+		if size > o.UDPSize() {
+			size = o.UDPSize()
+		}
 	}
 
 	// normalize size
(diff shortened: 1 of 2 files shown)


request/request.go
```diff
@@ -10,13 +10,19 @@ import (
 	"github.com/miekg/dns"
 )
 
+// DefaultMaxUDPSize is the maximum size of a packet that we would send over UDP.
+// 1232 is based on [DNS Flag Day 2020](https://dnsflagday.net/2020/) value.
+const DefaultMaxUDPSize = uint16(1232)
```
chantra, line 15: this is up for debate. In #4449 I had mentioned in the "consideration" section:

> this MaxUDPSize should probably have a sane default. I would be tempted to go with DNSFlagDay 2020 values, but for backward compatibility, it is probably better to use 4096?

chantra:
@chrisohaver @johnbelamaric @miekg @stp-ip @yongtang 

Sorry for the nudge.
Is there any opinions about this patch? Any way to move this forward?

The patch is relatively small, all unittests were passing out of the box, which I suppose indicate a fair amount of the current assumptions are kept unchanged.
If there is any other ideas on how to solve this, I would be happy to hear them.

#4449 has more details on the issue and how to possibly solve/incorporate that in the server.

PR #4465, opened 2021-02-12
[request] Also clear `do` and `size`

chantra:
Those 2 attributes were not cleared as part of `Clear()` call.

Ensure that all request caches are cleared.
I am not sure those were intentionally left out of the current `Clear()` function. If they were, I am happy to change this in a code documentation update.

None

None

None


miekg (approved):

miekg:
think omitted by accident

PR #4471, opened 2021-02-16
Add `entries by domain name` metric to the file plugin.

s-matyukevich:

This PR adds a new metric to the file plugin. This metric reports the number of records per domain name.  We need this metric in our environment because we dynamically generate zone file and this metric is very useful for troubleshooting cases when the file wasn't generated correctly. 

File plugin readme was updated.

No

### DIFF
--- a/plugin/file/README.md
+++ b/plugin/file/README.md
@@ -36,6 +36,12 @@ file DBFILE [ZONES... ] {
 
 If you need outgoing zone transfers, take a look at the *transfer* plugin.
 
+## Metrics
+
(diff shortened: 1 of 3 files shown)


miekg:
>  https://github.com/coredns/coredns/pull/4471

sorry, but this is a metric cardinality explosion waiting to happen, so no.

Maybe a total counter of records per zone, but that can also lead to too many metrics.

If you're are afraid of not generating your zones correctly you need to tackle that in the
generation path are have something like the new ZONEMD record.

PR #4473, opened 2021-02-18
plugin/rewrite: streamline the ResponseRule handling.

mandelsoft:

The functionality of a response rule is now completely encapsulated behind
a `ResponseRule` interface. This significantly simplifies the complete
processing flow, it enables more flexible response handling and it is possible
to eliminate lots of state flags, ifs and switches.

Based on the new flexibility the pull request also enables to support a
response name rewrite for all name rewrite types.
To be compatible, an explicit `answer auto` option is added for this.

Using this feature it is also possible now to rewrite a complete sub domain
hierarchy to a single domain name combined with a correct answer rewrite (coredns#2389).

The tests have been improved to test the new cases but also lots of older untested cases.

#2389, #2202

The documentation of the rewrite plugin has been adapted to reflect the new answer option.

No, the former behaviour completely remains. It just adds a new answer option for the rewrite plugin that can be used to influence the behaviour of all name rewrite types.

### DIFF
--- a/plugin/rewrite/README.md
+++ b/plugin/rewrite/README.md
@@ -13,23 +13,36 @@ Rewrites are invisible to the client. There are simple rewrites (fast) and compl
 
 A simplified/easy-to-digest syntax for *rewrite* is...
 ~~~
-rewrite [continue|stop] FIELD [FROM TO|FROM TTL]
+rewrite [continue|stop] FIELD [TYPE] [(FROM TO)|TTL] [OPTIONS]
 ~~~
 
 * **FIELD** indicates what part of the request/response is being re-written.
 
    * `type` - the type field of the request will be rewritten. FROM/TO must be a DNS record type (`A`, `MX`, etc.);
 e.g., to rewrite ANY queries to HINFO, use `rewrite type ANY HINFO`.
-   * `class` - the class of the message will be rewritten. FROM/TO must be a DNS class type (`IN`, `CH`, or `HS`); e.g., to rewrite CH queries to IN use `rewrite class CH IN`.
    * `name` - the query name in the _request_ is rewritten; by default this is a full match of the
      name, e.g., `rewrite name example.net example.org`. Other match types are supported, see the **Name Field Rewrites** section below.
-   * `answer name` - the query name in the _response_ is rewritten.  This option has special restrictions and requirements, in particular it must always combined with a `name` rewrite.  See below in the **Response Rewrites** section.
-   *  `edns0` - an EDNS0 option can be appended to the request as described below in the **EDNS0 Options** section.
+   * `class` - the class of the message will be rewritten. FROM/TO must be a DNS class type (`IN`, `CH`, or `HS`); e.g., to rewrite CH queries to IN use `rewrite class CH IN`.
+   * `edns0` - an EDNS0 option can be appended to the request as described below in the **EDNS0 Options** section.
    * `ttl` - the TTL value in the _response_ is rewritten.
 
+* **TYPE** this optional element can be specified for a `name` or `ttl` field.
+  If not given type `exact` will be assumed. If options should be specified the
+  type must be given.
 * **FROM** is the name (exact, suffix, prefix, substring, or regex) or type to match
 * **TO** is the destination name or type to rewrite to
-* **TTL** is the number of seconds to set the TTL value to
+* **TTL** is the number of seconds to set the TTL value to (only for field `ttl`)
+
+* **OPTIONS**
+
+  for field `name` further options are possible controlling the response rewrites.
+  All name matching types support the following options
+
+     * `answer auto` - the names in the _response_ is rewritten in a best effort manner.
+     * `answer name FROM TO` - the query name in the _response_ is rewritten matching the from regex pattern.
+     * `answer value FROM TO` - the names in the _response_ is rewritten matching the from regex pattern.
+
+  See below in the **Response Rewrites** section for further details.
 
 If you specify multiple rules and an incoming query matches multiple rules, the rewrite
 will behave as follows:
@@ -49,7 +62,7 @@ client.
 The syntax for name rewriting is as follows:
 
 ```
-rewrite [continue|stop] name [exact|prefix|suffix|substring|regex] STRING STRING
+rewrite [continue|stop] name [exact|prefix|suffix|substring|regex] STRING STRING [OPTIONS]
 ```
 
 The match type, e.g., `exact`, `substring`, etc., triggers rewrite:
@@ -60,7 +73,8 @@ The match type, e.g., `exact`, `substring`, etc., triggers rewrite:
 * **suffix**: when the name ends with the matching string
 * **regex**: when the name in the question section of a request matches a regular expression
 
-If the match type is omitted, the `exact` match type is assumed.
+If the match type is omitted, the `exact` match type is assumed. If OPTIONS are
+given, the type must be specified.
 
 The following instruction allows rewriting names in the query that
 contain the substring `service.us-west-1.example.org`:
@@ -95,7 +109,8 @@ rewrite name suffix .schmoogle.com. .google.com.
 
 ### Response Rewrites
 
-When rewriting incoming DNS requests' names, CoreDNS re-writes the `QUESTION SECTION`
+When rewriting incoming DNS requests' names (field `name`), CoreDNS re-writes
+the `QUESTION SECTION`
 section of the requests. It may be necessary to rewrite the `ANSWER SECTION` of the
 requests, because some DNS resolvers treat mismatches between the `QUESTION SECTION`
 and `ANSWER SECTION` as a man-in-the-middle attack (MITM).
@@ -127,6 +142,35 @@ ftp.service.us-west-1.consul. 0    IN A    10.30.30.30
 
 The above is a mismatch between the question asked and the answer provided.
 
+There are three possibilities to specify an answer rewrite:
+- A rewrite can request a best effort answer rewrite by adding the option `answer auto`.
+- A rewrite may specify a dedicated regex based response name rewrite with the
+  `answer name FROM TO` option.
+- A regex based rewrite of record values like `CNAME`, `SRV`, etc, can be requested by
+  an `answer value FROM TO` option.
+
+Hereby FROM/TO follow the rules for the `regex` name rewrite syntax.
+
+#### Auto Response Name Rewrite
+
+The following configuration snippet allows for rewriting of the
+`ANSWER SECTION` according to the rewrite of the `QUESTION SECTION`:
+
+```
+    rewrite stop {
+        name suffix .coredns.rocks .service.consul answer auto
+    }
+```
+
+Any occurrence of the rewritten question in the answer is mapped
+back to the original value before the rewrite.
+
+Please note that answers for rewrites of type `exact` are always rewritten.
+For a `suffix` name rule `auto` leads to a reverse suffix response rewrite,
+exchanging FROM and TO from the rewrite request.
+
+#### Explicit Response Name Rewrite
+
 The following configuration snippet allows for rewriting of the
 `ANSWER SECTION`, provided that the `QUESTION SECTION` was rewritten:
 
@@ -151,9 +195,11 @@ ftp-us-west-1.coredns.rocks. 0    IN A    10.20.20.20
 ftp-us-west-1.coredns.rocks. 0    IN A    10.30.30.30
 ```
 
+#### Rewriting other Response Values
+
 It is also possible to rewrite other values returned in the DNS response records
 (e.g. the server names returned in `SRV` and `MX` records). This can be enabled by adding
-the `answer value` to a name regex rule as specified below. `answer value` takes a
+the `answer value FROM TO` option to a name rule as specified below. `answer value` takes a
 regular expression and a rewrite name as parameters and works in the same way as the
 `answer name` rule.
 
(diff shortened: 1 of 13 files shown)


plugin/rewrite/README.md
```diff
@@ -127,6 +128,28 @@ ftp.service.us-west-1.consul. 0    IN A    10.30.30.30
...
+```
+    rewrite stop {
+        name suffix .coredns.rocks .service.consul answer auto
+    }
+```
+
+Any occurrence of the rewritten question in the answer is mapped
+back to the original value before the rewrite.
+
+Please not that answers for rewrites of type `exact` are always rewritten.
```
chrisohaver:
```suggestion
Please note that answers for rewrites of type `exact` are always rewritten.
```

plugin/rewrite/name.go
```diff
@@ -58,62 +114,62 @@ const (
 
 // Rewrite rewrites the current request based upon exact match of the name
 // in the question section of the request.
```
miekg, line 116: this comment is not true anymore, becasuse the rule is returned instead of executed?

plugin/rewrite/name.go
```diff
@@ -58,62 +114,62 @@ const (
...
+		return nil, RewriteIgnored
 	}
 	s := rule.Replacement
 	for groupIndex, groupValue := range regexGroups {
 		groupIndexStr := "{" + strconv.Itoa(groupIndex) + "}"
 		s = strings.Replace(s, groupIndexStr, groupValue, -1)
 	}
 	state.Req.Question[0].Name = s
-	return RewriteDone
+	return rule.Response(state)
```
miekg: why does this need to be returned? `rule.Response(state)` can also be called outside of this method

mandelsoft: This is a dedicated implementation for the name rewrite types (the other types (can) do it differently), that need to dynamically handle the auto and non-auto case. I've separated this into this response function, because it is identical for all name types. `Response` is just an attribute, here. The interface function to be used from the outside is still the `Rewrite` method of a rule object, which returns a dedicated, possibly request specific response rule . Because of this feature to provide a request specific response rule, a `GetResponse()` method cannot be used for the rewrite rule type object. The rewrite object must have the full control depending on the rewrite request to provide a dedicated response rule.
Basically all the attributes should be internal(lower case), but they were public so so far, therefore I didn't changed this.

mandelsoft:
I get this strange error in the kubernetes tests

```
go test -v ./...
go: github.com/coredns/caddy@v1.1.0: missing go.sum entry; to add it:
	go mod download github.com/coredns/caddy
```

Does anybody know how to get rid of it. I've done the download and the entry is already in the go.sum file.

mandelsoft:
I get this strange error in the kubernetes tests

```
go test -v ./...
go: github.com/coredns/caddy@v1.1.0: missing go.sum entry; to add it:
	go mod download github.com/coredns/caddy
```

Does anybody know how to get rid of it? I've done the download and the entry is already in the go.sum file.

plugin/rewrite/name.go
mandelsoft: May be it is better to encapsulate the common behaviour of those name rules into a dedicated base struct used for the name rule types. I could provide this, It could also handle the `Mode()` function.

mandelsoft: I've changed this now and introduced a common base part, which allowed to get rid of even more coding

miekg:
>@mandelsoft commented on this pull request.
>
>
>In plugin/rewrite/name.go:
>
>>       }
>        s := rule.Replacement
>        for groupIndex, groupValue := range regexGroups {
>                groupIndexStr := "{" + strconv.Itoa(groupIndex) + "}"
>                s = strings.Replace(s, groupIndexStr, groupValue, -1)
>        }
>        state.Req.Question[0].Name = s
>-       return RewriteDone
>+       return rule.Response(state)
>
>I've changed this now and introduced a common base part, which allowed to get
>rid of even more coding

excellent. I'll take another look

plugin/rewrite/name.go
miekg: making members/functions private is OK w/ me. I think this was mostly a side effect of style. There should be nothing consuming this as an API

plugin/rewrite/name.go
```diff
@@ -140,45 +193,56 @@ func newNameRule(nextAction string, args ...string) (Rule, error) {
...
+				return nil, fmt.Errorf("answer name mapping only for regexp rule")
+			}
+			if strings.ToLower(args[4]) != "name" {
+				return nil, fmt.Errorf("exceeded the number of arguments for a regex name rule")
+			}
+		}
 	}
 
 	if len(args) < 7 {
+		_ = auto
```
miekg: why?

plugin/rewrite/name.go
```diff
@@ -140,45 +193,56 @@ func newNameRule(nextAction string, args ...string) (Rule, error) {
...
 		switch matchType {
 		case ExactMatch:
-			rewriteAnswerFromPattern, err := isValidRegexPattern(rewriteQuestionTo, rewriteQuestionFrom)
+			_, err := isValidRegexPattern(rewriteQuestionTo, rewriteQuestionFrom)
 			if err != nil {
```
miekg, line 393: this if can no be folded into the check, saving a line

plugin/rewrite/name_test.go
```diff
@@ -75,13 +75,244 @@ func TestRewriteNamePrefixSuffix(t *testing.T) {
...
+}
+
 func TestNewNameRule(t *testing.T) {
 	tests := []struct {
 		next         string
 		args         []string
 		expectedFail bool
 	}{
-		{"stop", []string{"exact", "srv3.coredns.rocks", "srv4.coredns.rocks"}, false},
+		{"stop", []string{"exact", "rv3.coredns.rocks", "srv4.coredns.rocks"}, false},
```
miekg: why this change?

plugin/rewrite/reverter.go
```diff
@@ -23,15 +25,16 @@ type ResponseRule struct {
...
 	dns.ResponseWriter
 	originalQuestion dns.Question
-	ResponseRewrite  bool
 	ResponseRules    []ResponseRule
+	noSave           bool
 }
 
 // NewResponseReverter returns a pointer to a new ResponseReverter.
-func NewResponseReverter(w dns.ResponseWriter, r *dns.Msg) *ResponseReverter {
+func NewResponseReverter(w dns.ResponseWriter, r *dns.Msg, noSaveQuestion bool) *ResponseReverter {
```
miekg:
```suggestion
func NewResponseReverter(w dns.ResponseWriter, r *dns.Msg, restoreQuestion bool) *ResponseReverter {
```

plugin/rewrite/reverter.go
```diff
@@ -23,15 +25,16 @@ type ResponseRule struct {
...
 	return &ResponseReverter{
 		ResponseWriter:   w,
 		originalQuestion: r.Question[0],
+		noSave:           noSaveQuestion,
```
miekg:
```suggestion
		restore:           restoreQuestion,
```

plugin/rewrite/reverter.go
```diff
@@ -40,42 +43,13 @@ func (r *ResponseReverter) WriteMsg(res1 *dns.Msg) error {
 	// Deep copy 'res' as to not (e.g). rewrite a message that's also stored in the cache.
 	res := res1.Copy()
 
-	res.Question[0] = r.originalQuestion
-	if r.ResponseRewrite {
+	if !r.noSave {
```
miekg:
```suggestion
	if !r.restore {
```

plugin/rewrite/rewrite.go
```diff
@@ -33,34 +33,29 @@ const (
 type Rewrite struct {
 	Next     plugin.Handler
 	Rules    []Rule
-	noRevert bool
+	noRevert bool // for test
```
miekg: oh members that _are_ only used for testing, that sounds iffy. Can this be done in another why. I would very much not have this done this way

plugin/rewrite/ttl.go
```diff
@@ -9,81 +9,90 @@ import (
...
+
+func (r *ttlResponseRule) RewriteResponse(rr dns.RR) {
+	rr.Header().Ttl = r.TTL
+}
+
 type exactTTLRule struct {
 	NextAction string
 	From       string
-	ResponseRule
+	response   *ttlResponseRule
```
miekg: why do these need to be pointers now? I think added a NoopRule can also remove the need to make `nil` the thing that conveys 'noop'. Would also make this PR much smaller, 'cause these changes can all be reverted

plugin/rewrite/name.go
mandelsoft: sorry , an artefact from testing, I'll remove it

plugin/rewrite/reverter.go
mandelsoft: Better world be `noRestore`, the default (false) should be the standard behavior, which is restoring the original question in the response handling. For testing purposes this will then be set to true.

plugin/rewrite/rewrite.go
mandelsoft: The `noRevert` was already there, it has not been introduced by this PR. I just added another such field, and added the comment telling the reason for this field. I just found no other usage for the already existing field, besides various test cases.

plugin/rewrite/ttl.go
mandelsoft: I'll use the same pattern with a `ttlRuleBase` as for the name rules, which simplifies the complete rule implementations.

plugin/rewrite/name_test.go
mandelsoft: not intended, sorry, I'll revert to the original one

mandelsoft:
@miekg I've introduced a base rule for all the ttl rules, also, with a common response handling. This simplifies the whole rewrite and response handling. Additionally I icorprated all your remarks, but instead of using the field name `Restore` I uses `noRestore` because the default behavior should be selected by the default field value (false).

I've added the field comment`// for test` because this field (`noRevert` was not introduced by this PR) was only used for test purposes, so far. The same is true for the `noRestore` field I added. If you see any usecase for package users we should make these field public. As far as they are package internal, they are only used for testing. 

If you don't like such testing fields in non-testing code, do you have a better idea? There could be a package local function enabling the test mode, but this would prevent parallel tests and the conscious decision for a dedicated test case to explicitly avoid those steps to check intermediate processing step results. Another possibility would be to officially introduce a policy struct, which controls the revert and restoration behaviour. Nevertheless it is basically only used for test purposes. 

If you think this is better, I can do it on short notice.

mandelsoft:
Unfortunately the rewrite plugin has been changed in the meantime. Those changes are heavily conflicting, so I've to redo the changes.

mandelsoft:
@miekg I've rebased and extended the auto rewrite by the additional value rewrite features. Therefore I aligned the value rewrite coding with the new restructuring of the rewrite plugin. Thereby a simplified and more flexible option parsing has been introduced to support all the variants. Additionally response rewrite rules are possible now for all name rewrites.

mandelsoft:
@miekg Is there still something I can do?

miekg:
>@miekg Is there still something I can do?

I'm sorry I didn't have any time to review this week (and probably also this weekend). So
hopefully Monday

.github/workflows/go.test.yml
```diff
@@ -79,4 +79,4 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Test Makefile
-      run: make -f Makefile.release build
+      run:  make DOCKER=bla GITHUB_ACCESS_TOKEN=bla -n build docker github-push docker-push -f Makefile.release
```
miekg: is this a rebase gone wrong?

man/coredns-rewrite.7
```diff
@@ -18,7 +18,7 @@ A simplified/easy-to-digest syntax for \fIrewrite\fP is...
 .RS
 
 .nf
-rewrite [continue|stop] FIELD [FROM TO|FROM TTL]
+rewrite [continue|stop] FIELD [TYPE] [(FROM TO)|TTL] [OPTIONS]
```
miekg: revert any changes to this file, will be done automatically from a github workflow

man/coredns-sign.7
```diff
@@ -8,9 +8,9 @@
 .SH "DESCRIPTION"
```
miekg: please revert all these changes

plugin/rewrite/name.go
```diff
@@ -9,40 +9,117 @@ import (
...
+
+	"github.com/miekg/dns"
 )
 
-type exactNameRule struct {
-	NextAction string
-	From       string
-	To         string
-	ResponseRule
+////////////////////////////////////////////////////////////////////////////////
```
miekg: please remove these `////` blocks, it clutters the godoc

plugin/rewrite/rewrite.go
```diff
@@ -31,37 +31,29 @@ const (
 
 // Rewrite is a plugin to rewrite requests internally before being handled.
 type Rewrite struct {
-	Next     plugin.Handler
-	Rules    []Rule
-	noRevert bool
+	Next      plugin.Handler
+	Rules     []Rule
+	noRevert  bool // for test
```
miekg: could you add a TODO here, that we want to get rid of these 'used for testing members' in this struct.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] plugin/rewrit..." ]
***@***.*** I've rebased and extended the auto rewrite by the additional value
>rewrite features. Therefore I aligned the value rewrite coding with the new
>restructuring of the rewrite plugin. Thereby a simplified and more flexible
>option parsing has been introduced to support all the variants. Additionally
>response rewrite rules are possible now for all name rewrites.

yep, I think this is very much a wanted clean up and improvement of the code, thanks for
you're continued work on it.

Also saw the rebase going a bit haywire you have no picked up a couple of other changes.

One real comment on the member in the Rewrite struct, just used for testing (add a TODO
there), but otherwise I see no reason to delay this any further.

Cheers!

.github/workflows/go.test.yml
mandelsoft: I haven't touched the Makefile, but there is an additional change in the meantime. I just did a new rebase.

plugin/rewrite/rewrite.go
mandelsoft: I could propose a general `RevertPolicy` instead, that can officially be set using the API.
Please give me a trigger to push that proposal. So far, I just added the TODO.

mandelsoft:
@miekg I've replaced the revert options now by a RevertPolicy. This is a commit on top of the rest. So we can just discard it if you feel more comfortable with the TODO comment

mandelsoft:
Hi @miekg , could we bring the PR somehow to an end?

miekg (approved):

PR #4481, opened 2021-02-23
Fix go-test yaml

miekg:
unsure why this only pops up, but the yaml was not correct for this
workflow. Randomly added whitespace until it parsed as proper yaml

### DIFF
--- a/.github/workflows/go.test.yml
+++ b/.github/workflows/go.test.yml
@@ -71,7 +71,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-   - name: Install dependencies
+    - name: Install dependencies
       run: |
         sudo apt-get install make curl
 


miekg:
k8s tests now start failing w/
~~~
go: github.com/coredns/caddy@v1.1.0: missing go.sum entry; to add it:
	go mod download github.com/coredns/caddy
~~~
executing that yields
~~~
% go mod download github.com/coredns/caddy     
go mod download: skipping argument github.com/coredns/caddy that resolves to the main module
~~~

miekg:
ok, `go.sum` is not checked in coredns/ci.. what

yongtang:
@miekg I tried locally and it seems if `github.com/coredns/caddy` is clone in `/go/src/github.com/coredns/caddy` location then the build will go through. Though this is part of the `coredns/ci` script and I am still trying to find out which place will need to be updated on ci script. /cc @chrisohaver

miekg:
The ci lacks a go.sum, it's not checked in if I saw that right

chrisohaver:
> The ci lacks a go.sum, it's not checked in if I saw that right

I don't think that should be a problem. go.sum is always re-built at build time.

yongtang:
@chrisohaver The golang version now is 1.16 which removes the rebuild. See
https://blog.golang.org/go116-module-changes

I think clone github.com/coredns/caddy on the CI machine before invoke `go build .` will solve the issue.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] Fix go-test y..." ]
>    The ci lacks a go.sum, it's not checked in if I saw that right
>
>I don't think that should be a problem. go.sum is always re-built at build
>time.

that's def. the problem. I'll continue to ignore this test for now

chrisohaver:
> @chrisohaver The golang version now is 1.16 which removes the rebuild. See
https://blog.golang.org/go116-module-changes

Thanks, @yongtang ... that's a significant change.

In coredns/ci, go.mod points coredns to a local directory so it can test variable versions of coredns.   Prior to 1.16, go would automatically update go.mod dependencies, and generate go.sum, both of which will change depending on the dependencies of the coredns code base being tested.  In that regard, relying on a static go.sum would only work if it was generated with a depency-identical version of coredns being tested.

Fiddling locally, it looks like inserting a `go mod download` before `go test` will update the go.mod dependencies to be compatible with the local version of coredns, and generate go.sum, enabling the test to compile.

PR #4485, opened 2021-02-24
release: add 1.8.3 release notes

miekg:


miekg:
carbon copy of 1.8.2 with only the fix makefile PR added

PR #4487, opened 2021-02-24
Fix coredns/ci kubernetes test failure

chrisohaver:

With go 1.16, reconciliation of module dependencies is no longer automatic with build/test.
This adds that manual step before the ci test, which _should_ let the test compile successfully.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,6 +56,7 @@ jobs:
           name: Run Kubernetes tests
           command: |
             cd ~/go/src/${CIRCLE_PROJECT_USERNAME}/ci/test/kubernetes
+            go mod download
             go test -v ./...
 
 workflows:


miekg:
why can't you just check in the go.sum file instead of fiddling here?

chrisohaver:
> why can't you just check in the go.sum file instead of fiddling here?

The version of coredns in the test is dynamic (pulled from the PR), which dynamically changes the module dependencies and the sum, so it needs to be refigured each test.

miekg:
so, it should do both and at least have a go.sum there, and we should document this as well in the both repos

chrisohaver:
> so, it should do both and at least have a go.sum there, and we should document this as well in the both repos

Since coredns/ci's go.mod points to a local directory, a checked-in go.sum is meaningless.  It only reflects the go.sum with whatever the state of the local coredns directory was for last person who commits it.

PR #4488, opened 2021-02-24
Makefile.release: do a complete dry-run

miekg:
Just dry-run all important targets.


miekg:
@yongtang this is a much better (and faster) method and it can now also check the other targets, that bit me when releasing 1.8.2

yongtang (approved):
/lgtm

PR #4490, opened 2021-02-25
plugin/kubernetes: don't return when ServerVersion return an error

miekg:
When err=nil try to determine the version of the k8s cluster and disable
endpoint slices. Don't return from connecting to the cluster.

In the future we should just default to true, and delete all this code.

### DIFF
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -252,15 +252,14 @@ func (k *Kubernetes) InitKubeCache(ctx context.Context) (err error) {
 	}
 	// Disable use of endpoint slices for k8s versions 1.18 and earlier. Endpoint slices were
 	// introduced in 1.17 but EndpointSliceMirroring was not added until 1.19.
-	sv, err := kubeClient.ServerVersion()
-	if err != nil {
-		return err
-	}
-	major, _ := strconv.Atoi(sv.Major)
-	minor, _ := strconv.Atoi(strings.TrimRight(sv.Minor, "+"))
-	if k.opts.useEndpointSlices && major <= 1 && minor <= 18 {
-		log.Info("watching Endpoints instead of EndpointSlices in k8s versions < 1.19")
-		k.opts.useEndpointSlices = false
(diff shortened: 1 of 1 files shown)


chrisohaver:
I think this will be a problem if the API is not available at coredns start time for 1.18 and older k8s clusters.  It would use endpointslices, which will not work.  So the decision would need to be deferred.

miekg:
@johnbelamaric merge failed because the k8s e2e wasn't successful (different issue)

miekg:
that's the entire problem with version detection and determining what you should do; it's a bad idea.

not starting because the api is not there is much more grave

johnbelamaric:
Oh, @chrisohaver I see your point. I haven't looked closely at how we "wait and retry"; seems like this code should execute when the "retry" is successful.

chrisohaver:
> Oh, @chrisohaver I see your point. I haven't looked closely at how we "wait and retry"; seems like this code should execute when the "retry" is successful.

Yeah - the decision on which endpoint object we should watch should be made in the retry - currently it is not.

chrisohaver:
> I think this will be a problem if the API is not available at coredns start time for 1.18 and older k8s clusters. It would use endpointslices, which will not work. So the decision would need to be deferred.

Actually - the issue is the opposite of that... It will default to watching endpoints instead of slices if the API isnt up... which I think would cause problems on k8s 1.19 or greater.  I think endpoints are mirrored to endpointslices, but not the other way around... so we would not pick up all endpoints (for which the source of truth in k8s 19+ is endpointslices, not endpoints).

johnbelamaric:
Would it be easier to just periodically check and enable the slice watching
if it's available? Kinda stupid extra processing, but I haven't looked at
the code or how difficult it would be to do it in the right place.

chrisohaver:
> Would it be easier to just periodically check and enable the slice watching if it's available?

No, I think the logic can be moved into a retry loop. I'm looking into it now.

chrisohaver:
That said - the alternative is to default to always use endpoint slices.  But with coredns's single release branch, that would leave k8s 1.18 and older users behind.

PR #4338, opened 2020-12-15
Fix NPE isssue in plugin kubernetes 

standup-jb:
### 1. Why is this pull request needed and what does it do?
I faced the npe issue . the error log as below。I fixed this issue . add error check before use sv .
`github.com/coredns/coredns/plugin/kubernetes.(*Kubernetes).InitKubeCache(0xc0004a4480, 0x21f3940, 0xc0001080d8, 0x0, 0x1b856c0)
	/Users/xmly/go/src/coredns/plugin/kubernetes/kubernetes.go:256 +0x22d
github.com/coredns/coredns/plugin/kubernetes.setup(0xc000250cf0, 0x3, 0x1da9502)
	/Users/xmly/go/src/coredns/plugin/kubernetes/setup.go:42 +0x190
github.com/coredns/caddy.executeDirectives(0xc000048800, 0x7ffff3e08991, 0x15, 0x30a00e0, 0x2e, 0x2e, 0xc0002c1420, 0x1, 0x1, 0x0, ...)
	/Users/xmly/go/pkg/mod/github.com/coredns/caddy@v1.1.0/caddy.go:656 +0x69e
github.com/coredns/caddy.ValidateAndExecuteDirectives(0x21ebf80, 0xc0000e2640, 0xc000048800, 0x0, 0x1, 0xc000206100)
	/Users/xmly/go/pkg/mod/github.com/coredns/caddy@v1.1.0/caddy.go:612 +0x421
github.com/coredns/caddy.startWithListenerFds(0x21ebf80, 0xc0000e2640, 0xc000048800, 0x0, 0x0, 0x0)
	/Users/xmly/go/pkg/mod/github.com/coredns/caddy@v1.1.0/caddy.go:515 +0x13c
github.com/coredns/caddy.Start(0x21ebf80, 0xc0000e2640, 0x21ebf80, 0xc0000e2640, 0x0)
	/Users/xmly/go/pkg/mod/github.com/coredns/caddy@v1.1.0/caddy.go:472 +0xea
github.com/coredns/coredns/coremain.Run()
	/Users/xmly/go/src/coredns/coremain/run.go:78 +0x27c
main.main()
	/Users/xmly/go/src/coredns/coredns.go:12 +0x20`

PR #4492, opened 2021-02-25
plugin/kubernetes: do endpoint/slice check in retry loop

chrisohaver:

 If the API server is not available when CoreDNS starts, it may [not select the correct endpoint object type to watch](https://github.com/coredns/coredns/pull/4490#issuecomment-786072526).  When the wrong endpoint type is watched, DNS records for endpoints and headless services will be incomplete/missing.

This PR does the following:
- Moves the endpointslice check logic into a retry loop.
- Puts the code into a separate function so it's easier to identify and remove when the time comes.
- Documents reasoning and assumptions for the version check.

#4490 #4338

### DIFF
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -18,10 +18,10 @@ import (
 	"github.com/coredns/coredns/plugin/pkg/fall"
 	"github.com/coredns/coredns/plugin/pkg/upstream"
 	"github.com/coredns/coredns/request"
-	discovery "k8s.io/api/discovery/v1beta1"
 
 	"github.com/miekg/dns"
 	api "k8s.io/api/core/v1"
+	discovery "k8s.io/api/discovery/v1beta1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -6,9 +6,9 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"strconv"
 	"strings"
 
+	"github.com/coredns/caddy"
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/etcd/msg"
 	"github.com/coredns/coredns/plugin/kubernetes/object"
@@ -19,7 +19,6 @@ import (
 
 	"github.com/miekg/dns"
 	api "k8s.io/api/core/v1"
-	discovery "k8s.io/api/discovery/v1beta1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
@@ -213,7 +212,7 @@ func (k *Kubernetes) getClientConfig() (*rest.Config, error) {
 }
 
 // InitKubeCache initializes a new Kubernetes cache.
-func (k *Kubernetes) InitKubeCache(ctx context.Context) (err error) {
+func (k *Kubernetes) InitKubeCache(ctx context.Context, c *caddy.Controller) (err error) {
 	config, err := k.getClientConfig()
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


plugin/kubernetes/setup.go
```diff
@@ -60,10 +60,45 @@ func setup(c *caddy.Controller) error {
...
+			}
+			// Disable use of endpoint slices for k8s versions 1.18 and earlier. The Endpointslices API was enabled
+			// by default in 1.17 but Service -> Pod proxy continued to use Endpoints by default until 1.19.
+			// DNS results should be built from the same source data that the proxy uses.  This decision assumes
+			// k8s EndpointSliceProxying featuregate is at the default (i.e. only on for k8s >= 1.19).
+			major, _ := strconv.Atoi(sv.Major)
+			minor, _ := strconv.Atoi(strings.TrimRight(sv.Minor, "+"))
+			if k.opts.useEndpointSlices && major <= 1 && minor <= 18 {
+				log.Info("Watching Endpoints instead of EndpointSlices in k8s versions < 1.19")
+				k.opts.useEndpointSlices = false
```
miekg: this is now a data race on k.opts.useEndpointSlices

plugin/kubernetes/setup.go
```diff
@@ -60,10 +60,45 @@ func setup(c *caddy.Controller) error {
 	return nil
 }
 
+// selectEndpointType will select which endpoint object type to watch (endpointslices or endpoints)
+// based on the supportability of endpointslices in the API and server version.
+// If the API supports discovery v1 beta1, and the server versions >= 1.19, endpointslices will be watched.
+// This function should be removed, along with non-slice endpoint watch code, when support for k8s < 1.19 is dropped.
+func (k *Kubernetes) selectEndpointType(kubeClient *kubernetes.Clientset) {
+	ticker := time.NewTicker(100 * time.Millisecond)
```
miekg: ticker should be stopped when leaving the function

plugin/kubernetes/setup.go
```diff
@@ -60,10 +60,45 @@ func setup(c *caddy.Controller) error {
...
+				k.opts.useEndpointSlices = false
+			}
+			return
+		}
+	}
+}
+
 // RegisterKubeCache registers KubeCache start and stop functions with Caddy
-func (k *Kubernetes) RegisterKubeCache(c *caddy.Controller) {
+func (k *Kubernetes) RegisterKubeCache(c *caddy.Controller, kubeClient *kubernetes.Clientset) {
```
miekg: better to return a startup function that to keep passing caddy.Controllers

chrisohaver:
I reworked the code with the frame of mind that watching EndpointSlices is normal behavior, and watching Endpoints is a special exception. `newdnsController` now will always define an EndpointSlice lister and controller.
Later, during coredns start up, but before starting the api controllers, we check with the api server to see if we need to watch Endpoints instead, and replace controllers if necessary.  In the future this step can be removed, and `newdnsController` can remain as is.

chrisohaver:
If there is reluctance to implement endpointslice API support checks in a retry loop  ...

An alternative is to expose endpointslice use as a configurable option, and leave it to the user to configure which to use (defaulting to endpointslice).  This is more adaptable to custom k8s deployments, since a cluster can disable parts of endpointslice support via featuregate, which we don't currently detect (and I think difficult to detect dynamically).  It removes all k8s version checking from the code, but would require users of k8s < 1.19 to explicitly set the option. It adds yet-another-option to kubernetes, but the option could be deprecated later when we no longer support api.Endpoints.

Thoughts?

miekg:
I'm in favor of checking for the users and not pushing this down to them. We also introducing an option which then needs to be removed again with all the config pain associated with that.

plugin/kubernetes/kubernetes.go
```diff
@@ -246,25 +247,78 @@ func (k *Kubernetes) InitKubeCache(ctx context.Context) (err error) {
...
+// should be watched, and false when endpoints should be watched.
+// If the API supports discovery v1 beta1, and the server versions >= 1.19, endpointslices are watched.
+// This function should be removed, along with non-slice endpoint watch code, when support for k8s < 1.19 is dropped.
+func (k *Kubernetes) endpointSliceSupported(kubeClient *kubernetes.Clientset) bool {
+	useEndpointSlices := false
+	ticker := time.NewTicker(100 * time.Millisecond)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
```
miekg, line 301: if we know what to do we can stop this loop, no?

miekg: oh we do, vnm

plugin/kubernetes/kubernetes.go
```diff
@@ -246,25 +247,78 @@ func (k *Kubernetes) InitKubeCache(ctx context.Context) (err error) {
...
+		ticker := time.NewTicker(100 * time.Millisecond)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
+				if k.APIConn.HasSynced() {
+					return nil
+				}
+			case <-timeout:
+				return nil
```
miekg, line 278: log.Debug/log.Warning ?

miekg:
and some documentation updates (although it's getting a long README), but still wouldn't hurt

chrisohaver:
Looking at the code with fresh eyes - I see that this introduces a race condition on `epController`. 

edit: ... and `epLister`

plugin/kubernetes/kubernetes.go
chrisohaver: Trying to be concise with `starting server with unsynced Kubernetes API` ...

It's hard to be concise about the repercussions... which are that we continue to attempt to connect and sync in the background.  And until the API completely syncs for all watched objects, we will answer with SERVFAIL for any kubernetes query that does not resolve to a known record.

miekg (approved):
I think this looks OK

chrisohaver:
Added sections to read me to describe the (pre-existing) 5 second start up delay behavior, and the endpointslice/endpoint monitor behavior.

I'd like to test this more before merging especially after the locking changes in ab47772, though I have not had time to focus on this in the past week.

PR #4501, opened 2021-03-03
Add support for CIDR range upto /24

zamog:
This pull requests add support for CIDR range upto /24
[#2779 ](https://github.com/coredns/coredns/issues/2779)
None.
No.

### DIFF
--- a/plugin/kubernetes/setup.go
+++ b/plugin/kubernetes/setup.go
@@ -121,7 +121,17 @@ func ParseStanza(c *caddy.Controller) (*Kubernetes, error) {
 	zones := c.RemainingArgs()
 
 	if len(zones) != 0 {
-		k8s.Zones = zones
+		var classfulZones []string
+		for i := 0; i < len(zones); i++ {
+			tmp, err := plugin.ClassfulFromCIDR(zones[i])
+			if err == nil {
+				classfulZones = append(classfulZones, tmp...)
+
+			} else {
+				classfulZones = append(classfulZones, zones[i])
+			}
+		}
+		k8s.Zones = classfulZones
 		for i := 0; i < len(k8s.Zones); i++ {
 			k8s.Zones[i] = plugin.Host(k8s.Zones[i]).Normalize()
 		}
--- a/plugin/normalize.go
+++ b/plugin/normalize.go
@@ -166,24 +166,12 @@ func ClassfulFromCIDR(s string) ([]string, error) {
 	var cidrs []string
 	if err == nil {
 		ones, _ := n.Mask.Size()
-		if len(n.IP) == net.IPv6len {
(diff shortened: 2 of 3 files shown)


miekg:
fixing this would be swell. Why does this need big.Int? This is just some bit shifting?

zamog:
> fixing this would be swell. Why does this need big.Int? This is just some bit shifting?
For no good reason, updated to use bitwise - commit 25a3af8

miekg:
btw isn't there a neat go package that just does this?

zamog:
> btw isn't there a neat go package that just does this?

didn't find one that has exactly the needed functionality. there are some that implement full cidr calculator like:
https://github.com/apparentlymart/go-cidr - use big int
https://github.com/dspinhirne/netaddr-go - ipv4/ipv6 calculations
but i think it's an overkill to import them for this limited functionality

miekg:
fair enough, let me review

plugin/kubernetes/setup.go
```diff
@@ -121,7 +121,17 @@ func ParseStanza(c *caddy.Controller) (*Kubernetes, error) {
...
+		for i := 0; i < len(zones); i++ {
+			tmp, err := plugin.ClassfulFromCIDR(zones[i])
+			if err == nil {
+				classfulZones = append(classfulZones, tmp...)
+
+			} else {
+				classfulZones = append(classfulZones, zones[i])
+			}
+		}
+		k8s.Zones = classfulZones
```
miekg, line 134: if this requires plugin changes it's a no-go. If we can't do it on the server level than the whole 'reverse-gen' feature will be removed. As in: nice idea, didn't work too bad

plugin/normalize.go
```diff
@@ -141,3 +142,109 @@ func SplitHostPort(s string) (host, port string, ipnet *net.IPNet, err error) {
...
+			cidrs = append(cidrs, networks[i].String())
+		}
+	}
+	return cidrs, err
+}
+
+// NextSubnet returns the next available subnet of the desired mask size
+// starting for the maximum IP of the offset subnet
+// If the IP exceeds the maxium IP then the second return value is true
+func NextSubnet(network *net.IPNet, prefixLen int) (*net.IPNet, bool) {
```
miekg, line 186: looks copied from https://github.com/apparentlymart/go-cidr/blob/master/cidr/cidr.go#L188

zamog: Yes (added a note in the first commit #4501 11a688de5957b505c3b1d74eb09c8cd128ba4390) - seems like it drops after the change

plugin/kubernetes/setup.go
zamog: No problem to implement it on the server

miekg:
See #4538

PR #4504, opened 2021-03-04
feat: add docker source label

viceice:
This allows [renovate](/renovatebot/renovate) to link to source repo and fetch changelogs (if available)

not checked

no

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM debian:stable-slim
 
+LABEL org.opencontainers.image.source="https://github.com/coredns/coredns"
+
 RUN apt-get update && apt-get -uy upgrade
 RUN apt-get -y install ca-certificates && update-ca-certificates
 


yongtang:
@viceice Is this label a standard used in similar projects like Kubernetes/etc?

viceice:
Yes, official opencontainers label.

https://www.github.com/opencontainers/image-spec/tree/master/annotations.md

miekg:
there are many more labels, just adding a link to the repo doesn't seem like it adds much. From memory git hash and other fluff like it can also be added.

I don't think this adds much, and if you need it you can easily build your own pipeline and add labels

viceice:
I know, do you like me to add more labels?

We use it at renovate, reading it from the registry to find the source of the image. That way renovate can fetch release note from source repo to show in it's update pr's.

We also use it to link to the source from the pr.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] feat: add doc..." ]
>I know, do you like me to add more labels?
>
>We use it at renovate, reading it from the registry to find the source of the
>image. That way renovate can fetch release note from source repo to show in
>it's update pr's.
>
>We also use it to link to the source from the pr.

no, I don't think we should add any labels. The whole process you descibe seems to come
down to not having to google to view the changelog.

yongtang:
@viceice Honestly I would prefer not to add a specific label unless the label is used by a majority or a significant portion of the users. You can always build your own docker image with pre-built CoreDNS static binary. The effort is with very minimal.

viceice:
Ok, no problem.

viceice:
Our hope was to add those label to more images, so it will be adopted by more and more images over time.

PR #4505, opened 2021-03-04
Use latest Go version

miekg:
Update go.mod to 1.16 and make the workflows default to the latest
available.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/coredns/coredns
 
-go 1.13
+go 1.16
 
 require (
 	github.com/Azure/azure-sdk-for-go v40.6.0+incompatible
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/coredns/coredns
 
-go 1.13
+go 1.16
```
SuperQ, line 3: This is a "minimum version" spec. This should only be updated if there is code that won't compile on older versions of Go.

miekg:
and there we go
~~~
# github.com/fatih/faillint
Error: ../../../go/pkg/mod/github.com/fatih/faillint@v1.7.0/main.go:25:15: undefined: io.Discard
note: module requires Go 1.16
~~~

SuperQ (approved):
LGTM

Issue #4382, opened 2021-01-11
CoreDNS plugin to watch Kubernetes CRD for stub-domains


christianang:
**What would you like to be added**:

A CoreDNS plugin that watches a Kubernetes CRD that can configure a stub-domain in CoreDNS.

My strawdog proposal for a CRD is:

```yaml
---
apiVersion: coredns.addons.x-k8s.io/v1alpha1 (???)
kind: DNSZone
metadata:
  name: my-company.internal
  namespace: kube-system
spec:
  zoneName: my-company.internal
  forwardTo: 10.100.1.1
```

**Why is this needed**:

Currently, CoreDNS serves as the Kubernetes cluster DNS server for all applications running on the Kubernetes cluster. We would like to provide the ability for multiple collaborators (controllers or humans) to add upstream DNS resolvers for custom DNS zones so that applications running on the Kubernetes cluster can resolve custom domains.

While stubdomains and upstream nameservers can be configured with the forward plugin in CoreDNS, the Corefile for the Kubernetes cluster DNS is a global resource. This makes it difficult for a custom controller to add its own configuration to the Corefile. A custom controller might have to assume no one else would change the Corefile and overwrite any outside changes, which is not great because it reduces flexibility. Alternatively, a controller could have logic to determine what parts it added, and therefore can change, which isn’t impossible, but complex and potentially brittle e.g a human operator cannot tell if a part of the Corefile was added by a controller and they attempt to change it.

I’m interested in building this, but I’d like some feedback on if this was built would this be something we can get in-tree. The reason I’m interested in having this in-tree is because it removes the complexity of building my own version of coredns and replacing the image in `kube-system`. It also feels like a generally useful feature for Kubernetes that streamlines the process of adding a stub-domain. I’m also interested in thoughts on the structure of the CRD.

Additional context [here](https://github.com/kubernetes-sigs/cluster-addons/issues/88), I originally brought the idea up in the coredns-operator, but the idea that it may belong in CoreDNS itself was brought up.

Potentially relevant issue #1999 because it also suggests adding support for watching CRDs.

chrisohaver:
It would be cool if we could do a generic CRD watching plugin that passes off configurations to other plugins.  i.e. `<suspend disbelief>` CRD plugin detects change, processes it and sees it's a new domain for forward, so it informs the forward plugin of the new domain; forward plugin then updates its configuration`</suspend disbelief>`

[...]

PR #4512, opened 2021-03-09
Initial implementation of ForwardCRD plugin

christianang:

This PR is the initial implementation of the KubernetesCRD plugin that sets out to implement #4382. Namely, it allows CoreDNS to watch a DNSZone CRD to configure stub-domains for Kubernetes clusters.

Edit: This PR has been updated from just the readme to the initial implementation of the plugin.

#4382


miekg:
very interesting idea. What the data model for the zone data? And how do you envision lookup taking weird DNS stuff into account, like wildcards and in-zone CNAMEs? You want to depend on `file.Lookup` ? Would not be a bad idea to make the interface suitable for different backends (long standing idea in another issue somewhere)

christianang:
Currently I'm thinking this would handle the very narrow case of forwarding dns messages to other dns servers based on the zone name (basically implementing stub-domains, but through crds). Not currently thinking of handling any individual records for a zone. I imagine handling requests at that point would be similar to how CoreDNS handles dns mux-ing now for server blocks. E.g there will be a map of zone names that map to forward plugins and we would loop through the map looking for a match, popping off subdomains of the zone until a match is found and we call that forward plugin's `ServeDNS` function. I think this sidesteps a lot of dns complexity, but let me know if I'm overlooking something.

I can envision that this plugin can become more flexible, through additional crd(s). Not sure what this might look like yet, since there are probably a few ways to slice it so starting with a pretty limited crd.

christianang:
I've updated the PR to include my initial implementation of the plugin. Apologies for what is now a massive PR, it felt like it made sense to continue iterating on the single PR, but let me know if you prefer breaking this down in any way. I did try and keep commits focused, which may help digest the PR.

Would appreciate feedback, I imagine being new to this codebase there were design decisions I made that might not be the best approach for a given problem, so let me know if you spot ways this can be improved.

One design decision that might be worth pointing out, I added a function to the forward plugin `NewWithConfig` that allows the `kubernetescrd` plugin to create instances of forward on-demand. This duplicated some code, which could be "DRY"-ed up, but in favor of not making too far reaching changes to the code base I left it duplicated and am happy with making follow up PRs to de-duplicate code if the approach I used makes sense.

plugin/kubernetescrd/README.md
```diff
@@ -0,0 +1,161 @@
+# kubernetescrd
```
rosenhouse: Small note from reviewing this README:  maybe a few words or a section about security?

A malicious user could create these resources in order to steal traffic (as they could if they had write access to the `coredns` configmap in `kube-system`

May be worth recommending that users limit the namespace to `kube-system`, perhaps show that example first with some language recommending that?  Also worth noting that when touching on the RBAC topic, since RBAC on writes to the resource could be another way of locking down the security.

christianang: Good point. I've added a small section to readme wrt to this.

core/dnsserver/zdirectives.go
```diff
@@ -46,6 +46,7 @@ var Directives = []string{
 	"clouddns",
 	"k8s_external",
 	"kubernetes",
+	"kubernetescrd",
```
chrisohaver: The position of this plugin in the chain is driven by the DNS function performed, which for now is just forwarding.  If the plugin is expanded to expose other functions via CRDs, then picking a single position in the plugin chain that is appropriate for all functions may prove difficult/impossible. IOW, this plugin may need to remain specific to stub domain forwarding.

We might want to order this immediately before _forward_. In its current position, it executes before _file_, whereas _forward_ normally executes after _file_.  I think we should try to make the behavior be as close as possible to using the _forward_ plugin.

However, the typical way of configuring a stub domain is to define a serverblock for the zone.  A separate serverblock creates a plugin chain for that zone, allowing for different configurations per stubdomain  (e.g. cache settings).  Routing all the zones from a single point isn't quite as flexible. There is currently no way for plugins to "create" new server blocks.  I think it would be possible to modify core/dnsserver to enable that, but it could get complicated.

plugin/forward/forward.go
```diff
@@ -53,12 +56,112 @@ type Forward struct {
...
+
 // New returns a new Forward.
 func New() *Forward {
 	f := &Forward{maxfails: 2, tlsConfig: new(tls.Config), expire: defaultExpire, p: new(random), from: ".", hcInterval: hcInterval, opts: options{forceTCP: false, preferUDP: false, hcRecursionDesired: true}}
 	return f
 }
 
+// NewWithConfig returns a new Forward configured by the provided
+// ForwardConfig.
+func NewWithConfig(config ForwardConfig) (*Forward, error) {
```
chrisohaver, line 88: I think I like this direction.  I presume forward's existing Corefile parsing functions will also be refactored to create a `ForwardConfig`, and then use `NewWithConfig` to create the `Forward` struct.  This also opens the way for other plugins to reuse the "forwarding" function of the forward plugin (e.g. adding a forwarding action to coredns/policy).

@miekg, WDYT?

plugin/kubernetescrd/controller.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+		forwardConfig := forward.ForwardConfig{
+			From: dnsZone.Spec.ZoneName,
+			To:   []string{dnsZone.Spec.ForwardTo},
+		}
+		plugin, err := d.instancer(forwardConfig)
+		if err != nil {
+			return err
+		}
+		err = plugin.OnStartup()
```
chrisohaver: This doesnt capture the dnstap integration (from forward setup).  Not sure if it's achievable from this point.

chrisohaver (commented):
Thanks @christianang! Lots to digest, but from what I understand, it looks good so far.

core/dnsserver/zdirectives.go
chrisohaver: Another difference between routing queries through separate serverblocks vs single point will surface in metrics, since server scope metrics have a zone label of the enclosing serverblock.  When routing queries through one point, the resulting metrics would all be labeled with the same zone (the zone of the serverblock enclosing the plugin).

chrisohaver: > If the plugin is expanded to expose other functions via CRDs, then picking a single position in the plugin chain that is appropriate for all functions may prove difficult/impossible.

Regarding the above: This may be solved by instead of having the crd-monitoring plugin serving DNS requests, the other plugins did.  For example, if forward could handle multiple zones, learn of added/deleted zones from the crd-monitoring plugin while running, and start/stop serving those zones.  In this way, dns requests would be served by the forward plugin itself.  Other plugins could be adapted similarly, e.g. hosts, for crd controlled A/AAAA records.  The hosts and forward plugins require different positions in the plugin chain (hosts near the top, and forward near the bottom).

Note: this doesn't address the aforementioned "serverblock" differences.

christianang: > The position of this plugin in the chain is driven by the DNS function performed, which for now is just forwarding. If the plugin is expanded to expose other functions via CRDs, then picking a single position in the plugin chain that is appropriate for all functions may prove difficult/impossible. IOW, this plugin may need to remain specific to stub domain forwarding.

Positioning in the chain is an interesting point I didn't realize. I was optimistically aiming for this plugin to encapsulate other functions via CRDs, but it does seem difficult now. Perhaps it is best to rename this plugin if you think it is unlikely we can encapsulate other functionality into this plugin (maybe `forwardingcrd` or something)? It does feel like it may get unwieldy if there becomes a proliferation of plugins to implement more functionality via CRDs.

> We might want to order this immediately before forward. In its current position, it executes before file, whereas forward normally executes after file. I think we should try to make the behavior be as close as possible to using the forward plugin.

Makes sense, and I can make this change.

> However, the typical way of configuring a stub domain is to define a serverblock for the zone. A separate serverblock creates a plugin chain for that zone, allowing for different configurations per stubdomain (e.g. cache settings). Routing all the zones from a single point isn't quite as flexible. There is currently no way for plugins to "create" new server blocks. I think it would be possible to modify core/dnsserver to enable that, but it could get complicated.

Yeah, the current implementation is not very flexible in this regard. I like the idea of this plugin "creating" new server blocks, but it does seem very complex and I did not want to dig too deep into modifying core/dnsserver code on the initial implementation. I could see the plugin being refactored to use such functionality and perhaps the CRD at that point could allow a user to specify what plugins they want to enable and those could be translated into server blocks. Perhaps creating new server blocks would also helps with the plugin chain ordering problem and the metrics problem?

> Regarding the above: This may be solved by instead of having the crd-monitoring plugin serving DNS requests, the other plugins did. For example, if forward could handle multiple zones, learn of added/deleted zones from the crd-monitoring plugin while running, and start/stop serving those zones. In this way, dns requests would be served by the forward plugin itself. Other plugins could be adapted similarly, e.g. hosts, for crd controlled A/AAAA records. The hosts and forward plugins require different positions in the plugin chain (hosts near the top, and forward near the bottom).

That makes sense, but one thing I'm unclear on, do instances of plugins just exist that I can modify (assuming we gave that plugin the functionality to be externally modified), or would the crd-monitoring plugin still have to instantiate instances of the plugin (and would that mean it isn't part of the proper plugin chain)?

plugin/kubernetescrd/controller.go
christianang: Good catch, but what makes you say you don't think it is achievable? I think I can save off the `dnstap` plugin within setup and set the `tapPlugin` on `forward.NewWithConfig`, but let me know if there is something I'm missing.

core/dnsserver/zdirectives.go
chrisohaver: > Perhaps creating new server blocks would also helps with the plugin chain ordering problem and the metrics problem?

Yes - it would essentially mimic the current way stub zones are configured.  So, metrics would get labeled in the correct zone.  Plugin chain ordering would not matter, since each zone gets its own chain.

I've been looking into how this could be implemented, and what it looks like.  After a few hours fiddling with it, it can confirm, it _is_ complicated - and requires some (perhaps unacceptable) hacks.  ATM, I'm thinking that my tinkering wont produce something feasible, and we'll just document the differences in behavior between CRD stub zones and manually defined ones.

> That makes sense, but one thing I'm unclear on, do instances of plugins just exist that I can modify (assuming we gave that plugin the functionality to be externally modified), or would the crd-monitoring plugin still have to instantiate instances of the plugin (and would that mean it isn't part of the proper plugin chain)?

I was thinking we assume they exist.  Dynamically inserting another plugin into the chain while maintaining the correct execution order would almost impossible I think.  Assuming the plugin exists means we'd probably want the dns-serving plugins to be declared in a blank slate state (to later be configured via the crd interface). But in the CRD stub zone forward case, an instance of forward typically pre-exists (the default upstream).

The crd plugin would not sit in the plugin chain. It would not have a `ServeDNS()` function defined (wont implement a plugin Handler), and it would not be added to the Config in setup().  There are examples of this kind of "configuration" type plugin, e.g. *sign*

Another approach entirely could be to create "crd" versions of each plugin we want to CRD-ify... i.e. *forward-crd* , *hosts-crd*, ... and package some of the shared crd monitoring components to reduce code duplication. These crd plugins could be placed next to (just before, just after?) their non-crd counterparts. *forward-crd* essentially akin to the implementation in this PR.  This minimizes ordering angst, but i think there may still be a little WRT the before/after question (i.e. which one "wins" in the event of zone overlap).

plugin/kubernetescrd/controller.go
chrisohaver: Oh it could be - I didn't scratch below the surface ...  just saying I wasn't sure if it was or not.

christianang: Gotcha, I implemented dnstap and tested it locally.

core/dnsserver/zdirectives.go
christianang: Any preference on which implementation is preferable? Not sure if I lean towards any particular implementation since they both seem to have their own tradeoffs. Configuring a separate plugin means you have to add both plugins to the chain, which could lead to misconfiguration, but solves the ordering issue. CRD-ifying plugins solves adding both plugins to the chain, but there is now an ordering issue with their non-crd counterparts.

chrisohaver: > Any preference on which implementation is preferable?

I like the "crd plugin configures other plugins" solution because it mimics the behavior of having had configured those plugins via the Corefile, and it results in fewer plugins down the road.
However, I like the "forward-crd plugin serves dns itself via an instance of forward" because it would enable other plugins to leverage the forwarding capability of the forward plugin.

I'm leaning toward "forward-crd plugin serves dns itself via an instance of forward" ... since i'm on the fence, and you've more or less already done it this way.

christianang: Sounds good. I've done a quick rename from `kubernetescrd` to `forwardcrd`. I think that is all I might need to do at this point, but let me know if I've missed anything.

plugin/forwardcrd/README.md
```diff
@@ -0,0 +1,173 @@
+# forwardcrd
+
+## Name
+
+*forwardcrd* - enables proxying DNS messages to upstream resolvers by reading
+CRDs from a Kubernetes cluster
```
chrisohaver:
```suggestion
the `DNSZone` CRD from a Kubernetes cluster
```

plugin/forwardcrd/README.md
```diff
@@ -0,0 +1,173 @@
...
+  **Note**: It is recommended to limit the namespace (e.g to `kube-system`)
+  because this can be potentially misused. It is ideal to keep the level of
+  write access similar to the `coredns` configmap in the `kube-system`
+  namespace.
+
+## Ready
+
+This plugin reports readiness to the ready plugin. This will happen after it has
+synced to the Kubernetes API.
+
```
chrisohaver, line 69: Should add a section (here or elsewhere) describing how this works with and differs from the Corefile serverblock defined stub domains, and a forward plugin in the same block that covers the following points...
* when DNSZone CRDs and serverblock defined stub domains are used, domains defined in the Corefile take precedence (in the event of zone overlap). e.g. if the domain `example.com` is defined in the corefile as a stub domain, and a DNSZone CRD record defined for `sub.example.com`, then `sub.example.com` would get forwarded to the upstream defined in the corefile, not the DNSZone.
* metrics difference: DNSZone metrics all labeled in single zone (the zone of the enclosing serverblock).
* When using forwardcrd and forward in the same serverblock, DNSZones take precedence over forward plugin defined in the same serverblock. e.g. if a DNSZone is defined for `.`, then no queries would be forwarded to the upstream defined the forward plugin of the same serverblock.

precedence is serverblock -> forwardcrd -> forward

chrisohaver (commented):
> I added a function to the forward plugin NewWithConfig that allows the kubernetescrd plugin to create instances of forward on-demand. This duplicated some code, which could be "DRY"-ed up, but in favor of not making too far reaching changes to the code base I left it duplicated and am happy with making follow up PRs to de-duplicate code if the approach I used makes sense.

I think it makes sense to work that out now and include it in this PR.

plugin/forwardcrd/README.md
```diff
@@ -0,0 +1,173 @@
...
+  names for remote k8s connection.  This option is ignored if connecting
+  in-cluster (i.e. endpoint is not specified).
+* `kubeconfig` **KUBECONFIG [CONTEXT]** authenticates the connection to a remote
+  k8s cluster using a kubeconfig file.  **[CONTEXT]** is optional, if not set,
+  then the current context specified in kubeconfig will be used.  It supports
+  TLS, username and password, or token-based authentication.  This option is
+  ignored if connecting in-cluster (i.e., the endpoint is not specified).
+* `namespace` **NAMESPACE** only reads `DNSZone` resources from the namespace
+  listed. If this option is omitted then it will read from all namespaces.
+  **Note**: It is recommended to limit the namespace (e.g to `kube-system`)
```
chrisohaver: Ideally we should default to the recommended configuration.  e.g. Default to only reading CRDs from `kube-system`, then perhaps allowing just `namespace` without any namespaces listed mean "all namespaces".

christianang:
Something that occurs to me is there could be slight confusion that the plugin is called `forwardcrd`, but the actual CRD is named `DNSZone`. I wonder if there are any thoughts on me renaming the CRD to be called `Forward`?

plugin/forwardcrd/README.md
christianang: Makes sense. Added to the readme.

plugin/forwardcrd/README.md
christianang: Makes sense.

christianang:
Updated the PR with updates based on your feedback. I also did the refactor to forward setup so that is resolved as well.

plugin/forwardcrd/README.md
```diff
@@ -0,0 +1,206 @@
...
+synced to the Kubernetes API.
+
+## Ordering
+
+Coredns has following precedence:
+Corefile Server Block -> `forwardcrd` plugin -> `forward` plugin.
+
+When `DNSZone` CRDs and Server Blocks define stub domains that are used,
+domains defined in the Corefile take precedence (in the event of zone overlap).
+in the Corefile take precedence (in the event of zone overlap). e.g. if the
```
chrisohaver:
```suggestion
e.g. if the
```

plugin/forwardcrd/README.md
```diff
@@ -0,0 +1,206 @@
...
+  configmap in the `kube-system` namespace.
+
+## Ready
+
+This plugin reports readiness to the ready plugin. This will happen after it has
+synced to the Kubernetes API.
+
+## Ordering
+
+Coredns has following precedence:
```
chrisohaver:
```suggestion
Forward behavior can be defined in three ways, via a Server Block, via the *forwardcrd* plugin, and via the *forward* plugin.  If more than one of these methods is employed and a query falls within the zone of more than one, CoreDNS selects which one to use based on the following precedence:
```

plugin/forwardcrd/README.md
```diff
@@ -0,0 +1,206 @@
...
+Corefile Server Block -> `forwardcrd` plugin -> `forward` plugin.
```
chrisohaver:
```suggestion
Corefile Server Block -> *forwardcrd* plugin -> *forward* plugin.
```

chrisohaver: To be consistent with other plugin documentation, *emphasize* plugin names.

chrisohaver (commented):
LGTM!  Suggested some fixes and tweaks to docs.

chrisohaver:
> Something that occurs to me is there could be slight confusion that the plugin is called `forwardcrd`, but the actual CRD is named `DNSZone`. I wonder if there are any thoughts on me renaming the CRD to be called `Forward`?

Sounds reasonable.  `StubDomain` could also work - carrying forward the name Kubernetes uses to describe the feature.

christianang:
I ended up leaning towards renaming it to `Forward` for consistency with the plugin name.

plugin/forwardcrd/README.md
```diff
@@ -0,0 +1,207 @@
...
+domains defined in the Corefile take precedence (in the event of zone overlap).
+e.g. if the
+domain `example.com` is defined in the Corefile as a stub domain, and a
+`Forward` CRD record defined for `sub.example.com`, then `sub.example.com` would
+get forwarded to the upstream defined in the Corefile, not the `Forward` CRD.
+
+When using `forwardcrd` and `forward` in the same Server Block, `Forward` CRDs
+take precedence over the `forward` plugin defined in the same Server Block.
+e.g. if a `Forward` resource is defined for `.`, then no queries would be
+forwarded to the upstream defined the `forward` plugin of the same Server Block.
```
chrisohaver:
```suggestion
When using *forwardcrd* and *forward* plugins in the same Server Block, `Forward` CRDs
take precedence over the *forward* plugin defined in the same Server Block.
e.g. if a `Forward` CRD is defined for `.`, then no queries would be
forwarded to the upstream defined in the *forward* plugin of the same Server Block.
```

plugin/forwardcrd/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Forward resource
+
+Apply the `Forward` CRD to your Kubernetes cluster.
+
+```
+kubectl apply -f ./manifests/crds/coredns.io_forwards.yaml
+```
+
+Assuming the **forwardcrd** plugin has been configured to allow `Forward`
```
chrisohaver:
```suggestion
Assuming the *forwardcrd* plugin has been configured to allow `Forward`
```

plugin/forwardcrd/README.md
```diff
@@ -0,0 +1,207 @@
...
+take precedence over the *forward* plugin defined in the same Server Block.
+e.g. if a `Forward` CRD is defined for `.`, then no queries would be
+forwarded to the upstream defined in the *forward* plugin of the same Server Block.
+
+## Metrics
+
+`Forward` CRD metrics are all labeled in a single zone (the zone of the enclosing
+Server Block).
+
+## Examples
```
chrisohaver, line 104: A real-world example of how forwardcrd would be used in an actual CoreDNS Corefile as deployed in a kubernetes cluster would be helpful.

plugin.cfg
```diff
@@ -61,6 +61,7 @@ secondary:secondary
 etcd:etcd
 loop:loop
 forward:forward
+forwardcrd:forwardcrd
```
chrisohaver, line 65: _forwardcrd_ should come before _forward_. Otherwise, a default upstream forwarder, such as the one in the default K8s file `forward . /etc/resolv.conf`, would prevent _forwardcrd_ in the same server block from serving any queries.

chrisohaver (approved):

christianang:
Thanks for approving @chrisohaver. Can I ask what next steps are for merging? Let me know if there is anything else I should do to help this along.

chrisohaver:
@miekg, This LGTM. Are you OK to merge?

ohbus (approved):

plugin/forward/forward.go
v-byte-cpu: @chrisohaver I love this idea and implemented a similar solution in https://github.com/coredns/coredns/pull/4583. Anyway config parsing logic should be separated from plugin logic according to the single-responsibility principle. It will not only decouple components but also allow to thoroughly cover the parsing functionality with tests.

christianang:
I've rebased this PR on master. It looks like the major change I had to account for was the change to `Normalize` returning an array (#4621). When a forward crd is created and the zone is normalized, it only [uses the first element](https://github.com/coredns/coredns/pull/4512/files#diff-03444ae608d8d9e40e535b1a558123fdee46e72e62c051c72f3571c2104c687eR35), which seems to be similar to what the [forward plugin itself is doing](https://github.com/coredns/coredns/pull/4512/files#diff-2067160fadb0b4ccfe74a9fe1cb1d849527835d14d1ef4d732ce2be13e9ae796R90) because we don't expect the `from` for the forward plugin to ever be a non-octet reverse zone, but correct me if I'm wrong.

chrisohaver:
>... because we don't expect the `from` for the forward plugin to ever be a non-octet reverse zone, but correct me if I'm wrong.

Yes for now, but eventually this should be supported.  In #4639, I'm proposing that we warn users that are using such CIDRs until support is added.

christianang:
That makes sense. Made another rebase to carry over the warning to this PR.

christianang:
I've rebased my changes on master and also reduced the commits to make it easier to comb through.

@miekg if you have some time to review I'd appreciate it.

christianang:
Hi, revisiting this PR since its been some time. @miekg do you think you can give this a review? I would appreciate it.

chrisohaver:
@christianang, can you add this plugin and yourself as the owner, in the CODEOWNERS file?

christianang:
Done!

chrisohaver:
@miekg, is it OK with you if we merge this new plugin without your review, or would you like to review it first?

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Initial imple..." ]
***@***.***, if it OK with you if we merge this new plugin without your review, or
>would you like to review it first?

I don't have the time to thorougly review this. If I can find time for deeper coredns
reviews it would be focussed on some long standing PRs in the *file* plugin.

miekg:
ok, wsn't expediting that,. Think this is perfectly fine as an external plugin. Never saw the use-case why this needs to be in coredns proper

chrisohaver:
> ok, wsn't expediting that,. Think this is perfectly fine as an external plugin. Never saw the use-case why this needs to be in coredns proper

Sorry - I misinterpreted your response as an OK.  I'll revert it.

chrisohaver:
@christianang, sorry for the mixup. If you have time, please [register the plugin as an external plugin on coredns.io](https://coredns.io/2017/07/23/add-external-plugins/) ...

chrisohaver:
> ...  please register the plugin as an external plugin on coredns.io ...

Sorry - its been so long since I had looked at the code... quickly scanning it over, I remember now that this plugin required some refactoring of the _forward_ plugin... so it cannot stand as-is as an external plugin unless we also create a new external plugin from a copy of the forward plugin with those modifications ... or we merge the mods to the _forward_ plugin.

johnbelamaric:
Alternatives to consider, that (maybe) won't require changes to any existing plugins:

1) Operator that consumes CRDs and rewrites the ConfigMap of the Corefile;
2) New input flag(s) that read read directly from k8s CRDs to create a special CaddyfileInput a-la this example: https://github.com/coredns/learning-coredns/blob/master/dnscached/main.go (may require using CoreDNS as a library rather than directly)
3) Some sort of tweak to coredns/caddyfile.Dispenser to construct a Corefile out of CRD input (similar to (2) but different approach internally

Of course (1) requires zero changes to CoreDNS.

chrisohaver:
1 was originally suggested, but I had suggested a coredns plugin idea. Sorry folks!

christianang:
Sad to hear. I'll hit pause on making any attempts at this for now since (as @chrisohaver brought up), this does require a refactor to the forward plugin so it can be used by the newly introduced crd plugin so just making it an external plugin will take some effort.

I am open to continuing discussion on the feature and how best to implement if someone wants to continue (perhaps it makes sense to reopen and continue on #4382?), but for now I'll put down the effort.

PR #4526, opened 2021-03-17
workflow: sign commits

miekg:
This doesn't do much, but at least it doesn't litter PRs with non-signed
commits. This should keep the DCO clean(er)


miekg:
we probably may want to move all of these into a schedule that does 1 daily commit (if needed)

greenpau (approved):

SuperQ:
LGTM, +1 to consolidated auto-commits.

PR #4528, opened 2021-03-17
Use distroless base and also run coredns as nonroot.

vinayakankugoyal:
To run coredns as nonroot using the nonroot distroless base image.

nope

### How tested:
1. I created an alternative Dockerfile as shown below. This dockerfile uses debian:stable-slim so that I could exec into the container and run /bin/bash to see if the coredns is running as non-root and what capabilities it has on it.
```
FROM --platform=linux/$BUILDARCH k8s.gcr.io/build-image/setcap:buster-v1.4.0
COPY coredns /coredns
# We apply cap_net_bind_service so that coredns can be run as
# non-root and still listen on port less than 1024
RUN setcap cap_net_bind_service=+ep /coredns


FROM debian:stable-slim

RUN apt-get update && apt-get -uy upgrade
RUN apt-get -y install ca-certificates && update-ca-certificates
RUN apt-get -y install procps libcap2-bin

COPY --from=0 /coredns /coredns

USER 1000:1000
EXPOSE 53 53/udp
ENTRYPOINT ["/coredns"]
```
2. I built the image 
```
docker buildx build -t gcr.io/vinaygo/coredns --platform linux/amd64 --load .
```
3.  Ran the image
```
docker run -d gcr.io/vinaygo/coredns
```
4. Let's see what the container is
```
docker container ls                                                      
CONTAINER ID   IMAGE                    COMMAND                  CREATED          STATUS          PORTS                       NAMES
cdcc8a0a825b   gcr.io/vinaygo/coredns   "/coredns"               33 minutes ago   Up 33 minutes   53/tcp, 53/udp              dazzling_ardinghelli
```
5. Exec into that container
```
docker exec -it dazzling_ardinghelli  /bin/bash 
```
6.  lets see whats running
```
I have no name!@cdcc8a0a825b:/$ ps -aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
1000           1  0.4  0.0 748648 39428 ?        Ssl  15:56   0:00 /coredns
1000          20  0.6  0.0   3872  3268 pts/0    Ss   15:56   0:00 /bin/bash
1000          27  0.0  0.0   7644  2688 pts/0    R+   15:57   0:00 ps -aux
```
7. lets see what capabilities does coredns have
```
I have no name!@cdcc8a0a825b:/$ getcap coredns 
coredns = cap_net_bind_service+ep
```

Questions you might have:
Why do we need buildx?
buildx preserves xargs across build stages. Meaning that if we apply capabilities to a binary using setcap in one buildstage and then copy that binary to another then the capabilities stay intact.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,11 @@
-FROM debian:stable-slim
+FROM --platform=linux/$BUILDARCH k8s.gcr.io/build-image/setcap:buster-v1.4.0 AS setcap
+COPY coredns /coredns
+# We apply cap_net_bind_service so that coredns can be run as
+# non-root and still listen on port less than 1024
+RUN setcap cap_net_bind_service=+ep /coredns
 
-RUN apt-get update && apt-get -uy upgrade
-RUN apt-get -y install ca-certificates && update-ca-certificates
-
-FROM scratch
-
-COPY --from=0 /etc/ssl/certs /etc/ssl/certs
-ADD coredns /coredns
-
+FROM gcr.io/distroless/static:nonroot
+COPY --from=setcap /coredns /coredns
+USER nonroot:nonroot
 EXPOSE 53 53/udp
 ENTRYPOINT ["/coredns"]
--- a/Makefile.release
+++ b/Makefile.release
@@ -145,7 +145,7 @@ else
 	    mkdir -p build/docker/linux/$${arch} ;\
 	    tar -xzf release/$(NAME)_$(VERSION)_linux_$${arch}.tgz -C build/docker/linux/$${arch} ;\
 	    cp Dockerfile build/docker/linux/$${arch} ;\
-	    docker build -t $(DOCKER_IMAGE_NAME)-$${arch}:$(VERSION) build/docker/linux/$${arch} ;\
+	    DOCKER_CLI_EXPERIMENTAL=enabled docker buildx build --load --platform linux/$${arch} -t $(DOCKER_IMAGE_NAME)-$${arch}:$(VERSION) build/docker/linux/$${arch} ;\
 	    docker tag $(DOCKER_IMAGE_NAME)-$${arch}:$(VERSION) $(DOCKER_IMAGE_NAME)-$${arch}:latest ;\
 	done
 endif
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ initWorkingDir: &initWorkingDir
 
 integrationDefaults: &integrationDefaults
   machine:
-    image: ubuntu-1604:201903-01
+    image: ubuntu-2004:202010-01
   working_directory: ~/go/src/${CIRCLE_PROJECT_USERNAME}/coredns
   environment:
     - K8S_VERSION: v1.19.1
@@ -36,7 +36,7 @@ buildCoreDNSImage: &buildCoreDNSImage
       command: |
         cd ~/go/src/${CIRCLE_PROJECT_USERNAME}/coredns
         make coredns SYSTEM="GOOS=linux" && \
-        docker build -t coredns . && \
+        DOCKER_CLI_EXPERIMENTAL=enabled docker buildx build --load -t coredns . && \
         kind load docker-image coredns
 
 jobs:


vinayakankugoyal:
Wow thats a lot of reviewers!

Makefile.release
```diff
@@ -145,7 +145,7 @@ else
 	    mkdir -p build/docker/linux/$${arch} ;\
 	    tar -xzf release/$(NAME)_$(VERSION)_linux_$${arch}.tgz -C build/docker/linux/$${arch} ;\
 	    cp Dockerfile build/docker/linux/$${arch} ;\
-	    docker build -t $(DOCKER_IMAGE_NAME)-$${arch}:$(VERSION) build/docker/linux/$${arch} ;\
+	    DOCKER_CLI_EXPERIMENTAL=enabled docker buildx buil --load --platform linux/$${arch} -t $(DOCKER_IMAGE_NAME)-$${arch}:$(VERSION) build/docker/linux/$${arch} ;\
```
dekkagaijin: `buil` -> `build`

vinayakankugoyal: 🙈 🤦🏾

vinayakankugoyal: Fixed!

johnbelamaric:
CI docker build is failing, PTAL.

vinayakankugoyal:
> CI docker build is failing, PTAL.

My guess is that the CI is using a pretty old version of docker. Ill have to tinker with it, which would involve making a lot of updates to the CI config. So moving this PR to draft so that I don't spam people as I push changes to the tinker with the CI.

rajansandeep:
> My guess is that the CI is using a pretty old version of docker. Ill have to tinker with it, which would involve making a lot of updates to the CI config. So moving this PR to draft so that I don't spam people as I push changes to the tinker with the CI.

I think setting the docker version should do the trick.
https://circleci.com/docs/2.0/building-docker-images/#docker-version

vinayakankugoyal:
> > My guess is that the CI is using a pretty old version of docker. Ill have to tinker with it, which would involve making a lot of updates to the CI config. So moving this PR to draft so that I don't spam people as I push changes to the tinker with the CI.
> 
> I think setting the docker version should do the trick.
> https://circleci.com/docs/2.0/building-docker-images/#docker-version

Thanks @rajansandeep! but it looks like thats for the docker executor and we are using the machine executor. I think we just need to bump the ubuntu image version. The new recommended version is any ways ubuntu-2004:202010-01 according to this: https://circleci.com/docs/2.0/executor-intro/#machine

vinayakankugoyal:
Bumping up the machine image version worked in terms of fixing the build failure. But 1 test seems to be failing. I'll rerun the test to see if it is a flake.

rajansandeep:
Yes, it was a flake. Re-running it passed the test.

vinayakankugoyal:
Given that the tests are passing, moving this to "Ready for review".

dekkagaijin (approved):

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Use distroles..." ]
>Given that the tests are passing moving this to "Ready for review".

pretty concerning that yet again docker gives issues. Our release machine
runs docker version 5:20.10.5~3-0~debian-buster



/Miek

--
Miek Gieben

Dockerfile
```diff
@@ -1,12 +1,11 @@
...
-ADD coredns /coredns
-
+FROM --platform=linux/$BUILDARCH k8s.gcr.io/build-image/setcap:buster-v1.4.0
+COPY coredns /coredns
+# We apply cap_net_bind_service so that coredns can be run as
+# non-root and still listen on port less than 1024
+RUN setcap cap_net_bind_service=+ep /coredns
+
+FROM gcr.io/distroless/static:nonroot
+COPY --from=0 /coredns /coredns
```
SuperQ: You can alias layers in a multi-step build to make it slightly nicer.

```suggestion
FROM --platform=linux/$BUILDARCH k8s.gcr.io/build-image/setcap:buster-v1.4.0 AS setcap
COPY coredns /coredns
# We apply cap_net_bind_service so that coredns can be run as
# non-root and still listen on port less than 1024
RUN setcap cap_net_bind_service=+ep /coredns

FROM gcr.io/distroless/static:nonroot
COPY --from=setcap /coredns /coredns
```

vinayakankugoyal: my rationale for not doing that was because there is only one other stage, but looking at your suggested change it does look much better. Done. Thanks for the suggestion.

Makefile.release
```diff
@@ -145,7 +145,7 @@ else
 	    mkdir -p build/docker/linux/$${arch} ;\
 	    tar -xzf release/$(NAME)_$(VERSION)_linux_$${arch}.tgz -C build/docker/linux/$${arch} ;\
 	    cp Dockerfile build/docker/linux/$${arch} ;\
-	    docker build -t $(DOCKER_IMAGE_NAME)-$${arch}:$(VERSION) build/docker/linux/$${arch} ;\
+	    DOCKER_CLI_EXPERIMENTAL=enabled docker buildx build --load --platform linux/$${arch} -t $(DOCKER_IMAGE_NAME)-$${arch}:$(VERSION) build/docker/linux/$${arch} ;\
```
miekg, line 148: why does this need this experimental stuff? I don't trust docker to do the right thing and this has bitten us before. Why does an image change require these changes as well?

I think about 3 released failed because docker when fubar, and I'm close to ripping it all out because of the fragility of it all

vinayakankugoyal: It doesn't need the experimental stuff I think it only needs DOCKER_BUILDKIT=1. I know k8s switched to using it so I thought we can switch it too. 

> I think about 3 released failed because docker when fubar, and I'm close to ripping it all out because of the fragility of it all
What is the "it" that you plan to rip out?

vinayakankugoyal: We need the experimental stuff that because:
1. It lets us run part of the build stage in the arch of the machine building the image
2. It preserves xattrs so capabilities applied to the binary in one build-stage remain if the binary is copied to the other build stage. 
(FWIW k8s switched all their builds to it, but that might not be a valid argument for you to make the change.)

> I think about 3 released failed because docker when fubar, and I'm close to ripping it all out because of the fragility of it all

What is the "it" that you are close to ripping out?

zouyee: The explanation that probably makes sense, see https://github.com/kubernetes/kubernetes/pull/87954

But IMO, I don't think it makes too much sense to enable buildx. 😄

Dockerfile
```diff
@@ -1,12 +1,11 @@
...
-FROM scratch
-
-COPY --from=0 /etc/ssl/certs /etc/ssl/certs
-ADD coredns /coredns
-
+FROM --platform=linux/$BUILDARCH k8s.gcr.io/build-image/setcap:buster-v1.4.0 AS setcap
+COPY coredns /coredns
+# We apply cap_net_bind_service so that coredns can be run as
+# non-root and still listen on port less than 1024
+RUN setcap cap_net_bind_service=+ep /coredns
```
pacoxu, line 5: Per https://github.com/kubernetes/kubernetes/issues/102612, `sysctl net/ipv4/ip_unprivileged_port_start=53` maybe a future default behavior.

Makefile.release
BenTheElder: without buildx you cannot have xattrs in a dockerfile build. the image format supports it, but not "docker build". buildx does.

PR #4531, opened 2021-03-18
plugin/bind: Discard link-local addresses on binding by interface name

m-yosefpor:
Signed-off-by: Mohammad Yosefpor <myusefpur@gmail.com>

Binding doesn't work on link local IPv6 addresses (`fe80::/10`). I found multiple stackoverflow/github issues related to binding on IPv6 link-local (e.g. [this](https://stackoverflow.com/questions/30984656/socket-programming-bind-invalid-argument) or [this](https://github.com/cytopia/pwncat/issues/89) ) explaining why this does not work. It seems we need to also specify scope_id in binding. So it needs changes in coredns/core and caddy parts if we want to be able to bind on link local addresses as well.

So I suggest, we add a `IsLinkLocalUnicast()` check and discard link-local IPv6 addresses, and specify a `except for link-local addresses` in readme.

Without this, bind by interface name feature does not work with interfaces with link local address (which in some distros it has by default), so using those interface names will not work then. We should either discard them (what has been done in this PR), or change caddy code to allow binding on link-local addresses.

https://github.com/coredns/coredns/issues/4219
https://github.com/coredns/coredns/pull/4522

Changed README.md

It has backward compatibility.

### DIFF
--- a/plugin/bind/README.md
+++ b/plugin/bind/README.md
@@ -13,7 +13,7 @@ If several addresses are provided, a listener will be open on each of the IP pro
 
 Each address has to be an IP or name of one of the interfaces of the host. Bind by interface name, binds to the IPs on that interface at the time of startup or reload (reload will happen with a SIGHUP or if the config file changes).
 
-If the given argument is an interface name, and that interface has serveral IP addresses, CoreDNS will listen on all of the interface IP addresses (including IPv4 and IPv6), except for link-local addresses on that interface.
+If the given argument is an interface name, and that interface has serveral IP addresses, CoreDNS will listen on all of the interface IP addresses (including IPv4 and IPv6), except for IPv6 link-local addresses on that interface.
 
 ## Syntax
 
--- a/plugin/bind/README.md
+++ b/plugin/bind/README.md
@@ -13,7 +13,7 @@ If several addresses are provided, a listener will be open on each of the IP pro
 
 Each address has to be an IP or name of one of the interfaces of the host. Bind by interface name, binds to the IPs on that interface at the time of startup or reload (reload will happen with a SIGHUP or if the config file changes).
 
-If the given argument is an interface name, and that interface has serveral IP addresses, CoreDNS will listen on all of the interface IP addresses (including IPv4 and IPv6).
+If the given argument is an interface name, and that interface has serveral IP addresses, CoreDNS will listen on all of the interface IP addresses (including IPv4 and IPv6), except for link-local addresses on that interface.
 
 ## Syntax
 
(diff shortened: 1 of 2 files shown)


m-yosefpor:
I've also tested this and it's working as expected:

```
# cat Corefile
.:5053 {
    bind eth0
    forward . /etc/resolv.conf
}
```

```
# ip a s eth0
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether fa:16:3e:04:f6:73 brd ff:ff:ff:ff:ff:ff
    inet 172.21.49.102/22 brd 172.21.51.255 scope global dynamic eth0
       valid_lft 73697sec preferred_lft 73697sec
    inet6 2001:db8:0:f101::1/64 scope global
       valid_lft forever preferred_lft forever
    inet6 fe80::f816:3eff:fe04:f673/64 scope link
       valid_lft forever preferred_lft forever
 ```
 
 ```
 # ./coredns -conf Corefile
.:5053 on 2001:db8:0:f101::1
.:5053 on 172.21.49.102
CoreDNS-1.8.3
linux/amd64, go1.15.5, c6b1d878-dirty
```

As it can be seen, it listens on all interfaces (including IPv4 and IPv6 of `eth0` except for link-local `fe80::f816:3eff:fe04:f673`

plugin/bind/README.md
```diff
@@ -13,7 +13,7 @@ If several addresses are provided, a listener will be open on each of the IP pro
 
 Each address has to be an IP or name of one of the interfaces of the host. Bind by interface name, binds to the IPs on that interface at the time of startup or reload (reload will happen with a SIGHUP or if the config file changes).
 
-If the given argument is an interface name, and that interface has serveral IP addresses, CoreDNS will listen on all of the interface IP addresses (including IPv4 and IPv6).
+If the given argument is an interface name, and that interface has serveral IP addresses, CoreDNS will listen on all of the interface IP addresses (including IPv4 and IPv6), except for link-local addresses on that interface.
```
miekg: this includes the v4 range ? 169.254/16 as well?

miekg:
lgtrm, one question

m-yosefpor:
Yeah it works with IPv4 link-local address now. So I also changed the README.md to specify `IPv6 link-local` too.

```
# cat Corefile
.:5053 {
    bind eth0
    forward . /etc/resolv.conf
}
```
```
# ./coredns -conf Corefile
.:5053 on 169.254.2.2
.:5053 on 172.21.49.102
.:5053 on 2001:db8:0:f101::1
CoreDNS-1.8.3
linux/amd64, go1.15.5, 5141daf2-dirty
```

miekg (approved):

PR #4534, opened 2021-03-19
metrics: remove RR type

### LINKED ISSUE #4507: label cardinality: remove `type` from coredns_dns_request_duration_seconds
I was just manually looking at /metrics and that metric is *huge*, don't know yet when this was added or the original rational, but dropping the `type` here makes sense. This leaves server and zone (which can already be a lot)

miekg:
To combat label cardinality explosions remove the type from metrics.
This was most severe in the histogram for request duration, but remove
it everywhere. This also allow for the removal of dnsutil metric type
stuff; signaling "this is probably not a good idea".

It's also highlighted difference between grpc and forward code, where
forward did use type and grpc didn't; getting rid of all that "fixes"
that discrepancy

Fixes: #4507

### DIFF
--- a/plugin/metrics/vars/vars.go
+++ b/plugin/metrics/vars/vars.go
@@ -21,14 +21,14 @@ var (
 		Subsystem: subsystem,
 		Name:      "request_duration_seconds",
 		Buckets:   plugin.TimeBuckets,
-		Help:      "Histogram of the time (in seconds) each request took.",
-	}, []string{"server", "zone", "type"})
+		Help:      "Histogram of the time (in seconds) each request took per zone.",
+	}, []string{"server", "zone"})
 
 	RequestSize = promauto.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: plugin.Namespace,
 		Subsystem: subsystem,
 		Name:      "request_size_bytes",
-		Help:      "Size of the EDNS0 UDP buffer in bytes (64K for TCP).",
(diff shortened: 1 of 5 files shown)


plugin/metrics/vars/vars.go
```diff
@@ -14,21 +14,21 @@ var (
 		Subsystem: subsystem,
 		Name:      "requests_total",
 		Help:      "Counter of DNS requests made per zone, protocol and family.",
-	}, []string{"server", "zone", "proto", "family", "type"})
+	}, []string{"server", "zone", "proto", "family"})
```
SuperQ: I think we still want to keep the `type` here. I can see wanting to remove it on the duration histogram, but it's useful here.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] metrics: remo..." ]
>@SuperQ commented on this pull request.
>
>
>In plugin/metrics/vars/vars.go:
>
>> @@ -14,21 +14,21 @@ var (
>                Subsystem: subsystem,
>                Name:      "requests_total",
>                Help:      "Counter of DNS requests made per zone, protocol and family.",
>-       }, []string{"server", "zone", "proto", "family", "type"})
>+       }, []string{"server", "zone", "proto", "family"})
>
>I think we still want to keep the type here. I can see wanting to remove it on
>the duration histogram, but it's useful here.

yeah, was on the fence about this one. I'll add that back.

SuperQ (approved):
LGTM

PR #4537, opened 2021-03-21
plugin/dnssec: use entire RRset as key input

miekg:
This uses the entire rrset as input for the hash key; this is to detect
differences in the RRset and generate the correct signature.

As this would then lead to unbounded growth, we periodically (every 8h)
prune the cache of old entries. In theory we could rely on the random
eviction, but it seems nicer to do this in a maintannce loop so that we
remove the unused ones. This required adding a Walk function to the
plugin/pkg/cache.

### DIFF
--- a/plugin/dnssec/cache.go
+++ b/plugin/dnssec/cache.go
@@ -29,7 +29,7 @@ func periodicClean(c *cache.Cache, stop <-chan struct{}) {
 	for {
 		select {
 		case <-tick.C:
-			// we sign for 8 days, check if a signature in the cache reached 3/4 of that, if found delete
+			// we sign for 8 days, check if a signature in the cache reached 75% of that (i.e. 6), if found delete
 			// the signature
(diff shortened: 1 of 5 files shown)


plugin/dnssec/cache.go
```diff
@@ -3,25 +3,46 @@ package dnssec
...
+	return h.Sum64()
+}
+
+func periodicClean(c *cache.Cache, stop <-chan struct{}) {
+	tick := time.NewTicker(8 * time.Hour)
+	defer tick.Stop()
+	for {
+		select {
+		case <-tick.C:
+			// we sign for 8 days, check if a signature in the cache reached 3/4 of that, if found delete
```
chrisohaver:
```suggestion
			// we sign for 8 days, check if a signature in the cache reached 75% of that (i.e. 6), if found delete
```

chrisohaver (approved):

PR #4538, opened 2021-03-21
reverse zones

### LINKED ISSUE #2779: Reverse zones should be calculated for the full CIDR range

miekg:
This fixes the reverse zones handling. Add expanstion of the reverse
notation to all octet boundary subnets and add those to the config - just as if
they were directly typed in the config.

This takes inspiration from #4501, but that (even with DCO!!) seems to
be just using https://github.com/apparentlymart/go-cidr/ so use that
instead - I think a minor function is still needed that one is copied
from #4501.

Also sort the zones we are listing on startup - caught in this PR
because of the expanded zones being not listed next to each other.

This also removes the need for FilterFunc from the config, so this is
now gone as well, making the whole thing slightly more efficient.

Add couple of reverse unit tests and a e2e test that queries for the
correct (and incorrect) reverse zones and checks the reply.

Closes: #4501
Fixes: #2779


chrisohaver:
Thanks!

One issue I see: If a plugin chain is created for each resulting zone, this could run afoul with the duplicate plugin chain issue in coredns, since a single cidr can expand to many zones.  See https://github.com/coredns/coredns/pull/3862 for a proposed fix to the duplicate plugin chain issue.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] reverse zones..." ]
>Thanks!
>
>One issue I see: If a plugin chain is created for each resulting zone, this
>could run afoul with the duplicate plugin chain issue in coredns, since a
>single cidr can expand to many zones. See #3862 for a proposed fix to the
>duplicate plugin chain issue.

yes, but no different than typing these out. I was thinking about adding a warning if you
add more then 10 subnets or something but that seems arbitrary.

chrisohaver:
This looks ok to me, for a _serverblock_ level fix.

It doesn't fix this issue for individual plugin use of CIDR notation however, which is what #4501 aimed to solve (albeit only in one in one plugin: kubernetes).  A fix at the plugin level would require hands on every plugin that uses Normalize to read zones form the Corefile; perhaps by introducing a CIDR-friendly version of Normalize that returns a list of zones, and migrating plugins over to the new version.

miekg:
Good point. I hadn't considered that.

Agree with the approach, add new function and move (all?) in tree plugins
over. And leave the old function for backwards compatibility.

miekg:
let's first commit this one and then tackle the Normalize function and push that through

Issue #4435, opened 2021-01-29
Allow overriding RA (RecursionAvailable) flag value


s-matyukevich:
**What would you like to be added**:

Allow uses to specify whether `file` plugin (and probably some other plugins as well) set RA flag or not.

**Why is this needed**:

After https://github.com/coredns/coredns/pull/2408 we no longer set RA flag. However, this breaks our use case: we use CoreDNS as upstream resolver and dmsmasq as a local cache. dnsmasq doesn't cache responses from upstream servers if RA flag is unset. Here is the relevant dnsmasq code snippet

```
  /* Don't put stuff from a truncated packet into the cache.
     Don't cache replies from non-recursive nameservers, since we may get a 
     reply containing a CNAME but not its target, even though the target 
     does exist. */
  if (!(header->hb3 & HB3_TC) && 
      !(header->hb4 & HB4_CD) &&
      (header->hb4 & HB4_RA) &&
      !no_cache_dnssec)
    cache_end_insert();
```

It would be great if we can set the value for this flag in Corefile. for example 

```
.:53 {
    file /some-file {
        reload 30s
        recursion-available true
    }
}
```

miekg:
[ Quoting <notifications@github.com> in "[coredns/coredns] Allow overriding ..." ]
>What would you like to be added:
>
>Allow uses to specify whether file plugin (and probably some other plugins as
>well) set RA flag or not.
>
>Why is this needed:
>
>After #2408 we no longer set RA flag. However, this breaks our use case: we use
>CoreDNS as upstream resolver and dmsmasq as a local cache. dnsmasq doesn't
>cache responses from upstream servers if RA flag is unset. Here is the relevant
>dnsmasq code snippet
>
>  /* Don't put stuff from a truncated packet into the cache.
>     Don't cache replies from non-recursive nameservers, since we may get a
>     reply containing a CNAME but not its target, even though the target
>     does exist. */
>  if (!(header->hb3 & HB3_TC) &&
>      !(header->hb4 & HB4_CD) &&
>      (header->hb4 & HB4_RA) &&
>      !no_cache_dnssec)
>    cache_end_insert();
>
>It would be great if we can set the value for this flag in Corefile. for
>example
>
>.:53 {
>    file /some-file {
>        reload 30s
>        recursion-available true
>    }
>}

I consider this a bug in dnsmasq, it should be smarter then looking a 1 bit and making a
decision.

[...]

PR #4541, opened 2021-03-22
Add field to rewrite plugin to support flag rewrites.

mqasimsarfraz:
There are cases where users want to change the response flags to fit their needs. e.g `dnsmasqd` needs `ra` for caching so having a generic way to changes response might be helpful in such cases. 

Also, this is just a draft of an idea I had so I am sure it can be further improved. It would be great if I can please get the maintainers review on this? 
#4435
The usage for the plugin needs to be added to documentation.
No

### DIFF
--- a/plugin/rewrite/README.md
+++ b/plugin/rewrite/README.md
@@ -291,6 +291,18 @@ rewrite edns0 subnet set 24 56
 * If the query's source IP address is an IPv4 address, the first 24 bits in the IP will be the network subnet.
 * If the query's source IP address is an IPv6 address, the first 56 bits in the IP will be the network subnet.
 
+### Flag Response Rewrites
+Using the FIELD `flag` you can set or clear the flags in the responses. The syntax of the rule is as follows:
+
+```
+rewrite [continue|stop] flag [set|clear] STRING
+```
+Following example add recursive available flag to the response:
+```
+rewrite flag set ra
+```
+Currently supported flag include `aa`, `ra` and `rd`.
+
 ## Full Syntax
 
 The full plugin usage syntax is harder to digest...
--- a/plugin/rewrite/flag.go
+++ b/plugin/rewrite/flag.go
@@ -1,0 +1,79 @@
+// Package rewrite is a plugin for rewriting requests internally to something different.
+package rewrite
+
+import (
+	"context"
+	"fmt"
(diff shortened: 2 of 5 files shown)


miekg:
[ Quoting ***@***.***> in "[coredns/coredns] Add field to rewr..." ]
>
>There are cases where users want to change the response flags to fit their
>needs. e.g dnsmasqd needs ra for caching so having a generic way to changes
>response might be helpful in such cases.

I think we're piling to much stuff into this plugin and we should either not do it, or
make another plugin that just rewrites these bits.

mqasimsarfraz:
Thanks for the response @miekg! 
> I think we're piling to much stuff into this plugin and we should either not do it, or make another plugin that just rewrites these bits.

Yes, agreed. I also saw that we are doing [refactoring](https://github.com/coredns/coredns/pull/4473) around the same code as well. Would it make sense to add it to `rewrite` plugin afterwards if it is simpler? If not then I am happy to propose a separate plugin?

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Add field to ..." ]
>Thanks for the response @miekg!
>
>    I think we're piling to much stuff into this plugin and we should either
>    not do it, or make another plugin that just rewrites these bits.
>
>Yes, agreed. I also saw that we are doing refactoring around the same code as
>well. Would it make sense to add it to rewrite plugin afterwards if it is
>simpler? If not then I am happy to propose a separate plugin?

I think a new plugin makes sense, yes will copy some boilerplate code, but in the end such
separation is good.

Now it only needs a good name.

mqasimsarfraz:
> I think a new plugin makes sense, yes will copy some boilerplate code, but in the end such separation is good. Now it only needs a good name.

Ah, the hardest part! :) how about `dnsflag` , `flag` or `flagbits` ?

miekg:
Maybe just "header"?

mqasimsarfraz:
Yes, `header` sounds nice. I will propose a PR soon. Do you have any recommendations? I will go ahead and close this!

edgarasg:
@MQasimSarfraz will there be any progress with header plugin?

mqasimsarfraz:
@edgarasg Sorry I haven't had time to finalize anything but let me push my changes for review soon!

PR #4543, opened 2021-03-23
plugin/bind: exclude interface or ip address

m-yosefpor:
As per discussions in https://github.com/coredns/coredns/issues/4219 , this PR implements `except` directive for bind plugin.

```
. {
    bind lo {
        except 127.0.0.1
    }
}
```

**IMPORTANT NOTE:** This PR does not currently work with wildcard binding. e.g.

```
. {
  bind :: {
    except 127.0.0.1
  }
```

https://github.com/coredns/coredns/issues/4219

README.md has changed.

No it is backward compatible.

### DIFF
--- a/plugin/bind/README.md
+++ b/plugin/bind/README.md
@@ -80,6 +80,7 @@ You can exclude some addresses by their IP or interface name (The following will
         exclude 127.0.0.1
     }
 }
+~~~
 
 ## Bugs
 
--- a/plugin/bind/README.md
+++ b/plugin/bind/README.md
@@ -20,22 +20,22 @@ If the given argument is an interface name, and that interface has serveral IP a
 In its basic form, a simple bind uses this syntax:
 
 ~~~ txt
-bind ADDRESS  ...
+bind ADDRESS|IFACE  ...
 ~~~
 
 You can also exclude some addresses with their IP address or interface name in expanded syntax:
 
 ~~~
-bind ADDRESS ... {
-    except ADDRESS ...
+bind ADDRESS|IFACE ... {
+    except ADDRESS|IFACE ...
 }
 ~~~
 
 
 
-* **ADDRESS** is an IP address or interface name to bind to.
+* **ADDRESS|IFACE** is an IP address or interface name to bind to.
 When several addresses are provided a listener will be opened on each of the addresses. Please read the *Description* for more details.
-* `except`, exclude interfaces or ip addresses to bind to. `except` option only exclude addresses for the current `bind` directive if multiple `bind` directives are used in the same server block.
+* `except`, excludes interfaces or ip addresses to bind to. `except` option only excludes addresses for the current `bind` directive if multiple `bind` directives are used in the same server block.
 ## Examples
 
 To make your socket accessible only to that machine, bind to IP 127.0.0.1 (localhost):
@@ -72,12 +72,12 @@ The following server block, binds on localhost with its interface name (both "12
 }
 ~~~
 
-You can exclude some addresses by their IP or interface name (The following will only listen on `::1`):
+You can exclude some addresses by their IP or interface name (The following will only listen on `::1` or whatever addresses have been assigned to the `lo` interface):
 
 ~~~ corefile
 . {
     bind lo {
-        exclude 127.0.0.1
+        except 127.0.0.1
     }
 }
 ~~~
--- a/plugin/bind/README.md
+++ b/plugin/bind/README.md
@@ -17,13 +17,25 @@ If the given argument is an interface name, and that interface has serveral IP a
 
 ## Syntax
 
+In its basic form, a simple bind uses this syntax:
+
 ~~~ txt
 bind ADDRESS  ...
 ~~~
 
-**ADDRESS** is an IP address to bind to.
-When several addresses are provided a listener will be opened on each of the addresses.
+You can also exclude some addresses with their IP address or interface name in expanded syntax:
 
+~~~
+bind ADDRESS ... {
+    except ADDRESS ...
+}
+~~~
+
+
+
+* **ADDRESS** is an IP address or interface name to bind to.
+When several addresses are provided a listener will be opened on each of the addresses. Please read the *Description* for more details.
+* `except`, exclude interfaces or ip addresses to bind to. `except` option only exclude addresses for the current `bind` directive if multiple `bind` directives are used in the same server block.
 ## Examples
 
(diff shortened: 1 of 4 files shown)


miekg:
test fail
```
--- FAIL: TestSetup (0.00s)
    setup_test.go:29: Test 7, expected no errors, but got: plugin/bind: invalid option "except"
```

plugin/bind/README.md
```diff
@@ -17,13 +17,25 @@ If the given argument is an interface name, and that interface has serveral IP a
...
 ~~~ txt
 bind ADDRESS  ...
 ~~~
 
-**ADDRESS** is an IP address to bind to.
-When several addresses are provided a listener will be opened on each of the addresses.
+You can also exclude some addresses with their IP address or interface name in expanded syntax:
+
+~~~
+bind ADDRESS ... {
```
miekg: should this be ADDRESS|IFACE? and throughout ?

plugin/bind/README.md
```diff
@@ -60,6 +72,16 @@ The following server block, binds on localhost with its interface name (both "12
 }
 ~~~
 
+You can exclude some addresses by their IP or interface name (The following will only listen on `::1`):
```
miekg: (or whatever addresses have been assigned to the `lo` interface).

plugin/bind/bind.go
```diff
@@ -1,6 +1,17 @@
...
-import "github.com/coredns/coredns/plugin"
+import (
+	"github.com/coredns/coredns/plugin"
+)
 
 func init() { plugin.Register("bind", setup) }
+
+type bind struct {
+	Next     plugin.Handler
+	includes []string
```
miekg: 'includes' needs a better name, `addrs`, `ips`?

plugin/bind/README.md
m-yosefpor: I explained it throughly in description part, and referenced this section to there for more details

m-yosefpor:
@miekg  suggestions applied. Just to make sure you noticed the PR description:

IMPORTANT NOTE: This PR does not currently work with wildcard binding. e.g.
```
. {
  bind :: {
    except 127.0.0.1
  }
 ```

To implement this, we can bind on all interface IPs, instead of binding on `::`. I don't know if this this has the same behavior, or is different in some edge cases.

To elaborate more, if `::` is passed, and there is no `except` directive, bind on `::` (backward compatibility in edge cases), and if one or more `except` directive is provided, translate `::` to all IP addresses of all interfaces, and exclude given IP addresses in exclude. FYI this is also what dnsmasq is doing when passing `--bind-interface` flag. Shall I implement this like that? or is there any better way doing this?

P.S. I only always mention dnsmasq as an examples, as I have prior experience with it, nothing more.

**Comment by miekg** [comment]
[ Quoting ***@***.***> in "Re: [coredns/coredns] plugin/bind: ..." ]
>
>
>In plugin/bind/README.md:
>
>>  ~~~ txt
> bind ADDRESS  ...
> ~~~
>
>-**ADDRESS** is an IP address to bind to.
>-When several addresses are provided a listener will be opened on each of the addresses.
>+You can also exclude some addresses with their IP address or interface name in expanded syntax:
>+
>+~~~
>+bind ADDRESS ... {
>
>I explained it throughly in description part, and referenced this section to
>there for more details

while reading through the README I see I have more suggestions on how to make it better
and more consistent - I'll send a PR after this is merged.

File: plugin/bind/setup.go [hunk 4]
```diff
@@ -10,48 +11,101 @@ import (
...
 
 func setup(c *caddy.Controller) error {
-	config := dnsserver.GetConfig(c)
 
+	config := dnsserver.GetConfig(c)
 	// addresses will be consolidated over all BIND directives available in that BlocServer
 	all := []string{}
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return plugin.Error("bind", fmt.Errorf("failed to get interfaces list"))
```
miekg:
```suggestion
		return plugin.Error("bind", fmt.Errorf("failed to get interfaces list: %s", err))
```

plugin/bind/setup.go
```diff
@@ -10,48 +11,101 @@ import (
...
-					addrs, err := iface.Addrs()
-					if err != nil {
-						return plugin.Error("bind", fmt.Errorf("failed to get the IP addresses of the interface: %q", arg))
-					}
-					for _, addr := range addrs {
-						if ipnet, ok := addr.(*net.IPNet); ok {
-							if ipnet.IP.To4() != nil || (!ipnet.IP.IsLinkLocalMulticast() && !ipnet.IP.IsLinkLocalUnicast()) {
-								all = append(all, ipnet.IP.String())
-							}
+			return err
```
miekg: can you move the plugin.Error calls from `parse` here? So only this function has those? Had to look twice if this was done correctly in the current code

plugin/bind/README.md
```diff
@@ -17,13 +17,25 @@ If the given argument is an interface name, and that interface has serveral IP a
...
+}
+~~~
+
 
-**ADDRESS** is an IP address to bind to.
-When several addresses are provided a listener will be opened on each of the addresses.
 
+* **ADDRESS|IFACE** is an IP address or interface name to bind to.
+When several addresses are provided a listener will be opened on each of the addresses. Please read the *Description* for more details.
+* `except`, excludes interfaces or ip addresses to bind to. `except` option only excludes addresses for the current `bind` directive if multiple `bind` directives are used in the same server block.
```
miekg:
```suggestion
* `except`, excludes interfaces or IP addresses to bind to. `except` option only excludes addresses for the current `bind` directive if multiple `bind` directives are used in the same server block.
```

plugin/bind/setup.go
miekg: Or at least have a single function where plugin.Error is called, usually this is only done in `setup`

plugin/bind/README.md
m-yosefpor: fixed.

plugin/bind/setup.go
m-yosefpor: Nice catch 👌 I moved them around to new functions and missed that duplicate plugin.Error call. Now I moved all plugin.Error from `parse` and `listIP` to `setup` function..

plugin/bind/setup.go
m-yosefpor: done

miekg (approved):

PR #4025, opened 2020-07-24
using promauto package to ensure all created metrics are properly registered

zouyee:
### 1. Why is this pull request needed and what does it do?
check if metric is registererd 

### 2. Which issues (if any) are related?
fix #4024

### 3. Which documentation changes (if any) need to be made?
none
### 4. Does this introduce a backward incompatible change or deprecation?
none

PR #4546, opened 2021-03-24
fix: reload metrics failed count

pacoxu:
When I check the reload failed count metrics in my local cluster, it is not shown.

After some investigation, it may be a bug.

https://pkg.go.dev/github.com/inverse-inc/packetfence/go/caddy/caddy#Controller.OnRestart

It was only called when caddy restarts.
Reload is not `restart`?

None

None


pacoxu:
It was fixed by #4025 in 1.7.1

I was using 1.6.7 now.

PR #4556, opened 2021-04-01
Variable initialization update

Akiros001:
Variable initialization update


miekg:
please clearly state why you want to make such a trivial change and better yet first open an issue to discuss and prevent busy work for every one involved

PR #4557, opened 2021-04-02
Variable initialization update

Akiros001:
Variables are declared without an initial value, and it is recommended to define methods using var


miekg:
Sorry, but either way is fine in Go.

/close
/lock

PR #4570, opened 2021-04-12
plugin/kubernetes: Support both v1 and v1beta1 EndpointSlices

chrisohaver:

Adds support for both v1 and v1beta1 EndpointSlices.
 AFAIK, v1 and v1beta1 are functionally identical, however using v1beta1 on a K8s 1.21 cluster results in a deprecation notice warning in the log.  Using v1 on a K8s 1.21 cluster averts the warning. So, this code is essentially in place for now to remove a warning from the log. I welcome opinions as to whether or not it's worth adding this for that reason.

In this change, v1 will be used if available, if not, v1beta1 is used.

Like the dual Endpoint/EndpointSlice support, this code can be temporary until versions of K8s that do not support the v1 are no longer supported, at which point this code can be removed.

#4568

### DIFF
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -20,7 +20,8 @@ import (
 
 	"github.com/miekg/dns"
 	api "k8s.io/api/core/v1"
-	discovery "k8s.io/api/discovery/v1beta1"
+	discovery "k8s.io/api/discovery/v1"
+	discoveryV1beta1 "k8s.io/api/discovery/v1beta1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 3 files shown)


plugin/kubernetes/kubernetes.go
```diff
@@ -292,7 +298,8 @@ func (k *Kubernetes) InitKubeCache(ctx context.Context) (onStart func() error, o
 // should be watched, and false when endpoints should be watched.
 // If the API supports discovery v1 beta1, and the server versions >= 1.19, endpointslices are watched.
```
sgreene570: This comment should most likely be updated to reflect the new behavior of `endpointSliceSupported`.

plugin/kubernetes/kubernetes.go
```diff
@@ -256,10 +257,15 @@ func (k *Kubernetes) InitKubeCache(ctx context.Context) (onStart func() error, o
...
 			if initEndpointWatch {
 				// Revert to watching Endpoints for incompatible K8s.
-				// This can be remove when all supported k8s versions support endpointslices.
-				if ok := k.endpointSliceSupported(kubeClient); !ok {
+				// This can be removed when all supported k8s versions support endpointslices.
+				ok, v := k.endpointSliceSupported(kubeClient)
+				if !ok {
 					k.APIConn.(*dnsControl).WatchEndpoints(ctx)
 				}
+				// Revert to EndpointSlice v1beta1 if v1 is not supported
```
sgreene570, line 265: Should this have a comment along the lines of:
"This can removed when all supported k8s versions no longer support EndpointSlice v1beta1" ?

chrisohaver: Actually, the entire function should be removed before that happens (or at the same time). IOW, once all supported k8s versions support v1, we can simply use v1, with no need for legacy endpoints or beta slices fallback.

I updated the function's comments to reflect that.

sgreene570:
Friendly ping on this. Any updates here? :grin:

chrisohaver:
> Friendly ping on this. Any updates here? 😁

No updates. Just waiting on a review from a maintainer.

johnbelamaric (approved):

chrisohaver:
At first glance... it looks like merging this broke the unready endpoint fix ...

chrisohaver:
> At first glance... it looks like merging this broke the unready endpoint fix ...

should be fixed with ... #4616

PR #4580, opened 2021-04-23
plugin/kubernetes: Exclude unready endpoints from endpointslices

chrisohaver:

Records are incorrectly created for unready endpoints received via the Endpointslices watch.
The EndpointSlices API doesn't list unready endpoint addresses in a separate list as the Endpoints API did, so the condition must be explicitly checked.


rajansandeep (approved):
/lgtm

Should there be tests added to CI?

chrisohaver:
> /lgtm
> 
> Should there be tests added to CI?

Yes. (can be done after this merges)

PR #4582, opened 2021-04-23
plugin/forward: Add option for local address

chrisohaver:

An upstream DNS service might not be accessible via all local addresses that have a route to the upstream IP (e.g. the upstream might employ an ACL of some sort, or other some intervening network restriction).  In the case that more than one interface can route to the upstream, the system may automatically select an interface that is prevented from accessing the upstream DNS.

This PR adds an option to control which local address is used when forwarding requests upstream.

Included

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -51,7 +51,7 @@ forward FROM TO... {
     policy random|round_robin|sequential
     health_check DURATION [no_rec]
     max_concurrent MAX
-    local_address IP [to TO]
+    local_addr IP [to TO]
 }
 ~~~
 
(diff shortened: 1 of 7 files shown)


chrisohaver:
@miekg, PTAL is this something you'd consider for inclusion in the forward plugin?

miekg:
I think the same behavior can be accomplished by running multiple coredns instances, where one does the forward.

And once we add this, next thing that comes around is that it should be an interface name as well, so don't think this is a good enhancement

chrisohaver:
> I think the same behavior can be accomplished by running multiple coredns instances, where one does the forward.

More simply, it could be done with static routes.
If making the behavior apply only to dns traffic is a need, with iptables (probably - more or less).

Although, from a CoreDNS user perspective a Corefile option would be more convenient.

chrisohaver:
Closing per ...

> ... don't think this is a good enhancement

PR #4584, opened 2021-04-24
Add error check to avoid panic

wangchenglong01:
Add error check to avoid panic
If err is not nil, the `stat` will be nil. So the following code calling other method will cause panic.

### DIFF
--- a/plugin/hosts/hostsfile.go
+++ b/plugin/hosts/hostsfile.go
@@ -117,11 +117,14 @@ func (h *Hostsfile) readHosts() {
 	defer file.Close()
 
 	stat, err := file.Stat()
+	if err != nil {
+		return
+	}
 	h.RLock()
 	size := h.size
 	h.RUnlock()
 
-	if err == nil && h.mtime.Equal(stat.ModTime()) && size == stat.Size() {
+	if h.mtime.Equal(stat.ModTime()) && size == stat.Size() {
 		return
 	}
 
--- a/plugin/hosts/hostsfile.go
+++ b/plugin/hosts/hostsfile.go
@@ -117,11 +117,15 @@ func (h *Hostsfile) readHosts() {
 	defer file.Close()
 
 	stat, err := file.Stat()
+	if err != nil {
+		log.Warningf("Failed to get file info: %v", err)
+		return
+	}
 	h.RLock()
 	size := h.size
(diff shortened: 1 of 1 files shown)


luckymrwang:
> ### 1. Why is this pull request needed and what does it do?
> Add error check to avoid panic
> If err is not nil, the `stat` will be nil. So the following code calling other method will cause panic.
> 
> ### 2. Which issues (if any) are related?
> ### 3. Which documentation changes (if any) need to be made?
> ### 4. Does this introduce a backward incompatible change or deprecation?

image

**plugin/hosts/hostsfile.go  line:139~143**

yongtang (approved):
LGTM

wangchenglong01:
@johnbelamaric @pmoroney

ohbus (approved):
LGTM

plugin/hosts/hostsfile.go
```diff
@@ -117,11 +117,15 @@ func (h *Hostsfile) readHosts() {
...
 	stat, err := file.Stat()
+	if err != nil {
+		log.Warningf("Failed to get file info: %v", err)
+		return
+	}
 	h.RLock()
 	size := h.size
 	h.RUnlock()
 
-	if err == nil && h.mtime.Equal(stat.ModTime()) && size == stat.Size() {
```
zouyee, line 124: If the error is not null, the subsequent checks will not be executed.

plugin/hosts/hostsfile.go
```diff
@@ -117,11 +117,15 @@ func (h *Hostsfile) readHosts() {
 	defer file.Close()
 
 	stat, err := file.Stat()
+	if err != nil {
+		log.Warningf("Failed to get file info: %v", err)
```
zouyee, line 121: Determine if the error is empty, just to exit early, so the title need to be  modified.

chrisohaver: Perhaps we should skip logging a warning here for the same reason it is skipped above (`// We already log a warning if the file doesn't exist or can't be opened on setup. No need to return the error here.`).

wangchenglong01: You are right. I have removed the warning.

chrisohaver (approved):
Lgtm

PR #4591, opened 2021-04-26
Add `requests_total` metric to `hosts` plugin

f0o:
It adds a new Metric for RequestsCount similar to all other plugins that serve requests to monitor the amount of requests being served from hosts file

N/A

Potentially add newly exposed metric to docs, however I dont see the current metrics being listed there either.
So likely N/A

No

### DIFF
--- a/plugin/hosts/hosts.go
+++ b/plugin/hosts/hosts.go
@@ -69,6 +69,7 @@ func (h Hosts) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 	m.Answer = answers
 
 	w.WriteMsg(m)
+	RequestCount.WithLabelValues().Add(1)
 	return dns.RcodeSuccess, nil
 }
 
--- a/plugin/hosts/metrics.go
+++ b/plugin/hosts/metrics.go
@@ -22,4 +22,11 @@ var (
 		Name:      "reload_timestamp_seconds",
 		Help:      "The timestamp of the last reload of hosts file.",
 	})
+	// RequestCount is the amount of requests served from hosts-file/s
+	RequestCount = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "hosts",
+		Name:      "requests_total",
+		Help:      "Counter of requests made per upstream.",
+	}, []string{})
 )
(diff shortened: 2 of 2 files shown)


chrisohaver:
Adding this metric to _hosts_ has been discussed before: https://github.com/coredns/coredns/issues/3758
In the end, the decision was not to add this metric on a per-plugin basis.

f0o:
but isn't it on a per-plugin basis? I see it in numerous plugins (quite verbatim)

Above code is taken from the grpc plugin IIRC

chrisohaver:
> but isn't it on a per-plugin basis? I see it in numerous plugins (quite verbatim)

It is.  The decision at that time (about a year ago) was not to implement this independently for each plugin.

f0o:
Continuing the discussion in #3758 and closing this then

f0o:
Reopening as other plugins keep adding new counters even latest (1.8.5) versions - just like this one.

See: https://github.com/coredns/coredns/pull/4781/files

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Add `requests..." ]
>Reopening as other plugins keep adding new counters even latest (1.8.5)
>versions - just like this one.

really?

Having a counter that tells what plugin replied would be useful. Might even lead to a
reduction of total metrics

f0o:
> [ Quoting ***@***.***> in "Re: [coredns/coredns] Add `requests..." ]
> Reopening as other plugins keep adding new counters even latest (1.8.5) versions - just like this one.
> really? Having a counter that tells what plugin replied would be useful. Might even lead to a reduction of total metrics

I fully agree but it doesnt seem to be a priority so I dont see the rationale why other plugins can freely add more requests counters but specifically this one not.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Add `requests..." ]
>    [ Quoting @.***> in "Re: [coredns/coredns] Add `requests..." ]
>    Reopening as other plugins keep adding new counters even latest (1.8.5)
>    versions - just like this one.
>    really? Having a counter that tells what plugin replied would be useful.
>    Might even lead to a reduction of total metrics
>
>I fully agree but it doesnt seem to be a priority so I dont see the rationale
>why other plugins can freely add more requests counters but specifically this
>one not.

commit hash ? I have no idea what you're talking about?

PR #4592, opened 2021-04-26
plugin/kubernetes: fix SOA response for non-apex queries

### LINKED ISSUE #4585: SOA query on record does not respond with Authority Section in kubernetes
**What happened**:
We have a named/bind9 server which we use to resolve all our networks DNS. This server is configured with a forwarder for the `cluster.local.` domain to a coredns deployment running inside our kubernetes cluster. The named server is outside the cluster.

Most everything with this setup works as expected.

We configured a record (`test.record.tld`) in one of our zones to be a CNAME to the service inside kubernetes `sr.ns.svc.cluster.local.` as an example. This also seemed to work as expected.

Where we ran into issue was when attempting to issue a letsencrypt cerftificate on the above mentioned record using DNS01 challenge method and RFC2136 via cert-manager.io

RFC2136 (or cert-managers implementation) would fall over when attempting to do an SOA on `test.record.tld` due to our named server returning SERVFAIL.

The above request results in named making an SOA request to coredns inside our cluster for `sr.ns.svc.cluster.local.`. This resulted in named giving a "DNS format error, or FORMERR", with "reply has no answer".

Upon closer inspection, coredns is returning a NOERROR response with what seems to be the correct information. Looking closer still and after testing with several other domains running on different DNS servers/providers, we discovered that coredns was the only server that did not return the SOA in an authority section, and instead responded with an answer section. Our forwarder to coredns was the only forwarder we tested which caused named to fail the request.

We also noticed that changing the CNAME to just `cluster.local.` would actually work, even though the result was in an answer section. I suspect because we were requesting the root of the zone.

We expected named not to respond with SERVFAIL for forwarded SOA queries.

Our coredns config I believe is stock/default for kubernetes.
```
    .:53 {
        log
        errors
        health {
           lameduck 5s
        }
        ready
          pods insecure
          ttl 30
        }
        prometheus :9153
        forward . /etc/resolv.conf {
           max_concurrent 1000
        }
        cache 30
        loop
        reload
        loadbalance
    }
```

Running the following command

`dig @<core dns ip> svc.cluster.local soa`

results in

```
; <<>> DiG 9.16.13-Ubuntu <<>> @10.96.0.10 svc.cluster.local soa
; (1 server found)
;; global options: +cmd
;; Got answer:
;; WARNING: .local is reserved for Multicast DNS
;; You are currently testing what happens when an mDNS query is leaked to DNS
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 49913
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; COOKIE: 5a80dbacc96f81a9 (echoed)
;; QUESTION SECTION:
;svc.cluster.local.		IN	SOA

;; ANSWER SECTION:
cluster.local.		27	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1619328346 7200 1800 86400 30

;; Query time: 0 msec
;; SERVER: 10.96.0.10#53(10.96.0.10)
;; WHEN: Sun Apr 25 05:26:09 UTC 2021
;; MSG SIZE  rcvd: 151
```

We'd instead expect

```
; <<>> DiG 9.16.13-Ubuntu <<>> @10.96.0.10 svc.cluster.local soa
; (1 server found)
;; global options: +cmd
;; Got answer:
;; WARNING: .local is reserved for Multicast DNS
;; You are currently testing what happens when an mDNS query is leaked to DNS
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 47775

;; OPT PSEUDOSECTION:
; COOKIE: e640ac1a44271985 (echoed)
;; QUESTION SECTION:
;svc.cluster.local.		IN	SOA

;; AUTHORITY SECTION:
cluster.local.		19	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1619326708 7200 1800 86400 30

;; Query time: 0 msec
;; SERVER: 10.96.0.10#53(10.96.0.10)
;; WHEN: Sun Apr 25 05:29:02 UTC 2021
;; MSG SIZE  rcvd: 151
```

We worked around this by adding the following to our coredns config. It's not ideal, but we think it's an ok workaround.
```
        template IN SOA cluster.local {
          match ".*\.cluster\.local\.$"
          authority "cluster.local. 30 IN SOA ns.dns.cluster.local. hostmaster.cluster.local. 1619326708 7200 1800 86400 30"
          fallthrough
        }
```
With this configuration in place, everything works as expected, and SSL certificate renewal works.

Is this behavior expected? Should the workaround be necessary?

**Environment**:

- the version of CoreDNS: Tested with 1.7.0 and 1.8.3
- Corefile:
- logs, if applicable:
- Others:

chrisohaver:

Correctly handle non-apex SOA queries.

Fixes #4585

### DIFF
--- a/plugin/kubernetes/handler.go
+++ b/plugin/kubernetes/handler.go
@@ -43,13 +43,13 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 	case dns.TypeSRV:
 		records, extra, err = plugin.SRV(ctx, &k, zone, state, plugin.Options{})
 	case dns.TypeSOA:
-		if len(state.Name()) == len(zone) {
+		if qname == zone {
 			records, err = plugin.SOA(ctx, &k, zone, state, plugin.Options{})
 		}
 	case dns.TypeAXFR, dns.TypeIXFR:
 		return dns.RcodeRefused, nil
 	case dns.TypeNS:
-		if len(state.Name()) == len(zone) {
+		if state.Name() == zone {
 			records, extra, err = plugin.NS(ctx, &k, zone, state, plugin.Options{})
 			break
 		}
--- a/plugin/kubernetes/handler.go
+++ b/plugin/kubernetes/handler.go
@@ -43,7 +43,9 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 	case dns.TypeSRV:
 		records, extra, err = plugin.SRV(ctx, &k, zone, state, plugin.Options{})
 	case dns.TypeSOA:
-		records, err = plugin.SOA(ctx, &k, zone, state, plugin.Options{})
+		if zone == state.Name() {
+			records, err = plugin.SOA(ctx, &k, zone, state, plugin.Options{})
+		}
 	case dns.TypeAXFR, dns.TypeIXFR:
 		return dns.RcodeRefused, nil
 	case dns.TypeNS:
@@ -82,7 +84,6 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 	m.Authoritative = true
 	m.Answer = append(m.Answer, records...)
 	m.Extra = append(m.Extra, extra...)
-
 	w.WriteMsg(m)
 	return dns.RcodeSuccess, nil
 }
--- a/plugin/kubernetes/handler.go
+++ b/plugin/kubernetes/handler.go
@@ -43,13 +43,13 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
 	case dns.TypeSRV:
 		records, extra, err = plugin.SRV(ctx, &k, zone, state, plugin.Options{})
 	case dns.TypeSOA:
-		if zone == state.Name() {
+		if len(state.Name()) == len(zone) {
 			records, err = plugin.SOA(ctx, &k, zone, state, plugin.Options{})
 		}
 	case dns.TypeAXFR, dns.TypeIXFR:
 		return dns.RcodeRefused, nil
 	case dns.TypeNS:
-		if state.Name() == zone {
+		if len(state.Name()) == len(zone) {
 			records, extra, err = plugin.NS(ctx, &k, zone, state, plugin.Options{})
 			break
 		}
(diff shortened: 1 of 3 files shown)


plugin/kubernetes/handler.go
```diff
@@ -43,11 +43,13 @@ func (k Kubernetes) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.M
...
 	case dns.TypeSOA:
-		records, err = plugin.SOA(ctx, &k, zone, state, plugin.Options{})
+		if len(state.Name()) == len(zone) {
+			records, err = plugin.SOA(ctx, &k, zone, state, plugin.Options{})
+		}
 	case dns.TypeAXFR, dns.TypeIXFR:
 		return dns.RcodeRefused, nil
 	case dns.TypeNS:
-		if state.Name() == zone {
+		if len(state.Name()) == len(zone) {
```
miekg: why does this need to be len? Same above

chrisohaver: This is done to make the comparison case insensitive.

v-byte-cpu: But it does not check the content, better to use `strings.EqualFold`

chrisohaver: The content is validated earlier.

https://github.com/coredns/coredns/blob/17e0b5895da708f918bdf95ee9659adf18dfb77c/plugin/kubernetes/handler.go#L17-L20

chrisohaver: I think `if qname == zone {` would work too... I think the cases match there... Name() lowercases.

I think comparing len requires less operations, but perhaps thats's insignificant.

v-byte-cpu: If I may put my two cents in, slight improvement in performance is important, but if it does not harm the readability of the code. If two people have asked this question, then perhaps other developers will also ask similar questions.

chrisohaver: OK, doing string compare now.

I've also reverted the changes to the NS case, so it's doing the name compare as it was before. Fixing the case insensitivity/preservation for the NS case is off topic anyways.  It can be fixed in another PR.

lastthyme (approved):

miekg (approved):

PR #4593, opened 2021-04-30
Share plugins among zones in the same server block

chrisohaver:

Revisiting #3862, which aimed to consolidate duplicate plugin chains for each zone in a single block to a one plugin chain. This reason for doing this is to reduce memory consumption for server blocks with multiple zones.

The problem: A server block with multiple zones will create a separate instance of each plugin in the block for each zone in the block.  This can become a problem for plugins that have a large memory overhead. It also can be a problem for server blocks with a large number of zones.

The approach in #3862 required modifying every plugin to skip setup if the current zone (server block key) was not the first zone in the block. Since then, we have forked caddy, so we can consider making modifications to the caddy package to solve this globally instead of requiring changes in every handler plugin.  As such, this PR requires a modification to `coredns/caddy/caddy.go` `executeDirectives` function, to skip executing the directives of extra keys in the server block IOW: To skip executing plugin setup functions for zones after the first zone listed. As follows:

```diff
func executeDirectives(inst *Instance, filename string,
	directives []string, sblocks []caddyfile.ServerBlock, justValidate bool) error {
	// map of server block ID to map of directive name to whatever.
	storages := make(map[int]map[string]interface{})

	// It is crucial that directives are executed in the proper order.
	// We loop with the directives on the outer loop so we execute
	// a directive for all server blocks before going to the next directive.
	// This is important mainly due to the parsing callbacks (below).
	for _, dir := range directives {
		for i, sb := range sblocks {
			var once sync.Once
			if _, ok := storages[i]; !ok {
				storages[i] = make(map[string]interface{})
			}

			for j, key := range sb.Keys {
				// Execute directive if it is in the server block
				if tokens, ok := sb.Tokens[dir]; ok {
					controller := &Controller{
						instance:  inst,
						Key:       key,
						Dispenser: caddyfile.NewDispenserTokens(filename, tokens),
						OncePerServerBlock: func(f func() error) error {
							var err error
							once.Do(func() {
								err = f()
							})
							return err
						},
						ServerBlockIndex:    i,
						ServerBlockKeyIndex: j,
						ServerBlockKeys:     sb.Keys,
						ServerBlockStorage:  storages[i][dir],
					}

+					// only set up directives for the first key in a block
+					if j > 0 {
+						continue
+					}
+
					setup, err := DirectiveAction(inst.serverType, dir)
					if err != nil {
						return err
					}

					err = setup(controller)
					if err != nil {
						return err
					}

					storages[i][dir] = controller.ServerBlockStorage // persist for this server block
				}
			}
		}

		if !justValidate {
			// See if there are any callbacks to execute after this directive
			if allCallbacks, ok := parsingCallbacks[inst.serverType]; ok {
				callbacks := allCallbacks[dir]
				for _, callback := range callbacks {
					if err := callback(inst.context); err != nil {
						return err
					}
				}
			}
		}
	}

	return nil
}
```

This can result in subtle but significant changes in behavior. For example, zones sharing a single instance of *cache* would collectively be bound by the cache's maximum size.  In CoreDNS today, each zone gets its own instance, and the collective max size is the cache max size (as defined in the server block) multiplied by the number of zones.


chrisohaver:
Running a test using _auto_ for 100 small zones (just few records each). 
e.g 
```
zone1.com zone2.com zone3.com [...] zone100.com {
  cache
  auto {
    directory auto/
  } 
}
```

Memory footprint CoreDNS today: ~90 MBs

With the plugin instance sharing described in this PR ( plus required caddy change ): ~9 MBs

chrisohaver:
Another memory test with 1000 zones in a block containing _errors_, _cache_, _prometheus_, and _forward_.

CoreDNS today: 190 MB
With this fix: 13 MB

johnbelamaric:
Awesome, will look later this week.

core/dnsserver/register.go
```diff
@@ -106,6 +117,16 @@ func (h *dnsContext) MakeServers() ([]caddy.Server, error) {
 		return nil, errValid
 	}
 
+	// Set the plugins of each config in a block to the plugins of the
+	// first config in the block. Doing this results in zones
+	// sharing the same plugin instances as other zones in
+	// the same block.
+	for _, c := range h.configs {
+		if c.firstConfigInBlock != nil {
```
miekg: isn't this always true?

miekg:
I think this is sane change and also make coredns do what folks expect it does in this case.

unclear if only the caddy change is needed, or caddy + this PR?

chrisohaver:
> I think this is sane change and also make coredns do what folks expect it does in this case.
> 
> unclear if only the caddy change is needed, or caddy + this PR?

Both are needed...
Without the CoreDNS changes, only the first zone in the server block would function. Other zones would have empty plugin chains.
Without the Caddy changes, separate plugin instances are created for each zone. The CoreDNS changes would unreference them, so they would never serve DNS.

chrisohaver:
I opened PR coredns/caddy#1 with the required change.  Once that merges and is tagged, will update go.mod here to use the new tag.

johnbelamaric:
This looks great. Such a small change but big impact :)

johnbelamaric (approved):

rajansandeep (approved):

chrisohaver:
307e8a7 resolves issues for _bind_ and _debug_, so all Configs from the same block use the first Config's ListenHosts and Debug settings.

miekg:
can you rebase/resolve the go.mod conflict, then lgtm and this can be merged.

chrisohaver:
@miekg, resolved go.mod conflicts.

chrisohaver:
OK to merge?

chrisohaver:
@miekg, I added a test to validate `Debug` is set in every `Config` in a block containing the _debug_ plugin.  It uses reflect and unsafe sorcery to access unexported elements of the related structs (thus avoiding need to refactor code in caddy for the sake of a test).

miekg:
let's do it!

Issue #4290, opened 2020-11-14
Wrong answers returned when wildcard and concrete records exist


SivaKesava1:
Hi,

In the presence of wildcards and concrete records, the CoreDNS server returns the wrong answer.

Consider the following zone file:

|  	|  	|  	|  	|
|-	|-	|-	|-	|
| example.com. 	| 500 | SOA 	| ns1.outside.com. root.example.com. 3 604800 86400 2419200 604800 	|
| example.com. 	| 500 | NS 	| ns1.outside.com.  	|
| *.d.example.com. 	| 500 | A | 1.1.1.1  	|
| r.c.d.example.com. 	| 500 | A	| 2.2.2.2  	|

For the query _<something.r.c.d.example.com. , A>_, the Coredns server returns the following response:

```
          "opcode QUERY",
          "rcode NOERROR",
          "flags QR AA",
          ";QUESTION",
          "something.r.c.d.example.com. IN A",
          ";ANSWER",
          "something.r.c.d.example.com. 500 IN A 1.1.1.1",
          ";AUTHORITY",
          "example.com. 500 IN NS ns1.outside.com.",
          ";ADDITIONAL"
```

**What you expected to happen**:
The expected response is as follows:
```
          "opcode QUERY",
          "rcode NXDOMAIN",
          "flags QR AA",
          ";QUESTION",
          "something.r.c.d.example.com. IN A",
          ";ANSWER",
          ";AUTHORITY",
          "example.com. 500 IN SOA ns1.outside.com. root.example.com. 3 604800 86400 2419200 604800 ",
          ";ADDITIONAL"
```

As per RFC 4592, the query _<something.r.c.d.example.com. , A>_ should not match the wildcard record _<*.d.example.com. , A>_ as _<r.c.d.example.com. , A>_  exists. It is considered the closest encloser, and since there is neither _`something`_ domain nor a wildcard record under it, the answer has to be `NXDOMAIN.` 
**CoreDNS exhibit the same behavior even for queries like _<something.c.d.example.com. , A>_.**  The expected response is `NXDOMAIN,` but CoreDNS returns the wildcard record.

**Environment**:

- the version of CoreDNS: `CoreDNS-1.8.0  linux/amd64, go1.15.3, 054c9ae`
- Corefile:
```
example.com:53 {
    file db.example.com
    log
    errors
}
```
- OS: Ubuntu 18

chrisohaver:
[...]

PR #4599, opened 2021-05-03
plugin/file: Fix in wrong answers returned when wildcard and concrete records exist

xdu31:
It fixes in file plugin wrong answers returned when wildcard and concrete records exist:
<pre>
; example.org test file with wildcard corner cases
$TTL 3600
example.org.		IN	SOA	sns.dns.icann.org. noc.dns.icann.org. 2015082541 7200 3600 1209600 3600  
example.org.		IN	NS	b.iana-servers.net.  
*.d.example.org.        IN      A       127.0.1.53  
r.c.d.example.org.      IN      A       127.0.1.52
</pre>
As per RFC 4592, the query `something.r.c.d.example.com. , A` should not match the wildcard record `*.d.example.com. , A` as `r.c.d.example.com. , A` exists. It is considered the closest encloser, and since there is neither something domain nor a wildcard record under it, the answer has to be NXDOMAIN.
CoreDNS exhibit the same behavior even for queries like `something.c.d.example.com. , A`. The expected response is NXDOMAIN, but CoreDNS returns the wildcard record.

https://github.com/coredns/coredns/issues/4290

No


chrisohaver:
Thanks! In the test case zone `cornerCasesWildcard `, `something.c.d.example.org.` should also be NXDOMAIN, since `c.d.example.org.` is a non-wild empty non-terminal. Can you add this test case? Its a different case than the `something.r.c.d.example.org.` test case, since  `r.c.d.example.org.` is not empty.

plugin/file/lookup.go
```diff
@@ -197,11 +197,22 @@ func (z *Zone) Lookup(ctx context.Context, state request.Request, qname string)
...
 
+	ce, foundCe = z.ClosestEncloser(qname)
+
 	// Haven't found the original name.
 
 	// Found wildcard.
 	if wildElem != nil {
 		auth := ap.ns(do)
+		if foundCe && dns.CountLabel(ce.Name())+1 > dns.CountLabel(wildElem.Name()) {
+			// wildcard denial
```
miekg: this comment is a bit terse to explain the previous if statement, please expand a bit

plugin/file/lookup.go
```diff
@@ -232,6 +243,7 @@ func (z *Zone) Lookup(ctx context.Context, state request.Request, qname string)
 			rrs = append(rrs, sigs...)
 
 		}
+
```
miekg: superfluous change

plugin/file/lookup.go
```diff
@@ -258,10 +270,8 @@ func (z *Zone) Lookup(ctx context.Context, state request.Request, qname string)
 			goto Out
 		}
 
-		ce, found := z.ClosestEncloser(qname)
-
 		// wildcard denial only for NXDOMAIN
-		if found {
+		if foundCe {
```
miekg: is `found` actually still used?

miekg:
can you more carefully explain (in the commit mesg), why foundCe/found and ce our now necessary ? Current msg is rather sparse or github is playing tricks on me

plugin/file/lookup.go
xdu31: > is `found` actually still used?

Yes, in a few other places

https://github.com/coredns/coredns/blob/d6d8a9974bc45855a7d8a89e9e452cc9b6301d0e/plugin/file/lookup.go#L95

https://github.com/coredns/coredns/blob/d6d8a9974bc45855a7d8a89e9e452cc9b6301d0e/plugin/file/lookup.go#L162

https://github.com/coredns/coredns/blob/d6d8a9974bc45855a7d8a89e9e452cc9b6301d0e/plugin/file/lookup.go#L167

https://github.com/coredns/coredns/blob/d6d8a9974bc45855a7d8a89e9e452cc9b6301d0e/plugin/file/lookup.go#L250

plugin/file/lookup.go
xdu31: Hi @miekg This is the case described in https://github.com/coredns/coredns/issues/4290, when the record "something.r.c.d.example.com" is matching "*.d.example.com" wildcard, "r.c.d.example.com", its closest encloser (CE) can be found, however, its CE is not directly under the wildcard (cntLabel(CE) + 1 > cntLabel(wildcard)), the record needs to be resolved as nxdomain.

xdu31:
> Thanks! In the test case zone `cornerCasesWildcard `, `something.c.d.example.org.` should also be NXDOMAIN, since `c.d.example.org.` is a non-wild empty non-terminal. Can you add this test case? Its a different case than the `something.r.c.d.example.org.` test case, since `r.c.d.example.org.` is not empty.

Hi @chrisohaver @SivaKesava1 Should`something.c.d.example.org.` also return NXDOMAIN? It looks to be the same case as `else.something.d.example.org`, it doesn't have a encloser under the wildcard `*.d.example.org` and is a multi-segment wildcard completion.

SivaKesava1:
> > Thanks! In the test case zone `cornerCasesWildcard `, `something.c.d.example.org.` should also be NXDOMAIN, since `c.d.example.org.` is a non-wild empty non-terminal. Can you add this test case? It's a different case than the `something.r.c.d.example.org.` test case, since `r.c.d.example.org.` is not empty.
> 
> Hi @chrisohaver @SivaKesava1 Should`something.c.d.example.org.` also return NXDOMAIN? It looks to be the same case as `else.something.d.example.org`, it doesn't have a encloser under the wildcard `*.d.example.org` and is a multi-segment wildcard completion.
 
Yes, it should be an NXDOMAIN. Consider the domain names of the records as a tree, then follow the tree label by label from the root and stop when you can go further due to a non-match. If the node at which you stopped has a wildcard child, then you will perform wildcard synthesis; otherwise, you would return an NXDOMAIN.

```
    org
      |
   example
      |
      d
     / \
    c   *
    |
    r
```

When doing the matching, `something.c.d.example.org.` stops at the `c` node, and it does not have a wildcard child, so an NXDOMAIN. For `else.something.d.example.org,` it stops the matching at the `d` node, and it has a wildcard child, so it is a wildcard synthesis.
I am not aware of the internal data structures used by CoreDNS, so I can not elaborate further, but the above explanation is based on how RFCs view the zone file.

xdu31:
> > > Thanks! In the test case zone `cornerCasesWildcard `, `something.c.d.example.org.` should also be NXDOMAIN, since `c.d.example.org.` is a non-wild empty non-terminal. Can you add this test case? It's a different case than the `something.r.c.d.example.org.` test case, since `r.c.d.example.org.` is not empty.
> > 
> > 
> > Hi @chrisohaver @SivaKesava1 Should`something.c.d.example.org.` also return NXDOMAIN? It looks to be the same case as `else.something.d.example.org`, it doesn't have a encloser under the wildcard `*.d.example.org` and is a multi-segment wildcard completion.
> 
> Yes, it should be an NXDOMAIN. Consider the domain names of the records as a tree, then follow the tree label by label from the root and stop when you can go further due to a non-match. If the node at which you stopped has a wildcard child, then you will perform wildcard synthesis; otherwise, you would return an NXDOMAIN.
> 
> ```
>     org
>       |
>    example
>       |
>       d
>      / \
>     c   *
>     |
>     r
> ```
> 
> When doing the matching, `something.c.d.example.org.` stops at the `c` node, and it does not have a wildcard child, so an NXDOMAIN. For `else.something.d.example.org,` it stops the matching at the `d` node, and it has a wildcard child, so it is a wildcard synthesis.
> I am not aware of the internal data structures used by CoreDNS, so I can not elaborate further, but the above explanation is based on how RFCs view the zone file.

@SivaKesava1 Thanks for the clarification, I need more changes for handling this above case.

xdu31:
@chrisohaver @miekg @SivaKesava1 
1. Re-implementation: If the domain's terminal is neither the matching wildcard, nor a domian directly
under the wildcard, return NXDOMAIN
2. New test case: `something.c.d.example.org.`

plugin/file/lookup.go
```diff
@@ -239,8 +255,8 @@ func (z *Zone) Lookup(ctx context.Context, state request.Request, qname string)
 
 	// Hacky way to get around empty-non-terminals. If a longer name does exist, but this qname, does not, it
 	// must be an empty-non-terminal. If so, we do the proper NXDOMAIN handling, but set the rcode to be success.
-	if x, found := tr.Next(qname); found {
-		if dns.IsSubDomain(qname, x.Name()) {
+	if nextElem != nil {
```
chrisohaver: That this logic has essentially changed from `tr.Next(qname)` to `tr.Next(parts)`, seems like it might break empty non-terminal resolution.

chrisohaver: > That this logic has essentially changed from `tr.Next(qname)` to `tr.Next(parts)`, seems like it might break empty non-terminal resolution.

I don't see an existing test that validates queries for empty non-terminals should get a 'NOERROR' response. Should add it if one doesn't currently exist. e.g. in `lookup_test.go`...

`dbMiekNL` +=
```
test.ent	IN	A	139.162.196.79
```

and `dnsTestCases` +=
```
	{
		Qname: "ent.miek.nl.", Qtype: dns.TypeA,
		Ns: []dns.RR{
			test.SOA("miek.nl.	1800	IN	SOA	linode.atoom.net. miek.miek.nl. 1282630057 14400 3600 604800 14400"),
		},
	},
```

xdu31: Hi @chrisohaver I reverted back the tr.Next(qname) and added a new test case in lookup_test.go.

chrisohaver: Cool - thanks.  Looks like the other two wildcard related tests in azure and clouddns are passing now after that change too.

plugin/file/lookup.go
```diff
@@ -201,8 +207,18 @@ func (z *Zone) Lookup(ctx context.Context, state request.Request, qname string)
 
 	// Found wildcard.
 	if wildElem != nil {
-		auth := ap.ns(do)
+		// if the domain's ternimal is neither the matching wildcard, nor a domain directly under the wildcard:
```
chrisohaver:
```suggestion
		// if the domain's terminal is neither the matching wildcard, nor a domain directly under the wildcard:
```

plugin/file/lookup.go
xdu31: Hi @chrisohaver, the idea is to iteratively find the next record that is the subdomain of the parts, it will finally find the last(longest) matching subdomain record.

For the zone file in `cornerCasesWildcard`
<pre>
     org
      |
   example
      |
      d
     / \
    c   *
    |
    r
</pre>

`example.org.`, ne is `*.d.example.org.`
`d.example.org.` ne is `*.d.example.org.`

`c.d.example.org.` ne is `r.c.d.example.org.`
`r.c.d.example.org.` ne is `r.c.d.example.org.`
...

`something.d.example.org.` ne is `*.d.example.org.` (inherent from `d.example.org.` as `*.d.example.org.` is not a parent of `something.d.example.org.`)
...

If number of labels of ne is greater than number of labels of the matching wildcard, we want to return NXDOMAINS

plugin/file/lookup.go
```diff
@@ -92,6 +92,12 @@ func (z *Zone) Lookup(ctx context.Context, state request.Request, qname string)
 			break
 		}
 
+		if nextElem, found = tr.Next(parts); found {
```
chrisohaver, line 95: Tests fail if the following record is added to the zone: `x.&.d.example.org.        IN      A       127.0.1.53`.

I think it's because of an assumption that a `*` element will always be ordered in the tree before any other element in the same sub domain (and `&` is sorted before `*`).  
E.g. the following record does not cause tests to fail: `x.+.d.example.org.        IN      A       127.0.1.53` presumably because `+` sorts after `*`.

plugin/file/wildcard_test.go
```diff
@@ -296,3 +366,13 @@ example.org.		IN	NS	b.iana-servers.net.
...
 foo.example.org.        IN      A       127.0.0.54
 `
+
+const cornerCasesWildcard = `; example.org test file with wildcard corner cases
+$TTL 3600
+example.org.		IN	SOA	sns.dns.icann.org. noc.dns.icann.org. 2015082541 7200 3600 1209600 3600
+example.org.		IN	NS	b.iana-servers.net.
+*.d.example.org.        IN      A       127.0.1.53
+z.+.d.example.org.		IN		A		127.0.1.53
+x.&.d.example.org.		IN		A		127.0.1.53
```
chrisohaver: These IPs should be different from the wild card entry IP - So that an answer from these records can be discerned from an (incorrect) answer synthesized from the wildcard record.

chrisohaver (changes requested):
Looks good.  The latest version of the logic feels more intuitive IMO. Thanks!

One minor comment about the test data, but other than that LGTM.

plugin/file/wildcard_test.go
xdu31: @chrisohaver Change made.

chrisohaver (approved):
LGTM!  IMO an owner of the file plugin should review before merging.

xdu31:
Hi @miekg @stp-ip @yongtang could you help review the PR?

yongtang (approved):
LGTM

plugin/file/lookup.go
miekg: sorting and tr.Next don't do what you think. 

and in general I think this PR just happens to implement what the tests hand it

miekg:
meh, can't reopen old PRs.

anyways, finally have some time to comment/take a look. So the issue is/was that the current lookup algo should not have to change in response to this change. Adding a new element and taking Next() is thus probably not the right choice.

Also, I didn't think the tree is sorted and the layout of the elements depends on the insertion of them. I know, 'cause I tried doing this with NSEC and it did not work, although my tests all proved correct, but then I tried with larger zones and the whole thing fell apart.

chrisohaver:
FYI, @xdu31 copying @miekg's comments from https://github.com/coredns/coredns/pull/4599#issuecomment-850348435

> anyways, finally have some time to comment/take a look. So the issue is/was that the current lookup algo should not have to change in response to this change. Adding a new element and taking Next() is thus probably not the right choice.
>
> Also, I didn't think the tree is sorted and the layout of the elements depends on the insertion of them. I know, 'cause I tried doing this with NSEC and it did not work, although my tests all proved correct, but then I tried with larger zones and the whole thing fell apart.

PR #4606, opened 2021-05-05
DNSSEC check for two days of remaining validity

### LINKED ISSUE #4605: DNSSEC cached signatures prematurely deleted
**What happened**:

Cached DNSSEC sigs are deleted after two days.

Cached sigs should live for six days per #1378.

Rebuild with cache timings in seconds instead of days.

PR incoming

kcolemangt:
Signed-off-by: Keith C <keith@fraudmarc.com>

This should allow cached signatures to be cached for the intended six day duration.
fix #4605 
No


miekg:
is this testable?

PR #4621, opened 2021-05-12
reverse zone: make Normalize return proper reverse zones

miekg:
Make normalize return multiple "hosts" (= reverse zones) when a
non-octet boundary cidr is given.

Added pkg/cidr package that holds the cidr calculation routines; felt
they didn't really fit dnsutil.

This change means the IPNet return parameter isn't needed, the hosts are
all correct. The tests that tests this is also removed: TestSplitHostPortReverse
The fallout was that zoneAddr _also_ doesn't need the IPNet member, that
in turn make it visible that zoneAddr in address.go duplicated a bunch
of stuff from register.go; removed/refactored that too.

Created a plugin.OriginsFromArgsOrServerBlock to help plugins do the
right things, by consuming ZONE arguments; this now expands reverse
zones correctly.

This change is not backwards compatible to plugins that have a ZONE
argument that they parse in the setup util.

All in-tree plugins have been updated.


miekg:
<strike>there are still some test failures I dont understand yet:
~~~
--- FAIL: TestFileUpstreamError (0.01s)
    --- FAIL: TestFileUpstreamError/nxdomain (0.00s)
        file_upstream_test.go:156: expected rcode 3, got 2
    --- FAIL: TestFileUpstreamError/nxdomain-chain (0.00s)
        file_upstream_test.go:156: expected rcode 3, got 2
--- FAIL: TestFileUpstreamAdditional (0.00s)
    file_upstream_test.go:214: Rcode should not be dns.RcodeServerFailure
Not executing TestPrometheusImports: faillint not found
--- FAIL: TestTemplateUpstream (0.00s)
    template_upstream_test.go:41: Rcode should not be dns.RcodeServerFailure
FAIL
exit status 1
~~~
</strike>

<strike>and in *template*:
~~~
-- FAIL: TestMultiSection (0.00s)
    template_test.go:506: TestMultiSection expected response code REFUSED got: 3841
FAIL
exit status 1
~~~
for the latter it seems the test is actually wrong, but didn't fail before... :/ </strike>

Ok, there are some setups, that are somewhat tricky and tried optimizing them. Reverted

yongtang (approved):
LGTM

miekg:
hmmm it's in now, but may have been more prudent to wait for <strike>1.8.4</strike> 1.8.5

miekg:
this needs another PR because go-cidr barfs on it:
~~~
diff --git plugin/normalize.go plugin/normalize.go
index 96ec59c7..8724de53 100644
--- plugin/normalize.go
+++ plugin/normalize.go
@@ -108,11 +108,16 @@ func SplitHostPort(s string) (hosts []string, port string, err error) {
        }
 
        // Check if it parses as a reverse zone, if so we use that. Must be fully specified IP and mask.
-       _, n, err := net.ParseCIDR(s)
+       ip, n, err := net.ParseCIDR(s)
        if err != nil {
                return []string{s}, port, nil
        }
-
+       if ip.To4() == nil { // v6 address, if the mask if not on a octet, it's not a valid cidr (we don't need to split v6 like v4)
+               ones, _ := n.Mask.Size()
+               if ones%8 != 0 {
+                       return []string{s}, port, nil
+               }
+       }
        // now check if multiple hosts must be returned.
        nets := cidr.Class(n)
        hosts = cidr.Reverse(nets)
~~~

plugin/loop/README.md
```diff
@@ -13,6 +13,8 @@ The plugin will try to send the query for up to 30 seconds. This is done to give
 to start up. Once a query has been successfully sent, *loop* disables itself to prevent a query of
 death.
 
+Note that *loop* will _only_ send "looping queries" for the first zone given in the Server Block.
+
```
chrisohaver, line 17: Should mention the CIDR notation effect here. First zone of a CIDR notation "zone" = largest lowest prefix?  (or whatever the order of zones produced by `OriginsFromArgsOrServerBlock`)

plugin/forward/setup.go
```diff
@@ -92,7 +92,7 @@ func parseStanza(c *caddy.Controller) (*Forward, error) {
 	if !c.Args(&f.from) {
 		return f, c.ArgErr()
 	}
-	f.from = plugin.Host(f.from).Normalize()
+	f.from = plugin.Host(f.from).Normalize()[0] // there can only be one here, won't work with non-octet reverse
```
chrisohaver, line 95: Should be in README.

chrisohaver: Until it is supported, we should probably error out if `from` Normalizes to > 1 zone.

chrisohaver: Opened #4639 to address this in README and as a start up _warning_.

PR #4636, opened 2021-05-19
Check octect boundary in v6 addr.

miekg:
This prevents a crash when calling into go-cidr.

### DIFF
--- a/plugin/normalize.go
+++ b/plugin/normalize.go
@@ -158,11 +158,16 @@ func SplitHostPort(s string) (hosts []string, port string, err error) {
 	}
 
 	// Check if it parses as a reverse zone, if so we use that. Must be fully specified IP and mask.
-	_, n, err := net.ParseCIDR(s)
+	ip, n, err := net.ParseCIDR(s)
 	if err != nil {
 		return []string{s}, port, nil
 	}
-
+	if ip.To4() == nil { // v6 address, if the mask if not on a octet, it's not a valid cidr (we don't need to split v6 like v4)
+		ones, _ := n.Mask.Size()
+		if ones%8 != 0 {
+			return nil, "", fmt.Errorf("non-octet boundary is not supported for a v6 reverse: %s", ip)
+		}
+	}
 	// now check if multiple hosts must be returned.
 	nets := cidr.Class(n)
 	hosts = cidr.Reverse(nets)


miekg:
also found by fuzzing, /cc @yongtang with the normalize stuff in, this would be good to include as well.

plugin/normalize.go
```diff
@@ -108,11 +108,16 @@ func SplitHostPort(s string) (hosts []string, port string, err error) {
 	}
 
 	// Check if it parses as a reverse zone, if so we use that. Must be fully specified IP and mask.
-	_, n, err := net.ParseCIDR(s)
+	ip, n, err := net.ParseCIDR(s)
 	if err != nil {
 		return []string{s}, port, nil
 	}
-
+	if ip.To4() == nil { // v6 address, if the mask if not on a octet, it's not a valid cidr (we don't need to split v6 like v4)
```
chrisohaver, line 165: Is this true? I think its typical that ipv6 prefix fall on boundaries, but I didnt think it was required.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Check octect ..." ]
>
>
>In plugin/normalize.go:
>
>>       if err != nil {
>                return []string{s}, port, nil
>        }
>-
>+       if ip.To4() == nil { // v6 address, if the mask if not on a octet, it's not a valid cidr (we don't need to split v6 like v4)
>
>Is this true? I think its typical that ipv6 prefix fall on boundaries, but I
>didnt think it was required.

I would be amazed if it is written down in an RFC somewhere, but non-octet boundaries only
make sense for v4, with the limited address space.

In a mathematical sense; yes sure this can be done, in go-cidr an uint overflowed when
allocating the number of subnets...

chrisohaver:
> I would be amazed if it is written down in an RFC somewhere, but non-octet boundaries only make sense for v4, with the limited address space. In a mathematical sense; yes sure this can be done, in go-cidr an uint overflowed when allocating the number of subnets...

Makes sense.  Should we error out when we encounter a non-boundary ipv6 prefix instead of accepting it as a literal zone?

miekg:
thanks, I agree, erroring seems the saner option. I'll make that change.

miekg:
updated

plugin/normalize.go
```diff
@@ -158,11 +158,16 @@ func SplitHostPort(s string) (hosts []string, port string, err error) {
...
+		ones, _ := n.Mask.Size()
+		if ones%8 != 0 {
+			return nil, "", fmt.Errorf("non-octet boundary is not supported for a v6 reverse: %s", ip)
```
chrisohaver, line 168: For ipv6, reverse zones are divided into segments of 4 bits (one hex per segment).
So, should this be `if ones%4 != 0 {` ... and `"non-quartet boundary is not supported`

miekg: you're right! Also https://www.ripe.net/manage-ips-and-asns/db/support/configuring-reverse-dns uses /29 which negates this whole PR. Let me put the example in there in a test.

It may well be that 5d80a6e2 fixed the crash in go-cidr and this is now some superficial flavoring on top of it

miekg:
ok, so this Corefile
~~~
::fFFF:B:F/115
~~~
crashes on master with:
~~~
math/big.nat.make(...)
	/home/miek/up/go/src/math/big/nat.go:69
math/big.nat.shl(0xc00028bf50, 0x1, 0x1, 0xc00028bf50, 0x1, 0x1, 0xffffffffffffffa8, 0xc00002e2d0, 0x30, 0x20)
	/home/miek/up/go/src/math/big/nat.go:1010 +0x2a5
math/big.(*Int).Lsh(...)
	/home/miek/up/go/src/math/big/int.go:1009
github.com/apparentlymart/go-cidr/cidr.AddressRange(0xc00002e2d0, 0xc00078a7a0, 0xc00091f750, 0x4163f0, 0x0, 0x0, 0x7f623b84a5c0)
	/home/miek/go/pkg/mod/github.com/apparentlymart/go-cidr@v1.1.0/cidr/cidr.go:124 +0x165
github.com/apparentlymart/go-cidr/cidr.NextSubnet(0xc00002e2d0, 0x78, 0x1, 0x1)
	/home/miek/go/pkg/mod/github.com/apparentlymart/go-cidr@v1.1.0/cidr/cidr.go:189 +0x45
github.com/coredns/coredns/plugin/pkg/cidr.nets(0xc00002e270, 0x78, 0x10, 0x73, 0xa)
	/home/miek/src/github.com/coredns/coredns/plugin/pkg/cidr/cidr.go:35 +0x225
github.com/coredns/coredns/plugin/pkg/cidr.Class(0xc00002e270, 0xe, 0xc00028be90, 0x10)
	/home/miek/src/github.com/coredns/coredns/plugin/pkg/cidr/cidr.go:20 +0xb3
github.com/coredns/coredns/plugin.SplitHostPort(0xc0002aca80, 0xe, 0x206d7be, 0x3, 0xc0002aca80, 0xe, 0x1, 0xc000425e60, 0x0)
	/home/miek/src/github.com/coredns/coredns/plugin/normalize.go:167 +0x1cb
github.com/coredns/coredns/core/dnsserver.(*dnsContext).InspectServerBlocks(0xc000425e80, 0x20729d7, 0x8, 0xc000425e60, 0x1, 0x1, 0xc000425e60, 0x1, 0x1, 0x0, ...)
	/home/miek/src/github.com/coredns/coredns/core/dnsserver/register.go:68 +0xbec
github.com/coredns/caddy.ValidateAndExecuteDirectives(0x23b2910, 0xc00022b900, 0xc00032e700, 0x0, 0x1, 0xc000416e48)
	/home/miek/go/pkg/mod/github.com/coredns/caddy@v1.1.0/caddy.go:607 +0x2ca
github.com/coredns/caddy.startWithListenerFds(0x23b2910, 0xc00022b900, 0xc00032e700, 0x0, 0x0, 0x0)
	/home/miek/go/pkg/mod/github.com/coredns/caddy@v1.1.0/caddy.go:515 +0x135
github.com/coredns/caddy.Start(0x23b2910, 0xc00022b900, 0x23b2910, 0xc00022b900, 0x0)
	/home/miek/go/pkg/mod/github.com/coredns/caddy@v1.1.0/caddy.go:472 +0xf1
github.com/coredns/coredns/coremain.Run()
	/home/miek/src/github.com/coredns/coredns/coremain/run.go:63 +0x18e
main.main()
	/home/miek/src/github.com/coredns/coredns/coredns.go:12 +0x25
~~~

miekg:
maybe 'a' fix is to just check if the mask > X?

miekg:
I was waiting for this for 1.8.4 but lets defer it until 185

chrisohaver:
> maybe 'a' fix is to just check if the mask > X?

Was just thinking the same thing. Although, if the mask/prefix is on a 4 bit boundary, then we should be able to create the ipv6 reverse zone without doing any "math" regardless of length.

chrisohaver:
The following pass OK with existing code...

```
		{"2001:db8::/29", []string{
			"8.b.d.0.1.0.0.2.ip6.arpa.",
			"9.b.d.0.1.0.0.2.ip6.arpa.",
			"a.b.d.0.1.0.0.2.ip6.arpa.",
			"b.b.d.0.1.0.0.2.ip6.arpa.",
			"c.b.d.0.1.0.0.2.ip6.arpa.",
			"d.b.d.0.1.0.0.2.ip6.arpa.",
			"e.b.d.0.1.0.0.2.ip6.arpa.",
			"f.b.d.0.1.0.0.2.ip6.arpa.",
		}},
		{"2001:db8::/115", []string{
			"0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa.",
			"1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa.",
			"2.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa.",
			"3.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa.",
			"4.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa.",
			"5.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa.",
			"6.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa.",
			"7.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa.",
		}},
```

I think the problem with `::fFFF:B:F/115` may be that it has non-zero values outside of the mask, which are normally ignored  but might be confusing the algorithm we use.

chrisohaver:
Fiddling around a bit more, I think ipv6 produces incorrect answers in some cases.  I think in part due to `net.Class` assuming ipv4 boundary size.

chrisohaver:
I think the code below fixes the wrong answers for normal ipv6 cidrs. 

It also appears to prevent the crash for the abnormal `::fFFF:B:F/115` case... but it produces an odd/wrong answer... `["11.0.in-addr.arpa.", "11.0.in-addr.arpa."]` ... i'm guessing that it's not being detected as a valid IPv6 mask (due to the leading zeros), and falling back to ipv4 processing. We could expressly check for iv6 with leading zeros and error out.

```go
// Class return slice of "classful" (/8, /16, /24 or /32 only) CIDR's from the CIDR in net.
func Class(n *net.IPNet) []string {
	boundary := 8
	if ipv6 := n.IP.To16(); ipv6 != nil {
		boundary = 4
	}
	ones, _ := n.Mask.Size()
	if ones%boundary == 0 {
		return []string{n.String()}
	}

	mask := int(math.Ceil(float64(ones)/float64(boundary))) * boundary
	networks := nets(n, mask)
	cidrs := make([]string, len(networks))
	for i := range networks {
		cidrs[i] = networks[i].String()
	}
	return cidrs
}
```

chrisohaver:
Opened PR #4652 with fix [above](https://github.com/coredns/coredns/pull/4636#issuecomment-849829000).

miekg:
this pr is not correct, closing

PR #4637, opened 2021-05-19
Fix obsure crash in Corefile parsing

miekg:
This was found by fuzzing.

We need to make this a fully qualified domain name to catch all errors
in dnsserver/register.go and not later when plugin.Normalize() is called again on these
strings, with the prime difference being that the domain name is fully
qualified. This was found by fuzzing where "ȶ" is deemed OK, but "ȶ." is
not (might be a bug in miekg/dns actually). But here we were checking ȶ,
which is OK, and later we barf in ȶ. leading to "index out of range".

Added a tests and check manually if it would crash with the current code
(yes), and fail with an error in this PR (yes).


miekg:
similar to #4636, /cc @yongtang

yongtang (approved):
I tested locally, was able to reproduce the crash and the PR fix. Thanks for the fix!

PR #4647, opened 2021-05-25
plugin/kubernetes: consistently choose EndpointSlices or Endpoints for unstable API

uesyn:
Signed-off-by: uesyn <suemura@zlab.co.jp>

In the current implementation, CoreDNS gives up the use of EndpointSlices and falls back to Endpoints when Kubernetes API is temporarily unavailable. This behavior may cause inconsistent query results if there are multiple CoreDNS instances, and some use EndpointSlices, others use Endpoints.
EndpointSlices are not mirrored to Endpoints, so CoreDNS may fail to resolve a name for a service that has only EndpointSlice as backend if this problem occurs.
This PR makes CoreDNS consistently choose EndpointSlices or Endpoints.

### DIFF
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -22,6 +22,7 @@ import (
 	api "k8s.io/api/core/v1"
 	discovery "k8s.io/api/discovery/v1"
 	discoveryV1beta1 "k8s.io/api/discovery/v1beta1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
@@ -301,8 +302,6 @@ func (k *Kubernetes) InitKubeCache(ctx context.Context) (onStart func() error, o
 // v1 is not supported.
 // This function should be removed, when all supported versions of k8s support v1.
 func (k *Kubernetes) endpointSliceSupported(kubeClient *kubernetes.Clientset) (bool, string) {
-	var sliceVer string
-	useEndpointSlices := false
 	ticker := time.NewTicker(100 * time.Millisecond)
 	defer ticker.Stop()
 	for {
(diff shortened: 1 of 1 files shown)


plugin/kubernetes/kubernetes.go
```diff
@@ -301,36 +302,48 @@ func (k *Kubernetes) InitKubeCache(ctx context.Context) (onStart func() error, o
...
-	var sliceVer string
-	useEndpointSlices := false
 	ticker := time.NewTicker(100 * time.Millisecond)
 	defer ticker.Stop()
 	for {
 		select {
 		case <-ticker.C:
 			sv, err := kubeClient.ServerVersion()
 			if err != nil {
+				log.Warningf("Failed to get kubernetes version: %s", err)
```
chrisohaver: Please remove this warning, I think this would be too verbose.  It would get logged 10 times per second until the API is up.

plugin/kubernetes/kubernetes.go
```diff
@@ -301,36 +302,48 @@ func (k *Kubernetes) InitKubeCache(ctx context.Context) (onStart func() error, o
...
+				return false, ""
 			}
-			return useEndpointSlices, sliceVer
+
+			// Enable use of endpoint slices if the API supports the discovery api
+			_, err = kubeClient.Discovery().ServerResourcesForGroupVersion(discovery.SchemeGroupVersion.String())
+			if err == nil {
+				return true, discovery.SchemeGroupVersion.String()
+			} else if !kerrors.IsNotFound(err) {
+				log.Warningf("Failed to discovery %s: %s", discovery.SchemeGroupVersion.String(), err)
```
chrisohaver: This is also too verbose, if this state can persist for more than a millisecond. I think it can be removed

plugin/kubernetes/kubernetes.go
```diff
@@ -301,36 +302,48 @@ func (k *Kubernetes) InitKubeCache(ctx context.Context) (onStart func() error, o
...
+				continue
+			}
+
+			_, err = kubeClient.Discovery().ServerResourcesForGroupVersion(discoveryV1beta1.SchemeGroupVersion.String())
+			if err == nil {
+				return true, discoveryV1beta1.SchemeGroupVersion.String()
+			} else if !kerrors.IsNotFound(err) {
+				log.Warningf("Failed to discovery %s: %s", discoveryV1beta1.SchemeGroupVersion.String(), err)
```
chrisohaver: Same as above, re: verbosity.

plugin/kubernetes/kubernetes.go
```diff
@@ -301,36 +302,48 @@ func (k *Kubernetes) InitKubeCache(ctx context.Context) (onStart func() error, o
...
+				continue
+			}
+
+			// Disable use of endpoint slices in case that it is disabled in k8s versions 1.19 and newer.
+			log.Info("Watching Endpoints instead of EndpointSlices")
```
chrisohaver:
```suggestion
			log.Info("Endpointslices API disabled. Watching Endpoints instead. ")
```

chrisohaver (changes requested):
Thank you! I didn't think this would be an issue because the server version is checked first, and I had assumed that if the Kubernetes API could answer that, then the API was "available."  But I presume you have _observed_ this is not the case? Or is this to protect against the API becoming unavailable in between checking the version and checking for Endpointslice support?

Looks like the key change here is checking for errors other than the "not found" error.

I've suggested some changes to logged warnings - some of which would be logged too frequently while the API is not available.

uesyn:
@chrisohaver 
Thank you for your review! I applied all your suggestions!

> Thank you! I didn't think this would be an issue because the server version is checked first, and I had assumed that if the Kubernetes API could answer that, then the API was "available." But I presume you have observed this is not the case? Or is this to protect against the API becoming unavailable in between checking the version and checking for Endpointslice support?

I haven't observed at CoreDNS yet, but I think this could happen.
I tested the below code in the situation that kube-apiserver returns too many requests and got an error as expected.
This situation may be a rare case, but it is worth taking into account because the choice of either EndpointSlices or Endpoints affects until terminating a CoreDNS.

```go
func main() {
	clientset, err := NewClientSet()
	if err != nil {
		log.Fatalf("Failed to get clientset: %s", err)
	}

	for {
		var err error

		_, err = clientset.ServerVersion()
		if err != nil {
			continue
		}

		if _, err = clientset.Discovery().ServerResourcesForGroupVersion(discovery.SchemeGroupVersion.String()); err == nil {
			// Here is unreachable. In k8s v1.19, always got error, because K8s v1.19 doesn't support discovery v1.
			continue
		}

		if _, err = clientset.Discovery().ServerResourcesForGroupVersion(discoveryv1beta1.SchemeGroupVersion.String()); err == nil {
			continue
		}

		log.Printf("Got error: %s", err)
	}
}
```

chrisohaver (approved):
Thanks! I think this is safe and could protect against cases where the connection to the API is unstable.

PR #4648, opened 2021-05-25
deprecate Normalize and MustNormalize

chrisohaver:

The recent signature change to `Normalize` (#4621) is not compatible with existing external plugins.
To ease the transition, this PR ...
* Renames the fixed version of `Normalize` to `NormalizeExact`
* Restore the original `Normalize` and `MustNormalize` and documents them as Deprecated.

#4621

### DIFF
--- a/plugin/normalize.go
+++ b/plugin/normalize.go
@@ -63,8 +63,34 @@ type (
 
 // Normalize will return the host portion of host, stripping
 // of any port or transport. The host will also be fully qualified and lowercased.
+// An empty string is returned on failure
+// Deprecated: use OriginsFromArgsOrServerBlock or NormalizeExact
+func (h Host) Normalize() string {
+	// The error can be ignored here, because this function should only be called after the corefile has already been vetted.
+	host, _ := h.MustNormalize()
+	return host
+}
+
+// MustNormalize will return the host portion of host, stripping
+// of any port or transport. The host will also be fully qualified and lowercased.
+// An error is returned on error
+// Deprecated: use OriginsFromArgsOrServerBlock or NormalizeExact
+func (h Host) MustNormalize() (string, error) {
+	s := string(h)
+	_, s = parse.Transport(s)
+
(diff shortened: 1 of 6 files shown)


chrisohaver:
I'm not married to `NormalizeExact` ... but I found it less awkward than other candidates `NormalizeAll` `NormalizeFull` `Normalize2` `NormalizeWithCompleteCidrReverseZoneExpansion`

plugin/normalize.go
```diff
@@ -63,8 +63,34 @@ type (
 
 // Normalize will return the host portion of host, stripping
 // of any port or transport. The host will also be fully qualified and lowercased.
+// An empty string is returned on failure
+// Deprecated: use OriginsFromArgsOrServerBlock or NormalizeExact
+func (h Host) Normalize() string {
+	// The error can be ignored here, because this function should only be called after the corefile has already been vetted.
```
johnbelamaric: It would be nice to emit a warning, ideally at build time but since these are only called in setup, I would find runtime acceptable too.

chrisohaver: I believe `golint` will flag the use of deprecated functions (identified by "Deprecated:" in function comments).  But we don't automate code linting.

I can add a runtime warning.

plugin/normalize.go
```diff
@@ -76,6 +77,7 @@ func (h Host) Normalize() string {
 // An error is returned on error
 // Deprecated: use OriginsFromArgsOrServerBlock or NormalizeExact
 func (h Host) MustNormalize() (string, error) {
+	log.Warning("Host.Normalize and Host.MustNormalize are Deprecated. Use OriginsFromArgsOrServerBlock or NormalizeExact.")
```
johnbelamaric: This will confuse the ordinary user. I might say something like:

> An external plugin is using the deprecated x or y function. These will be removed in future versions of CoreDNS. The plugin should be updated to use a or b instead.

If we can identify the plugin from this location that would be even better.

PR #4650, opened 2021-05-26
Reduce the cardinality of health endpoint metrics

SuperQ:

The health endpoint histogram has a large amount of cardinality for a
simple endpoint. Introduce a new "Slim" set of buckets for `/health` to
reduce the metrics load on large deployments. Especially those that have
per-node DNS caching services.

Add a metric to count internal health check failures rather than use the
timeout value as side effect monitor of the check error. This avoids
incorrectly recording the timeout value if there is an error that is not
a timeout (ex. refused)

Updated health plugin README.

No

### DIFF
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -105,5 +105,8 @@ const Namespace = "coredns"
 // TimeBuckets is based on Prometheus client_golang prometheus.DefBuckets
 var TimeBuckets = prometheus.ExponentialBuckets(0.00025, 2, 16) // from 0.25ms to 8 seconds
 
+// SlimTimeBuckets is low cardinality set of duration buckets.
+var SlimTimeBuckets = prometheus.ExponentialBuckets(0.00025, 10, 5) // from 0.25ms to 2.5 seconds
+
 // ErrOnce is returned when a plugin doesn't support multiple setups per server.
 var ErrOnce = errors.New("this plugin can only be used once per Server Block")
--- a/plugin/health/README.md
+++ b/plugin/health/README.md
@@ -50,11 +50,13 @@ Doing this is supported but both endpoints ":8080" and ":8081" will export the e
 
 If monitoring is enabled (via the *prometheus* plugin) then the following metric is exported:
 
- *  `coredns_health_request_duration_seconds{}` - duration to process a HTTP query to the local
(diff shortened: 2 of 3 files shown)


chrisohaver (approved):
Seems sensible to me.

SuperQ:
I looked over the metrics for this a bit and decided to add an additional change to add a health check failure metric, rather than imply failure by observing the timeout value.

plugin/plugin.go
```diff
@@ -105,5 +105,8 @@ const Namespace = "coredns"
 // TimeBuckets is based on Prometheus client_golang prometheus.DefBuckets
 var TimeBuckets = prometheus.ExponentialBuckets(0.00025, 2, 16) // from 0.25ms to 8 seconds
 
+// SlimTimeBuckets is low cardinality set of duration buckets.
+var SlimTimeBuckets = prometheus.ExponentialBuckets(0.00025, 10, 5) // from 0.25ms to 2.5 seconds
```
miekg, line 109: I seen this be 6s in k8s deployments.....

miekg:
lgtm, but see my other comment

plugin/plugin.go
SuperQ: Are you talking about the health endpoint? That just means it fits in the `+Inf` bucket.

But, how can it be 6s when the timeout is 3s?

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Reduce the ca..." ]
>@SuperQ commented on this pull request.
>
>
>In plugin/plugin.go:
>
>> @@ -105,5 +105,8 @@ const Namespace = "coredns"
> // TimeBuckets is based on Prometheus client_golang prometheus.DefBuckets
> var TimeBuckets = prometheus.ExponentialBuckets(0.00025, 2, 16) // from 0.25ms to 8 seconds
>
>+// SlimTimeBuckets is low cardinality set of duration buckets.
>+var SlimTimeBuckets = prometheus.ExponentialBuckets(0.00025, 10, 5) // from 0.25ms to 2.5 seconds
>
>Are you talking about the health endpoint? That just means it fits in the +Inf
>bucket.

fair


>But, how can it be 6s when the timeout is 3s?

I don't now (and didn't care enough to delve deep), but even a timeout for a local request
is kinda crazy

PR #4652, opened 2021-05-27
Fix IPv6 case for CIDR format reverse zones

### LINKED ISSUE #4656: IPv6 CIDR style zones can result in incorrect reverse zones
**What happened**:

In CoreDNS 1.8.4, IPv6 CIDR style zones can expand to an incorrect set of reverse zones.

e.g.

`2001:db8::/115` Results in 2 `/120`'s...

```

/120  /116  /112      /96       /80       /64       /48       /32       /16
0.    0.    0.0.0.0.  0.0.0.0.  0.0.0.0.  0.0.0.0.  0.0.0.0.  8.b.d.0.  1.0.0.2.  ip6.arpa.
1.    0.    0.0.0.0.  0.0.0.0.  0.0.0.0.  0.0.0.0.  0.0.0.0.  8.b.d.0.  1.0.0.2.  ip6.arpa.
```

`2001:db8::/115` Expected result is 2 `/116`'s ...

```
/116  /112      /96       /80       /64       /48       /32       /16
0.    0.0.0.0.  0.0.0.0.  0.0.0.0.  0.0.0.0.  0.0.0.0.  8.b.d.0.  1.0.0.2.  ip6.arpa.
1.    0.0.0.0.  0.0.0.0.  0.0.0.0.  0.0.0.0.  0.0.0.0.  8.b.d.0.  1.0.0.2.  ip6.arpa.```

```

**Environment**:

- the version of CoreDNS:  1.8.4
- Corefile:
- logs, if applicable:
- Others:

chrisohaver:

Fixes handling of IPv6 CIDR reverse zones.

Supercedes #4636
Fixes: #4656


chrisohaver:
We might want to rename `cider.Class()` ... since it's not truly about IP classes. It just so happens that IPv4 reverse zones are segmented by IPv4 class, but this is not the case for IPv6 and since IPv6 "classes" aren't really a thing, it could be confusing.  `SplitOnReverseBoundaries` would be more descriptive of what the function does.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Fix IPv6 case..." ]
>We might want to rename cider.Class() ... since it's not truly about IP
>classes. It just so happens that IPv4 reverse zones are segmented by IPv4
>class, but this is not the case for IPv6 and since IPv6 "classes" aren't really
>a thing, it could be confusing. SplitOnReverseBoundaries would be more
>descriptive of what the function does.

yeah, maybe just 'Split' though? But I agree

chrisohaver:
Converting to draft.  I'm not quite sure about some of the expected results in the tests.

chrisohaver:
> Converting to draft. I'm not quite sure about some of the expected results in the tests.

ok - I counted out the bits, seems correct.

PR #4653, opened 2021-05-28
Add -p for port flag

miekg:
-p is the default flag in DNS software for setting the port, we use
'dns.port' because of preventing clashes with other caddy software
users. This is no longer an issue, so we can do what we want here.

Add -p to works like -dns.port.

(this PR includes generated manpage, but that shouldn't matter too much)


yongtang:
`Go coverage` tests was failing though I think that is just flaky case. Triggered the test to re-run.

yongtang (approved):
LGTM

yongtang:
`Go coverage test` passed now.

Issue #4256, opened 2020-11-04
Empty wildcard matching query returned with NXDOMAIN


SivaKesava1:
**What happened**:

Hi,
I was checking out CoreDNS as an authoritative DNS server and gave it a small test zone as input.

|  	|  	|  	|  	|
|-	|-	|-	|-	|
| example.com. 	| 500 | SOA 	| ns1.outside.com. root.example.com. 3 604800 86400 2419200 604800 	|
| example.com. 	| 500 | NS 	| ns1.outside.com.  	|
| f.*.e.example.com. 	| 500 | A| 2.2.2.2 	|
| t.e.example.com. 	| 500| A 	| 1.1.1.1  	|

For the query _<something.e.example.com. , A>_, the Coredns server returns the following response:
```
          "opcode QUERY",
          "rcode NXDOMAIN",
          "flags QR AA",
          ";QUESTION",
          "something.e.example.com. IN A",
          ";ANSWER",
          ";AUTHORITY",
          "example.com. 500 IN SOA ns1.outside.com. root.example.com. 3 604800 86400 2419200 604800",
          ";ADDITIONAL"
```

**What you expected to happen**:
The query name _something.e.example.com._ will match the empty wildcard node and so the rcode should be `NOERROR` instead of `NXDOMAIN.` The other implementations like BIND, NSD, PowerDNS, Knot return `NOERROR` code with everything else being the same.

**Environment**:

- the version of CoreDNS: `CoreDNS-1.8.0  linux/amd64, go1.15.3, 054c9ae`
- Corefile:
```
example.com:53 {
    file db.example.com
    log
    errors
}
```
- OS: Ubuntu 18

chrisohaver:
The _file_ plugin as written, assumes that wild cards will only contain a `*` in the leftmost segment of the record. IOW, that wildcard segments are never non-terminals (empty or not).  This is the behavior defined in [RFC 1034 4.3.3](https://tools.ietf.org/html/rfc1034#section-4.3.3), which is "Internet Standard".   In that case, I think a non-terminal `*` is  treated as a literal `*`.

[RFC 4592](https://tools.ietf.org/html/rfc4592#section-2.1.3) permits non-terminal and empty non-terminal wildcard segments. It is a "Proposed Standard" (albeit 14 years old).

SivaKesava1:
I did not realize that it was a proposed standard as most of the major implementations allow non-terminal empty wildcards.

miekg:
[...]

PR #4655, opened 2021-05-28
[Re-apply #4599]: plugin/file: Fix in wrong answers returned when wildcard and concrete records exist

xdu31:
It fixes in file plugin wrong answers returned when wildcard and concrete records exist:
```
; example.org test file with wildcard corner cases
$TTL 3600
example.org.		IN	SOA	sns.dns.icann.org. noc.dns.icann.org. 2015082541 7200 3600 1209600 3600  
example.org.		IN	NS	b.iana-servers.net.  
*.d.example.org.        IN      A       127.0.1.53  
r.c.d.example.org.      IN      A       127.0.1.52
```

As per RFC 4592, the query `something.r.c.d.example.com.` , A should not match the wildcard record `*.d.example.com.` , A as `r.c.d.example.com.` , A exists. It is considered the closest encloser, and since there is neither something domain nor a wildcard record under it, the answer has to be NXDOMAIN.
CoreDNS exhibit the same behavior even for queries like `something.c.d.example.com.` , A. The expected response is NXDOMAIN, but CoreDNS returns the wildcard record.

https://github.com/coredns/coredns/issues/4290

No

### DIFF
--- a/plugin/file/lookup.go
+++ b/plugin/file/lookup.go
@@ -56,10 +56,10 @@ func (z *Zone) Lookup(ctx context.Context, state request.Request, qname string) 
 	}
 
 	var (
-		found, shot    bool
-		parts          string
-		i              int
-		elem, wildElem *tree.Elem
+		found, shot              bool
+		parts                    string
+		i, maxLabelNum           int
+		elem, wildElem, nextElem *tree.Elem
 	)
 
 	loop, _ := ctx.Value(dnsserver.LoopKey{}).(int)
@@ -90,6 +90,12 @@ func (z *Zone) Lookup(ctx context.Context, state request.Request, qname string) 
 		// We overshot the name, break and check if we previously found something.
 		if shot {
 			break
+		}
+
+		if nextElem, found = tr.Next(parts); found {
+			if dns.IsSubDomain(parts, nextElem.Name()) {
+				maxLabelNum = z.origLen + i
+			}
 		}
 
 		elem, found = tr.Search(parts)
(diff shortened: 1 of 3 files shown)


chrisohaver:
FYI, @xdu31 copying @miekg's comments from https://github.com/coredns/coredns/pull/4599#issuecomment-850348435

> anyways, finally have some time to comment/take a look. So the issue is/was that the current lookup algo should not have to change in response to this change. Adding a new element and taking Next() is thus probably not the right choice.
>
> Also, I didn't think the tree is sorted and the layout of the elements depends on the insertion of them. I know, 'cause I tried doing this with NSEC and it did not work, although my tests all proved correct, but then I tried with larger zones and the whole thing fell apart.

xdu31:
> FYI, @xdu31 copying @miekg's comments from [#4599 (comment)](https://github.com/coredns/coredns/pull/4599#issuecomment-850348435)
> 
> > anyways, finally have some time to comment/take a look. So the issue is/was that the current lookup algo should not have to change in response to this change. Adding a new element and taking Next() is thus probably not the right choice.
> > Also, I didn't think the tree is sorted and the layout of the elements depends on the insertion of them. I know, 'cause I tried doing this with NSEC and it did not work, although my tests all proved correct, but then I tried with larger zones and the whole thing fell apart.

Hi @miekg, thanks for the review, I will try to see if it can be done in another way later some time. Also would it be possible to get a larger test zone test zone from you?

miekg:
have re-read the rfc. This is def. an issue. I'll take a look if the search algo needs big changes are not.
My understanding of wildcard domains didn't change only domains with left most label is "*" are wildcards, but the above is a bug

chrisohaver:
> My understanding of wildcard domains didn't change only domains with left most label is "*" are wildcards

Agreed.  Empty non-terminal wildcards (wildcard symbol not in left most label), are _proposed_ in https://tools.ietf.org/html/rfc4592#section-2.1.3.  Although it's clear that they propose lifting the restriction, the RFC is ambiguous on the matter of how such records would behave.  It also seems to suggest that the original restriction permitted them ... ` but the [section 4.3.3 of RFC 1034] restriction as stated still permits "example.*.example." for instance.` ... which to me is clearly not true.

Anyways, this PR and corresponding issue #4290 are not related to empty non-terminal wildcards.

Another open issue not related to this PR does touch on empty non-terminal wildcards: #4256.

chrisohaver (approved):

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] [Re-apply #45..." ]
***@***.*** approved this pull request.

still not sure this is the correct fix. Checking it is (somewhat) annoying, as I need to
implement the fix to see how it compares

chrisohaver:
> still not sure this is the correct fix. Checking it is (somewhat) annoying, as I need to implement the fix to see how it compares

Understood, I'm just carrying forward my approval from when the original was open.
I had previously tested locally using the GitHub cli tool... `gh pr checkout 4655`, which is how I confirmed some of the problems with the intermediate iterations of the PR (that were not yet covered by tests).

Issue #4668, opened 2021-06-06
plugin/route53: Cannot retrieve implicit AWS credentials on ECS Fargate


Licht-T:
**What happened**:
We can use Route53 plugin with implicit AWS credentials by attaching the "AmazonRoute53ReadOnlyAccess" policy to the EC2 instance. On the other hand, we cannot do so on ECS Fargate even if the appropriate policy is attached to its task execution role.

**What you expected to happen**:
Route53 plugin works with implicit AWS credentials on ECS Fargate.

**How to reproduce it (as minimally and precisely as possible)**:
- Attach the AmazonRoute53ReadOnlyAccess policy to the ECS task execution role.
- Run Route53 plugin enabled CoreDNS on ECS Fargate.

**Anything else we need to know?**:
No.

**Environment**:

- the version of CoreDNS: 1.8.4
- Corefile:
```
licht.local {
        route53 licht.local.:Z00037533HUIT2DY2YV46
}

. {
        forward . 8.8.8.8
}
```
- logs, if applicable:
```
2021-06-06 16:14:08 plugin/route53: Corefile:2 - Error during parsing: failed to create route53 plugin: NoCredentialProviders: no valid providers in chain. Deprecated.
2021-06-06 16:14:08 For verbose messaging see aws.Config.CredentialsChainVerboseErrors
```
- OS (e.g: `cat /etc/os-release`):
- Others:

Licht-T:
Will close since PR merged.

PR #4669, opened 2021-06-06
plugin/route53: Fix bug when retrieving implicit AWS credentials on ECS Fargate

Licht-T:
This enables Route53 plugin to retrieve implicit AWS credentials on ECS Fargate. Tested on both EC2 instance and ECS Fargate.

https://github.com/coredns/coredns/issues/4668

We don't need to change any documentation.

No, this does not.


yongtang (approved):
LGTM

PR #4680, opened 2021-06-08
plugin/log: fix closing of codeblock

miekg:
This fixes the rendering of this README as html. The codeblock was
closed with 4 ~, should be 3.

Signed-off-by: Miek Gieben <miek@miek.nl>

PR #4529, opened 2021-03-18
workflow: move committing workflows to schedule

miekg:
This moves all workflows that commit to a schedule, so that it _doesn't_
push into peoples PRs as that enlarges them and is just non-obvious.

They run throughout the week.

Also change the "user" we use for this, so that DCO believes it a real
email address (DCO is pretty stupid here). Update the authors target to
filter out this bot as well.

Signed-off-by: Miek Gieben <miek@miek.nl>

PR #4683, opened 2021-06-08
Fix timestamp issue for doc generating

yongtang:

This PR will fix the issue of doc is not automated rendered
through GitHub Actions anymore.

It looks the issue is cause by the fact that in `fixup_file_mtime.sh`
is only fixing files on source (.md) side, not on target (man/*.[1|5|7])
side. As a result Makefile will skip the rendering of doc as
it assume everything will be update to date.

This should fix the issue we were facing.

See https://github.com/coredns/coredns/pull/4680#issuecomment-856872811 for related discussion.

n/a

n/a

### DIFF
--- a/.github/fixup_file_mtime.sh
+++ b/.github/fixup_file_mtime.sh
@@ -9,7 +9,7 @@ if [[ ! -f 'coredns.1.md' ]]; then
   exit 1
 fi
 
-for file in coredns.1.md corefile.5.md plugin/*/README.md; do
+for file in coredns.1.md corefile.5.md plugin/*/README.md man/*.1 man/*.5 man/*.7; do
   time=$(git log --pretty=format:%cd -n 1 --date='format:%Y%m%d%H%M.%S' "${file}")
   touch -m -t "${time}" "${file}"
 done


yongtang:
/cc @miekg I think this will fix the issue. On local machine after this change, I am able to see following changes which indicates many docs are due for udpates.
```
git status
On branch doc-fix
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   man/coredns-bind.7
	modified:   man/coredns-cache.7
	modified:   man/coredns-dnssec.7
	modified:   man/coredns-file.7
	modified:   man/coredns-forward.7
	modified:   man/coredns-health.7
	modified:   man/coredns-kubernetes.7
	modified:   man/coredns-log.7
	modified:   man/coredns-loop.7
	modified:   man/coredns-rewrite.7
	modified:   man/coredns-secondary.7
	modified:   man/coredns-sign.7

no changes added to commit (use "git add" and/or "git commit -a")
```

miekg:
Nice. Thanks!

/lgtm

yongtang:
Merge the PR now. The GitHub Actions for doc building is based on cron schedule (PR #4529). I think by tomorrow we will see the doc changes applied automatically.

Issue #4684, opened 2021-06-09
TC Flag not set for Externalname kind service when the payload is large


TheRealGoku:
As explained in this article
https://easoncao.com/coredns-resolution-truncation-issue-on-kubernetes-kube-dns/

Ideal behavior - the response should contain not just cname but also all ip's that belongs to the cname. If the server(coredns) is not able to sandwich all the ip's due to payload limit of 512, it should send partial payload and with Truncate flag set, once this partial truncate response was received by the client, the client will use tcp as a mode to send same dns query again,

Since coredns failed to set the truncate flag and since the "A" authoritative flag was set, clients neither upgrading the request to tcp nor utilizing recursive strategy to send follow up "A" query to cname. The reason behind setting "A" could be acceptable due to the fact that coredns has dual responsibility (as a Authoritative domain for local kubernetes related services, as a recursive resolver for external internet domains)

The only concern here is a coredns just responding with cname without even setting truncate flag.

**Tested Potential workarounds:**

1. Forcefully make the clients(pods) to use just tcp mode when interacting with coredns pods. This could be simply configured with the help of use-vc flag like below
```
  template:
    metadata:
      labels:
        app: customer-dns
    spec:
      dnsConfig:
        options:
          - name: use-vc
      containers:
        - name: customer-dns
          image: nginx:latest
          ports:
            - containerPort: 8080
          resources:
            limits:
              cpu: 300m
              memory: 500Mi
            requests:
              cpu: 200m
              memory: 500Mi
```
 
2. As a win win situation, the customer could use nodelocaldns cache, as the communication from nodelocaldns cache to the coredns pod will be by default tcp, and the communication between pod to the nodelocaldns pod can either be tcp or udp.

As per my testing, implementing nodelocaldns cache solved the issue as now the issue is being masked since nodelocalcache dns implementation didnt act in that unexpected way.

chrisohaver:
I'm pretty certain this was fixed by https://github.com/miekg/dns/pull/1171, which is included in CoreDNS in the 1.8.0/1.8.1 timeframe.  The article referenced used CoreDNS 1.6.6, which did not have that fix.

TheRealGoku:
Thanks @chrisohaver , I have recently tested this behavior in EKS 1.18 running CoreDNS-1.7.0. Will test this now on EKS 1.20 which should have CoreDNS 1.8.3. do you recall which exact version this got merged in?

chrisohaver:
[...]

PR #4687, opened 2021-06-11
WIP: Reflect truncated state to client, if upstream lookup is truncated

chrisohaver:

When responding to a client with records that were obtained from a truncated upstream response, the response to the client should be marked truncated. We can fail to do this for a CNAME to external records, where the response from the upstream providing the records is truncated. If that happens, the answer to the client need to be marked truncated.

Note: Depending on individual record size and size of the truncated upstream response, this issue may or may not manifest. If the total length of CNAME + truncated A records happens to exceed the client's max buffer size, then the response to the client will be truncated at write time.  But if the total length of CNAME + truncated A records does _not_ exceed the clients buffer size, then the answer will be written with the TC bit off, even though the answer is not complete.  Responses are truncated on record boundaries - so this is dependent on the size of the individual records.

This is a big change, we may want to break it into pieces. I've tried to break the commits out per dependent groups.

_Possibly_ related to #4684, although the symptoms described are different.

### DIFF
--- a/plugin/auto/auto.go
+++ b/plugin/auto/auto.go
@@ -67,10 +67,11 @@ func (a Auto) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (i
 		return dns.RcodeRefused, nil
 	}
 
-	answer, ns, extra, result := z.Lookup(ctx, state, qname)
+	answer, ns, extra, result, tc := z.Lookup(ctx, state, qname)
 
 	m := new(dns.Msg)
 	m.SetReply(r)
+	m.Truncated = tc
 	m.Authoritative = true
 	m.Answer, m.Ns, m.Extra = answer, ns, extra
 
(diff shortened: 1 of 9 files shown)


chrisohaver:
@miekg, would you prefer I split this into separate PRs?

The `template` and `dns64` cases are trivial.
The `backend.go` fixes (`etcd` and `kubernetes`) are not too bad IMO.

I can understand your hesitation to fix this for `file` et al, since the structure there passes record sets around, not whole responses, it alters the signature of many methods.

But this is a real issue, and IMO we should at least fix it in the places where it's not too disruptive.

chrisohaver:
I've split the template and dns64 fixes out to separate PRs, since they are simple non-disruptive fixes.

chrisohaver:
> I've split the template and dns64 fixes out to separate PRs, since they are simple non-disruptive fixes.

... and  etcd/kubernetes as well.

PR #4688, opened 2021-06-13
Create geoip plugin

snebel29:
This plugin add support for configuring maxmind geoIP2 City database to add geo location data to request metadata.

Discussed in https://github.com/coredns/coredns/discussions/4685.

I still have to check everywhere a documentation change is required.

As a new plugin there is no backward incompatible changes.


plugin/mmdb/README.md
```diff
@@ -0,0 +1,56 @@
+# mmdb
```
miekg: wouldn't geo or geoip be more descriptive?

plugin/mmdb/README.md
```diff
@@ -0,0 +1,56 @@
...
+## Name
+*mmdb* - Lookup maxmind geoip2 databases then add associated data to the context request.
+
+## Description
+*mmdb* - Allows you to configure [geoIP2 maxmind databases](https://dev.maxmind.com/geoip/docs/databases) and automatically associate and DNS request data matching the client IP.
```
miekg:
```suggestion
The *mmdb* plugin allows configuring a  [geoIP2 maxmind databases](https://dev.maxmind.com/geoip/docs/databases) to automatically associate a geographic region by matching the client IP.
```

plugin/mmdb/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+Common use cases would be adding per request GEO location data, ISP, ASN, etc.
+
+The data will be added to the request context leveraging the [metadata](../metadata/README.md) plugin, and therefore data will be available by make use of it, label names will be created using the following convention.
```
miekg: this sentence doesn't work

plugin/mmdb/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Databases
+The following official [max mind databases](https://dev.maxmind.com/geoip/docs/databases) are allowed.
```
miekg:
```suggestion
The following  [max mind databases](https://dev.maxmind.com/geoip/docs/databases) are allowed.
```

plugin/mmdb/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+- Anonymous IP
+- ASN
+- City
+- ConnectionType
+- Country
+- Domain
+- Enterprise
+- ISP
```
miekg: unclear what these actually mean...?

plugin/mmdb/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+Databases with custom schemas are not allowed, but may be supported in the future.
```
miekg:
```suggestion
Databases with custom schemas are not supported yes.
```

plugin/mmdb/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Metadatada Label/Value Convention
```
miekg:
```suggestion
## Metadatada Labels
```

plugin/mmdb/README.md
```diff
@@ -0,0 +1,56 @@
...
+**Label**: plugin/mmdb/`KEYNAME`/...
+
+**Value**: Will be the associated field value in its string representation.
```
miekg:
```suggestion
```

plugin/mmdb/README.md
```diff
@@ -0,0 +1,56 @@
...
+- ISP
+
+Databases with custom schemas are not allowed, but may be supported in the future.
+
+## Metadatada Label/Value Convention
+**Label**: plugin/mmdb/`KEYNAME`/...
+
+**Value**: Will be the associated field value in its string representation. 
+
+> Note: Numberic types will be represented in base 10.
```
miekg:
```suggestion
Note that  numberic types will be represented in base 10.
```

plugin/mmdb/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+For example, in a `City` database but not limited to those fields:
+
+| Label                                                 | Value            |
+| :---------------------------------------------------- | :-------------   |
+| `plugin/mmdb/city/names/en`                           | `Cambridge`      |
```
miekg: think the label is just `mmdb..../`

plugin/mmdb/README.md
```diff
@@ -0,0 +1,56 @@
...
+| `plugin/mmdb/country/names/en`                        | `United Kingdom` |
+| `plugin/mmdb/continent/names/en`                      | `Europe`         |
+| `plugin/mmdb/registered_country/is_in_european_union` | `false`          |
+| `plugin/mmdb/location/longitude`                      | `0.1315`         |
```
miekg: can this be sortened? Ie. mmdb/longitude? Or do you get collisions then? What main categories are there? What sub categories are there, i.e. what is the meaning of these label after you've explained the syntax

plugin/mmdb/README.md
```diff
@@ -0,0 +1,56 @@
...
+| `plugin/mmdb/location/latitude`                       | `52.2242`        |
+
+## Syntax
+```txt
+mmdb [DBFILE]
```
miekg:
```suggestion
mmdb DBFILE [DBFILE]...
```

plugin/mmdb/fixtures/README.md
```diff
@@ -0,0 +1,7 @@
+# mmdb fixtures
```
miekg: should be in testdata

plugin/mmdb/fixtures/create_databases.go
```diff
@@ -0,0 +1,147 @@
...
+)
+
+const (
+	cdir          = "81.2.69.142/32"
+	cityDBName    = "GeoLite2-City.mmdb"
+	countryDBName = "GeoLite2-Country.mmdb"
+)
+
+// Create new mmdb database fixtures in this directory.
+func main() {
```
miekg: why is this needed when you check in the files anyway? And if needed it should probably be moved to a _test.go file

plugin/mmdb/mmdb.go
```diff
@@ -0,0 +1,100 @@
...
+	return plugin.NextOrFailure(pluginName, m.Next, ctx, w, r)
+}
+
+const labelPrefix = "plugin/" + pluginName + "/"
+
+// Metadata implements the metadata.Provider Interface in the metadata plugin, and is used to store
+// the data associated with the source IP of every request.
+func (m Mmdb) Metadata(ctx context.Context, state request.Request) context.Context {
+	srcIP := net.ParseIP(state.IP())
+	if srcIP == nil {
```
miekg: can't happen

plugin/mmdb/mmdb.go
```diff
@@ -0,0 +1,100 @@
...
+
+	for _, db := range m.databases {
+		// Enterprise schema is a superset of City schema which in turn is a superset of Country schema
+		// therefore we just query the schema that has the most data available.
+		switch {
+		case db.provides&enterprise == 1:
+			data, err := db.Enterprise(srcIP)
+			if err != nil {
+				// TODO(snebel29): Print a debug error without causing a logging flood?
+				break
```
miekg: probably need a way to set an error in a label?

plugin/mmdb/mmdb.go
```diff
@@ -0,0 +1,100 @@
...
+	srcIP := net.ParseIP(state.IP())
+	if srcIP == nil {
+		// This should be unreachable code.
+		// TODO(snebel29): Print a debug error without causing a logging flood?
+		// TODO(snebel29): Is there any Prometheus metric that we should provide?
+		return ctx
+	}
+
+	for _, db := range m.databases {
+		// Enterprise schema is a superset of City schema which in turn is a superset of Country schema
```
miekg: this ordering must be described in the README

plugin/mmdb/setup.go
```diff
@@ -0,0 +1,94 @@
...
+	dnsserver.GetConfig(c).AddPlugin(func(next plugin.Handler) plugin.Handler {
+		mmdb.Next = next
+		return mmdb
+	})
+
+	return nil
+}
+
+// probingIP is an arbitrary internet IP used to discover database types through database lookups.
+var probingIP = net.ParseIP("81.2.69.142")
```
miekg: use something from rfc 1918 if possible, otherwise the IPv4 documenation IP address

plugin/mmdb/setup.go
```diff
@@ -0,0 +1,94 @@
...
+
+func mmdbParse(c *caddy.Controller) (*Mmdb, error) {
+	dbPaths := make(map[string]bool)
```
miekg:
```suggestion
	dbPaths := make(map[string]struct{})
```

plugin/mmdb/setup.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	// TODO(snebel29): Warn or Fail when configuring overlapping db/fields?
+	// TODO(snebel29): Fail if we open more than once the same dbType?
+	dbs := make([]mmdb, 0, len(dbPaths))
+
+	for path := range dbPaths {
+		reader, err := geoip2.Open(path)
+		if err != nil {
+			return nil, plugin.Error(pluginName, c.Errf("failed to open database file: %v", err))
```
miekg: plugin.Error is already called when this function returns

plugin/mmdb/setup.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+		// Probe all databases to figure out the database information that it can provide.
+		for _, schema := range schemas {
+			if err := schema.validate(); err != nil {
+				// If we get an InvalidMethodError then we know this database does not provide enterprise schema.
+				if _, ok := err.(geoip2.InvalidMethodError); !ok {
+					return nil, plugin.Error(pluginName,
+						c.Errf("unexpected failure looking up database %q schema %q: %v", filepath.Base(path), schema.name, err))
+				}
+			} else {
```
miekg: danling else, can just to continue in the above if and remove the else entirely

plugin/mmdb/README.md
snebel29: I recognize I was on the fence choosing a name.

Technically speaking the plugin is providing support for `mmdb` format databases, some of which don't provide geoip data, see: Domain, ASN, connection type, ISP, anonymous IP, etc.

In the other hand `mmdb` databases seems to be better known and became popular providing geoip data, all of those databases (including the ones providing non geo data) are often referred as geoIP.

I really have no strong opinion on one or the other, at some point I even used geoIP2 as plugin name...then renamed to `mmdb` to be technically correct.

What do you think we should be using?

plugin/mmdb/README.md
snebel29: Many of those databases provides useful data which is not geographic, see: https://dev.maxmind.com/geoip/docs/databases I feel with little effort (by handling their specific schemas) this plugin could support them all, allowing users to hook any of those databases instead of just doing geo with City, Country and Enterprise databases.

What do you think?

plugin/mmdb/README.md
snebel29: Those are the names of the popular databases that maxmind provides and  which can be used to provide data associated with IP addresses, and which could be configured in this plugin, I didn't want to copy/paste the description on each of them and rather provided a link to their website, where they are detailed.

Do you feel we should be doing differently?

plugin/mmdb/README.md
snebel29: I guess you meant `s/yes/yet/` - will do.

plugin/mmdb/README.md
snebel29: Some fields could be shortened, but others would collide, see example for the City database https://github.com/oschwald/geoip2-golang/blob/main/reader.go#L85-L133 where geoNameId and names among others are used in different nested structs.

My thinking was that is better to provide them all, and let the user decide. Optionally fields could be filtered in/out by users through config.

The full list of fields for all those databases can be found in https://github.com/oschwald/geoip2-golang/blob/main/reader.go#L19-L202

Does it make sense?

plugin/mmdb/README.md
snebel29: I was thinking that path are potentially long, so instead of providing them all into a single `mmdb` statement, users could provide one path into one `mmdb` statement, but I'm happy to do as you suggest if that's the preferred way.

plugin/mmdb/fixtures/README.md
snebel29: ok, will match the hierarchy.

plugin/mmdb/fixtures/create_databases.go
snebel29: Creating the database and record was a bit painful so I felt keeping the boilerplate to recreate them in case further record fixtures (or databases) were needed, would be handy.

I can move them over a `_test.go` file, or even remove if you push for it.

plugin/mmdb/mmdb.go
snebel29: Ok, I was a bit scared of getting a nil pointer dereference and decided to do defensive code, but I'm happy to ignore.

plugin/mmdb/mmdb.go
snebel29: oh, interesting. I was naively thinking that consumers would simply look for the keys they are interested and simply handle scenarios when is available and not.

Do you think providing the error in a label would be handy for consumers? do you have an specific label convention in mind?

plugin/mmdb/mmdb.go
snebel29: Will add a TODO and get it done, in any case this should be transparent for users... this is just avoiding redundant lookups, to end up overriding existing labels.

plugin/mmdb/setup.go
snebel29: I guess you mean to use a private IP instead? of course that's no problem at all.

> otherwise the IPv4 documenation IP address

I didn't get that, sorry.

plugin/mmdb/setup.go
snebel29: ok, will double check and remove redundant error.

plugin/mmdb/setup.go
snebel29: Will do.

snebel29:
Thank you @miekg for spending you precious time providing early feedback, I feel the key point where we need to get aligned is, whether the plugin should generically support all publicly available with known schemas mmdb databases, or focus on the ones providing location data only.

As commented in a previous thread I feel with little effort we could support them all in the same plugin... but we could as well provide different plugins, for every database case with a predefined list of fields, starting with geoip.

I'm just afraid those would becomes a bit of a copy paste of the geoip plugin with a few changes in the accessed fields. But this also works, and is probably easier to understand for users.

Other open discussions such as shortening of labels may be affected because of this one.

plugin/mmdb/README.md
snebel29: ok, will give it a second look after all proposed changes have been merged.

plugin/mmdb/README.md
snebel29: Done

plugin/mmdb/README.md
snebel29: Narrowed down to only one type.

plugin/mmdb/README.md
snebel29: Changed to geoip

plugin/mmdb/README.md
snebel29: Simplified using a limited set of fields.

plugin/mmdb/README.md
snebel29: Plugin config block changed to
```
geoip [DBFILE] {
    languages [CODE]...
}
```

plugin.cfg
```diff
@@ -20,6 +20,7 @@
 # log:log
 
 metadata:metadata
+geoip:geoip
```
snebel29, line 23: I didn't know what the right position would be, I had the impression for this plugin (geoip) it doesn't matter much since its serveDNS function just run Next in the chain?

snebel29:
Hi @miekg 

I implemented my understanding of the suggested changes, notably reduced the scope to focus on `geoip` rather than trying to fit a more general `mmdb` plugin.

Also I provided a default and limited set of labels that I felt would be of use for most users, I guess another (more complex) option could be to allow users to configure the fields they want...

There is still a few conversations unresolved from your previous comments, take a look and let me know when you have chance.

This PR is still WIP mostly because I have to review documentation changes, in case its current form is accepted, but other than that I think it should be ready to review.

Thanks and regards.

snebel29:
In case it's useful to anyone, I did manually test the plugin by creating a local dummy interface.
```shell
$ sudo ip link add dummy1 type dummy; sudo ip addr add 81.2.69.142/32 dev dummy1; sudo ip link set dummy1 up
```
Creating `example.org` zone file.
```txt
$ORIGIN example.org.
@	3600 IN	SOA sns.dns.icann.org. noc.dns.icann.org. 2017042745 7200 3600 1209600 3600

www     IN A     127.0.0.1
        IN AAAA  ::1
```
Creating `Corefile`
```txt
example.org {
        file example.org
        debug
        log
        geoip YOUR_WORKSPACE/coredns/plugin/geoip/testdata/GeoLite2-City.mmdb
        metadata
}
```
Run `coredns` with the new config.
```
$ ./coredns -conf Corefile -dns.port 9053
```
Lookup coredns on the dummy interface.
```
$ dig +short www.example.org @81.2.69.142 -p 9053
```

The I played by adding code to some existing plugins (to avoid creating one from scratch) using the example code in the README.md.
```go
import (
    "strconv"
    "github.com/coredns/coredns/plugin/metadata"
)
// ...
if getLongitude := metadata.ValueFunc(ctx, "geoip/longitude"); getLongitude != nil {
    if longitude, err := strconv.ParseFloat(getLongitude(), 64); err == nil {
		// Do something useful with longitude.
	}
} else {
    // The metadata label geoip/longitude for some reason, was not set.
}
// ...
```

Once finished clean up the dummy interface
```shell
$ sudo ip link del dummy1
```

plugin.cfg
miekg: this looks fine

plugin/geoip/README.md
```diff
@@ -0,0 +1,100 @@
+# geoip
+
+## Name
+*geoip* - Lookup maxmind geoip2 databases using the client IP, then add associated geoip data to the context request.
+
+## Description
+The *geoip* plugin add geo location data associated with the client IP, it allows you to configure a [geoIP2 maxmind database](https://dev.maxmind.com/geoip/docs/databases) to add the geo location data associated with the IP address.
+
+The data is added leveraging the [metadata](../metadata/README.md) plugin, values can then be retrieved using it as well, for example:
```
miekg:
```suggestion
The data is added leveraging the *metadata* plugin, values can then be retrieved using it as well, for example:
```

plugin/geoip/README.md
```diff
@@ -0,0 +1,100 @@
...
+		// Do something useful with longitude.
+	}
+} else {
+    // The metadata label geoip/longitude for some reason, was not set.
+}
+// ...
+```
+
+## Databases
+Right now only databases providing the city schema such as `City` and `Enterprise` are supported, other databases types with different schemas are not supported yet.
```
miekg:
```suggestion
The supported databases use city schema such as `City` and `Enterprise` . Other databases types with different schemas are not supported yet. 
```

plugin/geoip/README.md
```diff
@@ -0,0 +1,100 @@
...
+} else {
+    // The metadata label geoip/longitude for some reason, was not set.
+}
+// ...
+```
+
+## Databases
+Right now only databases providing the city schema such as `City` and `Enterprise` are supported, other databases types with different schemas are not supported yet. 
+
+You can download a free and public City database [here](https://www.maxmind.com/en/geoip2-city).
```
miekg:
```suggestion
You can download a [free and public City database](https://www.maxmind.com/en/geoip2-city).
```

plugin/geoip/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Syntax
+* **DBFILE** the mmdb database file path.
```
miekg: move the DBFILE bullet below the syntax example

plugin/geoip/README.md
```diff
@@ -0,0 +1,100 @@
...
+```txt
+geoip [DBFILE]
+```
+
+You can specifiy a list of languages for `names` fields.
```
miekg: the names field is a thing in the geo IP db? Prolly worth explaining a tiny bit more, or remove this entirely and explain it below in the bullet item

plugin/geoip/README.md
```diff
@@ -0,0 +1,100 @@
...
+        languages en fr es zh-CN
+    }
+    metadata # Note that metadata plugin must be enabled as well.
+}
+```
+
+## Metadatada Labels
+A limited set of fields will be exported as labels, all values are stored using strings **regardless of their underlying value type**, and therefore you may have to convert it back to its original type, note that numeric values are always represented in base 10.
+
+**LANG** Language location ISO??? code, if database does not have a name for the selected language the label
```
miekg:
```suggestion
**LANG** is the language  ISO code, if database does not have a name for the selected language the label
```

plugin/geoip/README.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+* `languages` configures a list of languages for which you want to create a label, by default it's set to English: `en`, note that except for English many languages may not be available in the database, language must be available in your geoip2 database, a list of the codes supported can be found in [Locale Codes](#LocaleCodes).
+
+## Examples
+The following configuration configures the `City` database adding names in English, French, Spanish and Simplified Chinese.
+```txt
+. {
+    geoip /opt/geoip2/db/GeoLite2-City.mmdb {
+        languages en fr es zh-CN
```
miekg: looking at the use of LANG in the labels and I don't feel it's good enough. You're basically forcing the user of this data to looks for all possible LANGs and there is a coupling the corefile and the code.
Why is a language even needed? I think we can default to english and drop the whole LANG option

plugin/geoip/geoip_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/coredns/coredns/plugin/test"
+	"github.com/coredns/coredns/request"
+	"github.com/coredns/coredns/plugin/metadata"
+)
+
+func TestGeoIPName(t *testing.T) {
```
miekg: This doesn't really test anything useful, please remove

plugin/geoip/geoip.go
```diff
@@ -0,0 +1,108 @@
...
+	city
+	connectionType
+	country
+	domain
+	enterprise
+	isp
+)
+
+// probingIP is an arbitrary internet IP used to discover database types through database lookups.
+var probingIP = net.ParseIP("81.2.69.142")
```
miekg: this is a weird thing to have, please remove

miekg:
this is getting close, code starts to look good. Left a bunch of new comment and a discussion on the use of languages

plugin/geoip/README.md
snebel29: Done

plugin/geoip/README.md
snebel29: Deferring action until discussion around languages is clear, because we may end up removing languages option which would get this text removed entirely.

plugin/geoip/README.md
snebel29: My initial thought was that some users may have a use case for getting "names", such as city names, country names, etc. in a different language than English, and giving them a a good default (English) while still letting them change that choice could be useful.

I'm happy to remove the option and provide English only, and do something else in the future when a clear use case appears.

plugin/geoip/geoip_test.go
snebel29: Done

plugin/geoip/geoip.go
snebel29: The plugin validates that a `mmdb` database file provides the right schema by querying up the database then looking into the error returned to see if the schema is valid.

To do that the plugin is leveraging reader.City method whose signature requires a net.IP, we have to pass any valid net.IP value.

Would you be happy by changing the variable name and explaining a bit better why do we have that variable? do you propose to do the schema validation in a different way?

plugin/geoip/README.md
snebel29: Removed.

plugin/geoip/README.md
snebel29: Removed languages option, and default to English only.

snebel29:
> this is getting close, code starts to look good. Left a bunch of new comment and a discussion on the use of languages

Once again thank you, I pushed the suggested changes and leaved a reply into one of them.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Create geoip ..." ]
>such s cities, countries, etc. in a different language than English, and giving
>them a a good default (English) while still lettting them change that choice
>could be useful.
>
>I'm happy to remove the option and provide English only, and do something else
>in the future when a clear use case appears.

yes, please. I think the use-case if to know *where* the query comes from, not the precise
locale of the city name.

Also does this work with IPv6? Or only ipv4?

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Create geoip ..." ]
>
>
>In plugin/geoip/geoip.go:
>
>> +     provides int
>+}
>+
>+const (
>+       anonymousIP = 1 << iota
>+       asn
>+       city
>+       connectionType
>+       country
>+       domain
>+       enterprise
>+       isp
>+)
>+
>+// probingIP is an arbitrary internet IP used to discover database types through database lookups.
>+var probingIP = net.ParseIP("81.2.69.142")
>
>The plugin validates that a mmdb database file provides the right schema by
>querying up the database then looking into the error returned to see if the
>schema is valid.

then 127.0.0.1 seems as good as any. Also don't think this needs to be a global variable.

snebel29:
> [ Quoting ***@***.***> in "Re: [coredns/coredns] Create geoip ..." ]
> such s cities, countries, etc. in a different language than English, and giving them a a good default (English) while still lettting them change that choice could be useful. I'm happy to remove the option and provide English only, and do something else in the future when a clear use case appears.
> yes, please. I think the use-case if to know *where* the query comes from, not the precise locale of the city name. Also does this work with IPv6? Or only ipv4?

Languages option was removed in https://github.com/coredns/coredns/pull/4688/commits/806e70de76e15e2a3478f717f4dd3468f8ebcc51 and https://github.com/coredns/coredns/pull/4688/commits/09d381d751fc08d96d691b46f684dbe03d5076d2

Maxmind databases [support IPV6](https://support.maxmind.com/ipv6-data-faq/) out of the box, as well as the package used to interact with it.

snebel29:
> then 127.0.0.1 seems as good as any. 

For some reason the official mmdb writter does not let you to add reserved "networks".
```
2021/07/11 12:13:27 attempt to insert ::7f00:1/128, which is in a reserved network
```
So I've originally decided to use a non reserved one, even with that limitation and since we don't use database data during validation, I could use `127.0.0.1`, that's ok. <= Done.

> Also don't think this needs to be a global variable.

probingIP is also set to `nil` in one testing scenario, to simulate an unexpected validation error, having that in a global and un-exported variable makes it easy, is that ok for you?

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Create geoip ..." ]
>    then 127.0.0.1 seems as good as any.
>
>For some reason the official mmdb writter does not let you to add reserved
>"networks".
>
>2021/07/11 12:13:27 attempt to insert ::7f00:1/128, which is in a reserved network
>
>So I've decided to use a non reserved one, even with that limitation and since
>we don't use database data during validation, I could use 127.0.0.1, that's ok.
><= Done.
>
>    Also don't think this needs to be a global variable.
>
>probingIP is also set to nil in one testing scenario, to simulate an unexpected
>validation error, having that in a global and un-exported variable makes it
>easy, is that ok for you?

probably still doesn't need to be a global var.

If you need an example IP use one from the documentation range: 192.0.2.0/24
https://en.wikipedia.org/wiki/Reserved_IP_addresses

plugin/geoip/city.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/coredns/coredns/plugin/metadata"
+	"github.com/oschwald/geoip2-golang"
+)
+
+const defaultLang = "en"
+
+func (g GeoIP) setCityMetadata(ctx context.Context, data *geoip2.City) {
+	// Set labels for city, country and continent names.
+	if name, ok := data.City.Names[defaultLang]; ok {
```
miekg: should we set an error in the label, when the language is not available for this city? 'no city found for 'en'' or something?

plugin/geoip/geoip.go
```diff
@@ -0,0 +1,102 @@
...
+	*geoip2.Reader
+	// provides defines the schemas that can be obtained by querying this database, by using
+	// bitwise operations.
+	provides int
+}
+
+const (
+	anonymousIP = 1 << iota
+	asn
+	city
```
miekg: only city is used from this list of constant
(at least `domain` is not used - which I what I checked)

plugin/geoip/geoip.go
```diff
@@ -0,0 +1,102 @@
...
+// Metadata implements the metadata.Provider Interface in the metadata plugin, and is used to store
+// the data associated with the source IP of every request.
+func (g GeoIP) Metadata(ctx context.Context, state request.Request) context.Context {
+	srcIP := net.ParseIP(state.IP())
+
+	switch {
+	case g.db.provides&city == city:
+		data, err := g.db.City(srcIP)
+		if err != nil {
+			log.Debugf("setting up metadata failed due to database lookup error: %v", err)
```
miekg:
```suggestion
			log.Debugf("Setting up metadata failed due to database lookup error: %v", err)
```

plugin/geoip/geoip.go
```diff
@@ -0,0 +1,102 @@
...
+			break
```
miekg:
```suggestion
			return ctx
```

plugin/geoip/setup_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"strings"
+	"testing"
+
+	"github.com/coredns/caddy"
+	"github.com/coredns/coredns/core/dnsserver"
+)
+
+var (
+	_, file, _, _    = runtime.Caller(0)
+	fixturesDir      = filepath.Join(filepath.Dir(file), "testdata")
```
miekg: you can just open files in `testdata` relative from the tests, no need for `runtime.Caller(0)` stuff

plugin/geoip/setup_test.go
```diff
@@ -0,0 +1,114 @@
...
+	enterpriseDBPath = filepath.Join(fixturesDir, "GeoIP2-Enterprise.mmdb")
+	cityDBPath       = filepath.Join(fixturesDir, "GeoLite2-City.mmdb")
+	countryDBPath    = filepath.Join(fixturesDir, "GeoLite2-Country.mmdb")
```
miekg: I think we're only using `city`? So we don't need to test anything else here.

plugin/geoip/testdata/create_databases.go
```diff
@@ -0,0 +1,117 @@
+// This package is a helper function to create the mmdb databases used for testing in the
+// geoip plugin.
+package main
```
miekg: I don't think we need this binary, just check in a test mmdb, maybe put in some pointers on how to create one yourself.

snebel29:
> probably still doesn't need to be a global var. 

It doesn't need to be a global var, it just felt using a global var was the simplest way to achieve what I needed which is temporarily set the `net.IP` used for validation to `nil` to force getting a different error during validation, to test an error edge case.

I can try to do in a different way, using some struct and member field, or something like that. Is that what you want?

> If you need an example IP use one from the documentation range: 192.0.2.0/24 https://en.wikipedia.org/wiki/Reserved_IP_addresses

Hmmm, what do you mean? I'm already using `127.0.0.1` as you suggested for database schema validation, for the mmdb files used for testing I can't create records using private IP scopes, it fails to create records using private networks (as shown before) so I had to use a public IP scope. Am I missing something?

plugin/geoip/city.go
snebel29: My two cents,
To the best of my knowledge `en` language is always set in all name fields (city, country, continent) in the official `mmdb` databases, if it's not being set it would be a "bug" during database creation, of course I think this scenario needs to be handle.

The advantage I feel provides not setting the label (as it's done right now), is that is straightforward for consumers to know when the value of the label is not available and therefore not do anything with it, they simply load the ` metadata.ValueFunc` and if it's nil then it does not exist, the downside is that you don't know what's the reason for it to not exist.

In the contrary when we set the label to an error, as you suggested is not that straightforward to know for a consumer if the value is the real value name, or just an error. May cause consumers to rely and hard code the error text in their code, etc. The positive is that the error description is provided instead of simply not setting the label... although I'm not completely sure how a user would leverage it.

WDYT?

plugin/geoip/geoip.go
snebel29: I felt it make sense to keep the entire list for forward compatibility in case we decide to use other database types in the future, and to match the existing list in the [geopip2-golang library](https://github.com/oschwald/geoip2-golang/blob/main/reader.go#L206-L215).

Having our own list with different values for different schemas would make things harder and may get out of sync with the original database type list of values.

WDYT?

plugin/geoip/setup_test.go
snebel29: Done

plugin/geoip/setup_test.go
snebel29: You are right, we are using `city` schema only, however multiple database files provides that schema, `GeoIP2-Enterprise.mmdb`, `GeoLite2-City.mmdb` and `GeoLite2-Country.mmdb` all provide the city schema, Details can be found in the code of the geoip2-golang package, [see](https://github.com/oschwald/geoip2-golang/blob/main/reader.go#L273-L310).

The other database types are used to test unknown, or unsupported database types.

Does this make sense to you?

plugin/geoip/testdata/create_databases.go
snebel29: For the test cases to work is important to respect the schema and the values set in them, I also think this is the best way for future readers to discover and understand the mmdb file structure and current values.

Keeping the exact schema and values around is also very handy in case a simple modification like adding one extra value, or changing the value of a field is needed.

If you push for it I don't mind doing what you suggest, but I feel the value of keeping that code around and not just a few hints on how you could create one database, is higher than the value of removing it.

If it really bothers you to have a main file there, would it work for you to just move that to a `README.md` in the testdata directory?

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Create geoip ..." ]
>    probably still doesn't need to be a global var.
>
>It doesn't need to be a global var, it just felt using a global var was the
>simplest way to achieve what I needed which is temporarily set the net.IP used
>for validation to nil to force getting a different error during validation, to
>test an error edge case.
>
>I can try to do in a different way, using some struct and member field, or
>something like that. Is that what you want?
>
>    If you need an example IP use one from the documentation range: 192.0.2.0/
>    24 https://en.wikipedia.org/wiki/Reserved_IP_addresses
>
>Hmmm, what do you mean? I'm already using 127.0.0.1 as you suggested for
>database schema validation, for the mmdb files used for testing I can't create
>records using private IP scopes, it fails to create records using private
>networks (as shown before) so I had to use a public IP scope. Am I missing
>something?

ah, sorry, I thought 127.0.0.1 didn't work, if it does only better

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Create geoip ..." ]
>
>
>In plugin/geoip/city.go:
>
>> @@ -0,0 +1,60 @@
>+package geoip
>+
>+import (
>+       "context"
>+       "strconv"
>+
>+       "github.com/coredns/coredns/plugin/metadata"
>+       "github.com/oschwald/geoip2-golang"
>+)
>+
>+const defaultLang = "en"
>+
>+func (g GeoIP) setCityMetadata(ctx context.Context, data *geoip2.City) {
>+       // Set labels for city, country and continent names.
>+       if name, ok := data.City.Names[defaultLang]; ok {
>
>WDYT?

just use `name := data.City.Name[defaultLang]` this will be the default zero value if it
doesn't exist (IIRC) then.

I agree setting an error is making this more difficult than it needs to be.

/Miek

--
Miek Gieben

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Create geoip ..." ]
>WDYT?

remove what's not being used now. If we add stuff, we'll add it here as well.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Create geoip ..." ]
>Does this make sense to you?

yes, but seems a bit of an overkill, we already state in the doc what this plugin
supports.

Less is more at this point. If we extend things we'll add back in stuff.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Create geoip ..." ]
>
>
>In plugin/geoip/testdata/create_databases.go:
>
>> @@ -0,0 +1,117 @@
>+// This package is a helper function to create the mmdb databases used for testing in the
>+// geoip plugin.
>+package main
>
>For the test cases to work is important to respect the schema and the values
>set in them, I also think this is the best way for future readers to discover
>and understand the mmdb file structure and current values.

surely this is documented my mmdb?  And we can point to that documentation?

snebel29:
> [ Quoting ***@***.***> in "Re: [coredns/coredns] Create geoip ..." ] ***@***.*** commented on this pull request.
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ In plugin/geoip/city.go: > @@ -0,0 +1,60 @@ +package geoip + +import ( + "context" + "strconv" + + "github.com/coredns/coredns/plugin/metadata" + "github.com/oschwald/geoip2-golang" +) + +const defaultLang = "en" + +func (g GeoIP) setCityMetadata(ctx context.Context, data *geoip2.City) { + // Set labels for city, country and continent names. + if name, ok := data.City.Names[defaultLang]; ok { WDYT?
> just use `name := data.City.Name[defaultLang]` this will be the default zero value if it doesn't exist (IIRC) then. I agree setting an error is making this more difficult than it needs to be. /Miek
> […](#)
> -- Miek Gieben

Done

snebel29:
> [ Quoting ***@***.***> in "Re: [coredns/coredns] Create geoip ..." ]
> WDYT?
> remove what's not being used now. If we add stuff, we'll add it here as well.

OK, I removed all non used db types (Now there is city only)

snebel29:
> [ Quoting ***@***.***> in "Re: [coredns/coredns] Create geoip ..." ]
> Does this make sense to you?
> yes, but seems a bit of an overkill, we already state in the doc what this plugin supports. Less is more at this point. If we extend things we'll add back in stuff.

OK, I removed the Enterprise, Country and ISP databases and their associated test cases.

snebel29:
> [ Quoting ***@***.***> in "Re: [coredns/coredns] Create geoip ..." ] ***@***.*** commented on this pull request.
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ In plugin/geoip/testdata/create_databases.go: > @@ -0,0 +1,117 @@ +// This package is a helper function to create the mmdb databases used for testing in the +// geoip plugin. +package main For the test cases to work is important to respect the schema and the values set in them, I also think this is the best way for future readers to discover and understand the mmdb file structure and current values.
> surely this is documented my mmdb? And we can point to that documentation?

OK, I've added a `README.md` file into testdata and moved the code snipped there, please let me me know if that's what you wanted or if I misunderstood.

miekg:
very excellent. Thanks! 

Only the final step of adding yourself to the OWNERS file for this plugin. And I'll invite you to the repo as well

miekg:
*except* for the annoying merge conflicts in go.mod/sum

snebel29:
@miekg 

I just rebased all changes from master and resolved the conflicts in `go.mod` and `go.sum` followed by a `go mod tidy`, after that I also created a plugin line (sorted) in CODEOWNERS with you (should it be you and me or just me?) followed by me.

Is that everything?

snebel29:
Go Covergae check failed, with what smells some sort of flaky test

```
--- FAIL: TestMultiZoneBlockConfigs (0.00s)
    server_test.go:124: Could not get CoreDNS serving instance: Listen: listen tcp :40000: bind: address already in use
```

miekg:
yes, that test turns out to be flaky, in general fix port numbers are a bad idea, but there you sorta need it... Maybe some stupid retrying is in order there.

miekg (approved):

snebel29:
Thanks!

PR #4698, opened 2021-06-14
plugin/transfer: Support CIDR notation for 'to'

### LINKED ISSUE #4690: transfer's to does not allow IPv6 CIDR address
**What happened**:

The documentation contradicts the implementation regarding to the **to** statement in the **transfer module**.

Using

        transfer {
          to 2a0a:e5c0::/29
          to 2a01:4f8:150:7092::2/64
        }

I get the error message

```
plugin/transfer: must specify an IP address: `2a0a:e5c0::/29'
```

Support AXFR to hosts coming from 2a0a:e5c0::/29.

See above. 

The documentation says:

```
to ADDRESS… The hosts transfer will transfer to. Use * to permit transfers to all addresses. ADDRESS must be denoted in CIDR notation (e.g., 127.0.0.1⁄32) or just as plain addresses. to may be specified multiple times.
```

Above is clearly cidr and the intent is to allow AXRFs from the given networks.

**Environment**:

- the version of CoreDNS: k8s.gcr.io/coredns/coredns:v1.8.0
- Corefile: See above
- logs, if applicable:
- Others:

chrisohaver:

Adds support for allowed IP networks (CIDR). 

Fixes #4690

### DIFF
--- a/plugin/transfer/README.md
+++ b/plugin/transfer/README.md
@@ -11,7 +11,7 @@ This plugin answers zone transfers for authoritative plugins that implement `tra
 *transfer* answers full zone transfer (AXFR) requests and incremental zone transfer (IXFR) requests
 with AXFR fallback if the zone has changed.
 
-When a plugin wants to notify it's secondaries it will call back into the *transfer* plugin.
+When a plugin wants to notify its secondaries it will call back into the *transfer* plugin.
 
 The following plugins implement zone transfers using this plugin: *file*, *auto*, *secondary*, and
 *kubernetes*. See `transfer.go` for implementation details if you are a plugin author that wants to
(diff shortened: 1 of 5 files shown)


miekg:
is this inline with what acl does?

chrisohaver:
> is this inline with what acl does?

Ah - yeah I should probably make this respond with `REFUSED`, instead of passing to next plugin.

Furthermore ... CIDR filtering can be handled entirely by _acl_ instead... albeit less succinctly.

This config using _acl_ ...

```
  acl {
    allow type AXFR net 10.1.0.0/16
    allow type IXFR net 10.1.0.0/16
    block type AXFR net *
    block type IXFR net *
  }
  transfer {
    to *
  }
```

... would be equivalent to the following config enabled in this PR  ...

```
  transfer {
    to 10.1.0.0/16
  }
```

So strictly speaking this feature isn't needed, unless you see enough value in the shorter config to warrant the functional redundancy.

chrisohaver:
I opened #4711 with doc changes if we don't want to add CIDR restriction directly to `transfer` via this PR.

chrisohaver:
>>is this inline with what acl does?

> Ah - yeah I should probably make this respond with REFUSED, instead of passing to next plugin.

Transfer already does that.  So  yes, the behavior is in-line with how _acl_ does it.

chrisohaver:
Closing since #4711 merged.

PR #4700, opened 2021-06-16
fallthrough: make clear this is a bad idea

### LINKED ISSUE #4545: Add fallthrough to file plugin, or add wildcards to hosts plugin
Either of these:
1. fallthrough for `file` plugin
2. wildcard for `host` plugin

**Why is this needed**:
We need to have some static records (some of them wildcard) for a domain which coredns is not authoritative for. `hosts` plugin have fallthrough option, but does not support wildcard. `file` plugin supports wildcard, but doesn't allow to fallthrough if domain is not found.

example.com zone:
```
aa.example.com 1.2.3.4
bb.example.com 2.3.4.5
*.apps.example.com 5.6.7.8
```
but there are also other domains under `example.com` which is not managed by our team (`cc.example.com` and we want to fallthrough to `forward` plugin for them.

I know we can use `template` plugin for that, but there are many static records which will result in a very long and messy Corefile. Also it's much better to have a separate file for static records and keep Corefile for our  coredns configuration (separate reload, human errors, etc)

**related issues:**
https://github.com/coredns/coredns/issues/3601

miekg:
Update the docs to reflect what's in the docs of
plugins/pkg/fallthrough.

Closes: #4545

### DIFF
--- a/plugin.md
+++ b/plugin.md
@@ -125,11 +125,15 @@ In this example the *file* plugin is handling all names below (and including) `e
 a query comes in that is not a subdomain (or equal to) `example.org` the next plugin is called.
 
 Now, the world isn't perfect, and there are may be reasons to "fallthrough" to the next plugin,
-meaning a plugin is only responsible for a *subset* of names within the zone.
+meaning a plugin is only responsible for a *subset* of names within the zone. But this is a very
+dangerous route to take, because the multiple sources can have very different ideas on what names
+exist, this can confuse caches and resolvers.
 
 The `fallthrough` directive should optionally accept a list of zones. Only queries for records
 in one of those zones should be allowed to fallthrough. See `plugin/pkg/fallthrough` for the
 implementation.
+
+Use this directive should be a last resort and is almost never needed (nor a good idea).
 
 ## Mutating a Response
 


johnbelamaric:
Hmph. Not so sure I agree with you. I think having this makes us able to accomplish things that are very powerful. And I don't see why it's so bad.

SuperQ:
Maybe better than "good" vs "bad", we document valid use cases, and ones that are problematic / have sharp edges.

I find it helps users when the intended use case is explicitly spelled out.

chrisohaver:
> Maybe better than "good" vs "bad", we document valid use cases, and ones that are problematic / have sharp edges.
> 
> I find it helps users when the intended use case is explicitly spelled out.

Yeah, giving examples of how fallthrough can confuse caches and resolvers would help a plugin developer decide if they should implement it in their plugin. It would also help _users_ of fallthrough understand some of the pitfalls and traps they might walk into using the fallthrough option.

dilyevsky (approved):

miekg:
fair points, but I'm afraid this will turn into a large document. I'll try to extend this when I find some cycles

chrisohaver:
I know this has been sitting for a long time, but I have been going through old PRs and I thought I'd poke this with another comment.

> But [allowing fallthtough] is a very dangerous route to take, because the _multiple sources can have very different ideas on what names exist, this can confuse caches and resolvers.

Specifically related to caches, this seems more like a reason for dissuading split-horizon because in a split-horizon dns, answers can be different for different resolvers (resulting in cache poisoning). However in fall-through behavior, the answer is consistent.

One concrete problem with fall-though that comes to mind is _zone transfer_.  If a plugin implements zone transfer, then making it include "fall-through" records would be intractably complex to implement.  Doing so would, for example require the plugin that it falls-through to also implements zone transfer, and the first plugin merging the results. I think it's unlikely this would ever be implemented, and thus a plugin would not really transfer its entire zone.

PR #4702, opened 2021-06-17
Route53 support Failover

### LINKED ISSUE #4701: Route53 not support Failover
aws route53 supports failover, but not work in coredns route53 plugin

route53 failover config:
```
s-search-hdfs-nn.test.release | A | Failover | Secondary | 10.60.8.187 10.60.0.129 10.60.11.82
s-search-hdfs-nn.test.release | A | Failover | Primary | 10.60.13.249 10.60.6.242 10.60.12.249
```

The domain name is resolved on the eks node as follows：
```
# nslookup s-search-hdfs-nn.test.release
Server:		10.60.0.2
Address:	10.60.0.2#53

Non-authoritative answer:
Name:	s-search-hdfs-nn.test.release
Address: 10.60.13.249
Name:	s-search-hdfs-nn.test.release
Address: 10.60.12.249
Name:	s-search-hdfs-nn.test.release
Address: 10.60.6.242
```
But the result of resolving the domain name in the pod is as follows：
```
nslookup s-search-hdfs-nn.test.release
Server:         172.20.0.10
Address:        172.20.0.10#53

Name:   s-search-hdfs-nn.test.release
Address: 10.60.8.187
Name:   s-search-hdfs-nn.test.release
Address: 10.60.0.129
Name:   s-search-hdfs-nn.test.release
Address: 10.60.11.82
Name:   s-search-hdfs-nn.test.release
Address: 10.60.13.249
Name:   s-search-hdfs-nn.test.release
Address: 10.60.6.242
Name:   s-search-hdfs-nn.test.release
Address: 10.60.12.249

```

icyxp:
issue link: [issue-923363220](https://github.com/coredns/coredns/issues/4701#issue-923363220)

Closes: #4701

### DIFF
--- a/plugin/route53/route53.go
+++ b/plugin/route53/route53.go
@@ -254,12 +254,55 @@ func (h *Route53) updateZones(ctx context.Context) error {
 				}
 				err = h.client.ListResourceRecordSetsPagesWithContext(ctx, in,
 					func(out *route53.ListResourceRecordSetsOutput, last bool) bool {
+						failoverList := map[string]map[string]*route53.ResourceRecordSet{}
 						for _, rrs := range out.ResourceRecordSets {
+							if rrs.SetIdentifier != nil {
+								if failoverList[*rrs.Name] == nil {
+									failoverList[*rrs.Name] = map[string]*route53.ResourceRecordSet{}
+								}
+								failoverList[*rrs.Name][*rrs.Failover] = rrs
+								continue
+							}
+
 							if err := updateZoneFromRRS(rrs, newZ); err != nil {
 								// Maybe unsupported record type. Log and carry on.
 								log.Warningf("Failed to process resource record set: %v", err)
 							}
 						}
+
+						for k := range failoverList {
+							primary := failoverList[k]["PRIMARY"]
+							secondary := failoverList[k]["SECONDARY"]
+
+							hcInput := &route53.GetHealthCheckStatusInput{
+								HealthCheckId: primary.HealthCheckId,
+							}
+							status, _ := h.client.GetHealthCheckStatus(hcInput)
+
+							count := len(status.HealthCheckObservations)
+							failedNum := 0
+							for _, v := range status.HealthCheckObservations {
+								if strings.HasPrefix(*v.StatusReport.Status, "Failure") {
+									failedNum++
+								}
+							}
+
+							if failedNum >= count/2 {
+								log.Warningf("primary hc failed, failed num %d, skip %s primary record, use secondary record", failedNum, *primary.SetIdentifier)
+								if secondary != nil {
+									if err := updateZoneFromRRS(secondary, newZ); err != nil {
+										// Maybe unsupported record type. Log and carry on.
+										log.Warningf("Failed to process resource record set: %v", err)
+									}
+								}
+							} else {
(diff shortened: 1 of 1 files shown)


yongtang:
@icyxp Can you take a look at the DCO issue?

icyxp:
> @icyxp Can you take a look at the DCO issue?

Okay, I fixed it well.

plugin/route53/route53.go
```diff
@@ -254,12 +254,55 @@ func (h *Route53) updateZones(ctx context.Context) error {
 				}
 				err = h.client.ListResourceRecordSetsPagesWithContext(ctx, in,
 					func(out *route53.ListResourceRecordSetsOutput, last bool) bool {
+						failoverList := map[string]map[string]*route53.ResourceRecordSet{}
 						for _, rrs := range out.ResourceRecordSets {
+							if rrs.SetIdentifier != nil {
```
dilyevsky, line 259: This will match every non-simple RR type not just failover. From docs:

```
Resource record sets that have a routing policy other than simple...
```

plugin/route53/route53.go
```diff
@@ -254,12 +254,55 @@ func (h *Route53) updateZones(ctx context.Context) error {
...
+								continue
+							}
+
 							if err := updateZoneFromRRS(rrs, newZ); err != nil {
 								// Maybe unsupported record type. Log and carry on.
 								log.Warningf("Failed to process resource record set: %v", err)
 							}
 						}
+
+						for k := range failoverList {
```
dilyevsky, line 273: This should go into separate method and I don't think it's desired to have healtcheck performed within a hot loop - it should be performed async

plugin/route53/route53.go
```diff
@@ -254,12 +254,55 @@ func (h *Route53) updateZones(ctx context.Context) error {
...
+
+							hcInput := &route53.GetHealthCheckStatusInput{
+								HealthCheckId: primary.HealthCheckId,
+							}
+							status, _ := h.client.GetHealthCheckStatus(hcInput)
+
+							count := len(status.HealthCheckObservations)
+							failedNum := 0
+							for _, v := range status.HealthCheckObservations {
+								if strings.HasPrefix(*v.StatusReport.Status, "Failure") {
```
dilyevsky, line 285: Is there no better way to do this? Also should probably log unsuccessful checks.

dilyevsky: And short-circuit when threshold is exceeded.

plugin/route53/route53.go
```diff
@@ -254,12 +254,55 @@ func (h *Route53) updateZones(ctx context.Context) error {
...
+									failedNum++
+								}
+							}
+
+							if failedNum >= count/2 {
```
dilyevsky, line 290: Why more than half?

dilyevsky (changes requested):
Thanks for the PR!

Couple issues here:
* This only covers active-passive failover RRs which should be documented at least. For example, I could setup failover with just weighted records in active-active mode and I don't think this code handles this correctly.
* Would like to avoid doing expensive work and especially calling remote RPCs within a hot path
* Needs tests

chrisohaver:
@icyxp, do you plan to complete this PR.

PR #4712, opened 2021-06-22
plugin/dns64: Persist truncated state to client if AAAA response is truncated

chrisohaver:

Persists the TC bit to the client response if the `AAAA` lookup response is truncated.
If not done, the client will think it got a complete response when the upstream response was truncated, _resulting in a wrong answer given to the client_.


chrisohaver:
ptal, @SuperQ

SuperQ (approved):
Good catch.

LGTM

PR #4713, opened 2021-06-22
plugin/template: Persist truncated state to client if CNAME lookup response is truncated

chrisohaver:

### DIFF
--- a/plugin/template/template.go
+++ b/plugin/template/template.go
@@ -11,7 +11,6 @@ import (
 	"github.com/coredns/coredns/plugin/metadata"
 	"github.com/coredns/coredns/plugin/metrics"
 	"github.com/coredns/coredns/plugin/pkg/fall"
-	"github.com/coredns/coredns/plugin/pkg/upstream"
 	"github.com/coredns/coredns/request"
 
 	"github.com/miekg/dns"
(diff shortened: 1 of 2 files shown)


chrisohaver:
cc @rtreffer

plugin/template/template.go
```diff
@@ -100,8 +104,10 @@ func (h Handler) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 			}
 			msg.Answer = append(msg.Answer, rr)
 			if template.upstream != nil && (state.QType() == dns.TypeA || state.QType() == dns.TypeAAAA) && rr.Header().Rrtype == dns.TypeCNAME {
-				up, _ := template.upstream.Lookup(ctx, state, rr.(*dns.CNAME).Target, state.QType())
-				msg.Answer = append(msg.Answer, up.Answer...)
+				if up, err := template.upstream.Lookup(ctx, state, rr.(*dns.CNAME).Target, state.QType()); err == nil && up != nil {
```
rtreffer, line 107: the previous code accessed up without an explicit nil check, I think this is / should be part of the interface contract.

rtreffer (approved):
Yeah, dropping the truncate bit seems wrong. We append to the answer section, so dropping answers (by adding an incomplete list) should result in TC being set.

Thank you for adding a test, too :heart_eyes_cat: It could benefit from a positive case, but this is already way better than before.

And thanks for pinging again, this fell through :see_no_evil:

PR #4715, opened 2021-06-23
plugin/etcd+kubernetes: Persist truncated state to client if CNAME lookup response is truncated

chrisohaver:

For backend and backend using plugins _etcd_ and _kubernetes_...


chrisohaver:
Merge conflicts resolved.

PR #4716, opened 2021-06-23
plugin/k8s_external: Persist tc bit from lookup to client response

chrisohaver:


chrisohaver:
Merge conflicts resolved.

PR #4718, opened 2021-06-27
add configurable log level to errors plugin

Tantalor93:
Hello, in our company we are big fans of CoreDNS and we are running several of our services on top of CoreDNS with our custom made plugins and this is one of the changes of errors plugin we made to actually help us classify errors better (for example we are consolidating i/o timeout caused by unstable upstream as warning instead of errors). I would appreciate your feedback on this proposed change and if you require any changes I can make them :) 

This pull request adds a new optional option to errors plugin consolidate functionality and enables configuring log level on which the error is logged. It enables the user of CoreDNS to classify errors into more severe and less severe by assigning different kinds of errors lower log levels (like a warning). An example of usage is like this, which logs timeout errors as warnings
```
.:53 {
   forward . 8.8.4.4 8.8.8.8
   errors {
    consolidate 10s "read tcp ([0-9a-f.:]+)->([0-9a-f.:]+): i/o timeout" warn
    consolidate 10s "read udp ([0-9a-f.:]+)->([0-9a-f.:]+): i/o timeout" warn
  }
}
```

no issue is related

I have updated README of errors plugin, I don't know whether some other documentation is needed to be updated?

this feature should be backward compatible, as this option to configure log level for error consolidation is optional and if not specified, it fallbacks to error

### DIFF
--- a/plugin/errors/README.md
+++ b/plugin/errors/README.md
@@ -22,12 +22,12 @@ Extra knobs are available with an expanded syntax:
 
 ~~~
 errors {
-	consolidate DURATION REGEXP
+	consolidate DURATION REGEXP <LEVEL?>
 }
 ~~~
 
-Option `consolidate` allows collecting several error messages matching the regular expression **REGEXP** during **DURATION**. After the **DURATION** since receiving the first such message, the consolidated message will be printed to standard output, e.g.
-
+Option `consolidate` allows collecting several error messages matching the regular expression **REGEXP** during **DURATION**. After the **DURATION** since receiving the first such message, the consolidated message will be printed to standard output with
+log level, which is configurable by optional option **LEVEL**, e.g. Supported log levels for **LEVEL** option are `warn`,`error`,`info`,`debug`.
 ~~~
 2 errors like '^read udp .* i/o timeout$' occurred in last 30s
 ~~~
@@ -53,7 +53,7 @@ Use the *forward* to resolve queries via 8.8.8.8 and print consolidated error me
 . {
     forward . 8.8.8.8
     errors {
-        consolidate 5m ".* i/o timeout$"
+        consolidate 5m ".* i/o timeout$" warn
         consolidate 30s "^Failed to .+"
     }
 }
--- a/plugin/errors/README.md
+++ b/plugin/errors/README.md
@@ -22,12 +22,12 @@ Extra knobs are available with an expanded syntax:
 
 ~~~
 errors {
-	consolidate DURATION REGEXP <LEVEL?>
+	consolidate DURATION REGEXP [LEVEL]
 }
 ~~~
 
 Option `consolidate` allows collecting several error messages matching the regular expression **REGEXP** during **DURATION**. After the **DURATION** since receiving the first such message, the consolidated message will be printed to standard output with
-log level, which is configurable by optional option **LEVEL**, e.g. Supported log levels for **LEVEL** option are `warn`,`error`,`info`,`debug`.
+log level, which is configurable by optional option **LEVEL**. Supported options for **LEVEL** option are `warn`,`error`,`info` and `debug`.
 ~~~
 2 errors like '^read udp .* i/o timeout$' occurred in last 30s
 ~~~
--- a/plugin/errors/setup_test.go
+++ b/plugin/errors/setup_test.go
@@ -7,7 +7,7 @@ import (
 	"testing"
 
 	"github.com/coredns/caddy"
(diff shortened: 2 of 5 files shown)


miekg:
think this is a fair change. For the regexp matching we don't see the level? I.e. we only work on the error text? Other wise we could extend the regexp matching without adding a new keyword?

plugin/errors/setup_test.go
```diff
@@ -1,9 +1,13 @@
 package errors
 
 import (
+	"bytes"
+	golog "log"
+	"strings"
 	"testing"
 
 	"github.com/coredns/caddy"
+	corednslog "github.com/coredns/coredns/plugin/pkg/log"
```
miekg:
```suggestion
	clog "github.com/coredns/coredns/plugin/pkg/log"
```

Tantalor93: 👍 fixed

Tantalor93:
@miekg sadly for regexp matching we don't see log level, we only see the error message received from the next plugin in the chain (as you said). So for example for timeout error, the `errors` plugin sees error like `read udp 192.168.1.179:63338->9.2.3.3:53: i/o timeout`

plugin/errors/README.md
```diff
@@ -22,12 +22,12 @@ Extra knobs are available with an expanded syntax:
 
 ~~~
 errors {
-	consolidate DURATION REGEXP
+	consolidate DURATION REGEXP <LEVEL?>
```
miekg:
```suggestion
	consolidate DURATION REGEXP [LEVEL]...
```

miekg: Also do we want multiple levels? Or do we assume the same error text will never be in multiple levels?

Tantalor93: I think it does not make sense to log the error message on multiple log levels. The error should be logged either as ERROR, WARNING, DEBUG, or INFO

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] add configura..." ]
>@Tantalor93 commented on this pull request.
>
>
>In plugin/errors/README.md:
>
>> @@ -22,12 +22,12 @@ Extra knobs are available with an expanded syntax:
>
> ~~~
> errors {
>-       consolidate DURATION REGEXP
>+       consolidate DURATION REGEXP <LEVEL?>
>
>I think it does not make sense to log the error message on multiple log levels.
>The error should be logged either as ERROR, WARNING, DEBUG, or INFO

ack, that makes sense

plugin/errors/README.md
```diff
@@ -22,12 +22,12 @@ Extra knobs are available with an expanded syntax:
 
 ~~~
 errors {
-	consolidate DURATION REGEXP
+	consolidate DURATION REGEXP [LEVEL]...
```
miekg:
```suggestion
	consolidate DURATION REGEXP [LEVEL]
```
as discussed, there will only be one

plugin/errors/README.md
```diff
@@ -22,12 +22,12 @@ Extra knobs are available with an expanded syntax:
...
 }
 ~~~
 
-Option `consolidate` allows collecting several error messages matching the regular expression **REGEXP** during **DURATION**. After the **DURATION** since receiving the first such message, the consolidated message will be printed to standard output, e.g.
-
+Option `consolidate` allows collecting several error messages matching the regular expression **REGEXP** during **DURATION**. After the **DURATION** since receiving the first such message, the consolidated message will be printed to standard output with
+log level, which is configurable by optional option **LEVEL**, e.g. Supported log levels for **LEVEL** option are `warn`,`error`,`info`,`debug`.
```
miekg:
```suggestion
log level, which is configurable by optional option **LEVEL**. Supported options for **LEVEL**  are `warn`,`error`,`info` and `debug`.
```

plugin/errors/README.md
Tantalor93: 👍 right, fixed

plugin/errors/README.md
Tantalor93: thanks 👍 changed as you suggested

plugin/errors/README.md
```diff
@@ -53,7 +53,7 @@ Use the *forward* to resolve queries via 8.8.8.8 and print consolidated error me
 . {
     forward . 8.8.8.8
     errors {
-        consolidate 5m ".* i/o timeout$"
+        consolidate 5m ".* i/o timeout$" warn
```
chrisohaver, line 56: ~This addition should be reflected in the description of the example above it.~

Nevermind - already merged.

PR #4725, opened 2021-06-29
do not log NOERROR in log plugin when response is not available

### LINKED ISSUE #4724: Log plugin logs wrong rcode incase of error
**What happened**:
when no response is written from upstream plugins (for example due to timeout in forward plugin) the log plugin log incorrect rcode
```
[INFO] 127.0.0.1:63325 - 28625 "A IN 9gag.com. udp 49 false 4096" NOERROR - 0 6.010923182s
```

log plugin to log either SERVFAIL which is returned to the client or at least a placeholder instead (either is fine, but logging it as NOERROR is misleading)
```
[INFO] 127.0.0.1:63325 - 28625 "A IN 9gag.com. udp 49 false 4096" SERVFAIL - 0 6.010923182s
```

```
[INFO] 127.0.0.1:63325 - 28625 "A IN 9gag.com. udp 49 false 4096" - - 0 6.010923182s

```

run CoreDNS with Corefile (the invalid IP in forward plugin is there on purpose to better simulate errors)
```
.:53 {
    forward . 9.8.7.6
    log
}
```

run dig
```
dig @127.0.0.1 9gag.com
```

and look at the logs of CoreDNS
```
.:53
CoreDNS-1.8.4
darwin/amd64, go1.16, 
[INFO] 127.0.0.1:63325 - 28625 "A IN 9gag.com. udp 49 false 4096" NOERROR - 0 6.010923182s

```

**Environment**:

- the version of CoreDNS: 1.8.4
- Corefile:
```
.:53 {
    forward . 9.8.7.6
    log
}
```
- logs, if applicable:
- Others:

Tantalor93:

close #4724

the log plugin logs NOERROR rcode in case of no response is written, this PR instead changes this to log placeholder ( - ), which at least does not mislead the reader of logs

Fixes: #4724

I don't think we need to change documentation

I don't think so

### DIFF
--- a/plugin/pkg/replacer/replacer.go
+++ b/plugin/pkg/replacer/replacer.go
@@ -86,7 +86,7 @@ func appendValue(b []byte, state request.Request, rr *dnstest.Recorder, label st
 		return strconv.AppendInt(b, int64(state.Size()), 10)
 	// Recorded replacements.
 	case "{rcode}":
-		if rr == nil {
+		if rr == nil || rr.Msg == nil {
 			return append(b, EmptyValue...)
 		}
 		if rcode := dns.RcodeToString[rr.Rcode]; rcode != "" {
--- a/plugin/pkg/replacer/replacer_test.go
+++ b/plugin/pkg/replacer/replacer_test.go
@@ -393,3 +393,52 @@ func TestMetadataMalformed(t *testing.T) {
 		}
 	}
 }
+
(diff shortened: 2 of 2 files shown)


chrisohaver (approved):
LGTM, thanks!

FWIW, Observationally I don't think `rr` ever be nil. I wonder if the original intent was to do the same for the `rsize` and `duration` checks - i.e. to record empty value when the message is not written to the client (rr.Msg == nil)).

PR #4727, opened 2021-06-29
when no response is written, fallback to status of next plugin in prometheus plugin

### LINKED ISSUE #4726: when no response is written to client, prometheus plugin reports it as NOERROR response
**What happened**:
when no response is written from upstream plugins (for example due to timeout in forward plugin) the prometheus plugin reports it as NOERROR (successful) response

```
curl http://localhost:9153/metrics | grep coredns_dns_responses_total

coredns_dns_responses_total{rcode="NOERROR",server="dns://:53",zone="."} 2
```

I would expect that it will be reported as SERVFAIL as it is returned to client as SERVFAIL
```
coredns_dns_responses_total{rcode="SERVFAIL",server="dns://:53",zone="."} 2
```

run CoreDNS with Corefile (the invalid IP in forward plugin is there on purpose to better simulate errors)
```
.:53 {
    forward . 9.8.7.6
    prometheus []:9153
}
```

run dig
```
dig @127.0.0.1 9gag.com
```

get metrics and look at the coredns_dns_responses_total metric
```
curl http://localhost:9153/metrics | grep coredns_dns_responses_total

# HELP coredns_dns_responses_total Counter of response status codes.
# TYPE coredns_dns_responses_total counter
coredns_dns_responses_total{rcode="NOERROR",server="dns://:53",zone="."} 2
```

**Environment**:

- the version of CoreDNS: 1.8.4 (latest)
- Corefile:
```
.:53 {
    forward . 9.8.7.6
    prometheus []:9153
}
```
- logs, if applicable:
- Others:

Tantalor93:

close #4726

when no response is written from up the chain of plugins, the default value of dnstest.Recorder for rcode (0) is used as rcode reported to the `coredns_dns_responses_total` metric, which is misleading and wrong. This PR changes the behaviour that when no response is written, the return status of the next plugin is used.

#4726

I think no changes are needed

I don't think so, it fixes the behaviour

### DIFF
--- a/plugin/metrics/handler.go
+++ b/plugin/metrics/handler.go
@@ -28,6 +28,9 @@ func (m *Metrics) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 
 	rc := rw.Rcode
 	if !plugin.ClientWrite(status) {
+		// when no response was written, fallback to status returned from next plugin as this status
+		// is actually used as rcode of DNS response
+		// see https://github.com/coredns/coredns/blob/master/core/dnsserver/server.go#L318
 		rc = status
 	}
 	vars.Report(WithServer(ctx), state, zone, rcode.ToString(rc), rw.Len, rw.Start)
--- a/plugin/metrics/handler.go
+++ b/plugin/metrics/handler.go
@@ -26,7 +26,11 @@ func (m *Metrics) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	rw := dnstest.NewRecorder(w)
 	status, err := plugin.NextOrFailure(m.Name(), m.Next, ctx, rw, r)
 
-	vars.Report(WithServer(ctx), state, zone, rcode.ToString(rw.Rcode), rw.Len, rw.Start)
+	rc := rw.Rcode
+	if !plugin.ClientWrite(status) {
+		rc = status
+	}
+	vars.Report(WithServer(ctx), state, zone, rcode.ToString(rc), rw.Len, rw.Start)
 
 	return status, err
 }


plugin/metrics/handler.go
```diff
@@ -26,7 +26,11 @@ func (m *Metrics) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	rw := dnstest.NewRecorder(w)
 	status, err := plugin.NextOrFailure(m.Name(), m.Next, ctx, rw, r)
 
-	vars.Report(WithServer(ctx), state, zone, rcode.ToString(rw.Rcode), rw.Len, rw.Start)
+	rc := rw.Rcode
+	if !plugin.ClientWrite(status) {
```
miekg, line 30: so *status* is _NOT_ the rcode. Let me re-read your cover letter for the exact problem.
The status is _only_ used to determine if there still is a need to respond to the client (with an error)

miekg: We need some comments here, as we munging the status into an rcode. It might make sense to 
a) default to SERVFAIL in these cases
b) use "-" as the rcode string as we don't actually know what happened

WDYT?

Tantalor93: From what I understood from the CoreDNS codebase, if no plugin writes a response, the status (return code from the chain of plugins) is used to construct a DNS message with error https://github.com/coredns/coredns/blob/master/core/dnsserver/server.go#L318, but I agree that without this context it seems like mixing two things in here.

I think out of the two options you listed I prefer the a) as it seems to be the response the DNS client will receive from CoreDNS instance and I would expect that metrics like `coredns_dns_responses_total{rcode}` will be showing proper rcodes the client receives.

so something like would be fine with you?

```
rc := rw.Rcode
// no response was written, fallback to SERVFAL rcode for metrics
if !plugin.ClientWrite(status) {
		rc = dns.RcodeServerFailure
}
```

or should I just add clarifying comment to the current solution?

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] when no respo..." ]
>so something like would be fine with you?
>
>rc := rw.Rcode
>// no response was written, fallback to SERVFAL rcode for metrics
>if !plugin.ClientWrite(status) {
>                rc = dns.RcodeServerFailure
>}
>
>or should I just add clarifying comment to the current solution?

I see we actually do use this as rcode in core/dnsserver, `errorFunc` creates
a message with 'status' as rcode. So the previous code was correct. Might make sense to
reference that server code though in a comment?

Tantalor93:
@miekg sure, added the comment referencing the error handler, hope it is clear enough

miekg:
Ack. I just realized this may also impact the log plugin, but that's a
different pr anyway

Tantalor93:
Yea, the log plugin was also affected, I changed the behaviour as part of https://github.com/coredns/coredns/pull/4725

plugin/metrics/handler.go
miekg: what you have now looks good

miekg (approved):

Issue #4735, opened 2021-07-06
Regression: CoreDNS in version 1.8.0 fails to resolve domain (works in 1.6.9) 


lhotrifork:
**What happened**:
When using k3s with CoreDNS 1.8.0 for a specific domain (which is my corporate http proxy) returns 'rflags' in a different way than when using CoreDNS 1.6.9. 
In practice this means that OpenID Connect using OpenResty as a reverse-proxy fails. Downgrading to 1.6.9 fixes the issue. CoreDNS in version 1.7.X has the same issue.
I cannot easily upgrade to 1.8.3 as it depends on `endpointSlices`.

I've narrowed it down to the following difference (which may be a wild goose-chase).
For 1.6.9 CoreDNS logs the following way for the domain:
```
[INFO] 10.42.0.220:44598 - 59764 "A IN proxy.mydomain.com. udp 32 false 512" NOERROR qr,aa,rd,ra 336 0.000129347s
```
For 1.8.0 it logs:
```
[INFO] 10.42.0.228:33081 - 63992 "A IN proxy.mydomain.com. udp 32 false 512" NOERROR qr,rd,ra,ad 336 0.000702693s
```
The difference seems to be in the 'rflags'.

**Anything else we need to know?**:
As this is my proxy server a lot of requests also goes through it, and for those request there's also a difference between 1.6.9 and 1.8.0.
1.6.9:
```
[INFO] 10.42.0.213:36482 - 19579 "A IN proxy.mydomain.com.flux-system.svc.cluster.local. udp 62 false 512" NXDOMAIN qr,aa,rd 155 0.000166089s
[INFO] 10.42.0.213:35226 - 61848 "AAAA IN proxy.mydomain.com.flux-system.svc.cluster.local. udp 62 false 512" NXDOMAIN qr,aa,rd 155 0.003467135s
[INFO] 10.42.0.213:37512 - 34923 "A IN proxy.mydomain.com.svc.cluster.local. udp 50 false 512" NXDOMAIN qr,aa,rd 143 0.000105847s
[INFO] 10.42.0.213:59737 - 16377 "AAAA IN proxy.mydomain.com.svc.cluster.local. udp 50 false 512" NXDOMAIN qr,aa,rd 143 0.000061375s
[INFO] 10.42.0.213:55805 - 65513 "AAAA IN proxy.mydomain.com.cluster.local. udp 46 false 512" NXDOMAIN qr,aa,rd 139 0.000271467s
[...]

PR #4736, opened 2021-07-08
plugin/cache: Unset AD flag when DO is not set for cache miss

chrisohaver:

Unset AD header flag when DO is not set in request.
This is [already done for cache hits](https://github.com/coredns/coredns/blob/master/plugin/cache/item.go#L68-L70), but was not for cache miss.

related to #4735


zouyee (approved):

nextgens:
This is in breach of https://datatracker.ietf.org/doc/html/rfc6840#section-5.8 ... which makes perfectly clear that DO or AD in the query should be treated the same (and the AD flag preserved).

PR #4743, opened 2021-07-10
add metrics for total of log messages per log level

Tantalor93:

This PR adds a new Prometheus counter `coredns_log_messages_total{level}` which counts the number of logged messages per log level globally (across all plugins), for example how many errors were logged, how many warnings, etc. 

Example of new metrics
```
# HELP coredns_log_messages_total Counter of logged messages.
# TYPE coredns_log_messages_total counter
coredns_log_messages_total{level="error"} 2
coredns_log_messages_total{level="info"} 2
```

This enables us to monitor if there are any error logs being produced by the CoreDNS (we can set up an alert in Prometheus Alertmanager on this and if it fires, we can go to check the logs) or if there are too many warnings, etc.

no issue related

I guess none

no

### DIFF
--- a/plugin/pkg/log/log.go
+++ b/plugin/pkg/log/log.go
@@ -20,6 +20,14 @@ import (
 // it can not be unset.
 var D = &d{}
 
+var levelToLabel = map[string]string{
+	debug:   debugLabel,
+	info:    infoLabel,
+	warning: warningLabel,
+	err:     errorLabel,
+	fatal:   fatalLabel,
+}
+
 type d struct {
 	on bool
 	sync.RWMutex
@@ -50,11 +58,13 @@ func (d *d) Value() bool {
 // logf calls log.Printf prefixed with level.
 func logf(level, format string, v ...interface{}) {
 	golog.Print(level, fmt.Sprintf(format, v...))
+	LogMessagesTotal.WithLabelValues(levelToLabel[level]).Inc()
 }
 
 // log calls log.Print prefixed with level.
 func log(level string, v ...interface{}) {
 	golog.Print(level, fmt.Sprint(v...))
+	LogMessagesTotal.WithLabelValues(levelToLabel[level]).Inc()
 }
 
 // Debug is equivalent to log.Print(), but prefixed with "[DEBUG] ". It only outputs something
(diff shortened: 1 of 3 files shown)


miekg:
I think this is overly broad. If an error is logged and it's _that_ important there should probably be a metric for that thing specifcally. This also help with debugging as you don't directly need to go back to the logs, because the metric tells directly.

Tantalor93:
@miekg yea, I think I get your point. Although It still seems to me that having this kind of metrics is quite common in systems and applications, for example [Loki](https://grafana.com/docs/loki/latest/operations/observability/), [Spring boot applications using Logback has it OOB](https://spring.io/blog/2018/03/16/micrometer-spring-boot-2-s-new-application-metrics-collector).

In Wandera we had to painfully add this metric to CoreDNS via separate plugin (not proud of this code 😀). As we are often using this kind of metrics for alerting not only on errors, but also on warnings, when they exceed certain threshold (like errors that are not that critical) and having metric for everything seemed like a overkill for us.

yongtang:
@Tantalor93  I understand the traditional way is to add metrics to code path of the program  itself (e.g., coredns) whenever there is a need for adding new metrics. However, I do want to raised the prospect of alternative paths. In other words, is there a way to NOT add metrics inside the coredns, but add metrics as part of the log processing phase?

From my understand, the `metrics for total of log messages per log level` you are trying to add, is already available in the logging itself. Can we think about process the log and generate the metrics? This will avoid updating the code of coredns itself, and easier/quicker to see this happen (no need to wait for coredns update).

Tantalor93:
@yongtang i think what you are describing would also be possible 🙂 but as I said, we already have it solved using custom CoreDNS plugin, I was just looking for easier solution so I created this PR with what I came up with. I was looking for the input whether you like it or not. We can close this PR as declined if you are not interested in this enhancement.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] add metrics f..." ]
***@***.*** i think what you are describing would also be possible 🙂 but as I
>said, we already have it solved using custom CoreDNS plugin, I was just looking
>for easier solution so I created this PR with what I came up with. I was
>looking for the input whether you like it or not. We can close this PR as
>declined if you are not interested in this enhancement.

This is not a good way forward. Adding metrics always has a cost, so we should also be
careful when we add one.

Technically the thing you're after can also be done in your logging pipeline.

Tantalor93:
Ok, closing as agreed

PR #4749, opened 2021-07-12
publish trace id as metadata from trace plugin

Tantalor93:

Hello, I would appreciate your input/comments regarding this enhancement of trace plugin.

It adds zipkin/datadog trace identifier to the metadata published by `trace` plugin so that the trace ID can be reused from different plugins (like `log` plugin). By publishing trace ID as metadata, it can be used from `log` plugin => it can be logged as part of query logging and therefore enables correlating logs of CoreDNS with traces in tracing systems (like Zipkin) and enables better debugging of problematic DNS queries.

For example, let's have a CoreDNS instance with Corefile
```
.:53 {
    forward . 8.8.8.8
    log . "[{/trace/traceid}] '{type} {class} {name} {proto}:{port} {size} {>do} {>bufsize}' {rcode} {>rflags} {rsize} {duration}"
    trace zipkin 127.0.0.1:9411 {
        service dns-service
    }
    metadata
}
```

and Zipkin running in docker container
```
```

when I run DNS query against this CoreDNS instance
```
dig @127.0.0.1 example.com
```

the CoreDNS will log
```
[INFO] [019bfc0577fd9975] 'A IN example.com. udp:64410 52 false 4096' NOERROR qr,rd,ra,ad 67 0.010437056s
```

and by taking the trace Id from the log `019bfc0577fd9975`, I can now directly view trace in Zipkin UI http://localhost:9411/zipkin/traces/019bfc0577fd9975 without the need to search for the related trace

image

no issue related

I have enhanced the README of the trace plugin

no

### DIFF
--- a/plugin/trace/README.md
+++ b/plugin/trace/README.md
@@ -93,6 +93,13 @@ trace tracinghost:9411 {
 }
 ~~~
 
+## Metadata
+
+The trace plugin will publish the following metadata, if the *metadata*
+plugin is also enabled:
+
+* `trace/traceid`: identifier of (zipkin/datadog) trace of processed request 
+
 ## See Also
 
 See the *debug* plugin for more information about debug logging.
--- a/plugin/trace/trace.go
+++ b/plugin/trace/trace.go
@@ -8,6 +8,7 @@ import (
 	"sync/atomic"
 
 	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/metadata"
 	"github.com/coredns/coredns/plugin/pkg/dnstest"
 	"github.com/coredns/coredns/plugin/pkg/log"
 	"github.com/coredns/coredns/plugin/pkg/rcode"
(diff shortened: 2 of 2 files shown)


johnbelamaric:
This should already be in the context, I don't think we need to publish it in metadata?

Tantalor93:
@johnbelamaric yea, span context should be part of the context, but by exposing it as metadata it can be easily used from log plugin for logging query with trace id for correlation, that is ultimately what I want to achieve.

Other option coming to my mind is probably adding new placeholder ( https://github.com/coredns/coredns/tree/master/plugin/log#log-format )  to log plugin and when it is used in Corefile, the log plugin will pull the traceid from the context 🤔  What do you think?

johnbelamaric:
Yeah, come to think of it, this is probably better because it localizes the change here and doesn't create an inter-plugin dependency.

Thanks. What's the CI failure?

chrisohaver:
> What's the CI failure?

Circle CI flake.

miekg (approved):

PR #4751, opened 2021-07-14
deflake TestMultiZoneBlockConfigs

miekg:
Deflake by retrying and adding random port numbers. We try 3 times to
get an instance.

Also fix a bug where server.Stop() was called even if the server
creation failed - this was never hit due to t.Fatal() above it, but fix
that nontheless.

### DIFF
--- a/test/server_test.go
+++ b/test/server_test.go
@@ -1,10 +1,13 @@
 package test
 
 import (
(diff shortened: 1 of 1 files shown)


snebel29 (approved):
LGTM

PR #4752, opened 2021-07-14
plugin/header: Introduce header plugin

mqasimsarfraz:
There are cases where users want to change the response flags to fit their needs. e.g `dnsmasqd` needs `ra` for caching so having a generic way to changes response might be helpful in such cases. 

#4435 #4541

A usage guide for the plugin needs to be added to the documentation.
No

### DIFF
--- a/plugin/header/README.md
+++ b/plugin/header/README.md
@@ -1,0 +1,42 @@
+# header
+
+## Name
+
+*header* - modifies the header for all the responses.
+
+## Description
+It ensures that the flags are in the desired state for all the responses. The modifications are made transparently for the client.
+
+## Syntax
+
+~~~
+header {
+    ACTION FLAGS...
+    ACTION FLAGS...
+}
+~~~
+
+* **ACTION**  defines the state for dns flags, allowed values set/clear. 
+* **FLAGS** are the dns flags that will be modified e.g `ra`, `aa` etc.
+
+## Examples
+
+Make sure recursive available `ra` flag is set in all the responses:
+
+~~~ corefile
+. {
+    header {
+        set ra
+    }
+}
+~~~
+Make sure recursive available `ra` and authoritative `aa` flags are set and recursive desired is cleared in all the responses:
+
+~~~ corefile
+. {
+    header {
+        set ra aa 
+        clear rd
+    }
+}
+~~~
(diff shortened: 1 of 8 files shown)


plugin/header/README.md
```diff
@@ -0,0 +1,42 @@
...
+## Syntax
+
+~~~
+header {
+    ACTION FLAGS...
+    ACTION FLAGS...
+}
+~~~
+
+* **ACTION**  defines the state for dns flags, allowed values set/clear.
```
miekg:
```suggestion
* **ACTION** defines the state for dns flags, allowed values set/clear. 
```

plugin/header/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+~~~
+header {
+    ACTION FLAGS...
+    ACTION FLAGS...
+}
+~~~
+
+* **ACTION**  defines the state for dns flags, allowed values set/clear. 
+* **FLAGS** are the dns flags that will be modified e.g `ra`, `aa` etc.
```
miekg: should we support _all_ headers? If not, this should list the current supported set

plugin/header/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Examples
+
+Make sure recursive available `ra` flag is set in all the responses:
+
+~~~ corefile
+. {
+    header {
+        set ra
```
miekg, line 37: does clear get presedence? set ra; clear ra will clear ra?

miekg:
Nice idea! Could only find a couple of nits, basically ready to merge as-is

Thanks!

plugin/header/README.md
mqasimsarfraz: Yes, that is correct. I have added a note for that!

mqasimsarfraz (commented):
Thanks! :) I have updated PR to address all the nits. Please let me know if I can improve anything else as well!

miekg:
Good. then you can add yourself (and me?) in the CODEOWNERS file for this new plugin and I'll invite to the repo.

mqasimsarfraz:
Sounds good. I updated the PR.

miekg (approved):

miekg:
see few things failing, but not too bad, I'll send a small follow up PR to fix the remaining bits

mqasimsarfraz:
Sounds great. Thanks!

Issue #4309, opened 2020-11-24
[Cache] Add support for serving stale data according to RFC 8767


irlevesque:
I’m looking at implementing support for [RFC 8767](https://tools.ietf.org/html/rfc8767) (Serving Stale Data to Improve DNS Resiliency) in the Cache plugin. I’m trying to figure out an elegant solution to the RFC’s example method:

> Outside the period of the failure recheck timer, the resolver should start the query resolution timer and begin the iterative resolution process. This timer bounds the work done by the resolver when contacting external authorities and is commonly around 10 to 30 seconds. If this timer expires on an attempted lookup that is still being processed, the resolution effort is abandoned.
> 
> If the answer has not been completely determined by the time the client response timer has elapsed, the resolver should then check its cache to see whether there is expired data that would satisfy the request. If so, it adds that data to the response message with a TTL greater than 0 (as specified in Section 4). The response is then sent to the client while the resolver continues its attempt to refresh the data.

This behavior is of particular interest to us: it is preferable to request a response from the authoritative source before serving expired cache.

Currently, the Cache plugin either fields the request from cache and returns the result or calls the next plugin down the chain for a non-cached response. When `serve_stale` is set, it returns stale cached results before requesting an update. I’m thinking we’ll need to modify the `serve_stale` behavior to start a timer, call the next plugin and wait for a response. If a response was returned to the client, it should cache the response and exit. If the timer expires, however, it should return any stale cache bounded by its duration.

I propose modifying the Cache plugin's syntax accordingly:

* `serve_stale`, when serve_stale is set, cache will serve an expired entry to a client when backend resolution is
  unavailable. **DURATION** is how far back to consider stale responses as fresh. The default duration is 1h.
* `serve_stale_response_ttl` is the TTL to return with stale responses when using `serve_stale`. When `serve_stale_response_ttl` is unspecified, **DURATION** defaults to 30s.
* `serve_stale_backend_timeout` is a timer for cache's `serve_stale` to attempt resolution from the backend before serving stale responses. When `serve_stale_backend_timeout` is unspecified, **DURATION** defaults to 30s.

I'm working on a PR for this but would appreciate feedback from the team if this approach sounds sane.

chrisohaver:
> This behavior is of particular interest to us: it is preferable to request a response from the authoritative source before serving expired cache.

There are pros and cons to checking _before_ serving stale cache vs _after_ (as implemented currently).
[...]

Issue #4199, opened 2020-10-14
plugin/cache: serve_stale enhancements


miekg:
A recent discussion let me look into serve_stale in cache and it you must specify a duration which you need to figure out manually. Btw the discussion was how fast a k8s cluster breaks when the control plane goes down, coredns was fingered as one of the things the makes it go down more quickly. 

Another approach is when serve_stale is active to check the response from whatever the upstream is and if it's not an answer (like the thing we cached before, or potentially a new reply all together), we reply with the old one. This mimics "last-known-good-behavior" which is a good strategy for surviving outages. It's also more dynamic, doesn't require a duration option and may be actually be "on" by default.

bharath-123:
k8s 1.18 has made node local dns stable.[nodelocaldns](https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/)  it seems like its coredns on every node now. If the coredns deployment goes down for some reason, every node will have a node local dns cache which can serve them.

I feel that ideally, the node local coredns should not share configmaps with the cluster wide coredns deployment. So hopefully it does not go down due to a bad config in the main coredns. 

But with this i think that, if a node with the main coredns pods goes down things should still be fine because the node local dns cache can serve requests __if__ they are still in the cache.

What do you think?

chrisohaver:
> I feel that ideally, the node local coredns should not share configmaps with the cluster wide coredns deployment.

They don't.  They have their own config map, different than the cluster dns deployment.

> What do you think?

I think the current serve stale implementation does what you are hinting at.  To continue to serve stale cache entries beyond their original TTL when the upstream (cluster DNS deployment), is down.

chrisohaver:
#5131 implements what I think is described in this issue.

i.e. it checks upstream first, if no answer then reply with old (stale) cached entry.

Issue #4097, opened 2020-08-31
modify behavior for serve_stale 


ltagliamonte-dd:
As the documentation states for serve_stale:
```
cache will attempt to refresh the cache entry AFTER sending the expired cache entry to the client.
```

When serve_stale is in use the very first reply you receive from coreDNS is a stale reply and only the second one will be an update one.

In my case our main programming language is JVM based, in our apps we set a DNS cache ttl of 10s but if you use a serve_stale of 24hrs for example and you deploy everyday is it possible that a new container ends up on a node with a old and stale DNS entry and that container will get old IPs and it will be basically down for 10s (in our case) until it get the right IPs.

To better support use cases like mine i'd like to propose to change the `serve_stale` behavior and  fetch first from the upstream and in case of failure serve a stale entry if available.

chrisohaver:
FWIW, CoreDNS implements the same behavior as Unbound's `serve-expired` option.  So there is a precedent for the current behavior.

Doing the check upstream before sending an answer to the client would impact performance greatly in the case that the upstream is down.

ltagliamonte-dd:
@chrisohaver thank you for the reply could it be possible to add a flag that will change the order?
I'm my mind I would love to implement something like a circuit breaker... if the upstream fails N times start serving from the stale cache and start a corutine that with a backoff try to fetch the entry and when it succeed stop serving from the stale_cache.

chrisohaver:
Sounds like a reasonable approach, but I'm not sure if we want to maintain two significantly different implementations of more or less the same feature in the cache plugin.  @miekg ?

ltagliamonte-dd:
maybe a circuit breaker plugin?
To be honest I don't know so good the code and the complexity behind modifying the existing cache plugin or adding a new plugin.

miekg:
[ Quoting <notifications@github.com> in "Re: [coredns/coredns] modify behavi..." ]
>Sounds like a reasonable approach, but I'm not sure if we want to maintain two
>significantly different implementations of more or less the same feature in the
>cache plugin. @miekg ?

serve_stale is a mechanism that might save you a bit of time; it goes totally against the
spirit of DNS, so it's bound to be sub optimal in cases.

Making things more complex to have (slightly less?) sub optimal behavior in not something
we should do here.

/Miek

--
Miek Gieben

zmedico:
[...]

PR #4756, opened 2021-07-15
plugin/cache: Add refresh mode setting to serve_stale

Towerthousand:

This PR adds an optional `REFRESH_MODE` parameter on the `serve_stale` configuration directive of the cache plugin. For context, please see #4309

Reasoning and RFC discussed at #4309
Other similar issues are #4097 and #4199.

I updated the cache plugin `README.md` to document the new parameter on the PR.

No.

### DIFF
--- a/plugin/cache/handler.go
+++ b/plugin/cache/handler.go
@@ -38,9 +38,13 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 	if i != nil {
 		ttl = i.ttl(now)
 	}
-	if i == nil {
+	if i == nil || (ttl < 0 && c.staleFetchBefore) {
+		// Try to resolve upstream if we have no cached record or the cached one is stale and staleFetchBefore is set to true.
 		crr := &ResponseWriter{ResponseWriter: w, Cache: c, state: state, server: server, do: do}
-		return c.doRefresh(ctx, state, crr)
+		rcode, err := c.doRefresh(ctx, state, crr)
+		if i == nil || rcode == dns.RcodeSuccess {
+			return rcode, err
+		}
 	}
 	if ttl < 0 {
 		servedStale.WithLabelValues(server).Inc()
--- a/plugin/cache/README.md
+++ b/plugin/cache/README.md
@@ -37,7 +37,7 @@ cache [TTL] [ZONES...] {
     success CAPACITY [TTL] [MINTTL]
     denial CAPACITY [TTL] [MINTTL]
     prefetch AMOUNT [[DURATION] [PERCENTAGE%]]
-    serve_stale [DURATION]
+    serve_stale [DURATION] [REFRESH_MODE]
 }
 ~~~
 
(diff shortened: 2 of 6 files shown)


chrisohaver:
In retrospect, "before" and "after" could be a bit esoteric to a user. From a user perspective, the option switches between two behaviors:
1. serve stale cache immediately when a request comes in, and then attempt to refresh the cache asynchronously.
2. only serve stale cache after verifying that the upstream cannot answer the incoming request.

Naming the options "immediate" and "verified" might align closer to why someone might want to use each option.

plugin/cache/handler.go
```diff
@@ -38,9 +38,13 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 	if i != nil {
 		ttl = i.ttl(now)
 	}
-	if i == nil {
+	if i == nil || (ttl < 0 && c.staleFetchBefore) {
+		// Try to resolve upstream if we have no cached record or the cached one is stale and staleFetchBefore is set to true.
 		crr := &ResponseWriter{ResponseWriter: w, Cache: c, state: state, server: server, do: do}
-		return c.doRefresh(ctx, state, crr)
+		rcode, err := c.doRefresh(ctx, state, crr)
+		if i == nil || rcode == dns.RcodeSuccess {
```
chrisohaver, line 45: Does this intend to ignore any responses that are not "success" and serve stale cache _instead_?
I don't think we should do that.  For example, NXDOMAIN is a valid response, it means the record is currently (and authoritatively) non-existent and that should be returned to the client.  Also, if the response from "upstream" _was_ NXDOMAIN, then this would end up sending two separate responses to the client, the NXDOMAIN response, followed by the stale response (which would probably be ignored by the client).

I think the only reason to serve stale cache in the always-check-before-sending-stale mode, is if the upstream server does not respond (i.e. we don't know what the state of the upstream record is).  You might need to wrap the response writer with one that remembers/exposes if a message was written to the client.

chrisohaver: The relevant section of the RFC... https://datatracker.ietf.org/doc/html/rfc8767#section-4

>Answers from authoritative servers that have a DNS response code of
   either 0 (NoError) or 3 (NXDomain) and the Authoritative Answer (AA)
   bit set MUST be considered to have refreshed the data at the
   resolver.  Answers from authoritative servers that have any other
   response code SHOULD be considered a failure to refresh the data and
   therefore leave any previous state intact.  See Section 6 for a
   discussion.

chrisohaver: IOW - per the RFC - NOERROR and NXDOMAIN responses MUST get sent to the client, and the stale cache is replaced. For other response codes, e.g. SERVFAIL/REFUSED, the response SHOULD be ignored if stale cache is present, and stale cache sent to client. Section 6 of the RFC discusses rationale for treating various responses codes differently (in effect - a case for making the behavior configurable for response codes other than NOERROR/NXDOMAIN).

chrisohaver:
@Towerthousand, do you plan to continue with this PR?

Towerthousand:
@chrisohaver hey, sorry for not responding before, just noticed the PR got closed. We (Datadog) will pick this up again soon, we implemented a workaround on our side but we still wish to push it forward. I think it won't be me though, someone else will open a PR soon.

PR #4757, opened 2021-07-16
plugin/forward : add proxy address as tag

Tantalor93:

Hello, in order to improve the observability of CoreDNS, I would like to know to which upstream DNS nameserver the DNS request was forwarded to, so I can better analyze the issues from opentracing compatible toolings

This PR adds to each `connect` span of forward plugin (there can be multiple `connect` spans, depending on [policy](https://coredns.io/plugins/forward/)) a `peer.address` tag  as it is described in [opentracing conventions ](https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table) with an proxy address.

example of usage with CoreDNS with Corefile
```
.:53 {
    forward . 8.8.8.8
    trace zipkin 127.0.0.1:9411 {
        service dns-service
    }
    metadata
}
``` 

will produce trace which can be viewed in Zipkin like this
image

no issue related

I think no changes are needed?

no

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/miekg/dns"
 	ot "github.com/opentracing/opentracing-go"
+	otext "github.com/opentracing/opentracing-go/ext"
 )
 
 var log = clog.NewWithPlugin("forward")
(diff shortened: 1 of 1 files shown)


miekg:
/lgtm

Unsure if these trace tags need documenting. When do they show up? When you just look at a trace?

Tantalor93:
@miekg yea, when you visualize the trace in tools like Zipkin, Splunk APM, etc. There you can do some basic analytics and aggregations using these tags

PR #4767, opened 2021-07-22
plugin/kubernetes: Improve namespace usage

mlowery:

@yingnanzhang666 found that, in our clusters with well over 10K namespaces, that `namespaceExposed` was both CPU and memory intensive. This PR does the following:

* Use `GetByKey` instead of `List` in `GetNamespaceByName` (which is called by `namespaceExposed`). This avoids a slice allocation and iterating over it looking for a match.
* Add `ToNamespace` to reduce memory for namespace cache. While namespace specs are pretty small, our namespaces have plenty of metadata which is irrelevant to CoreDNS and can be thrown away, as is done with Endpoints, etc.

None.

None.

No.

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -39,7 +39,7 @@ type dnsController interface {
 	EpIndexReverse(string) []*object.Endpoints
 
 	GetNodeByName(context.Context, string) (*api.Node, error)
-	GetNamespaceByName(string) (*api.Namespace, error)
+	GetNamespaceByName(string) (*object.Namespace, error)
 
 	Run()
 	HasSynced() bool
@@ -150,14 +150,16 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
 		dns.epLock.Unlock()
 	}
 
-	dns.nsLister, dns.nsController = cache.NewInformer(
+	dns.nsLister, dns.nsController = object.NewIndexerInformer(
 		&cache.ListWatch{
 			ListFunc:  namespaceListFunc(ctx, dns.client, dns.namespaceSelector),
 			WatchFunc: namespaceWatchFunc(ctx, dns.client, dns.namespaceSelector),
 		},
 		&api.Namespace{},
-		defaultResyncPeriod,
-		cache.ResourceEventHandlerFuncs{})
+		cache.ResourceEventHandlerFuncs{},
+		cache.Indexers{},
+		object.DefaultProcessor(object.ToNamespace, nil),
+	)
 
 	return &dns
 }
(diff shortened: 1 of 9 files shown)


plugin/kubernetes/controller.go
```diff
@@ -540,17 +542,18 @@ func (dns *dnsControl) GetNodeByName(ctx context.Context, name string) (*api.Nod
...
-			return ns, nil
-		}
+	o, exists, err := dns.nsLister.GetByKey(name)
+	if err != nil {
+		return nil, err
+	}
+	if !exists {
+		return nil, fmt.Errorf("namespace not found")
+	}
+	ns, ok := o.(*api.Namespace)
```
chrisohaver, line 552:
```suggestion
	ns, ok := o.(*object.Namespace)
```

chrisohaver: Since the lister stores `object.Namespace` ... 

```suggestion
	ns, ok := o.(*object.Namespace)
```

Function signature also needs to change to return `*object.Namespace`, and all callers of `GetNamespaceByName`, unit tests etc.

chrisohaver (changes requested):

plugin/kubernetes/controller.go
mlowery: Done. Thanks!

mlowery:
This PR is ready for another look. All tests pass now.

chrisohaver (commented):
Thanks,  LGTM now.

Have you manually tested this on a Kubernetes cluster?

mlowery:
> Have you manually tested this on a Kubernetes cluster?

Yes. Tested with `namespaces` and `namespace_labels` options for `kubernetes` plugin.

As far as performance impact goes, I applied this patch to CoreDNS 1.7.0 and ran dnsperf against an instance of CoreDNS with `cache` plugin disabled. Number of namespaces in cluster was about 15K. dnsperf query file contained a single cluster-local query for a namespace that did not exist (to force it to walk the entire namespace slice for each request). 

CoreDNS instance was BestEffort. Before and after looked like this:

| Metric | Before | After |
| --- | --: | --: |
| `kubectl top` cores (captured in middle of test) | 56097m | 8594m |
| `kubectl top` bytes (captured in middle of test) | 265Mi | 104Mi |
| dnsperf queries per second | 21286 | 86959 |
| pprof profile | most time spent in `runtime.mapiternext` <sup>1</sup> | not in web view |
| pprof heap alloc_space | most from `cache.List` | relative size reduced greatly <sup>2</sup> |
| pprof heap inuse_space | most from `cache.List` | not in web view |

<sup>1</sup> called from `cache.List` which is called from `kubernetes.GetNamespaceByName`
<sup>2</sup> `kubernetes.GetNamespaceByName` still shows but allocations are from `fmt.Errorf` now

PR #4778, opened 2021-08-04
plugins/kubernetes: Switch to klog/v2

mrueg:
I believe coredns' kubernetes plugin should use klog/v2 since this is what the kubernetes libraries actually use. As per https://github.com/kubernetes/klog#how-to-use-klog I added a missing initialization call for flags as well.
None
None
No.


chrisohaver (approved):
Thanks, LGTM

PR #4779, opened 2021-08-04
Use docker buildx for release image

saschagrunert:
This allows to choose the correct architecture in the image manifest,
which defaulted to the host system before applying this patch.

Refers to https://github.com/kubernetes/kubernetes/issues/104085
None
Fixed architecture within manifest for non `amd64` images.


dims:
cc @johnbelamaric

johnbelamaric:
/approve
/merge

@chrisohaver @miekg 
@saschagrunert I assume this means our published manifests are wrong today?

dims:
@johnbelamaric unfortunately some of the arch(es) are wrong, https://github.com/kubernetes/kubernetes/issues/104085 has some details.

saschagrunert:
Yes the issue should mainly affect the manifests for the non amd64 bit  images, while the binaries within the images are correct.

miekg:
fyi: this change has been reverted because it broke the docker build.

if something like this is still needed, it cannot assume binfmt kernel module is loaded and it should not run a privileged docker container

PR #4780, opened 2021-08-04
chore: use multi-stage to build docker image

guilhem:

Simplify build process by only needing a `docker build .`

Building image from outside binary need a `--build-arg=BUILD_TYPE=outside`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ buildCoreDNSImage: &buildCoreDNSImage
       command: |
         cd ~/go/src/${CIRCLE_PROJECT_USERNAME}/coredns
         make coredns SYSTEM="GOOS=linux" && \
-        docker build -t coredns . && \
+        docker build -t coredns --build-arg=BUILD_TYPE=outside . && \
         kind load docker-image coredns
 
 jobs:
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,10 @@
-*
-!coredns
+# Docker
+Dockerfile
+.dockerignore
+
+# don't send git context
+#.git (GITCOMMIT isn't a ARG for the moment)
+
+# Don't send configuration changes
+*.md
+*.yaml
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,20 @@
-FROM debian:stable-slim
+ARG BUILD_TYPE=multi-stage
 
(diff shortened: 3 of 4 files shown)


chrisohaver:
Can you give a bit more explanation and context to this change?

guilhem:
> Can you give a bit more explanation and context to this change?

Of course.
Current build is made in 2 steps:
- build binary "locally"
- Add binary in an empty image

But this process make it impossible to use a Docker native builder (like ACR build, Cloud Build, etc.).

This PR make Dockerfile to be a "native" builder by default: `docker build .` and you have a working container image.

To be compatible with a process where we want the same binary in release and image, I kept the "outside" build process with a `docker build --build-arg`.

yongtang:
@guilhem I think one concern could be that the binary within the docker image will be different from the release binary as you are re-building the binary within the container?

guilhem:
> @guilhem I think one concern could be that the binary within the docker image will be different from the release binary as you are re-building the binary within the container?

Yes, that's why, for deployment purpose, I added (and configured in CI) a `--build-arg=BUILD_TYPE=outside` which use binary previously built.
The released container image will be the same as previously.

yongtang:
@guilhem I haven't used Code Build so I am not sure. But if you want to just build everything from within container, I think you can essentially run:
```
docker run --rm -i -t -v $PWD:/v -w /v golang:1.16 make
```

And that will give you the binary you need?

Not sure a multi-stage Dockerfile is needed?

chrisohaver:
IIUC, this change intends to make building a _containerized_ version CoreDNS easier, and compatible with other tools.

`docker run --rm -i -t -v $PWD:/v -w /v golang:1.16 make` will produce a binary of coredns, not a containerized coredns.

yongtang:
I think the concern is, the containerized CoreDNS image that can be generated by the new method will not be used to releasing CoreDNS in this repo. So it is just to serve individual usage. Maybe this can be applied locally with Azure ACR/AWS code pipeline than making it to coredns repo?

SuperQ:
+1, I don't see a good reason to change this.

guilhem:
This change absolutely doesn't change it.
I kept the way you was doing and want to do.
This PR _only_ add a way for a full container experience building.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] chore: use mu..." ]
>+1, I don't see a good reason to change this.

agreed

PR #4781, opened 2021-08-05
Update cache metrics

SuperQ:

Add a total cache request counter to follow Prometheus conventions[0].
Mark the existing cache miss metric as deprecated.

> Similarly, with hit or miss for caches, it’s better to have one
> metric for total and another for hits.

[0]: https://prometheus.io/docs/instrumenting/writing_exporters/#naming

Updated plugin README.

Backwards compatible, but deprecates the cache miss metric.


f0o:
@miekg this

PR #4792, opened 2021-08-09
Bump actions use go 1.16

chrisohaver:

Per recent action logs, looks like go 1.15 is being installed when the go version is not specified in action/setup-go.


rajansandeep (approved):

SuperQ (approved):
LGTM

PR #4793, opened 2021-08-09
plugin/forward: CNAME resolution support

tmeckel:

**Description**
The PR contains upstream i.e CNAME resolution support for the `forward` plugin. With upstream support enabled the forward plugin will try to resolve CNAMEs and only return the resulting A or AAAA address to the client.

**Use case**

Azure provides so called private endpoints for PaaS services, which will bind those services to IP addresses of a virtual (private) network in Azure. Along with this support, additional CNAME records are then created in Azure's Public DNS for the affected service instances. Clients that are then bound to appropriately configured DNS servers then receive the IP address of the endpoint from the virtual network and not the public address of the Azure service. This strategy creates additional traffic between the clients and the involved DNS Servers.

image

Using coredns and the forward plugin with upstream (CNAME resolution) support enabled will resolve the CNAME chain inside the dns server and thus reducing the traffic generated by the involved clients. With caching enabled this will speedup the resolution of addresses of private endpoints for clients as well.

[Azure Private Endpoint DNS configuration](https://docs.microsoft.com/en-us/azure/private-link/private-endpoint-dns)
[Virtual network and on-premises workloads using a DNS forwarder](https://docs.microsoft.com/en-us/azure/private-link/private-endpoint-dns#virtual-network-and-on-premises-workloads-using-a-dns-forwarder)

None

Documentation for the upstream support for the `forward` plugin has already been added

No

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -52,6 +52,7 @@ forward FROM TO... {
     policy random|round_robin|sequential
     health_check DURATION [no_rec]
     max_concurrent MAX
+    upstream [max_depth MAX]
 }
 ~~~
 
@@ -93,6 +94,11 @@ forward FROM TO... {
   response does not count as a health failure. When choosing a value for **MAX**, pick a number
   at least greater than the expected *upstream query rate* * *latency* of the upstream servers.
   As an upper bound for **MAX**, consider that each concurrent query will use about 2kb of memory.
+* `upstream` If upstream support is enabled the plugin will try to resolve CNAMEs and only return the resulting A or AAAA address.
+  If no A or AAAA record can be resolved the original (first) answer from a forwarder will be returned to the client. Circular dependencies are
+  detected and an error will be logged accordingly. In that case the original (first) answer from a forwarder will be returned to the client as well.
+  * `max_depth` **MAX** to limit the maximum calls to resolve a CNAME chain to the final A or AAAA record, a value `> 0` can be specified.
+    If the maximum depth is reached and no A or AAAA record could be found, the the original (first) from a forwarder answer will be returned to the client.
 
 Also note the TLS config is "global" for the whole forwarding proxy if you need a different
 `tls-name` for different upstreams you're out of luck.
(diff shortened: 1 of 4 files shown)


chrisohaver:
Before reviewing in detail, I think this could (and probably should) be done as a separate plugin rather than extending the forward plugin.

tmeckel:
@chrisohaver If the plugin will be one of the plugins that are available in the docker file and thus you're not required to compile yourself an own version of `coredns` I'm just fine with this 👍🏼 Any thought how the plugin could be named?

chrisohaver:
> @chrisohaver If the plugin will be one of the plugins that are available in the docker file and thus you're not required to compile yourself an own version of `coredns` I'm just fine with this 👍🏼 Any thought how the plugin could be named?

Yes, IMO as an internal plugin (i.e. included in the release). How about  _resolve_ as the name? It could later be extended to handle SRV additional record processing.  All just my opinion.  I'm not a _forward_ plugin owner.

tmeckel:
@chrisohaver Resolve sounds a bit to general to me 😀 the name should indicate what's really happening. So I'd tend to name it something like `resolve_cnames` but this would collide with the idea of you to handle SRV records as well.  So the question arises for me why SRV Records should be treated the same way. I mean, right now the code checks if the first answer is a CNAME and resolve this CNAME. Why should I process SRV records the same way? 🤔

chrisohaver:
> Resolve sounds a bit to general 

Maybe _resolve_targets_ - although, I'm not sure if targets is really a DNS term  (e.g. CNAME targets and SRV targets).

> So the question arises for me why SRV Records should be treated the same way. 

It's the same type of situation but for SRVs.  An SRV response that does not include target records in the additional section would prompt the client to look up the records it needs to use (extra lookups from the client).

> I mean, right now the code checks if the first answer is a CNAME and resolve this CNAME. Why should I process SRV records the same way?

By "It could _later_ be extended to handle SRV additional record processing" I mean "later" as "future"; functionality that could be added in the future.  Not necessarily in this first PR.

But all that said - we don't want to get jack-knifed over a name at this point, it can be easily changed during review if necessary.

miekg:
forward *forwards* it doesn't do anything else,. Any logic that cares about the type being resolved doesn't belong here; that's the job of a recursive resolver. 

If I find the piece of mind and time I might actually try to implement such a beast.

PR #4795, opened 2021-08-10
pkg/reuseport: Remove pre-go1.11 impl

mrueg:
As go <1.11 is pretty ancient, I believe it is safe to remove this code (unless I'm missing anything specific like windows builds or similar)
Drops support for go <1.11 (which I believe is the case anyway in other parts of the project due to go modules).


chrisohaver (approved):
Seems reasonably safe to me.

yongtang (approved):

chrisohaver:
This appears to have broken windows builds ...

Issue #4808, opened 2021-08-16
plugin/dns64 : support also intercepting DNS requests coming over IPv4 network


Tantalor93:
Hello, it seems we hit an issue related to this comment in `dns64` plugin https://github.com/coredns/coredns/blob/master/plugin/dns64/dns64.go#L78.
**What would you like to be added**:
We would like to add configuration to `dns64` plugin to enable intercepting DNS requests coming also over IPv4 network

**Why is this needed**:
our CoreDNS instance is sitting in IPv4 network and DNS requests are being forwarded to it from gateway in the same IPv4 network and we need to apply DNS64 to these forwarded requests, but CoreDNS does not apply DNS64 it due to them arriving from IPv4 address

our setup is something like this:
```
| IPv6-only| ----> |     IPv4 network     |
|  device  | ----> | gateway ----> CoreDNS|
```

SuperQ:
For the record, this is likely to cause problems if non-DNS64 devices are also making requests through the same server.

PR #4809, opened 2021-08-16
plugin/dns64 : add support for DNS requests over IPv4 network

Tantalor93:

This PR adds a configuration option to `dns64` plugin which enables intercepting also DNS requests from IPv4 network

https://github.com/coredns/coredns/issues/4808

updated README

no, if the option is not provided, it fallback to false and previous behaviour

### DIFF
--- a/plugin/dns64/README.md
+++ b/plugin/dns64/README.md
@@ -27,11 +27,13 @@ Or use this slightly longer form with more options:
 dns64 [PREFIX] {
     [translate_all]
     prefix PREFIX
+    [allow_ipv4]
 }
 ~~~
 
 * `prefix` specifies any local IPv6 prefix to use, instead of the well known prefix (64:ff9b::/96)
 * `translate_all` translates all queries, including responses that have AAAA results.
+* `allow_ipv4` allow DNS queries, ignoring whether they came from over IPv4 or IPv6 network
 
 ## Examples
 
@@ -66,6 +68,15 @@ Enable translation even if an existing AAAA record is present.
 . {
     dns64 {
         translate_all
+    }
+}
+~~~
+
+Apply translation even to the requests which arrived over IPv4 network.
+~~~ corefile
+. {
+    dns64 {
+        allow_ipv4
     }
 }
 ~~~
--- a/plugin/dns64/README.md
+++ b/plugin/dns64/README.md
@@ -72,7 +72,8 @@ Enable translation even if an existing AAAA record is present.
 }
 ~~~
 
-Apply translation even to the requests which arrived over IPv4 network.
+Apply translation even to the requests which arrived over IPv4 network. Beware that enabling this options may break dual-stack clients,
+by doing DNS64 translations for empty AAAA records (no IPv6 for the domain) and clients might prefer using synthetic records instead of using A records.
 ~~~ corefile
 . {
     dns64 {
--- a/plugin/dns64/README.md
+++ b/plugin/dns64/README.md
@@ -72,8 +72,8 @@ Enable translation even if an existing AAAA record is present.
 }
 ~~~
 
-Apply translation even to the requests which arrived over IPv4 network. Beware that enabling this option for dual-stack
-clients might cause problems.
+Apply translation even to the requests which arrived over IPv4 network. Beware that enabling this option in CoreDNS handling combination 
+of dual-stack and IPv6 clients might cause problems by also doing DNS64 translations for requests of dual-stack clients.
 ~~~ corefile
 . {
     dns64 {
--- a/plugin/dns64/README.md
+++ b/plugin/dns64/README.md
@@ -33,7 +33,7 @@ dns64 [PREFIX] {
 
 * `prefix` specifies any local IPv6 prefix to use, instead of the well known prefix (64:ff9b::/96)
 * `translate_all` translates all queries, including responses that have AAAA results.
-* `allow_ipv4` allow DNS queries, ignoring whether they came from over IPv4 or IPv6 network
+* `allow_ipv4` Allow translating queries if they come in over IPv4, default is IPv6 only translation.
 
 ## Examples
 
--- a/plugin/dns64/README.md
+++ b/plugin/dns64/README.md
@@ -72,8 +72,8 @@ Enable translation even if an existing AAAA record is present.
 }
 ~~~
 
-Apply translation even to the requests which arrived over IPv4 network. Beware that enabling this options may break dual-stack clients,
-by doing DNS64 translations for empty AAAA records (no IPv6 for the domain) and clients might prefer using synthetic records instead of using A records.
+Apply translation even to the requests which arrived over IPv4 network. Beware that enabling this option for dual-stack
+clients might cause problems.
 ~~~ corefile
 . {
     dns64 {
--- a/plugin/dns64/README.md
+++ b/plugin/dns64/README.md
@@ -72,8 +72,11 @@ Enable translation even if an existing AAAA record is present.
 }
 ~~~
 
-Apply translation even to the requests which arrived over IPv4 network. Beware that enabling this option in CoreDNS handling combination 
-of dual-stack and IPv6 clients might cause problems by also doing DNS64 translations for requests of dual-stack clients.
+Apply translation even to the requests which arrived over IPv4 network. Warning, the `allow_ipv4` feature will apply 
+translations to requests coming from dual-stack clients. This means that a request for a client that sends an `AAAA`
+that would normal result in an `NXDOMAIN` would get a translated result. 
+This may cause unwanted IPv6 dns64 traffic when a dualstack client would normally use the result of an `A` record request.
+
 ~~~ corefile
 . {
     dns64 {
--- a/plugin/dns64/dns64.go
+++ b/plugin/dns64/dns64.go
@@ -70,7 +70,7 @@ func (d *DNS64) Name() string { return "dns64" }
 
 // requestShouldIntercept returns true if the request represents one that is eligible
 // for DNS64 rewriting:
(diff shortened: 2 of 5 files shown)


plugin/dns64/dns64.go
```diff
@@ -69,13 +70,11 @@ func (d *DNS64) Name() string { return "dns64" }
 
 // requestShouldIntercept returns true if the request represents one that is eligible
 // for DNS64 rewriting:
-// 1. The request came in over IPv6 (not in RFC)
+// 1. The request came in over IPv6, unless AllowIPv4 option is not set
```
SuperQ: Clearer would be `or the 'allow_ipv4' option is set`. Avoids double negative wording.

plugin/dns64/README.md
```diff
@@ -27,11 +27,13 @@ Or use this slightly longer form with more options:
 dns64 [PREFIX] {
     [translate_all]
     prefix PREFIX
+    [allow_ipv4]
 }
 ~~~
 
 * `prefix` specifies any local IPv6 prefix to use, instead of the well known prefix (64:ff9b::/96)
 * `translate_all` translates all queries, including responses that have AAAA results.
+* `allow_ipv4` allow DNS queries, ignoring whether they came from over IPv4 or IPv6 network
```
SuperQ: Similar here, let's simplify the wording.

```suggestion
* `allow_ipv4` Allow translating queries if they come in over IPv4, default is IPv6 only translation.
```

plugin/dns64/dns64.go
Tantalor93: right 👍 fixed as you suggested

plugin/dns64/README.md
Tantalor93: yeah, that sounds much better 👍  fixed as you suggested

chrisohaver (approved):

chrisohaver (dismissed):
I kind of feel like this should be the default behavior.  But, makes sense to leave the default as-is for backward compatibility.

SuperQ:
@chrisohaver This being the default behavior is annoying and dangerous for a couple of reasons:

Say you have two networks behind a CoreDNS server, one IPv6-only and one Dual Stack.

The CoreDNS server sits in a dual-stack subnet so it can handle DNS requests from both networks.

If you enable this feature, the `A` requests from the dual stack network will get translated to IPv6/DNS64 rather than returned as normal.

IMO, this feature doesn't make any sense since for a IPv6-only network, you will need a IPv6-native DNS server anyway.

plugin/dns64/dns64.go
```diff
@@ -69,13 +70,11 @@ func (d *DNS64) Name() string { return "dns64" }
...
-// 1. The request came in over IPv6 (not in RFC)
+// 1. The request came in over IPv6 or the 'allow_ipv4' option is set
 // 2. The request is of type AAAA
 // 3. The request is of class INET
-func requestShouldIntercept(req *request.Request) bool {
-	// Only intercept with this when the request came in over IPv6. This is not mentioned in the RFC.
-	// File an issue if you think we should translate even requests made using IPv4, or have a configuration flag
-	if req.Family() == 1 { // If it came in over v4, don't do anything.
+func (d *DNS64) requestShouldIntercept(req *request.Request) bool {
+	if !d.AllowIPv4 && req.Family() == 1 { // If it came in over v4, don't do anything.
```
SuperQ: I think the comment here needs updating. Maybe remove it and put a better explanation in for the old comment.

plugin/dns64/README.md
```diff
@@ -70,6 +72,15 @@ Enable translation even if an existing AAAA record is present.
 }
 ~~~
 
+Apply translation even to the requests which arrived over IPv4 network.
```
SuperQ: Can we add a note that this may break `A` record lookup for users coming from dual-stack networks?

SuperQ (changes requested):
This needs some warnings about dual-stack users getting DNS64 translations for `A` record lookups.

chrisohaver:
> If you enable this feature, the `A` requests from the dual stack network will get translated to IPv6/DNS64 rather than returned as normal.

This plugin does not act on `A` requests.  It only acts on `AAAA` requests.

chrisohaver:
> This needs some warnings about dual-stack users getting DNS64 translations for A record lookups.

Ah, I think you mean to say that Dual Stack clients _should_ get an empty answer for queries that don't have any `AAAA` records.  But this option will cause the response to have `A`->`AAAA` translated records in the response. Gotcha.

IOW: "This needs some warnings about dual-stack users getting DNS64 translations for empty `AAAA` record lookups."

chrisohaver:
> > This needs some warnings about dual-stack users getting DNS64 translations for A record lookups.
> 
> Ah, I think you mean to say that Dual Stack clients _should_ get an empty answer for queries that don't have any `AAAA` records. But this option will cause the response to have `A`->`AAAA` translated records in the response. Gotcha.

Nevermind - that happens regardless of whether or not the setting is enabled.  It just illustrates that it's problematic to enable dns64 translation for a dual stack network at all.

plugin/dns64/README.md
Tantalor93: added a warning, hope the wording is understandable 👍

plugin/dns64/dns64.go
Tantalor93: updated

plugin/dns64/README.md
```diff
@@ -72,7 +72,8 @@ Enable translation even if an existing AAAA record is present.
 }
 ~~~
 
-Apply translation even to the requests which arrived over IPv4 network.
+Apply translation even to the requests which arrived over IPv4 network. Beware that enabling this options may break dual-stack clients,
+by doing DNS64 translations for empty AAAA records (no IPv6 for the domain) and clients might prefer using synthetic records instead of using A records.
```
chrisohaver: Wouldn't this happen even without the option enabled?

Tantalor93: ummm, right. Bad wording :D

Tantalor93: hope it makes sense now, I tried to describe the example the @SuperQ gave in the comments

plugin/dns64/README.md
```diff
@@ -70,6 +72,16 @@ Enable translation even if an existing AAAA record is present.
 }
 ~~~
 
+Apply translation even to the requests which arrived over IPv4 network. Beware that enabling this option in CoreDNS handling combination 
+of dual-stack and IPv6 clients might cause problems by also doing DNS64 translations for requests of dual-stack clients.
```
chrisohaver: ... but the plugin will do translations for dual-stack clients even if the option is not enabled. So, the behavior described is not specific to this option.  As long as a dual-stack client makes a query for a `AAAA` over IPv6 transport, this plugin will attempt a translation (when there are no `AAAA` records).

The difference In behavior would be for `AAAA` queries made over IPv4 transport.  By default, the plugin gives a different answer to the client depending on the transport version. On the surface that seems potentially problematic - different answers for same name/type depending on which IP you send the request out on.

The `allow_ipv4` option makes the answer consistent for both transports.

miekg:
is this going somewhere?

chrisohaver:
> is this going somewhere?

IIRC, IMO the documented caveat is invalid and should be removed. The code change is fine, although I think this should be the default behavior, not a new option.  I don't agree with the [arguments above](https://github.com/coredns/coredns/pull/4809#issuecomment-899657036) for not making this default behavior, or the [reasoning](https://github.com/coredns/coredns/pull/4809#pullrequestreview-730941508) for adding a warning in the documentation.

edit: I think the case for making this an option is for backward compatibility sake, in case someone is relying on the current split-horizon-like behavior.

Tantalor93:
I am kinda struggling with the right wording in the README, was hoping that @SuperQ would comment somehow as plugin owner. I am happy to make any adjustments to the PR, but dont know  what is needed from me @miekg @chrisohaver . Still I think the code changes are valid.

chrisohaver:
> Still I think the code changes are valid.

I agree. I don't think the warning is needed, but as I [mentioned](https://github.com/coredns/coredns/pull/4809#event-5164504947) above, I deferred approval to @SuperQ, the plugin owner.

plugin/dns64/README.md
SuperQ: How about this wording:

```
Warning, the `allow_ipv4` feature will apply translations to requests coming from dual-stack clients. This means that a request for a client that sends an `AAAA` that would normal result in an `NXDOMAIN` would get a translated result. This may cause unwanted IPv6 dns64 traffic when a dualstack client would normally use the result of an `A` record request.
```

@chrisohaver Does this better explain it/warn the user?

chrisohaver: fwiw, I don’t understand why it’s a warning. It seems like that is how it should behave by default. But if you are ok with it then let’s go with that.

SuperQ: It's a warning because users with functioning IPv4 will get re-routed to IPv6 unnecessarily. That's why we implemented it the current way by default. This plugin is meant for users who are IPv6-only, not dualstack. If the user can reach IPv4, they should get untranslated results.

Tantalor93: I have adjusted wording as you suggested @SuperQ, please take a look if it makes sense

chrisohaver:
@SuperQ, please approve if you are OK with this PR. Thanks!

SuperQ (approved):
LGTM, thanks for taking the time to get the warning done well. It's a useful feature, but it needs to be used with caution in the real world.

SuperQ:
Always translating should definitely not be the default behavior. The reasoning above is why we made it not do it this way in the beginning. IMO, it's a missing recommendation in the DNS64 RFCs.

chrisohaver:
Thanks @Tantalor93 and @SuperQ!

PR #4814, opened 2021-08-16
Revert "pkg/reuseport: Remove pre-go1.11 impl"

chrisohaver:
Looks like coredns/coredns#4795 caused windows builds to fail.


miekg:
Should rename the files as we did in miekg/dns to make clear this isnt a Go
version issue

PR #4816, opened 2021-08-16
Fix docker build issue

yongtang:
This PR fixes docker build issue, as it looks like
debian changed `stable/update` to `stable-security/update`.


miekg:
this change has been manually reverted, because it broke the docker build

Issue #4649, opened 2021-05-26
I use the etcd plug-in to forward all the analysis, and I encountered a CPU usage of 300%


goudai:
```txt
. {
    reload 10s
    errors
    log
    etcd {
        path /skydns/ka2
        endpoint    xx
        fallthrough
    }

    loadbalance round_robin

    # upstream dns
    forward . 10.8.8.8:53 {
      prefer_udp
    }

    cache 10
    prometheus 0.0.0.0:9123
}
```
Coredns version：1.8.3

I use the etcd plug-in to forward all the analysis, and I encountered a CPU usage of 300%

chrisohaver:
@goudai, can you provide a detailed reproducible scenario?

PR #4824, opened 2021-08-25
plugin:Reset dns question name if NS loopkup fail

### LINKED ISSUE #4544: forward causing high cpu usage
**What happened**:

when request . in NS use forward plugin，even all proxy return “2 . NS: dns: bad rdata”，it still in loop until defaultTimeout，causing high cpu usage

**Why is this needed**:

it‘s no need to request proxy when all return error，causing high cpu usage

return err after all proxy return err

DNS like this
QUESTION: 
 QNAME: ns.dns.. QTYPE: NS QCLASS: in

**Environment**:

- the version of CoreDNS: 1.6.6
- Corefile: 

```
. {
    reload 10s
    errors
    log
    etcd {
        path /skydns/ka2
        endpoint    xx
        fallthrough
    }

    loadbalance round_robin

    # upstream dns
    forward . 10.8.8.8:53 {
      prefer_udp
    }

    cache 10
    prometheus 0.0.0.0:9123
}
```

- logs, if applicable:
- OS (e.g: `cat /etc/os-release`): debian9
- Others:

changqingsun:
### **1. Why is this pull request needed and what does it do?**
When I used both etcd plugin and forward plugin, It cannot resolve the NS record of the root domain name.

cmd `dig -t NS .`
resolve time out.

At this time, I found two phenomena：

- coredns logs print "NS: dns: bad rdata".
- it still in loop until defaultTimeout，causing high cpu usage

The reason of this question is that the question section of packet which forwarded by forward plugin is` ns.dns..`  Instead of` . `. 
So the forward plugin print bad rdata Error.

Question section is setted in backend_lookup.go,  I think if NS lookup fail ,  it should reset this question section.

so I commit this PR.

### **2. Which issues (if any) are related?**
forward causing high cpu usage #4544
I use the etcd plug-in to forward all the analysis, and I encountered a CPU usage of 300% #4649 

### **3. Which documentation changes (if any) need to be made?**
No.

### **4. Does this introduce a backward incompatible change or deprecation?**
No.

I verified in our test environment that this commit  is  work.

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -422,13 +422,13 @@ func NS(ctx context.Context, b ServiceBackend, zone string, state request.Reques
 	old := state.QName()
 
 	state.Clear()
-	state.Req.Question[0].Name = dnsutil.Join("ns.dns.", zone)
+	state.Req.Question[0].Name = dnsutil.Join("ns.dns", zone)
 	services, err := b.Services(ctx, state, false, opt)
+	// ... and reset
+	state.Req.Question[0].Name = old
 	if err != nil {
 		return nil, nil, err
 	}
-	// ... and reset
-	state.Req.Question[0].Name = old
 
 	seen := map[string]bool{}
 


chrisohaver:
Hmm.. I think this change makes sense. But also ~IMO `dnsutil.Join` should be creating valid output... i.e. it should be fixed to  handle joining a root domain correctly.~

Actually - I think we're passing the "label" to `Join` incorrectly... https://github.com/coredns/coredns/pull/4824#discussion_r695784481

plugin/backend_lookup.go
```diff
@@ -425,6 +425,7 @@ func NS(ctx context.Context, b ServiceBackend, zone string, state request.Reques
 	state.Req.Question[0].Name = dnsutil.Join("ns.dns.", zone)
```
chrisohaver:
```suggestion
	state.Req.Question[0].Name = dnsutil.Join("ns.dns", zone)
```

changqingsun: I try this commit in my test env.  But i still can not work with command `"dig +t NS ."`

the cmd return:
```
dig -t NS . @localhost
;; Question section mismatch: got ns.dns/NS/IN
;; Question section mismatch: got ns.dns/NS/IN
;; Question section mismatch: got ns.dns/NS/IN
;; Question section mismatch: got ns.dns/NS/IN
```
Catch the packet 

```
request ： 	DNS	77	Standard query 0x83a0 NS ns.dns OPT
reply  	DNS	1063	Standard query response 0x83a0 No such name NS ns.dns SOA a.root-servers.net NSEC do NSEC aaa RRSIG RRSIG RRSIG OPT
```
So use question name ns.dns  can not get result from upstream server. 

But in my opinion, this commit is needed.  and  need modify function `dnsutil.Join`

```
func Join(labels ...string) string {
	ll := len(labels)
	if labels[ll-1] == "." {
		//return strings.Join(labels[:ll-1], ".") + "."
                return "."
	}
	return dns.Fqdn(strings.Join(labels, "."))
}
```

plugin/backend_lookup.go
```diff
@@ -482,6 +483,7 @@ func BackendError(ctx context.Context, b ServiceBackend, zone string, rcode int,
 	m := new(dns.Msg)
 	m.SetRcode(state.Req, rcode)
 	m.Authoritative = true
+	m.RecursionAvailable = true
```
chrisohaver: None of the backend plugins currently implement recursion - so setting this flag is not correct.

plugin/etcd/handler.go
```diff
@@ -69,6 +69,7 @@ func (e *Etcd) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (
 	m := new(dns.Msg)
 	m.SetReply(r)
 	m.Authoritative = true
+	m.RecursionAvailable = true
```
chrisohaver: The etcd plugin does not implement recursion.

plugin/backend_lookup.go
changqingsun: @chrisohaver  disscuss this issue ? 
i try  use this ? it does still not work.
```
func Join(labels ...string) string {
	ll := len(labels)
	if labels[ll-1] == "." {
		//return strings.Join(labels[:ll-1], ".") + "."
                return "."
	}
	return dns.Fqdn(strings.Join(labels, "."))
}
```

chrisohaver: > @chrisohaver disscuss this issue ?
> i try use this ? it does still not work.
> 
> ```
> func Join(labels ...string) string {
> 	ll := len(labels)
> 	if labels[ll-1] == "." {
> 		//return strings.Join(labels[:ll-1], ".") + "."
>                 return "."
> 	}
> 	return dns.Fqdn(strings.Join(labels, "."))
> }
> ```

This will ignore all other labels when last label is a ".", which is definitely not what we want to happen. Note that `dnsutil.Join` is a utility function used in many places - not just here.

changqingsun: the  commit which you suggestion  doesn't not work still ....  it can not relose my issue 
`state.Req.Question[0].Name = dnsutil.Join("ns.dns", zone)`

chrisohaver:
> the commit which you suggestion doesn't not work still ... `;; Question section mismatch: got ns.dns/NS/IN`

I suspect you tried it without your fix to restore the question - that's also needed.

plugin/backend_lookup.go
```diff
@@ -425,6 +425,7 @@ func NS(ctx context.Context, b ServiceBackend, zone string, state request.Reques
 	state.Req.Question[0].Name = dnsutil.Join("ns.dns.", zone)
 	services, err := b.Services(ctx, state, false, opt)
 	if err != nil {
+		state.Req.Question[0].Name = old
```
chrisohaver:
``` go
	services, err := b.Services(ctx, state, false, opt)
 	if err != nil {
 		return nil, nil, err
 	}
 	// ... and reset
```

shorter ... 

``` go
	services, err := b.Services(ctx, state, false, opt)
 	// ... and reset
 	if err != nil {
 		return nil, nil, err
 	}
```

changqingsun: OK, I agree with you this change. 
```
	services, err := b.Services(ctx, state, false, opt)
 	// ... and reset
 	if err != nil {
 		return nil, nil, err
 	}

```
this fix can reslove my issue. 

I modify this commit as the shorter, and remove  ra flag commit.

miekg:
the root domain is not supported in the etcd plugin and breaks stuff IIRC

chrisohaver:
If we merge, before merging, we need to be make sure we don't break K8s NS queries, which use the same backend function.
I don't recall if there is an end to end test for it.

changqingsun:
> If we merge, before merging, we need to be make sure we don't break K8s NS queries, which use the same backend function.
> I don't recall if there is an end to end test for it.

OK ,  I update this commit,  please help me to check it.   I can pass my test in my env.   but I didn't test  K8S NS queries.

I'll try my best  find K8S environment and test that.   Or is there any feasible test method on your side?

chrisohaver:
I’ll look into adding an end to end test for it.

changqingsun:
> I’ll look into adding an end to end test for it.

OK ,  Thanks a lot.

chrisohaver:
Added an NS query test to **ci/circleci: kubernetes-tests**

re-ran it in this PR ... and it passed.

chrisohaver (approved):

changqingsun:
> Added an NS query test to **ci/circleci: kubernetes-tests**
> 
> re-ran it in this PR ... and it passed.

OK,  Thank you very much for your help.

miekg:
This needs an intree test as well. Would like that to see fail unless this patch is applied

changqingsun:
> This needs an intree test as well. Would like that to see fail unless this patch is applied

What is intree test ？ How to do it.     I tested in my test env , it can not reslove root domain use etcd plugin and forward plugin  without this patch.

chrisohaver:
> > This needs an intree test as well. Would like that to see fail unless this patch is applied
> 
> What is intree test ？ How to do it. I tested in my test env , it can not reslove root domain use etcd plugin and forward plugin without this patch.

Add a unit test to this PR that tests the code in a way that it would fail if this fix is was not in place.

chrisohaver (changes requested):
Thanks.  Please add a unit test.

changqingsun:
> Thanks. Please add a unit test.

OK ， I will do it as soon as possible

changqingsun:
> Thanks. Please add a unit test.

I write a testcase in  plugin/etcd/lookup_test.go

but i run it ... show me  no tests to run .
```
MacBook etcd % ls 
README.md      cname_test.go  etcd.go        group_test.go  handler.go     log_test.go    lookup_test.go msg            multi_test.go  other_test.go  setup.go       setup_test.go  xfr.go
MacBook etcd % go test .
ok      github.com/coredns/coredns/plugin/etcd  (cached) [no tests to run]

```

Is my execution method wrong?

chrisohaver:
@changqingsun, that file has a build tag `// +build etcd`... so it doesn't get built unless you build with that tag.
It has a build tag because the test requires a local etcd server to be configured and running.

It would probably be better to test the `NS` function directly in a new file `backend_lookup_test.go`, passing a `ServiceBackend` test stub.

changqingsun:
> @changqingsun, that file has a build tag `// +build etcd`... so it doesn't get built unless you build with that tag. It has a build tag because the test requires a local etcd server to be configured and running.
> 
> It would probably be better to test the `NS` function directly in a new file `backend_lookup_test.go`, passing a `ServiceBackend` test stub.

hi, chrisohaver,   This issue only happened when use etcd and forward scenes.  only test NS,  i still use etcd servicebackend...   I don‘t konw  how to do it . can you help me ? 


input   NS .
expect result :  err   (real is err...)

chrisohaver:
> > @changqingsun, that file has a build tag `// +build etcd`... so it doesn't get built unless you build with that tag. It has a build tag because the test requires a local etcd server to be configured and running.
> > It would probably be better to test the `NS` function directly in a new file `backend_lookup_test.go`, passing a `ServiceBackend` test stub.
> 
> hi, chrisohaver, This issue only happened when use etcd and forward scenes. only test NS, i still use etcd servicebackend... I don‘t konw how to do it . can you help me ?
> 
> input NS . expect result : err (real is err...)

For the test, define an new object that minimally implements servicebackend. Then pass that object to the function.

chrisohaver:
@changqingsun, do you plan to complete this PR?

PR #4826, opened 2021-08-26
upstream lookups are done with original EDNS options

### LINKED ISSUE #4825: plugin/dns64 not respecting EDNS options in original AAAA requests
**What happened**:
Hello 👋  while using `dns64` plugin I noticed that plugin `dns64` is not respecting EDNS options in original AAAA requests, for example ECS option from the original AAAA request is not taken into consideration when doing synthesis from A record. This brings additional problems as we are encoding additional info into EDNS options and these are not passed to upstream.

I have backtracked it to this code https://github.com/coredns/coredns/blob/master/plugin/pkg/upstream/upstream.go#L32 where the original EDNS options are lost (or rather they are not copied), it seems that the same issue will be even in `kubernetes` and `etcd` plugins

EDNS options from original AAAA request be passed to upstream, see how to reproduce section

Let's have a Corefile
```
.:53 {
    dns64
    forward . 8.8.8.8
    log
}
```
when you do a DNS request with the ECS option with US subnet
```
dig @::1  aws.amazon.com +subnet=216.58.211.238/24 -t AAAA +short
tp.8e49140c2-frontier.amazon.com.
dr49lng3n1n2s.cloudfront.net.
64:ff9b::4109:5846
```

and another DNS request with ECS option with Czech subnet
```
dig @::1  aws.amazon.com +subnet=81.0.198.170/24 -t AAAA +short
tp.8e49140c2-frontier.amazon.com.
dr49lng3n1n2s.cloudfront.net.
64:ff9b::4109:5846
```

**you get the same synthesized address**, which is weird when you compare it with same requests done directly against GoogleDNS. You can see that these addresses differ and so should the synthesized AAAA records.
```
dig @8.8.8.8 aws.amazon.com +subnet=216.58.211.238/24 +short
tp.8e49140c2-frontier.amazon.com.
dr49lng3n1n2s.cloudfront.net.
13.35.123.36
```
```
dig @8.8.8.8 aws.amazon.com +subnet=81.0.198.170/24 +short
tp.8e49140c2-frontier.amazon.com.
dr49lng3n1n2s.cloudfront.net.
13.225.75.7
```

**Environment**:

- the version of CoreDNS: latest
- Corefile:
- logs, if applicable:
- Others:

Tantalor93:

copy whole original DNS request with all EDNS options to make lookups against upstream with all original EDNS0 options

https://github.com/coredns/coredns/issues/4825

I guess none

no

close #4825

### DIFF
--- a/plugin/pkg/upstream/upstream.go
+++ b/plugin/pkg/upstream/upstream.go
@@ -25,11 +25,9 @@ func (u *Upstream) Lookup(ctx context.Context, state request.Request, name strin
 		return nil, fmt.Errorf("no full server is running")
 	}
 
-	size := state.Size()
-	do := state.Do()
-	req := new(dns.Msg)
+	req := state.Req.Copy()
+	// overwrite copied ID to a new one
 	req.SetQuestion(name, typ)
-	req.SetEdns0(uint16(size), do)
 
 	nw := nonwriter.New(state.W)
 	server.ServeDNS(ctx, nw, req)
--- a/plugin/pkg/upstream/upstream.go
+++ b/plugin/pkg/upstream/upstream.go
@@ -24,13 +24,10 @@ func (u *Upstream) Lookup(ctx context.Context, state request.Request, name strin
 	if !ok {
 		return nil, fmt.Errorf("no full server is running")
 	}
-
-	req := state.Req.Copy()
-	// overwrite copied ID to a new one
-	req.SetQuestion(name, typ)
+	req := state.NewWithQuestion(name, typ)
 
 	nw := nonwriter.New(state.W)
-	server.ServeDNS(ctx, nw, req)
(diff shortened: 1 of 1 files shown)


chrisohaver:
Can we use ‘state.NewWithQuestion’ here?

Tantalor93:
@chrisohaver I guess we can 👍

chrisohaver:
> @chrisohaver I guess we can 👍

Looks like `SetQuestion` resets the message id, and sets RD bit ...

``` go
	dns.Id = Id()
	dns.RecursionDesired = true
```

... but `NewWithQuestion` retains the originals.  So - this would change things slightly...  IMO - we probably shouldn't be changing the RD from the original.  IIUC, I don't think resetting the Id is needed either.

So. LGTM I think.

chrisohaver (approved):

Issue #4398, opened 2021-01-17
CNAME need not be followed after a synthesized CNAME for a CNAME query


SivaKesava1:
Hi again,

A synthesized `CNAME` record is a perfect answer for a `CNAME` query and need not be followed. 

Consider these cases, 
|                    |           |                                                          |
|--------------------|-----------|----------------------------------------------------------|
|uni.net.        | 500  SOA  | ns1.uni.net. root.campus.edu. 3 86400 7200 604800 300 |
| *.uni.net.        | 500  CNAME   | ww.uni.net.                                         |
| ww.uni.net.    | 500 CNAME | ww.campus.edu.                                          |
| test.uni.net.    | 500 CNAME | ww.uni.net.                                          |

For the query, `< test.uni.net., CNAME>` CoreDNS returns the last `CNAME` records and **stops**. But for the query `<foo.uni.net, CNAME>` CoreDNS synthesizes a `CNAME` record from the wildcard record and then again **follows** it. The response was as follows:
```
          ";QUESTION",
          "foo.uni.net. IN CNAME",
          ";ANSWER",
          "foo.uni.net. 500 IN CNAME ww.uni.net.",
          "ww.uni.net. 500 IN CNAME ww.campus.edu.",
```

The processing should have been the same for both cases.

bkrugerp99:
Real world example?  (inside a container)

```
# nslookup dl-cdn.alpinelinux.org
Server:		10.100.0.10
Address:	10.100.0.10#53
Non-authoritative answer:
dl-cdn.alpinelinux.org.default.svc.cluster.local	canonical name = dl-cdn.alpinelinux.org.
dl-cdn.alpinelinux.org	canonical name = dualstack.d.sni.global.fastly.net.
Name:	dualstack.d.sni.global.fastly.net
Address: 151.101.54.133
** server can't find dualstack.d.sni.global.fastly.net: NXDOMAIN
```

ykhr53:
I can reproduce the behaviour.
Root cause is in the `externalLookup` function, CNAME RR (in this case, "ww.uni.net. 500 IN CNAME ww.campus.edu.") will be added to `rrs` which will be returned to a client.
https://github.com/coredns/coredns/blob/master/plugin/file/lookup.go#L333

The reason why we don't get this additional answer on `test.uni.net` is, we don't call `externalLookup` for non-wildcard RRs. Please check the differences.

The case we found entire name:
[...]

PR #4828, opened 2021-08-27
plugin/file: fix wildcard CNAME answer

ykhr53:
To fix the issue  https://github.com/coredns/coredns/issues/4398

Currently, we return different answers between wildcard CNAME and non-wildcard CNAME RRs for CNAME query.
When we receive a CNAME query for a wildcard CNAME RR, we try to chase chained names (like we do for A record query). 
For a non-wildcard CNAME RR, we don't chase chained names for a CNAME query because exceptional condition has been added in the same commit, the CNAME chain feature added. https://github.com/coredns/coredns/commit/4ef53081c536a66acf77c7b1e3df15c2fc0ca52b

To fix the inconsistent behavior, I added the same condition and some test cases.

https://github.com/coredns/coredns/issues/4398

No.

No.

### DIFF
--- a/plugin/file/lookup.go
+++ b/plugin/file/lookup.go
@@ -203,7 +203,7 @@ func (z *Zone) Lookup(ctx context.Context, state request.Request, qname string) 
 	if wildElem != nil {
 		auth := ap.ns(do)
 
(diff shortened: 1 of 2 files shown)


ykhr53:
Could anybody review the CR? thanks!

chrisohaver (approved):
Thanks! LGTM

PR #4829, opened 2021-08-27
Add ACME Plugin

chinzhiweiblank:
see [#3460](https://github.com/coredns/coredns/issues/3460)

### DIFF
--- a/plugin/acme/README.md
+++ b/plugin/acme/README.md
@@ -1,0 +1,55 @@
+# acme
+## Name
+*acme* - Performs the ACME protocol for the specified domain
+
+## Description
+The *acme* plugin is useful for automating certificate management through the `ACME`.
+
+Make sure that:
+* You own the domain
+* Your CoreDNS server is the authoritative nameserver for the domain
+
+## Syntax
+~~~ txt
+acme {
+  domain DOMAIN
+
+  # optional parameters
+  challenge <CHALLENGE> port <PORT>
+}
+~~~
+
+* `DOMAIN` is the domain name the plugin should be authoritative for.
+* By default, the **DNS** challenge will be used for ACME.
+
+You can specify one or more challenges the CA can use to verify your ownership of the domain.
+* `CHALLENGE` is the name of the challenge you will use for ACME. There are only two options: `tlsalpn` and `http01`.
+* `PORT` is the port number to use for each challenge. Make sure the ports are open and accessible.
+
+## Examples
+### Basic
+~~~ corefile
+acme {
+  domain example.org
+}
+~~~
+This will perform ACME for `example.org` and use the `DNS01` challenge only.
+
+### Advanced
+~~~ corefile
+acme {
+  domain example.com
+
+  challenge http port 90
+  challenge tlsalpn port 8080
+}
+~~~
+This will perform ACME for `example.com` and perform the following challenges:
+1. `HTTP` challenge on port **90**
+2. `TLSALPN` challenge on port **8080**
+3. `DNS` challenge
+
+## See Also
+1. [RFC for ACME](https://datatracker.ietf.org/doc/html/rfc8555/)
+2. [ACME Protocol](https://www.thesslstore.com/blog/acme-protocol-what-it-is-and-how-it-works/)
+3. [Challenge Types](https://letsencrypt.org/docs/challenge-types/)
--- a/plugin/acme/README.md
+++ b/plugin/acme/README.md
@@ -12,7 +12,7 @@ Make sure that:
 ## Syntax
 ~~~ txt
 acme {
-  domain DOMAIN
+  domain <DOMAIN>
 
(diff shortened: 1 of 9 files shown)


miekg:
Hi, this is neat. But before I do some reviewing, how does this interact with tls we use
in some plugins for supporting DoT and DoH ? Ideally I want to use this plugin to get them
certs and make stuff happening (ideally in a non-blocking [at startup] way).

Can you flesh out that scenario?

chinzhiweiblank:
> Hi, this is neat. But before I do some reviewing, how does this interact with tls we use in some plugins for supporting DoT and DoH ? Ideally I want to use this plugin to get them certs and make stuff happening (ideally in a non-blocking [at startup] way). Can you flesh out that scenario?

It obtains the TLS configuration for the server like what the tls plugin does in [here](https://github.com/coredns/coredns/blob/5b9b079dabc7f71463cea3f0c6a92f338935039d/plugin/tls/tls.go#L87).

yongtang:
@chinzhiweiblank Can you give an example Corefile setup?

chrisohaver:
> Closes [#3011](https://github.com/coredns/coredns/pull/3011)

do you mean #3460 ?

chinzhiweiblank:
> > Closes [#3011](https://github.com/coredns/coredns/pull/3011)
> 
> do you mean #3460 ?

My mistake. I changed it.

chinzhiweiblank:
> @chinzhiweiblank Can you give an example Corefile setup?
@miekg 
If I understand your question correctly, I should be explaining how the keys in the certificate obtained in this plugin are used in DoT and DoH in other plugins.

```txt
acme {
  domain example.org
}
```

This will obtain the TLS certificate for `example.org`.
This plugin will do the following in this [segment](https://github.com/coredns/coredns/pull/4829/files#diff-58e97756fcb7e20904ac1abc9fe07f1a4d66e5d2b6401e2bd7e07ddce1df9d65R18-R24):
1. Fetch the configuration of the DNS server
2. Obtain the cached certificate
3. Assign the certificate to the configuration

chrisohaver:
I think the question is "how does this interact with the _tls_ plugin ..."

chrisohaver:
Missing: acme plugin added to `plugin.cfg`.

chrisohaver:
> I think the question is "how does this interact with the _tls_ plugin ..."
> Missing: acme plugin added to plugin.cfg.

If placed after _tls_ in plugin.cfg: Looks like it would always override the _tls_ plugin.
If placed before _tls_ in plugin.cfg: incompatible with _tls_ causing CoreDNS not to start if both plugins are present.

chrisohaver:
@chinzhiweiblank,  A while into reviewing this, I'm unsure of what the state of it is ...

Can you summarize the current state of this PR?  Is the plugin functional in its current state or is it a work in progress?  Is it ready for review?

chinzhiweiblank:
> @chinzhiweiblank, A while into reviewing this, I'm unsure of what the state of it is ...
> 
> Can you summarize the current state of this PR? Is the plugin functional in its current state or is it a work in progress? Is it ready for review?

Hi Chris, I think it's a work in progress. It's not ready for review. Do you mind if I make it into a draft or close it for now? I would reopen it when I think it's ready.

chinzhiweiblank:
> > I think the question is "how does this interact with the _tls_ plugin ..."
> > Missing: acme plugin added to plugin.cfg.
> 
> If placed after _tls_ in plugin.cfg: Looks like it would always override the _tls_ plugin.
> If placed before _tls_ in plugin.cfg: incompatible with _tls_ causing CoreDNS not to start if both plugins are present.

I would place it after tls. If the user chooses to use acme, he or she should be aware that the server would be using the keys obtained from acme in the end.

chrisohaver:
> Hi Chris, I think it's a work in progress. It's not ready for review. Do you mind if I make it into a draft or close it for now? I would reopen it when I think it's ready.

Please mark it as a draft.

chrisohaver:
Should this instead be a feature that extends the existing _tls_ plugin? I.e. new options in the _tls_ plugin.  I don't see any obvious _functional_ advantages to them being separate, since one would functionally eclipse the other if both are present.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Add ACME Plug..." ]
>Should this instead be a feature that extends the existing tls plugin? I.e. new
>options in the tls plugin. I don't see any obvious functional advantages to
>them being separate, since one would functionally eclipse the other if both are
>present.

I think that makes sense, ie `tls acme { .... potential options here ....}`, and we need
to think throught the implications; are these certs stored on disk, what about storing
them in memcache? Should we explore caddy's Filesystem interface? This at minimum needs a
little design written down

chrisohaver:
Stale, and I think preferable to extend the existing _tls_ plugin than to add a new plugin.

PR #4831, opened 2021-08-27
Revert "upstream lookups are done with original EDNS options"

Tantalor93:
Reverts coredns/coredns#4826

@miekg as discussed


chrisohaver:
I don't think this needs to be reverted. https://github.com/coredns/coredns/issues/4825#issuecomment-907142104

PR #4835, opened 2021-08-28
introduce new plugin to prevent DNS cache snooping

Tantalor93:

Hello 👋 I would like to circulate an idea for a new CoreDNS plugin, I prepared a draft how it could look like and I would welcome comments

This PR adds a new plugin that can to some extent prevent [DNS cache snooping attacks](https://kb.isc.org/docs/aa-00509). This is inspired by the existing module in [Knot Resolver](https://knot-resolver.readthedocs.io/en/stable/modules-refuse_nord.html).

mention new plugin in documentation

no

### DIFF
--- a/core/dnsserver/zdirectives.go
+++ b/core/dnsserver/zdirectives.go
@@ -31,6 +31,7 @@ var Directives = []string{
 	"local",
 	"dns64",
 	"acl",
+	"refusenord",
 	"any",
 	"chaos",
 	"loadbalance",
--- a/core/plugin/zplugin.go
+++ b/core/plugin/zplugin.go
@@ -42,6 +42,7 @@ import (
 	_ "github.com/coredns/coredns/plugin/nsid"
 	_ "github.com/coredns/coredns/plugin/pprof"
 	_ "github.com/coredns/coredns/plugin/ready"
+	_ "github.com/coredns/coredns/plugin/refusenord"
 	_ "github.com/coredns/coredns/plugin/reload"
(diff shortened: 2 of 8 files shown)


chrisohaver:
The readme should make it clear that this intends to prevent cache snooping of _upstream recursive resolvers_, not of the CoreDNS cache.
The CoreDNS cache will send the request to the next plugin regardless of RD bit state, so an attacker cannot deduce whether or not an item is in the cache by setting the RD bit.

chrisohaver (commented):
Thanks!  

The readme should make it clear that this intends to prevent cache snooping of upstream recursive resolvers, not of the CoreDNS cache.
The CoreDNS cache will send the request to the next plugin regardless of RD bit state, so an attacker cannot deduce whether or not an item is in the cache by setting the RD bit.  

Thinking out loud - an alternate place to add this feature could be in the _forward_ plugin... since it only really applies when the forward plugin is involved.  I.e. an option in the _forward_ plugin that refuses to forward non RD queries.

Tantalor93:
oh right @chrisohaver, CoreDNS cache is not susceptible to cache snooping. This would only prevent upstream DNS cache snooping, so as you said, it would rather make more sense as an option to the `forward` plugin rather than separate plugin🤔  I'll close this 👍

chrisohaver:
I’m not decided on whether or not this makes more sense as a separate plugin or an extension of another. 

Another plugin that could be extended is the acl plugin, adding ability to make flag based rules, in which case there might be some other use cases.

PR #4840, opened 2021-08-30
plugin/rewrite: restore 1.8.3 question revert logic; add tests

### LINKED ISSUE #4834: Breaking changes in ReWrite Plugin CoreDNS 1.8.4
AKS is currently running CoreDNS 1.8.3 and upgrading to CoreDNS 1.8.4 in September.  A breaking change with the rewrite plugin with CoreDNS 1.8.4 has been identified. 
 

This is an example for 1.8.3

```
<domain to be rewritten>.com:53 {
        errors
        cache 30
        rewrite name substring <domain to be rewritten>.com default.svc.cluster.local
          pods insecure
        }
        forward .  /etc/resolv.conf # you can redirect this to a specific DNS server such as 10.0.0.10, but that server must be able to resolve the rewritten domain name
    }
```
The below is an example for 1.8.4 that won't break, which requires customers to update there configuration.
```
  <domain to be rewritten>.com:53 {
  log
  errors
  rewrite stop {
    name regex (.*)\.<domain to be rewritten>.com {1}.default.svc.cluster.local
    answer name (.*)\.default\.svc\.cluster\.local {1}.<domain to be rewritten>.com
  }
  forward . /etc/resolv.conf # you can redirect this to a specific DNS server such as 10.0.0.10, but that server must be able to resolve the rewritten domain name
```

**What happened**:  AKS is advising customers to update there rewrite plugin.

**What you expected to happen**: No breaking changes during patch version upgrades.

**How to reproduce it (as minimally and precisely as possible)**: Apply a configuration in 1.8.3 and upgrade to 1.8.4.

**Environment**:
Azure Kubernetes Service 1.20+

- the version of CoreDNS: 1.8.4
- Corefile:
- logs, if applicable:
- Others:

chrisohaver:

see #4834.  1.8.3 always restored the _question section_ in the response  to the original question.  [1.8.4 stopped doing this](https://github.com/coredns/coredns/pull/4473), resulting in question mismatch errors for most clients (e.g. including `dig`).  

**Note**: This addresses the mismatch between the original question and the  _question section_ in the response.  This is _not_ changing default behavior for the _answer section_ of the response, of which some clients are tolerant of a mismatch (e.g  `dig`).  For clients that are intolerant, a mismatch between the original question and the _answer section_  is resolvable (as it was in 1.8.3), using an `answer` rewrite rule.

Fixes #4834

None

No.

### DIFF
--- a/plugin/rewrite/rewrite.go
+++ b/plugin/rewrite/rewrite.go
@@ -54,7 +54,10 @@ func (rw Rewrite) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 			}
 			wr.ResponseRules = append(wr.ResponseRules, respRules...)
 			if rule.Mode() == Stop {
-				break
+				if !rw.RevertPolicy.DoRevert() {
+					return plugin.NextOrFailure(rw.Name(), rw.Next, ctx, w, r)
+				}
+				return plugin.NextOrFailure(rw.Name(), rw.Next, ctx, wr, r)
 			}
 		}
 	}
--- a/plugin/rewrite/rewrite_test.go
+++ b/plugin/rewrite/rewrite_test.go
@@ -163,6 +163,59 @@ func TestNewRule(t *testing.T) {
 	}
 }
 
+func TestRewriteDefaultRevertPolicy(t *testing.T) {
+	rules := []Rule{}
+
+	r, _ := newNameRule("stop", "prefix", "prefix", "to")
+	rules = append(rules, r)
+	r, _ = newNameRule("stop", "suffix", ".suffix.", ".nl.")
+	rules = append(rules, r)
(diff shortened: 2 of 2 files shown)


miekg:
We can't do a plain revert + potential fix/test because there isn't a small commit to revert?

Otherwise lgtm

chrisohaver:
> We can't do a plain revert + potential fix/test because there isn't a small commit to revert?

The commit included code clean up _and_ the new `answer auto` feature.  Reverting the entire commit will result in backward incompatibility between 1.8.4 and the next release, so I felt it was worth trying to dive in and see what actually broke instead of just reverting.

greenpau (approved):
LGTM 👍

djablonski-moia:
Do you an idea when this might be released?

PR #4841, opened 2021-09-01
plugin/forward: Use new msg.Id for upstream queries

IIvyPy:
change the message id when upstream.

There are two reasons to add  the code blow:
       one reason is the upstream message id reuse the client message id, which can increase the risk of cache-poisoning. For example: one can create a dns query which message id is defined by himself. At the same time, he uses the same message id to fake the response. 
       another reason is the DoH mod of iOS is uses zero as default message id, which increases the rate of FORMERR.

This code uses a random message id to upstream server. And recovery the origin message id after obtaining the result.

### DIFF
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -97,6 +97,13 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts options
 	}
 
 	pc.c.SetWriteDeadline(time.Now().Add(maxTimeout))
+	// records the origin Id before upstream.
+	originId := state.Req.Id
+	state.Req.Id = dns.Id()
+	defer func(){
+		state.Req.Id = originId
+	}()
+
 	if err := pc.c.WriteMsg(state.Req); err != nil {
 		pc.c.Close() // not giving it back
 		if err == io.EOF && cached {
@@ -114,6 +121,10 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts options
 			if err == io.EOF && cached {
 				return nil, ErrCachedClosed
 			}
+			// recovery the origin Id after upstream.
+			if ret != nil{
+				ret.Id = originId
+			}
 			return ret, err
 		}
 		// drop out-of-order responses
@@ -121,6 +132,8 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts options
 			break
 		}
 	}
+	// recovery the origin Id after upstream.
+	ret.Id = originId
 
 	p.transport.Yield(pc)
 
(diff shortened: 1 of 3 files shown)


chrisohaver:
Thanks! Can you rebase to remove the unrelated commits, and add a test?

rdrozhdzh:
will it work with TSIG or other signature records?

chrisohaver:
> will it work with TSIG or other signature records?

Ah. IIUC, it would break (TSIG). The client generated digest would not match the server generated digest.

IIvyPy:
> Thanks! Can you rebase to remove the unrelated commits, and add a test?

okay, I will add a test soon.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] plugin/forwar..." ]
>    will it work with TSIG or other signature records?
>
>Ah. IIUC, it would break. The client generated digest would not match the
>server generated digest.

which TSIG? The one coredns receives? Or one that has been generated in coredns?

chrisohaver:
> [ Quoting ***@***.***> in "Re: [coredns/coredns] plugin/forwar..." ]
> will it work with TSIG or other signature records? Ah. IIUC, it would break. The client generated digest would not match the server generated digest.
> which TSIG? The one coredns receives? Or one that has been generated in coredns?

It would break a tsig transaction between a client and upstream server where coredns sits between the two forwarding traffic. Ie client generates digest and sends request, coredns changes message id and forwards to upstream server, upstream server fails auth since digest won’t match.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] plugin/forwar..." ]
>    [ Quoting @.***> in "Re: [coredns/coredns] plugin/forwar..." ]
>    will it work with TSIG or other signature records? Ah. IIUC, it would
>    break. The client generated digest would not match the server generated
>    digest.
>    which TSIG? The one coredns receives? Or one that has been generated in
>    coredns?
>
>It would break a tsig transaction between a client and upstream server where
>coredns sits between the two forwarding traffic. Ie client generates digest and
>sends request, coredns changes message id and forwards to upstream server,
>upstream server fails auth since digest won’t match.

I couldn't (quickly) find if TSIG is hop-by-hop, but:

https://datatracker.ietf.org/doc/html/rfc8945#section-4.3.2

the ID in the tsig, so this should be fine.

Also (while scrolling through the thing):

"Special Considerations for Forwarding Servers"
https://datatracker.ietf.org/doc/html/rfc8945#section-5.5

(btw: *forward* doens't do any of that)

rdrozhdzh:
I just wanted to pay attention of the author to such cases - they should be testes carefully.

Eventually, it may work, since the forward plugin unpacks the external response and checks TSIG with altered ID, whereas the coredns then re-packs the response to client with original ID.

zengyuxing007:
maybe we should add a config paramter for this ~ #7019

PR #4842, opened 2021-09-02
Enable HTTP/2 in gRPC service

balboah:
### What

Same as for #4182, add "h2" in "Next Protos" when configuring tls.

### Why

This is required for being able to upgrade the connection to http/2 which in turn is what gRPC is built upon.
Without this, I was unable to use gRPC in this setup:

- coredns with the grpc plugin -> google cloud http/2 load balancer -> coredns grpc server

### DIFF
--- a/core/dnsserver/server_grpc.go
+++ b/core/dnsserver/server_grpc.go
@@ -40,6 +40,11 @@ func NewServergRPC(addr string, group []*Config) (*ServergRPC, error) {
 		// Should we error if some configs *don't* have TLS?
 		tlsConfig = conf.TLSConfig
 	}
+	// http/2 is required when using gRPC. We need to specify it in next protos
+	// or the upgrade won't happen.
+	if tlsConfig != nil {
+		tlsConfig.NextProtos = []string{"h2"}
+	}
 
 	return &ServergRPC{Server: s, tlsConfig: tlsConfig}, nil
 }


core/dnsserver/server_grpc.go
```diff
@@ -40,6 +40,9 @@ func NewServergRPC(addr string, group []*Config) (*ServergRPC, error) {
 		// Should we error if some configs *don't* have TLS?
 		tlsConfig = conf.TLSConfig
 	}
+	// http/2 is required when using gRPC. We need to specify it in next protos
+	// or the upgrade won't happen.
+	tlsConfig.NextProtos = []string{"h2"}
```
chrisohaver: ~move to inside for loop.~  Actually that wouldn't help, IIUC an explicit nil check is needed.

balboah: Right, the nil check was only in https server.
Do you think we should also require tls here like what is done in [https server](https://github.com/coredns/coredns/blob/master/core/dnsserver/server_https.go#L42) as this is basically also https?
I added this requirement in the commit, let me know and I'll update

chrisohaver: Perhaps, though that might introduce backward incompatibility if currently we permit an unencrypted gRPC listener.

I was thinking just a simple nil check before accessing `NextProtos`.  e.g.

```suggestion
	if (tlsConfig != nil) {
		tlsConfig.NextProtos = []string{"h2"}
	}
```

miekg: gprc with `WithInsecure()` is a normal thing to do when testing/playing with grpc, so we not make that impossible.
The h2 upgrade also makes sense in normal TLS setup... I think. It's not specified for DoT, but IIRC http/2 is negotiated, so it shouldn't harm either.

Meaning we can just put it here: https://github.com/coredns/coredns/blob/master/plugin/tls/tls.go#L22

balboah:
Updated following the suggestions

plugin/tls/tls.go
```diff
@@ -34,6 +34,8 @@ func setTLSDefaults(tls *ctls.Config) {
 		ctls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
 	}
 	tls.PreferServerCipherSuites = true
+	// Allow negotiating http/2 by default.
+	tls.NextProtos = []string{"h2"}
```
miekg: I believe if we do it here, we don't need to do it in dnsserver/server_grpc.go (but it's a bit hairy to follow that path)

chrisohaver: ... same as `dnsserver/server_https.go`.

balboah: yeah I was thinking about that. but I agree it's more clear to see it inside the server code instead of only hidden in tls package. So I'd suggest leave duplicate code or remove it from tls plugin.
Oh and currently https server also allows http/1.1 which is not something we need here.

What do you think?

miekg:
I think that could be solved by a comment, pointing to the tls plugin from
the server code. It makes tls somewhat special, but so are the bind and
root plugin

balboah:
@miekg and what about https config? Should we remove http/1.1 and use the same "h2" only config?

miekg:
If that doesn't break any clients I'm fine with such a change

balboah:
All up to date with latest and greatest

balboah:
@miekg @chrisohaver all good? 

The two gotchas I can think of is that DoH rfc only *recommends* http/2 as minimum version, doesn't *require*. In theory there may be clients that breaks because they don't support it, hard to know.

DoT doesn't use the h2 setting afaik.

So either we try it like this, or I change it back to be only in grpc server

chrisohaver:
> @miekg @chrisohaver all good?
> 
> The two gotchas I can think of is that DoH rfc only _recommends_ http/2 as minimum version, doesn't _require_. In theory there may be clients that breaks because they don't support it, hard to know.

I suppose server grpc could append "http/1.1" to nextprotos if there's a concern.  But I'm leaning toward "change it back to be only in grpc server,"  as long as an unencrypted grpc listener is still permitted.

balboah:
@chrisohaver yes I agree simple change wins :)
Updated. There are no http/1.1 specified for gRPC server as http/2 is required by gRPC afaik, unlike DoH

chrisohaver (approved):

PR #4844, opened 2021-09-03
The IPv4 parser no longer accepts leading zeros

### LINKED ISSUE #4843: plugin/hosts/hostsfile_test.go fail in golang 1.17
**What happened**:
plugin/hosts/hostsfile_test.go fail in golang 1.17

Run tests using golang-1.17

This was caused by net.ParseIP and net.ParseCIDR functions now rejecting IPv4 addresses which contain decimal components with leading zeros.
https://golang.org/doc/go1.17#minor_library_changes

**Environment**:
- golang 1.17

rouzier:
Fix test golang-1.17

Fixes #4843

None

Yes it does.
Any ip4 address with leading zeros will be skipped.


miekg (approved):
Lgtm

PR #4852, opened 2021-09-07
Update golang version to 1.17

yongtang:

This PR updates golang version to 1.17 when applicable.

n/a

n/a

n/a

### DIFF
--- a/.github/workflows/go.coverage.yml
+++ b/.github/workflows/go.coverage.yml
@@ -9,7 +9,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v2
       with:
-        go-version: '^1.16'
+        go-version: '^1.17'
       id: go
 
     - name: Check out code
--- a/.github/workflows/go.coverage.yml
+++ b/.github/workflows/go.coverage.yml
@@ -9,7 +9,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v2
       with:
-        go-version: '1.17'
+        go-version: '1.17.0'
       id: go
 
     - name: Check out code
--- a/.github/workflows/go.coverage.yml
+++ b/.github/workflows/go.coverage.yml
@@ -9,7 +9,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v2
       with:
-        go-version: '^1.17'
+        go-version: '1.17'
       id: go
 
     - name: Check out code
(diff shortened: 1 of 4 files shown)


chrisohaver:
`go-version: '^1.16'` _should_ install the latest available minor version >= 1.16 ...  So it should pick up 1.17 when it's available to the build platform.  It should only stop short of 2.0.0 (< 2.0.0).

Per comment ... https://github.com/actions/setup-go/issues/101#issuecomment-814127065
and https://www.npmjs.com/package/semver "Caret Ranges"

yongtang:
@chrisohaver  As of yesterday it looks like `go-version: '^1.16' ` still only install 1.16?

https://github.com/coredns/coredns/runs/3512203105?check_suite_focus=true

chrisohaver:
I see... according to docs it checks a local cache for a match first:

 "It will first check the local cache for a version match. If version is not found locally, It will pull it from main branch of go-versions repository and on miss or failure, it will fall back to the previous behavior of download directly from go dist."

Suggesting that 1.17 is not in the "local cache" yet, but 1.16.7 was, so it picked 1.16.7 instead of "pulling from the main branch".

Which means that if the "local cache" is not stable between builds, the precise version used for each run could be unpredictable.

So, yes `^1.17` should force at least 1.17.0, _but maybe the caret notation is not the best notation to use_.

chrisohaver:
> maybe the caret notation is not the best notation to use.

It would seem that the only way to ensure the latest version of Go is used is to specify its full version (major.minor.patch).  Which requires updating it everywhere every time we want to bump the go version.

yongtang:
@chrisohaver changed the commit to use `1.17` instead (no `1.17.0` as patch is not part of the tag for 1.17).

chrisohaver:
IIUC, the git action code corrects for the empty patch vs `0` patch... here... 
 
https://github.com/actions/setup-go/blob/76929df236370590fde8a3a25d829ad91a2b6ed6/src/installer.ts#L206-L211

... so `1.17.0` should match a version advertised as `1.17`, because the code appends a `.0` to the advertised version before checking for a match.

Not sure how an empty patch in the version spec like `1.17` is handled... if it's shorthand for "latest 1.17 patch available", then it's subject to settling on whatever happens to be the latest in the local cache vs what is actually the latest version available.

Technically, removing `^` doesn't avoid cache.  The cache is always checked.

chrisohaver:
It looks like Go 1.17 is advertised in the go-versions manifest as 1.17.0 anyways ... https://github.com/actions/go-versions/blob/main/versions-manifest.json

chrisohaver (approved):

PR #4866, opened 2021-09-14
docker: test makefile and Dockerfile change

miekg:
The fix in 4b9bc138d92b90adafeb977913f96e9030507278 breaks docker with

~~~
Step 2/9 : RUN sed -i.bak s@stable/update@stable-security/update@g /etc/apt/sources.list
 ---> Running in 29a0dbae5746
no status provided on response: unknown
~~~

So reverted here. This adds github workflows to test Makefile.docker
syntax as well. And small updates to the docker-coredns over in the
release repo.

### DIFF
--- a/Makefile.docker
+++ b/Makefile.docker
@@ -67,7 +67,7 @@ else
 	@# 1. Copy appropriate coredns binary to build/docker/linux/<arch>
 	@# 2. Copy Dockerfile to build/docker/linux/<arch>
 	for arch in $(LINUX_ARCH); do \
-	    docker build -t $(DOCKER_IMAGE_NAME)-$${arch}:$(VERSION) build/docker/linux/$${arch} ;\
+	    docker build -t $(DOCKER_IMAGE_NAME)-$${arch}:$(VERSION) build/docker/$${arch}  && \
 	    docker tag $(DOCKER_IMAGE_NAME)-$${arch}:$(VERSION) $(DOCKER_IMAGE_NAME)-$${arch}:latest ;\
 	done
 endif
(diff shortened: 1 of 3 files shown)


Makefile.docker
```diff
@@ -67,7 +67,7 @@ else
 	@# 1. Copy appropriate coredns binary to build/docker/linux/<arch>
 	@# 2. Copy Dockerfile to build/docker/linux/<arch>
```
chrisohaver, line 68: We no longer do either of these things in this target.
```suggestion

```

chrisohaver (changes requested):
Looks good I think.  Not sure about the 4b9bc13 part... but FWIW, when testing it locally, the docker build worked fine worked for me _with or without_ the 4b9bc13 fix in the Dockerfile. 

When testing it, did find some more lingering comments that are no longer relevant, which should be removed.

Makefile.docker
```diff
@@ -67,7 +67,7 @@ else
...
 	for arch in $(LINUX_ARCH); do \
-	    docker build -t $(DOCKER_IMAGE_NAME)-$${arch}:$(VERSION) build/docker/linux/$${arch} ;\
+	    docker build -t $(DOCKER_IMAGE_NAME)-$${arch}:$(VERSION) build/docker/$${arch}  && \
```
chrisohaver, line 70: I'm thinking  `image-download` should pull down images and put them into a directory structure that matches the release build paths, so that a user could build a custom coredns container (e.g. with external plugins), without needing to release the binaries to their GitHub fork as an intermediate step ... e.g. 

```
make -f Makefile.release build
make DOCKER=project -f Makefile.docker docker-build docker-push
```

This would also be useful for testing unreleased versions.

chrisohaver (approved):
lgtm

chrisohaver:
> When testing it, did find some more lingering comments that are no longer relevant, which should be removed.

Removed in #4868

PR #4868, opened 2021-09-14
Makefile.docker: Fix example

chrisohaver:

Fixes the example in the comments.
Also removes some stray comments that no longer apply after prior refactoring.

### DIFF
--- a/Makefile.docker
+++ b/Makefile.docker
@@ -14,9 +14,9 @@
 #
 # Testing docker is done e.g. via:
 #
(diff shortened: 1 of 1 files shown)


Makefile.docker
```diff
@@ -14,9 +14,9 @@
 #
 # Testing docker is done e.g. via:
 #
-#  export DOCKER_PASSWORD=<pass>
```
lionello, line 17: I actually think this space was intentional, so that `DOCKER_PASSWORD` would not get stored in the shell history..

PR #4875, opened 2021-09-21
Fix broken URL link in 1.8.0 release note

vanou:

This PR fixes broken URL link to plugins/dnstap in release note.

NONE

NONE

NONE

### DIFF
--- a/notes/coredns-1.8.0.md
+++ b/notes/coredns-1.8.0.md
@@ -14,7 +14,7 @@ If you are running 1.7.1 you want to upgrade for the *cache* plugin fixes.
 
 This release also adds three backwards incompatible changes. This will only affect you if you have an
 **external plugin** or use **outgoing zone transfers**. If you're using `dnstap` in your plugin,
-you'll need to upgrade to the new API as detailed in it's [documentation](/plugin/dnstap).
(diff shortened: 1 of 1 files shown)


chrisohaver (changes requested):
Thanks, I think an absolute link would work better than a relative one.  Please also sign the DCO.

vanou:
> Thanks, I think an absolute link would work better than a relative one. Please also sign the DCO.

Thanks @chrisohaver . I added DCO.
I think link notation like this ``/plugins/dnstap`` is better than absolute link because
- If CoreDNS change structure of their site, it's easy to modify URLs in their documentation because of consistency with another document (another documents use relative link ,e.g. ``/plugins/dnstap``)

chrisohaver:
> another documents use relative link ,e.g. /plugins/dnstap

Ah right, the links are intended to work in the context of coredns.io, not here (github.com).

vanou:
Thanks @chrisohaver . Can we move forward?

chrisohaver:
> Thanks @chrisohaver . Can we move forward?

Please sign the DCO

chrisohaver:
Ah I see it’s signed, but just the name on email doesn’t match.

Issue #4853, opened 2021-09-09
1.8.5 release tracking


yongtang:
This is an issue to track 1.8.5 releases. 

One reason for a release is #4834, together with what has already been merged.

- [x] Update release notes (PR #4854)
- [x] Update version string to 1.8.5 (PR #4855)
- [x] ~~PR #4842~~ 

/cc @miekg  @chrisohaver

yongtang:
I will take another look to see if there are any remaining items that are needed.

chrisohaver:
Perhaps #4842.

chrisohaver:
> Perhaps #4842.

Yeah, looks like all other open PRs are pending review/approval, or pending author action.

chrisohaver:
Would be nice to see new plugin #4512 merged (opened in March). Just wanted to get @miekg's blessing before merging.

yongtang:
I think #4512 can either go with 1.8.5 or go with 1.9.0 as it is a new plugin so technically "new feature"

For #4842, are we ok with the merge?

chrisohaver:
> I think #4512 can either go with 1.8.5 or go with 1.9.0 as it is a new plugin so technically "new feature"

There are two new plugins in the release already... In the past we have only bumped minor version for backward compatibility issues, not for new features.

> For #4842, are we ok with the merge?

I am (I approved it). But the last commit undid some suggestions that @miekg had made.

yongtang:
> But the last commit undid some suggestions that @miekg had made.

@chrisohaver  If #4842 is not to fix a critical bug then I think we can just release 1.8.5 now. And then go with release 1.9.0 to include #4512 and #4842. We don't need to wait too long to release 1.9.0 anyway.

If #4842 is a prioritized bug then we can wait for Miek's comments in review getting addressed before 1.8.5 release.

chrisohaver:
Yes, Let’s just release now, and we can release 1.8.6 soon after if there are any pressing bugs.

chrisohaver:
Just double-checked: No config backward incompatible changes in the list.  One PR deprecated the cache miss metric, but it remains present for metrics backward compatibility. Opened #4857 to update release notes.

yongtang:
With all related PR merged, I will try to trigger a release for 1.8.5 shortly.

yongtang:
Looks like we hit some issues with the release. Will give it another try, and, hopefully with success.

miekg:
[...]

PR #4877, opened 2021-09-21
Makefile.release: Force http 1.1

chrisohaver:

Perhaps not a great idea to do this long term, but for now force http 1.1 for curl ... taking a stab in the dark at sidestepping the recent http2 failures seen in #4853.


chrisohaver:
I suppose we can consider this if the [token troubleshooting](https://github.com/coredns/coredns/issues/4853#issuecomment-924040033) doesn't pan out.

miekg:
Not needed, see release issue comments

/close

PR #4891, opened 2021-09-27
plugin/k8s_external: fix external nsAddrs when CoreDNS Service has no External IPs

### LINKED ISSUE #4887: k8s_external returning pod IPs for ns1.dns AAAA records instead of service IP
**What happened**:
k8s_external returns pod IPs for ns1.dns when no Load balancer IP is available

k8s_external returns the service's (in this config, routable) service IP address when no LB IP exists

`dig ns1.dns.<k8s_external zone> @<kube-dns service> AAAA` with service type set to ClusterIP (in an IPv6 cluster?)

Untested on IPv4 as cluster in question is IPv6 only.  
Behavior resolved as soon as type was changed to LoadBalancer to allocate an LB IP

**Environment**:

- the version of CoreDNS: `coredns:1.8.4`
- Corefile: one that uses `k8s_external`
- logs, if applicable:
- OS (e.g: `cat /etc/os-release`): Debian buster under kubeadm-k8s
- Others: kube-dns configured without a LoadBalancer IP address (now fixed, but this is about the unexpected behavior of `externalAddrFunc` when no Ingress or LB exists)

chrisohaver:

Context _k8s_external_: When CoreDNS is running _in cluster_, but has no external IPs defined for its Service object, the additional section in an NS response should be empty - since there are no valid IP address to return.  Currently, we incorrectly fallback to returning local IP addresses, as if the CoreDNS instance was running outside of the cluster.

Closes #4887

### DIFF
--- a/plugin/kubernetes/ns.go
+++ b/plugin/kubernetes/ns.go
@@ -15,8 +15,9 @@ func isDefaultNS(name, zone string) bool {
 // it returns a record for the local address of the machine we're running on.
 func (k *Kubernetes) nsAddrs(external bool, zone string) []dns.RR {
 	var (
-		svcNames []string
-		svcIPs   []net.IP
+		svcNames     []string
+		svcIPs       []net.IP
+		foundService bool
 	)
 
 	// Find the CoreDNS Endpoints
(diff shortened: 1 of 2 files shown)


miekg:
Looking at this on mobile.

Do we need a Boolean for this? Can't we infer no svcs found from the
returned values being empty?

chrisohaver:
> Looking at this on mobile. Do we need a Boolean for this? Can't we infer no svcs found from the returned values being empty?

We do - but looking at this again, I should adjust the code to be more clear ...

We need to distinguish between "no endpoints matched" vs "an endpoint matched, but no services had an external IP".

**an endpoint matched, but no services had an external IP**: we assume CoreDNS is a Pod inside the cluster, so we return an empty list of addresses.

**no endpoints matched** : we assume CoreDNS is running external to the cluster, so we return the local ip addresses.


I think it would be more clear if I rename the bool to `foundEndpoint`, and set it to true when matching endpoints are found

chrisohaver:
looks like benchmark test flaked.

chrisohaver:
ping

Issue #4899, opened 2021-09-29
plugin/route53 CoreDNS fails to start when API throttling occurs


rnsc:
**What happened**:
Tried to start two instances of CoreDNS with the same AWS Route53 zones configured.
CoreDNS failed to start on both instances because they ate each other's request allocation on the AWS API.
Throttling kicks in and results in plugins failing.

**What you expected to happen**:
I would expect CoreDNS to at least retry recoverable errors on the AWS HTTP API instead of failing at the first error.

**How to reproduce it (as minimally and precisely as possible)**:
To reproduce this, you'll need a good amount of hosted zones (> 200) on AWS Route53 and start multiple instances of CoreDNS with that config.

This docker compose file is sufficient:
```yaml
---
version: '3'

services:
  dns_one: &route53
    image: coredns
    volumes:
      - $PWD/conf:/etc/coredns
    environment:
      AWS_ACCESS_KEY_ID: $AWS_ACCESS_KEY_ID
      AWS_SECRET_ACCESS_KEY: $AWS_SECRET_ACCESS_KEY
    restart: always
  dns_two: *route53
```

CoreDNS was built with the following Dockerfile:

```raw
FROM golang:1.16 as build_stage

RUN mkdir /v
ADD coredns /v

WORKDIR /v

RUN make

FROM scratch

COPY --from=build_stage /v/coredns /

CMD ["/coredns", "-conf", "/etc/coredns/Corefile"]
```

**Anything else we need to know?**:
I did some update to the route53 plugin to overcome this issue on my local setup by adding a MaxRetries value to the AWS session.
I can provide a PR if you want to discuss it.

**Environment**:

- the version of CoreDNS: CoreDNS-1.8.5
- Corefile: N/A
- logs, if applicable: N/A
- OS (e.g: `cat /etc/os-release`): N/A
- Others: N/A

Issue #4894, opened 2021-09-28
plugin/route53 blocking reload/restart


rnsc:
**What happened**:

Health and readiness endpoints are unresponsive when reloading a config with a few hundreds AWS Route53 zones.
It takes 5 minutes for CoreDNS to start (after a stop), all the while no health checks are responding, nor prometheus endpoints, and nothing can be served until all the AWS Route53 zones are fully loaded.

So, I have more than 500 AWS Route 53 hosted zones configured into CoreDNS.
The config also has a `.` default with a forward config to external DNS.
Each zone is configured with `ready` and `prometheus`.

My AWS hosted zones definition is not in the Corefile, but is imported with a simple `import`.

A start, or reload of the config always takes around 5 minutes.

Start:
```bash
Sep 28 11:20:27 systemd[1]: Started CoreDNS DNS server.
Sep 28 11:25:22 coredns[3734616]: .:53
```

Reload:
```bash
Sep 28 11:50:56 systemd[1]: Reloading CoreDNS DNS server.
Sep 28 11:50:56 coredns[3734616]: [INFO] SIGUSR1: Reloading
Sep 28 11:50:56 coredns[3734616]: [INFO] Reloading
Sep 28 11:50:56 systemd[1]: Reloaded CoreDNS DNS server.
Sep 28 11:55:44 coredns[3734616]: [INFO] Reloading complete
```

**What you expected to happen**:

I would at least expect the health/ready endpoints to answer something rather than not listening at all.
I understand that the DNS server itself is not available until all zones are loaded. Upon reload, the DNS server is still up and replies to requests properly, even if the health/ready endpoints are dead.
I guess it's also related, if I do a `systemctl reload coredns` while a start/reload is happening, systemd just times out on waiting and kills the process. So there's a thread locking the main process?

**How to reproduce it (as minimally and precisely as possible)**:

I pasted my Corefile below, the imported config is just 570 zones with the route53 plugin defined in each of them.
I'm provisioning them with Ansible and have a randomised refresh to avoid throttling on the AWS API.

**Anything else we need to know?**:

**Environment**:

- the version of CoreDNS: `CoreDNS-1.8.4`
- Corefile: 
```
. {
  errors
  health
  ready
  prometheus :9153
  cache 600
  minimal
  forward . 8.8.8.8 8.8.4.4
}

import aws-zones.config
```
- logs, if applicable: (see above)
- OS (e.g: `cat /etc/os-release`): 
```
[...]

PR #4902, opened 2021-09-30
Plugin/route53 AWS API MaxRetries parameter

rnsc:
This PR adds the possibility to tune the MaxRetries parameter of the AWS session created with the AWS SDK.

https://github.com/coredns/coredns/issues/4899

I already updated the README for the plugin.

No, it's a new parameter with a default.

### DIFF
--- a/plugin/route53/README.md
+++ b/plugin/route53/README.md
@@ -22,34 +22,38 @@ route53 [ZONE:HOSTED_ZONE_ID...] {
 }
 ~~~
 
-*   **ZONE** the name of the domain to be accessed. When there are multiple zones with overlapping
-    domains (private vs. public hosted zone), CoreDNS does the lookup in the given order here.
-    Therefore, for a non-existing resource record, SOA response will be from the rightmost zone.
+* **ZONE** the name of the domain to be accessed. When there are multiple zones with overlapping
+ domains (private vs. public hosted zone), CoreDNS does the lookup in the given order here.
+ Therefore, for a non-existing resource record, SOA response will be from the rightmost zone.
 
-*   **HOSTED\_ZONE\_ID** the ID of the hosted zone that contains the resource record sets to be
-    accessed.
+* **HOSTED\_ZONE\_ID** the ID of the hosted zone that contains the resource record sets to be
+ accessed.
 
-*   **AWS\_ACCESS\_KEY\_ID** and **AWS\_SECRET\_ACCESS\_KEY** the AWS access key ID and secret access key
-    to be used when query AWS (optional). If they are not provided, then coredns tries to access
-    AWS credentials the same way as AWS CLI, e.g., environmental variables, AWS credentials file,
-    instance profile credentials, etc.
+* **AWS\_ACCESS\_KEY\_ID** and **AWS\_SECRET\_ACCESS\_KEY** the AWS access key ID and secret access key
+ to be used when query AWS (optional). If they are not provided, then coredns tries to access
+ AWS credentials the same way as AWS CLI, e.g., environmental variables, AWS credentials file,
+ instance profile credentials, etc.
 
-*   `credentials` is used for reading the credential **FILENAME** and setting the **PROFILE** name for a given
-    zone. **PROFILE** is the AWS account profile name. Defaults to `default`. **FILENAME** is the
-    AWS credentials filename, defaults to `~/.aws/credentials`.
+* `credentials` is used for reading the credential **FILENAME** and setting the **PROFILE** name for a given
+ zone. **PROFILE** is the AWS account profile name. Defaults to `default`. **FILENAME** is the
+ AWS credentials filename, defaults to `~/.aws/credentials`.
 
-*   `fallthrough` If zone matches and no record can be generated, pass request to the next plugin.
-    If **ZONES** is omitted, then fallthrough happens for all zones for which the plugin is
-    authoritative. If specific zones are listed (for example `in-addr.arpa` and `ip6.arpa`), then
-    only queries for those zones will be subject to fallthrough.
+* `fallthrough` If zone matches and no record can be generated, pass request to the next plugin.
+ If **ZONES** is omitted, then fallthrough happens for all zones for which the plugin is
+ authoritative. If specific zones are listed (for example `in-addr.arpa` and `ip6.arpa`), then
+ only queries for those zones will be subject to fallthrough.
 
-*   `refresh` can be used to control how long between record retrievals from Route 53. It requires
-    a duration string as a parameter to specify the duration between update cycles. Each update
-    cycle may result in many AWS API calls depending on how many domains use this plugin and how
-    many records are in each. Adjusting the update frequency may help reduce the potential of API
-    rate-limiting imposed by AWS.
+* `refresh` can be used to control how long between record retrievals from Route 53. It requires
+ a duration string as a parameter to specify the duration between update cycles. Each update
+ cycle may result in many AWS API calls depending on how many domains use this plugin and how
+ many records are in each. Adjusting the update frequency may help reduce the potential of API
+ rate-limiting imposed by AWS.
 
-*   **DURATION** A duration string. Defaults to `1m`. If units are unspecified, seconds are assumed.
+* `aws_api_max_retries` can be used to retry recoverable errors on the AWS API. By default errors
+    are not retried and thus, in case of throttling on the AWS API, the plugin doesn't update the zone.
+    The default value is 0, meaning that the pluging doesn't retry.
+
+* **DURATION** A duration string. Defaults to `1m`. If units are unspecified, seconds are assumed.
 
 ## Examples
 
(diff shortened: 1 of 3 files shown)


plugin/route53/setup.go
```diff
@@ -109,6 +115,20 @@ func setup(c *caddy.Controller) error {
 				} else {
 					return plugin.Error("route53", c.ArgErr())
 				}
+			case "aws_api_max_retries":
```
miekg: this may be better named 'max_retries'

plugin/route53/README.md
```diff
@@ -22,42 +22,46 @@ route53 [ZONE:HOSTED_ZONE_ID...] {
...
-*   `refresh` can be used to control how long between record retrievals from Route 53. It requires
-    a duration string as a parameter to specify the duration between update cycles. Each update
-    cycle may result in many AWS API calls depending on how many domains use this plugin and how
-    many records are in each. Adjusting the update frequency may help reduce the potential of API
-    rate-limiting imposed by AWS.
+* `refresh` can be used to control how long between record retrievals from Route 53. It requires
+ a duration string as a parameter to specify the duration between update cycles. Each update
+ cycle may result in many AWS API calls depending on how many domains use this plugin and how
+ many records are in each. Adjusting the update frequency may help reduce the potential of API
+ rate-limiting imposed by AWS.
```
miekg: what's up with all these unrelated changes? Please revert those

rnsc: Right, my markdown linter in VSCode was complaining, thus me trying to fix it, I'll revert it, sorry

plugin/route53/setup.go
rnsc: Done

plugin/route53/README.md
rnsc: Reverted!

rnsc:
Not sure what happened with the reviewers addition following my conflict resolution... I'm sorry about the spam

yongtang:
@rnsc Can you try rebase with the latest master (assuming `upstream` is pointing to coredns/coredns git):
```
git checkout plugin/route53-aws-api-retry-parameter
git fetch upstream master
git rebase upstream/master
git push -f origin plugin/route53-aws-api-retry-parameter
```

rnsc:
> @rnsc Can you try rebase with the latest master (assuming `upstream` is pointing to coredns/coredns git):
> 
> ```
> git checkout plugin/route53-aws-api-retry-parameter
> git fetch upstream master
> git rebase upstream/master
> git push -f origin plugin/route53-aws-api-retry-parameter
> ```

Hello @yongtang I did just that, and I get conflicts on go.mod and go.sum, files which I haven't touched myself, wondering if I should fork from scratch?
image
image

yongtang:
@rnsc You can try manually remove the duplicate locally to resolve the conflict. From the screenshot it seems to be purely duplicate so you can delete either one (after the build I guess go will pick up the latest one anyway).

rnsc:
@yongtang Ok, I think I fixed it, thanks for the help!

plugin/route53/README.md
```diff
@@ -49,6 +49,10 @@ route53 [ZONE:HOSTED_ZONE_ID...] {
     many records are in each. Adjusting the update frequency may help reduce the potential of API
     rate-limiting imposed by AWS.
 
+*   `max_retries` can be used to retry recoverable errors on the AWS API. By default errors
+    are not retried and thus, in case of throttling on the AWS API, the plugin doesn't update the zone.
+    The default value is 0, meaning that the pluging doesn't retry.
```
chrisohaver, line 54:
```suggestion
    The default value is 0, meaning that the plugin doesn't retry.
```

chrisohaver: Does this apply while establishing a connection, or while the connection is active,  or both?

What is considered a "recoverable" vs "unrecoverable" error?

rnsc: @chrisohaver I believe this documentation describes this, maybe it should be mentioned here?The AWS Go documentation mentions "retryable errors", so my wording could be confusing because it doesn't match that:
https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws/retry

> By default the standard retryer uses the DefaultRetryables slice of IsErrorRetryable types to determine whether a given error is retryable. By default this list of retryables includes the following:
```raw
- Retrying errors that implement the RetryableError method, and return true.
- Connection Errors
  - Errors that implement a ConnectionError, Temporary, or Timeout method that return true.
  - Connection Reset Errors.
  - net.OpErr types that are dialing errors or are temporary.
  - HTTP Status Codes: 500, 502, 503, and 504.
- API Error Codes
  - RequestTimeout, RequestTimeoutException
  - Throttling, ThrottlingException, ThrottledException, RequestThrottledException, TooManyRequestsException,
    RequestThrottled, SlowDown, EC2ThrottledException
  - ProvisionedThroughputExceededException, RequestLimitExceeded, BandwidthLimitExceeded, LimitExceededException
  - TransactionInProgressException, PriorRequestNotComplete
```

rnsc: And as far as I understood the documentation, the parameter is set when initialising the client connection

chrisohaver: Thanks! ....  so I'm guessing it means pretty much any error other than an active connection refused or failed auth.  Maybe just don't mention "recoverable errors", since details of exactly what that means are a tad esoteric. e.g. 

> `max_retries` can be used to retry failed connections to the AWS API.

We should provide guidance here on when this should be used, and what value it should be set to.  Otherwise it may come across as a "magical number knob".

Are there drawbacks to setting this number very high? 

Is zero retries a sensible default?

chrisohaver: IOW - does this need to be configurable?  Can we reasonably pick some magic number that would work.

rnsc: @chrisohaver thank you for the better wording example, it does make sense to have it more generic.

In regards to defaults, as far as my testing went and how I understood the documentation, it's 0 by default already, thus my idea to define the variable and set it to 0 and have a non mandatory configuration option available to override it.

Now a disclaimer on the impact of setting that value should be written then, because due to the exponential backoff feature that is backed in the SDK, if you set the retries to, for example, 20, and you encounter a lot of errors, throttling, that could lead to a very long loading time.

rnsc: Maybe adding a max value check in the code could be worth it, to avoid people trying huge values that would be detrimental in case of high errors rate on the API?

chrisohaver: > if you set the retries to, for example, 20, and you encounter a lot of errors, throttling, that could lead to a very long loading time.

OK, but the alternative is CoreDNS exits ~immediately~ before starting.  

From a user standpoint, the actual number of retries is somewhat immaterial.  What actually matters is how long we would retry before giving up and aborting CoreDNS start up.

IMO, it might be better to avoid this by not requiring a successful API connection before CoreDNS starts.
A good behavior that doesn't require user config would be for CoreDNS to start, while the plugin retries connections in the background (logging errors on failure).

chrisohaver: >> Is zero retries a sensible default?

> In regards to defaults, as far as my testing went and how I understood the documentation, it's 0 by default already, thus my idea to define the variable and set it to 0 and have a non mandatory configuration option available to override it.

I intended to question whether or not zero retries makes sense as the default behavior.  i.e. ...
Perhaps a value other than zero would make a better default.
And then it follows that if we can pick a default that would work well for everyone, we don't need to expose a knob.

rnsc: @chrisohaver that's also fine with me, refactoring the plugin to support an async method to lazy load the configured zones is not really something I'd be comfortable with though (from a coding standpoint, I'd rather have someone with better Go knowledge take a look at it), I thought this was an ok middle ground solution before a better approach can be put in place, wdyt?

rnsc: > > > Is zero retries a sensible default?
> 
> > In regards to defaults, as far as my testing went and how I understood the documentation, it's 0 by default already, thus my idea to define the variable and set it to 0 and have a non mandatory configuration option available to override it.
> 
> I intended to question whether or not zero retries makes sense as the default behavior. i.e. ... Perhaps a value other than zero would make a better default. And then it follows that if we can pick a default that would work well for everyone, we don't need to expose a knob.

Ah, right, I see what you mean! I'm also fine with a default low retry value without exposing the value to users, that's probably better!

chrisohaver:
> A good behavior that doesn't require user config would be for CoreDNS to start, while the plugin retries connections in the background (logging errors on failure).

I think solving it this way would also fix #4894.

plugin/route53/README.md
chrisohaver: > I thought this was an ok middle ground solution before a better approach can be put in place, wdyt?

Yes, although ideally we shouldn't have to deprecate a user config when that happens, hence hoping we can pick a sane default without making it user configurable.

chrisohaver:
I think we can just add a set number of retries, without adding a user config. e.g. 5 or something that would fix #4899, without hijacking start up too long if a connection to the API is never successful.

PR #4908, opened 2021-10-06
plugin/kubernetes: Don't use pod names longer than 63 characters as dns labels

chrisohaver:

When the `endpoint_pod_names` option is used, only use Pod names shorter than 64 characters.  For longer pod names, fallback to standard endpoint name convention.

https://kubernetes.slack.com/archives/C09QYUH5W/p1633514843299500


chrisohaver:
I'll add a test to coredns/ci kubernetes-tests.

bradbeam (approved):

aojea:
+1 
Thanks for the quick response

PR #4914, opened 2021-10-08
Metrics: expand coredns_dns_responses_total with plugin label

miekg:
This adds (somewhat hacky?) code to add a plugin label to the
coredns_dns_responses_total metric. It's completely obvlious to the
plugin as we just check who called the *recorder.WriteMsg method. We use
runtime.Caller( 1 2 3) to get multiple levels of callers, this should be
deep enough, but it depends on the dns.ResponseWriter wrapping that's
occuring.

README.md of metrics updates and test added in test/metrics_test.go to
check for the label being set.

I went through the plugin to see what metrics could be removed, but
actually didn't find any, the plugin push out metrics that make sense.

Due to the path fiddling to figure out the plugin name I doubt this
works (out-of-the-box) for external plugins, but I haven't tested that.

### DIFF
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -328,7 +328,7 @@ func errorAndMetricsFunc(server string, w dns.ResponseWriter, r *dns.Msg, rc int
 	answer.SetRcode(r, rc)
 	state.SizeAndDo(answer)
 
-	vars.Report(server, state, vars.Dropped, rcode.ToString(rc), answer.Len(), time.Now())
+	vars.Report(server, state, vars.Dropped, rcode.ToString(rc), "", answer.Len(), time.Now())
 
 	w.WriteMsg(answer)
 }
--- a/core/dnsserver/server.go
+++ b/core/dnsserver/server.go
@@ -328,7 +328,7 @@ func errorAndMetricsFunc(server string, w dns.ResponseWriter, r *dns.Msg, rc int
 	answer.SetRcode(r, rc)
 	state.SizeAndDo(answer)
 
-	vars.Report(server, state, vars.Dropped, rcode.ToString(rc), "", answer.Len(), time.Now())
+	vars.Report(server, state, vars.Dropped, rcode.ToString(rc), "" /* plugin */, answer.Len(), time.Now())
 
 	w.WriteMsg(answer)
 }
--- a/plugin/metrics/handler.go
+++ b/plugin/metrics/handler.go
@@ -2,10 +2,10 @@ package metrics
 
 import (
 	"context"
+	"path/filepath"
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/metrics/vars"
-	"github.com/coredns/coredns/plugin/pkg/dnstest"
 	"github.com/coredns/coredns/plugin/pkg/rcode"
 	"github.com/coredns/coredns/request"
 
@@ -23,7 +23,7 @@ func (m *Metrics) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	}
 
 	// Record response to get status code and size of the reply.
-	rw := dnstest.NewRecorder(w)
+	rw := NewRecorder(w)
 	status, err := plugin.NextOrFailure(m.Name(), m.Next, ctx, rw, r)
 
 	rc := rw.Rcode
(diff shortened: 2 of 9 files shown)


miekg:
yeah, unrelated plugin has weird test failure .... 

... well that test code uses the responseWriter I changed here, but the change looks innocent... odd.

miekg:
Might be cleaner to do a metric's response writer in this case, i.e. noone else needs access to Callers. But would be nice to explain the trace failure

core/dnsserver/server.go
```diff
@@ -328,7 +328,7 @@ func errorAndMetricsFunc(server string, w dns.ResponseWriter, r *dns.Msg, rc int
 	answer.SetRcode(r, rc)
 	state.SizeAndDo(answer)
 
-	vars.Report(server, state, vars.Dropped, rcode.ToString(rc), answer.Len(), time.Now())
+	vars.Report(server, state, vars.Dropped, rcode.ToString(rc), "", answer.Len(), time.Now())
```
dilyevsky: nit:
```suggestion
	vars.Report(server, state, vars.Dropped, rcode.ToString(rc), "" /* plugin */, answer.Len(), time.Now())
```

plugin/metrics/recorder.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/miekg/dns"
+)
+
+// Recorder is a dnstest.Recorder specific to the metrics plugin.
+type Recorder struct {
+	*dnstest.Recorder
+	// CallerN holds the string return value of the call to runtime.Caller(N)
+	Caller1 string
```
dilyevsky: Could probably be `Caller [3]string` ?

dilyevsky (approved):

plugin/metrics/recorder.go
miekg: good point. Changed. removes a bunch of lines of code as well.

miekg:
interesting test failures, i suspect we're scraping the wrong prometheus or something.

Yep the metrics endpoint isn't completely downed before a new test runs, we've been playing with that startup/shutdown since forever, so I don't want to change that here, but the order of tests matters and metrics linger here.

chrisohaver:
Thanks, clever approach! Some initial thoughts...

1. IIRC, the cache plugin does the writing to the client for hits and misses (everything), so I think if the cache plugin is used, all answers would be attributed to cache.
2. If there is a significant performance impact of using this kind of introspection, it should be optional.

miekg:
If the cache plugin made the write than that is the info you get. Updating the metrics is also atomic, runtime.Caller doesn't say anything about slowness, so I think we should be ok.

plugin/metrics/handler.go
```diff
@@ -33,10 +33,25 @@ func (m *Metrics) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
...
+
+// authoritativePlugin returns which of made the write, if none is found the empty string is returned.
+func (m *Metrics) authoritativePlugin(caller [3]string) string {
+	// a b and c contain the full path of the caller, the plugin name 2nd last elements
+	// .../coredns/plugin/whoami/whoami.go --> whoami
+	// this is likely FS specific, so use filepath.
+	for _, c := range caller {
+		plug := filepath.Base(filepath.Dir(c))
+		if _, ok := m.plugins[plug]; ok {
+			return plug
```
chrisohaver, line 53: I tinkered with this locally a bit... 

if we continue to iterate through the `Caller` list in order and return the last one before we reach a "non-plugin" one, I think it will more accurately reflect the plugin that "initiated" the write.

e.g. reporting forward or cache depending on whether or not cache was actually used.

chrisohaver:
> If the cache plugin made the write than that is the info you get. Updating the metrics is also atomic, runtime.Caller doesn't say anything about slowness, so I think we should be ok.

FWIW, I think _log_ would also act undesirably (everything attributed to _log_), except that _log_ just happens to use a response writer defined in a shared package, so its not identified as a  "plugin"...

chrisohaver:
We should probably concretely document the pretty significant caveats here... e.g. that if you are using `cache`, this metric will only ever log `cache` as the plugin, regardless if the response was actually served from cache or elsewhere.  Cache is a pretty common plugin to use.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] Metrics: expa..." ]
>We should probably concretely document the pretty significant caveats here...
>e.g. that if you are using cache, this metric will only ever log cache as the
>plugin, regardless if the response was actually served from cache or elsewhere.
>Cache is a pretty common plugin to use.

I might not see it becasue I'm too in the code, but that's exactly the data people want to
know. You can't have it both ways

PR #4934, opened 2021-10-25
plugin/metrics: support HTTPS qType in requests count metric label

plieskovsky:
Since iOS14, apple devices are standardly using the HTTPS query (type 65). It would be beneficial to be able to distinguish the amount of these requests based on the metrics. Therefore I would like to add the HTTPS query type support into the metric labels.

none

none

no

### DIFF
--- a/plugin/metrics/README.md
+++ b/plugin/metrics/README.md
@@ -30,7 +30,7 @@ Extra labels used are:
 * `proto` which holds the transport of the response ("udp" or "tcp")
 * The address family (`family`) of the transport (1 = IP (IP version 4), 2 = IP6 (IP version 6)).
 * `type` which holds the query type. It holds most common types (A, AAAA, MX, SOA, CNAME, PTR, TXT,
-  NS, SRV, DS, DNSKEY, RRSIG, NSEC, NSEC3, IXFR, AXFR and ANY) and "other" which lumps together all
+  NS, SRV, DS, DNSKEY, RRSIG, NSEC, NSEC3, HTTPS, IXFR, AXFR and ANY) and "other" which lumps together all
   other types.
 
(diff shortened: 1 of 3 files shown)


SuperQ:

plieskovsky:
@SuperQ added, thank you for the info 👍

SuperQ (approved):
LGTM

miekg:
happy to add this, but isn't the HTTPS still in draft? Might be prudent to wait until RFC.

yeah it still draft: https://datatracker.ietf.org/doc/draft-ietf-dnsop-svcb-https/

plieskovsky:
@miekg ah, that's interesting site/fact. Didn't know about it.
On the other hand, iOS has been already using it for about 4 months. We can see a non-trivial amount (~18%) of HTTPS requests sent to our CoreDNS instance that faces the internet. So I guess we cannot go wrong with enabling the computation of it into the metric label.

But as you guys are the owners/maintainers of the repo, the decision is totally up to you :)

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] plugin/metric..." ]
***@***.*** ah, that's interesting site/fact. Didn't know about it.
>On the other hand, iOS has been already using it for about 4 months. We can see
>a non-trivial amount (~18%) of HTTPS requests sent to our CoreDNS instance that
>faces the internet. So I guess we cannot go wrong with enabling the computation
>of it into the metric label.
>
>But as you guys are the owners/maintainers of the repo, the decision is totally
>up to you :)

yeah, it's a label value so easily changed and I can see this RR being import, but it
might still see a name change. SVCB might be nice as well (same caveats)

/lgtm
/merge

PR #4944, opened 2021-11-01
Plugin/file: add minInterval avoid cpu load too high

sstoner:
Signed-off-by: vanceli <vanceli@tencent.com>

when plugin `file` reload duration is too low:
the CPU: 
image

wecom-temp-65bfbd4e2a0ed182c01534f267feb5cd

none

we should notify users when they set these args

https://github.com/coredns/coredns/blob/master/plugin/file/README.md#syntax

No

### DIFF
--- a/plugin/file/setup.go
+++ b/plugin/file/setup.go
@@ -108,11 +108,19 @@ func fileParse(c *caddy.Controller) (Zones, error) {
 		for c.NextBlock() {
 			switch c.Val() {
 			case "reload":
-				d, err := time.ParseDuration(c.RemainingArgs()[0])
-				if err != nil {
+				args := c.RemainingArgs()
+				if len(args) > 0 {
+					d, err := time.ParseDuration(args[0])
+					if err != nil {
+						return Zones{}, plugin.Error("file", err)
+					}
+					reload = d
+				}
+
+				if reload != 0 && reload < minInterval {
 					return Zones{}, plugin.Error("file", err)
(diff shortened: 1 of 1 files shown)


chrisohaver (changes requested):
Thanks!

I'm not sure if 2 seconds is too high, or even if we should impose a minimum at all. Nevertheless if we do, the README should document it.

Technically this change alters the syntax since it would allow a bare `reload` without a duration. I don't think that change is necessary.  Although, it would be much better to error out than to crash when DURATION is not specified (as the existing code in master does now).

sstoner:
> I'm not sure if 2 seconds is too high, or even if we should impose a minimum at all. Nevertheless if we do, the README should document it.

I will do the benchmark and post the result here 

> Technically this change alters the syntax since it would allow a bare reload without a duration.

the document said `Default is one minute`, but the default value is never used unless we set a specific value for `reload`. The crash may not be friendly...

sstoner:
I test again... CPU load will rise obviously when the duration is set below the millisecond level. If it's not necessary, I will close this MR

900 µs
image

1ms 
image

chrisohaver:
> the document said Default is one minute, but the default value is never used unless we set a specific value for reload.

IIRC, the default is used if `reload` is not specified at all.

I'm hesitant to merge this as is since it introduces a backward incompatibility for anyone with a reload set to less than 2 seconds.

PR #4955, opened 2021-11-01
plugin/file: Fix file resource leak in fileParse

victory460:

```reader, err := os.Open(fileName)``` 
reader need to close,when it used over
No.
No.
No.

### DIFF
--- a/plugin/file/setup.go
+++ b/plugin/file/setup.go
@@ -121,6 +121,7 @@ func fileParse(c *caddy.Controller) (Zones, error) {
 				return Zones{}, c.Errf("unknown property '%s'", c.Val())
 			}
 		}
+		reader.Close()
 	}
 
 	for origin := range z {

### CODE REVIEW

File: plugin/file/setup.go [hunk 1]
```diff
@@ -121,6 +121,7 @@ func fileParse(c *caddy.Controller) (Zones, error) {
 				return Zones{}, c.Errf("unknown property '%s'", c.Val())
 			}
 		}
+		reader.Close()
```
chrisohaver, line 124: Would be better to do `defer reader.Close()` right after opening.

chrisohaver (changes requested):

chrisohaver:
It looks like the zone parser actually closes the file when it hits the end... in `subNext()`.

I don't know if the zone parser could exit before it reaches the end without closing.

@victory460, have you observed a leak here? Or are you basing this on looking at the code.

plugin/file/setup.go
victory460: Oh, sorry,  the reader is closed in github.com/miekg/dns subNext() .
```
func (zp *ZoneParser) subNext() (RR, bool) {
	if rr, ok := zp.sub.Next(); ok {
		return rr, true
	}

	if zp.sub.osFile != nil {	//this 
		zp.sub.osFile.Close()
		zp.sub.osFile = nil
	}

	if zp.sub.Err() != nil {
		// We have errors to surface.
		return nil, false
	}

	zp.sub = nil
	return zp.Next()
}
```
Thanks.

victory460:
> It looks like the zone parser actually closes the file when it hits the end... in `subNext()`.
> 
> I don't know of the zone parser could exit before it reaches the end without closing.
> 
> @victory460, have you observed a leak here? Or are you basing this on looking at the code.

Thank you for your reply. the reader is closed in github.com/miekg/dns subNext() .

PR #4957, opened 2021-11-02
plugin/tsig: new plugin TSIG

chrisohaver:

*tsig* - validate TSIG requests and sign responses.

It does the following ...
1. Allows user to define TSIG secret keys.
2. Adds TSIG RRs to responses ( if request was TSIG'd ).

The `miekg/dns` package does the actual TSIG validation and signing.

This plugin does not support outgoing TSIG requests (e.g. Notifies, or proxy TSIG via forward).  That will require mods to those plugins (e.g. configuring which key/algo to use)


miekg:
neat!

core/dnsserver/config.go
```diff
@@ -40,6 +40,9 @@ type Config struct {
 	// TLSConfig when listening for encrypted connections (gRPC, DNS-over-TLS).
 	TLSConfig *tls.Config
 
+	// TSIG secrets, [name]key
```
miekg:
```suggestion
	// TSIG secrets, [name]key.
```

plugin/tsig/tsig.go
```diff
@@ -0,0 +1,119 @@
...
+	// Make sure the response has an EDNS OPT RR if the request had it.
+	// Otherwise ScrubWriter would append it *after* TSIG, making it a non-compliant DNS message.
+	state := request.Request{Req: r.req, W: r.ResponseWriter}
+	state.SizeAndDo(m)
+
+	if r.reqTSIG != nil {
+		var repTSIG dns.TSIG
+		repTSIG.Hdr = dns.RR_Header{Name: r.reqTSIG.Hdr.Name, Rrtype: dns.TypeTSIG, Class: dns.ClassANY}
+		repTSIG.Algorithm = r.reqTSIG.Algorithm
+		repTSIG.OrigId = m.MsgHdr.Id
```
jinmeiib, line 120: I suspect some other fields of `repTSIG` should be set here, including
- `OrigId` from `m.MsgHdr.Id`
-  `Error` from `r.reqTSIG.Error`
- `TimeSigned` appropriately depending on whether the error is `RcodeBadTime` or not (and, the "bad time" case may have to be detected in `TSIGServer.ServeDNS`)

chrisohaver: Thanks for catching that! I need to copy error resulting from tsig validation failure forward to the response.

`OrigId` is set to `m.MsgHdr.Id`

IIRC, `TimeSigned` gets overwritten with the current time when the MAC is calculated and written to the client in the base responsewriter. I'll take a second look at that.

chrisohaver: Ah ... RFC 2854 4.5.2. ... time error handling is more fiddly than I initially thought. 

> 4.5.2. TIME check and error handling
>
>  If the server time is outside the time interval specified by the
   request (which is: Time Signed, plus/minus Fudge), the server MUST
   generate an error response with RCODE 9 (NOTAUTH) and TSIG ERROR 18
   (BADTIME).  The server SHOULD also cache the most recent time signed
   value in a message generated by a key, and SHOULD return BADTIME if a
   message received later has an earlier time signed value.  A response
   indicating a BADTIME error MUST be signed by the same key as the
   request.  **It MUST include the client's current time in the time
   signed field, the server's current time (a u_int48_t) in the other
   data field, and 6 in the other data length field.**  This is done so
   that the client can verify a message with a BADTIME error without the
   verification failing due to another BADTIME error.  The data signed
   is specified in [4.3].  The server SHOULD log the error.

chrisohaver:
I'm not sure about the bad time error response handling ... Specifically about encoding the "server time" into the OtherData string field.  I'm assuming that the byte order should be network byte order (as TimeSigned is).

jinmeiib:
> I'm not sure about the bad time error response handling ... Specifically about encoding the "server time" into the OtherData string field. I'm assuming that the byte order should be network byte order (as TimeSigned is).

It should be in network byte order.  I'll share sample code with you separately.

chrisohaver:
Reading over RFC 8945, I see there are very many special cases that this plugin does not handle, in particular protocols for error responses that are not followed here.   I'll close this PR for now and and re-open after getting this PR in a better shape for review.  I think leaving this open for review in its current state will be a distraction, so I'm closing it for now to avoid wasting any reviewer's time.

chrisohaver:
Re-opening this now.  I have proposed fixes to the error response cases to be fixed in the dns lib. https://github.com/miekg/dns/pull/1316

Also added to this PR, changes to the transfer plugin to pull in TSIG secrets and set them in the `dns.Transfer` instance when sending the transfer response.

There are some dangling threads, all related to forwarding TSIG messages that I still need to address.  Specifically, if a request's key is unknown, if we forward the request upstream the TSIG RR must be left completely intact.
Also, if a request's key is known, and validates OK, then the forward should b sign the outgoing request using a key/algo chosen by forward (i.e. new option). And there is a third item: a special case related to responses with the AD flag.
These are all covered in https://datatracker.ietf.org/doc/html/rfc8945#section-5.5

But I think we can just list those as known issues/bugs/limitations for now.

plugin/tsig/tsig.go
```diff
@@ -0,0 +1,141 @@
...
+	// wrap the response writer so the response will be TSIG signed.
+	w = &restoreTsigWriter{w, r, tsigRR}
+
+	if err = w.TsigStatus(); err != nil {
+		log.Debugf("TSIG validation failed: %v %v", dns.TypeToString[state.QType()], err)
+		rcode = dns.RcodeNotAuth
+		switch err {
+		case dns.ErrSecret:
+			tsigRR.Error = dns.RcodeBadKey
+			// See RFC-2845: 4.7. Special considerations for forwarding servers
```
jinmeiib: RFC8945 Section 5.5 would be a better reference.  I'd also note that the entire (raw) DNS message would have to be kept and passed to the forwarding plugin; if, for example, the incoming message doesn't compress names, the forwarded message must not compress names either.  I suspect we'd need some more architectural extension to make it possible (like allowing plugins to access raw the incoming DNS message in addition to the "unpacked" `dns.Msg`), rather than just not using this plugin.

chrisohaver: This was more of a note to self that I forgot to remove...
  
Thanks though, I didn't think of the compression case. I'll add that to the "Bugs" section I'm working on.

chrisohaver:
Redoing manual testing after making some mods to the miekg/dns/tsig.go, this plugin stopped working as expected.  I'm trying to track down what's gone wrong.

Also looking into modifying the unit tests so they actually pass responses through the dns response writer that does the tsig signing.

chrisohaver:
Regarding the tsig secrets physically residing in the Corefile: Security wise, I think we should provide a means of loading keys from file.

chrisohaver:
> Redoing manual testing after making some mods to the miekg/dns/tsig.go, this plugin stopped working as expected. I'm trying to track down what's gone wrong.

OK - it's working now with latest commit with miekg/dns#1316.  I'd still like to get tests that validate the actual response signing/writing.  Would be nice if I can get it working as unit tests as opposed to end to end integration tests.

chrisohaver:
Added some integration tests.  
One positive test, and one BadKey test (which I expect to fail without miekg/dns#1316).

~Still need to work out how to test BadSig and BadTime errors.~ Done

chrisohaver:
> Regarding the tsig secrets physically residing in the Corefile: Security wise, I think we should provide a means of loading keys from file.

I have added this.  For file format, I chose to go with the format that `dig` accepts (which is also the "key" directive of named).  It's also the output of `tsig-keygen`.

chrisohaver:
There are some TODOs left here, but I think this PR is incrementally mergeable as is (once miekg/dns#1316 is included)

I don't want to overwhelm with a big PR.

TODOs:
* In _transfer_ plugin, sign zone change notifications.
* In _secondary_ plugin, sign zone transfer requests and validate TSIG status of response before updating zone.
* In _forward_, conform to RFC (covered in README).

chrisohaver:
Rebased with master, which includes miekg/dns#1316 fix merged in miekg/dns v1.1.45 tag.
The BADKEY response unit test passes now.

chrisohaver:
All tests passing.  I think the PR ready for final review/merge.

fcolista:
Hello!
Any update on this?
I'm eager to use this plugin!
Thanks.

chrisohaver:
> Hello!
> Any update on this?
> I'm eager to use this plugin!
> Thanks.

No updates, just awaiting further review and approval. It can take a long time before maintainers have time to review a new plugin.

chrisohaver:
It looks like no one has time to review this. I would like to merge it.  Any objections?

fcolista:
> It looks like no one has time to review this. I would like to merge it. Any objections?

Please go ahead!

PR #4959, opened 2021-11-04
plugin/file: Fix print tree error

victory460:

the print  by level of tree can not work correctly, if the tree has 4 nodes, it shoud print three line node , but it will print all node contain in the two line.
No.
No.
No.

### DIFF
--- a/plugin/file/tree/print.go
+++ b/plugin/file/tree/print.go
@@ -29,9 +29,9 @@ func (n *Node) print() {
 		}
 		if nodesInCurrentLevel == 0 {
 			fmt.Println()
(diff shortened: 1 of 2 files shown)


miekg:
can you add a test?

victory460:
> can you add a test?

Thank you very much for your review. I have now added test code for it.

victory460:
> > can you add a test?
> 
> Thank you very much for your review. I have now added test code for it.


Sorry, I accidentally closed this pull request by a mistake, then I mention one again.[plugin/file: Fix print tree error](https://github.com/coredns/coredns/pull/4962) @miekg

PR #4961, opened 2021-11-05
plugin/loadbalance: More consistent shuffling

### LINKED ISSUE #4960: If there are many service pods on the headless backend, the results of DNS lookup for a headless service will not be balanced
coredns has enabled loadbalance，the version is v1.8. If there are many service pods on the headless backend, the results of DNS lookup for a headless service will not be balanced.

For example, if 5 pods are mounted on the headless backend, the analysis results are balanced. If headless mounts 15 pods, the resolution result of dns will always be about 2 times higher than the other pods.

test method：
for i  in {1..10000};do  curl -vI  xxxx.svc.cluster.local cluster.local  >> /tmp/abc  2>&1  ;done
The analysis results of 5 pods in the backend：
 sh-4.2# cat /tmp/abc  | grep "Trying" |  egrep -o   '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}' | sort | uniq -c | sort -nr
   2128 172.20.8.54
   2037 172.20.8.55
   1957 172.20.8.53
   1943 172.20.5.33
   1935 172.20.5.32

The analysis results of 15 pods in the backend：
sh-4.2# cat /tmp/efg  | grep "Trying" |  egrep -o   '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}' | sort | uniq -c | sort -nr
   1338 172.20.8.54
    659 172.20.8.53
    654 172.20.3.100
    653 172.20.3.99
    648 172.20.3.102
    637 172.20.3.103
    628 172.20.3.95
    622 172.20.5.33
    613 172.20.3.94
    612 172.20.3.97
    611 172.20.3.101
    596 172.20.3.98
    595 172.20.3.96
    569 172.20.8.55
    565 172.20.5.32

172.20.8.54 response is much higher than other requested ip

chrisohaver:

The random shuffle algorithm is ... 

Given N records: Randomly select two records. Swap them. Repeat for N*(1-4) times ... 

But the randomized loop exit condition gets re-randomized every iteration.  The result is that it will shuffle less than intended - usually exiting after swapping a number of times slightly higher than the number of records (not 1-4 times number of records, closer to 1 time + a few extra swaps).  This would have a more pronounced effect for longer record sets, that few extra swaps being less significant amount of extra shuffling. Under-shuffling results in a higher incident of the first record not being shuffled.

Fixing the random exit condition to be the same for all iterations lessens the problem, but still under-shuffles slightly since 1:4 times it will shuffle only N times for N records (resulting in a high incident of the first record not being shuffled).

Observationally, fixing the the number of shuffles to 3N seems to distribute the records pretty evenly. 2N still favored the first record significantly.

Edit: A simpler algorithm works better and requires less swaps:  iterating through each record once randomly selecting a record in the remainder of the list, and swapping them.

#4960

### DIFF
--- a/plugin/loadbalance/loadbalance.go
+++ b/plugin/loadbalance/loadbalance.go
@@ -61,7 +61,7 @@ func roundRobinShuffle(records []dns.RR) {
 			records[0], records[1] = records[1], records[0]
 		}
 	default:
-		for j := 0; j < l*(int(dns.Id())%4+1); j++ {
+		for j := 0; j < l*3; j++ {
 			q := int(dns.Id()) % l
 			p := int(dns.Id()) % l
 			if q == p {
--- a/plugin/loadbalance/loadbalance.go
+++ b/plugin/loadbalance/loadbalance.go
@@ -61,13 +61,12 @@ func roundRobinShuffle(records []dns.RR) {
 			records[0], records[1] = records[1], records[0]
 		}
 	default:
-		for j := 0; j < l*3; j++ {
-			q := int(dns.Id()) % l
-			p := int(dns.Id()) % l
-			if q == p {
-				p = (p + 1) % l
+		for j := 0; j < l; j++ {
+			p := j + (int(dns.Id()) % (l-j))
+			if j == p {
+				continue
 			}
-			records[q], records[p] = records[p], records[q]
+			records[j], records[p] = records[p], records[j]
 		}
 	}
 }


chrisohaver:
Might be cheaper to alter the algorithm to instead of randomly pick two records to swap, iterate through each, and swap with a randomly determined other record.  This way each record gets moved at least once in with N swaps.

chrisohaver:
The latest commit updates the algorithm to iterate through each record once randomly selecting a record in the remainder of the list, and swap them.

This results in a more even distribution, and requires less swaps.

miekg (approved):

ashwin-subramanian (commented):
Lgtm

PR #4962, opened 2021-11-05
plugin/file: Fix print tree error

victory460:

The print method designed in the coredns is a perfect debugging tool, but I found a small problem. The steps to reproduce this problem:

https://github.com/coredns/coredns/blob/4c0eacb6574786c8b3db1e5f068ebce96f7545e1/plugin/file/lookup_test.go#L135-L141
Just need to add *zone.Tree.Print()* to L140, like following:

```
func TestLookup(t *testing.T) {
	zone, err := Parse(strings.NewReader(dbMiekNL), testzone, "stdin", 0)
	if err != nil {
		t.Fatalf("Expected no error when reading zone, got %q", err)
	}
	zone.Tree.Print() //add this
	fm := File{Next: test.ErrorHandler(), Zones: Zones{Z: map[string]*Zone{testzone: zone}, Names: []string{testzone}}}
```

Print result before fix,  it just print two row:
```
=== RUN   TestLookup
dname.miek.nl. 
a.miek.nl. www.miek.nl. miek.nl. archive.miek.nl. mx.miek.nl. *.y.miek.nl. ext-cname.miek.nl. srv.miek.nl. *.x.miek.nl. 
```
Print result after fix:
```
=== RUN   TestLookup
dname.miek.nl. 
a.miek.nl. www.miek.nl. 
miek.nl. archive.miek.nl. mx.miek.nl. *.y.miek.nl. 
ext-cname.miek.nl. srv.miek.nl. *.x.miek.nl. 
```
In summary, the print by level of tree can not work correctly,  if the tree has 4 nodes,  it shoud print three row node,  but it print all node in the two row now. 
No.
No.
No.

### DIFF
--- a/plugin/file/tree/print.go
+++ b/plugin/file/tree/print.go
@@ -29,9 +29,9 @@ func (n *Node) print() {
 		}
 		if nodesInCurrentLevel == 0 {
 			fmt.Println()
+			nodesInCurrentLevel = nodesInNextLevel
+			nodesInNextLevel = 0
 		}
-		nodesInCurrentLevel = nodesInNextLevel
-		nodesInNextLevel = 0
 	}
 	fmt.Println()
 }
--- a/plugin/file/tree/print_test.go
+++ b/plugin/file/tree/print_test.go
@@ -1,0 +1,102 @@
+package tree
+
+import (
+	"net"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/miekg/dns"
+)
+
+func TestPrint(t *testing.T) {
+	rr1 := dns.A{
+		Hdr: dns.RR_Header{
+			Name:     dns.Fqdn("server1.example.com"),
+			Rrtype:   1,
+			Class:    1,
+			Ttl:      3600,
+			Rdlength: 0,
+		},
+		A: net.IPv4(10, 0, 1, 1),
+	}
+	rr2 := dns.A{
+		Hdr: dns.RR_Header{
+			Name:     dns.Fqdn("server2.example.com"),
+			Rrtype:   1,
+			Class:    1,
+			Ttl:      3600,
+			Rdlength: 0,
+		},
+		A: net.IPv4(10, 0, 1, 2),
+	}
+	rr3 := dns.A{
+		Hdr: dns.RR_Header{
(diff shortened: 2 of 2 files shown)


chrisohaver:
@victory460, you can push new commits to an existing PR without closing and re-opening a new PR for each change.

re: #4945 #4947 #4948 #4955 #4959

victory460:
> @victory460, you can push new commits to an existing PR without closing and re-opening a new PR for each change.
> 
> re: #4945 #4947 #4948 #4955 #4959

OK. Thanks.

chrisohaver (changes requested):
The test provided doesn't have any fail conditions - so unless it crashes, it will always pass.

The unit test should fail if tested against the current code in master, and pass when tested against the fixed code in the PR.

victory460:
> The test provided doesn't have any fail conditions - so unless it crashes, it will always pass.
> 
> The unit test should fail if tested against the current code in master, and pass when tested against the fixed code in the PR.

Thank you for your review:smile: . According to your suggestions, I updated the unit test code.
If you have any other suggestions, I'm willing to improve it. Thanks. @chrisohaver

miekg:
1 more request. Can you put in the git commit what printed wrong now and how that looks with this fix?

victory460:
> 1 more request. Can you put in the git commit what printed wrong now and how that looks with this fix?

Thank you for your review gratefully @miekg . I updated the git commit content and the description of this PR.

yongtang:
@victory460 I noticed that your commit's email is not associated with your GitHub ID. I would suggest you to update the email of the commit, or add the email to GitHub's associated email address as otherwise your work will not be attributed to you.

victory460:
> @victory460 I noticed that your commit's email is not associated with your GitHub ID. I would suggest you to update the email of the commit, or add the email to GitHub's associated email address as otherwise your work will not be attributed to you.

Thanks a lot！According to your suggestion, I added my commit's email address to my github account in email setting now.

Rosalieww:
We also found the bug in this print tree code, can you merge the PR when you have time for it? Thanks.   @miekg @chrisohaver

chrisohaver (approved):

Issue #4943, opened 2021-10-28
plugin/route53 allow configuring of endpoint URL


56KBs:
**What would you like to be added**:
The ability to configure the `endpoint` in the Route53 plugin AWS client.

**Why is this needed**:
We are making use of the Route53 plugin in order to aid our integration tests, using a mocked version of Route53 served via [localstack](https://github.com/localstack/localstack), sadly, localstack does not provide for free the ability to serve the Route53, when combining it with coredns it does allow this. I have currently locally modified the `setup.go` for the Route53 plugin to support this in order to allow us to use the locally hosted Route53 instead of the true Route53.

Aut0R3V:
I'd like to work on this ,can someone provide a little bit of guidance on how to get this done?

chrisohaver:
> can someone provide a little bit of guidance on how to get this done?

@Aut0R3V, Add a new option (e.g. `endpoint`) to parsing in `setup.go`, and assign it to (probably) `Endpoint` in the `aws.Config`. Also add unit tests, and update README.

Aut0R3V:
Thank you!

PR #4963, opened 2021-11-05
plugin/route53: Configurable AWS Endpoint

56KBs:

Provide the ability to configure the AWS endpoint

#4943

Document the new option - Changes made in branch

No

### DIFF
--- a/plugin/route53/setup.go
+++ b/plugin/route53/setup.go
@@ -26,8 +26,8 @@ var log = clog.NewWithPlugin("route53")
 func init() { plugin.Register("route53", setup) }
 
 // exposed for testing
-var f = func(credential *credentials.Credentials) route53iface.Route53API {
-	return route53.New(session.Must(session.NewSession(&aws.Config{Credentials: credential})))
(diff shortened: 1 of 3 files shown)


plugin/route53/setup.go
```diff
@@ -26,8 +26,8 @@ var log = clog.NewWithPlugin("route53")
 func init() { plugin.Register("route53", setup) }
 
 // exposed for testing
-var f = func(credential *credentials.Credentials) route53iface.Route53API {
-	return route53.New(session.Must(session.NewSession(&aws.Config{Credentials: credential})))
+var f = func(credential *credentials.Credentials, endpoint *string) route53iface.Route53API {
```
dilyevsky, line 29: nit: doesn't seem like it _needs_ to be a pointer since you're always passing empty string as default?

dilyevsky: hm actually looks like both nil and empty string are acceptible as defaults per aws docs.

dilyevsky (approved):

zouyee (approved):

yongtang (approved):

PR #4971, opened 2021-11-09
plugin/loadbalance: remove cname case in roundRobin

victory460:

https://github.com/coredns/coredns/blob/06e3b2e85994c180e1f8febebae1b11370789596/plugin/loadbalance/loadbalance.go#L34-L48

As shown in the code above, the cname record is obtained, but random sorting of cname is meaningless as @chrisohaver said.

No.
No.
No.

### DIFF
--- a/plugin/loadbalance/loadbalance.go
+++ b/plugin/loadbalance/loadbalance.go
@@ -26,14 +26,11 @@ func (r *RoundRobinResponseWriter) WriteMsg(res *dns.Msg) error {
 }
 
 func roundRobin(in []dns.RR) []dns.RR {
-	cname := []dns.RR{}
 	address := []dns.RR{}
 	mx := []dns.RR{}
 	rest := []dns.RR{}
 	for _, r := range in {
 		switch r.Header().Rrtype {
-		case dns.TypeCNAME:
-			cname = append(cname, r)
 		case dns.TypeA, dns.TypeAAAA:
 			address = append(address, r)
 		case dns.TypeMX:
@@ -46,9 +43,8 @@ func roundRobin(in []dns.RR) []dns.RR {
 	roundRobinShuffle(address)
 	roundRobinShuffle(mx)
 
-	out := append(cname, rest...)
-	out = append(out, address...)
-	out = append(out, mx...)
+	out := append(address, mx...)
+	out = append(out, rest...)
 	return out
 }
 


chrisohaver:
Unlike some other record types, such as A or AAAA, CNAME record names must be unique.  IOW, if you query for `test.example.com` you can only ever get one answer CNAME record back with a name of `test.example.com`.

You _can_ get multiple CNAMEs of _different_ names in a response, where the multiple CNAME records form a chain.  But, most client resolvers are sensitive to CNAME order, and will get confused if a CNAME targets a record earlier in the list.  Even if clients could handle out of order CNAME chains, randomizing the chain order would not provide a randomized answer.

So, I don't think this is necessary.

If you actually encountered a scenario where you need to randomize CNAME order, can you describe it in detail?

chrisohaver (commented):
**original PR title: Plugin/loadbalance: add cname shuffle support plugin/loadbalance**

Unlike some other record types such as A or AAAA, CNAME record names must be unique.  IOW, if you query for `test.example.com` you can only ever get one CNAME record back with a name of `test.example.com`.

You _can_ get multiple CNAMEs of _different_ names in a response, where the multiple CNAME records form a chain.  But, most client resolvers are sensitive to CNAME order, and will get confused if a CNAME targets a record earlier in the list.  Even if clients could handle out of order CNAME chains, randomizing the chain order would not provide a randomized answer.

So, I don't think this is necessary.

If you actually encountered a scenario where you need to randomize CNAME order, can you describe it in detail?

victory460:
> Unlike some other record types such as A or AAAA, CNAME record names must be unique. IOW, if you query for `test.example.com` you can only ever get one CNAME record back with a name of `test.example.com`.
> 
> You _can_ get multiple CNAMEs of _different_ names in a response, where the multiple CNAME records form a chain. But, most client resolvers are sensitive to CNAME order, and will get confused if a CNAME targets a record earlier in the list. Even if clients could handle out of order CNAME chains, randomizing the chain order would not provide a randomized answer.
> 
> So, I don't think this is necessary.
> 
> If you actually encountered a scenario where you need to randomize CNAME order, can you describe it in detail?

Thank you for your review. The scenario we considered at the beginning was to use CDN services from multiple companies for load balancing. You are right, so I modified this PR that to move case dns.TypeCNAME to the default: branch.

plugin/loadbalance/loadbalance.go
```diff
@@ -46,9 +43,8 @@ func roundRobin(in []dns.RR) []dns.RR {
 	roundRobinShuffle(address)
 	roundRobinShuffle(mx)
 
-	out := append(cname, rest...)
```
chrisohaver, line 49: Removing this will place CNAMEs after their targets, 
which will break dns resolution for some/most clients.

chrisohaver (changes requested):
It’s unclear what bug you are aiming to fix. Can you give an example of a query with the current unexpected response, and what the expected response should be?

plugin/loadbalance/loadbalance.go
victory460: Thanks！

PR #4972, opened 2021-11-09
plugin/multicluster: Adding a new plugin supporting Multicluster DNS

vanekjar:

* A plugin implementing K8s multi-cluster services DNS spec from https://github.com/kubernetes/enhancements/pull/2577

* Most of the code is forked from the existing `kubernetes` plugin as multi-cluster DNS spec is very similar to the existing K8s DNS spec.

* Moving code from the separate repository at https://github.com/vanekjar/coredns-multicluster


chrisohaver:
Thanks! How far along is the KEP?  It's labeled "Work in Progress".

For now it makes sense to [register this plugin as an external plugin](https://coredns.io/2017/07/23/add-external-plugins/).  This is easy to do, since the plugin is already set up that way, and documented in the the README as such. 

When we include this in CoreDNS,  we need to make some effort to reduce the amount of code duplication, perhaps by packaging shared code, or by extending the existing _kubernetes_ plugin to support the proposed multi-cluster schema instead of implementing it as a separate plugin.

chrisohaver (commented):
Since so much of it is duplicated from the existing _kubernetes_ plugin, it would be very helpful to reviewers for you to highlight what has changed and what has been added.

vanekjar:
Chris, thank you for the feedback.

> For now it makes sense register this plugin as an external plugin. This is easy to do, since the plugin is already set up that way, and documented in the the README as such.

Totally makes sense. That's probably a sufficient solution until the KEP is finalized. But I believe that eventually this plugin should become an integral part of coredns alongside of existing kubernetes plugin. Just opened a separate PR for the external plugin (https://github.com/coredns/coredns.io/pull/241).

> When we include this in CoreDNS, we need to make some effort to reduce the amount of code duplication, perhaps by packaging shared code, or by extending the existing kubernetes plugin to support the proposed multi-cluster schema instead of implementing it as a separate plugin.

Right, the code duplication is annoying. The idea was to let both plugins evolve separately for now without causing any potential issues for the stable kubernetes plugin. But I agree that extracting shared code would be the right thing to do. I don't think plugins can be merged as the functionality is almost mutually exclusive - they track different resources (`Service` vs `ServiceImport`) and manage DNS requests for different zones (`cluster.local` vs `clusterset.local`).

miekg:
If we are going to carry this, the duplicated code should all be factored out. This will probably also benefit things like forwardCRD. Bonus points for making that factored out code reduce the load on the k8s control plane

chrisohaver:
@miekg Would you be OK with this external plugin living in a new repo in the coredns project? e.g. `coredns/multicluster`

miekg:
another repo under coredns is fine with me. Refactoring the k8s api conn. has come up again and again, so we should probably do that at some point....

runakash:
> another repo under coredns is fine with me.

@miekg / @chrisohaver Sounds good to us. Let us know once the repo is created. I can help with migrating the codebase.

vanekjar:
I hope we'll be able to transfer the existing repository instead creating a new one from scratch. That way all exisitng links can be correctly redirected to the new repo.

Adding link describing the process:
https://docs.github.com/en/repositories/creating-and-managing-repositories/transferring-a-repository

vanekjar:
Repo successfully transfered to `coredns` org, closing this PR as obsolete.

https://github.com/coredns/multicluster

Issue #4973, opened 2021-11-09
Add ability to add new server transport protocols


jdamick:
**What would you like to be added**:

The ability to add new server implementations for either existing transports (dns, https, etc.) and add new server transport (quic, or other experimental servers).

**Why is this needed**:

Currently it is not possible to add or override server transports without patching existing source.

chrisohaver:
Closing this issue, as a PR was opened at the same time.  I think it makes sense to have the discussion in one place, not split between an issue and a PR.

jdamick:
Ok, I didn't want to just submit an unsolicited PR with no context.

chrisohaver:
No problem. The context can go in the PR description.

IMO, The idea of raising an issue is to have a discussion and gather input from the community about it before agreeing upon and submitting a solution.  If you already have a solution built, there is little purpose to opening a separate issue.

chrisohaver:
I have copied the description here to the PR description.

PR #4974, opened 2021-11-09
Add the ability to add new server transport implemementations or override existing ones

jdamick:
Adds the ability to add new server implementations for either existing transports (dns, https, etc.) and add new server transport (quic, or other experimental servers). The pattern followed is similar to that of plugins, maintaining backward compatibility
with existing consts and functions.

#4973

documentation related to adding new servers, similar to plugins.

I don't believe so.


jdamick:
From issue:

**What would you like to be added:**

The ability to add new server implementations for either existing transports (dns, https, etc.) and add new server transport (quic, or other experimental servers).

**Why is this needed:**

Currently it is not possible to add or override server transports without patching existing source.

server.cfg
```diff
@@ -0,0 +1,22 @@
...
+# Or
+#     <plugin-name>:<fully-qualified-package-name>.<server creator function>
+#
+# External server example:
+#
+# quic:4242:github.com/coredns/coredns/dnsserver/experimental.NewServerQuic
+# Local plugin example:
+# tls:853:NewServerTLS
+
+tls:853:NewServerTLS
```
johnbelamaric, line 19: If you instead have a convention that it is always `NewServer`, and instead put each different server creation routine in its own package, then you can generate the correct imports and calls to in `zservers.go` such that external servers become possible. As it is here, wouldn't you need new servers defined in the `dnsserver` package?

johnbelamaric (commented):
Interesting idea. Any specific use cases?

plugin/pkg/transport/ztransport.go
```diff
@@ -0,0 +1,22 @@
+// generated by servers_generate.go; DO NOT EDIT
+
+package transport
+
+// These transports are supported by CoreDNS.
+const (
```
johnbelamaric, line 6: what's the point of these if they are generated anyway?

server.cfg
johnbelamaric: so, for example

```
tls:853:tls
dns:53:dns
foo:10000:github.com/myrepo/fooserver
```

would become

```
import "github.com/coredns/coredns/core/dnsserver/tlsserver" tlssrvr
import "github.com/coredns/coredns/core/dnsserver/dnsserver" dnssrvr
import "github.com/myrepo/fooserver" foosrvr

...
...

	RegisterTransportServer(transport.TLS, transport.TLSPort, func(addr string, group []*Config) (caddy.Server, error) {
		return tlssrv.NewServer(addr, group)
	})

	RegisterTransportServer(transport.DNS, transport.DNSPort, func(addr string, group []*Config) (caddy.Server, error) {
		return dnssrv.NewServer(addr, group)
	})

	RegisterTransportServer(transport.FOO, transport.FOOPort, func(addr string, group []*Config) (caddy.Server, error) {
		return foosrv.NewServer(addr, group)
	})

...
....

jdamick: Yes, I thought about that but I was trying to minimize the changes. Another thought to make it cleaner is to have an interface for it. Do you think I should  make any of these changes?

jdamick:
> Interesting idea. Any specific use cases?

Yes, I have 2 cases.  One is related to optimizations and framing around the DNS protocol, another is a custom protocol transport. 

Currently I have to maintain patches and apply them to add or override the servers, whereas if the servers could be added like plugins it would remove the patching pain.

plugin/pkg/transport/ztransport.go
jdamick: Do you mean maintaining the constants?  I was trying to maintain compatibility as much as possible.

chrisohaver:
> I have 2 use cases. One is related to optimizations and framing around the DNS protocol, 

Are these optimizations that could be integrated directly into the existing implementation? i.e. merged into CoreDNS? Or are they proprietary?

> another is a custom protocol transport.

I wonder how common this use case would be.  I would think not common enough to justify maintaining a plugin framework for it.

> Currently I have to maintain patches and apply them ... pain.

The code around these areas is relatively static - so I'd imagine that patching would _usually_ be more tedious than painful.

jdamick:
> > I have 2 use cases. One is related to optimizations and framing around the DNS protocol,
> 
> Are these optimizations that could be integrated directly into the existing implementation? i.e. merged into CoreDNS? Or are they proprietary?

They aren't really applicable to general use cases. But I will make a PR if there any areas that can be.

 
> > another is a custom protocol transport.
> 
> I wonder how common this use case would be. I would think not common enough to justify maintaining a plugin framework for it.

For a few other data points, if one wants to add experimental quic support now they will have to patch. If they want to add http (no s), if they have tls offloading, they have to patch.  

> 
> > Currently I have to maintain patches and apply them ... pain.
> 
> The code around these areas is relatively static - so I'd imagine that patching would _usually_ be more tedious than painful.

Not to say it's difficult, but unless you are overriding an existing protocol scheme, there are several files that have to be patched to add the new constants and update the multiple switch statements. I suppose if these were consolidated in 1 file or package it might be easier.

I've been looking at @johnbelamaric's suggestion of repackaging the servers to see how that plays out.  I'll update this PR here with that option, unless there is an objection?

chrisohaver:
> For a few other data points, if one wants to add experimental quic support now they will have to patch. If they want to add http (no s), if they have tls offloading, they have to patch.

Thanks.  HTTP and QUIC are good concrete examples.  I _think_ there has been opposition to adding plain HTTP in the past, so that supports the case for making it easier to add new protocols.

miekg:
I wonder why this is needed, quic will be added then it's ready and there is a good lib. I don't see why generic support for adding new stuff on this level should be added. Patching the source is also a working and good solution

jdamick:
> Patching the source is also a working and good solution

I dug into abstracting out the servers in separate packages, that appears to involve a lot of changes. Many of the servers reach in to use the Lock/Unlock in the base Server. There are also some areas that reach into the servers[tcp] and change it.  These can be fixed but the scope of the change increases.

@miekg  Would you be more open to refactoring the just the servers for now so they can be separate packages, then talk about the plugin of servers?

jdamick:
Based on feedback, i'll just close this for now and continue patching.

PR #4977, opened 2021-11-10
plugin/k8s_external: implement zone transfers

### LINKED ISSUE #4803: Implement zone transfer for k8s_external plugin

Implement zone transfer for the _k8s_external_ plugin.

**Why is this needed**:

To do zone transfers of zones served by  _k8s_external_.

chrisohaver:

Implements zone transfers for the k8s_external plugin. This PR does not implement Notifies.

#4803

This PR adds a usage example to the k8s_external README

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -47,13 +47,20 @@ type dnsController interface {
 
 	// Modified returns the timestamp of the most recent changes
 	Modified() int64
+
+	// ExtModified returns the timestamp of the most recent changes to
+	// services with external facing IP addresses
+	ExtModified() int64
 }
 
 type dnsControl struct {
-	// Modified tracks timestamp of the most recent changes
+	// modified tracks timestamp of the most recent changes
 	// It needs to be first because it is guaranteed to be 8-byte
 	// aligned ( we use sync.LoadAtomic with this )
 	modified int64
+	// extModified tracks timestamp of the most recent changes to
+	// services with external facing IP addresses
+	extModified int64
 
 	client kubernetes.Interface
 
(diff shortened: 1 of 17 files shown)


plugin/kubernetes/controller.go
```diff
@@ -646,17 +659,83 @@ func endpointsEquivalent(a, b *object.Endpoints) bool {
...
+
+	return intSvc, extSvc
+}
+
 func (dns *dnsControl) Modified() int64 {
 	unix := atomic.LoadInt64(&dns.modified)
 	return unix
 }
 
+func (dns *dnsControl) ExtModified() int64 {
```
miekg: why does this need *another* exported method?

chrisohaver: I was probably thinking it would be less disruptive than changing the signature of `Modified()` to handle both...  but you're right, I think it's better to keep the number of methods in the interface sh short as possible.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] plugin/k8s_ex..." ]
>
>
>In plugin/kubernetes/controller.go:
>
>>  func (dns *dnsControl) Modified() int64 {
>        unix := atomic.LoadInt64(&dns.modified)
>        return unix
> }
>
>+func (dns *dnsControl) ExtModified() int64 {
>
>I was probably thinking it would be less disruptive than changing the signature
>of Modified() to handle both... but you're right, I think it's better to keep
>the number of methods in the interface sh short as possible.

Think we can just push changes through here

plugin/kubernetes/controller.go
chrisohaver: Done - Consolidated the functionality to the existing `Modified` function.

chrisohaver:
ping

PR #4980, opened 2021-11-12
plugin/log: Add option to log slow dns responses

ScheererJ:
This change adds a flag to the log plugin to log slow dns responses.
The duration of what is defined as slow is configurable, default is
100ms.
Slow dns responses are important to keep track of as dns clients
may have a low timeout and a late response may therefore be considered
as no response whatsoever.

None.
The log plugin documentation is updated as part of the pull request.
No, the change should be completely compatible.

### DIFF
--- a/plugin/log/README.md
+++ b/plugin/log/README.md
@@ -30,11 +30,12 @@ log [NAMES...] [FORMAT]
   for the Common Log Format. You can also use `{combined}` for a format that adds the query opcode
   `{>opcode}` to the Common Log Format.
 
-You can further specify the classes of responses that get logged:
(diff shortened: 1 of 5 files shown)


chrisohaver:
Thanks.  I think the use case is valid, but probably better to track slow queries via existing Prometheus metrics than via log output.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] plugin/log: A..." ]
>Thanks. I think the use case is valid, but probably better to track slow
>queries via Prometheus metrics than via log output.

agreed, and this should alrady be tracked there.

Issue #4984, opened 2021-11-15
Information Disclosure Issue: Add feature to disable wildcard lookups


micahhausler:
Opening this as a public issue, as there are already [public discussions](https://twitter.com/raesene/status/1458724150210662400) and [integration into penetration testing tooling](https://twitter.com/jaybeale/status/1459201234066313218).

CoreDNS's wildcard feature (specifically [wildcards in Kubernetes](https://github.com/coredns/coredns/tree/163ac033aee0348ca5a4d718d234adb1a1eedce7/plugin/kubernetes#wildcards)) present an information disclosure.

CoreDNS has several plugins that support wildcard lookups, including the `file`, `etcd`, and `kubernetes` plugins. In environments such as Kubernetes where Service Discovery is desired, full enumeration of Services, Namespaces, and Endpoint IPs is possible even when Kubernetes API permissions or authentication is denied. 

Administrators can [disable the mounting of Kubernetes Service Account tokens](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server) and deny the pod any identity in the Kubernetes API. They can [disable Docker links compatibility](https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables) to limit same-namespace service discovery in a pod. They can also use [`spec.dnspolicy` or `spec.dnsConfig`](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy) to use alternate nameservers instead of CoreDNS, but that prevents any in-cluster service discovery. 

For an administrator who considers namespaces, service names, or pod IPs (or the number of pod IPs behind a service) to be confidential and still wants service discovery, they have no way to disable wildcard service listing in their cluster.

This is particularly concerning if anyone were to run untrusted code in a cluster where that code executes lookups against CoreDNS. Barriers such as node-affinity/taints and tolerations, or encapsulation mechanisms like [Google's gVisor](https://github.com/google/gvisor) or [Kata containers](https://katacontainers.io/) are of no benefit if DNS requests are sent to CoreDNS. 

If this were considered a vulnerability, my initial CVSS estimate would rate it as a [Medium 4.3
](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:A/AC:L/PR:N/UI:N/S:U/C:L/I:N/A:N)
**What would you like to be added**:

[...]

PR #4985, opened 2021-11-15
plugin/kubernetes: Add wildcards option

chrisohaver:

Add option to enable/disable wildcard queries. Default for now is `enabled`.

#4984

included

No.  But if we default to `disabled`, it would be.

### DIFF
--- a/plugin/kubernetes/README.md
+++ b/plugin/kubernetes/README.md
@@ -42,6 +42,7 @@ kubernetes [ZONES...] {
     noendpoints
     fallthrough [ZONES...]
     ignore empty_service
+    wildcards enabled|disabled
 }
 ```
 
@@ -101,7 +102,9 @@ kubernetes [ZONES...] {
 * `ignore empty_service` returns NXDOMAIN for services without any ready endpoint addresses (e.g., ready pods).
   This allows the querying pod to continue searching for the service in the search path.
   The search path could, for example, include another Kubernetes cluster.
-
+* `wildcards enabled|disabled` enables or disables the wildcard query feature (see Wildcards section below).
+  Default is enabled.
+  
 Enabling zone transfer is done by using the *transfer* plugin.
 
 ## Startup
--- a/plugin/kubernetes/external.go
+++ b/plugin/kubernetes/external.go
@@ -69,7 +69,7 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
 
 		for _, ip := range svc.ExternalIPs {
 			for _, p := range svc.Ports {
-				if !(match(port, p.Name) && match(protocol, string(p.Protocol))) {
+				if !(k.match(port, p.Name) && k.match(protocol, string(p.Protocol))) {
 					continue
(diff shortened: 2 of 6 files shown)


chrisohaver:
IMO, it would be nice if we could remove the feature entirely.  But if we do, we should give users a very long runway.
So, for now default to `enabled`, and make it an option to disable.
Next minor release perhaps, make it default to `disabled`.
In the future, maybe we can remove wildcard query support completely.

miekg:
Read the issue. Agree with Chris this should be removed. As it's
intertwined with etcd we should remove it there as well IMO.

Probably a knob know default to true, then false and the remove the code?

plugin/kubernetes/README.md
```diff
@@ -101,7 +102,9 @@ kubernetes [ZONES...] {
 * `ignore empty_service` returns NXDOMAIN for services without any ready endpoint addresses (e.g., ready pods).
   This allows the querying pod to continue searching for the service in the search path.
   The search path could, for example, include another Kubernetes cluster.
-
+* `wildcards enabled|disabled` enables or disables the wildcard query feature (see Wildcards section below).
+  Default is enabled.
```
miekg, line 106: Hmm, another option that requires deprecating. I wonder can't we just rip the wildcard stuff out at once and be done with it?

chrisohaver:
> Hmm, another option that requires deprecating. I wonder can't we just rip the wildcard stuff out at once and be done with it?

This is the user friendly / long runway approach, which I tend to gravitate to.  We could also just remove it - the maintainer friendly approach.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] plugin/kubern..." ]
>This is the user friendly / long runway approach, which I tend to gravitate to.
>We could also just remove it - the maintainer friendly approach.

I've seen people use it on twitter, as nice debug feature, but we have axfr now if you
want this. I'm inclined to say deprecate this in one go, and be done with it. DNS as such
can use it; you specifically need to craft a query to use this.

chrisohaver:
>... As it's intertwined with etcd we should remove it there as well IMO. 

Wildcard behavior is part of SkyDNS, which etcd emulates ...

https://github.com/skynetservices/skydns#wildcards

... So I'm not sure if we should remove it from the etcd plugin.

johnbelamaric:
yeah, seems if someone is using that plugin it's probably what they want. if it's not, perhaps a *new* etcd plugin is needed that doesn't use that functionality

chrisohaver:
We could skip adding an option, deprecate wildcards now and  remove them in next release with #5019 + #5030.

That said, the advantage of adding an option (this PR) is that users could disable wildcards before waiting for the next minor release, if they are concerned about the information disclosure (#4984).

chrisohaver:
Closing this.  Just deprecating and removing wildcards is simpler.  In the mean time, users can disable wildcard queries in kubernetes using the suggestions outlined in https://github.com/coredns/coredns/issues/4984#issuecomment-969272093.

PR #4986, opened 2021-11-16
plugin/auto: Fix panic caused by config invalid reload value

gomakesix:
Signed-off-by: hack <reiachvowfllz@gmail.com>

As stated in the readme of the auto plugin:
_reload interval default is one minute. Value of 0 means to not scan for changes and reload._ 
Configured in corefile like this：

```
  auto miek.nl {
        directory .
        reload 0s
    }
```
This reload parameter is involved in the *setup.go* of the auto plugin：

```
ticker := time.NewTicker(a.loader.ReloadInterval)
```
However, time.NewTicker cannot receive a value of 0, it will cause panic：

```
func NewTicker(d Duration) *Ticker {
	if d <= 0 {
		panic(errors.New("non-positive interval for NewTicker"))
	}
```
Follow the rules in auto plugin readme: _value of 0 means to not scan for changes and reload_, we fixed it.
No.
No.
No.

### DIFF
--- a/plugin/auto/setup_test.go
+++ b/plugin/auto/setup_test.go
@@ -126,3 +126,52 @@ func TestAutoParse(t *testing.T) {
 		}
 	}
 }
+
+func TestSetupReload(t *testing.T) {
+	tests := []struct {
+		name    string
+		config  string
+		wantErr bool
+	}{
+		{
+			name: "reload valid",
+			config: `auto {
+				directory .
+				reload 5s
+			}`,
+			wantErr: false,
+		},
+		{
+			name: "reload disable",
+			config: `auto {
+				directory .
+				reload 0
+			}`,
+			wantErr: false,
+		},
+		{
+			name: "reload invalid",
+			config: `auto {
+				directory .
+				reload -1s
+			}`,
+			wantErr: true,
+		},
+		{
+			name: "reload invalid",
+			config: `auto {
+				directory .
+				reload
+			}`,
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctr := caddy.NewTestController("dns", tt.config)
+			if err := setup(ctr); (err != nil) != tt.wantErr {
+				t.Errorf("Error: setup() error = %v, wantErr %v", err, tt.wantErr)
+			}
+		})
+	}
+}
--- a/plugin/auto/setup.go
+++ b/plugin/auto/setup.go
@@ -1,6 +1,7 @@
 package auto
 
 import (
+	"errors"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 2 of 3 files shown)


chrisohaver (commented):
Thanks! Please add a unit test. And can you also check that the file plugin doesnt have the same issue. They should be able to be configured the same way.

gomakesix:
> Thanks! Please add a unit test. And can you also check that the file plugin doesnt have the same issue. They should be able to be configured the same way.

Thanks for your review, file plugin doesnt have the same issue. In addition, I will add unit test later.

chrisohaver:
> Thanks for your review, file plugin doesnt have the same issue. In addition, I will add unit test later.

Thanks for checking.  It does look like both _file_ and _auto_ will crash if no argument is provided to `reload` though.  If you can fix that too while you're at it, that would be cool.  I think it's closely related enough to fix in the same PR.

gomakesix:
> It does look like both _file_ and _auto_ will crash if no argument is provided to `reload` though.

We check that if file plugin without provided reload config, it work fine.

gomakesix:
> > Thanks for your review, file plugin doesnt have the same issue. In addition, I will add unit test later.
> 
> Thanks for checking. It does look like both _file_ and _auto_ will crash if no argument is provided to `reload` though. If you can fix that too while you're at it, that would be cool. I think it's closely related enough to fix in the same PR.

Thanks for your review @chrisohaver , we add a unit test of auto plugin now.

chrisohaver:
> >  It does look like both file and auto will crash _if no argument is provided to reload_ though. If you can fix that too while you're at it, that would be cool. I think it's closely related enough to fix in the same PR.
> 
> We check that if file plugin without provided reload config, it work fine.

This will panic ...

```
  auto miek.nl {
        directory .
        reload
    }
```

so will ...

```
  file example.db {
        reload
    }
```

Of course I understand this is a different panic than the `0s` panic - but IMO, it's closely related enough to fix in the same PR _if you have time_.

plugin/auto/setup_test.go
```diff
@@ -126,3 +126,44 @@ func TestAutoParse(t *testing.T) {
...
 }
+
+func TestSetupReload(t *testing.T) {
+	tests := []struct {
+		name    string
+		config  string
+		wantErr bool
+	}{
+		{
+			name: "Blacklist 1",
```
chrisohaver: These are odd names for the tests ... I'm not getting how "Blacklist" relates to the test performed.  How about "reload".  e.g. reload valid, reload disable, reload invalid, ...

gomakesix: Thanks a lot! I fixed it now.

gomakesix:
> > > It does look like both file and auto will crash _if no argument is provided to reload_ though. If you can fix that too while you're at it, that would be cool. I think it's closely related enough to fix in the same PR.
> > 
> > 
> > We check that if file plugin without provided reload config, it work fine.
> 
> This will panic ...
> 
> ```
>   auto miek.nl {
>         directory .
>         reload
>     }
> ```
> 
> so will ...
> 
> ```
>   file example.db {
>         reload
>     }
> ```
> 
> Of course I understand this is a different panic than the `0s` panic - but IMO, it's closely related enough to fix in the same PR _if you have time_.

👌,  fixed it.

chrisohaver (approved):
Thanks! LGTM

PR #4987, opened 2021-11-17
Changes to cover noerror/nosoa condition

ponienselvan:
When upstream responses doesn't include SOA to indicate NODATA  but with NOERROR code - CoreDNS treats this as NOERROR and picks least TTL from any record included in the response.

This is typical behavior in case of F5 Global Loadbalancer dns responses:

1. When all service members fail GLB returns no records and NOERROR
2. Combined with recursion by an upstream server - responses may contain CNAME section
3. CoreDNS uses TTL on CNAME for caching an NODATA condition

This set of patches treats this condition as NODATA 

See Above

Not Applicable

Not Applicable

### DIFF
--- a/plugin/pkg/response/typify.go
+++ b/plugin/pkg/response/typify.go
@@ -86,12 +86,21 @@ func Typify(m *dns.Msg, t time.Time) (Type, *dns.OPT) {
 		}
 	}
 
-	if len(m.Answer) > 0 && m.Rcode == dns.RcodeSuccess {
+	answerMatch:= false
+	for _, r :=  range m.Answer{
+		if(m.Question[0].Qtype==r.Header().Rrtype){
+			answerMatch=true
+		}
+	}
+	
+	
+	if len(m.Answer) > 0 && m.Rcode == dns.RcodeSuccess && answerMatch {
 		return NoError, opt
 	}
 
 	soa := false
 	ns := 0
+	
 	for _, r := range m.Ns {
 		if r.Header().Rrtype == dns.TypeSOA {
 			soa = true
@@ -101,10 +110,15 @@ func Typify(m *dns.Msg, t time.Time) (Type, *dns.OPT) {
 			ns++
 		}
 	}
-
+	
 	if soa && m.Rcode == dns.RcodeSuccess {
 		return NoData, opt
 	}
+
+	if len(m.Answer)>0 && !answerMatch && m.Rcode == dns.RcodeSucess && !soa {
+		return NoData, opt
+	}
+	
 	if soa && m.Rcode == dns.RcodeNameError {
 		return NameError, opt
 	}
--- a/plugin/pkg/response/typify.go
+++ b/plugin/pkg/response/typify.go
@@ -115,7 +115,7 @@ func Typify(m *dns.Msg, t time.Time) (Type, *dns.OPT) {
 		return NoData, opt
 	}
 
-	if len(m.Answer)>0 && !answerMatch && m.Rcode == dns.RcodeSucess && !soa {
+	if len(m.Answer)>0 && !answerMatch && m.Rcode == dns.RcodeSuccess && !soa {
 		return NoData, opt
 	}
 	
--- a/plugin/pkg/response/typify.go
+++ b/plugin/pkg/response/typify.go
@@ -86,21 +86,20 @@ func Typify(m *dns.Msg, t time.Time) (Type, *dns.OPT) {
 		}
 	}
 
-	answerMatch:= false
-	for _, r :=  range m.Answer{
-		if(m.Question[0].Qtype==r.Header().Rrtype){
-			answerMatch=true
+	answerMatch := false
+	for _, r := range m.Answer {
+		if m.Question[0].Qtype == r.Header().Rrtype {
+			answerMatch = true
 		}
 	}
-	
-	
+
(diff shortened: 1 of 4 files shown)


chrisohaver:
Can you provide an example of how the current behavior can be problematic?

ponienselvan:
> Can you provide an example of how the current behavior can be problematic?


Yes. In the absence of proper classification, core dns picks up TTL values from a CNAME record instead defaulting to minTTL. 


Consider below cases
**Healthy service instances are present**
$ dig conntester.example.com
;; QUESTION SECTION:
;conntester.example.com. IN   A
 
;; ANSWER SECTION:
conntester.example.com. **900** IN CNAME      conntester-conntester-glb.example.com.
conntester-conntester-glb.example.com. 20 IN A 192.168.1.2
conntester-conntester-glb.example.com. 20 IN A 192.168.1.3
conntester-conntester-glb.example.com. 20 IN A 192.168.1.4
conntester-conntester-glb.example.com. 20 IN A 192.168.1.5
conntester-conntester-glb.example.com. 20 IN A 192.168.1.6
 
**No Healthy service instances are present**
$ dig conntester.example.com #when there are no service ip functional
;; QUESTION SECTION:
;conntester.example.com. IN   A
 
;; ANSWER SECTION:
conntester.example.com. **900** IN CNAME      conntester-conntester-glb.example.com.
;; response does not include  A records indicating no healthy service ips for the service (but service exists)
;; No SOA/NS records, NOERROR as well

In the second case CoreDNS uses the TTL on CNAME for caching and it takes 15 minutes before it re-resolves

chrisohaver:
Are there any examples that do not involve cache?  The title and description are quite general and actually don't mention cache. 

If your concern is only with _cache_ behavior I think this might be safer to fix more directly in cache than in utility packages. For example, `Typify` is used by several plugins.  Changing the behavior to suit this cache scenario could have unintended consequences in different contexts.

ponienselvan:
> Are there any examples that do not involve cache? The title and description are quite general and actually don't mention cache.
> 
> If your concern is only with _cache_ behavior I think this might be safer to fix more directly in cache than in utility packages. For example, `Typify` is used by several plugins. Changing the behavior to suit this cache scenario could have unintended consequences in different contexts.

Understand, but still think if response is of different type than question it should be classified as NODATA. Don't see problem in submitting new patch touching ttl.go . Let me know how to proceed.

miekg:
Might be worth doing it here so all plugins can benefit, I have to look
this over carefully though...

plugin/pkg/response/typify.go
```diff
@@ -101,10 +110,15 @@ func Typify(m *dns.Msg, t time.Time) (Type, *dns.OPT) {
 			ns++
 		}
 	}
-
+	
 	if soa && m.Rcode == dns.RcodeSuccess {
 		return NoData, opt
 	}
+
+	if len(m.Answer)>0 && !answerMatch && m.Rcode == dns.RcodeSucess && !soa {
```
chrisohaver:
```suggestion
	if len(m.Answer)>0 && !answerMatch && m.Rcode == dns.RcodeSuccess && !soa {
```

chrisohaver:
@ponienselvan can you test this and get it working locally?

ponienselvan:
> @ponienselvan can you test this and get it working locally?

Will do.

ponienselvan:
> > @ponienselvan can you test this and get it working locally?
> 
> Will do.

Done. Works as expected. Open to just fixing ttl.go as well.
if you want to start fresh pr - let me know.
Thank you for the patience!

miekg:
I think this completely break DNSSEC, and from the look of it, it's probably not even valid in plain DNS. Why can't these things return a SOA? 

I'm interested to learn what other resolver, *forward/cache* is not that, make of this.

ponienselvan:
> I think this completely break DNSSEC, and from the look of it, it's probably not even valid in plain DNS. Why can't these things return a SOA?

DNSSEC: Noticed that - the fix should be applicable only for A/AAA questions  
Why can't these return a SOA - valid question, we have No control/influence over these but trying to fix these as well.

> I'm interested to learn what other resolver, _forward/cache_ is not that, make of this.

Have tested with BlueCoat, BIND, unbind, PowerDNS  - All of them do the right thing ie do not use TTL on CNAME record(as negative cache) in the absence of SOA record. 

Additionally if the reply were to contain different types CoreDNS applies lowest TTL for all records, this is also passed on to downstream clients - Other caching dns (i tested just BIND, BlueCoat recently) tracks TTL on  every record type separately. This is completely different issue needs to be fixed at some point.

ponienselvan:
Let me know if a new PR is required.

chrisohaver:
> Let me know if a new PR is required.

I think that since you are focused on cache examples and precedent, you should fix this directly in cache.  Changing Typify comes with the added burden of _carefully_ determining how it affects all other contexts.

IOW, IMO, a PR that addresses this directly in cache would be easier to review, safer, and thus likely to be merged more quickly than this PR.

This PR can be left open for further review.  As miekg says, it could be the better solution.  But if a fix to cache is needed sooner, then fixing directly in cache is IMO a good fix for now, even if this PR ends up being merged eventually.

ponienselvan:
Will  create PR with  small change set, change will be in plugin/pkg/dnsutil/ttl.go and ttl_test.go

PR #4992, opened 2021-11-20
Encourage the use of tls_servername in forward plugin

OctoHuman:
Updates the docs to warn users to specify `tls_servername` when using the forward plugin to forward over TLS.

If `tls_servername` isn't specified then any TLS certificate that has been signed by a trusted CA will be accepted, which means anyone with a cert issued to another domain could man-in-the-middle you.

While there are some valid use cases for this, for most, it probably is something undesirable.

Considering that no errors or warnings are thrown when you don't specify a `tls_servername` in the config while using TLS forwarding, users might not realize that the TLS connection they are establishing is highly insecure.
N/A
N/A
N/A

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -79,7 +79,9 @@ forward FROM TO... {
 * `tls_servername` **NAME** allows you to set a server name in the TLS configuration; for instance 9.9.9.9
   needs this to be set to `dns.quad9.net`. Multiple upstreams are still allowed in this scenario,
   but they have to use the same `tls_servername`. E.g. mixing 9.9.9.9 (QuadDNS) with 1.1.1.1
-  (Cloudflare) will not work.
+  (Cloudflare) will not work. Using TLS forwarding but not setting `tls_servername` results in anyone
+  being able to man-in-the-middle your connection to the DNS server you are forwarding to. Because of this,
+  it is strongly recommended to set this value when using TLS forwarding.
 * `policy` specifies the policy to use for selecting upstream servers. The default is `random`.
   * `random` is a policy that implements random upstream selection.
   * `round_robin` is a policy that selects hosts based on round robin ordering.


miekg:
/lgtm

can you sign your commit? I.e. add 'Signed-off-by:', git commit -s does this for you.

OctoHuman:
@miekg Sorry, I just fixed the sign off

polarathene:
The [`forward` plugin docs](https://coredns.io/plugins/forward/) state:

> **TO**… are the destination endpoints to forward to. The **TO** syntax allows you to specify a protocol, `tls://9.9.9.9` or `dns://` (or no protocol) for plain DNS. The number of upstreams is limited to 15.

**I'd like some clarification of the protocol schemes mentioned:**

- `tls://` is [not a valid IANA URI scheme](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml)? Is this to indicate intent for a secure connection for both DoH and DoT? Is this CoreDNS specific? (_[DoT RFC 7858](https://datatracker.ietf.org/doc/html/rfc7858) mentions nothing about `tls://`_)
- `dns://` (_[RFC 4501 (2006)](https://datatracker.ietf.org/doc/html/rfc4501#section-4)_) is presumably equivalent to no protocol scheme for "plain DNS"?

From this PR:

> Using TLS forwarding but not setting `tls_servername` results in anyone
  being able to man-in-the-middle your connection to the DNS server you are forwarding to. Because of this,
  it is strongly recommended to set this value when using TLS forwarding.

To clarify, **is this referring to only upstreams using `tls://` before the IP?** Unrelated to a bare IP or `dns://` or applicable there too? (_I assume TLS forwarding is only intended for upstreams with `tls://` but that is not clear in the documentation_)

At the end of the docs page there are examples which have none for `dns://`, but there are some with `tls://` that is omitted in the last example with multiple `tls_servername`, **presumably the `tls://` or `dns://` is implicit based on if `tls_servername` is specified or not?**

---

> Considering that no errors or warnings are thrown when you don't specify a `tls_servername` in the config while using TLS forwarding, users might not realize that the TLS connection they are establishing is highly insecure.

Which leaves me to wonder, if that is the case (_implicit scheme_), then shouldn't an issue be raised to properly throw an error/warning if `tls_servername` is not specified but `tls://` is for the upstream IPs? (_The docs examples do state that `tls_servername` is mandatory for TLS negotiation_)

The docs PR here only helps if the user actually reads that page and specifically that part to be aware of avoiding the mistake, others may have read it prior and not been aware and continue to update their deployed CoreDNS without being wiser..

Presumably you can detect the `tls://` usage and a lack of an `tls_servername`, or is `tls://` and `dns://` just decorative for the upstream IPs and stripped/ignored?

**Is `tls_servername` the actual toggle for TLS support instead**, and all upstreams must support that (_as is kind of implied with mentions of `tls_servername` not being possible to mix FQDNs_), no `dns://` or bare IP (_or as assumed, that's irrelevant and it's all implicitly `tls://`?_)

~~I would have thought you could avoid the more verbose example at the end which seems like a workaround for multiple upstream vendors by supporting `tls://<FQDN here>`?~~ _Nope, misunderstanding on my part. The [hostnames for the IPs (`one.one.one.one` and `1dot1dot1dot1.cloudflare-dns.com`) are unrelated](https://developers.cloudflare.com/1.1.1.1/encrypted-dns/dns-over-tls#how-it-works), `cloudflare-dns.com` serves the cert from a different IP, but `dns.quad9.net` maps to `9.9.9.9`_ :man_shrugging:

Issue #4996, opened 2021-11-22
Release tracking: 1.8.7


miekg:
Let's release 1.8.7, bunch of things got merged, nothing earth shattering.

yongtang:
Will try to release 1.8.7 today. /cc @miekg  @chrisohaver

yongtang:
From the error looks like the golang download link should be updated.

/cc @miekg  @chrisohaver do you know where the script should be updated?

chrisohaver:
Fixed here I think ... https://github.com/coredns/release/blob/master/release-coredns.
And then I think needs to be pushed to drone by @miekg.

... or maybe another drone script pulls this? I don't know.

mrueg:
Are there any plans to continue on this release?

yongtang:
coredns 1.8.7 has been released successfully 👍 🎉 !

PR #5011, opened 2021-11-26
Bump version to 1.8.7

yongtang:
Bump version to 1.8.7, part of issue #4996

### DIFF
--- a/coremain/version.go
+++ b/coremain/version.go
@@ -2,7 +2,7 @@ package coremain
 
 // Various CoreDNS constants.
 const (
-	CoreVersion = "1.8.6"
+	CoreVersion = "1.8.7"
 	coreName    = "CoreDNS"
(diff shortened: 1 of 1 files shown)


yongtang:
@miekg  @chrisohaver Unless there is an objection, I think we can release 1.8.7 first? After that we can prepare a 1.9.0 which will consist of the wildcard changes.

chrisohaver:
@yongtang 

We should consider adding the deprecation warning #5030 in a release before removing wildcard feature.

yongtang:
@chrisohaver With PR #5030 merged, we should be ready to go for 1.8.7.

chrisohaver (approved):

yongtang:
Actually we probably want to carry #5020 as well.

yongtang:
Probably should have merge #5020 first.

PR #5019, opened 2021-12-01
plugin/kubernetes: remove wildcard query functionality

chrisohaver:

Removes wildcard functionality completely from the _kubernetes_ plugin.

This is an alternative to disabling wildcard queries via an option (#4985).

**Note**: It is expected that all wildcard query tests in the **ci/circleci: kubernetes-tests** will fail until those specific cases are updated/removed.

#4984

Wildcard queries section removed from README. Done.

Functionally, yes - wildcard queries will no longer work (NXDOMAIN returned instead).


johnbelamaric:
from a process perspective, I am OK with this as long as we start a deprecation process in the K8s community indicating when this is going away (presumably this is a 1.9.0 change). The k8s community can continue to ship 1.8.x for a K8s release or two.

I do worry that if there are people that *really* depend on this, they may fork over it.

chrisohaver:
> from a process perspective, I am OK with this as long as we start a deprecation process in the K8s community indicating when this is going away (presumably this is a 1.9.0 change). The k8s community can continue to ship 1.8.x for a K8s release or two.

Agreed. Until then, we could log warnings when wildcard queries are received.

> I do worry that if there are people that _really_ depend on this, they may fork over it.

Or, make a copy of the kubernetes plugin and maintain it privately as an external plugin.  I'm certain some people do this to implement kubernetes plugin feature requests we have rejected in the past. Although, strictly speaking, anyone using an external plugin likely forks coredns to maintain `plugin.cfg`.

johnbelamaric:
>
>
> Agreed. Until then, we could log warnings when wildcard queries are
> received.
>
>
>
That'd be great. Though if there are a LOT maybe we need to suppress logs
and only print them periodically (with a count or something). Never simple
:)

chrisohaver:
> That'd be great. Though if there are a LOT maybe we need to suppress logs and only print them periodically (with a count or something). Never simple :)

There’s the consolidate option in _errors_ which I think would work for that …

chrisohaver:
> > That'd be great. Though if there are a LOT maybe we need to suppress logs and only print them periodically (with a count or something). Never simple :)
> 
> There’s the consolidate option in _errors_ which I think would work for that …

Actually - I implemented it this way, and it plays out to be pretty awkward.  It requires a lengthy `consolidate` config in the Corefile, impacts performance, and the consolidated error messages end up being attributed to errors not kubernetes, e.g....

```
[WARNING] plugin/errors: 4 errors like 'Deprecated wildcard query received. Wildcard queries will no longer be supported in the next minor release.' occurred in last 30s
```

I think it would work better directly implemented in the kubernetes plugin, with a counter.

miekg:
we could add a metric for this specifically?

chrisohaver:
> we could add a metric for this specifically?

We could, in addition to a warning in the log. Since users would have to know about the new metric and add it to monitoring tools to track it, I don't think the metric by itself would suffice as a deprecation warning to users. It would be a short lived metric, requiring that we deprecate it when we remove the wildcard code.  I don't think the number of wildcard queries matters enough to track as a metric.

miekg:
[ Quoting ***@***.***> in "Re: [coredns/coredns] plugin/kubern..." ]
>    we could add a metric for this specifically?
>
>We could, in addition to a warning in the log. Since users would have to know
>about the new metric and add it to monitoring tools to track it, I don't think
>the metric by itself would suffice as a deprecation warning to users. It would
>be a short lived metric, requiring that we deprecate it when we remove the
>wildcard code. I don't think the number of wildcard queries matters enough to
>track as a metric.

fair point, logging sgtm

chrisohaver:
Not stale.

chrisohaver:
The circleci kubernetes wildcard query tests are removed in coredns/ci#150

PR #5021, opened 2021-12-02
plugin/file: Fix NXDomain status case

victory460:

When using the file plugin, lookup a domain that does not exist in zone file, it does not return the expected status NXDOMAIN but SERVFAIL instead.
In addition, it will print a error log :
`[ERROR] plugin/errors: 2 server1.example222.org. A: plugin/file: no next plugin found`
The reason for the error log is that no plugin behind the file that can be executed.

My test case corefile：
```
.:1053 {
    file db.example.org example.org
    log
    debug
}
```

the zone file db.example.org :
```
$ORIGIN example.org.
@  3600 IN    SOA sns.dns.icann.org. noc.dns.icann.org. (
            2017042745 ; serial
            7200       ; refresh (2 hours)
            3600       ; retry (1 hour)
            1209600    ; expire (2 weeks)
            3600       ; minimum (1 hour)
            )

server1  IN  A  10.0.1.3
         IN  A  10.0.1.5

```
Test dig a not exist domain  :
```
$ dig @127.0.0.1 -p 1053 server1.example222.org

; <<>> DiG 9.10.6 <<>> @127.0.0.1 -p 1053 server1.example222.org
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 14435        // <- notes the status

;; OPT PSEUDOSECTION:
;; QUESTION SECTION:
;server1.example222.org.		IN	A

;; Query time: 0 msec
;; SERVER: 127.0.0.1#1053(127.0.0.1)
;; WHEN: Thu Dec 02 21:44:53 CST 2021
;; MSG SIZE  rcvd: 51
```

After fixed when dig a not exist domain:
```
dig @127.0.0.1 -p 1053 server1.example222.org

; <<>> DiG 9.10.6 <<>> @127.0.0.1 -p 1053 server1.example222.org
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 12467      // <- notes the status
;; flags: qr aa rd; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
;; QUESTION SECTION:
;server1.example222.org.		IN	A

;; Query time: 0 msec
;; SERVER: 127.0.0.1#1053(127.0.0.1)
;; WHEN: Thu Dec 02 21:43:39 CST 2021
;; MSG SIZE  rcvd: 51
```

None.
None.
None.

### DIFF
--- a/plugin/file/file.go
+++ b/plugin/file/file.go
@@ -39,7 +39,12 @@ func (f File) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (i
 	// TODO(miek): match the qname better in the map
 	zone := plugin.Zones(f.Zones.Names).Matches(qname)
 	if zone == "" {
-		return plugin.NextOrFailure(f.Name(), f.Next, ctx, w, r)
+		m := new(dns.Msg)
+		m.SetReply(r)
+		m.Authoritative = true
+		m.Rcode = dns.RcodeNameError
+		w.WriteMsg(m)
+		return dns.RcodeNameError, nil
 	}
 
 	z, ok := f.Zones.Z[zone]


chrisohaver (changes requested):
The current behavior is correct and expected.

chrisohaver:
The correct Corefile would be...

```
example.org:1053 {
    file db.example.org example.org
    log
    debug
}
```

The proposed changes makes file grab all domains, making it impossible for plugins _after_ it to process requests.

chrisohaver:
Another way to do it would be to forward other domains to an upstream server...

```
.:1053 {
    file db.example.org example.org
    forward . 8.8.8.8
}
```

Note that with the proposed change, this would no longer work.

victory460:
> The current behavior is correct and expected.

When lookup a domain that does not exist in zone file, Is it correct behavior status NXDOMAIN or SERVFAIL ? @chrisohaver

chrisohaver:
If the requested name is not in the zone for which file is authortitative, then NXDOMAIN is the wrong answer.

victory460:
> If the requested name is not in the zone for which file is authortitative, then NXDOMAIN is the wrong answer.

Thanks.

chrisohaver:
You can configure coredns to act the way you want using the following Corefile ...

```
example.org:1053 {
    file db.example.org example.org
    log
    debug
}

.:1053 {
    template ANY ANY {
      rcode NXDOMAIN
    } 
}
```

victory460:
> Another way to do it would be to forward other domains to an upstream server...
> 
> ```
> .:1053 {
>     file db.example.org example.org
>     forward . 8.8.8.8
> }
> ```
> 
> Note that with the proposed change, this would no longer work.

yes, add forward plugin in next , it will work fine, because the request is forward to another NS. Thanks.

victory460:
> You can configure coredns to act the way you want using the following Corefile ...
> 
> ```
> example.org:1053 {
>     file db.example.org example.org
>     log
>     debug
> }
> 
> .:1053 {
>     template ANY ANY {
>       rcode NXDOMAIN
>     } 
> }
> ```

Thanks a lot. I test it, but not work with NXDOMAIN status. Did i make a mistake in some where？

```
$ dig @127.0.0.1 -p 1053 server1.example222.org

; <<>> DiG 9.10.6 <<>> @127.0.0.1 -p 1053 server1.example222.org
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 17213.  // <- notes the status

;; OPT PSEUDOSECTION:
;; QUESTION SECTION:
;server1.example222.org.		IN	A

;; Query time: 2 msec
;; SERVER: 127.0.0.1#1053(127.0.0.1)
;; WHEN: Thu Dec 02 23:43:45 CST 2021
;; MSG SIZE  rcvd: 51
```

chrisohaver:
Whats your use case?

chrisohaver:
> Thanks a lot. I test it, but not work with NXDOMAIN status. Did i make a mistake in some where?

It works for me.

victory460:
> > Thanks a lot. I test it, but not work with NXDOMAIN status. Did i make a mistake in some where?
> 
> It works for me.

Thank you for your careful answer, it is indeed work fine.

PR #5057, opened 2021-12-23
Fix

jinglina:
when nameserver not found, nodelocaldns using coredns won't get up, I think it is not a good idea.


SuperQ:
Please fix the branch history on this PR.

ihac:
Please pull the latest commit from the upstream master branch and cherry-pick your commits on top of it
image

SuperQ:
I don't think this is the correct fix. If there are no upstream servers, this is a problem. The original reason for this was a nil list of servers will panic.

chrisohaver:
Please …
1. update your `master` branch
2. rebase `fix` with `master`
3. force push `fix`

chrisohaver:
> I don't think this is the correct fix. If there are no upstream servers, this is a problem. The original reason for this was a nil list of servers will panic.

I agree, This change would require all callers of `HostPortOrFile` to check for a nil list otherwise they might panic with a nil ref. 

If a caller wants to tolerate an empty list, the *caller* should handle that specific error.

SuperQ:
We could move the error to an exported var so the caller can use `errors.Is()`

SuperQ:
Closing in favor of #5058.

chrisohaver:
> Closing in favor of #5058.

FWIW, #5058 doesn't itself solve the problem presented by @jinglina.
However, I think it's appropriate to close this PR, since it's not the correct approach.

#5058 makes it more graceful to handle this error from callers who might want behave specially when getting an empty list of hosts (without it, the caller would need to compare the error to a hard coded string, which is clumsy). 

I suspect @jinglina's concern specifically relates to the behavior of the _forward_ plugin.  In which case a solution would involve the _forward_ plugin tolerating this specific error, **and** at least some code in its `ServeDNS` to handle an empty list (like always returning SERVFAIL if there are no upstream servers).

PR #5064, opened 2022-01-01
fix: convert key to domain

soulfy:
If the Key ends with '/', there'll be an extra character '.' at the beginning of the Domain.

### DIFF
--- a/plugin/etcd/msg/path_test.go
+++ b/plugin/etcd/msg/path_test.go
@@ -11,7 +11,7 @@ func TestPath(t *testing.T) {
 	}
 }
 
-func TestPath(t *testing.T) {
+func TestDomain(t *testing.T) {
 	result1 := Domain("/skydns/local/cluster/staging/service/")
 	if result1 != "service.staging.cluster.local." {
 		t.Errorf("==>%s", result1)
--- a/plugin/etcd/msg/path_test.go
+++ b/plugin/etcd/msg/path_test.go
@@ -14,11 +14,11 @@ func TestPath(t *testing.T) {
 func TestDomain(t *testing.T) {
 	result1 := Domain("/skydns/local/cluster/staging/service/")
 	if result1 != "service.staging.cluster.local." {
-		t.Errorf("==>%s", result1)
+		t.Errorf("Failure to get domain from etcd key (with a trailing '/'), expect: 'service.staging.cluster.local.', actually get: '%s'", result1)
 	}
 
 	result2 := Domain("/skydns/local/cluster/staging/service")
 	if result2 != "service.staging.cluster.local." {
-		t.Errorf("==>%s", result2)
+		t.Errorf("Failure to get domain from etcd key (without trailing '/'), expect: 'service.staging.cluster.local.' actually get: '%s'", result2)
 	}
 }
--- a/plugin/etcd/msg/path_test.go
+++ b/plugin/etcd/msg/path_test.go
@@ -10,3 +10,15 @@ func TestPath(t *testing.T) {
 		}
 	}
 }
+
+func TestPath(t *testing.T) {
+	result1 := Domain("/skydns/local/cluster/staging/service/")
+	if result1 != "service.staging.cluster.local." {
+		t.Errorf("==>%s", result1)
+	}
+
+	result2 := Domain("/skydns/local/cluster/staging/service")
+	if result2 != "service.staging.cluster.local." {
+		t.Errorf("==>%s", result2)
+	}
+}
--- a/plugin/etcd/msg/path.go
+++ b/plugin/etcd/msg/path.go
@@ -22,6 +22,9 @@ func Path(s, prefix string) string {
 // Domain is the opposite of Path.
 func Domain(s string) string {
 	l := strings.Split(s, "/")
+	if l[len(l)-1] == "" {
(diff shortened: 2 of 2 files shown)


plugin/etcd/msg/path_test.go
```diff
@@ -10,3 +10,15 @@ func TestPath(t *testing.T) {
 		}
 	}
 }
+
+func TestPath(t *testing.T) {
```
chrisohaver: A function with the same name is declared 10 lines earlier in the same file.

plugin/etcd/msg/path_test.go
```diff
@@ -10,3 +10,15 @@ func TestPath(t *testing.T) {
...
+	result1 := Domain("/skydns/local/cluster/staging/service/")
+	if result1 != "service.staging.cluster.local." {
+		t.Errorf("==>%s", result1)
+	}
+
+	result2 := Domain("/skydns/local/cluster/staging/service")
+	if result2 != "service.staging.cluster.local." {
+		t.Errorf("==>%s", result2)
+	}
```
chrisohaver: These errors should be more descriptive.

chrisohaver:
Not sure if a trailing `/` in a key is valid in the SkyDNS schema?  It looks to be valid in etcd, but all SkyDNS examples that I see have no trailing slash. Guessing it is valid, or at least couldn't hurt to accept.

soulfy:
> Not sure if a trailing `/` in a key is valid in the SkyDNS schema? It looks to be valid in etcd, but all SkyDNS examples that I see have no trailing slash. Guessing it is valid, or at least couldn't hurt to accept.

Sorry, previously I only executed a single test case, but didn't finish compiling.

In my scenario, there is indeed a  trailing `/` in a key, and it's valid in etcd.

plugin/etcd/msg/path_test.go
```diff
@@ -10,3 +10,15 @@ func TestPath(t *testing.T) {
...
+func TestDomain(t *testing.T) {
+	result1 := Domain("/skydns/local/cluster/staging/service/")
+	if result1 != "service.staging.cluster.local." {
+		t.Errorf("==>%s", result1)
+	}
+
+	result2 := Domain("/skydns/local/cluster/staging/service")
+	if result2 != "service.staging.cluster.local." {
+		t.Errorf("==>%s", result2)
+	}
```
chrisohaver: These test failures should be more descriptive.

chrisohaver (changes requested):
Perhaps someone more familiar with SkyDNS and etcd can answer...
Is a trailing slash in a key is valid in the _SkyDNS_ schema? It is clearly a valid key in etcd (but that doesn't necessarily mean it's valid in SkyDNS).
Is a key without a trailing slash considered equal to the same key with a trailing slash in etcd?

soulfy:
> Perhaps someone more familiar with SkyDNS and etcd can answer... Is a trailing slash in a key is valid in the _SkyDNS_ schema? It is clearly a valid key in etcd (but that doesn't necessarily mean it's valid in SkyDNS). Is a key without a trailing slash considered equal to the same key with a trailing slash in etcd?

In my test, a key without a trailing slash is actually equal to the same key with a trailing slash in etcd.

Anyway, i think this fix can increase compatibility. :-)

chrisohaver:
>> It is clearly a valid key in etcd

> In my test, a key without a trailing slash is actually equal to the same key with a trailing slash _in etcd_.

That isn't in doubt.  The question is "is a trailing slash in a key is valid in _the SkyDNS schema_?".
SkyDNS being a _subset_ of etcd. I _think_ it probably is.  But someone who knows SkyDNS better should confirm.  If there is no one that can confirm, it looks ok to me.

soulfy:
In the customised skydns I'm involved in, that was it.

plugin/etcd/msg/path_test.go
chrisohaver: This was marked resolved - but no changes were made.  The errors are non-descriptive and identical, so it's impossible to know which one failed if one does, and the other does not.   An error should _at the very least_ provide a means of uniquely identifying which test failed.  Better, would be to also show the expected result and the incorrect result. Best would be to also briefly describe what is being tested.  Please see other tests in the project as examples.

_Edit:_ Actually, since the errors are on separate lines, a failure can be traced back to the specific failed test case but one should not have to trace a test case back to test code to determine what the expected value was. Please make the error more descriptive than "==>", as suggested above.

chrisohaver (changes requested):

soulfy:
> Edit: Actually, since the errors are on separate lines, a failure can be traced back to the specific failed test case but one should not have to trace a test case back to test code to determine what the expected value was. Please make the error more descriptive than "==>", as suggested above.

sorry, just realized what was wrong  🤣

chrisohaver (approved):

chrisohaver (approved):
LGTM. No other reviews, so will merge.

chrisohaver:
Oh.  @soulfy, please [sign the commits](https://github.com/coredns/coredns/pull/5064/checks?check_run_id=4961249280).

soulfy:
> Oh. @soulfy, please [sign the commits](https://github.com/coredns/coredns/pull/5064/checks?check_run_id=4961249280).

😢

PR #5070, opened 2022-01-04
fix goroutine leaks in TestRoute53

timmyyuan:

I find there is a goroutine leaking in `TestRoute53` and fix it:

```
[Goroutine 28 in state select, with github.com/coredns/coredns/plugin/route53.(*Route53).Run.func1 on top of the stack:
goroutine 28 [select]:
github.com/coredns/coredns/plugin/route53.(*Route53).Run.func1()
	/root/actions-runner/_work/synctrapper/synctrapper/coredns/plugin/route53/route53.go:87 +0x129
created by github.com/coredns/coredns/plugin/route53.(*Route53).Run
	/root/actions-runner/_work/synctrapper/synctrapper/coredns/plugin/route53/route53.go:85 +0xf9
]
```

I suspect `TestSetupRoute53` is also leak goroutines since it doesn't call any functions in `c.OnShutdown` but I don't know how to fix it.

No

Not needed

No

### DIFF
--- a/plugin/route53/route53_test.go
+++ b/plugin/route53/route53_test.go
@@ -77,7 +77,8 @@ func (fakeRoute53) ListResourceRecordSetsPagesWithContext(_ aws.Context, in *rou
 }
 
 func TestRoute53(t *testing.T) {
-	ctx := context.Background()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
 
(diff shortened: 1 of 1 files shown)


yongtang:
@timmyyuan Can you fix the DCO issue as was specified?
https://github.com/coredns/coredns/pull/5070/checks?check_run_id=4701467919

timmyyuan:
@yongtang Done

yongtang (approved):

PR #5076, opened 2022-01-07
plugin/file: fix a wildcard lookup bug in multi-label condition

victory460:

In a complex scenario, the wildcard lookup work not correctly.

Corefile:

```
example.org:1053 {
    file db.example.org example.org
    log
    debug
}
```

db.example.org:

```
$ORIGIN example.org.
@	3600 IN	SOA sns.dns.icann.org. noc.dns.icann.org. (
				2017042745 ; serial
				7200       ; refresh (2 hours)
				3600       ; retry (1 hour)
				1209600    ; expire (2 weeks)
				3600       ; minimum (1 hour)
				)

*  IN  A  10.1.1.1
server1  IN  A  10.1.1.2

```

When `dig test.server1.example.org` , it shoud answer `10.1.1.1`. But it answer none record.

```shell
$ dig @127.0.0.1 -p 1053 test.server1.example.org

; <<>> DiG 9.10.6 <<>> @127.0.0.1 -p 1053 test.server1.example.org
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 37049

;; OPT PSEUDOSECTION:
;; QUESTION SECTION:
;test.server1.example.org.	IN	A

;; AUTHORITY SECTION:
example.org.		3600	IN	SOA	sns.dns.icann.org. noc.dns.icann.org. 2017042745 7200 3600 1209600 3600

;; Query time: 0 msec
;; SERVER: 127.0.0.1#1053(127.0.0.1)
;; WHEN: Fri Jan 07 16:14:49 CST 2022
;; MSG SIZE  rcvd: 134
```

None.
None.
None.

### DIFF
--- a/plugin/file/lookup.go
+++ b/plugin/file/lookup.go
@@ -93,7 +93,66 @@ func (z *Zone) Lookup(ctx context.Context, state request.Request, qname string) 
 		}
 
 		elem, found = tr.Search(parts)
-		if !found {
+
+		if found {
+			// If we see DNAME records, we should return those.
+			if dnamerrs := elem.Type(dns.TypeDNAME); dnamerrs != nil {
+				// Only one DNAME is allowed per name. We just pick the first one to synthesize from.
+				dname := dnamerrs[0]
+				if cname := synthesizeCNAME(state.Name(), dname.(*dns.DNAME)); cname != nil {
+					var (
+						answer, ns, extra []dns.RR
+						rcode             Result
+					)
+
+					// We don't need to chase CNAME chain for synthesized CNAME
+					if qtype == dns.TypeCNAME {
+						answer = []dns.RR{cname}
+						ns = ap.ns(do)
+						extra = nil
+						rcode = Success
+					} else {
+						ctx = context.WithValue(ctx, dnsserver.LoopKey{}, loop+1)
+						answer, ns, extra, rcode = z.externalLookup(ctx, state, elem, []dns.RR{cname})
+					}
+
+					if do {
+						sigs := elem.Type(dns.TypeRRSIG)
+						sigs = rrutil.SubTypeSignature(sigs, dns.TypeDNAME)
+						dnamerrs = append(dnamerrs, sigs...)
+					}
+
+					// The relevant DNAME RR should be included in the answer section,
+					// if the DNAME is being employed as a substitution instruction.
+					answer = append(dnamerrs, answer...)
+
+					return answer, ns, extra, rcode
+				}
+				// The domain name that owns a DNAME record is allowed to have other RR types
+				// at that domain name, except those have restrictions on what they can coexist
+				// with (e.g. another DNAME). So there is nothing special left here.
+			}
+			// If we see NS records, it means the name as been delegated, and we should return the delegation.
+			if nsrrs := elem.Type(dns.TypeNS); nsrrs != nil {
+				// If the query is specifically for DS and the qname matches the delegated name, we should
+				// return the DS in the answer section and leave the rest empty, i.e. just continue the loop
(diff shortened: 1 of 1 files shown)


chrisohaver (changes requested):
> ```
> *  IN  A  10.1.1.1
> server1  IN  A  10.1.1.2
> ```
> When dig test.server1.example.org , it shoud answer 10.1.1.1. But it answer none record.

This is not the correct expectation. The `server1` record prevents wildcard record synthesization "below it".  NXDOMAIN is the correct answer.

From RFC 1034 ...

> Wildcard RRs do not apply:
> [...]
>   - When the query name or a name between the wildcard domain and
     the query name is know to exist.  For example, if a wildcard
     RR has an owner name of "*.X", and the zone also contains RRs
     attached to B.X, the wildcards would apply to queries for name
     Z.X (presuming there is no explicit information for Z.X), but
     not to B.X, A.B.X, or X.

chrisohaver:
> NXDOMAIN is the correct answer.

There should be a test added to validate this case; one that would fail with this PR.

victory460:
> > NXDOMAIN is the correct answer.
> 
> There should be a test added to validate this case; one that would fail with this PR.

Thanks, you are right. We add a lookup test case for wildcard RRs do not apply. #5078 
Want to you review it, when you have time.

victory460:
> > ```
> > *  IN  A  10.1.1.1
> > server1  IN  A  10.1.1.2
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > When dig test.server1.example.org , it shoud answer 10.1.1.1. But it answer none record.
> 
> This is not the correct expectation. The `server1` record prevents wildcard record synthesization "below it". NXDOMAIN is the correct answer.
> 
> From RFC 1034 ...
> 
> > Wildcard RRs do not apply:
> > [...]
> > 
> > * When the query name or a name between the wildcard domain and
> >   the query name is know to exist.  For example, if a wildcard
> >   RR has an owner name of "*.X", and the zone also contains RRs
> >   attached to B.X, the wildcards would apply to queries for name
> >   Z.X (presuming there is no explicit information for Z.X), but
> >   not to B.X, A.B.X, or X.

Thanks, you are right. I have a question that if db.example.org is :
```
$ORIGIN example.org.
@	3600 IN	SOA sns.dns.icann.org. noc.dns.icann.org. (
				2017042745 ; serial
				7200       ; refresh (2 hours)
				3600       ; retry (1 hour)
				1209600    ; expire (2 weeks)
				3600       ; minimum (1 hour)
				)

*  IN  A  10.1.1.1
b.c  IN  A  10.1.1.2
```
When `dig a.b.c.example.org` , what the correct expectation? NXDOMAIN ? @chrisohaver

chrisohaver:
> ```
> *  IN  A  10.1.1.1
> b.c  IN  A  10.1.1.2
> ```
> 
> 
>     
>   
> 
> When `dig a.b.c.example.org` , what the correct expectation? NXDOMAIN ?

Yes.

victory460:
> > ```
> > *  IN  A  10.1.1.1
> > b.c  IN  A  10.1.1.2
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > When `dig a.b.c.example.org` , what the correct expectation? NXDOMAIN ?
> 
> Yes.

Thanks. But now it answer 10.1.1.1,  we will try to fix it later.

chrisohaver:
> Thanks. But now it answer 10.1.1.1, we will try to fix it later

See #4655

PR #5077, opened 2022-01-07
Fallback to MinimalDefaultTTL when noerror no valid response for A? and no SOA

ponienselvan:
When upstream responses doesn't include SOA to indicate NODATA but with NOERROR code - CoreDNS treats this as NOERROR and picks least TTL from any record included in the response.

This is typical behavior in case of F5 Global Loadbalancer dns responses:

1. When all service members fail GLB returns no records and NOERROR
2. Combined with recursion by an upstream server - responses may contain CNAME   but NOERROR and NOSOA

This was also the background for unmerged PR https://github.com/coredns/coredns/pull/4987.
PR 4987 also modified message classification under such scenarios as NODATA, with this request changes are limited to minTTL calculation.

see Above
Not Applicable
Not Applicable

### DIFF
--- a/plugin/pkg/dnsutil/ttl.go
+++ b/plugin/pkg/dnsutil/ttl.go
@@ -20,6 +20,21 @@ func MinimalTTL(m *dns.Msg, mt response.Type) time.Duration {
 		return MinimalDefaultTTL
 	}
 
+	// Covers Cases where there is NoError also answer type doesnt match question and no SOA
+	answerMatch := false
+	if len(m.Question) > 0 && len(m.Ns) == 0 && len(m.Answer) > 0 {
+		if m.Question[0].Qtype == dns.TypeA || m.Question[0].Qtype == dns.TypeAAAA {
+			for _, r := range m.Answer {
+				if m.Question[0].Qtype == r.Header().Rrtype {
+					answerMatch = true
+				}
+			}
+			if !answerMatch {
+				return MinimalDefaultTTL
+			}
+		}
+	}
+
 	minTTL := MaximumDefaulTTL
 	for _, r := range m.Answer {
 		if r.Header().Ttl < uint32(minTTL.Seconds()) {
--- a/plugin/pkg/dnsutil/ttl_test.go
+++ b/plugin/pkg/dnsutil/ttl_test.go
@@ -41,6 +41,40 @@ func TestMinimalTTL(t *testing.T) {
 	}
 }
 
+func TestMinimalTTLNoSOA(t *testing.T) {
+	m := new(dns.Msg)
+	m.SetQuestion("bar.www.example.org.", dns.TypeA)
+	m.Rcode = dns.RcodeSuccess
+	m.Answer = []dns.RR{test.CNAME("bar.www.example.org. 3500 IN CNAME foo.example.org.")} // but we add a cname with the name!
+
+	utc := time.Now().UTC()
+	mt, _ := response.Typify(m, utc)
+	dur := MinimalTTL(m, mt)
+	if dur != MinimalDefaultTTL {
+		t.Fatalf("Expected minttl duration to be %d, got %d", MinimalDefaultTTL, dur)
+	}
+
+}
+
+func TestMinimalTTLNoErrorWithSOA(t *testing.T) {
+	m := new(dns.Msg)
+	m.SetQuestion("bar.www.example.org.", dns.TypeA)
+	m.Rcode = dns.RcodeSuccess
+	m.Answer = []dns.RR{test.CNAME("bar.www.example.org. 3600 IN CNAME foo.example.org.")} // but we add a cname with the name!
+	m.Ns = []dns.RR{
+		test.SOA("example.org.       900    IN      SOA     ns.example.org. dns.example.org. 2025042470 10000 2400 604800 3600"),
+	}
+
+	utc := time.Now().UTC()
+
+	mt, _ := response.Typify(m, utc)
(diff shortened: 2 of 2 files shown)


chrisohaver:
I don't see how the existing behavior can cause a problem in practice.  Take the following series of events for example...

1. A resolver requests an `A` record
2. CoreDNS forwards request to the upstream
3. The upstream returns a `NOERROR` response with only a single record in the answer section: a `CNAME` with no `A` record for the target, and no `SOA` in Authority section. (the case described by this PR)
4. CoreDNS caches this response for the TTL value of the `CNAME` record.
5. The resolver receives the answer from CoreDNS, see's that no `A` record is present in the answer for the `CNAME` target, and no `SOA` record is present in Authority.  It therefore makes a new query for the `CNAME` target.
6. CoreDNS receives the new query and forwards request to the upstream.
7. The upstream returns a response, and CoreDNS caches it using the TTL of the A Record (or SOA if no answer, etc).

The result is two cached responses for two separate questions, the initial and target queries.  Each has a different TTL.  If the TTL of the  initial response is longer than the target response TTL, then the target response will expire before the initial response.

The proposed change would cause CoreDNS to cache the initial repose for minTTL, which would result in the record expiring prematurely from cache.

ponienselvan:
> I don't see how the existing behavior can cause a problem in practice. Take the following series of events for example...
> 
> 1. A resolver requests an `A` record
> 2. CoreDNS forwards request to the upstream
> 3. The upstream returns a `NOERROR` response with only a single record in the answer section: a `CNAME` with no `A` record for the target, and no `SOA` in Authority section. (the case described by this PR)
> 4. CoreDNS caches this response for the TTL value of the `CNAME` record.
> 5. The resolver receives the answer from CoreDNS, see's that no `A` record is present in the answer for the `CNAME` target, and no `SOA` record is present in Authority.  It therefore makes a new query for the `CNAME` target.
> 6. CoreDNS receives the new query and forwards request to the upstream.
> 7. The upstream returns a response, and CoreDNS caches it using the TTL of the A Record (or SOA if no answer, etc).
> 
> The result is two cached responses for two separate questions, the initial and target queries. Each has a different TTL. If the TTL of the initial response is longer than the target response TTL, then the target response will expire before the initial response.
> 
> The proposed change would cause CoreDNS to cache the initial repose for minTTL, which would result in the record expiring prematurely from cache.

Everything you have described from step 5 onwards doesn't happen.
a) recursion desired bit is set and upstream does the necessary recursion
b) verified both getenthosts and getaddrinfo fail in this scenario.

chrisohaver:
Thanks.  In the scenario you are spotlighting ...

> When upstream response doesn't include SOA to indicate NODATA

Should we be caching at all in this scenario? Per RFC 2308 ...

>    Negative responses without SOA records SHOULD NOT be cached as there
   is no way to prevent the negative responses looping forever between a
   pair of servers even with a short TTL.

ponienselvan:
> Thanks. In the scenario you are spotlighting ...
> 
> > When upstream response doesn't include SOA to indicate NODATA
> 
> Should we be caching at all in this scenario? Per RFC 2308 ...
May be. Also the reason I started with Typify. RFC talks about NODATA/NXDOMAIN scenario - case covered in the patch is NOERROR case (note it would have been NODATA in original patch). I think defaulting to defaultmin ttl is better than wrongly caching with ttl on cname. 

> > Negative responses without SOA records SHOULD NOT be cached as there
> > is no way to prevent the negative responses looping forever between a
> > pair of servers even with a short TTL.

chrisohaver:
> I think defaulting to defaultmin ttl is better than wrongly caching with ttl on cname.

Ah, was this because default min ttl is closer RFCs guidance of not caching (i.e. zero)?
Seems it would be better to not cache at all than to cache for less time, if that that's what the RFC prescribes.

ponienselvan:
> > I think defaulting to defaultmin ttl is better than wrongly caching with ttl on cname.
> 
> Ah, was this because default min ttl is closer RFCs guidance of not caching (i.e. zero)? 

Yes. 

>Seems it would be better to not cache at all than to cache for less time, if that that's what the RFC prescribes.

chrisohaver:
> Yes

Seems it would be better to not cache at all than to cache for less time, if that that's what the RFC prescribes.

plugin/pkg/dnsutil/ttl.go
```diff
@@ -20,6 +20,21 @@ func MinimalTTL(m *dns.Msg, mt response.Type) time.Duration {
 		return MinimalDefaultTTL
 	}
 
+	// Covers Cases where there is NoError also answer type doesnt match question and no SOA
+	answerMatch := false
+	if len(m.Question) > 0 && len(m.Ns) == 0 && len(m.Answer) > 0 {
+		if m.Question[0].Qtype == dns.TypeA || m.Question[0].Qtype == dns.TypeAAAA {
```
chrisohaver, line 26: This needs to be more generalized, since CNAMEs can alias other record types than A and AAAA.

plugin/pkg/dnsutil/ttl.go
```diff
@@ -20,6 +20,21 @@ func MinimalTTL(m *dns.Msg, mt response.Type) time.Duration {
 		return MinimalDefaultTTL
 	}
 
+	// Covers Cases where there is NoError also answer type doesnt match question and no SOA
```
chrisohaver, line 23:
```suggestion
	// Handle case where CNAME is the only record type in the response to a non-CNAME query.
```

chrisohaver (changes requested):

PR #5078, opened 2022-01-08
plugin/file: add a lookup test case for wildcard RRs do not apply

victory460:

Add a lookup test case for wildcard RRs do not apply.   #5076 
From RFC 1034 ...

```
Wildcard RRs do not apply:
[...]

When the query name or a name between the wildcard domain and
the query name is know to exist. For example, if a wildcard
RR has an owner name of "*.X", and the zone also contains RRs
attached to B.X, the wildcards would apply to queries for name
Z.X (presuming there is no explicit information for Z.X), but
not to B.X, A.B.X, or X.
```

None.
None.
No.

### DIFF
--- a/plugin/file/lookup_test.go
+++ b/plugin/file/lookup_test.go
@@ -103,6 +103,13 @@ var dnsTestCases = []test.Case{
 		Ns: miekAuth,
 	},
 	{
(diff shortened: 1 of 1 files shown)


chrisohaver (approved):
Lgtm

yongtang (approved):
lgtn

PR #5079, opened 2022-01-10
plugin/file: add TXT test case

victory460:

Add TXT RR test case in lookup test.
None.
None.
No.


chrisohaver:
The SPF record type is deprecated.  Is there a compelling reason to add tests for it?

https://datatracker.ietf.org/doc/html/rfc7208#section-14.1

victory460:
> The SPF record type is deprecated. Is there a compelling reason to add tests for it?
> 
> https://datatracker.ietf.org/doc/html/rfc7208#section-14.1

Thanks for your review. I remove spf test case , just add a txt rr test case now.

PR #5086, opened 2022-01-11
plugin/file: unified NoData result

victory460:

When `len(rrs) == 0` in case of found entire name, it `return nil, ret, nil, NoData`, but in case of found wildcard, it `return nil, ret, nil, Success`.
This PR try to unified the result when `len(rrs) == 0`.
https://github.com/coredns/coredns/blob/6b537effda2c612b2bdbbd4a91bae231e9267e4c/plugin/file/lookup.go#L189-L197

https://github.com/coredns/coredns/blob/6b537effda2c612b2bdbbd4a91bae231e9267e4c/plugin/file/lookup.go#L226-L234
None.
None.
No.

### DIFF
--- a/plugin/file/lookup.go
+++ b/plugin/file/lookup.go
@@ -230,7 +230,7 @@ func (z *Zone) Lookup(ctx context.Context, state request.Request, qname string) 
 				nsec := typeFromElem(wildElem, dns.TypeNSEC, do)
 				ret = append(ret, nsec...)
 			}
-			return nil, ret, nil, Success
+			return nil, ret, nil, NoData
 		}
 
 		if do {
--- a/plugin/file/lookup_test.go
+++ b/plugin/file/lookup_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/coredns/coredns/plugin/pkg/dnstest"
 	"github.com/coredns/coredns/plugin/test"
+	"github.com/coredns/coredns/request"
 
 	"github.com/miekg/dns"
 )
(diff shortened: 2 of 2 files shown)


chrisohaver:
Can you provide a _specific_ example query and RR sets that would have been failing, but now fixed with this PR?

victory460:
> Can you provide a _specific_ example query and RR sets that would have been failing, but now fixed with this PR?

Thanks for your review.
https://github.com/coredns/coredns/blob/6b537effda2c612b2bdbbd4a91bae231e9267e4c/plugin/file/file.go#L92-L97

As the above code, because it didn't do specific anything when `result == NoData` now,  this PR is seem no effect.
This PR purpose to unified action logic when `len(rrs) == 0` between found entire name and found wildcard case.

chrisohaver:
> this PR is seem no effect 

I see.  So this PR doesn't fix anything that is outwardly broken, or change any outward behavior as far as you have observed.
_file_ and _auto_ treat `Nodata` and `Success` the same.

But other plugins _azure, cloudns, route53_ do have a following conditional that checks for `NoData` specifically.  So this change would presumably alter behavior in those plugins.

https://github.com/coredns/coredns/blob/6b537effda2c612b2bdbbd4a91bae231e9267e4c/plugin/azure/azure.go#L319-L328

chrisohaver:
> But other plugins azure, cloudns, route53 ...

Not sure if the backends for those plugins actually support wildcard records.  If not, then the condition could never occur. If they can then I'd think this probably fixes a bug: Where the search would continue to the next zone when a wildcard record exists for a type other than what is being requested, instead of returning an empty result to the client.

chrisohaver (approved):
I think this is correct, and probably safe.  I'd like to get other eyes on it though before it merges.

chrisohaver:
... and probably should get a unit test too.

I don't see any existing _direct_ unit tests for `Zone.Lookup`, but perhaps we should start one here, since it is a function used by many plugins.

victory460:
> ... and probably should get a unit test too.
> 
> I don't see any existing _direct_ unit tests for `Zone.Lookup`, but perhaps we should start one here, since it is a function used by many plugins.

Thanks for your review. Now we add NoData test case for lookup hit `len(rrs) == 0` that mock by query another type of none rr.  See #5087 .

chrisohaver:
> Thanks for your review. Now we add NoData test case for lookup hit len(rrs) == 0 that mock by query another type of none rr. See #5087 .

To clarify, I think a unit test should be added to this PR, since this PR also contains the fix it covers.
The unit test should test `Zone.Lookup` by calling `Zone.Lookup` directly (not like existing tests that test via ServeDNS). The new tests should validate the case fixed by this PR in such a way that it would fail without the fix in this PR.

victory460:
> > Thanks for your review. Now we add NoData test case for lookup hit len(rrs) == 0 that mock by query another type of none rr. See #5087 .
> 
> To clarify, I think a unit test should be added to this PR, since this PR also contains the fix it covers. The unit test should test `Zone.Lookup` by calling `Zone.Lookup` directly (not like existing tests that test via ServeDNS). The new tests should validate the case fixed by this PR in such a way that it would fail without the fix in this PR.

Thanks for your review. I  add a direct test according to your suggestion. Thanks a lot.

victory460:
@chrisohaver Thanks a lot!  I update it.

zouyee (approved):

zouyee:
@victory460 pls rebase pr, thanks!

victory460:
> @victory460 pls rebase pr, thanks!

@zouyee  Thanks! Done.

PR #5087, opened 2022-01-12
plugin/file: add NoData test case 

victory460:

Add NoData test case for lookup hit `len(rrs) == 0` that mock by query another type of none rr.
Background see https://github.com/coredns/coredns/issues/5086#issuecomment-1010219772_ .
None.
None.
No.

### DIFF
--- a/plugin/file/lookup_test.go
+++ b/plugin/file/lookup_test.go
@@ -42,6 +42,20 @@ var dnsTestCases = []test.Case{
 		},
 		Ns: miekAuth,
 	},
+	// the NoData test case for found entire name when lookup
+	{
+		Qname: "a.miek.nl.", Qtype: dns.TypeMX,
+		Ns: []dns.RR{
+			test.SOA("miek.nl.	1800	IN	SOA	linode.atoom.net. miek.miek.nl. 1282630057 14400 3600 604800 14400"),
+		},
+	},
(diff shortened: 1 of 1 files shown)


chrisohaver:
This isn't what I meant by ...

> I don't see any _existing_ direct unit tests for Zone.Lookup, but perhaps we should start one here, since it is a function used by many plugins.

By direct, I mean a direct test of the Lookup function. These existing tests are indirect through ServeDNS. In file's case, ServeDNS conflates NoData and Success results from Lookup, meaning that they cannot catch the bug fixed by #5086.

Hence I suggested adding a new set of tests that directly tests the Lookup function.

I also think the new test should be part of #5086. Not a separate PR.

victory460:
> This isn't what I meant by ...
> 
> > I don't see any _existing_ direct unit tests for Zone.Lookup, but perhaps we should start one here, since it is a function used by many plugins.
> 
> By direct, I mean a direct test of the Lookup function. These existing tests are indirect through ServeDNS. In file's case, ServeDNS conflates NoData and Success results from Lookup, meaning that they cannot catch the bug fixed by #5086.
> 
> Hence I suggested adding a new set of tests that directly tests the Lookup function.
> 
> I also think the new test should be part of #5086. Not a separate PR.

Thanks for your review. I will to add a direct test of the Lookup function later according to your suggestion.

PR #5088, opened 2022-01-12
plugin/file: fix lost extra section when delegated subdomain to another

victory460:

Fix lost extra section when delegated subdomain to another. For example:
test case:
```
	{
		Qname: "sub.miek.nl.", Qtype: dns.TypeNS,
		Ns: []dns.RR{
			test.NS("sub.miek.nl.	1800	IN	NS	a.miek.nl."),
		},
		Extra: []dns.RR{
			test.A("a.miek.nl.	1800	IN	A       139.162.196.78"),
			test.AAAA("a.miek.nl.	1800	IN	AAAA	2a01:7e00::f03c:91ff:fef1:6735"),
		},
	}
```
zone file:
```
sub          IN   NS     a.miek.nl.
a              IN      A       139.162.196.78
                IN      AAAA    2a01:7e00::f03c:91ff:fef1:6735
```
None.
None.
No.

### DIFF
--- a/plugin/file/lookup_test.go
+++ b/plugin/file/lookup_test.go
@@ -132,6 +132,16 @@ var dnsTestCases = []test.Case{
 		Rcode: dns.RcodeServerFailure,
 		Ns:    miekAuth,
 	},
+	{
+		Qname: "sub.miek.nl.", Qtype: dns.TypeNS,
+		Ns: []dns.RR{
+			test.NS("sub.miek.nl.	1800	IN	NS	a.miek.nl."),
+		},
+		Extra: []dns.RR{
+			test.A("a.miek.nl.	1800	IN	A       139.162.196.78"),
+			test.AAAA("a.miek.nl.	1800	IN	AAAA	2a01:7e00::f03c:91ff:fef1:6735"),
+		},
+	},
 }
 
 const (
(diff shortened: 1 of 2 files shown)


chrisohaver (changes requested):
I don't think this is a correct change.

`// Glue returns any potential glue records for nsrrs.`

If the Nameserver (`ns.Ns`) is not a a child of the delegated zone (`ns.Header().Name`), then there is no need for "glue" records, and therefore they would not exist.  Glue records only exist to avoid cyclical dependencies.

In your example, the Nameserver is not inside the delegated domain, so there is no cyclical dependency.  The resolver can separately look up `a.miek.nl`. No glue records are required.

If the Nameserver was instead `a.sub.miek.nl`, then there would be a cyclical dependency if the resolver tried to look it up separately. i.e. Whats the IP for `a.sub.miek.nl`? Ask `a.sub.miek.nl`, it handles `sub.miek.nl`.  OK, Whats the IP for `a.sub.miek.nl`? Ask `a.sub.miek.nl`, it handles `sub.miek.nl`. ... ad infinitum
So this case requires "glue" records to exist in the parent domain (even though that domain is delegated).

victory460:
> If the Nameserver (`ns.Ns`) is not a a child of the delegated zone (`ns.Header().Name`), then there is no need for "glue" records, and therefore they would not exist. Glue records only exist to avoid cyclical dependencies.

Thanks for your review.

PR #5091, opened 2022-01-13
Log panic cause for better debugging

Tantalor93:

Hello, we are using CoreDNS in our product and we recently noticed this behavior of CoreDNS with our custom plugins

Currently, when a panic occurs in CoreDNS it is logged like this:
```
[ERROR] Recovered from panic in server: "dns://:53"
```
which is not very helpful for debugging the issue related to the panic cause. 

As part of this PR, I propose to add the cause of panic as part of the log to improve these logs, so it will for example look like this (note the last part saying that the cause was `index out of range`)
```
[ERROR] Recovered from panic in server "dns://:53" runtime error: index out of range [1] with length 0
```

none

none

I don't think so


chrisohaver (approved):
LGTM.

yongtang (approved):

PR #5093, opened 2022-01-14
plugin/kubernetes: Accept IPv4s with leading zeros

chrisohaver:

Kubernetes 1.23 allows IPv4s with leading zeros, and always interprets them as decimal.  Since Go 1.17, if we encounter such an IP, we may no longer treat it as a valid IP due to changes in net.ParseIP/net.ParseCIDR.

This PR strips leading zeros as we receive them from the K8s informers, essentially treating them as decimal, so that they are not rejected later by net.ParseIP/net.ParseCIDR operations.

From the Kubertnetes 1.23 release notes:

> Since golang 1.17 both net.ParseIP and net.ParseCIDR rejects leading zeros in the dot-decimal notation of IPv4
> addresses, Kubernetes will keep allowing leading zeros on IPv4 address to not break the compatibility.
> IMPORTANT: **Kubernetes interprets leading zeros on IPv4 addresses as decimal**, users must not rely on parser alignment
> to not being impacted by the associated security advisory: CVE-2021-29923 golang standard library "net" - Improper
> Input Validation of octal literals in golang 1.16.2 and below standard library "net" results in indeterminate SSRF
> & RFI vulnerabilities.

My understanding is that this a temporary measure by Kubernetes to maintain backward compatibility, and eventually leading zeros will be rejected - at which point we can remove this code.

### DIFF
--- a/plugin/kubernetes/object/endpoint.go
+++ b/plugin/kubernetes/object/endpoint.go
@@ -73,7 +73,7 @@ func ToEndpoints(obj meta.Object) (meta.Object, error) {
 		}
 
 		for j, a := range eps.Addresses {
-			ea := EndpointAddress{IP: a.IP, Hostname: a.Hostname}
+			ea := EndpointAddress{IP: stripLeadingZerosIPv4(a.IP), Hostname: a.Hostname}
 			if a.NodeName != nil {
 				ea.NodeName = *a.NodeName
 			}
@@ -93,7 +93,7 @@ func ToEndpoints(obj meta.Object) (meta.Object, error) {
 
 	for _, eps := range end.Subsets {
 		for _, a := range eps.Addresses {
-			e.IndexIP = append(e.IndexIP, a.IP)
+			e.IndexIP = append(e.IndexIP, stripLeadingZerosIPv4(a.IP))
 		}
 	}
 
@@ -132,6 +132,7 @@ func EndpointSliceToEndpoints(obj meta.Object) (meta.Object, error) {
 			continue
 		}
 		for _, a := range end.Addresses {
+			a = stripLeadingZerosIPv4(a)
 			ea := EndpointAddress{IP: a}
 			if end.Hostname != nil {
 				ea.Hostname = *end.Hostname
@@ -183,6 +184,7 @@ func EndpointSliceV1beta1ToEndpoints(obj meta.Object) (meta.Object, error) {
 			continue
 		}
 		for _, a := range end.Addresses {
+			a = stripLeadingZerosIPv4(a)
 			ea := EndpointAddress{IP: a}
 			if end.Hostname != nil {
 				ea.Hostname = *end.Hostname
--- a/plugin/kubernetes/object/endpoint_test.go
+++ b/plugin/kubernetes/object/endpoint_test.go
@@ -1,0 +1,67 @@
+package object
+
+import (
+	"fmt"
+	"testing"
+
+	api "k8s.io/api/core/v1"
+	discovery "k8s.io/api/discovery/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestEndpointsIPStripLeadingZeros(t *testing.T) {
+	obj := &api.Endpoints{
+		ObjectMeta: metav1.ObjectMeta{Name: "endpoint1", Namespace: "test1"},
+		Subsets: []api.EndpointSubset{{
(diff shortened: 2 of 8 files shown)


chrisohaver:
fyi, @aojea

aojea:
Cc @thockin

aojea:
> My understanding is that this a temporary measure by Kubernetes to maintain backward compatibility, and eventually leading zeros will be rejected - at which point we can remove this code.

This is the thing:
- there is no easy way to migrate/upgrade the data stored in Kubernetes, so is hard to predict when this change can be reverted
- we don't know if these ips are being used at all, but since is a possibility and kubernetes doesn't break ... :shrug: 
- the blame of the CVE goes to the consumers of the API, that are the ones that should verify there is no parsing misalignment

thinking out loud, maybe leveraging that the ecosystem invalidate the IPs and users are forced to migrate their data with weird/funny IPs in order to keep their cluster working is not a bad thing, and Kubernetes can remove the forked parsers after a prudential time.

I know that is not the best UX, but having the whole Kubernetes project with its hand tied because some users decide to use a weird representation of IPs doesn't seem legit too, to be honest, I can't imagine any good reason why people would use IPv4 addresses with leading zeros.

aojea:
Is just my opinion ... but I feel bad you have to keep this inconsistencies between releases on your project ....

chrisohaver:
To confirm @aojea, you re saying these changes are not needed, right?  Thats what I understood this weekend when I initially read it.  But now re-reading it, I am not sure if that's what you meant.  Thanks!

aojea:
> To confirm @aojea, you re saying these changes are not needed, right? Thats what I understood this weekend when I initially read it. But now re-reading it, I am not sure if that's what you meant. Thanks!

My opinion is that the ecosystem should move on and harden the parsers, since the source of the bug is parsers misalignment, keep loosening them on other projects will extend the blast radius. This way, the evolution of the ecosystem will allow, eventually, to deprecate the loose parsers in Kubernetes (the data will be valid but unusable, maybe when 1.16 is EOL?)

The question is if CoreDNS is fine dropping IPs with leading zeros, despite the data is valid in Kubernetes ... I think that both approaches are defensible

chrisohaver:
My thinking was ...

* CoreDNS currently consumes IP addresses with leading zeros from the Kubernetes API, but fails to synthesize records for them (essentially because it's ambiguous whether or not the IPs are decimal or octal representations).  But ...
* The Kubernetes API documents that all IP address data with leading zeros are interpreted as decimal. So there is no ambiguity.

It follows that the CoreDNS _kubernetes_ plugin only (not the whole of CoreDNS) should treat IP address data with leading zeros consumed from the Kubernetes API as valid decimal IPs.

If we do not do this, then strictly speaking CoreDNS violates the K8s DNS Spec. Currently, the CoreDNS _kubernetes_ plugin does store the IPs to local cache and reverse indexes with their leading zeros if present.  However these will not produce records in responses to DNS queries, since record synthesis inevitably involves parsing the strings using ParseIP. 

This PR fixes this by stripping the leading zeros when writing them to local informer stores (this IMO is a cleaner fix than stripping them out every time a DNS response record is synthesized).

> The question is if CoreDNS is fine dropping IPs with leading zeros, despite the data is valid in Kubernetes ... I think that both approaches are defensible.

CoreDNS intends to implement the [k8s dns spec](https://github.com/kubernetes/dns/blob/master/docs/specification.md).  If producing records for services/endpoints with leading zero IPs is not required by the spec, then CoreDNS is fine with dropping IPs with leading zeros.

chrisohaver:
Anyways - I'm fine with taking a wait and see approach.  Optimistically, this will all blow over without anyone noticing.  So far I have only seen one issue raised related to leading zeros, and it wasn't Kubernetes related. CoreDNS 1.8.5 (first Go1.17 compiled release) has been out for a few months.  Kubeadm's first use of a Go1.17 compiled CoreDNS release was with Kubernetes 1.23, released last month.

aojea:
> Anyways - I'm fine with taking a wait and see approach. Optimistically, this will all blow over without anyone noticing. So far I have only seen one issue raised related to leading zeros, and it wasn't Kubernetes related. CoreDNS 1.8.5 (first Go1.17 compiled release) has been out for a few months. Kubeadm's first use of a Go1.17 compiled CoreDNS release was with Kubernetes 1.23, released last month.

that is my thinking too, otherwise will not be able to break the loop :+1:

PR #5109, opened 2022-01-20
Health-checks should respect force_tcp

tombokombo:
Signed-off-by: tombokombo <tombo@sysart.tech>

Forwarding with force_tcp is using UDP proto for health-checks. I would expect that setting force_tcp for communication with upstream will use tcp for health-checks as well, but this PR is needed to do so.

### DIFF
--- a/plugin/forward/setup.go
+++ b/plugin/forward/setup.go
@@ -144,7 +144,8 @@ func parseStanza(c *caddy.Controller) (*Forward, error) {
 		}
 		f.proxies[i].SetExpire(f.expire)
 		f.proxies[i].health.SetRecursionDesired(f.opts.hcRecursionDesired)
-		if f.opts.forceTCP {
+		// when TLS is used, checks are set to tcp-tls
+		if f.opts.forceTCP && transports[i] != transport.TLS {
 			f.proxies[i].health.SetTCPTransport()
 		}
 	}
--- a/plugin/forward/setup.go
+++ b/plugin/forward/setup.go
@@ -144,6 +144,9 @@ func parseStanza(c *caddy.Controller) (*Forward, error) {
 		}
 		f.proxies[i].SetExpire(f.expire)
 		f.proxies[i].health.SetRecursionDesired(f.opts.hcRecursionDesired)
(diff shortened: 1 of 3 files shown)


chrisohaver:
See #3938. The reason for this not being accepted before is that both TCP and UDP are required per RFCs.

Although a network event that affects UDP and not TCP is not impossible.

tombokombo:
> See #3938. The reason for this not being accepted before is that both TCP and UDP are required per RFCs.
> 
> Although a network event that affects UDP and not TCP is not impossible.

Ok, haven't seen that. But now, in kubernetes days and complicated networking models, if I've choose to use tcp for upstream connection and there is a reason behind it. So I would expect that health-check will behave the same and it shouldn't matter which protocols are supported by upstream, till it supports tcp, which I choose to use (forcibly).

chrisohaver (approved):
Seems reasonable to me, it doesn't introduce any new options, and it seems inline with the existing behavior (e.g. that health checks use tls when forwarding uses tls).

Although it does change default behavior, I think the new behavior is more intuitive. I don't think the change would cause a problem in any existing deployments for health checks to start occurring over TCP instead of UDP where `force_tcp` is used.

haad:
I would very much like to have this merged.

plugin/forward/setup.go
```diff
@@ -144,7 +144,8 @@ func parseStanza(c *caddy.Controller) (*Forward, error) {
 		}
 		f.proxies[i].SetExpire(f.expire)
 		f.proxies[i].health.SetRecursionDesired(f.opts.hcRecursionDesired)
-		if f.opts.forceTCP {
+		// when TLS is used, checks are set to tcp-tls
+		if f.opts.forceTCP && transports[i] != transport.TLS {
```
chrisohaver, line 148: Good catch

tombokombo: As you mentioned TLS, I better checked that.

yongtang:
I think this PR might be fit into next minor release `v1.9.0`, (in case we want to have a patch of `v1.8.8` in between).

PR #5110, opened 2022-01-21
plugin/file: fix the lost of elem.name

victory460:

Fix the lost of elem.name when insert rr.
None.
None.
No.

### DIFF
--- a/plugin/file/tree/elem.go
+++ b/plugin/file/tree/elem.go
@@ -12,6 +12,7 @@ type Elem struct {
 func newElem(rr dns.RR) *Elem {
 	e := Elem{m: make(map[uint16][]dns.RR)}
 	e.m[rr.Header().Rrtype] = []dns.RR{rr}
+	e.name = rr.Header().Name
 	return &e
 }
 


chrisohaver:
`name` is a private field. Its public accessor is `Name()` which sets it if it is not already set.

What actual problem does this fix?

victory460:
> `name` is a private field. Its public accessor is `Name()` which sets it if it is not already set.
> 
> What actual problem does this fix?

Thanks for your review. This is a potential problem, if we use elem.name to do something. Now it has not caused serious problems for the time being, because the bottom line is handled in this function:

https://github.com/coredns/coredns/blob/7754718e1ac58d15e16a51f011976f6283aad95e/plugin/file/tree/elem.go#L57-L67
the L59 condition `e.name != ""`  always false, then every time it do a for loop assign value to e.name.

chrisohaver:
I don’t think this should merge since the current implementation is fine and I really don’t see a need to change it.

victory460:
> I don’t think this should merge since the current implementation is fine and I really don’t see a need to change it.

My previous understanding was that：

- `func (e *Elem) Name() string` is designed for get elem.name, it's not a setter, inappropriate to set elem.name.
- `func newElem(rr dns.RR) *Elem` is designed for construct a elem, and set some member of elem , such as elem.m and elem.name.

Issue #5104, opened 2022-01-19
How come some resolvers (1.1.1.1) can't resolve CNAME pointing to external domains ?


guillaumefenollar:
First thank you for that great project :) 
First of all, I'm puzzled I don't find anything related to that problem because it seems a frequent usage. I may be misusing CoreDNS or RFC 1035 files, in case I apology :) But I'm facing this issue in several environments.

I'm trying to use CoreDNS 1.8.6 (docker img) with file plugin and it can't resolve CNAME records when they return external domain names  ( ie current server is not authoritative), using some resolvers, like cloudflare and others (as stated on dnschecker.org for instance).

Here's a simple Corefile:
```
. {
  errors
  log
  reload 30s
  hebergement-web.nc {
     file db.hebergement-web.nc
  }
}
```
And the zone content : 
```
$TTL 86400
$ORIGIN hebergement-web.nc.
@      IN SOA dns.fenollar.fr. guillaume.fenollar.fr. (
                2022011901 ; serial 
                86400       ; refresh (2 hours)
                86400       ; retry (1 hour)
                1209600    ; expire (2 weeks)
                86400       ; minimum (1 hour)
                )

    86400 IN NS vps.fenollar.fr.
    86400 IN NS dns.fenollar.fr.

test IN CNAME novit.io.
```

Now a dig using Google resolver 
```
; <<>> DiG 9.16.23 <<>> test.hebergement-web.nc @8.8.8.8
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 31696
;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;test.hebergement-web.nc.	IN	A

;; ANSWER SECTION:
test.hebergement-web.nc. 21600	IN	CNAME	novit.io.
novit.io.		3600	IN	A	164.132.97.235

;; Query time: 560 msec
;; SERVER: 8.8.8.8#53(8.8.8.8)
;; WHEN: Wed Jan 19 18:16:49 +11 2022
;; MSG SIZE  rcvd: 90
```
Next same request for 1.1.1.1
```
; <<>> DiG 9.16.23 <<>> test.hebergement-web.nc @1.1.1.1
;; global options: +cmd
;; Got answer:
[...]

PR #5120, opened 2022-01-24
plugin/file: External lookup errors should not result in SERVFAIL

chrisohaver:

When doing a CNAME target lookup for a target, if the _external_ lookup fails with an error, return NOERROR.  Returning a SERVFAIL could prevent a recursive resolver from looking up the CNAME target itself. Also added a comment in the code describing this, since it is counter intuitive to return NOERROR where there is an error.

#5104 - observations suggest that for Cloudflare, the SERVFAIL response prevents the recursive resolver from looking up the CNAME target itself.

#4863 - I had changed the failure case from NOERROR to SERVFAIL fairly recently (last September).

### DIFF
--- a/plugin/file/lookup.go
+++ b/plugin/file/lookup.go
@@ -381,12 +381,7 @@ Redo:
 }
 
 func (z *Zone) doLookup(ctx context.Context, state request.Request, target string, qtype uint16) ([]dns.RR, Result) {
-	m, e := z.Upstream.Lookup(ctx, state, target, qtype)
-	if e != nil {
-		// An error here usually indicates that we have no means or don't know how to do the lookup.
-		// Return a success here so a recursive resolver can accept the lone CNAME, and do the lookup itself.
-		return nil, Success
-	}
+	m, _ := z.Upstream.Lookup(ctx, state, target, qtype)
 	if m == nil {
 		return nil, Success
 	}
--- a/plugin/file/lookup.go
+++ b/plugin/file/lookup.go
@@ -331,17 +331,13 @@ func (z *Zone) externalLookup(ctx context.Context, state request.Request, elem *
 	}
 
 	targetName := rrs[0].(*dns.CNAME).Target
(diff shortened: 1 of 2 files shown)


chrisohaver:
The failed test case is directly caused by the change.  In that test case, the CNAME target is not defined in the current zone. It is defined in another zone, but that zone happens to be a subdomain/child of the first/parent. The parent has no knowledge of the child, so _should_ be assuming that it is authoritative for the zone.  This test depends on the parent sending the query upstream if no record is found in the zone.  On the surface it seems the test is not valid, and that the parent zone should be made aware of the subdomain it is not authoritative for (e.g. delegation - although that would be odd since it's the same server).

chrisohaver:
> On the surface it seems the test is not valid, and that the parent zone should be made aware of the subdomain it is not authoritative for

Nevertheless, I'm reverting that change so the test passes. Also adjusting PR description.

miekg:
this is wrong that status of such a lookup is the last one in the chain, not somewhere in the middle.

Which text/rfc are you referencing for this change?

chrisohaver:
>... that status of such a lookup is the last one in the chain, not somewhere in the middle.

Thats true, but in this case no return code is returned - an error object is returned, because there is no next link in the chain.  I. e. CoreDNS doesn't know how to lookup the domain. The last link in the chain - the CNAME, was NOERROR.

chrisohaver:
> I. e. CoreDNS doesn't know how to lookup the domain. The last link in the chain - the CNAME, was NOERROR.

Actually - I guess the last link in the chain is itself (coredns), when it sends the query back into itself to do the lookup. Even if it cannot handle the zone, it still replies with a response code, and we should relay that instead of a hard coded response.

chrisohaver:
> should relay that instead of a hard coded response

Although this would still send a SERVFAIL back to the resolver in #5104, so it wouldn't change behavior there.  So closing this.

PR #5127, opened 2022-01-26
plugin/forward: Enable multiple forward declarations

chrisohaver:

This PR enables multiple forward plugin definitions in a server block.  
When processing a query, each forward instance is evaluated in their order in the Corefile.

This implementation chains the multiple forward plugin instances to each other via the existing `forward.Next`.  This keeps the amount of code change to a minimum.

An alternate solution is to create a slice of `Forward` structs, and loop through each one, essentially acting as a single plugin that checks multiple `Forward` instances.

This limitation comes up now and then in Q&A.  

no

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -139,8 +139,8 @@ example.org {
 }
 ~~~
 
-Send all requests within `lab.example.local.` `10.20.0.1`, all requests within `example.local.` (and not in
-`lab.example.local.`) to `10.0.0.1`, and all others requests to the servers defined in `/etc/resolv.conf`, and
+Send all requests within `lab.example.local.` to `10.20.0.1`, all requests within `example.local.` (and not in
+`lab.example.local.`) to `10.0.0.1`, all others requests to the servers defined in `/etc/resolv.conf`, and
 caches results.
 
 ~~~ corefile
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -139,6 +139,37 @@ example.org {
 }
 ~~~
 
+Send all requests within `lab.example.local.` `10.20.0.1`, all requests within `example.local.` (and not in
+`lab.example.local.`) to `10.0.0.1`, and all others requests to the servers defined in `/etc/resolv.conf`, and
+caches results.
+
+~~~ corefile
+. {
+    cache
+    forward lab.example.local 10.20.0.1
+    forward example.local 10.0.0.1
+    forward . /etc/resolv.conf
+}
+~~~
+
+The example above is almost equivalent to the following example, except that example below defines three separate plugin
+chains (and thus 3 separate instances of _cache_).
+
+~~~ corefile
+lab.example.local {
+    cache
+    forward . 10.20.0.1
+}
+example.local {
+    cache
+    forward . 10.0.0.1
+}
+. {
+    cache
+    forward . /etc/resolv.conf
+}
+~~~
+
 Load balance all requests between three resolvers, one of which has a IPv6 address.
 
 ~~~ corefile
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -141,7 +141,10 @@ example.org {
 
 Send all requests within `lab.example.local.` to `10.20.0.1`, all requests within `example.local.` (and not in
 `lab.example.local.`) to `10.0.0.1`, all others requests to the servers defined in `/etc/resolv.conf`, and
-caches results.
+caches results. Note that a CoreDNS server configured with multiple _forward_ plugins in a server block will evaluate those
+forward plugins in the order they are listed when serving a request.  Therefore, subdomains should be
+placed before parent domains otherwise subdomain requests will be forwarded to the parent domain's upstream.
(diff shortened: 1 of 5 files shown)


chrisohaver:
I kind of prefer taking the "slice of Forward structs" approach outlined in the description.  But since this is a much smaller change, I figured I'd present it first.

chrisohaver:
FWIW - A case came up again in a recent [support issue](https://github.com/kubernetes/dns/issues/525#issuecomment-1162983491), where allowing multiple forward declarations in the root server block would enable a better solution.

Tantalor93 (commented):
@chrisohaver could you extend README of forward plugin and add mention that this plugin can be specified multiple times in a single server block (thats not common pattern in CoreDNS => worth mentioning in README), what is the behaviour in such case and maybe some usecases? Also please add an example into the README of Corefile with multiple forward declarations?

chrisohaver:
Thanks for reviewing!

> thats not common pattern in CoreDNS

Actually, it is a common pattern in CoreDNS, but worth illustrating with an example. I think plugins that enforce only one per server block are the exception, hence those plugins having a note in their README, "This plugin can only be used once per Server Block."

> Also please add an example into the README of Corefile with multiple forward declarations?

Agreed. I will add an example that illustrate a use case for multiple instances.

chrisohaver:
@Tantalor93, It's been a looong time since I looked at this (January).

I want to build it and play with it a bit to make sure it works as intended before merging.

plugin/forward/README.md
```diff
@@ -141,6 +139,37 @@ example.org {
 }
 ~~~
 
+Send all requests within `lab.example.local.` `10.20.0.1`, all requests within `example.local.` (and not in
```
Tantalor93: typo
```suggestion
Send all requests within `lab.example.local.` to `10.20.0.1`, all requests within `example.local.` (and not in
```

chrisohaver:
> It's been a looong time since I looked at this (January). I want to build it and play with it a bit to make sure it works as intended ...

OK, I have tested using local forwarding and the forwarding aspect works as expected.

I've also added a note about plugin order in the Corefile, since (as is the case in some other multi-instance plugins) relative order of the instances in the server block matters.

Tantalor93 (approved):
LGTM, thanks!

Tantalor93 (commented):
I just noticed that `parseForward` is not used anymore in production code (it is replaced by `parseMultiForward`), could you remove it?

chrisohaver:
> I just noticed that `parseForward` is not used anymore in production code (it is replaced by `parseMultiForward`), could you remove it?

IIRC it is used by the *alternate* external plugin. So for now, I was going to leave it in.

chrisohaver:
> > I just noticed that `parseForward` is not used anymore in production code (it is replaced by `parseMultiForward`), could you remove it?
> 
> IIRC it is used by the _alternate_ external plugin. So for now, I was going to leave it in.

Hmm. Maybe not that plugin, but I thought an external plugin used it? I should have commented why I left it in. This PR had been sitting waiting for a review for so long I forgot the details. 

Thanks for dedicating the time to reviewing PRs!

Tantalor93:
> > > I just noticed that `parseForward` is not used anymore in production code (it is replaced by `parseMultiForward`), could you remove it?
> > 
> > 
> > IIRC it is used by the _alternate_ external plugin. So for now, I was going to leave it in.
> 
> Hmm. Maybe not that plugin, but I thought an external plugin used it? I should have commented why I left it in. This PR had been sitting waiting for a review for so long I forgot the details.

I think that `parseForward` cannot be directly used by any other plugin, as it is not exported function of `forward` package and therefore not visible from other GO packages and any other plugin


> Thanks for dedicating the time to reviewing PRs!

no problem, this feature looks very useful 👍

chrisohaver:
> but I thought an external plugin used it?

it couldn’t have been used by another package because it’s a private function. Oh. I think I remember why I thought this. I think it was once a public function, and it was used by an external plugin, but later it was made private in a refactor, breaking the plugin. And the other plugin was refactored. 

TLDR, I can remove it.

zouyee (approved):

chrisohaver:
@Tantalor93, How do I interpret the linter failure?  It says there exists trailing whitespace but doesn't say where.

chrisohaver:
rebasing, to see if that helps the cryptic linter fail... no it didnt.

chrisohaver:
> rebasing, to see if that helps the cryptic linter fail... no it didnt.

I think I figured it out - golinter makes comments in-line in the commits.
I guess since the offending files are outside of this PR, they didn't boil up into the main PR page. Or maybe they are always buried like that.  I don't know.

Tantalor93:
@chrisohaver yea, golangci-lint annotates code directly (see https://github.com/golangci/golangci-lint-action/issues/5)

Tantalor93 (approved):

chrisohaver:
> @chrisohaver yea, golangci-lint annotates code directly (see [golangci/golangci-lint-action#5](https://github.com/golangci/golangci-lint-action/issues/5))

Yeah - weird that those comments don't show up in the main discussion.  That means for a large commit, you'd have to go hunting for them.

PR #5128, opened 2022-01-26
Add stale github action

yongtang:
This PR adds stale github action, so that stale issues or PRs will be merged when there is no activity for 30+7 days.

We can polish on fine control after some experiments.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,15 @@
+name: 'Close Stale Issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v4
+        with:
+          stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 7 days'
+          stale-pr-message: 'This pull request is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 7 days'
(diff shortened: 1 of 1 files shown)


greenpau (approved):

chrisohaver (approved):
I think this is OK for now.

It would be better if we had a way of knowing if the issue stale because it's waiting for a review from maintainers, or if it's waiting for the author to address review feedback.  I.e. if the ball is in the maintainer's court, we shouldn't be marking it stale. But without a workflow defined this will be impossible to script. I've been doing this manually, going back and reading comments in PRs and making that determination before pinging authors/closing PRs.

Maybe we can add a "ready for review" tag that users can set via a comment, and that we can unset when we review or something like that.

chrisohaver:
For whatever reason - the action only marked 7 issues stale.  The ones marked stale were more recent issues. Much older issues that should have been marked stale, were not.

Perhaps the stale action only looks so far back?

chrisohaver:
> Perhaps the stale action only looks so far back?

Perhaps it's `operations-per-run`, which defaults to 30:

https://github.com/actions/stale#operations-per-run

Maybe it fetches issues one at a time, or in small batches.  Fetch, add tag, comment, each being a separate operation per open issue/pr.

yongtang:
@chrisohaver  I think we can wait a little to see if remaining ones are being picked up accordingly over time. This is also helpful to us, as over the years many issues have been cumulated. We can take some (30) at a time (over the next several weeks ) to find out if we need to add additional flags or measures to make sure issues are not incorrectly being closed.

PR #5131, opened 2022-01-27
plugin/cache: Add refresh mode setting to serve_stale

atollena:

This PR adds an optional REFRESH_MODE parameter on the serve_stale configuration directive of the cache plugin. For context, please see #4309

Reasoning and RFC discussed at #4309
Other similar issues are #4097 and #4199.

I updated the cache plugin README.md to document the new parameter on the PR.

No.

### DIFF
--- a/plugin/cache/README.md
+++ b/plugin/cache/README.md
@@ -37,7 +37,7 @@ cache [TTL] [ZONES...] {
     success CAPACITY [TTL] [MINTTL]
     denial CAPACITY [TTL] [MINTTL]
     prefetch AMOUNT [[DURATION] [PERCENTAGE%]]
-    serve_stale [DURATION]
+    serve_stale [DURATION] [REFRESH_MODE]
 }
 ~~~
 
@@ -57,7 +57,12 @@ cache [TTL] [ZONES...] {
 * `serve_stale`, when serve\_stale is set, cache always will serve an expired entry to a client if there is one
   available.  When this happens, cache will attempt to refresh the cache entry after sending the expired cache
   entry to the client. The responses have a TTL of 0. **DURATION** is how far back to consider
-  stale responses as fresh. The default duration is 1h.
+  stale responses as fresh. The default duration is 1h. **REFRESH_MODE** controls when the attempt to refresh
+  the cache happens. `verified` will first verify that an entry is still unavailable from the source before sending
+  the stale response to the client. `immediate` will immediately send the expired response to the client before
+  checking to see if the entry is available from the source. **REFRESH_MODE** defaults to `immediate`. Setting this
+  value to `verified` can lead to increased latency when serving stale responses, but will prevent stale entries
+  from ever being served if an updated response can be retrieved from the source.
 
 ## Capacity and Eviction
 
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -38,7 +38,9 @@ type Cache struct {
 	duration   time.Duration
 	percentage int
 
-	staleUpTo time.Duration
+	// Stale serve
+	staleUpTo   time.Duration
+	verifyStale bool
 
 	// Testing.
 	now func() time.Time
@@ -225,6 +227,33 @@ func (w *ResponseWriter) Write(buf []byte) (int, error) {
 	}
 	n, err := w.ResponseWriter.Write(buf)
 	return n, err
+}
+
+// verifyStaleResponseWriter is a response writer that only writes messages if they should replace a
+// stale cache entry, and otherwise discards them.
+type verifyStaleResponseWriter struct {
+	*ResponseWriter
+	refreshed bool // set to true if the last WriteMsg wrote to ResponseWriter, false otherwise.
+}
+
+// newVerifyStaleResponseWriter returns a ResponseWriter to be used when verifying stale cache
+// entries. It only forward writes if an entry was successfully refreshed according to RFC8767,
+// section 4 (response is NoError or NXDomain), and ignores any other response.
+func newVerifyStaleResponseWriter(w *ResponseWriter) *verifyStaleResponseWriter {
+	return &verifyStaleResponseWriter{
+		w,
+		false,
+	}
+}
+
+// WriteMsg implements the dns.ResponseWriter interface.
+func (w *verifyStaleResponseWriter) WriteMsg(res *dns.Msg) error {
+	w.refreshed = false
+	if res.Rcode == dns.RcodeSuccess || res.Rcode == dns.RcodeNameError {
+		w.refreshed = true
+		return w.ResponseWriter.WriteMsg(res) // stores to the cache and send to client
+	}
+	return nil // else discard
 }
 
 const (
--- a/plugin/cache/handler.go
+++ b/plugin/cache/handler.go
@@ -35,19 +35,29 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 
 	ttl := 0
 	i := c.getIgnoreTTL(now, state, server)
-	if i != nil {
-		ttl = i.ttl(now)
-	}
 	if i == nil {
 		crr := &ResponseWriter{ResponseWriter: w, Cache: c, state: state, server: server, do: do}
 		return c.doRefresh(ctx, state, crr)
 	}
+	ttl = i.ttl(now)
 	if ttl < 0 {
-		servedStale.WithLabelValues(server, c.zonesMetricLabel).Inc()
+		// serve stale behavior
+		if c.verifyStale {
+			crr := &ResponseWriter{ResponseWriter: w, Cache: c, state: state, server: server, do: do}
+			cw := newVerifyStaleResponseWriter(crr)
+			ret, err := c.doRefresh(ctx, state, cw)
+			if cw.refreshed {
+				return ret, err
+			}
+		}
+
 		// Adjust the time to get a 0 TTL in the reply built from a stale item.
 		now = now.Add(time.Duration(ttl) * time.Second)
-		cw := newPrefetchResponseWriter(server, state, c)
-		go c.doPrefetch(ctx, state, cw, i, now)
+		if !c.verifyStale {
+			cw := newPrefetchResponseWriter(server, state, c)
+			go c.doPrefetch(ctx, state, cw, i, now)
+		}
+		servedStale.WithLabelValues(server, c.zonesMetricLabel).Inc()
 	} else if c.shouldPrefetch(i, now) {
 		cw := newPrefetchResponseWriter(server, state, c)
 		go c.doPrefetch(ctx, state, cw, i, now)
@@ -70,7 +80,7 @@ func (c *Cache) doPrefetch(ctx context.Context, state request.Request, cw *Respo
 	}
 }
 
-func (c *Cache) doRefresh(ctx context.Context, state request.Request, cw *ResponseWriter) (int, error) {
+func (c *Cache) doRefresh(ctx context.Context, state request.Request, cw dns.ResponseWriter) (int, error) {
 	if !state.Do() {
 		setDo(state.Req)
 	}
--- a/plugin/cache/cache_test.go
+++ b/plugin/cache/cache_test.go
@@ -266,7 +266,7 @@ func TestServeFromStaleCache(t *testing.T) {
 	req.SetQuestion("cached.org.", dns.TypeA)
 	ctx := context.TODO()
 
-	// Cache example.org.
+	// Cache cached.org. with 60s TTL
 	rec := dnstest.NewRecorder(&test.ResponseWriter{})
 	c.staleUpTo = 1 * time.Hour
 	c.ServeDNS(ctx, rec, req)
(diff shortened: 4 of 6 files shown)


atollena:
I'm taking over the work from @Towerthousand on this, started in https://github.com/coredns/coredns/pull/4756. This commit implements suggestions from @chrisohaver:

1. properly return stale records on nxdomain in accordance with rfc 8767
2. rename before/after to immediate/verify

I also refactored the tests slightly to make it easier to test the nxdomain case.

In addition, I think it would be valuable to go a bit further and implement the following bits of rfc8767:

1. Add a failure recheck timer to prevents querying the upstream for a period of time (potentially configurable) -- we have cases where we are worried that when an upstream is down, it gets hammered by clients when coming back. This would help in this scenario.
2. Add a client response timer to respond with stale records early, when a stale record is available. Since we plan to use this for intra-DC DNS, we can set this very low, which reduces latency before getting a stale entry.
3. make the TTL of stale entries  configurable (at the moment it is hardcoded at 0, which the RFC explicitly forbids). Ideally I think we'd go with 30s by default, as recommended by the RFC, but this is a backward incompatible change.

I'd be willing to do any of those items, either directly here or in a follow-up PR, but I'm interested to get feedback before implementing them.

plugin/cache/README.md
```diff
@@ -57,7 +57,10 @@ cache [TTL] [ZONES...] {
 * `serve_stale`, when serve\_stale is set, cache always will serve an expired entry to a client if there is one
   available.  When this happens, cache will attempt to refresh the cache entry after sending the expired cache
   entry to the client. The responses have a TTL of 0. **DURATION** is how far back to consider
-  stale responses as fresh. The default duration is 1h.
+  stale responses as fresh. The default duration is 1h. **REFRESH_MODE** controls whether the attempt to refresh
+  the cache happens before (`immediate`) or after (`verified`) returning a response. This defaults to `immediate`.
```
chrisohaver: > refresh the cache happens before (`immediate`) or after (`verified`) returning a response

"immediate" and "verified" are swapped here, resulting in incorrect description.
And the phrasing is confusing, probably better described using the option terminology.

* `verified` will first _verify_ that an entry is still unavailable from the source before sending the stale response to the client.
* `immediate` will _immediately_ send a response to the client before checking to see if the entry is available from the source.

plugin/cache/handler.go
```diff
@@ -38,9 +38,13 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
 	}
-	if i == nil {
+	if i == nil || (ttl < 0 && c.verifyStale) {
+		// try to resolve upstream if we have no cached record or the cached one is stale and serve_stale is set to verify.
 		crr := &ResponseWriter{ResponseWriter: w, Cache: c, state: state, server: server, do: do}
-		return c.doRefresh(ctx, state, crr)
+		rcode, err := c.doRefresh(ctx, state, crr)
+		if i == nil || (rcode == dns.RcodeSuccess || rcode == dns.RcodeNameError){
+			return rcode, err
+		}
```
chrisohaver: Calling `c.doRefresh` generally results in sending a response to the client if a response is received - no matter what the response code is. As proposed, if the response code is not NXDOMAIN or NOERROR, the client can be sent two responses, the first one being the current upstream response, and the second one (which would be ignored by the client) being the stale cached entry.

chrisohaver (changes requested):

plugin/cache/README.md
atollena: fixed, thank you for the suggestion.

plugin/cache/handler.go
atollena: OK, I realise that I did not understand your comment in https://github.com/coredns/coredns/pull/4756#discussion_r672453981.

I wrapped the response writer as you suggested, so that it only writes if the response is nxdomain or noerror and exposes whether it wrote. I'm not quite sure that the behaviour is correct in case the underlying `ResponseWriter` returns an error to `WriteMsg`. As proposed the code just ignores the error and proceeds with serving stale entry -- my concern is whether this could in certain cases lead to double writing.

atollena:
@chrisohaver thanks for the prompt review, I think I addressed your comments.

plugin/cache/README.md
```diff
@@ -57,7 +57,10 @@ cache [TTL] [ZONES...] {
 * `serve_stale`, when serve\_stale is set, cache always will serve an expired entry to a client if there is one
   available.  When this happens, cache will attempt to refresh the cache entry after sending the expired cache
   entry to the client. The responses have a TTL of 0. **DURATION** is how far back to consider
-  stale responses as fresh. The default duration is 1h.
+  stale responses as fresh. The default duration is 1h. **REFRESH_MODE** controls when the attempt to refresh
+  the cache happens. `verified` will first verify that an entry is still unavailable from the source before sending
+  the stale response to the client. `immediate` will immediately send the expired response to the client before 
+  checking to see if the entry is available from the source.  **REFRESH_MODE** defaults to `immediate`.
```
chrisohaver: Please add a warning about that latency will be higher when using `verified`.  It was present in an earlier commit, but looks like it was removed in this last commit.

plugin/cache/cache.go
```diff
@@ -225,6 +227,34 @@ func (w *ResponseWriter) Write(buf []byte) (int, error) {
...
+
+// WriteMsg implements the dns.ResponseWriter interface.
+func (w *verifyStaleResponseWriter) WriteMsg(res *dns.Msg) error {
+	mt, _ := response.Typify(res, w.now().UTC())
+	w.refreshed = false
+	if mt == response.NoError || mt == response.NameError {
+		w.refreshed = true
+		return w.ResponseWriter.WriteMsg(res) // stores to the cache and send to client
+	}
+	return nil // else discard
```
chrisohaver: This should check the response code of the message directly (`res.Rcode == dns.RcodeSuccess || res.Rcode == dns.RcodeNameError`), rather than running the message through `Typify` which returns pseudo-codes (like `NoData`, which has a `NOERROR` return code).

chrisohaver (changes requested):

plugin/cache/README.md
atollena: I restored the warning (I changed a few words but kept the content).

plugin/cache/cache.go
atollena: Done.

plugin/cache/README.md
```diff
@@ -57,7 +57,12 @@ cache [TTL] [ZONES...] {
 * `serve_stale`, when serve\_stale is set, cache always will serve an expired entry to a client if there is one
   available.  When this happens, cache will attempt to refresh the cache entry after sending the expired cache
   entry to the client. The responses have a TTL of 0. **DURATION** is how far back to consider
-  stale responses as fresh. The default duration is 1h.
+  stale responses as fresh. The default duration is 1h. **REFRESH_MODE** controls when the attempt to refresh
+  the cache happens. `verified` will first verify that an entry is still unavailable from the source before sending
+  the stale response to the client. `immediate` will immediately send the expired response to the client before 
+  checking to see if the entry is available from the source.  **REFRESH_MODE** defaults to `immediate`. Setting this
+  value to `verified` can lead to increased latency when serving stale responses for unavailable upstreams, but can
+  reduce the number of stale entries served, especially when resolution is infrequent.
```
chrisohaver: Latency is always increased for serving stale entries with `verified` ...  not just for unavailable upstreams/sources.  

"can reduce" -> "will prevent": `verified` prevents stale entries from ever being served if an updated response can be retrieved from the source. 

```suggestion
  value to `verified` can lead to increased latency when serving stale responses, but will
  prevent stale entries from ever being served if an updated response can be retrieved from the source.
```

> especially when resolution is infrequent

I think better to remove this qualification. `verified` always prevents this, regardless of resolution frequency.  Conversely, `immediate` will serve  1 extra stale result to a client before updating the cache.  If there is high concurrency on the cache item, then many of those stale responses can be written to clients before the cache is updated.

chrisohaver (changes requested):
Code LGTM.  Just some minor tweaks to the readme now.

plugin/cache/README.md
atollena: I took your suggestion.

chrisohaver (approved):

atollena:
I realised by looking at the coverage report that I didn't properly test discarding messages that are not success or nxdomain, so I improved the test a bit to write a response in case of servfail.

atollena:
Thanks again @chrisohaver for the quick review rounds. Is there anything I can or should do to help get this change merged? (e.g. squashing the git commits, or anything else)

chrisohaver:
> Thanks again @chrisohaver for the quick review rounds. Is there anything I can or should do to help get this change merged?

Unless a PR is really trivial, I usually defer final approval and merging to a codeowner of the related plugin. In doing reviews of code that I am not a codeowner for, my hope is that this effort makes their reviews easier and PRs merge more quickly.

atollena:
IIUC @miekg is the sole owner of the cache plugin. Miek, do you have any feedback for this change?

I'm also interested to have your input on [this comment above](https://github.com/coredns/coredns/pull/5131#issuecomment-1023208795) regarding further work on serve_stale behaviors -- let me know if you'd rather have that filed as an issue for further discussion.

chrisohaver:
I'm going to merge this in the next couple of days unless anyone objects.

atollena:
> I'm going to merge this in the next couple of days unless anyone objects.

Thanks. I merged master and squashed the commits.

chrisohaver (approved):

plugin/cache/README.md
```diff
@@ -57,7 +57,12 @@ cache [TTL] [ZONES...] {
 * `serve_stale`, when serve\_stale is set, cache always will serve an expired entry to a client if there is one
   available.  When this happens, cache will attempt to refresh the cache entry after sending the expired cache
   entry to the client. The responses have a TTL of 0. **DURATION** is how far back to consider
-  stale responses as fresh. The default duration is 1h.
+  stale responses as fresh. The default duration is 1h. **REFRESH_MODE** controls when the attempt to refresh
+  the cache happens. `verified` will first verify that an entry is still unavailable from the source before sending
```
lobshunter, line 61: A minor typo here, `verified` will cause a invalid value error

```suggestion
  the cache happens. `verify` will first verify that an entry is still unavailable from the source before sending
```

chrisohaver: @lobshunter, This PR merged a couple months ago, and the typo has since been fixed.

lobshunter: My bad, I checked out to this PR to see how it works, didn't notice the new fix. 😂

salla2:
> I'm taking over the work from @Towerthousand on this, started in #4756. This commit implements suggestions from @chrisohaver:
> 
> 1. properly return stale records on nxdomain in accordance with rfc 8767
> 2. rename before/after to immediate/verify
> 
> I also refactored the tests slightly to make it easier to test the nxdomain case.
> 
> In addition, I think it would be valuable to go a bit further and implement the following bits of rfc8767:
> 
> 1. Add a failure recheck timer to prevents querying the upstream for a period of time (potentially configurable) -- we have cases where we are worried that when an upstream is down, it gets hammered by clients when coming back. This would help in this scenario.
> 2. Add a client response timer to respond with stale records early, when a stale record is available. Since we plan to use this for intra-DC DNS, we can set this very low, which reduces latency before getting a stale entry.
> 3. make the TTL of stale entries  configurable (at the moment it is hardcoded at 0, which the RFC explicitly forbids). Ideally I think we'd go with 30s by default, as recommended by the RFC, but this is a backward incompatible change.
> 
> I'd be willing to do any of those items, either directly here or in a follow-up PR, but I'm interested to get feedback before implementing them.

@atollena Did you tried implementing above changes? especially failure recheck timer to prevents querying the upstream for a period of time ?

atollena:
> > I'm taking over the work from @Towerthousand on this, started in #4756. This commit implements suggestions from @chrisohaver:
> > 
> > 1. properly return stale records on nxdomain in accordance with rfc 8767
> > 2. rename before/after to immediate/verify
> > 
> > I also refactored the tests slightly to make it easier to test the nxdomain case.
> > In addition, I think it would be valuable to go a bit further and implement the following bits of rfc8767:
> > 
> > 1. Add a failure recheck timer to prevents querying the upstream for a period of time (potentially configurable) -- we have cases where we are worried that when an upstream is down, it gets hammered by clients when coming back. This would help in this scenario.
> > 2. Add a client response timer to respond with stale records early, when a stale record is available. Since we plan to use this for intra-DC DNS, we can set this very low, which reduces latency before getting a stale entry.
> > 3. make the TTL of stale entries  configurable (at the moment it is hardcoded at 0, which the RFC explicitly forbids). Ideally I think we'd go with 30s by default, as recommended by the RFC, but this is a backward incompatible change.
> > 
> > I'd be willing to do any of those items, either directly here or in a follow-up PR, but I'm interested to get feedback before implementing them.
> 
> @atollena Did you tried implementing above changes? especially failure recheck timer to prevents querying the upstream for a period of time ?

I did not and do not plan to work on this further. Feel free to take that over.

Issue #5144, opened 2022-01-31
1.9.0 release tracking


chrisohaver:
Proposing we release 1.9.0 now, which has some PRs queued up for it that change behavior in a backward incompatible way (although no configuration backward incompatibilities that would require Corefile updates)

* #5019 (issue: #4984)
* #5109

I don't think there are any pressing reasons to release a 1.8.8.

yongtang:
Created a PR #5145 to add initial release note.

I am also thinking about change to use GitHub Actions to build and push container to dockerhub. This can be done whenever a release tag is created and triggered automatically. We had some issues in the past with docker. I will see if I can get this part in before 1.9.0.

yongtang:
Created a PR #5146 for Github Actions to do docker release.

yongtang:
Let's get 1.9.0 out first, as all have been covered.

chrisohaver:
Has version been updated?

yongtang:
Yes https://github.com/coredns/coredns/pull/5145/commits/b88a5e1681ee0591052c0b2e397add4274c54e65

yongtang:
https://github.com/coredns/coredns/pull/5145/files

chrisohaver:
@yongtang, if you don't have time to troubleshoot the new docker release action workflow failures.
Maybe just release the old way (via comment).

yongtang:
The first error is due to rate limiting of GitHub API I believe so some download will be rate limited.

yongtang:
The second time error seems to be due to the dockerhub credential.

yongtang:
The 1.8.7's docker build failure might be also coming from rate limiting:
https://github.com/coredns/coredns/issues/4996#issuecomment-1013158130

So the response is not an tar.gz archive but HTTP rate limit I guess.

chrisohaver:
> The 1.8.7's docker build failure might be also coming from rate limiting:
https://github.com/coredns/coredns/issues/4996#issuecomment-1013158130

I think that 1.8.7 failure was due to the release artifacts not being completely uploaded yet. i.e. it downloaded an incomplete file because it was still in the process of uploading. IIRC, I ran `/docker 1.8.7` as soon as I saw notifications that the _release_ was created.  However, this was before the release _script_ finished and wrote its status to the issue.

yongtang:
With --retry this issue will be solved. Though the GitHub Actions runner and the corebot does not have the installed curl version newer enough to handle 404 retries.

[...]

PR #5146, opened 2022-02-01
Add github actions for docker release.

yongtang:
Once a release/tag shown up in GitHub, an GitHub Action will be automatically triggered for docker release.

This will help avoid additional steps when releasing.

See comment https://github.com/coredns/coredns/issues/5144#issuecomment-1026257712

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,22 @@
+name: Docker Release
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  docker-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+          set -x -e
+          echo "Docker release ${{ github.event.release.tag_name }}:"
+          release=${{ github.event.release.tag_name }}
+          version=${release:1}
+          export DOCKER_LOGIN=${{ secrets.DOCKERHUB_USERNAME }}
+          export DOCKER_PASSWORD=${{ secrets.DOCKERHUB_PASSWORD }}
+          make VERSION=${version} DOCKER=coredns -f Makefile.docker release
+          docker images
+          make VERSION=${version} DOCKER=coredns -f Makefile.docker docker-push
+          echo "Docker release ${{ github.event.release.tag_name }} successful"
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -19,11 +19,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Build Docker Images
-        run: |
-          set -x -e
-          make VERSION=${RELEASE:1} DOCKER=yongtang -f Makefile.docker release
-          docker images
+        run: make VERSION=${RELEASE:1} DOCKER=coredns -f Makefile.docker release
+      - name: Show Docker Images
+        run: docker images
       - name: Publish Docker Images
-        run: |
-          set -x -e
-          make VERSION=${RELEASE:1} DOCKER=yongtang -f Makefile.docker docker-push
+        run: make VERSION=${RELEASE:1} DOCKER=coredns -f Makefile.docker docker-push
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,20 +3,27 @@ name: Docker Release
 on:
   release:
     types: [published]
+  workflow_dispatch:
+    inputs:
+      release:
+        description: "Release (e.g., v1.9.0)"
+        required: true
 
 jobs:
   docker-release:
     runs-on: ubuntu-latest
+    env:
+      DOCKER_LOGIN: ${{ secrets.DOCKERHUB_USERNAME }}
+      DOCKER_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
+      RELEASE: ${{ github.event.inputs.release || github.event.release.tag_name }}
     steps:
       - uses: actions/checkout@v2
-      - run: |
+      - name: Build Docker Images
+        run: |
           set -x -e
-          echo "Docker release ${{ github.event.release.tag_name }}:"
-          release=${{ github.event.release.tag_name }}
-          version=${release:1}
-          export DOCKER_LOGIN=${{ secrets.DOCKERHUB_USERNAME }}
-          export DOCKER_PASSWORD=${{ secrets.DOCKERHUB_PASSWORD }}
-          make VERSION=${version} DOCKER=coredns -f Makefile.docker release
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,22 @@
...
+  release:
+    types: [published]
+
+jobs:
+  docker-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+          set -x -e
```
SuperQ: Rather than do this, please break up the steps.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,22 @@
...
+          echo "Docker release ${{ github.event.release.tag_name }}:"
+          release=${{ github.event.release.tag_name }}
+          version=${release:1}
+          export DOCKER_LOGIN=${{ secrets.DOCKERHUB_USERNAME }}
+          export DOCKER_PASSWORD=${{ secrets.DOCKERHUB_PASSWORD }}
```
SuperQ: These should be done as environment variables

```yaml
jobs:
  docker-release:
    env:
      DOCKER_LOGIN: ${{ secrets.DOCKERHUB_USERNAME }}
...
```

SuperQ (changes requested):
Please break up the one-big-step.

chrisohaver (commented):
Can this be manually triggered if needed? I'm thinking of a case where the action fails, and a workflow update is required to fix it.

.github/workflows/docker.yml
yongtang: @SuperQ  Updated now.

.github/workflows/docker.yml
yongtang: @SuperQ  Done.

yongtang:
> Can this be manually triggered if needed? I'm thinking of a case where the action fails, and a workflow update is required to fix it.

@chrisohaver  You can now go into "Actions" => "Docker Release" => "Run workflow" and provide a version to publish.

.github/workflows/docker.yml
SuperQ: This `set -x -e` is unnecessary. The GitHub action is already set this way.

```suggestion
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,29 @@
...
+      DOCKER_LOGIN: ${{ secrets.DOCKERHUB_USERNAME }}
+      DOCKER_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
+      RELEASE: ${{ github.event.inputs.release || github.event.release.tag_name }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build Docker Images
+        run: |
+          set -x -e
+          make VERSION=${RELEASE:1} DOCKER=yongtang -f Makefile.docker release
+          docker images
```
SuperQ: Again, this should be a separate step.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Publish Docker Images
+        run: |
+          set -x -e
+          make VERSION=${RELEASE:1} DOCKER=yongtang -f Makefile.docker docker-push
```
SuperQ: Simplify

```suggestion
        run: make VERSION=${RELEASE:1} DOCKER=yongtang -f Makefile.docker docker-push
```

SuperQ (changes requested):

chrisohaver (commented):
Is there a chance that this could automatically trigger before the release is complete?  For example that the tag is created before all the binaries are available to download?

yongtang:
@SuperQ  PR updated. Please take a look.

yongtang:
> Is there a chance that this could automatically trigger before the release is complete? For example that the tag is created before all the binaries are available to download?

@chrisohaver  The trigger is:
```
  release:
    types: [published]
```

A tag does not does not necessarily mean it is a release. Only a published release (with downloadable binaries in place) can trigger a build.

SuperQ (approved):
Nice!

chrisohaver:
> A tag does not does not necessarily mean it is a release. Only a published release (with downloadable binaries in place) can trigger a build.

Looks like first the release is made, and then files are uploaded to the release.

https://github.com/coredns/coredns/blob/7d9ad0872c644c693bbb9e8864e2c19627586ad2/Makefile.release#L89-L110

I think the action will trigger after the release is created on line 90, and then depending on how quickly the action triggers, it may try to download the binaries that are still in the process of uploading.

yongtang:
@chrisohaver  I think this can be resolved by only create draft when release is being created.

After draft is created, either manually click the button to release, or automatically actually publish the draft when all assets have been updated.

Let me create a separate PR to split out draft vs. publish case.

chrisohaver:
> Let me create a separate PR to split out draft vs. publish case.

Why a separate PR?  Seems like a small change, with the singular purpose of making this PR work should just be included in this PR.

yongtang:
@chrisohaver we are touching two areas. One is for release binary and another is the trigger depending on the other one. Unfortunately, until it truly invoked during the release, we don't know for sure if it works or not. 

I want to have the ability to revert either one (or both) if any issue happens.

chrisohaver:
> Unfortunately, until it truly invoked during the release, we don't know for sure if it works or not.

Perhaps it should be tested on a fork first.

yongtang:
@chrisohaver  The github actions for docker release has been tested on my own repo and it works fine.
https://github.com/yongtang/coredns/actions

But, the other one of Makefile.release is invoked through Miek's bot. I don't have access to that one. So I would rather split those two items to two PRs.

chrisohaver:
> The github actions for docker release has been tested on my own repo and it works fine.

That's surprising, perhaps the trigger is slow enough that the action ends up running after all the binaries are uploaded. Since it works fine in your fork, let's close the other PR.

chrisohaver:
One minor thing: I think moving to GitHub actions _might_ result in inconsistent docker versions, as it did with Go versions.
That was a bigger deal in Go, but for docker probably much less likely to cause an issue.  But something to keep in mind if things go awry.

yongtang:
@chrisohaver  The other PR is technically not needed. The docker release can be run manually with a click.

yongtang:
@chrisohaver No the `types: [published]` will fetch the version of the published release, not the current master.

chrisohaver:
> @chrisohaver No the `types: [published]` will fetch the version of the published release, not the current master.

No. I'm referring to the version of docker, and Golang.  Not coredns.

We had issues in the past with GitHub actions getting spun up on instances that ran inconsistent versions of Golang installed.  I am suggesting that the same could occur with the version of docker installed. Although, docker version differences are probably much less likely to cause problems.

Issue #3034, opened 2019-07-23
plugin/template Respond for zone match and regex mismatch?


stp-ip:
```
.:5053 {
      template IN PTR in-addr.arpa {
        # 1.0.0.127.in-addr.arpa
        match "^(?P<a>[0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])[.](?P<b>[0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])[.](?P<c>[0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])[.](?P<d>[0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])[.]in-addr[.]arpa[.]$"
        answer "{{ .Name }} 86400 IN PTR {{ .Group.d }}-{{ .Group.c }}-{{ .Group.b }}-{{ .Group.a }}.net.example"
      }
}
```

It seems currently we don't respond to anything that matches the zone, but not the regex.
Shouln't we at least SERVFAIL?

Usually `fallthrough` + forward will lead to a NXDOMAIN, but this feels like an unhandled case here, but could just be my thinking.

chrisohaver:
I think either a NODATA or NXDOMAIN response would be closer to "correct" than SERVFAIL.  But automatically determining the right kind of negative response would be hard - I think it would require parsing the regex itself to determine all valid sub domains - which is perhaps theoretically possible but impractical. 

I think in the "spirit" of _template_, we might want to introduce a new option, "default", which allows the user to control the response code when the query does not match. e.g. 

`default ( NXDOMAIN | NOERROR | SERVFAIL )`

stp-ip:
I think we should default to SERVFAIL at least and giving the user the option with default would be an additional iteration. Currently we seem to just not respond at all.

chrisohaver:
Ironically, IIUC SERVFAIL would never be the actual desired answer in any given single case.  NXDOMAIN would sometimes be the desired answer, as would NODATA sometimes be the desired answer. But since we cant be sure which is the correct answer, then I suppose SERVFAIL "works".

We could return NODATA for any queries that match the zone exactly but doesn't match the regex. This is technically correct if we assume that a plugin claiming to be authoritative for a zone actually is.

johnbelamaric:
[...]

PR #5156, opened 2022-02-07
Stale-bot: Do not mark issues stale when pending maintainer action

yongtang:
If an issue has been labeled as 'answered' or 'needs info', or a PR has been labeled with 'needs update' by maintainers,
then it will be marked with stale after 30 days and +7 days to close.

Any activity in between will unstale, and 'answered,needs info' in issues and 'needs update' in PRs will be removed.

So authors can exempt the issue or PR from stale as long as there is some activity.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -15,3 +15,6 @@ jobs:
           days-before-close: 7
           exempt-issue-labels: 'enhancement'
           exempt-all-milestones: true
+          labels-to-remove-when-unstale: 'answered,needs info,needs update'
+          any-of-issue-labels: 'answered,needs info'
+          any-of-pr-labels: 'needs update'
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -9,9 +9,9 @@ jobs:
     steps:
       - uses: actions/stale@v4
         with:
-          stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 7 days'
-          stale-pr-message: 'This pull request is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 7 days'
-          days-before-stale: 30
-          days-before-close: 7
+          stale-issue-message: 'This issue is stale because it has been open 90 days with no activity. Remove stale label or comment or this will be closed in 15 days'
+          stale-pr-message: 'This pull request is stale because it has been open 90 days with no activity. Remove stale label or comment or this will be closed in 15 days'
+          days-before-stale: 90
+          days-before-close: 15
           exempt-issue-labels: 'enhancement'
           exempt-all-milestones: true
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -9,9 +9,11 @@ jobs:
     steps:
       - uses: actions/stale@v4
         with:
-          stale-issue-message: 'This issue is stale because it has been open 90 days with no activity. Remove stale label or comment or this will be closed in 15 days'
-          stale-pr-message: 'This pull request is stale because it has been open 90 days with no activity. Remove stale label or comment or this will be closed in 15 days'
-          days-before-stale: 90
-          days-before-close: 15
+          stale-issue-message: 'This issue is stale because it has been open 60 days with no activity. Remove stale label or comment or this will be closed in 15 days'
+          stale-pr-message: 'This pull request is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 7 days'
+          days-before-issue-stale: 60
+          days-before-issue-close: 15
+          days-before-pr-stale: 30
+          days-before-pr-close: 7
           exempt-issue-labels: 'enhancement'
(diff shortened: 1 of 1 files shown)


chrisohaver (changes requested):
I dunno. I think 30+7 days is fine. I think it will keep things moving.   For the most part it will act like a monthly reminder to maintainers of open PRs awaiting review. A reminder once every 90 days wouldn’t be as effective.

yongtang:
@chrisohaver For PRs, 30+7 might be fine.

Though for issues however, 30+7 tends to be a little short. My observation is that discussions can last multiple months before it settle for a conclusion that a feature request is accepted (and confirmed with "enhancement" label which will be whitelisted by stale).

yongtang:
@chrisohaver  Note in stale's default value it is actually 60+7:
https://github.com/actions/stale

yongtang:
@chrisohaver  Updated with PR 30+7, issue 60+15.

chrisohaver:
I think 30 is fine for issues too. I don’t see a difference. If we don’t reply to an issue in 30 days, we probably won’t ever look at it, unless someone pings us. 90 just slows down the pings.

chrisohaver (changes requested):
I think the initial 30+7 numbers are fine for now.
If after a couple of months we find that notifications for idle issues/prs every 30 days is too frequent, we can increase them. But I don't think we can draw that conclusion after a just a week.

The already implemented exception is enhancements, which will frequently sit idle a long time before someone offers to implement them, and then idle a long time before a PR is opened.

Bugs/questions, and PRs are more urgent, and should get our more immediate attention.  Being reminded once every 30 days that these are lingering is fine IMO.  Longer than that kind of defeats the purpose.  IOW, it's not desirable for a question or bug or PR to sit idle for more than 30 days.

yongtang:
There are two scenarios:
1. An PR/issue has been reviewed by a maintainer and some opinion has been provided. The question is back to the original PR/issue author.
2. An PR/issue is awaiting maintainers' attention and actions/responses. (

Issue #3034 is an example of 2).

In case of 1), it is clearly the original PR/issue author's responsibility to reply back to unstale.

In case of 2), I am not sure we should let the original PR/issue author to constantly ping maintainers and interact with a stale-bot, as some people (in community) may not feel good with it.

It is understandable that many coredns maintainers work in spare time. The ongoing COVID also puts a difficult situation for many maintainers to response in time. However, at the minimum I think what coredns can do is to not let a community member to constantly ping or interact a stale bot, while awaiting maintainer's response.

I have added a label of "needs maintainer response", and created a PR #5165 to whitelist this label. That will server as an remainder to maintainers (and put the responsibility of reply to maintainers).

chrisohaver:
It's not easy for us to automate a means of determining if an issue/pr is pending maintainer action or author action.
This will require maintainer review on whether or not something should go stale.

A more conservative approach could be to leave stale labeling to maintainers, but still automate the closing of issues/prs that have been marked stale for a certain amount of time.  For example, maintainers add the stale label when it looks like an issue is stalled waiting for author action. Once that label is added, a comment is made on issue/pr will close in X days if they don't respond (comment could be automated by a bot when someone adds the stale label).  If someone does respond before that time the stale label is removed.

A slightly more automated approach: When an issue/pr is opened, we can safely assume the state is not awaiting the author at this point, and needs a review.  When the issue/pr is reviewed but not resolved (closed or merged), the responder/reviewer adds a label indicating that the issue is waiting for author's response (e.g. "pending author" or something like that).  A stale-bot would then only mark items stale that have that label, and then proceed to close them if X days pass with no updates.  This is in essence the inverse of "needs maintainer review" label you proposed, but I think works better since new issues/prs land in the correct state. In either inversion, we'd need to go through the existing backlog of issues/prs once, to make sure they are properly labeled (we don't have many, so this is not a big chore).

chrisohaver:
> A slightly more automated approach: When an issue/pr is opened, we can safely assume the state is not awaiting the author at this point, and needs a review. When the issue/pr is reviewed but not resolved (closed or merged), the responder/reviewer adds a label indicating that the issue is waiting for author's response (e.g. "pending author" or something like that). A stale-bot would then only mark items stale that have that label, and then proceed to close them if X days pass with no updates. This is in essence the inverse of "needs maintainer review" label you proposed, but I think works better since new issues/prs land in the correct state. In either inversion, we'd need to go through the existing backlog of issues/prs once, to make sure they are properly labeled (we don't have many, so this is not a big chore).

I think this will work, while preventing subscribers of issues/prs waiting for maintainer response from getting spammed by stale-bot.  Since in this model the stale bot does not act as a periodic reminder of issues/prs pending maintainer input, the stale/close times can be increased to accommodate author's schedules if desired.

yongtang:
If an issue has been labeled as  'answered' or 'needs info', or a PR has been labeled with 'needs update' by maintainers,
then it will be marked with stale after 30 days and +7 days to close.

Any activity in between will unstale, and 'answered,needs info' in issues and 'needs update' in PRs will be removed.

So authors can exempt the issue or PR from stale as long as there is some activity.

chrisohaver (approved):
LGTM!

PR #5165, opened 2022-02-07
Add "needs maintainer response" label and as part of whitelisted stale label.

yongtang:
This PR adds "needs maintainer response" label and as part of whitelisted stale label.

Previously the stale will whitelist a PR already in milestone, or an issue labeled as enhamcement (confirmed by coredns maintainers).

However, there are situations where an issue has not been confirmed by coredns maintainer as "enhancement" and is awaiting maintainer's opinion/response.

See PR #3034 as an example.

In that case, we don't want the community member to constantly ping maintainers constantly
to unstale the issue.

Instead, those issues labeled as "needs maintainer response" should caught maintainer's attention.

This PR will replace #5156

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -13,5 +13,5 @@ jobs:
           stale-pr-message: 'This pull request is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 7 days'
           days-before-stale: 30
           days-before-close: 7
-          exempt-issue-labels: 'enhancement'
+          exempt-issue-labels: 'enhancement,needs maintainer response'
           exempt-all-milestones: true


chrisohaver (changes requested):
I don’t think it’s a good idea to exclude PRs or issues that need our attention from generating reminders.  I think those reminders would help us help us catch forgotten issues/PRs that are “in our court”.  It’s has already helped.

yongtang:
Well it looks like there are some disagreement here:

In case of https://github.com/coredns/coredns/issues/3034, the issue was supposed to be closed in 7 days, unless one of the maintainers mentioned in the issue replied, then this issue is set to be closed.

I am concerned about who is supposed to be responsible to keep this issue open? Or if nobody in the issue answers, the issue will be closed.

But, just to point out that the last conversation from issue #3034 is from community member looking for response from maintainer. (This is different from some other scenarios where community member is expected to provide a response after maintainers provided a response.)

At the moment this stale bot is  making me feeling not comfortable when community members might have to deal with a bot waiting for a response from maintainers.

chrisohaver:
Let's move the conversation to #5156.  No need to have same conversation in two separate PRs. The changes in this PR should have been added #5156 anyways, as it is editing the same file in an attempt to solve the same problems, and a result of feedback from #5156.

yongtang:
I am seeing several issues:

1. 30 days is a  too short to me. It is very hard for me to request any community member to response back in 30 days, when the COVID pandemic is still roaming. It is not cool.
   (for the same reason, I don't want to set a rule for any maintainer to reply in 30 days as well)

2. When community members are asking questions and expect maintainers to answer. I don't want to see community members to interact with a bot every 30 days. It is also not so cool.

How to solve the above issues? Either we increase the length to be long enough (I was thinking at least 90 days), or we have a way to not ask the community member to interact with a bot constantly (when they are expecting some maintainer to provide a response).

Or we do both and in that case it will be in one PR. But, I am seeing disagreement on both proposals. And from my side, I also don't feel comfortable with the current status.

chrisohaver:
Please, let's move this conversation to one PR, #5156.

I think my comment in https://github.com/coredns/coredns/pull/5156#issuecomment-1032843984 addresses the above concerns.

chrisohaver:
Superseded by #5156

PR #5171, opened 2022-02-09
plugin/grpc: dont panic when from-zone cannot be normalized

chrisohaver:

Don't panic when from-zone cannot be normalized.

Addresses issue TOB-CDNS-15 in https://github.com/trailofbits/publications/blob/master/reviews/CoreDNS.pdf


zouyee (approved):
/lgtm

PR #5176, opened 2022-02-10
Fix response should keep option when request with it

sstoner:

As we know that the cache plugin will cache items by removing OPT RR. Then the response from cached will reuse the request's OPT RR. 

When the response msg with edns0 and without Option is not from the cache plugin, the client received MSG size is different.

#5182


chrisohaver:
@sstoner, Please give a detailed example of the problem you are facing.

sstoner:
@chrisohaver I created a related issue that I'm facing, please see #5182

PR #5180, opened 2022-02-10
plugin/template: Return SERVFAIL for zone-match regex-no-match case

### LINKED ISSUE #3034: plugin/template Respond for zone match and regex mismatch?

chrisohaver:

This fixes the condition when a template's zone matches, but its match statements no not match the query, and there is no `fallthrough` specified.

The intent of the original code was to return `NXDOMAIN`, but it just returned to the plugin chain without actually writing a response to client, resulting in no response to the client.  However, `NXDOMAIN` is not always the correct response anyways (see discussion in #3034).

So, this PR makes this condition a `SERVFAIL`. So instead of hanging the client, a `SERVFAIL` is returned.

A correct `NXDOMAIN` / `NODATA` response can be implemented using multiple _template_ definitions using the `fallthrough` option.

The PR also cleans up some minor README formatting issues, readability and doc/behavior drift.

closes #3034

Included

No. Although I suppose some people could intentionally be abusing this bug as a black hole mechanism.

### DIFF
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -17,23 +17,24 @@ template CLASS TYPE [ZONE...] {
     additional RR
     authority RR
     rcode CODE
-    fallthrough [ZONE...]
+    fallthrough [FALLTHROUGH-ZONE...]
 }
 ~~~
 
 * **CLASS** the query class (usually IN or ANY).
 * **TYPE** the query type (A, PTR, ... can be ANY to match all types).
 * **ZONE** the zone scope(s) for this template. Defaults to the server zones.
-* **REGEX** [Go regexp](https://golang.org/pkg/regexp/) that are matched against the incoming question name. Specifying no regex matches everything (default: `.*`). First matching regex wins.
+* `match` **REGEX** [Go regexp](https://golang.org/pkg/regexp/) that are matched against the incoming question name. 
+  Specifying no regex matches everything (default: `.*`). First matching regex wins.
 * `answer|additional|authority` **RR** A [RFC 1035](https://tools.ietf.org/html/rfc1035#section-5) style resource record fragment
-  built by a [Go template](https://golang.org/pkg/text/template/) that contains the reply.
+  built by a [Go template](https://golang.org/pkg/text/template/) that contains the reply. Specifying no answer will result
+  in a response with an empty answer section.
 * `rcode` **CODE** A response code (`NXDOMAIN, SERVFAIL, ...`). The default is `NOERROR`. Valid response code values are
   per the `RcodeToString` map defined by the `miekg/dns` package in `msg.go`.
-* `fallthrough` Continue with the next plugin if the zone matched but no regex matched.
-  If specific zones are listed (for example `in-addr.arpa` and `ip6.arpa`), then only queries for
-  those zones will be subject to fallthrough.
-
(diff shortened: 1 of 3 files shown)


stp-ip:
As before I still feel SERVFAIL as implemented here is good and agree with the fallthrough thinking on NXDOMAIN and NODATA.
Would love another opinion so as this does feel like a not clear cut thing.

chrisohaver:
> As before I still feel SERVFAIL as implemented here is good and agree with the fallthrough thinking on NXDOMAIN and NODATA.

I agree.

chrisohaver:
More detailed thoughts on this...

* SERVFAIL essentially reflects here that there is a hole in the configuration... kind of like the no-next-plugin case.
* It's practically impossible for us to automatically know what the right answer is (NXDOMAIN/NODATA) by analyzing the regex pattern, current request, and zone. This is because we need to return NODATA when the query is a super domain of any other possible query that would match the regex.
* It would also be impractical for us to analyze the zone/regex at setup time to determine if this condition can occur (i.e. to reject it as an invalid configuration).
* We discussed adding a "default" option in #3034, but this would not always work correctly, since the NXDOMAIN/NODATA answer would not be the same for all queries that fall in the zone, but do not match the regexes.
* If we were OK with being wrong sometimes, and we were to always respond with one of NXDOMAIN or NODATA, then the response should have an SOA in the Authority section.  The contents of which would need to be configured somewhere, which means more options.  We already have a way of adding an SOA to the Authority section, but would need a means of signifying that it is only to be included when no match regexes match. However, you can already do the equivalent of this using the existing `fallthrough` option to fall through to other template instances that responds with NXDOMAIN/NODATA and the SOA in the Authority section.  So there is no need to add new options. These fall through instances could also be crafted to return the correct NXDOMAIN or NODATA response in relation to the first instance's regex, if a user cares about returning the correct negative response.

Another consideration could be to _always_  fall through when this condition occurs (when a query matches the zone but not any regexes), even when `fallthrough` is not specified. IOW, auto fall though instead of returning SERVFAIL.  This is different than specifying `fallthrough` because it doesn't apply when the zone doesn't match.

chrisohaver:
Resolved conflicts.  Please review.

zouyee (approved):

PR #5183, opened 2022-02-11
geoip: read source IP from EDNS0 subnet if provided

julian7:
Signed-off-by: Balazs Nagy <julsevern@gmail.com>

This patch implements EDNS backend processing (similar in powerdns: https://doc.powerdns.com/authoritative/settings.html#setting-edns-subnet-processing). This feature comes very handy to test whether your geo config is working properly.

Initiating such a test is as simple as:

```shell
$ dig +subnet=1.2.3.4 @dns-server.example.com www.geo-aware.com
```

I couldn't find any issues related to this feature.

Probably it's a good idea to mention this in coredns-geoip manual.

This is handled as an option in powerdns, but I'm not sure whether it makes sense not to handle EDNS0 subnets.

/cc @miekg @snebel29

### DIFF
--- a/plugin/geoip/README.md
+++ b/plugin/geoip/README.md
@@ -42,18 +42,16 @@ or
 
 ```text
 geoip [DBFILE] {
-    edns-subnet
+    [edns-subnet]
 }
 ```
 
-By default, geoip plugin doesn't honor EDNS0 subnet extension, but it can be enabled using the extended syntax:
-
 * **DBFILE** the mmdb database file path.
-* `edns-subnet` (optional): by default, geoip plugin read client IP and ignores [EDNS0 subnet](https://en.wikipedia.org/wiki/EDNS_Client_Subnet) extension. Using this option enables reading source IP from EDNS0 subnet when available. This help identifying closest source IP address as well as making GeoIP test processing easy: `dig +subnet=1.2.3.4 @dns-server.example.com www.geo-aware.com`.
+* `edns-subnet`: Optional. Use [EDNS0 subnet](https://en.wikipedia.org/wiki/EDNS_Client_Subnet) (if present) for Geo IP instead of the source IP of the DNS request. This helps with identifying the closest source IP address as well as making GeoIP testing easy: `dig +subnet=1.2.3.4 @dns-server.example.com www.geo-aware.com`.
 
 ## Examples
 
-The following configuration configures the `City` database.
+The following configuration configures the `City` database, and looks up geolocation based on EDNS0 subnet if present.
 
 ```txt
 . {
--- a/plugin/geoip/README.md
+++ b/plugin/geoip/README.md
@@ -34,23 +34,22 @@ You can download a [free and public City database](https://dev.maxmind.com/geoip
 
 ## Syntax
 
-```txt
+```text
 geoip [DBFILE]
 ```
 
-* **DBFILE** the mmdb database file path.
+or
 
-By default, geoip plugin doesn't honor EDNS0 subnet extension, but it can be enabled using the extended syntax:
-
-```
+```text
 geoip [DBFILE] {
     edns-subnet-processing
 }
 ```
 
-* **DBFILE**: see above
-* `edns-subnet-processing` reads source IP from EDNS0 subnet when available. This makes
-  testing GeoIP processing easy: `dig +subnet=1.2.3.4 @dns-server.example.com www.geo-aware.com`.
+By default, geoip plugin doesn't honor EDNS0 subnet extension, but it can be enabled using the extended syntax:
+
+* **DBFILE** the mmdb database file path.
+* `edns-subnet-processing` (optional): by default, geoip plugin read client IP and ignores [EDNS0 subnet](https://en.wikipedia.org/wiki/EDNS_Client_Subnet) extension. Using this option enables reading source IP from EDNS0 subnet when available. This help identifying closest source IP address as well as making GeoIP test processing easy: `dig +subnet=1.2.3.4 @dns-server.example.com www.geo-aware.com`.
 
 ## Examples
 
@@ -58,7 +57,9 @@ The following configuration configures the `City` database.
 
 ```txt
 . {
-    geoip /opt/geoip2/db/GeoLite2-City.mmdb
+    geoip /opt/geoip2/db/GeoLite2-City.mmdb {
+      edns-subnet-processing
+    }
     metadata # Note that metadata plugin must be enabled as well.
 }
 ```
--- a/plugin/geoip/README.md
+++ b/plugin/geoip/README.md
@@ -47,7 +47,7 @@ geoip [DBFILE] {
 ```
 
 * **DBFILE** the mmdb database file path.
-* `edns-subnet`: Optional. Use [EDNS0 subnet](https://en.wikipedia.org/wiki/EDNS_Client_Subnet) (if present) for Geo IP instead of the source IP of the DNS request. This helps with identifying the closest source IP address as well as making GeoIP testing easy: `dig +subnet=1.2.3.4 @dns-server.example.com www.geo-aware.com`.
+* `edns-subnet`: Optional. Use [EDNS0 subnet](https://en.wikipedia.org/wiki/EDNS_Client_Subnet) (if present) for Geo IP instead of the source IP of the DNS request. This helps identifying the closest source IP address through intermediary DNS resolvers, and it also makes GeoIP testing easy: `dig +subnet=1.2.3.4 @dns-server.example.com www.geo-aware.com`.
 
 ## Examples
 
--- a/plugin/geoip/README.md
+++ b/plugin/geoip/README.md
@@ -46,8 +46,12 @@ geoip [DBFILE] {
 }
 ```
 
-* **DBFILE** the mmdb database file path.
+* **DBFILE** the mmdb database file path. We recommend updating your mmdb database periodically for more accurate results.
 * `edns-subnet`: Optional. Use [EDNS0 subnet](https://en.wikipedia.org/wiki/EDNS_Client_Subnet) (if present) for Geo IP instead of the source IP of the DNS request. This helps identifying the closest source IP address through intermediary DNS resolvers, and it also makes GeoIP testing easy: `dig +subnet=1.2.3.4 @dns-server.example.com www.geo-aware.com`.
+
+  **NOTE:** due to security reasons, recursive DNS resolvers may mask a few bits off of the clients' IP address, which can cause inaccuracies in GeoIP resolution.
+
+  There is no defined mask size in the standards, but there are examples: [RFC 7871's example](https://datatracker.ietf.org/doc/html/rfc7871#section-13) conceals the last 72 bits of an IPv6 source address, and NS1 Help Center [mentions](https://help.ns1.com/hc/en-us/articles/360020256573-About-the-EDNS-Client-Subnet-ECS-DNS-extension) that ECS-enabled DNS resolvers send only the first three octets (eg. /24) of the source IPv4 address.
 
 ## Examples
 
--- a/plugin/geoip/README.md
+++ b/plugin/geoip/README.md
@@ -40,6 +40,18 @@ geoip [DBFILE]
 
 * **DBFILE** the mmdb database file path.
 
+By default, geoip plugin doesn't honor EDNS0 subnet extension, but it can be enabled using the extended syntax:
+
+```
+geoip [DBFILE] {
+    edns-subnet-processing
+}
+```
+
+* **DBFILE**: see above
+* `edns-subnet-processing` reads source IP from EDNS0 subnet when available. This makes
+  testing GeoIP processing easy: `dig +subnet=1.2.3.4 @dns-server.example.com www.geo-aware.com`.
+
 ## Examples
 
 The following configuration configures the `City` database.
--- a/plugin/geoip/README.md
+++ b/plugin/geoip/README.md
@@ -1,9 +1,11 @@
 # geoip
 
 ## Name
+
 *geoip* - Lookup maxmind geoip2 databases using the client IP, then add associated geoip data to the context request.
 
 ## Description
+
 The *geoip* plugin add geo location data associated with the client IP, it allows you to configure a [geoIP2 maxmind database](https://dev.maxmind.com/geoip/docs/databases) to add the geo location data associated with the IP address.
 
 The data is added leveraging the *metadata* plugin, values can then be retrieved using it as well, for example:
@@ -16,8 +18,8 @@ import (
 // ...
 if getLongitude := metadata.ValueFunc(ctx, "geoip/longitude"); getLongitude != nil {
     if longitude, err := strconv.ParseFloat(getLongitude(), 64); err == nil {
-		// Do something useful with longitude.
-	}
+        // Do something useful with longitude.
+    }
 } else {
     // The metadata label geoip/longitude for some reason, was not set.
 }
@@ -25,18 +27,23 @@ if getLongitude := metadata.ValueFunc(ctx, "geoip/longitude"); getLongitude != n
 ```
 
 ## Databases
+
 The supported databases use city schema such as `City` and `Enterprise`. Other databases types with different schemas are not supported yet.
 
 You can download a [free and public City database](https://dev.maxmind.com/geoip/geolite2-free-geolocation-data).
 
 ## Syntax
+
 ```txt
 geoip [DBFILE]
 ```
+
 * **DBFILE** the mmdb database file path.
 
 ## Examples
+
 The following configuration configures the `City` database.
+
 ```txt
 . {
     geoip /opt/geoip2/db/GeoLite2-City.mmdb
@@ -45,6 +52,7 @@ The following configuration configures the `City` database.
 ```
 
 ## Metadata Labels
+
 A limited set of fields will be exported as labels, all values are stored using strings **regardless of their underlying value type**, and therefore you may have to convert it back to its original type, note that numeric values are always represented in base 10.
 
 | Label                                | Type      | Example          | Description
--- a/plugin/geoip/README.md
+++ b/plugin/geoip/README.md
@@ -42,14 +42,14 @@ or
 
 ```text
 geoip [DBFILE] {
-    edns-subnet-processing
+    edns-subnet
 }
 ```
 
 By default, geoip plugin doesn't honor EDNS0 subnet extension, but it can be enabled using the extended syntax:
 
 * **DBFILE** the mmdb database file path.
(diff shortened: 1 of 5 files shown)


snebel29 (commented):
I think it make sense to look for EDNS0 subnet option and use it instead of the request source IP, when it is provided.

The code looks good at first glance, I will have a second read later on.

Thanks for the additional cleanup, I think it is also good to wait for @miekg to have a look as well.

snebel29:
@julian7 I think a little mention of ENDS0 subnet option being read in the plugin documentation (README.md) would be helpful.

@miekg I think this is good to merge, would you like to thumb this up as well?

Cheers.

chrisohaver (changes requested):
Thanks! Please update /plugin/geoip/README.md.

The only reason I can think of making this an option is that it does make it a bit easier for someone to "spoof" or mask their geo location.  I think this is probably minor, since geo data is usually used for stats, and not for policy driven decisions. So not making this an option is _probably_ OK.  We can add an option later to disable it, if someone opens an issue down the road.

julian7:
After a long pondering break, I decided to put this extra functionality behind extra configuration. First, we're dealing with a GA feature, changing the behavior of such an item can cause trouble. Second, this is also not turned on in PowerDNS by default (perhaps because of the same reason), but at least it won't cause surprises.

I've added some documentation too.

plugin/geoip/README.md
```diff
@@ -40,6 +40,18 @@ geoip [DBFILE]
...
 
+By default, geoip plugin doesn't honor EDNS0 subnet extension, but it can be enabled using the extended syntax:
+
+```
+geoip [DBFILE] {
+    edns-subnet-processing
+}
+```
+
+* **DBFILE**: see above
```
snebel29: **Nit**

I find adding once again here DBFILE argument a bit redundant, given that is already defined above. I also changed the wording along with syntax and examples section... What do you think?

## Syntax
```txt
geoip [DBFILE] {
    edns-subnet-processing
}
```

* **DBFILE** the mmdb database file path.
* `edns-subnet-processing` By default, geoip plugin read client IP and ignores EDNS0 subnet extension, using this option enable reading source IP from [EDNS0 subnet](https://en.wikipedia.org/wiki/EDNS_Client_Subnet) when available. This help identifying closest source IP address as well as making GeoIP test processing easy: `dig +subnet=1.2.3.4 @dns-server.example.com www.geo-aware.com`.

## Examples

The following configuration configures the `City` database and enables EDNS subnet processing.

```txt
. {
    geoip /opt/geoip2/db/GeoLite2-City.mmdb {
        edns-subnet-processing
    }
    metadata # Note that metadata plugin must be enabled as well.
}
```

julian7: I think it's a good idea, but I'd mention that `edns-subnet-processing` is optional, as well as the { } block. Otherwise, I'm happy with the rewording.

snebel29: From my side, sounds good to be explicit with the optional nature of the argument.

julian7: Please have a look, and open a conversation if something else is needed.

plugin/geoip/README.md
```diff
@@ -16,35 +18,54 @@ import (
...
+
+or
+
+```text
+geoip [DBFILE] {
+    edns-subnet-processing
+}
+```
+
+By default, geoip plugin doesn't honor EDNS0 subnet extension, but it can be enabled using the extended syntax:
```
chrisohaver:
```suggestion
```

plugin/geoip/README.md
```diff
@@ -16,35 +18,54 @@ import (
...
+
+```text
 geoip [DBFILE]
 ```
+
+or
+
+```text
+geoip [DBFILE] {
+    edns-subnet-processing
```
chrisohaver:
```suggestion
    [edns-subnet-processing]
```

plugin/geoip/README.md
```diff
@@ -16,35 +18,54 @@ import (
...
+}
+```
+
+By default, geoip plugin doesn't honor EDNS0 subnet extension, but it can be enabled using the extended syntax:
+
 * **DBFILE** the mmdb database file path.
+* `edns-subnet-processing` (optional): by default, geoip plugin read client IP and ignores [EDNS0 subnet](https://en.wikipedia.org/wiki/EDNS_Client_Subnet) extension. Using this option enables reading source IP from EDNS0 subnet when available. This help identifying closest source IP address as well as making GeoIP test processing easy: `dig +subnet=1.2.3.4 @dns-server.example.com www.geo-aware.com`.
```
chrisohaver:
```suggestion
* `edns-subnet-processing`  Optional. For the geo IP address, use [EDNS0 subnet](https://en.wikipedia.org/wiki/EDNS_Client_Subnet) (if present) instead of the source IP of the DNS request. This helps with identifying the closest source IP address as well as making GeoIP testing easy: `dig +subnet=1.2.3.4 @dns-server.example.com www.geo-aware.com`.
```

It's not clear what you mean by "This helps with identifying the closest source IP address".

plugin/geoip/README.md
```diff
@@ -16,35 +18,54 @@ import (
...
 
 ## Examples
+
 The following configuration configures the `City` database.
+
 ```txt
 . {
-    geoip /opt/geoip2/db/GeoLite2-City.mmdb
+    geoip /opt/geoip2/db/GeoLite2-City.mmdb {
+      edns-subnet-processing
```
chrisohaver: The description of the example should reflect this. e.g. "The following configuration configures the `City` database, and looks up geo location based on edns0 subnet if present.

chrisohaver (dismissed):
Thanks!, Just some minor nits on readme.

plugin/geoip/README.md
chrisohaver: Another _very_ minor thing: `edns-subnet-processing` is kind of a long option name.  e.g. I think just `edns-subnet` would be sufficient.

julian7: `edns-subnet-processing` is taken from PowerDNS options, verbatim. We're not tied to pdns though :) I'm editing it shortly.

plugin/geoip/README.md
julian7: @chrisohaver I'm theorycrafting here, but in theory, public recursors can forward original requestors' IP addresses in an EDNS0 subnet attachment. We just open the gates for such a behavior.

Otherwise, good changes.

plugin/geoip/README.md
julian7: Good point, thanks.

plugin/geoip/README.md
julian7: huh did that line make it so far?

chrisohaver: I think that this is described sufficiently in the syntax section, and doesn't require more documentation.

chrisohaver:
>> It's not clear what you mean by "This helps with identifying the closest source IP address".

>  I'm theorycrafting here, but in theory, public recursors can forward original requestors' IP addresses in an EDNS0 subnet attachment. We just open the gates for such a behavior.

Thanks that is what I though you meant.  How about:

"This can help identify a source IP address closer to the client when the client's dns request is processed by middleboxes (e.g. forwarding, proxy, NAT)".

plugin/geoip/README.md
julian7: I was reusing this conversation for edns-subnet processing benefits. Now a possible benefit for intermediary dns recursors is mentioned at "edns-subnet" config option.

chrisohaver (commented):
Thanks! The readme looks good to me!

plugin/geoip/README.md
```diff
@@ -47,7 +47,7 @@ geoip [DBFILE] {
 ```
 
 * **DBFILE** the mmdb database file path.
-* `edns-subnet`: Optional. Use [EDNS0 subnet](https://en.wikipedia.org/wiki/EDNS_Client_Subnet) (if present) for Geo IP instead of the source IP of the DNS request. This helps with identifying the closest source IP address as well as making GeoIP testing easy: `dig +subnet=1.2.3.4 @dns-server.example.com www.geo-aware.com`.
+* `edns-subnet`: Optional. Use [EDNS0 subnet](https://en.wikipedia.org/wiki/EDNS_Client_Subnet) (if present) for Geo IP instead of the source IP of the DNS request. This helps identifying the closest source IP address through intermediary DNS resolvers, and it also makes GeoIP testing easy: `dig +subnet=1.2.3.4 @dns-server.example.com www.geo-aware.com`.
```
chrisohaver, line 50: I think "intermediary DNS resolvers" might be technically imperfect description here  - _but I think its fine as is_.  The specific situation that ends-subnet aims to assist is when any intermediate network "stuff" changes the source ip address of the request packet in transit, thus obfuscating the original client's IP.  More specifically, when the new IP address is not "near" the client.

IOW - it's not just "intermediate DNS resolvers" that can change the IP, other common situations can change the source IP, and ECS also solves those situations.

julian7: I agree, it's a nice 80% solution :) I'm fine with that though.

julian7:
Maybe I've missed it. Is there anything else I can do about this change?

chrisohaver:
> Maybe I've missed it. Is there anything else I can do about this change?

I don't think so.  The PR should be reviewed and approved by a code owner for this plugin.
I'm not a code owner for geoip, and I'm not at all familiar with the geoip backend database.
That said, looking at the code, my understanding is that edns0 subnet typically holds a an IP prefix aka subnet, not a complete IP address.  I see pulling the address portion of the subnet, and using that as a db lookup on IP addresses as problematic.

snebel29:
> I see pulling the address portion of the subnet, and using that as a db lookup on IP addresses as problematic.

Good point @chrisohaver, a simple solution could be to support complete addresses only by ignoring `EDNS0_SUBNET` if:

- Family == ipV4 && SourceMask != 32
- Family == ipV6 && SourceMask != 128

Thoughts?

chrisohaver:
> > I see pulling the address portion of the subnet, and using that as a db lookup on IP addresses as problematic.
> 
> Good point @chrisohaver, a simple solution could be to support complete addresses only by ignoring `EDNS0_SUBNET` if:
> 
> * Family == ipV4 && SourceMask != 32
> * Family == ipV6 && SourceMask != 128
> 
> Thoughts?

Yeah. That seems reasonable.

julian7:
> > > I see pulling the address portion of the subnet, and using that as a db lookup on IP addresses as problematic.
> > 
> > 
> > Good point @chrisohaver, a simple solution could be to support complete addresses only by ignoring `EDNS0_SUBNET` if:
> > 
> > * Family == ipV4 && SourceMask != 32
> > * Family == ipV6 && SourceMask != 128
> > 
> > Thoughts?
> 
> Yeah. That seems reasonable.

Counter-example: In [RFC 7871 section 13](https://datatracker.ietf.org/doc/html/rfc7871#section-13) example, the recursive DNS resolver conceals the last 72 bits of the source IP address, leaving only the first 56 through for geoip resolution. Likewise, in [NS1's ECS document](https://help.ns1.com/hc/en-us/articles/360020256573-About-the-EDNS-Client-Subnet-ECS-DNS-extension), they mention that recursive DNS resolvers often hide the last byte of an IPv4 address.

chrisohaver:
>> Yeah. That seems reasonable.

> Counter-example ...

That seems supportive rather than counter. For example, IPv4s in adjacent /25 subnets could be in different locations. If an IP from one of those /25s is masked at 24 bits, the look up algorithm could select the wrong location.

Perhaps there are some generally adopted practices that make this not a problem. E.g. for ipv4, if...
* geo location data is never more granular than /24 for ipv4
* resolvers never mask out more than the last byte of an ipv4

Then we could ignore ipv4 ECS when mask < 24, and accept everything >= 24.

But I suspect those bullets are not universally true.

julian7:
> > > Yeah. That seems reasonable.
> 
> > Counter-example ...
> 
> That seems supportive rather than counter. For example, IPv4s in adjacent /25 subnets could be in different locations. If an IP from one of those /25s is masked at 24 bits, the look up algorithm could select the wrong location.

The amount of bits hidden is not our choice. The one who made the choice has to deal with inaccuracies caused by hiding bits.

> Perhaps there are some generally adopted practices that make this not a problem. E.g. for ipv4, if...
> 
> * geo location data is never more granular than /24 for ipv4
> * resolvers never mask out more than the last byte of an ipv4
> 
> Then we could ignore ipv4 ECS when mask < 24, and accept everything >= 24.

Generally speaking, if any edns subnet is provided, it is more accurate than the source IP. The worst case is if the database can’t tell where an IP address (it works with ip ranges too) is coming from.

Usually, inaccuracies in geoip are both common and benign. There are cases when geoip is used as a geofence, but i’d rather recommend against making harder decisions like GDPR enforcement based on such information.

Most inaccuracies are coming from the database, as in most cases, a stale version of the free version is used. There were just one or two cases I could see anything significant because of an error like this, over five years of running a geoip-aware service.

If someone wants to implement such an extension to the module, that’s great. I, however, still think it is not worth the effort.

snebel29:
I guess users configuring `edns-subnet` option (which is not set by default) would have to evaluate themselves their use case, along with the geoIP database used and the impact of potential inaccuracies.

From my side, and unless anyone feel against I will have a second read of the entire change since it'd being a while the last time I've read it, then if everything is alright I'll merge.

Thanks.

chrisohaver:
> I guess users configuring edns-subnet option (which is not set by default) would have to evaluate themselves their use case, along with the geoIP database used and the impact of potential inaccuracies.

Those potential inaccuracies should be documented in the readme so users can evaluate it before deploying it.

snebel29:
> Those potential inaccuracies should be documented in the readme so users can evaluate it before deploying it.

Agreed.

snebel29:
@julian7 Thanks for your contribution and patience! Would you mind adding a note in the plugin documentation about those inaccuracies derived from (potentially) looking up using IP with "hidden" bits? That way users would be in a better position deciding to enable `edns-subnet` or not.

julian7:
> @julian7 Thanks for your contribution and patience! Would you mind adding a note in the plugin documentation about those inaccuracies derived from (potentially) looking up using IP with "hidden" bits? That way users would be in a better position deciding to enable `edns-subnet` or not.

@snebel29  I've made some changes, please have a look.

chrisohaver:
I'm still leery of using a masked IP to lookup a value in an unmasked IP based  dataset.
I expect it's possible that the masked IP could result in no geo location data returned.  When this happens, would it make sense to fall back to using the source IP?

@julian7, have you had an opportunity to test how masked IP lookups via the geoip db actually behave?

julian7:
> I'm still leery of using a masked IP to lookup a value in an unmasked IP based dataset. I expect it's possible that the masked IP could result in no geo location data returned. When this happens, would it make sense to fall back to using the source IP?
> 
> @julian7, have you had an opportunity to test how masked IP lookups via the geoip db actually behave?

I've ran a query on my public IP address, using a not too recent copy of GeoLite2 Cou:

```
❯ for i in A.B.C.D A.B.C.0 A.B.0.0 A.0.0.0; mmdblookup --file GeoLite2-City.mmdb --ip $i country iso_code; end | grep -Ev '^$'
  "HU" <utf8_string>
  "HU" <utf8_string>
  "HU" <utf8_string>
  "HR" <utf8_string>
```

where A, B, C and D are the octets of my IP address, and I indeed live in Hungary. I've tested it on city names too, and even the most recent GeoLite2-City couldn't tell my city with all four octets, however, the /24 and /16 queries could still tell the county name accurately.

Fun fact, the location in /8 form is in a range I'd rather go to, Zadar is less than 1:30h away from the place :)

chrisohaver:
> I've ran a query on my public IP address ...

Thanks for testing it!

I guess that the way the data is stored is by subnet (or ip-ranges) for regions (e.g. counties, cities, etc), and the algorithm checks those subnets/ranges against the  looked-up IP.  The problem with sending a masked IP is that the algorithm can't tell the difference between unmasked zeros, and masked zeros, and could incorrectly provide a precise location to a masked IP (treating masked zeros as real zeros).  I'm kind of surprised the lookup algorithm doesn't accept a mask argument to help prevent that, but perhaps that level of correctness is not deemed important given the already inherent inaccuracy of geo-ip location data.

snebel29:
Yesterday I took a *quick* look to geoIP PDNS module source code, which is
using the same backend and I had the impression they were also doing
lookups with just the IP address too, whether full or masked.

Even if you could lookup passing the mask, there is no guarantees the
“subnet” would match perfectly with the ones in the database, there could
be an overlap, intermediate resolver could have masked differently that
classified in the DB, etc.

I agree this is not the level of accuracy of what we would instinctively
expect and/or like it to be, I guess geoIP databases are inherently
inaccurate, more if you add edns-subnet into the mix.

chrisohaver:
> Even if you could lookup passing the mask, there is no guarantees the
“subnet” would match perfectly with the ones in the database, there could
be an overlap, intermediate resolver could have masked differently that
classified in the DB, etc.

It wouldn't be looking for a perfect match, rather looking for subnets/ranges that enclose the queried subnet. If it can't find one, then either ...
* the queried subnet contains IPs that have no location data.
* and or the queried subnet contains IPs with different locations.

Subnet range comparisons can be expensive, so perhaps while more correct, it's judged not worth the trouble especially given the level of quality of the base data.  Anyways, this is digression.  The geoip lookup algorithms are there way they are.

Jumping back to an earlier question: If the ECS IP query produces no location data, is no location data better than location data based on the source IP?  IOW Should this plugin fallback on using source IP when the ECS IP query fails?  Or is the presence of an ECS somewhat indicating that the source IP is distant enough from the original client's IP that a nil location is "less wrong" than reporting the source IP location.

snebel29:
> Jumping back to an earlier question: If the ECS IP query produces no location data, is no location data better than location data based on the source IP? IOW Should this plugin fallback on using source IP when the ECS IP query fails? Or is the presence of an ECS somewhat indicating that the source IP is distant enough from the original client's IP that a nil location is "less wrong" than reporting the source IP location.

It is a good question I have no strong opinion for, at first glance It does feel like different use cases may want different things, some may prefer to have "some" data while others would rather have no data.

Right now, I do feel we are probably better off by keeping it simple rather than trying to be smart, and adding complexity through the addition of options when a specific use case appears.

WDYT?

chrisohaver:
> Right now, I do feel we are probably better off by keeping it simple rather than trying to be smart, and adding complexity.

I wasn't proposing adding an option.  It could be implemented either way, one is no more complex than the other.  If you don't have an opinion, we can leave it as is - i.e. no data is better than source ip based data when ECS is present.  I think I lean in that direction as being the better behavior. IMO, ECS present in a request generally indicates that the source IP should be ignored in this context.

PR #5191, opened 2022-02-15
plugin/cache: Correct responses to Authenticated Data requests

### LINKED ISSUE #5189: CoreDNS cache violates rfc6840
Some users of https://github.com/Mailu/Mailu reported that DNSSec is not working properly with CoreDNS. We figured out that it is caused by a change to CoreDNS introduced with #4736 that violates rfc6840.

For Details, please see comment in https://github.com/coredns/coredns/pull/4736#issuecomment-1039996401 and discussion in https://github.com/Mailu/Mailu/issues/2239#issuecomment-1039965168

chrisohaver:

Cache interferes with AD bit as part of un-DNSSEC-ifying responses to non-DNSSEC requests. As part of that process, we strip DNSSEC records and unset the AD bit. However, Per https://datatracker.ietf.org/doc/html/rfc6840#section-5.7 and https://datatracker.ietf.org/doc/html/rfc6840#section-5.8,
A requester may specify the AD bit without setting the DO bit, to indicate that it cares about the AD it in the response, but does not care to receive DNSSEC records.  **Therefore, we should retain the AD bit of the cached response whenever the requester sets the AD bit.**

~One thing I'm unsure of here is how the addition of the DO bit to the upstream query will affect the AD bit in the response.  That is, will the AD bit in a response differ between a query with AD and DO set vs AD set and DO unset?  If so, then this fix would not work and we may need to reassess #4085, since the AD bit in the response will always be as if the DO bit was set in the original request.  This depends on what the legacy usage of the AD bit is that rfc6840 is aiming to accommodate.~ 
I'm going to assume that the only reason a resolver will validly set the AD bit and not the DO bit is for the reason mentioned in rfc6840 - that a resolver wants to know if the data has been authenticated by a trusted source, and does not care to verify itself.

closes #5189

### DIFF
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -107,6 +107,7 @@ type ResponseWriter struct {
 	server string // Server handling the request.
 
 	do         bool // When true the original request had the DO bit set.
+	ad         bool // When true the original request had the AD bit set.
 	prefetch   bool // When true write nothing back to the client.
 	remoteAddr net.Addr
 }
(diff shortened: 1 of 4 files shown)


nextgens:
> One thing I'm unsure of here is how the addition of the DO bit to the upstream query will affect the AD bit in the response. That is, will the AD bit in a response differ between a query with AD and DO set vs AD set and DO unset? 

Short of the obvious (where an upstream resolver is running coredns without this PR) I don't see a case where it would.

Issue #4939, opened 2021-10-25
Update protobuf Module URL


utkonos:
**What would you like to be added**:
Update protobuf module URL.

**Why is this needed**:
The one being used is deprecated.

```
go: module github.com/golang/protobuf is deprecated: Use the "google.golang.org/protobuf" module instead.
```

Bosegeek:
Does any know how can we get this resolved ?

**build constraints exclude all Go files in /Users/xyz/go/pkg/mod/google.golang.org/protobuf@v1.27.1**

Updating the go.mod and updating the references does not work. I did try to install the package globally but still getting above error.

miekg:
how/when do you see this error?

yongtang:
@utkonos the issue has been fixed. Now coredns does not have reference to deprecated protobuf module any more. The dependency libraries may still pointing to the old protobuf module though that is outside of coredns control.

PR #5193, opened 2022-02-16
Update to use the latest protobuf package to build pb

yongtang:
The pb package was generated some time ago with old version
of https://github.com/golang/protobuf which was deprecated
and in favor of google.golang.org/protobuf (see
deprecation notice in https://pkg.go.dev/github.com/golang/protobuf)

This PR updates the generation of pb package with
v1.27.1 of google.golang.org/protobuf.

This PR will partially resolve the issue raised in #4939 (additional update might be needed).


chrisohaver (commented):
Thanks for tackling this Yong. 

> This PR will partially resolve the issue raised in https://github.com/coredns/coredns/issues/4939 (additional update might be needed).

What else do you think might be needed?

yongtang:
There are indirect dependencies in go.mod, which can only be fixed after upgrading dependencies that also use old protobuf package. We have to see if they can be fixed by dependabot + go mod tidy (when invoked by bot).

chrisohaver (approved):

PR #5213, opened 2022-02-28
Fix pinning of github actions

yongtang:
It is recommended to pin github actions with hash so that an action is not posing an unknown security risk (as the actions itself is not written by us).

Note while pinning to commit ID may looks to be tedious, they will be automatically updated by dependabot. So there is no additional burden.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -20,15 +20,15 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v1
+        uses: github/codeql-action/init@75f07e7ab2ee63cba88752d8c696324e4df67466
         with:
           languages: ${{ matrix.language }}
 
       - name: Autobuild
-        uses: github/codeql-action/autobuild@v1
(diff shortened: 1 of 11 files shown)


chrisohaver (commented):
> they will be automatically updated by dependabot

Will they be updated to use the latest tag or commit hash? i.e. will dependabot undo the pinning?

chrisohaver:
> It is recommended to pin github actions with hash so that an action is not posing an unknown security risk (as the actions itself is not written by us).

I presume the risk here is that a bad actor could delete and recreate a tag such that it points to a different commit that contains malicious code.

yongtang:
@chrisohaver  The pinning will not be undone, see example: https://github.com/yongtang/coredns/pull/8

chrisohaver (approved):
LGTM.  I did not verify any of the commit hash values.

yongtang:
> I presume the risk here is that a bad actor could delete and recreate a tag such that it points to a different commit that contains malicious code.

@chrisohaver the integrity of the actions can be changed if only version is specified The github actions we use are from open source community. While we trust the community, it can be possible that some outside hackers hacked into some repos and modify the actions by changing the content referenced by versions. Hash at least guarantees the integrity of the actions used.

chrisohaver:
> > I presume the risk here is that a bad actor could delete and recreate a tag such that it points to a different commit that contains malicious code.
> 
> @chrisohaver the integrity of the actions can be changed if only version is specified The github actions we use are from open source community. While we trust the community, it can be possible that some outside hackers hacked into some repos and modify the actions by changing the content referenced by versions. Hash at least guarantees the integrity of the actions used.

Sounds like a long winded "yes".

PR #5214, opened 2022-02-28
[plugin/rewrite] Refactor to satisfy security scan

yongtang:
Re-arrange the logic so that it will not hit the false positive of DAST scan result from LTGM.com:
https://lgtm.com/projects/g/coredns/coredns/?mode=list

The re-arrange also avoid an extra assign (not a big diff, though).

### DIFF
--- a/plugin/rewrite/name.go
+++ b/plugin/rewrite/name.go
@@ -298,18 +298,19 @@ func newNameRule(nextAction string, args ...string) (Rule, error) {
 		return nil, fmt.Errorf("too few arguments for a name rule")
 	}
 	if len(args) == 2 {
-		matchType = "exact"
+		matchType = ExactMatch
 		rewriteQuestionFrom = plugin.Name(args[0]).Normalize()
 		rewriteQuestionTo = plugin.Name(args[1]).Normalize()
 	}
 	if len(args) >= 3 {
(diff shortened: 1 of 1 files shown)


greenpau (approved):

greenpau:
@yongtang , adding a test would help.

chrisohaver:
Please describe what plugin the fix is in in PR description/title or commit message.

chrisohaver:
> @yongtang , adding a test would help.

I don't think a test can be added, because this condition will never happen.

As a result of the logic prior to the check, args cannot be < 3 if matchType = RegexMatch.
If args < 2, we return early.
If args == 2, we set match type to exact.

chrisohaver (changes requested):
I don't think the range check is currently necessary due to the logic above it.

yongtang:
@greenpau @chrisohaver  It looks like this is a false positive of DAST scan result from LTGM.com:
https://lgtm.com/projects/g/coredns/coredns/?mode=list

While the index bound will not be hit, I did update the PR and re-arrange the logic so that it will not hit the false positive (this is the only positive remaining). The re-arrange also avoid an extra assign (not a big diff, though).

chrisohaver (approved):
LGTM - I'm not offended by nested ifs and else statements.

PR #5218, opened 2022-03-02
all: fix some typos

cuishuang:


yongtang:
@cuishuang  I think transferer is the correct term (alternative to  transferrer). We don't want to change it. If you are using some tools (e.g., misspell, lint, etc) and a false positive is invoked, you can ignore or whitelist.

cuishuang:
> @cuishuang I think transferer is the correct term (alternative to transferrer). We don't want to change it. If you are using some tools (e.g., misspell, lint, etc) and a false positive is invoked, you can ignore or whitelist.

Thanks for the detailed answer ^_^

PR #5220, opened 2022-03-03
Use tickers instead of time.After to avoid memory leak

chrisohaver:

`time.After` is not garbage collected after it fires  (See https://pkg.go.dev/time#After) 

This converts most uses of time.After to use tickers instead.

I don't think the remaining usages of time.After pose a potential leak problem, but can convert those too if we think it's better to just avoid using time.After all together .

Addresses issue TOB-CDNS-10 in https://github.com/trailofbits/publications/blob/master/reviews/CoreDNS.pdf

### DIFF
--- a/plugin/dnstap/io.go
+++ b/plugin/dnstap/io.go
@@ -92,8 +92,10 @@ func (d *dio) write(payload *tap.Dnstap) error {
 }
 
 func (d *dio) serve() {
-	timeout := time.After(d.flushTimeout)
+	timeout := time.NewTimer(d.flushTimeout)
+	defer timeout.Stop()
 	for {
+		timeout.Reset(d.flushTimeout)
 		select {
 		case <-d.quit:
 			if d.enc == nil {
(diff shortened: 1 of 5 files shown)


yongtang (approved):

plugin/dnstap/io.go
```diff
@@ -92,8 +92,10 @@ func (d *dio) write(payload *tap.Dnstap) error {
 }
 
 func (d *dio) serve() {
-	timeout := time.After(d.flushTimeout)
+	timeout := time.NewTimer(d.flushTimeout)
+	defer timeout.Stop()
 	for {
+		timeout.Reset(d.flushTimeout)
```
lobshunter, line 98: Should this `timeout.Reset` be placed under `case <-timeout.C` branch, so the code is logically identical to `time.After` version?

chrisohaver: I don’t think it makes a difference, since the only other case in the select will exit. Iow, the only time we get back to the top of the for loop is when timeout fires.

lobshunter: There is another branch `payload := <-d.queue`, reset timeout on top results in longer flush interval.

It doesn't make too much difference though, messages are flushed out once buffer is full.

PR #5221, opened 2022-03-04
Enabling cifuzz

nathannaveen:

This enables cifuzz. https://google.github.io/oss-fuzz/getting-started/continuous-integration/

None

None

None

cc @yongtang

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,27 @@
+name: CIFuzz
+on:
+  pull_request:
+    branches:
+      - master
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,27 @@
+name: CIFuzz
+on:
+  push:
+    branches:
+      - master
```
yongtang, line 5: @nathannaveen I think it might make sense to trigger cifuzz for pull_request instead (so commit can be fixed before the merge)?
```
  pull_request:
    branches:
      - master
```

nathannaveen: I have done that, thanks!

yongtang (approved):

PR #5224, opened 2022-03-04
Add limit to ioutil.ReadAll for request body

yongtang:
This PR adds limit to ioutil.ReadAll for DoH request body
so that it will not be subject to large requests.

Addresses issue TOB-CDNS-14 in https://github.com/trailofbits/publications/blob/master/reviews/CoreDNS.pdf


chrisohaver (approved):
LGTM - The maximum size of a DNS message supported by DoH protocol is 65535 bytes.

PR #5225, opened 2022-03-04
Update to avoid pseudo-random number

yongtang:
This PR update the usage of rand so that non-global seed is used.

Addresses issue TOB-CDNS-7 in https://github.com/trailofbits/publications/blob/master/reviews/CoreDNS.pdf

### DIFF
--- a/plugin/forward/policy.go
+++ b/plugin/forward/policy.go
@@ -3,6 +3,7 @@ package forward
 import (
 	"math/rand"
 	"sync/atomic"
+	"time"
 )
 
 // Policy defines a policy we use for selecting upstreams.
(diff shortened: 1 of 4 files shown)


chrisohaver (approved):
lgtm thanks!

chrisohaver (changes requested):
The failure in tests seems to be related?

edit: Ah, yes.  I initially thought only the Coverage test was failing, but I see `/test` is also failing in the same place.

yongtang:
@chrisohaver  I will take a look.

chrisohaver (approved):
LGTM!

PR #5226, opened 2022-03-04
[plugin/reload]: Change hash from md5 to sha512

yongtang:

This PR changes the reload plugin's hash from md5 to sha512,
for the purpose of avoid using md5. MD5 is a weak hash algorithm
and for security reasons we will avoid using it.

Addresses issue TOB-CDNS-6 in https://github.com/trailofbits/publications/blob/master/reviews/CoreDNS.pdf

reload plugin.

This is a potential backward incompatible change, so likely will have to wait until 1.10 release.

### DIFF
--- a/plugin/reload/README.md
+++ b/plugin/reload/README.md
@@ -10,7 +10,7 @@ This plugin allows automatic reload of a changed _Corefile_.
 To enable automatic reloading of _zone file_ changes, use the `auto` plugin.
 
 This plugin periodically checks if the Corefile has changed by reading
-it and calculating its MD5 checksum. If the file has changed, it reloads
+it and calculating its SHA512 checksum. If the file has changed, it reloads
 CoreDNS with the new Corefile. This eliminates the need to send a SIGHUP
 or SIGUSR1 after changing the Corefile.
 
@@ -101,7 +101,7 @@ CoreDNS v1.7.0 and later does parse the Corefile and supports detecting changes 
 * `coredns_reload_failed_total{}` - counts the number of failed reload attempts.
 * `coredns_reload_version_info{hash, value}` - record the hash value during reload.
 
-Currently the type of `hash` is "md5", the `value` is the returned hash value.
+Currently the type of `hash` is "sha512", the `value` is the returned hash value.
 
 ## See Also
 
--- a/plugin/reload/reload.go
+++ b/plugin/reload/reload.go
@@ -3,7 +3,7 @@ package reload
 
 import (
 	"bytes"
-	"crypto/md5"
+	"crypto/sha512"
 	"encoding/hex"
 	"encoding/json"
 	"sync"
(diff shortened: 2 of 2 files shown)


johnbelamaric:
1) I don't have any real problem with this.
2) It seems unnecessary as this hash is simply for detecting changes not for any security purpose.
3) I don't see how it's backwards incompatible. All it's used for is detecting changes and reloading if there is one. Nothing will break.
4) Will this increase CPU use?

yongtang:
@johnbelamaric The subtle difference is when monitoring is enabled, then the output:
```
coredns_reload_version_info{hash, value}
```
will be changed. Though to an extent unless user tries to pass the value of the monitoring output then yes there is no backward incompatible changes.

yongtang:
@johnbelamaric security wise the md5 has been known that a collision can be crafted (unlikely sha where collision is almost impossible). That means coredns may continues to use the outdated server configuration while operator may assume an updated version is in place.

johnbelamaric:
I see, we export the hash as part of the metrics. Hmm. I don't think we make guarantees with respect to the metrics, but I am OK with waiting for 1.10 if you think it's important.

Even if you can craft a collision with MD5, I don't see how that's particularly relevant in this case. If someone has access to the config in order to modify it to make a collision, then, well, they have access to the config and can just change it to do whatever. 

In any case, assuming there is no dramatic CPU increase due to this, I am OK with it. Note that the hash is recalculated every x seconds. But I can't imagine that's really an issue (though some folks do use this in very resource constrained environments with dodgy...well, slow...CPUs).

chrisohaver:
> @johnbelamaric security wise the md5 has been known that a collision can be crafted (unlikely sha where collision is almost impossible). That means coredns may continues to use the outdated server configuration while operator may assume an updated version is in place.

There's the chance of this accidentally happening (probably zero) vs intentionally being crafted.

Realistically speaking though, a crafted colliding MD5 is virtually certain to be an invalid Corefile. I'd guess that it's probably the case that no two different _valid_ Corefiles actually share the same MD5.

I don't see a real danger here.  Not sure if its worth doing this to allay concerns of users that might be wary of any usage of MD5.

yongtang:
> Realistically speaking though, a crafted colliding MD5 is virtually certain to be an invalid Corefile.

I am less certain about that. Normally the way hackers do is to first craft a Corefile file that is valid, then pad with non-essential strings until a collision is hit. Unless Corefile does not allow non-essential strings (can be extremely long) the chance of a collision with valid Corefile is decent.

chrisohaver:
> Even if you can craft a collision with MD5, I don't see how that's particularly relevant in this case. If someone has access to the config in order to modify it to make a collision, then, well, they have access to the config and can just change it to do whatever.

I agree. Although I suppose someone could use this to deploy a malicious "time bomb" config, that would not become active until the next time CoreDNS restarts.  But like you say,  if the vulnerability requires access to replace the Corefile, then that is in itself a worse vulnerability. The same time-bomb effect can be achieved in a simpler way without md5 collision (e.g. in two updates to Corefile, first removing reload).

But, if there are no drawbacks it might be worth doing this to allay concerns, even if the vulnerability is only perceived.

chrisohaver:
> Unless Corefile does not allow non-essential strings (can be extremely long) the chance of a collision with valid Corefile is decent.

FWIW, it would not have to be non-essential strings like comments (if their content affects the md5). I think all that is needed is 128 bits worth of arbitrary "stuff" in the config like a few long gibberish zones.  So definitely doable.

chrisohaver (approved):
I think it's worth doing this to allay concerns, even if the vulnerability is only perceived.  I think it would be common for folks, including security professionals, to reflexively flag any use of MD5 as a problem.

I suspect the performance impact is minimal.

PR #5227, opened 2022-03-05
[plugin/azure] Deprecate the usage of plaintext secret in azure plugin

yongtang:

This PR deprecates the usage of plaintext secret in azure plugin.
Instead, this PR utilizes the environment-based authentication
that is recommended by azure:
https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authorization#use-environment-based-authentication

Note in order to maintain partial backward-compatibility, when `tenant`,
`client`, `secret`, `subscription` are specified, the values are saved
to environmental variables of `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`,
`AZURE_CLIENT_SECRET`, `AZURE_SUBSCRIPTION_ID` respectively.

Note also only `secret` has been deprecated now, all other options can still
be used (will replace the environment variable when appropriate).

The purpose of this PR is to remove the plaintext secret in Corefile for security
reasons.

Addresses issue TOB-CDNS-12 in https://github.com/trailofbits/publications/blob/master/reviews/CoreDNS.pdf

plugin/azure

plugin/azure

### DIFF
--- a/plugin/azure/README.md
+++ b/plugin/azure/README.md
@@ -14,11 +14,11 @@ record types. NS record type is not supported by azure private DNS.
 
 ~~~ txt
 azure RESOURCE_GROUP:ZONE... {
-    tenant TENANT_ID
-    client CLIENT_ID
-    secret CLIENT_SECRET
-    subscription SUBSCRIPTION_ID
-    environment ENVIRONMENT
+    tenant AZURE_TENANT_ID
+    client AZURE_CLIENT_ID
+    secret AZURE_CLIENT_SECRET
+    subscription AZURE_SUBSCRIPTION_ID
+    environment AZURE_ENVIRONMENT
     fallthrough [ZONES...]
     access private
 }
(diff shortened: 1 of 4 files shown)


plugin/azure/README.md
```diff
@@ -27,18 +27,31 @@ azure RESOURCE_GROUP:ZONE... {
...
 
 *   `fallthrough` If zone matches and no record can be generated, pass request to the next plugin.
     If **ZONES** is omitted, then fallthrough happens for all zones for which the plugin is
     authoritative.
 
 *   `access`  specifies if the zone is `public` or `private`. Default is `public`.
 
+## Authentication
+
+Azure plugin uses [environment-based authentication](https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authorization#use-environment-based-authentication),
```
chrisohaver, line 49:
```suggestion
The _azure_ plugin uses [environment-based authentication](https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authorization#use-environment-based-authentication),
```

plugin/azure/README.md
```diff
@@ -27,18 +27,31 @@ azure RESOURCE_GROUP:ZONE... {
...
+where there is a list of accepted environmental variables. Note in additional to passing those environmental variables,
+it is also possible to use `tenant`, `client`, `subscription`, and `environment` to override corresponding environmental variables.
```
chrisohaver, line 51:
```suggestion
where there is a list of accepted environmental variables. You may overwrite some environment variables using the `tenant`, `client`, `subscription`, and `environment` options.
```

plugin/azure/README.md
```diff
@@ -27,18 +27,31 @@ azure RESOURCE_GROUP:ZONE... {
...
+Note `secret` field in Corefile has been deprecated and may be removed in future releases. Users should pass secret through
+environmental variable **AZURE_CLIENT_SECRET** instead.
```
chrisohaver, line 53:
```suggestion

```

chrisohaver (changes requested):
I think we should make it very clear that the options do not just override existing environment variables for the connection, they actually _overwrite_ the respective environment variables.

chrisohaver:
> I think we should make it very clear that the options do not just override existing environment variables for the connection, they actually _overwrite_ the respective environment variables.

We may need to disallow multiple instances of _azure_ in the same server block (if it's currently allowed).
And also document any possible dangers when using multiple instances of _azure_ (defined in separate server blocks), e.g. possible race conditions on setting env - or one instance clobbering existing vars before another that expects those existing vars reads them.

If these those situations can happen, it could be backward incompatible with existing deployments.

PR #5228, opened 2022-03-05
[plugin/route53] Deprecate plaintext secret in Corefile for route53 plugin

yongtang:

This PR deprecates plaintext secret in Corefile for route53 plugin (`aws_access_key`).
Since using environmental variables of `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`
have already been available, no other changes other than deprecation is needed.

This will avoid saving plaintext secret in Corefile which could be
of security concern.

Addresses issue TOB-CDNS-12 in https://github.com/trailofbits/publications/blob/master/reviews/CoreDNS.pdf

plugin/route53

plugin/route53


dilyevsky (approved):
🔐

Issue #5223, opened 2022-03-04
seems that log plugin format not correct


hingbong:
```
log . {combined} {/forward/upstream} {
        class denial error
    }
```
i set this in Corefile, but i only get `[INFO] 8.8.4.4:853`.
another is that the `log_dir` is not working, always logs to stdout

chrisohaver:
The log format needs to be one ~token~ argument. e.g. If there are spaces in it, it needs to be wrapped in quotes...

```
log . "{combined} {/forward/upstream}" {
        class denial error
    }
```

Otherwise I think it only uses the last ~token~ argument on the line.

hingbong:
> The log format needs to be one ~token~ argument. e.g. If there are spaces in it, it needs to be wrapped in quotes...
> 
> ```
> log . "{combined} {/forward/upstream}" {
>         class denial error
>     }
> ```
> 
> Otherwise I think it only uses the last ~token~ argument on the line.

now i get `[INFO] {combined} 8.8.4.4:853`

chrisohaver:
Looking at the code, looks like `{combined}` and `{common}` cannot be combined with any other placeholders.

This is not documented.  The documentation suggests that these act like other placeholders, but they do not (they are kind of "fake" placeholders).

chrisohaver:
The following should be the equivalent of what you want ...

```
log . "{remote}:{port} - {>id} \"{type} {class} {name} {proto} {size} {>do} {>bufsize}\" {rcode} {>rflags} {rsize} {duration}  \"{>opcode}\" {/forward/upstream}"
```

chrisohaver:
IMO - _log_ should be fixed to treat `{common}` and `{combined}` as normal placeholders that can be used in conjunction with other placeholders.

hingbong:
only keep two quotation marks, `"{remote}:{port} - {>id} {type} {class} {name} {proto} {size} {>do} {>bufsize} {rcode} {>rflags} {rsize} {duration} {>opcode} {/forward/upstream}"` and works

chrisohaver:
> only keep two quotation marks ...

I neglected to escape two of the double-quotes... fixed in post above.

hingbong:
and the log_dir is not working, always logs to stdout
![image](https://user-images.githubusercontent.com/26398240/156862759-fe2526a4-920d-42d6-95db-9ccaf3bcafba.png)
do i need to open a new issue

yongtang:
[...]

PR #5229, opened 2022-03-05
Fix the unwanted command line options

yongtang:
This PR is related to #5223 where coredns incorrectly print out
command line options (such as log_dir) where those options was
never processed.
The reason was that coredns uses global "flag" for process
flags, yet some dependency libraries will also use the global
"flag" and injects the flags implicitly.
This PR uses flag.NewFlatSet to explicitly use a dedicated
FlagSet to process flags.
Before:
```
./coredns --help
Usage of ./coredns:
  -alsologtostderr
    	log to standard error as well as files
  -conf string
    	Corefile to load (default "Corefile")
  -dns.port string
    	Default port (default "53")
  -log_backtrace_at value
    	when logging hits line file:N, emit a stack trace
  -log_dir string
    	If non-empty, write log files in this directory
  -logtostderr
    	log to standard error instead of files
  -p string
    	Default port (default "53")
  -pidfile string
    	Path to write pid file
  -plugins
    	List installed plugins
  -quiet
    	Quiet mode (no initialization output)
  -stderrthreshold value
    	logs at or above this threshold go to stderr
  -v value
    	log level for V logs
  -version
    	Show version
  -vmodule value
    	comma-separated list of pattern=N settings for file-filtered logging
```
After:
```
./coredns --help
Usage of ./coredns:
  -conf string
    	Corefile to load (default "Corefile")
  -dns.port string
    	Default port (default "53")
  -p string
    	Default port (default "53")
  -pidfile string
    	Path to write pid file
  -plugins
    	List installed plugins
  -quiet
    	Quiet mode (no initialization output)
  -version
    	Show version

```

### DIFF
--- a/core/dnsserver/register.go
+++ b/core/dnsserver/register.go
@@ -1,7 +1,6 @@
 package dnsserver
 
 import (
-	"flag"
 	"fmt"
 	"net"
 	"time"
@@ -9,6 +8,7 @@ import (
 	"github.com/coredns/caddy"
 	"github.com/coredns/caddy/caddyfile"
 	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/pkg/flag"
 	"github.com/coredns/coredns/plugin/pkg/parse"
 	"github.com/coredns/coredns/plugin/pkg/transport"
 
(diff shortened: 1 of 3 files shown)


chrisohaver (commented):
Will this be backward incompatible?  Whats the behavior if someone specifies a deprecated flag?  Is it just ignored or will it cause CoreDNS not to start?

PR #5242, opened 2022-03-07
Update James's github handle

yongtang:
Since James's github handle has been updated from @fastest963 to
@jameshartig (https://github.com/coredns/coredns/commits?author=jameshartig),
this PR updates the CODEOWNERS file.


jameshartig:
@yongtang Thanks!!

PR #5243, opened 2022-03-08
plugin/cache: Avoid copy of large value in `range`

yongtang:
This PR made a small fix to avoid copy of large value (c.shards (2048 bytes) )
in range by using pointer to slice.

### DIFF
--- a/plugin/pkg/cache/cache.go
+++ b/plugin/pkg/cache/cache.go
@@ -66,7 +66,7 @@ func (c *Cache) Remove(key uint64) {
 // Len returns the number of elements in the cache.
 func (c *Cache) Len() int {
 	l := 0
-	for _, s := range c.shards {
+	for _, s := range &c.shards {
 		l += s.Len()
 	}
 	return l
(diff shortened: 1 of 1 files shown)


chrisohaver (commented):
Can you elaborate a bit on the rationale here? Was this found in a code analysis tool or something?

plugin/pkg/cache/cache.go
```diff
@@ -66,15 +66,15 @@ func (c *Cache) Remove(key uint64) {
 // Len returns the number of elements in the cache.
 func (c *Cache) Len() int {
 	l := 0
-	for _, s := range c.shards {
+	for _, s := range &c.shards {
 		l += s.Len()
 	}
```
chrisohaver, line 71: Perhaps ...

```suggestion
	for i := range c.shards {
		l += s[i].Len()
	}
```

... but shards is a slice of _pointers_ - so this should not make a difference.

chrisohaver:
> Can you elaborate a bit on the rationale here? 

Ah. I found https://www.pixelstech.net/article/1583648868-Some-tricks-and-tips-for-using-for-range-in-GoLang, which describes this optimization in section 3.

chrisohaver (approved):
> This PR made a small fix to avoid copy of large value (c.shards (2048 bytes) )
in range by _using pointer to slice_.

It's actually a pointer to an _array_.  Ranging over a pointer to a slice is invalid in Go.

The following does a good job at explaining why this optimization works ...
 
https://garbagecollected.org/2017/02/22/go-range-loop-internals/

https://go.dev/play/p/4b181zkB1O

PR #5244, opened 2022-03-08
plugin/health : try to shutdown overloaded goroutine first to prevent warnings

### LINKED ISSUE #5249: Warnings by health plugin when CoreDNS is shutting down
**What happened**:
Hello, while running customized CoreDNS in our clusters, we noticed that sometimes during the shutdown of CoreDNS there are warnings from the `health` plugin. After some investigation, we think it is caused by the incorrect order of shutting down goroutines in `health` plugin => race condition in shutdown

```
[WARNING] plugin/health: Local health request to "http://:8080/health" failed: Get "http://:8080/health": dial tcp :8080: connect: connection refused
```

No such warning appears on CoreDNS shutdown, see linked PR https://github.com/coredns/coredns/pull/5244

This is not happening very often, but when CoreDNS receives a shutdown signal in a form of SIGTERM and starts shutting down, this warning log sometimes appears.

No

**Environment**:
K8s cluster

- the version of CoreDNS: 1.9.0
- Corefile:
```
.:53 {
  health
}
```
- logs, if applicable:
```
.:53
CoreDNS-1.9.0
darwin/amd64, go1.17.6, 
[INFO] SIGTERM: Shutting down servers then terminating
[WARNING] plugin/health: Local health request to "http://:8080/health" failed: Get "http://:8080/health": dial tcp :8080: connect: connection refused
```
- Others:

Tantalor93:
There are two goroutines in the `health` plugin, one is running a health HTTP server and the second is pinging the health HTTP server = function [overloaded](https://github.com/coredns/coredns/blob/master/plugin/health/overloaded.go#L14). When CoreDNS is shutting down, the health HTTP server is shutdown first, and then the `overloaded` goroutine second, therefore there is a possibility for the second goroutine to call the already closed health HTTP server.

example of logs
```
.:53
CoreDNS-1.9.0
darwin/amd64, go1.17.6, 
[INFO] SIGTERM: Shutting down servers then terminating
[WARNING] plugin/health: Local health request to "http://:8080/health" failed: Get "http://:8080/health": dial tcp :8080: connect: connection refused
```

This PR is changing the `overloaded` function to support graceful shutdown, by accepting context which is cancelled by parent goroutine upon CoreDNS shutdown, this way we can gracefully cancel the `overloaded` function even in the middle of HTTP request

#5249

none

no

### DIFF
--- a/plugin/health/health.go
+++ b/plugin/health/health.go
@@ -2,6 +2,7 @@
 package health
 
 import (
+	"context"
 	"io"
 	"net"
 	"net/http"
@@ -22,14 +23,13 @@ type health struct {
 	nlSetup bool
 	mux     *http.ServeMux
 
-	stop chan bool
+	stop context.CancelFunc
 }
 
 func (h *health) OnStartup() error {
 	if h.Addr == "" {
 		h.Addr = ":8080"
 	}
-	h.stop = make(chan bool)
 	ln, err := reuseport.Listen("tcp", h.Addr)
 	if err != nil {
 		return err
@@ -45,8 +45,11 @@ func (h *health) OnStartup() error {
 		io.WriteString(w, http.StatusText(http.StatusOK))
 	})
 
+	ctx := context.Background()
+	ctx, h.stop = context.WithCancel(ctx)
+
 	go func() { http.Serve(h.ln, h.mux) }()
-	go func() { h.overloaded() }()
+	go func() { h.overloaded(ctx) }()
 
 	return nil
 }
@@ -61,9 +64,9 @@ func (h *health) OnFinalShutdown() error {
 		time.Sleep(h.lameduck)
 	}
 
-	h.ln.Close()
+	h.stop()
 
+	h.ln.Close()
 	h.nlSetup = false
-	close(h.stop)
 	return nil
 }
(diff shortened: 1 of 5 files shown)


yongtang:
> not mitigate! that would require more changes

Thanks @Tantalor93  for the PR. Can you briefly list the possible changes that might be needed, if that is possible?

Tantalor93:
@yongtang I would say that `overloaded` goroutine would have to be closed first in a more robust way = adjust `overloaded` function to accept cancellable context and distinguish whether the HTTP request was cancelled due to CoreDNS shutdown or timeout.

 If you think it is worth it, I can prepare changes, I was just not sure whether you are open to bigger changes in this plugin just because of warning log on shutdown 🙂

yongtang:
@Tantalor93 We definitely want to polish and address any potential issues! Will be interested to see the proposed change 👍

Tantalor93:
@yongtang I have prepared the solution I was talking before, I would appreciate your code review & feedback 🙏

yongtang:
Thanks @Tantalor93 for the PR! Overall looks good to me. Let's keep it for several days to see if there are inputs from other maintainers.

Tantalor93:
Thanks! @yongtang

Issue #5246, opened 2022-03-09
1.9.1 release tracking 


yongtang:
CoreDNS 1.9.0 built with golang 1.17.6. However, golang 1.17.6 consists of several security vulnerabilities (CVE-2022-23772, CVE-2022-23773, CVE-2022-23806).

It makes sense to release a new coredns 1.9.1 with latest golang version (1.17.8).

yongtang:
PR #5247 was merged. Will try to do a 1.9.1 release soon.

yongtang:
1.9.1 has been released, closing. 👍 🎉

PR #5247, opened 2022-03-09
Bump version to 1.9.1, and add release notes

yongtang:
This PR bumps coredns to 1.9.1 and add release notes.

See #5246 for release tracking.


yongtang:
If no objections, I will just merge as there is a security related release update (golang 1.17.6 had several vulnerabilities, and we used 1.17.6 for coredns 1.9.0 release). See #5246.

PR #5250, opened 2022-03-12
Unit test - config.go

neilnaveen:
This is my first PR to this project. Let me know if I am missing something. 
- Included test for config.go.
- Included additional .gitignore for Goland.
- 100% coverage for config.go 

None
None
None


.gitignore
```diff
@@ -4,3 +4,6 @@ coredns.exe
 build/
 release/
 vendor/
+/.idea/coredns.iml
```
yongtang: @neilnaveen Normally don't want to add IDE to gitignore as golang itself is not tied to IDE. Can you remove the change to this file?

neilnaveen: I have removed all idea folders.

core/dnsserver/config_test.go
```diff
@@ -0,0 +1,122 @@
...
+				blocIndex:    1,
+				blocKeyIndex: 1,
+			},
+			want: "1:1",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := keyForConfig(tt.args.blocIndex, tt.args.blocKeyIndex); got != tt.want {
```
dilyevsky, line 54: Sorry, this function doesn't appear to be doing anything complicated - imo the disadvantages of extra boilerplate test code outweigh the possible benefits of test coverage (literally covering one line here).

core/dnsserver/config_test.go
```diff
@@ -0,0 +1,122 @@
...
+				Port: "Root",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := caddy.NewTestController("dns", tt.name)
+			c.ServerBlockKeyIndex = tt.args.keyIndex
+			c.ServerBlockIndex = tt.args.BlockIndex
+			if got := GetConfig(c); got.Port == tt.want.Port {
```
dilyevsky, line 117: I think ^^ reasoning applies here as well.

core/dnsserver/config_test.go
```diff
@@ -0,0 +1,122 @@
...
+		keyIndex   int
+		BlockIndex int
+	}
+	tests := []struct {
+		name string
+		args args
+		want *Config
+	}{
+		{
+			name: "Test #1",
```
dilyevsky, line 72: These test cases don't seem to be testing useful corner cases and instead test boolean permutation of non-boolean variables (`int`).

pro tip: use description of a test as a test name (e.g `"return nil on invalid key"`). If you can't think of a good description probably not a useful test then.

core/dnsserver/config_test.go
neilnaveen: So do you want me to remove this test code? Let me know.

neilnaveen: I have a different opinion on this. The tests makes sure that the `fmt.Sprintf` is working as expected with a `:`. I agree we don't need many tests. But having 1 tests will ensure that things don't break. Thoughts?

core/dnsserver/config_test.go
neilnaveen: The idea was to set the 
 ```
 c.ServerBlockKeyIndex = tt.args.keyIndex
c.ServerBlockIndex = tt.args.BlockIndex
```
So that `c` can be passed to `GetConfig`. If there are other corner cases please let me know and I can fix them.

Thanks

core/dnsserver/config_test.go
neilnaveen: Sorry I don't get it. Can you please explain?

PR #5263, opened 2022-03-15
Unit test - request/writer.go

nathannaveen:

- Implemented unit tests for request/writer.go

### DIFF
--- a/request/writer_test.go
+++ b/request/writer_test.go
@@ -1,0 +1,122 @@
+package request
+
+import (
+	"github.com/coredns/coredns/plugin/test"
+	"github.com/miekg/dns"
+	"reflect"
+	"testing"
+)
+
+func TestNewScrubWriter(t *testing.T) {
+	type args struct {
+		req *dns.Msg
+		w   dns.ResponseWriter
+	}
+	tests := []struct {
+		name string
+		args args
+		want *ScrubWriter
+	}{
+		{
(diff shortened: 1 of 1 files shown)


nathannaveen:
@yongtang 👀 please!

request/writer_test.go
```diff
@@ -0,0 +1,122 @@
+package request
+
+import (
+	"github.com/coredns/coredns/plugin/test"
+	"github.com/miekg/dns"
+	"reflect"
+	"testing"
+)
+
+func TestNewScrubWriter(t *testing.T) {
```
chrisohaver, line 10: This test is not needed.  `NewScrubWriter` is a one line function - too simple to warrant this.

request/writer_test.go
```diff
@@ -0,0 +1,122 @@
...
+	test.ResponseWriter
+	written []byte
+}
+
+func (r *inmemoryWriter) WriteMsg(m *dns.Msg) error {
+	r.written, _ = m.Pack()
+	return r.ResponseWriter.WriteMsg(m)
+}
+
+func TestScrubWriter_WriteMsg(t *testing.T) {
```
chrisohaver, line 67: This doesn't appear to be testing anything that `WriteMsg` does.

chrisohaver (changes requested):
I don't think these tests are helpful.

PR #5268, opened 2022-03-17
[plugin/route53] Eliminate the unnecessary passing of provider chain

### LINKED ISSUE #5153: Support IRSA in Route53 plugin
Apologize in advance if this is some user end config error that I fail to recognize.

I am trying to configure coredns to read records in some route53 private zone. My preference is to use IRSA so that I don't need to create static AWS access keys, but it doesn't work. My setup:

in coredns config:

```yaml
example.com:53 {
  route53 example.com:<my-private-zone-id>
}
```

And I have verified that the web identity token has been injected into the container successfully:

```yaml
  containers:
    env:
    - name: AWS_DEFAULT_REGION
      value: us-east-2
    - name: AWS_REGION
      value: us-east-2
    - name: AWS_ROLE_ARN
      value: arn:aws:iam::<arn-for-the-role>
    - name: AWS_WEB_IDENTITY_TOKEN_FILE
      value: /var/run/secrets/eks.amazonaws.com/serviceaccount/token
    image: 602401143452.dkr.ecr.us-east-2.amazonaws.com/eks/coredns:v1.8.4-eksbuild.1
...
    volumeMounts:
    - mountPath: /etc/coredns
      name: config-volume
      readOnly: true
    - mountPath: /tmp
      name: tmp
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: kube-api-access-kr77q
      readOnly: true
    - mountPath: /var/run/secrets/eks.amazonaws.com/serviceaccount
      name: aws-iam-token
      readOnly: true
...
  volumes:
  - name: aws-iam-token
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          audience: sts.amazonaws.com
          expirationSeconds: 86400
          path: token
```

However, in the log, I found that coredns is still trying to use the instance profile:

```
plugin/route53: /etc/coredns/Corefile:26 - Error during parsing: failed to create route53 plugin: AccessDenied: User: arn:aws:sts::<aws-account-id>:assumed-role/eks-worker-<eks-cluster-name>/i-<ec2-instance-id> is not authorized to perform: route53:ListHostedZonesByName because no identity-based policy allows the route53:ListHostedZonesByName action
        status code: 403, request id: xxxxxxxxxxxxxxxxxxxxxxxxx
```

Support for IRSA, or, if it is already supported, update the documentation to include instruction on how to use it.

**Why is this needed**:

Provision static AWS access keys, as suggested in the documentation, is a security risk. Using instance profile is doable, but provision a dedicated node group just for a a few coredns pods in our k8s clusters seems to be a little overkill and can create management overhead. 

If it is possible to support IRSA, we can grant Route53 read permission to coredns' service account and nothing else.

tommyk1m:
- Passed static credential is used as `opts` overriding all preceding config sources during the mergeConfigSrcs 
https://github.com/aws/aws-sdk-go/blob/92b5621bf8338d68f64c79cd094969d289428723/aws/session/session.go#L499

- The override prevents creating sessions using `envCfg` as we see in https://github.com/coredns/coredns/issues/5153

- Passed in aws_access_key pair and shared credential configs will be set as ENV variables and picked up as `envCfg`

https://github.com/coredns/coredns/issues/5153

plugin/route53

nope

### DIFF
--- a/plugin/route53/setup.go
+++ b/plugin/route53/setup.go
@@ -3,6 +3,7 @@ package route53
 import (
 	"context"
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -14,8 +15,6 @@ import (
 	clog "github.com/coredns/coredns/plugin/pkg/log"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/credentials"
-	"github.com/aws/aws-sdk-go/aws/defaults"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/route53"
 	"github.com/aws/aws-sdk-go/service/route53/route53iface"
@@ -26,8 +25,10 @@ var log = clog.NewWithPlugin("route53")
 func init() { plugin.Register("route53", setup) }
 
 // exposed for testing
-var f = func(credential *credentials.Credentials, endpoint *string) route53iface.Route53API {
-	return route53.New(session.Must(session.NewSession(&aws.Config{Credentials: credential, Endpoint: endpoint})))
+var f = func(endpoint *string) route53iface.Route53API {
+	cfg := aws.Config{}
+	cfg.Endpoint = endpoint
+	return route53.New(session.Must(session.NewSession(&cfg)))
 }
 
 func setup(c *caddy.Controller) error {
@@ -41,8 +42,6 @@ func setup(c *caddy.Controller) error {
 		// With that said, even though a user doesn't define any credentials in
 		// Corefile, we should still attempt to read the default credentials file,
 		// ~/.aws/credentials with the default profile.
-		sharedProvider := &credentials.SharedCredentialsProvider{}
-		var providers []credentials.Provider
 		var fall fall.F
 		var endpoint string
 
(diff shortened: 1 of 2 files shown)


plugin/route53/setup.go
```diff
@@ -91,12 +86,12 @@ func setup(c *caddy.Controller) error {
 				c.RemainingArgs() // eats args
 			case "credentials":
 				if c.NextArg() {
-					sharedProvider.Profile = c.Val()
+					os.Setenv("AWS_PROFILE", c.Val())
```
dilyevsky, line 89: ...actually this will potentially mess with other `route53` plugins using different credentials no? May need to use https://pkg.go.dev/github.com/aws/aws-sdk-go/aws/session#Options with `session.NewSessionWithOptions` call.

dilyevsky (changes requested):
Thanks! Don't really dig AWS's way of configuring session via env variables but oh well...

chrisohaver:
> Thanks! Don't really dig AWS's way of configuring session via env variables but oh well...

Yeah - I'm seeing it introduced in #5227 too.  IIUC, this is changing the environment, and therefore it should be documented that this is happening.

A side effect of this is collisions with other instances of the same plugin - e.g. connecting twice using different creds for two different zones. If the first instance initialized sets the option, it mutates the environment before the second instance initializes. Race conditions could arise if establishing connections is deferred execute on in server start, instead of during setup (depends on when the values are actually _read_ from the env). 

(edit: @dilyevsky, I just noticed you already mentioned this possibility above)

dilyevsky:
Yeah I believe just updating docs is probably not enough since it'll likely break some existing setups.

chrisohaver:
Superseded by #5370

PR #5269, opened 2022-03-17
Included additional tests for server.go

neilnaveen:
- Included additional tests for server.go

Signed-off-by: neil <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/core/dnsserver/server_test.go
+++ b/core/dnsserver/server_test.go
@@ -2,14 +2,32 @@ package dnsserver
 
 import (
 	"context"
+	"net"
+	"reflect"
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/pkg/log"
+	"github.com/coredns/coredns/plugin/pkg/trace"
 	"github.com/coredns/coredns/plugin/test"
-
 	"github.com/miekg/dns"
 )
+
+// testHandler is a net.Listener that implements the net.Listener interface.
+type testHandler struct {
+}
+
+func (h *testHandler) Accept() (net.Conn, error) {
+	return nil, nil
+}
+func (h *testHandler) Close() error {
+	return nil
+}
+func (h *testHandler) Addr() net.Addr {
+	return nil
+}
 
 type testPlugin struct{}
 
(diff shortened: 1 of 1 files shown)


neilnaveen:
@yongtang can you please review?

core/dnsserver/server_test.go
```diff
@@ -2,15 +2,33 @@ package dnsserver
...
+	"reflect"
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/pkg/log"
+	"github.com/coredns/coredns/plugin/pkg/trace"
 	"github.com/coredns/coredns/plugin/test"
-
```
chrisohaver, line 10: removing this line will cause import order/grouping tests to fail.

core/dnsserver/server_test.go
```diff
@@ -93,3 +111,149 @@ func BenchmarkCoreServeDNS(b *testing.B) {
...
+				classChaos:   tt.fields.classChaos,
+			}
+			if got := s.WrapListener(tt.args.ln); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("Server.WrapListener() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestServer_Address(t *testing.T) {
```
chrisohaver, line 193: `Server.Address` is a single line function that returns simply Addr field.
It's too simple to warrant a test of this bulk.

core/dnsserver/server_test.go
```diff
@@ -93,3 +111,149 @@ func BenchmarkCoreServeDNS(b *testing.B) {
...
+				dnsWg:        sync.WaitGroup{},
+				graceTimeout: time.Second,
+				trace:        nil,
+				debug:        false,
+				classChaos:   false,
+			},
+			want: "",
+		},
+		{
+			name: "nil case",
```
chrisohaver, line 226: "nil case" doesn't describe this case. ":53" isn't nil.

chrisohaver (changes requested):
Neither of the functions tested warrant such complex test cases.
They are both one line functions that simply return a single struct field.

Were these tests created by a unit test code generator?

PR #5271, opened 2022-03-18
update all +build statements

mqasimsarfraz:
Signed-off-by: MQasimSarfraz <syed.qasim.sarfraz@gmail.com>

As per [go 1.17](https://go.dev/doc/go1.17), "The go command now understands //go:build lines and prefers them over // +build lines". This PR includes changes after `gofmt` to add new pragmas. The changes are backward compatible and `+build` should be removed in future once needed.

### DIFF
--- a/directives_generate.go
+++ b/directives_generate.go
@@ -1,4 +1,4 @@
-//+build ignore
+//go:build ignore
 
 package main
 
--- a/owners_generate.go
+++ b/owners_generate.go
@@ -1,4 +1,4 @@
-//+build ignore
+//go:build ignore
 
 // generates plugin/chaos/zowners.go.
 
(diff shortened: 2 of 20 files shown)


chrisohaver:
I'm thinking we can remove the older style now, since we already require at least go 1.17.

mqasimsarfraz:
@chrisohaver I also think we should be fine removing old style. Especially since it will be removed with go 1.18. So if it makes sense I go ahead the clean them in the same PR?

chrisohaver:
> @chrisohaver I also think we should be fine removing old style. Especially since it will be removed with go 1.18. So if it makes sense I go ahead the clean them in the same PR?

Yes.

mqasimsarfraz:
@chrisohaver Can you please have a look again?

yongtang (approved):

PR #5281, opened 2022-03-22
plugin/forward: configurable domain support for healthcheck

### LINKED ISSUE #5121: plugin/forward: configurable healthcheck domain support

Make domain used by forward plugin's upstream health check configurable.

**Why is this needed**:

Currently forward plugin health check resolves `.` (see https://coredns.io/plugins/forward/#description). Configuring DNS firewalls to allow `.` to be resolvable, and not have that interpreted as allow all domains is not always possible.

hansedong:
Signed-off-by: hansedong <admin@yinxiaoluo.com>

Add support for configurable domains for the health_check configuration item in the forward plugin.

In some cases, CoreDNS is deployed in a company's internal network and cannot resolve `. NS` records. If health check is enabled, CoreDNS will use the `. NS` record as the request method for health check.

At this time, if the upstream DNS Server does not have the ability to resolve `. NS` records, or the response to resolve `. NS` records is very slow (for example, the DNS Server is not in the United States), but it is very fast to resolve other domestic domain names. In this case, using `. NS` records as a fixed detection method will cause the health check to fail all the time.

#5121 

https://coredns.io/plugins/forward/

No

### DIFF
--- a/plugin/forward/health_test.go
+++ b/plugin/forward/health_test.go
@@ -242,3 +242,42 @@ func TestHealthNoMaxFails(t *testing.T) {
 		t.Errorf("Expected number of health checks to be %d, got %d", 0, i1)
 	}
 }
+
+func TestHealthDomain(t *testing.T) {
+	hcReadTimeout = 10 * time.Millisecond
+	readTimeout = 10 * time.Millisecond
+	defaultTimeout = 10 * time.Millisecond
+	hcWriteTimeout = 10 * time.Millisecond
+	hcDomain := "example.org."
+	i := uint32(0)
+	q := uint32(0)
+	s := dnstest.NewServer(func(w dns.ResponseWriter, r *dns.Msg) {
+		if atomic.LoadUint32(&q) == 0 { //drop the first query to trigger health-checking
+			atomic.AddUint32(&q, 1)
+			return
+		}
+		if r.Question[0].Name == hcDomain && r.RecursionDesired == true {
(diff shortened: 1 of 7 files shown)


plugin/forward/health_test.go
```diff
@@ -242,3 +242,42 @@ func TestHealthNoMaxFails(t *testing.T) {
...
+
+	req := new(dns.Msg)
+	req.SetQuestion(".", dns.TypeNS)
+
+	f.ServeDNS(context.TODO(), &test.ResponseWriter{}, req)
+
+	time.Sleep(20 * time.Millisecond)
+	i1 := atomic.LoadUint32(&i)
+	if i1 != 1 {
+		t.Errorf("Expected number of health checks with RecursionDesired==false to be %d, got %d", 1, i1)
```
chrisohaver: Please adjust the message to fit the case being tested.

chrisohaver (changes requested):
Thanks. Please also update the README.

chrisohaver (approved):
lgtm thanks!

hansedong:
@chrisohaver Thanks, by the way, do I need to update the content related to the forward plugin in the [github.com/coredns/coredns.io](https://github.com/coredns/coredns.io) project?

chrisohaver:
> do I need to update the content related to the forward plugin in the [github.com/coredns/coredns.io](https://github.com/coredns/coredns.io) project?


No, we have a script that copies those from this repo when we cut new releases.

PR #5285, opened 2022-03-25
Pv firewall

kenshinx:
DNS firewall based coredns demo code


chrisohaver:
FYI @kenshinx, coredns maintains a policy based dns firewall external plugin : https://github.com/coredns/policy

PR #5291, opened 2022-03-29
plugin/backend: Fix direct CNAME loop check

chrisohaver:

Fixes the direct CNAME loop check done in a few places in `backend_lookup.go`.  The intent of the check is to detect a CNAME that directly targets itself (e.g. per comments `// x CNAME x is a direct loop, don't add those`).

However the check was using`Name(X).Matches(Y)` which despite its name returns true if X is a _subdomain_ of Y.  Therefore for example, the CNAME `an.example.com. CNAME example.com.` would be be detected as a direct loop, and dropped from responses.

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math"
 	"net"
+	"strings"
 
 	"github.com/coredns/coredns/plugin/etcd/msg"
 	"github.com/coredns/coredns/plugin/pkg/dnsutil"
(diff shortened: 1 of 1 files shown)


chrisohaver:
I think now this was originally done as a subdomain check specifically due to the "recursive" way that the _etcd_ plugin (skydns) works. Skydns's default mode is to return all records for the searched domain and subdomains.  Therefore a CNAME targeting a parent domain of itself would end up being a self loop in the _etcd_ plugin.

I'll revert the fix here, and add comments as to why the sub domain check is done.

PR #5292, opened 2022-03-29
Pin docker image by SHA

neilnaveen:
Signed-off-by: neil <42328488+neilnaveen@users.noreply.github.com>

- pinned docker image by SHA using crane digest https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_digest.md
- https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:stable-slim
+FROM debian:stable-slim@sha256:a04e8ec615db03154299695a65f7a85b8800a4e7878d241871419d3878e3662d
 
 RUN apt-get update && apt-get -uy upgrade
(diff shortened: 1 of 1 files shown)


neilnaveen:
@yongtang 👀

yongtang:
@neilnaveen we have to be careful about the change here. We are using generic debian image to build multi-arch docker images. Not sure ping to a specific version will cause any issues. Can you verify this change works for amd64 and arm64 at the same time?

yongtang:
@neilnaveen we have to be careful about the change here. We are using generic debian image to build multi-arch docker images. Not sure ping to a specific version will cause any issues. Can you verify this change works for amd64 and arm64 at the same time?

PR #5294, opened 2022-03-31
plugin/etcd: Comment reason for non-exact match in direct cname loop checks

chrisohaver:

**This only adds some code  comments** that clarify the counterintuitive way we check for direct CNAME loops. When checking for direct CNAME loops (where a name points too itself) we don't do an exact comparison of name and target. Instead we do a subdomain check, to see if the CNAME is a subdomain or equal to the CNAME's target. This is because in etcd/skydns `w.x CNAME x` is also a direct loop due to the "recursive" nature of how skydns processes lookups.

I found this confusing enough to warrant a comment. 

replaces #5291

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -30,6 +30,7 @@ func A(ctx context.Context, b ServiceBackend, zone string, state request.Request
 		case dns.TypeCNAME:
 			if Name(state.Name()).Matches(dns.Fqdn(serv.Host)) {
 				// x CNAME x is a direct loop, don't add those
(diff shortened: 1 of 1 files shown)


chrisohaver:
Note: This logic specifically caters to the _etcd_ plugin's skydns convention.  But being in the backend, it does also affect the _kubernetes_ plugin.  That is, it prevents a Kubernetes Externalname (which is represented as a CNAME) from targeting a superdomain of itself, which in normal DNS terms is valid to do.  However given the Kubernetes DNS schema, this doesn't limit any practical scenario that I can imagine.

PR #5311, opened 2022-04-07
Update ACL plugin's README to capture the edge case

### LINKED ISSUE #4697: ACLs seem to not take affect correctly (IPv6, kubernetes)
**What happened**:

ACLs have been defined to *block* certain traffic as follows:

```
  Corefile: |
    .:53 {
        acl k8s.place7.ungleich.ch {
             allow net ::/0
        }
        acl . {
              allow net 2a0a:e5c0:13::/48
              block
        }
        forward . /etc/resolv.conf {
           max_concurrent 1000
        }
```

This configuration **does** allow the use of CoreDNS as an Open resolver from the whole IPv6 Internet.
It should not, though. I've tested this from the world wild web on different endpoints which are certainly out of this range and I was able to resolv any domain name.

These ACLs should, in our opinion, do the following:

* allow any kind of request from 2a0a:e5c0:13::/48
* allow querying k8s.place7.ungleich.ch (the k8s domain) 
* block all other requests

Use above config 

**Environment**:

- the version of CoreDNS: k8s.gcr.io/coredns/coredns:v1.8.0
- Corefile: See above
- logs, if applicable:
- Others:

yongtang:
This PR adds a note in ACL plugin's README to capture the edge case
where source ip of the client may be different from the source ip
observed by coredns server (in situations such as Source NAT).

This PR closes #4697.

### DIFF
--- a/plugin/acl/README.md
+++ b/plugin/acl/README.md
@@ -8,8 +8,11 @@
 
 With `acl` enabled, users are able to block or filter suspicious DNS queries by configuring IP filter rule sets, i.e. allowing authorized queries to recurse or blocking unauthorized queries.
 
-Note the rule sets of `acl` are based on the source ip of the query observed by coredns server. If the source ip observed by coredns is different from
-the source ip of the sender client (e.g., in situations such as Source NAT), then rule sets based on client source ip may not work as expected.
+
+When evaluating the rule sets, _acl_ uses the source IP of the TCP/UDP headers of the DNS query received by CoreDNS.
+This source IP will be different than the IP of the client originating the request in cases where the source IP of the request is changed in transit.  For example:
+* if the request passes though an intermediate forwarding DNS server or recursive DNS server before reaching CoreDNS
+* if the request traverses a Source NAT before reaching CoreDNS
 
 This plugin can be used multiple times per Server Block.
 
--- a/plugin/acl/README.md
+++ b/plugin/acl/README.md
@@ -8,6 +8,9 @@
 
 With `acl` enabled, users are able to block or filter suspicious DNS queries by configuring IP filter rule sets, i.e. allowing authorized queries to recurse or blocking unauthorized queries.
 
+Note the rule sets of `acl` are based on the source ip of the query observed by coredns server. If the source ip observed by coredns is different from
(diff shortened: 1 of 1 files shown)


plugin/acl/README.md
```diff
@@ -8,6 +8,9 @@
 
 With `acl` enabled, users are able to block or filter suspicious DNS queries by configuring IP filter rule sets, i.e. allowing authorized queries to recurse or blocking unauthorized queries.
 
+Note the rule sets of `acl` are based on the source ip of the query observed by coredns server. If the source ip observed by coredns is different from
+the source ip of the sender client (e.g., in situations such as Source NAT), then rule sets based on client source ip may not work as expected.
+
```
chrisohaver:
```suggestion

When evaluating the rule sets, _acl_ uses the source IP of the TCP/UDP headers of the DNS query received by CoreDNS.
This source IP will be different than the IP of the client originating the request in cases where the source IP of the request is changed in transit.  For example:
* if the request passes though an intermediate forwarding DNS server or recursive DNS server before reaching CoreDNS
* if the request traverses a Source NAT before reaching CoreDNS

```

plugin/acl/README.md
```diff
@@ -8,6 +8,9 @@
 
 With `acl` enabled, users are able to block or filter suspicious DNS queries by configuring IP filter rule sets, i.e. allowing authorized queries to recurse or blocking unauthorized queries.
```
chrisohaver:
```suggestion
With `acl` enabled, users are able to block or filter suspicious DNS queries by configuring IP filter rule sets, i.e. allowing authorized queries or blocking unauthorized queries.
```

In addition to the source IP confusion, I think users in  #4697 may have misinterpreted "recurse" here, and thought this meant it would block forwarding to upstream recursive servers (but not to local authoritative zones).  What is actually meant here is that it either allows or blocks the query from passing to the rest of the plugin chain (but this is not "recursion" in a DNS sense).  Removing "to recurse" makes this clearer and more direct.

chrisohaver (changes requested):

chrisohaver (approved):

ihac:
Thanks!

PR #5328, opened 2022-04-22
plugin/template: fix rcode option documentation

### LINKED ISSUE #5327: template plugin documentation uses wrong default value for rcode
**What happened**:
I attempted to use the template plugin to cause CoreDNS to produce an empty response to requests for an AAAA record. CoreDNS complained that my configuration had an invalid value for rcode.

I expected AAAA requests to return successful, but empty responses.

template IN AAAA {
  match www.example.com.
  rcode SUCCESS
}

The problem turned out to be the documentation for the template plugin, which states that the default value for rcode is "SUCCESS", Looking at the source code, I discovered that the successful value is "NOERROR".

**Environment**:

- the version of CoreDNS: 1.8.6
- Azure Kubernetes Service
- Corefile: AKS-configured with "coredns-custom" ConfigMap.

chrisohaver:

For the `rcode` option, correct the default value and reference the source of all valid values (in miekg/dns package).

closes #5327


rtreffer (approved):
Right, the rcode is initialized as `0` by default and that is `NOERROR`.

This seems to be the case since my initial commit 0.o
Nice catch!

PR #5353, opened 2022-04-28
plugin/bind: fix a typo in test case

victory460:

 Fix a typo in bind plugin test case.

None.
None.
No.


chrisohaver (changes requested):
"noone" could be "no one" with the space removed.

This is not an important or necessary change.

victory460:
> space

Thanks!

PR #5354, opened 2022-04-28
plugin/bind: add support for bind IPv4 subnet ip

victory460:

 Add support for bind IPv4 subnet ip.
```
. {
    bind 10.0.0.0/16
}
```
None.

plugin/bind/README.md
No.

### DIFF
--- a/plugin/bind/README.md
+++ b/plugin/bind/README.md
@@ -82,6 +82,16 @@ You can exclude some addresses by their IP or interface name (The following will
 }
 ~~~
 
+If you want to bind one subnet hosts, you can use the following server block:
+
+```
+. {
+    bind 10.0.0.0/16
+}
+```
+
+Please make sure that these ip have been configured in advance on the network card.
+
 ## Bugs
 
 When defining more than one server block, take care not to bind more than one server to the same
--- a/plugin/bind/setup.go
+++ b/plugin/bind/setup.go
@@ -1,9 +1,11 @@
 package bind
 
 import (
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"net"
+	"strings"
 
 	"github.com/coredns/caddy"
 	"github.com/coredns/coredns/core/dnsserver"
@@ -91,10 +93,25 @@ func listIP(args []string, ifaces []net.Interface) ([]string, error) {
 			}
 		}
 		if !isIface {
-			if net.ParseIP(a) == nil {
-				return nil, fmt.Errorf("not a valid IP address or interface name: %q", a)
+			if strings.Contains(a, ".") && strings.Contains(a, "/") {
+				_, ipNet, err := net.ParseCIDR(a)
+				if err != nil {
+					return nil, fmt.Errorf("failed to get the ipNet: %q", a)
+				}
+				mask := binary.BigEndian.Uint32(ipNet.Mask)
+				start := binary.BigEndian.Uint32(ipNet.IP)
+				finish := (start & mask) | (mask ^ 0xffffffff)
+				for i := start; i <= finish; i++ {
+					ip := make(net.IP, 4)
+					binary.BigEndian.PutUint32(ip, i)
+					all = append(all, fmt.Sprint(ip))
+				}
+			} else {
+				if net.ParseIP(a) == nil {
+					return nil, fmt.Errorf("not a valid IP address or interface name: %q", a)
+				}
+				all = append(all, a)
 			}
-			all = append(all, a)
 		}
 	}
 	return all, nil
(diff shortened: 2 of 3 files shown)


chrisohaver (changes requested):
Please describe in detail what this new feature does (or rather what you intend it to do).

When you open a PR, the template asks some questions.  At a minimum, please answer those when you open PRs - especially when there is not a linked issue providing context.
* What does this PR do.
* Why is it needed.

victory460:
Thanks for you review.

> Please describe in detail what this new feature does (or rather what you intend it to do).
> 
> When you open a PR, the template asks some questions. At a minimum, please answer those when you open PRs - especially when there is not a linked issue providing context.
> 
> * What does this PR do.

This PR add ability to bind IPv4 subnet ip.
```
. {
    bind 10.0.0.0/16
}
```

> * Why is it needed.

Scenes: 
Different clients or different regions initiate DNS requests with different destination IP in the subnet .  
Then external auxiliary configuration routing will route these dns request traffic to the coredns with the pre-configured binding subnet ip.
Then you can get different serverIp, map to get different client/region, and then transfer it to next plugin to process it.

chrisohaver:
OK, that didn't clarify it for me at all.  But I'm guessing that...

By binding to a subnet, you intend it to be shorthand for binding to every address that falls within that subnet.

For example: Let's say you have a system with three addresses 10.25.0.10, 10.25.0.11, 10.25.30.10.
With this enhancement, `bind 10.25.0.0/24` would bind to 10.25.0.10 and 10.25.0.11, but not 10.25.30.10.
It would act the same as `bind 10.25.0.10 10.25.0.11 `

Does that accurately describe your intent?

chrisohaver:
> By binding to a subnet, you intend it to be shorthand for binding to every address that falls within that subnet.
>
> For example: Let's say you have a system with three addresses 10.25.0.10, 10.25.0.11, 10.25.30.10.
With this enhancement, bind 10.25.0.0/24 would bind to 10.25.0.10 and 10.25.0.11, but not 10.25.30.10.
It would act the same as bind 10.25.0.10 10.25.0.11


IIUC however, the code in the PR behaves differently.  It looks like it tries to bind to every possible IP address in the subnet, regardless of whether or not the addresses actually exists on the local system. If one of the addresses does not exist on the local system, I think CoreDNS will fail to start.

chrisohaver (changes requested):
It looks like the code tries to bind to every possible IP address in the subnet, regardless of whether or not the addresses actually exists on the local system. If any of the addresses do not exist on the local system, CoreDNS will fail to start.

The README needs to be updated to reflect the new feature.

victory460:
> OK, that didn't clarify it for me at all. But I'm guessing that...
> 
> By binding to a subnet, you intend it to be shorthand for binding to every address that falls within that subnet.
> 
> For example: Let's say you have a system with three addresses 10.25.0.10, 10.25.0.11, 10.25.30.10. With this enhancement, `bind 10.25.0.0/24` would bind to 10.25.0.10 and 10.25.0.11, but not 10.25.30.10. It would act the same as `bind 10.25.0.10 10.25.0.11 `
> 
> Does that accurately describe your intent?


For example scenes in the picture, configure different nameservers at machines in different vpcs, then we can analyze which vpc  the request comes from, according to the different server's ip obtained by coredns.

victory460:
> It looks like the code tries to bind to every possible IP address in the subnet, regardless of whether or not the addresses actually exists on the local system. If any of the addresses do not exist on the local system, CoreDNS will fail to start.
> 
> The README needs to be updated to reflect the new feature.

Thanks for your reply. I update the README about the new feature now.

chrisohaver (changes requested):
I'm not sold on inclusion of this feature.
* It doesn't add new functionality.  The same result could be achieved by listing each IP.
* It adds complexity.
* The use case it facilitates feels obscure and on the edge.

If we do decide this should be included, it should also ...
* support ipv6
* only bind to addresses that exist on the system AND fall within the subnet. IMO, this would make the feature slightly more useful (and may also simplify the code to iterate over local address instead of enumerating each ip in the subnet).

Freeaqingme:
I haven't thoroughly reviewed the code, but is this feature compatible with IPv6?

Especially this line makes me think that it was designed specifically for IPv4 only: 
> if strings.Contains(a, ".") && strings.Contains(a, "/") {

victory460:
> I haven't thoroughly reviewed the code, but is this feature compatible with IPv6?
> 
> Especially this line makes me think that it was designed specifically for IPv4 only:
> 
> > if strings.Contains(a, ".") && strings.Contains(a, "/") {

Thanks for you review. It was designed specifically for IPv4 only.

PR #5368, opened 2022-05-06
Don't add OPT RR to non-EDNS0 queries

### LINKED ISSUE #5366: Bufsize plugin causes non-RFC-compliant and sometimes too-large responses to non-EDNS0 clients
**What happened**:

If the bufsize plugin is enabled in CoreDNS, clients that don't support EDNS0 (and thus don't send an OPT RR) receive a reply with OPT RR that should not be sent. The reply is also sent using the maximum message size configured via bufsize (1232 in the example documentation) when the client only supports receiving 512-octet UDP responses. An oversize reply packet causes resolution failures in clients that don't support EDNS0 when querying a hostname and receiving a response containing >512 octets.

```
/ # nslookup longdns1.pahealy.com 10.0.70.209
Server:         10.0.70.209
Address:        10.0.70.209:53

Non-authoritative answer:
*** Can't find longdns1.pahealy.com: Parse error

*** Can't find longdns1.pahealy.com: No answer
```

Even if the bufsize plugin is enabled, responses to non-EDNS0 clients should still follow the original [RFC1035 section 2.3.4](https://datatracker.ietf.org/doc/html/rfc1035#section-2.3.4) requirements and send a maximum of 512 octets in the UDP payload. If the request is too large, it should be truncated and TC set. Additionally, per [RFC 6891 section 7](https://www.rfc-editor.org/rfc/rfc6891#section-7), no OPT RR record may be included if the client didn't send one in the initial request: ["Lack of presence of an OPT record in a request MUST be taken as an indication that the requestor does not implement any part of this specification and that the responder MUST NOT include an OPT record in its response."](https://www.rfc-editor.org/rfc/rfc6891#:~:text=Lack%20of%20presence%20of%20an%20OPT%20record%20in%20a%20request%20MUST%20be%20taken%20as%20an%0A%20%20%20indication%20that%20the%20requestor%20does%20not%20implement%20any%20part%20of%20this%0A%20%20%20specification%20and%20that%20the%20responder%20MUST%20NOT%20include%20an%20OPT%20record%0A%20%20%20in%20its%20response.) As called out in the [bufsize plugin considerations](https://github.com/coredns/coredns/blob/master/plugin/bufsize/README.md#considerations): "For now, if a client does not use EDNS, this plugin adds OPT RR", which violates the MUST NOT in the RFC.

Note that eventual resolution still fails in this example because busybox's nslookup doesn't support querying over TCP in the event of a record with TC set, and thus can't get the longer response. This bug is only about the failure to parse the initial UDP packet returned from CoreDNS.

```
/ # nslookup longdns1.pahealy.com 10.0.70.209
Server:         10.0.70.209
Address:        10.0.70.209:53

Non-authoritative answer:
longdns1.pahealy.com    canonical name = longdns-317c90a0654d341c6a770c08e5224f739331818a.pahealy.com
longdns-317c90a0654d341c6a770c08e5224f739331818a.pahealy.com    canonical name = longdns-72000071c119b7e8b6c2f4450d41916a6ef52542.pahealy.com
longdns-72000071c119b7e8b6c2f4450d41916a6ef52542.pahealy.com    canonical name = longdns-28f36a8c3d8e3c9fb1bfbfb3f1c0bb6bc3f4da0d.pahealy.com
longdns-28f36a8c3d8e3c9fb1bfbfb3f1c0bb6bc3f4da0d.pahealy.com    canonical name = longdns-1a85774a6a11551221a6cace8b6223a357082d36.pahealy.com
longdns-1a85774a6a11551221a6cace8b6223a357082d36.pahealy.com    canonical name = longdns-a6f98cf4df2b590df57a8f89648c39c4e01581ef.pahealy.com
longdns-a6f98cf4df2b590df57a8f89648c39c4e01581ef.pahealy.com    canonical name = longdns-1a68144a5b73afd68e204d954f7b490035f25b9b.pahealy.com
longdns-1a68144a5b73afd68e204d954f7b490035f25b9b.pahealy.com    canonical name = longdns-423187aaf2fd2ef1041866586fc46f9e0a9fcf79.pahealy.com

*** Can't find longdns1.pahealy.com: No answer
```

1. Deploy a coredns 1.9.1 instance (example kubernetes manifests: [deployment](https://gist.github.com/phealy/c6ed936dc77cb510183d0b7c88f55271), [configmap](https://gist.github.com/phealy/65f1160a4ff555f38838bf1e58408875), [service](https://gist.github.com/phealy/dba085c6a8c8387658ed07820b8ed84d)
2. Set "bufsize 1232" in the Corefile for `.:53` (see example Corefile below).
3. Use the nslookup from busybox 1.34.1 or comparable to query the coredns instance for a long domain entry that returns >512 octets in the response (I have an example up at `longdns1.pahealy.com`). The nslookup client from busybox is naive - it [hardcodes a length 512 unsigned char array](https://elixir.bootlin.com/busybox/latest/source/networking/nslookup.c#L518) for the reply and does not support TCP requests, so it won't do a valid resolution on a long answer, but it will at least show that the initial response parses successfully.

Example:

```
$ kubectl create namespace coredns-edns0
namespace/coredns-edns0 created

$ kubectl apply -n coredns-edns0 -f https://gist.githubusercontent.com/phealy/65f1160a4ff555f38838bf1e58408875/raw/2b32393e0af321dc32baedc6fd2305298ee90938/coredns-configmap.yaml
configmap/coredns created

$ kubectl apply -n coredns-edns0 -f https://gist.githubusercontent.com/phealy/c6ed936dc77cb510183d0b7c88f55271/raw/4c26aa97406cd9ca6624cb72635be57913354ee2/coredns-deployment.yaml
deployment.apps/coredns created

$ kubectl apply -n coredns-edns0 -f https://gist.githubusercontent.com/phealy/dba085c6a8c8387658ed07820b8ed84d/raw/9e4f84ce691c6760fc339c1211716269c5ddf625/coredns-service.yaml
service/coredns created

$ kubectl run busybox -n coredns-edns0 -it --rm --command --image=busybox:1.34.1 --overrides='{"spec":{"nodeSelector": {"kubernetes.io/os": "linux"}}}' -- nslookup longdns1.pahealy.com coredns.coredns.svc
If you don't see a command prompt, try pressing enter.
Server:         coredns.coredns.svc
Address:        10.0.70.209:53

Non-authoritative answer:
*** Can't find longdns1.pahealy.com: Parse error

*** Can't find longdns1.pahealy.com: No answer

Session ended, resume using 'kubectl attach busybox -c busybox -i -t' command when the pod is running
pod "busybox" deleted
```

**Environment**:

- the version of CoreDNS: verified on 1.9.1, 1.8.6, probably on all versions since the [introduction of bufsize in 1.6.6](https://github.com/coredns/coredns/pull/3401)
- Corefile:
  ```
  .:53 {
      errors
      log
      debug
      bufsize 1232
      ready
      health
      prometheus :9153
      forward . /etc/resolv.conf
      cache 30
      loop
      reload
      loadbalance
  }
  ```
- logs, if applicable:
  - bufsize 1232:
    ```
    coredns-577d746f8-8vv7t coredns [INFO] 172.26.4.131:56028 - 17408 "A IN longdns1.pahealy.com. udp 49 false 1232" NOERROR qr,rd,ra 2218 0.609804922s
    coredns-577d746f8-8vv7t coredns [INFO] 172.26.4.131:56028 - 17408 "AAAA IN longdns1.pahealy.com. udp 49 false 1232" NOERROR qr,aa,rd,ra 2142 0.000199002s
    ```

  - bufsize 512:
    ```
    coredns-7675bb7ccb-kgdlg coredns [INFO] 172.26.4.131:37205 - 21248 "A IN longdns1.pahealy.com. udp 49 false 512" NOERROR qr,tc,rd,ra 2218 0.454013244s
    coredns-7675bb7ccb-kgdlg coredns [INFO] 172.26.4.131:37205 - 21248 "AAAA IN longdns1.pahealy.com. udp 49 false 512" NOERROR qr,tc,rd,ra 2142 0.497982219s
    ```

  - no bufsize plugin:
    ```
    coredns-58674bb54b-j8w5x coredns [INFO] 172.26.4.131:41472 - 24832 "A IN longdns1.pahealy.com. udp 38 false 512" NOERROR qr,tc,rd,ra 2218 0.490688344s
    coredns-58674bb54b-j8w5x coredns [INFO] 172.26.4.131:41472 - 24832 "AAAA IN longdns1.pahealy.com. udp 38 false 512" NOERROR qr,tc,rd,ra 2142 0.526578513s
    ```

- OS (e.g: `cat /etc/os-release`): coredns/coredns:1.9.1 docker container
- Others:
[ZIP file containing packet captures of lookups from busybox nslookup with no bufsize plugin, bufsize 512, and bufsize 1232](https://github.com/coredns/coredns/files/8632228/packet-captures.zip)

phealy:
Signed-off-by: Patrick W. Healy <phealy@phealy.com>

It stops the bufsize plugin from adding an OPT RR to replies where the query didn't contain an OPT RR, making the bufsize plugin compliant with [RFC 6891 section 7](https://www.rfc-editor.org/rfc/rfc6891#section-7): ["Lack of presence of an OPT record in a request MUST be taken as an indication that the requestor does not implement any part of this specification and that the responder MUST NOT include an OPT record in its response."](https://www.rfc-editor.org/rfc/rfc6891#:~:text=Lack%20of%20presence%20of%20an%20OPT%20record%20in%20a%20request%20MUST%20be%20taken%20as%20an%0A%20%20%20indication%20that%20the%20requestor%20does%20not%20implement%20any%20part%20of%20this%0A%20%20%20specification%20and%20that%20the%20responder%20MUST%20NOT%20include%20an%20OPT%20record%0A%20%20%20in%20its%20response.)

Additionally, this behavior would cause the server to send responses that could be larger than the client could handle; non-EDNS0 clients only expect 512 bytes in a UDP reply, but if the plugin was enabled the server would send UDP packets of up to <bufsize>, causing resolution failures for any reply that was less than 512 bytes but less or equal than the configured value for bufsize.

#5366 - Bufsize plugin causes non-RFC-compliant and sometimes too-large responses to non-EDNS0 clients

Only README.md, which is already updated.

No

### DIFF
--- a/plugin/bufsize/README.md
+++ b/plugin/bufsize/README.md
@@ -5,6 +5,7 @@
 ## Description
 *bufsize* limits a requester's UDP payload size.
 It prevents IP fragmentation, mitigating certain DNS vulnerabilities.
+This will only affect query replies to EDNS0-capable clients; i.e. queries that have an OPT RR.
 
 ## Syntax
 ```txt
@@ -36,4 +37,3 @@ Enable limiting the buffer size as an authoritative nameserver:
 
 ## Considerations
 - Setting 1232 bytes to bufsize may avoid fragmentation on the majority of networks in use today, but it depends on the MTU of the physical network links.
-- For now, if a client does not use EDNS, this plugin adds OPT RR.
--- a/plugin/bufsize/README.md
+++ b/plugin/bufsize/README.md
@@ -5,7 +5,7 @@
 ## Description
 *bufsize* limits a requester's UDP payload size.
 It prevents IP fragmentation, mitigating certain DNS vulnerabilities.
-This will only affect query replies to EDNS0-capable clients; i.e. queries that have an OPT RR.
+This will only affect queries that have an OPT RR.
 
 ## Syntax
 ```txt
--- a/plugin/bufsize/bufsize.go
+++ b/plugin/bufsize/bufsize.go
@@ -19,9 +19,6 @@ type Bufsize struct {
 func (buf Bufsize) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 	if option := r.IsEdns0(); option != nil {
 		option.SetUDPSize(uint16(buf.Size))
-	} else {
-		// If a client does not use EDNS, add it
-		r.SetEdns0(uint16(buf.Size), false)
 	}
 
 	return plugin.NextOrFailure(buf.Name(), buf.Next, ctx, w, r)
--- a/plugin/bufsize/bufsize_test.go
+++ b/plugin/bufsize/bufsize_test.go
@@ -31,7 +31,7 @@ func TestBufsize(t *testing.T) {
 			outgoingBufsize: 512,
 			expectedErr:     nil,
 		},
-		// If EDNS is not enabled, this plugin adds it
+		// If EDNS is not enabled, this plugin should not add it
 		{
 			next:            whoami.Whoami{},
 			qname:           ".",
@@ -65,6 +65,14 @@ func TestBufsize(t *testing.T) {
 					}
 				} else {
 					t.Errorf("Test %d: Not found OPT RR.", i)
+				}
+			}
+		}
+
+		if tc.inputBufsize == 0 {
+			for _, extra := range req.Extra {
+				if _, ok := extra.(*dns.OPT); ok {
+					t.Errorf("Test %d: Found OPT RR on reply to query with no OPT RR.", i)
 				}
 			}
 		}


chrisohaver (approved):
LGTM

Despite being documented that the plugin adds EDNS to a non-EDNS request, no reason was given as to why.
I can't think of a situation where adding EDNS to a non-EDNS request would be desired.
... and doing to violates RFC.

In theory, this would be backward incompatible with someone depending on this behavior.  But since I can't think of how this behavior would actually benefit any working scenario,  I think it's ok to merge now without waiting until a next minor release.

plugin/bufsize/README.md
```diff
@@ -5,6 +5,7 @@
 ## Description
 *bufsize* limits a requester's UDP payload size.
 It prevents IP fragmentation, mitigating certain DNS vulnerabilities.
+This will only affect query replies to EDNS0-capable clients; i.e. queries that have an OPT RR.
```
chrisohaver:
```suggestion
This will only affect queries that have an OPT RR.
```

phealy:
@chrisohaver Good day! Had a follow-up question for you. Is there any chance of getting this PR and #5401 cherry-picked against 1.8.7 to make a 1.8.8 release? They're hitting a few of our end users, but we try not to rev core components across a minor version once they've been released (i.e. we released Kubernetes 1.23 with CoreDNS 1.8, so we need to stick with 1.8 for it to avoid introducing a breaking change during a patch release).

Thanks!

chrisohaver:
> @chrisohaver Good day! Had a follow-up question for you. Is there any chance of getting this PR and #5401 cherry-picked against 1.8.7 to make a 1.8.8 release? They're hitting a few of our end users, but we try not to rev core components across a minor version once they've been released (i.e. we released Kubernetes 1.23 with CoreDNS 1.8, so we need to stick with 1.8 for it to avoid introducing a breaking change during a patch release).
> 
> Thanks!

CoreDNS has been on a single release track thus far. We have not yet released a patch release that is lower than the latest release at that time.

What breaking change in 1.9 are you running into?

phealy:
Nothing in 1.9 specifically - but we have hit [breaking changes in the past](https://github.com/coredns/coredns/issues/4834) and so we're somewhat stuck with our default versioning policy, which is that we minimize the chance of changes causing issues by only applying patch releases to supporting components within a single Kubernetes version (so we won't upgrade customers on 1.23 past coredns 1.8.x by default). We'll pull in the 1.9 changes with our next release, but we want to see if we can get these fixed in the past version first.

If CoreDNS's release policy isn't compatible with a 1.8.8 release, we'll cherry pick the fixes to 1.8.7 internally and build a 1.8.7-patch release, but we wanted to ask here first.

Thanks!

PR #5377, opened 2022-05-11
plugin/dnstap: support FQDN TCP endpoint 

### LINKED ISSUE #5355: FQDN as dnstap TCP endpoint
At the moment, `coredns` supports `dnstap` with Unix sockets and TCP, which is awesome. however, the TCP support looks to be limited to IP:PORT. It would be useful to have `dnstap` with a FQDN remote socket, something like below:

`dnstap tcp://my-service.kube-system.svc.cluster.local:5000 full`

**Why is this needed**:

This makes it easier to manage `dnstap` logger pods in a k8s environment using the Service object type. At the moment there's a need to extract the Static IP of the service/pod and manually populate the ConfigMap which is not ideal.

lobshunter:
Previously used parsing function only allows IP address as TCP endpoint.
This PR resolves TCP endpoint using `url.Parse` instead to support FQDN.

Relates to #5355.

Included

No

### DIFF
--- a/plugin/dnstap/README.md
+++ b/plugin/dnstap/README.md
@@ -41,6 +41,12 @@ Log to a remote endpoint.
 dnstap tcp://127.0.0.1:6000 full
 ~~~
 
+Log to a remote endpoint by FQDN.
+
+~~~ txt
+dnstap tcp://example.com:6000 full
+~~~
+
 ## Command Line Tool
 
 Dnstap has a command line tool that can be used to inspect the logging. The tool can be found
--- a/plugin/dnstap/io.go
+++ b/plugin/dnstap/io.go
@@ -25,7 +25,6 @@ type tapper interface {
 type dio struct {
 	endpoint     string
 	proto        string
-	conn         net.Conn
 	enc          *encoder
 	queue        chan tap.Dnstap
 	dropped      uint32
--- a/plugin/dnstap/setup.go
+++ b/plugin/dnstap/setup.go
@@ -1,13 +1,13 @@
 package dnstap
 
 import (
+	"net/url"
 	"strings"
 
 	"github.com/coredns/caddy"
 	"github.com/coredns/coredns/core/dnsserver"
 	"github.com/coredns/coredns/plugin"
 	clog "github.com/coredns/coredns/plugin/pkg/log"
-	"github.com/coredns/coredns/plugin/pkg/parse"
 )
 
 var log = clog.NewWithPlugin("dnstap")
@@ -24,12 +24,12 @@ func parseConfig(c *caddy.Controller) (Dnstap, error) {
 	}
 
 	if strings.HasPrefix(endpoint, "tcp://") {
-		// remote IP endpoint
-		servers, err := parse.HostPortOrFile(endpoint[6:])
+		// remote network endpoint
+		endpointURL, err := url.Parse(endpoint)
 		if err != nil {
 			return d, c.ArgErr()
 		}
-		dio := newIO("tcp", servers[0])
+		dio := newIO("tcp", endpointURL.Host)
 		d = Dnstap{io: dio}
 	} else {
 		endpoint = strings.TrimPrefix(endpoint, "unix://")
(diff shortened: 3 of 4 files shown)


chrisohaver:
This won't close #5355, for the reasons discussed in that issue.

lobshunter:
Forgot the resolving FQDN part, my bad. Currently I am not sure what upstream server will CoreDNS use to resolve domain name. I will do some tesing on kubernetes and bare metal.

chrisohaver:
> Forgot the resolving FQDN part, my bad. Currently I am not sure what upstream server will CoreDNS use to resolve domain name. I will do some tesing on kubernetes and bare metal.

IIUC, it (`net.Dial`) will use the "system resolver", which is what is in the Pod's `/etc/resolv.conf`.  For the CoreDNS pod in K8s, this is the same as the `/etc/resolv.conf` on the node (which cannot resolve k8s service names).

lobshunter:
> > Forgot the resolving FQDN part, my bad. Currently I am not sure what upstream server will CoreDNS use to resolve domain name. I will do some tesing on kubernetes and bare metal.
> 
> IIUC, it (`net.Dial`) will use the "system resolver", which is what is in the Pod's `/etc/resolv.conf`. For the CoreDNS pod in K8s, this is the same as the `/etc/resolv.conf` on the node (which cannot resolve k8s service names).

My understanding is different. CoreDNS itself is a deployment pod in kubernetes cluster, so it's /etc/resolv.conf points to in-cluster DNS server(which is CoreDNS itself). In that case, resolution request of dnstap can be handled by kubernetes plugin properly. But I cannot be not fully sure without actual testing.

I shall get some time to setup a cluster to test it.

lobshunter:
> My understanding is different

And I am wrong. Kubernetes uses [dnsPolicy](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy) configuration for pod. To CoreDNS, the value is `Default`, so `/etc/resolv.conf` inherits from node.(Why wouldn't it, otherwise there could be DNS loop 😈)

Specifying an upstream DNS server should work, the server can be CoreDNS itself. Config block can be something like
```
dnstap SOCKET [full] [UPSTREAM]
dnstap tcp://example.com:5000 full 127.0.0.1:53
```
I am not sure if it is a good approach.

chrisohaver:
> Specifying an upstream DNS server should work, the server can be CoreDNS itself. Config block can be something like
>```
>dnstap SOCKET [full] [UPSTREAM]
>dnstap tcp://example.com:5000 full 127.0.0.1:53
>```
>I am not sure if it is a good approach.

Yes, my thoughts exactly.  It doesn't feel like a good approach because this isn't a problem that only affects _dnstap_, it potentially affects any plugin that uses a hostname to connect to something, so a more global way of configuring the resolver would seem more appropriate.  But as far as implementing that - in *dnstap* we actually to the dialing, and (I think) we can set the resolver.  For other plugins, the dial may happen in a client library, and the resolver may not be configurable by us.  And looking up the IP in advance and using the IP  to connect could disrupt TLS negotiations.  So, I don't immediately see a good solution.

I think the PR is fine as is.  On its own, it doesn't close #5355, but that's OK, it is still useful for when the name _is_ resolvable by upstream servers.
The solution to #5355 in particular could involve this PR, plus deploying an upstream coredns with kubernetes plugin (i.e. enabling upstream DNS to resolve cluster names).

lobshunter:
> The solution to https://github.com/coredns/coredns/issues/5355 in particular could involve this PR, plus deploying an upstream coredns with kubernetes plugin (i.e. enabling upstream DNS to resolve cluster names).

That's true. I made it work by setting CoreDNS's dnsPolicy to ClusterFirst, and modifying forward plugin setting not to use `/etc/resolv.conf` but an explicit upstream server. But I believe this is a weird approach. Enabling upstream DNS to resolve cluster names should be better than this.

Enabling this PR for it _could_ be useful sometimes.

chrisohaver:
>  I made it work by setting CoreDNS's dnsPolicy to ClusterFirst, and modifying forward plugin setting not to use /etc/resolv.conf but an explicit upstream server. But I believe this is a weird approach. Enabling upstream DNS to resolve cluster names should be better than this.

I think your method works better. It assumes the nodes all share the same upstreams, but I would think is usually the case.  Much easier to configure, and the request doesn't travel outside the cluster.

chrisohaver (approved):

johnbelamaric:
When serving as the cluster DNS, the CoreDNS pod uses the Default DNS
policy which uses the host resolv.conf. Otherwise it would be
self-referential and bad things would happen…

Issue #5376, opened 2022-05-11
plugin/acl: Not working when access with ipv6 link local address


BoringCat:
**What happened**:
Set acl to block all AAAA response. But still got response

**What you expected to happen**:
nslookup got
```
# nslookup -type=AAAA *SOMEDOMAIN* fe80::aaaa:bbbb:cccc:dddd%eth0
** server can't find *SOMEDOMAIN*: REFUSED
```

**How to reproduce it (as minimally and precisely as possible)**:
*tip: Same as Corefile below
1. listen zone `.`
2. set forward to /etc/resolv.conf
3. set acl `block type AAAA`
4. send request

**Anything else we need to know?**:
I do some test and find why this issue happened

- First, acl get client ip in here:  
  https://github.com/coredns/coredns/blob/master/plugin/acl/acl.go#L99  
  And it return by `dns.ResponseWriter.RemoteAddr().String()`  
  At  https://github.com/coredns/coredns/blob/master/request/request.go#L44-L57  
- Then, when client ip is ipv6 link local. `state.IP()` will return `Addr%ifname` instead `Addr`.   
  example: 
    - IPv6 global unicast: `2001:1111:2222:3333:4444:5555:6666:7777`  
    - IPv6 link local: `fe80::aaaa:bbbb:cccc:dddd%eth0`  
- But, `net.ParseIP` can not deal with it. Will return `<nil>`
- And at https://github.com/coredns/coredns/blob/master/plugin/acl/acl.go#L109-L112  
  when send `<nil>` to `policy.filter.GetByIP`. It return `<nil> false`. And passed `if !contained`

So acl think not need to handle this request

**Environment**:
```
Linux Router 5.4.162 #0 SMP PREEMPT Sat Dec 18 02:50:04 2021 aarch64 GNU/Linux
```

- the version of CoreDNS:
  ```
  CoreDNS-1.9.1
  linux/arm64, go1.17.8, 4b597f8
  ```
- Corefile:
  ```
  . {
      prometheus :9153
      acl {
          block type AAAA
      }
      forward . /etc/resolv.conf
  }
  ```
- logs, if applicable:  
  plugin/log is disabled.
- OS (e.g: `cat /etc/os-release`):
  ```
  NAME="OpenWrt"
  VERSION="SNAPSHOT"
  ID="openwrt"
  ID_LIKE="lede openwrt"
  PRETTY_NAME="OpenWrt SNAPSHOT"
  VERSION_ID="snapshot"
[...]

PR #5387, opened 2022-05-18
Fixing issue #5376 by adding a check to parse out Zone info

SamrathPalSingh:
This fixes issue #5376 by simply removing any Zone info in the IP address before net.ParseIP(). This also adds a test in the acl-test  file to test the new changes.

It also blocks any queries which fail in the parsing step.

### DIFF
--- a/plugin/acl/acl.go
+++ b/plugin/acl/acl.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/metrics"
+	"github.com/coredns/coredns/plugin/pkg/log"
 	"github.com/coredns/coredns/request"
 
 	"github.com/infobloxopen/go-trees/iptree"
@@ -98,11 +99,17 @@ RulesCheckLoop:
 func matchWithPolicies(policies []policy, w dns.ResponseWriter, r *dns.Msg) action {
 	state := request.Request{W: w, Req: r}
 
-	ip := net.ParseIP(strings.Split(state.IP(), "%")[0])
+	ip := []byte(nil)
+	if idx := strings.IndexByte(state.IP(), '%'); idx >= 0 {
+		ip = net.ParseIP(state.IP()[:idx])
+	} else {
+		ip = net.ParseIP(state.IP())
+	}
 
 	// if the parsing did not return a proper response then we simply return 'actionBlock' to
 	// block the query
-	if len(ip) == 0 {
+	if ip == nil {
+		log.Errorf("ParseIP failed for ip address: %v", state.IP())
 		return actionBlock
 	}
 	qtype := state.QType()
--- a/plugin/acl/acl.go
+++ b/plugin/acl/acl.go
@@ -8,11 +8,12 @@ import (
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/metrics"
-	"github.com/coredns/coredns/plugin/pkg/log"
 	"github.com/coredns/coredns/request"
 
 	"github.com/infobloxopen/go-trees/iptree"
 	"github.com/miekg/dns"
+
+	clog "github.com/coredns/coredns/plugin/pkg/log"
 )
 
 // ACL enforces access control policies on DNS queries.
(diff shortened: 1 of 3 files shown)


SamrathPalSingh:
@chrisohaver I'm not so sure about the second change, where we block any requests which fail the parsing step. 
```
	if len(ip) == 0 {
		return actionBlock
	}
```
Should I remove/edit it ?

plugin/acl/acl.go
```diff
@@ -96,7 +98,13 @@ RulesCheckLoop:
 func matchWithPolicies(policies []policy, w dns.ResponseWriter, r *dns.Msg) action {
 	state := request.Request{W: w, Req: r}
 
-	ip := net.ParseIP(state.IP())
+	ip := net.ParseIP(strings.Split(state.IP(), "%")[0])
+
+	// if the parsing did not return a proper response then we simply return 'actionBlock' to
+	// block the query
+	if len(ip) == 0 {
```
chrisohaver: Probably better to compare the result to nil to check if parsing failed. i.e. `if ip == nil {`

plugin/acl/acl.go
```diff
@@ -96,7 +98,13 @@ RulesCheckLoop:
 func matchWithPolicies(policies []policy, w dns.ResponseWriter, r *dns.Msg) action {
 	state := request.Request{W: w, Req: r}
 
-	ip := net.ParseIP(state.IP())
+	ip := net.ParseIP(strings.Split(state.IP(), "%")[0])
```
chrisohaver: It's probably slightly less expensive to use `strings.IndexByte` in place of `Split` (assuming Go compiler doesn't optimize based on the literal `[0]`)... e.g. something like the following:

```go
if idx := strings.IndexByte(state.IP(), '%'); idx >= 0 {
    ip = net.ParseIP(state.IP()[:idx])
}

File: plugin/acl/acl.go [hunk 3]
```diff
@@ -96,7 +98,13 @@ RulesCheckLoop:
...
+
+	// if the parsing did not return a proper response then we simply return 'actionBlock' to
+	// block the query
+	if len(ip) == 0 {
+		return actionBlock
```
chrisohaver, line 114: If we can't parse the trimmed IP string, we should log an error.
I agree that blocking is probably the best action here.

chrisohaver (changes requested):
Thanks!

plugin/acl/acl.go
SamrathPalSingh: @chrisohaver - added a log statement which mentions the failing IP

plugin/acl/acl.go
```diff
@@ -96,7 +99,19 @@ RulesCheckLoop:
 func matchWithPolicies(policies []policy, w dns.ResponseWriter, r *dns.Msg) action {
 	state := request.Request{W: w, Req: r}
 
-	ip := net.ParseIP(state.IP())
+	ip := []byte(nil)
```
chrisohaver:
```suggestion
	var ip net.IP
```

plugin/acl/acl.go
```diff
@@ -96,7 +99,19 @@ RulesCheckLoop:
...
+	if idx := strings.IndexByte(state.IP(), '%'); idx >= 0 {
+		ip = net.ParseIP(state.IP()[:idx])
+	} else {
+		ip = net.ParseIP(state.IP())
+	}
+
+	// if the parsing did not return a proper response then we simply return 'actionBlock' to
+	// block the query
+	if ip == nil {
+		log.Errorf("ParseIP failed for ip address: %v", state.IP())
```
chrisohaver: Please use the plugin logger  https://github.com/coredns/coredns/blob/master/plugin/pkg/log/plugin.go. It logs the plugin name along with the message.

example: https://github.com/coredns/coredns/blob/master/plugin/cache/setup.go#L17

A better error message would be "Blocking request. Unable to parse source address: ".

chrisohaver (changes requested):

plugin/acl/acl.go
SamrathPalSingh: Just for my knowledge, is there an actual difference between using the var keyword and the short declaration operator in this particular case ?

Changing []byte to net.IP, I realize is safer.

chrisohaver (commented):
Need to fix the import ordering/grouping, but looks good otherwise.  (this is also just a style thing, but it is enforced in the tests)

plugin/acl/acl.go
chrisohaver: no difference afaik.

IMO, if declaring a new empty variable, var is preferred style. If also assigning a value at the same time, then := is fine.

SamrathPalSingh:
@chrisohaver fixed the import ordering. All check have passed.

chrisohaver (approved):
Lgtm

yongtang (approved):

ihac:
LGTM

PR #5388, opened 2022-05-19
Correct timeout description

RetoHaslerMGB:

It corrects the timeout values in the documentation (currently the documentation says 100ms dial timeout, but actually its 1000ms=1s)

```
// plugin/forward/persistent.go starting at line 151
const (
	defaultExpire  = 10 * time.Second
	minDialTimeout = 1 * time.Second  // <<<<< DEFINITION IN THE CODE
	maxDialTimeout = 30 * time.Second
)
```

None

No futher documentation adjustment needed. This already fixes the documentation.

Since this only changes the documentation I would not excpect an incompatible change or deprecation.

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -103,7 +103,7 @@ Also note the TLS config is "global" for the whole forwarding proxy if you need 
 
 On each endpoint, the timeouts for communication are set as follows:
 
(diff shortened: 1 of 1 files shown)


yongtang:
@RetoHaslerMGB Can you fix the DCO error?

RetoHaslerMGB:
@yongtang  DCO should now be fixed.

yongtang (approved):

PR #5414, opened 2022-05-28
Warn when domain names are not in RFC1035 preferred syntax

### LINKED ISSUE #5386: Warn of potentially malformed domains in Corefiles
[Via a Slack conversation](https://cloud-native.slack.com/archives/C4DF7FP71/p1652741093205869) with @chrisohaver.

**What you would like to be added**: A warning about potentially malformed domains in `Corefile`s: e.g., `k8s.foo.bar.com::53` (with two colons). 

**Why this is needed**: CoreDNS will parse the domain as if `com:` were a valid domain ([it technically is](https://cloud-native.slack.com/archives/C4DF7FP71/p1652753134359439?thread_ts=1652741093.205869&cid=C4DF7FP71)), and serve requests referencing `k8s.foo.bar.com` elsewise, probably using the default provider. In my specific case, this resulted in `NXDOMAIN`s when `Pod`s tried to resolve cluster-local `Service`s.

mdsahiloss:

Log a warning when server block zones do not adhere to RFC1035's  "Preferred Syntax"

This PR fixes #5386

No documentation changes.

No.


mdsahiloss:
If Any reviewer find any issue with this PR so let me know I'll try to fix that.

core/dnsserver/onstartup.go
```diff
@@ -11,6 +12,13 @@ import (
 func startUpZones(protocol, addr string, zones map[string]*Config) string {
 	s := ""
 
+	// Below block of code checks whether the given domain-name is correct or not.
+	// If not correct then shows a warning against domain-name.
+	matchDomain, _ := regexp.Compile("[A-Za-z]+://([A-Za-z0-9-]+\\.[A-Za-z0-9-]+)*:[0-9]+")
+	if !matchDomain.MatchString(addr) {
+		fmt.Println("Warning: Malformed domain-name " + addr)
```
chrisohaver: Probably better too append the message to `s` than to print out directly.  Would make it easier to test too.

chrisohaver: Technically it is valid, and not malformed. The message should change to something like ...

`Warning: Domain does not follow RFC1035 preferred syntax`

core/dnsserver/onstartup.go
```diff
@@ -11,6 +12,13 @@ import (
 func startUpZones(protocol, addr string, zones map[string]*Config) string {
 	s := ""
 
+	// Below block of code checks whether the given domain-name is correct or not.
+	// If not correct then shows a warning against domain-name.
```
chrisohaver: replace "correct" with "following RFC1035 preferred syntax"

chrisohaver (changes requested):

chrisohaver (commented):
Please also add a unit test.

mdsahiloss:
@chrisohaver I think, Now It's good to go.

core/dnsserver/onstartup.go
```diff
@@ -2,15 +2,27 @@ package dnsserver
 
 import (
 	"fmt"
+	"regexp"
 	"sort"
 )
 
+// Below checkDomainName() function checks whether the given domain-name is following RFC1035 preferred syntax or not.
+// If not correct then shows a warning against domain-name.
```
chrisohaver:
```suggestion
// checkDomainName() returns true if the given domain name follows RFC1035 preferred syntax
```

core/dnsserver/onstartup.go
```diff
@@ -2,15 +2,27 @@ package dnsserver
...
+}
+
 // startUpZones creates the text that we show when starting up:
 // grpc://example.com.:1055
 // example.com.:1053 on 127.0.0.1
 func startUpZones(protocol, addr string, zones map[string]*Config) string {
 	s := ""
 
+	if !checkDomainName(addr) {
+		s += fmt.Sprintln("Warning: Domain does not follow RFC1035 preferred syntax")
```
chrisohaver: Should include the offending name in the warning.

core/dnsserver/onstartup_test.go
```diff
@@ -0,0 +1,9 @@
+package dnsserver
+
+import "testing"
+
+func TestCheckDomainName(t *testing.T) {
+	if !checkDomainName("dns://example.com:53") {
+		t.Fatalf(`Something went wrong, Now domain name doesn't match with Regular-Expression`)
+	}
```
chrisohaver: There should several positive and negative test cases.  i.e. some cases that you expect to get a true result, and some that expect a false result.  

The test failure message should describe the test case and the expected result.  
e.g. `Expected true for "an-example.com"`
or `Expected false for "example-?&^%$.com"`

core/dnsserver/onstartup.go
```diff
@@ -2,15 +2,27 @@ package dnsserver
...
+	return matchDomain.MatchString(addr)
+}
+
 // startUpZones creates the text that we show when starting up:
 // grpc://example.com.:1055
 // example.com.:1053 on 127.0.0.1
 func startUpZones(protocol, addr string, zones map[string]*Config) string {
 	s := ""
 
+	if !checkDomainName(addr) {
```
chrisohaver: `addr` is the wrong thing to validate here.  

Each zone key in `zones` needs to be validated.

core/dnsserver/onstartup.go
```diff
@@ -2,15 +2,27 @@ package dnsserver
...
 import (
 	"fmt"
+	"regexp"
 	"sort"
 )
 
+// Below checkDomainName() function checks whether the given domain-name is following RFC1035 preferred syntax or not.
+// If not correct then shows a warning against domain-name.
+func checkDomainName(addr string) bool {
+	matchDomain, _ := regexp.Compile("[A-Za-z]+://([A-Za-z0-9-]+\\.[A-Za-z0-9-]+)*:[0-9]+")
```
chrisohaver: This regex should only match a domain name, not the protocol prefix and port suffix.

chrisohaver (changes requested):

core/dnsserver/onstartup_test.go
```diff
@@ -0,0 +1,12 @@
...
+
+import "testing"
+
+func TestCheckDomainName(t *testing.T) {
+	if !checkDomainName("an-example.com") {
+		t.Fatalf(`Expected true for "an-example.com"`)
+	}
+	if !checkDomainName("example-?&^%$.com") {
+		t.Fatalf(`Expected false for "example-?&^$.com"`)
+	}
```
mariuskimmina: Why do both test cases have an `!`? 
I think something is wrong here.

mariuskimmina: The second test should fail if `checkDomainName` returns true, you have to remove the `!` there. Also your `checkDomainName` function does in fact return true in that second case, which is unintended I think.

core/dnsserver/onstartup.go
```diff
@@ -2,9 +2,16 @@ package dnsserver
 
 import (
 	"fmt"
+	"regexp"
 	"sort"
 )
 
+// checkDomainName() returns true if the given domain name follows RFC1035 preferred syntax.
+func checkDomainName(domainName string) bool {
+	matchDomain, _ := regexp.Compile("([A-Za-z0-9-]+\\.[A-Za-z0-9-]+)*")
```
chrisohaver: Pretty sure this will match a compliant substring in a non-compliant domain name.  
Need to anchor the beginning and end of the regex.  
also pretty sure domain always ends in a dot. 

Please compile coredns with your changes and test both compliant and non-compliant domains. 
Easier / quicker to see if your code works than to iterate with reviews.

core/dnsserver/onstartup_test.go
mdsahiloss: This is because I'm directly sending a domain-name as an argument to `checkDomainName(domainName)` function and it returns `true` value when domain-name matches Regex and i think we don't need to show  `Expected true for "an-example.com"` output when domain-name matches expected Regex.
whereas, when we pass an wrong `domain-name` as an argument then `checkDomainName(domainName)` returns `false` and I think at that It's needed to print `Expected false for "example-?&^$.com"` that's why I think putting a Not `!` before `checkDomainName(domainName)` gives exprected result.

mariuskimmina: But there is a problem, at the moment your function `checkDomainNames` returns `true` in both cases, which is is unintended and your test is not catching that.

I tried your function in a test program
```
package main

import (
        "fmt"
        "regexp"
)

func checkDomainName(domainName string) bool {
        matchDomain, _ := regexp.Compile("([A-Za-z0-9-]+\\.[A-Za-z0-9-]+)*")
        return matchDomain.MatchString(domainName)
}

func main() {
    output := checkDomainName("an-example.com")
    fmt.Println(output)
    output = checkDomainName("example-?&^%$.com")
    fmt.Println(output)
}
```

And it get the following output 

```
true
true
```

mdsahiloss: Thanks for letting me know,As I can see there is an issue with the Regex so I think I need to improve it.

core/dnsserver/onstartup.go
mdsahiloss: @chrisohaver  How should I anchor beginning and ending of a domain-name?, I think ending should be anchored by a single `:` but what about beginning.

chrisohaver: https://www.regular-expressions.info/anchors.html

mdsahiloss: Thanks for sharing the docs link, want to ask that, Are you sure? you want a dot `.` at the ending of Regex as ending anchor .

chrisohaver: Two unrelated things ...

1. The regex needs to be anchored at both ends, otherwise it could successfully match a substring of an invalid domain.  e.g. `#!@.example.com.`, would match your current regex because the substring `example.com` matches.
2. Not related to regex anchoring: the variable `zone` will always end in a dot (`.`) because it is a fully qualified domain name (normalized earlier in execution).


(This will be less mysterious if you compile your changes, execute the resulting coredns binary with various  compliant and non-compliant domains, and observe the output)

mdsahiloss: I have improved(anchored beginning and ending) Regex as per your suggestion, means there must be a dot `.` at the ending of Regex for an example if `zone == an-example.com.` then `checkDomainName()` will return true or else will return false although `zone == an-example.com`. Let me know if there is still any issue

chrisohaver:
Please rebase with latest master and sign the DCO.

core/dnsserver/onstartup.go
```diff
@@ -2,9 +2,16 @@ package dnsserver
 
 import (
 	"fmt"
+	"regexp"
 	"sort"
 )
 
+// checkDomainName() returns true if the given domain name follows RFC1035 preferred syntax.
+func checkDomainName(domainName string) bool {
+	matchDomain, _ := regexp.Compile("^[A-Za-z0-9-]{1,}\\.[A-Za-z0-9-]{1,}\\.$")
```
chrisohaver: This will fail to match the root domain, `.`. 

Can you share the Corefile you are trying with?

core/dnsserver/onstartup_test.go
```diff
@@ -0,0 +1,12 @@
+package dnsserver
+
+import "testing"
+
+func TestCheckDomainName(t *testing.T) {
+	if !checkDomainName("an-example.com") {
```
chrisohaver: Won’t this fail with your latest commit? There is no trailing dot in the tested input.

core/dnsserver/onstartup_test.go
```diff
@@ -0,0 +1,12 @@
...
+		t.Fatalf(`Expected true for "an-example.com"`)
+	}
+	if !checkDomainName("example-?&^%$.com") {
```
chrisohaver: This logic is backwards.

chrisohaver (changes requested):

core/dnsserver/onstartup_test.go
mariuskimmina: I think this should really be changed to something like:

```
if checkDomainName("example-?&^%$.com.") {
    t.Fatalf(`Expected false but received true for "example-?&^$.com."`)
}
```

mariuskimmina (changes requested):

core/dnsserver/onstartup.go
mdsahiloss: okay.. let me update Regex for root domain too.

core/dnsserver/onstartup_test.go
mdsahiloss: I think, this will fail, I need to update test too.

core/dnsserver/onstartup_test.go
mdsahiloss: Got it.

core/dnsserver/onstartup_test.go
```diff
@@ -0,0 +1,14 @@
...
+
+import "testing"
+
+func TestCheckDomainName(t *testing.T) {
+
+	// Note: checkDomainName() returns true for RFC1035 preferred syntax.
+	if !checkDomainName("an-example.com.") {
+		t.Fatalf(`Expected true but received false for "an-example.com"`)
+	}
+	if checkDomainName("example-?&^%$.com") {
```
chrisohaver: Add a trailing dot to this test.

core/dnsserver/onstartup_test.go
```diff
@@ -0,0 +1,14 @@
...
+		t.Fatalf(`Expected false but received true for "example-?&^$.com"`)
+	}
+}
```
chrisohaver, line 40: You need more test cases here.  At the very least you should add the root test case, since your regex didn't account for it initially.

coremain/version.go
```diff
@@ -2,7 +2,7 @@ package coremain
```
chrisohaver: This file should not be here.  Please update your master and rebase.

notes/coredns-1.9.3.md
```diff
@@ -0,0 +1,31 @@
++++
```
chrisohaver: This file should not be here.  Please update your master and rebase.

chrisohaver (changes requested):
This code always crashes.

It is expected for you to compile and test your changes manually and also run your unit tests before submitting PRs for review.  This is a base expectation of all contributors to open source projects.

Please let me know if you need help or guidance on how to build coredns with your changes.

core/dnsserver/onstartup.go
```diff
@@ -2,9 +2,16 @@ package dnsserver
 
 import (
 	"fmt"
+	"regexp"
 	"sort"
 )
 
+// checkDomainName() returns true if the given domain name follows RFC1035 preferred syntax.
+func checkDomainName(domainName string) bool {
+	matchDomain, _ := regexp.Compile("[A-Za-z0-9-]{1,}\\.[A-Za-z0-9-]{1,}\\.$)|^\\.$")
```
chrisohaver: Suggestion: It would be easier to first check for the root domain case separately than to figure out how to do it in regex.  Also, it would result in a shorter and easier to read regex.

Also, There is no benefit to compiling the regex in a separate step since it's only being evaluated once.  You could instead use single line [regexp.MatchString](https://pkg.go.dev/regexp#MatchString).  To avoid re-compiling the regex for every zone, the compile step needs to be done outside of the zone keys loop in `startUpZones`.

mdsahiloss: I think so, there is no benefit to compile the regex but if I use [regexp.MatchString](https://pkg.go.dev/regexp#MatchString), code will adjust in two steps itself like now, first step to handle return values from `regexp.MatchString()` like `bool` and `err` and then result will be returned in the next step so I think this is good to go If you feel any issue with it so let me know I will update it. 

and your suggestion to check root-domain separately looks good and will be easier for anyone to understand regex, I will update the regex again to make it better.

chrisohaver: > ... I think this is good to go ...

I think the compile step should be done outside of the zone keys loop in startUpZones.

MdSahil-oss: As you suggested earlier
> Each zone key in `zones` needs to be validated.

So, I think according to that suggestion It's perfect but If you want me to edit it I'll do it.

mdsahiloss: As we know `zones` is an Array of Strings, where right now only one string is present (That's root domain-name `.`), But in future when codebase will be deployed , There will be more than one domain-name in `zones` so for that we will need to
call `checkDomainName()` in a for-loop so that we can validate each `zone` of `zones` and now, we are doing the same if I call `checkDomainName()` outside of zone keys loop I'll have to create another for-loop in `startUpZones`.
Let me know if I'm wrong here or you have different vision for it. I'll enhance code as per your suggestion.

chrisohaver: @MdSahil-oss, 

> As we know `zones` is an Array of Strings ...

This is incorrect. `zones` is a map of `string` to `Config`, not an array.  `keys` is a [slice](https://www.godesignpatterns.com/2014/05/arrays-vs-slices.html) of domain names derived from the keys of the `zones` map.

> where right now only one string is present

That's not correct.  It can hold any number of strings.

It is inefficient to re-compile the regex for each domain string in the list.  It should be compiled once, and then the resulting compiled regex can be used to evaluate each domain.  It's not a major problem, since this only happens at startup, and the list of domain names will usually be short.  But IMO it's a good habit to be efficient in code when doing so is easy. Since it's easy to do here, it should be done.

>  I'll have to create another for-loop in startUpZones

No, you don't need to do that.

Please test your code using the Corefile below.  Put temporary print statements in the code (or use a debugger if you already have set up) to look at the value of `keys` in `startupZones`.

This is one Corefile, not 3.
```
# these zones are ok
example.com abc-123.com abc-123 abc.123-xyz a----3 {
 log
}

# these zones should display warnings
abc-^%$.example.com abc-.example.com 123-abc.example.com {
 log
}

# this zone is ok
. {
  log
}
```

chrisohaver (changes requested):
To summarize the issues that need to be addressed before this can merge:
1. The regex is invalid and does not compile successfully. This causes the subsequent code to panic.
2. The regex doesn't accurately match the RFC 1035 preferred syntax rules. In particular, 1035 has specific rules about permitted leading and trailing characters in each label that the regex does not enforce. 
3. The regex compilation should occur once, per server, and the resulting regex object re-used to match each zone in the server.
4. New test cases need to be added to better cover special cases such as disallowed leading/trailing characters, the root domain, and false positives on preferred-syntax substrings in a non-preferred-syntax domain.
5. Commit history needs to be cleaned up(remove unrelated commits from the PR). Rebasing with updated master should fix it up.

core/dnsserver/onstartup.go
mdsahiloss: only `abc-123` and `a----3` are failing from
> # these zones are ok
example.com abc-123.com abc-123 abc.123-xyz a----3 {
 log
}

and others are correct

chrisohaver: They fail because your regex requires at least two labels to be present.

SuperQ:
Maybe it would be better to pull in a more well maintained package like [govalidator](https://github.com/asaskevich/govalidator/).

chrisohaver:
> Maybe it would be better to pull in a more well maintained package like [govalidator](https://github.com/asaskevich/govalidator/).

We'd need to carefully evaluate govalidator's idea of what a valid DNS name is to make sure it aligns with what we need.  But IMO the RFC 1035 "preferred syntax" rules are simple and static. It's easy to represent in a short regex, so I think better to just do a regex in this case. 
e.g.  `^([A-Za-z]([A-Za-z0-9-]*[A-Za-z0-9])?\.)+$`

Note that here we are working with names that have already been validated as being valid DNS names with `dns.IsDomainName` (e.g. label length is already validated by this point).

This got me thinking though. This code should exclude reverse dns names from this validation because reverse names don't follow 1035 "preferred syntax" (they can have bare integers as labels).  i.e. if a domain name ends in `.arpa.`, the code should not check for preferred syntax.

SuperQ:
govalidator uses this regexp:

```
DNSName           string = `^([a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62}){1}(\.[a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62})*[\._]?$`
```

chrisohaver:
> govalidator uses this regexp:
> 
> ```
> DNSName           string = `^([a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62}){1}(\.[a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62})*[\._]?$`
> ```

OK, so it definitely doesn't match rfc1035's preferred syntax rules.

core/dnsserver/onstartup.go
mdsahiloss: @chrisohaver should I improve Regex for single label either ?

chrisohaver: A single label is permitted in the preferred syntax. So yes.

mdsahiloss: Okay... so I'm improving my Regex for single label too but don't forget there will be a dot `.` required as trailing characters for each `zone`. e.g
# these zones are ok
example.com. abc-123.com. abc-123. abc.123-xyz. a----3. {
 log
}

# these zones should display warnings
abc-^%$.example.com. abc-.example.com. 123-abc.example.com. {
 log
}

# this zone is ok
. {
  log
}

chrisohaver: Ok. Feel free to use the regex I [provided](https://github.com/coredns/coredns/pull/5414#issuecomment-1151114989) earlier. You’ll still need to check for the root domain separately, but I think its ok to do that.

core/dnsserver/onstartup_test.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	// Below conditions not checks for RFC 1035 "preferred syntax"
+	if regexpObj.MatchString("example-?&^%$.com.") {
+		t.Fatalf(`Expected false but received true for "example-?&^$.com."`)
+	}
+	if regexpObj.MatchString("-example.com.") {
+		t.Fatalf(`Expected true but received false for "-example.com."`)
+	}
+	if regexpObj.MatchString("example.com") {
```
chrisohaver: Message and logic mismatch.

core/dnsserver/onstartup_test.go
```diff
@@ -0,0 +1,46 @@
...
+	if !regexpObj.MatchString("an-example.com.") {
+		t.Fatalf(`Expected true but received false for "an-example.com."`)
+	}
+
+	// Below conditions not checks for RFC 1035 "preferred syntax"
+	if regexpObj.MatchString("example-?&^%$.com.") {
+		t.Fatalf(`Expected false but received true for "example-?&^$.com."`)
+	}
+	if regexpObj.MatchString("-example.com.") {
+		t.Fatalf(`Expected true but received false for "-example.com."`)
```
chrisohaver: Expected false, right?

core/dnsserver/onstartup.go
```diff
@@ -2,14 +2,22 @@ package dnsserver
 
 import (
 	"fmt"
+	"regexp"
 	"sort"
 )
 
+// compileRegexp() returns regexp-object for RFC1035 preferred syntax.
+func compileRegexp() *regexp.Regexp {
```
chrisohaver: Function name too generic: use something like regex1035PrefSyntax()

chrisohaver (changes requested):
Should check for reverse zones per earlier comment above. Add test case for that too. 

Please also provide a copy-paste sample of what the output actually looks like when running coredns with multiple zones where some get the warning.

core/dnsserver/onstartup_test.go
mdsahiloss: yeah! right

mdsahiloss:
@chrisohaver only `.arpa.` will specify that domain-name is reverse-domain-name ?

chrisohaver:
> @chrisohaver only `.arpa.` will specify that domain-name is reverse-domain-name ?

Use [`dnsutil.IsReverse(domain-name)`](https://github.com/coredns/coredns/blob/5b9b079dabc7f71463cea3f0c6a92f338935039d/plugin/pkg/dnsutil/reverse.go#L35) to check for it.

core/dnsserver/onstartup_test.go
```diff
@@ -43,4 +49,12 @@ func TestCheckDomainName(t *testing.T) {
...
 		t.Fatalf(`Expected false but received true for "123-abc.example.com."`)
 	}
+
+	// Below conditions checks for reverse-domain-name.
+	if dnsutil.IsReverse("example.com.") != 0 {
+		t.Fatalf(`Expected false for "reverse domain name" but received true for "example.com."`)
+	}
+	if dnsutil.IsReverse("com.example.") == 0 {
+		t.Fatalf(`Expected true for "reverse domain name" but received false for "com.example."`)
+	}
```
chrisohaver: In these tests, you should be testing `regexpObj.MatchString` ... not `dnsutil.IsReverse`.

core/dnsserver/onstartup_test.go
```diff
@@ -24,15 +30,15 @@ func TestCheckDomainName(t *testing.T) {
 		t.Fatalf(`Expected true but received false for "an-example.com."`)
 	}
 
-	// Below conditions not checks for RFC 1035 "preferred syntax"
+	// Below conditions not checks for RFC 1035 "preferred syntax".
```
chrisohaver: This comment is confusing, and I think unnecessary.

core/dnsserver/onstartup_test.go
```diff
@@ -1,16 +1,22 @@
 package dnsserver
 
-import "testing"
+import (
+	"testing"
+
+	"github.com/coredns/coredns/plugin/pkg/dnsutil"
+)
 
 func TestCheckDomainName(t *testing.T) {
```
chrisohaver: Function name should contain the function name being tested, e.g. `TestRegex1035PrefSyntax`

core/dnsserver/onstartup_test.go
```diff
@@ -1,16 +1,22 @@
...
 
 	// Note: compileRegexp() returns regexp-object for RFC1035 preferred syntax.
-	regexpObj := compileRegexp() // regexpObj initialised and declared with a regexp-object for RFC1035 preferred syntax.
+	regexpObj := regex1035PrefSyntax() // regexpObj initialised and declared with a regexp-object for RFC1035 preferred syntax.
 
-	// Below conditions checks for RFC 1035 "preferred syntax"
+	// Below condition checks for root domain-name.
```
chrisohaver: Comment not necessary. The test case is obvious.

core/dnsserver/onstartup_test.go
```diff
@@ -1,16 +1,22 @@
...
 	if !regexpObj.MatchString(".") {
 		t.Fatalf(`Expected true but received false for "." (root domain)`)
 	}
+
+	// Below conditions checks for RFC 1035 "preferred syntax".
```
chrisohaver: This comment is pretty much redundant.

Instead, group the tests into two categories, "positive cases" (where you expect to get a match) and "negative cases" (where you don't expect to get a match).

chrisohaver (changes requested):
Need to add tests for reverse zone queries.

For clarification on what reverse dns domains are: https://en.wikipedia.org/wiki/Reverse_DNS_lookup

mdsahiloss:
When only root-domain-name `.` is given in Zones then output:
~~~ txt
.:53 
CoreDNS-1.9.3 
linux/amd64, go1.17.10, 9d012a10-dirty
~~~
When multiple domain-names are given as input:
~~~ txt
package main

import (
	"fmt"
	"regexp"
)

func regex1035PrefSyntax() *regexp.Regexp {
	matchDomain, _ := regexp.Compile("(^([A-Za-z]([A-Za-z0-9-]*[A-Za-z0-9])(\\.([A-Za-z0-9-]+))*\\.)$)|^(\\.)$")
	return matchDomain
}

func main() {

	s := ""
	regexpObj := regex1035PrefSyntax()
	Zones := [9]string{".", "example.com.", "abc-123.com.", "abc-123.", "abc.123-xyz.", "a----3.", "abc-^%$.example.com.", "abc-.example.com.", "123-abc.example.com."}
	for _, zone := range Zones {
		if !regexpObj.MatchString(zone) {
			s += fmt.Sprintln("Warning: Domain " + zone + " does not follow RFC1035 preferred syntax")
		}
	}
	fmt.Println(s)
}

~~~
Output:
~~~ txt
Warning: Domain abc-^%$.example.com. does not follow RFC1035 preferred syntax
Warning: Domain abc-.example.com. does not follow RFC1035 preferred syntax
Warning: Domain 123-abc.example.com. does not follow RFC1035 preferred syntax
~~~

chrisohaver:
> When multiple domain-names are given as input:

What is the output of actually running coredns with a Corefile that has multiple valid and invalid domains?
i.e. like you did with the `.` zone example.... but with more than one zone.

core/dnsserver/onstartup.go
```diff
@@ -2,14 +2,23 @@ package dnsserver
...
 import (
 	"fmt"
+	"regexp"
 	"sort"
+	"github.com/coredns/coredns/plugin/pkg/dnsutil"
 )
 
+// regex1035PrefSyntax() returns regexp-object for RFC1035 preferred syntax.
+func regex1035PrefSyntax() *regexp.Regexp {
+	matchDomain, _ := regexp.Compile("(^([A-Za-z]([A-Za-z0-9-]*[A-Za-z0-9])(\\.([A-Za-z0-9-]+))*\\.)$)|^(\\.)$")
```
chrisohaver: I think it's cleaner to check for the root domain separately. This will simplify the regex (reducing nested parentheses), and make it easier to read.
Also if you use back tics to quote the string, you don't need to escape the `\.`s.  This will also make it easier to read.

core/dnsserver/onstartup_test.go
mdsahiloss: ohh! you mean I should remove these tests
~~~ txt
if dnsutil.IsReverse("example.com.") != 0 {
		t.Fatalf(`Expected false for "reverse domain name" but received true for "example.com."`)
	}
if dnsutil.IsReverse("com.example.") == 0 {
t.Fatalf(`Expected true for "reverse domain name" but received false for "com.example."`)
       }
~~~

core/dnsserver/onstartup.go
```diff
@@ -20,6 +31,9 @@ func startUpZones(protocol, addr string, zones map[string]*Config) string {
 	sort.Strings(keys)
 
 	for _, zone := range keys {
+		if dnsutil.IsReverse(zone) == 0 && !rootDomainRegex.MatchString(zone) && !otherDomainRegex.MatchString(zone) {
```
chrisohaver: It would be much simpler (and more efficient) to just do a string compare to check for the root zone instead of   using a regex. i.e.

```
		if zone != "." && dnsutil.IsReverse(zone) == 0 && !otherDomainRegex.MatchString(zone) {
```

core/dnsserver/onstartup_test.go
chrisohaver: Well not exactly.  Per the code comment, your intent here was to add test for reverse zone cases.  That is a good intent, and there should be tests for reverse zones.  However the test cases check `dnsutil.IsReverse`, not your function, also the zones tested are not reverse dns zones (reverse zones doesn't mean reverse order of domain labels).  So, reverse zone cases need to exist, but they need to be fixed.

chrisohaver (changes requested):
Since some of your logic is in the `startUpZones` function, you should also be testing that the output of `startUpZones` is as expected.  This should be done in a new test function `TestStartUpZones`. In your test when passing values to the function, the `*Config`s in the `zones` map can be `nil`, since they are not used by `startUpZones.

mdsahiloss:
@chrisohaver can you give me some examples for reverse domain name so that I can add them into Test cases

chrisohaver:
> @chrisohaver can you give me some examples for reverse domain name so that I can add them into Test cases

`0.10.in-addr.arpa.`
`1.0.0.2.ip6.arpa.`

mdsahiloss:
BTW, if I test like this :->
~~~ txt
// Positive test case
if dnsutil.IsReverse("an-example.com.") == 0 && !domainRegex.MatchString("an-example.com.") {
		t.Fatalf(`Expected true but received false for "an-example.com."`)
}

// Negative test case
if dnsutil.IsReverse("example-?&^%$.com.") == 0 && domainRegex.MatchString("example-?&^%$.com.") {
		t.Fatalf(`Expected false but received true for "example-?&^$.com."`)
}
~~~
I think we don't need to add additional test cases for `reverse-domain-name` .

chrisohaver:
> BTW, if I test like this :->
> 
> ```
> // Positive test case
> if dnsutil.IsReverse("an-example.com.") == 0 && !domainRegex.MatchString("an-example.com.") {
> 		t.Fatalf(`Expected true but received false for "an-example.com."`)
> }
> 
> // Negative test case
> if dnsutil.IsReverse("example-?&^%$.com.") == 0 && domainRegex.MatchString("example-?&^%$.com.") {
> 		t.Fatalf(`Expected false but received true for "example-?&^$.com."`)
> }
> ```
> 
> I think we don't need to add additional test cases for `reverse-domain-name` .

No, that does not add any additional coverage. The goal isn't to test the `dnsutil.IsReverse` function here.  The goal is to test the code you have added.  As I mentioned in the review comment, you'll need to add a new test function for `startUpZones`.

chrisohaver:
> you'll need to add a new test function for `startUpZones`.

Alternately, you can refactor all of your conditional logic into a new function, and test that new function instead of testing the regex directly as you are doing now. You can make the compiled 1035 regex a package variable to make this easier. e.g. 

```go
var regex1035PreferredSyntax, _ = regexp.Compile(`your regex here`)

func checkZoneSyntax(zone string) bool {
  return zone != "." && dnsutil.IsReverse(zone) == 0 && !regex1035PreferredSyntax.MatchString(zone)
}
```

chrisohaver:
> You can make the compiled 1035 regex a package variable to make this easier.

Actually, it's probably better, and simpler just to not use a pre-compiled regex, than to keep a compiled regex hanging around in memory.  The regex is only done once at start up for each zone, So it is not a big deal to re-parse the regex for each zone.  i.e. use [regexp.MatchString(pattern, s)](https://pkg.go.dev/regexp#MatchString)

core/dnsserver/onstartup.go
```diff
@@ -2,15 +2,23 @@ package dnsserver
...
 	"fmt"
+	"regexp"
 	"sort"
+
+	"github.com/coredns/coredns/plugin/pkg/dnsutil"
 )
 
+// checkZoneSyntax() checks whether the given string match 1035 Preferred Syntax or not.
+func checkZoneSyntax(zone string) bool {
+	regex1035PreferredSyntax, _ := regexp.MatchString(`^([A-Za-z][A-Za-z0-9-]*[A-Za-z](\.[A-Za-z][A-Za-z0-9-]*[A-Za-z])*\.)$`, zone)
```
chrisohaver: This regex looks more complex than it needs to be - and also looks wrong (doesn't match preferred 1035 syntax).  See the example regex I provided earlier.

Also, the regex match should not be executed if the zone is root or a reverse zone (since you can return true in those cases without needing to evaluate the regex).

core/dnsserver/onstartup_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	// Negative Test cases
+	if checkZoneSyntax("example-?&^%$.com.") {
+		t.Fatalf(`Expected false but received true for "example-?&^$.com."`)
+	}
+	if checkZoneSyntax("-example.com.") {
+		t.Fatalf(`Expected false but received false for "-example.com."`)
+	}
+	if checkZoneSyntax("example.com") {
+		t.Fatalf(`Expected false but received false for "example.com"`)
```
chrisohaver: "Expected false but received false" ?

core/dnsserver/onstartup_test.go
```diff
@@ -0,0 +1,45 @@
...
+	if !checkZoneSyntax("an-example.com.") {
+		t.Fatalf(`Expected true but received false for "an-example.com."`)
+	}
+
+	// Negative Test cases
+	if checkZoneSyntax("example-?&^%$.com.") {
+		t.Fatalf(`Expected false but received true for "example-?&^$.com."`)
+	}
+	if checkZoneSyntax("-example.com.") {
+		t.Fatalf(`Expected false but received false for "-example.com."`)
```
chrisohaver: "Expected false but received false" ?

core/dnsserver/onstartup_test.go
```diff
@@ -0,0 +1,45 @@
...
+	if !checkZoneSyntax(".") {
+		t.Fatalf(`Expected true but received false for "."`)
+	}
+	if !checkZoneSyntax("example.com.") {
+		t.Fatalf(`Expected true but received false for "example.com."`)
+	}
+	if !checkZoneSyntax("abc-123.com.") {
+		t.Fatalf(`Expected true but received false for "abc-123.com."`)
+	}
+	if !checkZoneSyntax("abc.123-xyz.") {
```
chrisohaver: incorrect expected result.  `abc.123-xyz.` should be _false_, because it has a label that starts with a non-letter.

chrisohaver (changes requested):
Still needs reverse zone test cases.

Also needs test cases for compliant and non compliant single label zones (e.g. `example.` == true, `example:.` == false), and single character labels (e.g. `a.example.com.` == true, `1.example.com.` == false)

Please carefully review the preferred syntax rules defined in rfc1035 and make sure that:
1. your test cases expected values adhere to that spec (do this first)
2. your regex accurately match that spec

It's important to get the tests right first - otherwise you may end up tweaking your function to match bad tests (I think this has started to happen, re: complexity added to regex to force a bad test pass).

core/dnsserver/onstartup.go
mdsahiloss: I think you are right here, I need to improve Regex also for trailing digits as per 1035 syntax.

mdsahiloss: the regex match should not being executed if the zone is root or a reverse zone.

mdsahiloss:
@chrisohaver Now I think, Regex is completely following 1035 Systax.
But there is a limitation I provided to Regex to keep it simple that each label must be at least 2 or more characters long and ends with a  trailing dot `.` e.g. 
`an.` will be `true`
but `a.` will be `false`

chrisohaver:
> @chrisohaver Now I think, Regex is completely following 1035 Systax.
> But there is a limitation I provided to Regex to keep it simple that each label must be at least 2 or more characters long and ends with a trailing dot `.` e.g.
> `an.` will be `true`
> but `a.` will be `false`

That’s not right. Can you just use the regex I gave earlier?

mdsahiloss:
@chrisohaver can you send me that Regex again? I think that is no longer on this conversation page.

chrisohaver:
> @chrisohaver can you send me that Regex again? I think that is no longer on this conversation page.

https://github.com/coredns/coredns/pull/5414#issuecomment-1151114989

mdsahiloss:
I think, ` ^([A-Za-z]([A-Za-z0-9-]*[A-Za-z0-9])?\.)+$` is good to use and will work fine for each label.

chrisohaver (changes requested):
Please address all feedback from prior review(s).

* Regex still longer and more complex than it should be.
* Reverse zone test cases still missing.
* Regex is still evaluated in root and reverse cases.
* etc

core/dnsserver/onstartup.go
```diff
@@ -2,24 +2,39 @@ package dnsserver
...
+
+	"github.com/coredns/coredns/plugin/pkg/dnsutil"
 )
 
+// checkZoneSyntax() checks whether the given string match 1035 Preferred Syntax or not.
+func checkZoneSyntax(zone string) bool {
+	if(zone == "." || dnsutil.IsReverse(zone) != 0){
+		return true
+	}
+	regex1035PreferredSyntax, _ := regexp.MatchString(`^([A-Za-z]([A-Za-z0-9-]*[A-Za-z0-9])?(\.[A-Za-z]([A-Za-z0-9-]*[A-Za-z0-9])?)*\.)$`, zone)
```
chrisohaver: Your regex will incorrectly match a zone starting with `.` such as `.example.`.
Please use the regex I have suggested. i.e.

```suggestion
	regex1035PreferredSyntax, _ := regexp.MatchString(`^([A-Za-z]([A-Za-z0-9-]*[A-Za-z0-9])?\.)+$`, zone)
```

core/dnsserver/onstartup_test.go
```diff
@@ -0,0 +1,70 @@
...
+	if checkZoneSyntax("abc-%$.example.com.") {
+		t.Fatalf(`Expected false but received true for "abc-^$.example.com."`)
+	}
+	if checkZoneSyntax("abc-.example.com.") {
+		t.Fatalf(`Expected false but received true for "abc-.example.com."`)
+	}
+	if checkZoneSyntax("123-abc.example.com.") {
+		t.Fatalf(`Expected false but received true for "123-abc.example.com."`)
+	}
+}
```
chrisohaver, line 40:
```suggestion
	if checkZoneSyntax(".example.com.") {
 		t.Fatal(`Expected false but received true for ".example.com."`)
 	}
}
```

core/dnsserver/onstartup_test.go
```diff
@@ -0,0 +1,70 @@
+package dnsserver
+
+import (
+	"testing"
+)
+
+func TestRegex1035PrefSyntax(t *testing.T) {
+
+	// Positive Test cases
+	if !checkZoneSyntax(".") {
```
chrisohaver: For all (these) test cases, fail with Error instead of Fatal, so the test will evaluate all cases without aborting on first failure.

Also, you should only use the format version of these functions when you are doing formatting/ value replacement.

Your test zone strings do not always match the zones printed in the error messages ( I suspect because in a couple of places you stripped the `%` from the error message to make the warnings about format strings go away, which was due to your using `Fatalf` instead of `Fatal`).

The copy-paste approach to the test is verbose prone to typo errors.  You should refactor these tests to a loop that checks a series of test cases defined in a slice.  This would reduce copy-paste code replication, making the code shorter, and also less prone to this kind of error. e.g. 

```go
testCases := []struct{
  zone string
  expected bool
}{
  {zone: ".", expected: true},
  {zone: "example.com.", expected: true},
  {zone: "example.", expected: true},
  // etc, add remaining positive and negative test cases here
}

for _, testCase := range testCases {
  if checkZoneSyntax(testCase.zone) != testCase.expected {
    t.Errorf("Expected %v for %q", testCase.expected, testCase.zone)
  }
}
```

(Please note: I just typed the above code example straight into the comment, I didn't check it for correctness, so there may be syntax issues, such as the one I just fixed.)

chrisohaver (changes requested):

core/dnsserver/onstartup.go
```diff
@@ -2,24 +2,39 @@ package dnsserver
...
+
 	for k := range zones {
 		keys[i] = k
 		i++
 	}
 	sort.Strings(keys)
 
 	for _, zone := range keys {
+		if !checkZoneSyntax(zone) {
+			s += fmt.Sprintln("Warning: Domain " + zone + " does not follow RFC1035 preferred syntax")
```
chrisohaver:
```suggestion
			s += fmt.Sprintf("Warning: Domain %q does not follow RFC1035 preferred syntax\n", zone)
```

core/dnsserver/onstartup.go
```diff
@@ -2,24 +2,39 @@ package dnsserver
 
 import (
 	"fmt"
+	"regexp"
 	"sort"
+
+	"github.com/coredns/coredns/plugin/pkg/dnsutil"
 )
 
+// checkZoneSyntax() checks whether the given string match 1035 Preferred Syntax or not.
```
chrisohaver, line 11:
```suggestion
// checkZoneSyntax() checks whether the given string match 1035 Preferred Syntax or not.
// The root zone, and all reverse zones always return true even though they technically don't meet 1035 Preferred Syntax 
```

chrisohaver (changes requested):
Almost there!

chrisohaver:
Please sign the commits, so the DCO check passes. https://github.com/coredns/coredns/pull/5414/checks?check_run_id=7212082925

PR #5425, opened 2022-06-05
Log the literal value, not a format string

alindeman:

The `log` plugin currently passes the replaced log string to `Infof` which expects a format string.

The query question or an IPv6 link local address could contain a `%` which the Go formatter would interpret as a verb, resulting in a log message such as:

```
Jun 05 16:41:03 router coredns[10786]: [INFO] [fe80::abcd:abcd:abcd:abcd%!l(MISSING)an0]:54939 - 57566 "A IN foo.%!s(MISSING).example. udp 43 false 4096" NXDOMAIN qr,aa,rd,ra 107 0.010446858s
```

N/A

N/A

No


chrisohaver (approved):
Thanks! LGTM

PR #5435, opened 2022-06-13
plugin/k8s_external: Add support for PTR requests

### LINKED ISSUE #5422: kubernetes plugin answers reverse lookups for External IPs
**What happened**:

When _kubernetes_ plugin is in use, and _k8s_external_ is not in use:
For reverse lookups of an IP that is an External IP of a Kubernetes Service, CoreDNS  will answer with that service.

When _k8s_external_ is not in use: CoreDNS should only answer with a service who's Cluster IP or Endpoint IP matches the Service.  In a default K8s coredns deployment, the request should fallthrough to the next plugin, and be forwarded upstream.

When _k8s_external_ is in use: One might expect CoreDNS to answer, but with the service name being in the _k8s_external_ zone, (not in the _kubernetes_ zone as happens currently). Although _k8s_external_ does not currently support PTR requests (documented in readme). 

Create a k8s cluster (with CoreDNS as dns discovery service).
Create a service with an External IP. 
Do a reverse lookup for that IP.

Introduced when _k8s_external_ was added in https://github.com/coredns/coredns/pull/2379
Issue raised in https://github.com/kubernetes/dns/issues/518

The root cause is that a single reverse index is maintained, and it holds both Cluster IPs and External IPs.

A workaround is to focus the reverse zones handled by the _kubernetes_ plugin such that it excludes the External IP range.  (Since the cluster Endpoint IP range cannot be programmatically determined, the reverse lookup zones default handle all reverse lookups everything.)

A complete solution is to split the reverse index into two indexes (one for Cluster IP and the other for External IPs), and add PTR query handling to the _k8s_external_ plugin.   _kubernetes_ would use the Cluster IP index, and _k8s_external_ would use the External IPs index.  Both _k8s_external_ and _kubernetes_ should answer PTR requests and fallthrough when no match is found. _k8s_external_ currently does not support PTR requests and does not support fallthrough.

A partial solution is to just remove External IPs from the reverse index and leave PTR unimplemented for _k8s_external_.

Regarding backward incompatibility: I think it's unlikely that a user depends on the current behavior, especially since the answer currently provided is non-reciprocal.  i.e. Looking up A/AAAA for the name returned by the External IP PTR request will result in the Service's Cluster IP (or EndpointIPs for headless Service).

chrisohaver:

Fixes issue #5422 by:
* Excluding External IP addresses from being added to the existing _kubernetes'_ plugin IP->Service index
* Adding support for PTR requests on External IPs of Services to the _k8s_external_ plugin

Closes #5422

### DIFF
--- a/plugin/k8s_external/README.md
+++ b/plugin/k8s_external/README.md
@@ -10,7 +10,7 @@ This plugin allows an additional zone to resolve the external IP address(es) of 
 service. This plugin is only useful if the *kubernetes* plugin is also loaded.
 
 The plugin uses an external zone to resolve in-cluster IP addresses. It only handles queries for A,
-AAAA and SRV records; all others result in NODATA responses. To make it a proper DNS zone, it handles
+AAAA, SRV, and PTR records; all others result in NODATA responses. To make it a proper DNS zone, it handles
 SOA and NS queries for the apex of the zone.
 
 By default the apex of the zone will look like the following (assuming the zone used is `example.org`):
@@ -101,6 +101,3 @@ zone transfers.  Notifies are not supported.
 For some background see [resolve external IP address](https://github.com/kubernetes/dns/issues/242).
 And [A records for services with Load Balancer IP](https://github.com/coredns/coredns/issues/1851).
 
-# Bugs
-
-PTR queries for the reverse zone is not supported.
--- a/plugin/k8s_external/external.go
+++ b/plugin/k8s_external/external.go
@@ -71,6 +71,20 @@ func (e *External) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 		return plugin.NextOrFailure(e.Name(), e.Next, ctx, w, r)
 	}
 
+	if state.QType() == dns.TypePTR {
+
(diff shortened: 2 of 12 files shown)


chrisohaver:
hmm - I think I need to handle the case where multiple services share a common external IP address.
While it isn't possible for more than one Service to share a Cluster IP or more than one Pod to share a Pod IP,
it's possible that two Services could share an External IP.

[Added in 1543d25]

chrisohaver:
> While it isn't possible for more than one Service to share a Cluster IP or more than one Pod to share a Pod IP,
> it's possible that two Services could share an External IP.

FWIW, it's also possible for multiple headless services to share a pod IP, which _kubernetes_ already handles.

chrisohaver:
Also need to add fallthrough option to k8s_external.

chrisohaver:
> Also need to add fallthrough option to k8s_external.

For now, let's assume that users will be able to tailor the zone CIDRs or reverse zones to fit their external IP ranges.
If this turns out to be a bad assumption (e.g. unpredictable or dynamic elastic ranges), we can add the fallthrough option to _k8s_external_.

Issue #5437, opened 2022-06-14
Enhancement: add k8s_event plugin


xh4n3:
**What would you like to be added**
k8s_event, a plugin that would send events to k8s.

**Why is this needed**
- Noticeable: Many CoreDNS users don’t setup CoreDNS metrics alerting before they had DNS breakdowns. If we push events like upstream DNS server timeout, Corefile reload failure to k8s, they should have notice some potential risk in the first place.
- Lasting & Precise: Metrics do not always provide enough information for finding root causes for DNS breakdowns. Container log are mostly mixed with access logs, and prone to get rotated. K8s events can provide slightly more lasting and more precise information.

**Consideration & Goals**
- Customizable:
  - event type: what kind (or level) of event should be reported
  - target k8s: which k8s apiserver endpoint to report events to
    - or are we gonna support other types of eventing in the future, e.g. cloud events?
- Less code: we better not to change every plugin's code to intergrate with this eventing plugin
- Avoid overloading k8s: include events from control plane, exclude events from each dns request

**Design** 
As Chris suggested, CoreDNS's logger package is responsible for the logging of all plugins, and is suitable for such a feature to be implement in. Besides, the k8s_event plugin could be the user interface to change the behaviors of this feature. Let me demonstrate below:

![image](https://user-images.githubusercontent.com/4469022/173582029-19afcc95-6b05-4045-951b-90745af081fa.png)

What k8s_event plugin does:
1. initiate a event recorder with k8s apiserver
2. wrap event recording logic in callback functions
3. call the logger's SetCallback with its callback functions

```go
func onStart() {
    kubeClient := getKubeClient()
    ref := getRef()
    recorder := getRecorder(kubeClient)
    log.SetCallback(logFunc, logfFunc)
}

func logfFunc(ref *corev1.ObjectReference, record record.EventRecorder, classes []Class) func(level, format string, v ...interface{}) {
	return func(level, format string, v ...interface{}) {
            if matchClass(level, classes) {
                record.Eventf(ref, corev1.EventTypeWarning, levelToReason(level), fmt.Sprintf(format, v))
            }
	}
}
```

What logger package does:
1. initiate a singleton variable to store callback functions, allow k8s_event to set
2. when its printer functions get called, it checks whether callback functions have been set
  a. if set, call the callback functions first, then call the Golang's log to print
  b. if not set, simply call the Golang's log to print
```go
// this is almost identical to the D singleton for Debug
var E = &e{}

type e struct {
[...]

PR #5451, opened 2022-06-22
core: add log listeners for k8s_event plugin

xh4n3:
I proposed a plugin k8s_event earlier in the issue https://github.com/coredns/coredns/issues/5437 .

image

In this pull request, I added the method `LoadInterceptor` to let event adaptors like k8s_event register themselves.

And I pushed a draft version of k8s_event to a personal repository https://github.com/xh4n3/k8s_event .

It may not be the final version, I'd like to hear some feedback first.

If we're all satisfied, maybe I can push the k8s_event to some repository under the CoreDNS organization?

https://github.com/coredns/coredns/issues/5437

This pr is not facing the end-users, so no doc change.
After this gets merged, I will create another plugin doc page on the coredns.io website.

No.


chrisohaver:
Thanks!

It appears that only one plugin can intercept log messages, the last one taking precedence. Can you add support for multiple  log intercepting plugins?

chrisohaver:
I think of this feature as implemented as more of a "mirror", since the message is still being logged by the log package, and also a copy of the log message is being sent to another function.

The name "interceptor" kind of implies that the log message would sent to an alternate function _instead_ of being logged by log.

(IOW: suggesting we rename to `mirror`, or something other than `interceptor`)

Actually: `listener` works better than `mirror`

yongtang:
I will take a detailed look during the weekend but I like the direction this plugin is heading. We can follow up on where to host this plug-in (in coredns org or repo).

xh4n3:
@chrisohaver Nice suggestions, I've pushed a new commit addressing them.

The new diagram would be like this:
image

BTW, here's an example of the reported events.
```bash
$ kubectl get ev -A -w
NAMESPACE   LAST SEEN   TYPE      REASON           OBJECT              MESSAGE
# when CoreDNS started
default     1s          Normal    CoreDNSInfo      namespace/default   plugin/reload: Running configuration SHA512 = 8d4db1d9149964e0cfda9274ea7500bed395a7d518ca3ef1b7df9980908fd447d9eafed6d017cf89f96ac4b406585b68aece7c108a30c5660f69532d1a74948f
# when i tried to dig with a unreachable upstream
default     0s          Warning   CoreDNSError     namespace/default   plugin/errors: 2 www.ab.com. A: read udp 30.225.16.62:57863->114.114.114.118:53: i/o timeout
default     1s          Warning   CoreDNSError     namespace/default   plugin/errors: 2 www.ab.com. A: read udp 30.225.16.62:52029->114.114.114.118:53: i/o timeout
# when i added the proxy plugin
default     1s          Warning   CoreDNSError     namespace/default   plugin/reload: Corefile changed but reload failed: starting with listener file descriptors: Corefile:16 - Error during parsing: Unknown directive 'proxy'
# when i removed proxy plugin
default     1s          Normal    CoreDNSInfo      namespace/default   plugin/reload: Running configuration SHA512 = 8d4db1d9149964e0cfda9274ea7500bed395a7d518ca3ef1b7df9980908fd447d9eafed6d017cf89f96ac4b406585b68aece7c108a30c5660f69532d1a74948f
default     1s          Warning   CoreDNSWarning   namespace/default   plugin/k8s_event: COREDNS_NAMESPACE or COREDNS_POD_NAME is not set in environment variables, reporting events to default namespace
```

chrisohaver (commented):
Thanks, looks good.

Thinking about if we need an RWMutex on `listeners`:  I don't _think_ so, since plugins do their setups serially, and I don't think any plugins try to write to the log in a go routine during setup (i.e. concurrently with other plugins setting up). And plugins should only register a listener during setup, before the server starts.

chrisohaver:
> And plugins should only register a listener during setup, before the server starts.

Although, I see in https://github.com/xh4n3/k8s_event/pull/1, that the listener registration is deferred to onStartup, not during plugin setup ... meaning that it could try to register a listener when other plugins are writing to the log.

So perhaps we _should_ add RWMutex to `listeners`.

xh4n3:
@chrisohaver Thanks, the RWMutex has been added.

@yongtang Any comments on this PR? Please let me know, thanks 😃

chrisohaver (approved):
Thanks! LGTM

PR #5452, opened 2022-06-22
plugin/trace: make zipkin and datadog reporters log errors using CoreDNS logger

Tantalor93:
### 1. Why is this pull request needed and what does it do?
This PR improves `trace` plugin in a way that errors logged by zipkin/datadog span reporters are logged in a CoreDNS format and not default format of the logger from the std library

example of logs before the change

let's have Corefile
```
.:53 {
  log
 trace zipkin 127.0.0.1:9411 {
   service coredns
 }
  forward . 8.8.8.8
}
```

and when we do some DNS query, the CoreDNS will log:
```
2022/06/22 21:53:18 failed to send the request: Post "http://127.0.0.1:9411/api/v2/spans": dial tcp 127.0.0.1:9411: connect: connection refused
```

this PR improves the logging in a way that it is logged as:
```
[WARNING] plugin/trace: failed to send the request: Post "http://127.0.0.1:9411/api/v2/spans": dial tcp 127.0.0.1:9411: connect: connection refused
```
This way the logs produced by CoreDNS are in a more unified format and better parsable by other log parsers

### 2. Which issues (if any) are related?
no issue related

### 3. Which documentation changes (if any) need to be made?
none needed

### 4. Does this introduce a backward incompatible change or deprecation?
no

PR #5457, opened 2022-06-23
log DoH HTTP server error logs in CoreDNS format

Tantalor93:

This PR changes default logger of  the HTTP server used for processing DoH requests to log error logs in a CoreDNS format (similar change to #5452), all errors are logged on DEBUG level as these errors are in most cases related to the client issues, but sometimes for debugging they can be useful.

example of behaviour before and after this change:
let's have an Corefile
```
https://.:8080 {
  log
  forward . 8.8.8.8
  debug
}
```

and when we do DoH query such as this
```
curl -k -H 'accept: application/dns-message' 'http://127.0.0.1:8080/dns-query?dns=;180BAAABAAAAAAAAA3d3dwdleGFtcGxlA2NvbQAAAQAB' -i
```

it will log
```
http: URL query contains semicolon, which is no longer a supported separator; parts of the query may be stripped when parsed; see golang.org/issue/25192
```

but this PR changes it that it will be logged as 
```
[DEBUG] http: URL query contains semicolon, which is no longer a supported separator; parts of the query may be stripped when parsed; see golang.org/issue/25192
```

none

none

no

### DIFF
--- a/core/dnsserver/server_https.go
+++ b/core/dnsserver/server_https.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	stdlog "log"
 	"net"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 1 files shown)


chrisohaver (approved):
lgtm thanks

PR #5459, opened 2022-06-25
Fix a crash in rewrite plugin when rule type is missing

yongtang:

This PR tries to fix a crash in rewrite plugin when rule type is missing.

n/a

n/a
n/a


greenpau (approved):
LGTM

PR #5491, opened 2022-07-06
Support deprecated "service.alpha.kubernetes.io/tolerate-unready-endpoints" annotation

zuzzas:

A deprecated annotation `service.alpha.kubernetes.io/tolerate-unready-endpoints`
With the 1.8.5 release of coredns an EndpointSlice support was introduced. EndpointSlices do not respect this deprecated annotation. The condition.ready field is not set to true if there is an annotation. Endpoint controller respects it.

Once we've updated coredns to >1.8.5, a lot of stuff stopped working. As it turns out, a lot of older Charts use this annotation, and migrating them all at once is quite prohibitive.

Is it possible for coredns to support this annotation?

None.

None.

It does not.

### DIFF
--- a/plugin/kubernetes/handler_test.go
+++ b/plugin/kubernetes/handler_test.go
@@ -684,6 +684,7 @@ var epsIndex = map[string][]*object.Endpoints{
 		Name:      "kubedns",
 		Namespace: "kube-system",
 		Index:     object.EndpointsKey("kubedns", "kube-system"),
+		Ready:     true,
 	}},
 	"svc1.testns": {{
 		Subsets: []object.EndpointSubset{
@@ -699,6 +700,7 @@ var epsIndex = map[string][]*object.Endpoints{
 		Name:      "svc1-slice1",
 		Namespace: "testns",
 		Index:     object.EndpointsKey("svc1", "testns"),
+		Ready:     true,
 	}},
 	"svcempty.testns": {{
 		Subsets: []object.EndpointSubset{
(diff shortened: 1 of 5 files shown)


chrisohaver (changes requested):
Ignoring deprecated annotations by default seems like the prudent approach.
Honoring deprecated annotations is problematic (technically a bug), and encourage use of deprecated configurations.

It potentially introduces backward incompatible behavior. If someone unknowingly has stray deprecated alpha notation present, and they are content with the current behavior of the annotations being ignored, then this change introduces a backward incompatible change.

I'm also hesitant to support this as an option without exploring other possible solutions.  If added as an option, it would essentially be a short-lived option that exists to support laggard helm charts. Once those helm chars are fixed, it would become option clutter.

zuzzas:
Indeed. That's why I was hesitant to open this PR in the first place.

Furthermore, it's not supported in the main branch anymore: https://github.com/kubernetes/kubernetes/pull/108020

zuzzas:
But the question stands: is it possible to merge this until more charts migrate to the proper Service .spec field?

Have you taken a look at the code? Even if it's not possible to merge, would it work properly on a patched coredns?

chrisohaver:
> is it possible to merge this until more charts migrate to the proper Service .spec field?

Definitely not as default behavior. Furthermore, I'm hesitant to support this as _optional_ behavior it would soon become useless option clutter once the laggard helm charts are fixed.  Its existance might also discourage/slow fixing the helm charts.

> Have you taken a look at the code? Even if it's not possible to merge, would it work properly on a patched coredns?

Yes I have, It looks like it would work.  

Another possible solution could be to use a mutating admission controller on Services to properly migrate this when Services are created.

zuzzas:
>Another possible solution could be to use a mutating admission controller on Services to properly migrate this when Services are created.

Nice! Thank you. That's a better migration path that I did not think of. It does not apply to existing Services, mind you, but looks great going forward.

PR #5505, opened 2022-07-11
plugin/k8s_external: Resolve headless services

TomasKohout:
Signed-off-by: Tomas Kohout <tomas.kohout1995@gmail.com>

We have a need for resolving headless services with the use of the k8s_external plugin. Our users use this feature with clients for MongoDB etc. because in those cases LoadBalancer IP is behaving strangely. 

I havn't found any.

Documentation related to the k8s_external plugin.

I don't think that this will introduce backward incompatibility because it's opt in feature.

## Call for aid

I'm strugling with the AXFR transfer test. Is there somebody who would lend me a hand please? :slightly_smiling_face:

### DIFF
--- a/plugin/k8s_external/transfer.go
+++ b/plugin/k8s_external/transfer.go
@@ -5,9 +5,8 @@ import (
 	"strings"
 
 	"github.com/coredns/coredns/plugin"
-	"github.com/coredns/coredns/plugin/kubernetes"
-
 	"github.com/coredns/coredns/plugin/etcd/msg"
+	"github.com/coredns/coredns/plugin/kubernetes"
 	"github.com/coredns/coredns/plugin/transfer"
 	"github.com/coredns/coredns/request"
 
--- a/plugin/k8s_external/transfer.go
+++ b/plugin/k8s_external/transfer.go
@@ -50,7 +50,7 @@ func (e *External) Transfer(zone string, serial uint32) (<-chan []dns.RR, error)
 
 		svcs := e.externalServicesFunc(zone, e.headless)
 		srvSeen := make(map[string]struct{})
-		// svcsGroup := make(map[string][]msg.Service)
+
 		for i := range svcs {
 			name := msg.Domain(svcs[i].Key)
 
--- a/plugin/k8s_external/transfer.go
+++ b/plugin/k8s_external/transfer.go
@@ -40,7 +40,7 @@ func (e *External) Transfer(zone string, serial uint32) (<-chan []dns.RR, error)
 		ch <- []dns.RR{&dns.NS{Hdr: nsHdr, Ns: nsName}}
 
 		// Add Nameserver A/AAAA records
-		nsRecords := e.externalAddrFunc(state)
+		nsRecords := e.externalAddrFunc(state, e.headless)
 		for i := range nsRecords {
 			// externalAddrFunc returns incomplete header names, correct here
 			nsRecords[i].Header().Name = nsName
@@ -48,10 +48,12 @@ func (e *External) Transfer(zone string, serial uint32) (<-chan []dns.RR, error)
 			ch <- []dns.RR{nsRecords[i]}
 		}
 
-		svcs := e.externalServicesFunc(zone)
+		svcs := e.externalServicesFunc(zone, e.headless)
 		srvSeen := make(map[string]struct{})
+		// svcsGroup := make(map[string][]msg.Service)
 		for i := range svcs {
 			name := msg.Domain(svcs[i].Key)
+
 			if svcs[i].TargetStrip == 0 {
 				// Add Service A/AAAA records
 				s := request.Request{Req: &dns.Msg{Question: []dns.Question{{Name: name}}}}
--- a/plugin/k8s_external/transfer.go
+++ b/plugin/k8s_external/transfer.go
@@ -2,8 +2,11 @@ package external
 
 import (
 	"context"
+	"strings"
 
 	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/kubernetes"
+
 	"github.com/coredns/coredns/plugin/etcd/msg"
 	"github.com/coredns/coredns/plugin/transfer"
 	"github.com/coredns/coredns/request"
@@ -48,7 +51,7 @@ func (e *External) Transfer(zone string, serial uint32) (<-chan []dns.RR, error)
 			ch <- []dns.RR{nsRecords[i]}
 		}
 
-		svcs := e.externalServicesFunc(zone, e.headless)
+		svcs, headlessSvcs := e.externalServicesFunc(zone, e.headless)
 		srvSeen := make(map[string]struct{})
 
 		for i := range svcs {
@@ -81,6 +84,55 @@ func (e *External) Transfer(zone string, serial uint32) (<-chan []dns.RR, error)
 				if !nameSeen(srvSeen, srv) {
 					ch <- []dns.RR{srv}
 				}
+			}
+		}
+		for key, svcs := range headlessSvcs {
+			// we have to strip the leading key because it's either port.protocol or endpoint
+			name := msg.Domain(key[:strings.LastIndex(key, "/")])
+			switchKey := key[strings.LastIndex(key, "/")+1:]
+			switch switchKey {
+			case kubernetes.Endpoint:
+				// headless.namespace.example.com records
+				s := request.Request{Req: &dns.Msg{Question: []dns.Question{{Name: name}}}}
+				as, _ := e.a(ctx, svcs, s)
+				if len(as) > 0 {
+					ch <- as
+				}
+				aaaas, _ := e.aaaa(ctx, svcs, s)
+				if len(aaaas) > 0 {
+					ch <- aaaas
+				}
+				// Add bare SRV record, ensuring uniqueness
+				recs, _ := e.srv(ctx, svcs, s)
+				ch <- recs
+				for _, srv := range recs {
+					ch <- []dns.RR{srv}
+				}
+
+				for i := range svcs {
+					// endpoint.headless.namespace.example.com record
+					s := request.Request{Req: &dns.Msg{Question: []dns.Question{{Name: msg.Domain(svcs[i].Key)}}}}
+
+					as, _ := e.a(ctx, []msg.Service{svcs[i]}, s)
+					if len(as) > 0 {
+						ch <- as
+					}
+					aaaas, _ := e.aaaa(ctx, []msg.Service{svcs[i]}, s)
+					if len(aaaas) > 0 {
+						ch <- aaaas
+					}
+					// Add bare SRV record, ensuring uniqueness
+					recs, _ := e.srv(ctx, []msg.Service{svcs[i]}, s)
+					ch <- recs
+					for _, srv := range recs {
+						ch <- []dns.RR{srv}
+					}
+				}
+
+			case kubernetes.PortProtocol:
+				s := request.Request{Req: &dns.Msg{Question: []dns.Question{{Name: name}}}}
+				recs, _ := e.srv(ctx, svcs, s)
+				ch <- recs
 			}
 		}
 		ch <- []dns.RR{soa}
--- a/plugin/k8s_external/transfer_test.go
+++ b/plugin/k8s_external/transfer_test.go
@@ -27,6 +27,7 @@ func TestTransferAXFR(t *testing.T) {
 	k.APIConn = &external{}
 
 	e := New()
+	e.headless = true
 	e.Zones = []string{"example.com."}
 	e.externalFunc = k.External
 	e.externalAddrFunc = externalAddress  // internal test function
@@ -64,6 +65,7 @@ func TestTransferAXFR(t *testing.T) {
 			if ans.Header().Rrtype == dns.TypePTR {
 				continue
 			}
+
 			expect = append(expect, ans)
 		}
 	}
@@ -92,6 +94,7 @@ func TestTransferIXFR(t *testing.T) {
 
 	e := New()
 	e.Zones = []string{"example.com."}
+	e.headless = true
 	e.externalFunc = k.External
 	e.externalAddrFunc = externalAddress  // internal test function
 	e.externalSerialFunc = externalSerial // internal test function
@@ -136,7 +139,9 @@ func difference(testRRs []dns.RR, gotRRs []dns.RR) []dns.RR {
 	}
 
 	foundRRs := []dns.RR{}
+	cmp := map[string]struct{}{}
 	for _, rr := range gotRRs {
+		cmp[rr.String()] = struct{}{}
 		if _, ok := expectedRRs[rr.String()]; !ok {
 			foundRRs = append(foundRRs, rr)
 		}
--- a/plugin/k8s_external/transfer_test.go
+++ b/plugin/k8s_external/transfer_test.go
@@ -139,9 +139,7 @@ func difference(testRRs []dns.RR, gotRRs []dns.RR) []dns.RR {
 	}
 
 	foundRRs := []dns.RR{}
-	cmp := map[string]struct{}{}
 	for _, rr := range gotRRs {
-		cmp[rr.String()] = struct{}{}
 		if _, ok := expectedRRs[rr.String()]; !ok {
 			foundRRs = append(foundRRs, rr)
 		}
--- a/plugin/kubernetes/external.go
+++ b/plugin/kubernetes/external.go
@@ -12,8 +12,8 @@ import (
 )
 
 // External implements the ExternalFunc call from the external plugin.
-// It returns any services matching in the services' ExternalIPs.
-func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
+// It returns any services matching in the services' ExternalIPs and if enabled, headless endpoints..
+func (k *Kubernetes) External(state request.Request, headless bool) ([]msg.Service, int) {
 	if state.QType() == dns.TypePTR {
 		ip := dnsutil.ExtractAddressFromReverse(state.Name())
 		if ip != "" {
@@ -36,7 +36,7 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
 	// We are dealing with a fairly normal domain name here, but we still need to have the service
 	// and the namespace:
 	// service.namespace.<base>
-	var port, protocol string
+	var port, protocol, endpoint string
 	namespace := segs[last]
 	if !k.namespaceExposed(namespace) {
 		return nil, dns.RcodeNameError
@@ -49,7 +49,10 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
 
 	service := segs[last]
 	last--
-	if last == 1 {
+	if last == 0 {
+		endpoint = stripUnderscore(segs[last])
+		last--
+	} else if last == 1 {
 		protocol = stripUnderscore(segs[last])
 		port = stripUnderscore(segs[last-1])
 		last -= 2
@@ -60,8 +63,15 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
 		return nil, dns.RcodeNameError
 	}
 
+	var (
+		endpointsListFunc func() []*object.Endpoints
+		endpointsList     []*object.Endpoints
+		serviceList       []*object.Service
+	)
+
 	idx := object.ServiceKey(service, namespace)
-	serviceList := k.APIConn.SvcIndex(idx)
+	serviceList = k.APIConn.SvcIndex(idx)
+	endpointsListFunc = func() []*object.Endpoints { return k.APIConn.EpIndex(idx) }
 
 	services := []msg.Service{}
 	zonePath := msg.Path(state.Zone, coredns)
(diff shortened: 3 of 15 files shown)


chrisohaver:
Can you describe the use case in more detail with some examples?  There may be another way to solve what you want to do with existing plugins.

TomasKohout:
@chrisohaver sure thing! :) 

We have our internal dns server with zone eg. example.com. Then we have subdomains in example.com that are forwarded to our CoreDNS pods in k8s clusters so eg. k8s-dev.example.com is forwarded to the k8s-dev cluster. Plain External plugin is not enough for us because it resolve only LoadBalancer and ExternalIPs, but users requested to resolve also pod IPs (which are routable in our network) so I've added this feature to CoreDNS v1.8 which was pretty straightforward, but things got complicated in 1.9. :slightly_smiling_face:

chrisohaver:
Thanks, the change seems thorough, covering NS record case, and zone transfers. But it does add quite a bit of code spread across two plugins, so I want to explore alternate implementations that may be less impactful.

If  ...
* _k8s_external_ supported the `fallthrough` option (e.g. like the _kubernetes_ plugin)
* and you re-order the position of the _rewrtite_ plugin in `plugin.cfg` to sit between _k8s_external_ and _kubernetes_

Then I think the following would almost work...

```
. {

  ...

  k8s_external example.com {
    fallthrough example.com
  }
  rewrite suffix example.com. cluster.local.  # order in the corefile doesn't matter, just listing here between the other two for clarity.
  kubernetes cluster.local

  ...

}
```

However it also exposes ClusterIP service IPs that don't have external IPs. 🤔

TomasKohout:
I didn't know that it's possible to do this. :slightly_smiling_face:  But I think that it could cause confusion on our users side. :slightly_frowning_face:

chrisohaver:
Since only endpoint queries require special attention, I think just _rewrite_ will work with stock CoreDNS.

```
. {
  rewrite regex ^(.+\..+\..+)\.example\.com\.$ {1}.svc.cluster.local. auto
  k8s_external example.com
  kubernetes cluster.local
}
```


With this, the query `ep1.svc1.default.example.com.` will get modified internally to `ep1.svc1.default.svc.cluster.local.` before being processed by _k8s_external_. It then get processed by the _kubernetes_ plugin instead of the _k8s_external_.  Before writing the response to the client, the name is restored to `ep1.svc1.default.example.com.`, so the operation is transparent to the client.

edit: I fixed missing `svc` label in above proposed solution.

chrisohaver:
@TomasKohout, I just realized my proposed fix was missing the `svc` label, and edited the comment to fix it.

TomasKohout:
I've change a bit configuration that you've suggested.

```
rewrite stop {
  name suffix .example.com .svc.cluster.local. answer auto
}
```

But this also expose ClusterIPs which I'd like to keep hidden in that example.com dns zone.

chrisohaver:
> I've change a bit configuration that you've suggested.
> 
> ```
> rewrite stop {
>   name suffix .example.com .svc.cluster.local. answer auto
> }
> ```
> 
> But this also expose ClusterIPs which I'd like to keep hidden in that example.com dns zone.

That is because you changed it to match any number of leading labels, which will match also service queries (2 leading labels).  My suggested solution will only affect endpoint queries, because it matches 3 labels.

But now I'm thinking that in addition to querying endpoints by name directly, you also want to be able to query a headless service by name (without an endpoints specified), and get a list of Pod IPs back.  In that case, my assumption above "Since only endpoint queries require special attention" is incorrect.  Service queries also need to change.  This makes it not possible with _rewrite_, since a headless service query and clusterIP service query are indistinguishable by name.

plugin/kubernetes/external.go
```diff
@@ -96,27 +142,54 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
...
+		// Endpoint query or headless service
+		if headless && svc.Headless() {
+			
+			idx := object.ServiceKey(svc.Name, svc.Namespace)
+		    endpointsList :=  k.APIConn.EpIndex(idx)
+	
+			for _, ep := range endpointsList {
+				if ep.Name != svc.Name || ep.Namespace != svc.Namespace {
+					continue
+				}
```
chrisohaver: This check is not necessary since `endpointsList` is the result of an index lookup, and the index is the name/namespace.

plugin/kubernetes/external.go
```diff
@@ -96,27 +142,54 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
...
+
+				for _, eps := range ep.Subsets {
+					
+					for _, addr := range eps.Addresses {
+
```
chrisohaver: Minor: Could remove the empty lines here.

TomasKohout:
So what do you think @chrisohaver? :slightly_smiling_face: Does this change make sense?

chrisohaver:
> So what do you think @chrisohaver? 🙂 Does this change make sense?

I'm reviewing the changes, though I won't have much free time in the coming week, so it may take some time before I can get to it.

TomasKohout:
Sure thing. :slightly_smiling_face: Have a nice time :-)

plugin/k8s_external/transfer.go
```diff
@@ -40,18 +40,20 @@ func (e *External) Transfer(zone string, serial uint32) (<-chan []dns.RR, error)
...
 			// externalAddrFunc returns incomplete header names, correct here
 			nsRecords[i].Header().Name = nsName
 			nsRecords[i].Header().Ttl = e.ttl
 			ch <- []dns.RR{nsRecords[i]}
 		}
 
-		svcs := e.externalServicesFunc(zone)
+		svcs := e.externalServicesFunc(zone, e.headless)
 		srvSeen := make(map[string]struct{})
+		// svcsGroup := make(map[string][]msg.Service)
```
chrisohaver:
```suggestion
```

plugin/k8s_external/transfer_test.go
```diff
@@ -136,7 +139,9 @@ func difference(testRRs []dns.RR, gotRRs []dns.RR) []dns.RR {
 	}
 
 	foundRRs := []dns.RR{}
+	cmp := map[string]struct{}{}
 	for _, rr := range gotRRs {
+		cmp[rr.String()] = struct{}{}
```
chrisohaver: `cmp` does not appear to be used by anything.

plugin/kubernetes/external.go
```diff
@@ -75,16 +85,52 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
 			continue
 		}
 
-		for _, ip := range svc.ExternalIPs {
-			for _, p := range svc.Ports {
-				if !(matchPortAndProtocol(port, p.Name, protocol, string(p.Protocol))) {
+		if svc.Headless() || endpoint != "" {
+			if endpointsList == nil {
+				endpointsList = endpointsListFunc()
```
chrisohaver: Is the function variable indirection necessary?  Can you just call `k.APIConn.EpIndex(idx)` directly?

plugin/kubernetes/external.go
```diff
@@ -75,16 +85,52 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
 			continue
 		}
 
-		for _, ip := range svc.ExternalIPs {
-			for _, p := range svc.Ports {
-				if !(matchPortAndProtocol(port, p.Name, protocol, string(p.Protocol))) {
+		if svc.Headless() || endpoint != "" {
```
chrisohaver: Missing `headless` check?

plugin/kubernetes/external.go
```diff
@@ -12,8 +12,8 @@ import (
 )
 
 // External implements the ExternalFunc call from the external plugin.
-// It returns any services matching in the services' ExternalIPs.
-func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
+// It returns any services matching in the services' ExternalIPs and if enabled, headless endpoints..
+func (k *Kubernetes) External(state request.Request, headless bool) ([]msg.Service, int) {
```
chrisohaver, line 28: I could be missing it, but I don't the see the `headless` bool parameter checked/used in this function.

plugin/kubernetes/ns_test.go
```diff
@@ -103,7 +103,7 @@ func TestNsAddrs(t *testing.T) {
 	k.APIConn = &APIConnTest{}
 	k.localIPs = []net.IP{net.ParseIP("10.244.0.20")}
 
-	cdrs := k.nsAddrs(false, k.Zones[0])
+	cdrs := k.nsAddrs(false, false, k.Zones[0])
```
chrisohaver, line 106: Should add tests for when `headless` is true.

plugin/kubernetes/external.go
```diff
@@ -75,16 +85,52 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
...
-				services = append(services, s)
+				for _, eps := range ep.Subsets {
+					for _, addr := range eps.Addresses {
+
+						// See comments in parse.go parseRequest about the endpoint handling.
+						if endpoint != "" {
+							if !match(endpoint, endpointHostname(addr, k.endpointNameMode)) {
+								continue
+							}
+						}
```
chrisohaver:
```suggestion
						
						if endpoint != "" && !match(endpoint, endpointHostname(addr, k.endpointNameMode)) {
							continue
						}
						
```

plugin/kubernetes/external.go
```diff
@@ -75,16 +85,52 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
...
 				}
-				rcode = dns.RcodeSuccess
-				s := msg.Service{Host: ip, Port: int(p.Port), TTL: k.ttl}
-				s.Key = strings.Join([]string{zonePath, svc.Namespace, svc.Name}, "/")
 
-				services = append(services, s)
+				for _, eps := range ep.Subsets {
+					for _, addr := range eps.Addresses {
+
+						// See comments in parse.go parseRequest about the endpoint handling.
```
chrisohaver: What specific comments there are you referring to, and how do they relate to this code?

plugin/kubernetes/external.go
```diff
@@ -96,27 +142,54 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
...
 	for _, svc := range k.APIConn.ServiceList() {
-		for _, ip := range svc.ExternalIPs {
-			for _, p := range svc.Ports {
-				s := msg.Service{Host: ip, Port: int(p.Port), TTL: k.ttl}
-				s.Key = strings.Join([]string{zonePath, svc.Namespace, svc.Name}, "/")
-				services = append(services, s)
-				s.Key = strings.Join(append([]string{zonePath, svc.Namespace, svc.Name}, strings.ToLower("_"+string(p.Protocol)), strings.ToLower("_"+string(p.Name))), "/")
-				s.TargetStrip = 2
-				services = append(services, s)
+		// Endpoint query or headless service
```
chrisohaver: "Endpoint query" doesn't make sense in this context, because this function is not used in the context of a record query.

plugin/kubernetes/external.go
```diff
@@ -96,27 +142,54 @@ func (k *Kubernetes) External(state request.Request) ([]msg.Service, int) {
...
+						for _, p := range eps.Ports {
+							s := msg.Service{Host: addr.IP, Port: int(p.Port), TTL: k.ttl}
+							baseSvc := strings.Join([]string{zonePath, svc.Namespace, svc.Name}, "/")
+							s.Key = strings.Join([]string{baseSvc, endpointHostname(addr, k.endpointNameMode)}, "/")
+							services = append(services, s)
+						}
+					}
+				}
+			}
+			continue
```
chrisohaver, line 195: Also need to append the headless service record (without endpoint label) to `services`.

chrisohaver (changes requested):
First pass, comments in line.  

If they are not present already, there should be tests for the headless option disabled, and enabled covering any code that checks for option.

Please clean up lint and gofmt issues.

plugin/kubernetes/external.go
TomasKohout: yep, it should be used in the if you've mentined before.

plugin/kubernetes/external.go
TomasKohout: In parse.go is explained how to get each element from a dns name. This specific code was taken from kubernetes.go line n. 543. 

IMHO it's related in the sense of understanding endpointHostname function.

chrisohaver: ~OK, how about...~


~// See comments in parse.go parseRequest about endpoint record format.~


edit: actually, disregard that.  I'm still confused by this comment. I'm not seeing how the comments in parse help provide clarity in the code following this comment.

plugin/kubernetes/ns_test.go
TomasKohout: Will do.

plugin/kubernetes/external.go
TomasKohout: I've deleted that comment.

plugin/kubernetes/external.go
TomasKohout: I've deleted the endpoint label and also added port and protocol with stripTarget set to 2. Hope it's ok.

plugin/kubernetes/external.go
TomasKohout: I think that it's added in previous section now. :slightly_smiling_face:

plugin/kubernetes/ns_test.go
TomasKohout: I've added some tests, when you'll have time, could you please check if it's enough? :slightly_smiling_face:

plugin/k8s_external/transfer.go
TomasKohout: Deleted.

plugin/k8s_external/transfer_test.go
TomasKohout: Removed.

plugin/kubernetes/external.go
TomasKohout: Yep, it's not cecessary, removed.

plugin/kubernetes/external.go
TomasKohout: True, fixed. :slightly_smiling_face:

plugin/kubernetes/external.go
TomasKohout: Removed. :slightly_smiling_face:

plugin/kubernetes/external.go
chrisohaver: I meant in the comment "Endpoint query" doesn't make sense in this function, because this function is only use for transfers, so it would never process a endpoint query.  I wasn't suggesting a code change here, just a remarking on the comment.  Perhaps `// Endpoints and headless services`  is better.

plugin/kubernetes/ns_test.go
```diff
@@ -52,16 +54,21 @@ var svcs = []*object.Service{
 		Namespace:  "kube-system",
 		ClusterIPs: []string{"10::111"},
 	},
+	{
+		Name:       "headless-service",
+		Namespace:  "kube-system",
+		ClusterIPs: []string{api.ClusterIPNone},
+	},
```
chrisohaver: Can you use the `hdls-dns-service`, which already exists?

plugin/kubernetes/ns_test.go
```diff
@@ -52,16 +54,21 @@ var svcs = []*object.Service{
...
 	return svcs
 }
 
 func (APIConnTest) EpIndexReverse(ip string) []*object.Endpoints {
-	if ip != "10.244.0.20" {
-		return nil
-	}
+	// if ip != "10.244.0.20" {
+	// 	return nil
+	// }
```
chrisohaver: This results in a mocked index that behaves incorrectly, returning values that don't match the index requested.

plugin/kubernetes/ns_test.go
```diff
@@ -103,10 +118,10 @@ func TestNsAddrs(t *testing.T) {
 	k.APIConn = &APIConnTest{}
 	k.localIPs = []net.IP{net.ParseIP("10.244.0.20")}
 
-	cdrs := k.nsAddrs(false, false, k.Zones[0])
+	cdrs := k.nsAddrs(false, true, k.Zones[0])
```
chrisohaver: Need too add a test case here, not change the existing on. In the new test case, the first argument should be true, since you want to test in the context of an external query. 

i.e. with the pre-existing services endpoints defined, 

`k.nsAddrs(true, true, k.Zones[0])` should have 1 result (hdls-dns-service)
`k.nsAddrs(true, false, k.Zones[0])` should have 0 results (Because no services in the test have an External IP.  Maybe one should be added so that's actually tested too).

`k.nsAddrs(false, true, k.Zones[0])` is kind of a non-case, since when determining the ns addresses from the internal context, the headless services should always included.  i.e. `k.nsAddrs(false, true, k.Zones[0])` and `k.nsAddrs(false, false, k.Zones[0])`  should yield the same result. IOW, when external is false, we should ignore the value for headless (its not an option in the internal context).  Otherwise, we would need to be very careful never to pass external=false and headless=false to nsAddrs or we'd get a wrong result.

chrisohaver:
Please also include documentation updates to the README - it will help with reviewing to know what the intended behavior is rather than trying to imply it from code (which may or may not actually be working as intended).

Kind of related to that: We should decide the best behavior for when a headless service also has an external IP.  If `headless` is set, should it...
* return records for ONLY the endpoint IPs (this is how the PR is currently implemented)
* return records for ONLY the external IPs
* return records for BOTH endpoint IPs and external IPs

If there are valid use cases for more than one of these behaviors, then we may need to make it a controllable option.

plugin/kubernetes/external.go
TomasKohout: I see. :smile:

TomasKohout:
> Please also include documentation updates to the README - it will help with reviewing to know what the intended behavior is rather than trying to imply it from code (which may or may not actually be working as intended).
> 
> Kind of related to that: We should decide the best behavior for when a headless service also has an external IP. If `headless` is set, should it...
> 
> * return records for ONLY the endpoint IPs (this is how the PR is currently implemented)
> * return records for ONLY the external IPs
> * return records for BOTH endpoint IPs and external IPs
> 
> If there are valid use cases for more than one of these behaviors, then we may need to make it a controllable option.

Not sure what is the best behavior here, but I'd vote for the second one since if you set External IP you probably want to use it. Also this won't break backward compatibility.

I'm going to update the readme now. :slightly_smiling_face:

TomasKohout:
I've changed behavior of ExternalServices function so now AFXR test transfer works. I'm not sure whether that change is acceptable though. :slightly_smiling_face:

chrisohaver:
> I've changed behavior of ExternalServices function so now AFXR test transfer works. I'm not sure whether that change is acceptable though. 🙂

What did you change?

chrisohaver:
> What did you change?

OK - I see.  Those changes don't seem like the right direction. I don't understand the explanation added to the code. Why are the changes needed?

TomasKohout:
> > What did you change?
> 
> OK - I see. Those changes don't seem like the right direction. I don't understand the explanation added to the code. Why are the changes needed?

Hmmm. I think that without those changes, transfer protocol woudn't work as expected because it would have wrong or missing records.

The reason behind this is that headless and "external" services require different handling. Because you want to have records for endpoint and also use those endpoints in service level record. Or am I missing something?

chrisohaver:
> The reason behind this is that headless and "external" services require different handling. Because you want to have records for endpoint and also use those endpoints in service level record. Or am I missing something?

I'll need to carve some time out to review more carefully.

IMO, the existing `etcd/msg.Service` layer here makes this pointlessly more difficult to solve - which is probably why we side-stepped it entirely in the _kubernetes_ transfer implementation (which emits records directly as dns.RRs).

FWIW, refactoring kubernetes and k8s_external to not use plugin.Backend, and etcd/msg.Service is a task gathering dust in my todo/wish list.  But it's a huge (and unexciting) task, so it has been too easy to find excuses not to do it.

chrisohaver (approved):
Sorry it took me so long to get back to this.  I think this looks OK.  It does reinforce my wish to refactor _kubernetes_ and _k8s_external_ away from the etcd.Msg object and plugin/backend.  But that isn't happening now.

Thanks!

chrisohaver (changes requested):
Actually I think there may be some remaining issues in  ns_tests.go.

I need to look those over again.

plugin/kubernetes/ns_test.go
TomasKohout: Sure thing!

plugin/kubernetes/ns_test.go
TomasKohout: From what I see in nsAddrs function, when external is false, the whole section is skipped, so that condition holds.

I'll create additional test.

TomasKohout: Actually, I've already created test case for headless service. So, it should be OK. 

I've added test for already existing service with externalIP. :slightly_smiling_face:

plugin/kubernetes/ns_test.go
TomasKohout: Uncommented.

TomasKohout:
I've changed things that you've pointed out. :slightly_smiling_face: Thanks for you attention!

chrisohaver (approved):

PR #5508, opened 2022-07-12
plugin/rewrite: Support min and max TTL values

### LINKED ISSUE #5501: Allow configuring min & max TTL in rewrite plugin
The rewrite plugin allows overwriting the TTL in DNS response to a fixed value.
I wanted to use this plugin to reduce TTL on some domains. But while using it I realized that if the existing TTL value is already lower than the configured value, the rewrite plugin actually increases the TTL value by overriding it - that is not the behavior I wanted.

I think overwriting TTL could be used more flexibly if the plugin accepts a **min-max** TTL range instead of a single value.
The logic would then be to set TTL to **min** if it is lower than **min**, set it to **max** if it is greater than **max**, or else leave TTL value unchanged.

For example setting **ttl suffix . 30-300** would restrict TTL values on all domains to the range 30s to 300s.
And to stay backwards compatible with a single TTL value, both **min** and **max** can be set to the single provided value.

Is anyone else interested in this feature? If yes, I could work on a pull request.

Does the syntax **min-max** make sense or would you prefer using 2 separate arguments for **min** and **max** with the 2nd being optional for backwards compatibility?

AndreasHuber-CH:
Signed-off-by: Andreas Huber <andreas.huber@nagra.com>

Add support for specifying min & max TTL values in rewrite plugin.

Implements enhancement described in #5501

I've updated README.md to give some examples of the new TTL config.

The change is backward compatible

### DIFF
--- a/plugin/rewrite/README.md
+++ b/plugin/rewrite/README.md
@@ -273,7 +273,27 @@ The syntax for the TTL rewrite rule is as follows. The meaning of
 An omitted type is defaulted to `exact`.
 
 ```
-rewrite [continue|stop] ttl [exact|prefix|suffix|substring|regex] STRING SECONDS
+rewrite [continue|stop] ttl [exact|prefix|suffix|substring|regex] STRING [SECONDS|MIN-MAX]
+```
+
+It is possible to supply a range of TTL values in the `SECONDS` parameters instead of a single value.
+If a range is supplied, the TTL value is set to `MIN` if it is below, or set to `MAX` if it is above.
+The TTL value is left unchanged if it is already inside the provided range.
+The ranges can be unbounded on either side.
+
+TTL examples with ranges:
+```
+# rewrite TTL to be between 30s and 300s
+rewrite ttl example.com. 30-300
+
+# cap TTL at 30s
+rewrite ttl example.com. -30 # equivalent to rewrite ttl example.com. 0-30
+
+# increase TTL to a minimum of 30s
+rewrite ttl example.com. 30-
+
+# set TTL to 30s
+rewrite ttl example.com. 30 # equivalent to rewrite ttl example.com. 30-30
 ```
 
 ## EDNS0 Options
--- a/plugin/rewrite/ttl.go
+++ b/plugin/rewrite/ttl.go
@@ -14,11 +14,16 @@ import (
 )
 
 type ttlResponseRule struct {
-	TTL uint32
+	minTTL uint32
+	maxTTL uint32
 }
 
 func (r *ttlResponseRule) RewriteResponse(rr dns.RR) {
-	rr.Header().Ttl = r.TTL
+	if rr.Header().Ttl < r.minTTL {
+		rr.Header().Ttl = r.minTTL
+	} else if rr.Header().Ttl > r.maxTTL {
+		rr.Header().Ttl = r.maxTTL
+	}
 }
 
 type ttlRuleBase struct {
@@ -26,10 +31,10 @@ type ttlRuleBase struct {
 	response   ttlResponseRule
 }
 
-func newTTLRuleBase(nextAction string, ttl uint32) ttlRuleBase {
+func newTTLRuleBase(nextAction string, minTtl, maxTtl uint32) ttlRuleBase {
 	return ttlRuleBase{
 		nextAction: nextAction,
-		response:   ttlResponseRule{TTL: ttl},
+		response:   ttlResponseRule{minTTL: minTtl, maxTTL: maxTtl},
 	}
 }
 
@@ -108,7 +113,7 @@ func newTTLRule(nextAction string, args ...string) (Rule, error) {
 	if len(args) == 3 {
 		s = args[2]
 	}
-	ttl, valid := isValidTTL(s)
+	minTtl, maxTtl, valid := isValidTTL(s)
 	if !valid {
 		return nil, fmt.Errorf("invalid TTL '%s' for a ttl rule", s)
 	}
@@ -116,22 +121,22 @@ func newTTLRule(nextAction string, args ...string) (Rule, error) {
 		switch strings.ToLower(args[0]) {
 		case ExactMatch:
 			return &exactTTLRule{
-				newTTLRuleBase(nextAction, ttl),
+				newTTLRuleBase(nextAction, minTtl, maxTtl),
 				plugin.Name(args[1]).Normalize(),
 			}, nil
 		case PrefixMatch:
 			return &prefixTTLRule{
-				newTTLRuleBase(nextAction, ttl),
+				newTTLRuleBase(nextAction, minTtl, maxTtl),
 				plugin.Name(args[1]).Normalize(),
 			}, nil
 		case SuffixMatch:
 			return &suffixTTLRule{
-				newTTLRuleBase(nextAction, ttl),
+				newTTLRuleBase(nextAction, minTtl, maxTtl),
 				plugin.Name(args[1]).Normalize(),
 			}, nil
 		case SubstringMatch:
 			return &substringTTLRule{
-				newTTLRuleBase(nextAction, ttl),
+				newTTLRuleBase(nextAction, minTtl, maxTtl),
 				plugin.Name(args[1]).Normalize(),
 			}, nil
 		case RegexMatch:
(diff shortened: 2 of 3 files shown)


plugin/rewrite/README.md
```diff
@@ -276,6 +276,27 @@ An omitted type is defaulted to `exact`.
 rewrite [continue|stop] ttl [exact|prefix|suffix|substring|regex] STRING SECONDS
 ```
 
+It is possible to supply a range of TTL values in the `SECONDS` parameters instead of a single value.
```
chrisohaver, line 279: The Syntax section also needs to be updated to reflect the new syntax.

chrisohaver (changes requested):

plugin/rewrite/ttl.go
```diff
@@ -14,22 +14,27 @@ import (
...
+	}
 }
 
 type ttlRuleBase struct {
 	nextAction string
 	response   ttlResponseRule
 }
 
-func newTTLRuleBase(nextAction string, ttl uint32) ttlRuleBase {
+func newTTLRuleBase(nextAction string, minTtl uint32, maxTtl uint32) ttlRuleBase {
```
Tantalor93:
```suggestion
func newTTLRuleBase(nextAction string, minTtl, maxTtl uint32) ttlRuleBase {
```

plugin/rewrite/ttl.go
```diff
@@ -151,22 +156,56 @@ func newTTLRule(nextAction string, args ...string) (Rule, error) {
...
-	if err != nil {
-		return uint32(0), false
-	}
-	if i > 2147483647 {
-		return uint32(0), false
+func isValidTTL(v string) (uint32, uint32, bool) {
+	s := strings.Split(v, "-")
+	if len(s) == 1 {
+		i, err := strconv.Atoi(s[0])
+		if err != nil || i > 2147483647 || i < 0 {
```
Tantalor93: what is number `2147483647`? Seems like a random magic number, maybe worth introducing named constant for this with proper explanation?

plugin/rewrite/ttl.go
```diff
@@ -151,22 +156,56 @@ func newTTLRule(nextAction string, args ...string) (Rule, error) {
...
+			return uint32(0), uint32(0), false
```
Tantalor93: no need for explicit type?
```suggestion
			return 0, 0, false
```

plugin/rewrite/ttl.go
```diff
@@ -151,22 +156,56 @@ func newTTLRule(nextAction string, args ...string) (Rule, error) {
...
-		return uint32(0), false
+	if len(s) == 2 {
+		var min, max int
+		var err error
+		if s[0] == "" {
+			min = 0
+		} else {
+			min, err = strconv.Atoi(s[0])
+			if err != nil {
+				return uint32(0), uint32(0), false
```
Tantalor93:
```suggestion
				return 0, 0, false
```

plugin/rewrite/ttl.go
```diff
@@ -151,22 +156,56 @@ func newTTLRule(nextAction string, args ...string) (Rule, error) {
...
+			}
+		}
+		if min > 2147483647 || min < 0 {
+			return uint32(0), uint32(0), false
```
Tantalor93:
```suggestion
			return 0, 0, false
```

plugin/rewrite/ttl.go
```diff
@@ -151,22 +156,56 @@ func newTTLRule(nextAction string, args ...string) (Rule, error) {
...
+			if s[0] == "" {
+				// explicitly reject ttl directive "-" that would otherwise be interpreted
+				// as 0-2147483647 which is pretty useless
+				return uint32(0), uint32(0), false
+			}
+			max = 2147483647
+		} else {
+			max, err = strconv.Atoi(s[1])
+			if err != nil {
+				return uint32(0), uint32(0), false
```
Tantalor93:
```suggestion
				return 0, 0, false
```

plugin/rewrite/ttl.go
```diff
@@ -151,22 +156,56 @@ func newTTLRule(nextAction string, args ...string) (Rule, error) {
...
+			}
+		}
+		if min > 2147483647 || min < 0 {
+			return uint32(0), uint32(0), false
+		}
+		if s[1] == "" {
+			if s[0] == "" {
+				// explicitly reject ttl directive "-" that would otherwise be interpreted
+				// as 0-2147483647 which is pretty useless
+				return uint32(0), uint32(0), false
```
Tantalor93:
```suggestion
				return 0, 0, false
```

plugin/rewrite/ttl.go
```diff
@@ -151,22 +156,56 @@ func newTTLRule(nextAction string, args ...string) (Rule, error) {
...
+			}
+			max = 2147483647
+		} else {
+			max, err = strconv.Atoi(s[1])
+			if err != nil {
+				return uint32(0), uint32(0), false
+			}
+		}
+		if max > 2147483647 || max < 0 {
+			return uint32(0), uint32(0), false
```
Tantalor93:
```suggestion
			return 0, 0, false
```

plugin/rewrite/ttl.go
```diff
@@ -151,22 +156,50 @@
...
+			max, err = strconv.Atoi(s[1])
+			if err != nil || max > 2147483647 || max < 0 {
+				return uint32(0), uint32(0), false
+			}
+		}
+		if min > max {
+			// reject invalid range
+			return uint32(0), uint32(0), false
+		}
+		return uint32(min), uint32(max), true
```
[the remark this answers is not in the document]
Tantalor93: I think you need to use `strconv.ParseUint(value, 10, 32)`

plugin/rewrite/ttl.go
```diff
@@ -151,22 +156,56 @@ func newTTLRule(nextAction string, args ...string) (Rule, error) {
...
-	i, err := strconv.Atoi(v)
-	if err != nil {
-		return uint32(0), false
-	}
-	if i > 2147483647 {
-		return uint32(0), false
+func isValidTTL(v string) (uint32, uint32, bool) {
+	s := strings.Split(v, "-")
+	if len(s) == 1 {
+		i, err := strconv.Atoi(s[0])
```
Tantalor93: `strconv.ParseUint(s[1], 10, 32)` would be better I think? That will also check that value will fit uint32 range. Also in other conversions

AndreasHuber-CH:
Hi,
I was on vacation the last weeks so it took a bit more time to respond.
I've pushed a simplification using strconv.ParseUint and removing the range checks.
The code I've written before was based on the existing code, but it certainly looks nicer now.

plugin/rewrite/README.md
```diff
@@ -273,7 +273,28 @@ The syntax for the TTL rewrite rule is as follows. The meaning of
 An omitted type is defaulted to `exact`.
 
 ```
-rewrite [continue|stop] ttl [exact|prefix|suffix|substring|regex] STRING SECONDS
+rewrite [continue|stop] ttl [exact|prefix|suffix|substring|regex] STRING [SECONDS|minSECONDS-maxSECONDS]
```
chrisohaver:
```suggestion
rewrite [continue|stop] ttl [exact|prefix|suffix|substring|regex] STRING [SECONDS|MIN-MAX]
```

plugin/rewrite/README.md
```diff
@@ -273,7 +273,28 @@ The syntax for the TTL rewrite rule is as follows. The meaning of
...
+```
+
+It is possible to supply a range of TTL values in the `SECONDS` parameters instead of a single value.
+If a range is supplied, the TTL value is increased to the configured minimum value if it is below,
```
chrisohaver:
```suggestion
If a range is supplied, the TTL value is set to `MIN` if it is below,
```

plugin/rewrite/README.md
```diff
@@ -273,7 +273,28 @@ The syntax for the TTL rewrite rule is as follows. The meaning of
...
+or decreased to the configured maximum value if it is above.
```
chrisohaver:
```suggestion
or set to MAX if it is above.
```

chrisohaver (changes requested):
Thanks.  Minor doc changes suggested.

Tantalor93 (commented):
Please fix the DCO, otherwise LGTM

Tantalor93 (approved):

PR #5511, opened 2022-07-14
Add test for DNS over TLS

mariuskimmina:
Signed-off-by: Marius Kimmina <mar.kimmina@gmail.com>

It adds a test for serving DNS over TLS (using the tls and whoami plugin). I was not able to find any test for this before.

None

None

No

### DIFF
--- a/test/tls_test.go
+++ b/test/tls_test.go
@@ -1,0 +1,60 @@
+package test
+
+import (
+	"crypto/tls"
+	"testing"
+
+	"github.com/miekg/dns"
+)
+
+func TestDNSoverTLS(t *testing.T) {
+	testcases := []struct {
+		name         string
+		config       string
+		Qname        string
+		Qtype        uint16
+		Answer       []dns.RR
+		AnswerLength int
+	}{
+		{
+			name: "Test whoami with DNS over TLS",
+			config: `tls://.:1053 {
+                tls ../plugin/tls/test_cert.pem ../plugin/tls/test_key.pem
+                whoami
+            }`,
+			Qname:        "example.com.",
+			Qtype:        dns.TypeA,
+			Answer:       []dns.RR{},
+			AnswerLength: 0,
+		},
+	}
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


test/tls_test.go
```diff
@@ -0,0 +1,60 @@
...
+                whoami
+            }`,
+			Qname:        "example.com.",
+			Qtype:        dns.TypeA,
+			Answer:       []dns.RR{},
+			AnswerLength: 0,
+		},
+	}
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
```
Tantalor93: why make it table test, when there is only single test case? This rather decreases readability

test/tls_test.go
```diff
@@ -0,0 +1,60 @@
...
+			client := dns.Client{
+				Net:       "tcp-tls",
+				TLSConfig: &tls.Config{InsecureSkipVerify: true},
+			}
+			r, _, err := client.Exchange(m, tcp)
+
+			if err != nil {
+				t.Fatalf("Could not exchange msg: %s", err)
+			}
+
```
Tantalor93, line 32: can you also check rcode?
```
	if r.Rcode != dns.RcodeSuccess {
		t.Errorf("Expected success but got %d", d.Rcode)
	}
```

Tantalor93 (commented):
also please fix linter issues

test/tls_test.go
mariuskimmina: The Idea was that this would make it easier to add other testcases in the future. Sure for one test it doesn't really add anything but I don't think it hurts either. Do you want me to get rid of it?

test/tls_test.go
mariuskimmina: Sure, added it - and also fixed the golangci-lint

test/tls_test.go
Tantalor93: it can be later easily rewritten as table test when needed, but for now I would go with a simple test, almost all these integration tests are written as non-table tests

mariuskimmina:
Thank you for your review @Tantalor93, I made all the changes you request.

zouyee (approved):

PR #5512, opened 2022-07-15
fix E2E tests execution

Tantalor93:

Currently, it seems that E2E tests are not properly executed due to missing deps, so I have executed recommended go get command
image

see for example https://github.com/coredns/coredns/runs/7074611979?check_suite_focus=true
none

none

none


Tantalor93:
nvm, it seems to be executed even without it this PR 🙂

Issue #5510, opened 2022-07-13
Reverse DNS lookups for local/private addresses are classified as error


Tantalor93:
**What happened**:
When doing reverse DNS lookups for local addresses like `dig @127.0.0.1 1.1.168.192.in-addr.arpa. -t PTR`. These requests are logged and classified as errors by `log` plugin, they are logged like this:
```
[INFO] 127.0.0.1:50337 - 13057 "PTR IN 1.1.168.192.in-addr.arpa. udp 65 false 1232" NXDOMAIN qr,rd,ra 53 0.01310145s
```
when running CoreDNS with Corefile
```
.:53 {
  log {
    class error
  }
  forward . 8.8.8.8
}
```

**What you expected to happen**:
These requests should not be logged, when `log` plugin is configured to log only errors as defined in documentation https://coredns.io/plugins/log/#syntax

**How to reproduce it (as minimally and precisely as possible)**:
Run CoreDNS with Corefile
```
.:53 {
  log {
    class error
  }
  forward . 8.8.8.8
}
```
and do DNS resolution like
```
dig @127.0.0.1 1.1.168.192.in-addr.arpa. -t PTR
```
and observe logs

**Anything else we need to know?**:
it seems to be classified as errors, because responses to these DNS requests does not have SOA record in authority section, they look like this
```
; <<>> DiG 9.17.17 <<>> @127.0.0.1 1.1.168.192.in-addr.arpa. -t PTR
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 13057
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;1.1.168.192.in-addr.arpa.	IN	PTR

;; Query time: 14 msec
;; SERVER: 127.0.0.1#53(127.0.0.1) (UDP)
;; WHEN: Wed Jul 13 15:59:38 CEST 2022
;; MSG SIZE  rcvd: 53
```

These local IP reverse DNS records are served from Blackhole servers https://en.wikipedia.org/wiki/Blackhole_server

**Environment**:

- the version of CoreDNS: 1.9.3
- Corefile:
```
.:53 {
  log {
    class error
  }
  forward . 8.8.8.8
}

```
- logs, if applicable:
```
[...]

Issue #5313, opened 2022-04-11
Allow caching of non-authoritative NXDOMAIN replies


lmarszal:
**AWS VPC DNS doesn't provide SOA section for cached NXDOMAIN replies**

Use ``for X in `seq 20` ; do dig not-existing-10.us-west-2.compute.internal @[IP-OF-VPC-DNS-SERVER] ; done`` to query AWS VPC-DNS a few times.  
You'll see that first reply always contains SOA section, but subsequent replied doesn't. You can test it with any other query that returns NXDOMAIN error - the result is the same.

My assumption is that AWS VPC-DNS caches DNS replies, but doesn't include SOA section for cached replies. This is clear violation of RFC2308.

According to RFC2308 ([https://datatracker.ietf.org/doc/html/rfc2308#section-6](https://datatracker.ietf.org/doc/html/rfc2308#section-6)):

>    When a server, in answering a query, encounters a cached negative
>    response it MUST add the cached SOA record to the authority section
>    of the response with the TTL decremented by the amount of time it was
>    stored in the cache.  This allows the NXDOMAIN / NODATA response to
>    time out correctly.

Also, same RFC ([https://datatracker.ietf.org/doc/html/rfc2308#section-3](https://datatracker.ietf.org/doc/html/rfc2308#section-3)):

>    Name servers authoritative for a zone MUST include the SOA record of
>    the zone in the authority section of the response when reporting an
>    NXDOMAIN or indicating that no data of the requested type exists.

I reported this issue to AWS support few months ago, but they hadn't fixed it yet.

**Why this is a problem for coredns?**

Coredns cache plugin caches only valid replies. NXDOMAIN replies from AWS VPC DNS that doesn't have SOA section are considered invalid and are not cached (see [this code](https://github.com/coredns/coredns/blob/882b2d913d28b2f80c4597a16b5043a1a92fb1b8/plugin/pkg/response/typify.go#L105-L107) for what's considered valid NXDOMAIN reply).

When using coredns in kubernetes (or EKS) NXDOMAIN replies are something normal because of [ndot:5 configuration and search domains](https://github.com/kubernetes/kubernetes/issues/33554#issuecomment-266251056). To make things worse, AWS by default adds `[region].compute.internal` to the list of your search domains on every node - meaning every FQDN dns query from apps running on k8s will eventually be forwarded to AWS VPC-DNS and won't be cached. This renders coredns caching feature useless in some cases.

**What would you like to be added**:

[...]

PR #5524, opened 2022-07-18
NXDOMAIN messages with no SOA authority are classified as NameError

Tantalor93:
This PR adjusts the classification of DNS messages with rcode NXDOMAIN and no SOA in the authority section to be considered as `NameError`, currently, it is classified as `OtherError`, which causes behavior pointed out by issues #5510 and #5313 . This adjustment is based on the issues stated before and also based on [RFC](https://datatracker.ietf.org/doc/html/RFC2308#section-2.1)

> Name errors (NXDOMAIN) are indicated by the presence of "Name Error" in
   the RCODE field.  In this case the domain referred to by the QNAME does
   not exist.  Note: the answer section may have SIG and CNAME RRs and
   authority section may have SOA, NXT and SIG RRsets.
   

This will cause : 
* `cache` plugin denial configuration will be applied to these messages
* `log` plugin will not classify messages like these #5510 as errors

effects on `sign`, `dns64`, and `minimal` plugins should be effectively none

#5510 and #5313

### DIFF
--- a/plugin/dnssec/dnssec.go
+++ b/plugin/dnssec/dnssec.go
@@ -52,7 +52,7 @@ func (d Dnssec) Sign(state request.Request, now time.Time, server string) *dns.M
 	}
 
 	if mt == response.NameError || mt == response.NoData {
-		if req.Ns[0].Header().Rrtype != dns.TypeSOA || len(req.Ns) > 1 {
+		if len(req.Ns) == 0 || req.Ns[0].Header().Rrtype != dns.TypeSOA || len(req.Ns) > 1 {
 			return req
 		}
 
--- a/plugin/dnssec/dnssec_test.go
+++ b/plugin/dnssec/dnssec_test.go
@@ -123,6 +123,20 @@ func TestSigningEmpty(t *testing.T) {
 	}
 }
 
+func TestSigningNameErrorWithNoSoa(t *testing.T) {
+	d, rm1, rm2 := newDnssec(t, []string{"miek.nl."})
+	defer rm1()
+	defer rm2()
+
+	m := &dns.Msg{}
+	m.SetQuestion("a.miek.nl.", dns.TypeA)
+	state := request.Request{Req: m, Zone: "miek.nl."}
+	m = d.Sign(state, time.Now().UTC(), server)
+	if !section(m.Ns, 0) {
+		t.Errorf("Authority section should have 0 RRSIGs")
+	}
+}
+
 func section(rss []dns.RR, nrSigs int) bool {
 	i := 0
 	for _, r := range rss {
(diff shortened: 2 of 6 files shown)


chrisohaver:
Will this cause NXDOMAIN responses without an SOA to be cached?  If so, that goes against RFC 2308.

Tantalor93:
> Will this cause NXDOMAIN responses without an SOA to be cached? If so, that goes against RFC 2308.

correct 🤔 but they are cached even in the current state due to https://github.com/coredns/coredns/blob/5b9b079dabc7f71463cea3f0c6a92f338935039d/plugin/cache/cache.go#L159
because such responses are classified as `OtherError`

chrisohaver:
>>Will this cause NXDOMAIN responses without an SOA to be cached? If so, that goes against RFC 2308.

>correct 🤔 but they are cached even in the current state ...

Isn't #5313 a request for us to change cache behavior so it will cache NXDOMAIN responses without an SOA, because we don't currently?

Tantalor93:
> Isn't https://github.com/coredns/coredns/issues/5313 a request for us to change cache behavior so it will cache NXDOMAIN responses without an SOA, because we don't currently?

yes, it is. It seems that issue is not correct? Because it looks like that we are caching them as positive answers

chrisohaver:
> yes, it is. It seems that issue is not correct? Because it looks like that we are caching them as positive answers

Have you verified this on a running coredns?  Or is this based only on code observation?

chrisohaver:
The following  Corefile demonstrates: 
* `valid` (NXDOMAIN with SOA) is cached
* `invalid` (NXDOMAIN without SOA) is not cached.

Since no _log_ plugin is present in the _cache_ server block, log entries will appear only on cache misses (when they reach the _template_ block).  Log messages appear every time for `invalid`, but only once every 30s for `valid`.

```
.:5399 {
  cache 30
  forward . 127.0.0.1:5398
}

.:5398 {
  log
  # NXDOMAIN without an SOA (invalid)
  template IN ANY invalid {
    rcode NXDOMAIN
  }
  # NXDOMAIN with an SOA (valid)
  template IN ANY valid {
    rcode NXDOMAIN
    authority "{{ .Zone }} 60 IN SOA ns.example.com hostmaster.example.com (1 60 60 60 60)"
  }
}
```

Tantalor93:
@chrisohaver right, your snippet is correct, they are not cached. I tried to verify it by a combination of observation and debugging, but it seems I was wrong. In that case, this PR would be against RFC you mentioned, I'll close this, this is not the proper way to fix those issues mentioned. Thanks for reviews!

chrisohaver:
Since unit tests passed in this PR, that means one or both of the following are true:
1. there is no unit test coverage for this behavior
2. the changes in the PR did not alter behavior

I suspect it's 1, in which case we should add a case that validates the NXDOMAIN without SOA case.

PR #5529, opened 2022-07-21
fix minor code unreachability error

Abirdcfly:
Signed-off-by: Abirdcfly <fp544037857@gmail.com>

t.Error* will report test failures but continue executing the test. 
t.Fatal* will report test failures and stop the test immediately.
**So some of the code is actually unreachable**

An example:
https://go.dev/play/p/5K3KZ5RzHhR
```go
package main

import (
	"testing"
)

// LastIndex returns the index of the last instance of x in list, or
// -1 if x is not present.
func LastIndex(list []int, x int) int {
	for i := len(list) - 1; i >= 0; i-- {
		if list[i] == x {
			return i
		}
	}
	return -1
}

func TestLastIndex(t *testing.T) {
	tests := []struct {
		list []int
		x    int
		want int
	}{
		{list: []int{1}, x: 1, want: 0},
		{list: []int{1, 1}, x: 1, want: 3}, // change want:1 to want:3, should fail
		{list: []int{2, 1}, x: 2, want: 0},
		{list: []int{1, 2, 1, 1}, x: 2, want: 1},
		{list: []int{1, 1, 1, 2, 2, 1}, x: 3, want: -1},
		{list: []int{3, 1, 2, 2, 1, 1}, x: 3, want: 111}, // change want:0 to want:111, should fail
	}
	for i, tt := range tests {
		t.Logf("index:%d", i)
		if got := LastIndex(tt.list, tt.x); got != tt.want {
			t.Errorf("LastIndex(%v, %v) = %v, want %v", tt.list, tt.x, got, tt.want)
		}
	}
}

func TestLastIndex2(t *testing.T) {
	tests := []struct {
		list []int
		x    int
		want int
	}{
		{list: []int{1}, x: 1, want: 0},
		{list: []int{1, 1}, x: 1, want: 3}, // change want:1 to want:3, should fail
		{list: []int{2, 1}, x: 2, want: 0},
		{list: []int{1, 2, 1, 1}, x: 2, want: 1},
		{list: []int{1, 1, 1, 2, 2, 1}, x: 3, want: -1},
		{list: []int{3, 1, 2, 2, 1, 1}, x: 3, want: 111}, // change want:0 to want:111, should fail
	}
	for i, tt := range tests {
		t.Logf("index:%d", i)
		if got := LastIndex(tt.list, tt.x); got != tt.want {
			t.Fatalf("LastIndex(%v, %v) = %v, want %v", tt.list, tt.x, got, tt.want) // Fatalf is equivalent to Logf followed by FailNow, so the following code will be unreachable
			continue  // unreachable
		}
	}
}

/* output:

=== RUN   TestLastIndex
    prog.go:32: index:0
    prog.go:32: index:1
    prog.go:34: LastIndex([1 1], 1) = 1, want 3
    prog.go:32: index:2
    prog.go:32: index:3
    prog.go:32: index:4
    prog.go:32: index:5
    prog.go:34: LastIndex([3 1 2 2 1 1], 3) = 0, want 111
--- FAIL: TestLastIndex (0.00s)
=== RUN   TestLastIndex2
    prog.go:53: index:0
    prog.go:53: index:1
    prog.go:55: LastIndex([1 1], 1) = 1, want 3
--- FAIL: TestLastIndex2 (0.00s)
FAIL

*/
```

No
No
No

### DIFF
--- a/plugin/rewrite/ttl_test.go
+++ b/plugin/rewrite/ttl_test.go
@@ -83,7 +83,6 @@ func TestTtlRewrite(t *testing.T) {
 		rule, err := newRule(r.args...)
 		if err != nil {
 			t.Fatalf("Rule %d: FAIL, %s: %s", i, r.args, err)
-			continue
 		}
 		if reflect.TypeOf(rule) != r.expectedType {
 			t.Fatalf("Rule %d: FAIL, %s: rule type mismatch, expected %q, but got %q", i, r.args, r.expectedType, rule)


plugin/rewrite/ttl_test.go
```diff
@@ -83,7 +83,6 @@ func TestTtlRewrite(t *testing.T) {
 		rule, err := newRule(r.args...)
 		if err != nil {
 			t.Fatalf("Rule %d: FAIL, %s: %s", i, r.args, err)
```
chrisohaver, line 85:
```suggestion
			t.Errorf("Rule %d: FAIL, %s: %s", i, r.args, err)
			continue
```

plugin/rewrite/ttl_test.go
```diff
@@ -83,7 +83,6 @@ func TestTtlRewrite(t *testing.T) {
...
-			continue
 		}
 		if reflect.TypeOf(rule) != r.expectedType {
 			t.Fatalf("Rule %d: FAIL, %s: rule type mismatch, expected %q, but got %q", i, r.args, r.expectedType, rule)
```
chrisohaver, line 88:
```suggestion
			t.Errorf("Rule %d: FAIL, %s: rule type mismatch, expected %q, but got %q", i, r.args, r.expectedType, rule)
			continue
```

chrisohaver (approved):
I initially suggested we Error instead of Fatal, so other test cases could be verified, but it looks like this test build rules for a second test, so if a rule is missing, the following test's expectations would be misaligned.  So using the existing use of Fatal is correct/required.

PR #5537, opened 2022-07-25
plugin/template: Add CNAME example to README

chrisohaver:

Just adds an example CNAME template since the behavior is slightly different from other record templates (i.e. it does the CNAME lookup, and includes the target record in the answer if upstream can resolve it).


Tantalor93 (approved):
LGTM

PR #5538, opened 2022-07-26
plugin/view: Advanced routing interface and new 'view' plugin

chrisohaver:

In a nutshell, this PR adds BIND-like view capability to CoreDNS.

More specifically, it adds an advanced routing interface (`dnsserver.Viewer`), that allows a plugin implementing it to decide if a query should be routed into its server block.

It also adds a new plugin *view*, that uses the new interface to enable a user to define expression based conditions that must be met for a query to be routed to its server block.  This is similar in function to BIND views. 

For example:
The following forwards clients in `10.9.0.0/24` to `10.9.0.100`, and all other clients to the servers in `/etc/resolv.conf`.

```
. {
  view example-view {
    expr incidr(client_ip(), '10.9.0.0/24')
  }
  forward . 10.9.0.100
}

. {
  forward . /etc/resolv.conf
}
```

Please see the *view* README for more examples and complete syntax.

#2833


core/dnsserver/server_https.go
```diff
@@ -50,9 +50,11 @@ func NewServerHTTPS(addr string, group []*Config) (*ServerHTTPS, error) {
...
 	// TLS configuration returns an error: it can only be specified once.
 	var tlsConfig *tls.Config
-	for _, conf := range s.zones {
-		// Should we error if some configs *don't* have TLS?
-		tlsConfig = conf.TLSConfig
+	for _, z := range s.zones {
+		for _, conf := range z {
+			// Should we error if some configs *don't* have TLS?
+			tlsConfig = conf.TLSConfig
+		}
```
chrisohaver, line 57: I just noticed that this will be a problem if multiple configs for the same zone have different TLS configs. Not sure if this is easy to fix.

core/dnsserver/server_grpc.go
```diff
@@ -37,9 +37,11 @@ func NewServergRPC(addr string, group []*Config) (*ServergRPC, error) {
...
 	// TLS configuration returns an error: it can only be specified once.
 	var tlsConfig *tls.Config
-	for _, conf := range s.zones {
-		// Should we error if some configs *don't* have TLS?
-		tlsConfig = conf.TLSConfig
+	for _, z := range s.zones {
+		for _, conf := range z {
+			// Should we error if some configs *don't* have TLS?
+			tlsConfig = conf.TLSConfig
+		}
```
chrisohaver, line 44: ditto here.

core/dnsserver/server_tls.go
```diff
@@ -28,9 +28,11 @@ func NewServerTLS(addr string, group []*Config) (*ServerTLS, error) {
...
 	// TLS configuration returns an error: it can only be specified once.
 	var tlsConfig *tls.Config
-	for _, conf := range s.zones {
-		// Should we error if some configs *don't* have TLS?
-		tlsConfig = conf.TLSConfig
+	for _, z := range s.zones {
+		for _, conf := range z {
+			// Should we error if some configs *don't* have TLS?
+			tlsConfig = conf.TLSConfig
+		}
```
chrisohaver, line 35: and here.

core/dnsserver/server_https.go
```diff
@@ -63,8 +65,10 @@ func NewServerHTTPS(addr string, group []*Config) (*ServerHTTPS, error) {
 
 	// Use a custom request validation func or use the standard DoH path check.
 	var validator func(*http.Request) bool
-	for _, conf := range s.zones {
-		validator = conf.HTTPRequestValidateFunc
+	for _, z := range s.zones {
+		for _, conf := range z {
+			validator = conf.HTTPRequestValidateFunc
+		}
```
chrisohaver, line 71: This is also a problem if multiple configs for the same zone have different http request validate funcs.

core/dnsserver/server_https.go
chrisohaver: Although, this already a problem for different server blocks that listen on the same port (i.e. same server).  Despite the comment above this, _tls_ does not in fact enforce one TLS definition per server, it just enforces one defined per Config (each server block creates a separate Config. each Server can serve multiple Configs).

So I don't think this introduces a new problem.  But the existing problem should be fixed - or documented.

core/dnsserver/server_https.go
chrisohaver: This has the same pre-existing issue as tls.

plugin/view/README.md
```diff
@@ -0,0 +1,117 @@
...
+}
+
+. {
+  forward . 10.0.0.1
+}
+```
+
+## Expressions
+
+Expressions use Kinetic.govaluate (https://github.com/Knetic/govaluate), which "Provides support for evaluating arbitrary
```
Tantalor93: I would be quite hesitant to add a new logic on top of [govaluate](https://github.com/Knetic/govaluate) project, it seems to be no longer actively maintained. Is there any alternative we can use? Would it be possible for example to use standard GO templates (like in [Helm](https://helm.sh/)) for this expression logic?

chrisohaver: Good point. On first thought I think Go template syntax is verbose and hard to read, but I’ll take another look.

chrisohaver: Googling for "govaluate alternatives" First hit is https://github.com/antonmedv/expr, which looks promising, and appears actively maintained.

Tantalor93: that seems better choice than govaluate, if you think that GO templates are not suited for this functionality 👍

chrisohaver: I think maybe we could make Go Template work. If we take the user input as a conditional, and insert it into a hard coded if statement?  e.g.

We can have a boilerplate template: 
```
{{ if <user expression> }}
ok
{{end}}
```

And when we evaluate a view expression, we replace `<user expression>` with the Go template expression defined by the user in the Corefile, and evaluate the template.  If it evaluates to the string "ok" then it's a pass.  This kind of feels like forcing the wrong puzzle piece.  Maybe there is a better way to use Go templates to do this - I don't have much experience with it.

The Go template expression syntax looks a bit unfriendly though:
* it appears to use prefix notation, which is not easy to read for most people.  
* function call syntax also doesn't appear to use parentheses to group all their arguments, so to read an expression that has a function call in it, you have to know how many arguments it takes.

I think those issues would be tolerable for very simple expressions that comprise a single function call, or a single comparison.

e.g. single operation no too bad
```
view incidr client_ip "1.2.3.0/24"
```

multiple operations ...  

```
view and eq class "IN" and eq type "A" incidr client_ip "1.2.3.0/24"
```

many operations ...
```
view and eq class "IN" or and eq type "A" incidr client_ip "1.2.3.0/24" and eq type "AAAA" incidr client_ip "1:2::3:4/64" 
```

What do you think?

chrisohaver: Hmm - looks like I could cut a bit less against the grain with Go templates for custom function calls, and attach methods to certain state fields. e.g. enabling cleaner syntax like:

```
view ClientIP.InCIDR "1.2.3.0/24"
```

But that doesn't really resolve the prefix notation thing.

The question I guess is, does the benefit of going with a Go-native tool outweigh its apparently loose technical fit, and cryptic syntax. (I think it may)

Tantalor93: true, the GO template might not be the best fit for your use case, it would probably somehow work, but it seems it would be too hacky.

from what I have tried the https://github.com/antonmedv/expr seems like a nice fit. API seems quite understandable, also it seems to be well documented and actively developed. Expression language seems understandable and easily customizable

chrisohaver:
I landed on making all query values accessible as functions instead of variables (e.g. `client_ip()` instead of `client_ip`), for a couple of reasons:
1. **For serve time efficiency**: As functions, creating the expression environment from the query state doesn't require evaluating every value up front, and so they are only evaluated if actually used by the compiled expression.
2. **For code simplicity**: As functions, compiling the expression during setup can be done with a nil state, instead of having to create a "fake" `request.Request` with staged values.

chrisohaver:
Adding metadata support, which works in real life, but need to update some scattered unit tests to account for changes I needed to make to metadata plugin.  The collection of metadata functions from each plugin used to be done by the metadata plugin itself in `metadata.ServeDNS` when it was executed in the plugin chain.  But now that we need to be able to evaluate metadata _before_ we select a plugin to enter, the collection of metadata functions is now instead done earlier in `server.ServeDNS`. So unit tests need to be updated to do that step manually before calling `metadata.ServerDNS` directly.

... ok looks like I got them all.

SuperQ (commented):
This is really great. I think this would be useful to better solve issues like handling dns64 responses from different networks (#4809).

I'm wondering if views should have names, like they do in bind. This way metrics can be separated by view name.

chrisohaver:
> This is really great. I think this would be useful to better solve issues like handling dns64 responses from different networks (#4809).
> 
> I'm wondering if views should have names, like they do in bind. This way metrics can be separated by view name.

I had forgotten to consider metrics, thanks for the reminder.  Yes, each server block should be uniquely identifiable with an identifier (e.g. name) so we can add that as a new vector to the server metrics.  I think I can do this by altering the Viewer interface to return the identifier of the corresponding view.  I think though that I should alter the syntax of the view plugin to better fit: only allow a single view plugin per server block; add a name option. e.g. 

```
view NAME {
  expr EXPRESSION
}
```

Or, just 

```
view NAME EXPRESSION
```

chrisohaver:
I have added view names, which triggered a bit of code adjustment.  Now, only one view plugin can be defined per server block. Furthermore, only one `Viewer` plugin can be defined per server block (although for now, *view* is the only implementor).  View name, is now added to server metrics if a view is defined (except for when we error out before entering a view, in which case the view blank).  The view name is also exported as metadata (e.g. for use by log plugin).

johnbelamaric:
Awesome. One question. I am not looking at the code so maybe this is a nonsense question. But:

> The collection of metadata functions from each plugin used to be done by the metadata plugin itself in `metadata.ServeDNS` when it was executed in the plugin chain. But now that we need to be able to evaluate metadata _before_ we select a plugin to enter, the collection of metadata functions is now instead done earlier in `server.ServeDNS`. So unit tests need to be updated to do that step manually before calling `metadata.ServerDNS` directly.

Does that mean this "collecction" is done whether or not metadata plugin is enabled? Does that impact performance?

chrisohaver:
> Does that mean this "collecction" is done whether or not metadata plugin is enabled? Does that impact performance?

For that reason, it’s only done if metadata is enabled. 

In more detail: At server start up the code checks for to see if there is a plugin in the chain that implements MetadataCollector (only metadata does). If there is, then it saves a pointer to the plugin in the Config. server.ServeDNS only calls Collect when that pointer is not nil.

johnbelamaric:
Nice, thanks.

core/dnsserver/config.go
```diff
@@ -60,8 +70,14 @@ type Config struct {
 	// firstConfigInBlock is used to reference the first config in a server block, for the
 	// purpose of sharing single instance of each plugin among all zones in a server block.
 	firstConfigInBlock *Config
+
+	// metaCollector references the first MetadataCollector plugin, if one exists
+	metaCollector *MetadataCollector
```
Tantalor93: I think we do not have to declare it as a pointer as it is an Interface type that is already a pointer to the real structure
```suggestion
	metaCollector MetadataCollector
```

plugin/view/setup.go
```diff
@@ -0,0 +1,65 @@
...
+		if len(args) != 1 {
+			return nil, c.ArgErr()
+		}
+		v.viewName = args[0]
+
+		for c.NextBlock() {
+			switch c.Val() {
+			case "expr":
+				args := c.RemainingArgs()
+				prog, err := expr.Compile(strings.Join(args, " "), expr.Env(expression.DefaultEnv(context.TODO(), nil)))
```
Tantalor93: can we rather use background context? I think it is more appropriate, TODO sounds like something that needs to be finished :)
```suggestion
				prog, err := expr.Compile(strings.Join(args, " "), expr.Env(expression.DefaultEnv(context.Background(), nil)))
```

SuperQ (approved):
LGTM from a metrics point of view.

Tantalor93 (commented):
Overall I like the way this is heading, just few minor comments. Also could you add a new integration tests suite into https://github.com/coredns/coredns/tree/master/test  ? Cover some typical use cases and maybe some edge cases for usage of views?

chrisohaver:
I've added some integration tests. We are a bit limited in integration tests, for example I don't have control over client IP address (for testing a client IP split horizon config).  But there is a test for splitting on query type, and leading label in query name.

chrisohaver:
> I've added some integration tests.

I have also added a metadata integration test.  It doesn't illustrate a meaningful use-case,  but it does verify that the server collects metadata as expected (and only when the metadata plugin is enabled), and that the metadata expression function correctly resolves metadata.

Tantalor93:
Thanks @chrisohaver for adding the tests and considering those minor comments 🙂 

also back to the comment from @SuperQ about metrics, what about metrics exposed by other plugins other than `prometheus` plugin, e.g. [acl](https://github.com/coredns/coredns/blob/master/plugin/acl/metrics.go#L17), [cache](https://github.com/coredns/coredns/blob/master/plugin/cache/metrics.go#L17), dnssec, template plugin, dns64 etc.  I think it is also worth adding the `view` dimension also into those metrics to be able to monitor those plugins effectively. WDYT? Or is there any reason not to add the new label to those metrics?

SuperQ:
Having the view label on pretty much everything would be a good idea. Doesn't have to happen all at once. We can follow up with more labeling in other PRs.

Tantalor93:
If you want to solve the rest of the metrics in the followup I am ok with that, still I think it is very much needed, without it the metrics can be very misleading (like cache size)

Tantalor93 (approved):

chrisohaver:
> If you want to solve the rest of the metrics in the followup I am ok with that, still I think it is very much needed, without it the metrics can be very misleading (like cache size)

It would be good to work out for at least one now, to make sure the solution fits well, and to see if we can make any changes here to make it cleaner to add to other plugins. Eg we may need to stuff the view name in the context and or provide a helper function to get the view name.

Tantalor93:
> > If you want to solve the rest of the metrics in the followup I am ok with that, still I think it is very much needed, without it the metrics can be very misleading (like cache size)
> 
> It would be good to work out for at least one now, to make sure the solution fits well, and to see if we can make any changes here to make it cleaner to add to other plugins. Eg we may need to stuff the view name in the context and or provide a helper function to get the view name.

agreed, should be quite straightforward as you are adding the view name into context using metadata https://github.com/coredns/coredns/pull/5538/files#diff-3afb7f5bbeea60f8b64de29eb72dbd78bbf40e1d374985871d2d393cb0ad4351R12, I think it would be good to start with `cache` as one of the more used CoreDNS plugins, but I'll leave it up to you 🙂

chrisohaver:
Although that’s only added if the metadata plugin is enabled.  It could also be retrieved in setup, by hunting for a Viewer plugin in the Config. No additions to context required.

chrisohaver:
I've added the view vector to _cache_ plugin metrics.  Since `ViewName` is already in the Config, there is no need to use the context or adding a helper function.

chrisohaver:
There is a slightly simpler way to implement this:

Instead of having plugins implement the `Viewer` interface, plugins could just directly modify the Config (`Config.ViewName` and `Config.FilterFuncs` ... via Setter/Getter). 

* This would allow *view*, to sit outside the plugin chain like other "config" type plugins (currently as implemented it has to live in the chain).  It sits at the end of the chain, so it's  normally never reached when serving a request. So this would be mostly an organizational benefit.

* However, if *view* outside the chain, we can then we can position it as the first plugin in `plugin.cfg`, meaning that it would get processed before all other plugins - and thus plugins like _cache_ would not have to defer retrieving the view name until server start.  This is a very minor benefit.

* Doing away with the interface and server code that looks for implementors would reduce lines of code a little, which may make it worth doing.

I'll play with it a bit and see how it works out.

chrisohaver:
> I'll play with it a bit and see how it works out.

Well, it is a bit nicer, but it runs afoul with the *metadata* plugin, which requires all  plugins that produce metadata be in the plugin chain.  So since *view* produces metadata, it must sit in the plugin chain. 

We could make the server code produce the metadata instead, and thus no longer require *view* to sit in the chain, but I think that's too hacky.  

I think the implementation is fine as is.

Tantalor93 (approved):
lgtm

chrisohaver:
I’m debating whether to merge this now for 1.9.4, or to wait until 1.10.0.  It’s not backward incompatible, but it is v1 of a big feature. 

Thoughts?

chrisohaver:
I’m leaning toward merge now.

johnbelamaric:
Two questions:
1) What else is in 1.9.4, any critical bug fixes?
2) Does this change introduce regression risk (specifically when it is *not* being used)?

chrisohaver:
> Two questions:
> 
> 1. What else is in 1.9.4, any critical bug fixes?
> 2. Does this change introduce regression risk (specifically when it is _not_ being used)?

No critical fixes that I know of (I think otherwise 1.9.4 would have been released earlier). The catalyst for this release was time elapsed, not a specific issue. 

There are changes to code in server path when views are not used (eg checking to see if views are defined), and changes to when metadata is collected whether or not views are enabled.

To be more cautious, we could very quickly release 1.10.0 with just this PR on the heels of 1.9.4.

johnbelamaric:
I am ok with either option.

yongtang:
I think this PR can wait for 1.10 as it is a fairly big change.

chrisohaver:
> I think this PR can wait for 1.10 as it is a fairly big change.

OK.  And I see you already kicked off the release. So we will go with that.

PR #5540, opened 2022-07-27
plugin/cache: Add cache disable option

chrisohaver:

Disabling negative cache is a common request. It's an option in most dns cache implementations.

Being able to exclude some zones from cache is useful for avoiding caching responses from dynamic backends, while still caching responses from upstream servers forwarded in the same server block.

For example, the _kubernetes_ plugin maintains its own cache, which is dynamically updated in real-ish time via the k8s API watch. Redundantly caching the _kubernetes_ records with the _cache_ plugin prevents clients from seeing the latest data held in the _kubernetes_ cache. Disabling cache entirely is not ideal, since for example responses from a _forward_ plugin in the same server block would no longer be cached. IOW, in this case _cache_ should not cache Kubernetes records, but should cache everything else.

This PR adds the ability to disable the positive and/or negative caches for a list of zones.

kubernetes/kubernetes#92559

### DIFF
--- a/plugin/cache/README.md
+++ b/plugin/cache/README.md
@@ -39,6 +39,7 @@ cache [TTL] [ZONES...] {
     prefetch AMOUNT [[DURATION] [PERCENTAGE%]]
     serve_stale [DURATION] [REFRESH_MODE]
     servfail DURATION
+    disable success|denial [ZONES...] 
 }
 ~~~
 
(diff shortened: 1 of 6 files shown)


Tantalor93 (approved):
LGTM

fm2022aa:
@chrisohaver,  
Hi, I want to ask: if the pod of a statefulset recreate quickly, then the ip of the pod is changed;  the cache may at most incorrect for 30 seconds?
 
the config is :
```
        cache 30 {
          disable denial cluster.local
        }
```

Issue #1060, opened 2017-09-11
dnstap contains neither identity nor version


jpmens:
I'm not sure whether this is a bug or just a feature request. I note that coredns-011 produces dnstap output which contains [neither identity nor version](https://github.com/dnstap/dnstap.pb/blob/master/dnstap.proto#L23).

It would be nice if _identity_ where set to the fqdn of the machine it's running on, and if _version_ could be prepopulated with`CoreDNS-011` or something configurable.

Here's an example from Knot:

```
type: MESSAGE
identity: "knot.ww.mens.de"
version: "Knot DNS 2.5.4"
message:
  type: AUTH_RESPONSE
```

PR #5555, opened 2022-08-04
Identity and version support for the dnstap plugin

djcode:
This PR makes dnstap send identity and version information with dnstap messages. Defaulting to the server hostname and the CoreDNS version, it also adds the ability to override with your custom strings.

```
dnstap dnstap.sock {
  identity NAME
  version VERSION
 }
```

This addresses issue #1060

The dnstap plugin documentation needs to be updated to reflect you can add identity and version parameters in a block.

If the block is not specified, the dnstap plugin will continue to function as it used to, however now sending the hostname and coredns version as the identity and version in dnstap messages.

### DIFF
--- a/plugin/dnstap/handler.go
+++ b/plugin/dnstap/handler.go
@@ -18,12 +18,14 @@ type Dnstap struct {
 
 	// IncludeRawMessage will include the raw DNS message into the dnstap messages if true.
 	IncludeRawMessage bool
+	Identity string
+	Version string
 }
 
 // TapMessage sends the message m to the dnstap interface.
 func (h Dnstap) TapMessage(m *tap.Message) {
 	t := tap.Dnstap_MESSAGE
-	h.io.Dnstap(&tap.Dnstap{Type: &t, Message: m})
+	h.io.Dnstap(&tap.Dnstap{Type: &t, Message: m, Identity: []byte(h.Identity), Version: []byte(h.Version)})
 }
 
 func (h Dnstap) tapQuery(w dns.ResponseWriter, query *dns.Msg, queryTime time.Time) {
--- a/plugin/dnstap/handler.go
+++ b/plugin/dnstap/handler.go
@@ -18,14 +18,14 @@ type Dnstap struct {
 
 	// IncludeRawMessage will include the raw DNS message into the dnstap messages if true.
 	IncludeRawMessage bool
-	Identity string
-	Version string
+	Identity []byte
+	Version []byte
 }
 
 // TapMessage sends the message m to the dnstap interface.
 func (h Dnstap) TapMessage(m *tap.Message) {
 	t := tap.Dnstap_MESSAGE
-	h.io.Dnstap(&tap.Dnstap{Type: &t, Message: m, Identity: []byte(h.Identity), Version: []byte(h.Version)})
+	h.io.Dnstap(&tap.Dnstap{Type: &t, Message: m, Identity: h.Identity, Version: h.Version})
 }
 
 func (h Dnstap) tapQuery(w dns.ResponseWriter, query *dns.Msg, queryTime time.Time) {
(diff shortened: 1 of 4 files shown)


plugin/dnstap/setup.go
```diff
@@ -38,6 +39,26 @@ func parseConfig(c *caddy.Controller) (Dnstap, error) {
...
+				d.Identity = c.Val()
+			}
+  		case "version": {
+				if !c.NextArg() {
+					return d, c.ArgErr()
+				}
+				d.Version = c.Val()
+			}
+		}
+	}
```
mariuskimmina: The formatting here looks a bit odd, did you use `go fmt`?

plugin/dnstap/setup.go
```diff
@@ -38,6 +39,26 @@ func parseConfig(c *caddy.Controller) (Dnstap, error) {
 	}
 
 	d.IncludeRawMessage = c.NextArg() && c.Val() == "full"
+	
+	d.Identity, _ = os.Hostname()
+	d.Version = caddy.AppName + "-" + caddy.AppVersion
```
Tantalor93: Any reason to not use `coremain.CoreVersion` ? https://github.com/coredns/coredns/blob/45b0a11294c59bfd806a57807aaa2a185f761cd5/coremain/version.go#L5

plugin/dnstap/handler.go
```diff
@@ -18,12 +18,14 @@ type Dnstap struct {
...
 	IncludeRawMessage bool
+	Identity string
+	Version string
 }
 
 // TapMessage sends the message m to the dnstap interface.
 func (h Dnstap) TapMessage(m *tap.Message) {
 	t := tap.Dnstap_MESSAGE
-	h.io.Dnstap(&tap.Dnstap{Type: &t, Message: m})
+	h.io.Dnstap(&tap.Dnstap{Type: &t, Message: m, Identity: []byte(h.Identity), Version: []byte(h.Version)})
```
Tantalor93: I would bet these byte slices conversions will cause heap allocations, can we rather store h.Identity and  h.Version as byte slices rather than as strings? To avoid heap allocations on request path

plugin/dnstap/setup.go
djcode: I'm happy to make that change. I just copied the version logic from the chaos plugin. https://github.com/coredns/coredns/blob/master/plugin/chaos/setup.go#L30
I much prefer this variable.

plugin/dnstap/setup.go
djcode: `go fmt` did not complain about this, but I agree! The indentation on github looks different to what I had in my editor.

plugin/dnstap/setup.go
djcode: On further investigation, it looks like coremain sets these values from those constants.
https://github.com/coredns/coredns/blob/master/coremain/run.go#L31
They're available as it is now without having to add an import. Any reason to not use these?

plugin/dnstap/handler.go
djcode: I've moved all the []byte stuff to setup.go

plugin/dnstap/setup.go
Tantalor93: I think it is okay to use these values to have consistency with chaos plugin

plugin/dnstap/handler.go
```diff
@@ -18,12 +18,14 @@ type Dnstap struct {
...
 	IncludeRawMessage bool
+	Identity []byte
+	Version []byte
 }
 
 // TapMessage sends the message m to the dnstap interface.
 func (h Dnstap) TapMessage(m *tap.Message) {
 	t := tap.Dnstap_MESSAGE
-	h.io.Dnstap(&tap.Dnstap{Type: &t, Message: m})
+	h.io.Dnstap(&tap.Dnstap{Type: &t, Message: m, Identity: h.Identity, Version: h.Version})
```
Tantalor93, line 28: Can you also extend handler test https://github.com/coredns/coredns/blob/0188090ae93557d02b7db25b13c83c86b412f172/plugin/dnstap/handler_test.go#L36 and verify identity and version is set as expected?

Tantalor93 (approved):
Thanks!

Issue #5548, opened 2022-07-31
Forward: Allow modification of query flags (e.g. always set RD)


jarus:
**What would you like to be added**:
Allow modification of DNS query flags within the forward plugin. Especially, I would like to set the RD (Recursion Desired) flag so that an upstream DNS server performs internal recursive resolution even the client or another recursive resolver didn't request recursive resolution.

**Why is this needed**:
I'm currently building a rewrite/forward setup towards the internal AWS EC2  and GCP VM DNS servers to allow resolution of EC2 or GCP VM instance names within an existing DNS infrastructure.

Overview and query path:
* Client: Queries `i-abcdef12345.aws-eu-west-1.example.com` through existing On-Prem DNS infrastructure
* On-Prem DNS Recursive Resolver (arbitrary recursive DNS resolver e.g. unbound/bind): Through `NS` records for `aws-eu-west-1.example.com` it knows which "authority nameserver" or in this case On-Prem DNS Forwarder servers should be queried.
* On-Prem DNS Forwarder (CoreDNS) acts as authority for `aws-eu-west-1.example.com` and forwards the query to another CoreDNS server in the AWS VPC. This may sound overkill but allows to use DNS over TLS, caching etc. on the On-Prem DNS Forwarder.
* CoreDNS server in AWS VPC rewrites query from `i-abcdef12345.aws-eu-west-1.example.com` to `i-abcdef12345.eu-west-1.compute.internal` and forwards query to AWS EC2 DNS server (169.254.169.253).
* AWS EC2 DNS server (169.254.169.253): Resolves `i-abcdef12345.eu-west-1.compute.internal` and generate response

[...]

PR #5556, opened 2022-08-06
plugin/header: Add support for query modification (#5548)

jarus:

The header plugin only supported modification of DNS flags in responses. With this pull request it is also possible to modify DNS flags in queries. For example requesting recursive resolution from subsequent plugins or DNS servers. 

#5548 

The documentation of the header plugin has been adjusted for the extended config syntax and capabilities.

The change is fully backward compatible but I assume in the long-run it makes sense to deprecate and remove the less explicit syntax. As I couldn't find a good example how plugins should communicate config deprecation, I would highly appreciate some guidance from maintainers.

### DIFF
--- a/plugin/header/README.md
+++ b/plugin/header/README.md
@@ -13,12 +13,12 @@ The modifications are made transparently for the client and subsequent plugins.
 
 ~~~
 header {
-    SELECTOR ACTION FLAGS...
-    SELECTOR ACTION FLAGS...
+    [SELECTOR] ACTION FLAGS...
+    [SELECTOR] ACTION FLAGS...
 }
 ~~~
 
-* **SELECTOR** defines if the action should be applied on `query` or `response`.
+* **SELECTOR** defines if the action should be applied on `query` or `response`. In future CoreDNS version the selector will be mandatory. For backwards compatibility the action will be applied on `response` if the selector is undefined.
 
 * **ACTION** defines the state for DNS message header flags. Actions are evaluated in the order they are defined so last one has the
   most precedence. Allowed values are:
--- a/plugin/header/README.md
+++ b/plugin/header/README.md
@@ -2,21 +2,23 @@
 
 ## Name
 
-*header* - modifies the header for responses.
+*header* - modifies the header for queries and responses.
 
 ## Description
 
-*header* ensures that the flags are in the desired state for responses. The modifications are made transparently for
-the client.
+*header* ensures that the flags are in the desired state for queries and responses.
+The modifications are made transparently for the client and subsequent plugins.
 
 ## Syntax
 
 ~~~
 header {
-    ACTION FLAGS...
-    ACTION FLAGS...
+    SELECTOR ACTION FLAGS...
+    SELECTOR ACTION FLAGS...
 }
 ~~~
+
+* **SELECTOR** defines if the action should be applied on `query` or `response`.
 
 * **ACTION** defines the state for DNS message header flags. Actions are evaluated in the order they are defined so last one has the
   most precedence. Allowed values are:
@@ -34,7 +36,7 @@ Make sure recursive available `ra` flag is set in all the responses:
 ~~~ corefile
 . {
     header {
-        set ra
+        response set ra
     }
 }
 ~~~
@@ -44,8 +46,18 @@ Make sure "recursion available" `ra` and "authoritative answer" `aa` flags are s
 ~~~ corefile
 . {
     header {
-        set ra aa
-        clear rd
+        response set ra aa
+        response clear rd
+    }
+}
+~~~
+
+Make sure "recursion desired" `rd` is set for all subsequent plugins::
+
+~~~ corefile
+. {
+    header {
+        query set rd
     }
 }
 ~~~
--- a/plugin/header/handler.go
+++ b/plugin/header/handler.go
@@ -8,7 +8,7 @@ import (
 	"github.com/miekg/dns"
 )
 
-// Header modifies dns.MsgHdr in the responses
+// Header modifies flags of dns.MsgHdr in queries and / or responses
 type Header struct {
 	QueryRules    []Rule
 	ResponseRules []Rule
--- a/plugin/header/handler.go
+++ b/plugin/header/handler.go
@@ -10,13 +10,16 @@ import (
 
 // Header modifies dns.MsgHdr in the responses
 type Header struct {
-	Rules []Rule
-	Next  plugin.Handler
+	QueryRules    []Rule
+	ResponseRules []Rule
+	Next          plugin.Handler
 }
 
 // ServeDNS implements the plugin.Handler interface.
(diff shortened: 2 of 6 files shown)


plugin/header/README.md
```diff
@@ -2,22 +2,24 @@
...
+*header* ensures that the flags are in the desired state for queries and responses.
+The modifications are made transparently for the client and subsequent plugins.
 
 ## Syntax
 
 ~~~
 header {
-    ACTION FLAGS...
-    ACTION FLAGS...
+    SELECTOR ACTION FLAGS...
```
mqasimsarfraz: `SELECTOR` should be optional for now with `[SELECTOR]`

plugin/header/README.md
```diff
@@ -2,22 +2,24 @@
...
+    SELECTOR ACTION FLAGS...
 }
 ~~~
 
+* **SELECTOR** defines if the action should be applied on `query` or `response`.
```
mqasimsarfraz: Add a note here that it will be mandatory to specify it and non selector behavior will be removed in future releases.

plugin/header/header_test.go
```diff
@@ -80,3 +80,22 @@ func TestHeader(t *testing.T) {
 		}
 	}
 }
+
+func TestHeaderQueryRules(t *testing.T) {
+	wr := dnstest.NewRecorder(&test.ResponseWriter{})
+	next := plugin.HandlerFunc(func(ctx context.Context, writer dns.ResponseWriter, msg *dns.Msg) (int, error) {
+		if !msg.RecursionDesired {
```
mqasimsarfraz: Why not add a table driven test case like we are doing for `TestHeaderResponseRules` for consistency.

plugin/header/setup.go
```diff
@@ -11,39 +12,60 @@ import (
...
+		var queryRules []Rule
+		var responseRules []Rule
+
 		for c.NextBlock() {
-			v := c.Val()
+			selector := strings.ToLower(c.Val())
+
+			var action string
+			if selector == "set" || selector == "clear" {
+				// In previous CoreDNS version, the headers plugin only modified
```
mqasimsarfraz: Do a `log.Warning` here with a deprecation notice rather than a comment.

mqasimsarfraz (changes requested):
Thanks for the contribution. Some minor comments otherwise looking good from my side! :+1:

plugin/header/README.md
jarus: Adjusted as requested.

plugin/header/README.md
jarus: Adjusted as requested.

plugin/header/setup.go
jarus: Adjusted as requested.

plugin/header/header_test.go
jarus: Sure. I initially tried it to keep the consistency. Sadly I have no real idea how the test could be written in this table style. I tried to implement some kind of nextAssert function which performs the assertion but I didn't manage it to make the compiler happy. As mentioned this are literally my first Go steps. I would highly appreciate some help here.

mqasimsarfraz: This [article](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) is great for introduction. We have more [test cases](https://github.com/coredns/coredns/blob/master/plugin/rewrite/name_test.go#L35) this like in `rewrite` plugin. Please let me know if it doesn't help?

plugin/header/header_test.go
```diff
@@ -80,3 +80,22 @@ func TestHeader(t *testing.T) {
...
+	})
+
+	handler := Header{
+		QueryRules: []Rule{{Flag: recursionDesired, State: false}},
+		Next:       next,
+	}
+
+	m := new(dns.Msg)
+	handler.ServeDNS(context.TODO(), wr, m)
+}
```
mqasimsarfraz:
```suggestion
func TestHeaderQueryRules(t *testing.T) {
	wr := dnstest.NewRecorder(&test.ResponseWriter{})
	next := plugin.HandlerFunc(func(ctx context.Context, writer dns.ResponseWriter, msg *dns.Msg) (int, error) {
		writer.WriteMsg(msg)
		return dns.RcodeSuccess, nil
	})

	tests := []struct {
		handler  plugin.Handler
		got      func(msg *dns.Msg) bool
		expected bool
	}{
		{
			handler: Header{
				QueryRules: []Rule{{Flag: recursionAvailable, State: true}},
				Next:       next,
			},
			got: func(msg *dns.Msg) bool {
				return msg.RecursionAvailable
			},
			expected: true,
		},
		{
			handler: Header{
				QueryRules: []Rule{{Flag: recursionDesired, State: true}},
				Next:       next,
			},
			got: func(msg *dns.Msg) bool {
				return msg.RecursionDesired
			},
			expected: true,
		},
		{
			handler: Header{
				QueryRules: []Rule{{Flag: recursionDesired, State: false}},
				Next:       next,
			},
			got: func(msg *dns.Msg) bool {
				return msg.RecursionDesired
			},
			expected: false,
		},
		{
			handler: Header{
				QueryRules: []Rule{{Flag: authoritative, State: true}},
				Next:       next,
			},
			got: func(msg *dns.Msg) bool {
				return msg.Authoritative
			},
			expected: true,
		},
	}

	for i, tc := range tests {
		m := new(dns.Msg)

		_, err := tc.handler.ServeDNS(context.TODO(), wr, m)
		if err != nil {
			t.Errorf("Test %d: Expected no error, but got %s", i, err)
			continue
		}

		if tc.got(m) != tc.expected {
			t.Errorf("Test %d: Expected flag state=%t, but got %t", i, tc.expected, tc.got(m))
			continue
		}
	}
}
```

plugin/header/header_test.go
jarus: Thank you very much for your help. Especially the test case that you wrote helped me a lot. I honestly didn't thought about checking the DNSMsg after passing it through the handler function. 🤦‍♂️  My brain stuck a bit too much on Go function arguments are Pass-by-Value and forgot about using Pass-by-Reference in this case.

plugin/header/handler.go
```diff
@@ -10,13 +10,16 @@ import (
 
 // Header modifies dns.MsgHdr in the responses
```
mqasimsarfraz: Comment needs an update.

plugin/header/header_test.go
mqasimsarfraz: Glad I was able to help! :+1:

plugin/header/handler.go
jarus: Updated comment. All other comments in `plugin/header` are still valid match with adjusted plugin scope.

mqasimsarfraz: :+1:

mqasimsarfraz (approved):
Thanks for the contribution, LGTM! :+1:

PR #5602, opened 2022-09-07
Do not expand query UDP buffer size if already set to a smaller value

ezaquarii:
Signed-off-by: Krzysztof Narkiewicz <knarkiewicz@bloomberg.net>

It prevents `bufsz` plugin from expanding a UDP buffer size if it is already set to a smaller value.

None, we found that the plugin expands the buffer size that other clients try to keep at smaller value.

None.

No. This change should only improve interoperability.

### DIFF
--- a/plugin/bufsize/bufsize.go
+++ b/plugin/bufsize/bufsize.go
@@ -1,4 +1,4 @@
-// Package bufsize implements a plugin that modifies EDNS0 buffer size.
+// Package bufsize implements a plugin that clamps EDNS0 buffer size preventing packet fragmentation.
 package bufsize
 
 import (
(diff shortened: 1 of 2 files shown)


ykhr53:
thanks for the PR, lgtm.

ezaquarii:
> thanks for the PR, lgtm.

@ykhr53 could you point me to the linter output? It fails because of some trivial formatting issue but I can't find the linter output. Details page contains log but not the error location.

ykhr53:
```
Error: unnecessary leading newline (whitespace)
```

The error is for unnecessary leading newline, but I believe it's basically came from [here](
https://github.com/coredns/coredns/blob/83663445290688ba5ee6dc0250a7c3a1237bf83b/.github/workflows/golangci-lint.yml), so probably you should run `golangci-lint` and check details on your side.

chrisohaver:
> > thanks for the PR, lgtm.
> 
> @ykhr53 could you point me to the linter output? It fails because of some trivial formatting issue but I can't find the linter output. Details page contains log but not the error location.

The linter buries comments inline with the commits.  They are hard to find.

chrisohaver (approved):

ykhr53 (approved):
lgtm

Issue #5584, opened 2022-08-26
Enhancement: tls plugin with acme support (automated certificates)


mariuskimmina:
I would like to introduce a CoreDNS plugin. It's a new tls plugin that can utilize the acme protocol for automated tls certificates (as long as CoreDNS is the authoritative DNS server for the domain). In developing this plugin I had help from @yongtang and @greenpau as my mentors during this years Google Summer of Code.

## What would you like to be added
tlsplus, a (tls) plugin that can obtain and renew certificates if coredns is setup to be the authoritative dns-server for a domain. This plugin can be used as a drop-in replacement for the current tls plugin, which means that it extends the current tls plugin but doesn't break any existing setups.

## Why is this needed
* this aims to solve #3460
* Automatic HTTPS has been one of the most praised features in Caddy, this plugin is an attempt to bring some of that functionality into CoreDNS

## Link to the Plugin (also contains documentation)
https://github.com/mariuskimmina/coredns-tlsplus

## Requirements
* Own a domain
* Setup CoreDNS on a publicly reachable IP
* Setup CoreDNS as the authoritative DNS server for your domain
* Port 53 - While CoreDNS may serve DNS over TLS on any port, during startup the plugin will use port 53 to solve the ACME Challenge

## Design
The Plugin utilizes `certmagic` (the same lib that caddy uses) to initalize the ACME Challenge. During the plugin setup, a `dns.Server` will be started that answers DNS requests that start with `_acme-challenge`. Once the challenge is solved this `dns.Server` gets shutdown again. 
The Plugin also starts a loop that runs in the background and checks if the cert needs renewal every 15 minutes (default, configurable). A Cert needs renewal if 70% of it's total validity period have passed. When a cert needs renewal CoreDNS will restart and thus the plugin setup runs again, renewing the certificate. The restart of CoreDNS is done similarly to the `reload` plugin

![how-it-works](https://user-images.githubusercontent.com/38843153/187634378-93fd623c-ae9e-46bf-acbb-72781e49db6a.png)

## Implementation
I thought about solving the ACME Challenge in a `ServeDNS` function but decided to start a `dns.Server` during plugin setup instead. The Problem with doing it in `serveDNS` is, that once CoreDNS Starts you can't configure it to use the certificate wtihout a restart (AFAIK). So, starting a `dns.Server` during the plugin setup phase made it possible to obtain a certificate, configure it and then start the actual CoreDNS server.

## Confiugration
The tlsplus plugin offers the following options:

[...]

PR #5608, opened 2022-09-11
plugin/tls: add automated certificates via acme

mariuskimmina:
Short summary: When CoreDNS is the authoritative DNS Server for a domain this updated tls plugin allows to obtain and renew certificates for this domain automatically.
For a more detailed description see #5584 

This aims to solve #3460.

The readme of the tls plugin needs to be updated, also done here.

The tls plugin can still be used normally, there should be no backwards incompatible change.

Pinging @yongtang and @greenpau who guided me during the development of this feature.

Edit1: On request, integration tests have now been included into this PR. These tests spin up a [Pebble](https://github.com/letsencrypt/pebble) server as a local CA and obtain a certificate from it.


yongtang:
@mariuskimmina   I think some of the build failures are a little concerning. Can you try to address those failure cases so that build can pass first?

mariuskimmina:
I got a bit confused by the import order check but now I understood it. All green now @yongtang

yongtang:
@mariuskimmina I think it makes sense to add integration tests into  the same PR as it give everyone confidence that the implementation works. Can you add integration tests into the PR?

mariuskimmina:
@yongtang Integration tests have now been added (and the original description has been updated).

test/test/config/pebble-config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "pebble": {
+    "listenAddress": "0.0.0.0:14000",
```
yongtang: Will it be possible not to use hard-coded port numbers? I think maybe we can render a random address within a chunk of golang code?

plugin/tls/setup.go
```diff
@@ -0,0 +1,209 @@
...
+	r              = renewCert{quit: make(chan bool), renew: make(chan bool)}
+	once, shutOnce sync.Once
+)
+
+const (
+	defaultCA            = "https://acme-v02.api.letsencrypt.org/directory"
+	defaultEmail         = "doesnotexist@test.com"
+	defaultCheckInterval = 15
+	defaultPort          = 53
+	defaultCertPath      = "./local/share/certmagic"
```
yongtang: Wondering if this is a generic path we want to use? What about non-Linux platforms? Can we just save the cert file within the same address as coredns binary or Corefile?

mariuskimmina: This is what `certmagic` uses by default. Saving it in the same location as the coredns binary would be possible. 
Since there are a couple more files than just one cert, how about creating a directory `tls` - then we would end up with a directory structure like this:

```
.
├── Corefile
├── coredns
└── tls
    ├── acme
    │   ├── acme-staging-v02.api.letsencrypt.org-directory
    │   │   └── users
    │   │       └── doesnotexist@test.com
    │   │           ├── doesnotexist.json
    │   │           └── doesnotexist.key
    │   └── acme-v02.api.letsencrypt.org-directory
    │       ├── challenge_tokens
    │       └── users
    │           └── doesnotexist@test.com
    │               ├── doesnotexist.json
    │               └── doesnotexist.key
    ├── certificates
    │   └── acme-staging-v02.api.letsencrypt.org-directory
    │       └── ns1.subdomain.mariuskimmina.com
    │           ├── ns1.subdomain.mariuskimmina.com.crt
    │           ├── ns1.subdomain.mariuskimmina.com.json
    │           └── ns1.subdomain.mariuskimmina.com.key
    ├── locks
    └── ocsp
        └── ns1.subdomain.mariuskimmina.com-9e3130e8
```

All directorys and filenames inside `tls` are then created by `certmagic`

yongtang: @mariuskimmina  sounds good, probably '.tls' to avoid possible collision?

test/test/config/pebble-config.json
mariuskimmina: I mean, in theory sure, but that is just the config of the CA for integration tests - I am not sure what the benefit of a random port over 14000 or any other number would be - can you elaborate? (It would certainly make the entire test more complex)

yongtang: This is just to avoid an increased chance of integration test failure, as I assume 14000 or 15000 are more commonly used (increased chance of collisions). If you don't want to complicate the randomization, I will suggest you may pick a hard-coded random port , e.g., 14234, 15345, etc. than a not-so-random 14000 or 15000.

plugin/tls/setup.go
mariuskimmina: `.tls` in the current directory is now being used.

yongtang:
Currently testing different scenarios. Noticed that in case a machine's dns pointing to coredns itself then there might be some dependency that needs to be resolved. Looking into it now.

mariuskimmina:
To clarify here a bit: Currently for this to work, the system resolver (`/etc/resolv.conf`) needs to point to some external resolver (like 8.8.8.8 or 1.1.1.1) because CoreDNS needs to be able to resolve `acme-v02.api.letsencrypt.org` (or the ACME Server Name of any other CA) during start-up. 

It would be nice to set the resolver for this specific request inside the code but currently this seems rather difficult to do (maybe I am missing something?), espically since this request is being created and send inside `certmagic`.

mariuskimmina:
I've written a full setup walktrough for this, which can be found here: https://mariuskimmina.com/blog/automated_certs_coredns/#example

greenpau:
> I've written a full setup walktrough for this, which can be found here:

@mariuskimmina , looks great! Awesome work!

mariuskimmina:
@chrisohaver any thoughts on this? If it feels like to niche of a use case I could also offer the plugin as an external plugin (tranfering the project to CoreDNS org)

chrisohaver:
> @chrisohaver any thoughts on this? If it feels like to niche of a use case I could also offer the plugin as an external plugin (tranfering the project to CoreDNS org)

Going through old PRs, this comment evaded my radar. I think this is a good feature to include in CoreDNS.

I have not reviewed the code.  I will review it if no other maintainers will.

mariuskimmina:
Glad to hear that, it's been a while since I last worked on this but I would still be happy to try and make any changes required to move forward

SuperQ:
I like the idea of having an acme client here, but this setup seems to assume the CoreDNS server is publicly available.

Maybe it would be better to use the [go-acme/lego](https://pkg.go.dev/github.com/go-acme/lego/v4) library. It supports a number of challenge configurations which would allow for a more flexible setup for getting certificates.

mariuskimmina:
Yes this does require the CoreDNS server to be publicly reachable. My understanding of the original discussion in #3460 was that they wanted a plugin that let's coredns solve the dns challenge itself - by setting a TXT record which the CA can verify. For that the dns server needs to be publicly reachable.

Doing this without being public reachable could be done in the same way e.g. caddy does it by using the API of another DNS Server to solve the challenge - is that what you have in mind @SuperQ ?

simonfelding:
As a coredns user administering a corporate network and trying to get rid of BIND, I would appreciate this feature. I don't mind having the DNS server publically available. If you want coredns to be more than "the kubernetes DNS server", I think this would be a very nice feature.

chrisohaver:
@SuperQ, @yongtang. Is further work needed on this PR?

SuperQ:
> Doing this without being public reachable could be done in the same way e.g. caddy does it by using the API of another DNS Server to solve the challenge - is that what you have in mind @SuperQ ?

Yes, I think we should have support for `DNS-01` challenges. I would love if CoreDNS could answer these itself, rather than rely on `HTTP-01` as the only way to get certs.

mariuskimmina:
> > Doing this without being public reachable could be done in the same way e.g. caddy does it by using the API of another DNS Server to solve the challenge - is that what you have in mind @SuperQ ?
> 
> Yes, I think we should have support for `DNS-01` challenges. I would love if CoreDNS could answer these itself, rather than rely on `HTTP-01` as the only way to get certs.

It is solving the DNS challenge by answering these itself, answering these requests does require CoreDNS to be able to receive them, which requires it being publicly reachable.

SuperQ:
Yes, self-solving would require public access, but only DNS, not HTTP. But with https://github.com/go-acme/lego configured with an external provider the solver could happen externally with no need for public reach.

mariuskimmina:
> Yes, self-solving would require public access, but only DNS, not HTTP. But with https://github.com/go-acme/lego configured with an external provider the solver could happen externally with no need for public reach.

Yes, I agree. That was basically my previous suggestions that we could use the API of an external DNS Server to solve the challenge, that way CoreDNS would not need to be reachable from the outside.
On the HTTP part tho, It's been a while since I worked on this but I don't remember doing HTTP anywhere here, where are you getting that from?

SuperQ:
Oh, I assumed that since you were using the Caddy solver it was using `HTTP-01` challenges. Since this is what Caddy defaults to.

mariuskimmina:
I have some free time on my hands and looked into https://github.com/go-acme/lego 
I am now working on a new version of this that focuses on using external providers and won't require CoreDNS to be exposed. 

I'll open a new PR for that once ready

houyuwushang:
Closing as superseded by #8310, which implemented the self-hosted DNS-01 certificate-management scope and closed #3460.

Thank you @mariuskimmina for the earlier implementation and design discussion. This work helped inform the lifecycle and dependency boundaries of the merged solution and is credited in #8310.

The external DNS-provider direction mentioned in the latest comment is not part of this PR and can be proposed separately if it is pursued.

PR #5609, opened 2022-09-11
plugin/template: Add parseInt template function

erikjoh:
Signed-off-by: Erik Johansson <ejohansson@spotify.com>

This change adds a template function `parseInt` that can be used in templates to for example parse an IPv4 address
in hexadecimal encoding.
This also allows replicating the PowerDNS [feature](https://doc.powerdns.com/authoritative/lua-records/functions.html#createForward) of getting auto generated A records with the hexadecimal format (for example `ip7b7b7b7b.example.`). Today, the template plugin only seems to allow for a dashed decimal IPv4 format (for example `ip-123-123-123-123.example.`).

I've tried to find a way to solve this with native template functions but it seems like adding `strconv.ParseUint` like this would be the best way to go.

No issue found.

Template plugin README.

No.


chrisohaver (changes requested):
Looks good.  Can you add an operational test?

erikjoh:
@chrisohaver Added [test/template_test.go](https://github.com/coredns/coredns/pull/5609/files#diff-dc80a0b200539d65f776e1cd5feef9a9407fafbfd08e512bc4eef5666d7e99be) since I didn't find any existing file which seemed appropriate.

chrisohaver:
I'm sorry I used confusing terminology - perhaps I should have said "functional test". IOW I meant a unit test that covers the functionality of the new operation added.  e.g. the tests in  `/plugin/template/template_test.go`.  The unit tests in `setup_test.go` confirm config parsing, but not the operation of the new feature.

`/tests/` is used for integration tests where we need to run a server; it's usually reserved for testing things that cannot be tested in a unit test (e.g. functionality across multiple plugins or integrations with dnsserver).

erikjoh:
Sorry for that misunderstanding on my part as well and sorry to everyone who got pinged due to CODEOWNERS! :sweat: 
Does this look like a step in the right direction?
The tests in `/plugin/template/template_test.go` currently mock all the Go templates so it seems more appropriate to have this functional test in `/plugin/template/setup_test.go` since what we essentially want to test is that `templateParse` actually is mapping `parseInt` to `strconv.ParseUint` and not some other function.

chrisohaver:
OK - I don't see that as mocking, but I see what you mean - the template objects are being created manually in the test.

So, either you'd need to replicate the function addition in the test, or better, break out the template creation into a function to be called from setup and the test  e.g. 

```Golang 
				for _, answer := range args {
					tmpl, err := gotmpl.New("answer").Funcs(funcMap).Parse(answer)
					if err != nil {
						return handler, c.Errf("could not compile template: %s, %v", c.Val(), err)
					}
					t.answer = append(t.answer, tmpl)
				}
```

becomes

```Golang 
				for _, answer := range args {
					err := t.addTemplate("answer", answer)
					if err != nil {
						return handler, c.Errf("could not compile template: %s, %v", c.Val(), err)
					}
				}
```

erikjoh:
Based on your suggestion, wouldn't `addTemplate` require either `reflect` or a `switch` like this to find the correct field to append the template to:

```golang
func (t template) addTemplate(field, text string) (error) {
  funcMap := gotmpl.FuncMap{
    "parseInt": strconv.ParseUint,
  }
  tmpl, err := gotmpl.New(field).Funcs(funcMap).Parse(text)
  if err != nil {
    return err
  }
  switch field {
  case "answer":
    t.answer = append(t.answer, tmpl)
  case "additional":
    t.additional = append(t.additional, tmpl)
  case "authority":
    t.authority = append(t.authority, tmpl)
  default:
    return errors.New("not a valid field")
  }
}
``` 

Is that what you had in mind or would it be enough to just have a function to get a new template?

```golang
func newTemplate(name, text string) (*gotmpl.Template, error) {
  funcMap := gotmpl.FuncMap{
    "parseInt": strconv.ParseUint,
  }
  return gotmpl.New(name).Funcs(funcMap).Parse(text)
}
```

Considering how the objects are created in the test now I think the second option would be simpler.

https://github.com/coredns/coredns/blob/4116a4ebdb0caae743cc9ebe0b462a0055c12754/plugin/template/template_test.go#L20-L27 

Also would you like me to remove the test from the last commit?
Currently that test should cover the functionality that is added.

chrisohaver:
Sorry. Fat fingered close on my phone.

chrisohaver:
Yes, the 2nd is what I had in mind, to fit easily into the existing tests. It’s not how it came out when I gave the code example though. Thanks for catching that.

chrisohaver:
> Also would you like me to remove the test from the last commit?
Currently that test should cover the functionality that is added.

Yes. I don’t think they are necessary. Sorry again about misleading you to write them.

chrisohaver (approved):
LGTM Thanks!

PR #5627, opened 2022-09-17
plugin/view : fix README

Tantalor93:

README of `view` plugin mentions that `server_port` is a client port, I think it is more correct to say that it is a port on a server-side


chrisohaver (approved):
Thanks

Issue #5614, opened 2022-09-12
Steering Committee Election


chrisohaver:
OK, let's kick off the (first) steering committee election process.

* We'll initially aim to fill 5 seats.  If there isn't enough interest, we should discuss if we should reduce the committee size to 3 seats.
* Folks who want to be in the steering committee please declare that here in this issue.
* After some time (two weeks?), we will start the voting.

How does this sound?

@bradbeam
@chrisohaver
@darshanime
@dilyevsky
@ekleiner
@greenpau
@ihac
@inigohu
@Isolus
@jameshartig
@johnbelamaric
@miekg
@mqasimsarfraz
@nchrisdk
@nitisht
@pmoroney
@rajansandeep
@rdrozhdzh
@rtreffer
@snebel29
@stp-ip
@SuperQ
@Tantalor93
@varyoo
@ykhr53
@yongtang
@zouyee

yongtang:
> We'll initially aim to fill 5 seats. If there isn't enough interest, we should discuss if we should reduce the committee size to 3 seats.

One of the concerns is that we want to make sure no single company can dominate the decision making process. Note CNCF   _explicitly_ does not want to see one or two company dominate the governance.

If a structure is going to be changed then there are multiple options available I believe.

chrisohaver:
> If a structure is going to be changed then there are multiple options available I believe.

Thanks.  We can discuss those if there isn't enough interest to fill 5 seats.

chrisohaver:
I'd like to run for a committee member seat. I am employed by Infoblox.

johnbelamaric:
I'd like to run for a committee member seat. I am employed by Google.

Tantalor93:
Even though I am a relatively new maintainer, I'd like to run for a committee member seat. I am employed by Jamf.

stp-ip:
Context: My time and therefore activity is quite low currently, but I am trying to stay up to date on issues as always and am happy to help out with the steering committee.

Therefore I am open to running for a committee member seat. I am "employed" by Okkur Labs and Rekkur Solutions.

yongtang:
If no other maintainers expresses interest I am open to running for a committee member seat as independent.

Open source activities and opinions are my own, and the company I currently work for (ivanti) is not involved. The company has no seat or involvement in CNCF, and has no business interest or activity in DNS area. Note while running as an independent (not representing a company), happy to step aside if needed so the restriction of no company taking more than one seat still applies.

SuperQ:
[...]

PR #5655, opened 2022-09-26
Update CODEOWNERS: Add steering committee members

chrisohaver:

This adds the maintainers expressing an interest in being on the steering committee per
https://github.com/coredns/coredns/issues/5614

I don't know everyone's emails, so I just left that out. I don't think it's a big deal, but if anyone finds this to be a problem, we can collect them and add them in.

All members added here are employed by different companies.  Should we also add "company" to the list?  If so, would all the members added in this PR please let me know if you consider your contributions as independent from your employer's interests.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,9 +1,9 @@
 # 5 steering committee members
-# steering committee member: <github handle>, <email>, term ends <term end date>
-# steering committee member: <github handle>, <email>, term ends <term end date>
-# steering committee member: <github handle>, <email>, term ends <term end date>
-# steering committee member: <github handle>, <email>, term ends <term end date>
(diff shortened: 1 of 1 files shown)


Tantalor93:
I dont think we need to explicitly write company names into Codeowners (this can be quite prone to changes).

Regarding the contributions, most of my contributions are done independent of my employer interest. Although some of the contributions were done after observing the behaviour of the CoreDNS in the customized solution of my Employer.

Tantalor93 (approved):

johnbelamaric (approved):
LGTM, +1

SuperQ (approved):

bradbeam (approved):

rajansandeep (approved):

PR #5657, opened 2022-09-30
plugin/k8s_external: Fix rcode for headless services

Miciah:
-------

The `TestExternal` test for the kubernetes plugin had a seemingly erroneous return statement that prevented the test from verifying any but the first test case.  This PR removes that bogus return statement.

The test cases also had several typos, which this PR corrects.

Fixing `TestExternal` revealed that `External` was returning the wrong rcode in some cases.  Instead of returning NOERROR, `External` was returning NXDOMAIN for headless services.  This PR corrects that error.

https://cloud-native.slack.com/archives/C4DF7FP71/p1664570787814109

Possibly a release note.

Not unless someone is depending on CoreDNS to return NXDOMAIN when resolving headless services.

### DIFF
--- a/plugin/kubernetes/external.go
+++ b/plugin/kubernetes/external.go
@@ -116,6 +116,7 @@ func (k *Kubernetes) External(state request.Request, headless bool) ([]msg.Servi
 							if !(matchPortAndProtocol(port, p.Name, protocol, p.Protocol)) {
 								continue
 							}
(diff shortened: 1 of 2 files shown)


chrisohaver (approved):
Thanks! LGTM

chrisohaver:
FYI @TomasKohout

TomasKohout:
Thanks! 🙂

PR #5658, opened 2022-10-02
plugin/file: making qname search more efficient

slayer1366:
When a list of zones in file plugin becomes larger than two or three hundreds coredns  performance is seriously degrading. Benchmarks show that for two thousands zones coredns could processed at most 8Krps and all 32 CPU cores are all utilized. 


slayer1366:
Try later with signed commit

PR #5661, opened 2022-10-02
plugin/file: make a qname search more efficient

slayer1366:
Signed-off-by: Oleg Gorokhov <slayer@yandex-team.ru>

When a list of zones in file plugin becomes larger (e.g. two or three hundreds zones) coredns performance is seriously degrading. Benchmarks show that coredns could processed two thousands zones in with file plugin at most 8Krps and all 32 CPU cores are utilized at 100%

### DIFF
--- a/plugin/file/file.go
+++ b/plugin/file/file.go
@@ -36,16 +36,35 @@ func (f File) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (i
 	state := request.Request{W: w, Req: r}
 
 	qname := state.Name()
-	// TODO(miek): match the qname better in the map
-	zone := plugin.Zones(f.Zones.Names).Matches(qname)
-	if zone == "" {
-		return plugin.NextOrFailure(f.Name(), f.Next, ctx, w, r)
-	}
 
-	z, ok := f.Zones.Z[zone]
-	if !ok || z == nil {
-		return dns.RcodeServerFailure, nil
-	}
+        var z *Zone
+
+        zone := ""
+        subdomain := qname
+        off := 0
+        end := false
+
+        for {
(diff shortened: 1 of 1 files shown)


chrisohaver:
Thanks!

Perhaps we should also add a benchmark that tests lookup performance on a zone file with 2-3 hundred records? The current benchmark tests on a small domain of less than 20 records (BenchmarkFileLookup).

chrisohaver:
> Perhaps we should also add a benchmark that tests lookup performance on a zone file with 2-3 hundred records? The current benchmark tests on a small domain of less than 20 records (BenchmarkFileLookup).

Ah. Correction, this optimization is about the number of zones, not records. So the test would instead increase number of zones, not records in the zone.

plugin/file/file.go
```diff
@@ -36,16 +36,35 @@ func (f File) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (i
 	state := request.Request{W: w, Req: r}
 
 	qname := state.Name()
-	// TODO(miek): match the qname better in the map
-	zone := plugin.Zones(f.Zones.Names).Matches(qname)
```
chrisohaver, line 40: Should this optimization be implemented in `Matches` instead?

slayer1366: I will try to make another PR fixing tests failed, and implement search as a function. But It could not be done as Matches - as we use map of *Zone and the last one is defined in file plugin inside.

chrisohaver:
Please update this PR instead of opening a new one.  This keeps all comments in one place.

slayer1366:
Closing w.r.t https://github.com/coredns/coredns/pull/5670

Issue #5616, opened 2022-09-13
Add a "weighted-round-robin" strategy to the loadbalance plugin


gabordozsa:
**What would you like to be added**:

Extend the *loadbalance* plugin with a new "weighted-round-robin" strategy.

**Why is this needed**:

The *loadbalance* plugin only provides a random shuffle of records currently.

Sometimes more control over the load distribution is required (e.g. hierarchical load balancing among non-uniform resources). Weighted-round-robin is a simple enough deterministic strategy where the weight value assigned to an address defines how many times that particular address will be returned as top entry in successive answers.

chrisohaver:
> where the weight value assigned to an address

Can you give an example of how/where a weight would be assigned to an address?

gabordozsa:
I would suggest the following extension to the *loadbalance* plugin. 

Example server block in Corefile:
```
demo.plmt {
        loadbalance weighted_round_robin ./demo.plmt.weights {
                reload 10s
        } 
        file ./demo.plt {
                reload 10s
        }
}
```
where `./demo.plmt.weights` is a file containing the weight values assigned to the addresses.
E.g., contents of `./demo.plmt.weights`:
```
web1-2.demo.plmt
200.64.1.1   4
200.64.1.2   3
200.64.1.3   2
```
where the first line is the domain name and the following three lines assign different weight values (4, 3 and 2) to the different addresses.

chrisohaver:
What should the behavior be if `web1-2.demo.plmt` starts returning IPs that do not have a weight defined?

gabordozsa:
Then the answer is not modified. We would check if the next expected "top" IP is part of the address list. If it is present then we would move it to the top (first) position. If it is not present then we would return the address records unmodified.

gabordozsa:
I have opened a PR https://github.com/coredns/coredns/pull/5662 to continue the discussion.

gabordozsa:
The follow up PR  https://github.com/coredns/coredns/pull/5662 is in Draft mode (mainly due to missing tests) but  I would appreciate any early feedback on the approach, code structure or the general concept behind the PR. Thanks!

gabordozsa:
The follow up PR https://github.com/coredns/coredns/pull/5662 is ready for review.

PR #5662, opened 2022-10-03
Add weighted-round-robin policy to loadbalance plugin

gabordozsa:
Weighted round robin policy gives more control over the load distribution then random shuffle. Weight values assigned to IPs determine the relative likelihood of an IP being returned as top record in the answer.

Signed-off-by: Gabor Dozsa <gabor.dozsa@ibm.com>

This PR extends the **loadbalance** plugin with a new "weighted-round-robin" strategy.  

The loadbalance plugin only provides a random shuffle of records currently. Sometimes more control over the load distribution is required (e.g. hierarchical load balancing among non-uniform resources). Weighted-round-robin is a simple enough strategy where the weight value assigned to an address defines the relative likelihood of an IP being returned as top record in the answer.

https://github.com/coredns/coredns/issues/5616

The PR proposes changes to the README of the **loadbalance** plugin.

No

### DIFF
--- a/plugin/loadbalance/README.md
+++ b/plugin/loadbalance/README.md
@@ -30,6 +30,30 @@ returned in the answer.
 
  * **DURATION** interval to reload `WEIGHTFILE` and update weight assignments if there are changes in the file. The default value is `30s`. A value of `0s` means to not scan for changes and reload.
 
+
+## Weightfile
+
+The generic weight file syntax:
+
+~~~
+# Comment lines are ignored
+
+domain-name1
+ip11 weight11
+ip12 weight12
+ip13 weight13
+
+domain-name2
+ip21 weight21
+ip22 weight22
+# ... etc.
+~~~
+
+where `ipXY` is an IP address for `domain-nameX` and `weightXY` is the weight value associated with that IP. The weight values are in the range of [1,255].
+
+The `weighted` policy selects one of the address record in the result list and moves it to the top (first) position in the list. The random selection takes into account the weight values assigned to the addresses in the weight file. If an address in the result list is associated with no weight value in the weight file then the default weight value "1" is assumed for it when the selection is performed.
+
+
 ## Examples
 
 Load balance replies coming back from Google Public DNS:
@@ -64,22 +88,3 @@ www.example.com
 100.64.1.3 2
 ~~~
 
-The generic weight file syntax:
-
-~~~
-# Comment lines are ignored
-
-domain-name1
-ip11 weight11
-ip12 weight12
-ip13 weight13
-
-domain-name2
-ip21 weight21
-ip22 weight22
-# ... etc.
-~~~
-
-where `ipXY` is an IP address for `domain-nameX` and `weightXY` is the weight value associated with that IP. The weight values are in the range of [1,255].
-
-The `weighted` policy selects one of the address record in the result list and moves it to the top (first) position in the list. The random selection takes into account the weight values assigned to the addresses in the weight file. If an address in the result list is associated with no weight value in the weight file then the default weight value "1" is assumed for it when the selection is performed.
--- a/plugin/loadbalance/README.md
+++ b/plugin/loadbalance/README.md
@@ -29,13 +29,13 @@ returned in the answer.
 
 
 ~~~
-loadbalance weighted_round_robin WEIGHTFILE
+loadbalance weighted WEIGHTFILE
 ~~~
 
 * **WEIGHTFILE** is the file containing the weight values assigned to IPs for various domain names. If the path is relative, the path from the **root** plugin will be prepended to it. The format is explained below in the *Examples* section.
 
 ~~~
-loadbalance weighted_round_robin WEIGHTFILE {
+loadbalance weighted WEIGHTFILE {
 			reload DURATION
 }
 ~~~
@@ -60,7 +60,7 @@ example.com {
         file ./db.example.com {
                 reload 10s
         }
-        loadbalance weighted_round_robin ./db.example.com.weights {
+        loadbalance weighted ./db.example.com.weights {
                     reload 10s
         }
 }
--- a/plugin/loadbalance/README.md
+++ b/plugin/loadbalance/README.md
@@ -62,14 +62,10 @@ More (optional) control for the "weighted_round_robin" policy:
 ~~~
 loadbalance weighted_round_robin WEIGHTFILE {
 			reload DURATION
-			[deterministic]
 }
 ~~~
 
 * **reload** the interval to reload **WEIGHTFILE** and update weight assignments if there are changes in the file. The default value is **30s**. A value of **0s** means to not scan for changes and reload.
-
-* **deterministic** switch to deterministic weighted-round-robin strategy. This remove randomness from answer re-ordering. A weight value defines exactly how many
-times a particular IP should be in the top record for consecutive queries to the same domain name.
 
 
 ## Examples
--- a/plugin/loadbalance/README.md
+++ b/plugin/loadbalance/README.md
@@ -23,9 +23,9 @@ loadbalance [POLICY]
 
 The "round_robin" strategy randomizes the order of  A, AAAA, and MX records applying a uniform probability distribution.
 
-The "weighted_round_robin" policy assigns weight values to IPs to control the relative likelihood of particular IPs to be returned as top
-entry (A or AAAA record) in the answer. See [Wikipedia](https://en.wikipedia.org/wiki/Weighted_round_robin) about a generic
-description of weighted round robin load balancing strategy.
+The "weighted_round_robin" policy assigns weight values to IPs to control the relative likelihood of particular IPs to be returned as the first
+(top) A/AAAA record in the answer. Note that it does not shuffle all the records in the answer, it is only concerned about the first A/AAAA record
+returned in the answer.
 
 Additional option required by "weighted_round_robin" policy:
 
@@ -35,9 +35,27 @@ loadbalance weighted_round_robin WEIGHTFILE
 
 * **WEIGHTFILE** the weight file containing the weight values assigned to IPs. If the path is relative, the path from the **root** plugin will be prepended to it.
 
-The weight file is parsed line-by-line. If the first character in the line is '#' then the line is ignored (i.e. comment line). Otherwise, if the line contains a single word then it is a domain name. If the line contains two words then the first one is an IP and the second one is a weight value for that IP. The IPs are addresses for the domain name which is defined above them in the file. Multiple domain names can be specified in the same weight file. A weight value must be in the range [1,255].
+Expected weight file syntax:
 
-If the server receives a query for which no domain name is specified in the corresponding weight file then the answer is returned unmodified. If the domain name is specified in the weight file but the answer does not include the expected top IP (or there isn't any IP/weight pair specified for the particular domain name) then the answer is returned unmodified.
+~~~
+# Comment lines are ignored
+
+domain-name1
+ip11 weight11
+ip12 weight12
+ip13 weight13
+
+domain-name2
+ip21 weight21
+ip22 weight22
+# ... etc.
+~~~
+
+where *ipXY* is an IP address for *domain-nameX* and *weightXY* is the weight value associated with that IP. The weight values are in the range of [1,255].
+
+For each new result list, the weighted-round-robin policy determines the "next expected top IP" from the weighted IPs list (defined in the weight file). Afterwards, it tries to move that particular IP to the first (top) position of the result list.
+
+If the server receives a query for which no domain name is specified in the corresponding weight file then the answer is returned unmodified. If the domain name is specified in the weight file but the answer does not include the expected top IP (or there isn't any IP/weight pair specified for the particular domain name) then the answer is returned unmodified, too.
 
 More (optional) control for the "weighted_round_robin" policy:
 
--- a/plugin/loadbalance/README.md
+++ b/plugin/loadbalance/README.md
@@ -19,7 +19,40 @@ implementations (like glibc) are particular about that.
 loadbalance [POLICY]
 ~~~
 
-* **POLICY** is how to balance. The default, and only option, is "round_robin".
+* **POLICY** is how to balance. Available options are "round_robin" and "weighted_round_robin". The default is "round_robin".
+
+The "round_robin" strategy randomizes the order of  A, AAAA, and MX records applying a uniform probability distribution.
+
+The "weighted_round_robin" policy assigns weight values to IPs to control the relative likelihood of particular IPs to be returned as top
+entry (A or AAAA record) in the answer. See [Wikipedia](https://en.wikipedia.org/wiki/Weighted_round_robin) about a generic
+description of weighted round robin load balancing strategy.
+
+Additional option required by "weighted_round_robin" policy:
+
+~~~
+loadbalance weighted_round_robin WEIGHTFILE
+~~~
+
+* **WEIGHTFILE** the weight file containing the weight values assigned to IPs. If the path is relative, the path from the **root** plugin will be prepended to it.
+
+The weight file is parsed line-by-line. If the first character in the line is '#' then the line is ignored (i.e. comment line). Otherwise, if the line contains a single word then it is a domain name. If the line contains two words then the first one is an IP and the second one is a weight value for that IP. The IPs are addresses for the domain name which is defined above them in the file. Multiple domain names can be specified in the same weight file. A weight value must be in the range [1,255].
+
+If the server receives a query for which no domain name is specified in the corresponding weight file then the answer is returned unmodified. If the domain name is specified in the weight file but the answer does not include the expected top IP (or there isn't any IP/weight pair specified for the particular domain name) then the answer is returned unmodified.
+
+More (optional) control for the "weighted_round_robin" policy:
+
+~~~
+loadbalance weighted_round_robin WEIGHTFILE {
+			reload DURATION
+			[deterministic]
+}
+~~~
+
+* **reload** the interval to reload **WEIGHTFILE** and update weight assignments if there are changes in the file. The default value is **30s**. A value of **0s** means to not scan for changes and reload.
+
+* **deterministic** switch to deterministic weighted-round-robin strategy. This remove randomness from answer re-ordering. A weight value defines exactly how many
+times a particular IP should be in the top record for consecutive queries to the same domain name.
+
 
 ## Examples
 
@@ -31,3 +64,29 @@ Load balance replies coming back from Google Public DNS:
     forward . 8.8.8.8 8.8.4.4
 }
 ~~~
+
+Use weighted round robin strategy to load balance replies defined by the **file** plugin:
+
+~~~ corefile
+demo.plmt {
+        file ./demo.plmt {
+                reload 10s
+        }
+        loadbalance weighted_round_robin ./demo.plt.weights {
+                    reload 10s
+        }
+}
+~~~
+
+where the weight file **./demo.plt.weights** contains:
+
+~~~
+wwww.demo.plmt
+100.64.1.1 3
+100.64.1.2 1
+100.64.1.3 2
+~~~
+
+This assigns weight vales **3**, **1** and **2** to the IPs **100.64.1.1**, **100.64.1.2** and **100.64.1.3**, respectively. These IPs are addresses in A records for the domain name "wwww.demo.plmt". (The IPs for the A records are defined in the ./demo.plmt file using the **file** plugin in this example).
+
+In this example, the ratio between the number of answers in which **100.64.1.1**, **100.64.1.2** or **100.64.1.3** is in the top A record should converge to  **3 : 1 : 2**.  (E.g. there should be twice as many answers with **100.64.1.3** in the top A record than with **100.64.1.2**).
--- a/plugin/loadbalance/README.md
+++ b/plugin/loadbalance/README.md
@@ -53,9 +53,7 @@ ip22 weight22
 
 where *ipXY* is an IP address for *domain-nameX* and *weightXY* is the weight value associated with that IP. The weight values are in the range of [1,255].
 
-For each new result list, the weighted-round-robin policy determines the "next expected top IP" from the weighted IPs list (defined in the weight file). Afterwards, it tries to move that particular IP to the first (top) position of the result list.
-
-If the server receives a query for which no domain name is specified in the corresponding weight file then the answer is returned unmodified. If the domain name is specified in the weight file but the answer does not include the expected top IP (or there isn't any IP/weight pair specified for the particular domain name) then the answer is returned unmodified, too.
+The weighted-round-robin policy selects one of the address record in the result list and moves it to the top (first) position in the list. The random selection takes into account the weight values assigned to the addresses in the weight file. If an address in the result list is associated with no weight value in the weight file then the default weight value "1" is assumed for it when the selection is performed.
 
 More (optional) control for the "weighted_round_robin" policy:
 
--- a/plugin/loadbalance/README.md
+++ b/plugin/loadbalance/README.md
@@ -26,7 +26,7 @@ loadbalance [round_robin | weighted WEIGHTFILE] {
 (top) A/AAAA record in the answer. Note that it does not shuffle all the records in the answer, it is only concerned about the first A/AAAA record
 returned in the answer.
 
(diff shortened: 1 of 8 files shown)


chrisohaver:
Thanks for your submission.

I presume that multiple names can exist in the weight file; This is not clear in the documentation.
Documentation should also clearly describe what happens when:
* a name is queried that is not in the file.
* a name is queried that is in the file, but the file does not define a weight for an IP address returned by the upstream.
* a name is queried that is in the file, but the file defines weights for IP addresses that are not in the answer from the upstream.

plugin/loadbalance/setup.go
```diff
@@ -10,34 +14,105 @@ import (
 )
 
 var log = clog.NewWithPlugin("loadbalance")
+var openErr = errors.New("Weight file open error")
```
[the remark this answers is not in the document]
gabordozsa: Ack

gabordozsa:
> Thanks for your submission.
> 

Thank you for the quick feedback!

> I presume that multiple names can exist in the weight file; This is not clear in the documentation. 

That's right, multiple names can be specified in the same weight file. I have added that info to the README. 

Documentation should also clearly describe what happens when:
> 
> * a name is queried that is not in the file.
> * a name is queried that is in the file, but the file does not define a weight for an IP address returned by the upstream.
> * a name is queried that is in the file, but the file defines weights for IP addresses that are not in the answer from the upstream.

I have added more explanation about these cases, too, to the README.

plugin/loadbalance/README.md
```diff
@@ -19,7 +19,40 @@ implementations (like glibc) are particular about that.
...
+
+~~~
+loadbalance weighted_round_robin WEIGHTFILE
+~~~
+
+* **WEIGHTFILE** the weight file containing the weight values assigned to IPs. If the path is relative, the path from the **root** plugin will be prepended to it.
+
+The weight file is parsed line-by-line. If the first character in the line is '#' then the line is ignored (i.e. comment line). Otherwise, if the line contains a single word then it is a domain name. If the line contains two words then the first one is an IP and the second one is a weight value for that IP. The IPs are addresses for the domain name which is defined above them in the file. Multiple domain names can be specified in the same weight file. A weight value must be in the range [1,255].
+
+If the server receives a query for which no domain name is specified in the corresponding weight file then the answer is returned unmodified. If the domain name is specified in the weight file but the answer does not include the expected top IP (or there isn't any IP/weight pair specified for the particular domain name) then the answer is returned unmodified.
```
chrisohaver: > If the domain name is specified in the weight file but the answer does not include the expected top IP ...

I don't understand what "the expected top IP" means here?   Do you mean "If the domain name is specified in the weight file but the list of IPs in the answer does not the same as the list of weighted IPs in the weight file for that domain ... "

plugin/loadbalance/README.md
```diff
@@ -19,7 +19,40 @@ implementations (like glibc) are particular about that.
...
+
+Additional option required by "weighted_round_robin" policy:
+
+~~~
+loadbalance weighted_round_robin WEIGHTFILE
+~~~
+
+* **WEIGHTFILE** the weight file containing the weight values assigned to IPs. If the path is relative, the path from the **root** plugin will be prepended to it.
+
+The weight file is parsed line-by-line. If the first character in the line is '#' then the line is ignored (i.e. comment line). Otherwise, if the line contains a single word then it is a domain name. If the line contains two words then the first one is an IP and the second one is a weight value for that IP. The IPs are addresses for the domain name which is defined above them in the file. Multiple domain names can be specified in the same weight file. A weight value must be in the range [1,255].
```
chrisohaver: I think this is probably describing verbatim the parsing algorithm... I think it would be clearer and shorter to just describe the expected syntax of the file, perhaps with an example file.

plugin/loadbalance/README.md
```diff
@@ -19,7 +19,40 @@ implementations (like glibc) are particular about that.
 loadbalance [POLICY]
 ~~~
 
-* **POLICY** is how to balance. The default, and only option, is "round_robin".
+* **POLICY** is how to balance. Available options are "round_robin" and "weighted_round_robin". The default is "round_robin".
+
+The "round_robin" strategy randomizes the order of  A, AAAA, and MX records applying a uniform probability distribution.
+
+The "weighted_round_robin" policy assigns weight values to IPs to control the relative likelihood of particular IPs to be returned as top
+entry (A or AAAA record) in the answer. See [Wikipedia](https://en.wikipedia.org/wiki/Weighted_round_robin) about a generic
```
chrisohaver: I think this reference is probably too generic and algorithmy to be helpful to someone who could not infer what weighted-round-robin is by its name.

plugin/loadbalance/weighted.go
```diff
@@ -0,0 +1,299 @@
...
+		index int
+		count uint8
+	}
+	// Randomized weighted-round-robin
+	randomizedWRR struct {
+		wsum uint
+	}
+)
+
+func (d *determininsticWRR) nextTopIP(weights []*weightItem, rn *rand.Rand) net.IP {
```
chrisohaver: I'm not certain this is needed. While in theory, a deterministic implementation is different than random, IMO in the context of a server that typically serves a large number of requests, it is statistically/virtually identical to random.

plugin/loadbalance/handler.go
```diff
@@ -3,21 +3,54 @@ package loadbalance
...
+	topIP interface {
+		nextTopIP(weights []*weightItem, rn *rand.Rand) net.IP
+	}
+)
 
 // ServeDNS implements the plugin.Handler interface.
 func (rr RoundRobin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
-	wrr := &RoundRobinResponseWriter{w}
-	return plugin.NextOrFailure(rr.Name(), rr.Next, ctx, wrr, r)
+	rrw := &RoundRobinResponseWriter{ResponseWriter: w, policy: rr.policy, weights: rr.weights}
```
chrisohaver: IIUC, the new weighted option only handles (or intends to handle) `A` and `AAAA` requests (see other note about SRV requests).  If so, we should not be using the RoundRobinResponseWriter when the query type is not `A`/`AAAA`, otherwise coredns could waste time looking through answers e.g. that won't have an A/AAAA records in them.

plugin/loadbalance/loadbalance.go
```diff
@@ -18,9 +27,14 @@ func (r *RoundRobinResponseWriter) WriteMsg(res *dns.Msg) error {
...
-	res.Answer = roundRobin(res.Answer)
-	res.Ns = roundRobin(res.Ns)
-	res.Extra = roundRobin(res.Extra)
+	switch r.policy {
+	case ramdomShufflePolicy:
+		res.Answer = roundRobin(res.Answer)
+		res.Ns = roundRobin(res.Ns)
+		res.Extra = roundRobin(res.Extra)
+	case weightedRoundRobinPolicy:
+		res.Answer = r.weights.weightedRoundRobin(res.Question[0].Name, res.Answer)
```
chrisohaver: I think the weighted shuffle implementation in its current state only shuffles answers to A/AAAA requests.  But if you want to shuffle responses to SRV requests too, for which any given SRV records can reference one _or more_ A/AAAA records in the `Extra` (aka Additional) section, you should also randomize Extra.  I think this would make the feature more complete - handling all (I think all) query types that can answer with A/AAAA records.

chrisohaver (changes requested):
Thanks for contributing!  I've made some comments.

plugin/loadbalance/README.md
gabordozsa: Not really. The weighted-round-robin policy does not shuffle all the elements of the result lists (as opposed to the default round-robin). It only tries to move a particular IP to the first position in the list.  So, whenever it gets a new result list it does the following two steps: 

1.  it computes the "next expected IP" from the weighted IPs list 
2.  it checks if that particular IP is in the result list. If so, then it moves that IP to the first (top) position in the list.

The rest of the result list is not modified (i.e. not shuffled). The weights only determine how likely an IP will be returned as the first item in the result list.

plugin/loadbalance/weighted.go
gabordozsa: I agree. I only added this for easier testing/demonstration of the behaviour (i.e. we can check the behaviour by running only a few queries instead of 100s or 1000s of queries). If you think that it is not necessary then I will remove it.

plugin/loadbalance/handler.go
gabordozsa: Right, I can change that.

But maybe I should change the entire weighted-round-robin approach. It could be an extra addition to the default round-robin policy in the sense that it would run default round-robin shuffle first always. After that, it would try to change the first item in the A/AAAA record lists (only) to honour the weights assigned to IPs. This way, it would handle the same requests as default round-robin does (with some extra tweaks for the A/AAAA records only). Would that be a better approach?

chrisohaver: On the surface that would seem to create extra work when processing the result for no benefit?

gabordozsa: Right, if the user is only interested in first A/AAAA record in the result list then it would not add any benefit. The only "benefit" would be a possible simpler implementation logic for the weighted-round-robin and more familiar results for users who are already familiar with the default round-robin. 

So I will stick with the current approach.  I'm going change the PR to handle only requests which have results with A/AAAA records in them.

plugin/loadbalance/loadbalance.go
gabordozsa: Right, I will add code to apply the logic to the Extra section, too.

plugin/loadbalance/README.md
gabordozsa: Agree. I will remove it.

plugin/loadbalance/README.md
gabordozsa: Okay, I will do that.

gabordozsa: Done.

plugin/loadbalance/README.md
gabordozsa: Done.

plugin/loadbalance/handler.go
gabordozsa: Done.

plugin/loadbalance/loadbalance.go
gabordozsa: Done.

plugin/loadbalance/weighted.go
gabordozsa: It is used in the unit tests, too, to verify expected results. (As opposed to the default round-robin policy which does not have any unit test to verify the expected order of the randomized result records.)

gabordozsa:
I accidentally removed the review request from @miekg by re-requesting review from @chrisohaver. I cannot revert that unfortunately but I still hope for as many reviews as possible.

plugin/loadbalance/README.md
chrisohaver: Thanks.  IMO the wording in the readme doesn't clearly capture that.

The behavior seems counterintuitive - that results will be sometimes weighted and sometimes not for the same query name and answer set.

It would seem more intuitive to only return the answer unaltered if the answer contains a record that is not defined in the weight file.  Otherwise, the answer records all have defined weights, and one can be selected to be the top/first record in the response.

An alternate approach would be to only return the answer unaltered if the answer does not contain any records defined in the weight file.  Otherwise, only records with defined weights are eligible to be selected for the top/first record in the response.  Additionally, when returning the answer "unaltered", it would probably be better to randomly select a top IP, as if all the records have equal weight (which in this case they do in a sense).

Yet another approach would be to define in the weight file _default_ weights to be used for IPs that do not have a weight defined.

plugin/loadbalance/weighted.go
chrisohaver: > It is used in the unit tests, too, to verify expected results.

Technically those tests are only testing the deterministic implementation.

plugin/loadbalance/README.md
gabordozsa: It makes sense, thanks for the suggestions! I changed the implementation to assign weight for every address record in the result list. If an address does not have a weight defined in the weight file then the default weight value "1" is assumed for it.  The answer is returned unaltered only if there isn't any address record.

The default weight value (1) is hardcoded currently as I'm not sure if it is really necessary to make that configurable (via the weight file).

plugin/loadbalance/weighted.go
gabordozsa: True. I have removed the deterministic implementation.

gabordozsa:
@chrisohaver I believe I have addressed all your review comments so far. Please let me know if something is still missing (or if I need to take any action to let this PR move ahead). Thanks!

chrisohaver:
> @chrisohaver I believe I have addressed all your review comments so far. Please let me know if something is still missing (or if I need to take any action to let this PR move ahead). Thanks!

Thanks - I need to reacquaint myself with the PR. Regarding the syntax description in the README: Can you reorganize it to be more inline with other plugin READMEs (a single syntax definition followed by bulleted list of descriptions of the options)?  Perhaps move the Weightfile description into a separate section (IMO its too long to be inline with the syntax options).

plugin/loadbalance/README.md
```diff
@@ -31,3 +76,29 @@ Load balance replies coming back from Google Public DNS:
     forward . 8.8.8.8 8.8.4.4
 }
 ~~~
+
+Use weighted round robin strategy to load balance replies defined by the **file** plugin:
+
+~~~ corefile
+demo.plmt {
```
chrisohaver: Can we use `example.com` as the example zone?  I'm not sure what `plmt` and `plt` are referencing.

plugin/loadbalance/README.md
```diff
@@ -19,7 +19,52 @@ implementations (like glibc) are particular about that.
 loadbalance [POLICY]
 ~~~
 
-* **POLICY** is how to balance. The default, and only option, is "round_robin".
+* **POLICY** is how to balance. Available options are "round_robin" and "weighted_round_robin". The default is "round_robin".
+
+The "round_robin" strategy randomizes the order of  A, AAAA, and MX records applying a uniform probability distribution.
+
+The "weighted_round_robin" policy assigns weight values to IPs to control the relative likelihood of particular IPs to be returned as the first
```
chrisohaver: I think just `weighted` should suffice for the name.

plugin/loadbalance/README.md
gabordozsa: Sure, it's done.

plugin/loadbalance/README.md
gabordozsa: Done.

gabordozsa:
> > @chrisohaver I believe I have addressed all your review comments so far. Please let me know if something is still missing (or if I need to take any action to let this PR move ahead). Thanks!
> 
> Thanks - I need to reacquaint myself with the PR. Regarding the syntax description in the README: Can you reorganize it to be more inline with other plugin READMEs (a single syntax definition followed by bulleted list of descriptions of the options)? Perhaps move the Weightfile description into a separate section (IMO its too long to be inline with the syntax options).

Thanks for looking at this PR again! I have re-organized the README. Please let me know if it needs any more changes!

plugin/loadbalance/README.md
gabordozsa: Ack

plugin/loadbalance/README.md
gabordozsa: Yes, I agree. It's done.

plugin/loadbalance/loadbalance.go
```diff
@@ -18,9 +27,20 @@ func (r *RoundRobinResponseWriter) WriteMsg(res *dns.Msg) error {
 		return r.ResponseWriter.WriteMsg(res)
 	}
 
-	res.Answer = roundRobin(res.Answer)
-	res.Ns = roundRobin(res.Ns)
-	res.Extra = roundRobin(res.Extra)
+	switch r.policy {
```
chrisohaver: Instead of policy being a string that is checked for every query, can you make it a function that is defined and set during setup? e.g. a func that operates on the dns message `func(*dns.Msg) *dns.Msg`

plugin/loadbalance/handler.go
```diff
@@ -3,21 +3,50 @@ package loadbalance
...
 	"github.com/miekg/dns"
 )
 
-// RoundRobin is a plugin to rewrite responses for "load balancing".
-type RoundRobin struct {
-	Next plugin.Handler
-}
+type (
+	// RoundRobin is a plugin to rewrite responses for "load balancing".
+	RoundRobin struct {
```
chrisohaver: We should probably rename the plugin struct to something more generic.  `LoadBalance` seems more appropriate now.

plugin/loadbalance/setup.go
```diff
@@ -10,34 +13,102 @@ import (
...
 func init() { plugin.Register("loadbalance", setup) }
 
 func setup(c *caddy.Controller) error {
-	err := parse(c)
+	policy, weighted, err := parse(c)
 	if err != nil {
 		return plugin.Error("loadbalance", err)
 	}
 
+	if policy == weightedRoundRobinPolicy {
```
chrisohaver: Can you move this logic into `parse`, and have `parse` return onStartup/onShutdown functions instead of `weightedRR` ?

chrisohaver (changes requested):

plugin/loadbalance/loadbalance.go
gabordozsa: Yes, that is a nice improvement, thanks for the suggestion! It's done.

plugin/loadbalance/handler.go
gabordozsa: Done.

plugin/loadbalance/setup.go
gabordozsa: Done.

plugin/loadbalance/handler.go
```diff
@@ -10,15 +10,16 @@ import (
...
+type LoadBalance struct {
+	Next    plugin.Handler
+	shuffle func(*dns.Msg) *dns.Msg
 }
 
 // ServeDNS implements the plugin.Handler interface.
-func (rr RoundRobin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
-	wrr := &RoundRobinResponseWriter{w}
-	return plugin.NextOrFailure(rr.Name(), rr.Next, ctx, wrr, r)
+func (rr LoadBalance) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
```
chrisohaver:
```suggestion
func (lb LoadBalance) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
```

plugin/loadbalance/README.md
```diff
@@ -19,7 +19,28 @@ implementations (like glibc) are particular about that.
...
+ The `round_robin` strategy randomizes the order of  A, AAAA, and MX records applying a uniform probability distribution.
+
+ The `weighted` policy assigns weight values to IPs to control the relative likelihood of particular IPs to be returned as the first
+(top) A/AAAA record in the answer. Note that it does not shuffle all the records in the answer, it is only concerned about the first A/AAAA record
+returned in the answer.
+
+
+~~~
+loadbalance weighted_round_robin WEIGHTFILE
+~~~
```
chrisohaver: Please remove this and consolidate it into with the first syntax definition (immediately under **Syntax** title).

plugin/loadbalance/README.md
```diff
@@ -19,7 +19,28 @@ implementations (like glibc) are particular about that.
...
+
+* **WEIGHTFILE** is the file containing the weight values assigned to IPs for various domain names. If the path is relative, the path from the **root** plugin will be prepended to it. The format is explained below in the *Examples* section.
+
+~~~
+loadbalance weighted_round_robin WEIGHTFILE {
+			reload DURATION
+}
+~~~
```
chrisohaver: Please remove this and consolidate it into with the first syntax definition (immediately under **Syntax** title).

plugin/loadbalance/README.md
```diff
@@ -31,3 +52,49 @@ Load balance replies coming back from Google Public DNS:
...
+www.example.com
+100.64.1.1 3
+100.64.1.2 1
+100.64.1.3 2
+~~~
+
+This assigns weight vales `3`, `1` and `2` to the IPs `100.64.1.1`, `100.64.1.2` and `100.64.1.3`, respectively. These IPs are addresses in A records for the domain name `www.example.com`. (The IPs for the A records are defined in the `./db.example.com` file using the **file** plugin in this example).
+
+In this example, the ratio between the number of answers in which `100.64.1.1`, `100.64.1.2` or `100.64.1.3` is in the top (first) A record should converge to  `3 : 1 : 2`.  (E.g. there should be twice as many answers with `100.64.1.3` in the top A record than with `100.64.1.2`).
```
chrisohaver: I think this description of what the example accomplishes should be stated before the example Corefile and weight file (i.e. moved to line 56 and maybe rephrased a bit).

chrisohaver (changes requested):
Thanks for being patient.  Ive added some comments on readme.

plugin/loadbalance/handler.go
gabordozsa: Done.

plugin/loadbalance/README.md
gabordozsa: Done.

plugin/loadbalance/README.md
gabordozsa: Done.

plugin/loadbalance/README.md
gabordozsa: Done.

plugin/loadbalance/handler.go
```diff
@@ -10,15 +10,16 @@ import (
...
-	wrr := &RoundRobinResponseWriter{w}
-	return plugin.NextOrFailure(rr.Name(), rr.Next, ctx, wrr, r)
+func (lb LoadBalance) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	rw := &LoadBalanceResponseWriter{ResponseWriter: w, shuffle: lb.shuffle}
+	return plugin.NextOrFailure(lb.Name(), lb.Next, ctx, rw, r)
 }
 
 // Name implements the Handler interface.
-func (rr RoundRobin) Name() string { return "loadbalance" }
+func (rr LoadBalance) Name() string { return "loadbalance" }
```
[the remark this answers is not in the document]
gabordozsa: Done

plugin/loadbalance/README.md
```diff
@@ -31,3 +40,46 @@ Load balance replies coming back from Google Public DNS:
...
+
+domain-name2
+ip21 weight21
+ip22 weight22
+# ... etc.
+~~~
+
+where `ipXY` is an IP address for `domain-nameX` and `weightXY` is the weight value associated with that IP. The weight values are in the range of [1,255].
+
+The `weighted` policy selects one of the address record in the result list and moves it to the top (first) position in the list. The random selection takes into account the weight values assigned to the addresses in the weight file. If an address in the result list is associated with no weight value in the weight file then the default weight value "1" is assumed for it when the selection is performed.
```
chrisohaver: Please move this general description of the weight file structure to a separate section `## Weightfile`, before  the Examples section.

gabordozsa: Done.

plugin/loadbalance/README.md
```diff
@@ -16,10 +16,43 @@ implementations (like glibc) are particular about that.
...
+			reload DURATION
+}
+~~~
+* `round_robin` policy randomizes the order of  A, AAAA, and MX records applying a uniform probability distribution. This is the default load balancing policy.
+
+* `weighted` policy assigns weight values to IPs to control the relative likelihood of particular IPs to be returned as the first
+(top) A/AAAA record in the answer. Note that it does not shuffle all the records in the answer, it is only concerned about the first A/AAAA record
+returned in the answer.
+
+ * **WEIGHTFILE** is the file containing the weight values assigned to IPs for various domain names. If the path is relative, the path from the **root** plugin will be prepended to it. The format is explained below in the *Examples* section.
```
chrisohaver:
```suggestion
 * **WEIGHTFILE** is the file containing the weight values assigned to IPs for various domain names. If the path is relative, the path from the **root** plugin will be prepended to it. The format is explained below in the *Weightfile* section.
```

gabordozsa: Ops, I missed that one. Done.

gabordozsa:
@chrisohaver please let me know if any further change is needed to move this PR forward. Thanks!

chrisohaver (approved):

PR #5670, opened 2022-10-04
plugin/file: make a qname search more efficient

slayer1366:
Signed-off-by: Oleg Gorokhov <slayer@yandex-team.ru>

When a list of zones in file plugin becomes larger (e.g. two or three hundreds zones) coredns performance is seriously degrading. Benchmarks show that coredns could processed two thousands zones in with file plugin at most 8Krps and all 32 CPU cores are utilized at 100%

### DIFF
--- a/plugin/file/file.go
+++ b/plugin/file/file.go
@@ -31,7 +31,7 @@ type (
 	}
 )
 
-// Making qname search in map
+// Matches implements qname search in a map.
 func (f File) Matches(qname string) (*Zone, string) {
         var z *Zone
         zone := ""
--- a/plugin/file/file.go
+++ b/plugin/file/file.go
@@ -31,36 +31,36 @@ type (
 	}
 )
 
-// Making qname search in map
+// Matches implements qname search in a map.
 func (f File) Matches(qname string) (*Zone, string) {
-        var z *Zone
-        zone := ""
+	var z *Zone
+	zone := ""
 
-        subdomain := qname
-        off := 0
-        end := false
+	subdomain := qname
+	off := 0
+	end := false
 
-        for {
-                if y, ok := f.Zones.Z[subdomain]; ok {
-                        z = y
-                        zone = subdomain
-                        break
-                }
+	for {
+		if y, ok := f.Zones.Z[subdomain]; ok {
+			z = y
+			zone = subdomain
+			break
+		}
 
-                if subdomain == "." {
-                        break
-                }
+		if subdomain == "." {
+			break
+		}
 
-                off, end = dns.NextLabel(qname, off)
-                if end {
-                        // Last dot should also be checked
-                        subdomain = "."
-                        continue
-                }
-                subdomain = qname[off:len(qname)]
-        }
+		off, end = dns.NextLabel(qname, off)
+		if end {
+			// Last dot should also be checked
+			subdomain = "."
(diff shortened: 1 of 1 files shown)


chrisohaver:
This is the 3rd PR opened for this issue.  Please make further updates to this PR instead of closing and opening a new PR for every change.

plugin/file/file.go
```diff
@@ -31,16 +31,53 @@ type (
...
+
+        z, zone := f.Matches(qname)
+
+        if zone == "" {
+                return plugin.NextOrFailure(f.Name(), f.Next, ctx, w, r)
+        }
+
+        if z == nil {
+                return dns.RcodeServerFailure, nil
+        }
```
Tantalor93: wrong formatting, please use tabs

plugin/file/file.go
```diff
@@ -31,16 +31,53 @@ type (
...
 	state := request.Request{W: w, Req: r}
 
 	qname := state.Name()
-	// TODO(miek): match the qname better in the map
-	zone := plugin.Zones(f.Zones.Names).Matches(qname)
-	if zone == "" {
-		return plugin.NextOrFailure(f.Name(), f.Next, ctx, w, r)
-	}
+
+        z, zone := f.Matches(qname)
```
Tantalor93: `z` is not used for anything, it is overwritten on line 82

plugin/file/file.go
```diff
@@ -31,16 +31,53 @@ type (
 	}
 )
 
+// Matches implements qname search in a map.
+func (f File) Matches(qname string) (*Zone, string) {
```
Tantalor93, line 35: can you write a benchmark test? Without it, there is no data to actually prove that your solution is actually better. See documentation how to write a benchmark test https://pkg.go.dev/testing#hdr-Benchmarks

plugin/file/file.go
```diff
@@ -31,16 +31,53 @@ type (
...
+                if end {
+                        // Last dot should also be checked
+                        subdomain = "."
+                        continue
+                }
+                subdomain = qname[off:len(qname)]
+        }
+
+        return z, zone
+}
```
Tantalor93, line 64: use tabs instead of spaces for consistency :)

plugin/file/file.go
Tantalor93: Also what I think @chrisohaver meant by this comment https://github.com/coredns/coredns/pull/5661#discussion_r985813914 is whether we can actually try to improve this method https://github.com/coredns/coredns/blob/master/plugin/normalize.go#L25 performance to transitively improve all other plugins using this matching?

Tantalor93 (changes requested):

Tantalor93:
also, please sign your commits https://github.com/coredns/coredns/pull/5670/checks?check_run_id=8708024371 🙂

plugin/file/file.go
chrisohaver: > Also what I think @chrisohaver meant ...

I did - but I didn't notice at the time that the optimization uses a map of zones, and the current function uses a slice of zones. Migrating the existing function to use a map would be too disruptive to existing internal and external plugins (since most just store a slice of zones).

Maybe we couple add a MatchMap function, and migrate plugins over to it.

Issue #5640, opened 2022-09-20
Bufsize gets overwritten to 2048 on cache miss when logic enables DNSSEC


gcs278:
**What happened**:
As I understand, forwarded/upstream DNS requests triggered by a cache miss, will have DNSSEC enabled (DO bit) regardless if the client set it. During this logic, when the DO bit is not set and an EDNS0 record exists, an existing EDNS0 record with the bufsize as specified in the corefile, is overwritten to 2048.

This change was introduced in v1.7.1 by commit https://github.com/coredns/coredns/commit/acf9a0fa19928e605ac8ac3314890c9fef73e16b. Prior to this change, the DNS cache refresh logic used the `bufsize` in the corefile. This becomes an issue when users' networks have UDP size limitations of less than 2048.

**What you expected to happen**:
The DNS cache refresh request should use `bufsize` as specified in the corefile as it did prior to v1.7.1.

**How to reproduce it (as minimally and precisely as possible)**:
1. Deploy a CoreDNS instance >= v1.7.1 with the corefile below (`./coredns --conf corefile`)
2. `dig @127.0.0.1 google.com`
4. Log output of coredns process will show payload/bufsize of 2048 (see logs below)

**Anything else we need to know?**:

**Environment**:

- the version of CoreDNS: >= v1.7.1
- Corefile:
```
.:53 {
  bufsize 512
  forward . 127.0.0.1:8053
  cache
}

.:8053 {
  log
  forward . 8.8.8.8
}
```
- logs, if applicable:
   - Client Sets Bufsize (UDPSize) of 4096, but DNS refresh sets to 2048 despite `bufsize 512` in corefile:
 ```
[INFO] 127.0.0.1:37106 - 63758 "A IN google.com. udp 51 true 2048" NOERROR qr,rd,ra 65 0.062822559s
 ```
- OS (e.g: `cat /etc/os-release`):
```
NAME="Red Hat Enterprise Linux"
VERSION="8.4 (Ootpa)"
```
- Others:
  - I'm pretty new to CoreDNS, so I don't really understand the complexities, but a solution I had success with was removing `o.SetUDPSize(defaultUDPBufSize)` from the `setDO` function if a EDNS0 record existed, so it wouldn't be overridden:
https://github.com/coredns/coredns/blob/85370b1662491cfee647066a8b89e00fb1bef241/plugin/cache/handler.go#L160-L165

Let me know if you have any questions and/or agree with my suggested way forward.

[...]

PR #5671, opened 2022-10-04
plugin/cache: cache now uses source query DNSSEC option for upstream

gcs278:
Signed-off-by: Grant Spence <gspence@redhat.com>

The DNSSEC cache solution introduced in https://github.com/coredns/coredns/commit/acf9a0fa19928e605ac8ac3314890c9fef73e16b changed the logic to always tag upstream cache refreshes with DNSSEC requests. Though a clever low-code approach, it has introduced a couple of problems:
1. Counter intuitive behavior to an observer: The fact that all upstream queries get tagged as requesting DNSSEC can be confusing to an observer when clients are not requesting DNSSEC.
2. All upstream DNS requests get hardcoded to bufsize of 2048
3. It has been raised as a concern by users who are acutely bandwidth conscious, as it will cause unnecessary overhead if an upstream server serves DNSSEC, but clients never actually request it. 

https://github.com/coredns/coredns/issues/5640

plugin/cache/README.md

No

### DIFF
--- a/plugin/cache/dnssec.go
+++ b/plugin/cache/dnssec.go
@@ -2,35 +2,13 @@ package cache
 
 import "github.com/miekg/dns"
 
-// isDNSSEC returns true if r is a DNSSEC record. NSEC,NSEC3,DS and RRSIG/SIG
-// are DNSSEC records. DNSKEYs is not in this list on the assumption that the
-// client explicitly asked for it.
-func isDNSSEC(r dns.RR) bool {
-	switch r.Header().Rrtype {
-	case dns.TypeNSEC:
-		return true
-	case dns.TypeNSEC3:
-		return true
-	case dns.TypeDS:
-		return true
-	case dns.TypeRRSIG:
-		return true
-	case dns.TypeSIG:
-		return true
-	}
-	return false
-}
-
-// filterRRSlice filters rrs and removes DNSSEC RRs when do is false. In the returned slice
-// the TTLs are set to ttl. If dup is true the RRs in rrs are _copied_ into the slice that is
+// filterRRSlice filters out OPT RRs, and sets all RR TTLs to ttl.
+// If dup is true the RRs in rrs are _copied_ into the slice that is
 // returned.
-func filterRRSlice(rrs []dns.RR, ttl uint32, do, dup bool) []dns.RR {
+func filterRRSlice(rrs []dns.RR, ttl uint32, dup bool) []dns.RR {
 	j := 0
 	rs := make([]dns.RR, len(rrs))
 	for _, r := range rrs {
-		if !do && isDNSSEC(r) {
-			continue
-		}
 		if r.Header().Rrtype == dns.TypeOPT {
 			continue
 		}
--- a/plugin/cache/README.md
+++ b/plugin/cache/README.md
@@ -10,8 +10,7 @@ With *cache* enabled, all records except zone transfers and metadata records wil
 3600s. Caching is mostly useful in a scenario when fetching data from the backend (upstream,
 database, etc.) is expensive.
 
-*Cache* will change the query to enable DNSSEC (DNSSEC OK; DO) if it passes through the plugin. If
-the client didn't request any DNSSEC (records), these are filtered out when replying.
+*Cache* will pass DNSSEC (DNSSEC OK; DO) options through the plugin for upstream queries.
 
 This plugin can only be used once per Server Block.
 
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -76,7 +76,7 @@ func New() *Cache {
 // key returns key under which we store the item, -1 will be returned if we don't store the message.
 // Currently we do not cache Truncated, errors zone transfers or dynamic update messages.
 // qname holds the already lowercased qname.
-func key(qname string, m *dns.Msg, t response.Type) (bool, uint64) {
+func key(qname string, m *dns.Msg, t response.Type, do bool) (bool, uint64) {
 	// We don't store truncated responses.
 	if m.Truncated {
 		return false, 0
@@ -86,11 +86,21 @@ func key(qname string, m *dns.Msg, t response.Type) (bool, uint64) {
 		return false, 0
 	}
 
-	return true, hash(qname, m.Question[0].Qtype)
+	return true, hash(qname, m.Question[0].Qtype, do)
 }
 
-func hash(qname string, qtype uint16) uint64 {
+var one = []byte("1")
+var zero = []byte("0")
+
+func hash(qname string, qtype uint16, do bool) uint64 {
 	h := fnv.New64()
+
+	if do {
+		h.Write(one)
+	} else {
+		h.Write(zero)
+	}
+
 	h.Write([]byte{byte(qtype >> 8)})
 	h.Write([]byte{byte(qtype)})
 	h.Write([]byte(qname))
@@ -145,6 +155,7 @@ func newPrefetchResponseWriter(server string, state request.Request, c *Cache) *
 		Cache:          c,
 		state:          state,
 		server:         server,
+		do:             state.Do(),
 		prefetch:       true,
 		remoteAddr:     addr,
 	}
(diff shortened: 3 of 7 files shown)


plugin/cache/dnssec.go
```diff
@@ -2,35 +2,13 @@ package cache
...
-// the TTLs are set to ttl. If dup is true the RRs in rrs are _copied_ into the slice that is
+// filterRRSlice sets the rrs TTLs to ttl.
+// If dup is true the RRs in rrs are _copied_ into the slice that is
 // returned.
-func filterRRSlice(rrs []dns.RR, ttl uint32, do, dup bool) []dns.RR {
+func filterRRSlice(rrs []dns.RR, ttl uint32, dup bool) []dns.RR {
 	j := 0
 	rs := make([]dns.RR, len(rrs))
 	for _, r := range rrs {
-		if !do && isDNSSEC(r) {
```
gcs278, line 31: This was a debate for me. In theory, we shouldn't be getting DNSSEC responses from upstream since the cache only requests DNSSEC when the query originally requested it. Right?

So I remove this logic as it felt unneeded. However, there may be edge cases I'm unaware of.

plugin/cache/dnssec.go
```diff
@@ -2,35 +2,13 @@ package cache
...
-		return true
-	case dns.TypeSIG:
-		return true
-	}
-	return false
-}
-
-// filterRRSlice filters rrs and removes DNSSEC RRs when do is false. In the returned slice
-// the TTLs are set to ttl. If dup is true the RRs in rrs are _copied_ into the slice that is
+// filterRRSlice sets the rrs TTLs to ttl.
```
chrisohaver:
```suggestion
// filterRRSlice filters out OPT RRs, and sets all RR TTLs to ttl.
```

plugin/cache/dnssec.go
chrisohaver: Seems correct to me.

chrisohaver (commented):
Thanks for taking this up.  I have not reviewed in detail yet.
One thing that came to mind while reviewing is prefetching.

When prefetching a cache item, we should build the prefetch request with the do bit as was in the original query that requested the cached item.  But we don't know what that was, so we should add the original do bit to the cache item.  Alternately we could infer this from the cached item's hash key by creating a hash key with do bit on and off, and seeing which one matches - but this is very hacky.

gcs278:
> When prefetching a cache item, we should build the prefetch request with the do bit as was in the original query that requested the cached item. But we don't know what that was, so we should add the original do bit to the cache item. Alternately we could infer this from the cached item's hash key by creating a hash key with do bit on and off, and seeing which one matches - but this is very hacky.

Thanks for your comments! I agree the prefetch should use the original query DO bit, but as far as I know, it should already use the original query DO bit based on the fact the cache is now keyed on the presence of the DO bit. Effectively creating unique cache entries based on the value of the DO bit. Now, unique entries based on the original query DO bit is not as memory efficient as combining them as mentioned here: https://github.com/coredns/coredns/pull/4085#issue-687952982

I've been making sure I don't regress https://github.com/coredns/coredns/issues/3836 which was the driver behind defaulting the upstream queries to always use DNSSEC. I created a bind9 server that doesn't support DNSSEC and removes the DO Bit. My testing shows that caching still works when with a query with DO bit and a response without DO bit because the cache is keyed on the query DO bit.

**So I'll outline the options:**
1. Cache the query DO bit (this PR)
  a. **Pros**: No cache issues with query DO true and response DO false (https://github.com/coredns/coredns/issues/3836) and upstream queries match client queries
  b. **Cons**: Less memory efficient. Two cache entries now based on queries with and without DO bit, regardless if server supports.
2. Cache the response DO bit (prior to https://github.com/coredns/coredns/commit/acf9a0fa19928e605ac8ac3314890c9fef73e16b)
  a. **Pros**: Memory efficient because it's based on what DO bit the server has, not queries and upstream queries match client queries
  b. **Cons**: Exposes https://github.com/coredns/coredns/issues/3836
3. Always do DNSSEC on cache refresh (current implementation, after https://github.com/coredns/coredns/commit/acf9a0fa19928e605ac8ac3314890c9fef73e16b)
  a. **Pros**: No cache issues with query DO true and response DO false (https://github.com/coredns/coredns/issues/3836) and memory efficient because we only store 1 type of response (DO bit true)
  b. **Cons**: Listed here: https://github.com/coredns/coredns/pull/5671#issue-1396944091

Is this increase in memory justified? I feel like there might be a more complex solution here, but I'm not sure complexity is worth it.

plugin/cache/dnssec.go
gcs278: Done

chrisohaver:
>  as far as I know, it should already use the original query DO bit based on the fact the cache is now keyed on the presence of the DO bit.

The value of the request DO bit is not stored anywhere except in the hash key. The cache item itself only holds the query name and type (because those are in the response), but not the request DO bit.  So when looking at an item in cache, we don't know what the request DO bit value was. Technically you could infer the DO bit value, by building the hash key with it both on and off, and seeing which one matches the item's hash key, but this is very hacky.  It would be better to add a bool value for it to the cache item struct.

In the current "always set DO bit" solution, it follows that prefetch would also always set the DO bit - so it doesn't need to know the original request value.

In the proposed solution, a prefetch with the _response_ DO bit (which we already have available in the cache item) would fail to correctly handle the case where a server was not answering DNSSEC queries with DNSSEC responses at the time it was cached, but has since started.  Therefore, we should prefetch with the _request_ DO bit.


When adding DO bit to the hash key, it must be the _request_ DO bit.  Using the response DO bit in the key would result in 100% cache miss when the request sets DO, and the response is not DNSSEC. Using the response DO bit essentially saves on cache size by not caching this case (although it is writing to cache, just not to a place that the same request would retrieve it).

Adding the DO bit to the key potentially doubles cache size (no matter if it's from the response or request).
But IMO correct answers wins over memory efficiency in this context.

gcs278:
> The value of the request DO bit is not stored anywhere except in the hash key. 

Sorry, I might be missing something, but since CoreDNS doesn't prefetch asynchronously (in the background), we have the incoming query's DO Bit that is triggering the prefetch, of which we used to match the item from the cache with the same query DO bit. So this PR does that at the moment, the prefetch query DO bit is set using the query that triggered the prefetch, and subsequently matched the cache item with the same DO Bit. I tested locally with queries with a DO Bit and an upstream DNS that removes the DO Bit. The prefetches still use the query DO bit regardless of the response.

I see your point if prefetching was asynchronous to queries. In that case, we would not know what the original query DO bit was, as it is not stored, and we don't have an incoming query object to decode the DO bit of the item in the cache.

Does this make sense? Are you saying we should still store the query DO bit in the cache, even though the query that triggered the prefetch encodes this information for us?

> Adding the DO bit to the key potentially doubles cache size (no matter if it's from the response or request).
But IMO correct answers wins over memory efficiency in this context.

Agreed :+1:

chrisohaver:
> Sorry, I might be missing something, but since CoreDNS doesn't prefetch asynchronously

Ah - I has forgotten and was thinking it was asynchronous. Not an issue then. Thanks.

gcs278:
Thanks for the draft reviews to make sure I'm on track. The PR is ready for the official review process.

One thing I noticed, regardless of the existing implementation or this new PR, is that the forward or cache plugin will always respond back with the query DO bit. So when query DO bit=true and response DO bit=false, both plugins will return DO bit=true. If that's something we want to fix, for the cache plugin, we'd actually need to store the response DO bit and return it in the response.

Pic is wireshark of master branch CoreDNS, Dig Query -> Forward Query -> Upstream Query -> Upstream Response -> Forward Response -> Dig Response. You can see the DO bit getting set to the original query DO bit in the forward/cache responses.
image

chrisohaver:
> One thing I noticed, regardless of the existing implementation or this new PR, is that the forward or cache plugin will always respond back with the query DO bit.

That seems like it would violate DNS rules.  Certainly seems like it should be fixed.

gcs278:
> That seems like it would violate DNS rules. Certainly seems like it should be fixed.

Agreed. I found the logic that is doing it for all plugins. It takes the query DO bit and overrides the response DO bit with it. The solution is just to remove this logic (which fixes forward plugin) and store the DO bit of the response in the cache:
https://github.com/coredns/coredns/blob/26aae951e76fd4d110dce382791df34246e3a34e/request/request.go#L196-L198

Seems like it's been around forever in the code a long time(since https://github.com/coredns/coredns/commit/16e50ec5f82aa3969b27f5b8fa750919ad0a115a), but it wasn't used globally until https://github.com/coredns/coredns/pull/2357. 

I'm guessing the motivation was EDNS0 compliance as stated in https://github.com/coredns/coredns/pull/2357#issue-386481100.

I feel like this is a separate issue than was is addressed in the PR, based on the fact it's been like this forever. So I can open another issue for discussion. Let me know what you think.

chrisohaver (approved):
LGTM

PR #5687, opened 2022-10-11
plugin/file: fix file leak

### LINKED ISSUE #5684: plugin/file: zones files are opened but never closed inside fileParse
Reported by [Static Major from Orijtech Inc](https://github.com/marketplace/actions/staticmajor-analyzer), we can see this 
```shell
/coredns/plugin/file/setup.go:114:6: leaking resource created on line 91
/coredns/plugin/file/setup.go:118:6: leaking resource created on line 91
/coredns/plugin/file/setup.go:126:5: leaking resource created on line 91
```

which originates from https://github.com/coredns/coredns/blob/2f24b512d3999e9687ffbfaf8f8226c64185dca1/plugin/file/setup.go#L91 for which there is a missing reader.Close() which can be achieved by putting that code in an extracted closure/function and adding a `defer reader.Close()`

Kindly FYI-ing @willpoint @jhusdero @elias-orijtech @kirbyquerby

yongtang:
This PR fixes file leak in plugin/file.

This PR fixes #5684.

/cc  @odeke-em

### DIFF
--- a/plugin/file/setup.go
+++ b/plugin/file/setup.go
@@ -92,19 +92,27 @@ func fileParse(c *caddy.Controller) (Zones, error) {
 		if err != nil {
 			openErr = err
 		}
-		defer reader.Close()
 
-		for i := range origins {
-			z[origins[i]] = NewZone(origins[i], fileName)
-			if openErr == nil {
-				reader.Seek(0, 0)
-				zone, err := Parse(reader, origins[i], fileName, 0)
-				if err != nil {
-					return Zones{}, err
+		err = func() error {
+			defer reader.Close()
+
+			for i := range origins {
+				z[origins[i]] = NewZone(origins[i], fileName)
+				if openErr == nil {
+					reader.Seek(0, 0)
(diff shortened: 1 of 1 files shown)


odeke-em (changes requested):
Thank you for the PR @yongtang! Unfortunately this change isn't correct, invoking defer in a for loop will pile up the defers and they'll only be run when the for loop exits. The proper fix as I described in the issue is to turn that into a closure so

```go

   reader, err := ...
   if err != nil {
       return ...
   }
   func() {
         defer reader.Close()
         for i := range origins {
             ...
         }
    }
}()
```

yongtang:
@odeke-em Ah that is my fault. Let me update the PR.

yongtang:
@odeke-em  PR has been updated. Please take a look.

odeke-em (approved):
Nice, thank you @yongtang and LGTM!

Tantalor93 (approved):
lgtm, thx!

Tantalor93 (approved):

PR #5706, opened 2022-10-21
plugin/file: remove un-used function

### LINKED ISSUE #5705: LLRB print_test incorrect
https://github.com/coredns/coredns/blob/6cb5e107077e2081ba13f1755ee9da34e7f6fc96/plugin/file/tree/print_test.go#L63
server3.example.com. should be server4.example.com. left sub-tree
but code show server3.example.com. is server1.example.com. left sub-tree

chrisohaver:

Removes an unused and buggy test function.

#4959 #4962 #5705


Tantalor93:
is the problem only with the test? Isn't the whole `Tree.Print`  wrong? Shouldn't we remove the tested function also? If the function `Tree.Print` works as expected, maybe we should instead aim to fix the test?

chrisohaver:
> is the problem only with the test? Isn't the whole `Tree.Print` wrong? Shouldn't we remove the tested function also? If the function `Tree.Print` works as expected, maybe we should instead aim to fix the test?

I intended to also remove `Tree.Print`. and `Node.print`, but accidentally left it out of the PR.

As far as I can tell, these functions are not used anywhere, and they don't implement any interfaces.

Issue #5703, opened 2022-10-18
dot is broken in some cases


sanyo0714:
**What happened**:
dot is broken in the case of 【kdig version 3.2.1+】 and 【coredns 1.8.5+ build at golang 1.18.1+】

doh&dot use same [TLSConfig]  with corefile 
```
https://.:8443 tls://.:8080 {
	tls plugin/tls/test_cert.pem  plugin/tls/test_key.pem  plugin/tls/test_ca.pem
}
```
doh set tlsConfig.NextProtos = []string{"h2", "http/1.1"}
dot also has tlsConfig.NextProtos = []string{"h2", "http/1.1"}
```
the code @core/dnsserver/register.go line 150
c.TLSConfig = c.firstConfigInBlock.TLSConfig
```

**What you expected to happen**:
with 【coredns 1.8.5+ build at golang 1.18.1+】 and 【kdig version 3.2.1+】 the dot is work

**How to reproduce it (as minimally and precisely as possible)**:

【coredns 1.8.5+ build at golang 1.18.1+】 and 【kdig version 3.2.1+】 , with corefile 

```
https://.:8443 tls://.:8080 {
	tls plugin/tls/test_cert.pem  plugin/tls/test_key.pem  plugin/tls/test_ca.pem
}
```

**Anything else we need to know?**:

fix the bug with
```
@core/dnsserver/register.go line 150
c.TLSConfig = c.firstConfigInBlock.TLSConfig.Clone()
```
**Environment**:

- the version of CoreDNS: 1.8.5+
- Corefile:
```
https://.:8443 tls://.:8080 {
	tls plugin/tls/test_cert.pem  plugin/tls/test_key.pem  plugin/tls/test_ca.pem
}
```
- logs, if applicable:

![image](https://user-images.githubusercontent.com/14814461/196366199-da1a41ff-c3c5-45a5-b611-2edd933e287d.png)

- Others:
tools: 
kdig (Knot DNS), version 3.2.1
![image](https://user-images.githubusercontent.com/14814461/196914286-763a5a19-9ff0-476e-b81a-2fe7a0a6f62f.png)

golang 1.18.1+
@ src/crypto/tls/handshake_server.go line 220 & line 281
```
func (hs *serverHandshakeState) processClientHello() error {
 ...
      selectedProto, err := negotiateALPN(c.config.NextProtos, hs.clientHello.alpnProtocols)
      if err != nil {
          c.sendAlert(alertNoApplicationProtocol)
          return err
      }
[...]

PR #5710, opened 2022-10-24
Fork TLSConfig for each encrypted connection

sanyo0714:
Signed-off-by: sanyo <yeshengan.ysa@alibaba-inc.com>

Fork TLSConfig for each encrypted connection

#5703

### DIFF
--- a/test/tls_test.go
+++ b/test/tls_test.go
@@ -8,41 +8,68 @@ import (
 	"github.com/miekg/dns"
 )
 
-func TestDNSoverTLS(t *testing.T) {
-	dot, doh := ":1053", ":8443"
-	corefile := `tls://.%s https://.%s {
+func TestTLS(t *testing.T) {
+	tempCorefile := `%s {
         tls ../plugin/tls/test_cert.pem ../plugin/tls/test_key.pem
         whoami
     }`
-	qname := "example.com."
-	qtype := dns.TypeA
+
+	dot, doh := ":1053", ":8443"
+	m := new(dns.Msg)
+	m.SetQuestion("example.com.", dns.TypeA)
 	answerLength := 0
 
-	ex, _, _, err := CoreDNSServerAndPorts(fmt.Sprintf(corefile, dot, doh))
-	if err != nil {
-		t.Fatalf("Could not get CoreDNS serving instance: %s", err)
+	tests := []struct {
+		server string
+		client dns.Client
+	}{
+		{fmt.Sprintf("tls://.%s", dot),
+			dns.Client{
+				Net:       "tcp-tls",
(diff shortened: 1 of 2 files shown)


chrisohaver:
Can you add a test?

sanyo0714:
> Can you add a test?


Can I add the test case this way?
```
caddy.Quiet = true
dnsserver.Quiet = true

corefile := `
https://.:8443 tls://.:8080 {
tls test_cert.pem test_key.pem test_ca.pem
}`
server, err := caddy.Start(caddy.CaddyfileInput{
    Contents:       []byte(corefile),
    Filepath:       "config",
    ServerTypeName: "dns",
})
if err != nil {
    t.Errorf("Failed to start server for input %q:\n%s", err, corefile)
}

serItemSet := make(map[string]struct{}, 2)
for _, s := range server.Servers() {
    sv := reflect.ValueOf(s)
    if sv.CanInterface() {
        ser := sv.FieldByName("server").Elem()
        switch ser.Type().String() {

        // ServerTLS format &{0xc00029e540 0xc00041ec00}
        // ServerHTTPS format &{0xc00029e5a0 0xc0004422a0 0xc00012b110 0xc00041ec00 0x170e6a0}
        // check that all items of ServerTLS differ from ServerHTTPS
        case "*dnsserver.ServerTLS", "*dnsserver.ServerHTTPS":
            ser := fmt.Sprintf("%v", ser)
            for _, item := range strings.Split(ser[2:len(ser)-1], " ") {
                if _, ok := serItemSet[item]; ok {
                    log.Fatalf("dnsserver.ServerTLS,dnsserver.ServerHTTPS has same item:[%s]", item)
                } else {
                    serItemSet[item] = struct{}{}
                }
            }
        }
    }
}
server.Stop()
```

chrisohaver:
> > Can you add a test?
> 
> Can I add the test case this way?
> ...

That suggested test case would verify that your solution is in place, but doesn't really verify that the solution resolves the problem ("dot is broken").

A better test would be a test like the ones in `/test` that runs the server and sends it test queries that verify that dot functions as expected in the given configuration.

test/tls_test.go
```diff
@@ -2,21 +2,23 @@ package test
 
 import (
 	"crypto/tls"
+	"fmt"
 	"testing"
 
 	"github.com/miekg/dns"
 )
 
 func TestDNSoverTLS(t *testing.T) {
```
chrisohaver: Can you rename this test to `TestTLS`, and make it execute two test cases: the existing unaltered dot case, and the new dot+doh case?

chrisohaver (commented):
Thanks for finding and fixing this. The test case looks good, one minor comment.

chrisohaver (approved):

PR #5722, opened 2022-10-26
plugin/acl: adding ability to drop queries

rsclarke:

Both block and filter actions write responses to the client based upon the source IP address of the UDP packet containing the query.  An attacker spoofing the source IP address to that of their target, can elicit a response to be sent to the victim host, known as DNS Reflection.  If an attacker is able to elicit a large response from a relatively small query, with a spoofed source IP address, they are able to increase the amount of data sent to the victim, known as DNS Amplification.  Scaling this from one to many queries allows an attacker to perform an effective Denial of Service (DoS) attack against their target.

Adding the drop action enables CoreDNS to ignore queries of a given type or network range from being processed and a response written, where an operator knows ahead of time, should not originate or be destined to.

Updated acl plugin documentation to describe the `drop` action and the difference from `block` and `filter` along with example configuration.  A new metric is also listed for the total number of dropped requests, an indicator when significant number of queries are ignored.

No

### DIFF
--- a/plugin/acl/acl_test.go
+++ b/plugin/acl/acl_test.go
@@ -51,6 +51,7 @@ func TestACLServeDNS(t *testing.T) {
 		wantRcode             int
 		wantErr               bool
 		wantExtendedErrorCode uint16
+		expectNoResponse      bool
 	}{
 		// IPv4 tests.
 		{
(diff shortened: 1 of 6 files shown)


plugin/acl/acl_test.go
```diff
@@ -205,6 +206,79 @@ func TestACLServeDNS(t *testing.T) {
...
+				qtype:    dns.TypeA,
+			},
+			wantRcode:             dns.RcodeRefused,
+			wantExtendedErrorCode: dns.ExtendedErrorCodeBlocked,
+		},
+		{
+			name: "Subnet-Order 2 DROPPED",
+			config: `acl example.org {
+				drop net 192.168.0.0/16
+				block net 192.168.2.0/24
```
ihac: I feel `block net 192.168.1.0/24` would be better here considering you used `192.168.1.1` as the source IP.

ihac (commented):
Overall LGTM. One minor suggestion regarding the test case.

plugin/acl/acl_test.go
rsclarke: Thanks and I agree, updated to `192.168.1.0/24`.

ihac (approved):
The change itself looks good to me. Will let more members to comment on its necessity.

Thanks for the contribution.

chrisohaver:
The ability to drop requests without a response is a useful and valid option for mitigating UDP reflection attacks.
I believe there is already a way to accomplish dropping responses destined to a specific network in CoreDNS (1.10) with the _view_ and _erratic_ plugins. for example:

```
. {
  view example {
    expr incidr(client_ip(), '10.0.0.0/24')
  }
  erratic {
    drop 1       # drop 1 in every 1 queries, i.e. all queries
  }
}

. {
  forward . 8.8.8.8
}
```

Regarding the reflection attack mitigation specifically:
You could also restrict to just udp: `expr incidr(client_ip(), '10.0.0.0/24') && proto() == 'udp'`
Also, there is the external plugin [rrl](https://github.com/coredns/rrl) that is designed to protect against reflection attacks.

chrisohaver:
> I believe there is already a way to accomplish dropping responses destined to a specific network in CoreDNS (1.10) with the view and erratic plugins.

That said, the proposed _acl_ syntax is less verbose.  Also since _acl_ is more purpose-built for this case (less generic/flexible than _views_), it might accomplish the task in less operations per request (I have not tested this).

Functionally however, I think anything _acl_ can currently do can also be accomplished with the new _view_ plugin in conjunction with other plugins (e.g. _template_ to render REFUSED or NODATA responses).

Tantalor93 (approved):
The feature as it is makes sense to me, although as @chrisohaver said same effect can be achieved using other combination of plugins having this feature as part of acl plugin for simple usage makes sense as usage of views is quite advanced feature, so I would vote for merging this PR unless @chrisohaver is against it

chrisohaver:
This is such a small amount of code, I think it's ok to have it as a redundant feature.

Tantalor93:
Thanks for contributing!

PR #5737, opened 2022-11-03
plugin/kubernetes: error NXDOMAIN for TXT lookups

### LINKED ISSUE #5736: plugin/kubernetes should return an NXDOMAIN for missing TXT queries
**What happened**:
TXT queries are incorrectly returning with an empty NOERROR response rather than NXDOMAIN. Beyond just being incorrect this matters because as a valid response it means that DNS clients won't try additional domains in their search path to hopefully find the right thing.

CoreDNS should return an NXDOMAIN as the record doesn't exist

```
[]$ export CLUSTERDOMAIN=prod.cluster.local
[]$ dig foobarbaz.svc.$CLUSTERDOMAIN TXT @192.168.240.10 | grep status
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 45013
```

Expected:
```
[]$ dig foo.google.com TXT @8.8.8.8 | grep status
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 53366
```

This works properly for SRV and A records (anything the Kubernetes plugin actually supports):
```
[]$ dig foo.svc.$CLUSTERDOMAIN SRV @192.168.240.10 | grep status
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 58698

[]$ dig foo.svc.$CLUSTERDOMAIN A @192.168.240.10 | grep status
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 10092
```

**Environment**:

- the version of CoreDNS: 1.9.1
- Corefile:
```
.:53 {
    log
    errors
    health
    ready
    kubernetes prod.cluster.local
    forward . 8.8.8.8
    prometheus :9153
    cache 10
    loop
    reload
    loadbalance
}

```
- logs, if applicable:
- Others:

LAMRobinson:
All non "dns-version" TXT queries should return a NXDOMAIN rather than NOERROR

#5736

### DIFF
--- a/plugin/kubernetes/handler_test.go
+++ b/plugin/kubernetes/handler_test.go
@@ -249,7 +249,7 @@ var dnsTestCases = []kubeTestCase{
 	}},
 	// A TXT record does not exist but another record for the same FQDN does
 	{Case: test.Case{
-		Qname: "kubernetes.default.svc.cluster.local.", Qtype: dns.TypeTXT,
+		Qname: "svc1.testns.svc.cluster.local.", Qtype: dns.TypeTXT,
 		Rcode:  dns.RcodeSuccess,
 		Answer: []dns.RR{},
 	}},
--- a/plugin/kubernetes/handler_test.go
+++ b/plugin/kubernetes/handler_test.go
@@ -251,9 +251,7 @@ var dnsTestCases = []kubeTestCase{
 	{Case: test.Case{
 		Qname: "kubernetes.default.svc.cluster.local.", Qtype: dns.TypeTXT,
 		Rcode:  dns.RcodeSuccess,
-		Answer: []dns.RR{
-			test.SOA("cluster.local.	5	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1499347823 7200 1800 86400 5"),
-        },
+		Answer: []dns.RR{},
 	}},
 	// A TXT record does not exist and neither does another record for the same FQDN
 	{Case: test.Case{
--- a/plugin/kubernetes/handler_test.go
+++ b/plugin/kubernetes/handler_test.go
@@ -250,7 +250,7 @@ var dnsTestCases = []kubeTestCase{
 	// A TXT record does not exist but another record for the same FQDN does
 	{Case: test.Case{
 		Qname: "kubernetes.default.svc.cluster.local.", Qtype: dns.TypeTXT,
-		Rcode: dns.RcodeSuccess,
+		Rcode:  dns.RcodeSuccess,
 		Answer: []dns.RR{},
 	}},
 	// A TXT record does not exist and neither does another record for the same FQDN
--- a/plugin/kubernetes/handler_test.go
+++ b/plugin/kubernetes/handler_test.go
@@ -250,7 +250,7 @@ var dnsTestCases = []kubeTestCase{
 	// A TXT record does not exist but another record for the same FQDN does
 	{Case: test.Case{
 		Qname: "svc1.testns.svc.cluster.local.", Qtype: dns.TypeTXT,
-		Rcode:  dns.RcodeSuccess,
+		Rcode: dns.RcodeSuccess,
 		Ns: []dns.RR{
 			test.SOA("cluster.local.	5	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1499347823 7200 1800 86400 5"),
 		},
--- a/plugin/kubernetes/handler_test.go
+++ b/plugin/kubernetes/handler_test.go
@@ -247,6 +247,20 @@ var dnsTestCases = []kubeTestCase{
 			test.TXT("dns-version.cluster.local 28800 IN TXT 1.1.0"),
 		},
 	}},
+	// A TXT record does not exist but another record for the same FQDN does
+	{Case: test.Case{
+		Qname: "kubernetes.default.svc.cluster.local.", Qtype: dns.TypeTXT,
+		Rcode: dns.RcodeSuccess,
+		Answer: []dns.RR{},
+	}},
+	// A TXT record does not exist and neither does another record for the same FQDN
+	{Case: test.Case{
+		Qname: "svc0.svc-nons.svc.cluster.local.", Qtype: dns.TypeTXT,
+		Rcode: dns.RcodeNameError,
+		Ns: []dns.RR{
+			test.SOA("cluster.local.	5	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1499347823 7200 1800 86400 5"),
+		},
+	}},
 	// A Service (Headless) does not exist
 	{Case: test.Case{
 		Qname: "bogusendpoint.hdls1.testns.svc.cluster.local.", Qtype: dns.TypeA,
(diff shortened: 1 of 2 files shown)


plugin/kubernetes/kubernetes.go
```diff
@@ -102,10 +102,10 @@ func (k *Kubernetes) Services(ctx context.Context, state request.Request, exact
 
 		segs := dns.SplitDomainName(t)
 		if len(segs) != 1 {
-			return nil, nil
+			return nil, errInvalidRequest
```
chrisohaver: The answer should be the same as in the next `if`.
This and the following `if` can be combined into one.

plugin/kubernetes/kubernetes.go
```diff
@@ -102,10 +102,10 @@ func (k *Kubernetes) Services(ctx context.Context, state request.Request, exact
...
 		}
 		if segs[0] != "dns-version" {
-			return nil, nil
+			return nil, errNoItems
```
chrisohaver: Strictly speaking, in the case that the TXT query is for a name that exists for another record type, the answer should be "NODATA".  This is expected DNS behavior.

So, before returning `errNoItems` (which results in NXDOMAIN), this code would need to check to see if a service with the name exists.  If one does exist, then the error returned should be `nil` (which results in NODATA).

chrisohaver (changes requested):
Thanks! Comments in line. Please also add test cases.

plugin/kubernetes/kubernetes.go
LAMRobinson: Good point - I've got this working:
```
$ dig foo.svc.prod.cluster.local TXT @192.168.240.10 | grep status
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 49658

$ dig kubernetes.default.svc.prod.cluster.local TXT @192.168.240.10 | grep status
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 15694
```

chrisohaver (commented):
Looks good.  Can you sign the commits?

plugin/kubernetes/kubernetes.go
LAMRobinson: I ended up restructuring this a bit, rather than checking for a negative (make sure this isn't a query for "dns-version"), we now check if it is a query for "dns-version" and serve that if so, otherwise return NXDOMAIN

LAMRobinson:
OK should be good to go, let me know !

```
$ dig kubernetes.default.svc.prod.cluster.local TXT @192.168.240.10 | grep status
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 60508

$ dig foo.svc.prod.cluster.local TXT @192.168.240.10 | grep status
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 13919

$ dig dns-version.prod.cluster.local TXT @192.168.240.10 | grep version
; <<>> DiG 9.11.4-P2-RedHat-9.11.4-26.P2.el7_9.9 <<>> dns-version.prod.cluster.local TXT @10.229.2.32
;dns-version.prod.cluster.local.        IN      TXT
dns-version.prod.cluster.local. 10 IN   TXT     "1.1.0"
```

plugin/kubernetes/kubernetes.go
```diff
@@ -110,7 +110,7 @@ func (k *Kubernetes) Services(ctx context.Context, state request.Request, exact
 
 		// Hard code the only valid TXT - "dns-version.<zone>"
```
chrisohaver: I think this check should happen before the `Records` check, so the more expensive Records check is only performed for TXT queries that do not match the static dns-version record.

chrisohaver (commented):
Thanks. Please also fix the reviewdog whitespace notifications, imbedded in-line in Files Changed.

plugin/kubernetes/handler_test.go
```diff
@@ -251,7 +251,9 @@ var dnsTestCases = []kubeTestCase{
 	{Case: test.Case{
 		Qname: "kubernetes.default.svc.cluster.local.", Qtype: dns.TypeTXT,
 		Rcode:  dns.RcodeSuccess,
-		Answer: []dns.RR{},
+		Answer: []dns.RR{
+			test.SOA("cluster.local.	5	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1499347823 7200 1800 86400 5"),
+        },
```
chrisohaver: This doesn't look right.

For a NODATA response, the SOA record should be in the Authority section (`Ns`).

plugin/kubernetes/kubernetes.go
LAMRobinson: good idea, done

plugin/kubernetes/handler_test.go
LAMRobinson: yeah I expected as such, but was trying to troubleshoot why a test was failing - ill revert

LAMRobinson:
@chrisohaver can you give me a pointer for identifying this failing test? don't seem to be able to identify it in the source

```
--- FAIL: TestServeDNS (0.00s)
    handler_test.go:449: Test 24, rcode is "NXDOMAIN", expected "NOERROR"
FAIL
```

chrisohaver:
> @chrisohaver can you give me a pointer for identifying this failing test? don't seem to be able to identify it in the source
> 
> ```
> --- FAIL: TestServeDNS (0.00s)
>     handler_test.go:449: Test 24, rcode is "NXDOMAIN", expected "NOERROR"
> FAIL
> ```

The 24th test case in handler_test.go TestServeDNS is the first one you added in the PR.

```
	// A TXT record does not exist but another record for the same FQDN does
 	{Case: test.Case{
 		Qname: "kubernetes.default.svc.cluster.local.", Qtype: dns.TypeTXT,
 		Rcode:  dns.RcodeSuccess,
 		Answer: []dns.RR{},
 	}},
```

The service `kubernetes.default.svc.cluster.local.` probably doesn't exist in the set of test data, so NXDOMAIN is expected.  You'll want to replace this query name with a name of a service that exists in the test data (e.g. from the first test case: `svc1.testns.svc.cluster.local.`).

chrisohaver:
Regarding the new failure.  Note that NXDOMAIN responses are expected to have an authority section.  There should be some examples in other NXDOMAIN test cases.

Edit:  The same goes for "NODATA" responses (NOERROR + empty answer section) as in the failing test case - the SOA in the Authority section is expected for NODATA as well.  The SOA should be the same as in an NXDOMAIN response.

existing example: 

``` go
	// AAAA Service (with an existing A record, but no AAAA record)
	{Case: test.Case{
		Qname: "svc1.testns.svc.cluster.local.", Qtype: dns.TypeAAAA,
		Rcode: dns.RcodeSuccess,
		Ns: []dns.RR{
			test.SOA("cluster.local.	5	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1499347823 7200 1800 86400 5"),
		},
	}},
```

chrisohaver (changes requested):
Looks good.  Please fix line 253 per the reviewdog comment in handler_test.go, and it should be good to merge.

LAMRobinson:
Fab ! Sorry for the review churn here, really appreciate your help

chrisohaver (approved):

PR #5750, opened 2022-11-12
plugin/etcd: fix NS records when PathPrefix contains multiple path components

fuhry:
The `plugin/etcd/msg.Domain()` function currently operates on the assumption that the prefix always contains a single path component: `/foo` or `foo`, not `foo/bar`. However, nothing prohibits the etcd plugin from being configured with a PathPrefix like `/skydns/foo`.

When the `etcd` plugin is configured this way, NS queries expose all of the path components after the first one. So a prefix of `/skydns/foo` and a zone of `bar.local` results in an NS record at `/skydns/foo/local/bar/dns/ns` and returns `ns.dns.bar.local.foo.` as the nameserver, where the correct answer would be `ns.dns.bar.local.`.

To handle these cases, we add a new `RName` property to the `Service` struct, which allows the plugin to override the name that would be generated by `msg.Domain()`. The method `.Domain()` on that struct returns the `RName` property if it's non-empty, and otherwise calls `msg.Domain()` to invoke the previous default behavior.

At present, the etcd plugin is the only plugin that has been updated to override RName. Other plugins may need this fixed as well; a new DomainWithPrefix public helper function is exported from `msg` to fit this use case.

The existing `TestOtherLookup` test case has been modified to verify the new etcd plugin behavior.

Signed-Off-By: Dan Fuhry <dan@fuhry.com>

This pull request is needed because the etcd plugin currently behaves incorrectly under a valid configuration.

It fixes the incorrect behavior by allowing plugins to override the record name.

A quick search for "etcd" in currently open issues shows no results.

None.

No.

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -243,7 +243,7 @@ func SRV(ctx context.Context, b ServiceBackend, zone string, state request.Reque
 
 		case dns.TypeA, dns.TypeAAAA:
 			addr := serv.Host
-			serv.Host = msg.Domain(serv.Key)
+			serv.Host = serv.Domain()
 			srv := serv.NewSRV(state.QName(), weight)
 
 			if ok := isDuplicate(dup, srv.Target, "", srv.Port); !ok {
@@ -309,7 +309,7 @@ func MX(ctx context.Context, b ServiceBackend, zone string, state request.Reques
 
 		case dns.TypeA, dns.TypeAAAA:
 			addr := serv.Host
-			serv.Host = msg.Domain(serv.Key)
+			serv.Host = serv.Domain()
 			mx := serv.NewMX(state.QName())
 
 			if ok := isDuplicate(dup, mx.Mx, "", mx.Preference); !ok {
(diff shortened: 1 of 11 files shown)


fuhry:
@chrisohaver @yongtang Sorry for the ping - just highlighting a couple of names listed as reviewers on this PR who also have recent commits in master. Would anyone be able to take a look at this? Thank you!

chrisohaver:
@nitisht, please take a look.

Primary changes here are to the _etcd_ plugin. The changes to other plugins are just due to refactoring of etcd package's public methods (hence the long list of reviewers - automatically requested per the CODEOWNERS file).

fuhry:
Hi @nitisht, friendly ping on this PR, thank you!

fuhry:
@chrisohaver @yongtang Would someone be able to take a look at this? Thanks!

chrisohaver:
Looks like we need a new volunteer to own this plugin.

chrisohaver (commented):
Would another shorter way of addressing this be to strip the prefix from the keys when retrieving services from etcd, and adjust Domain to assume no prefix is present?

PR #5752, opened 2022-11-13
docs : add running go binary to readme

gabrielwere:
Signed-off-by: gabrielwere <weregabriel91@gmail.com>

The pull request is a minor change to the Readme.MD file including instructions on how to run the go binary file and how to handle the default port(53) not working.I believe it could be helpful to beginners like myself .

No issues

The pull request includes changes to the Readme.MD

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,11 @@ $ docker run --rm -i -t -v $PWD:/v -w /v golang:1.18 make
 
 The above command alone will have `coredns` binary generated.
 
+## Running the binary
+
+Run `go run coredns.go`.\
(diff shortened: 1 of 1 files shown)


chrisohaver (changes requested):
The dns.port flag is already described in the example section that follows. IMO, the better/recommended way of controlling the port is via the corefile, not via the flag.

Also instead of `go run`, one should use `make` to build the binary (as in the preceding section) and then execute the resulting binary directly (`./coredns`).

gabrielwere:
Thanks for the feedback!
Will use the corefile instead of the flag from now on.

PR #5762, opened 2022-11-15
[DoH] Make protocol parsing more flexible

sebdah:
This change avoid the hard coding of HTTPS as the protocol, which will e.g. make testing in external libraries easier and it will allow for additional flexibility in whether HTTP or HTTPS should be used for the call.

`https://` will still be the default protocol and thus the change is backwards compatible.

The change is also updating the test to be a table based test with more scenarios and also utilizing Go sub tests for nicer test output.

None.

None.

No.

### DIFF
--- a/plugin/pkg/doh/doh.go
+++ b/plugin/pkg/doh/doh.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 
 	"github.com/miekg/dns"
 )
@@ -16,18 +17,30 @@ const MimeType = "application/dns-message"
 // Path is the URL path that should be used.
 const Path = "/dns-query"
 
-// NewRequest returns a new DoH request given a method, URL (without any paths, so exclude /dns-query) and dns.Msg.
+// NewRequest returns a new DoH request given a HTTP method, URL and dns.Msg.
+//
+// The URL should not have a path, so please exclude /dns-query. The URL will
+// be prefixed with https:// by default, unless it's already prefixed with
+// either http:// or https://.
 func NewRequest(method, url string, m *dns.Msg) (*http.Request, error) {
 	buf, err := m.Pack()
 	if err != nil {
 		return nil, err
 	}
 
+	if !strings.HasPrefix(url, "http://") && !strings.HasPrefix(url, "https://") {
+		url = fmt.Sprintf("https://%s", url)
+	}
+
 	switch method {
 	case http.MethodGet:
 		b64 := base64.RawURLEncoding.EncodeToString(buf)
 
-		req, err := http.NewRequest(http.MethodGet, "https://"+url+Path+"?dns="+b64, nil)
+		req, err := http.NewRequest(
+			http.MethodGet,
+			fmt.Sprintf("%s%s?dns=%s", url, Path, b64),
+			nil,
+		)
 		if err != nil {
 			return req, err
 		}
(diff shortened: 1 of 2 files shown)


chrisohaver:
Thanks. Can you share the project you are working on that requires this change?

sebdah:
@chrisohaver I can unfortunately not share the code of the project that will be using this. But the scenario is that in the normal case (production code) we'll be using `https://` as the protocol, of course. However, when using `httptest.NewServer()` for setting up a server to mock the DoH response, it'll be using HTTP. 

So this change is needed to help with the latter case.

sebdah:
I see that one test failed because of a mismatch in my email signature, I'll fix that a little later today. Need to step out for vaccination :).

plugin/pkg/doh/doh.go
```diff
@@ -23,11 +24,19 @@ func NewRequest(method, url string, m *dns.Msg) (*http.Request, error) {
 		return nil, err
 	}
 
+	if !strings.HasPrefix(url, "http://") && !strings.HasPrefix(url, "https://") {
+		url = fmt.Sprintf("https://%s", url)
+	}
+
```
mariuskimmina, line 34: The comment on the functions says: "NewRequest returns a new DoH request given a method, URL (without any paths, so exclude /dns-query) and dns.Msg."

This wouldn't be exactly true anymore since the request might also use HTTP?

sebdah: I've updated the comment to be a little more elaborate in my fixup commit 6c8d8119f20077129f57cf04e11993d826011a95.

mariuskimmina:
Thanks for updating the comment, it still bugs me a bit that it says "returns a DoH request". DoH is defined as DNS over HTTPS, so it should not be possible for this to return an HTTP request from my understanding. At the same time I can see why this might be useful to you. 
Thoughts @chrisohaver ?

chrisohaver:
If only required for unit tests, the URL of the request can be modified directly in the request during the test.  IOW, this could probably be resolved without modifying "production" code.

sebdah:
@mariuskimmina Thanks for your comment. I don't quite follow on this:

> Thanks for updating the comment, it still bugs me a bit that it says "returns a DoH request".

The method does return an `*http.Request` both before and after this change. I'm certainly happy to change the comment if you have something in mind, though :).

sebdah:
@chrisohaver You are correct in that I could mock this thing away and mock the `*http.Request` (while it's considerably more complex in my case, because this request creation happens rather deep in the call structure).

However, I still think this change is valuable in the sense that it's more resilient for HTTPS users, while also giving flexibility to HTTP users. Previously requests would fail if the `url` was prefixed with `https://` as we'd then prepend it with `https://`, yielding a URL like `https://https://localhost:443` or such.

chrisohaver:
The request/client portions of this package aren't used anywhere in coredns.  They appear to be a partially completed/abandoned effort to add doh to the _forward_ plugin (perhaps inherited from an older implementation of the since deprecated _proxy_ plugin). They probably should have been removed when that effort was abandoned.

`miekg/dns` would seem to be a better home for this package.  Although searching for "doh" in miekg/dns yields that the maintainers are opposed to supporting doh in that library (4 years ago).

That said. I'm OK with these changes.

mariuskimmina:
I don't know enough about these parts of the code base so if it's okay for @chrisohaver then I don't mind it either. Just to elaborate what I meant to @sebdah, the method returns an `*http.request` but before it was not possible for this request to use HTTP, with this change it becomes possible, so now this method might return an DNS over HTTPS (DoH) request but it might also return an DNS over HTTP (?) request - which felt a bit odd to me.

chrisohaver:
> I don't know enough about these parts

I don't either really. It seems to be "cruft" left over from a development effort abandoned 4 years ago.

sebdah:
Yeah, I agree that e.g. `miekg/dns` could be a more suitable place for this piece of code to live. But given there seems to be little interest in supporting DoH there, I'd really appreciate if we could merge this in here ❤️.

chrisohaver (approved):

chrisohaver:
I believe this can merge if the commits are signed.

sebdah:
Thanks @chrisohaver, I have signed the commit!

PR #5773, opened 2022-11-22
plugin/dnstap: Fix behavior when multiple dnstap plugins specified

chrisohaver:

Prior to this fix, when more than one dnstap plugin was configured, the last instance would overwrite others instead of establishing multiple taps. With this fix, each instance of dnstap remains independent, permitting for example sending dnstap to multiple endpoints.

This PR makes no changes to the dnstap plugin core logic, it just fixes to the setup functions to properly the link multiple instances in the plugin chain.


Tantalor93 (approved):
LGTM! thanks!

Issue #5530, opened 2022-07-21
feat: add opentelemetry plugin


dongjiang1989:
**What would you like to be added**:

[opentelemetry](https://opentelemetry.io/) as an observability framework for cloud-native software is ♨️.

**Why is this needed**:

CoreDNS as a network centre handler, it's observability is critical, It will be helpful for CoreDns‘s simplicity connect with open-telemetry.

[Golang SDK](https://github.com/open-telemetry/opentelemetry-go)

Tantalor93:
To add some more missing reasoning:

currently, for observability (tracing) purposes, there is an existing [trace](https://coredns.io/plugins/trace/) plugin, which is built on top of [Opentracing](https://opentracing.io/), but Opentracing as a project was deprecated and it is recommended to migrate to a new standard which is a [Opentelemetry](https://opentelemetry.io/docs/migration/opentracing/). Overall I think it makes sense to introduce a new plugin built on top of Opentelemetry and deprecate `trace` plugin.

Marking as a nice candidate for a first contribution to the project 👍 If you are up to it, you can pick it up @dongjiang1989

dongjiang1989:
@Tantalor93  Thanks for your reply.  I want to pick up this feat.

gkum99:
Hi @Tantalor93 i am new to open source and I would like to work on this issue.

Tantalor93:
@2022H1030014G as there is not any activity on this, I think you can freely work on this and when you have some skeleton ready, just open a new draft PR, where we can talk on the details

75asu:
Hi @Tantalor93 , @2022H1030014G , If there isn't any movement on this one then I would like to pick this up and requesting @Tantalor93 to provide me some resources and file location to get this done

Dentrax:
Any updates on this? Looking forward to this feature! @eminaktas

KiranSatyaRaj:
is anyone working on this issue, has any pull request been made regarding this, if not I would love to take it up. @dongjiang1989

Edit: I started working on it folks. 
Edit: This is my first time working on a technical issue, So I might take more time, if anyone wants to work on it please go ahead

Tantalor93:
It seems that @KiranSatyaRaj unassigned himself, I will assign myself to the issue and handle it as part of our company hacksgiving next week

chrisohaver:
Closing this, per [comments](https://github.com/coredns/coredns/pull/5777#issuecomment-1327205020).

PR #5777, opened 2022-11-24
Introduce a new opentelemetry plugin for observability

Tantalor93:
This PR creates a new `opentelemetry` plugin, which should replace `trace` plugin in providing tracing (Observability) capabilities, it is built on top of the new standard [OpenTelemetry](https://opentelemetry.io/docs/), which replaces [OpenTracing](https://opentracing.io/) (OpenTracing standard is now being deprecated). The new plugin supports exporting traces into Zipkin collectors and OpenTelemetry collectors

#5530

documentation of a new plugin

will deprecate `trace` plugin


.gitignore
```diff
@@ -4,3 +4,4 @@ coredns.exe
 build/
 release/
 vendor/
+.idea/
```
SuperQ: Please don't include your local editor ignores in the repo. Use [`git config --global core.excludesFile`](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/).

SuperQ (changes requested):

SuperQ:
I'm not sure I want opentelemetry in CoreDNS. I'd rather go with a lighter weight tracing library. OTel brings in a lot of excess baggage.

Perhaps this would be more appropriate as an external plugin.

plugin/opentelemetry/setup.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func opentelemetryParse(c *caddy.Controller) (*opentelemetry, error) {
+	var (
+		ot  = &opentelemetry{serviceName: defServiceName}
+		err error
+	)
+
+	for c.Next() {
+		var err error
```
mariuskimmina: Creating `var err error` once should be enough here I think.

plugin/opentelemetry/setup.go
```diff
@@ -0,0 +1,165 @@
...
+		ep = supportedProviders[epType]
+	}
+
+	if epType == "zipkin" {
+		if !strings.Contains(ep, "http") {
+			ep = "http://" + ep + "/api/v2/spans"
+		}
+	} else if epType == "otelhttp" {
+		if strings.Contains(ep, "http") || strings.Contains(ep, "/") {
+			return "", "", fmt.Errorf("invalid format of opentelemetry endpoint: '%s', valid format: 'HOST:PORT'", ep)
```
mariuskimmina: I think we should use the official `OpenTelemetry` name with uppercase O and T.

mariuskimmina (commented):
Overall this looks good to me, just some small comments. 

Should be able to try this out in the next few days and give further feedback

Tantalor93:
@SuperQ I think I dont see much of an option to have this as external plugin as it needs to access (and modify) core functionality to be able to instrument plugin chain 🤔 (this https://github.com/Tantalor93/coredns/blob/01a9635a84890289ead663de1154efd126864406/plugin/plugin.go#L82)

Tantalor93:
And I dont think that OpenTelemetry GO SDK is that heavyweight, based on go.mod changes, it basically adds few dependencies, where the biggest are zipkin and otel exporters, and as this new plugin will basically replace `trace` plugin, in future releases we could remove `trace` plugin altogether with all its OpenTracing and Datadog dependencies and make CoreDNS more lightweight. Or @SuperQ what makes you think that OpenTelemetry is too heavyweight?

SuperQ:
While it doesn't look like much, the imports bring in about 15k LoC in dependencies based on a scan of the changes to `go mod vendor`. I would be interested in how this compares to simplifying our `trace` plugin to a more dedicated tracing library like [zipkin-go](https://github.com/openzipkin/zipkin-go).

Tantalor93:
Simplifying trace plugin to depend only on zipkin-go might be a way, although OpenTelemetry as a observability standard is more extensible and would allow CoreDNS to work with a more type of collectors rather than just Zipkin (opentelemetry plugin in this state of PR is able to operate with both Zipkin and OpenTel collectors). Also adding OpenTelemetry into CoreDNS allows even third party plugins to further built and extend observability capabilities of their plugins on top of OpenTelemetry. Binding observability capabilities of CoreDNS only to one concrete technology (Zipkin) doesnt sound right to me.

SuperQ:
> Also adding OpenTelemetry into CoreDNS allows even third party plugins to further built and extend observability capabilities of their plugins on top of OpenTelemetry.

This sounds a bit like a solution in search of a problem. What actual problems does this solve for CoreDNS?

For example, the existing Prometheusmetrics plugin allows all CoreDNS plugins to integrate into that observability framework.

Can other plugins not use the existing `trace` plugin?

> Binding observability capabilities of CoreDNS only to one concrete technology (Zipkin) doesnt sound right to me.

But this is just tracing. Is tracing really so complicated that we can't just produce a format that other systems can consume it?

Again, I feel like this is a solution in search of a problem. Why is this a separate plugin and not changes to the existing `trace` plugin?

Tantalor93:
> This sounds a bit like a solution in search of a problem. What actual problems does this solve for CoreDNS?

This PR is trying to solve the problem that the OpenTracing project, upon which the `trace` plugin is built is now an archived project as [announced](https://github.com/opentracing/specification/issues/163) by maintainers of the project and also by CNCF ([announcement](https://www.cncf.io/blog/2022/01/31/cncf-archives-the-opentracing-project/)) and it is recommended to migrate to an OpenTelemetry

> But this is just tracing. Is tracing really so complicated that we can't just produce a format that other systems can consume it?

I think this is also one of the points of this PR, to add support for tracing in a new uniform tracing standard 

> Why is this a separate plugin and not changes to the existing trace plugin?

support for OpenTelemetry has been added as a separate plugin in order to maintain backward compatibility and to allow users of CoreDNS to migrate to an OpenTelemetry standard consciously

.gitignore
Tantalor93: good tip! removed

plugin/opentelemetry/setup.go
Tantalor93: you are right, fixed!

plugin/opentelemetry/setup.go
Tantalor93: yea, seems better, fixed!

chrisohaver:
I agree that there is a clear need to move to OpenTelemetry, since OpenTracing has been deprecated. 

I'm not sure about new plugin vs updating the existing plugin. On the surface, it seems that updating the existing _trace_ plugin to use openTelemetry would be less disruptive to users unless it is truly backward incompatible. Can we leverage the [opentracing compatibility layer](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/opentracing.md)?

mariuskimmina:
Got around to test it out a bit - I am not very familiar with OpenTelemetry, and yet it was easy to get this to work (with zipkin) - good job there!

mariuskimmina:
One question: 
I used the following Corefile:
```
.:54 {
    log
    opentelemetry zipkin http://localhost:9411/api/v2/spans
    whoami
}
```

I am using `http` here without the `insecure` flag and it seems to work just fine - is that intended?
From your documentation it sounds like http should not be possible without the `insecure` flag

Tantalor93:
@mariuskimmina yea that `insecure` flag is mainly to enable using OpenTelemetry exporter via HTTP, Zipkin is not that strict and works with HTTP even without an insecure option

@chrisohaver probably we could use that opentracing-opentelemetry bridge and migrate `trace` plugin in a backward compatible way (I would have to check this), although that will not enable us to get rid of OpenTracing dependency anytime soon. So my idea was to go around this and rather built a new plugin on top of a new SDK and have a plugin with a new cleaner API. While maintaining the old `trace` plugin for a few releases and then the drop `trace` plugin.

chrisohaver:
Yes, the compatibility layer doesn't give us anything at this point, since the new implementation is already completed.

I think given that it's already implemented as a separate plugin at this point, it's unlikely we'll identify any reason compelling enough to warrant reworking this into the existing trace plugin.

Tantalor93:
Are there any actions required from me or any questions that need to be answered in order to move forward with this PR? @chrisohaver @SuperQ or will we wait for more opinions from other maintainers? I just want to avoid this to go stale

chrisohaver:
I’ll try to dedicate time to fully review this coming week.

chrisohaver:
> I’ll try to dedicate time to fully review this coming week.

Actually I’ve got some older PRs in my queue, so I probably will not get to this one next week.

plugin/trace/README.md
```diff
@@ -9,6 +9,8 @@
 With *trace* you enable OpenTracing of how a request flows through CoreDNS. Enable the *debug*
 plugin to get logs from the trace plugin.
 
+**Consider using [opentelemetry plugin](../opentelemetry) for tracing of DNS requests, because [OpenTracing project as such is deprecated](https://github.com/opentracing/specification/issues/163).**
```
chrisohaver: I'm thinking we should deprecate _trace_ now, and remove it in the next minor release.

chrisohaver (approved):
LGTM

plugin/trace/README.md
Tantalor93: ok, what should be done to deprecate the plugin @chrisohaver?

chrisohaver: If others agree we should deprecate now, we just mention that the plugin is deprecated prominently in the plugin README. And in next release, we'd mention the plugin has been deprecated in the release notes.

Later in a future minor release, we'd remove the plugin.  If I recall, our policy is to wait at lease a couple of releases before removing a plugin/disabling a feature after deprecation.

Tantalor93: I have added a deprecation notice into trace plugin README 👍

Dentrax:
Great work!

Cc'ing @harshanarayana for awareness. 😇

Tantalor93:
@chrisohaver I am wondering whether we should wait for more maintainers to approve/disapprove this PR or how to approach this PR? WDYT?

chrisohaver:
> I'm not sure I want opentelemetry in CoreDNS. I'd rather go with a lighter weight tracing library. OTel brings in a lot of excess baggage.
> Perhaps this would be more appropriate as an external plugin.

@superq, have you changed your position on replacing the trace plugin with this opentelemetry plugin?

chrisohaver:
Let's merge this.  @Tantalor93 can you update go mod/sum?

SuperQ:
@chrisohaver No, not really. I'm not a fan of Open Telemetry as a project. I don't think it's going in the right direction. It's a huge kitchen sink of stuff.

chrisohaver:
> @chrisohaver No, not really. I'm not a fan of Open Telemetry as a project. I don't think it's going in the right direction. It's a huge kitchen sink of stuff.

Thanks for responding. Can you suggest alternatives, path forward, what to do with existing trace plugin, and how to support tracing natively in CoreDNS in the future?

SuperQ:
Like I posted above, [zipkin-go](https://github.com/openzipkin/zipkin-go) might be a good option. From what I've seen in tracing, zipkin is a reasonably supported format that can be accepted by most telemetry receivers. So even if someone wants to use OpenTelemetry, it will happily accept zipkin as the format.

SuperQ:
And again, even if we do want to use OTel library, why is this a separate plugin and not an update to the existing tracing plugin?

chrisohaver:
> And again, even if we do want to use OTel library, why is this a separate plugin and not an update to the existing tracing plugin?

If it was an update to the existing trace plugin, would that change your position on including it in coredns?

edit: FWIW, a reason for making it a separate plugin was given [above](https://github.com/coredns/coredns/pull/5777#issuecomment-1328233569), but I suspect this doesn't matter since you are opposed to including the opentel package in CoreDNS - I don't think it would matter which plugin includes it.

chrisohaver:
> Like I posted above, [zipkin-go](https://github.com/openzipkin/zipkin-go) might be a good option. From what I've seen in tracing, zipkin is a reasonably supported format that can be accepted by most telemetry receivers. So even if someone wants to use OpenTelemetry, it will happily accept zipkin as the format.

Ok that sounds reasonable.

thipokch:
Where can I find this external plugin?

chrisohaver:
I don’t think there is one.

Issue #5780, opened 2022-11-28
Allow the server timeouts and queries per connection to be configured


rlees85:
**What would you like to be added**:

I would like to be able to configure the timeouts on the DNS server. This is NOT for the forwarding servers or the forward plugin. It is for the listener(s) on CoreDNS itself. In addition to configuring the timeouts I would also like to configure the amount of requests per connection or simply remove the limit.

CoreDNS uses: https://github.com/miekg/dns as its DNS server and this allows configuration of timeouts and requests per connection.

I would like to configure:

* ReadTimeout (https)
* WriteTimeout (https)
* IdleTimeout (https)
* ReadTimeout (tcp/tls)
* WriteTimeout (tcp/tls)
* IdleTimeout (tcp/tls)
* MaxTCPQueries (tcp/tls) - happy to just disable this by hardcoding to `-1` if the community feels this is better since I seen a comment on an issue on the miekg DNS repo that he didn't really see the need for the limit but it is left there anyway.

**Why is this needed**:

CoreDNS currently has a hard-coded 2 second read timeout for TCP/TLS connections and 5 seconds for HTTPS. This appears to be optimised for local network and/or Kubernetes environments where CoreDNS is most often used. Where CoreDNS is acting as a resolver on the public Internet this 2 second timeout can cause problems particularly on 3G/4G/5G connections where there might be lag.

**Additional details**:

I plan to prepare this pull request myself. I have never written Golang before so it should be fairly hilarious. I have already set hard-coded values myself and it builds but I would like to create something that can be merged back into official CoreDNS releases so it must be refined.

I would just like some pointers on how I should implement this? Since its not a plugin configuration it seems there is no way to put it into `Corefile` in a reasonable way. I am thinking command line parameters?

Once the PR is created hopefully the community can help a bit further around refining and tests/etc. Should hopefully only be a small one!

chrisohaver:
Sounds reasonable to me! Though I think it would be better to do as a new plugin if we can.  This would be similar in category to other plugins such as _tls_, which do not directly process DNS requests but instead configure DNS server options.

rlees85:
Yes after understanding plugins a bit more I agree also...

```$ ./coredns --help
Usage of ./coredns:
  -alsologtostderr
    	log to standard error as well as files
  -conf string
    	Corefile to load (default "Corefile")
  -dns.https_idle_timeout duration
    	Default DNS over HTTPS idle timeout (default 30s)
  -dns.https_read_timeout duration
[...]

PR #5784, opened 2022-11-29
plugin/timeouts - Allow ability to configure listening server timeouts

rlees85:

Creates a "timeouts" plugin for configuring the listening server timeouts. Applies to TCP, DNS-over-TLS, DNS-over-HTTPS.

CoreDNS currently has a hard-coded 2 second read timeout for TCP/TLS connections and 5 seconds for HTTPS. This appears to be optimised for local network and/or Kubernetes environments where CoreDNS is most often used. Where CoreDNS is acting as a resolver on the public Internet this 2 second timeout can cause problems particularly on 3G/4G/5G connections where there might be lag.

It also removes the limit of queries per connection since we are enforcing timeouts this makes no sense and can cause some devices to think the DNS server is faulty.

It also removes the 128 requests per connection limit.

`Corefile` is also added to `.gitignore` for easier testing. Happy to revert this if preferred.

See more details in below.

https://github.com/coredns/coredns/issues/5780

Plugin documentation - included.

The only possible issue would be clients that *expect* the server to close the connection after 128 requests which feels very unlikely.


rlees85:
I think this is as far as I can take this PR. Any guidance in making it good enough for merge would be appreciated. I am not 100% familiar with everything that's going on in CoreDNS and this is my first attempt at Go so only happy to take criticism/suggestions.

core/dnsserver/server.go
```diff
@@ -81,6 +87,18 @@ func NewServer(addr string, group []*Config) (*Server, error) {
 		// append the config to the zone's configs
 		s.zones[site.Zone] = append(s.zones[site.Zone], site)
 
+		// set timeouts
+		for key, timeout := range site.Timeouts {
+			switch key {
+			case "idle":
```
Tantalor93: I am not a fan of these strings as keys, I would expect some static enums here, something like 
```
		for key, timeout := range site.Timeouts {
			switch key {
			case IdleTimeout:
				s.idleTimeout = timeout
			case ReadTimeout:
				s.readTimeout = timeout
			case WriteTimeout:
				s.writeTimeout = timeout
			}
		}
```
see other comments for config declaration

core/dnsserver/config.go
```diff
@@ -53,6 +54,9 @@ type Config struct {
 	// TLSConfig when listening for encrypted connections (gRPC, DNS-over-TLS).
 	TLSConfig *tls.Config
 
+	// Timeouts for TCP, TLS and HTTPS servers.
+	Timeouts map[string]time.Duration
```
Tantalor93:
```suggestion
	Timeouts map[TimeoutType]time.Duration
```

and create new enum type
```
type TimeoutType uint8

const (
	IdleTimeout TimeoutType = iota
	ReadTimeout
	WriteTimeout
)
```

plugin/timeouts/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+Additionally some routers hold open connections when using DNS over TLS or DNS over HTTPS. Allowing
+a longer idle timeout helps performance and reduces issues with such routers.
+
+The *timeouts* "plugin" allows you to configure CoreDNS server read, write and idle timeouts.
+
+## Syntax
+
+~~~ txt
+timeouts READ [WRITE] [IDLE]
```
Tantalor93: I would choose different approach to passing timeout durations, your approach forces me to provide all configurations when I want to override just IDLE timeout, so I would go with extra knobs configuration

```suggestion
timeouts {
	read DURATION
	write DURATION
	idle DURATION
}
```
=> If I want to override just write timeout I can do
```
timeouts {
	write 5s
}
```
if I want to override just idle timeout I can do 
```
timeouts {
	idle 5s
}
```

plugin/timeouts/README.md
```diff
@@ -0,0 +1,58 @@
...
+## Examples
+
+Start a DNS-over-TLS server that picks up incoming DNS-over-TLS queries on port 5553 and uses the
+nameservers defined in `/etc/resolv.conf` to resolve the query. This proxy path uses plain old DNS.
+A 10 second read timeout, 20 second write timeout and a 60 second idle timeout have been configured.
+
+~~~
+tls://.:5553 {
+	tls cert.pem key.pem ca.pem
+	timeouts 10 20 60
```
Tantalor93: also I would expect these arguments to accept format in GO duration, meaning  `5s`, `500ms`, etc

Tantalor93 (changes requested):
Hello, thanks for your contribution! Although I have some objections regarding plugin configuration, please consider adjusting how the plugin will be configured

core/dnsserver/config.go
chrisohaver: Could instead just be 3 separate properties in the struct. No need that I see for them to be in a map.

plugin/timeouts/README.md
rlees85: This makes a lot more sense. I am not entirely sure why I didn't do this to start with.

plugin/timeouts/README.md
rlees85: The revision I am going to submit will accept Go duration but if a plain number is specified it will assume that seconds was meant.

core/dnsserver/server.go
```diff
@@ -81,6 +87,17 @@ func NewServer(addr string, group []*Config) (*Server, error) {
...
+			s.readTimeout = site.ReadTimeout
+		}
+		if site.WriteTimeout != time.Duration(0) {
+			s.writeTimeout = site.WriteTimeout
+		}
+		if site.IdleTimeout != time.Duration(0) {
+			s.idleTimeout = site.IdleTimeout
+		}
+
 		// copy tsig secrets
```
rlees85, line 101: Wondering if there is a better way to do this (multiple if)

core/dnsserver/server.go
rlees85: I think this is resolved by no longer using a map. Please let me know if not

rlees85:
Changes submitted for how the plugin should be configured

core/dnsserver/server_tls.go
```diff
@@ -87,3 +97,7 @@ func (s *ServerTLS) OnStartupComplete() {
 		fmt.Print(out)
 	}
 }
+
+const (
+	tlsMaxQueries = -1
```
chrisohaver, line 102: This change doesn't seem related to timeout configuration, and appears to be changing the default value from 0 to -1.

core/dnsserver/server.go
chrisohaver: Maybe?  But it looks perfectly fine and easy to read as is.

core/dnsserver/server_tls.go
rlees85: You are right I think I didn't mention it in the pull request, but did in the issue its linked to, so my bad. The default was never set originally by Coredns so is defaulted to 128 here: https://github.com/miekg/dns/blob/master/server.go#18

The change is to remove the limit since we are having timeouts for idle connections anyway and an active connection has no good reason to limit the amount of requests it can make. Especially given a malicious actor could just keep reconnecting and sending another 128 anyway.

I'm also sure I found a post on Github somewhere where miekg mentioned he wasn't sure the point of the limit especially with timeouts set hence why I hard-coded to -1 (no limit) rather than trying to make it configurable.

chrisohaver: > I think I didn't mention it in the pull request ...

You did mention it in the PR description, I just didn't see it before.  Thanks!

From the PR description...
> It also removes the limit of queries per connection since we are enforcing timeouts this makes no sense and can cause some devices to think the DNS server is faulty.

plugin/pkg/timeouts/timeouts.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"fmt"
+	"strconv"
+	"time"
+)
+
+// NewTimeoutFromArg returns a time.Duration from a configuration argument
+// (string) which has come from the Corefile. The argument has some basic
+// validation applied before returning a time.Duration.
+func NewTimeoutFromArg(arg string) (time.Duration, error) {
```
Tantalor93: why do we want this externalized? can we rather have it only as part of `plugin/timeouts` package?

plugin/pkg/timeouts/timeouts.go
```diff
@@ -0,0 +1,30 @@
...
+	_, err := strconv.Atoi(arg)
+	if err == nil {
+		// If no time unit is specified default to seconds rather than
+		// GO's default of nanoseconds.
+		arg = arg + "s"
+	}
+
+	d, err := time.ParseDuration(arg)
+	if err != nil {
+		return time.Duration(0), fmt.Errorf("failed to parse timeout duration '%s'", arg)
```
Tantalor93: just a nitpick, unnecessary boxing of 0 to time.Duration (I think there are multiple occurrences of this unnecessary boxing)
```suggestion
		return 0, fmt.Errorf("failed to parse timeout duration '%s'", arg)
```

plugin/timeouts/timeouts.go
```diff
@@ -0,0 +1,63 @@
...
+		for c.NextBlock() {
+			block := c.Val()
+			timeoutArgs := c.RemainingArgs()
+			if len(timeoutArgs) != 1 {
+				return c.ArgErr()
+			}
+
+			timeout, err := timeouts.NewTimeoutFromArg(timeoutArgs[0])
+			if err != nil {
+				return err
```
Tantalor93:
```suggestion
		return c.Err(err.Error())
```
to also propagate the line number of Corefile to the error message

plugin/pkg/timeouts/timeouts_test.go
```diff
@@ -0,0 +1,65 @@
+package timeouts
+
+import (
+	"testing"
+	"time"
+)
+
+func TestNewTimeoutFromArg(t *testing.T) {
```
Tantalor93: I think this would be better as table test
```
func TestNewTimeoutFromArg(t *testing.T) {
	tests := []struct {
		name    string
		arg     string
		wantErr bool
		want    time.Duration
	}{
		{
			name: "valid GO duration - seconds",
			arg:  "30s",
			want: 30 * time.Second,
		},
		{
			name: "valid GO duration - minutes",
			arg:  "2m",
			want: 2 * time.Minute,
		},
		{
			name: "number - fallback to seconds",
			arg:  "30",
			want: 30 * time.Second,
		},
		{
			name:    "invalid timeout",
			arg:     "twenty seconds",
			wantErr: true,
		},
		{
			name:    "timeout is too low",
			arg:     "0",
			wantErr: true,
		},
		{
			name:    "timeout is too high",
			arg:     "24h1s",
			wantErr: true,
		},
	}
	for _, test := range tests {
		t.Run(test.name, func(t *testing.T) {
			actual, err := NewTimeoutFromArg(test.arg)
			if test.wantErr && err == nil {
				t.Error("error was expected")
			}
			if !test.wantErr && err != nil {
				t.Error("error was not expected")
			}

			if test.want != actual {
				t.Errorf("expected '%v' got '%v'", test.want, actual)
			}
		})
	}
}
```

Tantalor93 (changes requested):
overall looks good to me, I would just make few quality improvements 👍

plugin/pkg/timeouts/timeouts.go
Tantalor93: if we want this externalized I think we should try to think of better name of a package and function, this is quite common duration parsing, not necessarily timeout
```suggestion
package durations

func NewDurationFromArg(arg string) (time.Duration, error) {
```

rlees85: I have no good reason to externalise it. Looking at the TLS plugin it looked like anything beyond very basic parsing was externalised. I would be happy to try and bring it together if it was felt it would be better.

rlees85:
Thanks @Tantalor93 I will try and get on these improvements when I get a moment

plugin/pkg/timeouts/timeouts.go
rlees85: I think I would leave it (mostly) externalised but rename to durations. It would do the basic parsing and default durations passed as integers to seconds as it does currently. This might be useful in other areas of the app when parsing durations rather than having to duplicate code.

The check for "max" and "min" durations will be moved into the `plugin/timeouts` bit since durations parsed in other parts of the code may not have such constraints.

Would you be happy with that approach? i will convert the tests to table tests as well

Tantalor93: Sounds good 👍

rlees85:
Hopefully that should cover the comments unless I've messed something up (i've ran go tests/etc locally so hopefully not!)

Tantalor93 (approved):
LGTM, thanks for your contribution!

Issue #5785, opened 2022-11-29
Timeout instead of BADVERS in EDNS (DNSDist + CoreDNS)


kaplanben:
**What happened**:
During tests of EDNS compliance of our DNS infrastructure, we noticed that some ENDS queries are not returned with their expected responses. We confirm this behavior with multiple EDNS compliance tools and all returned the same results.

The 2 types of requests are `Unknown Version` and `Unknown Version with Unknown Option` see [link](https://kb.isc.org/docs/edns-compatibility-dig-queries)

Instead of returning `BADVERS` the query fail on Timeout error (`connection timed out; no servers could be reached`).

**What you expected to happen**:
Both queries should return: 
 - `BADVERS` status
 - OPT record with EDNS version set to 0
 - not to see SOA record in the ANSWER section of the query response

**How to reproduce it (as minimally and precisely as possible)**:
1. CoreDNS with default values and a simple example DNS zone.
2. DNSdist to forward DNS queries to CoreDNS instance from step 1.
3. Run the following dig clis
   - `dig +norec +edns=100 +noednsneg soa $zone @$dnsdist_server`
   - `dig +norec +edns=100 +noednsneg +ednsopt=100 soa $zone @$dnsdist_server`

**Environment**:
 - OS: ubuntu 18.04 <--> 22.04
 - Software: CoreDNS-1.10.0 (dnsdist 1.6.1)
 
**Anything else we need to know?**:
We tried many different setups (for ruling out possible causes) and since this is a tricky issue I'll list all of our findings here.
Three test cases, where **Case C** is the one with the mentioned issue.

> The following two test cases are not included in this issue, since they all worked as expected, but should be considered when 
   doing a triage:
>  - `DNSDist` forwarding queries to a `bind9` - *works as expected*
>  - `DNSDist` forwarding queries to a `PowerDNS` - *works as expected* 

1. Case A - Using only `bind9` server (to be as a test case) - *works as expected*
2. Case B - Using only `CoreDNS` server (trying to rule out CoreDNS)  - *works as expected*
3. Case C - `DNSDist` forwarding queries to a `CoreDNS` service as its backend.  - **Timeout**

The following DNS zone file (`/tmp/example.org`) is shared on all the cases below:

```
$TTL    1
@       IN      SOA     ns1.example.org. admin.example.org. (
                              5
                         604800
                          86400
                        2419200
                         604800 )
@       IN      NS      ns1.example.org.
@       IN      A       127.0.0.1
[...]

PR #5787, opened 2022-11-30
plugin/edns: remove truncating of question section on bad EDNS version

kaplanben:
EDNS requests of "Unknown Version" removed the query section altogether. 
Not sure why since this is not required (see [link](https://kb.isc.org/docs/edns-compatibility-dig-queries))

This causes issues with DNS solutions that use this information (initial queried name, type, and class) in order to route the response to the right client (e.g. PDNS).

The change here is to keep the initial query section as is.

[5785](https://github.com/coredns/coredns/issues/5785)

None that I'm aware of since the ENDS "Unknown Version" requirements are clear and do not include removing the query section.

### DIFF
--- a/plugin/pkg/edns/edns_test.go
+++ b/plugin/pkg/edns/edns_test.go
@@ -29,9 +29,12 @@ func TestVersionNoEdns(t *testing.T) {
 	m := ednsMsg()
 	m.Extra = nil
 
-	_, err := Version(m)
+	r, err := Version(m)
 	if err != nil {
 		t.Errorf("Expected no error, but got one: %s", err)
+	}
+	if r != nil {
+		t.Errorf("Expected nil since not an EDNS0 request, but did not got nil")
 	}
 }
 
--- a/plugin/pkg/edns/edns_test.go
+++ b/plugin/pkg/edns/edns_test.go
@@ -10,9 +10,18 @@ func TestVersion(t *testing.T) {
 	m := ednsMsg()
 	m.Extra[0].(*dns.OPT).SetVersion(2)
 
-	_, err := Version(m)
+	r, err := Version(m)
 	if err == nil {
 		t.Errorf("Expected wrong version, but got OK")
+	}
+	if r.Question == nil {
+		t.Errorf("Expected question section, but got nil")
+	}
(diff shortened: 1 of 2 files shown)


chrisohaver:
> EDNS requests of "Unknown Version" removed the query section altogether. Not sure why ...

Yeah.  The colorful comments don't explain _why_. I mean, it's fair to assume there was an actual reason, but we don't know what that reason was. Presumably "something" didn't work when there was a question (during local manual testing?). But the comments are at least helpful by implying the need to remove the question was deemed dubious at the time. The code is 7 years old now, and the original author (and CoreDNS founder) is no longer an active maintainer.

chrisohaver (approved):

Tantalor93:
I think it would be worth to cover this functionality better by unit tests https://github.com/coredns/coredns/blob/master/plugin/pkg/edns/edns_test.go#L9 WDYT @chrisohaver? Like check the returned dns.Msg

Tantalor93:
but otherwise, I agree with this change

chrisohaver:
> I think it would be worth to cover this functionality better by unit tests

Yeah - the existing tests cover very poorly - ignoring the returned message and only checking the error returned.

chrisohaver (changes requested):
@kaplanben can you please fix the existing unit tests to actually validate the messages returned by `Versions`.  Thanks!

plugin/pkg/edns/edns_test.go
```diff
@@ -10,10 +10,19 @@ func TestVersion(t *testing.T) {
...
+	}
+	if r.Rcode != dns.RcodeBadVers {
+		t.Errorf("Expected Rcode to be of BADVER (16), but got %d", r.Rcode)
+	}
+	if r.Extra == nil {
+		t.Errorf("Expected OPT section, but got nil")
+	}
 }
 
 func TestVersionNoEdns(t *testing.T) {
```
Tantalor93, line 28: can you add an assertion here that the response message is nil? So it is actually asserted that it is expected in case of no EDNS?

Tantalor93 (approved):

chrisohaver (approved):
Lgtm thanks!

PR #5795, opened 2022-12-06
plugin/geoip: Add GLSB example

chrisohaver:

Adds example of using _geoip_ + _view_ plugins to provide GLSB functionality.

This example could arguably be placed in either the _view_ plugin README or _geoip_ README.
But I think from a use case perspective, it makes more sense in the _geoip_ README.


chrisohaver:
FYI: @SuperQ

SuperQ:
Nice, does the order matter?

chrisohaver:
> Nice, does the order matter?

It does.  Server blocks are evaluated in the order listed in the Corefile. 
A default "viewless" server block must always be the last server block of its domain (otherwise CoreDNS won't start).

Added a note in the example about order.

SuperQ (approved):
LGTM. 

For order, I was more thinking about does the `geoip` need to be loaded before the `view`.

I guess inside the block it doesn't matter.

snebel29 (approved):
LGTM, combining those two plugins is cool!

Issue #5478, opened 2022-07-01
Domains cache cannot be updated correctly when the server time changes


luozhiwenn:
**What happened**:
When the time of the server where the coredns resides change(set the time to a certain time in the past), the domains cache cannot be updated correctly.

**What you expected to happen**:
domains cache can be updated correctly
I think coredns should use [golang monotonic clock](https://golang.org/pkg/time/#hdr-Monotonic_Clocks) to calculate ttl in cache plugin

**How to reproduce it (as minimally and precisely as possible)**:
- run CoreDNS in kubernetes cluster as DNS server
- create a service `test-svc`
- initiate a domain name resolution request for this service, get the service ClusterIP
![image](https://user-images.githubusercontent.com/74550097/176848260-ca5e337d-5f78-425e-ab96-a291c6b4e330.png)
- use `date -s "xxx"` to change the time one hour back
- delete the created service `test-svc` then recreate, kubernete will allocate a new ClusterIP for this service
- initiate a domain name resolution request again but got a stale ClusterIP that first time created, even though the record ttl has passed
![image](https://user-images.githubusercontent.com/74550097/176848325-0982f163-3db1-410b-9350-08501146c567.png)

**Anything else we need to know?**:

**Environment**:

- the version of CoreDNS: v1.8.3
- Corefile: 
- logs, if applicable:
- OS (e.g: `cat /etc/os-release`): EulerOS
- Others:

chrisohaver:
At a glance, it looks like _cache_ handles this OK.  Monotonic clock operations are handled implicitly by Golang.  Time.time contains a monotonic time and a local time (aka wall clock); operations such as `Sub` use the monotonic times if available.  I don't see anything that would strip the monotonic times.

A replication without the _kubernetes_ plugin would be more minimal.  Many confounding factors in a K8s environment, e.g.  time serialization between multiple k8s processes, multiple k8s nodes each with their own clock ...

luozhiwenn:
Maybe the `now.UTC()` strip the monotonic times?
https://github.com/coredns/coredns/blob/c4d02442ef493b4892316fea00128c982d48a914/plugin/cache/item.go#L97

I found such a sentence in [golang monotonic clock](https://golang.org/pkg/time/#hdr-Monotonic_Clocks):
![image](https://user-images.githubusercontent.com/74550097/176995539-6cb5082c-a52a-45c6-a7e8-4bae423ac49d.png)
[](url)

luozhiwenn:
[...]

PR #5798, opened 2022-12-10
plugin/cache: remove cache item while system time adjust backward

mike96265:
coredns cache plugin record when the dns reply has been received. base on the store time and ttl to detect outdated cache item. but when system time adjust backward, the cache item might take effect much longer than it should be.

### DIFF
--- a/plugin/cache/handler.go
+++ b/plugin/cache/handler.go
@@ -121,18 +121,26 @@ func (c *Cache) getIgnoreTTL(now time.Time, state request.Request, server string
 
 	if i, ok := c.ncache.Get(k); ok {
 		itm := i.(*item)
-		ttl := itm.ttl(now)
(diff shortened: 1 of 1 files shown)


johnbelamaric:
This adds a conditional in the hot path for something the almost never happens. Instead, an independent go routine that periodically (once per 5 minutes?) checks for this condition and just flushes the entire cache if it exists would be better.

That would have zero impact on the hot path.

chrisohaver:
I think careful handling of time will solve this ( e.g. not using time functions that strip the monotonic time).  This was raised before, but it's been low on my priority due to it being a corner case.

here: https://github.com/coredns/coredns/issues/5478

chrisohaver (changes requested):
I think this would be better solved by refactoring the existing code to not use time functions that strip monotonic time. See: https://github.com/coredns/coredns/issues/5478

PR #5806, opened 2022-12-13
Drop obsolete client-go auth plugins

skitt:

The OpenStack plugin is no longer available, even in version 0.24.4 of client-go; see
https://github.com/kubernetes/client-go/blob/v0.24.4/plugin/pkg/client/auth/openstack/openstack_stub.go It is replaced by the client-keystone-auth credential plugin. The plugin has been entirely removed in client-go 0.26.0, which breaks the build when any other dependency pulls in client-go 0.26.0 or later.

The GCP plugin is deprecated in K8s 1.22+ and unavailable in 1.26+ (although it is still stubbed in client-go 0.26.0). It is replaced by the gke-gcloud-auth-plugin credential plugin.

None that I’m aware of.

None.

It drops direct support for GCP and OpenStack authentication, but they were already unusable anyway (the stubbed plugins always return an error).

### DIFF
--- a/plugin/kubernetes/setup.go
+++ b/plugin/kubernetes/setup.go
@@ -17,9 +17,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/miekg/dns"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
-	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"       // pull this in here, because we want it excluded if plugin.cfg doesn't have k8s
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"      // pull this in here, because we want it excluded if plugin.cfg doesn't have k8s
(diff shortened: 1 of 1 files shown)


plugin/kubernetes/setup.go
```diff
@@ -17,9 +17,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/miekg/dns"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
-	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"       // pull this in here, because we want it excluded if plugin.cfg doesn't have k8s
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"      // pull this in here, because we want it excluded if plugin.cfg doesn't have k8s
```
chrisohaver:
```suggestion
	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc" // pull this in here, because we want it excluded if plugin.cfg doesn't have k8s
```

chrisohaver (approved):
Thanks!

PR #5828, opened 2022-12-28
plugin/forward race policy

zzhuang94:
`plugin/forward` supports race policy
1. we have multiple upstream hosts
2. we want to send request to all of them at once, and pick the fastest result
3. so we can use race policy
4. just like `dnsmasq` multiple upstream hosts usage

no issues

`plugin/forward/README.md` already update

no

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -47,7 +47,7 @@ forward FROM TO... {
     max_fails INTEGER
     tls CERT KEY CA
     tls_servername NAME
-    policy random|round_robin|sequential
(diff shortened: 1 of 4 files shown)


Tantalor93:
I quite like this feature, but there is already external plugin which provides the desired functionality https://github.com/networkservicemesh/fanout

zzhuang94:
> I quite like this feature, but there is already external plugin which provides the desired functionality https://github.com/networkservicemesh/fanout

thank you very much!

PR #5833, opened 2023-01-05
plugin/nomad: Add a Nomad plugin

### LINKED ISSUE #5829: New plugin: Nomad

Nomad recently launched support for [service discovery](https://www.hashicorp.com/blog/nomad-service-discovery), which enables discovering services without the use of Consul. Nomad maintains a service catalogue of all the services which are registered, with it's address/port, tags etc available over APIs.

**Why is this needed**:

A Nomad plugin would enable the usecase of discovering Nomad services via DNS. Currently that isn't possible and the only way to use Nomad services is via `consul-template` or querying Nomad HTTP APIs directly. However, querying services via DNS is a common use case and the plugin will allow for that.

I'd like to build the plugin and send a PR if it's okay to add support for this.

mr-karan:
This commit introduces a `nomad` plugin which provides a DNS interface for querying Nomad services.

This PR adds a `nomad` plugin. It provides the capability to query [Nomad services API](https://developer.hashicorp.com/nomad/api-docs/services) via DNS.

https://github.com/coredns/coredns/issues/5829

The plugin has a README.md so I believe that is sufficient.

No.


plugin/nomad/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+;; Query time: 0 msec
+;; SERVER: 127.0.0.1#1053(127.0.0.1) (UDP)
+;; WHEN: Thu Jan 05 12:12:20 IST 2023
+;; MSG SIZE  rcvd: 339
+```
+
+### plugin.cfg
+
+This plugin is intended to appear twoard the end of the plugin list, usually near the `proxy` plugin declaration.
```
chrisohaver, line 144: This should be removed.

plugin/nomad/Corefile.example
```diff
@@ -0,0 +1,12 @@
...
+    debug
+    health
+    log
+    nomad {
+	  	address http://127.0.0.1:4646
+        ttl 10
+    }
+    prometheus :9153
+    cache 30
+}
```
chrisohaver, line 12: Examples should be inline in the plugin README instead of a separate file.

plugin/nomad/ready.go
```diff
@@ -0,0 +1,6 @@
+package nomad
+
+// Ready signals when the plugin is ready for use.
+// In case of Nomad, when the ping to the Nomad API is successful
+// the plugin is ready.
+func (n Nomad) Ready() bool { return true }
```
chrisohaver, line 6: Instead returning a hard coded "true" this should return the value of a property in the nomad struct.

plugin/nomad/setup.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	// Do a ping check to check if the Nomad server is reachable.
+	_, err := n.client.Agent().Self()
+	if err != nil {
+		return plugin.Error("nomad", err)
+	}
+	// Mark the plugin as ready to use.
+	// https://github.com/coredns/coredns/blob/master/plugin.md#readiness
+	n.Ready()
```
chrisohaver, line 33: Should be deferred to OnStartup, and instead of calling n.Ready(), this should set a ready boolean variable in the nomad struct to true ...  and the implementation of Ready should return that value instead of hard coded true.

plugin/nomad/setup.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	if err := parse(c, n); err != nil {
+		return plugin.Error("nomad", err)
+	}
+
+	// Do a ping check to check if the Nomad server is reachable.
+	_, err := n.client.Agent().Self()
+	if err != nil {
+		return plugin.Error("nomad", err)
+	}
```
chrisohaver, line 30: It would be better to defer the check to c.OnStartup. That way a down nomad api isn't preventing coredns from starting (other plugins may be able to answer).

plugin/nomad/nomad.go
```diff
@@ -0,0 +1,149 @@
...
+// in a Server.
+func (n Nomad) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	state := request.Request{W: w, Req: r}
+	qname := state.Name()
+	qtype := state.QType()
+
+	// Split the query name with a `.` as the delimiter and extract namespace and service name.
+	// If the query is not for a Nomad service, return.
+	qnameSplit := dns.SplitDomainName(qname)
+	if len(qnameSplit) < 3 || qnameSplit[2] != "nomad" {
```
chrisohaver, line 44: I think this would unintentionally fail to catch queries like `www.example.nomad.com`.
Safer to use plugin.Name.Matches ... e.g. `plugin.Name("nomad").Matches(state.Name())`

IMO "nomad" should not be hard coded, but should be configurable as the zone of the plugin.  e.g. 

```
nomad ZONE {
     address URL
     token TOKEN
     ttl DURATION
}
```

plugin/nomad/nomad.go
```diff
@@ -0,0 +1,149 @@
...
+	namespace := qnameSplit[1]
+	serviceName := qnameSplit[0]
+
+	m := new(dns.Msg)
+	m.SetReply(r)
+	m.Authoritative = true
+	m.Compress = true
+	m.Rcode = dns.RcodeSuccess
+	m.Answer = []dns.RR{}
+	m.Extra = []dns.RR{}
```
chrisohaver, line 56: The initialization to default values can be removed.

plugin/nomad/nomad.go
```diff
@@ -0,0 +1,149 @@
...
+						Ttl:    n.ttl,
+					},
+					A: addr,
+				})
+			}
+		default:
+			m.Rcode = dns.RcodeNotImplemented
+			w.WriteMsg(m)
+			requestFailedCount.WithLabelValues(metrics.WithServer(ctx), namespace).Inc()
+			return dns.RcodeNotImplemented, nil
```
chrisohaver, line 139: I think this will result in sending two responses to the client (the client would completely ignore the second response, so you would not see it).

The first response written above `w.WriteMsg(m)`, the second written by the server when it sees the "failure" response code return value returned from the plugin chain  (dns.RcodeNotImplemented is categorized as a failure by the server). The server uses `ClientWrite` func below to determine if it needs to send an error.

https://github.com/coredns/coredns/blob/e7ad486b50d8f831b7dd4a0840d300a3bcfac471/plugin/plugin.go#L86-L100

Best fix is to remove the `w.WriteMsg(m)` line. 
If you find the response sent by the server is not correct, or needs tweaking, write the response manually as you are doing now, but return dns.RcodeSuccess instead so the server does not send a second response.

plugin/nomad/nomad.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	// Fetch service registrations for the given service.
+	log.Debugf("Looking up record for svc: %s namespace: %s", serviceName, namespace)
+	svcRegistrations, _, err := n.client.Services().Get(serviceName, (&api.QueryOptions{Namespace: namespace}))
+	if err != nil {
+		m.Rcode = dns.RcodeServerFailure
+		w.WriteMsg(m)
+		requestFailedCount.WithLabelValues(metrics.WithServer(ctx), namespace).Inc()
+		return dns.RcodeServerFailure, fmt.Errorf("error fetching service detail: %w", err)
```
chrisohaver, line 72: See my other comment about writing two response to the client. dns.RcodeServerFailure is also categorized as a failure.

plugin/nomad/nomad.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	// If no service registrations are found, ignore this service.
+	if len(svcRegistrations) == 0 {
+		m.Rcode = dns.RcodeNameError
+		w.WriteMsg(m)
+		requestFailedCount.WithLabelValues(metrics.WithServer(ctx), namespace).Inc()
+		return dns.RcodeNameError, nil
```
chrisohaver, line 80: RE: double writes: This is OK as is. dns.RcodeNameError is _not_ categorized as a failure.

However, NXDOMAIN responses also require an SOA record returned in the Authority section (`Ns`).

plugin/nomad/nomad.go
```diff
@@ -0,0 +1,149 @@
...
+		if addr == nil {
+			m.Rcode = dns.RcodeServerFailure
+			w.WriteMsg(m)
+			requestFailedCount.WithLabelValues(metrics.WithServer(ctx), namespace).Inc()
+			return dns.RcodeServerFailure, fmt.Errorf("error parsing IP address: %w", err)
+		}
+
+		// Check the query type to format the appriopriate response.
+		switch qtype {
+		case dns.TypeA:
```
chrisohaver, line 96: I think before appending, you need to make sure here that the `addr` is actually a v4 IP.
ditto for `AAAA` case.

plugin/nomad/nomad.go
```diff
@@ -0,0 +1,149 @@
...
+				})
+			}
+		default:
+			m.Rcode = dns.RcodeNotImplemented
+			w.WriteMsg(m)
+			requestFailedCount.WithLabelValues(metrics.WithServer(ctx), namespace).Inc()
+			return dns.RcodeNotImplemented, nil
+		}
+	}
+
```
chrisohaver, line 142: Need to add a check here in the event that there were no matching registrations of the queried type.  e.g. if a service has an ipv6 address registered, but no ipv4, and user queried for type `A`.  In this case, we need to add an SOA record to the Authority section of the response, since the sewer will be empty (a "NODATA" response).

chrisohaver (changes requested):
Thanks for contributing!

mr-karan:
Sorry, lost track of this and got busy. I'll work on changes mentioned soon.

jdoss:
@mr-karan Can this get reopened and pushed over the finish line? Having this plugin added to coreDNS would be a fantastic addition.

blinkinglight:
hey, 
how can i continue this PR?

chrisohaver:
@mr-karan, is it ok with you if someone else picks up your proposed code to complete the feature?

SuperQ:
I think this would make more sense to start out as an external plugin.

blinkinglight:
https://github.com/ituoga/coredns-nomad just good enought for now, builds via cd/cd but works with weave.works so good for now.

blinkinglight:
if you think that something is broken and you have no time, please, tell for "next" (just with comment and PR)  what it can do.

blinkinglight:
you don't have to finish what you started just because you started it. but leave not to next, where to start ;)

mr-karan:
This was available as an external plugin since a very long time: https://github.com/mr-karan/coredns-nomad

I agree I should have been more proactive/communicative in this thread. Couldn't make/find time to continue with it. Apologies, will do better next time.

blinkinglight:
> This was available as an external plugin since a very long time: https://github.com/mr-karan/coredns-nomad
> 
> I agree I should have been more proactive/communicative in this thread. Couldn't make/find time to continue with it. Apologies, will do better next time.

just too long :) do you have time to merge and review PR's? ? so maybe we could move github actions part to your repo?

mr-karan:
Sure, please feel free to open a PR. Will get this shipped :rocket:

caiodelgadonew:
@mr-karan when do you plan to have this shipped? will be amazing to use it <3

mr-karan:
@caiodelgadonew It's available as immediate use even now: https://github.com/mr-karan/coredns-nomad

@blinkinglight is helping with releasing a docker image for this plugin.

PR #5847, opened 2023-01-11
plugin/etcd: update documentation

vicoooo26:
If an entry `/skydns/test/skydns/mx` have entries `/skydns/test/skydns/mx/a` and `/skydns/test/skydns/mx/b`, a query for `mx.skydns.text` will return `REFUSED` instead of records of `a` and `b` while a query for `mx.skydns.test` will return expected records.

None
None
None


yongtang:
@vicoooo26  Can you fix the DCO issue shown in the build?

vicoooo26:
Yes, the latest commit has solved the DCO issue.

PR #5879, opened 2023-01-25
plugin/cache: add a new keepttl option

MrFreezeex:
Whenever CoreDNS is used as an authoritative server it could be useful to serve a consistent TTL (not decreased based on time) to clients and still use caching to spare request to the backend.

This commit adds a new option `keepttl` to the cache plugin to do just that. This new option doesn't update any caching logic but just override the cache plugin answer.

To the cache plugin (included in this PR)

no

### DIFF
--- a/plugin/cache/README.md
+++ b/plugin/cache/README.md
@@ -39,6 +39,7 @@ cache [TTL] [ZONES...] {
     serve_stale [DURATION] [REFRESH_MODE]
     servfail DURATION
     disable success|denial [ZONES...]
+    keepttl
 }
 ~~~
 
@@ -69,6 +70,11 @@ cache [TTL] [ZONES...] {
   greater than 5 minutes.
 * `disable`  disable the success or denial cache for the listed **ZONES**.  If no **ZONES** are given, the specified
   cache will be disabled for all zones.
+* `keepttl` do not age TTL when serving responses from cache. The entry will still be removed from cache
+  when the TTL expires as normal, but until it expires responses will include the original TTL instead
+  of the remaining TTL. This can be useful if CoreDNS is used as an authoritative server and you want
+  to serve a consistent TTL to downstream clients. This is **NOT** recommended when CoreDNS is caching
+  records it is not authoritative for because it could result in downstream clients using stale answers.
 
 ## Capacity and Eviction
 
--- a/plugin/cache/handler.go
+++ b/plugin/cache/handler.go
@@ -71,6 +71,11 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) 
 		})
 	}
 
+	if c.keepttl {
+		// If keepttl is enabled we fake the current time to the stored
+		// one so that we always get the original TTL
+		now = i.stored
+	}
 	resp := i.toMsg(r, now, do, ad)
 	w.WriteMsg(resp)
 	return dns.RcodeSuccess, nil
--- a/plugin/cache/cache.go
+++ b/plugin/cache/cache.go
@@ -48,6 +48,9 @@ type Cache struct {
 	pexcept []string
 	nexcept []string
 
+	// Keep ttl option
+	keepttl bool
+
 	// Testing.
 	now func() time.Time
 }
--- a/plugin/cache/cache_test.go
+++ b/plugin/cache/cache_test.go
@@ -690,6 +690,44 @@ func TestCacheWildcardMetadata(t *testing.T) {
 	}
 }
 
+func TestCacheKeepTTL(t *testing.T) {
+	defaultTtl := 60
+
+	c := New()
+	c.Next = ttlBackend(defaultTtl)
+
+	req := new(dns.Msg)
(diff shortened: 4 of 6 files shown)


chrisohaver:
> Whenever CoreDNS is used as an authoritative server it could be useful to serve a consistent TTL to clients and still use caching to spare request to the backend.

It seems like setting the maximum TTL to the longest allowed TTL may be an easier way to solve that situation with the currently available options.  It would leave the TTL unchanged, but also keep record in cache per their original TTL value thus sparing requests to the back end more so than caching for a shorted period of time.

MrFreezeex:
> It seems like setting the maximum TTL to the longest allowed TTL may be an easier way to solve that situation with the currently available options. It would leave the TTL unchanged, but also keep record in cache per their original TTL value thus sparing requests to the back end more so than caching for a shorted period of time.

Hi! thanks for the quick reply :). I am not exactly sure what you mean though. The TTL seem to be always computed by subtracting the time in the cache plugin AFAIU: https://github.com/coredns/coredns/blob/d3e4fc78c3b80c1768b0f95ad849b07ba9968ba5/plugin/cache/item.go#L98. So you would get the "original ttl" the first time it gets cached and then it would always be decreased.

We are trying to do something similar to the dontAge option from dnsdist (https://dnsdist.org/reference/config.html?highlight=dontage):
```
dontAge (bool) – Don’t reduce TTLs when serving from the cache. Use this when dnsdist fronts a cluster of authoritative servers
```

EDIT: I just updated the doc it might be slightly clearer like that

chrisohaver:
Thanks for the clarification.  I think the feature would be problematic applied to any other situations.  E.g. when caching responses retrieved from upstream servers.

I think we should also add a note in the README that enabling this feature could result in downstream caching stale answers from upstream servers.  Or perhaps that we don't recommend using this option for caching answers from upstream dns servers.  IOW, it should not be used when caching records CoreDNS is _not_ authoritative for because it could result in downstream clients using stale answers.

MrFreezeex:
Indeed better documentation is always good :D. I just added a note in the cache plugin readme like you advised!

plugin/cache/README.md
```diff
@@ -69,6 +70,10 @@ cache [TTL] [ZONES...] {
   greater than 5 minutes.
 * `disable`  disable the success or denial cache for the listed **ZONES**.  If no **ZONES** are given, the specified
   cache will be disabled for all zones.
+* `keepttl` enforce the cache plugin to not reduce the TTL he got from its backend. This can be useful
+  if CoreDNS is used as an authoritative server and you want to serve a consistent TTL to the clients.
+  This is **NOT** recommended when CoreDNS is not an authoritative server as using this option could result
```
chrisohaver:
```suggestion
  This is **NOT** recommended when CoreDNS is caching records it is not authoritative for because it could result
```

plugin/cache/README.md
```diff
@@ -69,6 +70,10 @@ cache [TTL] [ZONES...] {
   greater than 5 minutes.
 * `disable`  disable the success or denial cache for the listed **ZONES**.  If no **ZONES** are given, the specified
   cache will be disabled for all zones.
+* `keepttl` enforce the cache plugin to not reduce the TTL he got from its backend. This can be useful
```
chrisohaver:
```suggestion
* `keepttl`  do not age the TTL when serving responses from cache. The entry will still be removed from cache when the TTL expires as normal, but until it expires responses will include the original TTL instead of the remaining TTL. This can be useful 
```

chrisohaver (commented):
Thanks! Just suggesting some tweaks to the wording.

plugin/cache/README.md
chrisohaver: Suggesting this because CoreDNS can be hybrid authoritative for some zones, and non-authoritative others.

MrFreezeex:
Thanks for the wording suggestions, looks indeed better this way!

plugin/cache/README.md
```diff
@@ -69,6 +70,11 @@ cache [TTL] [ZONES...] {
   greater than 5 minutes.
 * `disable`  disable the success or denial cache for the listed **ZONES**.  If no **ZONES** are given, the specified
   cache will be disabled for all zones.
+* `keepttl` do not age TTL when serving responses from cache. The entry will still be removed from cache
+  when the TTL expires as normal, but until it expires responses will include the responses will include the
```
chrisohaver:
```suggestion
  when the TTL expires as normal, but until it expires responses will include the
```

MrFreezeex: Oups my bad indeed sorry for that

plugin/cache/handler.go
```diff
@@ -71,6 +71,11 @@ func (c *Cache) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 		})
 	}
 
+	if c.keepttl {
+		// If ttl is enabled we fake the current time to the stored
```
chrisohaver:
```suggestion
		// If keepttl is enabled we fake the current time to the stored
```

chrisohaver (approved):

PR #5890, opened 2023-01-30
plugin/forward: fix broken tap plugins when dnstap plugins specified

koho:

Prior to this fix, when one or more dnstap plugins were configured, the dnstap in forward plugin is not working due to the following reasons:

1. The type assertion of `Dnstap` is incorrect, which causes an empty plugin list.
2. No iteration over the plugin chain. The plugin list only contains the first dnstap plugin.
3. The query tap message instance is shared across multiple dnstap plugins. If one of the tap plugins overwrites the `QueryMessage` field of the query message, the remaining plugins will use the same message regardless of its own configuration.

This PR fixes the type assertion and iterates over the plugin chain to form a full list of tap plugins. As for the third problem, a new tap message instance will be created for each plugin, just like the response message.

#5773

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -67,9 +67,10 @@ func (f *Forward) SetProxy(p *Proxy) {
 }
 
 // SetTapPlugin appends one or more dnstap plugins to the tap plugin list.
-func (f *Forward) SetTapPlugin(h plugin.Handler) {
-	for tapPlugin, ok := h.(*dnstap.Dnstap); ok; tapPlugin, ok = tapPlugin.Next.(*dnstap.Dnstap) {
-		f.tapPlugins = append(f.tapPlugins, tapPlugin)
+func (f *Forward) SetTapPlugin(tapPlugin *dnstap.Dnstap) {
+	f.tapPlugins = append(f.tapPlugins, tapPlugin)
+	if nextPlugin, ok := tapPlugin.Next.(*dnstap.Dnstap); ok {
+		f.SetTapPlugin(nextPlugin)
 	}
 }
 
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -66,6 +66,13 @@ func (f *Forward) SetProxy(p *Proxy) {
 	p.start(f.hcInterval)
 }
 
+// SetTapPlugin appends one or more dnstap plugins to the tap plugin list.
+func (f *Forward) SetTapPlugin(h plugin.Handler) {
+	for tapPlugin, ok := h.(*dnstap.Dnstap); ok; tapPlugin, ok = tapPlugin.Next.(*dnstap.Dnstap) {
+		f.tapPlugins = append(f.tapPlugins, tapPlugin)
+	}
+}
+
 // Len returns the number of configured proxies.
 func (f *Forward) Len() int { return len(f.proxies) }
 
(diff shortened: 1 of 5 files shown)


plugin/forward/forward.go
```diff
@@ -66,6 +66,13 @@ func (f *Forward) SetProxy(p *Proxy) {
 	p.start(f.hcInterval)
 }
 
+// SetTapPlugin appends one or more dnstap plugins to the tap plugin list.
+func (f *Forward) SetTapPlugin(h plugin.Handler) {
+	for tapPlugin, ok := h.(*dnstap.Dnstap); ok; tapPlugin, ok = tapPlugin.Next.(*dnstap.Dnstap) {
```
chrisohaver: ~This looks like it would be an infinite loop if `h` is a `*dnstap.Dnstap`~

nevermind - I didn't have my coffee yet. 😄

plugin/forward/forward_test.go
```diff
@@ -22,3 +24,32 @@ func TestList(t *testing.T) {
 		}
 	}
 }
+
+func TestSetTapPlugin(t *testing.T) {
+	input := `
+      dnstap /tmp/dnstap.sock full
+      dnstap tcp://example.com:6000
```
chrisohaver: The input should probably also include some plugins that are not dnstap.

chrisohaver (changes requested):
Thanks for contributing!

>  If one of the tap plugins overwrites the QueryMessage field of the query message,

Does this actually happen?  _dnstap_ just relays messages to the endpoint, there is no reason that it should be mutating the messages before/after they go out.

plugin/forward/forward.go
```diff
@@ -66,6 +66,13 @@ func (f *Forward) SetProxy(p *Proxy) {
...
+		f.tapPlugins = append(f.tapPlugins, tapPlugin)
+	}
+}
```
chrisohaver: I think since we know that this will only be called when `h` = the first dnstap plugin (the logic in the function relies on that), we don't need to reassert that in the function, and we can type the parameter more tightly...

```go
func (f *Forward) SetTapPlugin(tapPlugin *dnstap.Dnstap) {
  for {
    f.tapPlugins = append(f.tapPlugins, tapPlugin)
    tapPlugin, ok := tapPlugin.Next.(*dnstap.Dnstap)
    if !ok {
      break
    }
  }
}
```

or a tad shorter with recursion ...

```go
func (f *Forward) SetTapPlugin(tapPlugin *dnstap.Dnstap) {
  f.tapPlugins = append(f.tapPlugins, tapPlugin)
  if  nextPlugin, ok := tapPlugin.Next.(*dnstap.Dnstap); ok {
    f.SetTapPlugin(nextPlugin)
  }
}
```

koho: Sure.

koho:
> Does this actually happen? 

Actually, it happened with the example in readme:

```
dnstap /tmp/dnstap.sock full
dnstap tcp://example.com:6000
```

The first dnstap plugin has the option `IncludeRawMessage = true`, the second `IncludeRawMessage = false`.

https://github.com/coredns/coredns/blob/68b2aa6708cca8d9e7e32f493e07849543839b9d/plugin/forward/dnstap.go#L42-L48

During the first iteration, `q.QueryMessage` is set. But `q.QueryMessage` is not empty in the second iteration, which violates the configuration. Also, `q` is passed by reference. If we put an else clause here to reset `q.QueryMessage`, it will lead to inconsistent data when the channel is busy. The safe way is to make a new message instance per tap plugin.

chrisohaver (approved):

chrisohaver (approved):

PR #5893, opened 2023-02-01
plugin/forward: add new metric for upstream connection failures.

ventifus:
Signed-off-by: Andrew Denton <adenton@redhat.com>

This adds a metric, `coredns_forward_connection_failures_total` that records the number of times the forward plugin failed to communicate with an upstream host, and why. For example:

```text
# HELP coredns_forward_connection_failures_total Counter of connection failures per remote IP address.
# TYPE coredns_forward_connection_failures_total counter
coredns_forward_connection_failures_total{proto="tcp",to="8.8.8.8:53",type="connect: connection refused"} 537
coredns_forward_connection_failures_total{proto="tcp",to="8.8.8.8:53",type="connect: network is unreachable"} 474
coredns_forward_connection_failures_total{proto="tcp",to="8.8.8.8:53",type="i/o timeout"} 20
coredns_forward_connection_failures_total{proto="udp",to="8.8.8.8:53",type="write: operation not permitted"} 60706
```

This can provide additional insight into resolution failures, particularly if they're intermittent, and without having to parse logs.

README.md and coredns-forward(7) manpage have been updated.

No

### DIFF
--- a/plugin/forward/metrics.go
+++ b/plugin/forward/metrics.go
@@ -28,6 +28,12 @@ var (
 		Buckets:   plugin.TimeBuckets,
 		Help:      "Histogram of the time each request took.",
 	}, []string{"to", "rcode"})
+	RequestFailures = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "forward",
+		Name:      "connect_failures_total",
+		Help:      "Counter of connection failures per remote IP address.",
+	}, []string{"to", "proto", "type"})
 	HealthcheckFailureCount = promauto.NewCounterVec(prometheus.CounterOpts{
 		Namespace: plugin.Namespace,
 		Subsystem: "forward",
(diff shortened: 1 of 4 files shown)


plugin/forward/metrics.go
```diff
@@ -28,6 +28,12 @@ var (
 		Buckets:   plugin.TimeBuckets,
 		Help:      "Histogram of the time each request took.",
 	}, []string{"to", "rcode"})
+	RequestFailures = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "forward",
+		Name:      "connect_failures_total",
+		Help:      "Counter of connection failures per remote IP address.",
+	}, []string{"to", "proto", "type"})
```
SuperQ, line 36: `type` is too generic a label name.

But in this case it seems to be the error message. This is probably not a label we want here, as it is likely to generate too much label cardinality.

Please remove the error code from the labels.

SuperQ (changes requested):

SuperQ:
Thanks, I like the counter. But remember that metrics are designed to help us determine when there is a problem, not diagnose all problems. We want to keep cardinality to the bare minimum necessary.

plugin/forward/metrics.go
chrisohaver: Perhaps we could do our own categorization of error types into a small number of values (2-4). Without the error type, I think this ends up being redundant with other existing metrics.

SuperQ: Yea, I would be more comfortable with a stronger filter.

PR #5899, opened 2023-02-05
plugin/dnssec: on delegation, sign DS or NSEC of no DS.

jeremiejig:
* When returning NS for delegation point, we sign any DS Record or if not found we generate a NSEC proving absence of DS. This follow behaviour describe in rfc4035 (Section 3.1.4)
* DS request at apex behave as before.
* Fix edge case of requesting NSEC which prove that NSEC does not exist.

Signed-off-by: Jeremiejig <me@jeremiejig.fr>

This commit fix incorrect behaviour of plugin/dnssec which lead to a **broken** chain of trust 
for an unsigned delegated zone.
It was also the occasion to properly add DS/NSEC RRSig signature when answering a delegation
matching [rfc4035](https://www.rfc-editor.org/rfc/rfc4035) (See section 3.1.4)

None

### DIFF
--- a/plugin/dnssec/black_lies.go
+++ b/plugin/dnssec/black_lies.go
@@ -1,6 +1,8 @@
 package dnssec
 
 import (
+	"strings"
+
 	"github.com/coredns/coredns/plugin/pkg/response"
 	"github.com/coredns/coredns/request"
 
@@ -11,15 +13,27 @@ import (
 // See https://tools.ietf.org/html/draft-valsorda-dnsop-black-lies-00
 // For example, a request for the non-existing name a.example.com would
 // cause the following NSEC record to be generated:
+//
 //	a.example.com. 3600 IN NSEC \000.a.example.com. ( RRSIG NSEC ... )
+//
 // This inturn makes every NXDOMAIN answer a NODATA one, don't forget to flip
 // the header rcode to NOERROR.
 func (d Dnssec) nsec(state request.Request, mt response.Type, ttl, incep, expir uint32, server string) ([]dns.RR, error) {
 	nsec := &dns.NSEC{}
 	nsec.Hdr = dns.RR_Header{Name: state.QName(), Ttl: ttl, Class: dns.ClassINET, Rrtype: dns.TypeNSEC}
 	nsec.NextDomain = "\\000." + state.QName()
+	if state.QName() == "." {
+		nsec.NextDomain = "\\000." // If You want to play as root server
+	}
 	if state.Name() == state.Zone {
 		nsec.TypeBitMap = filter18(state.QType(), apexBitmap, mt)
+	} else if mt == response.Delegation || state.QType() == dns.TypeDS {
+		nsec.TypeBitMap = delegationBitmap[:]
+		if mt == response.Delegation {
+			labels := dns.SplitDomainName(state.QName())
+			labels[0] += "\\000"
+			nsec.NextDomain = strings.Join(labels, ".") + "."
+		}
 	} else {
 		nsec.TypeBitMap = filter14(state.QType(), zoneBitmap, mt)
 	}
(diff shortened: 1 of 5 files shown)


jeremiejig:
If need be I can provide some test script to illustrate the main problem solved by this patch.

My main motivation behind this patch was the fact that when we try to resolve a dns query delegated by a dnssec trusted coredns server to an unsigned delegated domain, we get this error with delv:
`;; resolution failed: broken trust chain`, which translate to no resolution with dnssec enforced resolver (named, unbound, knot-resolver will respond with SERVFAIL).

The previous behaviour of a resolver prior to this patch (any extra glue is omitted):
* Question: `web.sub.example.com. IN A flags: dodnssec`
* Resolver receive normal dnssec response from `.` and `.com` to `example.com. NS`, meaning we have a chain of DS and DNSKEY to trust from root to `example.com.`
* Resolver ask the following to `example.com.` coredns nameserver:
  * Question: `example.com. IN NS` we receive the following RR in answer section:
    * `example.com. IN NS ns.example.com.`
    * `example.com. IN RRSIG A ...`
  * Question: `example.com. IN DNSKEY`
    * `example.com. IN DNSKEY ...` any number of DNSKEY
    * `example.com. IN RRSIG DNSKEY` any number of RRSIG signed by CSK or KSK.
  * The resolver now trust `example.com.` with the *DS* record it received at the same time as *NS* record from `com.` nameserver.
  * Question: `sub.example.com. IN A`, we receive the following in authority section:
    * `sub.example.com. IN NS ns.sub.example.com.` (we do not have a DS record, as the domain is unsigned)
    * No signature for a delegation ([rfc4035#section-2.2](https://www.rfc-editor.org/rfc/rfc4035#section-2.2))
  * As we are a dnssec aware resolver we query the **missing** `DS` record for `sub.example.com.` (this is even described in [rfc4035#section-4.2](https://www.rfc-editor.org/rfc/rfc4035#section-4.2)) we received the following in authority section:
    * `example.com. IN SOA ns.example.com. ...`
    * `example.com. IN RRSIG SOA ...`
    * `sub.example.com. IN NSEC \000.example.com. A AAAA NSEC RRSIG TXT...` but no SOA, **NS** nor DS in the type list.
    * `sub.example.com. IN RRSIG NSEC ...`
* Resolver may have ask the following to `sub.example.com.` name server prior to asking for `sub.example.com. DS` at `example.com.` coredns
  * Question: `web.sub.example.com. IN A flags: dodnssec`
    * `web.sub.example.com. IN A 192.0.2.10` (with or without RRSIG does not matter)
* We now have all the pieces to trust unsigned answer from `sub.example.com.`, but we have an issue with the following record:
  * `sub.example.com. IN NSEC \000.example.com. A AAAA NSEC RRSIG...(but no NS)` this response authenticated by the associated RRSIG prove effectively that *DS* record does not exist, but the absence of **NS** in the type list also prove that **`sub.example.com.`** is not a delegation point, which force the resolver to not trust the previous delegation response (there cannot be an unsigned delegation at `sub.example.com.` in dnssec sense).

chrisohaver:
Thanks for the PR, and _thanks very much for the detailed explanation with examples and RFC section reference_.  Just acknowledging for now. I will try to look at this closely soon.

Isolus:
I'm sorry for responding so late. It took me some time to look at the details. I would also like to thank you for the PR and the detailed explanation.

I think the pull request solves the problem.

But since our way to use NSEC type bit maps is rather a trick related to black lies and not standardized (but standard compliant), I would be happy if someone else could take a look at it as well so that no edge case is overlooked.

chrisohaver:
@Isolus, I don't think anyone else is going to be able to review this soon.  It's been almost 2 months.

Are you comfortable as is to merge this?  If so, please do.

chrisohaver:
Oh - and @jeremiejig, please sign the commits. Thanks!

Isolus:
@chrisohaver Yes, I looked at it again and I'm comfortable with it. I'll merge it as soon as @jeremiejig signs the commit.

jeremiejig:
@Isolus @chrisohaver I just rebase my branch to have only my commit with the sign-off.

PR #5906, opened 2023-02-07
plugin/loadbalance: Improve weights update

gabordozsa:
Make the weights update atomic which only happens if the weight file is parsed without any error. Also, add missing check to reject zero weight values.

Signed-off-by: Gabor Dozsa <gabor.dozsa@ibm.com>

It improves the weights update for the weighted policy in the loadbalance plugin. Two main changes: 
-  make the weights update "atomic" in the sense that it happens as a single assignment after the weight file get parsed successfully
-  add a missing check in the weight file parser to reject zero weight values. (Valid weight values are [1,255] as stated in the README)  

None

None

No

### DIFF
--- a/plugin/loadbalance/weighted.go
+++ b/plugin/loadbalance/weighted.go
@@ -254,26 +254,26 @@ func (w *weightedRR) updateWeights() error {
 	scanner := bufio.NewScanner(&buf)
 
 	// Parse the weight file contents
-	err = w.parseWeights(scanner)
+	domains, err := w.parseWeights(scanner)
 	if err != nil {
 		return err
 	}
+
+	// access to weights must be protected
+	w.mutex.Lock()
+	w.domains = domains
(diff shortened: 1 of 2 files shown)


gabordozsa:
@chrisohaver Could you please take a look at this PR? The  go-bencher check failed due to some connection issue and I do not know how to re-run that check.

chrisohaver:
AFAIK, There is no way to re-run go-bencher tests without recommitting. (there is a re-run button in actions, but it doesn't appear to work).  Try amending your last commit message and force push it to the PR.

chrisohaver (approved):
> make the weights update "atomic" in the sense that it happens as a single assignment after the weight file get parsed successfully

OK.  Not technically "atomic" as I understand it, but this change is good because it reduces the amount of time the weights are locked by deferring the assignment until after parsing the file.

PR #5907, opened 2023-02-08
Close the body after reading the response body information

Fish-pro:
Signed-off-by: Fish-pro <zechun.chen@daocloud.io>

Close the body after reading the response body information


chrisohaver (approved):
Looks harmless, changes in tests only.

PR #5908, opened 2023-02-08
Use http constants instead of string

Fish-pro:
Signed-off-by: Fish-pro <zechun.chen@daocloud.io>

Using strings as arguments to http package functions is error-prone, so call http package functions in the standard way


bradbeam (approved):

SuperQ (approved):
Thanks!

PR #5931, opened 2023-02-23
Introduces the ARG in Dockerfile  

yashsingh74:
The PR introduces the ARG in the Dockerfile which will help to updated the values like images using the `--build-arg` without updating the Dockerfile.


Tantalor93:
> The PR introduces the ARG in the Dockerfile which will help to updated the values like images using the --build-arg without updating the Dockerfile.

hello, thanks for the PR, but sadly the reasoning is not cleat to me, what issue does it solve, or what is the benefit?

yashsingh74:
@Tantalor93 
I reason the introduced this change is because there are some cases where I want to build the coredns not from scratch or  from different BASE_IMAGE. Now with the --build-args parameter I don't need to change the Dockerfile and hardcoding the values instead I can build the images using the CLI by passing the `--build-args` parameters like [here](https://docs.docker.com/engine/reference/commandline/build/#build-arg). This will also helps to keep the code clean without manually overwriting the values.

mariuskimmina:
Thanks for the PR @yashsingh74 . 
@Tantalor93  it sounds reasonable to me, e.g. when you want to spin up a CoreDNS container to test certain things and you want to get a shell in the container, from scratch gives you basically nothing and in such a case quickly building a CoreDNS image with a different base could be nice.

ivyostosh:
Any update on when this PR can be merged?

SuperQ:
This needs a rebase.

SuperQ (approved):
Thanks!

ivyostosh:
I wonder if the failing CI job just needs a re-run?

yashsingh74:
@SuperQ can you help in merging this PR.

Tantalor93 (approved):
LGTM, thx!

Tantalor93 (approved):
LGTM, thx!

Issue #5895, opened 2023-02-04
Forward using specific interface


imacks:
Example: machine has 2 net iface, and I want the forward plugin to use eth0 when querying tls://8.8.8.8 and eth1 for tls://9.9.9.9.

I didn't look closely but it seems to me miekg/dns pkg doesn't allow you to set the LocalAddr when sending a dns query?

Without this feature I have to use iptables, making the setup more complicated.

aashh:
This is addressed in #5933

PR #5933, opened 2023-02-25
plugin/forward: Add local address option

aashh:

This adds a `local` option to the `forward` plugin, to allow specifying the source address to use for outbound queries. 

This is useful on systems where a destination is reachable through multiple interfaces. It's comparable to the BIND `query-source address` option, the Unbound `outgoing-interface` option, etc.

#5895 

Update to plugin README, included in this change.

No

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -50,6 +50,7 @@ forward FROM TO... {
     policy random|round_robin|sequential
     health_check DURATION [no_rec] [domain FQDN]
     max_concurrent MAX
+    local ADDRESS
 }
 ~~~
 
@@ -95,6 +96,7 @@ forward FROM TO... {
   response does not count as a health failure. When choosing a value for **MAX**, pick a number
   at least greater than the expected *upstream query rate* * *latency* of the upstream servers.
   As an upper bound for **MAX**, consider that each concurrent query will use about 2kb of memory.
+* `local` **ADDRESS** specifies the source address to use for outbound queries.
 
 Also note the TLS config is "global" for the whole forwarding proxy if you need a different
 `tls-name` for different upstreams you're out of luck.
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -50,7 +50,7 @@ forward FROM TO... {
     policy random|round_robin|sequential
     health_check DURATION [no_rec] [domain FQDN]
     max_concurrent MAX
-    local ADDRESS
+    source ADDRESS
 }
 ~~~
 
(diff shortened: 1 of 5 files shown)


aashh:
This change duplicates code from `DialTimeout()` and `DialTimeoutWithTLS()` in https://github.com/miekg/dns because those functions don't offer any way to set the local (source) network address.

The pull request miekg/dns#1426 would allow avoiding this duplication of code on the CoreDNS side.

chrisohaver (commented):
Thanks for contributing. Some thoughts before digging into implementation specifics...

* A trivial matter, but I find `local` to be too generic - other possible names  to consider `source` `bind` `from` ... or maybe longer .. `send_using` `send_from` etc
* I think this syntax will be limiting when multiple upstreams are involved.  It's possible that the upstream dns servers would be on different networks that cannot be reached from a single local address.  Or, it's possible that an admin may want to specify a different source address to use for for each upstream for other reasons.
* FWIW, IIUC think this function is already achievable externally of CoreDNS using routing tools such as iptables. In ithe current state the PR is not complex, but something to consider if this PR becomes more complex - is the convenience worth the added complexity.

aashh:
> * FWIW, IIUC think this function is already achievable externally of CoreDNS using routing tools such as iptables. In ithe current state the PR is not complex, but something to consider if this PR becomes more complex - is the convenience worth the added complexity.

The use case here is for addresses that can be reached via multiple next-hops. In these cases, when we want a particular next-hop to be used, it's typical to add functionality to an application to influence the choice. That is what other DNS servers (and many other mainstream services) offer, so I think it's at least a reasonable thing for CoreDNS to support.

> * I think this syntax will be limiting when multiple upstreams are involved.  It's possible that the upstream dns servers would be on different networks that cannot be reached from a single local address.  Or, it's possible that an admin may want to specify a different source address to use for for each upstream for other reasons.

I'd be interested to hear of a configuration with multiple upstreams that are _both_ reachable via multiple next-hops _and_ required to be reached via _different_ next-hops. The other major DNS servers don't seem compelled to support this, so until a real use case is encountered, is it worth adding complexity (internally and user-facing) to support it?

aashh:
> * A trivial matter, but I find `local` to be too generic - other possible names  to consider `source` `bind` `from` ... or maybe longer .. `send_using` `send_from` etc

Renamed to `source`

chrisohaver:
> I'd be interested to hear of a configuration with multiple upstreams that are both reachable via multiple next-hops and required to be reached via different next-hops. 

That doesn't have to be the case to run into the limitation.  All that is required is your use case (where an upstream is reachable on multiple interfaces), and another upstream on a distinct network interface, e.g. a local lab network.  With this option, the one upstream would not be reachable since the source address is not in the second network.  Essentially, this option requires that all upstreams are reachable from the configured source address.

However, without the flexibility in this option, I think it could still be achieved (albeit awkwardly) by using local forwarding and local server instances - allowing you to segregate upstreams on like networks into separate forward/server instances.  So, since there is a workaround, I think the limitation would be OK.

Please explore the possibility of adding tests for this new feature.

PR #5943, opened 2023-03-02
add support unix socket for GRPC

junhwong:
add support unix socket for GRPC to adapt to some local scenarios that are inconvenient to use TCP

No

### DIFF
--- a/plugin/grpc/README.md
+++ b/plugin/grpc/README.md
@@ -129,6 +129,14 @@ Or with multiple upstreams from the same provider
 }
 ~~~
 
+Forward requests to a local upstream listening on a Unix domain socket.
+
+~~~ corefile
+. {
+    grpc . unix:///path/to/grpc.sock
+}
+~~~
+
 ## Bugs
 
 The TLS config is global for the whole grpc proxy if you need a different `tls_servername` for
--- a/plugin/pkg/parse/host.go
+++ b/plugin/pkg/parse/host.go
@@ -33,6 +33,14 @@ func HostPortOrFile(s ...string) ([]string, error) {
 	var servers []string
 	for _, h := range s {
 		trans, host := Transport(h)
+		if len(host) == 0 {
+			return servers, fmt.Errorf("invalid address: %q", h)
+		}
+
+		if trans == transport.UNIX {
+			servers = append(servers, trans+"://"+host)
+			continue
+		}
 
 		addr, _, err := net.SplitHostPort(host)
 
--- a/plugin/grpc/proxy_test.go
+++ b/plugin/grpc/proxy_test.go
@@ -3,9 +3,15 @@ package grpc
 import (
 	"context"
 	"errors"
+	"net"
+	"os"
+	"path"
 	"testing"
 
+	"github.com/coredns/caddy"
 	"github.com/coredns/coredns/pb"
+	"github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/coredns/coredns/plugin/test"
 
 	"github.com/miekg/dns"
 	"google.golang.org/grpc"
@@ -63,4 +69,57 @@ type testServiceClient struct {
 
 func (m testServiceClient) Query(ctx context.Context, in *pb.DnsPacket, opts ...grpc.CallOption) (*pb.DnsPacket, error) {
 	return m.dnsPacket, m.err
+}
+
+func TestProxyUnix(t *testing.T) {
+	tdir, err := os.MkdirTemp("", "tmp*")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tdir)
+
+	fd := path.Join(tdir, "test.grpc")
+	listener, err := net.Listen("unix", fd)
+	if err != nil {
+		t.Fatal("Failed to listen: ", err)
+	}
+	defer listener.Close()
+
+	server := grpc.NewServer()
+	pb.RegisterDnsServiceServer(server, &grpcDnsServiceServer{})
+
+	go server.Serve(listener)
+	defer server.Stop()
+
+	c := caddy.NewTestController("dns", "grpc . unix://"+fd)
+	g, err := parseGRPC(c)
+
+	if err != nil {
+		t.Errorf("Failed to create forwarder: %s", err)
+	}
+
+	m := new(dns.Msg)
+	m.SetQuestion("example.org.", dns.TypeA)
+	rec := dnstest.NewRecorder(&test.ResponseWriter{})
+
+	if _, err := g.ServeDNS(context.TODO(), rec, m); err != nil {
+		t.Fatal("Expected to receive reply, but didn't")
+	}
+	if x := rec.Msg.Answer[0].Header().Name; x != "example.org." {
+		t.Errorf("Expected %s, got %s", "example.org.", x)
+	}
+}
+
+type grpcDnsServiceServer struct {
+	pb.UnimplementedDnsServiceServer
+}
+
+func (*grpcDnsServiceServer) Query(ctx context.Context, in *pb.DnsPacket) (*pb.DnsPacket, error) {
+	msg := &dns.Msg{}
+	msg.Unpack(in.GetMsg())
+	answer := new(dns.Msg)
+	answer.Answer = append(answer.Answer, test.A("example.org. IN A 127.0.0.1"))
+	answer.SetRcode(msg, dns.RcodeSuccess)
+	buf, _ := answer.Pack()
+	return &pb.DnsPacket{Msg: buf}, nil
 }
--- a/plugin/grpc/setup_test.go
+++ b/plugin/grpc/setup_test.go
@@ -25,6 +25,7 @@ func TestSetup(t *testing.T) {
 		{"grpc . 127.0.0.1:8080", false, ".", nil, ""},
 		{"grpc . [::1]:53", false, ".", nil, ""},
 		{"grpc . [2003::1]:53", false, ".", nil, ""},
+		{"grpc . unix:///var/run/g.sock", false, ".", nil, ""},
 		// negative
 		{"grpc . a27.0.0.1", true, "", nil, "not an IP"},
 		{"grpc . 127.0.0.1 {\nblaatl\n}\n", true, "", nil, "unknown property"},
--- a/plugin/pkg/parse/host_test.go
+++ b/plugin/pkg/parse/host_test.go
@@ -58,6 +58,16 @@ func TestHostPortOrFile(t *testing.T) {
 			"",
 			true,
 		},
+		{
+			"unix:///var/run/g.sock",
+			"unix:///var/run/g.sock",
+			false,
+		},
+		{
+			"unix://",
+			"",
+			true,
+		},
 	}
 
 	err := os.WriteFile("resolv.conf", []byte("nameserver 127.0.0.1\n"), 0600)
--- a/plugin/pkg/parse/transport.go
+++ b/plugin/pkg/parse/transport.go
@@ -27,6 +27,9 @@ func Transport(s string) (trans string, addr string) {
 		s = s[len(transport.HTTPS+"://"):]
 
 		return transport.HTTPS, s
+	case strings.HasPrefix(s, transport.UNIX+"://"):
+		s = s[len(transport.UNIX+"://"):]
+		return transport.UNIX, s
 	}
 
 	return transport.DNS, s
--- a/plugin/pkg/transport/transport.go
+++ b/plugin/pkg/transport/transport.go
@@ -6,6 +6,7 @@ const (
 	TLS   = "tls"
 	GRPC  = "grpc"
 	HTTPS = "https"
+	UNIX  = "unix"
 )
 
 // Port numbers for the various transports.


plugin/pkg/parse/host.go
```diff
@@ -63,6 +63,12 @@ func HostPortOrFile(s ...string) ([]string, error) {
 		}
 
 		if net.ParseIP(stripZone(addr)) == nil {
+			if strings.HasPrefix(host, "unix://") {
+				// TODO 检查是否存在
```
inigohu: Could you write it in English?

junhwong: ok, changed and commit

plugin/pkg/parse/host.go
```diff
@@ -63,6 +63,12 @@ func HostPortOrFile(s ...string) ([]string, error) {
 		}
 
 		if net.ParseIP(stripZone(addr)) == nil {
+			if strings.HasPrefix(host, "unix://") {
```
inigohu: We could include the unix schema in [plugin/pkg/parse/transport.go](https://github.com/coredns/coredns/blob/master/plugin/pkg/parse/transport.go)

go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/Azure/azure-sdk-for-go v68.0.0+incompatible
 	github.com/Azure/go-autorest/autorest v0.11.28
 	github.com/Azure/go-autorest/autorest/azure/auth v0.5.12
-	github.com/antonmedv/expr v1.12.0
+	github.com/antonmedv/expr v1.12.1
```
inigohu: why this change?

plugin/pkg/parse/host.go
```diff
@@ -33,6 +33,15 @@ func HostPortOrFile(s ...string) ([]string, error) {
 	var servers []string
 	for _, h := range s {
 		trans, host := Transport(h)
+		if len(host) == 0 {
+			return servers, fmt.Errorf("invalid address: %q", h)
+		}
+
+		switch trans {
```
inigohu: Use the `if` statement for a single case.

inigohu:
Could you squash your commits together to make it more concise?

go.mod
junhwong: A misoperation in solving DCO

inigohu: If this is an unrelated bug fix, please open a new PR

junhwong: OK, that's no problem. although I don't think it has any impact.

junhwong:
Any other questions? @inigohu

inigohu:
>Any other questions? @inigohu

Please resolve pending comments

plugin/pkg/parse/host.go
junhwong: see plugin/pkg/parse/host_test.go:66

junhwong:
All resolved.
In addition to `go.mod`, if there is no problem, I will submit a new PR
@inigohu

chrisohaver:
Can you clean the commit history?  You have grabbed an unrelated commit (453f89d70ad9bcf7accf0810f09954fb302fcafe)

inigohu:
Can you include an example of a UNIX socket in `coredns/plugin/grpc/README.md`?

junhwong:
Already added @inigohu 

I don't know how to separate them @chrisohaver

junhwong:
Why use **~~~**  instead of **```** in README.md? 
@inigohu

chrisohaver:
> I don't know how to separate them @chrisohaver

Looking at your fork, I see you are working directly in the master branch ...

Try to do a `rebase` of your master branch to remove the redundant commit.  Try `git rebase -i HEAD~2`, and then drop the unwanted commit.  Then force push master to your fork. This should probably fix things. But git can be confusing.

junhwong:
> 

tks @chrisohaver ,  got it! 


cc @inigohu

inigohu:
>Why use ~~~ instead of ``` in README.md?

They are the same. https://www.markdownguide.org/extended-syntax/#fenced-code-blocks
>Depending on your Markdown processor or editor, you’ll use three backticks (```) or three tildes (~~~)

plugin/grpc/README.md
```diff
@@ -129,7 +129,15 @@ Or with multiple upstreams from the same provider
 }
 ~~~
 
+For local upstreams, can using `UNIX Socket` protocol.
+
+~~~ corefile
+. {
+    grpc . path/to/grpc.sock
```
chrisohaver: I think this example would fail to parse correctly. shouldn't it be `grpc . unix:///path/to/grpc.sock`?

plugin/grpc/README.md
```diff
@@ -129,7 +129,15 @@ Or with multiple upstreams from the same provider
 }
 ~~~
 
+For local upstreams, can using `UNIX Socket` protocol.
```
chrisohaver:
```suggestion
Forward requests to a local upstream listening on a unix socket `/path/to/grpc.sock`.
```

chrisohaver (changes requested):

plugin/grpc/README.md
junhwong: No, you're wrong

Grpc is an outer application protocol. The specific communication protocol is unix. They are not a concept.

junhwong: No, you're wrong

Grpc is an outer application protocol. The specific communication protocol is unix. They are not a concept

chrisohaver: I think you're not understanding my comment.  Let me rephrase...

Will your example parse without error?  In your unit test case, the sock path includes the unix uri prefix.  If this example is valid, please add this style of config to the unit tests.

IOW: If the URI prefix is not required, please add a unit test that does not include the unix uri prefix.

junhwong: I understand your maind. What I want to express is that what you said is not accurate.
The  `unix://` prefix is necessary, otherwise the resolution will conflict, and the `resolve.conf` definition will be found by default.

chrisohaver: > The unix:// prefix is necessary

If it is necessary, your example will fail to work as described.  It would assume the path is to a resolv.conf file, parse it as such, and thus be an invalid example of how to use a unix socket.

chrisohaver:
```suggestion
    grpc . unix:///path/to/grpc.sock
```

chrisohaver: Please let reviewers resolve conversations. Thanks.

junhwong: That was my negligence ...

junhwong (commented):
.

chrisohaver:
Curious how you have tested this?

junhwong:
> Curious how you have tested this?

1. Unit tests: eg.  plugin/grpc/setup_test.go
2.  What you mentioned above is only a document issue
3.  This is a simple PR, and it has already worked on my application scenario

chrisohaver:
> it has already worked on my application scenario

This is what I am asking about.  In that scenario, what are you using as the grpc local socket listener?

junhwong:
We use [dns lib](//github.com/miekg/dns) to implementation grpc(`dns.proto`)interface . @chrisohaver

plugin/pkg/parse/host.go
inigohu: I meant that comparing only the `transport.UNIX` case, it would be better to use an `if` statement instead of `swtich`.

```
if trans == transport.UNIX {
	servers = append(servers, trans+"://"+host)
        continue
}
```

junhwong: ...  I misinterpreted it as  `single case test`

junhwong: I am accustomed to use `switch`, but i accept your propose and will submit it later.

chrisohaver:
Can you also extend the CoreDNS gRPC server to allow listening on a unix socket?

junhwong:
> Can you also extend the CoreDNS gRPC server to allow listening on a unix socket?

Personally, I think the scene is limited.
CoreDNS is mostly used as a cluster upstream or local cache front-end.

chrisohaver:
> > Can you also extend the CoreDNS gRPC server to allow listening on a unix socket?
> 
> Personally, I think the scene is limited.

As is forwarding gRPC to a local unix socket.

chrisohaver:
I'd like to validate that this feature works as proposed.  Can you recommend to a method for testing it?

junhwong:
> Can you recommend to a method for testing it?

test what?
if it's CoreDNS extend, i don't have a particularly  good way,  you can try two stages:
```
domain-up:unix://path/to { ... }    # for local extend
:1053{
    grpc domain-up unix://path/to 
}
```

junhwong:
extend CoreDNS for support unix socket is also easy，like the example above
`domain-up:unix://path/to { ... }`

I can submit a new PR after this PR is completed, but I still do not recommend extending. There are no instance application scenarios.

junhwong:
@inigohu @chrisohaver 
What is the next process for this PR?

chrisohaver:
> test what?

Your proposed changes. The changes you are requesting we pull into coredns in this pull request. I’d like to test them to make sure they work.

junhwong:
For this PR? They can definitely work!
The unit test can confirm that the configuration has passed, which is enough。
If you want to reconfirm, you can implement a minimum server, the following is my code section, which you can refer to:

```go
func (s *grpcDnsServiceServer) Query(ctx context.Context, in *pb.DnsPacket) (*pb.DnsPacket, error) {
	msg := &dns.Msg{}
	if err := msg.Unpack(in.GetMsg()); err != nil {
		log.Error(err)
		return nil, err
	}
	w := &grpcResponseWriter{
		local: s.local,
	}
	if p, ok := peer.FromContext(ctx); ok {
		w.remote = p.Addr
	}
	s.handler.ServeDNS(w, msg)
	return &pb.DnsPacket{Msg: w.packed}, nil
}
```

```go
listener, err := net.Listen("unix", "path")
server := grpc.NewServer(opts...)
pb.RegisterDnsServiceServer(server, impl)
server.Serve(listener)
```


in `Query` func , print any msg indicates success

chrisohaver:
> For this PR? They can definitely work! 

> The unit test can confirm that the configuration has passed, which is enough

IIUC, The unit tests in your PR merely verify configuration parsing - not functionality.

> ... you can implement a minimum server, the following is my code section, which you can refer to:
> 
> ```go
> func (s *grpcDnsServiceServer) Query(ctx context.Context, in *pb.DnsPacket) (*pb.DnsPacket, error) {
> 	msg := &dns.Msg{}
> 	if err := msg.Unpack(in.GetMsg()); err != nil {
> 		log.Error(err)
> 		return nil, err
> 	}
> 	w := &grpcResponseWriter{
> 		local: s.local,
> 	}
> 	if p, ok := peer.FromContext(ctx); ok {
> 		w.remote = p.Addr
> 	}
> 	s.handler.ServeDNS(w, msg)
> 	return &pb.DnsPacket{Msg: w.packed}, nil
> }
> ```
> 
>  ```go
> listener, err := net.Listen("unix", "path")
> server := grpc.NewServer(opts...)
> pb.RegisterDnsServiceServer(server, impl)
> server.Serve(listener)
> ```


Cool.  Can you integrate this into a test in /test ?

junhwong:
I don't think it's necessary. Testing the configuration is enough

1. Connectivity is done by GRPC itself
2. Protocol correctness is accomplished by the CoreDNS core
3. Combination They were completed by the `Grpc plugin` and have now been completed very well

The test is repeated again, which does not make much sense

chrisohaver:
I'm not comfortable merging this without a functional test.  But if another maintainer is OK with the PR as is, then it can merge.

johnbelamaric:
@chrisohaver agreed. 

Recall, the point of the test being automated is to ensure it still works when things change. It is certainly possible for upstream dependencies to change, and break our functionality. Without a functional automated test, we would not find this when we bump the dependency version, and could release a non-working binary.

junhwong (commented):
TBR  @chrisohaver @inigohu

plugin/pkg/parse/host.go
inigohu: PTAL

junhwong: It’s really strange.I already had this modified and submitted it. I even double-checked it.

junhwong: @inigohu

test/grpc_test.go
```diff
@@ -56,3 +59,75 @@ func TestGrpc(t *testing.T) {
...
+	msg.Unpack(in.GetMsg())
+	answer := new(dns.Msg)
+	answer.SetRcode(msg, dns.RcodeSuccess)
+	buf, _ := answer.Pack()
+	return &pb.DnsPacket{Msg: buf}, nil
+}
+
+func TestGrpcUnix(t *testing.T) {
+
+	fd := "/tmp/test745260595.grpc"
```
inigohu: Take a look at how [unixsock_test.go](https://go.dev/src/net/unixsock_test.go) does it.
[testUnixAddr](https://go.dev/src/net/mockserver_test.go) uses `os.MkdirTemp` to get a name that is unique

junhwong: ok

test/grpc_test.go
```diff
@@ -56,3 +59,77 @@ func TestGrpc(t *testing.T) {
...
+	buf, _ := answer.Pack()
+	return &pb.DnsPacket{Msg: buf}, nil
+}
+
+func TestGrpcUnix(t *testing.T) {
+	fd, err := os.MkdirTemp("", "*.grpc")
+	if err != nil {
+		t.Fatal(err)
+	}
+	os.Remove(fd)
```
inigohu, line 81: No need to remove it before use.

test/grpc_test.go
```diff
@@ -56,3 +59,77 @@ func TestGrpc(t *testing.T) {
...
+	msg := &dns.Msg{}
+	msg.Unpack(in.GetMsg())
+	answer := new(dns.Msg)
+	answer.SetRcode(msg, dns.RcodeSuccess)
+	buf, _ := answer.Pack()
+	return &pb.DnsPacket{Msg: buf}, nil
+}
+
+func TestGrpcUnix(t *testing.T) {
+	fd, err := os.MkdirTemp("", "*.grpc")
```
inigohu: `MkdirTemp` creates the directory. It uses an empty pattern to create only the directory and uses `filepath.Join(d, "sock")` to create the file.

test/grpc_test.go
inigohu: No need to clean before use

junhwong: PTAL @inigohu

inigohu:
LGTM. 
WDYT @chrisohaver ?

chrisohaver:
> LGTM.
> WDYT @chrisohaver ?

I think it looks good. Thanks for adding the test!

One comment I have is that the organization of the added test is a tad confusing.

The existing test in `grpc_test.go`  (`TestGrpc`) is there to test the grpc _server_ - not the grpc plugin.

The test added (`TestGrpcUnix`) is there to test the grpc _plugin_ .
And the added test arbitrarily uses a grpc listener, which isn't really necessary, and adds a little complexity to the test. That isn't a huge deal, but it makes the goal of the test a bit less clear. I suspect this was just copied from `TestGrpc`.

I think renaming the test to `TestGrpcPluginUnix` would clarify things a little. Maybe also add comment on what is being tested.
And ideally, the coredns server should be listening over normal dns, not grpc.

chrisohaver:
@inigohu, This is a bit off topic, but locally I modified the test in this PR to test concurrent queries ( just for peace of mind and to satisfy some doubts stemming from my ignorance of how grpc and unix domain sockets work).

While I was testing that I noticed that grpc plugin calls `state.Match(ret)`. Under what circumstances would the return not match?  Should we instead check that the Id of the response matches?

junhwong:
Thumbs up for your meticulousness! @chrisohaver 

I have considered where to place it, but this test is not within the scope of a unit test, so it can only be considered as an integration test. Therefore, it is placed here now.

junhwong:
I think there is also ambiguity in TestGrpcPluginUnix. When taken apart, Unix itself becomes a plugin, but this is clearly not the case, so I haven't changed it for now

junhwong:
I just submitted it due to formatting issues during workflow testing. Please take a look again @inigohu @chrisohaver

chrisohaver (changes requested):
Please rename the TestGrpcUbix test to TestGrpcPluginUnix . 
Please add a code comment that describes what is being tested by that test .
In that test, please make the coredns server listen on plain dns, and have the client request over plain dns, not grpc.

inigohu:
>While I was testing that I noticed that grpc plugin calls `state.Match(ret)`. Under what circumstances would the return not match? 

The DNS service could return a malformed response and we should check this.

>Should we instead check that the Id of the response matches?

After reading [this RFC](https://datatracker.ietf.org/doc/html/rfc1035#section-7.3), I would definitely include the ID check in the first place.
>Processing responses
...
The recommended strategy is to do a preliminary matching using the ID
field in the domain header, and then to verify that the question section
corresponds to the information currently desired.

junhwong:
> In that test, please make the coredns server listen on plain dns, and have the client request over plain dns, not grpc.

`plain dns`?  I don't understand what you mean

@chrisohaver

chrisohaver:
> plain dns? I don't understand what you mean

I mean dns over udp, not dns over grpc.

junhwong:
What do you want to test? 
Can you write a simple demo to help me understand?

chrisohaver:
Currently your unit test is structured as follows:

1. A client makes a gRPC dns query to a CoreDNS gRPC server.
2. The _grpc_ plugin proxy forwards the request over gRPC on a unix domain socket ...

Please change the first step to ...

1. A client makes a **UDP** DNS query to a CoreDNS **UDP** server

junhwong:
Why use `UDP DNS`?
I still don't understand what you want to test or achieve.

chrisohaver:
The aim is to simplify the test, removing unnecessary complications that do not improve coverage. Actually an even better way would be to create an instance of the grpc plugin, and call ServeDNS, passing the dns query message to it directly. There are some examples of this for other plugins, eg forward.

junhwong:
Can you give an example or extract some relevant code fragments?

chrisohaver:
TestProxy does this with the _forward_ plugin:

instantiate the plugin (in this example, the plugin is created by parsing a config string, but should also be able to do this by just creating the grpc struct and setting values directly if easier): 

https://github.com/coredns/coredns/blob/604a902e2c7e0317aecaa3666124079c75a31573/plugin/forward/proxy_test.go#L23-L30

create and send a request to the plugin: 

https://github.com/coredns/coredns/blob/604a902e2c7e0317aecaa3666124079c75a31573/plugin/forward/proxy_test.go#L32-L36

junhwong (commented):
PTAL @chrisohaver

chrisohaver (approved):
LGTM. Thanks!

plugin/grpc/README.md
```diff
@@ -129,7 +129,15 @@ Or with multiple upstreams from the same provider
 }
 ~~~
 
+Forward requests to a local upstream listening on a `unix socket`
```
chrisohaver:
```suggestion
Forward requests to a local upstream listening on a Unix domain socket.
```

junhwong:
PTAL @chrisohaver @inigohu @jameshartig @dilyevsky @SuperQ @Isolus @greenpau @bradbeam @Tantalor93 @pmoroney @rajansandeep
Thanks!

plugin/grpc/README.md
```diff
@@ -129,7 +129,15 @@ Or with multiple upstreams from the same provider
 }
 ~~~
 
+Forward requests to a local upstream listening on a `unix domain socket`
```
chrisohaver:
```suggestion
Forward requests to a local upstream listening on a Unix domain socket.
```

chrisohaver (changes requested):
LGTM, Just minor readme formatting tweaks.  Thanks!

chrisohaver (approved):

Tantalor93 (approved):

Issue #5910, opened 2023-02-09
[Bug] tls plugin does not use path set by root plugin


rahilarious:
**What happened**:
Following config does not load /config/cert.pem but tries to load /cert.pem (from execution dir)
```
tls://example.com:853 {
      root /config
     tls cert.pem key.pem
    whoami
}
```

**What you expected to happen**:
It should be loading from /config/cert.pem

**Anything else we need to know?**:
Probably happens because https://github.com/coredns/coredns/blob/055b2c31a9cf28321734e5f71613ea080d216cd3/plugin.cfg#L25 is loading before https://github.com/coredns/coredns/blob/055b2c31a9cf28321734e5f71613ea080d216cd3/plugin.cfg#L30 in plugin.cfg

**Environment**:
- the version of CoreDNS: 1.10.1
- OS: official docker image

mariuskimmina:
I can confirm the issue and I also tried to rebuild coredns with `root:root` before `tls:tls` in my `plugin.cfg` but without success, it still looks for the cert in the current directory.

mariuskimmina:
The file plugin actively checks for `config.Root` in it's `setup.go` file - the tls plugin does not. I think I can make a fix for this in the next few days - @chrisohaver can you assign this to me?

rahilarious:
> The file plugin actively checks for `config.Root` in it's `setup.go` file - the tls plugin does not. I think I can make a fix for this in the next few days - @chrisohaver can you assign this to me?

Hey, thank you for looking into this.

PR #5944, opened 2023-03-05
plugin/tls/etcd/forward/grpc: respect the path specified by root plugin

mariuskimmina:

Use the path specified by the root plugin to look for certs

#5910 

For this to work, the plugin root must come before the tls plugin in plugin.cfg. 
There may also be folks who are using the root and tls plugin at the moment with their current behaviour, this change could lead to problems for them.

### DIFF
--- a/plugin.cfg
+++ b/plugin.cfg
@@ -19,6 +19,7 @@
 # Local plugin example:
 # log:log
 
+root:root
 metadata:metadata
 geoip:geoip
 cancel:cancel
@@ -26,7 +27,6 @@ tls:tls
 reload:reload
 nsid:nsid
 bufsize:bufsize
-root:root
 bind:bind
 debug:debug
 trace:trace
--- a/plugin/etcd/setup.go
+++ b/plugin/etcd/setup.go
@@ -2,6 +2,7 @@ package etcd
 
 import (
 	"crypto/tls"
+	"path/filepath"
 
 	"github.com/coredns/caddy"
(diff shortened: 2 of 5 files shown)


mariuskimmina:
I believe this covers all cases now - I haven't manually tested all of them.
I think this change makes sense, but it could potentially be breaking for some setups. 
What do you think @chrisohaver ?

chrisohaver:
>There may also be folks who are using the root and tls plugin at the moment with their current behaviour, this change could lead to problems for them.

Tagging this for 1.11.0 milestone, since it is potentially backward incompatible.

chrisohaver:
I merged this prematurely...  It needs an update to the root plugin README, since _root_ is no longer specific to finding zone files (as described in the README).  Updated the PR title to reflect that this changes not just _tls_ plugin behavior.
Also `core/dnsserver/zdirectives.go` is not updated to reflect the changes to plugin order.

And I'm thinking the _root_ unit tests should be added or extended to validate new behavior.

chrisohaver:
@mariuskimmina, My apologies for prematurely merging this.  Since this PR has already been merged, it cannot be reopened. Please reopen a new PR when you get a chance that addresses the above (unit tests, README update, and zdirectives update).  Thanks!

mariuskimmina:
@chrisohaver I should be able to get to that in the next 3 to 4 days

chrisohaver:
> @chrisohaver I should be able to get to that in the next 3 to 4 days

Awesome thanks!

PR #5951, opened 2023-03-06
plugin/forward: Allow Proxy to be used outside of forward plugin.

patdowney:
This pull request changes the visibility of the `Start` method and `Options` struct so that they can be used outside the `forward` plugin. An example use case is for other plugins to proxy requests to dynamic upstream hosts.

None

`Options` has been updated to document it's fields.

No backward incompatible changes as members were not available for use outside of the `forward` plugin.

### DIFF
--- a/plugin/forward/connect.go
+++ b/plugin/forward/connect.go
@@ -72,14 +72,14 @@ func (t *Transport) Dial(proto string) (*persistConn, bool, error) {
 }
 
 // Connect selects an upstream, sends the request and waits for a response.
-func (p *Proxy) Connect(ctx context.Context, state request.Request, opts options) (*dns.Msg, error) {
+func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options) (*dns.Msg, error) {
 	start := time.Now()
 
 	proto := ""
 	switch {
-	case opts.forceTCP: // TCP flag has precedence over UDP flag
+	case opts.ForceTCP: // TCP flag has precedence over UDP flag
 		proto = "tcp"
-	case opts.preferUDP:
+	case opts.PreferUDP:
 		proto = "udp"
 	default:
 		proto = state.Proto()
--- a/plugin/forward/dnstap.go
+++ b/plugin/forward/dnstap.go
@@ -13,7 +13,7 @@ import (
 )
 
 // toDnstap will send the forward and received message to the dnstap plugin.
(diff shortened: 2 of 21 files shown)


patdowney:
Perhaps an alternative location for a sharable `Proxy` struct would be to move into `plugin/pkg/proxy`?

chrisohaver:
> Perhaps an alternative location for a sharable `Proxy` struct would be to move into `plugin/pkg/proxy`?

Yes, that would be better.

chrisohaver:
is TestHealthMaxFails getting dropped here intentionally?  Or did it move and I am not seeing its new location.

patdowney:
@chrisohaver thanks for picking that up, I didn't miss it out on purpose! I've reinstated `TestHealthMaxFails`

patdowney:
Please don't merge yet, have just realised I've moved the timeout variables that used to make the tests fastish and not made the behaviour  of the `Forward` tests similar after moving `Proxy` into it's own location

patdowney:
OK, I've made some changes so the tests take a comparable time.

```
user@host forward % git branch --show-current
master
user@host  forward % go test ./... -count=1
ok      github.com/coredns/coredns/plugin/forward       2.750s
```

```
user@host forward % git branch --show-current
feat/export-forward-proxy
user@host forward % go test ./... -count=1
ok      github.com/coredns/coredns/plugin/forward       2.561s
```

chrisohaver (approved):
LGTM

chrisohaver:
> An example use case is for other plugins to proxy requests to dynamic upstream hosts.

Can you give an example of this example use case?

patdowney:
@chrisohaver are you looking for a code example or written?  I have a fairly niche use case where I need to dynamically choose a backend DoT service to forward `TXT` requests to depending on the name that was initially queried. The `Proxy` type was exactly what I was looking for to do the forwarding.

chrisohaver:
Take a look at the _view_ plugin.  It sounds like that could solve your problem without the need of a custom plugin.

chrisohaver:
e.g. the following (untested) config will forward TXT queries in `example.com` to 10.0.0.10, and everything else to 8.8.8.8 ...

```
example.com {
  view my-view {
    expr type == 'TXT'
  }
  forward . 10.0.0.10
}

. {
  forward . 8.8.8.8
}
```

patdowney:
Thanks @chrisohaver for the pointer, will take a look

PR #5957, opened 2023-03-10
Update README.md

lixin963:
than->from


chrisohaver (changes requested):
The current wording is clear and grammatically correct as is.  I don't think this change adds any clarity.

Issue #5958, opened 2023-03-10
plugin/k8s_external add support fallthrough


sstoner:
**What would you like to be added**:

I would like to add support for `fallthrough` in `plugin/k8s_external`
```yaml
k8s_external [ZONE...] {
    fallthrough [ZONES...]
}
```

**Why is this needed**:

This is needed in case the zone matches, but no record can be generated. In that case, the request can be passed to the next plugin.

sstoner:
### Possible issue reproduction:

#### CoreDNS Version: 1.10.1

#### The Corefile I used

```yaml
.:53 {
      k8s_external cls1.la in-addr.arpa ip6.arpa {
         apex test.ptr
      }
      kubernetes cluster.local in-addr.arpa ip6.arpa {
        pods insecure
        fallthrough in-addr.arpa ip6.arpa
      }
    }
```

We offer a service

```yaml
$ kubectl get svc demo
NAME   TYPE           CLUSTER-IP       EXTERNAL-IP            PORT(S)          AGE
demo   LoadBalancer   192.168.13.101   1.1.1.1                8080:30888/TCP   107s
```

**This is what we expected, and it is what we got.**

```yaml
# 172.16.21.72 is the coredns server serves that Corefile
[~/vanceli]# dig -t PTR 1.1.1.1.in-addr.arpa  @172.16.21.72

; <<>> DiG 9.11.4-P2-RedHat-9.11.4-26.P2.tl2.5 <<>> -t PTR 1.1.1.1.in-addr.arpa @172.16.21.72
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 10063
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;1.1.1.1.in-addr.arpa.          IN      PTR

;; ANSWER SECTION:
1.1.1.1.in-addr.arpa.   5       IN      PTR     demo.default.cls1.la.

;; Query time: 0 msec
;; SERVER: 172.16.21.72#53(172.16.21.72)
;; WHEN: 一 3月 13 18:40:44 CST 2023
;; MSG SIZE  rcvd: 103
```

**The PTR record of the cluster IP cannot be obtained unless we remove k8s_external from Corefile.**

[...]

PR #5959, opened 2023-03-13
plugin/k8s_extenral: Supports fallthrough option

sstoner:
To enable fallthrough for the`k8s_external` plugin, add the fallthrough option to the plugin configuration. This will allow DNS queries that are not handled by the plugin to be passed through to the next plugin in the Corefile. 
The fallthrough option can be useful if you want to use the`k8s_external` plugin for resolving external IP addresses, but fall back to a different plugin for other DNS queries.

#5958

The documentation for [k8s_external](https://coredns.io/plugins/k8s_external/) should include the new options added to the`k8s_external` plugin, along with their descriptions, syntax, and usage examples.

I have made some changes to the documentation, but it may still require further improvement. Please let me know if you have any suggestions so that I can make further revisions.

This pull request changes the behavior of `k8s_external` when a user misconfigures it. Previously, the plugin would silently move on to the next plugin, but now it will return an error message to users and terminate execution.

### DIFF
--- a/plugin/k8s_external/README.md
+++ b/plugin/k8s_external/README.md
@@ -10,8 +10,7 @@ This plugin allows an additional zone to resolve the external IP address(es) of 
 service and headless services. This plugin is only useful if the *kubernetes* plugin is also loaded.
 
 The plugin uses an external zone to resolve in-cluster IP addresses. It only handles queries for A,
-AAAA, SRV, and PTR records; all others result in NODATA responses. To make it a proper DNS zone, it handles
-SOA and NS queries for the apex of the zone.
+AAAA, SRV, and PTR records; To make it a proper DNS zone, it handles SOA and NS queries for the apex of the zone.
 
 By default the apex of the zone will look like the following (assuming the zone used is `example.org`):
 
@@ -67,6 +66,14 @@ k8s_external [ZONE...] {
 
 * if there is a headless service with external IPs set, external IPs will be resolved
 
+If the queried domain does not exist, you can fall through to next plugin by adding the `fallthrough` option.
+
+~~~
+k8s_external [ZONE...] {
+    fallthrough [ZONE...]
+}
+~~~
+
 ## Examples
 
 Enable names under `example.org` to be resolved to in-cluster DNS addresses.
@@ -105,6 +112,18 @@ zone transfers.  Notifies are not supported.
          k8s_external example.org
      }
  ~~~
+
+With the `fallthrough` option, if the queried domain does not exist, it will be passed to the next plugin that matches the zone.
+
+~~~
+. {
+   kubernetes cluster.local
+   k8s_external example.org {
+     fallthrough
+   }
+   forward . 8.8.8.8
+}
+~~~
 
 # See Also
 
--- a/plugin/k8s_external/external.go
+++ b/plugin/k8s_external/external.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/etcd/msg"
+	"github.com/coredns/coredns/plugin/pkg/fall"
 	"github.com/coredns/coredns/plugin/pkg/upstream"
 	"github.com/coredns/coredns/request"
 
@@ -39,6 +40,7 @@ type Externaler interface {
 type External struct {
 	Next  plugin.Handler
 	Zones []string
+	Fall  fall.F
 
 	hostmaster string
 	apex       string
@@ -93,6 +95,10 @@ func (e *External) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 	m.Authoritative = true
 
 	if len(svc) == 0 {
+		if e.Fall.Through(state.Name()) && rcode == dns.RcodeNameError {
+			return plugin.NextOrFailure(e.Name(), e.Next, ctx, w, r)
+		}
+
 		m.Rcode = rcode
 		m.Ns = []dns.RR{e.soa(state)}
 		w.WriteMsg(m)
--- a/plugin/k8s_external/external.go
+++ b/plugin/k8s_external/external.go
@@ -70,10 +70,6 @@ func (e *External) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 		return plugin.NextOrFailure(e.Name(), e.Next, ctx, w, r)
 	}
 
-	if e.externalFunc == nil {
-		return plugin.NextOrFailure(e.Name(), e.Next, ctx, w, r)
-	}
-
 	state.Zone = zone
 	for _, z := range e.Zones {
 		// TODO(miek): save this in the External struct.
--- a/plugin/k8s_external/setup.go
+++ b/plugin/k8s_external/setup.go
@@ -1,6 +1,7 @@
 package external
 
 import (
+	"errors"
 	"strconv"
 
 	"github.com/coredns/caddy"
@@ -9,7 +10,9 @@ import (
 	"github.com/coredns/coredns/plugin/pkg/upstream"
 )
 
-func init() { plugin.Register("k8s_external", setup) }
+const pluginName = "k8s_external"
+
+func init() { plugin.Register(pluginName, setup) }
 
 func setup(c *caddy.Controller) error {
 	e, err := parse(c)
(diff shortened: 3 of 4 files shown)


plugin/k8s_external/external.go
```diff
@@ -93,6 +102,10 @@ func (e *External) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 	m.Authoritative = true
 
 	if len(svc) == 0 {
+		if e.Fall.Through(state.Name()) {
```
chrisohaver: The behavior for `fallthough` we try to adhere to globally in CoreDNS is to only fall through when the domain queried does not exist, and not to fall through if the domain exists but hold no records of the queried type ...

```suggestion
		if e.Fall.Through(state.Name()) && rcode == dns.RcodeNameError {
```

plugin/k8s_external/external.go
```diff
@@ -68,8 +73,12 @@ func (e *External) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 		return plugin.NextOrFailure(e.Name(), e.Next, ctx, w, r)
 	}
 
-	if e.externalFunc == nil {
-		return plugin.NextOrFailure(e.Name(), e.Next, ctx, w, r)
+	if !e.kubernetesPluginLoaded {
+		if e.Fall.Through(state.Name()) {
+			return plugin.NextOrFailure(e.Name(), e.Next, ctx, w, r)
+		}
```
chrisohaver: I appreciate the added error here.  If the kubernetes plugin is not loaded, the configuration is essentially invalid, the user needs to know about it and fix it, but I think we should not fall through.

chrisohaver (changes requested):
thanks!

plugin/k8s_external/external.go
sstoner: Thank you for your feedback! It appears that this proposed change prioritizes the accuracy and reliability of DNS queries.

plugin/k8s_external/external.go
sstoner: > I appreciate the added error here. If the kubernetes plugin is not loaded, the configuration is essentially invalid, the user needs to know about it and fix it, but I think we should not fall through.

I also think so, but previously, if the Kubernetes plugin was not loaded, we defaulted to doing a fall through for the user. If we don't do a fall through now, we should return this error in advance and prevent the program from starting during setup phase instead of just logging it?

Here, I am complying with the previous agreement where 'fall through' is the default for users. However, it is now necessary to enable the `fallthrough` option.

chrisohaver: > we should return this error in advance and prevent the program from starting during setup phase 

If we could that would be ideal, but IIRC we cannot detect this during setup due to the plugin initialization order. We have to wait until all plugins are loaded, so we defer this check to onStartup.

> Here, I am complying with the previous agreement where 'fall through' is the default for users.
> However, it is now necessary to enable the fallthrough option.

If the configuration is bad (missing kubernetes), the default is to fall through to next plugin.  _I think this existing behavior is flawed_, it hides a broken configuration from the user, and would never result in a running-as-intended instance of CoreDNS. It is better to error when we know the configuration is wrong than to be silent and behave in a way that we know is not intended - e.g. sending queries upstream that could result in wrong answers sent to a client.

I appreciate and prefer your addition of adding an error here.  IMO, you are fixing a bug.  But again, I think the error should always be returned when the configuration is bad, even when the user is using `fallthrough`.

sstoner: Thank you for the explanation. I've also observed that enabling the `errors` plugin results in an extra error log being printed twice. Furthermore, I've made changes to the default behavior such that if Kubernetes isn't loaded, we'll return a `SERVFAIL` response instead of proceeding to the next plugin.

After reviewing the code and testing, I suggest we check for this issue during startup and failed start attempts. I will address this in a future pull request. Does that sound good to you?

chrisohaver: > After reviewing the code and testing, I suggest we check for this issue during startup and failed start attempts. I will address this in a future pull request. Does that sound good to you?

This isn't possible for the reason I explained above.

plugin/k8s_external/external.go
```diff
@@ -68,8 +73,11 @@ func (e *External) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 		return plugin.NextOrFailure(e.Name(), e.Next, ctx, w, r)
 	}
 
-	if e.externalFunc == nil {
-		return plugin.NextOrFailure(e.Name(), e.Next, ctx, w, r)
+	// TODO: If Kubernetes is not loaded, return an error directly on startup.
+	if !e.kubernetesPluginLoaded {
+		log.Error("the kubernetes plugin failed to load or misconfigured")
```
chrisohaver: We should not log an error here explicitly with a logger.  We should only return the servfail.

plugin/k8s_external/external.go
```diff
@@ -68,8 +73,11 @@ func (e *External) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Ms
 		return plugin.NextOrFailure(e.Name(), e.Next, ctx, w, r)
 	}
 
-	if e.externalFunc == nil {
-		return plugin.NextOrFailure(e.Name(), e.Next, ctx, w, r)
+	// TODO: If Kubernetes is not loaded, return an error directly on startup.
```
chrisohaver: This is not possible to do due to the order that plugins register.

plugin/k8s_external/README.md
```diff
@@ -106,6 +113,16 @@ zone transfers.  Notifies are not supported.
      }
  ~~~
 
+With the fallthrough option, if the query is not found, it will be passed to the next plugin.
+~~~
+. {
+   kubernetes cluster.local
+   k8s_external example.org {
+     fallthrough
+   }
```
chrisohaver, line 123: Should include a next plugin in this example (e.g. _forward_), otherwise the example only accomplishes turning NXDOMAIN responses into SERVFAIL responses.

plugin/k8s_external/setup.go
```diff
@@ -29,6 +34,15 @@ func setup(c *caddy.Controller) error {
 			e.externalServicesFunc = x.ExternalServices
 			e.externalSerialFunc = x.ExternalSerial
 		}
+
+		if e.externalFunc == nil ||
+			e.externalAddrFunc == nil ||
+			e.externalSerialFunc == nil ||
+			e.externalServicesFunc == nil {
+			return nil
```
chrisohaver: I think we could log a warning/error here... 

I don't know if it's appropriate to panic and shutdown though at this point.  I don't think it is.

chrisohaver: ... actually this code is never reached unless there is a plugin named `kubernetes` that does not implement `Externaler`... which basically would never happen in any reasonable scenario. 

Practically speaking this check is unnecessary, and can be safely be removed.

chrisohaver: A warning could be logged in the `m == nil` check above.

plugin/k8s_external/setup.go
```diff
@@ -29,6 +34,15 @@ func setup(c *caddy.Controller) error {
...
+		}
+
+		e.kubernetesPluginLoaded = true
```
chrisohaver: should move this into the existing `if x, ok := m.(Externaler); ok {` block.

plugin/k8s_external/setup.go
sstoner: > I don't know if it's appropriate to panic and shutdown though at this point. I don't think it is.

I think we should end the process promptly to avoid users being unaware of their misconfiguration. If the Kubernetes plugin is not configured or Externalizer is not implemented, we can initiate a failure by returning an error.

sstoner:
I simplified the process in `ServeDNS` for`k8s_external` by removing the detection of whether the `kubernetes` plugin is loaded or not, and instead moved it to startup. This ensures that if there is a configuration error causing `kubernetes` to not load, we will fail to start and output an error message to the user.

plugin/k8s_external/README.md
```diff
@@ -106,6 +113,17 @@ zone transfers.  Notifies are not supported.
      }
  ~~~
 
+With the fallthrough option, if the query is not found, it will be passed to the next plugin that matches the zone.
```
chrisohaver:
```suggestion
With the `fallthrough` option, if the queried domain does not exist, it will be passed to the next plugin that matches the zone.
```

plugin/k8s_external/README.md
```diff
@@ -67,6 +66,14 @@ k8s_external [ZONE...] {
 
 * if there is a headless service with external IPs set, external IPs will be resolved
 
+if zone matches and there are not any external IPs, you can fallthrough to next plugin by adding `fallthrough` option.
```
chrisohaver:
```suggestion
If the queried domain does not exist, you can fall through to next plugin by adding the `fallthrough` option.
```

chrisohaver (commented):
LGTM Thanks!  Ive suggested some minor tweaks in README language to be more correct about NODATA vs NXDOMAIN behavior.

sstoner:
> LGTM Thanks! Ive suggested some minor tweaks in README language to be more correct about NODATA vs NXDOMAIN behavior.

Thank you for your feedback and suggestions. I appreciate the attention to detail and effort you put into helping ensure the accuracy of the PR.

plugin/k8s_external/README.md
sstoner: Fixed

chrisohaver (approved):

PR #5969, opened 2023-03-15
Run coredns as non root.

vinayakankugoyal:
To run coredns as nonroot using the nonroot distroless base image.

No

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM debian:stable-slim
+FROM --platform=$BUILDPLATFORM debian:stable-slim AS build
 SHELL [ "/bin/sh", "-ec" ]
 
 RUN export DEBCONF_NONINTERACTIVE_SEEN=true \
@@ -7,13 +7,14 @@ RUN export DEBCONF_NONINTERACTIVE_SEEN=true \
            TERM=linux ; \
     apt-get -qq update ; \
     apt-get -yyqq upgrade ; \
-    apt-get -yyqq install ca-certificates ; \
+    apt-get -yyqq install ca-certificates libcap2-bin; \
     apt-get clean
+COPY coredns /coredns
+RUN setcap cap_net_bind_service=+ep /coredns
 
-FROM --platform=$TARGETPLATFORM scratch
-
-COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
-ADD coredns /coredns
-
+FROM --platform=$TARGETPLATFORM gcr.io/distroless/static-debian11:nonroot
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY --from=build /coredns /coredns
+USER nonroot:nonroot
 EXPOSE 53 53/udp
 ENTRYPOINT ["/coredns"]


chrisohaver:
(closed and reopened to re-trigger the failed circleci test)

chrisohaver (approved):

Dockerfile
```diff
@@ -10,10 +10,13 @@ RUN export DEBCONF_NONINTERACTIVE_SEEN=true \
     apt-get -yyqq install ca-certificates ; \
     apt-get clean
 
-FROM --platform=$TARGETPLATFORM scratch
-
-COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
-ADD coredns /coredns
+FROM --platform=$TARGETPLATFORM k8s.gcr.io/build-image/setcap:bullseye-v1.4.2 AS setcap
```
SuperQ: Could we combine this with the certs `FROM` and have a single `build` layer? I don't see the need for two steps here.

SuperQ (changes requested):

Dockerfile
```diff
@@ -10,10 +10,13 @@ RUN export DEBCONF_NONINTERACTIVE_SEEN=true \
     apt-get -yyqq install ca-certificates ; \
```
SuperQ: We could just add `setcap` to this install list, rather than pulling in a whole second Debian layer into the image.

Dockerfile
vinayakankugoyal: yup will do

Dockerfile
vinayakankugoyal: good call, lets do that.

SuperQ (approved):
LGTM

SuperQ:
Weird, CircleCI is having issues.

chrisohaver:
> /retest

That doesn't work in this repo.  The Circle CI failures are some kind of Circle CI infrastructure issue.

FYI, the error doesnt go away after a successful run... but in this case, we can see that he kubernetes ci tests are not present, so Circle CI failed to start them.

Please amend your commit and force push - that should re-run the tests.
Be aware, If the kubernetes ci tests do get run successfully - I think the old Circle CI failure will remain (I have observed this in some other recent PRs).

chrisohaver:
closing and re-opening to re-run circle ci tests.

dmotte:
Guys, just to let you know: after upgrading, my CoreDNS Docker container broke because of this change.

The command I was using to run the container was: `docker run ... -v $PWD/Corefile:/Corefile ... docker.io/coredns/coredns:latest`, but after this change the **working directory** of the container has become `/home/nonroot` instead of `/`, so CoreDNS could not find the `Corefile` anymore and was falling back to the default `whoami` configuration **silently**.

Adding `-conf /Corefile` as CLI option solved my problem, but maybe it's worth to document this **breaking change** somewhere. Moreover, I know that the config file path should always be specified as a best practice, but IMHO falling back to the default config silently may not be the best option.

Not complaining tho 🙂 Just posting my experience here in the hope that it may be useful to someone

polarathene:
> Adding `-conf /Corefile` as CLI option solved my problem

You could also use `--workdir` (`working_dir` in `compose.yaml`).

Although I agree with you, that seems like an oversight for a minor version bump.

chrisohaver:
This change is listed in the release notes. https://github.com/coredns/coredns/blob/master/notes/coredns-1.11.0.md

polarathene:
> This change is listed in the release notes.

Yes, but it doesn't mention anything about the implicit `WORKDIR` change which was probably unintentional?

---

Likewise there was no discussion here about the breakage it introduced for those who already were using unprivileged ports and explicitly dropping the capability that is now mandatory for the binary to run regardless even if you don't bind to privileged ports.

According to the capabilities man page, the `setcap` approach is a "capability-dumb binary" which causes this problem (_failing if the expected capability is not available_). I am not a Go dev, but from what the man page says, it should be possible for CoreDNS to be smarter and run without `setcap` by using the API call to request the capability if binding to privileged ports? Then everyone should be happy and it's not just container specific? 🤷‍♂️ 

For example: https://github.com/ik5/gocapng

chrisohaver:
I think we should consider rolling this change back, since it's causing problems for people.

polarathene:
> since it's causing problems for people

Yes, please be consistent with the binary 🙏 

If you want to address the non-root capability concern, do so properly as my last comment suggests:

> _I am not a Go dev, but from what the man page says, it should be possible for CoreDNS to be smarter and run without `setcap` by using the API call to request the capability if binding to privileged ports? Then everyone should be happy and it's not just container specific?_ 🤷‍♂️
> 
> _For example:_ https://github.com/ik5/gocapng

Please revert the `WORKDIR` as well, and communicate that clearly in release notes if possible as it was not previously.

---

I looked into this and documented justification extensively at the time via the PR: https://github.com/coredns/coredns/pull/6320

I know it is very verbose, but I am confident it's the better approach.

The only blocker was the [kubernetes test failure](https://github.com/coredns/coredns/pull/6320#issuecomment-1721209536), which I cannot afford to investigate. If you're able to implement the Go API instead, that would presumably fix it and benefit anyone wanting the same feature by default without the Docker image.

dgl:
> If you want to address the non-root capability concern, do so properly as my last comment suggests:
[...]
> For example: https://github.com/ik5/gocapng

That "proper" way clearly states the need to have setcap on the binary: https://github.com/ik5/gocapng?tab=readme-ov-file#building-requirements

That's why dropping setcap is likely breaking Kubernetes. (After waiting for 5 minutes CircleCI isn't loading the log, so I can't be sure...)

The correct fix _using capabilities_ is to use ambient capabilities, but those aren't currently supported in Kubernetes: https://github.com/kubernetes/kubernetes/issues/56374 (and likely won't be in the short term)...

Probably the better fix is to do similar to Docker does as you point out, and set `net.ipv4.ip_local_port_range`, however that isn't supported on some older kernels, particularly the one that RHEL 7 ships (which Kubernetes still supports -- for now: https://github.com/kubernetes/kubernetes/issues/116799).

polarathene:
Apologies for the verbosity in response below, and my lack of k8s knowledge which probably shows below 😅 

**TL;DR:** Better to switch back to `scratch` base image with root user as the default?
- Non-root is still possible without `setcap`, just requires communicating to users how to configure for that explicitly (eg: `net.ipv4.ip_unprivileged_port_start=0`), should they require a privileged port internally. That is a better security policy and UX in my opinion.
- The kernel checks from `setcap` also apply to running as root, which outputs an unhelpful error to those using root but dropping all capabilities by default.
- [`setcap` ironically also impacts users with non-root](https://github.com/coredns/coredns/issues/6249#issuecomment-1691209105), that have a more strict security policy and want to drop the capability as they're satisfied with the internal container port binding to an unprivileged port, but due to the kernel check enforcing the capability requirement (even when it's not needed), this is not possible to support.

---

<details>
<summary>Original response (click to expand)</summary>

> That "proper" way clearly states the need to have `setcap` on the binary

Thanks for clarifying that. I had a misunderstanding and for non-root you're right, but if switching back to how CoreDNS was distributed as an image before (which was fine IMO), the API can be used and `setcap` won't be necessary.

I don't think the minor convenience should be favored from using it for non-root. Instead that should be clearly documented for the security conscious that insist on running it as non-root.

I am of the impression that they're simply interested in doing it for the sake of it because they don't know any better and are overly cautious/paranoid, plus it's simpler for them than to be explicit about the security context needed. If they care enough about it, better to inform correct configuration via documentation instead.

---

https://man7.org/linux/man-pages/man7/capabilities.7.html

> ### Safety checking for **capability-dumb binaries**
> A capability-dumb binary is an application that has been marked
 to have file capabilities, but has **not** been converted to **use the
 [`libcap`(3)](https://man7.org/linux/man-pages/man3/libcap.3.html) API** to manipulate its capabilities.
> 
> (In other words,
 this is a traditional set-user-ID-root program that has been
 switched to use file capabilities, but whose **code has not been
 modified to understand capabilities**.)
> 
> For such applications, the effective capability bit is set on the file, so that the file permitted capabilities are automatically enabled in the process effective set when executing the file.
> The **kernel recognizes** a file which has the effective capability bit set as capability-dumb for the **purpose of the check** described here.

Relying on `setcap` for file permitted capabilities to enforce a kernel check on granting the required capabilities when running the process unprivileged. This is only relevant because the process is being run without root.

The API in that context cannot manage the capabilities in the same way at runtime, unless a parent process with privileges is running it and then manages privileges for the child process.

Services like Postfix commonly do this, but if you insist on running the container without root user, you're stuck with the capability-dumb approach via file attributes. 

--

> If the process did not obtain the full set of file permitted capabilities, then [`execve`(2)](https://man7.org/linux/man-pages/man2/execve.2.html) **fails with the error `EPERM`**.
> This prevents possible security risks that could arise when a capability-dumb application is executed with **less privilege than it needs**.
> Note that, by definition, the **application could not itself recognize this problem**, since it does not employ the
 [`libcap`(3)](https://man7.org/linux/man-pages/man3/libcap.3.html) API.
 
This error is not helpful as shown in #6320 you get: `exec /coredns: operation not permitted`, as opposed to if the process avoided the check and encountered the actual error related to lack of the required capability: `Listen: listen tcp :53: bind: permission denied`.

Prior to the introduction of `setcap`, distroless was not used, the final image was `scratch` with CoreDNS binary copied over (_along with the `ca-certificates`_). The risk of running the binary as root within the container was minimal.

Outside of containerization, you'd often run CoreDNS as root yes? It's not distributed with the same capability-dumb approach outside of the Docker image and worked fine prior to introducing that into the image. The only reason it was merged was a vague request to support running as non-root.

You could run as non-root and bind to an unprivileged port, no issues. It doesn't change the fact that the privileged port is being run from the host as root? If running a Docker container via a rootless daemon, then you could have root in the container and bind to `:53` internally, but on the actual host you'd have to bind to an unprivileged port. No specific use-case was provided in where it was important to support `:53` via a non-root user within the container alone.

A container exploit does not require root to break out, there's even less attack surface with `scratch`. If a user wants to run as non-root user all they need to do is set the UID+GID, while either way both can drop all capabilities by default and only enable those which are required.

For non-root user, the only reason the capability is required is to bind a port lower than 1024. This isn't an issue on Docker by default as I've illustrated in #6302 while elsewhere you should be capable of the same approach, setting the `sysctl` to lower the port binding. You're permitting the process to bind to any port as non-root regardless, so I'd be interested in what actual attack vector you're trying to protect against (_even a past vulnerability example that could be relatable_), otherwise it's security paranoid parroting 🤷‍♂️ 

You can apply the `sysctl` in k8s as well via security context, as you would the other related settings. It's far better to have that explicit config than rely on implicit config that is subject to change (_perfect example with CoreDNS image already as changes have been introduced that were breaking for existing users without mention in the changelog_).

---

> That's why dropping setcap is likely breaking Kubernetes.

Yes, understood. It's because it's running unprivileged and relies on file attribute granted from root to permit the capability. Personally, I don't like how that also affects the binary when run as root, since being capability-dumb still incurs the early exit check with the unhelpful error.

> The correct fix using capabilities is to use ambient capabilities, but those aren't currently supported in Kubernetes

Not available in Docker either AFAIK.

However both allow you to control the capabilities from the context of the container as a whole. Since only one process is run this seems adequate. Those that are security conscious and really care about the additional security practices should be comfortable with the explicit approach as described earlier.

As root, CoreDNS could drop all capabilities by default and only keep those that are necessary. Although I'm not sure how well that'd work due to the plugin system, they'd likewise have to express the capabilities they need. Alternatively, the capabilities can be checked, along with if running privileged or unprivileged, if CoreDNS should better communicate a failure and advise a user (eg: link to documentation, or suggest using `setcap`).

---

> however that isn't supported on some older kernels, particularly the one that RHEL 7 ships (which Kubernetes still supports

That's almost a decade old release series. Is there known deployments of CoreDNS with k8s on RHEL 7 currently?

Is that really worth supporting, or just an unknown? [RHEL 7 is presently in maintenance](https://access.redhat.com/support/policy/updates/errata#Life_Cycle_Dates) and soon to switch to extended support, but I imagine those unable to migrate to newer version of RHEL aren't as critical about security as they are other factors?

They are likely able to afford the expertise to maintain their own customizations to support their needs, or are comfortable remaining on an older CoreDNS release (_that problem will affect them elsewhere in that same sense, so it's not adding any new friction_)

A 3.x kernel is very old for running a containerized environment on. There's no cgroups v2 support, which is important for running Docker for some time IIRC, and I wouldn't be surprised if that applies to k8s. I really don't see RHEL 7 as a valid blocker? :shrug:

</details>

---

**UPDATE:** Actually it looks like the main problem is just setting `e` (effective) on the binary with `setcap`. I think everyone can be happy if `setcap` just sets `p` (permitted) capability and uses API at runtime to raise the capability into the effective set (_allowed since it's marked as permitted_).

That would prevent the binary failing from a kernel check on effective capabilities being set too early (via `setcap`). The associated sysctl could additionally be checked to verify if it's actually necessary to add `CAP_NET_BIND_SERVICE` into the effective set, in addition to running unprivileged 🤔

PR #5990, opened 2023-03-24
plugin/forward: fails to forward requests when the fail counter overflows

sstoner:

If a proxy repeatedly fails, it will be removed from the forward list. However, if the fail counter (uint32) overflows, the fail counter of proxy will be reset to an overflowed value of `0`. 
For example: https://go.dev/play/p/aDvMM3TeJHf?v=goprev

This can cause the proxy to be added back to the forward list and may result in some recursive requests failing during the duration of `duration` multiplied by `max_fails`.

No

No

No, it does not. However, it enhances the stability and reliability of the system.

BTW, I suggest introducing a success counter for the `health_check` probe to determine if a failed proxy is ready after multiple successful probes, instead of relying on just one successful probe.

### DIFF
--- a/plugin/pkg/proxy/health.go
+++ b/plugin/pkg/proxy/health.go
@@ -105,7 +105,7 @@ func (h *dnsHc) Check(p *Proxy) error {
 	err := h.send(p.addr)
 	if err != nil {
 		HealthcheckFailureCount.WithLabelValues(p.addr).Add(1)
-		atomic.AddUint32(&p.fails, 1)
+		p.incrementFails()
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


Tantalor93:
thanks for opening the PR @sstoner, imho it makes sense. Can you resolve the conflicts, otherwise LGTM

sstoner:
> thanks for opening the PR @sstoner, imho it makes sense. Can you resolve the conflicts, otherwise LGTM

Thanks, on it

Tantalor93 (approved):
I'll wait few days if any other maintainer want to review this also and then I'll merge this

chrisohaver (approved):

Issue #5998, opened 2023-03-30
CoreDNS 1.10.1 Return SERVFAIL Error with "dns: overflow unpacking uint16" when Upstream DNS response is more than 512 bytes


robbiezhang:
**What happened**:
Use CoreDNS 1.10.1 to query DNS `cluster-us-east-1-152269.prevops.com` using AzureDNS as upstream DNS server in an AKS cluster. It returns SERVFAIL after ~10 seconds
```
nslookup cluster-us-east-1-152269.prevops.com.
Server:         10.0.0.10
Address:        10.0.0.10#53

** server can't find cluster-us-east-1-152269.prevops.com: SERVFAIL
```

**What you expected to happen**:
It should return a list of IPs

**How to reproduce it (as minimally and precisely as possible)**:
Create an AKS cluster
Change the coredns version to 1.10.1.
Deploy a dnsutils pod, and run `nslookup cluster-us-east-1-152269.prevops.com.`

**Anything else we need to know?**:
The issue doesn't repro in 1.10.0, 1.9.4, 1.9.3 

**Environment**:
AKS, AzureDNS

- the version of CoreDNS: 1.10.1
- Corefile:
```
    .:53 {
        errors
        ready
        health {
          lameduck 5s
        }
        kubernetes cluster.local in-addr.arpa ip6.arpa {
          pods insecure
          fallthrough in-addr.arpa ip6.arpa
          ttl 30
        }
        prometheus :9153
        forward . /etc/resolv.conf
        cache 30
        loop
        reload
        loadbalance
    }
```
- logs, if applicable:
1.10.1 (Failed)
```
[INFO] 10.244.1.4:53490 - 19745 "A IN cluster-us-east-1-152269.prevops.com. udp 54 false 512" - - 0 5.001039164s
[ERROR] plugin/errors: 2 cluster-us-east-1-152269.prevops.com. A: dns: overflow unpacking uint16
[INFO] 10.244.1.4:53490 - 19745 "A IN cluster-us-east-1-152269.prevops.com. udp 54 false 512" - - 0 5.001601765s
[ERROR] plugin/errors: 2 cluster-us-east-1-152269.prevops.com. A: dns: overflow unpacking uint16
```
1.10.0 (Succeeded)
```
[INFO] 10.244.1.4:39792 - 57873 "A IN cluster-us-east-1-152269.prevops.com. udp 54 false 512" NOERROR qr,aa,tc,rd,ra 2134 0.0001473s
[...]

PR #6003, opened 2023-03-31
Handle UDP responses that overflow with tc bit

### LINKED ISSUE #5953: CoreDNS resiliency with non-compliant upstream
**What happened**:
I'm trying to work through an issue in which a OKD user reported (https://issues.redhat.com/browse/OCPBUGS-6829) CoreDNS failing on upstream/forwarded requests with the failure:
```
CoreDNS logs: [ERROR] plugin/errors: 2 oauth-openshift.apps.okd-admin.muc.lv1871.de. AAAA: dns: overflowing header size 
```

The only way this logic could be getting executed is if the upstream is non-compliant regarding UDP payloads sizes. We specify `bufsize: 512` and it appears the upstream is coming back with more than 512 and not setting the truncated bit. The user is using a Windows 2016 or 2019 AD DC integrated DNS server for upstream DNS.

I've reproduced this error with 2 CoreDNS binaries: one being a vanilla CoreDNS (first hop/downstream) and another being a hacked CoreDNS (upstream) that doesn't truncate and sends UDP responses larger than UDP payload requested. I "hacked" CoreDNS to never truncate by putting a `return` [here](https://github.com/openshift/coredns/blob/9aaa7e0a86b69bafb9f544a0e5cb1873535a8f6b/vendor/github.com/miekg/dns/msg_truncate.go#L50). So it looks like:
_Dig Client --> Vanilla Downstream CoreDNS 1.10.0 --> Upstream Hacked-to-not-truncate CoreDNS 1.10.0_

In this reproducer, the chain of events is as follows:
1. Client requests a larger-than-512-byte DNS Record
2. The downstream CoreDNS forwards to the upstream CoreDNS, but the upstream hacked-CoreDNS doesn't truncate it's response
3. Downstream CoreDNS repeatedly fails and tries sending more requests to upstream (in UDP) for 5 seconds straight. The error is always one of the "overflow" errors I mentioned above.
4. The request times out after 5 seconds ([hard coded](https://github.com/coredns/coredns/blob/b7279d1f668ad2fe401c254543f9dcbc878b0378/plugin/forward/forward.go#L247)) and the downstream returns the "overflow" message that it had been receiving for 5 seconds of non-stop querying

I understand the upstream is non-compliant, but it seems non-compliant server exist in the wild (see https://github.com/coredns/coredns/issues/3941 and https://github.com/coredns/coredns/issues/3305). The user's solution here is to increase `bufsize` (seems a bit arbitrary) or turn on `force_tcp`.

_My question here is:_ Could CoreDNS retry TCP after a UDP Payload failure automatically? It helplessly retries for 5 seconds, but would it increase resiliency to attempt a TCP retry upon one of these failure cases?

I talked about creating the reproducer in the **What happened** section.

N/A

**Environment**:

- the version of CoreDNS:
1.10.0
- Corefile:
```
  Corefile: |
    .:5353 {
        bufsize 512
        errors
        log . {
            class error
        }
        health {
            lameduck 20s
        }
        ready
            pods insecure
        }
        prometheus 127.0.0.1:9153
        forward . /etc/resolv.conf {
            policy sequential
        }
        cache 900 {
            denial 9984 30
        }
        reload
    }
```
- logs, if applicable:
```
[ERROR] plugin/errors: 2 www.example.org. AAAA: dns: overflowing header size
```
- RHEL 8.7
- Others:

phealy:
Prior to #5671 being merged, the cache plugin did two things that it doesn't do now:

1. Force the UDP buffer size to 2048 bytes
1. Always set the DNSSEC DO bit, which also meant that an eDNS0 OPT RR was always included with the appropriate buffer size

This caused problems for clients when the DNS server would ignore the bufsize in the OPT RR and return a UDP response >512 octets; it would fail to unpack because it didn't fit in the buffer, and also other confusion as enumerated in the PR.

However, the pre-5671 behavior would also prevent issues with noncompliant upstream DNS servers that returned UDP responses >512 octets even when the client didn't supply an eDNS0 header, because every request sent eDNS0. The too-large-for-the-client response would get put into the cache, and coredns would properly send a partial response with the truncate bit set, letting the client know it needed to retry over TCP. The retry would then be served from the cache. This only worked if cache was enabled.

This PR prevents request failure in the case of an upstream that sends an overly large UDP response - when handling the error from miekg/dns attempting to unpack the too-large message, if the error indicates an overflow we return an empty response with the truncate bit set. This behavior matches what large providers are doing (Cloudflare on 1.1.1.1, Google DNS on 8.8.8.8) and indicates to the client that it needs to try over TCP.

While this does have the downside of not getting the UDP response into cache, that's a more complicated change - coredns would have to use a large enough buffer to get the message successfully unpacked, with no idea from the client side what the maximum that could be sent by the server would be. This would lead to higher memory utilization on each query, and would be useless if cache wasn't enabled or the client wasn't able to retry via TCP (see busybox's nslookup).

I proposed two alternate solutions in #5953, and this would be a required first step for either of them:
1. Allow the configuration to force a larger buffer size - in order to avoid the problem that led to #5671 in the first place, we still need to handle overlarge responses and set the truncate bit.
2. Allow the configuration to force eDNS0 on all upstream requests - if the server sends an overly large response even with eDNS0 turned on, we still need to truncate appropriately instead of failing to respond to the client.

#5953
#5998

I didn't see any.

No.

### DIFF
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -124,7 +124,12 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 			// client retry over TCP (if that's supported) or at least receive a clean
 			// error. The connection is still good so we break before the close.
 			if proto == "udp" && strings.Contains(err.Error(), "overflow") {
-				ret.Truncated = true
+				newRet := state.Req.Copy()
+				newRet.AuthenticatedData = false
+				newRet.RecursionAvailable = ret.RecursionAvailable
+				newRet.Response = true
+				newRet.Truncated = true
+				ret = newRet
 				break
 			}
 			pc.c.Close() // not giving it back
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"io"
 	"strconv"
+	"strings"
 	"sync/atomic"
 	"time"
 
@@ -117,6 +118,15 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 	for {
 		ret, err = pc.c.ReadMsg()
 		if err != nil {
+			// If the error is an overflow, we probably have a misbehaving upstream
+			// that is sending out-of-spec (>512 byte) UDP responses without an eDNS0
+			// OPT RR. Set the truncate bit and return the result - this will make the
+			// client retry over TCP (if that's supported) or at least receive a clean
+			// error. The connection is still good so we break before the close.
+			if proto == "udp" && strings.Contains(err.Error(), "overflow") {
+				ret.Truncated = true
+				break
+			}
 			pc.c.Close() // not giving it back
 			if err == io.EOF && cached {
 				return nil, ErrCachedClosed


plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +118,15 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 	for {
 		ret, err = pc.c.ReadMsg()
 		if err != nil {
+			// If the error is an overflow, we probably have a misbehaving upstream
+			// that is sending out-of-spec (>512 byte) UDP responses without an eDNS0
+			// OPT RR. Set the truncate bit and return the result - this will make the
+			// client retry over TCP (if that's supported) or at least receive a clean
+			// error. The connection is still good so we break before the close.
+			if proto == "udp" && strings.Contains(err.Error(), "overflow") {
```
chrisohaver, line 126: can we more accurately identify the error by type rather than string compares?

chrisohaver (changes requested):
My initial preference was to have forward itself do the retry over TCP. However that solution is only very subtly better.  It results in the cache for the requested record being filled very slightly earlier than in this proposed method (one round trip to client earlier). This would only benefit for two concurrent queries of an record that is not yet in cache.

Since that is such a small differcence, I am OK with this approach.  This approach also has forward doing "less magic behind the curtains" for the client.

Can you add some tests that verify this in conjunction with cache? Cross plugin "integration" tests go in `/tests`. Adding tests is significantly more work than the change here so far, but I think it important to have them, especially since this was more or less a regression.

plugin/pkg/proxy/connect.go
phealy: I'd love to, but the [upstream message helpers that return the overflow errors](https://github.com/miekg/dns/blob/d83d31bedae5ce562b4686f7fbd70d4cba94424d/msg_helpers.go) aren't using any of [the exported defined errors from the package](https://github.com/miekg/dns/blob/master/msg.go#L51-L71).

phealy:
> Can you add some tests that verify this in conjunction with cache? Cross plugin "integration" tests go in `/tests`.

Happy to - I just wanted to get validation that you were OK with the change itself before I spent time writing all the tests for it.

phealy:
> My initial preference was to have forward itself do the retry over TCP. However that solution is only very subtly better. It results in the cache for the requested record being filled very slightly earlier than in this proposed method (one round trip to client earlier). This would only benefit for two concurrent queries of an record that is not yet in cache.

I do think this is a good benefit, but I didn't want to do it immediately for a few reasons:
1. It adds delay to the initial response when the client may never come back for it - especially important for something like busybox or musl-based apps (though that's finally getting fixed in 1.2.4) that *can't* come back via TCP.
2. It's useless if cache isn't enabled (we could check for that before retrying, of course).
3. It's a more complicated change, and I don't know cache and the plugin stack well enough yet to make this change quickly and fix the regression.
4. It adds magic, as you said.

I did wonder about this, as a possible solution - could we use something akin to the [prefetch capability in the cache plugin](https://coredns.io/plugins/cache/#:~:text=are%20never%20cached.-,prefetch,.,-serve_stale) to get the best of both worlds? For example:
1.  client hits us up with UDP query, no eDNS0, response is >512 bytes
2. we return the UDP response immediately with tc set
3. if cache is enabled, we do a background query to fill the cache in the anticipation of the client coming back via TCP.

That way we get the cache warming if desired, we don't query when the cache isn't there to take the result, and it can be made configurable by the end user. I don't know if this would be best integrated in cache or as a separate plugin that could inspect the request at the tail end. Thoughts?

chrisohaver:
> I do think this is a good benefit, but I didn't want to do it immediately for a few reasons:

I am (still) OK with the currently proposed PR, with tests added.

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +118,15 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 	for {
 		ret, err = pc.c.ReadMsg()
 		if err != nil {
+			// If the error is an overflow, we probably have a misbehaving upstream
+			// that is sending out-of-spec (>512 byte) UDP responses without an eDNS0
+			// OPT RR. Set the truncate bit and return the result - this will make the
```
gcs278, line 123: Thanks for the PR! This sounds like a simpler solution than having CoreDNS retry TCP itself.

My question is, will this handle the situation where an upstream sets the truncated bit, but actually doesn't fully truncate the result? In my case, the upstream does appear to set the TC bit and partially truncate the response, but it doesn't apply any truncation on the Authority RRs, which pushes it over the bufsize requested (and CoreDNS overflows).

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +118,15 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+			// client retry over TCP (if that's supported) or at least receive a clean
+			// error. The connection is still good so we break before the close.
+			if proto == "udp" && strings.Contains(err.Error(), "overflow") {
+				ret.Truncated = true
+				break
```
chrisohaver, line 133: Only should break here if response message id matches the request's message id.

plugin/pkg/proxy/connect.go
phealy: It should, but I think it would actually be cleaner to build a new response object that we guarantee is in a clean state. It'll be in the next commit.

I changed to that method based on the documentation:

> ReadMsg reads a message from the connection co. If the received message contains a TSIG record the transaction signature is verified. This method always tries to return the message, **however if an error is returned there are no guarantees that the returned message is a valid representation of the packet read.**

plugin/pkg/proxy/connect.go
phealy: Fixed, thanks.

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +118,20 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
 		ret, err = pc.c.ReadMsg()
 		if err != nil {
+			// If the error is an overflow, we probably have a misbehaving upstream
+			// that is sending out-of-spec (>512 byte) UDP responses without an eDNS0
+			// OPT RR. Set the truncate bit and return the result - this will make the
+			// client retry over TCP (if that's supported) or at least receive a clean
+			// error. The connection is still good so we break before the close.
+			if proto == "udp" && strings.Contains(err.Error(), "overflow") {
+				newRet := state.Req.Copy()
+				newRet.AuthenticatedData = false
```
chrisohaver, line 128: Please add a note here for why AD bit is cleared.  e.g. something like ...

Clear AD bit in case request had set the AD bit.  The empty response is not authenticated.

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +118,20 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+				newRet.RecursionAvailable = ret.RecursionAvailable
```
chrisohaver, line 129: This probably also warrants an explanatory code comment.

Should AA be handled the same way?

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +118,20 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 	for {
 		ret, err = pc.c.ReadMsg()
 		if err != nil {
+			// If the error is an overflow, we probably have a misbehaving upstream
+			// that is sending out-of-spec (>512 byte) UDP responses without an eDNS0
+			// OPT RR. Set the truncate bit and return the result - this will make the
```
chrisohaver, line 123:
```suggestion
			// If the error is an overflow, we probably have an upstream misbehaving in some way
			// (e.g. sending >512 byte UDP responses without an eDNS0 OPT RR).
			// Instead of returning an error, return an empty response with TC bit set. This will make the
```

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +118,20 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+			// client retry over TCP (if that's supported) or at least receive a clean
+			// error. The connection is still good so we break before the close.
+			if proto == "udp" && strings.Contains(err.Error(), "overflow") {
```
gcs278, line 126: So I recreated a scenario in which upstream sets the truncated flag, but doesn't truncate the response. I get a `buffer size too small` error instead of overflow. Would we want to support retrying for this scenario?

chrisohaver: what's the buffer size set to?

chrisohaver: ... I see. I think its just awkward reverse wording...  IMO should have been "message is too large for buffer". or "message exceeds buffer size" ("buffer size too small" had me thinking buffer was set below the minimum allowed value)

This is a public error var in the dns package.  Yes, it should be included.

```
	ErrBuf           error = &Error{err: "buffer size too small"}          // ErrBuf indicates that the buffer used is too small for the message.

```

plugin/pkg/proxy/connect.go
gcs278: Is it worth pursuing an update in https://github.com/miekg/dns to type-out these errors? It requires waiting for a miekg/dns release and re-vendoring, so I'm assuming that would take longer.

gcs278:
@phealy Just a friendly bump - are you planning to pick this PR back up? I'm very interested in this fix.

gcs278:
@phealy not sure if you got caught up in other priorities, but could I help you finish out this PR? I don't want to step on any toes, but I do have some cycles to get this update through if you want me to take a stab at the review comments and testing. Let me know your estimated timeline to getting back to this or if I can help out here.

Issue #5527, opened 2022-07-19
plugin/rewrite: Support CNAME target record answer rewrite


ryaneorth:
**What would you like to be added**:
Add a new `cname` option for the FIELD parameter in the rewrite plugin

**Why is this needed**:
The use-case this new field option would meet can found in [this question](https://github.com/coredns/coredns/issues/5513).

chrisohaver:
The description needs to fully describe what the requested option does.

chrisohaver:
Actually - while I was typing up an example description for you, I realize this request is more complicated than it might initially seem.

**Effectively, it must mutate the CNAME targets, _and also_ the attached records for those targets (both name and IPs). And to do that it needs to perform lookups for them ... which is something _rewrite_ doesn't do.**

I'm not sure where this kind of function should be implemented.  Maybe extending _rewrite_ to do this sort of thing is the way to go - maybe another plugin is a better approach.

chrisohaver:
I'm thinking this should be OK to include in _rewrite_, as opposed to a new plugin.  It is unique compared to other rewrite operations, but I think it still fits.  The lookup should be easy to do with the `upstream` package.

@greenpau, @johnbelamaric, WDYT?

chrisohaver:
This kind of dips _another_ toe into growing _rewrite_ into a fully fledged _answer_ rewriter (with the ability to ad-hoc rewrite any fields of records in an answer).  When deciding the syntax for this function, we should keep in mind possible future syntax for that kind of ad-hoc answer rewrite function.

Amila-Rukshan:
Hi @chrisohaver 
I came across this issue when looking for an issue to contribute to in this repo. I understand the problem in the linked original issue but could not get my head around the suggested solution. Are you able to provide some corefile reflecting the implemented feature (using cname FIELD in rewrite) and input query and output response to better understand it?

Thanks for your time!

chrisohaver:
Thanks for considering contributing!

A query for name `abc.example.com`, type `A`

The response without rewrite...
```
abc.example.com IN CNAME def.example.com
def.example.com IN A 1.2.3.4
```

The response with CNAME target answer rewrite,
using the proposed syntax `rewrite cname def.example.com. xyz.example.com.`
```
abc.example.com IN CNAME xyz.example.com
xyz.example.com IN A 5.6.7.8
```

Amila-Rukshan:
@chrisohaver, Thanks for the quick response! 
[...]

PR #6004, opened 2023-04-02
plugin/rewrite: Introduce cname target rewrite rule

Amila-Rukshan:
This PR extends the rewrite plugin capable of rewriting `cname` `taget` in the DNS response. 

https://github.com/coredns/coredns/issues/5527

https://github.com/coredns/coredns/blob/master/plugin/rewrite/README.md
This rewrite plugin documentation needs to be updated

No

### DIFF
--- a/plugin/rewrite/cname_target.go
+++ b/plugin/rewrite/cname_target.go
@@ -131,6 +131,9 @@ func newCNAMERule(nextAction string, args ...string) (Rule, error) {
 			return nil, fmt.Errorf("unknown cname rewrite type: %s", rewriteType)
 		}
 		paramFromTarget, paramToTarget = strings.ToLower(args[1]), strings.ToLower(args[2])
+	} else if len(args) == 2 {
+		rewriteType = CNameExactMatch
+		paramFromTarget, paramToTarget = strings.ToLower(args[0]), strings.ToLower(args[1])
 	} else {
 		return nil, fmt.Errorf("too few (%d) arguments for a cname rule", len(args))
 	}
--- a/plugin/rewrite/cname_target.go
+++ b/plugin/rewrite/cname_target.go
@@ -3,21 +3,31 @@ package rewrite
 import (
 	"context"
 	"fmt"
+	"net"
 	"strings"
 
 	"github.com/coredns/coredns/plugin/pkg/log"
+	"github.com/coredns/coredns/plugin/pkg/upstream"
 	"github.com/coredns/coredns/request"
 	"github.com/miekg/dns"
 )
+
+// UpstreamInt wraps the Upstream API for dependency injection during testing
+type UpstreamInt interface {
+	Lookup(ctx context.Context, state request.Request, name string, typ uint16) (*dns.Msg, error)
+}
 
 // cNameResponseRule is cname target rewrite rule.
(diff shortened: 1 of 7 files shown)


greenpau (changes requested):
@Amila-Rukshan , what should be added to the `README.md` considering this change?

Amila-Rukshan:
@greenpau, I have now updated the PR with the readme file

greenpau:
@Amila-Rukshan , will review early next week. Meanwhile, please look at the coverage report and see any opportunities for improvement.

plugin/rewrite/cname_target.go
```diff
@@ -0,0 +1,161 @@
...
+	CNameExactMatch RewriteType = "exact"
+	// CNamePrefixMatch is the type for prefix match of cname target
+	CNamePrefixMatch RewriteType = "prefix"
+	// CNameSuffixMatch is the type for suffix match of cname target
+	CNameSuffixMatch RewriteType = "suffix"
+	// CNameSubstringMatch is the type for substring match of cname target
+	CNameSubstringMatch RewriteType = "substring"
+	// CNameRegexMatch is the type for regex match of cname target
+	CNameRegexMatch RewriteType = "regex"
+)
```
greenpau: @Amila-Rukshan , why do you need to add the above when we already have the following?

https://github.com/coredns/coredns/blob/0063d7a80c0db18069429c775e4b95a5c0b4b69c/plugin/rewrite/name.go#L115-L124

greenpau:
@Amila-Rukshan , this is more of a general question. 

We have existing `name` rule type.

https://github.com/coredns/coredns/blob/0063d7a80c0db18069429c775e4b95a5c0b4b69c/plugin/rewrite/rewrite.go#L126-L127

Is it possible to incorporate `cname` rule type in `name` rule type? What would be pros and cons here? It appears that the approach in this PR is straight-forward. I am thinking about duplication/de-duplication aspects of this PR.

plugin/rewrite/cname_target.go
Amila-Rukshan: Updated the code to reuse the rewrite types

Amila-Rukshan:
@greenpau, It's better if we can combine `name` and `cname` two to reduce the number of lines. But `name` rewrite does not cause another upstream DNS resolution as we do with `cname` rule to create a new DNS response based on the new cname target. There is that functional difference between `name` and `cname` rewrite rules and I feel like combing them will reduce the simplicity of keeping individual rules. Considering that didn't look into combining them. `rewrite` plugin codebase already has individual rules for `ttl`, `edns0`, `class`..etc as well.

greenpau (approved):

greenpau:
> Considering that didn't look into combining them. rewrite plugin codebase already has individual rules for ttl, edns0, class..etc as well.

@Amila-Rukshan , it makes sense. 

Please check the failing tests.

Amila-Rukshan:
seems the coredns repo was not accessible to the workflow:
```
Error: fatal: unable to access 'https://github.com/coredns/coredns/': The requested URL returned error: 429
```
@greenpau, will you able to retrigger it?

greenpau:
@Amila-Rukshan , thank you for contributing! 👍

PR #6016, opened 2023-04-05
Build binary in docker and add docker build-args

ivyostosh:
This PR allows full customization on build environment through docker build, including binary build golang image version, debian build image version, and base layer image version.

N/A but can create one if needed.

Change is self-explanatory and doesn't require additional documentation. Users can still run docker build the same way as before.

No.


SuperQ (changes requested):
Sorry, but we prefer to keep the binary build outside of the Dockerfile to keep the image layers lightweight.

ivyostosh:
@SuperQ Understand the community's intent to keep the image lightweight. Will you all be open to the idea of having this option as a separate Make target and its own Dockerfile? By doing so, the default Docker behavior wouldn't be impacted, and users that would like to build binary in docker can have it as an option.

ivyostosh:
@SuperQ Any thoughts on this [question](https://github.com/coredns/coredns/pull/6016#issuecomment-1499224266)?

chrisohaver:
I think a custom Dockerfile would be somewhat straightforward to maintain external to this repo.

Are there compelling reasons for requiring this degree of customization to be maintained in coredns/coredns?

ivyostosh:
@chrisohaver Agree this can be maintained externally - we plan to do so if this PR doesn't go through :)

For some context, I work for AWS and we have multiple teams leverage the core-dns github repo. We want to build the binary in Docker so we have full control of the go binary used and address CVEs faster. As this has impacts across different teams, having this change upstream is easier to maintain. 

Now, I certainly cannot speak for users outside of our org on if this is a desired use case, but figured I would raise the PR to get some feedback first :)

chrisohaver:
I think a custom Dockerfile would be somewhat straightforward to maintain external to this repo.

PR #6017, opened 2023-04-07
Add new plugin 

snail2sky:
Because origin mysql plugin performance is not enough in the production environment, and it strongly depends on the stability of mysql.
The `mysql_extend` fixed above question
None
No


SuperQ (changes requested):
This needs justification.

* Why is this a separate plugin, and not improvements to the original plugin?
* How is this an improvement?
* Why is "performance not enough"?

snail2sky:
It may be because of the different ideas. I have made a lot of changes to the original plug-in, which will definitely cause the original plug-in to be incompatible. The fields stored in the DB used by the original plug-in contain json strings, and I think users maintain json in the database It is extremely inconvenient. At that time, a PR was also opened to the original author, but it may be because of incompatibility that it was impossible to merge.&nbsp;My improved plugin address is&nbsp;https://github.com/snail2sky/coredns_mysql.git


First of all, by optimizing the DB connection pool, the connection pool is only established once when the plug-in is started, and the remaining requests will reuse this connection instead of re-establishing the TCP connection. The original plug-in will re-establish the tcp connection every time a new request is processed. This also answers your third question. Every query of the plug-in is like http1.0, and the underlying connection cannot be reused, so it must be inefficient. At that time, I also conducted a stress test and found that the performance was very poor. , only a few thousand QPS, and the stability of the original plug-in and its dependence on mysql, which is almost unacceptable in a production environment. Even if there is a problem with mysql, DNS should continue to serve, otherwise this kind of infrastructure would not dare used in production


In addition, the original plug-in does not expose any indicators. We have no way to monitor the plug-in running in this way, and we cannot achieve closed-loop management. There are also some other modifications, such as adding some debug logs. Once the plug-in itself has problems, you can Easy to debug. And even if this plug-in cannot be officially included, I will actively maintain it, because this is the first time I have participated in an open source project, and it is not necessarily open source that is officially maintained, as long as it can solve everyone's problems, that is enough. 



thanks for your question!








                       
Original Email
                       
                     

Sent Time:2023/4/7 14:10

Cc recipient:"snail"< ***@***.*** &gt;;"Author"< ***@***.*** &gt;;

Subject:Re: [coredns/coredns] Add new plugin  (PR #6017)




 
@SuperQ requested changes on this pull request.
 
This needs justification.
  
How is this an improvement?
 
Why is "performance not enough"?

SuperQ:
Those are all great improvements.

I agree, the json blobs in the database are not good. Having a proper DNS schema is an improvement.

Also having a connection pool is something I would consider a requirement for this kind of plugin.

I normally suggest attempting to work with the upstream plugin maintainer to incrementally improve on their project. Only when a good faith attempt has failed is it really necessary to hard fork a project. I looked at the original repo and the PR you opened. You simply dumped a huge messy PR request on the original repo with no context. This approach is hard to review and merge.

Right now the mysql plugin is external, I think we need to first decide if we want to have an internal plugin. I think this version is much closer to something that would be acceptable as an internal plugin.

snail2sky:
Yes, I seem to have made a common novice mistake and put all the changes in a PR, which is really bad, but it is also difficult to deal with, because it consumes a lot of my time to make changes based on the original plugin I thought it would take a long time to revise and submit the PR bit by bit from the beginning, so I might as well rewrite it myself. To be honest, this is a bit disrespectful to the original author, but his code is really difficult to modify, a bit pain~~

In addition, I see that the original author's warehouse hasn't been updated for a long time, and he didn't reply a word to my PR, which gave me the impression that he didn't seem to be maintaining this plugin, which made me want to give up the idea of ​​following







                       
Original Email
                       
                     

Sent Time:2023/4/7 14:53

Subject:Re: [coredns/coredns] Add new plugin  (PR #6017)




 
Those are all great improvements.
 
I agree, the json blobs in the database are not good. Having a proper DNS schema is an improvement.
 
Also having a connection pool is something I would consider a requirement for this kind of plugin.
 
I normally suggest attempting to work with the upstream plugin maintainer to incrementally improve on their project. Only when a good faith attempt has failed is it really necessary to hard fork a project. I looked at the original repo and the PR you opened. You simply dumped a huge messy PR request on the original repo with no context. This approach is hard to review and merge.
 
Right now the mysql plugin is external, I think we need to first decide if we want to have an internal plugin. I think this version is much closer to something that would be acceptable as an internal plugin.

SuperQ:
Yea, I get it. Sometimes it's better to rewrite something from scratch. Given the amount of changes you've made, I think it's justified in this case.

If the old plugin is unmaintained, we should drop or replace it.

snail2sky:
Original Email
                       
                     

Sent Time:2023/4/7 15:25

Subject:Re: [coredns/coredns] Add new plugin  (PR #6017)




 
Yea, I get it. Sometimes it's better to rewrite something from scratch. Given the amount of changes you've made, I think it's justified in this case.
 
If the old plugin is unmaintained, we should drop or replace it.

PR #6018, opened 2023-04-07
Add new plugin

snail2sky:
How is this an improvement? and Why is "performance not enough"?

It may be because of the different ideas. I have made a lot of changes to the original plug-in, which will definitely cause the original plug-in to be incompatible. The fields stored in the DB used by the original plug-in contain json strings, and I think users maintain json in the database It is extremely inconvenient. At that time, a PR was also opened to the original author, but it may be because of incompatibility that it was impossible to merge. My improved plugin address is https://github.com/snail2sky/coredns_mysql.git

None
No


SuperQ:
Please don't frequently open and close pull requests and then open new requests. It causes a lot of noise for maintainers.

It is better to simply push and/or force-push updates.

I would also recommend using [fork and branch workflows](https://blog.scottlowe.org/2015/01/27/using-fork-branch-git-workflow/) rather than push-to-master on your fork. This will allow for easier maintenance of your changes.

snail2sky:
Okay, I will do this. When I submitted the pr this time, I didn’t add signed, which caused the automatic operation to fail.


                       
Original Email
                       
                     

Sent Time:2023/4/7 15:28

Subject:Re: [coredns/coredns] Add new plugin (PR #6018)




 
Please don't frequently open and close pull requests and then open new requests. It causes a lot of noise for maintainers.
 
It is better to simply push and/or force-push updates.
 
I would also recommend using fork and branch workflows rather than push-to-master on your fork. This will allow for easier maintenance of your changes.

Issue #1663, opened 2018-04-06
feature request: sql plugin


glerchundi:
I would like know if you guys are open to have a plugin which talks to a SQL database. As far as I can see there are two ways to implement this:
1. Create generic table with all the records in there and governed by another column which defines the zone in order to query entries. Entry data would look like a blob or bytes array encoded in a way coredns can decode. This makes it difficult to feed this database from the outside because feeder should know the schema as well as how it should be encoded.
2. Create tables per record type like the way [`coredns-dbcheck`](https://github.com/creamfinance/coredns-dbcheck) does.

Some examples would look like these:

Standard schema, the one proposed by CoreDNS:

```
database postgres://user:pass@host?db . {
}
```

Custom schema:
```
database postgres://user:pass@host?db . {
   a_query = "SELECT r.p1,r.p2,r.p3,r.p4 FROM a_records r INNER JOIN zones z ON r.zone_id = z.id WHERE z.zone = '{{.}}'"
}
```

I would go with the second option, WDYT? @miekg 

I'm open to contribute on this as well as in #1226.

stp-ip:
One option would be to extend the file plugin with a way to grab files form a db/object store instead of just the filesystem.
The schema is just plain dns zone files and on the CoreDNS side it's a slight change.

With the etcd plugin and later the etcd3 support there is already some sort of DB function available. It has a special syntax, but this would be the same with any other DB implementation except for just storing zone file blobs in it (see above).

glerchundi:
> One option would be to extend the file plugin with a way to grab files form a db/object store instead of just the filesystem.

That would be interesting but I think the use case for the database is different. Although an object store sounds like a very good idea! If we abstract the file plugin with something like [afero](https://github.com/spf13/afero), [storage](https://github.com/sajari/storage), [billy](https://github.com/src-d/go-billy) or [vfs](https://github.com/blang/vfs) it will open CoreDNS up to a wide range of possibilities! 

> With the etcd plugin and later the etcd3 support there is already some sort of DB function available. It has a special syntax, but this would be the same with any other DB implementation except for just storing zone file blobs in it (see above).

It is and we've evaluated it but it doesn't fit within our use case. We want to have as less moving parts as possible and having another component like etcd, consul or whatever when we are already operating another datastore makes operations harder. Not to mention that sadly the big cloud players don't offer any managed services for them :(

[...]

PR #6019, opened 2023-04-07
New plugin named mysql_extend

snail2sky:
How is this an improvement? and Why is "performance not enough"?

It may be because of the different ideas. I have made a lot of changes to the original plug-in, which will definitely cause the original plug-in to be incompatible. The fields stored in the DB used by the original plug-in contain json strings, and I think users maintain json in the database It is extremely inconvenient. At that time, a PR was also opened to the original author, but it may be because of incompatibility that it was impossible to merge. My improved plugin address is https://github.com/snail2sky/coredns_mysql.git
None
No


plugin/mysql_extend/init.go
```diff
@@ -0,0 +1,185 @@
...
+	} else {
+		createTableCount.With(prometheus.Labels{"status": "success", "table_name": m.zonesTable}).Inc()
+	}
+
+	_, err = m.db.Exec(`
+        CREATE TABLE IF NOT EXISTS ` + m.recordsTable + ` (
+            id INT NOT NULL AUTO_INCREMENT,
+            zone_id INT NOT NULL,
+            hostname VARCHAR(512) NOT NULL,
+            type VARCHAR(10) NOT NULL,
```
SuperQ, line 170: I wonder if the record type should be an [ENUM column](https://dev.mysql.com/doc/refman/8.0/en/enum.html) of the most common record types.

plugin/mysql_extend/hooks.go
```diff
@@ -0,0 +1,173 @@
...
+		return
+	}
+	logger.Debugf("Success to dump data to local: %#v", pureRecord)
+	dumpLocalData.With(prometheus.Labels{"status": "success"}).Inc()
+}
+
+func (m *Mysql) openDB() (*sql.DB, error) {
+	db, err := sql.Open("mysql", m.dsn)
+	if err != nil {
+		openMysqlCount.With(prometheus.Labels{"status": "fail"}).Inc()
```
SuperQ, line 133: Rather than a `status="fail"` label, it's usually better to have two metrics.

```
coredns_mysql_connections_total
coredns_mysql_connections_failed_total
```

This is a typical Prometheus pattern. One metric to count all connection attempts, one to count just the failures.

plugin/mysql_extend/init.go
```diff
@@ -0,0 +1,185 @@
...
+				m.queryRecordSQL = c.Val()
+			default:
+				return c.Errf("unknown property '%s'", c.Val())
+			}
+		}
+	}
+	return nil
+}
+
+func (m *Mysql) createTables() {
```
SuperQ, line 149: I don't think the plugin should create the tables. The DNS server should not have any write access to the database.

SuperQ:
If this is a full replacement for the current external mysql plugin, let's just call it `mysql` rather than `mysql_extend`.

I'm also wondering if we should re-open the old SQL plugin feature request: https://github.com/coredns/coredns/issues/1663 to create a more generic SQL plugin that supports multiple database backends.

snail2sky:
Makes sense, I can modify this


                       
Original Email
                       
                     

Sent Time:2023/4/7 15:16

Subject:Re: [coredns/coredns] New plugin named mysql_extend (PR #6019)




 
@SuperQ commented on this pull request.
 
 
In plugin/mysql_extend/init.go:
 &gt; +            UNIQUE KEY (zone_name) +        ); +    `) +	if err != nil { +		createTableCount.With(prometheus.Labels{"status": "fail", "table_name": m.zonesTable}).Inc() +		logger.Error(err) +	} else { +		createTableCount.With(prometheus.Labels{"status": "success", "table_name": m.zonesTable}).Inc() +	} + +	_, err = m.db.Exec(` +        CREATE TABLE IF NOT EXISTS ` + m.recordsTable + ` ( +            id INT NOT NULL AUTO_INCREMENT, +            zone_id INT NOT NULL, +            hostname VARCHAR(512) NOT NULL, +            type VARCHAR(10) NOT NULL,  
I wonder if the record type should be an ENUM column of the most common record types.

snail2sky:
Okay, I will also modify this, I thought that some other status or latitude information can be directly used tags~&nbsp;







                       
Original Email
                       
                     

Sent Time:2023/4/7 15:20

Subject:Re: [coredns/coredns] New plugin named mysql_extend (PR #6019)




 
@SuperQ commented on this pull request.
 
 
In plugin/mysql_extend/hooks.go:
 &gt; +		return +	} +	if err := os.WriteFile(m.dumpFile, content, safeMode); err != nil { +		logger.Error(err) +		logger.Errorf("Failed to dump data to local: %s", err) +		dumpLocalData.With(prometheus.Labels{"status": "fail"}).Inc() +		return +	} +	logger.Debugf("Success to dump data to local: %#v", pureRecord) +	dumpLocalData.With(prometheus.Labels{"status": "success"}).Inc() +} + +func (m *Mysql) openDB() (*sql.DB, error) { +	db, err := sql.Open("mysql", m.dsn) +	if err != nil { +		openMysqlCount.With(prometheus.Labels{"status": "fail"}).Inc()  
Rather than a status="fail" label, it's usually better to have two metrics.
 coredns_mysql_connections_total coredns_mysql_connections_failed_total  
This is a typical Prometheus pattern.

snail2sky:
That’s right, you are right, the main reason is that it was more convenient to test the database at that time, and coredns only has read permission for mysql during the normal connection process







                       
Original Email
                       
                     

Sent Time:2023/4/7 15:23

Subject:Re: [coredns/coredns] New plugin named mysql_extend (PR #6019)




 
@SuperQ commented on this pull request.
 
 
In plugin/mysql_extend/init.go:
 &gt; +				} +				m.queryZoneSQL = c.Val() +			case "query_record_sql": +				if !c.NextArg() { +					return c.ArgErr() +				} +				m.queryRecordSQL = c.Val() +			default: +				return c.Errf("unknown property '%s'", c.Val()) +			} +		} +	} +	return nil +} + +func (m *Mysql) createTables() {  
I don't think the plugin should create the tables. The DNS server should not have any write access to the database.

snail2sky:
I think this is a very good idea, because if it is a general-purpose database engine, then users will no longer have to worry about whether to add a new database maintained by themselves in order to use this plug-in, and it can also greatly reduce the user's workload. pressure. I thought about it when I first wrote this plug-in, but unfortunately I think my programming ability is a bit weak, and I use very few databases, so in the end only mysql database is supported as the storage backend







                       
Original Email
                       
                     

Sent Time:2023/4/7 15:30

Subject:Re: [coredns/coredns] New plugin named mysql_extend (PR #6019)




 
If this is a full replacement for the current external mysql plugin, let's just call it mysql rather than mysql_extend.
 
I'm also wondering if we should re-open the old SQL plugin feature request: #1663 to create a more generic SQL plugin that supports multiple database backends.

plugin/mysql_extend/init.go
snail2sky: I agree with this. I'll go back and edit this next time

plugin/mysql_extend/hooks.go
snail2sky: I'm really not sure about this, thanks for reminding me

plugin/mysql_extend/init.go
snail2sky: This is just for the convenience of development. Although it has no special impact on the DB, it is not very good overall and does not conform to the rules. But because I never thought that the plug-in I wrote could be officially included by coredns, so Just added this feature, I will remove it right away

chrisohaver:
This should be an external plugin.

chrisohaver:
My apologies, I think there may have been a misunderstanding. @snail2sky had opened https://github.com/coredns/coredns.io/pull/290 to add the `mysql_extend` plugin as an external plugin. However, it was added to the wrong directory (`plugins` instead of `explugins`), and I had carelessly merged it without noticing that error.  Once the error was brought to my attention, I reverted the change, and requested that the plugin be moved to the correct location (`explugins`).  It seems my request may have been misunderstood as an approval to adopt the plugin into coredns's internal plugins. 

I agree with @SuperQ in that a generic database backend might be more appropriate for inclusion in coredns than a mysql specific one.

snail2sky:
> My apologies, I think there may have been a misunderstanding. @snail2sky had opened [coredns/coredns.io#290](https://github.com/coredns/coredns.io/pull/290) to add the `mysql_extend` plugin as an external plugin. However, it was added to the wrong directory (`plugins` instead of `explugins`), and I had carelessly merged it without noticing that error. Once the error was brought to my attention, I reverted the change, and requested that the plugin be moved to the correct location (`explugins`). It seems my request may have been misunderstood as an approval to adopt the plugin into coredns's internal plugins.抱歉，我想可能有误会。 @snail2sky 已打开 coredns/coredns.io#290 将 mysql_extend 插件添加为外部插件。然而，它被添加到错误的目录（plugins 而不是 explugins），我不小心合并了它而没有注意到那个错误。一旦错误引起我的注意，我就恢复了更改，并要求将插件移动到正确的位置 (explugins)。看来我的请求可能被误解为批准将插件采用到 coredns 的内部插件中。
> 
> I agree with @SuperQ in that a generic database backend might be more appropriate for inclusion in coredns than a mysql specific one.我同意@SuperQ 的观点，通用数据库后端可能比特定于 mysql 的后端更适合包含在 coredns 中。

Yes, I also felt a little curious at the beginning. How can I say that my plug-in will be directly included in the internal plug-in~ I was flattered all day long, haha

snail2sky:
> My apologies, I think there may have been a misunderstanding. @snail2sky had opened [coredns/coredns.io#290](https://github.com/coredns/coredns.io/pull/290) to add the `mysql_extend` plugin as an external plugin. However, it was added to the wrong directory (`plugins` instead of `explugins`), and I had carelessly merged it without noticing that error. Once the error was brought to my attention, I reverted the change, and requested that the plugin be moved to the correct location (`explugins`). It seems my request may have been misunderstood as an approval to adopt the plugin into coredns's internal plugins.抱歉，我想可能有误会。 @snail2sky 已打开 coredns/coredns.io#290 将 mysql_extend 插件添加为外部插件。然而，它被添加到错误的目录（plugins 而不是 explugins），我不小心合并了它而没有注意到那个错误。一旦错误引起我的注意，我就恢复了更改，并要求将插件移动到正确的位置 (explugins)。看来我的请求可能被误解为批准将插件采用到 coredns 的内部插件中。
> 
> I agree with @SuperQ in that a generic database backend might be more appropriate for inclusion in coredns than a mysql specific one.我同意@SuperQ 的观点，通用数据库后端可能比特定于 mysql 的后端更适合包含在 coredns 中。

So should I re-issue a pr to the coredns.io repo and add it to the external plugin?

chrisohaver:
> So should I re-issue a pr to the coredns.io repo and add it to the external plugin?

That aligns best with your original intent in https://github.com/coredns/coredns.io/pull/290.

PR #6022, opened 2023-04-11
Updates for using ecs in metadata

YoungjuWang:

For COREDNS to respond utilizing the ECS value as shown in the example below.

```bash
# dig @8.8.8.8 o-o.myaddr.l.google.com txt +subnet=100.20.30.0/24 +short
"173.194.168.3"
"edns0-client-subnet 100.20.30.0/24"
```

q? how to use 'ecs' metadata?
a
```
  template IN TXT example.com {
        match whoami\.(.*)
        answer "{{ .Name }} 60 IN TXT {{ .Remote }}"
        answer "{{ .Name }} 60 IN TXT \"edns0-client-subnet {{ .Meta \"geoip/ecs\" }}\""
        fallthrough
  }
```

Maybe geoip plugin's README.md is need to be updated.

no

### DIFF
--- a/plugin/geoip/city.go
+++ b/plugin/geoip/city.go
@@ -55,4 +55,8 @@ func (g GeoIP) setCityMetadata(ctx context.Context, data *geoip2.City) {
 	metadata.SetValueFunc(ctx, pluginName+"/postalcode", func() string {
 		return postalCode
 	})
+	ecs := g.ecs.address + "/" + g.ecs.netmask
+	metadata.SetValueFunc(ctx, pluginName+"/ecs", func() string {
+		return ecs
+	})
 }
--- a/plugin/geoip/geoip.go
+++ b/plugin/geoip/geoip.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"path/filepath"
+	"strconv"
 
 	"github.com/coredns/coredns/plugin"
 	clog "github.com/coredns/coredns/plugin/pkg/log"
@@ -23,6 +24,12 @@ type GeoIP struct {
 	Next  plugin.Handler
 	db    db
 	edns0 bool
+	ecs   ecs
+}
+
+type ecs struct {
+	address string
+	netmask string
 }
 
 type db struct {
@@ -85,6 +92,9 @@ func (g GeoIP) Metadata(ctx context.Context, state request.Request) context.Cont
 			for _, s := range o.Option {
 				if e, ok := s.(*dns.EDNS0_SUBNET); ok {
 					srcIP = e.Address
+					// Save ECS contents from request.
+					g.ecs.address = e.Address.String()
+					g.ecs.netmask = strconv.Itoa(int(e.SourceNetmask))
 					break
 				}
 			}


snebel29:
Hi @YoungjuWang 

First of all thanks for your contribution, before I comment on your pull request and the code changes it contains I would like to understand what are you trying to achieve, but specially the use cases that you have in mind.

If I understood correctly, you are trying to set the address and netmask as metadata value keys when EDNS client subnet was provided in the request for later consumption by other plugins, is that right?

What are the use cases that you have in mind, and you are trying to support?

Providing that additional context would be very helpful, to provide you with good code review feedback.

Thanks in advance.

snebel29 (commented):
Hi @YoungjuWang

First of all thanks for your contribution, before I comment on your pull request and the code changes it contains I would like to understand what are you trying to achieve, but specially the use cases that you have in mind.

If I understood correctly, you are trying to set the address and netmask as metadata value keys when EDNS client subnet was provided in the request for later consumption by other plugins, is that right?

What are the use cases that you have in mind, and you are trying to support?

Providing that additional context would be very helpful, to provide you with good code review feedback.

Thanks in advance.

YoungjuWang:
Hi @snebel29

I appreciate your feedback.

> If I understood correctly, you are trying to set the address and netmask as metadata value keys when EDNS client subnet was provided in the request for later consumption by other plugins, is that right?

Yes, I want to make the ECS address and netmask available to other plugins as metadata.

Thanks

snebel29:
> Hi @snebel29
> 
> 
> 
> I appreciate your feedback.
> 
> 
> 
> > If I understood correctly, you are trying to set the address and netmask as metadata value keys when EDNS client subnet was provided in the request for later consumption by other plugins, is that right?
> 
> 
> 
> Yes, I want to make the ECS address and netmask available to other plugins as metadata.
> 
> 
> 
> Thanks

What about use cases? Do you have in mind already some? Or you just want to add this metadata with the expectation someone will find one?

Cheers.

YoungjuWang:
> What about use cases? Do you have in mind already some? Or you just want to add this metadata with the expectation someone will find one?
> 
> Cheers.

If the ecs metadata will be available in the view plugin, it seems like it could be utilized to set the view based on the actual client ip (if passed as an ecs value) rather than the client_ip i.e. the resolver's ip.

It could also be utilized to determine the ecs value currently being used by the client, such as `o-o.myaddr.l.google.com` in the example. This could also be used to verify that the metadata(i.e City) returned by the geoIP plugin is correct for the currently passed ecs.

Thanks.

snebel29:
Putting the current code of this PR aside, I personally wouldn't mind adding ECS as metadata from the geoIP plugin if this end up being needed or convenient, and I am happy to be convinced this is the case.

But at first glance geoIP does not feel the right place to me giving that the ECS option should be universally available to all and any plugin through `request.Request` without having to depend on the geoIP plugin.

> If the ecs metadata will be available in the view plugin, it seems like it could be utilised to set the view based on the actual client ip (if passed as an ecs value) rather than the client_ip i.e. the resolver's ip.

I am not sure, that to enable the use of ECS value in the "view" plugin the best approach is to create a dependency on the geoIP plugin, wouldn't it be simpler (and cleaner) to just add a new DNS query function?

https://github.com/coredns/coredns/blob/0063d7a80c0db18069429c775e4b95a5c0b4b69c/plugin/pkg/expression/expression.go#L12-L47

> It could also be utilized to determine the ecs value currently being used by the client, such as o-o.myaddr.l.google.com in the example. This could also be used to verify that the metadata(i.e City) returned by the geoIP plugin is correct for the currently passed ecs.

I am not sure I fully understand your point here, how exactly are you planning to consume that metadata value and from what other plugin? I wonder if all you are suggesting and may need is to add this field to the "log" plugin? Can you clarify?

Since the discussion is crossing the boundaries of the geoIP plugin, and to have additional feeback Do you have any opinion @chrisohaver @johnbelamaric @SuperQ ?

Cheers.

chrisohaver:
I also think this should not be attached to the geo-ip plugin.

Maybe include this as part of the default metadata created by the metadata plugin?

But perhaps it is better suited in a separate new plugin such as `edns_metadata` or just `edns`, that registers edns data as coredns metadata ... 

IIRC, it would be similar to the external plugin ... https://github.com/coredns/metadata_edns0.  Maybe we should extend/rename that and pull it in to be an internal plugin.

YoungjuWang:
Let me check [metadata_edns0](https://github.com/coredns/metadata_edns0) plugin.
So I'll just close my PR, right?

snebel29:
> Let me check [metadata_edns0](https://github.com/coredns/metadata_edns0) plugin. So I'll just close my PR, right?

My suggestion would be to close this PR, look at the couple of options discussed with pros/cons to support your uses cases and open a discussion to propose a concrete approach, once you get buy-in from maintainers you could work on a new PR.

I hope that helps, although I am only owner of geoIP plugin I would not mind to be keep on the loop (and contribute, even if only with code reviews to start with) for a potential new plugin.

Cheers.

PR #5991, opened 2023-03-24
plugin/kubernetes: expose client-go internal request metrics

iyashu:
### 1. Why is this pull request needed and what does it do?
This pull request exposes following K8s client level metrics to monitor K8s requests latency & status codes. It is useful in setting up alerts (if required) in case coredns fails to query kube-apiserver (due to networking issues or application level issues) making it to serve stale records.
```
# HELP coredns_kubernetes_rest_client_rate_limiter_duration_seconds Client side rate limiter latency in seconds. Broken down by verb and URL.
# TYPE coredns_kubernetes_rest_client_rate_limiter_duration_seconds histogram
coredns_kubernetes_rest_client_rate_limiter_duration_seconds_bucket{url="10.24.1.228:6443",verb="GET",le="0.001"} 14
coredns_kubernetes_rest_client_rate_limiter_duration_seconds_bucket{url="10.24.1.228:6443",verb="GET",le="0.002"} 14
coredns_kubernetes_rest_client_rate_limiter_duration_seconds_bucket{url="10.24.1.228:6443",verb="GET",le="0.004"} 14
coredns_kubernetes_rest_client_rate_limiter_duration_seconds_bucket{url="10.24.1.228:6443",verb="GET",le="0.008"} 14
coredns_kubernetes_rest_client_rate_limiter_duration_seconds_bucket{url="10.24.1.228:6443",verb="GET",le="0.016"} 14
coredns_kubernetes_rest_client_rate_limiter_duration_seconds_bucket{url="10.24.1.228:6443",verb="GET",le="0.032"} 14
coredns_kubernetes_rest_client_rate_limiter_duration_seconds_bucket{url="10.24.1.228:6443",verb="GET",le="0.064"} 14
coredns_kubernetes_rest_client_rate_limiter_duration_seconds_bucket{url="10.24.1.228:6443",verb="GET",le="0.128"} 14
coredns_kubernetes_rest_client_rate_limiter_duration_seconds_bucket{url="10.24.1.228:6443",verb="GET",le="0.256"} 14
coredns_kubernetes_rest_client_rate_limiter_duration_seconds_bucket{url="10.24.1.228:6443",verb="GET",le="0.512"} 14
coredns_kubernetes_rest_client_rate_limiter_duration_seconds_bucket{url="10.24.1.228:6443",verb="GET",le="+Inf"} 14
coredns_kubernetes_rest_client_rate_limiter_duration_seconds_sum{url="10.24.1.228:6443",verb="GET"} 0.000188625
[...]

PR #6040, opened 2023-04-23
plugin/kubernetes: rename client-go metrics label "url" to "host"

iyashu:
See https://github.com/coredns/coredns/pull/5991#issuecomment-1518706256 for more details. As we're using apiserver request url host as label value, I think it makes sense to update label name as `host` instead of `url`.  

NA.

No backward incompatible changes from last official release.


iyashu:
@chrisohaver @Tantalor93 @SuperQ ptal.

chrisohaver (approved):

PR #6052, opened 2023-04-25
fix some comments

cuishuang:
fix some comments

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -605,7 +605,7 @@ func (dns *dnsControl) detectChanges(oldObj, newObj interface{}) {
 			dns.updateModified()
 		}
 		if emod {
-			dns.updateExtModifed()
+			dns.updateExtModified()
 		}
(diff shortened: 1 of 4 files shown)


plugin/kubernetes/controller.go
```diff
@@ -748,7 +748,7 @@ func (dns *dnsControl) updateModified() {
 	atomic.StoreInt64(&dns.modified, unix)
 }
 
-// updateExtModified set dns.extModified to the current time.
+// updateExtModifed set dns.extModified to the current time.
 func (dns *dnsControl) updateExtModifed() {
```
chrisohaver: Please instead change the function name to `updateExtModified`.

chrisohaver (changes requested):

plugin/kubernetes/controller.go
cuishuang: Thanks.Done

chrisohaver (approved):

PR #6068, opened 2023-05-03
Fix temp file close error

SuperQ:

Avoid Go 1.20 test error by not attempting to close the testing temp file unless there was an error in Read().
* Use a CreateTemp() to create unique test files.
* Defer the deletion of the temp file.
* Fix typo in `err` handling.

Workaround for: https://github.com/golang/go/issues/59938


SuperQ:
Thanks for the pointers @bitoku

chrisohaver (approved):

PR #6074, opened 2023-05-04
bugfix prometheus/pprof port not released on upgrade

kokizzu:
metrics (prometheus) and pprof plugin does not release listening port when doing coredns upgrade
```
kill -USR2 `cat /tmp/coredns.pid` 
```
so need to listen to SIGUSR2 signal, then shutdown the server so next fork (upgraded binary) will not have an error.

-

-

no


chrisohaver:
Can this be tested in CI?
Does this also need to be applied to other plugins that listen on ports? e.g. _health_, _ready_?

kokizzu:
or maybe I should create PR on caddy, to add OnUpgrade -- to attach callback for plugins that have custom listen port that should be shutted down before upgrade?

kokizzu:
rip the pull bot is resetting the changes

PR #6100, opened 2023-05-15
Handle UDP responses that overflow with TC bit with test case

SriHarsha001:

Description in this PR - https://github.com/coredns/coredns/pull/6003 

Code changes in plugin/pkg/proxy/connect.go file is copied from above PR - https://github.com/coredns/coredns/pull/6003 

https://github.com/coredns/coredns/issues/5953
https://github.com/coredns/coredns/issues/5998

Not sure. I didn't see any.

No.

### DIFF
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -18,20 +18,6 @@ import (
 	"github.com/miekg/dns"
 )
 
-// // Error represents a DNS error.
-// type Error struct{ err string }
-
-// func (e *Error) Error() string {
-// 	if e == nil {
-// 		return "dns: <nil>"
-// 	}
-// 	return "dns: " + e.err
-// }
-
-// var (
-// 	ErrBufFromhere error = &Error{err: "buffer size too small"}
-// )
-
 // limitTimeout is a utility function to auto-tune timeout values
 // average observed time is moved towards the last observed delay moderated by a weight
 // next timeout to use will be the double of the computed average, limited by min and max frame.
@@ -142,15 +128,6 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 			// Instead of returning an error, return an empty response with TC bit set. This will make the
 			// client retry over TCP (if that's supported) or at least receive a clean
 			// error. The connection is still good so we break before the close.
-
-			// This is for scenario in which upstream sets the truncated flag, but doesn't truncate the response.
-			// We get a "buffer size too small" error instead of overflow. Retrying for this scenario.
-			// dnsErrBufOccured := false
-			// if dnsErr, ok := err.(*dns.Error); ok {
-			// 	if dnsErr.Error() == dns.ErrBuf.Error() {
-			// 		dnsErrBufOccured = ok
-			// 	}
-			// }
 
 			dnsErrBufOccured := false
 			var perr *dns.Error
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -6,6 +6,7 @@ package proxy
 
 import (
 	"context"
+	"errors"
 	"io"
 	"strconv"
 	"strings"
@@ -16,6 +17,20 @@ import (
 
 	"github.com/miekg/dns"
 )
+
+// // Error represents a DNS error.
+// type Error struct{ err string }
+
+// func (e *Error) Error() string {
+// 	if e == nil {
+// 		return "dns: <nil>"
+// 	}
+// 	return "dns: " + e.err
+// }
+
+// var (
+// 	ErrBufFromhere error = &Error{err: "buffer size too small"}
+// )
 
 // limitTimeout is a utility function to auto-tune timeout values
 // average observed time is moved towards the last observed delay moderated by a weight
@@ -127,8 +142,27 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 			// Instead of returning an error, return an empty response with TC bit set. This will make the
 			// client retry over TCP (if that's supported) or at least receive a clean
 			// error. The connection is still good so we break before the close.
-			if proto == "udp" && strings.Contains(err.Error(), "overflow") {
+
+			// This is for scenario in which upstream sets the truncated flag, but doesn't truncate the response.
+			// We get a "buffer size too small" error instead of overflow. Retrying for this scenario.
+			// dnsErrBufOccured := false
+			// if dnsErr, ok := err.(*dns.Error); ok {
+			// 	if dnsErr.Error() == dns.ErrBuf.Error() {
+			// 		dnsErrBufOccured = ok
+			// 	}
+			// }
+
+			dnsErrBufOccured := false
+			var perr *dns.Error
+			if errors.As(err, &perr) {
+				if errors.Is(err, dns.ErrBuf) {
+					dnsErrBufOccured = true
+				}
+			}
+
+			if proto == "udp" && ((strings.Contains(err.Error(), "overflow")) || dnsErrBufOccured) {
 				newRet := state.Req.Copy()
+
 				// Clear AD bit in case request had set the AD bit. The empty response is not authenticated.
 				newRet.AuthenticatedData = false
 				newRet.RecursionAvailable = ret.RecursionAvailable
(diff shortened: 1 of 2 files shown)


chrisohaver:
> Copy of description from https://github.com/coredns/coredns/pull/6003

Please just make a reference to the original PR instead of copy-pasting the original PR description.  In the description, please make it clear that you are copying the code of from @phealy's PR, which appears to have stalled.  Please also review all outstanding feedback on the original PR, and list the changes you have made to address that feedback. Thanks!

plugin/pkg/proxy/connect.go
```diff
@@ -121,9 +121,24 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
-			if proto == "udp" {
-				if _, ok := err.(net.Error); !ok {
-					continue
+
+			// If the error is an overflow, we probably have an upstream misbehaving in some way.
+			// (e.g. sending >512 byte UDP responses without an eDNS0 OPT RR).
+			// Instead of returning an error, return an empty response with TC bit set. This will make the
+			// client retry over TCP (if that's supported) or at least receive a clean
+			// error. The connection is still good so we break before the close.
+			if proto == "udp" && strings.Contains(err.Error(), "overflow") {
```
gcs278: Could we include `ErrBuf`? https://github.com/coredns/coredns/pull/6003/files#r1156290153

gcs278: I had a question regarding if we should create types for these errors, searching for the string "overflow" seems a bit dangerous to me: https://github.com/coredns/coredns/pull/6003/files#r1156333895

gcs278 (commented):
thanks for the PR

chrisohaver:
Why is this PR copied from another active  PR?

chrisohaver (changes requested):
Thanks for picking this up!

* In the description, please make it clear that you are copying the code of from @phealy's PR, and the reason (it appears to have stalled) 
* Please also review all outstanding feedback on the original PR, and list the changes you have made to address that feedback. (e.g. I see that you added a unit test, but there was also other feedback that does not appear to be addressed)

plugin/pkg/proxy/connect.go
SriHarsha001: This is done with below change - 

			dnsErrBufOccured := false
			var perr *dns.Error
			if errors.As(err, &perr) {
				if errors.Is(err, dns.ErrBuf) {
					dnsErrBufOccured = true
				}
			}

			if proto == "udp" && ((strings.Contains(err.Error(), "overflow")) || dnsErrBufOccured) {

SriHarsha001: I have create an issue here - https://github.com/miekg/dns/issues/1463 and will have to wait before I can update this code - if proto == "udp" && ((strings.Contains(err.Error(), "overflow")) || dnsErrBufOccured)

plugin/pkg/proxy/connect.go
```diff
@@ -121,12 +122,38 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+
+			if proto == "udp" && ((strings.Contains(err.Error(), "overflow")) || dnsErrBufOccured) {
+				newRet := state.Req.Copy()
+
+				// Clear AD bit in case request had set the AD bit. The empty response is not authenticated.
+				newRet.AuthenticatedData = false
+				newRet.RecursionAvailable = ret.RecursionAvailable
+				newRet.Response = true
+				newRet.Truncated = true
+				ret = newRet
```
chrisohaver: Please see my comments for these lines of code in the original PR.

plugin/pkg/proxy/connect.go
SriHarsha001: Created a PR for this - https://github.com/miekg/dns/pull/1465

chrisohaver: Thanks!  In that PR, maybe put a blank line between `ErrTime`  and `ErrOverflowUnpackingA`, that way indentation doesn't change on existing set of errors, and the diff looks cleaner.

SriHarsha001: @chrisohaver , @gcs278 - While waiting for this PR https://github.com/miekg/dns/pull/1465 to be reviewed, I wanted to check with you if I need to check for all the overflow errors exported here - https://github.com/miekg/dns/pull/1465/files. Because Overflow error due to 512 cache size can occur due to any of the reasons.

For example - 

if errors.Is(err, dns.ErrOverflowHeaderSize)
if errors.Is(err, dns.ErrOverflowPackingUint8)
if errors.Is(err, dns.ErrBuf) .... so on

SriHarsha001: > I had a question regarding if we should create types for these errors, searching for the string "overflow" seems a bit dangerous to me: https://github.com/coredns/coredns/pull/6003/files#r1156333895

@miekg - What would you recommend in this case ? Reference PR - https://github.com/miekg/dns/pull/1465

plugin/pkg/proxy/connect.go
SriHarsha001: Set the AA bit to false in response.

plugin/pkg/proxy/connect.go
SriHarsha001: @miekg - Any suggestions will be appreciated. I will look forward to hearing from you.

plugin/pkg/proxy/connect.go
```diff
@@ -142,6 +143,10 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 
 				// Clear AD bit in case request had set the AD bit. The empty response is not authenticated.
 				newRet.AuthenticatedData = false
+
+				// Clear AA bit in case request had set the AA bit.
```
chrisohaver, line 147: Why do you clear the AA bit?

plugin/pkg/proxy/connect.go
SriHarsha001: @chrisohaver - Do you have any suggestions here ?

plugin/pkg/proxy/connect.go
SriHarsha001: that means that this request isn't an authoritative (i.e. that server owns the final answer) response. Since it's not a complete answer, it's good to keep it set to false. @chrisohaver - Please let me know if you see any concerns here.

plugin/pkg/proxy/connect.go
SriHarsha001: @chrisohaver @miekg - I kindly request your review and suggest on what we can do here to move forward on this change. We need to fix this to move forward with upcoming releases.

chrisohaver: Miek is on indefinite sabbatical from the CoreDNS project, for about a year now.

Looks like https://github.com/miekg/dns/pull/1465 was rejected.

Based on the [comment](https://github.com/miekg/dns/pull/1465#issuecomment-1584116670), I would suggest re-submitting it with following changes:

Don't create new class variables for every possible overflow case.
Create a single public overflow error type that implements error, and raise this error for overflows.
You can make the error text different per overflow case, but the error type would be the same for all.

plugin/pkg/proxy/connect.go
chrisohaver: Cool, no concerns. Please add that description to the existing comment.

chrisohaver: >  it's good to keep it set to false

Actually, are you basing that on common sense, or an RFC requirement?
We may want to follow what is done for actual truncated responses.  

If RFC says AA should be set to false when replying with a truncated response, then we should definitely do that here.
if not, then we need to discuss pros/cons.  E.g. what's is the negative of leaving AA bit as it was.

plugin/pkg/proxy/connect.go
SriHarsha001: Thank you @chrisohaver - I have submitted a PR with the changes here - https://github.com/miekg/dns/pull/1475, waiting for it to be reviewed.

plugin/pkg/proxy/connect.go
SriHarsha001: I couldn't find any specific RFC that directly addresses the handling of the authorization bit (AA) in truncated DNS responses. There is no specific handling or interpretation defined for the authorization bit (AA) in truncated responses

RFC 1035 does not specifically cover the handling of the AA bit in truncated responses
https://tools.ietf.org/html/rfc1035
https://datatracker.ietf.org/doc/html/rfc1035#section-4.2.1

When a response is truncated, the focus is on addressing the truncation itself and ensuring the complete response is obtained. The authorization bit is not directly related to handling truncation but rather serves as an indicator of authority.
The authorization bit is considered meaningful only in non-truncated responses, where it signifies the authoritative nature of the responding server.

AFAIK - It is better to keep it set to false here.

PR #6107, opened 2023-05-22
chore: add ut

yyzxw:

Improve unit test coverage.

### DIFF
--- a/plugin/bind/setup_test.go
+++ b/plugin/bind/setup_test.go
@@ -45,3 +45,39 @@ func TestSetup(t *testing.T) {
 		}
 	}
 }
+
+func Test_isIn(t *testing.T) {
+	type args struct {
+		s    string
+		list []string
+	}
+	tests := []struct {
+		name string
+		args args
+		want bool
+	}{
+		{
+			name: "test not exist",
+			args: args{
+				s:    "test",
+				list: []string{"test1", "test2", "test3"},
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ Corefile
 build/
 release/
 vendor/
+./.idea
```
chrisohaver: please revert this

plugin/bind/setup_test.go
```diff
@@ -45,3 +45,39 @@ func TestSetup(t *testing.T) {
 		}
 	}
 }
+
+func Test_isIn(t *testing.T) {
```
Tantalor93, line 49: this is admirable, but can we rather remove the function `isIn` and rather replace it with `slices.Contains(except, ip)` from [standard library](https://golangtutorial.dev/tips/golang-slice-contains-method/)?

request/request_test.go
```diff
@@ -281,3 +282,123 @@ func TestRequestClear(t *testing.T) {
 		t.Errorf("Expected st.port to be cleared after Clear")
 	}
 }
+
+type fakeResponseWriter struct {
```
Tantalor93, line 286: can we rather add this functionality into `test.responsewriter`?

Tantalor93 (changes requested):

PR #6125, opened 2023-05-28
refactor: use standard library instead of `isIn`

yyzxw:

use standard library  `slices.Contains` instead of `isIn`


Tantalor93 (approved):
lgtm

PR #6137, opened 2023-06-02
plugin/kubernetes: fix headless/endpoint query panics when endpoints are disabled

### LINKED ISSUE #6087: Plugin kubernetes panics if noendpoints is set and it received a request for a headless service
**What happened**:
Plugin `kubernetes` panics in `kubernetes.(*dnsControl).EpIndex` if `noendpoints` is set and it receives a request for a headless service.

It should return `NXDOMAIN` as [documented](https://coredns.io/plugins/kubernetes/).

Start CoreDNS with `kubernetes` plugin enabled and `noendpoints` set.

Create a headless service in kubernetes, e.g. foo in default namespace.

`dig @10.96.0.10 foo.default.svc.cluster.local.`

CoreDNS returns `SERVFAIL` and logs a panic message.

It seems PR [#2128](https://github.com/coredns/coredns/pull/2128) removes a lot of is-nil checks. Any special considerations?

**Environment**:

- the version of CoreDNS: v1.10.1
- Corefile:
```
...
  errors
  kubernetes ... {
    noendpoints
    ttl 30
    fallthrough
  }
...
```
- logs, if applicable:
- Others: k8s v1.26.4

chrisohaver:

Fixes headless and endpoint query panics when endpoints are disabled. This fix always creates listers for all objects, even when an object such as Endpoints is disabled. This way, the listers are never nil, and we don't have to check them every time before accessing them. However, when endpoints is disabled, the _controller_ is not started, so the lister remains completely empty (no items in it and empty indexes).

This PR also fixes tests and fixtures in controller_test.go, which were broken in many ways, rendering the existing benchmark therein somewhat meaningless.  Fixing the existing test was necessary to get the new tests to work.

- API controllers were never started during the test, so the lister stores and indexes were never filled with the test data. The benchmark has been running on an empty set of data, receiving NXDOMAIN for everything. - fixed by starting the controller. 
- Endpoint test data were created as Endpoints, but controller defaults to watch EndpointSlices, so even once the test was fixed to start the controller, it was not seeing the endpoints created - fixed by updating them to be created as EndpointSlices.
- the created service names and endpoint service names were mismatched, creating dangling endpoints with non-existent services - fixed by making all service names consistent 

fixes #6087

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -128,33 +128,37 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
 		object.DefaultProcessor(object.ToService, nil),
 	)
 
+	podLister, podController := object.NewIndexerInformer(
+		&cache.ListWatch{
+			ListFunc:  podListFunc(ctx, dns.client, api.NamespaceAll, dns.selector),
+			WatchFunc: podWatchFunc(ctx, dns.client, api.NamespaceAll, dns.selector),
+		},
+		&api.Pod{},
+		cache.ResourceEventHandlerFuncs{AddFunc: dns.Add, UpdateFunc: dns.Update, DeleteFunc: dns.Delete},
+		cache.Indexers{podIPIndex: podIPIndexFunc},
+		object.DefaultProcessor(object.ToPod, nil),
+	)
+	dns.podLister = podLister
 	if opts.initPodCache {
-		dns.podLister, dns.podController = object.NewIndexerInformer(
-			&cache.ListWatch{
-				ListFunc:  podListFunc(ctx, dns.client, api.NamespaceAll, dns.selector),
-				WatchFunc: podWatchFunc(ctx, dns.client, api.NamespaceAll, dns.selector),
-			},
-			&api.Pod{},
-			cache.ResourceEventHandlerFuncs{AddFunc: dns.Add, UpdateFunc: dns.Update, DeleteFunc: dns.Delete},
-			cache.Indexers{podIPIndex: podIPIndexFunc},
-			object.DefaultProcessor(object.ToPod, nil),
-		)
+		dns.podController = podController
 	}
 
+	epLister, epController := object.NewIndexerInformer(
+		&cache.ListWatch{
+			ListFunc:  endpointSliceListFunc(ctx, dns.client, api.NamespaceAll, dns.selector),
+			WatchFunc: endpointSliceWatchFunc(ctx, dns.client, api.NamespaceAll, dns.selector),
+		},
+		&discovery.EndpointSlice{},
+		cache.ResourceEventHandlerFuncs{AddFunc: dns.Add, UpdateFunc: dns.Update, DeleteFunc: dns.Delete},
+		cache.Indexers{epNameNamespaceIndex: epNameNamespaceIndexFunc, epIPIndex: epIPIndexFunc},
+		object.DefaultProcessor(object.EndpointSliceToEndpoints, dns.EndpointSliceLatencyRecorder()),
+	)
+	dns.epLock.Lock()
+	dns.epLister = epLister
 	if opts.initEndpointsCache {
-		dns.epLock.Lock()
-		dns.epLister, dns.epController = object.NewIndexerInformer(
-			&cache.ListWatch{
-				ListFunc:  endpointSliceListFunc(ctx, dns.client, api.NamespaceAll, dns.selector),
-				WatchFunc: endpointSliceWatchFunc(ctx, dns.client, api.NamespaceAll, dns.selector),
-			},
-			&discovery.EndpointSlice{},
-			cache.ResourceEventHandlerFuncs{AddFunc: dns.Add, UpdateFunc: dns.Update, DeleteFunc: dns.Delete},
-			cache.Indexers{epNameNamespaceIndex: epNameNamespaceIndexFunc, epIPIndex: epIPIndexFunc},
-			object.DefaultProcessor(object.EndpointSliceToEndpoints, dns.EndpointSliceLatencyRecorder()),
-		)
-		dns.epLock.Unlock()
+		dns.epController = epController
 	}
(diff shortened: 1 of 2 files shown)


plugin/kubernetes/controller_test.go
```diff
@@ -25,62 +28,122 @@ func inc(ip net.IP) {
...
 
 	// Add resources
-	generateEndpoints(cidr, client)
-	generateSvcs(cidr, "all", client)
-	m := new(dns.Msg)
-	m.SetQuestion("svc1.testns.svc.cluster.local.", dns.TypeA)
+	_, err := client.CoreV1().Namespaces().Create(ctx, &api.Namespace{ObjectMeta: meta.ObjectMeta{Name: "testns"}}, meta.CreateOptions{})
+	if err != nil {
+		println(err.Error())
+		log.Fatal(err)
```
Tantalor93: why this double logging? I would just keep `log.Fatal(err)` and remove the other

plugin/kubernetes/controller_test.go
```diff
@@ -25,62 +28,122 @@ func inc(ip net.IP) {
...
+func TestEndpointsDisabled(t *testing.T) {
+	ctx := context.Background()
+	k := kubernertesWithFakeClient(ctx, "cluster.local.", "10.0.0.0/30", false, "headless")
+	k.opts.initEndpointsCache = false
+
+	go k.APIConn.Run()
+	for !k.APIConn.HasSynced() {
+		time.Sleep(time.Millisecond)
+	}
+
```
Tantalor93, line 81: `defer k.APIConn.Stop()`
to avoid running wild goroutine after the test is done?

plugin/kubernetes/controller_test.go
```diff
@@ -25,62 +28,122 @@ func inc(ip net.IP) {
...
+func TestEndpointsEnabled(t *testing.T) {
+	ctx := context.Background()
+	k := kubernertesWithFakeClient(ctx, "cluster.local.", "10.0.0.0/30", true, "headless")
+	k.opts.initEndpointsCache = true
+
+	go k.APIConn.Run()
+	for !k.APIConn.HasSynced() {
+		time.Sleep(time.Millisecond)
+	}
+
```
Tantalor93, line 101:
```suggestion
	defer k.APIConn.Stop()
```

plugin/kubernetes/controller_test.go
```diff
@@ -25,62 +28,122 @@ func inc(ip net.IP) {
...
+func BenchmarkController(b *testing.B) {
+	ctx := context.Background()
+	k := kubernertesWithFakeClient(ctx, "cluster.local.", "10.0.0.0/24", true, "all")
+
+	go k.APIConn.Run()
+	for !k.APIConn.HasSynced() {
+		time.Sleep(time.Millisecond)
+	}
+
```
Tantalor93, line 60:
```suggestion
	defer k.APIConn.Stop()

```

plugin/kubernetes/controller_test.go
```diff
@@ -25,62 +28,122 @@ func inc(ip net.IP) {
 	}
 }
 
-func BenchmarkController(b *testing.B) {
+func kubernertesWithFakeClient(ctx context.Context, zone, cidr string, initEndpointsCache bool, svcType string) *Kubernetes {
```
Tantalor93: typo
```suggestion
func kubernetesWithFakeClient(ctx context.Context, zone, cidr string, initEndpointsCache bool, svcType string) *Kubernetes {
```

plugin/kubernetes/controller_test.go
chrisohaver: stray debug.  will remove.

Tantalor93 (approved):

plugin/kubernetes/controller.go
```diff
@@ -128,33 +128,37 @@ func newdnsController(ctx context.Context, kubeClient kubernetes.Interface, opts
...
+		&cache.ListWatch{
+			ListFunc:  endpointSliceListFunc(ctx, dns.client, api.NamespaceAll, dns.selector),
+			WatchFunc: endpointSliceWatchFunc(ctx, dns.client, api.NamespaceAll, dns.selector),
+		},
+		&discovery.EndpointSlice{},
+		cache.ResourceEventHandlerFuncs{AddFunc: dns.Add, UpdateFunc: dns.Update, DeleteFunc: dns.Delete},
+		cache.Indexers{epNameNamespaceIndex: epNameNamespaceIndexFunc, epIPIndex: epIPIndexFunc},
+		object.DefaultProcessor(object.EndpointSliceToEndpoints, dns.EndpointSliceLatencyRecorder()),
+	)
+	dns.epLock.Lock()
```
bradbeam, line 156: this lock seems unnecessary. is there a reason we are doing it that I'm missing?

bradbeam (approved):

plugin/kubernetes/controller.go
chrisohaver: required for endpoints vs endpointslice watch selection.

chrisohaver: > required for endpoints vs endpointslice watch selection.

the dynamic  endpoints/endpointslice detection can eventually go away when we no longer need to support k8s 1.19.0? IIRC? is that EOL yet?  maybe we can drop it now.

chrisohaver: OK k8s 1.19 went EOL over a year ago, so I think we can move the endpoint watch code over to endpointslice only now... but not in this PR.

bradbeam: Was more a question around the lock itself since dns gets created at the start of the function and I dont believe anything else has reference to it at this point.
But good find nonetheless :)

chrisohaver: it is referenced elsewhere, during startup when determining which object to watch.

chrisohaver: ah- I see what you mean. (yes - it could be removed from this spot)

PR #6138, opened 2023-06-03
plugin/tls: respect the path specified by root plugin

mariuskimmina:
Follow up on #5944

### DIFF
--- a/plugin/root/README.md
+++ b/plugin/root/README.md
@@ -2,14 +2,18 @@
 
 ## Name
 
-*root* - simply specifies the root of where to find (zone) files.
+*root* - simply specifies the root of where to find files. 
 
 ## Description
 
 The default root is the current working directory of CoreDNS. The *root* plugin allows you to change
-this. A relative root path is relative to the current working directory.
+this. A relative root path is relative to the current working directory. 
+The *root* directory is currently supported by the following plugins:
 
-This plugin can only be used once per Server Block.
+* file
+* tls
+
+This plugin can only be used once per Server Block. 
 
 ## Syntax
 
@@ -26,5 +30,16 @@ Serve zone data (when the *file* plugin is used) from `/etc/coredns/zones`:
 ~~~ corefile
 . {
     root /etc/coredns/zones
+}
+~~~
+
+When you use the *root* and *tls* plugin together, your cert and key should also be placed in the *root* directory.
+The example below will look for `/config/cert.pem` and `/config/key.pem`
+
+~~~ txt
+tls://example.com:853 {
+    root /config
+    tls cert.pem key.pem
+    whoami
 }
 ~~~
--- a/plugin/root/README.md
+++ b/plugin/root/README.md
@@ -8,7 +8,8 @@
 
 The default root is the current working directory of CoreDNS. The *root* plugin allows you to change
 this. A relative root path is relative to the current working directory. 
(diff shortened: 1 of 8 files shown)


plugin/root/README.md
```diff
@@ -2,14 +2,15 @@
 
 ## Name
 
-*root* - simply specifies the root of where to find (zone) files.
+*root* - simply specifies the root of where to find config files.
```
chrisohaver: I don't think all plugins check root, so we cant claim exactly this yet. Perhaps we need to list plugins that do, or add root check to plugins that don't.

Tantalor93: I would not explicitly list the plugins, rather I would try to make it consistent and fix it for all the plugins

mariuskimmina: So what other files could there be?

What I found so far:
* Kubernetes kubeconfig
* dnssec key file (this might already respect the root path?)
* forward
* sign plugin (also seems to already respect the root path)

What else comes to your mind?

chrisohaver: Check also for indirect file access, where a path or file name is passed to an imported package. Eg cloudns and tsig secrets, and others.

mariuskimmina:
Hi @chrisohaver 

As I have come across some issues when trying to make this work for **all** config files I went back to applying this change only to the tls plugin for now. In the long term it would of course still be better to have it for all config files but I'd like for this PR to not get into a rotten state and we can then migrate over other plugins one by one. 
Please let me know what you think

chrisohaver:
> Hi @chrisohaver
> 
> As I have come across some issues when trying to make this work for **all** config files I went back to applying this change only to the tls plugin for now. In the long term it would of course still be better to have it for all config files but I'd like for this PR to not get into a rotten state and we can then migrate over other plugins one by one. Please let me know what you think

Ok, please update the list of plugins in the README that currently respect the *root* plugin when accessing files.

plugin/root/README.md
```diff
@@ -2,14 +2,18 @@
...
 
-*root* - simply specifies the root of where to find (zone) files.
+*root* - simply specifies the root of where to find files. 
 
 ## Description
 
 The default root is the current working directory of CoreDNS. The *root* plugin allows you to change
-this. A relative root path is relative to the current working directory.
+this. A relative root path is relative to the current working directory. 
+The *root* directory is currently supported by the following plugins:
```
chrisohaver:
```suggestion

**NOTE: The *root* directory is NOT currently supported by all plugins.** 

Currently the following plugins respect the *root* plugin configuration:
```

Im thinking we should add a Bugs section (like some other plugins), and add this note there as well.

chrisohaver (changes requested):

plugin/root/README.md
mariuskimmina: Adjusted it, I have also added the "Bugs" section.

chrisohaver (approved):

chrisohaver:
Looks like we aren't going to get a review from plugin owners. Sorry for the delay. Please resolve the conflicts and I will merge it.

mariuskimmina:
> Looks like we aren't going to get a review from plugin owners. Sorry for the delay. Please resolve the conflicts and I will merge it.

Done

greenpau:
@mariuskimmina , well done! 🎉 Happy to see you contributing! 👍 
cc: @yongtang

PR #6139, opened 2023-06-03
replace reviewdog fully by golangci-lint

Tantalor93:
This PR replaces the current [Reviewdog functionality](https://github.com/coredns/coredns/blob/master/.github/workflows/reviewdog.yml) fully with [golangci-lint](https://github.com/coredns/coredns/blob/master/.golangci.yml) linter. Aim of this PR is to consolidate CoreDNS linting and unify it under single technology (golangci-lint), which is better configurable, maintainable. 

`whitespace` check is already enabled for `golangci-lint` we need to additionaly enable `gofmt`

none

none

no


SuperQ (approved):
Nice

PR #6148, opened 2023-06-08
Bind on hostname

rsclarke:

The bind plugin currently supports either literal IP addresses or interface names to open a listener on.  This pull request adds the ability to add a listener for an IP address (or IP addresses) resolved through a given hostname.

N/A

Updated bind plugin documentation to indicate hostname support and usage.

Order is preserved in resolution of the `bind` and `except` sub-directive with the hostname being resolved last.  Any interface/hostname collision (i.e interface `lo` and a hostname `lo`) will maintain backwards compatibility in that the interface takes precedence.

Users specifying an invalid hostname to bind for the system receive the same error message as if given an incorrect IP address.

### DIFF
--- a/plugin/bind/README.md
+++ b/plugin/bind/README.md
@@ -7,35 +7,36 @@
 ## Description
 
 Normally, the listener binds to the wildcard host. However, you may want the listener to bind to
-another IP instead.
+another IP address instead.
 
-If several addresses are provided, a listener will be open on each of the IP provided.
+To specify the IP address of the listener the `bind` directive accepts an IP address, interface name or hostname.  Interface names and hostnames will be resolved to IP addresses. A listener will be open on each provided and resolved IP address.
 
-Each address has to be an IP or name of one of the interfaces of the host. Bind by interface name, binds to the IPs on that interface at the time of startup or reload (reload will happen with a SIGHUP or if the config file changes).
+Bind by interface name, binds to the IP addresses on that interface at the time of startup or reload (reload will happen with a SIGHUP or if the config file changes).
 
-If the given argument is an interface name, and that interface has serveral IP addresses, CoreDNS will listen on all of the interface IP addresses (including IPv4 and IPv6), except for IPv6 link-local addresses on that interface.
+Bind by hostname, binds to the IP addresses resolved by the default resolver at the time of startup or reload.
+
+If the given argument is an interface name or hostname, and that interface or hostname has serveral IP addresses, CoreDNS will listen on all of the IP addresses (including IPv4 and IPv6), except for IPv6 link-local addresses on an interface.
 
 ## Syntax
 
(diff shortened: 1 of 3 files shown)


chrisohaver:
Interesting. Can you provide a practical use case example?

rsclarke:
> Interesting. Can you provide a practical use case example?

In my case the `fly-global-services`[^1] address is not known when configuring coredns and should be resolved at runtime.  From what I understand it is not the same across multiple machines either when running multiple instances but the hostname is.

[^1]: https://fly.io/docs/app-guides/udp-and-tcp/#the-fly-global-services-address

chrisohaver:
In your case I presume it's defined in /etc/hosts? If not, then thats super kludgey to ask some other server what the local IP is.

Is the local IP also exported to an enviroment variable?  If so, you could bind to the ENV variable.

chrisohaver:
not an obvious coredns feature: you can insert environment variables into the Corefile. https://coredns.io/manual/configuration/#environment-variables

so, If fly.io service makes the IP address available in an env variable, then you could bind to it with existing functionality.

rsclarke:
Yes it is defined in `/etc/hosts` provided by the Fly platform.

Unfortunately it is not exposed as an environment variable, though I understand I could introduce an entry point script to resolve this and expose it to coredns using environment variable substitution in the Corefile as you suggested.  I'd rather favour coredns being able to do this than introduce additional utilities into the container image.

chrisohaver:
> Unfortunately it is not exposed as an environment variable

That is unfortunate.

In general, for a DNS service _specifically_ - it feels "wrong" that a the service would need to bind to a hostname because the IP is not well known.  How would clients learn of the IP when they make a DNS lookup?

In your use-case, I presume clients would make queries against some public IP, and then the requests get routed (somehow) to the correct service IP.

I understand the need in your specific use case, but I'm hesitant to support this as a general feature. Perhaps we can look the host up directly in /etc/hosts instead of going through the system resolver?

rsclarke:
> I presume clients would make queries against some public IP, and then the requests get routed

Yes, Fly operates a proxy performing NAT.  My scenario is operating coredns as an authoritative server, the public IP addresses are static and would be listed as the NS records for the domains.

I'd be hesitant to introduce another dependency and or additional logic just to parse and lookup within `/etc/hosts` or equivalent on other operating systems when this may be achieved through `net.LookupHost`.

Rather than my specific use case, I liken this to specifying an interface, the address or addresses are not known until runtime.  The discovery of the DNS server from a client perspective is outside the scope, be it advertised through DHCP or otherwise.

I also think this brings familiarity when specifying a listener address, just like one might specify for the health or metrics plugins `localhost:8080` or `127.0.0.1:9091`, it allows the `bind` directive to support the same, excluding the port.

PR #6162, opened 2023-06-15
plugin/kubernetes: filter ExternalName service queries for subdomains of subdomains

### LINKED ISSUE #6161: plugin/kubernetes returns response for subdomain of subdomain of ExternalName Service
**What happened**:
A query for any.website.com from a kubernetes namespace which includes an ExternalName Service with the name of `com` will return the ExternalName record. The same would happen for a query from any.website.io if there is an ExternalName Service named `io`.

The plugin should respond with NXDOMAIN so the request can be processed by other plugins.

```bash
kubectl run --image=kennethreitz/httpbin --port=80 --expose httpbin
kubectl create service externalname --external-name=httpbin.default.svc.cluster.local com
kubectl get svc
sleep 10
kubectl run -it --rm --restart=Never --image=tutum/dnsutils temp  -- /bin/bash -c "nslookup www.google.com"
```

services
```bash
NAME         TYPE           CLUSTER-IP     EXTERNAL-IP                         PORT(S)   AGE
com          ExternalName   <none>         httpbin.default.svc.cluster.local   <none>    0s
httpbin      ClusterIP      10.96.13.146   <none>                              80/TCP    0s
...
```
coredns query response
```bash
:$kubectl run -it --rm --restart=Never --image=tutum/dnsutils temp  -- /bin/bash -c "nslookup www.google.com"
Server:		10.96.0.10
Address:	10.96.0.10#53

www.google.com.default.svc.cluster.local	canonical name = httpbin.default.svc.cluster.local.
Name:	httpbin.default.svc.cluster.local
Address: 10.96.13.146
```

cleanup
```bash
kubectl delete svc httpbin com
kubectl delete pod httpbin
```

This is happening because the default search suffix for the full namespace `default.svc.cluster.local` is added onto the query sent to coredns resulting in `www.google.com.default.svc.cluster.local`. When this is [parsed](https://github.com/coredns/coredns/blob/master/plugin/kubernetes/parse.go#L68) it assumes that the query is for an SRV record and assigns the labels www and google to protocol and port. Then during [matching](https://github.com/coredns/coredns/blob/master/plugin/kubernetes/kubernetes.go#L449) of the services from discovery those fields are ignored.

I believe the fix here is very similar to #5733 where ExternalName services need to be skipped if the port or protocol is set.

If you agree I can add a PR for this.

Any workarounds you could suggest?
Thank you!

**Environment**:

coredns v1.10.1
standard config from minikube

emaildanwilson:
This prevents queries for `*.*.service.namespace.svc.cluster.local` from being resolved when the service is type ExternalName. The `*` is not for a wildcard DNS query here, just to indicate that any valid dns label can be used in the query.

For example, a query for `www.google.com.default.svc.cluster.local` will currently match and return the result of an ExternalName service named `com` in the `default` namespace.

This change prevents this from happening.

fixes #6161

none

no, it fixes queries that should be failing but are returning an invalid response.

### DIFF
--- a/plugin/kubernetes/handler_test.go
+++ b/plugin/kubernetes/handler_test.go
@@ -408,6 +408,14 @@ var dnsTestCases = []kubeTestCase{
 			test.SOA("cluster.local.	5	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1499347823 7200 1800 86400 5"),
 		},
 	}},
+	// A query for a subdomain of a subdomain of an external service should not resolve to the external service
+	{Case: test.Case{
+		Qname: "subdomain.subdomain.external.testns.svc.cluster.local.", Qtype: dns.TypeCNAME,
+		Rcode: dns.RcodeNameError,
+		Ns: []dns.RR{
+			test.SOA("cluster.local.	5	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1499347823 7200 1800 86400 5"),
+		},
+	}},
 }
 
 func TestServeDNS(t *testing.T) {
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -445,8 +445,8 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 
 		// External service
 		if svc.Type == api.ServiceTypeExternalName {
-			//External services cannot have endpoints, so skip this service if an endpoint is present in the request
-			if r.endpoint != "" {
+			// External services do not have endpoints, nor can we accept port/protocol pseudo subdomains in an SRV query, so skip this service if endpoint, port, or protocol is non-empty in the request
+			if r.endpoint != "" || r.port != "" || r.protocol != "" {
 				continue
 			}
 			s := msg.Service{Key: strings.Join([]string{zonePath, Svc, svc.Namespace, svc.Name}, "/"), Host: svc.ExternalName, TTL: k.ttl}


plugin/kubernetes/handler_test.go
```diff
@@ -408,6 +408,14 @@ var dnsTestCases = []kubeTestCase{
 			test.SOA("cluster.local.	5	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1499347823 7200 1800 86400 5"),
 		},
 	}},
+	// A query for a double subdomain of an external service should not resolve to the external service
```
chrisohaver:
```suggestion
	// A query for subdomains of an external service should not resolve to the external service
```

plugin/kubernetes/handler_test.go
```diff
@@ -408,6 +408,14 @@ var dnsTestCases = []kubeTestCase{
...
+	{Case: test.Case{
+		Qname: "double.endpoint.external.testns.svc.cluster.local.", Qtype: dns.TypeCNAME,
```
chrisohaver: I think `double.endpoint.external.testns.svc.cluster.local.` should already be invalid when it is parsed ...

type = "svc"
namespace = "testns"
service name = "external"
endpoint name = "endpoint"
protocol = "double"  ... which is invalid because protocol has to begin with a "_"
port = not specified (I don't recall if having only one of port/protocol is valid)

chrisohaver: actually parse lets invalid port(service) and protocol pseudo-subdomains through.
It should probably block them there.

Do we / should we permit SRV queries on external services?  Does DNS spec permit this? I know that SRVs cannot have _targets_ that are CNAMEs, but this is not that.
e.g. for an external service external-svc.default.svc.cluster.local. that points to my.service.com., should a SRV query for external-svc.default.svc.cluster.local. return CNAME and SRV record? 
should _http._tcp.external-svc.default.svc.cluster.local. be allowed? (dubious if we think of the pseudo-segments as real subdomains...)

emaildanwilson: I started with only setting the port/protocol pseudo segments if they started with an underscore but then went with this other approach because it could also effect headless service lookups. I can revisit this approach if it's preferred. hopefully some others can chime in on the SRV lookup questions.

chrisohaver: OK, the fix is OK as is.  Though I think the use of the terms `double` and `endpoint` is confusing here.  They are really just subdomains.  In fact, if the `endpoint` segment was actually being parsed as an "endpoint", the existing check would have caught it.

```suggestion
		Qname: "subdomain1.subdomain2.external.testns.svc.cluster.local.", Qtype: dns.TypeCNAME,
```

plugin/kubernetes/kubernetes.go
```diff
@@ -446,7 +446,7 @@ func (k *Kubernetes) findServices(r recordRequest, zone string) (services []msg.
 		// External service
 		if svc.Type == api.ServiceTypeExternalName {
 			//External services cannot have endpoints, so skip this service if an endpoint is present in the request
```
chrisohaver:
```suggestion
			// External services do not have endpoints, nor can we accept port/protocol pseudo subdomains in an SRV query, so skip this service if endpoint, port, or protocol is non-empty in the request
```

plugin/kubernetes/handler_test.go
emaildanwilson: sounds good. I'll make the updates to the names so it's less confusing.

chrisohaver:
FWIW, the title of the PR mentions "double wildcard", and the example query contains `*`'s, which used to have a special nonstandard behavior in CoreDNS's _kubernetes_ plugin.

I thought I should clarify that CoreDNS no longer supports "wildcard" queries like this.  It treats a `*` segment as a literal `*`, per standard DNS behavior since 1.9.0.

It doesnt affect the validity of the issue or fix.  The issue exists for any number of subdomains under the external service domain > 1 .

plugin/kubernetes/handler_test.go
```diff
@@ -408,6 +408,14 @@ var dnsTestCases = []kubeTestCase{
 			test.SOA("cluster.local.	5	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1499347823 7200 1800 86400 5"),
 		},
 	}},
+	// A query for a subdomain of a subdomain of an external service should not resolve to the external service
+	{Case: test.Case{
+		Qname: "subdoman.subdomain.external.testns.svc.cluster.local.", Qtype: dns.TypeCNAME,
```
chrisohaver: I know it doesn't matter,  but this will bother someone, might as well fix now.

```suggestion
		Qname: "subdomain.subdomain.external.testns.svc.cluster.local.", Qtype: dns.TypeCNAME,
```

emaildanwilson:
> FWIW, the title of the PR mentions "double wildcard", and the example query contains `*`'s, which used to have a special nonstandard behavior in CoreDNS's _kubernetes_ plugin.
> 
> I thought I should clarify that CoreDNS no longer supports "wildcard" queries like this. It treats a `*` segment as a literal `*`, per standard DNS behavior since 1.9.0.
> 
> It doesnt affect the validity of the issue or fix. The issue exists for any number of subdomains under the external service domain > 1 .

Thank you! I updated the title and added clarity to the description so that this doesn't get mixed up with wildcard dns query functionality.

plugin/kubernetes/handler_test.go
emaildanwilson: shoot. will fix.

PR #6164, opened 2023-06-16
test: use `t.TempDir` to create temporary test directory

Juneezee:

This pull request replaces `os.MkdirTemp` with `t.TempDir`. We can use the `t.TempDir` function from the `testing` package to create temporary directory. The directory created by `t.TempDir` is automatically removed when the test and all its subtests complete. 

Reference: https://pkg.go.dev/testing#T.TempDir

```go
func TestFoo(t *testing.T) {
	// before
	dir, err := os.MkdirTemp("", "")
	if err != nil {
		t.Fatal(err)
	}
	defer os.RemoveAll(dir)

	// now
	dir := t.TempDir()
}
```

No.

### DIFF
--- a/plugin/auto/walk_test.go
+++ b/plugin/auto/walk_test.go
@@ -18,14 +18,10 @@ www IN A 127.0.0.1
 `
 
 func TestWalk(t *testing.T) {
-	tempdir, err := createFiles()
+	tempdir, err := createFiles(t)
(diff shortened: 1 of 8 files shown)


SuperQ (approved):
LGTM

Tantalor93 (approved):

PR #6165, opened 2023-06-17
Update workflow versioning

SuperQ:
* Use major version pinning for actions to reduce dependeabot churn.
* Use tilde matching for Go versions to use the latest patch release.
* Checkout before installing Go to fix go module caching.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -27,15 +27,15 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+        uses: actions/checkout@v3
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@cdcdbb579706841c47f7063dda365e292e5cad7a
+        uses: github/codeql-action/init@v2
         with:
           languages: ${{ matrix.language }}
 
       - name: Autobuild
-        uses: github/codeql-action/autobuild@cdcdbb579706841c47f7063dda365e292e5cad7a
+        uses: github/codeql-action/autobuild@v2
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@cdcdbb579706841c47f7063dda365e292e5cad7a
+        uses: github/codeql-action/analyze@v2
--- a/.github/workflows/go.coverage.yml
+++ b/.github/workflows/go.coverage.yml
@@ -8,14 +8,14 @@ jobs:
     name: Coverage
     runs-on: ubuntu-latest
     steps:
-      - name: Install Go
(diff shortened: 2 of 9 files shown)


SuperQ:
Moved the checkout action to first so that caching in setup-go works.

```
Warning: Restore cache failed: Dependencies file is not found in /home/runner/work/coredns/coredns. Supported file pattern: go.sum
```

.github/workflows/codeql-analysis.yml
```diff
@@ -27,7 +27,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+        uses: actions/checkout@v3
 
       - name: Initialize CodeQL
         uses: github/codeql-action/init@cdcdbb579706841c47f7063dda365e292e5cad7a
```
Tantalor93: what about these codeql-action versions?

SuperQ: Sure

.github/workflows/go.coverage.yml
```diff
@@ -8,15 +8,15 @@ jobs:
     name: Coverage
     runs-on: ubuntu-latest
     steps:
+      - name: Check out code
```
Tantalor93, line 11: can we also pin `codecov/codecov-action`

.github/workflows/scorecards.yml
```diff
@@ -23,7 +23,7 @@ jobs:
 
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+        uses: actions/checkout@v3
```
Tantalor93, line 26: can we also pin version actions/upload-artifact and ossf/scoreboards action ?

.github/workflows/yamllint.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: 'Checkout'
-        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+        uses: actions/checkout@v3
```
Tantalor93, line 11: can we pin yamllint action also?

Tantalor93 (commented):
nice! this looks very useful, i would just make sure we do this consistently for all workflows and their used actions

bradbeam (approved):

.github/workflows/yamllint.yml
SuperQ: That one doesn't seem to promote this kind of pinning, can try tho.

SuperQ: Nope, this one didn't work.

Tantalor93 (approved):

SuperQ:
Now, hopefully dependabot doesn't just revert all these version pins. :grin:

PR #6175, opened 2023-06-19
plugin/hosts: add Readiness implementation, add no_file option

xervon:

This change:
1. adds a `no_file` option to the `hosts` plugin that prevents loading of any hosts file and instead relies entirely on inline hosts definitions
2. adds an implementation for the `ready.Readiness` interface that signals ready once both the `hosts`-file and all inline definitions have been /successfully/ loaded. This means if the hosts file does not exist (yet), it will only signal ready once it does and has been loaded. If `no_file` is set that  part of the readiness check is skipped.

This probably should be two PRs though I don't really know how to split them properly, because I don't know how to create dependant PRs on Github.

None (that I know of)

The plugin README has been updated. Is anything else needed?

A Coredns setup that uses the `hosts` and `ready` plugins but can't guarantee the existence of the `hosts` file would've signaled success on the `ready` endpoint while emitting a warning about the missing `hosts` file.
With this change the `ready` endpoint will never return a `200 OK` response. This could cause readiness checks to fail.

One way to fix this could be to disable the readiness check by default and adding a new option `ready_check_file` that enables the readiness check on the hosts file.

### DIFF
--- a/plugin/hosts/README.md
+++ b/plugin/hosts/README.md
@@ -47,6 +47,7 @@ entries) and cannot be created manually.
 ~~~
 hosts [FILE [ZONES...]] {
     [INLINE]
+    no_file
     ttl SECONDS
     no_reverse
     reload DURATION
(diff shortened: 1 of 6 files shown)


xervon:
On further thought this does actually introduce a backwards incompatible change in it's current form:

A Coredns setup that uses the `hosts` and `ready` plugins but can't guarantee the existence of the `hosts` file would've signaled success on the `ready` endpoint while emitting a warning about the missing `hosts` file.
With this change the `ready` endpoint will never return a `200 OK` response. This could cause readiness checks to fail.

One way to fix this could be to disable the readiness check by default and adding a new option `ready_check_file` that enables the readiness check on the hosts file.

I've also updated the original post to reflect this.

Tantalor93:
> adds a no_file option to the hosts plugin that prevents loading of any hosts file and instead relies entirely on inline hosts definitions

I think that if a user wants to use only inlined definitions in Corefile, he can use [template](https://coredns.io/plugins/template/) plugin instead of using **hosts** plugin. The hosts plugin source of truth should be an existing file in hosts format. So I am not fan of adding `no_file` option

> A Coredns setup that uses the hosts and ready plugins but can't guarantee the existence of the hosts file would've signaled success on the ready endpoint while emitting a warning about the missing hosts file.
With this change the ready endpoint will never return a 200 OK response. This could cause readiness checks to fail.

yes, that looks like backward incompatible change. I would suggest an option `fail_readiness_on_missing_file` with default value false

Issue #6160, opened 2023-06-13
forward plugin metrics renamed


Tantalor93:
As part of https://github.com/coredns/coredns/pull/5951 most of the forward plugin metrics were moved under different namespace (`forward` =>`proxy`), which caused that those metrics are renamed. This change is not backward compatible and is quite dangerous without proper announcement

also the documentation is no longer true https://github.com/coredns/coredns/tree/master/plugin/forward#metrics

for example `coredns_forward_healthcheck_failures_total` => `coredns_proxy_healthcheck_failures_total`

I think it would be safest to revert the mentioned PR, WDYT @chrisohaver

chrisohaver:
Why not fix the namespace. 

(Ie change the namespace back to forward instead of proxy, instead of reverting the entire PR)

chrisohaver:
@patdowney, FYI

Tantalor93:
> Why not fix the namespace.
> 
> (Ie change the namespace back to forward instead of proxy, instead of reverting the entire PR)

that would work as well 🙂

patdowney:
Thanks @chrisohaver - sorry @Tantalor93 - let me take a look

patdowney:
There's some conversation going on on #6178 that is leaning towards removing the duplicate metrics and documenting the changes as breaking. The `README.md` is being updated appropriately. Might need to rename the PR!

PR #6135, opened 2023-06-01
1.11.0 release prep

chrisohaver:
Adds 1.11.0 release notes

PR #6178, opened 2023-06-23
Fix forward metrics for backwards compatibility

patdowney:
This pull request is necessary to fix monitoring for users of the commonly used `forward` plugin, that was broken by a PR I previously worked on to pull the `Proxy` functionality out of `forward` into somewhere more accessible.

This pull request adds the ability for consumers of `plugin/pkg/proxy`, like `forward` to specify an isolated metrics collector for the plugin.

#6160 

Fixing this for backwards compatibility will reconcile the documentation that currently refers to `coredns_forward_` metrics.

Hopefully not, it's fixing something I broker earlier!


SuperQ (changes requested):
I think the big issue here is that, in my opinion, metrics should not escape the plugin.

Rather than allow new metrics to be created with new names, we should add a third label to the metrics and require `NewProxy` require a name.

So something like:
```go
func NewProxy(addr, trans, proxyName string)
```

The the metrics would have names like:
```go
	requestCount = promauto.NewCounterVec(prometheus.CounterOpts{
		Namespace: plugin.Namespace,
		Subsystem: "proxy",
		Name:      "requests_total",
		Help:      "Counter of requests made per upstream.",
	}, []string{"proxy_name", "to"})
```

This way the metrics in one plugin don't leak into another.

patdowney:
Thanks @SuperQ - I'm happy to make that change, however it won't fix the backwards compatibility issue that's been raised/I broke initially, as that's dependent on the metric names having the coredns_forward_ prefix.

Do you have any suggestions how we handle that aspect? Or is it effectively a breaking change and I/we need to treat it as such?

SuperQ:
Metric names are allowed to change, we don't have specific backwards compatibility guarantees around them.

We do need to document the change as breaking.

patdowney:
Perhaps another alternative is to not record the metrics in proxy.Proxy but instead move that into the calling plugin?

Or maybe both for the case of forward?

SuperQ:
Having metrics in the calling plugin is also an option. Typically I like to have each plugin responsible for its own metrics.

Anything proxy specific can stay in the proxy plugin.

patdowney:
@SuperQ - I've added `proxy_name` label where applicable. I've also added duplicate metrics in `forward` to keep some backwards compatibility:
* `coredns_forward_requests_total{to}`,
* `coredns_forward_responses_total{to, rcode}`, 
* `coredns_forward_request_duration_seconds{to, rcode}`
 
However  there are some that aren't as easy to do a similar thing for:
* `coredns_forward_healthcheck_failures_total{to, rcode}`
* `coredns_forward_conn_cache_hits_total{to, proto}`
* `coredns_forward_conn_cache_misses_total{to, proto}`

So given that we can't easily keep backwards compatibility for all of the metrics, I'm wondering if, given the cost of clutter/maintenance, if it's worth keeping backwards compatibility for any of them?

I've updated the `forward` `README.md` with the new metric names. Where should I  document the breaking change?

plugin/forward/metrics.go
```diff
@@ -1,14 +1,38 @@
...
 	"github.com/coredns/coredns/plugin"
 
+	"github.com/miekg/dns"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
 // Variables declared for monitoring.
 var (
+	RequestCount = promauto.NewCounterVec(prometheus.CounterOpts{
```
SuperQ: Please make all of these metric vars unexported so they can't be used outside of the package.

```suggestion
	requestCount = promauto.NewCounterVec(prometheus.CounterOpts{
```

plugin/pkg/proxy/connect.go
```diff
@@ -152,9 +152,9 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 		rc = strconv.Itoa(ret.Rcode)
 	}
 
-	RequestCount.WithLabelValues(p.addr).Add(1)
-	RcodeCount.WithLabelValues(rc, p.addr).Add(1)
-	RequestDuration.WithLabelValues(p.addr, rc).Observe(time.Since(start).Seconds())
+	RequestCount.WithLabelValues(p.proxyName, p.addr).Add(1)
+	RcodeCount.WithLabelValues(p.proxyName, p.addr, rc).Add(1)
+	RequestDuration.WithLabelValues(p.proxyName, p.addr, rc).Observe(time.Since(start).Seconds())
```
SuperQ: Same with the proxy metrics. While we're here let's fix them to be unexported.

```suggestion
	requestCount.WithLabelValues(p.proxyName, p.addr).Add(1)
	rcodeCount.WithLabelValues(p.proxyName, p.addr, rc).Add(1)
	requestDuration.WithLabelValues(p.proxyName, p.addr, rc).Observe(time.Since(start).Seconds())
```

plugin/pkg/proxy/connect.go
```diff
@@ -152,9 +152,9 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 		rc = strconv.Itoa(ret.Rcode)
 	}
 
-	RequestCount.WithLabelValues(p.addr).Add(1)
-	RcodeCount.WithLabelValues(rc, p.addr).Add(1)
-	RequestDuration.WithLabelValues(p.addr, rc).Observe(time.Since(start).Seconds())
+	RequestCount.WithLabelValues(p.proxyName, p.addr).Add(1)
+	RcodeCount.WithLabelValues(p.proxyName, p.addr, rc).Add(1)
```
SuperQ: The `RequestCount` and `RcodeCount` metrics seem to be duplicates of each other, as well as duplicating the `_count` value of the `RequestDuration` metric.

Let's drop them as they are just extra cardinality data.

plugin/forward/metrics.go
```diff
@@ -22,3 +46,17 @@ var (
...
+func recordReqMetrics(addr string, start time.Time, res *dns.Msg) {
+	RequestCount.WithLabelValues(addr).Add(1)
+
+	if res != nil {
+		// Record metrics
+		rc, ok := dns.RcodeToString[res.Rcode]
+		if !ok {
+			rc = strconv.Itoa(res.Rcode)
+		}
+		RcodeCount.WithLabelValues(addr, rc).Add(1)
```
SuperQ: Similar here, we don't actually need `RcodeCount` since we get the same values from the `RequestDuration` metric `_count` data.

We can remove this and mark it as a breaking change.

SuperQ (changes requested):
Thanks, that is a good improvement.

Some other comments inline.

SuperQ:
> So given that we can't easily keep backwards compatibility for all of the metrics, I'm wondering if, given the cost of clutter/maintenance, if it's worth keeping backwards compatibility for any of them?

When the underlying code changes such that metrics lose their meaning it's fine to make breaking changes to metrics.

plugin/forward/metrics.go
patdowney: I've removed the duplicate metrics from `forward` completely.

plugin/pkg/proxy/connect.go
patdowney: Have removed these and updated the `README.md` with pointers to derive the value from the `_count` value of the `requestDuration` metric.

plugin/pkg/proxy/connect.go
patdowney: Have unexported all metrics defined within the `proxy` plugin.

plugin/forward/metrics.go
patdowney: Have unexported all metrics defined within the `forward` plugin.

SuperQ:
Does it make sense to keep the `coredns_forward_requests_total{to}` metric since it records requests where the proxy returned nil?

patdowney:
Good eye! The original behaviour of the metrics recording looked like this:
```
	RequestCount.WithLabelValues(p.addr).Add(1)
	RcodeCount.WithLabelValues(rc, p.addr).Add(1)
	RequestDuration.WithLabelValues(p.addr, rc).Observe(time.Since(start).Seconds())
```
The behaviour of recording `RequestCount` if the `dns.Msg` was `nil` was introduced by me during this PR (in commit 5bea543f) because I wasn't paying enough attention to the original behaviour (there's a check for `nil` earlier on in the original function that I missed).

I'm not sure about adding it back in, at least with the original name, as the metric value won't be what people monitoring it are expecting.

SuperQ:
Makes sense. The key is that users will want to have a metric that captures all requests regardless of valid response. This allows them to detect lost responses.

If the metrics capture this, then we're all good.

SuperQ (approved):
Looks good, thanks for the deprecated metric translation documentation.

zouyee (approved):

Tantalor93 (approved):
LGTM, but we should make sure this is mentioned in release notes for 1.11.0 (https://github.com/coredns/coredns/pull/6135)

Issue #5583, opened 2022-08-24
Support For DoQ (DNS over QUIC)


FruityWelsh:
**What would you like to be added**:
DoQ (DNS over QUIC) query servering support. 
**Why is this needed**:
DoQ would allow an easier way to take advantage of the speed improvements of QUIC vs TLS including the included streaming (similar to http/2) and encryption (taking over the role of tls). This would be similar to DoH/3 support except it should be simpler to implement right now, uses a different standard port, and less data transmitted (which is a benifit for some for privacy reasons).

Relevant RFC from IETF: https://datatracker.ietf.org/doc/rfc9250/
Reference implentation: https://github.com/huitema/dnsoquic

tanujd11:
@FruityWelsh @chrisohaver I want to contribute to this. Do you have any pointers for me before I start? One of the issue I can see at the moment is that coredns fork of caddy doesn't have the HTTP3 support that is there in the upstream at the moment. How to workaround that situation?

Amila-Rukshan:
Here is a WIP implementation of DoQ -  https://github.com/DNS-OARC/golang-dns-server-doq

jaehnri:
For documentation purposes, I plan to work on this feature during GSoC '23. 🤞 
The proposal was discussed here: https://github.com/coredns/coredns/discussions/6000.

prabin-acharya:
Hi @jaehnri, are you working on this?
If not, I would love to contribute.

jaehnri:
Yes, I'm working on it.

antonio-pedro99:
Hi @FruityWelsh, I am applying for LFX Mentorship and I saw that this project is also participating. As I saw @jaehnri is currently working in this issue, I would like to get some confirmation from you so that I can apply.
cc: @yongtang

houyuwushang:
This server-side request was implemented by #6182, merged in cc7a364. Current master supports quic:// server blocks, and I revalidated the current implementation with DoQ unit and race tests plus integration tests covering real queries, protocol errors, and stream limits.

#6245 is a separate feature direction: forwarding queries from CoreDNS to a DoQ upstream. It therefore does not need to keep this server-listener issue open and should be tracked independently if that work resumes.

Closing as completed by #6182.

PR #6182, opened 2023-06-29
[RFC-9250]: Add QUIC server support 

jaehnri:
This pull request adds support for listening on QUIC.

References https://github.com/coredns/coredns/issues/5583

The main README was updated to contain examples of a CoreDNS DoQ server. More documentation may be necessary.

No.


core/dnsserver/server_quic.go
```diff
@@ -0,0 +1,287 @@
...
+	if out != "" {
+		fmt.Print(out)
+	}
+}
+
+// Stop stops the server. It blocks until the server is totally stopped.
+func (s *ServerQUIC) Stop() error { return nil }
+
+// Shutdown stops the server (non gracefully).
+func (s *ServerQUIC) Shutdown() error { return nil }
```
Tantalor93: shoudn't you close the listener here?
```suggestion
func (s *ServerQUIC) Shutdown() error { 
	if s.quicListener != nil {
		return s.quicListener.Close()
	}
	return nil
}
```

core/dnsserver/server_quic.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	out := startUpZones(transport.QUIC+"://", s.Addr, s.zones)
+	if out != "" {
+		fmt.Print(out)
+	}
+}
+
+// Stop stops the server. It blocks until the server is totally stopped.
+func (s *ServerQUIC) Stop() error { return nil }
```
Tantalor93: Stop the quic server here?
```suggestion
func (s *ServerQUIC) Stop() error {
	return s.Shutdown() 
}
```

core/dnsserver/server_quic.go
jaehnri: Did it, thanks!

core/dnsserver/server_quic.go
jaehnri: I opted to remove this method.

chrisohaver (commented):
Thanks! Looks good so far!

I think it would be good to add some integration style tests (in `/test`), that perform some test DoQ queries against an instance of CoreDNS configured with DoQ.

core/dnsserver/server_quic.go
```diff
@@ -0,0 +1,293 @@
...
+		return
+	}
+
+	req := &dns.Msg{}
+
+	// Drafts of the RFC9250 did not require the 2-byte prefixed message length.
+	// Thus, we are only supporting the official version (DoQ v1).
+	err = req.Unpack(buf[2:])
+	if err != nil {
+		clog.Debug("unpacking quic packet: %s", err)
```
Tantalor93:
```suggestion
		clog.Debugf("unpacking quic packet: %s", err)
```

core/dnsserver/server_quic.go
```diff
@@ -0,0 +1,293 @@
...
+
+// Serve implements caddy.TCPServer interface.
+func (s *ServerQUIC) Serve(l net.Listener) error { return nil }
+
+// Listen implements caddy.TCPServer interface.
+func (s *ServerQUIC) Listen() (net.Listener, error) { return nil, nil }
+
+// closeQUICConn quietly closes the QUIC connection.
+func closeQUICConn(conn quic.Connection, code quic.ApplicationErrorCode) {
+	clog.Debug("closing quic conn %s with code %d", conn.LocalAddr(), code)
```
Tantalor93:
```suggestion
	clog.Debugf("closing quic conn %s with code %d", conn.LocalAddr(), code)
```

core/dnsserver/server_quic.go
```diff
@@ -0,0 +1,293 @@
...
+
+	err := conn.CloseWithError(code, "")
+	if err != nil {
+		clog.Debug("closing quic connection with code %d: %s", code, err)
```
Tantalor93:
```suggestion
		clog.Debugf("closing quic connection with code %d: %s", code, err)
```

core/dnsserver/server_quic.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+// ListenPacket implements caddy.UDPServer interface.
+func (s *ServerQUIC) ListenPacket() (net.PacketConn, error) {
+	p, err := reuseport.ListenPacket("udp", s.Addr[len(transport.QUIC+"://"):])
+	if err != nil {
+		return nil, err
+	}
+
+	s.quicListener, err = quic.Listen(p, s.tlsConfig, s.quicConfig)
```
Tantalor93, line 182: why is this initialization of listener not protected by the mutex, which is used for some reason when you are stopping it? https://github.com/coredns/coredns/pull/6182/files#diff-f75b035c9af77b269e4c75e4603ab2e964916b15d7e7906fddc1bc4866bc6de2R207

core/dnsserver/server_quic.go
```diff
@@ -0,0 +1,293 @@
...
+			// indicate, after the last response, through the STREAM FIN
+			// mechanism that no further data will be sent on that stream.
+			_ = stream.Close()
+		}()
+
+	}
+}
+
+func (s *ServerQUIC) serveQUICStream(stream quic.Stream, conn quic.Connection) {
+	bufPtr := s.bytesPool.Get().(*[]byte)
```
Tantalor93: I like the usage of pooled resource, but it hides the fact that we are allocating memory buffer 65535+2 bytes for message regardless of actual size of DNS message, that seems like not efficient usage of memory, can't we rather:
* read 2-byte [prefix](https://github.com/coredns/coredns/pull/6182/files#diff-d4ed07e777c92c5e42daca148e60fbb401f40fa8ba727323ed782e6a4012fcf0R32) to get DNS message length
* allocate buffer according to DNS message length
* read rest of the stream into the buffer?

jaehnri: I agree. The problem with this approach is that (at least for sync.Pool) all buffers must be of the same size.
We could use the following library that offers a sized pool of buffers: https://github.com/libp2p/go-buffer-pool. What do you think?

Tantalor93: imho I would keep it simple for now and ditch the resource pool, it is a premature resource optimization
=> just allocate a new buffer for each request and let the GO runtime garbage collect it as soon as you are done with it

core/dnsserver/server_quic.go
jaehnri: To be honest, I don't really understand the necessity of these mutex locks. I asked Miek about it and it looks like it is needed to avoid data races when shutting down the server. 

Previously, I put the locks according to the other servers (`server_grpc` and `server_https`, mainly) and none of them protect the listener initialization.

Anyway, it does make sense to do it, so I protected it as you mentioned.

core/dnsserver/server_quic.go
```diff
@@ -0,0 +1,254 @@
...
+			// indicate, after the last response, through the STREAM FIN
+			// mechanism that no further data will be sent on that stream.
+			_ = stream.Close()
+		}()
+
+	}
+}
+
+func (s *ServerQUIC) serveQUICStream(stream quic.Stream, conn quic.Connection) {
+	buf, err := io.ReadAll(stream)
```
Tantalor93: AFAIK `io.ReadAll` is not very optimal regarding the memory usage, it internally [reallocates memory buffer until the stream fits.](https://github.com/golang/go/blob/master/src/io/io.go#L715) As I suggested, would it be possible to go with:
1. read 2-byte [prefix](https://github.com/coredns/coredns/pull/6182/files#diff-d4ed07e777c92c5e42daca148e60fbb401f40fa8ba727323ed782e6a4012fcf0R32) to get DNS message length
2. allocate buffer according to DNS message length
3. read rest of the stream into the buffer?

jaehnri: Sorry, I pushed and instantly noticed it. Now, it is done as you've suggested. Thanks for the careful review!

core/dnsserver/quic.go
```diff
@@ -0,0 +1,47 @@
...
+func AddPrefix(b []byte) (m []byte) {
+	m = make([]byte, 2+len(b))
+	binary.BigEndian.PutUint16(m, uint16(len(b)))
+	copy(m[2:], b)
+
+	return m
+}
+
+// These methods implement the dns.ResponseWriter interface from Go DNS.
+func (w *DoQWriter) Close() error          { return nil }
```
Tantalor93: shouldnt you close the stream here?

core/dnsserver/server_quic.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+func (s *ServerQUIC) serveQUICStream(stream quic.Stream, conn quic.Connection) {
+	buf, err := readDOQMessage(stream)
+
+	// io.EOF does not really mean that there's any error, it is just
+	// the STREAM FIN indicating that there will be no data to read
+	// anymore from this stream.
+	if err != nil && err != io.EOF {
+		closeQUICConn(conn, DoQCodeProtocolError)
```
Tantalor93: why are you forcing whole connection to close in this case?

core/dnsserver/server_quic.go
```diff
@@ -0,0 +1,302 @@
...
+
+		return
+	}
+
+	req := &dns.Msg{}
+	err = req.Unpack(buf)
+	if err != nil {
+		clog.Debugf("unpacking quic packet: %s", err)
+		closeQUICConn(conn, DoQCodeProtocolError)
```
Tantalor93: shouldnt you close only stream?

core/dnsserver/server_quic.go
```diff
@@ -0,0 +1,302 @@
...
+
+		return
+	}
+
+	if !validRequest(req) {
+		// If a peer encounters such an error condition, it is considered a
+		// fatal error. It SHOULD forcibly abort the connection using QUIC's
+		// CONNECTION_CLOSE mechanism and SHOULD use the DoQ error code
+		// DOQ_PROTOCOL_ERROR.
+		closeQUICConn(conn, DoQCodeProtocolError)
```
Tantalor93: single invalid request closes whole connection? Is this in RFC?

core/dnsserver/quic.go
jaehnri: Makes sense. I moved the `STREAM FIN` sending mechanism to this method. I noticed it is not called in the `serveDNS` routine, so I explicitly called it in `w.WriteMsg`.

core/dnsserver/server_quic.go
jaehnri: Yes, see https://www.rfc-editor.org/rfc/rfc9250#section-4.3.3-3

core/dnsserver/server_quic.go
jaehnri: There are two scenarios in which this code is called:
1) Server receives a `STREAM FIN` before receiving all the bytes for a message indicated in the 2-octet length field.
In this scenario, it is a protocol error and the connection must be closed.

2) Something bad happened while reading the stream. I am not sure if this will happen due to an internal server error or due to a bad message sent by the client. In both cases, the RFC does not specify what should happen with the stream and/or connection.

Thus, I think it is fair to close the connection and use the `DOQ_PROTOCOL_ERROR` signal, even though it MIGHT be a waste of resources. What do you think?

core/dnsserver/server_quic.go
jaehnri: > All DNS messages (queries and responses) sent over DoQ connections MUST be encoded as a 2-octet length field followed by the message content as specified in [[RFC1035](https://www.rfc-editor.org/rfc/rfc9250#RFC1035)].

I am assuming that if `req.Unpack` fails, it means the message was not according to `RFC1035`, thus a `DOQ_PROTOCOL_ERROR`. Then, we close the connection according to https://www.rfc-editor.org/rfc/rfc9250#section-4.3.3-3

core/dnsserver/server_quic.go
```diff
@@ -0,0 +1,302 @@
...
+
+	return s.ServeQUIC()
+}
+
+// ServeQUIC listens for incoming QUIC packets.
+func (s *ServerQUIC) ServeQUIC() error {
+	for {
+		conn, err := s.quicListener.Accept(context.Background())
+		if err != nil {
+			closeQUICConn(conn, DoQCodeInternalError)
```
jaehnri: @Tantalor93 I am unsure about this part. I am considering that, if we are unable to accept the connection/stream, it represents a `DOQ_INTERNAL_ERROR`.

In case of internal error, the RFC does not specify if we should drop the connection or only stream. What do you think?

Tantalor93: I think in this case, we should close the connection

but actually this connection closing is quite weird, because you are potentially closing already closed or non existing connection (because there was an error while accepting a new connection). so I think that `closeQUICConn` should be more robust to avoid panics

```
func closeQUICConn(conn quic.Connection, code quic.ApplicationErrorCode) {
	if conn == nil {
		// nothing to close
		return
	}
	...
```

Tantalor93:
@jaehnri looks good otherwise, so far from my testing it seems to be working 👍 But please consider adding integration test as @chrisohaver suggested https://github.com/coredns/coredns/pull/6182#pullrequestreview-1514888047

core/dnsserver/server_quic.go
```diff
@@ -0,0 +1,302 @@
...
+	// See https://www.rfc-editor.org/rfc/rfc9250.html#section-4.2-4
+	sizeBuf := make([]byte, 2)
+	_, err := io.ReadFull(r, sizeBuf)
+	if err != nil {
+		return nil, err
+	}
+
+	size := binary.BigEndian.Uint16(sizeBuf)
+	buf := make([]byte, size)
+	_, err = readAll(r, buf)
```
Tantalor93: why would it be bad to use `io.ReadFull` here?

core/dnsserver/server_quic.go
jaehnri: True. I updated the method to include this check. Thanks!

core/dnsserver/server_quic.go
jaehnri: Didn't realize we could drop that ReadAll. Thanks!

jaehnri:
I fixed the CI problems we had previously and added a simple test coverage for the server and its writer. LMK if I can enhance anything else. Thanks!!

core/dnsserver/server_quic.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	w := &DoQWriter{
+		localAddr:  conn.LocalAddr(),
+		remoteAddr: conn.RemoteAddr(),
+		stream:     stream,
+		Msg:        req,
+	}
+
+	dnsCtx := context.WithValue(context.Background(), Key{}, s.Server)
```
Tantalor93: so the stream cancellation is propagated correctly to plugins
```suggestion
	dnsCtx := context.WithValue(stream.Context(), Key{}, s.Server)
```

jaehnri:
Did a little oopsie while rebasing but it should be fixed now.

Tantalor93:
the server code LGTM, last thing I am thinking is about adding a metric for the quic server about what kind of responses it returns, similarly as we have for DoH server https://github.com/coredns/coredns/blob/c1b479fd024c91d03608587765dc9ae8745aef0c/plugin/metrics/vars/vars.go#L69, would be beneficial to have breakdown per quic status code, WDYT @jaehnri ?

jaehnri:
@Tantalor93 I've added support for the metrics plugin. 

While testing, I noticed a few possible outcomes of `conn.Accept()` and `conn.AcceptStream()` and thus enhanced the error checking there.  Please, review the last 2 commits. Thank you!

core/dnsserver/server_quic.go
```diff
@@ -202,16 +219,20 @@ func (s *ServerQUIC) Serve(l net.Listener) error { return nil }
...
 
 	clog.Debugf("closing quic conn %s with code %d", conn.LocalAddr(), code)
-
 	err := conn.CloseWithError(code, "")
 	if err != nil {
-		clog.Debugf("closing quic connection with code %d: %s", code, err)
+		clog.Debugf("failed to close quic connection with code %d: %s", code, err)
+	}
+
+	// DoQCodeNoError metrics are already registered after s.ServeDNS()
```
jaehnri, line 233: This part is a little tricky: metrics of successful queries will be registered after ServeDNS. 

However, for failed responses, I chose to report their metrics after closing their connections. This is the way I found to have one response per stream.

Tantalor93 (approved):
LGTM, great work 👍 

I think that would fit also nicely into 1.11.0, but I will let @chrisohaver and @yongtang if the release is not too big

also, I would recommend squashing the commits, because having commits like `Fix import ordering` and `Run go mod tidy` in the master branch is not that great 🙂

chrisohaver:
> I would recommend squashing the commits

They will squash automatically when merged.

jaehnri:
Glad to have my first meaningful contribution here. Thank you, @Tantalor93 and @chrisohaver, for the insightful review and guidance.

PR #6183, opened 2023-06-30
Change default value to 1232

pemensik:
As specified by [DNS flag day 2020](http://www.dnsflagday.net/2020/#action-dns-software-vendors), good and decent default value avoiding fragmentation issues should be 1232. It is quite likely 1500 would work reliably on local ethernet networks.

Value 512 is set implicitly and must be used for all clients, which did not include OPT RR with explicit value they support.

Since MR #5368 it should work correctly.

Changes EDNS0 default value to recommended value.

- #5366
- https://github.com/coredns/coredns/issues/5953

- https://github.com/coredns/coredns.io/issues/295

It removes the confusion around EDNS0 size. It may create behaviour change.

### DIFF
--- a/plugin/bufsize/README.md
+++ b/plugin/bufsize/README.md
@@ -1,11 +1,15 @@
 # bufsize
 ## Name
-*bufsize* - sizes EDNS0 buffer size to prevent IP fragmentation.
+*bufsize* - limits EDNS0 buffer size to prevent IP fragmentation.
 
 ## Description
-*bufsize* limits a requester's UDP payload size.
+*bufsize* limits a requester's UDP payload size to within a maximum value.
+If a request with an OPT RR has a bufsize greater than the limit, the bufsize
+of the request will be reduced. Otherwise the request is unaffected.
 It prevents IP fragmentation, mitigating certain DNS vulnerabilities.
-This will only affect queries that have an OPT RR.
+It cannot increase UDP size requested by the client, it can be reduced only.
+This will only affect queries that have
+an OPT RR ([EDNS(0)](https://www.rfc-editor.org/rfc/rfc6891)).
 
 ## Syntax
 ```txt
@@ -13,14 +17,14 @@ bufsize [SIZE]
 ```
 
 **[SIZE]** is an int value for setting the buffer size.
-The default value is 512, and the value must be within 512 - 4096.
+The default value is 1232, and the value must be within 512 - 4096.
 Only one argument is acceptable, and it covers both IPv4 and IPv6.
 
 ## Examples
 Enable limiting the buffer size of outgoing query to the resolver (172.31.0.10):
 ```corefile
 . {
-    bufsize 512
+    bufsize 1100
     forward . 172.31.0.10
     log
 }
@@ -29,7 +33,7 @@ Enable limiting the buffer size of outgoing query to the resolver (172.31.0.10):
 Enable limiting the buffer size as an authoritative nameserver:
 ```corefile
 . {
-    bufsize 512
+    bufsize 1220
     file db.example.org
     log
 }
--- a/plugin/bufsize/setup.go
+++ b/plugin/bufsize/setup.go
@@ -24,12 +24,13 @@ func setup(c *caddy.Controller) error {
 }
 
 func parse(c *caddy.Controller) (int, error) {
-	const defaultBufSize = 512
+	// value from http://www.dnsflagday.net/2020/
+	const defaultBufSize = 1232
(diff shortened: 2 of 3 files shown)


pemensik:
I think this plugin is somehow too simple. It can only further reduce EDNS accepted size after https://github.com/coredns/coredns/issues/5366. It is not able to add EDNS header to forwarded query, which I think documentation for it suggests.

It can only reduce accepted response size for queries, which already has EDNS0 header.

pemensik:
I think the example, which uses bufsize with forward plugin, creates confusion. It were originally able to add EDNS0 header to forwarded query. After MR https://github.com/coredns/coredns/pull/5368 it cannot do that anymore. There are cases where adding EDNS0 header is desired. But in case the response is larger than originating client query allowed, it would have to return response with TC bit set *and* EDNS0 header removed. That seems not possible anymore.

pemensik:
Related change in OpenShift: https://github.com/openshift/cluster-dns-operator/pull/370

pemensik:
It seems to me similar algorithm should be made as in https://github.com/coredns/coredns/pull/4085. When incoming query arrives, it should be safe to set internal supported EDNS0 size on the forwarded query. When the answer arrives, OPT RR may need to be removed from response, if it were not in original request.

Tantalor93 (approved):
Thanks for the PR, LGTM! would be nice to include it into `1.11.0` release
FYI @chrisohaver

Tantalor93 (changes requested):
oh, looks like `TestSetupBufsize` needs to be [updated](https://github.com/coredns/coredns/actions/runs/5425254281/jobs/9927741740?pr=6183). Shame on me for approving before triggering pipelines 🙂

chrisohaver:
I’d like to review all recent PRs/issues relating to Edns before they merge.

Tantalor93:
FYI @pemensik all commits must be signed https://github.com/coredns/coredns/pull/6183/checks?check_run_id=14770904337 🙂

pemensik:
Okay, I think it is ready for review now.

Tantalor93 (approved):

plugin/bufsize/README.md
```diff
@@ -1,26 +1,28 @@
 # bufsize
 ## Name
-*bufsize* - sizes EDNS0 buffer size to prevent IP fragmentation.
+*bufsize* - reduces EDNS0 buffer size to prevent IP fragmentation.
 
 ## Description
-*bufsize* limits a requester's UDP payload size.
+*bufsize* reduces a requester's UDP payload size.
```
chrisohaver:
```suggestion
*bufsize* limits a requester's UDP payload size to within a maximum value. If a request with an OPT RR has a bufsize greater than the limit, the bufsize of the request will be reduced. Otherwise the request is unaffected.
```

IMO "limit" is the correct word here.  It implies more directly that the value is limited by the configured value, and would not be "reduced", if below the limit.

plugin/bufsize/README.md
```diff
@@ -1,26 +1,28 @@
 # bufsize
 ## Name
-*bufsize* - sizes EDNS0 buffer size to prevent IP fragmentation.
+*bufsize* - reduces EDNS0 buffer size to prevent IP fragmentation.
```
chrisohaver:
```suggestion
*bufsize* - limits EDNS0 buffer size to prevent IP fragmentation.
```

chrisohaver (changes requested):
Thanks! The flag-day advised default value LGTM.  Minor changes to wording in docs requested.

plugin/bufsize/README.md
pemensik: Okay, explaining it more makes sense. Added just few line breaks to your version.

chrisohaver (approved):
lgtm thanks!

Tantalor93 (changes requested):
`TestSetupBufsize` needs adjustment https://github.com/coredns/coredns/actions/runs/5478594030/jobs/9979672394?pr=6183

```
Test 0: Bufsize not correctly set for input bufsize. Expected: 512, actual: 1232
```

pemensik:
> `TestSetupBufsize` needs adjustment https://github.com/coredns/coredns/actions/runs/5478594030/jobs/9979672394?pr=6183
> 
> ```
> Test 0: Bufsize not correctly set for input bufsize. Expected: 512, actual: 1232
> ```

Ouch, I have force-pushed previous version with just corrected documentation. I have already fixed this.

Tantalor93 (approved):
LGTM

PR #6194, opened 2023-07-04
install faillint properly

Tantalor93:
during execution of test coverage some tests are skipped due to missing `faillint` installed
image

and installation of faillint for E2E tests does not work properly either without specification of version
image

none

none

no

### DIFF
--- a/.github/workflows/go.coverage.yml
+++ b/.github/workflows/go.coverage.yml
@@ -22,7 +22,7 @@ jobs:
 
       - name: Test With Coverage
         run: |
-          go install github.com/fatih/faillint@latest || true
+          go install github.com/fatih/faillint@latest
           for d in request core coremain plugin test; do \
              ( cd $d; go test -coverprofile=cover.out -covermode=atomic -race ./...; [ -f cover.out ] && cat cover.out >> ../coverage.txt ); \
           done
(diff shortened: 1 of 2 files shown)


.github/workflows/go.coverage.yml
```diff
@@ -22,6 +22,7 @@ jobs:
 
       - name: Test With Coverage
         run: |
+          go install github.com/fatih/faillint@latest || true
```
SuperQ: `|| true` would ignore install errors, which is probably not what we want.

```suggestion
          go install github.com/fatih/faillint@latest
```

.github/workflows/go.test.yml
```diff
@@ -63,7 +63,7 @@ jobs:
 
       - name: Test
         run: |
-          go install github.com/fatih/faillint || true
+          go install github.com/fatih/faillint@latest || true
```
SuperQ: Same here, let's not ignore errors.

```suggestion
          go install github.com/fatih/faillint@latest
```

SuperQ (changes requested):

.github/workflows/go.coverage.yml
Tantalor93: ok 👍 fixed

.github/workflows/go.test.yml
Tantalor93: fixed 👍

SuperQ (approved):

PR #6195, opened 2023-07-06
Add support for RISC-V

chazapis:

This PR adds RISC-V support, by generating a RISC-V binary and container image.

No relevant issues found.

None.

No.

### Notes

By adding the architecture to `Makefile.release`, I was able to easily generate a binary for `riscv64`.

The container image generation was a bit trickier, as the base images used do not currently support RISC-V (`debian:stable-slim` and `gcr.io/distroless/static-debian11:nonroot`). Debian supports RISC-V in the "unstable" branch and distroless has [deferred releasing RISC-V images until the unstable branch is moved into stable](https://github.com/GoogleContainerTools/distroless/issues/1269) (estimated in 2025). ~Until then, the cleanest solution I could come up with is to use `Dockerfile.riscv64` for RISC-V with a simple and generic change in `Makefile.docker` to select the architecture-specific `Dockerfile` if available.~ Edit: For RISC-V we use unstable-derived base images, until there are RISC-V variants of the images used for all other architectures.

The resulting container has been tested in QEMU with a [RISC-V build of K3s](https://github.com/k3s-io/k3s/pull/7778).


Dockerfile.riscv64
```diff
@@ -0,0 +1,18 @@
+ARG DEBIAN_IMAGE=debian:unstable-slim
```
SuperQ: There's no need to have a separate Dockerfile for this. We don't use any architecture-specific pieces in the Docker build.

SuperQ (changes requested):

SuperQ:
Secondary, we don't need a whole new Dockerfile, we can simply adjust the image with a `DEBIAN_IMAGE` env var.

Dockerfile.riscv64
chazapis: This is not the only reason for the separate `Dockerfile`. There is also the problem of not having a distroless image with a `nonroot` user. Yes, the Debian image could be adjusted by setting the variable, but are there any better ideas on what to do with the missing distroless image?

chazapis: For example, using Alpine for the final image would be a solution, as Alpine supports RISC-V, but this is a rather big change (moving to Alpine for all architectures).

chazapis: Ok, I just found into [pre-made images of distroless for RISC-V](https://github.com/go-riscv/distroless/pkgs/container/distroless%2Fstatic-unstable). Somehow I missed that. I will try with the existing `Dockerfile`.

chazapis:
> Secondary, we don't need a whole new Dockerfile, we can simply adjust the image with a `DEBIAN_IMAGE` env var.

You are right. I uploaded a new version that just sets the base images to "unstable" variants with RISC-V support. I have tested this in a QEMU setup and it looks like it is working fine.

SuperQ (approved):
Nice, thanks!

PR #6204, opened 2023-07-14
plugin/rewrite: add rcode as a rewrite option

pschou:
### 1. Why is this pull request needed, and what does it do?

Adds the ability to mask server failure replies from a proxied DNS server.  This change makes CoreDNS "just work" in broken/firewalled environments.

For example, an IPv6 incompatible DNS appliance-- enables a stand-in proxy fix where IPv6 is not fully supported yet.  This option allows you to "ignore errors" and reply with an empty record.  When docker or Kubernetes sits behind a firewall appliance that blocks AAAA records from being resolved, the docker API sits and waits seconds to do multiple tries in hopes of getting a SUCCESS reply.  This flag masks this issue go away, and thus, the infrastructure runs faster.

In gist, it takes any SERVER-FAIL replies and switches them to SUCCESS to give the application a "no records found" reply.

None (that I know of)

The plugin README has been updated.

No


chrisohaver:
Thanks for contributing! 

> When docker or Kubernetes sits behind a firewall appliance that blocks AAAA records from being resolved, the docker API sits and waits seconds to do multiple tries

There may be other ways to solve the issue, with existing plugins such as _acl_ or _template_.
e.g. you can make coredns respond to all AAAA requests with empty response (aka nodata) with _acl_ plugin using `filter type AAAA`.

pschou:
> Thanks for contributing!

Thank you for the consideration and reply!

> There may be other ways to solve the issue, with existing plugins such as _acl_ or _template_. e.g. you can make coredns respond to all AAAA requests with empty response (aka nodata) with _acl_ plugin using `filter type AAAA`.

I agree that this is a potential avenue of resolution if we want all AAAA records to be taken away, but I still want IPv6 to be functional and enabled as much as possible-- so turning off all IPv6 goes against the plan to make IPv6 work where ever possible.

The SERVER-FAIL is something that the DNS appliance keeps throwing back.  It seems the DNS administrators think that SERVER-FAIL reply is the right course of action, but this keeps tripping up my Kubernetes deployments because they just stall whenever I try to deploy on a system.

palmerrb2000:
This fixed a longstanding problem for me. A sincere thank you to all who contribute!

chrisohaver:
Conceptually, this class of feature is a better fit for the _rewrite_ plugin.  It's a plugin that rewrites/overrides aspects of request and response.  E.g. It could be implemented by adding a new `FIELD` type option to the _rewrite_ plugin: `code FROM TO`.

Also if implemented in the _rewrite_ plugin, the response code override could be applied to a response from any plugin, not just _forward_ plugin.  So it would be more utilitarian, applicable to more situations.

pschou:
> Conceptually, this class of feature is a better fit for the _rewrite_ plugin. It's a plugin that rewrites/overrides aspects of request and response. E.g. It could be implemented by adding a new `FIELD` type option to the _rewrite_ plugin: `code FROM TO`.
> 
> Also if implemented in the _rewrite_ plugin, the response code override could be applied to a response from any plugin, not just _forward_ plugin. So it would be more utilitarian, applicable to more situations.

This is a very good point and I concur that a rewrite type would be a better fit for more diverse use cases.  I have amended the PR to realize this suggestion.

plugin/rewrite/README.md
```diff
@@ -335,6 +336,54 @@ rewrite ttl example.com. 30-
...
+
+The syntax for the RCODE rewrite rule is as follows. The meaning of
+`exact|prefix|suffix|substring|regex` is the same as with the name rewrite rules.
+An omitted type is defaulted to `exact`.
+
+```
+rewrite [continue|stop] rcode [exact|prefix|suffix|substring|regex] STRING OLD NEW
+```
+
+The values of OLD and NEW can be any of the following:
```
chrisohaver: Should mention that you accept both numeric and string representation for OLD and NEW response code.

plugin/rewrite/README.md
```diff
@@ -335,6 +336,54 @@ rewrite ttl example.com. 30-
 rewrite ttl example.com. 30 # equivalent to rewrite ttl example.com. 30-30
 ```
 
+### RCODE Field Rewrites
+
+At times, the need to rewrite a RCODE value could arise. For example, a DNS server
+may respond with a SERVFAIL instead of NOERROR records when AAAA records are requested.
+
+In the below example, the TTL in the answers for `coredns.rocks` domain are
+being set to `15`:
```
chrisohaver: Please update sentence for rcode example

plugin/rewrite/rcode.go
```diff
@@ -0,0 +1,180 @@
...
+	// try parsing integer based rcode
+	if err == nil && i <= 23 {
+		return int(i), true
+	}
+
+	for RCodeInt, RCodeStr := range dns.RcodeToString {
+		if strings.EqualFold(RCodeStr, v) {
+			return RCodeInt, true
+		}
+	}
```
chrisohaver: You can use the `dns.RcodeToString` map here.

pschou:
@chrisohaver Do you think the documentation should use OLD/NEW or FROM/TO?  I am considering switching it, but I would like to see what you think.

chrisohaver:
> @chrisohaver Do you think the documentation should use OLD/NEW or FROM/TO? I am considering switching it, but I would like to see what you think.

Best to be consistent with existing wording.   I think it uses from/to.

chrisohaver (commented):
Thanks!

wondywang:
#6214

plugin/rewrite/rcode.go
pschou: Good catch, I didn't see that map!

wondywang:
This PR also addresses a particular scenario I encountered. A big thank you to all contributors!

chrisohaver:
Fix import ordering and LGTM. Deferring final approval/merge to the rewrite plugin codeowners.

greenpau:
@pschou , please address the go coverage errors. Otherwise, LGTM.

greenpau (changes requested):
LGTM. Please address presubmit check.

greenpau:
@pschou , there are some other test failing too:

- DCO Action required after 1s — DCO

- ci/circleci: kubernetes-tests — Your tests failed on CircleCI

pschou:
@greenpau, this build process is fragile.

The pull approval was delayed all because of a single new line character, "\n", the test to verify that the imports are in the right format fail-- ok, I understand that consistency is important.  So, I add a single new line character to space the imports apart, and now other checks that previously passed are failing due to a curl command and because of an external API change at go.dev.  This issue is not related to this pull request, out of scope.

When pulling down this endpoint, one will find two lines are returned:
```
$ curl 'https://go.dev/VERSION?m=text'
go1.21.0
time 2023-08-04T20:14:06Z
```

The error in the kubernetes check looks like this:
```
LATEST=$(curl -s https://go.dev/VERSION?m=text)
curl https://dl.google.com/go/${LATEST}.linux-amd64.tar.gz | sudo tar xz -C $GOROOT --strip-components=1
curl: (6) Could not resolve host: time
curl: (3) URL using bad/illegal format or missing URL
```

Maybe a proposed fix would look like this:
```
LATEST=($(curl -s https://go.dev/VERSION?m=text))
curl "https://dl.google.com/go/${LATEST[0]}.linux-amd64.tar.gz" | sudo tar xz -C $GOROOT --strip-components=1
```

Note that every other check passed before this external API response change:
  https://app.circleci.com/pipelines/github/coredns/coredns/6794/workflows/4c459d4c-0fb8-4049-a554-743276e69d85/jobs/6867

Now it is failing due to the external API change:
  https://app.circleci.com/pipelines/github/coredns/coredns/6951/workflows/336e30ed-1b21-4de5-bcc5-8b83de43df91/jobs/7025

chrisohaver:
Please rebase (the go version check was fixed a few weeks ago)

pschou:
Ok, done 👍

greenpau (approved):
LGTM

PR #6212, opened 2023-07-19
plugin/forward: add continue_on_error as an option

pschou:

Adds the ability to continue processing a DNS response after a failure in the forward plugin. This change enables CoreDNS to process additional DNS lookups when a proxy fails to provide a valid response.  One can thus follow a forward plugin up with additional handlers to handle the error cases, such as by providing a "catch-all" DNS entry or an empty response.

None (that I know of)

The plugin README has been updated.

No


chrisohaver:
You can achieve this sort of behavior with the [alternate](https://github.com/coredns/alternate) external plugin.

chrisohaver:
Another external plugin to checkout out is [_fanout_](https://coredns.io/explugins/fanout/)

PR #6213, opened 2023-07-19
plugin/template: documentation - update the URL for caddyfile and add an example

pschou:

Documentation update and addition

None

This is only documentation

No

### DIFF
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -96,7 +96,7 @@ The most simplistic template is
 2. All queries will be answered (no `fallthrough`)
 3. The answer is always NXDOMAIN
 
-### Resolving SOA with dynamic HOSTNAME and static NS entries
+### Resolving SOA/NS for a dynamic domain
 
 Return a start of authority record with a dynamic hostname:
 
@@ -104,7 +104,7 @@ Return a start of authority record with a dynamic hostname:
 engineering.coredns.io. {
     template IN SOA {
       match ^engineering\.coredns\.io\.$
-      answer "{{ .Zone }} 3600 IN SOA {$HOSTNAME}. engineering.coredns.io. (1 60 120 86400 3600)"
+      answer "{{ .Zone }} 3600 IN SOA ns1.{{ .Zone }} admin.{{ .Zone }} (1 60 120 86400 3600)"
       fallthrough
     }
     template IN SOA {
@@ -123,9 +123,9 @@ engineering.coredns.io. {
 ~~~
 
 1. This template uses the defined zone (`engineering.coredns.io.`)
-2. All queries to SOA will be answered, the fallthrough on the first SOA block allows the second SOA block to be parsed as the match will only match subdomains and non-matching results will be handled by the second.
-3. A SOA record is sent with the local HOSTNAME defined by the environment.
-4. A pair of NS record is sent for any NS requests to the zone directly, while requests to subdomains will be served an SOA record.
+2. The hostname used as the SOA record name is retrieved from the HOSTNAME environment variable. HOSTNAME contains a fully qualified domain name, in the zone `engineering.coredns.io.`.
+3. All SOA/NS queries for the exact hostname will be answered with respective answers.
+4. For subdomains of the hostname, the templates will `fallthrough` to the next template and return an empty answer (NODATA) and include the SOA record in the authority section (as required for negative answers).
 
 ### Resolve .invalid as NXDOMAIN
 
--- a/plugin/template/README.md
+++ b/plugin/template/README.md
@@ -96,6 +96,37 @@ The most simplistic template is
 2. All queries will be answered (no `fallthrough`)
 3. The answer is always NXDOMAIN
 
+### Resolving SOA with dynamic HOSTNAME and static NS entries
+
+Return a start of authority record with a dynamic hostname:
+
+~~~ corefile
+engineering.coredns.io. {
+    template IN SOA {
+      match ^engineering\.coredns\.io\.$
+      answer "{{ .Zone }} 3600 IN SOA {$HOSTNAME}. engineering.coredns.io. (1 60 120 86400 3600)"
+      fallthrough
+    }
+    template IN SOA {
+      authority "{{ .Zone }} 3600 IN SOA {$HOSTNAME}. engineering.coredns.io. (1 60 120 86400 3600)"
+    }
+    template IN NS {
+      match ^engineering\.coredns\.io\.$
+      answer "{{ .Zone }} 900 IN NS ns1.engineering.coredns.io."
+      answer "{{ .Zone }} 900 IN NS ns2.engineering.coredns.io."
+      fallthrough
+    }
+    template IN NS {
+      authority "{{ .Zone }} 3600 IN SOA {$HOSTNAME}. engineering.coredns.io. (1 60 120 86400 3600)"
+    }
+}
+~~~
+
+1. This template uses the defined zone (`engineering.coredns.io.`)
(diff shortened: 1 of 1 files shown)


plugin/template/README.md
```diff
@@ -96,6 +96,37 @@ The most simplistic template is
...
+      fallthrough
+    }
+    template IN NS {
+      authority "{{ .Zone }} 3600 IN SOA {$HOSTNAME}. engineering.coredns.io. (1 60 120 86400 3600)"
+    }
+}
+~~~
+
+1. This template uses the defined zone (`engineering.coredns.io.`)
+2. All queries to SOA will be answered, the fallthrough on the first SOA block allows the second SOA block to be parsed as the match will only match subdomains and non-matching results will be handled by the second.
```
chrisohaver:
```suggestion
3. All SOA/NS queries for the exact hostname will be answered with respective answers.
4. For subdomains of the hostname, the templates will `fallthrough` to the next template and return an empty answer (NODATA) and include the SOA record in the authority section (as required for negative answers).
```

plugin/template/README.md
```diff
@@ -96,6 +96,37 @@ The most simplistic template is
...
+3. A SOA record is sent with the local HOSTNAME defined by the environment.
```
chrisohaver:
```suggestion
```

plugin/template/README.md
```diff
@@ -96,6 +96,37 @@ The most simplistic template is
...
+4. A pair of NS record is sent for any NS requests to the zone directly, while requests to subdomains will be served an SOA record.
```
chrisohaver:
```suggestion
```

plugin/template/README.md
```diff
@@ -96,6 +96,37 @@ The most simplistic template is
 2. All queries will be answered (no `fallthrough`)
 3. The answer is always NXDOMAIN
 
+### Resolving SOA with dynamic HOSTNAME and static NS entries
```
chrisohaver:
```suggestion
### Resolving SOA/NS for a dynamic domain
```

plugin/template/README.md
```diff
@@ -96,6 +96,37 @@ The most simplistic template is
...
+      answer "{{ .Zone }} 900 IN NS ns2.engineering.coredns.io."
+      fallthrough
+    }
+    template IN NS {
+      authority "{{ .Zone }} 3600 IN SOA {$HOSTNAME}. engineering.coredns.io. (1 60 120 86400 3600)"
+    }
+}
+~~~
+
+1. This template uses the defined zone (`engineering.coredns.io.`)
```
chrisohaver, line 125:
```suggestion
1. This template uses the defined zone (`engineering.coredns.io.`)
2. The hostname used as the SOA record name is retrieved from the HOSTNAME environment variable. HOSTNAME contains a fully qualified domain name, in the zone `engineering.coredns.io.`.
```

plugin/template/README.md
```diff
@@ -96,6 +96,37 @@ The most simplistic template is
...
 
+### Resolving SOA with dynamic HOSTNAME and static NS entries
+
+Return a start of authority record with a dynamic hostname:
+
+~~~ corefile
+engineering.coredns.io. {
+    template IN SOA {
+      match ^engineering\.coredns\.io\.$
+      answer "{{ .Zone }} 3600 IN SOA {$HOSTNAME}. engineering.coredns.io. (1 60 120 86400 3600)"
```
chrisohaver: I don't think the SOA field values are correct here.

First field after record type should be the primary master namserver.  e.g. `ns1.engineering.coredns.io.`
Second field after record type should be an encoded admin email address. e.g. `admin.engineering.coredns.io.`

```suggestion
      answer "{{ .Zone }} 3600 IN SOA ns1.{{ .Zone }} admin.{{ .Zone }} (1 60 120 86400 3600)"
```

This would be the same for all returned SOA records.  But then, the dynamic hostname isn't used anywhere, so. I'm confused by the example. I'm probably missing something.

chrisohaver (commented):
I'm probably missing something, but I cant make sense of the example.  The example appears to make the SOA primary nameserver dynamic?  What is the use case for this?

pschou:
> I'm probably missing something, but I cant make sense of the example. The example appears to make the SOA primary nameserver dynamic? What is the use case for this?

Thank you for the point of information request.  I was not clear!  This configuration block, by being dynamic, can be backed up into a git repository and used for keeping multiple DNS servers in sync.  By having the configuration be a dynamic variable, one can do a git pull on the target server and reload/bounce the process without having to do manual edits to a flat text file.

plugin/template/README.md
```diff
@@ -96,6 +96,37 @@ The most simplistic template is
...
+
+~~~ corefile
+engineering.coredns.io. {
+    template IN SOA {
+      match ^engineering\.coredns\.io\.$
+      answer "{{ .Zone }} 3600 IN SOA ns1.{{ .Zone }} admin.{{ .Zone }} (1 60 120 86400 3600)"
+      fallthrough
+    }
+    template IN SOA {
+      authority "{{ .Zone }} 3600 IN SOA {$HOSTNAME}. engineering.coredns.io. (1 60 120 86400 3600)"
```
chrisohaver, line 111:
```suggestion
      authority "{{ .Zone }} 3600 IN SOA ns1.{{ .Zone }} admin.{{ .Zone }} (1 60 120 86400 3600)"
```

plugin/template/README.md
```diff
@@ -96,6 +96,37 @@ The most simplistic template is
...
+    }
+    template IN NS {
+      match ^engineering\.coredns\.io\.$
+      answer "{{ .Zone }} 900 IN NS ns1.engineering.coredns.io."
+      answer "{{ .Zone }} 900 IN NS ns2.engineering.coredns.io."
+      fallthrough
+    }
+    template IN NS {
+      authority "{{ .Zone }} 3600 IN SOA {$HOSTNAME}. engineering.coredns.io. (1 60 120 86400 3600)"
```
chrisohaver, line 120:
```suggestion
      authority "{{ .Zone }} 3600 IN SOA ns1.{{ .Zone }} admin.{{ .Zone }} (1 60 120 86400 3600)"
```

chrisohaver:
> This configuration block, by being dynamic, can be backed up into a git repository and used for keeping multiple DNS servers in sync. 

In sync to do what though?  That part is still unclear.

pschou:
> In sync to do what though? That part is still unclear.

Basically, to track changes and keep two servers in sync.

Server A and Server B both have a hosts.txt file and act as authoritative servers.  They both do a git pull from a git repo to pull changes into the local folder.  The config files are stored in the git repo like on disk.  Should the git repo fail to provide updates, both servers will continue to serve the last known good update.  All updates to the hosts.txt file are documented with git change commit logs.  The flat text file seems to work well for both git and ease of reading/finding entries.

chrisohaver:
If the two servers are serving the same zone, shouldn't the SOAs be more or less identical?  Why do they need be dynamic based on the name of the host?

pschou:
> If the two servers are serving the same zone, shouldn't the SOAs be more or less identical? Why do they need to be dynamic based on the name of the host?

Yes.  The idea of using HOSTNAME was to make whichever server, A or B, declare itself as being the SOA for a client query.

I realize now that this would not be a good plan when DNSSEC is brought into consideration...

chrisohaver:
> declare itself as being the SOA for a client query

Not sure what you mean by this.  Do you mean declare itself as the primary nameserver (MNAME)?

There should only be one SOA record per zone.  Two servers serving the same zone should serve the same SOA record.

I think something like this could possibly make sense if the two servers are essentially disconnected, and a given client querying one of the servers has no knowledge of the other server copy.   But it is a pretty esoteric situation, so I don't think it's a good fit as a template example.

PR #6216, opened 2023-07-20
plugin/recursion: enable server side CNAME recursive lookups

pschou:

Enables recursive lookups by catching CNAME entries without corresponding request type entry and tries to resolve them by doing sub-queries out to the intended target.

None that I know of.

Note:  This plugin can be a replacement for `explugins/unbound` for recursive queries.  `recursion` compiles into a static binary while `explugins/unbound` has the bug in which it cannot be compiled into a static binary.

Added README for plugin

No.

### DIFF
--- a/core/dnsserver/zdirectives.go
+++ b/core/dnsserver/zdirectives.go
@@ -34,6 +34,7 @@ var Directives = []string{
 	"acl",
 	"any",
 	"chaos",
+	"recursion",
 	"loadbalance",
 	"tsig",
 	"cache",
--- a/core/plugin/zplugin.go
+++ b/core/plugin/zplugin.go
@@ -42,6 +42,7 @@ import (
 	_ "github.com/coredns/coredns/plugin/nsid"
 	_ "github.com/coredns/coredns/plugin/pprof"
 	_ "github.com/coredns/coredns/plugin/ready"
+	_ "github.com/coredns/coredns/plugin/recursion"
 	_ "github.com/coredns/coredns/plugin/reload"
 	_ "github.com/coredns/coredns/plugin/rewrite"
 	_ "github.com/coredns/coredns/plugin/root"
--- a/plugin.cfg
+++ b/plugin.cfg
@@ -43,6 +43,7 @@ dns64:dns64
 acl:acl
 any:any
 chaos:chaos
+recursion:recursion
 loadbalance:loadbalance
 tsig:tsig
 cache:cache
--- a/plugin/recursion/README.md
+++ b/plugin/recursion/README.md
@@ -1,0 +1,79 @@
+# recursion
+
+## Name
+
+*recursion* - queries recursively for a resource.
+
+## Description
+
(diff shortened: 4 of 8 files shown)


chrisohaver:
Thanks for contributing! Some initial thoughts:

For all new plugins, we should first establish if there is a general need for the plugin, to determine if this should be an external plugin, or included here as an internal plugin. 

If there is a specific case that you need this function for, please describe that (and explain why the CNAME comes back without the target record).

CNAME resolution is already inherent in the DNS spec. In the scenario that you describe, the target record type, if it actually exists, will already be included in the result, so this type of function should not be needed.

"recursion" as a name would likely be confused with DNS Recursion (which is not the same thing as CNAME resolution), giving people the wrong idea of what the plugin does based on its name.  A better name would be _cname_ or *cname_lookup*.

**Possible workaround**: In a pinch, you _might_ be able to use the new cname rewrite option in the _rewrite_ plugin to do this?  I don't recall exactly how it operates, but there is a chance that a no-op cname rewrite (rewrite target to itself) may result in looking up and adding the missing target record.  IMO worth checking out.

pschou:
> Thanks for contributing! Some initial thoughts:

You're welcome

> For all new plugins, we should first establish if there is a general need for the plugin, to determine if this should be an external plugin, or included here as an internal plugin.

The real need is a 'built-in' capability so one does not have to compile with a non-static binary.  I appreciate the idea of keeping things simple and to the point unless additional functionality is needed.

> If there is a specific case that you need this function for, please describe that (and explain why the CNAME comes back without the target record).

The challenge comes in when a CNAME points to another server, which has a CNAME record there (like a load balancer).  For simplicity, if the client gets the final set of usable records when asking for "RecursionDesired", this would enable this flag being parsed even when the lookup has to be done elsewhere one or more hops.

> CNAME resolution is already inherent in the DNS spec. In the scenario that you describe, the target record type, if it actually exists, will already be included in the result, so this type of function should not be needed.

Yes, when testing (see README.md), the results returned are not recursively looked up.

> "recursion" as a name would likely be confused with DNS Recursion (which is not the same thing as CNAME resolution), giving people the wrong idea of what the plugin does based on its name. A better name would be _cname_ or _cname_lookup_.

Agreed, the name recursion has been overloaded and the actual implementation needs to be clarified.  The CNAME flattening is what is desirable.

> **Possible workaround**: In a pinch, you _might_ be able to use the new cname rewrite option in the _rewrite_ plugin to do this? I don't recall exactly how it operates, but there is a chance that a no-op cname rewrite (rewrite target to itself) may result in looking up and adding the missing target record. IMO worth checking out.

I appreciate the pointer and I do see on line 75 the function call to an upstream lookup.  

```
        r.state.Req.Question[0].Name = toTarget
        upRes, err := r.Upstream.Lookup(r.ctx, r.state, toTarget, r.state.Req.Question[0].Qtype)
```

Nice!  I didn't realize this hidden gem existed.

PR #6224, opened 2023-07-25
Maintaining prior releases

yashsingh74:
Currently, for coredns a release is maintained from the master branch. All the patches, bug fixes, and CVEs are cut from the master branch. Once the releases are cut, there is no scope for maintaining the older release versions. 
This proposal allows to achieve the following goals:

- Maintaining the older versions up to N-2 major or minor releases. 
- It also includes a branching strategy, which will help to cut the number of minor and patch releases while also making it easier to maintain the branches. 

Upstream repositories like Kubernetes, containerd, etc. use the same procedure. 

Based on the discussion here - https://github.com/coredns/coredns/discussions/6150

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,0 +1,40 @@
+# Releases Guide
+
+Here, we will briefly understand the release plan and versioning steps. 
+This will cover the release process, major and minor releases, backporting the PR and manual release steps.
+
+If there is something that you require or this document leaves out, please
+reach out by [filing an issue](https://github.com/coredns/coredns/issues).
+
+## Releases
+
+All the major and minor releases will be made from the master. The tag will be of the format `v<major>.<minor>.<patch>`.
+
+After a minor release, a branch will be created with the format of `release-<major>.<minor>` from the minor tag. Once the branch is ready, all the patches will be released from that branch. 
+For example, once we release `v1.10.0`, a branch will be created as `release-1.10` with the following `v1.10.0` tag. Next, all future patches will be done against that branch. i.e. `v1.10.1, v1.10.2, ...`.
+
+## Next Release
+
+The next _minor_ release will be tracked as a GitHub Milestone.
+
+## Coredns Support
+
+The support will be identified with several states:
+
+- __*Active*__: The release is a stable branch that is currently supported and accepting patches.
+- __*End of Life*__: The release branch is no longer supported, and no new patches will be accepted.
+
+Currently, we are planning to provide maintenance support for the two latest release branches **(N, N-1)**. Examples: If the latest release is 1.10, we will support 1.9 and 1.10.  If the latest release is 2.0, then we will support 2.0, and the most recent release branch of 1.X. 
+
+Maintenance releases will be supported for upto 1 year until the end of life is announced for the branch i.e. v1.9 **(N-2)** will be EOL when v1.11 (N) is released.
+These branches will accept bug reports and backports until the end of life.
+
+## Backporting
+
+If there are important bug or security fixes that need to be backported please let us know in one of three ways:
+- Open an issue.
+- Open a PR with a cherry-picked change from the master branch.
+- Open a PR with a ported fix.
+
+If there is no existing fix in the master, you should first fix it in the master. Once the PR is merged, back port the fix to the supported release branches.
+If the issue is only in the release branch and not in the master, then open a PR and make the specific changes on the master branch to fix the issue.


chrisohaver:
@johnbelamaric, @stp-ip, @superq, @tantalor93

Moving from master branch only, to multiple minor release branch support is one of those things that warrants a committee vote IMO, since it does entail taking on more work.

IMO I think we should go this route (supporting multiple minor releases concurrently).  Worst case scenario is that we lag behind and don't have time to maintain prior minor releases, in which case we no worse off than where we are now.  The user community may get frustrated about that but I'm sure many are already frustrated with our current single branch support policy.

One change I'd suggest would be to make support for prior minor release based on time (e.g. 1 year), rather than a set number of releases, since our release cadence is not regular.

RELEASES.md
```diff
@@ -0,0 +1,78 @@
...
+
+| Release                                                              | Status        | Start         | End of Life             |
+| ---------                                                            | ------------- |---------------|-------------------------|
+| 1.8                                                            | End of Life | Oct 22, 2020  | Jan 14, 2022            |
+| 1.9                                                            | Maintenance | Feb 10, 2022  | Until 1.11 is released. |
+| 1.10                                                            | Active | Sept 19, 2022 | 6 months or more        |
+| 1.11                                                            | Next | TBD           | TBD                     |
+
+
+## Kubernetes Support
```
chrisohaver: I think this section intends to say that we will ensure that there is always a supported coredns release branch that supports every supported version of K8s.  FWIW, We have already been doing this for the master branch, so this doesn't change anything, but it's good to formalize it here.  In _theory_ this could put us on the hook for maintaining releases older than N-2, but this would probably not happen.

RELEASES.md
```diff
@@ -0,0 +1,78 @@
...
+
+The Kubernetes matrix represents the versions of coredns that are used in the Kubernetes release. Kubernetes only supports `N-3` minor release versions, and it will ensure there is always a supported version of coredns available.
+
+| Kubernetes Version | Coredns Version |
```
chrisohaver: I don't think the matrix of versions is necessary here (it would need constant updates every time k8s releases).  It also kind of implies that we should support the minor release listed here (or at least could reasonably be interpreted that way).  E.g. When we release 1.11.0, it implies that we should continue to support 1.8.  But k8s 1.24 works with coredns 1.10, so that is not the case.

RELEASES.md
```diff
@@ -0,0 +1,78 @@
...
+
+The next release will be tracked in the milestone under the issue section, with Release tracking. 
+Also, it can be filtered with the label release. If your issue or PR is not present in a milestone, please create a PR to track the next release.
+
+## Coredns Support
+
+The support will be identified with several states:
+
+- __*Active*__: The release is a stable branch that is currently supported and accepting patches.
+- __*Maintenance*__: The release that is currently supported and accepting patches.
```
AmitKumarDas: suggest: This still does not clarify the differences between Active vs. Maintenance branches.

RELEASES.md
```diff
@@ -0,0 +1,78 @@
...
+- __*End of Life*__: The release branch is no longer supported, and no new patches will be accepted.
+
+Currently, we are planning to support at least two latest major and minor releases **(N, N-1)**. The latest releases can be any combination of major or minor releases. Examples: a major release followed by two or more minor releases (e.g. 1.9, 1.10). 
+It is unknown when the next release will be made available. So, until then, the current or latest release will be considered the Active release.
+
+For e.g.
+
+The active release is *v1.10 (N)* and *v1.9 (N-1)* is in maintenance. But when the next release rolls out i.e. *v1.11 (N+1)* then *N, N+1* will be the active release, and *N-1* will be in maintenance mode.
```
AmitKumarDas: Clarification: Do we intend to have N-1 i.e. v1.9 in maintenance when both N & N+1 are active releases.

RELEASES.md
```diff
@@ -0,0 +1,78 @@
...
+
+Maintenance releases will be supported for upto 6 months or more until the end of life is announced for the branch i.e. v1.9 will be EOL when v1.11 is released, as shown in the diagram.
```
AmitKumarDas: This line contradicts with above line

RELEASES.md
```diff
@@ -0,0 +1,78 @@
...
+
+Releases will be created from the release branches. The release will follow the following convention as `<major>.<minor>.<patch>`. 
+All the major and minor releases will be made from the master. The tag will be of the format `v<major>.<minor>.<patch>`.
+
+After a minor release, a branch will be created with the format of `release-<major>.<minor>` from the minor tag. Once the branch is ready, all the patches will be released from that branch. 
+For example, once we release `v1.10.0`, a branch will be created as `release-1.10` with the following `v1.10` tag. Next, all future patches will be done against that branch. i.e. `v1.10.0, v1.10.1`.
+
+## Next Release
+
+The next release will be tracked in the milestone under the issue section, with Release tracking.
```
chrisohaver:
```suggestion
The next _minor_ release will be tracked as a GitHub Milestone. 
```

RELEASES.md
```diff
@@ -0,0 +1,78 @@
...
+Releases will be created from the release branches. The release will follow the following convention as `<major>.<minor>.<patch>`. 
+All the major and minor releases will be made from the master. The tag will be of the format `v<major>.<minor>.<patch>`.
+
+After a minor release, a branch will be created with the format of `release-<major>.<minor>` from the minor tag. Once the branch is ready, all the patches will be released from that branch. 
+For example, once we release `v1.10.0`, a branch will be created as `release-1.10` with the following `v1.10` tag. Next, all future patches will be done against that branch. i.e. `v1.10.0, v1.10.1`.
+
+## Next Release
+
+The next release will be tracked in the milestone under the issue section, with Release tracking. 
+Also, it can be filtered with the label release. If your issue or PR is not present in a milestone, please create a PR to track the next release.
```
chrisohaver:
```suggestion
```

RELEASES.md
chrisohaver: Agreed - I'm not sure why there is a distinction between active ad maintenance branches. A branch is either supported, or not supported.  AFAIK, there are no operational differences between the most recent release, and the prior one.

RELEASES.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Coredns Support
+
+The support will be identified with several states:
+
+- __*Active*__: The release is a stable branch that is currently supported and accepting patches.
+- __*Maintenance*__: The release that is currently supported and accepting patches.
+- __*End of Life*__: The release branch is no longer supported, and no new patches will be accepted.
+
+Currently, we are planning to support at least two latest major and minor releases **(N, N-1)**. The latest releases can be any combination of major or minor releases. Examples: a major release followed by two or more minor releases (e.g. 1.9, 1.10).
```
chrisohaver: Im confused by wording here and use of major/minor. Suggesting changes/ for clarity...

```suggestion
Currently, we are planning to provide maintenance support for the two latest release branches **(N, N-1)**. Examples: If the latest release is 1.10, we will support 1.9 and 1.10.  If the latest release is 2.0, then we will support 2.0, and the most recent release branch of 1.X.
```

RELEASES.md
chrisohaver: The wording is confusing to me, but I think this attempts to explain a common sense aspect of release progression that doesn't warrant spelling out.

```suggestion
```

RELEASES.md
chrisohaver: IMO we should stick to either date based XOR number of releases based support expirations (XOR, not both).

RELEASES.md
```diff
@@ -0,0 +1,78 @@
...
+
+The current state of available support will be:
+
+
+| Release                                                              | Status        | Start         | End of Life             |
+| ---------                                                            | ------------- |---------------|-------------------------|
+| 1.8                                                            | End of Life | Oct 22, 2020  | Jan 14, 2022            |
+| 1.9                                                            | Maintenance | Feb 10, 2022  | Until 1.11 is released. |
+| 1.10                                                            | Active | Sept 19, 2022 | 6 months or more        |
+| 1.11                                                            | Next | TBD           | TBD                     |
```
chrisohaver: Ideally our support policy should be obvious to the point that this table is not necessary.  I think that the proposal is a combination of both release counts and original release dates makes this too complex.

RELEASES.md
```diff
@@ -0,0 +1,78 @@
...
+
+The maintainers are responsible for the backporting process. They will ensure any bug fixes that need to be in active release or any new features for the next major and minor releases.
+
+If there are important fixes that need to be backported, please let us know in one of three ways:
+- Open an issue.
+- Open a PR with a cherry-picked change from the master branch.
+- Open a PR with a ported fix.
+
+If there is no existing fix in the master, you should first fix it in the master by creating an issue. After that, once the PR is merged, open an issue on the release branch and backport it.
+If the issue is only in the release branch and not in the master, then open a PR and make the specific changes on the release branch to fix the issue.
```
chrisohaver:
```suggestion
If the issue is only in the release branch and not in the master, then open a PR and make the specific changes on the master branch to fix the issue.
```

RELEASES.md
```diff
@@ -0,0 +1,78 @@
...
+## Backporting
+
+The maintainers are responsible for the backporting process. They will ensure any bug fixes that need to be in active release or any new features for the next major and minor releases.
+
+If there are important fixes that need to be backported, please let us know in one of three ways:
+- Open an issue.
+- Open a PR with a cherry-picked change from the master branch.
+- Open a PR with a ported fix.
+
+If there is no existing fix in the master, you should first fix it in the master by creating an issue. After that, once the PR is merged, open an issue on the release branch and backport it.
```
chrisohaver:
```suggestion
If there is no existing fix in the master, you should first fix it in the master. Once the PR is merged, back port the fix to the supported release branches.
```

RELEASES.md
```diff
@@ -0,0 +1,78 @@
...
+|--------------------|--------------------|
+| [1.24](https://github.com/kubernetes/kubernetes/blob/release-1.24/build/dependencies.yaml#L32)           | 1.8 (to be specific 1.8.6)    |
+| [1.25](https://github.com/kubernetes/kubernetes/blob/release-1.25/build/dependencies.yaml#L30)           | 1.9 (To be specific 1.9.3)    |
+| [1.26](https://github.com/kubernetes/kubernetes/blob/release-1.26/build/dependencies.yaml#L30)           | 1.9 (To be specific 1.9.3)    |
+| [1.27](https://github.com/kubernetes/kubernetes/blob/release-1.27/build/dependencies.yaml#L30)           | 1.10 (To be specific 1.10.1)   |
+
+
+## Backporting
+
+The maintainers are responsible for the backporting process. They will ensure any bug fixes that need to be in active release or any new features for the next major and minor releases.
```
chrisohaver: This kind of implies that only maintainers are supposed to back port fixes.  I don't think this should be the case, and I would expect any contributing community member to be able to open a PR to backport a fix.

If instead this is just saying that maintainers are _ultimately_ responsible (e.g. merging fixes), then this goes without saying, since that is an implied responsibility of maintainers for all PRs already.

```suggestion
The maintainers are responsible for the backporting process. They will ensure any bug fixes that need to be in active release or any new features for the next major and minor releases.
```

RELEASES.md
```diff
@@ -0,0 +1,78 @@
...
+- Open a PR with a cherry-picked change from the master branch.
+- Open a PR with a ported fix.
+
+If there is no existing fix in the master, you should first fix it in the master by creating an issue. After that, once the PR is merged, open an issue on the release branch and backport it.
+If the issue is only in the release branch and not in the master, then open a PR and make the specific changes on the release branch to fix the issue.
+
+
+## Manual steps
+
+Manual steps for creating the major or minor release is present [here](https://github.com/coredns/coredns/blob/master/Makefile.release).
```
chrisohaver: I don't think this section needs to be in this document.

```suggestion
## Manual steps

Manual steps for creating the major or minor release is present [here](https://github.com/coredns/coredns/blob/master/Makefile.release).
```

chrisohaver (changes requested):
I think this document should be re-aligned/trimmed down to be more directly about what releases we support (e.g. 2 latest minor releases), what is included in that support (e.g. bug fixes, security updates), and (perhaps) how that support is done.

I don't think it should try to cover the entire release process.

RELEASES.md
yashsingh74: Done

RELEASES.md
yashsingh74: Removed as per the suggestion.

RELEASES.md
yashsingh74: As, I was not sure previously till when the support will be provided when the branch is not active. So, I kept in maintenance mode which is still accepting fixes. But now N, N-1 will be active releases as per the maintainers suggestion.

RELEASES.md
yashsingh74: Done

RELEASES.md
yashsingh74: I removed this statement as this statement is confusing.

yashsingh74: Done.

RELEASES.md
yashsingh74: Updated the statement.

RELEASES.md
yashsingh74: Done

RELEASES.md
yashsingh74: Done

RELEASES.md
yashsingh74: Done

RELEASES.md
yashsingh74: Done

RELEASES.md
```diff
@@ -0,0 +1,41 @@
...
+
+If there is something that you require or this document leaves out, please
+reach out by [filing an issue](https://github.com/coredns/coredns/issues).
+
+## Releases
+
+Releases will be created from the release branches. The release will follow the following convention as `<major>.<minor>.<patch>`. 
+All the major and minor releases will be made from the master. The tag will be of the format `v<major>.<minor>.<patch>`.
+
+After a minor release, a branch will be created with the format of `release-<major>.<minor>` from the minor tag. Once the branch is ready, all the patches will be released from that branch.
```
chrisohaver, line 13:
```suggestion
After a minor release, a branch will be created with the format of `v<major>.<minor>` from the minor tag. Once the branch is ready, all the patches will be released from that branch. 
```

chrisohaver, line 13: The most recent release (N) can be supported and built from master, as we do now.
Creating a branch for the N release would mean having to merge every fix from master to it. This is a lot of additional error prone clerical work.

A branch should only be needed for the N-1 release.

IOW, if current release is 1.10...

N (1.10) branch = master
N-1 (1.9) branch = v1.9

If we find that we need some multi-PR collaborative work that would cross release boundaries, a new development branch should be created for that work, and merged to master when ready to release it.  This is not common, but has been done in the past a few times.

RELEASES.md
```diff
@@ -0,0 +1,41 @@
...
+Here, we will briefly understand the release plan and versioning steps. 
+This will cover the release process, major and minor releases, backporting the PR and manual release steps.
+
+If there is something that you require or this document leaves out, please
+reach out by [filing an issue](https://github.com/coredns/coredns/issues).
+
+## Releases
+
+Releases will be created from the release branches. The release will follow the following convention as `<major>.<minor>.<patch>`. 
+All the major and minor releases will be made from the master. The tag will be of the format `v<major>.<minor>.<patch>`.
```
chrisohaver: This is hard to follow:  

> Releases will be created from the release branches.

and 

> All the major and minor releases will be made from the master.

These seem to be directly contradictory.

RELEASES.md
```diff
@@ -0,0 +1,41 @@
+# Releases Guide
+
+Here, we will briefly understand the release plan and versioning steps. 
+This will cover the release process, major and minor releases, backporting the PR and manual release steps.
+
+If there is something that you require or this document leaves out, please
+reach out by [filing an issue](https://github.com/coredns/coredns/issues).
+
+## Releases
```
chrisohaver, line 9: This section may be easier to follow if presented as a numbered list of steps take for each release.

chrisohaver (changes requested):
For the new _minor_ release process, I'm thinking the following (e.g. releasing new minor version 1.11):

1. tag the release v1.11.0 in master branch
2. build and publish the release (tag v1.11.0)
3. create a new branch from the prior patch release (v1.10.1), named v1.10

On going maintenance: At this stage as part of on going maintenance, patches that should go into 1.10 should be applied to the v1.10 branch.

For the new _patch_ release process (e.g. v1.11.1 and v1.10.2):

1. tag the release v1.11.1 in master branch
2. build and publish the release (tag v1.11.1)
3. tag the release v1.10.2 in v1.10 branch
4. build and publish the release (tag v1.10.2)

RELEASES.md
yashsingh74: I think using `release-` will be a better approach because it is most common convention used throughout in Kubernetes and its core components. I believe we should align to that approach instead of using the `v`.

yashsingh74: Yes, I agree this will create a lot more additional work and cherry-picking. But yes we can start creating a branches which are going to be in maintenance mode i.e.. (N-1).
```
N (1.10) branch = master
N-1 (1.9) branch = release-1.9
```

chrisohaver:
It just occurred to me that supporting prior minor releases can have implications for external plugins that coredns maintains.  That is, for certain backward incompatible changes, an external plugin may also need multiple branches to support two incompatible supported coredns releases.

For now, I propose that for external plugins that coredns maintains, we are only going to support the latest coredns release.

yashsingh74:
> For now, I propose that for external plugins that coredns maintains, we are only going to support the latest coredns release.

I am still confused are we still planning to maintain the older releases.

chrisohaver:
> > For now, I propose that for external plugins that coredns maintains, we are only going to support the latest coredns release.
> 
> I am still confused are we still planning to maintain the older releases.

Yes, that is still proposed, but not for external plugins in the CoreDNS project.  For the reasons [above](https://github.com/coredns/coredns/pull/6224#issuecomment-1673395577), I propose that we should only support CoreDNS maintained _external plugins_ on the latest CoreDNS release.

RELEASES.md
yashsingh74: Done

chrisohaver:
Do any other maintainers have an opinion on this?

chrisohaver:
TLDR: This PR proposes that we start supporting patch releases for prior minor releases.

chrisohaver:
IMO, before spending more time hammering out the details of this document, we should get a consensus on whether or not CoreDNS maintainers are OK with taking on extra load of maintaining multiple releases.

One significant complication could be adapting [Corefile Migration](https://github.com/coredns/corefile-migration) to handle migrations through multiple parallel release branches.

RELEASES.md
```diff
@@ -0,0 +1,40 @@
...
+
+All the major and minor releases will be made from the master. The tag will be of the format `v<major>.<minor>.<patch>`.
+
+After a minor release, a branch will be created with the format of `release-<major>.<minor>` from the minor tag. Once the branch is ready, all the patches will be released from that branch. 
+For example, once we release `v1.10.0`, a branch will be created as `release-1.10` with the following `v1.10` tag. Next, all future patches will be done against that branch. i.e. `v1.10.0, v1.10.1`.
```
SuperQ: Let's not make the Go mistake of having non-semver first releases.

```suggestion
For example, once we release `v1.10.0`, a branch will be created as `release-1.10` with the following `v1.10.0` tag. Next, all future patches will be done against that branch. i.e. `v1.10.1, v1.10.2, ...`.
```

SuperQ:
We have two ways of doing release branches in the Prometheus project:

To start, we only add bugfixes (which includes security fixes) to our patch releases.

In `prometheus/prometheus` we fork the main branch at release time.
* Fixes are cherry-picked from `main` into the `release-X.Y` branch.
* Sometimes fixes are committed directly to `release-X.Y`.
* A changelog and version bump are added directly to `release-X.Y`.
* Changes and changelog entries are cherry-picked from `release-X.Y` back to `main`.

I personally find this workflow annoying and don't do this for the majority of Prometheus repos (exporters). For these we do a simpler workflow of all changes to `main` and then cherry-pick as PRs to `release-X.Y`. This is far easier IMO.

As for how we as a group want to support old releases, I would prefer a flexible "as needed" basis. If we make a big change, but there's a critical security bug we need to backport, we can put in the effort to backport. Basically, old releases supported as best effort.

RELEASES.md
```diff
@@ -0,0 +1,40 @@
...
+- __*End of Life*__: The release branch is no longer supported, and no new patches will be accepted.
+
+Currently, we are planning to provide maintenance support for the two latest release branches **(N, N-1)**. Examples: If the latest release is 1.10, we will support 1.9 and 1.10.  If the latest release is 2.0, then we will support 2.0, and the most recent release branch of 1.X. 
+
+Maintenance releases will be supported for upto 1 year until the end of life is announced for the branch i.e. v1.9 **(N-2)** will be EOL when v1.11 (N) is released.
+These branches will accept bug reports and backports until the end of life.
+
+## Backporting
+
+If there are important fixes, new features for the next minor releases that need to be backported, please let us know in one of three ways:
```
SuperQ: I don't think new features should be backported. This is technically not allowed in semver.

```suggestion
If there are important bug or security fixes that need to be backported please let us know in one of three ways:
```

yashsingh74:
Is there any changes required for the PR?

RELEASES.md
```diff
@@ -0,0 +1,40 @@
...
+## Coredns Support
+
+The support will be identified with several states:
+
+- __*Active*__: The release is a stable branch that is currently supported and accepting patches.
+- __*End of Life*__: The release branch is no longer supported, and no new patches will be accepted.
+
+Currently, we are planning to provide maintenance support for the two latest release branches **(N, N-1)**. Examples: If the latest release is 1.10, we will support 1.9 and 1.10.  If the latest release is 2.0, then we will support 2.0, and the most recent release branch of 1.X. 
+
+Maintenance releases will be supported for upto 1 year until the end of life is announced for the branch i.e. v1.9 **(N-2)** will be EOL when v1.11 (N) is released.
```
chrisohaver, line 29: Is a "maintenance release" same as a "minor release"

Issue #6092, opened 2023-05-11
Support extra data in dnstap payload


swapmat:
**What would you like to be added**:

dnstap supports sending `extra` bytes in the payload. This is defined in the [dnstap.proto](https://github.com/dnstap/dnstap.pb/blob/master/dnstap.proto#L40)
However, the dnstap plugin and forward plugin which uses dnstap do not allow sending this extra field. Hence, this enhancement request to update the dnstap plugin to support sending the extra bytes.

There is no limit to the size of the `extra` but we can cap it based on a config.

**Why is this needed**:

We have a use-case where we want to add some extra meta-data/context in the dns log to be sent to the dnstap receiver.

Tantalor93:
Hello, can you more describe your use case? What would you like to see in the extra section?

chrisohaver:
IIUC, the use case is to add the dnstap `extra` field to the dnstap message, set to a CoreDNS metadata value that the user would define in the Corefile.  e.g. configured via an option like `extra METADATA`, where METADATA defines a CoreDNS metadata value.

Tantalor93:
so if I understand correctly, something like this
```
.:53 {
	dnstap /tmp/dnstap.sock {
  		identity my-dns-server1
  		version MyDNSServer-1.2.3
		extra forward/upstream
	}
	forward . 8.8.8.8
}
```
would add to dnstap `extra` field a value of `forward/upstream` [metadata](https://github.com/coredns/coredns/tree/master/plugin/forward#metadata) from forward plugin?

chrisohaver:
yes.  metadata plugin also required ... e.g. 

```
.:53 {
        metadata
	dnstap /tmp/dnstap.sock {
  		identity my-dns-server1
  		version MyDNSServer-1.2.3
		extra forward/upstream
	}
	forward . 8.8.8.8
}
```

Tantalor93:
ok, makes sense! 👍

Nitesh639:
@Tantalor93, can I work on this issue?

chrisohaver:
I think this would be a good first issue since it doesn't require deep DNS protocol knowledge.

Nitesh639:
@chrisohaver Thank you for suggesting this issue as a good first task. I'm new to the organization and excited to contribute. I agree that it's a great choice since it doesn't require deep DNS protocol knowledge. I want to start working on it.

Tantalor93:
@Nitesh639 sure, go ahead 👍

[...]

PR #6226, opened 2023-07-25
Plugin dnstap: add support for "extra" field in payload

chenyuheng:
As proposed in #6092 , this PR adds supports to customize "extra" field in dnstap payload, the field could be set in the config file.

The "extra" field in defined in [dnstap.proto](https://github.com/dnstap/dnstap.pb/blob/master/dnstap.proto#L40).

#6092 

The plugin README: https://github.com/coredns/coredns/blob/master/plugin/dnstap/README.md

No

### DIFF
--- a/plugin/dnstap/README.md
+++ b/plugin/dnstap/README.md
@@ -18,6 +18,7 @@ Every message is sent to the socket as soon as it comes in, the *dnstap* plugin 
 dnstap SOCKET [full] {
   [identity IDENTITY]
   [version VERSION]
+  [extra EXTRA]
   [skipverify]
 }
 ~~~
@@ -60,6 +61,14 @@ Log to a socket, overriding the default identity and version.
 dnstap /tmp/dnstap.sock {
   identity my-dns-server1
   version MyDNSServer-1.2.3
+}
+~~~
+
+Log to a socket, customize the "extra" field in dnstap payload.
+
+~~~ txt
+dnstap /tmp/dnstap.sock {
+  extra forward/upstream
 }
 ~~~
 
--- a/plugin/dnstap/README.md
+++ b/plugin/dnstap/README.md
@@ -64,11 +64,13 @@ dnstap /tmp/dnstap.sock {
 }
 ~~~
 
-Log to a socket, customize the "extra" field in dnstap payload.
+Log to a socket, customize the "extra" field in dnstap payload. You may use metadata provided by other plugins in the extra field.
 
 ~~~ txt
+forward . 8.8.8.8
+metadata
 dnstap /tmp/dnstap.sock {
-  extra forward/upstream
+  extra "upstream: {/forward/upstream}"
 }
 ~~~
 
@@ -133,7 +135,10 @@ And then in your plugin:
 
 ~~~ go
 import (
-  github.com/coredns/coredns/plugin/dnstap/msg
+  "github.com/coredns/coredns/plugin/dnstap/msg"
+  "github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/coredns/coredns/request"
+
   tap "github.com/dnstap/golang-dnstap"
 )
 
@@ -147,7 +152,12 @@ func (x ExamplePlugin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dn
             q.QueryMessage = buf
         }
         msg.SetType(q, tap.Message_CLIENT_QUERY)
+        
+        // if no metadata interpretation is needed, just send the message
         tapPlugin.TapMessage(q)
+
+        // OR: to interpret the metadata in "extra" field, give more context info
+        tapPlugin.TapMessageWithMetadata(q, ctx, request.Request{W: w, Req: query}, dnstest.NewRecorder(w))
     }
     // ...
 }
--- a/plugin/dnstap/README.md
+++ b/plugin/dnstap/README.md
@@ -136,7 +136,7 @@ And then in your plugin:
 ~~~ go
 import (
   "github.com/coredns/coredns/plugin/dnstap/msg"
-	"github.com/coredns/coredns/request"
+  "github.com/coredns/coredns/request"
 
   tap "github.com/dnstap/golang-dnstap"
 )
--- a/plugin/dnstap/README.md
+++ b/plugin/dnstap/README.md
@@ -27,6 +27,7 @@ dnstap SOCKET [full] {
 * `full` to include the wire-format DNS message.
 * **IDENTITY** to override the identity of the server. Defaults to the hostname.
 * **VERSION** to override the version field. Defaults to the CoreDNS version.
+* **EXTRA** to define "extra" field in dnstap payload, [metadata](../metadata/) replacement available here.
 * `skipverify` to skip tls verification during connection. Default to be secure
 
 ## Examples
--- a/plugin/dnstap/README.md
+++ b/plugin/dnstap/README.md
@@ -136,7 +136,6 @@ And then in your plugin:
 ~~~ go
 import (
   "github.com/coredns/coredns/plugin/dnstap/msg"
-  "github.com/coredns/coredns/plugin/pkg/dnstest"
 	"github.com/coredns/coredns/request"
 
   tap "github.com/dnstap/golang-dnstap"
@@ -157,7 +156,7 @@ func (x ExamplePlugin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dn
         tapPlugin.TapMessage(q)
 
         // OR: to interpret the metadata in "extra" field, give more context info
-        tapPlugin.TapMessageWithMetadata(q, ctx, request.Request{W: w, Req: query}, dnstest.NewRecorder(w))
+        tapPlugin.TapMessageWithMetadata(ctx, q, request.Request{W: w, Req: query})
     }
     // ...
 }
--- a/plugin/dnstap/handler.go
+++ b/plugin/dnstap/handler.go
@@ -26,20 +26,20 @@ type Dnstap struct {
 	ExtraFormat       string
 }
 
-// TapMessage sends the message m to the dnstap interface, without interpreting the "Extra" field using metadata.
+// TapMessage sends the message m to the dnstap interface, without populating "Extra" field.
 func (h Dnstap) TapMessage(m *tap.Message) {
 	h.TapMessageWithMetadata(nil, m, request.Request{})
 }
 
-// TapMessageWithMetadata sends the message m to the dnstap interface, with "Extra" field being interpreted by the provided metadata context.
+// TapMessageWithMetadata sends the message m to the dnstap interface, with "Extra" field being populated.
 func (h Dnstap) TapMessageWithMetadata(ctx context.Context, m *tap.Message, state request.Request) {
 	t := tap.Dnstap_MESSAGE
 	extraStr := h.ExtraFormat
-	if ctx != nil {
-		extraStr = h.repl.Replace(ctx, state, nil, extraStr)
-	}
 	var extra []byte
 	if extraStr != "" {
+		if ctx != nil {
+			extraStr = h.repl.Replace(ctx, state, nil, extraStr)
+		}
 		extra = []byte(extraStr)
 	}
 	dt := &tap.Dnstap{
--- a/plugin/dnstap/handler.go
+++ b/plugin/dnstap/handler.go
@@ -28,28 +28,26 @@ type Dnstap struct {
 
 // TapMessage sends the message m to the dnstap interface, without populating "Extra" field.
 func (h Dnstap) TapMessage(m *tap.Message) {
-	h.TapMessageWithMetadata(context.TODO(), m, request.Request{})
+	if h.ExtraFormat == "" {
+		h.tapWithExtra(m, nil)
+	} else {
+		h.tapWithExtra(m, []byte(h.ExtraFormat))
+	}
 }
 
 // TapMessageWithMetadata sends the message m to the dnstap interface, with "Extra" field being populated.
 func (h Dnstap) TapMessageWithMetadata(ctx context.Context, m *tap.Message, state request.Request) {
+	if h.ExtraFormat == "" {
+		h.tapWithExtra(m, nil)
+		return
+	}
+	extraStr := h.repl.Replace(ctx, state, nil, h.ExtraFormat)
+	h.tapWithExtra(m, []byte(extraStr))
+}
+
+func (h Dnstap) tapWithExtra(m *tap.Message, extra []byte) {
 	t := tap.Dnstap_MESSAGE
-	extraStr := h.ExtraFormat
-	var extra []byte
-	if extraStr != "" {
-		if ctx != context.TODO() {
-			extraStr = h.repl.Replace(ctx, state, nil, extraStr)
-		}
-		extra = []byte(extraStr)
-	}
-	dt := &tap.Dnstap{
-		Type:     &t,
-		Message:  m,
-		Identity: h.Identity,
-		Version:  h.Version,
-		Extra:    extra,
-	}
-	h.io.Dnstap(dt)
+	h.io.Dnstap(&tap.Dnstap{Type: &t, Message: m, Identity: h.Identity, Version: h.Version, Extra: extra})
 }
 
 func (h Dnstap) tapQuery(ctx context.Context, w dns.ResponseWriter, query *dns.Msg, queryTime time.Time) {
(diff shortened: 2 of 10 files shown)


chenyuheng:
This comment shows the result of manually testting for this change.

Note that the current dnstap command line tool [won't output "extra" field](https://github.com/dnstap/golang-dnstap/blob/ebb538e7d351a58861a8f348491828214a1d8db2/YamlFormat.go#L112C1-L113C1), so I tested this change with my edited dnstap command line tool, I will try to make another PR to update the change to the [golang-dnstap repository](https://github.com/dnstap/golang-dnstap) later.

With Corefile:
```
.:8053 {
	dnstap /tmp/dnstap.sock {
		identity test_1
		version test0.1
		extra EXTRA_FIELD_TEST_233
	}
	forward . 8.8.8.8
}
```

The dnstap command line tool will have:

image

plugin/dnstap/handler.go
```diff
@@ -20,12 +20,20 @@ type Dnstap struct {
...
 // TapMessage sends the message m to the dnstap interface.
 func (h Dnstap) TapMessage(m *tap.Message) {
 	t := tap.Dnstap_MESSAGE
-	h.io.Dnstap(&tap.Dnstap{Type: &t, Message: m, Identity: h.Identity, Version: h.Version})
+	dt := &tap.Dnstap{
+		Type: &t,
+		Message: m,
+		Identity: h.Identity,
+		Version: h.Version,
+		Extra: h.Extra,
```
rdrozhdzh: what you implemented here is assigning a preconfigured text to `Extra` (which can be useful as well), whereas https://github.com/coredns/coredns/issues/6092 suggests extracting and assigning value of preconfigured metadata label.

The metadata value is normally created by other plugin, e.g. if you specify label "forward/upstream", it is expected that plugin 'forward' should provide this value. The plugin which provides metadata can generate different values for different queries, and this is the key difference from what you have implemented.

Please get familiar with plugin metadata https://github.com/coredns/coredns/tree/master/plugin/metadata as well as the package https://github.com/coredns/coredns/tree/master/plugin/pkg/replacer which can make your work easier

I also recommend to look at plugins 'rewrite' and 'log' as examples of plugins which read metadata from other plugins.

plugin/dnstap/README.md
```diff
@@ -63,6 +64,14 @@ dnstap /tmp/dnstap.sock {
 }
 ~~~
 
+Log to a socket, customize the "extra" field in dnstap payload.
+
+~~~ txt
+dnstap /tmp/dnstap.sock {
+  extra forward/upstream
```
rdrozhdzh: I think, this should be configured as `{/forward/upstream}` if we want to specify metadata label. At least, other plugins (rewrite, log) use this format, and this is what package https://github.com/coredns/coredns/tree/master/plugin/pkg/replacer expects.

The text `forward/upstream` still can be interpreted as a constant string (literal).

plugin/dnstap/handler.go
chenyuheng: Thanks for your kind review! I have introduced the `replacer` package in my new commits.

Since the replacer requires context, status (`request.Request`) and rr (`*dnstest.Recorder`), I have added a new function called `TapMessageWithMetadata` to receive these context objects and process the metadata replacement accordingly. 

The old function `TapMessage` has not been removed, as external plugins may have dependencies on it.

plugin/dnstap/README.md
```diff
@@ -133,7 +135,10 @@ And then in your plugin:
 
 ~~~ go
 import (
-  github.com/coredns/coredns/plugin/dnstap/msg
+  "github.com/coredns/coredns/plugin/dnstap/msg"
+  "github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/coredns/coredns/request"
```
rdrozhdzh: inconsistent tabbing

plugin/dnstap/README.md
```diff
@@ -147,7 +152,12 @@ func (x ExamplePlugin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dn
             q.QueryMessage = buf
         }
         msg.SetType(q, tap.Message_CLIENT_QUERY)
+        
+        // if no metadata interpretation is needed, just send the message
         tapPlugin.TapMessage(q)
+
+        // OR: to interpret the metadata in "extra" field, give more context info
+        tapPlugin.TapMessageWithMetadata(q, ctx, request.Request{W: w, Req: query}, dnstest.NewRecorder(w))
```
rdrozhdzh: The recorder is needed for intercepting DNS response, and it should be created before writing DNS response.

Creating and passing recorder this way doesn't make sense. It will always have empty DNS response.

plugin/dnstap/handler.go
```diff
@@ -27,17 +27,17 @@ type Dnstap struct {
...
-// TapMessage sends the message m to the dnstap interface.
+// TapMessage sends the message m to the dnstap interface, without interpreting the "Extra" field using metadata.
 func (h Dnstap) TapMessage(m *tap.Message) {
-	h.TapMessageWithMetadata(m, nil, nil, nil)
+	h.TapMessageWithMetadata(m, nil, request.Request{}, nil)
 }
 
 // TapMessageWithMetadata sends the message m to the dnstap interface, with "Extra" field being interpreted by the provided metadata context.
-func (h Dnstap) TapMessageWithMetadata(m *tap.Message, ctx context.Context, state *request.Request, rr *dnstest.Recorder) {
+func (h Dnstap) TapMessageWithMetadata(m *tap.Message, ctx context.Context, state request.Request, rr *dnstest.Recorder) {
```
rdrozhdzh: I would suggest to remove the `rr *dnstest.Recorder` at all, and always pass `nil` to `Replacer`. I believe this parameter creates more inconveniences than benefits.

The recorder parameter is needed only for handling several standard labels like
https://github.com/coredns/coredns/blob/b95bc0a7cac365b79d46288e1433a5c5ef903e02/plugin/pkg/replacer/replacer.go#L87-L111

This code can safely work with `nil`, i.e. it will not panic.

The reason why it's not so important to support these DNS response labels (as well as other DNS labels) is because DNSTAP already provides an alternative standardized way for transferring this data.

Regarding parameter `state`, this is more commonly used data type, and we can keep it. This should not be a big problem for developers of other plugins to provide it. However, if we decide to not support the standard DNS labels (as explained above), we can eliminate this parameter as well. You can pass empty `Request{}` to the replacer  (this needs to be tested with all standard labels to avoid panics)

The only important parameter here is Context, which provides access to metadata.

I would appreciate a feedback from community regarding support of standard DNS labels (do we really need them?) for DNSTAP extra section, i.e.
https://github.com/coredns/coredns/blob/b95bc0a7cac365b79d46288e1433a5c5ef903e02/plugin/pkg/replacer/replacer.go#L37-L57

plugin/forward/dnstap.go
```diff
@@ -1,20 +1,22 @@
...
 	"github.com/coredns/coredns/request"
+	"github.com/coredns/coredns/plugin/pkg/dnstest"
 
 	tap "github.com/dnstap/golang-dnstap"
 	"github.com/miekg/dns"
 )
 
 // toDnstap will send the forward and received message to the dnstap plugin.
-func toDnstap(f *Forward, host string, state request.Request, opts proxy.Options, reply *dns.Msg, start time.Time) {
+func toDnstap(f *Forward, host string, state request.Request, opts proxy.Options, reply *dns.Msg, start time.Time, ctx context.Context, w dns.ResponseWriter) {
```
rdrozhdzh: context should be the first parameter

plugin/dnstap/handler.go
rdrozhdzh: context should be the first parameter

rdrozhdzh (commented):
Please add also more unit tests to make sure the new functionality works as expected

plugin/dnstap/handler.go
chenyuheng: Thanks! I have updated these codes in my new commits. I will update the unit tests tomorrow.

For the `state` parameter, maybe we can just keep the function as:
```go
func (h Dnstap) TapMessageWithMetadata(ctx context.Context, m *tap.Message, state request.Request) 
```
Leave the choice to plugin developers, if they think their plugins does not need to support these standard labels, they simply just pass the empty `State{}`

plugin/dnstap/README.md
```diff
@@ -18,6 +18,7 @@ Every message is sent to the socket as soon as it comes in, the *dnstap* plugin
 dnstap SOCKET [full] {
   [identity IDENTITY]
   [version VERSION]
+  [extra EXTRA]
```
rdrozhdzh, line 21: please add the explanation of this parameter, see other parameters below as example

plugin/dnstap/handler.go
```diff
@@ -15,20 +17,42 @@ import (
...
 
 	// IncludeRawMessage will include the raw DNS message into the dnstap messages if true.
 	IncludeRawMessage bool
 	Identity          []byte
 	Version           []byte
+	ExtraFormat       string
 }
 
-// TapMessage sends the message m to the dnstap interface.
+// TapMessage sends the message m to the dnstap interface, without interpreting the "Extra" field using metadata.
```
rdrozhdzh: , without populating "Extra" field.

plugin/dnstap/handler.go
```diff
@@ -15,20 +17,42 @@ import (
...
 func (h Dnstap) TapMessage(m *tap.Message) {
+	h.TapMessageWithMetadata(nil, m, request.Request{})
+}
+
+// TapMessageWithMetadata sends the message m to the dnstap interface, with "Extra" field being interpreted by the provided metadata context.
+func (h Dnstap) TapMessageWithMetadata(ctx context.Context, m *tap.Message, state request.Request) {
 	t := tap.Dnstap_MESSAGE
-	h.io.Dnstap(&tap.Dnstap{Type: &t, Message: m, Identity: h.Identity, Version: h.Version})
+	extraStr := h.ExtraFormat
+	if ctx != nil {
```
rdrozhdzh: please move this `if` under `if extraStr != ""` below. We don't want spending resources on Replacer if `extra` was not configured. So, only the users who configured `extra` will pay for it.
Because I see this method is called now even from internal functions like `tapQuery()`

plugin/dnstap/handler_test.go
```diff
@@ -21,24 +22,32 @@ func testCase(t *testing.T, tapq, tapr *tap.Message, q, r *dns.Msg) {
 			return 0, w.WriteMsg(r)
 		}),
 		io: &w,
+		ExtraFormat: "extra_field_{/metadata/test}_{type}_{name}_{class}_{proto}_{size}_{remote}_{port}_{local}",
```
rdrozhdzh: better to pass `ExtraFormat` as parameter instead of hardcoding it here

plugin/dnstap/handler_test.go
```diff
@@ -82,3 +100,24 @@ func testMessage() *tap.Message {
...
+	h := Dnstap{
+		Next: test.HandlerFunc(func(_ context.Context,
+			w dns.ResponseWriter, r *dns.Msg) (int, error) {
+			return 0, w.WriteMsg(r)
+		}),
+		io: &w,
+		ExtraFormat: "extra_field_{/metadata/test}_{type}_{name}_{class}_{proto}_{size}_{remote}_{port}_{local}",
+	}
+	h.TapMessage(tapq.Message)
+}
```
rdrozhdzh: bad EOF

plugin/dnstap/setup_test.go
```diff
@@ -24,27 +25,33 @@ func TestConfig(t *testing.T) {
...
+		{"dnstap tcp://example.com:6000", false, []results{{"example.com:6000", false, "tcp", []byte(hostname), []byte("-"), ""}}},
+		{"dnstap", true, []results{{"fail", false, "tcp", []byte(hostname), []byte("-"), ""}}},
+		{"dnstap dnstap.sock full {\nidentity NAME\nversion VER\n}\n", false, []results{{"dnstap.sock", true, "unix", []byte("NAME"), []byte("VER"), ""}}},
+		{"dnstap dnstap.sock full {\nidentity NAME\nversion VER\nextra EXTRA\n}\n", false, []results{{"dnstap.sock", true, "unix", []byte("NAME"), []byte("VER"), "EXTRA"}}},
+		{"dnstap dnstap.sock {\nidentity NAME\nversion VER\nextra EXTRA\n}\n", false, []results{{"dnstap.sock", false, "unix", []byte("NAME"), []byte("VER"), "EXTRA"}}},
+		{"dnstap {\nidentity NAME\nversion VER\nextra EXTRA\n}\n", true, []results{{"fail", false, "tcp", []byte("NAME"), []byte("VER"), "EXTRA"}}},
 		{`dnstap dnstap.sock full {
                 identity NAME
                 version VER
+				extra EXTRA
```
rdrozhdzh: here and below inconsistent tabs

plugin/dnstap/writer.go
```diff
@@ -13,6 +15,7 @@ import (
 type ResponseWriter struct {
 	queryTime time.Time
 	query     *dns.Msg
+	ctx       *context.Context
```
rdrozhdzh: do you really need a pointer here?

plugin/pkg/replacer/replacer_test.go
```diff
@@ -247,6 +247,8 @@ func TestLabels(t *testing.T) {
 
 	for lbl := range labels {
 		repl := replacer.Replace(ctx, state, w, lbl)
+		// test empty state won't panic
+		replacer.Replace(ctx, request.Request{}, w, lbl)
```
rdrozhdzh: you can pass `nil` here as the 3rd parameter, to test if it doesn't panic as well

rdrozhdzh (commented):
In general looks good, see minor comments

plugin/dnstap/writer.go
chenyuheng: The ctx field is needed in the WriteMsg function below to call TapMessageWithMetadata.

Or do you mean to use the context struct directly without using a pointer? Like ctx context.Context? I think using a pointer can help to avoid an extra copy of the context structure, thus reducing memory usage.

plugin/dnstap/handler.go
```diff
@@ -15,20 +17,42 @@ import (
...
+}
+
+// TapMessageWithMetadata sends the message m to the dnstap interface, with "Extra" field being populated.
+func (h Dnstap) TapMessageWithMetadata(ctx context.Context, m *tap.Message, state request.Request) {
 	t := tap.Dnstap_MESSAGE
-	h.io.Dnstap(&tap.Dnstap{Type: &t, Message: m, Identity: h.Identity, Version: h.Version})
+	extraStr := h.ExtraFormat
+	var extra []byte
+	if extraStr != "" {
+		if ctx != context.TODO() {
```
rdrozhdzh: this looks not reliable, TODO() may come from outside as well.

I would suggest splitting this to 3 functions like
```go
func (h Dnstap) TapMessage(m *tap.Message) {
  h.tapWithExtra(m, nil)
}
func (h Dnstap) TapMessageWithMetadata(ctx context.Context, m *tap.Message, state request.Request) {
  if h.ExtraFormat == "" {
    h.tapWithExtra(m, nil)
    return
  }
  extraStr := h.repl.Replace(ctx, state, nil, h.ExtraFormat)
  h.tapWithExtra(m, []byte(extraStr))
}
func (h Dnstap) tapWithExtra((m *tap.Message, extra []byte) {
  t := tap.Dnstap_MESSAGE
  h.io.Dnstap(&tap.Dnstap{Type: &t, Message: m, Identity: h.Identity, Version: h.Version, Extra: extra})
}
```

plugin/dnstap/writer.go
rdrozhdzh: Yes, I mean storing context struct directly. I don't think this is critical memory overhead. This is like storing/passing pointers to strings everywhere instead of storing/passing strings.
Also, pointers are managed by garbage collector subsystem which may spend some extra memory/CPU on serving pointers.

plugin/pkg/replacer/replacer_test.go
```diff
@@ -247,6 +247,10 @@ func TestLabels(t *testing.T) {
 
 	for lbl := range labels {
 		repl := replacer.Replace(ctx, state, w, lbl)
+		// test empty state and/or recorder won't panic
+		replacer.Replace(ctx, request.Request{}, w, lbl)
+		replacer.Replace(ctx, state, nil, lbl)
```
rdrozhdzh: I mean, you need to keep the original call
```go
  repl := replacer.Replace(ctx, state, w, lbl)
```
and the second call with empty Request and Recorder
```go
  repl := replacer.Replace(ctx, request.Request{}, nil, lbl)
```
Four calls looks like overkill. Anyway, depending on `lbl` it will check either Request{} or Recorder.

rdrozhdzh (approved):
looks good to me

dmachard:
@chenyuheng  FYI the  [DNS-collector ](https://github.com/dmachard/go-dnscollector ) tool supports the extra field, any contributions will be appreciated.

PR #6227, opened 2023-07-26
removed needless condition

b00tkitism:
I've removed needless condition
No issues, but optimizer code
No changes
No


chrisohaver (changes requested):
This changes behavior, resulting in not returning an error message unless the last call to Notify is a failure.

PR #6228, opened 2023-07-26
Removed unnecessary conditions

b00tkitism:
I've removed unnecessary conditions
It just make code more optimize I think
nothing needs to change
No

### DIFF
--- a/plugin/dnstap/io.go
+++ b/plugin/dnstap/io.go
@@ -62,16 +62,14 @@ func (d *dio) dial() error {
 			Timeout: d.tcpTimeout,
 		}
 		conn, err = tls.DialWithDialer(dialer, "tcp", d.endpoint, config)
-		if err != nil {
-			return err
-		}
 	} else {
 		conn, err = net.DialTimeout(d.proto, d.endpoint, d.tcpTimeout)
-		if err != nil {
-			return err
-		}
 	}
 
+	if err != nil {
+		return err
+	}
+	
 	if tcpConn, ok := conn.(*net.TCPConn); ok {
 		tcpConn.SetWriteBuffer(tcpWriteBufSize)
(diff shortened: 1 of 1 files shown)


SuperQ:

chrisohaver:
I don't think this adds any _optimizations_ - just code style preference.

I actually prefer the code as it was before the suggested changes.

plugin/dnstap/io.go
```diff
@@ -62,16 +62,14 @@ func (d *dio) dial() error {
...
 		}
 		conn, err = tls.DialWithDialer(dialer, "tcp", d.endpoint, config)
-		if err != nil {
-			return err
-		}
 	} else {
 		conn, err = net.DialTimeout(d.proto, d.endpoint, d.tcpTimeout)
-		if err != nil {
-			return err
-		}
```
chrisohaver, line 72: I prefer the error checks here, since it makes it immediately clear that the errors are handled.

plugin/dnstap/io.go
```diff
@@ -103,9 +101,7 @@ func (d *dio) close() { close(d.quit) }
 func (d *dio) write(payload *tap.Dnstap) error {
 	if d.enc == nil {
 		atomic.AddUint32(&d.dropped, 1)
-		return nil
-	}
-	if err := d.enc.writeMsg(payload); err != nil {
+	} else if err := d.enc.writeMsg(payload); err != nil {
```
chrisohaver, line 104: This change does not appear to present any obvious optimization.

I prefer the early return and separate if instead of else-if.

chrisohaver (changes requested):

SuperQ:
I agree with @chrisohaver. This makes the code less readable.

PR #6230, opened 2023-07-26
implement Name function on secondary plugin

diogenxs:
implement Name() func of handler interface on secondary plugin

[prometheus metric](https://coredns.io/plugins/metrics/) `coredns_plugin_enabled` shows zones loaded with `secondary` as `file`

NA

NA


Tantalor93 (approved):
lgtm, thanks!

PR #6231, opened 2023-07-28
update outdated blog links in documents

chenyuheng:
The old blog site (blog.coredns.io) is moved to coredns.io/blog and the old domain name is not working.

no


chrisohaver (approved):
thanks!

PR #6235, opened 2023-08-01
Fix typos

hezhizhen:

Fixed some typos.

It may be worth noting that `dnsutil.MaximumDefaulTTL` is renamed as `dnsutil.MaximumDefaultTTL`.

### DIFF
--- a/plugin/pkg/dnsutil/ttl.go
+++ b/plugin/pkg/dnsutil/ttl.go
@@ -20,7 +20,7 @@ func MinimalTTL(m *dns.Msg, mt response.Type) time.Duration {
 		return MinimalDefaultTTL
 	}
 
-	minTTL := MaximumDefaulTTL
+	minTTL := MaximumDefaultTTL
 	for _, r := range m.Answer {
 		if r.Header().Ttl < uint32(minTTL.Seconds()) {
 			minTTL = time.Duration(r.Header().Ttl) * time.Second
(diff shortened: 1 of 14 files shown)


plugin/pkg/dnsutil/ttl.go
```diff
@@ -47,6 +47,6 @@ func MinimalTTL(m *dns.Msg, mt response.Type) time.Duration {
 const (
 	// MinimalDefaultTTL is the absolute lowest TTL we use in CoreDNS.
 	MinimalDefaultTTL = 5 * time.Second
-	// MaximumDefaulTTL is the maximum TTL was use on RRsets in CoreDNS.
-	MaximumDefaulTTL = 1 * time.Hour
+	// MaximumDefaultTTL is the maximum TTL was use on RRsets in CoreDNS.
+	MaximumDefaultTTL = 1 * time.Hour
```
chrisohaver: This change will be backward incompatible for any external plugins that use these public constants.

hezhizhen: How about marking as TODO and revert them first?

chrisohaver:
Added to the 1.12 milestone, due to public constant name change: `dnsutil.MaximumDefaulTTL`.

hezhizhen:
> Added to the 1.12 milestone, due to public constant name change: `dnsutil.MaximumDefaulTTL`.

Now that this PR is added to 1.12, should I revert these name changes or not?

chrisohaver:
> > Added to the 1.12 milestone, due to public constant name change: `dnsutil.MaximumDefaulTTL`.
> 
> Now that this PR is added to 1.12, should I revert these name changes or not?

You can, and I will remove it from 1.12 milestone to go out in next patch.

hezhizhen:
> > > Added to the 1.12 milestone, due to public constant name change: `dnsutil.MaximumDefaulTTL`.
> > 
> > 
> > Now that this PR is added to 1.12, should I revert these name changes or not?
> 
> You can, and I will remove it from 1.12 milestone to go out in next patch.

I've reverted this change and added a TODO.

PTAL.

chrisohaver (approved):

Issue #6243, opened 2023-08-07
plugin/reload: potential race conditon with imported files


cbarbian-sap:
**What happened**:

If the `Corefile` uses imported files, a race condition may happen during reload, in the sense that the restarted instance may not run with the most recent configuration. Explanation attempt:

As of now, the `InstanceStartupEvent` event hook in https://github.com/coredns/coredns/blob/master/plugin/reload/reload.go#L76 reads and remembers an initial hash value of the current parsed `Corefile`. Here, 'current' means that it indeed (and correctly IMHO) uses the `Corefile` content that was loaded by current instance's `Start()` or `Restart()` methods. However, any imported files are read at execution time of the time of the reload's event `hook()` function, when calling the `parse()` function. 

This can create a race condition. Imagine the following sequence:
1. A reload is triggered for some reason, e.g. by a change of the Corefile or one of the imported files
2. The currently running reload hook does its job, and triggers a `Restart()` of the instance (https://github.com/coredns/coredns/blob/master/plugin/reload/reload.go#L108)
3. Now imagine, another change happens in one of the imported files, during or after executing the `Corefile` (https://github.com/coredns/caddy/blob/master/caddy.go#L246), but before emitting the `InstanceStartupEvent` (https://github.com/coredns/caddy/blob/master/caddy.go#L264). Then some or all of the plugins might use the previous, outdated content of the imported file, but the new reload handler started in the `hook()` function will remember the SHA of the current content. As a consequence, no further reload will happen (unless another change occurs), but plugins might run with a non-up-to-date configuration.

IMHO the actual problem is that the `caddy.Instance`  stores the content of the current `Corefile`, but not the content of potentially imported files. Probably, changing this would be a bigger effort.

As an alternative solution, I created a pull request (https://github.com/coredns/coredns/pull/6244).

**What you expected to happen**:

Such a race condition should not happen.

**How to reproduce it (as minimally and precisely as possible)**:

Import a frequently changing file, and let reload run at a high frequency; and wait ...

**Anything else we need to know?**: no

**Environment**:

- the version of CoreDNS: 1.10.1, master
- Corefile: 
    ```
    .:62529 {
        bind 127.0.0.1
        errors
        log .
	    prometheus
        kubernetes cluster.local in-addr.arpa ip6.arpa {
            kubeconfig /var/folders/qq/3flcldfx1x3gvszg3xfv_13m0000gn/T/2860259645/kubeconfig
            pods insecure
[...]

PR #6244, opened 2023-08-07
plugin/reload: avoid race conditions if imported files change during reload / fix issues with sigusr1 reload

cbarbian-sap:

Try to fix race condition described in https://github.com/coredns/coredns/issues/6243

https://github.com/coredns/coredns/issues/6243

None I think

Hopefully not


chrisohaver:
Thanks for the fix!

Adding a test may be a pain, but I think one approach would be to craft an ad-hoc test plugin with a locked OnShutdown function that the executing test can unlock (i.e. pausing the reload to induce the race condition).

cbarbian-sap:
I have added a corresponding test case. As you suggested, it uses a plugin (called `dummy`) which basically reads one integer argument from the configuration;  such as
```
dummy 1
```
In addition the plugin registers an `OnShutdown` callback, which updates that imported file by incrementing the integer argument, not forever, but the first couple of times. This ensures that the problematic race condition reliably occurs
(I think the locking thing you proposed is not even required to make this happen).

The test then starts up a coredns instance with a simple Corefile:
```
.:0 {
  import *.conf
  reload 2s 1s
}
```
and kicks off a change in the imported file, and waits until the plugin is finally running with the latest configuration version.

With the upstream logic, the test fails, with the changed logic in the PR it succeeds.

cbarbian-sap:
However I do not completely understand why the test is failing.
Running it individually **always** works:
```
go test -count 1 -v -run ^TestReloadWithImports$ ./test
```
However running the package tests as a whole never works. Could it be the case that this is because the global plugins (in that case, reload) are then shared across the various tests in the package, which is problematic if they carry a state? 

If so, what could I do to overcome this ?

cbarbian-sap:
It really seems to be impossible to use certain plugins (at least the reload plugin is affected) in that shared `./test` package in several test cases resp. in different caddy server instances (e.g. because the callbacks used by the plugin are registered only once, with the very first instance using it).

At the moment the only test case using the reload plugin in `./test` seems to be `readme_test.go`, and that one only instantiates the plugin, but does not really perform any functional tests. This is why it seems to work at the moment.

Anyway, to overcome the problem, I moved the test introduced by this PR into a separate package, `./test/reload`....

cbarbian-sap:
Following the discussion in #6243 I created a PR on coredns/caddy: https://github.com/coredns/caddy/pull/4, and refactored the reload logic to make use of it.

NOTES:
- Because of that PR not yet merged resp. released, the code in this PR currently uses my fork (via replace in go.mod):
  ```
  replace github.com/coredns/caddy => github.com/cbarbian-sap/caddy v0.0.1
  ```
- This PR now also addresses #6263.

PR #6245, opened 2023-08-08
New Plugin:  QUIC proxy [RFC9250]

### LINKED ISSUE #5583: Support For DoQ (DNS over QUIC)

jaehnri:

This PR adds support for QUIC proxying in CoreDNS 

Closes https://github.com/coredns/coredns/issues/5583

README documentation for the plugin was added, but it is possible that there are more places to update.

No.


plugin/quic/quic.go
```diff
@@ -0,0 +1,142 @@
...
+	span = ot.SpanFromContext(ctx)
+	list := q.list()
+	deadline := time.Now().Add(defaultTimeout)
+
+	for time.Now().Before(deadline) {
+		if i >= len(list) {
+			// reached the end of list without any answer
+			if ret != nil {
+				// write empty response and finish
+				w.WriteMsg(ret)
```
chrisohaver: I think that writing an empty dns message to client would result in client-side parsing/validation errors. I think the proper response id with a SERVFAIL response code, with question correctly set.  Returning a servfail and error instead of writing a response would work too (base server code writes an error response to client when ServeDNS returns a SERVFAIL code).

```suggestion
				return dns.RcodeServerFailure, <some-error>
```

jaehnri: Agreed and fixed. I got this code from the `grpc` plugin, should we port this fix there too?
https://github.com/coredns/coredns/blob/b5e6291115d1e60fed561c64d70341b354e69504/plugin/grpc/grpc.go#L51-L54

plugin/quic/policy.go
```diff
@@ -0,0 +1,68 @@
+package quic
```
jaehnri, line 1: This policy selection code has been repeated across several plugins now (`forward`, `grpc`, `quic`). Should we unify it? If so, where should it be? `plugin/pkg`?

test/quic_test.go
```diff
@@ -117,6 +121,38 @@ func TestQUICProtocolError(t *testing.T) {
 	}
 }
 
+func TestQUICPlugin(t *testing.T) {
```
jaehnri, line 124: The plugin integration test ended up in the `test` folder. Unlike `forward`, setting up a mock `quic` server to answer a simple DNS query is not easy because you need not only to create a listener but also accept connections, streams, write, and close to the stream, **all manually**.

The best way I found to make it work and avoid flakiness was to use the `CoreDNSServerAndPorts` function to spin up a CoreDNS QUIC server and test the plugin against it.

plugin/quic/setup.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	if q.tlsConfig == nil {
+		q.tlsConfig = new(tls.Config)
+	}
+
+	if q.tlsServerName != "" {
+		q.tlsConfig.ServerName = q.tlsServerName
+	} else {
+		q.tlsConfig.InsecureSkipVerify = true
```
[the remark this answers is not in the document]
jaehnri: I'll have to find a way to circumvent this. `quic-go` doesn't work without `tls.Config` and I couldn't set the `tls_servername` using `plugin/tls` test certificates.

plugin/quic/policy.go
chrisohaver: That would be a good place. Thanks!

plugin/quic/quic.go
chrisohaver: > should we port this fix there too?

Probably, but not in this PR.

plugin/quic/setup.go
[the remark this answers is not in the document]
ghost: @jaehnri the plugin/dns test certs common name is empty, therefore the hostname cannot match. My suggestion would be to drop the tls_servername option, because it should never be needed (the tls package checks if the common name, subject alternative names or ip addresses contains the hosts name) and instead introduce the option tls_skip_verify. For test cases u can set the common name manually via the tlsConfig

chrisohaver:
@jaehnri, What is the state of this PR.  Is it pending a final review, or are there parts yet to complete?

qm3ster:
:horse:

houyuwushang:
@jaehnri, thank you for the original DoQ proxy implementation and the work captured here.

The inbound DoQ server requested by #5583 has since been completed by #6182, but outbound DoQ forwarding is still missing from current master. The tree has also evolved in a way that prevents this branch from being revived as-is: the `quic` directive is now used for inbound server configuration by #7296, and `forward` plus the shared proxy layer have changed substantially.

I opened #8473 to continue the outbound capability from current master as `forward` support for `quic://` upstreams. I used this PR as prior art and preserved credit for your original work in the replacement implementation, now available in #8474. Please let us know if you would like to resume this PR or collaborate on the replacement.

houyuwushang:
Closing as superseded by #8474, which has now merged and implements outbound DoQ support in `forward` via `quic://` upstreams. Thank you @jaehnri for the original implementation and prior art.

PR #6271, opened 2023-08-15
plugin/reload: make reload work after SIGUSR1

### LINKED ISSUE #6263: plugin/reload does not work with SIGUSR1 triggered reload
**What happened**:

It seems that the reload plugin stops to work after a reload was triggered by sending SIGUSR1 to the coredns process.

Reload plugin should not break due sending SIGUSR1.

Start a coredns server with the following Corefile:
```
.:1053 {
        reload 2s 1s
        erratic
        # forward . 8.8.8.8
}
```

Confirm that it works:
```
$ dig @localhost -p 1053 dns.google

# (snip)
;; QUESTION SECTION:
;dns.google.			IN	A

;; ANSWER SECTION:
dns.google.		0	IN	A	192.0.2.53

# (snip)
```

Ok fine, the `192.0.2.53` response comes from the erratic plugin.

Now trigger a reload by sending SIGUSR1 to the coredns process: `killall -USR1 coredns`. Log output:

```
[INFO] SIGUSR1: Reloading
[INFO] Reloading
[INFO] Reloading complete
```

Still fine. Now modify the Corefile to

```
.:1053 {
	reload 2s 1s
  	# erratic
	forward . 8.8.8.8
}
```

Log output:
```
[INFO] Reloading
[ERROR] Restart failed: getting old listener file: file tcp [::]:1053: use of closed network connection
[ERROR] plugin/reload: Corefile changed but reload failed: starting with listener file descriptors: getting old listener file: file tcp [::]:1053: use of closed network connection
```

And it indeed did not reload; running the above dig command again still returns `192.0.2.53`,  instead of the expected response `4.4.4.4` and `8.8.8.8` that would have been returned by the forward plugin.

From that point on, no further reload will work.

I tried to understand what happens. First of all it should be noted that the reload plugin registers an event hook (and an `OnFinalShutdown` callback) in its `setup` function: 

```
once.Do(func() {
	caddy.RegisterEventHook("reload", hook)
})
// re-register on finalShutDown as the instance most-likely will be changed
shutOnce.Do(func() {
	c.OnFinalShutdown(func() error {
		r.quit <- true
		return nil
	})
})
```
(see https://github.com/coredns/coredns/blob/master/plugin/reload/setup.go#L69C1-L78C4), and it does that only once.
Which by the way seems wrong to me for the callback, because that callback is then missing for all future instances that are created by the plugin upon reloading; but that's maybe a different story. 
In case of the event hook at least it seems to be correct and logical to do it just once.

BUT: the signal handler for the USR1 signal clears all event hooks before restarting the instance:
```
purgeEventHooks()

// Kick off the restart; our work is done
EmitEvent(InstanceRestartEvent, nil)
_, err = inst.Restart(caddyfileToUse)
if err != nil {
	restoreEventHooks(oldEventHooks)

	log.Printf("[ERROR] SIGUSR1: %v", err)
}
```
(see https://github.com/coredns/caddy/blob/master/sigtrap_posix.go#L84). It seems this was introduced around https://github.com/caddyserver/caddy/issues/2044. Which by the way makes that `EmitEvent(InstanceRestartEvent, nil)`  statement completely meaningless, because no hook will ever handle it, but that's maybe a different story, too.

Whatever, because of that, the reload `hook()` function (https://github.com/coredns/coredns/blob/master/plugin/reload/reload.go#L63) will not run, and therefore not spin up a new go routine (https://github.com/coredns/coredns/blob/master/plugin/reload/reload.go#L84) for the new instance created through the `Restart()` invocation in the signal handler.

So, when changing the Corefile afterwards,  the go routine that was started for the old instance is still active, and tries to `Restart()` the instance it knows; however that instance was already stopped through the `Restart()` done by the USR1 signal handler. Which - as far as I understand it - explains the error message 'use of closed network connection'.

Now, even if that `purgeEventHooks()` logic would not be in place, another bad thing would have happened, because then, a second reloading go routine would have been started, while the old one would still be running. Then I guess, the new one would do its reloading  job properly, but the old one would forever pollute the logs with error messages (and each reload would add another such zombie go routine).

**Environment**:

- the version of CoreDNS: 1.11.0
- Corefile: see above
- logs, if applicable: see above
- OS (e.g: `cat /etc/os-release`): linux, mac, does not matter
- Others: n/a

cbarbian-sap:

Try to fix #6263: it introduces two relevant changes:
1. Register the reload event hook (https://github.com/coredns/coredns/blob/master/plugin/reload/reload.go#L63) not only once (as it happens today, see https://github.com/coredns/coredns/blob/master/plugin/reload/setup.go#L69), but every time the `setup` function of the plugin is called.
This addresses the problem that the USR1 signal handler currently flushes all event hooks (https://github.com/coredns/caddy/blob/master/sigtrap_posix.go#L84).
For the moment, a workaround (https://github.com/cbarbian-sap/coredns/blob/a2f8cf8f35f1a3393d875bb466509afab68eede8/plugin/reload/setup.go#L78) is needed to do this re-registration, because https://github.com/coredns/caddy currently only offers the function `RegisterEventHook()` which panics in case of duplicate registration.  In case this PR is accepted, it might be worth to expose such a function ('register event hook if not yet registered') directly in the coredns/caddy module.
2. Terminate the background reload event handler (https://github.com/coredns/coredns/blob/master/plugin/reload/reload.go#L84) more aggressively, whenever the instance it handles is shutdown: https://github.com/cbarbian-sap/coredns/blob/a2f8cf8f35f1a3393d875bb466509afab68eede8/plugin/reload/reload.go#L78. This is IMHO not wrong, as I do not see why that go routine should continue to run if the instance it is managing is already stopped. With that change, the whole part around the `OnFinalShutdown` callback registration in `setup.go` becomes entirely obsolete.

NB: this PR assumes that shutdown callbacks will be called at most once (since the callback proposed by this PR does a send to a buffered channel (`quit`) of length 1). Is that assumption valid?

Issue #6263 

None.

Hopefully not.

### DIFF
--- a/plugin/reload/reload.go
+++ b/plugin/reload/reload.go
@@ -22,10 +22,9 @@ const (
 )
 
 type reload struct {
-	dur  time.Duration
-	u    int
-	mtx  sync.RWMutex
-	quit chan bool
+	dur time.Duration
+	u   int
+	mtx sync.RWMutex
 }
 
 func (r *reload) setUsage(u int) {
@@ -73,14 +72,23 @@ func hook(event caddy.EventName, info interface{}) error {
 
 	// this should be an instance. ok to panic if not
 	instance := info.(*caddy.Instance)
+
+	// stop the reload handler if the instance is shut down
+	quit := make(chan bool, 1)
+	instance.OnShutdown = append(instance.OnShutdown, func() error {
+		quit <- true
+		return nil
+	})
+
+	// calculate current configuration hash
 	parsedCorefile, err := parse(instance.Caddyfile())
 	if err != nil {
 		return err
 	}
-
 	sha512sum := sha512.Sum512(parsedCorefile)
 	log.Infof("Running configuration SHA512 = %x\n", sha512sum)
 
+	// start reload handler
 	go func() {
 		tick := time.NewTicker(r.interval())
 		defer tick.Stop()
@@ -118,7 +126,7 @@ func hook(event caddy.EventName, info interface{}) error {
 					}
 					return
 				}
-			case <-r.quit:
+			case <-quit:
 				return
 			}
 		}
--- a/plugin/reload/setup.go
+++ b/plugin/reload/setup.go
@@ -3,7 +3,6 @@ package reload
 import (
 	"fmt"
 	"math/rand"
-	"sync"
 	"time"
 
 	"github.com/coredns/caddy"
@@ -17,11 +16,9 @@ func init() { plugin.Register("reload", setup) }
 
 // the info reload is global to all application, whatever number of reloads.
 // it is used to transmit data between Setup and start of the hook called 'onInstanceStartup'
-// channel for QUIT is never changed in purpose.
 // WARNING: this data may be unsync after an invalid attempt of reload Corefile.
 var (
-	r              = reload{dur: defaultInterval, u: unused, quit: make(chan bool)}
-	once, shutOnce sync.Once
+	r = reload{dur: defaultInterval, u: unused}
 )
 
 func setup(c *caddy.Controller) error {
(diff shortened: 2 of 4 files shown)


cbarbian-sap:
Closing in favour of #6244.

PR #6277, opened 2023-08-17
Handle UDP responses that overflow with TC bit

SriHarsha001:
Description in this PR - https://github.com/coredns/coredns/pull/6003
Code changes in plugin/pkg/proxy/connect.go file is copied from above PR - https://github.com/coredns/coredns/pull/6003

This PR - https://github.com/coredns/coredns/pull/6100 was closed by mistake. So submitting a new PR here with the same code changes. 

Getting this PR https://github.com/miekg/dns/pull/1475 reviewed is taking time. Can we get the bug fix merged here with "overflow" string comparison. And we can address this "To accurately identify the error by type rather than string compares" as a follow up PR while I work with maintainers of miekg package. 

https://github.com/coredns/coredns/issues/5953
https://github.com/coredns/coredns/issues/5998

Not sure. 

No.

### DIFF
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -125,18 +125,17 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 				// Instead of returning an error, return an empty response with TC bit set. This will make the
 				// client retry over TCP (if that's supported) or at least receive a clean
 				// error. The connection is still good so we break before the close.
-				isDNSBufferError := false
-				isDNSOverflowError := false
+				shouldTruncate := false
 
 				// This is to handle a scenario in which upstream sets the TC bit, but doesn't truncate the response
 				// and we get ErrBuf instead of overflow.
 				if _, isDNSErr := err.(*dns.Error); isDNSErr && errors.Is(err, dns.ErrBuf) {
-					isDNSBufferError = true
+					shouldTruncate = true
 				} else if strings.Contains(err.Error(), "overflow") {
-					isDNSOverflowError = true
+					shouldTruncate = true
 				}
 
-				if isDNSBufferError || isDNSOverflowError {
+				if shouldTruncate {
 					// Only if response message id matches the request message id - return a truncated response.
 					if ret != nil && (state.Req.Id == ret.Id) {
 						ret = truncateResponse(ret)
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -187,17 +187,13 @@ const cumulativeAvgWeight = 4
 // This is to mitigate the effects of an upstream server sending UDP responses that exceed the maximum size of 512 bytes.
 func truncateResponse(response *dns.Msg) *dns.Msg {
 
-	truncatedResponse := response.Copy()
-
-	// Clear out Answer, Extra, and AuthenticatedData sections
-	truncatedResponse.Answer = nil
-	truncatedResponse.Extra = nil
-
-	// Clearing AD bit to indicate that the response is not authenticated.
-	truncatedResponse.AuthenticatedData = false
+	// Clear out Answer, Extra, and Ns sections
+	response.Answer = nil
+	response.Extra = nil
+	response.Ns = nil
 
 	// Set TC bit to indicate truncation.
-	truncatedResponse.Truncated = true
+	response.Truncated = true
 
-	return truncatedResponse
+	return response
 }
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -124,27 +124,33 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 			// Instead of returning an error, return an empty response with TC bit set. This will make the
 			// client retry over TCP (if that's supported) or at least receive a clean
 			// error. The connection is still good so we break before the close.
+
 			dnsErrBufOccured := false
+			dnsErrOverflowOccured := false
 			var perr *dns.Error
 
-			if errors.As(err, &perr) {
-				if errors.Is(err, dns.ErrBuf) {
-					dnsErrBufOccured = true
+			if proto == "udp" {
+				if errors.As(err, &perr) {
+					if errors.Is(err, dns.ErrBuf) {
+						dnsErrBufOccured = true
+					}
+				}
+
+				if strings.Contains(err.Error(), "overflow") {
+					dnsErrOverflowOccured = true
 				}
 			}
 
-			if proto == "udp" && ((strings.Contains(err.Error(), "overflow")) || dnsErrBufOccured) {
-				newRet := state.Req.Copy()
-
-				newRet.RecursionAvailable = ret.RecursionAvailable
-				newRet.Response = true
-
-				// Set TC bit to indicate truncation.
-				newRet.Truncated = true
-				ret = newRet
-
-				// break here only if response message id matches the request's message id.
+			if dnsErrBufOccured || dnsErrOverflowOccured {
+				// Only if response message id matches the request message id -
+				// Copy the state, set TC bit and break.
 				if state.Req.Id == ret.Id {
+					newRet := state.Req.Copy()
+
+					// Set TC bit to indicate truncation.
+					newRet.Truncated = true
+
+					ret = newRet
 					break
 				}
 			}
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -119,17 +119,17 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 	for {
 		ret, err = pc.c.ReadMsg()
 		if err != nil {
-			// If the error is an overflow, we probably have an upstream misbehaving in some way.
-			// (e.g. sending >512 byte UDP responses without an eDNS0 OPT RR).
-			// Instead of returning an error, return an empty response with TC bit set. This will make the
-			// client retry over TCP (if that's supported) or at least receive a clean
-			// error. The connection is still good so we break before the close.
-
-			isDNSBufferError := false
-			isDNSOverflowError := false
-			var perr *dns.Error
-
 			if proto == "udp" {
+				// For UDP, if the error is an overflow, we probably have an upstream misbehaving in some way.
+				// (e.g. sending >512 byte responses without an eDNS0 OPT RR).
+				// Instead of returning an error, return an empty response with TC bit set. This will make the
+				// client retry over TCP (if that's supported) or at least receive a clean
+				// error. The connection is still good so we break before the close.
+
+				isDNSBufferError := false
+				isDNSOverflowError := false
+				var perr *dns.Error
+
 				// This is to handle a scenario in which upstream sets the TC bit, but doesn't truncate the response
 				// and we get ErrBuf instead of overflow.
 				if errors.As(err, &perr) {
@@ -141,13 +141,13 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 				if strings.Contains(err.Error(), "overflow") {
 					isDNSOverflowError = true
 				}
-			}
 
-			if isDNSBufferError || isDNSOverflowError {
-				// Only if response message id matches the request message id - return a truncated response.
-				if ret != nil && (state.Req.Id == ret.Id) {
-					ret = truncateResponse(ret)
-					break
+				if isDNSBufferError || isDNSOverflowError {
+					// Only if response message id matches the request message id - return a truncated response.
+					if ret != nil && (state.Req.Id == ret.Id) {
+						ret = truncateResponse(ret)
+						break
+					}
 				}
 			}
 
@@ -184,7 +184,6 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 const cumulativeAvgWeight = 4
 
 // Function to return an empty response with TC (truncated) bit set.
-// This is to mitigate the effects of an upstream server sending UDP responses that exceed the maximum size of 512 bytes.
 func truncateResponse(response *dns.Msg) *dns.Msg {
 
 	// Clear out Answer, Extra, and Ns sections
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -127,9 +127,9 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 				// client retry over TCP (if that's supported) or at least receive a clean
 				// error. The connection is still good so we break before the close.
 
-				if shouldTruncateResponse(err) {
-					// Only if response message id matches the request message id - return a truncated response.
-					if ret != nil && (state.Req.Id == ret.Id) {
+				// Only if response message id matches the request message id - check for truncation and truncate response.
+				if ret != nil && (state.Req.Id == ret.Id) {
+					if shouldTruncateResponse(err) {
 						ret = truncateResponse(ret)
 						break
 					}
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -119,23 +119,15 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 	for {
 		ret, err = pc.c.ReadMsg()
 		if err != nil {
-			if proto == "udp" {
+			// If TLS has been configured, then actual protocol used is TCP.
+			if p.transport.transportTypeFromConn(pc) == typeUDP {
 				// For UDP, if the error is an overflow, we probably have an upstream misbehaving in some way.
 				// (e.g. sending >512 byte responses without an eDNS0 OPT RR).
 				// Instead of returning an error, return an empty response with TC bit set. This will make the
 				// client retry over TCP (if that's supported) or at least receive a clean
 				// error. The connection is still good so we break before the close.
-				shouldTruncate := false
 
-				// This is to handle a scenario in which upstream sets the TC bit, but doesn't truncate the response
-				// and we get ErrBuf instead of overflow.
-				if _, isDNSErr := err.(*dns.Error); isDNSErr && errors.Is(err, dns.ErrBuf) {
-					shouldTruncate = true
-				} else if strings.Contains(err.Error(), "overflow") {
-					shouldTruncate = true
-				}
-
-				if shouldTruncate {
+				if shouldTruncateResponse(err) {
 					// Only if response message id matches the request message id - return a truncated response.
 					if ret != nil && (state.Req.Id == ret.Id) {
 						ret = truncateResponse(ret)
@@ -176,9 +168,20 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 
 const cumulativeAvgWeight = 4
 
+// Function to determine if a response should be truncated.
+func shouldTruncateResponse(err error) bool {
+	// This is to handle a scenario in which upstream sets the TC bit, but doesn't truncate the response
+	// and we get ErrBuf instead of overflow.
+	if _, isDNSErr := err.(*dns.Error); isDNSErr && errors.Is(err, dns.ErrBuf) {
+		return true
+	} else if strings.Contains(err.Error(), "overflow") {
+		return true
+	}
+	return false
+}
+
 // Function to return an empty response with TC (truncated) bit set.
 func truncateResponse(response *dns.Msg) *dns.Msg {
-
 	// Clear out Answer, Extra, and Ns sections
 	response.Answer = nil
 	response.Extra = nil
@@ -186,6 +189,5 @@ func truncateResponse(response *dns.Msg) *dns.Msg {
 
 	// Set TC bit to indicate truncation.
 	response.Truncated = true
-
 	return response
 }
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -151,7 +151,7 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 					}
 				}
 			}
-
+}
 			pc.c.Close() // not giving it back
 			if err == io.EOF && cached {
 				return nil, ErrCachedClosed
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -120,13 +120,11 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 		ret, err = pc.c.ReadMsg()
 		if err != nil {
 			if proto == "udp" {
-
 				// For UDP, if the error is an overflow, we probably have an upstream misbehaving in some way.
 				// (e.g. sending >512 byte responses without an eDNS0 OPT RR).
 				// Instead of returning an error, return an empty response with TC bit set. This will make the
 				// client retry over TCP (if that's supported) or at least receive a clean
 				// error. The connection is still good so we break before the close.
-
 				isDNSBufferError := false
 				isDNSOverflowError := false
 
@@ -146,7 +144,7 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 					}
 				}
 			}
-}
+
 			pc.c.Close() // not giving it back
 			if err == io.EOF && cached {
 				return nil, ErrCachedClosed
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -147,6 +147,9 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 				if state.Req.Id == ret.Id {
 					newRet := state.Req.Copy()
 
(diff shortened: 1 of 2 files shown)


SriHarsha001:
@chrisohaver - I apologize, I had accidently deleted my fork which closed this PR - https://github.com/coredns/coredns/pull/6100 , hence submitted this PR with the same code changes. 

As trying to get this PR https://github.com/miekg/dns/pull/1475 reviewed and merged is taking a long time, can we get the bug fix merged here with "overflow" string comparison.

Can we address this comment "To accurately identify the error by type rather than string compares" as a follow up PR while I work with maintainers of miekg package.

SriHarsha001:
@chrisohaver - When you get a chance, I kindly request you to take a look at this request.

chrisohaver:
Is there anything new to review here vs where https://github.com/coredns/coredns/pull/6100 was left off?

SriHarsha001:
@chrisohaver - Thank you for your reply. I wanted to check if we get the bug fix merged here with "overflow" string comparison. And we can address this "To accurately identify the error by type rather than string compares" as a follow up PR while I work with maintainers of miekg package on this PR - https://github.com/miekg/dns/pull/1475

chrisohaver:
> @chrisohaver - Thank you for your reply. I wanted to check if we get the bug fix merged here with "overflow" string comparison. And we can address this "To accurately identify the error by type rather than string compares" as a follow up PR while I work with maintainers of miekg package on this PR - [miekg/dns#1475](https://github.com/miekg/dns/pull/1475)

Aside from that, were there other outstanding issues that needed to be addressed?  It looks like there were ... relating to AA bit?  Maybe elsewhere?  Can you review the feedback and requested changes in https://github.com/coredns/coredns/pull/6100?

SriHarsha001:
@chrisohaver - yes yes, we had discussed about setting AA bit to false here - https://github.com/coredns/coredns/pull/6100#discussion_r1242570014 

Please let me know, if there is anything outstanding that needs to be address.

chrisohaver:
> I couldn't find any specific RFC that directly addresses the handling of the authorization bit (AA) in truncated DNS responses. There is no specific handling or interpretation defined for the authorization bit (AA) in truncated responses

I think in that case we should not alter the AA bit from the original.

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,42 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+			}
+
+			if proto == "udp" && ((strings.Contains(err.Error(), "overflow")) || dnsErrBufOccured) {
+				newRet := state.Req.Copy()
+
+				// Clear AD bit in case request had set the AD bit. The empty response is not authenticated.
+				newRet.AuthenticatedData = false
+
+				// Clear AA bit in case request had set the AA bit.
+				newRet.Authoritative = false
```
chrisohaver: Unless we can show that upstream implementations clear the AA and AD bits on a response they truncate _themselves_, I don't think we should alter the header bits from the original, with the exception of setting the TC bit

chrisohaver (changes requested):

plugin/pkg/proxy/connect.go
SriHarsha001: Thank you. Sure, I have made the suggested changes.

plugin/pkg/proxy/proxy_test.go
```diff
@@ -128,3 +128,52 @@ func TestProxyIncrementFails(t *testing.T) {
...
+		ret.Answer = append(ret.Answer, test.A("example.org. IN A 127.0.0.11"))
+		ret.Answer = append(ret.Answer, test.A("example.org. IN A 127.0.0.12"))
+		ret.Answer = append(ret.Answer, test.A("example.org. IN A 127.0.0.13"))
+		ret.Answer = append(ret.Answer, test.A("example.org. IN A 127.0.0.14"))
+		ret.Answer = append(ret.Answer, test.A("example.org. IN A 127.0.0.15"))
+		ret.Answer = append(ret.Answer, test.A("example.org. IN A 127.0.0.16"))
+		ret.Answer = append(ret.Answer, test.A("example.org. IN A 127.0.0.17"))
+		ret.Answer = append(ret.Answer, test.A("example.org. IN A 127.0.0.18"))
+		ret.Answer = append(ret.Answer, test.A("example.org. IN A 127.0.0.19"))
+		ret.Answer = append(ret.Answer, test.A("example.org. IN A 127.0.0.20"))
```
chrisohaver: This isn't a performance concern since it's in a test, but to shorten code and make it more readable, please assign `ret.Answer` a single slice of tests, instead appending them one at a time.

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,36 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+			// client retry over TCP (if that's supported) or at least receive a clean
+			// error. The connection is still good so we break before the close.
+			dnsErrBufOccured := false
+			var perr *dns.Error
+
+			if errors.As(err, &perr) {
+				if errors.Is(err, dns.ErrBuf) {
+					dnsErrBufOccured = true
+				}
+			}
```
chrisohaver: We only need to check this if the protocol is udp.

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,36 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+
+			if proto == "udp" && ((strings.Contains(err.Error(), "overflow")) || dnsErrBufOccured) {
```
chrisohaver: I'm not sure how Go compiler optimizes and evals `||` expressions. Assuming it just checks left to right, it would be better to check `dnsErrBufOccured` before doing a the substring compare for "overflow".

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,36 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+				newRet := state.Req.Copy()
+
+				newRet.RecursionAvailable = ret.RecursionAvailable
```
chrisohaver: Why is this explicitly being copied over?  Is it not copied in `Copy()`?

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,36 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+				newRet.Response = true
```
chrisohaver: IIUC, this must have already been set in `state.Req`.
So we need to explicitly set it here?

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,36 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+
+				// Set TC bit to indicate truncation.
+				newRet.Truncated = true
+				ret = newRet
+
+				// break here only if response message id matches the request's message id.
+				if state.Req.Id == ret.Id {
+					break
+				}
```
chrisohaver: This case should be checked before making a copy of the response.  IIUC, it should be inverted and checked earlier.  Something like the following:

```
// drop response if its message id doesn't match the request message id.
if state.Req.Id != ret.Id {
  continue 
}
```

chrisohaver (changes requested):

plugin/pkg/proxy/connect.go
SriHarsha001: Sure, moved the protocol check. Once my PR for Miekg package is merged, I plan to update this code with below - 

			if proto == "udp" {
				switch err.(type) {
				case *dns.Error:
					if errors.Is(err, dns.ErrBuf) {
						dnsErrBufOccured = true
					}
				case *dns.OverflowError:
					dnsErrOverflowOccured = true
				default:
					dnsErrBufOccured = false
					dnsErrOverflowOccured = false
				}
			}

plugin/pkg/proxy/proxy_test.go
SriHarsha001: Sure, made the suggested changes.

plugin/pkg/proxy/connect.go
SriHarsha001: Sure, made the suggested changes.

plugin/pkg/proxy/connect.go
SriHarsha001: IIUC - When a DNS server receives a query, it can respond with the RecursionAvailable bit set to indicate that it supports recursive resolution. So preserving RecursionAvailable field from the original response (ret).

plugin/pkg/proxy/connect.go
SriHarsha001: The Response bit is set to indicate that message is a response sent by a DNS server to a client in answer to a query. The original request (state.Req.Copy()) will not have response field set. So preserving response field from the original response (ret).

plugin/pkg/proxy/connect.go
SriHarsha001: Moved this if condition up. Copying the original request state and setting TC bit only if the resp message ID matches request message ID.

plugin/pkg/proxy/connect.go
chrisohaver: Yeah,  But isn't it already copied via `state.Req.Copy()`?

So setting it explicitly here is not necessary, right?

chrisohaver: Ah, nevermind - I just realized this is copied from the _request_ - not the _response_...  which is counter intuitive.

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,45 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+				if strings.Contains(err.Error(), "overflow") {
+					dnsErrOverflowOccured = true
+				}
+			}
+
+			if dnsErrBufOccured || dnsErrOverflowOccured {
+				// Only if response message id matches the request message id -
+				// Copy the state, set TC bit and break.
+				if state.Req.Id == ret.Id {
+					newRet := state.Req.Copy()
```
chrisohaver: Why is a new return message created by copying the _request_ (`state.Req`) instead of the response (`ret`)?

Can the code instead alter `ret` (i.e. set TC bit, and empty the answer/extra/auth sections?)

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,45 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+
+				if strings.Contains(err.Error(), "overflow") {
+					dnsErrOverflowOccured = true
+				}
+			}
+
+			if dnsErrBufOccured || dnsErrOverflowOccured {
+				// Only if response message id matches the request message id -
+				// Copy the state, set TC bit and break.
+				if state.Req.Id == ret.Id {
```
chrisohaver: Can `ret` be nil?

chrisohaver (changes requested):

plugin/pkg/proxy/connect.go
SriHarsha001: I was thinking about it on why the copy was made from the original request instead of original response. Not sure if there was any specific reason (which I don't know) why this was done by Patrick in his original PR. I will reach out to Patrick and see if I am missing something.

chrisohaver: It looks like this was discussed in comments @phealy's original PR.  Original he used the `ret`, but now I see it changed, with the following rationale ...

> ... I think it would actually be cleaner to build a new response object that we guarantee is in a clean state.

I'm not convinced this is better though.  I think it would be simpler to use the original `ret`, set truncation bit, and clear out answer/extra/auth sections.

plugin/pkg/proxy/connect.go
SriHarsha001: yes, added a check in the if condition.

plugin/pkg/proxy/connect.go
SriHarsha001: Yes, I have made the suggested changes and moved the logic in a new function. Would you mind taking another look please.

plugin/pkg/proxy/connect.go
```diff
@@ -148,3 +182,22 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 }
 
 const cumulativeAvgWeight = 4
+
+// Function to return an empty response with TC (truncated) bit set.
+// This is to mitigate the effects of an upstream server sending UDP responses that exceed the maximum size of 512 bytes.
+func truncateResponse(response *dns.Msg) *dns.Msg {
+
+	truncatedResponse := response.Copy()
```
chrisohaver: Is the copy necessary?  or can we modify the original response in place?

plugin/pkg/proxy/connect.go
```diff
@@ -148,3 +182,22 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+
+	// Clear out Answer, Extra, and AuthenticatedData sections
+	truncatedResponse.Answer = nil
+	truncatedResponse.Extra = nil
+
```
chrisohaver: Also need to clear out the Authority section (it's called Ns in the msg struct).

plugin/pkg/proxy/connect.go
```diff
@@ -148,3 +182,22 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+	// Clearing AD bit to indicate that the response is not authenticated.
+	truncatedResponse.AuthenticatedData = false
```
chrisohaver: IMO, we should not modify this bit unless you have a compelling use case you can illustrate that this fixes.

chrisohaver (changes requested):

plugin/pkg/proxy/connect.go
SriHarsha001: Sure, modifying the original resp in place.

plugin/pkg/proxy/connect.go
SriHarsha001: Sure, clearing authority section.

plugin/pkg/proxy/connect.go
SriHarsha001: Sure, removed resetting AD bit here.

plugin/pkg/proxy/connect.go
```diff
@@ -148,3 +182,18 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 }
 
 const cumulativeAvgWeight = 4
+
+// Function to return an empty response with TC (truncated) bit set.
+// This is to mitigate the effects of an upstream server sending UDP responses that exceed the maximum size of 512 bytes.
```
chrisohaver:
```suggestion

```

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,38 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+			// (e.g. sending >512 byte UDP responses without an eDNS0 OPT RR).
+			// Instead of returning an error, return an empty response with TC bit set. This will make the
+			// client retry over TCP (if that's supported) or at least receive a clean
+			// error. The connection is still good so we break before the close.
+
+			isDNSBufferError := false
+			isDNSOverflowError := false
+			var perr *dns.Error
+
+			if proto == "udp" {
```
chrisohaver:
```suggestion
			if proto == "udp" {
			// For UDP, if the error is an overflow, we probably have an upstream misbehaving in some way.
			// (e.g. sending >512 byte responses without an eDNS0 OPT RR).
			// Instead of returning an error, return an empty response with TC bit set. This will make the
			// client retry over TCP (if that's supported) or at least receive a clean
			// error. The connection is still good so we break before the close.

			isDNSBufferError := false
			isDNSOverflowError := false
			var perr *dns.Error

			
```

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,38 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+				if errors.As(err, &perr) {
+					if errors.Is(err, dns.ErrBuf) {
+						isDNSBufferError = true
+					}
+				}
+
+				if strings.Contains(err.Error(), "overflow") {
+					isDNSOverflowError = true
+				}
+			}
```
chrisohaver:
```suggestion
			
```

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,38 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+
+			if isDNSBufferError || isDNSOverflowError {
+				// Only if response message id matches the request message id - return a truncated response.
+				if ret != nil && (state.Req.Id == ret.Id) {
+					ret = truncateResponse(ret)
+					break
+				}
+			}
+
```
chrisohaver, line 133:
```suggestion
}
```

chrisohaver (changes requested):
requested changes may be confusing...  I think all the error/overflow/truncation logic can be moved into the `if proto == "udp" { ... }` block.

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,37 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+				// This is to handle a scenario in which upstream sets the TC bit, but doesn't truncate the response
+				// and we get ErrBuf instead of overflow.
+				if errors.As(err, &perr) {
+					if errors.Is(err, dns.ErrBuf) {
+						isDNSBufferError = true
+					}
+				}
+
+				if strings.Contains(err.Error(), "overflow") {
```
chrisohaver: If `isDNSBufferError` is true, we don't need to do this substring string compare.

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,37 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+				// Instead of returning an error, return an empty response with TC bit set. This will make the
+				// client retry over TCP (if that's supported) or at least receive a clean
+				// error. The connection is still good so we break before the close.
+				isDNSBufferError := false
+				isDNSOverflowError := false
+				var perr *dns.Error
+
+				// This is to handle a scenario in which upstream sets the TC bit, but doesn't truncate the response
+				// and we get ErrBuf instead of overflow.
+				if errors.As(err, &perr) {
```
chrisohaver: where is `perr` used?

chrisohaver (changes requested):

plugin/pkg/proxy/connect.go
chrisohaver: And why is it named `perr`?  What does the `p` stand for?

plugin/pkg/proxy/connect.go
SriHarsha001: Thank you very much. I have addressed this in the if condition.

plugin/pkg/proxy/connect.go
SriHarsha001: Thank you very much. I have addressed this in the if condition and not using this variable.

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,32 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+
+				// This is to handle a scenario in which upstream sets the TC bit, but doesn't truncate the response
+				// and we get ErrBuf instead of overflow.
+				if _, isDNSErr := err.(*dns.Error); isDNSErr && errors.Is(err, dns.ErrBuf) {
+					isDNSBufferError = true
+				} else if strings.Contains(err.Error(), "overflow") {
+					isDNSOverflowError = true
+				}
+
+				if isDNSBufferError || isDNSOverflowError {
```
chrisohaver: I think only one boolean is needed...
 
```suggestion
				shouldTruncate := false

				// This is to handle a scenario in which upstream sets the TC bit, but doesn't truncate the response
				// and we get ErrBuf instead of overflow.
				if _, isDNSErr := err.(*dns.Error); isDNSErr && errors.Is(err, dns.ErrBuf) {
					shouldTruncate = true
				} else if strings.Contains(err.Error(), "overflow") {
					shouldTruncate = true
				}

				if shouldTruncate {
```

chrisohaver (changes requested):

plugin/pkg/proxy/connect.go
SriHarsha001: Ah yes yes. Thank you. I will do better next time.

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,31 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+				// error. The connection is still good so we break before the close.
+				shouldTruncate := false
+
+				// This is to handle a scenario in which upstream sets the TC bit, but doesn't truncate the response
+				// and we get ErrBuf instead of overflow.
+				if _, isDNSErr := err.(*dns.Error); isDNSErr && errors.Is(err, dns.ErrBuf) {
+					shouldTruncate = true
+				} else if strings.Contains(err.Error(), "overflow") {
+					shouldTruncate = true
+				}
```
chrisohaver: Please make this into a function, and add unit tests for it.

chrisohaver (changes requested):

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,31 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 	for {
 		ret, err = pc.c.ReadMsg()
 		if err != nil {
+			if proto == "udp" {
```
chrisohaver: `proto` isn't necessarily reliable to determine the actual protocol used with the upstream host.  For example, when forwarding UDP requests to upstream over DoT (dns over tls) connections, the value of proto will be "udp" here, while the actual protocol is over tcp.

SriHarsha001: oh ok understood. Changed the condition here - if p.transport.transportTypeFromConn(pc) == typeUDP

plugin/pkg/proxy/connect.go
SriHarsha001: Implemented both the suggested changes.

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,23 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+			if p.transport.transportTypeFromConn(pc) == typeUDP {
+				// For UDP, if the error is an overflow, we probably have an upstream misbehaving in some way.
+				// (e.g. sending >512 byte responses without an eDNS0 OPT RR).
+				// Instead of returning an error, return an empty response with TC bit set. This will make the
+				// client retry over TCP (if that's supported) or at least receive a clean
+				// error. The connection is still good so we break before the close.
+
+				if shouldTruncateResponse(err) {
+					// Only if response message id matches the request message id - return a truncated response.
+					if ret != nil && (state.Req.Id == ret.Id) {
```
chrisohaver: If the IDs don't match, we can skip checking the response truncation errors.  IOW, the ID match check should happen before the call to `shouldTruncateResponse(err)`.  ~It probably can go before the protocol check actually.~

chrisohaver (changes requested):

plugin/pkg/proxy/connect.go
chrisohaver: >  It probably can go before the protocol check actually.

Actually - no I don't think it can.  But it should definitely go before the call to `shouldTruncateResponse`.

SriHarsha001: Sure sure, addressed this. Thank you.

SriHarsha001:
@chrisohaver - When you get few minutes, I kindly request you to review the latest changes in this PR. Thanking you in advance.

plugin/pkg/proxy/connect.go
```diff
@@ -117,6 +119,25 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
...
+				// For UDP, if the error is an overflow, we probably have an upstream misbehaving in some way.
+				// (e.g. sending >512 byte responses without an eDNS0 OPT RR).
+				// Instead of returning an error, return an empty response with TC bit set. This will make the
+				// client retry over TCP (if that's supported) or at least receive a clean
+				// error. The connection is still good so we break before the close.
+
+				// Only if response message id matches the request message id - check for truncation and truncate response.
+				if ret != nil && (state.Req.Id == ret.Id) {
+					// Check if the response should be truncated based on the error.
+					if shouldTruncateResponse(err) {
```
chrisohaver: These three nested `if`s can be combined into one since instructions are only present in the inner most `if`. e.g. 

```
if ret != nil && state.Req.Id == ret.Id &&
	p.transport.transportTypeFromConn(pc) == typeUDP && shouldTruncateResponse(err) {
```

chrisohaver (changes requested):

plugin/pkg/proxy/connect.go
SriHarsha001: Sure, made the suggested changes. Thank you.

chrisohaver (approved):

PR #6304, opened 2023-09-07
Add PITS Global Data Recovery Services as an adopter

pheianox:
This pull request adds new entry in the adopters list.
N/A
Not sure.
Not sure.


yongtang (approved):
Thanks!

Issue #6249, opened 2023-08-09
Docker working directory changed


izenn:
Upgraded from 1.10.1 to 1.11.0 and coredns was not reading the config from /Corefile or /hosts anymore

steps to reproduce:
deploy docker image with volume mounts to /Corefile and /hosts

per docker inspect, the working directory for the docker container was changed from / to /home/nonroot

using coredns/coredns:latest from hub.docker.com

NikVandewouwer:
Not sure if related, but also for me coredns started crashing this morning after pulling latest (1.11.0) due latest tag:
```
  coredns:
    image: coredns/coredns:latest
    container_name: coredns
    volumes:
      - ${ROOT_DIR}/docker/coredns:/etc/coredns
    command: -conf /etc/coredns/Corefile
    expose:
      - 53
      - 53/udp
    ports:
      - 53:53
      - 53:53/udp
    restart: unless-stopped
```

 I've received continuously the following error in the logs every minute:
`Listen: listen tcp :53: bind: permission denied`

After reverting back to 1.10.1 everything works again:
```
  coredns:
    image: coredns/coredns:1.10.1
    container_name: coredns
    volumes:
      - ${ROOT_DIR}/docker/coredns:/etc/coredns
    command: -conf /etc/coredns/Corefile
    expose:
      - 53
      - 53/udp
    ports:
      - 53:53
      - 53:53/udp
    restart: unless-stopped
```

(I'm using docker-compose on synology nas)

SuperQ:
I'm guessing this is related to the user ID change.

https://github.com/coredns/coredns/pull/5969

NikVandewouwer:
@SuperQ could you guide me in the right direction what to do to able to run 1.11.0?

chrisohaver:
@NikVandewouwer, Try adding NET_BIND_SERVICE capability

chrisohaver:
> @NikVandewouwer, Try adding NET_BIND_SERVICE capability

Nevermind - I was thinking this was a privileged port bind issue, not a directory issue.  I wasn't paying attention.

chrisohaver:
> > @NikVandewouwer, Try adding NET_BIND_SERVICE capability
> 
> Nevermind - I was thinking this was a privileged port bind issue, not a directory issue. I wasn't paying attention.

OK, I see there are two issue reported here. Second is a port bind issue: `Listen: listen tcp :53: bind: permission denied`
Adding NET_BIND_SERVICE capability should resolve that.

[...]

Issue #5968, opened 2023-03-15
coredns should run as a non-root user


destijl:
Hi coredns folks,

Today the coredns container runs as root on the host, but it doesn't really need that level of privilege. Current security best practice is to run as an unprivileged user, which helps protect the host and other adjacent containers in case coredns is compromised by an attacker. @vinayakankugoyal sent a PR a couple of years ago to migrate it to an unprivileged user: https://github.com/coredns/coredns/pull/4528 but it didn't end up merging. From the comments on the PR I can't really tell why.

I'm filing this feature request so that hopefully we can get it migrated in the future. It would be great if coredns users were protected without having to figure out that non-root configuration for themselves.

I think Vinayak's code should still work if there's appetite to make the change.

Here's how you can verify that it's running as root today:

```
$ docker run coredns/coredns:latest
.:53
CoreDNS-1.10.1
linux/amd64, go1.20, 055b2c3
```

```
$ ps -C coredns -o user,uid,cmd
USER       UID CMD
root         0 /coredns
```

Vinayak and I are presenting about how to migrate containers to non-root in this Kubecon EU talk: https://sched.co/1HyX4

johnbelamaric:
GKE doesn't run CoreDNS. But other  than that 100% agree. Thought we did this a long time ago!

vinayakankugoyal:
I opened another PR to do this: https://github.com/coredns/coredns/pull/5969

chrisohaver:
> @vinayakankugoyal sent a PR a couple of years ago to migrate it to an unprivileged user: https://github.com/coredns/coredns/pull/4528 but it didn't end up merging. **From the comments on the PR I can't really tell why.**

The comments show there was unresolved opposition from a couple of maintainers at the time to introducing buildx.

destijl:
Thanks, updated comment about GKE. I thought this got on our radar through the GKE non-root effort, but it was through other work with kubeadm that vinayak came across it and tried to get it fixed.

johnbelamaric:
Sure. We do run CoreDNS in Anthos clusters.

destijl:
Woohoo! Thanks for getting it done!

skrobul:
I reached this issue when searching why my Kubernetes cluster broke after restarting all nodes hosting CoreDNS pods with the following message:

```
[...]

PR #6320, opened 2023-09-15
refactor: Simplify `Dockerfile`

polarathene:

**TL;DR:**
- Enforcing the `NET_BIND_SERVICE` capability for any user to bind privileged ports isn't required with Docker unless running with `--network host`?
- You could have a much simpler `Dockerfile` and allow those who enforce dropping `NET_BIND_SERVICE` to still run the `coredns` binary on **unprivileged ports** which is more secure of a choice to support. `setcap` applying the capability with _effective_ + _permitted_ set is enforcing a requirement for privileges that aren't always necessary. You can still run as non-root without this.

---

[Introduced in March 2023](https://github.com/coredns/coredns/pull/5969) to support non-root feature. It seems largely unnecessary though?
- Everything removed was to support [`setcap` mandating `NET_BIND_SERVICE`](https://man7.org/linux/man-pages/man7/capabilities.7.html) on the `/coredns` binary.
- There is no need to `COPY` for `ca-certificates.crt`, the [image base `static-debian11:nonroot` already has that](https://github.com/GoogleContainerTools/distroless/blob/0e43ec744c69497ab7aa6f06fec0bd15160c8b7c/base/README.md) (_unless the `BASE` reference `ARG` is changed_). I assume that the base image won't be changing much due to the expectation of `USER nonroot:nonroot` to be successful, thus value of overwriting the contents seems questionable?
  - **EDIT:** Originally [this feature was added when the `scratch` image was used for the final stage](https://github.com/coredns/coredns/pull/5931#issuecomment-1442786431), it was more compatible then, but was not merged until the switch to `nonroot` image which added the `USER` directive.
  - **EDIT:** The `ca-certificates.crt` is [also from earlier as an improvement for the final stage with `scratch` ](https://github.com/coredns/coredns/pull/5571) of the image build. Thus no longer important.

```bash
#! /bin/bash

# Custom config:
cat > /tmp/Corefile <<EOF
.:5353 {
  whoami
}
EOF


# No problems binding with non-root to port >1024
# in container, and < 1024 on host:
docker run --rm \
  --sysctl 'net.ipv4.ip_unprivileged_port_start=1024' \
  --cap-drop NET_BIND_SERVICE \
  --user 1000:1000
  --workdir / --volume '/tmp/Corefile:/Corefile' \
  -p 53:5353/udp
  coredns/coredns:1.10.1
```

On a host network, or perhaps a different container run-time you may need to opt for the `setcap` approach? However due to the change in Docker `20.10.0` (Dec 2020), non-root users can already bind to ports below 1024:

```console
# --cap-add NET_BIND_SERVICE is default capability,
# Dropping it has no impact due to Docker modifying sysctl:
$ docker run --rm --cap-drop NET_BIND_SERVICE coredns/coredns:1.10.1
$ docker run --rm --cap-drop NET_BIND_SERVICE --user 1000:1000 coredns/coredns:1.10.1

.:53
CoreDNS-1.10.1
linux/amd64, go1.20, 055b2c3
```

Restore the sysctl restriction and the capability will have an effect when not applied:

```console
# --cap-add NET_BIND_SERVICE is default capability, reset sysctl to 1024:
$ docker run --rm --sysctl 'net.ipv4.ip_unprivileged_port_start=1024'  coredns/coredns:1.10.1
.:53
CoreDNS-1.10.1
linux/amd64, go1.20, 055b2c3


# Root (drop capability) and user (does not inherit?) - Both fail to bind now:
$ docker run --rm --sysctl 'net.ipv4.ip_unprivileged_port_start=1024' --cap-drop NET_BIND_SERVICE  coredns/coredns:1.10.1
$ docker run --rm --sysctl 'net.ipv4.ip_unprivileged_port_start=1024'  --user '1000:1000'  coredns/coredns:1.10.1
Listen: listen tcp :53: bind: permission denied
```

The only reason to use the `setcap` call in `Dockerfile` is to make the `/coredns` binary treated as a [_"capability-dumb binary"_](https://man7.org/linux/man-pages/man7/capabilities.7.html):

```console
# `setcap cap_net_bind_service=ep` via 1.11.1 tag,
# Capability requirement is enforced, sysctl has no relevance:
$ docker run --rm \
  --sysctl 'net.ipv4.ip_unprivileged_port_start=1024'  \
  --cap-add NET_BIND_SERVICE \
  coredns/coredns:1.11.1

.:53
CoreDNS-1.11.1
linux/amd64, go1.20.7, ae2bbc2

# Capability dropped, fails check and refuses to run binary:
$ docker run --rm \
  --sysctl 'net.ipv4.ip_unprivileged_port_start=1024' \
  --cap-drop NET_BIND_SERVICE \
  coredns/coredns:1.11.1

exec /coredns: operation not permitted


# Even root:
$ docker run --rm \
  --cap-drop NET_BIND_SERVICE \
  --user '0:0'
  coredns/coredns:1.11.1

exec /coredns: operation not permitted
```

- Not a helpful error message to encounter.
- Root user does not necessarily need to bind port lower than 1024 (_and with Docker can regardless by default_)
- When the capability is dropped, not even the `root` user can bypass the restriction.
- `setcap` is only in use to bypass the non-root restriction being respected (_for **any** user in the container to leverage, with any port_).

Last issue raised requesting non-root feature:
- https://github.com/coredns/coredns/issues/5968

Reference of unexpected problems from users affected by the non-root approach (_which also changed default `WORKDIR`_):
- https://github.com/coredns/coredns/issues/6249#issuecomment-1720646718
- https://github.com/coredns/coredns/pull/5969#issuecomment-1712368430

Earlier attempts to apply the change:
- https://github.com/coredns/coredns/pull/3923#issuecomment-637969242
- https://github.com/coredns/coredns/pull/4528#issue-833964301

**NOTE:** PR #4528 at least included a comment providing better context of the capability intent, but it's not been necessary since [Docker `20.10.0`](https://docs.docker.com/engine/release-notes/20.10/#security-2) (Dec 2020) which for Docker managed networks [drops the `<=1024` privileged ports](https://github.com/moby/moby/pull/41030/files#diff-9f91bff23e0bd70d6429b63d9db2d8180d2e89cdb64db4fb3e10a96f74d36271R790).

N/A?

- Reverts back to `1.10.1` Docker image expectation without mandatory capability enforced. The entire `build` stage is not needed, it's only purpose was to apply `setcap` to a pre-built `coredns` binary.
- Has not reverted the expectation of `1.10.1` working directory (_I could include that, should just be `WORKDIR /`_).
- The `USER nonroot:nonroot` directive shouldn't be necessary as that's already what the image should run with, I left it alone for the visibility. `EXPOSE` likewise is not necessary and is [effectively documentation](https://docs.docker.com/engine/reference/builder/#expose).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,7 @@
-ARG DEBIAN_IMAGE=debian:stable-slim
 ARG BASE=gcr.io/distroless/static-debian11:nonroot
-FROM --platform=$BUILDPLATFORM ${DEBIAN_IMAGE} AS build
-SHELL [ "/bin/sh", "-ec" ]
-
-RUN export DEBCONF_NONINTERACTIVE_SEEN=true \
-           DEBIAN_FRONTEND=noninteractive \
-           DEBIAN_PRIORITY=critical \
-           TERM=linux ; \
-    apt-get -qq update ; \
-    apt-get -yyqq upgrade ; \
-    apt-get -yyqq install ca-certificates libcap2-bin; \
-    apt-get clean
-COPY coredns /coredns
-RUN setcap cap_net_bind_service=+ep /coredns
 
 FROM --platform=$TARGETPLATFORM ${BASE}
-COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
-COPY --from=build /coredns /coredns
+COPY coredns /coredns
 USER nonroot:nonroot
 EXPOSE 53 53/udp
 ENTRYPOINT ["/coredns"]


chrisohaver:
Please look into the failed ci/circleci kubernetes-test.  On the surface, the change appears to have rendered coredns unresponsive to dns queries, but could of course be some flaw or inadequacy with the test.

polarathene:
> Please look into the failed ci/circleci kubernetes-test.

I don't have experience with k8s.

```
srv_test.go:52: Could not load corefile: timeout waiting for coredns to be ready. coredns log: Listen: listen tcp :53: bind: permission denied
```

I am familiar with the error though. I would say it's due to the non-root user binding below port 1024. This wouldn't normally be allowed on the host with CoreDNS would it? The binary isn't built/packaged with the capability enforced?

It is fine for Docker due to their internal bridge networks relaxing the port restriction.

**UPDATE Nov 2025:** (_[investigated the failure further myself](https://github.com/coredns/coredns/pull/7435#issuecomment-3162622921)_)
- This was due to the non-root process running with defaults of containerd 1.x where enabling privileged ports is an opt-in feature (_unlike Docker where it's been a default since 20.10_).
- With containerd 2.0 that became enabled by default for their CRI plugin, but the CoreDNS CI runner uses `kind` which needs to bumped to the [`0.27.0` release](https://github.com/kubernetes-sigs/kind/releases/tag/v0.27.0) (Feb 2025) that includes containerd 2.0 support.

---

My concern is from [their comment in k8s](https://github.com/kubernetes/kubernetes/issues/7961#issuecomment-801130176) by @vinayakankugoyal , where perhaps they relied on the container to bind to `:53` internally for convenience and their network/runtime does not relax the privileged ports (_which is what the capability would bypass_).

It would seem that you can only keep one group happy 😅 

- For DNS I can understand wanting to favor binding to `:53` and needing that ability in other contexts beyond Dockers internal bridge networks.
- For those that bind to a `> 1024` port internally and then publish that via `:53`, it's also warranted to not require `NET_BIND_SERVICE` capability in the image, since non-root user still works fine for that.

So I guess it's your choice as maintainer which one to favour. I don't have much input for k8s as I don't know how flexible that is with DNS configuration. Docker doesn't seem to allow changing the DNS port for containers with `--dns` for example. but is otherwise functional (_like `dig` with `-p` which can choose the custom port to connect to_).

SuperQ:

polarathene:
> This needs a DCO sign-off.

I'm aware, but in light of the k8s test failure I'll wait for maintainer decision on which route the Docker image should cater to.

If `setcap` is kept, I'll just raise a new PR (_or adjust this one if preferred_) to handle the `WORKDIR` and remove the `ca-certificates.crt` copy.

chrisohaver:
> https://github.com/coredns/coredns/pull/5969 to support non-root feature. It seems largely unnecessary though?

Perhaps we should roll that back. Thoughts?

marc1404:
@polarathene I'm sorry for 'reviving' this PR after almost a year. Are you still interested in maintaining it and eventually merging it? As you and others have pointed out and as it has been referenced multiple times it's a bit cumbersome having to add the `NET_BIND_SERVICE` capability when running CoreDNS on non-privileged ports.

I tried retracing the discussions in https://github.com/coredns/coredns/issues/6249, https://github.com/coredns/coredns/pull/5969 and here. As I understand it, this change would be accepted and rolling back #5969 could also be done.

I'd be happy to see your change merged :)

polarathene:
> Are you still interested in maintaining it and eventually merging it?

There's not much to maintain, as you can see from the diff it simplified it to just a `COPY` without any messing around with enforcing capabilities.

I'm presently at full capacity elsewhere and likely will be for a while. If someone else wants to pursue the change I welcome that, as it's been effectively stale for over a year since I attempted to push this forward. I'll otherwise eventually get back to this some time in 2025.

---

> As you and others have pointed out and as it has been referenced multiple times it's a bit cumbersome having to add the `NET_BIND_SERVICE` capability when running CoreDNS on non-privileged ports.

The capability is always required for the functionality. The `setcap` approach grants it to the binary for a non-root user to have provided it's still within the bounding set (_dropping all caps for the container would prevent that_). This method is just a lazy way to go about enabling that for non-root.

The correct approach is for the devs to implement raising the capability when it's actually needed, otherwise failing. It should be roughly a few lines to implement that, I know how to do so with Rust but I'm not a Go dev, shouldn't be hard though.

It would then work regardless of in a container environment or not, no `setcap` enforcement needed (_you'd still do this for non-root to be permitted to request the capability, just not to enforce it to run the binary_). So you may want to do that for the container still, or those who want rootless could run the container as rootless (_with a root user internally mapped to a non-root on the host_).

AFAIK there isn't much concern about running as root if you drop the capabilities accordingly, the attacker would need some exploit to escape the container and that does not always require a root user within the container to do so. The capability required here isn't one that is needed for Docker IIRC, and I believe k8s also is making it redundant by setting the associated sysctl to `0` by default.

---

> I tried retracing the discussions in #6249, #5969 and here. As I understand it, this change would be accepted and rolling back #5969 could also be done.

AFAIK the main benefit of running as non-root is the user does not need to think about extra precautions like dropping capabilities you don't need. I don't stay on top of CVEs related to this but to my knowledge the defaults (at least for Docker) are fairly reasonable and compromising the container should be quite difficult unless the user does something unwise, meanwhile those that want to take the additional precautions are likely to do so already rather than trust various upstreams do it correctly.

My experience with rootless is minimal, I've done so recently with Podman, but that seems a far better way to approach it for those that are more security conscious.

polarathene:
> I'd be happy to see your change merged :)

Just to clarify, you don't seem to be marked as involved with the CoreDNS project directly?

The project maintainers still seem to be interested in the non-root approach they have in place?:
- https://github.com/coredns/coredns/pull/6416
- https://github.com/coredns/coredns/issues/6900

marc1404:
> Just to clarify, you don't seem to be marked as involved with the CoreDNS project directly?

@polarathene Correct, I'm not directly involved with the CoreDNS project. Our project ([gardener/gardener](https://github.com/gardener/gardener)) consumes CoreDNS, and while not critically important, it'd be nice if we weren't forced to grant the `NET_BIND_SERVICE` capability when running CoreDNS on ports `> 1024`.

Thank you for your quick response and detailed explanation. I'll try contributing a change where the capability is raised dynamically when required 🤞

polarathene:
> it'd be nice if we weren't forced to grant the `NET_BIND_SERVICE` capability when running CoreDNS on ports `> 1024`.

Oh ok. The only reason you wouldn't be forced is when the related systctl makes it irrelevant I think, otherwise you'd need to grant the capability (even for root if it were dropped IIRC).

If it helps, this is from an old comment of mine on the topic, demonstrating a non-root user with `setcap` only applying `+p`:

```console
# Now with `+p`:
$ docker run --rm -it --user 1000 \
  --sysctl net.ipv4.ip_unprivileged_port_start=1024 \
  --cap-drop ALL --cap-add CAP_NET_BIND_SERVICE \
  cap-test-p inspect
Ambient: []
Inheritable: []
Effective: []
Permitted: ["CAP_NET_BIND_SERVICE"]
Bounding: ["CAP_NET_BIND_SERVICE"]

# Without the `--aware` flag this would fail
# since the cap would not be in the effective set:
$ docker run --rm -it --user 1000 \
  --sysctl net.ipv4.ip_unprivileged_port_start=1024 \
  --cap-drop ALL --cap-add CAP_NET_BIND_SERVICE \
  cap-test-p --aware

CAP_NET_BIND_SERVICE is required to bind a privileged port
CAP_NET_BIND_SERVICE is permitted but missing from the effective set
Successfully added CAP_NET_BIND_SERVICE into the Effective set
Successfully bound to: TcpListener { addr: 127.0.0.1:80, fd: 3 }
```

So if the desired port to bind is below `net.ipv4.ip_unprivileged_port_start`, you'll need to check that `CAP_NET_BIND_SERVICE` is in the permitted set (_it should be by default for the root user, unless it was dropped from the bounding set_). If it's permitted, then you can raise to effective set (_allowing you to use that capability_).

The above output is from a small rust program I wrote a while back to support justifying the change to not require it with CoreDNS, but I didn't get around to publishing it 😓 

You can also test similar outside of Docker with systemd with ambient capabilities:

```console
$ sudo systemd-run --system --uid 1000 \
  --unit cap-test-none \
  --collect --pty --quiet \
  --property AmbientCapabilities=CAP_NET_BIND_SERVICE \
  target/x86_64-unknown-linux-musl/release/capability-aware --aware

CAP_NET_BIND_SERVICE is required to bind a privileged port
The effective set already includes CAP_NET_BIND_SERVICE
Successfully bound to: TcpListener { addr: 127.0.0.1:80, fd: 3 }

# The `inspect` subcommand output (bounding set excluded):
Ambient: ["CAP_NET_BIND_SERVICE"]
Inheritable: ["CAP_NET_BIND_SERVICE"]
Permitted: ["CAP_NET_BIND_SERVICE"]
Effective: ["CAP_NET_BIND_SERVICE"]
```

PR #6333, opened 2023-09-19
plugin/auto: warn when auto is unable to read elements of the directory tree

### LINKED ISSUE #6318: [ERROR] plugin/errors: 2 example.org. A: plugin/auto: no next plugin found
**What happened**:
Between CoreDNS  v1.10.1 and v1.11.1 the behaviour of the `auto` plugin seems to have changed. The same Corefile working with 1.10.1 (Docker image `docker.io/coredns/coredns@sha256:a0ead06651cf580044aeb0a0feba63591858fb2e43ade8c9dea45a6a89ae7e5e`) now leads to the error message mentioned above with the latest v1.11.1 (`docker.io/coredns/coredns@sha256:1eeb4c7316bacb1d4c8ead65571cd92dd21e27359f0d4917f1a5822a73b75db1`)

no behavioral change between those two versions

```
example.org example.com {
  auto {
    directory /zonefiles/ (.*)\.fwd\.signed {1}
  }
  errors
  log
}
```
CoreDNS is started with cmdline flag `-dns.port 53` as non-root, usage of port 53 to standard user is granted through `net.ipv4.ip_unprivileged_port_start` sysctl
/zonefiles/example.org.fwd.signed and /zonefiles/example.com.fwd.signed exist and contain valid signed zonefiles.
I currently serve them through separate `file` stanzas as a workaround.

going back to v1.10.1 also works with the above Corefile.

**Environment**:

- the version of CoreDNS: v1.11.1
- Corefile: see above
- logs, if applicable: dig output

;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 9192

- OS (e.g: `cat /etc/os-release`): standard coredns docker image (hashes see above)

chrisohaver:

Log a warning when _auto_ is unable to read elements of its directory tree.  Displaying a warning message will help with troubleshooting directory/file permission issues. Currently, no message is logged when the configured directory, or elements within are not listable/readable.

closes #6318

### DIFF
--- a/plugin/auto/walk.go
+++ b/plugin/auto/walk.go
@@ -23,11 +23,7 @@ func (a Auto) Walk() error {
 		if e != nil {
 			log.Warningf("error reading %v: %v", path, e)
 		}
-		if info == nil {
-			return nil
-		}
-
-		if info.IsDir() {
+		if info == nil || info.IsDir() {
 			return nil
 		}
 
--- a/plugin/auto/walk.go
+++ b/plugin/auto/walk.go
@@ -19,8 +19,15 @@ func (a Auto) Walk() error {
 		toDelete[n] = true
 	}
 
-	filepath.Walk(a.loader.directory, func(path string, info os.FileInfo, _ error) error {
-		if info == nil || info.IsDir() {
+	filepath.Walk(a.loader.directory, func(path string, info os.FileInfo, e error) error {
+		if e != nil {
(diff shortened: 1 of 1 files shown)


chrisohaver:
I'm hoping there does not exist a valid use case where this warning would be too spammy.

Tantalor93 (approved):
lgtm

PR #6337, opened 2023-09-23
oneaddr plugin

Snawoot:

This pull request adds a small plugin *oneaddr* which might be useful in conjunction with *loadbalance* plugin. *oneaddr* plugin leaves only one address record in the response for each name.

It is intended to workaround a problem with DNS clients which re-order addresses from response, defying the purpose of round-robin. This way, clients will get only one address at a time, which leaves them no space to misbehave.

Another use case is load balancing across significant (tens or more) worker servers. Exposing all of them in the DNS response does not make sense, it bloats response with a lot of records which client won't use anyway.

None.

Probably add plugin page on coredns.io site.

No.

### DIFF
--- a/core/dnsserver/zdirectives.go
+++ b/core/dnsserver/zdirectives.go
@@ -34,6 +34,7 @@ var Directives = []string{
 	"acl",
 	"any",
 	"chaos",
+	"oneaddr",
 	"loadbalance",
 	"tsig",
 	"cache",
--- a/core/plugin/zplugin.go
+++ b/core/plugin/zplugin.go
@@ -40,6 +40,7 @@ import (
 	_ "github.com/coredns/coredns/plugin/metrics"
 	_ "github.com/coredns/coredns/plugin/minimal"
 	_ "github.com/coredns/coredns/plugin/nsid"
+	_ "github.com/coredns/coredns/plugin/oneaddr"
 	_ "github.com/coredns/coredns/plugin/pprof"
 	_ "github.com/coredns/coredns/plugin/ready"
 	_ "github.com/coredns/coredns/plugin/reload"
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ Corefile
 build/
 release/
 vendor/
+db.*
```
SuperQ: What is this? This doesn't seem like it is necessary.

SuperQ (changes requested):

.gitignore
Snawoot: Removed

Snawoot:
@SuperQ Thanks! Sign-off done.

chrisohaver:
Thanks for your contribution! There exists a similar “onlyone” external plugin: https://github.com/coredns/learning-coredns/tree/master/plugins/onlyone, which is featured in the Learning CoreDNS book.

Snawoot:
@chrisohaver Thanks! I didn't knew about that. Indeed, *onlyone* plugin is superior in many aspects, but doesn't stacks with *loadbalance* in weighted configuration. Anyway, it would be nice to have either of them in the mainline CoreDNS. Clearly necessity for something like that is in the air.

SuperQ:
Yes, I agree, it would be nice to have one of the two "return one" plugins added to our standard set. Client re-ordering and things like RFC 3484's numeric sorting implementations have lead to many, many DNS load-balancing pains in my past.

chrisohaver:
>  _onlyone_ plugin [...] doesn't stacks with _loadbalance_ in weighted configuration.

Can you explain what the incompatibility is?

Perhaps we can extend the _loadbalance_ plugin with a new `onlyone`/`oneaddr` option instead of adding a new plugin?

danhanks:
If we extend the loadbalance plugin with this feature, I'd suggest making the number of records returned configurable: `onlyreturn N`. Return only N records, where N _could_ be 1.

PR #6351, opened 2023-10-09
plugin/etcd: the etcd client adds the DialKeepAliveTime parameter

journey-c:
When the etcd node network is unavailable, such as abnormal shutdown, etcd client will still use the connection of the node, resulting in service unavailability.

It can be simulated with the following command
```bash
iptables -A INPUT -p tcp --dport 2379 -j DROP
```
None
None
None

### DIFF
--- a/plugin/etcd/setup.go
+++ b/plugin/etcd/setup.go
@@ -101,7 +101,7 @@ func newEtcdClient(endpoints []string, cc *tls.Config, username, password string
 	etcdCfg := etcdcv3.Config{
 		Endpoints:         endpoints,
 		TLS:               cc,
-		DialKeepAliveTime: etcdTimeout + etcdTimeout/2,
+		DialKeepAliveTime: etcdTimeout,
 	}
 	if username != "" && password != "" {
 		etcdCfg.Username = username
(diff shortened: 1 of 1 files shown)


plugin/etcd/setup.go
```diff
@@ -99,8 +99,9 @@ func etcdParse(c *caddy.Controller) (*Etcd, error) {
 
 func newEtcdClient(endpoints []string, cc *tls.Config, username, password string) (*etcdcv3.Client, error) {
 	etcdCfg := etcdcv3.Config{
-		Endpoints: endpoints,
-		TLS:       cc,
+		Endpoints:         endpoints,
+		TLS:               cc,
+		DialKeepAliveTime: etcdTimeout + etcdTimeout/2,
```
yongtang: @journey-c Sorry for the late response. Is there any reference in why this value is chosen?

journey-c: @yongtang The value is based on the operation timeout and buffer time, in order to timely remove faulty nodes when the ETCD network is not working

yongtang: @journey-c  So that is the empirical configuration from your working environment?

journey-c: @yongtang Yes, but does require "DialKeepAliveTime"

yongtang: @journey-c  My concern is that "etcdTimeout + etcdTimeout/2" is a magic number. This will be harder to maintain in the future (as people may not know why `1+ 1/2` is the right number. - What about just use `etcdTimeout`? (not `1+1/2`)

journey-c: @yongtang It should be possible too

yongtang (approved):

Issue #6186, opened 2023-07-01
CD bit response is cached and served later


pemensik:
**What happened**:
- If CD bit is set in query, it disables validation at remote server

**What you expected to happen**:
- CD queries may pass, but the same answer must not be served to queries without CD bit set

**How to reproduce it (as minimally and precisely as possible)**:
- (re)start coredns
- ``dig @localhost -p 3053 +cd dnssec-failed.org``
- sleep 3
- ``dig @localhost -p 3053 +nocd dnssec-failed.org``

```dig
; <<>> DiG 9.18.16 <<>> @localhost -p 3053 +nocd dnssec-failed.org
; (2 servers found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 38206
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: 7d7b4ddf7c182cc7 (echoed)
;; QUESTION SECTION:
;dnssec-failed.org.		IN	A

;; ANSWER SECTION:
dnssec-failed.org.	295	IN	A	96.99.227.255

;; Query time: 0 msec
;; SERVER: ::1#3053(localhost) (UDP)
;; WHEN: Sat Jul 01 12:01:39 CEST 2023
;; MSG SIZE  rcvd: 91
```

- TTL shows it were cached, but CD bit is not included this time
- It allowed reaching bogus DNSSEC record without CD bit, which should not have let us. forwarder is doing DNSSEC validation.

**Anything else we need to know?**:
- RFC reference: https://www.rfc-editor.org/rfc/rfc4035.html#section-3.2.2
- RFC allows caching, but if that is done, the cache has to be separate. Similar to current DO bit separate caching would work. More at https://www.rfc-editor.org/rfc/rfc4035.html#section-4.7

**Environment**:

- the version of CoreDNS: current master, commit 6e1263d3d9ae1deef399df23d4ae47d2c3154e03
- Corefile:
```corefile
.:3053 {
  cache
  forward . 9.9.9.9
  log
}
```
- logs, if applicable:
```
.:3053
CoreDNS-1.10.1
linux/amd64, go1.20.5, 6e1263d3
[...]

PR #6354, opened 2023-10-11
plugin/cache: key cache on Checking Disabled (CD) bit

gcs278:
This update keys the cache on CD bit, which effectively separates the entries for queries with CD disabled or enabled.

The cache plugin currently does not have separate entries for CD disabled/enabled. When a user initially queries WITH the CD bit set, the cache stores the a potentially invalid cache entry. Then if a user queries without CD bit set, CoreDNS will return the invalid cache entry (as if the CD bit was set) which violates [rfc4035#section-4.7](https://datatracker.ietf.org/doc/html/rfc4035#section-4.7):

> Resolvers MUST NOT return RRsets from the BAD cache unless the resolver is not required to validate the signatures of the RRsets in question under the rules given in [Section 4.2](https://datatracker.ietf.org/doc/html/rfc4035#section-4.2) of this document.

And [rfc4035.html#section-3.2.2](https://www.rfc-editor.org/rfc/rfc4035.html#section-3.2.2) specifically says our cache should depend on the CD Bit of the original query:

> If the resolver side implements a BAD cache (see [Section 4.7](https://www.rfc-editor.org/rfc/rfc4035.html#section-4.7)) and the name server side receives a query that matches an entry in the resolver side's BAD cache, the name server side's response depends on the state of the CD bit in the original query.

https://github.com/coredns/coredns/issues/6186

N/A

No


gcs278:
WIP: Limited testing by hand is done. Need to right a unit test.

gcs278:
@chrisohaver could you put this on your queue to review when you get a chance? Thanks!

plugin/cache/cache_test.go
```diff
@@ -144,20 +149,21 @@ var cacheTestCases = []cacheTestCase{
 				test.MX("miek.nl.	3600	IN	MX	10 aspmx2.googlemail.com."),
 				test.RRSIG("miek.nl.	1800	IN	RRSIG	MX 8 2 1800 20160521031301 20160421031301 12051 miek.nl. lAaEzB5teQLLKyDenatmyhca7blLRg9DoGNrhe3NReBZN5C5/pMQk8Jc u25hv2fW23/SLm5IC2zaDpp2Fzgm6Jf7e90/yLcwQPuE7JjS55WMF+HE LEh7Z6AEb+Iq4BWmNhUz6gPxD4d9eRMs7EAzk13o1NYi5/JhfL6IlaYy qkc="),
 			},
+			RecursionAvailable: true,
 		},
 		shouldCache: false,
```
Miciah: Why is `out` specified for this test case with `shouldCache: false`?

plugin/cache/cache_test.go
```diff
@@ -17,124 +17,129 @@ import (
...
 type cacheTestCase struct {
-	test.Case                    // the expected message coming "out" of cache
-	in                 test.Case // the test message going "in" to cache
-	AuthenticatedData  bool
-	RecursionAvailable bool
-	Truncated          bool
-	shouldCache        bool
+	out         test.Case // the expected message coming "out" of cache
+	in          test.Case // the test message going "in" to cache
+	shouldCache bool
```
Miciah: At first glance, it does seem cleaner to move `AuthenticatedData`, `RecursionAvailable`, and `Truncated` into `test.Case` and to have an explicit expected `out` value for each test case.  However, it seems that `TestCacheInsertion` doesn't actually check these fields in `out`:  I can remove everything except `Answer` or `Ns` from `out` in each of the test cases, and tests still pass.  That seems strange to me.

plugin/cache/cache_test.go
```diff
@@ -728,7 +763,142 @@ func TestCacheKeepTTL(t *testing.T) {
...
+		{
+			name: "truncated bit should NOT be unique",
+			initial: test.Case{
+				Qname: "example.org.",
+				Qtype: dns.TypeA,
+			},
+			query: test.Case{
+				Qname:     "example.org.",
+				Qtype:     dns.TypeA,
+				Truncated: true,
```
Miciah: Does it make sense for `query` to be truncated?  Anyway, the test still passes if I delete this line.

plugin/cache/prefech_test.go
```diff
@@ -102,23 +164,24 @@ func TestPrefetch(t *testing.T) {
...
 				if v.fetch {
 					select {
 					case <-fetchc:
-						if !v.fetch {
-							t.Fatalf("After %s: want request to trigger a prefetch", v.after)
-						}
+						break
 					case <-time.After(time.Second):
 						t.Fatalf("After %s: want request to trigger a prefetch", v.after)
 					}
```
Miciah: While you're in here, does `want, got := rec.Rcode, dns.RcodeSuccess` a few lines below have the values reversed?

chrisohaver (approved):
LGTM thanks!

And thank you for your superb PR description!

plugin/cache/cache_test.go
gcs278: Good point. It never uses `out`. I'll clean it up.

plugin/cache/cache_test.go
gcs278: Great catch. I think the right thing to do is to add the various header flag checks to the `Header` function: https://github.com/coredns/coredns/blob/4652bf17a95844f18008ce782201eb33f95ba6dd/plugin/test/helpers.go#L127

However, i started down this path and saved my progress here: https://github.com/coredns/coredns/compare/master...gcs278:coredns:unit-test-header-checking, but this seems like a large unit test change and a bit out of the scope of this PR.

So, for the time being, I added the individual checks in `TestCacheInsertion` (AuthenticatedData, RecursionAvailable, CheckingDisabled, Authoritative). Note that Truncated actually doesn't really need to be checked, because the expectation is that Truncated responses should never be cached.

I can follow up with another PR with the unit test changes. It seems like a gap in coverage that we aren't consistently checking the header flags for the plugins.

plugin/cache/cache_test.go
gcs278: Nope that doesn't make sense, I think I was just iterating through the header fields, but TC bit only matters for responses. We won't cache truncated entries, and that test is covered in `TestCacheInsertion`, so I'll delete this.

plugin/cache/prefech_test.go
gcs278: Done.

gcs278 (commented):
Thanks for the review @Miciah. Changes pushed and ready for another. https://github.com/coredns/coredns/compare/4652bf17a95844f18008ce782201eb33f95ba6dd..2ddedca40c1ef16b6741ae2a17b6b373384b5fe0

plugin/cache/cache_test.go
gcs278: I also realize this PR leaves things in a bit of a transitional phase in which these fields are move into `test.Case`, but other unit tests are still utilizing adding the header bits at the top level test case structure such as in `cname_test.go`:
https://github.com/coredns/coredns/blob/2fe5890d8b7fc7801d308dbb51bf29c57ee9a92b/plugin/template/cname_test.go#L66

I don't think this is too problematic to be in this transitional state, since I feel like it like the header flags should be transitioned to be specified in the `test.Case`, but feel free to let me know if you feel differently.

plugin/cache/cache_test.go
```diff
@@ -256,54 +273,73 @@ func newTestCache(ttl time.Duration) (*Cache, *ResponseWriter) {
...
+				t.Error("Expected Authoritative Answer bit to be true, but was false")
+			}
+			if resp.AuthenticatedData != tc.out.AuthenticatedData {
+				t.Errorf("Expected Authenticated Data bit to be %t, but got %t", tc.out.AuthenticatedData, resp.AuthenticatedData)
+			}
+			if resp.RecursionAvailable != tc.out.RecursionAvailable {
+				t.Errorf("Expected Recursion Available bit to be %t, but got %t", tc.out.RecursionAvailable, resp.RecursionAvailable)
+			}
+			if resp.CheckingDisabled != tc.out.CheckingDisabled {
+				t.Errorf("Expected Checking Disabled bit to be %t, but got %t", tc.out.CheckingDisabled, resp.CheckingDisabled)
```
Miciah: The error messages above include the index of the test case, so it would make sense for these error messages to include it as well:
```suggestion
				t.Errorf("Test %d: Expected Authoritative Answer bit to be true, but was false", n)
			}
			if resp.AuthenticatedData != tc.out.AuthenticatedData {
				t.Errorf("Test %d: Expected Authenticated Data bit to be %t, but got %t", n, tc.out.AuthenticatedData, resp.AuthenticatedData)
			}
			if resp.RecursionAvailable != tc.out.RecursionAvailable {
				t.Errorf("Test %d: Expected Recursion Available bit to be %t, but got %t", n, tc.out.RecursionAvailable, resp.RecursionAvailable)
			}
			if resp.CheckingDisabled != tc.out.CheckingDisabled {
				t.Errorf("Test %d: Expected Checking Disabled bit to be %t, but got %t", n, tc.out.CheckingDisabled, resp.CheckingDisabled)
```
On the other hand, I notice that some `t.Error`s below don't print the index.  Rather than changing all the error messages, might be simpler to rewrite this test to use `t.Run`.  Alternatively, if you want to keep things simple in this PR, just `t.Log("Test", n)` at the top of the loop.  

On the gripping hand, it's just an error message in a test that shouldn't be failing anyway, so ¯\\\_ (ツ)\_/¯.

plugin/cache/prefech_test.go
Miciah, line 1: The file name has a typo: `prefech_test.go` should be `prefetch_test.go`.  Potential follow-up clean-up.

plugin/cache/prefech_test.go
```diff
@@ -102,28 +164,29 @@ func TestPrefetch(t *testing.T) {
...
+				req.SetEdns0(512, v.do)
+
 				c.ServeDNS(context.TODO(), rec, req)
 				if v.fetch {
 					select {
 					case <-fetchc:
-						if !v.fetch {
-							t.Fatalf("After %s: want request to trigger a prefetch", v.after)
-						}
+						break
```
Miciah: Why do you need `break` here?  Wouldn't an empty `case` clause have the same effect?  I know I reviewed this code before and didn't comment on this line then, but I can't see a reason for it now, so if there is a valid reason, it's subtle enough that it warrants a comment.

plugin/cache/cache_test.go
```diff
@@ -728,7 +764,129 @@ func TestCacheKeepTTL(t *testing.T) {
...
+				Qtype: dns.TypeA,
+			},
+			query: test.Case{
+				Qname:            "example.org.",
+				Qtype:            dns.TypeA,
+				CheckingDisabled: true,
+			},
+		},
+		{
+			name: "CD bit and DO Bit should be unique",
```
Miciah: Capitalization is inconsistent here and a few other places: "bit" versus "Bit".  (I hesitate to comment on such a trivial thing...)

Miciah (commented):
I have a few comments, but it's all trivial things that you can feel free to ignore or leave to follow-up clean-up PRs.
/lgtm

plugin/cache/cache_test.go
gcs278: Good point. I bit bullet and added another commit that wraps `TestCacheInsertion` in `t.Run`. I moved the test case struct into the unit test function, so you can reap the benefit of running individual test cases (at least with GoLand and I presume `go test` as well).

plugin/cache/cache_test.go
gcs278: All good, I don't mind fixing nits while I'm fixing other things.

plugin/cache/prefech_test.go
gcs278: Oh wow, didn't even notice. Added as a separate commit, but let me know if it's too much for this PR.

plugin/cache/prefech_test.go
gcs278: Good point - removed, and added a comment for readability.

plugin/cache/cache_test.go
```diff
@@ -275,74 +43,303 @@ func newTestCache(ttl time.Duration) (*Cache, *ResponseWriter) {
...
+			if !tc.shouldCache && found {
+				t.Errorf("Cached message that should not have been cached: %s", state.Name())
 			}
-			if resp.RecursionAvailable != tc.out.RecursionAvailable {
-				t.Errorf("Expected Recursion Available bit to be %t, but got %t", tc.out.RecursionAvailable, resp.RecursionAvailable)
-			}
-			if resp.CheckingDisabled != tc.out.CheckingDisabled {
-				t.Errorf("Expected Checking Disabled bit to be %t, but got %t", tc.out.CheckingDisabled, resp.CheckingDisabled)
+			if tc.shouldCache && !found {
+				t.Errorf("Did not cache message that should have been cached: %s", state.Name())
```
Miciah: These should be `t.Fatalf` now.  You still want to go to the next test case (which pre-`t.Run`-refactoring was done by `continue`) rather than do all the `if resp.foo != tc.out.foo { t.Errorf(...) }` checks below if you didn't find a cache entry or got an unexpected cache entry.  (Alternatively, you could make the `if found` into an `else if found`.)

gcs278: Done - I debated internally about this and thought it wouldn't hurt to continue checking the improperly found cache item, but now that seems pointless since the test is already invalid.

Miciah:
Thanks!
/lgtm

gcs278:
@chrisohaver ready for your final review

PR #6360, opened 2023-10-20
bump vendor: golang.org/x/net v0.17.0

zouyee:
vendor: golang.org/x/sys v0.13.0
full diff: https://github.com/golang/sys/compare/v0.10.0...v0.13.0

vendor: golang.org/x/net v0.17.0
full diff: https://github.com/golang/text/compare/v0.13.0...v0.17.0

This fixes the same CVE as go1.21.3 and go1.20.10;

net/http: rapid stream resets can cause excessive work

A malicious HTTP/2 client which rapidly creates requests and
immediately resets them can cause excessive server resource consumption.
While the total number of requests is bounded to the
http2.Server.MaxConcurrentStreams setting, resetting an in-progress
request allows the attacker to create a new request while the existing
one is still executing.

HTTP/2 servers now bound the number of simultaneously executing
handler goroutines to the stream concurrency limit. New requests
arriving when at the limit (which can only happen after the client
has reset an existing, in-flight request) will be queued until a
handler exits. If the request queue grows too large, the server
will terminate the connection.

This issue is also fixed in golang.org/x/net/http2 v0.17.0,
for users manually configuring HTTP/2.

The default stream concurrency limit is 250 streams (requests)
per HTTP/2 connection. This value may be adjusted using the
golang.org/x/net/http2 package; see the Server.MaxConcurrentStreams
setting and the ConfigureServer function.

This is https://github.com/advisories/GHSA-4374-p667-p6c8 and Go issue https://go.dev/issue/63417.
This is also tracked by https://github.com/advisories/GHSA-qppj-fm5r-hxr3.


SuperQ:
Fixed in https://github.com/coredns/coredns/pull/6355

Issue #6370, opened 2023-10-23
Query corednse2e.com using AzureDNS as upstream server. Resp is truncated to 512 bytes, but TC flag is not set. 


SriHarsha001:
**What happened**:
Use CoreDNS > v1.10.1 to query corednse2e.com using AzureDNS as upstream DNS server in an AKS cluster. Response is truncated to 512 bytes and TC flag is not set. 

```
CoreDNS using AzureDNS as upstream server. There is no TC flag, but the response is truncated. 
---------------------------------------------------------------------------------------------------
kubectl exec dnsutils -- dig corednse2e.com +ignore +noedns +search +noshowsearch +time=10 +tries=6 @10.96.0.10
; <<>> DiG 9.9.5-9+deb8u19-Debian <<>> corednse2e.com +ignore +noedns +search +noshowsearch +time=10 +tries=6 @10.96.0.10
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 44737
;; flags: qr rd ra; QUERY: 1, ANSWER: 30, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;corednse2e.com.                        IN      A

;; ANSWER SECTION:
corednse2e.com.         30      IN      A       10.96.0.247
corednse2e.com.         30      IN      A       10.96.0.55
corednse2e.com.         30      IN      A       10.96.0.97
corednse2e.com.         30      IN      A       10.96.0.242
corednse2e.com.         30      IN      A       10.96.0.254
corednse2e.com.         30      IN      A       10.96.0.248
corednse2e.com.         30      IN      A       10.96.0.96
corednse2e.com.         30      IN      A       10.96.0.95
corednse2e.com.         30      IN      A       10.96.0.252
corednse2e.com.         30      IN      A       10.96.0.255
corednse2e.com.         30      IN      A       10.96.0.50
corednse2e.com.         30      IN      A       10.96.0.99
corednse2e.com.         30      IN      A       10.96.0.51
corednse2e.com.         30      IN      A       76.223.105.230
corednse2e.com.         30      IN      A       10.96.0.52
corednse2e.com.         30      IN      A       10.96.0.246
corednse2e.com.         30      IN      A       10.96.0.94
corednse2e.com.         30      IN      A       10.96.0.93
corednse2e.com.         30      IN      A       10.96.0.54
corednse2e.com.         30      IN      A       10.96.0.92
[...]

PR #6371, opened 2023-10-24
fix a bug for udp protocol when response len is >= UDPsize

SriHarsha001:
The PR is to fix this issue - Using CoreDNS with versions greater than v1.10.1 to query corednse2e.com/corednsaks.com using AzureDNS as upstream DNS server in an AKS cluster, response is truncated to 512 bytes but TC flag is not set. 

If an upstream server does not set TC bit and only if response length is greater than pc.c.UDPSize or 512 default, then we set TC bit for UDP protocol, and client can retry over TCP. This is a corner case where offset is equal to len(msg) as calculated in unpackHeader defined here github.com/miekg/msg_helpers.go. In this case, there is no Overflow error returned, so the response will be truncated without TC bit set. 

For example  - 
- When domain name has 15 characters, offset after unpackQuestion will be 32. 
- Each loop adds 30 to offset and when number of A records in response is more than 15, then offset calculated in unpackRRslice will keep increasing in this sequence 62,92,122,152,182.....482,512
- In 16th loop, offset (32 question bytes + 30 * 16) == 512 which will be == len(msg) for default UDPsize, so in this case, there is no Overflow error returned. 
- And the response will be truncated (ie response does not contain all the A records) and TC flag is false. This is a failure scenario. 

Similarly, 
- When domain name has 19 characters, offset after unpackQuestion will be 36. 
- Each loop adds 34 to offset and when number of A records in response is more than 14, then offset calculated in unpackRRslice will keep increasing in this sequence 70,104,138,.....512
- In 14th loop, offset (36 question bytes + 34 * 14) == 512 which will be == len(msg) for default UDPsize, so in this case, there is no Overflow error returned. 
- And the response will be truncated (ie response does not contain all the A records) and TC flag is false. This is a failure scenario. 

As we can see, response is truncated and TC bit is false. 
kubectl exec dnsutils -- dig corednse2e.com +ignore +noedns +search +noshowsearch +time=10 +tries=6 @10.96.0.10
; <<>> DiG 9.9.5-9+deb8u19-Debian <<>> corednse2e.com +ignore +noedns +search +noshowsearch +time=10 +tries=6 @10.96.0.10
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 35567
;; flags: qr rd ra; QUERY: 1, ANSWER: 30, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;corednse2e.com.                        IN      A

;; ANSWER SECTION:
corednse2e.com.         30      IN      A       10.96.0.55
corednse2e.com.         30      IN      A       10.96.0.51
corednse2e.com.         30      IN      A       10.96.0.246
corednse2e.com.         30      IN      A       10.96.0.93
corednse2e.com.         30      IN      A       10.96.0.54
corednse2e.com.         30      IN      A       10.96.0.97
corednse2e.com.         30      IN      A       76.223.105.230
corednse2e.com.         30      IN      A       10.96.0.98
corednse2e.com.         30      IN      A       10.96.0.242
corednse2e.com.         30      IN      A       10.96.0.50
corednse2e.com.         30      IN      A       10.96.0.91
corednse2e.com.         30      IN      A       10.96.0.96
corednse2e.com.         30      IN      A       10.96.0.250
corednse2e.com.         30      IN      A       10.96.0.252
corednse2e.com.         30      IN      A       10.96.0.99
corednse2e.com.         30      IN      A       10.96.0.248
corednse2e.com.         30      IN      A       13.248.243.5
corednse2e.com.         30      IN      A       10.96.0.254
corednse2e.com.         30      IN      A       10.96.0.92
corednse2e.com.         30      IN      A       10.96.0.249
corednse2e.com.         30      IN      A       10.96.0.52
corednse2e.com.         30      IN      A       10.96.0.251
corednse2e.com.         30      IN      A       10.96.0.244
corednse2e.com.         30      IN      A       10.96.0.53
corednse2e.com.         30      IN      A       10.96.0.243
corednse2e.com.         30      IN      A       10.96.0.247
corednse2e.com.         30      IN      A       10.96.0.253
corednse2e.com.         30      IN      A       10.96.0.255
corednse2e.com.         30      IN      A       10.96.0.95
corednse2e.com.         30      IN      A       10.96.0.94

;; Query time: 37 msec
;; SERVER: 10.96.0.10#53(10.96.0.10)
;; WHEN: Thu Oct 26 05:31:44 UTC 2023
;; MSG SIZE  rcvd: 512

Whereas full response should have 64 records. 
kubectl exec dnsutils -- dig corednse2e.com +ignore +noedns +search +noshowsearch +time=10 +tries=6 @168.63.129.16
; <<>> DiG 9.9.5-9+deb8u19-Debian <<>> corednse2e.com +ignore +noedns +search +noshowsearch +time=10 +tries=6 @168.63.129.16
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 59586
;; flags: qr rd ra; QUERY: 1, ANSWER: 64, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;corednse2e.com.                        IN      A

;; ANSWER SECTION:
corednse2e.com.         600     IN      A       76.223.105.230
corednse2e.com.         600     IN      A       13.248.243.5
corednse2e.com.         600     IN      A       10.96.0.91
corednse2e.com.         600     IN      A       10.96.0.92
corednse2e.com.         600     IN      A       10.96.0.93
corednse2e.com.         600     IN      A       10.96.0.94
corednse2e.com.         600     IN      A       10.96.0.95
corednse2e.com.         600     IN      A       10.96.0.96
corednse2e.com.         600     IN      A       10.96.0.97
corednse2e.com.         600     IN      A       10.96.0.98
corednse2e.com.         600     IN      A       10.96.0.242
corednse2e.com.         600     IN      A       10.96.0.243
corednse2e.com.         600     IN      A       10.96.0.99
corednse2e.com.         600     IN      A       10.96.0.244
corednse2e.com.         600     IN      A       10.96.0.246
corednse2e.com.         600     IN      A       10.96.0.247
corednse2e.com.         600     IN      A       10.96.0.248
corednse2e.com.         600     IN      A       10.96.0.249
corednse2e.com.         600     IN      A       10.96.0.250
corednse2e.com.         600     IN      A       10.96.0.251
corednse2e.com.         600     IN      A       10.96.0.252
corednse2e.com.         600     IN      A       10.96.0.253
corednse2e.com.         600     IN      A       10.96.0.254
corednse2e.com.         600     IN      A       10.96.0.255
corednse2e.com.         600     IN      A       10.96.0.50
corednse2e.com.         600     IN      A       10.96.0.51
corednse2e.com.         600     IN      A       10.96.0.52
corednse2e.com.         600     IN      A       10.96.0.53
corednse2e.com.         600     IN      A       10.96.0.54
corednse2e.com.         600     IN      A       10.96.0.55
corednse2e.com.         600     IN      A       10.96.0.57
corednse2e.com.         600     IN      A       10.96.0.58
corednse2e.com.         600     IN      A       10.96.0.59
corednse2e.com.         600     IN      A       10.96.0.60
corednse2e.com.         600     IN      A       10.96.0.61
corednse2e.com.         600     IN      A       10.96.0.62
corednse2e.com.         600     IN      A       10.96.0.63
corednse2e.com.         600     IN      A       10.96.0.64
corednse2e.com.         600     IN      A       10.96.0.65
corednse2e.com.         600     IN      A       10.96.0.66
corednse2e.com.         600     IN      A       10.96.0.67
corednse2e.com.         600     IN      A       10.96.0.68
corednse2e.com.         600     IN      A       10.96.0.69
corednse2e.com.         600     IN      A       10.96.0.70
corednse2e.com.         600     IN      A       10.96.0.71
corednse2e.com.         600     IN      A       10.96.0.72
corednse2e.com.         600     IN      A       10.96.0.73
corednse2e.com.         600     IN      A       10.96.0.74
corednse2e.com.         600     IN      A       10.96.0.75
corednse2e.com.         600     IN      A       10.96.0.76
corednse2e.com.         600     IN      A       10.96.0.77
corednse2e.com.         600     IN      A       10.96.0.78
corednse2e.com.         600     IN      A       10.96.0.79
corednse2e.com.         600     IN      A       10.96.0.80
corednse2e.com.         600     IN      A       10.96.0.81
corednse2e.com.         600     IN      A       10.96.0.82
corednse2e.com.         600     IN      A       10.96.0.83
corednse2e.com.         600     IN      A       10.96.0.84
corednse2e.com.         600     IN      A       10.96.0.85
corednse2e.com.         600     IN      A       10.96.0.86
corednse2e.com.         600     IN      A       10.96.0.87
corednse2e.com.         600     IN      A       10.96.0.88
corednse2e.com.         600     IN      A       10.96.0.89
corednse2e.com.         600     IN      A       10.96.0.90

;; Query time: 38 msec
;; SERVER: 168.63.129.16#53(168.63.129.16)
;; WHEN: Thu Oct 26 05:32:22 UTC 2023
;; MSG SIZE  rcvd: 1056

With the proposed fix, below is the expected response. 
kubectl exec dnsutils -- dig corednse2e.com +ignore +noedns +search +noshowsearch +time=10 +tries=6 @10.96.0.10
; <<>> DiG 9.9.5-9+deb8u19-Debian <<>> corednse2e.com +ignore +noedns +search +noshowsearch +time=10 +tries=6 @10.96.0.10
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 49592
;; flags: qr tc rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;corednse2e.com.                        IN      A

;; Query time: 82 msec
;; SERVER: 10.96.0.10#53(10.96.0.10)
;; WHEN: Thu Oct 26 05:35:25 UTC 2023
;; MSG SIZE  rcvd: 32

https://github.com/coredns/coredns/issues/6370

Not sure

No

### DIFF
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -142,7 +142,7 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 			return ret, err
 		}
 		// If an upstream server does not set TC bit and response length is greater than state.Size() or 512 default buffer size, we need to set TC bit for UDP protocol, and client can retry over TCP.
-		if ret.Len() > int(pc.c.UDPSize) && p.transport.transportTypeFromConn(pc) == typeUDP {
+		if ret.Len() > int(pc.c.UDPSize) && p.transport.transportTypeFromConn(pc) == typeUDP && !ret.Truncated {
 			// Truncate the response.
 			ret = truncateResponse(ret)
 			break
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -141,9 +141,8 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 			}
 			return ret, err
 		}
-		// If an upstream server does not set TC bit and response length is greater or equal to state.Size() or 512 default buffer size, then we cannot be sure if the response is actually truncated or not.
-		// So we need to set TC bit for UDP protocol, and client can retry over TCP.
-		if ret.Len() >= int(pc.c.UDPSize) && p.transport.transportTypeFromConn(pc) == typeUDP {
+		// If an upstream server does not set TC bit and response length is greater than state.Size() or 512 default buffer size, we need to set TC bit for UDP protocol, and client can retry over TCP.
+		if ret.Len() > int(pc.c.UDPSize) && p.transport.transportTypeFromConn(pc) == typeUDP {
 			// Truncate the response.
 			ret = truncateResponse(ret)
 			break
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -143,7 +143,7 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 		}
 		// If an upstream server does not set TC bit and response length is greater or equal to state.Size() or 512 default buffer size, then we cannot be sure if the response is actually truncated or not.
 		// So we need to set TC bit for UDP protocol, and client can retry over TCP.
-		if ret.Len() == int(pc.c.UDPSize) && p.transport.transportTypeFromConn(pc) == typeUDP {
+		if ret.Len() >= int(pc.c.UDPSize) && p.transport.transportTypeFromConn(pc) == typeUDP {
 			// Truncate the response.
 			ret = truncateResponse(ret)
 			break
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -141,6 +141,13 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 			}
 			return ret, err
 		}
+		// If an upstream server does not set TC bit and response length is greater or equal to state.Size() or 512 default buffer size, then we cannot be sure if the response is actually truncated or not.
+		// So we need to set TC bit for UDP protocol, and client can retry over TCP.
+		if ret.Len() == int(pc.c.UDPSize) && p.transport.transportTypeFromConn(pc) == typeUDP {
+			// Truncate the response.
+			ret = truncateResponse(ret)
+			break
+		}
 		// drop out-of-order responses
 		if state.Req.Id == ret.Id {
 			break
--- a/plugin/pkg/proxy/proxy_test.go
+++ b/plugin/pkg/proxy/proxy_test.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"math"
+	"strconv"
 	"testing"
 	"time"
 
@@ -130,58 +131,63 @@ func TestProxyIncrementFails(t *testing.T) {
 	}
 }
 
-func TestCoreDNSOverflow(t *testing.T) {
+func TestCoreDNSOverflowWithoutEdns(t *testing.T) {
+
 	s := dnstest.NewServer(func(w dns.ResponseWriter, r *dns.Msg) {
 		ret := new(dns.Msg)
 		ret.SetReply(r)
 
 		answers := []dns.RR{
-			test.A("example.org. IN A 127.0.0.1"),
-			test.A("example.org. IN A 127.0.0.2"),
-			test.A("example.org. IN A 127.0.0.3"),
-			test.A("example.org. IN A 127.0.0.4"),
-			test.A("example.org. IN A 127.0.0.5"),
-			test.A("example.org. IN A 127.0.0.6"),
-			test.A("example.org. IN A 127.0.0.7"),
-			test.A("example.org. IN A 127.0.0.8"),
-			test.A("example.org. IN A 127.0.0.9"),
-			test.A("example.org. IN A 127.0.0.10"),
-			test.A("example.org. IN A 127.0.0.11"),
-			test.A("example.org. IN A 127.0.0.12"),
-			test.A("example.org. IN A 127.0.0.13"),
-			test.A("example.org. IN A 127.0.0.14"),
-			test.A("example.org. IN A 127.0.0.15"),
-			test.A("example.org. IN A 127.0.0.16"),
-			test.A("example.org. IN A 127.0.0.17"),
-			test.A("example.org. IN A 127.0.0.18"),
-			test.A("example.org. IN A 127.0.0.19"),
-			test.A("example.org. IN A 127.0.0.20"),
+			test.A("example123.org. IN A 127.0.0.1"),
+			test.A("example123.org. IN A 127.0.0.2"),
+			test.A("example123.org. IN A 127.0.0.3"),
+			test.A("example123.org. IN A 127.0.0.4"),
+			test.A("example123.org. IN A 127.0.0.5"),
+			test.A("example123.org. IN A 127.0.0.6"),
+			test.A("example123.org. IN A 127.0.0.7"),
+			test.A("example123.org. IN A 127.0.0.8"),
+			test.A("example123.org. IN A 127.0.0.9"),
+			test.A("example123.org. IN A 127.0.0.10"),
+			test.A("example123.org. IN A 127.0.0.11"),
+			test.A("example123.org. IN A 127.0.0.12"),
+			test.A("example123.org. IN A 127.0.0.13"),
+			test.A("example123.org. IN A 127.0.0.14"),
+			test.A("example123.org. IN A 127.0.0.15"),
+			test.A("example123.org. IN A 127.0.0.16"),
+			test.A("example123.org. IN A 127.0.0.17"),
+			test.A("example123.org. IN A 127.0.0.18"),
+			test.A("example123.org. IN A 127.0.0.19"),
+			test.A("example123.org. IN A 127.0.0.20"),
 		}
 		ret.Answer = answers
 		w.WriteMsg(ret)
 	})
-	defer s.Close()
 
+	defer s.Close()
 	p := NewProxy("TestCoreDNSOverflow", s.Addr, transport.DNS)
 	p.readTimeout = 10 * time.Millisecond
 	p.Start(5 * time.Second)
+
 	defer p.Stop()
 
 	// Test different connection modes
 	testConnection := func(proto string, options Options, expectTruncated bool) {
-		t.Helper()
 
+		t.Helper()
 		queryMsg := new(dns.Msg)
-		queryMsg.SetQuestion("example.org.", dns.TypeA)
+		queryMsg.SetQuestion("example123.org.", dns.TypeA)
 
 		recorder := dnstest.NewRecorder(&test.ResponseWriter{})
 		request := request.Request{Req: queryMsg, W: recorder}
 
 		response, err := p.Connect(context.Background(), request, options)
+
 		if err != nil {
 			t.Errorf("Failed to connect to testdnsserver: %s", err)
 		}
-
+		if response.Rcode != dns.RcodeSuccess {
+			t.Errorf("Expected No error, but got %v", response.Rcode)
+		}
 		if response.Truncated != expectTruncated {
 			t.Errorf("Expected truncated response for %s, but got TC flag %v", proto, response.Truncated)
 		}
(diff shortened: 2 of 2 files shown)


chrisohaver:
> If an upstream server does not set TC bit and response length is greater or equal to state.Size() or 512 default buffer size, then we cannot be sure if the response is actually truncated or not. This issue happens when offset in unpackHeader function ([here](https://github.com/miekg/dns/blob/0d504a67cffe8a05b3ff4057c7602b4afe0b26ef/msg_helpers.go#L75C4-L75C4)) is equal to UDP size. So we need to set TC bit for UDP protocol, and client can retry over TCP.


This would seem to introduce a new bug in CoreDNS's code (sometimes setting TC bit when we don't have to), in which case I'm wary of this fix.

SriHarsha001:
@chrisohaver - Thank you for taking a look at this PR. 

TC bit would be set only when length of ret is greater than UDPSize and for UDP protocol. 

Currently, in CoreDNS > v1.10.1 querying "corednse2e.com/corednsaks.com" for example fails when using AzureDNS as upstream DNS server. Response is truncated to 512 bytes, but TC flag is not set. 

Bug reference - https://github.com/coredns/coredns/issues/6370

chrisohaver:
Would it not be a bug to add the TC bit to a response that is not actually truncated?

SriHarsha001:
I realized a mistake. I should not be setting TC bit to true, when ret length is equal to UDPSize. I have made corrections in my previous comments and the code in connect.go

Not sure how to simulate a non compliant DNS server, for the test cases. I will try to figure this out. 

I am trying my best to explain the scenario. 

For example -
- When domain name has 15 characters, offset after unpackQuestion will be 32.
- Each loop adds 30 to offset and when number of A records in response is more than 15, then offset calculated in unpackRRslice will keep increasing in this sequence 62,92,122,152,182.....482,512
- In 16th loop, offset (32 question bytes + 30 * 16) == 512 which will be == len(msg) for default UDPsize, so in this case, there is no Overflow error returned.
- And the response will be truncated (ie response does not contain all the A records) and TC flag is false. This is a failure scenario.

Similarly,

- When domain name has 19 characters, offset after unpackQuestion will be 36.
- Each loop adds 34 to offset and when number of A records in response is more than 14, then offset calculated in unpackRRslice will keep increasing in this sequence 70,104,138,.....512
- In 14th loop, offset (36 question bytes + 34 * 14) == 512 which will be == len(msg) for default UDPsize, so in this case, there is no Overflow error returned.
- And the response will be truncated (ie response does not contain all the A records) and TC flag is false. This is a failure scenario.

If an upstream server does not set TC bit and only if response length is greater than pc.c.UDPSize or 512 default buffer size, then we set TC bit for UDP protocol, and client can retry over TCP.

plugin/pkg/proxy/connect.go
```diff
@@ -141,6 +141,12 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 			}
 			return ret, err
 		}
+		// If an upstream server does not set TC bit and response length is greater than state.Size() or 512 default buffer size, we need to set TC bit for UDP protocol, and client can retry over TCP.
+		if ret.Len() > int(pc.c.UDPSize) && p.transport.transportTypeFromConn(pc) == typeUDP && !ret.Truncated {
```
rdrozhdzh, line 145: 1. method `Len()` doesn't give you the size of original message received from UDP conn.
2. method `Len()` is quite expensive. It calculates (predicts) the size of DNS message after encoding it back to binary format from the golang structures. It's not desirable to call expensive function for each and every DNS response.
3. since the `ret` is truncated message, the `ret.Len()` will calculate the size of truncated message, i.e. I think, it will never exceed the size of allocated buffer. Also, the allocated buffer is not always equal to `pc.c.UDPSize()`, see 
https://github.com/miekg/dns/blob/0d504a67cffe8a05b3ff4057c7602b4afe0b26ef/client.go#L279-L283

4. To detect the truncation problem, you could rather compare the count of records in message header and the actual number of records in RR slices. Hope, this will work.

SriHarsha001: Thank you @rdrozhdzh for taking a look and suggestions. 

ret, err = pc.c.ReadMsg() - Here both number of records in message header and number of records in RR slice are already truncated. So unable to detect the truncation problem here.  

The other option I see is to add the below if condition in msg.go in Miekg package. Not sure if this will be accepted, but I can submit a PR and request a review.

	dns.Answer, off, err = unpackRRslice(int(dh.Ancount), msg, off)
	if err == nil && int(dh.Ancount) > int(len(dns.Answer)) {
		dns.Truncated = true
	}

rdrozhdzh: > ret, err = pc.c.ReadMsg() - Here both number of records in message header and number of records in RR slice are already truncated. So unable to detect the truncation problem here.

ok, I see `setHdr()` doesn't copy those counters to message :( 
So, looks like we cannot avoid modifications in miekg/dns lib.

> The other option I see is to add the below if condition in msg.go in Miekg package. Not sure if this will be accepted, but I can submit a PR and request a review.

1. Method `unpack()` is not a good place to override TC bit. Ideally, this method should return the message as closely to binary representation as possible. In case of inconsistencies detected, it should return error instead. The external code (e.g. coredns) can analyze this error and update TC bit accordingly (for UDP only!).
2. If you want to compare counters in header with actual records counts - this should be done for **all** answer, ns and extra sections. Ideally, I would rewrite `unpack()` to read all records from answer, ns and extra in a single `unpackRRslice()` call, since the RRs format is the same for all sections. After reading all records we can check total number and then distribute them between sections according to counters in header.
3. One more option to detect the buffer overflow is to allocate the message buffer one byte longer than we currently do for UDP. So, if we see that buffer is full after reading the message from UDP connection - that indicates overflow. Not sure what is better. Solution in p.2 looks simpler/clearer but it cannot find out the exact reason of inconsistency (malformed/hacked message or insufficient buffer size). The logic in p.3 can more reliably detect the buffer overflow condition.

SriHarsha001: Thank you for your suggestions @rdrozhdzh. 

p.1 solution looks to be the best - In case of inconsistencies detected, it should return error instead. The external code (e.g. coredns) can analyze this error and update TC bit accordingly (for UDP only!). 

Please let me know your thoughts on below proposed changes. 
	
For each section, just before the header count is updated, check if the original count received from upstream is greater than what got extracted in unpackRRslice(), then return ErrBuf. Do you foresee any issue ? 
	
	dns.Answer, off, err = unpackRRslice(int(dh.Ancount), msg, off)
	if err == nil && int(dh.Ancount) > int(len(dns.Answer)) {
		return ErrBuf
	}
	// The header counts might have been wrong so we need to update it
	dh.Ancount = uint16(len(dns.Answer))
	if err == nil {
		dns.Ns, off, err = unpackRRslice(int(dh.Nscount), msg, off)
	}
	if err == nil && int(dh.Nscount) > int(len(dns.Ns)) {
		return ErrBuf
	}
	// The header counts might have been wrong so we need to update it
	dh.Nscount = uint16(len(dns.Ns))
	if err == nil {
		dns.Extra, _, err = unpackRRslice(int(dh.Arcount), msg, off)
	}
	if err == nil && int(dh.Arcount) > int(len(dns.Extra)) {
		return ErrBuf
	}
	// The header counts might have been wrong so we need to update it
	dh.Arcount = uint16(len(dns.Extra))
	
p.2 - In the unpack() function, dh header contains count of records for each section sent by an upstream server. So I think it is not necessary to read all records from answer, ns and extra in a single unpackRRslice() call, check total number and then distribute them between sections according to counters in header. 

p.3 - This seem to involve lot of logic changes and complexity.

rdrozhdzh: The proposed code looks good to me, except the lines like
```go
dh.Ancount = uint16(len(dns.Answer))
```
First, with your changes, I don't expect mismatches anymore between counters in header and the size of slices. Second, the updated counters in `hd` don't have any effect (even in existing code), because they only update counters in the **copy of the header** being passed as input parameter. These counters don't affect anything outside `unpack()` function.

rdrozhdzh: probably, it would be even better, if you put this check into `unpackRRslice()`

SriHarsha001: Thank you again for taking a look @rdrozhdzh 

Re-assignments of counters in header is the original code. This will hold the actual records that got processed by CoreDNS and sent back to the clients. My code change is only the "if" condition to return ErrBuf, when number of records processed by CoreDNS is lesser than actual records sent by upstream (due to buff size). 

I will submit a PR for Miekg package and request a review.

SriHarsha001: Created a bug here - https://github.com/miekg/dns/issues/1492

Submitted a PR here - https://github.com/miekg/dns/pull/1493 

So will close this PR.

Issue #6188, opened 2023-07-01
EDNS0 header should not be copied to response as it arrived


pemensik:
**What happened**:
- EDNS0 header is copied back as it is. That is incorrect.

**What you expected to happen**:
- Only UDP size and DO bit is copied to response OPT header.
- Other extensions should be added to response only when the server understands them and wants to send data back to clients in it.

**How to reproduce it (as minimally and precisely as possible)**:
- ``dig +ednsopt=NSID:6578616d706c65 @localhost -p 3053`` - send query with EDNS0 extension

```dig
$ dig +ednsopt=NSID:6578616d706c65 @localhost -p 3053

; <<>> DiG 9.18.16 <<>> +ednsopt @localhost -p 3053
; (2 servers found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 1225
;; flags: qr aa rd ra ad; QUERY: 1, ANSWER: 13, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: 059cc5c1f6b85787 (echoed)
; NSID: 65 78 61 6d 70 6c 65 ("example")
;; QUESTION SECTION:
;.				IN	NS

;; ANSWER SECTION:
.			2420	IN	NS	c.root-servers.net.
.			2420	IN	NS	k.root-servers.net.
.			2420	IN	NS	j.root-servers.net.
.			2420	IN	NS	a.root-servers.net.
.			2420	IN	NS	l.root-servers.net.
.			2420	IN	NS	d.root-servers.net.
.			2420	IN	NS	g.root-servers.net.
.			2420	IN	NS	b.root-servers.net.
.			2420	IN	NS	h.root-servers.net.
.			2420	IN	NS	f.root-servers.net.
.			2420	IN	NS	i.root-servers.net.
.			2420	IN	NS	e.root-servers.net.
.			2420	IN	NS	m.root-servers.net.

;; Query time: 0 msec
;; SERVER: ::1#3053(localhost) (UDP)
;; WHEN: Sat Jul 01 13:42:53 CEST 2023
;; MSG SIZE  rcvd: 454
```

- EDNS0 is copied from the request without understanding each extension and reacting to it. ISC BIND9 and unbound send
- ``dig +ednsopt=NSID @localhost`` should not result in empty but present NSID.

**Anything else we need to know?**:
[...]

PR #6415, opened 2023-11-21
WIP: Don't copy back or forward EDNS query options

gcs278:
WIP

TODO:
- Need to filter out EDNS OPTs on forward (hop-by-hop)
- Unit testing
- Proper reference of RFCs for motivation
- Understand why rewrite plugin was adding EDNS in rewrite plugin config back to response
- Investigate: bufsize shouldn't be copied from query to response

Related RFCs:
- Cookies: https://www.rfc-editor.org/rfc/rfc7873.html#section-5.2
- NSID: https://www.rfc-editor.org/rfc/rfc5001.html#section-2.1
- Expire: https://www.rfc-editor.org/rfc/rfc7314.html#section-3.3
- TCP Keepalive: https://datatracker.ietf.org/doc/html/rfc7828#section-3.3.2
- Middlebox: https://datatracker.ietf.org/doc/html/rfc6891#section-6.2.6
- EDNS: https://datatracker.ietf.org/doc/html/rfc6891#section-6.1.2

Doesn't copy query EDNS Options to response and doesn't forward  them either.

https://github.com/coredns/coredns/issues/6188

TBD

TBD


gcs278:
I haven't been able to work on this, so I'm going to close, but anyone can feel free to revive.
/close

PR #6456, opened 2024-01-08
bump golangci-lint and disable setup-go cache

Tantalor93:
this should mitigate problems with random build failures when running linter, from my experience this happens due to conflicts of cache inside `setup-go` action and `golangci-lint` action


SuperQ:
I've not seen the cache be a problem with golangci-lint.

Tantalor93:
@SuperQ feel free to merge your bump and see if it helps 👍 https://github.com/coredns/coredns/pull/6455

Issue #4493, opened 2021-02-26
Forward plugin timeout configuration


arno-pons:
According to forward plugin documentation, the timeouts for communication are set as follows:
```
The dial timeout by default is 30s, and can decrease automatically down to 100ms based on early results.
The read timeout is static at 2s.
```

Sometimes we are experencing some timeout issues in DNS resolution using coredns in our Kubernetes clusters.
Theses timeout are probably caused by network latencies, and I want to understand if these forward plugin timeout values are configurable ? If not could at least explains how `dial timeout` is working (or point to a documentation if exists) ?

Thanks and regards.

zouyee:
related pr:
https://github.com/coredns/coredns/pull/2319

It seems not be configured.

PR #6483, opened 2024-01-30
forward: respect context

Snawoot:

1. Stop useless query attempts when context is already expired.
2. Allows to control effective forward timeout with `cancel` plugin.

Related #4493 

None, it's close to correct behaviour we would naturally expect.

No, there is no way one could successfully query some server relying on canceled context as base context for their queries.

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -110,7 +110,7 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	list := f.List()
 	deadline := time.Now().Add(defaultTimeout)
 	start := time.Now()
-	for time.Now().Before(deadline) {
+	for time.Now().Before(deadline) && ctx.Err() == nil {
 		if i >= len(list) {
 			// reached the end of list, reset to begin
 			i = 0


plugin/forward/forward.go
```diff
@@ -110,7 +110,7 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 	list := f.List()
 	deadline := time.Now().Add(defaultTimeout)
 	start := time.Now()
-	for time.Now().Before(deadline) {
+	for time.Now().Before(deadline) && ctx.Err() == nil {
```
Tantalor93, line 113: I would expect that context cancellation should be handled by `proxy.Connect` https://github.com/coredns/coredns/blob/master/plugin/forward/forward.go#L152
but currently it isn't as it seems?

IMHO the fix you propose will not correctly cancel querying when forward plugin is [retrying query over TCP](https://github.com/coredns/coredns/blob/master/plugin/forward/forward.go#L151-L163) for example

Snawoot: > I would expect that context cancellation should be handled by proxy.Connect https://github.com/coredns/coredns/blob/master/plugin/forward/forward.go#L152
but currently it isn't as it seems?

It may pass context to subsequent dials and if they respect context cancelation too, these functions will return error and it's all context handling expected to be done from this side. But the outer loop explicitly waits till certain amount of time, hence that wait has to take context into account.

Snawoot: Oh, now I see. proxy.Connect accepts ctx, but doesn't actually use it. Apparently that has to be addressed separately.

Tantalor93: > Oh, now I see. proxy.Connect accepts ctx, but doesn't actually use it. Apparently that has to be addressed separately.

agreed, can be handled separately

Tantalor93 (approved):

plugin/forward/forward.go
rdrozhdzh: We have multiple places in the loop body where we break the loop because of some errors, e.g. see lines 184, 194. Should we follow the same pattern for consistency?
With proposed implementation the error message returned by the forward plugin will be incorrect.

rdrozhdzh: Though, the error message like "no healthy proxies" may appear in existing code as well, when default deadline expires, but not all proxies were tried yet. So, this PR doesn't add new issue. Should be ok as a simple fix.

rdrozhdzh (approved):

PR #6484, opened 2024-01-30
plugin/kubernetes: Add useragent

mrueg:
CoreDNS doesn't set a UserAgent and client-go will fallback to the default one https://github.com/kubernetes/client-go/blob/master/rest/config.go#L498

In kubernetes' audit logs you'll see:
```
"userAgent":"coredns/v0.0.0 (linux/amd64) kubernetes/$Format"
```

This change adds a userAgent to the requests made by CoreDNS against the kubernetes API:

```
"userAgent":"CoreDNS/v1.11.1+ae2bbc29be1aaae0b3ded5d188968a6c97bb3144 (linux/amd64)"
```

-
-
-

### DIFF
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -6,9 +6,11 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"runtime"
 	"strings"
 	"time"
 
+	"github.com/coredns/coredns/coremain"
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/etcd/msg"
 	"github.com/coredns/coredns/plugin/kubernetes/object"
@@ -192,6 +194,7 @@ func (k *Kubernetes) getClientConfig() (*rest.Config, error) {
 			return nil, err
 		}
 		cc.ContentType = "application/vnd.kubernetes.protobuf"
+		cc.UserAgent = fmt.Sprintf("%s/%s git_commit:%s (%s/%s/%s)", coremain.CoreName, coremain.CoreVersion, coremain.GitCommit, runtime.GOOS, runtime.GOARCH, runtime.Version())
 		return cc, err
 	}
 
@@ -218,6 +221,7 @@ func (k *Kubernetes) getClientConfig() (*rest.Config, error) {
 		return nil, err
 	}
 	cc.ContentType = "application/vnd.kubernetes.protobuf"
+	cc.UserAgent = fmt.Sprintf("%s/%s git_commit:%s (%s/%s/%s)", coremain.CoreName, coremain.CoreVersion, coremain.GitCommit, runtime.GOOS, runtime.GOARCH, runtime.Version())
 	return cc, err
 }
 
--- a/coremain/run.go
+++ b/coremain/run.go
@@ -31,7 +31,7 @@ func init() {
 	flag.StringVar(&dnsserver.Port, serverType+".port", dnsserver.DefaultPort, "Default port")
 	flag.StringVar(&dnsserver.Port, "p", dnsserver.DefaultPort, "Default port")
 
-	caddy.AppName = coreName
+	caddy.AppName = CoreName
 	caddy.AppVersion = CoreVersion
 }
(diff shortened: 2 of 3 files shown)


SuperQ (commented):
I like it, minor nit about formatting.

Also, what about including `runtime.Version()` so we know what compiler version was used?

plugin/kubernetes/kubernetes.go
```diff
@@ -192,6 +194,7 @@ func (k *Kubernetes) getClientConfig() (*rest.Config, error) {
 			return nil, err
 		}
 		cc.ContentType = "application/vnd.kubernetes.protobuf"
+		cc.UserAgent = fmt.Sprintf("%s/%s-%s (%s/%s)", coremain.CoreName, coremain.CoreVersion, coremain.GitCommit, runtime.GOOS, runtime.GOARCH)
```
SuperQ: I'm not sure I like the version concatenated with the git commit. I would prefer a space separator there.

Maybe:
```suggestion
		cc.UserAgent = fmt.Sprintf("%s/%s git_commit:%s (%s/%s)", coremain.CoreName, coremain.CoreVersion, coremain.GitCommit, runtime.GOOS, runtime.GOARCH)
```

mrueg:
> I like it, minor nit about formatting.
> 
> Also, what about including `runtime.Version()` so we know what compiler version was used?

Sure we can add it, not sure if the compiled golang version provides any additional value here?

plugin/kubernetes/kubernetes.go
mrueg:
``` 
./coremain/run.go:                      appVersion = fmt.Sprintf("%s (+%s %s)", strings.TrimPrefix(gitNearestTag, "v"), GitCommit, buildDate)
```
Should we try to make it a short commit format?

SuperQ:
Usually when trying to identify a build, I like to have as much unique info as possible. I also usually like to include the git branch name. Looks like we don't have this in the current coremain package.

plugin/kubernetes/kubernetes.go
SuperQ: Long hash is fine IMO.

chrisohaver:
Have the comments been addressed?  Not sure since commits were squashed.

mrueg:
I haven't addressed the comments yet. 
I can change the format of the useragent to state git_commit explicitly.
I'm still torn a bit about exposing runtime.Version() in the useragent. Any thoughts on this @chrisohaver ?

chrisohaver:
> I'm still torn a bit about exposing runtime.Version() in the useragent. Any thoughts on this @chrisohaver ?

What's an example of how it could be useful?

mrueg:
> > I'm still torn a bit about exposing runtime.Version() in the useragent. Any thoughts on this @chrisohaver ?
> 
> What's an example of how it could be useful?

I don't have one (and I would favor not exposing it). @SuperQ any thoughts why you would need this?

mrueg:
@SuperQ @chrisohaver anything else I need to do before this is ready to be merged?

SuperQ:
Having the compiler version can be useful in finding bugs/security issues.

mrueg:
> Having the compiler version can be useful in finding bugs/security issues.

Wouldn't you rather rely on container scanning or similar for that? I think the purpose of a useragent is to help identifying the client and not to provide extended metadata (because right now it says kubernetes, which is misleading).

SuperQ:
Scanners are solving a different issue. This is about having the context inline at the time of use.

mrueg:
I've included the runtime.Version(), I personally still lack the benefit but eventually rather have a few bytes more in my log storage system than having trying to guess which coredns version logged it. :) 

This is what it will look like now:

```
"userAgent":"CoreDNS/v1.11.1 git_commit:ae2bbc29be1aaae0b3ded5d188968a6c97bb3144 (linux/amd64/go1.22)"
```

SuperQ (approved):
Great!

mrueg:
@SuperQ @chrisohaver anything else that needs to be done to get this merged? :)

PR #6486, opened 2024-01-31
Add the CoreDNS maintained alternate plugin to the build by default.

ThreeSixes:
This PR compiles the CoreDNS alternate package as part of the standard binary and Docker image that ship with various distributions or CoreDNS. This is especially useful for Kubernetes cluster operators that provide internal DNS servers which resolve specific domains locally to make sure requests stay inside a specific network or cluster unless the host resolving the address is outside the cluster for both fault tolerance and other purposes, but also leverage something like cert-manager which requires external TXT records to be generated and resolved for the DNS-01 resolution mechanism to function. Using the alternate plugin provides a mechanism to resolve entries that are only available on public DNS servers to allow these mechanisms to appropriately function without maintaining self-compiled and hosted versions of CoreDNS.

None

Any relevant documentation about included plugins should be updated to include information indicating the alternate plugin has been included and should link to the repo.

No.


chrisohaver:
Any one opposed to doing this?

plugin.cfg
```diff
@@ -71,3 +71,4 @@ whoami:whoami
 on:github.com/coredns/caddy/onevent
 sign:sign
 view:view
+alternate:github.com/coredns/alternate
```
chrisohaver: Putting _alternate_ as the last plugin would make it non-functional.
It needs to be ordered somewhere before _forward_, e.g. immediately before would work.

https://github.com/coredns/alternate?tab=readme-ov-file#building-coredns-with-alternate

chrisohaver (commented):
I'm not opposed to making alternate built-in by default.  I _think_ it's one of the more commonly used external plugins.

If we do leave alternate in a separate repository and include it as we would an external plugin, we'll need to work out how we want to handle the docs. I think the easiest way would be to create /plugins/alternate/README.md in this repo, copying/tweaking the README.md from the alternate repo.  This should allow it to get automatically picked up by coredns.io site builder.

chrisohaver:
I think the other thing to consider is if we should extend the forward plugin to handle the alternate feature as a new configurable option, instead of it being a separate plugin.  I suspect it was originally proposed to be, but then implemented as a separate plugin after not being accepted at the time.

plugin.cfg
ThreeSixes: I made the mistake of not putting -s on a commit and now I'm unable to push or rebase. :) Any advice on how to proceed?

chrisohaver:
> I made the mistake of not putting -s on a commit and now I'm **unable to push or rebase**. :) Any advice on how to proceed?

Ok by me if you nuke it and open a new PR.   Ive never had the issue you describe before, but I also don't submit PRs from my fork's master, I created a new branch for each PR.

There's probably a "git way" to back out of your situation, but I don't know what it off hand, and sometimes nuking and starting over is faster than figuring out how to git it right.

PR #6511, opened 2024-02-21
 plugin/loadbalance: add limit_answers option

danhanks:
This PR proposes an addition to the `loadbalance` plugin enabling it to limit the number of RRs in the Answers section of a DNS response after it has been shuffled / weighted by the plugin.

A similar capability is discussed in #6337 with the proposed `oneaddr` plugin. I commented there about the possibility of expanding that capability from 1 to N. This is the first step in that direction.

Changes to the README.md (as outlined in this first commit) for the `loadbalance` plugin, also probably plugin docs on coredns.io.

No.

Assuming the syntax / format of this addition are approved, I plan to push additional commits to this PR to implement this proposed change.

### DIFF
--- a/plugin/loadbalance/README.md
+++ b/plugin/loadbalance/README.md
@@ -18,6 +18,7 @@ implementations (like glibc) are particular about that.
 ~~~
 loadbalance [round_robin | weighted WEIGHTFILE] {
 			reload DURATION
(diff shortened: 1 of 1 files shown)


chrisohaver:
syntax looks good

PR #6524, opened 2024-03-04
Enable Prometheus native histograms

SuperQ:
Add a NativeHistogramBucketFactor parameter to the use of `NewHistogramVec` in order to enable use of Prometheus Native Histograms.

This will store automatically computed sparse buckets in CoreDNS. If a compatible Prometeus requests native histograms this data will returned instead of the static buckets.

The default factor of 1.05 should provide high quality resolution data.


jameshartig:
Does it make sense to set `NativeHistogramMaxBucketNumber` and `NativeHistogramMaxZeroThreshold`?

SuperQ:
@jameshartig I'm not sure those are needed right away. I think we can adjust those later if we run into issues.

Issue #6454, opened 2024-01-08
CoreDNS 1.11.2 Release


chrisohaver:
Let's do a release soon.

Any currently open PRs that we want to push to get into the next release?

SuperQ:
It looks like we have a bunch of golangci-lint issues to fix.

chrisohaver:
Yes, they've been failing I think since November.

chrisohaver:
I suspect perhaps a change in the linter?

Tantalor93:
oh I think I know, why it fails, this should fix it https://github.com/coredns/coredns/pull/6456
usually Dependabot pull requests failed due to conflict in cache of `setup-go` and `golangci-lint` action

SuperQ:
Here's the current list of non-chore changes:

* plugin/rewrite: add rcode as a rewrite option by @pschou in https://github.com/coredns/coredns/pull/6204
* Handle UDP responses that overflow with TC bit by @SriHarsha001 in https://github.com/coredns/coredns/pull/6277
* Use the correct root domain name in the proxy plugin's TestHealthX tests by @tmthrgd in https://github.com/coredns/coredns/pull/6395
* plugin/cache: key cache on Checking Disabled (CD) bit by @gcs278 in https://github.com/coredns/coredns/pull/6354
* plugin/etcd: the etcd client adds the DialKeepAliveTime parameter by @journey-c in https://github.com/coredns/coredns/pull/6351
* plugin/auto: warn when auto is unable to read elements of the directory tree by @chrisohaver in https://github.com/c
oredns/coredns/pull/6333
* plugin/tls: respect the path specified by root plugin by @mariuskimmina in https://github.com/coredns/coredns/pull/6138
* rewrite: fix multi request concurrency issue in cname rewrite  by @Amila-Rukshan in https://github.com/coredns/coredns/pull/6407

cattyhouse:
what about this : 
https://github.com/coredns/coredns/issues/6069

jason-bivins:
Hey, we have a few customers waiting on the UDP overflow fix.  Is there an ETA for the release yet?

berks-slack:
Is there an estimated date on this? There are quite a few dependency vulnerabilities in 1.11.1

chrisohaver:
The docker build/publish has failed due to authentication failure (in 2 attempts).

chrisohaver:
I do not have push access to the coredns docker hub repo, so I cannot resolve the docker push auth failure.

[...]

PR #6526, opened 2024-03-05
Repo controlled Go version

xdu31:

Repo controlled Go version, go version is controlled in a central place for:
1. binary build
2. tests
3. image build
4. github workflow

A Go version bump actually changes the build of coredns, thus needs to be reflected as a new commit hash, for CVE fixes


.go-version
```diff
@@ -0,0 +1 @@
+1.21.7
```
pacoxu: Per https://github.com/coredns/coredns/issues/6454#issuecomment-1979466900, we may use 1.21.8 for CVE update.

xdu31: > Per [#6454 (comment)](https://github.com/coredns/coredns/issues/6454#issuecomment-1979466900), we may use 1.21.8 for CVE update.

@pacoxu just bumped to 1.21.8 in a new commit

pacoxu (approved):

Issue #6536, opened 2024-03-10
Bind link-local addresses for interfaces using scope


riedel:
**What would you like to be added**:

 #4531 removed linking link-local addresses altogether. 

However, I am not sure how this is any wanted behavior: it was rather done to work around a bug.

IMHO this just a misunderstanding, how binding link-local IPv6 addresses work. Typically you just need to provide a scope, which is actually given by the interface. This means that the semantics would be fully defined when using e.g. `bind eth0`

See eg.  https://stackoverflow.com/questions/2460841/how-to-bind-a-link-local-address-to-an-ipv6-socket.

The problem is that this would require some more rewriting as the net.ParseIP does not support the scope (see e.g. https://github.com/ooni/probe/issues/2036 here) . So I was just wondering before submitting a pull request if for whatever reason the link-local exception is wanted behaviour.

**Why is this needed**:

To me this is a real problem at the moment, when trying to use CoreDNS for local DNS rewriting.  The problem with using a global address is that in my dynamic DNS setting the prefix may change, so I cannot hardcode the DNS Server for my network.

Feedback and workarounds welcome

chrisohaver:
From reading https://github.com/coredns/coredns/pull/4531, it seems the change was done because it was easier to disable than fix.  If you'd like to propose a fix you may do so.

riedel:
This is my PR in go but I tried to address it and it seemed quite straight forward actually:  #6547

PR #6547, opened 2024-03-11
plugin/bind: add zone for link-local IPv6 instead of skipping

riedel:
This rewrites #4531 to actually not skip link-local IPs.  See #6536.

### DIFF
--- a/plugin/bind/setup.go
+++ b/plugin/bind/setup.go
@@ -99,8 +99,7 @@ func listIP(args []string, ifaces []net.Interface) ([]string, error) {
 			}
 		}
 		if !isIface {
-			_, err := net.ResolveIPAddr("ip", a)
-			if err != nil {
+			if net.ParseIP(a) == nil {
 				return nil, fmt.Errorf("not a valid IP address or interface name: %q", a)
 			}
 			all = append(all, a)
--- a/plugin/bind/setup.go
+++ b/plugin/bind/setup.go
@@ -84,15 +84,23 @@ func listIP(args []string, ifaces []net.Interface) ([]string, error) {
 				}
 				for _, addr := range addrs {
 					if ipnet, ok := addr.(*net.IPNet); ok {
-						if ipnet.IP.To4() != nil || (!ipnet.IP.IsLinkLocalMulticast() && !ipnet.IP.IsLinkLocalUnicast()) {
-							all = append(all, ipnet.IP.String())
+						ipa, err := net.ResolveIPAddr("ip", ipnet.IP.String())
+						if err == nil {
+							if len(ipnet.IP) == net.IPv6len &&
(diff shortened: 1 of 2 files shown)


riedel:
Should have fixed the gofmt now. No clue how to put the relevant part into the test coverage. Would require the CI to bring up an interface with a link-local address.

I tested just tested on windows 11 and ubuntu 22.4 now.

plugin/bind/setup.go
```diff
@@ -84,15 +84,23 @@ func listIP(args []string, ifaces []net.Interface) ([]string, error) {
...
+							}
+							all = append(all, ipa.String())
 						}
 					}
 				}
 			}
 		}
 		if !isIface {
-			if net.ParseIP(a) == nil {
+			_, err := net.ResolveIPAddr("ip", a)
```
chrisohaver: From https://pkg.go.dev/net#ResolveIPAddr ...

> ResolveIPAddr returns an address of IP end point.
>
> The network must be an IP network name.
>
> **_If the host in the address parameter is not a literal IP address, ResolveIPAddr resolves the address to an address of IP end point._** Otherwise, it parses the address as a literal IP address. The address parameter can use a host name, but this is not recommended, because it will return at most one of the host name's IP addresses. 

Am I reading the above go doc correctly that `net.ResolveIPAddr` will resolve hostnames to IPs (i.e. via system DNS resolver)? If so, it could have weird consequences if for example a user typo's an interface name.  The interface name would not be a known interface, so it would fall through to here and potentially get resolved to a foreign ip address instead of returning an error.

chrisohaver (commented):
Thanks for contributing!

plugin/bind/setup.go
riedel: Thanks for catching this. 

Hadn't thought of this.

 My suggestion: I revert that second part, because if an interface is given this can be guaranteed. 
I am using this function only because it maintains the "zone" (aka scope).  And then the problem can be split off into another PR, as this is actually independent of the first part.

However, I wonder what the adverse effects could be (Attack scenario: provide a hostname that mimics an interface but provide a different IP), as the bind should fail and this is only a function for error handling. The only alternative I think would be to reimplement the scope parsing completely?!

ideas:
1. I could limit the impact if I only resort to ResolveIPAddr if ParseIP fails.
2. I could stringify the address again and do an equality check to make sure no resolving happens.
3. Find a better go function? Any ideas?

riedel: Or  declare it a feature that you now can also bind "localhost" ;) and "fix" the documentation (actually checked: it works with the downstream bind).

riedel:
I reverted the unrelated error handling and will create a new PR for that

riedel:
@chrisohaver: any more doubts after I removed the ResolveIPAddr from the error handler? Should now just behave like before except for binding all addresses of an interface (not skipping link local)

chrisohaver (approved):

PR #6549, opened 2024-03-13
Add the CoreDNS maintained alternate plugin to the build by default.

ThreeSixes:
This PR compiles the CoreDNS alternate package as part of the standard binary and Docker image that ship with various distributions or CoreDNS. This is especially useful for Kubernetes cluster operators that provide internal DNS servers which resolve specific domains locally to make sure requests stay inside a specific network or cluster unless the host resolving the address is outside the cluster for both fault tolerance and other purposes, but also leverage something like cert-manager which requires external TXT records to be generated and resolved for the DNS-01 resolution mechanism to function. Using the alternate plugin provides a mechanism to resolve entries that are only available on public DNS servers to allow these mechanisms to appropriately function without maintaining self-compiled and hosted versions of CoreDNS.

None

Any relevant documentation about included plugins should be updated to include information indicating the alternate plugin has been included and should link to the repo.

No.

### DIFF
--- a/core/dnsserver/zdirectives.go
+++ b/core/dnsserver/zdirectives.go
@@ -55,6 +55,7 @@ var Directives = []string{
 	"secondary",
 	"etcd",
 	"loop",
+	"alternate",
 	"forward",
 	"grpc",
 	"erratic",
--- a/core/plugin/zplugin.go
+++ b/core/plugin/zplugin.go
@@ -4,6 +4,7 @@ package plugin
 
 import (
 	// Include all plugins.
+	_ "github.com/coredns/alternate"
 	_ "github.com/coredns/caddy/onevent"
 	_ "github.com/coredns/coredns/plugin/acl"
 	_ "github.com/coredns/coredns/plugin/any"
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/antonmedv/expr v1.15.5
 	github.com/apparentlymart/go-cidr v1.1.0
 	github.com/aws/aws-sdk-go v1.50.10
+	github.com/coredns/alternate v0.2.9
 	github.com/coredns/caddy v1.1.1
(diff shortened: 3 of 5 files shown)


ThreeSixes:
@chrisohaver here's the new PR.

chrisohaver:
I'm not opposed to making alternate built-in by default. I think it's one of the more commonly used external plugins.

If we do leave alternate in a separate repository and include it as we would an external plugin, we'll need to work out how we want to handle the docs. I think the easiest way would be to create /plugins/alternate/README.md in this repo, copying/tweaking the README.md from the alternate repo. This should allow it to get automatically picked up by coredns.io site builder.

Another thing to consider is if we should extend the _forward_ plugin to handle the alternate feature as a new configurable option, instead of it being a separate plugin. I suspect it was originally proposed to be, but then implemented as a separate plugin after not being accepted at the time.

chrisohaver:
I think it would be significantly less code to add an `alternate` option to the existing forward plugin, than to include the _alternate_ plugin as a separate plugin.  I think this also may be more intuitive from a configuration standpoint. 

One way to do it would be to add the option `alternate RCODE_1 [,RCODE_2, RCODE_3...]`, which will pass the request to the next _forward_ instance  before writing to client if the response from upstream matches any of the RCODEs. This would also allow chaining more than one alternate without having to leverage local forwarding and multiple server blocks. 

e.g. The following would try 1.2.3.4 first. If the response is NXDOMAIN, try 5.6.7.8. If the response from 5.6.7.8 is NXDOMAIN, try 9.0.1.2.

```
. {
  forward . 1.2.3.4 {
    alternate NXDOMAIN
  }
  forward . 5.6.7.8 {
    alternate NXDOMAIN
  }
  forward . 9.0.1.2 {
  }
}
```

chrisohaver:
closing in favor of https://github.com/coredns/coredns/pull/6681

PR #6566, opened 2024-03-25
Use Docker username/password variables safely

max-allan-cgr:
@chrisohaver 's suspicion the docker creds are causing the docker push to fail. This change replaces the Makefile variables  with shell variables (which will be used literally as they are in shell).

Make will expand any `$` in a variable value as an embedded variable. So a string like `foo$bar` will end up as `fooar` - because variable `b` is unset. Using `$${VAR}` takes VAR's value direct from the shell environment without any further processing.

Of course I can't really test this fixes the problem without making a release. But using  `$(VAR)` definitely experiences the problem described above with `$` in the value, and this change fixes that. So this seems like a sensible change to make regardless and is quicker/easier than moving everyone/everything over to ghcr.io. 

[#6454
](https://github.com/coredns/coredns/issues/6454)

None

No

### DIFF
--- a/Makefile.docker
+++ b/Makefile.docker
@@ -111,7 +111,7 @@ else
 	docker manifest create --amend $(DOCKER_IMAGE_NAME):latest $(DOCKER_IMAGE_LIST_VERSIONED)
 	docker manifest push --purge $(DOCKER_IMAGE_NAME):$(VERSION)
 	docker manifest push --purge $(DOCKER_IMAGE_NAME):latest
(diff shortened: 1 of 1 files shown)


SuperQ:
Maybe we should remove the login from the Makfile and switch to [docker/login-action](https://github.com/docker/login-action).

max-allan-cgr:
@SuperQ that sounds like more than a few characters worth of easy fix! I was going for the least change, quickest way to get it working and a release out. 

If people want to fix it a different way that's fine, but please can they do it soon so we get a release?

SuperQ:
Fix here: https://github.com/coredns/coredns/pull/6567

chrisohaver:
@max-allan-cgr , thanks for the initiative to fix this.  Lets see if we can get https://github.com/coredns/coredns/pull/6567 to work quickly, and if not fall back to this minimal fix.

chrisohaver:
Superseded by https://github.com/coredns/coredns/pull/6567

PR #6567, opened 2024-03-25
Use Docker login action

SuperQ:
Use `docker/login-action` instead of handling Docker login in the Makfile. This fixes special character handling in the Makefile as well as allowing future pushing to other container registries.

https://github.com/coredns/coredns/issues/6454

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -25,5 +25,10 @@ jobs:
         run: make VERSION=${RELEASE:1} DOCKER=coredns -f Makefile.docker release
       - name: Show Docker Images
         run: docker images
+      - name: Docker login
+        uses: docker/login-action@e92390c5fb421da1463c202d546fed0ec5c39f20  # v3.1.0
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
       - name: Publish Docker Images
         run: make VERSION=${RELEASE:1} DOCKER=coredns -f Makefile.docker docker-push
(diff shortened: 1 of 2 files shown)


max-allan-cgr:
I thought: "You can remove lines 19 and 20 from the action now. The DOCKER_LOGIN and PASSWORD variables are presumably no longer needed... "

And then spotted line 114 of the Makefile.docker. Which creates a TOKEN. So that'll probably still need my `$${...}` applying to it. Unless there is a way to get a token from the action?

SuperQ:
Yes, I saw that. I hadn't finished looking into what the token is used for.

chrisohaver:
@SuperQ thanks for taking this on!

SuperQ:
Yea, I don't know what we're doing with the DELETE. But this should work for the publish side.

chrisohaver:
>  I don't know what we're doing with the DELETE. But this should work for the publish side.

If I recall correctly, the multi-platform build leaves a trail of intermediate single platform images.  If we don't do that step we leave crumbs behind - but Its not crucial to clean them up, it's just a "being tidy" thing.

chrisohaver (approved):

SuperQ:
I'm pretty sure you want those images, as they're needed by the manifest.

SuperQ:
I'm going to leave the workflow as-is and evaluate what to do with the Docker stuff for later.

chrisohaver:
> I'm pretty sure you want those images, as they're needed by the manifest.

My understanding is that the delete op just deletes the entries from the docker-hub UI. It doesn’t delete the underlying images referenced by the multi arch manifest.

PR #6575, opened 2024-03-29
add (redundant) write permission to release workflow

chrisohaver:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,8 @@ jobs:
   release:
     name: Release
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


chrisohaver:
Per [this comment](https://github.com/softprops/action-gh-release/issues/400#issuecomment-1909907285), setting this in the yaml is required. 

Although I agree with [the follow up comment](https://github.com/softprops/action-gh-release/issues/400#issuecomment-1931756304) that the repo setting should be enough, no one has actually confirmed this.  Other comments only confirm that setting permission in the workflow file works.

chrisohaver:
#6454

PR #6576, opened 2024-03-29
Use latest commit for softprops/action-gh-release

chrisohaver:

Per https://github.com/softprops/action-gh-release/issues/411

trying using latest commit to see it it fixes release creation failure


chrisohaver:
#6454

PR #6577, opened 2024-04-01
plugin/ready: fix Reset list of readiness plugins

yangjunmyfm192085:
When ready starts, it adds all implementors of readiness to its list of plugins to check for readiness. 
We need to clears out the list,  on every reload (such as kubernetes).

However, the current method of clears out the list is placed in the c.OnStartup function, which causes each plug-in to be cleared  out, and there will be only one plug-in in the final list.

### DIFF
--- a/plugin/ready/setup.go
+++ b/plugin/ready/setup.go
@@ -11,6 +11,7 @@ import (
 func init() { plugin.Register("ready", setup) }
 
 func setup(c *caddy.Controller) error {
+	plugins.Reset()
 	addr, err := parse(c)
 	if err != nil {
 		return plugin.Error("ready", err)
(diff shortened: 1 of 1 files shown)


chrisohaver:
I'm not clear on precisely what is currently broken.

yangjunmyfm192085:
For example, there are two plug-ins, plugin1 and plugin2, both requiring readiness checks, but according to the current implementation, only one plug's readiness check will take effect.

chrisohaver:
> For example, there are two plug-ins, plugin1 and plugin2, both requiring readiness checks, but according to the current implementation, only one plug's readiness check will take effect.

What do you mean by only one taking effect? 

What problem does this cause?

PR #6588, opened 2024-04-05
Port to AWS Go SDK v2

skitt:

v1 is deprecated, and support will end a little over a year from now. This ports the route53 plugin to v2.

The major difference is that listing resource records no longer uses a callback function; the results are retrieved directly, but paging has to be handled by the caller.

None.

None.

None.


plugin/route53/setup.go
```diff
@@ -122,8 +130,13 @@ func setup(c *caddy.Controller) error {
 			}
 		}
 
-		client := f(opts)
 		ctx, cancel := context.WithCancel(context.Background())
+		cfg, err := config.LoadDefaultConfig(ctx, cfgOpts...)
```
skitt: This can’t happen here, it fails in tests if the relevant AWS setup isn’t present.

jayanthvn:
Hi, Is there any plans to merge this PR?

jayanthvn:
@dilyevsky - Is it possible to please expedite the PR review? We are waiting for the release.

dilyevsky (approved):

yongtang (approved):

jaydeokar:
Can we fix the merge conflicts ? @skitt

dims:
cc @johnbelamaric @yongtang

jaydeokar:
Testing these changes -
It seems like you have to pass a region or a non-null string while creating a route53 client. 
https://github.com/aws/aws-sdk-go-v2/issues/1778 - Similar issue. 

This can be fixed by passing an env variable `AWS_REGION`, but it might break existing users of this plugin.

plugin/route53/setup.go
```diff
@@ -14,21 +14,29 @@ import (
...
+	ListHostedZonesByName(ctx context.Context, params *route53.ListHostedZonesByNameInput, optFns ...func(*route53.Options)) (*route53.ListHostedZonesByNameOutput, error)
+	ListResourceRecordSets(ctx context.Context, params *route53.ListResourceRecordSetsInput, optFns ...func(*route53.Options)) (*route53.ListResourceRecordSetsOutput, error)
+}
+
+var f = func(ctx context.Context, cfgOpts []func(*config.LoadOptions) error, clientOpts []func(*route53.Options)) (route53Client, error) {
+	cfg, err := config.LoadDefaultConfig(ctx, cfgOpts...)
+	if err != nil {
+		return nil, err
+	}
+	return route53.NewFromConfig(cfg, clientOpts...), nil
```
jaydeokar, line 49: Can we use IMDS to get the AWS_REGION if the region flag is not set ?

```
	if cfg.Region == "" {
	   imdsClient := imds.NewFromConfig(cfg)
	   region, err := imdsClient.GetRegion(ctx, &imds.GetRegionInput{})
	   if err != nil {
	       return nil, log.Fatalf("failed to get region from IMDS: %v", e)
	   }
	   cfg.Region = r.Region
	 }
```

The reason being SDK v1 used to have a default fallback region (aws global partition), which no longer exists in sdk v2. So unsuspecting users of this plugin who have not been setting a region flag will start seeing failures post upgrade..

skitt: Thanks for looking into this, I’ll update the PR when I return from PTO next week.

jaydeokar (approved):

PR #6589, opened 2024-04-07
feature: log queue and buffer memory size configuration

chenylh:
Add configuration functions for dnstap plugin.  Specify the size of the log queue and connection memory buffer.

Not involved

https://coredns.io/plugins/dnstap/

Not involved


chenylh:
Log queue and buffer memory size configuration

chenylh:
取消

PR #6623, opened 2024-04-20
Only create PTR records for endpoints with hostname defined

chrisohaver:

Per discussion in https://kubernetes.slack.com/archives/C09QYUH5W/p1713088975540039,

We should only be creating PTR records for Endpoints that have a hostname defined in the endpoint address object.
This should normally result in producing one PTR record for each IP, even if multiple Services select the same Pod.  K8s enforces that only one endpoints has a hostname defined at any time.  If for some reason more than one have hostname defined, we will return both rather than select one arbitrarily - but IIUC, this would only happen if something is awry in k8s.

https://kubernetes.slack.com/archives/C09QYUH5W/p1713088975540039

### DIFF
--- a/plugin/kubernetes/reverse.go
+++ b/plugin/kubernetes/reverse.go
@@ -44,8 +44,13 @@ func (k *Kubernetes) serviceRecordForIP(ip, name string) []msg.Service {
 		}
 		for _, eps := range ep.Subsets {
 			for _, addr := range eps.Addresses {
-				if addr.IP == ip {
-					domain := strings.Join([]string{endpointHostname(addr, k.endpointNameMode), ep.Index, Svc, k.primaryZone()}, ".")
+				// The endpoint's Hostname will be set if this endpoint is supposed to generate a PTR.
+				// So only return reverse records that match the IP AND have a non-empty hostname.
+				// Kubernetes more or less keeps this to one canonical service/endpoint per IP, but in the odd event there
+				// are multiple endpoints for the same IP with hostname set, return them all rather than selecting one
+				// arbitrarily.
+				if addr.IP == ip && addr.Hostname != "" {
+					domain := strings.Join([]string{addr.Hostname, ep.Index, Svc, k.primaryZone()}, ".")
 					svcs = append(svcs, msg.Service{Host: domain, TTL: k.ttl})
 				}
 			}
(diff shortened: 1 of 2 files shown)


chrisohaver:
Circleci k8s PTR tests need adjusted.

bradbeam (approved):
Aside from fixing tests it seems reasonable

chrisohaver:
State of the tests...

```
--- FAIL: TestKubernetesPTR (3.53s)
    --- PASS: TestKubernetesPTR/100.0.96.10.in-addr.arpa._PTR (0.13s)
    --- FAIL: TestKubernetesPTR/253.0.17.172.in-addr.arpa._PTR (0.24s)
    --- FAIL: TestKubernetesPTR/1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.d.c.b.a.4.3.2.1.ip6.arpa._PTR (0.19s)
    --- PASS: TestKubernetesPTR/120.0.96.10.in-addr.arpa._PTR (0.10s)
    --- PASS: TestKubernetesPTR/252.0.17.172.in-addr.arpa._PTR (0.10s)
    --- PASS: TestKubernetesPTR/200.0.17.172.in-addr.arpa._PTR (0.10s)
```

The latter 3 tests (4th-6th) pass because they are negative tests expecting NXDOMAINs.

The 2nd and 3rd fail because they are explicitly defined Endpoints without hostname defined.  The test uses explicitly defined Endpoints so that IPs for headless services are static.  Hopefully we can explicitly define hostnames for these endpoints.  I'll have to play with it. 

The test object creation yaml ...

https://github.com/coredns/ci/blob/master/build/kubernetes/dns-test.yaml

chrisohaver:
Adjusted tests: https://github.com/coredns/ci/pull/153

chrisohaver:
Since with this change we will no longer produce PTR records that we did previously, I think we may want to defer this change to the next minor release - or make the next release a minor release (per coredns backward compatibility change policy).

chrisohaver:
I've reverted the adjusted tests, since they don't work as expected.

I suspect we may need to re-write all A/AAAA, SRV, and PTR tests for Headless Services to work with dynamically assigned Pod IP instead of relying on manually defined Endpoints.

chrisohaver:
rebooting to rerun tests

chrisohaver:
ugh. circle-ci is now immediately failing before running tests for some unknown mysterious reason.

adrianmoisey:
Linking https://github.com/coredns/coredns/pull/6898 to keep the trail alive

Issue #6640, opened 2024-04-25
proposal(plugin/dnstap): add configuration for filtering messages 


mathetake:
**What would you like to be added**:

I would like to add a plugin option of `dnstap` to filter out certain messages.
Such configuration might be based on the [message type](https://github.com/dnstap/dnstap.pb/blob/1061e3ed4430f68a0adb87eecadbb9208e7b51dd/dnstap.proto#L156-L228) of dnstap Message. But, it would be better for our usecase if there's a way to limit the messages to only the successful client responses (the one with RcodeSuccess) since we are only interested in the successful client responses. The latter can be complex and may result in a difficult API, so I am open to discuss the option API design here. I would appreciate it if anyone can enlighten me on the existing "filtering" API design if there's already such thing in another CoreDNS plugin.

**Why is this needed**:

Currently, dnstap plugin sends three [message type](https://github.com/dnstap/dnstap.pb/blob/1061e3ed4430f68a0adb87eecadbb9208e7b51dd/dnstap.proto#L156-L228):

* `FORWARDER_QUERY`: is sent by `forwarder` plugin whenever dnstap plugin is enabled.
* `CLIENT_RESPONSE`: is sent when the wrapped `WriteMsg` is called.
* `CLIENT_QUERY`: is sent when Dnstap.Serve is called.

Even though the dnstap is [designed](https://dnstap.info/Architecture/) to minimize the impacts on the query processing critical path, it would be better if there's a way to configure which message is supposed to get sent to dnstap upstream servers to minimize the unnecessary CPU/memory pressure on CoreDNS process.

I am happy to work on this by myself once approved.

rdrozhdzh:
This may be partially related to https://github.com/coredns/coredns/issues/4198

mathetake:
@rdrozhdzh thanks, that issue looks almost the same as this one. More importantly, I didn't know the fact that the similar configuration existed before. I will go ahead and open a PR soon, and will ping you there.

mathetake:
https://github.com/coredns/coredns/pull/6643

PR #6643, opened 2024-04-26
dnstap: adds message_types option for filtering

### LINKED ISSUE #4198: Enhance configuration of dnstap plugin

Plugin `dnstap` should support configuration options for allowed types of DNSTAP messages.
The config may look like
```
  dnstap tcp://127.0.0.1:9123 full CR FR
```
where all parameters after endpoint are optional
or
```
  dnstap tcp://127.0.0.1:9123 full {
    allow CR FR
  }
```
Other plugins should be able to check these config options. Bunch of booleans is ok.
See also https://github.com/coredns/coredns/pull/4179#discussion_r502541328 for reference

**Why is this needed**:
PR #4179 removed the ability to disable CQ and/or CR messages with flags passes via context. 
Adding the configuration options to dnstap plugin will restore ability to control generation of dnstap messages of particular types.

Note, other popular DNS servers (e.g. unbound) also have similar configuration options.

mathetake:
Note: I am completely open to discuss the naming of the option

This resolves #4198 and #6640 by introducing `message_types` configuration in the `dnstap` plugin.

#4198 and #6640 

dnstap/README.md

No.

---

~I left this as draft until we agreed on the naming of the option. I would appreciate it if you folks can take a look @rdrozhdzh  @miekg (pinging the people in the original issue #4198)~

### DIFF
--- a/plugin/dnstap/README.md
+++ b/plugin/dnstap/README.md
@@ -20,6 +20,7 @@ dnstap SOCKET [full] {
   [version VERSION]
   [extra EXTRA]
   [skipverify]
+  [message_types MESSAGE_TYPES]
 }
 ~~~
 
@@ -29,6 +30,12 @@ dnstap SOCKET [full] {
 * **VERSION** to override the version field. Defaults to the CoreDNS version.
 * **EXTRA** to define "extra" field in dnstap payload, [metadata](../metadata/) replacement available here.
 * `skipverify` to skip tls verification during connection. Default to be secure
+* **MESSAGE_TYPES** is a space-separated list of [dnstap message types](https://github.com/dnstap/dnstap.pb/blob/1061e3ed4430f68a0adb87eecadbb9208e7b51dd/dnstap.proto#L156-L229) to be tapped.
+Defaults to tapping all message types. Currently, this plugin only supports the following message types: 
+  * `CLIENT_QUERY`
+  * `CLIENT_RESPONSE`
+  * `FORWARDER_QUERY`: only available when the forward plugin is used.
+  * `FORWARDER_RESPONSE`: only available when the forward plugin is used.
 
 ## Examples
 
@@ -80,6 +87,14 @@ Log to a remote TLS endpoint.
 ~~~ txt
 dnstap tls://127.0.0.1:6000 full {
   skipverify
+}
+~~~
+
+Log only `CLIENT_QUERY` and `CLIENT_RESPONSE` messages to a remote endpoint.
+
+~~~ txt
+dnstap tcp://example.com:6000 full {
+  message_types CLIENT_QUERY CLIENT_RESPONSE
 }
 ~~~
 
--- a/plugin/dnstap/README.md
+++ b/plugin/dnstap/README.md
@@ -31,11 +31,12 @@ dnstap SOCKET [full] {
 * **EXTRA** to define "extra" field in dnstap payload, [metadata](../metadata/) replacement available here.
 * `skipverify` to skip tls verification during connection. Default to be secure
 * **MESSAGE_TYPES** is a space-separated list of [dnstap message types](https://github.com/dnstap/dnstap.pb/blob/1061e3ed4430f68a0adb87eecadbb9208e7b51dd/dnstap.proto#L156-L229) to be tapped.
-Defaults to tapping all message types. Currently, this plugin only supports the following message types: 
+Defaults to tapping all message types. Currently, this plugin only supports the following message types by default:
   * `CLIENT_QUERY`
   * `CLIENT_RESPONSE`
   * `FORWARDER_QUERY`: only available when the forward plugin is used.
   * `FORWARDER_RESPONSE`: only available when the forward plugin is used.
+  * Other message types might be tapped by the custom plugins that use the dnstap as described in the [Using Dnstap in your plugin](#using-dnstap-in-your-plugin) section.
 
 ## Examples
 
--- a/plugin/dnstap/handler.go
+++ b/plugin/dnstap/handler.go
@@ -71,21 +71,26 @@ func (h *Dnstap) tapClientQuery(ctx context.Context, w dns.ResponseWriter, query
 
 // ServeDNS logs the client query and response to dnstap and passes the dnstap Context.
 func (h *Dnstap) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
-	if h.MessageTypeEnabled(tap.Message_CLIENT_RESPONSE) {
+	crEnabled, cqEnabled := h.MessageTypeEnabled(tap.Message_CLIENT_RESPONSE), h.MessageTypeEnabled(tap.Message_CLIENT_QUERY)
+	var qt time.Time
+	if crEnabled || cqEnabled {
+		qt = time.Now()
+	}
+	if crEnabled {
 		// Custom ResponseWriter is only used to tap CLIENT_RESPONSE messages, so we only create it if needed.
 		w = &ResponseWriter{
 			ResponseWriter: w,
 			Dnstap:         h,
 			query:          r,
 			ctx:            ctx,
-			queryTime:      time.Now(),
+			queryTime:      qt,
 		}
 	}
 
 	// The query tap message should be sent before sending the query to the
 	// forwarder. Otherwise, the tap messages will come out out of order.
-	if h.MessageTypeEnabled(tap.Message_CLIENT_QUERY) {
-		h.tapClientQuery(ctx, w, r, time.Now())
+	if cqEnabled {
+		h.tapClientQuery(ctx, w, r, qt)
 	}
 	return plugin.NextOrFailure(h.Name(), h.Next, ctx, w, r)
 }
--- a/plugin/dnstap/handler.go
+++ b/plugin/dnstap/handler.go
@@ -71,8 +71,8 @@ func (h *Dnstap) tapClientQuery(ctx context.Context, w dns.ResponseWriter, query
 
 // ServeDNS logs the client query and response to dnstap and passes the dnstap Context.
 func (h *Dnstap) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
-	qt := time.Now()
 	if h.MessageTypeEnabled(tap.Message_CLIENT_RESPONSE) {
+		qt := time.Now()
 		// Custom ResponseWriter is only used to tap CLIENT_RESPONSE messages, so we only create it if needed.
 		w = &ResponseWriter{
 			ResponseWriter: w,
@@ -86,6 +86,7 @@ func (h *Dnstap) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 	// The query tap message should be sent before sending the query to the
 	// forwarder. Otherwise, the tap messages will come out out of order.
 	if h.MessageTypeEnabled(tap.Message_CLIENT_QUERY) {
+		qt := time.Now()
 		h.tapClientQuery(ctx, w, r, qt)
 	}
 	return plugin.NextOrFailure(h.Name(), h.Next, ctx, w, r)
--- a/plugin/dnstap/handler.go
+++ b/plugin/dnstap/handler.go
@@ -24,7 +24,7 @@ type Dnstap struct {
 	Identity          []byte
 	Version           []byte
 	ExtraFormat       string
-	// enabledMessageTypes is a bitfield of enabled tap.Message_Types.
+	// enabledMessageTypes is a bitmap of enabled tap.Message_Types.
 	// There's 14 message types in total, so uint64 is enough to store all of them.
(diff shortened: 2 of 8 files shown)


plugin/dnstap/handler.go
```diff
@@ -66,20 +71,31 @@ func (h Dnstap) tapQuery(ctx context.Context, w dns.ResponseWriter, query *dns.M
 
 // ServeDNS logs the client query and response to dnstap and passes the dnstap Context.
 func (h Dnstap) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
-	rw := &ResponseWriter{
-		ResponseWriter: w,
-		Dnstap:         h,
-		query:          r,
-		ctx:            ctx,
-		queryTime:      time.Now(),
+	qt := time.Now()
```
codefromthecrypt: if we are filtering.. should we only do time.Now inside the ifs? that would reduce overhead when we know it won't be sent

plugin/dnstap/handler.go
```diff
@@ -66,20 +71,30 @@ func (h *Dnstap) tapQuery(ctx context.Context, w dns.ResponseWriter, query *dns.
...
-		ctx:            ctx,
-		queryTime:      time.Now(),
+	if h.MessageTypeEnabled(tap.Message_CLIENT_RESPONSE) {
+		// Custom ResponseWriter is only used to tap CLIENT_RESPONSE messages, so we only create it if needed.
+		w = &ResponseWriter{
+			ResponseWriter: w,
+			Dnstap:         h,
+			query:          r,
+			ctx:            ctx,
+			queryTime:      time.Now(),
```
rdrozhdzh: would be good to have here the same timestamp as at the line 88, i.e. call the `time.Now()` once, and use the same value in both places.
External DNSTAP processing software may rely on this timestamp to find the correlation between CR and CQ messages if both are enabled.

plugin/dnstap/handler_test.go
```diff
@@ -134,3 +135,20 @@ func TestTapMessage(t *testing.T) {
...
+			t.Errorf("Expected %s to be enabled", name)
+		}
+	}
+	h.enabledMessageTypes = parseMessageTypes("CLIENT_QUERY CLIENT_RESPONSE")
+	if !h.MessageTypeEnabled(tap.Message_CLIENT_QUERY) {
+		t.Error("Expected CLIENT_QUERY to be enabled")
+	}
+	if !h.MessageTypeEnabled(tap.Message_CLIENT_RESPONSE) {
+		t.Error("Expected CLIENT_RESPONSE to be enabled")
+	}
```
rdrozhdzh, line 153: please add at least one negative case, like
```go
	if h.MessageTypeEnabled(tap.Message_AUTH_RESPONSE) {
		t.Error("Expected AUTH_RESPONSE to be disabled")
	}
```

plugin/dnstap/setup.go
```diff
@@ -138,3 +151,21 @@ func setup(c *caddy.Controller) error {
 
 	return nil
 }
+
+// parseMessageTypes parses a space-separated list of message types and returns a bitmap of the tap.Message_Type values.
+func parseMessageTypes(raw string) uint64 {
+	var bitmap uint64
+	for _, str := range strings.Split(raw, " ") {
```
rdrozhdzh: what if user inserts two spaces between message types?
I think it would be better to delegate extracting the arguments to caddy lib, e.g. you can use the `caddy.Controller.RemainingArgs()` in `func parseConfig()` like here
https://github.com/coredns/coredns/blob/a7ed346585e30b99317d36e4d007b7b19a228ea5/plugin/forward/setup.go#L173-L177

plugin/dnstap/setup_test.go
```diff
@@ -135,3 +140,27 @@ func TestMultiDnstap(t *testing.T) {
...
+func Test_parseMessageTypes(t *testing.T) {
+	tests := []struct {
+		in     string
+		expect uint64
+	}{
+		{in: "", expect: ^uint64(0)},
+		{in: "CLIENT_QUERY", expect: 1 << tap.Message_CLIENT_QUERY},
+		{
+			in:     "CLIENT_QUERY CLIENT_RESPONSE",
+			expect: (1 << tap.Message_CLIENT_QUERY) | (1 << tap.Message_CLIENT_RESPONSE),
```
rdrozhdzh: Here you calculate the expected value the same way as you calculate it in the function you are testing. What if this calculation is wrong? Your test will not be able to catch this. I would just hardcode the expected value here.

On the other hand, the values like `Message_CLIENT_QUERY` are controlled by external project. If we hardcode the expected value, one day this test may surprisingly fail due to updates in external package which is not good. Though, I think it's unlikely that they (github.com/dnstap/golang-dnstap) will changes the message type constants.

plugin/forward/dnstap.go
```diff
@@ -34,33 +34,37 @@ func toDnstap(ctx context.Context, f *Forward, host string, state request.Reques
...
-		q := new(tap.Message)
-		msg.SetQueryTime(q, start)
-		// Forwarder dnstap messages are from the perspective of the downstream server
-		// (upstream is the forward server)
-		msg.SetQueryAddress(q, state.W.RemoteAddr())
-		msg.SetResponseAddress(q, ta)
-		if t.IncludeRawMessage {
-			buf, _ := state.Req.Pack()
-			q.QueryMessage = buf
+		if t.MessageTypeEnabled(tap.Message_FORWARDER_QUERY) {
```
rdrozhdzh, line 37: It might make sense to check the enabled message types at very begin of this function and save resources on processing hosts, ports, etc. Or even better, you can check the types in
https://github.com/coredns/coredns/blob/a7ed346585e30b99317d36e4d007b7b19a228ea5/plugin/forward/forward.go#L76-L81
i.e. skip the tap plugin, if both FQ and FR messages are disabled

rdrozhdzh (commented):
https://github.com/coredns/coredns/blob/master/plugin/dnstap/README.md should be updated as well

plugin/dnstap/handler.go
mathetake: https://github.com/coredns/coredns/pull/6643/commits/4db7c5a7ffa3783d95c62ae1b89bb3e918293909

plugin/dnstap/handler_test.go
mathetake: https://github.com/coredns/coredns/pull/6643/commits/00edfa61b874bfc3e14e679abb76cb800cbb2705

plugin/dnstap/setup.go
mathetake: https://github.com/coredns/coredns/pull/6643/commits/e7ce1267050a9c3e69ec042d919bb35b67d7233a

plugin/dnstap/setup_test.go
mathetake: https://github.com/coredns/coredns/pull/6643/commits/6e9188f5af01462bdecb3f393c065cbf30d51854

plugin/forward/dnstap.go
mathetake: https://github.com/coredns/coredns/pull/6643/commits/109b2048ad82c8ecf5715e32f5f1347e8eb22ed4

mathetake:
@rdrozhdzh thank you for the thorough reviews. I think I addressed all your comments, so I would appreciate it if you can take a look again!

plugin/dnstap/README.md
```diff
@@ -29,6 +30,12 @@ dnstap SOCKET [full] {
 * **VERSION** to override the version field. Defaults to the CoreDNS version.
 * **EXTRA** to define "extra" field in dnstap payload, [metadata](../metadata/) replacement available here.
 * `skipverify` to skip tls verification during connection. Default to be secure
+* **MESSAGE_TYPES** is a space-separated list of [dnstap message types](https://github.com/dnstap/dnstap.pb/blob/1061e3ed4430f68a0adb87eecadbb9208e7b51dd/dnstap.proto#L156-L229) to be tapped.
+Defaults to tapping all message types. Currently, this plugin only supports the following message types: 
+  * `CLIENT_QUERY`
+  * `CLIENT_RESPONSE`
+  * `FORWARDER_QUERY`: only available when the forward plugin is used.
+  * `FORWARDER_RESPONSE`: only available when the forward plugin is used.
```
rdrozhdzh, line 38: you can mention that other custom plugins can implement support for other message types similarly to implementation in forward plugin.

plugin/dnstap/setup_test.go
```diff
@@ -44,14 +45,16 @@ func TestConfig(t *testing.T) {
                 identity NAME2
                 version VER2
                 extra EXTRA2
+				message_types CLIENT_RESPONSE CLIENT_QUERY
               }`, false, []results{
-			{"dnstap.sock", true, "unix", []byte("NAME"), []byte("VER"), "EXTRA"},
-			{"127.0.0.1:6000", false, "tcp", []byte("NAME2"), []byte("VER2"), "EXTRA2"},
+			{"dnstap.sock", true, "unix", []byte("NAME"), []byte("VER"), "EXTRA", defaultEnabledMessageTypes},
+			{"127.0.0.1:6000", false, "tcp", []byte("NAME2"), []byte("VER2"), "EXTRA2", 0b1100000},
```
rdrozhdzh: you can still add a comment here to designate which message types are encoded in bit mask

plugin/forward/forward_test.go
```diff
@@ -40,37 +38,148 @@ func TestList(t *testing.T) {
...
+		name,
+		tapConfig string
+		assert func(t *testing.T, src *dnstap.Dnstap, actualTaps []*dnstap.Dnstap)
+	}{
+		{
+			name:      "one dnstap without message_types",
+			tapConfig: `dnstap tcp://example.com:6000`,
+			assert: func(t *testing.T, src *dnstap.Dnstap, actualTaps []*dnstap.Dnstap) {
+				if len(actualTaps) != 1 {
+					t.Fatalf("Expected: 1 results, got: %v", len(actualTaps))
```
rdrozhdzh: here and in other places, let's replace the `t.Fatalf()` with `t.Errorf(); return` to continue running other test cases

plugin/forward/forward_test.go
```diff
@@ -40,37 +38,148 @@ func TestList(t *testing.T) {
...
+			tapConfig: `dnstap tcp://example.com:6000 {
+	message_types CLIENT_RESPONSE
+}
+dnstap /tmp/dnstap.sock full`,
+			assert: func(t *testing.T, src *dnstap.Dnstap, actualTaps []*dnstap.Dnstap) {
+				if len(actualTaps) != 1 {
+					t.Fatalf("Expected: 2 results, got: %v", len(actualTaps))
+				}
+				if actualTaps[0] != src.Next {
+					t.Error("Unexpected order of dnstap plugins")
```
rdrozhdzh: "Unexpected dnstap plugin"

plugin/forward/forward_test.go
```diff
@@ -40,37 +38,148 @@ func TestList(t *testing.T) {
...
+		},
+		{
+			name: "two dnstaps where one has only non-forward message_types",
+			tapConfig: `dnstap tcp://example.com:6000 {
+	message_types CLIENT_RESPONSE
+}
+dnstap /tmp/dnstap.sock full`,
+			assert: func(t *testing.T, src *dnstap.Dnstap, actualTaps []*dnstap.Dnstap) {
+				if len(actualTaps) != 1 {
+					t.Fatalf("Expected: 2 results, got: %v", len(actualTaps))
```
rdrozhdzh: Expected: 1 result

plugin/forward/forward_test.go
```diff
@@ -40,37 +38,148 @@ func TestList(t *testing.T) {
...
+				}
+				if actualTaps[0] != src.Next {
+					t.Error("Unexpected order of dnstap plugins")
+				}
+			},
+		},
+		{
+			name: "three dnstaps with only non-forward message types",
```
rdrozhdzh: two dnstaps

rdrozhdzh:
@mathetake , just curious, are you going to implement the dnstap message filtering based on the DNS response RCODE, as you mentioned in your original request?

mathetake:
@rdrozhdzh actually, I am not planning to introduce that RCODE based filter as that seems not worth it vs the complexity in the API.

rdrozhdzh (approved):
Thanks, looks good to me

mathetake:
kindly ping any codeowner for a quick look?

mathetake:
I _really_ hate to be nagging, and I understand all of you guys are busy doing other stuff, but by any chance could one of you code owners take a look? @johnbelamaric @miekg @Tantalor93 @chrisohaver @yongtang

mathetake:
sorry let me ping again - this is the last time I won't do that again.

mathetake:
Unfortunately at this point I have to think at the point CoreDNS is no longer willing to accept contributions any more despite all the helps from multiple people in the community. I'm really sorry to see this happen but I am no longer able to engage here. 😔 closing.

PR #6681, opened 2024-05-14
Add 'next' option to forward plugin

dihmandrake:
Allows the forward plugin to execute the next plugin based on the return code. Similar to the externally mainted alternate plugin https://github.com/coredns/alternate

Based on the idea of chrisohaver@ in https://github.com/coredns/coredns/pull/6549#issuecomment-2079950889

I am having issues adding a proper test for functionality. Primarily, I do not know the code base enough and having multiple `dnstest.NewServer` with ResponseWriter does not work. From my testing these are "Singletons'' and only the last defined response writer is used for all servers

This PR introduces an alternate option within the CoreDNS forward plugin, eliminating the need for separate compiled plugins. This option proves particularly beneficial for Kubernetes cluster operators managing internal DNS servers. These servers resolve specific domains locally, ensuring requests remain within the designated network or cluster. This approach enhances fault tolerance and serves other purposes. Additionally, the internal DNS servers may leverage tools like cert-manager, which necessitates external TXT record generation and resolution for the DNS-01 challenge mechanism.

The new alternate option within the forward plugin allows resolving entries exclusively available on public DNS servers. This enables mechanisms like cert-manager to function correctly without requiring operators to maintain self-compiled and hosted versions of CoreDNS. This approach simplifies deployment and maintenance while ensuring proper functionality.

PR: https://github.com/coredns/coredns/pull/6549

Added as part of this PR

No

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -43,6 +43,8 @@ type Forward struct {
 	from    string
 	ignored []string
 
+	nextAlternateRcodes []int
+
 	tlsConfig     *tls.Config
 	tlsServerName string
 	maxfails      uint32
@@ -192,6 +194,15 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 			formerr.SetRcode(state.Req, dns.RcodeFormatError)
 			w.WriteMsg(formerr)
 			return 0, nil
+		}
+
+		// Check if we have an alternate Rcode defined, check if we match on the code
+		for _, alternateRcode := range f.nextAlternateRcodes {
+			if alternateRcode == ret.Rcode && f.Next != nil { // In case we do not have a Next handler, just continue normally
+				if _, ok := f.Next.(*Forward); ok { // Only continue if the next forwarder is also a Forworder
+					return plugin.NextOrFailure(f.Name(), f.Next, ctx, w, r)
+				}
+			}
 		}
 
 		w.WriteMsg(ret)
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -50,6 +50,7 @@ forward FROM TO... {
     policy random|round_robin|sequential
     health_check DURATION [no_rec] [domain FQDN]
     max_concurrent MAX
+    next RCODE_1 [RCODE_2] [RCODE_3...]
 }
 ~~~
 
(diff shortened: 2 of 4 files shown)


plugin/forward/forward.go
```diff
@@ -194,6 +196,15 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 			return 0, nil
 		}
 
+		// Check if we have an alternate Rcode defined, check if we match on the code
+		for _, alternateRcode := range f.alternateRcodes {
+			if alternateRcode == ret.Rcode && f.Next != nil { // In case we do not have a Next handler, just continue normally
+				//				if _, ok := f.Next.(*Forward); ok { // Only continue if the next forwarder is also a Forworder
```
chrisohaver: This is how I was thinking it would work... only going to next if its another instance of the forward plugin.

chrisohaver (commented):
Thanks! I think `next` might be a better name for the option from an intuitive sense.
There is the history of the external plugin being called `alternate` ... and before that I think it was called `fallback`, but I dont think its important from a clarity pov to maintain the name going forward.

What do you think?

dihmandrake:
> Thanks! I think `next` might be a better name for the option from an intuitive sense. There is the history of the external plugin being called `alternate` ... and before that I think it was called `fallback`, but I dont think its important from a clarity pov to maintain the name going forward.
> 
> What do you think?

Hi,

thank you for the review and the suggestion. I do not have a strong feeling about naming. `alternate` might be more consistent while I understand that `next` might appear clearer. Since I do not have preference, I have changed the config value to `next` and noted also in the Readme that this will only work if the next handler is a `forward` plugin.
The only thing where I left `alternate` is in the variable name for `alternateRcodes` as I think it helps possible in the future with clarity.

Would you mind taking another look @chrisohaver

chrisohaver (approved):

Jamal-B:
@chrisohaver, thank you for the review and for the merge (and of course, thanks to @dihmandrake for the PR).
Do you have any idea when a version including this feature will be released?

chrisohaver:
It will be in the next release - date TBD.

Jonomir:
> It will be in the next release - date TBD.

Any idea when this will be?
I'm currently building something that would benefit a lot from this.

ThreeSixes:
+1 on wondering if there's a plan to release this and when.

PR #6692, opened 2024-05-20
Generate zplugin.go correctly with third-party plugins

### LINKED ISSUE #6691: zplugin.go isn't generated correctly with third-party plugins
**What happened**:

When running `make gen` or `go generate coredns.go` with a third-party external CoreDNS plugin enabled, the `zplugin.go` will cause the `TestImportOrdering` unit test to fail. This is because the logic in `directives_generate.go` does not separate third-party external plugins from CoreDNS plugins in the go import block. This leads to the `TestImportOrdering` unit test failing, as it requires that third-party imports paths should be in a separate import block.

`go generate coredns.go` should generate `zplugin.go` according to the standards enforced in `TestImportOrdering`.

1. Add an external plugin to CoreDNS by adding it to `plugin.cfg`, e.g. add `ocp_dnsnameresolver:github.com/openshift/coredns-ocp-dnsnameresolver` anywhere in `plugin.cfg`.
2. Run `TestImportOrdering` unit test:
   ```
   === RUN   TestImportOrdering
    presubmit_test.go:195: import path for "github.com/openshift/coredns-ocp-dnsnameresolver" is not of the same type 
   "coredns" in "/home/gspence/src/github.com/coredns/coredns/core/plugin/zplugin.go"
   --- FAIL: TestImportOrdering (0.13s)
   ```

While this issue does not affect the main CoreDNS repo, it can cause unit test failures in forks of CoreDNS that include external third-party plugins. So I understand it's a low priority.

**Environment**:

- the version of CoreDNS: 1.11.3
- Corefile: N/A
- logs, if applicable: N/A
- OS (e.g: `cat /etc/os-release`): N/A
- Others: N/A

gcs278:

Previously, the generation of zplugin.go would not separate third-party external plugins from CoreDNS plugins in the go import block. This leads to the TestImportOrdering unit test failing, as it requires that third-party imports paths to be in a separate import block.

While this issue does not affect the main CoreDNS repo, it can cause unit test failures in forks of CoreDNS that include external third-party plugins.

Fixes https://github.com/coredns/coredns/issues/6691

None

No


gcs278:
@chrisohaver or whoever appropriate, could I get a quick review? Should be low-risk as it doesn't impact this CoreDNS repo, but allows for forks to pass unit tests if they are adding 3rd party plugins.

PR #6695, opened 2024-05-21
Enhance CoreDNS A and AAAA methods to utilize upstream for A record resolution when plugin fails

### LINKED ISSUE #6694: A better way to handle plugin Lookup
## The background

I developed the `inner` plugin based on the CoreDNS plugin development mode within the company, used for handling DNS services for internal domain names of the company. The issues I encountered are as follows:

I added a CNAME record in the plugin: `m.xxx.com CNAME -> 12345.yyy.com`

Among them:

- `m.xxx.com` is the internal domain name of the company
- `12345.yyy.com` is the public domain name

My expectations are:

1. I only need to add a CNAME record in the `inner` plugin: `m.xxx.com CNAME -> 12345.yyy.com`.
2. `12345.yyy.com` is a domain name that exists on the public network, so theoretically I don't need to add an A record for `12345.yyy.com` in the `inner` plugin to complete the resolution.

However, the actual situation is:

- If the domain name `12345.yyy.com` is not added to the `inner` plugin. Then if I query m.xxx.com, CoreDNS will eventually fail to resolve and will not return any DNS records.
- If the domain name `12345.yyy.com` is added with relevant A records to the `inner` plugin. Then CoreDNS internally first queries CNAME records, then queries A records, and CoreDNS will eventually complete resolution. By using dig for querying, you will get 2 records: one CNAME record and one A record.

I analyzed the code of CoreDNS and found a possible reason for this issue：

In `plugin/backend_lookup.go`, there is the following code segment:

```go
// A returns A records from Backend or an error.
func A(ctx context.Context, b ServiceBackend, zone string, state request.Request, previousRecords []dns.RR, opt Options) (records []dns.RR, truncated bool, err error) {
	services, err := checkForApex(ctx, b, zone, state, opt)
	if err != nil {
		return nil, false, err
	}

	dup := make(map[string]struct{})

	for _, serv := range services {
		what, ip := serv.HostType()

		switch what {
		case dns.TypeCNAME:
			if Name(state.Name()).Matches(dns.Fqdn(serv.Host)) {
				// x CNAME x is a direct loop, don't add those
				// in etcd/skydns w.x CNAME x is also direct loop due to the "recursive" nature of search results
				continue
			}

			newRecord := serv.NewCNAME(state.QName(), serv.Host)
			if len(previousRecords) > 7 {
				// don't add it, and just continue
				continue
			}
			if dnsutil.DuplicateCNAME(newRecord, previousRecords) {
				continue
			}
			fmt.Println("zone:", zone, "dns.Fqdn(serv.Host):", dns.Fqdn(serv.Host), "res", dns.IsSubDomain(zone, dns.Fqdn(serv.Host)))
			if dns.IsSubDomain(zone, dns.Fqdn(serv.Host)) {
				state1 := state.NewWithQuestion(serv.Host, state.QType())
				state1.Zone = zone
				nextRecords, tc, err := A(ctx, b, zone, state1, append(previousRecords, newRecord), opt)

				if err == nil {
					// Not only have we found something we should add the CNAME and the IP addresses.
					if len(nextRecords) > 0 {
						records = append(records, newRecord)
						records = append(records, nextRecords...)
					}
				}
				if tc {
					truncated = true
				}
				continue
			}
			// This means we can not complete the CNAME, try to look else where.
			target := newRecord.Target
			// Lookup
			m1, e1 := b.Lookup(ctx, state, target, state.QType())
			if e1 != nil {
				continue
			}
			if m1.Truncated {
				truncated = true
			}
			// Len(m1.Answer) > 0 here is well?
			records = append(records, newRecord)
			records = append(records, m1.Answer...)
			continue

		case dns.TypeA:
			if _, ok := dup[serv.Host]; !ok {
				dup[serv.Host] = struct{}{}
				records = append(records, serv.NewA(state.QName(), ip))
			}

		case dns.TypeAAAA:
			// nada
		}
	}
	return records, truncated, nil
}
```

The core logic is:

1. First, use checkForApex to query if there are corresponding DNS records in the plugin.
2. If there is a CNAME record, and the Zone to which the Value resolved by CNAME belongs can be handled by the current plugin, then query the A record corresponding to the CNAME record through the plugin.
3. If neither of the above 2 steps work, then use Lookup to query other plugins in a chained manner.

However, there is a preset logic here:

> If the Zone of a domain's CNAME record belongs to the current plugin's processing scope, then the A record of the CNAME Value must also exist in the plugin. Otherwise, an incorrect A record will be found. As a result, the entire resolution chain will be interrupted.

The code I tried to fix is roughly as follows:

```go
// A returns A records from Backend or an error.
func A(ctx context.Context, b ServiceBackend, zone string, state request.Request, previousRecords []dns.RR, opt Options) (records []dns.RR, truncated bool, err error) {
	services, err := checkForApex(ctx, b, zone, state, opt)
	if err != nil {
		return nil, false, err
	}

	dup := make(map[string]struct{})

	for _, serv := range services {
		what, ip := serv.HostType()

		switch what {
		case dns.TypeCNAME:
			if Name(state.Name()).Matches(dns.Fqdn(serv.Host)) {
				// x CNAME x is a direct loop, don't add those
				// in etcd/skydns w.x CNAME x is also direct loop due to the "recursive" nature of search results
				continue
			}

			newRecord := serv.NewCNAME(state.QName(), serv.Host)
			if len(previousRecords) > 7 {
				// don't add it, and just continue
				continue
			}
			if dnsutil.DuplicateCNAME(newRecord, previousRecords) {
				continue
			}
                        // Here! Add an identifier here to indicate that if the current plugin cannot resolve the CNAME record, it needs to be processed further through the upstream of the plugin.
			needLookup := true
			if dns.IsSubDomain(zone, dns.Fqdn(serv.Host)) {
				state1 := state.NewWithQuestion(serv.Host, state.QType())
				state1.Zone = zone
				nextRecords, tc, err := A(ctx, b, zone, state1, append(previousRecords, newRecord), opt)

				if err == nil {
					// Not only have we found something we should add the CNAME and the IP addresses.
					if len(nextRecords) > 0 {
						records = append(records, newRecord)
						records = append(records, nextRecords...)
					}
					needLookup = false
					}
				if tc {
					truncated = true
				}
			}
                         
			if needLookup {
				// This means we can not complete the CNAME, try to look else where.
				target := newRecord.Target
				// Lookup
				m1, e1 := b.Lookup(ctx, state, target, state.QType())
				if e1 != nil {
					continue
				}
				if m1.Truncated {
					truncated = true
				}
				// Len(m1.Answer) > 0 here is well?
				records = append(records, newRecord)
				records = append(records, m1.Answer...)
			}

		case dns.TypeA:
			if _, ok := dup[serv.Host]; !ok {
				dup[serv.Host] = struct{}{}
				records = append(records, serv.NewA(state.QName(), ip))
			}

		case dns.TypeAAAA:
			// nada
		}
	}
	return records, truncated, nil
}
```

hansedong:

try to fix #6694 

#6694 

none

no

### DIFF
--- a/plugin/rewrite/README.md
+++ b/plugin/rewrite/README.md
@@ -51,6 +51,7 @@ will behave as follows:
 
    * `continue` will continue applying the next rule in the rule list.
    * `stop` will consider the current rule the last rule and will not continue.  The default behaviour is `stop`
+   * When multiple rules are matched, the request rewrite follows the line order in the configuration, while the response rewrite(`answer` option) is executed in reverse order.
 
 ## Examples
 
--- a/plugin/rewrite/reverter.go
+++ b/plugin/rewrite/reverter.go
@@ -90,8 +90,9 @@ func (r *ResponseReverter) WriteMsg(res1 *dns.Msg) error {
 }
 
 func (r *ResponseReverter) rewriteResourceRecord(res *dns.Msg, rr dns.RR) {
-	for _, rule := range r.ResponseRules {
-		rule.RewriteResponse(res, rr)
+	// The reverting rules need to be done in reversed order.
+	for i := len(r.ResponseRules) - 1; i >= 0; i-- {
+		r.ResponseRules[i].RewriteResponse(res, rr)
 	}
 }
 
--- a/plugin/rewrite/reverter_test.go
+++ b/plugin/rewrite/reverter_test.go
@@ -126,6 +126,24 @@ func TestValueResponseReverter(t *testing.T) {
 	rules = append(rules, r)
 
 	doValueReverterTests("suffix", rules, t)
+
+	// multiple rules
+	rules = []Rule{}
+	r, err = newNameRule("continue", "suffix", `.domain.uk`, ".domain.us", "answer", "auto")
+	if err != nil {
+		t.Errorf("cannot parse rule: %s", err)
+		return
+	}
+	rules = append(rules, r)
+
+	r, err = newNameRule("stop", "suffix", `.domain.us`, ".cluster.local", "answer", "auto")
+	if err != nil {
+		t.Errorf("cannot parse rule: %s", err)
+		return
+	}
+	rules = append(rules, r)
+
+	doValueReverterTests("suffix_multiple", rules, t)
 }
 
 func doValueReverterTests(name string, rules []Rule, t *testing.T) {


hansedong:
@chrisohaver @miekg 
Could you please take the time to review whether there is anything that needs further discussion about the problem this PR is trying to solve?

chrisohaver:
You’re in the queue and I will review as time allows if another maintainer does not review first.

Issue #6661, opened 2024-05-01
Missing Docker image for 1.11.3


chrisohaver:
The docker image for CoreDNS release 1.11.3 is not available due to persistent issues in the CoreDNS release process.
Once these issue are resolved, the image will be made available.

jauderho:
For some reason, this is also not being tagged as the latest release when you go to the releases page.

chrisohaver:
> For some reason, this is also not being tagged as the latest release when you go to the releases page.

This because 1.11.3 is not fully released.  The docker images are not published, so the release is not available.  The latest release actually _available_ is still 1.11.1.

dafanasiev:
NOTE: right now there is no way to detect `latest` release using github api releases-list endpoint (`https://api.github.com/repos/coredns/coredns/releases`) because both `draft` and `prerelease` are `false`:

```json
# curl https://api.github.com/repos/coredns/coredns/releases | jq '.[] | select(.name=="v1.11.3")'

{
  "url": "https://api.github.com/repos/coredns/coredns/releases/153656321",
  "assets_url": "https://api.github.com/repos/coredns/coredns/releases/153656321/assets",
  "upload_url": "https://uploads.github.com/repos/coredns/coredns/releases/153656321/assets{?name,label}",
  "html_url": "https://github.com/coredns/coredns/releases/tag/v1.11.3",
  "id": 153656321,
  "author": {....},
  "node_id": "RE_kwDOAzt_0s4JKJwB",
  "tag_name": "v1.11.3",
  "target_commitish": "a7ed346585e30b99317d36e4d007b7b19a228ea5",
  "name": "v1.11.3",
  "draft": false,
  "prerelease": false,
  "created_at": "2024-04-26T19:08:47Z",
  "published_at": "2024-05-01T12:01:11Z",
  "assets": [....],
  "tarball_url": "https://api.github.com/repos/coredns/coredns/tarball/v1.11.3",
  "zipball_url": "https://api.github.com/repos/coredns/coredns/zipball/v1.11.3",
[...]

Issue #6638, opened 2024-04-24
CoreDNS 1.11.3 Release


chrisohaver:
Issue for tracking the 1.11.3 release.

chrisohaver:
If someone has time to grab the reins on this, please do.
If not, I will get to it as time permits.

tedaford:
I put together a PR basically following the format of the previous release, feel free to edit or merge. Just figured I'd help speed things along, I'm waiting on this release for a feature, haha.

https://github.com/coredns/coredns/pull/6639

chrisohaver:
Cutting release now...

chrisohaver:
Docker publish failed:  Authentication failure (same failure as in 1.11.2)

chrisohaver:
**Please note: The docker image is not available due to ongoing issues with the release process.**

https://github.com/coredns/coredns/issues/6661

zhangguanzhang:
any update? There are many cve vulnerabilities in version `1.11.1`

toredash:
Can be closed?

PR #6723, opened 2024-06-03
Use dockerhub token instead of password

heiko-braun:
It's an attempt to fix (or help with) #6661 

#6661 
#6638

The release process docs would need be changed to mention that a personal access token for dockerhub is needed: https://github.com/docker/login-action?tab=readme-ov-file#docker-hub

Not that I am aware of

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       DOCKER_LOGIN: ${{ secrets.DOCKERHUB_USERNAME }}
-      DOCKER_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
+      DOCKER_PASSWORD: ${{ secrets.DOCKERHUB_TOKEN }}
       RELEASE: ${{ github.event.inputs.release || github.event.release.tag_name }}
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


heiko-braun:
@chrisohaver You would need to create a dockerhub token and update the action secrets accordingly

chrisohaver:
Thank You!  I am not an admin in the coredns dockerhub org,  so I don't have permission to generate a token for it.
Strictly speaking if this fix works this PR is not necessary. All we need to do is find someone who is an admin in coredns dockerhub org and have them generate a token and update the existing secrets.

heiko-braun:
I didn't look into the login-action deeply, but you might be right. Eventually we can keep DOCKERHUB_USERNAME and simply update it. I have updated it to TOKEN, since it's seems to be best practice.

johnbelamaric:
> Thank You! I am not an admin in the coredns dockerhub org, so I don't have permission to generate a token for it. Strictly speaking if this fix works this PR is not necessary. All we need to do is find someone who is an admin in coredns dockerhub org and have them generate a token and update the existing secrets.

Chris, I am an owner of that repo. What do you need? I can also make you an admin and/or owner.

chrisohaver:
@johnbelamaric thanks John! For now, can you generate a token for your account and update the existing dockerhub_password secret to that token (not your password). Hopefully that works!

thanks!

PR #6731, opened 2024-06-10
fix(cwd): Restored backwards compatibility of Current Workdir

wmkuipers:
This merge requests restores the default working directory for the coredns container that now runs nonroot.

An issue has not been opened yet, after implementation of https://github.com/coredns/coredns/commit/d21537f93197f82f9abd66569ea78f0fe629061f the base-image of the container has changed, resulting in the Working Directory being changed from `/` to `/home/nonroot`. This causes the container to, by default, no longer use `/Corefile` as a default conf, but instead use `/home/nonroot/Corefile`

None, the documentation already suggests is runs from /Corefile

It fixes backwards incompatibility.[ In the PR of the commit](https://github.com/coredns/coredns/pull/5969), it was suggested that there would be no backwards compatibility breaking, however this change was breaking the default config file location.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,5 +18,6 @@ FROM --platform=$TARGETPLATFORM ${BASE}
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY --from=build /coredns /coredns
 USER nonroot:nonroot
+WORKDIR /
 EXPOSE 53 53/udp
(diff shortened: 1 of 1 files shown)


rjacobs90:
Nice finding @wmkuipers!!! 👍

chrisohaver (approved):

chrisohaver:
Thanks for the fix!

PR #6753, opened 2024-06-20
`root`: explicit mark `dnssec` support

NicolaiSoeborg:

Improve documentation. DNSSEC keys can be specified relative to `root`: https://github.com/coredns/coredns/blob/master/plugin/dnssec/setup.go#L135

Didn't open issue for this

Aligns docs with code

Nope


mariuskimmina (approved):
seems to be correct

PR #6764, opened 2024-06-27
core: set cache-control max-age as integer, not float

balboah:
## What

When CoreDNS is used as a dns-over-https server, it sets cache control headers.
However, the max-age is set as a float, instead of an integer as specified in RFC 2616.

## How

This PR truncates the float into an integer, using `uint32` which is also what the dnsutil is using.

## Why

Setting the max-age as a float can make the value get ignored in some cases, such as when fronted by a caching CDN.


chrisohaver:
LGTM thanks!  Can you sign the commit to the DCO check passes? Thanks!

balboah:
replaced with signed commit, hope everything checks out!

PR #6765, opened 2024-06-28
dns.IsSubDomain is Inefficient methods

cedar-gao:

### DIFF
--- a/plugin/normalize.go
+++ b/plugin/normalize.go
@@ -25,7 +25,7 @@ type Zones []string
 func (z Zones) Matches(qname string) string {
 	zone := ""
 	for _, zname := range z {
-		if dns.IsSubDomain(zname, qname) {
+		if IsSubDomain(zname, qname) {
 			// We want the *longest* matching zone, otherwise we may end up in a parent
 			if len(zname) > len(zone) {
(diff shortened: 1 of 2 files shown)


plugin/normalize.go
```diff
@@ -35,6 +35,21 @@ func (z Zones) Matches(qname string) string {
 	return zone
 }
 
+// dns.IsSubDomain is Inefficient methods
+func IsSubDomain(zone, qname string) bool {
```
Tantalor93, line 39: I think that community would benefit more if you would rather optimize the dns.IsSubDomain in https://github.com/miekg/dns than optimize it here

plugin/normalize_test.go
```diff
@@ -138,3 +140,23 @@ func TestHostNormalizeExact(t *testing.T) {
 		}
 	}
 }
+func TestIsSubDomain(t *testing.T) {
```
Tantalor93, line 143: this should have been rather Benchmark test https://pkg.go.dev/testing#hdr-Benchmarks

SuperQ (changes requested):
This change needs to be made upstream in https://github.com/miekg/dns.

PR #6774, opened 2024-07-07
plugin/tls: add automated certificates via acme

mariuskimmina:
Similar to https://github.com/coredns/coredns/pull/5608 this PR aims to automate the process of obtaining TLS certificates to make it easier for users to serve DNS over TLS.

The difference between the original PR and this PR is that with this implementation the CoreDNS server itself does not need to be exposed to the Internet. It uses the APIs of other publicly available DNS providers to solve the DNS01 Challenge. That is currently the only automated way this implementation supports for obtaining certificates. Setting certificates manually will of course also continue to work. This change is based on a discussion with @SuperQ on the original PR. 

**A note on using certmagic over lego**: I tried to use lego this time at first but since it only concerns itself with obtaning the certificate and does not offer any aid in managing the certificate afterwards, I found that using certmagic would be easier to maintain. I also saw the ACME implementation in Traefik, which is build on top of lego, and while it seems very mature I ultimately found it simpler to stick with certmagic.

https://github.com/coredns/coredns/issues/3460

### 3. Things left to do
There are a couple things left to do here but I wanted to open the PR already anyway to gather some feedback.

- [ ] Update the documentation of the tls plugin
- [ ] Add more provider
- [ ] Add tests
- [ ] Clean up logs

### 4. Limitations

- This implementation currently does not support obtaining certificates for more than 1 domain
- CoreDNS has to restart itself after a renewing a certificate

AFAIK both of these limitations are also true when using the current TLS plugin so they shouldn't be a deal breaker.

### 5. Example Corefile
Here is a Corefile showing how this would work - I own the `coredns-acme.xyz` domain and have it registered at digitalocean. I also have to set the `DO_AUTH_TOKEN` environment variable to authenticate to digitalocean.

```
tls://.:5555 {
    debug
    tls acme {
        domain coredns-acme.xyz
        dnsProvider digitalocean
        email mar.kimmina@gmail.com
    }

    forward . tls://9.9.9.9 {
      tls_servername dns.quad9.net
    }
}
```

Digitalocean is currently the only provider I tested with (and implemented, see providers.go) I plan to add all the providers from this list in the next few days tho, just a bit of grind work https://caddyserver.com/docs/json/admin/identity/issuers/acme/challenges/dns/provider/

An alternative to this giant switch case statement would be to do it the same way caddy apparently does it which is to offer custom builds for each dns provider (https://caddy.community/t/how-to-use-dns-provider-modules-in-caddy-2/8148) that would of course keep the binary smaller but it's also somewhat inconvenient.  I currently don't think packing all the provider implementations is an issue but that can be up for discussion.


plugin/tls/provider.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/libdns/netlify"
+	"github.com/libdns/powerdns"
+	"github.com/libdns/route53"
+	"github.com/libdns/vercel"
+)
+
+func createDNSSolver(provider string) *certmagic.DNS01Solver {
+	var dnsSolver *certmagic.DNS01Solver
+	switch provider {
+	case "digitalocean":
```
mariuskimmina, line 26: Okay I am starting to see why caddy doesn't pack all the providers - I haven't added all of them and the binary is already 13M bigger than before. 

Any thoughts on this @SuperQ @chrisohaver ?

Does it make sense to limit this to only the most popular ones (and if so who decides which are those)?

mariuskimmina: Perhaps it could make sense to provide these as external plugins, one for each provider, this would keep the binary size small. The only downside I see with that is that it would be a lot of code duplication as these would be mostly identical. 

Users would then just have to pick the plugin for their dns provider

mariuskimmina: Created one as an example: https://github.com/mariuskimmina/acme_digitalocean

PR #6776, opened 2024-07-08
fix: plugin/kubernetes: Fix autopath for host network pods

### LINKED ISSUE #4299: Enables server-side domain search path completion for pods running in the host network
Currently, we could use autopath plugin together with kubernetes plugin to enable the server-side domain search path completion in Kubernetes clusters. However, it doesn't support the pods running in the host network due to no private ip.
Is there any ideas to support it? Is it reasonable to get search path from the dns request directly? The reason we need the private ip is to find the namespace a pod running in. However, we may get it from the dns request(e.g. XX.namespaces.svc.cluster.local) and we could check whether it existed or not in kubernetes cluster also.

**Why is this needed**:
 It would reduce the latency of the external domain name resolution for the pods running in the host network and in many scenarios, it needs to run the container in host mode.

Adrian-St:

This PR fixes the issue that the autopath plugin isn't working for pods running in the host network.
The reason this is marked as a fix and not a feature is because the current behaviour is causing issues in combination with caching and has inconsistencies.

Currently the AutoPath function fetches the pods by request IP to check where the request is coming from. For pods running in the host network this would be the Node IP. So if you have multiple pods running in host networking it would return multiple pods but only the first one is being used. This means that autopath might work for a pod running on one node because it happens to be the first one in the list, but it doesn't work on another node. For normal pods in the same namespace the same DNS query would also work with autopath.

In itself this is not a big issue, but in combination with caching this can cause lookup failures because A records and AAAA records are cached differently and in our resolvers if one of them returns an answer and the other one doesn't then the whole query fails with NXDOMAIN and it doesn't try the rest of the search path.

This implementation takes all host network pods on the node, matches the query against them and takes the namespace from the pod that matches. Of course any pod could request a DNS name in any namespace, but I don't really see why that is an issue, since the client would have to actively try to break DNS lookup.

The only potential issue I see is if the first query is lost and the second query is matched by accident, it could result in a wrong "base".
Example: If there is a namespace called "other" and a pod in this namespace does a DNS request for "example.other", the first query would be "example.other.other.svc.cluster.local" and the second "example.other.svc.cluster.local". Now if the first query is lost, the second query would use "example" as a base and append the items from the search path, which could result on wrong DNS responses.
But this was already an issue before, this implementation just slightly increases the risk of this happening because it would also apply to other host network pods on the same node as the pod in the "other" namespace.

This would fix: https://github.com/coredns/coredns/issues/4299

None that I found, currently it's undocumented that host network pods don't work with autopath as far as I can tell.
 
No, unless clients rely on the inconsistent behaviour of autopath, which would be strange.

### DIFF
--- a/plugin/kubernetes/autopath.go
+++ b/plugin/kubernetes/autopath.go
@@ -1,6 +1,10 @@
 package kubernetes
 
 import (
+	"fmt"
+	"regexp"
+	"strings"
+
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/kubernetes/object"
 	"github.com/coredns/coredns/request"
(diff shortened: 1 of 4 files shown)


plugin/kubernetes/autopath.go
```diff
@@ -49,14 +75,27 @@ func (k *Kubernetes) AutoPath(state request.Request) []string {
...
+func namespaceRegex(pods []*object.Pod) string {
+	nsMap := map[string]struct{}{}
+	for _, pod := range pods {
+		nsMap[pod.Namespace] = struct{}{}
+	}
+	var namespaces []string
+	for ns := range nsMap {
+		namespaces = append(namespaces, ns)
+	}
+	return fmt.Sprintf("(%s)", strings.Join(namespaces, "|"))
```
chrisohaver, line 87: Can this be optimized as a single loop?

plugin/kubernetes/autopath.go
```diff
@@ -27,19 +31,41 @@ func (k *Kubernetes) AutoPath(state request.Request) []string {
...
+		namespace = pods[0].Namespace
+	} else {
+		var expr *regexp.Regexp
+		nsMatcher := namespaceRegex(pods)
+		if zone == "." {
+			expr = regexp.MustCompile(fmt.Sprintf(`.+\.%s\.svc\.`, nsMatcher))
+		} else {
+			regexSafeZone := strings.ReplaceAll(zone, ".", "\\.")
+			expr = regexp.MustCompile(fmt.Sprintf(`.+\.%s\.svc\.%s`, nsMatcher, regexSafeZone))
+		}
```
chrisohaver, line 52: I think the regex expressions need to be anchored, else they might match a subdomain of a domain that would not otherwise match.

plugin/kubernetes/autopath.go
```diff
@@ -27,19 +31,41 @@ func (k *Kubernetes) AutoPath(state request.Request) []string {
...
+		matches := expr.FindStringSubmatch(state.Name())
```
chrisohaver, line 53: I suspect that doing this without regex may be more performant.

chrisohaver (commented):
Thanks for your contribution!

PR #6783, opened 2024-07-09
fix: multiarch builds and multiple repos using buildx

### LINKED ISSUE #6661: Missing Docker image for 1.11.3

grumps:

This should help finally getting v1.11.3 images built and released. Once the setup of ghcr.io packages and fixing the docker hub token is resolved.

This PR fixes code and reliability issues of getting images built and pushed for multi-arch support. It also adds support for ghcr.io. However it *will not fix the token for the Docker Hub registry*. Additionally there's repo configuration steps required to get ghcr.io to work as well.

Refactors `Makefile.docker` to use buildx and to support multiple repos.

`docker buildx build` removes the need to push multiple tags, use `docker manifest create` and delete tags so that all arches appear on a single tag. This was not feasible to complete when adding ghcr.io hosted images.

Updates to Dockerfile to use TARGETPLATFORM to pickup the correct binary.

Updates to E2E tests for new target and test mode.

Example Images
- [ghcr.io images](https://github.com/grumps/coredns/pkgs/container/coredns)
- [hub.docker.com](https://hub.docker.com/repository/docker/grumps/coredns/general)

Example Job:
- [Docker release (action)](https://github.com/grumps/coredns/actions/runs/9851072955/job/27197308346)

ref: #6661

Maybe update the image can now be pulled from ghcr.io once this is merged.

No.

One can't simply merge this PR and resolve #6661.
1. An admin for coredns needs to get a new token form [hub.docker.com](https://hub.docker.com/) then add it to the coredns secrets. e.g. 

2. Make sure that actions have read/write permissions
write-workflow
3. Allow actions to write to packages. This is the part L'm not 100% sure the exact steps here. I found that I had to manually push a container so that I could link it to the coredns fork on the `grumps` [packages page.](https://github.com/grumps?tab=packages). You will also need to make sure that it's public (the default to private). i expect this to cause the docker release action/job to fail a few times since this would be the coredns orgs first github package and I found the docs lacking how to do this the first time.

Also there's an test failing but it's been failing on master prior to my commit.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -11,6 +11,7 @@ on:
 
 permissions:
   contents: read
+  packages: write
 
 jobs:
   docker-release:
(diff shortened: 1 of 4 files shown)


chrisohaver:
I’d like to get the token issue resolved first. Then we can evaluate these changes.

Issue #6803, opened 2024-07-31
[Forward plugin] Imbalance with multiple servers behind single cluster IP


Arkelenia:
**What happened**:

We have a setup with a coredns running on each node in a Kubernetes cluster, acting as a node local cache and a Kubernetes-cluster-wide coredns cluster serving as the authority for our internal domain. The latter is behind a single cluster IP and our node local cache is configured with this single target in the forward plugin configuration.

After upgrading our authoritative coredns cluster from 1.8.1 to 1.11.3, we experienced a lot of imbalance in traffic. There are a few factors causing this issue:
- A [regression](https://github.com/coredns/coredns/commit/e7ad486b50d8f831b7dd4a0840d300a3bcfac471#diff-5da03c210bd07af1f13de1b99491ce78028868ad9a981c30184792a60ec565e0R154) that caused a coredns server to keep connections open forever instead of closing them after receiving 128 DNS queries
- The proxy is inherently unstable because its [connection pool favors the last used connection](https://github.com/coredns/coredns/blob/master/plugin/pkg/proxy/persistent.go#L58), which is often the one to the slowest server, and consequently the one to the most loaded server
- The client only resets connections after being idle for too long, which means some connections can have an infinite lifespan

**What you expected to happen**:

We expected our authoritative coredns servers to handle approximately the same amount of traffic. Those screenshots show the number of requests handled by different servers in the same Kubernetes cluster before and after the upgrade:

<img width="1464" alt="Screenshot 2024-07-30 at 8 47 34 PM" src="https://github.com/user-attachments/assets/eefc9f9d-144a-4ab1-a7db-7d5718b03d16">

**How to reproduce it (as minimally and precisely as possible)**:

You can set up a Kubernetes deployment running 2 pods with coredns 1.11.3 configured with the Kubernetes plugin:

```
cluster.local {
    kubernetes *.cluster.local. {
        pods disabled
    }
}
```

You can set up a Kubernetes service providing a static cluster IP:

```
apiVersion: v1
kind: Service
metadata:
  name: cluster-dns
  namespace: default
spec:
  type: ClusterIP
  clusterIP: 172.16.0.10
  selector:
    app: coredns
    service: cluster-dns
  ports:
  - name: dns
    port: 53
    protocol: UDP
    targetPort: 53
  - name: dns-tcp
    port: 53
    protocol: TCP
    targetPort: 53
```

Then you can set up a second deployment (or a daemonset) simulating the node local cache configured with:

```
[...]

PR #6804, opened 2024-07-31
[Forward plugin] Improve load balancing in connection pool

Arkelenia:

More details are available in the issue. In a nutshell, when multiple servers back a single destination endpoint (through a Kubernetes clusterIP, for example) in the forward plugin, the traffic is very unbalanced. The connection pool favors slow servers when returning cached connections.

This PR proposes the following changes:
- Change from LIFO to FIFO behavior when returning a cached connection from the pool.
- Add two explicit parameters: `idle_timeout` and `max_connection_age`. `expire` is deprecated in favor of `idle_timeout`, hence, `idle_timeout` takes precedence if both are defined but `expire` can still be used. For backward compatibility, `max_connection_age` defaults to infinite if not defined.

This PR is only meant as a suggestion to gather feedback.

This PR is a suggestion for this [issue](https://github.com/coredns/coredns/issues/6803).

Document the new parameters `idle_timeout` and `max_connection_age` and the "deprecation" of `expire`

This PR doesn't introduce a breaking change because an existing configuration would still work after applying this change.

### DIFF
--- a/plugin/pkg/proxy/persistent.go
+++ b/plugin/pkg/proxy/persistent.go
@@ -2,7 +2,6 @@ package proxy
 
 import (
 	"crypto/tls"
-	"sort"
 	"time"
 
 	"github.com/miekg/dns"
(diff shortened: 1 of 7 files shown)


plugin/pkg/proxy/persistent.go
```diff
@@ -46,78 +47,49 @@ func newTransport(proxyName, addr string) *Transport {
...
-			continue
-		}
-		if all {
-			t.conns[transtype] = nil
-			// now, the connections being passed to closeConns() are not reachable from
-			// transport methods anymore. So, it's safe to close them in a separate goroutine
-			go closeConns(stack)
+func (t *Transport) popValidConn(transtype transportType) *persistConn {
+	now := time.Now()
+	for idx, pc := range t.conns[transtype] {
```
Arkelenia, line 72: This becomes O(n) instead of the previous O(1) fetching of a connection. On normal operations, I would expect most connections to be valid and only needing to iterate through a few connections before finding a valid one. Benchmarks indicate that each step of the loop is about 5ns.

thevilledev (changes requested):
Stale PR and conflicts with the changes introduced in #7970. The conn manager no longer exists and the proxy implementation now uses a connection pool.

PR #6807, opened 2024-08-01
Add configuration knob for max queries per connection

Arkelenia:

Coredns underlying DNS library exposes a `MaxTCPQueries` parameter that controls how many queries can be sent on a given connection and defaults to 128 if unset. This previous change to coredns changed the default to `-1` (unlimited). This PR is an attempt at making this value configurable.

This new parameter is added in the `timeouts` plugin because it is the closest plugin I could think of but I might be wrong.

This is related to the regression mentioned in this issue https://github.com/coredns/coredns/issues/6803

The documentation of the `timeouts` plugin needs to be changed to add the new parameter.

It doesn't since the new parameter still defaults to "unlimited".

### DIFF
--- a/core/dnsserver/config.go
+++ b/core/dnsserver/config.go
@@ -54,10 +54,11 @@ type Config struct {
 	// TLSConfig when listening for encrypted connections (gRPC, DNS-over-TLS).
 	TLSConfig *tls.Config
 
-	// Timeouts for TCP, TLS and HTTPS servers.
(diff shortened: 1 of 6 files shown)


thevilledev:
Hello @Arkelenia! I think this is a valid thing still in 2026. May I ask you to rebase this PR and resolve merge conflicts? I'll park this as a draft in the meanwhile.

PR #6814, opened 2024-08-09
Upgrade antonmedv/expr to expr-lang/expr

lyda:

The expr project has moved.  This also upgrades from 1.15.5 to 1.16.9. The expr project lacks a changelog but tests pass and the changes don't seem to change it much - but there were a lot of changes.

None.

None.

It shouldn't.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ initWorkingDir: &initWorkingDir
 
 integrationDefaults: &integrationDefaults
   machine:
-    image: ubuntu-2004:2022.04.2
+    image: default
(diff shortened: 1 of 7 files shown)


chrisohaver (approved):
Thanks!

Tantalor93 (approved):

lyda:
I didn't know how else to goose the CircleCI run so I added a change to move to an image that wouldn't be deprecated at the end of the year.  But to get it to run on a medium instance I had to set it to default.  The pro is that it will never be deprecated. The con is that it might randomly upgrade and mess things up. I'm lazy and usually pick the option that upgrades automatically and then get burned by the Upgrade Gods (Cthulhu fhtagn) every few years but works out to less effort than manually upgrading. Obviously other opinions are valid.

PR #6862, opened 2024-09-03
Multi sub support

panic-kbutton:

In current form, the Azure plugin will parse subscriptions across many zones and apply the last zone's subscription ID to all plugin definitions. When used across multiple subscriptions with distinct resource group names (which most subscriptions in a large tenant will have), this causes 404s when interacting with the Azure API. DNS Clients must be created within the context of a subscription; this PR moves to a client per zone.

N/A

None - this corrects behavior to match expectations / existing documentation.

I don't believe so - multi-subscription support appears to be non-existent as is.

### DIFF
--- a/plugin/azure/setup.go
+++ b/plugin/azure/setup.go
@@ -10,8 +10,6 @@ import (
 	"github.com/coredns/coredns/plugin/pkg/fall"
 	clog "github.com/coredns/coredns/plugin/pkg/log"
 
-	publicAzureDNS "github.com/Azure/azure-sdk-for-go/profiles/latest/dns/mgmt/dns"
-	privateAzureDNS "github.com/Azure/azure-sdk-for-go/profiles/latest/privatedns/mgmt/privatedns"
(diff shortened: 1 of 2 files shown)


plugin/azure/setup.go
```diff
@@ -136,9 +122,12 @@ func parse(c *caddy.Controller) (auth.EnvironmentSettings, map[string][]string,
 				return env, resourceGroupMapping, accessMap, fall, c.Errf("unknown property: %q", c.Val())
 			}
 		}
+		for k, v := range resourceGroupMapping {
+			subResourceGroupMapping[env.Values[auth.SubscriptionID]+"/"+k] = v
```
alexzielenski, line 126: Rather than string manipulation I would make this key a strongly typed struct

alexzielenski (approved):
minor nit

panic-kbutton:
This PR can be considered obsolete if #6864 is accepted and merged

PR #6874, opened 2024-09-14
Optimize deletion just regenerate Names

chenylh:


chrisohaver (changes requested):
please add tests that check for edge cases:
removing first zone in slice
removing last zone in slice
removing only zone in slice

PR #6881, opened 2024-09-18
plugin/cache: Add option to handle 0 ttl, when responding from cache

Isaiselvan-Intelsat:

We need to avoid having 0ttl in the DNS responses. because some legacy devices/PC uses this 0 ttl as infinite wait. 

When serving from the cache, There is a background cache update taking place which would need some buffer time to complete. So by setting non 0 ttl value we can avoid more requests during this update transition. 

Not related to any issue. 

plugin/cache readme has been updated for this. 

Without the new option the functionality remains the same. So no backward compatibility issues.

### DIFF
--- a/plugin/cache/README.md
+++ b/plugin/cache/README.md
@@ -142,3 +142,13 @@ example.org {
     }
 }
 ~~~
+
+Enable zerottl for `example.org`, set 30seconds when the cache is served from the stale record:
+
+~~~ corefile
+example.org {
+    cache {
+        zerottl 30s
+    }
+}
+~~~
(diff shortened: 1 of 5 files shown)


chrisohaver (commented):
Are you using the `serve_stale` option?  IIUC, the only way cache returns a 0 TTL, is when it serves a stale item.

In that case, I think this should be a new parameter added to the `serve_stale` option instead of a separate option.

Isaiselvan-Intelsat:
Changes are incorporated in https://github.com/coredns/coredns/pull/8411
Hence closing the pull request.

Issue #5595, opened 2022-09-05
coredns doesn't perform better despite having more cores


gpl:
We are running CoreDNS 1.9.3 (retrieved from the official releases on GitHub), and have been having difficulty with increasing performance of a single instance of coredns.

With GOMAXPROCS set to 1, we observe ~60k qps and full utilization of one core.

With GOMAXPROCS set to 2, we seem to hit a performance limit of ~90-100k qps, but it consumes almost entirely two cores.

With GOMAXPROCS set to 4, we observe that coredns will use all 4 cores - but throughput does not increase, and latency seems to be the same.

With GOMAXPROCS set to 8-64, we observe the same CPU usage and throughput.

We have the following corefile:
```
.:55 {
  file db.example.org example.org
  cache 100
  whoami
}
```

db.example.org
```
$ORIGIN example.org.
@       3600 IN SOA sns.dns.icann.org. noc.dns.icann.org. 2017042745 7200 3600 1209600 3600
        3600 IN NS a.iana-servers.net.
        3600 IN NS b.iana-servers.net.

www     IN A     127.0.0.1
        IN AAAA  ::1
```

We are using `dnsperf`:  https://github.com/DNS-OARC/dnsperf

And the following command:
```
  dnsperf -d test.txt -s 127.0.0.1 -p 55 -Q 10000000 -c 1 -l 10000000 -S .1 -t 8
```
test.txt:
```
www.example.com AAAA
```

Is there anything we could be missing?

Thanks!

johnbelamaric:
Perhaps you are saturating the NIC throughput.

johnbelamaric:
I guess with local host that shouldn’t be the case.

Tantalor93:
Hello, If you could collect profiling data (CPU profile) exposed by pprof plugin, this could greatly benefit the investigation @gpl

gpl:
Attaching profiles for gomaxprocs 1,2,4,8,16.
[coredns-gomaxprocs.zip](https://github.com/coredns/coredns/files/9489572/coredns-gomaxprocs.zip)

Tantalor93:
Based on a quick look at the profiles, it seems that most of the CPU time serving DNS requests was spent on writing responses to the client
<img width="1679" alt="image" src="https://user-images.githubusercontent.com/5522817/189220405-d9423faf-53ef-4b4d-9eec-3586b081ebdc.png">


[...]

PR #6882, opened 2024-09-20
multisocket plugin

Shmillerov:

Added the ability to run multiple servers on the same port using the **multisocket** plugin. This increases the throughput of CoreDNS in environments with a large number of CPU cores.

Syntax:
```
multisocket [NUM_SOCKETS]
```

default value of NUM_SOCKETS is equal to GOMAXPROCS.

**Performance testing:**
https://github.com/coredns/coredns/pull/6882#discussion_r1818522596

https://github.com/coredns/coredns/issues/5595

I added a readme for the plugin. After merging to CoreDNS repo, need to re-generate https://coredns.io/plugins/ using https://github.com/coredns/coredns.io

Changes are backward compatible


SuperQ (dismissed):
~~I'm not sure we want to implement this. Multi-process makes a number of things much worse. For example monitoring metrics and similar things.~~

~~I think we should put more effort into making multi-threaded scaling better rather than paper over the issue with multiple processes.~~

chrisohaver:
This is more or less doable currently by running multiple instances of coredns.

johnbelamaric:
> I think we should put more effort into making multi-threaded scaling better rather than paper over the issue with multiple processes.

Isn't that what this does? It allows multiple internal servers (same process) to run on the same address and port. Basically, in #5595 it was determined that the single socket used to read the queries was a bottleneck; this alleviates that, IIUC.

plugin/reuseport/reuseport.go
```diff
@@ -0,0 +1,40 @@
...
+
+func parseNumSocks(c *caddy.Controller) error {
+	config := dnsserver.GetConfig(c)
+
+	for c.Next() {
+		args := c.RemainingArgs()
+		if len(args) != 1 {
+			return c.ArgErr()
+		}
+		numSocks, err := strconv.Atoi(args[0])
```
johnbelamaric: Would a default of "auto" set to GOMAXPROCS (or some function thereof) make sense? How does GOMAXPROCS get set in a containerized environment, is it container-aware? It must be...??

SuperQ: GOMAXPROCS is not container aware.

We would need to add something like https://pkg.go.dev/go.uber.org/automaxprocs for this.

Shmillerov:
Let me tell you about our scenario.
We want to run СoreDNS as a DaemonSet in Kubernetes. This means that there is only one pod per node. We can't scale CoreDNS horizontally in that case. And vertically too, because no matter how many resources you add, the CoreDNS will not perform better (https://github.com/coredns/coredns/issues/5595)

This fix allows to scale CoreDNS vertically.

> This is more or less doable currently by running multiple instances of coredns.

In general, I agree. But it is more difficult to control these instances.

If we run several processes in a container, it is not clear how to detect main process. If the instance goes down, who will restart it? We would like to have one process so that Kubernetes itself controls the lifecycle of our container.

plugin/reuseport/reuseport.go
Shmillerov: Hello @johnbelamaric.
My understanding is right that your idea is that you want the plugin to adjust to the available CPU and set num_socks itself?

According to my and @iyashu measurements, one replica of the CoreDNS consumes about 3 CPUs. In this case, we can set num_sock as cpu_limit/3. But in our measurements, we used the simplest logic with the file plugin and cache. There may be different plugins that will consume different CPU.

Also, I noticed that the performance does not increase with num_sock > 8. On the contrary, it begins to decrease gradually.
Therefore, the automatic setting of NUM_SOCK = GOMAXPROCS may not be the best idea.

It seems that the value should be set only after measurements with specific plugins and on a specific environment.

SuperQ:
Sorry, I clearly didn't read the code / proposal well enough. Setting up multiple listeners from the same process is the goal. 

I think this is fine for now, at least until Go has better support for this.

chrisohaver:
Perhaps we can consider an alternate plugin name? “reuseport” is a bit confusing given that coredns already binds using the reuseport option. Perhaps “multiprocess” or “multiserver” or something.

plugin/reuseport/reuseport.go
SuperQ: > Also, I noticed that the performance does not increase with num_sock > 8. On the contrary, it begins to decrease gradually.

Slightly wild guessing. This may be related to the number of network interface queues. It's common on some network interfaces to have 8 CPUs assigned to interrupt handling.

johnbelamaric: My point is we need a sensible default for the number of sockets. That could be determined as a function of the CPUs, or it could be a fixed number. But I don't want users to have to come up with a number in most cases. Ideally it should set "basically right" wherever it lands.

johnbelamaric:
> Perhaps we can consider an alternate plugin name? “reuseport” is a bit confusing given that coredns already binds using the reuseport option. Perhaps “multiprocess” or “multiserver” or something. 

"numsockets" would most clearly describe what it does, I think. It's not about multiple processes.

glebkin:
Hi!

Please take into account, that we also need additional changes in Caddy server - https://github.com/coredns/caddy/pull/6 to address this issue - https://github.com/coredns/coredns/issues/5595#issuecomment-2329195223 with performance degradation after server reload.

SuperQ:
Maybe this should be a command line flag and not something reloadable?

Shmillerov:
> Maybe this should be a command line flag and not something reloadable?

During reloading, the server restarts. When restarting, all servers start listening to a single socket because here: https://github.com/coredns/caddy/blob/master/caddy.go#L221, we have a map of server addresses to PacketConn.

With numsockets, we are starting multiple servers on the same address. Therefore, reloading is a a problem for us.

If we want this functionality with numsockets, we need to fix this somehow. It doesn't matter whether it is a plugin or a CLI flag.

glebkin:
Can anyone please take a look at this PR - https://github.com/coredns/caddy/pull/6?

plugin/reuseport/reuseport.go
BenTheElder: automaxprocs may be a desirable change anyhow, but that seems like something we should split out to another PR?

Shmillerov: Before making any "auto" setting, I would take measurements on different hardware and OS to figure out how to calculate this automatic value. It seems to me that these measurements can give different results, and it will be difficult to decide how to calculate "basically right". Windows doesn't support SO_REUSEPORT, MacOS supports, but it [doesn't work the way it does on Linux](https://github.com/uNetworking/uWebSockets/issues/1194#issuecomment-777044197).

Different hardware, different OS, different plugins that can consume different CPUs... It seems so far that it would be more reliable to use some static value.

For most people, this plugin is not needed. CoreDNS perfectly scales horizontally. But if someone want to use this plugin, it means that he have some performance issues and wants to scale vertically and he will be able to compare the results before and choose the appropriate value for him.

I can describe in the documentation that this setting should be determined individually for each case and it is better to take measurements on your environment before setting it.

BenTheElder: (Er, I meant specifically making GOMAXPROCS default cgroups aware, seperate from auto-computing a value for maxsockets, I'm not sure regarding the latter)

> For most people, this plugin is not needed. CoreDNS perfectly scales horizontally. But if someone want to use this plugin, it means that he have some performance issues and wants to scale vertically and he will be able to compare the results before and choose the appropriate value for him.

I think a lot of users will reach to enable this plugin if they find performance issues without necessarily being familiar with stress testing and optimizing the configuration, we could probably detect an auto max procs without changing the GOMAXPROCS default yet.

What's the downside to having an attempt at defaulting here when the plugin is enabled? That seems strictly better than forcing the user to pick a number. Advanced users will of course find an optimal value.

Shmillerov: I did measurements on my environment. After numsockets 8, the performance started to decrease and I don't know the reason. Perhaps this number may be different on other environments. I wouldn't like to make a mistake here by setting it incorrectly. So I didn't want to take responsibility for setting this value :)

But in general I understand you too. You would like everything to work out of the box, without any additional settings.

I'll try to take measurements with numsockets 8 (as the best value I got) with different numbers of cores. If numsockets reduces performance with a small number of CPUs, I'll consider the option of dynamic configuration depending on the CPU limit/GOMAXPROCS.

Shmillerov: Hello!
I did performance testing on 96 CPU Intel(R) Xeon(R) Gold 6248R CPU @ 3.00GHz.
CoreDNS configured with a plugin that simply takes a value from a map and respond with it.
All tests were run on k8s with different CPU limits on the container.
I took `numsockets 8` as maximum because with this number I got the best QPS on the same env without any limits.

Results: [numsockets test.md](https://github.com/user-attachments/files/17244241/numsockets.test.md)

Table:

|          | numsockets 1 | numsockets 2 | numsockets 4 | numsockets 8 |
|----------|--------------|--------------|--------------|--------------|
| limit 2  | 12381 QPS    | 11998 QPS    | 10498 QPS    | 8045 QPS     |
| limit 4  | 24920 QPS    | 24435 QPS    | 21597 QPS    | 16483 QPS    |
| limit 8  | 27855 QPS    | 50814 QPS    | 45068 QPS    | 34422 QPS    |
| limit 16 | 27386 QPS    | 50962 QPS    | 93010 QPS    | 83023 QPS    |
| limit 32 | the same     | the same     | 96244 QPS    | 234660 QPS   |

As you can see, performance degrades as `numsockets` increases under the same limits until it hits the maximum QPS. In my case, when I have just a simple logic to take a value from map and return it.
So based on this test, optimal settings are
4 CPU = numsockets 1
8 CPU = numsockets 2
16 CPU = numsockets 4
32 CPU = numsockets 8

But I did another test with a plugin that consumes more CPU.
Here is the result of testing with forward plugin and different limits:
|          | numsockets 1 | numsockets 2 | numsockets 4 | numsockets 8 |
|----------|--------------|--------------|--------------|--------------|
| limit 8  | 17248 QPS    | 12758 QPS    | 8591 QPS     | 7993 QPS     |
| limit 16 | 22577 QPS    | 24160 QPS    | 20164 QPS    | 18888 QPS    |
| limit 32 | 22666 QPS    | 28986 QPS    | 36919 QPS    | 38241 QPS    |

based on this test, there will be other optimal values.

And I didn't find a way to define "basically right" here. As I said, it depends on the plugin set. We can set it to 2 by default. But if the user decides to use this plugin "just in case", he may get performance degradation.

Based on the results of these measurements, I wouldn't set any default values ​​for the numsockets.
So I'm waiting for your decision on what to do.

johnbelamaric: Ok. What I would do then is no "auto" setting, no default; we will just make the user pick a number. But in the documentation say something like "the correct number will vary based on your environment and configured plugins, but a useful number to try is CPUs / 4 with a maximum of 8". Basically, let's at least give them a hint as to what to try. Maybe you can do a blog with your testing and add a link, too.

Shmillerov: Please check how it looks now :)

https://github.com/coredns/coredns/blob/57ff6e49f8cb16a21026b88eadbea618d04b47ce/plugin/numsockets/README.md#recommendations

plugin/numsockets/README.md
```diff
@@ -0,0 +1,60 @@
...
+option allows to open multiple listening sockets at the same address and port. In this case, kernel distributes incoming 
+connections between sockets.
+
+Enabling this option allows to start multiple servers, which increases the throughput of CoreDNS in environments with a 
+large number of CPU cores.
+
+## Syntax
+
+~~~
+numsockets [NUM_SOCKETS]
```
johnbelamaric: The brackets `[ ]` means it's optional - but it's not. So, remove the brackets.

johnbelamaric:
So, right now all the tests run without this set. Therefore, we don't have any real test coverage. Could we add some runs of some of the basic server tests with this value set? Or the integration tests? I want something that will show that basic functionality continues working with this setting.

Shmillerov:
Yes, I agree. I didn't notice that there are more tests besides plugin tests :)
I added a test that starts a server with different numsockets and checks at how many servers actually start and what port they are on.
Could you check if this is enough?

johnbelamaric:
That's a good start, but I do want to run through some actual functional tests with the value at something other than "1", and make sure there are no functional issues.

Plus the restart test you will need as you mentioned in https://github.com/coredns/caddy/pull/6#issuecomment-2402245401.

So, if you add all those tests here, they should all work except the restart test. Once we merge the Caddy fix, and update the `go.mod` to point to that version, we should see the restart test start working.

I'd like those tests in place before merging either PR. Thanks!

Shmillerov:
Hello, I added a `restart` and `whoami` tests with numsockets plugin
https://github.com/coredns/coredns/blob/42896b8c21923ed4abcfea5e66d51c55cf8484bb/test/numsockets_test.go
Do you have any other ideas what can be added? Or do you think this is enough? :)

johnbelamaric:
> Or do you think this is enough? :)

`whoami` shows that the reading of the requests and the replies work properly still, and of course the restart test is critical. So I think that's probably good.

johnbelamaric:
Shouldn't there be at least one failing test, since we haven't merged https://github.com/coredns/caddy/pull/6 ? I had expected to see that failing, then merge the caddy change, then update the go.mod to point to the new caddy version, and see the test clear up. The fact that it's not failing now makes me think our testing is incomplete?

Shmillerov:
https://github.com/coredns/coredns/pull/6882/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R146
In this PR I did replace like
```
// Temp replace. Will be removed from this PR when the caddy changes are merged
replace github.com/coredns/caddy v1.1.1 => github.com/glebkin/caddy v0.0.0-20240924070517-c7def4b62439
```
So when caddy changes will be merged, I will remove it from here and will update the caddy version

plugin/reuseport/reuseport.go
Shmillerov: Hello!
I redid the same test on the same env, with the same CPU limits, but used GOMAXPROCS additionally with value as a CPU limit. 
Previously, the GOMAXPROCS was always 96 (as a number of cores). Because of this, the results were inaccurate.

**Environment** 
96 CPU
Intel(R) Xeon(R) Gold 6248R CPU @ 3.00GHz
NUMA nodes: 2

**DNSPerf**
`taskset -c 90-95 dnsperf -d internal.txt -s 2a02:5501:28:1::f89f -p 53 -c 100 -l 30 -T 4`

**CoreDNS**
CoreDNS was running as a container in kubernetes.
Config looks like:
```
numsockets 10
reload 2s
errors
log . {
    class denial error
}
pprof :{$PPROF_PORT}
prometheus :{$METRICS_PORT}
health :{$LIVE_PORT}
ready :{$READY_PORT}

my_cache # respond with record from cache

forward . address1 address2 # forward if no records found in cache
```

addresses specified in forward can achieve  ~50k RPS

**Summary**
- Need to align GOMAXPROCS with the limit. It significantly improves performance. I suggest to add https://github.com/uber-go/automaxprocs to CoreDNS. But maybe in separate PR.
- QPS for Forward doesn't reach more than 67612. This limit is because of performance of DNS servers that I used in forward. So I suggest focusing on Cache QPS/RATE, and I’m adding here Forward just for understanding.
- RATE for Cache remains at the same level at different limits and numsockets (excluding 1 CPU). It means now there is no degradation due to increasing numsockets/limits.
- Even with numsockets > 8, performance continues to improve. So I want to say that we can use numsockets > 8 without any problems.

**Results**

numsockets 1

limit | QPS | CPU(%) | RATE (QPS/CPU) | QPS | CPU(%) | RATE (QPS/CPU)
-- | -- | -- | -- | -- | -- | --
1 CPU | 16778 | 100 | 167.8 | 36344 | 100 | 363.4
2 CPU | 23413 | 200 | 117.1 | 40977 | 200 | 204.9
3 CPU | 31307 | 300 | 104.4 | 38563 | 273 | 141.3
4 CPU | 38948 | 393 | 99.1 | 34965 | 290 | 120.6
5 CPU | 48664 | 494 | 98.5 | 33651 | 304 | 110.7
6 CPU | 53108 | 566 | 93.8 | 32618 | 309 | 105.6
NO LIM | 23083 | 1119 | 20.6 | 28465 | 442 | 64.4

numsockets 2
limit | Forward QPS | CPU (%) | RATE (QPS/CPU) | Cache QPS | CPU (%) | RATE (QPS/CPU)
-- | -- | -- | -- | -- | -- | --
1 CPU | 16253 | 100 | 162.5 | 37484 | 100 | 374.8
2 CPU | 21489 | 200 | 107.4 | 42884 | 200 | 214.4
3 CPU | 29569 | 300 | 98.6 | 62647 | 300 | 208.8
4 CPU | 44070 | 400 | 110.2 | 84664 | 400 | 211.7
5 CPU | 47977 | 494 | 97.1 | 110568 | 493 | 224.3
6 CPU | 55026 | 590 | 93.3 | 101200 | 520 | 194.6

numsockets 3
limit | Forward QPS | CPU (%) | RATE (QPS/CPU) | Cache QPS | CPU (%) | RATE (QPS/CPU)
-- | -- | -- | -- | -- | -- | --
1 CPU | 16964 | 100 | 169.6 | 35927 | 100 | 359.3
2 CPU | 22315 | 200 | 111.6 | 44558 | 200 | 222.8
3 CPU | 34324 | 299 | 114.8 | 59068 | 299 | 197.6
4 CPU | 43660 | 397 | 110 | 84431 | 398 | 212.1
5 CPU | 47525 | 494 | 96.2 | 106006 | 493 | 215
6 CPU | 58703 | 580 | 101.2 | 137948 | 590 | 233.8
8 CPU | 55511 | 732 | 75.8 | 158704 | 690 | 230

numsockets 4
limit | Forward QPS | CPU (%) | RATE (QPS/CPU) | Cache QPS | CPU (%) | RATE (QPS/CPU)
-- | -- | -- | -- | -- | -- | --
4 CPU | 37779 | 395 | 95.6 | 86625 | 398 | 217.7
6 CPU | 49612 | 590 | 84.1 | 129863 | 592 | 219.4
8 CPU | 55525 | 700 | 79.3 | 176373 | 783 | 225.3
10 CPU | 60136 | 820 | 73.3 | 206945 | 870 | 237.9

numsockets 5
limit | Forward QPS | CPU (%) | RATE (QPS/CPU) | Cache QPS | CPU (%) | RATE (QPS/CPU)
-- | -- | -- | -- | -- | -- | --
4 CPU | 41217 | 395 | 104.3 | 86308 | 398 | 216.9
6 CPU | 58107 | 580 | 100.2 | 138444 | 594 | 233.1
8 CPU | 67612 | 715 | 94.6 | 184441 | 783 | 235.6
10 CPU | 63655 | 827 | 77 | 228928 | 964 | 237.5
12 CPU | 61124 | 935 | 65.4 | 257965 | 1060 | 243.4

numsockets 8
limit | Forward QPS | CPU (%) | RATE (QPS/CPU) | Cache QPS | CPU (%) | RATE (QPS/CPU)
-- | -- | -- | -- | -- | -- | --
4 CPU | 41026 | 398 | 103.1 | 87958 | 399 | 220.4
6 CPU | 48723 | 580 | 84 | 138505 | 595 | 232.8
8 CPU | 58256 | 700 | 83.2 | 190950 | 785 | 243.2
10 CPU | 60110 | 870 | 69.1 | 242212 | 970 | 249.7
12 CPU | 58728 | 970 | 60.5 | 252645 | 1125 | 224.6
14 CPU | 60963 | 1100 | 55.4 | 271986 | 1260 | 215.9

numsockets 10
limit | Forward QPS | CPU (%) | RATE (QPS/CPU) | Cache QPS | CPU (%) | RATE (QPS/CPU)
-- | -- | -- | -- | -- | -- | --
12 CPU | 55799 | 920 | 60.7 | 260453 | 1120 | 232.5
14 CPU | 63630 | 1060 | 60 | 332332 | 1308 | 254.1

Shmillerov: Hello @johnbelamaric

numsockets plugin works really good, there is no performance degradation when I set more numsockets with the same limits. And RATE (for cache) is always 200+ for different limits and numsockets.

Based on these tests, I removed the restrictions from the documentation and set default value `2` for numsockets plugin.
Why 2? Because it will significantly increase throughput (as we can see in cache tests, 40977 QPS/2 CPU -> 110568 QPS/5 CPU).
But if this is not enough, it can be specified manually in the plugin configuration based on the target set of plugins and environment. Instructions for calculating numsockets are still in the documentation.

Let's decide what to do with this PR. What should be done to merge it to master?

johnbelamaric: This is awesome, thanks. I merged the coredns/caddy PR, so please update this to use that, and then I can approve it.

johnbelamaric:
@Shmillerov if you happen to be at KubeCon NA in Salt Lake City next month, would be great if you want to talk about this in the CoreDNS maintainer track session. Otherwise I will cover it. Great stuff.

Shmillerov:
@johnbelamaric, thank you, but I can't participate :) I believe you will cover me!

johnbelamaric (approved):

johnbelamaric:
@Shmillerov just FYI since I don't think you are on the CoreDNS [Slack](https://cloud-native.slack.com/archives/C4DF7FP71). We are hoping to do the following in the next few days:
- merge low risk PRs and cut a patch release
- merge this (and maybe others?) and cut a minor release

johnbelamaric (approved):
Do not merge until we cut a patch release with low-risk changes first.

PR #6898, opened 2024-09-30
only create PTR records for endpoints with hostname defined

chrisohaver:

Per discussion in https://kubernetes.slack.com/archives/C09QYUH5W/p1713088975540039,

We should only be creating PTR records for Endpoints that have a hostname defined in the endpoint address object.
This should normally result in producing one PTR record for each IP, even if multiple Services select the same Pod.  K8s enforces that only one endpoints has a hostname defined at any time.  If for some reason more than one have hostname defined, we will return both rather than select one arbitrarily - but IIUC, this would only happen if something is awry in k8s.

https://kubernetes.slack.com/archives/C09QYUH5W/p1713088975540039

### DIFF
--- a/plugin/kubernetes/reverse.go
+++ b/plugin/kubernetes/reverse.go
@@ -44,8 +44,13 @@ func (k *Kubernetes) serviceRecordForIP(ip, name string) []msg.Service {
 		}
 		for _, eps := range ep.Subsets {
 			for _, addr := range eps.Addresses {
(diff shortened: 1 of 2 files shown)


Michael-Sinz:
This turns out to break our load balancer since now there are no DNS entries for the individual pods.

Strange that one would be upset at multiple entries being returned when multiple services are being defined for the pod.  I would have expected multiple in that case.

PR #6914, opened 2024-10-07
build(deps):  deps/zap 1.17.0 to 1.27.0

gorsing:
This pull request is needed to update the dependency `go.uber.org/zap` from version 1.17.0 to 1.27.0. Version 1.17 was released in 2021 and contained some known memory leak issues. The newer versions include fixes for these problems, which will help improve the performance and stability of the application.

This pull request is related to resolved memory leak issues in newer versions of `go.uber.org/zap` that were identified in version 1.17.0.

No documentation changes are required. Current usage examples for `zap` will remain relevant, as the changes in version do not introduce significant alterations to the API.

This pull request does not introduce a backward incompatible change. The dependency update maintains backward compatibility with existing functionality, and all existing interfaces remain unchanged.


chrisohaver:
This looks like an indirect dependency, so I _think_ the proper way to update this is in the package that requires it directly, `go.etcd.io/etcd/client/v3` ...

```
go mod why go.uber.org/zap
# go.uber.org/zap
github.com/coredns/coredns/plugin/etcd
go.etcd.io/etcd/client/v3
go.uber.org/zap
```

chrisohaver (changes requested):
IMO, this needs to be done upstream instead of here.

Issue #6645, opened 2024-04-28
etcd connection leakage when reload


Yukielz:
**What happened**:
etcd connection leakage when reload.
**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
use etcd plugin and modify corefile to trigger reload.
**Anything else we need to know?**:

**Environment**:

- the version of CoreDNS: v1.11.1
- Corefile:
```yaml
  .:53 {
      errors
      health {
          lameduck 5s
      }
      ready
      etcd com {
          path /skydns/ka2
          endpoint http://192.168.0.1:2379 http://192.168.0.2:2379 http://192.168.0.3:2379
          fallthrough
      }
      cache 10
      loop
      reload
      log
      loadbalance
  }
```
- logs, if applicable:
- OS (e.g: `cat /etc/os-release`):
- Others:

yangsenzk:
This bug can be reproduced. The **process_open_fds** metric of backend ETCD will be continuously increased by modify corefile that triggered coredns reload.

Yukielz:
add more infomation:
when i add `log` to tigger coredns reload, it was obvious that the old connection was not released
![image](https://github.com/coredns/coredns/assets/56835284/74042e3d-bd29-4251-b5a7-a74f3fd12289)

Yukielz:
@chrisohaver cc

Yukielz:
@chrisohaver  Hi, chrisohaver. can you check this issue and review this [pr](https://github.com/coredns/coredns/pull/6646).
To avoid this, we must restart all coerdns pod when etcd endpoints changed.

Yukielz:
#6646

PR #6916, opened 2024-10-08
fix: reload close channel

gorsing:
This pull request introduces improvements to the reload functionality of CoreDNS. It refines the logic for periodically checking for changes in the Corefile configuration and enhances the shutdown procedure for the reload goroutine. Specifically, it:

1. Implements a more reliable quit channel mechanism using `chan struct{}` to signal the reload routine to stop.
2. Ensures that the interval and jitter settings are adjustable and properly logged, enhancing the usability of the reload plugin.
3. Prevents race conditions and ensures synchronized access to shared variables between setup and the `onInstanceStartup` hook.

This PR addresses the need for a more robust and synchronized reload mechanism in CoreDNS, improving overall stability.
The problem remains #6646 #6645

No documentation changes are needed.

No, this pull request does not introduce any backward incompatible changes or deprecations. It enhances existing functionality while maintaining compatibility with current configurations.


chrisohaver:
I cant put my finger on the significance of these changes.  Palpably, what problem do these changes resolve?

gorsing:
Problem with the problem code https://github.com/coredns/coredns/pull/6646 https://github.com/coredns/coredns/issues/6645
The code opens Goroutines but does not close it. and this in combination with the reload and etcd plugins causes a memory leak

chrisohaver:
> Problem with the problem code #6646 #6645 The code opens Goroutines but does not close it. and this in combination with the reload and etcd plugins causes a memory leak

I don't see how this code changes the behavior.  It looks like go routines were closed before, and they are still closed after these changes.  Can you show how in the previous code that the go routines were not closed?

thevilledev (changes requested):
This PR is very much stale and also in conflict with changes like #7562. I think the issue requires demonstration through a test case.

Issue #6915, opened 2024-10-08
Incoordination between rewrite/cache plugins


jasonjoo2010:
Hi community,  

We deeply used rewrite plugin internally on CoreDNS for various purposes, like traffic redirection and application performance improvement, and it worked quite good and never disappointed users. But we encountered a strange issue and finally confirmed it's one thing we need to resolve in the code logic to make the rewrite/cache plugins better coordinated. After fixing it internally, I want to post the story here and ask about people's opinion if it's a necessary fix for open source.  

As people all know, the order of plugins getting called is very important in different aspects of performance and plugin effect. For example, the rate limit plugin should be in the front tier to block unwanted requests. The default call order for cache/rewrite is cache first, rewrite next. Meaning CoreDNS will use more space(memory) to trade for time(CPU). We swap the order of these 2 plugins considering the appropriate amount of rewrite rules doesn't consume much CPU, meanwhile we want to reduce the cache cost in some cases:  

An internal client library uses a UUID style domain when accessing the storage backend by design. The domain is in this form: `<uuid prefix>.store.intl.x-corp.com`. This behaviour exhausts the memory cache shortly and CoreDNS begins evicting cache entries. Although the cache eviction is not a big problem and it still runs efficiently, the queries are still not cached by CoreDNS. Furthermore, we usually take the evicting rate as one of the performance relevant indicators for config tuning. So we added new rewrite rule for this UUID domain to remove the UUID part from it to accelerate the resolution. But without moving the cache plugin behind the rewrite plugin, it won't be cached as expected. So we adjusted their order for this purpose and deployed to all production.  

All things are good until we encountered a strange issue: the domain name in the answer sometimes doesn't match the name in the query. It indicates the name reversion is not working although the rewrite rule works under certain conditions. After inspecting this issue, we get the reason as follows.  

- There is a rewrite rule of suffix type rewriting `.com.internal.domain` to `.com`.
- Client A queries `TEST1.EXAMPLE.COM` and get an answer for `TEST1.EXAMPLE.COM` which is good.
- Client B queries `test1.example.com` and get an answer for `TEST1.EXAMPLE.COM` which is a bit strange but still good. Because the domain resolution is case insensitive in the design.
- Client C queries `test1.example.com.internal.domain` and get an answer for `TEST1.EXAMPLE.COM` which is not good. The name in the answer doesn't match the name in the query so the client won't accept it.

There are basically 2-3 ways to fix it:  

- Persist the domain name in the cache in lower case only.
[...]

PR #6917, opened 2024-10-08
fix: make name reverter case insensitive

jasonjoo2010:
This MR fixes the issue #6915 

When users adjusted the plugin call order and put rewrite before cache, CoreDNS may misbehave with mismatched domain in the answer section in the response.

#6915 

No document change.

It's backward compatible.

### DIFF
--- a/plugin/rewrite/name.go
+++ b/plugin/rewrite/name.go
@@ -59,13 +59,20 @@ func newRemapStringRewriter(orig, replacement string) stringRewriter {
 }
 
 func (r *remapStringRewriter) rewriteString(src string) string {
-	if src == r.orig {
-		return r.replacement
+	switch {
+	case len(r.orig) > len(src):
+		return src
+	case len(r.orig) == len(src):
+		if strings.EqualFold(src, r.orig) {
+			return r.replacement
+		}
+		return src
+	default:
+		if hasSuffixFold(src, "."+r.orig) {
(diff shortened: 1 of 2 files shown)


greenpau:
@jasonjoo2010 , please improve the coverage.

Also, it makes sense to explain the logic in the documentation, thoughts?

jasonjoo2010:
Hi @greenpau 
Thanks for your quick response and appreciate.

I have improved the test coverage and it should be good now.
Coming to the document, we can discuss a bit more.

According to [RFC4343](https://datatracker.ietf.org/doc/html/rfc4343), it explained in all aspects about the case insensitivity in DNS protocol and this behaviour should be expected in all systems as default. It only becomes a surprise if a service breaks that.
Coming to the case here, for `auto` there's nothing the users can do about the case and it's handled by CoreDNS internally with best effort. So no matter which case used in a query, the server should reply the same one or it's an unexpected behaviour. Not like explicitly rewrite the answer using regular expression which is also supported, the user has the responsibility to give a "proper" expression(which we don't actually use in our cases and it's also not intuitive from the user side tbh).

This PR fixes the `auto` support in 2 modes: regexp/suffix.

Another thing which worths a mention is that the behaviour of CoreDNS and other DNS servers is a bit different in the case dealing, other DNS like Windows DNS regards the domain name in the query but CoreDNS regards the domain name in the cache. It's not a big problem because the RFC said it's OK.

Wdyt?

Issue #6937, opened 2024-10-25
DNS seems flaky


matteriben:
### What happened?

I created a headless service without a selector and defined an endpointslice using podIPs. Inside the pods I queried DNS. It seemed to take ~20 seconds for the first successful response. Following this a subsequent response apparently failed unexpectedly (line 9 of the massaged core dns log file snippet).

```
     1	2024-10-25T00:53:46.182233945Z [INFO] 10.244.0.82:44628 - 30082 "A IN node0.pod.namespace.svc.cluster.local. udp 72 false 512" NXDOMAIN qr,aa,rd 165 0.000117891s
     2	2024-10-25T00:53:46.253100368Z [INFO] 10.244.0.93:52617 - 56694 "A IN node0.pod.namespace.svc.cluster.local. udp 72 false 512" NXDOMAIN qr,aa,rd 165 0.000105911s
     3	2024-10-25T00:53:46.480963630Z [INFO] 10.244.0.92:52986 - 22363 "A IN node0.pod.namespace.svc.cluster.local. udp 72 false 512" NXDOMAIN qr,aa,rd 165 0.000093061s
     4	2024-10-25T00:53:47.204138260Z [INFO] 10.244.0.82:51733 - 7928 "A IN node0.pod.namespace.svc.cluster.local. udp 72 false 512" NXDOMAIN qr,aa,rd 165 0.00015467s
     5	2024-10-25T00:53:47.272560170Z [INFO] 10.244.0.93:53722 - 42551 "A IN node0.pod.namespace.svc.cluster.local. udp 72 false 512" NOERROR qr,aa,rd 142 0.00012205s
     6	2024-10-25T00:53:47.273148393Z [INFO] 10.244.0.93:45857 - 16439 "AAAA IN node0.pod.namespace.svc.cluster.local. udp 72 false 512" NOERROR qr,aa,rd 165 0.000229791s
     7	2024-10-25T00:53:47.500587592Z [INFO] 10.244.0.92:52084 - 56243 "A IN node0.pod.namespace.svc.cluster.local. udp 72 false 512" NOERROR qr,aa,rd 142 0.000108021s
     8	2024-10-25T00:53:47.500996354Z [INFO] 10.244.0.92:34111 - 60259 "AAAA IN node0.pod.namespace.svc.cluster.local. udp 72 false 512" NOERROR qr,aa,rd 165 0.000133821s
[...]

PR #6939, opened 2024-10-26
fix: order kubernetes plugin before cache plugin

matteriben:
I'm new to this project, and this change has not been tested yet, but it is intended to address issues I've seen in kubernetes which seem involve caching of the zone which kubernetes is authoritative for.

https://github.com/coredns/coredns/issues/6937

I have observed delays ~30 seconds to resolve names for slow starting pods (taking more than 30 seconds for IP assignment), also sometimes the old address initially being returned for ~30 seconds with fast starting pods, and an occasional unexpected response (such as NXDOMAIN following NOERROR) around the time resolution seems to start working.

Not sure.

### DIFF
--- a/plugin.cfg
+++ b/plugin.cfg
@@ -45,7 +45,6 @@ any:any
 chaos:chaos
 loadbalance:loadbalance
 tsig:tsig
-cache:cache
 rewrite:rewrite
 header:header
 dnssec:dnssec
@@ -61,8 +60,9 @@ k8s_external:k8s_external
 kubernetes:kubernetes
 file:file
 auto:auto
-secondary:secondary
 etcd:etcd
+cache:cache
+secondary:secondary
 loop:loop
 forward:forward
 grpc:grpc


chrisohaver:
If your goal is to exclude the Kubernetes cluster domain from the cache, you can do that with the `disable` option in cache plugin.

matteriben:
> If your goal is to exclude the Kubernetes cluster domain from the cache, you can do that with the `disable` option in cache plugin.

My goal was to address the root cause of an issue I'm seeing involving the cache for a zone the kubernetes plugin is authoritative for. You are right that in my specific case I can disable cache for zones which the kubernetes plugin is authoritative for and do not fallthrough.

I'm new to this project, but it seems to me that the root cause is the plugin order with cache before an authoritative plugin. To me it seems the cache should only be used after a fallthrough from plugins handling authoritative zones. Disabling zones which other plugins are authoritative for seems to me like a workaround to the plugin ordering which seems like the root cause to me.

I would expect in general that authoritative zones and non-authoritative zones would have very different requirements in terms of caching. I would expect plugins which are authoritative to implement their own "caching" such that they can provide responses which are authoritative and fast without relying on a general purpose cache plugin. Perhaps my expectations are completely wrong here.

chrisohaver:
I don’t think changing plugin order is the right solution.  Disabling cache for the domain is more flexible, backward compatible, and since it’s already available, it is not prone to introducing regressions as would changing the plugin order.

johnbelamaric:
> I have observed delays ~30 seconds to resolve names for slow starting pods (taking more than 30 seconds for IP assignment), also sometimes the old address initially being returned for ~30 seconds with fast starting pods, and an occasional unexpected response (such as NXDOMAIN following NOERROR) around the time resolution seems to start working.

This may be due to latency in the K8s endpoints controller rather than CoreDNS. For example, a slow starting pod will not be added to the service until it shows `Ready`. If it's not in the service, it won't show up in CoreDNS. I suggest you check if that could be the cause of the issue before looking at CoreDNS.

Note that cache only caches things for plugins that come AFTER it. We want to cache things even if we are authoritative, in general.

chrisohaver:
> We want to cache things even if we are authoritative, in general.

In that case, we may want to revert the recent change made to k8s coredns configmap that disables cache for the cluster domain. 

https://github.com/kubernetes/kubernetes/pull/128359

chrisohaver:
Although IMO we shouldn’t be caching the cluster domain since the k8s api client keeps its own cache of sorts, which is kept up to date by the service/endpoint watches. With cache caching too, the data is double cached, and with the default 5 second ttl, we could be serving outdated information from cache while we have updated info in the k8s client watch indexed store. So better not to cache it.

johnbelamaric:
Yeah, k8s plugin keeps everything in memory so cache serves no purpose

PR #6948, opened 2024-10-30
automaxprocs

Shmillerov:

Based on https://github.com/coredns/coredns/pull/6882#discussion_r1818522596 it's better to align GOMAXPROCS with CPU limit. Suggest to use [automaxprocs](https://github.com/uber-go/automaxprocs) for that

no

I think no

No


coremain/run.go
```diff
@@ -47,6 +49,11 @@ func Run() {
 	log.SetOutput(os.Stdout)
 	log.SetFlags(LogFlags)
 
+	_, err := maxprocs.Set(maxprocs.Logger(log.Printf))
+	if err != nil {
+		mustLogFatal(err)
```
SuperQ: I don't think we should make this fatal.

Shmillerov: By default, automaxprocs ignore the error in init https://github.com/uber-go/automaxprocs/blob/master/automaxprocs.go#L32.

But since the log in this case contains time, I wanted to align the logs with the current ones in CoreDNS and therefore moved maxprocs.Set to run.go.

What do you prefer? ignore, log or fatal ? :)

johnbelamaric:
Once we add this, does it makes sense with your latest testing to determine `numsockets` defaults from GOMAXPROCS? Even if it's not the *optimal* value in all situations, is it better than "2" as a default?

If so, maybe we should rename "numsockets" to "multisocket"? I think that's a little more descriptive in the Corefile.

coremain/run.go
SuperQ: I think a warning log would be sufficient.

Shmillerov: Changed to log without fatal.

I added `[WARNING]` before message to understand that this is a warning. The same as in `log` plugin.

Maybe it looks strange, but in CoreDNS we have default go logger. And I think it's not a good idea to use `log` plugin in main.

Shmillerov:
> Once we add this, does it makes sense with your latest testing to determine `numsockets` defaults from GOMAXPROCS? Even if it's not the _optimal_ value in all situations, is it better than "2" as a default?
> 
> If so, maybe we should rename "numsockets" to "multisocket"? I think that's a little more descriptive in the Corefile.

I'm ok, I will do it. But I think this change is not related to numsockets/multusocket plugin, so if in this PR everything is ok, we can merge it first

johnbelamaric (approved):

Issue #6949, opened 2024-10-30
view plugin performance


Shmillerov:
**What would you like to be added**:
CoreDNS version: v1.11.3

I have a CoreDNS server with 15 `view` configurations that filters my request by client IP like:
```
. {
    view 1 {
        expr client_ip() == '127.0.0.1'
    }

    rewrite edns0 local set 0xffee 1

    log . {
        class denial error
    }
    errors

    import fanout/fanout.conf
}

. {
    view 2 {
        expr client_ip() == '127.0.0.2'
    }

    rewrite edns0 local set 0xffee 2

    log . {
        class denial error
    }
    errors

    import fanout/fanout.conf
}

. {
    view 3 {
        expr client_ip() == '127.0.0.3'
    }

    rewrite edns0 local set 0xffee 3

    log . {
        class denial error
    }
    errors

    import fanout/fanout.conf
}

. {
    reload 2s
    prometheus :{$METRICS_PORT}
    pprof :{$PPROF_PORT}
    health :{$LIVE_PORT}
    ready :{$READY_PORT}
}
```

And based on client_ip rewrites eDNS option.

I performed test with 1 CPU and 1 GOMAXPROCS.

And here is a profile of test
[pprof.dns-proxy.samples.cpu.001.pb.gz](https://github.com/user-attachments/files/17573511/pprof.dns-proxy.samples.cpu.001.pb.gz)

![image](https://github.com/user-attachments/assets/6eb994cc-7f76-49e4-8110-bc1ae41e105f)

`view` plugin took 22.1% of CPU time. But I just need to filter request by client ip.
Also 7.4% GC worked. Most of the garbage is created by `view`.

Problem places:
https://github.com/coredns/coredns/blob/master/plugin/pkg/expression/expression.go#L13 - collecting a map of variables for each request. It took 13.2%
https://github.com/coredns/coredns/blob/master/plugin/pkg/expression/expression.go#L38 - сalculating client IP from request, which requires conversion from addr to string. It looks like we did this for every request and for every view configuration. It took 6.7%

**Why is this needed**:
So I think 25% of CPU can be released if I give up the `view` and just make a map and filter by IP. Because of performance I can't use `view` plugin. It would be great to make the OOB solution optimal so that I don't have to write a custom plugin.

Shmillerov:
I improved fanout plugin a bit and now I got 29% CPU consumed by view plugin
[...]

Issue #6900, opened 2024-10-01
Planning next release 1.11.4 or 1.12.0


chrisohaver:
Let’s identify open PRs that should be in the next release and discuss other release issues here.

chrisohaver:
In the case we have not merged any code after the last release that warrants a minor feature bump: Since we are still releasing from a single branch, perhaps we should first release backward compatible fixes as 1.11.4 then immediately follow up with a 1.12.0 release that includes backward incompatible changes and/or other fixes that warrant a minor version bump.

Xdynix:
Can I ask about the planned timeline for this release? It actually has the fix I'm looking for (#6416). Or is there anything I can do to speed it up?

chrisohaver:
All plans are here in the open.  No plans of an actual date yet.

Xdynix:
Got it, thanks for the reply.

ufenx:
Are there any releasing plans determined here? The current release 1.11.3 used [go version](https://github.com/coredns/coredns/blob/v1.11.3/.go-version) 1.21.8 and a go version with at least 1.22.7+ or 1.23.1+ is required. Do you have a plan of the actual release date before December and make sure the go version is updated to remediate the CVEs? 

Thank you so much.

chrisohaver:
> Are there any releasing plans determined here? The current release 1.11.3 used [go version](https://github.com/coredns/coredns/blob/v1.11.3/.go-version) 1.21.8 and a go version with at least 1.22.7+ or 1.23.1+ is required. Do you have a plan of the actual release date before December and make sure the go version is updated to remediate the CVEs?
> 
> Thank you so much.

This is the planning thread. We plan to release soon.

ufenx:
> > Are there any releasing plans determined here? The current release 1.11.3 used [go version](https://github.com/coredns/coredns/blob/v1.11.3/.go-version) 1.21.8 and a go version with at least 1.22.7+ or 1.23.1+ is required. Do you have a plan of the actual release date before December and make sure the go version is updated to remediate the CVEs?
> > Thank you so much.
> 
> This is the planning thread. We plan to release soon.

Thanks for your reply. But do you have a planned release date? I'm asking because if we can not get it before December, we have to make a backup plan to build a version on our own starting from now.

chrisohaver:
We should release before December.

johnbelamaric:
@chrisohaver as we discussed on Slack, let's do both :)

What needs to merge to get the next 1.11 release out?

[...]

PR #6957, opened 2024-11-07
[metadata] Add ability to define metadata based on request

Shmillerov:
I have a scenario - need to map client IP to some ID and write this ID as edns option. I used view + rewrite for this. But it works slowly and configuration looks strange, because for 50 IPs I need to create 50 server blocks.

Therefore, I improved the metadata plugin, which now allows to set metadata based on DNS request characteristic (using go template) and pre-defined in config mapping

Example of config:
```
metadata {
    map {{.Type}} key1     key2 {
        A         val11    val12
        AAAA      val21    val22
        default   val31    val32
    }
}
rewrite edns0 local set 0xffee {metadata/key1}
```

it looks like switch case by Request.Type. When type is matches A, it adds `metadata/key1 = val11` and `metadata/key2 = val12` to metadata. And as usual we can use it in other plugins, like rewrite edns0 as in this example

Updated metadata docs: https://github.com/Shmillerov/coredns/blob/feature/metadata-map/plugin/metadata/README.md

https://github.com/coredns/coredns/issues/6949

Need to update coredns.io after the merge

no


johnbelamaric:
Does this include all the functionality of https://github.com/coredns/metadata_edns0 ?

johnbelamaric:
> Does this include all the functionality of https://github.com/coredns/metadata_edns0 ?

Oh, I misunderstood. This modifies the edns data in the request?

I don't think I want that done in the `metadata` plugin. So far, the metadata plugin is just a way to tell CoreDNS "enable metadata functionality and expose stuff to other plugins in the chain". It doesn't touch the actual request or response; that is left to other plugins.

`metadata_edns0` is similar, it is about *decoding* edns0 and mapping it into the metadata.

What you have here is *writing* edns0 data. It would seem more appropriate as part of `rewrite` in that case. Or as a new plugin for appending edns0 data (that tickles a memory, are we sure @rajansandeep didn't write something like that a few years ago?).

Shmillerov:
Hello, sorry. Looks like I made a bad description :) It's not related to edns0 at all.

I made a plugin that allows to set metadata based on request characteristic and specified mapping.
But as we discussed with @chrisohaver in slack, it can be a part of metadata plugin, so I did it here

Let me provide you an example of config
```
metadata {
    map {{.Type}} key1     key2 {
        A         val11    val12
        AAAA      val21    val22
        default   val31    val32
    }
}
rewrite edns0 local set 0xffee {metadata/key1}
```

it looks like switch case by Request.Type. When type is matches A, it adds `metadata/key1 = val11` and `metadata/key2 = val12` to metadata. And as usual we can use it in other plugins, like rewrite edns0 as in this example

chrisohaver:
Its not clear what fields/parts of the request are available in the templating... Would for example the extra records be available? i.e. OPT/edns0?

Shmillerov:
> Its not clear what fields/parts of the request are available in the templating... Would for example the extra records be available? i.e. OPT/edns0?

It is possible to use any public methods and fields of Request structure https://pkg.go.dev/github.com/coredns/coredns/request#Request in go template. So you can reach edns0 option using smth like `{{.Req.IsEdns0.Option}}`

```
func Test(t *testing.T) {
	tmpl := template.Must(template.New("{{.Req.IsEdns0.Option}}").Parse("{{.Req.IsEdns0.Option}}"))

	msg := dns.Msg{}
	msg.SetEdns0(4096, true)
	msg.IsEdns0().Option = append(msg.IsEdns0().Option, &dns.EDNS0_LOCAL{
		Code: 123,
		Data: []byte("test"),
	})
	r := request.Request{
		Req:  &msg,
		W:    nil,
		Zone: "",
	}

	tmpl.Execute(os.Stdout, &r) // prints [123:0x74657374]
}
```

Maybe we can make more usable custom functions for accessing edns0 options. But it's possible as for now

johnbelamaric:
Ok, I am not sure about this yet and won't have the bandwidth to think about it until after KubeCon. Can we focus on getting the low risk changes out and then merging the multisocket stuff and doing another release? I would like to be able to talk about it as "done" next week rather than "coming soon", if possible. ie, https://github.com/coredns/coredns/issues/6900

Shmillerov:
@johnbelamaric, @chrisohaver, did you get a chance to review this?

Shmillerov:
Hello. Do you have a plan to check this PR? Do you agree to take this into an internal plugin or should I make an external one with similar functionality?

chrisohaver:
> Hello. Do you have a plan to check this PR? Do you agree to take this into an internal plugin or should I make an external one with similar functionality?

I'm sorry for taking so long to get back to this.  When we originally discussed this feature in slack several months ago, I had the wrong idea of what the goal was.  I had a simpler function in mind.

The syntax is a bit complex, but I think the function could be of general use?  One syntactical issue is that we have not yet (that I know of) introduced nested `{}` blocks inside plugin definitions.  This might require updates to external corefile parsing/migration code. 

Functionally, the feature feels like a switch-case ... e.g. 

```
switch TYPE
  case "A": foo="123"
  case "AAAA": bar="abc"
  default: baz="xyz"
```

To better fit coredns corefile style, perhaps this should be a series of  separate statements instead of a block. e.g.

```
metadata {
    if {{.Type}}=='A' key1=val11 key2=val12
    elseif {{.Type}}=='AAAA' key1=val21    key2=val22
    else   key1=val31    key2=val32
}
```

This ends up being more verbose, and harder to parse.  It is more flexible but is this over tooling something that would get little use?  It seems like there is a general use case here.  Maybe we can think of some other use cases.

Shmillerov:
Thanks for your reply!

> The syntax is a bit complex, but I think the function could be of general use?

As far as I understand, Corefile uses Caddyfile syntax. I found https://caddyserver.com/docs/caddyfile/directives/map and implemented it the same way.

> One syntactical issue is that we have not yet (that I know of) introduced nested {} blocks inside plugin definitions. This might require updates to external corefile parsing/migration code.

I don't remember, I did this 4-5 months ago, but from unit tests it looks like everything works fine for my PR without any updates to external corefile parsing/migration code.

I originally did it as a separate `metamap` plugin and there was no nesting. If you don't like the nesting, we can go back to this option

> To better fit coredns corefile style, perhaps this should be a series of separate statements instead of a block. e.g.

About `if, elseif, else`. I think there can be several types of configuration: static and dynamic.
Dynamic configuration is now supported by the reload plugin. There are 2 options:
1. edit the whole Corefile
2. move the dynamic part to a file/files, import them and edit these files.

It seems that method 2 is easier for machine configuration, and for my option it would look like this:
```
    metadata {
        map {{.IP}} vrf {
            import ./mapping/*
            default REFUSED
        }
    }
```

where `mapping` can be one or more files in any order. But for `if` and `elseif` the order is important. Also "config writer" should think that the first line should be written with `if`, the second with "elseif". I don't think it's a good idea.

> Maybe we can think of some other use cases.

Let's think about whether this functionality is needed at all? I use it, but maybe I have some specific needs. Maybe it will be easier for users to make their own plugin than to configure these conditions with pseudocode.

chrisohaver:
> I don't remember, I did this 4-5 months ago, but from unit tests it looks like everything works fine for my PR without any updates to external corefile parsing/migration code.

I was thinking of the coredns/corefile-migration project.  I think it may require updates to support nested plugin options.

> Let's think about whether this functionality is needed at all? I use it, but maybe I have some specific needs. Maybe it will be easier for users to make their own plugin than to configure these conditions with pseudocode.

Unless we can identify some other compelling use cases, I think the map functionality as-is not generally utilitarian enough to justify adding it to coredns.

PR #6967, opened 2024-11-11
Bump version to 1.11.4 and add notes

chrisohaver:
See any other fixes/features we should highlight?  Please amend the notes.


johnbelamaric:
LGTM, thanks @chrisohaver. I think maybe we should include the distroless update in #6968 - that seems a reasonable thing to do in the "less risky" release before 1.12.0.

PR #6975, opened 2024-11-15
Do not interrupt querying readiness probes for plugins

glebkin:
Kubernetes constantly polls the readiness of the applications running in cluster. Getting the OK status does not guarantee that the application will continue to be ready to serve traffic. For example, our DNS server can use an external application as a source for DNS records. At some point, it may stop responding, we might want to signal this to k8s and stop receiving traffic until the connection is restored.

No issues

I've already updated documentation for `ready` plugin

It's backwards compatible as I introduced new functionality via additional configuration param.

### DIFF
--- a/plugin/ready/README.md
+++ b/plugin/ready/README.md
@@ -19,7 +19,7 @@ their readiness reported as the union of their respective readinesses.
 
 ~~~
 ready [ADDRESS] {
-    monitor [until-ready | continuously]
+    monitor until-ready|continuously
 }
 ~~~
 
--- a/plugin/ready/README.md
+++ b/plugin/ready/README.md
@@ -8,8 +8,7 @@
 
 By enabling *ready* an HTTP endpoint on port 8181 will return 200 OK, when all plugins that are able
 to signal readiness have done so. If some are not ready yet the endpoint will return a 503 with the
-body containing the list of plugins that are not ready. Once a plugin has signaled it is ready it
-will not be queried again.
+body containing the list of plugins that are not ready.
 
 Each Server Block that enables the *ready* plugin will have the plugins *in that server block*
 report readiness into the /ready endpoint that runs on the same port. This also means that the
--- a/plugin/ready/README.md
+++ b/plugin/ready/README.md
@@ -18,12 +18,19 @@ their readiness reported as the union of their respective readinesses.
 ## Syntax
 
 ~~~
-ready [ADDRESS]
+ready [ADDRESS] {
+    monitor [until-ready | continuously]
+}
 ~~~
 
 *ready* optionally takes an address; the default is `:8181`. The path is fixed to `/ready`. The
 readiness endpoint returns a 200 response code and the word "OK" when this server is ready. It
 returns a 503 otherwise *and* the list of plugins that are not ready.
+
+The *ready* directive can include an optional `monitor` parameter, defaulting to `until-ready`. The following values are supported:
+
+* `until-ready` - once a plugin signals it is ready, it will not be checked again. This mode assumes stability after the initial readiness confirmation.
+* `continuously` - in this mode, plugins are continuously monitored for readiness. This means a plugin may transition between ready and not ready states, providing real-time status updates.
 
 ## Plugins
 
--- a/plugin/ready/README.md
+++ b/plugin/ready/README.md
@@ -25,7 +25,8 @@ ready [ADDRESS] {
 
 *ready* optionally takes an address; the default is `:8181`. The path is fixed to `/ready`. The
 readiness endpoint returns a 200 response code and the word "OK" when this server is ready. It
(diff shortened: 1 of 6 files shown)


glebkin:
Guys, would you please kindly take a look into this PR?

chrisohaver:
>  I didn't find any plugins in CoreDNS implementing ready.Readiness interface.

The kubernetes plugin does. This is a very significant implementation that we want to be extra careful not to introduce changes in behavior that could disrupt existing processes.

chrisohaver:
I think this may change the existing behavior of Kubernetes readiness checks.  It depends on the behavior of the `HasSynced` implementations in the k8s client API.  This change might cause a pod to start reporting unready if `HasSynced` ever becomes false after it is true after the initial synchronization following connection.  This may not be a problematic change, but we still need to be extra careful if introducing any change in behavior.

To be safe, we could bake the old behavior into the kubernetes readiness implementation (e.g. once HasSynced is true, then always return true for subsequent calls without checking HasSynced again).

glebkin:
We could also introduce a new flag in the ready plugin to make it backward compatible. If this is true, we will not remove the plugin from the list if its Ready implementation returns true.

chrisohaver:
> We could also introduce a new flag in the ready plugin to make it backward compatible. If this is true, we will not remove the plugin from the list if its Ready implementation returns true.

yes - as long as the default behavior is the old behavior, this would work as well - but it does add one more mostly superfluous option to the configuration.

Shmillerov:
Hello!

I think it would be right to make a `ready` plugin with regular checks.

From k8s docs: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes

> Readiness probes runs on the container during its whole lifecycle.

I think it's not very good that CoreDNS `ready` plugin can't change the status at runtime.

@chrisohaver, @glebkin tell me what is needed to get things moving?

About `kubernetes` plugin with `HasSynced()`. I looked at the docs and impl and it looks like this parameter is set to true once and can't be changed anymore.

https://pkg.go.dev/k8s.io/client-go/tools/cache#SharedInformer

```
// HasSynced returns true if the shared informer's store has been
// informed by at least one full LIST of the authoritative state
// of the informer's object collection.  This is unrelated to "resync".
//
// Note that this doesn't tell you if an individual handler is synced!!
// For that, please call HasSynced on the handle returned by
// AddEventHandler.
HasSynced() [bool](https://pkg.go.dev/builtin#bool)
```
	
https://pkg.go.dev/k8s.io/client-go/tools/cache#FIFO.HasSynced
```
HasSynced returns true if an Add/Update/Delete/AddIfNotPresent are called first, or the first batch of items inserted by Replace() has been popped.
```

chrisohaver:
I’d prefer this be enabled by flag/option. Else we should defer it to the next minor release.  How about `monitor (until-ready | continuously)`? I’m open to other name ideas.

glebkin:
> I’d prefer this be enabled by flag/option. Else we should defer it to the next minor release. How about `monitor (until-ready | continuously)`? I’m open to other name ideas.

I've updated PR, please take a look

plugin/ready/README.md
```diff
@@ -8,8 +8,7 @@
 
 By enabling *ready* an HTTP endpoint on port 8181 will return 200 OK, when all plugins that are able
 to signal readiness have done so. If some are not ready yet the endpoint will return a 503 with the
-body containing the list of plugins that are not ready. Once a plugin has signaled it is ready it
-will not be queried again.
+body containing the list of plugins that are not ready.
```
chrisohaver, line 11: Add “By default once a plugin has signaled it is ready it will not be queried again.”

glebkin: Done

plugin/ready/setup.go
```diff
@@ -11,10 +11,17 @@ import (
 func init() { plugin.Register("ready", setup) }
 
 func setup(c *caddy.Controller) error {
-	addr, err := parse(c)
+	addr, monType, err := parse(c)
 	if err != nil {
 		return plugin.Error("ready", err)
 	}
+
+	if monType == monitorTypeContinuously {
```
chrisohaver, line 19: Please make this only accept valid values `monitorTypeContinuously` or `monitorTypeUntilReady`, and error if any other value is given.

glebkin: But we are already checking for valid values in [parseMonitorType](https://github.com/coredns/coredns/pull/6975/files#diff-0781887bd10cada1112cdfde220be229f8a5ef79695d93bb4ae1610cff63ad81R111). Are you sure that we need additional validation in the setup function? I don't see any cases where we might end up with an invalid value after the parse function has been called.

chrisohaver: I didn't see it.

chrisohaver (approved):

PR #6997, opened 2024-11-22
Fix release notes

Shmillerov:
coredns.io sync doesn't work correctly because of missing `"` in release notes.

Changes:
- Added a missing `"` in coredns-1.11.3.md header
- Added a header to coredns-1.11.3.md and coredns-1.12.0.md


chrisohaver:
Thanks!

Issue #6991, opened 2024-11-21
Enable exponential backoff to forwarder plugin


salla2:
**What would you like to be added**:
I would like the exponential backoff period feature to be added to the forwarder plugin. This feature should be enabled when all upstream resolvers are not responding.

**Why is this needed**:
We are currently using the serve stale feature to respond from a stale cache when upstream resolvers are not responding. However, we've noticed that when the upstream resolvers are not responding, every request from the pod ends up being forwarded to the upstream servers. This situation bombards the upstream servers and delays their recovery. If we add the backoff feature, it will stop sending every request to the upstream servers after a certain period, while still responding from the cache until the serve stale period expires. This would give the upstream resolvers some breathing space to recover.

chrisohaver:
A better solution IMO to the stated problem is to introduce an option to make the forward plugin return SERVFAIL immediately when processing a request while all upstreams are down, instead of trying to send it to a random upstream.

The random upstream behavior is explained per code comments as an assumption that the health checker must be faulty or not working.  However in an environment where you know that the health checker is functional, then it makes more sense to return SERVFAIL right away than to send the query to an upstream you know will not respond.

salla2:
This change will greatly help, thanks again for your prompt response. I'm assuming our health checks will be performed every 0.5 sec and it will one call per node , so it will be way far better.

Issue #5296, opened 2022-04-01
CoreDNS Cache's serve_stale


roldancer:
Hi All, We are planning to use on Kubernetes the CoreDNS cache's serve_stale parameter for a custon forwarded zone, do you recommed to use serve_stale in production?, we want to improve DNS resilience.

Many thanks for your feedback

chrisohaver:
Should be fine. Be aware that the way the `serve_stale` feature in CoreDNS works follows unbound's `serve-expired` implementation, not RFC-8767.

chrisohaver:
> Be aware that the way the `serve_stale` feature in CoreDNS works follows unbound's `serve-expired` implementation, not RFC-8767.

Of note, by design the implementation favors fast response time over answer correctness.  It will always serve an expired cache entry (younger than DURATION), _before_ checking to see if there is an updated answer.  After responding with the expired entry, it resolves the request, and updates cache if it can resolve.  Therefore, it will respond with outdated answers  to clients until the cache updates (which can be incorrect if the record has changed since it was last cached).

For records that change frequently, an incorrect answer would occur more frequently. E.g. headless services in Kubernetes point to pod IPs, which are not very static (changing each time a pod is respawned). Although if only enabling `serve_stale` for a custom/private zone, Kubernetes records would of course not be affected.

If this caveat is acceptable, then it should be fine in production.

roldancer:
Hi @chrisohaver, Many thanks for your feedback, do you know if someone is using CoreDNS cache's serve_stale parameter for a custon forwarded zone, I mean is there any enterprise reference?

Many thanks

chrisohaver:
I personally don't have any enterprise references.

roldancer:
Hi @chrisohaver , We are planning to use CoreDNS Cache's serve_stale parameter for a custom zone, this zone has DNS records with TTL of 1h, 8h and 24h, what are the most important recommendations or use case for Cache's serve_stale parameters keeping in mind that we need to improve DNS resilience. 
Many thanks in advance for any recommendation, guide, use case or real experience ....

Thanks.

chrisohaver:
To allow CoreDNS to serve answers from cache for a DNS name in a private zone while that zone's DNS server is non-responsive, you will want to set the stale_cache duration to be about as long as the maximum time you would expect the zone's DNS server to be non-responsive.

I'd recommend enabling stale_cache only for the cache plugin in the private zone server block, not in the root server block that contains the kubernetes plugin.

PR #6999, opened 2024-11-25
plugin/forward: added option `failfast_all_unhealthy_upstreams` to return servfail if all upstreams are down

puneetloya:

If all upstream proxies are unavailable, coredns forward plugin returns a servfail to the client and does not call the upstream unless proxies are back to being healthy.

### 2. Issue link?
It addresses #6991 

Forward plugin documentation: 

<pre><code>
forward FROM TO... {
    except IGNORED_NAMES...
    force_tcp
    prefer_udp
    expire DURATION
    max_fails INTEGER
    tls CERT KEY CA
    tls_servername NAME
    policy random|round_robin|sequential
    health_check DURATION [no_rec] [domain FQDN]
    max_concurrent MAX
    next RCODE_1 [RCODE_2] [RCODE_3...]
    <b>failfast_all_unhealthy_upstreams</b>  <i>(THIS IS A NEW OPTION ADDED BY THIS PR)</i> 
}
</code>
</pre>

```
failfast_all_unhealthy_upstreams - determines the handling of requests when all upstream servers are 
unhealthy and unresponsive to health checks. Enabling this option will immediately return SERVFAIL
 responses for all requests. By default, requests are sent to a random upstream.
```

No, it is backward compatible.

### DIFF
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -48,9 +48,10 @@ forward FROM TO... {
     tls CERT KEY CA
     tls_servername NAME
     policy random|round_robin|sequential
-    health_check DURATION [no_rec] [domain FQDN] [on_fail ONFAIL]
+    health_check DURATION [no_rec] [domain FQDN]
     max_concurrent MAX
     next RCODE_1 [RCODE_2] [RCODE_3...]
+    failfast_all_unhealthy_upstreams
 }
 ~~~
 
@@ -91,13 +92,13 @@ forward FROM TO... {
     The flag is default `true`.
   * `domain FQDN` - set the domain name used for health checks to **FQDN**.
     If not configured, the domain name used for health checks is `.`.
-  * `on_fail ONFAIL` - controls how requests are handled when _all_ upstream servers are unhealthy and unresponsive to health checks. Allowed values for **ONFAIL** are `servfail` and `spray`.  `servfail` will immediately return SERVFAIL responses for all requests. `spray` will instead send requests to a random upstream.  The default behavior is `spray`.
 * `max_concurrent` **MAX** will limit the number of concurrent queries to **MAX**.  Any new query that would
   raise the number of concurrent queries above the **MAX** will result in a REFUSED response. This
   response does not count as a health failure. When choosing a value for **MAX**, pick a number
   at least greater than the expected *upstream query rate* * *latency* of the upstream servers.
   As an upper bound for **MAX**, consider that each concurrent query will use about 2kb of memory.
 * `next` If the `RCODE` (i.e. `NXDOMAIN`) is returned by the remote then execute the next plugin. If no next plugin is defined, or the next plugin is not a `forward` plugin, this setting is ignored
+* `failfast_all_unhealthy_upstreams` - determines the handling of requests when all upstream servers are unhealthy and unresponsive to health checks. Enabling this option will immediately return SERVFAIL responses for all requests. By default, requests are sent to a random upstream.
 
 Also note the TLS config is "global" for the whole forwarding proxy if you need a different
 `tls_servername` for different upstreams you're out of luck.
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -48,7 +48,7 @@ forward FROM TO... {
     tls CERT KEY CA
     tls_servername NAME
     policy random|round_robin|sequential
-    health_check DURATION [no_rec] [domain FQDN]
+    health_check DURATION [no_rec] [domain FQDN] [on_fail ONFAIL]
     max_concurrent MAX
     next RCODE_1 [RCODE_2] [RCODE_3...]
 }
@@ -91,6 +91,7 @@ forward FROM TO... {
     The flag is default `true`.
   * `domain FQDN` - set the domain name used for health checks to **FQDN**.
     If not configured, the domain name used for health checks is `.`.
+  * `on_fail ONFAIL` - Allowed values for ONFAIL are **servfail** or **spray**. If healthcheck is broken for an upstream the setting servfail would return servfail response to clients and would not contact the upstreams whereas spray would keep health checking the upstreams even if upstreams are failing. It defaults to **spray**.
 * `max_concurrent` **MAX** will limit the number of concurrent queries to **MAX**.  Any new query that would
   raise the number of concurrent queries above the **MAX** will result in a REFUSED response. This
   response does not count as a health failure. When choosing a value for **MAX**, pick a number
--- a/plugin/forward/README.md
+++ b/plugin/forward/README.md
@@ -91,7 +91,7 @@ forward FROM TO... {
     The flag is default `true`.
   * `domain FQDN` - set the domain name used for health checks to **FQDN**.
     If not configured, the domain name used for health checks is `.`.
-  * `on_fail ONFAIL` - Allowed values for ONFAIL are **servfail** or **spray**. If healthcheck is broken for an upstream the setting servfail would return servfail response to clients and would not contact the upstreams whereas spray would keep health checking the upstreams even if upstreams are failing. It defaults to **spray**.
+  * `on_fail ONFAIL` - controls how requests are handled when _all_ upstream servers are unhealthy and unresponsive to health checks. Allowed values for **ONFAIL** are `servfail` and `spray`.  `servfail` will immediately return SERVFAIL responses for all requests. `spray` will instead send requests to a random upstream.  The default behavior is `spray`.
 * `max_concurrent` **MAX** will limit the number of concurrent queries to **MAX**.  Any new query that would
   raise the number of concurrent queries above the **MAX** will result in a REFUSED response. This
   response does not count as a health failure. When choosing a value for **MAX**, pick a number
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
+	"fmt"
 	"sync/atomic"
 	"time"
 
@@ -45,11 +46,12 @@ type Forward struct {
 
 	nextAlternateRcodes []int
 
-	tlsConfig     *tls.Config
-	tlsServerName string
-	maxfails      uint32
-	expire        time.Duration
-	maxConcurrent int64
+	tlsConfig       *tls.Config
+	tlsServerName   string
+	maxfails        uint32
+	expire          time.Duration
+	maxConcurrent   int64
+	hcFailureAction string
 
 	opts proxy.Options // also here for testing
 
@@ -126,12 +128,18 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 			if fails < len(f.proxies) {
 				continue
 			}
-			// All upstream proxies are dead, assume healthcheck is completely broken and randomly
-			// select an upstream to connect to.
-			r := new(random)
-			proxy = r.List(f.proxies)[0]
 
 			healthcheckBrokenCount.Add(1)
+			// All upstreams are dead, return servfail if all upstreams are down
+			if f.hcFailureAction == "servfail" {
+				upstreamErr = fmt.Errorf("all upstreams down: %s", proxy.Addr())
+				break
+			} else {
+				// assume healthcheck is completely broken and randomly
+				// select an upstream to connect to.
+				r := new(random)
+				proxy = r.List(f.proxies)[0]
+			}
 		}
 
 		if span != nil {
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -46,12 +46,12 @@ type Forward struct {
 
 	nextAlternateRcodes []int
 
-	tlsConfig       *tls.Config
-	tlsServerName   string
-	maxfails        uint32
-	expire          time.Duration
-	maxConcurrent   int64
-	hcFailureAction string
+	tlsConfig      *tls.Config
+	tlsServerName  string
+	maxfails       uint32
+	expire         time.Duration
+	maxConcurrent  int64
+	hcOnFailAction string
 
 	opts proxy.Options // also here for testing
 
@@ -131,7 +131,7 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 
 			healthcheckBrokenCount.Add(1)
 			// All upstreams are dead, return servfail if all upstreams are down
-			if f.hcFailureAction == "servfail" {
+			if f.hcOnFailAction == "servfail" {
 				upstreamErr = fmt.Errorf("all upstreams down: %s", proxy.Addr())
 				break
 			} else {
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
-	"fmt"
 	"sync/atomic"
 	"time"
 
@@ -46,12 +45,12 @@ type Forward struct {
 
 	nextAlternateRcodes []int
 
-	tlsConfig      *tls.Config
-	tlsServerName  string
-	maxfails       uint32
-	expire         time.Duration
-	maxConcurrent  int64
-	hcOnFailAction string
+	tlsConfig                  *tls.Config
+	tlsServerName              string
+	maxfails                   uint32
+	expire                     time.Duration
+	maxConcurrent              int64
+	failfastUnhealthyUpstreams bool
 
 	opts proxy.Options // also here for testing
 
@@ -131,15 +130,13 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 
 			healthcheckBrokenCount.Add(1)
 			// All upstreams are dead, return servfail if all upstreams are down
-			if f.hcOnFailAction == "servfail" {
-				upstreamErr = fmt.Errorf("all upstreams down: %s", proxy.Addr())
+			if f.failfastUnhealthyUpstreams {
 				break
-			} else {
-				// assume healthcheck is completely broken and randomly
-				// select an upstream to connect to.
-				r := new(random)
-				proxy = r.List(f.proxies)[0]
 			}
+			// assume healthcheck is completely broken and randomly
+			// select an upstream to connect to.
+			r := new(random)
+			proxy = r.List(f.proxies)[0]
 		}
 
 		if span != nil {
--- a/plugin/forward/health_test.go
+++ b/plugin/forward/health_test.go
@@ -279,12 +279,15 @@ func TestHealthDomain(t *testing.T) {
 }
 
 func TestHealthAllUpstreamsDownServeFail(t *testing.T) {
-
+	defaultTimeout = 20 * time.Second
 	// Setting bad address for healthcheck to fail
-	p := proxy.NewProxy("TestHealthAllUpstreamsDown", "test-serv-fail.coredns.local", transport.DNS)
+	p := proxy.NewProxy("TestHealthAllUpstreamsDown", "8.8.8.8:5003", transport.DNS)
+	p1 := proxy.NewProxy("TestHealthAllUpstreamsDown", "8.8.8.8:5004", transport.DNS)
 	f := New()
-	f.hcOnFailAction = "servfail"
 	f.SetProxy(p)
+	f.SetProxy(p1)
+	f.hcOnFailAction = "servfail"
+
 	defer f.OnShutdown()
 
 	req := new(dns.Msg)
--- a/plugin/forward/health_test.go
+++ b/plugin/forward/health_test.go
@@ -283,7 +283,7 @@ func TestHealthAllUpstreamsDownServeFail(t *testing.T) {
 	// Setting bad address for healthcheck to fail
 	p := proxy.NewProxy("TestHealthAllUpstreamsDown", "test-serv-fail.coredns.local", transport.DNS)
 	f := New()
-	f.hcFailureAction = "servfail"
+	f.hcOnFailAction = "servfail"
 	f.SetProxy(p)
 	defer f.OnShutdown()
 
--- a/plugin/forward/health_test.go
+++ b/plugin/forward/health_test.go
@@ -277,3 +277,21 @@ func TestHealthDomain(t *testing.T) {
 		t.Errorf("Expected number of health checks with Domain==%s to be %d, got %d", hcDomain, 1, i1)
 	}
 }
+
+func TestHealthAllUpstreamsDownServeFail(t *testing.T) {
+
+	// Setting bad address for healthcheck to fail
+	p := proxy.NewProxy("TestHealthAllUpstreamsDown", "test-serv-fail.coredns.local", transport.DNS)
+	f := New()
+	f.hcFailureAction = "servfail"
+	f.SetProxy(p)
+	defer f.OnShutdown()
+
+	req := new(dns.Msg)
+	req.SetQuestion("example.org.", dns.TypeA)
+
+	resp, _ := f.ServeDNS(context.TODO(), &test.ResponseWriter{}, req)
+	if resp != dns.RcodeServerFailure {
+		t.Errorf("Expected Response code: %d, Got: %d", dns.RcodeServerFailure, resp)
+	}
+}
--- a/plugin/forward/health_test.go
+++ b/plugin/forward/health_test.go
@@ -278,23 +278,78 @@ func TestHealthDomain(t *testing.T) {
 	}
 }
 
-func TestHealthAllUpstreamsDownServeFail(t *testing.T) {
-	defaultTimeout = 20 * time.Second
-	// Setting bad address for healthcheck to fail
-	p := proxy.NewProxy("TestHealthAllUpstreamsDown", "8.8.8.8:5003", transport.DNS)
-	p1 := proxy.NewProxy("TestHealthAllUpstreamsDown", "8.8.8.8:5004", transport.DNS)
+func TestAllUpstreamsDown(t *testing.T) {
+	qs := uint32(0)
+	s := dnstest.NewServer(func(w dns.ResponseWriter, r *dns.Msg) {
+		// count non-healthcheck queries
+		if r.Question[0].Name != "." {
+			atomic.AddUint32(&qs, 1)
+		}
+		// timeout
+	})
+	defer s.Close()
+
+	s1 := dnstest.NewServer(func(w dns.ResponseWriter, r *dns.Msg) {
+		// count non-healthcheck queries
+		if r.Question[0].Name != "." {
+			atomic.AddUint32(&qs, 1)
+		}
+		// timeout
+	})
+	defer s1.Close()
+
+	p := proxy.NewProxy("TestHealthAllUpstreamsDown", s.Addr, transport.DNS)
+	p1 := proxy.NewProxy("TestHealthAllUpstreamsDown2", s1.Addr, transport.DNS)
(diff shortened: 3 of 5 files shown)


puneetloya:
This PR based on the idea from @chrisohaver https://github.com/coredns/coredns/issues/6991#issuecomment-2491978581

puneetloya:
@chrisohaver @johnbelamaric @Tantalor93 @miekg can any of you please review this? 

I've been looking at the other issues related to this and all the history? #4309 and #5296 and RFC 8767.

I think this behavior

https://github.com/coredns/coredns/blob/master/plugin/cache/handler.go#L57

image

This causes more problems, as apps will ask for resolution every time (with TTL being 0) and it only increases overhead during a time when upstream servers are probably in duress.

chrisohaver:
> @chrisohaver @johnbelamaric @Tantalor93 @miekg can any of you please review this?
> 
> I've been looking at the other issues related to this and all the history? #4309 and #5296 and RFC 8767.
> 
> I think this behavior
> 
> https://github.com/coredns/coredns/blob/master/plugin/cache/handler.go#L57
> 
> image
> This causes more problems, as apps will ask for resolution every time (with TTL being 0) and it only increases overhead during a time when upstream servers are probably in duress.

this observation seems unrelated to this PR

chrisohaver (changes requested):
> need to add to health_check `onfailure retry/servfail`

I think `spray` better describes the default behavior than `retry`: e.g. `on_fail ( servfail | spray )` 

`health_check DURATION [no_rec] [domain FQDN] [on_fail ONFAIL]`

where `ONFAIL` can be `servfail` or `spray`

> Infact retry is just dummy, currently I do not check for that string. The option is only useful if you set servfail, if not is assumed it can retry.

It shouldn't just accept anything like that.  It should accept only from a set of values, and other values should render an error.

puneetloya:
> > need to add to health_check `onfailure retry/servfail`
> 
> I think `spray` better describes the default behavior than `retry`: e.g. `on_fail ( servfail | spray )`
> 
> `health_check DURATION [no_rec] [domain FQDN] [on_fail ONFAIL]`
> 
> where `ONFAIL` can be `servfail` or `spray`
 
Changed this  in the code and added to Readme.

>  Infact retry is just dummy, currently I do not check for that string. The option is only useful if you set servfail, if not is assumed it can retry.
> It shouldn't just accept anything like that. It should accept only from a set of values, and other values should render an error.

Ya I should have worded that better. When I meant "dummy" I was actually checking for those values but setting retry (now spray) would be the default behavior.

puneetloya:
> > This causes more problems, as apps will ask for resolution every time (with TTL being 0) and it only increases overhead during a time when upstream servers are probably in duress.
> 
> this observation seems unrelated to this PR

I can create a separate PR for it. What should the default behavior(ttl) be ? Should it default to keepttl when it is serving from stale cache?

puneetloya:
@chrisohaver can you please review?

plugin/forward/README.md
```diff
@@ -91,6 +91,7 @@ forward FROM TO... {
     The flag is default `true`.
   * `domain FQDN` - set the domain name used for health checks to **FQDN**.
     If not configured, the domain name used for health checks is `.`.
+  * `on_fail ONFAIL` - Allowed values for ONFAIL are **servfail** or **spray**. If healthcheck is broken for an upstream the setting servfail would return servfail response to clients and would not contact the upstreams whereas spray would keep health checking the upstreams even if upstreams are failing. It defaults to **spray**.
```
chrisohaver: * the description of how spray works is incorrect.  
* health checking always occurs when an upstream is unhealthy, regardless of this setting. 
* this new `on_fail` setting only comes into effect when _all_ upstreams are down, not just any one. 
 
Please use the following description instead:

```suggestion
  * `on_fail ONFAIL` - controls how requests are handled when _all_ upstream servers are unhealthy and unresponsive to health checks. Allowed values for **ONFAIL** are `servfail` and `spray`.  `servfail` will immediately return SERVFAIL responses for all requests. `spray` will instead send requests to a random upstream.  The default behavior is `spray`.
```

chrisohaver (changes requested):

plugin/forward/README.md
puneetloya: reworded the description as per your suggestion

puneetloya:
@chrisohaver I have done the changes based on the [Codecov](https://app.codecov.io/gh/coredns/coredns/pull/6999?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=coredns) report, can you please check again?

puneetloya:
@chrisohaver @johnbelamaric @Tantalor93 @miekg can any of you please review this?

puneetloya:
Hi Folks, Just trying to bring this up again? Can you please take a look?

plugin/forward/README.md
```diff
@@ -48,7 +48,7 @@ forward FROM TO... {
     tls CERT KEY CA
     tls_servername NAME
     policy random|round_robin|sequential
-    health_check DURATION [no_rec] [domain FQDN]
+    health_check DURATION [no_rec] [domain FQDN] [on_fail ONFAIL]
```
rdrozhdzh: frankly speaking, the new setting is not a health check setting. Also, multiple optional parameters complicate parsing and look cumbersome.
Consider adding a separate option like `trust_health_check` or `dont_try_dead_proxies` or smth like this.

plugin/forward/forward.go
```diff
@@ -45,11 +46,12 @@ type Forward struct {
...
-	tlsServerName string
-	maxfails      uint32
-	expire        time.Duration
-	maxConcurrent int64
+	tlsConfig      *tls.Config
+	tlsServerName  string
+	maxfails       uint32
+	expire         time.Duration
+	maxConcurrent  int64
+	hcOnFailAction string
```
rdrozhdzh: boolean is enough to control the new behavior. Checking boolean in operation flow is more concise and cheaper

plugin/forward/health_test.go
```diff
@@ -277,3 +277,24 @@ func TestHealthDomain(t *testing.T) {
 		t.Errorf("Expected number of health checks with Domain==%s to be %d, got %d", hcDomain, 1, i1)
 	}
 }
+
+func TestHealthAllUpstreamsDownServeFail(t *testing.T) {
+	defaultTimeout = 20 * time.Second
```
rdrozhdzh: here and in other tests - modifying global variable may affect other tests. Also, please justify, why have you chosen 20s? This potentially may affect the test running time.

plugin/forward/health_test.go
```diff
@@ -277,3 +277,24 @@ func TestHealthDomain(t *testing.T) {
...
+	// Setting bad address for healthcheck to fail
+	p := proxy.NewProxy("TestHealthAllUpstreamsDown", "8.8.8.8:5003", transport.DNS)
+	p1 := proxy.NewProxy("TestHealthAllUpstreamsDown", "8.8.8.8:5004", transport.DNS)
```
rdrozhdzh: please don't use the real worlds IP addresses. 1. You should not create unneeded load to a third-party services. 2. You never know which services are going to be run on these ports now and in future.

plugin/forward/forward.go
```diff
@@ -126,12 +128,18 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
...
-			// select an upstream to connect to.
-			r := new(random)
-			proxy = r.List(f.proxies)[0]
 
 			healthcheckBrokenCount.Add(1)
+			// All upstreams are dead, return servfail if all upstreams are down
+			if f.hcOnFailAction == "servfail" {
+				upstreamErr = fmt.Errorf("all upstreams down: %s", proxy.Addr())
+				break
+			} else {
```
rdrozhdzh: the `else` is not needed after `break`

plugin/forward/README.md
puneetloya: I agree. What about `skip_unhealthy_upstreams` ? I am ok with `dont_try_dead_proxies`. Its just that sometimes negative options tend to cause confusion.

puneetloya: Naming this is getting tricky. I asked genai about it. Let me know if you think anything seems appropriate?

image

image

puneetloya: another one `backoff_retrying_dead_proxies`

plugin/forward/health_test.go
puneetloya: My test case is: I need a way for checking upstreams that are down or unreachable but not invalid. If I give invalid  addresses like example2.local:53, I am getting 
`dial udp: lookup example2.local` 
but I am actually trying to get: 
`all upstreams down`.

plugin/forward/health_test.go
puneetloya: This was probably because I am using real world IP addresses. I have explained below why I was trying to use a valid IP.

plugin/forward/README.md
rdrozhdzh: I know, it's not easy to come up with a good name, I was not able to :)

`skip_unhealthy_upstreams` - I though about this as well, but in fact we normally skip unhealthy upstreams even in existing code if some of them are healthy and another ones are unhealthy. We need to denote the case when ALL upstreams are unhealthy.

From the variants suggested by AI, I like `failfast_all_unhealthy_upstreams`. Lets see what other people think.

plugin/forward/health_test.go
rdrozhdzh: You can try loopback address like `127.0.5.3:5003`.

plugin/forward/health_test.go
```diff
@@ -277,3 +277,24 @@ func TestHealthDomain(t *testing.T) {
...
+	f.SetProxy(p1)
+	f.hcOnFailAction = "servfail"
+
+	defer f.OnShutdown()
+
+	req := new(dns.Msg)
+	req.SetQuestion("example.org.", dns.TypeA)
+
+	resp, _ := f.ServeDNS(context.TODO(), &test.ResponseWriter{}, req)
+	if resp != dns.RcodeServerFailure {
```
rdrozhdzh, line 328: this comment somehow missed in my previous review..
With the condition above, you cannot check if your new code works properly. It returns SERVFAIL even if you retry when all proxies are died.

puneetloya: Agreed. I updated the test to check for the error message "all upstreams down" but its not working with loopback addresses. It does not work for 8.8.8.8 addresses either.

image
image

puneetloya: Fixed this, there was a way to mimic the proxy server behavior in other tests. I am checking for servfail and `all upstreams down` message.

plugin/forward/forward.go
puneetloya: Fixed

plugin/forward/health_test.go
puneetloya: Fixed, the test is moved to forward_test.go as its no more a healthcheck option

plugin/forward/health_test.go
puneetloya: Fixed. Removed this global setting from the test.

plugin/forward/forward.go
puneetloya: Fixed

plugin/forward/README.md
puneetloya: For now I have added `failfast_all_unhealthy_upstreams`. I can easily change if we have a better name.

puneetloya: Updated the  description of the PR as well.

plugin/forward/README.md
```diff
@@ -97,6 +97,7 @@ forward FROM TO... {
   at least greater than the expected *upstream query rate* * *latency* of the upstream servers.
   As an upper bound for **MAX**, consider that each concurrent query will use about 2kb of memory.
 * `next` If the `RCODE` (i.e. `NXDOMAIN`) is returned by the remote then execute the next plugin. If no next plugin is defined, or the next plugin is not a `forward` plugin, this setting is ignored
+* `failfast_all_unhealthy_upstreams` - determines the handling of requests when all upstream servers are unhealthy and unresponsive to health checks. Enabling this option will immediately return SERVFAIL responses for all requests. By default, requests are sent to a random upstream.
```
rdrozhdzh, line 101: please update also syntax definition above (as you did in PR description)

plugin/forward/forward_test.go
```diff
@@ -74,3 +79,45 @@ func TestSetTapPlugin(t *testing.T) {
...
+	})
+
+	s1 := dnstest.NewServer(func(w dns.ResponseWriter, r *dns.Msg) {
+		// timeout
+	})
+	p := proxy.NewProxy("TestHealthAllUpstreamsDown", s.Addr, transport.DNS)
+	p1 := proxy.NewProxy("TestHealthAllUpstreamsDown2", s1.Addr, transport.DNS)
+
+	p.GetHealthchecker().SetReadTimeout(10 * time.Millisecond)
+	p.GetHealthchecker().SetWriteTimeout(20 * time.Millisecond)
```
rdrozhdzh: inconsistent value, why 20 in this case?

plugin/forward/forward_test.go
```diff
@@ -74,3 +79,45 @@ func TestSetTapPlugin(t *testing.T) {
...
+	p.GetHealthchecker().SetDomain("example.org.")
+	p1.GetHealthchecker().SetReadTimeout(10 * time.Millisecond)
+	p1.GetHealthchecker().SetWriteTimeout(10 * time.Millisecond)
+	p1.GetHealthchecker().SetDomain("example1.org.")
```
rdrozhdzh: SetDomain() and SetWriteTimeout() are not really needed in your test, the default values should be good

plugin/forward/forward_test.go
```diff
@@ -74,3 +79,45 @@ func TestSetTapPlugin(t *testing.T) {
...
+	f := New()
+	f.SetProxy(p)
+	f.SetProxy(p1)
+	f.failfastUnhealthyUpstreams = true
+
+	defer f.OnShutdown()
+
+	var resp int
+	var msg error
+	for i := 0; i < 5; i++ {
```
rdrozhdzh: this loop looks awkward

plugin/forward/forward_test.go
```diff
@@ -74,3 +79,45 @@ func TestSetTapPlugin(t *testing.T) {
...
+	}
+
+	if resp != dns.RcodeServerFailure {
+		t.Errorf("Expected Response code: %d, Got: %d", dns.RcodeServerFailure, resp)
+	}
+
+	if !strings.Contains(msg.Error(), "all upstreams down") {
+		t.Errorf("Expected error message: all upstreams down , Got: %s", msg.Error())
+	}
+}
```
rdrozhdzh: your test still doesn't verify the functionality you added.

Basically, you should run query two times - one with failfastUnhealthyUpstreams enabled, another one with failfastUnhealthyUpstreams disabled. Your test should assure that behavior in the two cases is different.

For example, 
- you can update the Handlers in your test servers to count the requests for your test domain e.g. "example.org." (don't count probe queries from healthcheck)
- you can consider configuring forward plugin with the options which facilitate quick switching proxies to unhealthy state faster, e.g. `f.maxfails = 1` or `f.hcInterval = 10 * time.Millisecond`
- or you can quickly switch proxies to failed state by direct calls like `p.GetHealthchecker().Check(p)`
- in preparation phase, you should make sure all proxies are down (p.Down()) before sending the test queries
- enable failfast option, send test query to forward, check the number of intercepted queries by your test servers
- disable failfast option, send test query to forward, check the number of intercepted queries by your test servers

plugin/forward/setup_test.go
```diff
@@ -382,3 +382,45 @@ func TestNextAlternate(t *testing.T) {
 		}
 	}
 }
+
+func TestFailfastAllUnhealthyUpstreams(t *testing.T) {
+	tests := []struct {
+		input          string
+		shouldErr      bool
```
rdrozhdzh: this field looks redundant, it can be inferred as `expectedErr != ""`. Redundant data complicates code and may lead to ambiguities, inconsistencies and contradictions.

rdrozhdzh:
in PR description

4. Does this introduce a backward incompatible change or deprecation?

Yes it is backward compatible.

should read as

4. Does this introduce a backward incompatible change or deprecation?

**No**, it is backward compatible.

plugin/forward/forward_test.go
puneetloya: Nothing specific, I can change that.

plugin/forward/forward_test.go
puneetloya: The reason for this loop was, if I immediately request just once, it finishes before Proxy is marked down and hence does not go into the check of all_upstreams_down. If there is a way to mark Proxy's down beforehand. I see your comment below let me see if that works.

plugin/forward/forward_test.go
puneetloya: The test does check the functionality by checking the error message which is only returned when `failfast_all_unhealthy_upstreams` is set.

puneetloya: I will definitely add the cases you stated above

plugin/forward/README.md
puneetloya: Fixed

plugin/forward/forward_test.go
puneetloya: Fixed. I've put what you suggested in the same test case.

plugin/forward/setup_test.go
puneetloya: Fixed.

plugin/forward/forward_test.go
puneetloya: This is removed. Fixed.

plugin/forward/forward_test.go
puneetloya: Removed

plugin/forward/forward_test.go
puneetloya: Fixed

puneetloya:
@rdrozhdzh I may have taken the liberty to resolve some of the conversations to keep things clear on this PR. Please unresolve anything if required.

plugin/forward/README.md
```diff
@@ -48,9 +48,10 @@ forward FROM TO... {
     tls CERT KEY CA
     tls_servername NAME
     policy random|round_robin|sequential
-    health_check DURATION [no_rec] [domain FQDN]
+    health_check DURATION [no_rec] [domain FQDN] [on_fail ONFAIL]
```
rdrozhdzh: please restore this `health_check` option syntax

plugin/forward/forward.go
```diff
@@ -126,12 +128,18 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 			if fails < len(f.proxies) {
 				continue
 			}
-			// All upstream proxies are dead, assume healthcheck is completely broken and randomly
+
+			healthcheckBrokenCount.Add(1)
+			// All upstreams are dead, return servfail if all upstreams are down
+			if f.failfastUnhealthyUpstreams {
+				upstreamErr = fmt.Errorf("all upstreams down: %s", proxy.Addr())
```
rdrozhdzh: [lower importance] 
The error message looks inconsistent, it mentions "**all** upstreams" whereas you append only one of upstreams to the message. Do you really need to mention this upstream?

Returning the pre-built error like `ErrNoHealthy` is cheaper, so you can just break here. `ServeDNS()` will return either last `upstreamErr` (if set) or `ErrNoHealthy`.

Another possible benefit of reusing the existing error message is that other plugins (either standard or custom) may analyze this error message for their own purposes. Introducing new error message (especially with non-constant text) may lead to unnecessary complications.

plugin/forward/forward_test.go
```diff
@@ -74,3 +80,80 @@ func TestSetTapPlugin(t *testing.T) {
...
+	// i.e, fails > maxfails
+	for range 2 {
+		p.GetHealthchecker().Check(p)
+		p1.GetHealthchecker().Check(p1)
+	}
+
+	defer f.OnShutdown()
+
+	// Check if all proxies are down
+	if p.Down(1) && p1.Down(1) {
```
rdrozhdzh: Don't hardcode `1` here, you should use exactly what configured in forward plugin, the same way as `ServeDNS()` does, i.e.
```go
  if p.Down(f.maxfails) && p1.Down(f.maxfails) {
    // test code
```
or even better (more idiomatic)
```go
  if !p.Down(f.maxfails) || !p1.Down(f.maxfails) {
    t.Fatalf("Expected all proxies to be down")
  }
  // test code
```

plugin/forward/forward_test.go
```diff
@@ -74,3 +80,80 @@ func TestSetTapPlugin(t *testing.T) {
...
+		req := new(dns.Msg)
+		req.SetQuestion("example.org.", dns.TypeA)
+		resp, msg := f.ServeDNS(context.TODO(), &test.ResponseWriter{}, req)
```
rdrozhdzh: [minor]
name `msg` looks confusing, it gives impression that returned type is string. Consider naming `err` instead

plugin/forward/forward_test.go
```diff
@@ -74,3 +80,80 @@ func TestSetTapPlugin(t *testing.T) {
...
+
+		req = new(dns.Msg)
+		req.SetQuestion("example.org.", dns.TypeA)
+		_, msg = f.ServeDNS(context.TODO(), &test.ResponseWriter{}, req)
+		if strings.Contains(msg.Error(), "all upstreams down") {
+			t.Errorf("Expected error message: all upstreams down , Got: %s", msg.Error())
+		}
+
+		if qs == 0 {
+			t.Errorf("Expected queries to the upstream: 1, Got: %d", qs)
```
rdrozhdzh: inconsistency between condition and message text

plugin/forward/forward_test.go
rdrozhdzh: > The test does check the functionality by checking the error message which is only returned when `failfast_all_unhealthy_upstreams` is set.

Sorry, I overlooked that the error message is different. Anyway, you may decide to remove this new message, as I suggested above.

plugin/forward/setup_test.go
```diff
@@ -382,3 +382,40 @@ func TestNextAlternate(t *testing.T) {
...
+			if !strings.Contains(err.Error(), test.expectedErr) {
+				t.Errorf("Test %d: expected error to contain: %v, found error: %v, input: %s", i, test.expectedErr, err, test.input)
+			}
+		}
+
+		if test.expectedErr != "" {
+			continue
+		}
+
+		f := fs[0]
```
rdrozhdzh, line 420: this may panic if `parseForward(c)` returns zero-length slice (because of error or by other reasons). Is this what intended?

plugin/forward/setup_test.go
```diff
@@ -382,3 +382,40 @@ func TestNextAlternate(t *testing.T) {
...
+		fs, err := parseForward(c)
+
+		if err != nil {
+			if test.expectedErr == "" {
+				t.Errorf("Test %d: expected no error but found one for input %s, got: %v", i, test.input, err)
+			}
+			if !strings.Contains(err.Error(), test.expectedErr) {
+				t.Errorf("Test %d: expected error to contain: %v, found error: %v, input: %s", i, test.expectedErr, err, test.input)
+			}
+		}
```
rdrozhdzh, line 414:
```go
} else {
  // check the case when you expect error, but parseForward(c) succeeds
```

plugin/forward/forward_test.go
```diff
@@ -74,3 +80,80 @@ func TestSetTapPlugin(t *testing.T) {
...
+	p1.GetHealthchecker().SetReadTimeout(10 * time.Millisecond)
+
+	f := New()
+	f.SetProxy(p)
+	f.SetProxy(p1)
+	f.failfastUnhealthyUpstreams = true
+	f.maxfails = 1
+	// Make proxys fail by checking health twice
+	// i.e, fails > maxfails
+	for range 2 {
```
rdrozhdzh: consider rewriting as `f.maxfails + 1` instead of `2`

plugin/forward/forward_test.go
puneetloya: Fixed

plugin/forward/forward_test.go
puneetloya: I am checking for the failure condition, if there are zero queries to the upstream but I am expecting 1.

plugin/forward/forward_test.go
puneetloya: Fixed

plugin/forward/forward_test.go
puneetloya: Fixed

plugin/forward/forward.go
puneetloya: Fixed

plugin/forward/README.md
puneetloya: Fixed

plugin/forward/setup_test.go
puneetloya: I copied this from a test above. Error checking is already done, if it errors it would not come to this line. I can do a length check if you are suggesting that?

plugin/forward/forward_test.go
puneetloya: Fixed, using `qs != 1` instead.

plugin/forward/setup_test.go
rdrozhdzh: you prevent executing this code if you expect error. But in case you get error unexpectedly, after reporting the check error, the test will crash. Not so critical, since the test framework will catch this panic anyway.

plugin/forward/setup_test.go
```diff
@@ -382,3 +382,44 @@ func TestNextAlternate(t *testing.T) {
...
+		if err != nil {
+			if test.expectedErr == "" {
+				t.Errorf("Test %d: expected no error but found one for input %s, got: %v", i, test.input, err)
+			}
+			if !strings.Contains(err.Error(), test.expectedErr) {
+				t.Errorf("Test %d: expected error to contain: %v, found error: %v, input: %s", i, test.expectedErr, err, test.input)
+			}
+		} else {
+			if test.expectedErr != "" {
+				t.Errorf("Test %d: expected error but found %s for input %s", i, err, test.input)
```
rdrozhdzh: ... but found no error ...

plugin/forward/forward_test.go
```diff
@@ -74,3 +80,78 @@ func TestSetTapPlugin(t *testing.T) {
...
+		t.Errorf("Expected queries to the upstream: 0, Got: %d", qs)
+	}
+
+	// set failfast to false to check if queries get answered
+	f.failfastUnhealthyUpstreams = false
+
+	req = new(dns.Msg)
+	req.SetQuestion("example.org.", dns.TypeA)
+	_, err = f.ServeDNS(context.TODO(), &test.ResponseWriter{}, req)
+	if err == ErrNoHealthy {
```
rdrozhdzh: will it return kind of "read UDP timeout" in this case?

rdrozhdzh (approved):
Overall, I'm Ok with this change, see minor comments
This PR is still likely has to be approved by any of maintainers

plugin/forward/forward_test.go
puneetloya: Yes it will.

puneetloya: The idea was to check that it does not return `ErrNoHealthy`.

plugin/forward/setup_test.go
puneetloya: fixed

chrisohaver (approved):

puneetloya:
@chrisohaver I've fixed the linting fixes but these test failures are not related to my PR. https://github.com/coredns/coredns/actions/runs/13714431365/job/38374388890

PR #7010, opened 2024-11-26
Try to parse X-Real-IP header in DoH server

poscat0x04:
to determine the requesting IP address when behind a reverse proxy

Currently coredns cannot correctly determine the origin IP address when used behind a reverse proxy
such as nginx. This can lead to issues for plugins that depends on origin IP addresses
such as ends0 rewrite (which will set the the subnet IP to be set to 127.0.0.1 when used in a typical configuration and
cause issues).

AFAIK none.

I don't think this the behavior was documented so I guess none.

I'm not very familiar with the coredns codebase but I would assume no.

### DIFF
--- a/core/dnsserver/server_https.go
+++ b/core/dnsserver/server_https.go
@@ -160,9 +160,16 @@ func (s *ServerHTTPS) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Create a DoHWriter with the correct addresses in it.
 	h, p, _ := net.SplitHostPort(r.RemoteAddr)
 	port, _ := strconv.Atoi(p)
+	realIp := r.Header.Get("X-Real-IP") // Since we might be placed behind a reverse proxy
+	ip := net.ParseIP(realIp)
+
+	if ip == nil {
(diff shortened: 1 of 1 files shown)


poscat0x04:
cc @chrisohaver any progress?

thevilledev:
Hey,

Thank you for the PR and for surfacing the problem.

However, I think this approach has a fundamental problem. Unconditionally trusting a user-supplied HTTP header introduces additional risks. Spoofing the source address could break IP-based ACLs, rate limiting, EDNS0 rewrites and so on. Anything that trusts the client source IP address. The header should only be honored when the direct connection comes from a known proxy address.

A feature like this would have to be opt-in and configurable. At least:

- A configuration directive to enable trusted-proxy behaviour
- A way to specify trusted proxy CIDR ranges
- Consideration for which headers to support (`X-Real-IP`, `X-Forwarded-For`, `Forwarded` per [RFC 7239](https://datatracker.ietf.org/doc/html/rfc7239))
- Tests and documentation

That being said, this is a non-trivial feature. I would suggest opening an issue first to discuss the approach before implementation.

PR #7017, opened 2024-12-03
Add Windows Service support

ZetaTwo:

This PR adds support to run CoreDNS as a Windows service.

1. It splits Run into two functions: `Run` which contains most of the old function except it returns the instance object instead of waiting for it, and `RunForever` which creates and instance and waits for it (i.e. exactly the same behaviour as Run used to have).
2. It moves the flag.Parse call earlier on the execution
3. It adds a Windows-specific entrypoint which can run the program just like before or in service mode
4. If executed in service mode, it starts the Windows Service event loop, starts the server instance and maintains it until the service exits.

In short, there are no changes in behaviour except when launched as a service on Windows or when launched with the `-service` flag on Windows which will start the service debug mode.

There is no issue for this

A note could be added that CoreDNS can be run as a Windows service. The new `-service` flag on Windows allows to run the service in debug mode (i.e. in a terminal but using the service interface).

If someone was previously calling `.Run()` from other code they would now have to call `.RunForever`. If this is unacceptable, it's easy to change so that the the functions are called for example `.CreateInstance` and `.Run` so that `.Run` maintains it function.

### DIFF
--- a/coredns.go
+++ b/coredns.go
@@ -1,13 +1,18 @@
+//go:build !windows
+
 package main
 
 //go:generate go run directives_generate.go
 //go:generate go run owners_generate.go
 
 import (
+	"flag"
+
 	_ "github.com/coredns/coredns/core/plugin" // Plug in CoreDNS.
 	"github.com/coredns/coredns/coremain"
 )
 
 func main() {
-	coremain.Run()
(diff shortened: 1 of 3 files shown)


thevilledev:
Done in #7962

PR #7018, opened 2024-12-04
Set IP_TRANSPARENT on the sockets

kosmas-valianos:
It enables coreDNS to work as a transparent DNS proxy. You can capture DNS requests going to IP A and redirect them to coreDNS IP B. A classic example setup is to use `TPROXY` in the `mangle` table in the `PREROUTING` chain to redirect the packets to coreDNS

Not related to any Open issue. Needed in our setup to have coreDNS as transparent DNS Proxy and at the moment we have to patch coreDNS.

In places that it is mentioned that coreDNS supports/sets `SO_REUSEPORT` it can be mentioned that it also supports/sets `IP_TRANSPARENT`

`IP_TRANSPARENT` is supported since Linux 2.6.24

### DIFF
--- a/plugin/pkg/reuseport/listen_reuseport.go
+++ b/plugin/pkg/reuseport/listen_reuseport.go
@@ -17,6 +17,9 @@ func control(network, address string, c syscall.RawConn) error {
 		if err := unix.SetsockoptInt(int(fd), unix.SOL_SOCKET, unix.SO_REUSEPORT, 1); err != nil {
 			log.Warningf("Failed to set SO_REUSEPORT on socket: %s", err)
 		}
+		if err := unix.SetsockoptInt(int(fd), unix.SOL_IP, unix.IP_TRANSPARENT, 1); err != nil {
+			log.Warningf("Failed to set IP_TRANSPARENT on socket: %s", err)
+		}
 	})
 	return nil
 }


kosmas-valianos:
Any feedback here?

dilyevsky (changes requested):
This change will require CAP_NET_ADMIN which I don't believe we currently require. I don't think this should be turned on by default.

kosmas-valianos:
> This change will require CAP_NET_ADMIN which I don't believe we currently require. I don't think this should be turned on by default.

How should it be controlled then? Ideally it would be with a "core" plugin, but plugins are high level and they don't hook in such low level things

dilyevsky:
I think top-level `ip_transparent` config option should suffice. The other thing I noticed is `reuseport.Listen` is called from a bunch of places (pprof and others) so another reason why it can't be default.

kosmas-valianos:
> I think top-level `ip_transparent` config option should suffice. The other thing I noticed is `reuseport.Listen` is called from a bunch of places (pprof and others) so another reason why it can't be default.

Hmm do you mean something like the `bind` where it  basically edits the `type Config struct {` ? To set a boolean in that Config struct which can been toggled by `ip_transparent`. That would work to set the Config but then somehow needs to make it all the way to the `reuseport` package and that package is Config agnostic. To make matters worse those Listen() in there are basically called from
```
// ListenPacket implements caddy.UDPServer interface.
func (s *Server) ListenPacket() (net.PacketConn, error) {
	p, err := reuseport.ListenPacket("udp", s.Addr[len(transport.DNS+"://"):])
```
which means you cannot simply pass Config in there. I don't see any non invasive way to do it :cry:  If there is an example in the codebase with such a low level thing being configurable point me at it.

PR #7047, opened 2024-12-24
Add ECS support to cache and add ECS plugin

### LINKED ISSUE #6023: EDNS0 CLIENT-SUBNET entry does not match the client request.
**What happened**:

EDNS0 CLIENT-SUBNET entry does not match the client request.

```
# dig @localhost www.example.com +subnet=110.10.24.0/24
...
;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; CLIENT-SUBNET: 127.0.0.1/32/0
...
```

CLIENT-SUBNET is changed to the requested value (subnet). 

```
# dig @localhost www.example.com +subnet=110.10.24.0/24
...
;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; CLIENT-SUBNET: 110.10.24.0/24/0
...
```

coredns version
```
# ./coredns --version
CoreDNS-1.10.1
linux/amd64, go1.18.6,
```

my corefile
```
. {
  reload 5s
  log . common
  nsid
  bufsize 1232
  rewrite edns0 subnet set 32 56
  rewrite edns0 local set 0x8 {geoip/ecs} # it's not working
  geoip /home/coredns/GeoLite2-City_20230411/GeoLite2-City.mmdb {
        edns-subnet
  }
  metadata
  auto {
        directory /tmp/zones ^db\.(.*) {1}
        reload 10s
        }
}
```

zone info
```
# cat /tmp/zones/db.example.com
$ORIGIN example.com.
@       3600 IN SOA sns.dns.icann.org. noc.dns.icann.org. 5 7200 3600 1209600 3600

www     IN A     127.0.0.4
        IN AAAA  ::1
```

**Environment**:

- the version of CoreDNS:
- Corefile:
- logs, if applicable:
```
# cat /etc/os-release
NAME="AlmaLinux"
VERSION="9.0 (Emerald Puma)"
ID="almalinux"
ID_LIKE="rhel centos fedora"
VERSION_ID="9.0"
PLATFORM_ID="platform:el9"
PRETTY_NAME="AlmaLinux 9.0 (Emerald Puma)"
ANSI_COLOR="0;34"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:almalinux:almalinux:9::baseos"
HOME_URL="https://almalinux.org/"
DOCUMENTATION_URL="https://wiki.almalinux.org/"
BUG_REPORT_URL="https://bugs.almalinux.org/"

ALMALINUX_MANTISBT_PROJECT="AlmaLinux-9"
ALMALINUX_MANTISBT_PROJECT_VERSION="9.0"
REDHAT_SUPPORT_PRODUCT="AlmaLinux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.0"
```
- Others:

### LINKED ISSUE #7011: coredns does not retry the request with ecs disabled when it receives a REFUSED message from an upstream recursive resolver
**What happened**:
When requesting an upstream with ecs enabled, coredns does not retry the request with ecs disabled when it receives a REFUSED message.

According to the [rfc](https://datatracker.ietf.org/doc/html/rfc7871#section-7.1.3):

>  If the Forwarding Resolver receives a REFUSED response when it sends a query that includes a non-zero ADDRESS, it MUST
retry with no ADDRESS.

I expect coredns to follow the specification.

Start the server with
```
dns://.5353 {
  bind lo
  forward . 8.8.8.8:53
}
```

Run `dig @127.0.0.1 +subnet=1.0.0.0/24 e2490.dscf.akamaiedge.net`.
Optionally run `dig @127.0.0.1 e2490.dscf.akamaiedge.net` to observe that the upstream does not return `REFUSED` when ecs is disabled.

**Environment**:

- the version of CoreDNS:
- Corefile: CoreDNS-1.11.3
- logs, if applicable: N/A
- OS (e.g: `cat /etc/os-release`): N/A

### LINKED ISSUE #7036: cache plugin does not use the returned EDNS Client Subnet in the key (ECS)
**What happened**:

The cache plugin uses only the qname, qtype, do and cd flags when caching responses containing to queries containing EDNS client subnet options (ECS).

This does not respect cache considerations in [RFC 7871](https://datatracker.ietf.org/doc/html/rfc7871#section-7.3).

The cache plugin must use the subnet and prefix returned in the response, after appropriate validation as specified in the RFC.

```
. {
  cache
  forward . 8.8.8.8
}
```

```
dig google.com @127.0.0.1 +subnet=208.67.222.0/24
dig google.com @127.0.0.1 +subnet=102.38.255.0/24
```

Must return different IP addresses, but it doesn't.

**Environment**:

- the version of CoreDNS: CoreDNS-1.12.0
- OS (e.g: `cat /etc/os-release`): darwin/arm64, go1.23.3

danog:

This pull requests implements [RFC 7871](https://www.rfc-editor.org/rfc/rfc7871), EDNS client subnet support.

It does so by:

1. Adding an optional `ecs` plugin that adds the ECS option to requests, based on the source IP of the user
2. Modifying the cache plugin by caching results based on the ECS subnets specified in the requests and responses, according to the RFC.
3. Modifying the forward plugin to retry queries without ECS data in case of failure, following the RFC.

Fixes https://github.com/coredns/coredns/issues/7036
Fixes https://github.com/coredns/coredns/issues/6023
Fixes https://github.com/coredns/coredns/issues/7011

All added to this pull request.

No.


danog:
Now also fixes https://github.com/coredns/coredns/issues/7011

plugin/forward/forward.go
```diff
@@ -161,6 +172,15 @@ func (f *Forward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg
 				opts.ForceTCP = true
 				continue
 			}
+
+			// Retry without ECS option in case of REFUSED as per RFC 7871
```
chrisohaver: It would be better IMO, if we can move the retry logic out of forward and into the ecs plugin itself (the ecs plugin sending the query back down the chain with out the ecs data if the response was REFUSED etc).  IIUC, the server side retry logic is only really necessary when ecs is being added server side (by the ecs plugin) in which case the client is unaware of the ecs in the request.  In the case it is already specified in the incoming request, the client should be aware and ready to handle REFUSED responses how it deems fit.

plugin/cache/README.md
```diff
@@ -75,6 +77,8 @@ cache [TTL] [ZONES...] {
   of the remaining TTL. This can be useful if CoreDNS is used as an authoritative server and you want
   to serve a consistent TTL to downstream clients. This is **NOT** recommended when CoreDNS is caching
   records it is not authoritative for because it could result in downstream clients using stale answers.
+* `mask_v4` specifies the maximum cachable IPv4 prefix size, for queries containing ECS (EDNS0 Client Subnet) data.
+* `mask_v6` specifies the maximum cachable IPv6 prefix size, for queries containing ECS (EDNS0 Client Subnet) data.
```
chrisohaver, line 81: Looking forward to other possible edns0 protocol support, I think this would be better formatted as one option. Something like `edns PROTOCOL [OPTIONS...]`, where PROTOCOL is `ecs`, and the options configure the subnet masks. for example:

`edns ecs v4 32 v6 128`

danog:
Ping @miekg? :)

danog:
Pinging @miekg again :)

chrisohaver:
Miek is not currently active in the project.

I just noticed I have some pending feedback from months ago that I didn’t submit. Submitting now…

plugin/cache/README.md
danog: Wouldn't this decrease readability?

plugin/forward/forward.go
danog: Considering that ECS-specific logic was also added to the `cache` plugin, it makes no sense to extract it for the `forward` plugin IMO, the retrying logic is required for forwarding resolvers by [section 7.1.3](https://www.rfc-editor.org/rfc/rfc7871#section-7.1.3).

danog:
Ping!

Anankke:
Any update on this?

plugin/forward/forward.go
Kentzo: I agree with @chrisohaver. CoreDNS should only retry forwarding request without ECS _iff_ CoreDNS added it. If the option was added by the client, CoreDNS should not retry but instead send REFUSED back.

Kentzo:
It seems to partially duplicate functionality of the [rewrite](https://coredns.io/plugins/rewrite/#edns0_subnet) plugin: `rewrite edns0 subnet set 24 56`.

Filippo125:
Any update on this?

Dolores-123456:
mark

sindef:
Any update on this one? Would be fantastic to see in!

PR #7048, opened 2024-12-24
Canonicalize IPv4 addresses in acl

danog:

This pull requests canonicalizes IPv4 addresses (turning them into a 4-byte slice) before using them in the iptree, due to a bug in the `iptree` library which ignores IPv4 addresses if they're in a 16-byte slice (see https://github.com/infobloxopen/go-trees/pull/27 for more info).


SuperQ:
Would you mind adding some tests for this behavior?

danog:
Actually, looking into it some more, this specific bug is not triggered in the ACL codepaths as `net.ParseCIDR` does canonicalization (only for the IPNet, not for the IP).

This is still a nasty footgun, there should be at least some comments in place for future editors of acl (in case someone will want to manually create a subnet starting from a non-canonicalized net.IP...).

PR #7082, opened 2025-01-20
multicluster: add the multicluster plugin in-tree

### LINKED ISSUE #6921: Merging the coredns/multicluster plugin in tree
Merging the coredns/multicluster plugin (https://github.com/coredns/multicluster) in-tree/in the coredns main branch.

**Why is this needed**:
The CoreDNS multicluster plugin is the implementation of KEP-1645 (https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/1645-multi-cluster-services-api/README.md#dns) and from what is publicly available is used by
- AWS (https://github.com/aws/aws-cloud-map-mcs-controller-for-k8s but alpha though)
- Alibaba (https://www.alibabacloud.com/help/en/ack/manage-multi-cluster-services-by-using-the-cli?spm=a2c63.p38356.0.0.445d6a42eAVFn4#uicontrol-8rm-imv-mer)
- Cilium (https://docs.cilium.io/en/latest/network/clustermesh/mcsapi/#installing-coredns-multicluster which will be released in Cilium 1.17 in ~Jan/Feb of next year)
- In e2e tests of the mcs-api repo (https://github.com/kubernetes-sigs/mcs-api/blob/eff1ba8c3ab2d50a7faa0a02d18dc6ef9f58a21d/scripts/up.sh#L52)

Outside of that the Kubernetes SIG handling this KEP (SIG Multicluster) is looking at graduating to beta in the coming months: https://github.com/kubernetes-sigs/mcs-api/issues/48.

On my side I represent mostly the Cilium side of things and we would aim to somehow simplify the installation process of the coredns multicluster plugin which triggered this proposal/question. We could start by providing a docker image with the multicluster plugin compiled in/included but unlike AWS and Alibaba we run on many different types of Kubernetes installation and we don't control tightly everyone's setup so this comes with the risk that the Kubernetes cluster has installed a specific version of CoreDNS with some additional plugins. Including this in-tree would increase significantly the chance that the multicluster plugin would be compiled-in in the long term in the coredns image already installed in the cluster.

I would be happy to do all the work of getting the plugin in-tree and any potential improvements in order to do that. And as Cilium is aiming to be conformant and follow closely the KEP I would be happy to work on the maintenance of the plugin to keep it up to date with the KEP.

Let me know if getting the plugin in-tree looks reasonable to you, if yes I would start drafting some PR/checking more in-depth how to do this. Thanks!

MrFreezeex:
This adds the https://github.com/coredns/multicluster in-tree which is a DNS implementation of MCS-API/KEP-1645 (https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/1645-multi-cluster-services-api/README.md#dns). See #6921 for more details

The first commit includes all the multicluster code as is/unmodified, I was not able to do a git subtree (even squashed) as it seems not possible to add the signed-of to the subtree specific commits, the code shouldn't be changed at all in this commit. The following commits do the following:
- "integrate" the multicluster plugin in the repo (remove mentions to plugin external plugin installation, conform to coredns import style, ...)
- Two small fixes
- Add a logger adapter from the kubernetes plugin as I need to expose a function from the kubernetes plugin it's easier to coordinate this directly inside the coredns repo
- Add support for NS queries also by exploiting the existing kubernetes plugin code

AFAICT it currently still miss one feature that the kubernetes plugin already have: support for transfer/IXFR. I would like to tackle that once this plugin is merged since it would probably require a bit more code than the modification to the kubernetes/multicluster plugin done in this PR.

Fixes #6921

No


MrFreezeex:
This should be ready and the test failure should be fixed by #7083.

chrisohaver:
Thanks for the contribution and sorry for the delay in us reviewing this.

Would it be feasible to extend the kubernetes plugin to support multicluster via an option instead of making it a separate plugin?  This could reduce the amount of code duplication and subsequent consolidation refactoring. What do you think would be the drawbacks of that approach?

MrFreezeex:
> Thanks for the contribution and sorry for the delay in us reviewing this.

No worries, thanks a lot for picking up this :heart: 

> Would it be feasible to extend the kubernetes plugin to support multicluster via an option instead of making it a separate plugin? This could reduce the amount of code duplication and subsequent consolidation refactoring. What do you think would be the drawbacks of that approach?

I think it's certainly feasible, I don't know yet the full drawbacks I would need to try a bit to see how it goes. It would certainly help in reducing all the tests duplicated for sure and a bit more on the rest of the non test related code too.

One of my aim was to reduce the amount of code change to the current multicluster plugin but maybe you want to just review this from zero and in this case it won't really help.

So happy to try out to do this and if I see a major problem with that will report back here or just comes up with a new PR doing this and you will be able to judge which approach is best and we can iterate on it with reviews!

MrFreezeex:
@chrisohaver I created the PR #7266 that add this functionality directly into the kubernetes plugin, let me know which approach you prefer and feel free to comment on either PR! :pray:

PR #7083, opened 2025-01-20
test: fix cname with proxy test

MrFreezeex:
www.example.net is now behind akamai with various IP answered and a chain of CNAME. Let's replace www.example.net by one of the root server which answer a single IP and hopefully should remain this way.

no documentation changes
no


MrFreezeex:
For instance here is a PR where this test fails: https://github.com/coredns/coredns/pull/7081, and it also fails in the master branch as well.

SuperQ (approved):
Maybe longer term, we should setup `coredns.io` hostnames for these tests.

PR #7084, opened 2025-01-21
update to go1.23.5 (fix CVE-2024-45341, CVE-2024-45336)

zouyee:

go1.23.5 (released 2025-01-16) includes security fixes to the crypto/x509 and net/http packages, as well as bug fixes to the compiler, the runtime, and the net package. See the Go 1.23.5 milestone on our issue tracker for details;

https://github.com/golang/go/issues?q=milestone%3AGo1.23.5+label%3ACherryPickApproved full diff: golang/go@go1.23.4...go1.23.5
Hello gophers,

We have just released Go versions 1.23.5 and 1.22.11, minor point releases.

These minor releases include 2 security fixes following the security policy:

crypto/x509: usage of IPv6 zone IDs can bypass URI name constraints

A certificate with a URI which has a IPv6 address with a zone ID may incorrectly satisfy a URI name constraint that applies to the certificate chain.

Certificates containing URIs are not permitted in the web PKI, so this only affects users of private PKIs which make use of URIs.

Thanks to Juho Forsén of Mattermost for reporting this issue.

This is CVE-2024-45341 and Go issue https://go.dev/issue/71156.

net/http: sensitive headers incorrectly sent after cross-domain redirect

The HTTP client drops sensitive headers after following a cross-domain redirect. For example, a request to a.com/ containing an Authorization header which is redirected to b.com/ will not send that header to b.com.

In the event that the client received a subsequent same-domain redirect, however, the sensitive headers would be restored. For example, a chain of redirects from a.com/, to b.com/1, and finally to b.com/2 would incorrectly send the Authorization header to b.com/2.

Thanks to Kyle Seely for reporting this issue.

This is CVE-2024-45336 and Go issue https://go.dev/issue/70530.


zouyee:
```
--- FAIL: TestZoneExternalCNAMELookupWithProxy (0.03s)
    file_cname_proxy_test.go:75: Expected 2 RRs in answer section got 5
```

xref: https://github.com/coredns/coredns/pull/7083

SuperQ:
Fixed master branch.

zhangguanzhang:
need update to go1.23.6
```
CVE-2025-22866 │       │ 1.22.12, 1.23.6, 1.24.0-rc.3
```

chrisohaver:
Go version has since been updated to 1.24.1

PR #7105, opened 2025-01-29
dynamicforward: the new plugin implementation

Paramoshka:

This pull request introduces a new CoreDNS plugin, dynamicforward, designed for dynamic tracking of Kubernetes Services and seamless updates to the forward server list. The plugin monitors EndpointSlices associated with a specified Service, dynamically adjusting the DNS forwarding configuration as endpoints are added, removed, or updated.

The dynamicforward plugin  is particularly useful for organizing node-local-dns caching mechanism. In [standard scheme](https://github.com/kubernetes/kubernetes/blob/8294abc599696e0d1b5aa734afa7ae1e4f5059a0/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml#L100)  we use ClusterIP as upstream, so we are delegating responsibility for load balancing to CNI. In case of apiserver failure, CNI don't know which upstream endpoints are alive or not.
But coredns itself can handle upstream healthchecks and in case of apiserver failure it will still keep the endpoints list and will load balance the requests. 

There are some design imperfections:
* The plugin don't support the full list of  classic forward options. It uses the forward plugin for serving DNS under the hood, but it don't have public interface for configuring options.
* Controversial name of the plugin.

Added man.

No.

### DIFF
--- a/plugin/dynamicforward/dynamicforward.go
+++ b/plugin/dynamicforward/dynamicforward.go
@@ -43,7 +43,14 @@ func (df *DynamicForward) UpdateForwardServers(newServers []string, config Dynam
 	df.cond.L.Lock()
 
 	// Clear proxy list `Forwarder`
+	if df.forwarder != nil {
+		for _, oldProxy := range df.forwarder.List() {
+			oldProxy.Stop()
+		}
+	}
+
 	df.forwarder = forward.New()
+
 	// Fill up list servers
 	df.forwardTo = newServers
 
--- a/plugin/dynamicforward/dynamicforward.go
+++ b/plugin/dynamicforward/dynamicforward.go
@@ -1,0 +1,67 @@
+package dynamicforward
+
+import (
+	"context"
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/forward"
+	"github.com/coredns/coredns/plugin/pkg/proxy"
+	"github.com/coredns/coredns/plugin/pkg/transport"
+	"github.com/miekg/dns"
+	"log"
+	"sync"
+)
+
+// DynamicForward main struct of plugin
+type DynamicForward struct {
+	Next        plugin.Handler
+	Namespace   string
+	ServiceName string
+	forwardTo   []string
+	mu          sync.RWMutex
+	forwarder   *forward.Forward
+	options     *forward.Options
+	cond        *sync.Cond
+}
+
+func (df *DynamicForward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	//wait forwarder
+	df.cond.L.Lock()
+
+	if df.forwarder == nil {
+		df.cond.Wait()
+	}
+
+	df.cond.L.Unlock()
+
+	return df.forwarder.ServeDNS(ctx, w, r)
+}
+
+// UpdateForwardServers update list servers for forward requests
+func (df *DynamicForward) UpdateForwardServers(newServers []string, config DynamicForwardConfig) {
+	//df.mu.Lock()
+	//defer df.mu.Unlock()
+	df.cond.L.Lock()
+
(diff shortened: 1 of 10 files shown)


plugin/dynamicforward/dynamicforward.go
```diff
@@ -0,0 +1,62 @@
...
+	ServiceName string
+	forwardTo   []string
+	mu          sync.RWMutex
+	forwarder   *forward.Forward
+	options     *forward.Options
+}
+
+func (df *DynamicForward) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	//wait forwarder
+	df.mu.RLock()
```
apolovov: There isn't guarantee the ServeDNS will be called after df.forwarder will be initialized and the mutex won't help. Need another solution.

AndreyPavlovFlant (approved):

johnbelamaric:
I haven't really looked at the code, but a few organizational points:

1) The idea here is bypass kube-proxy load balancing and directly read the upstream service endpoint slices? I guess that reduces latency a little bit, does it really make that much difference?
1) A name like `kubeforward` is probably better - it relies on kube existing, it's not independent of kube.
1) Since it's really just for node-local-dns, why is this not part of https://github.com/kubernetes/dns/tree/master/cmd/node-cache ? I don't think it makes sense in upstream CoreDNS.

apolovov:
Thank you, @johnbelamaric, for your feedback. Let me answer your questions:

> 1. The idea here is bypass kube-proxy load balancing and directly read the upstream service endpoint slices? I guess that reduces latency a little bit, does it really make that much difference?

The main purpose of the plugin is not to reduce latency by eliminating an extra "hop," but to increase reliability.
Current scheme with the standard forward plugin in the event of an apiserver failure:
* The kube-proxy will not handle kube-dns pod outages and will not exclude them from load balancing.
* Consequently, coredns will continue sending requests to non-functioning upstream servers, leading to application issues.

Scheme with the new plugin:
* Coredns will keep a list of upstream servers at the time of the apiserver failure.
* It will perform its own `health_check`s and exclude any failed upstream servers.

> 2. A name like `kubeforward` is probably better - it relies on kube existing, it's not independent of kube.

Good name, it is better, will rename if there won't be better options.

> 3. Since it's really just for node-local-dns, why is this not part of https://github.com/kubernetes/dns/tree/master/cmd/node-cache ? I don't think it makes sense in upstream CoreDNS.

This tool uses coredns and it's Corefile under the hood, so here we are. Also, this tool is handy when you use some classic CNI with kube-proxy as Service implementation, but in clusters with eBPF-powered kube-proxy implementations (i.e. Cilium), it is overhead.

chrisohaver:
I think this should be an external plugin that node local dns can pull in.

Paramoshka:
Thank you @chrisohaver for the idea, I think this is a good option for our plugin to make it external. We will look at the documentation soon and redo the current PR. Is there no objection if we add an interface function `SetProxyOptions` to pass parameters to the `forward` plugin?
For the moment we accept the name `kubeforward` for our plugin, if there are other suggestions we will gladly consider them. Thanks for your comments!

Paramoshka:
Hello, @chrisohaver and @johnbelamaric,

Following the suggestions above, we have decided to move our dynamicforward - now under a new name [kubeforward](https://github.com/coredns/coredns.io/pull/323) - to the external plugins In addition, we identified the need for an interface for passing parameters to the forward plugin. We implemented a separate [PR](https://github.com/coredns/coredns/pull/7229) for this purpose.

Thank you for your recommendations and feedback!

thevilledev:
Hey @Paramoshka, shall we close the PR now that kubeforward is handled as an external plugin?

Paramoshka:
> shall we close the PR now that kubeforward is handled as an external plugin?

@thevilledev Yeah, I think we can close, thanks for the reminder.

Issue #7120, opened 2025-02-06
The Autopath plugin should recognize hostNetwork pods


fzu-huang:
**What would you like to be added**:
I think we should add a field to the pod object in the kubernetes plugin cache that describes whether it is hostNetwork or not. Like:

```
// Pod is a stripped down api.Pod with only the items we need for CoreDNS.
type Pod struct {
	// Don't add new fields to this struct without talking to the CoreDNS maintainers.
	Version   string
	PodIP     string
	Name      string
	Namespace string

	Hostnetwork bool       // need to add

	*Empty
}

```

In addition, kubernetse plug-in has a method to query pod based on IP index: podWithIP, here we should judge whether the obtained pod is hostNetwork, if it is, it returns nil .

```
// podWithIP returns the api.Pod for source IP. It returns nil if nothing can be found.
func (k *Kubernetes) podWithIP(ip string) *object.Pod {
	if k.podMode != podModeVerified {
		return nil
	}
	ps := k.APIConn.PodIndex(ip)
	if len(ps) == 0 {
		return nil
	}
         // need to add
	if ps[0].HostNetwork {
		return nil
	}
	return ps[0]
}
```

In this way, we can ensure that there is no miscalculation when "IP-based query pod" is used

**Why is this needed**:

There is currently an bug with the autopath plugin: It relies on the client IP to identify the pod, including identifying the namespace of the pod, which leads to the possibility that when the pod IP is quickly reused, the pod it identifies based on the IP is an old object that may not be in the same namespace as the intended pod.

This problem tends to arise in two scenarios:
1. After the pod is deleted, its IP is quickly reassigned to a new pod. The new pod is not in the same namespace as the old pod, and autopath does not quickly synchronize cache;
2. Use hostNetwork and dnsPolicy for ClusterFirstWithHostnetwork pod, IP is the node IP of this kind of pod, so there may be multiple pod using the same IP, When autopath executes k.asponn.podindex (ip), it always gets the first pod, so other pods will encounter dns resolution exceptions if they belong to other namespaces.

I think that by introducing the hostNetwork field, we can fix the incompatibility of the autopath plugin with the special pod in point 2 above.

PR #7122, opened 2025-02-06
do not autopath for hostnetwork pod

fzu-huang:
Improved availability of the autopath plugin
https://github.com/coredns/coredns/issues/7120


chrisohaver:
Thanks for the fix!  Please add a mention of this behavior in the Autopath section of the Kubernetes plugin readme. 
Something like...

## AutoPath

The *kubernetes* plugin can be used in conjunction with the *autopath* plugin.  Using this
feature enables server-side domain search path completion in Kubernetes clusters.  Note: 
* The `pods` option must be set to `verified` for this to function properly
* The remote IP address in the DNS packet received by CoreDNS must match the IP address of the Pod that sent the request
* The pod that sent the request must not be using host networking. Since host network pods don't have their own IP address, such pods cannot be uniquely identified.

fzu-huang:
> Thanks for the fix! Please add a mention of this behavior in the Autopath section of the Kubernetes plugin readme. Something like...
> 
> ## AutoPath
> The _kubernetes_ plugin can be used in conjunction with the _autopath_ plugin. Using this feature enables server-side domain search path completion in Kubernetes clusters. Note:
> 
> * The `pods` option must be set to `verified` for this to function properly
> * The remote IP address in the DNS packet received by CoreDNS must match the IP address of the Pod that sent the request
> * The pod that sent the request must not be using host networking. Since host network pods don't have their own IP address, such pods cannot be uniquely identified.

Sorry for the delay in continuing the progress of this PR. I have re-based the latest code and added the sign-off information.

@chrisohaver

plugin/autopath/README.md
thevilledev, line 1: As already mentioned by @chrisohaver, I too would put this to the Kubernetes plugin README. It already covers `pods verified` and it would be a more natural place for source IP requirements. See https://coredns.io/plugins/kubernetes/#autopath

fzu-huang:
The readme files for the autopath plugin and the Kubernetes plugin both provide relevant explanations. Take a look. @chrisohaver @thevilledev

fzu-huang:
This PR has been blocked for a long time. Could you please review it and move it forward? @chrisohaver @yongtang

PR #7129, opened 2025-02-09
chore: enable early-return and superfluous-else from revive

mmorel-35:

Use revive rules
* [early-return](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#early-return) : In Go it is idiomatic to minimize nesting statements, a typical example is to avoid if-then-else constructions.
* [superfluous-else](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#superfluous-else): To improve the readability of code, it is recommended to reduce the indentation as much as possible. This rule highlights redundant else-blocks that can be eliminated from the code.

None

N/A

No

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,4 +1,9 @@
 version: "2"
+
+issues:
+  max-issues-per-linter: 0
+  max-same-issues: 0
+
 linters:
   default: none
   enable:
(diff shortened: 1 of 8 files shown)


SuperQ (approved):

thevilledev (approved):
LGTM, thanks!

PR #7142, opened 2025-02-13
fix: Ensure DNS query name reset in plugin.NS error path

hansedong:

Currently, the implementation of the NS function in `plugin/backend_lookup.go` of CoreDNS is as follows:

```go
func NS(ctx context.Context, b ServiceBackend, zone string, state request.Request, opt Options) (records, extra []dns.RR, err error) {
    // NS record for this zone live in a special place, ns.dns.<zone>. Fake our lookup.
    // only a tad bit fishy...
    old := state.QName()

    state.Clear()
    state.Req.Question[0].Name = dnsutil.Join("ns.dns.", zone)
    services, err := b.Services(ctx, state, false, opt)
    if err != nil {
        return nil, nil, err
    }
    // ... and reset

    seen := map[string]bool{}

    for _, serv := range services {
        what, ip := serv.HostType()
        switch what {
        case dns.TypeCNAME:
            return nil, nil, fmt.Errorf("NS record must be an IP address: %s", serv.Host)

        case dns.TypeA, dns.TypeAAAA:
            serv.Host = msg.Domain(serv.Key)
            ns := serv.NewNS(state.QName())
            extra = append(extra, newAddress(serv, ns.Ns, ip, what))
            if _, ok := seen[ns.Ns]; ok {
                continue
            }
            seen[ns.Ns] = true
            records = append(records, ns)
        }
    }
    return records, extra, nil
}
```

In `plugin.NS`, when `Services()` returns an error, the function returns directly without resetting the query name back to original. This leaves the query name set to the internal lookup value ("ns.dns.<zone>") in error cases.

So, in this case, the DNS domain name has actually been changed, not the original value. And when the caller of `plugin.NS` (possibly a plugin) needs to pass the request to the next plugin for processing, the domain name processed by the next plugin is not actually the origin domain name.

This PR ensures the query name is always reset to original value before error handling, making the error path consistent with the success path.

Before:

```go
if err != nil {
    return nil, nil, err  // Query name remains as "ns.dns.<zone>"
}
// ... and reset
```

After:

```go
// Always reset the query name to original first
if err != nil {
    return nil, nil, err  // Query name is properly reset
}
```

None

No

No

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -427,20 +427,20 @@ func PTR(ctx context.Context, b ServiceBackend, zone string, state request.Reque
 	return records, nil
 }
 
-// NS returns NS records from  the backend
+// NS returns NS records from the backend
 func NS(ctx context.Context, b ServiceBackend, zone string, state request.Request, opt Options) (records, extra []dns.RR, err error) {
-	// NS record for this zone live in a special place, ns.dns.<zone>. Fake our lookup.
-	// only a tad bit fishy...
+	// NS record for this zone lives in a special place, ns.dns.<zone>. Fake our lookup.
+	// Only a tad bit fishy...
 	old := state.QName()
 
 	state.Clear()
 	state.Req.Question[0].Name = dnsutil.Join("ns.dns.", zone)
 	services, err := b.Services(ctx, state, false, opt)
+	// reset the query name to the original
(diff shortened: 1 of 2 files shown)


hansedong:
Guys, would you please kindly take a look into this PR?

yongtang (approved):

Issue #7109, opened 2025-01-31
Enable TLS for coreDNS metrics endpoint


smoshiur1237:
**What would you like to be added**:
Request to add TLS support for coreDNS metrics endpoint. To enable TLS for CoreDNS metrics endpoint,  need to configure CoreDNS to expose metrics over HTTPS.

**Proposed Enhancement**:

- TLS support in coredns metrics server.
- Provide configuration options to enable/disable and have variable to store the cert and key for TLS
- Create a Secret for the TLS Certificate and Key
- Modify the CoreDNS ConfigMap to add TLS option in the configuration. For example
 ```
prometheus :9153 {
        tls /etc/coredns/coredns.crt /etc/coredns/coredns.key
    }

```
-  Mount the Secret into the CoreDNS Pod
-  Update Prometheus Scraping Configuration, make sure to update the Prometheus scraping configuration to use HTTPS

**Why is this needed**:
TLS support is needed to ensure and enhance security of metrics data, so that it is securely transmitted and compatible with best practices. Currently coreDNS have metrics endpoint with a port number which exposes metrics with plaintext and doesn't support secure transmission of metrics. So the coredns configmap has the following configuration which is not supproting TLS:

 ```
 }
        errors
        ready
        prometheus :9153
```
This would need the support for TLS to ensure the security, configuration would have the following addition for TLS
```
  }
        errors
        ready
        prometheus :9153 {
        tls /etc/coredns/coredns.crt /etc/coredns/coredns.key
    }
```

smoshiur1237:
/cc @johnbelamaric 
I would appreciate, if you have any suggestion on how to support TLS for metrics endpoint.  Also please share how to take this enhancement request forward.

TomyLobo:
Do you have a requirement that requires you to have the traffic between CoreDNS and your ingress controller encrypted?
If yes, get a CNI that does that for you, like [Calico](https://docs.tigera.io/calico/latest/network-policy/encrypt-cluster-pod-traffic).
If not, use an ingress with TLS enabled.

Neither option requires this feature.
Both are more secure.

tuminoid:
Issue with the CNI encryption is that it is not end-to-end encryption. It leaves host-to-pod traffic unencrypted, as [Calico docs state clearly](https://docs.tigera.io/calico/latest/network-policy/encrypt-cluster-pod-traffic).  Ingress has little relevance here as CoreDNS metrics can be exposed only internally to Prometheus and traffic does not go thru Ingress.

[...]

PR #7144, opened 2025-02-17
Support optional TLS in metrics

mquhuy:

This PR adds optional TLS support for the CoreDNS metrics endpoint to enhance security by encrypting metric data communications. Currently, the metrics endpoint operates over unencrypted HTTP, which could expose sensitive operational data in transit. This PR implements:

- A configuration option in Corefile to enable TLS for the metrics endpoint
- Certificate and key configuration for securing metrics traffic
- Backward compatibility by making TLS optional for metrics
- Proper error handling for metrics endpoint TLS-related issues
- Appropriate logging for metrics TLS events

This enhancement is important for environments where secure monitoring is required, especially when metrics might contain sensitive operational data or when compliance regulations mandate encryption for all monitoring traffic.

#7109 

1. Update to the metrics plugin documentation to describe the new TLS configuration options
2. Addition of configuration examples showing how to enable and configure TLS for the metrics endpoint
3. Updates to monitoring and observability documentation sections
4. Addition of details on:
   - How to configure certificates for the metrics endpoint
   - How to connect monitoring tools to the TLS-enabled metrics endpoint
   - Compatible Prometheus configurations for scraping secure metrics
   - Troubleshooting secure metrics endpoint connectivity

No, this PR does not introduce any backward incompatible changes or deprecations. The TLS support for the metrics endpoint is implemented as an optional feature that is disabled by default.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,6 +28,7 @@ require (
 	github.com/prometheus/client_golang v1.20.5
 	github.com/prometheus/client_model v0.6.1
 	github.com/prometheus/common v0.62.0
+	github.com/prometheus/exporter-toolkit v0.14.0
 	github.com/quic-go/quic-go v0.49.0
 	go.etcd.io/etcd/api/v3 v3.5.18
 	go.etcd.io/etcd/client/v3 v3.5.18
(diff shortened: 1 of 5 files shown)


JanMkl:
@chrisohaver, @miekg or @johnbelamaric would you have time to review this? Thanks!

chrisohaver:
Thanks for contributing!  I expect this is already possible with a side car or separate process to do tls decryption, though in-app support would be more convenient.

Not sure if we need to do the tls in the app. 

What do other maintainers think?

JanMkl:
@chrisohaver Thanks for the fast response. 

In my opinion you bring the main point up in your comment and that is that in-app configuration is more convenient and less error-prone from the security standpoint. This approach would be accepted in environments where requirement is to have even internal interfaces encrypted (this case by not having an interface between the app and sidecar). Lastly but not least reducing possible attack surface which could have vulnerabilities.

Aiming for less complex and less moving parts would be from security standpoint better solution in the long run.

SuperQ:
We have a standard webserver config wrapper for this kind of thing in the Prometheus [exporter-toolkit](https://pkg.go.dev/github.com/prometheus/exporter-toolkit@v0.14.0/web) for exactly the use case of TLS and auth for metrics endpoints.

mquhuy:
> Thanks for contributing! I expect this is already possible with a side car or separate process to do tls decryption, though in-app support would be more convenient.

Thank you @chrisohaver  very much for the comment! As you and @JanMkl mentioned, built-in tls support is more convenient and less error-prone, and since it's an optional feature, it shouldn't break anything.

> We have a standard webserver config wrapper for this kind of thing in the Prometheus [exporter-toolkit](https://pkg.go.dev/github.com/prometheus/exporter-toolkit@v0.14.0/web) for exactly the use case of TLS and auth for metrics endpoints.

Thank you @SuperQ  a lot! I have not known of this package, sounds like a good improvement from what this PR currently does! I'll make an update soon.

mquhuy:
Hi @SuperQ, I've made a new commit changing the TLS server creation from using normal net/http package to the `exporter-toolkit`. Basically everything is fine, but a small issue I encountered is that as the exporter-toolkit expects usage of [slog.Logger](https://pkg.go.dev/log/slog#Logger) as the logger, I had to make an adapter to adapt coredns normal logger to the slog.Logger interface, which lengthened the solution quite a bit. Do you think it should be fine like that, or would you have any suggestion for me on how to better-handle this situation.

Otherwise, could you have a look at the current solution and give me a review? Thank you very much in advance!

SuperQ:
Thanks, we should probably migrate coredns over to `log/slog` eventually.

mquhuy:
Hello. For personal reasons I decided to not continue with this PR, and hereby close it. If anyone's interested in making this contribution, please feel free to do so!

Issue #7125, opened 2025-02-07
Not getting response back from endpoints with CNAME chains/lookups greater than 7.


uuoonim:
### Discussed in https://github.com/coredns/coredns/discussions/7123

<div type='discussions-op-text'>

<sup>Originally posted by **uuoonim** February  6, 2025</sup>
It looks like A, AAAA, TXT record look ups have a max # of lookup of 7 to limit CNAME chaining in backend_lookup.go. We have a use case where we have more than 7 lookups/CNAME chain. Many endpoints from Microsoft openai has chains greater than 7 now, I'm wondering if we can increase this limit to 10. One example would be an endpoint thats on their website: `docs-test-001.openai.azure.com` with 8 lookups to get to the A record. Because of this, we don't even see a response back, instead we see messages like `bad address`.

Any oppositions for increasing this arbitrary limit of 7 to something greater like 10?

PR #7153, opened 2025-02-21
Increase CNAME lookup limit from 7 to 10

uuoonim:
Found a use case where we need to do more than 7 CNAME lookups. For example, hitting an OpenAI endpoint from azure. One example of the endpoint they have on their website is docs-test-001.openai.azure.com, which exceeds 7 lookups, cannot get to the A record.

https://github.com/coredns/coredns/discussions/7123
https://github.com/coredns/coredns/issues/7125

This isn't documented anywhere already. Not sure if it needs to be

No

### DIFF
--- a/plugin/backend_lookup.go
+++ b/plugin/backend_lookup.go
@@ -13,6 +13,8 @@ import (
 	"github.com/miekg/dns"
 )
 
+const maxCnameChainLength = 10
+
 // A returns A records from Backend or an error.
 func A(ctx context.Context, b ServiceBackend, zone string, state request.Request, previousRecords []dns.RR, opt Options) (records []dns.RR, truncated bool, err error) {
 	services, err := checkForApex(ctx, b, zone, state, opt)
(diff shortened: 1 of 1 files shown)


SuperQ:
So digging around, I found a thread https://github.com/NLnetLabs/unbound/issues/438 where a similar limit was increased to 11.

There is https://lists.isc.org/pipermail/bind-users/2010-July/080543.html that claims BIND will do 16.

Perhaps it's worth making this a configuration option at the same time we bump it?

uuoonim:
> Perhaps it's worth making this a configuration option at the same time we bump it?

Thought about making it a config option too but wasn't sure where to begin. Ill give it a shot though

Question: would this be a proper use case to create a plugin? or can this be done on existing code?

uuoonim:
Can we get this merged for now? I can work on getting it a configurable value after, not sure how long it'll take for me.

SuperQ (approved):

SuperQ:
Yes, I'm OK with this for now.

Maybe at a minimum, change the value to a package level `const` so it's consistent. We can refactor it to a config var later.

uuoonim:
> Yes, I'm OK with this for now.
> 
> Maybe at a minimum, change the value to a package level `const` so it's consistent. We can refactor it to a config var later.

Sorry for the delay. done.

SuperQ (approved):

johnbelamaric:
@uuoonim you need to fix the CI failures before we can merge this

PR #7178, opened 2025-03-06
Fix autopath for host network pods

### LINKED ISSUE #4299: Enables server-side domain search path completion for pods running in the host network

Adrian-St:
This is a follow up PR for: https://github.com/coredns/coredns/pull/6776
@chrisohaver, let me know if you're happy with the changes.

1. Why is this pull request needed and what does it do?
This PR fixes the issue that the autopath plugin isn't working for pods running in the host network.
The reason this is marked as a fix and not a feature is because the current behaviour is causing issues in combination with caching and has inconsistencies.

Currently the AutoPath function fetches the pods by request IP to check where the request is coming from. For pods running in the host network this would be the Node IP. So if you have multiple pods running in host networking it would return multiple pods but only the first one is being used. This means that autopath might work for a pod running on one node because it happens to be the first one in the list, but it doesn't work on another node. For normal pods in the same namespace the same DNS query would also work with autopath.

In itself this is not a big issue, but in combination with caching this can cause lookup failures because A records and AAAA records are cached differently and in our resolvers if one of them returns an answer and the other one doesn't then the whole query fails with NXDOMAIN and it doesn't try the rest of the search path.

This implementation takes all host network pods on the node, matches the query against them and takes the namespace from the pod that matches. Of course any pod could request a DNS name in any namespace, but I don't really see why that is an issue, since the client would have to actively try to break DNS lookup.

The only potential issue I see is if the first query is lost/answered by another pod and the second query is matched by accident, it could result in a wrong "base".
Example: If there is a namespace called "other" and a pod in this namespace does a DNS request for "example.other", the first query would be "example.other.other.svc.cluster.local" and the second "example.other.svc.cluster.local". Now if the first query is lost, the second query would use "example" as a base and append the items from the search path, which could result on wrong DNS responses.
But this was already an issue before, this change just makes it more predicable (happening for every host network pod).

2. Which issues (if any) are related?
This would fix: https://github.com/coredns/coredns/issues/4299 and the bug described in https://github.com/coredns/coredns/issues/7120

3. Which documentation changes (if any) need to be made?
None that I found, currently it's undocumented that host network pods don't work with autopath as far as I can tell.

4. Does this introduce a backward incompatible change or deprecation?
No, unless clients rely on the inconsistent behaviour of autopath, which would be strange.


chrisohaver:
Sorry for the delay in responding... couple of things...

* I dont understand how using the client namespace helps solve the problem, can you explain by example?
* This solution appears to require metadata to be enabled. This would need to be mentioned in docs.  What's the behavior if metadata is not enabled?

Adrian-St:
@chrisohaver 
Let's say that a host network pod from the `default` namespace makes a DNS request to exampe.com.
The first DNS query would be:
```A example.com.default.svc.cluster.local.```
and the IP of the request would be the node IP.

In the old behaviour of the `podWithIP` function would query the pods by IP, get all host network pods on this node and return the first one. This might be the correct pod in the `default` namespace, but it also could be one in another namespace, in which case it would not do autopath, causing the bug because another CoreDNS instance might do autopath.

With my update the `podWithIP` function turns into `podsWithIP` and the logic will cross-check the namespace used in the query against all the pods that run on the node as host network pods. If ANY of them match it will do autopath, which prevents the bug from happening since there's no "luck of draw" involved.

The metadata update is just to fix the metadata behaviour, since currently it will attach the metadata of the first host network pod, which might be wrong. With the update the metadata could still be wrong if there's two host network pods in the same namespace, but it greatly improves the behaviour. 

Hope that clears things up! Btw, the CI pipeline failed because of an issue unrelated to my change I believe, it would be great if you could retrigger the pipeline.

Issue #7177, opened 2025-03-04
PR #6898 unexpectedly breaks creation of PTR records


nwohlgemuth:
**What happened**:
I have a single headless service associated to a deployment. Each pod in the deployment has two containers. One container is responsible for serving incoming gRPC requests (some compute heavy task). The other container is responsible for communicating with a custom load balancer service we have. On startup, it first does a reverse DNS lookup to get the DNS name of the pod. This DNS name is given to the load balancer service so whichever client needs to connect to it, does so through the DNS name of the specific pod. This way, we can precisely control which clients go to which service pods.

**What you expected to happen**:
That I can do a reverse DNS lookup of a pod and get a DNS record. In other words, if a pod has IP address 1.2.3.4, I should be able to run the following command and see the following output:

getent hosts 1.2.3.4
1.2.3.4   1-2-3-4.my-service.my-namespace.svc.cluster.local

However, as a result of https://github.com/coredns/coredns/pull/6898, I no longer get the DNS record returned when running the command. If I set the hostname to some value 'foo', I will get a DNS record returned. However, the DNS record is now the same for all pods as it is based on that value 'foo'. This completely breaks my load balancing service as I can no longer route requests to a specific pod. What I need, is a way to get back to the behavior before PR #6898 was merged. I do not have multiple services, only one and I would like the PTR records created as they were before. It seems to me like the change should have been scoped to the special case when you have pods associated with multiple services.

**How to reproduce it (as minimally and precisely as possible)**:
Create a headless service associated to a deployment. Run getent hosts on the IP address for one of the pods. Before PR #6898, you will get a DNS record returned. After, you will not.

**Anything else we need to know?**:

**Environment**:
The environment is not relevant as this issue is a direct result of the change in behavior as caused by PR #6898

aojea:
cross posting from the original kubernetes issue https://github.com/kubernetes/kubernetes/issues/127716#issuecomment-2700445224

Don't you get the pod name using the reverse lookup?

Michael-Sinz:
> cross posting from the original kubernetes issue [kubernetes/kubernetes#127716 (comment)](https://github.com/kubernetes/kubernetes/issues/127716#issuecomment-2700445224)
> 
> Don't you get the pod name using the reverse lookup?

No - without setting hostname and subdomain and this new coredns, we now get nothing.  The proper change could have checked if hostname and subdomain are not set to continue with prior behavior and if set use the new behavior.

[...]

PR #7180, opened 2025-03-06
Fix Kubernetes regression of not returning PTR records

adrianmoisey:
Only ever return a single PTR record. If a PTR record has a hostname, prefer it.

Fix for https://github.com/coredns/coredns/issues/7177

Fix for https://github.com/coredns/coredns/issues/7177

# Test cases

## StatefulSet with multiple Services:

(This is the original scenario, starting in https://github.com/kubernetes/kubernetes/issues/124207 leading to https://github.com/coredns/coredns/pull/6898)

<details>
<summary>Testing YAML</summary>

```
---
apiVersion: v1
kind: Service
metadata:
  name: headless-svc-dummy1
  namespace: test
spec:
  type: ClusterIP
  clusterIP: None
  publishNotReadyAddresses: true
  selector:
    app: dummy
---
apiVersion: v1
kind: Service
metadata:
  name: headless-svc-dummy2
  namespace: test
spec:
  type: ClusterIP
  clusterIP: None
  publishNotReadyAddresses: true
  selector:
    app: dummy
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: dummy
  namespace: test
  labels:
    app: dummy
spec:
  replicas: 2
  podManagementPolicy: Parallel
  selector:
    matchLabels:
      app: dummy
  serviceName: headless-svc-dummy1
  template:
    metadata:
      labels:
        app: dummy
    spec:
      containers:
        - name: alpine
          image: alpine:latest
          command: ["sleep", "infinity"]
          resources:
            requests:
              cpu: 20m
              memory: 1Mi
```
</details>

**CoreDNS 1.11.3:**
```
# dig +short -x 10.244.0.16
dummy-0.headless-svc-dummy1.test.svc.cluster.local.
10-244-0-16.headless-svc-dummy2.test.svc.cluster.local.
```
Two records returned, incorrect behaviour ❌ 

**CoreDNS 1.11.4:**
```
# dig +short -x 10.244.0.16
dummy-0.headless-svc-dummy1.test.svc.cluster.local.
```
A single hostname returned, correct behaviour ✅ 

**CoreDNS patched:**
```
# dig +short -x 10.244.0.16
dummy-0.headless-svc-dummy1.test.svc.cluster.local.
```
A single hostname returned, correct behaviour maintained ✅ 

## Deployment with single Service:

(The regression...)

<details>
<summary>Testing YAML</summary>

```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: dummy-deployment
  namespace: test
  labels:
    app: dummy-deployment
spec:
  replicas: 2
  selector:
    matchLabels:
      app: dummy-deployment
  template:
    metadata:
      labels:
        app: dummy-deployment
    spec:
      containers:
      - name: dummy-container
        image: nginx:latest
        ports:
        - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: headless-svc-deployment
  namespace: test
spec:
  type: ClusterIP
  clusterIP: None
  publishNotReadyAddresses: true
  selector:
    app: dummy-deployment
```
</details>

**CoreDNS 1.11.3:**
```
# dig +short -x 10.244.0.24
10-244-0-24.headless-svc-deployment.test.svc.cluster.local.
```
Single records returned, correct behaviour ✅

**CoreDNS 1.11.4:**
```
# dig +short -x 10.244.0.24
#
```
No record returned, incorrect behaviour ❌ 

**CoreDNS patched:**
```
# dig +short -x 10.244.0.24
10-244-0-24.headless-svc-deployment.test.svc.cluster.local.
```
A single hostname returned, correct behaviour restored ✅

### DIFF
--- a/plugin/kubernetes/reverse.go
+++ b/plugin/kubernetes/reverse.go
@@ -49,9 +49,17 @@ func (k *Kubernetes) serviceRecordForIP(ip, name string) []msg.Service {
 				// Kubernetes more or less keeps this to one canonical service/endpoint per IP, but in the odd event there
 				// are multiple endpoints for the same IP with hostname set, return them all rather than selecting one
 				// arbitrarily.
-				if addr.IP == ip && addr.Hostname != "" {
-					domain := strings.Join([]string{addr.Hostname, ep.Index, Svc, k.primaryZone()}, ".")
-					svcs = append(svcs, msg.Service{Host: domain, TTL: k.ttl})
+				if addr.IP == ip {
+					var domain string
+					// If the pod specifies a hostname (e.g., in a StatefulSet), use it for the PTR record.
+					if addr.Hostname != "" {
+						domain = strings.Join([]string{addr.Hostname, ep.Index, Svc, k.primaryZone()}, ".")
+						svcs = []msg.Service{{Host: domain, TTL: k.ttl}}
+						return svcs
+					} else {
+						domain = strings.Join([]string{endpointHostname(addr, k.endpointNameMode), ep.Index, Svc, k.primaryZone()}, ".")
+						svcs = []msg.Service{{Host: domain, TTL: k.ttl}}
+					}
 				}
 			}
 		}
--- a/plugin/kubernetes/reverse_test.go
+++ b/plugin/kubernetes/reverse_test.go
@@ -121,6 +121,52 @@ func (APIConnReverseTest) EpIndexReverse(ip string) []*object.Endpoints {
 		Namespace: "testns",
 		Index:     object.EndpointsKey("svc2", "testns"),
 	}
+	ep3s1 := object.Endpoints{
+		Subsets: []object.EndpointSubset{
+			{
+				Addresses: []object.EndpointAddress{
+					{IP: "10.0.0.98", Hostname: ""}, // has no hostname but we expect ep3s2 to take priority
+				},
+				Ports: []object.EndpointPort{
+					{Port: 80, Protocol: "tcp", Name: "http"},
+				},
+			},
+		},
+		Name:      "svc3-slice1",
+		Namespace: "testns",
+		Index:     object.EndpointsKey("svc3s1", "testns"),
+	}
+	ep3s2 := object.Endpoints{
+		Subsets: []object.EndpointSubset{
+			{
+				Addresses: []object.EndpointAddress{
+					{IP: "10.0.0.98", Hostname: "ep3s2"}, // has hostname and takes priority over ep3s1
+				},
+				Ports: []object.EndpointPort{
+					{Port: 80, Protocol: "tcp", Name: "http"},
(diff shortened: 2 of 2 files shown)


johnbelamaric:
Umm...looks like you and @jackfrancis both pushed PRs. This one sticks to "one PTR per IP", whereas #7181 will send multiple PTRs for different services, but not multiple PTRs for the same service (one without hostname, and one with), which is what #6898 was fixing, I think. Did I get that right?

plugin/kubernetes/reverse.go
```diff
@@ -49,9 +49,17 @@ func (k *Kubernetes) serviceRecordForIP(ip, name string) []msg.Service {
...
+				if addr.IP == ip {
+					var domain string
+					// If the pod specifies a hostname (e.g., in a StatefulSet), use it for the PTR record.
+					if addr.Hostname != "" {
+						domain = strings.Join([]string{addr.Hostname, ep.Index, Svc, k.primaryZone()}, ".")
+						svcs = []msg.Service{{Host: domain, TTL: k.ttl}}
+						return svcs
+					} else {
+						domain = strings.Join([]string{endpointHostname(addr, k.endpointNameMode), ep.Index, Svc, k.primaryZone()}, ".")
+						svcs = []msg.Service{{Host: domain, TTL: k.ttl}}
```
Michael-Sinz, line 61: This line and the one in the positive case are the same and could be placed outside of the if/else block.

Michael-Sinz (approved):
Thank you - I think this at least restores the DNS records for common stateless sets.

There is still the behavior difference with a hostname but that is the actual intention of the original change so that seems good.

jackfrancis:
> Umm...looks like you and @jackfrancis both pushed PRs. This one sticks to "one PTR per IP", whereas #7181 will send multiple PTRs for different services, but not multiple PTRs for the same service (one without hostname, and one with), which is what #6898 was fixing, I think. Did I get that right?

This is what you get when two people are heads down and not aware of the other. :)

@johnbelamaric I'm not sure that https://github.com/coredns/coredns/pull/6898 was aiming to fix returning multiple PTR records. From the PR description:

> If for some reason more than one have hostname defined, we will return both rather than select one arbitrarily - but IIUC, this would only happen if something is awry in k8s.

My PR was designed explicitly to build upon https://github.com/coredns/coredns/pull/6898 (I am assuming that the implementation of https://github.com/coredns/coredns/pull/6898, including multiple PTR return values, was intentional).

This PR seems to want to change things further so that only one PTR record is ever returned. No strong opinion about which approach should be preferred. Both seem to restore the previous behavior of returning a PTR record for pod IPs that do not have a hostname set.

adrianmoisey:
Replaced by https://github.com/coredns/coredns/pull/7194

PR #7181, opened 2025-03-06
return PTR record if pod hostname is not set

### LINKED ISSUE #7177: PR #6898 unexpectedly breaks creation of PTR records

jackfrancis:

This PR restores the previous behavior of returning a well-formed PTR record for DNS records that correlate with pods without a hostname. It is compatible with the changes in #6898 whose intention (among other things) was to improve the deterministic response of PTR records for pod IPs sharing a common hostname.

Fixes #7177

### DIFF
--- a/plugin/kubernetes/reverse.go
+++ b/plugin/kubernetes/reverse.go
@@ -37,24 +37,37 @@ func (k *Kubernetes) serviceRecordForIP(ip, name string) []msg.Service {
 		return []msg.Service{{Host: domain, TTL: k.ttl}}
 	}
 	// If no cluster ips match, search endpoints
-	var svcs []msg.Service
+	// Keep a discrete accounting of services with Hostnames set and those without
+	// So that we don't return a mixed set of service records with and without Hostnames set
+	var svcsHostname, svcsIPs []msg.Service
 	for _, ep := range k.APIConn.EpIndexReverse(ip) {
 		if len(k.Namespaces) > 0 && !k.namespaceExposed(ep.Namespace) {
 			continue
 		}
 		for _, eps := range ep.Subsets {
 			for _, addr := range eps.Addresses {
-				// The endpoint's Hostname will be set if this endpoint is supposed to generate a PTR.
-				// So only return reverse records that match the IP AND have a non-empty hostname.
-				// Kubernetes more or less keeps this to one canonical service/endpoint per IP, but in the odd event there
-				// are multiple endpoints for the same IP with hostname set, return them all rather than selecting one
-				// arbitrarily.
-				if addr.IP == ip && addr.Hostname != "" {
-					domain := strings.Join([]string{addr.Hostname, ep.Index, Svc, k.primaryZone()}, ".")
-					svcs = append(svcs, msg.Service{Host: domain, TTL: k.ttl})
+				var domain string
+				if addr.IP == ip {
+					if addr.Hostname != "" {
+						// If the endpoint's Hostname is set, compose a compliant PTR record composed from that hostname
+						// Kubernetes more or less keeps this to one canonical service/endpoint per IP, but in the odd event there
+						// are multiple endpoints for the same IP with hostname set, return them all rather than selecting one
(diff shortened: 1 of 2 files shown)


plugin/kubernetes/reverse.go
```diff
@@ -37,24 +37,37 @@ func (k *Kubernetes) serviceRecordForIP(ip, name string) []msg.Service {
...
-				// arbitrarily.
-				if addr.IP == ip && addr.Hostname != "" {
-					domain := strings.Join([]string{addr.Hostname, ep.Index, Svc, k.primaryZone()}, ".")
-					svcs = append(svcs, msg.Service{Host: domain, TTL: k.ttl})
+				var domain string
+				if addr.IP == ip {
+					if addr.Hostname != "" {
+						// If the endpoint's Hostname is set, compose a compliant PTR record composed from that hostname
+						// Kubernetes more or less keeps this to one canonical service/endpoint per IP, but in the odd event there
+						// are multiple endpoints for the same IP with hostname set, return them all rather than selecting one
```
johnbelamaric, line 54: I don't see a test for this behavior. The two-service selection test returns a single PTR.

Also why should we only return multiple in the case where hostname is set? I am confused as to what we actually *want* right now.

johnbelamaric:
Can we first decide on #7177 the behavior we want, then we can decide on the two competing PRs #7181 and #7180 ...

plugin/kubernetes/reverse.go
jackfrancis: This is not (meant to be) net new behavior, it is a reformulation of the existing behavior. I'm relying on the existing UT to validate that this change doesn't yield any new behaviors.

Also, to be clear, this change returns multiple records in both scenarios (hostname or no hostname). It just doesn't return a mixed set (multiple records, some of which have hostnames, some which do not). That preserves the post-6898 behaviors.

Taking a step back I agree that returning multiple records is weird, but that is what the current code at HEAD currently does. Here's a UT case to prove it (to be clear I don't think this is good! just trying not to break existing behaviors):

https://github.com/coredns/coredns/pull/7182

jackfrancis: Added more UT cases to this PR as well to show more permutations

houyuwushang:
@jackfrancis Thanks for investigating this and adding the test cases. The regression in #7177 was resolved by #7194 and released in v1.12.1. Current master restores PTR records for endpoints without an explicit hostname while preserving the existing multiple-PTR behavior.

Closing this PR as superseded. Additional regression coverage can be handled separately in a test-only PR.

PR #7194, opened 2025-03-10
Revert "only create PTR records for endpoints with hostname defined"

adrianmoisey:
Reverts coredns/coredns#6898

See discussion in https://github.com/coredns/coredns/issues/7177

### DIFF
--- a/plugin/kubernetes/reverse.go
+++ b/plugin/kubernetes/reverse.go
@@ -44,13 +44,8 @@ func (k *Kubernetes) serviceRecordForIP(ip, name string) []msg.Service {
 		}
 		for _, eps := range ep.Subsets {
 			for _, addr := range eps.Addresses {
-				// The endpoint's Hostname will be set if this endpoint is supposed to generate a PTR.
-				// So only return reverse records that match the IP AND have a non-empty hostname.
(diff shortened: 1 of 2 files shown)


johnbelamaric:
@chrisohaver ptal

I still think we should answer [these questions](https://github.com/coredns/coredns/issues/7177#issuecomment-2704941640), and probably update the [DNS spec](https://github.com/kubernetes/dns/blob/master/docs/specification.md), and implement to match. Even if the answer to the questions are "do what we used to do", we should document them clearly.

chrisohaver:
> @chrisohaver ptal
> 
> I still think we should answer [these questions](https://github.com/coredns/coredns/issues/7177#issuecomment-2704941640), and probably update the [DNS spec](https://github.com/kubernetes/dns/blob/master/docs/specification.md), and implement to match. Even if the answer to the questions are "do what we used to do", we should document them clearly.

I think we should revert.

chrisohaver:
IMO, “what we used to do” made sense, and tbh despite having executed the fix, I don’t fully understand the reasoning behind the hostname-only fix and why it was needed. 

(That’s not to say I don’t think there _was_ a good reason for it - I’m just admitting that I executed the fix without fully understanding reasoning)

chrisohaver (approved):

Issue #7196, opened 2025-03-11
Ability to optionally configure the Name and Type via `request_duration_seconds` metrics


harshanarayana:
**What would you like to be added**:
Current `forward` plugin of CoreDNS supports a metrics named `request_duration_seconds`. But this contains only `proxy_name` `to` and `rcode`.

I am requesting if there is a possibility to enable `name` and `type` as well with these metrics to be able to generate a better representation of what the `forward` workflow looks like. 

I am aware of the fact that this can increase the metrics cardinality. So, I am suggesting we create a new metrics entity and make that configurable via the `forward` plugin configuration.

```Corefile
forward . <dns> {
    qtype_for_metrics A AAAA ...
}
```
This can generate a new metrics `request_duration_seconds_with_name` which can have additional tag/label for `name` and `type`. 

**Why is this needed**:

We recently had a reported case of DNS queries for cluster internal Names leaking to upstream DNS.  This happened to be a case of some misconfiguration. Which we are working on preventing. 

When evaluating that, the topic of metrics came up and looking around, did not provide any available metrics with these informations. 

Having these metrics with Name and Type against the `forward` plugin can easily help isolate and identify the leaks happening in DNS Configuration and mitigate them at the source or with a suitable configuration easily. 

Below is a quick test patch I did to evaluate the feasibility of implementing this. Posting this here for the sake of discussion. 

```diff
commit 6170ab6835988e7b228e86e76421254d46a844cf
Author: Harsha Narayana <harshnar@cisco.com>
Date:   Sat Mar 8 18:31:41 2025 +0530

    enable query name and type as part of the metrics optionally

diff --git a/plugin/forward/forward.go b/plugin/forward/forward.go
index 3ea255161..4bb0fdbf6 100644
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -61,6 +61,8 @@ type Forward struct {
 	tapPlugins []*dnstap.Dnstap // when dnstap plugins are loaded, we use to this to send messages out.
 
 	Next plugin.Handler
+
+	qTypeForMetrics []string
 }
 
 // New returns a new Forward.
diff --git a/plugin/forward/setup.go b/plugin/forward/setup.go
index be5b12f94..6866e9dce 100644
--- a/plugin/forward/setup.go
+++ b/plugin/forward/setup.go
@@ -312,6 +312,12 @@ func parseBlock(c *caddy.Controller, f *Forward) error {
 			return c.ArgErr()
 		}
[...]

PR #7197, opened 2025-03-11
enabled query name and type as part of the metrics optionally

harshanarayana:
As part of this change, we are introducing a new configuration for the `forward` plugin named `qtype_for_metrics` that can be used to configure the query types that needs to be generated into metrics which contains the additional information of `name` and `type`.

The new metrics will be `request_duration_seconds_with_name` and will contain `"proxy_name", "to", "rcode", "name", "type"` as the labels/tags in the metrics.

#7196 

1. Document on the `forward` plugin needs to be updated to contain the additional details. https://github.com/coredns/coredns.io/blob/master/content/plugins/forward.md

No

### DIFF
--- a/plugin/forward/forward.go
+++ b/plugin/forward/forward.go
@@ -61,6 +61,8 @@ type Forward struct {
 	tapPlugins []*dnstap.Dnstap // when dnstap plugins are loaded, we use to this to send messages out.
 
 	Next plugin.Handler
+
+	qTypeForMetrics []string
 }
 
 // New returns a new Forward.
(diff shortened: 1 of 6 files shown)


Shmillerov:
Hello. I'm not a maintainer, but I have a couple of thoughts on this.

Here you add "name" (DNS domain name) as a label to the metric. In this case, each unique DNS query will be a separate row in `/metrics`.

I don't think this is a good idea. It will cause performance issues on the CoreDNS and metrics collector side.

You can read https://prometheus.io/docs/practices/naming/#labels **CAUTION** for more details.

PR #7249, opened 2025-04-12
fix(plugin/secondary): make transfer property mandatory

wafuwafu13:

- Fix this TODO.
https://github.com/coredns/coredns/blob/52b3172b2efffc7e662caf3cb1bb2b0c8743839f/plugin/secondary/setup_test.go#L18
- Add test coverage for the following error.
https://github.com/coredns/coredns/blob/52b3172b2efffc7e662caf3cb1bb2b0c8743839f/plugin/secondary/setup.go#L86

None

Already documented
https://coredns.io/plugins/secondary/
>A working syntax would be:
>secondary [zones...] {
>   transfer from ADDRESS [ADDRESS...]
> }

No. This change enforces that secondary zones must have a `transfer from` property, which was already an implicit requirement and is documented in the plugin documentation. 
The change makes this requirement explicit in the code.


plugin/secondary/setup_test.go
```diff
@@ -35,6 +29,20 @@ func TestSecondaryParse(t *testing.T) {
...
 		},
+		{
+			`secondary`,
+			true,
+			"",
+			nil,
+		},
+		{
+			`secondary example.org {
+				transferr from 127.0.0.1
```
wafuwafu13, line 40: Intentionally made a typo.

chrisohaver (approved):

PR #7255, opened 2025-04-15
Add optional TLS support to /metrics endpoint

peppi-lotta:
## 1. Why is this pull request needed?

It adds optional TLS support to the CoreDNS metrics endpoint, allowing metrics to be served over HTTPS. This improves security by encrypting metrics traffic. TLS is disabled by default, so existing configurations continue to work.

## 2. Related issues

#7109

## 3. Documentation changes

- Update metrics plugin docs to include TLS settings and examples
- Add instructions for providing certificates and configuring Prometheus to scrape over HTTPS
- Add troubleshooting notes

## 4. Backward compatibility

No breaking changes. TLS is optional and off by default.


peppi-lotta:
@chrisohaver @SuperQ Would you have time to review? :)

kashifest:
@chrisohaver @SuperQ can you please take a look on this PR? It would be nice to have this PR moving forward

kashifest:
It would be really nice to have someone review this and start the discussion rolling

amshankaran:
It would be helpful if someone could take a look at this.

plugin/metrics/metrics.go
```diff
@@ -97,14 +249,84 @@ func (m *Metrics) OnStartup() error {
...
+	if err != nil {
+		return fmt.Errorf("failed to create temporary TLS config file: %w", err)
+	}
+
+	minVersion, err := tlsVersionToString(m.tlsConfig.minVersion)
+	if err != nil {
+		return fmt.Errorf("failed to convert TLS version to string: %w", err)
+	}
+
+	yamlConfig := fmt.Sprintf(`
```
SuperQ: Oof, I'm really not a fan of having this write out a templated yaml file like this.

Maybe we first need to refactor the exporter-toolkit slightly so that you can pass the `web.Config` struct directly instead of having to use the config file only.

peppi-lotta: Thank you for commenting, I really appreciate it! Is this a blocker for this PR? If so, I'm willing to provide the refactoring PR in exporter-toolkit if needed :)

kashifest: Would the current way be acceptable as an interim solution so as to secure the metrics endpoint with TLS and meanwhile we improve the exporter-toolkit to include the `web.Config` as a Struct as suggested and eventually use that here?

peppi-lotta: @SuperQ Is this as acceptable as an interim solution? Can this get merged?

peppi-lotta:
@jameshartig @miekg @superq @greenpau @Tantalor93 Can I get a review on this? What are the next steps needed to get this merged?

miekg:
I consider TLS on metric endpoints to be an anti feature

JanMkl:
> I consider TLS on metric endpoints to be an anti feature

I understand the desire to minimise complexity—especially on internal metrics endpoints—but in a zero-trust architecture, all network traffic, regardless of source or function, must be both authenticated and encrypted.

Tampering with metrics may seem low-impact, but it can lead to unpredictable or harmful behavior. For example:
- It can reveal service topology and workload characteristics.
- It can trigger alerts or automated remediation based on falsified data.

In multi-tenant or regulated environments where the network is not implicitly trusted, TLS is not an anti-feature—it’s a baseline security control. Even when the content isn’t sensitive, preserving integrity and controlling access are essential.

To be clear, the PR does not propose mandatory TLS—it introduces optional support, as indicated in the title: “Add optional TLS support to /metrics endpoint”. Rejecting TLS outright disregards the realities of deployment models built on zero-trust principles.

peppi-lotta:
I've simplified this TLS implementation quite a bit. It's still using the Prometheus exporter-toolkit but now in the intended way. There is no more creating any temp files (this was the concern brought up in this comment: https://github.com/coredns/coredns/pull/7255#discussion_r2073357522). 

@jameshartig @miekg @SuperQ @greenpau @Tantalor93 Can I get a review on this? What are the next steps needed to get this merged?

kashifest (approved):
lgtm, it looks ok. It would nice to have someone with review rights have a look into this.

kashifest:
@johnbelamaric would you please take a look at this one?

yongtang:
@SuperQ any additional feedback on the updated PR?

kashifest:
@johnbelamaric @SuperQ Can we please decide how to proceed with this PR?

johnbelamaric:
@peppi-lotta can you rebase?

peppi-lotta:
@johnbelamaric I've rebased now :)

plugin/metrics/metrics.go
```diff
@@ -34,15 +38,18 @@ type Metrics struct {
...
 	met := &Metrics{
-		Addr:    addr,
-		Reg:     prometheus.DefaultRegisterer.(*prometheus.Registry),
-		zoneMap: make(map[string]struct{}),
-		plugins: pluginList(caddy.ListPlugins()),
+		Addr:          addr,
+		Reg:           prometheus.DefaultRegisterer.(*prometheus.Registry),
+		zoneMap:       make(map[string]struct{}),
+		plugins:       pluginList(caddy.ListPlugins()),
+		tlsConfigPath: "",
```
johnbelamaric: I don't think we need this, it will already be the empty string

plugin/metrics/metrics.go
```diff
@@ -99,17 +106,60 @@ func (m *Metrics) OnStartup() error {
...
+			startResult <- err
+		}
+		// If we get here without error, server is running
 	}()
 
+	// Wait for startup errors
+	select {
+	case err := <-startResult:
+		return err
+	case <-time.After(200 * time.Millisecond):
```
johnbelamaric: I don't love this. Is there instead a status or something we can check on the control structure?

plugin/metrics/metrics.go
```diff
@@ -99,17 +106,60 @@ func (m *Metrics) OnStartup() error {
...
+	}
+
+	// Create web config for ListenAndServe
+	webConfig := &web.FlagConfig{
+		WebListenAddresses: &[]string{m.Addr},
+		WebSystemdSocket:   new(bool), // false by default
+		WebConfigFile:      &m.tlsConfigPath,
+	}
+
+	logger := slog.New(slog.NewTextHandler(os.Stderr, nil))
```
johnbelamaric: I assume we need this particular object for the prometheus exporter? How does this interplay with our existing logging?

plugin/metrics/metrics.go
peppi-lotta: fixed

plugin/metrics/metrics.go
peppi-lotta: Yes the prometheus web toolkit requires a slog.Logger type object to be passed to it. I think it interplays nicely. There is not a lot of logging when the the TLS setup is successful as can be seen from bellow. 

```
$ kubectl logs -n kube-system          coredns-67ddfdcbd-6lhfq 
maxprocs: Leaving GOMAXPROCS=22: CPU quota undefined
time=2026-03-11T06:43:05.148Z level=INFO msg="Listening on" address=[::]:9153
time=2026-03-11T06:43:05.148Z level=INFO msg="TLS is enabled." http2=true address=[::]:9153
.:53
[INFO] plugin/reload: Running configuration SHA512 = 815c08e687bdc067019d64a774daf517a0fd3bb19531a7596741f57e78e3abb3dade4ae3bbe4898839f7bab7622c2bb91b27ce2f2c92d52b53e992dcd26ecbad
CoreDNS-1.14.2
linux/amd64, go1.25.5, 
```

plugin/metrics/metrics.go
terror96: Hi @johnbelamaric, joining the discussion.

I have been discussing this with @peppi-lotta, and I think there is a way to avoid waiting a predefined time for the service to be ready. I also agree that a timer-based approach can be problematic for various reasons. The `ServeTLS` method in `net/http`, which `web.Serve` ultimately calls, unfortunately does not provide a usable hook for checking service status. I don't think building an additional health-check mechanism that sends probe messages is a feasible solution either.

However, we know that `ServeTLS` essentially follows two paths: it either exits with an error or it calls `Accept`. `ServeTLS` creates a new listener but uses the listener argument to accept connections. Therefore, we can wrap the listener passed to `web.Serve` like this, e.g.:

```
// startupListener wraps a net.Listener to detect when Accept() is first called
type startupListener struct {
	readyOnce sync.Once
	ready     chan struct{}
	net.Listener
}

func newStartupListener(l net.Listener) *startupListener {
	return &startupListener{
		Listener: l,
		ready:    make(chan struct{}),
	}
}

func (sl *startupListener) Accept() (net.Conn, error) {
	// Signal ready on first Accept() call (server is running)
	sl.readyOnce.Do(func() {
		close(sl.ready)
	})
	return sl.Listener.Accept()
}

func (sl *startupListener) Ready() <-chan struct{} {
	return sl.ready
}

...
	ln, err := reuseport.Listen("tcp", m.Addr)
	m.ln := newStartupListener(ln)
...
	err := web.Serve(m.ln, server, webConfig, logger)
	// Wait for server to be ready (first Accept() called) or error
	select {
	case <-startupListener.Ready():
		log.Println("Server is ready and accepting connections")
	case err := <-serverErr:
		log.Fatalf("Server failed to start: %v", err)
	}

```
What do you think?

peppi-lotta: I pushed a commit implementing the suggested approach. Seems to work well :)

johnbelamaric:
thanks!

PR #7283, opened 2025-05-07
Drop RetryOnError from cache.Config

skitt:

This allows CoreDNS to build with Kubernetes 1.33 without requiring it. RetryOnError was removed in
https://github.com/kubernetes/kubernetes/pull/129568/commits/8e77ac000131019d5aa49c19aa1f477f6dac4d59 and defaults to false anyway.

None.

None.

No.


cbron:
Looks like this was fixed here: https://github.com/coredns/coredns/pull/7293

skitt:
> Looks like this was fixed here: #7293

Yes, shame this didn’t get any attention at the time ;-).

PR #7293, opened 2025-05-13
Prepare for the k8s api update

lyda:
Don't explicitly set `RetryOnError` to `false`.  It won't exist in the next version of the k8s api and it won't make a difference in this code since the struct would default to that. It also means that projects including codedns can build with Go 1.24.x.

This field is removed in the next release of the k8s client libraries. Currently we're not seeing them because this is built with go 1.23.x.

None - this is the default value of this field while it exists.

None.

No.

### DIFF
--- a/plugin/kubernetes/object/informer.go
+++ b/plugin/kubernetes/object/informer.go
@@ -17,7 +17,6 @@ func NewIndexerInformer(lw cache.ListerWatcher, objType runtime.Object, h cache.
 		ListerWatcher:    lw,
 		ObjectType:       objType,
 		FullResyncPeriod: defaultResyncPeriod,
-		RetryOnError:     false,
 		Process:          builder(clientState, h),
 	}
 	return clientState, cache.New(cfg)


SuperQ:
We already use dependabot to manage dependency updates.

go.mod
```diff
@@ -1,15 +1,15 @@
 module github.com/coredns/coredns
 
-go 1.23.0
+go 1.24.0
```
SuperQ: We follow upstream Go's release process. Go supports the last two Go releases, which we want to follow here.

lyda:
Ah...

```
go: k8s.io/api@v0.33.0 requires go@1.24.0, but 1.23.0 is requested
```

OK, reverting the `go.*` changes but leaving the other one. `false` is the zero value for `bool` so you don't need to set it. And when you do upgrade to 1.24.x, it will unlock the `k8s.io/api@v0.33.0` upgrade which will need this change. So might as well do it now to make dependabot's life easier.

SuperQ (approved):
LGTM, thanks

SuperQ:
Yea, it's really unfortunate [KEP-3744](https://github.com/kubernetes/enhancements/tree/master/keps/sig-release/3744-stay-on-supported-go-versions) misunderstand Go's compatibility and has them updating the minimum supported version long before it's out of support. This makes it difficult for downstrema users to upgraded to the latest K8s libraries while still supporting current Go versions.

I would argue with them, but honestly I don't have the time.

PR #7295, opened 2025-05-14
fix(plugin/bind): remove zone for link-local IPv4

mmickan:

On a host with an IPv4 link-local address assigned to an interface that CoreDNS is configured to bind to, this happens:

```
$ ./coredns -conf=/etc/coredns/Corefile
maxprocs: Leaving GOMAXPROCS=4: CPU quota undefined
lookup 169.254.1.1%dummy0: no such host
```

This PR fixes an incorrect test for an IPv6 address so it only adds the zone to IPv6 addresses, never to IPv4.

https://github.com/coredns/coredns/pull/6547 introduced the bug.

None, this makes it work the way the documentation already says it works. 😃 

No.

### DIFF
--- a/plugin/bind/setup.go
+++ b/plugin/bind/setup.go
@@ -85,7 +85,7 @@ func listIP(args []string, ifaces []net.Interface) ([]string, error) {
 					if ipnet, ok := addr.(*net.IPNet); ok {
 						ipa, err := net.ResolveIPAddr("ip", ipnet.IP.String())
 						if err == nil {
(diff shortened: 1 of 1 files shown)


yongtang:
@mmickan Would you mind fix the DCO issue?

https://github.com/coredns/coredns/pull/7295/checks?check_run_id=42175895880

mmickan:
Hi @yongtang ,

The DCO check is now passing and I've rebased on master while I was at it.

Thanks,
Mark

yongtang (approved):

PR #7310, opened 2025-05-20
feat: enable plugins via environment during build

dislogical:
This change will make it easier to automate builds using external plugins. For example, if I'm building my own image using `coredns-abc` and `coredns-def` plugins, I can just set `COREDNS_PLUGINS=abs:github.com/whatever/coredns-abc,def:github.com/something/coredns-def` when I initiate my build.

None.

Docs changes are not strictly necessary, but I've added a note to the README on how to use this feature.

Nope!

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,8 @@ $ cd coredns
 $ make
 ~~~
 
+> **_NOTE:_**  extra plugins may be enabled when building by setting the `COREDNS_PLUGINS` environment variable with comma separate list of plugins in the same format as plugin.cfg
+
 This should yield a `coredns` binary.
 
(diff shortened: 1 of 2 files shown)


dislogical:
> FAIL	github.com/coredns/coredns/test	600.093s

I'm willing to bet this is a timeout given the roundness of that time, is this a normal flake or should I take a look at fixing something?

thevilledev:
> is this a normal flake or should I take a look at fixing something?

A recurring issue yes. I filed an issue about it now and will have a look later this week: #7311 

Re-ran the failing tests and all green now. :)

yongtang (approved):

PR #7312, opened 2025-05-20
chore: update CODEOWNERS

thevilledev:

GitHub UI shows an error in the [current CODEOWNERS file](https://github.com/coredns/coredns/blob/master/CODEOWNERS) about users who no longer have write access to the repository. This PR removes those users from the file.

None.

None.

None.


SuperQ (approved):

yongtang:
@thevilledev @SuperQ  I think we can leave the names alone - we don't remove maintainers unless maintainers decide to step down themselves:

https://github.com/coredns/coredns/blob/master/GOVERNANCE.md#changes-in-maintainership

thevilledev:
Got it!

PR #7337, opened 2025-05-30
Keep golang to 1.24.2 due to build issues in 1.24.3

yongtang:
### 1. Why is this pull request needed and what does it do?
This PR keeep golang to 1.24.2 due to build issues in 1.24.3:

https://github.com/golang/go/issues/73617

This happens when make -f Makefile.release release is invoked in coredns.

go: downloading github.com/josharian/intern v1.0.0
go: downloading github.com/ryanuber/go-glob v1.0.0
# github.com/coredns/coredns
link: duplicated definition of symbol dlopen, from github.com/ebitengine/purego and github.com/ebitengine/purego
make[1]: *** [Makefile:20: coredns] Error 1
make[1]: Leaving directory '/go/src/github.com/coredns/coredns'
make: *** [Makefile.release:71: build] Error 2

### 2. Which issues (if any) are related?

N/A

### 3. Which documentation changes (if any) need to be made?
N/A
### 4. Does this introduce a backward incompatible change or deprecation?
N/A

PR #7338, opened 2025-05-30
Actually invoke make release -f Makefile.release during test

yongtang:

In previous setup, we do release test but only with dry-run:

This caused problems and we couldn't detect the 1.24.3's build issue (See #7337)

This PR actually test the release with make release -f Makefile.release.

N/A

N/A

N/A


thevilledev (approved):
LGTM

PR #7266, opened 2025-04-23
kubernetes: add multicluster support

MrFreezeex:
### 1. Why is this pull request needed and what does it do?
This adds the https://github.com/coredns/multicluster functionality which is a DNS implementation of MCS-API/KEP-1645 (https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/1645-multi-cluster-services-api/README.md#dns) in the kubernetes plugin. See https://github.com/coredns/coredns/issues/6921 for more details

This is an alternative PR compared to #7082 which was aiming to integrate the multicluster in-tree, this implementation focuses on integrating it directly in the kubernetes plugin and as such be able to reuse more code from the kubernetes plugin.

### 2. Which issues (if any) are related?
Fixes https://github.com/coredns/coredns/issues/6921

### 3. Which documentation changes (if any) need to be made?
Documentation about the new multicluster option

### 4. Does this introduce a backward incompatible change or deprecation?
no

PR #7350, opened 2025-06-03
fix(kubernetes): multicluster DNS race condition

thevilledev:

I was working on #7349 and I noticed this peculiar test failure on a completely unrelated test:

```
--- FAIL: TestMultiClusterHeadless (0.00s)
    controller_test.go:134: Expected SUCCESS, got NXDOMAIN
[coredns.local.]
[coredns.local. clusterset.local.]
```

So I started looking. The `HasSynced` method had incorrect variable assignments for the ServiceImport and MultiCluster Endpoints controllers. Both were incorrectly assigning their sync status to variable `c` instead of their respective variables `e` and `f`.

This caused a race condition where `HasSynced()` would return `true` even when all controllers weren't actually synced, leading to intermittent NXDOMAIN responses.

The fix ensures all controller sync states are properly checked before allowing DNS queries to proceed.

Originally implemented in #7266.

None.

None.

No.

### DIFF
--- a/plugin/kubernetes/controller.go
+++ b/plugin/kubernetes/controller.go
@@ -456,11 +456,11 @@ func (dns *dnsControl) HasSynced() bool {
 	d := dns.nsController.HasSynced()
 	e := true
 	if dns.svcImportController != nil {
-		c = dns.svcImportController.HasSynced()
+		e = dns.svcImportController.HasSynced()
 	}
 	f := true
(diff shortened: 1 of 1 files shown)


MrFreezeex (approved):
Woops sorry my bad, copy paste has gone a bit too wrong here :/, thanks a lot for fixing this ❤️

thevilledev:
Any time, happy to help!

PR #7381, opened 2025-06-17
fix(auto/file): return REFUSED when no next plugin is available

### LINKED ISSUE #7346: auto: Incorrect rcode returned when CNAME points to other zone - with proposed solutions
**What happened**:
When using the `file` or `auto` plugins inside a catch-all server block, queries for CNAMEs pointing to out-of-zone names return SERVFAIL due to CoreDNS' internal lookup of the CNAME target failing due to CoreDNS not being authoritative for the target.

Such queries should return NOERROR. Returning the rcode from the internal CNAME target resolution to the client is only correct for a recursive server, whereas the `file` and `auto` plugins act as authoritative servers and should return NOERROR so that the resolver can obtain the CNAME record and look up the target itself.

`Corefile`:
```
. {
    errors
    auto {
        directory /zones/
    }
}
```

`/zones/db.example.com`:
```
$ORIGIN example.com.
@	IN	SOA ns1.example.com. noc.example.com. 1 1 1 1 1

test	IN	CNAME example.org.
```

```
dig @127.0.0.1 test.example.com
...
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 30536
...
```

This came up before in #5104 and the workaround reached by @chrisohaver was to avoid using a catch-all server block, however that isn't really compatible with the deployment I have in mind, where the zones `auto` loads from disk may change over time.

I have two solutions in mind:
1. Tweak the `auto` and `file` plugins to explicitly return REFUSED (instead of letting `plugin.NextOrFailure` return SERVFAIL) when there is no next plugin configured. The code in the `file` plugin that does the CNAME target lookup already interprets REFUSED correctly so no further change there is needed.
2. Create a new tiny plugin called something like `finally` that can be configured to unconditionally return the desired error code(REFUSED in this case) as a last resort if no other plugin handles the request. This could also be useful for other catch-all setups that don't use the `auto` or `file` plugins.

I'm happy to implement either of these solutions as a PR, just looking for some guidance from the team as to what would be preferred and accepted.

I think I'm leaning towards option 1, as `file` and `auto` are by definition authoritative servers, and it is semantically correct for an authoritative server to return REFUSED when queried for zones it's not authoritative for.

**Environment**:

- the version of CoreDNS: 1.12.1
- Corefile: see above
- logs, if applicable:
    ```
    [ERROR] plugin/errors: 2 example.org. A: plugin/auto: no next plugin found
    ```
- OS (e.g: `cat /etc/os-release`): official docker container
- Others:

Tugzrida:
It is standard behaviour of authoritative DNS servers to return REFUSED when responding to a request they are not authoritative for. CoreDNS correctly follows this when responding to queries that don't match any server blocks, however when broader server blocks are present (such as a catch-all), queries not responded to by any plugin are typically responded to with SERVFAIL, as this is defined in `plugin.NextOrFailure()`.

By itself, this is not really an issue, however as described in #7346, this causes CoreDNS to return SERVFAIL when resolving out-of-bailiwick CNAME targets, which breaks these CNAMEs.

While it is difficult to map the traditional authoritative/resolver framework onto CoreDNS, for the case of the `auto` and `file` plugins, they fall more on the authoritative side, particularly when they are the last plugin in the chain.

Therefore, this PR adds logic to the `auto` and `file` plugins that, when they are the last plugin, they return REFUSED for unrecognised names, instead of SERVFAIL.

fixes #7346 

None, this corrects the behaviour of the examples already in the `auto` documentation

No

### DIFF
--- a/plugin/auto/auto.go
+++ b/plugin/auto/auto.go
@@ -51,6 +51,10 @@ func (a Auto) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (i
 	// Now the real zone.
 	zone = plugin.Zones(a.Zones.Names()).Matches(qname)
 	if zone == "" {
+		// If no next plugin is configured, it's more correct to return REFUSED as auto acts as an authoritative server
+		if a.Next == nil {
+			return dns.RcodeRefused, nil
+		}
 		return plugin.NextOrFailure(a.Name(), a.Next, ctx, w, r)
 	}
 
--- a/plugin/auto/auto_test.go
+++ b/plugin/auto/auto_test.go
@@ -104,39 +104,44 @@ func TestAutoServeDNSZoneMatching(t *testing.T) {
 	t.Parallel()
 
 	tests := []struct {
-		name    string
-		origins []string
-		names   []string
-		qname   string
-		hasZone bool
+		name         string
+		origins      []string
+		names        []string
+		qname        string
+		hasZone      bool
+		shouldRefuse bool
 	}{
 		{
-			name:    "exact zone match",
-			origins: []string{"example.org."},
-			names:   []string{"example.org."},
-			qname:   "test.example.org.",
-			hasZone: true,
+			name:         "exact zone match",
+			origins:      []string{"example.org."},
+			names:        []string{"example.org."},
+			qname:        "test.example.org.",
+			hasZone:      true,
+			shouldRefuse: false,
 		},
 		{
-			name:    "subdomain zone match",
-			origins: []string{"example.org."},
-			names:   []string{"example.org."},
-			qname:   "sub.test.example.org.",
-			hasZone: true,
+			name:         "subdomain zone match",
+			origins:      []string{"example.org."},
+			names:        []string{"example.org."},
+			qname:        "sub.test.example.org.",
+			hasZone:      true,
+			shouldRefuse: false,
 		},
 		{
-			name:    "no origin match",
-			origins: []string{"other.org."},
-			names:   []string{"example.org."},
-			qname:   "test.example.org.",
-			hasZone: false,
+			name:         "no origin match",
+			origins:      []string{"other.org."},
+			names:        []string{"example.org."},
+			qname:        "test.example.org.",
+			hasZone:      false,
+			shouldRefuse: false,
 		},
 		{
-			name:    "origin match but no name match",
-			origins: []string{"example.org."},
-			names:   []string{"other.org."},
-			qname:   "test.example.org.",
-			hasZone: false,
+			name:         "origin match but no name match",
+			origins:      []string{"example.org."},
+			names:        []string{"other.org."},
+			qname:        "test.example.org.",
+			hasZone:      false,
+			shouldRefuse: true,
 		},
 	}
 
@@ -163,14 +168,18 @@ func TestAutoServeDNSZoneMatching(t *testing.T) {
 			rec := dnstest.NewRecorder(&test.ResponseWriter{})
 			ctx := context.Background()
 
-			_, err := a.ServeDNS(ctx, rec, m)
+			code, err := a.ServeDNS(ctx, rec, m)
 
 			if tt.hasZone {
 				if err != nil {
 					t.Errorf("Expected no error for zone match, got: %v", err)
 				}
 			} else {
-				if err == nil {
+				if tt.shouldRefuse {
+					if code != dns.RcodeRefused {
+						t.Errorf("Expected code %d, got %d", dns.RcodeRefused, code)
+					}
+				} else if err == nil {
 					t.Errorf("Expected error for no zone match, got nil")
 				}
 			}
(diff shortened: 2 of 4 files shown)


Tugzrida:
Apologies, I didn't realise the existence of the e2e tests. Those fails should be fixed now.

yongtang (approved):

tsarna:
> By itself, this is not really an issue, however as described in https://github.com/coredns/coredns/issues/7346, this causes CoreDNS to return SERVFAIL when resolving out-of-bailiwick CNAME targets, which breaks these CNAMEs.

As far as I can see, this is still broken in 1.14.2:

zone example.com has:

xyz._domainkey IN CNAME xyz._domainkey.amazonses.com.

querying for TXT xyz._domainkey.example.com returns SERVFAIL with the CNAME in the answer.
querying for CNAME xyz._domainkey.example.com query returns the CNAME.

Tugzrida:
Hey @tsarna, what does your Corefile look like when you're seeing this behaviour?

SadDrummer:
I faced with same problem in CoreDNS 1.14.2

Corefile:

```
example.com {
    # Internal view
    errors
    log
    import config/split-horizon-internal.conf
    file zones/example.com.internal.zone
}

example.com. {
    # External view
    errors
    log
    file zones/example.com.external.zone
}
```

In zone files, both external and internal:
```
cdn             IN      CNAME          cdn.external.com.  
```
Yes dot at end of external cdn exist

Resolver behaviors:
Direct request to my DNS (from local network and global)
```
cdn.example.com	canonical name = cdn.external.com.
Name:	cdn.external.com
Address: <IP>
```

1.1.1.1:
```
;; Got SERVFAIL reply from 1.1.1.1
Server:		1.1.1.1
Address:	1.1.1.1#53

** server can't find cdn.example.com: SERVFAIL
```

8.8.8.8:
```
Server:		8.8.8.8
Address:	8.8.8.8#53

Non-authoritative answer:
cdn.example.com	canonical name = cdn.external.com.
Name:	cdn.external.com
Address: <IP>
```

Other external resolvers 50/50 get NOERROR with resolve or SERVFAIL

SadDrummer:
Additional info
CoreDNS logs:
```
Jun 26 08:27:21 server coredns[247701]: [INFO] <REDACTED>36577 - 20433 "A IN cdn.example.com. udp 46 true 1400" SERVFAIL qr,aa,cd 177 0.008681597s
Jun 26 08:27:21 server coredns[247701]: [INFO] <REDACTED>45996 - 19765 "HTTPS IN cdn.example.com. udp 46 true 1400" SERVFAIL qr,aa,cd 177 0.00946232s
Jun 26 08:27:21 server coredns[247701]: [INFO] <REDACTED>47466 - 44021 "HTTPS IN cdn.example.com. udp 46 true 1400" SERVFAIL qr,aa,cd 177 0.000047011s
Jun 26 08:27:21 server coredns[247701]: [INFO] <REDACTED>40723 - 56448 "A IN cdn.example.com. udp 35 false 512" SERVFAIL qr,aa,cd 177 0.008787609s
Jun 26 08:27:21 server coredns[247701]: [INFO] <REDACTED>34926 - 29331 "HTTPS IN cdn.example.com. udp 35 false 512" SERVFAIL qr,aa,cd 177 0.009293348s
Jun 26 08:27:48 server coredns[247701]: [INFO] <REDACTED>10319 - 29185 "A IN cdn.example.com. udp 46 true 1232" SERVFAIL qr,aa 177 0.008553715s
Jun 26 08:27:48 server coredns[247701]: [INFO] <REDACTED>8111 - 14980 "HTTPS IN cdn.example.com. udp 46 true 1232" SERVFAIL qr,aa 177 0.009348328s
Jun 26 08:27:57 server coredns[247701]: [INFO] <REDACTED>58273 - 940 "A IN cdn.example.com. udp 58 false 1232" NOERROR qr,aa,rd 255 0.018172438s
Jun 26 08:29:05 server coredns[247701]: [INFO] <REDACTED>42871 - 48934 "AAAA IN cdn.example.com. udp 58 false 1232" NOERROR qr,aa,rd 177 0.017048428s
Jun 26 08:29:10 server coredns[247701]: [INFO] <REDACTED>22359 - 10259 "AAAA IN cdn.example.com. udp 46 true 1232" SERVFAIL qr,aa 177 0.008508414s
Jun 26 08:29:10 server coredns[247701]: [INFO] <REDACTED>5633 - 2203 "A IN cdn.example.com. udp 46 true 1232" SERVFAIL qr,aa 177 0.009940097s
Jun 26 08:29:10 server coredns[247701]: [INFO] <REDACTED>11745 - 56035 "AAAA IN cdn.example.com. udp 46 true 1232" SERVFAIL qr,aa 177 0.000133502s
```

Seems like trouble with DNSSEC check, but "false - SERVFAIL" also spotted
`dnssec` plugin not enabled

And
Not all CNAME broken
More info about my internal zone file:

```
proxy               IN      A              10.10.10.10
some-project        IN      CNAME          proxy
cdn                 IN      CNAME          cdn.external.com.

```
Request for some-project.example.com work fine
For external zone same chain also work fine
cnd.example.com - SERVFAIL

PR #7390, opened 2025-06-27
chore: fix docs incompatibility

yquansah:
The CoreDNS current go.mod requires [go:1.23](https://github.com/coredns/coredns/blob/master/go.mod#L5). I ran the docker command in the README and it failed. So just contributing this change for a higher go version to run CoreDNS with docker.


thevilledev (approved):
Thanks!

PR #7402, opened 2025-07-09
plugin/file: preserve case in SRV record names and targets per RFC 6763

### LINKED ISSUE #7237: auto: Does not preserve case of SRV replies
**What happened**:

Reply for the SRV resource record squashed uppercase characters in the query

Reply should have preserved case in the query

Corefile:
```
. {
    log
    errors
    auto {
        directory ...
        reload 2s
    }
}
```

Zone:
```
$ORIGIN home.arpa.
$TTL 5
@ IN SOA @ nobody.invalid. 8 3600 1200 604800 5
     NS @
     AAAA ...
_smb._tcp PTR Home\ Media._smb._tcp
Home\ Media._smb._tcp SRV 0 0 445 samba
```

```shell
$ dig @... -t PTR "_smb._tcp.home.arpa."
;; QUESTION SECTION:
;_smb._tcp.home.arpa.		IN	PTR

;; ANSWER SECTION:
_smb._tcp.home.arpa.	5	IN	PTR	Home\032Media._smb._tcp.home.arpa.

$ dig @... -t SRV "Home Media._smb._tcp.home.arpa."
;; QUESTION SECTION:
;Home\032Media._smb._tcp.home.arpa. IN	SRV

;; ANSWER SECTION:
home\032media._smb._tcp.home.arpa. 5 IN	SRV	0 0 445 samba.home.arpa.
```

[RFC 6763, Section 4.1.1](https://datatracker.ietf.org/doc/html/rfc6763#section-4.1.1):
> The <Instance> portion of the Service Instance Name is a user-
   friendly name consisting of arbitrary Net-Unicode text [[RFC5198](https://datatracker.ietf.org/doc/html/rfc5198)].  It
   MUST NOT contain ASCII control characters (byte values 0x00-0x1F and
   0x7F) [[RFC20](https://datatracker.ietf.org/doc/html/rfc20)] but otherwise is allowed to contain any characters,
   without restriction, including spaces, uppercase, lowercase,
   punctuation -- including dots -- accented characters, non-Roman text,
   and anything else that may be represented using Net-Unicode.

**Environment**:

- the version of CoreDNS: 1.22

syedazeez337:
### What this PR does

This PR ensures SRV record names and targets preserve their original case in responses, aligning with [RFC 6763 §4.1.1](https://datatracker.ietf.org/doc/html/rfc6763#section-4.1.1), which requires that service instance names be treated as user-friendly Unicode text and must retain casing and formatting.

---

### What Changed

- ✅ SRV **target** is no longer lowercased in `Insert()` (e.g. `samba.home.arpa.`)
- ✅ SRV **record name** now bypasses forced lowercasing during insert
- ✅ `less()` in `plugin/file/tree/less.go` now compares labels **case-insensitively**
- ✅ Verified with manual test zone using `dig`

```dns
_smb._tcp                     PTR Home\032Media._smb._tcp.home.arpa.
Home\032Media._smb._tcp       SRV 0 0 445 samba.home.arpa.
````

Query:

```bash
dig @127.0.0.1 -p 53 SRV "Home\\032Media._smb._tcp.home.arpa."
```

Expected response:

```dns
Home\032Media._smb._tcp.home.arpa. 5 IN SRV 0 0 445 samba.home.arpa.
```

---

### Why this is needed

Although DNS is case-insensitive, [[RFC 6763](https://datatracker.ietf.org/doc/html/rfc6763#section-4.1.1)](https://datatracker.ietf.org/doc/html/rfc6763#section-4.1.1) mandates that **SRV record names used in service discovery retain original case and formatting**, especially for mDNS, Bonjour, and DNS-SD clients.

Prior to this fix, CoreDNS returned lowercased instance names, breaking spec-compliant discovery clients.

---

### Limitations

This PR does **not preserve the casing of the query name**, as it is already normalized (lowercased) by CoreDNS or `miekg/dns` before plugin processing. This fix ensures the **record data** (response) complies with the spec.

---

### Related

Fixes https://github.com/coredns/coredns/issues/7237

### DIFF
--- a/plugin/file/tree/less.go
+++ b/plugin/file/tree/less.go
@@ -2,6 +2,7 @@ package tree
 
 import (
 	"bytes"
+	"strings"
 
 	"github.com/miekg/dns"
 )
@@ -27,8 +28,8 @@ func less(a, b string) int {
 
 		// sadly this []byte will allocate... TODO(miek): check if this is needed
 		// for a name, otherwise compare the strings.
-		ab := []byte(a[ai:aj])
-		bb := []byte(b[bi:bj])
+		ab := []byte(strings.ToLower(a[ai:aj]))
+		bb := []byte(strings.ToLower(b[bi:bj]))
 		doDDD(ab)
 		doDDD(bb)
 
--- a/plugin/file/zone.go
+++ b/plugin/file/zone.go
@@ -73,7 +73,10 @@ func (z *Zone) CopyWithoutApex() *Zone {
 
 // Insert inserts r into z.
 func (z *Zone) Insert(r dns.RR) error {
-	r.Header().Name = strings.ToLower(r.Header().Name)
+	// r.Header().Name = strings.ToLower(r.Header().Name)
+	if r.Header().Rrtype != dns.TypeSRV {
+		r.Header().Name = strings.ToLower(r.Header().Name)
+	}
 
 	switch h := r.Header().Rrtype; h {
 	case dns.TypeNS:
@@ -108,7 +111,7 @@ func (z *Zone) Insert(r dns.RR) error {
 	case dns.TypeMX:
 		r.(*dns.MX).Mx = strings.ToLower(r.(*dns.MX).Mx)
 	case dns.TypeSRV:
-		r.(*dns.SRV).Target = strings.ToLower(r.(*dns.SRV).Target)
+		// r.(*dns.SRV).Target = strings.ToLower(r.(*dns.SRV).Target)
 	}
 
 	z.Tree.Insert(r)
--- a/plugin/file/zone_test.go
+++ b/plugin/file/zone_test.go
@@ -1,6 +1,12 @@
 package file
 
-import "testing"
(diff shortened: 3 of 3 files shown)


yongtang:
@syedazeez337 will it make sense to add a test?

yongtang:
@syedazeez337 will it make sense to add a test?

syedazeez337:
Hi @yongtang, I have added the test case for my code changes. Let me know if it rightly addresses the code guidelines of the organisation. Happy to contribute the changes needed. Thank you.
image
Also I have addressed the format issue that was failing it should pass now.

syedazeez337:
Screenshot 2025-07-12 142413
I have fixed the linting issues, it should pass now. Let me know if the logic checks out. Thank you.

syedazeez337:
Hi @yongtang, I have resolved all the issues that were failing on the github Checks. I will wait for your review on the logic. Thank you.

yongtang (approved):

PR #7413, opened 2025-07-15
plugin/test: fix TXT record comparison logic for multi-string vs multi-record scenarios

### LINKED ISSUE #7404: Confusion between multiple TXT records and TXT records longer than 255 bytes for the test plugin
**What happened**:

When testing multiple TXT records, the test plugin asserts wrongly the records after the first one on the value of the first one.

TXT records are correctly asserted.

Check https://github.com/k8s-gateway/k8s_gateway/pull/108

There seem to be a loop in https://github.com/coredns/coredns/blob/master/plugin/test/helpers.go#L209 that does not distinguish between 2 different TXT records and a TXT record split because it is longer than 255 bytes.

syedazeez337:
## Summary

This PR fixes a subtle but important bug in the `test` plugin of CoreDNS where TXT record assertions failed to correctly distinguish between:

- **Multiple distinct TXT records**, and  
- **A single TXT record split across multiple 255-byte strings** (as allowed by [RFC 1035 §3.3.14](https://datatracker.ietf.org/doc/html/rfc1035#section-3.3.14)).

The result was inaccurate test validations — either false positives or failures — especially in cases where long TXT records were involved.

---

## Problem Description

In DNS, a single TXT record may be split into multiple substrings due to the 255-byte limit. For example:

```dns
example.org. 3600 IN TXT "chunk1..." "chunk2..."
```

This is **one TXT record**, not two.

The `test` plugin previously compared TXT records using raw `[]string` slices from `dns.TXT.Txt`, like:

```go
for j, txt := range x.Txt {
    if txt != expected.Txt[j] {
        ...
    }
}
```

This caused incorrect results when:

- A single record was split across multiple strings (`len(Txt) > 1`), and
- The expected result consisted of multiple separate TXT RRs.

---

## Solution

The comparison logic has been updated to:

- Join the substrings using `strings.Join(...)`
- Compare the full logical string instead of chunk-by-chunk

### Example Fix:
```go
actualTxt := strings.Join(x.Txt, "")
expectedTxt := strings.Join(section[i].(*dns.TXT).Txt, "")
```

---

## Additional Changes

- ✅ Added a regression test: `plugin/test/testdata/txtrecordsplit.test`
  - Covers both split and multi-record TXT scenarios
- ✅ Created a stubbed `RunTestFile` in `testdata_test.go` to enable isolated testing
- ✅ Ensured the code passes `go fmt` and `golangci-lint` with no issues

---

## Impact

This improves the reliability of the `test` plugin when verifying TXT records, especially in cases involving long or split TXT strings. It ensures future plugin development or zone file testing behaves consistently with DNS standards.

---

## Related Issue

Fixes [#7404](https://github.com/coredns/coredns/issues/7404)


yongtang (approved):

micheelengronne:
Thank you @syedazeez337

PR #7415, opened 2025-07-18
plugin/proxy: add upstream_response_duration_seconds metric (#7382)

### LINKED ISSUE #7382: Add a pure upstream duration metric.

Add a more refined metric that represents the actual time taken by the upstream.

**Why is this needed**:

The metric `coredns_proxy_request_duration_seconds_bucket` includes the time spent during the sending phase of the message. We expect to add a more pure metric that starts timing after the request has been sent and ends when the upstream response is received. This would provide a clearer understanding of whether the upstream response is genuinely slow. The existing metric includes scenarios where CoreDNS is overloaded and the requests are blocked, which can skew the results. 

Therefore, we hope to add a new metric. thanks

syedazeez337:

This pull request introduces a new Prometheus metric `proxy_upstream_response_duration_seconds` to the `proxy` plugin.  
Unlike the existing `request_duration_seconds` metric, this new histogram measures the precise time taken between when a DNS request is sent to the upstream server and when a response is received — excluding any CoreDNS-internal delays such as scheduling, queueing, or write overhead.

This gives operators deeper insight into upstream latency and helps diagnose slow external DNS behavior more accurately.

Fixes: https://github.com/coredns/coredns/issues/7382

No breaking or user-visible documentation changes required.  
Metric is self-describing via Prometheus output.  
A note can optionally be added to plugin documentation or the release notes.

No. This is a backward-compatible enhancement that only adds a new metric. Existing metrics and behaviors are untouched.

### DIFF
--- a/plugin/pkg/proxy/connect.go
+++ b/plugin/pkg/proxy/connect.go
@@ -141,6 +141,8 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 		return nil, err
 	}
 
+	upstreamStart := time.Now()
+
 	var ret *dns.Msg
 	pc.c.SetReadDeadline(time.Now().Add(p.readTimeout))
 	for {
@@ -173,6 +175,7 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 			break
 		}
 	}
+
 	// recovery the origin Id after upstream.
 	ret.Id = originId
 
@@ -182,6 +185,8 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 	if !ok {
 		rc = strconv.Itoa(ret.Rcode)
 	}
+
+	upstreamResponseDuration.WithLabelValues(p.addr, proto, rc, p.proxyName).Observe(time.Since(upstreamStart).Seconds())
 
 	requestDuration.WithLabelValues(p.proxyName, p.addr, rc).Observe(time.Since(start).Seconds())
 
--- a/plugin/pkg/proxy/metrics.go
+++ b/plugin/pkg/proxy/metrics.go
@@ -38,4 +38,12 @@ var (
 		Name:      "conn_cache_misses_total",
 		Help:      "Counter of connection cache misses per upstream and protocol.",
 	}, []string{"proxy_name", "to", "proto"})
+
+	upstreamResponseDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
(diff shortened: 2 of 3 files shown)


plugin/pkg/proxy/connect.go
```diff
@@ -176,6 +178,9 @@ func (p *Proxy) Connect(ctx context.Context, state request.Request, opts Options
 			break
 		}
 	}
+
+	upstreamDuration := time.Since(upstreamStart)
```
SuperQ:
```suggestion
	upstreamResponseDuration.WithLabelValues(p.addr, proto, rc, p.proxyName).Observe(time.Since(upstreamStart).Seconds())
```

SuperQ (changes requested):
Rather than alloc a new value, just record the metric observation at the desired line.

plugin/pkg/proxy/connect.go
SuperQ: Oh, I see, this wants to use `rc`. I don't think the `if` statement is going to add that much latency. My first thought is that the alloc of `upstreamDuration` is going to be worse.

plugin/pkg/proxy/metrics.go
```diff
@@ -38,4 +38,12 @@ var (
...
 		Help:      "Counter of connection cache misses per upstream and protocol.",
 	}, []string{"proxy_name", "to", "proto"})
+
+	upstreamResponseDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "proxy",
+		Name:      "upstream_response_duration_seconds",
+		Buckets:   plugin.TimeBuckets,
+		Help:      "Histogram of time taken from sending a DNS request to receiving a response from the upstream (excluding CoreDNS delays).",
+	}, []string{"to", "protocol", "rcode", "server"})
```
SuperQ, line 48: This has a different label set than `requestDuration`. Can you describe why this should be different?

syedazeez337: Hi @SuperQ, thanks for the feedback.
I included the server label to distinguish metrics from different plugin blocks using the same upstream. This becomes useful in multi-zone setups, without it, metrics from distinct zones could get merged, leading to ambiguity.

Let me know if you'd prefer renaming the label or handling it differently.

Screenshot From 2025-07-19 14-44-52
Screenshot From 2025-07-19 14-45-11
Screenshot From 2025-07-19 14-45-47

PR #7428, opened 2025-07-28
chore: Minor changes to `Dockerfile`

polarathene:

- [`-qq` implies `-y`](https://manpages.ubuntu.com/manpages/lunar/man8/apt-get.8.html) for `apt-get`. Also there never was `-yy` (_introduced from [this PR](https://github.com/coredns/coredns/pull/5571), likely with other changes that aren't actually all that relevant_), only `-y`.
- `--no-install-recommends` was also added via [PR feedback request](https://github.com/coredns/coredns/pull/7428#discussion_r2241137962).
- Add a reference comment for better context of why `WORKDIR` is explicitly set to the default, as [this was a fix](https://github.com/coredns/coredns/pull/6731) for an [implicit change when the base image was swapped](https://github.com/coredns/coredns/pull/5969#issuecomment-1712368430) (_originally a breaking change for users of the image at the time_).

#6249

#7009 

None

No

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,8 @@ RUN export DEBCONF_NONINTERACTIVE_SEEN=true \
            DEBIAN_PRIORITY=critical \
            TERM=linux ; \
     apt-get -qq update ; \
-    apt-get -yyqq upgrade ; \
-    apt-get -yyqq install ca-certificates libcap2-bin; \
+    apt-get -qq upgrade ; \
+    apt-get -qq --no-install-recommends install ca-certificates libcap2-bin; \
     apt-get clean
 COPY coredns /coredns
 RUN setcap cap_net_bind_service=+ep /coredns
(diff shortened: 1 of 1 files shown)


SuperQ:

yongtang:
@polarathene can you fix the DCO issue?

Dockerfile
```diff
@@ -8,8 +8,8 @@ RUN export DEBCONF_NONINTERACTIVE_SEEN=true \
            DEBIAN_PRIORITY=critical \
            TERM=linux ; \
     apt-get -qq update ; \
-    apt-get -yyqq upgrade ; \
-    apt-get -yyqq install ca-certificates libcap2-bin; \
+    apt-get -qq upgrade ; \
+    apt-get -qq install ca-certificates libcap2-bin; \
```
mrueg:
```suggestion
    apt-get -qq --no-install-recommends install ca-certificates libcap2-bin; \
```
might be worth adding this option as well, then it won't install https://packages.debian.org/sid/libpam-cap

polarathene: Done 👍

polarathene:
> This needs a DCO sign-off. You can use `git commit -s --amend` to add it.

> @polarathene can you fix the DCO issue?

Sorted 👍

yongtang (approved):

PR #7431, opened 2025-08-01
Properly create hostname from IPv6

### LINKED ISSUE #7430: Invalid records for endpoint for IPv6 targets
**What happened**:

Core DNS can generate invalid records for endpoint for IPv6 targets.
We encountered cases where SRV queries on an endpoint would return invalid records if any of the pods had an IP ending with `::`.

For instance if a pod had an IP similar to this: `2001:db8:3333:4444:5555:6666::` the generated record would be `2001-db8-3333-4444-5555-6666--.endpoint.svc.namespace.cluster.local`. Such records are not valid because the hostname ends with `--`.

I was expecting the record to be valid.

Create a headless service.
Create a pod matching this service selector. Not sure how to force the IP address.
Query the service `dig -t SRV <service fqdn>`

The response will contain something simlar to this: `2001-db8-3333-4444-5555-6666--.endpoint.svc.namespace.cluster.local`.

This unit test would demonstrate the issue: https://github.com/coredns/coredns/pull/7431/files#diff-02e61cd1e31f52c578534b589f4e39f91472b2415234bddfdc1cd833f5270ff6R34

**Environment**:

- the version of CoreDNS: v1.11.4 but the issue appears to still be on master branch.
- Corefile:
- logs, if applicable:
- Others: only on IPv6 clusters

gjacquet:

Hostname for an IPV6 address can be invalid if the address ends with `::`.

Fixes: https://github.com/coredns/coredns/issues/7430

Generated hostname will be different if IP address ends with `::`.

### DIFF
--- a/plugin/kubernetes/kubernetes.go
+++ b/plugin/kubernetes/kubernetes.go
@@ -351,7 +351,11 @@ func endpointHostname(addr object.EndpointAddress, endpointNameMode bool) string
 		return strings.ReplaceAll(addr.IP, ".", "-")
 	}
 	if strings.Contains(addr.IP, ":") {
-		return strings.ReplaceAll(addr.IP, ":", "-")
+		ipv6Hostname := strings.ReplaceAll(addr.IP, ":", "-")
+		if strings.HasSuffix(ipv6Hostname, "-") {
+			return ipv6Hostname + "0"
+		}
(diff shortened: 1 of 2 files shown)


SuperQ (approved):
Thanks!

